PR #1875, opened 2016-06-07
Access Policy: Included Pseudo Statements for 'none' policy

### LINKED ISSUE #1738: mc access: incorrectly displaying success when 'None' policy is being set
Nesting conflict when setting the 'none' policy isn't being properly displayed in mc. Anytime a 'none' policy is set, it is always shown to be successful even when a nesting conflict occurs. 
```
$ mc access readonly myminio/mybucket/
Access permission for ‘myminio/mybucket/’ is set to ‘readonly’
$ mc access none myminio/mybucket/abc
Access permission for 'myminio/mybucket/abc' is set to 'none'
```

mkash32:
Fixes https://github.com/minio/mc/issues/1738
Pseudo Statements for 'none' are included to verify that 'none' is a valid policy for given prefix (returns error if there is nesting issue)
The Pseudo none statements are removed immediately after checking. In this way 'none' policy will be treated and checked the same as 'readonly', 'writeonly', and 'readwrite'.

Modifications on minio-go: https://github.com/minio/minio-go/pull/414

### DIFF
--- a/bucket-policy-parser.go
+++ b/bucket-policy-parser.go
@@ -33,6 +33,7 @@ const (
 )
 
 // supportedActionMap - lists all the actions supported by minio.
+// actions beginning with s3:None are pseudo statements used to prevent nesting. These statements are only encountered in Put Bucket Policy request and are promptly removed, they will never be saved in access-policy.
 var supportedActionMap = map[string]struct{}{
 	"s3:GetObject":                  {},
 	"s3:ListBucket":                 {},
@@ -42,6 +43,10 @@ var supportedActionMap = map[string]struct{}{
 	"s3:AbortMultipartUpload":       {},
 	"s3:ListBucketMultipartUploads": {},
 	"s3:ListMultipartUploadParts":   {},
+
+	// Pseudo Statements
+	"s3:None":                       {},
+	"s3:NoneBucket":                 {},
 }
 
 // supported Conditions type.
@@ -197,6 +202,7 @@ var invalidPrefixActions = map[string]struct{}{
 	"s3:GetBucketLocation":          {},
 	"s3:ListBucket":                 {},
 	"s3:ListBucketMultipartUploads": {},
+	"s3:NoneBucket":                 {},
 	// Add actions which do not honor prefixes.
 }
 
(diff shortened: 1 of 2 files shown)


mkash32:
@hackintoshrao Could you check if the fix looks proper? If it looks good I will add test cases.

bucket-policy-parser.go
```diff
@@ -33,6 +33,7 @@ const (
 )
 
 // supportedActionMap - lists all the actions supported by minio.
+// actions beginning with s3:None are pseudo statements used to prevent nesting. These statements are only encountered in Put Bucket Policy request and are promptly removed, they will never be saved in access-policy.
```
harshavardhana, line 36: Enter here to wrap the comment to the next line.

bucket-policy-parser.go
```diff
@@ -264,6 +270,28 @@ func checkBucketPolicyResources(bucket string, bucketPolicy BucketPolicy) APIErr
...
 }
 
+
+// removeNonePseudoStatements - removes the pseudo 'none' statements (s3:None and s3:NoneBucket) from bucket policy
+func removePseudoNoneStatements(bucketPolicy BucketPolicy) (statements []policyStatement, changed bool) {
+	// Swap all of the none statements to the beginning and return slice of the remaining portion
+	swapIndex := 0
+	for index, statement := range bucketPolicy.Statements {
+		for _, action := range statement.Actions {
+			if strings.HasPrefix(action, "s3:None") {
```
harshavardhana, line 280: Is 's3:None' a known common prefix ? - can we cite the relevant doc from s3 ?

mkash32, line 280: It isn't a common prefix, I was just using it so the server can check for any nesting conflicts when we are trying to set 'none' policy. Right after we [check for nesting](https://github.com/minio/minio/pull/1875/files#diff-4801f881a6fb2336ea40167d248cab0cR208), the [none statements are removed](https://github.com/minio/minio/pull/1875/files#diff-4801f881a6fb2336ea40167d248cab0cR214).

mkash32, line 280: I labelled them as "Pseudo statements". I'm not sure if this is the best way to solve the bug but the only other option is checking for nesting on the client side and that may be redundant since the server is already doing it. So I tried fitting 'none' into the server's nesting check by making a statement for it.
These statements are deleted directly after checking, they are never saved.

harshavardhana, line 280: > I labelled them as "Pseudo statements". I'm not sure if this is the best way to solve the bug but the only other option is checking for nesting on the client side and that may be redundant since the server is already doing it. So I tried fitting 'none' into the server's nesting check by making a statement for it.
> These statements are deleted directly after checking, they are never saved.

This isn't the best way to do it, we are introducing new fields which are not necessary and we are also making some custom policy actions.

harshavardhana:
> Pseudo Statements for 'none' are included to verify that 'none' is a valid policy for given prefix (returns error if there is nesting issue)
> The Pseudo none statements are removed immediately after checking. In this way 'none' policy will be treated and checked the same as 'readonly', 'writeonly', and 'readwrite'.

This won't be the right way to fix this, this would break s3 compatibility.  We shouldn't add reserved fields to fix a bug.

mkash32:
Oh darn, yeah it would break the compatibility. Closing this pr

PR #1876, opened 2016-06-07
XL: set write quorum (no. of disk / 2) + 2

balamurugana:
Previously write quorum was set to (no. of disk / 2) + 3.  As per new
change, the write quorum is set to (no. of disk / 2) + 2.


harshavardhana:
Can you add more rationale on this change? - `As per new change` doesn't add much, perhaps add some quorum matrix etc. 

We can capture it this time which was not captured before.

harshavardhana:
Can you rebase this @balamurugana ?

PR #1882, opened 2016-06-08
XL/FS: Returned error when content-length mismatches with size of part

### LINKED ISSUE #1868: XL: Incorrect behaviour for CompleteMultipartUpload.
Here is the test case to reproduce https://gist.github.com/hackintoshrao/17195ba97df5057ad4082ac8f2649532
## Expected behaviour

See the test. 
## Actual behaviour

See the test.

Run the test. 
## Minio version (paste output of `minio version`)

Latest master
## System information

OSX El Captain 64 bit. 

Here is the brief description of he issue. 
- A Part with the incorrect size parameter is inserted [here](https://gist.github.com/hackintoshrao/17195ba97df5057ad4082ac8f2649532#file-object-api-multipart_test-go-L50) . The actual size of the part is 4, but 6 is send as parameter. (Larger than the actual size). 
- Since there is no check on the actual number of bytes of data read from the reader and the size parameter, the wrong value will be inserted into the json configuration of the part. One solution of the solution could be to put a check inside `PutObjectPart` for mismatch between the size argument and the actual number of bytes read from the `ReadCloser`. Gotto check with the s3 standards that a Part with the incorrect size parameter can be inserted. 
- This leads to incorrect size stored for the part in the config. 
- An incorrect info about the size of  the part in the its json configuration is misleading. 
- Then a completePart array is formed using the malformed part [here](https://gist.github.com/hackintoshrao/17195ba97df5057ad4082ac8f2649532#file-object-api-multipart_test-go-L85) . 
- Then the malformed part is send as input for `CompleteMultipartUpload` [here](https://gist.github.com/hackintoshrao/17195ba97df5057ad4082ac8f2649532#file-object-api-multipart_test-go-L141). 

And when you run the test, FS fails with an `EOF` trying to read the part and surprisingly XL passes. 
Here is the result of the run.

```
=== RUN   TestObjectCompleteMultipartUpload
--- FAIL: TestObjectCompleteMultipartUpload (0.53s)
    object-api-multipart_test.go:1508: Test 1: SingleNode: Expected to fail with error "One or more of the specified parts could not be found", but instead failed with error "EOF"
    object-api-multipart_test.go:1503: Test 1: XL: Expected to fail with <ERROR> "One or more of the specified parts could not be found", but passed instead
FAIL
exit status 1

```

krisis:
- Fixed an existing test in object api test suite to supply correct content-length

Fixes #1868

### DIFF
--- a/fs-v1-multipart.go
+++ b/fs-v1-multipart.go
@@ -300,6 +300,7 @@ func (fs fsObjects) PutObjectPart(bucket, object, uploadID string, partID int, s
 	md5Writer := md5.New()
 
 	var buf = make([]byte, blockSizeV1)
+	var bytesRead int64
 	for {
 		n, err := io.ReadFull(data, buf)
 		if err == io.EOF {
@@ -308,6 +309,7 @@ func (fs fsObjects) PutObjectPart(bucket, object, uploadID string, partID int, s
 		if err != nil && err != io.ErrUnexpectedEOF {
 			return "", toObjectErr(err, bucket, object)
 		}
+		bytesRead += int64(n)
 		// Update md5 writer.
 		md5Writer.Write(buf[:n])
 		m, err := fs.storage.AppendFile(minioMetaBucket, tmpPartPath, buf[:n])
@@ -322,8 +324,15 @@ func (fs fsObjects) PutObjectPart(bucket, object, uploadID string, partID int, s
 	newMD5Hex := hex.EncodeToString(md5Writer.Sum(nil))
 	if md5Hex != "" {
 		if newMD5Hex != md5Hex {
+			fs.storage.DeleteFile(minioMetaBucket, tmpPartPath)
 			return "", BadDigest{md5Hex, newMD5Hex}
 		}
+	}
+
+	if size != bytesRead {
+		// content length doesn't match with part size, delete the part
+		fs.storage.DeleteFile(minioMetaBucket, tmpPartPath)
+		return "", errSignatureMismatch
 	}
 
 	// Hold write lock as we are updating fs.json
--- a/xl-v1-multipart.go
+++ b/xl-v1-multipart.go
@@ -346,9 +346,6 @@ func (xl xlObjects) putObjectPart(bucket string, object string, uploadID string,
 	if err != nil {
 		return "", toObjectErr(err, minioMetaBucket, tmpPartPath)
 	}
-	if size == -1 {
-		size = n
-	}
 	// Calculate new md5sum.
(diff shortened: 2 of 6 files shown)


fs-v1-multipart.go
```diff
@@ -322,10 +324,17 @@ func (fs fsObjects) PutObjectPart(bucket, object, uploadID string, partID int, s
 	newMD5Hex := hex.EncodeToString(md5Writer.Sum(nil))
 	if md5Hex != "" {
 		if newMD5Hex != md5Hex {
+			fs.storage.DeleteFile(minioMetaBucket, tmpPartPath)
 			return "", BadDigest{md5Hex, newMD5Hex}
 		}
 	}
 
+	if size != bytesRead {
+		// content length doesn't match with part size, delete the part
```
harshavardhana, line 333: '// content' should be capitalized as '// Content'

xl-v1-multipart.go
```diff
@@ -359,6 +356,11 @@ func (xl xlObjects) putObjectPart(bucket string, object string, uploadID string,
 			return "", BadDigest{md5Hex, newMD5Hex}
 		}
 	}
+	if size != n {
+		// content-length mismatch, delete the temporary object.
```
harshavardhana, line 360: Same here.

fs-v1-multipart.go
```diff
@@ -322,10 +324,17 @@ func (fs fsObjects) PutObjectPart(bucket, object, uploadID string, partID int, s
...
+		fs.storage.DeleteFile(minioMetaBucket, tmpPartPath)
+		return "", errSignatureMismatch
```
harshavardhana, line 335: Returning this error even in case of 'anonymous' requests is going to be wrong and unexpected which will happen since there is not real validation should this be returned in such a case. 

If you look at the flow the signature errors are not returned from this point on-wards.  Signature verification happens at object-handlers through a io.Pipe(). The reason to do that was signature verification should happen outside of the object API - since all the HTTP related details are not spilled over into this layer.

harshavardhana: Returning this error even in case of 'anonymous' requests is going to be wrong and unexpected which will happen since there is not real validation should this be returned in such a case. 

If you look at the flow the signature errors are not returned from this point on-wards.  Signature verification happens at object-handlers through a io.Pipe(). The reason to do that was signature verification should happen outside of the object API - since all the HTTP related details are not spilled over into this layer.

krisis:
Closing issue, ref: https://github.com/minio/minio/issues/1868#issuecomment-225499718

PR #1883, opened 2016-06-08
api: Add new bucket policy nesting error

mkash32:
Fixes https://github.com/minio/mc/issues/1706
- Added ErrPolicyNesting which is returned when nesting of policies has occured
- Replaces ErrMalformedPolicy in the case of nesting

### DIFF
--- a/api-errors.go
+++ b/api-errors.go
@@ -109,6 +109,7 @@ const (
 	ErrWriteQuorum
 	ErrStorageFull
 	ErrObjectExistsAsDirectory
+	ErrPolicyNesting
 )
 
 // error code to APIError structure, these fields carry respective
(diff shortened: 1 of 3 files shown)


harshavardhana:
You should rebase the changes rather than merging the branch, to avoid the merge commit.

mkash32:
My bad, I didn't run the tests locally. Found the problem.

api-errors.go
```diff
@@ -415,6 +416,11 @@ var errorCodeResponse = map[APIErrorCode]APIError{
 		Description:    "Multiple disks failures, unable to write data.",
 		HTTPStatusCode: http.StatusServiceUnavailable,
 	},
+	ErrPolicyNesting: {
+		Code:           "XMinioPolicyNesting",
+		Description:    "Policy nesting conflict has occurred.",
+		HTTPStatusCode: http.StatusBadRequest,
```
harshavardhana: We can return back 'http.StatusConflict'

PR #1887, opened 2016-06-09
posix: Support UNC paths on windows.

### LINKED ISSUE #1620: posix: Evaluate UNC path support.
https://blog.klauspost.com/long-windows-paths-unc-paths-in-go/

```
import "strings"
import "regexp"

// Pattern to match a windows absolute path: "c:\" and similar
var isAbsWinDrive = regexp.MustCompile(`^[a-zA-Z]\:\\`)

// UncPath converts an absolute Windows path
// to a UNC long path.
func UncPath(s string) string {
    // UNC can NOT use "/", so convert all to "\"
    s = strings.Replace(s, `/`, `\`, -1)

    // If prefix is "\\", we already have a UNC path or server.
    if strings.HasPrefix(s, `\\`) {

        // If already long path, just keep it
        if strings.HasPrefix(s, `\\?\`) {
            return s
        }

        // Trim "\\" from path and add UNC prefix.
        return "\\\\?\UNC\\" + strings.TrimPrefix(s, `\\`)
    }

    if isAbsWinDrive.MatchString(s) {
        return `\\?\` + s
    }

    return s
}
```

If it is any consolation, dot net users have the same issue - https://github.com/dotnet/corefx/issues/645

harshavardhana:
This allows us to now use 32K paths names on windows.

Fixes #1620

### DIFF
--- a/config-migrate.go
+++ b/config-migrate.go
@@ -49,7 +49,7 @@ func purgeV1() {
 		fatalIf(err, "Unable to retrieve config path.")
 
 		configFile := filepath.Join(configPath, "fsUsers.json")
-		os.RemoveAll(configFile)
+		removeAll(configFile)
 	}
 	fatalIf(errors.New(""), "Failed to migrate unrecognized config version ‘"+cv1.Version+"’.")
 }
--- a/object-api-getobjectinfo_test.go
+++ b/object-api-getobjectinfo_test.go
@@ -21,7 +21,6 @@ import (
 	"crypto/md5"
 	"encoding/hex"
 	"io/ioutil"
-	"os"
 	"strconv"
 	"testing"
 )
@@ -116,7 +115,7 @@ func BenchmarkGetObjectFS(b *testing.B) {
 	if err != nil {
 		b.Fatal(err)
 	}
-	defer os.RemoveAll(directory)
+	defer removeAll(directory)
 
 	// Create the obj.
 	obj, err := newFSObjects(directory)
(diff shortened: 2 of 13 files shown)


harshavardhana:
Unit tests have been updated as well - please review.

balamurugana:
Below are current observation
-  If a drive is given as disk (eg `D:` is given as disk) UNC path does not work to create directory tree ie its not possible to create `\\?\D:\mybucket`.  We get error `mkdir \\?: The filename, directory name or volume name syntax is incorrect.` 
  This check needs to be done.
- Each path segment length is 255 characters long.   We would need to add this check also.

harshavardhana:
> If a drive is given as disk (eg D: is given as disk) UNC path does not work to create directory tree ie its not possible to create \?\D:\mybucket. We get error mkdir \?: The filename, directory name or volume name syntax is incorrect. 
> This check needs to be done.

This is not a bug because of UNC but its a bug because of 'mkdirAll()' - there is a way to fix this. Will update the 'mkdirAll()' code.

harshavardhana:
> Each path segment length is 255 characters long. We would need to add this check also.

This can be added as well in checkPathLength.

harshavardhana:
> If a drive is given as disk (eg D: is given as disk) UNC path does not work to create directory tree ie its not possible to create \?\D:\mybucket. We get error mkdir \?: The filename, directory name or volume name syntax is incorrect. 
> This check needs to be done.

This is fixed, please validate.

harshavardhana:
> Each path segment length is 255 characters long. We would need to add this check also.
> This can be added as well in checkPathLength.

Existing code checkPathLength already does this no other fix is needed.

harshavardhana:
Is there anything else needed from this patch?

PR #1892, opened 2016-06-10
[WIP-donotmerge] XL - object healing.

krishnasrinivas:
Need discussion on objectAPI.HealObject return value:
i.e `func (xl xlObjects) HealObject(bucket, object string) error`

It does not return anything other than error.
It should return some information on the status of the object like:
- size
- number of disks on which it is available.

What other info is needed?
Should the response be in XML or JSON?
- JSON will be preferable for the webclient
- XML would mean consistency in the response format at the S3 layer
  
  probably we can do XML at S3 layer and JSON for webrpc.

### DIFF
--- a/erasure-healfile.go
+++ b/erasure-healfile.go
@@ -1,0 +1,100 @@
+/*
+ * Minio Cloud Storage, (C) 2016 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package main
+
+import "encoding/hex"
+
+// erasureHealFile heals part file of an object. It never writes the healed file to the actual
+// location, but to a temporary location ".minio/tmpPath" as specified by the caller. It will
+// be the responsibility of the caller to rename to the actual location.
+func erasureHealFile(latestDisks []StorageAPI, outDatedDisks []StorageAPI,
+	volume, // Volume of the object.
+	path, // Path to the part file.
+	partName, // Name of the part.
+	metaBucket, // .minio
+	tmpPath string, // Temporary file path where healed part will be written to.
+	size int64, // Size of the part file.
+	erasure erasureInfo) ([]checkSumInfo, error) {
+	var offset int64
+	remainingSize := size
+
+	// Sha512 hash for bitrot protection.
+	hashWriters := newHashWriters(len(outDatedDisks))
+
+	for remainingSize > 0 {
+		curBlockSize := erasure.BlockSize
+		if remainingSize < curBlockSize {
+			curBlockSize = remainingSize
+		}
+
+		// Calculate the block size that needs to be read from each disk.
+		curEncBlockSize := getEncodedBlockLen(curBlockSize, erasure.DataBlocks)
+
+		// Memory for reading data from disks and reconstructing missing data using erasure coding.
+		enBlocks := make([][]byte, len(latestDisks))
+
+		// Read data from the latest disks.
+		// FIXME: can be optimized to read from DataBlocks only if possible.
+		for index, disk := range latestDisks {
+			if disk == nil {
+				continue
+			}
+			blockIndex := erasure.Distribution[index] - 1
+			enBlocks[blockIndex] = make([]byte, curEncBlockSize)
+			_, err := disk.ReadFile(volume, path, offset,
+				enBlocks[blockIndex])
+			if err != nil {
+				enBlocks[blockIndex] = nil
+			}
+		}
+
+		// Reconstruct missing data.
+		err := decodeData(enBlocks, erasure.DataBlocks, erasure.ParityBlocks)
+		if err != nil {
+			return nil, err
+		}
+
+		// Write to the temporary heal file.
+		for index, disk := range outDatedDisks {
+			if disk == nil {
+				continue
+			}
+			blockIndex := erasure.Distribution[index] - 1
+			_, err := disk.AppendFile(metaBucket, tmpPath, enBlocks[blockIndex])
+			if err != nil {
+				return nil, err
+			}
+			hashWriters[index].Write(enBlocks[blockIndex])
+		}
+		remainingSize -= curBlockSize
+		offset += curEncBlockSize
+	}
+
+	// Sha512 checksums for the healed parts.
+	checkSums := make([]checkSumInfo, len(outDatedDisks))
+	for index, disk := range outDatedDisks {
+		if disk == nil {
+			continue
+		}
+		checkSums[index] = checkSumInfo{
+			Name:      partName,
+			Algorithm: "sha512",
+			Hash:      hex.EncodeToString(hashWriters[index].Sum(nil)),
+		}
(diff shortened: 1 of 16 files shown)


web-router.go
```diff
@@ -75,6 +75,7 @@ func registerWebRouter(mux *router.Router, web *webAPIHandlers) {
 	webBrowserRouter.Methods("POST").Path("/webrpc").Handler(webRPC)
 	webBrowserRouter.Methods("PUT").Path("/upload/{bucket}/{object:.+}").HandlerFunc(web.Upload)
 	webBrowserRouter.Methods("GET").Path("/download/{bucket}/{object:.+}").Queries("token", "{token:.*}").HandlerFunc(web.Download)
+	webBrowserRouter.Methods("GET").Path("/heal/{bucket}/{object:.+}").HandlerFunc(web.Heal)
```
harshavardhana: Any particular reason to keep it as a separate handler? why not use the '/webrpc' itself?

xl-v1-object.go
```diff
@@ -125,6 +125,115 @@ func (xl xlObjects) GetObject(bucket, object string, startOffset int64, length i
...
+		}
+	}
+
+	// Used for temp location.
+	tmpID := getUUID()
+	// Sha512 of the part files.
+	checkSumInfos := make([][]checkSumInfo, len(xl.storageDisks))
+
+	// Heal each part. erasureHealFile() will write the healed part to
+	// .minio/tmpID/ which needs to be renamed later.
```
harshavardhana: `.minio/tmp/uuid`

xl-v1-object.go
```diff
@@ -125,6 +125,115 @@ func (xl xlObjects) GetObject(bucket, object string, startOffset int64, length i
...
+		}
+		for index, sum := range checkSums {
+			if outDatedDisks[index] == nil {
+				continue
+			}
+			checkSumInfos[index] = append(checkSumInfos[index], sum)
+		}
+	}
+
+	// Write xl.json to all the outdate disks.
```
harshavardhana: `s/outdate/outdated/g`  or  `s/outdate/out of date/g`

harshavardhana:
Needs a unit test as well along with this change.

harshavardhana:
Another missing part here is Healing of object should also heal its buckets as well.

web-router.go
harshavardhana: Is there a reason to keep it like this ? why isn't webrpc sufficient for this use?

xl-v1-object.go
```diff
@@ -125,6 +125,116 @@ func (xl xlObjects) GetObject(bucket, object string, startOffset int64, length i
 	return nil
 }
 
+// HealObject - heal the object.
+// FIXME: If an object object was deleted and one disk was down, and later the disk comes back
+// up again, heal on the object should delete it.
```
harshavardhana, line 130: Can you open a bug for this to be fixed in future?

xl-v1-list-objects.go
```diff
@@ -26,7 +26,8 @@ func (xl xlObjects) listObjects(bucket, prefix, marker, delimiter string, maxKey
 		recursive = false
 	}
 
-	walkResultCh, endWalkCh := xl.listPool.Release(listParams{bucket, recursive, marker, prefix})
+	heal := false
```
harshavardhana, line 29: Needs a comment here.

xl-v1-multipart.go
```diff
@@ -81,7 +81,8 @@ func (xl xlObjects) listMultipartUploads(bucket, prefix, keyMarker, uploadIDMark
 	}
 	// Validate if we need to list further depending on maxUploads.
 	if maxUploads > 0 {
-		walkerCh, walkerDoneCh := xl.listPool.Release(listParams{minioMetaBucket, recursive, multipartMarkerPath, multipartPrefixPath})
+		heal := false
```
harshavardhana, line 84: same here.

tree-walk-heal-xl.go
```diff
@@ -22,13 +22,7 @@ import (
...
 
-// listDir - lists all the entries at a given prefix, takes additional params as filter and leaf detection.
-// filter is required to filter out the listed entries usually this function is supposed to return
-// true or false.
-// isLeaf is required to differentiate between directories and objects, this is a special requirement for XL
-// backend since objects are kept as directories, the only way to know if a directory is truly an object
-// we validate if 'xl.json' exists at the leaf. isLeaf replies true/false based on the outcome of a Stat
-// operation.
+// listDirHeal - Sorted union of entries from all the disks under prefixDir.
 func (xl xlObjects) listDirHeal(bucket, prefixDir string, filter func(entry string) bool) (mergedentries []string, err error) {
```
harshavardhana, line 26: 'mergedentries'  -> 'mergedEntries'

harshavardhana:
One more thing to do here is that we should send two separate PR's one for the HealObject() and then after that ListObjectsHeal() - don't see interdependence between both the API's.

harshavardhana:
> Should the response be in XML or JSON?
> 
> JSON will be preferable for the webclient
> XML would mean consistency in the response format at the S3 layer
> 
> probably we can do XML at S3 layer and JSON for webrpc.

There is no webrpc related change in this patch set - are you going to update this patch?

harshavardhana:
> One more thing to do here is that we should send two separate PR's one for the HealObject() and then after that ListObjectsHeal() - don't see interdependence between both the API's.

It would make it easier to implement unit tests for each of these APIs as well.

erasure-healfile.go
```diff
@@ -0,0 +1,100 @@
...
+	path, // Path to the part file.
+	partName, // Name of the part.
+	metaBucket, // .minio
+	tmpPath string, // Temporary file path where healed part will be written to.
+	size int64, // Size of the part file.
+	erasure erasureInfo) ([]checkSumInfo, error) {
+	var offset int64
+	remainingSize := size
+
+	// Sha512 hash for bitrot protection.
```
harshavardhana, line 35: '// Sha512 hash for bitrot protection' -> '// Sha512 hash for bit rot protection'

erasure-healfile.go
```diff
@@ -0,0 +1,100 @@
...
+
+		// Read data from the latest disks.
+		// FIXME: can be optimized to read from DataBlocks only if possible.
+		for index, disk := range latestDisks {
+			if disk == nil {
+				continue
+			}
+			blockIndex := erasure.Distribution[index] - 1
+			enBlocks[blockIndex] = make([]byte, curEncBlockSize)
+			_, err := disk.ReadFile(volume, path, offset,
```
harshavardhana, line 58: Add 'FIXME' here to support 'isValidBlock()' as well in future - otherwise disk.ReadFile would corrupt data blocks in such scenarios during decodeData().  

We could potentially heal in such scenarios but it is little tricky.

fs-v1.go
```diff
@@ -441,3 +441,7 @@ func (fs fsObjects) listObjects(bucket, prefix, marker, delimiter string, maxKey
 func (fs fsObjects) ListObjects(bucket, prefix, marker, delimiter string, maxKeys int) (ListObjectsInfo, error) {
 	return fs.listObjects(bucket, prefix, marker, delimiter, maxKeys)
 }
+
+func (fs fsObjects) Heal(bucket, prefix string, recursive bool) error {
+	return nil
```
harshavardhana, line 448: Add a comment on why its not implemented.

krishnasrinivas:
@harshavardhana will fix comments after release as this patch was decided not to be a blocker for the release. Main issue was, in the meeting, we could not figure out the best way to trigger the healing process and what the heal-response should look like.

harshavardhana:
> @harshavardhana will fix comments after release as this patch was decided not to be a blocker for the release. Main issue was, in the meeting, we could not figure out the best way to trigger the healing process and what the heal-response should look like.

Okay!

krishnasrinivas:
working on a different branch https://github.com/minio/minio/pull/2381
closing this

PR #1894, opened 2016-06-11
XL/format: Fix a bug in checkDisksConsistency.

harshavardhana:
Current checkDisksConsistency was not exhaustive enough check fix it 
and add relevant unit tests as well.

### DIFF
--- a/format-config-v1.go
+++ b/format-config-v1.go
@@ -224,10 +224,46 @@ func genericFormatCheck(formatConfigs []*formatConfigV1, sErrs []error) (err err
 	return nil
 }
 
+// errDiskOrderMismatch - returned when disk UUID is not in consistent JBOD order.
+var errDiskOrderMismatch = errors.New("disk order mismatch")
+
+// isSavedUUIDInOrder - validates if disk uuid is present and valid in all
+// available format config JBOD. This function also validates if the disk UUID
+// is always available on all JBOD under the same order.
+func isSavedUUIDInOrder(uuid string, formatConfigs []*formatConfigV1) bool {
+	var orderIndexes []int
+	// Validate each for format.json for relevant uuid.
+	for _, formatConfig := range formatConfigs {
+		if formatConfig == nil {
+			continue
+		}
+		// Validate if UUID is present in JBOD.
+		uuidIndex := findDiskIndex(uuid, formatConfig.XL.JBOD)
+		if uuidIndex == -1 {
+			// UUID not found.
+			errorIf(errDiskNotFound, "Disk %s not found in JBOD list", uuid)
+			return false
+		}
+		// Save the position of UUID present in JBOD.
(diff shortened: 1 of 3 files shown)


format-config-v1.go
```diff
@@ -224,10 +224,41 @@ func genericFormatCheck(formatConfigs []*formatConfigV1, sErrs []error) (err err
...
+	var orderIndexes []int
+	// Validate each for format.json for relevant uuid.
+	for _, formatConfig := range formatConfigs {
+		if formatConfig == nil {
+			continue
+		}
+		// Validate if UUID is present in JBOD.
+		uuidIndex := findDiskIndex(uuid, formatConfig.XL.JBOD)
+		if uuidIndex == -1 {
+			// UUID not found.
```
abperiasamy, line 243: We should log here. We have all the context to this error.

format-config-v1.go
```diff
@@ -224,10 +224,41 @@ func genericFormatCheck(formatConfigs []*formatConfigV1, sErrs []error) (err err
...
+		}
+		// Save the position of UUID present in JBOD.
+		orderIndexes = append(orderIndexes, uuidIndex+1)
+	}
+	// Once uuid is found, verify if the uuid
+	// present in same order across all format configs.
+	prevOrderIndex := orderIndexes[0]
+	for _, orderIndex := range orderIndexes {
+		if prevOrderIndex != orderIndex {
+			return false
```
abperiasamy, line 256: Here as well.

format-config-v1_test.go
```diff
@@ -0,0 +1,191 @@
...
+	for i, testCase := range testCases {
+		err := checkFormatXL(testCase.formatConfigs)
+		if err != nil && testCase.shouldPass {
+			t.Errorf("Test %d: Expected to pass but failed with %s", i+1, err)
+		}
+		if err == nil && !testCase.shouldPass {
+			t.Errorf("Test %d: Expected to fail but passed instead", i+1)
+		}
+	}
+}
```
abperiasamy, line 193: Instead of this style, it is better to write 1-1 unit tests. For example, isSavedUUIDInOrder will have isSavedUUIDInOrderTest. This test will only look out for variations in UUID order. This way, we will have a cleaner coverage for all functions. If we only test higher level functions, details may be lost. It is hard to remember all the tricky corner cases at higher level functions. Test functions for higher level functions will only test the glue code, how lower level functions are stitched together.

harshavardhana, line 193: > Instead of this style, it is better to write 1-1 unit tests. For example, isSavedUUIDInOrder will have isSavedUUIDInOrderTest. This test will only look out for variations in UUID order. This way, we will have a cleaner coverage for all functions. If we only test higher level functions, details may be lost. It is hard to remember all the tricky corner cases at higher level functions. Test functions for higher level functions will only test the glue code, how lower level functions are stitched together.

Okay this make sense.

PR #1896, opened 2016-06-13
FS: Adding format.json during FS initialization

### LINKED ISSUE #1696: fs: Add pending format.json

hackintoshrao:
fixes https://github.com/minio/minio/issues/1696 .

### DIFF
--- a/fs-v1-metadata.go
+++ b/fs-v1-metadata.go
@@ -7,7 +7,8 @@ import (
 )
 
 const (
-	fsMetaJSONFile = "fs.json"
+	fsMetaJSONFile   = "fs.json"
+	fsFormatJSONFile = "format.json"
 )
 
 // A fsMetaV1 represents a metadata header mapping keys to sets of values.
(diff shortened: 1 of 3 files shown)


fs-v1-metadata.go
```diff
@@ -78,6 +79,33 @@ func newFSMetaV1() (fsMeta fsMetaV1) {
 	return fsMeta
 }
 
+// newFSFormatV1 - initializes new formatConfigV1 with FS format info.
+func newFSFormatV1() (fsFormatMeta formatConfigV1) {
+	fsFormatMeta = formatConfigV1{}
```
harshavardhana: do not call `fsFormatMeta` and change the function. 

```
func newFSFormatV1() (format *formatConfigV1) {
return &formatConfigV1{
      Version: "1",
      Format: "fs",
      FS: &fsFormat{
             Version: "1",
      },
    }, nil 
} 
```

fs-v1.go
```diff
@@ -39,6 +39,11 @@ type fsObjects struct {
 	listObjectMapMutex *sync.Mutex
 }
 
+// creates format.json, the FS format info in minioMetaBucket.
+func createMetaFormatJSON(storageDisk StorageAPI) {
```
harshavardhana: createMetaFormatJSON -- initFormatFS() ..

fs-v1-metadata.go
```diff
@@ -78,6 +79,33 @@ func newFSMetaV1() (fsMeta fsMetaV1) {
...
+}
+
+// writes FS format (format.json) into minioMetaBucket.
+func writeFSFormatData(storage StorageAPI, fsFormat formatConfigV1) error {
+	metadataBytes, err := json.Marshal(fsFormat)
+	if err != nil {
+		return err
+	}
+	// fsFormatJSONFile - format.json file stored in minioMetaBucket(.minio) directory.
+	n, err := storage.AppendFile(minioMetaBucket, fsFormatJSONFile, metadataBytes)
```
harshavardhana, line 100: what happens when the file already exists? this will corrupt the previous `format.json`

fs-v1.go
```diff
@@ -48,6 +53,7 @@ func newFSObjects(disk string) (ObjectLayer, error) {
 
 	// Runs house keeping code, like creating minioMetaBucket, cleaning up tmp files etc.
 	fsHouseKeeping(storage)
+	createMetaFormatJSON(storage)
```
harshavardhana: Handle the errors and return. do not ignore it. It is okay for 'fsHouseKeeping()'

fs-v1.go
hackintoshrao: So, I'll Stat first on the file and write only if the file doesn't exist ?

fs-v1-metadata.go
hackintoshrao, line 100: So, I'll Stat first on the file and write only if the file doesn't exist ?

hackintoshrao:
@harshavardhana : Need to do code changes to handle the case of the format.json being already existing. Please review rest of the code.
Should I now stat for the format.json and write only if doesn't exist ?

harshavardhana:
Yes you also need to load it. Check xl-v1.go for more details. 

A simple pseudo code is as below.

```
err = loadFormatFS() 
If err == errUnformattedDisk {
    initFormatFS()
}
```

PR #1897, opened 2016-06-14
object layer: Send 200 OK and whitespace chars

### LINKED ISSUE #1866: fs: CompleteMultipartUpload should send white space characters till the operation is complete.
#### Expected behavior.

Server should send 200 OK and then continue to send white space chars till the operation is complete.

http://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadComplete.html

`Processing of a Complete Multipart Upload request could take several minutes to complete. After Amazon S3 begins processing the request, it sends an HTTP response header that specifies a 200 OK response. While processing is in progress, Amazon S3 periodically sends whitespace characters to keep the connection from timing out. Because a request could fail after the initial 200 OK response has been sent, it is important that you check the response body to determine whether the request succeeded. Not sure how server indicates a failure after 200 OK with white space characters, needs to be figured out.`
### Actual behaviour

Server does not send 200 OK immediately, it waits till concatenation of all the parts is complete which might take a long time.
#### Minio version

master

krisis:
... thereof until CompleteMultipartUpload finishes at the object
layer.
Ref: http://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadComplete.html

Closes #1866

### DIFF
--- a/object-handlers.go
+++ b/object-handlers.go
@@ -843,6 +843,8 @@ func (api objectAPIHandlers) AbortMultipartUploadHandler(w http.ResponseWriter, 
 	writeSuccessNoContent(w)
 }
 
+// Send whitespace character, once every 5secs, until CompleteMultipartUpload is done.
+// CompleteMultipartUpload method of the object layer indicates that it's done via doneCh
 func sendWhiteSpaceChars(w http.ResponseWriter, doneCh <-chan struct{}) {
 	for {
 		select {
@@ -962,12 +964,12 @@ func (api objectAPIHandlers) CompleteMultipartUploadHandler(w http.ResponseWrite
 	w.WriteHeader(http.StatusOK)
 
 	doneCh := make(chan struct{})
+	// Signal that completeMultipartUpload is over via doneCh
 	go func(doneCh chan<- struct{}) {
 		md5Sum, err = api.ObjectAPI.CompleteMultipartUpload(bucket, object, uploadID, completeParts)
 		doneCh <- struct{}{}
 	}(doneCh)
 
-	// Send whitespace character, once every 5secs, until CompleteMultipartUpload is done
 	sendWhiteSpaceChars(w, doneCh)
 
 	if err != nil {
--- a/object-handlers.go
+++ b/object-handlers.go
@@ -946,21 +946,40 @@ func (api objectAPIHandlers) CompleteMultipartUploadHandler(w http.ResponseWrite
 		completeParts = append(completeParts, part)
 	}
 	// Complete multipart upload.
+	// Send 200 OK
+	setCommonHeaders(w)
+	w.WriteHeader(http.StatusOK)
+
+	//Send whitespace character, once every 5secs, until CompleteMultipartUpload is done
+	done := make(chan bool)
+	go func() {
+		for {
+			select {
+			case <-time.After(5 * time.Second):
+				w.Write([]byte(" "))
+			case <-done:
+				return
+			}
+		}
+	}()
+
 	md5Sum, err = api.ObjectAPI.CompleteMultipartUpload(bucket, object, uploadID, completeParts)
 	if err != nil {
 		errorIf(err, "Unable to complete multipart upload.")
 		writeErrorResponse(w, r, toAPIErrorCode(err), r.URL.Path)
+		done <- true
 		return
 	}
+	done <- true
+
 	// Get object location.
 	location := getLocation(r)
 	// Generate complete multipart response.
 	response := generateCompleteMultpartUploadResponse(bucket, object, location, md5Sum)
 	encodedSuccessResponse := encodeResponse(response)
-	// Write headers.
-	setCommonHeaders(w)
 	// write success response.
-	writeSuccessResponse(w, encodedSuccessResponse)
+	w.Write(encodedSuccessResponse)
+	w.(http.Flusher).Flush()
 }
 
 /// Delete objectAPIHandlers
--- a/object-handlers.go
+++ b/object-handlers.go
@@ -843,6 +843,17 @@ func (api objectAPIHandlers) AbortMultipartUploadHandler(w http.ResponseWriter, 
 	writeSuccessNoContent(w)
 }
 
+func sendWhiteSpaceChars(w http.ResponseWriter, doneCh <-chan struct{}) {
+	for {
+		select {
+		case <-time.After(5 * time.Second):
+			w.Write([]byte(" "))
+		case <-doneCh:
+			return
+		}
+	}
+}
+
 // ListObjectPartsHandler - List object parts
 func (api objectAPIHandlers) ListObjectPartsHandler(w http.ResponseWriter, r *http.Request) {
 	vars := mux.Vars(r)
@@ -950,27 +961,20 @@ func (api objectAPIHandlers) CompleteMultipartUploadHandler(w http.ResponseWrite
 	setCommonHeaders(w)
 	w.WriteHeader(http.StatusOK)
 
-	//Send whitespace character, once every 5secs, until CompleteMultipartUpload is done
-	doneCh := make(chan bool)
-	go func(doneCh <-chan bool) {
-		for {
-			select {
-			case <-time.After(5 * time.Second):
-				w.Write([]byte(" "))
-			case <-doneCh:
-				return
-			}
-		}
+	doneCh := make(chan struct{})
+	go func(doneCh chan<- struct{}) {
+		md5Sum, err = api.ObjectAPI.CompleteMultipartUpload(bucket, object, uploadID, completeParts)
+		doneCh <- struct{}{}
 	}(doneCh)
 
-	md5Sum, err = api.ObjectAPI.CompleteMultipartUpload(bucket, object, uploadID, completeParts)
+	// Send whitespace character, once every 5secs, until CompleteMultipartUpload is done
+	sendWhiteSpaceChars(w, doneCh)
+
 	if err != nil {
 		errorIf(err, "Unable to complete multipart upload.")
-		doneCh <- true
 		writeErrorResponseNoHeader(w, r, getAPIError(toAPIErrorCode(err)), r.URL.Path)
 		return
 	}
-	doneCh <- true
 
 	// Get object location.
 	location := getLocation(r)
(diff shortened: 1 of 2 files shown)


object-handlers.go
```diff
@@ -946,21 +946,40 @@ func (api objectAPIHandlers) CompleteMultipartUploadHandler(w http.ResponseWrite
 		completeParts = append(completeParts, part)
 	}
 	// Complete multipart upload.
+	// Send 200 OK
+	setCommonHeaders(w)
+	w.WriteHeader(http.StatusOK)
+
+	//Send whitespace character, once every 5secs, until CompleteMultipartUpload is done
```
harshavardhana: Needs a 'space' after '//'

object-handlers.go
```diff
@@ -946,21 +946,40 @@ func (api objectAPIHandlers) CompleteMultipartUploadHandler(w http.ResponseWrite
 		completeParts = append(completeParts, part)
 	}
 	// Complete multipart upload.
+	// Send 200 OK
+	setCommonHeaders(w)
+	w.WriteHeader(http.StatusOK)
```
harshavardhana, line 964: Writing this status will cause a problem with the error response later, you cannot write HTTP status twice to HTTP response writer. 

Once you write a test case you will see the problem reproducing an error condition.

object-handlers.go
```diff
@@ -946,21 +946,40 @@ func (api objectAPIHandlers) CompleteMultipartUploadHandler(w http.ResponseWrite
...
+
+	//Send whitespace character, once every 5secs, until CompleteMultipartUpload is done
+	done := make(chan bool)
+	go func() {
+		for {
```
abperiasamy: pass doneCh as arg.

object-handlers.go
```diff
@@ -946,21 +946,40 @@ func (api objectAPIHandlers) CompleteMultipartUploadHandler(w http.ResponseWrite
 		completeParts = append(completeParts, part)
 	}
 	// Complete multipart upload.
+	// Send 200 OK
+	setCommonHeaders(w)
+	w.WriteHeader(http.StatusOK)
+
+	//Send whitespace character, once every 5secs, until CompleteMultipartUpload is done
+	done := make(chan bool)
+	go func() {
```
abperiasamy: name it as doneCh

object-handlers.go
```diff
@@ -946,21 +946,40 @@ func (api objectAPIHandlers) CompleteMultipartUploadHandler(w http.ResponseWrite
...
+				return
+			}
+		}
+	}()
+
 	md5Sum, err = api.ObjectAPI.CompleteMultipartUpload(bucket, object, uploadID, completeParts)
 	if err != nil {
 		errorIf(err, "Unable to complete multipart upload.")
 		writeErrorResponse(w, r, toAPIErrorCode(err), r.URL.Path)
+		done <- true
```
harshavardhana: We need some sort of API level change to accommodate this an API level change would be completeMultipartUpload should return a channel and we attempt a read on that channel will unblock itself once the completeMultipartUpload is complete or error out.

object-handlers.go
krisis, line 964: @harshavardhana you are right, I shouldn't be using `writeErrorResponse` in the `if-err-not-nil-block`. As per AWS docs, it is OK to send `200 OK` and send error as part of the body.

object-handlers.go
harshavardhana: With this approach we can safely write http.StatusOK and also white space characters.

abperiasamy: doneCh <- true should be above writeErrorResponse

harshavardhana:
> @harshavardhana you are right, I shouldn't be using writeErrorResponse in the if-err-not-nil-block. As per AWS docs, it is OK to send 200 OK and send error as part of the body.

Can you show the documentation on this?  it would mean all our client libs have to be changed for CompleteMultipartUpload to look for the response body instead of http.StatusOK.  AFAIK this style was only true with CopyObject API. 

Interestingly i do not see any special handling in any of the aws-sdk's related to this whitespace characters - so i wonder if the documentation is correct or uptodate.  Have we seen S3 send white space characters other than the documentation? - Since we have transferred lots of data to S3 we have never encountered and issue based on this.

krishnasrinivas:
@harshavardhana 
http://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadComplete.html

```
HTTP/1.1 200 OK
x-amz-id-2: Uuag1LuByRx9e6j5Onimru9pO4ZVKnJ2Qz7/C1NPcfTWAtRPfTaOFg==
x-amz-request-id: 656c76696e6727732072657175657374
Date:  Mon, 1 Nov 2010 20:34:56 GMT
Connection: close
Server: AmazonS3

<Error>
  <Code>InternalError</Code>
  <Message>We encountered an internal error. Please try again.</Message>
  <RequestId>656c76696e6727732072657175657374</RequestId>
  <HostId>Uuag1LuByRx9e6j5Onimru9pO4ZVKnJ2Qz7/C1NPcfTWAtRPfTaOFg==</HostId>
</Error>
```

we haven't tested if S3 sends whitespace characters.
Also in client libs we actually don't need any changes to account for whitespace characters because XML parser will ignore the white spaces (?)

krisis:
Tested the PR with,
- `mc cp <large-file> <alias>/bucket/object`
- `mc mirror $GOPATH/src/github.com/minio-go <alias>/bucket/gosrc`

Both tests ran to successful completion

object-handlers.go
```diff
@@ -951,26 +951,26 @@ func (api objectAPIHandlers) CompleteMultipartUploadHandler(w http.ResponseWrite
...
 
 	//Send whitespace character, once every 5secs, until CompleteMultipartUpload is done
-	done := make(chan bool)
-	go func() {
+	doneCh := make(chan bool)
+	go func(doneCh <-chan bool) {
 		for {
 			select {
 			case <-time.After(5 * time.Second):
 				w.Write([]byte(" "))
```
krishnasrinivas: There is a race condition where w.Write() here can happen after writeErrorResponseNoHeader()

krishnasrinivas: you can do this i think:
go func() {
    md5Sum, err = api.ObjectAPI.CompleteMultipartUpload(bucket, object, uploadID, completeParts)
    doneCh <- struct{}{}
}

for {
select{
    case <-time.After(5 * time.Second): w.Write([]byte(" "))
    case <-doneCh:break
}
}
if err != nil {
    writeErrorResponseNoHeader()
    return
}

object-handlers.go
```diff
@@ -946,21 +946,40 @@ func (api objectAPIHandlers) CompleteMultipartUploadHandler(w http.ResponseWrite
...
+	setCommonHeaders(w)
+	w.WriteHeader(http.StatusOK)
+
+	//Send whitespace character, once every 5secs, until CompleteMultipartUpload is done
+	doneCh := make(chan bool)
+	go func(doneCh <-chan bool) {
+		for {
+			select {
+			case <-time.After(5 * time.Second):
+				w.Write([]byte(" "))
```
krishnasrinivas: There is a race condition where w.Write() here can happen after writeErrorResponseNoHeader()

you can do this i think:

```
go func() {
    md5Sum, err = api.ObjectAPI.CompleteMultipartUpload(bucket, object, uploadID, completeParts)
    doneCh <- struct{}{}
}()

for {
select{
    case <-time.After(5 * time.Second): w.Write([]byte(" "))
    case <-doneCh:break
}
}
if err != nil {
    writeErrorResponseNoHeader()
    return
}
```

harshavardhana:
> @harshavardhana 
> http://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadComplete.html
> 
> HTTP/1.1 200 OK
> x-amz-id-2: Uuag1LuByRx9e6j5Onimru9pO4ZVKnJ2Qz7/C1NPcfTWAtRPfTaOFg==
> x-amz-request-id: 656c76696e6727732072657175657374
> Date:  Mon, 1 Nov 2010 20:34:56 GMT
> Connection: close
> Server: AmazonS3
> 
> <?xml version="1.0" encoding="UTF-8"?>
> 
> <Error>
>   <Code>InternalError</Code>
>   <Message>We encountered an internal error. Please try again.</Message>
>   <RequestId>656c76696e6727732072657175657374</RequestId>
>   <HostId>Uuag1LuByRx9e6j5Onimru9pO4ZVKnJ2Qz7/C1NPcfTWAtRPfTaOFg==</HostId>
> </Error>
> we haven't tested if S3 sends whitespace characters.
> Also in client libs we actually don't need any changes to account for whitespace characters because XML parser will ignore the white spaces (?)

Talked at length with @krisis on this,  looks like our client libs have to handle it and not look for http.StatusOK for CompleteMultipartUpload instead read the response body.

object-handlers.go
krisis: Good catch! Thanks. I shall change that and update the PR.

object-handlers.go
```diff
@@ -843,6 +843,17 @@ func (api objectAPIHandlers) AbortMultipartUploadHandler(w http.ResponseWriter,
 	writeSuccessNoContent(w)
 }
 
+func sendWhiteSpaceChars(w http.ResponseWriter, doneCh <-chan struct{}) {
```
harshavardhana, line 848: Needs a comment for this function.

object-handlers.go
```diff
@@ -950,27 +961,20 @@ func (api objectAPIHandlers) CompleteMultipartUploadHandler(w http.ResponseWrite
...
-		for {
-			select {
-			case <-time.After(5 * time.Second):
-				w.Write([]byte(" "))
-			case <-doneCh:
-				return
-			}
-		}
+	doneCh := make(chan struct{})
+	go func(doneCh chan<- struct{}) {
```
harshavardhana, line 968: Needs a comment for this 'go-routine'

object-handlers.go
```diff
@@ -951,26 +951,26 @@ func (api objectAPIHandlers) CompleteMultipartUploadHandler(w http.ResponseWrite
...
 				w.Write([]byte(" "))
-			case <-done:
+			case <-doneCh:
 				return
 			}
 		}
-	}()
+	}(doneCh)
 
 	md5Sum, err = api.ObjectAPI.CompleteMultipartUpload(bucket, object, uploadID, completeParts)
```
harshavardhana: Move the doneCh <- true - right before 'if err != nil'

krishnasrinivas: @harshavardhana there is another commit on top of this which supersedes this commit

PR #1901, opened 2016-06-14
posix: RenameFile() should remove empty source parent directories aft…

### LINKED ISSUE #1900: posix: RenameFile should remove empty parent directories
posix.RenameFile() should remove empty parent directories just like posix.DeleteFile() does.

krishnasrinivas:
…er rename.

fixes #1900

### DIFF
--- a/posix.go
+++ b/posix.go
@@ -695,5 +695,7 @@ func (s posix) RenameFile(srcVolume, srcPath, dstVolume, dstPath string) (err er
 		}
 		return err
 	}
-	return nil
+	srcParentDir := slashpath.Join(srcVolumeDir, slashpath.Dir(srcPath))
+	// Remove the parent directories of the source if they are empty.
+	return deleteFile(srcVolumeDir, srcParentDir)
 }


balamurugana:
@krishnasrinivas removing directory tree after renaming causes multipart upload id removal ie it removes .minio/tmp/UUID directory tree. This makes multipart related calls failure.

krishnasrinivas:
> @krishnasrinivas removing directory tree after renaming causes multipart upload id removal ie it removes .minio/tmp/UUID directory tree. This makes multipart related calls failure.

@balamurugana only empty directories should be removed (just like how we remove directories during DeleteFile)  ... I haven't digged into the code flow of why multipart related calls fail

harshavardhana:
> @balamurugana only empty directories should be removed (just like how we remove directories during DeleteFile) ... I haven't digged into the code flow of why multipart related calls fail

There is a small window where the aforementioned directory is indeed empty as in UPLOADID directory for multipart in progress, that is the reason the tests are failing.

posix.go
```diff
@@ -695,5 +695,7 @@ func (s posix) RenameFile(srcVolume, srcPath, dstVolume, dstPath string) (err er
 		}
 		return err
 	}
-	return nil
+	srcParentDir := slashpath.Join(srcVolumeDir, slashpath.Dir(srcPath))
+	// Remove the parent directories of the source if they are empty.
+	return deleteFile(srcVolumeDir, srcParentDir)
```
balamurugana, line 700: The right way to address this issue is like below

```
    // Remove stale directories.
    srcParentDir := slashpath.Join(srcVolumeDir, slashpath.Dir(srcPath))
    srcBaseDir := srcVolumeDir
    // tmp directory should not be removed.
    if srcPath == "tmp" || strings.HasPrefix(srcPath, "tmp/") {
        srcBaseDir = slashpath.Join(srcBaseDir, "tmp")
    }

    // Remove the parent directories of the source if they are empty.
    err = deleteFile(srcBaseDir, srcParentDir)
    if err != nil {
        // As the actual rename succeeded, log the error.
        errorIf(err, "failed to delete directory", srcBaseDir, srcParentDir)
    }

    return nil

```

However unit test also fails due to missing upload id directory.

balamurugana:
https://github.com/minio/minio/pull/2007 is a discussed way of handling this issue.

harshavardhana:
Closing this, as this is superseded by #2007 - please re-open otherwise.

PR #1903, opened 2016-06-15
fs-v1: don't guess Content-Type by extension

### LINKED ISSUE #1902: Don't set default Content-Type
## Expected behaviour

GetObjectHandler should respond with a best-guess Content-Type not depending on the extension.
## Actual behaviour

Files without extension are reported as `application/octet-stream`.

Upload a .png, create a presigned get url and check the Content-Type with curl:

```
$ file test.png 
test.png: PNG image data, 264 x 384, 8-bit/color RGBA, interlaced
$ mc cp test.png ${host}/test/png_without_extension
$ mc share download ${host}/test/png_without_extension
.... copy URL from snippet ....
$ curl -v ${DOWNLOAD_URL} > /dev/null
....
 < Content-Type: application/octet-stream
....
```
## Minio version

git afc310248895
## System information

ubuntu 14.04

I'm aware that the XL Backend can store a predefined Content-Type and also responds it, which is a good thing. What I don't like is that both the fs and the XL backend default to `application/octet-stream`. Just leaving the Content-Type out in `setObjectHeaders` magically fixes this and fills in the correct content type.

Note that this only affects the `api` but not the `web` interface, as it seems that the `web` interface ignores the Content-Type from `GetObjectInfo` and thus triggers the mentioned magic.

diresi:
The go server does this automatically, and better.

Fixes #1902.

### DIFF
--- a/api-headers.go
+++ b/api-headers.go
@@ -66,7 +66,9 @@ func setObjectHeaders(w http.ResponseWriter, objInfo ObjectInfo, contentRange *h
 	lastModified := objInfo.ModTime.UTC().Format(http.TimeFormat)
 	w.Header().Set("Last-Modified", lastModified)
 
-	w.Header().Set("Content-Type", objInfo.ContentType)
+	if objInfo.ContentType != "" {
+		w.Header().Set("Content-Type", objInfo.ContentType)
+	}
 	if objInfo.MD5Sum != "" {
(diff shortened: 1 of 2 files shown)


diresi:
Well, that doesn't look too good ...

harshavardhana:
Thanks @diresi for reporting the issue should be - Fixed by #1918

Issue #1657, opened 2016-05-17
Unit tests for PUT object when the object already exists


krishnasrinivas:
unit test case for the commit https://github.com/minio/minio/commit/8099396ff0db764baa7102e86868c21f510657ca
, unit test case should do this:

```
PUT small object 1/2/3/4 should succeed
PUT small object 1/2/3 should fail
PUT small object 1/2/3/4 should fail
PUT small object 1/2/3/4/5 should fail

PUT large object 1/2/3 should fail
PUT large object 1/2/3/4 should fail
PUT large object 1/2/3/4/5 should fail

DELETE small object 1/2/3/4 should succeed


PUT large object 1/2/3/4 should succeed
PUT small object 1/2/3 should fail
PUT small object 1/2/3/4 should fail
PUT small object 1/2/3/4/5 should fail

PUT large object 1/2/3 should fail
PUT large object 1/2/3/4 should fail
PUT large object 1/2/3/4/5 should fail

DELETE large object 1/2/3/4 should succeed
```

harshavardhana:
Assigned to @hackintoshrao

GarimaKapoor:
please update timeline for closing the issue.

mkash32:
@GarimaKapoor I am working on it, will be able to send pr by tomorrow.

mkash32:
@GarimaKapoor sorry, this took longer than I expected

PR #1904, opened 2016-06-15
Unit tests for PUT 

mkash32:
Unit tests for PUT object when object already exists
Issue: https://github.com/minio/minio/issues/1657

### DIFF
--- a/server_xl_test.go
+++ b/server_xl_test.go
@@ -226,6 +226,111 @@ func (s *MyAPIXLSuite) newRequest(method, urlStr string, contentLength int64, bo
 	return req, nil
 }
 
+// putSimpleObjectMultipart uploads a multipart object consisting of 2 parts, repeated constant byte string and a single byte ('0').
+func (s *MyAPIXLSuite) putSimpleObjectMultipart(bucketName, object string) (response *http.Response, err error) {
+	request, err := s.newRequest("POST", testAPIXLServer.URL+"/"+bucketName+"/"+object+"?uploads", 0, nil)
+	if err != nil {
+		return nil, err
+	}
+	client := http.Client{}
+	response, err = client.Do(request)
+	if err != nil {
+		return nil, err
+	}
+	if response.StatusCode != http.StatusOK {
+		return response, nil
+	}
+
+	decoder := xml.NewDecoder(response.Body)
+	newResponse := &InitiateMultipartUploadResponse{}
+
+	err = decoder.Decode(newResponse)
+	if err != nil {
+		return nil, err
+	}
+
+	uploadID := newResponse.UploadID
+
+	// Create a byte array of 5MB.
+	data := bytes.Repeat([]byte("0123456789abcdef"), 5*1024*1024/16)
+
+	hasher := md5.New()
+	hasher.Write(data)
+	md5Sum := hasher.Sum(nil)
+
+	buffer1 := bytes.NewReader(data)
+	request, err = s.newRequest("PUT", testAPIXLServer.URL+"/"+bucketName+"/"+object+"?uploadId="+uploadID+"&partNumber=1", int64(buffer1.Len()), buffer1)
+	request.Header.Set("Content-Md5", base64.StdEncoding.EncodeToString(md5Sum))
+	if err != nil {
+		return nil, err
+	}
+
+	client = http.Client{}
+	response1, err := client.Do(request)
+	if err != nil {
+		return nil, err
+	}
+	if response1.StatusCode != http.StatusOK {
+		return response1, nil
+	}
+
+	// Byte array one 1 byte.
+	data = []byte("0")
+
+	hasher = md5.New()
+	hasher.Write(data)
+	md5Sum = hasher.Sum(nil)
+
+	buffer2 := bytes.NewReader(data)
+	request, err = s.newRequest("PUT", testAPIXLServer.URL+"/"+bucketName+"/"+object+"?uploadId="+uploadID+"&partNumber=2", int64(buffer2.Len()), buffer2)
+	request.Header.Set("Content-Md5", base64.StdEncoding.EncodeToString(md5Sum))
+	if err != nil {
+		return nil, err
+	}
+
+	client = http.Client{}
+	response2, err := client.Do(request)
+	if err != nil {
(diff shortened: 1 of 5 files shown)


server_xl_test.go
```diff
@@ -1482,3 +1587,114 @@ func (s *MyAPIXLSuite) TestObjectMultipartOverwriteSinglePut(c *C) {
...
+
+	// Put object a/b/c, should fail because it is a directory
+	buffer2 := bytes.NewReader([]byte("hello two"))
+	request, err = s.newRequest("PUT", testAPIXLServer.URL+"/multipleobjects/a/b/c", int64(buffer2.Len()), buffer2)
+	c.Assert(err, IsNil)
+
+	client = http.Client{}
+	response, err = client.Do(request)
+	c.Assert(err, IsNil)
+	verifyError(c, response, "InternalError", "We encountered an internal error, please try again.", http.StatusInternalServerError)
```
mkash32: @krishnasrinivas This request should result in an error but it is successful. The same test works on fs, but on xl we are getting unexpected results: success for small/large put object a/b/c when an object a/b/c/d is present.

xl-v1-multipart.go
```diff
@@ -631,9 +631,11 @@ func (xl xlObjects) CompleteMultipartUpload(bucket string, object string, upload
 
 	// Rename if an object already exists to temporary location.
 	uniqueID := getUUID()
-	err = xl.renameObject(bucket, object, minioMetaBucket, path.Join(tmpMetaPrefix, uniqueID))
-	if err != nil {
-		return "", toObjectErr(err, bucket, object)
+	if xl.isObject(bucket, object) {
+		err = xl.renameObject(bucket, object, minioMetaBucket, path.Join(tmpMetaPrefix, uniqueID))
```
harshavardhana, line 653: renameObject() already handles this inside if the `object` is not found the error is ignored, we don't need to do 'isObject()' again.

xl-v1-object.go
```diff
@@ -343,9 +343,11 @@ func (xl xlObjects) PutObject(bucket string, object string, size int64, data io.
 
 	// Rename if an object already exists to temporary location.
 	newUniqueID := getUUID()
-	err = xl.renameObject(bucket, object, minioMetaBucket, path.Join(tmpMetaPrefix, newUniqueID))
-	if err != nil {
-		return "", toObjectErr(err, bucket, object)
+	if xl.isObject(bucket, object) {
+		err = xl.renameObject(bucket, object, minioMetaBucket, path.Join(tmpMetaPrefix, newUniqueID))
```
harshavardhana, line 339: same as above.

xl-v1-multipart.go
krishnasrinivas, line 653: The problem was renameObject() was renaming the parent directory.

for ex: if object is `a/b/c/d` rename() on `a/b/` would succeed.

probably xl.isObject() can be moved inside xl.renameObject()

harshavardhana, line 653: > The problem was renameObject() was renaming the parent directory.
> 
> for ex: if object is a/b/c/d rename() on a/b/ would succeed.
> 
> probably xl.isObject() can be moved inside xl.renameObject()

why would there be a rename on `a/b/` ?  when the object is `a/b/c/d`,  seems like the bug is somewhere else.

krishnasrinivas, line 653: > why would there be a rename on a/b/ ? when the object is a/b/c/d, seems like the bug is somewhere else.

check the unit test case.
1) create object a/b/c/d
2) create object a/b - now this is expected fail - because ultimately when we do renam(tmp, "a/b') the rename would fail complaining that distination directory "a/b" already exists.

But create object "a/b" is succeeding because we would have renamed original "a/b" to a tmp path. (we should rename only if it was an object, not prefix of an object)

harshavardhana, line 653: > check the unit test case.
> 1) create object a/b/c/d
> 2) create object a/b - now this is expected fail - because ultimately when we do renam(tmp, "a/b') the rename would fail complaining that distination directory "a/b" already exists.
> 
> But create object "a/b" is succeeding because we would have renamed original "a/b" to a tmp path. (we should rename only if it was an object, not prefix of an object)

Then i think we need to move this check inside `renameObject()` .

mkash32, line 653: > Then i think we need to move this check inside renameObject() .

Is the check needed inside `renameObject()`? Another call is made to rename the temporary object to final location, we already know the source is an object in that case so is the check necessary?

krishnasrinivas, line 653: > Is the check needed inside renameObject()? Another call is made to rename the temporary object to final location, we already know the source is an object in that case so is the check necessary?

extra call inside xl.renameObject() is fine even if we know that the source is an object. It will avoid the bugs like we see now.

mkash32, line 653:
```
@@ -171,6 +171,9 @@ func (xl xlObjects) getObjectInfo(bucket, object string) (ob
 // its proper location.
 func (xl xlObjects) renameObject(srcBucket, srcObject, dstBucket, dstObject str
+       // Do not rename if source is not an object.
+       if !xl.isObject(srcBucket, srcObject) {
+               return nil
+       }
        // Initialize sync waitgroup.
        var wg = &sync.WaitGroup{}

        // Initialize list of errors.
```

I made the above change and it is causing an issue with xl multipart uploads. After any multipart upload, a GET request is resulting in 404 status code.

mkash32, line 653: `renameObject()` call in putObjectPart in xl-v1-multipart.go (L370) is the reason for the problem. I think `tmpPartPath` is the path to the part (which will give false in the xl.isObject check) and therefore renameObject will return without renaming.

harshavardhana:
Can you rebase this patch - not sure why windows tests are failing.

mkash32:
@harshavardhana Is it failing because of something in the test (like a/b/c/d since we are using literal '/') or is it because of the fix?

harshavardhana:
> @harshavardhana Is it failing because of something in the test (like a/b/c/d since we are using literal '/') or is it because of the fix?

Do not know, will need to validate this properly on windows, will be doing that shortly! - will report back on my findings.

PR #1909, opened 2016-06-16
XL: Remove usage of reduceErr and make it isQuorum verification.

### LINKED ISSUE #1908: reduceError() incorrect reduction
reduceErrors() returns nil when it should return an error.
For example: on 4x4 if fileNotFoundCount is 3 and diskNotFoundCount is 2, reduceError() should return `error` but it returns `nil`

harshavardhana:
Fixes #1908

### DIFF
--- a/posix.go
+++ b/posix.go
@@ -458,9 +458,6 @@ func (s posix) AppendFile(volume, path string, buf []byte) (n int64, err error) 
 		}
 		return 0, err
 	}
-	if err = checkDiskFree(s.diskPath, s.minFreeDisk); err != nil {
-		return 0, err
-	}
 	filePath := pathJoin(volumeDir, path)
 	if err = checkPathLength(filePath); err != nil {
 		return 0, err
@@ -670,9 +667,17 @@ func (s posix) RenameFile(srcVolume, srcPath, dstVolume, dstPath string) (err er
 	if !(srcIsDir && dstIsDir || !srcIsDir && !dstIsDir) {
 		return errFileAccessDenied
 	}
+	srcFilePath := slashpath.Join(srcVolumeDir, srcPath)
+	if err = checkPathLength(srcFilePath); err != nil {
+		return err
+	}
+	dstFilePath := slashpath.Join(dstVolumeDir, dstPath)
+	if err = checkPathLength(dstFilePath); err != nil {
+		return err
+	}
 	if srcIsDir {
 		// If source is a directory we expect the destination to be non-existent always.
-		_, err = os.Stat(preparePath(slashpath.Join(dstVolumeDir, dstPath)))
+		_, err = os.Stat(preparePath(dstFilePath))
 		if err == nil {
 			return errFileAccessDenied
 		}
@@ -681,14 +686,14 @@ func (s posix) RenameFile(srcVolume, srcPath, dstVolume, dstPath string) (err er
 		}
 		// Destination does not exist, hence proceed with the rename.
 	}
-	if err = mkdirAll(preparePath(slashpath.Dir(slashpath.Join(dstVolumeDir, dstPath))), 0755); err != nil {
+	if err = mkdirAll(preparePath(slashpath.Dir(dstFilePath)), 0755); err != nil {
 		// File path cannot be verified since one of the parents is a file.
 		if strings.Contains(err.Error(), "not a directory") {
 			return errFileAccessDenied
 		}
 		return err
 	}
-	err = os.Rename(preparePath(slashpath.Join(srcVolumeDir, srcPath)), preparePath(slashpath.Join(dstVolumeDir, dstPath)))
+	err = os.Rename(preparePath(srcFilePath), preparePath(dstFilePath))
 	if err != nil {
 		if os.IsNotExist(err) {
 			return errFileNotFound
--- a/xl-v1-bucket.go
+++ b/xl-v1-bucket.go
@@ -33,10 +33,6 @@ func (xl xlObjects) MakeBucket(bucket string) error {
 	nsMutex.Lock(bucket, "")
 	defer nsMutex.Unlock(bucket, "")
 
-	// Err counters.
-	createVolErr := 0       // Count generic create vol errs.
-	volumeExistsErrCnt := 0 // Count all errVolumeExists errs.
-
 	// Initialize sync waitgroup.
 	var wg = &sync.WaitGroup{}
 
@@ -63,29 +59,47 @@ func (xl xlObjects) MakeBucket(bucket string) error {
 	// Wait for all make vol to finish.
 	wg.Wait()
 
-	// Look for specific errors and count them to be verified later.
-	for _, err := range dErrs {
-		if err == nil {
-			continue
-		}
-		// if volume already exists, count them.
-		if err == errVolumeExists {
-			volumeExistsErrCnt++
-			continue
-		}
-
-		// Update error counter separately.
-		createVolErr++
-	}
-
-	// Return err if all disks report volume exists.
-	if volumeExistsErrCnt > len(xl.storageDisks)-xl.readQuorum {
-		return toObjectErr(errVolumeExists, bucket)
-	} else if createVolErr > len(xl.storageDisks)-xl.writeQuorum {
-		// Return errXLWriteQuorum if errors were more than allowed write quorum.
+	// Do we have write quorum?.
+	if !isQuorum(dErrs, xl.writeQuorum) {
+		// Purge successfully created buckets if we don't have writeQuorum.
+		xl.undoMakeBucket(bucket)
 		return toObjectErr(errXLWriteQuorum, bucket)
 	}
+
+	// Verify we have any other errors which should undo make bucket.
+	for _, err := range dErrs {
+		// Bucket already exists, return BucketExists error.
+		if err == errVolumeExists {
+			return toObjectErr(errVolumeExists, bucket)
+		}
+		// Undo make bucket for any other errors.
+		if err != nil && err != errDiskNotFound {
+			xl.undoMakeBucket(bucket)
+			return toObjectErr(err, bucket)
+		}
+	}
 	return nil
+}
+
+// undo make bucket operation upon quorum failure.
+func (xl xlObjects) undoMakeBucket(bucket string) {
+	// Initialize sync waitgroup.
+	var wg = &sync.WaitGroup{}
+	// Undo previous make bucket entry on all underlying storage disks.
+	for index, disk := range xl.storageDisks {
+		if disk == nil {
+			continue
+		}
+		wg.Add(1)
+		// Delete a bucket inside a go-routine.
+		go func(index int, disk StorageAPI) {
+			defer wg.Done()
+			_ = disk.DeleteVol(bucket)
+		}(index, disk)
+	}
+
+	// Wait for all make vol to finish.
+	wg.Wait()
 }
 
 // getBucketInfo - returns the BucketInfo from one of the load balanced disks.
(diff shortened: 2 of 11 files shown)


xl-v1-utils.go
```diff
@@ -24,6 +24,72 @@ import (
...
+	if fileNotFoundCount+diskNotFoundCount > len(errs)-quorumCount {
+		return errFileNotFound
+	} else if fileLongNameCount+diskNotFoundCount > len(errs)-quorumCount {
+		return errFileNameTooLong
+	} else if volumeNotFoundCount+diskNotFoundCount > len(errs)-quorumCount {
+		return errVolumeNotFound
+	} else if fileAccessDeniedCount+diskNotFoundCount > len(errs)-quorumCount {
+		return errFileAccessDenied
+	}
+	return nil
```
krishnasrinivas: still does not cover all cases, for example in 4x4 if fileNotFoundCount is 3 and volumeNotFoundCount is 2, reduceError() should return error but it returns nil

posix.go
```diff
@@ -458,11 +458,9 @@ func (s posix) AppendFile(volume, path string, buf []byte) (n int64, err error)
 		}
 		return 0, err
 	}
-	if err = checkDiskFree(s.diskPath, s.minFreeDisk); err != nil {
-		return 0, err
-	}
```
krishnasrinivas, line 463: checkDiskFree() check is not necessary?

posix.go
```diff
@@ -458,11 +458,9 @@ func (s posix) AppendFile(volume, path string, buf []byte) (n int64, err error)
...
 	filePath := pathJoin(volumeDir, path)
-	if err = checkPathLength(filePath); err != nil {
+	err = checkPathLength(filePath)
+	if err != nil {
 		return 0, err
```
krishnasrinivas, line 463: how come you broke `if err = checkPathLength(filePath); err != nil {}` style to this style? does it give any advantage?

xl-v1-healing.go
```diff
@@ -7,8 +23,7 @@ import (
 )
 
 // Get the highest integer from a given integer slice.
-func highestInt(intSlice []int64) (highestInteger int64) {
-	highestInteger = int64(1)
+func highestInt(intSlice []int64, highestInt int64) (highestInteger int64) {
```
krishnasrinivas, line 26: `highestInt int64` is not getting used in the function?

harshavardhana:
> how come you broke if err = checkPathLength(filePath); err != nil {} style to this style? does it give any advantage?

Ah must be related to some debugging which i was doing, will need to revert it back.

harshavardhana:
> checkDiskFree() check is not necessary?

Nope looks like it was already there, it was redundant.

harshavardhana:
@krishnasrinivas were you able to review?

krishnasrinivas:
> @krishnasrinivas were you able to review?

will review now

xl-v1-utils.go
```diff
@@ -24,6 +24,33 @@ import (
...
+		if err == errDiskNotFound {
+			diskNotFoundCount++
+		}
+	}
+	if diskNotFoundCount > len(errs)-minQuorumCount {
+		return false
+	}
+	return true
+}
+
```
krishnasrinivas, line 38: logic based on diskFoundCount instead of diskNotFoundCount is more readable, like:

```
func isQuorum(errs []error, minQuorumCount int) bool {
        var diskFoundCount int
        for _, err := range errs {
                if err != errDiskNotFound {
                        diskFoundCount++
                }
        }
        return diskFoundCount >= minQuorumCount
}
```

xl-v1-bucket.go
```diff
@@ -63,27 +79,16 @@ func (xl xlObjects) MakeBucket(bucket string) error {
...
-	if volumeExistsErrCnt > len(xl.storageDisks)-xl.readQuorum {
-		return toObjectErr(errVolumeExists, bucket)
-	} else if createVolErr > len(xl.storageDisks)-xl.writeQuorum {
-		// Return errXLWriteQuorum if errors were more than allowed write quorum.
-		return toObjectErr(errXLWriteQuorum, bucket)
+	for _, err := range dErrs {
+		if err != nil && err != errDiskNotFound {
+			xl.undoMakeBucket(bucket)
+			return toObjectErr(err, bucket)
+		}
```
krishnasrinivas, line 79: this has undesired consequence:
1) `mc mb myminio/newbucket` - creates newbucket
2) again do `mc mb myminio/newbucket` - deletes the bucket created in step-1

xl-v1-metadata.go
```diff
@@ -259,39 +282,18 @@ func (xl xlObjects) renameXLMetadata(srcBucket, srcPrefix, dstBucket, dstPrefix
...
-	var errCount = 0
-	for _, err := range mErrs {
-		if err == nil {
-			continue
+	if !isQuorum(mErrs, xl.writeQuorum) {
+		if isQuorum(mErrs, xl.readQuorum) {
+			return nil
 		}
-		errCount++
+		xl.undoRenameXLMetadata(srcBucket, srcPrefix, dstBucket, dstPrefix, mErrs)
```
krishnasrinivas: `renameXLMetadata()` source is always a tmp location so undoRenameXLMetadata to rename it back to a tmp file will not be useful

xl-v1-metadata.go
```diff
@@ -259,39 +282,18 @@ func (xl xlObjects) renameXLMetadata(srcBucket, srcPrefix, dstBucket, dstPrefix
...
-			go func(index int, disk StorageAPI) {
-				defer wg.Done()
-				if mErrs[index] != nil {
-					return
-				}
-				_ = disk.RenameFile(dstBucket, dstJSONFile, srcBucket, srcJSONFile)
-			}(index, disk)
+	for _, err := range mErrs {
+		if err != nil && err != errDiskNotFound {
+			xl.undoRenameXLMetadata(srcBucket, srcPrefix, dstBucket, dstPrefix, mErrs)
```
krishnasrinivas: doing undoRenameXLMetadata during complete-multipart-upload will be harmful because it will remove xl.json from the .minio/multipart/object/ directory

xl-v1-metadata.go
```diff
@@ -259,39 +282,18 @@ func (xl xlObjects) renameXLMetadata(srcBucket, srcPrefix, dstBucket, dstPrefix
...
+			return err
```
krishnasrinivas: returning `err` without doing `undoRenameXLMetadata()` would be good

xl-v1-metadata.go
```diff
@@ -352,38 +375,23 @@ func (xl xlObjects) writeUniqueXLMetadata(bucket, prefix string, xlMetas []xlMet
...
-			}(index, disk)
-		}
-		wg.Wait()
+		xl.deleteAllXLMetadata(bucket, prefix, mErrs)
 		return errXLWriteQuorum
 	}
+
+	for _, err := range mErrs {
+		if err != nil && err != errDiskNotFound {
+			xl.deleteAllXLMetadata(bucket, prefix, mErrs)
```
krishnasrinivas: since `writeUniqueXLMetadata` writes to the tmp location doing `xl.deleteAllXLMetadata` is not needed

xl-v1-metadata.go
```diff
@@ -417,37 +425,21 @@ func (xl xlObjects) writeSameXLMetadata(bucket, prefix string, xlMeta xlMetaV1)
...
```
krishnasrinivas: since `writeUniqueXLMetadata` writes to the tmp location doing `xl.deleteAllXLMetadata` is not needed

BTW is writeSameXLMetadata needed? we can use writeUniqueXLMetadata right?

harshavardhana:
> 2) again do mc mb myminio/newbucket - deletes the bucket created in step-1

Ah yes need to ignore 'errVolumeExists'

xl-v1-metadata.go
harshavardhana: > since writeUniqueXLMetadata writes to the tmp location doing xl.deleteAllXLMetadata is not needed

The function itself has no such awareness of tmp location so that is why it is done this way.

harshavardhana:
> BTW is writeSameXLMetadata needed? we can use writeUniqueXLMetadata right?

Both have different purpose, in one case xlMetaV1 is different for each index whereas writeSameXL writes same xlMetaV1 across all disks.

xl-v1-utils.go
harshavardhana, line 38: > logic based on diskFoundCount instead of diskNotFoundCount is more readable, like:
> 
> func isQuorum(errs []error, minQuorumCount int) bool {
>         var diskFoundCount int
>         for _, err := range errs {
>                 if err != errDiskNotFound {
>                         diskFoundCount++
>                 }
>         }
>         return diskFoundCount >= minQuorumCount
> }

Makes sense.

xl-v1-metadata.go
harshavardhana: > renameXLMetadata() source is always a tmp location so undoRenameXLMetadata to rename it back to a tmp file will not be useful

It would be since 'tmp/' location can be cleaned naturally without issues, any dangling files would be an issue to detect and cleanup later.

harshavardhana:
> doing undoRenameXLMetadata during complete-multipart-upload will be harmful because it will remove xl.json from the .minio/multipart/object/ directory

This is only on error when it is necessary, it will only remove the dangling ones since other operations would have naturally failed.

harshavardhana:
> returning err without doing undoRenameXLMetadata() would be good

That would still leave dangling files, since all this operation is based on tmp path. we can't have dangling files on actual namespace. it is okay to rename them back to tmp so that upon server restart atleast we can clean them up.

harshavardhana:
@krishnasrinivas - updated with new changes.

krishnasrinivas:
> @krishnasrinivas - updated with new changes.

checking...

krishnasrinivas:
> > doing undoRenameXLMetadata during complete-multipart-upload will be harmful because it will remove xl.json from the .minio/multipart/object/ directory
> 
> This is only on error when it is necessary, it will only remove the dangling ones since other operations would have naturally failed.

No, here's the situation,
CompleteMultipart would call:
rErr := xl.renameXLMetadata(minioMetaBucket, tempUploadIDPath, minioMetaBucket, uploadIDPath)

here we would already have `uploadIDPath/xl.json` which will be replaced by `tempUploadIDPath/xl.json` but for some reason we do `undoRenameXLMetadata()` it would mean that `uploadIDPath` will not have xl.json at all (the original xl.json would have been replaced)

harshavardhana:
> No, here's the situation,
> CompleteMultipart would call:
> rErr := xl.renameXLMetadata(minioMetaBucket, tempUploadIDPath, minioMetaBucket, uploadIDPath)
> 
> here we would already have uploadIDPath/xl.json which will be replaced by tempUploadIDPath/xl.json but for some reason we do undoRenameXLMetadata() it would mean that uploadIDPath will not have xl.json at all (the original xl.json would have been replaced)

It is a tricky case i agree but leaving a dangling / improper `xl.json` during a quorum failure would be hard to detect, in updated patch renameXLMetadata undo'es the operation only during quorum failure. Any other errors are returned as is.

krishnasrinivas:
> It is a tricky case i agree but leaving a dangling / improper xl.json

it's not dangling right? it would have just replaced previous xl.json

harshavardhana:
> it's not dangling right? it would have just replaced previous xl.json

Now i see, since this renameXL is not used elsewhere this logic doesn't make sense. `xl.json` wouldn't be a dangling one.

harshavardhana:
> Now i see, since this renameXL is not used elsewhere this logic doesn't make sense. xl.json wouldn't be a dangling one.

Also it is not appropriate name for the function, moving it to xl-multipart-common.go

harshavardhana:
Done updated!

krishnasrinivas:
looks good

PR #1910, opened 2016-06-16
erasure-readfile: write to given Writer than returning buffer.

### LINKED ISSUE #1889: XL: Read of a large part will cause out of memory error
```
98                  var buffer []byte                                                                                                                                             
99                  buffer, err = erasureReadFile(onlineDisks, bucket, pathJoin(object, partName), partName, partSize, eInfos)                                                    
100                 if err != nil {                                                                                                                                               
101                         return err                                                                                                                                            
102                 }
```

If it was a PUT operation of a large file (for example a 4G file on a VM with 3G RAM) this operation will fail with oom error.
@harshavardhana  Was it a conscious decision to do it this way?

balamurugana:
Fixes #1889

### DIFF
--- a/erasure-readfile.go
+++ b/erasure-readfile.go
@@ -17,26 +17,22 @@
 package main
 
 import (
+	"bytes"
 	"encoding/hex"
 	"errors"
+	"io"
 
 	"github.com/klauspost/reedsolomon"
 )
 
-// erasureReadFile - read an entire erasure coded file at into a byte
-// array. Erasure coded parts are often few mega bytes in size and it
-// is convenient to return them as byte slice. This function also
-// supports bit-rot detection by verifying checksum of individual
-// block's checksum.
-func erasureReadFile(disks []StorageAPI, volume string, path string, partName string, size int64, eInfos []erasureInfo) ([]byte, error) {
-	// Return data buffer.
-	var buffer []byte
-
-	// Total size left
-	totalSizeLeft := size
-
-	// Starting offset for reading.
-	startOffset := int64(0)
+// erasureReadFile - read bytes from erasure coded files and writes to given writer.
+// Erasure coded files are read block by block as per given erasureInfo and data chunks
+// are decoded into a data block.  Data block is trimmed for given offset and length,
+// then written to given writer.  This function also supports bit-rot detection by
+// verifying checksum of individual block's checksum.
+func erasureReadFile(writer io.Writer, disks []StorageAPI, volume string, path string, partName string, eInfos []erasureInfo, offset int64, length int64) (int64, error) {
+	// Total bytes written to writer
+	bytesWritten := int64(0)
 
 	// Gather previously calculated block checksums.
 	blockCheckSums := metaPartBlockChecksums(disks, eInfos, partName)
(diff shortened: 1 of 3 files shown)


erasure-readfile.go
```diff
@@ -97,25 +99,34 @@ func erasureReadFile(disks []StorageAPI, volume string, path string, partName st
...
+			return bytesWritten, err
 		}
 
 		// Copy data blocks.
-		buffer = append(buffer, dataBlocks...)
+		var n int64
+		if offset > 0 {
+			n, err = io.Copy(writer, bytes.NewBuffer(dataBlocks[offset:]))
+		} else {
+			n, err = io.Copy(writer, bytes.NewBuffer(dataBlocks))
```
harshavardhana: Use bytes.NewReader()

erasure-readfile.go
```diff
@@ -97,25 +99,34 @@ func erasureReadFile(disks []StorageAPI, volume string, path string, partName st
...
+		}
+		bytesWritten += int64(n)
+		if err != nil {
+			return bytesWritten, err
```
harshavardhana, line 136: Another problem is you are not taking in the 'length' requirement you are writing more than requested.

xl-v1-object.go
```diff
@@ -95,26 +94,16 @@ func (xl xlObjects) GetObject(bucket, object string, startOffset int64, length i
...
-			var m int64
-			m, err = io.Copy(writer, bytes.NewReader(buffer[partOffset:]))
-			if err != nil {
-				return err
-			}
-			length -= m
+		if length > n {
+			length -= n
 		} else {
-			_, err = io.CopyN(writer, bytes.NewReader(buffer[partOffset:]), length)
```
harshavardhana, line 108: This case in not handled inside erasureReadFile()

erasure-readfile.go
balamurugana: Done

erasure-readfile.go
balamurugana, line 136: Done. Fixing it.

xl-v1-object.go
balamurugana, line 108: Done. Fixing it.

erasure-readfile.go
```diff
@@ -17,54 +17,47 @@
...
 	"errors"
+	"io"
 
 	"github.com/klauspost/reedsolomon"
 )
 
-// erasureReadFile - read an entire erasure coded file at into a byte
-// array. Erasure coded parts are often few mega bytes in size and it
-// is convenient to return them as byte slice. This function also
-// supports bit-rot detection by verifying checksum of individual
```
harshavardhana, line 29: Why this comment is removed? - does this function remove bit-rot detection?

erasure-readfile.go
```diff
@@ -74,13 +67,33 @@ func erasureReadFile(disks []StorageAPI, volume string, path string, partName st
...
+
+			// Remember bytes read at first time.
+			if lastReadSize == 0 {
+				lastReadSize = n
+			}
+
+			// TODO: In a suitation, there may be very few corrupted shards among total shards
+			// ie in 16 disks setup, there may be size mismatching corrupted shards
+			// (assume first three shards are invalid).  This is still valid to retrieve
+			// the data by reconstruct without those shards.
```
harshavardhana: Do not call shards, call 'blocks' as consistently used everywhere.

erasure-readfile.go
```diff
@@ -17,54 +17,47 @@
...
-	for totalSizeLeft > 0 {
-		// Calculate the proper block size.
-		var curBlockSize int64
-		if eInfo.BlockSize < totalSizeLeft {
-			curBlockSize = eInfo.BlockSize
-		} else {
-			curBlockSize = totalSizeLeft
-		}
+	// Calculate start block for given offset and how many bytes to skip to get the offset.
+	startBlock := offset / eInfo.BlockSize
```
harshavardhana: Move these into a function so that it's useful when we bring in healing code.

erasure-readfile.go
```diff
@@ -17,54 +17,47 @@
...
+	for block := startBlock; block <= endBlock; block++ {
 		// Allocate encoded blocks up to storage disks.
 		enBlocks := make([][]byte, len(disks))
 
 		// Counter to keep success data blocks.
 		var successDataBlocksCount = 0
 		var noReconstruct bool // Set for no reconstruction.
 
+		// Keep how many bytes are read for this block.
+		// In most cases, last block in the file is short than eInfo.BlockSize
```
harshavardhana: s/'last block in the file is short than'/'last block in the file is shorter than'

erasure-readfile.go
```diff
@@ -74,13 +67,33 @@ func erasureReadFile(disks []StorageAPI, volume string, path string, partName st
...
+				// As the data we got is smaller than eInfo.BlockSize, keep only required shard slice
+				enBlocks[blockIndex] = append([]byte{}, enBlocks[blockIndex][:n]...)
+			}
+
+			// Remember bytes read at first time.
+			if lastReadSize == 0 {
+				lastReadSize = n
+			}
+
+			// TODO: In a suitation, there may be very few corrupted shards among total shards
```
harshavardhana: can you open a bug on this and explain the situation there? - need to see how this plays out, since a corrupted block would have been skipped by now after the `isValidBlock`. Only situation is a small window where other disks also get corrupted after the sha512 validation.

erasure-readfile.go
balamurugana, line 29: > does this function remove bit-rot detection?

No change is made wrt block verification.  Adding the comment again.

erasure-readfile.go
balamurugana: Are you talking about moving start/end block calculation into a separate function?

erasure-readfile.go
balamurugana: Done

erasure-readfile.go
balamurugana: Done.

erasure-readfile.go
balamurugana: Done.

harshavardhana:
> Are you talking about moving start/end block calculation into a separate function?

Yeah

PR #1911, opened 2016-06-16
erasure: Add erasure encode/decode unit tests.

### LINKED ISSUE #1819: XL : Tests:  Unit tests for erasure coding logic.
This is an extremely important test. We could even consider fuzzing here. Tests for `erasure-*` files.

harshavardhana:
Fixes #1819


erasure_test.go
```diff
@@ -0,0 +1,133 @@
...
+			encodedData:  genInsufficientData(data, 4, 4),
+			dataBlocks:   4,
+			parityBlocks: 4,
+			blockSize:    len(data),
+			shouldPass:   false,
+		},
+	}
+
+	for i, testCase := range testDecodeCases {
+		err := decodeData(testCase.encodedData, testCase.dataBlocks, testCase.parityBlocks)
```
abperiasamy: Comments required here. We should also generalize these tests for 4x4, 6x6 and 8x8 configs.

erasure_test.go
```diff
@@ -0,0 +1,231 @@
...
+	encodedData, err := encodeData(data, dataBlocks, parityBlocks)
+	if err != nil {
+		// Upon failure panic this function.
+		panic(err)
+	}
+	return encodedData
+}
+
+// Generates good encoded data with one parity block and data block missing.
+var getGoodEncodedData = func(data []byte, dataBlocks, parityBlocks int) [][]byte {
```
abperiasamy: func getGoodEncodedData(..)  is simpler. Is there a good reason to use anonymous func here?

erasure_test.go
```diff
@@ -0,0 +1,231 @@
...
+	return encodedData[:dataBlocks/2]
+}
+
+// TestErasureEncode checks for encoding for different data sets.
+func TestErasureEncode(t *testing.T) {
+	data := []byte("Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book. It has survived not only five centuries, but also the leap into electronic typesetting, remaining essentially unchanged. It was popularised in the 1960s with the release of Letraset sheets containing Lorem Ipsum passages, and more recently with desktop publishing software like Aldus PageMaker including versions of Lorem Ipsum.")
+
+	// List of data blocks the test will run on.
+	var listDataBlocks = []int{
+		4,
```
abperiasamy: Why was 1 data block and 4 parity chosen, when the comment says test will run on data blocks?

harshavardhana:
> Why was 1 data block and 4 parity chosen, when the comment says test will run on data blocks?

Data is entire data set,  data block is still '4'

erasure_test.go
```diff
@@ -0,0 +1,186 @@
...
+	// - validates insufficient data.
+	testDecodeCases := []struct {
+		enFn       func([]byte, int, int) [][]byte
+		blockSize  int
+		shouldPass bool
+	}{
+		// Generates bad encoded data.
+		{
+			enFn:       getBadEncodedData,
+			blockSize:  len(data),
```
abperiasamy: even blockSize can be directly referred in the loop.

PR #1931, opened 2016-06-19
XL: CompleteMultipart should ignore last part is 0bytes.

### LINKED ISSUE #1917: XL: Skip zero sized last part in complete-multipart call
Currently we store zero sized last part in complete-multipart and check whether last part size is zero or not in GetObject().

We could avoid the check by skipping zero sized last part in CompleteMultipart call.

harshavardhana:
Fixes #1917


balamurugana:
:+1: LGTM

Issue #1893, opened 2016-06-11
storageAPI:  New changes to ReadFile and AppendFile


krishnasrinivas:
(Discussed during the meeting)
The changes to avoid looping at the object layer.

current storage api:

```
        ReadFile(volume string, path string, offset int64, buf []byte) (n int64, err error)                                                                                             
        AppendFile(volume string, path string, buf []byte) (n int64, err error)                                                                                                         
```

new api:

```
        // ReadFile assures that it returns 'n' bytes, if 'n' bytes are not available
        // it will set io.EOF as error. (similar to io.ReadFull behavior)
        ReadFile(volume string, path string, offset int64, n int64) (buf []byte, err error)                                                                                                      
        // AppendFile assures that it appends the entire buffer else returns error.
        AppendFile(volume string, path string, buf []byte) (err error)
```

abperiasamy:
+1

abperiasamy:
call n as length

harshavardhana:
Will take over this work, if it's not finished yet.  Let me know the current status.

krishnasrinivas:
@harshavardhana sure you can take over, havent started this work. But see if this is needed for this release as it won't be affecting backend/performance.

harshavardhana:
Closing the issue as AppendFile change is finished and ReadFile change has been deferred.

PR #1932, opened 2016-06-19
XL: Change AppendFile() to return only error

harshavardhana:
AppendFile ensures that it appends the entire buffer. Returns
an error otherwise, this patch removes the necessity for the
caller to look for 'n' return on short writes.

Ref #1893


harshavardhana:
// cc @krishnasrinivas can you review?

PR #1935, opened 2016-06-20
XL: Change ReadFile() to ensure returning 'length' bytes.

### LINKED ISSUE #1893: storageAPI:  New changes to ReadFile and AppendFile

harshavardhana:
ReadFile() returns a buffer of 'length' bytes as requested.
If length bytes are not available it returns ' < length ' bytes
but never more than what was requested.

Error of io.EOF is returned if no bytes were read, returns any
other errors as well.

Fixes #1893

### DIFF
--- a/erasure-utils.go
+++ b/erasure-utils.go
@@ -19,7 +19,6 @@ package main
 import (
 	"crypto/sha512"
 	"hash"
-	"io"
 
 	"github.com/klauspost/reedsolomon"
 )
@@ -44,21 +43,18 @@ func newHash(algo string) hash.Hash {
 	}
 }
 
+// hashSum calculates the hash of the entire path and returns.
 func hashSum(disk StorageAPI, volume, path string, writer hash.Hash) ([]byte, error) {
-	startOffset := int64(0)
-	// Read until io.EOF.
-	for {
-		buf := make([]byte, blockSizeV1)
-		n, err := disk.ReadFile(volume, path, startOffset, buf)
-		if err == io.EOF {
-			break
-		}
-		if err != nil {
-			return nil, err
-		}
-		writer.Write(buf[:n])
-		startOffset += n
+	fi, err := disk.StatFile(volume, path)
+	if err != nil {
+		return nil, err
 	}
+	var startOffset = int64(0)
+	var totalLeft = fi.Size
+	if err = writeN(disk, volume, path, startOffset, totalLeft, writer); err != nil {
+		return nil, err
+	}
+	// Return the final hash sum.
 	return writer.Sum(nil), nil
 }
 
--- a/fs-v1-multipart.go
+++ b/fs-v1-multipart.go
@@ -487,7 +487,6 @@ func (fs fsObjects) CompleteMultipartUpload(bucket string, object string, upload
 	}
 
 	tempObj := path.Join(tmpMetaPrefix, uploadID, "object1")
-	var buffer = make([]byte, blockSizeV1)
 
 	// Loop through all parts, validate them and then commit to disk.
 	for i, part := range parts {
@@ -505,22 +504,29 @@ func (fs fsObjects) CompleteMultipartUpload(bucket string, object string, upload
 		// Construct part suffix.
 		partSuffix := fmt.Sprintf("object%d", part.PartNumber)
 		multipartPartFile := path.Join(mpartMetaPrefix, bucket, object, uploadID, partSuffix)
-		offset := int64(0)
+		startOffset := int64(0)
 		totalLeft := fsMeta.Parts[partIdx].Size
 		for totalLeft > 0 {
-			var n int64
-			n, err = fs.storage.ReadFile(minioMetaBucket, multipartPartFile, offset, buffer)
+			// Figure out the right blockSize to read.
+			var curBlockSize int64
+			if blockSizeV1 < totalLeft {
+				curBlockSize = blockSizeV1
+			} else {
+				curBlockSize = totalLeft
+			}
+			var buf []byte
+			buf, err = fs.storage.ReadFile(minioMetaBucket, multipartPartFile, startOffset, curBlockSize)
 			if err != nil {
 				if err == errFileNotFound {
 					return "", InvalidPart{}
 				}
 				return "", toObjectErr(err, minioMetaBucket, multipartPartFile)
 			}
-			if err = fs.storage.AppendFile(minioMetaBucket, tempObj, buffer[:n]); err != nil {
+			if err = fs.storage.AppendFile(minioMetaBucket, tempObj, buf); err != nil {
 				return "", toObjectErr(err, minioMetaBucket, tempObj)
 			}
-			offset += n
-			totalLeft -= n
+			startOffset += int64(len(buf))
+			totalLeft -= int64(len(buf))
 		}
 	}
 
--- a/fs-v1.go
+++ b/fs-v1.go
@@ -188,25 +188,8 @@ func (fs fsObjects) GetObject(bucket, object string, startOffset int64, length i
 		return ObjectNameInvalid{Bucket: bucket, Object: object}
 	}
 	var totalLeft = length
-	for totalLeft > 0 {
-		// Figure out the right blockSize as it was encoded before.
-		var curBlockSize int64
-		if blockSizeV1 < totalLeft {
-			curBlockSize = blockSizeV1
-		} else {
-			curBlockSize = totalLeft
-		}
-		buf := make([]byte, curBlockSize)
-		n, err := fs.storage.ReadFile(bucket, object, startOffset, buf)
-		if err != nil {
-			return toObjectErr(err, bucket, object)
-		}
-		_, err = writer.Write(buf[:n])
-		if err != nil {
-			return toObjectErr(err, bucket, object)
-		}
-		totalLeft -= n
-		startOffset += n
+	if err := writeN(fs.storage, bucket, object, startOffset, totalLeft, writer); err != nil {
+		return toObjectErr(err, bucket, object)
 	}
 	return nil
 }
--- a/posix.go
+++ b/posix.go
@@ -352,12 +352,12 @@ func (s *posix) ListDir(volume, dirPath string) (entries []string, err error) {
 	return readDir(pathJoin(volumeDir, dirPath))
 }
 
-// ReadFile reads exactly len(buf) bytes into buf. It returns the
-// number of bytes copied. The error is EOF only if no bytes were
-// read. On return, n == len(buf) if and only if err == nil. n == 0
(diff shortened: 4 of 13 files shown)


posix.go
```diff
@@ -352,82 +352,97 @@ func (s *posix) ListDir(volume, dirPath string) (entries []string, err error) {
...
+	// Read file.
+	m, err := io.ReadFull(file, readBuf)
+	if err != nil {
+		// On EOF no bytes were read.
+		if err == io.EOF {
+			// Return io.EOF
+			return nil, io.EOF
+		}
+		// For any other errors return failure.
+		if err != io.ErrUnexpectedEOF {
```
krishnasrinivas, line 433: For ReadFile, if `m` < length, returning `nil` error indicates that there is more data to be read. Since ReadFile behaves like ReadFull we can set `io.ErrUnexpectedEOF` as the error. it also avoids the higher layer not to do an extra call.

harshavardhana, line 433: > For ReadFile, if m < length, returning nil error indicates that there is more data to be read. Since ReadFile behaves like ReadFull we can set io.ErrUnexpectedEOF as the error. it also avoids the higher layer not to do an extra call.

Makes sense.

krishnasrinivas, line 433: But returning `nil` makes the higher layer code simpler as it does not need to check for io.ErrUnexpectedEOF

for ex:

```
enBlocks[blockIndex], err = disk.ReadFile(volume, path, block*eInfo.BlockSize, eInfo.BlockSize)
if err != nil {
    ....
```

krishnasrinivas, line 433: i think returning `nil` itself like how current code does is better because of easier logic for higher layers.
this behavior is like File.Read() call

krishnasrinivas, line 433: > > For ReadFile, if m < length, returning nil error indicates that there is more data to be read. Since ReadFile behaves like ReadFull we can set io.ErrUnexpectedEOF as the error. it also avoids the higher layer not to do an extra call.
> 
> Makes sense.

OK then actually see what is better.

krishnasrinivas, line 433: > > > For ReadFile, if m < length, returning nil error indicates that there is more data to be read. Since ReadFile behaves like ReadFull we can set io.ErrUnexpectedEOF as the error. it also avoids the higher layer not to do an extra call.
> > 
> > Makes sense.
> 
> OK then actually see what is better.

actually regardless of whether we return `ErrUnexpectedEOF` or `nil` higher layer will always knows how many bytes to read (erasure-readfile gets length from the caller). So higher layer will be simple in either case. Since ReadFile behavior is similar to ReadFull `ErrUnexpectedEOF` will be better.

posix.go
```diff
@@ -352,82 +352,95 @@ func (s *posix) ListDir(volume, dirPath string) (entries []string, err error) {
...
-	// Read file.
-	m, err := io.ReadFull(file, buf)
+	// Allocate read buffer.
+	readBuf := make([]byte, length)
 
-	// Error unexpected is valid, set this back to nil.
-	if err == io.ErrUnexpectedEOF {
-		err = nil
+	// Read file.
+	m, err := io.ReadFull(file, readBuf)
```
krishnasrinivas, line 425: you can do:

```
m, err := io.ReadFull(file, readBuf)
return readBuf[:m], err
```

fs-v1-multipart.go
```diff
@@ -505,22 +504,32 @@ func (fs fsObjects) CompleteMultipartUpload(bucket string, object string, upload
...
-			if err != nil {
+			// Figure out the right blockSize to read.
+			var curBlockSize int64
+			if blockSizeV1 < totalLeft {
+				curBlockSize = blockSizeV1
+			} else {
+				curBlockSize = totalLeft
+			}
+			var buf []byte
+			buf, err = fs.storage.ReadFile(minioMetaBucket, multipartPartFile, startOffset, curBlockSize)
```
krishnasrinivas, line 518: `err` should never be non-nil here. so:

```
if err != nil {
    return...
}
```

will do

erasure-utils.go
```diff
@@ -44,21 +44,22 @@ func newHash(algo string) hash.Hash {
...
 
+// hashSum calculates the hash of the entire path and returns.
 func hashSum(disk StorageAPI, volume, path string, writer hash.Hash) ([]byte, error) {
 	startOffset := int64(0)
-	// Read until io.EOF.
+	// Read until io.EOF and write the read buffer to hash writer.
 	for {
-		buf := make([]byte, blockSizeV1)
-		n, err := disk.ReadFile(volume, path, startOffset, buf)
+		buf, err := disk.ReadFile(volume, path, startOffset, blockSizeV1)
```
krishnasrinivas: to avoid extra ReadFile() call (which gets io.EOF) to the disk this can be done:

```
func hashSum(disk StorageAPI, volume, path string, writer hash.Hash) ([]byte, error) { 
        startOffset := int64(0)                                                        
        // Read until io.EOF and write the read buffer to hash writer.                 
        for {                                                                          
                buf, err := disk.ReadFile(volume, path, startOffset, blockSizeV1)      
                if len(buf) > 0 {                                                      
                        writer.Write(buf)                                              
                        startOffset += int64(len(buf))                                 
                }                                                                      
                if err == io.EOF || err == io.ErrUnexpectedEOF {                       
                        break                                                          
                }                                                                      
                if err != nil {                                                        
                        return nil, err                                                
                }                                                                      
        }                                                                              
        // Return the final hash sum.                                                  
        return writer.Sum(nil), nil                                                    
}  
```

fs-v1.go
```diff
@@ -189,24 +189,26 @@ func (fs fsObjects) GetObject(bucket, object string, startOffset int64, length i
...
 		var curBlockSize int64
 		if blockSizeV1 < totalLeft {
 			curBlockSize = blockSizeV1
 		} else {
 			curBlockSize = totalLeft
 		}
-		buf := make([]byte, curBlockSize)
-		n, err := fs.storage.ReadFile(bucket, object, startOffset, buf)
-		if err != nil {
+		buf, err := fs.storage.ReadFile(bucket, object, startOffset, curBlockSize)
```
krishnasrinivas: here also err should never be non-nil, if it is then there is a bug somewhere.
so

```
if err != nil {
    return...
}
```

will do

xl-v1-utils.go
```diff
@@ -92,20 +89,18 @@ func readXLMeta(disk StorageAPI, bucket string, object string) ([]byte, error) {
 	var writer = new(bytes.Buffer)
 	startOffset := int64(0)
 
-	// Allocate 2MiB buffer, this is sufficient for the most of `xl.json`.
-	buf := make([]byte, 2*1024*1024)
-
 	// Read until io.EOF.
 	for {
```
krishnasrinivas: you can reuse readAll right?
`return readAll(disk, bucket, path.Join(object, xlMetaJSONFile))`

erasure-utils.go
harshavardhana: > if err == io.EOF || err == io.ErrUnexpectedEOF {

This is incorrect io.ErrUnexpectedEOF has data in it.

harshavardhana:
> return readAll(disk, bucket, path.Join(object, xlMetaJSONFile))

No we wanted to use a smaller sized buffer than 10MB a maximum of xl.json ever grows till 2MB+ for a 10,000 part upload. It is just an optimization.

erasure-utils.go
harshavardhana: > to avoid extra ReadFile() call (which gets io.EOF) to the disk this can be done:

to what benefit? should we avoid that?

fs-v1.go
harshavardhana: Yeah this will do.

harshavardhana:
Moved the code for readAll and writeN into storage-utils.go - readAll now always reads exactly total size, additionally does a stat to figure out the right size.

erasure-utils.go
krishnasrinivas: > This is incorrect io.ErrUnexpectedEOF has data in it.

it would have already done `writer.Write(buf)`

```
                if len(buf) > 0 {                                                      
                        writer.Write(buf)                                              
                        startOffset += int64(len(buf))                                 
                }                                                                      
                if err == io.EOF || err == io.ErrUnexpectedEOF {                       
                        break                                                          
                }           
```

krishnasrinivas: > > to avoid extra ReadFile() call (which gets io.EOF) to the disk this can be done:
> 
> to what benefit? should we avoid that?

because the extra disk call is redundant and avoiding it does not complicate the code. i.e there is no benefit in doing the extra call.

storage-utils.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+	return nil
+}
+
+// readAll - returns contents from volume/path as byte array.
+func readAll(disk StorageAPI, volume string, path string) ([]byte, error) {
+	fi, err := disk.StatFile(volume, path)
+	if err != nil {
+		return nil, err
+	}
```
krishnasrinivas, line 56: you can just do:
`return disk.ReadFile(volume, path, int64(0), fi.Size)`

xl-v1-utils.go
```diff
@@ -64,48 +61,7 @@ func randInts(count int) []int {
...
-			break
-		}
-		if err != nil && err != io.EOF {
-			return nil, err
-		}
-		writer.Write(buf[:n])
-		startOffset += n
-	}
-	return writer.Bytes(), nil
+	return readAll(disk, bucket, object)
```
krishnasrinivas: need to do this! :
readAll(disk, bucket, path.Join(object, xlMetaJSONFile))

krishnasrinivas: `readXLMeta()` can actually return xlMetaV1{} instead of []byte

harshavardhana:
> return disk.ReadFile(volume, path, int64(0), fi.Size)

Yes that is right, let the lower level return the entire buffer.

krishnasrinivas:
looks good.

abperiasamy:
Wait till @krishnasrinivas parallel read PR.

harshavardhana:
This change has been deferred and not needed anymore.. Closing

PR #1936, opened 2016-06-20
INSTALLGO.md mentions Go 1.5+ for OS X

yrashk:
However, current requirement is 1.6, so the file has been updated to reflect that.


harshavardhana:
Thanks @yrashk

PR #1941, opened 2016-06-20
XL: fix getBlockInfo() to return correct end block

balamurugana:
If requested offset/length of an object is equal to
erasureInfo.BlockSize, getBlockInfo() returns one more block added to
actual end block.  This patch fixes the issue.

This patch also adds unit test for get objects with big files.

### DIFF
--- a/erasure-utils.go
+++ b/erasure-utils.go
@@ -96,6 +96,9 @@ func getBlockInfo(offset, length, blockSize int64) (startBlock, bytesToSkip, end
 
 	// Calculate end block for given size to read
 	endBlock = (offset + length) / blockSize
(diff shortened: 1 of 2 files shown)


erasure-utils.go
```diff
@@ -96,6 +96,9 @@ func getBlockInfo(offset, length, blockSize int64) (startBlock, bytesToSkip, end
 
 	// Calculate end block for given size to read
 	endBlock = (offset + length) / blockSize
+	if endBlock > 0 && (offset+length)%blockSize == 0 {
+		endBlock--
```
harshavardhana, line 100: Can you also add a comment here.

PR #1943, opened 2016-06-20
Make ioErrCount updates go-routine safe

krisis:


posix.go
```diff
@@ -39,7 +40,7 @@ const (
 type posix struct {
 	diskPath    string
 	minFreeDisk int64
-	ioErrCount  int
+	ioErrCount  *int32
```
harshavardhana: you can keep this as int32 and use '&s.ioErrCount'

PR #1945, opened 2016-06-20
XL: Avoid memory copy, write to writer directly all the dataBlocks.

harshavardhana:
getDataBlocks internally uses bytes avoid it by using writer
directly and write to it.


krisis:
Looks good to me.

harshavardhana:
Already fixed by #1959

PR #1950, opened 2016-06-21
Added couple of unit-tests to xlObjects

krisis:


xl-v1-object_test.go
```diff
@@ -0,0 +1,146 @@
...
+import (
+	"bytes"
+	"crypto/md5"
+	"encoding/hex"
+	"io/ioutil"
+	"os"
+	"testing"
+)
+
+func initializeObjLayer() (objLayer ObjectLayer, err error) {
```
hackintoshrao: The functionalities [here](https://github.com/minio/minio/blob/master/test-utils_test.go#L330) can be used.  Here is a sample implementation [here](https://github.com/minio/minio/blob/master/object-api-listobjects_test.go)

xl-v1-object_test.go
```diff
@@ -20,40 +20,13 @@ import (
...
-}
-
-func cleanupObjLayer(xl xlObjects) {
-	disks := xl.physicalDisks
-	for i := range disks {
-		os.RemoveAll(disks[i])
-	}
-}
-
 func failDisks(xl xlObjects, n int) (removedDisks []StorageAPI) {
```
harshavardhana, line 26: why not do `removeRandomDisk()` ?

krisis, line 26: Since the `erasureInfo.Distribution` randomizes the placement of data and parity blocks amongst the disks, it suffices to pick any `k` successive disks out of the total.

PR #1967, opened 2016-06-23
tests: add unit test for posix.readDir()

### LINKED ISSUE #1820: XL: Tests: Unit tests for POSIX layer functionalities.
Need good coverage for posix layer functionalities, `posix-*` files on the repo.

balamurugana:
Fixes #1820

### DIFF
--- a/posix-list-dir_test.go
+++ b/posix-list-dir_test.go
@@ -1,0 +1,157 @@
+/*
+ * Minio Cloud Storage, (C) 2016 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"sort"
+	"testing"
+)
+
+// Test to check for different input arguments.
+func TestReadDir0(t *testing.T) {
+	// Check non existent directory.
+	if _, err := readDir("/tmp/non-existent-directory"); err != errFileNotFound {
+		t.Fatalf("expected = %s, got: %s", errFileNotFound, err)
+	}
+
+	// Check if file is given.
+	if _, err := readDir("/etc/issue/mydir"); err != errFileNotFound {
+		t.Fatalf("expected = %s, got: %s", errFileNotFound, err)
+	}
+
+	// Check if permission denied.
+	if _, err := readDir("/proc/1/fd"); err == nil {
+		t.Fatalf("expected = an error, got: nil")
+	}
+}
+
+type result struct {
+	dir     string
+	entries []string
+}
+
+var testResults = []result{}
+
+// Test to read empty directory.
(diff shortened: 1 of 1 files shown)


posix-list-dir-others.go
```diff
@@ -28,6 +29,7 @@ import (
 func readDir(dirPath string) (entries []string, err error) {
 	d, err := os.Open(preparePath(dirPath))
 	if err != nil {
+		fmt.Println(">>> DEBUG:", dirPath)
```
harshavardhana: is this necessary?

posix-list-dir_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+	s.dir = dir
+}
+
+func (s *PosixTestSuite) TeardownTestReadDir1(c *C) {
+	os.RemoveAll(s.dir)
+}
+
+// Test to read empty directory.
+func (s *PosixTestSuite) TestReadDir1(c *C) {
```
hackintoshrao: Instead of multiple test functions, its better to generate a array of struct with all inputs and expected outputs and iterate through them, here is an example https://github.com/minio/minio/blob/master/object-api-multipart_test.go#L132 . This allows for adding more cases easily.

posix-list-dir_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+	. "gopkg.in/check.v1"
+)
+
+type PosixTestSuite struct {
+	dir   string
+	names []string
+}
+
+var _ = Suite(&PosixTestSuite{})
```
hackintoshrao: We have followed the style of using Go's testing package for unit tests and `gocheck suite` only for functional tests. It'll great to use just the testing package for unit tests. Any setup/taerdown can be made inside the test itself, or feel free to add an util function to do so in `test-utils_test.go`

posix-list-dir_test.go
balamurugana: If you see, each test has its own setup/input separately.  The combined tests of all looks unclean.

posix-list-dir-others.go
balamurugana: Oops! I missed removing this debug line.  Fixing it.

posix-list-dir_test.go
```diff
@@ -0,0 +1,137 @@
...
+			return false
+		}
+	}
+
+	return true
+}
+
+func TestReadDir(t *testing.T) {
+	defer teardown()
+	setupTestReadDir1(t)
```
hackintoshrao, line 142: The setup data can collected into cases like these https://github.com/minio/minio/blob/master/object-api-multipart_test.go#L1158 . Instead of 3 different setup functions, the setup data can be collected into an array as above and then you can iterate over them to complete the setup like shown in the above link.

posix-list-dir_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+	for i := range expected {
+		if expected[i] != got[i] {
+			return false
+		}
+	}
+
+	return true
+}
+
```
hackintoshrao, line 138: A Comment on the intent of the test needed.

posix-list-dir_test.go
```diff
@@ -0,0 +1,137 @@
...
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"sort"
+	"testing"
+)
+
```
hackintoshrao, line 27: A Comment on the intent of the test needed.

posix-list-dir_test.go
```diff
@@ -0,0 +1,137 @@
...
+		testResults = append(testResults, result{dir, entries})
+	}
+}
+
+func teardown() {
+	for _, r := range testResults {
+		os.RemoveAll(r.dir)
+	}
+}
+
```
hackintoshrao, line 120: A comment needed.

hackintoshrao:
Good to go for the first cut. We can make it more organized in the next iteration. LGTM.

Issue #1825, opened 2016-06-01
XL: Tests: tree-walk tests for fs and XL. 


hackintoshrao:
Tests for tree-walk implementation of fs and XL needed. Its a high priority test.

GarimaKapoor:
please put a timeline for the fix.

balamurugana:
To be fixed: 11-Jun-2016

GarimaKapoor:
Please update the timeline

krisis:
Will try to send a PR for this issue by Jun 23rd.

PR #1969, opened 2016-06-23
Added unit-tests for treeWalkPool

krisis:
#1825


tree-walk-pool_test.go
```diff
@@ -0,0 +1,100 @@
...
+	select {
+	case <-time.After(1 * time.Second):
+		break
+	default:
+		if c1, _ := tw.Release(params); c1 == nil {
+			t.Error("treeWalk go-routine got freed before timeout")
+		}
+	}
+}
+
```
hackintoshrao, line 59: Need a brief Summary of Tests intent here.

tree-walk-pool_test.go
```diff
@@ -0,0 +1,100 @@
...
+ * limitations under the License.
+ */
+
+package main
+
+import (
+	"testing"
+	"time"
+)
+
```
hackintoshrao, line 23: Since its a exposed function, it looks very plain without an comment. A comment with brief summary of the purpose of the test would look good

hackintoshrao:
Tests need more comments with the intent for each step/case like followed here https://github.com/minio/minio/blob/master/object-api-multipart_test.go#L108 . Tests are entry points for team and community to start understanding the code, so they gotto be more readable than the code itself with exhaustive commenting for each step.

tree-walk-fs_test.go
```diff
@@ -0,0 +1,105 @@
...
+		if !strings.HasPrefix(res.entry, "d/") {
+			t.Errorf("Entry %s doesn't match prefix d/", res.entry)
+		}
+	}
+}
+
+// Test if entries received on tree walk's channel appear after the supplied marker
+func TestTreeWalkMarker(t *testing.T) {
+	// Create a FS-backend
+	objLayer, disks, err := getSingleNodeObjectLayer()
```
hackintoshrao: You can use the `ExecObjectLayerTest` wrapper to run your test on both XL and FS setup and then to do the cleanup, here is an example https://github.com/minio/minio/blob/master/object-api-multipart_test.go#L229 .

tree-walk-fs_test.go
```diff
@@ -0,0 +1,105 @@
...
+package main
+
+import (
+	"bytes"
+	"strings"
+	"testing"
+)
+
+// Test if tree walker returns entries matching prefix alone are received
+// when a non empty prefix is supplied
```
hackintoshrao: a full stop at the end of comments required for all.

tree-walk-fs_test.go
krisis: @hackintoshrao, this file is dedicated for unit tests for treeWalk implementation for FS backend. It doesn't make sense to run these tests for both XL and FS.

hackintoshrao: @krisis : Hmm, then can you split `ExecObjectLayerTest` to use 2 functions. One is setup FS and call the test and the other to setup `XL` and call the test. Then you can use `FS` util to call the test. Its better to separate them as test util functionalities.

hackintoshrao: @krisis : I'm working on this split. Necessary for benchmarking.

krisis: @hackintoshrao OK. Is there anything else pending to be done in this PR?

With this PR the coverage for `tree-walk-pool-fs.go` has moved from 64.7% to 97.1%.
See `play/krisis/treewalk-pool.html` and `play/krisis/ff9fc22-cov.html` for comparison.

hackintoshrao: @krisis : Take a look at `makeTestBackend` from teset-utils_test.go. That should help get just a FS/XL instance.

krishnasrinivas:
@krisis https://github.com/minio/minio/pull/2027

PR #1971, opened 2016-06-24
pkg: Remove unused safe package.

harshavardhana:


harshavardhana:
Still being used by pkg/quick closing.

PR #1974, opened 2016-06-24
XL: Implement new ReadAll API for files which are read in single call.

harshavardhana:


posix.go
```diff
@@ -352,6 +352,94 @@ func (s *posix) ListDir(volume, dirPath string) (entries []string, err error) {
...
+	buf := bytes.NewBuffer(make([]byte, 0, capacity))
+	// If the buffer overflows, we will get bytes.ErrTooLarge.
+	// Return that as an error. Any other panic remains.
+	defer func() {
+		e := recover()
+		if e == nil {
+			return
+		}
+		if panicErr, ok := e.(error); ok && panicErr == bytes.ErrTooLarge {
+			err = panicErr
```
abperiasamy: return(panicErr)
to be consistent with other return.

posix.go
```diff
@@ -352,6 +352,94 @@ func (s *posix) ListDir(volume, dirPath string) (entries []string, err error) {
...
+	// Close the reader.
+	defer file.Close()
+
+	// Read all.
+	return readAll(file, st.Size())
+}
+
+// readAll reads from r until an error or EOF and returns the data it read
+// from the internal buffer allocated with a specified capacity.
+func readAll(r io.Reader, capacity int64) (b []byte, err error) {
```
abperiasamy: I am thinking if really need this protection?. Since this ReadAll is only used for reading metadata, we can simply use ioutil.ReadAll.

PR #1977, opened 2016-06-24
server: Add connection rate limiter for server.

harshavardhana:


rate-limit-handler.go
```diff
@@ -0,0 +1,72 @@
...
+	c.handler.ServeHTTP(w, r)
+
+	// Release by draining the channel once.
+	c.releaseOnce.Do(c.release)
+}
+
+// setRateLimitHandler limits the number of concurrent http requests based
+// on MINIO_MAXCONN.
+func setRateLimitHandler(handler http.Handler) http.Handler {
+	maxConnStr := os.Getenv("MINIO_MAXCONN")
```
abperiasamy: Parse MINIO_MAXCONN env variable during the server startup and set globalMaxConn. If it is not set, then setRateLimitHandler will be NOOP.

globals.go
```diff
@@ -41,6 +41,13 @@ var (
 	// Add new global flags here.
 )
 
+var (
```
abperiasamy: this can be merged into globalTrace and Quiet flag list.

server-main.go
```diff
@@ -137,6 +137,15 @@ func initServerConfig(c *cli.Context) {
...
 
+	// Fetch max conn limit from environment variable.
+	maxConnStr := os.Getenv("MINIO_MAXCONN")
+	if maxConnStr != "" {
+		// We need to parse to its integer value.
+		maxConn, err := strconv.Atoi(maxConnStr)
+		fatalIf(err, "Unable to convert MINIO_MAXCONN=%s environment variable into its integer value.", maxConnStr)
+		globalMaxConn = maxConn
+	}
+
```
abperiasamy, line 143: // Enable connection limit if MINIO_MAXCONN is set.
if maxConnStr := os.Getenv("MINIO_MAXCONN");  maxConnStr != "" {
        globalMaxConn, err := strconv.Atoi(maxConnStr)
        fatalIf(err, "Unable to convert MINIO_MAXCONN=%s environment variable into its integer value.", maxConnStr)
}

PR #1985, opened 2016-06-26
main: minio --help should print regardless of root.

### LINKED ISSUE #1964: reading --help output shouldn't require to use non-root user.
## Expected behaviour

I ran `minio --help` as root. I was expecting to see the `--help` output, probably with a warning to run as root.
## Actual behaviour

No output, just a denial to display anything as root.

Install minio. Try to read the `--help` output as root.
## Minio version
  Version: 2016-06-03T19:32:05Z
  Release-Tag: RELEASE.2016-06-03T19-32-05Z
  Commit-ID: 73ddb5be758566dde18b519393c7554ae280b074
## System information

Linux enceladus 3.10.0-229.el7.x86_64 #1 SMP Fri Mar 6 11:36:42 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux

harshavardhana:
Remove root check entirely.

Fixes #1964

### DIFF
--- a/main.go
+++ b/main.go
@@ -64,12 +64,6 @@ func init() {
 
 	// Set global trace flag.
 	globalTrace = os.Getenv("MINIO_TRACE") == "1"
-
-	// It is an unsafe practice to run network services as
-	// root. Containers are an exception.
-	if !isContainerized() && os.Geteuid() == 0 {
-		console.Fatalln("Please run ‘minio’ as a non-root user.")
-	}
 }
 
 func migrate() {
--- a/runtime-checks.go
+++ b/runtime-checks.go
@@ -18,37 +18,11 @@ package main
 
 import (
 	"fmt"
-	"io/ioutil"
-	"os"
 	"runtime"
-	"strings"
 
 	"github.com/hashicorp/go-version"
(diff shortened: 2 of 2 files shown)


globals.go
```diff
@@ -38,6 +38,10 @@ const (
 var (
 	globalQuiet = false // Quiet flag set via command line
 	globalTrace = false // Trace flag set via environment setting.
+
+	// By default root is disallowed.
+	globalRootDisallow = true
+
```
abperiasamy: global variables are useful to save states for the life of a program. In this case, this would mean that certain functions that are not allowed in root-mode will alone be disabled. Simpler approach is to simply check at the right place and exit.

main.go
```diff
@@ -155,6 +153,11 @@ func mustGetProfilePath() string {
 func main() {
 	app := registerApp()
 	app.Before = func(c *cli.Context) error {
+		// Check if running as 'root' is allowed.
+		if globalRootDisallow {
```
abperiasamy: here is where you would check and exit directly. Another alternative is to remove isContainerized and root check. Leave it up to the user to take care. More and more Unix privileges are becoming irrelevant.

PR #1986, opened 2016-06-26
XL: Handle quorum situations properly for write operations.

### LINKED ISSUE #1951: XL: With insufficient write quorum, put object still succeeds
In a 8 disks setup, where three disks are down, put object still succeeds.   As per write quorum calculation, the setup can withstand max of no.of disks - no. of disks / 2 + 2 ie 2 disk failure.

harshavardhana:
Adds two test cases one for
- PutObject when write quorum is not available.
- PutObjectPart when write quorum is not available.

Fixes #1951


object-api-multipart_test.go
```diff
@@ -99,6 +99,88 @@ func testObjectAPIIsUploadIDExists(obj ObjectLayer, instanceType string, t *test
...
 
+// Wrapper for calling TestPutObjectPartDiskNotFound tests for both XL
+// write quorum.
+func TestPutObjectPartDiskNotFound(t *testing.T) {
+	ExecObjectLayerDiskNotFoundTest(t, testPutObjectPartDiskNotFound)
+}
+
+// testPutObjectPartDiskNotFound - Tests validate PutObjectPart behavior when disks go offline.
+func testPutObjectPartDiskNotFound(obj ObjectLayer, instanceType string, disks []string, t *testing.T) {
+	bucketNames := []string{"minio-bucket", "mibnio-2-bucket"}
```
krisis: s/mibnio/minio

object-api-multipart_test.go
```diff
@@ -99,6 +99,88 @@ func testObjectAPIIsUploadIDExists(obj ObjectLayer, instanceType string, t *test
...
+	objectNames := []string{"minio-object-1.txt"}
+	uploadIDs := []string{}
+
+	// bucketnames[0].
+	// objectNames[0].
+	// uploadIds [0].
+	// Create bucket before intiating NewMultipartUpload.
```
krisis, line 117: s/intiating/initiating/

object-api-putobject_test.go
```diff
@@ -0,0 +1,213 @@
...
+		if actualErr == nil && testCase.shouldPass {
+			// Asserting whether the md5 output is correct.
+			if testCase.inputMeta["md5Sum"] != actualMd5Hex {
+				t.Errorf("Test %d: %s: Calculated Md5 different from the actual one %s.", i+1, instanceType, actualMd5Hex)
+			}
+		}
+	}
+
+	// This causes quorum failure verify.
+	removeAll(disks[len(disks)-1])
```
krisis, line 197: Wouldn't it be simpler to move the last test case from the `testcases` slice, since we perform different checks instead of restting index to point to the last element in the slice?

object-api-putobject_test.go
```diff
@@ -0,0 +1,213 @@
...
+		{bucket, object, "efgh", map[string]string{"md5Sum": "1f7690ebdd9b4caf8fab49ca1757bf27"}, int64(len("efgh")), true, "", nil},
+		{bucket, object, "ijkl", map[string]string{"md5Sum": "09a0877d04abf8759f99adec02baf579"}, int64(len("ijkl")), true, "", nil},
+		{bucket, object, "mnop", map[string]string{"md5Sum": "e132e96a5ddad6da8b07bba6f6131fef"}, int64(len("mnop")), true, "", nil},
+		{bucket, object, "mnop", map[string]string{"md5Sum": "e132e96a5ddad6da8b07bba6f6131fef"}, int64(len("mnop")), false, "", InsufficientWriteQuorum{}},
+	}
+
+	for i, testCase := range testCases[:3] {
+		actualMd5Hex, actualErr := obj.PutObject(testCase.bucketName, testCase.objName, testCase.intputDataSize, bytes.NewBufferString(testCase.inputReaderData), testCase.inputMeta)
+		if actualErr != nil && testCase.shouldPass {
+			t.Errorf("Test %d: %s: Expected to pass, but failed with: <ERROR> %s.", i+1, instanceType, actualErr.Error())
```
krisis, line 175: Question: Would it make sense to make comparison functions for fields of types like partInfo et al, that we use often?

hackintoshrao:
LGTM 👍

object-api-putobject_test.go
harshavardhana, line 197: > Wouldn't it be simpler to move the last test case from the testcases slice, since we perform different checks instead of restting index to point to the last element in the slice?

Done

PR #2005, opened 2016-06-27
API: add writePartTooSmallErrorResponse to extend standard error responses.

### LINKED ISSUE #1536: handlers: APIErrorResponse does not have all the fields
with S3, PutObjectPart succeeds for parts with < 5MB
But complete multipart failse with:

```
{ [S3Error: Your proposed upload is smaller than the minimum allowed size]
  code: 'EntityTooSmall',
  proposedsize: '1048576',
  minsizeallowed: '5242880',
  partnumber: '1',
  etag: 'b6d81b360a5672d80c27430f39153e2c',
  requestid: 'A5675AEAF4F07375',
  hostid: 'Ei6zxS8gWBHnK8TYsc/+TAlwftj9d6GFinPeh+4uMfEuHWgUOCJRil6B8ppjC65Yr37UB/d30yc=',
  amzRequestid: null,
  amzId2: null,
  amzBucketRegion: null } undefined

```

we are not able to send partnumber, proposedsize and minsizeallwoed in response XML as APIErrorResponse does not have those fields.

harshavardhana:
This function is added to extend the standard error responses.
Which is needed in some cases for example CompleteMultipartUpload
should respond with ErrPartTooSmall error when parts uploaded are
lesser than 5MB (i.e minimum allowed size per part).

Fixes #1536

### DIFF
--- a/api-response-multipart.go
+++ b/api-response-multipart.go
@@ -1,0 +1,54 @@
+/*
+ * Minio Cloud Storage, (C) 2016 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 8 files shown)


krisis:
Looks good to me.

PR #2007, opened 2016-06-27
XL: allow meta bucket name appended with tmp meta prefix.

balamurugana:


harshavardhana:
@balamurugana - can you create test to catch this possibility?

harshavardhana:
// cc @krishnasrinivas - suggested we need to handle this case for 'multipart' directory as well.

krishnasrinivas:
> // cc @krishnasrinivas - suggested we need to handle this case for 'multipart' directory as well.

the same approach will not work for multipart because we can have a path like `.minio/multipart/dir1/dir2/dir3/obj` and after rename we will be left with directory `.minio/multipart/dir1/dir2/dir3/`

harshavardhana:
Can you rebase this @balamurugana ?

harshavardhana:
> the same approach will not work for multipart because we can have a path like .minio/multipart/dir1/dir2/dir3/obj and after rename we will be left with directory .minio/multipart/dir1/dir2/dir3/

It is validated that this is not true 'multipart' is indeed removed properly and not left over.  It is the 'tmp' that needs to be addressed 
- FS 'tmp' stays for PutObject and Multipart
- XL 'tmp' stays only for PutObject

xl-v1-object.go
```diff
@@ -400,18 +401,18 @@ func (xl xlObjects) PutObject(bucket string, object string, size int64, data io.
...
 	// Rename the successfully written temporary object to final location.
-	err = xl.renameObject(minioMetaBucket, tempObj, bucket, object)
+	err = xl.renameObject(minioMetaTmpBucket, tempObj, bucket, object)
 	if err != nil {
 		return "", toObjectErr(err, bucket, object)
 	}
 
 	// Delete the temporary object.
-	xl.deleteObject(minioMetaBucket, path.Join(tmpMetaPrefix, newUniqueID))
+	xl.deleteObject(minioMetaTmpBucket, newUniqueID)
```
harshavardhana, line 424: FS code also needs to be modified with this fix.

harshavardhana:
@balamurugana - tested this, this still leaves the 'tmp' directory on FS,  did we agree that leaving 'tmp' is okay? i forgot.

krishnasrinivas:
> @balamurugana - tested this, this still leaves the 'tmp' directory on FS, did we agree that leaving 'tmp' is okay?

its not ok on FS, its ok on XL.

harshavardhana:
> its not ok on FS, its ok on XL.

With this argument then we shouldn't store '.minio' either , '.minio' was agreed to keep state, so possibly we can keep 'tmp' as well which is inside '.minio' anyways. 

Since upon shutdown these are cleaned properly anyways, the fix is right in anycase.

PR #2008, opened 2016-06-27
XL/erasure: refactor erasureReadFile. Move parallelRead into a separate function.

krishnasrinivas:
- getReadDisk was not a recursive function. Hence parallelRead as recursion looked odd and forced.
- move parallelRead() as a separate function so that it can be unit tested.

### DIFF
--- a/erasure-readfile.go
+++ b/erasure-readfile.go
@@ -125,6 +125,50 @@ func getReadDisks(orderedDisks []StorageAPI, index int, dataBlocks int) (readDis
 	return nil, 0, errXLReadQuorum
 }
 
+// parallelRead - reads chunks in parallel from the disks specified in []readDisks.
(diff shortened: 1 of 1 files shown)


erasure-readfile.go
```diff
@@ -125,6 +125,50 @@ func getReadDisks(orderedDisks []StorageAPI, index int, dataBlocks int) (readDis
 	return nil, 0, errXLReadQuorum
 }
 
+// parallelRead - reads chunks in parallel from the disks specified in []readDisks.
+func parallelRead(volume, path string, readDisks []StorageAPI, orderedDisks []StorageAPI, enBlocks [][]byte, blockOffset int64, curChunkSize int64, bitRotVerify func(diskIndex int) bool) {
```
harshavardhana, line 129: Can you explore the possibility of adding a unit test for this code path?

krishnasrinivas, line 129: will do

krishnasrinivas, line 129: @harshavardhana will add the unit test as part of https://github.com/minio/minio/issues/2010

harshavardhana:
> @harshavardhana will add the unit test as part of #2010

Okay

PR #2009, opened 2016-06-27
api: Sent ErrPreconditionFailed on If-Match failure

### LINKED ISSUE #1981: FS: GetObject wrong responses for If-Match and If-None-Match
## Expected behaviour

If-Match when given a bad ETag should give back 412 Precondition Failed XML error. Instead it is sending the object back.

If-None-Match when given a good ETag should give back 304 Not Modified and no object. Instead it is giving 200 OK and sending the object back. 
## Actual behaviour

If-Match is sending the object back.
If-None-Match is sending the object back.

run the tests for GetObject on s3verify
## Minio version

krisis:
ref: http://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html#ErrorCodeList

Fixes #1981 (previous fix didn't add error response to response.Body on If-Match condition failure)

### DIFF
--- a/server_xl_test.go
+++ b/server_xl_test.go
@@ -841,6 +841,78 @@ func (s *MyAPIXLSuite) TestNotBeAbleToCreateObjectInNonexistentBucket(c *C) {
 	verifyError(c, response, "NoSuchBucket", "The specified bucket does not exist.", http.StatusNotFound)
 }
 
+// TestGetOnObject - Asserts properties for GET on an object.
+// GET requests on an object retrieves the object from server.
+// Tests behaviour when If-Match/If-None-Match headers are set on the request
+func (s *MyAPIXLSuite) TestGetOnObject(c *C) {
+	// generate a random bucket name.
+	bucketName := getRandomBucketName()
+	// make HTTP request to create the bucket.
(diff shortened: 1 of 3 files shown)


harshavardhana:
Would you mind adding a test case? 

```
func (s *MyAPISuite) TestHeadOnObject(c *C) {
```

Already tests for If-Modified-Since.

server_xl_test.go
```diff
@@ -841,6 +841,78 @@ func (s *MyAPIXLSuite) TestNotBeAbleToCreateObjectInNonexistentBucket(c *C) {
...
+	c.Assert(err, IsNil)
+	c.Assert(string(body), Equals, "hello world")
+
+	// GetObject with If-Match sending mismatching etag in request headers
+	// is expected to return an error response with ErrPreconditionFailed.
+	request, err = newTestRequest("GET", s.endPoint+"/"+bucketName+"/object1",
+		0, nil, s.accessKey, s.secretKey)
+	request.Header.Set("If-Match", etag[1:])
+	response, err = client.Do(request)
+	c.Assert(response.StatusCode, Equals, http.StatusPreconditionFailed)
```
harshavardhana: you can use 'verifyError' wrapper here.

Issue #2002, opened 2016-06-27
Tests: Comments and cleanup for server_xl_test.go


hackintoshrao:
In progress......

PR #2011, opened 2016-06-27
Cleanup: Adding detailed comments for server_xl_test.

hackintoshrao:
Fixes https://github.com/minio/minio/issues/2002

### DIFF
--- a/object-api-multipart_test.go
+++ b/object-api-multipart_test.go
@@ -1726,7 +1726,7 @@ func testObjectCompleteMultipartUpload(obj ObjectLayer, instanceType string, t *
 	uploadIDs = append(uploadIDs, uploadID)
 	// Parts with size greater than 5 MB.
 	// Generating a 6MB byte array.
-	validPart := bytes.Repeat([]byte("0123456789abcdef"), 1024*1024)
+	validPart := bytes.Repeat([]byte("abcdef"), 1024*1024)
 	validPartMD5 := findMD5(validPart)
 	// Create multipart parts.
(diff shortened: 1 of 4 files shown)


object-api-multipart_test.go
```diff
@@ -1725,7 +1725,7 @@ func testObjectCompleteMultipartUpload(obj ObjectLayer, instanceType string, t *
 
 	uploadIDs = append(uploadIDs, uploadID)
 	// Parts with size greater than 5 MB.
-	// Generating a 6MB byte array.
+	// Generating a 16MB byte array.
 	validPart := bytes.Repeat([]byte("0123456789abcdef"), 1024*1024)
```
harshavardhana: Is there a particular reason to make this 16MB? if they are valid parts why not just make it like 6MB or 8MB, 16MB for each part is not necessary for these tests.

test-utils_test.go
```diff
@@ -439,6 +439,60 @@ func getDeleteBucketURL(endPoint, bucketName string) string {
...
+// return URL for completing multipart upload.
+// complete multipart upload request is sent after all parts are uploaded.
+func getCompleteMultipartUploadURL(endPoint, bucketName, objectName, uploadID string) string {
+	queryValue := url.Values{}
+	queryValue.Set("uploadId", uploadID)
+	return makeTestTargetURL(endPoint, bucketName, objectName, queryValue)
+}
+
 // returns temp root directory. `
 func getTestRoot() (string, error) {
```
harshavardhana, line 497: Everything else looks clean.

harshavardhana:
One question apart from that everything else looks clean.

PR #2027, opened 2016-06-28
tree-walk: unify FS and XL tree-walk with functional approach.

krishnasrinivas:
getListDir, doTreeWalk and startTreeWalk are mere functions now without any dependency on FS/XL.
More easily unit-testable now.


tree-walk.go
```diff
@@ -28,51 +28,62 @@ type treeWalkResult struct {
...
+			entries, err = disk.ListDir(bucket, prefixDir)
+			if err != nil {
+				// For any reason disk was deleted or goes offline, continue
+				// and list form other disks if possible.
+				if err == errDiskNotFound || err == errFaultyDisk {
+					continue
+				}
+				break
 			}
-			if strings.HasSuffix(entry, slashSeparator) && isLeaf(bucket, pathJoin(prefixDir, entry)) {
```
krisis, line 58: The `strings.HasSuffix` check has been removed. How is it handled now?

krishnasrinivas, line 58: `strings.HasSuffix` here was never needed as isLeaf will take care of it

fs-v1-multipart.go
```diff
@@ -70,9 +70,8 @@ func (fs fsObjects) listMultipartUploads(bucket, prefix, keyMarker, uploadIDMark
 		walkResultCh, endWalkCh := fs.listPool.Release(listParams{minioMetaBucket, recursive, multipartMarkerPath, multipartPrefixPath})
 		if walkResultCh == nil {
 			endWalkCh = make(chan struct{})
-			walkResultCh = fs.startTreeWalk(minioMetaBucket, multipartPrefixPath, multipartMarkerPath, recursive, func(bucket, object string) bool {
-				return fs.isMultipartUpload(bucket, object)
-			}, endWalkCh)
+			listDir := getListDir(fs.isMultipartUpload, fs.storage)
```
harshavardhana: getListDir() should be named appropriately here, perhaps like a factory or something. 

listDirFactory() would make sense it is returning a listDirFn

tree-walk.go
```diff
@@ -28,51 +28,62 @@ type treeWalkResult struct {
...
-			// For any reason disk was deleted or goes offline, continue
-			// and list form other disks if possible.
-			if err == errDiskNotFound || err == errFaultyDisk {
+// "listDir" function of type listDirFn returned by getListDir() - explained below.
+type listDirFn func(bucket, prefixDir string, filter func(entry string) bool) (entries []string, err error)
+
+// Returns function "listDir" of the type listDirFn.
+// isLeaf - is used by listDir function to check if an entry is a leaf or non-leaf entry.
+// disks - used for doing disk.ListDir(). FS passes single disk argument, XL passes a list of disks.
+func getListDir(isLeaf func(string, string) bool, disks ...StorageAPI) listDirFn {
```
harshavardhana: isLeaf should come as the last argument here think from the filepath.Walk()

tree-walk.go
```diff
@@ -28,51 +28,62 @@ type treeWalkResult struct {
...
+	// listDir - lists all the entries at a given prefix, takes additional filter function as parameter.
```
harshavardhana: the disks slice can be filtered out here. we don't need to do 

```
for _, disk := range disks {
      if disk == nil {
             continue
      }
}
```

For each loop.

tree-walk.go
```diff
@@ -28,51 +28,62 @@ type treeWalkResult struct {
...
-		if disk == nil {
-			continue
-		}
-		entries, err = disk.ListDir(bucket, prefixDir)
-		if err != nil {
-			// For any reason disk was deleted or goes offline, continue
-			// and list form other disks if possible.
-			if err == errDiskNotFound || err == errFaultyDisk {
+// "listDir" function of type listDirFn returned by getListDir() - explained below.
+type listDirFn func(bucket, prefixDir string, filter func(entry string) bool) (entries []string, err error)
```
harshavardhana: Call this 'listDirFunc' use the short form for the initialized func. 

```
listDirFn := func()
```

tree-walk.go
krishnasrinivas: as `getListDir` is a variadic function `disks ...StporageAPI` has to be the last argument

fs-v1-multipart.go
krishnasrinivas: will call it listDirFactory()

tree-walk.go
krishnasrinivas: everywhere we are passing the disks slice as is and functions take care of skipping the disk if it is `nil` hence better to be consistent, so the onus of filtering of nil disks is never on the caller.

tree-walk.go
krishnasrinivas: will call it `listDirFunc`

tree-walk_test.go
```diff
@@ -131,142 +84,178 @@ func testTreeWalkMarker(obj ObjectLayer, instanceType string, t *testing.T) {
...
+	entries, err = listDir(volume, "", "")
+	if err != errDiskNotFound {
+		t.Error("expected errDiskNotFound error.")
+	}
 }
-
-// FIXME: Test the abort timeout when the tree-walk go routine is 'parked' in
-// the pool.  Currently, we need to create objects greater than maxObjectList
-// (== 1000) which would increase time to run the test. If (and when) we decide
-// to make maxObjectList configurable we can re-evaluate adding a unit test for
```
harshavardhana, line 271: @krishnasrinivas has updated tests @krisis can you review?

harshavardhana:
Also can @krishnasrinivas or @krisis please paste here the code coverage after this change here?.  Will be automating this soon. 

Current code coverage is as shown below, would love to see if these values changed in anyway.

```
tree-walk-fs.go: 94.9% 
tree-walk-pool.go: 97.1%
tree-walk-xl.go: 94.4%
```

tree-walk.go
```diff
@@ -28,51 +28,60 @@ type treeWalkResult struct {
...
-		}
-		// Skip the entries which do not match the filter.
-		for i, entry := range entries {
-			if !filter(entry) {
-				entries[i] = ""
-				continue
+			entries, err = disk.ListDir(bucket, prefixDir)
+			if err != nil {
+				// For any reason disk was deleted or goes offline, continue
+				// and list form other disks if possible.
```
krisis: typo: s/list form/list from/

tree-walk_test.go
```diff
@@ -17,76 +17,48 @@
...
-				return !strings.HasSuffix(object, slashSeparator)
-			}, endWalkCh)
-	case xlObjects:
-		twResultCh = typ.startTreeWalk(bucket, prefix, marker, true,
-			typ.isObject, endWalkCh)
-	}
-	return twResultCh
+// Sample entries for the namespace.
+var volume = "testvolume"
+var files = []string{
```
krisis, line 29: nit: we should keep our jargon consistent. We mean objects but we refer them as files in variable names This is confusing.
e.g `var files []string`

krisis:
The changes look good to me. I have one stylistic objection. This shouldn't stop the patch from being merged.

What  this PR does could be achieved using an interface that groups ObjectLayer functions and isLeaf resulting in an (golang) idiomatic method dispatch.

In this PR, functions are used as carriers of specialized implementations and a factory function constructs one based on an input function. In effect, the function to be applied inside tree walk functions are picked by the developer, at compile-time.

If we used interfaces to reuse common code, we would pass an interface that implements ObjectLayer functions and isLeaf specifics. The appropriate implementation will be dispatched by the runtime, with the usual type guarantees that come with interfaces. In this approach, the specific implementations remain close to other aspects of a specific layer in the source code.

krisis:
The coverage results with this PR are 

```
tree-walk.go - 95.8%
tree-walk-pool.go: 97.1%
```

tree-walk_test.go
krishnasrinivas, line 29: here's the convention:
when using StorageAPI use volume and files
when working in the ObjectLayer context xl-v1-\* fs-v1-\* use bucket and object.

krisis, line 29: OK.

krishnasrinivas:
is this what you mean?:

```
type ObjectLayerPlus interface {
    ObjectLayer
    isLeaf
}

func listDir(olPlus ObjectLayerPlus, bucket, prefixDir, prefixEntry string, disks ...StorageAPI) (entries []string, err error) {                                                
        // use oPlus.isLeaf                                                                                                                                                     
}
```

> What this PR does could be achieved using an interface that groups ObjectLayer functions and isLeaf resulting in an (golang) idiomatic method dispatch.

this is not possible because isLeaf has to cater for multipart listing too, i.e isLeaf implementation checks for xl.json for regular object listing and uploads.json for multipart listing. Even if it was possible, introducing an interface type to solve this will be less desirable to using plain functions as we don't need to introduce another type for just this purpose - using higher order functions are perfectly fine in Go.

> In this PR, functions are used as carriers of specialized implementations and a factory function constructs one based on an input function. In effect, the function to be applied inside tree walk functions are picked by the developer, at compile-time.

This would a common pattern right? You can abstract implementation details in a function (isLeaf) and pass it to build higher order functions(listDirFactory) even if it is at a compile time. Infact if something can be decided at the compile time it should be decided at the compile time instead of deferring till the runtime.

> If we used interfaces to reuse common code, we would pass an interface that implements ObjectLayer functions and isLeaf specifics. The appropriate implementation will be dispatched by the runtime, with the usual type guarantees that come with interfaces.

Explained above on why function is better than interface.

> In this approach, the specific implementations remain close to other aspects of a specific layer in the source code.

you feel that implementation details aren't abstracted from one another? i am unable to see the mixup.

krisis:
> this is not possible because isLeaf has to cater for multipart listing too, i.e isLeaf implementation checks for xl.json for regular object listing and uploads.json for multipart listing. Even if it was possible, introducing an interface type to solve this will be less desirable to using plain functions as we don't need to introduce another type for just this purpose - using higher order functions are perfectly fine in Go.

I realise now that creating new interface (carrier) types for each value of isLeaf, for multipart uploads namespace and object namespace, is cumbersome. This aspect makes using interfaces less desirable for abstracting tree walk. Thanks for the explanation.

PR #2029, opened 2016-06-28
Benchmarks for ObjectLayer.PutObject()

hackintoshrao:
Fixes https://github.com/minio/minio/issues/1999

### DIFF
--- a/benchmark-utils_test.go
+++ b/benchmark-utils_test.go
@@ -26,6 +26,69 @@ import (
 	"time"
 )
 
+// Benchmark utility functions for ObjectLayer.PutObject().
+// Creates Object layer setup ( MakeBucket ) and then runs the PutObject benchmark.
+func runPutObjectBenchmark(b *testing.B, obj ObjectLayer, objSize int) {
+	var err error
+	// obtains random bucket name.
+	bucket := getRandomBucketName()
+	// create bucket.
+	err = obj.MakeBucket(bucket)
+	if err != nil {
+		b.Fatal(err)
+	}
+
+	// PutObject returns md5Sum of the object inserted.
+	// md5Sum variable is assigned with that value.
+	var md5Sum string
+	// get text data generated for number of bytes equal to object size.
+	textData := generateBytesData(objSize)
+	// generate md5sum for the generated data.
+	// md5sum of the data to written is required as input for PutObject.
+	hasher := md5.New()
+	hasher.Write([]byte(textData))
+	metadata := make(map[string]string)
+	metadata["md5Sum"] = hex.EncodeToString(hasher.Sum(nil))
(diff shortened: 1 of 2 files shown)


benchmark-utils_test.go
```diff
@@ -26,6 +26,68 @@ import (
...
+	}
+	// cleaning up the backend by removing all the directories and files created on function return.
+	defer removeRoots(disks)
+	// calling runPutObjectBenchmark which uses *testing.B and the object Layer to run the benchmark.
+	runBenchMark(b, objLayer)
+}
+
+// closure for returning the put object benchmark executor for given object size in bytes.
+func returnPutObjectBenchmark(objSize int) func(*testing.B, ObjectLayer) {
+	return func(b *testing.B, obj ObjectLayer) {
```
harshavardhana, line 86: why this closure? what purpose does this serve?

benchmark-utils_test.go
```diff
@@ -75,6 +137,8 @@ func runGetObjectBenchmark(b *testing.B, obj ObjectLayer, objSize int) {
 	b.StopTimer()
 
 }
+
+// picks a random byte and repeats it to size bytes.
 func generateBytesData(size int) []byte {
 	// randomly picks a character and returns its equivalent byte array.
 	getRandomByte := func() []byte {
```
harshavardhana: isn't getRandomByte() useful outside of this function? - why is it kept like lambda?

harshavardhana: > isn't getRandomByte() useful outside of this function? - why is it kept like lambda?

If it isn't used outside why keep it like lambda since the end result is repeat of just one byte. We should avoid unnecessary closure type functions they don't add to the readability part. 

Additionally you could use randomString(1) provides a better distribution overall.

benchmark-utils_test.go
hackintoshrao, line 86: With the closure the usage of the API to create PutObject benchmark for various sizes becomes easier, like this https://github.com/minio/minio/pull/2029/commits/dc908684308825d8d544f478b12935b338a5f97e#diff-753f78fe05fe3698678fc421b0ceb521R124 . When different functionalities are separated out the closure helps to make API more extensible.

benchmark-utils_test.go
hackintoshrao: > Additionally you could use randomString(1) provides a better distribution overall. 

Where is this function located ?

benchmark-utils_test.go
harshavardhana, line 86: > With the closure the usage of the API to create PutObject benchmark for various sizes becomes easier, like this dc90868#diff-753f78fe05fe3698678fc421b0ceb521R124 . When different functionalities are separated out the closure helps to make API more extensible.

For that purpose you should do interface. Closure is for some other purpose they are not to provide signature reflection.

PR #2033, opened 2016-06-29
posix: remove disk free space check for read-only and delete methods.

balamurugana:


posix.go
```diff
@@ -111,16 +111,23 @@ func newPosix(diskPath string) (StorageAPI, error) {
 	return fs, nil
 }
 
+// getDiskInfo returns given disk information.
```
harshavardhana, line 118: Let's add a test case for this.. @balamurugana

balamurugana, line 118: Done.

PR #2039, opened 2016-06-30
getCertsPath should use getConfigPath instead of defaulting to users homedir.

### LINKED ISSUE #2028: Different certificates directory for differents '-config-dir' are not working.
## Expected behaviour

I set -config-dir=${APP_CONFIG_DIR}, then I suppose that ${APP_CONFIG_DIR}/certs should be the certificates directory of Minio.
## Actual behaviour

But It Always point to ${HOME_DIR}/.minio/certs

Set your certificates in ~/.minio/certs (private.key and public.crt)
Put two minio servers working at same time passing a different config-dir to one of them.

Both of them will use the certificates in ~/.minio/certs
## System information

Windows 10 64-bit. Go 1.6.2

dnp1:
Considering that "-certs-path' is not a flag, I assume that it is always a subpath from '-config-dir', so I changed getCertsPath to call getConfigPath instead of homeDir directly.

### DIFF
--- a/globals.go
+++ b/globals.go
@@ -33,6 +33,7 @@ const (
 	globalMinioConfigFile    = "config.json"
 	globalMinioProfilePath   = "profile"
 	// Add new global values here.
+	globalMinioCertsDirname = "certs"
 )
 
 var (
--- a/globals.go
+++ b/globals.go
@@ -27,13 +27,12 @@ const (
 const (
 	globalMinioConfigVersion = "4"
 	globalMinioConfigDir     = ".minio"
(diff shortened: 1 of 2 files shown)


globals.go
```diff
@@ -33,6 +33,7 @@ const (
 	globalMinioConfigFile    = "config.json"
 	globalMinioProfilePath   = "profile"
 	// Add new global values here.
+	globalMinioCertsDirname = "certs"
```
harshavardhana: You could use 'globalMinioCertsDir' which is already defined.

harshavardhana:
LGTM, Thanks :+1:

PR #2048, opened 2016-07-01
posix: Return errDiskNotWritable during disk initialization.

### LINKED ISSUE #2035: server returns "volume not found" when user doesn't have write permissions on the disk(s)
## Expected behaviour

When minio server doesn't have permission on disks it is run with it should return "Permission denied" as opposed to "volume not found".
## Actual behaviour

When minio server is run for the first time on disks, the user it is run as, doesn't have permissions for, it returns "volume not found" error. 
- Create enough disks for using with minio as root (say).
- Run minio server with those disks as a different user.
## Minio version
- commit id - 57bc08c
## System information

`Linux 4.4.0-24-generic #43-Ubuntu SMP`

harshavardhana:
It can happen that minio server might not have
writable permissions on the input export paths
on command line.

Fixes #2035

### DIFF
--- a/fs-v1.go
+++ b/fs-v1.go
@@ -85,8 +85,21 @@ func newFSObjects(disk string) (ObjectLayer, error) {
 		return nil, err
 	}
 
+	// Attempt to create `.minio`.
+	err = storage.MakeVol(minioMetaBucket)
+	if err != nil {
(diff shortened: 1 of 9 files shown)


abperiasamy:
looks good now.

Issue #1990, opened 2016-06-26
tests: Address poor coverage for bucket policy tests.


harshavardhana:
bucket-handlers.go: 29% 
bucket-policy-handlers.go: 39%

hackintoshrao:
Timeline: 2nd July.

PR #2055, opened 2016-07-01
Bucket Policy: Fixes and tests. 

hackintoshrao:
With changes in code sending this as a separate commit. 
- Addition to allowed actions.
- Changes to ResourceMatch logic.
- Test for action match function.

### DIFF
--- a/bucket-policy-parser.go
+++ b/bucket-policy-parser.go
@@ -34,6 +34,8 @@ const (
 
 // supportedActionMap - lists all the actions supported by minio.
 var supportedActionMap = map[string]struct{}{
+	"*":                             {},
+	"s3:*":                          {},
 	"s3:GetObject":                  {},
 	"s3:ListBucket":                 {},
 	"s3:PutObject":                  {},
--- a/bucket-policy-handlers.go
+++ b/bucket-policy-handlers.go
@@ -21,7 +21,6 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
-	"regexp"
 	"strings"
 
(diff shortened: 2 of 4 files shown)


bucket-policy-parser.go
```diff
@@ -34,6 +34,8 @@ const (
 
 // supportedActionMap - lists all the actions supported by minio.
 var supportedActionMap = map[string]struct{}{
+	"*":                             {},
+	"s3*":                           {},
```
harshavardhana: This should be "s3:*" , can you check?

harshavardhana:
Can you also paste new coverage following these changes? - will see about automating this for minio server as well.

hackintoshrao:
> Can you also paste new coverage following these changes?

The coverage is still bad since the GET,PUT and DELETE bucket handlers needs tests. Working on it. 

> will see about automating this for minio server as well.

What do you mean by this ? I'll be adding some functional tests as well with these changes.

bucket-policy-parser.go
hackintoshrao: Changing it to `s3:*`

harshavardhana:
> What do you mean by this ? I'll be adding some functional tests as well with these changes.I

Will add support for running coverage all the time, so that developers get to see if their patch decreases the coverage such patches will not be accepted.  (there are some exceptions though)

hackintoshrao:
@harshavardhana : As you pointed out there are exceptions.

hackintoshrao:
Part of fix for https://github.com/minio/minio/issues/1990

PR #2071, opened 2016-07-02
api/bucket-policy:  Refactor, Handler test.

hackintoshrao:
PR contains,
- New setup utilities for running object handler tests. Here is why they are essential, 
  - Unit tests have to be run in isolation without being have to be associated with other functionalities which are not under test. 
  - The integration tests follows the philosophy of running a Test Server and registers all handlers and fires HTTP requests over the socket to simulate the system functionality under usual work load scenarios and test for correctness. But this philosophy cannot be adopted for running unit tests for HTTP handlers. 
  - Running Unit tests for API handlers, 
    - Shouldn't run a test server. Should purely call the handlers `ServeHTTP` under isolation. 
    - Shouldn't register all handlers, should only register handlers under test and so that the system is close to be in an isolated setup. 
- As an example PutBucketPolicy test is illustrated using the new setup. Exhaustive cases has to be added and has been listen in TODO for now.

### DIFF
--- a/test-utils_test.go
+++ b/test-utils_test.go
@@ -116,7 +116,7 @@ func StartTestServer(t TestErrHandler, instanceType string) TestServer {
 
 	credentials, root, err := initTestConfig("us-east-1")
 	if err != nil {
-		t.Fatalf(err.Error())
+		t.Fatalf("%s", err)
 	}
 	testServer.Root = root
 	testServer.Disks = erasureDisks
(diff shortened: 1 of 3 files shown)


hackintoshrao:
Part of fix for https://github.com/minio/minio/issues/1990 . Since there are changes to the testing setup itself, this needs a separate commit. Will be sending tests for 2 other bucket handlers in the next PR.

test-utils_test.go
```diff
@@ -111,34 +113,41 @@ func StartTestServer(t TestErrHandler, instanceType string) TestServer {
...
 		t.Fatalf("Failed obtaining Temp Backend: <ERROR> %s", err)
 	}
-	testServer.Disks = erasureDisks
-	// Obtain temp root.
-	root, err := getTestRoot()
+
+	credentials, root, err := initTestConfig("us-east-1")
 	if err != nil {
-		t.Fatalf("Failed obtaining Temp Root for the backend Backend: <ERROR> %s", err)
+		t.Fatalf(err.Error())
```
harshavardhana: why still 'err.Error()' ?

PR #2083, opened 2016-07-04
FS: check whether disk format is FS or not.

### LINKED ISSUE #2060: Simple command error could easily lead to a corrupted backend format msg
It is easy to corrupt the disks (from the user's point of view):
1.  $ mkdir -p /tmp/fs1/export{1..16}
2.  $ minio server /tmp/fs1/export{1..16}
3. In a new terminal: 
$ mc mb myminio/bucket/ && mc cp file myminio/bucket/
4. Now, you run mistakenly this command and you correct it later 
$ minio server /tmp/fs1/export1
CTRL-C 
$ minio server /tmp/fs1/export{1..16}

FATA[0000] Unable to intialize object layer.             cause=corrupted backend format sysInfo=map[host.arch:amd64 host.name:xps-u host.cpus:4 mem.used:3.2MB mem.total:6.5MB mem.heap.used:3.2MB mem.heap.total:4.7MB host.os:linux host.lang:go1.6.2] type=*errors.errorString

balamurugana:
Fixes #2060

### DIFF
--- a/fs-v1-errors.go
+++ b/fs-v1-errors.go
@@ -1,0 +1,22 @@
+/*
+ * Minio Cloud Storage, (C) 2016 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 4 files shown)


harshavardhana:
Can we add a test case here @balamurugana ?

balamurugana:
> Can we add a test case here @balamurugana ?

Done

PR #2089, opened 2016-07-04
server-tests: unify XL and FS tests into common code. server_test.go contains common test code. server_xl_test.go contains XL tests specific to XL.

krishnasrinivas:


server_test.go
```diff
@@ -33,32 +33,36 @@ import (
...
 	secretKey  string
 }
 
-// Initializing the test suite.
-var _ = Suite(&TestSuiteFS{})
+// Init and run test on FS backend.
+var _ = Suite(&TestSuiteCommon{serverType: "FS"})
+
+// Init and run test on XL backend.
+var _ = Suite(&TestSuiteCommon{serverType: "XL"})
```
hackintoshrao, line 49: This guarantees running of the tests twice right ? one for "XL" and one for "FS" ? Just checking, haven't used it like that before.

hackintoshrao:
This is a  much needed change, cleaner/simple/extensible. Have just one comment on the code, if thats good then LGTM 👍 . Thanks !

server_test.go
harshavardhana, line 49: > This guarantees running of the tests twice right ? one for "XL" and one for "FS" ? Just checking, haven't used it like that before.

One has to validate how many tests were run using 'go test'

krishnasrinivas, line 49: > This guarantees running of the tests twice right ? one for "XL" and one for "FS" ? Just checking, haven't used it like that before.

Verified once again, yes works as expected.

PR #2090, opened 2016-07-04
unit-tests: Unit tests for functions in erasure-readfile.go

krishnasrinivas:
@harshavardhana there is no easy way to unit test erasureReadFile and parallelRead.

The only way is to do xl.PutObject and use the resulting environment to test erasureReadFile and parallelRead, no other way right?

### DIFF
--- a/erasure-readfile_test.go
+++ b/erasure-readfile_test.go
@@ -1,0 +1,206 @@
+/*
+ * Minio Cloud Storage, (C) 2015, 2016 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package main
+
+import "testing"
+import "reflect"
+
+// Tests getReadDisks which returns readable disks slice from which we can
+// read parallelly.
+func testGetReadDisks(t *testing.T, xl xlObjects) {
+	d := xl.storageDisks
+	testCases := []struct {
+		index     int          // index argument for getReadDisks
+		argDisks  []StorageAPI // disks argument for getReadDisks
+		retDisks  []StorageAPI // disks return value from getReadDisks
+		nextIndex int          // return value from getReadDisks
+		err       error        // error return value from getReadDisks
+	}{
+		{
+			0,
+			// When all disks are available, should return data disks.
+			[]StorageAPI{d[0], d[1], d[2], d[3], d[4], d[5], d[6], d[7], d[8], d[9], d[10], d[11], d[12], d[13], d[14], d[15]},
+			[]StorageAPI{d[0], d[1], d[2], d[3], d[4], d[5], d[6], d[7], nil, nil, nil, nil, nil, nil, nil, nil},
+			8,
+			nil,
+		},
+		{
+			0,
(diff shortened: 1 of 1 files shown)


erasure-readfile_test.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+	disks[0] = nil
+	// With one data disk down, it should return dataBlocks-1 data disks and 2 parity disks.
+	readDisks, _, err = getReadDisks(disks, 0, xl.dataBlocks)
+	if err != nil {
+		t.Errorf("getReadDisks returned error: %s", err)
+		return
+	}
+
+	for index := 0; index < len(readDisks); index++ {
```
harshavardhana: This whole block needs comments..

erasure-readfile_test.go
```diff
@@ -0,0 +1,148 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package main
+
+import "testing"
+
+// Test for getReadDisks.
```
harshavardhana: Comment should be about 'non-obvious' - should be about the gist of the function body.

erasure-readfile_test.go
```diff
@@ -0,0 +1,148 @@
...
+			} else {
+				if readDisks[index] != nil {
+					t.Errorf("readDisks[%d] should be nil", index)
+				}
+			}
+		}
+	}
+}
+
+// Test getOrderedDisks.
```
harshavardhana: Same here.

erasure-readfile_test.go
```diff
@@ -0,0 +1,148 @@
...
+func testGetOrderedDisks(t *testing.T, xl xlObjects) {
+	disks := xl.storageDisks
+	blockCheckSums := make([]checkSumInfo, len(disks))
+	distribution := []int{16, 14, 12, 10, 8, 6, 4, 2, 1, 3, 5, 7, 9, 11, 13, 15}
+	orderedDisks, _ := getOrderedDisks(distribution, disks, blockCheckSums)
+	if orderedDisks[0] != disks[8] ||
```
harshavardhana, line 118: Comment here.

erasure-readfile_test.go
```diff
@@ -0,0 +1,148 @@
...
+	enBlocks[dataBlocks-1] = nil
+	// When only parity blocks are available.
+	if isSuccessDecodeBlocks(enBlocks, dataBlocks) == true {
+		t.Errorf("isSuccessDecodeBlocks should return false.")
+	}
+}
+
+// Wrapper function for testGetReadDisks, testGetOrderedDisks.
+func TestErasureReadFile(t *testing.T) {
+	objLayer, dirs, err := getXLObjectLayer()
```
hackintoshrao, line 198: can you use `makeTestBackend` function from `test-utils_test.go`, its a wrapper and used around in the tests/benchmarks.

hackintoshrao:
- It'll be good if unit tests follow the style of formulating inputs and expected output in an array of structs, the cases should contain erroneous inputs and the errors should be verified, that'll make the tests easily extensibe, it'll be more encouraging to add more test cases in the future.
- The exhaustive of the test can be verified by deleting any `if err!=nil` conditional check of errors and see whether the test catches them. Basically test should also work as code preservatives, any undesirable code changes should be caught by the tests ( thats the challenge I'm witnessing w.r.t to the current implementation of locks, not able to write test to catch the removal or addition of any lock related changes in the code).

harshavardhana:
> @harshavardhana there is no easy way to unit test erasureReadFile and parallelRead.
> 
> The only way is to do xl.PutObject and use the resulting environment to test erasureReadFile and parallelRead, no other way right?

This is acceptable, once we have an object through PutObject.
-  We can corrupt them by appending to the file directly using posix.  
-  We can remove the disks entirely.
-  We can remove all but one data disks to hit the parallel read special case. 

All these cases can be simulated and we should do it otherwise we won't be able to catch issues.

PR #2095, opened 2016-07-05
fix: allocate buffer to required size than readSizeV1

balamurugana:


erasure-utils.go
```diff
@@ -152,13 +152,17 @@ func getEncodedBlockLen(inputLen int64, dataBlocks int) (curEncBlockSize int64)
 // err == nil, not err == EOF. Additionally offset can be provided to start
 // the read at. copyN returns io.EOF if there aren't enough data to be read.
 func copyN(writer io.Writer, disk StorageAPI, volume string, path string, offset int64, length int64) (err error) {
-	// Use 128KiB staging buffer to read up to length.
-	buf := make([]byte, readSizeV1)
+	// Use staging buffer to read up to length.
+	bufSize := readSizeV1
```
harshavardhana: Better use 'int64(readSizeV1)' so that you don't have to do 'int()' on length.

erasure-utils.go
```diff
@@ -152,13 +152,17 @@ func getEncodedBlockLen(inputLen int64, dataBlocks int) (curEncBlockSize int64)
...
+	if int64(bufSize) > length {
+		bufSize = int(length)
+	}
+	buf := make([]byte, bufSize)
 
 	// Read into writer until length.
 	for length > 0 {
-		curLength := int64(readSizeV1)
-		if length < readSizeV1 {
+		curLength := int64(bufSize)
```
harshavardhana: This is not needed if bufSize is int64.

harshavardhana:
Can you run 'go test' locally tests are hanging.

erasure-utils.go
balamurugana: Done

balamurugana:
The test object-api-multipart_test.go::TestObjectAPIPutObjectPart hangs on the case

``` go
        // Test case - 12.
        // Input to replicate Md5 mismatch.
        {bucket, object, uploadID, 1, "", "a35", 0, false, "",
            fmt.Errorf("%s", "Bad digest: Expected a35 is not valid with what we calculated "+"d41d8cd98f00b204e9800998ecf8427e")},

```

balamurugana:
https://github.com/minio/minio/blob/master/fs-v1-multipart.go#L299 does not break as https://github.com/minio/minio/blob/master/fs-v1-multipart.go#L300 returns `0, nil`.  This causes blocking infinite loop.

PR #2108, opened 2016-07-05
object-cache: use golang bytes.Buffer and bytes.NewReader instead of custom implementation.

### LINKED ISSUE #2107: XL/objectCache: two parallel GetObject can result in partial data
If there are two parallel GetObjects:
1) one GetObject will be filling the cache, wouldn't have completed.
2) second GetObject will find the entry in the cache and return partial data.

@harshavardhana i have the fix, pusing it.

krishnasrinivas:
Also fixes #2107


pkg/objcache/objcache.go
```diff
@@ -112,9 +100,32 @@ func (c *Cache) Create(key string, size int64) (writer io.Writer, err error) {
...
-	c.entries[key] = NewBuffer(make([]byte, 0, int(size)))
-	c.currentSize += valueLen
-	return c.entries[key], nil
+
+	// Will hold the object contents.
+	buf := bytes.NewBuffer(make([]byte, 0, size))
+	// Account for the memory allocated above.
+	c.currentSize += uint64(size)
+	// Implements io.WriteCloser
+	return struct {
```
harshavardhana: Write this more cleanly do not use an anonymous struct, implement a proper struct and fill it with io.Writer and Closer.

pkg/objcache/objcache.go
```diff
@@ -112,9 +100,32 @@ func (c *Cache) Create(key string, size int64) (writer io.Writer, err error) {
...
+		addToCache(func() error {
+			c.mutex.Lock()
+			defer c.mutex.Unlock()
+			if buf.Len() != int(size) {
+				// Full object not available hence do not save buf to object cache.
+				c.currentSize -= uint64(size)
+				return nil
+			}
+			// Full object available in buf, save it to cache.
+			c.entries[key] = buf.Bytes()
```
harshavardhana: Did you test the overall memory usage of using this style v/s having our own buffer implementation?

krishnasrinivas:
> This causes a copy, the reason to have our own buffer is to be able to avoid having an extra copy of the same buffer.

There is no copy as it's a slice.

pkg/objcache/objcache.go
krishnasrinivas: pushed a commit with better readability.

pkg/objcache/objcache.go
```diff
@@ -74,11 +74,15 @@ var ErrKeyNotFoundInCache = errors.New("Key not found in cache")
...
-type addToCache func() error
+// Used for adding entry to the object cache. Implements io.WriteCloser
+type cacheBuffer struct {
+	*bytes.Buffer // Implements io.Writer
+	onClose       func()
+}
 
-func (add addToCache) Close() error {
-	return add()
+func (c cacheBuffer) Close() error {
```
harshavardhana, line 85: Close requires comment.. and an explanation of what it does.

pkg/objcache/objcache.go
```diff
@@ -105,26 +109,21 @@ func (c *Cache) Create(key string, size int64) (w io.WriteCloser, err error) {
...
-		io.Writer
-		io.Closer
-	}{
-		// Implements io.Writer
+
+	return cacheBuffer{
 		buf,
-		// Implements io.Closer
-		addToCache(func() error {
+		func() {
```
harshavardhana: Needs a comment.

pkg/objcache/objcache.go
```diff
@@ -74,11 +74,15 @@ var ErrKeyNotFoundInCache = errors.New("Key not found in cache")
 // ErrCacheFull - cache is full.
 var ErrCacheFull = errors.New("Not enough space in cache")
 
-// Used for adding entry to the object cache. Implements io.Closer
-type addToCache func() error
+// Used for adding entry to the object cache. Implements io.WriteCloser
+type cacheBuffer struct {
+	*bytes.Buffer // Implements io.Writer
+	onClose       func()
```
harshavardhana, line 80: why are we not returning error anymore?

pkg/objcache/objcache.go
```diff
@@ -105,26 +109,21 @@ func (c *Cache) Create(key string, size int64) (w io.WriteCloser, err error) {
 	buf := bytes.NewBuffer(make([]byte, 0, size))
 	// Account for the memory allocated above.
 	c.currentSize += uint64(size)
-	// Implements io.WriteCloser
-	return struct {
-		io.Writer
-		io.Closer
-	}{
-		// Implements io.Writer
+
```
harshavardhana, line 114: Needs a comment.

harshavardhana:
> why are we not returning error anymore?

I see its not needed, fair enough to not return any.

PR #2112, opened 2016-07-06
http: handle request range like Amazon S3.

### LINKED ISSUE #2098: api: getObjectHandler returns InvalidRange for offsets with in the length of data.
## Expected behaviour

Should not return an error for offset value panning with in the length of the data.
## Actual behaviour

We return InvalidRange error 
## Steps to reproduce the behavior

Following code runs fine on S3 but Minio server returns InvalidRange.

```
func (s *TestSuiteCommon) TestGetObjectRange(c *C) {
        // generate a random bucket name.
        bucketName := getRandomBucketName()
        // HTTP request to create the bucket.
        request, err := newTestRequest("PUT", getMakeBucketURL(s.endPoint, bucketName),
                0, nil, s.accessKey, s.secretKey)
        c.Assert(err, IsNil)

        client := http.Client{}
        // execute the HTTP request to create bucket.
        response, err := client.Do(request)
        c.Assert(err, IsNil)
        c.Assert(response.StatusCode, Equals, http.StatusOK)

        // content for the object to be uploaded.
        buffer1 := bytes.NewReader([]byte("Hello World"))

        objectName := "test-object"
        // HTTP request to upload the object.
        request, err = newTestRequest("PUT", getPutObjectURL(s.endPoint, bucketName, objectName),
                int64(buffer1.Len()), buffer1, s.accessKey, s.secretKey)
        c.Assert(err, IsNil)

        client = http.Client{}
        // execute the HTTP request to upload the object.
        response, err = client.Do(request)
        c.Assert(err, IsNil)
        // verify whether upload was successful.
        c.Assert(response.StatusCode, Equals, http.StatusOK)

        // HTTP request to download the object.
        request, err = newTestRequest("GET", getGetObjectURL(s.endPoint, bucketName, objectName),
                0, nil, s.accessKey, s.secretKey)
        // Invalid byte range set.
        request.Header.Add("Range", "bytes=7-6")
        c.Assert(err, IsNil)

        client = http.Client{}
        // execute the HTTP request.
        response, err = client.Do(request)
        c.Assert(err, IsNil)
        c.Assert(response.StatusCode, Equals, http.StatusOK)
}
```
## Minio version
  0540863663b9f515a0a5bd803af34adca08468dc
## System information

Ubuntu 16.04

balamurugana:
Fixes #2098

### DIFF
--- a/httprange.go
+++ b/httprange.go
@@ -1,5 +1,5 @@
 /*
- * Minio Cloud Storage, (C) 2015 Minio, Inc.
+ * Minio Cloud Storage, (C) 2015, 2016 Minio, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -24,10 +24,13 @@ import (
 )
 
 const (
-	b = "bytes="
+	byteRangePrefix = "bytes="
 )
 
-// InvalidRange - invalid range
+// errInvalidRange - returned when given range value is not valid.
+var errInvalidRange = errors.New("Invalid range")
+
+// InvalidRange - invalid range typed error.
 type InvalidRange struct{}
 
 func (e InvalidRange) Error() string {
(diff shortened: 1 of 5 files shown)


httprange.go
```diff
@@ -36,93 +39,77 @@ func (e InvalidRange) Error() string {
...
 	}
-	return nil
-}
 
-// parseRange parses a Range header string as per RFC 2616.
-func (r *httpRange) parseRange(s string) error {
-	if s == "" {
-		return errors.New("header not present")
+	// Trim byte range prefix.
+	byteRangeString := rangeString[len(byteRangePrefix):]
```
krishnasrinivas: minor nitpick, you can use strings.TrimPrefix

httprange.go
```diff
@@ -36,93 +39,77 @@ func (e InvalidRange) Error() string {
...
+
+	// Check if range string contains delimiter '-', else return error.
+	sepIndex := strings.Index(byteRangeString, "-")
```
krishnasrinivas, line 67: this is fine, but you could use helper functions in strings package:

if !stirngs.Contains(byteRangeString, "-") {return...}

posStrs := strings.Split(byteRangeString, "-")

httprange.go
```diff
@@ -36,93 +39,77 @@ func (e InvalidRange) Error() string {
...
-		return InvalidRange{}
+
+	firstBytePosString := byteRangeString[:sepIndex]
+	lastBytePosString := byteRangeString[sepIndex+1:]
+
+	firstBytePos := int64(-1)
+	lastBytePos := int64(-1)
+
+	// Convert firstBytePosString only if its not empty.
+	if len(firstBytePosString) > 0 {
```
krishnasrinivas, line 79: this is also fine, another way:
if firstBytePosString != "" {...}

httprange.go
```diff
@@ -36,93 +39,77 @@ func (e InvalidRange) Error() string {
...
+		if firstBytePos, err = strconv.ParseInt(firstBytePosString, 10, 64); err != nil {
```
krishnasrinivas, line 80: strconv.Atoi is an easier API as it works on base 10 automatically

httprange.go
```diff
@@ -36,93 +39,77 @@ func (e InvalidRange) Error() string {
 
 // HttpRange specifies the byte range to be sent to the client.
 type httpRange struct {
```
krishnasrinivas, line 41: i think `offset, length, size` would be more native instead of positions

httprange.go
```diff
@@ -36,93 +39,77 @@ func (e InvalidRange) Error() string {
...
-			i, err := strconv.ParseInt(end, 10, 64)
-			if err != nil || r.start > i {
-				return InvalidRange{}
-			}
-			if i >= r.size {
-				i = r.size - 1
-			}
-			r.length = i - r.start + 1
-		}
+func parseRequestRange(rangeString string, size int64) (hrange *httpRange, err error) {
```
krishnasrinivas, line 57: need not return a pointer.

httprange.go
```diff
@@ -36,93 +39,77 @@ func (e InvalidRange) Error() string {
 
 // HttpRange specifies the byte range to be sent to the client.
 type httpRange struct {
-	start, length, size int64
+	firstBytePos int64
+	lastBytePos  int64
+	size         int64
 }
 
 // String populate range stringer interface
```
krishnasrinivas, line 47: need not be pointer receiver methods here.

httprange.go
balamurugana: Done

httprange.go
balamurugana, line 67: We would not need to split by multiple "-" as strings.Split() does.  If `rangeString  = "bytes=2-5,8-10"` we would need to do array math to check if its invalid or not.

You can refer httprange_test.go for valid cases.

httprange.go
balamurugana, line 80: `strconv.Atoi()` returns `int`, but we need to work with `int64`

httprange.go
balamurugana, line 41: for HTTP range, as per [rfc7233](https://tools.ietf.org/html/rfc7233) it calls first/last byte position which is more appropriate I think.

httprange.go
balamurugana, line 57: This is on purpose to avoid default field value of httprange structure.  0 is valid first/last byte position.

httprange.go
balamurugana, line 47: Done

httprange.go
krishnasrinivas, line 41: BytePos is for the protocol.
By native I meant, for us, once we do parseRequestRange() to get httpRange we never use lastBytePos and we have to call getLength() to get the length. The term byte position will not hold meaning for us, it will be offset and length.

httprange.go
krishnasrinivas, line 57: > This is on purpose to avoid default field value of httprange structure. 0 is valid first/last byte position.

Did not get this, how does returning pointer help?

httprange.go
```diff
@@ -36,93 +39,77 @@ func (e InvalidRange) Error() string {
...
+	// Check if range string contains delimiter '-', else return error.
+	sepIndex := strings.Index(byteRangeString, "-")
+	if sepIndex == -1 {
+		return nil, fmt.Errorf("invalid range string '%s'", rangeString)
 	}
-	if !strings.HasPrefix(s, b) {
-		return InvalidRange{}
+
+	firstBytePosString := byteRangeString[:sepIndex]
+	lastBytePosString := byteRangeString[sepIndex+1:]
```
krishnasrinivas, line 73: you can do:

if !strings.Contains(byteRangeString, "-") {return err}

strs := strings.Split("-")
if len(strs) != 2 {return err}

object-handlers.go
```diff
@@ -104,12 +104,15 @@ func (api objectAPIHandlers) GetObjectHandler(w http.ResponseWriter, r *http.Req
...
-	// Caculate the http Range.
+	// Get request range.
 	var hrange *httpRange
-	hrange, err = getRequestedRange(r.Header.Get("Range"), objInfo.Size)
-	if err != nil {
-		writeErrorResponse(w, r, ErrInvalidRange, r.URL.Path)
-		return
+	if hrange, err = parseRequestRange(r.Header.Get("Range"), objInfo.Size); err != nil {
+		// Handle only errInvalidRange
+		// Ignore other parse error and treat it as regular Get request like Amazon S3.
```
harshavardhana, line 111: We can log the errors which are not handled here.

httprange_test.go
```diff
@@ -0,0 +1,67 @@
...
+		if hrange.lastBytePos != successCase.lastBytePos {
+			t.Fatalf("expected: %d, got: %d", successCase.lastBytePos, hrange.lastBytePos)
+		}
+		if hrange.getLength() != successCase.length {
+			t.Fatalf("expected: %d, got: %d", successCase.length, hrange.getLength())
+		}
+	}
+
+	// Test invalid range strings.
+	invalidRangeStrings := []string{"", "2-5", "bytes= 2-5", "bytes=2 - 5", "bytes=0-0,-1", "bytes=5-2"}
```
harshavardhana: add one more case like "bytes=2-5 " - notice space at the end.

harshavardhana: > add one more case like "bytes=2-5 " - notice space at the end.

One more case 'bytes=2--5' - notice double '-'

httprange.go
```diff
@@ -36,93 +39,77 @@ func (e InvalidRange) Error() string {
...
 
-// parseRange parses a Range header string as per RFC 2616.
-func (r *httpRange) parseRange(s string) error {
-	if s == "" {
-		return errors.New("header not present")
+	// Trim byte range prefix.
+	byteRangeString := strings.TrimPrefix(rangeString, byteRangePrefix)
+
+	// Check if range string contains delimiter '-', else return error.
+	sepIndex := strings.Index(byteRangeString, "-")
```
harshavardhana, line 67: strings.IndexRune() is better quite common in parsing http headers.

httprange.go
```diff
@@ -24,10 +24,13 @@ import (
...
-	b = "bytes="
+	byteRangePrefix = "bytes="
 )
 
-// InvalidRange - invalid range
+// errInvalidRange - returned when given range value is not valid.
+var errInvalidRange = errors.New("Invalid range")
+
+// InvalidRange - invalid range typed error.
 type InvalidRange struct{}
```
harshavardhana, line 34: Remove this error if its not needed anymore.

httprange_test.go
balamurugana: Done

object-handlers.go
balamurugana, line 111: Done

httprange.go
balamurugana, line 34: it is used by xl-v1-metadata.go:182

httprange.go
balamurugana, line 67: As per rfc, i don't see its unicode. Why "-" in strings.IndexRune() needs to be treated as unicode?

harshavardhana, line 67: > As per rfc, i don't see its unicode. Why "-" in strings.IndexRune() needs to be treated as unicode?

IndexRune is better understood since its a single character.

PR #2114, opened 2016-07-06
URL Encode X-Amz-Copy-Source as per the spec

ncw:
The documents for COPY state that the X-Amz-Copy-Source must be URL encoded.

http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectCOPY.html

I discovered this while experimenting with [rclone](http://github.com/ncw/rclone) and minio.

### DIFF
--- a/object-handlers.go
+++ b/object-handlers.go
@@ -341,7 +341,12 @@ func (api objectAPIHandlers) CopyObjectHandler(w http.ResponseWriter, r *http.Re
 	// don't even read it.
 
 	// objectSource
-	objectSource := r.Header.Get("X-Amz-Copy-Source")
+	objectSource, err := url.QueryUnescape(r.Header.Get("X-Amz-Copy-Source"))
+	if err != nil {
+		writeErrorResponse(w, r, ErrCouldntDecodeCopySource, r.URL.Path)
+		return
+	}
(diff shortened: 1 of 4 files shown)


object-handlers.go
```diff
@@ -341,7 +341,12 @@ func (api objectAPIHandlers) CopyObjectHandler(w http.ResponseWriter, r *http.Re
 	// don't even read it.
 
 	// objectSource
-	objectSource := r.Header.Get("X-Amz-Copy-Source")
+	objectSource, err := url.QueryUnescape(r.Header.Get("X-Amz-Copy-Source"))
+	if err != nil {
+		writeErrorResponse(w, r, ErrCouldntDecodeCopySource, r.URL.Path)
+		return
+	}
+	log.Errorf("source %q\n", objectSource)
```
harshavardhana, line 349: We shouldn't log here.

harshavardhana:
Thanks for the patch @ncw - will make necessary changes locally and merge.

object-handlers.go
ncw, line 349: Oops left that in from debugging, sorry.

harshavardhana:
Merged with some changes @ncw

ncw:
Thanks!

PR #2115, opened 2016-07-07
XL/PutObject: Add single putObject and multipart caching.

harshavardhana:
Additionally adds test cases as well for object cache.

### DIFF
--- a/pkg/objcache/objcache.go
+++ b/pkg/objcache/objcache.go
@@ -26,15 +26,25 @@ import (
 	"time"
 )
 
-// NoExpiration represents caches to be permanent and can only be deleted.
-var NoExpiration = time.Duration(0)
+// NoExpiry represents caches to be permanent and can only be deleted.
+var NoExpiry = time.Duration(0)
+
+// DefaultExpiry represents default time duration value when individual entries will be expired.
+var DefaultExpiry = time.Duration(72 * time.Hour) // 72hrs.
+
+// buffer represents the in memory cache of a single entry.
+// buffer carries value of the data and last accessed time.
+type buffer struct {
+	value        []byte    // Value of the entry.
+	lastAccessed time.Time // Represents time when value was last accessed.
+}
 
 // Cache holds the required variables to compose an in memory cache system
 // which also provides expiring key mechanism and also maxSize.
 type Cache struct {
 	// Mutex is used for handling the concurrent
 	// read/write requests for cache
-	mutex *sync.RWMutex
+	mutex sync.Mutex
 
 	// maxSize is a total size for overall cache
 	maxSize uint64
@@ -43,29 +53,39 @@ type Cache struct {
 	currentSize uint64
 
 	// OnEviction - callback function for eviction
-	OnEviction func(a ...interface{})
+	OnEviction func(key string)
 
-	// totalEvicted counter to keep track of total expirations
+	// totalEvicted counter to keep track of total expirys
 	totalEvicted int
 
 	// map of objectName and its contents
-	entries map[string][]byte
+	entries map[string]*buffer
 
-	// Expiration in time duration.
+	// Expiry in time duration.
 	expiry time.Duration
+
+	// Stop garbage collection routine, stops any running GC routine.
+	stopGC chan struct{}
 }
 
-// New creates an inmemory cache
-//
-// maxSize is used for expiring objects before we run out of memory
-// expiration is used for expiration of a key from cache
+// New - Return a new cache with a given default expiry duration.
+// If the expiry duration is less than one (or NoExpiry),
+// the items in the cache never expire (by default), and must be deleted
+// manually.
 func New(maxSize uint64, expiry time.Duration) *Cache {
-	return &Cache{
-		mutex:   &sync.RWMutex{},
+	C := &Cache{
 		maxSize: maxSize,
-		entries: make(map[string][]byte),
+		entries: make(map[string]*buffer),
 		expiry:  expiry,
 	}
+	// We have expiry start the janitor routine.
+	if expiry > 0 {
+		C.stopGC = make(chan struct{})
+
+		// Start garbage collection routine to expire objects.
+		C.startGC()
+	}
+	return C
 }
 
 // ErrKeyNotFoundInCache - key not found in cache.
@@ -74,17 +94,19 @@ var ErrKeyNotFoundInCache = errors.New("Key not found in cache")
 // ErrCacheFull - cache is full.
 var ErrCacheFull = errors.New("Not enough space in cache")
 
+// ErrExcessData - excess data was attempted to be written on cache.
+var ErrExcessData = errors.New("Attempted excess write on cache")
+
 // Used for adding entry to the object cache. Implements io.WriteCloser
 type cacheBuffer struct {
 	*bytes.Buffer // Implements io.Writer
-	onClose       func()
+	onClose       func() error
 }
 
 // On close, onClose() is called which checks if all object contents
 // have been written so that it can save the buffer to the cache.
-func (c cacheBuffer) Close() error {
-	c.onClose()
-	return nil
+func (c cacheBuffer) Close() (err error) {
+	return c.onClose()
 }
 
 // Create - validates if object size fits with in cache size limit and returns a io.WriteCloser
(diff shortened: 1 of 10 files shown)


pkg/objcache/objcache.go
```diff
@@ -77,14 +77,16 @@ var ErrCacheFull = errors.New("Not enough space in cache")
...
 }
 
 // On close, onClose() is called which checks if all object contents
 // have been written so that it can save the buffer to the cache.
-func (c cacheBuffer) Close() error {
-	c.onClose()
-	return nil
+func (c cacheBuffer) Close() (err error) {
+	err = c.onClose()
+	// Reset the buffer since we are not going to use it anymore.
```
krishnasrinivas: better not do c.Reset() because it indicates that we are going to re-use the buffer.

xl-v1-object.go
```diff
@@ -429,6 +452,7 @@ func (xl xlObjects) PutObject(bucket string, object string, size int64, data io.
 		if vErr := data.(*signVerifyReader).Verify(); vErr != nil {
 			// Incoming payload wrong, delete the temporary object.
 			xl.deleteObject(minioMetaTmpBucket, tempObj)
+
 			// Error return.
```
krishnasrinivas, line 475: before returning cache used size must be reduced `c.currentSize -= uint64(size)` should be done, by doing maybe newBuffer.CloseWIthError(err)
better not do caching for put-object for 2 reasons:
1) we can keep the cache code simple by not handling special case of sha/signature mismatch.
2) we will never cache something that will never be read.

harshavardhana, line 475: > before returning cache used size must be reduced c.currentSize -= uint64(size) should be done, by doing maybe newBuffer.CloseWIthError(err)
> better not do caching for put-object for 2 reasons:
> 1) we can keep the cache code simple by not handling special case of sha/signature mismatch.
> 2) we will never cache something that will never be read.

We are not even writing to the cache why should it be reduced?

harshavardhana:
> better not do c.Reset() because it indicates that we are going to re-use the buffer.

Makes sense

xl-v1-object.go
krishnasrinivas, line 475: in cache.create we do `c.currentSize += uint64(size)` while allocating new buffer. In that case we need not increment during cache.Create() and increment the cache size during Close() when adding to the cache. If the size exceeds cache size the []byte need not be added to the cache.

harshavardhana, line 475: > in cache.create we do c.currentSize += uint64(size) while allocating new buffer. In that case we need not increment during cache.Create() and increment the cache size during Close() when adding to the cache. If the size exceeds cache size the []byte need not be added to the cache.

yes since Close() is used now.

krishnasrinivas:
looks good

krishnasrinivas:
Added comments.

krishnasrinivas:
My comments are gone? adding the comments again.

pkg/objcache/janitor.go
```diff
@@ -0,0 +1,53 @@
...
+ *
+ */
+
+package objcache
+
+import "time"
+
+// FIXME comments.
+
+type janitor struct {
```
krishnasrinivas: i think having a janitor method on `cache struct` is more readable and understand what's happening instead of janitor being separate from cache.

pkg/objcache/objcache.go
```diff
@@ -22,23 +22,42 @@ import (
...
 // NoExpiration represents caches to be permanent and can only be deleted.
 var NoExpiration = time.Duration(0)
 
+// DefaultExpiration represents default time duration value when individual
+// entries will be expired.
+var DefaultExpiration = time.Duration(24 * time.Hour)
+
+// DefaultCleanupInterval represents default value when the go-routine kicks
+// in and cleanups all the expired entries.
+var DefaultCleanupInterval = time.Duration(25 * time.Hour)
```
krishnasrinivas: default value can be much less like 30 minutes.

pkg/objcache/objcache.go
```diff
@@ -59,20 +78,40 @@ type Cache struct {
...
+		entries:         make(map[string]buffer),
+		expiration:      expiration,
+		cleanupInterval: cleanupInterval,
+	}
+	// This trick ensures that the janitor goroutine (which--granted it
+	// was enabled--is running DeleteExpired on c forever) does not keep
+	// the returned C object from being garbage collected. When it is
+	// garbage collected, the finalizer stops the janitor goroutine, after
+	// which c can be collected.
+	C := &Cache{c}
```
krishnasrinivas: This trick is hard to read. You can simply have `func (c *Cache) Destroy()` that stops the janitor goroutine.

pkg/objcache/objcache.go
```diff
@@ -119,19 +155,19 @@ func (c *Cache) Create(key string, size int64) (w io.WriteCloser, err error) {
 	// Function called on close which saves the object contents
 	// to the object cache.
 	onClose := func() error {
-		c.mutex.Lock()
-		defer c.mutex.Unlock()
 		if buf.Len() != int(size) {
```
krishnasrinivas: previous code of keeping `Lock` and `Unlock` together was better as separating it apart does not give much advantage as it's not doing any locking calls.

pkg/objcache/objcache.go
```diff
@@ -145,36 +181,50 @@ func (c *Cache) Create(key string, size int64) (w io.WriteCloser, err error) {
 
 // Open - open the in-memory file, returns an in memory read seeker.
 // returns an error ErrNotFoundInCache, if the key does not exist.
-func (c *Cache) Open(key string) (io.ReadSeeker, error) {
-	c.mutex.RLock()
-	defer c.mutex.RUnlock()
-
+func (c *cache) Open(key string) (io.ReadSeeker, error) {
 	// Entry exists, return the readable buffer.
```
krishnasrinivas, line 167: same comment here: previous code of keeping Lock and Unlock together was better as separating it apart does not give much advantage as it's not doing any locking calls.

pkg/objcache/objcache.go
```diff
@@ -22,23 +22,42 @@ import (
...
 // which also provides expiring key mechanism and also maxSize.
 type Cache struct {
+	*cache
+}
+
+// why this? explained in New().
+type cache struct {
 	// Mutex is used for handling the concurrent
 	// read/write requests for cache
 	mutex *sync.RWMutex
```
krishnasrinivas: https://twitter.com/davecheney/status/743289114980548612

harshavardhana: > https://twitter.com/davecheney/status/743289114980548612

Yes

pkg/objcache/objcache.go
harshavardhana, line 167: > same comment here: previous code of keeping Lock and Unlock together was better as separating it apart does not give much advantage as it's not doing any locking calls.

The advantage is avoiding entire body locking, the inference here is to indicate that we are locking only the map reads/writes. I prefer this approach over other.

harshavardhana, line 167: > same comment here: previous code of keeping Lock and Unlock together was better as separating it apart does not give much advantage as it's not doing any locking calls.
> The advantage is avoiding entire body locking, the inference here is to indicate that we are locking only the map reads/writes. I prefer this approach over other.

Reverted back to the previous style, since we don't need RLock anymore.

harshavardhana:
The patch is updated for both putobject and complete multipart caching - there is one odd issue which i am still fixing.

xl-v1-multipart.go
```diff
@@ -677,7 +677,13 @@ func (xl xlObjects) CompleteMultipartUpload(bucket string, object string, upload
 	}
 	// Hold write lock on the destination before rename.
 	nsMutex.Lock(bucket, object)
-	defer nsMutex.Unlock(bucket, object)
+	defer func() {
+		// Unlock.
+		nsMutex.Unlock(bucket, object)
+
+		// Fill the cache once we have
+		go xl.fillCache(bucket, object)
```
krishnasrinivas: you could just call xl.GetObject and reuse the code.

krishnasrinivas:
@harshavardhana i would prefer garbage-collection problem and janitor go-routine to be implemented like this https://github.com/krishnasrinivas/minio/commit/8169bca56a1e0d63f3798da28889c3ad84817bd0 as it is simpler

Also i feel it's better to let GetObject fill the object-cache instead of us reading from the disk after complete-multipart-upload because if the object is never read we will be wasting the cache.

harshavardhana:
> Also i feel it's better to let GetObject fill the object-cache instead of us reading from the disk after complete-multipart-upload because if the object is never read we will be wasting the cache.

The reasoning is to handle a good case when PUT followed by GET. Basically someone uploads a bunch of files and shares the prefix with others, coming from FTP point of view it is a more useful scenario. This idea precedes the idea that object will not be read. 

GetObject() by itself will not help, let me use a combination of DeleteObject() and GetObject(). 

> @harshavardhana i would prefer garbage-collection problem and janitor go-routine to be implemented like this krishnasrinivas@8169bca as it is simpler

Yes i wanted to argue a case, was going to do what you suggested anyways.

pkg/objcache/objcache.go
```diff
@@ -146,35 +169,74 @@ func (c *Cache) Create(key string, size int64) (w io.WriteCloser, err error) {
...
 
-// Deletes the entry that was found.
-func (c *Cache) deleteEntry(key string, size int) {
-	delete(c.entries, key)
-	c.currentSize -= uint64(size)
-	c.totalEvicted++
-	if c.OnEviction != nil {
-		c.OnEviction(key)
+// StopJanitor sends a message to the janitor routine to stop.
+func (c *Cache) StopJanitor() {
```
krishnasrinivas: can be `Stop` or `Destroy` as `janitor` term need not be exposed to the cache consumer

harshavardhana: > can be Stop or Destroy as janitor term need not be exposed to the cache consumer

It is needed since consumer has to call it, quite useful when we are going to do graceful shutdown.

pkg/objcache/objcache.go
```diff
@@ -146,35 +169,74 @@ func (c *Cache) Create(key string, size int64) (w io.WriteCloser, err error) {
...
+	if c.stopJanitor != nil {
+		c.stopJanitor <- struct{}{}
+	}
+}
+
+// Start running janitor routine ticking at expiration interval, on each interval
+// this routine does a sweep across the cache entries and expires all the expired entries.
+func (c *Cache) startJanitor() {
+	ticker := time.NewTicker(c.expiration)
```
krishnasrinivas: janitor run interval should be different than expiration value. i.e janitor should run more frequently and not default 72hrs.

also conceptually time.After() is more correct than time.NewTicker() because if you use time.NewTicker() technically it can tick when c.DeleteExpired() is actually happening.

PR #2133, opened 2016-07-07
ObjectLayer/GetObject: Should return the right error value. Fix done in FS and XL.

### LINKED ISSUE #2117: XL/FS: ObjectLayer.GetObject() incorrect behaviour while trying to fetch from an non-existent bucket.
## Expected behaviour

To error out with `Bucket not exist`
## Actual behaviour
-  FS strangely passes the test while its expected to error, here is the log.  Not sure how does it pass the test on an non-existent bucket. 
  `object-api-getobject_test.go:118: Test 1: SingleNode: Expected to fail with <ERROR> "Bucket not found: abc", but passed instead`
-  XL fails with the following error.
  `object-api-getobject_test.go:123: Test 1: XL: Expected to fail with error "Bucket not found: abc", but instead failed with error "The requested range is not satisfiable" instead`

run the test here https://github.com/hackintoshrao/minio/commit/6ff382bc018177433846a090b4d7675f61716a52
## Minio version

Latest master at this point 169c72cdab14ac0f993161f6ed97e8462eccfb97 . 
## System information

OSX ElCaptain.

krishnasrinivas:
fixes #2117

@harshavardhana, recudeErrs() was the easiest solution I could think of, not sure if it's the cleanest.

Also this behavior is not observed in an actual setup as GetObjectInfo is done in GetObjectHandler() which returned the right error.

### DIFF
--- a/xl-v1-object.go
+++ b/xl-v1-object.go
@@ -67,6 +67,11 @@ func (xl xlObjects) GetObject(bucket, object string, startOffset int64, length i
 		return toObjectErr(errXLReadQuorum, bucket, object)
 	}
 
+	// If all the disks returned error, we return error.
+	if err := reduceErrs(errs); err != nil {
+		return toObjectErr(err, bucket, object)
+	}
+
 	// List all online disks.
 	onlineDisks, highestVersion, err := xl.listOnlineDisks(metaArr, errs)
 	if err != nil {
--- a/xl-v1-utils.go
+++ b/xl-v1-utils.go
@@ -23,6 +23,59 @@ import (
 	"path"
 )
 
+// Returns nil even if one of the slice elements is nil.
+// Else returns the error which occours the most.
+func reduceErrs(errs []error) error {
+	// In case the error type is not in the known error list.
+	var unknownErr = errors.New("unknown error")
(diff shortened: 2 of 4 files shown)


xl-v1-object.go
```diff
@@ -63,6 +63,10 @@ func (xl xlObjects) GetObject(bucket, object string, startOffset int64, length i
 		return toObjectErr(errXLReadQuorum, bucket, object)
 	}
 
+	if err := reduceErrs(errs); err != nil {
```
harshavardhana, line 71: A comment here

xl-v1-utils.go
```diff
@@ -23,6 +23,21 @@ import (
 	"time"
 )
 
+// Returns nil even if one of the slice elements is nil.
+// Else returns the error in the first slice element.
```
harshavardhana: We could explain what 'reduceErrs' do this is not explaining the non-obvious of why it is done as such. 

```
reduceErrs()
```

reduces the slice of error elements into a singleton error. Also we could modify this to return only the error where all elements have returned the same error.  Current function is non-deterministic as you would see while writing a unit test.

A unit test is needed here.

PR #2135, opened 2016-07-07
XL/GetObject: If quorum not available during GetObject appropriate error should be returned.

### LINKED ISSUE #2129: XL: when readQuorum fails, GetObject returns 200OK with no content
in 8:8 setup, when 8 disks are down:

```
krishna@escape:~$ mc --debug cat myminio/testbucket/passwd
mc: <DEBUG> GET /testbucket/?location= HTTP/1.1
Host: localhost:9000
User-Agent: Minio (linux; amd64) minio-go/1.0.1 mc/2016-07-06T11:47:13Z
Authorization: AWS4-HMAC-SHA256 Credential=**REDACTED**/20160707/us-east-1/s3/aws4_request, SignedHeaders=host;x-amz-content-sha256;x-amz-date, Signature=**REDACTED**
X-Amz-Date: 20160707T142610Z
Accept-Encoding: gzip

mc: <DEBUG> HTTP/1.1 200 OK
Transfer-Encoding: chunked
Accept-Ranges: bytes
Content-Type: text/xml; charset=utf-8
Date: Thu, 07 Jul 2016 14:26:10 GMT
Server: Minio/DEVELOPMENT.GOGET (linux; amd64)
Vary: Origin
X-Amz-Request-Id: 9O64XXWK7YYGQHKA

mc: <DEBUG> Response Time:  1.763014ms

mc: <DEBUG> GET /testbucket/passwd HTTP/1.1
Host: localhost:9000
User-Agent: Minio (linux; amd64) minio-go/1.0.1 mc/2016-07-06T11:47:13Z
Authorization: AWS4-HMAC-SHA256 Credential=**REDACTED**/20160707/us-east-1/s3/aws4_request, SignedHeaders=expect;host;x-amz-content-sha256;x-amz-date, Signature=**REDACTED**
Expect: 100-continue
X-Amz-Date: 20160707T142610Z
Accept-Encoding: gzip

mc: <DEBUG> HTTP/1.1 200 OK
Content-Length: 2204
Accept-Ranges: bytes
Date: Thu, 07 Jul 2016 14:26:10 GMT
Etag: "b9a59b85f2dd2b7fa57a82be7b15c806"
Last-Modified: Thu, 07 Jul 2016 14:24:08 GMT
Server: Minio/DEVELOPMENT.GOGET (linux; amd64)
Vary: Origin
X-Amz-Request-Id: WOE87UCMUTPI49PI

mc: <DEBUG> Response Time:  1.946786ms

krishna@escape:~$ 
```

Should have returned quorum/internal error.

krishnasrinivas:
fixes #2129

@harshavardhana I don't know how it behaves if headers have already been sent down the wire because parial data was already written on the response writer. But it works as expected if no data was sent down the socket when the quorum error is returned.

### DIFF
--- a/object-handlers.go
+++ b/object-handlers.go
@@ -65,6 +65,13 @@ func errAllowableObjectNotFound(bucket string, r *http.Request) APIErrorCode {
 	return ErrNoSuchKey
 }
 
+// Simple way to convert a func to io.Writer type.
+type funcToWriter func([]byte) (int, error)
+
+func (f funcToWriter) Write(p []byte) (int, error) {
+	return f(p)
+}
+
 // GetObjectHandler - GET Object
 // ----------
 // This implementation of the GET operation retrieves object. To use GET,
@@ -106,23 +113,21 @@ func (api objectAPIHandlers) GetObjectHandler(w http.ResponseWriter, r *http.Req
 
 	// Get request range.
 	var hrange *httpRange
-	if hrange, err = parseRequestRange(r.Header.Get("Range"), objInfo.Size); err != nil {
-		// Handle only errInvalidRange
-		// Ignore other parse error and treat it as regular Get request like Amazon S3.
-		if err == errInvalidRange {
-			writeErrorResponse(w, r, ErrInvalidRange, r.URL.Path)
-			return
+	rangeHeader := r.Header.Get("Range")
+	if rangeHeader != "" {
+		if hrange, err = parseRequestRange(rangeHeader, objInfo.Size); err != nil {
+			// Handle only errInvalidRange
+			// Ignore other parse error and treat it as regular Get request like Amazon S3.
+			if err == errInvalidRange {
+				writeErrorResponse(w, r, ErrInvalidRange, r.URL.Path)
+				return
+			}
+
+			// log the error.
+			errorIf(err, "Invalid request range")
 		}
 
-		// log the error.
-		errorIf(err, "Invalid request range")
 	}
-
-	// Set standard object headers.
-	setObjectHeaders(w, objInfo, hrange)
-
-	// Set any additional requested response headers.
-	setGetRespHeaders(w, r.URL.Query())
 
 	// Verify 'If-Modified-Since' and 'If-Unmodified-Since'.
 	lastModified := objInfo.ModTime
@@ -141,13 +146,44 @@ func (api objectAPIHandlers) GetObjectHandler(w http.ResponseWriter, r *http.Req
 		startOffset = hrange.offsetBegin
 		length = hrange.getLength()
 	}
+	// Indicates if any data was written to the http.ResponseWriter
+	dataWritten := false
+	// io.Writer type which keeps track if any data was written.
+	writer := funcToWriter(func(p []byte) (int, error) {
+		if !dataWritten {
+			// Set headers on the first write.
+			// Set standard object headers.
+			setObjectHeaders(w, objInfo, hrange)
+
+			// Set any additional requested response headers.
+			setGetRespHeaders(w, r.URL.Query())
+
+			dataWritten = true
+		}
+		return w.Write(p)
+	})
 	// Reads the object at startOffset and writes to mw.
-	if err := api.ObjectAPI.GetObject(bucket, object, startOffset, length, w); err != nil {
+	if err := api.ObjectAPI.GetObject(bucket, object, startOffset, length, writer); err != nil {
 		errorIf(err, "Unable to write to client.")
-		// Do not send error response here, client would have already died.
+		if !dataWritten {
+			// Error response only if no data has been written. i.e if
+			// partial data has already been written before an error
+			// occoured then no point in setting StatusCode and
+			// sending error XML.
+			apiErr := toAPIErrorCode(err)
+			writeErrorResponse(w, r, apiErr, r.URL.Path)
+		}
 		return
 	}
-	// Success.
+	if !dataWritten {
+		// If ObjectAPI.GetObject did not return error and no data has
+		// been written it would mean that it is a 0-byte object.
+		// Appropriate headers need to be set.
+		setObjectHeaders(w, objInfo, hrange)
+
+		// Set any additional requested response headers.
+		setGetRespHeaders(w, r.URL.Query())
+	}
 }
 
 // checkLastModified implements If-Modified-Since and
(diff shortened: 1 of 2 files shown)


object-handlers.go
```diff
@@ -144,7 +144,10 @@ func (api objectAPIHandlers) GetObjectHandler(w http.ResponseWriter, r *http.Req
 	// Reads the object at startOffset and writes to mw.
 	if err := api.ObjectAPI.GetObject(bucket, object, startOffset, length, w); err != nil {
 		errorIf(err, "Unable to write to client.")
-		// Do not send error response here, client would have already died.
+		// In case headers haven't been written to client, unset the headers and send an error response.
+		unsetObjectHeaders(w)
+		apiErr := toAPIErrorCode(err)
+		writeErrorResponse(w, r, apiErr, r.URL.Path)
```
harshavardhana: You cannot do this, response writer cannot be written again once you have started writing.

harshavardhana:
> @harshavardhana I don't know how it behaves if headers have already been sent down the wire because parial data was already written on the response writer. But it works as expected if no data was sent down the socket when the quorum error is returned.

Golang implementation doesn't allow this to happen, once you have started writing to response writer you cannot again write to it even if there is a partial data.

harshavardhana:
Also you need to be absolutely sure that the data was not written yet, which is not possible in this scenario. As this might lead to races for http.ResponseWriter which starts printing messages on the screen. 

The idea is simple you need to write a non http.StatusOK on the first write to the response writer could be an error etc this could be a custom closer wrapped in so that if the Close() occurs before the Write() happened we write non http.StatusOK by doing w.responseWriter.WriteHeader(errstatus)

object-handlers.go
```diff
@@ -106,24 +113,22 @@ func (api objectAPIHandlers) GetObjectHandler(w http.ResponseWriter, r *http.Req
...
-		// log the error.
-		errorIf(err, "Invalid request range")
 	}
 
-	// Set standard object headers.
-	setObjectHeaders(w, objInfo, hrange)
-
-	// Set any additional requested response headers.
-	setGetRespHeaders(w, r.URL.Query())
-
```
harshavardhana, line 126: Now that you moved this would introduce a bug in checkSourceETag() subsequently, change the function to take in 'ETag' rather than look for responseWriter to have ETag set.

object-handlers.go
```diff
@@ -141,13 +146,44 @@ func (api objectAPIHandlers) GetObjectHandler(w http.ResponseWriter, r *http.Req
...
+			// partial data has already been written before an error
+			// occoured then no point in setting StatusCode and
+			// sending error XML.
+			apiErr := toAPIErrorCode(err)
+			writeErrorResponse(w, r, apiErr, r.URL.Path)
+		}
 		return
 	}
-	// Success.
+	if !dataWritten {
```
harshavardhana, line 173: This is harder to follow, why even read if its a '0byte' object? - we have that value with us anyways in objInfo.Size - we could set the headers and return? not even have funcToWriter() for it.

object-handlers.go
```diff
@@ -141,13 +146,44 @@ func (api objectAPIHandlers) GetObjectHandler(w http.ResponseWriter, r *http.Req
...
+	})
 	// Reads the object at startOffset and writes to mw.
-	if err := api.ObjectAPI.GetObject(bucket, object, startOffset, length, w); err != nil {
+	if err := api.ObjectAPI.GetObject(bucket, object, startOffset, length, writer); err != nil {
 		errorIf(err, "Unable to write to client.")
-		// Do not send error response here, client would have already died.
+		if !dataWritten {
+			// Error response only if no data has been written. i.e if
+			// partial data has already been written before an error
+			// occoured then no point in setting StatusCode and
```
harshavardhana: 'occurred'

object-handlers.go
```diff
@@ -141,13 +146,44 @@ func (api objectAPIHandlers) GetObjectHandler(w http.ResponseWriter, r *http.Req
...
+		}
+		return w.Write(p)
+	})
 	// Reads the object at startOffset and writes to mw.
-	if err := api.ObjectAPI.GetObject(bucket, object, startOffset, length, w); err != nil {
+	if err := api.ObjectAPI.GetObject(bucket, object, startOffset, length, writer); err != nil {
 		errorIf(err, "Unable to write to client.")
-		// Do not send error response here, client would have already died.
+		if !dataWritten {
+			// Error response only if no data has been written. i.e if
```
harshavardhana: 'Write error response only if not data has been written yet to the client'

harshavardhana:
@krishnasrinivas can you address or reply to comments? - lets get this merged soon.

object-handlers.go
```diff
@@ -178,57 +173,85 @@ func (api objectAPIHandlers) GetObjectHandler(w http.ResponseWriter, r *http.Req
...
+	return false
+}
+
+// returns true if object was modified after givenTime.
+func ifModifiedSince(objTime time.Time, givenTimeStr string) bool {
+	givenTime, err := time.Parse(http.TimeFormat, givenTimeStr)
+	if err != nil {
+		return true
+	}
+	if objTime.After(givenTime) {
```
harshavardhana, line 252: `return objTime.After(givenTime)`

krishnasrinivas:
@harshavardhana I haven't taken care of this:

```
-       // The Date-Modified header truncates sub-second precision, so
-       // use mtied me < t+1s instead of mtime <= t to check for unmodified.
```

See if it is needed in the refactored code.

Also I tried doing:

> > This is harder to follow, why even read if its a '0byte' object? - we have that value with us anyways in objInfo.Size - we could set the headers and return? not even have funcToWriter() for it.

But realized i get offset and length and quorum validation for free if i call GetObject() hence simplified the code to this:

```
        if !dataWritten {                                                                                                                                                                                                         
                // If ObjectAPI.GetObject did not return error and no data has                                                                                                                                                    
                // been written it would mean that it is a 0-byte object.                                                                                                                                                         
                // call wrter.Write(nil) to set appropriate headers.                                                                                                                                                              
                writer.Write(nil)                                                                                                                                                                                                 
        }           
```

harshavardhana:
> @harshavardhana I haven't taken care of this:
> -       // The Date-Modified header truncates sub-second precision, so
> -       // use mtied me < t+1s instead of mtime <= t to check for unmodified.
>   See if it is needed in the refactored code.

Yeah this is needed, there is a bug which can happen with applications.

PR #2144, opened 2016-07-08
Move isLeaf check to treeWalk from listDir

krisis:


krishnasrinivas:
@krisis isLeaf() check has to happen before we do `sort.Strings(entries)` hence it is done inside listDir() that is:
`a-b/` < `a/` but if we strip trailing `/` then `a` < `a-b` i.e sort() will give different results if you include trailing `/`

krisis:
@krishnasrinivas correct. I missed this part. This fix will not work, so closing this PR.

PR #2145, opened 2016-07-08
Add PutObjectPart benchmark

vadmeste:


benchmark-utils_test.go
```diff
@@ -67,6 +67,72 @@ func runPutObjectBenchmark(b *testing.B, obj ObjectLayer, objSize int) {
...
+	hasher.Write([]byte(textData))
+	metadata := make(map[string]string)
+	metadata["md5Sum"] = hex.EncodeToString(hasher.Sum(nil))
+	// benchmark utility which helps obtain number of allocations and bytes allocated per ops.
+	b.ReportAllocs()
+	// the actual benchmark for PutObjectPart starts here. Reset the benchmark timer.
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		// insert the object.
+		uploadID, err = obj.NewMultipartUpload(bucket, "object"+strconv.Itoa(i), metadata)
```
hackintoshrao: The NewMultipart shouldn't happen in loop... This is will benchmark NewMultipart and insertion of one part. 
- The initiation should happen outside the loop.  
- There should be a function which takes the size of each part, upload ID and uploads a 1GB file.
- Something like this `PutPart1GB(partsize int, uploadID string )`. 
- The above function should be benchmarked for various sizes of parts.

vadmeste: > The NewMultipart shouldn't happen in loop

I thought about that too.. However, it seems for me that we need to insert a different object in each benchmark iteration to have a more correct bench numbers, and that will work only if we call NewMultipartUpload for each iteration too.. What do you think?

hackintoshrao: Missed out on the comment. Sorry for responding late. The problem here is the we'll not be benchmarking a practical scenario. In a practical workload case with thousands of uploads the `NewMultipart` will only be initiated once per each upload and this will be followed by thousands of parts being uploaded for a big object.

hackintoshrao: What we are really interested here is to evaluate the upload of tens/hundreds/thousands of parts once a multi part operation is initiated. Also the difference in performance with difference part size for upload of object of same size.

harshavardhana:
LGTM, Thanks :+1:

PR #2158, opened 2016-07-10
server: Implement --ignore-disks for ignoring disks from healing.

### LINKED ISSUE #2128: XL: detect if a disk has been unmounted
If minio is started as:
`minio server /mnt/disk1 /mnt/disk2 /mnt/disk3...`

`/mnt/disk1 /mnt/disk2 /mnt/disk3...` are mount points for the disks. Now if user does `umount /mnt/disk` we have no way to tell that the disk is not available, we will just see empty `/mnt/disk` and try to initialize it.

harshavardhana:
By default server heals/creates missing directories and re-populates `format.json`, in some scenarios when disk is down for maintenance it would be beneficial for users to ignore such disks rather than
mistakenly using `root` partition.

Fixes #2128

### DIFF
--- a/routers.go
+++ b/routers.go
@@ -23,16 +23,15 @@ import (
 	router "github.com/gorilla/mux"
 )
 
-// newObjectLayer - initialize any object layer depending on the
-// number of export paths.
-func newObjectLayer(exportPaths []string) (ObjectLayer, error) {
-	if len(exportPaths) == 1 {
-		exportPath := exportPaths[0]
+// newObjectLayer - initialize any object layer depending on the number of disks.
+func newObjectLayer(disks, ignoredDisks []string) (ObjectLayer, error) {
+	if len(disks) == 1 {
+		exportPath := disks[0]
 		// Initialize FS object layer.
 		return newFSObjects(exportPath)
 	}
 	// Initialize XL object layer.
-	objAPI, err := newXLObjects(exportPaths)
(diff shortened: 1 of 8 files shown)


routers.go
```diff
@@ -41,11 +40,11 @@ func newObjectLayer(exportPaths []string) (ObjectLayer, error) {
 
 // configureServer handler returns final handler for the http server.
 func configureServerHandler(srvCmdConfig serverCmdConfig) http.Handler {
-	objAPI, err := newObjectLayer(srvCmdConfig.exportPaths)
+	objAPI, err := newObjectLayer(srvCmdConfig.disks, srvCmdConfig.ignoredDisks)
 	fatalIf(err, "Unable to intialize object layer.")
 
 	// Initialize storage rpc server.
-	storageRPC, err := newRPCServer(srvCmdConfig.exportPaths[0]) // FIXME: should only have one path.
+	storageRPC, err := newRPCServer(srvCmdConfig.disks[0]) // FIXME: should only have one path.
```
abperiasamy, line 47: RPC is initialized only for first disk?. If it is broken, we should either fix it or disable it till distributed version.

abperiasamy:
I will merge it for now. RPC layer can be fixed in a separate PR.

routers.go
abperiasamy, line 47: There is a design problem here. They have to be initialized differently. Let me accept this patch and I will propose new changes separately.

PR #2166, opened 2016-07-11
XL: listOnlineDisks should use modTime instead of version.

harshavardhana:
This change is needed to make reading from objects future proof
in-terms of handling online disks. Our current counter is not
based on affirmative knowledge and relies on arithmetic sequence
which can lead to bugs.

Using modTime simplifies the understanding of `xl.json` and future
tooling / debugging of the format.


abperiasamy:
We need to use modTime (string match) instead of a monotonic counter to determine if the commits indeed belong to the same batch. Incremental counters are expensive and slow.

krishnasrinivas:
@harshavardhana can you tell what is the problem with current version numbering? and how is commitID better?

harshavardhana:
> @harshavardhana can you tell what is the problem with current version numbering? and how is commitID better?

The patch is changed now @abperiasamy's comment. We are removing the 'Version' field and using modTime itself to provide us this information.

harshavardhana:
@krisis found a problem in this approach for 'modTime' even if we use UTC() since its dependency on the system clock would lead us to an unknown situation when a system clock if it goes back in time.. could lead us to a different object content with same modtime.

harshavardhana:
Patch updated with using.. UUID instead

xl-v1-healing.go
```diff
@@ -100,23 +157,22 @@ func (xl xlObjects) shouldHeal(onlineDisks []StorageAPI) (heal bool) {
...
 	onlineDisks = make([]StorageAPI, len(xl.storageDisks))
 
-	// List all the file versions from partsMetadata list.
-	versions := listObjectVersions(partsMetadata, errs)
+	// List all the file commit ids from parts metadata.
+	uuids := listObjectUUIDs(partsMetadata, errs)
 
-	// Get highest object version.
-	highestVersion := highestInt(versions, int64(1))
+	// Reduce list of UUIDs to a single common value.
```
krisis, line 141: s/to a single common value/to a maximally occurring value/

xl-v1-multipart.go
```diff
@@ -716,6 +710,7 @@ func (xl xlObjects) CompleteMultipartUpload(bucket string, object string, upload
 	// Save the final object size and modtime.
 	xlMeta.Stat.Size = objectSize
 	xlMeta.Stat.ModTime = time.Now().UTC()
+	xlMeta.Stat.UUID = getUUID()
```
krisis: do we need a new UUID during completeMultipartUpload? Can't we retain the UUID assigned during putObjectPart request for the same uploadID?

harshavardhana: > do we need a new UUID during completeMultipartUpload? Can't we retain the UUID assigned during putObjectPart request for the same uploadID?

It is better to handle scenarios of failure as well here.

harshavardhana:
> s/to a single common value/to a maximally occurring value/

Correct.

xl-v1-healing.go
```diff
@@ -16,33 +16,90 @@
...
 	}
-	return highestInteger
+
+	// Eventually if we see the occurrence of a certain element
+	// is higher than maxima, we return the last element.
+	if count > maxima {
+		commonUUID = prevUUID
+	} // else for all other cases we return the choosen common element.
+	return commonUUID
+}
```
krishnasrinivas, line 45: It was hard to read the logic, easier would be to use map (though map will be inefficient than slice):

```
func commonUUID(uuids []string) (uuid string) {                                                                                                                                 
        uuidMap := make(map[string]int)                                                                                                                                         
        for _, uuid := range uuids {                                                                                                                                            
                if uuid == uuidSentinel {                                                                                                                                       
                        continue                                                                                                                                                
                }                                                                                                                                                               
                uuidMap[uuid]++                                                                                                                                                 
        }                                                                                                                                                                       
        commonUUID := ""                                                                                                                                                        
        maxCount := 0                                                                                                                                                           
        for uuid, count := range uuidMap {                                                                                                                                      
                if count > maxCount {                                                                                                                                           
                        maxCount = count                                                                                                                                        
                        commonUUID = uuid                                                                                                                                       
                }                                                                                                                                                               
        }                                                                                                                                                                       
        return commonUUID                                                                                                                                                       
}                           
```

xl-v1-healing.go
```diff
@@ -16,33 +16,90 @@
...
+	// Variables.
+	var (
+		pIndex     int    // Holds the value of th first non-sentinel element.
+		prevUUID   string // Holds the value of the first non-sentinel element.
+		commonUUID string // Holds the final value of uuid most occurred.
+		count      int    // Counter for same occurrence of elements.
+		maxima     int    // Counter for remembering max occurrence of elements.
+	)
+	// Sort so that the sentinel value are staggerred at the beginning.
+	sort.Strings(uuids)
```
krishnasrinivas: this affects the original uuids slice

xl-v1-healing.go
```diff
@@ -100,23 +157,22 @@ func (xl xlObjects) shouldHeal(onlineDisks []StorageAPI) (heal bool) {
...
-	// Get highest object version.
-	highestVersion := highestInt(versions, int64(1))
+	// Reduce list of UUIDs to a single common value.
+	uuid = commonUUID(uuids)
 
-	// Pick online disks with version set to highestVersion.
-	for index, version := range versions {
-		if version == highestVersion {
+	// Create a new online disks slice, which have common uuid.
+	for index, u := range uuids {
```
krishnasrinivas: sorting in commonUUID() will cause problems here

xl-v1-utils.go
```diff
@@ -78,14 +78,14 @@ func reduceErrs(errs []error) error {
...
 	for _, err := range errs {
-		if err == errDiskNotFound {
+		if err == errDiskNotFound || err == errFaultyDisk || err == errDiskAccessDenied {
 			continue
 		}
-		diskFoundCount++
+		errCount++
 	}
-	return diskFoundCount >= minQuorumCount
+	return errCount >= minQuorumCount
```
krishnasrinivas, line 88: instead of errCount, it can be named just count.
Because it reads as `if error count is more than quorumCount we have met the quorum` when it should mean the opposite

harshavardhana, line 88: > instead of errCount, it can be named just count.
> Because it reads as if error count is more than quorumCount we have met the quorum when it should mean the opposite

Oops yeah.

PR #2173, opened 2016-07-11
XL/bucket: Remove bucket only if incomplete uploads is empty as well.

### LINKED ISSUE #2170: XL: removing a bucket won't clean incomplete uploads
For mc users, it won't be possible to clean incomplete uploads after removing a bucket:
`mc rm -r --force myminio/bucket/`

harshavardhana:
Fixes #2170

### DIFF
--- a/object-common.go
+++ b/object-common.go
@@ -171,22 +171,22 @@ func cleanupDir(storage StorageAPI, volume, dirPath string) error {
 	// Function to delete entries recursively.
 	delFunc = func(entryPath string) error {
 		if !strings.HasSuffix(entryPath, slashSeparator) {
-			// No trailing "/" means that this is a file which can be deleted.
+			// Delete the file entry.
 			return storage.DeleteFile(volume, entryPath)
 		}
+
 		// If it's a directory, list and call delFunc() for each entry.
 		entries, err := storage.ListDir(volume, entryPath)
-		if err != nil {
-			if err == errFileNotFound {
-				// if dirPath prefix never existed.
-				return nil
-			}
+		// If entryPath prefix never existed, safe to ignore.
+		if err == errFileNotFound {
+			return nil
+		} else if err != nil { // For any other errors fail.
 			return err
-		}
-		for _, entry := range entries {
-			err = delFunc(pathJoin(entryPath, entry))
+		} // else on success..
 
-			if err != nil {
+		// Recurse and delete all other entries.
+		for _, entry := range entries {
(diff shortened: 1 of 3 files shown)


object-common.go
```diff
@@ -171,21 +172,28 @@ func cleanupDir(storage StorageAPI, volume, dirPath string) error {
...
-				return nil
-			}
+		if err != nil && err != errFileNotFound {
+			// If entryPath prefix never existed, safe to ignore.
+			// For any other errors fail.
 			return err
 		}
+		if len(entries) == 0 {
+			// Delete volume when entries are empty.
+			err = storage.DeleteVol(path.Join(volume, entryPath))
```
krishnasrinivas: This gives the impression that Volume and directories are same, Because, if there is an empty dir somewhere insde .minio/multipart upload it gets cleanedup by a DeleteVol call

Calling an explicit DeleteVol on `.minio/multipart` or `.minio/tmp` is fine because we treat them as volumes, but DeleteVol should not be done on anything inside it

object-common.go
```diff
@@ -171,21 +172,28 @@ func cleanupDir(storage StorageAPI, volume, dirPath string) error {
...
 		// If it's a directory, list and call delFunc() for each entry.
 		entries, err := storage.ListDir(volume, entryPath)
-		if err != nil {
-			if err == errFileNotFound {
-				// if dirPath prefix never existed.
-				return nil
-			}
+		if err != nil && err != errFileNotFound {
+			// If entryPath prefix never existed, safe to ignore.
+			// For any other errors fail.
```
krishnasrinivas: previous code was correct here right? i.e return any error other than `errFileNotFound`

xl-v1-bucket.go
```diff
@@ -238,7 +239,14 @@ func (xl xlObjects) DeleteBucket(bucket string) error {
 		// Delete volume inside a go-routine.
 		go func(index int, disk StorageAPI) {
 			defer wg.Done()
-			err := disk.DeleteVol(bucket)
+			// Cleanup all the previously incomplete multiparts.
```
krishnasrinivas: better to cleanup incomplete uploads after DeleteVol succeeds. i.e if DeleteVol fails for some reason (ex if it has objects in it) we should not delete the multipart entries.

harshavardhana: > better to cleanup incomplete uploads after DeleteVol succeeds. i.e if DeleteVol fails for some reason (ex if it has objects in it) we should not delete the multipart entries.

That is correct.

krishnasrinivas:
looks good

PR #2184, opened 2016-07-12
Store and retrieve user-defined object metadata in a .metadata bucket

yrashk:
Also fixes a problem with a non-canonical header naming validation

P.S. I wasn't quite sure where to write an actual HTTP PUT/GET test (where are these tests in the test suite, a cursory glance didn't reveal them?), but if you can advise me on this, I will happily amend this PR.

### DIFF
--- a/fs-v1-multipart.go
+++ b/fs-v1-multipart.go
@@ -240,6 +240,12 @@ func (fs fsObjects) newMultipartUpload(bucket string, object string, meta map[st
 		}
 		return "", toObjectErr(err, minioMetaBucket, uploadIDPath)
 	}
+
+	err = PutObjectMetadata(fs.storage, bucket, object, meta)
+	if err != nil {
+		return "", toObjectErr(err, bucket, object)
+	}
+
 	// Return success.
 	return uploadID, nil
 }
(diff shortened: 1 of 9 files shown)


harshavardhana:
Thanks for the patch @yrashk.  

Quick question is there a problem in the XL code where you are not able to retrieve the metadata? currently the meta is saved as part of the `xl.json` ?

GetObjectInfo() itself should be used and extended to support these.. and it can be easily achieved. 

In case of FS we deliberately kept it simpler and consciously decided not to store this information.

yrashk:
I didn't use XL for our own needs — simply filled in the blanks :) 

Is there any chance you might be persuaded to allow the storage of user metadata for the FS backend? There are use cases when it's useful.

harshavardhana:
> Is there any chance you might be persuaded to allow the storage of user metadata for the FS backend? There are use cases when it's useful.

Will discuss this with @abperiasamy and update here.  It is necessary to some extent for applications relying on custom headers, but what we wished to avoid is creating custom files in the backend for a normal FS user.

Problem can aggravate when someone runs FS on bunch of existing files, which is a general use case for FS. User could potentially delete this and corruption in this meta information is just not protected enough. 

Anyhow will update here.

harshavardhana:
Thanks for all the work @yrashk

krishnasrinivas:
@yrashk just curious,  can you tell how you guys are using minio FS, i.e the use case.

PR #2189, opened 2016-07-12
Fix metadata storage for compatibility with S3

utsl42:
This fixes a bug in PutObjectHandler that keeps it from saving metadata headers, and enables returning them back through ObjectInfo.


harshavardhana:
Thanks @utsl42 - this looks good.

PR #2193, opened 2016-07-12
Refactor of xl.PutObjectPart and erasureCreateFile.

krishnasrinivas:
Needs review and testing.

erasureCreateFile is much simpler -
- it does not worry about disk ordering
- knowledge of XL's data structures is taken out

PutObjectPart refactored -
- works on orderedDisks which makes reading code easier.
- simplified code.

### DIFF
--- a/xl-v1-multipart.go
+++ b/xl-v1-multipart.go
@@ -354,10 +354,13 @@ func (xl xlObjects) PutObjectPart(bucket, object, uploadID string, partID int, s
 	nsMutex.RUnlock(minioMetaBucket, uploadIDPath)
 
 	// List all online disks.
-	onlineDisks, _ := xl.listOnlineDisks(partsMetadata, errs)
+	onlineDisks, modTime := listOnlineDisks(xl.storageDisks, partsMetadata, errs)
 
 	// Pick one from the first valid metadata.
-	xlMeta := pickValidXLMeta(partsMetadata)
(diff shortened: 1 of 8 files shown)


xl-v1-multipart.go
```diff
@@ -433,70 +429,54 @@ func (xl xlObjects) PutObjectPart(bucket, object, uploadID string, partID int, s
 		return "", toObjectErr(err, minioMetaBucket, partPath)
 	}
 
-	// Read metadata (again) associated with the object from all disks.
-	partsMetadata, errs = readAllXLMetadata(onlineDisks, minioMetaBucket,
+	// Read metadata associated with the object from all disks.
+	partsMetadata, errs = readAllXLMetadata(xl.storageDisks, minioMetaBucket,
 		uploadIDPath)
 	if !isQuorum(errs, xl.writeQuorum) {
+		nsMutex.RUnlock(minioMetaBucket, uploadIDPath)
```
harshavardhana: This is not needed, would lead to panic.

krishnasrinivas:
@harshavardhana i think this can be merged for the release, the code is simpler and the logic is not too different from the previous code.

harshavardhana:
> @harshavardhana i think this can be merged for the release, the code is simpler and the logic is not too different from the previous code.

Will wait this one out as it doesn't fix anything blocking and merge it after we make the release @krishnasrinivas

krishnasrinivas:
@harshavardhana rebased.

PR #2199, opened 2016-07-13
xl: Make namespace locking granular for PutObject

krisis:


xl-v1-utils.go
```diff
@@ -23,57 +23,27 @@ import (
...
-	// For each err in []errs increment the corresponding count value.
+// Returns number of errors that occurred the most (incl. nil) and the
+// corresponding error value. N B when there is more than one error value that
+// occurs equal and maximum number of times, the error value returned depends
+// on how golang's map datastructure arranges its keys. This doesn't affect
+// correctness as long as quorum value is greater than or equal to simple
+// majority, since none of the equally maximal values would not occur quorum or
+// more number of times.
+func reduceErrs(errs []error) (int, error) {
+	errorCounts := make(map[error]int)
```
krishnasrinivas, line 34: ah, map is a better way. what was I thinking :+1:

harshavardhana:
> ah, map is a better way. what was I thinking 

lemons :lemon:

xl-v1-utils.go
```diff
@@ -23,57 +23,27 @@ import (
...
-	// to store it so that it can be used for the return error type.
-	var unknownErrType error
-
-	// For each err in []errs increment the corresponding count value.
+// Returns number of errors that occurred the most (incl. nil) and the
+// corresponding error value. N B when there is more than one error value that
+// occurs equal and maximum number of times, the error value returned depends
+// on how golang's map datastructure arranges its keys. This doesn't affect
+// correctness as long as quorum value is greater than or equal to simple
+// majority, since none of the equally maximal values would not occur quorum or
```
krisis: s/would not occur/would occur

xl-v1-object.go
```diff
@@ -68,8 +68,12 @@ func (xl xlObjects) GetObject(bucket, object string, startOffset int64, length i
 	}
 
 	// If all the disks returned error, we return error.
-	if err := reduceErrs(errs); err != nil {
-		return toObjectErr(err, bucket, object)
+	if errCount, err := reduceErrs(errs); err != nil {
+		if errCount < xl.readQuorum {
```
krishnasrinivas, line 72: did you mean `if errCount > xl.readQuorum {` ?

already a `if !isDiskQuorum(errs, xl.readQuorum) {}` check happens just before this, so returning `err` here will give more insight on the error type than a generic errXLReadQuorum.
For example let's say, all the disks returned errVolumeNotFound, we should return errVolumeNotFound and not errXLReadQuorum

xl-v1-object.go
```diff
@@ -368,20 +372,28 @@ func (xl xlObjects) PutObject(bucket string, object string, size int64, data io.
...
-
-	// Initialize xl meta.
-	xlMeta := newXLMetaV1(object, xl.dataBlocks, xl.parityBlocks)
-
+	var partsMetadata []xlMetaV1
+	var errs []error
+	nsMutex.RLock(bucket, object)
 	// Read metadata associated with the object from all disks.
-	partsMetadata, errs := readAllXLMetadata(xl.storageDisks, bucket, object)
+	partsMetadata, errs = readAllXLMetadata(xl.storageDisks, bucket, object)
```
krishnasrinivas: you could use previous style `partsMetadata, errs := readAllXLMetadata(xl.storageDisks, bucket, object)` right?

krisis: Yes. I did this anticipating that I might have to re-read it under the write lock to fetch updated values. But that wasn't required. I shall change it to look as before.

PR #2206, opened 2016-07-14
vendorize: update all vendorized packages.

harshavardhana:
Bring in new changes from upstream for all the packages.

Important ones include
- gorilla/mux
- logrus
- jwt


krishnasrinivas:
:+1:

PR #2211, opened 2016-07-15
Refactor xl.GetObject and erasureReadFile.

krishnasrinivas:


xl-v1-object.go
```diff
@@ -173,8 +169,17 @@ func (xl xlObjects) GetObject(bucket, object string, startOffset int64, length i
...
+		for index := range onlineDisks {
+			checkSums[index], _, err = metaArr[index].GetCheckSum(partName)
+			if err != nil {
+				return toObjectErr(err, bucket, object)
+			}
+		}
+
 		// Start reading the part name.
-		n, err := erasureReadFile(mw, onlineDisks, bucket, pathJoin(object, partName), partName, eInfos, partOffset, readSize, partSize)
+		n, err := erasureReadFile(mw, onlineDisks, bucket, pathJoin(object, partName), partOffset, readSize, partSize, xlMeta.Erasure.BlockSize, xl.dataBlocks, xl.parityBlocks, checkSums)
```
harshavardhana: you need to pass `xlMeta.Erasure.DataBlocks`, `xlMeta.Erasure.ParityBlocks`

xl-v1-object.go
```diff
@@ -173,8 +169,17 @@ func (xl xlObjects) GetObject(bucket, object string, startOffset int64, length i
 			readSize = length - totalBytesRead
 		}
 
+		// Get the checksums of the current part.
+		checkSums := make([]string, len(onlineDisks))
+		for index := range onlineDisks {
+			checkSums[index], _, err = metaArr[index].GetCheckSum(partName)
```
harshavardhana, line 175: change the function signature to not return an error instead rely on the returned values. Since there is only one error being returned 'errUnexpected' it is better handled here. 

```
checkSums[index], _ = metaAttr[index].GetCheckSum(partName) 
if checkSums[index] == "" {
     return toObjectError(errUnexpected, bucket, object)
}
```

erasure-readfile.go
```diff
@@ -277,25 +234,25 @@ func erasureReadFile(writer io.Writer, disks []StorageAPI, volume string, path s
...
 			var readDisks []StorageAPI
 			var err error
-			readDisks, nextIndex, err = getReadDisks(orderedDisks, nextIndex, eInfo.DataBlocks)
+			readDisks, nextIndex, err = getReadDisks(disks, nextIndex, dataBlocks)
 			if err != nil {
 				return bytesWritten, err
 			}
-			parallelRead(volume, path, readDisks, orderedDisks, enBlocks, blockOffset, curChunkSize, bitRotVerify)
-			if isSuccessDecodeBlocks(enBlocks, eInfo.DataBlocks) {
+			parallelRead(volume, path, readDisks, disks, enBlocks, blockOffset, curChunkSize, bitRotVerify)
```
harshavardhana, line 243: Add more comments here.

erasure-readfile.go
```diff
@@ -277,25 +234,25 @@ func erasureReadFile(writer io.Writer, disks []StorageAPI, volume string, path s
 			// readDisks - disks from which we need to read in parallel.
 			var readDisks []StorageAPI
 			var err error
-			readDisks, nextIndex, err = getReadDisks(orderedDisks, nextIndex, eInfo.DataBlocks)
+			readDisks, nextIndex, err = getReadDisks(disks, nextIndex, dataBlocks)
```
harshavardhana, line 238: Comments here.

xl-v1-object.go
krishnasrinivas, line 175: > change the function signature to not return an error instead rely on the returned values

that would be idiomatic `C` not idiomatic `Go` way of returning errors.

harshavardhana:
> that would be idiomatic C not idiomatic Go way of returning errors.

Agree if it is `os.Open()` or `rpc.Call` but not here.

harshavardhana:
Will merge this for now, will revisit the xl-v1-metadata.go functions again.

PR #2218, opened 2016-07-17
api: ListBuckets doesn't have a body, we should never read the body.

harshavardhana:
ListBuckets was incorrectly reading the body of the request, fix it.


krisis:
LGTM

PR #2219, opened 2016-07-17
api: Requests should be differentiated if possible based on http router.

harshavardhana:
In current master ListObjectsV2 was merged into ListObjectsHandler
which also implements V1 API as well.

Move the detection of ListObject types to its rightful place
in http router.


krisis:
LGTM

PR #2220, opened 2016-07-17
tree-walk: optimize tree walk such that leaf detection of entries is delayed till the entry is sent on the treeWalkResult channel.

krishnasrinivas:
On minio2 for 100k entries (uuids as filenames)
`time mc ls myminio/testbucket/|head`
without delayIsLeaf optimization : 90 secs
with delayIsLeaf optimization: 5 secs

### DIFF
--- a/tree-walk.go
+++ b/tree-walk.go
@@ -37,43 +37,108 @@ type treeWalkResult struct {
 	end   bool
 }
 
+// posix.ListDir returns entries with trailing "/" for directories. At the object layer
+// we need to remove this trailing "/" for objects and retain "/" for prefixes before
+// sorting because the trailing "/" can affect the sorting results for certain cases.
+// Ex. lets say entries = ["a-b/", "a/"] and both are objects.
+//     sorting with out trailing "/" = ["a", "a-b"]
+//     sorting with trailing "/"     = ["a-b/", "a/"]
+// Hence if entries[] does not have a case like the above example then isLeaf() check
+// can be delayed till the entry is pushed into the treeWalkResult channel.
+// delayIsLeafCheck() returns true if isLeaf can be delayed or false if
+// isLeaf should be done in listDir()
+func delayIsLeafCheck(entries []string) bool {
+	for i, entry := range entries {
+		if i == len(entries)-1 {
+			break
+		}
+		// If any byte in the "entry" string is less than '/' then the
+		// next "entry" should not contain '/' at the same same byte position.
+		for j := 0; j < len(entry); j++ {
+			if entry[j] < '/' {
+				if len(entries[i+1]) > j {
+					if entries[i+1][j] == '/' {
+						return false
+					}
+				}
+			}
+		}
+	}
+	return true
+}
+
+// Return entries that have prefix prefixEntry.
+// Note: input entries are expected to be sorted.
+func filterMatchingPrefix(entries []string, prefixEntry string) []string {
+	start := 0
+	end := len(entries)
+	for {
+		if start == end {
+			break
+		}
+		if strings.HasPrefix(entries[start], prefixEntry) {
+			break
+		}
+		start++
+	}
+	for {
+		if start == end {
+			break
+		}
+		if strings.HasPrefix(entries[end-1], prefixEntry) {
+			break
+		}
+		end--
+	}
+	return entries[start:end]
+}
+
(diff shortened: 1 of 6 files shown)


tree-walk.go
```diff
@@ -37,43 +37,104 @@ type treeWalkResult struct {
...
 			if err == nil {
-				// Skip the entries which do not match the prefixEntry.
-				for i, entry := range entries {
-					if !strings.HasPrefix(entry, prefixEntry) {
-						entries[i] = ""
-						continue
+				// Sort and filter out the entries that does not have prefix prefixEntry.
+				sort.Strings(entries)
+				start := 0
+				end := len(entries)
```
krishnasrinivas: @harshavardhana haven't tried to add comments for the code below as code explains the logic better than words. have added the gist `Sort and filter out the entries that does not have prefix prefixEntry.` above

harshavardhana:
Can you add test cases as well for various styles of input for 'delayIsLeafCheck()' ?

tree-walk.go
```diff
@@ -37,43 +37,104 @@ type treeWalkResult struct {
...
+				for {
+					if start == end {
+						break
 					}
```
harshavardhana: This code block has become bigger, we should move this a function which basically just skips all entries which do not have the prefix.

harshavardhana: > This code block has become bigger, we should move this a function which basically just skips all entries which do not have the prefix.

Along with a test case as well.

harshavardhana:
> On minio2 for 100k entries (uuids as filenames)
> time mc ls myminio/testbucket/|head
> without delayIsLeaf optimization : 90 secs
> with delayIsLeaf optimization: 5 secs

Can you do more testing ? basically provide a more average baseline here. Would be nice to know the overall benefit gradient with 
- 100k, 500k, 1mil, 2mil in same directory. 
- 100k, 500k, 1mil, 2mil spread out.

krishnasrinivas:
> > On minio2 for 100k entries (uuids as filenames)
> > time mc ls myminio/testbucket/|head
> > without delayIsLeaf optimization : 90 secs
> > with delayIsLeaf optimization: 5 secs
> 
> Can you do more testing ? basically provide a more average baseline here. Would be nice to know the overall benefit gradient with
> 
> 100k, 500k, 1mil, 2mil in same directory.
> 100k, 500k, 1mil, 2mil spread out.

From previous experience, the numbers can be extrapolated, but I will check it once again.

tree-walk.go
```diff
@@ -37,43 +37,108 @@ type treeWalkResult struct {
...
+// Hence if entries[] does not have a case like the above example then isLeaf() check
+// can be delayed till the entry is pushed into the treeWalkResult channel.
+// delayIsLeafCheck() returns true if isLeaf can be delayed or false if
+// isLeaf should be done in listDir()
+func delayIsLeafCheck(entries []string) bool {
+	for i, entry := range entries {
+		if i == len(entries)-1 {
+			break
+		}
+		// For each rune in the entry string, if any rune is less than '/'
```
krisis: `entry[j] < '/'` is performing a byte comparison and not a rune comparison. Please update the comment to reflect the same.

harshavardhana:
This patch can be accepted.. just that while we are at it we could finish it.

tree-walk.go
krishnasrinivas: ah, I thought indexing into a string fetches a rune and not byte, but you are right:

```
krishna@escape:/tmp$ cat a.go 
package main

import "fmt"

func main() {
    a := "⌘"
    for i := range a {
        fmt.Println(a[i])
    }
    fmt.Println("====")
    b := []byte(a)
    for i := range b {
        fmt.Println(b[i])
    }
}
krishna@escape:/tmp$ go run a.go
226
====
226
140
152
krishna@escape:/tmp$ 
```

krishnasrinivas:
> This patch can be accepted.. just that while we are at it we could finish it.

In that case you can take the patch in, I am confident that the numbers can be extrapolated. For majority of the cases, the only overhead with the current patch is delayIsLeafCheck which was quick on my laptop (2 secs for 10 mil entries)

I will do  performance testing for listing later in the week once I get other things off my list.

harshavardhana:
LGTM, Thanks :+1:

PR #2228, opened 2016-07-18
FS/XL: Return IncompleteBody{} error for short writes

### LINKED ISSUE #2221: fs/xl: PutObject should return IncompleteBody when request.Body fewer than specified bytes
## Expected behaviour

if request.Body has fewer bytes than the size provided in Content-length field in the request header, PutObject should fail with `IncompleteBody` error.
## Actual behaviour

fs: Object is not created and no error is returned.
xl: Object is created successfully.

Found while performing unit test at the xl/fs object layer. 
## Minio version

commit id `27a5b61`

krisis:
Fixes #2221

### DIFF
--- a/erasure-createfile.go
+++ b/erasure-createfile.go
@@ -28,7 +28,7 @@ import (
 // erasureCreateFile - writes an entire stream by erasure coding to
 // all the disks, writes also calculate individual block's checksum
 // for future bit-rot protection.
-func erasureCreateFile(disks []StorageAPI, volume, path string, reader io.Reader, blockSize int64, dataBlocks int, parityBlocks int, writeQuorum int) (size int64, checkSums []string, err error) {
+func erasureCreateFile(disks []StorageAPI, volume, path string, reader io.Reader, size int64, blockSize int64, dataBlocks int, parityBlocks int, writeQuorum int) (bytesWritten int64, checkSums []string, err error) {
 	// Allocated blockSized buffer for reading.
 	buf := make([]byte, blockSize)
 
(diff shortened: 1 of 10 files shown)


erasure-createfile.go
```diff
@@ -44,10 +44,15 @@ func erasureCreateFile(disks []StorageAPI, volume, path string, reader io.Reader
 			return 0, nil, rErr
 		}
 		if rErr == io.EOF {
+			// Should return IncompleteBody when reader contains
+			// fewer than specified bytes
+			if bytesWritten < size {
+				return bytesWritten, nil, io.ErrShortWrite
```
harshavardhana: why can't we handle this outside of 'erasureCreateFile' ? at the object layer?

fs-v1-multipart.go
```diff
@@ -321,23 +321,26 @@ func (fs fsObjects) PutObjectPart(bucket, object, uploadID string, partID int, s
...
+	bytesWritten := int64(0)
+	// Read the buffer till io.EOF and append the read data to the temporary file.
 	for {
-		n, err := io.ReadFull(limitDataReader, buf)
-		if err == io.EOF {
-			break
-		}
-		if err != nil && err != io.ErrUnexpectedEOF {
-			return "", toObjectErr(err, bucket, object)
+		n, rErr := limitDataReader.Read(buf)
```
harshavardhana: .Read is not efficient for us, it will lead to many more calls since Read also doesn't return full 'buf' bytes.  io.ReadFull was used for this purpose only. 

Also move this into separate function as 'fs-createfile.go'

harshavardhana: > .Read is not efficient for us, it will lead to many more calls since Read also doesn't return full 'buf' bytes. io.ReadFull was used for this purpose only.

Actually wait .Read is fine, my bad. 

> Also move this into separate function as 'fs-createfile.go'

Just that move this into separate file / function. use it also in PutObject()

erasure-createfile.go
krisis: @harshavardhana Nothing stops us from handling it at the object layer. I shall update the PR handling it at the object layer for XL.

krisis:
> you could delete the tmp file too.
> @krishnasrinivas so, there are other such places in xl-v1-object.go where we don't delete the temp object. Should I fix all those too in this PR?

krishnasrinivas:
> @krishnasrinivas so, there are other such places in xl-v1-object.go where we don't delete the temp object. Should I fix all those too in this PR?

oh! if there are a quite a few places where we leave tmp files without deleting then it's better to have a separate 'tmp file cleanup' PR and not mix with this PR

harshavardhana:
> @krishnasrinivas so, there are other such places in xl-v1-object.go where we don't delete the temp object. Should I fix all those too in this PR?
> oh! if there are a quite a few places where we leave tmp files without deleting then it's better to have a separate 'tmp file cleanup' PR and not mix with this PR

@krisis can you show where are these places where are not deleting tmp files?

PR #2229, opened 2016-07-18
erasureReadFile and erasureCreateFile testcases.

krishnasrinivas:


erasure-readfile_test.go
```diff
@@ -203,3 +207,149 @@ func TestErasureReadFile(t *testing.T) {
...
+		// 2nd block.
+		{blockSize, blockSize},
+		// Test cases for unusual offsets and lengths.
+		{blockSize - 1, 2},
+		{blockSize - 1, blockSize + 1},
+		{blockSize + 1, blockSize - 1},
+		{blockSize + 1, blockSize},
+		{blockSize + 1, blockSize + 1},
+		{blockSize*2 - 1, blockSize*3 + 1},
+	}
```
harshavardhana, line 346: Add a comment here.

erasure-readfile_test.go
```diff
@@ -203,3 +207,149 @@ func TestErasureReadFile(t *testing.T) {
...
+	for i, testCase := range testCases {
+		expected := data[testCase.offset:(testCase.offset + testCase.length)]
+		buf := &bytes.Buffer{}
+		size, err = erasureReadFile(buf, disks, "testbucket", "testobject", testCase.offset, testCase.length, length, blockSize, dataBlocks, parityBlocks, checkSums)
+		if err != nil {
+			t.Error(err)
+			continue
+		}
+		got := buf.Bytes()
+		if bytes.Compare(expected, got) != 0 {
```
harshavardhana: use 'bytes.Equal' instead.

erasure-readfile.go
```diff
@@ -262,14 +262,17 @@ func erasureReadFile(writer io.Writer, disks []StorageAPI, volume string, path s
 		}
 
 		var outSize, outOffset int64
+
+		// Total data to be read.
+		outSize = blockSize
```
harshavardhana, line 267: We need to think about moving this logic into a separate function or something. That way we can test thoroughly this skipping logic.

harshavardhana, line 267: Not needed in this patch.

krishnasrinivas, line 267: > We need to think about moving this logic into a separate function or something. That way we can test thoroughly this skipping logic.

true, will check on this. Also there are so many variables in erasureReadFile() it's difficult to read, will see how it can be improved.

PR #2238, opened 2016-07-19
Preallocate space before append to avoid disk full error

### LINKED ISSUE #2194: [distributed] fallocate/fadvise to provide hint for I/O
Under `CompleteMultipartUpload` we should use fallocate/fadvise when available to allocate disk space.

vadmeste:
Fixes #2194 

Please do not merge immediately under considered really stable.

### DIFF
--- a/fallocate.go
+++ b/fallocate.go
@@ -1,0 +1,25 @@
+// +build !linux
+
+/*
+ * Minio Cloud Storage, (C) 2016 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package main
+
+// Fallocate is not POSIX and not supported under Windows
+// Always return successful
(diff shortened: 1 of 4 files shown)


krishnasrinivas:
we do AppendFile in 128K chunks so Fallocate at posix won't be helpful.

harshavardhana:
> we do AppendFile in 128K chunks so Fallocate at posix won't be helpful.

Perhaps if posix knows what will be the final size..

vadmeste:
> we do AppendFile in 128K chunks so Fallocate at posix won't be helpful.

yeah, I noticed that.. actually I suggested to add a new func PrepareFile() in storage-interface for this purpose, but the way why it is 128K, couldn't we increase it ?

harshavardhana:
> yeah, I noticed that.. actually I suggested to add a new func PrepareFile() in storage-interface for this purpose, but the way why it is 128K, couldn't we increase it ?

It something coming from erasure itself, because of the block size i.e 10MB.

vadmeste:
Anyway, as a general rule it is good to use fallocate to make disks less fragmented, fallocate() is really fast

krishnasrinivas:
> Anyway, as a general rule it is good to use fallocate to make disks less fragmented, fallocate() is really fast

if you write() 128k in one go the FS implementations make sure that it's not fragmented. so in this situation fallocate will not help. If you write in smaller chunks say 4K with time gap between the writes then it will lead to fragmentation (in that situation fallocate will help)

harshavardhana:
Can this patch be rebased? @vadmeste

krishnasrinivas:
> Can this patch be rebased? @vadmeste

@harshavardhana this commit is not adding value, check the discussion above

harshavardhana:
> @harshavardhana this commit is not adding value, check the discussion above

Yes ended up reading a lot about fallocate, it is not useful for us.  Closing this PR for now.

PR #2243, opened 2016-07-20
XL: Do not rely on getLoadBalancedQuorumDisks for NS consistency.

### LINKED ISSUE #2242: XL: Namespace operations like isBucketExists fail non-deterministically
## Expected behaviour

isBucketExists/isObject like functions on the namespace  should return same error (incl. nil) when invoked with different (number of) disks failing in a XL setup.
## Actual behaviour

isObject fails with ObjectNotFound at times and returns successfully other times in a non-deterministic manner. This becomes difficult in the context of unit-tests.

```
OK: 96 passed
--- FAIL: TestXLDeleteObjectDiskNotFound (0.03s)
    xl-v1-object_test.go:134: Object not found: bucket#object
```

```
OK: 96 passed
--- FAIL: TestPutObjectPartDiskNotFound (0.03s)
    object-api-multipart_test.go:181: Test XL: expected error Storage resources are insufficient for the write operation., got Invalid upload id 8b4ea0a4-9370-4846-b621-19999e51e2e2 instead.
FAIL
```
- Run unit-tests on latest master `cef26fd`
  ## Minio version
- master branch `cef26fd`

harshavardhana:
The reason is any function relying on `getLoadBalancedQuorumDisks`
cannot possibly have an idempotent behavior.

The problem comes from given a set of N disks returning just a
shuffled N/2 disks.  In case of a scenario where we have N/2
number of failed disks, the returned value of `getLoadBalancedQuorumDisks`
is not equal to the same failed disks so essentially calls using such
disks might succeed or fail randomly at different intervals in time.

This proposal change is we move to `getLoadBalancedDisks()`
and use the shuffled N disks as a whole. Since most of the time we might
hit a good disk since we are not reducing our solution space. This
also provides consistent behavior for all the functions which rely
on shuffled disks.

Fixes #2242

### DIFF
--- a/xl-v1-bucket.go
+++ b/xl-v1-bucket.go
@@ -30,6 +30,10 @@ func (xl xlObjects) MakeBucket(bucket string) error {
 	if !IsValidBucketName(bucket) {
 		return BucketNameInvalid{Bucket: bucket}
 	}
+	// Verify if bucket is found.
+	if xl.isBucketExist(bucket) {
+		return toObjectErr(errVolumeExists, bucket)
+	}
 
 	nsMutex.Lock(bucket, "")
 	defer nsMutex.Unlock(bucket, "")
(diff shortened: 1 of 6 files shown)


xl-v1-bucket.go
```diff
@@ -220,13 +240,15 @@ func (xl xlObjects) DeleteBucket(bucket string) error {
 	if !IsValidBucketName(bucket) {
 		return BucketNameInvalid{Bucket: bucket}
 	}
+	// Verify if bucket is found.
+	if !xl.isBucketExist(bucket) {
```
krishnasrinivas, line 244: why is this check needed? because the results of DeleteVol() should return the error.

xl-v1-bucket.go
```diff
@@ -220,13 +240,15 @@ func (xl xlObjects) DeleteBucket(bucket string) error {
 	if !IsValidBucketName(bucket) {
 		return BucketNameInvalid{Bucket: bucket}
 	}
+	// Verify if bucket is found.
```
krishnasrinivas, line 243: comment is not adding value

Issue #2244, opened 2016-07-20
api - content-encoding missing


brackxm:
## Expected behaviour

getObject returns the Content-Encoding
## Actual behaviour

getObject does not return the Content-Encoding
## Steps to reproduce the behaviour

putObject with Content-Encoding
then getObject
## Minio version

Version: 2016-07-13T21:46:05Z
Release-Tag: RELEASE.2016-07-13T21-46-05Z
Commit-ID: 3f27734c22212f224037a223439a425e6d2b653a
## System information

aws java sdk 1.11.5

krishnasrinivas:
@brackxm are you using minio FS or XL? FS does not store content-type or content-encoding in the backend.

brackxm:
The server was started like in the readme/quickstart, with a directory argument.
Assuming this is FS, how do i start an XL server?

brackxm:
same behaviour with the erasure coded backend (starting the server with 6 dirs).
Maybe that is XL, as there are xl.json files now.

krishnasrinivas:
> same behaviour with the erasure coded backend (starting the server with 6 dirs).
> Maybe that is XL, as there are xl.json files now.

That is right

harshavardhana:
Fixed by #2245

PR #2245, opened 2016-07-20
api: Set content-encoding properly if set.

harshavardhana:
Additionally don't set content-type if not present, golang http
server automaticaly handles this and sets it automatically.


harshavardhana:
Still not done.. adding a test case.

harshavardhana:
Fixes #2244

PR #2247, opened 2016-07-20
Serving http and https on the same port using a customized server

vadmeste:


server-mux.go
```diff
@@ -0,0 +1,151 @@
...
+	listener, err := net.Listen("tcp", m.Server.Addr)
+	mux, err := NewMuxListener(listener, mustGetCertFile(), mustGetKeyFile())
+	if err != nil {
+		fatalIf(err, "Failed to start minio server.")
+	}
+	err = http.Serve(mux,
+		http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			if r.TLS == nil {
+				u := url.URL{
+					Scheme:   "https",
```
harshavardhana, line 178: Shouldn't this be 'http' with r.TLS == nil ?

server-mux.go
```diff
@@ -0,0 +1,151 @@
...
+type MuxListener struct {
+	net.Listener
+	config tls.Config
+}
+
+func NewMuxListener(listener net.Listener, certPath, keyPath string) (MuxListener, error) {
+	var err error
+	config := &tls.Config{}
+	if config.NextProtos == nil {
+		config.NextProtos = []string{"http/1.1"}
```
harshavardhana: Have you thought about `http/2.0` support on what would this do?

server-mux.go
```diff
@@ -0,0 +1,151 @@
...
+
+type MuxServer struct {
+	http.Server
+}
+
+func (m *MuxServer) ListenAndServeTLS(certFile, keyFile string) error {
+	listener, err := net.Listen("tcp", m.Server.Addr)
+	mux, err := NewMuxListener(listener, mustGetCertFile(), mustGetKeyFile())
+	if err != nil {
+		fatalIf(err, "Failed to start minio server.")
```
harshavardhana: We should return a error here.. since this function is required to.

server-mux.go
```diff
@@ -0,0 +1,151 @@
...
+	net.Conn
+	buffer  []byte
+	reset   bool
+	pointer int
+	peekLen int
+}
+
+func NewMuxConn(c net.Conn) *MuxConn {
+	maxLen := 0
+	for _, m := range defaultHTTPMethods {
```
harshavardhana: Move this into a function of its own.

server-mux.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+func (c *MuxConn) peekProtocol() (string, error) {
+	n, err := c.Conn.Read(c.buffer)
+	if n == 0 || err != nil {
+		return "", err
+	}
+	c.reset = true
+	for _, m := range defaultHTTPMethods {
+		if strings.HasPrefix(string(c.buffer), m+" ") {
```
harshavardhana: Don't understand the logic here.. you might have to write a proper comment talking about the HTTP headers themselves.

harshavardhana: > Don't understand the logic here.. you might have to write a proper comment talking about the HTTP headers themselves.

Ah the TLS part i see how it is doing, you will be able to peek only HTTP not HTTPS.

harshavardhana:
This is a good work.. nice start for us. Potentially can be done as a package also.

server-mux.go
vadmeste, line 178: No, we are constructing a new redirect URL which will be https://...:9000/

server-mux.go
vadmeste: :+1:

server-mux.go
```diff
@@ -0,0 +1,151 @@
...
+		}
+		return c.pointer, nil
+	}
+	n, err := c.Conn.Read(b)
+	return n, err
+}
+
+type MuxListener struct {
+	net.Listener
+	config tls.Config
```
harshavardhana: you might have to use `config *tls.Config`

vadmeste: why ?

harshavardhana:
> why ?

Check the travis error.

server-mux.go
vadmeste: It seems to me it won't matter a pointer or not in this case

harshavardhana:
Can you test the HTTP 2.0 support with this change?  - https://http2.github.io/

harshavardhana:
also are there any performance implications?

vadmeste:
> Can you test the HTTP 2.0 support with this change? - https://http2.github.io/
- HTTP 2.0 is already activated by default in the server (since Go1.6).
- I added  http2 (h2) support to TLS.
- Go still doesn't support plain http2 (h2c) but it still won't make difference when it will be activated
- Sometimes, a client sends a direct HTTP2 request (PRI) when it assumes that the server is already HTTP2. That should be supported too (but not tested)

I used nghttp as a tool to test Minio with http2+tls

vadmeste:
> also are there any performance implications?

When the connection is initiated between the client and Minio, the code has to peek the protocol used in the connection. I can use Patricia algorithm to accelerate search (I remember you used that once), but when the connection is established, the only overhead is a boolean comparaison (which must be one cpu cycle operation). Well to be sure about what I said, I will use jmeter to test performance.

vadmeste:
> one cpu cycle

sorry, I mean one cpu instruction

Makefile
```diff
@@ -61,6 +61,7 @@ checks:
 	@(env bash $(PWD)/buildscripts/checkgopath.sh)
 
 getdeps: checks
+	@go get -u github.com/hashicorp/go-immutable-radix
```
harshavardhana: do you sill need this?

server-mux.go
```diff
@@ -0,0 +1,189 @@
...
+	"DELETE",
+	"TRACE",
+	"CONNECT",
+}
+
+// MuxConn - implements a Read() which streams twice the firs bytes from
+// the incoming connection, to help peeking protocol
+type MuxConn struct {
+	net.Conn
+	buffer  []byte
```
harshavardhana: Move this into a struct of its own for simplicity. 

```
var buf struct {
    data []byte
    offset int 
    unRead bool
} 
```

server-mux.go
```diff
@@ -0,0 +1,189 @@
...
+		}
+	}
+	return "tls", nil
+}
+
+// Read - streams the MuxConn buffer when reset flag is activated, otherwise
+// streams from the incoming network connection
+func (c *MuxConn) Read(b []byte) (int, error) {
+	if c.reset {
+		copy(b, c.buffer[c.pointer:])
```
harshavardhana: Make sure to serve the entire 'len(b)' to avoid second Read() here, since you would have returned a smaller buffer that requested.

server-mux.go
```diff
@@ -0,0 +1,214 @@
...
+		}))
+	return err
+}
+
+// ListenAndServe - Same as the http.Server version
+func (m *MuxServer) ListenAndServe() error {
+	return m.Server.ListenAndServe()
+}
+
+func longestWord(strings []string) int {
```
0xmohit, line 200: What is the use of the parameter being passed here?  AFAICT, it's not being used.

Even if the parameter `strings` were to be used inside the function, the choice of name isn't great as it collides with that of a package in the standard library.

harshavardhana, line 200: Good catch its a bug @vadmeste wanted to perhaps pass the HTTP methods.

vadmeste, line 200: > What is the use of the parameter being passed here? AFAICT, it's not being used.
> 
> Even if the parameter strings were to be used inside the function, the choice of name isn't great as it collides with that of a package in the standard library.

Thanks. I will fix this @0xmohit. By the way, I didn't find any API similar to longestWord, could you point it to me?

0xmohit, line 200: > By the way, I didn't find any API similar to longestWord, could you point it to me?

@vadmeste I'm not aware of any such API in the Go standard lib.

Just out of curiosity: Instead of cooking up own implementation for multiplexing, couldn't we use an alternative (such as https://github.com/soheilhy/cmux).

harshavardhana, line 200: > Just out of curiosity: Instead of cooking up own implementation for multiplexing, couldn't we use an alternative (such as https://github.com/soheilhy/cmux).

No it was on purpose to not use it the cmux API's are not convenient and we don't need a complex implementation like what they have.

PR #2254, opened 2016-07-21
Intrumentation:  Leak test framework

hackintoshrao:
- The idea is to make this part of every unit test, integration test, benchamark to detect leaks. Eventually most of the leaks will be caught when this is made part of tests. 
- here is one way to test/use the leak detector: 

Go to any unit test and add this one line and then run the test, `go test -run=TestGetObject`, you'll find the leaked routine and verify them later with a defer statement. 

```
func TestGetObject(t *testing.T) {
    defer DetectTestLeak(t)()
    ExecObjectLayerTest(t, testGetObject)
}

```

Same method can by employed to enable this in benchmarks. 

Here is how to add this in suite tests. 

```
// API suite container common to both FS and XL.
type TestSuiteCommon struct {
        findLeak leakDetect

        serverType string
        testServer TestServer
        endPoint   string
        accessKey  string
        secretKey  string
}

// Init and run test on FS backend.
var _ = Suite(&TestSuiteCommon{serverType: "FS"})

// Init and run test on XL backend.
// var _ = Suite(&TestSuiteCommon{serverType: "XL"})

// Setting up the test suite.
// Starting the Test server with temporary FS backend.
func (s *TestSuiteCommon) SetUpSuite(c *C) {
        s.testServer = StartTestServer(c, s.serverType)
        s.endPoint = s.testServer.Server.URL
        s.accessKey = s.testServer.AccessKey
        s.secretKey = s.testServer.SecretKey

        s.findLeak = NewLeakDetect()
}

// Called implicitly by "gopkg.in/check.v1" after all tests are run.
func (s *TestSuiteCommon) TearDownSuite(c *C) {
        s.testServer.Stop()
        s.findLeak.DetectLeak(c)
}
```

There are several leaks which are detected. Once this is reviewed and merged will integrate these and make a PR and later those leans has to be fixed.

### DIFF
--- a/leak-detect_test.go
+++ b/leak-detect_test.go
@@ -1,0 +1,154 @@
+// Copyright 2013 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package leaktest provides tools to detect leaked goroutines in tests.
+// To use it, call "defer leaktest.AfterTest(t)()" at the beginning of each
+// test that may use goroutines.
+package main
+
+import (
+	"runtime"
+	"sort"
+	"strings"
+	"time"
+)
+
+const (
+	// deadline (in seconds) upto which the go routine leak detection has to be retried.
+	leakDetectDeadline = 5
+	// pause time (in milliseconds) between each snapshot at the end of the go routine leak detection.
+	leakDetectPauseTimeMs = 50
+)
+
+// LeakDetect - type with  methods for go routine leak detection.
+type LeakDetect struct {
+	relevantRoutines map[string]bool
+}
+
+// NewLeakDetect - Initialize a LeakDetector with the snapshot of relevant Go routines.
+func NewLeakDetect() LeakDetect {
+	snapshot := LeakDetect{
+		relevantRoutines: make(map[string]bool),
+	}
+	for _, g := range pickRelevantGoroutines() {
+		snapshot.relevantRoutines[g] = true
+	}
+	return snapshot
+}
+
+// CompareCurrentSnapshot - Comapres the initial relevant stack trace with the current one (during the time of invocation).
+func (initialSnapShot LeakDetect) CompareCurrentSnapshot() []string {
+	var stackDiff []string
+	for _, g := range pickRelevantGoroutines() {
+		// Identify the Go routines those were not present in the initial snapshot.
+		// In other words a stack diff.
+		if !initialSnapShot.relevantRoutines[g] {
+			stackDiff = append(stackDiff, g)
+		}
+	}
+	return stackDiff
+}
+
+// DetectLeak - Creates a snapshot of runtiem stack and compares it with  the initial stack snapshot.
+func (initialSnapShot LeakDetect) DetectLeak(t TestErrHandler) {
+	if t.Failed() {
+		return
+	}
+	// Loop, waiting for goroutines to shut down.
+	// Wait up to 5 seconds, but finish as quickly as possible.
+	deadline := time.Now().Add(leakDetectDeadline * time.Second)
+	for {
+		// get sack snapshot of relevant go routines.
+		leaked := initialSnapShot.CompareCurrentSnapshot()
+		// current stack snapshot matches the initial one, no leaks, return.
+		if len(leaked) == 0 {
+			return
+		}
+		// wait a test again will deadline.
+		if time.Now().Before(deadline) {
+			time.Sleep(leakDetectPauseTimeMs * time.Millisecond)
+			continue
+		}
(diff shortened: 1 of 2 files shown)


hackintoshrao:
Its based on the concept off `StackDiff` before the test and after the test.  A snapshot of relevant go routines are taken when tests begin and its compared with the stack snapshot of the relevant go routines after the test to detect any pending routines.

leak-detect_test.go
```diff
@@ -0,0 +1,132 @@
...
+package main
+
+import (
+	"runtime"
+	"sort"
+	"strings"
+	"time"
+)
+
+const (
```
harshavardhana, line 17: Needs comments..

leak-detect_test.go
```diff
@@ -0,0 +1,132 @@
+// Copyright 2013 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
```
harshavardhana, line 3: if this is a golang package why can't we directly reference it and vendorize it?

hackintoshrao, line 3: As the current implementation doesn't maintain state info across calls these won't be useful in case of integration test scenarios. So made changes so that the state information can be optionally carried across.

leak-detect_test.go
hackintoshrao, line 17: done.

leak-detect_test.go
```diff
@@ -0,0 +1,134 @@
...
+		if len(sl) != 2 {
+			continue
+		}
+		stack := strings.TrimSpace(sl[1])
+		// ignore the testing go routine.
+		// since the tests will be invoking the leaktest it would contain the test go routine.
+		if strings.HasPrefix(stack, "testing.RunTests") {
+			continue
+		}
+		// Ingnore the following go routines.
```
harshavardhana: 's/Ingnore/Ignore'

leak-detect_test.go
```diff
@@ -0,0 +1,134 @@
...
+		// testing and run time go routines should be ignored, only the application generated go routines should be taken into account.
+		if stack == "" ||
+			// Below are the stacks ignored by the upstream leaktest code.
+			strings.Contains(stack, "testing.Main(") ||
```
harshavardhana: Follow a different style here.. move 

```
var ignoredStackFns = []string{
    "testing.Main(",
    ...
    ...
    // Add more stack functions to ignore.
}

func isIgnoredStackFn(stack string) (ok bool) {
    ok = true
    for _, stackFn := range ignoredStackFns {
         if !strings.Contains(stack, stackFn) {
                  ok = false
                  continue
         }
         break
    }
    return ok
}
```

leak-detect_test.go
```diff
@@ -0,0 +1,134 @@
...
+	buf := make([]byte, 2<<20)
+	buf = buf[:runtime.Stack(buf, true)]
+	// runtime stack of go routines will be listed with 2 blank spaces between each of them, so split on "\n\n" .
+	for _, g := range strings.Split(string(buf), "\n\n") {
+		// Again split on a new line, the first line of the second half contaisn the info about the go routine.
+		sl := strings.SplitN(g, "\n", 2)
+		if len(sl) != 2 {
+			continue
+		}
+		stack := strings.TrimSpace(sl[1])
```
harshavardhana, line 139: here is 'stack' a function? or a stack itself?

leak-detect_test.go
```diff
@@ -0,0 +1,134 @@
...
+// goroutines leaked.
+// Usage: `defer DetectTestLeak(t)()` in beginning line of benchmarks or unit tests.
+func DetectTestLeak(t TestErrHandler) func() {
+	initialStackSnapShot := NewLeakDetect()
+	return func() {
+		initialStackSnapShot.DetectLeak(t)
+	}
+}
+
+// interestingGoroutines returns all goroutines we care about for the purpose
```
harshavardhana: this doesn't reflect the function name

leak-detect_test.go
```diff
@@ -0,0 +1,134 @@
...
+// of leak checking. It excludes testing or runtime ones.
+func snapShotRelevantGoroutines() (gs []string) {
```
harshavardhana: how about 'pickRelevantGoroutines()' ?

leak-detect_test.go
hackintoshrao: This is neat. Thanks.

leak-detect_test.go
hackintoshrao, line 139: Its the stack entry for a function.

leak-detect_test.go
hackintoshrao: Ya, that sounds better.

PR #2259, opened 2016-07-21
optimize memory allocation during erasure-read by using temporary buffer pool.

krishnasrinivas:
With the change the buffer needed during GetObject by erasureReadFile is allocated only once.

### DIFF
--- a/erasure-utils.go
+++ b/erasure-utils.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"hash"
 	"io"
+	"sync"
 
 	"github.com/klauspost/reedsolomon"
 	"github.com/minio/blake2b-simd"
@@ -234,4 +235,49 @@ func copyBuffer(writer io.Writer, disk StorageAPI, volume string, path string, b
 
 	// Success.
 	return nil
+}
+
+// Error returned when there are no more free buffers in the scratch pool.
+// Normally this error should never be returned as scratch pool should always be
+// initialized with dataBlocks+parityBlocks number of buffers.
+var errScratchNoBuf = errors.New("no free buffer")
+
+// Temporary pool of buffers that can be used for scratch space.
+type scratchPool struct {
+	buf  [][]byte
+	used []bool
+	sync.Mutex
+}
+
+// Returns an unused buffer.
+func (s *scratchPool) get() (buf []byte, err error) {
+	s.Lock()
+	defer s.Unlock()
+	for i := 0; i < len(s.used); i++ {
+		if !s.used[i] {
+			s.used[i] = true
+			return s.buf[i], nil
+		}
+	}
+	return nil, errScratchNoBuf
+}
+
+// Marks all buffers as unused.
+func (s *scratchPool) reset() {
+	s.Lock()
+	defer s.Unlock()
+	for i := 0; i < len(s.used); i++ {
+		s.used[i] = false
+		s.buf[i] = s.buf[i][:0]
+	}
+}
+
+// Returns new scratch pool with n number of buffers each with size specified as arguemnt.
+func newScratchPool(size int64, n int) *scratchPool {
+	used := make([]bool, n)
+	buf := make([][]byte, n)
+	for i := 0; i < n; i++ {
+		buf[i] = make([]byte, size)
+	}
+	return &scratchPool{buf, used, sync.Mutex{}}
 }
--- a/erasure-utils.go
+++ b/erasure-utils.go
@@ -21,7 +21,6 @@ import (
 	"errors"
 	"hash"
 	"io"
-	"sync"
 
 	"github.com/klauspost/reedsolomon"
 	"github.com/minio/blake2b-simd"
@@ -235,49 +234,4 @@ func copyBuffer(writer io.Writer, disk StorageAPI, volume string, path string, b
 
 	// Success.
 	return nil
-}
-
-// Error returned when there are no more free buffers in the scratch pool.
-// Normally this error should never be returned as scratch pool should always be
-// initialized with dataBlocks+parityBlocks number of buffers.
-var errScratchNoBuf = errors.New("no free buffer")
-
-// Temporary pool of buffers that can be used for scratch space.
-type scratchPool struct {
-	buf  [][]byte
-	used []bool
-	sync.Mutex
-}
-
-// Returns an unused buffer.
-func (s *scratchPool) get() (buf []byte, err error) {
-	s.Lock()
-	defer s.Unlock()
-	for i := 0; i < len(s.used); i++ {
-		if !s.used[i] {
-			s.used[i] = true
-			return s.buf[i], nil
-		}
-	}
-	return nil, errScratchNoBuf
-}
-
-// Marks all buffers as unused.
-func (s *scratchPool) reset() {
-	s.Lock()
-	defer s.Unlock()
-	for i := 0; i < len(s.used); i++ {
-		s.used[i] = false
-		s.buf[i] = s.buf[i][:0]
-	}
-}
-
-// Returns new scratch pool with n number of buffers each with size specified as arguemnt.
-func newScratchPool(size int64, n int) *scratchPool {
-	used := make([]bool, n)
-	buf := make([][]byte, n)
-	for i := 0; i < n; i++ {
-		buf[i] = make([]byte, size)
-	}
-	return &scratchPool{buf, used, sync.Mutex{}}
 }
--- a/erasure-readfile.go
+++ b/erasure-readfile.go
@@ -17,7 +17,6 @@
 package main
 
 import (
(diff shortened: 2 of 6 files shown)


erasure-utils.go
```diff
@@ -235,3 +236,40 @@ func copyBuffer(writer io.Writer, disk StorageAPI, volume string, path string, b
 	// Success.
 	return nil
 }
+
+var errScratchNoBuf = errors.New("no free buffer")
```
harshavardhana: why not use 'sync.Pool'

erasure-utils.go
```diff
@@ -235,3 +236,48 @@ func copyBuffer(writer io.Writer, disk StorageAPI, volume string, path string, b
 	// Success.
 	return nil
 }
+
+// Error returned when there are no more free buffers in the scratch pool.
+// Normally this error should never be returned as scratch pool should always be
```
harshavardhana: I am not entirely comfortable with this approach, it creates complicated code base. 

We should use 'sync.Pool' if necessary but previously using that never really helped entirely.  

Code as below is not well tested and i would avoid adding these type of primitives in the codebase for simplicity sake.

harshavardhana:
> With the change the buffer needed during GetObject by erasureReadFile is allocated only once.

What optimization are we talking about here? why isn't `sync.Pool` preferred? at GetObject(). Need more evidence with varying workload.   If there are no real benefits we should avoid using pool primitives. 

Creating our own pool primitive is not safe and can lead to bugs.  Along with that we haven't written a unit test for that either we should keep 'WIP' until that is done.  Something new like also disruptive in the code flow would require clean unit tests.

krishnasrinivas:
> What optimization are we talking about here? why isn't sync.Pool preferred? at GetObject(). Need more evidence with varying workload. If there are no real benefits we should avoid using pool primitives.

correct, looking at sync.Pool

hackintoshrao:
When 500 serial `mc cat` run on a 1Gb file the code in this branch completed in 18s whereas the master code took 36s, there were no data sanctity issues, just an observation.

krishnasrinivas:
> When 500 serial mc cat run on a 1Gb file the code in this branch completed in 18s whereas the master code took 36s, there were data sanctity issues, just an observation.

how was the data sanctity issue discovered? also minio-cache was disabled right?

hackintoshrao:
Cache was disabled, I check for md5sum after mc-cat in the script

harshavardhana:
Is there a setup that this can be looked at @krishnasrinivas @hackintoshrao ?  do you see errors with minio master as well without this change?

hackintoshrao:
@harshavardhana , the setup exists on Minio3.

harshavardhana:
> @harshavardhana , the setup exists on Minio3.

Do you see the problem on 'master' as well?

harshavardhana:
@hackintoshrao confirmed this was a typo there are no data sanctity issues @krishnasrinivas

krishnasrinivas:
the following benchmark test showed improvement for the optimized code:
with optimization time taken for erasureReadFile: 4-6 secs consistently
without optimization: 6-10 secs
memory wise HeapSys and HeapAlloc remained same for both

will test in an actual setup to see if it's worth using the pool (as `premature optimization is root of all evil` )

```
func BenchmarkErasureReadFile(b *testing.B) {
        // Initialize environment needed for the test.
        dataBlocks := 7
        parityBlocks := 7
        blockSize := int64(blockSizeV1)
        setup, err := newErasureTestSetup(dataBlocks, parityBlocks, blockSize)
        if err != nil {
                b.Error(err)
                return
        }
        defer setup.Remove()

        disks := setup.disks
        length := int64(5242880000)
        reader, err := os.Open("/home/krishna/5g-non0")
        if err != nil {
               b.Fatal(err)
        }
        // Create a test file to read from.
        size, checkSums, err := erasureCreateFile(disks, "testbucket", "testobject", reader, blockSize, dataBlocks, parityBlocks, dataBlocks+1)
        if err != nil {
                b.Fatal(err)
        }
        if size != length {
                b.Errorf("erasureCreateFile returned %d, expected %d", size, length)
        }

        // create scratch pool buffer which will be used by erasureReadFile as scratch space for
        // reading from disks and erasure decoding.
        chunkSize := getChunkSize(blockSize, dataBlocks)
        // scratch := newScratchPool(chunkSize, len(disks))
        pool := &sync.Pool{
                New: func() interface{} {
                        return make([]byte, chunkSize)
                },
        }
        b.ResetTimer()
        for i := 0; i < b.N; i++ {
                size, err = erasureReadFile(ioutil.Discard, disks, "testbucket", "testobject", 0, length, length, blockSize, dataBlocks, parityBlocks, checkSums, pool)
                if err != nil {
                        b.Error(err)
                }
        }
        var m runtime.MemStats
        runtime.ReadMemStats(&m)
        b.Logf("%d, %d, %d\n", m.HeapSys, m.HeapAlloc, m.HeapIdle)

}
```

harshavardhana:
> the following benchmark test showed improvement for the optimized code:
> with optimization time taken for erasureReadFile: 4-6 secs consistently
> without optimization: 6-10 secs
> memory wise HeapSys and HeapAlloc remained same for both
> 
> will test in an actual setup to see if it's worth using the pool (as premature optimization is root of all evil )

Agreed cumulatively if there is a great improvement i.e as a function of file size then it makes sense to implement and merge this.  The inherent complexity here arises because of managing buffers in this manner.

It is a tricky part of the code to keep it clean / documented for the future i would even suggest we do this activity after we have reached good coverage with our unit tests and our libraries are updated as well with unit tests/functional tests.

harshavardhana:
@hackintoshrao posted some numbers assuming this is a serial 'mc cat' with 500 such operations - this is a real boost and quite nicely reduces the time spent. 

```
real 18m23.736s (with Krishna's changes)
real 34m55.389s with current master
```

Now we need to make the change to use 'sync.Pool' and test again.  Additionally we should also get a memory profiler output to see what is the overall memory usage drop as well. 

// cc @krishnasrinivas

krishnasrinivas:
@harshavardhana @hackintoshrao with my tests the numbers were same with and without sync.Pool how ever it was different with scratchPool

100 parallel download on 5GB file:
without optimization 5min 700% CPU
with sync.Pool         5min  700% CPU
with scratchPool      4min  600% CPU (so there is 20-25% improvement for time)

for 10 serial downloads the numbers were same for all 3 cases. (in your case there was 50% improvement with scratchPool) but I had not done 3> drop_caches

krishnasrinivas:
will check again with drop_caches

harshavardhana:
> @harshavardhana @hackintoshrao with my tests the numbers were same with and without sync.Pool how ever it was different with scratchPool
> 
> 100 parallel download on 5GB file:
> without optimization 5min 700% CPU
> with sync.Pool 5min 700% CPU
> with scratchPool 4min 600% CPU (so there is 20-25% improvement for time)
> 
> for 10 serial downloads the numbers were same for all 3 cases. (in your case there was 50% improvement with scratchPool) but I had not done 3> drop_caches

👍  - these are serious benefits, we should definitely get this in.

krishnasrinivas:
for serial download,
master branch: 190% CPU
sync.Pool: 190% CPU
scratchPool: 130% CPU

harshavardhana:
> for serial download,
> master branch: 190% CPU
> sync.Pool: 190% CPU
> scratchPool: 130% CPU

Can you post your sync.Pool implementation?

erasure-utils.go
```diff
@@ -235,3 +236,48 @@ func copyBuffer(writer io.Writer, disk StorageAPI, volume string, path string, b
...
+// initialized with dataBlocks+parityBlocks number of buffers.
+var errScratchNoBuf = errors.New("no free buffer")
+
+// Temporary pool of buffers that can be used for scratch space.
+type scratchPool struct {
```
harshavardhana: Rename this as 'bufferPool'  and make this re-purposable package.

krishnasrinivas:
@harshavardhana https://github.com/krishnasrinivas/minio/tree/bytepool bpool's performance was not impressive:
100 parallel download on 5GB file: 700% CPU
5G serial download: 210% CPU

i.e no better than sync.Pool - Maybe because of the channel overhead.

scratchPool performs the best.

harshavardhana:
> @harshavardhana https://github.com/krishnasrinivas/minio/tree/bytepool bpool's performance was not impressive:
> 100 parallel download on 5GB file: 700% CPU
> 5G serial download: 210% CPU
> 
> i.e no better than sync.Pool - Maybe because of the channel overhead.
> 
> scratchPool performs the best.

Then let's move the scratchPool as re-purposable package inside `pkg/bufpool/` and then let's merge this.

harshavardhana:
LGTM, Thanks :+1:

PR #2260, opened 2016-07-21
xl-v1-metadata.go, xl-v1-multipart-common.go

krishnasrinivas:
remaining code will be covered when we do testing with simulation of quorum error.
Need to implement tests by simulating disk failures similar to how it is done
in erasure-readfile_test.go and erasure-createfile_test.go This should fix coverage
in other files too which has code path for quorum error.

### DIFF
--- a/fs-v1-multipart-common.go
+++ b/fs-v1-multipart-common.go
@@ -19,7 +19,6 @@ package main
 import (
 	"encoding/json"
 	"path"
-	"strings"
 	"time"
 )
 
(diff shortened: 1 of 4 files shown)


harshavardhana:
> remaining code will be covered when we do testing with simulation of quorum error.
> Need to implement tests by simulating disk failures similar to how it is done
> in erasure-readfile_test.go and erasure-createfile_test.go This should fix coverage
> in other files too which has code path for quorum error.

Where are the unit tests in this patch? i see only handler level test.

krishnasrinivas:
> Where are the unit tests in this patch? i see only handler level test.

I changed s/unit-tests/tests/ in the commit message

PR #2271, opened 2016-07-23
api: Implement bucket notification.

harshavardhana:
Supports two different queue types
- AMQP
- ElasticSearch.


harshavardhana:
Supports redis as well now.

PR #2276, opened 2016-07-24
XL/GetObject: When disk is not available, checksum should be empty.

harshavardhana:
Do not proceed to verify if we have checksum.


vadmeste:
:+1:

krisis:
:+1:

PR #2289, opened 2016-07-26
api: Add bucket notification util tests.

harshavardhana:


bucket-notification-utils_test.go
```diff
@@ -15,3 +15,203 @@
...
+		// Validate if 'suffix' is correct.
+		{
+			filterName: "suffix",
+			status:     true,
+		},
+		// Invalid prefix should return false.
+		{
+			filterName: "",
+			status:     false,
+		},
```
hackintoshrao, line 41: 2 more cases with strings with wrong prefix and suffix values would be good.

bucket-notification-utils_test.go
```diff
@@ -15,3 +15,203 @@
...
+	}
+
+	for i, testCase := range testCases {
+		status := isValidFilterName(testCase.filterName)
+		if testCase.status != status {
+			t.Fatalf("Test %d: Expected \"%t\", got \"%t\"", i+1, testCase.status, status)
```
hackintoshrao: Need not stop the test here in case of failure, let all failed cases get reported, `Errorf` would be good instead of `Fatalf`.

bucket-notification-utils_test.go
```diff
@@ -15,3 +15,203 @@
...
+// Tests queue arn validation.
+func TestQueueArn(t *testing.T) {
+	savedServerConfig := serverConfig
+	defer func() {
+		serverConfig = savedServerConfig
+	}()
+	serverConfig = nil
+
+	// Test initialized config file.
+	path, err := ioutil.TempDir("", "minio-")
```
hackintoshrao: Can the function initTestConfig from [here](https://github.com/minio/minio/blob/master/test-utils_test.go#L138) be used. If not it'll be good to add this into test utils.

harshavardhana:
> Can the function initTestConfig from here be used. If not it'll be good to add this into test utils.

That is doable, but initTestConfig() itself is not generic. Let me make one.. There are few places this can be used.

bucket-notification-utils_test.go
```diff
@@ -15,3 +15,218 @@
...
+		if testCase.errCode != errCode {
+			t.Errorf("Test %d: Expected \"%d\", got \"%d\"", i+1, testCase.errCode, errCode)
+		}
+	}
+}
+
+// Tests queue arn validation.
+func TestQueueArn(t *testing.T) {
+	savedServerConfig := serverConfig
+	defer func() {
```
hackintoshrao: A comment on what this is doing and  defer is necessary will be helpful while reading code.

bucket-notification-utils_test.go
```diff
@@ -15,3 +15,218 @@
...
+		if testCase.errCode != errCode {
+			t.Errorf("Test %d: Expected \"%d\", got \"%d\"", i+1, testCase.errCode, errCode)
+		}
+	}
+}
+
+// Test unmarshal queue arn.
+func TestUnmarshalSqsArn(t *testing.T) {
+	savedServerConfig := serverConfig
+	defer func() {
```
hackintoshrao: A comment on what this is doing and defer is necessary will be helpful while reading code.

hackintoshrao:
LGTM 👍

PR #2297, opened 2016-07-26
Heal corrupted formats of disks already containing objects

### LINKED ISSUE #2182: XL/format: Some disks are corrupt i.e missing `format.json` but has bucket/objects
Currently we return failure, we need to handle this case by re-populating `format.json` by inspecting the objects.

vadmeste:
- reduceFormatErrs return multiple errors because we can have different disks with different erros
- Add healFormatXLCorruptedDisks()

Fixes #2182

### DIFF
--- a/format-config-v1.go
+++ b/format-config-v1.go
@@ -118,12 +118,18 @@ var errDiskOrderMismatch = errors.New("disk order mismatch")
 func reduceFormatErrs(errs []error, diskCount int) (err error) {
 	var errUnformattedDiskCount = 0
 	var errDiskNotFoundCount = 0
+	var errCorruptedFormatCount = 0
 	for _, dErr := range errs {
 		if dErr == errUnformattedDisk {
 			errUnformattedDiskCount++
 		} else if dErr == errDiskNotFound {
 			errDiskNotFoundCount++
+		} else if dErr == errCorruptedFormat {
+			errCorruptedFormatCount++
 		}
+	}
+	if errCorruptedFormatCount > 0 {
+		return errCorruptedFormat
 	}
 	// Unformatted disks found, we need to figure out if any disks are offline.
 	if errUnformattedDiskCount > 0 {
@@ -222,8 +228,8 @@ func genericFormatCheck(formatConfigs []*formatConfigV1, sErrs []error) (err err
 		return errXLReadQuorum
 	}
 
-	// One of the disk has corrupt format, return error.
-	if errCorruptFormatCount > 0 {
+	// Check if number of corrupted format under quorum
+	if errCorruptFormatCount > len(formatConfigs)-readQuorum {
 		return errCorruptedFormat
 	}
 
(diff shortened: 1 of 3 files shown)


format-config-v1.go
```diff
@@ -115,32 +115,40 @@ var errSomeDiskOffline = errors.New("some disks are offline")
...
 	for _, dErr := range errs {
 		if dErr == errUnformattedDisk {
 			errUnformattedDiskCount++
 		} else if dErr == errDiskNotFound {
 			errDiskNotFoundCount++
+		} else if dErr == errCorruptedFormat {
+			errCorruptedFormatCount++
 		}
 	}
+	if errCorruptedFormatCount > 0 {
```
harshavardhana, line 131: Do not understand this logic here, can you explain?

format-config-v1.go
```diff
@@ -222,8 +230,8 @@ func genericFormatCheck(formatConfigs []*formatConfigV1, sErrs []error) (err err
 		return errXLReadQuorum
 	}
 
-	// One of the disk has corrupt format, return error.
-	if errCorruptFormatCount > 0 {
+	// Check if number of corrupted format under tolerent limit
```
harshavardhana: under 'quorum' not 'tolerant limit'

format-config-v1.go
```diff
@@ -522,6 +530,131 @@ func healFormatXLFreshDisks(storageDisks []StorageAPI) error {
...
+		vols, err := storageDisks[index].ListVols()
+		if err != nil || len(vols) == 0 {
+			continue
+		}
+		objects, err := storageDisks[index].ListDir(vols[0].Name, "")
+		if err != nil || len(objects) == 0 {
+			continue
+		}
+		xlData, err := readXLMeta(storageDisks[index], vols[0].Name, objects[0])
+		if err != nil {
```
harshavardhana: We need to skip and continue only for 'err == errFileNotFound' , rest it should be an unfix-able error and we shouldn't start the server.

format-config-v1.go
```diff
@@ -522,6 +530,131 @@ func healFormatXLFreshDisks(storageDisks []StorageAPI) error {
...
+	for index, disk := range orderedDisks {
+		if disk == nil {
+			newJBOD[index] = getUUID()
+		}
+	}
+
+	// For disks with corrupted formats, inspect the disk index from any object:
+	//  1. Search for any object
+	//  2. Parse its metafile and calculate its index based on index and distribution information
+	for index, format := range formatConfigs {
```
harshavardhana: Move this entire block into a function.

format-config-v1.go
```diff
@@ -522,6 +530,131 @@ func healFormatXLFreshDisks(storageDisks []StorageAPI) error {
...
+		}
+		diskIndex := -1
+		for i, d := range xlData.Erasure.Distribution {
+			if d == xlData.Erasure.Index {
+				diskIndex = i
+			}
+		}
+		// Check for found results
+		if diskIndex == -1 || orderedDisks[diskIndex] != nil {
+			// Some inconsistant data are found, exit immediatly
```
harshavardhana: 'inconsistant' --> 'inconsistent'

harshavardhana: immediatly --> immediately.

format-config-v1.go
```diff
@@ -522,6 +530,131 @@ func healFormatXLFreshDisks(storageDisks []StorageAPI) error {
...
+		objects, err := storageDisks[index].ListDir(vols[0].Name, "")
+		if err != nil || len(objects) == 0 {
+			continue
+		}
+		xlData, err := readXLMeta(storageDisks[index], vols[0].Name, objects[0])
+		if err != nil {
+			continue
+		}
+		diskIndex := -1
+		for i, d := range xlData.Erasure.Distribution {
```
harshavardhana: All you need would be `.Index` - Distribution has no bearing here,  all objects have their own distribution. This will be inconsistent.

format-config-v1.go
```diff
@@ -522,6 +530,131 @@ func healFormatXLFreshDisks(storageDisks []StorageAPI) error {
...
+	for index, format := range formatConfigs {
+		if format != nil {
+			continue
+		}
+		vols, err := storageDisks[index].ListVols()
+		if err != nil || len(vols) == 0 {
+			continue
+		}
+		objects, err := storageDisks[index].ListDir(vols[0].Name, "")
+		if err != nil || len(objects) == 0 {
```
harshavardhana: Cases like this we should just return and continue.

format-config-v1.go
```diff
@@ -522,6 +530,131 @@ func healFormatXLFreshDisks(storageDisks []StorageAPI) error {
...
+	// For disks with corrupted formats, inspect the disk index from any object:
+	//  1. Search for any object
+	//  2. Parse its metafile and calculate its index based on index and distribution information
+	for index, format := range formatConfigs {
+		if format != nil {
+			continue
+		}
+		vols, err := storageDisks[index].ListVols()
+		if err != nil || len(vols) == 0 {
+			continue
```
harshavardhana: Here as well.

xl-v1.go
```diff
@@ -144,21 +144,27 @@ func newXLObjects(disks, ignoredDisks []string) (ObjectLayer, error) {
 	}
 
 	// Handles different cases properly.
-	switch reduceFormatErrs(sErrs, len(storageDisks)) {
-	case errUnformattedDisk:
-		// All drives online but fresh, initialize format.
-		if err := initFormatXL(storageDisks); err != nil {
-			return nil, fmt.Errorf("Unable to initialize format, %s", err)
+	for _, err := range reduceFormatErrs(sErrs, len(storageDisks)) {
+		switch err {
```
harshavardhana: Is the logic here to run across all errors and write 'format.json' again and again?

vadmeste: Well, I thought that I need to process multiple errors when some disks are corrupted and some are fresh. But I see now this is a mistake. However, healFormatXLCorruptedDisks() will need to init unformatted disks too.. I really wanted to separate healing process from adding fresh disks but the information distributed over all the disks are fully connected.

So I see why you are considering the state of all disks rather than the state of each one of the disks. Will fix that

format-config-v1.go
```diff
@@ -522,6 +528,170 @@ func healFormatXLFreshDisks(storageDisks []StorageAPI) error {
...
+				found = true
+				break
+			}
+		}
+		if !found {
+			unassignedDisks = append(unassignedDisks, storageDisks[i])
+		}
+	}
+
+	// Assign unsiggned disks to nil elements in orderedDisks
```
harshavardhana: 'unAssigned'

format-config-v1.go
```diff
@@ -522,6 +528,170 @@ func healFormatXLFreshDisks(storageDisks []StorageAPI) error {
...
+			return nil
+		}
+		orderedDisks[diskIndex] = storageDisks[index]
+		assignedDisks = append(assignedDisks, storageDisks[index])
+	}
+
+	// At this stage, all disks with corrupted formats but with objects inside found their way.
+	// Now take care of unformatted disks, which are the `unassignedDisks`
+
+	// Populate unassignDisks
```
harshavardhana: 'unAssignedDisks'

format-config-v1.go
harshavardhana: It would nicer to move this into a function of its own. Easy to test.

format-config-v1.go
```diff
@@ -522,6 +528,170 @@ func healFormatXLFreshDisks(storageDisks []StorageAPI) error {
...
+		}
+		xlData, err := readXLMeta(storageDisks[index], vols[0].Name, objects[0])
+		if err != nil {
+			if err == errFileNotFound {
+				continue
+			}
+			return err
+		}
+		diskIndex := -1
+		for i, d := range xlData.Erasure.Distribution {
```
harshavardhana: You can implement a function which does this.

format-config-v1.go
```diff
@@ -522,6 +528,170 @@ func healFormatXLFreshDisks(storageDisks []StorageAPI) error {
...
+		}
+		orderedDisks[diskIndex] = storageDisks[index]
+		assignedDisks = append(assignedDisks, storageDisks[index])
+	}
+
+	// At this stage, all disks with corrupted formats but with objects inside found their way.
+	// Now take care of unformatted disks, which are the `unassignedDisks`
+
+	// Populate unassignDisks
+	for i := range storageDisks {
```
harshavardhana: Even put this into a function as well.

vadmeste:
:+1:

format-config-v1.go
```diff
@@ -522,6 +528,177 @@ func healFormatXLFreshDisks(storageDisks []StorageAPI) error {
 	return saveFormatXL(orderedDisks, newFormatConfigs)
 }
```
harshavardhana, line 530: Add a comment.

format-config-v1.go
```diff
@@ -522,6 +528,177 @@ func healFormatXLFreshDisks(storageDisks []StorageAPI) error {
...
+			}
+		}
+		if !found {
+			unAssignedDisks = append(unAssignedDisks, storageDisks[i])
+		}
+	}
+	return assignedDisks, unAssignedDisks
+}
+
+func reorderDisksByInspection(orderedDisks, storageDisks []StorageAPI, formatConfigs []*formatConfigV1) ([]StorageAPI, error) {
```
harshavardhana, line 552: Add a comment.

PR #2299, opened 2016-07-27
XL/Erasure: Make bit-rot verification based on xl.json algo.

harshavardhana:
Currently `xl.json` saves algorithm information for bit-rot
verification. Since the bit-rot algo's can change in the
future make sure the erasureReadFile doesn't default to
a particular algo. Instead use the checkSumInfo.

### DIFF
--- a/erasure-createfile.go
+++ b/erasure-createfile.go
@@ -28,11 +28,11 @@ import (
 // erasureCreateFile - writes an entire stream by erasure coding to
 // all the disks, writes also calculate individual block's checksum
 // for future bit-rot protection.
-func erasureCreateFile(disks []StorageAPI, volume, path string, reader io.Reader, blockSize int64, dataBlocks int, parityBlocks int, writeQuorum int) (bytesWritten int64, checkSums []string, err error) {
+func erasureCreateFile(disks []StorageAPI, volume, path string, reader io.Reader, blockSize int64, dataBlocks int, parityBlocks int, algo string, writeQuorum int) (bytesWritten int64, checkSums []string, err error) {
 	// Allocated blockSized buffer for reading.
 	buf := make([]byte, blockSize)
 
-	hashWriters := newHashWriters(len(disks))
+	hashWriters := newHashWriters(len(disks), algo)
 
 	// Read until io.EOF, erasure codes data and writes to all disks.
 	for {
--- a/erasure-readfile.go
+++ b/erasure-readfile.go
@@ -159,7 +159,7 @@ func parallelRead(volume, path string, readDisks []StorageAPI, orderedDisks []St
 // are decoded into a data block. Data block is trimmed for given offset and length,
 // then written to given writer. This function also supports bit-rot detection by
 // verifying checksum of individual block's checksum.
-func erasureReadFile(writer io.Writer, disks []StorageAPI, volume string, path string, offset int64, length int64, totalLength int64, blockSize int64, dataBlocks int, parityBlocks int, checkSums []string, pool *bpool.BytePool) (int64, error) {
+func erasureReadFile(writer io.Writer, disks []StorageAPI, volume string, path string, offset int64, length int64, totalLength int64, blockSize int64, dataBlocks int, parityBlocks int, checkSums []string, algo string, pool *bpool.BytePool) (int64, error) {
 	// Offset and length cannot be negative.
 	if offset < 0 || length < 0 {
 		return 0, errUnexpected
@@ -186,7 +186,7 @@ func erasureReadFile(writer io.Writer, disks []StorageAPI, volume string, path s
 				return true
 			}
 			// Is this a valid block?
-			isValid := isValidBlock(disks[diskIndex], volume, path, checkSums[diskIndex])
+			isValid := isValidBlock(disks[diskIndex], volume, path, checkSums[diskIndex], algo)
 			verified[diskIndex] = isValid
 			return isValid
 		}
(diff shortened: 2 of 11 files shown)


krishnasrinivas:
reviewing.

erasure-createfile.go
```diff
@@ -28,11 +28,11 @@ import (
 // erasureCreateFile - writes an entire stream by erasure coding to
 // all the disks, writes also calculate individual block's checksum
 // for future bit-rot protection.
-func erasureCreateFile(disks []StorageAPI, volume, path string, reader io.Reader, blockSize int64, dataBlocks int, parityBlocks int, writeQuorum int) (bytesWritten int64, checkSums []string, err error) {
+func erasureCreateFile(disks []StorageAPI, volume, path string, reader io.Reader, blockSize int64, dataBlocks int, parityBlocks int, algo string, writeQuorum int) (bytesWritten int64, checkSums []checkSumInfo, err error) {
 	// Allocated blockSized buffer for reading.
```
krishnasrinivas, line 32: it's correct to pass algo as a string instead of hard coding.

but code is more readable if erasureCreateFile returns `checkSums []string` and constructing checkSumInfo happens at one place (xl.PutObject) than filling some fields in erasureCreateFile and some fields in xl.PutObject

erasure-readfile.go
```diff
@@ -159,7 +159,7 @@ func parallelRead(volume, path string, readDisks []StorageAPI, orderedDisks []St
 // are decoded into a data block. Data block is trimmed for given offset and length,
 // then written to given writer. This function also supports bit-rot detection by
 // verifying checksum of individual block's checksum.
-func erasureReadFile(writer io.Writer, disks []StorageAPI, volume string, path string, offset int64, length int64, totalLength int64, blockSize int64, dataBlocks int, parityBlocks int, checkSums []string, pool *bpool.BytePool) (int64, error) {
+func erasureReadFile(writer io.Writer, disks []StorageAPI, volume string, path string, offset int64, length int64, totalLength int64, blockSize int64, dataBlocks int, parityBlocks int, checkSums []checkSumInfo, pool *bpool.BytePool) (int64, error) {
 	// Offset and length cannot be negative.
```
krishnasrinivas, line 163: For erasureReadFile also, it's better to pass `algo string` and `checkSums []string` separately than passing `[]checkSumInfo` as it expresses the function dependency explicitly than to pass checkSumInfo which has the partName also which is not needed by erasureReadFile.

passing []checkSumInfo makes the code reader of erasureReadFile believe that each disk will have different algo

checkSumInfo belongs to a higher layer as it contains partName which is not used by erasureCreateFile and erasureReadFile

xl-v1-object.go
```diff
@@ -413,7 +413,7 @@ func (xl xlObjects) PutObject(bucket string, object string, size int64, data io.
 	onlineDisks := getOrderedDisks(xlMeta.Erasure.Distribution, xl.storageDisks)
 
 	// Erasure code data and write across all disks.
-	sizeWritten, checkSums, err := erasureCreateFile(onlineDisks, minioMetaBucket, tempErasureObj, teeReader, xlMeta.Erasure.BlockSize, xlMeta.Erasure.DataBlocks, xlMeta.Erasure.ParityBlocks, xl.writeQuorum)
+	sizeWritten, checkSums, err := erasureCreateFile(onlineDisks, minioMetaBucket, tempErasureObj, teeReader, xlMeta.Erasure.BlockSize, xlMeta.Erasure.DataBlocks, xlMeta.Erasure.ParityBlocks, blakeAlgo, xl.writeQuorum)
```
krishnasrinivas: in erasureCreateFile and erasureReadFile you can pass `bitRotAlgo` and define `bitRotAlgo` to `blakeAlgo` so if you decide to change to bit-rot algo you just need to define `bitRotAlgo` to something else

erasure-createfile.go
harshavardhana, line 32: > but code is more readable if erasureCreateFile returns checkSums []string and constructing checkSumInfo happens at one place (xl.PutObject) than filling some fields in erasureCreateFile and some fields in xl.PutObject

This is correct. Makes sense will do those changes.

xl-v1-metadata.go
```diff
@@ -52,6 +52,12 @@ type checkSumInfo struct {
 	Hash      string `json:"hash"`
 }
 
+// Default bit-rot checksum technique used in version '1'.
+const (
+	bitRotAlgoV1 = "blake2b"
```
krishnasrinivas: it will be incorrect to call it `bitRotAlgoV1` because V1 format allows different objects to have different algos, i.e if we change the algo even then xl format is still V1

krishnasrinivas:
looks good apart from one comment on the name `bitRotAlgoV1`

xl-v1-metadata.go
harshavardhana: > it will be incorrect to call it bitRotAlgoV1 because V1 format allows different objects to have different algos, i.e if we change the algo even then xl format is still V1

What it indicates is that we are using "blake2b" for V1 format.  Also while it is possible we are never going to write objects with different checksum algos with in a given instance of a running server - there is no facility in erasureCreateFile() to do that, we can read different algos though.

harshavardhana: This value indicates how the bit-rot will be initially created along with V1 format.

krishnasrinivas: let's say we move to a different algo 'sha512', how will you make the changes in the names bitRotAlgoV1 and bitRotAlgo to accommodate the new algo?

harshavardhana: > let's say we move to a different algo 'sha512', how will you make the changes in the names bitRotAlgoV1 and bitRotAlgo to accommodate the new algo?

```
 bitRotAlgoV2 = "sha512" 
 bitRotAlgo = bitRotAlgoV2
```

krishnasrinivas: `V` here stands for version. "sha512" will be a different algo from "blake2b" so it can't be a version increment.

Did you mean to call them bitRotAlgo1 and bitRotAlgo2?

if `V` version is for xl format version, even then xl version would still be same so calling it bitRotAlgoV2 will be inappropriate

harshavardhana: > V here stands for version. "sha512" will be a different algo from "blake2b" so it can't be a version increment.
> 
> Did you mean to call them bitRotAlgo1 and bitRotAlgo2?
> 
> if V version is for xl format version, even then xl version would still be same so calling it bitRotAlgoV2 will be inappropriate

Actually you are right it won't make sense.. Since we won't bump `xl.json` with algo changes.

krishnasrinivas:
LGTM

PR #2311, opened 2016-07-28
Cleanup: Code organizing. 

hackintoshrao:
Remove listOnlineDisk() from x-v1-healing.go and place it in xl-v1-utils.go


hackintoshrao:
Since healing requires listing of online disks, its good remain there.

Issue #2290, opened 2016-07-26
api: Post Presigned post policy should use less memory and stage the i/o.


harshavardhana:
```
func extractHTTPFormValues(reader *multipart.Reader) (io.Reader, map[string]string, error) {
        /// HTML Form values
        formValues := make(map[string]string)
        filePart := new(bytes.Buffer)
        var err error
        for err == nil {
                var part *multipart.Part
                part, err = reader.NextPart()
                if part != nil {
                        if part.FileName() == "" {
                                var buffer []byte
                                buffer, err = ioutil.ReadAll(part) ---> Code here could server to OOM. 
                                if err != nil {
                                        return nil, nil, err
                                }
                                formValues[http.CanonicalHeaderKey(part.FormName())] = string(buffer)
                        } else {
                                if _, err = io.Copy(filePart, part); err != nil {
                                        return nil, nil, err
                                }
                        }
                }
        }
```

harshavardhana:
Assigned to @vadmeste

vadmeste:
@harshavardhana, it seems to me that this part of code is useless and someone has misunderstood the S3 spec, let's discuss further

harshavardhana:
@vadmeste after the recent patch merge is this still valid?

vadmeste:
@harshavardhana, this is not fixed yet

harshavardhana:
Fixed by #2317

PR #2317, opened 2016-07-28
Limit POST form fields and file size + Generic Request Size limiter 

### LINKED ISSUE #2316: Limit request size body to a maximum size
A malicious client can send a very large request which consumes first the memory then the disks in temporary files. Secure that with a golang API: https://golang.org/pkg/net/http/#MaxBytesReader

vadmeste:
Fixes #2316  and #2290

### DIFF
--- a/generic-handlers.go
+++ b/generic-handlers.go
@@ -39,6 +39,27 @@ func registerHandlers(mux *router.Router, handlerFns ...HandlerFunc) http.Handle
 	return f
 }
 
+// Adds limiting body size middleware
+
+// Set the body size limit to 6 Gb = Maximum object size + other possible data
+// in the same request
+const requestMaxBodySize = 1024 * 1024 * 1024 * (5 + 1)
+
+type requestSizeLimitHandler struct {
+	handler     http.Handler
+	maxBodySize int64
+}
+
+func setRequestSizeLimitHandler(h http.Handler) http.Handler {
+	return requestSizeLimitHandler{handler: h, maxBodySize: requestMaxBodySize}
+}
+
+func (h requestSizeLimitHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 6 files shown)


generic-handlers.go
```diff
@@ -39,6 +39,26 @@ func registerHandlers(mux *router.Router, handlerFns ...HandlerFunc) http.Handle
...
+	handler     http.Handler
+	maxBodySize int64
+}
+
+func setRequestSizeLimitHandler(h http.Handler) http.Handler {
+	return requestSizeLimitHandler{handler: h, maxBodySize: requestMaxBodySize}
+}
+
+func (h requestSizeLimitHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	r.Body = http.MaxBytesReader(w, r.Body, h.maxBodySize)
```
harshavardhana, line 59: A comment here.

handler-utils.go
```diff
@@ -99,32 +98,33 @@ func extractMetadataFromHeader(header http.Header) map[string]string {
...
 		var part *multipart.Part
 		part, err = reader.NextPart()
 		if part != nil {
 			canonicalFormName := http.CanonicalHeaderKey(part.FormName())
 			if canonicalFormName != "File" {
 				var buffer []byte
-				buffer, err = ioutil.ReadAll(part)
+				// Limit fields size (except file) to 1Mib since Policy document
+				// can reach that size according to https://aws.amazon.com/articles/1434
+				limitReader := io.LimitReader(part, 1024*1024)
```
harshavardhana: Add this as global constant and cite the article.

harshavardhana: Also do 1024*1024+1 - so that ioutil.ReadAll() returns a buffer we can validate if it is bigger than 1Mib and throw an appropriate error as well.

PR #2319, opened 2016-07-28
tests: Add missing unit test for posix.ReadFile.

harshavardhana:


posix_test.go
```diff
@@ -401,6 +401,91 @@ func TestDeleteFile(t *testing.T) {
...
+
+	var buf = make([]byte, 5)
+
+	// Create test files for further reading.
+	for _, testCase := range testCases {
+		if err := posix.AppendFile("success-vol", testCase.fileName, []byte("hello, world")); err != testCase.expectedErr {
```
hackintoshrao: Why asserting for expectedErr in `AppendFile` ?

posix_test.go
```diff
@@ -401,6 +401,91 @@ func TestDeleteFile(t *testing.T) {
...
+			t.Fatalf("case: %s, expected: %s, got: %s", testCase, testCase.expectedErr, err)
+		}
+	}
+
+	for _, testCase := range testCases {
+		if _, err := posix.ReadFile("success-vol", testCase.fileName, 0, buf); err != testCase.expectedErr {
+			t.Fatalf("case: %s, expected: %s, got: %s", testCase, testCase.expectedErr, err)
```
hackintoshrao: Avoid `Fatalf` wherever possible .

posix_test.go
```diff
@@ -401,6 +401,91 @@ func TestDeleteFile(t *testing.T) {
...
+
+	// Create test files for further reading.
+	for _, testCase := range testCases {
+		if err := posix.AppendFile("success-vol", testCase.fileName, []byte("hello, world")); err != testCase.expectedErr {
+			t.Fatalf("case: %s, expected: %s, got: %s", testCase, testCase.expectedErr, err)
+		}
+	}
+
+	for _, testCase := range testCases {
+		if _, err := posix.ReadFile("success-vol", testCase.fileName, 0, buf); err != testCase.expectedErr {
```
hackintoshrao: It'll be good to not to ignore a return value, its good to assert it for expected value.

posix_test.go
```diff
@@ -401,6 +401,91 @@ func TestDeleteFile(t *testing.T) {
...
+		expectedErr error
+	}{
+		{"myobject", nil},
+		{"path/to/my/object", nil},
+		// Test to read from previously created file.
+		{"myobject", nil},
+		// Test to use same path of previously created file.
+		{"path/to/my/testobject", nil},
+		// One path segment length is 255 chars long.
+		{"path/to/my/object000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001", nil},
```
hackintoshrao: Need a case on non-existatnt file.

posix_test.go
```diff
@@ -401,6 +402,102 @@ func TestDeleteFile(t *testing.T) {
...
+	// Create directory to make errIsNotRegular
+	if err = os.Mkdir(slashpath.Join(path, "success-vol", "object-as-dir"), 0777); err != nil {
+		t.Fatalf("Unable to create directory, %s", err)
+	}
+
+	testCases := []struct {
+		fileName    string
+		expectedErr error
+	}{
+		{"myobject", nil},
```
hackintoshrao: 4 more cases would be great 
- case with non-existant object. 
- case where fd.Seek fails, may be invalid offset. 
- case where EOF is returned by ReadFull. 
- case for errFaultyDisk.

hackintoshrao: Need a case wherein the buffer size is bigger than that of the actual object size.

posix_test.go
```diff
@@ -401,6 +405,150 @@ func TestDeleteFile(t *testing.T) {
...
+	// Create directory to make errIsNotRegular
+	if err = os.Mkdir(slashpath.Join(path, "success-vol", "object-as-dir"), 0777); err != nil {
+		t.Fatalf("Unable to create directory, %s", err)
+	}
+
+	testCases := []struct {
+		fileName    string
+		offset      int64
+		bufSize     int
+		expectedBuf []byte
```
hackintoshrao, line 436: It'll be good to follow the format of using `shouldPass` flag for tests, it'll be consistent with the other tests pattern.  Rest looks really beautiful.

posix_test.go
```diff
@@ -401,6 +405,159 @@ func TestDeleteFile(t *testing.T) {
...
+		expectedErr error
+	}{
```
hackintoshrao, line 438: It'll be good to add case numbers before the cases `// Test case  - n`, easy to follow in case of failures.

posix_test.go
```diff
@@ -401,6 +404,167 @@ func TestDeleteFile(t *testing.T) {
...
+			"myobject",
+			7, 8,
+			[]byte("world"),
+			io.ErrUnexpectedEOF,
+		},
+		// Seeking into a wrong offset, return PathError. - 11
+		{
+			"myobject",
+			-1, 5,
+			[]byte("hello"),
```
hackintoshrao: `expectedResult`  should be empty here.

posix_test.go
```diff
@@ -401,6 +404,167 @@ func TestDeleteFile(t *testing.T) {
...
+			[]byte("world"), nil,
+		},
+		// Reading from an object but buffer size greater. - 10
+		{
+			"myobject",
+			7, 8,
+			[]byte("world"),
+			io.ErrUnexpectedEOF,
+		},
+		// Seeking into a wrong offset, return PathError. - 11
```
hackintoshrao, line 496: Can just one case of seeking ahead of limits be added.

posix_test.go
```diff
@@ -401,6 +404,167 @@ func TestDeleteFile(t *testing.T) {
...
+		{
+			"myobject", 7, 5,
+			[]byte("world"), nil,
+		},
+		// Reading from an object but buffer size greater. - 10
+		{
+			"myobject",
+			7, 8,
+			[]byte("world"),
+			io.ErrUnexpectedEOF,
```
hackintoshrao, line 494: This is a special case wherein both error and the result is returned, not sure how to validate for both, may be as a separate case since its very crucial case. CopyBuffer -> Hashsum -> IsValidBlock -> BitRotverify -> ReadParallel -> GetOBject -> , the entire execution tree will be influenced by this particular case. Will handle this copyBuffer tests too.

harshavardhana:
> This is a special case wherein both error and the result is returned, not sure how to validate for both, may be as a separate case since its very crucial case. CopyBuffer -> Hashsum -> IsValidBlock -> BitRotverify -> ReadParallel -> GetOBject -> , the entire execution tree will be influenced by this particular case. Will handle this copyBuffer tests too.

Yes it is treated as error nonetheless since we are expecting it to fail, how the top level caller deals with this is entirely different.

harshavardhana:
> Can just one case of seeking ahead of limits be added.

Done

hackintoshrao:
Looks Great .. Good and important set of tests. LGTM 👍  Thanks.

PR #2321, opened 2016-07-28
bucket-policy: Migrate bucket policy to minioMetaBuket/buckets

### LINKED ISSUE #2214: Move bucket-policy as part of the special namespace itself.
- policy.json 
- notification.json 

Should go into XL backend itself and be protected, treat this like user set data.

krisis:
The bucket policy is moved into the minioMetaBucket similar to
notification config. This means that the bucket policies are also
erasure coded when used with multi-disk setup (XL).

Fixes #2214

### DIFF
--- a/bucket-policy-migrate.go
+++ b/bucket-policy-migrate.go
@@ -37,30 +37,9 @@ func cleanupOldBucketPolicyConfigs() error {
 	oldBucketsConfigDir, err := getOldBucketsConfigPath()
 	fatalIf(err, "Unable to fetch buckets config path to migrate bucket policy")
 
-	// Check if config directory holding bucket policy exists before
-	// starting to clean up.
-	_, err = os.Stat(oldBucketsConfigDir)
-	if os.IsNotExist(err) {
-		return nil
-	}
-	if err != nil {
-		return err
-	}
-
-	// WalkFunc that deletes bucket policy files in the config directory
-	// after successfully migrating all of them.
-	cleanupOldBucketPolicy := func(policyPath string, fileInfo os.FileInfo, err error) error {
-		if err != nil {
-			return err
-		}
-		// Skip entries that aren't bucket policy files.
-		if fileInfo.Name() != "access-policy.json" {
-			return nil
-		}
-		// Delete the bucket policy file from config directory.
-		return os.Remove(policyPath)
-	}
-	return filepath.Walk(oldBucketsConfigDir, cleanupOldBucketPolicy)
+	// Recursively remove configDir/buckets/ - old bucket policy config location.
+	// N B This is called only if all bucket policies were successfully migrated.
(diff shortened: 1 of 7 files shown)


bucket-policy-migrate.go
```diff
@@ -0,0 +1,99 @@
+/*
+ * Minio Cloud Storage, (C) 2015, 2016 Minio, Inc.
```
harshavardhana: Remove the '2015' license header here.

bucket-policy-migrate.go
```diff
@@ -0,0 +1,99 @@
...
+
+func cleanupOldBucketPolicyConfigs() error {
+	// Get old bucket policy config directory.
+	oldBucketsConfigDir, err := getOldBucketsConfigPath()
+	fatalIf(err, "Unable to fetch buckets config path to migrate bucket policy")
+
+	// Check if config directory holding bucket policy exists before
+	// starting to clean up.
+	_, err = os.Stat(oldBucketsConfigDir)
+	if err != nil {
```
harshavardhana: Shouldn't only return nil if 'os.IsNotExist(err)' is true.

bucket-policy-migrate.go
```diff
@@ -0,0 +1,99 @@
...
+		return nil
+	}
+
+	// WalkFunc that deletes bucket policy files in the config directory
+	// after successfully migrating all of them.
+	var cleanupOldBucketPolicy filepath.WalkFunc
```
harshavardhana: Why have a type, this will be inferred anyways.

krisis: I meant to leave the type there as documentation. Given that this is idiomatic, other Go developers can infer the type too. I shall change it to use `:=` operator instead.

bucket-policy-migrate.go
krisis: OK. I shall change the error handling to return nil only when the config directory is missing.

bucket-policy-migrate.go
krisis: Will remove it in the next update to the PR.

bucket-policy.go
```diff
@@ -32,97 +31,56 @@ func getBucketsConfigPath() (string, error) {
...
-		if os.IsNotExist(err) {
+		if _, ok := err.(ObjectNotFound); ok {
 			return nil, BucketPolicyNotFound{Bucket: bucket}
 		}
 		return nil, err
 	}
-	return ioutil.ReadFile(bucketPolicyFile)
+	var buffer bytes.Buffer
+	err = api.ObjectAPI.GetObject(minioMetaBucket, policyPath, 0, objInfo.Size, &buffer)
+	if err != nil {
```
harshavardhana, line 59: We should return 'BucketPolicyNotFound' here as well. just like GetObjectInfo.

PR #2322, opened 2016-07-29
Adding return error value to checkPortAvailability to enable testing …

jesselucas:
Adding checkport_test.go to test checkPortAvailability. Updated server-main.go to use error value from checkPortAvailability and calls fatalIf if an error is returned.

### DIFF
--- a/checkport_test.go
+++ b/checkport_test.go
@@ -1,0 +1,54 @@
+/*
+ * Minio Cloud Storage, (C) 2016 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package main
+
+import (
+	"fmt"
+	"net"
+	"runtime"
+	"testing"
+)
+
+func TestCheckPortAvailability(t *testing.T) {
+	tests := []struct {
+		port int
+	}{
+		{9000},
+		{10000},
+	}
+	for _, test := range tests {
+		// This test should pass if the ports are available
+		err := checkPortAvailability(test.port)
+		if err != nil {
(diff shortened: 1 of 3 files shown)


harshavardhana:
Thanks @jesselucas for the patch.  Looks like the behavior is slightly different on windows.

checkport_test.go
```diff
@@ -0,0 +1,35 @@
+package main
```
harshavardhana, line 17: Missing license header.  Should be fixed.

checkport_test.go
```diff
@@ -0,0 +1,35 @@
...
+	}
+	for _, test := range tests {
+		// This test should pass if the ports are available
+		err := checkPortAvailability(test.port)
+		if err != nil {
+			t.Fatalf("checkPortAvailability test failed for port: %d. Error: %s", test.port, err)
+		}
+
+		// Now use the ports and check again
+		ln, err := net.Listen("tcp", net.JoinHostPort("127.0.0.1", strconv.Itoa(test.port)))
```
harshavardhana: The best way to generate a problem  on windows would be to listen on 

```
net.Listen("tcp", fmt.Sprintf(":%d", test.port))
```

That way it listens on all interfaces. There are other problems though over this verification 

https://github.com/golang/go/issues/7598

checkport_test.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+	for _, test := range tests {
+		// This test should pass if the ports are available
+		err := checkPortAvailability(test.port)
+		if err != nil {
+			t.Fatalf("checkPortAvailability test failed for port: %d. Error: %v", test.port, err)
+		}
+
+		// Now use the ports and check again
+		ln, err := net.Listen("tcp", fmt.Sprintf(":%d", test.port))
```
harshavardhana, line 41: Looks like this won't be valid on windows apparently due to a Go bug.  We could skip this check for 'windows' .

checkport_test.go
```diff
@@ -0,0 +1,51 @@
...
+		if err != nil {
+			t.Fail()
+		}
+		defer ln.Close()
+
+		err = checkPortAvailability(test.port)
+		if err == nil {
```
harshavardhana: Perhaps a workaround for now, we could cite here as a small comment on why it is done this way. 

```
if err == nil && runtime.GOOS != windows {
```

jesselucas: Trying one more thing and if this doesn't work I'll implement the windows check. Thank you for all the help!

jesselucas: Added in the os check for windows. Let me know if you see anything else.

harshavardhana:
LGTM, Thanks :+1:

PR #2325, opened 2016-07-29
Refactoring checkPortAvailability to check each tcp network (tcp, tcp…

jesselucas:
After reading through https://github.com/golang/go/issues/7598 it gave me the idea to check each network rather than each address.

### DIFF
--- a/checkport.go
+++ b/checkport.go
@@ -17,11 +17,8 @@
 package main
 
 import (
-	"errors"
 	"fmt"
 	"net"
-	"os"
-	"syscall"
 )
 
 // Make sure that none of the other processes are listening on the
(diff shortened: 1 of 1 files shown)


checkport.go
```diff
@@ -32,70 +29,15 @@ import (
...
-				// Ignore other errors.
-				continue
-
-			}
-
-			if err = l.Close(); err != nil {
-				return err
-			}
-		}
+		l.Close()
```
harshavardhana: We could look for the error returned here, so that we don't have dangling connections left after this check, while the occurrence is highly unlikely.

harshavardhana:
LGTM, Thanks :+1:

PR #2331, opened 2016-07-30
server/config: Create 'certs' directory in initServerConfig().

harshavardhana:
certs directory was created only if config was not present, our
expectancy is we need 'certs' directory to be present all the
time making it easier to be documented.

### DIFF
--- a/server-main.go
+++ b/server-main.go
@@ -154,8 +154,12 @@ func finalizeEndpoints(tls bool, apiServer *http.Server) (endPoints []string) {
 
 // initServerConfig initialize server config.
 func initServerConfig(c *cli.Context) {
+	// Create certs path.
+	err := createCertsPath()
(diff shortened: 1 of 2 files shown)


server-main.go
```diff
@@ -154,8 +154,12 @@ func finalizeEndpoints(tls bool, apiServer *http.Server) (endPoints []string) {
 
 // initServerConfig initialize server config.
 func initServerConfig(c *cli.Context) {
+	// Create certs path.
+	err := createCertsPath()
```
krishnasrinivas, line 158: createCertsPath returns error of the dirs already exist

harshavardhana:
> createCertsPath returns error of the dirs already exist

why do you think so? i don't see any check in the code path. 'os.MkdirAll' shouldn't return an error on existing dir.

PR #2332, opened 2016-07-30
server: set globalCacheSize honoring system limits for max memory.

harshavardhana:
On unix systems it is possible to set max memory used by
running processes using 'ulimit -m' or 'syscall.RLIMIT_AS'.

A process whence exceeds this limit, kernel would pro-actively
kill such a server with OOM. To avoid this problem of defaulting
our cache size to 8GB we should look for if the current system
limits are lower and set the cache size appropriately. So that
we do not crash minio server by over allocating that what
is allowed.

### DIFF
--- a/server-rlimit-nix.go
+++ b/server-rlimit-nix.go
@@ -30,7 +30,7 @@ func setMaxOpenFiles() error {
 		return err
 	}
 	// Set the current limit to Max, it is usually around 4096.
-	// TO increate this limit further user has to manually edit
+	// TO increase this limit further user has to manually edit
 	// `/etc/security/limits.conf`
 	rLimit.Cur = rLimit.Max
 	err = syscall.Setrlimit(syscall.RLIMIT_NOFILE, &rLimit)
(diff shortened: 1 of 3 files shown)


server-rlimit-nix.go
```diff
@@ -43,3 +43,34 @@ func setMaxOpenFiles() error {
...
+	if err != nil {
+		return err
+	}
+	err = syscall.Getrlimit(syscall.RLIMIT_AS, &rLimit)
+	if err != nil {
+		return err
+	}
+	// Validate if rlimit memory is set to lower
+	// than max cache size. Then we should use such value.
+	if rLimit.Cur < globalMaxCacheSize {
```
krishnasrinivas, line 72: if rLimit.Cur is `RLIM_INFINITY` then we should limit globalMaxCacheSize so that it is < RAM size

harshavardhana:
> if rLimit.Cur is RLIM_INFINITY then we should limit globalMaxCacheSize so that it is < RAM size

Yeah good idea @krishnasrinivas, please make that change.

PR #2338, opened 2016-08-01
Object-cache: enforce cache size to be less than RAM.

### LINKED ISSUE #2337: limit max object cache size based on physical memory
Linux has `syscall.Sysinfo()` other OSes don't, need to see the best way to limit object cache size.

krishnasrinivas:
Implemented on linux because of availability of syscall.SysInfo()
fixes #2337

For other OSes there seemed to be no way of doing this other than reading /proc/

### DIFF
--- a/server-rlimit-nix.go
+++ b/server-rlimit-nix.go
@@ -18,7 +18,12 @@
 
 package main
 
-import "syscall"
+import (
+	"fmt"
+	"syscall"
+
+	"github.com/minio/minio/pkg/sys"
+)
 
 // For all unixes we need to bump allowed number of open files to a
 // higher value than its usual default of '1024'. The reasoning is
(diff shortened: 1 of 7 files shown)


server-rlimit-nix.go
```diff
@@ -59,6 +64,7 @@ func setMaxMemory() error {
 	// TO decrease this limit further user has to manually edit
 	// `/etc/security/limits.conf`
 	rLimit.Cur = rLimit.Max
+	fmt.Println(rLimit.Max)
```
harshavardhana: Remove print.

krishnasrinivas:
moved ErrNotImplemented to stats.go because:

```
# github.com/minio/minio
./server-rlimit-nix.go:84: undefined: "github.com/minio/minio/pkg/sys".ErrNotImplemented
```

PR #2341, opened 2016-08-02
Graceful shutdown for ServerMux

jesselucas:
First attempt at adding graceful shutdown to ServerMux

### DIFF
--- a/server-mux.go
+++ b/server-mux.go
@@ -18,11 +18,14 @@ package main
 
 import (
 	"crypto/tls"
+	"errors"
 	"io"
 	"net"
 	"net/http"
 	"net/url"
 	"strings"
+	"sync"
+	"time"
 )
 
 var defaultHTTP2Methods = []string{
@@ -120,31 +123,41 @@ func (c *MuxConn) Read(b []byte) (int, error) {
 type MuxListener struct {
 	net.Listener
 	config *tls.Config
+	wg     *sync.WaitGroup
 }
 
 // NewMuxListener - creates new MuxListener, returns error when cert/key files are not found
 // or invalid
-func NewMuxListener(listener net.Listener, certPath, keyPath string) (MuxListener, error) {
+func NewMuxListener(listener net.Listener, wg *sync.WaitGroup, certPath, keyPath string) (*MuxListener, error) {
 	var err error
-	config := &tls.Config{}
-	if config.NextProtos == nil {
-		config.NextProtos = []string{"http/1.1", "h2"}
+	var config *tls.Config
+	config = nil
+
+	if certPath != "" {
+		config = &tls.Config{}
+		if config.NextProtos == nil {
+			config.NextProtos = []string{"http/1.1", "h2"}
+		}
+		config.Certificates = make([]tls.Certificate, 1)
+		config.Certificates[0], err = tls.LoadX509KeyPair(certPath, keyPath)
+		if err != nil {
+			return &MuxListener{}, err
+		}
 	}
-	config.Certificates = make([]tls.Certificate, 1)
-	config.Certificates[0], err = tls.LoadX509KeyPair(certPath, keyPath)
-	if err != nil {
-		return MuxListener{}, err
-	}
-	return MuxListener{Listener: listener, config: config}, nil
+
+	l := &MuxListener{Listener: listener, config: config, wg: wg}
+
+	return l, nil
 }
 
 // Accept - peek the protocol to decide if we should wrap the
 // network stream with the TLS server
-func (l MuxListener) Accept() (net.Conn, error) {
+func (l *MuxListener) Accept() (net.Conn, error) {
 	c, err := l.Listener.Accept()
 	if err != nil {
 		return c, err
 	}
+
 	cmux := NewMuxConn(c)
 	protocol := cmux.PeekProtocol()
 	if protocol == "tls" {
@@ -153,9 +166,46 @@ func (l MuxListener) Accept() (net.Conn, error) {
 	return cmux, nil
 }
 
+// Close Listener
+func (l *MuxListener) Close() error {
+	if l == nil {
+		return nil
+	}
+
+	return l.Listener.Close()
+}
+
 // MuxServer - the main mux server
 type MuxServer struct {
 	http.Server
+	listener        *MuxListener
+	WaitGroup       *sync.WaitGroup
+	GracefulTimeout time.Duration
+	mu              sync.Mutex // guards closed and conns
+	closed          bool
+	conns           map[net.Conn]http.ConnState // except terminal states
+}
+
+// NewMuxServer constructor to create a MuxServer
+func NewMuxServer(addr string, handler http.Handler) *MuxServer {
+	m := &MuxServer{
+		Server: http.Server{
+			Addr: addr,
+			// Adding timeout of 10 minutes for unresponsive client connections.
+			ReadTimeout:    10 * time.Minute,
+			WriteTimeout:   10 * time.Minute,
+			Handler:        handler,
+			MaxHeaderBytes: 1 << 20,
+		},
+		WaitGroup:       &sync.WaitGroup{},
+		GracefulTimeout: 5 * time.Second,
+	}
+
+	// Track connection state
+	m.connState()
+
+	// Returns configured HTTP server.
+	return m
 }
 
 // ListenAndServeTLS - similar to the http.Server version. However, it has the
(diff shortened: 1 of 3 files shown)


harshavardhana:
// @vadmeste for your review.

server-mux.go
```diff
@@ -53,17 +55,58 @@ type MuxConn struct {
...
+			ReadTimeout:    10 * time.Minute,
+			WriteTimeout:   10 * time.Minute,
+			Handler:        configureServerHandler(srvCmdConfig),
+			MaxHeaderBytes: 1 << 20,
+		},
+		wg:              new(sync.WaitGroup),
+		GracefulTimeout: 5 * time.Second,
+	}
+
+	m.Server.ConnState = func(c net.Conn, cs http.ConnState) {
```
vadmeste: Could you explain why we need this ?

server-mux.go
```diff
@@ -115,47 +158,75 @@ func (c *MuxConn) Read(b []byte) (int, error) {
...
 }
 
+// Close the connection and decrements the WaitGroup
+func (c *MuxConn) Close() error {
+	err := c.Conn.Close()
+	if err != nil {
+		return err
+	}
+
+	c.wg.Done()
```
vadmeste: I suggest to call c.wg.Done() even when c.Conn.Close() returns an error here..

server-mux.go
```diff
@@ -115,47 +158,75 @@ func (c *MuxConn) Read(b []byte) (int, error) {
...
 	}
-	cmux := NewMuxConn(c)
+	cmux := NewMuxConn(c, l.wg)
 	protocol := cmux.PeekProtocol()
 	if protocol == "tls" {
 		return tls.Server(cmux, l.config), nil
 	}
+
+	// Increment the WaitGroup to enable graceful shutdown
+	l.wg.Add(1)
```
vadmeste: This is not called when a TLS connection is selected (`return tls.Server(cmux, l.config), nil`)

server-mux.go
```diff
@@ -115,47 +158,75 @@ func (c *MuxConn) Read(b []byte) (int, error) {
...
 // network stream with the TLS server
-func (l MuxListener) Accept() (net.Conn, error) {
+func (l *MuxListener) Accept() (net.Conn, error) {
 	c, err := l.Listener.Accept()
 	if err != nil {
 		return c, err
 	}
-	cmux := NewMuxConn(c)
+	cmux := NewMuxConn(c, l.wg)
 	protocol := cmux.PeekProtocol()
```
vadmeste, line 162: Currently, PeekProtocol is a blocking call.. you can open an idle telnet connection and see how even a force shutdown doesn't work

server-mux.go
```diff
@@ -115,47 +158,75 @@ func (c *MuxConn) Read(b []byte) (int, error) {
 	return c.Conn.Read(b)
 }
 
+// Close the connection and decrements the WaitGroup
+func (c *MuxConn) Close() error {
```
vadmeste: Are you sure that Close() is always called here? since it seems this is the only way to decrements Waitgroup

vadmeste:
@jesselucas, thanks for your work.. let's discuss further to see if we could enhance your change

server-mux.go
```diff
@@ -53,17 +55,58 @@ type MuxConn struct {
...
+			MaxHeaderBytes: 1 << 20,
+		},
+		wg:              new(sync.WaitGroup),
+		GracefulTimeout: 5 * time.Second,
+	}
+
+	m.Server.ConnState = func(c net.Conn, cs http.ConnState) {
+		if cs == http.StateIdle {
+			// Close any idle connections
+			if m.listener == nil {
```
vadmeste: I don't think that listener could be nil here in any case

jesselucas: Thanks for the feedback! I'll refactor and update soon.

harshavardhana:
> Thanks for the feedback! I'll refactor and update soon.

Can you rebase this patch?

server-mux.go
```diff
@@ -120,31 +122,43 @@ func (c *MuxConn) Read(b []byte) (int, error) {
...
 // network stream with the TLS server
-func (l MuxListener) Accept() (net.Conn, error) {
+func (l *MuxListener) Accept() (net.Conn, error) {
 	c, err := l.Listener.Accept()
 	if err != nil {
 		return c, err
 	}
+
+	// Wait till the listener is closed then cleanup
+	go l.cleanup(c)
```
vadmeste: One problem here is that it is creating many (infinite) go routines in the normal condition. Accept() is called for each new connection and cleanup() will just wait for shutdown signal.

server-mux.go
```diff
@@ -153,9 +167,62 @@ func (l MuxListener) Accept() (net.Conn, error) {
...
+}
+
+// Close Listener
+func (l *MuxListener) Close() error {
+	if l == nil {
+		return nil
+	}
+
+	// send nil to unblock gofunc to close l.Listener in Accept()
+	l.close <- nil
```
vadmeste: Actually, this doesn't guarantee that other go routines will be unblocked since you are receiving the 'nil' message just in the next line.

server-mux.go
```diff
@@ -153,9 +167,62 @@ func (l MuxListener) Accept() (net.Conn, error) {
...
 
+func (l *MuxListener) cleanup(c net.Conn) {
+	// Block until MuxServer close receives a value then cleanup listener
+	<-l.close
+
+	if c != nil {
+		c.Close()
+	}
+
+	l.close <- l.Listener.Close()
```
vadmeste: By the way, there is only one Listener instance and many Conn instances, we will have to close listener only once.

server-mux.go
jesselucas: https://play.golang.org/p/2iLaYXv1Ch

server-mux.go
jesselucas: Thank you! I see the issue. I was a little obsessed with closing connections. The point is to give them a chance to close themselves then close the listener within a certain amount of time. I'll refactor now.

server-mux.go
jesselucas: Great point! This should be removed with the refactor.

server-mux.go
```diff
@@ -153,9 +166,49 @@ func (l MuxListener) Accept() (net.Conn, error) {
 	return cmux, nil
 }
 
+// Close Listener
+func (l *MuxListener) Close() error {
+	if l == nil {
+		return nil
+	}
+
+	err := l.Listener.Close()
```
harshavardhana: This could be just 

```
return l.Listener.Close()
```

server-mux.go
```diff
@@ -153,9 +166,49 @@ func (l MuxListener) Accept() (net.Conn, error) {
...
+
+	// Return any error from l.Listener.Close().
+	return err
+}
+
 // MuxServer - the main mux server
 type MuxServer struct {
 	http.Server
+	listener        *MuxListener
+	wg              *sync.WaitGroup
```
harshavardhana: This can be made an upper case variable? 

```
WaitGroup *sync.WaitGroup
```

server-mux.go
```diff
@@ -153,9 +166,49 @@ func (l MuxListener) Accept() (net.Conn, error) {
...
+	m := &MuxServer{
+		Server: http.Server{
+			Addr: addr,
+			// Adding timeout of 10 minutes for unresponsive client connections.
+			ReadTimeout:    10 * time.Minute,
+			WriteTimeout:   10 * time.Minute,
+			Handler:        handler,
+			MaxHeaderBytes: 1 << 20,
+		},
+		wg:              new(sync.WaitGroup),
```
harshavardhana: You could use the consistent style of 

```
&sync.WaitGroup{}
```

We are largely deprecating the use of `new(sync.WaitGroup)`

server-mux.go
```diff
@@ -212,3 +280,83 @@ func longestWord(strings []string) int {
...
+				// Force close any idle and new connections.
+				if st == http.StateIdle || st == http.StateNew {
+					c.Close()
+				}
+			}
+			m.mu.Unlock()
+
+			// Wait for all connections to be gracefully closed
+			m.wg.Wait()
+			close(wait)
```
harshavardhana: This can be 'defer close(wait)' at the top.

server-mux.go
```diff
@@ -212,3 +280,83 @@ func longestWord(strings []string) int {
...
+		}()
+
+		select {
+		case <-time.After(m.GracefulTimeout):
+			return nil
+		case <-wait:
+			return nil
+		}
+	} else {
```
harshavardhana: I guess the logic would be reverted to avoid the 'if block' indentation. 

```
if m.closed {
   return errors.New("Server has been closed")
}
...
... the rest
```

server-mux.go
```diff
@@ -212,3 +280,83 @@ func longestWord(strings []string) int {
...
+		return errors.New("Server has been closed")
+	}
+}
+
```
harshavardhana, line 320: Needs a comment.

server-mux.go
```diff
@@ -212,3 +280,83 @@ func longestWord(strings []string) int {
...
+func (m *MuxServer) connState() {
+	oldHook := m.Server.ConnState
+	m.Server.ConnState = func(c net.Conn, cs http.ConnState) {
+		m.mu.Lock()
+		defer m.mu.Unlock()
+
+		switch cs {
```
harshavardhana, line 328: This entire block would need commenting since this is a critical portion of state handling.

jesselucas, line 328: ConnState hook was used to store any previously setup ConnState function and then call it from within the new one. I removed this since it doesn't seem like anyone, other than this function, will be adjusting the ConnState.

jesselucas:
@harshavardhana Updated. Let me know if you see anything else. 🎉

Issue #2252, opened 2016-07-20
Feature: lock XL disks to forbid reading/writing


vadmeste:
That will be necessar for fsck.

harshavardhana:
Can you add more gist here @vadmeste ?

vadmeste:
fsck will be an independant tool that will check the consistency of the XL data. For that, Minio server should be stopped. However we need to protect data from all malicious and mistaken actions, for example, running a new minio while fsck is checking etc..

Also probably fsck will change data in the XL disks themselves, and there is no safe way to do that only by stopping Minio.

vadmeste:
We need a proper shutdown mechanism for this.. could we discuss to see if we will really need that shutdown mechanism for other purposes? (cleaning stale files, etc..)

vadmeste:
Timeline: 5 August

harshavardhana:
> That will be necessar for fsck.

Locking disks and fsck is out of scope for Minio.  Since most of the time we are non-root we will never be able to disallow access to disks. 

Closing this as not a bug.

PR #2344, opened 2016-08-02
Add a generic registerShutdown function for graceful exit

vadmeste:


utils.go
```diff
@@ -73,3 +75,42 @@ func contains(stringList []string, element string) bool {
...
+		trapCh := signalTrap(os.Interrupt, syscall.SIGTERM)
+		for {
+			select {
+			case <-trapCh:
+				// Start a graceful shutdown call
+				shutdownSignal <- true
+			case <-shutdownSignal:
+				// Call all callbacks and exit for emergency
+				for _, callback := range shutdownCallbacks {
+					i := callback()
```
harshavardhana: what is 'i' here?  shouldn't use 'i' use 'err' to be conventionally correct.

fs-v1.go
```diff
@@ -133,8 +133,8 @@ func newFSObjects(disk string) (ObjectLayer, error) {
 	}
 
 	// Register the callback that should be called when the process shuts down.
-	registerShutdown(func() {
-		shutdownFS(storage)
+	registerShutdown(func() int {
```
harshavardhana: the int should take a type 

```
type errCode int

const (
exitFailure errCode = -1
exitSuccess errCode = 0 
)
```

utils.go
vadmeste: yes, it is just because shutdownFS() was exiting using a return code and I didn't want to break things.. shall we rename `i` to `exitCode` or we use `nil` and `error` instead ?

fs-v1.go
vadmeste: :+1:

vadmeste:
@harshavardhana, I think we need a specific order when executing callbacks.. like gracefully shutdown all webservices first, then cleaning stale files, etc.. right?

harshavardhana:
> @harshavardhana, I think we need a specific order when executing callbacks.. like gracefully shutdown all webservices first, then cleaning stale files, etc.. right?

Yes that is correct that is the order.

vadmeste:
Data synchronization process will be the last step before shutdown. Also, the order of executing callbacks will be the same as the order of registering callbacks, so the caller needs to take care of that too.

vadmeste:
@harshavardhana is this ready for merge or you have some comments?

This is one step to fix #2252

harshavardhana:
Can you rebase this patch @vadmeste ?

harshavardhana:
Need to add some test cases here @vadmeste

PR #2348, opened 2016-08-03
[wip][update] fs: Background commit in complete multipart upload

### LINKED ISSUE #2334: Don't concatenate multi-part upload during completion
The FS version concatenates all the parts of a multi-part upload when the multi-upload is completed. Multi-part uploads are often used for huge files, so this takes a lot of time and also spikes the IOPS for a while. In some situations (like ours), it would be much better to use the XL method to keep the parts separate and combine them during download.

Of course, the parts cannot be accessed the regular way (via the filesystem) anymore. But most systems that use S3 only use S3 to access them (you can't do otherwise in AWS), so this wouldn't be a big problem.

Concatenating files might be the sensible default option, but when performance and I/O is critical, then keeping the files separate might be the better option. Now, we reverted using the XL method, solely for this feature, but this costs us a lot of additional disk space.

vadmeste:
Some shortcomings with this patch:
*) the object will be not seen just after upload until the background commit will terminate

Fixes #2334

### DIFF
--- a/cmd/faulty-disk_test.go
+++ b/cmd/faulty-disk_test.go
@@ -16,6 +16,8 @@
 
 package cmd
 
+import "time"
+
 // Simulates disk returning errFaultyDisk on all methods of StorageAPI
 // interface after successCount number of successes.
 type faultyDisk struct {
@@ -112,4 +114,12 @@ func (f *faultyDisk) ReadAll(volume string, path string) (buf []byte, err error)
 		return f.disk.ReadAll(volume, path)
 	}
 	return nil, errFaultyDisk
+}
+
+func (f *faultyDisk) Chtimes(volume, path string, ctime, atime time.Time) (err error) {
+	return nil
+}
+
+func (f *faultyDisk) PrepareFile(volume, path string, size int64) (err error) {
+	return nil
 }
--- a/cmd/fs-v1-multipart.go
+++ b/cmd/fs-v1-multipart.go
@@ -521,12 +521,9 @@ func (fs fsObjects) CompleteMultipartUpload(bucket string, object string, upload
 		return "", err
 	}
 
-	tempObj := path.Join(tmpMetaPrefix, uploadID+"-"+"part.1")
+	var totalSize int64
 
-	// Allocate staging buffer.
-	var buf = make([]byte, readSizeV1)
-
-	// Loop through all parts, validate them and then commit to disk.
+	// Loop through all parts to validate them before starting the background commit
 	for i, part := range parts {
 		partIdx := fsMeta.ObjectPartIndex(part.PartNumber)
 		if partIdx == -1 {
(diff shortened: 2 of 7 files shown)


vadmeste:
Also, one test unit case is removed, which was testing is an upload id is invalid after a successful multipart complete (since old upload ids are removed), we should not check this anymore since removing the upload id is put in a goroutine

krishnasrinivas:
> we should not check this anymore since removing the upload id is put in a goroutine

you can Sleep() before checking because the test case is valid.

vadmeste:
> you can Sleep() before checking because the test case is valid.

yes we can, but it is just how much time it will sleep.. 1 seconds, 10 seconds? on travis it could be 1 min, that's why I am suggested to remove this one even if it is an important one, anyway let's discuss the overall feature before we talk about this particular point.

fs-v1-multipart.go
```diff
@@ -542,91 +537,132 @@ func (fs fsObjects) CompleteMultipartUpload(bucket string, object string, upload
...
 		multipartPartFile := path.Join(mpartMetaPrefix, bucket, object, uploadID, partSuffix)
-		offset := int64(0)
-		totalLeft := fsMeta.Parts[partIdx].Size
-		for totalLeft > 0 {
-			curLeft := int64(readSizeV1)
-			if totalLeft < readSizeV1 {
-				curLeft = totalLeft
+		_, err = fs.storage.StatFile(minioMetaBucket, multipartPartFile)
+		if err != nil {
+			return "", err
```
harshavardhana: You need to see a different error here `InvalidPart{}`

fs-v1-multipart.go
```diff
@@ -542,91 +537,132 @@ func (fs fsObjects) CompleteMultipartUpload(bucket string, object string, upload
...
+		}
+
+	}
+
+	// Background commit
+	go func() {
```
harshavardhana: Keep this in a separate function, take inputs for all the parts.

fs-v1-multipart.go
```diff
@@ -542,91 +537,132 @@ func (fs fsObjects) CompleteMultipartUpload(bucket string, object string, upload
...
+				//FIXME: alert if needed
 			}
-			var n int64
-			n, err = fs.storage.ReadFile(minioMetaBucket, multipartPartFile, offset, buf[:curLeft])
-			if n > 0 {
-				if err = fs.storage.AppendFile(minioMetaBucket, tempObj, buf[:n]); err != nil {
-					return "", toObjectErr(err, minioMetaBucket, tempObj)
+		}()
+		// Hold lock in the uploaded object until background commiting terminates
+		nsMutex.Lock(bucket, object)
```
harshavardhana: No need to hold lock until 'rename'.

harshavardhana:
Can you rebase this patch @vadmeste ?

fs-v1-multipart.go
```diff
@@ -542,6 +539,48 @@ func (fs fsObjects) CompleteMultipartUpload(bucket string, object string, upload
...
+		}
+
+		totalSize += fsMeta.Parts[partIdx].Size
+	}
+
+	// Check if we have enough disk space to make the commit. The parts size + the target file (totalSize*2)
+	// should respect the minimum free space requirement.
+	storageInfo := fs.StorageInfo()
+	if storageInfo.Free+totalSize*2 < storageInfo.Total*(1-fsMinSpacePercent) {
+		return "", errDiskFull
```
harshavardhana:
```
toObjectErr(errDiskFull, bucket, object) 
```

harshavardhana:
@vadmeste is this still a WIP?

vadmeste:
@harshavardhana, no, will do your notice and sync again

vadmeste:
Do not merge yet, one problem with this patch is that it is modifying original file (os.Truncate) during complete multipart

harshavardhana:
@vadmeste  we decided to do a separate backend for this by making XL itself support Single disk and multi disk.

Single disk means -- no erasure but all other things are available (saving metadata and bit-rot)
Multiple disks means -- all things same as is.

vadmeste:
Brilliant simplification!

harshavardhana:
This fix is not needed anymore.. @vadmeste you can close.

PR #2351, opened 2016-08-03
webapi: ServerInfo returns the list variables/values in the environnm…

vadmeste:
…ent of the Minio server

We will need this later in Minio Browser to know if we should forbid the change of access and secret keys if they are already set in the minio environment

### DIFF
--- a/web-handlers.go
+++ b/web-handlers.go
@@ -74,6 +74,7 @@ type ServerInfoRep struct {
 	MinioMemory   string
 	MinioPlatform string
 	MinioRuntime  string
+	MinioEnvVars  []string
 	UIVersion     string `json:"uiVersion"`
 }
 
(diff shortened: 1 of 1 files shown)


web-handlers.go
```diff
@@ -98,8 +99,13 @@ func (web *webAPIHandlers) ServerInfo(r *http.Request, args *WebGenericArgs, rep
 		runtime.GOOS,
 		runtime.GOARCH)
 	goruntime := fmt.Sprintf("Version: %s | CPUs: %s", runtime.Version(), strconv.Itoa(runtime.NumCPU()))
+	envs := make([]string, 0)
```
harshavardhana: This is frivolous `var envs []string` is sufficient.

vadmeste: > This is frivolous var envs []string is sufficient.

yes.. even `reply.MinioEnvVars = os.Environ()` is enough

harshavardhana:
> yes.. even reply.MinioEnvVars = os.Environ() is enough

:+1:

Issue #2335, opened 2016-08-01
server: validate whether server supports bucket policy with '?' character.


balamurugana:
As per Amazon S3 spec, resource may have `?` character to represent matching any single character.  We would need to validate whether its supported in the server currently.

hackintoshrao:
The spec [here](http://docs.aws.amazon.com/AmazonS3/latest/dev/s3-arn-format.html) says 

> This ARN uses both wildcards, '_', and '?', in the relative-ID part. It identifies all objects in buckets such as example1bucket, example2bucket, example3bucket and so on.
> arn:aws:s3:::example?bucket/_

Adding support for `?`

hackintoshrao:
This is fixed, can be closed now @balamurugana .

PR #2353, opened 2016-08-04
Bucket policy: Support for '?' wildcard.

hackintoshrao:
- Support for '?' wildcard for resource matching.
- Tests for the same for the wide range of cases.


hackintoshrao:
Fixes https://github.com/minio/minio/issues/2335

bucket-policy-parser_test.go
```diff
@@ -690,3 +690,500 @@ func TestParseBucketPolicy(t *testing.T) {
...
+	testCases := []struct {
+		pattern string
+		text    string
+		matched bool
+	}{
+		// Test case - 1.
+		// Test case with pattern "*". Expected to match any text.
+		{
+			pattern: "*",
+			text:    "s3:GetObject",
```
harshavardhana: Is `wildCardResourceMatch` ever used for Actions?

PR #2358, opened 2016-08-04
getbucketlocation: should handle UNSIGNED-PAYLOAD for sha256 header for signature calculation.

### LINKED ISSUE #2355: minio complains over UNSIGNED-PAYLOAD on secure server
## Expected behaviour

I might be wrong :) but I'd expect that `presignedGetObject` requests receive the same response from minio when properly authenticated.
## Actual behavior

Minio returns a `ErrContentSHA256Mismatch` ("The provided 'x-amz-content-sha256' header does not match what was computed.") when minio-js makes a request to do `presignedGetObject` on a secure server. In this case, sha256 summing should be disabled and the node client sends `UNSIGNED-PAYLOAD` as the header.

Minio works correctly on a server that isn't secure, which triggers the sha256 sum to take place.
## Steps to reproduce the behavior

This can be tested between `play.minio.io` (secure) and a local, out-of-the-box minio install. Run this code to see the difference between the two, adjusting authentication parameters as needed.

``` javascript
const Minio = require('minio');

/*var client = new Minio({
    endPoint: 'play.minio.io',
    port: 9000,
    secure: true,
    accessKey: 'Q3AM3UQ867SPQQA43P2F',
    secretKey: 'zuf+tfteSlswRu7BJ86wekitnifILbZam1KYY3TG'
});*/

/*var client = new Minio({
    endPoint: 'localhost',
    port: 9000,
    secure: false,
    ...
});*/

client.presignedGetObject('pictures', 'house.png', 100, function(err, url) {
    if (err) throw err;
    console.log(url);
});

```
## Minio version
- `play.minio.io` = `2016-07-28T03:19:46Z`
- `localhost` = `2016-08-03T07:43:43Z`
## System information

Unsure about the minio server, but locally it's a Macbook running OSX 
10.11.6.

krishnasrinivas:
fixes #2355

### DIFF
--- a/bucket-handlers.go
+++ b/bucket-handlers.go
@@ -17,12 +17,8 @@
 package main
 
 import (
-	"bytes"
-	"encoding/base64"
-	"encoding/hex"
 	"encoding/xml"
 	"io"
-	"io/ioutil"
 	"net/http"
 	"net/url"
 	"path"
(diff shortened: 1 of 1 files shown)


harshavardhana:
LGTM, Thanks :+1:

PR #2362, opened 2016-08-05
tests: tests for signature v4 parser.

hackintoshrao:


signature-v4-parser_test.go
```diff
@@ -0,0 +1,427 @@
+/*
+ * Minio Cloud Storage, (C) 2015 Minio, Inc.
```
harshavardhana: License header should be this year.

signature-v4-parser_test.go
```diff
@@ -0,0 +1,427 @@
...
+		},
+	}
+
+	for i, testCase := range testCases {
+		actualSignedHeaders, actualErrCode := parseSignedHeaders(testCase.inputSignElement)
+		if strings.Join(testCase.expectedSignedHeaders, ",") != strings.Join(actualSignedHeaders, ",") {
+			t.Errorf("Test %d: Expected the result to be \"%v\", but got \"%v\". ", i+1, testCase.expectedSignedHeaders, actualSignedHeaders)
+
+		}
+		if testCase.expectedErrCode != actualErrCode {
```
harshavardhana, line 287: shouldn't this be validated first? becasue 'actualSignedHeaders' would be 'nil' for actualErrCode is not ErrNone.

hackintoshrao, line 287: I'm treating both result and error as values here . Works good those `nil` cases, but will make a check to assert for values only if err == errNone.

PR #2373, opened 2016-08-06
Add ServerCtrl web api for shutdown

vadmeste:
Simple web api for shutdown/restart

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -21,6 +21,7 @@ import (
 	"encoding/xml"
 	"io"
 	"os"
+	"os/exec"
 	"strings"
 	"syscall"
 )
@@ -76,9 +77,16 @@ func contains(stringList []string, element string) bool {
 	return false
 }
 
-// shutdownSignal - is the channel that receives any boolean when
-// we want broadcast the start of shutdown
-var shutdownSignal chan bool
+// shutdownSignal to specify the shutdown policy (halt/restart)
+type shutdownSignal string
+
+const (
+	shutdownHalt    shutdownSignal = "halt"
+	shutdownRestart                = "restart"
+)
+
+// shutdownSignal - is the channel that receives the shutdown order
+var shutdownSignalCh chan shutdownSignal
 
 // shutdownCallbacks - is the list of function callbacks executed one by one
 // when a shutdown starts. A callback returns 0 for success and 1 for failure.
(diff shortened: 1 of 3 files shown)


utils.go
```diff
@@ -127,6 +135,21 @@ func monitorShutdownSignal() {
...
+				if signal == shutdownRestart {
+					path := os.Args[0]
+					cmdArgs := os.Args[1:]
+					cmd := exec.Command(path, cmdArgs...)
+					cmd.Stdout = os.Stdout
+					cmd.Stderr = os.Stderr
+
+					err := cmd.Start()
+					if err != nil {
+						// FIXME: log to the appropriate engine
```
harshavardhana, line 149: you can use `errorIf(err,...)`

utils.go
```diff
@@ -127,6 +135,21 @@ func monitorShutdownSignal() {
 					}
 
 				}
+				// All shutdown callbacks ensure that the server is safely terminated
+				// and any concurrent process could be started again
+				if signal == shutdownRestart {
+					path := os.Args[0]
+					cmdArgs := os.Args[1:]
+					cmd := exec.Command(path, cmdArgs...)
```
harshavardhana, line 143: Have you looked at saving the known FD's and restarting? - https://golang.org/pkg/os/#StartProcess

vadmeste, line 143: @harshavardhana, why will we need that ?

harshavardhana, line 143: > @harshavardhana, why will we need that ?

you need to persist the env for the existing process.  Does this support seamless migration of env's, pwd etc?

web-handlers.go
```diff
@@ -68,6 +68,33 @@ type WebGenericRep struct {
...
+	Status    string
+	UIVersion string `json:"uiVersion"`
+}
+
+// ServerCtrlArgs - server control args
+type ServerCtrlArgs struct {
+	Cmd string `json:"Cmd"`
+}
+
+// ServerInfo - get server info.
```
harshavardhana, line 82: Fix the comment.

utils.go
harshavardhana, line 143: > you need to persist the env for the existing process. Does this support seamless migration of env's, pwd etc?

Also when the browser sends the signal how seamless does JSON-RPC calls work? don't see any explanation in the patch what is the preceding nature. Is this a precursor to 'minio control <command>' ?

krishnasrinivas:
@vadmeste the shutdown handler will have to be registered with net/rpc server (not json-rpc) , i will push the code for healing handler by sunday. you can extend it. sorry for the delay.

utils.go
vadmeste, line 143: > you need to persist the env for the existing process. Does this support seamless migration of env's, pwd etc?

yes, the same env variables are still accessible after restart, even FDs and already sockets could be transfered too if we will need that later

> Also when the browser sends the signal how seamless does JSON-RPC calls work? don't see any explanation in the patch what is the preceding nature. Is this a precursor to 'minio control <command>' ?

sorry I didn't mention that but as you said, it is an introduction to minio control command

PR #2381, opened 2016-08-07
contoller: Implement controlled healing and trigger 

krishnasrinivas:
This patch introduces new command line 'control' 
- minio control 

TO manage minio server connecting through GoRPC API frontend.
- minio control heal 

Is implemented for healing objects.


control-main.go
```diff
@@ -27,32 +27,86 @@ import (
...
+	fatalIf(err, "Unable to parse URL")
 
-	_, err = rpc.DialHTTPPath("tcp", parsedURL.Host, healPath)
+	path := parsedURL.Path
+	if path == "" || path == slashSeparator {
+		cli.ShowCommandHelpAndExit(c, "heal", 1)
+	}
+
+	path = path[1:]
+	slashIndex := strings.Index(path, slashSeparator)
```
harshavardhana: make this a function.

control-main.go
```diff
@@ -27,32 +27,86 @@ import (
...
+		bucket = path[:slashIndex]
+		object = path[slashIndex+1:]
+	}
+	client, err := rpc.DialHTTPPath("tcp", parsedURL.Host, healPath)
 	fatalIf(err, "error connecting to %s", parsedURL.Host)
+
+	if object != "" && !strings.HasSuffix(object, slashSeparator) {
+		args := &HealObjectArgs{bucket, object}
+		reply := &HealObjectReply{}
+		err := client.Call("Heal.HealObject", args, reply)
```
harshavardhana: error shadowing here..

rpc-control.go
```diff
@@ -35,6 +35,26 @@ type HealListReply struct {
...
+		return err
+	}
+	reply.IsTruncated = info.IsTruncated
+	reply.NextMarker = info.NextMarker
+	for _, obj := range info.Objects {
+		reply.Objects = append(reply.Objects, obj.Name)
+	}
 	return nil
 }
+
```
harshavardhana, line 72: Comments..

krishnasrinivas:
@harshavardhana known issues:
 there are a couple of FIXMEs

```
217 // FIXME: If an object object was deleted and one disk was down, and later the disk comes back      
218 // up again, heal on the object should delete it.           
```

```
// FIXME: till net/rpc auth is brought in "minio control" can be enabled only though this env variable
```

HealObject(): should return the status of the file - on how many disks it is available.
HealList(): returns all the objects, I think we can return only objects that needs healing as there is no use of returning all the objects. HealList will again have the random-io problem because of stat of xl.json, need to see how it can be fixed.

Output of "minio control heal" is crude now. Needs to be fixed.

Add unit test cases.

control-main.go
```diff
@@ -0,0 +1,132 @@
...
+	"net/rpc"
+	"net/url"
+
+	"github.com/minio/cli"
+)
+
+// "minio control" command.
+var controlCmd = cli.Command{
+	Name:   "control",
+	Usage:  "minio control commands.",
```
harshavardhana: "Control and manage minio server."

control-main.go
```diff
@@ -0,0 +1,132 @@
...
+	Action: healControl,
+	CustomHelpTemplate: `NAME:
+  minio {{.Name}} - {{.Usage}}
+
+USAGE:
+  minio {{.Name}} [OPTIONS] URL
+
+EAMPLES:
+  1. Heal an object.
+     $ minio control heal http://localhost:9000/songs/classical/western/piano.mp3
```
harshavardhana, line 55: 'minio {{.Name}} heal'

control-main.go
```diff
@@ -0,0 +1,132 @@
...
+		if slashIndex == -1 {
+			bucket = path
+		} else {
+			bucket = path[:slashIndex]
+			object = path[slashIndex+1:]
+		}
+		return bucket, object
+	}
+
+	if len(c.Args()) != 1 {
```
harshavardhana, line 85: `if !c.Args().Present() {`

control-main.go
```diff
@@ -0,0 +1,132 @@
...
+  3. Heall all objects with a given prefix recursively.
+     $ minio control heal http://localhost:9000/songs/classical/
+`,
+}
+
+// "minio control heal" entry point.
+func healControl(c *cli.Context) {
+	// Parse bucket and object from url.URL.Path
+	parseBucketObject := func(path string) (bucket string, object string) {
+		path = path[1:]
```
harshavardhana: This can panic if path is empty string wouldn't it be simpler to use

```
        splits := strings.SplitN(path, string(slashSeparator), 3)
        switch len(splits) {
        case 0, 1:
                bucketName = ""
                objectName = ""
        case 2:
                bucketName = splits[1]
                objectName = ""
        case 3:
                bucketName = splits[1]
                objectName = splits[2]
        }
```

Rather than manipulating string indexes?

control-main.go
```diff
@@ -0,0 +1,132 @@
...
+
+	path := parsedURL.Path
+	if path == "" || path == slashSeparator {
+		cli.ShowCommandHelpAndExit(c, "heal", 1)
+	}
+
+	bucket, object = parseBucketObject(path)
+
+	client, err := rpc.DialHTTPPath("tcp", parsedURL.Host, healPath)
+	fatalIf(err, "error connecting to %s", parsedURL.Host)
```
harshavardhana: "error connecting to %s" --> "Unable to connect to %s"

erasure-healfile.go
```diff
@@ -0,0 +1,83 @@
...
+		}
+
+		// Calculate the block size that needs to be read from each disk.
+		curEncBlockSize := getChunkSize(curBlockSize, dataBlocks)
+
+		// Memory for reading data from disks and reconstructing missing data using erasure coding.
+		enBlocks := make([][]byte, len(latestDisks))
+
+		// Read data from the latest disks.
+		// FIXME: no need to read from all the disks. dataBlocks+1 is enough.
```
harshavardhana, line 42: Looks like this code is not parallelized yet.

object-errors.go
```diff
@@ -257,3 +257,10 @@ type PartTooSmall struct {
 func (e PartTooSmall) Error() string {
 	return fmt.Sprintf("Part size for %d should be atleast 5MB", e.PartNumber)
 }
+
+// NotImplemented If a feature is not implemented
+type NotImplemented struct{}
+
+func (e NotImplemented) Error() string {
+	return "Not Implemented"
```
harshavardhana, line 265: It would be better "API Not Implemented"

rpc-control.go
```diff
@@ -0,0 +1,69 @@
+package main
```
harshavardhana, line 17: license header.

xl-v1-object.go
```diff
@@ -213,6 +213,134 @@ func (xl xlObjects) GetObject(bucket, object string, startOffset int64, length i
...
+
+	for index, disk := range outDatedDisks {
+		// Before healing outdated disks, we need to remove xl.json
+		// and part files from "bucket/object/" so that
+		// rename(".minio.sys", "tmp/tmpuuid/", "bucket", "object/") succeeds.
+		if disk == nil {
+			// Not an outdated disk.
+			continue
+		}
+		if errs[index] != nil {
```
harshavardhana, line 259: Isn't the

`errs[index] != nil` already gives an outdated disk? according to xl.outDateDisks logic?

xl-v1-object.go
```diff
@@ -213,6 +213,134 @@ func (xl xlObjects) GetObject(bucket, object string, startOffset int64, length i
...
+			// If there was an error (most likely errFileNotFound)
+			continue
+		}
+		// Outdated object with the same name exists that needs to be deleted.
+		outDatedMeta := partsMetadata[index]
```
harshavardhana, line 264: We have to do this in parallel.

erasure-healfile.go
```diff
@@ -0,0 +1,83 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package main
+
+import "encoding/hex"
+
+func erasureHealFile(latestDisks []StorageAPI, outDatedDisks []StorageAPI, volume, path, healBucket, healPath string, size int64, blockSize int64, dataBlocks int, parityBlocks int, algo string) (checkSums []string, err error) {
```
harshavardhana, line 22: Needs comment.

control-main.go
krishnasrinivas, line 85: Right now it works for only one command

control-main.go
krishnasrinivas: I do a check:

```
89         if path == "" || path == slashSeparator {                                                                                                                            
90                 cli.ShowCommandHelpAndExit(c, "heal", 1)                                                                                                                     
91         }     
```

but it's cleaner and selfcontained i.e doesn't depend on this check.

erasure-healfile.go
krishnasrinivas, line 42: yes, since healing can be slow i have left it like this, will fix it in subsequent patches. it was written before parallelRead() was implemented, will check how i can reuse it.

object-errors.go
krishnasrinivas, line 265: It can be used to for features not implemented too

xl-v1-object.go
krishnasrinivas, line 259: in this loop the object of an older version is getting deleted, so if `errs[index] != nil` we do not try to delete the old object as it does not exist

xl-v1-object.go
krishnasrinivas, line 264: i have kept everything serial since healing can be slow, will incrementally make it parallel. wanted to get healing working first so kept things simple.

xl-v1-object.go
harshavardhana, line 259: Wouldnt it be explicit check for that. Otherwise this logic assumes that it is always errFileNotFound

krishnasrinivas, line 259: > Wouldnt it be explicit check for that. Otherwise this logic assumes that it is always errFileNotFound

did not get it, this logic does not assume that it's always errFileNotFound, can you check it again?

harshavardhana, line 259:
```
+           // If there was an error (most likely errFileNotFound)
 +          continue
 +      }
```

This comment most likely.

krishnasrinivas, line 259: yes in case of errFileNotFound or any other error like errDiskNotFound we don't try to delete the outdated files on it because if we do we will get error.

PR #2389, opened 2016-08-09
handlers: Remove 'notification.xml' when bucket is deleted.

harshavardhana:
Do not pass around objectHandlers object, input argument
should comply to a type for only that would be used inside
the function body.

### DIFF
--- a/bucket-handlers.go
+++ b/bucket-handlers.go
@@ -28,9 +28,9 @@ import (
 )
 
 // http://docs.aws.amazon.com/AmazonS3/latest/dev/using-with-s3-actions.html
-func enforceBucketPolicy(api objectAPIHandlers, action string, bucket string, reqURL *url.URL) (s3Error APIErrorCode) {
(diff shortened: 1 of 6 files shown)


bucket-handlers.go
```diff
@@ -30,7 +30,7 @@ import (
 // http://docs.aws.amazon.com/AmazonS3/latest/dev/using-with-s3-actions.html
 func enforceBucketPolicy(api objectAPIHandlers, action string, bucket string, reqURL *url.URL) (s3Error APIErrorCode) {
```
krisis: It would be uniform if all functions that work on bucket policy take ObjectLayer interface as an argument. i.e,

```
func enforceBucketPolicy(obj ObjectLayer, action string, bucket string, reqURL *url.URL) (s3Error APIErrorCode)
```

harshavardhana: Ah i see

harshavardhana: > It would be uniform if all functions that work on bucket policy take ObjectLayer interface as an argument. i.e,

There is another structural problem in this code, will need to fix it after this.

PR #2392, opened 2016-08-09
posix: bug fix: should return <nil, err> on error.

hackintoshrao:


hackintoshrao:
The top level decides the further action, no need for change at posix layer.

PR #2394, opened 2016-08-09
tests: add tests for certs.go and trie.go

brendanashworth:
Just adds some really basic tests. It doesn't test everything in both files, but its a start.

### DIFF
--- a/certs_test.go
+++ b/certs_test.go
@@ -1,0 +1,60 @@
+/*
+ * Minio Cloud Storage, (C) 2015, 2016 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package main
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+// Make sure we have a valid certs path.
+func TestGetCertsPath(t *testing.T) {
+	path, err := getCertsPath()
+	if err != nil {
+		t.Error(err)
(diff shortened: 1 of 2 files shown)


brendanashworth:
Windows doesn't like this 😢 will fix.

certs_test.go
```diff
@@ -0,0 +1,49 @@
...
+ */
+
+package main
+
+import (
+	"strings"
+	"testing"
+)
+
+func TestGetCertsPath(t *testing.T) {
```
harshavardhana, line 27: Comments are missing.

certs_test.go
```diff
@@ -0,0 +1,49 @@
...
+	if path == "" {
+		t.Fail()
+	}
+	if !strings.ContainsRune(path, '/') {
+		t.Fail()
+	}
+
+	mustGetCertsPath()
+}
+
```
harshavardhana, line 47: Comment..

trie_test.go
```diff
@@ -0,0 +1,61 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package main
+
+import (
+	"testing"
+)
+
```
harshavardhana, line 22: Write comments for each.

certs_test.go
```diff
@@ -0,0 +1,49 @@
...
+import (
+	"strings"
+	"testing"
+)
+
+func TestGetCertsPath(t *testing.T) {
+	path, err := getCertsPath()
+	if err != nil {
+		t.Fail()
```
harshavardhana: Make sure to fail with a proper message.

harshavardhana: Same everywhere.

certs_test.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+	if certsPath == "" {
+		t.Errorf("expected path to not be an empty string, got: '%s'", certsPath)
+	}
+	// Ensure it contains some sort of path separator.
+	if !strings.ContainsRune(certsPath, os.PathSeparator) {
+		t.Errorf("expected path to contain file separator")
+	}
+	// It should also be an absolute path.
+	if !path.IsAbs(certsPath) {
```
harshavardhana: you have to use 'filepath.IsAbs()' since 'getCertsPath' returns a windows specific path.

brendanashworth: I'm just guessing now because I don't have any windows machines :) thanks for the tip, I'll fix that.

PR #2396, opened 2016-08-09
Object layer initialization using distributed locking

krisis:
The following operations work,
- `mc mb dist/krisis`
- `mc cp small-file dist/krisis`
- `mc cat dist/krisis/small-file`

Known issues
- Multipart upload fails with "Unable to calculate checksum error", while the data is uploaded fine.

### DIFF
--- a/routers.go
+++ b/routers.go
@@ -49,6 +49,11 @@ func newObjectLayerFactory(disks, ignoredDisks []string) func() ObjectLayer {
 		if objAPI != nil {
 			return objAPI
 		}
+
+		// Acquire a distributed lock to ensure only one of the nodes
+		// initializes the format.json.
+		nsMutex.Lock(minioMetaBucket, formatConfigFile)
+		defer nsMutex.Unlock(minioMetaBucket, formatConfigFile)
 		objAPI, err = newObjectLayer(disks, ignoredDisks)
 		if err != nil {
 			return nil
--- a/routers.go
+++ b/routers.go
@@ -32,6 +32,7 @@ func newObjectLayer(disks, ignoredDisks []string) (ObjectLayer, error) {
 		// Initialize FS object layer.
 		return newFSObjects(exportPath)
 	}
+	// TODO: use dsync to block other concurrently booting up nodes.
 	// Initialize XL object layer.
 	objAPI, err := newXLObjects(disks, ignoredDisks)
 	if err == errXLWriteQuorum {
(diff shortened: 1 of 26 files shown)


routers.go
```diff
@@ -40,41 +41,56 @@ func newObjectLayer(disks, ignoredDisks []string) (ObjectLayer, error) {
 	return objAPI, err
 }
 
+func newObjectLayerFactory(disks, ignoredDisks []string) func() ObjectLayer {
+	var objAPI ObjectLayer
+	// FIXME: This needs to be go-routine safe.
+	return func() ObjectLayer {
+		var err error
+		if objAPI != nil {
```
harshavardhana, line 49: what is this for? how is this expected to be initialized?

routers.go
```diff
@@ -40,41 +41,56 @@ func newObjectLayer(disks, ignoredDisks []string) (ObjectLayer, error) {
...
+			return objAPI
+		}
+		objAPI, err = newObjectLayer(disks, ignoredDisks)
+		if err != nil {
+			return nil
```
harshavardhana, line 59: log the error whatever happened here..

rpc-server.go
```diff
@@ -111,22 +130,50 @@ func (s *storageServer) RenameFileHandler(arg *RenameFileArgs, reply *GenericRep
 }
 
 // Initialize new storage rpc.
-func newRPCServer(exportPath string) (*storageServer, error) {
+func newRPCServer(serverConfig serverCmdConfig) (servers []*storageServer, err error) {
```
harshavardhana, line 133: Can you add 'disks, ignoreDisks' ? rather than 'serverConfig?' - also there is a struct which is 'serverConfig' which would conflict here.

xl-v1.go
```diff
@@ -124,6 +124,10 @@ func newXLObjects(disks, ignoredDisks []string) (ObjectLayer, error) {
 		// to handle these errors internally.
 		storageDisks[index], err = newStorageAPI(disk)
 		if err != nil && err != errDiskNotFound {
+			switch diskType := storageDisks[index].(type) {
+			case networkStorage:
+				diskType.rpcClient.Close()
```
harshavardhana, line 129: Return an error that happened here..

harshavardhana:
Let's not do a master branch patch just yet,  send this patch to 'distributed branch' instead.

krisis:
I have sent a PR on top of distributed branch. Closing this PR.

PR #2401, opened 2016-08-10
wip: unify single-node and distributed namespace locking

krisis:


namespace-lock.go
```diff
@@ -30,37 +76,36 @@ type nsParam struct {
 
 // nsLock - provides primitives for locking critical namespace regions.
 type nsLock struct {
-	*dsync.DRWMutex
-	ref uint
+	rwlock RWLocker
```
harshavardhana: You can keep it like before 

```
type nsLock struct {
    RWLocker
    ref uint 
}
```

namespace-lock.go
```diff
@@ -30,37 +76,36 @@ type nsParam struct {
...
+	var nsLk *nsLock
 	n.mutex.Lock()
 
 	param := nsParam{volume, path}
 	nsLk, found := n.lockMap[param]
 	if !found {
-		nsLk = &nsLock{
-			DRWMutex: dsync.NewDRWMutex(volume + path),
-			ref:      0,
+		if n.isDist {
```
harshavardhana: You can simplify this by doing 

```
nsLk = &nsLock{
     RWLocker: func() RWLocker {
                  if n.isDist {
                         return dsync.NewDRWMutex(path.Join(volume, path))
                  }
                  return &sync.RWMutex{}
          },
          ref: 0,
}
```

harshavardhana: Just more nicer looking and avoids the top level 'var nsLk *nsLock'

test-utils_test.go
```diff
@@ -43,7 +43,7 @@ import (
 // Tests should initNSLock only once.
 func init() {
 	// Initialize name space lock.
-	initNSLock()
+	initNSLock(false)
```
harshavardhana: Use a separate variable 

```
distributed := false
initNSLock(distributed)
```

namespace-lock.go
```diff
@@ -30,37 +76,36 @@ type nsParam struct {
...
-	*dsync.DRWMutex
-	ref uint
+	rwlock RWLocker
+	ref    uint
 }
 
 // nsLockMap - namespace lock map, provides primitives to Lock,
 // Unlock, RLock and RUnlock.
 type nsLockMap struct {
+	isDist  bool
```
harshavardhana: Let's document the fields here this time. We need to specify what 'isDir' stands for.

namespace-lock.go
```diff
@@ -30,37 +76,36 @@ type nsParam struct {
...
 	lockMap map[nsParam]*nsLock
 	mutex   sync.Mutex
```
harshavardhana: Rename this as 'lockMapMutex'

PR #2409, opened 2016-08-12
[DO NOT MERGE] Use go test standard coverage calculation in place of codecov.io algorithm

vadmeste:
https://github.com/codecov/support/wiki/Report-Fixes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,10 +17,7 @@ script:
 - make coverage
 
 after_success:
-- bash <(curl -s https://codecov.io/bash)
-
-after_success:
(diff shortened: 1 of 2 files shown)


vadmeste:
The code coverage rate is different between codecov.io and go test, the difference reaches sometimes 10% for some files.. -X fix was supposed to fix that but it is not working until now. Will see this tomorrow.

vadmeste:
-X fix was not sufficient to have similar results between go coverage tool and codecov.io, opened an issue to understand the reason https://github.com/codecov/support/issues/278

Issue #2416, opened 2016-08-12
tests: Need better coverage for format-config-v1.go


hackintoshrao:

vadmeste:
This patch https://github.com/minio/minio/pull/2424 enhanced the code coverge by 0.5%, will see if I can do better

harshavardhana:
Can you add tests for these -cases ://codecov.io/gh/minio/minio/src/385543ebe06ca191cd1af1233f56a3a9dffedd0e/cmd/format-config-v1.go @vadmeste

vadmeste:
@harshavardhana, codecov.io is hard with us :)

Since the only remaining code not covered are when errors occur (disk not found, etc..), I guess I will write something similar to faulty disks which returns programmed errors, like:

`disk = FaultyDisk{Errors: []DiskError{NoError, NoError, DiskNotFound, ..}, ...}` 

What do you think?

harshavardhana:
> Since the only remaining code not covered are when errors occur (disk not found, etc..), I guess I will write something similar to faulty disks which returns programmed errors, like:
> 
> disk = FaultyDisk{Errors: []DiskError{NoError, NoError, DiskNotFound, ..}, ...}
> 
> What do you think?

Yes that would make sense.

harshavardhana:
Closing as we have reached sufficient coverage for `format-config-v1.go`

PR #2426, opened 2016-08-12
Heal format by inspection should avoid picking minioMetaBucket dir if…

vadmeste:
… the latter comes first in the list (listDir not ordered)

Also increase test coverage #2416


harshavardhana:
LGTM, Thanks :+1:

PR #2429, opened 2016-08-12
tests: Web handlers

vadmeste:


web-handlers_test.go
```diff
@@ -0,0 +1,993 @@
...
+			t.Fatalf("Test %d: Expected to succeed but it failed, %v", i+1, err)
+		}
+		if err == nil && !testCase.success {
+			t.Fatalf("Test %d: Expected to fail but it didn't", i+1, err)
+		}
+
+	}
+}
+
+type StorageInfoRequest struct {
```
harshavardhana: Move this to `StorageInfoRequest struct{}` avoid the '}' in newline.

web-handlers_test.go
```diff
@@ -0,0 +1,993 @@
...
+	BucketName string `json:"bucketName"`
+	Prefix     string `json:"prefix"`
+}
+
+type ListObjectsReply struct {
+	Objects []struct {
+		Name         string `json:"name"`
+		LastModified string `json:"lastModified"`
+		Size         int64  `json:"size"`
+		contentType  string `json:"contentType"`
```
harshavardhana: contentType should exported to have a json tag.

web-handlers_test.go
```diff
@@ -0,0 +1,993 @@
...
+	if listBucketsReply.Buckets[0].Name != bucketName {
+		t.Fatalf("Found another bucket other than already created by MakeBucket")
+	}
+}
+
+type ListObjectsRequest struct {
+	BucketName string `json:"bucketName"`
+	Prefix     string `json:"prefix"`
+}
+
```
harshavardhana: Why are we defining new structs, can't we use the same ones we have have ?

vadmeste:
@harshavardhana, fixed.. but will need to merge https://github.com/minio/minio/pull/2435 first.

PR #2430, opened 2016-08-13
Routing key was misspelled as routine key

yrashk:


harshavardhana:
@yrashk 
LGTM, Thanks :+1:

PR #2432, opened 2016-08-13
Protect shutdown callbacks lists with a mutex

vadmeste:


lbguilherme:
Note: registering a shutdown callback from a shutdown callback will deadlock

utils.go
```diff
@@ -90,19 +91,32 @@ var shutdownCallbacks []func() errCode
...
 	shutdownObjectStorageCallbacks = append(shutdownObjectStorageCallbacks, callback)
+	shutdownMutex.Unlock()
 }
 
 // Represents a type of an exit func which will be invoked during shutdown signal.
 type onExitFunc func(code int)
 
+// Initialize graceful shutdown mechanism
+func prepareGracefulShutdown() {
+	shutdownMutex = &sync.Mutex{}
```
harshavardhana: To have a mutex protection for 'callBack' list it is cleaner to put them in a struct together otherwise the mutex relation is harder to justify.

utils.go
```diff
@@ -90,19 +91,32 @@ var shutdownCallbacks []func() errCode
...
+	shutdownMutex.Lock()
 	shutdownObjectStorageCallbacks = append(shutdownObjectStorageCallbacks, callback)
+	shutdownMutex.Unlock()
 }
 
 // Represents a type of an exit func which will be invoked during shutdown signal.
 type onExitFunc func(code int)
 
+// Initialize graceful shutdown mechanism
+func prepareGracefulShutdown() {
```
harshavardhana: Once the latter comment is addressed you can call this function as 'initGracefulShutdown()'

main.go
```diff
@@ -190,7 +190,7 @@ func main() {
 	}
 
 	// Initialize and monitor shutdown signal
-	shutdownSignal = make(chan bool, 1)
+	prepareGracefulShutdown()
 	monitorShutdownSignal(os.Exit)
```
harshavardhana: This function can go potentially inside 'initGracefulShutdown()' - that way everywhere only one function has to be called.

utils.go
```diff
@@ -77,78 +77,80 @@ func contains(stringList []string, element string) bool {
...
-// we want broadcast the start of shutdown
-var shutdownSignal chan bool
+// Represents a type of an exit func which will be invoked during shutdown signal.
+type onExitFunc func(code int)
 
-// shutdownCallbacks - is the list of function callbacks executed one by one
-// when a shutdown starts. A callback returns 0 for success and 1 for failure.
-// Failure is considered an emergency error that needs an immediate exit
-var shutdownCallbacks []func() errCode
+type ShutdownCallbacks struct {
```
harshavardhana: keep shutdownCallbacks as unexported struct.

utils.go
```diff
@@ -77,78 +77,80 @@ func contains(stringList []string, element string) bool {
...
-// the general callback shutdowns
-var shutdownObjectStorageCallbacks []func() errCode
+// shutdownSignal receives any boolean when we want broadcast
+// the start of the shutdown process
+var shutdownSignal chan bool
 
-// Protect shutdownCallbacks list from a concurrent access
-var shutdownMutex *sync.Mutex
+// shutdownCBs stores regular and object storages callbacks
+var shutdownCBs ShutdownCallbacks
```
harshavardhana: make this a 'pointer here' 

```
var shutdownCBs *shutdownCallbacks
```

harshavardhana:
Sent you a PR on this branch @vadmeste

PR #2435, opened 2016-08-14
Check if eventN is initialized before notifying in Upload web handler

vadmeste:
We need this for tests which initialize webrpc handlers without the notification framework

### DIFF
--- a/event-notifier.go
+++ b/event-notifier.go
@@ -138,7 +138,10 @@ func (en *eventNotifier) RemoveLambdaTarget(lambdaARN string, listenerCh chan []
 }
 
 // Returns true if bucket notification is set for the bucket, false otherwise.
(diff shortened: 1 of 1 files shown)


web-handlers.go
```diff
@@ -393,7 +393,7 @@ func (web *webAPIHandlers) Upload(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if eventN.IsBucketNotificationSet(bucket) {
+	if eventN != nil && eventN.IsBucketNotificationSet(bucket) {
```
harshavardhana: There are many places this has to be done. Better to move this one into the 'IsBucketNotificationSet' itself where you can verify. 

https://play.golang.org/p/T_WgFhGxwY

PR #2436, opened 2016-08-14
storage/rpc-client: Lazily reconnect after first disconnect

krisis:


storage-rpc-client.go
```diff
@@ -199,6 +199,9 @@ func (n networkStorage) AppendFile(volume, path string, buffer []byte) (err erro
 	if n.rpcClient == nil {
```
harshavardhana, line 199: Looks like this is added twice..

rpc-client.go
```diff
@@ -0,0 +1,89 @@
...
+// failure, false otherwise.
+func IsRPCError(err error) bool {
+	if err == nil {
+		return false
+	}
+	// The following are net/rpc specific errors that indicate that
+	// the connection may have been reset. Reset rpcClient.rpc to nil
+	// to trigger a reconnect in future.
+	if err == rpc.ErrShutdown {
+		return true
```
harshavardhana, line 86: Quite simply do this `return err == rpc.ErrShutdown`

PR #2437, opened 2016-08-14
Implement RLock, RUnlock rpc handlers

krisis:
This would simplify dsync.RWMutex's algorithm for acquiring a distributed read lock and be tolerant to N/2-1 node failures.

### DIFF
--- a/lock-rpc-server.go
+++ b/lock-rpc-server.go
@@ -31,7 +31,9 @@ const lockRPCPath = "/minio/lock"
 type lockServer struct {
 	rpcPath string
 	mutex   sync.Mutex
-	lockMap map[string]struct{}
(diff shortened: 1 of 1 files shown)


vadmeste:
Generally speaking, I prefer quorum = N/2+1 instead. Because it represents the majority 51%. In this case, nodes won't be able to perform two conflicting actions at the same time. This would simplify the algorithm.

krisis:
@vadmeste at the server-side, the server process is not aware of quorum. The `DRWMutex` object maintains the state of nodes/servers where the locks were successfully acquired. So the quorum computation happens there. See `dmutex.go` in `minio/dsync` for implementation details. Let me know if they are not clear. N B The server-side RLock/RUnlock hasn't yet been put to use in `minio/dsync` package. I am working on a PR that would do that.

PR #2439, opened 2016-08-15
bug-fix: fix for tests failure when cache is disabled

hackintoshrao:
When cache is disabled and tests are run, it was failiong with following we error, 

```
OK: 102 passed
--- FAIL: TestGetObjectDiskNotFound (0.03s)
        object-api-getobject_test.go:318: Test 6: XL: Expected to fail with error "EOF", but instead failed with error "Storage resources are insufficient for the read operation." instead.
        object-api-getobject_test.go:310: Test 7: XL:  Expected to pass, but failed with: <ERROR> Storage resources are insufficient for the read operation.
        object-api-getobject_test.go:310: Test 8: XL:  Expected to pass, but failed with: <ERROR> Storage resources are insufficient for the read operation.
        object-api-getobject_test.go:310: Test 9: XL:  Expected to pass, but failed with: <ERROR> Storage resources are insufficient for the read operation.
FAIL
```

The test had 8 out 16 disks down. 

Fix relaxes the checks which were not necessary.

### DIFF
--- a/erasure-readfile.go
+++ b/erasure-readfile.go
@@ -43,8 +43,8 @@ func isSuccessDecodeBlocks(enBlocks [][]byte, dataBlocks int) bool {
 		} // else { // update parity block count.
 		successParityBlocksCount++
 	}
-	// Returns true if we have atleast dataBlocks + 1 parity.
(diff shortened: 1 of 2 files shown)


harshavardhana:
Look like codecoverage hit the same problem again.

hackintoshrao:
Ya, the report is not right.

PR #2440, opened 2016-08-15
Problem: AWS documentation defines event timestamp as ISO-8601

yrashk:
It specifies the following format: 1970-01-01T00:00:00.000Z

While the one Minio is using 20160814T124605Z

(See http://docs.aws.amazon.com/AmazonS3/latest/dev/notification-content-structure.html)

Solution: adhere to AWS documentation

### DIFF
--- a/event-notifier.go
+++ b/event-notifier.go
@@ -60,7 +60,7 @@ func newNotificationEvent(event eventData) NotificationEvent {
 		EventVersion:      "2.0",
 		EventSource:       "aws:s3",
 		AwsRegion:         region,
-		EventTime:         tnow.Format(iso8601Format),
+		EventTime:         tnow.Format(timeFormatAMZ),
 		EventName:         event.Type.String(),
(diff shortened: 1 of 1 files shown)


event-notifier.go
```diff
@@ -60,7 +60,7 @@ func newNotificationEvent(event eventData) NotificationEvent {
 		EventVersion:      "2.0",
 		EventSource:       "aws:s3",
 		AwsRegion:         region,
-		EventTime:         tnow.Format(iso8601Format),
+		EventTime:         tnow.Format("1970-01-01T00:00:00.000Z"),
```
harshavardhana: @yrashk - you can use. `time.RFC3339`

harshavardhana: > @yrashk - you can use. time.RFC3339

or @yrashk - `tnow.Format(timeFormatAMZ)`

harshavardhana:
LGTM, Thanks :+1:

Issue #2400, opened 2016-08-10
[distributed] Network path parsing fails to detect windows disk label correctly


krisis:
## Expected behaviour

This issue is specific to the `distributed` branch of this repo.

harshavardhana:
Fixed already by @awwalker

PR #2443, opened 2016-08-15
splitNetPath: Add support for windows paths including volumeNames e.g…

awwalker:
… ip:C:\network\path

Fixes: #2400, and #2446

### DIFF
--- a/storage-rpc-client.go
+++ b/storage-rpc-client.go
@@ -19,10 +19,7 @@ package main
 import (
 	"errors"
 	"io"
-	"net"
 	"path"
-	"path/filepath"
-	"runtime"
 	"strconv"
 	"strings"
 )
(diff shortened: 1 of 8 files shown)


storage-rpc-client.go
```diff
@@ -38,9 +40,18 @@ const (
 
 // splits network path into its components Address and Path.
 func splitNetPath(networkPath string) (netAddr, netPath string) {
+	// For paths of form ip:path
 	index := strings.LastIndex(networkPath, ":")
 	netAddr = networkPath[:index]
 	netPath = networkPath[index+1:]
+	if runtime.GOOS == "windows" { // Windows paths can come in form of ip:volumeName:\\network\\path.
+		volumeName := filepath.VolumeName(networkPath)
```
harshavardhana: You have to move this outside .. before we do the split.

utils_test.go
```diff
@@ -16,7 +16,41 @@
```
harshavardhana, line 16: This test needs to be moved as 'utils_nix_test.go'

utils.go
```diff
@@ -20,12 +20,37 @@ import (
 	"encoding/base64"
```
harshavardhana: You forgot to add `` // +build !windows``

utils_test.go
harshavardhana, line 16: You forgot to add `// +build !windows`

PR #2444, opened 2016-08-15
When updating the meta file, write to temp file first and then rename to the actual location.

krishnasrinivas:
This prevents appending the metadata to the metadata-file when a file is reuploaded.


harshavardhana:
Merged already..

PR #2451, opened 2016-08-16
tests: Object handlers tests. 

hackintoshrao:
Tests for 
- GetObject API handler. 
- CopyObject API handler.

### DIFF
--- a/object-handlers_test.go
+++ b/object-handlers_test.go
@@ -1,0 +1,334 @@
+/*
+ * Minio Cloud Storage, (C) 2016 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


object-handlers_test.go
```diff
@@ -0,0 +1,358 @@
...
+			expectedRespStatus: http.StatusRequestedRangeNotSatisfiable,
+		},
+		// Test case - 5.
+		// Test case with byte range exceeding the object size.
+		// Expected to read till end of the object.
+		{
+			bucketName:         bucketName,
+			objectName:         objectName,
+			byteRange:          "bytes=10-1000000000000000",
+			accessKey:          credentials.AccessKeyID,
```
harshavardhana: This is incorrect style, we shouldn't put everything into a single struct like this. Each requests don't have new access keys. We should split the cases with or without signed requests in different tests. 

This way its easy to fix bugs and run only those tests selectively.

hackintoshrao: Removing the keys from the struct. Will have separate tests for unsigned cases. That sounds good.

hackintoshrao:
Have made the suggested changes @harshavardhana .

PR #2467, opened 2016-08-17
Refactoring MuxServer Close() method to always wait for conns to close

jesselucas:
After the GracefulTimeout happens active connections weren't being forcefully closed. Also adding additional test to server-mux_test.go to make sure open connections are closed.


harshavardhana:
LGTM, Thanks :+1:

PR #2485, opened 2016-08-18
server-mux_test.go updating test to fix race condition

jesselucas:


server-mux_test.go
```diff
@@ -146,17 +142,41 @@ func TestServerCloseBlocking(t *testing.T) {
 }
 
 func TestListenAndServe(t *testing.T) {
-	m := NewMuxServer("", nil)
-	stopc := make(chan struct{})
+	wait := make(chan struct{})
+
+	// get an open port
+	ln, err := net.Listen("tcp", "127.0.0.1:0")
```
harshavardhana: You can get an open port by using a helper function we already have `getFreePort()`

Issue #2483, opened 2016-08-17
bucket notification API: prefix and suffix don't support asterisk character


vadmeste:
```
$ mc listen myminio/testbucket/ --suffix "*.go"
mc: <ERROR> Cannot watch on the specified bucket. prefix rule value cannot exceed 1024 characters
```

harshavardhana:
> $ mc listen myminio/testbucket/ --suffix "*.go"
> mc: <ERROR> Cannot watch on the specified bucket. prefix rule value cannot exceed 1024 characters

Does 's3' support?

vadmeste:
> Does 's3' support?

yes, S3 supports asterisk character, by the way, it doesn't interpret it

harshavardhana:
> yes, S3 supports asterisk character, by the way, it doesn't interpret it

Suffix doesn't have regex support, also the error message here is wrong.

harshavardhana:
Fixed by #2487

PR #2487, opened 2016-08-18
Allow asterisk character in key names

vadmeste:
http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html#object-keys, in section `Safe Characters`, asterisk is listed.

I successfully tested simple mc commands (ls, cp, cat) with keys that contain asterisks. Otherwise, I don't know if there is a wise reason to do not allow asterisk, drop the patch if any.

Refer #2483

### DIFF
--- a/api-errors.go
+++ b/api-errors.go
@@ -118,7 +118,7 @@ const (
 	ErrFilterNameInvalid
 	ErrFilterNamePrefix
 	ErrFilterNameSuffix
-	ErrFilterPrefixValueInvalid
+	ErrFilterValueInvalid
 
 	// S3 extended errors.
 	ErrContentSHA256Mismatch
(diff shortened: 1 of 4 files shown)


api-errors.go
```diff
@@ -500,9 +500,9 @@ var errorCodeResponse = map[APIErrorCode]APIError{
 		Description:    "Cannot specify more than one suffix rule in a filter.",
 		HTTPStatusCode: http.StatusBadRequest,
 	},
-	ErrFilterPrefixValueInvalid: {
+	ErrFilterValueInvalid: {
 		Code:           "InvalidArgument",
-		Description:    "prefix rule value cannot exceed 1024 characters",
+		Description:    "filter rule value is too long (> 1024) or contains unsupported characters",
```
harshavardhana: Wait this error string is s3 compatible?

vadmeste: > Wait this error string is s3 compatible?

we should output the same error string? this is s3 msg: `Size of filter rule value cannot exceed 1024 bytes in UTF-8 representation`

harshavardhana: Yes we have to

harshavardhana: Error messages are compliant with s3 with only extended errors have other meaning.

object-utils.go
```diff
@@ -103,7 +103,7 @@ func IsValidObjectPrefix(object string) bool {
 		return false
 	}
 	// Reject unsupported characters in object name.
-	if strings.ContainsAny(object, "`^*|\\\"") {
+	if strings.ContainsAny(object, "`^|\\\"") {
```
harshavardhana, line 106: We need to add a test case here to validate this is 'true' now.

harshavardhana, line 106:
```
object-utils_test.go:79:func TestIsValidObjectName(t *testing.T) {
object-utils_test.go:101:               isValidObjectName := IsValidObjectName(testCase.objectName)
```

PR #2492, opened 2016-08-19
Fix occasional Server Mux test due to a normal error returned by ListenAndServe()

vadmeste:
The error which was occuring is returned by ListenAndServe() when we close the server and considered normal. For example, this is how Caddy treats this case:

https://github.com/mholt/caddy/blob/68be4a91610a17c95716bd06c3191282dd9ce47c/caddy.go#L657

### DIFF
--- a/cmd/server-mux_test.go
+++ b/cmd/server-mux_test.go
@@ -23,7 +23,10 @@ import (
 	"net"
 	"net/http"
 	"net/http/httptest"
+	"strconv"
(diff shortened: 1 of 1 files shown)


harshavardhana:
Merged a change from @jesselucas can you check again @vadmeste ?

vadmeste:
@harshavardhana, @jesselucas's patch does not fail anymore, you can close this PR

PR #2493, opened 2016-08-19
Adding test coverage for server-mux.go by creating TestListenAndServeTLS()

jesselucas:


cmd/config.go
```diff
@@ -19,20 +19,27 @@ package cmd
 import (
 	"os"
```
harshavardhana, line 20: Can your commit be separated into new lines? following an 80 column rule. 

It gets harder to read 'git logs'

harshavardhana, line 20: Thanks for all the hardwork on this btw :+1:

PR #2512, opened 2016-08-22
Automatic re-connect and JWT token re-negotiation on disconnect

krisis:


harshavardhana:
@krisis this has an additional patch can you rebase it properly?

PR #2513, opened 2016-08-22
Append the parts to a file as and when the parts arrive

krishnasrinivas:
@harshavardhana this introduces two new files in the backend fs-append.json and fs-append.data though it does not change any format structures.

### DIFF
--- a/cmd/fs-v1-metadata.go
+++ b/cmd/fs-v1-metadata.go
@@ -77,9 +77,9 @@ func (m *fsMetaV1) AddObjectPart(partNumber int, partName string, partETag strin
 }
 
 // readFSMetadata - returns the object metadata `fs.json` content.
-func readFSMetadata(disk StorageAPI, bucket, object string) (fsMeta fsMetaV1, err error) {
+func readFSMetadata(disk StorageAPI, bucket, filePath string) (fsMeta fsMetaV1, err error) {
 	// Read all `fs.json`.
-	buf, err := disk.ReadAll(bucket, path.Join(object, fsMetaJSONFile))
+	buf, err := disk.ReadAll(bucket, filePath)
 	if err != nil {
 		return fsMetaV1{}, err
 	}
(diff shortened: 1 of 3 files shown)


krishnasrinivas:
Needs thorough review so that there are no locking issues or data corruption issues.

gowithplanb:
Tested this patch with CloudBerry Labs MBS client, time-out issues with large files are solved now. Also no issues with concurrent jobs.

krishnasrinivas:
@harshavardhana can this be merged?

harshavardhana:
> @harshavardhana can this be merged?

One question i have with this approach is what happens when a client retries and uploads the same part?  how will such a part be written back to disk - would it be appended to same file which is being appended? 

Or is this validation lazy?  i.e upon complete-multi-part we attempt to concat again?

krishnasrinivas:
> One question i have with this approach is what happens when a client retries and uploads the same part? how will such a part be written back to disk - would it be appended to same file which is being appended?
> 
> Or is this validation lazy? i.e upon complete-multi-part we attempt to concat again?

yes, if a part is reuploaded with different checksum then isPartsSame() check during completeMultilartUpload() fails and it falls back to the old concatenation way.

harshavardhana:
> yes, if a part is reuploaded with different checksum then isPartsSame() check during completeMultilartUpload() fails and it falls back to the old concatenation way.

The question i had was if the same part was re-uploaded would the AppendFile progress forward? since there is no way to Seek() .. and still end up thinking that the file is in-fact where it might not be at the end?

krishnasrinivas:
> > yes, if a part is reuploaded with different checksum then isPartsSame() check during completeMultilartUpload() fails and it falls back to the old concatenation way.
> 
> The question i had was if the same part was re-uploaded would the AppendFile progress forward? since there is no way to Seek() .. and still end up thinking that the file is in-fact where it might not be at the end?

if the part is reuploaded the appendPart() will progress and will not take into account the new part. Hence in the end when completeMultipartUpload there will be isPartsSame() check which will fail. So the appended file will be ignored and it will be concatenated again.

harshavardhana:
> if the part is reuploaded the appendPart() will progress and will not take into account the new part. Hence in the end when completeMultipartUpload there will be isPartsSame() check which will fail. So the appended file will be ignored and it will be concatenated again.

Okay cool, let me talk to @abperiasamy about this patch.

harshavardhana:
@krishnasrinivas we decided to do a separate backend for this by making XL itself support Single disk and multi disk. 

Single disk means -- no erasure but all other things are available (saving metadata and bit-rot)
Multiple disks means -- all things same as is.

harshavardhana:
> @krishnasrinivas we decided to do a separate backend for this by making XL itself support Single disk and multi disk.
> 
> Single disk means -- no erasure but all other things are available (saving metadata and bit-rot)
> Multiple disks means -- all things same as is.

This patch is ready for merge. anything else pending @krishnasrinivas ?

harshavardhana:
LGTM, Thanks :+1:

gowithplanb:
Thanks for merging this patch guys!

PR #2528, opened 2016-08-22
XL/boot: For XL setup, MINIO_ACCESS_KEY and MINIO_SECRET_KEY are mand…

### LINKED ISSUE #2509: [distributed]: Show error when keys are not set.
When Access and Secret keys are not set and RPC authentication across node fails no error is shown, It just doesn't work. 

A meaningful error message has to be displayed.

krishnasrinivas:
…atory. (for healing and storage RPC calls)

fixes #2509


harshavardhana:
This is an incorrect fix . We are not forcing people to set envs all the time. In a simpler setup we are asking them to copy `config.json`

harshavardhana:
Is there a reason you would force people to have envs ?

krishnasrinivas:
> This is an incorrect fix . We are not forcing people to set envs all the time. In a simpler setup we are asking them to copy config.json

I see .. for some reason I had assumed that for distributed XL keys are always from env var, let me fix it then.

harshavardhana:
Closing this PR - please re-open @krishnasrinivas if needed freshly.

PR #2532, opened 2016-08-22
Add Heal Disk metadata (format)

vadmeste:
Still in progress:
*) Missing some locks
*) Some serious testing


harshavardhana:
Can you submit this to distributed branch instead @vadmeste ?

PR #2534, opened 2016-08-23
signature: Use the client sent region if region is properly validated.

### LINKED ISSUE #2521: Check malformed authorization header to report good info instead of SignatureDoesNotMatch error
Currently, if the "region"/"bucket_location" setup in the client is wrong, the s3 v4 signature will fail as the server uses it in his own check of the signature.
The client will then receive a "SignatureDoesNotMatch" error.

As does aws s3, it could be good to report a "AuthorizationHeaderMalformed" error instead with the indication of the expected region, so the client can fix his configuration.

Setup s3cmd with the following config:

```
access_key = Q3AM3UQ867SPQQA43P2F
secret_key = zuf+tfteSlswRu7BJ86wekitnifILbZam1KYY3TG
host_base = play.minio.io:9000
host_bucket = play.minio.io:9000
use_https = True
signature_v2 = False
```

Ensure to have "**bucket_location**" unset.
Run s3cmd with -d to have the debug log:
`s3cmd -d -c playminio_config ls`
## Minio version
- play.minio.io:9000 (today 22/08/2016)
## System information

Client: s3cmd on Linux
## The error that should be returned by Minio in my opinion

In the case of s3cmd, as often the "region" setting is not provided by the user, we use a boggus "US" region for the first signature v4 request and then use the AuthorizationHeaderMalformed error reply to have the "correct" region to use for the following requests. All of that automatically.

Note: I think that this kind of issue is like the one encountered by a mc user here: https://github.com/minio/mc/issues/1707

In the following, I have pasted what reply amazon s3 when I issue such a request with the region.
Region name i try to use can be checked in the authorization header:
"Credential=ABC....ZQ/20160822/**US**/s3/aws4_request..."
and aws s3 gives the correct region to use in the reply:
"<Code>AuthorizationHeaderMalformed</Code><Message>The authorization header is malformed; the region \'US\' is wrong; expecting \'eu-central-1\'</Message>**<Region>eu-central-1</Region>**"

```
DEBUG: signature-v4 headers: {'x-amz-content-sha256': 'e....44', 'x-amz-date': '20160822T125922Z', 'Authorization': 'AWS4-HMAC-SHA256 Credential=ABC....ZQ/20160822/US/s3/aws4_request,SignedHeaders=host;x-amz-content-sha256;x-amz-date,Signature=54..c8'}
DEBUG: Processing request, please wait...
DEBUG: get_hostname(None): s3.eu-central-1.amazonaws.com
DEBUG: ConnMan.get(): re-using connection: https://s3.eu-central-1.amazonaws.com#1
DEBUG: format_uri(): /
DEBUG: Sending request method_string='GET', uri='/', headers={'x-amz-content-sha256': 'e...44', 'x-amz-date': '20160822T125922Z', 'Authorization': 'AWS4-HMAC-SHA256 Credential=ABC....ZQ/20160822/US/s3/aws4_request,SignedHeaders=host;x-amz-content-sha256;x-amz-date,Signature=e...44'}, body=(0 bytes)
DEBUG: ConnMan.put(): connection put back to pool (https://s3.eu-central-1.amazonaws.com#2)
DEBUG: Response:
{'data': '<?xml version="1.0" encoding="UTF-8"?>\n<Error><Code>AuthorizationHeaderMalformed</Code><Message>The authorization header is malformed; the region \'US\' is wrong; expecting \'eu-central-1\'</Message><Region>eu-central-1</Region><RequestId>FDA...61</RequestId><HostId>dpRazc2J...BgNy1X6GQK8CO3l1m=</HostId></Error>',
 'headers': {'connection': 'close',
             'content-type': 'application/xml',
             'date': 'Mon, 22 Aug 2016 13:00:13 GMT',
             'server': 'AmazonS3',
             'transfer-encoding': 'chunked',
             'x-amz-id-2': 'dp...NY=',
             'x-amz-request-id': 'FDA...61'},
 'reason': 'Bad Request',
 'status': 400}
INFO: Forwarding request to eu-central-1
```

Ending note: Thank you for providing the "playminio" public testing account, that is very convenient to test Minio.

harshavardhana:
A properly validated region string from the client should be once
validated properly.

Fixes #2521

### DIFF
--- a/cmd/signature-v4.go
+++ b/cmd/signature-v4.go
@@ -361,9 +361,8 @@ func doesSignatureMatch(hashedPayload string, r *http.Request, validateRegion bo
 		if !isValidRegion(sRegion, region) {
 			return ErrInvalidRegion
 		}
-	} else {
-		region = sRegion
 	}
+	region = sRegion
 
 	// Extract date, if not present throw error.
 	var date string


fviard:
Thank you for the fast fix :-)

PR #2546, opened 2016-08-24
List only objects that need healing

vadmeste:


cmd/xl-v1-list-objects-heal.go
```diff
@@ -137,12 +137,18 @@ func (xl xlObjects) listObjectsHeal(bucket, prefix, marker, delimiter string, ma
...
 		}
-		result.Objects = append(result.Objects, ObjectInfo{
-			Name:    objInfo.Name,
-			ModTime: objInfo.ModTime,
-			Size:    objInfo.Size,
-			IsDir:   false,
-		})
+		// Check if the current object needs healing
+		nsMutex.RLock(bucket, objInfo.Name)
+		partsMetadata, errs := readAllXLMetadata(xl.storageDisks, bucket, objInfo.Name)
```
vadmeste, line 142: Here I avoided to test on readAllXLMetadata() returned errors and I let xlShouldHeal() decides instead. I think that here we should show that there is something bad occurring and needs to be healed _even_ when we are not able to heal it. @krishnasrinivas, what do you think?

krishnasrinivas: @vadmeste true, in that case we can make errDiskNotFound the only pardonable error in xlShouldHeal, so xlShouldHeal will be:

```
// Returns if the object should be healed.                                                                                                                                      
func xlShouldHeal(partsMetadata []xlMetaV1, errs []error) bool {                                                                                                                
        modTime := commonTime(listObjectModtimes(partsMetadata, errs))                                                                                                          
        for index := range partsMetadata {                                                                                                                                      
                if errs[index] == errDiskNotFound {                                                                                                                             
                        continue                                                                                                                                                
                }                                                                                                                                                               
                if errs[index] != nil {                                                                                                                                         
                        return true                                                                                                                                             
                }                                                                                                                                                               
                if modTime != partsMetadata[index].Stat.ModTime {                                                                                                               
                        return true                                                                                                                                             
                }                                                                                                                                                               
        }                                                                                                                                                                       
        return false                                                                                                                                                            
}
```

vadmeste: :+1:

PR #2558, opened 2016-08-26
Stack trace in case of errors.

krishnasrinivas:


cmd/bucket-policy.go
```diff
@@ -128,6 +133,10 @@ func readBucketPolicy(bucket string, objAPI ObjectLayer) (*bucketPolicy, error)
 	}
 	policyPath := pathJoin(bucketConfigPrefix, bucket, policyJSON)
 	objInfo, err := objAPI.GetObjectInfo(minioMetaBucket, policyPath)
+	errorIf(err, "Unable to get policy.")
```
harshavardhana: `errorIf(err, "Unable to read bucket %s policy", bucket)`

cmd/bucket-policy.go
```diff
@@ -128,6 +133,10 @@ func readBucketPolicy(bucket string, objAPI ObjectLayer) (*bucketPolicy, error)
...
+	if e, ok := err.(*Error); ok {
```
harshavardhana: Can we put a function on this?

cmd/bucket-policy.go
```diff
@@ -136,6 +145,10 @@ func readBucketPolicy(bucket string, objAPI ObjectLayer) (*bucketPolicy, error)
 	}
 	var buffer bytes.Buffer
 	err = objAPI.GetObject(minioMetaBucket, policyPath, 0, objInfo.Size, &buffer)
+	errorIf(err, "Unable to get policy.")
```
harshavardhana: errorIf(err, "Unable to read bucket %s policy", bucket)

cmd/bucket-policy.go
```diff
@@ -161,6 +174,10 @@ func removeBucketPolicy(bucket string, objAPI ObjectLayer) error {
 	}
 	policyPath := pathJoin(bucketConfigPrefix, bucket, policyJSON)
 	if err := objAPI.DeleteObject(minioMetaBucket, policyPath); err != nil {
+		errorIf(err, "Unable to remove bucket policy.")
```
harshavardhana: `errorIf(err, "Unable to remove bucket %s policy.", bucket)`

cmd/bucket-policy.go
```diff
@@ -178,5 +195,9 @@ func writeBucketPolicy(bucket string, objAPI ObjectLayer, reader io.Reader, size
 
 	policyPath := pathJoin(bucketConfigPrefix, bucket, policyJSON)
 	_, err := objAPI.PutObject(minioMetaBucket, policyPath, size, reader, nil)
+	errorIf(err, "Unable to set policy.")
```
harshavardhana: `errorIf(err, "Unable to save bucket %s policy.", bucket)`

cmd/erasure-createfile_test.go
```diff
@@ -93,8 +94,9 @@ func TestErasureCreateFile(t *testing.T) {
 	// 1 more disk down. 7 disk down in total. Should return quorum error.
 	disks[10] = AppendDiskDown{disks[10].(*posix)}
 	_, _, err = erasureCreateFile(disks, "testbucket", "testobject4", bytes.NewReader(data), blockSize, dataBlocks, parityBlocks, bitRotAlgo, dataBlocks+1)
-	if err != errXLWriteQuorum {
-		t.Errorf("erasureCreateFile returned expected errXLWriteQuorum error, got %s", err)
+	if !isErrorOfType(err, errXLWriteQuorum) {
+		fmt.Println(err)
```
harshavardhana: Remove this 'fmt.Println'

cmd/errors.go
```diff
@@ -0,0 +1,83 @@
+package cmd
```
harshavardhana, line 17: Missing license header.

cmd/errors.go
```diff
@@ -0,0 +1,83 @@
...
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+)
+
```
harshavardhana, line 26: Comment ...

cmd/errors.go
```diff
@@ -0,0 +1,83 @@
...
+func initError() {
+	// Root path is automatically determined from the calling function's source file location.
+	// Catch the calling function's source file path.
+	_, file, _, _ := runtime.Caller(1)
+	// Save the directory alone.
+	rootPath = filepath.Dir(file)
+	fmt.Println(file)
+}
+
```
harshavardhana, line 38: Comment for the struct.

cmd/event-notifier.go
```diff
@@ -225,6 +225,9 @@ func loadNotificationConfig(bucket string, objAPI ObjectLayer) (*notificationCon
 	// Construct the notification config path.
 	notificationConfigPath := path.Join(bucketConfigPrefix, bucket, bucketNotificationConfig)
 	objInfo, err := objAPI.GetObjectInfo(minioMetaBucket, notificationConfigPath)
```
harshavardhana, line 227: Log here 

`errorIf(err, "Unable to get bucket %s notification.", bucket)`

cmd/event-notifier.go
```diff
@@ -237,6 +240,9 @@ func loadNotificationConfig(bucket string, objAPI ObjectLayer) (*notificationCon
 	}
 	var buffer bytes.Buffer
 	err = objAPI.GetObject(minioMetaBucket, notificationConfigPath, 0, objInfo.Size, &buffer)
```
harshavardhana, line 241: Same here. 

`errorIf(err, "Unable to get bucket %s notification.", bucket)`

cmd/logger.go
```diff
@@ -67,9 +67,10 @@ func errorIf(err error, msg string, data ...interface{}) {
 	fields := logrus.Fields{
 		"cause": err.Error(),
 	}
-	if globalTrace {
-		fields["stack"] = "\n" + stackInfo()
+	if e, ok := err.(*Error); ok {
+		fields["stack"] = strings.Join(e.Trace(), " ")
```
harshavardhana, line 71: We can still put some 'stackInfo()' when err is not compatible with 'TraceError'

cmd/main.go
```diff
@@ -160,6 +160,8 @@ func Main() {
 		// Enable all loggers by now.
 		enableLoggers()
```
harshavardhana, line 162: Comment.

cmd/object-api-getobject_test.go
```diff
@@ -159,6 +159,7 @@ func testGetObject(obj ObjectLayer, instanceType string, t TestErrHandler) {
 		if err != nil && !testCase.shouldPass {
 			if !strings.Contains(err.Error(), testCase.err.Error()) {
 				t.Errorf("Test %d: %s: Expected to fail with error \"%s\", but instead failed with error \"%s\" instead.", i+1, instanceType, testCase.err.Error(), err.Error())
+				errorIf(err, "error")
```
harshavardhana: any reason for double 'errors'?

cmd/object-api-listobjects_test.go
```diff
@@ -517,6 +517,7 @@ func testListObjects(obj ObjectLayer, instanceType string, t TestErrHandler) {
 	for i, testCase := range testCases {
 		result, err := obj.ListObjects(testCase.bucketName, testCase.prefix, testCase.marker, testCase.delimeter, testCase.maxKeys)
 		if err != nil && testCase.shouldPass {
+			errorIf(err, "error")
```
harshavardhana: Same here..

cmd/object-common.go
```diff
@@ -33,8 +33,18 @@ const (
 	bucketMetaPrefix = "buckets"
 )
```
harshavardhana, line 35: comment..

cmd/object-errors.go
```diff
@@ -264,3 +273,17 @@ type NotImplemented struct{}
 func (e NotImplemented) Error() string {
 	return "Not Implemented"
 }
+
+// InvalidArgument - invalid argument error
```
harshavardhana: why adding unused struct?

cmd/xl-v1-healing.go
```diff
@@ -138,7 +138,7 @@ func xlShouldHeal(partsMetadata []xlMetaV1, errs []error) bool {
 			return true
 		}
 		if errs[index] != nil {
-			continue
+			return true
```
harshavardhana, line 141: Was it decided to `return true` only for err == errDiskNotFound and mismatching modtime?

cmd/xl-v1-utils.go
```diff
@@ -46,20 +58,36 @@ func reduceErrs(errs []error, ignoredErrs []error) error {
...
+		return nil
+	}
+	return traceError(errMax, errs...)
 }
 
 // Validates if we have quorum based on the errors related to disk only.
 // Returns 'true' if we have quorum, 'false' if we don't.
 func isDiskQuorum(errs []error, minQuorumCount int) bool {
 	var count int
+	Errs := make([]error, len(errs))
```
harshavardhana: convert this to a function. 

`toTraceErrs(errs []error) traceErrs []error`

krishnasrinivas:
@harshavardhana the patch is not review worthy hence put `[wip-donotreview]` but just wanted to show the direction it is going. if you see the check `if e, ok := err.(*Error);..` everywhere which does not look neat, we could do away with `error` interface for error returns and use `Error` directly. There will also be better type checking, because as of now `err == errFileNotFound` will be false even if `err`' is an interface variable that wraps errFileNotFound in Error, but there will be no compile time error. It was tough to track down such situations while doing this patch.

harshavardhana:
> @harshavardhana the patch is not review worthy hence put [wip-donotreview] but just wanted to show the direction it is going. if you see the check if e, ok := err.(*Error);.. everywhere which does not look neat, we could do away with error interface for error returns and use Error directly. There will also be better type checking, because as of now err == errFileNotFound will be false even if err' is an interface variable that wraps errFileNotFound in Error, but there will be no compile time error. It was tough to track down such situations while doing this patch.

Oh my bad forgot that.

cmd/xl-v1-healing.go
krishnasrinivas: oh, no, xlShouldHeal should have been:

```
func xlShouldHeal(partsMetadata []xlMetaV1, errs []error) bool {                                                                                                                
        modTime := commonTime(listObjectModtimes(partsMetadata, errs))                                                                                                          
        for index := range partsMetadata {                                                                                                                                      
                if errs[index] == errDiskNotFound {                                                                                                                             
                        continue                                                                                                                                                
                }                                                                                                                                                               
                if errs[index] != nil {                                                                                                                                         
                        return true                                                                                                                                             
                }                                                                                                                                                               
                if modTime != partsMetadata[index].Stat.ModTime {                                                                                                               
                        return true                                                                                                                                             
                }                                                                                                                                                               
        }                                                                                                                                                                       
        return false                                                                                                                                                            
}   
```

i.e any error other than errDiskNotFound should be means it needs healing.

cmd/object-api-getobject_test.go
krishnasrinivas: i had added `errorIf(err, "error")` to get the stack trace to see where the error was originating, will remove it

cmd/logger.go
krishnasrinivas: @harshavardhana it is difficult to put a stack trace (using `stackInfo()`) when it is not available in `err` because it will mean sometimes stack trace represents origin of error and sometimes it represents the place where the errorIf is done - it will be confusing.

harshavardhana:
> @harshavardhana it is difficult to put a stack trace (using stackInfo()) when it is not available in err because it will mean sometimes stack trace represents origin of error and sometimes it represents the place where the errorIf is done - it will be confusing.

Agreed

harshavardhana:
Is this ready @krishnasrinivas ?

krishnasrinivas:
@harshavardhana yes

PR #2565, opened 2016-08-29
Docker multinode documentation

vadmeste:
This is the simplest method that I could find to run minio in multinodes mode, which is not handy only if the user uses script shells.

The other solution, more comfortable but little long to configure the first time, is using a service discovery, for example: docker swarm + etcd or consul or other. For that, the user should first specify in the service discovery backend how many nodes he wants to run. Then, we add a script shell in minio container, to register the container itself, check the number of nodes specified by the user, then check if all nodes  registered themselves, get their IP addresses and start minio server with the right arguments.


krishnasrinivas:
@vadmeste i find manually starting docker containers and testing it simpler as it needs less docker expertise, in case it's useful for others, i do it this way:
- open 6 terminals
- in each terminal run docker as:
  `docker run -it  -v /home/krishna:/root ubuntu /bin/bash`
- now i have a docker container with bash access in each of the 6 terminals, so it's like 6 VMs now.
- now the host minio binary will be available as:

```
root@8579ccd573a2:~# ls gocode/bin/minio  <- gocode is the GOPATH
gocode/bin/minio
root@8579ccd573a2:~#
```
- you can start minio in each container:

```
/root/gocode/bin/minio server 169.254.0.1:/export 169.254.0.2:/export 169.254.0.3:/export 169.254.0.4:/export 169.254.0.5:/export 169.254.0.6:/export
```
- on the host system i can connect to any of the minio processes
- i can ctrl-C, start the minio processes in the containers with bash access just like a regular VM
- in case i need to add debug statements, i just do it on the host system and compile it, and it will be available to all the containers inside which you can ctrl-C and start the minio process

harshavardhana:
> @vadmeste i find manually starting docker containers and testing it simpler as it needs less docker expertise, in case it's useful for others, i do it this way:
> 
> open 6 terminals
> 
> in each terminal run docker as:
> docker run -it -v /home/krishna:/root ubuntu /bin/bash
> 
> now i have a docker container with bash access in each of the 6 terminals, so it's like 6 VMs now.
> 
> now the host minio binary will be available as:
> 
> root@8579ccd573a2:~# ls gocode/bin/minio  <- gocode is the GOPATH
> gocode/bin/minio
> root@8579ccd573a2:~#
> you can start minio in each container:
> /root/gocode/bin/minio server 169.254.0.1:/export 169.254.0.2:/export 169.254.0.3:/export 169.254.0.4:/export 169.254.0.5:/export 169.254.0.6:/export
> on the host system i can connect to any of the minio processes
> 
> i can ctrl-C, start the minio processes in the containers with bash access just like a regular VM
> 
> in case i need to add debug statements, i just do it on the host system and compile it, and it will be available to all the containers inside which you can ctrl-C and start the minio process

@krishnasrinivas we need to document this for a production deployment, not for testing. Also we need to use the published container image here not ubuntu and download binaries.

harshavardhana:
this branch has conflicts @vadmeste can you rebase?

vadmeste:
> this branch has conflicts @vadmeste can you rebase?

:+1: 

we have many conflicts here because we are force rewriting commits history in this branch

PR #2572, opened 2016-08-29
Move formatting of disks out of object layer initialization

krisis:


krisis:
Still need to do tests in a distributed setup.

cmd/controller-handlers.go
```diff
@@ -100,7 +100,16 @@ func (c *controllerAPIHandlers) HealObjectHandler(args *HealObjectArgs, reply *G
...
 		return errInvalidToken
 	}
-	return objAPI.HealObject(args.Bucket, args.Object)
+	err := objAPI.HealObject(args.Bucket, args.Object)
+	if err != nil {
+		return err
+	}
+	// Notify waitForFormatting function that a disk was healed.
+	// N B Using a go-routine to avoid send on channel to block.
+	go func() {
```
harshavardhana: We shouldn't add the `globalWakeupCh` here it should be added `HealDiskMetadataHandler`

krisis:
Basic sanity tests on a cluster setup with 4 disks and 2 nodes worked.

PR #2583, opened 2016-08-30
posix: Avoid relying on string errors messages

vadmeste:
Needs to test with travis first, then read the code again and add some comments


cmd/posix.go
```diff
@@ -88,6 +88,58 @@ func isDirEmpty(dirname string) bool {
 	return false
 }
 
+func isSysErrNotDir(err error) bool {
```
harshavardhana: These should go into 'posix-errors.go' and needs comments..

PR #2585, opened 2016-08-31
docker: Minimize image size

thatguystone:
This takes the image size from 823.1MB (current edge) to 328MB (go1.6) // 281.8MB (go1.7) and eliminates a few layers.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,14 +1,17 @@
-FROM golang:1.6
+FROM golang:1.6-alpine
 
-RUN mkdir -p /go/src/app
 WORKDIR /go/src/app
+ENV ALLOW_CONTAINER_ROOT=1
 
 COPY . /go/src/app
-RUN go-wrapper download
-RUN go-wrapper install
-
-ENV ALLOW_CONTAINER_ROOT=1
-RUN mkdir -p /export/docker && cp /go/src/app/docs/Docker.md /export/docker/
+RUN \
+	apk add --no-cache git && \
+	go-wrapper download && \
+	go-wrapper install && \
+	mkdir -p /export/docker && \
+	cp /go/src/app/docs/Docker.md /export/docker/ && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.6
+FROM golang:1.7
```
harshavardhana, line 2: We are not planning to move to go 1.7 just yet

Dockerfile
```diff
@@ -1,14 +1,17 @@
-FROM golang:1.7
+FROM golang:1.7-alpine
```
harshavardhana, line 2: @rajivm would love your feedback on this.

Dockerfile
thatguystone: No problem. I'll change.

rajivm:
Last time I tried this I ran into some issues, if its working fully that would be awesome. Please let me know if you'd like me to do some testing.

thatguystone:
What issues were you running into? Everything has been working fine for me so far.

harshavardhana:
> What issues were you running into? Everything has been working fine for me so far.

Issues were related to certificates not being shipped in the default alpine distro @rajivm was using HTTPS and it led to problems. 

Can you validate that as well @thatguystone ?

thatguystone:
`/etc/ssl/certs/` is present in the image, and it has a bunch of certs registered. Fetching over https with curl works (it uses `/etc/ssl/certs/ca-certificates.crt`). Running the server with some generated certs also works fine.

harshavardhana:
> /etc/ssl/certs/ is present in the image, and it has a bunch of certs registered. Fetching over https with curl works (it uses /etc/ssl/certs/ca-certificates.crt). Running the server with some generated certs also works fine.

Very cool :+1:

rajivm:
👍

Issue #2597, opened 2016-09-01
docker: Document how to use docker compose and provide a sample compose yaml


harshavardhana:
```
$ cat ~/docker-compose/docker-compose.yml 
version: '2'
services:
  minio1:
    image: minio/minio:dist
    environment:
      - MINIO_ACCESS_KEY=abcd123
      - MINIO_SECRET_KEY=abcd12345
    ports:
      - "9001:9000"
    volumes:
      - /tmp/1:/export
    command: minio1:/export minio2:/export minio3:/export minio4:/export

  minio2:
    image: minio/minio:dist
    environment:
      - MINIO_ACCESS_KEY=abcd123
      - MINIO_SECRET_KEY=abcd12345
    ports:
      - "9002:9000"
    volumes:
      - /tmp/2:/export      
    command: minio1:/export minio2:/export minio3:/export minio4:/export

  minio3:
    image: minio/minio:dist
    environment:
      - MINIO_ACCESS_KEY=abcd123
      - MINIO_SECRET_KEY=abcd12345
    ports:
      - "9003:9000"
    volumes:
      - /tmp/3:/export      
    command: minio1:/export minio2:/export minio3:/export minio4:/export


  minio4:
    image: minio/minio:dist
    environment:
      - MINIO_ACCESS_KEY=abcd123
      - MINIO_SECRET_KEY=abcd12345
    ports:
      - "9004:9000"
    volumes:
      - /tmp/4:/export      
    command: minio1:/export minio2:/export minio3:/export minio4:/export    

```

harshavardhana:
Fixed by #2606

PR #2606, opened 2016-09-01
Add docker-compose file to run Minio in distributed mode

donatello:
Serves as a starting point to run a Minio cluster using Docker. The
file can be used as configuration for the docker-compose tool to start
4 Minio servers in distributed mode.
- Add a docker-compose.yml file to run 4 minio server instances in
  distributed mode
- Update Docker.md with command to use the file

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,46 @@
+version: '2'
+
+# starts 4 docker containers running minio server instances. Each
+# minio server's web interface will be accessible on the host at port
+# 9001 through 9004.
+services:
+  minio1:
+    image: minio/minio
+    ports:
+      - "9001:9000"
+    volumes:
+      - /mnt/export/minio1:/export
+    environment:
+      MINIO_ACCESS_KEY: abcd1
+      MINIO_SECRET_KEY: abcd1234
+    command: minio1:/export minio2:/export minio3:/export minio4:/export
+  minio2:
+    image: minio/minio
+    ports:
+      - "9002:9000"
+    volumes:
+      - /mnt/export/minio2:/export
+    environment:
+      MINIO_ACCESS_KEY: abcd1
+      MINIO_SECRET_KEY: abcd1234
+    command: minio1:/export minio2:/export minio3:/export minio4:/export
+  minio3:
+    image: minio/minio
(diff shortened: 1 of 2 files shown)


donatello:
Fixes #2597

glensc:
how is this supposed to work? seems to give error as it can't resolve the hostnames.

it failed to start giving me error:

```
FATA[0002] Unable to intialize object layer.             cause=dial tcp: missing port in address minio1
```

so i modified `docker-compose.yml` to use just `/export` as path and entered the container to debug:

```
# docker-compose exec minio1 sh

/go/src/app # minio server minio1:/export --address 9001
FATA[0002] Unable to intialize object layer.             cause=dial tcp: missing port in address minio1
/go/src/app # ping minio1
PING minio1 (172.18.0.2): 56 data bytes
64 bytes from 172.18.0.2: seq=0 ttl=64 time=0.069 ms
^C
--- minio1 ping statistics ---
1 packets transmitted, 1 packets received, 0% packet loss
round-trip min/avg/max = 0.069/0.069/0.069 ms
/go/src/app # minio version
Version: 2016-09-11T17:42:18Z
Release-Tag: RELEASE.2016-09-11T17-42-18Z
Commit-ID: 85e2d886bcb005d49f3876d6849a2b5a55e03cd3
/go/src/app # 
```

ps: where this cluster mode is documented? i.e that argumend can be host:/path not just local /path?

also, shouldn't the containers be [linked together](https://docs.docker.com/compose/compose-file/#/links) to be access each other's tcp ports?

harshavardhana:
> how is this supposed to work? seems to give error as it can't resolve the hostnames.

This is only for the `minio:edge` tag not for the 'latest' if use change the 'minio/minio:edge' you will be able to use run distributed mode.

glensc:
could you also add link to documentation here (about distributed mode) in case somebody else stumbles here (i.e by looking at docker-composer.yml file history)

harshavardhana:
> could you also add link to documentation here (about distributed mode) in case somebody else stumbles here (i.e by looking at docker-composer.yml file history)

We are in process of updating the doc for the time being and updated docker-compose.yml as well to reflect the changes for now.

PR #2609, opened 2016-09-02
notifications: Fix bucket notifications for DeleteMultipleObjects.

harshavardhana:
Now reports bucket notifications for DeleteMultipleObjects API.
Also deletes multiple objects in parallel.


krishnasrinivas:
:+1:

PR #2633, opened 2016-09-07
S3-layer: Multiple-delete should return success for non-existent objects.

### LINKED ISSUE #2630: Minio returns an error code when deleting a non-existent object through DeleteMultiObjects API
Hello, we use Minio as a replacement for S3 during development. This behavior cropped up when we ran an existing unit test against Minio. When deleting an object that doesn't exist, we normally expect to receive no error code. It seems as though Minio's response provides an error code (`NoSuchKey`) which causes an exception in the Java SDK.

As an aside... I took a look at the code and it doesn't seem like the error gets propagated...

``` Go
/// http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectDELETE.html
/// Ignore delete object errors, since we are suppposed to reply
/// only 204.
if err := api.ObjectAPI.DeleteObject(bucket, object); err != nil {
    writeSuccessNoContent(w)
    return
}
```

So this might be a side effect of how the SDK is interpreting the response from Minio.
## Expected behaviour

When running against Amazon AWS with Java SDK version 1.11.8, the response is populated with the deleted objects and no errors are put into the response list. See the response object populated from Amazon AWS' response below:

Amazon AWS Response
## Actual behaviour

The SDK generates a `MultiObjectDeleteException` after receiving `NoSuchKey` in errors list... so it seems like Minio does the opposite. See the response object populated from Minio's response below:

Minio Response
1. Try to delete a non-existent object.
## Minio version

Version: 2016-08-21T02:44:47Z
Release-Tag: RELEASE.2016-08-21T02-44-47Z
Commit-ID: 975eb319730c8db093b4744bf9e012356d61eef2
## System information
- Windows 7
- AWS Java SDK 1.11.8

krishnasrinivas:
Fixes #2630

### DIFF
--- a/cmd/bucket-handlers.go
+++ b/cmd/bucket-handlers.go
@@ -267,6 +267,12 @@ func (api objectAPIHandlers) DeleteMultipleObjectsHandler(w http.ResponseWriter,
 			deletedObjects = append(deletedObjects, object)
 			continue
 		}
+		if _, ok := err.(ObjectNotFound); ok {
+			// If the object is not found it should be
+			// accounted as deleted as per S3 spec.
+			deletedObjects = append(deletedObjects, object)
+			continue
+		}
 		errorIf(err, "Unable to delete object. %s", object.ObjectName)
 		// Error during delete should be collected separately.
 		deleteErrors = append(deleteErrors, DeleteError{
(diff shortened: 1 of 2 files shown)


harshavardhana:
LGTM, Thanks :+1:

PR #2643, opened 2016-09-09
Signature-V4: Include content-length for signature calculation.

krishnasrinivas:
This is to be compatible with clients which includes content-length for signature calculation
(and hence deviating from AWS Signature-v4 spec)

### DIFF
--- a/cmd/signature-v4.go
+++ b/cmd/signature-v4.go
@@ -342,8 +342,21 @@ func doesSignatureMatch(hashedPayload string, r *http.Request, validateRegion bo
 		return ErrContentSHA256Mismatch
 	}
 
+	header := req.Header
+
+	// Signature-V4 spec excludes Content-Length from signed headers list for signature calculation.
+	// But some clients deviate from this rule. Hence we consider Content-Length for signature
+	// calculation to be compatible with such clients.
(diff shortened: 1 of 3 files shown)


krishnasrinivas:
copyHeader() is done as mutating req.Header did not look clean.

cmd/signature-v4.go
```diff
@@ -26,15 +26,14 @@ package cmd
...
+	"crypto/sha256"
 	"encoding/hex"
 	"net/http"
 	"net/url"
 	"sort"
 	"strconv"
 	"strings"
 	"time"
-
-	"github.com/minio/sha256-simd"
```
harshavardhana, line 37: You should use 'sha256-simd' here, do you not have a local copy?

cmd/utils.go
```diff
@@ -30,6 +31,18 @@ import (
 	"github.com/pkg/profile"
 )
 
+// make a copy of http.Header
+func cloneHeader(h http.Header) http.Header {
+	h2 := make(http.Header, len(h))
```
harshavardhana, line 36: add a test perhaps for this..

harshavardhana:
> copyHeader() is done as mutating req.Header did not look clean.

:+1:

PR #2654, opened 2016-09-10
Return Owner information in ListObjectV2 only when fetch-owner is spe…

### LINKED ISSUE #2647: s3 list objects v2 owner
## Expected behaviour

AWS S3 does not return the object owner with list objects v2
## Actual behaviour

minio returns the object owner with list objects v2
## Minio version

Version: 2016-08-21T02:44:47Z
Release-Tag: RELEASE.2016-08-21T02-44-47Z
Commit-ID: 975eb319730c8db093b4744bf9e012356d61eef2

vadmeste:
…cified

Fixes #2647


harshavardhana:
Can you add a test here @vadmeste ?

PR #2668, opened 2016-09-11
Avoid hiding disk errors in some cases in FS Shutdown

vadmeste:


harshavardhana:
Can we add a testcase here @vadmeste ?

vadmeste:
@harshavardhana, the patch is ready

PR #2676, opened 2016-09-12
tests: tests for NewMultipart Upload HTTP handler.

hackintoshrao:


cmd/object-handlers_test.go
```diff
@@ -438,3 +440,171 @@ func testAPICopyObjectHandler(obj ObjectLayer, instanceType, bucketName string,
...
+		}
+		// compare the uploadID obtained from response with uploadID obtained from parsing upload.json .
+		// Assert the number of uploadID entries for the object in upload.json .
+
+		if len(testUploads.uploads) != len(uploadsJSON.Uploads) {
+			t.Fatalf("Minio %s: Expected record of %d uploadID's in upload.json for given object, but got %d.", instanceType, len(testUploads.uploads), len(uploadsJSON.Uploads))
+		}
+		// Entry for the upload ID's generated from the response should be found in upload.json .
+
+	CheckXlUploadID:
```
harshavardhana, line 573: ??? - we should avoid these..  is this really needed?

cmd/object-handlers_test.go
```diff
@@ -438,3 +440,171 @@ func testAPICopyObjectHandler(obj ObjectLayer, instanceType, bucketName string,
...
+		if err != nil {
+			t.Fatalf("Reading uploads json from FS backend failed: <ERROR> %s.", err)
+		} // compare the uploadID obtained from response with uploadID obtained from parsing upload.json .
+		// Assert the number of uploadID entries for the object in upload.json .
+
+		if len(testUploads.uploads) != len(uploadsJSON.Uploads) {
+			t.Fatalf("Minio %s:Expected record of %d uploadID's in upload.json for given object, but got %d.", instanceType, len(testUploads.uploads), len(uploadsJSON.Uploads))
+		}
+		// Entry for the upload ID's generated from the response should be found in upload.json .
+	CheckFsUploadID:
```
harshavardhana, line 596: ??? - we should avoid these..  is this really needed?

cmd/object-handlers_test.go
```diff
@@ -438,3 +440,171 @@ func testAPICopyObjectHandler(obj ObjectLayer, instanceType, bucketName string,
...
+	// Wait till all go routines finishes execution.
+	wg.Wait()
+
+	// verify the existance and correctness of upload.json.
+	var uploadsJSON uploadsV1
+	var err error
+	switch v := obj.(type) {
+	case xlObjects:
+		// read the upload.json for xl backend.
+		uploadsJSON, err = v.readUploadsJSON(bucketName, objectName)
```
harshavardhana, line 561: Calling this handler test we shouldn't use internal functions.. this is a bad idea and hard to extend these tests.  What is the purpose here?

cmd/object-handlers_test.go
```diff
@@ -438,3 +440,171 @@ func testAPICopyObjectHandler(obj ObjectLayer, instanceType, bucketName string,
...
+	err = decoder.Decode(multipartResponse)
+	if err != nil {
+		t.Fatalf("Error decoding the recorded response Body")
+	}
+	// verify the existance and correctness of upload.json.
+	var uploadsJSON uploadsV1
+	switch v := obj.(type) {
+	case xlObjects:
+		// read the upload.json for xl backend.
+		uploadsJSON, err = v.readUploadsJSON(bucketName, objectName)
```
harshavardhana, line 482: We should only validate indeed if the response was correct we don't need to see if the backend was read properly.  In this situation we are skipping locking and all other forms of sanity checks leading to a bad test.

cmd/object-handlers_test.go
hackintoshrao: Need to break away from outer loop too.

cmd/object-handlers_test.go
hackintoshrao: Will remove validation of upload.json .

hackintoshrao:
Closing the PR, will be making it a short tests without the upload.json assertions, Will send it along with other Multipart handler tests.

PR #2686, opened 2016-09-14
tests: Implemented more tests for fs-v1*.go

vadmeste:
- Wrote more tests for all private functions.

PR #2693, opened 2016-09-14
Fix FS remove bucket regression bug

vadmeste:
This was detected by a test in https://github.com/minio/minio/pull/2686

@krishnasrinivas, can you review this?

### DIFF
--- a/cmd/fs-v1.go
+++ b/cmd/fs-v1.go
@@ -117,26 +117,24 @@ func (fs fsObjects) Shutdown() error {
 	_, err := fs.storage.ListDir(minioMetaBucket, mpartMetaPrefix)
 	if err != errFileNotFound {
 		// A nil err means that multipart directory is not empty hence do not remove '.minio.sys' volume.
-		// A non nil err means that an unexpected error occurred
-		return err
+		// A non nil err means that an unexpected error occured
+		return toObjectErr(traceError(err))
 	}
 	// List if there are any bucket configuration entries.
 	_, err = fs.storage.ListDir(minioMetaBucket, bucketConfigPrefix)
 	if err != errFileNotFound {
 		// A nil err means that bucket config directory is not empty hence do not remove '.minio.sys' volume.
-		// A non nil err means that an unexpected error occurred
-		return err
+		// A non nil err means that an unexpected error occured
+		return toObjectErr(traceError(err))
 	}
 	// Cleanup everything else.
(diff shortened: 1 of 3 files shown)


harshavardhana (approved):

harshavardhana:
> @krishnasrinivas, can you review this?

Do you see anything? @krishnasrinivas

krishnasrinivas:
> Do you see anything? @krishnasrinivas

checking

cmd/object-common.go
```diff
@@ -225,7 +225,7 @@ func cleanupDir(storage StorageAPI, volume, dirPath string) error {
 		if err == errFileNotFound {
 			return nil
 		} else if err != nil { // For any other errors fail.
-			return traceError(err)
```
krishnasrinivas, line 228: Better to do traceError() at a point where the error happens, because if we decide enhance traceError() to add more context, for ex:

traceError(err, fmt.Sprintf("failed to ListDir on %s %s", volume, entryPath))

it can't be done if we move it to a different place.

what was the reason you moved it down? to avoid stack info for a recursive call?

harshavardhana: :+1:

cmd/fs-v1.go
```diff
@@ -118,14 +118,14 @@ func (fs fsObjects) Shutdown() error {
 	if err != errFileNotFound {
 		// A nil err means that multipart directory is not empty hence do not remove '.minio.sys' volume.
 		// A non nil err means that an unexpected error occured
-		return err
+		return toObjectErr(traceError(err))
```
krishnasrinivas, line 121: toObjectErr() will be useful only for s3 calls, ie it will be used to construct error response XML, as Shutdown() is an internal call we need not do toObjectErr(), but it is better to be consistent code wise i.e from the object layer we always return object-error not posix error. So doing toObjectErr() is fine.

Since we are returning trace info in the error now, you can delete the errorIf() statements in this function and instead do one errorIf() at the caller. (routers.go)

cmd/object-common.go
vadmeste: > what was the reason you moved it down? to avoid stack info for a recursive call?

I just wanted to simplify. In this case, I guess I should make something like `return traceError(storage.DeleteFile(volume, entryPath))` in L219 (little above)

vadmeste:
this should be ready for merging

Issue #2678, opened 2016-09-12
ServeHTTP Rate-limit bug in acquire()


donatello:
In [rate-limit-handler.go](https://github.com/minio/minio/blob/master/cmd/rate-limit-handler.go) (at this time of writing), there is a function `acquire()`, intended to allow at most `globalMaxConn` connections to be processed simultaneously, and to limit the number of waiting connections to `globalMaxConn*4`:

``` go
// acquire and release implement a way to send and receive from the
// channel this is in-turn used to rate limit incoming connections in
// ServeHTTP() http.Handler method.
func (c *rateLimit) acquire() error {
    // Kick out clients when it is really crowded
    if len(c.waitQueue) == cap(c.waitQueue) {
        return errTooManyRequests
    }

    // Add new element in waitQueue to keep track of clients
    // wanting to process their requests
    c.waitQueue <- struct{}{}

    // Moving from wait to work queue is protected by a mutex
    // to avoid draining waitQueue with multiple simultaneous clients.
    c.lock.Lock()
    c.workQueue <- <-c.waitQueue
    c.lock.Unlock()
    return nil
}
```

Consider the scenario, that `globalMaxConn = 1`, and `len(c.waitQueue) == 3`. Now assume, 2 connections simultaneously enter the `acquire()` method - both may pass the `if` block (`len(c.waitQueue) == 3 and cap(c.waitQueue) == 4`), then, one will enter wait queue, and the other will block waiting on the same queue.

This a bug because the intended semantics appears to be that no more than `globalConn * 4` connections should be waiting, but in this case there will be 5 waiting. With only one spot available in the queue, if 10 connections come in simultaneously, all 10 could be waiting instead of 9 of them getting rate-limit exceeded errors back.

Proposed fix:

``` go
func (c *rateLimit) acquire() error {
    c.lock.Lock()
    // Kick out clients when it is really crowded
    if len(c.waitQueue) == cap(c.waitQueue) {
        defer c.lock.Unlock() //don't forget to unlock
        return errTooManyRequests
    }

    // Add new element in waitQueue to keep track of clients
    // wanting to process their requests
    c.waitQueue <- struct{}{}
    c.lock.Unlock()

    c.workQueue <- <-c.waitQueue
    return nil
}
```

We prevent two go-routines from simultaneously checking for a spot in the queue and entering it.

No synchronization for moving an item from the `waitQueue` to `workQueue` is needed, as it is thread safe already.

donatello:
[...]

PR #2694, opened 2016-09-14
Lock-free rate-limit algorithm + bug-fix

donatello:
The fix for issue #2678 in commit https://github.com/minio/minio/pull/2679/commits/3ba086ca4c48fc3dfc5294b41184cfa74d87ce9f has a bug in this line:

``` go
    c.workQueue <- <-c.waitQueue
```

Every connection that enters `c.waitQueue` will come to this step, get dequeued and block to enter `c.workQueue`. The number of such waiters is unbounded, as after each dequeue, another slot in the wait queue becomes available.

Instead of some form of lock to resolve this problem, a cleaner solution without locks was found. This PR has the implementation.

Please review @vadmeste @harshavardhana @krisis.

### DIFF
--- a/cmd/rate-limit-handler.go
+++ b/cmd/rate-limit-handler.go
@@ -19,7 +19,6 @@ package cmd
 import (
 	"errors"
 	"net/http"
-	"sync"
 )
 
 var errTooManyRequests = errors.New("Too many clients in the waiting list")
@@ -28,7 +27,6 @@ var errTooManyRequests = errors.New("Too many clients in the waiting list")
 // limit the number of concurrent http requests.
 type rateLimit struct {
 	handler   http.Handler
-	lock      sync.Mutex
 	workQueue chan struct{}
 	waitQueue chan struct{}
 }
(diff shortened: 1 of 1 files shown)


harshavardhana:
This is clean

cmd/rate-limit-handler.go
```diff
@@ -37,26 +35,26 @@ type rateLimit struct {
...
-	// waiting connections.)
-	c.lock.Unlock()
-
-	// Block to put a waiting go-routine into processing mode.
-	c.workQueue <- <-c.waitQueue
+	//block attempting to enter the workQueue. If the workQueue is
+	//full, there can be at most cap(waitQueue) == 4*globalMaxConn
+	//goroutines waiting here because of the select above.
+	select {
+	case c.workQueue <- struct{}{}:
```
krisis, line 53: Neat!

krisis:
LGTM

vadmeste:
I am going to test with httperf

harshavardhana (approved):

vadmeste:
with httperf, this is how to send 10000 requests, with 1000 each second. 
`httperf --server localhost --port 9000 --rate=1000 --num-conns=10000 --uri=/testbucket/file`. 

It is working perfectly @donatello :+1:

donatello:
@vadmeste Thanks for the testing and httpperf example!

PR #2707, opened 2016-09-15
XL and FS use different tree walk ignored errors

vadmeste:
I removed these errors from FS walk tree ignored errors: errDiskNotFound, errDiskAccessDenied and  errFaultyDisk.

### DIFF
--- a/cmd/fs-v1.go
+++ b/cmd/fs-v1.go
@@ -38,6 +38,12 @@ type fsObjects struct {
 	listPool *treeWalkPool
 }
 
+// list of all errors that can be ignored in tree walk operation in FS
+var fsTreeWalkIgnoredErrs = []error{
+	errFileNotFound,
+	errVolumeNotFound,
+}
+
 // creates format.json, the FS format info in minioMetaBucket.
 func initFormatFS(storageDisk StorageAPI) error {
 	return writeFSFormatData(storageDisk, newFSFormatV1())
@@ -602,7 +608,7 @@ func (fs fsObjects) ListObjects(bucket, prefix, marker, delimiter string, maxKey
 			// object string does not end with "/".
 			return !strings.HasSuffix(object, slashSeparator)
 		}
-		listDir := listDirFactory(isLeaf, fs.storage)
+		listDir := listDirFactory(isLeaf, fsTreeWalkIgnoredErrs, fs.storage)
 		walkResultCh = startTreeWalk(bucket, prefix, marker, recursive, listDir, isLeaf, endWalkCh)
 	}
 	var fileInfos []FileInfo
--- a/cmd/tree-walk.go
+++ b/cmd/tree-walk.go
@@ -21,15 +21,6 @@ import (
 	"strings"
 )
 
-// list of all errors that can be ignored in tree walk operation.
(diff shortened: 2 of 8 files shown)


cmd/fs-v1.go
```diff
@@ -111,6 +111,14 @@ func newFSObjects(disk string) (ObjectLayer, error) {
 	return fs, nil
 }
 
+// getDiskIgnoredErrors - returns the list of disk errors that we can ignore
```
harshavardhana: Why a function like this? why not 

```
var fsTreeWalkIgnoredErrs = 
```

cmd/tree-walk_test.go
```diff
@@ -29,6 +29,15 @@ import (
 // Fixed volume name that could be used across tests
 const volume = "testvolume"
 
+// list of all errors that can be ignored in tree walk operation.
+var walkResultIgnoredErrs = []error{
```
harshavardhana: This is redundant we can keep this naming as 

```
var walkResultIgnoredErrsXL = 
```

and 

```
var walkResultIgnoredErrsFS =
```

cmd/xl-v1-common.go
```diff
@@ -62,10 +62,22 @@ func (xl xlObjects) isObject(bucket, prefix string) (ok bool) {
...
 			continue
 		}
 		errorIf(err, "Unable to stat a file %s/%s/%s", bucket, prefix, xlMetaJSONFile)
 	} // Exhausted all disks - return false.
 	return false
 }
+
+// treeWalkIgnoredErrs - returns the list of tree walk errors that we can
+// ignore when we have a quorum
+func (xl xlObjects) treeWalkIgnoredErrs() []error {
```
harshavardhana: Making these functions don't help really.

harshavardhana (changes requested):

cmd/tree-walk_test.go
krishnasrinivas: your earlier suggestion : `var fsTreeWalkIgnoredErrs =` is more readable i.e fsTreeWalkIgnoredErrs and xlTreeWalkIgnoredErrs

cmd/tree-walk.go
```diff
@@ -21,15 +21,21 @@ import (
...
 	errDiskAccessDenied,
 	errFaultyDisk,
 }
 
+// list of all errors that can be ignored in tree walk operation in FS
+var fsTreeWalkIgnoredErrs = []error{
+	errFileNotFound,
+	errVolumeNotFound,
+}
+
```
krishnasrinivas: It is better these two are declared in fs-v1\* xl-v1\* files. because if you see tree-walk.go is abstracted from xl or fs

cmd/tree-walk_test.go
harshavardhana: The reason to keep it this way is for 'emacs and vim' auto completes so that its easier to find these, but i guess either way is fine.

cmd/tree-walk.go
harshavardhana: Yeah agreed.

vadmeste:
are we good for merge?

harshavardhana (approved):

PR #2708, opened 2016-09-15
Upgrade to new dsync version incl. stale lock detection

fwessels:
Upgrade to the latest dsync version that includes amongst others a stale lock detection mechanism.

Other changes:
- Compatible behaviour of DRWMutex to sync.Mutex (so can call multiple RLocks() on same object)

dsync improvements:
- Stale lock detection mechanism
- Uid added to (r)lock message
- Added node & rpcPath from originator to (r)lock message so it is known to which server to query back
- Added new test cases for single & multiple stale locks, and single and multiple clients with locks that crash

### DIFF
--- a/cmd/lock-rpc-server.go
+++ b/cmd/lock-rpc-server.go
@@ -39,8 +39,8 @@ type LockArgs struct {
 	Token     string
 	Timestamp time.Time
 	Node      string
-	RpcPath   string
-	Uid       string
+	RPCPath   string
+	UID       string
 }
 
 // SetToken - sets the token to the supplied value.
@@ -115,7 +115,7 @@ func (l *lockServer) Lock(args *LockArgs, reply *bool) error {
 	}
 	_, *reply = l.lockMap[args.Name]
 	if !*reply { // No locks held on the given name, so claim write lock
-		l.lockMap[args.Name] = []lockRequesterInfo{lockRequesterInfo{writer: true, node: args.Node, rpcPath: args.RpcPath, uid: args.Uid, timestamp: time.Now(), timeLastCheck: time.Now()}}
+		l.lockMap[args.Name] = []lockRequesterInfo{lockRequesterInfo{writer: true, node: args.Node, rpcPath: args.RPCPath, uid: args.UID, timestamp: time.Now(), timeLastCheck: time.Now()}}
 	}
 	*reply = !*reply // Negate *reply to return true when lock is granted or false otherwise
 	return nil
@@ -136,10 +136,10 @@ func (l *lockServer) Unlock(args *LockArgs, reply *bool) error {
 	if *reply = isWriteLock(lri); !*reply { // Unless it is a write lock
 		return fmt.Errorf("Unlock attempted on a read locked entity: %s (%d read locks active)", args.Name, len(lri))
 	}
-	if l.removeEntry(args.Name, args.Uid, &lri) {
+	if l.removeEntry(args.Name, args.UID, &lri) {
 		return nil
 	}
-	return fmt.Errorf("Unlock unable to find corresponding lock for uuid: %s", args.Uid)
+	return fmt.Errorf("Unlock unable to find corresponding lock for uid: %s", args.UID)
 }
 
 // RLock - rpc handler for read lock operation.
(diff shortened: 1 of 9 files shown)


cmd/lock-rpc-server.go
```diff
@@ -158,26 +169,134 @@ func (l *lockServer) RUnlock(args *LockArgs, reply *bool) error {
...
+		var active bool
+
+		// Call back to original server verify whether the lock is still active (based on name & uid)
+		if err := c.Call("Dsync.Active", &LockArgs{Name: nlrip.name, UID: nlrip.lri.uid}, &active); err != nil {
+			// We failed to connect back to the server that originated the lock, this can either be due to
+			// - server at client down
+			// - some network error (and server is up normally)
+			//
+			// We will ignore the error, and we will retry later to get resolve on this lock
+			log.Println("  Dsync.Active failed:", err)
```
harshavardhana: We can use 'errorIf()' here which prints the stack info as well rather than using 'log.Println()'

cmd/lock-rpc-server.go
```diff
@@ -158,26 +169,134 @@ func (l *lockServer) RUnlock(args *LockArgs, reply *bool) error {
...
+			if !active { // The lock is no longer active at server that originated the lock
+				// so remove the lock from the map
+				l.mutex.Lock()
+				// Check if entry is still in map (could have been removed altogether by 'concurrent' (R)Unlock of last entry)
+				if lri, ok := l.lockMap[nlrip.name]; ok {
+					if !l.removeEntry(nlrip.name, nlrip.lri.uid, &lri) {
+						// Remove failed, in case it is a:
+						if nlrip.lri.writer {
+							// Writer: this should never happen as the whole (mapped) entry should have been deleted
+							log.Println("Lock maintenance failed to remove entry for write lock (should never happen)", nlrip.name, nlrip.lri, lri)
```
harshavardhana: Same here.

cmd/lock-rpc-server.go
```diff
@@ -158,26 +169,134 @@ func (l *lockServer) RUnlock(args *LockArgs, reply *bool) error {
...
+							// has been released concurrently (so it is fine)
+						}
+					} else {
+						// remove went okay, all is fine
+					}
+				}
+				l.mutex.Unlock()
+			}
+		}
+	}
```
harshavardhana, line 297: Can we bring the test case from dsync here for server side testing?

cmd/lock-rpc-server.go
fwessels: Done

cmd/lock-rpc-server.go
fwessels: Done

cmd/lock-rpc-server.go
fwessels: That would be nice, but we'd have to think about how to do this (you would have to fire up multiple 'minio server's on localhost, perform operations on it so that distributed locks are used and then kill servers and wait for maintenance to occur, and retry again).

cmd/lock-rpc-server.go
```diff
@@ -269,11 +269,10 @@ func (l *lockServer) lockMaintenance(interval time.Duration) {
 			// - some network error (and server is up normally)
 			//
 			// We will ignore the error, and we will retry later to get resolve on this lock
-			log.Println("  Dsync.Active failed:", err)
+			log.Infoln("Dsync.Active failed:", err)
```
harshavardhana: @fwessels we have a function called `errorIf` for these purposes in logger.go

fwessels: This is not an error perse, in fact it can legitimately happen while either a server to be contacted is down or the network is down (and during a subsequent run of maintenance, this lock will be retried).

So this is more of an 'informative' message (and hence the 'Infoln'), or we should silently remove it altogether.

harshavardhana: Even in that case errorIf() can give context here. Perhaps prints the stack as well upon, MINIO_TRACE enabling to see where exactly or who is calling this.

harshavardhana: Since we don't use generic 'log' anywhere in the code base these logs would look inconsistent with logrus logs.That is why `errorIf` can solve two purposes here.

fwessels: I'll remove the logging line then because the user really shouldn't be interested in the actual error.

harshavardhana: Okay makes sense.

PR #2718, opened 2016-09-16
lock: Make some cleanup and moving the code around.

harshavardhana:
This patch just avoids lot of ifs and inverts some logic.

### DIFF
--- a/cmd/lock-rpc-server-common.go
+++ b/cmd/lock-rpc-server-common.go
@@ -1,0 +1,85 @@
+/*
+ * Minio Cloud Storage, (C) 2016 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cmd
+
+import (
+	"errors"
+	"time"
+)
+
+// Similar to removeEntry but only removes an entry only if the lock entry exists in map.
+func (l *lockServer) removeEntryIfExists(nlrip nameLockRequesterInfoPair) {
+	// Check if entry is still in map (could have been removed altogether by 'concurrent' (R)Unlock of last entry)
+	if lri, ok := l.lockMap[nlrip.name]; ok {
+		if !l.removeEntry(nlrip.name, nlrip.lri.uid, &lri) {
+			// Remove failed, in case it is a:
+			if nlrip.lri.writer {
+				// Writer: this should never happen as the whole (mapped) entry should have been deleted
+				errorIf(errors.New(""), "Lock maintenance failed to remove entry for write lock (should never happen)", nlrip.name, nlrip.lri.uid, lri)
+			} // Reader: this can happen if multiple read locks were active and
+			// the one we are looking for has been released concurrently (so it is fine).
+		} // Removal went okay, all is fine.
+	}
+}
+
+// removeEntry either, based on the uid of the lock message, removes a single entry from the
+// lockRequesterInfo array or the whole array from the map (in case of a write lock or last read lock)
+func (l *lockServer) removeEntry(name, uid string, lri *[]lockRequesterInfo) bool {
+	// Find correct entry to remove based on uid.
+	for index, entry := range *lri {
+		if entry.uid == uid {
+			if len(*lri) == 1 {
+				// Remove the (last) lock.
+				delete(l.lockMap, name)
+			} else {
+				// Remove the appropriate read lock.
+				*lri = append((*lri)[:index], (*lri)[index+1:]...)
+				l.lockMap[name] = *lri
+			}
+			return true
+		}
+	}
+	// None found return false, perhaps entry removed in previous run.
+	return false
+}
+
+// Validate lock args.
+func (l *lockServer) validateLockArgs(args *LockArgs) error {
+	if !l.timestamp.Equal(args.Timestamp) {
+		return errInvalidTimestamp
+	}
+	if !isRPCTokenValid(args.Token) {
+		return errInvalidToken
+	}
+	return nil
+}
+
+// getLongLivedLocks returns locks that are older than a certain time and
+// have not been 'checked' for validity too soon enough
+func getLongLivedLocks(m map[string][]lockRequesterInfo, interval time.Duration) []nameLockRequesterInfoPair {
+	rslt := []nameLockRequesterInfoPair{}
+	for name, lriArray := range m {
+		for idx := range lriArray {
+			// Check whether enough time has gone by since last check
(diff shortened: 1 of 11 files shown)


harshavardhana:
@krisis @fwessels - your reviews.. needed

cmd/lock-rpc-server.go
```diff
@@ -189,14 +215,14 @@ func (l *lockServer) Active(args *LockArgs, reply *bool) error {
...
-	var lri []lockRequesterInfo
-	if lri, *reply = l.lockMap[args.Name]; !*reply {
-		return nil // No lock is held on the given name so return false
-	}
-	// Check whether uid is still active
-	for _, entry := range lri {
-		if *reply = entry.uid == args.UID; *reply {
-			return nil // When uid found return true
+	lri, ok := l.lockMap[args.Name]
+	// No lock is held on the given name so return false
```
fwessels: Comment not longer valid

cmd/lock-rpc-server.go
```diff
@@ -250,49 +276,46 @@ func getLongLivedLocks(m map[string][]lockRequesterInfo, interval time.Duration)
...
-							// has been released concurrently (so it is fine)
-						}
-					} else {
-						// remove went okay, all is fine
-					}
-				}
-				l.mutex.Unlock()
+		}
+		c.Close()
+		if !active {
```
fwessels: With `active` initialized to false (line 288) and in case of any errors for the Dsync.Active call, we would now remove the lock (!)

Locks should only be removed when `err == nil && !active`

cmd/lock-rpc-server.go
```diff
@@ -149,15 +160,30 @@ func (l *lockServer) RLock(args *LockArgs, reply *bool) error {
...
 		return err
 	}
-	var lri []lockRequesterInfo
-	lri, *reply = l.lockMap[args.Name]
-	if !*reply { // No locks held on the given name, so claim (first) read lock
-		l.lockMap[args.Name] = []lockRequesterInfo{{writer: false, node: args.Node, rpcPath: args.RPCPath, uid: args.UID, timestamp: time.Now(), timeLastCheck: time.Now()}}
-		*reply = true
-	} else {
+	lri, ok := l.lockMap[args.Name]
+	if ok {
```
fwessels: Can combine these two lines into single if statement

cmd/lock-rpc-server.go
```diff
@@ -189,14 +215,14 @@ func (l *lockServer) Active(args *LockArgs, reply *bool) error {
...
-	if lri, *reply = l.lockMap[args.Name]; !*reply {
-		return nil // No lock is held on the given name so return false
-	}
-	// Check whether uid is still active
-	for _, entry := range lri {
-		if *reply = entry.uid == args.UID; *reply {
-			return nil // When uid found return true
+	lri, ok := l.lockMap[args.Name]
+	// No lock is held on the given name so return false
+	if ok {
```
fwessels: Can combine into single if statement

fwessels (changes requested):
See comments, especially removal of stale locks (line 303)

harshavardhana:
Patch has been updated still writing some tests..

harshavardhana:
Added tests as well..

cmd/lock-rpc-server-common.go
```diff
@@ -0,0 +1,85 @@
...
+// removeEntry either, based on the uid of the lock message, removes a single entry from the
+// lockRequesterInfo array or the whole array from the map (in case of a write lock or last read lock)
+func (l *lockServer) removeEntry(name, uid string, lri *[]lockRequesterInfo) bool {
+	// Find correct entry to remove based on uid.
+	for index, entry := range *lri {
+		if entry.uid == uid {
+			if len(*lri) == 1 {
+				// Remove the (last) lock.
+				delete(l.lockMap, name)
+				continue
```
fwessels: Should not continue here (leads to `return false`) -- remove `continue`, put lines 50-52 in `else` and always `return true` in line 53

cmd/lock-rpc-server.go
```diff
@@ -125,39 +185,44 @@ func (l *lockServer) Lock(args *LockArgs, reply *bool) error {
...
+	if err := l.validateLockArgs(args); err != nil {
 		return err
 	}
-	var lri []lockRequesterInfo
-	lri, *reply = l.lockMap[args.Name]
-	if !*reply { // No locks held on the given name, so claim (first) read lock
-		l.lockMap[args.Name] = []lockRequesterInfo{{writer: false, node: args.Node, rpcPath: args.RPCPath, uid: args.UID, timestamp: time.Now(), timeLastCheck: time.Now()}}
-		*reply = true
-	} else {
+	lrInfo := lockRequesterInfo{
```
fwessels, line 211: Better like this with single instantiation of variable 👍

cmd/lock-rpc-server.go
```diff
@@ -227,132 +272,41 @@ type nameLockRequesterInfoPair struct {
...
-				// Start with random sleep time, so as to avoid "synchronous checks" between servers
-				time.Sleep(time.Duration(rand.Float64() * float64(lockMaintenanceLoop)))
-				for {
-					time.Sleep(lockMaintenanceLoop)
-					locker.lockMaintenance(lockCheckValidityInterval)
-				}
-			}()
-
-			lockServers = append(lockServers, locker)
+		err := c.Call("Dsync.Active", &LockArgs{
```
fwessels: If we negate the function of 'Dsync.Active', thus instead do a 'Dsync.Expired', then we can do as follows (and get rid of `err` altogether)

```
var expired bool
// Purposely ignore error
c.Call("Dsync.Expired", &LockArgs{.....}, &expired)
c.Close()
if expired { .....
}
```

cmd/namespace-lock.go
```diff
@@ -205,9 +207,10 @@ func (n *nsLockMap) unlock(volume, path, opsID string, readLock bool) {
 // allocated name space lock or initializing a new one.
 func (n *nsLockMap) Lock(volume, path, opsID string) {
 	var lockOrigin string
-	// lock debugging enabled. The caller information of the lock held has be obtained here before calling any other function.
+	// Lock debugging enabled. The caller information of the lock held has
+	// be obtained here before calling any other function.
```
fwessels: be --> been

cmd/namespace-lock.go
```diff
@@ -229,10 +232,11 @@ func (n *nsLockMap) Unlock(volume, path, opsID string) {
 func (n *nsLockMap) RLock(volume, path, opsID string) {
 	var lockOrigin string
 	readLock := true
-	// lock debugging enabled. The caller information of the lock held has be obtained here before calling any other function.
+	// Lock debugging enabled. The caller information of the lock held has
+	// be obtained here before calling any other function.
```
fwessels: be --> been

fwessels (changes requested):
See comments, nice addition of test cases.

See also the idea using `Dsync.Expired` instead of `Dsync.Active`, will test this in `minio/dsync(/chaos)` as well

harshavardhana:
> See comments, nice addition of test cases.
> 
> See also the idea using Dsync.Expired instead of Dsync.Active, will test this in minio/dsync(/chaos) as well

Made the relevant changes minus `Dsync.Expired` change.

fwessels:
See https://github.com/minio/dsync/pull/26 for (tested) `DSync.Expired` implementation.

harshavardhana:
All changes in @fwessels

fwessels:
Looks good to me.

Issue #2690, opened 2016-09-14
api/completeMultipart: Incorrect HTTP response for certain cases. 


hackintoshrao:
## Expected behaviour

To adhere to the spec as for the cases below. 

<img width="901" alt="screen shot 2016-09-14 at 9 29 23 am" src="https://cloud.githubusercontent.com/assets/2609511/18512289/0c69577a-7aa6-11e6-988b-40c65252af3b.png">
## Actual behaviour

200 OK response.

Will send the fix along with https://github.com/minio/minio/issues/2671 .

Issue #2671, opened 2016-09-11
[distributed]: Tests for Complete Multipart object Handler.


hackintoshrao:
Need tests for Complete Multipart  handler from object-handler-tests.

harshavardhana:
ping @hackintoshrao - which one you are working on .. so that these bugs can be assigned properly.

Issue #2669, opened 2016-09-11
[distributed]: Tests for DeleteObject handler. 


hackintoshrao:
Need tests for delete object handler, from `object-handler-tests`.

harshavardhana:
ping @hackintoshrao - which one you are working on .. so that these bugs can be assigned properly.

PR #2719, opened 2016-09-16
api/complete-multipart: fixes and tests.

hackintoshrao:
- Removing the logic of sending white space characters during complete multipart upload. 
- This simplifies the code and helps to meet the requirements mentioned at https://github.com/minio/minio/issues/2690 in a cleaner way.
- Tests for New Multipart Upload and Complete Multipart Upload API handlers.
- Tests for Delete Object Handler. 
- Fixes https://github.com/minio/minio/issues/2690 and https://github.com/minio/minio/issues/2671  and https://github.com/minio/minio/issues/2669 . 

With detailed discussion with @krishnasrinivas , we found out this approach to be the simplest amongst the other alternatives.

### DIFF
--- a/cmd/api-response-multipart.go
+++ b/cmd/api-response-multipart.go
@@ -20,6 +20,21 @@ package cmd
 
 import "net/http"
 
+// Represents additional fields necessary for ErrPartTooSmall S3 error.
+type completeMultipartAPIError struct {
+	// Proposed size represents uploaded size of the part.
+	ProposedSize int64
+	// Minimum size allowed epresents the minimum size allowed per
+	// part. Defaults to 5MB.
+	MinSizeAllowed int64
+	// Part number of the part which is incorrect.
+	PartNumber int
+	// ETag of the part which is incorrect.
+	PartETag string
+	// Other default XML error responses.
+	APIErrorResponse
+}
+
 // writeErrorResponsePartTooSmall - function is used specifically to
 // construct a proper error response during CompleteMultipartUpload
 // when one of the parts is < 5MB.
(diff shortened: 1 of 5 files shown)


krishnasrinivas (commented):
@hackintoshrao @harshavardhana the complication for completemultipart upload is arising because we have to accomodate the sendwhitespace() function. sendwhitespace() is not useful for XL, it was written because of slow concat process of FS. But since we now do append in background as parts come, FS's completemultipart() is also quick now. That means we can just remove sendwhitespace() functionality and avoid all complications.

Downside is we will not send whitespace chars for the slow concat fall back in FS. But it's a compromise we can make to keep code simple. Because anyhow we saw cloudberry timeout even if we sent white space chars. So for well behaved client uploads we work just fine.

harshavardhana:
> @hackintoshrao @harshavardhana the complication for completemultipart upload is arising because we have to accomodate the sendwhitespace() function. sendwhitespace() is not useful for XL, it was written because of slow concat process of FS. But since we now do append in background as parts come, FS's completemultipart() is also quick now. That means we can just remove sendwhitespace() functionality and avoid all complications.
> 
> Downside is we will not send whitespace chars for the slow concat fall back in FS. But it's a compromise we can make to keep code simple. Because anyhow we saw cloudberry timeout even if we sent white space chars. So for well behaved client uploads we work just fine.

I agree that is what i was leaning towards, we are okay to not do that since sending whitespace character is a S3 spec and as we have seen with experience by now the doc SPEC is not the direct representation of their implementation. 

Let's remove the go-routine at the top and we can avoid all the complications that came with it.

cmd/api-response-multipart.go
```diff
@@ -20,6 +20,21 @@ package cmd
 
 import "net/http"
 
+// Represents additional fields necessary for ErrPartTooSmall S3 error.
+type completeMultipartAPIError struct {
```
harshavardhana, line 24: Do not introduce these custom errors just yet, i am fixing this differently. For now do not validate other fields other than the common ones that occur in `APIErrorResponse`

cmd/object-handlers.go
```diff
@@ -806,28 +837,19 @@ func (api objectAPIHandlers) CompleteMultipartUploadHandler(w http.ResponseWrite
...
-
-	// Signal that completeMultipartUpload is over via doneCh
+	// Signal that joining parts is over is over via doneCh
 	go func(doneCh chan<- struct{}) {
-		md5Sum, err = objectAPI.CompleteMultipartUpload(bucket, object, uploadID, completeParts)
+		md5Sum, err = joinParts()
 		doneCh <- struct{}{}
 	}(doneCh)
 
 	sendWhiteSpaceChars(w, doneCh)
```
harshavardhana: Let's remove this and we will solve all the problems and we will not need to change the function signature of `CompleteMultipartUpload`

hackintoshrao: So I'll keep the CompleteMultipart API signature as it was and remove the sending of white space characters and running complete multipart in a new routine.

cmd/api-response-multipart.go
hackintoshrao: OK, for now I'll make changes to send usual error response.

PR #2731, opened 2016-09-19
server: Add more elaborate startup messages.

harshavardhana:
These messages based on our prep stage during XL
and prints more informative message regarding
drive information.


harshavardhana:
ping @krisis can you review?

cmd/prepare-storage.go
```diff
@@ -49,16 +52,15 @@ func init() {
...
   +----------+--------------------------+-----------------------+
   | All      | None Formatted           |  FormatDisks          |
   |	     |				|  and initObjectLayer  |
   |          | 		             	|                   	|
   +----------+--------------------------+-----------------------+
-  |	     |      			|  Wait for notify from |
-  | Quorum   | 				|  "Heal via minioctl"  |
-  |          | Quorum UnFormatted   	|                   	|
+  |	     |      			|                       |
+  | Quorum   | Quorum UnFormatted   	|      Failed        	|
```
krisis: What does `Failed` signify?

cmd/prepare-storage.go
```diff
@@ -114,90 +116,199 @@ func prepForInit(disks []string, sErrs []error, diskCount int) InitActions {
...
+}
+
+func retryFormattingDisks(firstDisk bool, storageDisks []StorageAPI) error {
+	if storageDisks == nil {
+		return errInvalidArgument
+	}
 	nextBackoff := time.Duration(0)
 	var err error
-	done := false
+	var done bool
```
krisis: Why is the above preferred over `done := false`? As per [Go's memory model](https://golang.org/ref/mem), initialization of a variable with its type's zero value is considered as a write. This makes both these statements operationally equivalent.

cmd/prepare-storage.go
```diff
@@ -114,90 +116,199 @@ func prepForInit(disks []string, sErrs []error, diskCount int) InitActions {
...
-		// Some disks possibly corrupted.
+		// Some of the formatted disks are possibly corrupted or unformatted, heal them.
 		return WaitForHeal
+	} // No quorum wait for quorum number of disks.
+	return WaitForQuorum
+}
+
+type printFunc func(msg string)
+
+func newPrintFn() printFunc {
```
krisis: newPrintFn doesn't convey its purpose. printOnce seems like a good alternative.

cmd/prepare-storage.go
```diff
@@ -114,90 +116,199 @@ func prepForInit(disks []string, sErrs []error, diskCount int) InitActions {
...
+				console.Printf(
+					"Initializing data volume. Waiting for minimum %d servers to come online.\n",
+					len(storageDisks)/2+1,
+				)
+			case WaitForAll:
+				console.Println("Initializing data volume for first time. Waiting for other servers to come online.")
+			case WaitForFormatting:
+				console.Println("Initializing data volume for first time. Waiting for first server to come online.")
 			}
+			nextBackoff++
```
krisis: Could we cycle through backoff times that are either distributed randomly within an upper-bound OR a fixed slice of suitably separated integers?

cmd/server-main.go
```diff
@@ -342,7 +311,7 @@ func serverMain(c *cli.Context) {
 	// Server address.
 	serverAddress := c.String("address")
 
-	// Check if requested port is available.
+	// Check if requesbted port is available.
```
krisis: s/requested/requebsted/

cmd/event-notifier_test.go
```diff
@@ -93,7 +93,7 @@ func TestInitEventNotifier(t *testing.T) {
 	if err != nil {
 		t.Fatal("Unable to create directories for FS backend. ", err)
 	}
-	fs, err := getSingleNodeObjectLayer(disk[0])
+	fs, _, err := getSingleNodeObjectLayer(disk[0])
```
krisis: All callers of getSingleNodeObjectLayer and getXLObjectLayer are ignoring the returned []StorageAPI. Is it worth returning it if no one requires it now?

cmd/prepare-storage.go
harshavardhana: The reason being consistent with the above `var err error` - 'done :=' is basically allocating we don't need to do that and let compiler have the default value based on the type. 

Since it is global to the loop this seemed better.

cmd/prepare-storage.go
harshavardhana: This function is not ready yet, it needs to do `printk` style messaging.

cmd/server-main.go
harshavardhana: done

cmd/event-notifier_test.go
harshavardhana: It is required in `makeTestBackend()` for controller RPC tests.

cmd/prepare-storage.go
harshavardhana: Right now its crude, perhaps in future we can do jitter based backoff.

cmd/prepare-storage.go
harshavardhana: 'Abort'  - need to rename it.

harshavardhana:
ping can you review again @krisis .

PR #2735, opened 2016-09-19
configMigrate() returns errors + tests

### LINKED ISSUE #2659: [distributed] tests: Add tests for config migration code
https://codecov.io/gh/minio/minio/src/8acf4d112a01f3071d008ae3bc559e2ec40a6dee/cmd/config-migrate.go

For more awareness.

vadmeste:
Fixes #2659

### DIFF
--- a/cmd/config-migrate_test.go
+++ b/cmd/config-migrate_test.go
@@ -1,0 +1,196 @@
+/*
+ * Minio Cloud Storage, (C) 2016 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cmd
+
+import (
+	"io/ioutil"
+	"os"
+	"strconv"
+	"testing"
+)
+
+const lastConfigVersion = 7
+
+// TestServerConfigMigrateV1 - tests if a config v1 is purged
+func TestServerConfigMigrateV1(t *testing.T) {
+	rootPath, err := newTestConfig("us-east-1")
+	if err != nil {
+		t.Fatalf("Init Test config failed")
+	}
+	// remove the root folder after the test ends.
+	defer removeAll(rootPath)
+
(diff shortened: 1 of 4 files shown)


vadmeste:
Increase of 2.5% in code coverage is suspected. Please don't merge this.

harshavardhana:
> Increase of 2.5% in code coverage is suspected. Please don't merge this.

Yes won't merge, add WIP in front.

vadmeste:
@harshavardhana, I've found that tests execute fatalIf (then os.Exit()) in config.go and that should be bad for testing itself (it seems go test just exits happily without continuing to test everything)

harshavardhana:
> @harshavardhana, I've found that tests execute fatalIf (then os.Exit()) in config.go and that should be bad for testing itself (it seems go test just exits happily without continuing to test everything)

Yes but it will only continue other non dependent tests but not the current set i believe?  - also is this patch ready for review?

vadmeste:
> Yes but it will only continue other non dependent tests but not the current set i believe?

No. The same thing happened when I wrote a unit test for shutdown callbacks. I can confirm again with a quick code.

> also is this patch ready for review?

Yes.

cmd/config-migrate_test.go
```diff
@@ -0,0 +1,196 @@
...
+	if err != nil {
+		t.Fatalf("Init Test config failed")
+	}
+	// remove the root folder after the test ends.
+	defer removeAll(rootPath)
+
+	setGlobalConfigPath(rootPath)
+	configPath := rootPath + "/" + globalMinioConfigFile
+
+	/* for i := 2; i <= lastConfigVersion; i++ {
```
harshavardhana: what is this for?

cmd/config.go
```diff
@@ -54,7 +54,9 @@ func getConfigPath() (string, error) {
 // mustGetConfigPath must get server config path.
 func mustGetConfigPath() string {
 	configPath, err := getConfigPath()
-	fatalIf(err, "Unable to get config path.")
+	if err != nil {
```
harshavardhana, line 57: You should still log 'errorIf()'

cmd/main.go
```diff
@@ -157,8 +159,13 @@ func checkMainSyntax(c *cli.Context) {
 func Main() {
 	app := registerApp()
 	app.Before = func(c *cli.Context) error {
+
+		configDir := c.GlobalString("config-dir")
+		if configDir == "" {
+			fatalIf(errors.New("config-dir is empty"), "Unable to get config file.")
```
harshavardhana: Rename this as `fatalf(errors.New("Config directory is empty"), "Unable to get config file.")`

cmd/config.go
vadmeste: > You should still log 'errorIf()'

Currently, minio doesn't work if it fails to get config path, and main.go will show an error anyway.

PR #2754, opened 2016-09-21
web-handlers: support for presigned-get json-rpc call for MinioBrowser's "share" feature.

krishnasrinivas:
@harshavardhana implementing presigned-get in minio-browser is more complicated than implementing it in minio-server hence done this way.

### DIFF
--- a/cmd/signature-v4.go
+++ b/cmd/signature-v4.go
@@ -27,8 +27,10 @@ package cmd
 import (
 	"bytes"
 	"encoding/hex"
+	"fmt"
 	"net/http"
 	"net/url"
+	"path"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


cmd/signature-v4.go
```diff
@@ -417,3 +419,36 @@ func doesSignatureMatch(hashedPayload string, r *http.Request, validateRegion bo
...
+		"X-Amz-Algorithm=" + signV4Algorithm,
+		"X-Amz-Credential=" + strings.Replace(credential, "/", "%2F", -1),
+		"X-Amz-Date=" + dateStr,
+		"X-Amz-Expires=" + "604800",
+		"X-Amz-SignedHeaders=host",
+	}, "&")
+
+	path := "/" + path.Join(bucket, object)
+
+	var extractedSignedHeaders http.Header // will be nil as "host" is the only signed header.
```
krisis, line 446: I couldn't understand the comment about `host`  being the only signed header. Could you explain that?

krisis (commented):
Besides the question I have asked on the comment, rest of the changes LGTM.

harshavardhana:
> I couldn't understand the comment about host being the only signed header. Could you explain that?

Host is the only header that is needed at least for presigned URLs to work.

harshavardhana:
@krishnasrinivas please add a test case.

harshavardhana:
> @harshavardhana implementing presigned-get in minio-browser is more complicated than implementing it in minio-server hence done this way.

This looks good.

PR #2756, opened 2016-09-22
web: add method to get all policies for given bucket name.

balamurugana:


harshavardhana:
Can you add a test case as well here @balamurugana ?

PR #2768, opened 2016-09-23
Minio MS Windows + SSL

Andrey198:
Good day.
Cant start minio on Windows system with SSL - cant find any documents on it or help in minio --help
Found this topic https://docs.minio.io/docs/generate-let-s-encypt-certificate-using-concert-for-minio 
but no instruction for MS Windows

Please, help.
Does minio can start and works through SSL?

Thanx!


harshavardhana:
> Does minio can start and works through SSL?

Same instructions work for Windows as well @Andrey198

Issue #2760, opened 2016-09-22
[distributed] Purge cached objects when disk has more recent copy


krisis:
## Expected behaviour
- Object layer cache could return stale copy of an object if the node receiving the GET request doesn't  receive the latest PUT request on the same object.
## Actual behaviour
- Object layer cache should purge its old copy of an object when discovered.
## Steps to reproduce the behaviour
- Setup:
  - 8 nodes each hosting a disk
- Issue a PUT object on node1
- Issue a PUT object with different data on node2
- Issue a GET object on node1
## System information

All platforms

harshavardhana:
Fixed by #2770

PR #2770, opened 2016-09-23
Purge stale object cache entry

krisis:
- Add unit test to validate that ModTime based cache entry purging works.


krisis:
Fixes #2760

PR #2788, opened 2016-09-27
Recalculate free minimum disk space

### LINKED ISSUE #2780: [fs] : Relax 5% disk usage to a more cumulative number based on the total disk capacity.
## Expected behaviour

Successful start of object-storage browser
## Actual behaviour

Error: Minimum '4' disks are required to enable erasure code

Start server on non-root HDD/non-home? 
i.e. `minio server ~/test` works. `minio server /files/test` does not. Both are newly created, empty directories. /files is a mount-point to my big hdds
## Minio version

Version: DEVELOPMENT.GOGET
Release-Tag: DEVELOPMENT.GOGET
Commit-ID: DEVELOPMENT.GOGET

Commit-ID according to git: cfbab22237cb29d06b3df7d684ab208d8bc661e7
## System information

Ubuntu 14.04 x64
Go Version 1.7.1

vadmeste:
Fixes #2780

Review is needed for this PR.

I had this use case:

```
$ df -h 
/dev/nvme0n1p7   24G   21G  1.4G  95% /mnt/p7
```

but the code considers that the disk has about 2.0Gib free, the reason is that the code is calculating the free disk space using free blocks instead of available blocks.

```
               fsblkcnt_t f_bfree;   /* Free blocks in filesystem */
               fsblkcnt_t f_bavail;  /* Free blocks available to
                                        unprivileged user */
```

One other thing, is that I noticed that we already remove 5% from the calculated free disk space (for journal and inodes). So it seems to me that we can set a fixed free disk space (100mb) instead of using a percentage calculation (or dynamic percentage calculation)

What do you think?

### DIFF
--- a/cmd/posix.go
+++ b/cmd/posix.go
@@ -33,15 +33,17 @@ import (
 )
 
 const (
-	fsMinSpacePercent = 5
-	maxAllowedIOError = 5
+	fsMinFreeSpace         = 1024 * 1024 * 1024
+	fsMinFreeInodesPercent = 5
+	maxAllowedIOError      = 5
 )
 
 // posix - implements StorageAPI interface.
 type posix struct {
-	ioErrCount  int32 // ref: https://golang.org/pkg/sync/atomic/#pkg-note-BUG
-	diskPath    string
-	minFreeDisk int64
+	ioErrCount    int32 // ref: https://golang.org/pkg/sync/atomic/#pkg-note-BUG
+	diskPath      string
+	minFreeSpace  int64
+	minFreeInodes int64
 }
 
 var errFaultyDisk = errors.New("Faulty disk")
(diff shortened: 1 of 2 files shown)


harshavardhana:
> One other thing, is that I noticed that we already remove 5% from the calculated free disk space (for journal and inodes). So it seems to me that we can set a fixed free disk space (100mb) instead of using a percentage calculation (or dynamic percentage calculation)

We can move this to '1g' instead of 100M is too small.

cmd/posix.go
```diff
@@ -101,7 +101,7 @@ func newPosix(diskPath string) (StorageAPI, error) {
 	}
 	fs := &posix{
 		diskPath:    diskPath,
-		minFreeDisk: fsMinSpacePercent, // Minimum 5% disk should be free.
+		minFreeDisk: fsMinFreeSpace,
```
harshavardhana: This logic needs to change since the previous code is using percent, so it won't work magically.

vadmeste:
Ready for review and merge

PR #2795, opened 2016-09-27
Add NATS notifier

### LINKED ISSUE #2783: Add support of nats.io for bucket notification
http://nats.io/

vadmeste:
Fixes #2783

### DIFF
--- a/cmd/notify-nats.go
+++ b/cmd/notify-nats.go
@@ -1,0 +1,108 @@
+/*
+ * Minio Cloud Storage, (C) 2016 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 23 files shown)


cmd/notify-nats.go
```diff
@@ -0,0 +1,108 @@
...
+	// Set default JSON formatter.
+	natsLog.Formatter = new(logrus.JSONFormatter)
+
+	// Successfully enabled all NATSs.
+	return natsLog, nil
+}
+
+// Fire is called when an event should be sent to the message broker
+func (n natsConn) Fire(entry *logrus.Entry) error {
+	ch := n.Conn
```
harshavardhana, line 91: Does the nats library provides a transparent reconnect logic upon network disconnections?

vadmeste: By default, it tries 60 times and waits 2 seconds between retries, not clear how to make it infinite though (even the library code is not clear enough)

PR #2803, opened 2016-09-28
build: Deprecate requirement of GOROOT

harshavardhana:


krishnasrinivas:
but play is write accessible to anyone right? so it will be risky to download binaries from here and run

PR #2805, opened 2016-09-29
signature: Region changes should be handled just like AWS.

harshavardhana:
- PutBucket happens with 'us-east-1'.
- ListBuckets happens with any region.
- GetBucketLocation happens with 'us-east-1' and location is returned.

### DIFF
--- a/cmd/auth-handler.go
+++ b/cmd/auth-handler.go
@@ -105,7 +105,7 @@ func sumMD5(data []byte) []byte {
 }
 
 // Verify if request has valid AWS Signature Version '4'.
-func isReqAuthenticated(r *http.Request) (s3Error APIErrorCode) {
+func isReqAuthenticated(r *http.Request, region string) (s3Error APIErrorCode) {
 	if r == nil {
 		return ErrInternalError
 	}
@@ -121,7 +121,6 @@ func isReqAuthenticated(r *http.Request) (s3Error APIErrorCode) {
 	}
 	// Populate back the payload.
 	r.Body = ioutil.NopCloser(bytes.NewReader(payload))
-	validateRegion := true // Validate region.
 	var sha256sum string
 	// Skips calculating sha256 on the payload on server,
 	// if client requested for it.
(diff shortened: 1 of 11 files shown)


harshavardhana:
ping @krishnasrinivas - need some careful review on this one. Have tested s3cmd, aws cli and mc with the changes.  These changes are in lieu if someone changes the default region for Minio.

krishnasrinivas:
checking.

cmd/bucket-handlers.go
```diff
@@ -197,7 +199,17 @@ func (api objectAPIHandlers) ListBucketsHandler(w http.ResponseWriter, r *http.R
 	}
 
 	// List buckets does not support bucket policies, no need to enforce it.
-	if s3Error := checkAuth(r); s3Error != ErrNone {
+	// Proceed to validate signature.
```
krishnasrinivas, line 200: step-by-step check is more readable than using anonymous function. Also errorIf can print appropriate message when `aType != authTypePresigned && aType != authTypeSigned`

cmd/bucket-handlers-listobjects.go
```diff
@@ -82,7 +82,8 @@ func (api objectAPIHandlers) ListObjectsV2Handler(w http.ResponseWriter, r *http
 			return
 		}
 	case authTypeSigned, authTypePresigned:
-		if s3Error := isReqAuthenticated(r); s3Error != ErrNone {
+		validateRegion := true // Validate region.
```
krishnasrinivas: the comment `// Validate region.` is redundant, here and at other places

cmd/auth-handler.go
```diff
@@ -105,7 +105,7 @@ func sumMD5(data []byte) []byte {
 }
 
 // Verify if request has valid AWS Signature Version '4'.
-func isReqAuthenticated(r *http.Request) (s3Error APIErrorCode) {
+func isReqAuthenticated(r *http.Request, validateRegion bool) (s3Error APIErrorCode) {
```
krishnasrinivas: request needs to be validated for `us-east-1` region in case of putbucket listbuckets getbucketlocation, hence instead of validateRegion, this function can take a region string directly. For putbucket listbuckets getbucketlocation calls we pass `us-east-1` and for other calls we pass `serverConfig.GetRegion()`

krishnasrinivas (changes requested):

cmd/auth-handler.go
harshavardhana: :+1:

cmd/bucket-handlers.go
harshavardhana: errorIf is not useful since we are going to send back AccessDenied in such a case. Let me making a readable function.

PR #2813, opened 2016-09-30
Verify signature early without waiting for sha256 calculation.

krishnasrinivas:
sha256 is calculated and verified along with md5sum at the object layer.

Tests for sha256 match/mismatch needs to be added.

### DIFF
--- a/cmd/api-errors.go
+++ b/cmd/api-errors.go
@@ -617,6 +617,8 @@ func toAPIErrorCode(err error) (apiErr APIErrorCode) {
 		apiErr = ErrNoSuchUpload
 	case PartTooSmall:
 		apiErr = ErrEntityTooSmall
+	case SHA256Mismatch:
+		apiErr = ErrContentSHA256Mismatch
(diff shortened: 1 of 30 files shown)


harshavardhana:
@krishnasrinivas can you add a test case here ? simply setting a sha value would clearly help in existing test cases.

krishnasrinivas:
@harshavardhana yes, it was still WIP,
now i have added valid/invalid sha256 test case to testObjectAPIPutObjectPart and testObjectAPIPutObject

i will do another round of refactor in a different commit - md5sum/sha256 calculation has to be abstracted out of XL/FS somehow in a clean way without affecting functionality.

harshavardhana:
> i will do another round of refactor in a different commit - md5sum/sha256 calculation has to be abstracted out of XL/FS somehow in a clean way without affecting functionality.

We can take a io.ReadCloser() instead of io.Reader and wrap it around so that a Close() validates internally both md5 and sha as well.

PR #2819, opened 2016-10-01
web: fix jwt token expiry set to one day by default.

### LINKED ISSUE #2818: [distributed] Browser token is set to 100yrs fix it.

balamurugana:
Fixes #2818


krisis (commented):
How do clients renegotiate their jet tokens on expiry?

balamurugana:
> How do clients renegotiate their jet tokens on expiry?

In current code, JWT token is created with 100 years as expiry for inter-node communication.  So it never expires in server/cluster lifetime.

Currently this patch is not complete due to removal of  inter-node JWT having 100 years as expiry.

PR #2824, opened 2016-10-03
Postgres notifier

donatello:
Ready for review.


cmd/config-migrate.go
```diff
@@ -434,6 +435,11 @@ func migrateV7ToV8() error {
 	} else {
 		srvConfig.Notify.Redis = cv7.Notify.Redis
 	}
+	if len(cv7.Notify.PostgreSQL) == 0 {
```
harshavardhana: This is dangerous and should be avoided,  implement a proper config migration. do not edit configs like this.  Migrate to v9.

cmd/notify-postgresql.go
```diff
@@ -0,0 +1,208 @@
...
+			"Records": entry.Data["Records"],
+		})
+		if err != nil {
+			return fmt.Errorf(
+				"Unable to encode event %v to JSON - got error - %v",
+				entry.Data["Records"], err,
+			)
+		}
+
+		// attempt to insert into the table
```
harshavardhana: Why does this say "attempt" ? - isn't this atomic?. Does this retry transparently if the attempt fails?

cmd/notify-postgresql.go
```diff
@@ -0,0 +1,208 @@
...
+	// TODO: check that table exists?
+
+	// create prepared statements
+	stmts := make(map[string]*sql.Stmt)
+	// insert or update statement
+	stmts["upsertRow"], err = db.Prepare(fmt.Sprintf(upsertRow, pgN.Table))
+	if err != nil {
+		return pgConn{}, errors.New("Unable to create upsert prepared statement for postgres notification.")
+	}
+	stmts["deleteRow"], err = db.Prepare(fmt.Sprintf(deleteRow, pgN.Table))
```
harshavardhana, line 167: Use the same error or embed this into a newly formed error. using `fmt.Errorf`

harshavardhana (changes requested):

cmd/config-migrate.go
donatello: Fixed, but updating the test code.

cmd/notify-postgresql.go
donatello: Fixed, and tried to make errors in this section uniform with `fmt.Errorf`

cmd/notify-postgresql.go
donatello: Fixed this comment. It is atomic, and failures are only related to connection problems and other un-controllables.

donatello:
Going to add table creation as well if table is not present at start up.

donatello:
@harshavardhana Now table is created if it does not exist at startup.

harshavardhana (approved):

PR #2862, opened 2016-10-06
Refactor streaming signatureV4 w/ state machine

krisis:
- Used state machine to make transitions between reading chunk header,
  chunk data and trailer explicit.


krisis:
The difference in behaviour from existing streaming chunk reader `Read` implementation is that existing one could return `n != 0` when `err == io.EOF`. I don't yet know if it could have caused the sporadic issue we have observed. This PR attempts to make the implementation easy to reason with and debug by printing the (explicit) state transitions in the process of streaming signature verification.

harshavardhana:
This change looks good @krisis . Can you address the coverage we can take this in ?.

PR #2877, opened 2016-10-07
Peers RPC for Bucket Notification and Listeners

donatello:

Issue #2772, opened 2016-09-24
[distributed] notifications and policy config in distributed XL


krishnasrinivas:
notifications and policy config should be applied to all the minio servers in a cluster.

harshavardhana:
> if notification is applied to only one minio server in d-XL, objects uploaded to other minio servers will not send notification event.

Glad you opened it, was working on a fix for this.

harshavardhana:
> notifications and policy config should be applied to all the minio servers in a cluster.

Notifications are written to the backend so the notification.xml is erasure coded and cached as well. So all nodes do indeed see the same configuration. 

There is another problem of double caching at object-handler layer which i am removing as its not needed.  On the other hand there is one more bug which Aditya is fixing which is ListenBucketNotification() in a distributed mode.

harshavardhana:
Fixed by #2891 and #2877

PR #2891, opened 2016-10-10
Use Peer RPC to propagate bucket policy changes

donatello:
This PR as of now works on top of https://github.com/minio/minio/pull/2877

It is intended to resolve the second part of #2772 - bucket policy changes propagation.

### DIFF
--- a/cmd/bucket-policy-handlers.go
+++ b/cmd/bucket-policy-handlers.go
@@ -182,7 +182,7 @@ func (api objectAPIHandlers) PutBucketPolicyHandler(w http.ResponseWriter, r *ht
 	}
 
 	// Save bucket policy.
-	if err = PutBucketPolicy(bucket, policy, objAPI); err != nil {
+	if err = persistAndNotifyBucketPolicyChange(bucket, policyChange{false, policy}, objAPI); err != nil {
 		switch err.(type) {
 		case BucketNameInvalid:
 			writeErrorResponse(w, r, ErrInvalidBucketName, r.URL.Path)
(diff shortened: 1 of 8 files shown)


donatello:
Currently tests related to the changes have been disabled.

harshavardhana:
> Currently tests related to the changes have been disabled.

Let me know when they are added back will start review or would you prefer reviewing without tests?

donatello:
@harshavardhana I think you can start reviewing. I've broken down the changes in both the PRs - this PR builds on the PR for changes on bucket notifications and listen bucket. The changes are broken into separate commits for functionality and tests fix (mainly disabling for now).

Also, the CI system(s) seems to have gotten confused because i modified some older commits and force pushed it. When I compile all changes together on my machine it works fine.

I've also done some basic testing locally and it appears to work.

The problem with tests is that now it needs a somewhat big refactor - both FS and XL now use the same mechanism to notify about changes in policy/notifications - a local RPC call is needed. The tests didn't require this previously. I am still working out how best to achieve this.

cmd/bucket-policy-handlers.go
```diff
@@ -192,13 +192,38 @@ func (api objectAPIHandlers) PutBucketPolicyHandler(w http.ResponseWriter, r *ht
...
+// removed.
+func PutBucketPolicy(bucket string, bktpolicy *bucketPolicy, objAPI ObjectLayer) error {
+	// FIXME: Race exists between the bucket existence check and
+	// then updating the bucket policy.
+	if err := isBucketExist(bucket, objAPI); err != nil {
+		return err
+	}
+
+	// persist to object layer
+	if bktpolicy == nil {
```
harshavardhana: I think using 'nil' as an indicator for 'removeBucketPolicy' is a bad idea as it doesn't convey anything one has to dig deeper to know what is the correct behavior. 

It is better kept in two separate functions and called by the caller separately.

cmd/bucket-policy.go
```diff
@@ -202,16 +196,23 @@ func removeBucketPolicy(bucket string, objAPI ObjectLayer) error {
...
+	buf, err := json.Marshal(bpy)
+	if err != nil {
+		errorIf(err, "Unable to marshal bucket policy '%v' to JSON", *bpy)
+		return err
+	}
 	policyPath := pathJoin(bucketConfigPrefix, bucket, policyJSON)
-	sha256sum := ""
-	if _, err := objAPI.PutObject(minioMetaBucket, policyPath, size, reader, nil, sha256sum); err != nil {
+	if _, err := objAPI.PutObject(minioMetaBucket, policyPath,
+		int64(len(buf)), bytes.NewReader(buf), nil, ""); err != nil {
```
harshavardhana: Tabbing and indentation like this is not needed in Golang and uncommon.  Keep the indentation as before.

cmd/s3-peer-client.go
```diff
@@ -174,3 +174,15 @@ func S3PeersUpdateBucketListener(bucket string, lcfg []listenerConfig) {
...
 	}
 }
+
+// S3PeersUpdateBucketPolicy - Sends update bucket policy request to
+// all peers. Currently we log an error and continue.
+func S3PeersUpdateBucketPolicy(bucket string, bply *bucketPolicy) {
+	setBPPArgs := &SetBPPArgs{Bucket: bucket, PCfg: bply}
+	peers := globalS3Peers.GetPeers()
+	errsMap := globalS3Peers.SendRPC(peers, "S3.SetBucketPolicyPeer",
+		setBPPArgs)
```
harshavardhana: Same here, there is no 80 line rule in Golang.

cmd/bucket-policy-handlers.go
donatello: Ok I've stopped using `nil` as a zero value here and added a `.IsEmpty` for the struct to identify an empty policy. This lets the behaviour be uniform even in case an empty policy was sent to PutBucketPolicy API. Additionally, we repeat less code by not introducing a high level remove policy function (there is already one at the lower level to delete the policy from storage).

cmd/bucket-policy.go
donatello: Done, though I do have some reservations.

cmd/s3-peer-client.go
donatello: Also done.

cmd/bucket-policy.go
harshavardhana: This is only true for other languages in Go it is expected. Unless `gofmt` does this we shouldn't manually indent like this.

cmd/bucket-policy-handlers.go
harshavardhana: A PutBucketPolicy cannot have an empty body, this signifies that it can and we treat it like RemoveBucketPolicy.  This is not the same as bucket notification we shouldn't convey that in the flow. It is okay to have a separate remove bucket policy.  Any code here pretty much has to convey the S3 spec.

donatello:
@harshavardhana I've taken a final whack at it while trying to still not repeat code. The helper function is renamed to now be more explicit in its purpose, and it takes an extra argument to indicate a policy deletion.

web-handlers are also now fixed to notify peers - something I had forgotten earlier.

Since policy changes are done via 4 places (two api handlers, and two places from the web-handler), I think it is better to use a single function that combines the (racy) `isBucketExists` check, the persistence to storage, and the notification to peers.

PR #2892, opened 2016-10-10
Fix dockerfile container image.

harshavardhana:


Dockerfile
```diff
@@ -1,19 +1,17 @@
...
-	go-wrapper install && \
+	go-wrapper install -ldflags "$(go run buildscripts/gen-ldflags.go)" && \
 	mkdir -p /export/docker && \
 	cp /go/src/app/docs/Docker.md /export/docker/ && \
 	rm -rf /go/pkg /go/src && \
 	apk del git
 
 EXPOSE 9000
-ENTRYPOINT ["go-wrapper", "run", "server"]
+ENTRYPOINT ["go-wrapper", "run"]
```
olalonde, line 16: Maybe add a default `CMD`? this commit broke my docker-compose setup. Took me  a while to figure out to because it was only failing on my CI server which pulled a fresh minio/minio image.

I think this might work:

``` bash
CMD ["server"]
```

harshavardhana: We had to make this breaking change since minio has other sub-commands, it was our mistake that "server" was included as part of the entrypoint earlier. 

To run docker properly now the right command is 

```
docker run -p 9000:9000 minio/minio server /export
```

PR #2910, opened 2016-10-12
Handle err returned by rpc.Server.RegisterName

### LINKED ISSUE #2897: `rpc.Server.RegisterName` can return error but we ignore it
https://golang.org/pkg/net/rpc/#RegisterName

This function is used in multiple places in the code - https://github.com/minio/minio/search?l=go&q=RegisterName&type=Code&utf8=%E2%9C%93 but error checking is not being done. This needs to be fixed.

krisis:
Fixes #2897

### DIFF
--- a/cmd/control-router.go
+++ b/cmd/control-router.go
@@ -79,7 +79,7 @@ type controlAPIHandlers struct {
 }
 
 // Register control RPC handlers.
-func registerControlRPCRouter(mux *router.Router, srvCmdConfig serverCmdConfig) {
+func registerControlRPCRouter(mux *router.Router, srvCmdConfig serverCmdConfig) (err error) {
 	// Initialize Control.
 	ctrlHandlers := &controlAPIHandlers{
 		ObjectAPI:      newObjectLayerFn,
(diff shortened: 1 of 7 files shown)


cmd/control-router.go
```diff
@@ -79,7 +79,7 @@ type controlAPIHandlers struct {
 }
 
 // Register control RPC handlers.
-func registerControlRPCRouter(mux *router.Router, srvCmdConfig serverCmdConfig) {
+func registerControlRPCRouter(mux *router.Router, srvCmdConfig serverCmdConfig) (err error) {
```
bobinette, line 82: Do you really need a named return here? You do not `defer` nor use naked returns

krisis: @bobinette nice catch. It doesn't serve its intended purpose in this function (and in `registerDistNSLockRouter`). I use the named return approach in `configureServerHandler` function, where `err` is assigned possibly non-nil values inside an if-block, to avoid declaring `var err error` outside of the if-block.
I shall address this comment in my next update to this PR.

bobinette: :+1: 
Do you mean here: https://github.com/minio/minio/pull/2910/files#diff-66094e7896517f1e92ccf82a355c267dR70? It seems you do not use it either :wink:
I also wonder why you'd rather declare the `var err error` before the if block rather than `err := ...` since you always `return` right after when `err != nil`

krisis: Yes, that too. I have updated my PR.

donatello:
LGTM.

krisis:
> Can you add tests for each condition here ? You have to call each register\* funcs twice to generate an error

@harshavardhana calling register\* functions twice will not fail, since each call creates a new set of `rpc.Server`. It is not possible to trigger rpc.RegisterName invocations in our code base to return non-nil error AFAICS.

PR #2911, opened 2016-10-12
Listen bucket notification for multiple prefixes/suffixes

### LINKED ISSUE #2871: Feature request: Listen bucket notification for multiple suffixes
Currently, listenBucketNotification only accepts single 'suffix', single 'prefix' query params. We should take a list and allow listening on them.

wendigo:
Fixes #2871

### DIFF
--- a/cmd/bucket-notification-handlers.go
+++ b/cmd/bucket-notification-handlers.go
@@ -252,18 +252,14 @@ func (api objectAPIHandlers) ListenBucketNotificationHandler(w http.ResponseWrit
 	// Parse listen bucket notification resources.
 	prefixes, suffixes, events := getListenBucketNotificationResources(r.URL.Query())
 
-	for _, prefix := range prefixes {
-		if !IsValidObjectPrefix(prefix) {
-			writeErrorResponse(w, r, ErrFilterValueInvalid, r.URL.Path)
-			return
-		}
+	if err := validateFilterValues(prefixes); err != ErrNone {
+		writeErrorResponse(w, r, err, r.URL.Path)
+		return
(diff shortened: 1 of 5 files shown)


harshavardhana:
This is good Thanks. Will review shortly.

cmd/bucket-notification-handlers.go
```diff
@@ -250,10 +250,20 @@ func (api objectAPIHandlers) ListenBucketNotificationHandler(w http.ResponseWrit
...
+	prefixes, suffixes, events := getListenBucketNotificationResources(r.URL.Query())
+
+	for _, prefix := range prefixes {
+		if !IsValidObjectPrefix(prefix) {
+			writeErrorResponse(w, r, ErrFilterValueInvalid, r.URL.Path)
+			return
+		}
+	}
+
+	for _, suffix := range suffixes {
```
harshavardhana: We could move this into a single function. Which validates and returns 'true/false' like following. 

```
validateFilterValues(values []string) APIErrorCode {
      for _, value := range values {
               if !IsValidObjectPrefix(value) {
                         return ErrFilterValueInvalid
               } 
      }
      return ErrNone
}
```

Now you can use this cleanly 

```
if apiErr := validateFilterValues(prefixes); apiErr != ErrNone {
              writeErrorResponse(w, r, apiErr, r.URL.Path)
}
```

This way you can add test cases for "validateFilterValues()"

wendigo: Sure, refactored + added tests for new function

wendigo:
Thx! :)

PR #2929, opened 2016-10-14
Propagate creds change to cluster (Fixes #2855)

### LINKED ISSUE #2855: [distributed] browser: Need to propagate change password to all minio servers in the cluster

donatello:


cmd/web-handlers.go
```diff
@@ -351,27 +351,65 @@ func (web *webAPIHandlers) SetAuth(r *http.Request, args *SetAuthArgs, reply *Se
...
+	// Notify all Minio Peers to update credentials
+	errsMap := updateCredsOnPeers(cred)
+	var peerErr *json2.Error
+	if len(errsMap) != 0 {
+		peerErr = &json2.Error{
+			Message: "Could not update credentials on one or more peers.",
+			Data:    errsMap,
+		}
+	}
+
```
krisis, line 371: if `updateCredsOnPeers` fail on one ore more peers, why do we save the new credential in the local config file?

cmd/web-handlers.go
```diff
@@ -351,27 +351,65 @@ func (web *webAPIHandlers) SetAuth(r *http.Request, args *SetAuthArgs, reply *Se
...
-		return &json2.Error{Message: err.Error()}
+		return appendErr(
+			peerErr,
+			&json2.Error{Message: err.Error()},
+		)
 	}
 	reply.Token = token
 	reply.UIVersion = miniobrowser.UIVersion
-	return nil
+	return peerErr
```
krisis: `net/rpc` expects rpc handlers to return `nil` for error if the call is successful. For readers to grasp that this code path represents success, return `nil` explicitly instead of relying on `peerErr` to hold `nil` value.

cmd/web-handlers.go
donatello: It is somewhat arbitrary, as we cannot say how many peers may succeed in updating the credentials. If there are one or more failures, there is a partition - nodes with the new password and nodes with the old. Nodes in the same partition may communicate with each other but not with nodes outside the partition.

Here we are just being optimistic and trying to maximize the number of servers where the password change takes place. At the end we will display a message to the user saying on what servers the update could not be made.

cmd/web-handlers.go
donatello: Ah! I think this is going to help solve the bug I have on this one! :+1:

cmd/web-peer-rpc.go
```diff
@@ -0,0 +1,145 @@
...
+	// A result type for responses from peers
+	type callResult struct {
+		ix  int
+		err error
+	}
+
+	// Channel to collect results from goroutines
+	resChan := make(chan callResult)
+
+	// Launch go routines to send request to each peer in
```
harshavardhana, line 86: To make this cleaner and simpler use waitgroups and then make a 'callResult' a slice of 'peers' in size.  Take a look at how we do this in inside XL.

cmd/web-handlers.go
```diff
@@ -351,26 +352,60 @@ func (web *webAPIHandlers) SetAuth(r *http.Request, args *SetAuthArgs, reply *Se
...
+	errsMap := updateCredsOnPeers(cred)
+	// Collect errors from peers
+	peerErrMsgs := []string{}
+	for peer, errVal := range errsMap {
+		peerErrMsgs = append(
+			peerErrMsgs,
+			fmt.Sprintf("Password change error on %s - got: %v", peer, errVal),
+		)
+	}
+	combineErr := func(newErr *json2.Error) *json2.Error {
```
harshavardhana: Move this out and make this a testable code like 'reduceErrs()'

cmd/web-rpc-router.go
```diff
@@ -0,0 +1,44 @@
...
+ * limitations under the License.
+ */
+
+package cmd
+
+import (
+	"net/rpc"
+
+	router "github.com/gorilla/mux"
+)
```
harshavardhana, line 23: All of these needs comments.. on why this is done and implemented.

cmd/web-handlers.go
```diff
@@ -354,58 +354,68 @@ func (web *webAPIHandlers) SetAuth(r *http.Request, args *SetAuthArgs, reply *Se
...
-	// Collect errors from peers
-	peerErrMsgs := []string{}
-	for peer, errVal := range errsMap {
-		peerErrMsgs = append(
-			peerErrMsgs,
-			fmt.Sprintf("Password change error on %s - got: %v", peer, errVal),
+	unexpErrsMsg := "ALERT: Unexpected error(s) happened - please check the server logs."
+	gaveUpMsg := func(errMsg error, moreErrors bool) *json2.Error {
+		msg := fmt.Sprintf(
+			("ALERT: We gave due to: '%s', but there were more errors. Please check the server logs."),
```
harshavardhana: "We gave up" ?

donatello: Oops - fixed!

Issue #2832, opened 2016-10-03
[fs] warn users about buckets being skipped in a more friendlier manner.


harshavardhana:
http://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html 
- Bucket names must be at least 3 and no more than 63 characters long.
- Bucket names must be a series of one or more labels. Adjacent labels are separated by a single period (.). Bucket names can contain lowercase letters, numbers, and hyphens. Each label must start and end with a lowercase letter or a number.
- Bucket names must not be formatted as an IP address (e.g., 192.168.5.4).
- When using virtual hosted–style buckets with SSL, the SSL wildcard certificate only matches buckets that do not contain periods. To work around this, use HTTP or write your own certificate verification logic. We recommend that you do not use periods (".") in bucket names.

We apply these restrictions on Minio server as well, in the case of 'filesystem' backend which can run on existing data. We skip the buckets which do not conform to the above restrictions. 

We need to print a nice friendlier message indicating that we are skipping these non-compliant directories.

harshavardhana:
This is already fixed by @MikeRalphson in master. Closing.

PR #2947, opened 2016-10-16
Add link explaining bucket name restrictions

MikeRalphson:
Does this count as friendlier? Or is the intention to explain the actual rule the bucket-name has failed to validate against? Ref #2832

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+go_import_path: github.com/minio/minio
 sudo: required
 
 dist: trusty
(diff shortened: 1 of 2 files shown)


harshavardhana:
@MikeRalphson - Thanks for sending the fix. But the issue is not related to changing this error string. Instead it is about printing a message when ListBuckets() filters out invalid bucket entries.. 

Basically saying something like 

"Found invalid bucket names skipping them" as a informative message.  But note that this is only valid for FS backend which serves existing files and folders.

MikeRalphson:
@harshavardhana Many thanks for the clarification.

harshavardhana:
Thanks for the patch @MikeRalphson merged with some local changes. #hacktoberfest

PR #2964, opened 2016-10-17
Avoid returning disk corrupted by servers in the middle of init all disks formats

### LINKED ISSUE #2948: [distributed] minio bootstrap fails with 16 disks
I tried to understand what's going wrong, but I've tested minio with 16 disks using the attached docker-compose but bootstrapping is always failing (15 disks and 1 failed)..

I've noticed this in the log
time="2016-10-16T18:23:34Z" level=fatal msg="formatting storage disks failed" cause="corrupted backend format"
[docker-log.txt](https://github.com/minio/minio/files/532025/docker-log.txt)

Here is the docker-compose file (I renamed it to .txt because github won't accept yml as an attachement)
[docker-compose.txt](https://github.com/minio/minio/files/532019/docker-compose.txt)

You need to remove all export dirs before testing.

I've checked docker-compose.yml multiple times.. but please do too.

vadmeste:
Fix #2948

### DIFF
--- a/cmd/prepare-storage.go
+++ b/cmd/prepare-storage.go
@@ -165,7 +165,7 @@ func prepForInitXL(firstDisk bool, sErrs []error, diskCount int) InitActions {
 	// Total disks unformatted are in quorum verify if we have some offline disks.
 	if disksUnformatted >= quorum {
 		// Some disks offline and some disks unformatted, wait for all of them to come online.
-		if disksUnformatted+disksOffline == diskCount {
+		if disksUnformatted+disksFormatted+disksOffline == diskCount {
 			return WaitForAll
 		}
 		// Some disks possibly corrupted and too many unformatted disks.


cmd/prepare-storage.go
```diff
@@ -165,7 +165,7 @@ func prepForInitXL(firstDisk bool, sErrs []error, diskCount int) InitActions {
 	// Total disks unformatted are in quorum verify if we have some offline disks.
 	if disksUnformatted >= quorum {
 		// Some disks offline and some disks unformatted, wait for all of them to come online.
-		if disksUnformatted+disksOffline == diskCount {
+		if disksUnformatted+disksFormatted+disksOffline == diskCount {
 			return WaitForAll
 		}
 		// Some disks possibly corrupted and too many unformatted disks.
```
krisis, line 171: Should we return Abort even if too many disks are unformatted without checking if disksCorrupted is non-zero?

vadmeste: > Should we return Abort even if too many disks are unformatted without checking if disksCorrupted is non-zero?

@krisis, in this case, I think if we have 
`disksUnformatted+disksFormatted+disksOffline` not equal to `diskCount`, this means that we have a corrupted disk.

PR #2988, opened 2016-10-18
Update ssl rpc support

vadmeste:


cmd/net-rpc-client.go
```diff
@@ -83,6 +84,11 @@ func (rpcClient *RPCClient) dialRPCClient() (*rpc.Client, error) {
 		conn, err = net.DialTimeout("tcp", rpcClient.node, 3*time.Second)
 	}
 	if err != nil {
+		// Print RPC connection errors that are worthy to display in log
+		switch err.(type) {
+		case x509.HostnameError:
+			errorIf(err, "Cannot establish RPC")
```
harshavardhana: Change this to following message errorIf(err, "Unable to establish RPC to %s", rpcClient.node)

donatello (approved):

PR #2989, opened 2016-10-18
More informative message when erasure fails to read a part of an object

vadmeste:


cmd/xl-v1-object.go
```diff
@@ -200,6 +200,7 @@ func (xl xlObjects) GetObject(bucket, object string, startOffset int64, length i
 		// Start erasure decoding and writing to the client.
 		n, err := erasureReadFile(mw, onlineDisks, bucket, pathJoin(object, partName), partOffset, readSize, partSize, xlMeta.Erasure.BlockSize, xlMeta.Erasure.DataBlocks, xlMeta.Erasure.ParityBlocks, checkSums, ckSumAlgo, pool)
 		if err != nil {
+			errorIf(err, "Unable to read part %d of the object `%s/%s`.", partIndex, bucket, object)
```
harshavardhana: You can use 'partName' instead of 'partIndex

PR #2995, opened 2016-10-19
Improve Peer RPC error handling (Fixes #2992)

### LINKED ISSUE #2992: [distributed] Need improved error handling of network errors for notifications/policy RPCs
Notifications/Policy peer RPCs do not retry requests in case of broken connections at this point. This can lead to stale configuration on nodes.

This can be assigned to me.

donatello:
- Check for RPC connection shutdown and try again just once.
- Refactor SendRPC to use sync.WaitGroup
- Use simple equality checking instead of err.Error() ==
  rpc.ErrShutdown.Error()

### DIFF
--- a/cmd/notify-listener.go
+++ b/cmd/notify-listener.go
@@ -19,12 +19,13 @@ package cmd
 import (
 	"fmt"
 	"io/ioutil"
+	"net/rpc"
 
 	"github.com/Sirupsen/logrus"
 )
 
 type listenerConn struct {
-	Client      *AuthRPCClient
+	TargetAddr  string
 	ListenerARN string
 }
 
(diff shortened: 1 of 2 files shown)


cmd/net-rpc-client.go
```diff
@@ -137,7 +137,7 @@ func (rpcClient *RPCClient) Call(serviceMethod string, args interface{}, reply i
 	// rpc.Client for a subsequent reconnect.
 	err := rpcLocalStack.Call(serviceMethod, args, reply)
 	if err != nil {
-		if err.Error() == rpc.ErrShutdown.Error() {
+		if err == rpc.ErrShutdown {
```
harshavardhana: Are you sure the err type is not lost?

donatello: Yes, because from golang spec: `Interface values are comparable. Two interface values are equal if they have identical dynamic types and equal dynamic values or if both have value nil.`

harshavardhana: GIven that type is not lost if err is of type fmt.Errorf() and rpc.ErrShutdown is errors.New() they are not comparable even if the function returns with error interface compatible. This is the whole reason we have toStorageErr()

In this example you can see if Call() gets wrapped by something else you would loose that information.  

https://play.golang.org/p/6W8Sj7_5c_ 

Another interesting bug if you rely on this is shown here https://gist.github.com/harshavardhana/c53bb4b2efcdf3d66a2c

donatello: Ok, I have changed it back to comparing with the `.Error()` string value.

PR #2999, opened 2016-10-19
Support to run one minio process per export

krishnasrinivas:


cmd/server-main.go
```diff
@@ -343,18 +324,23 @@ func serverMain(c *cli.Context) {
 	serverAddr := c.String("address")
 
 	// Check if requested port is available.
-	host, port := splitHostPort(serverAddr)
-	fatalIf(checkPortAvailability(port), "Port unavailable %d", port)
+	host, portStr, err := net.SplitHostPort(serverAddr)
```
harshavardhana, line 377: The splitHostPort job was to work when someone specifies minio server like this , perhaps it never did. 

```
minio server --address "10.1.10.1" 
```

Assuming port 80 in case of http and 443 in case of https. You might have to keep spitHostPort and use it to address  this issue.

cmd/globals.go
```diff
@@ -57,6 +57,7 @@ var (
 	globalMinioAddr = ""
 	// Minio default port, can be changed through command line.
 	globalMinioPort = 9000
```
harshavardhana, line 59: is this needed?

cmd/lock-rpc-server.go
```diff
@@ -88,27 +86,25 @@ func newLockServers(serverConfig serverCmdConfig) (lockServers []*lockServer) {
...
 
-	// Save ignored disks in a map
-	// Initialize ignored disks in a new set.
-	ignoredSet := set.NewStringSet()
-	if len(ignoredExports) > 0 {
-		ignoredSet = set.CreateStringSet(ignoredExports...)
-	}
 	for _, export := range exports {
-		if ignoredSet.Contains(export) {
+		ignored := false
```
harshavardhana: how is this better than using set.NewStringSet() ?

cmd/prepare-storage.go
```diff
@@ -250,27 +249,32 @@ func retryFormattingDisks(firstDisk bool, firstEndpoint string, storageDisks []S
...
 		storage, err := newStorageAPI(disks[0])
 		if err != nil && err != errDiskNotFound {
 			return nil, err
 		}
 		return []StorageAPI{storage}, nil
-	} // Otherwise proceed with XL setup.
+	}
+	// Otherwise proceed with XL setup.
+	// FIXME: this check is already done by validateDisks() hence redundant.
 	if err := checkSufficientDisks(disks); err != nil {
```
harshavardhana: We should remove this..

cmd/server-main.go
```diff
@@ -97,12 +98,57 @@ EXAMPLES:
...
+		str = append(str, strconv.Itoa(ep.port))
+	}
+	if ep.path != "" {
+		str = append(str, ep.path)
+	}
+	return strings.Join(str, ":")
+}
+
+func parseStorageEndPoint(ep string, defaultPort int) storageEndPoint {
+	parts := strings.Split(ep, ":")
```
harshavardhana, line 174: 'strings.SplitN' would be better?

cmd/storage-rpc-server.go
```diff
@@ -224,32 +222,27 @@ func newRPCServer(serverConfig serverCmdConfig) (servers []*storageServer, err e
...
 
-	// Initialize ignored disks in a new set.
-	ignoredSet := set.NewStringSet()
-	if len(ignoredExports) > 0 {
-		ignoredSet = set.CreateStringSet(ignoredExports...)
-	}
 	for _, export := range exports {
-		if ignoredSet.Contains(export) {
+		ignored := false
+		for _, ignoredDisk := range ignoredExports {
```
harshavardhana: This can be moved to function, since its getting repeated now.

cmd/utils.go
```diff
@@ -72,6 +72,33 @@ func checkDuplicates(list []string) error {
...
 }
 
+// checkDuplicates - function to validate if there are duplicates in a slice of strings.
+func checkDuplicateEndPoints(list []storageEndPoint) error {
+	// Empty lists are not allowed.
+	if len(list) == 0 {
+		return errInvalidArgument
+	}
+	// Empty end points are not allowed.
+	for _, ep := range list {
```
harshavardhana: why do this? can checkDuplicateStrings be used here as well? since you already have "ep.String()" method.

cmd/storage-rpc-server.go
harshavardhana: set.NewStringSet() can still be used inside a function as 'isIgnoredDisk()'

cmd/utils.go
```diff
@@ -72,7 +72,7 @@ func checkDuplicateStrings(list []string) error {
 	return nil
 }
 
-// checkDuplicates - function to validate if there are duplicates in a slice of strings.
+// checkDuplicates - function to validate if there are duplicates in a slice of endPoints.
```
harshavardhana, line 75: 'checkDuplicateEndpoints'

cmd/server-main.go
krishnasrinivas: actually previously it was getPort(), but i introduced splitHostPort and later removed it and used net.SplitHostPort directly.

Because by default we use port 9000, and if `--address "10.1.10.1"` is used the default port gets changed to 80/443 this would be inconsistent behavior. it will be easy for the users to remember that if port is not specified it defaults 9000 always. right?

cmd/globals.go
krishnasrinivas: yes, it gets used primarily in isLocalStorage() that gets called from various places.

but will see if it can be avoided to be put in global name space, instead put in serverCmdConfig and somehow pass it around. will explore that in a future patch.

cmd/lock-rpc-server.go
krishnasrinivas: we can use the storageEndPoint struct directly for comparison to check if the disk from ignored list is part of the main list.

otherwise if we have to use `set` we'll have to convert storageEndPoint to String and then use it with `set` package and this does not give any advantage.

we can do away with `set` package without compromising on number of lines or simplicity in logic.

cmd/prepare-storage.go
krishnasrinivas: removing it

cmd/utils.go
krishnasrinivas: yes ... let me change it to:

```
var strs []string
for _, ep := range list {
    strs := append(strs, ep.String())
}
return checkDuplicateStrings(strs)
```

cmd/server-main.go
krishnasrinivas: strings.Split() will be better because `127.0.0.1:80:/disk1:/disk1` will result in parse error in the default case of the switch statement here, if SplitN is used it will not catch it.

cmd/lock-rpc-server.go
krishnasrinivas: check the following code for comparison:

```
// Returns if ep is present in the eps list.                                           
func (ep storageEndPoint) presentIn(eps []storageEndPoint) bool {                      
        for _, entry := range eps {                                                    
                if entry == ep {                                                       
                        return true                                                    
                }                                                                      
        }                                                                              
        return false                                                                   
}                                                                                      

// Returns if ep is present in the eps list.                                           
func (ep storageEndPoint) presentIn(eps []storageEndPoint) bool {                      
        disksSet := set.NewStringSet()                                                 
        for _, entry := range eps {                                                    
                disksSet.Add(entry.String())                                           
        }                                                                              
        return disksSet.Contains(ep.String())                                          
}
```

harshavardhana:
> // Returns if ep is present in the eps list.  
> func (ep storageEndPoint) presentIn(eps []storageEndPoint) bool {  
>         for _, entry := range eps {  
>                 if entry == ep {  
>                         return true  
>                 }  
>         }  
>         return false  
> }                                                                                      
> 
> // Returns if ep is present in the eps list.  
> func (ep storageEndPoint) presentIn(eps []storageEndPoint) bool {  
>         disksSet := set.NewStringSet()  
>         for _, entry := range eps {  
>                 disksSet.Add(entry.String())  
>         }  
>         return disksSet.Contains(ep.String())  
> }

You are right, composition with the first one is nicer.

harshavardhana (approved):
Windows needs to be fixed.. .everything else is :+1:

cmd/server-main.go
harshavardhana: Okay makes sense.

harshavardhana:
Fixed and merged locally.. Closing

PR #3030, opened 2016-10-21
Add CMD command to Dockerfile

olalonde:
See https://github.com/minio/minio/pull/2892#discussion_r84454591

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,4 +14,5 @@ RUN \
 
 EXPOSE 9000
 ENTRYPOINT ["go-wrapper", "run"]
+CMD ["server"]
 VOLUME ["/export"]


harshavardhana:
Check the comment regarding the issue @olalonde with this patch.

Dockerfile
```diff
@@ -14,4 +14,5 @@ RUN \
 
 EXPOSE 9000
 ENTRYPOINT ["go-wrapper", "run"]
+CMD ["server"]
```
harshavardhana, line 17: Distributed release brings in new features like 'control, healing' if we default this it complicates the documentation. It was a mistake to keep "server" as part of the entry point previously so we had to fix the bug which led to a breaking change. I apologize for that :( but as it stands today is the correct style of the docker image we wanted. It behaves exactly like our released binaries.

harshavardhana:
Thanks for the patch @olalonde - but we are not going to default any commands for docker container . Docs become inconsistent with regular binary releases 

Closing this, feel free to open other issues if you face with docker container.

olalonde:
@harshavardhana that's just the "Docker way". The default command can just be overridden by `docker run` or `command` in docker-compose.yml if necessary.

Most Dockerfiles are expected to have a default command. For example, nginx: https://github.com/nginxinc/docker-nginx/blob/master/mainline/jessie/Dockerfile#L27, redis: https://github.com/docker-library/redis/blob/master/3.0/Dockerfile#L51, etc.

AFAIK, the docs wouldn't need to be changed at all, e.g. `docker run -p 9000:9000 minio/minio server /export` would still work as expected.

olalonde:
Ah, I missed [that comment](https://github.com/minio/minio/pull/3030#discussion_r84485127). My feeling is that anyone who's seriously using distributed minio in production will have to use a custom command anyways and should have a good understanding of how docker works (no confusion). Anyways, your call!

PR #3032, opened 2016-10-21
Add support of fallocate in FS and XL

vadmeste:
Tests are done under a SATA disk of 3.7 Tb and formatted with XFS.

Speed performance when reading files with and without fallocate (with `drop_caches` and `mc cat`)

| filesize | parts | no fallocate | fallocate |
| --- | --- | --- | --- |
| 350mb | 51 | 3.7s | 2.4s |
| 248mb | 11 | 1.9s | 1.6s |
| 628mb | 2 | 6.1s | 4.2s |


cmd/fs-v1-multipart.go
```diff
@@ -669,6 +678,20 @@ func (fs fsObjects) CompleteMultipartUpload(bucket string, object string, upload
 		// Allocate staging buffer.
 		var buf = make([]byte, readSizeV1)
 
+		objSize := int64(0)
+		for _, part := range parts {
```
harshavardhana: Make this a reusable function. Useful for both xl and fs.

cmd/fs-v1-multipart.go
```diff
@@ -669,6 +678,20 @@ func (fs fsObjects) CompleteMultipartUpload(bucket string, object string, upload
...
+			if partIdx == -1 {
+				return "", traceError(InvalidPart{})
+			}
+			objSize += fsMeta.Parts[partIdx].Size
+		}
+
+		if objSize > 0 {
+			// Prepare file to avoid disk fragmentation
+			_ = fs.storage.PrepareFile(minioMetaBucket, tempObj, objSize)
+		}
```
harshavardhana, line 704: Do not ignore the errors.

harshavardhana: Atleast print it here..

cmd/xl-v1-multipart.go
```diff
@@ -416,6 +416,14 @@ func (xl xlObjects) PutObjectPart(bucket, object, uploadID string, partID int, s
 	// Delete the temporary object part. If PutObjectPart succeeds there would be nothing to delete.
 	defer xl.deleteObject(minioMetaBucket, tmpPartPath)
 
+	if size > 0 {
+		for _, disk := range onlineDisks {
+			if disk != nil {
+				_ = disk.PrepareFile(minioMetaBucket, tmpPartPath, size/2+size%2)
+			}
```
harshavardhana, line 432: Do you need to know exact padding ?

cmd/posix.go
```diff
@@ -541,6 +541,99 @@ func (s *posix) ReadFile(volume string, path string, offset int64, buf []byte) (
...
+			return errFileAccessDenied
+		}
+		return err
+	}
+
+	// Close upon return.
+	defer w.Close()
+
+	if fileSize <= 0 {
+		return nil
```
harshavardhana: This should be an error condition and we should fail PrepareFile if the fileSize is negative.

cmd/fs-v1-multipart.go
vadmeste: > Do not ignore the errors.

yes right, however the xl version of PrepareFile can ignore errors, right?

cmd/xl-v1-multipart.go
vadmeste: > Do you need to know exact padding ?

we can put any number > 0 here, however we will lose the benefit of fallocate if we put a number inferior to  `size/2+size%2`

By the way, I need to see if I can move this to a more low level function which knows in advance how to calculate the size of output files (data + parity)

harshavardhana:
Need some tests here @vadmeste

cmd/posix.go
```diff
@@ -541,6 +541,99 @@ func (s *posix) ReadFile(volume string, path string, offset int64, buf []byte) (
...
+		if isSysErrNotDir(err) {
+			return errFileAccessDenied
+		}
+		return err
+	}
+
+	// Close upon return.
+	defer w.Close()
+
+	if fileSize <= 0 {
```
harshavardhana: Validate this earlier before creating a file and its parents. Otherwise this would unnecessarily leaves stale information.

cmd/posix.go
```diff
@@ -541,6 +541,99 @@ func (s *posix) ReadFile(volume string, path string, offset int64, buf []byte) (
...
+		return errors.New("Unable to prepare a negative sized file.")
+	}
+
+	// Allocate needed disk space to append data
+	e := Fallocate(int(w.Fd()), 0, fileSize)
+
+	// Ignore errors when Fallocate is not supported in the current system
+	if e != nil && e != syscall.ENOSYS && e != syscall.EOPNOTSUPP {
```
harshavardhana: This has to be converted to a function as well into `posix-errors.go`

cmd/posix.go
vadmeste: > Validate this earlier before creating a file and its parents. Otherwise this would unnecessarily leaves stale information.

yes, your are right, my original idea was about "preparing" a file so we do return nil even if fallocate fails because the file was already created at this point. But still like this it is more complicated. I'll fix it.

cmd/posix.go
vadmeste: > This has to be converted to a function as well into posix-errors.go

I don't see any benefit of doing that especially because this is little complicated here, notice that I am not returning an error here when fallocate is not supported. I don't prefer to hide important details inside another function to avoid bugs.

harshavardhana: What i wanted is something similar on the lines of os.IsNotExist() something like `isNotSupported()` .. did you get it?

vadmeste: > What i wanted is something similar on the lines of os.IsNotExist() something like isNotSupported() .. did you get it?

yeah, clear now.

cmd/fs-v1-multipart.go
```diff
@@ -450,6 +450,15 @@ func (fs fsObjects) PutObjectPart(bucket, object, uploadID string, partID int, s
 		bufSize = size
 	}
 	buf := make([]byte, int(bufSize))
+
+	if size > 0 {
+		// Prepare file to avoid disk fragmentation
+		err := fs.storage.PrepareFile(minioMetaBucket, tmpPartPath, size)
```
krishnasrinivas, line 456: this is not needed in putobjectpart as this file doesn't get used after concatenation process

cmd/storage-errors.go
```diff
@@ -59,3 +59,6 @@ var errVolumeAccessDenied = errors.New("volume access denied")
 
 // errVolumeAccessDenied - cannot access file, insufficient permissions.
 var errFileAccessDenied = errors.New("file access denied")
+
+// errFileTooBig - cannot write, file too big
+var errFileTooBig = errors.New("file too big")
```
krishnasrinivas: we need not introduce this as it does not get used at object-layer.

cmd/xl-v1-multipart.go
```diff
@@ -416,6 +416,19 @@ func (xl xlObjects) PutObjectPart(bucket, object, uploadID string, partID int, s
 	// Delete the temporary object part. If PutObjectPart succeeds there would be nothing to delete.
 	defer xl.deleteObject(minioMetaBucket, tmpPartPath)
 
+	if size > 0 {
+		for _, disk := range onlineDisks {
+			if disk != nil {
+				sizeInDisk := size / int64(xlMeta.Erasure.DataBlocks)
```
krishnasrinivas: exact sizeInDisk will be:

```
numBlocks := size/xlMeta.Erasure.BlockSize

chunkSize := getChunkSize(xlMeta.Erasure.BlockSize, dataBlocks)

sizeInDisk := numBlocks * chunkSize

remaining := size%xlMeta.Erasure.BlockSize
if remaining > 0 {
    sizeInDisk += getChunkSize(remaining, dataBlocks)
}
```

But instead of having the above complicated calculation we can take an approximate value of :
`sizeInDisk := size / int64(xlMeta.Erasure.DataBlocks)`

this is good enough for fallocate. You can add a comment saying this is an approx. value not the exact value.

and delete the following as it is not correct:

```
if size%int64(xlMeta.Erasure.DataBlocks) > 0 {
    sizeInDisk++
}
```

cmd/posix.go
```diff
@@ -541,68 +541,139 @@ func (s *posix) ReadFile(volume string, path string, offset int64, buf []byte) (
...
 	}
+
 	// Verify if the file already exists and is not of regular type.
 	var st os.FileInfo
 	if st, err = os.Stat(preparePath(filePath)); err == nil {
 		if !st.Mode().IsRegular() {
-			return errIsNotRegular
+			return nil, errIsNotRegular
 		}
 	}
```
krishnasrinivas: minor optimization, you can put and else {...} statement here and do mkdirAll in the else statement.
this optimization is not related to the current change set.

cmd/xl-v1-object.go
```diff
@@ -448,6 +448,18 @@ func (xl xlObjects) PutObject(bucket string, object string, size int64, data io.
 	// delete.
 	defer xl.deleteObject(minioMetaTmpBucket, tempObj)
 
+	if size > 0 {
+		for _, disk := range onlineDisks {
+			if disk != nil {
+				sizeInDisk := size / int64(xlMeta.Erasure.DataBlocks)
+				if size%int64(xlMeta.Erasure.DataBlocks) > 0 {
+					sizeInDisk++
+				}
```
krishnasrinivas: same as above comment in putobjectpart

krishnasrinivas (changes requested):
@vadmeste did you get the time increase or decrease because of this change? Even if there is no time improvement we have to know that we are taking this change in purely to avoid disk fragmentation. But at the same time there should not be too much of decrease in performance if there is any.

On a side note, PrepareFile does not help in most of the case of multipart upload with FS backend, as appendParts() will not be able to make use of it.

cmd/fs-v1-multipart.go
vadmeste: > this is not needed in putobjectpart as this file doesn't get used after concatenation process

as my general understanding, fallocate + remove is better than no_fallocate + remove because it will still reduce disk fragmentation

cmd/xl-v1-multipart.go
vadmeste: > But instead of having the above complicated calculation we can take an approximate value of :
> sizeInDisk := size / int64(xlMeta.Erasure.DataBlocks)

not sure about this one.. the file still could be fragmented if we allocate with a size inferior to the last final size.

cmd/fs-v1-multipart.go
krishnasrinivas: you are right. fallocate + remove is better than no_fallocate + remove

cmd/xl-v1-multipart.go
krishnasrinivas: Since the FS block size is 4K there is a good chance that it does not get fragmented, depends on the difference between actual and approx size. but sure you can allocate the exact amount.

harshavardhana:
Can you rebase this with current master @vadmeste ?

harshavardhana:
> @vadmeste did you get the time increase or decrease because of this change? Even if there is no time improvement we have to know that we are taking this change in purely to avoid disk fragmentation. But at the same time there should not be too much of decrease in performance if there is any.
> 
> On a side note, PrepareFile does not help in most of the case of multipart upload with FS backend, as appendParts() will not be able to make use of it.

Can you add your results as part of the patch description ?

harshavardhana:
@vadmeste can you rebase and fix conflicts?

vadmeste:
> Can you add your results as part of the patch description ?

:+1:  see the description please

PR #3035, opened 2016-10-21
logging: print file:line:funcName of the caller of errorIf and fatalIf

krishnasrinivas:
previously:

```
krishna@escape:~$ minio server /tmp/export1 /tmp/export1 /tmp/export3 /tmp/export4
FATA[0000] Invalid disk arguments for server.            cause=Duplicate key: "/tmp/export1" found of count: "2"
krishna@escape:~$
```

now:

```
krishna@escape:~$ minio server /tmp/export1 /tmp/export1 /tmp/export3 /tmp/export4
FATA[0000] Invalid disk arguments for server.            cause=Duplicate key: "/tmp/export1" found of count: "2" location=server-main.go:339:validateDisks
krishna@escape:~$
```

### DIFF
--- a/cmd/logger.go
+++ b/cmd/logger.go
@@ -20,6 +20,7 @@ import (
 	"bufio"
 	"bytes"
 	"fmt"
+	"os"
 	"path"
 	"path/filepath"
 	"runtime"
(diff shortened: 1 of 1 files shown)


cmd/logger.go
```diff
@@ -77,13 +78,31 @@ func stackInfo() string {
 	return strings.Replace(stackBuf.String(), filepath.ToSlash(GOPATH)+"/src/", "", -1)
 }
 
+// Get file:line:funcName of the caller.
+func callerLocation() string {
+	pc, file, line, ok := runtime.Caller(2)
+	var name string
+	var location string
+	if ok {
```
harshavardhana: you can use `funcFromPC` inside logger.go

harshavardhana: You can simply make a better function here.. 

```
func callerLocation(stackFrameN int) {
       pc, file, line, success := runtime.Caller(stackFrameN)
        if !success {
                file = "???"
                line = 0
        }
        shortFile := true // We are only interested in short file form.
        callerLoc := funcFromPC(pc, file, line, shortFile)
        return callerLoc
}
```

PR #3051, opened 2016-10-24
rename completeMultipartMD5() into getCompleteMultipartMD5().

balamurugana:


harshavardhana:
Can you add a test for this @balamurugana ?

balamurugana:
> Can you add a test for this @balamurugana ?

Done. Fixed and refreshed.

PR #3055, opened 2016-10-24
Use net.ParseCIDR instead of custom-built parsers

krisis:
## Description

During server initialization we parse endpoint network addresses by parsing their string representation directly. This PR uses standard `net` package function, `ParseCIDR` instead.
- Remove avoidable conversion to and from net.IP to string.
## Motivation and Context

This change makes parsing of network addresses provided as part of server endpoint address more reliable by depending on standard packages.
## How Has This Been Tested?

Existing unit tests cover this change.
## Types of changes

## Checklist:

### DIFF
--- a/cmd/host-to-ip.go
+++ b/cmd/host-to-ip.go
@@ -36,14 +36,18 @@ func (n byLastOctet) Less(i, j int) bool {
 func sortIPsByOctet(ips []string) error {
 	var nips []net.IP
 	for _, ip := range ips {
-		nip := net.ParseIP(ip)
-		if nip == nil {
+		nip, _, err := net.ParseCIDR(ip)
+		if err != nil {
 			return fmt.Errorf("Unable to parse invalid ip %s", ip)
+		}
+		if nip.To4() == nil {
+			continue
 		}
 		nips = append(nips, nip)
 	}
 	// Reverse sort ips by their last octet.
 	sort.Sort(sort.Reverse(byLastOctet(nips)))
+	ips = ips[:len(nips)]
 	for i, nip := range nips {
 		ips[i] = nip.String()
 	}
--- a/cmd/host-to-ip.go
+++ b/cmd/host-to-ip.go
@@ -17,7 +17,6 @@
 package cmd
 
 import (
-	"fmt"
 	"net"
 	"sort"
 )
@@ -33,19 +32,7 @@ func (n byLastOctet) Less(i, j int) bool {
 }
 
 // sortIPsByOctet - returns a reverse sorted list of hosts based on the last octet value.
-func sortIPsByOctet(ips []string) error {
-	var nips []net.IP
-	for _, ip := range ips {
-		nip := net.ParseIP(ip)
-		if nip == nil {
-			return fmt.Errorf("Unable to parse invalid ip %s", ip)
-		}
-		nips = append(nips, nip)
-	}
+func sortIPsByOctet(ips []net.IP) {
 	// Reverse sort ips by their last octet.
-	sort.Sort(sort.Reverse(byLastOctet(nips)))
-	for i, nip := range nips {
-		ips[i] = nip.String()
-	}
-	return nil
+	sort.Sort(sort.Reverse(byLastOctet(ips)))
 }
--- a/cmd/host-to-ip.go
+++ b/cmd/host-to-ip.go
@@ -17,6 +17,7 @@
 package cmd
 
 import (
+	"fmt"
 	"net"
 	"sort"
 )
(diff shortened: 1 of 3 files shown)


cmd/server-main.go
```diff
@@ -223,21 +224,29 @@ func getListenIPs(httpServerConf *http.Server) (hosts []string, port string, err
...
 		for _, addr := range addrs {
 			if addr.Network() == "ip+net" {
-				hostname := strings.Split(addr.String(), "/")[0]
-				if ip := net.ParseIP(hostname); ip.To4() != nil {
-					hosts = append(hosts, hostname)
+				ip, _, err := net.ParseCIDR(addr.String())
+				if err != nil {
+					return nil, port, fmt.Errorf("Unable to parse IP from network interface address. %s", err)
+				}
+				// FIXME: For now, we support only IPv4 addresses. This is mainly due to our sortIPsByOctet.
```
harshavardhana: We shouldnt say this. We don't support ipv6 for lot of other reasons this is misleading.

cmd/host-to-ip.go
```diff
@@ -33,19 +32,7 @@ func (n byLastOctet) Less(i, j int) bool {
...
-			return fmt.Errorf("Unable to parse invalid ip %s", ip)
-		}
-		nips = append(nips, nip)
-	}
+func sortIPsByOctet(ips []net.IP) {
 	// Reverse sort ips by their last octet.
-	sort.Sort(sort.Reverse(byLastOctet(nips)))
-	for i, nip := range nips {
-		ips[i] = nip.String()
-	}
```
harshavardhana, line 49: Now there is no use for this function.  It was kept this way to not use net.IP() instead just take strings and sort it appropriately. You should keep the strings approach back.

cmd/server-main.go
```diff
@@ -223,21 +224,29 @@ func getListenIPs(httpServerConf *http.Server) (hosts []string, port string, err
...
 	if host == "" {
 		var addrs []net.Addr
+		var ips []net.IP
 		addrs, err = net.InterfaceAddrs()
 		if err != nil {
 			return nil, port, fmt.Errorf("Unable to determine network interface address. %s", err)
 		}
 		for _, addr := range addrs {
 			if addr.Network() == "ip+net" {
-				hostname := strings.Split(addr.String(), "/")[0]
```
harshavardhana, line 232: You can move this entire logic to "sortIPsByOctet"

cmd/server-main.go
```diff
@@ -223,21 +224,29 @@ func getListenIPs(httpServerConf *http.Server) (hosts []string, port string, err
...
 		}
-		err = sortIPsByOctet(hosts)
-		if err != nil {
-			return nil, port, fmt.Errorf("Unable reverse sorted ips from hosts %s", err)
+		if len(ips) == 0 {
+			return nil, port, errNoAddr
+		}
+		sortIPsByOctet(ips)
+		for _, ip := range ips {
+			hosts = append(hosts, ip.String())
```
harshavardhana: once this sortIPsByOctet returns strings this is not needed.. You could just return whatever you received.

cmd/server-main.go
krisis: I placed this comment to bring about the discussion. I shall remove this in the next update to this PR.

cmd/host-to-ip.go
krisis: There are two reasons for I reason I sent this PR.
1. We should standard use `net`  package functions like `net.ParseCIDR` to parse network address as opposed to using functions from `strings` package.
2. The existing code, does one too many of string --> net.IP --> string, once for address validation and once for sorting.

cmd/server-main.go
krisis: OK. I shall change the name of the function appropriately.

cmd/server-main.go
krisis: OK

cmd/host-to-ip.go
```diff
@@ -36,14 +36,18 @@ func (n byLastOctet) Less(i, j int) bool {
 func sortIPsByOctet(ips []string) error {
 	var nips []net.IP
 	for _, ip := range ips {
-		nip := net.ParseIP(ip)
-		if nip == nil {
+		nip, _, err := net.ParseCIDR(ip)
+		if err != nil {
 			return fmt.Errorf("Unable to parse invalid ip %s", ip)
 		}
+		if nip.To4() == nil {
```
harshavardhana, line 43: logic should be reverted

```
if nip.To4() != nil {
```

krisis: It needn't be. The check ensures that we skip IP addresses that are not IPv4, which is what is desired.

harshavardhana:
Fixed by #3055, closing.

Issue #3066, opened 2016-10-24
[tests] Need better coverage for `checkPostPolicy`.


hackintoshrao:
Need better test coverage for `checkPostPolicy` from `signature-v4-postpolicyform.go`.

krishnasrinivas:
@harshavardhana can you keep this open till https://github.com/minio/minio/issues/3103 is fixed, will send in more testcases after that is fixed.

harshavardhana:
> @harshavardhana can you keep this open till #3103 is fixed, will send in more testcases after that is fixed.

Only mention the github issue next time in the PR,  if you say "fixes" then it would auto close the github issue.

GarimaKapoor:
@krishnasrinivas please update the timeline for the fix.

krishnasrinivas:
timeline 12-11

vadmeste:
Taking this.

vadmeste:
You can close this @harshavardhana, checkPostPolicy is all green https://codecov.io/gh/minio/minio/src/4daa0d2cee9cf4b0ec7a443416a4c453779eec98/cmd/postpolicyform.go

vadmeste:
Fixed by https://github.com/minio/minio/commit/372da5eaf51a25159a7396d5f114e181135ef91c

harshavardhana:
Closing @vadmeste

Issue #3059, opened 2016-10-24
[tests] Need better test coverage for `PostPolicyBucketHandler` .


hackintoshrao:
Need better test coverage for `PostPolicyBucketHandler` from object-handlers.

krishnasrinivas:
Added few testcases in https://github.com/minio/minio/pull/3104.

More will be added after fixing https://github.com/minio/minio/issues/3103

krishnasrinivas:
timeline 12-11

vadmeste:
I am taking this

PR #3076, opened 2016-10-24
content-length-range policy should be honored for the uploaded object sizes.

krishnasrinivas:
@harshavardhana exhaustive test cases will be added under the issues #3059 #3066

### DIFF
--- a/cmd/bucket-handlers.go
+++ b/cmd/bucket-handlers.go
@@ -17,6 +17,7 @@
 package cmd
 
 import (
+	"encoding/base64"
 	"encoding/xml"
 	"io"
 	"net/http"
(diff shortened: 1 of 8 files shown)


cmd/signature-v4-postpolicyform_test.go
```diff
@@ -33,8 +34,6 @@ func TestPostPolicyForm(t *testing.T) {
 		ErrCode       APIErrorCode
 	}
 	testCases := []testCase{
-		// Mal formed policy document
-		{Bucket: "testbucket", Key: "user/user1/filename/${filename}/myfile.txt", XAmzDate: "20160727T000000Z", XAmzAlgorithm: "AWS4-HMAC-SHA256", ContentType: "image/jpeg", Policy: "==", ErrCode: ErrMalformedPOSTRequest},
 		// Different AMZ date
```
krishnasrinivas, line 37: removed this test case as it was testing base64.StdEncoding.DecodeString

harshavardhana: why shouldn't it be tested?

krishnasrinivas: base64.StdEncoding.DecodeString is part of the standard library, so we need not test it, note that base64.StdEncoding.DecodeString is not used inside checkPostPolicy anymore

harshavardhana: Ah yes i see, now it makes sense.

cmd/bucket-handlers.go
```diff
@@ -450,11 +451,35 @@ func (api objectAPIHandlers) PostPolicyBucketHandler(w http.ResponseWriter, r *h
 		writeErrorResponse(w, r, apiErr, r.URL.Path)
 		return
 	}
-	if apiErr = checkPostPolicy(formValues); apiErr != ErrNone {
+
+	policyBytes, err := base64.StdEncoding.DecodeString(formValues["Policy"])
```
harshavardhana, line 455: why move this out , is it to extract post policy form?

cmd/signature-v4-postpolicyform.go
```diff
@@ -152,19 +149,17 @@ func parsePostPolicyFormV4(policy string) (PostPolicyForm, error) {
...
-	if err != nil {
-		return ErrMalformedPOSTRequest
-	}
+func checkPostPolicy(formValues map[string]string, postPolicyForm PostPolicyForm) APIErrorCode {
 	if !postPolicyForm.Expiration.After(time.Now().UTC()) {
 		return ErrPolicyAlreadyExpired
 	}
+	if postPolicyForm.Conditions.ContentLengthRange.Max > maxObjectSize {
+		// Upload object size can not be more than what is allowed.
+		return ErrNotImplemented
```
harshavardhana: This for API not implemented,  Should be a different error.

harshavardhana: Does s3 return this error?

cmd/bucket-handlers.go
krishnasrinivas: correct, because postpolicyform is needed in the caller function to limit the body reader

cmd/signature-v4-postpolicyform.go
krishnasrinivas: let me check

krishnasrinivas: aws-s3 is not responding with proper message, with (min, max) = (30, 20) and file size of 40, it gives error:
`<Message>Your proposed upload is smaller than the minimum allowed size</Message><ProposedSize>8</ProposedSize>`

what we can do is remove this check let limitReader handle the error case

harshavardhana: okay

krishnasrinivas: aws-s3 is not doing this check, returns error only if the size of the actual uploaded object is > 5G, we can do it the same way.

PR #3083, opened 2016-10-25
Add TLS based tests to functional test suite

### LINKED ISSUE #3020: [distributed] Enable SSL based tests as well for all handler/controller/rpc/api communication
We should use a self-signed certificate and validate it.

krisis:
## Description

Add TLS based functional tests for minio server.
## Motivation and Context

The existing tests only test HTTP API requests. See #3020 
## How Has This Been Tested?

By running, `go test -check.v` and will be run as part of CI on every commit.
## Types of changes

## Checklist:

Fixes #3020


krisis:
The addition of TLS based functional tests has increased the execution time of `make test`. As part of review, do suggest tests that can be removed.

test-data/server.key
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
harshavardhana: You don't need to keep this in a file you can just directly keep this in file as a global []byte array, avoids directories like this.

harshavardhana (commented):
> The addition of TLS based functional tests has increased the execution time of make test. As part of review, do suggest tests that can be removed.

How substantial is the increase?

harshavardhana:
> The addition of TLS based functional tests has increased the execution time of make test. As part of review, do suggest tests that can be removed.
> How substantial is the increase?

`go test -race` doesn't show any substantial increase, btw you don't need to copy all the tests. You can configure the Suite to have TLS and Non - TLS. 

```
// Init and run test on FS backend.
var _ = Suite(&TestSuiteCommon{serverType: "FS"})

// Init and run test on XL backend.
var _ = Suite(&TestSuiteCommon{serverType: "XL"})
```

Like how it is done here..

PR #3092, opened 2016-10-26
Move to URL based syntax formatting.

### LINKED ISSUE #3090: [distributed] Deprecate the current minio distributed syntax and default to URI style approach only.
## Expected Behavior

```
minio server minio server http://192.168.1.11:/media/myvol1 http://192.168.1.11:/media/myvol2 http://192.168.1.12:/media/myvol1 http://192.168.1.12:/media/myvol2 
```

There are not other special syntax's here for distributed setup we are explicitly asking for HTTP or HTTPs. 

This syntax also supports passing accesskey and secretkey through URI. 

## Current Behavior

We support following syntax

```
minio server 192.168.1.11:/media/myvol1 192.168.1.11:/media/myvol2 ... 
minio server --address "192.168.1.11:9000" 192.168.1.11:9000:/media/myvol1 192.168.1.11:9001:/media/myvol1 
```

http or otherwise is inferred from certs configuration.

## Context

Due to inconsistent UI we need to fix this before the release.

harshavardhana:
## Description

Common minio setup. 
- `minio server <dir>` (FS mode) (Special case `minio server <remote>:/dir`)
- `minio server <dir1> .... <dirN>` (XL mode where N is min 4 and max 16)

Distributed XL supports couple of more styles.

(Distributed XL mode where N is min 4 and max 16) - this when SSL is not configured.
- `minio server http://ip:port/dir1 .... http://ip:port/dirN` 

(Distributed XL mode where N is min 4 and max 16) - this when SSL is configured.
- `minio server https://ip:port/dir1 .... https://ip:port/dirN` 

(Distributed XL mode where N is min 4 and max 16) - this when access keys are provided as username and password. 
- `minio server https://username:password@ip:port/dir1 .... https://username:password@ip:port/dirN` 

## Motivation and Context

To fix #3090 

## Types of changes

## Checklist:

Fixes #3090

### DIFF
--- a/cmd/lock-rpc-server.go
+++ b/cmd/lock-rpc-server.go
@@ -81,10 +81,10 @@ func registerDistNSLockRouter(mux *router.Router, serverConfig serverCmdConfig) 
 }
 
 // Create one lock server for every local storage rpc server.
-func newLockServers(serverConfig serverCmdConfig) (lockServers []*lockServer) {
-	for _, ep := range serverConfig.endPoints {
-		if ep.presentIn(serverConfig.ignoredEndPoints) {
-			// Skip initializing ignored end point.
+func newLockServers(srvConfig serverCmdConfig) (lockServers []*lockServer) {
+	for _, ep := range srvConfig.endpoints {
+		if containsEndpoint(srvConfig.ignoredEndpoints, ep) {
+			// Skip initializing ignored endpoint.
 			continue
 		}
 
@@ -92,9 +92,10 @@ func newLockServers(serverConfig serverCmdConfig) (lockServers []*lockServer) {
 		if !isLocalStorage(ep) {
 			continue
 		}
+
 		// Create handler for lock RPCs
 		locker := &lockServer{
-			rpcPath: ep.path,
+			rpcPath: getPath(ep),
 			mutex:   sync.Mutex{},
 			lockMap: make(map[string][]lockRequesterInfo),
 		}
--- a/cmd/object-common.go
+++ b/cmd/object-common.go
@@ -18,6 +18,8 @@ package cmd
 
 import (
 	"net"
+	"net/url"
+	"runtime"
 	"strings"
 	"sync"
 )
@@ -104,21 +106,27 @@ func houseKeeping(storageDisks []StorageAPI) error {
 }
 
 // Check if a network path is local to this node.
-func isLocalStorage(ep storageEndPoint) bool {
-	if ep.host == "" {
+func isLocalStorage(ep *url.URL) bool {
+	if ep.Host == "" {
 		return true
 	}
-	if globalMinioHost != "" {
-		// if --address host:port was specified for distXL we short circuit only the endPoint
-		// that matches host:port
-		if globalMinioHost == ep.host && globalMinioPort == ep.port {
+	if globalMinioHost != "" && globalMinioPort != "" {
+		// if --address host:port was specified for distXL we short
+		// circuit only the endPoint that matches host:port
+		if net.JoinHostPort(globalMinioHost, globalMinioPort) == ep.Host {
 			return true
 		}
 		return false
 	}
+	// Split host to extract host information.
+	host, _, err := net.SplitHostPort(ep.Host)
+	if err != nil {
+		errorIf(err, "Cannot split host port")
+		return false
+	}
 	// Resolve host to address to check if the IP is loopback.
 	// If address resolution fails, assume it's a non-local host.
-	addrs, err := net.LookupHost(ep.host)
+	addrs, err := net.LookupHost(host)
 	if err != nil {
 		errorIf(err, "Failed to lookup host")
 		return false
@@ -149,12 +157,37 @@ func isLocalStorage(ep storageEndPoint) bool {
 	return false
 }
 
-// Depending on the disk type network or local, initialize storage API.
-func newStorageAPI(ep storageEndPoint) (storage StorageAPI, err error) {
-	if isLocalStorage(ep) {
-		return newPosix(ep.path)
+// Fetch the path component from *url.URL*.
+func getPath(ep *url.URL) string {
+	if ep == nil {
+		return ""
 	}
-	return newRPCClient(ep)
+	var diskPath string
+	// For windows ep.Path is usually empty
+	if runtime.GOOS == "windows" {
+		// For full URLs windows drive is part of URL path.
+		// Eg: http://ip:port/C:\mydrive
+		if ep.Scheme == "http" || ep.Scheme == "https" {
+			// For windows trim off the preceding "/".
+			diskPath = ep.Path[1:]
+		} else {
+			// For the rest url splits drive letter into
+			// Scheme contruct the disk path back.
+			diskPath = ep.Scheme + ":" + ep.Opaque
+		}
+	} else {
+		// For other operating systems ep.Path is non empty.
+		diskPath = ep.Path
+	}
+	return diskPath
+}
+
+// Depending on the disk type network or local, initialize storage API.
+func newStorageAPI(ep *url.URL) (storage StorageAPI, err error) {
+	if isLocalStorage(ep) {
+		return newPosix(getPath(ep))
+	}
+	return newStorageRPC(ep)
 }
 
 // Initializes meta volume on all input storage disks.
--- a/cmd/posix-utils_windows_test.go
+++ b/cmd/posix-utils_windows_test.go
@@ -48,7 +48,7 @@ func TestUNCPaths(t *testing.T) {
 	defer os.RemoveAll("c:\\testdisk")
 
 	var fs StorageAPI
-	fs, err = newPosix("c:\\testdisk")
+	fs, err = newPosix(`c:\testdisk`)
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -66,7 +66,6 @@ func TestUNCPaths(t *testing.T) {
 		} else if err == nil && !test.pass {
 			t.Error(err)
 		}
-
 		fs.DeleteFile("voldir", test.objName)
 	}
 }
@@ -81,8 +80,9 @@ func TestUNCPathENOTDIR(t *testing.T) {
 	}
 	// Cleanup on exit of test
 	defer os.RemoveAll("c:\\testdisk")
+
 	var fs StorageAPI
-	fs, err = newPosix("c:\\testdisk")
+	fs, err = newPosix(`c:\testdisk`)
 	if err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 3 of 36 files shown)


cmd/utils_test.go
```diff
@@ -247,25 +255,35 @@ func TestLocalAddress(t *testing.T) {
...
 				serverAddr: fmt.Sprintf(":%d", globalMinioPort),
 				isDistXL:   false,
-				endPoints: []storageEndPoint{
-					{path: "/mnt/disk1"},
-					{path: "/mnt/disk2"},
-					{path: "/mnt/disk3"},
-					{path: "/mnt/disk4"},
-				},
+				endpoints: []*url.URL{{
+					Path: "/mtn/disk1",
```
donatello: Typo: `/mtn` -> `/mnt`

cmd/utils_test.go
```diff
@@ -276,12 +294,15 @@ func TestLocalAddress(t *testing.T) {
...
 				serverAddr: "play.minio.io:9000",
 				isDistXL:   false,
-				endPoints: []storageEndPoint{
-					{path: "/mnt/disk1"},
-					{path: "/mnt/disk2"},
-					{path: "/mnt/disk3"},
-					{path: "/mnt/disk4"},
-				},
+				endpoints: []*url.URL{{
+					Path: "/mtn/disk1",
```
donatello: Same typo.

cmd/lock-rpc-server.go
```diff
@@ -80,11 +81,21 @@ func registerDistNSLockRouter(mux *router.Router, serverConfig serverCmdConfig)
 	return registerStorageLockers(mux, lockServers)
 }
 
+// Contains endpoint returns true if endpoint found in the list of input endpoints.
```
donatello: More readable:

```
// containsEndpoint returns true if `endpoint` is found in input `endpoints`.
```

cmd/posix-utils_windows_test.go
```diff
@@ -81,8 +81,13 @@ func TestUNCPathENOTDIR(t *testing.T) {
 	}
 	// Cleanup on exit of test
 	defer os.RemoveAll("c:\\testdisk")
+
+	endpoints, err := parseStorageEndpoints([]string{`c:\testdik`})
```
donatello: A bit rude maybe? :-)

cmd/server-main_test.go
```diff
@@ -219,9 +219,9 @@ func TestIsDistributedSetup(t *testing.T) {
 			disks  []string
 			result bool
 		}{
-			{[]string{`4.4.4.4:c:\mnt\disk1`, `4.4.4.4:c:\mnt\disk2`}, true},
-			{[]string{`4.4.4.4:c:\mnt\disk1`, `localhost:c:\mnt\disk2`}, true},
-			{[]string{`localhost:c:\mnt\disk1`, `localhost:c:\mnt\disk2`}, false},
+			{[]string{`http://4.4.4.4:80/c:\mnt\disk1`, `http://4.4.4.4:80/c:\mnt\disk2`}, true},
+			{[]string{`http://4.4.4.4:9000/c:\mnt\disk1`, `http://localhost:9000/c:\mnt\disk2`}, true},
+			{[]string{`http://localhost:9000/c:\mnt\disk1`, `http://localhost:9001/c:\mnt\disk2`}, false},
```
donatello: Do we need a test to make sure that http(s) is given in CLI for dist XL => meaning without a protocol part, it errors?

donatello (commented):
LGTM.

cmd/server-main_test.go
harshavardhana: isDistributedSetup would never receive that in the code flow. Since by the time it receives it already had url.URL.

harshavardhana: Frivolous checks on that looks a bit ugly.. But will add more failure cases here tomorrow.

cmd/posix_test.go
```diff
@@ -35,7 +36,13 @@ func newPosixTestSetup() (StorageAPI, string, error) {
...
 		return nil, "", err
 	}
-	posixStorage, err := newPosix(diskPath)
+	var endpoints []*url.URL
+	// Initialize a new posix layer.
+	endpoints, err = parseStorageEndpoints([]string{diskPath})
+	if err != nil {
+		return nil, "", err
+	}
+	posixStorage, err := newPosix(endpoints[0])
```
krishnasrinivas: if you make newPosix() accept a path instead of *url.URL above code can be avoided, also passing url.URL is not adding value to newPosix.

The same pattern can be seen everywhere else, i.e url.URL is constructed just before calling newPosix() and inside newPosix() only Path gets used

cmd/server-main.go
```diff
@@ -101,115 +98,24 @@ EXAMPLES:
...
+func parseStorageEndpoints(eps []string) (endpoints []*url.URL, err error) {
 	for _, ep := range eps {
 		if ep == "" {
-			continue
+			return nil, errInvalidArgument
 		}
-		var endpoint storageEndPoint
-		endpoint, err = parseStorageEndPoint(ep, defaultPort)
+		var u *url.URL
+		u, err = url.Parse(ep)
```
krishnasrinivas, line 120: does url.Parse support parsing http://host:9000/D:/export ? if not how does the syntax look for windows?

cmd/posix_test.go
harshavardhana: the reason to add that is to bring in support for connecting to a remote RPC disk as well. I haven't added that yet.

cmd/server-main.go
harshavardhana: Yes it does you just have to trim the beginning.

cmd/posix_test.go
krishnasrinivas: even for remote rpc disk, the stack will look like:
fsObjects->rpcClient->rpcServer->posix

here rpcServer just needs to do newPosix(path) instead of newPosix(url)

harshavardhana: newFSObjects takes storageDisk not  `*url.URL` .  The logic was on purpose done this way to move the parsing logic outside and storage layer is initialized outside of object layer.  Primarily for backend format formatting etc.

harshavardhana: Actually newStorageAPI already does this, this logic is not needed there.  One particular reason one needs is when printing the local disks we loose the information that provided on command line. 

For example in current master 

```
Initializing data volume.
[01/16] /mnt/export1/minio - 218 GiB online
[02/16] /mnt/export2/minio - 218 GiB online
[03/16] minio2:/mnt/export1/minio - 218 GiB online
[04/16] minio2:/mnt/export2/minio - 218 GiB online
[05/16] minio3:/mnt/export1/minio - 218 GiB online
[06/16] minio3:/mnt/export2/minio - 218 GiB online
[07/16] minio4:/mnt/export1/minio - 218 GiB online
[08/16] minio4:/mnt/export2/minio - 218 GiB online
[09/16] minio5:/mnt/export1/minio - 218 GiB online
[10/16] minio5:/mnt/export2/minio - 218 GiB online
[11/16] minio6:/mnt/export1/minio - 218 GiB online
[12/16] minio6:/mnt/export2/minio - 218 GiB online
[13/16] minio7:/mnt/export1/minio - 218 GiB online
[14/16] minio7:/mnt/export2/minio - 218 GiB online
[15/16] minio8:/mnt/export1/minio - 218 GiB online
[16/16] minio8:/mnt/export2/minio - 218 GiB online
```

It is neat to know what was the suppliedOriginal path.

harshavardhana: Thinking... how it can be done.

cmd/lock-rpc-server.go
```diff
@@ -80,11 +81,21 @@ func registerDistNSLockRouter(mux *router.Router, serverConfig serverCmdConfig)
...
+func containsEndpoint(endpoints []*url.URL, endpoint *url.URL) bool {
+	for _, ep := range endpoints {
+		if ep.Host == endpoint.Host && ep.Scheme == endpoint.Scheme && ep.Path == endpoint.Path {
```
krishnasrinivas: `if *ep == *endpoint {` would be better?

cmd/object-common.go
```diff
@@ -104,21 +105,19 @@ func houseKeeping(storageDisks []StorageAPI) error {
...
 		return true
 	}
-	if globalMinioHost != "" {
-		// if --address host:port was specified for distXL we short circuit only the endPoint
-		// that matches host:port
-		if globalMinioHost == ep.host && globalMinioPort == ep.port {
-			return true
-		}
+	// Split host to extract host information.
+	host, _, err := net.SplitHostPort(ep.Host)
```
krishnasrinivas, line 122: support for process per disk is lost here?

harshavardhana: yes on purpose, i broke it. Will bring in back separately.

cmd/lock-rpc-server.go
harshavardhana: :+1:

cmd/storage-rpc-client.go
```diff
@@ -93,15 +93,14 @@ func toStorageErr(err error) error {
 }
 
 // Initialize new rpc client.
-func newRPCClient(ep storageEndPoint) (StorageAPI, error) {
-	// Input validation.
-	if ep.host == "" || ep.port == 0 || ep.path == "" {
+func newPosixRPC(ep *url.URL) (StorageAPI, error) {
```
krishnasrinivas: calling it `newRPCClient` is more accurate because the function returns a structure related to rpcClient

also hypothetically if the remote storage can be a DB, calling this newRPCClient makes more sense

harshavardhana: newRPCClient is too generic its closer to 'net/rpc' .Also here RPC client is a StorageAPI. 

newRemoteStorage() makes sense.

cmd/posix_test.go
krishnasrinivas: it will be available in serverCmdConfig{}, there was a globalvariable holding this right?

cmd/storage-rpc-client.go
krishnasrinivas: but the convention is newSomething() should return `Something` or `*Something`

krishnasrinivas: also unrelated to this patch:

```
type networkStorage struct {                                                                                                                                                    
        netAddr   string                                                                                                                                                        
        netPath   string                                                                                                                                                        
        rpcClient *AuthRPCClient                                                                                                                                                
}
```

this and AuthRPCClient can be combined into one struct and call it maybe rpcClient. Because networkStorage is not doing anything extra other than making calls to AuthRPCClient

cmd/posix_test.go
harshavardhana: Yes but globalVariable association is cryptic and while the rest using disk.String().

cmd/posix.go
```diff
@@ -98,13 +99,30 @@ func isDirEmpty(dirname string) bool {
...
 	}
-	suppliedDiskPath := diskPath
-	var err error
+	suppliedDiskPath := ep.String()
+	var diskPath string
+	// For windows ep.Path is usually empty
+	if runtime.GOOS == "windows" {
+		// For full URLs windows drive is part of URL path.
+		// Eg: http://ip:port/C:\mydrive
+		if ep.Scheme == "http" || ep.Scheme == "https" {
```
krishnasrinivas: at the posix layer, dealing with URL related fields does not look right.
If http/hostname is needed for printing, printing should be done differently (outside posix). Because why should posix's String() print http/ipaddress related info?

generally if a function takes arguments that only it is concerned with it will result in a simpler code with right abstraction.

harshavardhana: You are right, i need to fix this behavior. The overarching problem is loosing command line arguments.

harshavardhana:
Patch is updated, start reviewing - will be adding tests subsequently.

cmd/object-common.go
```diff
@@ -104,21 +106,25 @@ func houseKeeping(storageDisks []StorageAPI) error {
...
-	if ep.host == "" {
+func isLocalStorage(ep *url.URL) bool {
+	if ep.Host == "" {
 		return true
 	}
-	if globalMinioHost != "" {
-		// if --address host:port was specified for distXL we short circuit only the endPoint
-		// that matches host:port
-		if globalMinioHost == ep.host && globalMinioPort == ep.port {
+	if globalMinioHost != "" && globalMinioPort != "" {
```
krishnasrinivas, line 113: you have deleted the comment explaining the short circuit case

cmd/posix.go
```diff
@@ -100,22 +99,21 @@ func isDirEmpty(dirname string) bool {
...
+func newPosix(path string) (StorageAPI, error) {
+	if path == "" {
 		return nil, errInvalidArgument
 	}
-	suppliedDiskPath := diskPath
-	var err error
+	diskPath := path
 	// Disallow relative paths, figure out absolute paths.
+	var err error
 	diskPath, err = filepath.Abs(diskPath)
```
krishnasrinivas: you can do `diskPath, err := filepath.Abs(diskPath)` and delete 2 lines above it

cmd/server-main.go
```diff
@@ -91,125 +88,35 @@ EXAMPLES:
 	  /mnt/export8/ /mnt/export9/ /mnt/export10/ /mnt/export11/ /mnt/export12/
 
   6. Start minio server on a 4 node distributed setup. Type the following command on all the 4 nodes.
-      $ export MINIO_ACCESS_KEY=minio
-      $ export MINIO_SECRET_KEY=miniostorage
-      $ minio {{.Name}} 192.168.1.11:/mnt/export/ 192.168.1.12:/mnt/export/ \
-          192.168.1.13:/mnt/export/ 192.168.1.14:/mnt/export/
+      $ minio {{.Name}} http://minio:miniostorage@192.168.1.11/mnt/export/ http://minio:miniostorage@192.168.1.12/mnt/export/ \
+          http://minio:miniostorage@192.168.1.13/mnt/export/ http://minio:miniostorage@192.168.1.14/mnt/export/
```
krishnasrinivas: instead of `minio:miniostorage` it can say 'accessKey:secretKey`

harshavardhana: This example is usually a working example where they can copy paste it.  Like how we did with `mc config host`. It is understood that these should be replaced with whatever they want.

krishnasrinivas: since we support MINIO_ACCESS_KEY/MINIO_SECRET_KEY, that will be a simpler syntax to follow than this one as the command looks very long and also it is not obvious that `minio` is the accessKey and `miniostorage` is the secretKey here

previous example had `export MINIO_ACCESS_KEY=minio` `export MINIO_SECRET_KEY=miniostorage` which made sure that user knows that minio is the accessKey and miniostorage is the secretKey

better provide this new syntax in documentation or as an example number 7

Issue #3073, opened 2016-10-24
[distributed] Add generation id to uploads.json


donatello:
## Expected Behavior

The uploads.json file used for multipart uploads needs to have a generation id, that helps identify the latest version of the file that was written "quorum" successfully.
## Current Behavior

Currently we are using the version from the first disk on which we successfully read the file. This can lead to stale information.
## Possible Solution

Add generation id to uploads.json.
## Context

Stale information in uploads.json is a source of bugs. This was discovered while reading source code in relation to chasing down bug #3014

harshavardhana:
This is not valid anymore closing. @donatello open a relevant bug which doesn't talk about generation id.

PR #3093, opened 2016-10-26
[WIP] Add generation id to uploads.json (#3073)

donatello:
## Description

Adding generation id to the uploads.json makes sure that we do not
read a stale version of the file.
- Since we always read before we write uploads.json, we can use a 64-bit
  counter as the generation id.
- Reading uploads.json now requires a read quorum to agree on the
  generation id in XL.
- The generation id is incremented whenever an uploadID is added to or
  removed from uploads.json.
- Fix a bug in FS where a read uploads.json error was ignored.
- Minor refactoring to make the code easier to follow.
## Motivation and Context

https://github.com/minio/minio/issues/3073
## Types of changes

## Checklist:

### DIFF
--- a/cmd/object-multipart-common.go
+++ b/cmd/object-multipart-common.go
@@ -35,6 +35,7 @@ type uploadInfo struct {
 type uploadsV1 struct {
 	Version string       `json:"version"`   // Version of the current `uploads.json`
 	Format  string       `json:"format"`    // Format of the current `uploads.json`
+	GenID   int64        `json:"GenId"`     // A generation id to identify latest written file.
 	Uploads []uploadInfo `json:"uploadIds"` // Captures all the upload ids for a given object.
 }
 
@@ -47,13 +48,15 @@ func (t byInitiatedTime) Less(i, j int) bool {
 	return t[i].Initiated.Before(t[j].Initiated)
 }
 
-// AddUploadID - adds a new upload id in order of its initiated time.
+// AddUploadID - adds a new upload id in order of its initiated
+// time. Also increments the Generation ID.
 func (u *uploadsV1) AddUploadID(uploadID string, initiated time.Time) {
 	u.Uploads = append(u.Uploads, uploadInfo{
 		UploadID:  uploadID,
 		Initiated: initiated,
 	})
 	sort.Sort(byInitiatedTime(u.Uploads))
+	u.GenID++
 }
 
 // Index - returns the index of matching the upload id.
@@ -64,6 +67,16 @@ func (u uploadsV1) Index(uploadID string) int {
 		}
 	}
 	return -1
+}
+
+// RemoveUploadIDAtIndex - removes upload ID at given index. Also
+// increments Generation ID.
+func (u uploadsV1) RemoveUploadIDAtIndex(idx int) {
+	// we ignore invalid index
+	if len(u.Uploads) > idx {
+		u.Uploads = append(u.Uploads[:idx], u.Uploads[idx+1:]...)
+		u.GenID++
+	}
 }
 
 // readUploadsJSON - get all the saved uploads JSON.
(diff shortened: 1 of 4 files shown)


cmd/object-multipart-common.go
```diff
@@ -35,6 +35,7 @@ type uploadInfo struct {
 type uploadsV1 struct {
 	Version string       `json:"version"`   // Version of the current `uploads.json`
 	Format  string       `json:"format"`    // Format of the current `uploads.json`
+	GenID   int64        `json:"GenId"`     // A generation id to identify latest written file.
```
harshavardhana, line 38: Don't use GenID stick to the same style as "xl.json" .

cmd/object-multipart-common.go
```diff
@@ -66,6 +69,16 @@ func (u uploadsV1) Index(uploadID string) int {
 	return -1
 }
 
+// RemoveUploadIDAtIndex - removes upload ID at given index. Also
+// increments Generation ID.
+func (u uploadsV1) RemoveUploadIDAtIndex(idx int) {
+	// we ignore invalid index
+	if len(u.Uploads) > idx {
+		u.Uploads = append(u.Uploads[:idx], u.Uploads[idx+1:]...)
+		u.GenID++
```
harshavardhana, line 78: We used this approach before for `xl.json` - time is more meaningful.

cmd/xl-v1-multipart-common.go
```diff
@@ -87,22 +87,55 @@ func (xl xlObjects) updateUploadsJSON(bucket, object string, uploadsJSON uploads
...
+	genIDCount := make(map[int64]int)
+	for i := 0; i < len(uploadResp); i++ {
+		if errs[i] != nil {
 			continue
 		}
-		break
+		// update the seen generation ID counter
+		gID := uploadResp[i].GenID
+		genIDCount[gID]++
+		// is the current gen-id "quorum" seen?
```
harshavardhana, line 131: Look at how `xl.json` is read in this scenario we need to implement something in similar lines of `readAllXlMetadata()`

donatello:
@krishnasrinivas Yes checked with harsha. The change is needed because the existing behaviour of reading the first available upload.json can lead to stale information (a previous write failure to a disk that overall succeeded in quorum, could be the version that is being read). A notion of quorum is required in reading uploads.json. Adding the generation id is a way to do that.

The backend format change is not serious here. We can without loss of (any extra) correctness, assume that existing uploads.json (written by a version of minio without this change) have a generation id of 0 and continue from there. Of course the patch is not yet complete as tests don't yet pass.

Regarding using time as a generation id: Suppose `t` is the timestamp in the existing uploads.json. It is possible (though somewhat unlikely) that a server may select and use the same timestamp `t` when updating uploads.json - this can happen due to time-skew in different servers. But this is a problem because if only some of the uploads.json files are updated, it is not possible to tell which is the more recent version. In the case of uploads.json we are able to circumvent this using an incrementing counter as a time independent logical clock because we have the property that uploads.json is always read before being written/updated - thus we always have the previous value of the counter and are able to increment it with no extra cost.

harshavardhana (commented):
> Regarding using time as a generation id: Suppose t is the timestamp in the existing uploads.json. It is possible (though somewhat unlikely) that a server may select and use the same timestamp t when updating uploads.json - this can happen due to time-skew in different servers. But this is a problem because if only some of the uploads.json files are updated, it is not possible to tell which is the more recent version. In the case of uploads.json we are able to circumvent this using an incrementing counter as a time independent logical clock because we have the property that uploads.json is always read before being written/updated - thus we always have the previous value of the counter and are able to increment it with no extra cost.

This i already explained that GenID is not useful. Modtime is the right value even if they are skewed we are in quorum whatever is in quorum is always the right copy. 

We don't need to invent monotonic clock style behavior here.

harshavardhana:
> The backend format change is not serious here. We can without loss of (any extra) correctness, assume that existing uploads.json (written by a version of minio without this change) have a generation id of 0 and continue from there. Of course the patch is not yet complete as tests don't yet pass.

This we need to confirm first with AB as well.  This patch was to open up the discussion.

donatello:
After discussing with AB, this is going to be tackled with a new algorithm. I will open a new PR as it is significantly different from this approach.

Issue #3071, opened 2016-10-24
erasure: Consider lowering minimum number for XL/Distributed to 2


fwessels:
Now that we have relaxed the read quorum to N/2, we could consider lowering the minimum number for XL/Distributed from 4 to 2.

Effectively this would mean replication over 2 disks/nodes while still having read capability when either one of the disks or either one of the nodes is down (no master node).

Obviously write capability would be hampered when one of the disks/nodes is down (unless we rely on healing which would not be more than a mere copy).

Could be a "poor man's" alternative to a replicated setup of two FS servers with one master and one slave (with the added benefit that a 2 server distributed cluster would not have a designated master but a single round robin load balancer would end up at the server that is still up).

harshavardhana:
> Now that we have relaxed the read quorum to N/2, we could consider lowering the minimum number for XL/Distributed from 4 to 2.
> 
> Effectively this would mean replication over 2 disks/nodes while still having read capability when either one of the disks or either one of the nodes is down (no master node).
> 
> Obviously write capability would be hampered when one of the disks/nodes is down (unless we rely on healing which would not be more than a mere copy).
> 
> Could be a "poor man's" alternative to a replicated setup of two FS servers with one master and one slave (with the added benefit that a 2 server distributed cluster would not have a designated master but a single round robin load balancer would end up at the server that is still up).

This will be more unreliable to do @fwessels , it is better to setup two Minio FS servers and do 'mc mirror'. 

Write requires N/2+1 quorum to provide sanctity during network partitions it makes it harder to make it bug free and we will loose data.  I would prefer we shouldn't support an unreliable setup.

fwessels:
@harshavardhana Agreed, naturally will be less reliable compared to a 4 disk/node sytem. And of course provides less guarantees against data protection (but not any less than two FS servers syncing between them).

harshavardhana:
> @harshavardhana Agreed, naturally will be less reliable compared to a 4 disk/node sytem. And of course provides less guarantees against data protection (but not any less than two FS servers syncing between them).

@fwessels any specific reason for this or just to have a "poor mans" replication?

fwessels:
No specific reason, just maybe a simpler alternative to FS replication. But maybe the downside is worse than the upside...

GarimaKapoor:
cc'ing @abperiasamy

harshavardhana:
Since we decided to not do it for now, closing this issue for now. We can re-open again as needed.

[...]

PR #3101, opened 2016-10-26
Lower quorum minimum to 2

fwessels:
With both read quorum for erasure coding determined at N/2 and read lock (in dsync) adjusted to N/2, we can lower the minimum number of nodes for Distributed from 4 to 2.

Fixes https://github.com/minio/minio/issues/3071


harshavardhana:
Closing this PR since this change is pushed to future, we can bring this patch back in if and when we decide to do this.

PR #3107, opened 2016-10-26
Show offline nodes after a fixed number of init retry

vadmeste:
## Description

After 5 format retries, show which nodes are offline
## Motivation and Context

The user needs a better information about what's wrong when nodes fail to init after some moments
## How Has This Been Tested?

Manually
## Types of changes
## Checklist:

### DIFF
--- a/cmd/prepare-storage.go
+++ b/cmd/prepare-storage.go
@@ -196,9 +196,16 @@ func retryFormattingDisks(firstDisk bool, firstEndpoint string, storageDisks []S
 	defer close(doneCh)
 
 	// Wait on the jitter retry loop.
-	for range newRetryTimer(time.Second, time.Second*30, MaxJitter, doneCh) {
+	for retryCounter := range newRetryTimer(time.Second, time.Second*30, MaxJitter, doneCh) {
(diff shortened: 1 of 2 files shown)


cmd/prepare-storage.go
```diff
@@ -196,9 +196,16 @@ func retryFormattingDisks(firstDisk bool, firstEndpoint string, storageDisks []S
 	defer close(doneCh)
 
 	// Wait on the jitter retry loop.
-	for range newRetryTimer(time.Second, time.Second*30, MaxJitter, doneCh) {
+	for retryCounter := range newRetryTimer(time.Second, time.Second*30, MaxJitter, doneCh) {
 		// Attempt to load all `format.json`.
 		formatConfigs, sErrs := loadAllFormats(storageDisks)
+		if retryCounter > 5 {
+			for i, e := range sErrs {
```
harshavardhana, line 203: What about the other errors? - we can just print the errors directly. @vadmeste

vadmeste: > What about the other errors?

Here, we delay showing some errors that could disappear later like disconnected node in our case. Other errors could be shown in other PRs

PR #3114, opened 2016-10-27
boot: check parameter syntax before initializing the system.

### LINKED ISSUE #3112: [distributed] Check argument syntax before doing anything else
If an unrecognized argument syntax is presented we should print error, help and exit.
## Expected Behavior

For ex.

```
minio server server1:/export/disk1 server2:/export/disk2 server3:/export/disk3 server4:/export/disk4
```

We should error out for this.

## Current Behavior

Right now we start one minio process with all disks as local disks.

krishnasrinivas:
## Description

`minio server <parameters>`

`<parameters>` syntax should be checked before doing anything else.
## Motivation and Context

Details here https://github.com/minio/minio/issues/3112
## How Has This Been Tested?

manual testing for minio-fs, singlenode-xl, distributed-xl
## Types of changes

## Checklist:

  fixes #3112

### DIFF
--- a/cmd/server-main.go
+++ b/cmd/server-main.go
@@ -26,6 +26,9 @@ import (
 	"strings"
 	"time"
 
+	"regexp"
+	"runtime"
+
 	"github.com/minio/cli"
 )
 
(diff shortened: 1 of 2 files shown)


cmd/server-main.go
```diff
@@ -276,6 +259,99 @@ func isDistributedSetup(eps []*url.URL) (isDist bool) {
 	return isDist
 }
 
+// We just exit for invalid endpoints.
+func checkEndpointsSyntax(eps []*url.URL, disks []string) {
+	for i, u := range eps {
+		// Scheme is "" for FS and singlenode-XL
+		if u.Scheme != "" && u.Scheme != "http" && u.Scheme != "https" {
```
harshavardhana: This not correct on windows when you specify C:\disk windows keeps "C" as part of Scheme.  You can possibly validate this on windows in this manner.  You have to handle this just like how getPath() handles the right path from looking at the list of URLs on windows.

harshavardhana:
The windows changes are good but you need to check why its failing again.  It is still one of those `fatalf`

harshavardhana:
Can you add some test cases here @krishnasrinivas ?

krishnasrinivas:
> Can you add some test cases here @krishnasrinivas ?

yes

krishnasrinivas:
don't merge this yet, some parsing problem in windows

cmd/server-main.go
```diff
@@ -119,8 +119,7 @@ func parseStorageEndpoints(eps []string) (endpoints []*url.URL, err error) {
 		if ep == "" {
 			return nil, errInvalidArgument
 		}
-		var u *url.URL
-		u, err = url.Parse(ep)
+		u, err := url.Parse(ep)
```
harshavardhana: This causes shadowing with "err error" return value. We need to keep it the same way as before or not return "err error" .

krishnasrinivas: ok

krishnasrinivas:
> don't merge this yet, some parsing problem in windows

not related to this patch, will send separate fix later

PR #3124, opened 2016-10-28
Read/write quorum algo with uploads.json in xl (Fixes #3123)

### LINKED ISSUE #3123: [distributed] Fix reading uploads.json behavior in multipart uploads
## Expected Behavior

Reading and updating uploads.json needs a quorum notion to avoid reading stale data.
## Current Behavior

Currently we are using the version from the first disk on which we successfully read the file. This can lead to stale information. (same as from #3073 )
## Context

Working on a new approach without backend format changes and better than the approach in #3093

donatello:
## Description

- Reads and writes of uploads.json in XL now uses quorum for
  newMultipart, completeMultipart and abortMultipart operations.
- Each disk's `uploads.json` file is read and updated independently for
  adding or removing an upload id from the file. Quorum is used to
  decide if the high-level operation actually succeeded.
- Refactor FS code to simplify the flow, and fix a bug while reading
  uploads.json.
## Motivation and Context

Fixes #3123 
## How Has This Been Tested?

All existing unit-tests pass.

Manual testing is underway.
## Types of changes

## Checklist:

### DIFF
--- a/cmd/fs-v1-multipart-common.go
+++ b/cmd/fs-v1-multipart-common.go
@@ -16,11 +16,7 @@
 
 package cmd
 
-import (
-	"encoding/json"
-	"path"
-	"time"
-)
+import "path"
 
 // Returns if the prefix is a multipart upload.
 func (fs fsObjects) isMultipartUpload(bucket, prefix string) bool {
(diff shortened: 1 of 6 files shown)


cmd/fs-v1-multipart-common.go
```diff
@@ -57,56 +56,56 @@ func (fs fsObjects) isUploadIDExists(bucket, object, uploadID string) bool {
...
 
-	// Update `uploads.json` on all disks.
-	uploadsJSONBytes, wErr := json.Marshal(&uploadsJSON)
-	if wErr != nil {
-		return traceError(wErr)
+	uploadsJSON, err := readUploadsJSON(bucket, object, fs.storage)
+	if errorCause(err) == errFileNotFound {
+		// If file is not found, we assume a default (empty)
+		// upload info.
+		uploadsJSON, err = newUploadsV1("xl"), nil
```
harshavardhana: newUploadsV1("xl") is not correct for FS.

cmd/xl-v1-multipart-common.go
```diff
@@ -20,169 +20,147 @@ import (
...
+					return
+				}
+				wErr = disk.RenameFile(minioMetaBucket, tmpUploadsPath, minioMetaBucket, uploadsPath)
+				if wErr != nil {
+					if dErr := disk.DeleteFile(minioMetaBucket, tmpUploadsPath); dErr != nil {
+						// we store the most recent error.
+						errs[index] = traceError(dErr)
+						return
+					}
+					errs[index] = traceError(wErr)
```
harshavardhana: This whole code here can be moved to a single function. Its common across FS and XL.

cmd/xl-v1-multipart-common.go
```diff
@@ -20,169 +20,147 @@ import (
...
 			}
 			// Undo rename `uploads.json` in parallel.
 			wg.Add(1)
 			go func(index int, disk StorageAPI) {
 				defer wg.Done()
-				if errs[index] != nil {
-					return
+
+				if !isDelete[index] {
+					// we don't want a partially
```
harshavardhana: Don't manually CR comments as i said before there is no 80 line requirement in Golang. This kind of looks ugly. This comment can be made better simply by saying 

```
// Remove partially written files.
```

cmd/xl-v1-multipart-common.go
```diff
@@ -20,169 +20,147 @@ import (
...
+	// Do we have write quorum?.
 	if !isDiskQuorum(errs, xl.writeQuorum) {
-		// Do we have readQuorum?.
-		if isDiskQuorum(errs, xl.readQuorum) {
-			return nil
-		}
-		// Rename `uploads.json` left over back to tmp location.
-		for index, disk := range xl.storageDisks {
-			if disk == nil {
+		// No quorum. Perform cleanup on the minority of disks
```
harshavardhana, line 92:
```
// No quorum. Perform cleanup on the minority of disks on which the operation succeeded. 
```

cmd/fs-v1-multipart-common.go
donatello: Fixed

cmd/xl-v1-multipart-common.go
donatello: Done.

cmd/xl-v1-multipart-common.go
donatello: Fixed.

cmd/xl-v1-multipart-common.go
donatello: Improved this by putting the comments a little earlier, and so less indented.

harshavardhana:
ping @donatello - let me know when this is ready to be merged.

donatello:
@harshavardhana This patch is ready for merge.

krishnasrinivas:
FS related files are getting affected for quorum fixes, is it because of uploads.json are same for both FS and XL?

In that case, FS uploads.json structure should be different from XL's uploads.json just like we have fsMetaV1 different from xlMetaV1

krishnasrinivas:
> FS related files are getting affected for quorum fixes, is it because of uploads.json are same for both FS and XL?
> 
> In that case, FS uploads.json structure should be different from XL's uploads.json just like we have fsMetaV1 different from xlMetaV1

I see that there is shared code between FS and XL when updating uploads.json if it's convenient to share the code then sharing uploads.json makes sense

PR #3132, opened 2016-10-31
posix: Fix windows performance issues.

harshavardhana:
Do not attempt to fetch volume/drive information for
each i/o situation. In our case we do this in all calls
`posix.go` this in-turn created a terrible situation for
windows. This issue does not affect the i/o path on Unix
platforms since statvfs calls are in the range of micro
seconds on these platforms.

This verification is only needed during startup and we
let things fail at a later stage on windows.

## Description

Do not attempt to fetch volume/drive information for each i/o situation on windows.

## Motivation and Context

Fixes lingering windows performance issues.

## How Has This Been Tested?

Compile and run on windows. 

## Types of changes

## Checklist:

### DIFF
--- a/cmd/posix.go
+++ b/cmd/posix.go
@@ -34,9 +34,9 @@ import (
 )
 
 const (
-	fsMinFreeSpace         = 1024 * 1024 * 1024
-	fsMinFreeInodesPercent = 5
-	maxAllowedIOError      = 5
+	fsMinFreeSpace    = 1024 * 1024 * 1024 // Min 1GiB free space.
+	fsMinFreeInodes   = 10000              // Min 10000.
+	maxAllowedIOError = 5
 )
 
 // posix - implements StorageAPI interface.
@@ -56,6 +56,9 @@ func checkPathLength(pathName string) error {
 	if runtime.GOOS == "darwin" && len(pathName) > 1016 {
 		return errFileNameTooLong
 	}
+
+	// Convert any '\' to '/'.
+	pathName = filepath.ToSlash(pathName)
 
 	// Check each path segment length is > 255
 	for len(pathName) > 0 && pathName != "." && pathName != "/" {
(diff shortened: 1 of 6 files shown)


cmd/posix.go
```diff
@@ -34,9 +34,9 @@ import (
 )
 
 const (
-	fsMinFreeSpace         = 1024 * 1024 * 1024
-	fsMinFreeInodesPercent = 5
-	maxAllowedIOError      = 5
+	fsMinFreeSpace    = 1024 * 1024 * 1024 // Min 1GiB free space.
+	fsMinFreeInodes   = 10000 + 2          // Min 10000 (max multiparts) + 2 (uploads.json, xl.json or fs.json) free inodes.
```
krishnasrinivas: you could keep it as `fsMinFreeInodes   = 10000` removing `+ 2` the extra `2` isn't adding much value

cmd/posix.go
```diff
@@ -166,7 +180,7 @@ func (s *posix) checkDiskFree() (err error) {
 	// Allow for the available disk to be separately validate and we will validate inodes only if
 	// total inodes are provided by the underlying filesystem.
 	if di.Files != 0 {
-		availableFiles := 100 * float64(di.Ffree) / float64(di.Files)
+		availableFiles := float64(di.Ffree) * 0.95
```
krishnasrinivas: shouldnt it be `availableFiles := di.Ffree`

cmd/posix.go
harshavardhana: Just adding to encompass the whole transaction. 10000 is also fine.

cmd/posix.go
harshavardhana: oh yes you are right.

krishnasrinivas: i see the comment : `// Remove 5% from free space for cumulative disk space used for journalling, inodes etc.`

i think you can increase `fsMinFreeInodes   = 10000 + 2` or ignore the 5% for journaling and inodes to keep the code simple

krishnasrinivas: > oh yes you are right.

ok the 5% is `10000` inodes now

harshavardhana:
ping @krishnasrinivas

krishnasrinivas:
With this fix, getting upload speeds of 60MBps on localhost (with fdcache it's 130MBps) 60MBps speed is good considering that network will be throttled by two uploads on a 1gige.

krishnasrinivas (approved):

PR #3135, opened 2016-10-31
docker: Move the docker document and compose to a separate dir.

harshavardhana:
## Types of changes

## Checklist:

### DIFF
--- a/docs/docker/README.md
+++ b/docs/docker/README.md
@@ -37,13 +37,14 @@ docker run -p 9000:9000 --name minio1 \
 
 ```
 
-## 4. Run Distributed Minio on Docker
+## 4. Test Distributed Minio on Docker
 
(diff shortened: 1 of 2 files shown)


docs/docker/README.md
```diff
@@ -39,7 +39,7 @@ docker run -p 9000:9000 --name minio1 \
 
 ## 4. Run Distributed Minio on Docker
```
abperiasamy: Test Distributed..

abperiasamy (commented):
Compose is only meant for testing (simulating) a distributed Minio setup on a single node. It is not meant for real workload. In that case, it is OK to simply use internal ~/.minio and /export

neomantra:
The `docker-compose.yml` link on [this page](https://docs.minio.io/docs/docker-compose.yml) is currently broken.  Maybe it needs to be regenerated due to this commit?

harshavardhana:
> The docker-compose.yml link on this page is currently broken. Maybe it needs to be regenerated due to this commit?

No looks like the path is wrong. Will fix it thanks for reporting @neomantra it is accessible here - https://github.com/minio/minio/blob/master/docs/docker/docker-compose.yml

PR #3139, opened 2016-10-31
fs: Remove object append file if found since it is not always used

### LINKED ISSUE #3137: fs: uploading the same object in a loop will fill the disk
some files are not cleaned under tmp/

```while true; do mc cp 70mb myminio/testbucket/ ; done```

### PR DESCRIPTION
**Pull request by vadmeste**
## Description
Disk gets filled in FS because append file in tmp/ is not removed when not used

## Motivation and Context
Fixes #3137 

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:

### DIFF
--- a/cmd/fs-v1-multipart.go
+++ b/cmd/fs-v1-multipart.go
@@ -669,6 +669,8 @@ func (fs fsObjects) CompleteMultipartUpload(bucket string, object string, upload
 		return "", err
 	}
 
+	fsAppendDataPath := getFSAppendDataPath(uploadID)
+
 	// Read saved fs metadata for ongoing multipart.
 	fsMetaPath := pathJoin(uploadIDPath, fsMetaJSONFile)
 	fsMeta, err := readFSMetadata(fs.storage, minioMetaBucket, fsMetaPath)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by harshavardhana — APPROVED** [review]

**Comment by harshavardhana** [comment]
ping @krishnasrinivas

**Comment by krishnasrinivas** [comment]
The tmp data file should have got appended properly and renamed, the problem is somewhere else. Checking it.

**Comment by vadmeste** [comment]
> The tmp data file should have got appended properly and renamed, the problem is somewhere else. Checking it.

As I understood, there is one case when a client calls Complete MultiPart before append all parts is done in background, that is how we have this tmp file not renamed and not used, am I correct ?

**Comment by krishnasrinivas** [comment]
> As I understood, there is one case when a client calls Complete MultiPart before append all parts is done in background, that is how we have this tmp file not renamed and not used, am I correct ?

correct.

@vadmeste a race i had dismissed (to keep the code simple) that it will not happen is happening. Basically after PutObjectPart is done we call `go appendParts(fs.storage, bucket, object, uploadID, opsID)` but before this new go routine is spawned if we get `CompleteMultipartUpload()` then the behavior we see here can happen. 

try:

```
while true; do mc --debug cp 70mb myminio/testbucket/ ; done
```

i.e with --debug flag, this race will not be seen because the client will be a little slow in sending CompleteMultipartUpload()

fix is that before spawning go routine we have to do `nsMutex.Lock(minioMetaBucket, fsAppendMetaPath, opsID)` but thinking how it will affect the code flow.

**Comment by krishnasrinivas** [comment]
@vadmeste your current fix will still be needed to clean things up, but there is a comment i have added.

I am fixing the bug where the appended file won't go waste because of the race condition.

File: cmd/fs-v1-multipart.go [hunk 1]
```diff
@@ -678,12 +680,9 @@ func (fs fsObjects) CompleteMultipartUpload(bucket string, object string, upload
 
 	fsAppendMeta, err := readFSMetadata(fs.storage, minioMetaBucket, fsAppendMetaPath)
 	if err == nil && isPartsSame(fsAppendMeta.Parts, parts) {
-		fsAppendDataPath := getFSAppendDataPath(uploadID)
 		if err = fs.storage.RenameFile(minioMetaBucket, fsAppendDataPath, bucket, object); err != nil {
 			return "", toObjectErr(traceError(err), minioMetaBucket, fsAppendDataPath)
 		}
-		// Remove the append-file metadata file in tmp location as we no longer need it.
-		fs.storage.DeleteFile(minioMetaBucket, fsAppendMetaPath)
```
krishnasrinivas, line 686: you can keep this line and move the below new two lines you have added into else block. That way it will be more clear.

krishnasrinivas (changes requested):
minor comment added.

harshavardhana:
what is the final consensus here @krishnasrinivas ?

krishnasrinivas (approved):

krishnasrinivas:
@harshavardhana can be merged

harshavardhana:
> I am fixing the bug where the appended file won't go waste because of the race condition.

We should open a bug for this @vadmeste @krishnasrinivas

PR #3145, opened 2016-11-01
Start all listeners when a given host resolves to multiple IPs.

harshavardhana:
## Description
Default golang net.Listen only listens on the first IP when
host resolves to multiple IPs.

This change addresses a problem for example your ``/etc/hosts``
has entries as following

```
127.0.1.1  minio1

192.168.1.10 minio1
```

Trying to start minio as

```
minio server --address "minio1:9001" ~/Photos
```

Causes the minio server to be bound only to "127.0.1.1" which
is an incorrect behavior since we are generally interested in
`192.168.1.10` as well.

This patch addresses this issue if the hostname is resolvable
and gives back list of addresses associated with that hostname
we just bind on all of them as it is the expected behavior.

## How Has This Been Tested?
Add multiple IPs to share the same hostname in your `/etc/hosts` and then start minio server

`` minio server --address "hostname:port" ~/Photos ``

## Types of changes

## Checklist:

### DIFF
--- a/cmd/server-main.go
+++ b/cmd/server-main.go
@@ -96,12 +96,6 @@ EXAMPLES:
       $ minio {{.Name}} http://192.168.1.11/mnt/export/ http://192.168.1.12/mnt/export/ \
           http://192.168.1.13/mnt/export/ http://192.168.1.14/mnt/export/
 
-  7. Start minio server on a 4 node distributed setup. Type the following command on all the 4 nodes exactly.
-      $ minio {{.Name}} http://minio:miniostorage@192.168.1.11/mnt/export/ \
-          http://minio:miniostorage@192.168.1.12/mnt/export/ \
-          http://minio:miniostorage@192.168.1.13/mnt/export/ \
-          http://minio:miniostorage@192.168.1.14/mnt/export/
-
 `,
 }
 
--- a/cmd/server-mux.go
+++ b/cmd/server-mux.go
@@ -167,7 +167,7 @@ func (l *ListenerMux) Close() error {
 // ServerMux - the main mux server
 type ServerMux struct {
 	http.Server
-	listener        *ListenerMux
+	listeners       []*ListenerMux
 	WaitGroup       *sync.WaitGroup
 	GracefulTimeout time.Duration
 	mu              sync.Mutex // guards closed, conns, and listener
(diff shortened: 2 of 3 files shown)


cmd/server-mux.go
```diff
@@ -199,6 +199,51 @@ func NewServerMux(addr string, handler http.Handler) *ServerMux {
...
+	var addrs []string
+	if net.ParseIP(host) != nil {
+		addrs = append(addrs, host)
+	} else {
+		addrs, err = net.LookupHost(host)
+		if err != nil {
+			return nil, err
+		}
+		if len(addrs) == 0 {
+			return nil, errors.New("No known address to listen on exiting.")
```
donatello: `"No known address to listen on. Exiting."`

cmd/server-main.go
```diff
@@ -96,12 +96,6 @@ EXAMPLES:
       $ minio {{.Name}} http://192.168.1.11/mnt/export/ http://192.168.1.12/mnt/export/ \
           http://192.168.1.13/mnt/export/ http://192.168.1.14/mnt/export/
 
-  7. Start minio server on a 4 node distributed setup. Type the following command on all the 4 nodes exactly.
-      $ minio {{.Name}} http://minio:miniostorage@192.168.1.11/mnt/export/ \
-          http://minio:miniostorage@192.168.1.12/mnt/export/ \
-          http://minio:miniostorage@192.168.1.13/mnt/export/ \
-          http://minio:miniostorage@192.168.1.14/mnt/export/
-
 `,
```
donatello, line 99: Not passing creds in URL anymore or just reducing the verbosity of the help message here?

cmd/server-mux.go
```diff
@@ -215,18 +260,38 @@ func (m *ServerMux) ListenAndServeTLS(certFile, keyFile string) (err error) {
...
+					}
+					http.Redirect(w, r, u.String(), http.StatusMovedPermanently)
+				} else {
+					// Execute registered handlers
+					m.Server.Handler.ServeHTTP(w, r)
+				}
+			}),
+		)
+	}
+	err = http.Serve(listeners[len(listeners)-1],
```
donatello: Check for error only on the last listener? Seems like we should check for all? It is worth commenting this.

cmd/server-mux.go
```diff
@@ -258,18 +324,19 @@ func (m *ServerMux) ListenAndServeTLS(certFile, keyFile string) (err error) {
...
 	m.mu.Lock()
-	m.listener = listenerMux
+	m.listeners = listeners
 	m.mu.Unlock()
 
-	err = m.Server.Serve(listenerMux)
+	for _, listener := range listeners[:len(listeners)-1] {
+		go m.Server.Serve(listener)
+	}
+	err = m.Server.Serve(listeners[len(listeners)-1])
```
donatello: Same here. Why only check for error on the last listener alone?  It is worth commenting this.

donatello: If I remember the earlier discussion about this correctly, we should be checking for error when listening to any of the resolved addresses, and aborting even if any one of them errors out. [Case when a host name resolves to both a local interface and a non-local IP too]

cmd/server-main.go
harshavardhana: Not needed to provide this as example since it is a security risk, it is just a good to know feature :-)

donatello (commented):
LGTM

cmd/server-mux.go
```diff
@@ -215,81 +260,91 @@ func (m *ServerMux) ListenAndServeTLS(certFile, keyFile string) (err error) {
...
 func (m *ServerMux) Close() error {
 	m.mu.Lock()
 	if m.closed {
+		m.mu.Unlock()
 		return errors.New("Server has been closed")
 	}
 	// Closed completely.
 	m.closed = true
 
 	// Close the listener.
```
fwessels: Close the listener**s**.

fwessels:
LGTM (just one typo in comments)

PR #3152, opened 2016-11-01
Add bucket metadata state client/handler (Fixes #3022)

### LINKED ISSUE #3022: [distributed] Local peers shouldn't connect to itself for updating notifications

donatello:
## Description

- Adds an interface to update in-memory bucket metadata state called
  BucketMetaState - this interface has functions to:
     - update bucket notification configuration,
     - bucket listener configuration,
     - bucket policy configuration, and
     - send bucket event

- This interface is implemented by `localBMS` a type for manipulating
  local node in-memory bucket metadata, and by `remoteBMS` a type for
  manipulating remote node in-memory bucket metadata.

- The remote node interface, makes an RPC call, but the local node
  interface does not - it updates in-memory bucket state directly.

## Motivation and Context

#3022 

We avoid making a RPC call to the same (self) node in this change.

## How Has This Been Tested?

Tested on single and distributed minio policy and listen features.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/bucket-metadata.go
+++ b/cmd/bucket-metadata.go
@@ -1,0 +1,152 @@
+/*
+ * Minio Cloud Storage, (C) 2014-2016 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cmd
+
+import (
+	"encoding/json"
+	"net/rpc"
+)
+
+// BucketMetaState - Interface to update bucket metadata in-memory
+// state.
+type BucketMetaState interface {
+	// Updates bucket notification
+	UpdateBucketNotification(args *SetBNPArgs) error
+
+	// Updates bucket listener
+	UpdateBucketListener(args *SetBLPArgs) error
+
+	// Updates bucket policy
+	UpdateBucketPolicy(args *SetBPPArgs) error
+
+	// Sends event
+	SendEvent(args *EventArgs) error
+}
+
+// Type that implements BucketMetaState for local node.
+type localBMS struct {
+	ObjectAPI func() ObjectLayer
+}
+
+// localBMS.UpdateBucketNotification - updates in-memory global bucket
+// notification info.
+func (lc *localBMS) UpdateBucketNotification(args *SetBNPArgs) error {
+	// check if object layer is available.
+	objAPI := lc.ObjectAPI()
+	if objAPI == nil {
+		return errServerNotInitialized
+	}
+
+	globalEventNotifier.SetBucketNotificationConfig(args.Bucket, args.NCfg)
+
+	return nil
+}
+
+// localBMS.UpdateBucketListener - updates in-memory global bucket
+// listeners info.
+func (lc *localBMS) UpdateBucketListener(args *SetBLPArgs) error {
+	// check if object layer is available.
+	objAPI := lc.ObjectAPI()
+	if objAPI == nil {
+		return errServerNotInitialized
+	}
+
+	// Update in-memory notification config.
+	return globalEventNotifier.SetBucketListenerConfig(args.Bucket, args.LCfg)
+}
+
+// localBMS.UpdateBucketPolicy - updates in-memory global bucket
+// policy info.
+func (lc *localBMS) UpdateBucketPolicy(args *SetBPPArgs) error {
+	// check if object layer is available.
+	objAPI := lc.ObjectAPI()
+	if objAPI == nil {
+		return errServerNotInitialized
+	}
+
+	var pCh policyChange
+	if err := json.Unmarshal(args.PChBytes, &pCh); err != nil {
+		return err
+	}
+
+	return globalBucketPolicies.SetBucketPolicy(args.Bucket, pCh)
+}
+
+// localBMS.SendEvent - sends event to local event notifier via
+// `globalEventNotifier`
+func (lc *localBMS) SendEvent(args *EventArgs) error {
+	// check if object layer is available.
+	objAPI := lc.ObjectAPI()
+	if objAPI == nil {
+		return errServerNotInitialized
+	}
+
+	return globalEventNotifier.SendListenerEvent(args.Arn, args.Event)
+}
+
+// Type that implements BucketMetaState for remote node.
+type remoteBMS struct {
+	*AuthRPCClient
+}
+
+// remoteBMS.UpdateBucketNotification - sends bucket notification
+// change to remote peer via RPC call.
+func (rc *remoteBMS) UpdateBucketNotification(args *SetBNPArgs) error {
+	reply := GenericReply{}
+	err := rc.Call("S3.SetBucketNotificationPeer", args, &reply)
+	// Check for network error and retry once.
+	if err != nil && err.Error() == rpc.ErrShutdown.Error() {
+		err = rc.Call("S3.SetBucketNotificationPeer", args, &reply)
+	}
+	return err
+}
+
+// remoteBMS.UpdateBucketListener - sends bucket listener change to
+// remote peer via RPC call.
+func (rc *remoteBMS) UpdateBucketListener(args *SetBLPArgs) error {
+	reply := GenericReply{}
+	err := rc.Call("S3.SetBucketListenerPeer", args, &reply)
+	// Check for network error and retry once.
+	if err != nil && err.Error() == rpc.ErrShutdown.Error() {
+		err = rc.Call("S3.SetBucketListenerPeer", args, &reply)
+	}
+	return err
+}
+
+// remoteBMS.UpdateBucketPolicy - sends bucket policy change to remote
+// peer via RPC call.
+func (rc *remoteBMS) UpdateBucketPolicy(args *SetBPPArgs) error {
+	reply := GenericReply{}
+	err := rc.Call("S3.SetBucketPolicyPeer", args, &reply)
+	// Check for network error and retry once.
+	if err != nil && err.Error() == rpc.ErrShutdown.Error() {
+		err = rc.Call("S3.SetBucketPolicyPeer", args, &reply)
+	}
+	return err
+}
+
+// remoteBMS.SendEvent - sends event for bucket listener to remote
+// peer via RPC call.
+func (rc *remoteBMS) SendEvent(args *EventArgs) error {
+	reply := GenericReply{}
+	err := rc.Call("S3.Event", args, &reply)
+	// Check for network error and retry once.
+	if err != nil && err.Error() == rpc.ErrShutdown.Error() {
+		err = rc.Call("S3.Event", args, &reply)
+	}
+	return err
+}
--- a/cmd/lock-rpc-server_test.go
+++ b/cmd/lock-rpc-server_test.go
@@ -447,13 +447,14 @@ func TestLockServers(t *testing.T) {
 	}
 	globalMinioHost = ""
 	testCases := []struct {
+		isDistXL         bool
 		srvCmdConfig     serverCmdConfig
 		totalLockServers int
 	}{
 		// Test - 1 one lock server initialized.
 		{
+			isDistXL: true,
 			srvCmdConfig: serverCmdConfig{
-				isDistXL: true,
 				endpoints: []*url.URL{{
 					Scheme: "http",
 					Host:   "localhost:9000",
@@ -476,8 +477,8 @@ func TestLockServers(t *testing.T) {
 		},
 		// Test - 2 two servers possible, 1 ignored.
 		{
+			isDistXL: true,
 			srvCmdConfig: serverCmdConfig{
-				isDistXL: true,
 				endpoints: []*url.URL{{
 					Scheme: "http",
 					Host:   "localhost:9000",
@@ -507,6 +508,7 @@ func TestLockServers(t *testing.T) {
 
 	// Validates lock server initialization.
 	for i, testCase := range testCases {
+		globalIsDistXL = testCase.isDistXL
 		lockServers := newLockServers(testCase.srvCmdConfig)
 		if len(lockServers) != testCase.totalLockServers {
 			t.Fatalf("Test %d: Expected total %d, got %d", i+1, testCase.totalLockServers, len(lockServers))
--- a/cmd/s3-peer-client.go
+++ b/cmd/s3-peer-client.go
@@ -19,179 +19,168 @@ package cmd
 import (
 	"encoding/json"
 	"fmt"
-	"net/rpc"
 	"net/url"
 	"path"
 	"sync"
-	"time"
 )
 
-type s3Peers struct {
-	// A map of peer server address (in `host:port` format) to RPC
-	// client connections.
-	rpcClients map[string]*AuthRPCClient
-
-	mutex *sync.RWMutex
-
-	// Is single-node?
-	isDistXL bool
-
-	// Slice of all peer addresses (in `host:port` format).
-	peers []string
+// s3Peer structs contains the address of a peer in the cluster, and
+// its BucketMetaState interface objects.
+type s3Peer struct {
+	// address in `host:port` format
+	addr string
+	// BucketMetaState client interface
+	bmsClient BucketMetaState
 }
 
-func initGlobalS3Peers(eps []*url.URL) {
-	// Get list of de-duplicated peers.
-	peers := getAllPeers(eps)
+// type representing all peers in the cluster
+type s3Peers []s3Peer
 
-	// Initialize global state.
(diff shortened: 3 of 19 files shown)


cmd/s3-peer-client.go
```diff
@@ -19,112 +19,66 @@ package cmd
...
+	// A map of peer server address (in `host:port` format) to
+	// BucketMetaState interface
+	BMSClients map[string]BucketMetaState
 
-	mutex *sync.RWMutex
-
-	// Is single-node?
-	isDistXL bool
+	// Is distributed?
+	IsDistXL bool
```
harshavardhana: Why expose this value?

donatello: So both `IsDistXL` and `Peers` are members of this struct that are directly used in many places in the code, especially outside of the `s3-peer-client.go`, by accessing via the `globalS3Peers` global. It is just to indicate that.

harshavardhana:
ping @vadmeste @hackintoshrao @balamurugana can you guys review this patch?

vadmeste:
Don't we need to use some distributed locking here?

cmd/s3-peer-client.go
```diff
@@ -19,125 +19,74 @@ package cmd
...
 
 	// Slice of all peer addresses (in `host:port` format).
-	peers []string
+	Peers []string
```
fwessels: Peers includes 'itself' right? In that sense the name 'peers' may be a bit misleading

fwessels:
LGTM

donatello:
> Don't we need to use some distributed locking here?

@vadmeste The change: https://github.com/minio/minio/commit/3977d6b7bd9fae800b869569f4b8059bc9c2975d already took care of the required locking. In this change we are only making sure that "self" is notified directly and not via an RPC call like previously.

cmd/s3-peer-client.go
donatello: @fwessels Yes, Peers does include "itself". I was thinking of peers (as in servers with identical roles) as the group of servers in the cluster, and not with respect to the current server, so the name didn't seem misleading to me. Let me know if changing it to "Nodes" or "AllNodes" or something else would help readability, and I can do it.

fwessels: @donatello That is also a valid way of looking at it as they are all peers of each other of course. Let's keep it like this.

vadmeste:
> @vadmeste The change: 3977d6b already took care of the required locking. In this change we are only making sure that "self" is notified directly and not via an RPC call like previously.

:+1:

harshavardhana:
ping @krishnasrinivas can you review this?

krishnasrinivas:
@harshavardhana yes

cmd/bucket-metadata.go
```diff
@@ -0,0 +1,136 @@
...
+// Type that implements BucketMetaState for remote node.
+type remoteBMS struct {
+	*AuthRPCClient
+}
+
+func (rc *remoteBMS) UpdateBucketNotification(args *SetBNPArgs) error {
+	reply := GenericReply{}
+	err := rc.Call("S3.SetBucketNotificationPeer", args, &reply)
+	// Check for network error and retry once.
+	if err != nil && err.Error() == rpc.ErrShutdown.Error() {
```
krishnasrinivas, line 112: will directly comparing `err` and `rpc.ErrShutdown` be a problem?

harshavardhana: yes, the type is lost internally. this is wrapped into an unexported error by RPC which has caused issue in the past.

cmd/s3-peer-client.go
```diff
@@ -19,125 +19,74 @@ package cmd
...
-	peers = []string{globalMinioAddr} // Starts with a default peer.
-	for _, ep := range eps {
-		if ep == nil {
-			return nil
-		}
-		// Rest of the peers configured.
-		peers = append(peers, ep.Host)
-	}
-	return peers
+func (s3p *s3Peers) GetPeerClient(peer string) BucketMetaState {
```
krishnasrinivas: GetPeerClient() isn't adding value here, i.e if this was a library package and we did not want to expose s3p.bmsClients then GetPeerClient() could give us the control. Since `s3Peers` gets used by the same package, you can use s3p.bmsClients[peer] directly

cmd/s3-peer-client.go
```diff
@@ -19,125 +19,74 @@ package cmd
...
-func (s3p *s3Peers) SendRPC(peers []string, method string, args interface {
-	SetToken(token string)
-	SetTimestamp(tstamp time.Time)
-}) map[string]error {
+// Send bucket metadata updates to all given nodes. The update calls
+// are sent in parallel, and errors are returned per node in a
+// map. The updates are sent to a type implementing the
+// BucketMetaState interface. This makes sure that the local node is
+// directly updated, and remote nodes are updated via RPC calls.
+func (s3p *s3Peers) SendUpdate(peers []string, args interface{}) map[string]error {
```
krishnasrinivas: i think SendUpdate can just return `[]error` instead of a map.

cmd/s3-peer-client.go
krishnasrinivas: I see that you have removed the Lock/Unlock for map access, is it because all accesses are read only access and hence safe?

cmd/s3-peer-client.go
```diff
@@ -187,9 +142,7 @@ func (s3p *s3Peers) SendRPC(peers []string, method string, args interface {
 // request to all peers. Currently we log an error and continue.
 func S3PeersUpdateBucketNotification(bucket string, ncfg *notificationConfig) {
 	setBNPArgs := &SetBNPArgs{Bucket: bucket, NCfg: ncfg}
-	peers := globalS3Peers.GetPeers()
-	errsMap := globalS3Peers.SendRPC(peers, "S3.SetBucketNotificationPeer",
-		setBNPArgs)
+	errsMap := globalS3Peers.SendUpdate(globalS3Peers.Peers, setBNPArgs)
```
krishnasrinivas: globalS3Peers.Peers need not be sent as SendUpdate itself can access it using s3p.Peers

cmd/s3-peer-client.go
donatello: Yes the `s3Peers.mutex` was only to make it safe to read and write the map inside `s3Peers`. But it turns out that map is only written once at initialization and then only concurrently read - this will work fine - https://golang.org/doc/go1.6#runtime (and from many other sources on the net).

cmd/s3-peer-client.go
krishnasrinivas: > So both IsDistXL and Peers are members of this struct that are directly used in many places in the code, especially outside of the s3-peer-client.go, by accessing via the globalS3Peers global. It is just to indicate that.

is this a generally followed convention in the main package? if not we need to decide if we should follow this convention because we currently don't.

cmd/s3-peer-client.go
```diff
@@ -19,125 +19,74 @@ package cmd
...
 
 	// Slice of all peer addresses (in `host:port` format).
-	peers []string
+	Peers []string
 }
 
-func initGlobalS3Peers(eps []*url.URL) {
-	// Get list of de-duplicated peers.
-	peers := getAllPeers(eps)
+func mkPeersFromEndpoints(eps []*url.URL) s3Peers {
```
krishnasrinivas: this can be named `makeS3Peers` omit FromEndPoints since we are not making s3Peers from other ways.

cmd/s3-peer-client.go
donatello: In the case where some peers return errors, we could have a policy of doing a controlled number of retries (only on the failing peers). `SendUpdate` has this function type to allow for this type of retry - we need to only pass the peers that had failures. This is also the reason it returns an error map.

We currently do not do any retries, since there seems to be no objective way to decide the maximum number of retries or retries for certain types of errors only - we have to consider that bucket read locks are held during this whole operation of notifying peers. But in the future if we come up with reasons to do this, we can just use the same function.

cmd/s3-peer-client.go
donatello: Same reason as the other review comment for `SendUpdate` (appears below this comment for me).

cmd/s3-peer-client.go
```diff
@@ -19,125 +19,74 @@ package cmd
...
-		secretKey:   serverConfig.GetCredential().SecretAccessKey,
-		address:     peer,
-		secureConn:  isSSL(),
-		path:        path.Join(reservedBucket, s3Path),
-		loginMethod: "S3.LoginHandler",
+
+	return s3Peers{
+		bmsClients: bmsc,
+		IsDistXL:   len(peers) > 1,
+		Peers:      peers,
```
krishnasrinivas: can Peers field be eliminated? because the info is already present in bmsClients.

or better bmsClients can be removed and instead Peers can be []struct{host string, *AuthRPCClient}

cmd/s3-peer-client.go
krishnasrinivas: > In the case where some peers return errors, we could have a policy of doing a controlled number of retries (only on the failing peers). SendUpdate has this function type to allow for this type of retry - we need to only pass the peers that had failures. This is also the reason it returns an error map.

SendUpdate can return a slice of errors and still achieve what map can achieve and SendUpdate() function becomes simpler. a `nil` error in the slice indicates that the corresponding peer succeeded, `non-nil` indicates error.

cmd/s3-peer-client.go
donatello: Ah ok, I think this could be done. Let me rework the patch and get back on this.

cmd/bucket-metadata.go
krishnasrinivas: do you remember when that happens? if it does then it looks like a bug in rpc package and we should raise an issue.

cmd/s3-peer-client.go
```diff
@@ -19,125 +19,74 @@ package cmd
...
+	peers := []string{globalMinioAddr}
 
-	// Save new peers
-	globalS3Peers.peers = peers
-
-	// store if this is a distributed setup or not.
-	globalS3Peers.isDistXL = len(globalS3Peers.peers) > 1
-}
+	for _, ep := range eps {
+		if ep == nil || ep.Host == "" || ep.Host == globalMinioAddr {
```
krishnasrinivas: parseStorageEndpoints() ensures that ep is never `nil`, so the check `ep == nil` can be removed.

cmd/s3-peer-client.go
donatello: There is trade-off I discovered while re-working this patch. `GetPeerClient` is used in a frequently hit code path in [`notify-listener.go`](https://github.com/minio/minio/blob/master/cmd/notify-listener.go#L61) to send event notifications to peers connected to listeners. If we do not use a map, we will need to iterate over a 16-length slice each time to locate the client object to send the notification - if we store a map, we do a fast(er) map lookup.

I'm of the opinion that we use the map since this usage of (a small amount of) memory is helping performance. What say?

donatello: Hmm, perhaps it can still be done without a map. Let me think a bit more thoroughly on this and get back.

cmd/s3-peer-client.go
abperiasamy: This code is not going to turn in to a package in the future. So it is OK to keep it as lower case. Either way is not wrong, but we need to keep it consistent.

cmd/bucket-metadata.go
```diff
@@ -0,0 +1,136 @@
...
+	// Sends event
+	SendEvent(args *EventArgs) error
+}
+
+// Type that implements BucketMetaState for local node.
+type localBMS struct {
+	ObjectAPI func() ObjectLayer
+}
+
+func (lc *localBMS) UpdateBucketNotification(args *SetBNPArgs) error {
```
abperiasamy, line 47: Even the methods (both public and private) needs comments.

cmd/s3-peer-client.go
donatello: Ok lowercased.

cmd/s3-peer-client.go
donatello: Done.

cmd/s3-peer-client.go
donatello: Done. Also removed `ep.Host == globalMinioAddr` as it is just making the code look more complex but this case is handled later in the loop.

cmd/bucket-metadata.go
donatello: Done.

cmd/bucket-metadata.go
donatello: This happens because in some cases we wrap errors with `traceError()` - this hides the underlying type, so comparing `Error()` string values seems to be the only reliable way to check. It's not related to `rpc` package.

harshavardhana: If its just traceError then "errorCause()" can be used. to get the underlying type.

cmd/s3-peer-client.go
donatello: This is now done. Please review @krishnasrinivas @harshavardhana . 

`SendUpdate` now receives an array of peer indexes with a special nil value to indicate that all peers should be notified. This is done to avoid creating a trivial array of all peer indices and passing to `SendUpdate` for the common case. IMO the code does look a bit more complicated that before, but we now do not store a map, and there is no performance effect due to it.

cmd/bucket-metadata.go
donatello: @harshavardhana We had a discussion about this earlier regarding this line https://github.com/minio/minio/blob/master/cmd/net-rpc-client.go#L145 - I had proposed changing it to `if err == rpc.ErrShutdown` - my understanding was that the way we could lose/hide the type was by using things like `traceError()`, otherwise this change would work. We do not at this time use `traceError()` here (in net-rpc-client.go) since the lowerlevel code is directly in the net/rpc package. It's just that we do `traceError()` with certain goals, and we don't do it uniformly, so comparing `.Error()` values is more reliable, in case future code changed underlying error types.

Perhaps we could consider this aspect in a separate issue and move ahead with this change?

donatello:
All review comments have been addressed at this point. Could we have another round of reviews please? @krishnasrinivas @harshavardhana @fwessels @vadmeste @abperiasamy

harshavardhana:
Can you rebase this change @donatello ?

donatello:
@harshavardhana rebased now.

cmd/s3-peer-client.go
```diff
@@ -19,191 +19,185 @@ package cmd
...
-	rpcClients map[string]*AuthRPCClient
-
-	mutex *sync.RWMutex
+	// Collection of peers
+	peers struct {
+		// address of peer in `host:port` format
+		addr []string
+		// BucketMetaState object for peers
+		bmsClient []BucketMetaState
+	}
```
krishnasrinivas: this is a case for:

`struct s3Peer {addr string, bmsClient BucketMetaState}`

`peers []s3Peer`

makes the code more readable at the cost of an extra type (s3Peer)

cmd/bucket-metadata.go
krishnasrinivas: @donatello causeError() takes care of the situation, i.e in case err does not implement Error

```
func errorCause(err error) error {                                                                                                                                              
        if e, ok := err.(*Error); ok {                                                                                                                                          
                err = e.e                                                                                                                                                       
        }                                                                                                                                                                       
        return err                                                                                                                                                              
}
```

so using it safe.

this change can be done as part of different PR, sure

cmd/s3-peer-client.go
donatello: Yep this is better. Additionally we can eliminate `isDistXL` from here and make it global. It is already used in `server-main.go`. Let me update the PR again.

donatello: This is updated now.

cmd/bucket-metadata.go
harshavardhana: This assumption is still incorrect @krishnasrinivas - `errorCause()` doesn't work in all situations, it is limited to traceError() alone and it shouldn't be recommended for every situation. 

Fundamentally  in Go error is just a string so comparing String() is not that all bad as it is considered otherwise.  A wrapped error into another type is the same error as the cause underneath.  A type of 'error' is an interface and Error() is a string so in essence `rpc.ErrShutdown` is same as `fmt.Errorf("%s", rpc.ErrShutdown)` .

krishnasrinivas: @harshavardhana  my suggestion to use errorCause() was based on this information:

> This happens because in some cases we wrap errors with traceError() - this hides the underlying type, so comparing Error() string values seems to be the only reliable way to check. It's not related to rpc package.

cmd/lock-rpc-server_test.go
```diff
@@ -507,6 +508,7 @@ func TestLockServers(t *testing.T) {
 
 	// Validates lock server initialization.
 	for i, testCase := range testCases {
+		globalIsDistXL = testCase.isDistXL
```
krishnasrinivas, line 511: advantage of having global variables is simpler code base but slightly complicated test cases because we have to make sure that one test case does not have a side effect on another test case by affecting the global variable :D

but if the testing framework provides a hook before running every test case we can set the global variables to the default values. this can be explored later.

krishnasrinivas (approved):

cmd/lock-rpc-server_test.go
harshavardhana: Can you open a bug for this @krishnasrinivas ?

PR #3162, opened 2016-11-02
sigv4: Trim and shrink spaces in headers values

### LINKED ISSUE #3158: Put object with user metadata which contains two and more spaces always return exception "SignatureDoesNotMatch"
## Expected Behavior

Object registered into storage with metadata.

## Current Behavior

Always return the same error:
```xml
<Error>
    <Code>SignatureDoesNotMatch</Code>
    <Message>The request signature we calculated does not match the signature you provided. Check your key and signing method.
    </Message>
    <Key></Key>
    <BucketName></BucketName>
    <Resource>/testbucket/2016/11/2/myFileName</Resource>
    <RequestId>3L137</RequestId>
    <HostId>3L137</HostId>
</Error>
```

1. Create bucket with name "testbucket"
2. Try put content into Minio object store.

Here is java code sample:
```java
import com.amazonaws.services.s3.AmazonS3;
import com.amazonaws.services.s3.model.ObjectMetadata;

private AmazonS3 s3Client;

public void writeDocument() {
    final byte[] content = "binaryContent".getBytes(Charset.forName("UTF-8"));

    final ObjectMetadata objectMetadata = new ObjectMetadata();
    // "foo  bar" two and more spaces between foo and bar
    objectMetadata.addUserMetadata("my-meta", "foo  bar");
    objectMetadata.setContentLength(content.length);
    try (ByteArrayInputStream bis = new ByteArrayInputStream(content)) {
        s3Client.putObject("testbucket", "2016/11/2/myFileName", bis, objectMetadata);
    } catch (IOException e) {
        throw new IllegalStateException(e);
    }
}
```
## Your Environment

Java version: 1.8
Client: com.amazonaws:aws-java-sdk-s3:1.11.11
Minio: Docker latest (the same problem in Docker edge)

vadmeste:
## Description
The current code doesn't use Trimall() described in  http://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html to create the canonical request and calculate the signature

## Motivation and Context
Fixes #3158 

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:

### DIFF
--- a/cmd/signature-v4-utils.go
+++ b/cmd/signature-v4-utils.go
@@ -157,3 +157,19 @@ func extractSignedHeaders(signedHeaders []string, reqHeaders http.Header) (http.
 
 	return extractedSignedHeaders, ErrNone
 }
+
+// Trim leading and trailing spaces and replace sequential spaces with one space, following Trimall()
+// in http://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html
+func signV4TrimAll(input string) string {
+	// Remove all whitespaces first
+	cleanWhiteSpaces := func(r rune) rune {
+		switch r {
+		case '\t', '\n', '\u000b', '\r', '\f':
+			return ' '
+		}
+		return r
+	}
+	input = strings.Map(cleanWhiteSpaces, input)
+	// Compress adjacent spaces to one space and return
+	return strings.Join(strings.Fields(input), " ")
+}
--- a/cmd/signature-v4-utils_test.go
+++ b/cmd/signature-v4-utils_test.go
@@ -211,3 +211,31 @@ func TestFindHost(t *testing.T) {
 		t.Fatalf("Expected the APIErrorCode to be %d, but got %d", ErrNone, errCode)
 	}
 }
+
+// TestSigV4TrimAll - tests the logic of TrimAll() function
+func TestSigV4TrimAll(t *testing.T) {
+	testCases := []struct {
+		// Input.
+		inputStr string
+		// Expected result.
+		result string
+	}{
+		{"本語", "本語"},
+		{" abc ", "abc"},
+		{" a b ", "a b"},
+		{"a b ", "a b"},
+		{"a  b", "a b"},
+		{"a   b", "a b"},
+		{"   a   b  c   ", "a b c"},
+		{"a \t b  c   ", "a b c"},
+		{"\"a \t b  c   ", "\"a b c"},
+	}
+
+	// Tests generated values from url encoded name.
+	for i, testCase := range testCases {
+		result := signV4TrimAll(testCase.inputStr)
+		if testCase.result != result {
+			t.Errorf("Test %d: Expected sigV4TrimAll result to be \"%s\", but found it to be \"%s\" instead", i+1, testCase.result, result)
+		}
+	}
+}
--- a/cmd/signature-v4.go
+++ b/cmd/signature-v4.go
@@ -68,7 +68,7 @@ func getCanonicalHeaders(signedHeaders http.Header, host string) string {
 				if idx > 0 {
 					buf.WriteByte(',')
 				}
-				buf.WriteString(v)
+				buf.WriteString(signV4TrimAll(v))
 			}
 			buf.WriteByte('\n')
 		}


harshavardhana (approved):

donatello:
> The current code doesn't use Trimall() described in http://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html to create the canonical request and calculate the signature

Just want to note an inconsistency in AWS documentation about this - https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html does not mention compressing contiguous whitespace into a single whitespace. `Trimall()` is mentioned in their general API documentation only.

cmd/signature-v4-utils.go
```diff
@@ -157,3 +157,15 @@ func extractSignedHeaders(signedHeaders []string, reqHeaders http.Header) (http.
...
+var seqSpaces = regexp.MustCompile("  +")
+
+// Trim leading and trailing spaces and replace sequential spaces with one space, following Trimall()
+// in http://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html
+func sigV4TrimAll(input string) string {
+	input = strings.TrimPrefix(input, " ")
+	input = strings.TrimSuffix(input, " ")
+	replaced := seqSpaces.ReplaceAllString(input, " ")
+	return replaced
+}
```
donatello, line 175: The documentation ("spec") seems to be a bit vague `The Trimall function removes excess white space before and after values, and converts sequential spaces to a single space.` - so what do we do in case prefix is whitespace other than spaces (like tabs) - e.g. in case input is `\tabc\t sc\t\t` - what should be used in the canonical request is a bit unclear. Maybe we can refer to what an AWS released library does in these cases.

cmd/signature-v4-utils.go
```diff
@@ -157,3 +157,15 @@ func extractSignedHeaders(signedHeaders []string, reqHeaders http.Header) (http.
...
 }
+
+// Sequential spaces regular expression
+var seqSpaces = regexp.MustCompile("  +")
+
+// Trim leading and trailing spaces and replace sequential spaces with one space, following Trimall()
+// in http://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html
+func sigV4TrimAll(input string) string {
+	input = strings.TrimPrefix(input, " ")
+	input = strings.TrimSuffix(input, " ")
```
donatello: According to the spec, I think we need to remove every leading and trailing space. `TrimPrefix` or `TrimSuffix` do not do that. See the example from the documentation https://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html -

Original headers

```
Host:iam.amazonaws.com\n
Content-Type:application/x-www-form-urlencoded; charset=utf-8\n
My-header1:    a   b   c  \n
X-Amz-Date:20150830T123600Z\n
My-Header2:    "a   b   c"  \n

```

Canonical form

```
content-type:application/x-www-form-urlencoded; charset=utf-8\n
host:iam.amazonaws.com\n
my-header1:a b c\n
my-header2:"a b c"\n
x-amz-date:20150830T123600Z\n
```

cmd/signature-v4-utils_test.go
```diff
@@ -211,3 +211,29 @@ func TestFindHost(t *testing.T) {
...
+		// Expected result.
+		result string
+	}{
+		{"本語", "本語"},
+		{" abc ", "abc"},
+		{" a b ", "a b"},
+		{"a b ", "a b"},
+		{"a  b", "a b"},
+		{"a   b", "a b"},
+		{"   a   b  c   ", " a b c "},
```
donatello: This case ought to be like:

```
{"     a     b    c     ", "a b c"},
```

donatello (changes requested):

cmd/signature-v4-utils.go
harshavardhana: @vadmeste we need to use strings.TrimSpace() instead of strings.TrimPrefix or Suffix for single space characters.

cmd/signature-v4-utils.go
harshavardhana: This requires us to test it properly with S3 and see how it behaves and comply.

cmd/signature-v4-utils.go
vadmeste: :+1:

vadmeste: @harshavardhana, actually, it looks like all the trimming stuffs are done before the code reaches at this point, that's why things are working as expected though sigV4TrimAll() is buggy

vadmeste: Well, I understand now.. it is aws-java-sdk which is cleaning the header value before sending it to the server.

cmd/signature-v4-utils.go
vadmeste: Well, I've just tested this header value which works with S3 but not with Minio: `\"\tfoo  bar\"`

vadmeste: https://github.com/aws/aws-sdk-java/blob/8b2d07337f838a75b4275e8a5eb89f035ff0f06d/aws-java-sdk-core/src/main/java/com/amazonaws/util/StringUtils.java#L276 

This is the function which cleans header value, it replaces white spaces (\t, \n and others) with real spaces and compress spaces later.

cmd/signature-v4-utils.go
```diff
@@ -157,3 +157,25 @@ func extractSignedHeaders(signedHeaders []string, reqHeaders http.Header) (http.
...
+		case '\t', '\n', '\u000b', '\r', '\f':
+			return ' '
+		}
+		return r
+	}
+	input = strings.Map(cleanWhiteSpaces, input)
+	// Clean leading and trailing spaces
+	input = strings.TrimSpace(input)
+	// Compress adjacent spaces to one space
+	replaced := seqSpaces.ReplaceAllString(input, " ")
```
donatello: I think this function can be simplified to just one line:

```
return strings.Join(strings.Fields(input), " ")
```

Just found https://golang.org/pkg/strings/#Fields - splits string after compressing consecutive whitespace. Uses unicode's definition of whitespace: https://golang.org/pkg/unicode/#IsSpace https://en.wikipedia.org/wiki/Whitespace_character#Unicode

cmd/signature-v4-utils.go
```diff
@@ -157,3 +157,25 @@ func extractSignedHeaders(signedHeaders []string, reqHeaders http.Header) (http.
 
 	return extractedSignedHeaders, ErrNone
 }
+
+// Sequential spaces regular expression
+var seqSpaces = regexp.MustCompile("  +")
+
+// Trim leading and trailing spaces and replace sequential spaces with one space, following Trimall()
+// in http://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html
+func sigV4TrimAll(input string) string {
```
krishnasrinivas: can be called `signV4TrimAll` instead of `sigV4TrimAll` as we consistently use `sign` instead of `sig`

krishnasrinivas: the code in object-handlers_test.go and test-utils_test.go uses SigV4 though: malformChunkSizeSigV4() malformDataSigV4() truncateChunkByHalfSigv4() it can be fixed in a cleanup patch

harshavardhana: That should be renamed properly.

PR #3172, opened 2016-11-03
router: PathPrefix router was wrong.

harshavardhana:
## Description
Current router path handling was wrong, it was working fine due to some mutability inside gorilla/mux but this is an important bug. 

## Types of changes

## Checklist:

### DIFF
--- a/cmd/api-router.go
+++ b/cmd/api-router.go
@@ -34,7 +34,7 @@ func registerAPIRouter(mux *router.Router) {
 	}
 
 	// API Router
-	apiRouter := router.NewRouter().PathPrefix("/").Subrouter()
+	apiRouter := mux.NewRoute().PathPrefix("").Subrouter()
 
 	// Bucket router
 	bucket := apiRouter.PathPrefix("/{bucket}").Subrouter()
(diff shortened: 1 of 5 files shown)


vadmeste:
Can we discuss this before you merge? there is a story behind this.

harshavardhana:
> Can we discuss this before you merge? there is a story behind this.

Just pinged you what is the story?

fwessels:
LGTM

krishnasrinivas (approved):

cmd/api-router.go
```diff
@@ -34,7 +34,7 @@ func registerAPIRouter(mux *router.Router) {
 	}
 
 	// API Router
-	apiRouter := router.NewRouter().PathPrefix("/").Subrouter()
+	apiRouter := mux.NewRoute().PathPrefix("/").Subrouter()
```
vadmeste: This has bad effect on the handlers per route below (time validity handler), but looks like it works like this:

```
apiRouter := mux.NewRoute().PathPrefix("").Subrouter()
```

Isn't this more correct ? especially '/' is already added in bucket router.

harshavardhana: I dont see any tests failing ?

harshavardhana: Also ping me on how to reproduce it will see a cleaner fix. The existing route is wrong and can lead to bugs.

vadmeste (approved):

PR #3177, opened 2016-11-04
connection muxer should use bufio.Reader to be simpler.

harshavardhana:
## Description
This change is rather regarding simplicity and readability of the net.Conn wrapper. 

## Motivation and Context
To simplify. 

## How Has This Been Tested?
Manually and using 'go tests'. 

## Types of changes

## Checklist:

### DIFF
--- a/cmd/server-mux.go
+++ b/cmd/server-mux.go
@@ -17,15 +17,23 @@
 package cmd
 
 import (
+	"bufio"
 	"crypto/tls"
 	"errors"
-	"io"
 	"net"
 	"net/http"
 	"net/url"
 	"strings"
 	"sync"
 	"time"
+)
+
+// The value choosen below is longest word choosen
+// from all the http verbs comprising of
+// "PRI", "OPTIONS", "GET", "HEAD", "POST",
+// "PUT", "DELETE", "TRACE", "CONNECT".
+const (
+	maxHTTPVerbLen = 7
 )
 
 var defaultHTTP2Methods = []string{
@@ -43,64 +51,40 @@ var defaultHTTP1Methods = []string{
 	"CONNECT",
 }
 
-// ConnBuf - contains network buffer to record data
-type ConnBuf struct {
-	buffer []byte
-	unRead bool
-	offset int
-}
-
-// ConnMux - implements a Read() which streams twice the firs bytes from
-// the incoming connection, to help peeking protocol
+// ConnMux - Peeks into the incoming connection for relevant
+// protocol without advancing the underlying net.Conn (io.Reader).
+// ConnMux - allows us to multiplex between TLS and Regular HTTP
+// connections on the same listeners.
 type ConnMux struct {
 	net.Conn
-	lastError error
-	dataBuf   ConnBuf
-}
-
-func longestWord(strings []string) int {
-	maxLen := 0
-	for _, m := range defaultHTTP1Methods {
-		if maxLen < len(m) {
-			maxLen = len(m)
-		}
-	}
-	for _, m := range defaultHTTP2Methods {
-		if maxLen < len(m) {
-			maxLen = len(m)
-		}
-	}
-
-	return maxLen
+	bufrw *bufio.ReadWriter
 }
 
 // NewConnMux - creates a new ConnMux instance
 func NewConnMux(c net.Conn) *ConnMux {
-	h1 := longestWord(defaultHTTP1Methods)
-	h2 := longestWord(defaultHTTP2Methods)
-	max := h1
-	if h2 > max {
-		max = h2
+	br := bufio.NewReader(c)
+	bw := bufio.NewWriter(c)
+	return &ConnMux{
+		Conn:  c,
+		bufrw: bufio.NewReadWriter(br, bw),
 	}
-	return &ConnMux{Conn: c, dataBuf: ConnBuf{buffer: make([]byte, max+1)}}
 }
 
 // PeekProtocol - reads the first bytes, then checks if it is similar
 // to one of the default http methods
 func (c *ConnMux) PeekProtocol() string {
-	var n int
-	n, c.lastError = c.Conn.Read(c.dataBuf.buffer)
-	if n == 0 || (c.lastError != nil && c.lastError != io.EOF) {
-		return ""
+	buf, err := c.bufrw.Peek(maxHTTPVerbLen)
+	if err != nil {
+		errorIf(err, "Unable to peek into the protocol")
+		return "http"
 	}
-	c.dataBuf.unRead = true
 	for _, m := range defaultHTTP1Methods {
-		if strings.HasPrefix(string(c.dataBuf.buffer), m) {
+		if strings.HasPrefix(string(buf), m) {
 			return "http"
 		}
 	}
 	for _, m := range defaultHTTP2Methods {
-		if strings.HasPrefix(string(c.dataBuf.buffer), m) {
+		if strings.HasPrefix(string(buf), m) {
 			return "http2"
 		}
 	}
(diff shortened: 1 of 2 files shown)


cmd/server-mux.go
```diff
@@ -110,28 +105,25 @@ func (c *ConnMux) PeekProtocol() string {
...
+	}
+	if c.bufwr == nil {
+		return 0, io.EOF
+	}
+	return c.bufwr.Read(b)
+}
+
+// Close the connection.
+func (c *ConnMux) Close() (err error) {
+	fmt.Println("Closing. ", c.RemoteAddr())
```
vadmeste: uneeded log, otherwise LGTM

cmd/server-mux.go
```diff
@@ -70,37 +65,35 @@ func longestWord(strings []string) int {
...
-	h1 := longestWord(defaultHTTP1Methods)
-	h2 := longestWord(defaultHTTP2Methods)
-	max := h1
-	if h2 > max {
-		max = h2
+	br := bufio.NewReader(c)
+	bw := bufio.NewWriter(c)
+	return &ConnMux{
+		Conn:    c,
+		bufwr:   bufio.NewReadWriter(br, bw),
```
krishnasrinivas: you could do `bufwr:   bufio.NewReadWriter(c, c)`?

cmd/server-mux.go
```diff
@@ -70,37 +65,35 @@ func longestWord(strings []string) int {
...
 
 // PeekProtocol - reads the first bytes, then checks if it is similar
 // to one of the default http methods
 func (c *ConnMux) PeekProtocol() string {
-	var n int
-	n, c.lastError = c.Conn.Read(c.dataBuf.buffer)
-	if n == 0 || (c.lastError != nil && c.lastError != io.EOF) {
-		return ""
+	var buf []byte
+	buf, c.lastError = c.bufwr.Peek(c.readLen)
```
krishnasrinivas: since c.readLen will always be the same value, it can be inited to a global constant and removed from ConnMux

cmd/server-mux.go
```diff
@@ -110,28 +103,21 @@ func (c *ConnMux) PeekProtocol() string {
...
-		return n, c.lastError
+	if c.lastError != nil {
+		return 0, c.lastError
+	}
+	return c.bufwr.Read(b)
+}
+
+// Close the connection.
+func (c *ConnMux) Close() (err error) {
+	if c.bufwr != nil {
```
krishnasrinivas: since c.bufwr is always `non-nil` this check can be removed

cmd/server-mux.go
```diff
@@ -110,28 +103,21 @@ func (c *ConnMux) PeekProtocol() string {
...
+		c.bufwr.Flush()
+	}
+	if c.Conn != nil {
```
krishnasrinivas: c.Conn is also never `nil` hence the check can be removed

cmd/server-mux.go
```diff
@@ -110,28 +103,21 @@ func (c *ConnMux) PeekProtocol() string {
...
-				nr, err := c.Conn.Read(tmpBuffer)
-				for idx, val := range tmpBuffer {
-					b[n+idx] = val
-				}
-				return n + nr, err
-			}
-		}
-		// We here return the last error
-		return n, c.lastError
+	if c.lastError != nil {
```
krishnasrinivas: `lastError` field can be removed from ConnMux as c.bufwr.Read(b) below will return error if any.

cmd/server-mux.go
```diff
@@ -110,28 +103,21 @@ func (c *ConnMux) PeekProtocol() string {
...
+		return 0, c.lastError
+	}
+	return c.bufwr.Read(b)
+}
+
+// Close the connection.
+func (c *ConnMux) Close() (err error) {
+	if c.bufwr != nil {
+		c.bufwr.Flush()
```
krishnasrinivas: Flush() returns `error` which should be checked

cmd/server-mux.go
```diff
@@ -43,22 +43,17 @@ var defaultHTTP1Methods = []string{
...
-	dataBuf   ConnBuf
+	readLen   int
+	bufwr     *bufio.ReadWriter
 }
 
-func longestWord(strings []string) int {
+// Find the longest word of all the HTTP methods.
+func longestWord() int {
 	maxLen := 0
 	for _, m := range defaultHTTP1Methods {
```
krishnasrinivas: instead of having 2 loops one for defaultHTTP1Methods and one for defaultHTTP2Methods, this can be done:

`for _, m := range append(defaultHTTP1Methods, defaultHTTP2Methods...) {...}`

cmd/server-mux.go
```diff
@@ -43,22 +43,17 @@ var defaultHTTP1Methods = []string{
...
 
-// ConnBuf - contains network buffer to record data
-type ConnBuf struct {
-	buffer []byte
-	unRead bool
-	offset int
-}
-
 // ConnMux - implements a Read() which streams twice the firs bytes from
 // the incoming connection, to help peeking protocol
```
krishnasrinivas: comment needs to be updated to reflect the new code

cmd/server-mux.go
harshavardhana: No bufio.NewReadWriter requires separate arguments.

cmd/server-mux.go
harshavardhana: This needs to be returned before proceeding to read. Its a behavior to indicate that we saw an error  during read before.

krishnasrinivas: removing `lastError` does not cause any harm because `c.bufwr.Read(b)` will return error anyway if `buf, c.lastError = c.bufwr.Peek(c.readLen)` Peek() had returned error. the benefit is simpler code as you don't need to maintain error state from Peek() in ConnMux.

cmd/server-mux.go
krishnasrinivas: `bufwr` sounds like `buffered write` which is not correct because it also does buffered read.
`bufConn` is a better name - i.e buffered version of `Conn`

harshavardhana: It means bufwr - bufferred writer and reader.  It is not a buffered version of Conn.

krishnasrinivas: > It means bufwr - bufferred writer and reader.

you know it because you wrote it, but for someone new reading the code bufwr reads like `buffer write`

> It is not a buffered version of Conn.

it is buffered Conn:

```
br := bufio.NewReader(c)
bw := bufio.NewWriter(c)
bufwr: bufio.NewReadWriter(br, bw)
```

harshavardhana: It is not buffered Conn, it has a whole new meaning.  Also i don't see how anyone would read a return back from bufio.NewReadWriter() as just buffered writer.

harshavardhana:
Anything else pending on this @krishnasrinivas @vadmeste ?

vadmeste (approved):

cmd/server-mux.go
```diff
@@ -28,6 +28,14 @@ import (
 	"time"
 )
 
+// The value choosen below is longest word choosen
+// from all the http verbs comprising of
+// "PRI", "OPTIONS", "GET", "HEAD", "POST",
+// "PUT", "DELETE", "TRACE", "CONNECT".
+const (
+	longestWord = 7
```
krishnasrinivas: can be `maxHTTPVerbLen` because `longestWord` is too generic a name for a global constant

cmd/server-mux.go
```diff
@@ -43,64 +51,42 @@ var defaultHTTP1Methods = []string{
...
 }
 
 // PeekProtocol - reads the first bytes, then checks if it is similar
 // to one of the default http methods
 func (c *ConnMux) PeekProtocol() string {
-	var n int
-	n, c.lastError = c.Conn.Read(c.dataBuf.buffer)
-	if n == 0 || (c.lastError != nil && c.lastError != io.EOF) {
-		return ""
+	buf, err := c.bufrw.Peek(c.readLen)
```
krishnasrinivas: global `longestWord` can be used directly here and ConnMux.readLen can be eliminated

cmd/server-mux.go
```diff
@@ -43,64 +51,42 @@ var defaultHTTP1Methods = []string{
...
-	}
-	for _, m := range defaultHTTP2Methods {
-		if maxLen < len(m) {
-			maxLen = len(m)
-		}
-	}
-
-	return maxLen
+	readLen int
+	bufrw   *bufio.ReadWriter
```
krishnasrinivas: :+1:

cmd/server-mux.go
```diff
@@ -220,7 +193,7 @@ func (l *ListenerMux) Accept() (net.Conn, error) {
 	if err != nil {
 		return conn, err
 	}
-	connMux := NewConnMux(conn)
+	connMux := NewConnMux(conn, longestWord+1)
```
krishnasrinivas: read length should be `longestWord` and not `longestWord+1` right?

fwessels (commented):
LGTM, nice simplification

cmd/server-mux.go
vadmeste: The purpose of adding one was to include the space which is always present after the HTTP verb, but it looks like we are not using it anymore.

cmd/server-mux.go
krishnasrinivas: defining it as const like this is also fine

another way, maxHTTPVerbLen can also be inited programatically :

```
var maxHTTPVerbLen = func() int {                                                      
        max := 0                                                                       
        for _, i := range append(defaultHTTP1Methods, defaultHTTP2Methods...) {        
                if len(i) > max {                                                      
                        max = len(i)                                                   
                }                                                                      
        }                                                                              
        return max                                                                     
}()
```

cmd/server-mux.go
harshavardhana: Removed it not required as such.

krishnasrinivas (approved):

PR #3179, opened 2016-11-04
Refactor: simplify signV4TrimAll()

donatello:
## Description

`strings.Fields` already takes care of removing space characters that is being done with `cleanWhiteSpaces`. This change is removing some unneeded code and adds a test case to show that the whitespace characters `"\t\n\u000b\r\f"` are being removed as required.

## Motivation and Context

Makes code a bit simpler.

## How Has This Been Tested?

Appropriate test case has been added.

## Types of changes

- [x] Refactor maintaining existing functionality.

## Checklist:

### DIFF
--- a/cmd/signature-v4-utils.go
+++ b/cmd/signature-v4-utils.go
@@ -161,15 +161,7 @@ func extractSignedHeaders(signedHeaders []string, reqHeaders http.Header) (http.
 // Trim leading and trailing spaces and replace sequential spaces with one space, following Trimall()
 // in http://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html
 func signV4TrimAll(input string) string {
-	// Remove all whitespaces first
-	cleanWhiteSpaces := func(r rune) rune {
-		switch r {
-		case '\t', '\n', '\u000b', '\r', '\f':
-			return ' '
-		}
-		return r
-	}
-	input = strings.Map(cleanWhiteSpaces, input)
-	// Compress adjacent spaces to one space and return
+	// Compress adjacent spaces (a space is determined by
+	// unicode.IsSpace() internally here) to one space and return
 	return strings.Join(strings.Fields(input), " ")
 }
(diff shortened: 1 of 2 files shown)


vadmeste (approved):

cmd/signature-v4-utils.go
```diff
@@ -161,15 +161,7 @@ func extractSignedHeaders(signedHeaders []string, reqHeaders http.Header) (http.
...
-		switch r {
-		case '\t', '\n', '\u000b', '\r', '\f':
-			return ' '
-		}
-		return r
-	}
-	input = strings.Map(cleanWhiteSpaces, input)
-	// Compress adjacent spaces to one space and return
+	// Compress adjacent spaces (a space is determined by
+	// unicode.IsSpace() internally here) to one space and return
```
krishnasrinivas, line 165: unicode.IsSpace considers `'\t', '\n', '\v', '\f', '\r', ' ', 0x85, 0xA0:` as white spaces, whereas aws-sdk-java considers `'\t', '\n', '\u000b', '\r', '\f'` as white spaces. It should be fine, we have a much simpler function.

krishnasrinivas (approved):

cmd/signature-v4-utils.go
donatello: Sorry to nit-pick but `unicode.IsSpace` actually considers all unicode classified space characters too - the above list is only in the Latin-1 subset - relevant doc snippet:

```
 Other definitions of spacing characters are set by category Z and property Pattern_White_Space. 
```

Still code is simpler, and I don't think it will every break - header fields are only ever expected to be generally compatible when encoded as ASCII (https://stackoverflow.com/questions/4400678/what-character-encoding-should-i-use-for-a-http-header#4410331)

krishnasrinivas: > I don't think it will every break

that's why i said `It should be fine`, mostly white space chars in header values will be ' ' or '\t`, if it ever breaks this function can be re-written.

PR #3187, opened 2016-11-05
Rewrite connection muxer to avoid blocking server by silent clients

### LINKED ISSUE #3159: No response when use mc or client-go to connnect to minio server?
I have find that it I can not get any response when  use mc to connnect to minio server? It seems like it's blocked there, but webui workes fine. At the same time if I refresh the webui, the result is output immediately, I do not know why. This is not always the case, when the service has just started, it is very easy to appear.

vadmeste:
Need rebase and code reread and test again for edge cases

## Description
Fix one problem with connection peeker which was blocking when a client opens a connection and doesn't send any data. Peeking is moved in a go routine, more changes are done to satisfy the new code flow.

## Motivation and Context
Fixes #3159 

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:

### DIFF
--- a/cmd/server-mux.go
+++ b/cmd/server-mux.go
@@ -134,9 +134,50 @@ func (c *ConnMux) Close() (err error) {
 type ListenerMux struct {
 	net.Listener
 	config *tls.Config
+	// acceptResCh is a channel for transporting wrapped net.Conn (regular or tls)
+	// after peeking the content of the latter
+	acceptResCh chan ListenerMuxAcceptRes
 	// Cond is used to signal Close when there are no references to the listener.
 	cond *sync.Cond
 	refs int
+}
+
+// ListenerMuxAcceptRes contains then final net.Conn data (wrapper by tls or not) to be sent to the http handler
+type ListenerMuxAcceptRes struct {
+	conn net.Conn
+	err  error
+}
+
+// newListenerMux listens and wraps accepted connections with tls after protocol peeking
+func newListenerMux(listener net.Listener, config *tls.Config) *ListenerMux {
+	l := ListenerMux{
+		Listener:    listener,
+		config:      config,
+		cond:        sync.NewCond(&sync.Mutex{}),
+		acceptResCh: make(chan ListenerMuxAcceptRes),
+	}
+	// Start listening, wrap connections with tls when needed
+	go func() {
+		// Loop for accepting new connections
+		for {
+			conn, err := l.Listener.Accept()
+			if err != nil {
+				l.acceptResCh <- ListenerMuxAcceptRes{err: err}
+				return
+			}
+			// Wrap the connection with ConnMux to be able to peek the data in the incoming connection
+			// and decide if we need to wrap the connection itself with a TLS or not
+			go func(conn net.Conn) {
+				connMux := NewConnMux(conn)
+				if connMux.PeekProtocol() == "tls" {
+					l.acceptResCh <- ListenerMuxAcceptRes{conn: tls.Server(connMux, l.config)}
+				} else {
+					l.acceptResCh <- ListenerMuxAcceptRes{conn: connMux}
+				}
+			}(conn)
+		}
+	}()
+	return &l
 }
 
 // IsClosed - Returns if the underlying listener is closed fully.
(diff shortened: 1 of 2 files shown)


cmd/server-mux.go
```diff
@@ -281,6 +276,50 @@ func initListeners(serverAddr string, tls *tls.Config) ([]*ListenerMux, error) {
...
+	// Listen and serve for plain http requests
+	httpListener := customListener{Listener: l, ch: make(chan net.Conn)}
+	go http.Serve(httpListener, handler)
+
+	// Listen and serve for https requests
+	tlsListener := customListener{Listener: l, ch: make(chan net.Conn)}
+	go http.Serve(tlsListener, handler)
+
+	for {
+		// Wait for new connections
```
harshavardhana: This change looks pretty good :+1:

harshavardhana:
Can you test this with - https://github.com/minio/minio/files/572316/gof3r.gz ? concurrency tests.. 

```
gof3r put -p ~/xubuntu-xenial-mate-20160528-longsleep-pine64-8GB.zip -k obj -b testbucket --endpoint=localhost:9000 --path-style --no-ssl
```

harshavardhana:
Also please test with "aws cli" with large uploads.

krishnasrinivas:
@vadmeste this way will be simpler https://github.com/krishnasrinivas/minio/commit/61f2a87de9c62b21c1feb0106312549772129cb3

i.e without using another net.Listener abstraction, the channel can be made part of ListenerMux itself.

(the for loop in newListenerMux() in there can be moved out to separate function too)

Also renaming ListenerMux to ListenerCloser is not correct. i.e ListenerCloser would be right name for the type:

```
type ListenerCloser interface {
    net.Listener
    io.Closer
}
```

vadmeste:
Thanks @krishnasrinivas, yes, it is more simpler that way, I've also made some simple corrections.

@harshavardhana, I made tests with 8gb files with aws cli and gof3r and both are fine with the new server-mux mechanism.

cmd/server-mux.go
```diff
@@ -133,12 +133,47 @@ func (c *ConnMux) Close() (err error) {
...
+
+// newListenerMux - wraps the accepted connection with tls after protocol peeking
+func newListenerMux(listener net.Listener, config *tls.Config) *ListenerMux {
+	l := ListenerMux{
+		Listener:    listener,
+		config:      config,
+		cond:        sync.NewCond(&sync.Mutex{}),
+		acceptResCh: make(chan ListenerMuxAcceptRes),
+	}
+	go func() {
```
harshavardhana, line 160: Needs comments here for each `go func()`,  `for {`  and `go func()`

cmd/server-mux.go
```diff
@@ -133,12 +133,47 @@ func (c *ConnMux) Close() (err error) {
...
+		for {
+			conn, err := l.Listener.Accept()
+			if err != nil {
+				l.acceptResCh <- ListenerMuxAcceptRes{conn: conn, err: err}
+				return
+			}
+			go func(conn net.Conn) {
+				connMux := NewConnMux(conn)
+				if connMux.PeekProtocol() == "tls" {
+					l.acceptResCh <- ListenerMuxAcceptRes{tls.Server(connMux, l.config), nil}
```
harshavardhana: No need to pass 'explicit' nil.  it is understood. Used the named variables instead.

cmd/server-mux.go
```diff
@@ -133,12 +133,47 @@ func (c *ConnMux) Close() (err error) {
...
+		Listener:    listener,
+		config:      config,
+		cond:        sync.NewCond(&sync.Mutex{}),
+		acceptResCh: make(chan ListenerMuxAcceptRes),
+	}
+	go func() {
+		for {
+			conn, err := l.Listener.Accept()
+			if err != nil {
+				l.acceptResCh <- ListenerMuxAcceptRes{conn: conn, err: err}
```
harshavardhana: Here lets not send 'conn' as if its valid, we should just set `err: err`

cmd/server-mux.go
```diff
@@ -133,12 +133,47 @@ func (c *ConnMux) Close() (err error) {
 // from returning from Close till Accept returns an error to the user.
 type ListenerMux struct {
 	net.Listener
-	config *tls.Config
+	config      *tls.Config
+	acceptResCh chan ListenerMuxAcceptRes
```
harshavardhana, line 139: Write a comment about the use of this field.

harshavardhana:
> @harshavardhana, I made tests with 8gb files with aws cli and gof3r and both are fine with the new server-mux mechanism.

Okay cool - address the comments and this is ready to be merged. :+1:

fwessels:
LGTM

PR #3202, opened 2016-11-08
Add `withQuorum` flag to isUploadIDExists:

### LINKED ISSUE #3149: [distributed] Stopping and restarting a node in the middle of a upload *always* returns XMinioWriteQuorum error
## Expected Behavior
If we have 4 running nodes, stopping and restarting a node during a *multipart* file upload shouldn't prevent that file to be uploaded again  

## Current Behavior
Cannot upload by that name anymore, even if I restart all nodes or restart upload.

1. Start 4 nodes from fresh disks
2. Upload a 70mb file
3. Stop any container and start it again
4. Try to upload the same file again without changing the dest name.

## Your Environment
Docker edge

donatello:
## Description

- withQuorum flag is used to check if the uploadID exists on quorum
  disks and is used in all mutating operations in multipart
  uploads (i.e. put object part, abort multipart and complete
  multipart).

- in listObjectParts, isUploadIDExists is called with withQuorum=false
  for better average case performance.

## Motivation and Context

Possible way to resolve https://github.com/minio/minio/issues/3149:

In case a stale upload id is attempted to be resumed, instead of returning misleading/buggy write quorum errors, the server will return that the upload id is invalid.

## How Has This Been Tested?

Existing test case validates it.

Manual testing with mc, now causes mc to return:

```
$ mc cp /tmp/stop.mp4 dminio1/testb
mc: <ERROR> Failed to copy ‘/tmp/stop.mp4’. The specified multipart upload does not exist. The upload ID may be invalid, or the upload may have been aborted or completed.
```

mc is not returning a nonsensical quorum error any more.

[The fact that it is failing to upload is itself a bug, but an [independent one](https://github.com/minio/mc/issues/1874)]

## Types of changes

## Checklist:

### DIFF
--- a/cmd/xl-v1-multipart.go
+++ b/cmd/xl-v1-multipart.go
@@ -365,7 +365,7 @@ func (xl xlObjects) PutObjectPart(bucket, object, uploadID string, partID int, s
 
 	nsMutex.RLock(minioMetaBucket, uploadIDPath, opsID)
 	// Validates if upload ID exists.
-	if !xl.isUploadIDExists(bucket, object, uploadID) {
+	if !xl.isUploadIDExists(bucket, object, uploadID, true) {
 		nsMutex.RUnlock(minioMetaBucket, uploadIDPath, opsID)
 		return "", traceError(InvalidUploadID{UploadID: uploadID})
 	}
@@ -476,7 +476,7 @@ func (xl xlObjects) PutObjectPart(bucket, object, uploadID string, partID int, s
 	defer nsMutex.Unlock(minioMetaBucket, uploadIDPath, opsID)
 
 	// Validate again if upload ID still exists.
-	if !xl.isUploadIDExists(bucket, object, uploadID) {
+	if !xl.isUploadIDExists(bucket, object, uploadID, true) {
 		return "", traceError(InvalidUploadID{UploadID: uploadID})
 	}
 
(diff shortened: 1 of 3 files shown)


fwessels:
LGTM

cmd/xl-v1-multipart.go
```diff
@@ -624,7 +624,7 @@ func (xl xlObjects) ListObjectParts(bucket, object, uploadID string, partNumberM
 	nsMutex.Lock(minioMetaBucket, pathJoin(mpartMetaPrefix, bucket, object, uploadID), opsID)
 	defer nsMutex.Unlock(minioMetaBucket, pathJoin(mpartMetaPrefix, bucket, object, uploadID), opsID)
 
-	if !xl.isUploadIDExists(bucket, object, uploadID) {
+	if !xl.isUploadIDExists(bucket, object, uploadID, false) {
```
krishnasrinivas, line 627: if this is the only place where it passes `false` and hence an exception then we can see if can make the code simpler by always having valid-quorum for isUploadIDExists() so that we don't have two variants of with and without quorum.

If we do with-quorum-always:
Upside is simpler code (i.e just one variant of the function)
Downside is ListObjectParts is slightly slower.

So doing with-quorum-always looks a better choice more so if we assume that reuploads and hence ListObjectParts calls are rare. Another thing is ListObjectParts calls are rare compared to PutObjectPart where we do with-quorum-always, hence even if there is performance problem because of with-quorum-always check then it will hit PutObjectPart rather than ListObjectParts.

donatello: True, but @harshavardhana says generally we want listing to be fast. This PR was to discuss these things and finalize with @abperiasamy - so I am just waiting for more inputs.

harshavardhana:
This is deferred after discussion should be implemented differently i.e to heal inconsistencies during the server startup itself including `uploads.json`.

harshavardhana:
Closing this patch.

Issue #3160, opened 2016-11-02
fs: Explore the possibility of bringing in shared locking on distributed disk backends


harshavardhana:
## Expected Behavior
Evaluate the possibility of using distributed shared backend but Minio serving only as a S3 gateway. 

## Current Behavior
Currently we don't support it cleanly and can have undesired effects/data loss. 

## Possible Solution
A way to in-fact use - https://golang.org/src/syscall/flock.go and use this exclusivity . Distributed backends should support SETLK. 

## Steps to Reproduce (for bugs)
Not a bug its a feature

## Context
There are many deployment scenarios when users wish to use just Minio standalone FS server with shared backend.

harshavardhana:
A patch has been sent but there are pitfalls on Linux which we need to discuss further, moving this to pending-discussion.

harshavardhana:
To be fixed in - https://github.com/minio/minio/pull/3397

harshavardhana:
Fixed by #3397

PR #3218, opened 2016-11-09
lock: Implement shared locking used for shared backend semantics.

harshavardhana:
## Description
This patch implements two things.

- Implements lock package implementing filesystem locks using
  FcntlFlock() syscall, also implements a compatible version
  on windows as well using win32 APIs.

- Adds a new ENV which indicates minio server to enable shared
  backend feature, while holding the locks on `format.json`

So two competing locks from two different minio processes compete
at filesystem level on a shared backend while performing an i/o.

This code is an attempt to allow this feature.

## Motivation and Context
Refer #3160 

## How Has This Been Tested?
make test and local shared backend setup.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/fs-v1-multipart.go
+++ b/cmd/fs-v1-multipart.go
@@ -59,7 +59,7 @@ func (fs fsObjects) listMultipartUploads(bucket, prefix, keyMarker, uploadIDMark
 	var err error
 	var eof bool
 	if uploadIDMarker != "" {
-		keyMarkerLock := nsMutex.NewNSLock(minioMetaMultipartBucket,
+		keyMarkerLock := globalNSMutex.NewNSLock(minioMetaMultipartBucket,
 			pathJoin(bucket, keyMarker))
 		keyMarkerLock.RLock()
 		uploads, _, err = listMultipartUploadIDs(bucket, keyMarker, uploadIDMarker, maxUploads, fs.storage)
@@ -114,7 +114,7 @@ func (fs fsObjects) listMultipartUploads(bucket, prefix, keyMarker, uploadIDMark
 			var end bool
 			uploadIDMarker = ""
 
-			entryLock := nsMutex.NewNSLock(minioMetaMultipartBucket,
+			entryLock := globalNSMutex.NewNSLock(minioMetaMultipartBucket,
 				pathJoin(bucket, entry))
 			entryLock.RLock()
 			tmpUploads, end, err = listMultipartUploadIDs(bucket, entry, uploadIDMarker, maxUploads, fs.storage)
@@ -231,7 +231,7 @@ func (fs fsObjects) newMultipartUpload(bucket string, object string, meta map[st
 
 	// This lock needs to be held for any changes to the directory
 	// contents of ".minio.sys/multipart/object/"
-	objectMPartPathLock := nsMutex.NewNSLock(minioMetaMultipartBucket,
+	objectMPartPathLock := globalNSMutex.NewNSLock(minioMetaMultipartBucket,
 		pathJoin(bucket, object))
 	objectMPartPathLock.Lock()
 	defer objectMPartPathLock.Unlock()
@@ -304,7 +304,7 @@ func (fs fsObjects) PutObjectPart(bucket, object, uploadID string, partID int, s
 
 	uploadIDPath := path.Join(bucket, object, uploadID)
 
-	preUploadIDLock := nsMutex.NewNSLock(minioMetaMultipartBucket, uploadIDPath)
+	preUploadIDLock := globalNSMutex.NewNSLock(minioMetaMultipartBucket, uploadIDPath)
 	preUploadIDLock.RLock()
 	// Just check if the uploadID exists to avoid copy if it doesn't.
 	uploadIDExists := fs.isUploadIDExists(bucket, object, uploadID)
@@ -384,7 +384,7 @@ func (fs fsObjects) PutObjectPart(bucket, object, uploadID string, partID int, s
 	}
 
 	// Hold write lock as we are updating fs.json
-	postUploadIDLock := nsMutex.NewNSLock(minioMetaMultipartBucket, uploadIDPath)
+	postUploadIDLock := globalNSMutex.NewNSLock(minioMetaMultipartBucket, uploadIDPath)
 	postUploadIDLock.Lock()
 	defer postUploadIDLock.Unlock()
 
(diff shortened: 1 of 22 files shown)


harshavardhana:
Do not need to review this just yet, there is more work to be done.

pkg/flock/lock.go
```diff
@@ -0,0 +1,93 @@
+package flock
+
+import (
+	"os"
+	"sync"
+)
+
+// FLocker - lock/unlock implementation backed by.
```
krisis: backed by a shared FS which implements fcntl(2).

pkg/flock/lock_test.go
```diff
@@ -0,0 +1,339 @@
...
+	}
+	f.Close()
+	defer func() {
+		err = os.Remove(f.Name())
+		if err != nil {
+			t.Fatal(err)
+		}
+	}()
+
+	frw := NewFlock(f.Name())
```
krisis: Looks like the caller should create the file on which we hold lock. What should the caller do to hold a lock on a file yet to be created?

pkg/flock/lock.go
```diff
@@ -0,0 +1,93 @@
...
+
+type lockedFile struct {
+	*os.File
+}
+
+// Private file mode.
+var privateFileMode = os.FileMode(0600)
+
+// RWLocker - interface that any read-write locking library should implement.
+type RWLocker interface {
```
krisis: Why do we export this interface?

pkg/flock/lock_solaris.go
```diff
@@ -0,0 +1,64 @@
...
+		Len:    0,
+		Pid:    0,
+	}
+)
+
+func open(path string, flag int, perm os.FileMode) (*os.File, error) {
+	return os.OpenFile(path, flag, perm)
+}
+
+func lock(f *os.File) error {
```
krisis: Plain old F_SETLKW wouldn't conflict with a lock on a different fd on the same name (i.e, open file description) by the same minio server process.

krisis (commented):
Open file description (OFD) locks on Linux by fcntl(2) and BSD locks
by flock(2) don't provide the semantics that we
require for locking over shared filesystems like NFS or GlusterFS.
- NFS clients translate flock calls to POSIX byte-range locks (without
  OFD semantics) using the NLM protocol. In this situation, two
  requests holding 'conflicting' locks will be considered as an
  attempt to lock by the same (minio server) process and NOT conflict,
  as we require.
- GlusterFS clients are limited by the fact that FUSE kernel module
  [doesn't have OFD support yet](https://sourceforge.net/p/fuse/mailman/message/35301755/)

pkg/flock/lock_test.go
harshavardhana: We fail with ENOENT.

pkg/flock/lock.go
harshavardhana: Not needed.. actually.

harshavardhana:
Simple PUT operations are working fine, doing `mc mirror` in parallel is holding the server properly on an NFS mount 

```
mc mirror /usr/bin test1/testbucket/ & mc mirror /usr/bin test2/testbucket/
```

There is a problem in multipart seems to be that tempobj which is being appended in a go-routine has `uploadID` of the multipart session associated with it, it is not a unique file name. This will cause problems regardless since we are doing an AppendFile() we cannot differentiate. 

Will be making changes for multipart to use a separate temp UUID and remember it.

cmd/fs-v1-multipart.go
```diff
@@ -417,12 +417,11 @@ func (fs fsObjects) PutObjectPart(bucket, object, uploadID string, partID int, s
 	}
 
 	go func() {
-		// Append the part in background.
-		errCh := fs.bgAppend.append(fs.storage, bucket, object, uploadID, fsMeta)
 		// Also receive the error so that the appendParts go-routine does not block on send.
 		// But the error received is ignored as fs.PutObjectPart() would have already
 		// returned success to the client.
-		<-errCh
+		// Append the part in background.
```
krisis: nit: The comment needs to change to accomadate the code movement.

cmd/namespace-lock.go
```diff
@@ -137,8 +133,14 @@ func (n *nsLockMap) lock(volume, path string, lockSource, opsID string, readLock
 
 	// Locking here can block.
 	if readLock {
+		if globalFSMutex != nil {
```
krisis, line 137: The namespace lock should be taken before the shared filesystem lock is attempted. This is the only we can prevent races between two request handlers attempting a lock on the same file from the same minio server process.

pkg/lock/lock_nix.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+)
+
+func open(path string, flag int, perm os.FileMode) (*os.File, error) {
+	return os.OpenFile(path, flag, perm)
+}
+
+func lock(f *os.File) error {
+	lock := wrlck
+	lock.Pid = int32(os.Getpid())
```
krisis: The pid of minio server process will not change. You can initialize the Pid field at the time of initializing ``wrlck`` and ``rdlck`` variables.

pkg/lock/lock_nix.go
```diff
@@ -0,0 +1,70 @@
...
+func rlock(f *os.File) error {
+	lock := rdlck
+	lock.Pid = int32(os.Getpid())
+	return syscall.FcntlFlock(f.Fd(), syscall.F_SETLKW, &lock)
+}
+
+func runlock(f *os.File) error {
+	lock := rdlck
+	lock.Pid = int32(os.Getpid())
+	if err := syscall.FcntlFlock(f.Fd(), syscall.F_GETLK, &lock); err != nil {
```
krisis: What is this GETLK for?

pkg/lock/lock_nix.go
```diff
@@ -0,0 +1,70 @@
...
+	"syscall"
+)
+
+var (
+	wrlck = syscall.Flock_t{
+		Type:   syscall.F_WRLCK,
+		Whence: 0,
+		Start:  0,
+		Len:    0,
+		Pid:    0,
```
krisis: The Pid can be initialized here, see comment below.

pkg/lock/lock_nix.go
harshavardhana: I think the sending an runlock on an runlocked file returns error?

cmd/fs-v1-multipart.go
harshavardhana: the code shouldn't be moved was a testing change reverting.

cmd/namespace-lock.go
harshavardhana: Good catch, let me change it.

pkg/lock/lock_nix.go
harshavardhana: done

pkg/lock/lock_nix.go
harshavardhana: done

harshavardhana:
We decided to do a completely different style of approach here by removing the necessity for `posix.go` at FS layer.  and let FS layer directly talk to backend.  There is a reason behind that - it provides a way for us to hold locks transparently for each i/o - current semantics of `fcntl` will work for us. 

It also solves the problem of running multiple minio servers on same node with same backend but using different access keys. Will explain more on this will call.

harshavardhana:
Closing, superseded by #3397

PR #3221, opened 2016-11-09
rpc: Our rpcClient should make an attempt to reconnect.

### LINKED ISSUE #3205: [distributed] replace disk, heal disk does not work the first time
* start 2:2 dXL
* bring down 4th server
* replace disk of 4th server (rm -rf and mkdir)
* bring up 4th server
* heal the disk:
```
krishna@escape:~$ minio control heal http://localhost:9001/
krishna@escape:~$ ls /export/disk4/ -a                             <---- disk not healed
.  ..
krishna@escape:~$ minio control heal http://localhost:9001/
krishna@escape:~$ ls /export/disk4/ -a
.  ..  .minio.sys                                                                    <- disk healed on 2nd try
krishna@escape:~$ 
```

i.e `minio control heal http://localhost:9001/` works only on the 2nd time.

## Expected Behavior

`minio control heal http://localhost:9001/` should work the first time.

## Current Behavior

`minio control heal http://localhost:9001/` works only on the 2nd time.

### LINKED ISSUE #3206: [distributed] replace disk, bucket heal does not happen in one situation
* start 2:2 dXL
* bring down 4th server
* replace disk of 4th server (rm -rf and mkdir)
* bring up 4th server
* heal the disk:
```
krishna@escape:~$ minio control heal http://localhost:9004/
krishna@escape:~$ ls /export/disk4/ -a
.  ..  .minio.sys
krishna@escape:~$ 
```
* heal the bucket:
```
krishna@escape:~$ minio control heal http://localhost:9004/test

krishna@escape:~$ ls /export/disk4/ -a
.  ..  .minio.sys
krishna@escape:~$
```

Bucket is not healed. But it heals if healing is pointed towards localhost:9001/9002/9003

harshavardhana:
## Description
rpcClient should attempt a reconnect if the call fails
with 'rpc.ErrShutdown' this is needed since at times when
the servers are taken down and brought back up.

The hijacked connection from net.Dial is usually closed.

So upon first attempt rpcClient might falsely indicate that
disk to be down, to avoid this state make another dial attempt
to really fail.

## Motivation and Context
Fixes #3206
Fixes #3205

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:

### DIFF
--- a/cmd/format-config-v1.go
+++ b/cmd/format-config-v1.go
@@ -246,7 +246,7 @@ func genericFormatCheck(formatConfigs []*formatConfigV1, sErrs []error) (err err
 	}
 
 	// Calculate read quorum.
-	readQuorum := len(formatConfigs)/2 + 1
+	readQuorum := len(formatConfigs) / 2
 
 	// Validate the err count under tolerant limit.
 	if errCount > len(formatConfigs)-readQuorum {
--- a/cmd/net-rpc-client.go
+++ b/cmd/net-rpc-client.go
@@ -142,17 +142,26 @@ func (rpcClient *RPCClient) Call(serviceMethod string, args interface{}, reply i
 	// rpc.Client for a subsequent reconnect.
 	err := rpcLocalStack.Call(serviceMethod, args, reply)
 	if err != nil {
-		if err.Error() == rpc.ErrShutdown.Error() {
-			// Reset rpcClient.rpc to nil to trigger a reconnect in future
-			// and close the underlying connection.
-			rpcClient.clearRPCClient()
+		// Any errors other than rpc.ErrShutdown just return quickly.
+		if err != rpc.ErrShutdown {
+			return err
+		} // else rpc.ErrShutdown returned by rpc.Call
 
-			// Close the underlying connection.
-			rpcLocalStack.Close()
+		// Reset the underlying rpc connection before
+		// moving to reconnect.
+		rpcClient.clearRPCClient()
 
-			// Set rpc error as rpc.ErrShutdown type.
-			err = rpc.ErrShutdown
+		// Close the underlying connection before reconnect.
+		rpcLocalStack.Close()
+
+		// Try once more to re-connect.
+		rpcLocalStack, err = rpcClient.dialRPCClient()
+		if err != nil {
+			return err
 		}
+
+		// Attempt the rpc.Call once again, upon any error now just give up.
+		err = rpcLocalStack.Call(serviceMethod, args, reply)
 	}
 	return err
 }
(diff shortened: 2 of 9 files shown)


cmd/format-config-v1.go
```diff
@@ -230,7 +230,8 @@ func genericFormatCheck(formatConfigs []*formatConfigV1, sErrs []error) (err err
 	)
 
 	// Through all errors calculate the actual errors.
-	for _, lErr := range sErrs {
+	for index, lErr := range sErrs {
+		fmt.Println(lErr, index)
```
vadmeste: This is not needed.

cmd/net-rpc-client.go
```diff
@@ -150,8 +150,18 @@ func (rpcClient *RPCClient) Call(serviceMethod string, args interface{}, reply i
...
+			rpcLocalStack, err = rpcClient.dialRPCClient()
+			if err != nil {
+				return err
+			}
+
+			// Attempt the service call oncne again, upon any error now just give up.
+			err = rpcLocalStack.Call(serviceMethod, args, reply)
+			if err != nil && err.Error() == rpc.ErrShutdown.Error() {
+				// Set rpc error as rpc.ErrShutdown type.
+				err = rpc.ErrShutdown
```
vadmeste: Isn't it useless to set err here ? because we enter here only when err is already rpc.ErrShutdown

harshavardhana: no we need to set it back to a proper type so that the type is not lost. otherwise rpc.Call() converts this into rpc.ServerError() at times not all the time.

cmd/format-config-v1.go
harshavardhana: Oops yeah.

cmd/net-rpc-client.go
harshavardhana: Let me try again locally and see if the type is not lost in these situations if the type is not lost then we can keep it rpc.ErrShudown.

harshavardhana: what happened @donatello don't see your comment anymore?

harshavardhana: Just tried the type is lost. it doesn't work.

cmd/net-rpc-client.go
```diff
@@ -141,20 +141,26 @@ func (rpcClient *RPCClient) Call(serviceMethod string, args interface{}, reply i
...
 	err := rpcLocalStack.Call(serviceMethod, args, reply)
-	if err != nil {
-		if err.Error() == rpc.ErrShutdown.Error() {
-			// Reset rpcClient.rpc to nil to trigger a reconnect in future
-			// and close the underlying connection.
-			rpcClient.clearRPCClient()
+	// Any errors other than rpc.ErrShutdown just return quickly.
+	if err != nil && err != rpc.ErrShutdown {
+		return err
+	} // else rpc.ErrShutdown returned by rpc.Call
```
donatello: doesnt seem to be right. It can be that `err == nil` after the if block..

harshavardhana: yes fixing this.

donatello (approved):

PR #3222, opened 2016-11-09
v4/presign: Fix presign requests when there are more signed headers.

### LINKED ISSUE #3217: Single-part upload using pre-signed url failed with response status 400
Attempt to upload a small file to a pre-signed failed with http response status 400.

## Expected Behavior

Uploading a small file (<5 MB, so no multi-part upload was used) to a bucket using a pre-signed url should succeed.  Because it works with aws S3.

## Current Behavior

It doesn't work with minio server.

## Possible Solution

For single-part upload, there is no upload ID.  So minio server should just save the file into the specified bucket. 

1. Generate a request using s3 api S3.PutObjectRequest
2. Get a pre-signed url using s3 api Request.Presign
3. Send the file to the pre-signed url using http PUT
4. Now you will get http response with status 400

## Context

For small files, we don't use multi-part upload, which is intended for big files.

## Your Environment

* Version used:
We build minio server from source code downloaded on 11/09/2016 using go get -u github.com/minio/minio

* Server type and version:
minio server, no SSL, 

Ubuntu 16.04

* Link to your project:

harshavardhana:
## Description
This fix removes a wrong logic which fails for requests which
have more signed headers in a presign request.

## Motivation and Context
Fixes #3217

## How Has This Been Tested?
Manually and using aws-sdk-go. 

## Types of changes

## Checklist:

### DIFF
--- a/cmd/signature-v4-utils.go
+++ b/cmd/signature-v4-utils.go
@@ -33,10 +33,17 @@ const unsignedPayload = "UNSIGNED-PAYLOAD"
 
 // http Header "x-amz-content-sha256" == "UNSIGNED-PAYLOAD" indicates that the
 // client did not calculate sha256 of the payload. Hence we skip calculating sha256.
-// We also skip calculating sha256 for presigned requests without "x-amz-content-sha256" header.
+// We also skip calculating sha256 for presigned requests without "x-amz-content-sha256"
+// query header.
 func skipContentSha256Cksum(r *http.Request) bool {
-	contentSha256 := r.Header.Get("X-Amz-Content-Sha256")
-	return isRequestUnsignedPayload(r) || (isRequestPresignedSignatureV4(r) && contentSha256 == "")
+	queryContentSha256 := r.URL.Query().Get("X-Amz-Content-Sha256")
+	isRequestPresignedUnsignedPayload := func(r *http.Request) bool {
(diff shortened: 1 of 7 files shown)


fwessels (commented):
LGTM

krishnasrinivas (commented):
Looks good but needs a test case.

harshavardhana:
> Looks good but needs a test case.

Added a test case @krishnasrinivas

cmd/signature-v4-utils.go
```diff
@@ -115,11 +121,14 @@ func extractSignedHeaders(signedHeaders []string, reqHeaders http.Header) (http.
 	if errCode != ErrNone {
 		return nil, errCode
 	}
+	// With only one signed header, we don't need to look for any other headers.
+	if len(signedHeaders) == 1 {
+		return nil, ErrNone
+	}
```
krishnasrinivas, line 122: actually findHost function can be removed and this above change `if len(signedHeaders) == 1 {` is not needed.

i.e here:

```
if !contains(signedHeaders, "host") {
    // host header signing is mandatory.
    return ErrUnsignedHeaders
}
```

can be done

krishnasrinivas (approved):
Added one comment about removal of findHost though.

PR #3228, opened 2016-11-10
fix: Ignore object not found error in RemoveObject() in web-handler.

### LINKED ISSUE #3181: [browser] Removing object throws "Object not found" for previously deleted object.
## Expected Behavior
It should not throw an error

## Current Behavior
"Object not found" error is shown and "Are you sure want to delete?" modal doesn't disappear.

## Possible Solution
It could be easy to fix at web handler side to not throw this error.

1. Open minio browser for a bucket.
2. Remove an existing object using `mc rm`
3. Now try to remove the same object in the browser.

## Minio version:
`Commit-ID: bf3c93a8cc264a2add6d3074b78d3a7d8954a523`

balamurugana:
Fixes #3181

### DIFF
--- a/cmd/web-handlers.go
+++ b/cmd/web-handlers.go
@@ -271,7 +271,15 @@ func (web *webAPIHandlers) RemoveObject(r *http.Request, args *RemoveObjectArgs,
 		return &json2.Error{Message: errAuthentication.Error()}
 	}
 	if err := objectAPI.DeleteObject(args.BucketName, args.ObjectName); err != nil {
-		return &json2.Error{Message: err.Error()}
+		objErr := errorCause(err)
+		switch objErr.(type) {
+		case ObjectNotFound:
+			// Ignore object not found error.
+			reply.UIVersion = miniobrowser.UIVersion
+			return nil
+		default:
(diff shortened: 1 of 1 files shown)


cmd/web-handlers.go
```diff
@@ -271,7 +272,9 @@ func (web *webAPIHandlers) RemoveObject(r *http.Request, args *RemoveObjectArgs,
 		return &json2.Error{Message: errAuthentication.Error()}
 	}
 	if err := objectAPI.DeleteObject(args.BucketName, args.ObjectName); err != nil {
-		return &json2.Error{Message: err.Error()}
+		if reflect.TypeOf(errorCause(err)) != reflect.TypeOf(ObjectNotFound{}) {
```
harshavardhana: We should use 

```
switch err.(type) {
```

Instead than importing "reflect" package.

krishnasrinivas:
looks good apart from @harshavardhana 's comment

cmd/web-handlers.go
balamurugana: Done

PR #3252, opened 2016-11-13
heal: Add healing support for bucket metadata files.

### LINKED ISSUE #2980: [distributed]: Enable quick heal for  'policy.json', 'notification.xml' , 'listeners.json'
Basically all bucket metadata files as well.

### LINKED ISSUE #3237: [distributed] HealObject() tries to heal when one of the servers is down
## Expected Behavior
HealObject() should return immediately as there is nothing to heal.

## Current Behavior
ListHealObject() tries to heal the object thinking that the object in the down server is outdated.

1. start 3:3 dxl
3. Upload a `bucket/object`
2. Bring down 6th server
3. Exec heal command on the bucket: (this takes up CPU healing the object)

```
krishna@escape:~$ MINIO_ACCESS_KEY=SXO8VW2OFKKP2OG7AC85 MINIO_SECRET_KEY=CKWSSgrUgvfUMTaNBkB63exet4WW+uNhQvi91Bc3 minio control heal http://localhost:9005/testbucket
[0] ◐ SUCCESS  5g                                                                                                    
krishna@escape:~$ 
```

It actually tried to heal the object and said success.

Fix:
ListObjectsHeal() should not have returned the object
HealObject() should not try to heal the object (xlShouldHeal should return false)

harshavardhana:
## Description
This change implements healing of bucket metadata under `policy.json, notification.xml and listener.json`

## Motivation and Context
Fixes #2980 

## Types of changes

## Checklist:

### DIFF
--- a/cmd/xl-v1-healing.go
+++ b/cmd/xl-v1-healing.go
@@ -16,9 +16,47 @@
 
 package cmd
 
-import "sync"
+import (
+	"fmt"
+	"path"
+	"sync"
+)
 
-// Heals a bucket if it doesn't exist on one of the disks.
+// healFormatXL - heals missing `format.json` on freshly or corrupted
+// disks (missing format.json but does have erasure coded data in it).
+func healFormatXL(storageDisks []StorageAPI) (err error) {
+	// Attempt to load all `format.json`.
+	formatConfigs, sErrs := loadAllFormats(storageDisks)
+
+	// Generic format check validates
+	// if (no quorum) return error
+	// if (disks not recognized) // Always error.
+	if err = genericFormatCheck(formatConfigs, sErrs); err != nil {
+		return err
+	}
+
+	// Handles different cases properly.
+	switch reduceFormatErrs(sErrs, len(storageDisks)) {
+	case errCorruptedFormat:
+		if err = healFormatXLCorruptedDisks(storageDisks); err != nil {
+			return fmt.Errorf("Unable to repair corrupted format, %s", err)
+		}
+	case errSomeDiskUnformatted:
+		// All drives online but some report missing format.json.
+		if err = healFormatXLFreshDisks(storageDisks); err != nil {
+			// There was an unexpected unrecoverable error during healing.
+			return fmt.Errorf("Unable to heal backend %s", err)
+		}
+	case errSomeDiskOffline:
+		// FIXME: in future.
+		return fmt.Errorf("Unable to initialize format %s and %s", errSomeDiskOffline, errSomeDiskUnformatted)
+	}
+	return nil
+}
+
+// Heals a bucket if it doesn't exist on one of the disks, additionally
+// also heals the missing entries for bucket metadata files
+// `policy.json, notification.xml, listeners.json`.
 func (xl xlObjects) HealBucket(bucket string) error {
 	// Verify if bucket is valid.
 	if !IsValidBucketName(bucket) {
@@ -30,8 +68,17 @@ func (xl xlObjects) HealBucket(bucket string) error {
 		return traceError(BucketNotFound{Bucket: bucket})
 	}
 
-	// Heal bucket - create buckets on disks where it does not exist.
+	// Heal bucket.
+	if err := healBucket(xl.storageDisks, bucket, xl.writeQuorum); err != nil {
+		return err
+	}
 
+	// Proceed to heal bucket metadata.
+	return healBucketMetadata(xl.storageDisks, bucket)
+}
+
+func healBucket(storageDisks []StorageAPI, bucket string, writeQuorum int) error {
+	// Heal bucket - create buckets on disks where it does not exist.
 	bucketLock := nsMutex.NewNSLock(bucket, "")
 	bucketLock.Lock()
 	defer bucketLock.Unlock()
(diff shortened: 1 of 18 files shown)


harshavardhana:
Tests are pending working on it.. but this is ready for review.

cmd/xl-v1-healing.go
```diff
@@ -82,29 +86,56 @@ func (xl xlObjects) HealBucket(bucket string) error {
...
-	if !IsValidBucketName(bucket) {
-		return traceError(BucketNameInvalid{Bucket: bucket})
+// Heals all the metadata associated for a given bucket.
+func healBucketMetadata(storageDisks []StorageAPI, bucket string) error {
+	policyPath := pathJoin(bucketConfigPrefix, bucket, policyJSON)
+	// Lock the `policy.json` before healing.
+	policyLock := nsMutex.NewNSLock(minioMetaBucket, policyPath)
+	policyLock.RLock()
+	defer policyLock.RUnlock()
+
```
donatello, line 148: Since this mutates the file on on some disks, shouldn't this take a write lock instead of read?

donatello: Also, it appears that this keeps the lock longer than necessary. It may be better to write this section with this scheme:

``` go
f := func(){
    policyLock := ...
    policyLock.RLock() // or write lock - not sure
    defer policyLock.RUnlock()
    return healObject(...)
}
if err := f(); err != nil {
    if !isErrObjectNotFound(err) {
        return err
    }
}
```

cmd/xl-v1-healing.go
```diff
@@ -82,29 +86,56 @@ func (xl xlObjects) HealBucket(bucket string) error {
...
 
-	// Verify if object is valid.
-	if !IsValidObjectName(object) {
-		return traceError(ObjectNameInvalid{Bucket: bucket, Object: object})
+	nConfigPath := path.Join(bucketConfigPrefix, bucket, bucketNotificationConfig)
+	// Lock the `notificiation.xml` before healing.
+	ncLock := nsMutex.NewNSLock(minioMetaBucket, nConfigPath)
+	ncLock.RLock()
+	defer ncLock.RUnlock()
+
```
donatello: Same points as for policy above.

cmd/xl-v1-healing.go
```diff
@@ -82,29 +86,56 @@ func (xl xlObjects) HealBucket(bucket string) error {
...
-	// Lock the object before healing.
-	objectLock := nsMutex.NewNSLock(bucket, object)
-	objectLock.RLock()
-	defer objectLock.RUnlock()
+	// Lock the `listeners.json` before healing.
+	lConfigPath := path.Join(bucketConfigPrefix, bucket, bucketListenerConfig)
+	listenLock := nsMutex.NewNSLock(minioMetaBucket, lConfigPath)
+	listenLock.RLock()
+	defer listenLock.RUnlock()
+
```
donatello, line 165: Same points as for policy above.

cmd/xl-v1-healing.go
harshavardhana: > Since this mutates the file on on some disks, shouldn't this take a write lock instead of read?

The reason to not hold Write lock is that it would block the Read operations, also since we are only writing to missing disks. Additional advantage here is that Write() outside of this call at this path will still block because of our Read lock. So all in all it works without much trouble. 

This is how object healing is implemented and it is a correct behavior.

cmd/xl-v1-healing.go
harshavardhana: Same as above.

cmd/xl-v1-healing.go
harshavardhana: Same as above.

cmd/xl-v1-healing.go
fwessels: What happens if a heal on the same object would happen simultaneously on more than one nodes? Any chance of corrupted data?

harshavardhana: Whoever completes last gets to create their missing blocks which are always the same given the state of the object.

cmd/xl-v1-healing.go
```diff
@@ -82,29 +86,47 @@ func (xl xlObjects) HealBucket(bucket string) error {
...
+// Heals all the metadata associated for a given bucket, this function
+// heals `policy.json`, `notification.xml` and `listeners.json`.
+func healBucketMetadata(storageDisks []StorageAPI, bucket string) error {
+	healBucketMetaFn := func(metaPath string) error {
+		metaLock := nsMutex.NewNSLock(minioMetaBucket, metaPath)
+		metaLock.RLock()
+		defer metaLock.RUnlock()
+		// Heals the metaPath.
+		if err := healObject(storageDisks, minioMetaBucket, metaPath); err != nil {
+			if !isErrObjectNotFound(err) {
```
fwessels: Can collapse two if statements into a single one

fwessels (commented):
LGTM

cmd/xl-v1-healing.go
harshavardhana: Yes will do

cmd/xl-v1-healing.go
donatello: > The reason to not hold Write lock is that it would block the Read operations, also since we are only writing to missing disks. Additional advantage here is that Write() outside of this call at this path will still block because of our Read lock. So all in all it works without much trouble. 

Ok. The code is however holding onto the lock a bit longer than necessary - e.g. policy path lock is held during notification config heal, and notification path lock is held during listener config heal.

harshavardhana: I already addressed that @donatello - check the latest patch.

harshavardhana:
This is ready for review, this patch only adds quickHeal part.

harshavardhana:
Any reviews here @minio/core ?

donatello:
I will do shortly.

donatello (approved):

PR #3273, opened 2016-11-17
Cleanup and fixes

balamurugana:
## Description
* newRequestID() (previously generateUploadID()) returns string than byte array.
* Remove unclear comments and added appropriate comments.
* SHA-256, MD5 Hash functions return Hex/Base64 encoded string than byte array.
* Remove duplicate MD5 hasher functions.
* Rename listObjectsValidateArgs() into validateListObjectsArgs()
* Remove repeated auth check code in all bucket request handlers.
* Remove abbreviated names in bucket-metadata
* Avoid nested if in bucketPolicyMatchStatement()
* Use ioutil.ReadFile() instead of os.Open() and ioutil.ReadAll()
* Set crossDomainXML as constant.

## How Has This Been Tested?
`make test`

## Types of changes

## Checklist:

### DIFF
--- a/cmd/auth-handler.go
+++ b/cmd/auth-handler.go
@@ -18,10 +18,6 @@ package cmd
 
 import (
 	"bytes"
-	"crypto/md5"
-	"crypto/sha256"
-	"encoding/base64"
-	"encoding/hex"
 	"io/ioutil"
 	"net/http"
 	"strings"
@@ -108,18 +104,32 @@ func getRequestAuthType(r *http.Request) authType {
 	return authTypeUnknown
 }
 
-// sum256 calculate sha256 sum for an input byte array
-func sum256(data []byte) []byte {
-	hash := sha256.New()
-	hash.Write(data)
-	return hash.Sum(nil)
-}
+func checkRequestAuthType(r *http.Request, bucket, policyAction, region string) APIErrorCode {
+	reqAuthType := getRequestAuthType(r)
 
-// sumMD5 calculate md5 sum for an input byte array
-func sumMD5(data []byte) []byte {
-	hash := md5.New()
-	hash.Write(data)
-	return hash.Sum(nil)
+	switch reqAuthType {
+	case authTypePresignedV2, authTypeSignedV2:
+		// Signature V2 validation.
+		s3Error := isReqAuthenticatedV2(r)
+		if s3Error != ErrNone {
+			errorIf(errSignatureMismatch, dumpRequest(r))
+		}
+		return s3Error
+	case authTypeSigned, authTypePresigned:
+		s3Error := isReqAuthenticated(r, region)
+		if s3Error != ErrNone {
+			errorIf(errSignatureMismatch, dumpRequest(r))
+		}
+		return s3Error
+	}
+
+	if reqAuthType == authTypeAnonymous && policyAction != "" {
+		// http://docs.aws.amazon.com/AmazonS3/latest/dev/using-with-s3-actions.html
+		return enforceBucketPolicy(bucket, policyAction, r.URL)
+	}
+
+	// By default return ErrAccessDenied
+	return ErrAccessDenied
 }
 
 // Verify if request has valid AWS Signature Version '2'.
@@ -157,7 +167,7 @@ func isReqAuthenticated(r *http.Request, region string) (s3Error APIErrorCode) {
 	}
 	// Verify Content-Md5, if payload is set.
 	if r.Header.Get("Content-Md5") != "" {
-		if r.Header.Get("Content-Md5") != base64.StdEncoding.EncodeToString(sumMD5(payload)) {
+		if r.Header.Get("Content-Md5") != getMD5HashBase64(payload) {
 			return ErrBadDigest
 		}
 	}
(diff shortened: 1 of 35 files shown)


harshavardhana:
@balamurugana let's follow the PR template, we shouldn't delete the description template. We should add some info that way gist is captured for each PR.   Take a look at other PRs, are you sending the PR from command line? 

Also this commit message can be better.

cmd/bucket-handlers-listobjects.go
```diff
@@ -52,6 +52,41 @@ func validateListObjectsArgs(prefix, marker, delimiter string, maxKeys int) APIE
...
+		s3Error := enforceBucketPolicy(bucket, policyAction, r.URL)
+		if s3Error != ErrNone {
+			writeErrorResponse(w, r, s3Error, r.URL.Path)
+		}
+		return s3Error
+	}
+
+	// By default return ErrAccessDenied
+	writeErrorResponse(w, r, ErrAccessDenied, r.URL.Path)
+	return ErrAccessDenied
```
harshavardhana: If you already wrote back to client, we don't need to return anything here.. otherwise we can return and write after the check with ErrNone.

cmd/bucket-handlers-listobjects.go
```diff
@@ -52,6 +52,41 @@ func validateListObjectsArgs(prefix, marker, delimiter string, maxKeys int) APIE
...
+			writeErrorResponse(w, r, s3Error, r.URL.Path)
+		}
+		return s3Error
+	}
+
+	if reqAuthType == authTypeAnonymous && policyAction != "" {
+		// http://docs.aws.amazon.com/AmazonS3/latest/dev/using-with-s3-actions.html
+		s3Error := enforceBucketPolicy(bucket, policyAction, r.URL)
+		if s3Error != ErrNone {
+			writeErrorResponse(w, r, s3Error, r.URL.Path)
```
harshavardhana: Same as below comment.

cmd/bucket-handlers-listobjects.go
```diff
@@ -52,6 +52,41 @@ func validateListObjectsArgs(prefix, marker, delimiter string, maxKeys int) APIE
...
+		if s3Error != ErrNone {
+			errorIf(errSignatureMismatch, dumpRequest(r))
+			writeErrorResponse(w, r, s3Error, r.URL.Path)
+		}
+		return s3Error
+	case authTypeSigned, authTypePresigned:
+		s3Error := isReqAuthenticated(r, region)
+		if s3Error != ErrNone {
+			errorIf(errSignatureMismatch, dumpRequest(r))
+			writeErrorResponse(w, r, s3Error, r.URL.Path)
```
harshavardhana: Same as below comment.

cmd/bucket-handlers-listobjects.go
```diff
@@ -52,6 +52,41 @@ func validateListObjectsArgs(prefix, marker, delimiter string, maxKeys int) APIE
...
+func checkRequestAuthType(w http.ResponseWriter, r *http.Request, bucket, policyAction, region string) APIErrorCode {
+	reqAuthType := getRequestAuthType(r)
+
+	switch reqAuthType {
+	case authTypePresignedV2, authTypeSignedV2:
+		// Signature V2 validation.
+		s3Error := isReqAuthenticatedV2(r)
+		if s3Error != ErrNone {
+			errorIf(errSignatureMismatch, dumpRequest(r))
+			writeErrorResponse(w, r, s3Error, r.URL.Path)
```
harshavardhana: Same as below comment.

cmd/bucket-handlers-listobjects.go
balamurugana: Yes. The function returning boolean is good enough. As I am still reading other handlers, I made it as interim. I will fix it in upcoming update of this PR.

harshavardhana: If you return APIErrorCode it is easy to test this function. Rather writing to client internally.

cmd/bucket-policy-handlers.go
```diff
@@ -50,17 +50,10 @@ func bucketPolicyEvalStatements(action string, resource string, conditions map[s
...
-		// Verify if resource matches.
-		if bucketPolicyResourceMatch(resource, statement) {
-			// Verify if condition matches.
-			if bucketPolicyConditionMatch(conditions, statement) {
-				return true
-			}
-		}
-	}
-	return false
+	// Verify if action, resource and condition in given statement.
```
harshavardhana: `// Verify if action, resource and condition match in a given statement.`

cmd/hasher.go
```diff
@@ -0,0 +1,48 @@
...
+
+// getSHA256Hash returns SHA-256 hash of given data.
+func getSHA256Hash(data []byte) string {
+	hash := sha256.New()
+	hash.Write(data)
+	return hex.EncodeToString(hash.Sum(nil))
+}
+
+// getMD5 returns MD5 data of given data.
+func getMD5(data []byte) []byte {
```
donatello: This function is actually returning the binary "hash" value for MD5 hash function. It is better named `getMD5Raw` - as this is the raw binary output of the hash function. The others are just encodings (see below).

cmd/hasher.go
```diff
@@ -0,0 +1,48 @@
...
+	hash := md5.New()
+	hash.Write(data)
+	return hash.Sum(nil)
+}
+
+// getMD5Hash returns MD5 hash of given data.
+func getMD5Hash(data []byte) string {
```
donatello, line 41: This is the hex encoding of the hash. It is better named as `getMD5Hex`.

cmd/hasher.go
```diff
@@ -0,0 +1,48 @@
...
+	return hex.EncodeToString(getMD5(data))
+}
+
+// getMD5HashBase64 returns MD5 hash in base64 encoding of given data.
+func getMD5HashBase64(data []byte) string {
```
donatello, line 46: `getMD5Base64` - is a better name for this function. If all the renames are done, we have 3 coherent MD5 functions - `getMD5Raw`, `getMD5Hex` and `getMD5Base64`, instead of `getMD5`, `getMD5Hash` and `getMD5HashBase64` which are all similar and confusing names.

cmd/hasher.go
```diff
@@ -0,0 +1,48 @@
...
+
+import (
+	"crypto/md5"
+	"crypto/sha256"
+	"encoding/base64"
+	"encoding/hex"
+)
+
+// getSHA256Hash returns SHA-256 hash of given data.
+func getSHA256Hash(data []byte) string {
```
donatello, line 27: This is better names as `getSHA256Hex`

cmd/auth-handler.go
```diff
@@ -104,6 +104,34 @@ func getRequestAuthType(r *http.Request) authType {
 	return authTypeUnknown
 }
 
+func checkRequestAuthType(r *http.Request, bucket, policyAction, region string) APIErrorCode {
```
donatello, line 107: Please add function comment - why `policyAction` is needed for a function named `checkRequestAuthType`? This is doing both authentication and authorisation (enforcing bucket policy for anonymous request).

donatello (commented):
Love the clean up commits removing tons of code!

LGTM - I just suggested some better names for some functions.

harshavardhana:
ping @balamurugana

PR #3282, opened 2016-11-18
multipart-fs: Append multiparts parts in a Go routine in background.

krishnasrinivas:
## Description

Do the background append process in a go-routine. Simplifies the code flow.

## Motivation and Context
Simplify the appendParts code flow. Also fix a race where completeMultipart overtakes appendPart of the last part. Explained in https://github.com/minio/minio/pull/3139 review comments.

## How Has This Been Tested?
manual testing and `make test`

## Types of changes

## Checklist:

### DIFF
--- a/cmd/fs-v1-append-parts.go
+++ b/cmd/fs-v1-append-parts.go
@@ -1,0 +1,170 @@
+package cmd
+
+import (
+	"errors"
+	"io"
+	"reflect"
+	"sync"
+	"time"
+)
+
+// Error sent by appendRoutine when there are holes in parts.
+// For ex. let's say client uploads part-2 before part-1 in which case we
+// can not append and have to wait till part-1 is uploaded. Hence we return
+// this error. Currently this error is not used in the caller.
+var errPartsMissing = errors.New("required parts missing")
+
+// Error sent when appendRoutine has waited long enough and timedout.
+var errAppendRoutineTimeout = errors.New("timeout")
+
+// Timeout value for the appendRoutine go-routine.
+var appendRoutineTimeout = 24 * 60 * 60 * time.Second
+
+// Holds a map of uploadID->appendRoutine
+type appendPartFS struct {
+	appendRoutineMap map[string]appendRoutineInfo
+	mu               sync.Mutex
+}
+
+// Input to the appendRoutine
+type appendRoutineInput struct {
+	meta  fsMetaV1   // list of parts that need to be appended
+	errCh chan error // error sent by appendRoutine
+}
+
+// Identifies an appendRoutine.
+type appendRoutineInfo struct {
+	inputCh   chan appendRoutineInput
+	timeoutCh chan struct{} // closed by appendRoutine when it timesout
+	endCh     chan struct{} // closed after complete/abort of upload to end the appendRoutine
+}
+
+// Called after a part is uploaded so that it can be appended in the background.
+func (m *appendPartFS) processPart(disk StorageAPI, bucket, object, uploadID string, meta fsMetaV1) chan error {
+	m.mu.Lock()
+	info, ok := m.appendRoutineMap[uploadID]
+	if !ok {
+		// Corresponding appendRoutine was not found, create a new one. Would happen when the first
+		// part of a multipart upload is uploaded.
+		inputCh := make(chan appendRoutineInput)
+		timeoutCh := make(chan struct{})
+		endCh := make(chan struct{})
+
+		info = appendRoutineInfo{inputCh, timeoutCh, endCh}
+		m.appendRoutineMap[uploadID] = info
+
+		go m.appendRoutine(disk, bucket, object, uploadID, info)
+	}
+	m.mu.Unlock()
+	errCh := make(chan error)
+	go func() {
+		// send input in a goroutine as send on the inputCh can block if appendRoutine
+		// is busy appending a part.
+		select {
+		case <-info.timeoutCh:
+			// This is to handle a rare race condition where we found info in m.appendRoutineMap
+			// but soon after that appendRoutine timedouted out.
+			errCh <- errAppendRoutineTimeout
+		case info.inputCh <- appendRoutineInput{meta, errCh}:
+		}
+	}()
+	return errCh
+}
+
+// Called after complete-multipart-upload or abort-multipart-upload so that the appendRoutine is not left dangling.
+func (m *appendPartFS) endAppendRoutine(uploadID string) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	info, ok := m.appendRoutineMap[uploadID]
+	if !ok {
+		return
+	}
+	delete(m.appendRoutineMap, uploadID)
+	close(info.endCh)
+}
+
+// The go-routine that appends the parts in the background.
+func (m *appendPartFS) appendRoutine(disk StorageAPI, bucket, object, uploadID string, info appendRoutineInfo) {
+	// Holds the list of parts that is already appended to the "append" file.
+	appendMeta := fsMetaV1{}
+	for {
+		select {
+		case input := <-info.inputCh:
+			// We receive on this channel when new part gets uploaded or when complete-multipart sends
+			// a value on this channel to confirm if all the required parts are appended.
+			meta := input.meta
+			for {
+				// Append should be done such a way that if part-3 and part-2 is uploaded before part-1, we
+				// wait till part-1 is uploaded after which we append part-2 and part-3 as well in this for-loop.
+				part, appendNeeded := partToAppend(meta, appendMeta)
+				if !appendNeeded {
+					if reflect.DeepEqual(meta.Parts, appendMeta.Parts) {
+						// Sending nil is useful so that the complete-multipart-upload knows that
+						// all the required parts have been appended.
+						input.errCh <- nil
+					} else {
+						// Sending error is useful so that complete-multipart-upload can fall-back to
+						// its own append process.
+						input.errCh <- errPartsMissing
+					}
+					break
+				}
+				if err := appendPart(disk, bucket, object, uploadID, part); err != nil {
+					input.errCh <- err
+					break
+				}
+				appendMeta.AddObjectPart(part.Number, part.Name, part.ETag, part.Size)
+			}
+		case <-info.endCh:
+			// Either complete-multipart-upload or abort-multipart-upload closed endCh to end the appendRoutine.
+			appendFilePath := getFSAppendDataPath(uploadID)
+			disk.DeleteFile(bucket, appendFilePath) // Delete the tmp append file in case abort-multipart was called.
+			return
+		case <-time.After(appendRoutineTimeout):
+			// Timeout the goroutine to garbage collect its resources. This would happen if the client initates
+			// a multipart upload and does not complete/abort it.
+			m.mu.Lock()
+			delete(m.appendRoutineMap, uploadID)
+			m.mu.Unlock()
+			close(info.timeoutCh)
+			// Delete the temporary append file as well.
+			appendFilePath := getFSAppendDataPath(uploadID)
+			disk.DeleteFile(bucket, appendFilePath)
+		}
+	}
+}
+
+// Appends the "part" to the append-file inside "tmp/" that finally gets moved to the actual location
+// upon complete-multipart-upload.
+func appendPart(disk StorageAPI, bucket, object, uploadID string, part objectPartInfo) error {
+	partPath := pathJoin(mpartMetaPrefix, bucket, object, uploadID, part.Name)
+	appendFilePath := getFSAppendDataPath(uploadID)
+
+	offset := int64(0)
+	totalLeft := part.Size
+	buf := make([]byte, readSizeV1)
+	for totalLeft > 0 {
+		curLeft := int64(readSizeV1)
+		if totalLeft < readSizeV1 {
+			curLeft = totalLeft
+		}
+		var n int64
+		n, err := disk.ReadFile(minioMetaBucket, partPath, offset, buf[:curLeft])
+		if n > 0 {
+			if err = disk.AppendFile(minioMetaBucket, appendFilePath, buf[:n]); err != nil {
+				disk.DeleteFile(bucket, appendFilePath)
+				return err
+			}
+		}
+		if err != nil {
+			if err == io.EOF || err == io.ErrUnexpectedEOF {
+				break
+			}
+			disk.DeleteFile(bucket, appendFilePath)
+			return err
+		}
+		offset += n
+		totalLeft -= n
+	}
+	return nil
+}
--- a/cmd/fs-v1-append-parts.go
+++ b/cmd/fs-v1-append-parts.go
@@ -8,83 +8,108 @@ import (
 	"time"
 )
 
-// Error sent by appendRoutine when there are holes in parts.
+// Error sent by appendParts go-routine when there are holes in parts.
 // For ex. let's say client uploads part-2 before part-1 in which case we
 // can not append and have to wait till part-1 is uploaded. Hence we return
 // this error. Currently this error is not used in the caller.
 var errPartsMissing = errors.New("required parts missing")
 
-// Error sent when appendRoutine has waited long enough and timedout.
-var errAppendRoutineTimeout = errors.New("timeout")
+// Error sent when appendParts go-routine has waited long enough and timedout.
+var errAppendPartsTimeout = errors.New("appendParts goroutine timeout")
 
-// Timeout value for the appendRoutine go-routine.
-var appendRoutineTimeout = 24 * 60 * 60 * time.Second
+// Timeout value for the appendParts go-routine.
+var appendPartsTimeout = 24 * 60 * 60 * time.Second
 
-// Holds a map of uploadID->appendRoutine
-type appendPartFS struct {
-	appendRoutineMap map[string]appendRoutineInfo
-	mu               sync.Mutex
+// Holds a map of uploadID->appendParts go-routine
+type backgroundAppend struct {
+	infoMap map[string]bgAppendPartsInfo
+	sync.Mutex
 }
 
-// Input to the appendRoutine
-type appendRoutineInput struct {
+// Input to the appendParts go-routine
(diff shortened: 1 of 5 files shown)


cmd/fs-v1-append-parts.go
```diff
@@ -0,0 +1,170 @@
+package cmd
```
harshavardhana: Add license header.

cmd/fs-v1-append-parts.go
```diff
@@ -0,0 +1,170 @@
...
+)
+
+// Error sent by appendRoutine when there are holes in parts.
+// For ex. let's say client uploads part-2 before part-1 in which case we
+// can not append and have to wait till part-1 is uploaded. Hence we return
+// this error. Currently this error is not used in the caller.
+var errPartsMissing = errors.New("required parts missing")
+
+// Error sent when appendRoutine has waited long enough and timedout.
+var errAppendRoutineTimeout = errors.New("timeout")
```
harshavardhana: we can call this better..

cmd/fs-v1-append-parts.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+	m.mu.Unlock()
+	errCh := make(chan error)
+	go func() {
+		// send input in a goroutine as send on the inputCh can block if appendRoutine
+		// is busy appending a part.
+		select {
+		case <-info.timeoutCh:
+			// This is to handle a rare race condition where we found info in m.appendRoutineMap
+			// but soon after that appendRoutine timedouted out.
```
harshavardhana: `timedouted out.` --> `appendRoutine has timed out`

cmd/fs-v1-append-parts.go
```diff
@@ -0,0 +1,170 @@
...
+
+// Identifies an appendRoutine.
+type appendRoutineInfo struct {
+	inputCh   chan appendRoutineInput
+	timeoutCh chan struct{} // closed by appendRoutine when it timesout
+	endCh     chan struct{} // closed after complete/abort of upload to end the appendRoutine
+}
+
+// Called after a part is uploaded so that it can be appended in the background.
+func (m *appendPartFS) processPart(disk StorageAPI, bucket, object, uploadID string, meta fsMetaV1) chan error {
```
harshavardhana: `processStart`  ---> `appendProcessStart` ?

cmd/fs-v1-append-parts.go
```diff
@@ -0,0 +1,170 @@
...
+	info, ok := m.appendRoutineMap[uploadID]
+	if !ok {
+		return
+	}
+	delete(m.appendRoutineMap, uploadID)
+	close(info.endCh)
+}
+
+// The go-routine that appends the parts in the background.
+func (m *appendPartFS) appendRoutine(disk StorageAPI, bucket, object, uploadID string, info appendRoutineInfo) {
```
harshavardhana: `appendRoutine` --> `appendInRoutine`

cmd/fs-v1-append-parts.go
krishnasrinivas: it's `processPart` ... since it's a function name I named it like a verb as in process the new part that got uploaded.

cmd/fs-v1-append-parts.go
krishnasrinivas: since it's a go-routine that runs in the background, i named it like a noun i.e it's a goroutine that's responsible for appending so the go-routine is called `appendRoutine` if it was a function then appendInRoutine would be better as it's a verb

cmd/fs-v1-append-parts.go
harshavardhana: we can say what it is going to do, 'processPart' doesn't convey what it is going to do. It is too generic.

cmd/fs-v1-append-parts.go
harshavardhana: The problem is `go` indicates that it's a go-routine() naming that as part of the function is just redundant and not conveying what it is really doing. Its really doing is `appendParts()`

harshavardhana (approved):

harshavardhana:
Can you guys review this @krisis and @fwessels ?

cmd/fs-v1-background-append.go
```diff
@@ -0,0 +1,212 @@
...
+	}()
+}
+
+// Called on complete-multipart-upload. Returns nil if app the required parts have been appended.
+func (b *backgroundAppend) complete(disk StorageAPI, bucket, object, uploadID string, meta fsMetaV1) error {
+	b.Lock()
+	info, ok := b.infoMap[uploadID]
+	b.Unlock()
+	if !ok {
+		return errPartsMissing
```
fwessels, line 97: Should we not b.remove(uploadID) the 'incomplete' upload?

cmd/fs-v1-background-append.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+	errCh := make(chan error)
+	select {
+	case <-info.timeoutCh:
+		// This is to handle a rare race condition where we found info in b.infoMap
+		// but soon after that appendParts go-routine timedouted out.
+		return errAppendPartsTimeout
```
fwessels, line 104: Same as above

cmd/fs-v1-background-append.go
```diff
@@ -0,0 +1,212 @@
...
+				}
+				appendMeta.AddObjectPart(part.Number, part.Name, part.ETag, part.Size)
+			}
+		case <-info.endCh:
+			// Either complete-multipart-upload or abort-multipart-upload closed endCh to end the appendParts go-routine.
+			appendFilePath := getFSAppendDataPath(uploadID)
+			disk.DeleteFile(bucket, appendFilePath) // Delete the tmp append file in case abort-multipart was called.
+			return
+		case <-time.After(appendPartsTimeout):
+			// Timeout the goroutine to garbage collect its resources. This would happen if the client initates
```
fwessels: init**i**ates

cmd/fs-v1-background-append.go
```diff
@@ -0,0 +1,212 @@
...
+// For ex. let's say client uploads part-2 before part-1 in which case we
+// can not append and have to wait till part-1 is uploaded. Hence we return
+// this error. Currently this error is not used in the caller.
+var errPartsMissing = errors.New("required parts missing")
+
+// Error sent when appendParts go-routine has waited long enough and timedout.
+var errAppendPartsTimeout = errors.New("appendParts goroutine timeout")
+
+// Timeout value for the appendParts go-routine.
+var appendPartsTimeout = 24 * 60 * 60 * time.Second
```
krisis, line 36: What is the purpose of the above timeout value? The way the timeout is used in `appendParts` method, it appears it includes the time when parts are received via inputCh and being appended. Isn't it to avoid a go-routine meant for appending parts run without doing useful work?

cmd/fs-v1-background-append.go
```diff
@@ -0,0 +1,212 @@
...
+func (b *backgroundAppend) complete(disk StorageAPI, bucket, object, uploadID string, meta fsMetaV1) error {
+	b.Lock()
+	info, ok := b.infoMap[uploadID]
+	b.Unlock()
+	if !ok {
+		return errPartsMissing
+	}
+	errCh := make(chan error)
+	select {
+	case <-info.timeoutCh:
```
krisis, line 101: What happens if both cases can run and the select statement picks the other case at random?

cmd/fs-v1-background-append.go
```diff
@@ -0,0 +1,212 @@
...
+				if err := appendPart(disk, bucket, object, uploadID, part); err != nil {
+					input.errCh <- err
+					break
+				}
+				appendMeta.AddObjectPart(part.Number, part.Name, part.ETag, part.Size)
+			}
+		case <-info.endCh:
+			// Either complete-multipart-upload or abort-multipart-upload closed endCh to end the appendParts go-routine.
+			appendFilePath := getFSAppendDataPath(uploadID)
+			disk.DeleteFile(bucket, appendFilePath) // Delete the tmp append file in case abort-multipart was called.
```
krisis: The comment above says that either completeMultipart or abortMultipart request could have closed `endCh` and that you should delete the temporary append file only if it were an abortMultipart request, but the delete happens unconditionally. Is the comment stale or is there something missing in the code?

krisis (commented):
Could we have a smaller filename instead of `fs-v1-background-append.go`? e.g, `fs-v1-bg-append.go`

cmd/fs-v1-background-append.go
krishnasrinivas: not needed because whatever caused `delete(b.infoMap, uploadID)` would have also done the necessary cleanup. i.e:

```
        delete(b.infoMap, uploadID)                                                                                                                                             
        close(info.endCh)
```

causes the following cleanup:

```
                case <-info.endCh:                                                                                                                                              
                        // Either complete-multipart-upload or abort-multipart-upload closed endCh to end the appendParts go-routine.                                           
                        appendFilePath := getFSAppendDataPath(uploadID)                                                                                                         
                        disk.DeleteFile(bucket, appendFilePath) // Delete the tmp append file in case abort-multipart was called.                                               
                        return
```

another place is during the timeout:

```
                case <-time.After(appendPartsTimeout):                                                                                                                          
                        // Timeout the goroutine to garbage collect its resources. This would happen if the client initates                                                     
                        // a multipart upload and does not complete/abort it.                                                                                                   
                        b.Lock()                                                                                                                                                
                        delete(b.infoMap, uploadID)                                                                                                                             
                        b.Unlock()                                                                                                                                              
                        // Delete the temporary append file as well.                                                                                                            
                        appendFilePath := getFSAppendDataPath(uploadID)                                                                                                         
                        disk.DeleteFile(bucket, appendFilePath)                                                                                                                 

                        close(info.timeoutCh)
```

cmd/fs-v1-background-append.go
krishnasrinivas: the necessary cleanup happens when the timeout happens inside the go-routine:

```
                case <-time.After(appendPartsTimeout):                                                                                                                          
                        // Timeout the goroutine to garbage collect its resources. This would happen if the client initates                                                     
                        // a multipart upload and does not complete/abort it.                                                                                                   
                        b.Lock()                                                                                                                                                
                        delete(b.infoMap, uploadID)                                                                                                                             
                        b.Unlock()                                                                                                                                              
                        // Delete the temporary append file as well.                                                                                                            
                        appendFilePath := getFSAppendDataPath(uploadID)                                                                                                         
                        disk.DeleteFile(bucket, appendFilePath)                                                                                                                 

                        close(info.timeoutCh)
```

cmd/fs-v1-background-append.go
krishnasrinivas: the timeout is needed so that a go-routine is not left dangling (and consuming resources unnecessarily) when a client initiates a multipart upload and does not do abort/complete.

> The way the timeout is used in appendParts method, it appears it includes the time when parts are received via inputCh and being appended.

that is correct behavior right? i.e if the go-routine has not received a new part for a long time, time it out.

>  Isn't it to avoid a go-routine meant for appending parts run without doing useful work?

did not get this :-(

cmd/fs-v1-background-append.go
krishnasrinivas: only one of the cases can happen, it depends on what happens inside the go-routine, i.e inside appendParts go-routine either receive on timeout channel or receive on inputCh can happen.

cmd/fs-v1-background-append.go
krishnasrinivas: comment is stale, will fix it.

cmd/fs-v1-background-append.go
krisis: Yeah, only one of those cases happen. This is because, after the go-routine exits/returns, info.timeoutCh is closed and there is no go-routine to receive anything sent on info.inputCh. So, a send on info.inputCh _will_ block and therefore the receive on info.timeoutCh would be the only non-blocking case statement. The other possibility needs no explanation.

cmd/fs-v1-background-append.go
krisis: I was wrong. It doesn't include the time spent in appending parts (ref: https://talks.golang.org/2012/concurrency.slide#35). This looks fine.

cmd/fs-v1-background-append.go
```diff
@@ -0,0 +1,212 @@
...
+		case info.inputCh <- bgAppendPartsInput{meta, errCh}:
+			// Receive the error so that the appendParts go-routine does not block on send.
+			// But the error received is ignored as fs.PutObjectPart() would have already
+			// returned success to the client.
+			<-errCh
+		}
+	}()
+}
+
+// Called on complete-multipart-upload. Returns nil if app the required parts have been appended.
```
donatello: Typo: "nil if app the required parts..."

cmd/fs-v1-background-append.go
```diff
@@ -0,0 +1,212 @@
...
+	totalLeft := part.Size
+	buf := make([]byte, readSizeV1)
+	for totalLeft > 0 {
+		curLeft := int64(readSizeV1)
+		if totalLeft < readSizeV1 {
+			curLeft = totalLeft
+		}
+		var n int64
+		n, err := disk.ReadFile(minioMetaBucket, partPath, offset, buf[:curLeft])
+		if n > 0 {
```
donatello: Better to check for error first - it is more idiomatic in go. Also, `disk.ReadFile()` may be leaving an invalid n > 0 value and giving a non-nil error (it may not be the case now, but if `ReadFile` is changed in such a way, it could break here as an unintended consequence).

krishnasrinivas: we had discussed that the contract semantics of `disk.ReadFile()` will be same as `io.ReadFull()` so the consumer of this API can safely assume it and this code should always work. Code elsewhere also has been written based on this assumption. Should be put in the comment of `posix.ReadFile`

Also realized that ErrUnexpectedEOF should never happen as we always know what size to read.

Since it is more readable to eliminate error conditions first (as you mentioned) re-writing this to:

```
func appendPart(disk StorageAPI, bucket, object, uploadID string, part objectPartInfo) error {                                                                                  
        partPath := pathJoin(mpartMetaPrefix, bucket, object, uploadID, part.Name)                                                                                              
        appendFilePath := getFSAppendDataPath(uploadID)                                                                                                                         

        offset := int64(0)                                                                                                                                                      
        totalLeft := part.Size                                                                                                                                                  
        buf := make([]byte, readSizeV1)                                                                                                                                         
        for totalLeft > 0 {                                                                                                                                                     
                curLeft := int64(readSizeV1)                                                                                                                                    
                if totalLeft < readSizeV1 {                                                                                                                                     
                        curLeft = totalLeft                                                                                                                                     
                }                                                                                                                                                               
                var n int64                                                                                                                                                     
                n, err := disk.ReadFile(minioMetaBucket, partPath, offset, buf[:curLeft])                                                                                       
                if err == io.EOF {                                                                                                                                              
                        break                                                                                                                                                   
                }                                                                                                                                                               
                if err != nil {                                                                                                                                                 
                        disk.DeleteFile(bucket, appendFilePath)                                                                                                                 
                        return err                                                                                                                                              
                }                                                                                                                                                               
                if err = disk.AppendFile(minioMetaBucket, appendFilePath, buf[:n]); err != nil {                                                                                
                        disk.DeleteFile(bucket, appendFilePath)                                                                                                                 
                        return err                                                                                                                                              
                }                                                                                                                                                               
                offset += n                                                                                                                                                     
                totalLeft -= n                                                                                                                                                  
        }                                                                                                                                                                       
        return nil                                                                                                                                                              
}
```

krishnasrinivas:
> Could we have a smaller filename instead of fs-v1-background-append.go? e.g, fs-v1-bg-append.go
> Hide outdated

@krisis fs-v1-background-append.go looks somehow better/readable than fs-v1-bg-append.go even though longer, there does not seem to be a better shorter name too. So for now kept as it is, but if you are particular then we can get inputs on what others think is a better name.

cmd/fs-v1-background-append.go
```diff
@@ -0,0 +1,213 @@
...
+	totalLeft := part.Size
+	buf := make([]byte, readSizeV1)
+	for totalLeft > 0 {
+		curLeft := int64(readSizeV1)
+		if totalLeft < readSizeV1 {
+			curLeft = totalLeft
+		}
+		var n int64
+		n, err := disk.ReadFile(minioMetaBucket, partPath, offset, buf[:curLeft])
+		if err == io.EOF {
```
donatello: This check also does not seem to be required. `io.EOF` is returned only when no bytes were read - this cannot happen if `curLeft > 0` (and this is true inside the loop). If it does happen, it is an actual error that should not be ignored -> we were unable to read the file up to the length that we thought the file had.

krishnasrinivas: @donatello you are right

harshavardhana:
@krishnasrinivas can you rebase?

krisis (approved):

cmd/fs-v1-background-append.go
```diff
@@ -0,0 +1,210 @@
...
+		n, err := disk.ReadFile(minioMetaBucket, partPath, offset, buf[:curLeft])
+		if err != nil {
+			// Check for EOF/ErrUnexpectedEOF not needed as it should never happen as we know
+			// the exact size of the file and hence know the size of buf[]
+			// EOF/ErrUnexpectedEOF indicates that the length of file was shorter than part.Size and
+			// hence considered as an error condition.
+			disk.DeleteFile(bucket, appendFilePath)
+			return err
+		}
+		if err = disk.AppendFile(minioMetaBucket, appendFilePath, buf[:n]); err != nil {
```
vadmeste, line 202: In general, tmp files should be accessed using minioMetaTmpBucket. This is a recent need and guarantees that tmp directory is not removed/created concurrently.

vadmeste: Will check this and make needed changes

krishnasrinivas: @vadmeste sorry I sent a PR without seeing this comment https://github.com/minio/minio/pull/3304

PR #3283, opened 2016-11-18
xl: GetObject should validate parts before reading.

### LINKED ISSUE #3278: [distributed] corrupted data when disk becomes full
## Expected Behavior
Data are always sane

## Current Behavior
I can download a corrupted file without having mc returning an error

1. Start from fresh disks.. but we need to have 2.3Gb remaining space
2. mc mb myminio/testbucket
3. mc cp 1gb.file myminio/testbucket/
4. mc cp 1gb.file myminio/testbucket/ (again)

It is normal if the second cp fails.. however mc cp didn't indicate any problem.. In addition, mc cat myminio/testbucket/1gb.file | md5sum will give an incorrect hash.

Minio prints some errors:
```
time="2016-11-18T14:09:15Z" level=error msg="Unable to create object part." cause="mkdir /export/minio3/.minio.sys/tmp/8994253a-3756-40c9-9d1f-1698680fe5a3: no space left on device" location="[object-handlers.go:691:objectAPIHandlers.PutObjectPartHandler()]" stack="xl-v1-utils.go:52:reduceErrs xl-v1-metadata.go:343:writeUniqueXLMetadata xl-v1-multipart.go:521:xlObjects.PutObjectPart <autogenerated>:854:(*xlObjects).PutObjectPart object-handlers.go:688:objectAPIHandlers.PutObjectPartHandler api-router.go:44:(objectAPIHandlers).PutObjectPartHandler-fm" 
time="2016-11-18T14:09:15Z" level=error msg="Unable to create object part." cause="mkdir /export/minio4/.minio.sys/tmp/3ef3aba5-d4d0-4137-919b-e6bdbad457b9: no space left on device" location="[object-handlers.go:691:objectAPIHandlers.PutObjectPartHandler()]" stack="xl-v1-utils.go:52:reduceErrs xl-v1-metadata.go:343:writeUniqueXLMetadata xl-v1-multipart.go:521:xlObjects.PutObjectPart <autogenerated>:854:(*xlObjects).PutObjectPart object-handlers.go:688:objectAPIHandlers.PutObjectPartHandler api-router.go:44:(objectAPIHandlers).PutObjectPartHandler-fm" 
time="2016-11-18T14:09:22Z" level=error msg="Unable to read part.14 of the object `testbucket/1gb.bin`." cause="Read failed. Insufficient number of disks online" location="[xl-v1-object.go:201:xlObjects.GetObject()]" stack="erasure-readfile.go:251:erasureReadFile xl-v1-object.go:199:xlObjects.GetObject /home/vadmeste/work/go/src/runtime/asm_amd64.s:2086:runtime.goexit" 
```

## Context
Testing

## Your Environment
All, commit-id 05dc52a206a47abd71aa1d81e5eb3053285f034a

harshavardhana:
## Description
This patch addresses a specific situation when `xl.json`
has an additional `part.N` which was never written fully
to the disk or renamed from temporary location.

This can potentially occur in cases where we don't have
enough disk space.

## Motivation and Context
Fixes #3278

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:

### DIFF
--- a/cmd/object-errors.go
+++ b/cmd/object-errors.go
@@ -325,6 +325,17 @@ func (e NotImplemented) Error() string {
 	return "Not Implemented"
 }
 
+// InvalidObjectState represents a state of an object either is its corrupted
+// or unreadable.
+type InvalidObjectState struct {
+	Bucket string
+	Object string
+}
+
+func (e InvalidObjectState) Error() string {
+	return "Bucket: " + e.Bucket + " Object " + e.Object + " is in invalid state"
+}
+
 // Check if error type is ObjectNameInvalid.
 func isErrObjectNameInvalid(err error) bool {
 	err = errorCause(err)
--- a/cmd/xl-v1-object.go
+++ b/cmd/xl-v1-object.go
@@ -156,6 +156,19 @@ func (xl xlObjects) GetObject(bucket, object string, startOffset int64, length i
 		}
 	}
 
+	// Validate if all the parts exist before proceeding to read from the disk.
+	for _, part := range xlMeta.Parts {
+		// Pick the first disk from online disks and validate if we have all the parts.
+		_, err = onlineDisks[0].StatFile(bucket, path.Join(object, part.Name))
+		if err != nil {
+			errorIf(err, "Unable to stat the file %s/%s", object, part.Name)
+			return traceError(InvalidObjectState{
+				Bucket: bucket,
+				Object: object,
+			})
+		}
+	}
+
 	totalBytesRead := int64(0)
 
 	chunkSize := getChunkSize(xlMeta.Erasure.BlockSize, xlMeta.Erasure.DataBlocks)
--- a/cmd/xl-v1-object_test.go
+++ b/cmd/xl-v1-object_test.go
@@ -163,6 +163,12 @@ func TestXLDeleteObjectDiskNotFound(t *testing.T) {
 }
 
 func TestGetObjectNoQuorum(t *testing.T) {
+	root, err := newTestConfig("us-east-1")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer removeAll(root)
+
 	// Create an instance of xl backend.
 	obj, fsDirs, err := prepareXL()
 	if err != nil {
@@ -189,7 +195,7 @@ func TestGetObjectNoQuorum(t *testing.T) {
 	// Make 9 disks offline, which leaves less than quorum number of disks
 	// in a 16 disk XL setup. The original disks are 'replaced' with
 	// naughtyDisks that fail after 'f' successful StorageAPI method
-	// invocations, where f - [0,2)
+	// invocations, where f - [0,2]
 	for f := 0; f < 2; f++ {
 		diskErrors := make(map[int]error)
 		for i := 0; i <= f; i++ {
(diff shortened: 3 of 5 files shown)


cmd/xl-v1-object.go
```diff
@@ -156,6 +156,19 @@ func (xl xlObjects) GetObject(bucket, object string, startOffset int64, length i
 		}
 	}
 
+	// Validate if all the parts exist before proceeding to read from the disk.
+	for _, part := range xlMeta.Parts {
+		// Pick the first disk from online disks and validate if we have all the parts.
+		_, err = onlineDisks[0].StatFile(bucket, path.Join(object, part.Name))
```
vadmeste, line 162: onlineDisks[0] ? won't this cause a permanent error when the first disk is unaccessible but we have quorum ?

cmd/object-errors.go
```diff
@@ -325,6 +325,17 @@ func (e NotImplemented) Error() string {
 	return "Not Implemented"
 }
 
+// InvalidObjectState represents a state of an object either is its corrupted
+// or unreadable.
+type InvalidObjectState struct {
```
donatello, line 330: Comment can be improved: "InvalidObjectState represents that the object with bucket and object name is invalid (either corrupted or unreadable)." Currently it sounds like the type holds two values - corrupted and unreadable.

cmd/xl-v1-object.go
harshavardhana: We can validate across all disks and see the palpable errors..

cmd/xl-v1-object_test.go
```diff
@@ -189,7 +195,7 @@ func TestGetObjectNoQuorum(t *testing.T) {
 	// Make 9 disks offline, which leaves less than quorum number of disks
 	// in a 16 disk XL setup. The original disks are 'replaced' with
 	// naughtyDisks that fail after 'f' successful StorageAPI method
-	// invocations, where f - [0,2)
+	// invocations, where f - [0,2]
```
donatello, line 198: By `[0,2)` they meant half open interval, including 0, but not 2 - meaning just 0 or 1; see https://en.wikipedia.org/wiki/Interval_(mathematics)#Terminology - so if we don't want to use such deep-math notation, it should be written `[0,1]` here as we mean `f` can be 0 or 1.

cmd/xl-v1-object_test.go
```diff
@@ -206,9 +212,13 @@ func TestGetObjectNoQuorum(t *testing.T) {
...
 		err = xl.GetObject(bucket, object, 0, int64(len("abcd")), ioutil.Discard)
 		err = errorCause(err)
-		if err != toObjectErr(errXLReadQuorum, bucket, object) {
-			t.Errorf("Expected putObject to fail with %v, but failed with %v", toObjectErr(errXLWriteQuorum, bucket, object), err)
+		if _, ok := err.(InvalidObjectState); ok {
+			continue
+		}
+		if _, ok := err.(InsufficientReadQuorum); ok {
+			continue
 		}
```
donatello, line 220: The preceding two `if`'s are more neatly expressed with a type switch:

``` go
switch _ := err.(type) {
case InvalidObjectState:
        fallthrough
case InsufficientReadQuorum:
        continue
}
```

cmd/xl-v1-object.go
donatello: Another question - why is it needed? If some parts don't exist, GetObject should fail midway. So this is just to error out a bit early?

donatello:
I don't understand how this is addressing #3278. If the second PUT object was actually failing because of lack of disk space, it should return an error - not corrupt an existing object. The check in this PR, though it tells that the object is corrupt, does not seem to stop the corruption which should not happen in the first place.

harshavardhana:
This is an incorrect fix the root cause needs to be known, closing this.

PR #3288, opened 2016-11-19
auth/rpc: Take remote disk offline after maximum allowed attempts.

### LINKED ISSUE #3286: [distributed] rpc/client: Track errors upto 5 attempts and do not attempt Login() after.
## Expected Behavior
Currently if a disk is down rpcClient would attempt a dial for each such an attempt. Instead the remote disk should be remembered as dead unless explicitly toggled by healing or management API call etc.

## Current Behavior
Each login attempts and fails but net.Dial can be an expensive operation.

## Possible Solution
Remember the errors and then permanently take the disk offline after upto 5 attempts.

Just take one of the server down.

## Context
Posix has a similar functionality of syscall.EIO errors, similar case should be added for rpcClient.

## Your Environment

* Version used: master branch.

harshavardhana:
## Description
Disks when are offline for a long period of time, we should
ignore the disk after trying Login upto 5 times.

This is to reduce the network chattiness, this also reduces
the overall time spent on `net.Dial`.

## Motivation and Context
Fixes #3286

## How Has This Been Tested?
Manually and through chaos test.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/auth-rpc-client.go
+++ b/cmd/auth-rpc-client.go
@@ -97,12 +97,13 @@ type authConfig struct {
 
 // AuthRPCClient is a wrapper type for RPCClient which provides JWT based authentication across reconnects.
 type AuthRPCClient struct {
-	mu            sync.Mutex
-	config        *authConfig
-	rpc           *RPCClient // reconnect'able rpc client built on top of net/rpc Client
-	isLoggedIn    bool       // Indicates if the auth client has been logged in and token is valid.
-	token         string     // JWT based token
-	serverVersion string     // Server version exchanged by the RPC.
+	mu             sync.Mutex
+	config         *authConfig
+	rpc            *RPCClient // reconnect'able rpc client built on top of net/rpc Client
+	isLoggedIn     bool       // Indicates if the auth client has been logged in and token is valid.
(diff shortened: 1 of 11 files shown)


harshavardhana:
Ping @krisis @krishnasrinivas can you review this?

cmd/auth-rpc-client.go
```diff
@@ -127,30 +128,50 @@ func (authClient *AuthRPCClient) Close() error {
...
+	// if there is an error.
+	defer func() {
+		if err != nil {
+			authClient.serverIOErrCnt++
+		}
+	}()
+
+	// Take remote disk offline if the total server errors
+	// are more than maximum allowable IO error limit.
+	if authClient.serverIOErrCnt > maxAllowedIOError {
```
donatello: Better to put this `if` block before calling the defer. If the threshold errors have been reached, then calling the defer it unnecessary.

donatello (approved):
LGTM - except for one minor suggested change.

cmd/auth-rpc-client.go
harshavardhana: This is something that needs to be done for posix.go as well.  - Can you open a bug for this optimization?

PR #3126, opened 2016-10-29
[WIP] Make XML error response compatible with AWS S3

harshavardhana:
## Description

This patch addresses a lot of compatibility issues which arised after keeping layers of errors. This change is all the more needed for our final stretch towards S3 compatibility.

## Motivation and Context

Fixes the problems of S3 compatible error messages fixes #2386

## How Has This Been Tested?

Using s3 tools like s3cmd, aws cli and mc.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

Fixes #2386

PR #3296, opened 2016-11-21
web: Add more data for jsonrpc responses.

harshavardhana:
## Description
This change adds more richer error response
for JSON-RPC by interpreting object layer
errors to corresponding meaningful errors
for the web browser.

```go
&json2.Error{
   Message: "Bucket Name Invalid, Only lowercase letters, full stops, and numbers are allowed.",
}
```

Additionally this patch also allows PresignedGetObject()
to take expiry parameter to have variable expiry.

## Motivation and Context
Refer https://github.com/minio/miniobrowser/issues/224 

## How Has This Been Tested?
With miniobrowser locally.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/web-handlers.go
+++ b/cmd/web-handlers.go
@@ -19,6 +19,7 @@ package cmd
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"net/http"
@@ -41,7 +42,7 @@ import (
 // isJWTReqAuthenticated validates if any incoming request to be a
 // valid JWT authenticated request.
 func isJWTReqAuthenticated(req *http.Request) bool {
-	jwt, err := newJWT(defaultJWTExpiry)
+	jwt, err := newJWT(defaultJWTExpiry, serverConfig.GetCredential())
 	if err != nil {
 		errorIf(err, "unable to initialize a new JWT")
 		return false
@@ -85,7 +86,7 @@ type ServerInfoRep struct {
 // ServerInfo - get server info.
 func (web *webAPIHandlers) ServerInfo(r *http.Request, args *WebGenericArgs, reply *ServerInfoRep) error {
 	if !isJWTReqAuthenticated(r) {
-		return &json2.Error{Message: errAuthentication.Error()}
+		return toJSONError(errAuthentication)
 	}
 	host, err := os.Hostname()
 	if err != nil {
@@ -123,10 +124,10 @@ type StorageInfoRep struct {
 func (web *webAPIHandlers) StorageInfo(r *http.Request, args *GenericArgs, reply *StorageInfoRep) error {
 	objectAPI := web.ObjectAPI()
 	if objectAPI == nil {
-		return &json2.Error{Message: errServerNotInitialized.Error()}
+		return toJSONError(errServerNotInitialized)
 	}
 	if !isJWTReqAuthenticated(r) {
-		return &json2.Error{Message: errAuthentication.Error()}
+		return toJSONError(errAuthentication)
 	}
 	reply.StorageInfo = objectAPI.StorageInfo()
 	reply.UIVersion = miniobrowser.UIVersion
@@ -142,13 +143,13 @@ type MakeBucketArgs struct {
 func (web *webAPIHandlers) MakeBucket(r *http.Request, args *MakeBucketArgs, reply *WebGenericRep) error {
 	objectAPI := web.ObjectAPI()
 	if objectAPI == nil {
-		return &json2.Error{Message: errServerNotInitialized.Error()}
+		return toJSONError(errServerNotInitialized)
 	}
 	if !isJWTReqAuthenticated(r) {
-		return &json2.Error{Message: errAuthentication.Error()}
+		return toJSONError(errAuthentication)
 	}
 	if err := objectAPI.MakeBucket(args.BucketName); err != nil {
-		return &json2.Error{Message: err.Error()}
+		return toJSONError(err, args.BucketName)
 	}
 	reply.UIVersion = miniobrowser.UIVersion
 	return nil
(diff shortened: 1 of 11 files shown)


krishnasrinivas:
This commit is not helping https://github.com/minio/miniobrowser/issues/224. Because our webAPI layer is a simple layer on top of object-layer. i.e it just does one-to-one translation to object-layer API. The object-layer API just returns `error` without any other context. If the `error` returned by object-layer embeds more context than just an error string then `json2.Error.Data` field will be useful, in which case we should not be putting the codes like `InvalidBucketName` in the Data field but rather why the bucket name is invalid, i.e "bucket name should not contain numbers" etc

harshavardhana:
> This commit is not helping minio/miniobrowser#224. Because our webAPI layer is a simple layer on top of object-layer. i.e it just does one-to-one translation to object-layer API. The object-layer API just returns error without any other context. If the error returned by object-layer embeds more context than just an error string then json2.Error.Data field will be useful, in which case we should not be putting the codes like InvalidBucketName in the Data field but rather why the bucket name is invalid, i.e "bucket name should not contain numbers" etc

You are right the context should be addressed - https://github.com/minio/minio/pull/3126 patch addresses this problem. 

But think of it like `mc` . Browser needs to interpret these errors and shouldn't be blind to errors. It makes sense for it to know what type of error is being returned so that it can craft a better message rather than just displaying S3 protocol specific errors.

harshavardhana:
> You are right the context should be addressed - #3126 patch addresses this problem.
> 
> But think of it like mc . Browser needs to interpret these errors and shouldn't be blind to errors. It makes sense for it to know what type of error is being returned so that it can craft a better message rather than just displaying S3 protocol specific errors.

Actually rather than browser, these handlers can return better errors. Since these are the wrappers of object layer.  Let me re-do this.

harshavardhana:
ping @krishnasrinivas @krisis - the patch is updated.

cmd/web-handlers.go
```diff
@@ -771,22 +739,23 @@ type PresignedGetRep struct {
...
-	}
 	if !isJWTReqAuthenticated(r) {
-		return &json2.Error{Message: errAuthentication.Error()}
+		return toJSONError(errAuthentication)
 	}
+
 	if args.BucketName == "" || args.ObjectName == "" {
-		return &json2.Error{Message: "Bucket, Object are mandatory arguments."}
+		return &json2.Error{
+			Message: "Bucket, Object are mandatory arguments.",
```
krisis: "Bucket and Object are mandatory arguments"

cmd/web-handlers.go
```diff
@@ -818,3 +791,94 @@ func presignedGet(host, bucket, object string) string {
...
+// toJSONError converts regular errors into more user friendly
+// and consumable error message for the browser UI.
+func toJSONError(err error, params ...string) (jerr *json2.Error) {
+	apiErr := toWebAPIError(err)
+	switch apiErr.Code {
+	// Bucket name invalid with custom error message.
+	case "InvalidBucketName":
+		if len(params) > 0 {
+			jerr = &json2.Error{
+				Message: fmt.Sprintf("Bucket Name %s Invalid, Only lowercase letters, full stops, and numbers are allowed.",
```
krisis: "Bucket name %s is invalid. Lowercase letters, period and numerals are the only allowed characters."

cmd/web-handlers.go
harshavardhana: done

cmd/web-handlers.go
harshavardhana: done

cmd/web-handlers.go
```diff
@@ -818,3 +789,94 @@ func presignedGet(host, bucket, object string) string {
...
 	return host + path + "?" + query + "&" + "X-Amz-Signature=" + signature
 }
+
+// toJSONError converts regular errors into more user friendly
+// and consumable error message for the browser UI.
+func toJSONError(err error, params ...string) (jerr *json2.Error) {
+	apiErr := toWebAPIError(err)
+	switch apiErr.Code {
+	// Bucket name invalid with custom error message.
+	case "InvalidBucketName":
```
krishnasrinivas, line 802: if the caller forgets to send bucketname: `toJSONError(err)` then jerr will be `nil`

krishnasrinivas (approved):

cmd/web-handlers.go
```diff
@@ -818,3 +789,93 @@ func presignedGet(host, bucket, object string) string {
...
+	apiErr := toWebAPIError(err)
+	jerr = &json2.Error{
+		Message: apiErr.Description,
+	}
+	switch apiErr.Code {
+	// Bucket name invalid with custom error message.
+	case "InvalidBucketName":
+		if len(params) > 0 {
+			jerr = &json2.Error{
+				Message: fmt.Sprintf("Bucket Name %s is invalid. Lowercase letters, period and numerals are the only allowed characters.",
```
justinclift, line 805: Not sure how to word it, but the S3 restrictions around period handling (in bucket names) might need to be referenced too.  For example, if someone creates a bucket with the first or last character being a period - an invalid name - this error message wouldn't help explain what's wrong.

Is there any way to include a link to the docs for further info?  eg: "Further information: https://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html#bucketnamingrules"

harshavardhana:
> Not sure how to word it, but the S3 restrictions around period handling (in bucket names) might need to be referenced too. For example, if someone creates a bucket with the first or last character being a period - an invalid name - this error message wouldn't help explain what's wrong.
> 
> Is there any way to include a link to the docs for further info? eg: "Further information: https://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html#bucketnamingrules"

In such situations we can just explain.  The all encompassing errors and messages themselves are not that useful.  This error message serves 90% of use-cases and is minimal.

justinclift:
Ahhh yeah, that makes sense. :smile:

PR #3300, opened 2016-11-21
utils: reduceErrs returns and validates quorum errors.

### LINKED ISSUE #3298: reduceErrs should return maximally occurring error and its count
[reduceErrs](https://github.com/minio/minio/blob/master/cmd/xl-v1-utils.go#L35) returns the error which occurs the most, filtering out ``ignoredErrs``. This is insufficient to know if the corresponding operation had the same result in quorum or more number of nodes.

## Expected Behavior

``reduceErrs`` should return maximally occurring error and its count
## Current Behavior

``reduceErrs`` returns the maximally occurring error. e.g, ``reduceErrs([]error{errFileNotFound, errFileNotFound, nil, nil}, nil}`` will return ``nil`` or ``errFileNotFound`` depending on the order in which the keys where inserted into the underlying map, which is used to count the number of occurrences of an error.

## Possible Solution

``reduceErrs`` should return the number of occurrences of the maximally occurring error.

N/A

## Context

With ``reduceErrs`` returning an error (incl. ``nil``) which may not have quorum may cause data to diverge among disks in a erasure code backend.

## Your Environment

* Version used:
* Server type and version:
* Link to your project:

harshavardhana:
## Description
This is needed as explained by @krisis 

Lets say we have following errors.

```
[]error{nil, errFileNotFound, errDiskAccessDenied, errDiskAccesDenied}
```

Since the last two errors are filtered, the maximum is nil,
depending on map order.

Let's say we get nil from reduceErr. Clearly at this point
we don't have quorum nodes agreeing about the data and since
GetObject only requires N/2 (Read quorum) and isDiskQuorum
would have returned true. This is problematic and can lead to
undersiable consequences.

## Motivation and Context
Fixes #3298 

## How Has This Been Tested?
Theoretical 

## Types of changes

## Checklist:

### DIFF
--- a/cmd/xl-v1-metadata.go
+++ b/cmd/xl-v1-metadata.go
@@ -335,8 +335,7 @@ func writeUniqueXLMetadata(disks []StorageAPI, bucket, prefix string, xlMetas []
 		deleteAllXLMetadata(disks, bucket, prefix, mErrs)
 		return traceError(errXLWriteQuorum)
 	}
-
-	return reduceErrs(mErrs, objectOpIgnoredErrs)
+	return reduceWriteQuorumErrs(mErrs, objectOpIgnoredErrs, quorum)
 }
 
 // writeSameXLMetadata - write `xl.json` on all disks in order.
@@ -375,6 +374,5 @@ func writeSameXLMetadata(disks []StorageAPI, bucket, prefix string, xlMeta xlMet
 		deleteAllXLMetadata(disks, bucket, prefix, mErrs)
 		return traceError(errXLWriteQuorum)
 	}
-
-	return reduceErrs(mErrs, objectOpIgnoredErrs)
+	return reduceWriteQuorumErrs(mErrs, objectOpIgnoredErrs, writeQuorum)
 }
--- a/cmd/xl-v1-object.go
+++ b/cmd/xl-v1-object.go
@@ -79,7 +79,7 @@ func (xl xlObjects) GetObject(bucket, object string, startOffset int64, length i
 		return traceError(InsufficientReadQuorum{}, errs...)
 	}
 
-	if reducedErr := reduceErrs(errs, objectOpIgnoredErrs); reducedErr != nil {
+	if reducedErr := reduceReadQuorumErrs(errs, objectOpIgnoredErrs, xl.readQuorum); reducedErr != nil {
 		return toObjectErr(reducedErr, bucket, object)
 	}
 
@@ -339,8 +339,7 @@ func rename(disks []StorageAPI, srcBucket, srcEntry, dstBucket, dstEntry string,
 		undoRename(disks, srcBucket, srcEntry, dstBucket, dstEntry, isPart, errs)
 		return traceError(errXLWriteQuorum)
 	}
-	// Return on first error, also undo any partially successful rename operations.
-	return reduceErrs(errs, objectOpIgnoredErrs)
+	return reduceWriteQuorumErrs(errs, objectOpIgnoredErrs, quorum)
 }
 
 // renamePart - renames a part of the source object to the destination
--- a/cmd/xl-v1-utils.go
+++ b/cmd/xl-v1-utils.go
@@ -31,8 +31,7 @@ import (
 // golang's map orders keys. This doesn't affect correctness as long as quorum
 // value is greater than or equal to simple majority, since none of the equally
 // maximal values would occur quorum or more number of times.
-
-func reduceErrs(errs []error, ignoredErrs []error) error {
+func reduceErrs(errs []error, ignoredErrs []error) (maxCount int, maxErr error) {
 	errorCounts := make(map[error]int)
 	errs = errorsCause(errs)
 	for _, err := range errs {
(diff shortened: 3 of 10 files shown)


cmd/xl-v1-metadata.go
```diff
@@ -336,7 +336,12 @@ func writeUniqueXLMetadata(disks []StorageAPI, bucket, prefix string, xlMetas []
 		return traceError(errXLWriteQuorum)
 	}
 
-	return reduceErrs(mErrs, objectOpIgnoredErrs)
+	if maxCount, reducedErr := reduceErrs(mErrs, objectOpIgnoredErrs); reducedErr != nil {
+		if maxCount >= quorum {
+			return reducedErr
```
krisis: if the maxCount is less than quorum we should return the corresponding quorum error (read or write).
```go
else {
  return traceError(errXLWriteQuorum)
}
```

cmd/xl-v1-utils.go
```diff
@@ -42,14 +41,20 @@ func reduceErrs(errs []error, ignoredErrs []error) error {
...
 	max := 0
-	var errMax error
 	for err, count := range errorCounts {
 		if max < count {
 			max = count
-			errMax = err
+			maxErr = err
+		}
+	}
+	if maxErr != nil {
```
krisis: Even if ``maxErr`` is ``nil`` the count must be greater than or equal to the quorum value.

cmd/xl-v1-utils.go
```diff
@@ -42,14 +41,20 @@ func reduceErrs(errs []error, ignoredErrs []error) error {
...
+		if max >= quorum {
+			maxErr = traceError(maxErr, errs...)
+		} else {
+			maxErr = traceError(errXLReadQuorum, errs...)
```
krisis: It could be read or write quorum depending on the operation. It can't be errXLReadQuorum unconditionally.

cmd/xl-v1-object.go
```diff
@@ -339,8 +339,11 @@ func rename(disks []StorageAPI, srcBucket, srcEntry, dstBucket, dstEntry string,
 		undoRename(disks, srcBucket, srcEntry, dstBucket, dstEntry, isPart, errs)
 		return traceError(errXLWriteQuorum)
 	}
-	// Return on first error, also undo any partially successful rename operations.
-	return reduceErrs(errs, objectOpIgnoredErrs)
+	// Return on quorum error.
+	if reducedErr := reduceWriteQuorumErrs(errs, objectOpIgnoredErrs, quorum); reducedErr != nil {
```
krisis: you could do ``return reduceWriteQuorumErrs(...)``. reduceWriteQuorumErrs handles all error  values alike, including nil.

cmd/xl-v1-utils.go
```diff
@@ -42,14 +41,45 @@ func reduceErrs(errs []error, ignoredErrs []error) error {
...
+	}
+	return max, maxErr
+}
+
+// reduceReadQuorumErrs behaves like reduceErrs but only for returning
+// values of maximally occuring errors validated against readQuorum.
+func reduceReadQuorumErrs(errs []error, ignoredErrs []error, readQuorum int) (maxErr error) {
+	maxCount, maxErr := reduceErrs(errs, ignoredErrs)
+	if maxCount >= readQuorum {
+		if maxErr != nil {
```
krisis: While this is correct, we can simplify as follows. There is no need to handle ``nil`` differently.
```go
if maxCount >= readQuorum {
  return traceError(maxErr, errs)
} else {
  return traceError(errXLReadQuorum, errs)
}
```

cmd/xl-v1-utils.go
```diff
@@ -42,14 +41,45 @@ func reduceErrs(errs []error, ignoredErrs []error) error {
...
+		}
+		return maxErr
+	}
+	maxErr = traceError(errXLReadQuorum, errs...)
+	return
+}
+
+// reduceWriteQuorumErrs behaves like reduceErrs but only for returning
+// values of maximally occuring errors validated against writeQuorum.
+func reduceWriteQuorumErrs(errs []error, ignoredErrs []error, writeQuorum int) (maxErr error) {
```
krisis, line 81: reduceReadQuorumErrs and reduceWriteQuorumErrs can be unified as below,
```go
func reduceQuorumErrs(errs, ignoredErrs []error, quorum int, quorumErr error) error {
  maxCount, maxErr := reduceErrs(errs, ignoredErrs)
  if maxCount >= quorum {
    return traceError(maxErr, errs)
  } 
  return traceError(quorumErr, errs)
}

func reduceReadQuorumErrs(errs, ignoredErrs []error, readQuorum int) {
  return reduceQuorumErrs(errs, ignoredErrs, readQuorum, errXLReadQuorum)
}

func reduceWriteQuorumErrs(errs, ignoredErrs []error, writeQuorum int) {
  return reduceQuorumErrs(errs, ignoredErrs, writeQuorum, errXLWriteQuorum)
}
```

krisis (approved):
LGTM

PR #3309, opened 2016-11-22
Remove uploadIDChange structure.

balamurugana:
addUploadID() and removeUploadID() are wrappers to updateUploadJSON()
which is called with respective arguments.  New/Abort/Complete MultipartUpload methods call wrapper functions.

## Motivation and Context
Make cleaner readable code for better understanding.

## How Has This Been Tested?
`make test`

## Types of changes

## Checklist:

### DIFF
--- a/cmd/fs-v1-multipart-common.go
+++ b/cmd/fs-v1-multipart-common.go
@@ -16,7 +16,10 @@
 
 package cmd
 
-import "path"
+import (
+	"path"
+	"time"
+)
 
 // Returns if the prefix is a multipart upload.
 func (fs fsObjects) isMultipartUpload(bucket, prefix string) bool {
@@ -52,11 +55,10 @@ func (fs fsObjects) isUploadIDExists(bucket, object, uploadID string) bool {
 	return true
 }
 
-// writeUploadJSON - create `uploads.json` or update it with new uploadID.
-func (fs fsObjects) updateUploadJSON(bucket, object string, uCh uploadIDChange) error {
+// updateUploadJSON - add or remove upload ID info in all `uploads.json`.
+func (fs fsObjects) updateUploadJSON(bucket, object, uploadID string, initiated time.Time, isRemove bool) error {
 	uploadsPath := path.Join(bucket, object, uploadsJSONFile)
-	uniqueID := getUUID()
-	tmpUploadsPath := uniqueID
+	tmpUploadsPath := getUUID()
 
 	uploadsJSON, err := readUploadsJSON(bucket, object, fs.storage)
 	if errorCause(err) == errFileNotFound {
@@ -69,12 +71,12 @@ func (fs fsObjects) updateUploadJSON(bucket, object string, uCh uploadIDChange) 
 	}
 
 	// update the uploadsJSON struct
-	if !uCh.isRemove {
+	if !isRemove {
 		// Add the uploadID
-		uploadsJSON.AddUploadID(uCh.uploadID, uCh.initiated)
+		uploadsJSON.AddUploadID(uploadID, initiated)
 	} else {
 		// Remove the upload ID
-		uploadsJSON.RemoveUploadID(uCh.uploadID)
+		uploadsJSON.RemoveUploadID(uploadID)
 	}
 
 	// update the file or delete it?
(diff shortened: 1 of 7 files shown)


cmd/xl-v1-multipart-common.go
```diff
@@ -19,14 +19,13 @@ package cmd
...
 	"path"
 	"sync"
+	"time"
 )
 
-// writeUploadJSON - create `uploads.json` or update it with change
-// described in uCh.
-func (xl xlObjects) updateUploadJSON(bucket, object string, uCh uploadIDChange) error {
+// updateUploadJSON - add or remove upload ID info in all `uploads.json`.
+func (xl xlObjects) updateUploadJSON(bucket, object, uploadID string, initiated *time.Time, isRemove bool) error {
```
harshavardhana: There is no need to use `*.time.Time` here @balamurugana

cmd/fs-v1-multipart-common.go
```diff
@@ -52,11 +55,10 @@ func (fs fsObjects) isUploadIDExists(bucket, object, uploadID string) bool {
 	return true
 }
 
-// writeUploadJSON - create `uploads.json` or update it with new uploadID.
-func (fs fsObjects) updateUploadJSON(bucket, object string, uCh uploadIDChange) error {
+// updateUploadJSON - add or remove upload ID info in all `uploads.json`.
+func (fs fsObjects) updateUploadJSON(bucket, object, uploadID string, initiated *time.Time, isRemove bool) error {
```
harshavardhana: Same as below.

cmd/xl-v1-multipart-common.go
balamurugana: If its not `*time.Time`, then removeUploadID() would need to send `time.Time{}` which is not necessary.  Currently it calls `updateUploadJSON()` with `nil`

harshavardhana: No we shouldn't add it like this time.Time{} is never a pointer and no functions in Go return or interpret it like a pointer.  We should avoid pointers. nil has nothing special time.Time{} is atleast nicer and points to `time.Unix(0, 0)` unix epoch time which is nil for its type.

balamurugana: I guess `time.Time{}` or `time.Unix(0,0)` creates an memory usage, but `nil` is not.  That's the difference.  However its all inside/masked by `addUploadID()` and `removeUploadID()` functions

harshavardhana: it is not idiomatic Go style, you are bringing C style here you are forcing a pointer which has no real purpose here.

balamurugana: I see `initiated` is used only if `isRemove` is `false`.  I don't mind of using a value than a pointer.  But I don't understand why pointer using is not idiomatic Go style.  Its all depending on the context.  The question here is whether we want to allocate dummy structure or not.

harshavardhana: Its not about dummy structure, pointer is only used in Go when you have a mutating value and you are interested in doing some mutation on it.  But it is not correct to treat it like argument validation.

donatello: Adding a pointer argument requires you to now check if it is `nil` before safely using it. This is another reason to not use pointer here.

donatello:
The argument for readability here is that having extra add/remove function wrappers is better than a documented struct that compactly achieves the same behavior.

I may be biased having written the original code, but my view is that this is sacrificing brevity for little to no readability improvement, especially considering the original code's struct was sufficiently documented. My view is that more code is not always improving readability and more code is definitely more maintenance and source of bugs.

cmd/xl-v1-multipart-common.go
balamurugana: @donatello If thats the case, you should check `Time.IsZero()` along the side. My points are specific to this patch and this context.

As I see empty time structure is the way to go, I am modifying in the next refresh.

donatello: @balamurugana My point was that if pointer is used, you could pass nil (which crashes - worst kind of bug), zero time (which is a bug), and a non-zero time which may or may not be a bug. So if type is not a pointer, the programmer only needs to make sure that a correct value is given (no possibility of nil value crash) - but if type is a pointer, then programmer needs to make sure value is not accidentally nil, or check for nil.

harshavardhana:
> I may be biased having written the original code, but my view is that this is sacrificing brevity for little to no readability improvement, especially considering the original code's struct was sufficiently documented. My view is that more code is not always improving readability and more code is definitely more maintenance and source of bugs.

Actually struct may be a preferred approach but only when this struct is consumed elsewhere too and methods attached to it are useful in manipulating the struct.  The best example of such a struct is `xlMetaV1{}`  but not the one which is removed here.

PR #3315, opened 2016-11-22
Protect multipart dir from removing when it is empty

### LINKED ISSUE #3313: distributed/fs: Protect multipart directory from deletion
We don't really need to remove .minio.sys/multipart directory when this latter becomes empty, we can avoid eventual bugs like that.

vadmeste:
## Description
Do not remove multipart dir since it has no real gain

## Motivation and Context
Fix #3313 

## How Has This Been Tested?
Manually, go test, s3verify
 
## Types of changes

## Checklist:

### DIFF
--- a/cmd/fs-v1-multipart-common.go
+++ b/cmd/fs-v1-multipart-common.go
@@ -40,13 +40,13 @@ func (fs fsObjects) isBucketExist(bucket string) bool {
 
 // isUploadIDExists - verify if a given uploadID exists and is valid.
 func (fs fsObjects) isUploadIDExists(bucket, object, uploadID string) bool {
-	uploadIDPath := path.Join(mpartMetaPrefix, bucket, object, uploadID)
-	_, err := fs.storage.StatFile(minioMetaBucket, path.Join(uploadIDPath, fsMetaJSONFile))
+	uploadIDPath := path.Join(bucket, object, uploadID)
+	_, err := fs.storage.StatFile(minioMetaMultipartBucket, path.Join(uploadIDPath, fsMetaJSONFile))
 	if err != nil {
 		if err == errFileNotFound {
 			return false
 		}
-		errorIf(err, "Unable to access upload id"+uploadIDPath)
+		errorIf(err, "Unable to access upload id "+pathJoin(minioMetaMultipartBucket, uploadIDPath))
 		return false
 	}
 	return true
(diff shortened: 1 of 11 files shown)


cmd/fs-v1-multipart-common.go
```diff
@@ -40,21 +40,21 @@ func (fs fsObjects) isBucketExist(bucket string) bool {
...
-	uploadIDPath := path.Join(mpartMetaPrefix, bucket, object, uploadID)
-	_, err := fs.storage.StatFile(minioMetaBucket, path.Join(uploadIDPath, fsMetaJSONFile))
+	uploadIDPath := path.Join(bucket, object, uploadID)
+	_, err := fs.storage.StatFile(minioMetaMultipartBucket, path.Join(uploadIDPath, fsMetaJSONFile))
 	if err != nil {
 		if err == errFileNotFound {
 			return false
 		}
-		errorIf(err, "Unable to access upload id"+uploadIDPath)
+		errorIf(err, "Unable to access upload id"+minioMetaMultipartBucket+"/"+uploadIDPath)
```
harshavardhana: use `pathJoin`

cmd/fs-v1.go
```diff
@@ -76,7 +76,7 @@ func newFSObjects(storage StorageAPI) (ObjectLayer, error) {
 // Should be called when process shuts down.
 func (fs fsObjects) Shutdown() error {
 	// List if there are any multipart entries.
-	_, err := fs.storage.ListDir(minioMetaBucket, mpartMetaPrefix)
+	_, err := fs.storage.ListDir(minioMetaBucket, "multipart")
```
harshavardhana: You can keep "multipart" constant so that we have a single place for this.

cmd/fs-v1_test.go
```diff
@@ -126,8 +126,9 @@ func TestFSShutdown(t *testing.T) {
 	}
 	removeAll(disk)
 
+	// FIXME: Check why Shutdown returns success when second posix call returns faulty disk error
```
harshavardhana, line 129: Are you testing this?

cmd/object-utils.go
```diff
@@ -31,8 +31,8 @@ import (
 const (
 	// Minio meta bucket.
 	minioMetaBucket = ".minio.sys"
-	// Multipart meta prefix.
-	mpartMetaPrefix = "multipart"
```
harshavardhana, line 35: keep this, can be used.

cmd/fs-v1.go
vadmeste: > You can keep "multipart" constant so that we have a single place for this.

yes.. I wanted to discuss why we need to keep incomplete uploads data in FS after shutdown.. but let's fix the code and talk about this later.

cmd/fs-v1_test.go
vadmeste: > Are you testing this?

no. I disabled it intentionally because it is causing some problems.. This tests if shutdown returns faulty disks errors with naughty disks, can we merge it like that for the moment?

harshavardhana: we can you open a bug for it?

PR #3318, opened 2016-11-22
server: forbid zero port in address flag since it confuses clients and other Minio nodes too.

vadmeste:
## Description
address flag in server should not accept zero port since it confuses clients and other Minio nodes too.

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:

### DIFF
--- a/cmd/server-main.go
+++ b/cmd/server-main.go
@@ -369,6 +369,12 @@ func serverMain(c *cli.Context) {
 	// Check if requested port is available.
 	host, portStr, err := net.SplitHostPort(serverAddr)
 	fatalIf(err, "Unable to parse %s.", serverAddr)
+	if portStr == "0" || portStr == "" {
+		// Port zero or empty means use requested to choose any freely available
+		// port. Avoid this since it won't work with any configured clients,
+		// can lead to serious loss of availability.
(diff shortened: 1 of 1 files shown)


cmd/server-main.go
```diff
@@ -369,6 +369,11 @@ func serverMain(c *cli.Context) {
 	// Check if requested port is available.
 	host, portStr, err := net.SplitHostPort(serverAddr)
 	fatalIf(err, "Unable to parse %s.", serverAddr)
+	if portStr == "0" {
+		// Zero port means choose any available port, avoids this since it won't work
```
harshavardhana:
```
// Port zero means use requested to choose any freely available 
// port. Avoid this since it won't work with any configured clients,
// can lead to serious loss of availability.
```

cmd/server-main.go
```diff
@@ -369,6 +369,11 @@ func serverMain(c *cli.Context) {
...
+		// with the distributed config and all clients will be confused too.
+		fatalIf(errInvalidArgument, "Cannot accept zero port number")
```
harshavardhana: ``fatalIf(errInvalidArgument, "Invalid port %s, please use `--address` to pick a specific port")``

PR #3337, opened 2016-11-23
init: Honor config-dir flag when it is passed as global or local flag

vadmeste:
## Description
setGlobalsFromContext() is added to sets global variable after parsing
command line arguments.

## Motivation and Context
minio server /tmp/fs -C configdir doesn't work 

## Types of changes

## Checklist:

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -148,70 +148,84 @@ func checkMainSyntax(c *cli.Context) {
 	}
 }
 
+// Parse command arguments and set global variables accordingly
+func setGlobalsFromContext(c *cli.Context) {
+	// Set config dir
+	switch {
+	case c.IsSet("config-dir"):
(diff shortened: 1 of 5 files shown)


cmd/main.go
```diff
@@ -148,56 +148,66 @@ func checkMainSyntax(c *cli.Context) {
...
+
+// Main main for minio server.
+func Main() {
+	app := registerApp()
+	app.Before = func(c *cli.Context) error {
+		// Valid input arguments to main.
+		checkMainSyntax(c)
 
 		// Set global quiet flag.
 		globalQuiet = c.Bool("quiet") || c.GlobalBool("quiet")
```
harshavardhana: Even this flag needs to be moved into setGlobalsFromContext

vadmeste: Done

PR #3345, opened 2016-11-24
Docs: Fix command sample to be copy-pastable

donatello:
## Description

Minor docs fix for making command copy-pastable and alignment for readability.

## Types of changes

## Checklist:

### DIFF
--- a/docs/distributed/README.md
+++ b/docs/distributed/README.md
@@ -14,17 +14,17 @@ Distributed Minio provides protection against multiple node/drive failures and [
 
 A stand-alone Minio server would go down if the server hosting the disks goes offline. In contrast, a distributed Minio setup with _n_ disks will have your data safe as long as _n/2_ or more disks are online. You'll need a minimum of _(n/2 + 1)_ [Quorum](https://github.com/minio/dsync#lock-process) disks to create new objects though.
 
-For example, a 8 nodes distributed Minio setup, with 1 disk per node would stay put, even if upto 4 nodes are offline. But, you'll need atleast 5 nodes online to create new objects.
+For example, an 8-node distributed Minio setup, with 1 disk per node would stay put, even if upto 4 nodes are offline. But, you'll need atleast 5 nodes online to create new objects.
 
 ## Limitations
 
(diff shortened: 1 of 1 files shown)


docs/distributed/README.md
```diff
@@ -14,17 +14,17 @@ Distributed Minio provides protection against multiple node/drive failures and [
 
 A stand-alone Minio server would go down if the server hosting the disks goes offline. In contrast, a distributed Minio setup with _n_ disks will have your data safe as long as _n/2_ or more disks are online. You'll need a minimum of _(n/2 + 1)_ [Quorum](https://github.com/minio/dsync#lock-process) disks to create new objects though.
 
-For example, a 8 nodes distributed Minio setup, with 1 disk per node would stay put, even if upto 4 nodes are offline. But, you'll need atleast 5 nodes online to create new objects.
+For example, an 8-node distributed Minio setup, with 1 disk per node would stay put, even if upto 4 nodes are offline. But, you'll need atleast 5 nodes online to create new objects.
```
nitisht, line 17: Agree with an instead of a. But I am not in favour of -, as in 8-node, I could not find any standard guide that clearly says whether - should be used in such cases. Looks like more of a personal preference to me.

nitisht:
LGTM 👍

docs/distributed/README.md
donatello: Here's one: https://en.oxforddictionaries.com/punctuation/hyphen - quoting a relevant bit:

> It’s important to use hyphens in compound adjectives describing ages and lengths of time: leaving them out can make the meaning ambiguous. For example, 250-year-old trees clearly refers to trees that are 250 years old, while 250 year old trees could equally refer to 250 trees that are all one year old.

donatello: I understand there is no actual ambiguity in our doc but using the hyphen makes our meaning explicit.

nitisht: > compound adjectives describing ages and lengths of time: 

There is no age/time involved here. Though 8-node is a compound adjective that comes before the noun, so we can probably use the hyphen.

Issue #2991, opened 2016-10-18
erasure/distributed: Sync in-memory bucket metadata automatically


donatello:
In some cases, in-memory bucket metadata such as policy configuration, notifications configuration or listener configuration can go stale due to dropped/missed messages. We need a simple framework that automatically heals/fixes this.

deekoder:
needs to be discussed.

donatello:
On discussion with @abperiasamy it was decided to write a periodic checker that lists all buckets and syncs the bucket configurations by reading from object layer.

harshavardhana:
> On discussion with @abperiasamy it was decided to write a periodic checker that lists all buckets and syncs the bucket configurations by reading from object layer.

Are you still doing this @donatello ?

harshavardhana:
Notification subsystem and policy subsystem has been fully refactored, this situation is handled more cleanly now.

PR #3367, opened 2016-11-29
Avoid interface usage in sendUpdate()

balamurugana:
## How Has This Been Tested?

`make test`

## Types of changes

## Checklist:

### DIFF
--- a/cmd/s3-peer-client.go
+++ b/cmd/s3-peer-client.go
@@ -18,7 +18,6 @@ package cmd
 
 import (
 	"encoding/json"
-	"fmt"
 	"net/url"
 	"path"
 	"sync"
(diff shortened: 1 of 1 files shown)


donatello (changes requested):
This change is not useful. What is the reason to avoid interface here? Three highly repetitive functions have been added, and the ability to uniformly perform retries for erring nodes has been removed.

harshavardhana:
> This change is not useful. What is the reason to avoid interface here? Three highly repetitive functions have been added, and the ability to uniformly perform retries for erring nodes has been removed.

Generally a catch all interface is not the correct implementation typed is more preferred, while there is repetition in code which should be reduced in a separate manner.  

```interface{}```  has many pitfalls and lacks certain safely guarantees.

balamurugana:
> What is the reason to avoid interface here?

Another advantage of strict typing brings compilers to catch errors than failure at run time.  The best practice to avoid `interface` as argument/return type unless otherwise needed eg json/xml parsers.

donatello:
> >   This change is not useful. What is the reason to avoid interface here? Three highly repetitive functions have been added, and the ability to uniformly perform retries for erring nodes has been removed.
>
> Generally a catch all interface is not the correct implementation typed is more preferred, while there is repetition in code which should be reduced in a separate manner.
>
> interface{} has many pitfalls and lacks certain safely guarantees.

@harshavardhana It is better to bring in the repetition reducing code in along with this change - I did not use an interface lightly here as I am aware of certain pitfalls with interfaces, but I do not believe there is a simple way to reduce the repetition here. There is no need to bring in this change when there is no demonstrable benefit - just saying that interfaces are bad doesn't convey any concrete problem here.

donatello:
>>  What is the reason to avoid interface here?
>
> Another advantage of strict typing brings compilers to catch errors than failure at run time. The best practice to avoid interface as argument/return type unless otherwise needed eg json/xml parsers.

@balamurugana Yes, I agree and do understand the advantages of strict typing. I made a judicious choice to use an interface in an internal function to not only reduce code-repetition, but to also have a simple way to uniformly retry erring remote nodes.

Interfaces are the only feature of golang that let us write polymorphic code. It is very limited, but we either have to deal carefully with it or we end up repeating so much code and having a much larger surface area for bugs that are hard to detect via code inspection.

harshavardhana:
> @harshavardhana It is better to bring in the repetition reducing code in along with this change - I did not use an interface lightly here as I am aware of certain pitfalls with interfaces, but I do not believe there is a simple way to reduce the repetition here. There is no need to bring in this change when there is no demonstrable benefit - just saying that interfaces are bad doesn't convey any concrete problem here.

The concrete problem is `interface{}` itself it makes a highly unstable function.  Repetition at the benefit of safety, we prefer safety at this point.  The problem was SendUpdate() as a function does many things and no one can understand that by just looking at the function name and its input arguments - when its input argument can be anything, internally it does three different disconnected things.

krisis:
>The concrete problem is interface{} itself it makes a highly unstable function. Repetition at the benefit of safety, we prefer safety at this point. The problem was SendUpdate() as a function does many things and no one can understand that by just looking at the function name and its input arguments - when its input argument can be anything, internally it does three different disconnected things.

SendUpdate() is unifying the code responsible for sending bucket notification, policy notifications and listen bucket notifications. While they do different things, they share the same structure, i.e using a sync.WaitGroup to coordinate among many go-routines. This is being abstracted out. Why would someone find it difficult to understand? How would use of interface make this function unstable? I agree that we will only know at runtime if the type was wrong, if there exists a bug in our code due to oversight. But the function is neither complex nor is it likely to do more things.

donatello:
> The concrete problem is interface{} itself it makes a highly unstable function. Repetition at the benefit of safety, we prefer safety at this point.

It is stable by simple inspection - the code related to this particular interface{} usage is quite small and is present in the same (small) file. 

Repetition too is a form of losing safety. It is very easy to miss a subtle difference when reading very similar things. Even in this case, what if there are more kinds of updates to broadcast in the future? It will end up with even more repetition. What if we want to do controlled retries for errors? More repetition.

There are some future tasks that are likely to come back and modify the related code here - https://github.com/minio/minio/issues/2991 - here we _may_ have to broadcast more kinds of updates - this change is just going to make that task introduce even more repetitive code.

> The problem was SendUpdate() as a function does many things and no one can understand that by just looking at the function name and its input arguments - when its input argument can be anything, internally it does three different disconnected things.

From reading both the copious comments and the code, anyone could understand that the `s3Peers.SendUpdate()` function is simply doing what the (whole) name suggests - sending messages of certain types to peers in the cluster.

krishnasrinivas:
https://github.com/krishnasrinivas/minio/commit/9f6cddaa7d435b2a681ebbd8fc27adcf3346adad

above change would provide best of both worlds i.e no code repetition and better type checking

harshavardhana:
> krishnasrinivas@9f6cdda
> 
> above change would provide best of both worlds i.e no code repetition and better type checking

Yes this was the next suggestion. Thanks @krishnasrinivas

krisis:
> krishnasrinivas@9f6cdda
above change would provide best of both worlds i.e no code repetition and better type checking

👍

donatello:
> krishnasrinivas@9f6cdda
> 
> above change would provide best of both worlds i.e no code repetition and better type checking

This is great. We should incorporate this.

harshavardhana:
Who is sending the updated PR?

krishnasrinivas:
> Who is sending the updated PR?

sent https://github.com/minio/minio/issues/3375

harshavardhana:
A different change has been merged this is not needed anymore @balamurugana

harshavardhana:
Closing this PR since another change superseded this change.

PR #3368, opened 2016-11-29
api: Allow reconnection of policy/notification rpc clients.

harshavardhana:
## Description
Since we moved out reconnection logic from net-rpc-client.go
we should do it from the top-layer properly and bring back
the code to reconnect properly in-case the connection is lost.

## Motivation and Context
Reconnection is handled by top layer not internally in net-rpc-client.go

## How Has This Been Tested?
make test and distributed setup. 

## Types of changes

## Checklist:

### DIFF
--- a/cmd/bucket-metadata.go
+++ b/cmd/bucket-metadata.go
@@ -16,7 +16,10 @@
 
 package cmd
 
-import "encoding/json"
+import (
+	"encoding/json"
+	"net/rpc"
+)
 
 // BucketMetaState - Interface to update bucket metadata in-memory
 // state.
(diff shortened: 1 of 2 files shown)


donatello (approved):

krisis:
@harshavardhana I still see code in the Call() method, in net-rpc-client.go, to connect if the underlying rpc connection was found disconnected. Do we still need to try once more for policy change notifications alone?

harshavardhana:
> @harshavardhana I still see code in the Call() method, in net-rpc-client.go, to connect if the underlying rpc connection was found disconnected. Do we still need to try once more for policy change notifications alone?

Where do you see? @krisis - rpclocalStack is not nil on the first attempt here since it is not proactively closed yet.

harshavardhana:
> Where do you see? @krisis - rpclocalStack is not nil on the first attempt here since it is not proactively closed yet.

This now tells me that i need to Close() first.

krisis:
@harshavardhana Ok, I overlooked the fact that the Call() method doesn't call clearRPCClient(), to trigger a reconnect on subsequent invocation of Call().

PR #3381, opened 2016-12-01
lock: Moving locking to handler layer.

### LINKED ISSUE #2912: [lock] Possible race between GetObject and GetObjectInfo.
In get object we do:
objectAPI.GetObjectInfo()

// some checks

objectAPI.GetObject()

when `// some checks` are performed, if the object is replaced it will cause race condition.

### LINKED ISSUE #3370: Safe file uploading
When I use `mc cp` to upload file and the other process is simultaneously downloading object with same name using `mc cp` too, eventually downloading breaks down due to file size changing. How can I safely upload file to minio bucket?

harshavardhana:
## Description
This is implemented so that the issues like in the
following flow don't affect the behavior of operation.

```
GetObjectInfo()
.... --> Time window for mutation (no lock held)
.... --> Time window for mutation (no lock held)
GetObject()
```

This happens when two simultaneous uploads are made
to the same object the object has returned wrong
info to the client.

Another classic example is "CopyObject" API itself
which reads from a source object and copies to
destination object.

## Motivation and Context
Fixes #3370
Fixes #2912

## How Has This Been Tested?
go test and `mc cp`

## Types of changes

## Checklist:

### DIFF
--- a/cmd/bucket-handlers.go
+++ b/cmd/bucket-handlers.go
@@ -334,6 +334,10 @@ func (api objectAPIHandlers) PutBucketHandler(w http.ResponseWriter, r *http.Req
 		return
 	}
 
+	bucketLock := globalNSMutex.NewNSLock(bucket, "")
+	bucketLock.Lock()
+	defer bucketLock.Unlock()
+
 	// Proceed to creating a bucket.
 	err := objectAPI.MakeBucket(bucket)
 	if err != nil {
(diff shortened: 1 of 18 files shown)


harshavardhana:
This is pending on @abperiasamy to be reviewed, posting it here for discussion.

harshavardhana:
This is an important change need reviews here @krisis @krishnasrinivas @donatello

donatello (approved):
LGTM

PR #3383, opened 2016-12-02
fs/xl: Combine input checks into re-usable functions.

harshavardhana:
## Description
Repeated code around both object layers are moved
and combined into simple re-usable functions.

## Motivation and Context
Code cleanup and maintenance. 

## How Has This Been Tested?
go test and other local tools.

## Types of changes

## Checklist:


harshavardhana:
This is a precursor cleanup before we bring the changes in "ErrorResponse" cleanup patch - https://github.com/minio/minio/pull/3126  - since it has become a large patch. Breaking them down so that it becomes easy to review.

cmd/xl-v1-multipart.go
```diff
@@ -783,7 +708,7 @@ func (xl xlObjects) CompleteMultipartUpload(bucket string, object string, upload
 	destLock := nsMutex.NewNSLock(bucket, object)
 	destLock.Lock()
 	defer func() {
-		// A new complete multipart upload invalidates any
+		// A new complete multipart upload inchecks any
```
donatello: s/inchecks/expires/ or s/inchecks/invalidates

cmd/xl-v1-object.go
```diff
@@ -404,7 +383,7 @@ func (xl xlObjects) PutObject(bucket string, object string, size int64, data io.
 
 	// If caching is enabled, proceed to set the cache.
 	if size > 0 && xl.objCacheEnabled {
-		// PutObject invalidates any previously cached object in memory.
+		// PutObject inchecks any previously cached object in memory.
```
donatello: `inchecks` is not really a word. `invalidates` or `expires` is better.

cmd/object-api-input-checks.go
```diff
@@ -0,0 +1,151 @@
...
+	return checkPutObjectArgs(bucket, object, obj)
+}
+
+// Checks for PutObject arguments validity, also validates if bucket exists.
+func checkPutObjectArgs(bucket, object string, obj ObjectLayer) error {
+	// Verify bucket exists.
+	if err := checkBucketExist(bucket, obj); err != nil {
+		return traceError(err)
+	}
+	if !IsValidObjectName(object) {
```
donatello, line 142: `IsValidObjectName` is a simpler check (involves only string processing) and it should be done before `checkBucketExist` as that is a more expensive object layer check.

cmd/object-api-input-checks.go
```diff
@@ -0,0 +1,151 @@
...
+	return nil
+}
+
+// Checks for all ListObjects arguments validity.
+func checkListObjsArgs(bucket, prefix, marker, delimiter string, obj ObjectLayer) error {
+	// Verify bucket exists.
+	if err := checkBucketExist(bucket, obj); err != nil {
+		return traceError(err)
+	}
+	if !IsValidObjectPrefix(prefix) {
```
donatello, line 59: `IsValidObjectPrefix` is a simple string processing check and should be done before the IO performing `checkBucketExist` check, and in fact after all simple checks where the order is of no consequence - here it should be done as the last check.

cmd/xl-v1-healing.go
```diff
@@ -28,7 +28,7 @@ func healFormatXL(storageDisks []StorageAPI) (err error) {
 	// Attempt to load all `format.json`.
 	formatConfigs, sErrs := loadAllFormats(storageDisks)
 
-	// Generic format check validates
+	// Generic format check checks
```
donatello: This change has made the comment more cryptic - I suggest: `Generic format check steps:`

cmd/object-api-input-checks.go
```diff
@@ -0,0 +1,151 @@
...
+	return checkBucketAndObject(bucket, object)
+}
+
+// Checks on DeleteObject arguments, bucket and object.
+func checkDelObjArgs(bucket, object string) error {
+	return checkBucketAndObject(bucket, object)
+}
+
+// Checks bucket and object validity, returns nil if both are valid.
+func checkBucketAndObject(bucket, object string) error {
```
donatello: Is `checkBucketAndObjectNames` a better name here? It indicates the actual intent of the function - rather than about the validity of object store entities like buckets and objects.

donatello (commented):
This is a good change overall.

cmd/xl-v1-object.go
harshavardhana: Oops sed and replace issue. I wondered something like this would happen.

cmd/object-api-input-checks.go
harshavardhana: No there is a reason behind it. The order of errors is important. A bucket not found should happen before invalid prefix if both are given.

cmd/xl-v1-healing.go
harshavardhana: Again sed and replace. Let me fix them.

cmd/object-api-input-checks.go
harshavardhana: Yeah bucket names is better.

cmd/object-api-input-checks.go
harshavardhana: As explained above.

cmd/object-api-input-checks.go
harshavardhana: There is something new came to my mind, let me send a fix with that change. to simplify this further.

harshavardhana: Okay that change will happen here - https://github.com/minio/minio/pull/3381 - basically validating buckets right at the handler layer with in the lock rather than here at i/o layer.

harshavardhana: So this code will change but as i explained why the order matters can't change it in this patch.

donatello: :+1:

donatello (approved):

cmd/fs-v1-multipart-common_test.go
```diff
@@ -23,7 +23,7 @@ import (
 	"time"
 )
 
-// TestFSIsBucketExist - complete test of isBucketExist
```
krisis, line 26: Do we still need this unit test given that the function is refactored to work uniformly with XL and FS backends.

cmd/object-api-input-checks.go
```diff
@@ -0,0 +1,161 @@
...
+	return checkPutObjectArgs(bucket, object, obj)
+}
+
+// Checks for PutObject arguments validity, also validates if bucket exists.
+func checkPutObjectArgs(bucket, object string, obj ObjectLayer) error {
+	// Verify if bucket exists before validating object name.
+	// This is done on purpose since the order of errors is
+	// important here bucket does not exist error should
+	// happen before we return an error for invalid object name.
+	// FIXME: should be moved to handler layer.
```
krisis, line 137: Is this FIXME still valid?

krisis (commented):
Overall the changes look good to me.

cmd/object-api-input-checks.go
harshavardhana: yes it is valid.. the fixme will be fixed in handler level locking changes.

harshavardhana: https://github.com/minio/minio/pull/3381

krisis (approved):

cmd/object-api-input-checks.go
```diff
@@ -0,0 +1,161 @@
...
+			Object: object,
+		})
+	}
+	return nil
+}
+
+// Checks whether bucket exists and returns appropriate error if not.
+func checkBucketExist(bucket string, obj ObjectLayer) error {
+	if !IsValidBucketName(bucket) {
+		return BucketNameInvalid{Bucket: bucket}
```
vadmeste, line 154: traceError() missing ? and below in the same function too. LGTM otherwise.

harshavardhana: ah shucks.. will address it.

PR #3397, opened 2016-12-05
fs: Re-implement object layer to remember the fd.

harshavardhana:
This is to bring in support for shared backend.

## Types of changes

## Checklist:


harshavardhana:
Critical issue on windows is fixed now.. Any help on the testing and reviews would be good now.

harshavardhana:
Looks like the tests failed on appveyor, some permission/capability issues.  Will find a way to fix it.

harshavardhana:
A new problem on Windows still fixing.

cmd/posix.go
```diff
@@ -602,7 +602,7 @@ func (s *posix) createFile(volume, path string) (f *os.File, err error) {
 	if err != nil {
 		return nil, err
 	}
-	// Stat a volume entry.
+	// Stat a volume ekntry.
```
krisis: typo: s/ekntry/entry/

cmd/fs-v1.go
```diff
@@ -42,91 +54,194 @@ type fsObjects struct {
...
 }
 
-// list of all errors that can be ignored in tree walk operation in FS
-var fsTreeWalkIgnoredErrs = []error{
-	errFileNotFound,
-	errVolumeNotFound,
+// Initializes meta volume on all the fs path.
+func initMetaVolumeFS(fsPath, fsUUID string) error {
+	// This happens for the first time, but keep this here since this
+	// is the only place where it can be made expensive optimizing all
```
krisis: s/made expensive/made less expensive/ ?

cmd/fs-v1.go
```diff
@@ -42,91 +54,194 @@ type fsObjects struct {
...
+	// other calls. Create minio meta volume, if it doesn't exist yet.
+	// Create `.minio.sys`.
```
krisis: Create minio meta volume is the same as creating .minio.sys/ directory right?

cmd/fs-createfile.go
```diff
@@ -16,24 +16,116 @@
...
+		}
+		return err
+	}
+	if pathSt.IsDir() && !isDirEmpty(deletePath) {
+		// Verify if directory is empty.
+		return nil
+	}
+	// Attempt to lock the file if it exists,
+	// - if the file doesn't exist. We don't need to do anything,
+	//   we proceed and rename the file to actual namespace.
```
krisis: s/rename/remove

cmd/fs-createfile.go
```diff
@@ -16,24 +16,116 @@
...
+	// - if the file exists. Then we try to get a write lock this
+	//   will block if we can't get a lock perhaps another write
+	//   or read is in progress. Concurrent calls are protected
+	//   by the global namspace lock with in the same process.
```
krisis: s/with in/within/

cmd/fs-v1-background-append.go
```diff
@@ -186,29 +190,25 @@ func (b *backgroundAppend) appendParts(disk StorageAPI, bucket, object, uploadID
...
+	rfile, err := lock.LockedOpenFile(preparePath(partPath), os.O_RDONLY, 0666)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return errPartsMissing
 		}
-		offset += n
-		totalLeft -= n
+		return err
+	}
+	defer rfile.Close()
```
krisis: wouldn't this close other read locks held on this path?

krisis (commented):
These are initial comments, mostly minor.

cmd/fs-v1-background-append.go
harshavardhana: I think you are right, i was testing another theory to hold lock and validate if the file exists.. i tried to explain that in fsRenameFile(). 

But i think there is a problem in that too and fdMap is the only way to go, will re-do these changes.

harshavardhana:
Comments are certainly missing and perhaps has typos, please leave review comments with those mentions as well.  So that i can keep track and add them 

Please do test out the binary that i shared on minio channel.

pkg/lock/lock_nix.go
```diff
@@ -0,0 +1,44 @@
...
+
+// LockedOpenFile - initializes a new lock and protects the file from
+// concurrent access across mount points.
+func LockedOpenFile(path string, flag int, perm os.FileMode) (*LockedFile, error) {
+	f, err := os.OpenFile(path, flag, perm)
+	if err != nil {
+		return nil, err
+	}
+	switch {
+	case flag&os.O_RDONLY != 0:
```
balamurugana: you could use readable way of `flag & os.O_RDONLY == os.O_RDONLY`

pkg/lock/lock_nix.go
```diff
@@ -0,0 +1,44 @@
...
+		err = syscall.Flock(int(f.Fd()), syscall.LOCK_SH)
+	case flag&os.O_WRONLY != 0:
```
balamurugana: Same as above

pkg/lock/lock_nix.go
harshavardhana: Done

pkg/lock/lock_nix.go
harshavardhana: Done

balamurugana (commented):
Partial review

pkg/lock/lock_nix.go
balamurugana: how about flag having `os.RDWR`?  actually `os.RDWR & os.WRONLY != os.WRONLY`

harshavardhana: Let me make it generic..

pkg/lock/.gitignore
```diff
@@ -0,0 +1,2 @@
+*.test
```
donatello: Why is `.gitignore` file needed here?

pkg/lock/lock.go
```diff
@@ -0,0 +1,119 @@
...
+	Size() int64
+}
+
+// LockedFile - locked file is a locked instance, protected for
+// any simultaneous i/o operation.
+type LockedFile struct {
+	*os.File
+}
+
+// RLockedFile represents a rlocked file, implements a special
```
donatello: `RLockedFile represents a rlocked file...` is a bit circular.

pkg/lock/lock_windows.go
```diff
@@ -0,0 +1,142 @@
...
+
+func makeInheritSa() *syscall.SecurityAttributes {
+	var sa syscall.SecurityAttributes
+	sa.Length = uint32(unsafe.Sizeof(sa))
+	sa.InheritHandle = 1
+	return &sa
+}
+
+// perm param is ignored, on windows file perms/NT acls
+// are not octect combinations. Providing access to
```
donatello: `octet`

pkg/lock/lock.go
```diff
@@ -0,0 +1,119 @@
...
+	}, nil
+}
+
+// WLockedFile represents a wrapped write locked file.
+type WLockedFile struct {
+	*LockedFile
+}
+
+// WLockedOpenFile - returns a wrapped write locked file.
+func WLockedOpenFile(path string) (*WLockedFile, error) {
```
donatello: It is not clear what happens if the file is already open in read mode - it would be good to comment it.

cmd/fs-v1.go
```diff
@@ -232,76 +412,64 @@ func (fs fsObjects) GetObject(bucket, object string, offset int64, length int64,
 	}
 
 	// Stat the file to get file size.
-	fi, err := fs.storage.StatFile(bucket, object)
+	fi, err := os.Stat(preparePath(pathJoin(fs.fsPath, bucket, object)))
```
donatello: There seems to be a race here.

If we go via `os.Stat` here to find the size of the file without using the fd from fs.readersPool, it looks like we may read the info of a different file because a racing putobject might change the file before we go to reading the file here.

donatello (changes requested):
Partial review. I think I may have found a race.

cmd/fs-v1.go
```diff
@@ -20,113 +20,239 @@ import (
...
+	"syscall"
 
+	"github.com/minio/minio/pkg/disk"
+	"github.com/minio/minio/pkg/lock"
 	"github.com/minio/minio/pkg/mimedb"
 )
 
 // fsObjects - Implements fs object layer.
 type fsObjects struct {
-	storage StorageAPI
```
balamurugana, line 36: I see we can't use existing `posix.go` and we short-circuit here.  I would think of an `locked-posix.go` having shared `FS` friendly layer.  Note that `locked-posix.go` != `StorageAPI`.  This will bring some clean abstraction.

harshavardhana: It is intentional to make it monolith there is no abstraction. Just use it directly.

pkg/lock/lock.go
harshavardhana: Yep will write

pkg/lock/lock.go
harshavardhana: It should be read lock.

cmd/fs-v1.go
harshavardhana: There is no race. This entire thing is locked by namespace. This is why object level locking moved to handler layer.

harshavardhana: Will see if we can avoid the Stat()

krishnasrinivas:
Two parallel `mc cp` to the same object name but different minio servers will cause problems when two parallel CompleteMultipartUpload races. i.e one CompleteMultipartUpload can invoke fsRemoveUploadIDPath while the 2nd CompleteMultipartUpload is still appending which will see part file not found error. Hence the client will not see invalid uploadid.

If a PutObjectPart competes with CompleteMultipartUpload then it can leave an inconsistent backend in `.minio.sys/multipart`

cmd/fs-v1.go
balamurugana: The idea behind `locked-posix.go` is to make posix related operation separately and make testable locked posix operations. Now we need to test `FS` for such things. This also makes avoiding repetition of logic.

My proposal is to extend `posix.go` to `locked-posix.go` and have specific methods related to shared fs backend.

Another issue in this PR is it does not handle I/O errors properly like `posix.go`

harshavardhana: There are some other fundamental things that need to be changed in this patch working on a different approach.

harshavardhana:
> Two parallel mc cp to the same object name but different minio servers will cause problems when two parallel CompleteMultipartUpload races. i.e one CompleteMultipartUpload can invoke fsRemoveUploadIDPath while the 2nd CompleteMultipartUpload is still appending which will see part file not found error. Hence the client will not see invalid uploadid.
> 
> If a PutObjectPart competes with CompleteMultipartUpload then it can leave an inconsistent backend in .minio.sys/multipart

This is fixed @krishnasrinivas

krishnasrinivas:
> This is fixed @krishnasrinivas

reviewing

harshavardhana:
Some parts of the code can be moved together but it is not done yet.. will do as i get reviews.

cmd/fs-v1.go
```diff
@@ -353,17 +550,41 @@ func (fs fsObjects) PutObject(bucket string, object string, size int64, data io.
...
 		return ObjectInfo{}, err
 	}
+
 	// No metadata is set, allocate a new one.
 	if metadata == nil {
 		metadata = make(map[string]string)
 	}
 
-	uniqueID := mustGetUUID()
+	var fsMeta fsMetaV1
```
krisis: How do we synchronize between PutObject requests from different minio server instances where ``bucket == minioMetaBucket``?

pkg/lock/lock.go
```diff
@@ -0,0 +1,120 @@
...
+	if err != nil {
+		lkFile.Close()
+		return nil, err
+	}
+	return &RWLockedFile{
+		RLockedFile: &RLockedFile{
+			File:      lkFile,
+			rlockRefs: 1,
+			size:      st.Size(),
+			isClosed:  false,
```
krisis: ``isClosed`` isn't being used. It can be removed.

pkg/lock/lock.go
```diff
@@ -0,0 +1,120 @@
...
+	st, err := lkFile.Stat()
+	if err != nil {
+		lkFile.Close()
+		return nil, err
+	}
+	return &RLockedFile{
+		File:      lkFile,
+		rlockRefs: 1,
+		size:      st.Size(),
+		isClosed:  false,
```
krisis: ``isClosed`` is not being used. It can be removed.

pkg/lock/lock.go
```diff
@@ -0,0 +1,120 @@
...
+)
+
+// RLockedFile represents a read locked file, implements a special
+// closer which only closes the associated fd when the lockCount
+// has reached zero, i.e when all the readers have given up their locks.
+type RLockedFile struct {
+	*os.File
+	mutex     sync.Mutex
+	isClosed  bool
+	rlockRefs int // Holds read lock refs.
```
krisis: we could call it simply ``ref`` instead of ``rlockRefs`` since we use it for ``RWLockedFile`` too.

pkg/lock/lock.go
```diff
@@ -0,0 +1,120 @@
...
+	size      int64
+}
+
+// RWLockedFile represents a read/write locked file, implements a special
+// closer which only closes the associated fd when the lockCount
```
krisis: nit: s/associated fd/associated *os.File/

cmd/fs-v1.go
```diff
@@ -353,17 +550,41 @@ func (fs fsObjects) PutObject(bucket string, object string, size int64, data io.
...
+	var wlk *lock.RWLockedFile
+	if bucket != minioMetaBucket {
+		// Save objects' metadata in `fs.json`.
+		// Skip creating fs.json if bucket is .minio.sys as the object would have been created
+		// by minio's S3 layer (ex. policy.json)
+		fsMeta = newFSMetaV1()
+		fsMeta.Meta = metadata
+
+		fsMetaPath := pathJoin(fs.fsPath, minioMetaBucket, bucketMetaPrefix, bucket, object, fsMetaJSONFile)
+		wlk, err = fs.rwPool.Create(fsMetaPath)
```
krisis, line 545: There is no use in making ``Create`` a method of ``fs.rwPool``, unlike ``Open``. ``Create`` should be made a simple function or possibly a method of ``fsObjects`` if there is any benefit out of it.

cmd/fs-v1.go
```diff
@@ -474,20 +685,88 @@ func (fs fsObjects) DeleteObject(bucket, object string) error {
 		return err
 	}
 
+	bucketDir, err := fs.getBucketDir(bucket)
+	if err != nil {
+		return toObjectErr(err, bucket)
+	}
+
+	var wlk *lock.RWLockedFile
 	if bucket != minioMetaBucket {
```
krisis, line 671: How do we synchronize PutObject requests from different minio server instances for policy.json like objects? It appears that ``wlk == nil`` in those cases.

cmd/fs-v1-background-append.go
```diff
@@ -186,29 +188,27 @@ func (b *backgroundAppend) appendParts(disk StorageAPI, bucket, object, uploadID
...
+			return errPartsMissing
 		}
-		offset += n
-		totalLeft -= n
+		return err
+	}
+	defer file.Close()
+
+	tmpObjPath := pathJoin(fs.fsPath, minioMetaTmpBucket, fs.fsUUID, uploadID)
+	// No need to hold a lock, this is a unique file and will be only written
```
krisis, line 206: This is not guaranteed. The user must configure such that all multipart requests corresponding uploadID arrive on the same minio server instance.
e.g,
server1                                                                
- newMultipartUpload -> upload1                       
- putObjectPart(upload1, part1, part2, part4)
- completeMultipart(upload1, ...)

server2
- putObjectPart(upload1, part3, part4)
- completeMultipart(upload1, ...)

The background append process in server1 and server2 would be competing to write on the temporary file, which is not unique intrinsically and needs to be configured at load-balancer layer.

krisis (commented):
These are initial review comments.

cmd/fs-v1-background-append.go
harshavardhana: > The background append process in server1 and server2 would be competing to write on the temporary file, which is not unique intrinsically and needs to be configured at load-balancer layer.

The temporary file is unique FS is using `<fs_path>/<tmp>/<uuid>/<uploadid>` look at the fsUUID value.

harshavardhana: This fsUUID is unique per each instance.

pkg/lock/lock.go
harshavardhana: Done.

pkg/lock/lock.go
harshavardhana: Done

pkg/lock/lock.go
harshavardhana: Done

pkg/lock/lock.go
harshavardhana: Done

cmd/fs-v1.go
harshavardhana: The reason it is kept is for consistency since it was easy to read Open,Create -  right now there is no such need.

cmd/fs-v1.go
harshavardhana: Ah good catch, will fix it.

cmd/fs-v1.go
harshavardhana: Ah good catch, will fix it.

cmd/fs-v1.go
```diff
@@ -231,115 +390,66 @@ func (fs fsObjects) GetObject(bucket, object string, offset int64, length int64,
...
-			break
-		}
-		if totalLeft == 0 {
-			break
-		}
-	}
 
-	// Returns any error.
-	return toObjectErr(err, bucket, object)
+	_, err = io.CopyBuffer(writer, reader, buf)
```
krishnasrinivas: length is not honored if length > readSizeV1

harshavardhana: done - fixed.

pkg/lock/lock_nix.go
```diff
@@ -0,0 +1,58 @@
...
+	case syscall.O_RDWR:
+		fallthrough
+	case syscall.O_WRONLY | syscall.O_CREAT:
+		fallthrough
+	case syscall.O_RDWR | syscall.O_CREAT:
+		lockType = syscall.LOCK_EX
+	default:
+		return nil, fmt.Errorf("Unsupported flag (%d)", flag)
+	}
+	f, err := os.OpenFile(path, flag, perm)
```
krishnasrinivas: can always do O_SYNC so that after write we dont have to do a sync(), since this is always a metadata file it's fine to do O_SYNC here

cmd/fs-v1.go
```diff
@@ -355,17 +465,31 @@ func (fs fsObjects) PutObject(bucket string, object string, size int64, data io.
...
-	uniqueID := mustGetUUID()
+	fsMeta := newFSMetaV1()
+	fsMeta.Meta = metadata
+
+	fsMetaPath := pathJoin(fs.fsPath, minioMetaBucket, bucketMetaPrefix, bucket, object, fsMetaJSONFile)
+	wlk, err := fs.rwPool.Create(fsMetaPath)
+	if err != nil {
+		return ObjectInfo{}, toObjectErr(err, bucket, object)
+	}
+	if err = fs.writeFSMetadata(wlk, fsMeta); err != nil {
```
krishnasrinivas: fs.json should be written after the object is renamed

cmd/fs-v1-metadata.go
```diff
@@ -161,3 +173,39 @@ func isPartsSame(uploadedParts []objectPartInfo, completeParts []completePart) b
...
+	return fsMeta, nil
+}
+
+// Write fsMeta to fs.json.
+func (fs fsObjects) writeFSMetadata(writer io.Writer, fsMeta fsMetaV1) error {
+	metadataBytes, err := json.Marshal(fsMeta)
+	if err != nil {
+		return traceError(err)
+	}
+	_, err = writer.Write(metadataBytes)
```
krishnasrinivas: should truncate before writing because previously uploaded object metadata might exist.

cmd/fs-rwpool.go
```diff
@@ -0,0 +1,101 @@
...
+// ReaderAt is provided so that the fd is non seekable, since
+// we are sharing fd's with concurrent threads, we don't want
+// all readers to change offsets on each other during such
+// concurrent operations. Using ReadAt allows us to read from
+// any offsets.
+//
+// Closer is implemented to track total readers and to close
+// only when there no more readers, the fd is purged if the lock
+// count has reached zero.
+func (fsi *fsIOPool) Open(path string) (*lock.RLockedFile, error) {
```
krishnasrinivas, line 46: can this directly return io.SectionReader?

cmd/fs-rwpool.go
```diff
@@ -0,0 +1,101 @@
...
+//   will block if we can't get a lock perhaps another write
+//   or read is in progress. Concurrent calls are protected
+//   by the global namspace lock within the same process.
+func (fsi *fsIOPool) Create(path string) (wlk *lock.RWLockedFile, err error) {
+	if err = mkdirAll(pathutil.Dir(path), 0777); err != nil {
+		return nil, err
+	}
+	wlk, err = lock.RWLockedOpenFile(preparePath(path))
+	if err != nil {
+		if err = os.Remove(preparePath(pathutil.Dir(path))); err != nil {
```
krishnasrinivas: it's safer not to remove in case of error so that already existing file is not removed because of a temporary IO error etc

cmd/fs-rwpool.go
```diff
@@ -0,0 +1,101 @@
...
+// all readers to change offsets on each other during such
+// concurrent operations. Using ReadAt allows us to read from
+// any offsets.
+//
+// Closer is implemented to track total readers and to close
+// only when there no more readers, the fd is purged if the lock
+// count has reached zero.
+func (fsi *fsIOPool) Open(path string) (*lock.RLockedFile, error) {
+	fsi.Lock()
+	defer fsi.Unlock()
```
krishnasrinivas: this lock should be held only accessing the map, else Open will be slow when there are parallel opens.

cmd/fs-rwpool.go
harshavardhana: io.SectionReader doesn't exist.. we have to return something which has ReaderAt.

cmd/fs-rwpool.go
harshavardhana: agreed, done.

cmd/fs-helpers.go
```diff
@@ -16,76 +16,245 @@
...
-			break
+		return err
+	}
+	return nil
+}
+
+func fsRemoveAll(dirPath string) error {
+	err := removeAll(dirPath)
+	if err != nil {
+		if os.IsExist(err) {
```
vadmeste: Does this mean os.IsNotExist ?

cmd/fs-helpers.go
```diff
@@ -16,76 +16,245 @@
...
+		// File path cannot be verified since one of the parents is a file.
+		if isSysErrNotDir(err) {
+			return 0, errFileAccessDenied
 		}
+		return 0, err
+	}
+	defer writer.Close()
+	// Fallocate negative, fall back to maximum object size.
+	if fallocSize < 0 {
+		fallocSize = maxObjectSize
```
vadmeste: Allocating maxObjectSize when we don't know the real size is aggressive because fallocate *does* consume space in the disk in case of Linux and we'll end up allocating much more space than needed.

cmd/fs-helpers.go
```diff
@@ -16,76 +16,245 @@
...
+// Renames source to destination in a protected manner
+// by blocking concurrent requests to not rename destination
+// if there are parallel reads in progress, file system
+// locks are held to manage concurrent access.
+func fsRenameFile(sourcePath, destPath string) error {
+	if err := mkdirAll(path.Dir(destPath), 0777); err != nil {
+		return traceError(err)
+	}
+	if err := os.Rename(preparePath(sourcePath), preparePath(destPath)); err != nil {
+		if rerr := os.Remove(preparePath(sourcePath)); rerr != nil {
```
vadmeste: Not clear why are we removing sourcePath here in case of Rename error

cmd/fs-v1-multipart-common.go
```diff
@@ -17,72 +17,101 @@
...
+
+	"github.com/minio/minio/pkg/lock"
 )
 
 // Returns if the prefix is a multipart upload.
 func (fs fsObjects) isMultipartUpload(bucket, prefix string) bool {
-	_, err := fs.storage.StatFile(bucket, pathJoin(prefix, uploadsJSONFile))
-	return err == nil
+	uploadsJSONPath := pathJoin(bucket, prefix, uploadsJSONFile)
+	_, err := fsStatFile(pathJoin(pathJoin(fs.fsPath, minioMetaMultipartBucket, uploadsJSONPath)))
```
vadmeste: I guess this is a not needed double pathJoin()

cmd/fs-helpers.go
harshavardhana: Usually it's a temporary path. But it's not clear here.

pkg/lock/lock_nix.go
```diff
@@ -0,0 +1,58 @@
...
+	case syscall.O_RDWR:
+		fallthrough
+	case syscall.O_WRONLY | syscall.O_CREAT:
+		fallthrough
+	case syscall.O_RDWR | syscall.O_CREAT:
+		lockType = syscall.LOCK_EX
+	default:
+		return nil, fmt.Errorf("Unsupported flag (%d)", flag)
+	}
+	f, err := os.OpenFile(path, flag|syscall.O_DSYNC, perm)
```
krishnasrinivas: looks like O_DSYNC is not available on freebsd

cmd/fs-rwpool.go
```diff
@@ -0,0 +1,101 @@
...
+// count has reached zero.
+func (fsi *fsIOPool) Open(path string) (*lock.RLockedFile, error) {
+	fsi.Lock()
+	rlkFile, ok := fsi.readersMap[path]
+	fsi.Unlock()
+	// File reference exists on map, validate if its
+	// really closed and we are safe to purge it.
+	if ok {
+		// If the file is closed, remove it from the reader pool map.
+		if rlkFile.IsClosed() {
```
krishnasrinivas, line 53: it should be implemented in such a way that Close() removes the entry from the map immediately when ref is 0 and the file is closed. Else we might end up with a lot of stale entries in this map.

cmd/fs-helpers.go
harshavardhana: Actually its not needed since temp path is cleared by the caller upon error.

cmd/fs-v1-multipart-common.go
harshavardhana: yes its not needed.

cmd/fs-helpers.go
harshavardhana: Yes a typo.

cmd/fs-helpers.go
harshavardhana: Perhaps we can avoid doing fallocate..

harshavardhana:
Adding new tests to increase the coverage.. Please review again @minio/core

cmd/fs-v1-multipart.go
```diff
@@ -247,10 +331,29 @@ func (fs fsObjects) PutObjectPart(bucket, object, uploadID string, partID int, s
...
 	preUploadIDLock.RLock()
+	rwlk, err := fs.rwPool.Open(pathJoin(fs.fsPath, minioMetaMultipartBucket, bucket, object, uploadsJSONFile))
+	if err != nil {
+		preUploadIDLock.RUnlock()
+		if err == errFileNotFound {
+			return "", traceError(InvalidUploadID{UploadID: uploadID})
+		}
+		return "", toObjectErr(traceError(err), bucket, object)
+	}
+	defer rwlk.Close()
```
krishnasrinivas: rwlk.Close() should be done sooner (after preUploadIDLock.RUnlock()) so that parallel upload of parts happen.

the flock should be held again during rename of the part.

cmd/fs-v1-multipart.go
```diff
@@ -334,36 +430,42 @@ func (fs fsObjects) PutObjectPart(bucket, object, uploadID string, partID int, s
...
 		partLock.Unlock()
-		return "", toObjectErr(traceError(err), minioMetaMultipartBucket, partPath)
+		return "", toObjectErr(err, minioMetaMultipartBucket, partPath)
 	}
-	uploadIDPath = path.Join(bucket, object, uploadID)
-	if err = writeFSMetadata(fs.storage, minioMetaMultipartBucket, path.Join(uploadIDPath, fsMetaJSONFile), fsMeta); err != nil {
+	if _, err = fsMeta.WriteTo(metaFile); err != nil {
 		partLock.Unlock()
 		return "", toObjectErr(err, minioMetaMultipartBucket, uploadIDPath)
 	}
```
krishnasrinivas, line 467: fsNSPartPath should be locked and opended and the reader should be passed to fs.append() so that another server does not replace it when the background append is reading from the part file.

cmd/fs-v1-multipart.go
```diff
@@ -474,7 +605,22 @@ func (fs fsObjects) CompleteMultipartUpload(bucket string, object string, upload
...
+		return "", toObjectErr(traceError(err), bucket)
+	}
+
+	uploadIDPath := pathJoin(bucket, object, uploadID)
+
+	rwlk, err := fs.rwPool.Create(pathJoin(fs.fsPath, minioMetaMultipartBucket, bucket, object, uploadsJSONFile))
+	if err != nil {
+		return "", toObjectErr(traceError(err), bucket, object)
+	}
+	defer rwlk.Close()
```
krishnasrinivas: this should be done at the bottom of the function below these lines:
```
        // Hold the lock so that two parallel complete-multipart-uploads                                                                                                        
        // do not leave a stale uploads.json behind.                                                                                                                            
        objectMPartPathLock := globalNSMutex.NewNSLock(minioMetaMultipartBucket,                                                                                                
                pathJoin(bucket, object))                                                                                                                                       
        objectMPartPathLock.Lock()                                                                                                                                              
        defer objectMPartPathLock.Unlock() 
```

cmd/fs-v1-multipart-common.go
```diff
@@ -17,72 +17,74 @@
...
 		return err
 	}
+	uploadIDs.RemoveUploadID(uploadID)
+	if uploadIDs.IsEmpty() {
+		// No more uploads left, so we delete `uploads.json` file.
+		uploadsPath := pathJoin(bucket, object, uploadsJSONFile)
+		multipartPathBucket := pathJoin(fs.fsPath, minioMetaMultipartBucket)
+		uploadPathBucket := pathJoin(multipartPathBucket, uploadsPath)
+		return fsDeleteFile(multipartPathBucket, uploadPathBucket)
+	} // else not empty
```
krishnasrinivas, line 69: should do truncate() here

cmd/fs-v1-multipart-common.go
```diff
@@ -17,72 +17,74 @@
...
-	} else {
-		// no uploads, so we delete the file.
-		if err = fs.storage.DeleteFile(minioMetaMultipartBucket, uploadsPath); err != nil {
-			return toObjectErr(traceError(err), minioMetaMultipartBucket, uploadsPath)
-		}
+	if errorCause(err) == io.EOF {
+		// If we couldn't read anything, we assume a default (empty) upload info.
+		uploadIDs = newUploadsV1("fs")
 	}
+	uploadIDs.AddUploadID(uploadID, initiated)
```
krishnasrinivas, line 87: shoud do a truncate() before write()

cmd/fs-v1-multipart.go
```diff
@@ -19,16 +19,61 @@ package cmd
...
 	"strings"
 	"time"
 
 	"github.com/minio/sha256-simd"
 )
 
+// listMultipartUploadIDs - list all the upload ids from a marker up to 'count'.
+func (fs fsObjects) listMultipartUploadIDs(bucketName, objectName, uploadIDMarker string, count int) ([]uploadMetadata, bool, error) {
+	var uploads []uploadMetadata
+
```
krishnasrinivas, line 38: should hold a NS lock here

cmd/fs-v1-multipart-common.go
harshavardhana: WriteTo already does that

cmd/fs-v1-multipart-common.go
harshavardhana: WriteTo already does that.

cmd/fs-v1-multipart.go
harshavardhana: It is on purpose disallowed upload of parallel parts, want it to block and complete one if them completed. If needed parts can be uploaded to same server.  This optimization we can do later.

cmd/fs-v1-multipart.go
harshavardhana: Since the above case is moot we don't need to do this, another thing is that fsNSPartPath we can't hold a write lock since we write to temporary location. It invites same issues which we tried to fix from an earlier version of this patch.

cmd/fs-v1-multipart.go
harshavardhana: why is that?

cmd/fs-v1-multipart.go
harshavardhana: we never used to hold locks don't know what changed now. Also we have a lock at the parent for 'bucket, object' held.

cmd/fs-rwpool.go
harshavardhana: done

pkg/lock/lock_nix.go
harshavardhana: done

cmd/fs-v1-multipart.go
krishnasrinivas: now since LOCK_SH flock is held (rlk) what can happen is, a competing LOCK_EX in the same minio process will succeed after which if rlk.Close() happens then LOCK_EX will be lost. hence another minio server can hold LOCK_EX but we think we have LOCK_EX. i.e two processes can think that they have LOCK_EX at the same time.

previously there was no scope for this situation to arise.

harshavardhana: But we are holding a lock on the parent wouldn't that be sufficient? why would a EX succeed with in the same process if the parent is locked.. ? EX on `uploads.json` gets only modified when parent is granted..

cmd/fs-v1-multipart.go
harshavardhana: done

cmd/fs-v1-multipart.go
krishnasrinivas: there is a corner case where the part gets replaced by another minio server before `os.OpenFile(preparePath(tmpObjPath), os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0666)` happens in appendPart(). In which case the minio server will append a different part but the ETAG that it remembers will not be the md5sum of the part that it appended.

i think this can be fixed in a different patch as it will complicate this patch

cmd/fs-v1-multipart.go
krishnasrinivas: sequence of lock should always be NSLock and then fcntl lock, else fcntl locks will merge.

for ex.
* fs.rwPool.Open at some other place succeeds
* fs.rwPool.Create succeeds (merge happens) i.e read and write locks are "held" at the same time.
* fd closed got using fs.rwPool.Open closes, this will release the lock held by fs.rwPool.Create which will cause problems.

harshavardhana: Ah that way. Let me see if there are more places which doesn't follow this sequence.

cmd/fs-v1-multipart.go
krishnasrinivas: ah, i see that the the following happens in the caller:

```
                        entryLock := globalNSMutex.NewNSLock(minioMetaMultipartBucket,                                                                                          
                                pathJoin(bucket, entry))                                                                                                                        
                        entryLock.RLock()
```

cmd/fs-v1-multipart.go
krishnasrinivas: not just to allow parallel parts uploads, but also NS lock should always be held when fcntl lock is held. else two fcntl locks will merge.

sequence should be:
NS lock
fcntl lock
fcntl unlock
NS unlock

harshavardhana: Yes agreed.. will do it that way everywhere.

cmd/fs-v1-multipart.go
harshavardhana: Looks like all places that i saw have been addressed with the above style.

harshavardhana:
Closing this PR will open a new.. one so that new comments can be provided. There page has gotten really long.

Issue #3403, opened 2016-12-05
Unusual memory consumption with minio distributed server (XL) 


vadmeste:
@dcharbonnier:
GET and PUT during at least one hour, all the bandwidth that the servers can consume, with ~150 clients connected to 2 clusters of 4 machines, between 500Mbits and 1Gbit and it's a lot of read/write with files of ~50Mo

## Expected Behavior
Normal execution

## Current Behavior
OOM killer killing Minio

## Your Environment
All

vadmeste:
@dcharbonnier, are you already balancing connections? approximately how many clients are concurrently connected to one node in your cluster ?

harshavardhana:
```
[1459571.159608] [ pid ]   uid  tgid total_vm      rss nr_ptes nr_pmds swapents oom_score_adj name
[1459571.159614] [  296]     0   296    12934        1      31       3       80             0 systemd-journal
[1459571.159617] [  334]     0   334    11180       12      22       3      303         -1000 systemd-udevd
[1459571.159620] [  602]   106   602    10725        3      26       3      119          -900 dbus-daemon
[1459571.159622] [  608]   104   608    64099       81      27       3      224             0 rsyslogd
[1459571.159624] [  609]     0   609    69056       45      35       4      262             0 accounts-daemon
[1459571.159627] [  611]     0   611     7137        6      18       3       81             0 systemd-logind
[1459571.159629] [  617]     0   617     7241        7      20       3       65             0 cron
[1459571.159631] [  658]     0   658     3381       27      11       3       51             0 mdadm
[1459571.159633] [  692]     0   692     4868       28      14       3       33             0 irqbalance
[1459571.159635] [  771]     0   771     4030        0      11       3      216             0 dhclient
[1459571.159638] [  810]   108   810   107587        0      59       3     3161             0 named
[1459571.159640] [  817]     0   817    16380       12      36       3      169         -1000 sshd
[1459571.159642] [  829]     0   829     3224        0      11       3       54             0 ntpd
[1459571.159644] [  830]   109   830     3262        0      12       3       81             0 ntpd
[1459571.159646] [  831]   109   831     4303        1      14       3       63             0 ntpd
[1459571.159648] [  836]     0   836     3974        0      13       3       40             0 agetty
[1459571.159651] [30306]     0 30306   225425       31      87       6     6297          -500 dockerd
[1459571.159653] [30340]     0 30340   112458       15      34       6     1415          -500 docker-containe
[...]

PR #3405, opened 2016-12-06
caching: Optimize memory allocations.

harshavardhana:
## Description
This change brings in changes at multiple places

 - Reuse buffers at almost all locations ranging
   from rpc, fs, xl, checksum etc.
 - Change caching behavior to disable itself
   under low memory conditions i.e < 8GB of RAM.
 - Only objects cached are of size 1/10th the size
   of the cache for example if 4GB is the cache
   the maximum object size which will be cached
   is going to be 400MB. This change is an
   optimization to cache more objects rather
   than few larger objects.
 - If object cache is enabled default GC
   percent has been reduced to 20% in lieu
   with newly found behavior of GC. As soon
   as the cache size reaches 75% utilization
   this percentage is changed to 10% to make
   GC more aggressive.
 - Do not use *bytes.Buffer* due to its growth
   requirements. For every allocation *bytes.Buffer*
   allocates an additional buffer for its internal
   purposes. This is undesirable for us, so
   implemented a new cappedWriter which is capped to a
   desired size, beyond this all writes re rejected.

## Motivation and Context
Possible fix for #3403

## How Has This Been Tested?
Tested continuously with large clusters under distributed setup.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/erasure-createfile.go
+++ b/cmd/erasure-createfile.go
@@ -29,7 +29,7 @@ import (
 // all the disks, writes also calculate individual block's checksum
 // for future bit-rot protection.
 func erasureCreateFile(disks []StorageAPI, volume, path string, reader io.Reader, blockSize int64, dataBlocks int, parityBlocks int, algo string, writeQuorum int) (bytesWritten int64, checkSums []string, err error) {
-	// Allocated blockSized buffer for reading.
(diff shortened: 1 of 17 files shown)


harshavardhana:
This patch has been tested under various conditions, taking this in for now as is.

PR #3410, opened 2016-12-06
Fix max cache size calculation when system RAM is inferior to the default cache size

vadmeste:
## Description
objcache: New() panics when max cache size is zero
cache: Set correct maxCacheSize when avail. mem is <= 8Gb 

## Motivation and Context
Found bug after manual testing

## How Has This Been Tested?
go test

## Types of changes

## Checklist:


harshavardhana:
Cache size is never set to infinite in any manner @vadmeste - globalMaxCacheSize always has a real value not 0 or < 0. 

Also the check should be done when objcache.New() is initialized.

Issue #3198, opened 2016-11-07
tests: add a hook for tests to restore global variables to defaults


krishnasrinivas:
Currently the test cases modify the global variables. There is a good chance that test cases modify the global variables and affect other test cases. So far we have fixed this behavior by restoring the global variable  at the beginning of the test cases which gets affected. Easier way is to see if testing framework provides a hook that gets called before every test, so we can set global variables to their defaults when the hook is run.

hackintoshrao:
timeline - 10th November.

hackintoshrao:
In progress.

GarimaKapoor:
@hackintoshrao can you update on the status of this fix.

hackintoshrao:
Will be fixed today.

hackintoshrao:
Couldn't proceed, finishing the tests and reporting the issues. Will finish this weekend.

harshavardhana:
> Couldn't proceed, finishing the tests and reporting the issues. Will finish this weekend.

ping @hackintoshrao are you still working on this?

hackintoshrao:
Yes, will continue on this in couple of days.

GarimaKapoor:
@hackintoshrao any updates on this?

hackintoshrao:
Working on this currently ....

krishnasrinivas:
@hackintoshrao is there a way to trigger resetTestGlobals before each test automatically? like a pre-hook?

hackintoshrao:
@krishnasrinivas : There's no facility to trigger this as a hook before every hook, may be we could have a wrapper and use it for calling the actual tests. Collecting all the globals and resetting them in this PR. Should finish it in couple of days.

harshavardhana:
ping @hackintoshrao - any progress on this?

hackintoshrao:
The function to reset globals variables is added, need to add reset few more variables in it. Will add them by 5th Jan.

hackintoshrao:
Working on it currently.

hackintoshrao:
@krisis  @harshavardhana : Since the mutex to be held on the `globalNsLock` is one of the fileds of the struct, I don't find a way to set `globalNSLock` to `nil` atomically. That would be possible if the mutex lock lives outside the struct, and that would complicate the usage of entire lock usage.

```go
// reset global NSLock.
func resetGlobalNSLock() {
        if globalNSMutex != nil {
        // this is not possible. But Lock is essential for safe implementation.
              globalNSMutex.Lock()
[...]

PR #3411, opened 2016-12-07
Use a non member mutex lock for serverConfig access.

hackintoshrao:
Use a non member mutex lock for serverConfig access.
## Description

 - This is to ensure that the any new config references made to the
      serverConfig is also backed by a mutex lock.
 - Otherwise any new config assigment will also replace the member mutex
      which is currently used for safe access.
## Motivation and Context

- Useful to also guarantee that tests don't influence each other by new assignment of config.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/config-v10.go
+++ b/cmd/config-v10.go
@@ -23,8 +23,11 @@ import (
 	"github.com/minio/minio/pkg/quick"
 )
 
+// Read Write mutex for safe access to ServerConfig.
+var serverConfigMu sync.RWMutex
+
 // serverConfigV10 server configuration version '10' which is like version '9'
-// except it drops support of syslog config
+// except it drops support of syslog config.
 type serverConfigV10 struct {
 	Version string `json:"version"`
 
@@ -37,9 +40,6 @@ type serverConfigV10 struct {
 
 	// Notification queue configuration.
 	Notify notifier `json:"notify"`
-
-	// Read Write mutex.
-	rwMutex *sync.RWMutex
 }
 
 // initConfig - initialize server config and indicate if we are creating a new file or we are just loading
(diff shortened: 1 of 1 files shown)


hackintoshrao:
Part - 1 necessary changes for https://github.com/minio/minio/issues/3198 .

cmd/config-v10.go
```diff
@@ -23,8 +23,11 @@ import (
 	"github.com/minio/minio/pkg/quick"
 )
 
+// Read Write mute for safe access to ServerConfig.
```
donatello: `mutex` instead of `mute`.

donatello (approved):
LGTM - except for typo in comment.

cmd/config-v10.go
```diff
@@ -23,8 +23,11 @@ import (
...
+var severConfigMu sync.RWMutex
```
krisis: typo: s/severConfigMu/serverConfigMu

cmd/config-v10.go
```diff
@@ -112,168 +117,191 @@ var serverConfig *serverConfigV10
 
 // GetVersion get current config version.
 func (s serverConfigV10) GetVersion() string {
-	s.rwMutex.RLock()
-	defer s.rwMutex.RUnlock()
+	severConfigMu.Lock()
```
krisis: This should be a RLock()/RUnlock() here and in other places where members of serverConfigV10 objects are being read.

krisis (changes requested):

cmd/config-v10.go
```diff
@@ -37,13 +40,11 @@ type serverConfigV10 struct {
...
 	// Notification queue configuration.
 	Notify notifier `json:"notify"`
-
-	// Read Write mutex.
-	rwMutex *sync.RWMutex
 }
 
 // initConfig - initialize server config and indicate if we are creating a new file or we are just loading
 func initConfig() (bool, error) {
+
```
harshavardhana: Unnecessary new line.

cmd/config-v10.go
```diff
@@ -68,17 +69,18 @@ func initConfig() (bool, error) {
...
 		// Create config path.
 		err := createConfigPath()
 		if err != nil {
 			return false, err
 		}
-
+		// hold the mutex lock before a new config is assigned.
+		serverConfigMu.Lock()
 		// Save the new config globally.
 		serverConfig = srvCfg
```
harshavardhana: Move this up a bit assign to the variable like this

```
serverConfigMu.Lock()
serverConfig = srvCfg
serverConfigMu.Unlock()
```

cmd/config-v10.go
hackintoshrao: Removing.

cmd/config-v10.go
hackintoshrao: Does it mean that the assignment with the mutex locks should be in 3 consecutive lines?

krisis (approved):

cmd/config-v10.go
harshavardhana: Yes that's why it is readable.

hackintoshrao: done.

PR #3417, opened 2016-12-08
Add service API handler stubs for status, stop and restart

### LINKED ISSUE #3393: api/mgmt: Implement new service management API's (Shutdown, Restart)
## Expected Behavior
Management API to cater for `minio control service` command which was previously removed.

## Current Behavior
No way to restart or shutdown a running cluster of servers.

## Possible Solution

A new API perhaps like this. 

 - Service(type serviceType) 

```
type serviceType enum {
      Shutdown
      Restart
      Status
}
```

## Context
To bring in management functionalities through this API.

krisis:
Add service REST API handlers for status, stop and restart operations. These operations are apply uniformly to single node FS, XL and distributed setup. In a distributed setup, these operations are applied to all the nodes participating in the cluster. i.e a stop command would stop all the minio servers.

## Description

Adds service REST API handling. In a distributed setup, as part of handling the API, other nodes are intimated of the operation via RPC.

## Motivation and Context

These APIs will be consumed by ``mc admin service`` subcommands to manage a minio setup.

## How Has This Been Tested?

A sample go program which makes the corresponding REST calls was used to test, besides unit tests.

## Types of changes

## Checklist:

Fixes #3393


krisis:
This PR still has an issue. A Stop/Restart service REST API call would stop/restart the node to which the request was sent. There is an issue with internal admin RPC registration. I am working on this. Meanwhile reviews can start.

krisis:
The admin RPC issue is fixed. It was an issue with the order in which routes where registered on the mux (gorilla/mux).

cmd/admin-rpc-client.go
```diff
@@ -0,0 +1,144 @@
...
+	args := GenericArgs{}
+	reply := GenericReply{}
+	err := rc.Call("Service.Restart", &args, &reply)
+	if err != nil && err == rpc.ErrShutdown {
+		rc.Close()
+	}
+	return err
+}
+
+type controlPeer struct {
```
harshavardhana: we can avoid the use of "control" word here..

cmd/admin-rpc-client.go
```diff
@@ -0,0 +1,144 @@
...
+	addr    string
+	svcClnt stopRestarter
+}
+type controlPeers []controlPeer
```
harshavardhana: We moved to calling these admin so we should name them as such.

cmd/service.go
```diff
@@ -80,36 +80,37 @@ func (m *ServerMux) handleServiceSignals() error {
 		globalServiceDoneCh <- struct{}{}
 	}
 
-	// Start listening on service signal. Monitor signals.
+	// Wait for SIGTERM in a go-routine.
 	trapCh := signalTrap(os.Interrupt, syscall.SIGTERM)
+	go func(<-chan bool) {
```
harshavardhana, line 85: is there a particular reason to remove select usage?

cmd/admin-handlers_test.go
```diff
@@ -68,7 +68,7 @@ func (c cmdType) toServiceSignal() serviceSignal {
 	return serviceStatus
 }
 
-func serviceSignalReceiver(cmd cmdType, t *testing.T) {
+func mockServiceSignalReceiver(cmd cmdType, t *testing.T) {
```
harshavardhana: we can avoid the name of mock and use "initTestServiceSignalReceiver"

cmd/admin-handlers.go
```diff
@@ -16,42 +16,41 @@
...
 )
 
 func (adminAPI adminAPIHandlers) ServiceStatusHandler(w http.ResponseWriter, r *http.Request) {
-	fmt.Println("status: ", r)
 	adminErr := checkRequestAuthType(r, "", "", "")
 	if adminErr != ErrNone {
 		writeErrorResponse(w, r, adminErr, r.URL.Path)
 		return
 	}
+	// FIXME: Servicestatus command is not yet implemented in handleServiceSignals
```
harshavardhana: Is this implemented?

cmd/admin-handlers.go
```diff
@@ -26,7 +26,7 @@ const (
 	minioAdminOpHeader   = "X-Minio-Operation"
 )
 
-func (api objectAPIHandlers) ServiceStatusHandler(w http.ResponseWriter, r *http.Request) {
+func (adminAPI adminAPIHandlers) ServiceStatusHandler(w http.ResponseWriter, r *http.Request) {
 	fmt.Println("status: ", r)
 	adminErr := checkRequestAuthType(r, "", "", "")
```
harshavardhana: adminAPIErr

cmd/admin-rpc-server.go
```diff
@@ -16,12 +16,20 @@
 
 package cmd
 
+import (
+	"net/rpc"
+
+	router "github.com/gorilla/mux"
+)
+
+const servicePath = "/service"
```
harshavardhana: We can move this to "/admin/service"

harshavardhana (commented):
Initial comments.

cmd/service.go
krisis: The way we were handling os signals required us to have globalServiceSignalCh buffered. Separating in this fashion allows us to use an unbuffered channel. Unbuffered channels are easy to reason with and write deterministic unit tests.

cmd/admin-rpc-client.go
krisis: Yes, I shall change it.

cmd/admin-rpc-client.go
krisis: OK

cmd/admin-rpc-server.go
krisis: OK.

cmd/admin-handlers.go
krisis: done.

cmd/admin-handlers.go
krisis: This is an old comment. It is implemented and the FIXME is removed.

cmd/admin-handlers_test.go
krisis: It is mocking the signal receiver code in service.go and named accordingly. It is not initializing anything so it doesn't make sense naming it so.

docs/service-api.md
```diff
@@ -0,0 +1,33 @@
+# Service REST API
```
harshavardhana: Move this into `docs/admin-api/service.md` ?

cmd/admin-rpc-client.go
```diff
@@ -117,7 +117,7 @@ func initGlobalAdminPeers(eps []*url.URL) {
 	globalAdminPeers = makeControlPeers(eps)
 }
```
harshavardhana, line 135: Let's write comment for each of these functions..

cmd/admin-rpc-client.go
```diff
@@ -65,20 +65,20 @@ func (rc remoteAdminClient) Restart() error {
 	return err
 }
 
-type controlPeer struct {
+type adminPeer struct {
```
harshavardhana, line 80: We should write some comments about the actual usage of these peers and the structure used here.

cmd/admin-handlers_test.go
```diff
@@ -96,6 +96,14 @@ func testServicesCmdHandler(cmd cmdType, t *testing.T) {
 		t.Fatalf("Unable to initialize server config. %s", err)
 	}
 	defer removeAll(rootPath)
+
+	// Initialize admin peers to make admin RPC calls.
+	eps, err := parseStorageEndpoints([]string{"http://localhost"})
```
harshavardhana, line 101: what will be port used here? "9000"?

cmd/admin-rpc-client.go
```diff
@@ -0,0 +1,144 @@
...
+package cmd
+
+import (
+	"net/rpc"
+	"net/url"
+	"path"
+	"sync"
+)
+
+type localAdminClient struct {
```
harshavardhana, line 27: Approach looks good,  add comments..

cmd/admin-rpc-client.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+type controlPeer struct {
+	addr    string
+	svcClnt stopRestarter
+}
+type controlPeers []controlPeer
+
+func makeControlPeers(eps []*url.URL) controlPeers {
+	var ret []controlPeer
```
harshavardhana: what is "ret" ? we could use a non generic name..

cmd/admin-handlers_test.go
harshavardhana: My concern is the name itself, let's not use `mock`..  We can prefix it with  `test`  rather than `mock`.

harshavardhana (commented):
Few more..

docs/service-api.md
krisis: OK.

cmd/admin-handlers_test.go
krisis: parseStorageEndpoints() defaults to 9000 if no port is supplied.

PR #3419, opened 2016-12-08
FS/Multipart: Fix race between PutObjectPart and Complete/Abort multi…

### LINKED ISSUE #3351: fs: TestAPIAbortMultipartHandler sporadically fails under Windows
Below is the log of the failed test: (I added a log to print the real cause of the error)

```
Unable to abort multipart upload. remove \\?\c:\Users\appveyor\AppData\Local\Temp\1\minio-412581580\.minio.sys\multipart\1ks3s3fokor62m02gaeqhc9apmkxwes6qpv933o5fx4pbk9dex5p9mpszry5\test-object-new-multipart\6902e697-4b13-4771-9d3f-489a1703053c\object2: The process cannot access the file because it is being used by another process.
--- FAIL: TestAPIAbortMultipartHandler (0.45s)
	test-utils_test.go:1759: Minio FS: Anonymous HTTP request test fail for "TestAPIAbortMultipartHandler": 
		<Error> Expected the anonymous HTTP request to be served after the policy changes
		,Expected response HTTP status code to be 204, got 500
```

In Windows, I can reproduce this error ```: The process cannot access the file because it is being used by another process.``` with the following code:

```
package main

import "log"
import "os"

func main() {	
	file, err := os.Open("myfile") // For read access.
	if err != nil {
		log.Fatal(err)
	}
	if err := os.Remove("myfile"); err != nil {
		log.Fatal(err)
	}
	file.Close()
}

c:\test>go run test.go
2016/11/25 23:14:08 remove myfile: The process cannot access the file because it is being used by another process.
exit status 1
```

The sample code opens a file and tries to remove it before closing the fd.

So, it looks like the problem is that we are aborting the multipart upload (which means deleting multipart files, https://github.com/minio/minio/blob/master/cmd/fs-v1-multipart.go#L709) while appendPart is still not finished (appendPart is reading from the part and appends to a tmp file, https://github.com/minio/minio/blob/master/cmd/fs-v1-background-append.go#L196 )

What do you think ?

krishnasrinivas:
## Description

Fix race between PutObjectPart and Complete/Abort multipart.

## Motivation and Context

Fixes #3351

## How Has This Been Tested?

`make test`

## Types of changes

## Checklist:

### DIFF
--- a/cmd/fs-v1-multipart.go
+++ b/cmd/fs-v1-multipart.go
@@ -361,9 +361,9 @@ func (fs fsObjects) PutObjectPart(bucket, object, uploadID string, partID int, s
 		return "", toObjectErr(err, minioMetaMultipartBucket, uploadIDPath)
 	}
 
+	// Append the part in background.
+	errCh := fs.bgAppend.append(fs.storage, bucket, object, uploadID, fsMeta)
 	go func() {
-		// Append the part in background.
-		errCh := fs.bgAppend.append(fs.storage, bucket, object, uploadID, fsMeta)
 		// Also receive the error so that the appendParts go-routine does not block on send.
 		// But the error received is ignored as fs.PutObjectPart() would have already
 		// returned success to the client.
--- a/cmd/fs-v1-background-append.go
+++ b/cmd/fs-v1-background-append.go
@@ -161,9 +161,11 @@ func (b *backgroundAppend) appendParts(disk StorageAPI, bucket, object, uploadID
 		case <-info.abortCh:
 			// abort-multipart-upload closed abortCh to end the appendParts go-routine.
 			disk.DeleteFile(minioMetaTmpBucket, uploadID)
+			close(info.timeoutCh) // So that any racing PutObjectPart does not leave a dangling go-routine.
 			return
 		case <-info.completeCh:
 			// complete-multipart-upload closed completeCh to end the appendParts go-routine.
+			close(info.timeoutCh) // So that any racing PutObjectPart does not leave a dangling go-routine.
 			return
 		case <-time.After(appendPartsTimeout):
 			// Timeout the goroutine to garbage collect its resources. This would happen if the client initiates
--- a/cmd/fs-v1-background-append.go
+++ b/cmd/fs-v1-background-append.go
@@ -91,9 +91,9 @@ func (b *backgroundAppend) append(disk StorageAPI, bucket, object, uploadID stri
 // Called on complete-multipart-upload. Returns nil if the required parts have been appended.
 func (b *backgroundAppend) complete(disk StorageAPI, bucket, object, uploadID string, meta fsMetaV1) error {
 	b.Lock()
+	defer b.Unlock()
 	info, ok := b.infoMap[uploadID]
 	delete(b.infoMap, uploadID)
-	b.Unlock()
 	if !ok {
 		return errPartsMissing
 	}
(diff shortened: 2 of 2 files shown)


krisis:
@krishnasrinivas Is this still a test PR? The following scenario could cause a PutObjectPart request to block forever, with these changes.

E.g,
- Abortmultipart req is waiting on abortCh. This is inside the namespace lock.

- Putobjectpart req1 is waiting on inputCh in the go routine spawned by append. This is _outside_ namespace lock.

- appendParts go-routine, spawned by an earlier PutObject request, is assembling incoming parts by looking at multipart volume under .Minio.sys. this is outside the namespace lock.

In this case, AbortMultipart request could get 'selected' by appendParts and make the go-routine return before Putobjectpart req1 does. This will make Putobjectpart req1 to block forever.

krishnasrinivas:
> @krishnasrinivas Is this still a test PR? The following scenario could cause a PutObjectPart request to block forever, with these changes.
> 
> E.g,
> 
> Abortmultipart req is waiting on abortCh. This is inside the namespace lock.
> 
> Putobjectpart req1 is waiting on inputCh in the go routine spawned by append. This is outside namespace lock.
> 
> appendParts go-routine, spawned by an earlier PutObject request, is assembling incoming parts by looking at multipart volume under .Minio.sys. this is outside the namespace lock.
> 
> In this case, AbortMultipart request could get 'selected' by appendParts and make the go-routine return before Putobjectpart req1 does. This will make Putobjectpart req1 to block forever.

The PutObjectPart itself will not block as the append() process is tried on a go-routine but yes it can leave a dangling go-routine because of the race. I pushed a commit, can you confirm if it fixes this situation.

krisis:
>The PutObjectPart itself will not block as the append() process is tried on a go-routine but yes it can leave a dangling go-routine because of the race. I pushed a commit, can you confirm if it fixes this situation.

Running in a different go-routine doesn't change anything. It will still block on ``info.inputCh <- bgAppendPartsInput{...}`` if the appendParts routine is assembling parts as part of a previously 'selected' ``<-info.input.Ch`` case.

cmd/fs-v1-multipart.go
```diff
@@ -361,9 +361,9 @@ func (fs fsObjects) PutObjectPart(bucket, object, uploadID string, partID int, s
 		return "", toObjectErr(err, minioMetaMultipartBucket, uploadIDPath)
 	}
 
+	// Append the part in background.
+	errCh := fs.bgAppend.append(fs.storage, bucket, object, uploadID, fsMeta)
```
krisis, line 365: If [this case](https://github.com/minio/minio/pull/3419/files#diff-835acf3d529848c880629f8de34156ecR81) in the select block executes (i.e, the timeoutCh was closed by a competing AbortMultipartUpload say), then a receive on ``errCh`` would be blocked forever. Further, namespace lock taken on the part wouldn't be released. This would result in a namespace lock leak and a go-routine leak.

krishnasrinivas: But we do `errCh <- errAppendPartsTimeout` on the other side so receive on errCh would not block.

krisis: Yes, you are right. It wouldn't block for the reason I stated. I failed to notice the send on errCh.

krisis: It looks like [when](https://github.com/minio/minio/pull/3419/files#diff-835acf3d529848c880629f8de34156ecR156) ``appendNeeded`` is true and ``appendPart`` function returned ``nil`` nothing is sent on ``errCh``.

krishnasrinivas: note that this happens in a for loop, i.e the for loop breaks only if append not needed or if there was an error in appendPart(), before breaking it does `errCh <-`

krisis: OK.

krisis (approved):
LGTM

harshavardhana:
Can you update description of the PR since it's not a WIP anymore @krishnasrinivas ?

krishnasrinivas:
@harshavardhana WIP was removed yesterday

harshavardhana:
> @harshavardhana WIP was removed yesterday

The description here still says "Test PR to check if it fixes windows race condition."

krishnasrinivas:
> The description here still says "Test PR to check if it fixes windows race condition."

Ah, fixed.

PR #3423, opened 2016-12-09
env: Bring back MINIO_BROWSER env.

### LINKED ISSUE #3422: Please re-instantiate MINIO_BROWSER environment variable to turn off webinterface
## Expected Behavior
The ability to turn off the public webinterface that Minio provides when there is no use for it to be there (for example: in production environments where the only access to Minio is done with S3 API and not manually)

This should be done by re-instantiating the environment variable MINIO_BROWSER=off that was removed in pull request  #3230.

## Current Behavior
The webinterface is always on and cannot be turned off anymore.

## Possible Solution
Re-instantiate the environment variable MINIO_BROWSER again to allow turning off the webinterface.

1. The webinterface is always available.

## Context
Disallow any access to Minio via the webinterface. Our Minio instance is running on a public port (443) and will be accessed by robots and/or users all the time. Since there is no use for it, I'd like to disable it.

## Bonus
Allow to redirect to another user-defined URL when MINIO_BROWSER=off. In requests where you would otherwise show the browser (non API requests) which is now disabled, you could redirect to a webpage of choosing.

## Your Environment
* Version used:  2016-11-26T02:23:47Z
* Operating System and version: Debian Jessie

harshavardhana:
## Description
Set MINIO_BROWSER=off to disable web browser completely.

## Motivation and Context
Fixes #3422 

## How Has This Been Tested?
Manually.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/routers.go
+++ b/cmd/routers.go
@@ -75,6 +75,30 @@ func newObjectLayer(storageDisks []StorageAPI) (ObjectLayer, error) {
 	return objAPI, nil
 }
 
+// Composed function registering routers for only distributed XL setup.
+func registerDistXLRouters(mux *router.Router, srvCmdConfig serverCmdConfig) error {
+	// Register storage rpc router only if its a distributed setup.
+	err := registerStorageRPCRouters(mux, srvCmdConfig)
+	if err != nil {
+		return err
+	}
+
+	// Register distributed namespace lock.
+	err = registerDistNSLockRouter(mux, srvCmdConfig)
+	if err != nil {
+		return err
+	}
+
+	// Register S3 peer communication router.
+	err = registerS3PeerRPCRouter(mux)
+	if err != nil {
+		return err
+	}
+
+	// Register RPC router for web related calls.
+	return registerBrowserPeerRPCRouter(mux)
+}
+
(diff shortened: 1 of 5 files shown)


cmd/routers.go
```diff
@@ -83,34 +107,16 @@ func configureServerHandler(srvCmdConfig serverCmdConfig) (http.Handler, error)
...
-			return nil, err
-		}
+		registerDistXLRouters(mux, srvCmdConfig)
+	}
 
-		// Register distributed namespace lock.
-		err = registerDistNSLockRouter(mux, srvCmdConfig)
-		if err != nil {
+	// Register web router only if its not disabled.
+	if !globalIsBrowserDisabled {
```
krisis: to avoid the above double-negation, when read in English, we can call the variable globalIsBrowserEnabled and initialize it accordingly. These kind of option names confuse users. e.g, nfs.disable option in GlusterFS.

krisis (changes requested):

cmd/routers.go
harshavardhana: Agreed let me move the logic to the `globals.go`

krisis (approved):

PR #3426, opened 2016-12-10
Management API library - Service API

vadmeste:
Just a first attempt of a code that works but probably needs some discussion

### DIFF
--- a/pkg/madmin/client_test.go
+++ b/pkg/madmin/client_test.go
@@ -1,0 +1,24 @@
+/*
+ * Minio Cloud Storage, (C) 2016 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+// Package madmin_test
+package madmin_test
+
+import "testing"
+
+func TestMAdminClient(t *testing.T) {
+}
--- a/pkg/madmin/requests.go
+++ b/pkg/madmin/requests.go
@@ -1,0 +1,480 @@
+/*
+ * Minio Cloud Storage, (C) 2016 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package madmin
+
+import (
+	"bytes"
+	"encoding/base64"
+	"encoding/hex"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"math/rand"
+	"net/http"
+	"net/http/httputil"
+	"net/url"
+	"os"
+	"regexp"
+	"runtime"
+	"strings"
+
+	"github.com/minio/minio-go/pkg/s3signer"
+	"github.com/minio/minio-go/pkg/s3utils"
+)
+
+// Client implements Amazon S3 compatible methods.
+type Client struct {
+	///  Standard options.
+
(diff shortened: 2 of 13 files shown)


pkg/madmin/utils.go
```diff
@@ -0,0 +1,207 @@
...
+}
+
+// Verify if input endpoint URL is valid.
+func isValidEndpointURL(endpointURL string) error {
+	if endpointURL == "" {
+		return ErrInvalidArgument("Endpoint url cannot be empty.")
+	}
+	url, err := url.Parse(endpointURL)
+	if err != nil {
+		return ErrInvalidArgument("Endpoint url cannot have fully qualified paths.")
```
krisis: Isn't this error different from the one below? This means the URL was malformed and not parseable.

krisis (commented):
Initial comments.

pkg/madmin/utils.go
vadmeste: > Isn't this error different from the one below? This means the URL was malformed and not parseable.

Agree

vadmeste:
@harshavardhana, issues were addressed, but I am thinking to add a new package under minio-go/pkg/utils which has some useful functions like isValidEndpointURL, isValidDomain, isValidIP etc, .. what do you think ?

harshavardhana:
> @harshavardhana, issues were addressed, but I am thinking to add a new package under minio-go/pkg/utils which has some useful functions like isValidEndpointURL, isValidDomain, isValidIP etc, .. what do you think ?

@vadmeste if we can make it generic enough but as a thumb rule naming `utils` as a package can cause lot of unrelated functions to be dumped there which is a bad practice for a package itself -  `urlutil` is a better word for this, like httputil package. Deals with anything URL related.

harshavardhana:
Is this still a WIP @vadmeste ?

vadmeste:
> Is this still a WIP @vadmeste ?

No, I've updated the title.

pkg/madmin/client_test.go
```diff
@@ -0,0 +1,29 @@
...
+ *
+ */
+
+// Package madmin_test
+package madmin_test
+
+import "testing"
+
+func TestMAdminClient(t *testing.T) {
+	/* cl, _ := madmin.NewAdminClient("localhost:9000", "minio", "minio123", false)
```
harshavardhana: Can we remove the comment?

pkg/madmin/service-api.go
```diff
@@ -0,0 +1,140 @@
...
+		Type BackendType
+		// Following fields are only meaningful if BackendType is XL.
+		OnlineDisks  int // Online disks during server startup.
+		OfflineDisks int // Offline disks during server startup.
+		ReadQuorum   int // Minimum disks required for successful read operations.
+		WriteQuorum  int // Minimum disks required for successful write operations.
+	}
+}
+
+// ServiceStatus - Call Service Status API
```
harshavardhana: Little more descriptive comment describing the purpose.

pkg/madmin/service-api.go
```diff
@@ -0,0 +1,140 @@
...
+
+	err = json.Unmarshal(respBytes, &storageInfo)
+	if err != nil {
+		return ServiceStatusMetadata{}, err
+	}
+
+	return storageInfo, nil
+}
+
+// ServiceStop - Call Service Stop API
```
harshavardhana: Little more descriptive comment describing the purpose.

pkg/madmin/service-api.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return errors.New("Got " + resp.Status)
+	}
+
+	return nil
+}
+
+// ServiceRestart - Call Service Restart API
```
harshavardhana: Little more descriptive comment describing the purpose.

pkg/madmin/utils.go
```diff
@@ -0,0 +1,175 @@
...
+}
+
+// getEndpointURL - construct a new endpoint.
+func getEndpointURL(endpoint string, secure bool) (*url.URL, error) {
+	if strings.Contains(endpoint, ":") {
+		host, _, err := net.SplitHostPort(endpoint)
+		if err != nil {
+			return nil, err
+		}
+		if !isValidIP(host) && !isValidDomain(host) {
```
harshavardhana: You can use s3utils here..

pkg/madmin/requests.go
```diff
@@ -0,0 +1,482 @@
...
+	accessKeyID string
+	// SecretAccessKey required for authorized requests.
+	secretAccessKey string
+
+	// User supplied.
+	appInfo struct {
+		appName    string
+		appVersion string
+	}
+	endpointURL string
```
harshavardhana: You can use endpointURL to be url.URL here not string. Makes validation easier.

harshavardhana (changes requested):
Everything else looks good.

vadmeste:
All issues are addressed @harshavardhana, also I vendorized s3utils package.

vadmeste:
All issues are addressed @harshavardhana, also I vendorized s3utils and s3signer packages.

harshavardhana (approved):

PR #3428, opened 2016-12-11
presign v2: include resp headers in signature calc

### LINKED ISSUE #3400: sig v2: generated presigned GET doesn't work when `response-content-type` or other is specified
Using boto2 with signature v2, the generated presigned GET url works but it doesn't when a response header is specified. Below is a sample code which is not working, 

`pip install boto` to install boto

```python
from boto.s3.connection import S3Connection
from boto.s3.bucketlistresultset import bucket_lister

conn = S3Connection(aws_access_key_id="minio", aws_secret_access_key="minio123", is_secure=False, port=9000, host="localhost", debug=1, calling_format='boto.s3.connection.OrdinaryCallingFormat', path='/', provider='aws')

bucket = conn.get_bucket('testbucket', validate=True)

for key in bucket_lister(bucket):
    print(key)

url = conn.generate_url(3600, 'GET', bucket='testbucket', key='frag.c', response_headers={'response-content-type': 'application/octet-stream'})
print(url)
```

Opening the generated url gives this:
```
<Error><Code>SignatureDoesNotMatch</Code><Message>The request signature we calculated does not match the signature you provided. Check your key and signing method.</Message><Key></Key><BucketName></BucketName><Resource>/testbucket/frag.c</Resource><RequestId>3L137</RequestId><HostId>3L137</HostId></Error>
```

vadmeste:
## Description
Include response headers when presigning an url using signature v2 algorithm

Following: http://docs.aws.amazon.com/AmazonS3/latest/dev/RESTAuthentication.html#RESTAuthenticationStringToSign

## Motivation and Context
Fixes #3400 

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:

### DIFF
--- a/cmd/signature-v2.go
+++ b/cmd/signature-v2.go
@@ -37,6 +37,7 @@ const (
 // http://docs.aws.amazon.com/AmazonS3/latest/dev/RESTAuthentication.html#RESTAuthenticationStringToSign
 
 // Whitelist resource list that will be used in query string for signature-V2 calculation.
+// The list should be alphabetically sorted
 var resourceList = []string{
 	"acl",
 	"delete",
@@ -47,6 +48,12 @@ var resourceList = []string{
 	"partNumber",
 	"policy",
 	"requestPayment",
+	"response-cache-control",
+	"response-content-disposition",
+	"response-content-encoding",
+	"response-content-language",
+	"response-content-type",
+	"response-expires",
 	"torrent",
 	"uploadId",
 	"uploads",


cmd/signature-v2.go
```diff
@@ -54,6 +54,12 @@ var resourceList = []string{
 	"versioning",
 	"versions",
 	"website",
+	"response-content-type",
+	"response-content-language",
+	"response-expires",
+	"response-cache-control",
+	"response-content-disposition",
+	"response-content-encoding",
```
krishnasrinivas: `resourceList` should be alphabetically sorted.

vadmeste: yes thanks.. I ran make test in parallel and testing already detected that. Will fix it.

PR #3440, opened 2016-12-13
docs: removed ```sh and ```shell for code block.

koolhead17:
removed ```sh and ```shell for code block.
## Description
removed ```sh and ```shell for code block.

## Motivation and Context

to show the code block correctly in docs.minio.io

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,13 +6,13 @@ Minio server is light enough to be bundled with the application stack, similar t
 
 ## Docker Container
 ### Stable
-```sh
+```
 $ docker pull minio/minio
 $ docker run -p 9000:9000 minio/minio server /export
 ```
 
 ### Edge
-```sh
+```
 $ docker pull minio/minio:edge
 $ docker run -p 9000:9000 minio/minio:edge server /export
 ```
@@ -22,7 +22,7 @@ Please visit Minio Docker quickstart guide for more [here](https://docs.minio.io
 ### Homebrew
 Install minio packages using [Homebrew](http://brew.sh/)
 
-```sh
+```
 $ brew install minio
 $ minio server ~/Photos
 ```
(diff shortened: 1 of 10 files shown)


donatello:
This disables shell related syntax highlighting on Github.

On docs.minio.io currently, there is no syntax highlighting for shell blocks:

image

So why is this change being done? Does removing the language hint in the code block fix highlighting in Doctor?

deekoder:
@koolhead17 - We need to remove the highlighting only in 2 places mc quickstart guide and mc complete guide. The big shell block of commands showing all mc options are not highlighted correctly. 
Everything else can have ```sh ``` because it displays correct syntax highlighting.

donatello:
@deekoder I don't believe removing markup is the right approach to solve this problem. The problem here seems to be that the syntax highlighter is not handling shell blocks correctly. 

I tried to do some digging on the problem. The rendering in doctor happens in this line - https://github.com/minio/doctor/blob/master/app/controllers/documents_controller.rb#L35 - using the redcarpet library's default syntax highlighter. This library has a plugin system to use other syntax highlighters - I found one called [Rouge](http://rouge.jneen.net/) and checked that shell blocks are working as intended - see the following screenshot from their demo page:

image


The main thing to note about the highlighting in the above, is that shell key words like `export` and `cd` are highlighted and the use of export in the minio command does not cause `export` to be highlighted as well.


I also found that (luckily) Rouge works with redcarpet too - https://github.com/jneen/rouge#you-can-even-use-it-with-redcarpet

I suggest we patch Doctor to use Rouge - if I was more confident with my Ruby skills, I would have sent a patch too. Do you think we can do this?

koolhead17:
Closing this as it is not required.

PR #3451, opened 2016-12-14
server: checkEndpoints syntax properly.

harshavardhana:
## Description
Same as title.

## Motivation and Context
Cleanup.

## How Has This Been Tested?
Manually using `go test`

## Types of changes

## Checklist:

### DIFF
--- a/cmd/server-main.go
+++ b/cmd/server-main.go
@@ -17,7 +17,6 @@
 package cmd
 
 import (
-	"errors"
 	"fmt"
 	"io/ioutil"
 	"net"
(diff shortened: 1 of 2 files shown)


cmd/server-main.go
```diff
@@ -261,48 +260,36 @@ func isDistributedSetup(eps []*url.URL) bool {
...
+// Check if the endpoints are following expected syntax, i.e
+// valid scheme, valid path across all platforms.
+func checkEndpointsSyntax(eps []*url.URL, disks []string) error {
 	for i, u := range eps {
 		switch u.Scheme {
-		case "", "http", "https":
-			// Scheme is "" for FS and singlenode-XL, hence pass.
+		case "":
+			// "/" is not allowed.
+			if u.Path == "" || u.Path == "/" {
```
vadmeste, line 270: Can we show a better message when '/' is specified? other than Empty disk path.. like `Root path is not allowed`

vadmeste: The reason is when the user runs FS mode => `minio server /`, Empty disk path is not meaningful

harshavardhana: Agreed

krishnasrinivas (commented):
Looks good. One change that can be done is returning traceError() to know the line where error happened.

cmd/server-main.go
```diff
@@ -261,48 +260,36 @@ func isDistributedSetup(eps []*url.URL) bool {
...
+			}
+		case "http", "https":
+			// "http://server1/" is not allowed
+			if u.Path == "" || u.Path == "/" || u.Path == "\\" {
+				return fmt.Errorf("Root path is not allowed : %s (%s)", u.Path, disks[i])
+			}
 		default:
 			if runtime.GOOS == "windows" {
 				// On windows for "C:\export" scheme will be "C"
 				matched, err := regexp.MatchString("^[a-zA-Z]$", u.Scheme)
```
balamurugana, line 281: You could compile the regex only once and reuse it everywhere
```
windowsSchemeRegex := regexp.MustCompile("^[a-zA-Z]$")
```

later

```
windowsSchemeRegex.MatchString(u.Scheme)
```

Another thought. As we use single alphabet check, is it possible to avoid regex?

harshavardhana: It is possible.. send a patch Bala.

PR #3454, opened 2016-12-15
Browser: Added character limit for Bucket Name.

koolhead17:
Currently ``cmd/web-handlers.go`` on Bucket Name option does not mention about  minimum 3 characters limit for  Bucket Name. This PR adds the same. 

## Description
Enhanced the output message to include minimum 3 characters Bucket Name limit. Current error code output does not mention about it.

## Motivation and Context
This PR will make the error output more informative especially when someone selects a Bucket Name with 2 characters. 

## Types of changes

## Checklist:

Name limit.

### DIFF
--- a/cmd/web-handlers.go
+++ b/cmd/web-handlers.go
@@ -805,7 +805,7 @@ func toJSONError(err error, params ...string) (jerr *json2.Error) {
 	case "InvalidBucketName":
 		if len(params) > 0 {
 			jerr = &json2.Error{
-				Message: fmt.Sprintf("Bucket Name %s is invalid. Lowercase letters, period and numerals are the only allowed characters. Bucket Name must be at least 3 characters.", params[0]),
+				Message: fmt.Sprintf("Bucket Name %s is invalid. Lowercase letters, period, numerals are the only allowed with minimum number of 3 characters.", params[0]),
 			}
 		}
 	// Bucket not found custom error message.
--- a/cmd/web-handlers.go
+++ b/cmd/web-handlers.go
@@ -805,7 +805,7 @@ func toJSONError(err error, params ...string) (jerr *json2.Error) {
 	case "InvalidBucketName":
 		if len(params) > 0 {
 			jerr = &json2.Error{
-				Message: fmt.Sprintf("Bucket Name %s is invalid. Lowercase letters, period and numerals are the only allowed characters.", params[0]),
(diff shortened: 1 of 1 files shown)


harshavardhana:
Fix the commit message and the title here to be proper you don't have to say "Code:"  we don't need to mention the obvious.  The thumb rule is reference the component which is being modified choose

`web:`  or `browser:` 

In a commit title do not include more than 60 characters, more information if needed goes into the body of the commit itself. Title should be concise and meaningful.

koolhead17:
@harshavardhana modified as suggested.

harshavardhana:
Not just in the PR also in the commit message

cmd/web-handlers.go
```diff
@@ -805,7 +805,7 @@ func toJSONError(err error, params ...string) (jerr *json2.Error) {
 	case "InvalidBucketName":
 		if len(params) > 0 {
 			jerr = &json2.Error{
-				Message: fmt.Sprintf("Bucket Name %s is invalid. Lowercase letters, period and numerals are the only allowed characters.", params[0]),
+				Message: fmt.Sprintf("Bucket Name %s is invalid. Lowercase letters, period and numerals are the only allowed characters. Bucket Name must be at least 3 characters.", params[0]),
```
harshavardhana: "Must be atleast 3 characters long" No need to add "Bucket name" again.

harshavardhana (changes requested):

cmd/web-handlers.go
koolhead17: ok. let me modify and send it.

harshavardhana (approved):

cmd/web-handlers.go
```diff
@@ -805,7 +805,7 @@ func toJSONError(err error, params ...string) (jerr *json2.Error) {
 	case "InvalidBucketName":
 		if len(params) > 0 {
 			jerr = &json2.Error{
-				Message: fmt.Sprintf("Bucket Name %s is invalid. Lowercase letters, period and numerals are the only allowed characters. Bucket Name must be at least 3 characters.", params[0]),
+				Message: fmt.Sprintf("Bucket Name %s is invalid. Lowercase letters, period, numerals are the only allowed with minimum number of 3 characters.", params[0]),
```
harshavardhana: Let's fix this further to

"Bucket Name %s is invalid. Lowercase letters, period, numerals are the only allowed characters and should be minimum 3 characters in length."

koolhead17: sure. just made the modification.

PR #3456, opened 2016-12-15
Remove regexp usage

balamurugana:
This patch removes regexp usage in cacheControlHandler.ServeHTTP() and
server-main.checkEndpointsSyntax()

## How Has This Been Tested?

`make test`

## Types of changes

## Checklist:

### DIFF
--- a/cmd/generic-handlers.go
+++ b/cmd/generic-handlers.go
@@ -19,7 +19,6 @@ package cmd
 import (
 	"net/http"
 	"path"
-	"regexp"
 	"strings"
 	"time"
 
@@ -142,21 +141,18 @@ func setBrowserCacheControlHandler(h http.Handler) http.Handler {
 func (h cacheControlHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	if r.Method == "GET" && guessIsBrowserReq(r) && globalIsBrowserEnabled {
 		// For all browser requests set appropriate Cache-Control policies
-		match, err := regexp.Match(reservedBucket+`/([^/]+\.js|favicon.ico)`, []byte(r.URL.Path))
-		if err != nil {
-			errorIf(err, "Unable to match incoming URL %s", r.URL)
-			writeErrorResponse(w, r, ErrInternalError, r.URL.Path)
-			return
-		}
-		if match {
-			// For assets set cache expiry of one year. For each release, the name
-			// of the asset name will change and hence it can not be served from cache.
-			w.Header().Set("Cache-Control", "max-age=31536000")
-		} else if strings.HasPrefix(r.URL.Path, reservedBucket+"/") {
-			// For non asset requests we serve index.html which will never be cached.
-			w.Header().Set("Cache-Control", "no-store")
+		if strings.HasPrefix(r.URL.Path, reservedBucket+"/") {
+			if strings.HasSuffix(r.URL.Path, ".js") || r.URL.Path == reservedBucket+"/favicon.ico" {
+				// For assets set cache expiry of one year. For each release, the name
+				// of the asset name will change and hence it can not be served from cache.
+				w.Header().Set("Cache-Control", "max-age=31536000")
+			} else {
+				// For non asset requests we serve index.html which will never be cached.
+				w.Header().Set("Cache-Control", "no-store")
+			}
 		}
 	}
+
 	h.handler.ServeHTTP(w, r)
 }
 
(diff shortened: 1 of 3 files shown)


krisis (approved):

cmd/generic-handlers.go
```diff
@@ -139,24 +139,23 @@ func setBrowserCacheControlHandler(h http.Handler) http.Handler {
...
+		if strings.HasPrefix(r.URL.Path, reservedBucket+"/") {
+			if validFaviconRegex.MatchString(r.URL.Path[len(reservedBucket)+1:]) {
+				// For assets set cache expiry of one year. For each release, the name
+				// of the asset name will change and hence it can not be served from cache.
+				w.Header().Set("Cache-Control", "max-age=31536000")
+			} else {
+				// For non asset requests we serve index.html which will never be cached.
+				w.Header().Set("Cache-Control", "no-store")
+			}
 		}
```
donatello, line 153: The `if strings.HasPrefix(..` block needs an else clause that returns `ErrInternalError` to be same as the previous code. Why is it not needed?

balamurugana: `regexp.Match()` may return `err` due to invalid `pattern` hence `ErrInternalError`
As new `validFaviconRegex` is `regexp.MustCompile()`, its guarantied that no error would occur than panic.

So the fix does expected thing.

cmd/generic-handlers.go
```diff
@@ -139,24 +139,23 @@ func setBrowserCacheControlHandler(h http.Handler) http.Handler {
 	return cacheControlHandler{h}
 }
 
+var validFaviconRegex = regexp.MustCompile(`([^/]+\.js|favicon.ico)`)
```
krishnasrinivas: inludes js files also, not just favicon so the name `validFaviconRegex` is not correct

balamurugana: Yes. But thinking why we need to use regex here

harshavardhana: It is needed to add "Cache-Control" for .js and `favicon.ico` , we shouldn't do that for  `.html` ..

cmd/server-main.go
```diff
@@ -260,6 +260,8 @@ func isDistributedSetup(eps []*url.URL) bool {
 	return false
 }
 
+var windowsDriveMatch = regexp.MustCompile("^[a-zA-Z]$")
```
harshavardhana: Add a comment here, also i thought we are going to explore a non-regex approach here..

cmd/server-main.go
```diff
@@ -260,6 +259,8 @@ func isDistributedSetup(eps []*url.URL) bool {
 	return false
 }
 
+const alphabets = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"
+
```
donatello, line 294: It is faster and possibly more readable to use a function - 

```go
func IsLetter(string s) bool {
    return len(s) == 1 && ((s[0] >= 'a' && s[0] <= 'z') ||
         (s[0] >= 'A' && s[0] <= 'Z'))
}
```

`strings.Contains()` is significantly more expensive than the above because this is a 52 length constant.

cmd/generic-handlers.go
donatello: :+1:

cmd/server-main.go
harshavardhana: IsDriveLetter() is better..

balamurugana: Done. without a function.

donatello (approved):

cmd/server-main.go
```diff
@@ -277,18 +276,16 @@ func checkEndpointsSyntax(eps []*url.URL, disks []string) error {
...
-				if matched {
-					break
+				// On Windows, URL.Scheme is "C" for endpoint "C:\export" and
+				// error out if URL.Scheme is not a single letter alphabet to represent a drive.
+				if !(len(u.Scheme) == 1 && ((u.Scheme[0] >= 'a' && u.Scheme[0] <= 'z') ||
+					(u.Scheme[0] >= 'A' && u.Scheme[0] <= 'Z'))) {
+					return fmt.Errorf("Invalid scheme : %s (%s)", u.Scheme, disks[i])
 				}
 			}
-			return fmt.Errorf("Invalid scheme : %s (%s)", u.Scheme, disks[i])
```
krishnasrinivas, line 289: this line is needed ... for ex. to error out on `ftp://server1/path1`

harshavardhana: We should add a test for this..

balamurugana: Done

PR #3458, opened 2016-12-15
signv2: Do not use path encoding for query values.

harshavardhana:
## Description
Use query escape instead for comparing signature.

## Motivation and Context
Reported by community member @bsideup

## How Has This Been Tested?
Manually and with help from @vadmeste 

## Types of changes

## Checklist:

### DIFF
--- a/cmd/signature-v2.go
+++ b/cmd/signature-v2.go
@@ -22,6 +22,7 @@ import (
 	"encoding/base64"
 	"fmt"
 	"net/http"
+	"net/url"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


bsideup:
I can test this change once it's merged and deployed as `edge` tag :+1:

bsideup:
@harshavardhana actually, just checked locally by building an image and it works! 💯

cmd/signature-v2.go
```diff
@@ -128,12 +130,13 @@ func doesPresignV2SignatureMatch(r *http.Request) APIErrorCode {
...
 	}
 
+	// Check if the presigned URL has expired.
 	if expiresInt < time.Now().UTC().Unix() {
 		return ErrExpiredPresignRequest
 	}
 
 	expectedSignature := preSignatureV2(r.Method, encodedResource, strings.Join(filteredQueries, "&"), r.Header, expires)
-	if gotSignature != getURLEncodedName(expectedSignature) {
+	if gotSignature != url.QueryEscape(expectedSignature) {
```
vadmeste: If the client is free to encode or not to encode query values, this change will be wrong: The client can send here a raw '/' without escaping it and this comparaison line won't work. So, if we want to support both escaped and raw query values, we can use url.QueryUnescape(gotSignature) instead.

krishnasrinivas: http://docs.aws.amazon.com/general/latest/gr/signature-version-2.html says:

`The resulting signature must be base-64 encoded and then URI encoded.`

so this looks fine

krishnasrinivas (approved):

vadmeste (approved):

PR #3459, opened 2016-12-15
tests: Use testTmpDir var to specify tmp directory

vadmeste:
## Description
To be able to specify the directory where tests will be done.  This way,
it will be easier to run Minio tests on a mounted directory like NFS, ..

## Motivation and Context
Testing go test on NFS

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:

### DIFF
--- a/cmd/test-utils_test.go
+++ b/cmd/test-utils_test.go
@@ -133,6 +133,9 @@ const (
 var randN uint32
 var randmu sync.Mutex
 
+// Temp files created in default Tmp dir
+var globalTestTmpDir = os.TempDir()
+
 // reseed - returns a new seed every time the function is called.
 func reseed() uint32 {
 	return uint32(time.Now().UnixNano() + int64(os.Getpid()))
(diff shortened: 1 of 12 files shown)


cmd/test-utils_test.go
```diff
@@ -133,6 +133,9 @@ const (
 var randN uint32
 var randmu sync.Mutex
 
+// Temp files created in default Tmp dir
+var testTmpDir = os.TempDir()
```
harshavardhana: name this as 'globalTestTmpDir'

harshavardhana (changes requested):
How do you specify NFS directory?

vadmeste:
> How do you specify NFS directory?

Currently no way, I am thinking of adding a function which returns the content of an environment variable if non nil or default value

harshavardhana (approved):

harshavardhana:
Don't bring any new environment variables. See if go test can take custom cli options.

PR #3467, opened 2016-12-16
postForm: Follow success_action_status requirement

### LINKED ISSUE #3464: Support success_action_status header
## Expected Behavior
Minio should return a status code equal to the `success_action_status` header during a POST or PUT request.

## Current Behavior
It always returns a 204.

Make a POST request to create any object with the `success_action_status` header set to 200 and see the response status code is set to 204. Haven't tested with a PUT but assume it will work the same as a POST request.

## Context
Trying to use FineUploader and it seems to only accept a 200 response code as a success.

## Your Environment
* Version used: RELEASE.2016-12-13T17-19-42Z
* Environment name and version (e.g. nginx 1.9.1): docker

vadmeste:
## Description
S3 spec requires that Post Object response depends on the passed success_action_status. This commit implements that requirement.

## Motivation and Context
Fixes #3464 

## How Has This Been Tested?
go test and manually

## Types of changes

## Checklist:

### DIFF
--- a/cmd/bucket-handlers.go
+++ b/cmd/bucket-handlers.go
@@ -448,11 +448,22 @@ func (api objectAPIHandlers) PostPolicyBucketHandler(w http.ResponseWriter, r *h
 	w.Header().Set("ETag", "\""+objInfo.MD5Sum+"\"")
 	w.Header().Set("Location", getObjectLocation(bucket, object))
 
-	// Set common headers.
-	setCommonHeaders(w)
+	// Decide what http response to send depending on success_action_status parameter
+	switch formValues[http.CanonicalHeaderKey("success_action_status")] {
+	case "201":
+		resp := encodeResponse(PostResponse{
+			Bucket:   bucket,
+			Key:      object,
+			ETag:     "\"" + objInfo.MD5Sum + "\"",
+			Location: getObjectLocation(bucket, object),
+		})
+		writeResponse(w, http.StatusCreated, resp)
 
-	// Write successful response.
-	writeSuccessNoContent(w)
+	case "200":
+		writeSuccessResponse(w, nil)
+	default:
+		writeSuccessNoContent(w)
+	}
 
 	// Notify object created event.
 	eventNotify(eventData{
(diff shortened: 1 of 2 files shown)


cmd/bucket-handlers.go
```diff
@@ -448,11 +448,22 @@ func (api objectAPIHandlers) PostPolicyBucketHandler(w http.ResponseWriter, r *h
 	w.Header().Set("ETag", "\""+objInfo.MD5Sum+"\"")
 	w.Header().Set("Location", getObjectLocation(bucket, object))
 
-	// Set common headers.
-	setCommonHeaders(w)
+	// Decide what http response to send depending on success_action_status parameter
+	switch formValues["Success_action_status"] {
```
harshavardhana: Shouldn't "S" be small case here?

harshavardhana (commented):
Only one comment.

cmd/bucket-handlers.go
```diff
@@ -448,11 +448,22 @@ func (api objectAPIHandlers) PostPolicyBucketHandler(w http.ResponseWriter, r *h
...
+	case "201":
+		resp := encodeResponse(PostResponse{
```
harshavardhana, line 454: Don't use unkeyed values place the fields properly .Since bucket and object are of string type it would be hard if they get changed and we wouldn't know. Subtle bugs like this can arise.

harshavardhana (changes requested):
One more.

cmd/bucket-handlers.go
vadmeste: > Don't use unkeyed values place the fields properly .Since bucket and object are of string type it would be hard if they get changed and we wouldn't know. Subtle bugs like this can arise.

yes, you are right. I fixed it.

cmd/bucket-handlers.go
vadmeste: Currently, all fields in formValues have canonical names that's why it starts by a capital S.

harshavardhana (approved):

cmd/bucket-handlers.go
harshavardhana: Perhaps you can do http.Canonical so that it gets easier to understand the purpose.

PR #3469, opened 2016-12-17
server: Sort disk arguments for consistent ordering

### LINKED ISSUE #3425: cli: validate if the order of arguments are consistent across all nodes for the first time.
## Expected Behavior
Order of arguments are validated properly across all nodes for the first time.

## Current Behavior
Currently if the order of arguments are given wrong for the first time it can end up corrupting the data and create a wrong backend format.

## Possible Solution
Validate the args early on before even proceeding to generate `format` configs.

Quite easy start fresh servers with incorrect orders across nodes, as soon as the nodes `format` themselves they fail with inconsistent `format.json`. 

## Context
This is required since it is a common mistake and we should error out early on.

krisis:
This is important in a distributed setup, where the server hosting the
first disk formats a fresh setup. Sorting ensures that all servers
arrive at the same 'first' server.

Note: This change doesn't protect against different disk arguments
with some disks being same across servers.

## Description

This change implements the sort interface for url.URL type such that disk arguments are sorted based on concatenated Host (incl. port too) and Path. The sorted disk arguments ensure consistent order across servers in a distributed setup.
## Motivation and Context

Fixes #3425 
## How Has This Been Tested?

Manually tested using steps described in #3425 
## Types of changes

## Checklist:

### DIFF
--- a/cmd/url-sort.go
+++ b/cmd/url-sort.go
@@ -1,0 +1,34 @@
+/*
+ * Minio Cloud Storage, (C) 2016 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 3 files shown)


cmd/url-sort.go
```diff
@@ -0,0 +1,36 @@
...
+
+func (s orderedURLs) Swap(i, j int) {
+	s[i], s[j] = s[j], s[i]
+}
+
+func (s orderedURLs) Len() int {
+	return len(s)
+}
+
+type byHostPath struct{ orderedURLs }
```
harshavardhana: You don't need two structs.

krisis: Done.

harshavardhana:
You might have to rebase this. @krisis

harshavardhana:
Arch reviewed, merging.

PR #3471, opened 2016-12-18
rpcclient: fix leaky connection

balamurugana:
Previously, more than one goroutine calls RPCClient.dial(), each
goroutine gets a new rpc.Client but only one such client is stored
into RPCClient object.  This leads to leaky connection at the server
side.  This is fixed by taking lock at top of dial() and release on
return.

## How Has This Been Tested?

`make test`

## Types of changes

## Checklist:


harshavardhana:
LGTM

PR #3420, opened 2016-12-08
replace blake2b implementation

aead:
Replace the blake2b implementation

## Description
Replace the minio/blake2b-simd implementation with the implementation of golang/x/crypto/blake2b

## Motivation and Context
The implementation at golang.org/x/crypto is faster in my benchmarks

## How Has This Been Tested?
I ran the tests in the /cmd and the /pkg/*
Tests were ran under linux/amd64 with and without AVX2

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #3481, opened 2016-12-20
replace blake2b implementation

aead:
replace the blake2b-simd with the golang/x/crypto implementation

```
name        old time/op    new time/op     delta
Size64-8       715ns ±13%      614ns ± 3%    ~     (p=0.084 n=6+6)
Size128-8      612ns ± 5%      634ns ± 8%    ~     (p=0.084 n=6+6)
Size1K-8      2.18µs ± 5%     2.09µs ± 7%    ~     (p=0.084 n=6+6)
Size8K-8      13.1µs ± 2%     13.4µs ± 3%    ~     (p=0.084 n=6+6)
Size32K-8     48.5µs ± 1%     49.5µs ± 3%    ~     (p=0.775 n=6+6)
Size128K-8     199µs ± 0%      198µs ± 3%    ~     (p=0.468 n=6+6)

name        old speed      new speed       delta
Size64-8    92.6MB/s ±11%  104.2MB/s ± 3%    ~     (p=0.139 n=6+6)
Size128-8    208MB/s ± 6%    202MB/s ± 8%    ~     (p=0.102 n=6+6)
Size1K-8     466MB/s ± 7%    492MB/s ± 7%    ~     (p=0.139 n=6+6)
Size8K-8     621MB/s ± 2%    610MB/s ± 3%    ~     (p=0.102 n=6+6)
Size32K-8    672MB/s ± 2%    669MB/s ± 1%    ~     (p=0.818 n=6+6)
Size128K-8   657MB/s ± 1%    672MB/s ± 0%  +2.28%  (p=0.002 n=6+6)

name        old time/op   new time/op   delta
Size64-4      334ns ± 1%    243ns ± 0%  -27.14%  (p=0.029 n=4+4)
Size128-4     296ns ± 1%    242ns ± 0%  -18.21%  (p=0.029 n=4+4)
Size1K-4     1.44µs ± 0%   1.28µs ± 0%  -10.83%  (p=0.029 n=4+4)
Size8K-4     10.0µs ± 0%    9.4µs ± 0%   -6.23%  (p=0.029 n=4+4)
Size32K-4    39.8µs ± 1%   37.3µs ± 0%   -6.31%  (p=0.029 n=4+4)
Size128K-4    162µs ± 3%    149µs ± 0%   -7.72%  (p=0.029 n=4+4)

name        old speed     new speed     delta
Size64-4    192MB/s ± 1%  263MB/s ± 0%  +37.24%  (p=0.029 n=4+4)
Size128-4   431MB/s ± 0%  526MB/s ± 0%  +22.04%  (p=0.029 n=4+4)
Size1K-4    713MB/s ± 0%  800MB/s ± 0%  +12.17%  (p=0.029 n=4+4)
Size8K-4    815MB/s ± 0%  869MB/s ± 0%   +6.64%  (p=0.029 n=4+4)
Size32K-4   823MB/s ± 1%  878MB/s ± 0%   +6.72%  (p=0.029 n=4+4)
Size128K-4  810MB/s ± 3%  877MB/s ± 0%   +8.23%  (p=0.029 n=4+4)
```
See: https://go-review.googlesource.com/#/c/34319/

## Description
replace the blake2b-simd implementation with the 'offical' golang implementation 

## Motivation and Context
The x/crypto/blake2b implementation is faster - see benchmarks

## How Has This Been Tested?
Run all test in /cmd and /pkg

## Types of changes

## Checklist:


aead:
This is a duplicate of  #3420 - as soon as this PR is merged, the PR #3420 should probably be closed

cmd/erasure-utils.go
```diff
@@ -40,11 +41,19 @@ func newHashWriters(diskCount int, algo string) []hash.Hash {
 func newHash(algo string) hash.Hash {
 	switch algo {
 	case "blake2b":
-		return blake2b.New512()
+		h, err := blake2b.New512(nil)
+		if err != nil {
+			panic(fmt.Sprintf("impossible situation: %s", err))
```
harshavardhana: Why is this returning an error? seems like all hashing `New()` never returns an error. any specific reason?

harshavardhana (commented):
One question.

cmd/erasure-utils.go
harshavardhana: Also lets not do `fmt.Sprintf` here, we should be little more descriptive. 

```
panic(errUnexpected)
```

aead: > Why is this returning an error? seems like all hashing New() never returns an error. any specific reason?

The blake2b implementation is based on [RFC 7693](https://tools.ietf.org/html/rfc7693) - blake2b can be used as a MAC, but the key must be between 0 and 64 bytes long - if it's longer, the New* functions will return an non-nil error.
But your right, the code should never return an error as long as no key is passed.

> Also lets not do fmt.Sprintf here, we should be little more descriptive.

Maybe something like: "used blake2b as MAC with an invalid key" or is ignoring the error the better behavior?

harshavardhana: > Maybe something like: "used blake2b as MAC with an invalid key" or is ignoring the error the better behavior?

We can do this as well use blake2b.New512() from crypto if it returns an error , fall back to blake2b-simd project.

aead: > We can do this as well use blake2b.New512() from crypto if it returns an error , fall back to blake2b-simd project. 

The line `h, err := blake2b.New512(nil)` can never fail. 
All `New` functions in the blake2b package only return an error, if a key is passed (this means: instead of using blake2b as hash function, use it as MAC, similar to HMAC)
But passing a key in this context (minio uses blake2b as hash function) would be a compatibility breaking change.

Maybe this example shows the implications:
```
// this never fails, because we use blake2b as a hash function
// theoretically the error can be ignored
h, err := blake2b.New512(nil)  

key := make([]byte, 32)
h, err := blake2b.New512(key)  // this line doesn't fail, but if somebody do:

key := make([]byte, 65)
h, err := blake2b.New512(key) // this line fails, because we use blake2b as MAC and the key is to long ( > 64 bytes)
```
I added the error checking just because I don't think completely ignoring errors is good style, but in this case it would be okay...
So blake2b-simd implementation would be in the repo, but never used...

harshavardhana: Then i guess its fine.. we can ignore and add a comment why we are ignoring and why it wouldn't fail ever.

PR #3486, opened 2016-12-21
fs/xl: Simplify bucket metadata reading.

harshavardhana:
## Description
ObjectLayer GetObject() now returns the entire object
if starting offset is 0 and length is negative. This
also allows to simplify handler layer code where
we always had to use GetObjectInfo() before proceeding
to read bucket metadata files example: `policy.json`.

This also reduces one additional call overhead.

## Motivation and Context
Simplification.

## How Has This Been Tested?
Manually.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/fs-v1.go
+++ b/cmd/fs-v1.go
@@ -222,8 +222,8 @@ func (fs fsObjects) GetObject(bucket, object string, offset int64, length int64,
 	if err = checkGetObjArgs(bucket, object); err != nil {
 		return err
 	}
-	// Offset and length cannot be negative.
-	if offset < 0 || length < 0 {
+	// Offset cannot be negative.
+	if offset < 0 {
 		return toObjectErr(traceError(errUnexpected), bucket, object)
 	}
 	// Writer cannot be nil.
@@ -237,12 +237,13 @@ func (fs fsObjects) GetObject(bucket, object string, offset int64, length int64,
 		return toObjectErr(traceError(err), bucket, object)
 	}
 
-	// Reply back invalid range if the input offset and length fall out of range.
-	if offset > fi.Size || length > fi.Size {
-		return traceError(InvalidRange{offset, length, fi.Size})
+	// For negative length we read everything.
+	if length < 0 {
+		length = fi.Size - offset
 	}
-	// Reply if we have inputs with offset and length falling out of file size range.
-	if offset+length > fi.Size {
+
+	// Reply back invalid range if the input offset and length fall out of range.
+	if offset > fi.Size || offset+length > fi.Size {
 		return traceError(InvalidRange{offset, length, fi.Size})
 	}
 
@@ -291,6 +292,7 @@ func (fs fsObjects) GetObject(bucket, object string, offset int64, length int64,
 			break
 		}
 	}
+
 	// Returns any error.
 	return toObjectErr(err, bucket, object)
 }
--- a/cmd/xl-v1-object.go
+++ b/cmd/xl-v1-object.go
@@ -53,10 +53,12 @@ func (xl xlObjects) GetObject(bucket, object string, startOffset int64, length i
 	if err := checkGetObjArgs(bucket, object); err != nil {
 		return err
 	}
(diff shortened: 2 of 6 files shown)


cmd/xl-v1-object.go
```diff
@@ -88,6 +90,11 @@ func (xl xlObjects) GetObject(bucket, object string, startOffset int64, length i
 	// Reorder parts metadata based on erasure distribution order.
 	metaArr = getOrderedPartsMetadata(xlMeta.Erasure.Distribution, metaArr)
 
+	// For offset '0' and negative length read everything.
+	if startOffset == 0 && length < 0 {
```
krisis: what happens when ``startOffset != 0`` and ``length == -1`` ?

cmd/fs-v1.go
```diff
@@ -237,10 +237,16 @@ func (fs fsObjects) GetObject(bucket, object string, offset int64, length int64,
 		return toObjectErr(traceError(err), bucket, object)
 	}
 
+	// For offset '0' and negative length read everything.
+	if offset == 0 && length < 0 {
+		length = fi.Size
+	}
+
 	// Reply back invalid range if the input offset and length fall out of range.
 	if offset > fi.Size || length > fi.Size {
```
donatello: We can remove the following `if offset+length > fi.Size {` block if we just modify this condition to `if offset > fi.Size || offset + length > fi.Size {`.

Also we seem to be missing a check for an error case where `offset > 0 && length < 0` before (or even in) this if block.

donatello (commented):
Seems to be missing some corner cases.

cmd/xl-v1-object.go
harshavardhana: We should error out.

cmd/fs-v1.go
harshavardhana: We have to error out. Will add a if condition early on.

harshavardhana:
Actually we don't need to error out for offset > 0 and length < 0 , we can just read from the offset and the length upto the size of the object.

harshavardhana:
ping @krisis @donatello

cmd/fs-v1.go
```diff
@@ -237,12 +237,13 @@ func (fs fsObjects) GetObject(bucket, object string, offset int64, length int64,
 		return toObjectErr(traceError(err), bucket, object)
 	}
 
-	// Reply back invalid range if the input offset and length fall out of range.
-	if offset > fi.Size || length > fi.Size {
-		return traceError(InvalidRange{offset, length, fi.Size})
+	// For negative length we read everything.
+	if length < 0 {
+		length = fi.Size
```
donatello: Right side should be `fi.Size - offset` otherwise the next if block will wrongly error if the offset is positive.

cmd/xl-v1-object.go
```diff
@@ -88,13 +90,13 @@ func (xl xlObjects) GetObject(bucket, object string, startOffset int64, length i
 	// Reorder parts metadata based on erasure distribution order.
 	metaArr = getOrderedPartsMetadata(xlMeta.Erasure.Distribution, metaArr)
 
-	// Reply back invalid range if the input offset and length fall out of range.
-	if startOffset > xlMeta.Stat.Size || length > xlMeta.Stat.Size {
-		return traceError(InvalidRange{startOffset, length, xlMeta.Stat.Size})
+	// For negative length read everything.
+	if length < 0 {
+		length = xlMeta.Stat.Size
```
donatello: Same problem here.

donatello (changes requested):
Still appears to have bugs.

cmd/fs-v1.go
harshavardhana: Ah you are right..

cmd/xl-v1-object.go
harshavardhana: Done

harshavardhana:
A test is added as well to cover those changes. @donatello

donatello (approved):

PR #3487, opened 2016-12-21
objAPI: Implement CopyObject API at object layer.

### LINKED ISSUE #3316: api: CopyObject should support updating object metadata.
In order to update metadata from the AWS Java API you must use a copyobject call where the source and the destination object are the same but the metadata is different. This opperation fails with a " This copy request is illegal because it is trying to copy an object to itself" error when using minio
## Expected Behavior
Copy of an object over itself should not throw an exception. Especially if the metadata has changed.

## Current Behavior
Fails with 
2016-11-22 12:53:26,856 [sdfs] [org.opendedup.sdfs.filestore.cloud.BatchAwsS3ChunkStore$DeleteObject] [2509] [pool-10-thread-3]  - Unable to delete object -6061387925563889599
com.amazonaws.services.s3.model.AmazonS3Exception: This copy request is illegal because it is trying to copy an object to itself. (Service: Amazon S3; Status Code: 400; Error Code: InvalidRequest; Request ID: 3L137), S3 Extended Request ID: 3L137
        at com.amazonaws.http.AmazonHttpClient.handleErrorResponse(AmazonHttpClient.java:1341)
        at com.amazonaws.http.AmazonHttpClient.executeOneRequest(AmazonHttpClient.java:959)
        at com.amazonaws.http.AmazonHttpClient.executeHelper(AmazonHttpClient.java:736)
        at com.amazonaws.http.AmazonHttpClient.doExecute(AmazonHttpClient.java:487)
        at com.amazonaws.http.AmazonHttpClient.executeWithTimer(AmazonHttpClient.java:448)
        at com.amazonaws.http.AmazonHttpClient.execute(AmazonHttpClient.java:397)
        at com.amazonaws.http.AmazonHttpClient.execute(AmazonHttpClient.java:378)
        at com.amazonaws.services.s3.AmazonS3Client.invoke(AmazonS3Client.java:4039)
        at com.amazonaws.services.s3.AmazonS3Client.copyObject(AmazonS3Client.java:1714)
        at org.opendedup.sdfs.filestore.cloud.BatchAwsS3ChunkStore$DeleteObject.run(BatchAwsS3ChunkStore.java:2506)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
        at java.lang.Thread.run(Thread.java:745)
## Possible Solution
remove 

// Source and destination objects cannot be same, reply back error.
	if sourceObject == object && sourceBucket == bucket {
		writeErrorResponse(w, r, ErrInvalidCopyDest, r.URL.Path)
		return
	}

from CopyObjectHandler

1. create and object
2.update object metadata
3. copy opbject with new metadata over itself
4.

## Context
Unable to run Minio with OpenDedup as an object storage backend for backup

## Your Environment
OpenDedupe Target
On Ubuntu Linux
Minio 2016-11-22T00:35:30Z running in a docker container

harshavardhana:
## Description
This is written so that to simplify our handler code
and provide a way to only update metadata instead of
the data when source and destination in CopyObject
request are same.

## Motivation and Context
Fixes #3316

## How Has This Been Tested?
Manually.

## Types of changes

## Checklist:


harshavardhana:
Still adding tests.. - reviews welcome.

harshavardhana:
> Still adding tests.. - reviews welcome.

Tests done as well.

cmd/fs-helpers.go
```diff
@@ -0,0 +1,90 @@
...
+ */
+
+package cmd
+
+import "io"
+
+// Reads from the requested local location uses a staging buffer. Restricts
+// reads upto requested range of length and offset. If successful staging
+// buffer is written to the incoming stream. Returns errors if any.
+func fsReadFile(disk StorageAPI, bucket, object string, writer io.Writer, totalLeft, startOffset, bufSize int64, buf []byte) (err error) {
```
vadmeste: Can't we omit bufSize argument and use len(buf) instead ?

cmd/handler-utils.go
```diff
@@ -65,6 +65,31 @@ var supportedHeaders = []string{
 	// Add more supported headers here.
 }
 
+// Checks if metadata list in the request has updated by
+// verifying source metadata list.
+func isMetadataUpdate(srcMetadata, newMetadata map[string]string) bool {
```
vadmeste: I think it is better if you rename this function to 'needMetadataUpdate'

cmd/handler-utils.go
```diff
@@ -65,6 +65,31 @@ var supportedHeaders = []string{
...
+		}
+		if sVal != nVal {
+			return true
+		}
+	}
+	return false
+}
+
+// merge new metadata with existing source metadata.
+func mergeNewMetadata(srcMetadata, newMetadata map[string]string) {
```
krishnasrinivas: `func mergeMetadata(metadata ...map[string]string) map[string]string`

or

`func mergeMetadata(metadata1, metadata2 map[string]string) map[string]string`

will be better functions

cmd/fs-v1.go
```diff
@@ -217,7 +217,62 @@ func (fs fsObjects) DeleteBucket(bucket string) error {
...
 
-// GetObject - get an object.
+// CopyObject - copy object source object to destination object.
+func (fs fsObjects) CopyObject(srcBucket, srcObject, dstBucket, dstObject string, metadata map[string]string, updateMetadata bool) (ObjectInfo, error) {
+	// Stat the file to get file size.
+	fi, err := fs.storage.StatFile(srcBucket, srcObject)
+	if err != nil {
+		return ObjectInfo{}, toObjectErr(traceError(err), srcBucket, srcObject)
+	}
+
```
krishnasrinivas, line 229: if `updateMetadata` is true but src and dst are different the dst object is not written, just the metadata gets written.

cmd/object-handlers.go
```diff
@@ -302,46 +299,33 @@ func (api objectAPIHandlers) CopyObjectHandler(w http.ResponseWriter, r *http.Re
...
-	go func() {
-		startOffset := int64(0) // Read the whole file.
-		// Get the object.
-		gErr := objectAPI.GetObject(sourceBucket, sourceObject, startOffset, size, pipeWriter)
-		if gErr != nil {
-			errorIf(gErr, "Unable to read an object.")
-			pipeWriter.CloseWithError(gErr)
+	srcMetadata := objInfo.UserDefined
+	updateMetadata := isMetadataUpdate(srcMetadata, newMetadata)
+	if updateMetadata {
```
krishnasrinivas: if `updateMetadata` is true and dst != src, then lock is not held on dst during copyobject

cmd/object-handlers.go
```diff
@@ -302,46 +299,33 @@ func (api objectAPIHandlers) CopyObjectHandler(w http.ResponseWriter, r *http.Re
...
+		objectDWLock.Lock()
+		defer objectDWLock.Unlock()
+	}
 
-	// Create the object.
-	objInfo, err = objectAPI.PutObject(bucket, object, size, pipeReader, metadata, sha256sum)
+	// Copy source object to destination, if `updateMetadata` is set
+	// only metadata is updated. Source object and destination object
+	// should be same when metadata is being updated.
+	objInfo, err = objectAPI.CopyObject(srcBucket, srcObject, dstBucket, dstObject, srcMetadata, updateMetadata)
```
krishnasrinivas: `objInfo, err = objectAPI.CopyObject(srcBucket, srcObject, dstBucket, dstObject, mergedMetadata)`

will be a better API because, to decide whether to update just the metadata or not, the object layer can derive it by comparing src/dst

krishnasrinivas:
Looks like we are not doing it right when it comes to metadata, here's relevant info:
http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectCOPY.html

x-amz-metadata-directive:
```
Specifies whether the metadata is copied from the source object or replaced with metadata provided in the request.

* If copied, the metadata, except for the version ID, remains unchanged. In addition, the server-side-encryption, storage-class, and website-redirect-location metadata from the source is not copied. If you specify this metadata explicitly in the copy request, Amazon S3 adds this metadata to the resulting object. If you specify headers in the request specifying any user-defined metadata, Amazon S3 ignores these headers.
* If replaced, all original metadata is replaced by the metadata you specify.

Type: String

Default: COPY

Valid values: COPY | REPLACE

Constraints: Values other than COPY or REPLACE result in an immediate 400-based error response. You cannot copy an object to itself unless the MetadataDirective header is specified and its value set to REPLACE.
```

harshavardhana:
> Looks like we are not doing it right when it comes to metadata, here's relevant info:
> http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectCOPY.html

Ah looks like i assumed few things from the original bug report. Let me do it this way.

cmd/fs-v1.go
harshavardhana: Done

harshavardhana:
Spec changes have been implemented, please review again @vadmeste @krishnasrinivas

cmd/fs-helpers.go
harshavardhana: done.

cmd/object-handlers.go
```diff
@@ -298,50 +285,57 @@ func (api objectAPIHandlers) CopyObjectHandler(w http.ResponseWriter, r *http.Re
...
+	// Source and destination objects cannot be same if x-amz-metadata-directive is not set.
+	if cpMetadataOnly {
+		// Check if x-amz-metadata-direction was set to replace metadata.
+		if isMetadataReplace(r.Header) {
+			// Replace metadata requested proceed to extract
+			// relevant metadata from the request headers.
+			newMetadata = extractMetadataFromHeader(r.Header)
+
+			// Check if x-amz-metadata-direction was set to copy metadata.
+		} else if isMetadataCopy(r.Header) {
```
krishnasrinivas: this is an error case as well:

`You cannot copy an object to itself unless the MetadataDirective header is specified and its value set to REPLACE.`

cmd/object-handlers.go
```diff
@@ -298,50 +285,57 @@ func (api objectAPIHandlers) CopyObjectHandler(w http.ResponseWriter, r *http.Re
...
-		if gErr != nil {
-			errorIf(gErr, "Unable to read an object.")
-			pipeWriter.CloseWithError(gErr)
+		writeErrorResponse(w, r, ErrEntityTooLarge, cpSrcPath)
+		return
+	}
+
+	cpMetadataOnly := strings.EqualFold(cpSrcPath, cpDestPath)
+	var newMetadata map[string]string
+	// Source and destination objects cannot be same if x-amz-metadata-directive is not set.
```
krishnasrinivas: this comment is not correct because, src and dst can't be same even if x-amz-metadata-directive is set to COPY

cmd/object-handlers.go
```diff
@@ -298,50 +285,57 @@ func (api objectAPIHandlers) CopyObjectHandler(w http.ResponseWriter, r *http.Re
...
+		objectSRLock := globalNSMutex.NewNSLock(srcBucket, srcObject)
+		objectSRLock.RLock()
+		defer objectSRLock.RUnlock()
+	}
 
-	// Create the object.
-	objInfo, err = objectAPI.PutObject(bucket, object, size, pipeReader, metadata, sha256sum)
+	// Copy source object to destination, if source and destination
+	// object is same then only metadata is updated.
+	objInfo, err = objectAPI.CopyObject(srcBucket, srcObject, dstBucket, dstObject, newMetadata)
```
krishnasrinivas, line 348: note that in the code x-amz-metadata-directive is considered only when src and dst are same. but it needs to be considered even when src != dst

krishnasrinivas: > note that x-amz-metadata-directive is considered only when src and dst are same. but it needs to be considered even when src != dst

as a consequence of this, you can't release the read lock at line:270 because if there is a parallel PUT on the object, we might end up copying object content from newer object and metadata from older object.

krishnasrinivas: to simplify locking logic, this can be done at the beginning:
if (src == dst) then hold write lock on src
else hold read lock on src and write lock on dst

cmd/xl-v1-object.go
```diff
@@ -42,13 +42,105 @@ var objectOpIgnoredErrs = []error{
...
+		startOffset := int64(0) // Read the whole file.
+		if gerr := xl.GetObject(srcBucket, srcObject, startOffset, length, pipeWriter); gerr != nil {
+			errorIf(gerr, "Unable to read %s of the object `%s/%s`.", srcBucket, srcObject)
+			pipeWriter.CloseWithError(toObjectErr(gerr, srcBucket, srcObject))
+			return
+		}
+		pipeWriter.Close() // Close writer explicitly signalling we wrote all data.
+	}()
+
+	objInfo, err := xl.PutObject(dstBucket, dstObject, length, pipeReader, xlMeta.Meta, "")
```
krishnasrinivas: metadata needs to be considered instead of xlMeta.Meta.
the right metadata will be passed by the S3 layer depending on what the `x-amz-metadata-directive` is.

krishnasrinivas (changes requested):

cmd/object-handlers.go
harshavardhana: done

cmd/object-handlers.go
harshavardhana: done

cmd/object-handlers.go
harshavardhana: done

cmd/xl-v1-object.go
harshavardhana: done

harshavardhana:
Please review @vadmeste @krishnasrinivas ..

cmd/object-handlers.go
```diff
@@ -298,50 +316,28 @@ func (api objectAPIHandlers) CopyObjectHandler(w http.ResponseWriter, r *http.Re
...
-	// then its ETag will not be MD5sum of the object.
-	delete(metadata, "md5Sum")
-
-	sha256sum := ""
-
-	objectWLock := globalNSMutex.NewNSLock(bucket, object)
-	objectWLock.Lock()
-	defer objectWLock.Unlock()
+	defaultMeta := objInfo.UserDefined
+	newMetadata := getCpObjMetadataFromHeader(r.Header, defaultMeta)
```
vadmeste, line 336: newMetadata can be nil here though COPY is the default operation in S3 spec, Also we need to return http 400 when x-amz-metadata-directive is not COPY neither REPLACE

harshavardhana: > newMetadata can be nil here though COPY is the default operation in S3 spec, Also we need to return http 400 when x-amz-metadata-directive is not COPY neither REPLACE

Done.

vadmeste (approved):

PR #3491, opened 2016-12-22
Implement list and clear locks REST API

krisis:
## Description

Implements REST API for listing locks matching bucket , prefix and older than
and clear locks on a bucke/object.

## Motivation and Context

Essential for minio server administration.

## How Has This Been Tested?

Unit tests and simple functional tests.

## Types of changes

## Checklist:


krisis:
Changes to github.com/minio/pkg/madmin are pending.

cmd/admin-handlers.go
```diff
@@ -80,3 +80,77 @@ func (adminAPI adminAPIHandlers) ServiceRestartHandler(w http.ResponseWriter, r
...
+	// now (i.e, relTime is 0s in the past).
+	volLocks := filterLocks(bucket, object, time.Duration(0))
+	jsonBytes, err := json.Marshal(volLocks)
+	if err != nil {
+		writeErrorResponseNoHeader(w, r, ErrInternalError, r.URL.Path)
+		errorIf(err, "Failed to marshal lock information into json.")
+		return
+	}
+	// Remove lock matching bucket/object
+	globalNSMutex.ForceUnlock(bucket, object)
```
vadmeste: Can we force unlock before filtering locks above?

krisis: @vadmeste we are calling filterLocks(...) to find the locks that would be cleared for the given bucket and object. If we called filterLocks(...) after clearing the locks, we wouldn't see them.

vadmeste: > @vadmeste we are calling filterLocks(...) to find the locks that would be cleared for the given bucket and object. If we called filterLocks(...) after clearing the locks, we wouldn't see them.

You are right, I missed the logic of this code

pkg/madmin/lock-commands.go
```diff
@@ -0,0 +1,134 @@
...
+	Total int64
+	// No. of locks acquired on bucket, object.
+	Acquired int64
+	// No. of locks blocked on bucket, object.
+	Blocked int64
+	// Details on API operation that attempted the lock, when, for
+	// how long etc.
+	Details []LockDetail
+}
+
```
harshavardhana, line 71: Add a comment.

pkg/madmin/lock-commands.go
```diff
@@ -0,0 +1,134 @@
...
+func getLockInfos(body io.Reader) ([]LockInfo, error) {
+	respBytes, err := ioutil.ReadAll(body)
+	if err != nil {
+		return []LockInfo{}, err
```
harshavardhana: You can just return 'nil' for empty slices.

cmd/lockinfo-handlers.go
```diff
@@ -64,30 +67,79 @@ func getSystemLockState() (SystemLockState, error) {
...
 				Since:       lockInfo.since,
-				Duration:    time.Now().UTC().Sub(lockInfo.since),
+				Duration:    timeNow.Sub(lockInfo.since),
 			})
 		}
 		lockState.LocksInfoPerObject = append(lockState.LocksInfoPerObject, volLockInfo)
 	}
 	return lockState, nil
 }
+
```
harshavardhana: Needs a comment..

cmd/admin-handlers.go
```diff
@@ -80,3 +80,77 @@ func (adminAPI adminAPIHandlers) ServiceRestartHandler(w http.ResponseWriter, r
...
+	bucket := vars["bucket"]
+	prefix := vars["prefix"]
+	relTimeStr := vars["older-than"]
+
+	if !IsValidBucketName(bucket) {
+		writeErrorResponse(w, r, ErrInvalidBucketName, r.URL.Path)
+		return
+	}
+	if !IsValidObjectPrefix(prefix) {
+		writeErrorResponse(w, r, ErrInvalidObjectName, r.URL.Path)
```
harshavardhana: missing 'return'

cmd/admin-handlers.go
```diff
@@ -80,3 +80,77 @@ func (adminAPI adminAPIHandlers) ServiceRestartHandler(w http.ResponseWriter, r
...
+	}
+	relTime, err := time.ParseDuration(relTimeStr)
+	if err != nil {
+		writeErrorResponse(w, r, ErrInvalidDuration, r.URL.Path)
```
harshavardhana: Log the error on why it cannot be parsed..

cmd/admin-handlers.go
```diff
@@ -80,3 +80,77 @@ func (adminAPI adminAPIHandlers) ServiceRestartHandler(w http.ResponseWriter, r
...
+		errorIf(err, "Failed to marshal lock information into json.")
+		return
+	}
+	writeSuccessResponse(w, jsonBytes)
+}
+
+// ClearLocksHandler - POST /?lock&bucket=mybucket&object=myobject
+// HTTP header x-minio-operation: clear
+// ---------
+// Clear locks held on a given bucket, object and relative time.
```
harshavardhana: relative time is not used here?

cmd/admin-handlers.go
```diff
@@ -80,3 +80,77 @@ func (adminAPI adminAPIHandlers) ServiceRestartHandler(w http.ResponseWriter, r
...
+	volLocks := filterLocks(bucket, object, time.Duration(0))
+	jsonBytes, err := json.Marshal(volLocks)
+	if err != nil {
+		writeErrorResponseNoHeader(w, r, ErrInternalError, r.URL.Path)
+		errorIf(err, "Failed to marshal lock information into json.")
+		return
+	}
+	// Remove lock matching bucket/object
+	globalNSMutex.ForceUnlock(bucket, object)
+	writeSuccessResponse(w, jsonBytes)
```
harshavardhana, line 208: Is clear locks sending a response to provide rest of the list of locks? ..

cmd/admin-handlers.go
harshavardhana: Force unlocks should happen on all locks matching a prefix it is similar to 'rm' from `mc` point of view 
```
mc rm --recursive s3/miniocloud/object
```

Deletes all objects matching prefix `object`  and itself . Same applies to lock clear as well 

```
mc admin lock clear minio/testbucket/object
``` 

Clears the lock on the object itself , whereas 

```
mc admin clear --recursive minio/testbucket/object 
```

harshavardhana: Similarly like

```
mc rm --recursive --older-than=1 minio/testbucket/object
``` 

Locks are clear older than `1` day. 

```
mc admin lock clear --recursive --older-than=1 minio/testbucket/object
```

cmd/admin-handlers.go
```diff
@@ -80,3 +80,77 @@ func (adminAPI adminAPIHandlers) ServiceRestartHandler(w http.ResponseWriter, r
...
+	vars := mux.Vars(r)
+	bucket := vars["bucket"]
+	object := vars["object"]
+
+	if !IsValidBucketName(bucket) {
+		writeErrorResponse(w, r, ErrInvalidBucketName, r.URL.Path)
+		return
+	}
+	if !IsValidObjectName(object) {
+		writeErrorResponse(w, r, ErrInvalidObjectName, r.URL.Path)
```
harshavardhana: return is missing.

cmd/erasure-utils.go
```diff
@@ -42,15 +42,15 @@ func newHash(algo string) hash.Hash {
 	case "blake2b":
 		// ignore the error, because New512 without a key never fails
 		// New512 only returns a non-nil error, if the length of the passed
-		// key > 64 bytes - but we use blake2b as hash fucntion (no key)
+		// key > 64 bytes - but we use blake2b as hash function (no key)
```
harshavardhana, line 50: Already fixed. not needed..

harshavardhana (changes requested):

cmd/admin-handlers.go
krisis: The error message is returned as part of error response. How does it help to log it on server-side too?

cmd/admin-handlers.go
krisis: The corresponding implementation is missing in dsync and minio server. Should we hold this PR back until we have force unlocking on matching prefix and relative time?

cmd/admin-handlers.go
krisis: It returns the list of locks that belong to bucket, object that is supplied as query values. We use 0s as relative time to select all locks held on bucket, object.

cmd/admin-handlers_test.go
```diff
@@ -162,3 +163,151 @@ func TestServiceStopHandler(t *testing.T) {
 func TestServiceRestartHandler(t *testing.T) {
 	testServicesCmdHandler(restartCmd, t)
 }
+
```
abperiasamy, line 194: This file lacks comments. Every function (public and private )requires comment. Even code blocks like for and if.

krisis: Yes, I will add comments in the next update. Awaiting more reviews before updating the PR.

pkg/madmin/examples/lock-list.go
```diff
@@ -0,0 +1,48 @@
...
+	"time"
+
+	"github.com/minio/minio/pkg/madmin"
+)
+
+func main() {
+	// Note: YOUR-ACCESSKEYID, YOUR-SECRETACCESSKEY are
+	// dummy values, please replace them with original values.
+
+	// Requests are always secure (HTTPS) by default. Set secure=false to enable insecure (HTTP) access.
```
vadmeste: This comment doesn't make sense, no default parameter in go

docs/admin-api/service.md
```diff
@@ -31,3 +31,22 @@
...
   - x-minio-operation: status
   - Response: On success 200, return json formatted StorageInfo object.
+
+### Lock Management APIs
+* ListLocks
+  - GET /?lock&bucket=mybucket&prefix=myprefix&older-than=rel_time
+  - x-minio-operation: list
+  - Response: On success 200, json encoded response containing all locks held, older than rel_time. e.g, older than 3 hours.
+  - Possible error responses
+    - ErrInvalidBucketName
```
vadmeste: Can you add a simple paragraph somewhere to explain the generic format of the error ? (it must be something like what a S3 would return, right ?)

cmd/admin-rpc-client.go
```diff
@@ -155,9 +180,51 @@ func sendServiceCmd(cps adminPeers, cmd serviceSignal) {
...
+	// Used to aggregate volume lock information from all nodes.
+	allLocks := make([][]VolumeLockInfo, len(cps))
+	errs := make([]error, len(cps))
+	var wg sync.WaitGroup
+	remotePeers := cps[1:]
+	for i := range remotePeers {
+		wg.Add(1)
+		go func(idx int) {
+			defer wg.Done()
+			allLocks[idx+1], errs[idx] = cps[idx].cmdRunner.ListLocks(bucket, prefix, relTime)
```
vadmeste: Here, cps[idx] could be cps[0] because idx is zero in the first loop. The same goes for errs.

cmd/admin-rpc-client.go
```diff
@@ -155,9 +180,51 @@ func sendServiceCmd(cps adminPeers, cmd serviceSignal) {
...
+		}(i)
+	}
+	wg.Wait()
+	allLocks[0], errs[0] = cps[0].cmdRunner.ListLocks(bucket, prefix, relTime)
+
+	// Summarizing errors received for ListLocks RPC across all nodes.
+	errCount, err := reduceErrs(errs, []error{})
+	if errCount > (len(cps)/2+1) && err != nil {
```
vadmeste: This means we return an error only when errCount == 4 in case of 4 nodes. It needs a little review here

harshavardhana:
ping @krisis is this ready for merge?

krisis:
@harshavardhana another round of review of the recent changes would help? @vadmeste I have addressed your comments.

harshavardhana:
> @harshavardhana another round of review of the recent changes would help? @vadmeste I have addressed your comments.

Will do @krisis

cmd/admin-router.go
```diff
@@ -29,12 +29,20 @@ func registerAdminRouter(mux *router.Router) {
...
 	adminRouter.Methods("GET").Queries("service", "").Headers(minioAdminOpHeader, "status").HandlerFunc(adminAPI.ServiceStatusHandler)
 	// Service stop
 	adminRouter.Methods("POST").Queries("service", "").Headers(minioAdminOpHeader, "stop").HandlerFunc(adminAPI.ServiceStopHandler)
 	// Service restart
 	adminRouter.Methods("POST").Queries("service", "").Headers(minioAdminOpHeader, "restart").HandlerFunc(adminAPI.ServiceRestartHandler)
+
+	/// Lock operations
+
+	// List Locks
+	adminRouter.Methods("GET").Queries("lock", "", "bucket", "{bucket:.+}", "prefix", "{prefix:.+}", "older-than", "{older-than:.*}").Headers(minioAdminOpHeader, "list").HandlerFunc(adminAPI.ListLocksHandler)
```
vadmeste: I think it is better to avoid filtering requests here by bucket, prefix and older-than, so we can return a proper error message when lock is specified but other variables are not.. otherwise the request will be routed somewhere else, usually to object API

vadmeste: Also I think prefix here can be empty to list all hold locks

krisis: Yep, makes sense. I will make the necessary changes.

cmd/admin-handlers.go
```diff
@@ -80,3 +81,137 @@ func (adminAPI adminAPIHandlers) ServiceRestartHandler(w http.ResponseWriter, r
...
+// Only valid query params for list/clear locks management APIs.
+const (
+	lockBucket    lockQueryKey = "bucket"
+	lockPrefix    lockQueryKey = "prefix"
+	lockOlderThan lockQueryKey = "older-than"
+)
+
+// getFirstQueryVal - Returns the first query value for a given key and empty string otherwise.
+func getFirstQueryVal(vars url.Values, key lockQueryKey) string {
+	if qVals := vars[string(key)]; len(qVals) >= 1 {
```
harshavardhana: why can't you use vars.Get(string(key)) ?

cmd/admin-handlers.go
```diff
@@ -80,3 +81,137 @@ func (adminAPI adminAPIHandlers) ServiceRestartHandler(w http.ResponseWriter, r
...
+	bucket := getFirstQueryVal(vars, lockBucket)
+	prefix := getFirstQueryVal(vars, lockPrefix)
+	relTimeStr := getFirstQueryVal(vars, lockOlderThan)
+
+	// N B empty bucket name is invalid
+	if !IsValidBucketName(bucket) {
+		return "", "", time.Duration(0), ErrInvalidBucketName
+	}
+
+	// ... while an empty prefix is valid.
```
harshavardhana: Why "..." ?

cmd/admin-rpc-client.go
```diff
@@ -146,9 +167,58 @@ func sendServiceCmd(cps adminPeers, cmd serviceSignal) {
...
-			errs[idx] = invokeServiceCmd(remotePeers[idx], cmd)
+			// we use idx+1 because remotePeers slice is 1 position shifted w.r.t cps
+			errs[idx+1] = invokeServiceCmd(remotePeers[idx], cmd)
 		}(i)
 	}
 	wg.Wait()
 	errs[0] = invokeServiceCmd(cps[0], cmd)
 }
+
+func fetchLocks(cps adminPeers, bucket, prefix string, relTime time.Duration) ([]VolumeLockInfo, error) {
```
harshavardhana: this can be named as `listLocksPeers()`

harshavardhana: and a comment as well.

harshavardhana: also what does `cps` variable name stand for?  should be simply `peers` i think.

cmd/admin-rpc-client.go
```diff
@@ -146,9 +167,58 @@ func sendServiceCmd(cps adminPeers, cmd serviceSignal) {
...
+	// Used to aggregate volume lock information from all nodes.
+	allLocks := make([][]VolumeLockInfo, len(cps))
+	errs := make([]error, len(cps))
+	var wg sync.WaitGroup
+	remotePeers := cps[1:]
+	for i := range remotePeers {
```
harshavardhana: No reason to save `remotePeers` you can directly do `for i := range peers[1:] { } `

cmd/admin-rpc-client.go
```diff
@@ -146,9 +167,58 @@ func sendServiceCmd(cps adminPeers, cmd serviceSignal) {
...
+
+	// Summarizing errors received for ListLocks RPC across all
+	// nodes.  N B the possible unavailability of quorum in errors
+	// applies only to distributed setup.
+	errCount, err := reduceErrs(errs, []error{})
+	if err != nil {
+		if errCount >= (len(cps)/2 + 1) {
+			return nil, err
+		}
+		return nil, errXLWriteQuorum
```
harshavardhana: I am wondering if we need to do this, how important are these errors for the user..From listing point of view they are not that interesting, just logging on the server might be good enough. Only in case of clear locks we might have to send an error.

cmd/lockinfo-handlers.go
```diff
@@ -84,10 +89,57 @@ func getSystemLockState() (SystemLockState, error) {
...
+// filterLocks - Fetches locks held on bucket, matching prefix older than relTime.
```
harshavardhana: You don't need to say this as "filterLocks()"  - this is listLocks() - input params looks quite similar to listObjects(). relTime is like marker.

pkg/madmin/examples/lock-clear.go
```diff
@@ -0,0 +1,47 @@
...
+
+	// API requests are secure (HTTPS) if secure=true and insecure (HTTPS) otherwise.
+	// New returns an Minio Admin client object.
+	madmClnt, err := madmin.New("your-minio.example.com:9000", "YOUR-ACCESSKEYID", "YOUR-SECRETACCESSKEY", true)
+	if err != nil {
+		log.Fatalln(err)
+	}
+
+	// Clear locks held on mybucket/myprefix older than olderThan seconds.
+	olderThan := time.Duration(10 * time.Second)
```
harshavardhana: Use perhaps `30 * time.Minute` as a more palpable value for common use case.

pkg/madmin/lock-commands.go
```diff
@@ -0,0 +1,136 @@
...
+	// Execute POST on /?lock to clear locks.
+	resp, err := adm.executeMethod("POST", reqData)
+
+	defer closeResponse(resp)
+	if err != nil {
+		return nil, err
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, errors.New("Got " + resp.Status)
```
harshavardhana: Make this a little better atleast "Got HTTP Status: "

pkg/madmin/lock-commands_test.go
```diff
@@ -0,0 +1,56 @@
...
+package madmin
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"reflect"
+	"testing"
+)
+
```
harshavardhana, line 26: Comment

cmd/admin-handlers.go
krisis: I didn't know it was part of url package. I shall use it.

cmd/admin-rpc-client.go
krisis: listLocksPeers doesn't sound right. I shall add a comment and change cps to something more meaningful.

cmd/admin-rpc-client.go
krisis: It is a matter of style. I shall change to peers[1:].

cmd/admin-rpc-client.go
krisis: The errors may not be interesting to the user. In the event of an error, we need to decide what values we return to the user. IMO, we can't return a list of locks that aren't agreed upon.

pkg/madmin/lock-commands.go
krisis: OK, I shall change it uniformly in all places in pkg/madmin incl. service commands.

cmd/admin-rpc-client.go
krisis: On second thoughts, I prefer leaving it as is. It helps the reader to appreciate the difference in handling between local and remote peers.

harshavardhana: in that case you can have 

```
remotePeers
localPeer
``` 

For completeness.

cmd/lockinfo-handlers.go
```diff
@@ -84,10 +89,57 @@ func getSystemLockState() (SystemLockState, error) {
...
+	for param, debugLock := range globalNSMutex.debugLockMap {
+		if param.volume != bucket {
+			continue
+		}
+		// N B empty prefix matches all param.path.
+		if !strings.HasPrefix(param.path, prefix) {
+			continue
+		}
+
+		volLockInfo := VolumeLockInfo{
```
vadmeste, line 118: volLockInfo is always added to the returned slice regardless of the specified relTime, this means clients will have entries for any older-than value specified

vadmeste:
One thing, some locks are hold with this pair (.minio.sys/multipart, bucket/prefix), did we think about this case ?

harshavardhana:
Ping @krisis can you rebase this again?

krisis:
@vadmeste Good catch, IsValidBucketName check fails if  ".minio.sys/multipart" is the 'bucket' on which a lock is held. @harshavardhana Should we ignore these locks as internal and not expose them via ``mc admin locks {list|clear}`` subcommands? Alternately, we can relax the valid bucket name check to skip check on internal bucket names like ".minio.sys/multipart". Thoughts?

harshavardhana:
> @vadmeste Good catch, IsValidBucketName check fails if ".minio.sys/multipart" is the 'bucket' on which a lock is held. @harshavardhana Should we ignore these locks as internal and not expose them via mc admin locks {list|clear} subcommands? Alternately, we can relax the valid bucket name check to skip check on internal bucket names like ".minio.sys/multipart". Thoughts?

IsValidBucketName() we already support '.minio.sys' since we treat `.minio.sys/multipart` also as a bucket we can apply the same restrictions.

harshavardhana:
> @harshavardhana Should we ignore these locks as internal and not expose them via mc admin locks {list|clear} subcommands? Alternately, we can relax the valid bucket name check to skip check on internal bucket names like ".minio.sys/multipart". Thoughts?

In previous implementation we didn't ignore it @krisis it is perhaps important to clear them as well, potentially they are part of the namespace as well.

cmd/lockinfo-handlers.go
harshavardhana: @krisis i am not sure what is this? - can you clarify more @vadmeste .

vadmeste: @harshavardhana, @krisis already fixed this. Before the fix, the returned VolumeLockInfo slice wasn't empty even when no locks were hold

harshavardhana:
what else is pending on this @krisis ?

krisis:
@harshavardhana I have added minioMetaMultipartBucket to the list of valid bucket names. AFAICS, this patch is complete.

harshavardhana (approved):

Issue #3475, opened 2016-12-20
server: Start time sanity check across all servers


harshavardhana:
Create a new ServerInfo() RPC call to query all its neighbors to check for sane environment. 

- Architecture (OS and ARCH has to match)
- cmdline (CLI Args)

## Expected Behavior
We need to verify if all servers in a distributed setup are sane. 

## Current Behavior
No verification done can lead to corrupted state.

## Possible Solution
Explained above.

## Steps to Reproduce (for bugs)
Bringing any servers with any of the above environments can cause issues.

## Context
Sane environment check is needed and is an important check before we proceed to initialize object layer.

## Your Environment
Current master.

GarimaKapoor:
@vadmeste is this issue still work in progress

harshavardhana:
Actually @balamurugana is working on a related issue.

harshavardhana:
ping @balamurugana

deekoder:
Moving this to next milestone since this involves several moving parts that need to be changed eg #3970

harshavardhana:
This change is necessary starting with change https://github.com/minio/minio/pull/5420 @donatello @nitisht

harshavardhana:
config has moved to backend so all servers are always going to be same, closing this as not needed anymore.

PR #3492, opened 2016-12-22
[WIP] s3peer: Rewrite to add support for new peer clients 

vadmeste:
## Description
Make it easier to add new s3 peer clients and avoid being limited to only BMS client

## Motivation and Context
Pave the way for new RPC clients for inter nodes communications, like Get Server Info

## How Has This Been Tested?
go test

## Types of changes

## Checklist:

### DIFF
--- a/cmd/s3-peer-bms-client.go
+++ b/cmd/s3-peer-bms-client.go
@@ -1,0 +1,149 @@
+/*
+ * Minio Cloud Storage, (C) 2014-2016 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cmd
+
+import (
+	"encoding/json"
+	"fmt"
+	"sync"
+)
+
+// Bucket Meta State interface to be registered in S3 Peers API
+type s3PeerBucketMetaStateAPI struct{}
+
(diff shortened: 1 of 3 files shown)


cmd/s3-peer-bms-client.go
```diff
@@ -0,0 +1,149 @@
...
+	"fmt"
+	"sync"
+)
+
+// Bucket Meta State interface to be registered in S3 Peers API
+type s3PeerBucketMetaStateAPI struct{}
+
+// Bucket Meta State API name
+func (c s3PeerBucketMetaStateAPI) APIName() string {
+	return "bms"
```
harshavardhana, line 30: You should type the API name than using generic string like this, hard to catch bugs.

cmd/s3-peer-bms-client.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+// Return Bucket Meta State API for local use
+func (c s3PeerBucketMetaStateAPI) NewLocalClient() interface{} {
```
harshavardhana, line 34: Catch all interface should be avoided in the return statement.. The functionaliy which localBucketMetaState would provide should be put into a concrete type.

harshavardhana: A concrete type here should put into an interface like below 

```
type UpdateBucketMetadata interface {
... add relevant methods generic for all metadata updates ...
}
```

harshavardhana (changes requested):
Overall this is a good change but we need to make sure to concrete types either by writing a common interface compatible type or use a proper struct.

cmd/s3-peer-bms-client.go
vadmeste: Actually this PR will help to implement new calls (local and remote) like Get Storage Info (https://github.com/minio/minio/issues/3475).. that's why I am using generic interface{}

harshavardhana: Still we need to explore all possible ways before we can say that we have to use a generic interface{} like this.

harshavardhana:
@vadmeste is this still valid ?

vadmeste:
> @vadmeste is this still valid ?

Not valid anymore since @balamurugana is taking over this

PR #3494, opened 2016-12-23
nats: Add support of NATS.io Streaming server

### LINKED ISSUE #3466: notification: Support NATS streaming server for bucket notifications
Current nats support only uses `gnatsd` which is a non persistent queue. 

## Expected Behavior
Support persistent streaming NATS server.

## Current Behavior
We do not support persistent NATS server.

## Possible Solution
https://github.com/nats-io/go-nats-streaming use this. 

## Context
This was requested by a community member.

vadmeste:
## Description
Introduce support of NATS streaming server (or STAN), streaming field is added in nats json config

Below is the new format:

``` 
               "nats": {
                       "1": {
                               "enable": true,
                               "address": "localhost:4222",
                               "subject": "mysubject",
                               "username": "admin",
                               "password": "admin",
                               "token": "",
                               "secure": false,
                               "pingInterval": 0,
                               "streaming": {
                                       "enable": false,
                                       "clusterID": "",
                                       "clientID": "",
                                       "async": false,
                                       "maxPubAcksInflight": 0
                               }
                       }
               },
```

## Motivation and Context
Fixes #3466 

## How Has This Been Tested?
Manually + go test

## Types of changes

## Checklist:

- [] I have added tests to cover my changes.


vadmeste:
Ready for review and merge

harshavardhana:
We need to decide on keeping support for existing NATs support. Can you explore if we need to keep the existing non persistent nats support basically why streaming is better than existing gnatsd ?. I am not sure we need to support it just that it will cause confusion and we can keep it simple. By supporting only one way of doing NATs if clearly streaming is better.

Will discuss with AB further on the final call.

vadmeste:
> We need to decide on keeping support for existing NATs support. Can you explore if we need to keep the existing non persistent nats support basically why streaming is better than existing gnatsd ?. I am not sure we need to support it just that it will cause confusion and we can keep it simple. By supporting only one way of doing NATs if clearly streaming is better.
> 
> Will discuss with AB further on the final call.

My findings:
NATS streaming focuses on message delivery reliability and helps clients to retrieve old sent messages when they were disconnected. Though streaming and core (gnatsd) are inter operable (https://nats.io/documentation/faq/), I really didn't see any encouragements in forums to move to streaming version though this latter looked more generic to me, probably because the core is faster.

It makes more sense to me to use plugins for notifications when go plugin will be ready or probably even now. It would be easier since that plugin will receive any parameters from a config file, and has two interfaces Connect() and Send().

harshavardhana:
> It makes more sense to me to use plugins for notifications when go plugin will be ready or probably even now. It would be easier since that plugin will receive any parameters from a config file, and has two interfaces Connect() and Send().

This is a good idea but we need to wait for 1.8

harshavardhana:
Can you add some tests @vadmeste ?

vadmeste:
> Can you add some tests @vadmeste ?

Tests for config migration are done, the remaining code needs real NATS/STAN servers for testing

harshavardhana (approved):

PR #3502, opened 2016-12-28
Improve reconnection logic, allow jitters.

### LINKED ISSUE #3372: inconsistent StorageInfo on cluster
4 physical servers running the edge version of minio with docker.
It's a staging environment and the minio are restarted frequently in //
Sometimes the StorageInfo call does not return the same info for all servers, restarting the "failing" server fix the problem, the data are **not changed** between the restarts, we do **not** remove the exported folder.

It would be better to consistently restart successfully but way more important, all StorageInfo should report strictly the same information about Connected and failing disks.

```
{
     "host": "http://node0.st2.------tech.org:9000",
     "data": {
       "storageInfo": {
         "Total": 15872866181120,
         "Free": 14900683038720,
         "Backend": {
           "Type": 2,
           "OnlineDisks": 4,
           "OfflineDisks": 0,
           "ReadQuorum": 2,
           "WriteQuorum": 3
         }
       },
       "uiVersion": "2016-11-24T00:39:02Z"
     }
   },
   {
     "host": "http://node1.st2.------tech.org:9000",
     "data": {
       "storageInfo": {
         "Total": 11904649635840,
         "Free": 11174937053184,
         "Backend": {
           "Type": 2,
           "OnlineDisks": 3,
           "OfflineDisks": 1,
           "ReadQuorum": 2,
           "WriteQuorum": 3
         }
       },
       "uiVersion": "2016-11-24T00:39:02Z"
     }
   },
   {
     "host": "http://node2.st2.------tech.org:9000",
     "data": {
       "storageInfo": {
         "Total": 15872866181120,
         "Free": 14900683038720,
         "Backend": {
           "Type": 2,
           "OnlineDisks": 4,
           "OfflineDisks": 0,
           "ReadQuorum": 2,
           "WriteQuorum": 3
         }
       },
       "uiVersion": "2016-11-24T00:39:02Z"
     }
   },
   {
     "host": "http://node3.st2.------tech.org:9000",
     "data": {
       "storageInfo": {
         "Total": 15872866181120,
         "Free": 14900683038720,
         "Backend": {
           "Type": 2,
           "OnlineDisks": 4,
           "OfflineDisks": 0,
           "ReadQuorum": 2,
           "WriteQuorum": 3
         }
       },
       "uiVersion": "2016-11-24T00:39:02Z"
     }
   }
```

harshavardhana:
## Description
This is needed since any network operation error
is converted to disk not found but we also need
to make sure really if disk is not available.

Additionally we also need to retry more than
once because the server might be in startup
sequence which would render other servers to
wrongly think that the server is offline.

## Motivation and Context
Fixes #3372 

## How Has This Been Tested?
Manually and using a distributed setup using `y4m4/minio` docker image.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/retry-storage.go
+++ b/cmd/retry-storage.go
@@ -17,7 +17,7 @@
 package cmd
 
 import (
-	"net/rpc"
+	"time"
 
 	"github.com/minio/minio/pkg/disk"
 )
@@ -48,7 +48,7 @@ func (f retryStorage) Close() (err error) {
 // DiskInfo - a retryable implementation of disk info.
 func (f retryStorage) DiskInfo() (info disk.Info, err error) {
 	info, err = f.remoteStorage.DiskInfo()
-	if err == rpc.ErrShutdown {
+	if err == errDiskNotFound {
 		err = f.reInit()
 		if err == nil {
 			return f.remoteStorage.DiskInfo()
@@ -60,7 +60,7 @@ func (f retryStorage) DiskInfo() (info disk.Info, err error) {
 // MakeVol - a retryable implementation of creating a volume.
 func (f retryStorage) MakeVol(volume string) (err error) {
 	err = f.remoteStorage.MakeVol(volume)
-	if err == rpc.ErrShutdown {
+	if err == errDiskNotFound {
 		err = f.reInit()
 		if err == nil {
 			return f.remoteStorage.MakeVol(volume)
@@ -72,7 +72,7 @@ func (f retryStorage) MakeVol(volume string) (err error) {
 // ListVols - a retryable implementation of listing all the volumes.
 func (f retryStorage) ListVols() (vols []VolInfo, err error) {
 	vols, err = f.remoteStorage.ListVols()
-	if err == rpc.ErrShutdown {
+	if err == errDiskNotFound {
 		err = f.reInit()
 		if err == nil {
 			return f.remoteStorage.ListVols()
@@ -84,7 +84,7 @@ func (f retryStorage) ListVols() (vols []VolInfo, err error) {
 // StatVol - a retryable implementation of stating a volume.
 func (f retryStorage) StatVol(volume string) (vol VolInfo, err error) {
 	vol, err = f.remoteStorage.StatVol(volume)
-	if err == rpc.ErrShutdown {
+	if err == errDiskNotFound {
 		err = f.reInit()
 		if err == nil {
 			return f.remoteStorage.StatVol(volume)
@@ -96,7 +96,7 @@ func (f retryStorage) StatVol(volume string) (vol VolInfo, err error) {
 // DeleteVol - a retryable implementation of deleting a volume.
 func (f retryStorage) DeleteVol(volume string) (err error) {
 	err = f.remoteStorage.DeleteVol(volume)
-	if err == rpc.ErrShutdown {
+	if err == errDiskNotFound {
 		err = f.reInit()
 		if err == nil {
 			return f.remoteStorage.DeleteVol(volume)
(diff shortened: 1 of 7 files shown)


cmd/storage-rpc-client.go
```diff
@@ -143,13 +143,15 @@ const maxAllowedNetworkIOError = 1024
 // Initializes the remote RPC connection by attempting a login attempt.
 func (n *networkStorage) Init() (err error) {
 	// Attempt a login to reconnect.
-	return n.rpcClient.Login()
+	err = n.rpcClient.Login()
+	return toStorageErr(err)
```
krisis, line 151: Which errors are meaningfully converted using ``toStorageErr`` in the case of Login?

cmd/retry-storage.go
```diff
@@ -84,7 +85,7 @@ func (f retryStorage) ListVols() (vols []VolInfo, err error) {
 // StatVol - a retryable implementation of stating a volume.
 func (f retryStorage) StatVol(volume string) (vol VolInfo, err error) {
 	vol, err = f.remoteStorage.StatVol(volume)
-	if err == rpc.ErrShutdown {
+	if err == rpc.ErrShutdown || err == errDiskNotFound {
```
krisis: We could wrap this (repeating) error value checks into a function, say 
```go
func isDiskAvailablle(err error) bool
```

cmd/retry-storage.go
```diff
@@ -189,36 +190,59 @@ func (f retryStorage) DeleteFile(volume, path string) (err error) {
...
 		}
 	}
 	return err
 }
+
+// Connect and attempt to load the format from a disconnected node,
+// attempts three times before giving up.
+func (f retryStorage) reInit() (err error) {
+	err = f.remoteStorage.Close()
+	if err != nil {
```
krisis: Should we exit on receiving an error here? It is likely that we might see the same network error again, especially if the underlying connection was closed in an earlier RPC call. I think we can ignore error on close in ``reInit``. Thoughts?

cmd/storage-rpc-client.go
harshavardhana: all network errors.. that's all.

cmd/retry-storage.go
harshavardhana: Done

cmd/retry-storage.go
harshavardhana: Yeah Close can be ignored.. there won't be any error of interest here..

cmd/storage-rpc-client.go
krisis: ``rpc.ErrShutdown`` isn't a ``net.OpError``, should we then handle that too?

harshavardhana: I think we can keep it as errDiskNotFound and simply the checks

krisis (approved):

dcharbonnier:
and if I don't want my minio to start if all the servers are not connected ? can I get an option to force a start even if a disk is lost and not by default ? and if I think that 1:30 minute is magic and I prefer 5 because 5 is a better number than 3 and also maybe because in case of a datacenter going all  down I'm not sure all machines will come back at the same moment ? and maybe I think that retrying every 30 seconds is not very useful and it would be better to try every second, after all it cost nothing to try more times ? we are not speaking about a microservice that is design to tolerate failures, minio is the storage, one of the most critical part of the infrastructure, "giving up" on disks is not something easy especially when there is no automatic way to recover when a server come back

harshavardhana:
> and if I don't want my minio to start if all the servers are not connected ? can I get an option to force a start even if a disk is lost and not by default ? and if I think that 1:30 minute is magic and I prefer 5 because 5 is a better number than 3 and also maybe because in case of a datacenter going all down I'm not sure all machines will come back at the same moment ? and maybe I think that retrying every 30 seconds is not very useful and it would be better to try every second, after all it cost nothing to try more times ? we are not speaking about a microservice that is design to tolerate failures, minio is the storage, one of the most critical part of the infrastructure, "giving up" on disks is not something easy especially when there is no automatic way to recover when a server come back

The fix is incorrect here @dcharbonnier will fix it in a different manner.

PR #3506, opened 2016-12-29
Auto-reconnect for regular authRPC client.

harshavardhana:
## Description
Implement a storage rpc specific rpc client,
which does not reconnect unnecessarily.

Instead reconnect is handled at a different
layer for storage alone.

Rest of the calls using AuthRPC automatically
reconnect, i.e upon an error equal to `rpc.ErrShutdown`
they dial again and call the requested method again.

## Motivation and Context
To have auto reconnect login for all AuthRPC consumers.

## How Has This Been Tested?
Manually, distributed setup. 

## Types of changes

## Checklist:

### DIFF
--- a/cmd/auth-rpc-client.go
+++ b/cmd/auth-rpc-client.go
@@ -76,7 +76,6 @@ type AuthRPCClient struct {
 	mu            sync.Mutex
 	config        *authConfig
 	rpc           *RPCClient // reconnect'able rpc client built on top of net/rpc Client
-	isLoggedIn    bool       // Indicates if the auth client has been logged in and token is valid.
 	serverToken   string     // Disk rpc JWT based token.
 	serverVersion string     // Server version exchanged by the RPC.
 }
@@ -88,8 +87,6 @@ func newAuthClient(cfg *authConfig) *AuthRPCClient {
 		config: cfg,
 		// Initialize a new reconnectable rpc client.
 		rpc: newRPCClient(cfg.address, cfg.path, cfg.secureConn),
-		// Allocated auth client not logged in yet.
-		isLoggedIn: false,
 	}
 }
 
@@ -97,7 +94,7 @@ func newAuthClient(cfg *authConfig) *AuthRPCClient {
 func (authClient *AuthRPCClient) Close() error {
 	authClient.mu.Lock()
 	// reset token on closing a connection
-	authClient.isLoggedIn = false
+	authClient.serverToken = ""
 	authClient.mu.Unlock()
 	return authClient.rpc.Close()
 }
@@ -109,7 +106,7 @@ func (authClient *AuthRPCClient) Login() (err error) {
 	defer authClient.mu.Unlock()
 
 	// Return if already logged in.
-	if authClient.isLoggedIn {
+	if authClient.serverToken != "" {
 		return nil
 	}
 
@@ -135,7 +132,6 @@ func (authClient *AuthRPCClient) Login() (err error) {
 	// Set token, time stamp as received from a successful login call.
 	authClient.serverToken = reply.Token
 	authClient.serverVersion = reply.ServerVersion
-	authClient.isLoggedIn = true
 	return nil
 }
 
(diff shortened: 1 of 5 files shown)


cmd/auth-rpc-client.go
```diff
@@ -146,20 +146,26 @@ func (authClient *AuthRPCClient) Call(serviceMethod string, args interface {
...
-		// Call the underlying rpc.
-		err = authClient.rpc.Call(serviceMethod, args, reply)
-
-		// Invalidate token, and mark it for re-login on subsequent reconnect.
+		// Finally make the network call using net/rpc client.
+		return authClient.rpc.Call(serviceMethod, args, reply)
+	}
+	if err = loginAndCallFn(); err != nil {
+		// Invalidate token, and mark it for re-login and reconnect.
 		if err == rpc.ErrShutdown {
```
balamurugana: You could have `if err = loginAndCallFn(); err == rpc.ErrShutdown {`. No one more level of `if` is needed.

cmd/auth-rpc-client.go
```diff
@@ -146,20 +146,26 @@ func (authClient *AuthRPCClient) Call(serviceMethod string, args interface {
 	SetToken(token string)
 	SetTimestamp(tstamp time.Time)
 }, reply interface{}) (err error) {
-	// On successful login, attempt the call.
-	if err = authClient.Login(); err == nil {
+	loginAndCallFn := func() error {
+		// On successful login, proceed to attempt the requested service method.
+		if err = authClient.Login(); err != nil {
```
balamurugana: You could have simpler `if` by
```golang
if err = authClient.Login(); err == nil {
    args.SetToken(authClient.serverToken)
    args.SetTimestamp(time.Now().UTC())
    err = authClient.rpc.Call(serviceMethod, args, reply)
}

return err
```

cmd/auth-rpc-client.go
```diff
@@ -146,20 +146,26 @@ func (authClient *AuthRPCClient) Call(serviceMethod string, args interface {
...
+		// Invalidate token, and mark it for re-login and reconnect.
 		if err == rpc.ErrShutdown {
-			authClient.mu.Lock()
-			authClient.isLoggedIn = false
-			authClient.mu.Unlock()
+			// Close the underlying connection.
+			authClient.Close()
+
+			// Call login and call again.
+			err = loginAndCallFn()
```
balamurugana: This logic makes one more attempt to execute on `ErrShutdown`.  It would be nice to have a logic to make multiple attempts if its wanted in future meaning just changing a constant will do more or less attempts.  In short, you could have a `for` loop.

harshavardhana: Can do a retry based on newRetryTimer.. let me add that.

cmd/auth-rpc-client.go
harshavardhana: done

cmd/auth-rpc-client.go
harshavardhana: done

cmd/auth-rpc-client.go
```diff
@@ -146,21 +142,34 @@ func (authClient *AuthRPCClient) Call(serviceMethod string, args interface {
...
+		if err = authClient.Login(); err == nil {
+			// Set token and timestamp before the rpc call.
+			args.SetToken(authClient.serverToken)
+			args.SetTimestamp(time.Now().UTC())
+
+			// Finally make the network call using net/rpc client.
+			err = authClient.rpc.Call(serviceMethod, args, reply)
 		}
+		return err
+	}
```
balamurugana, line 156: It would nice to have newline to separate block of statements.

balamurugana (approved):

cmd/auth-rpc-client.go
harshavardhana: I don't like that style, frivolous spaces should be added using `gofmt` if gofmt doesn't complain then its not needed. I don't see it anywhere in official Go style etc.

balamurugana: `gofmt` is a tool to make formats for syntax beautification. It doesn't do logic separation/beautification of the code blocks.  That's why `gofmt` doesn't worry about single newline within code blocks.

If code is not logically separated, its hard to read/understand.

harshavardhana: Logically it is separated using "}" braces. I don't see this to be norm and have never seen it in Golang. Following unconventional stuff from other languages is wrong and i don't see it add anything for readability. I am perfectly able to read code with or without it . Cannot see how it is hard.

balamurugana: The practice is not to any specific language and Golang doesn't have such enforcement about not to have block separation.

I think you refer Golang as its library code base.  I see golang internal libraries do not follow such convention because of very specific/simpler/few line of code within a function, but golang tools follow that style.

I checked Kubernetes/Docker applications written in Go.  They follow the style I mentioned.

ATM there is no tool can beautify a code by logic, such tool is hard to write :smile:

harshavardhana: Can you share snippet of code i don't see it anywhere.. 

- https://github.com/docker/docker/blob/master/cmd/dockerd/daemon.go#L176
- https://github.com/docker/docker/blob/master/cmd/dockerd/daemon.go#L88
- https://github.com/docker/docker/blob/master/cmd/dockerd/daemon.go#L258

if it is done somewhere it is not consistent, in-fact docker also lacks comments too so i wouldn't buy into it too much into it.

The style itself is fine, but it is not a sacrosanct necessity if Go upstream project can live with it i don't see why we would have any issues and we don't even have 1/10th of the contributors they have.

PR #3509, opened 2016-12-30
fs: Re-implement object layer to remember the fd.

harshavardhana:
## Description
This patch re-writes FS backend to support shared backend sharing locks for safe concurrent access across multiple servers.

## Motivation and Context
To facilitate a way to run many minio servers on the same shared backend.

## How Has This Been Tested?
Manually using special setup with shared backend, NFS, XFS and SMB (Windows). 

## Types of changes

## Checklist:


cmd/fs-v1-multipart.go
```diff
@@ -601,52 +755,42 @@ func (fs fsObjects) CompleteMultipartUpload(bucket string, object string, upload
...
+	if err != nil {
+		return "", toObjectErr(traceError(err), bucket, object)
 	}
+	defer rwlk.Close()
 
-	// Return md5sum.
-	return s3MD5, nil
-}
+	// Close lock held on bucket/object/uploadid/fs.json
+	fs.rwPool.Close(fsMetaPathMultipart)
```
krishnasrinivas, line 764: this is already done at line 642 and hence doing it again can cause undesired behavior

cmd/fs-v1-multipart.go
```diff
@@ -666,17 +810,52 @@ func (fs fsObjects) AbortMultipartUpload(bucket, object, uploadID string) error
...
-		return traceError(InvalidUploadID{UploadID: uploadID})
+	fsMetaPath := pathJoin(fs.fsPath, minioMetaMultipartBucket, uploadIDPath, fsMetaJSONFile)
+	_, err = fs.rwPool.Open(fsMetaPath)
+	if err != nil {
+		if err == errFileNotFound {
+			return traceError(InvalidUploadID{UploadID: uploadID})
+		}
+		return toObjectErr(traceError(err), bucket, object)
 	}
+	fs.rwPool.Close(fsMetaPath)
```
krishnasrinivas: should do `defer fs.rwPool.Close(fsMetaPath)`

cmd/fs-v1-rwpool.go
```diff
@@ -0,0 +1,156 @@
...
+// count has reached zero.
+func (fsi *fsIOPool) Open(path string) (*lock.RLockedFile, error) {
+	fsi.Lock()
+	rlkFile, ok := fsi.readersMap[path]
+
+	// File reference exists on map, validate if its
+	// really closed and we are safe to purge it.
+	if ok {
+		// If the file is closed, remove it from the reader pool map.
+		if rlkFile.IsClosed() {
```
krishnasrinivas, line 58: should never reach here, we can do an errorIf() to catch a bug

cmd/fs-v1-multipart.go
harshavardhana: Needed on windows, its not really undesired.

cmd/fs-v1-multipart.go
harshavardhana: Same windows its needed.

cmd/fs-v1-rwpool.go
harshavardhana: Done

cmd/fs-v1-multipart.go
krishnasrinivas: double Close() will cause problems, i.e it can close the fd even when its not supposed to.

harshavardhana: done

cmd/fs-v1-multipart.go
harshavardhana: done

PR #3515, opened 2017-01-02
browser: Allow anonymous browsing of readable buckets.

krishnasrinivas:
## Description

Allow anonymous browsing of readable buckets.

## How Has This Been Tested?

Manual testing

## Types of changes

## Checklist:

### DIFF
--- a/cmd/jwt.go
+++ b/cmd/jwt.go
@@ -43,6 +43,7 @@ var errInvalidSecretKeyLength = errors.New("Invalid secret key, secret key shoul
 
 var errInvalidAccessKeyID = errors.New("The access key ID you provided does not exist in our records")
 var errAuthentication = errors.New("Authentication failed, check your access credentials")
+var errNoAuthToken = errors.New("JWT token missing")
 
 func authenticateJWT(accessKey, secretKey string, expiry time.Duration) (string, error) {
 	// Trim spaces.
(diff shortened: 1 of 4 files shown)


cmd/jwt.go
```diff
@@ -114,3 +115,18 @@ func isHTTPRequestValid(req *http.Request) bool {
 
 	return jwtToken.Valid
 }
+
```
harshavardhana, line 112: Comment needed..

cmd/jwt.go
```diff
@@ -43,6 +43,7 @@ var errInvalidSecretKeyLength = errors.New("Invalid secret key, secret key shoul
 
 var errInvalidAccessKeyID = errors.New("The access key ID you provided does not exist in our records")
 var errAuthentication = errors.New("Authentication failed, check your access credentials")
+var errNoAuthToken = errors.New("JWT token missing")
```
harshavardhana, line 46: Add this to "toWebAPIError()" function as well.

harshavardhana (changes requested):

harshavardhana:
@krishnasrinivas - can you add some test cases?

krishnasrinivas:
@harshavardhana the right test cases for this feature is to first apply a READ policy and try a `download` and `list` using JSONRPC without token and it should work. Then apply a WRITE policy and then verify that `upload` using JSONRPC without token should work.

I thought of augmenting bucket policy tests but we don't have test cases to test the actual working of bucket policies in bucket-policy-handlers_test.go using GET/PUT requests. i.e we just verify that put-policy get-policy delete-policy calls succeed, we don't verify if the policies itself are effective. We'll have to write if policy is effective with regular GET/PUT and also over JSON-RPC, since the tests will address a wider scope of handlers (REST in addition to JSONRPC) I will write the tests as part of separate commit if that's acceptable.

cmd/jwt.go
krishnasrinivas: we don't call toJSONError(errNoAuthToken) anywhere hence it's not needed. This error value is mainly needed to differentiate between errNoAuthToken and errAuthentication so that we reply with "auth error" when call is made with incorrect token on a, for example, public bucket instead of serving the request.

harshavardhana:
Take a look at ``ExecObjectLayerAPIAnonTest`` @krishnasrinivas

cmd/jwt.go
harshavardhana: Ah it is treated like anonymous.

harshavardhana (approved):

PR #3517, opened 2017-01-03
rpc: add post login and retry control callbacks in AuthRPCClient.

balamurugana:
AuthRPCClient.Call() has two addition parameters for post login
callback function and retry configuration.

Post login callback function is called upon successful login where
additional validation of the server can be done.  In
storage-rpc-client, it validates the server by reading/unmarshalling
format.json.

retry configuration has `isRetryableErr` callback function which is
called upon RPC call failure.  isRetryableErr could control to retry
the same RPC call once again by returning true.

## How Has This Been Tested?

`make test`

## Types of changes

## Checklist:


cmd/admin-rpc-client.go
```diff
@@ -53,18 +54,25 @@ func (lc localAdminClient) Restart() error {
 	return nil
 }
 
+func (rc remoteAdminClient) call(serviceMethod string, args interface {
```
harshavardhana: A comment for the wrapper..

cmd/storage-rpc-client.go
```diff
@@ -115,20 +128,33 @@ func newStorageRPC(ep *url.URL) (StorageAPI, error) {
...
+		secureConn:      isSSL(),
+		serviceName:     "Storage",
+	})
+	netStorage.retryCfg = rCfg
+	netStorage.verifyServerFunc = func() error {
+		// FIXME: This is basic server validation check by reading format.json.
+		//        However there could be many possible misconfiguration currently
+		//        we don't handle.
+		// Read format config file to validate the server.
+		buf, err := netStorage.readAll(minioMetaBucket, formatConfigFile, nil)
```
harshavardhana, line 146: To avoid accumulation of network error count and mutate netStorage, we can directly use the rpcClient to do "Storage.ReadAllHandler()" - it avoids further mutation of netStorage.

cmd/storage-rpc-client.go
```diff
@@ -356,15 +369,17 @@ func (n *networkStorage) ReadAll(volume, path string) (buf []byte, err error) {
 		return nil, errFaultyRemoteDisk
 	}
 
-	if err = n.rpcClient.Call("Storage.ReadAllHandler", &ReadAllArgs{
-		Vol:  volume,
-		Path: path,
-	}, &buf); err != nil {
+	args := ReadAllArgs{Vol: volume, Path: path}
```
harshavardhana, line 373: We could have called `Storage.ReadAllHandler` from the `verifyServerFunc()` directly. That way we can avoid incrementing `networkIOErrCount`

cmd/test-utils_test.go
```diff
@@ -366,6 +367,20 @@ func StartTestStorageRPCServer(t TestErrHandler, instanceType string, diskN int)
 	testRPCServer.AccessKey = credentials.AccessKey
 	testRPCServer.SecretKey = credentials.SecretKey
 
+	// Create dummy format json file.
+	emptyConfig := []byte("{}")
```
harshavardhana: Better to write a proper `format.json` hard to remember the purpose here.  Just use `saveFormatFS` ?

harshavardhana (changes requested):

cmd/admin-rpc-client.go
balamurugana: Done

cmd/storage-rpc-client.go
balamurugana: I couldn't able to get why we don't want to track network error.  In the previous style, using `loadFormat()` causes the same behavior like this code.

cmd/test-utils_test.go
balamurugana: This is just a workaround to make `verifyServerFunc()` to work.  Even if we have real/working configuration, its no use.

harshavardhana: It is since it allows to validate. The problem right is that no one will remember why this is done so in the first place.

cmd/storage-rpc-client.go
harshavardhana: The previous code it wasn't possible but it is important since you would want a clean state of StorageAPI. Also additional benefit is to avoid mutation inside a callback.

harshavardhana: Actually keep it like this.. it is necessary for retries.

cmd/test-utils_test.go
harshavardhana: We need to keep it similar to what would be expected in a good condition, a dummy has no real meaning here and no one can build context by looking at these lines.

balamurugana: The real meaning wrt to current logic is, `{}` works which is bad though.  This should be fixed.  Once this is fixed the test case should be fixed else it fails to work.

I can add a `FIXME` me to tell why empty json is used.

harshavardhana: That would make sense.

balamurugana: Done

harshavardhana (approved):

harshavardhana:
@balamurugana can you rebase ?

harshavardhana:
```
	bucket-handlers_test.go:752: Test 3 : Minio FS: Object content differs from expected value.
```

harshavardhana:
Looks like unrelated error.

balamurugana:
> Looks like unrelated error.

Yep. It succeeded second time. I added `expected:` and `got:` content in the error message to debug more.

harshavardhana:
This script hangs

```
$ cat run-dist.sh
#!/bin/bash

$(minio server --address localhost:9001 http://localhost:9001/tmp/1 http://localhost:9002/tmp/2 http://localhost:9003/tmp/3 http://localhost:9004/tmp/4)&
$(minio server --address localhost:9002 http://localhost:9001/tmp/1 http://localhost:9002/tmp/2 http://localhost:9003/tmp/3 http://localhost:9004/tmp/4)&
$(minio server --address localhost:9003 http://localhost:9001/tmp/1 http://localhost:9002/tmp/2 http://localhost:9003/tmp/3 http://localhost:9004/tmp/4)&
minio server --address localhost:9004 http://localhost:9001/tmp/1 http://localhost:9002/tmp/2 http://localhost:9003/tmp/3 http://localhost:9004/tmp/4`
```
with this change. Can you check @balamurugana ?

harshavardhana:
Works without the patch 

```
$ ./run-dist.sh 
Initializing data volume. Waiting for minimum 3 servers to come online. (elapsed 2s)

Initializing data volume.
[01/04] http://localhost:9001/tmp/1 - 55 GiB online
[02/04] http://localhost:9002/tmp/2 - 55 GiB online
[03/04] http://localhost:9003/tmp/3 - 55 GiB online
[04/04] http://localhost:9004/tmp/4 - 55 GiB online

Endpoint:  http://localhost:9004
AccessKey: USWUXHGYZQYFYFFIT3RE 
SecretKey: MOJRH0mkL1IPauahWITSVvyDrQbEEIwljvmxdq03 
Region:    us-east-1
SQS ARNs:  arn:minio:sqs:us-east-1:2:amqp arn:minio:sqs:us-east-1:1:redis

Browser Access:
   http://localhost:9004

   $ mc config host add myminio http://localhost:9004 USWUXHGYZQYFYFFIT3RE MOJRH0mkL1IPauahWITSVvyDrQbEEIwljvmxdq03

Drive Capacity: 58 GiB Free, 110 GiB Total
Status:         4 Online, 0 Offline. We can withstand [2] more drive failure(s).
```

balamurugana:
> with this change. Can you check @balamurugana ?

The problem is with our initialization logic.

When all nodes are fresh, the first node supposes to do the setup.  To do the setup, the first node logs into other nodes.  In the login process, other nodes are trying to verify first node's format.json which is missing.

This should be fixed first.

However I am already working on startup/bootstrap cleanups.  That will fix the prob.

harshavardhana:
> The problem is with our initialization logic.
> 
> When all nodes are fresh, the first node supposes to do the setup. To do the setup, the first node logs into other nodes. In the login process, other nodes are trying to verify first node's format.json which is missing.
> 
> This should be fixed first.
> 
> However I am already working on startup/bootstrap cleanups. That will fix the prob.

But this is working with the master which is what is happening even there.. how is it different in this patch? @balamurugana

harshavardhana:
> When all nodes are fresh, the first node supposes to do the setup. To do the setup, the first node logs into other nodes. In the login process, other nodes are trying to verify first node's format.json which is missing.

In-fact i got this working, but there are some other serious issues.. like ignoring errors etc. making RPC retry like this is tricky since at top level there are many errors that need to be properly handled.

harshavardhana:
@balamurugana is this patch still valid, you might have to rebase.

balamurugana:
> @balamurugana is this patch still valid, you might have to rebase.

Yes.  It is still a valid patch.  There are other fixes required/merged before rebasing this.

krishnasrinivas:
> @deekoder deekoder requested a review from krishnasrinivas 2 hours ago

@deekoder bala says in one of the comments: 
> There are other fixes required/merged before rebasing this.

I will review this sometime though.

deekoder:
Let's discuss on thu. We have to merge or close this pr.

deekoder:
@balamurugana acked on closing this.

PR #3533, opened 2017-01-05
Implement mgmt REST APIs for heal subcommands

krisis:
The heal APIs supported in this change are,
- listing of objects to be healed.
- healing a bucket.
- healing an object.

## Description

* Implement ListObjectsHealHandler and corresponding pkg/madmin API
* Update pkg/madmin API doc
* Add unit test for query param validation function.
* Add unit test for ListObjectsHeal API handler
* Implement HealBucket REST API handler
* Implement HealObject REST API Handler
* Make ListObjectsHeal API similar to ListObjectsV1

## How Has This Been Tested?

Tested using pkg/madmin/examples locally.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/admin-handlers.go
+++ b/cmd/admin-handlers.go
@@ -1,5 +1,5 @@
 /*
- * Minio Cloud Storage, (C) 2016 Minio, Inc.
+ * Minio Cloud Storage, (C) 2016, 2017 Minio, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -20,11 +20,27 @@ import (
 	"encoding/json"
 	"net/http"
 	"net/url"
+	"strconv"
 	"time"
 )
 
 const (
 	minioAdminOpHeader = "X-Minio-Operation"
+)
+
+// Type-safe query params.
+type mgmtQueryKey string
+
+// Only valid query params for list/clear locks management APIs.
+const (
+	mgmtBucket    mgmtQueryKey = "bucket"
+	mgmtObject    mgmtQueryKey = "object"
+	mgmtPrefix    mgmtQueryKey = "prefix"
+	mgmtOlderThan mgmtQueryKey = "older-than"
+	mgmtDelimiter mgmtQueryKey = "delimiter"
+	mgmtMarker    mgmtQueryKey = "marker"
+	mgmtMaxKey    mgmtQueryKey = "max-key"
+	mgmtDryRun    mgmtQueryKey = "dry-run"
 )
 
 // ServiceStatusHandler - GET /?service
@@ -63,32 +79,21 @@ func (adminAPI adminAPIHandlers) ServiceRestartHandler(w http.ResponseWriter, r 
 	}
 
 	// Reply to the client before restarting minio server.
-	w.WriteHeader(http.StatusOK)
+	writeSuccessResponseHeadersOnly(w)
 
 	sendServiceCmd(globalAdminPeers, serviceRestart)
 }
 
-// Type-safe lock query params.
-type lockQueryKey string
-
-// Only valid query params for list/clear locks management APIs.
-const (
-	lockBucket    lockQueryKey = "bucket"
-	lockPrefix    lockQueryKey = "prefix"
-	lockOlderThan lockQueryKey = "older-than"
-)
-
 // validateLockQueryParams - Validates query params for list/clear locks management APIs.
 func validateLockQueryParams(vars url.Values) (string, string, time.Duration, APIErrorCode) {
-	bucket := vars.Get(string(lockBucket))
-	prefix := vars.Get(string(lockPrefix))
-	relTimeStr := vars.Get(string(lockOlderThan))
+	bucket := vars.Get(string(mgmtBucket))
+	prefix := vars.Get(string(mgmtPrefix))
+	relTimeStr := vars.Get(string(mgmtOlderThan))
 
 	// N B empty bucket name is invalid
 	if !IsValidBucketName(bucket) {
 		return "", "", time.Duration(0), ErrInvalidBucketName
 	}
-
 	// empty prefix is valid.
 	if !IsValidObjectPrefix(prefix) {
 		return "", "", time.Duration(0), ErrInvalidObjectName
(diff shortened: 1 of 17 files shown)


cmd/admin-handlers.go
```diff
@@ -207,3 +211,72 @@ func (adminAPI adminAPIHandlers) ClearLocksHandler(w http.ResponseWriter, r *htt
...
+
+	vars := r.URL.Query()
+	bucket, prefix, marker, delimiter, maxKey, adminAPIErr := validateHealQueryParams(vars)
+	if adminAPIErr != ErrNone {
+		writeErrorResponse(w, r, adminAPIErr, r.URL.Path)
+		return
+	}
+
+	objLayer := newObjectLayerFn()
+	if objLayer == nil {
```
harshavardhana, line 282: Return this error in the beginning, don't validate incoming arguments if you can't serve them.

cmd/xl-v1-healing.go
```diff
@@ -170,7 +170,8 @@ func listBucketNames(storageDisks []StorageAPI) (bucketNames map[string]struct{}
 					continue
 				}
 				// Ignore the volume special bucket.
-				if volInfo.Name == minioMetaBucket {
+				if volInfo.Name == minioMetaBucket ||
+					volInfo.Name == minioMetaMultipartBucket {
```
harshavardhana: Convert to a function?

cmd/admin-handlers.go
```diff
@@ -213,3 +219,156 @@ func (adminAPI adminAPIHandlers) ClearLocksHandler(w http.ResponseWriter, r *htt
 	// Reply with list of locks cleared, as json.
 	writeSuccessResponseJSON(w, jsonBytes)
 }
+
+func validateHealQueryParams(vars url.Values) (string, string, string, string, int, APIErrorCode) {
```
harshavardhana, line 205: Oru comment!

pkg/madmin/heal-commands.go
```diff
@@ -0,0 +1,263 @@
...
+import (
+	"encoding/xml"
+	"fmt"
+	"net/http"
+	"net/url"
+	"time"
+)
+
+// listBucketResult container for listObjects response.
+type listBucketResult struct {
```
harshavardhana: s/listBucketResult/listBucketHealResult/

pkg/madmin/heal-commands.go
```diff
@@ -0,0 +1,263 @@
...
+	Prefix     string
+}
+
+// commonPrefix container for prefix response.
+type commonPrefix struct {
+	Prefix string
+}
+
+// ObjectInfo container for object metadata.
+type ObjectInfo struct {
```
harshavardhana, line 86: s/ObjectInfo/ObjectHealInfo

harshavardhana: We might have to add more information like how many disks the data exists or only parity exists etc.

pkg/madmin/heal-commands.go
```diff
@@ -0,0 +1,263 @@
...
+	queryVal.Set("heal", "")
+	queryVal.Set(string(healBucket), bucket)
+	queryVal.Set(string(healPrefix), prefix)
+	queryVal.Set(string(healMarker), marker)
+	queryVal.Set(string(healDelimiter), delimiter)
+	queryVal.Set(string(healMaxKey), maxKeyStr)
+	return queryVal
+}
+
+func (adm *AdminClient) listObjectsHeal(bucket, prefix, delimiter, marker string, maxKeys int) (listBucketResult, error) {
```
harshavardhana: Comment

cmd/admin-handlers.go
```diff
@@ -213,3 +219,156 @@ func (adminAPI adminAPIHandlers) ClearLocksHandler(w http.ResponseWriter, r *htt
...
+	}
+
+	// Get the list objects to be healed.
+	objectInfos, err := objLayer.ListObjectsHeal(bucket, prefix, marker, delimiter, maxKey)
+	if err != nil {
+		writeErrorResponse(w, toAPIErrorCode(err), r.URL)
+		return
+	}
+
+	listResponse := generateListObjectsV1Response(bucket, prefix, marker, delimiter, maxKey, objectInfos)
```
harshavardhana, line 271: You should use a separate function for this for clarity.  We shouldn't reply on the same XML, we need little more data than what is returned right now. For example objects that needs to be healed should return information like 

- How many shards are available? 
- Can healing be performed or not - like object is lost and cannot be recovered?

pkg/madmin/heal-commands.go
```diff
@@ -0,0 +1,263 @@
...
+	resp, err := adm.executeMethod("POST", reqData)
+
+	defer closeResponse(resp)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// HealObject - Heal the given bucket
```
harshavardhana: This is object heal, comment should reflect that.

cmd/admin-handlers.go
```diff
@@ -213,3 +219,176 @@ func (adminAPI adminAPIHandlers) ClearLocksHandler(w http.ResponseWriter, r *htt
...
+
+// validateHealQueryParams - Validates query params for heal list management API.
+func validateHealQueryParams(vars url.Values) (string, string, string, string, int, APIErrorCode) {
+	bucket := vars.Get(string(mgmtBucket))
+	prefix := vars.Get(string(mgmtPrefix))
+	marker := vars.Get(string(mgmtMarker))
+	delimiter := vars.Get(string(mgmtDelimiter))
+	maxKeyStr := vars.Get(string(mgmtMaxKey))
+
+	// N B empty bucket name is invalid
```
vadmeste, line 212: Not sure what 'N B' stands for.

harshavardhana: Its a latin phrase ``nota bene, NB``

cmd/api-response.go
```diff
@@ -197,6 +197,7 @@ type Object struct {
 
 	// The class of storage used to store the object.
 	StorageClass string
+	HealInfo     HealInfo `xml:"HealInfo,omitempty"`
```
vadmeste: omitempty doesn't work here, only with pointers I think.

cmd/object-api-datatypes.go
```diff
@@ -89,6 +95,7 @@ type ObjectInfo struct {
 
 	// User-Defined metadata
 	UserDefined map[string]string
+	HealInfo    HealInfo `xml:"HealInfo,omitempty"`
```
vadmeste: omitempty doesn't work here.

cmd/xl-v1-healing-common.go
```diff
@@ -133,3 +133,21 @@ func xlShouldHeal(partsMetadata []xlMetaV1, errs []error) bool {
 	}
 	return false
 }
+
+func xlHealStat(partsMetadata []xlMetaV1, errs []error) HealInfo {
```
vadmeste: Can you add a comment here ?

cmd/object-api-datatypes.go
```diff
@@ -59,6 +59,12 @@ type BucketInfo struct {
 	Created time.Time
 }
 
+// HealInfo - represents erasure coded chunks in disks
+// TBD: Final choice of datatype
+type HealInfo struct {
+	Shards []bool
```
harshavardhana: We also need to add meaningful info like "Depending" on the shards can this object healed or not on what state the object is.  This would allow `mc` to make intelligent decisions. This value is needed but might not be readily consumable the client and make decisions.

pkg/madmin/API.md
```diff
@@ -121,15 +121,106 @@ If successful restarts the running minio service, for distributed setup restarts
...
+    doneCh := make(chan struct{})
+
+    // Indicate to our routine to exit cleanly upon return.
+    defer close(doneCh)
+
+    // Set true if recursive listing is needed.
+    isRecursive := true
+    // List objects that need healing for a given bucket and
+    // prefix.
+    healObjectCh, err := madmClnt.ListObjectsHeal("mybucket", "myprefix", isRecursive, doneCh)
```
harshavardhana, line 164: This example can talk about ListObjectsHeal and calling -HealObject() on it. With some conditional changes based on HealInfo.

pkg/madmin/heal-commands.go
```diff
@@ -0,0 +1,294 @@
...
+	// Execute POST on /?heal&bucket=mybucket&object=myobject to heal an object.
+	resp, err := adm.executeMethod("POST", reqData)
+
+	defer closeResponse(resp)
+	if err != nil {
+		return err
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return errors.New("Got HTTP Status: " + resp.Status)
```
harshavardhana, line 304: We might have to fix this error message and perhaps provide some meaning like how it is done in minio-go/api-error-response.go

pkg/madmin/heal-commands.go
```diff
@@ -0,0 +1,294 @@
...
+	// Execute POST on /?heal&bucket=mybucket to heal a bucket.
+	resp, err := adm.executeMethod("POST", reqData)
+
+	defer closeResponse(resp)
+	if err != nil {
+		return err
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return errors.New("Got HTTP Status: " + resp.Status)
```
harshavardhana, line 270: Same as below.  Because the returned value might have an ErrorXML we need to extract it.

pkg/madmin/heal-commands.go
```diff
@@ -0,0 +1,294 @@
...
+		DisplayName string `json:"name"`
+		ID          string `json:"id"`
+	} `json:"owner"`
+
+	// The class of storage used to store the object.
+	StorageClass string `json:"storageClass"`
+
+	// Error
+	Err      error `json:"-"`
+	HealInfo *HealInfo
```
harshavardhana: This HealInfo should also have a json representation `json: healInfo`

harshavardhana (changes requested):

cmd/object-api-datatypes.go
krisis: ``mc`` can use ``HealInfo`` to decide if the object can be healed. Could you give an example of what kind of information would be useful?

harshavardhana: This is too low level for `mc` to decide - mc cannot know what a shard means -  server itself should send some meaningful info.  For example like below

```
type StatusType int 

const (
      CanHeal                  StatusType = 0 (Object can be healed)
      Corrupted                StatusType = 1 (Cannot heal)
      QuorumUnavailabe StatusType = 2 (Can attempt heal if quorum is available)
      /// Add more states if necessary. 
)

type HealInfo struct {
      Status                       StatusType
      MissingDataCount    int
      MissingPartityCount int
      HealSize                   int64 (total amount of data + parity to be healed in size)
}
```

`Corrupted` state signifies if `xl.json` is missing from quorum disks,  `xl.json` is corrupted cannot be unmarshalled on any one of the quorum disks.
`QuorumUnavailable` state signifies if insufficient quorum present i.e to many `errDiskNotFound`

`CanHeal` means `xl.json` needs to be healed, along with healing other parts if needed. or just parts need to be healed.

krisis: - ``Corrupted`` status type represents the state where ``N/2`` or more copies of ``xl.json`` are corrupted and __not__ when one of them is, right?

- ``CanHeal`` status type represents the state where there is ``N/2+1`` or more copies of ``xl.json`` 
  are present, right?

pkg/madmin/API.md
krisis: OK. Will do.

pkg/madmin/heal-commands.go
krisis: OK.

pkg/madmin/heal-commands.go
krisis: Yeah, this is missing for other admin commands too. I shall add it for heal subcommands in this PR and handle the remaining in a subsequent PR. Does that make sense?

vadmeste: I am doing this too.. Let me write this and merge my patch first, then we fix others in a separate PR

cmd/object-api-utils.go
```diff
@@ -43,13 +43,24 @@ const (
 var validBucket = regexp.MustCompile(`^[a-z0-9][a-z0-9\.\-]{1,61}[a-z0-9]$`)
 var isIPAddress = regexp.MustCompile(`^(\d+\.){3}\d+$`)
 
+// isMinioBucket returns true if given bucket is a Minio internal
+// bucket and false otherwise.
+func isMinioMetaBucketName(bucket string) bool {
+	if bucket == minioMetaBucket ||
```
harshavardhana: We can simply write this in one liner

```
return bucket == minioMetaBucket || 
                        bucket == minioMetaMultipartBucket ||
                        bucket == minioMetaTmpBucket
```

cmd/xl-v1-healing-common.go
```diff
@@ -133,3 +137,61 @@ func xlShouldHeal(partsMetadata []xlMetaV1, errs []error) bool {
 	}
 	return false
 }
+
+// xlHealStat - returns a structure which describes how many erasure
+// chunks are missing, how many bytes of the object needs to be healed
```
harshavardhana: ``how many bytes`` not anymore i think.

cmd/xl-v1-healing-common.go
```diff
@@ -133,3 +137,61 @@ func xlShouldHeal(partsMetadata []xlMetaV1, errs []error) bool {
...
+	// data and missing parity count.
+	for i, err := range errs {
+		if err == nil {
+			successCount++
+			continue
+		}
+
+		// xl.json is not found, which implies the erasure
+		// coded chunks are unavailable in the corresponding disk.
+		// First half of the disks are data and the rest are parity.
```
harshavardhana, line 174: Actually this logic is not true. - data and parity are truly randomized so basically object1's data disk is perhaps a parity disk for object2 . This shifting of disks happens for each `object` name.

harshavardhana: Its the content of the partsMetadata which will indicate this value.

cmd/xl-v1-healing-common.go
```diff
@@ -133,3 +137,55 @@ func xlShouldHeal(partsMetadata []xlMetaV1, errs []error) bool {
...
+	// Compute heal statistics like bytes to be healed, missing
+	// data and missing parity count.
+	missingDataCount := 0
+	missingParityCount := 0
+
+	for i, err := range errs {
+		// xl.json is not found, which implies the erasure
+		// coded blocks are unavailable in the corresponding disk.
+		// First half of the disks are data and the rest are parity.
+		if realErr := errorCause(err); realErr == errFileNotFound || realErr == errDiskNotFound {
```
harshavardhana, line 175: We can convert this into a function in the next round.

harshavardhana (approved):

PR #3537, opened 2017-01-06
api: Set appropriate content-type for success/error responses.

harshavardhana:
## Description
Golang HTTP client automatically detects content-type but
for S3 clients this content-type might be incorrect or
might misbehave.

For example:

```
Content-Type: text/xml; charset=utf-8
```

Should be
```
Content-Type: application/xml
```

Allow this to be set properly.

## Motivation and Context
S3 compliance.

## How Has This Been Tested?
Manually using - https://gist.github.com/TJC/654e5b2c2e7b27a81929c762addb535e

## Types of changes

## Checklist:

### DIFF
--- a/cmd/api-response.go
+++ b/cmd/api-response.go
@@ -19,6 +19,7 @@ package cmd
 import (
 	"encoding/xml"
 	"net/http"
+	"net/url"
 	"path"
 	"time"
 )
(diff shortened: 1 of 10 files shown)


cmd/api-response.go
```diff
@@ -482,51 +483,55 @@ func generateMultiDeleteResponse(quiet bool, deletedObjects []ObjectIdentifier,
 	return deleteResp
 }
 
-func writeResponse(w http.ResponseWriter, statusCode int, response []byte) {
+func writeResponse(w http.ResponseWriter, statusCode int, response []byte, mimeType string) {
```
krisis: To avoid typos when providing mimeType arguments to cause errors in runtime, we can have a type alias which enumerates commonly used mimetypes in our code. 
e.g,
```go
type mimeType string
const (
  mimeNone mimeType = ""
  mimeJSON mimeType = "application/json"
  ...
)

func writeResponse(w http.ResponseWriter, statusCode int, response []byte, mType mimeType)
```

krisis (approved):

PR #3439, opened 2016-12-13
Add support for Kafka as a notifications target (#2869)

donatello:
## Description

Same as title

## Motivation and Context

Fixes #2869

## How Has This Been Tested?

Testing was done with a Kafka setup and a Kafka consumer to check that events are sent to Kafka. Instructions to reproduce the tests are here: https://gist.github.com/donatello/a6b5ac908b5b7a4e6fb233e84608bb28

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #3550, opened 2017-01-09
Add notifications by webhook. 

alexellis:
## Description

Add notifications by webhook. 

## How Has This Been Tested?

Test by adding this to your `.minio/config.json` file:

```
		"webhook": {
			"1": {
				"enable": true,
				"address": "http://requestb.in/1i9al7m1"
			}
		}
```

Create a bucket called test on myminio, then subscribe to events:

```
mc mb myminio/test
mc events add --json --debug myminio/test arn:minio:sqs:us-east-1:1:webhook
```

Finally, invoke an event on the bucket:

```
mc cp ~/.ssh/id_rsa.pub myminio/test
```

## Motivation and Context

There is extensive coverage for event queues etc but the primitive of the Internet is HTTP, so this fills in that gap.

## Types of changes

## Checklist:

> I had an error about 'DISK' but it was happening before the changes

### DIFF
--- a/cmd/bucket-notification-utils.go
+++ b/cmd/bucket-notification-utils.go
@@ -131,6 +131,7 @@ func isValidQueueID(queueARN string) bool {
 	// Unmarshals QueueARN into structured object.
 	sqsARN := unmarshalSqsARN(queueARN)
 	// Is Queue identifier valid?.
+
 	if isAMQPQueue(sqsARN) { // AMQP eueue.
 		amqpN := serverConfig.GetAMQPNotifyByID(sqsARN.AccountID)
 		return amqpN.Enable && amqpN.URL != ""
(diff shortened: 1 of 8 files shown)


alexellis:
An edit may be required for config-migrate.go

donatello:
> An edit may be required for config-migrate.go

@alexellis Yes - we need to put the new configuration item in a new config version type - it is very similar to the changes done to add kafka support for minio - https://github.com/minio/minio/pull/3439/files (you can ignore the changes in `vendor/*` ofc).

alexellis:
I would think we can make a very minimal version of that script? Do we just need to initialize the new webhooks field with an empty object and we're done?

I assume we also need to create a new class called server-configv12.go (or similar) too? @donatello

harshavardhana:
> I would think we can make a very minimal version of that script? Do we just need to initialize the new webhooks field with an empty object and we're done?

Yes we need to do @alexellis - will send a fix for this PR.

harshavardhana:
Closing this since another PR is opened up with config changes #3566

PR #3557, opened 2017-01-10
event: Enhance event message struct to provide origin server.

### LINKED ISSUE #3556: Notification message (amqp) doesn't contain correct source minio IP
There is no good way to connect back to the minio server that created a notification message.

Note also how my IP in this example is affected by the docker setup and docker links I'm using between minio and rabbitmq and therefore not usable at all.

```
      ...
      "userIdentity": {
        "principalId": "minio"
      },
      "requestParameters": {
        "sourceIPAddress": "172.17.0.1:48212"
      },
      "responseElements": {},
      "s3": {
        "s3SchemaVersion": "1.0",
        "configurationId": "Config",
        "bucket": {
          "name": "video",
          "ownerIdentity": {
            "principalId": "minio"
          },
          "arn": "arn:aws:s3:::video"
        },
        "object": {
          "key": "d%2Fd7%2Fd7b3b561fe44f31e.mp4",
          "size": 15540066,
          "eTag": "087d63de0bcacc7aa787adf18277463c",
          "sequencer": "149804955F81CCFF"
        }
        ...
```

## Expected Behavior
There should be some way of connecting back to Minio from any value in the notification message JSON.

## Current Behavior
There is no such information in the messages and I haven't found any way of forwarding additional values or object/bucket metadata as part of the message data.

## Possible Solution
- the sourceIPAddress? (if it would be the actual IP of the upload request to minio)
or
- principalId? (if I could lead principalId back to an access key instead of the (hardcoded?) value of minio)
- s3['arn'] containing a hostname or minio instance name?
- something else?

1. Configure AMQP and s3:ObjectCreated:* event
2. Upload file
3. Try to connect back to Minio using nothing but message content

## Context
I have multiple minio instances publishing events to the same exchange in rabbitmq and need a way to connect back to the minio instance that created an event.
There are no values in the message that I can use to connect back to the correct minio instance.

## Your Environment
docker setup using minio:edge (more than one container) and rabbitmq using docker links for network.

harshavardhana:
## Description
`principalId` i.e user identity is kept as AccessKey in
accordance with S3 spec.

Additionally responseElements{} are added starting with

`x-amz-request-id` is a hexadecimal of the event time itself in nanosecs.
`x-minio-origin-server` - points to the server generating the event.

## Motivation and Context
Fixes #3556

## How Has This Been Tested?
Manually and using 

```go
package main

import (
	"encoding/json"
	"log"

	"github.com/minio/minio-go"
)

func main() {
	minioClient, err := minio.New("localhost:9000", "USWUXHGYZQYFYFFIT3RE", "MOJRH0mkL1IPauahWITSVvyDrQbEEIwljvmxdq03", false)
	if err != nil {
		log.Fatalln(err)
	}

	// Create a done channel to control 'ListenBucketNotification' go routine.
	doneCh := make(chan struct{})

	// Indicate to our routine to exit cleanly upon return.
	defer close(doneCh)

	// Listen for bucket notifications on "mybucket" filtered by prefix, suffix and events.
	for notificationInfo := range minioClient.ListenBucketNotification("testbucket", "", "", []string{
		"s3:ObjectCreated:*",
	}, doneCh) {
		if notificationInfo.Err != nil {
			log.Fatalln(notificationInfo.Err)
		}
		buf, _ := json.MarshalIndent(notificationInfo, "", "\t")
		log.Println(string(buf))
	}
}
```

## Types of changes

## Checklist:


cmd/server-main.go
```diff
@@ -473,14 +408,20 @@ func serverMain(c *cli.Context) {
 	// Initialize a new HTTP server.
 	apiServer := NewServerMux(serverAddr, handler)
 
-	// If https.
+	// Initialize local server address
+	localMinioAddr := getLocalAddress(srvConfig)
+
+	// Set the global minio addr for this server.
+	globalMinioAddr = localMinioAddr
```
vadmeste: Why we do not set globalMinioAddr directly to getLocalAddress(srvConfig) ?

harshavardhana: Done

vadmeste (approved):

cmd/certs.go
```diff
@@ -145,3 +145,22 @@ func parseCertificateChain(bytes []byte) ([]*x509.Certificate, error) {
...
+// Currently under Windows, there is no way to load system + user CAs at the same time
+func loadRootCAs() {
+	caFiles := mustGetCAFiles()
+	if len(caFiles) == 0 {
+		return
+	}
+	// Get system cert pool, and empty cert pool under Windows because it is not supported
+	globalRootCAs = mustGetSystemCertPool()
+	// Load custom root CAs for client requests
+	for _, caFile := range mustGetCAFiles() {
```
fwessels: Can use caFiles here (instead of calling func again)

fwessels:
LGTM (just one remark above)

PR #3558, opened 2017-01-11
ssl: Set a global boolean to enable SSL across Minio

harshavardhana:
## Description
We have been using `isSSL()` everywhere we can set
a global value once and re-use it again.

## Motivation and Context
Cleanup and some simplification.

## How Has This Been Tested?
Manually using local certs.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/server-main.go
+++ b/cmd/server-main.go
@@ -17,12 +17,14 @@
 package cmd
 
 import (
+	"errors"
 	"fmt"
 	"net"
 	"net/url"
 	"os"
 	"path"
 	"sort"
+	"strconv"
 	"strings"
 
 	"runtime"
@@ -296,11 +298,10 @@ func checkServerSyntax(c *cli.Context) {
 		}
 	}
 
-	tls := isSSL()
 	for _, ep := range endpoints {
-		if ep.Scheme == "https" && !tls {
+		if ep.Scheme == "https" && !globalIsSSL {
 			// Certificates should be provided for https configuration.
-			fatalIf(errInvalidArgument, "Certificates not provided for https")
+			fatalIf(errInvalidArgument, "Certificates not provided for secure configuration")
 		}
 	}
 }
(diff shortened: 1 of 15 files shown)


cmd/server-main.go
```diff
@@ -317,38 +317,55 @@ func isAnyEndpointLocal(eps []*url.URL) bool {
...
+		return "", "", fmt.Errorf("Port cannot be empty or '0', please use `--address` to pick a specific port")
+	}
+
+	// Check if port is available.
+	if err = checkPortAvailability(port); err != nil {
+		return "", "", err
+	}
+
+	// Parse port.
+	if _, err = strconv.Atoi(port); err != nil {
```
krisis: we should parse port before checking port availability.

cmd/server-main_test.go
```diff
@@ -63,20 +63,66 @@ func TestGetListenIPs(t *testing.T) {
...
+	}{
+		// Test 1 - successful.
+		{
+			addr: ":" + getFreePort(),
+			err:  nil,
+		},
+		// Test 2 port empty.
+		{
+			addr: ":0",
+			err:  errors.New("Port cannot be empty or '0', please use `--address` to pick a specific port"),
```
bweston92: @harshavardhana should we not define these somewhere so we can just reference to them?

main.go
```
var errEmptyPortSpecified = errors.New("Port cannot be empty or '0', please use `--address` to pick a specific port")

...
return errEmptyPortSpecified
...
```

main_test.go
```
{
  addr: ":" + getFreePort(),
  err: errEmptyPortSpecified,
}
```

krisis (changes requested):

cmd/server-main.go
harshavardhana: Oh yes that is correct.

cmd/server-main_test.go
harshavardhana: We could define but since they are not used elsewhere it's not defined in that manner, but not a bad idea either.

bweston92: Ok will leave it in your hands :)

vadmeste (approved):

PR #3574, opened 2017-01-13
docker: Add ARM docker container dockerfile.

harshavardhana:
## Description
Docker ARM docker container.

## Motivation and Context
Attempt to get ARM docker container for Minio working. 

## How Has This Been Tested?
Not yet. 

## Types of changes

## Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,9 @@
 go_import_path: github.com/minio/minio
 sudo: required
 
+services:
+ - docker
+
 dist: trusty
 
 language: go
@@ -13,12 +16,28 @@ env:
 - ARCH=i686
 
 script:
+# Refer https://blog.hypriot.com/post/setup-simple-ci-pipeline-for-arm-images/
+# prepare qemu
+- docker run --rm --privileged multiarch/qemu-user-static:register --reset
+# build image
+- docker build -t minio/minio:edge-armhf . -f Dockerfile.armhf
+# test image
+- docker run --rm minio/minio:edge-armhf version
+
+## Run all the tests
 - make
 - make test GOFLAGS="-timeout 15m -race -v"
 - make coverage
+
+# push image
+- >
+  if [ "$TRAVIS_BRANCH" == "master" ] && [ "$TRAVIS_PULL_REQUEST" == "false" ]; then
+    docker login -u="$DOCKER_USER" -p="$DOCKER_PASS"
+    docker push minio/minio:edge-armhf
+  fi
 
 after_success:
 - bash <(curl -s https://codecov.io/bash)
 
 go:
-- 1.7.4
+- 1.7.5
--- a/Dockerfile.armhf
+++ b/Dockerfile.armhf
@@ -1,4 +1,4 @@
-FROM alpine:3.5
+FROM resin/armhf-alpine:3.5
 
 ENV GOPATH /go
 ENV PATH $PATH:$GOPATH/bin


Dockerfile.armhf
```diff
@@ -0,0 +1,13 @@
+FROM armhf/alpine:latest
+
+ENV GOPATH /go
+
+RUN apk add --update --no-cache curl ca-certificates git golang && \
```
endersonmaia: The package name should be `go`

https://pkgs.alpinelinux.org/package/v3.4/community/armhf/go

endersonmaia:
I had to change the package name form `golang` to `go` and I get this error : 


    /go/src/github.com/minio/minio # go install -ldflags "$(go run buildscripts/gen-ldflags.go)"
    # github.com/minio/minio/cmd
    cmd/certs.go:81: undefined: x509.SystemCertPool
    cmd/notify-webhook.go:77: unknown http.Transport field 'DialContext' in struct literal
    /go/src/github.com/minio/minio # echo $?
    2

harshavardhana:
> /go/src/github.com/minio/minio # go install -ldflags "$(go run buildscripts/gen-ldflags.go)"
> \# github.com/minio/minio/cmd
> cmd/certs.go:81: undefined: x509.SystemCertPool
> cmd/notify-webhook.go:77: unknown http.Transport field 'DialContext' in struct literal
> /go/src/github.com/minio/minio # echo $?
> 2

@endersonmaia the dockerfiles have been updated to use new alpine images for both armhf and aarch64.  Would you mind validating it again and let us know? 

Thanks

alexellis:
Also pinging @vielmetti the Pack.net ARM64 champion ;-)

@harshavardhana - I can test the armhf version - do you have an image pushed to the hub already?

Dockerfile.armhf
```diff
@@ -0,0 +1,20 @@
+FROM resin/armhf-alpine:3.5
```
alexellis, line 1: I'd recommend trying armhf/alpine:3.5 which is an "official" version from the Docker team via @tianon

harshavardhana:
> @harshavardhana - I can test the armhf version - do you have an image pushed to the hub already?

Not yet will update in a bit.

Dockerfile.armhf
```diff
@@ -0,0 +1,20 @@
...
+
+ENV GOPATH /go
+ENV PATH $PATH:$GOPATH/bin
+
+RUN apk update -y
+RUN apk add --no-cache git ca-certificates go g++
```
alexellis: I think a much smaller binary could be produced if you built and published the ARM binary through one Dockerfile and then used `ADD https://....` in another. Then you'd only need Alpine or Scratch as a base image and this could be tiny.

Dockerfile.aarch64
```diff
@@ -0,0 +1,20 @@
+FROM resin/aarch64-debian:3.5
```
alexellis: Please compare with https://hub.docker.com/r/aarch64/debian/ (which is an official Docker image)

.travis.yml
```diff
@@ -13,6 +16,15 @@ env:
 - ARCH=i686
 
 script:
+# prepare qemu
```
alexellis: Could you include a link to Stefan's blog post on qemu in this comment? I assume that's where the line is from right?

alexellis (changes requested):
Suggesting some changes.

.travis.yml
harshavardhana: yes @alexellis

Dockerfile.armhf
harshavardhana: The reason do it this way is to build for every commit to the master, downloading only from the released binary won't produce right builds for building directly from master purposes.

Dockerfile.aarch64
harshavardhana: Actually i moved and changed it to `minio/golang-aarch64:1.7-alpine`

Dockerfile.armhf
harshavardhana: The reason to use `resin/armhf` since it bundles qemu so that it can be built on travis.

harshavardhana: Interesting thing about resin/armhf-alpine is that its based on `armhf/alpine` with quemu support transparent.

harshavardhana:
@alexellis - Got rid of the ARM64 for now, and address it later with some other proper testing. armhf for the time being builds properly - shouldn't have issues.

vielmetti:
When you're ready to test the ARMv8 (aarch64) version I'll be here, @harshavardhana @alexellis .

I'll start a separate issue for `multiarch` support for minio, since that's a next logical path.

PR #3580, opened 2017-01-14
admin: Add service set creds API

### LINKED ISSUE #3569: [admin] Add an API to change credentials using `mc admin service`
## Expected Behavior
Provide a way to change the server credentials, it shouldn't be allow if environment variable is set. (Just like web-browser)

## Current Behavior
There is no way to change the credentials using `mc admin` 

## Possible Solution
Implement a new API to change credentials and but disallow this if https is not configured on the client or the request is over plain HTTP (to avoid possible security issues)

This is a feature.

## Context
Being able to change passwords. 

## Your Environment
Current minio master.

vadmeste:
## Description
Add Set Creds admin API

## Motivation and Context
Fixes #3569 

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:


cmd/admin-handlers.go
```diff
@@ -86,6 +86,61 @@ func (adminAPI adminAPIHandlers) ServiceRestartHandler(w http.ResponseWriter, r
 	sendServiceCmd(globalAdminPeers, serviceRestart)
 }
 
+// Type-safe service query params.
+type serviceQueryKey string
+
+const (
+	serviceAccessKey serviceQueryKey = "accessKey"
```
harshavardhana: We should send access and secretKey as part of the body insteadof query params for security reasons. On a HTTPS connection this becomes more secure just like how browser actually sets the credentials.

cmd/api-errors.go
```diff
@@ -574,6 +575,12 @@ var errorCodeResponse = map[APIErrorCode]APIError{
 		Description:    "Server not initialized, please try again.",
 		HTTPStatusCode: http.StatusServiceUnavailable,
 	},
+	ErrOpNotAcceptable: {
+		Code:           "XMinioOpNotAcceptable",
+		Description:    "Requested operation is not accepted due to unsatisfied conditions.",
```
harshavardhana:
```
MethodNotAllowed	The specified method is not allowed against this resource.	405 Method Not Allowed	Client
``` 

http://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html  - is perhaps better here.

pkg/madmin/service.go
```diff
@@ -141,3 +141,28 @@ func (adm *AdminClient) ServiceRestart() error {
...
+	// Execute GET on bucket to list objects.
+	resp, err := adm.executeMethod("POST", reqData)
+
+	defer closeResponse(resp)
+	if err != nil {
+		return err
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return errors.New("Got HTTP Status: " + resp.Status)
```
harshavardhana: Extract the response XML body. For `mc` 'Got HTTP Status' is useless.

cmd/admin-handlers.go
vadmeste: > We should send access and secretKey as part of the body insteadof query params for security reasons. On a HTTPS connection this becomes more secure just like how browser actually sets the credentials.

With HTTPS, even query params are encrypted, However, it seems that it is still uncommon and could be leaked in server logs etc.., so will do it.

harshavardhana: But query params are visible to naked eye.

vadmeste:
Comments are addressed, but I did other changes too,

@harshavardhana, @krisis, can you review this PR ?

cmd/admin-router.go
```diff
@@ -36,6 +36,8 @@ func registerAdminRouter(mux *router.Router) {
 
 	// Service restart
 	adminRouter.Methods("POST").Queries("service", "").Headers(minioAdminOpHeader, "restart").HandlerFunc(adminAPI.ServiceRestartHandler)
+	// Service update credentials
+	adminRouter.Methods("POST").Queries("service", "").Headers(minioAdminOpHeader, "setCredentials").HandlerFunc(adminAPI.ServiceCredentialsHandler)
```
harshavardhana: For http methods and header - to be consistent we can use 'set-credentials'  similar to how 'region' is set as 'us-east-1'

cmd/api-errors.go
```diff
@@ -574,6 +576,12 @@ var errorCodeResponse = map[APIErrorCode]APIError{
 		Description:    "Server not initialized, please try again.",
 		HTTPStatusCode: http.StatusServiceUnavailable,
 	},
+	ErrAdminBadCreds: {
+		Code:           "XMinioBadCreds",
+		Description:    "Credentials are not suitable for use.",
```
harshavardhana: Propose a more useful new message.

docs/admin-api/management-api.md
```diff
@@ -32,6 +27,12 @@
   - x-minio-operation: status
   - Response: On success 200, return json formatted StorageInfo object.
 
+* SetCredentials
+  - GET /?service
+  - x-minio-operation: setCredentials
```
harshavardhana: `x-minio-operation: set-credentials`

harshavardhana: You missed this @vadmeste

vadmeste: > You missed this @vadmeste

yes thanks. I am also going to add errors description that could be returned by this API.

cmd/admin-handlers_test.go
```diff
@@ -172,20 +187,37 @@ func testServicesCmdHandler(cmd cmdType, t *testing.T) {
 		}
 	}
 
+	if cmd == setCreds {
+		// Check current
```
krisis: Could you expand the comment to explain the intent clearly?

cmd/api-errors.go
```diff
@@ -574,6 +576,12 @@ var errorCodeResponse = map[APIErrorCode]APIError{
 		Description:    "Server not initialized, please try again.",
 		HTTPStatusCode: http.StatusServiceUnavailable,
 	},
+	ErrAdminBadCred: {
+		Code:           "XMinioBadCred",
+		Description:    "Credential is mal-formed.",
```
krisis: s/mal-formed/malformed/
malformed is a non-hyphenated word.

cmd/admin-handlers.go
```diff
@@ -68,6 +70,72 @@ func (adminAPI adminAPIHandlers) ServiceRestartHandler(w http.ResponseWriter, r
...
+	if err != nil {
+		errorIf(err, "Cannot unmarshal credentials request")
+		writeErrorResponse(w, ErrMalformedXML, r.URL)
+		return
+	}
+
+	// Check passed credentials
+	cred, err := getCredential(req.Username, req.Password)
+	if err != nil {
+		writeErrorResponse(w, ErrAdminBadCred, r.URL)
```
krisis: Do we need a new error value ``ErrAdminBadCred``? ``getCredential`` would return one of ``errInvalidAcessKeyLength`` or ``errInvalidSecretKeyLength``. They would be more specific than ``ErrAdminBadCred``, i.e, whether the accesskey is incorrect or the secretkey is incorrect (or both).

vadmeste: > Do we need a new error value ErrAdminBadCred? getCredential would return one of errInvalidAcessKeyLength or errInvalidSecretKeyLength. They would be more specific than ErrAdminBadCred, i.e, whether the accesskey is incorrect or the secretkey is incorrect (or both).

errInvalidAcessKeyLength and errInvalidSecretKeyLength are standard errors, we need an ErrorResponse type which represents the regular S3 xml error like:
```
 27 <?xml version="1.0" encoding="UTF-8"?>
 28 <Error>
 29    <Code>AccessDenied</Code>
 30    <Message>Access Denied</Message>
 31    <BucketName>bucketName</BucketName>
 32    <Key>objectName</Key>
 33    <RequestId>F19772218238A85A</RequestId>
 34    <HostId>GuWkjyviSiGHizehqpmsD1ndz5NClSP19DOT+s2mv7gXGQ8/X1lhbDGiIJEXpGFD</HostId>
 35 </Error>
```

vadmeste:
I addressed your issues @krisis, also I added set credentials example under pkg/madmin/examples/

pkg/madmin/examples/service-set-credentials.go
```diff
@@ -0,0 +1,44 @@
...
+	madmClnt, err := madmin.New("your-minio.example.com:9000", "YOUR-ACCESSKEYID", "YOUR-SECRETACCESSKEY", true)
+	if err != nil {
+		log.Fatalln(err)
+	}
+
+	err = madmClnt.ServiceSetCredentials("YOUR-NEW-ACCESSKEY", "YOUR-NEW-SECRETKEY")
+	if err != nil {
+		log.Fatalln(err)
+	}
+	log.Println("New credentials successfully sent.")
```
krisis: nit: "New credentials successfully set" (not sent).

vadmeste: Done

krisis (approved):

cmd/admin-handlers.go
harshavardhana: We can construct them to a more palpable behavior and provide two distinct errors one for secret and another for access key rather than one single error "Malformed creds". This provides more context.

harshavardhana:
Can you resolve the conflicts @vadmeste ?

vadmeste:
I increased the timeout of go test here because it seems it is little slow under Windows. Otherwise, all issues are addressed.

harshavardhana:
> I increased the timeout of go test here because it seems it is little slow under Windows. Otherwise, all issues are addressed.

We need to investigate - can you open a bug?

harshavardhana (approved):

vadmeste:
> We need to investigate - can you open a bug?

I noticed that there is one test which put 1000 small objects and that is taking 2 minutes under Windows. By the way, -race flag is activated. Anyway, as we are writing new tests, execution will take more time and probably this is not a bug. Will open a new issue.

harshavardhana:
> I noticed that there is one test which put 1000 small objects and that is taking 2 minutes under Windows. By the way, -race flag is activated. Anyway, as we are writing new tests, execution will take more time and probably this is not a bug. Will open a new issue

Agreed but we need to investigate further.

PR #3597, opened 2017-01-19
check if have free space before creating file

andreikop:
Check if having available space before writing file to the storage

## Description

This PR adds check for free space for PutObject PutObjectPart and CopyObject.

## Motivation and Context

We often face a situation when 
1. Files are written to minio until the storage is 100% full
2. Minio get's stopped
3. Minio refuses to start because of [this check](https://github.com/minio/minio/blob/master/cmd/fs-v1.go#L157)

It is impossible to start minio and delete some files from S3 to free space. The only way to get minio working again is mounting it's volume and manually deleting some files.

After this patch it will be still possible to get 100% full storage, i.e. by concurrent requests, but the probability is much less. Need to discuss, why minio refuses to start if FS is full. I'd prefer to have storage available for Read and Delete.

## How Has This Been Tested?

There are tests for PutObject and PutObjectPart.
I've tried to make a test for CopyObject, but it requres stateful storage backend mock and is much more complicated then the code itself.

## Types of changes

This patch decreases available storage capacity by `minFreeSpace`

## Checklist:

### DIFF
--- a/cmd/fs-v1.go
+++ b/cmd/fs-v1.go
@@ -83,6 +83,12 @@ func initMetaVolumeFS(fsPath, fsUUID string) error {
 
 }
 
+// Get disk space which is available for storing data
+func availableDiskSpace(totalSpace int64) int64 {
+	// Remove 5% from free space for cumulative disk space used for journalling, inodes etc.
+	return int64(float64(totalSpace) * 0.95)
+}
+
 // newFSObjectLayer - initialize new fs object layer.
 func newFSObjectLayer(fsPath string) (ObjectLayer, error) {
 	if fsPath == "" {
@@ -189,9 +195,7 @@ func (fs fsObjects) checkDiskFree() (err error) {
 		return err
 	}
 
-	// Remove 5% from free space for cumulative disk space used for journalling, inodes etc.
-	availableDiskSpace := float64(di.Free) * 0.95
-	if int64(availableDiskSpace) <= fs.minFreeSpace {
+	if availableDiskSpace(di.Free) <= fs.minFreeSpace {
 		return errDiskFull
 	}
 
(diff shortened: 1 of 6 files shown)


harshavardhana:
Minio already uses `PrepareFIle()` API for this which should disallow any file creates if the i/o.. which version of Minio are you running? also are you using XL or Standalone FS backend?

harshavardhana:
Can you validate this @vadmeste ?  current master v/s released Minio server?

andreikop:
It seems like we are using version, which doesn't have `PrepareFile`, because we face 100 full disk.
I've tested master before patching it. The problem is reproducible. `PrepareFile` checks if disk is free before writing a file, but it doesn't check if available space would be critical after writing a file.
It should be possible to modify check in `PrepareFile`.

andreikop:
I'd prefer closing this PR and merging #3606

vadmeste:
FS is user friendly and can be started on the rootfs or C:\. Users can fill their drive with any possible stuffs and not always through Minio, so I think it is better to avoid checking free space at boot time as @andreikop suggested.

However, @andreikop if you think that the disk got filled with Minio itself, then this is another bug that need to be fixed.

vadmeste:
@harshavardhana, with FS, when I have 1.6 Gb remaining in the disk space and I want to upload a 1GB file, the disk will get filled and out of space error will appear. It is probably because in FS, we need more space than expected (2 * 1 Gb ?)

harshavardhana:
> @harshavardhana, with FS, when I have 1.6 Gb remaining in the disk space and I want to upload a 1GB file, the disk will get filled and out of space error will appear. It is probably because in FS, we need more space than expected (2 * 1 Gb ?)

Yes for multipart we will be using 2 * x - due to background concat operation. @vadmeste

harshavardhana:
> I'd prefer closing this PR and merging #3606

Closing this PR.

PR #3606, opened 2017-01-20
fs: start even if there are not enough free space

andreikop:
This is continuation of activity started at #3597 

Since minio already has check for free space in `PrepareFile`, it would be better if it starts regardless of available free space. So, user can delete some files and free space, but can't put new files until minio has enough space.


harshavardhana:
Thanks @andreikop

PR #3607, opened 2017-01-20
server-mux: Add tcp idle read timeout

vadmeste:
## Description
This PR closes connections opened by silent clients, the default timeout time is set to 30 seconds.

## Motivation and Context
Avoid many idle client connections i.e client didn't send any data until a given stipulated amount of time. 

## How Has This Been Tested?
Manually using a telnet-bomb.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/server-mux.go
+++ b/cmd/server-mux.go
@@ -103,6 +103,8 @@ func (c *ConnMux) PeekProtocol() string {
 // Read - streams the ConnMux buffer when reset flag is activated, otherwise
 // streams from the incoming network connection
 func (c *ConnMux) Read(b []byte) (int, error) {
+	// Push read deadline
+	c.Conn.SetReadDeadline(time.Now().Add(defaultTCPReadTimeout))
 	return c.bufrw.Read(b)
 }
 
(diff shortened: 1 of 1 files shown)


cmd/server-mux.go
```diff
@@ -194,6 +199,9 @@ func newListenerMux(listener net.Listener, config *tls.Config) *ListenerMux {
 				return
 			}
 
+			// Enable Read timeout
+			conn.SetReadDeadline(time.Now().Add(defaultTcpReadTimeout))
+
 			// Enable keep alive for each connection.
```
harshavardhana, line 205: Tested this thoroughly with some other changes the problem is that once this happens for example 

```
2017/01/22 11:52:22 http: Accept error: accept tcp [::]:9000: accept4: too many open files; retrying in 5ms
``` 

Golang stops responding especially "AccepTCP()" is hung for ever.

harshavardhana: What we need is doing "conn" list ourselves i.e basically make sure that we don't run out of RLimit.Max and handle it ourselves.. So that we can reject new connection attempts gracefully.

vadmeste: > Tested this thoroughly with some other changes the problem is that once this happens for example
> 
> 2017/01/22 11:52:22 http: Accept error: accept tcp [::]:9000: accept4: too many open files; retrying in 5ms
> Golang stops responding especially "AccepTCP()" is hung for ever.

This change doesn't avoid "too many open files" error, it just evicts silent clients, that are silent for a long time (we can define what this long time is), the server could be easily DOSed otherwise.

vadmeste: > What we need is doing "conn" list ourselves i.e basically make sure that we don't run out of RLimit.Max and handle it ourselves..

This is not only about the maximum opened files but also about having limited port range (< 65536) 

> So that we can reject new connection attempts gracefully.

Probably this is more than luxury, this is a rare scenario and our users will already try to avoid being limited by port numbers by using other methods.

harshavardhana (approved):

PR #3616, opened 2017-01-24
signature-v2: Use request.RequestURI for signature calculation.

krishnasrinivas:
## Description

http.Request.RequestURI should be used for Signature V2 calculation because:
* using url.Path will give incorrect results because it would contain decoded values. i.e if client used %2F for signature calculation url.Path will contain "/" and hence cause incorrect signature calculation
* using url.RawPath is unreliable, for ex. it does not get set if %20 is used for space

http.Request.RequestURI is the only reliable way to calculate signature v2.

## How Has This Been Tested?

manual and make test

## Types of changes

## Checklist:

### DIFF
--- a/cmd/signature-v2.go
+++ b/cmd/signature-v2.go
@@ -85,15 +85,12 @@ func doesPresignV2SignatureMatch(r *http.Request) APIErrorCode {
 	// Access credentials.
 	cred := serverConfig.GetCredential()
 
-	// url.RawPath will be valid if path has any encoded characters, if not it will
-	// be empty - in which case we need to consider url.Path (bug in net/http?)
-	encodedResource := r.URL.RawPath
-	encodedQuery := r.URL.RawQuery
-	if encodedResource == "" {
-		splits := strings.Split(r.URL.Path, "?")
-		if len(splits) > 0 {
-			encodedResource = getURLEncodedName(splits[0])
-		}
+	// r.RequestURI will have raw encoded URI as sent by the client.
+	splits := strings.Split(r.RequestURI, "?")
+	encodedResource := splits[0]
+	encodedQuery := ""
+	if len(splits) > 1 {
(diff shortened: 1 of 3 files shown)


cmd/signature-v2.go
```diff
@@ -85,15 +85,12 @@ func doesPresignV2SignatureMatch(r *http.Request) APIErrorCode {
...
-	// be empty - in which case we need to consider url.Path (bug in net/http?)
-	encodedResource := r.URL.RawPath
-	encodedQuery := r.URL.RawQuery
-	if encodedResource == "" {
-		splits := strings.Split(r.URL.Path, "?")
-		if len(splits) > 0 {
-			encodedResource = getURLEncodedName(splits[0])
-		}
+	// r.RequestURI will have raw encoded URI as sent by the client.
+	splits := strings.Split(r.RequestURI, "?")
```
vadmeste: We have splitStr() which could simplify a little this bloc

```
splits := splitStr(r.RequestURI, "?", 2)
encodedResource, encodedQuery := splits[0], splits[1]
```

cmd/signature-v2.go
```diff
@@ -213,20 +210,14 @@ func doesSignV2Match(r *http.Request) APIErrorCode {
...
-	//   url.RawPath will be valid if path has any encoded characters, if not it will
-	//   be empty - in which case we need to consider url.Path (bug in net/http?)
-	encodedResource := r.URL.RawPath
-	if encodedResource == "" {
-		splits := strings.Split(r.URL.Path, "?")
-		if len(splits) > 0 {
-			encodedResource = getURLEncodedName(splits[0])
-		}
+	// r.RequestURI will have raw encoded URI as sent by the client.
+	splits := strings.Split(r.RequestURI, "?")
```
vadmeste: Same as above.

vadmeste (approved):

PR #3618, opened 2017-01-24
Add dry-run query param for HealFormat API

krisis:
## Description

This change supports dry-run option in heal-format API, similar to heal-object and heal-bucket.

## Motivation and Context

This allows users to validate the heal format command before executing it for real.
 
## How Has This Been Tested?

Manually tested.
## Types of changes

## Checklist:

- [  ] I have added tests to cover my changes.

### DIFF
--- a/cmd/admin-handlers.go
+++ b/cmd/admin-handlers.go
@@ -398,7 +398,7 @@ func (adminAPI adminAPIHandlers) ListBucketsHealHandler(w http.ResponseWriter, r
 	writeSuccessResponseXML(w, encodeResponse(listResponse))
 }
 
-// HealBucketHandler - POST /?heal&bucket=mybucket
+// HealBucketHandler - POST /?heal&bucket=mybucket&dry-run
 // - x-minio-operation = bucket
 // - bucket is mandatory query parameter
 // Heal a given bucket, if present.
@@ -425,7 +425,7 @@ func (adminAPI adminAPIHandlers) HealBucketHandler(w http.ResponseWriter, r *htt
 		return
 	}
 
-	// if dry-run=yes, then only perform validations and return success.
+	// if dry-run is present in query-params, then only perform validations and return success.
 	if isDryRun(vars) {
 		writeSuccessResponseHeadersOnly(w)
 		return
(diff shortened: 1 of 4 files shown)


pkg/madmin/examples/heal-format.go
```diff
@@ -34,13 +34,23 @@ func main() {
 
 	// API requests are secure (HTTPS) if secure=true and insecure (HTTPS) otherwise.
 	// New returns an Minio Admin client object.
-	madmClnt, err := madmin.New("your-minio.example.com:9000", "YOUR-ACCESSKEYID", "YOUR-SECRETACCESSKEY", true)
+	madmClnt, err := madmin.New("localhost:9000", "minio", "minio123", false)
```
harshavardhana: any reason to pick this or was it just testing locally?

pkg/madmin/examples/heal-format.go
```diff
@@ -34,13 +34,23 @@ func main() {
...
-	// Heal storage format on available disks.
-	err = madmClnt.HealFormat()
+	// Attempt healing format in dry-run mode.
+	isDryRun := true
+	err = madmClnt.HealFormat(isDryRun)
+	if err != nil {
+		log.Fatalln(err)
+	}
+
+	log.Println("dry run to heal format on available disks was successful.")
```
harshavardhana: We don't need this i guess if you wish to keep we need to word this better.

cmd/admin-handlers.go
```diff
@@ -501,7 +502,7 @@ func (adminAPI adminAPIHandlers) HealObjectHandler(w http.ResponseWriter, r *htt
 	writeSuccessResponseHeadersOnly(w)
 }
 
-// HealFormatHandler - POST /?heal
+// HealFormatHandler - POST /?heal&dry-run=[yes|no]
```
harshavardhana: Was thinking that we can avoid making "yes" or no for this. But instead have queryParam itself optional i.e without any value.

This means to heal
```
POST /?heal
``` 

This means fake heal
```
POST /?heal&dry-run
```

Thoughts?

pkg/madmin/examples/heal-format.go
krisis: It needs to be removed. I used it to test the changes locally.

pkg/madmin/examples/heal-format.go
krisis: OK

cmd/admin-handlers.go
krisis: Ok, I shall make the necessary changes to the other heal APIs that support this too.

PR #3620, opened 2017-01-24
madmin: Add SetTimeout API for admin http requests

vadmeste:
## Description
Add new SetTimeout API

## Motivation and Context
We didn't have this API before, though we need it to configure a http request timeout 

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:

### DIFF
--- a/pkg/madmin/api.go
+++ b/pkg/madmin/api.go
@@ -31,6 +31,7 @@ import (
 	"regexp"
 	"runtime"
 	"strings"
+	"time"
 
 	"github.com/minio/minio-go/pkg/s3signer"
 	"github.com/minio/minio-go/pkg/s3utils"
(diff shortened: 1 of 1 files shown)


harshavardhana:
why this type of an API? @vadmeste what is useful is SetCustomTransport which is already there.. Can you elaborate on the use case for this?

vadmeste:
> why this type of an API? @vadmeste what is useful is SetCustomTransport which is already there.. Can you elaborate on the use case for this?

As I found, Transport doesn't help to set real timeout (there is only Idle timeout and Header timeout), http.Client's timeout is more generic and covers more cases.

harshavardhana:
> As I found, Transport doesn't help to set real timeout (there is only Idle timeout and Header timeout), http.Client's timeout is more generic and covers more cases.

But why do we need any other timeout other than idle or header? i am not sure we need other timeouts.

vadmeste:
For now, the user can set timeout for header response which is enough to cover most of real world cases, closing this PR.

PR #3621, opened 2017-01-24
browser: add yarn.lock and .gitignore files.

krishnasrinivas:
## Description

Add yarn.lock and .gitignore for browser/


browser/.gitignore
```diff
@@ -0,0 +1,18 @@
+**/*.swp
+cover.out
+*~
+minio
+!*/
+site/
```
harshavardhana, line 6: This needs 'node_modules' as well.

harshavardhana: and also 'production' directory generated by 'yarn release'

PR #3629, opened 2017-01-25
madmin: Fix args order in listObjectsHeal()

vadmeste:
## Description
The order of args in listObjectsHeal() internal function were switched, which will give wrong result in case
of a non recursive objects heal list.

## Motivation and Context
Fix a found bug

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:

### DIFF
--- a/pkg/madmin/heal-commands.go
+++ b/pkg/madmin/heal-commands.go
@@ -182,7 +182,7 @@ func mkHealQueryVal(bucket, prefix, marker, delimiter, maxKeyStr string) url.Val
 }
 
 // listObjectsHeal - issues heal list API request for a batch of maxKeys objects to be healed.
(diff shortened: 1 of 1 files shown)


harshavardhana (commented):
The commit message should be wrapped upto 80 lines, don't write the entire thing in a single line. Split them into paragraphs. 

`git log` should be easy to read.

vadmeste:
> The commit message should be wrapped upto 80 lines, don't write the entire thing in a single line. Split them into paragraphs.
> 
> git log should be easy to read.

Done.

harshavardhana (approved):

harshavardhana:
BTW can we add a test for this to avoid order mistakes like this @vadmeste ?

vadmeste:
> BTW can we add a test for this to avoid order mistakes like this @vadmeste ?

@krsis and I were thinking of a way of testing (we also need it in minio-go to avoid testing against a real server). One possible solution is to import minio/minio/cmd package, run Minio and run tests against it.

PR #3635, opened 2017-01-26
Make Docker command more useful

alexellis:
Updated command to:

- Run in background and show logs for Access keys
- Give the container a name so it is easily accessible later
- Provided example with bind-mount because beyond experimenting this is the core way to use Minio in Docker.

@harshavardhana

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,17 +5,34 @@ Minio is an object storage server released under Apache License v2.0. It is comp
 Minio server is light enough to be bundled with the application stack, similar to NodeJS, Redis and MySQL.
 
 ## Docker Container
+
 ### Stable
+
+```sh
+docker run -p 9000:9000 --name minio -d minio/minio server /export
 ```
-docker pull minio/minio
-docker run -p 9000:9000 minio/minio server /export
+
+
+The `--name` parameter means you can check the Minio container easily, the `-p` override sets the port Minio will be accessible by, `9000` is the default and `-d` runs Minio in the background as a daemon.
+
+Then retrieve your `AccessKey` and `SecretKey` with the following command:
+
+```sh
+docker logs minio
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -5,17 +5,34 @@ Minio is an object storage server released under Apache License v2.0. It is comp
 Minio server is light enough to be bundled with the application stack, similar to NodeJS, Redis and MySQL.
 
 ## Docker Container
+
 ### Stable
+
+```sh
+docker run -p 9000:9000 --name minio -d minio/minio server /export
+```
```
harshavardhana, line 13: We can move this to `docs/docker` instead provide one single line command here.. Just like how it is done. with Binary downloads.

alexellis: I think at least the single command needs to be more verbose to be useful - especially with the `-d` flag for those who are new to Docker, it will be confusing to have Minio running in the foreground. We're very familiar with the CLI etc, but think about the range of Minio users.

Also as we discussed the `pull` command is not necessary.

harshavardhana: That is true but verbosity here is large and shadowing others making it inconsistent.. That is why it is not needed to add options like `--name` and `-d` . They can be in more explanatory section and that can be referenced here to proceed for further reading. 

Just to get this running all they need is below to have Minio running. 

```
docker run -p 9000:9000 minio/minio server /export
``` 

Even with the explanation it is still incomplete we need to explain about how to have persistent access keys etc using environment variables or being able to persist `config.json` etc.

harshavardhana:
http://docs.minio.io/docs/minio-docker-quickstart-guide has been updated with more explanations so this change is not needed anymore.

@alexellis feel free to close this PR or send any updated fixes for docker README.md

PR #3636, opened 2017-01-26
Simplify running cmd.Main() from external tests

vadmeste:
## Description

An external test that runs cmd.Main() has a difficulty to set cmd arguments
and MINIO_{ACCESS,SECRET}_KEY values, this commit changes a little the current
behavior in a way that helps external tests.

## Motivation and Context
Packages such madmin or minio-go rely on an already configured server to run tests on, this
PR is a first step to simplify running Minio server from external tests.

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -190,6 +190,8 @@ func minioInit(ctx *cli.Context) {
 	enableLoggers()
 
 	// Fetch access keys from environment variables and update the config.
+	globalEnvAccessKey = os.Getenv("MINIO_ACCESS_KEY")
+	globalEnvSecretKey = os.Getenv("MINIO_SECRET_KEY")
 	if globalEnvAccessKey != "" && globalEnvSecretKey != "" {
 		// Set new credentials.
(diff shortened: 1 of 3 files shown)


cmd/main.go
```diff
@@ -224,5 +226,7 @@ func Main() {
 	}
 
 	// Run the app - exit on error.
-	app.RunAndExitOnError()
+	if err := app.Run(args); err != nil {
+		os.Exit(1)
```
harshavardhana: You should take a callback for example for `cmd.Main(args []string, exitFunc func())`  - this way you can change the behavior by default it will be `os.Exit(1)` upon error. 

Otherwise for tests you can set a exit mock function.

vadmeste: > You should take a callback for example for cmd.Main(args []string, exitFunc func()) - this way you can change the behavior by default it will be os.Exit(1) upon error.
> 
> Otherwise for tests you can set a exit mock function.

Done!

Issue #3540, opened 2017-01-07
Restricting Access to a Bucket and/or File to HTTP Referrer


ZiiX:
## Expected Behavior
S3 buckets allow you to restrict what sites can show files as [outlined on their site](https://docs.aws.amazon.com/AmazonS3/latest/dev/example-bucket-policies.html#example-bucket-policies-use-case-4).

## Current Behavior
Minio's difference only allows for a few [policy features](http://docs.minio.io/docs/minio-client-complete-guide#policy) and [share options](https://docs.minio.io/docs/minio-client-complete-guide#share).

## Possible Solution
It would be great to incorporate this feature on an individual file 'share' level as well as on a per-bucket 'policy' level for one, many or *.domains.  Also, adding to the browser-based create bucket button with the option to limit to referring domain(s) as well as an individual new "copy link" feature-set to the online minio browser with http referrer limit, time limit, etc options would be great.

Other solutions I'm not aware of to increase privacy and security would be welcome too!
## Context
I want to: 
- save bandwidth from folks posting my media links to their sites & sharing them from my minio server
-  limit hotlinking and protect files better that are currently served behind a password & force that file to stay behind a password protected site. 

## Your Environment
My environment is NGINX + Minio + Lets Encrypt

thank you for your time & my awesome minio!

harshavardhana:
Ping @krisis are planning to look into this ?

krisis:
Yes, I will be working on it.

PR #3641, opened 2017-01-27
Add aws:Referer condition key support.

krisis:
## Description

This change implements bucket policy enhancements required to restrict access based on HTTP referer.
See https://docs.aws.amazon.com/AmazonS3/latest/dev/example-bucket-policies.html#example-bucket-policies-use-case-4 for more information.

## Motivation and Context

See https://github.com/minio/minio/issues/3540
## How Has This Been Tested?

Manually and with unit tests.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/auth-handler.go
+++ b/cmd/auth-handler.go
@@ -125,7 +125,8 @@ func checkRequestAuthType(r *http.Request, bucket, policyAction, region string) 
 
 	if reqAuthType == authTypeAnonymous && policyAction != "" {
 		// http://docs.aws.amazon.com/AmazonS3/latest/dev/using-with-s3-actions.html
-		return enforceBucketPolicy(bucket, policyAction, r.URL)
+		return enforceBucketPolicy(bucket, policyAction, r.URL.Path,
+			r.Referer(), r.URL.Query())
 	}
 
 	// By default return ErrAccessDenied
--- a/cmd/bucket-handlers.go
+++ b/cmd/bucket-handlers.go
@@ -33,7 +33,7 @@ import (
 
 // http://docs.aws.amazon.com/AmazonS3/latest/dev/using-with-s3-actions.html
 // Enforces bucket policies for a bucket for a given tatusaction.
-func enforceBucketPolicy(bucket string, action string, reqURL *url.URL) (s3Error APIErrorCode) {
+func enforceBucketPolicy(bucket, action, resource, referer string, queryParams url.Values) (s3Error APIErrorCode) {
 	// Verify if bucket actually exists
 	if err := checkBucketExist(bucket, newObjectLayerFn()); err != nil {
 		err = errorCause(err)
@@ -57,16 +57,21 @@ func enforceBucketPolicy(bucket string, action string, reqURL *url.URL) (s3Error
 	}
 
 	// Construct resource in 'arn:aws:s3:::examplebucket/object' format.
-	resource := bucketARNPrefix + strings.TrimSuffix(strings.TrimPrefix(reqURL.Path, "/"), "/")
+	arn := bucketARNPrefix + strings.TrimSuffix(strings.TrimPrefix(resource, "/"), "/")
 
 	// Get conditions for policy verification.
 	conditionKeyMap := make(map[string]set.StringSet)
-	for queryParam := range reqURL.Query() {
-		conditionKeyMap[queryParam] = set.CreateStringSet(reqURL.Query().Get(queryParam))
+	for queryParam := range queryParams {
+		conditionKeyMap[queryParam] = set.CreateStringSet(queryParams.Get(queryParam))
+	}
+
+	// Add request referer to conditionKeyMap if present.
+	if referer != "" {
+		conditionKeyMap["referer"] = set.CreateStringSet(referer)
 	}
 
 	// Validate action, resource and conditions with current policy statements.
-	if !bucketPolicyEvalStatements(action, resource, conditionKeyMap, policy.Statements) {
+	if !bucketPolicyEvalStatements(action, arn, conditionKeyMap, policy.Statements) {
 		return ErrAccessDenied
 	}
 	return ErrNone
--- a/cmd/object-handlers.go
+++ b/cmd/object-handlers.go
@@ -56,10 +56,9 @@ func setGetRespHeaders(w http.ResponseWriter, reqParams url.Values) {
 func errAllowableObjectNotFound(bucket string, r *http.Request) APIErrorCode {
 	if getRequestAuthType(r) == authTypeAnonymous {
 		//we care about the bucket as a whole, not a particular resource
-		url := *r.URL
-		url.Path = "/" + bucket
-
-		if s3Error := enforceBucketPolicy(bucket, "s3:ListBucket", &url); s3Error != ErrNone {
+		resource := "/" + bucket
+		if s3Error := enforceBucketPolicy(bucket, "s3:ListBucket", resource,
+			r.Referer(), r.URL.Query()); s3Error != ErrNone {
 			return ErrAccessDenied
 		}
 	}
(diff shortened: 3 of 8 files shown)


krisis:
@harshavardhana When I manually test this change with curl, I get the following error html. Is there something I might be missing?
The policy used is,
```
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Sid": "Sample get object bucket policy",
      "Action": [
        "s3:GetObject"
      ],
      "Effect": "Allow",
      "Principal": {
        "AWS": [
          "*"
        ]
      },
      "Resource": [
        "arn:aws:s3:::test/*"
      ],
      "Condition": {
        "StringLike": {
          "aws:Referer": "www.example.*"
        }
      }
    }
  ]
}

```
```sh
➜ curl GET --referer "www.example.com"  http://localhost:9000/test/obj1
<html>
<head><title>403 Forbidden</title></head>
<body bgcolor="white">
<center><h1>403 Forbidden</h1></center>
<hr><center>nginx</center>
</body>
</html>
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=16.04
DISTRIB_CODENAME=xenial
DISTRIB_DESCRIPTION="Ubuntu 16.04.1 LTS"
```

harshavardhana:
Can you paste the nginx proxy config as well ? @krisis

krisis:
@harshavardhana, I didn't use ngnix proxy.

harshavardhana:
> @harshavardhana, I didn't use ngnix proxy.

The error HTML looks like comes from nginx 
```
➜ curl GET --referer "www.example.com"  http://localhost:9000/test/obj1
<html>
<head><title>403 Forbidden</title></head>
<body bgcolor="white">
<center><h1>403 Forbidden</h1></center>
<hr><center>nginx</center>
</body>
</html>
```

harshavardhana:
This seems to be working fine locally @krisis 

```
$ curl --referer "www.example.com" http://localhost:9000/test/1
Ubuntu 16.04.1 LTS \n \l
```

Oh looks like your curl command is wrong `GET` should be provided with `-X GET` 
```
curl GET --referer "www.example.com"  http://localhost:9000/test/obj1
```

krishnasrinivas:
how do we allow setting of referrer? using `mc`?

harshavardhana:
> how do we allow setting of referrer? using mc?

This is not for `mc` it is for users who have a special `policy.json` requirement which contains `referrer`.

cmd/bucket-handlers.go
```diff
@@ -33,7 +32,7 @@ import (
 
 // http://docs.aws.amazon.com/AmazonS3/latest/dev/using-with-s3-actions.html
 // Enforces bucket policies for a bucket for a given tatusaction.
-func enforceBucketPolicy(bucket string, action string, reqURL *url.URL) (s3Error APIErrorCode) {
+func enforceBucketPolicy(bucket string, action string, req http.Request) (s3Error APIErrorCode) {
```
krishnasrinivas: I think it's generally a better practice to pass only the values that a function needs, in this case req.URL and referrer and not pass the request as a whole which contains other things not needed by this function. Function signature becomes self documenting - in the sense it tells us exactly what it depends on.

But if the function depends on most of the things contained in http.Request then it would be better to pass it entirely instead of making the argument very long compromising on code readability.

cmd/object-handlers.go
```diff
@@ -440,7 +443,7 @@ func (api objectAPIHandlers) PutObjectHandler(w http.ResponseWriter, r *http.Req
 		return
 	case authTypeAnonymous:
 		// http://docs.aws.amazon.com/AmazonS3/latest/dev/using-with-s3-actions.html
-		if s3Error := enforceBucketPolicy(bucket, "s3:PutObject", r.URL); s3Error != ErrNone {
+		if s3Error := enforceBucketPolicy(bucket, "s3:PutObject", *r); s3Error != ErrNone {
```
krishnasrinivas: doing `*r` does not provide much protection as many of the fields in Request structure are reference based and hence not provide protection against mutability.

cmd/auth-handler.go
```diff
@@ -125,7 +125,7 @@ func checkRequestAuthType(r *http.Request, bucket, policyAction, region string)
```
harshavardhana, line 125: We should also update `docs/bucket/policy/README.md`

krishnasrinivas:
> how do we allow setting of referrer? using mc?

>> This is not for mc it is for users who have a special policy.json requirement which contains referrer.

editing things by hand might be tricky. also we have to change the minio client libs so that they are aware of `referrer` when updating the server.

harshavardhana:
Tricky or otherwise that is what AWS SDK users do. This has no meaning for us in our SDKs nor on mc. We can certainly enhance them by providing this functionality but a canned approach was purposefully choosen to avoid doing these type of low level things. The users of this feature generally would use an existing policy on S3 and use it with minio . AWS policy generator allows for generating simple policy JSON files so it is not something new.

cmd/auth-handler.go
krisis: An entry for each of StringLike, StringNotLike and AWS:Referer is already present as part of this PR. Let me know if the document needs additional changes.

cmd/bucket-handlers.go
krisis: OK, I was not aware of this practice in our code base. I will address this.

cmd/object-handlers.go
krisis: OK, I was not very hopeful about protection against mutability here :) My attempt was only symbolic within the limits of the language.

krisis:
Discussions around whether we should have this feature should be moved to the issue. It will allow the author of this issue to weigh in.

cmd/bucket-handlers.go
krishnasrinivas: note: we don't follow this practice in our code base, just a personal observation that it's a better practice

krishnasrinivas:
> This has no meaning for us in our SDKs nor on mc

we'll have to be mindful of referrer settings that the policy might already have when we make changes to the bucket policy.

harshavardhana:
Certainly all existing policies should persist in all our SDK implementations. Basically if someone attempts to remove it won't be removed . If this is not how it is done so then it's a bug.

harshavardhana (approved):

ZiiX:
I realize y'all are trying to stick to S3 bucket commands & I'm certain this will be appreciated for cdn folks and advanced users. The bucket policies do seem to be passed from some apps to aws that I have found so this is certainly a great route with folks that know how to use the aws policy generator (it's overwhelming to me but my simple policy i use is to restrict it from being shown outside my site with the example next).  I don't pass things from site to bucket and would like the minio to restrict server-side as opposed to being passed commands if that's what you wanted clarification on? Please let me know if it's appropriate for me to comment here. 
`{
	"Version": "2012-10-17",
	"Id": "http referer policy example",
	"Statement": [
		{
			"Sid": "Allow get requests originating from www.example.com and example.com.",
			"Effect": "Allow",
			"Principal": "*",
			"Action": "s3:GetObject",
			"Resource": "arn:aws:s3:::bucket-name/*",
			"Condition": {
				"StringLike": {
					"aws:Referer": [
						"https://www.domain.tld/*",
						"https://domain.tld/*"
					]
				}
			}
		},
		{
			"Effect": "Allow",
			"Principal": {
				"AWS": "arn:aws:iam::#########:user/#######"
			},
			"Action": "s3:*",
			"Resource": [
				"arn:aws:s3:::bucket-name",
				"arn:aws:s3:::bucket-name/*"
			]
		}
	]
}`
However, if I may, I thought I'd bring up another approach could also be to incorporate a host tld whitelist per-bucket(s) on the UI? I'm not a dev and would like it as a simple file storage solution to deliver my website images/videos and have found the aws arena somewhat poor in UX for novice folks like myself. For example, my vimeo video settings allow videos to be only played on a specific domain list (https/http) with a simple add/delete on the UI.  Also, being able to incorporate time-expiring links would be nifty too.

Either way, this is great work and it's appreciated. Y'all are amazing. I like my minio as a decentralized aws standalone file storage bucket & look forward to making my own redundant cdn

harshavardhana:
> I realize y'all are trying to stick to S3 bucket commands & I'm certain this will be appreciated for cdn folks and advanced users. The bucket policies do seem to be passed from some apps to aws that I have found so this is certainly a great route with folks that know how to use the aws policy generator (it's overwhelming to me but my simple policy i use is to restrict it from being shown outside my site with the example next). I don't pass things from site to bucket and would like the minio to restrict server-side as opposed to being passed commands if that's what you wanted clarification on? Please let me know if it's appropriate for me to comment here.

It is true the S3 API in this regard is overwhelming, we can indeed add an API for this in minio-go and our SDKs to make this simpler. But making this as a default behavior on server for all buckets sets us away from S3 and causes confusion with existing APIs.

Please open a feature/enhancement request which talks in detail about the simplification that you are looking for we can discuss internally and let you know.

ZiiX:
Will do. But yes, this feature y'all have added will help me/others protect media (& bandwidth) the same way I've done with aws by locking it to my server.tld as referrer only. Thank you.

PR #3643, opened 2017-01-27
Validate command line arguments for valid endpoints.

balamurugana:
This patch validates all combination of endpoints for FS/XL/Distribute
setup.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/endpoint.go
+++ b/cmd/endpoint.go
@@ -1,0 +1,189 @@
+/*
+ * Minio Cloud Storage, (C) 2017 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cmd
+
+import (
+	"fmt"
+	"net"
+	"net/url"
+	"sort"
+	"strconv"
+	"strings"
+
+	"github.com/minio/minio-go/pkg/set"
+)
+
+// IsValidDistribution - checks whether given count is a valid distribution for erasure coding.
+func IsValidDistribution(count int) bool {
+	return (count >= 4 && count <= 16 && count%2 == 0)
+}
+
+// CheckLocalServerAddr - checks if serverAddr is valid and local host.
+func CheckLocalServerAddr(serverAddr string) error {
+	serverAddr = strings.TrimSpace(serverAddr)
+	if serverAddr == "" {
+		// empty server address means local host.
+		return nil
+	}
+
+	host, port, err := net.SplitHostPort(serverAddr)
+	if err != nil {
+		return err
+	}
+
+	// Check whether port is a valid port number.
+	p, err := strconv.Atoi(port)
+	if err != nil {
+		return fmt.Errorf("invalid port number in server address")
+	} else if p < 1 {
+		return fmt.Errorf("port number should be greater than zero in server address")
+	}
+
+	if host != "" {
+		localIPs := mustGetLocalIP4()
+		hostIPs, err := getHostIP4(host)
+		if err != nil {
+			return err
+		}
+
+		if localIPs.Intersection(hostIPs).IsEmpty() {
+			return fmt.Errorf("host in server address should be this server")
+		}
+	}
+
+	return nil
+}
+
+// IsEmptyPath - check whether given path is not empty.
+func IsEmptyPath(path string) bool {
+	return path == "" || path == "." || path == "/" || path == `\`
+}
+
+// EndpointType - enum for endpoint type.
+type EndpointType int
+
+const (
+	// UnknownEndpointType - unknown endpoint type enum.
+	UnknownEndpointType EndpointType = 0
+
+	// PathEndpointType - path style endpoint type enum.
+	PathEndpointType = 1
+
+	// URLEndpointType - URL style endpoint type enum.
+	URLEndpointType = 2
+)
+
+// Endpoint - any type of endpoint.
+type Endpoint struct {
+	Value      string
+	URL        *url.URL
+	IsDisabled bool
+	IsLocal    bool
+}
+
+// Type - returns type of endpoint.
+func (endpoint Endpoint) Type() EndpointType {
+	if IsEmptyPath(endpoint.Value) {
+		return UnknownEndpointType
+	}
+
+	if endpoint.URL != nil {
+		return URLEndpointType
+	}
+
+	return PathEndpointType
+}
+
+// NewEndpoint - returns new endpoint based on given arguments.
+func NewEndpoint(arg string) (Endpoint, error) {
+	if IsEmptyPath(arg) {
+		return Endpoint{}, fmt.Errorf("Empty or root endpoint is not supported")
+	}
+
+	u, err := url.Parse(arg)
+	if err == nil && u.Host != "" {
+		// URL style of endpoint.
+		if !((u.Scheme == "http" || u.Scheme == "https") &&
+			u.Opaque == "" && u.ForceQuery == false && u.RawQuery == "" && u.Fragment == "") {
+			return Endpoint{}, fmt.Errorf("Unknown endpoint format")
+		}
+
+		host, port, err := splitHostPort(u.Host)
+		if !(err == nil && host != "") {
+			return Endpoint{}, fmt.Errorf("Invalid host in endpoint format")
+		}
+
+		if IsEmptyPath(u.Path) {
+			return Endpoint{}, fmt.Errorf("Empty or root path is not supported in URL endpoint")
+		}
+
+		u.Host = net.JoinHostPort(host, port)
+		arg = u.String()
+	} else {
+		u = nil
+	}
+
+	return Endpoint{
+		Value: arg,
+		URL:   u,
+	}, nil
+}
+
+// EndpointList - list of same type of endpoint.
+type EndpointList []Endpoint
+
+// NewEndpointList - returns new endpoint list based on input args.
+func NewEndpointList(args ...string) (endpoints EndpointList, err error) {
+	// Check whether given args contain duplicates.
+	if uniqueArgs := set.CreateStringSet(args...); len(uniqueArgs) != len(args) {
+		return nil, fmt.Errorf("duplicate endpoints found")
+	}
+
+	// Check whether no. of args are valid for XL distribution.
+	if !IsValidDistribution(len(args)) {
+		return nil, fmt.Errorf("total endpoints %d found. For XL/Distribute, it should be 4, 6, 8, 10, 12, 14 or 16", len(args))
+	}
+
+	sort.Strings(args)
+
+	var endpointType EndpointType
+	var scheme string
+
+	// Loop through args and adds to endpoint list.
+	for i, arg := range args {
+		endpoint, err := NewEndpoint(arg)
+		if err != nil {
+			return nil, fmt.Errorf("unknown endpoint format %s", arg)
+		}
+
+		// All endpoints have to be same type and scheme if applicable.
+		if i == 0 {
+			endpointType = endpoint.Type()
+			if endpoint.URL != nil {
+				scheme = endpoint.URL.Scheme
+			}
+		} else if endpoint.Type() != endpointType {
+			return nil, fmt.Errorf("mixed style endpoints are not supported")
+		} else if endpoint.URL != nil && scheme != endpoint.URL.Scheme {
+			return nil, fmt.Errorf("mixed scheme is not supported")
+		}
+
+		endpoints = append(endpoints, endpoint)
+	}
+
+	return endpoints, nil
+}
--- a/cmd/net.go
+++ b/cmd/net.go
@@ -1,0 +1,105 @@
+/*
+ * Minio Cloud Storage, (C) 2017 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cmd
+
+import (
+	"fmt"
+	"net"
+	"strconv"
+	"strings"
+
+	"github.com/minio/minio-go/pkg/set"
+)
+
+const defaultPort = "9000"
+
+// splitHostPort is wrapper to net.SplitHostPort(), but ignores "missing port in address".
+// If port is missing, default port "9000" is returned.
+func splitHostPort(hostPort string) (host, port string, err error) {
+	host, port, err = net.SplitHostPort(hostPort)
+	if err != nil {
+		// Ignore "missing port in address" error as we use default port.
+		// This happens when hostPort = "" or hostPort = "server".
+		if !strings.Contains(err.Error(), "missing port in address") {
+			return "", "", err
+		}
+
+		host = hostPort
+		port = defaultPort
+		err = nil
+	}
+
+	p, err := strconv.Atoi(port)
+	if err != nil {
+		return "", "", fmt.Errorf("Invalid port number")
+	} else if p < 1 {
+		return "", "", fmt.Errorf("Port number should be greater than zero")
+	}
+
+	return host, port, err
+}
+
+func mustSplitHostPort(hostPort string) (host, port string) {
+	host, port, err := splitHostPort(hostPort)
+	if err != nil {
+		fatalIf(err, "Unable to split host port %s", hostPort)
+	}
+
+	return host, port
+}
+
+// mustGetLocalIP4 returns IPv4 addresses of local host.  It panics on error.
+func mustGetLocalIP4() (ipList set.StringSet) {
+	ipList = set.NewStringSet()
+	addrs, err := net.InterfaceAddrs()
+	if err != nil {
+		fatalIf(err, "Unable to get IP addresses of this host.")
+	}
+
+	for _, addr := range addrs {
+		var ip net.IP
+		switch v := addr.(type) {
+		case *net.IPNet:
+			ip = v.IP
+		case *net.IPAddr:
+			ip = v.IP
+		}
+
+		if ip.To4() != nil {
+			ipList.Add(ip.String())
+		}
+	}
+
+	return ipList
+}
+
+// getHostIP4 returns IPv4 address of given host.
+func getHostIP4(host string) (ipList set.StringSet, err error) {
+	ipList = set.NewStringSet()
+	ips, err := net.LookupIP(host)
+	if err != nil {
+		return ipList, err
+	}
+
+	for _, ip := range ips {
+		if ip.To4() != nil {
+			ipList.Add(ip.String())
+		}
+	}
+
+	return ipList, err
+}
--- a/cmd/setup.go
+++ b/cmd/setup.go
@@ -1,0 +1,303 @@
+/*
+ * Minio Cloud Storage, (C) 2017 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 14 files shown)


harshavardhana:
@balamurugana is this patch ready to be merged? - i see a FIXME is it still a WIP?

balamurugana:
> @balamurugana is this patch ready to be merged? - i see a FIXME is it still a WIP?

@harshavardhana Its ready to merge.  The FIXME is to use return value of `NewSetup()` where the global endpoints and other variables will be removed by that.

harshavardhana:
@vadmeste and @krishnasrinivas can you guys review this ?

harshavardhana:
> @harshavardhana Its ready to merge. The FIXME is to use return value of NewSetup() where the global endpoints and other variables will be removed by that.

Can you open a github issue for this FIXME?

harshavardhana:
> Can you open a github issue for this FIXME?

Once this patch is merged..

cmd/net.go
```diff
@@ -0,0 +1,110 @@
...
+// splitHostPort is wrapper to net.SplitHostPort(), but ignores "missing port in address".
+// If port is missing, default port "9000" is returned.
+func splitHostPort(hostPort string) (host, port string, err error) {
+	host, port, err = net.SplitHostPort(hostPort)
+	// Ignore "missing port in address" error as we use default port.
+	if err != nil && strings.Contains(err.Error(), "missing port in address") {
+		host = hostPort
+		port = "9000"
+		err = nil
+	}
```
vadmeste: 'else' missing here? when err != nil and the error is not missing port in address

cmd/net.go
```diff
@@ -0,0 +1,110 @@
...
+	host, port, err := splitHostPort(hostPort)
+	if err != nil {
+		panic(err)
+	}
+
+	return host, port
+}
+
+// mustGetLocalIP4 returns IPv4 addresses of local host.  It panics on error.
+func mustGetLocalIP4() (ipList set.StringSet) {
```
vadmeste, line 66: It looks like we are only supporting IP v4 for the moment, can't we add support of IP v6 at the same time?

cmd/endpoint.go
```diff
@@ -0,0 +1,145 @@
...
+
+const (
+	// UnknownEndpointType - unknown endpoint type enum.
+	UnknownEndpointType EndpointType = 0
+
+	// FSEndpointType - FS endpoint type enum.
+	FSEndpointType = 1
+
+	// XLEndpointType - XL endpoint type enum.
+	XLEndpointType = 2
```
vadmeste: I think it will be less confusing if we call this RemoteEndpointType

cmd/endpoint.go
```diff
@@ -0,0 +1,145 @@
...
+func NewEndpoint(arg string, index int) (Endpoint, error) {
+	if IsEmptyPath(arg) {
+		return Endpoint{}, fmt.Errorf("Empty or root endpoint is not supported")
+	}
+
+	u, err := url.Parse(arg)
+	if err == nil && u.Host != "" {
+		// May be XL type endpoint arg.
+		host, port, err := splitHostPort(u.Host)
+		if !(err == nil && host != "" && (u.Scheme == "http" || u.Scheme == "https") &&
```
vadmeste: At this point, I think we can check for Scheme, Opaque, ForceQuery, RawQuery and Fragment and also IsEmptyPath() just before calling splitHostPost(). It will be more simple to read.

cmd/endpoint.go
```diff
@@ -0,0 +1,145 @@
...
+
+// NewEndpointList - returns new endpoint list based on input args.
+func NewEndpointList(args ...string) (endpoints EndpointList, err error) {
+	// Check whether given args contain duplicates.
+	if uniqueArgs := set.CreateStringSet(args...); len(uniqueArgs) != len(args) {
+		return nil, fmt.Errorf("duplicate endpoints found")
+	}
+
+	// Check whether no. of args are valid for XL distribution.
+	if !IsValidDistribution(len(args)) {
```
vadmeste, line 157: IsValidDistribution() doesn't accept len(args) = 1, which means NewEndpointList works only with XL/Distributed (not obvious), can we make this check outside this function, it will be less confusing.

cmd/endpoint.go
balamurugana: I feel `RemoteEndpointType` is unclear.  Basically endpoint can be either path style or URL style.  How about changing something like `PathEndpointType` and `URLEndpointType` ?

cmd/endpoint.go
balamurugana: @vadmeste `if` has `host != ""`.  The check requires after splitting `u.Host`.  Do you feel to move out this check alone outside of this `if`?

cmd/endpoint.go
balamurugana: I couldn't able to get you.

NewEndpointList() is applicable only to XL or Distribute setup.  The arguments can be multiples to such type of setup and we would need to fail if `len(args)` are unsupported.

Could you tell me why `len(args) = 1` is a valid case here?

cmd/net.go
balamurugana: Done

cmd/net.go
balamurugana: As per my reading of current code base, we do not support/use IPv6 and the current code checks for IPv4 only.

balamurugana:
> Can you open a github issue for this FIXME?

@harshavardhana done

cmd/endpoint.go
vadmeste: > I feel RemoteEndpointType is unclear. Basically endpoint can be either path style or URL style. How about changing something like PathEndpointType and URLEndpointType ?

:+1:

cmd/endpoint.go
vadmeste: > Do you feel to move out this check alone outside of this if?

Yes. Splitting the if would be better for readability.

cmd/endpoint.go
vadmeste: > NewEndpointList() is applicable only to XL or Distribute setup.

My point is that NewEndpointList() name doesn't indicate that this is about XL or Distributed. And, moving IsValidDistribution() outside NewEndpointList() will make the latter work with FS too (in case if we need something like this later)

cmd/endpoint.go
balamurugana: Done

cmd/endpoint.go
balamurugana: I think you are talking about FS type of `Setup` which is already taken care in `FSSetup`.

As you see in `NewEndpointList()`, any type of endpoints can be added but they are all have to be same type ie either `FSEndpointType` or `XLEndpointType`

If you are taking about FS/XL/Distribute setup, they are defined in `setup.go` where they use `NewEndpointList()`.

The idea of keeping/doing check of `IsValidDistribution()` at `NewEndpointList()` is `EndpointList` can not have a wrong set of endpoints.

Do you say `EndpointList` could have any length of endpoints where `XLSetup` and `DistributeSetup` should take care of distribution check?

vadmeste: > Do you say EndpointList could have any length of endpoints where XLSetup and DistributeSetup should take care of distribution check?

Yes. Or you can rename NewEndpointList() to NewXLEndpointList(), my concern is mostly about naming.

balamurugana: @vadmeste 

> Yes. Or you can rename NewEndpointList() to NewXLEndpointList(), my concern is mostly about naming.

I like the renaming approach, but one thing about `NewXLEndpointList()` is something not same as golang's style like a `<T>` and `New<T>()` for creating new `<T>` object.

Let me know whether I go ahead of renaming.

harshavardhana:
// cc @krishnasrinivas can you review this?

krishnasrinivas (commented):
@balamurugana tried to review it but found it difficult. Because the code has been written for future use. And without knowing how the changes in this commit fits w.r.t the other code we have it is difficult to review.

For ex:
```
+type Endpoint struct {
 +	Index      int
 +	Value      string
 +	IsDisabled bool
 +	URL        *url.URL
 +}
```
When you see this code, the following questions arise:
* How/Where does the Index field gets used?
* How/Where is IsDisabled used?

Another example:
NewSetup() returns an interface of the type Setup, again the questions are:
* is this interface really useful?
* how does it get used with other parts of our code base?

I think this is what you would want to address when the FIXME you put in there is fixed? It is best to bring that change as well in the commit so that everything can be reviewed as a whole to make sure the code we take in is doing the right thing.

balamurugana:
> I think this is what you would want to address when the FIXME you put in there is fixed? It is best to bring that change as well in the commit so that everything can be reviewed as a whole to make sure the code we take in is doing the right thing.

:+1: I am doing this.

harshavardhana:
@balamurugana do you know when this will be ready to be reviewed again ?

harshavardhana:
There are some conflicts in this patch @balamurugana can you rebase?

cmd/setup.go
```diff
@@ -0,0 +1,303 @@
...
+	// DistXLSetupType - Distributed XL setup type enum.
+	DistXLSetupType = 3
+)
+
+// Setup - interface denotes any minio setup.
+type Setup interface {
+	ServerAddr() string
+	Endpoints() EndpointList
+	Type() SetupType
+}
```
krishnasrinivas, line 46: @balamurugana having an interface type is not providing any advantage here. i.e you can do it as struct instead of interface and you will see that a lot of code will go away. i.e this struct is all you'll need.

```
type Setup Struct {
    serverAddr string
    endPoints []Endpoint
    type SetupType
}
```

But what is the goal of doing these changes? is it to bring all the setup related global variables into a single global struct so as not to pollute the global name space?

on a side note:
```
    serverAddr string
    endPoints []Endpoint
```

does not seem to be useful once init is completed.

cmd/endpoint.go
```diff
@@ -0,0 +1,189 @@
...
+	Value      string
+	URL        *url.URL
+	IsDisabled bool
+	IsLocal    bool
+}
+
+// Type - returns type of endpoint.
+func (endpoint Endpoint) Type() EndpointType {
+	if IsEmptyPath(endpoint.Value) {
+		return UnknownEndpointType
```
krishnasrinivas, line 101: We'll never have an `UnknownEndpointType` because NewEndpoint would have returned error if it was empty.

so `Value` field in the Endpoint struct is not useful.

looking at other variables:
`IsDisabled` - is not used anywhere
`IsLocal` - is not needed because we already have isLocalStorage() which is used only during init process. If isLocalStorage() was used during operations (i.e not just during init) then it would make sense to store it in a struct variable like this so that we don't unnecessarily call isLocalStorage() after init process

so having Endpoint struct like this does not look like it is adding any advantage. url.URL is enough to represent an endpoint.

cmd/setup.go
krishnasrinivas: note that:
```
    serverAddr string
    endPoints []Endpoint
```

does not seem to be useful once init is completed.

cmd/net.go
```diff
@@ -0,0 +1,106 @@
...
+import (
+	"fmt"
+	"net"
+	"strconv"
+	"strings"
+
+	"github.com/minio/minio-go/pkg/set"
+)
+
+const defaultPort = "9000"
```
krishnasrinivas, line 28: Note that this is not needed as we already have:
```
var serverFlags = []cli.Flag{
        cli.StringFlag{
                Name:  "address",
                Value: ":9000",
                Usage: `Bind to a specific IP:PORT. Defaults to ":9000".`,
        },
}
```

cmd/net.go
```diff
@@ -0,0 +1,106 @@
...
+	} else if p < 1 {
+		return "", "", fmt.Errorf("Port number should be greater than zero")
+	}
+
+	return host, port, err
+}
+
+func mustSplitHostPort(hostPort string) (host, port string) {
+	host, port, err := splitHostPort(hostPort)
+	if err != nil {
```
krishnasrinivas, line 58: instead of panic we can do fatalIf and print a friendly message

cmd/net.go
```diff
@@ -0,0 +1,106 @@
...
+
+	"github.com/minio/minio-go/pkg/set"
+)
+
+const defaultPort = "9000"
+
+// splitHostPort is wrapper to net.SplitHostPort(), but ignores "missing port in address".
+// If port is missing, default port "9000" is returned.
+func splitHostPort(hostPort string) (host, port string, err error) {
+	host, port, err = net.SplitHostPort(hostPort)
```
krishnasrinivas, line 33: the following code from previous code getHostPort is lost:
```
-	// Empty ports.		
 -	if port == "0" || port == "" {		
 -		// Port zero or empty means use requested to choose any freely available		
 -		// port. Avoid this since it won't work with any configured clients,		
 -		// can lead to serious loss of availability.		
 -		return "", "", errEmptyPort		
 -	}
```

cmd/net.go
```diff
@@ -0,0 +1,106 @@
...
+	return host, port
+}
+
+// mustGetLocalIP4 returns IPv4 addresses of local host.  It panics on error.
+func mustGetLocalIP4() (ipList set.StringSet) {
+	ipList = set.NewStringSet()
+	addrs, err := net.InterfaceAddrs()
+	if err != nil {
+		fmt.Println("Error: Unable to get IP addresses of this host.")
+		panic(err)
```
krishnasrinivas: use of fatalIf is better

cmd/endpoint.go
```diff
@@ -0,0 +1,189 @@
...
+func NewEndpoint(arg string) (Endpoint, error) {
+	if IsEmptyPath(arg) {
+		return Endpoint{}, fmt.Errorf("Empty or root endpoint is not supported")
+	}
+
+	u, err := url.Parse(arg)
+	if err == nil && u.Host != "" {
+		// URL style of endpoint.
+		if !((u.Scheme == "http" || u.Scheme == "https") &&
+			u.Opaque == "" && u.ForceQuery == false && u.RawQuery == "" && u.Fragment == "") {
```
krishnasrinivas, line 121: can you explain/add comments on what the code is trying to do here?

krishnasrinivas: we can disallow any scheme that's not http/https

cmd/net.go
```diff
@@ -0,0 +1,106 @@
...
+	host, port, err = net.SplitHostPort(hostPort)
+	if err != nil {
+		// Ignore "missing port in address" error as we use default port.
+		// This happens when hostPort = "" or hostPort = "server".
+		if !strings.Contains(err.Error(), "missing port in address") {
+			return "", "", err
+		}
+
+		host = hostPort
+		port = defaultPort
```
krishnasrinivas, line 42: we should have set globalMinioPort by now and use it instead of defaultPort

cmd/endpoint.go
krishnasrinivas: url.URL is fine to represent both path and host:port/path .... so it's better to avoid adding new datastructures unless it makes the code cleaner.

balamurugana: > url.URL is fine to represent both path and host:port/path .... so it's better to avoid adding new datastructures unless it makes the code cleaner.

having `url.URL` for other than URL style endpoint is not a fit.  Look into the below path style endpoints

```
foo
/foo
\foo
C:\foo
C:/foo
C:\\foo
http:path
http:/path
http:///path
```

In these cases, sometimes we would need to use URL.Scheme, URL.Opaque and/or URL.Path.  This would make logical errors.

balamurugana: > looking at other variables:
> IsDisabled - is not used anywhere

Will remove that.


> IsLocal - is not needed because we already have isLocalStorage() which is used only during init process. If isLocalStorage() was used during operations (i.e not just during init) then it would make sense to store it in a struct variable like this so that we don't unnecessarily call isLocalStorage() after init process
> 

In validation process, we know what are all endpoints are local.  There is no need to use another function to do the same, hence `IsLocal`


> so having Endpoint struct like this does not look like it is adding any advantage. url.URL is enough to represent an endpoint.

When `format.json` verification is moved into object layer, the content of `format.json` will be stored into `Endpoint` for every reconnect validation.

cmd/endpoint.go
balamurugana: Below all conditions have to be met for a valid URL style endpoint.
1. Scheme has to be `http` or `https`
2.  Opaque, ForceQuery, RawQuery and Fragment have to be empty.

If not, its not a valid URL style endpoint.  The check is negation of valid case.

cmd/net.go
balamurugana: This is not default server address `:9000`. It is default port a endpoint where port is not specified.

You could check `TestSplitHostPort()` for valid/invalid cases.

cmd/net.go
balamurugana: Not required.  You would need to check `TestCheckLocalServerAddr()`

cmd/net.go
balamurugana: `globalMinioPort` is used as port value for current process, not as a constant.  Here it says about defaultPort `9000`

cmd/net.go
balamurugana: Done

cmd/net.go
balamurugana: Done

cmd/setup.go
balamurugana: > But what is the goal of doing these changes? is it to bring all the setup related global variables into a single global struct so as not to pollute the global name space?

This change will make globals removed and setup definition is encapsulated into one object and that will bring a context ability which will help unit tests have it own context to run (not in parallel).

> note that:
> 
>     serverAddr string
>     endPoints []Endpoint
> 
> does not seem to be useful once init is completed.

No. On every login to servers need to validate endpoints and associated `format.json` of remote servers.  endpoints are needed to be sent/received.  This is not done in the current code base.

cmd/endpoint.go
krishnasrinivas: > In these cases, sometimes we would need to use URL.Scheme, URL.Opaque and/or URL.Path. This would make logical errors.

getPath() abstracts such details. But I agree that in case of "c:\path" storing "c" in u.Scheme does not look intuitive and also ep.Path is more readable than getPath(ep)

> In validation process, we know what are all endpoints are local. There is no need to use another function to do the same, hence IsLocal

isLocalStorage() will still be needed even if you have IsLocal, If we decide to have a separate Endpoint struct then we can have IsLocal too instead of calling isLocalStorage multiple times even during init process.

So a better Endpoint struct looks to be:
```
type Endpoint struct {
    path string
    url *url.URL
    isLocal bool
}
```
Either path will be valid or url will be valid depending on whether endpoint represents a path or a url.  This will be more readable than having "Value" field which will be valid for both path and url ... plus Value is not useful as only url field gets used if the endpoint represents a URL.

cmd/endpoint.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+
+	return Endpoint{
+		Value: arg,
+		URL:   u,
+	}, nil
+}
+
+// EndpointList - list of same type of endpoint.
+type EndpointList []Endpoint
```
krishnasrinivas, line 147: better do away this new type and use []Endpoint where ever we use slice of endpoints

cmd/net.go
krishnasrinivas: no, because this will not work:

minio server --address :8000 http://host1/path1 http://host2/path2....

i.e we should consider 8000 and not 9000

cmd/setup.go
krishnasrinivas: I see, so will you be making the changes suggested regarding using struct instead of interface?

krishnasrinivas:
@balamurugana also note that all the new code (i.e NewSetup()) is not being made use of. Hence I would suggest that after taking care of the code comments, you remove the old code that the new code intends to replace (for ex. parseStorageEndpoints(), checkEndpointsSyntax(), checkDuplicateEndpoints()) and make sure that new code is made use of everywhere (i.e there will be no use of functions like getPath()) then test (basic testing like copying one object) minio with different kinds of setups in both linux and windows:

* simple setup: minio server /export and minio server C:\export
* on windows when your PWD is on C: but you do minio server D:\export
* use --address to change the default port in XL/DistXL (this will fail this your current patch because of the explanation in the comment) on both windows and linux
* use --address to run multiple servers on the same machine simulating a DistXL (should be tested on both linux/windows)

once the code comments has been addressed and you confirm that the above setups has been tested the patch can be taken in.

cmd/endpoint.go
balamurugana: The idea behind this is to have methods for `format.json` get/get/verify in future where endpoint order is use.

balamurugana:
> you remove the old code that the new code intends to replace (for ex. parseStorageEndpoints(), checkEndpointsSyntax(), checkDuplicateEndpoints()) and make sure that new code is made use of everywhere (i.e there will be no use of functions like getPath())

If we scope for everything in this PR, it would need unit/functional test startups to be in big change.   However I will be keep updating this PR till it gets merged

harshavardhana:
@balamurugana - can you rebase this PR?  

Since this is a major change touches code written by @vadmeste @krishnasrinivas @krisis - can you guys review this and :+1:

harshavardhana:
@balamurugana do we have a time when this PR will be ready to be reviewed again?  are still planning to open new bugs per PR which you are going to send?

harshavardhana:
If this PR is not going to be rebased or submitted to anymore - should we go ahead and close it @balamurugana ?

harshavardhana:
> If this PR is not going to be rebased or submitted to anymore - should we go ahead and close it @balamurugana ?

ping @balamurugana

harshavardhana:
@balamurugana please close this PR since #3975

harshavardhana:
> @balamurugana please close this PR since #3975

My bad looks like #3975 is not the same.

deekoder:
I will mark this PR blocked.

deekoder:
disussed with @balamurugana  closing it as per conversation.

PR #3646, opened 2017-01-27
xl: Fix GET of an empty multiparted object

### LINKED ISSUE #3644: xl: Cannot download an empty object uploaded by multipart
## Expected Behavior
To able to download an empty object uploaded using multipart

## Current Behavior
Not able to download 
```
ERRO[0060] Unable to write to client.                    cause=The requested range "bytes 0-0/0" is not satisfiable. source=[object-handlers.go:162:objectAPIHandlers.GetObjectHandler()] stack=xl-v1-object.go:204:xlObjects.GetObject <autogenerated>:856:(*xlObjects).GetObject object-handlers.go:161:objectAPIHandlers.GetObjectHandler api-router.go:54:(objectAPIHandlers).GetObjectHandler-fm
```

## Possible Solution
Fix calculating part index when offset and length are equal to zero

1. `aws s3api --endpoint-url "http://localhost:9000" create-multipart-upload --bucket testbucket --key 'multipart/01'`
2. `aws s3api --endpoint-url "http://localhost:9000" upload-part --bucket testbucket --key 'multipart/01' --part-number 1 --body part01 --upload-id  "59d05370-8c7c-40f9-ab5b-0b7ef8353498"`
3. `aws s3api --endpoint-url "http://localhost:9000" complete-multipart-upload --multipart-upload file://complete.file --bucket testbucket --key 'multipart/01' --upload-id 59d05370-8c7c-40f9-ab5b-0b7ef8353498`
4. `aws s3api --endpoint-url "http://localhost:9000" get-object --bucket testbucket --key 'multipart/01' /tmp/output`

## Context
Manual testing

## Your Environment
Commit 85f2b74cfda5eb4ebc660aff273d1f7e41ee86bd

vadmeste:
## Description
GetObject returns unsatisfied range error when we try to download an object
uploaded using multipart mechanism.

## Motivation and Context
Fixes #3644 

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:

### DIFF
--- a/cmd/xl-v1-object.go
+++ b/cmd/xl-v1-object.go
@@ -198,8 +198,14 @@ func (xl xlObjects) GetObject(bucket, object string, startOffset int64, length i
 		return traceError(InvalidRange{startOffset, length, xlMeta.Stat.Size})
 	}
 
+	// Calculate endOffset according to length
+	endOffset := startOffset
+	if length > 0 {
+		endOffset += length - 1
+	}
+
 	// Get last part index to read given length.
-	lastPartIndex, _, err := xlMeta.ObjectToPartOffset(startOffset + length - 1)
+	lastPartIndex, _, err := xlMeta.ObjectToPartOffset(endOffset)
 	if err != nil {
 		return traceError(InvalidRange{startOffset, length, xlMeta.Stat.Size})
 	}


cmd/xl-v1-object.go
```diff
@@ -198,8 +198,14 @@ func (xl xlObjects) GetObject(bucket, object string, startOffset int64, length i
 		return traceError(InvalidRange{startOffset, length, xlMeta.Stat.Size})
 	}
 
+	// Calculate endOffset according to length
```
harshavardhana, line 201: We can return if 'offset == 0 && length == 0' as success at object-handlers.go as an additional optimization so that the i/o doesn't even reach the backend as its not needed anyways.

harshavardhana: This part is still needed..

harshavardhana (approved):

cmd/xl-v1-object.go
krishnasrinivas: note that i/o does not hit because we wont enter the loop:
`for ; partIndex <= lastPartIndex; partIndex++ {`

also if we return in the object layer itself then we will miss out on checking the quorum error.

cmd/xl-v1-object.go
```diff
@@ -198,8 +198,14 @@ func (xl xlObjects) GetObject(bucket, object string, startOffset int64, length i
...
+	endOffset := startOffset
```
krishnasrinivas, line 202: we can maybe call this lastByteOffset so that `-1` that we do is self explanatory, or add a comment on why `-1` is done.

cmd/xl-v1-object.go
krishnasrinivas: sorry, the loop is entered but we break because of `if length == totalBytesRead {` check.

harshavardhana: It is a trade off to avoid hitting the disk unnecessarily when client didn't want data in the first place. What needs to be sent to client is already available with GetObjectInfo. 

That is what it means by 'offset=0' and 'length=0' , value of validating quorum exists only if we need to serve data - when we are not going to serve anything it doesn't make sense to validate quorum. 0bytes file itself is quite meaningless on object storage - we kept it for compatibility reasons alone. 

I don't see any negative impacts of doing this.

PR #3648, opened 2017-01-27
docs: Restructure docs, move the files to their relevant location.

harshavardhana:
## Description
Moving docs to their relevant location. Also combines windows TLS docs with single doc with Linux.

## Motivation and Context
Cleanup

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:


camarcut (changes requested):
redacted. submitted incorrectly. See below.

deekoder (dismissed):
Linux/Readme :
5: - Systemd script is ... -> This seems very abrupt way to start this doc. Is it a Pre-Requisite?
6: - Download and find the relevant links.. What should user look for. Are we making any assumptions on what the user knows when reading this doc. If so we should state them.

12: Change "Give execute permissions". Sounds non standard. We could say " Change or Add permission".

20: Remove "Don't forget to" : Use neutral language. We do not use a conversational style in our docs anywhere else.

25: Use if you want to : Unusual start of sentence. Say something like "Optional : "

33 By default, minio reads... -> missing comma

45: configuration to override default keys. Is that a sentence fragment?

46: Systemctl is an inadequate heading. More description would be useful.

tls/README.md

Remove your from heading.

15: Let'sEncrypt is missing apostrophe.
47: Add the directory to your PATH and restart : could you add shell to show how it is done literally. Also we do this to cut copy paste and modify easily.
58: The easiest way is to generate certificate is to specify -> Remove is.

100: Now, it is time to generate the public certificate using this command. Remove Now, it is time to (embellishments). Just state "Generate public certificate using".

harshavardhana:
Fixed all the comments @deekoder

deekoder:
@harshavardhana  LGTM after one more change : 
"To make Minio aware about your generated key and certificate..."

- "aware about" doesn't sound right in this context. I think it should be "aware of".

harshavardhana:
> "aware about" doesn't sound right in this context. I think it should be "aware of".

Fixed that as well @deekoder - let me know what else.

deekoder:
👍Approved.

PR #3651, opened 2017-01-28
PutObject: use parts mechanism when storing objects

### LINKED ISSUE #3559: api: GetObject() transfer rate observed to have TTFB (Time To First Byte) ~5s delay
Downloading a file is slow

## Expected Behavior
When I download a file the TTFB should be few milliseconds

## Current Behavior
Downloading a 500Mo file take about 5s before the first bit
```
time curl -o /dev/null -s -w "Connect: %{time_connect} TTFB: %{time_starttransfer} Total time: %{time_total} \n" XXXXXXXX
Connect: 0.042 TTFB: 3.966 Total time: 31.460
```

## Your Environment
Cluster of 4 machines running on docker.

```
docker exec -ti 886727bc922e minio version
Version: 2017-01-09T23:49:09Z
Release-Tag: DEVELOPMENT.2017-01-09T23-49-09Z
Commit-ID: eb6d53d2f5d66e16a621141df111bd78bb843ecb
```
Bandwidth between the servers 1Gbit, the machines are on the same datacenter, no connectivity problem, when the download start we can observe up to 800Mbits/s between the machines, the BW is not the issue.
The problem can also be observed remotely and locally, 5s can't be explained by the latency between the client and the server anyway.
The rest of the transfert seems also quite slow compared to the network performances but this need more investigation on my side to confirm this part.

vadmeste:
## Description
For faster time-to-first-byte when we try to download a big object

## Motivation and Context
Fixes https://github.com/minio/minio/issues/3559

## How Has This Been Tested?
go test

## Types of changes

- [] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/cmd/xl-v1-multipart.go
+++ b/cmd/xl-v1-multipart.go
@@ -627,8 +627,11 @@ func (xl xlObjects) PutObjectPart(bucket, object, uploadID string, partID int, s
 		}
 	}
 
+	// We always allow empty part.
+	allowEmpty := true
+
 	// Erasure code data and write across all disks.
-	sizeWritten, checkSums, err := erasureCreateFile(onlineDisks, minioMetaTmpBucket, tmpPartPath, teeReader, xlMeta.Erasure.BlockSize, xl.dataBlocks, xl.parityBlocks, bitRotAlgo, xl.writeQuorum)
+	sizeWritten, checkSums, err := erasureCreateFile(onlineDisks, minioMetaTmpBucket, tmpPartPath, teeReader, allowEmpty, xlMeta.Erasure.BlockSize, xl.dataBlocks, xl.parityBlocks, bitRotAlgo, xl.writeQuorum)
 	if err != nil {
 		return "", toObjectErr(err, bucket, object)
 	}
--- a/cmd/xl-v1-object.go
+++ b/cmd/xl-v1-object.go
@@ -22,6 +22,7 @@ import (
 	"hash"
 	"io"
 	"path"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
@@ -461,7 +462,6 @@ func (xl xlObjects) PutObject(bucket string, object string, size int64, data io.
 	}
 
 	uniqueID := mustGetUUID()
-	tempErasureObj := path.Join(uniqueID, "part.1")
 	tempObj := uniqueID
 
 	// Initialize md5 writer.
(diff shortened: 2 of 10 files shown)


harshavardhana (commented):
Shadowing issue in the build - can you please fix @vadmeste

vadmeste:
> Shadowing issue in the build - can you please fix @vadmeste

Sure @harshavardhana, I still have one failed unit test before we go for review

vadmeste:
@harshavardhana, your comments are addressed, let's make another review round.

harshavardhana (approved):
This is neat! Can you remove the WIP? - Will wait for @krishnasrinivas review for any structural issues in this patch.

cmd/xl-v1-object.go
```diff
@@ -461,13 +462,16 @@ func (xl xlObjects) PutObject(bucket string, object string, size int64, data io.
...
-	tempErasureObj := path.Join(uniqueID, "part.1")
 	tempObj := uniqueID
 
 	// Initialize md5 writer.
 	md5Writer := md5.New()
 
-	writers := []io.Writer{md5Writer}
+	// Initialize md5 writer for parts, it will be
+	// reset before reading each new part
+	partMD5Writer := md5.New()
```
krishnasrinivas: i think leaving out part's md5sum will be better as it does not serve any purpose and it would just consume extra cpu.

cmd/xl-v1-object.go
```diff
@@ -512,40 +516,96 @@ func (xl xlObjects) PutObject(bucket string, object string, size int64, data io.
 	// Tee reader combines incoming data stream and md5, data read from input stream is written to md5.
 	teeReader := io.TeeReader(limitDataReader, mw)
 
+	// Initialize parts metadata
+	partsMetadata := make([]xlMetaV1, len(xl.storageDisks))
+
 	// Initialize xl meta.
-	xlMeta := newXLMetaV1(object, xl.dataBlocks, xl.parityBlocks)
+	for index := range partsMetadata {
+		partsMetadata[index] = newXLMetaV1(object, xl.dataBlocks, xl.parityBlocks)
```
krishnasrinivas: here each disk will get a different Distribution, make sure that xlMeta.Erasure.Distribution is same for all the disks

cmd/xl-v1-object.go
```diff
@@ -512,40 +516,96 @@ func (xl xlObjects) PutObject(bucket string, object string, size int64, data io.
...
+
+		// Erasure code data and write across all disks.
+		partSizeWritten, checkSums, erasureErr := erasureCreateFile(onlineDisks, minioMetaTmpBucket, tempErasureObj, partReader, partsMetadata[0].Erasure.BlockSize, partsMetadata[0].Erasure.DataBlocks, partsMetadata[0].Erasure.ParityBlocks, bitRotAlgo, xl.writeQuorum)
+		if erasureErr != nil {
+			return ObjectInfo{}, toObjectErr(erasureErr, minioMetaTmpBucket, tempErasureObj)
+		}
+
+		// Should return IncompleteBody{} error when reader has fewer bytes
+		// than specified in request header.
+		if partSizeWritten < int64(curPartSize) {
```
krishnasrinivas, line 573: if curPartSize is -1 or 0 this will always return error

cmd/xl-v1-object.go
```diff
@@ -512,40 +516,96 @@ func (xl xlObjects) PutObject(bucket string, object string, size int64, data io.
...
+			partsMetadata[index].Erasure.AddCheckSumInfo(checkSumInfo{
+				Name:      partName,
+				Hash:      checkSums[index],
+				Algorithm: bitRotAlgo,
+			})
+		}
+
+		// If we didn't write anything or we know that the next part doesn't have any
+		// data to write, we should quit this loop immediately
+		if partSizeWritten == 0 || getPartSizeFromIdx(size, globalPutPartSize, partIdx+1) == 0 {
```
krishnasrinivas, line 596: if size is -1, then we end up creating one extra part which is empty which can be avoided by checking if partSizeWritten is < 64MB. But if the stream size is a multiple of 64MB then it's a little tricky, so we can just delete the extra empty part before comitting.

krishnasrinivas (changes requested):

cmd/xl-v1-object.go
harshavardhana: On the other hand we need to compute the entire md5sum.

cmd/xl-v1-object.go
vadmeste: > if curPartSize is -1 or 0 this will always return error

Actually it is quite the opposite, if curPartSize is -1 or 0 it will never return an error.

cmd/xl-v1-object.go
vadmeste: > i think leaving out part's md5sum will be better as it does not serve any purpose and it would just consume extra cpu.

:+1:

cmd/xl-v1-object.go
vadmeste: > here each disk will get a different Distribution

It doesn't, Distribution is generated using hashOrder() which returns the same information if we provide the same input which is (object, xl.dataBlocks, xl.parityBlocks) in our case

cmd/xl-v1-object.go
vadmeste: > But if the stream size is a multiple of 64MB then it's a little tricky, so we can just delete the extra empty part before comitting.

You are right. Other than removing empty parts after creating them, we probably can pass a parameter to erasureCreateFile() to tell if it should create empty files or not, what do you think ?

cmd/xl-v1-object.go
krishnasrinivas: Ah yes, you are right

cmd/xl-v1-object.go
krishnasrinivas: passing a parameter to erasureCreateFile will be simpler :+1:

vadmeste:
All comments addressed, can you review again ?

cmd/xl-v1-multipart.go
```diff
@@ -627,8 +627,10 @@ func (xl xlObjects) PutObjectPart(bucket, object, uploadID string, partID int, s
 		}
 	}
 
+	allowEmptyErasure := true
```
harshavardhana: allowEmpty is fine.. 'erasureCreateFile' is already the function indicating 'erasure' we don't need to redundantly say it again.

krishnasrinivas (approved):

cmd/xl-v1-object.go
harshavardhana: But it is better to allocate this once and use it when we know this metadata is going to be consistent. So to indicate that distribution is only calculated once per upload. Relatively it would be faster and doesn't employ doing crc32 again and again on the same object name.

cmd/xl-v1-object.go
```diff
@@ -643,12 +690,12 @@ func (xl xlObjects) PutObject(bucket string, object string, size int64, data io.
 		IsDir:           false,
 		Bucket:          bucket,
 		Name:            object,
-		Size:            xlMeta.Stat.Size,
-		ModTime:         xlMeta.Stat.ModTime,
-		MD5Sum:          xlMeta.Meta["md5Sum"],
-		ContentType:     xlMeta.Meta["content-type"],
-		ContentEncoding: xlMeta.Meta["content-encoding"],
-		UserDefined:     xlMeta.Meta,
+		Size:            partsMetadata[0].Stat.Size,
```
harshavardhana: Indicate why we are using first element in the 'partsMetadata' slice.  Rather than changing this just set 

```
xlMeta := partsMetadata[0]
```

cmd/xl-v1-object.go
```diff
@@ -512,40 +512,98 @@ func (xl xlObjects) PutObject(bucket string, object string, size int64, data io.
...
-	// Should return IncompleteBody{} error when reader has fewer bytes
-	// than specified in request header.
-	if sizeWritten < size {
-		return ObjectInfo{}, traceError(IncompleteBody{})
+		// partReader streams at most maximum part size
+		partReader := io.LimitReader(teeReader, globalPutPartSize)
+
+		// Allow creating empty earsure file only when this is the first part. This flag is useful
+		// when size == -1 because in this case, we are not able to predict how many parts we will have.
+		allowEmptyErasure := partIdx == 1
```
harshavardhana: `allowEmptyObject`

cmd/xl-v1-object.go
```diff
@@ -512,40 +512,98 @@ func (xl xlObjects) PutObject(bucket string, object string, size int64, data io.
...
+		// than specified in request header.
+		if partSizeWritten < int64(curPartSize) {
+			return ObjectInfo{}, traceError(IncompleteBody{})
+		}
+
+		// Update the total written size
+		sizeWritten += partSizeWritten
+
+		// If erasure stored some data in the loop or created an empty file
+		if partSizeWritten > 0 || allowEmptyErasure {
```
harshavardhana: `allowEmptyErasure` can be `allowEmptyObject`

cmd/xl-v1-object.go
harshavardhana: also return this value with `getPartSizeFromIdx` this function.

cmd/xl-v1-utils.go
```diff
@@ -343,3 +343,24 @@ func getOrderedDisks(distribution []int, disks []StorageAPI) (orderedDisks []Sto
 	}
 	return orderedDisks
 }
+
+// getPartSizeFromIdx predicts the part size according to its index. It also
+// returns -1 when totalSize is also -1.
+func getPartSizeFromIdx(totalSize int64, partSize int64, partIndex int) int64 {
```
harshavardhana, line 349: We need to add tests for this function since its a critical piece -

harshavardhana (changes requested):

cmd/xl-v1-object.go
vadmeste: > allowEmptyObject

probably allowEmptyPart is better?

harshavardhana: Sure

vadmeste: > also return this value with getPartSizeFromIdx this function.

You mean like `partSize, allowEmptyPart := getPartSizeFromIdx()` ?

Computing  allowEmptyPart inside getPartSizeFromIdx() will confuse the reader if we can't find the right name of getPartSizeFromIdx() or it doesn't ?

harshavardhana: Yes we need to find a right name for it. Or have another function which returns both.

cmd/xl-v1-utils.go
vadmeste: Done, also I've added some panic() inside getPartSizeFromIdx for unmeaningful parameters like negative part index

cmd/xl-v1-object.go
vadmeste: > Indicate why we are using first element in the 'partsMetadata' slice. Rather than changing this just set
> 
> xlMeta := partsMetadata[0]

Done

harshavardhana (approved):

PR #3656, opened 2017-01-30
browser: Component attr names should not clash with redux state keys.

krishnasrinivas:
fixes #270

## Description

showAbout and showSettings attr values were being over ridden by redux state values.

## Motivation and Context

Fixes https://github.com/minio/miniobrowser/issues/270

## How Has This Been Tested?

manual

## Types of changes

## Checklist:

### DIFF
--- a/browser/app/js/components/Browse.js
+++ b/browser/app/js/components/Browse.js
@@ -414,10 +414,10 @@ export default class Browse extends React.Component {
     let freePercent = free * 100 / total
 
     if (web.LoggedIn()) {
-      browserDropdownButton = <BrowserDropdown fullScreen={ this.fullScreen.bind(this) }
-                                showAbout={ this.showAbout.bind(this) }
-                                showSettings={ this.showSettings.bind(this) }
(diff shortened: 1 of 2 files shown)


harshavardhana:
@krishnasrinivas can we fix the commit message to be 80 column only right now it has exceeded github's UI and wrapped to description. 

As practice we can keep the 80 column limit and rest can go to description/body of the commit.

krishnasrinivas:
@harshavardhana right now the commit message is 74 chars:

```
krishna@escape:~$ wc
browser: Component attr names should not clash with redux state key names.      0      12      74
```

So github UI must be having an even lower limit.

harshavardhana:
> So github UI must be having an even lower limit.

Yes we can keep 60 - i usually try to be 50-60.

krishnasrinivas:
github subject limit is 72:
```
krishna@escape:~$ wc
browser: Component attr names should not clash with redux state key n…      0      12      72
```

I have amended the commit message

PR #3664, opened 2017-01-31
Change query param name to duration in list/clear locks API

krisis:
Following is a sample list lock API request schematic,

  /?lock&bucket=mybucket&prefix=myprefix&duration=holdDuration
  x-minio-operation: list

The response would contain the list of locks held on mybucket matching
myprefix for a duration longer than holdDuration.

## Description

This change is to make the query params intention clear to lock management API consumers.
Previously, ``older-than`` query parameter was used to include locks that were held a duration of ``older-than`` units from current time. With this change, we rename the query parameter to ``duration`` to indicate that list/clear locks API will include locks held for a duration longer than the value specified using the ``duration`` query parameter.

## Motivation and Context

This change intends to simplify the lock management APIs.

## How Has This Been Tested?

unit tests

## Types of changes

## Checklist:

### DIFF
--- a/cmd/admin-handlers.go
+++ b/cmd/admin-handlers.go
@@ -36,14 +36,14 @@ type mgmtQueryKey string
 
 // Only valid query params for list/clear locks management APIs.
 const (
-	mgmtBucket    mgmtQueryKey = "bucket"
-	mgmtObject    mgmtQueryKey = "object"
-	mgmtPrefix    mgmtQueryKey = "prefix"
-	mgmtOlderThan mgmtQueryKey = "older-than"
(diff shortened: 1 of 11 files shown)


harshavardhana:
Can you rebase this patch @krisis ?

harshavardhana:
// cc @vadmeste mc needs to be updated as well

vadmeste:
> // cc @vadmeste mc needs to be updated as well

Shall I also modify the flag --older-than to --duration in mc ?

harshavardhana:
> Shall I also modify the flag --older-than to --duration in mc ?

I think so @vadmeste

Issue #3666, opened 2017-01-31
[distributed]: Severe increase in number of goroutines under load.


hackintoshrao:
There is a huge bump in number of go routines in the system observed during  most of the [exec-concurrent](https://github.com/minio/perftest/tree/master/exec-concurrent) tests.  The tests are run at a concurrency of 10.

## Expected Behavior

- Not a severe rise in number of go routines in the system after the load test. 

## Current Behavior

- Severe rise in the number of go routines in the system.
- Here is one such example 

```
time ./exec-concurrent TestMakeBucketRegions
Init:  14
intial :  14 , final :  1219 , diff:  1205
```
You can witness a bump of 1200+ go routines in the system after the tests.

however the number goes down in few minutes.

## Possible Solution

- From the stack trace the issue originates from [here](https://github.com/minio/minio/blob/master/cmd/retry.go#L103), with hundreds of go routine sleeping.
- Since the Go routine has gone to sleep its not receptive to the messages from done channel the back off time is finished.
- It should be made responsive. Have a working patch.
## Steps to Reproduce (for bugs)

1. Run  4 node Minio distributed server.
2. Add debug endpoints for getting number of go routines in the system and use them before and after load test. 
3. Run test at concurrency of 10 `./exec-concurrent TestMakeBucketRegions`

4. Observe the number of go routines in the system before and after the test.

## Context

Test for Go routine leaks.
## Your Environment

* Version used: Latest master at cb48517a78f34fbc3a6cfcf625e813abd9f9ee70
* Operating System and version: Ubuntu 16.04

harshavardhana:
> however the number goes down in few minutes.

How is this a leak after it goes down in few minutes?

hackintoshrao:
Leak may not be the appropriate term, I'm reporting a really large increase in number of go routines in the system even under not so hectic load. With increasing load the number increases to few ten thousands.

harshavardhana:
> Leak may not be the appropriate term, I'm reporting a really large increase in number of go routines in the system even under not so hectic load. With increasing load the number increases to few ten thousands.

Does the server become unavailable? slower? un-responsive? - if the resources are relinquished what is the issue here?

PR #3667, opened 2017-01-31
Fix go routine leaks.

hackintoshrao:
- Fix the Sleep call to make the done channel responsive.

## Description

The issue is described in https://github.com/minio/minio/issues/3666 . 
## Motivation and Context

Fix the Go routine leak by making the done channel responsive all the time.
## How Has This Been Tested?

The test is described in https://github.com/minio/minio/issues/3666 , 

The number of goroutines in the system of after the load test stabilized with this fix.
## Types of changes

## Checklist:

### DIFF
--- a/cmd/retry.go
+++ b/cmd/retry.go
@@ -91,16 +91,26 @@ func newRetryTimer(unit time.Duration, cap time.Duration, jitter float64, doneCh
 	go func() {
 		defer close(attemptCh)
 		nextBackoff := 0
+		// Channel used to signal after the expiry of backoff wait seconds.
+		var timer *time.Timer
 		for {
-			select {
-			// Attempts starts.
+			select { // Attempts starts.
 			case attemptCh <- nextBackoff:
 				nextBackoff++
 			case <-doneCh:
 				// Stop the routine.
 				return
 			}
-			time.Sleep(exponentialBackoffWait(nextBackoff))
+			timer = time.NewTimer(exponentialBackoffWait(nextBackoff))
+			// wait till next backoff time or till doneCh gets a message.
+			select {
+			case <-timer.C:
+			case <-doneCh:
+				// stop the timer and return.
+				timer.Stop()
+				return
+			}
+
 		}
 	}()
 	return attemptCh


cmd/retry.go
```diff
@@ -91,16 +91,28 @@ func newRetryTimer(unit time.Duration, cap time.Duration, jitter float64, doneCh
...
+				// enable the channel to notify after nextbackoff number of seconds.
+				backoffWait = time.After(exponentialBackoffWait(nextBackoff))
 			case <-doneCh:
 				// Stop the routine.
 				return
+			case <-backoffWait:
+				// disable till next backoff time is set in first case.
+				backoffWait = nil
+				// enable attempt channel.
+				attemptCh = make(chan int)
```
vadmeste: Actually, attemptCh channel is returned to the caller of newRetryTimer(), setting it here means we will not able to signal the caller to make its next try

hackintoshrao: Right. Should see how else the `done` channel be made responsive without the routine going to sleep.

hackintoshrao: Will send an alternative fix.

cmd/retry.go
```diff
@@ -91,6 +91,8 @@ func newRetryTimer(unit time.Duration, cap time.Duration, jitter float64, doneCh
 	go func() {
 		defer close(attemptCh)
 		nextBackoff := 0
+		// Channel used to signal after the expiry of backoff wait seconds.
+		var backoffWait <-chan time.Time
```
harshavardhana: Don't use this you don't need this variable at all.  what you need to do is a simple behavior change. 

All you need is this 

```go
$ git diff cmd/retry.go
diff --git a/cmd/retry.go b/cmd/retry.go
index d3ad59e..6a5fe6b 100644
--- a/cmd/retry.go
+++ b/cmd/retry.go
@@ -91,16 +91,19 @@ func newRetryTimer(unit time.Duration, cap time.Duration, jitter float64, doneCh
        go func() {
                defer close(attemptCh)
                nextBackoff := 0
+               var timer *time.Timer
                for {
                        select {
                        // Attempts starts.
                        case attemptCh <- nextBackoff:
                                nextBackoff++
+                               timer = time.NewTimer(exponentialBackoffWait(nextBackoff))
                        case <-doneCh:
                                // Stop the routine.
+                               timer.Stop()
                                return
                        }
-                       time.Sleep(exponentialBackoffWait(nextBackoff))
+                       <-timer.C
                }
        }()
        return attemptCh
```

harshavardhana (changes requested):

cmd/retry.go
hackintoshrao: What if `doneCh` gets a message when the goroutine blocked on `<-timer.C`? Its still not responsive right ?

cmd/retry.go
```diff
@@ -91,16 +91,26 @@ func newRetryTimer(unit time.Duration, cap time.Duration, jitter float64, doneCh
...
 		nextBackoff := 0
+		// Channel used to signal after the expiry of backoff wait seconds.
+		var timer *time.Timer
 		for {
-			select {
-			// Attempts starts.
+			select { // Attempts starts.
 			case attemptCh <- nextBackoff:
 				nextBackoff++
+				timer = time.NewTimer(exponentialBackoffWait(nextBackoff))
```
harshavardhana: You can actually move this out of the select.. so that we can avoid allocating a timer when a doneCh is involved right after timer is allocated.

cmd/retry.go
```diff
@@ -91,16 +91,26 @@ func newRetryTimer(unit time.Duration, cap time.Duration, jitter float64, doneCh
...
 				return
 			}
-			time.Sleep(exponentialBackoffWait(nextBackoff))
+			// wait till next backoff time or till doneCh gets a message.
+			select {
+			case <-timer.C:
+			case <-doneCh:
+				timer.Stop()
+				// stop the go routine and return.
+				return
```
krisis: nit: stop the timer and return.
... and the comment should appear before the code.

krisis (changes requested):
Please add a unit test for this function.

harshavardhana (approved):

PR #3670, opened 2017-02-01
xl: Abort multipart upload should honor quorum properly.

### LINKED ISSUE #3665: [distributed]: Abort Multipart fails when one of the node is permanently is down.
Abort Multipart fails when one of the node is permanently is down.
## Expected Behavior

- Abort Multipart to work when one of the node is down.
## Current Behavior

- Failing when one of the node is permanently down.

```
--- FAIL: TestObjectMultipartAbort (0.48s)
        server_test.go:3807: Expected response status 204, got 500.

```
## Possible Solution

Not sure.

1. Run Minio distributed on 4 nodes.
2. Stop one of the node.
3. Run [distributed functional test](https://github.com/minio/perftest/tree/master/distributed-functional-test)

## Context

Testing with one node down.
## Your Environment

* Version used: Latest master at cb48517a78f34fbc3a6cfcf625e813abd9f9ee70 .
* Operating System and version: Ubuntu 16.04

harshavardhana:
## Description
Current implementation didn't honor quorum properly and didn't
handle the errors generated properly. This patch addresses that
and also moves common code `cleanupMultipartUploads` into xl
specific private function.

## Motivation and Context
Fixes #3665

## How Has This Been Tested?
Manually taking down the 4th server.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/xl-v1-multipart.go
+++ b/cmd/xl-v1-multipart.go
@@ -96,8 +96,8 @@ func (xl xlObjects) updateUploadJSON(bucket, object, uploadID string, initiated 
 	// Wait for all the writes to finish.
 	wg.Wait()
 
-	// Do we have write quorum?
-	if !isDiskQuorum(errs, xl.writeQuorum) {
+	err := reduceWriteQuorumErrs(errs, objectOpIgnoredErrs, xl.writeQuorum)
+	if errorCause(err) == errXLWriteQuorum {
 		// No quorum. Perform cleanup on the minority of disks
 		// on which the operation succeeded.
 
@@ -131,7 +131,7 @@ func (xl xlObjects) updateUploadJSON(bucket, object, uploadID string, initiated 
 			}(index, disk)
 		}
 		wg.Wait()
-		return traceError(errXLWriteQuorum)
+		return err
 	}
 
 	// we do have quorum, so in case of delete upload.json file
@@ -148,11 +148,7 @@ func (xl xlObjects) updateUploadJSON(bucket, object, uploadID string, initiated 
 		}(index, disk)
 	}
 	wg.Wait()
-
-	if reducedErr := reduceWriteQuorumErrs(errs, objectOpIgnoredErrs, xl.writeQuorum); reducedErr != nil {
-		return reducedErr
-	}
-	return nil
+	return err
 }
 
 // addUploadID - add upload ID and its initiated time to 'uploads.json'.
@@ -266,17 +262,12 @@ func commitXLMetadata(disks []StorageAPI, srcBucket, srcPrefix, dstBucket, dstPr
 	// Wait for all the routines.
 	wg.Wait()
 
-	// Do we have write Quorum?.
-	if !isDiskQuorum(mErrs, quorum) {
+	err := reduceWriteQuorumErrs(mErrs, objectOpIgnoredErrs, quorum)
+	if errorCause(err) == errXLWriteQuorum {
 		// Delete all `xl.json` successfully renamed.
 		deleteAllXLMetadata(disks, dstBucket, dstPrefix, mErrs)
-		return traceError(errXLWriteQuorum)
 	}
-
-	if reducedErr := reduceWriteQuorumErrs(mErrs, objectOpIgnoredErrs, quorum); reducedErr != nil {
-		return reducedErr
-	}
-	return nil
+	return err
 }
 
 // listMultipartUploads - lists all multipart uploads.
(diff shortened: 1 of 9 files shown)


cmd/xl-v1-multipart.go
```diff
@@ -1077,13 +1077,51 @@ func (xl xlObjects) CompleteMultipartUpload(bucket string, object string, upload
...
+}
+
 // abortMultipartUpload - wrapper for purging an ongoing multipart
 // transaction, deletes uploadID entry from `uploads.json` and purges
 // the directory at '.minio.sys/multipart/bucket/object/uploadID' holding
 // all the upload parts.
 func (xl xlObjects) abortMultipartUpload(bucket, object, uploadID string) (err error) {
 	// Cleanup all uploaded parts.
-	if err = cleanupUploadedParts(bucket, object, uploadID, xl.storageDisks...); err != nil {
+	if err = xl.cleanupUploadedParts(bucket, object, uploadID, xl.storageDisks); err != nil {
```
krisis: ``cleanUploadedParts`` is a method of ``xlObjects`` and has access to ``storageDisks``. We shouldn't  be passing it as a parameter.

cmd/xl-v1-multipart.go
```diff
@@ -1077,13 +1077,51 @@ func (xl xlObjects) CompleteMultipartUpload(bucket string, object string, upload
...
+		if disk == nil {
+			errs[index] = traceError(errDiskNotFound)
+			continue
+		}
+		wg.Add(1)
+		// Cleanup each uploadID in a routine.
+		go func(index int, disk StorageAPI) {
+			defer wg.Done()
+			err := cleanupDir(disk, minioMetaMultipartBucket, uploadIDPath)
+			if err != nil && errorCause(err) != errVolumeNotFound {
```
krisis: Why is ``errVolumeNotFound`` treated differently?

cmd/xl-v1-multipart.go
```diff
@@ -1077,13 +1077,51 @@ func (xl xlObjects) CompleteMultipartUpload(bucket string, object string, upload
...
+				errs[index] = err
+			}
+		}(index, disk)
+	}
+
+	// Wait for all the cleanups to finish.
+	wg.Wait()
+
+	// Do we have write quorum?
+	if !isDiskQuorum(errs, xl.writeQuorum) {
```
krisis: ``isDiskQuorum`` looks suspect. It counts "non-ignored" errors and checks if the total is greater than the supplied quorum value. This is the opposite of what we want. We wish to know if there were quorum or more success (i.e, ``err == nil``) to deem the operation a success. This is not related to this patch, but this change relies on it.

krisis (changes requested):

cmd/xl-v1-multipart.go
harshavardhana: oh yes.. now we don't need.

cmd/xl-v1-multipart.go
harshavardhana: removed..

cmd/xl-v1-multipart.go
harshavardhana: It is because `minioMetaMultipartBucket` might not existing but we don't need to throw an error for that and safely ignore it..  But it was an early requirement since sometime we don't delete `minioMetaMultipartBucket` ever. So this case won't happen. 

This can be safely removed.

cmd/xl-v1-multipart.go
```diff
@@ -902,9 +895,9 @@ func (xl xlObjects) CompleteMultipartUpload(bucket string, object string, upload
 
 	// Read metadata associated with the object from all disks.
 	partsMetadata, errs := readAllXLMetadata(xl.storageDisks, minioMetaMultipartBucket, uploadIDPath)
-	// Do we have writeQuorum?.
-	if !isDiskQuorum(errs, xl.writeQuorum) {
-		return ObjectInfo{}, toObjectErr(traceError(errXLWriteQuorum), bucket, object)
+	reducedErr := reduceWriteQuorumErrs(errs, objectOpIgnoredErrs, xl.writeQuorum)
+	if errorCause(reducedErr) == errXLWriteQuorum {
+		return ObjectInfo{}, toObjectErr(traceError(reducedErr), bucket, object)
```
krisis: ``traceError`` is redundant for ``reducedErr``.

krisis (commented):
Could you add a test case for toObjectError(...) to ensure that the following property is maintained,
```go
toObjectErr(nil) == nil
```

cmd/xl-v1-multipart.go
harshavardhana: done

harshavardhana:
> toObjectErr(nil) == nil

Done add for toAPIErrCode and toStorageErr as well.

PR #3671, opened 2017-02-01
Require content-length in POST & Upload requests

vadmeste:
## Description
Avoid passing size = -1 to PutObject API by requiring content-length
header in POST request (as AWS S3 does) and in Upload web handler.
Post handler is modified to completely store multipart file to know
its size before sending it to PutObject().

## Motivation and Context
Avoid passing size = -1 to PutObject()

## How Has This Been Tested?
go test & manually

## Types of changes

## Checklist:

### DIFF
--- a/cmd/bucket-handlers.go
+++ b/cmd/bucket-handlers.go
@@ -393,6 +393,13 @@ func (api objectAPIHandlers) PostPolicyBucketHandler(w http.ResponseWriter, r *h
 		return
 	}
 
+	// Require Content-Length to be set in the request
+	size := r.ContentLength
+	if size < 0 {
+		writeErrorResponse(w, ErrMissingContentLength, r.URL)
+		return
+	}
+
 	// Here the parameter is the size of the form data that should
 	// be loaded in memory, the remaining being put in temporary files.
 	reader, err := r.MultipartReader()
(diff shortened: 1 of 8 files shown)


cmd/bucket-handlers.go
```diff
@@ -402,12 +409,28 @@ func (api objectAPIHandlers) PostPolicyBucketHandler(w http.ResponseWriter, r *h
 		return
 	}
 
-	fileBody, fileName, formValues, err := extractPostPolicyFormValues(reader)
+	// Read multipart data and save in memory and in the disk if needed
+	form, err := reader.ReadForm(1024 * 1024 * 10)
```
harshavardhana: Move this as a constant and btw why 10MB only?

vadmeste: > Move this as a constant and btw why 10MB only?

This is an arbitrary value, it means how much memory we allocate per POST multipart request, data that exceeds will be saved in the disk

harshavardhana: Let's choose a smaller value then perhaps like 5MB

vadmeste: > Let's choose a smaller value then perhaps like 5MB

Done

harshavardhana:
Can you bring in some test cases for this @vadmeste ?

harshavardhana (approved):

PR #3682, opened 2017-02-04
all/windows: Be case in-sensitive about pattern matching.

harshavardhana:
## Description
Resource strings and paths are case insensitive on windows
deployments but if user happens to use upper case instead of
lower case for certain configuration params like bucket
policies and bucket notification config. We might not honor
them which leads to a wrong behavior on windows.

## Motivation and Context
This is windows only behavior, for all other platforms case
is still kept sensitive.

## How Has This Been Tested?
Manually on Windows.

## Types of changes

## Checklist:


cmd/posix-list-dir-nix.go
```diff
@@ -68,10 +68,6 @@ func parseDirents(dirPath string, buf []byte) (entries []string, err error) {
 		if name == "." || name == ".." {
 			continue
 		}
-		// Skip special files.
-		if hasPosixReservedPrefix(name) {
-			continue
-		}
```
krisis, line 74: Why is this being removed?

harshavardhana: It should have been removed long time ago. This was pre-XL days.

krisis (approved):

Issue #3679, opened 2017-02-03
admin: Introduce a new API ServerInfo()


harshavardhana:
Provides following output, new command.

```
$ mc admin info play --json | jq .
{
  "status": "success",
  "storage": {
    "total": 675875831808,
    "free": 602565951488,
    "backend": {
      "type": "XL",
      "onlineDisks": 8,
      "offlineDisks": 0
    }
  },
  "server": {
    "uptime": 0000,  
    "version": "2017-02-02T20:07:53Z",
    "commitID": "aacfb8654ee399922ca70a6f2d86e23e84ce39e5",
    "sqsARN": [ "arn:minio:sqs:us-east-2:1:webhook", ...],
    "region": "us-east-2"
  },
  "network": {
    "transferred": 0000,
    "received": 00000,
    "thoughput": 0000,
  }
}
```

```
$ mc admin info play
Region:    us-east-2
SQS ARNs:  arn:minio:sqs:us-east-2:1:webhook
Total: 630GiB, Free: 561GiB. Online Disks: 8, Offline Disks: 0
Received: NNGiB, Transferred: NNGiB since <human-duration>
```

## Expected Behavior
Expected to have `ServerInfo`

## Current Behavior
Currently serverInfo is part of status Info but we need to introduce network transfer rates as well.

## Possible Solution
Implement network monitors at handler layer. 

## Context
Enhancement to existing functionality. 

## Your Environment
Current `mc master` and `minio master`

gowithplanb:
Please include the number of requests handled as well, that will make for nice graphing too.

harshavardhana:
> Please include the number of requests handled as well, that will make for nice graphing too.

Thanks @gowithplanb looks like a good idea :+1:

krisis:
@harshavardhana I can take this up.

krisis:
@harshavardhana how do we represent ``server`` key in the case of a distributed setup? e.g, ``version`` and ``uptime`` could be different for nodes in a distributed setup. One way to handle this is make ``server`` key an json array with one element per node. The same question applies to ``network`` key as well. Thoughts?

vadmeste:
by the way @krisis, I forgot to mention here that I already started to write a handler to count transferred bytes and requests count etc.. I mean all network related information. I'll commit it today and you can take the rest of the task if you want.

vadmeste:
Storage Info (free/total storage) is removed in a previous commit, can you confirm that we don't need it anymore or it will be just moved somewhere?

[...]

PR #3686, opened 2017-02-05
stats: Add network and http statisics

vadmeste:
## Description
Network: total bytes of incoming and outgoing server's data
by taking advantage of our ConnMux Read/Write wrapping

HTTP: total number of different http verbs passed in http
requests and different status codes passed in http responses.
This is counted in a new http handler.

## Motivation and Context
Partially f.i.x.e.s #3679 

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:

### DIFF
--- a/cmd/stats.go
+++ b/cmd/stats.go
@@ -1,0 +1,129 @@
+/*
+ * Minio Cloud Storage, (C) 2017 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cmd
+
+import (
+	"net/http"
+	"sync/atomic"
+)
+
+// counter - simplify atomic counting
+type counter struct {
+	val uint64
+}
+
+// Inc increases counter atomically
+func (c *counter) Inc(n uint64) {
+	atomic.AddUint64(&c.val, n)
+}
+
+// Value fetches counter's value atomically
+func (c *counter) Value() uint64 {
+	return atomic.LoadUint64(&c.val)
+}
+
+// ConnStats - Network statistics
+// Count total input/output transferred bytes during
+// the server's life.
+type ConnStats struct {
+	totalInputBytes  counter
+	totalOutputBytes counter
+}
+
+// Increase total input bytes
+func (s *ConnStats) incInputBytes(n int) {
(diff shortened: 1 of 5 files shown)


vadmeste:
@harshavardhana, I tested with a sampe code, using atomic is more faster, thanks for pointing it out.

```
package main

import (
        "sync"
        "sync/atomic"
        "testing"
)

type safeTest struct {
        atomicCounter uint64

        mutexCounter  uint64
        mu sync.Mutex
}

// Benchmark increasing counter with mutex lock/unlock
func BenchmarkSafeMutex(b *testing.B) {
        s := &safeTest{}
        for i := 0; i < b.N; i++ {
                s.mu.Lock()
                s.mutexCounter++
                s.mu.Unlock()
        }
}

// Benchmark increasing counter with atomic Add
func BenchmarkSafeAtomic(b *testing.B) {
        s := &safeTest{}
        for i := 0; i < b.N; i++ {
                atomic.AddUint64(&s.atomicCounter, 1)
        }
}
```

Result:
```
BenchmarkSafeMutex-4            50000000                25.1 ns/op
BenchmarkSafeAtomic-4           200000000                9.19 ns/op
PASS
ok      _/tmp/test      4.053s
```

vadmeste:
@harshavardhana, I used atomic wherever it is possible. Not possible to use atomic with map.

cmd/stats.go
```diff
@@ -0,0 +1,113 @@
...
+
+import (
+	"sync"
+	"sync/atomic"
+)
+
+// ConnStats - Network statistics
+// Count total input/output transferred bytes during
+// the server's life.
+type ConnStats struct {
```
harshavardhana, line 42: Lets actually generalize this 

```go
type uint64 counterer

func (c counterer) Inc(uint64 i) {
}

func (c counterer) Value() uint64 {
}
```

vadmeste:
@harshavardhana, the code below shows that map is slower than atomically increasing static variables even with using a switch statement.

So the new approach looks good.

```
package main

import (
        "sync"
        "sync/atomic"
        "testing"
)

func BenchmarkUpdateMap(b *testing.B) {
        method := "PUT"
        mu := sync.Mutex{}
        m := make(map[string]uint64)

        for i := 0; i < b.N; i++ {
                mu.Lock()
                m[method]++
                mu.Unlock()
        }
}

func BenchmarkUpdateStatic(b *testing.B) {
        method := "PUT"
        staticPUT := uint64(0)
        staticOTHER := uint64(0)

        for i := 0; i < b.N; i++ {
                switch method {
                case "PUT":
                        atomic.AddUint64(&staticPUT, 1)
                case "POST":
                        atomic.AddUint64(&staticOTHER, 1)
                case "GET":
                        atomic.AddUint64(&staticOTHER, 1)
                case "HEAD":
                        atomic.AddUint64(&staticOTHER, 1)
                case "DELETE":
                        atomic.AddUint64(&staticOTHER, 1)
                }
        }
}
```
```
testing: warning: no tests to run
BenchmarkUpdateMap-4            20000000                56.5 ns/op
BenchmarkUpdateStatic-4         100000000               16.9 ns/op
PASS
ok      _/tmp/test      2.903s
```

PR #3690, opened 2017-02-06
Add uptime to ServiceStatus

### LINKED ISSUE #3678: admin: ServiceStatus should return back server uptime
```
$ mc admin service status play
Total: 630GiB, Free: 561GiB. Online Disks: 8, Offline Disks: 0.

$ mc admin service status play --json
{"status":"success","service":true,"storageInfo":{"total":675875831808,"free":602565951488,"backend":{"backendType":"XL","onlineDisks":8,"offlineDisks":0}},"server":{"version":"2017-02-02T20:07:53Z","commitID":"aacfb8654ee399922ca70a6f2d86e23e84ce39e5"}}

``` 
## Expected Behavior
Server should report back the server uptime i.e  duration since when the server is online.

## Current Behavior
Do not know since when the server is up.

## Possible Solution
Remember the time when the server has been started after object layer initialization.

## Context
Usability improvements.

## Your Environment
Current `mc master` and `minio master`

krisis:
## Description

This change brings uptime information into ServiceStatus admin REST API.
Fixes #3678 

## Motivation and Context

See #3678 

## How Has This Been Tested?

Manually tested.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/admin-handlers.go
+++ b/cmd/admin-handlers.go
@@ -19,7 +19,6 @@ package cmd
 import (
 	"encoding/json"
 	"encoding/xml"
-	"fmt"
 	"io/ioutil"
 	"net/http"
 	"net/url"
@@ -54,8 +53,8 @@ type ServerVersion struct {
 
 // ServerStatus - contains the response of service status API
 type ServerStatus struct {
-	StorageInfo   StorageInfo   `json:"storageInfo"`
 	ServerVersion ServerVersion `json:"serverVersion"`
+	Uptime        time.Duration `json:"uptime"`
 }
 
 // ServiceStatusHandler - GET /?service
@@ -70,15 +69,22 @@ func (adminAPI adminAPIHandlers) ServiceStatusHandler(w http.ResponseWriter, r *
 		return
 	}
 
-	// Fetch storage backend information
-	storageInfo := newObjectLayerFn().StorageInfo()
 	// Fetch server version
 	serverVersion := ServerVersion{Version: Version, CommitID: CommitID}
 
+	// Fetch uptimes from all peers. This may fail to due to lack
+	// of read-quorum availability.
+	uptime, err := getPeerUptimes(globalAdminPeers)
+	if err != nil {
+		writeErrorResponse(w, toAPIErrorCode(err), r.URL)
+		errorIf(err, "Possibly failed to get uptime from majority of servers.")
+		return
+	}
+
 	// Create API response
 	serverStatus := ServerStatus{
-		StorageInfo:   storageInfo,
 		ServerVersion: serverVersion,
+		Uptime:        uptime,
 	}
 
 	// Marshal API response
@@ -541,7 +547,6 @@ func (adminAPI adminAPIHandlers) HealFormatHandler(w http.ResponseWriter, r *htt
 	// Create a new set of storage instances to heal format.json.
 	bootstrapDisks, err := initStorageDisks(globalEndpoints)
 	if err != nil {
-		fmt.Println(traceError(err))
 		writeErrorResponse(w, toAPIErrorCode(err), r.URL)
 		return
 	}
@@ -549,7 +554,6 @@ func (adminAPI adminAPIHandlers) HealFormatHandler(w http.ResponseWriter, r *htt
 	// Heal format.json on available storage.
 	err = healFormatXL(bootstrapDisks)
 	if err != nil {
-		fmt.Println(traceError(err))
 		writeErrorResponse(w, toAPIErrorCode(err), r.URL)
 		return
 	}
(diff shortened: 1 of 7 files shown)


cmd/admin-rpc-client.go
```diff
@@ -39,6 +40,7 @@ type adminCmdRunner interface {
 	Restart() error
 	ListLocks(bucket, prefix string, duration time.Duration) ([]VolumeLockInfo, error)
 	ReInitDisks() error
+	Uptime() (time.Time, error)
```
vadmeste: Shouldn't uptime be time.Duration and not time.Time ?

krisis: @vadmeste this can be done with ``time.Duration`` too. I happen to pick ``time.Time`` for the comparisons in a distributed setup to compute the uptime of the distributed setup. I have taken uptime to be the duration for which the distributed setup has been available w.r.t read operations.

vadmeste: > @vadmeste this can be done with time.Duration too. I happen to pick time.Time for the comparisons in a distributed setup to compute the uptime of the distributed setup. I have taken uptime to be the duration for which the distributed setup has been available w.r.t read operations.

In that case, I think this API should be called BootTime, but also I really think that we should use duration instead, since duration will be correctly calculated starting from go 1.9 , otherwise you'll need to be sure that the system's time is correct.

krisis: @vadmeste Makes sense, I shall update the PR to use ``time.Duration`` everywhere.

cmd/admin-rpc-client.go
```diff
@@ -88,6 +90,28 @@ func (rc remoteAdminClient) ReInitDisks() error {
...
 }
 
+// Uptime - Returns the uptime of this server. Timestamp is taken
+// after object layer is initialized.
+func (lc localAdminClient) Uptime() (time.Duration, error) {
+	if globalBootTime.IsZero() {
+		return time.Duration(0), errServerNotInitialized
+	}
+
+	return time.Now().UTC().Sub(globalBootTime), nil
```
vadmeste, line 100: Can you add a notice that this is sometimes buggy but will be automatically fixed in go 1.9 ? https://github.com/golang/go/issues/12914

cmd/admin-rpc-server.go
```diff
@@ -105,6 +111,23 @@ func (s *adminCmd) ReInitDisks(args *AuthRPCArgs, reply *AuthRPCReply) error {
...
+	if err := args.IsAuthenticated(); err != nil {
+		return err
+	}
+
+	if globalBootTime.IsZero() {
+		return errServerNotInitialized
+	}
+
+	*reply = UptimeReply{
+		Uptime: time.Now().UTC().Sub(globalBootTime),
```
vadmeste: Can you add a notice that this is sometimes buggy but will be automatically fixed in go 1.9 ? golang/go#12914

cmd/admin-handlers.go
```diff
@@ -56,6 +55,7 @@ type ServerVersion struct {
 type ServerStatus struct {
 	StorageInfo   StorageInfo   `json:"storageInfo"`
```
harshavardhana: We need to get rid of 'StorageInfo' from here and move this to 'ServerInfo()'

cmd/admin-handlers.go
```diff
@@ -70,15 +70,31 @@ func (adminAPI adminAPIHandlers) ServiceStatusHandler(w http.ResponseWriter, r *
...
 
+	objLayer := newObjectLayerFn()
+	if objLayer == nil {
+		writeErrorResponse(w, ErrServerNotInitialized, r.URL)
+		return
+	}
+
 	// Fetch storage backend information
-	storageInfo := newObjectLayerFn().StorageInfo()
+	storageInfo := objLayer.StorageInfo()
```
harshavardhana: We don't need this..

harshavardhana: Change the output JSON for 'status' as below. 

```
$ mc admin service status play/ --json | jq .
{
  "status": "success",
  "service": true,
  "storageInfo": {
    "total": 675875831808,
    "free": 602565951488,
    "backend": {
      "backendType": "XL",
      "onlineDisks": 8,
      "offlineDisks": 0
    }
  },
  "server": {
    "version": "2017-02-02T20:07:53Z",
    "commitID": "aacfb8654ee399922ca70a6f2d86e23e84ce39e5"
  }
}
```

To

```
$ mc admin service status play/ --json | jq .
{
  "status": "success",
  "service": "online|offline",
  "uptime": 0000,
}
```

krisis: @harshavardhana what does ``status`` field in the above json stand for? IIUC, ``service`` could be ``online`` or ``offline``, possibly ``maintenance`` in the future (for servers with all disks unformatted in a distributed setup). Is that right?

harshavardhana: `status` is `mc` output - each command in `mc` reports 'status' - shouldn't be concerned with the API you are going to implement. It is just a full picture here, see `mc ls --json` for more details or `mc mirror --json`. 

I leave it upto to you send a string or some typed enum based on `iota`, it can be easily converted by `mc` for 'service'

harshavardhana (approved):

PR #3694, opened 2017-02-06
[WIP] reduce the min prepare storage requirement.

### LINKED ISSUE #3234: erasure: Replace n/2 disks, new servers do not come up
1. run 2:2 DXL
2. bring down minio3 and minio4 (i.e 3rd and 4th server)
3. rm -rf /export/disk{3..4} (i.e remove the down servers disks)
4. mkdir /export/disk{3..4}
5. start minio3 and minio4

minio 3 and 4 does not come up and prints eternally:
`Initializing data volume. Waiting for minimum 3 servers to come online.`

harshavardhana:
## Description
Reduce the min prepare storage requirement.

## Motivation and Context
The motivation is to address a case where we can't replace fresh disks when N == readQuorum 

Fixes #3234 

## How Has This Been Tested?
Manually. 

## Types of changes

## Checklist:

### DIFF
--- a/cmd/prepare-storage.go
+++ b/cmd/prepare-storage.go
@@ -131,22 +131,10 @@ func prepForInitXL(firstDisk bool, sErrs []error, diskCount int) InitActions {
 		return WaitForConfig
 	}
 
-	quorum := diskCount/2 + 1
+	readQuorum := diskCount / 2
 	disksOffline := errMap[errDiskNotFound]
 	disksFormatted := errMap[nil]
 	disksUnformatted := errMap[errUnformattedDisk]
-	disksCorrupted := errMap[errCorruptedFormat]
-
-	// No Quorum lots of offline disks, wait for quorum.
-	if disksOffline >= quorum {
-		return WaitForQuorum
-	}
-
-	// There is quorum or more corrupted disks, there is not enough good
-	// disks to reconstruct format.json.
-	if disksCorrupted >= quorum {
-		return Abort
-	}
 
 	// All disks are unformatted, proceed to formatting disks.
 	if disksUnformatted == diskCount {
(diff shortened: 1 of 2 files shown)


harshavardhana:
We need to review this cleanly and dig through all the cases where this can fail - @vadmeste @krishnasrinivas @krisis

cmd/prepare-storage.go
```diff
@@ -131,20 +131,20 @@ func prepForInitXL(firstDisk bool, sErrs []error, diskCount int) InitActions {
...
-	quorum := diskCount/2 + 1
+	readQuorum := diskCount / 2
 	disksOffline := errMap[errDiskNotFound]
 	disksFormatted := errMap[nil]
 	disksUnformatted := errMap[errUnformattedDisk]
 	disksCorrupted := errMap[errCorruptedFormat]
 
 	// No Quorum lots of offline disks, wait for quorum.
-	if disksOffline >= quorum {
+	if disksOffline >= readQuorum {
```
vadmeste: If we have 2 offline disks of 4 total disks, disksOffline will be 2 and readQuorum will be also 2, doesn't this mean that we will never be able to run Minio, right?

harshavardhana: Yes good catch should be fixed.

krisis:
What happens if we have split-brain with format.json and use 'read-quorum' for prepare storage phase?

vadmeste:
If we have 4 disks, one disk formatted, one unformatted and one corrupted and one healthy, all nodes will wait for quorum (WaitForQuorum)

krisis:
@vadmeste In a 4-node setup, what happens when two nodes agree on a version of format.json (say f1) and the other two nodes agree on a version of format.json (say f2), which one should we use as the format.json for this setup for initialization and format-healing purposes?

vadmeste:
> @vadmeste In a 4-node setup, what happens when two nodes agree on a version of format.json (say f1) and the other two nodes agree on a version of format.json (say f2), which one should we use as the format.json for this setup for initialization and format-healing purposes?

What I think:
- We should consider the case when the user mistakenly provides the paths of nodes of another cluster, we should not take action until the user explicitly ask for it.
- It is more simple to quit immediately when we detect a weird thing and let heal does its job instead,

harshavardhana:
Will close this PR this is not the right style.. and we have to think about this properly.

PR #3695, opened 2017-02-06
auth/rpc: Make auth rpc client retry configurable.

harshavardhana:
## Description
auth/rpc: Make auth rpc client retry configurable.

## Motivation and Context
Currently the auth rpc client defaults to to a maximum
cap of 30seconds timeout. Make this to be configurable
by the caller of authRPCClient during initialization, if no
such config is provided then default to 30 seconds.

## How Has This Been Tested?
Manually.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/retry.go
+++ b/cmd/retry.go
@@ -48,7 +48,8 @@ func (r *lockedRandSource) Seed(seed int64) {
 // MaxJitter will randomize over the full exponential backoff time
 const MaxJitter = 1.0
 
-// NoJitter disables the use of jitter for randomizing the exponential backoff time
+// NoJitter disables the use of jitter for randomizing the
+// exponential backoff time
 const NoJitter = 0.0
 
 // Global random source for fetching random values.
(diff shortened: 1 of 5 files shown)


cmd/retry_test.go
```diff
@@ -0,0 +1,82 @@
...
+			close(doneCh)
+		}
+	}
+	_, ok := <-attemptCh
+	if ok {
+		t.Fatal("Attempt counter should be closed")
+	}
+}
+
+// Test retry time with jitter 2.0 set back to 1.0
```
krisis: The comment is not clear, it could be "Test retry-timer with jitter greater then MaxJitter of 1.0"

krisis (changes requested):
The changes look good except for the one comment.

cmd/retry_test.go
harshavardhana: done

krisis (approved):

cmd/retry.go
```diff
@@ -113,5 +116,27 @@ func newRetryTimer(unit time.Duration, cap time.Duration, jitter float64, doneCh
...
 	}()
+
+	// Start reading..
 	return attemptCh
 }
+
+// Default retry constants.
+var (
+	defaultRetryUnit = time.Second      // 1 second.
+	defaultRetryCap  = 30 * time.Second // 30 seconds.
```
balamurugana, line 127: repeating unit in `defaultRetryCap` makes confusion here.  `newRetryTimerWithJitter()` could have `cap` a int as argument and used as `cap * unit` inside.

harshavardhana: Cap is not a multiple of Unit - Unit can be '2 * time.Second' as well and as well as something else too.

PR #3714, opened 2017-02-08
[WIP] collapse GetObjectInfo usage in Get handler into GetObject()

harshavardhana:
This is done to avoid major races that can occur with shared
backend, extremely rare but it can happen.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/httprange.go
+++ b/cmd/httprange.go
@@ -23,99 +23,102 @@ import (
 	"strings"
 )
 
-const (
-	byteRangePrefix = "bytes="
-)
-
 // Valid byte position regexp
 var validBytePos = regexp.MustCompile(`^[0-9]+$`)
 
 // HttpRange specifies the byte range to be sent to the client.
-type httpRange struct {
+type objectRange struct {
 	offsetBegin  int64
 	offsetEnd    int64
 	resourceSize int64
 }
 
 // String populate range stringer interface
-func (hrange httpRange) String() string {
-	return fmt.Sprintf("bytes %d-%d/%d", hrange.offsetBegin, hrange.offsetEnd, hrange.resourceSize)
+func (r objectRange) String() string {
+	return fmt.Sprintf("bytes %d-%d/%d", r.offsetBegin, r.offsetEnd, r.resourceSize)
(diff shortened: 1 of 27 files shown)


cmd/httprange.go
```diff
@@ -27,108 +26,51 @@ const (
...
-	return fmt.Sprintf("bytes %d-%d/%d", hrange.offsetBegin, hrange.offsetEnd, hrange.resourceSize)
-}
-
-// getlength - get length from the range.
-func (hrange httpRange) getLength() int64 {
-	return 1 + hrange.offsetEnd - hrange.offsetBegin
-}
-
-func parseRequestRange(rangeString string, resourceSize int64) (hrange *httpRange, err error) {
+func parseRangeStr(rangeStr string) (offsetBegin, offsetEnd int64, err error) {
```
balamurugana: This function works for invalid range strings
```
		"bytes=+2-5",
		"bytes=2-+5",
		"bytes=2--5",
```

cmd/object-api-common.go
```diff
@@ -309,3 +310,186 @@ func cleanupDir(storage StorageAPI, volume, dirPath string) error {
 	err := delFunc(retainSlash(pathJoin(dirPath)))
 	return err
 }
+
+func checkRange(offsetBegin, offsetEnd int64, resourceSize int64) (startOffset, length int64, err error) {
```
balamurugana: I think you could keep `type httpRange struct`  and have `newHTTPRange()` would do this and you could use `getLength()` and `String()` instead of scattering range logic.

cmd/httprange.go
harshavardhana: Its not finished yet once the WIP is removed it can be reviewed.

cmd/object-api-common.go
harshavardhana: No it is on purpose.. httpRange is not the right data structure at object layer and also its not finished yet since its a critical change.  Reviews can be done only after unit tests are added.

harshavardhana:
Closing this as old PR should be worked on differently.

Issue #3700, opened 2017-02-07
config/cleanup: Re-do the serverConfig struct to honor ENV behavior properly


harshavardhana:
## Expected Behavior
When ENV variable is set we shouldn't update the `config.json` if it already exists, save it only if its a fresh start.

## Current Behavior
We override the config and regenerate it.

## Possible Solution
Do not regenerate. 

## Steps to Reproduce (for bugs)
Just set the `config.json` with custom keys and start with ENV server will re-write `config.json` with the ENV keys.

harshavardhana:
This is fixed by #3721 and #3703

PR #3721, opened 2017-02-09
config: setter/getter for Notifier and Logger into its own struct.

harshavardhana:
## Description
This is an attempt cleanup code and keep the top level config
functions simpler and easy to understand where as move the
notifier related code and logger setter/getter methods as part
of their own struct.

Locks are now held properly not globally by configMutex, but
instead as private variables.

## Motivation and Context
Final fix for #3700

## How Has This Been Tested?
Cleanup

## Types of changes

## Checklist:

### DIFF
--- a/cmd/notifiers.go
+++ b/cmd/notifiers.go
@@ -54,24 +54,12 @@ const (
 
 var errNotifyNotEnabled = errors.New("requested notifier not enabled")
 
-// Notifier represents collection of supported notification queues.
-type notifier struct {
-	AMQP          map[string]amqpNotify          `json:"amqp"`
-	NATS          map[string]natsNotify          `json:"nats"`
-	ElasticSearch map[string]elasticSearchNotify `json:"elasticsearch"`
-	Redis         map[string]redisNotify         `json:"redis"`
-	PostgreSQL    map[string]postgreSQLNotify    `json:"postgresql"`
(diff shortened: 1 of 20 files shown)


cmd/notifiers.go
```diff
@@ -66,12 +68,128 @@ type notifier struct {
...
 
+// SetAMQPByID -
+func (n *notifier) SetAMQPByID(accountID string, amqpn amqpNotify) {
+	n.Lock()
+	defer n.Unlock()
+	n.AMQP[accountID] = amqpn
+}
+
+func (n *notifier) GetAMQP() map[string]amqpNotify {
+	return n.AMQP
```
vadmeste: Can we hold a lock here and return a copy of AMQP ? the same of other types of notification.

vadmeste: LGTM otherwise

harshavardhana: Why would you hold a lock ? We are not modifying the members.

vadmeste: > Why would you hold a lock ? We are not modifying the members.

Currently, the caller of GetAMQP will iterate over n.AMQP elements, we can get a problem with a concurrent write

harshavardhana: You are right memory is same.

vadmeste (approved):

Issue #3722, opened 2017-02-09
docs: fix docs to describe different minio setups


krishnasrinivas:
Related: https://github.com/minio/minio/issues/3239

```
The issue is closed as I mentioned what are the agreed terminologies and our documentation is using these terminologies already. It is how we usually use it needs to be changed. We confused the community with XL etc our docs don't talk about it.

Also there are no abbrev .. list of type of setups possible with Minio

Standalone FS
Standalone Erasure
Distributed FS (shared)
Distributed Erasure
```

> The issue is closed as I mentioned what are the agreed terminologies and our documentation is using these terminologies already.

Our docs don't use these terminologies already. They should be fixed.

harshavardhana:
> Our docs don't use these terminologies already. They should be fixed.

@krishnasrinivas  Can you list all the docs which do not follow this in consistent manner? It would help @NitishT

GarimaKapoor:
@koolhead17 please review all the docs for this inconsistency and work with @NitishT to close this issue.

deekoder:
Need a thorough pass on this during the USS Akira release.

krishnasrinivas:
For example:
https://docs.minio.io/docs/minio-quickstart-guide This does not use the term `Standalone FS`

https://docs.minio.io/docs/minio-erasure-code-quickstart-guide does not use the term `Standalone Erasure`

https://docs.minio.io/docs/distributed-minio-quickstart-guide does not use the term `Distributed Erasure`

nitisht:
Can be closed in favor of https://github.com/minio/minio/issues/4529

harshavardhana:
Closing this is not a valid issue anymore. #4529

PR #3725, opened 2017-02-10
Remove XL references in public docs to Erasure.

harshavardhana:
## Description
XL references to be moved to Erasure

## Motivation and Context
Ref #3722

## How Has This Been Tested?
Manual inspection

## Types of changes

## Checklist:


cmd/globals.go
```diff
@@ -131,7 +131,7 @@ var (
 	globalHTTPStats = newHTTPStats()
 
 	// Time when object layer was initialized on start up.
-	globalBootTime time.Time
+	globalBootTime = time.Now().UTC()
```
krisis: Why is the globalBootTime being changed in this PR? If we are defining globalBootTime as time when object layer was initialized, it is wrong to initialize it here.

harshavardhana: It was for testing.. moving back.

Issue #2854, opened 2016-10-05
browser: Implement delete with multi select and confirmation dialog


krishnasrinivas:
Single object delete works. Need to implement multi-object delete.

krishnasrinivas:
timeline 9-oct

GarimaKapoor:
@krishnasrinivas can we not do this earlier. The release is on 10th. We will need to test it as well.

harshavardhana:
Moved out of milestone.

GarimaKapoor:
@krishnasrinivas do you have a timeline for the fix?

krishnasrinivas:
@rushenn are the UI changes done?

rushenn:
@krishnasrinivas UI is ready and converting the style naming conventions. Will push the changes shortly.

harshavardhana:
Fixed by #3730

krishnasrinivas:
@rushenn can you take care of the UI changes after https://github.com/minio/minio/pull/3810 is merged?

rushenn:
@krishnasrinivas sure

harshavardhana:
This is fixed finally by #3859

PR #3730, opened 2017-02-10
browser: Implement multi select user interface for object listings

rushenn:
## Description
Implement multi-select user interface for object listings. This is only the non-functional view part and need further work to make it functional. 

## Motivation and Context
https://github.com/minio/minio/issues/2854

## How Has This Been Tested?
Tested only the UI manually

## Types of changes

## Checklist:


rushenn:
Sample markup for action header:

action-header

```
<div className="list-actions">
      <span className="la-label">
           <i className="fa fa-check-circle"></i> 
           25 Objects selected
      </span>
      <span className="la-actions pull-right">
             <button>Download all as zip</button>
      </span>
</div>
```

harshavardhana:
ping @rushenn @krishnasrinivas is this ready for review?

rushenn:
@krishnasrinivas

harshavardhana:
> @krishnasrinivas

@rushenn what else is pending here?

rushenn:
@harshavardhana working on the functionality.

krishnasrinivas:
checking

krishnasrinivas:
@harshavardhana this has the html/less changes, now @rushenn is working on the JS/redux changes. We had a chat on how to go about the JS changes for zip/multi-delete issues.

.minio.sys/buckets/browser/policy.json
```diff
@@ -0,0 +1 @@
+{"Version":"2012-10-17","Statement":[{"Action":["s3:GetBucketLocation"],"Effect":"Allow","Principal":{"AWS":["*"]},"Resource":["arn:aws:s3:::browser"],"Sid":""},{"Action":["s3:ListBucket"],"Condition":{"StringEquals":{"s3:prefix":["/docs","cmd","fefe fef "]}},"Effect":"Allow","Principal":{"AWS":["*"]},"Resource":["arn:aws:s3:::browser"],"Sid":""},{"Action":["s3:GetObject"],"Effect":"Allow","Principal":{"AWS":["*"]},"Resource":["arn:aws:s3:::browser//docs*","arn:aws:s3:::browser/cmd*","arn:aws:s3:::browser/fefe fef *"],"Sid":""}]}
```
harshavardhana: You shouldn't add these files into your commit.  Also make sure to run minio on a separate directory.

harshavardhana (changes requested):

harshavardhana:
Tests are failing 

```
time="2017-02-20T22:25:32Z" level=error msg="Unable to parse JWT token string" cause="token contains an invalid number of segments" source="[jwt.go:113:isAuthTokenValid()]" 
--- FAIL: TestWebHandlerDownloadZip (0.26s)
	web-handlers_test.go:863: Expected to receive authentication error
```

browser/app/js/actions.js
```diff
@@ -402,6 +404,24 @@ export const setLoginError = () => {
   }
 }
 
+export const downloadAllasZip = (url, req, xhr) => {
+  return (dispatch) => {
+    xhr.open('PUT', url, true)
```
harshavardhana: Why would this be a PUT request, it should be "GET"

cmd/web-router.go
```diff
@@ -84,7 +84,7 @@ func registerWebRouter(mux *router.Router) error {
 	webBrowserRouter.Methods("POST").Path("/webrpc").Handler(webRPC)
 	webBrowserRouter.Methods("PUT").Path("/upload/{bucket}/{object:.+}").HandlerFunc(web.Upload)
 	webBrowserRouter.Methods("GET").Path("/download/{bucket}/{object:.+}").Queries("token", "{token:.*}").HandlerFunc(web.Download)
-	webBrowserRouter.Methods("GET").Path("/zip").Queries("token", "{token:.*}").HandlerFunc(web.DownloadZip)
```
harshavardhana, line 87: We should keep it as "GET" not PUT.  Get returns the reader not PUT.

harshavardhana (changes requested):

browser/app/js/actions.js
rushenn: We couldn't send content using GET. @krishnasrinivas can give a better explanation on this

harshavardhana: in that case we should use POST instead of PUT,  PUT is not expected to do what is being done here.

harshavardhana: Also update the test case that is failing along with it.

krishnasrinivas:
@harshavardhana after changing from PUT to POST it does not work for rushan (MacOS chrome) but works for me (linux chrome)
can you check if it works for you?

krishnasrinivas:
> can you check if it works for you?

OK, it works for him too

harshavardhana:
Odd that windows tests are failing, rebuilding them again.

harshavardhana (approved):

harshavardhana:
There is some problem in the PR which is causing some problems on windows, not sure what would trigger this only change is POST handler.

krishnasrinivas:
```
... obtained int = 411
... expected int = 200
```

411 is content-length required, weird
is triggering the test again a good idea? just in case it was some transient error.

harshavardhana:
> is triggering the test again a good idea? just in case it was some transient error.

Already did it twice. I think we can pull this locally and see what is going on. Must be handlers are tripping over each other after adding POST>

krishnasrinivas:
> Already did it twice. I think we can pull this locally and see what is going on. Must be handlers are tripping over each other after adding POST>
>  Show all reviewers

let me check on a windows machine

krishnasrinivas:
@rushenn can you rebase and push it again? the windows appveyor issue is fixed in master.

harshavardhana (approved):

PR #3741, opened 2017-02-13
Implement ServerConfig admin REST API

krisis:
Returns a valid config.json of the setup. In case of distributed
setup, it checks if quorum or more number of nodes have the same
config.json.

## Description

Fetches config.json from all disks and returns a valid config.json if present.

## Motivation and Context

For users to be able to read config.json of a minio instance without having to log onto the corresponding server and fetch the config.json.
## How Has This Been Tested?

Manually tested using pkg/madmin API also part of this PR.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/admin-handlers.go
+++ b/cmd/admin-handlers.go
@@ -672,3 +672,32 @@ func (adminAPI adminAPIHandlers) HealFormatHandler(w http.ResponseWriter, r *htt
 	// Return 200 on success.
 	writeSuccessResponseHeadersOnly(w)
 }
+
+// GetConfigHandler - GET /?config
+// - x-minio-operation = get
+// Get config.json of this minio setup.
+func (adminAPI adminAPIHandlers) GetConfigHandler(w http.ResponseWriter, r *http.Request) {
+	// Validate request signature.
+	adminAPIErr := checkRequestAuthType(r, "", "", "")
+	if adminAPIErr != ErrNone {
+		writeErrorResponse(w, adminAPIErr, r.URL)
+		return
+	}
+
+	// check if objectLayer is initialized, if not return.
+	if newObjectLayerFn() == nil {
+		writeErrorResponse(w, ErrServerNotInitialized, r.URL)
+		return
+	}
+
+	// Get config.json from all nodes. In a single node setup, it
+	// returns local config.json.
+	configBytes, err := getPeerConfig(globalAdminPeers)
+	if err != nil {
+		writeErrorResponse(w, toAPIErrorCode(err), r.URL)
+		errorIf(err, "Failed to get config from peers")
+		return
+	}
+
+	writeSuccessResponseJSON(w, configBytes)
+}
--- a/cmd/admin-rpc-client.go
+++ b/cmd/admin-rpc-client.go
@@ -17,8 +17,11 @@
 package cmd
 
 import (
+	"encoding/json"
+	"errors"
 	"net/url"
 	"path"
+	"reflect"
 	"sort"
 	"sync"
 	"time"
@@ -41,6 +44,7 @@ type adminCmdRunner interface {
 	ListLocks(bucket, prefix string, duration time.Duration) ([]VolumeLockInfo, error)
 	ReInitDisks() error
 	Uptime() (time.Duration, error)
+	GetConfig() ([]byte, error)
 }
 
 // Restart - Sends a message over channel to the go-routine
@@ -110,6 +114,25 @@ func (rc remoteAdminClient) Uptime() (time.Duration, error) {
 	}
 
 	return reply.Uptime, nil
+}
+
+// GetConfig - returns config.json of the local server.
+func (lc localAdminClient) GetConfig() ([]byte, error) {
+	if serverConfig == nil {
+		return nil, errors.New("config not present")
+	}
+
+	return json.Marshal(serverConfig)
+}
+
+// GetConfig - returns config.json of the remote server.
+func (rc remoteAdminClient) GetConfig() ([]byte, error) {
+	args := AuthRPCArgs{}
+	reply := ConfigReply{}
+	if err := rc.Call("Admin.GetConfig", &args, &reply); err != nil {
+		return nil, err
+	}
+	return reply.Config, nil
 }
 
 // adminPeer - represents an entity that implements Restart methods.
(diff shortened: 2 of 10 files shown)


krisis:
TBD:
- Type to represent config.json in pkg/madmin. For now it is the same as that used in minio server.

Rationale for getPeerConfig to use ``interface{}`` to ``Unmarshal`` the json bytes
- type-specific un-marshalling need not be unique for a given value of the type.
- un-marshalling to ``interface{}`` type invokes the rules as described [here](https://golang.org/pkg/encoding/json/#Unmarshal). This is guaranteed to be canonical.

harshavardhana:
> Rationale for getPeerConfig to use interface{} to Unmarshal the json bytes
> 
> type-specific un-marshalling need not be unique for a given value of the type.
> un-marshalling to interface{} type invokes the rules as described here. This is guaranteed to be canonical.

Not sure i understand this can you elaborate? @krisis

krisis:
@harshavardhana there are 2 reasons to use `interface{}` as the map key and not `ServerConfigV13`,
1. `ServerConfigV13` is not a valid key type - fails to compile.
2. `json` Unmarshal/Marshal for a given type may change with newer versions of the package.

harshavardhana:
> ServerConfigV13 is not a valid key type - fails to compile.

Why do you need `ServerConfigV13` ? you can use what is available `serverConfigV13` 

> json Unmarshal/Marshal for a given type may change with newer versions of the package.

That is why we need to keep the migration code - interface{} doesn't add any value - all the more it will mask issues.

krisis:
> >   ServerConfigV13 is not a valid key type - fails to compile.
> > Why do you need ServerConfigV13 ? you can use what is available serverConfigV13

`serverConfigV13` is the same as `ServerConfigV13`. We are capitalizing the type name to export it in pkg/madmin. So, both the types are not valid key type.

cmd/admin-handlers.go
```diff
@@ -672,3 +672,23 @@ func (adminAPI adminAPIHandlers) HealFormatHandler(w http.ResponseWriter, r *htt
 	// Return 200 on success.
 	writeSuccessResponseHeadersOnly(w)
 }
+
```
harshavardhana, line 675: Can you write a comment here.

cmd/admin-handlers.go
```diff
@@ -672,3 +672,23 @@ func (adminAPI adminAPIHandlers) HealFormatHandler(w http.ResponseWriter, r *htt
...
+func (adminAPI adminAPIHandlers) GetConfigHandler(w http.ResponseWriter, r *http.Request) {
+	// Validate request signature.
+	adminAPIErr := checkRequestAuthType(r, "", "", "")
+	if adminAPIErr != ErrNone {
+		writeErrorResponse(w, adminAPIErr, r.URL)
+		return
+	}
```
harshavardhana, line 685: i guess we should avoid meddling with server config if object layer is not initialized, safer that way.

cmd/admin-rpc-client.go
```diff
@@ -112,6 +116,23 @@ func (rc remoteAdminClient) Uptime() (time.Duration, error) {
 	return reply.Uptime, nil
 }
 
+func (lc localAdminClient) GetConfig() ([]byte, error) {
```
harshavardhana: Comment.

cmd/admin-rpc-client.go
```diff
@@ -112,6 +116,23 @@ func (rc remoteAdminClient) Uptime() (time.Duration, error) {
...
+	if serverConfig == nil {
+		return nil, errors.New("config not present")
+	}
+
+	return json.Marshal(serverConfig)
+}
+
```
harshavardhana, line 127: Comment.

cmd/admin-rpc-client.go
```diff
@@ -336,3 +357,116 @@ func getPeerUptimes(peers adminPeers) (time.Duration, error) {
 
 	return latestUptime, nil
 }
+
+func getPeerConfig(peers adminPeers) ([]byte, error) {
```
harshavardhana: A comment.

cmd/admin-rpc-client.go
```diff
@@ -336,3 +357,116 @@ func getPeerUptimes(peers adminPeers) (time.Duration, error) {
...
+	if err != nil {
+		errorIf(err, "Unable to find a valid server config")
+		return nil, traceError(err)
+	}
+
+	// Return the config.json that was present quorum or more
+	// number of disks.
+	return json.Marshal(configJSON)
+}
+
```
harshavardhana, line 411: Comment.

harshavardhana (changes requested):

cmd/admin-handlers.go
krisis: Done

cmd/admin-handlers.go
krisis: Ok

cmd/admin-rpc-client.go
krisis: Done.

cmd/admin-rpc-client.go
krisis: Done

cmd/admin-rpc-client.go
```diff
@@ -336,3 +359,120 @@ func getPeerUptimes(peers adminPeers) (time.Duration, error) {
...
+	// serverConfigs == [c1, c2, c1, c1]
+	// configCounter == [3, 1, 0, 0]
+	// c1, c2 are the only distinct values that appear.  c1 is
+	// identified by 0, the lowest index it appears in and c2 is
+	// identified by 1. So, we need to find the number of times
+	// each of these distinct values occur.
+
+	// For each of the serverConfig ...
+	for i := range serverConfigs {
+		// Skip nodes were getConfig failed.
```
donatello: `s/were/where/`

cmd/admin-rpc-client.go
```diff
@@ -336,3 +359,120 @@ func getPeerUptimes(peers adminPeers) (time.Duration, error) {
...
+				// is the lower index.
+				configCounter[j]++
+				break
+			} else if j == i {
+				// serverConfigs[i] is equal to no
+				// other value seen before. i.e, it is
+				// unique so far.
+				configCounter[i] = 1
+				break
+			} // else invariance is violated.
```
donatello: What is the invariance? It doesn't seem to be precisely stated anywhere.

The invariance is after `i-`th iteration, `sum of configCounter slice = i - errors seen upto i-th iteration`. Hopefully there's a nicer way to state it.

cmd/admin-rpc-client.go
```diff
@@ -336,3 +359,120 @@ func getPeerUptimes(peers adminPeers) (time.Duration, error) {
...
+	for i := range serverConfigs {
+		// Skip nodes were getConfig failed.
+		if errs[i] != nil {
+			continue
+		}
+		// Find the already seen copy it is equal to.
+		for j := 0; j <= i; j++ {
+			if j < i && configCounter[j] == 0 {
+				// serverConfigs[j] is known to be equal to a
+				// value that was already seen. See e.g above
```
donatello: `s/e.g/example/`

cmd/admin-rpc-client.go
```diff
@@ -336,3 +359,120 @@ func getPeerUptimes(peers adminPeers) (time.Duration, error) {
...
+			} else if j < i && reflect.DeepEqual(serverConfigs[i], serverConfigs[j]) {
+				// serverConfigs[i] is equal to
+				// serverConfigs[j], update
+				// serverConfigs[j]'s counter since it
+				// is the lower index.
+				configCounter[j]++
+				break
+			} else if j == i {
+				// serverConfigs[i] is equal to no
+				// other value seen before. i.e, it is
```
donatello: `s/i.e, it/It`

cmd/admin-rpc-client.go
```diff
@@ -336,3 +359,120 @@ func getPeerUptimes(peers adminPeers) (time.Duration, error) {
...
+// quorum or more number of servers.
+func getValidServerConfig(serverConfigs []serverConfigV13, errs []error) (serverConfigV13, error) {
+	// majority-based quorum
+	quorum := len(serverConfigs)/2 + 1
+
+	// Count the number of disks a config.json was found in.
+	configCounter := make([]int, len(serverConfigs))
+
+	// We group equal serverConfigs by the lowest index of the
+	// same value.  e.g, let us take the following serverConfigs
```
donatello: `s/value. e.g,/value; e.g./`

donatello (approved):
LGTM

harshavardhana (approved):

PR #3743, opened 2017-02-13
admin: Add ServerInfo API()

### LINKED ISSUE #3679: admin: Introduce a new API ServerInfo()

vadmeste:
## Description
Only throughput is missing, not sure how to calculate it yet.

## Motivation and Context
Fixes #3679 

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:

### DIFF
--- a/cmd/admin-handlers.go
+++ b/cmd/admin-handlers.go
@@ -188,6 +188,98 @@ func (adminAPI adminAPIHandlers) ServiceCredentialsHandler(w http.ResponseWriter
 	w.WriteHeader(http.StatusOK)
 }
 
+// ServerProperties holds some server information such as, version, region
+// uptime, etc..
+type ServerProperties struct {
+	Uptime   time.Duration `json:"uptime"`
+	Version  string        `json:"version"`
+	CommitID string        `json:"commitID"`
+	Region   string        `json:"region"`
+	SQSARN   []string      `json:"sqsARN"`
+}
+
+// ServerConnStats holds transferred bytes from/to the server
+type ServerConnStats struct {
+	TotalInputBytes  uint64 `json:"transferred"`
+	TotalOutputBytes uint64 `json:"received"`
+	Throughput       uint64 `json:"throughput,omitempty"`
+}
+
+// ServerInfo holds the information that will be returned by ServerInfo API
+type ServerInfo struct {
+	StorageInfo StorageInfo      `json:"storage"`
+	ConnStats   ServerConnStats  `json:"network"`
+	Properties  ServerProperties `json:"server"`
+}
+
+// ServerInfoHandler - GET /?server-info
+// ----------
+// Get server information
+func (adminAPI adminAPIHandlers) ServerInfoHandler(w http.ResponseWriter, r *http.Request) {
+	// Authenticate request
+	adminAPIErr := checkRequestAuthType(r, "", "", "")
+	if adminAPIErr != ErrNone {
+		writeErrorResponse(w, adminAPIErr, r.URL)
+		return
+	}
+
+	// Build storage info
+	objLayer := newObjectLayerFn()
+	if objLayer == nil {
+		writeErrorResponse(w, ErrServerNotInitialized, r.URL)
+		return
+	}
+	storage := objLayer.StorageInfo()
+
+	// Build list of enabled ARNs queues
+	var arns []string
+	for queueArn := range globalEventNotifier.GetAllExternalTargets() {
+		arns = append(arns, queueArn)
+	}
+
+	// Fetch uptimes from all peers. This may fail to due to lack
+	// of read-quorum availability.
+	uptime, err := getPeerUptimes(globalAdminPeers)
+	if err != nil {
+		writeErrorResponse(w, toAPIErrorCode(err), r.URL)
+		errorIf(err, "Unable to get uptime from majority of servers.")
+		return
+	}
+
+	// Build server properties information
+	properties := ServerProperties{
+		Version:  Version,
+		CommitID: CommitID,
+		Region:   serverConfig.GetRegion(),
+		SQSARN:   arns,
+		Uptime:   uptime,
+	}
+
+	// Build network info
+	connStats := ServerConnStats{
+		TotalInputBytes:  globalConnStats.getTotalInputBytes(),
+		TotalOutputBytes: globalConnStats.getTotalOutputBytes(),
+	}
+
+	// Build the whole returned information
+	info := ServerInfo{
+		StorageInfo: storage,
+		ConnStats:   connStats,
+		Properties:  properties,
+	}
+
+	// Marshal API response
+	jsonBytes, err := json.Marshal(info)
+	if err != nil {
+		writeErrorResponse(w, ErrInternalError, r.URL)
+		errorIf(err, "Failed to marshal storage info into json.")
+		return
+	}
+	// Reply with storage information (across nodes in a
+	// distributed setup) as json.
+	writeSuccessResponseJSON(w, jsonBytes)
+}
+
 // validateLockQueryParams - Validates query params for list/clear locks management APIs.
 func validateLockQueryParams(vars url.Values) (string, string, time.Duration, APIErrorCode) {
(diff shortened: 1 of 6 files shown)


vadmeste:
@harshavardhana, can you review how SetCredentials is renamed ? and how I removed operation header. (This probably needs a little discussion)

cmd/admin-handlers.go
```diff
@@ -123,12 +123,11 @@ type setCredsReq struct {
...
 }
 
-// ServiceCredsHandler - POST /?service
-// HTTP header x-minio-operation: creds
+// CredendialsHandler - POST /?credentials
 // ----------
 // Update credentials in a minio server. In a distributed setup, update all the servers
 // in the cluster.
-func (adminAPI adminAPIHandlers) ServiceCredentialsHandler(w http.ResponseWriter, r *http.Request) {
+func (adminAPI adminAPIHandlers) CredentialsHandler(w http.ResponseWriter, r *http.Request) {
```
harshavardhana: Don't change Credentials handler in this patch send it as separate fix.

cmd/admin-handlers.go
```diff
@@ -188,6 +187,92 @@ func (adminAPI adminAPIHandlers) ServiceCredentialsHandler(w http.ResponseWriter
...
+	Version  string        `json:"version"`
+	CommitID string        `json:"commitID"`
+	Region   string        `json:"region"`
+	SQSARN   []string      `json:"sqsARN"`
+}
+
+// Holds transferred bytes from/to the server
+type ServerConnStats struct {
+	TotalInputBytes  uint64 `json:"transferred"`
+	TotalOutputBytes uint64 `json:"received"`
```
harshavardhana: Keep the throughput header here and keep it `omitempty` if its not set. We will figure out the algorithm for that.

cmd/admin-handlers.go
```diff
@@ -188,6 +187,92 @@ func (adminAPI adminAPIHandlers) ServiceCredentialsHandler(w http.ResponseWriter
...
+	for queueArn := range globalEventNotifier.external.targets {
+		arns = append(arns, queueArn)
+	}
+
+	// Fetch uptimes from all peers. This may fail to due to lack
+	// of read-quorum availability.
+	uptime, err := getPeerUptimes(globalAdminPeers)
+	if err != nil {
+		writeErrorResponse(w, toAPIErrorCode(err), r.URL)
+		errorIf(err, "Possibly failed to get uptime from majority of servers.")
```
harshavardhana: "Unable to get uptime from majority of servers."

cmd/admin-handlers.go
```diff
@@ -188,6 +187,92 @@ func (adminAPI adminAPIHandlers) ServiceCredentialsHandler(w http.ResponseWriter
...
+	// Authenticate request
+	adminAPIErr := checkRequestAuthType(r, "", "", "")
+	if adminAPIErr != ErrNone {
+		writeErrorResponse(w, adminAPIErr, r.URL)
+		return
+	}
+
+	// Build list of enabled ARNs queues
+	arns := []string{}
+	for queueArn := range globalEventNotifier.external.targets {
```
harshavardhana: Don't access them like this  its not protected, add a function to `serverConfig` to provide this value.

cmd/admin-handlers.go
```diff
@@ -188,6 +187,92 @@ func (adminAPI adminAPIHandlers) ServiceCredentialsHandler(w http.ResponseWriter
...
+	}
+
+	// Build network info
+	connStats := ServerConnStats{
+		TotalInputBytes:  globalConnStats.getTotalInputBytes(),
+		TotalOutputBytes: globalConnStats.getTotalOutputBytes(),
+	}
+
+	// Build storage info
+	storage := newObjectLayerFn().StorageInfo()
```
harshavardhana: Make sure to check for if object layer is initialized or not do not enter here if object layer is not initialized.

harshavardhana (changes requested):

cmd/admin-handlers.go
vadmeste: > Don't access them like this its not protected, add a function to serverConfig to provide this value.

Comments in the code indicates that this map will never be modified after startup, that's why it doesn't have any protection.

cmd/admin-handlers.go
vadmeste: > "Unable to get uptime from majority of servers."

Done

cmd/admin-handlers.go
vadmeste: > Make sure to check for if object layer is initialized or not do not enter here if object layer is not initialized.

Done

vadmeste:
All comments are addressed.

cmd/admin-handlers.go
```diff
@@ -188,6 +188,98 @@ func (adminAPI adminAPIHandlers) ServiceCredentialsHandler(w http.ResponseWriter
...
+	objLayer := newObjectLayerFn()
+	if objLayer == nil {
+		writeErrorResponse(w, ErrServerNotInitialized, r.URL)
+		return
+	}
+	storage := objLayer.StorageInfo()
+
+	// Build list of enabled ARNs queues
+	arns := []string{}
+	for queueArn := range globalEventNotifier.external.targets {
```
harshavardhana: Can you find a way to get ARNs not by accessing internal variables like this? - add a method perhaps to `globalEventNotifier` ?

cmd/admin-handlers.go
```diff
@@ -188,6 +188,98 @@ func (adminAPI adminAPIHandlers) ServiceCredentialsHandler(w http.ResponseWriter
...
+	// Build storage info
+	objLayer := newObjectLayerFn()
+	if objLayer == nil {
+		writeErrorResponse(w, ErrServerNotInitialized, r.URL)
+		return
+	}
+	storage := objLayer.StorageInfo()
+
+	// Build list of enabled ARNs queues
+	arns := []string{}
```
harshavardhana: This is not needed just use `var args []string` more idomatic Go.

cmd/admin-router.go
```diff
@@ -39,6 +39,9 @@ func registerAdminRouter(mux *router.Router) {
 	// Service update credentials
 	adminRouter.Methods("POST").Queries("service", "").Headers(minioAdminOpHeader, "set-credentials").HandlerFunc(adminAPI.ServiceCredentialsHandler)
 
+	// Info operations
+	adminRouter.Methods("GET").Queries("server-info", "").HandlerFunc(adminAPI.ServerInfoHandler)
```
harshavardhana: May be just 'info' it can have other params in future i.e if at all `info` gets an argument for example 

```
info=disk
```

Just returns only disk, currently since it is without any argument returns everything.

harshavardhana (changes requested):

harshavardhana (approved):

PR #3780, opened 2017-02-21
update: For source builds look for absolute path.

harshavardhana:
## Description
update: For source builds look for absolute path.

## Motivation and Context
os.Args[0] doesn't point to absolute path we need
use exec.LookPath to find the absolute path before
sending os.Stat().

## How Has This Been Tested?
Manually after using `go install -v` 

## Types of changes

## Checklist:


cmd/update-main.go
```diff
@@ -62,15 +63,22 @@ EXAMPLES:
 `,
 }
 
-const releaseTagTimeLayout = "2006-01-02T15-04-05Z"
-
-const minioReleaseURL = "https://dl.minio.io/server/minio/release/" + runtime.GOOS + "-" + runtime.GOARCH + "/"
+const (
+	minioReleaseTagFormat = "2006-01-02T15-04-05Z"
```
balamurugana: the name `minioReleaseTagFormat` looks like a release tag format however the const is a time format portion in the release tag.

golang calls time format as layout as its not a format string like in `strftime()`

harshavardhana: That is just a variable name - const are called what they really are - RFC3339 not RFC3339Layout.  Atleast that's how i read it.

balamurugana: I think the const name needs to be properly named to differentiate release tag format and time portion in release tag.

cmd/update-main_test.go
```diff
@@ -68,9 +68,9 @@ func TestGetCurrentReleaseTime(t *testing.T) {
 		{minioVersion2, minioBinaryPath2, releaseTime2, nil},
 		{"junk", minioBinaryPath2, releaseTime2, nil},
 		{"3.2.0", minioBinaryPath2, releaseTime2, nil},
-		{minioVersion2, "", time.Time{}, fmt.Errorf(errorMessage1)},
-		{"junk", "non-existing-file", time.Time{}, fmt.Errorf(errorMessage2)},
-		{"3.2.0", "non-existing-file", time.Time{}, fmt.Errorf(errorMessage2)},
+		{minioVersion2, "", time.Time{}, errors.New(errorMessage1)},
+		{"junk", "non-existing-file", time.Time{}, errors.New(errorMessage2)},
+		{"3.2.0", "non-existing-file", time.Time{}, errors.New(errorMessage2)},
```
balamurugana: You would need to add specific test cases for this PR ie exec name is absolute path.

harshavardhana: That is already added automatically `minioBinaryPath2` is already absolute.

harshavardhana: adding one more case with non-abs path

cmd/update-main.go
harshavardhana: Alright will keep it as before but with `minio` name added.

cmd/update-main_test.go
balamurugana: That's the only success cases.  How about `C:\non-existing-file` and `/usr/bin/non-existing-file`, and existing file in the path with relative path like `ls` command from $PATH.

These tests are required to increase code coverage.

cmd/update-main_test.go
```diff
@@ -17,44 +17,64 @@
...
+	minioBinaryPath2 := tmpfile1.Name()
+	fi, err := tmpfile1.Stat()
 	if err != nil {
 		t.Fatalf("Unable to get temporary file info. %s", err)
 	}
-	if err = tmpfile.Close(); err != nil {
+	if err = tmpfile1.Close(); err != nil {
 		t.Fatalf("Unable to create temporary file. %s", err)
 	}
 	releaseTime2 := fi.ModTime().UTC()
```
balamurugana, line 52: It would be nice to keep `minioVersion2`, `minioBinaryPath2` and `releaseTime2` init statements grouped ie `minioVersion3` and `minioBinaryPath3` and `releaseTime3` are inserted in between

cmd/update-main_test.go
```diff
@@ -17,44 +17,64 @@
...
 	"net/http"
 	"net/http/httptest"
 	"os"
+	"os/exec"
 	"runtime"
 	"testing"
 	"time"
 )
 
+func goBinModTime(goBinPath string, t *testing.T) (modTime time.Time) {
```
balamurugana: You could move entire logic into `TestGetCurrentReleaseTime()`

cmd/update-main_test.go
```diff
@@ -66,11 +86,12 @@ func TestGetCurrentReleaseTime(t *testing.T) {
 		{minioVersion1, "", releaseTime1, nil},
 		{minioVersion1, minioBinaryPath2, releaseTime1, nil},
 		{minioVersion2, minioBinaryPath2, releaseTime2, nil},
+		{minioVersion2, minioBinaryPath3, releaseTime3, nil},
```
balamurugana, line 102: Also I think we would need to add nonexistent absolute path for minio binary test cases.

cmd/update-main_test.go
harshavardhana: It is kept here for readability.

cmd/update-main_test.go
harshavardhana: Let me do that.

cmd/update-main_test.go
harshavardhana: Let me add that as well

cmd/update-main.go
```diff
@@ -62,15 +64,24 @@ EXAMPLES:
...
+		// Make sure to look for the absolute path of the binary.
+		minioBinaryPath, err = exec.LookPath(minioBinaryPath)
+		if err != nil {
+			return releaseTime, err
+		}
+	}
+
 	// Looks like version is minio non-standard, we use minio binary's ModTime as release time.
 	fi, err := os.Stat(minioBinaryPath)
 	if err != nil {
```
balamurugana, line 87: When nonexistent bin name is given as absolute path this error `err = fmt.Errorf("Unable to get ModTime of %s. %s", minioBinaryPath, err)` is going to occur. no test case covers this

harshavardhana: This never occurs.. it never occurs in an ideal world because exec.LookPath() would have failed already.

cmd/server_test.go
```diff
@@ -1,2588 +0,0 @@
-/*
```
balamurugana: Is it expected to removal of this file?

harshavardhana: oops no.

balamurugana (approved):

Issue #3784, opened 2017-02-21
Case of missing some files after uploading large amount of data


dcharbonnier:
## Expected Behavior
I expect to retrieve my file after uploading them

## Current Behavior
Version RELEASE.2017-02-16T01-47-30Z is corrupting the files, it return a 200 on a PUT but then the server return 503 or 404 depending of the case.
3 files on 100000's so don't expect to spot it with a simple test.

harshavardhana:
@dcharbonnier are you using multipart or single PUT to upload the files ?

harshavardhana:
Also any logs that you might share would be helpful . We can look at the backend directly and see if the files really uploaded or not. It PUT said 200 it is guaranteed that the file is indeed written to disk. GET receiving 404 or 503 is either quorum related issue which is perhaps generated due to some network issue.

harshavardhana:
https://gist.github.com/dcharbonnier/7bbf9195f803866cba608f8a20a542fc - missing files looks like a successful rename.

harshavardhana:
Relevant `xl.json` value 

```
{"version":"1.0.0","format":"xl","stat":{"size":4224360,"modTime":"2017-02-21T12:41:30.87916948Z"},"erasure":{"algorithm":"klauspost/reedsolomon/vandermonde","data":2,"parity":2,"blockSize":10485760,"index":1,"distribution":[3,4,1,2],"checksum":[{"name":"part.1","algorithm":"blake2b","hash":"786a02f742015903c6c6fd852552d272912f4740e15847618a86e217f71f5419d25e1031afee585313896444934eb04b903a685b1448b755d56f701afe9be2ce"}]},"minio":{"release":"RELEASE.2017-02-16T01-47-30Z"},"meta":{"md5Sum":"2a5fb825c508bc66870b63a86f0b2566"},"parts":[{"number":1,"name":"part.1","etag":"","size":4224360}]}
```

```
time="2017-02-21T15:28:24Z" level=error msg="Unable to create an object." cause="#has incomplete body" source="[object-handlers.go:479:objectAPIHandlers.PutObjectHandler()]" stack="erasure-createfile.go:44:erasureCreateFile xl-v1-object.go:558:xlObjects.PutObject <autogenerated>:848:(*xlObjects).PutObject object-handlers.go:476:objectAPIHandlers.PutObjectHandler api-router.go:60:(objectAPIHandlers).PutObjectHandler-fm"
[...]

PR #3785, opened 2017-02-21
xl/multipart: Make sure to delete temp renamed object.

harshavardhana:
## Description
Make sure to delete temp renamed object.

## Motivation and Context
Existing objects before overwrites are renamed to
temp location in completeMultipart. We make sure
that we delete it even if subsequenty calls fail.

Additionally move verifying of parent dir is a
file earlier to fail the entire operation.

Ref #3784

## How Has This Been Tested?
Manually and `go test`

## Types of changes

## Checklist:

### DIFF
--- a/cmd/object-handlers.go
+++ b/cmd/object-handlers.go
@@ -483,7 +483,7 @@ func (api objectAPIHandlers) PutObjectHandler(w http.ResponseWriter, r *http.Req
 		objInfo, err = objectAPI.PutObject(bucket, object, size, r.Body, metadata, sha256sum)
 	}
 	if err != nil {
-		errorIf(err, "Unable to create an object.")
+		errorIf(err, "Unable to create an object. %s", r.URL.Path)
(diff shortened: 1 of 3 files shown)


krishnasrinivas:
I am unable to understand what sequence would have led to the state in https://gist.github.com/dcharbonnier/7bbf9195f803866cba608f8a20a542fc
and how this patch fixes it. Can you explain more?

Also note that '/minio/working-files/b8527691-9fb8-45b8-af2c-a45cca3953d0' here working-files would be the bucket and b8527691-9fb8-45b8-af2c-a45cca3953d0 would be the object. It is not in the tmp directory.

harshavardhana:
> I am unable to understand what sequence would have led to the state in https://gist.github.com/dcharbonnier/7bbf9195f803866cba608f8a20a542fc
> and how this patch fixes it. Can you explain more?

This doesn't fix anything related to that is why its just a reference @krishnasrinivas . This is just a change after doing code review while investigating that issue.

krishnasrinivas (approved):

PR #3792, opened 2017-02-23
Implement SetConfig admin API handler.

### LINKED ISSUE #3699: admin: Add SetConfig API takes `config.json` as input
```
mc admin config set <filename-to-upload.json>	if not reads from 'stdin' (sets the server `config.json`)
```

## Expected Behavior
Implement a way to edit/set `config.json` from the `mc` side. 

## Current Behavior
No way to set the logger and notification parameters across all nodes.

## Possible Solution
Implement admin API SetConfig()

krisis:
## Description

This PR implements SetConfig admin API REST handler and the corresponding pkg/madmin API.
Fixes #3699 

## Motivation and Context

See #3699 
## How Has This Been Tested?

Manually and unit tests.
## Types of changes

## Checklist:

### DIFF
--- a/cmd/admin-handlers.go
+++ b/cmd/admin-handlers.go
@@ -17,8 +17,10 @@
 package cmd
 
 import (
+	"bytes"
 	"encoding/json"
 	"encoding/xml"
+	"fmt"
 	"io/ioutil"
 	"net/http"
 	"net/url"
@@ -27,7 +29,8 @@ import (
 )
 
 const (
-	minioAdminOpHeader = "X-Minio-Operation"
+	minioAdminOpHeader   = "X-Minio-Operation"
+	minioConfigTmpFormat = "config-%s.json"
 )
 
 // Type-safe query params.
@@ -694,10 +697,127 @@ func (adminAPI adminAPIHandlers) GetConfigHandler(w http.ResponseWriter, r *http
 	// returns local config.json.
 	configBytes, err := getPeerConfig(globalAdminPeers)
 	if err != nil {
-		writeErrorResponse(w, toAPIErrorCode(err), r.URL)
 		errorIf(err, "Failed to get config from peers")
+		writeErrorResponse(w, toAdminAPIErrCode(err), r.URL)
 		return
 	}
 
 	writeSuccessResponseJSON(w, configBytes)
+}
+
+// toAdminAPIErrCode - converts errXLWriteQuorum error to admin API
+// specific error.
+func toAdminAPIErrCode(err error) APIErrorCode {
+	switch err {
+	case errXLWriteQuorum:
+		return ErrAdminConfigNoQuorum
+	}
+	return toAPIErrorCode(err)
+}
+
+// SetConfigResult - represents detailed results of a set-config
+// operation.
+type nodeSummary struct {
+	Name string `json:"name"`
+	Err  string `json:"err"`
+}
+type setConfigResult struct {
+	NodeResults []nodeSummary `json:"nodeResults"`
+	Status      bool          `json:"status"`
+}
+
+// writeSetConfigResponse - writes setConfigResult value as json depending on the status.
+func writeSetConfigResponse(w http.ResponseWriter, peers adminPeers, errs []error, status bool, reqURL *url.URL) {
+	var nodeResults []nodeSummary
+	// Build nodeResults based on error values received during
+	// set-config operation.
+	for i := range errs {
+		nodeResults = append(nodeResults, nodeSummary{
+			Name: peers[i].addr,
+			Err:  fmt.Sprintf("%v", errs[i]),
+		})
+
+	}
+
+	result := setConfigResult{
+		Status:      status,
+		NodeResults: nodeResults,
+	}
+
+	// The following elaborate json encoding is to avoid escaping
+	// '<', '>' in <nil>. Note: json.Encoder.Encode() adds a
+	// gratuitous "\n".
+	var resultBuf bytes.Buffer
+	enc := json.NewEncoder(&resultBuf)
+	enc.SetEscapeHTML(false)
+	jsonErr := enc.Encode(result)
+	if jsonErr != nil {
+		writeErrorResponse(w, toAPIErrorCode(jsonErr), reqURL)
+		return
+	}
+
+	writeSuccessResponseJSON(w, resultBuf.Bytes())
+	return
+}
+
+// SetConfigHandler - PUT /?config
+// - x-minio-operation = set
+func (adminAPI adminAPIHandlers) SetConfigHandler(w http.ResponseWriter, r *http.Request) {
+	// Get current object layer instance.
+	objectAPI := newObjectLayerFn()
+	if objectAPI == nil {
+		writeErrorResponse(w, ErrServerNotInitialized, r.URL)
+		return
+	}
+
+	// Validate request signature.
+	adminAPIErr := checkRequestAuthType(r, "", "", "")
+	if adminAPIErr != ErrNone {
+		writeErrorResponse(w, adminAPIErr, r.URL)
+		return
+	}
+
+	// Read configuration bytes from request body.
+	configBytes, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		errorIf(err, "Failed to read config from request body.")
+		writeErrorResponse(w, toAPIErrorCode(err), r.URL)
+		return
+	}
+
+	// Write config received from request onto a temporary file on
+	// all nodes.
+	tmpFileName := fmt.Sprintf(minioConfigTmpFormat, mustGetUUID())
+	errs := writeTmpConfigPeers(globalAdminPeers, tmpFileName, configBytes)
+
+	// Check if the operation succeeded in quorum or more nodes.
+	rErr := reduceWriteQuorumErrs(errs, nil, len(globalAdminPeers)/2+1)
+	if rErr != nil {
+		writeSetConfigResponse(w, globalAdminPeers, errs, false, r.URL)
+		return
+	}
+
+	// Take a lock on minio/config.json. NB minio is a reserved
+	// bucket name and wouldn't conflict with normal object
+	// operations.
+	configLock := globalNSMutex.NewNSLock(minioReservedBucket, globalMinioConfigFile)
+	configLock.Lock()
+	defer configLock.Unlock()
+
+	// Rename the temporary config file to config.json
+	errs = commitConfigPeers(globalAdminPeers, tmpFileName)
+	rErr = reduceWriteQuorumErrs(errs, nil, len(globalAdminPeers)/2+1)
+	if rErr != nil {
+		writeSetConfigResponse(w, globalAdminPeers, errs, false, r.URL)
+		return
+	}
+
+	// serverMux (cmd/server-mux.go) implements graceful shutdown,
+	// where all listeners are closed and process restart/shutdown
+	// happens after 5s or completion of all ongoing http
+	// requests, whichever is earlier.
+	writeSetConfigResponse(w, globalAdminPeers, errs, true, r.URL)
+
+	// Restart all node for the modified config to take effect.
+	sendServiceCmd(globalAdminPeers, serviceRestart)
 }
--- a/cmd/admin-rpc-client.go
+++ b/cmd/admin-rpc-client.go
@@ -20,11 +20,24 @@ import (
 	"encoding/json"
 	"errors"
 	"net/url"
+	"os"
 	"path"
+	"path/filepath"
 	"reflect"
 	"sort"
 	"sync"
 	"time"
+)
+
+const (
+	// Admin service names
+	serviceRestartRPC = "Admin.Restart"
+	listLocksRPC      = "Admin.ListLocks"
+	reInitDisksRPC    = "Admin.ReInitDisks"
+	uptimeRPC         = "Admin.Uptime"
+	getConfigRPC      = "Admin.GetConfig"
+	writeTmpConfigRPC = "Admin.WriteTmpConfig"
+	commitConfigRPC   = "Admin.CommitConfig"
 )
 
 // localAdminClient - represents admin operation to be executed locally.
@@ -45,6 +58,8 @@ type adminCmdRunner interface {
 	ReInitDisks() error
 	Uptime() (time.Duration, error)
 	GetConfig() ([]byte, error)
+	WriteTmpConfig(tmpFileName string, configBytes []byte) error
+	CommitConfig(tmpFileName string) error
 }
 
 // Restart - Sends a message over channel to the go-routine
@@ -63,7 +78,7 @@ func (lc localAdminClient) ListLocks(bucket, prefix string, duration time.Durati
 func (rc remoteAdminClient) Restart() error {
 	args := AuthRPCArgs{}
 	reply := AuthRPCReply{}
-	return rc.Call("Admin.Restart", &args, &reply)
+	return rc.Call(serviceRestartRPC, &args, &reply)
 }
 
 // ListLocks - Sends list locks command to remote server via RPC.
@@ -74,7 +89,7 @@ func (rc remoteAdminClient) ListLocks(bucket, prefix string, duration time.Durat
 		duration: duration,
 	}
 	var reply ListLocksReply
-	if err := rc.Call("Admin.ListLocks", &listArgs, &reply); err != nil {
+	if err := rc.Call(listLocksRPC, &listArgs, &reply); err != nil {
 		return nil, err
 	}
 	return reply.volLocks, nil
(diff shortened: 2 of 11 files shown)


cmd/admin-handlers.go
```diff
@@ -694,10 +694,76 @@ func (adminAPI adminAPIHandlers) GetConfigHandler(w http.ResponseWriter, r *http
 	// returns local config.json.
 	configBytes, err := getPeerConfig(globalAdminPeers)
 	if err != nil {
-		writeErrorResponse(w, toAPIErrorCode(err), r.URL)
+		var adminErrCode APIErrorCode
```
vadmeste: Can we write toAdminAPiErrorCode() since we need it in more than one place

cmd/admin-handlers.go
```diff
@@ -694,10 +694,76 @@ func (adminAPI adminAPIHandlers) GetConfigHandler(w http.ResponseWriter, r *http
...
+		} else {
+			adminErrCode = toAPIErrorCode(err)
+		}
+		writeErrorResponse(w, adminErrCode, r.URL)
+		return
+	}
+
+	// Send the length config file contents.
+	contentLengthStr := strconv.FormatInt(int64(n), 10)
+	w.Header().Set("Content-Length", contentLengthStr)
```
vadmeste: Can you add a comment to explain why we need to set content-length ?

cmd/admin-rpc-client.go
```diff
@@ -135,6 +142,77 @@ func (rc remoteAdminClient) GetConfig() ([]byte, error) {
...
+}
+
+func (lc localAdminClient) CommitConfig(tmpFileName string) error {
+	configDir, err := getConfigPath()
+	if err != nil {
+		errorIf(err, "Failed to get config path.")
+		return err
+	}
+
+	configFilePath := filepath.Join(configDir, "config.json")
```
vadmeste: You can directly use getConfigFile()

krisis: We need config file's path and not an `os.File` of it.

cmd/admin-rpc-client.go
```diff
@@ -489,3 +567,65 @@ func getValidServerConfig(serverConfigs []serverConfigV13, errs []error) (server
...
+		wg.Add(1)
+		go func(idx int, peer adminPeer) {
+			defer wg.Done()
+			_, errs[idx] = peer.cmdRunner.WriteTmpConfig(tmpFileName, configBytes)
+		}(i, peer)
+	}
+	wg.Wait()
+
+	// Check if the write succeeded in quorum or more number of nodes.
+	rErr := reduceWriteQuorumErrs(errs, nil, len(peers)/2+1)
```
vadmeste: Probably here, we should return all errors, instead of reducing them, this is an Admin API, better for the admin to see all problems instead.

cmd/admin-rpc-client.go
vadmeste: > We need config file's path and not an os.File of it.

getConfigFile() returns a string and it does exactly what you did here.

cmd/admin-rpc-client.go
```diff
@@ -489,3 +567,65 @@ func getValidServerConfig(serverConfigs []serverConfigV13, errs []error) (server
...
+	for i, peer := range peers {
+		wg.Add(1)
+		go func(idx int, peer adminPeer) {
+			defer wg.Done()
+			errs[idx] = peer.cmdRunner.CommitConfig(tmpFileName)
+		}(i, peer)
+	}
+	wg.Wait()
+
+	rErr := reduceWriteQuorumErrs(errs, nil, len(peers)/2+1)
```
vadmeste: Probably here, we should return all errors, instead of reducing them, this is an Admin API, better for the admin to see all problems instead.

cmd/admin-rpc-client.go
krisis: We reduce the errors to know how many nodes this operation succeeded on, to be able to proceed with setting of config across all the nodes in a minio distributed setup. If we don't have a majority of success, we fail the operation. Failing on a single failure, will make it difficult to change config when a node is permanently down.

cmd/admin-rpc-client.go
krisis: Yes, you are right. I am making a change, as we speak, to write the temporary config file inside the config dir and not os.TempDir() as it is now. When I tested locally on a distributed Minio test setup, `/tmp/` is being shared by all minio server instances running on the node. Config directory OTOH could be provided separate for each instance.

cmd/admin-handlers.go
```diff
@@ -694,10 +694,76 @@ func (adminAPI adminAPIHandlers) GetConfigHandler(w http.ResponseWriter, r *http
...
+
+	// Send the length of config file contents. This is to be inline
+	// with any Write like operation. We return the number of
+	// bytes written.
+	contentLengthStr := strconv.FormatInt(int64(n), 10)
+	w.Header().Set("Content-Length", contentLengthStr)
+	writeSuccessResponseHeadersOnly(w)
+
+	// Restart all node for the modified config to take effect.
+	sendServiceCmd(globalAdminPeers, serviceRestart)
```
donatello: The restart event could be racing with the response reaching the client.

The client should ideally have a good opportunity to know that the config update request is successful. If it gets an error response because the server is restarting the client may issue the request again.

Perhaps the restart request can be done in a defer with a 3 second delay? The delay is just to give the client a good chance to receive the response.

cmd/admin-handlers.go
```diff
@@ -694,10 +694,76 @@ func (adminAPI adminAPIHandlers) GetConfigHandler(w http.ResponseWriter, r *http
...
+	}
+
+	// Validate request signature.
+	adminAPIErr := checkRequestAuthType(r, "", "", "")
+	if adminAPIErr != ErrNone {
+		writeErrorResponse(w, adminAPIErr, r.URL)
+		return
+	}
+
+	tmpFileName := "config-" + mustGetUUID()
```
donatello: Perhaps adding a `.json` suffix to the tmp file name will make a stray file look more obvious for a server admin/operator.

cmd/admin-rpc-client.go
donatello: OTOH, if a minority of servers fail and remain in the older configuration, we should let the server operator/admin know so they can take manual action.

cmd/api-errors.go
```diff
@@ -593,6 +594,11 @@ var errorCodeResponse = map[APIErrorCode]APIError{
 		Description:    "The secret key is invalid.",
 		HTTPStatusCode: http.StatusBadRequest,
 	},
+	ErrAdminConfigNoQuorum: {
+		Code:           "XMinioAdminConfigNoQuorum",
+		Description:    "Minio Config doesn't have quorum.",
```
donatello: Isn't this a bit cryptic? How about "Configuration update failed because a server quorum was not met"?

cmd/admin-rpc-client.go
krisis: Hmm, then we could send a json output as below, where
status - "Success" when quorum or more nodes succeeded in setting config.
errors - List of errors seen on respective nodes. 

```json
"SetConfigResult" : {
  "errors" : ["DiskNotFound", "nil", "nil", "nil"],
  "status" : "Success"
}
```

cmd/admin-handlers.go
krisis: @donatello serverMux (``cmd/server-mux.go``) implements graceful shutdown, where all listeners are closed and process restart/shutdown happens after 5s or completion of all ongoing http requests, whichever is earlier.

cmd/api-errors.go
krisis: Yes, will change this.

krisis:
The following is a sample output from a 4-node distributed minio setup.
```shell
➜  madmin git:(issue/3699) ✗ go run examples/set-config.go
SetConfig:  {
	"nodeResults": [
		{
			"name": "localhost:9001",
			"err": "nil"
		},
		{
			"name": "localhost:9002",
			"err": "nil"
		},
		{
			"name": "localhost:9003",
			"err": "nil"
		},
		{
			"name": "localhost:9004",
			"err": "nil"
		}
	],
	"status": true
}

```

cmd/admin-rpc-client.go
```diff
@@ -17,16 +17,30 @@
...
+	"io"
 	"net/url"
+	"os"
 	"path"
+	"path/filepath"
 	"reflect"
 	"sort"
 	"sync"
 	"time"
 )
```
harshavardhana, line 31: Add a comment..

cmd/admin-handlers.go
```diff
@@ -694,10 +696,121 @@ func (adminAPI adminAPIHandlers) GetConfigHandler(w http.ResponseWriter, r *http
...
+
+	// Write config received from request onto a temporary file on
+	// all nodes.
+	tmpFileName := fmt.Sprintf(tmpFileFormat, mustGetUUID())
+	_, errs := writeTmpConfigPeers(globalAdminPeers, tmpFileName, configBytes)
+
+	// Check if the operation succeeded in quorum or more nodes.
+	rErr := reduceWriteQuorumErrs(errs, nil, len(globalAdminPeers)/2+1)
+	if rErr != nil {
+		writeSetConfigResponse(w, errs, false, r.URL)
```
harshavardhana: A 200 OK here is not expected, when sending errors which didn't satisfy write quorum - we should send 500 or something so that the client can retry if necessary.

cmd/admin-handlers.go
```diff
@@ -28,6 +29,7 @@ import (
 
 const (
 	minioAdminOpHeader = "X-Minio-Operation"
+	tmpFileFormat      = "config-%s.json"
```
harshavardhana: Too generic name keep it as `minioConfigTmpFileFormat`

cmd/admin-handlers.go
```diff
@@ -694,10 +696,121 @@ func (adminAPI adminAPIHandlers) GetConfigHandler(w http.ResponseWriter, r *http
...
 	}
 
 	writeSuccessResponseJSON(w, configBytes)
 }
+
+// toAdminAPIErrCode - converts errXLWriteQuorum error to admin API
+// specific error.
+func toAdminAPIErrCode(err error) APIErrorCode {
+	switch err {
+	case errXLWriteQuorum:
```
harshavardhana, line 712: is there no case of `errXLReadQuorum` ?  - perhaps adding a unit test would help.

cmd/admin-handlers.go
```diff
@@ -694,10 +696,121 @@ func (adminAPI adminAPIHandlers) GetConfigHandler(w http.ResponseWriter, r *http
...
+// operation.
+type nodeSummary struct {
+	Name string `json:"name"`
+	Err  string `json:"err"`
+}
+type setConfigResult struct {
+	NodeResults []nodeSummary `json:"nodeResults"`
+	Status      bool          `json:"status"`
+}
+
```
harshavardhana, line 728: A comment.

cmd/admin-handlers.go
```diff
@@ -694,10 +696,121 @@ func (adminAPI adminAPIHandlers) GetConfigHandler(w http.ResponseWriter, r *http
...
+func writeSetConfigResponse(w http.ResponseWriter, errs []error, status bool, reqURL *url.URL) {
+	var nodeResults []nodeSummary
+	for i := range errs {
```
harshavardhana: A comment for the `for loop`.

cmd/admin-handlers.go
```diff
@@ -694,10 +696,121 @@ func (adminAPI adminAPIHandlers) GetConfigHandler(w http.ResponseWriter, r *http
...
+		if errs[i] != nil {
+			nodeResults = append(nodeResults, nodeSummary{
+				Name: globalAdminPeers[i].addr,
+				Err:  fmt.Sprintf("%s", errs[i]),
+			})
+
+		} else {
+			nodeResults = append(nodeResults, nodeSummary{
+				Name: globalAdminPeers[i].addr,
+				Err:  "nil",
```
harshavardhana: You can collapse because `fmt.Sprintf("%s", errs[i])` if `errs[i] is `nil` it results in string being "\<nil\>" automatically.

cmd/admin-rpc-client.go
```diff
@@ -129,12 +145,92 @@ func (lc localAdminClient) GetConfig() ([]byte, error) {
...
+
+	tmpFile, err := os.Create(filepath.Join(configDir, tmpFileName))
+	if err != nil {
+		errorIf(err, "Failed to create temporary config file.")
+		return 0, err
+	}
+
+	defer tmpFile.Close()
+
+	config := bytes.NewReader(configBytes)
```
harshavardhana: May be just use `ioutil.WriteFile`  avoids all the clutter of `os.Create, io.Copy and bytes.NewReader`

cmd/admin-rpc-server.go
```diff
@@ -158,6 +162,84 @@ func (s *adminCmd) GetConfig(args *AuthRPCArgs, reply *ConfigReply) error {
...
+		errorIf(err, "Failed to get config path")
+		return err
+	}
+
+	// Create temporary config file at configDir/TmpFileName.
+	tmpFile, err := os.Create(filepath.Join(configDir, wArgs.TmpFileName))
+	if err != nil {
+		errorIf(err, "Failed to create temporary config file.")
+		return err
+	}
```
harshavardhana: Just use `ioutil.WriteFile` , also looks like a duplicate code to me.. `WriteTmpConfig()` can be re-purposed here.

cmd/admin-rpc-server.go
```diff
@@ -158,6 +162,84 @@ func (s *adminCmd) GetConfig(args *AuthRPCArgs, reply *ConfigReply) error {
...
+	defer tmpFile.Close()
+
+	reader := bytes.NewReader(wArgs.Buf)
+	n, err := io.Copy(tmpFile, reader)
+	if err != nil {
+		errorIf(err, "Failed to write temporary config.")
+		return err
+	}
+
+	wReply.N = int64(n)
```
harshavardhana: This is not useful in this situation to be returned back 'err' or no 'err' is fine. No one is consuming this information at the handler level right now. It is not useful if you use `ioutil.WriteFile` which will return error if it cannot write the entire buffer.

harshavardhana (changes requested):

cmd/admin-handlers.go
krisis: No. SetConfig is a write operation.

cmd/admin-handlers.go
krisis: This is what happens if errs[i] is nil - https://play.golang.org/p/5Zy1cuuSzL
```
{"Err":"\u003cnil\u003e"}
```

harshavardhana: Aah JSON you might have to tell that to not escape them.

krisis: This is possible if we used `%v` as the format specifier and use `json.Encoder` (and `SetEscapeHTML(false)`, with a caveat that we can't avoid the extra "\n" [`encoding/json`](https://golang.org/pkg/encoding/json/#Encoder.Encode) adds gratuitously. Would that be OK?

harshavardhana: Yeah that's OK.

cmd/admin-handlers.go
harshavardhana: oh is this just to setConfigHandler function? - because the function name is very generic `toAdminAPIErrCode(err)`

krisis: No. It is meant for all admin API handlers. It is currently used only by SetConfigHandler. As other uses appear, we can extend the function to return specific APIErrorCode values.

cmd/admin-handlers.go
```diff
@@ -694,10 +697,133 @@ func (adminAPI adminAPIHandlers) GetConfigHandler(w http.ResponseWriter, r *http
...
+		writeErrorResponse(w, toAPIErrorCode(jsonErr), reqURL)
+		return
+	}
+
+	if status {
+		writeSuccessResponseJSON(w, resultBuf.Bytes())
+	} else {
+		// Internal error signals to SDKs that this operation
+		// can be retried.
+		writeErrResponseJSON(w, http.StatusServiceUnavailable, resultBuf.Bytes())
```
vadmeste: I really prefer a clear separation between when we return xml response and json response. 500 error is already XML in S3 spec. Besides, since we are returning "status" in the json response, I think it is okay to return 200 status code.

harshavardhana: I see what you are saying is that it causes parsing problem.. on the `pkg/madm` package?  That can indeed cause confusion. I think then its okay to keep it 200 for now and revisit this later.

harshavardhana:
Some conflicts have occurred can you resolve them @krisis

harshavardhana (approved):

cmd/admin-handlers.go
vadmeste: > I see what you are saying is that it causes parsing problem.. on the pkg/madm package? That can indeed cause confusion. I think then its okay to keep it 200 for now and revisit this later.

:+1:

PR #3795, opened 2017-02-23
xl.PutObject: Ignore disks which return an error

vadmeste:
## Description
Ignore a disk which wasn't able to successfully perform an action to
avoid eventual perturbations when the disk comes back in the middle
of PutObject().

## Motivation and Context
Review of code and possible f.i.x of https://github.com/minio/minio/issues/3784 

## How Has This Been Tested?
go test

## Types of changes

## Checklist:

### DIFF
--- a/cmd/erasure-createfile.go
+++ b/cmd/erasure-createfile.go
@@ -114,6 +114,7 @@ func appendFile(disks []StorageAPI, volume, path string, enBlocks [][]byte, hash
 	// Write encoded data to quorum disks in parallel.
 	for index, disk := range disks {
 		if disk == nil {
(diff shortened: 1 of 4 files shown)


harshavardhana (commented):
This change looks good, can you also explore if there are any other places where this needs to be added?

Also one of the tests actually fail.

vadmeste:
> Read quorum never returns write quorum. we should use write quorum for all write operations the regardless if it does read in between .

yes my bad, I'll revert.

krishnasrinivas (approved):

harshavardhana (approved):
Can you add some tests for these scenarios?

PR #3798, opened 2017-02-24
event: use common initialization logic

balamurugana:
Previously creating and adding targets for each notification type was
repeated.  This patch fixes it.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/event-notifier.go
+++ b/cmd/event-notifier.go
@@ -537,6 +537,28 @@ func loadAllBucketNotifications(objAPI ObjectLayer) (map[string]*notificationCon
 	return nConfigs, lConfigs, nil
 }
 
+// addQueueTarget - calls newTargetFunc function and adds its returned value to queueTargets
+func addQueueTarget(queueTargets map[string]*logrus.Logger,
+	accountID, queueType string,
+	newTargetFunc func(string) (*logrus.Logger, error)) (string, error) {
+
+	// Construct the queue ARN for AMQP.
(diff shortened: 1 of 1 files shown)


cmd/event-notifier.go
```diff
@@ -537,6 +537,28 @@ func loadAllBucketNotifications(objAPI ObjectLayer) (map[string]*notificationCon
...
+
+	// Queue target if already initialized we move to the next ARN.
+	if _, ok := queueTargets[queueARN]; ok {
+		return queueARN, nil
+	}
+
+	// Using accountID we can now initialize a new AMQP logrus instance.
+	logger, err := newTargetFunc(accountID)
+	if err == nil {
+		queueTargets[queueARN] = logger
```
harshavardhana, line 556: Wouldn't it be easier if you provide this as method 

```
type queueTargets map[string]*logrus.Logger

func (qs *queueTargets) Set(accountID string) (err error)
func (qs *queueTargets) Del(accountID)
func (qs *queueTargets) Get(accountID) *logrus.Logger
```

IMHO this would make it easily readable.

balamurugana: This would make significant changes in current implementation.  Currently `Set()` and `Get()` equivalent code is used only in this function and no `Del()` anywhere.

harshavardhana: Okay agreed, this change looks good.

harshavardhana (approved):

PR #3803, opened 2017-02-24
cli: Use ADDRESS:PORT to clarify --address behavior

### LINKED ISSUE #3799: Allow binding to a hostname
The minio server provides a `--address` flag for binding to an IP, but it would be more convenient to be able to bind to a hostname. Could we add this ability to `--address`?

harshavardhana:
## Description
Currently we document as IP:PORT which doesn't provide
if someone can use HOSTNAME:PORT. This is a change
to clarify this by calling it as ADDRESS:PORT which
encompasses both a HOSTNAME and an IP.

## Motivation and Context
Fixes #3799

## How Has This Been Tested?
Manually UI.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/server-main.go
+++ b/cmd/server-main.go
@@ -36,7 +36,7 @@ var serverFlags = []cli.Flag{
 	cli.StringFlag{
 		Name:  "address",
 		Value: ":9000",
-		Usage: `Bind to a specific IP:PORT. Defaults to ":9000".`,
+		Usage: "Bind to a specific ADDRESS:PORT, ADDRESS can be an IP or hostname.",
 	},
 }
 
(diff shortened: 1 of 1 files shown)


cmd/server-main.go
```diff
@@ -36,7 +36,7 @@ var serverFlags = []cli.Flag{
 	cli.StringFlag{
 		Name:  "address",
 		Value: ":9000",
-		Usage: `Bind to a specific IP:PORT. Defaults to ":9000".`,
+		Usage: "Bind to a specific ADDRESS:PORT.",
```
krishnasrinivas: Can be:
Usage: "Bind to a specific ADDRESS:PORT. (ADDRESS can be IP or Hostname)"

krishnasrinivas (approved):

cmd/server-main.go
harshavardhana: :+1:

PR #3804, opened 2017-02-24
rpc: Remove time check for each RPC calls.

harshavardhana:
## Description
This removal comes to avoid some redundant requirements
which are adding more problems on a production setup.

Here are the list of checks for time as they happen

 - Fresh connect (during server startup) - CORRECT
 - A reconnect after network disconnect - CORRECT
 - For each RPC call - INCORRECT.

Verifying time for each RPC aggravates a situation
where a RPC call is rejected in a sequence of events
due to enough load on a production setup. 3 second
might not be enough time window for the call to be
initiated and received by the server.

## Motivation and Context
Removing unnecessary optimizations

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:

### DIFF
--- a/cmd/admin-rpc-server_test.go
+++ b/cmd/admin-rpc-server_test.go
@@ -53,7 +53,7 @@ func testAdminCmd(cmd cmdType, t *testing.T) {
 		<-globalServiceSignalCh
 	}()
 
-	ga := AuthRPCArgs{AuthToken: reply.AuthToken, RequestTime: time.Now().UTC()}
+	ga := AuthRPCArgs{AuthToken: reply.AuthToken}
 	genReply := AuthRPCReply{}
 	switch cmd {
 	case restartCmd:
@@ -108,8 +108,7 @@ func TestReInitDisks(t *testing.T) {
 	}
 
 	authArgs := AuthRPCArgs{
-		AuthToken:   reply.AuthToken,
-		RequestTime: time.Now().UTC(),
+		AuthToken: reply.AuthToken,
 	}
 	authReply := AuthRPCReply{}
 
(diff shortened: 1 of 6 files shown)


harshavardhana:
This is what @abperiasamy and I discussed makes sense for RPC calls.  We don't need to verify for each RPC call instead only verify either during startup Login or after rpc is disconnected/reconnect Login is sufficient.

Doing for each RPC calls invites more trouble than it solves under load and it was reproduced recently on David's cluster of 30TB with 100,000 files being uploaded.

krishnasrinivas:
can't the same thing be done with `AuthToken` as well?

harshavardhana:
> can't the same thing be done with AuthToken as well?

AuthToken as requirement is different it is part of the security and validity of the call itself coming from an rpc client.  JWT allows us to have a unique mechanism to validate if the request does indeed originate from an authenticated client which had a previous successful Login request. 

AuthToken is used like an ID to validate if the RPC call is coming from such a client which has correct credentials, correct time - all which were validated during the Login request.  This way a rouge RPC client cannot spoof the server. 

RequestTime on the other hand is not useful given that Token is already there, given that it is sufficient to validate it during connect/re-connect attempts.

cmd/typed-errors.go
```diff
@@ -49,4 +49,4 @@ var errServerNotInitialized = errors.New("Server not initialized, please try aga
 var errServerVersionMismatch = errors.New("Server versions do not match")
 
 // errServerTimeMismatch - server times are too far apart.
-var errServerTimeMismatch = errors.New("Server times are too far apart")
+var errServerTimeMismatch = errors.New("Server times are too far apart, please use http://www.ntp.org/ to synchronize the clocks")
```
balamurugana: I think its not a good idea about this suggestion in typed error.

balamurugana (commented):
Minor comment.   Rest of the changes are good to go.

cmd/typed-errors.go
harshavardhana: Sure.. will bring it in a separate change later.

balamurugana (approved):

PR #3808, opened 2017-02-27
XL: Make listOnlineDisks and outDatedDisks consistent w/ each other.

krisis:
## Description

- Checks if object parts match their checksum in `xl.json`
- Checks if all parts mentioned in `xl.json` are present.

## Motivation and Context

When disks in a distributed (Erasure coded) Minio setup have `xl.json` but have one or more parts missing or corrupt for a given object, we should heal them if there are enough 'good' parts.
## Types of changes

## Checklist:

### DIFF
--- a/cmd/xl-v1-healing-common.go
+++ b/cmd/xl-v1-healing-common.go
@@ -76,6 +76,26 @@ func listObjectModtimes(partsMetadata []xlMetaV1, errs []error) (modTimes []time
 	return modTimes
 }
 
+// Notes:
+// There are 5 possible states a disk could be in,
+// 1. __online__             - has the latest copy of xl.json - returned by listOnlineDisks
+//
+// 2. __offline__            - err == errDiskNotFound
+//
+// 3. __availableWithParts__ - has the latest copy of xl.json and has all
+//                             parts with checksums matching; returned by disksWithAllParts
+//
+// 4. __outdated__           - returned by outDatedDisk, provided []StorageAPI
+//                             returned by diskWithAllParts is passed for latestDisks.
+//    - has an old copy of xl.json
+//    - doesn't have xl.json (errFileNotFound)
+//    - has the latest xl.json but one or more parts are corrupt
+//
+// 5. __missingParts__       - has the latest copy of xl.json but has some parts
+// missing.  This is identified separately since this may need manual
+// inspection to understand the root cause. E.g, this could be due to
+// backend filesystem corruption.
+
 // listOnlineDisks - returns
 // - a slice of disks where disk having 'older' xl.json (or nothing)
 // are set to nil.
@@ -101,39 +121,22 @@ func listOnlineDisks(disks []StorageAPI, partsMetadata []xlMetaV1, errs []error)
 }
 
 // outDatedDisks - return disks which don't have the latest object (i.e xl.json).
-func outDatedDisks(disks, latestDisks []StorageAPI, partsMetadata []xlMetaV1,
+// disks that are offline are not 'marked' outdated.
+func outDatedDisks(disks, latestDisks []StorageAPI, errs []error, partsMetadata []xlMetaV1,
 	bucket, object string) (outDatedDisks []StorageAPI) {
 
 	outDatedDisks = make([]StorageAPI, len(disks))
-	for index, disk := range latestDisks {
-		// disk either has an older xl.json or is not online.
-		if disk == nil {
-			outDatedDisks[index] = disks[index]
+	for index, latestDisk := range latestDisks {
+		if latestDisk != nil {
 			continue
 		}
-
-		// disk has a valid xl.json but may not have all the
-		// parts. This is considered an outdated disk, since
-		// it needs healing too.
-		for pIndex, part := range partsMetadata[index].Parts {
-			// compute blake2b sum of part.
-			hash := newHash(blake2bAlgo)
-			blakeBytes, hErr := hashSum(disk, bucket, filepath.Join(object, part.Name), hash)
-			if hErr != nil {
-				outDatedDisks[index] = disk
-				break
-			}
-
-			blakeSum := hex.EncodeToString(blakeBytes)
-			// if blake2b sum doesn't match for a part
-			// then this disk is outdated and needs
-			// healing.
-			if blakeSum != partsMetadata[index].Erasure.Checksum[pIndex].Hash {
-				outDatedDisks[index] = disk
-				break
-			}
+		// disk either has an older xl.json or doesn't have one.
+		switch errorCause(errs[index]) {
+		case nil, errFileNotFound:
+			outDatedDisks[index] = disks[index]
 		}
 	}
+
 	return outDatedDisks
 }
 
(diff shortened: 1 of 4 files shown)


cmd/xl-v1-healing-common.go
```diff
@@ -59,29 +59,23 @@ func bootModtimes(diskCount int) []time.Time {
...
+		modTimes[index] = metadata.Stat.ModTime
 	}
 	return modTimes
 }
 
-// Returns slice of online disks needed.
-// - slice returing readable disks.
-// - modTime of the Object
+// listOnlineDisks - returns
+// - a slice of disks where disk having 'older' xl.json (or nothing)
```
vadmeste: you mean the most recent instead of older?

krisis: yeah, I meant to write the most recent. Thanks for catching this. I shall change it in the next round. 
This PR is not complete. I discovered that we don't check if parts present in online disks are corrupted (in the bitrot  sense) when we compute outdated disks. I am adding the necessary changes and the unit tests.

krisis: @vadmeste The comment is right as it stands. It says the returned []StorageAPI would have nil in place of disks where xl.json is either older or unavailable.

vadmeste: > @vadmeste The comment is right as it stands. It says the returned []StorageAPI would have nil in place of disks where xl.json is either older or unavailable.

yes, I understand now..

cmd/xl-v1-healing-common.go
```diff
@@ -97,11 +101,50 @@ func listOnlineDisks(disks []StorageAPI, partsMetadata []xlMetaV1, errs []error)
...
 		if disk == nil {
 			outDatedDisks[index] = disks[index]
+			continue
+		}
+
+		// disk has a valid xl.json but may not have all the
+		// parts. This is considered an outdated disk, since
+		// it needs healing too.
+		for pIndex, part := range partsMetadata[index].Parts {
+			if _, sErr := disk.StatFile(bucket,
```
harshavardhana: Just use hashSum function this whole thing collapses into one function.

harshavardhana (changes requested):

cmd/xl-v1-healing-common.go
krisis: Done.

cmd/xl-v1-healing-common.go
```diff
@@ -97,11 +101,50 @@ func listOnlineDisks(disks []StorageAPI, partsMetadata []xlMetaV1, errs []error)
...
+				filepath.Join(object, part.Name)); sErr != nil {
+				outDatedDisks[index] = disk
+				break
+			}
+
+			// Read the part from disk.
```
harshavardhana: is this needed prematurely? - is it to verify file itself is readable? we could do this with hashSum() itself serves the purpose of ReadAll()

cmd/xl-v1-healing-common.go
harshavardhana: is this needed prematurely? - is it to verify file itself is readable? we could do this with hashSum() itself serves the purpose of StatFile()

cmd/xl-v1-healing-common.go
krisis: This is not strictly required, using hashSum takes care of parts that weren't written fully too.

cmd/xl-v1-healing-common.go
krisis: Are you suggesting that hashSum would return `errFileNotFound` in the case of missing parts? Yes, that makes sense.

harshavardhana: yes it would. internally hashSum using disk.ReadFile() so it does return errFileNotFound.

harshavardhana:
ping @krishnasrinivas can you take a look at this change?

cmd/xl-v1-healing-common.go
krisis: Done.

cmd/xl-v1-healing-common.go
krisis: Done.

harshavardhana (approved):

cmd/xl-v1-healing-common.go
```diff
@@ -102,18 +96,10 @@ func listOnlineDisks(disks []StorageAPI, partsMetadata []xlMetaV1, errs []error)
 	return onlineDisks, modTime
 }
 
-// Return disks with the outdated or missing object.
-func outDatedDisks(disks []StorageAPI, partsMetadata []xlMetaV1, errs []error) (outDatedDisks []StorageAPI) {
+// outDatedDisks - return disks which don't have the latest object (i.e xl.json).
+func outDatedDisks(disks, latestDisks []StorageAPI, partsMetadata []xlMetaV1, errs []error) (outDatedDisks []StorageAPI) {
 	outDatedDisks = make([]StorageAPI, len(disks))
-	latestDisks, _ := listOnlineDisks(disks, partsMetadata, errs)
 	for index, disk := range latestDisks {
```
krishnasrinivas: errors other than errFileNotFound should not be considered as outdatedDisk, because if the error was errDiskNotFound erasureHealFile will try to write on it and fail - so healing will fail.

cmd/xl-v1-healing-common.go
```diff
@@ -97,11 +101,37 @@ func listOnlineDisks(disks []StorageAPI, partsMetadata []xlMetaV1, errs []error)
...
+		}
+
+		// disk has a valid xl.json but may not have all the
+		// parts. This is considered an outdated disk, since
+		// it needs healing too.
+		for pIndex, part := range partsMetadata[index].Parts {
+			// compute blake2b sum of part.
+			hash := newHash(blake2bAlgo)
+			blakeBytes, hErr := hashSum(disk, bucket, filepath.Join(object, part.Name), hash)
+			if hErr != nil {
```
krishnasrinivas: a temporary network outage might cause hashSum to return error for all latest disks and hence cause healObject() to remove objects from all the latest disks.
To keep healing code simple, on encountering any kind of unexpected error we can just bailout and return error, the client can send the heal request again.

cmd/xl-v1-healing-common.go
krishnasrinivas: errors other than errFileNotFound should not be considered as outdatedDisk, because if the error was errDiskNotFound erasureHealFile will try to write on it and fail - so healing will fail.

cmd/xl-v1-healing-common.go
krisis: OK, will do.

krisis:
@krishnasrinivas, to summarise,
1. Will change `outDatedDisks` to return disks that,
- Have no `xl.json` - `disk == nil && errs[index] == errFileNotFound`
- OR have an older copy of `xl.json` - `disk == nil && errs[index] == nil`,
where `disk` refers to online disks gotten from `listOnlineDisks` and `errs []error` shall be passed to `outDatedDisks` function in the updated PR.

2. Return error if blake2b checksum computation fails due to error returned by `hashSum`.
[EDIT] we can excuse `errFileNotFound` returned by `hashSum` - it indicates a missing part.

krishnasrinivas:
> Will change outDatedDisks to return disks that,
> Have no xl.json - disk == nil && errs[index] == errFileNotFound
> OR have an older copy of xl.json - disk == nil && errs[index] == nil,
> where disk refers to online disks gotten from listOnlineDisks and errs []error shall be passed to outDatedDisks function in the updated PR.

Yes, outDatedDisks =  missing objects and outdated objects

> we can excuse errFileNotFound returned by hashSum - it indicates a missing part.

Better not to excuse errFileNotFound too because it is a critical error that needs intervention and can be manually fixed and root cause should be found. Else it will hide the bug. We should fix it the way we fixed a recent bug in David Charbonneir's setup.

krisis:
>Better not to excuse errFileNotFound too because it is a critical error that needs intervention and can be manually fixed and root cause should be found. Else it will hide the bug. We should fix it the way we fixed a recent bug in David Charbonneir's setup.

IIUC, we won't be healing object if one of the parts is missing on a disk that had the latest xl.json, right? I am not aware of this bug you refer to. Could you point me to the github issue?

krisis:
@krishnasrinivas all the comments are addressed. It is ready for another round of review.

cmd/xl-v1-healing-common.go
```diff
@@ -59,29 +63,43 @@ func bootModtimes(diskCount int) []time.Time {
...
 	}
 	return modTimes
 }
 
-// Returns slice of online disks needed.
-// - slice returing readable disks.
-// - modTime of the Object
+// Notes:
+// There are 5 possible states a disk could be in,
+// 1. __online__             - has the latest copy of xl.json - returned by listOnlineDisks
```
vadmeste, line 81: What latest copy of xl.json does mean? quorum disks having the same xl.json ?

krisis: @vadmeste - `listOnlineDisks` returns disks that agree on a xl.json, which occurs maximally in the list of disks and has the latest `ModTime`.

harshavardhana (approved):

PR #3822, opened 2017-03-01
admin: Set Config returns errSet and errMsg

vadmeste:
## Description
There is no way to see if a node encountered an error
when trying to set a new config set, this commit adds
a bool errSet field.

## Motivation and Context
Implement mc set config 

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:

### DIFF
--- a/cmd/admin-handlers.go
+++ b/cmd/admin-handlers.go
@@ -718,9 +718,11 @@ func toAdminAPIErrCode(err error) APIErrorCode {
 // SetConfigResult - represents detailed results of a set-config
 // operation.
 type nodeSummary struct {
-	Name string `json:"name"`
-	Err  string `json:"err"`
+	Name   string `json:"name"`
+	ErrSet bool   `json:"errSet"`
+	ErrMsg string `json:"errMsg"`
 }
+
 type setConfigResult struct {
 	NodeResults []nodeSummary `json:"nodeResults"`
 	Status      bool          `json:"status"`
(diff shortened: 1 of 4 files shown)


krisis:
@vadmeste If set-config command succeeded the `Err` field would have `<nil>`, which is the string representation of `nil` in json. Do you need `ErrSet` in addition to this?

vadmeste:
> @vadmeste If set-config command succeeded the Err field would have <nil>, which is the string representation of nil in json. Do you need ErrSet in addition to this?

It is just easier for the developer to understand the API. ErrSet + ErrMsg more intuitive than Err type string

cmd/admin-handlers.go
```diff
@@ -732,9 +734,12 @@ func writeSetConfigResponse(w http.ResponseWriter, peers adminPeers, errs []erro
...
 	// set-config operation.
 	for i := range errs {
+		errSet := errs[i] != nil
+		errMsg := fmt.Sprintf("%v", errs[i])
 		nodeResults = append(nodeResults, nodeSummary{
-			Name: peers[i].addr,
-			Err:  fmt.Sprintf("%v", errs[i]),
+			Name:   peers[i].addr,
+			ErrSet: errSet,
+			ErrMsg: errMsg,
```
donatello: For the sake of brevity and readability (less context/local variables to remember), I would rather write the previous two lines as:

```go
   ErrSet: errs[i] != nil,
   ErrMsg: fmt.Sprintf("%v", errs[i]),
```

and remove the two earlier assignments ('+' lines 737 and 738).

vadmeste: Done

krisis (approved):

harshavardhana:
@vadmeste can you address the build issues?

vadmeste:
> @vadmeste can you address the build issues?

Done

harshavardhana (approved):

PR #3826, opened 2017-03-02
Remove globalMaxCacheSize and globalCacheExpiry variables

balamurugana:
This patch fixes below
* Remove global variables globalMaxCacheSize and globalCacheExpiry.
* Make global variables into constant in objcache package.

## How Has This Been Tested?

`make test`

## Types of changes

## Checklist:

### DIFF
--- a/pkg/objcache/objcache.go
+++ b/pkg/objcache/objcache.go
@@ -1,5 +1,5 @@
 /*
- * Minio Cloud Storage, (C) 2016 Minio, Inc.
+ * Minio Cloud Storage, (C) 2016, 2017 Minio, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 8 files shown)


pkg/objcache/objcache.go
```diff
@@ -27,18 +27,20 @@ import (
...
-// individual cache entry buffer size.
-var DefaultBufferRatio = uint64(10)
+	// defaultBufferRatio represents default ratio used to calculate the
+	// individual cache entry buffer size.
+	defaultBufferRatio = uint64(10)
 
-// DefaultGCPercent represents default garbage collection target percentage.
-var DefaultGCPercent = 20
+	// defaultGCPercent represents default garbage collection target percentage.
+	defaultGCPercent = 20
```
harshavardhana, line 42: It is kept public to be configurable as a generic package. Please keep it public.

pkg/objcache/objcache.go
```diff
@@ -27,18 +27,20 @@ import (
...
+	NoExpiry = time.Duration(0)
 
-// DefaultExpiry represents default time duration value when individual entries will be expired.
-var DefaultExpiry = time.Duration(72 * time.Hour) // 72hrs.
+	// DefaultExpiry represents three days time duration when individual entries will be expired.
+	DefaultExpiry = time.Duration(3 * 24 * time.Hour)
 
-// DefaultBufferRatio represents default ratio used to calculate the
-// individual cache entry buffer size.
-var DefaultBufferRatio = uint64(10)
```
harshavardhana, line 38: It is kept public to be configurable as a generic package. Please keep it public.

pkg/objcache/objcache.go
balamurugana: if defaultBufferRatio and defaultGCPercent are variables, when are those non-default values modified?

If they are tunable for `Cache`, then `New()` needs to accept as arguments.

harshavardhana: No not necessarily. It is a tunable which can be set during a init() of a program. So either add a method to make them configurable or keep it the way it is.

balamurugana: I see `New()` is using these values.  If any setter is called after `New()` there is no effect.
When are they getting used other than `New()`?

Problem with these globals are we should use like below
```go
cache1 := New(...)
// Modify globals
cache2 := New(...)
// Modify globals
cache3 := New(...)

```

This is not a good way of doing it.  Also it is problematic in concurrent programming.

harshavardhana: In go this is how you would have done it 

```
package main

func init() {
    objCache.DefaultBufferRatio = 5
    objCache.DefaultGCPercent = 10
}

func main() {
      cache1 := New()
}
```

You can bring the functionality back by adding two additional methods which allow this to be changed at runtime. It is useful for a cache because one needs to really make sure that GC kicks in quickly before running out of memory.

harshavardhana: For the time being i think this change is fine.. There are some other structural issues to address this by adding methods. This PR is good to merge.

harshavardhana (approved):

pkg/objcache/objcache.go
balamurugana: `init()` won't solve the problem of creating multiple caches with different properties at different places.

Either these tuneables at `New()` or changing them after `New()`.

Currently this package doesn't support changing them after `New()`.  If you want this feature, this should be a separate PR.

PR #3832, opened 2017-03-02
Fix check for bucket name validity

donatello:
## Description

The existing bucket name validity check is incorrect - it does not meet the spec at https://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html. 

## Motivation and Context

The rules state that:  

- Bucket names must be at least 3 and no more than 63 characters long.
- Bucket names must be a series of one or more labels. Adjacent labels are separated by a single period (.). Bucket names can contain lowercase letters, numbers, and hyphens. Each label must start and end with a lowercase letter or a number.
- Bucket names must not be formatted as an IP address (e.g., 192.168.5.4).

The existing validation allows bucket names such as `abc-.pqr` which are actually invalid.

## How Has This Been Tested?

New test cases have been added to address the corrected bucket naming rules.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/object-api-utils.go
+++ b/cmd/object-api-utils.go
@@ -21,7 +21,6 @@ import (
 	"fmt"
 	"io"
 	"path"
-	"regexp"
 	"runtime"
 	"strings"
 	"unicode/utf8"
@@ -38,11 +37,9 @@ const (
 	minioMetaMultipartBucket = minioMetaBucket + "/" + mpartMetaPrefix
 	// Minio Tmp meta prefix.
 	minioMetaTmpBucket = minioMetaBucket + "/tmp"
+	// DNS separator (period), used for bucket name validation.
+	dnsDelimiter = "."
 )
-
-// validBucket regexp.
-var validBucket = regexp.MustCompile(`^[a-z0-9][a-z0-9\.\-]{1,61}[a-z0-9]$`)
-var isIPAddress = regexp.MustCompile(`^(\d+\.){3}\d+$`)
 
 // isMinioBucket returns true if given bucket is a Minio internal
 // bucket and false otherwise.
@@ -52,10 +49,13 @@ func isMinioMetaBucketName(bucket string) bool {
 		bucket == minioMetaTmpBucket
 }
 
-// IsValidBucketName verifies a bucket name in accordance with Amazon's
-// requirements. It must be 3-63 characters long, can contain dashes
-// and periods, but must begin and end with a lowercase letter or a number.
-// See: http://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html
+// IsValidBucketName verifies that a bucket name is in accordance with
+// Amazon's requirements (i.e. DNS naming conventions). It must be 3-63
+// characters long, and it must be a sequence of one or more labels
+// separated by periods. Each label can contain lowercase ascii
+// letters, decimal digits and hyphens, but must not begin or end with
+// a hyphen. See:
+// http://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html
 func IsValidBucketName(bucket string) bool {
 	// Special case when bucket is equal to one of the meta buckets.
 	if isMinioMetaBucketName(bucket) {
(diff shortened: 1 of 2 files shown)


donatello:
This also needs to be fixed in SDKs.

balamurugana:
> The rules state that:
> 
>     Bucket names must be at least 3 and no more than 63 characters long.
>     Bucket names must be a series of one or more labels. Adjacent labels are separated by a single period (.). Bucket names can contain lowercase letters, numbers, and hyphens. Each label must start and end with a lowercase letter or a number.
>     Bucket names must not be formatted as an IP address (e.g., 192.168.5.4).
> 
> The existing validation allows bucket names such as abc-.pqr which are actually invalid.

Minio does not restrict bucket names other than the length whereas Amazon S3 comes with these restriction to work better for URL style endpoints like `bucket.s3.amazonaws.com`.  Minio does not support this ATM.

I think its OK to defer.

> This also needs to be fixed in SDKs.

:+1:  All Minio SDKs should support these restrictions.

cmd/object-api-utils.go
```diff
@@ -64,12 +69,13 @@ func IsValidBucketName(bucket string) bool {
...
-		return false
+	labels := strings.Split(bucket, ".")
+	for _, label := range labels {
+		if !validLabel.MatchString(label) {
+			return false
+		}
 	}
-	return (validBucket.MatchString(bucket) &&
-		!isIPAddress.MatchString(bucket) &&
-		!strings.Contains(bucket, ".."))
```
harshavardhana: We should move this check above the regex check to get some fair optimization - regex usage is memory heavy in Go - can be seen for all concurrent workloads.

cmd/object-api-utils.go
```diff
@@ -64,12 +69,13 @@ func IsValidBucketName(bucket string) bool {
 	if len(bucket) < 3 || len(bucket) > 63 {
 		return false
 	}
-	if bucket[0] == '.' || bucket[len(bucket)-1] == '.' {
```
harshavardhana, line 67: Keep this check as it is to avoid the regex checks..

harshavardhana (changes requested):

krisis:
> I think its OK to defer.

When users try to mirror objects from a Minio server to AWS S3, the operation would fail with invalid bucket or object name, as the case may be. I think this is an important reason to fix it now.

cmd/object-api-utils.go
krisis: For all valid bucket names this check would be false and we would be evaluating regex matches in all these cases. It isn't worth optimising for invalid bucket names.

cmd/object-api-utils_test.go
```diff
@@ -59,11 +61,14 @@ func TestIsValidBucketName(t *testing.T) {
...
-		{"THIS-BEINGS-WITH-UPPERCASe", false},
+		{"THIS-BEGINS-WITH-UPPERCASe", false},
 		{"tHIS-ENDS-WITH-UPPERCASE", false},
-		{"ThisBeginsAndEndsWithUpperCase", false},
+		{"ThisBeginsAndEndsWithUpperCasE", false},
 		{"una ñina", false},
-		{"lalalallalallalalalallalallalala-theString-size-is-greater-than-64", false},
+		{"dash-.may-not-appear-next-to-dot", false},
+		{"dash.-may-not-appear-next-to-dot", false},
+		{"dash-.-may-not-appear-next-to-dot", false},
```
krisis, line 70: shouldn't this test case be "dash.-may-not-..." so that the regex wouldn't fail before it checked for label ("-may-not-appear-next-to-dot") starting with a "-".

harshavardhana: > When users try to mirror objects from a Minio server to AWS S3, the operation would fail with invalid bucket or object name, as the case may be. I think this is an important reason to fix it now.

Yeah agreed we don't need to defer, since we defer for all other invalid bucket names.

cmd/object-api-utils.go
harshavardhana: The optimization is needed for the purpose i explained below, memory related issues with regexp.  One more thing we should do is avoid using string and use bytes instead since MatchString involves a memory copy inside. 

```
validBucket.Match([]byte(string))
```

krisis: The above condition would evaluate to false for all valid bucket names. Doesn't it perform regex matching in that case? So, the optimization applies only when the bucket name starts or ends with a '.', which is an invalid bucket name. I understand that we should avoid performing regex matches if we could, but we save nearly nothing in this case.

cmd/object-api-utils.go
```diff
@@ -40,8 +40,11 @@ const (
 	minioMetaTmpBucket = minioMetaBucket + "/tmp"
 )
 
-// validBucket regexp.
-var validBucket = regexp.MustCompile(`^[a-z0-9][a-z0-9\.\-]{1,61}[a-z0-9]$`)
+// IsValidBucketName related regexps. validLabel checks that a string
+// contains lowercase letters, decimal digits and hyphens, but does
+// not start or end with a hyphen. isIPAddress checks if a string
+// looks like an IP address.
+var validLabel = regexp.MustCompile(`^[a-z0-9]([a-z0-9\-]*[0-9a-z]){0,1}$`)
```
harshavardhana: Its not a label.. `isDnsCompatible` is perhaps better.

cmd/object-api-utils.go
```diff
@@ -64,12 +69,13 @@ func IsValidBucketName(bucket string) bool {
 	if len(bucket) < 3 || len(bucket) > 63 {
 		return false
 	}
-	if bucket[0] == '.' || bucket[len(bucket)-1] == '.' {
-		return false
+	labels := strings.Split(bucket, ".")
```
harshavardhana: Avoid saving into a variable since its not useful after this loop 

```
for _, label := range strings.Split(bucket, ".") {
}
```

Also lets keep `"."` as DNS delimiter constant to avoid mistakes.

cmd/object-api-utils.go
harshavardhana: What i am saying is we can do rune compare without going into regex, regex we can use it for composite checks which are harder to achieve simply by rune but in lieu of readability.  I think this PR's approach is fine.

harshavardhana (changes requested):

cmd/object-api-utils.go
donatello: Done.

cmd/object-api-utils.go
donatello: No regexps any more!

cmd/object-api-utils_test.go
donatello: The cases just enumerate a set of cases that are invalid (hyphen and dot adjacent in a few ways).

donatello:
@harshavardhana @krisis Please review again - now regular expression usage is removed.

harshavardhana (approved):

PR #3834, opened 2017-03-03
api: Increase the maximum object size limit from 5GiB to 16GiB.

### LINKED ISSUE #3825: make maxObjectSize a config option
maxObjectSize should be a setting

The maxObjectSize is an arbitrary value based on a strict equality with the s3 limits, in my case this limit is really annoying and force to bring to the last level of services the complexity of the authentification and the s3 API instead of having very simple PUT/GET on shared url.
To bypass this I have a proxy that take the PUT with a simple custom signature and send it to the storage with the S3 API.
This is useful 1 time every 100 000 PUT, one object is between 5Go and 10Go every 100 000, the rest are 99% < 100Mo et 1% is between 100Mo and 5Go.
I have no intention to switch to S3, even if the minio compatibility is interesting it's only on one side, most of the software that can use S3 as a backend can use minio or require minimal changes but I personally don't care about the opposite.
Having a warning saying that if you change that for a <5Go limit you break the possibility to switch from s3 to minio and if  > 5Go from minio to S3 is enough IMO.

harshavardhana:
## Description
The globalMaxObjectSize limit is instilled in S3 spec perhaps
due to certain limitations on S3 infrastructure. For minio we
don't have such limitations and we can stream a larger file
instead.

So we are going to bump this limit to 16GiB.

## Motivation and Context

Fixes #3825

## How Has This Been Tested?
Manually locally

## Types of changes

## Checklist:

### DIFF
--- a/cmd/admin-handlers_test.go
+++ b/cmd/admin-handlers_test.go
@@ -1239,16 +1239,16 @@ func TestWriteSetConfigResponse(t *testing.T) {
 	}
 
 	testPeers := []adminPeer{
-		adminPeer{
+		{
 			addr: "localhost:9001",
 		},
-		adminPeer{
+		{
 			addr: "localhost:9002",
 		},
-		adminPeer{
+		{
 			addr: "localhost:9003",
 		},
-		adminPeer{
+		{
 			addr: "localhost:9004",
 		},
 	}
--- a/cmd/bucket-handlers.go
+++ b/cmd/bucket-handlers.go
@@ -486,7 +486,7 @@ func (api objectAPIHandlers) PostPolicyBucketHandler(w http.ResponseWriter, r *h
 			return
 		}
 
-		if fileSize > lengthRange.Max || fileSize > maxObjectSize {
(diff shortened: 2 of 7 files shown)


dcharbonnier:
https://github.com/minio/minio/blob/master/docs/minio-limitations.md require an update too

harshavardhana:
> https://github.com/minio/minio/blob/master/docs/minio-limitations.md require an update too

We decided not to update that, we expect client SDKs to still keep S3 spec behavior and this document should just reflect that as well. 

If needed for special scenarios it allows beyond 5GB.

krisis (commented):
Code changes LGTM. curl + presignedURL put worked locally. Checking if copyObjectPart works with the increased maxObjectSize.

harshavardhana:
> Code changes LGTM. curl + presignedURL put worked locally. Checking if copyObjectPart works with the increased maxObjectSize.

Test copyObjectPart after the fixes in #3838

donatello (approved):

PR #3838, opened 2017-03-03
api: CopyObjectPart was copying wrong offsets due to shadowing.

### LINKED ISSUE #2205: registry:2 corrupts uploaded layers (filesystem layer verification failed) when using 3rd-party S3
Edit 2: It looks like an issue of handling multipart offset on minio's part. The developers are actively trying to resolve the issue, and they proposed a fix (pending validation).

---

Edit 1: It is possible that 63468ef4a85fabb756e799c2534f0df2f3c8167c introduces this regression

---

Numerous similar issues: #2168, docker/docker#21290, etc

Running on Kubernetes cluster:
```
Client Version: version.Info{Major:"1", Minor:"5", GitVersion:"v1.5.1", GitCommit:"82450d03cb057bab0950214ef122b67c83fb11df", GitTreeState:"clean", BuildDate:"2016-12-14T00:57:05Z", GoVersion:"go1.7.4", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"5", GitVersion:"v1.5.1+coreos.0", GitCommit:"cc65f5321f9230bf9a3fa171155c1213d6e3480e", GitTreeState:"clean", BuildDate:"2016-12-14T04:08:28Z", GoVersion:"go1.7.4", Compiler:"gc", Platform:"linux/amd64"}
```

Registry deployment file:
```json
{
    "apiVersion": "extensions/v1beta1",
    "kind": "Deployment",
    "metadata": {
        "name": "kube-registry",
        "namespace": "kube-system"
    },
    "spec": {
        "replicas": 3,
        "selector": {
            "matchLabels": {
                "app": "kube-registry"
            }
        },
        "template": {
            "metadata": {
                "labels": {
                    "app": "kube-registry"
                }
            },
            "spec": {
                "volumes": [{
                    "name": "auth-signing",
                    "secret": {
                        "secretName": "auth-signing"
                    }
                }],
                "containers": [{
                    "name": "sysctl-buddy",
                    "image": "alpine:latest",
                    "command": [
                        "/bin/sh",
                        "-c",
                        "while true; do\n  sysctl -w net.core.somaxconn=32768 > /dev/null 2>&1\n  sleep 10\ndone\n"
                    ],
                    "imagePullPolicy": "IfNotPresent",
                    "securityContext": {
                        "privileged": true
                    }
                }, {
                    "name": "registry",
                    "image": "registry:2",
                    "resources": {
                        "requests": {
                            "cpu": "200m",
                            "memory": "1024Mi"
                        },
                        "limits": {
                            "cpu": "2",
                            "memory": "2048Mi"
                        }
                    },
                    "env": [{
                            "name": "REGISTRY_LOG_LEVEL",
                            "value": "warn"
                        },
                        {
                            "name": "REGISTRY_HTTP_ADDR",
                            "value": ":5000"
                        },
                        {
                            "name": "REGISTRY_HTTP_HOST",
                            "value": "[redacted]"
                        },
                        {
                            "name": "REGISTRY_HTTP_SECRET",
                            "value": "[redacted]"
                        },
                        {
                            "name": "REGISTRY_STORAGE",
                            "value": "s3"
                        },
                        {
                            "name": "REGISTRY_STORAGE_DELETE_ENABLED",
                            "value": "true"
                        },
                        {
                            "name": "REGISTRY_STORAGE_S3_REGION",
                            "value": "us-east-1"
                        },
                        {
                            "name": "REGISTRY_STORAGE_S3_REGIONENDPOINT",
                            "value": "[redacted]"
                        },
                        {
                            "name": "REGISTRY_STORAGE_S3_BUCKET",
                            "value": "images"
                        },
                        {
                            "name": "REGISTRY_STORAGE_S3_ACCESSKEY",
                            "value": "docker"
                        },
                        {
                            "name": "REGISTRY_STORAGE_S3_SECRETKEY",
                            "value": "[redacted]"
                        },
                        {
                            "name": "REGISTRY_STORAGE_S3_ENCRYPT",
                            "value": "false"
                        },
                        {
                            "name": "REGISTRY_STORAGE_S3_SECURE",
                            "value": "false"
                        },
                        {
                            "name": "REGISTRY_STORAGE_S3_CHUNKSIZE",
                            "value": "20971520"
                        },
                        {
                            "name": "REGISTRY_STORAGE_CACHE_BLOBDESCRIPTOR",
                            "value": "redis"
                        },
                        {
                            "name": "REGISTRY_REDIS_ADDR",
                            "value": "docker-redis-cache:6379"
                        },
                        {
                            "name": "REGISTRY_AUTH",
                            "value": "token"
                        },
                        {
                            "name": "REGISTRY_AUTH_TOKEN_REALM",
                            "value": "[redacted]"
                        },
                        {
                            "name": "REGISTRY_AUTH_TOKEN_SERVICE",
                            "value": "[redacted]"
                        },
                        {
                            "name": "REGISTRY_AUTH_TOKEN_ISSUER",
                            "value": "[redacted]"
                        },
                        {
                            "name": "REGISTRY_AUTH_TOKEN_ROOTCERTBUNDLE",
                            "value": "/auth/auth.crt"
                        }
                    ],
                    "ports": [{
                        "containerPort": 5000,
                        "name": "registry",
                        "protocol": "TCP"
                    }],
                    "volumeMounts": [{
                        "name": "auth-signing",
                        "mountPath": "/auth",
                        "readOnly": true
                    }]
                }]
            }
        }
    }
}
```

Registry running with a 3rd-party S3 backend (minio).

---

Findings: `registry:2` breaks any layers larger than 100MB (`docker push` is fine, however `docker pull` gives error).

However, changing the image to `registry:2.5.1` or `registry:2.5`, and no more corruptions happened. Something must have changed between 2.5.x and 2.6.

harshavardhana:
## Description
startOffset was re-assigned to '0' so it would end up
copying wrong content ignoring the requested startOffset.

This also fixes the corruption issue we observed with docker registry.

## Motivation and Context
Fixes https://github.com/docker/distribution/issues/2205

## How Has This Been Tested?
With `go test` and docker registry. 

## Types of changes

## Checklist:

### DIFF
--- a/cmd/copy-part-range.go
+++ b/cmd/copy-part-range.go
@@ -1,0 +1,106 @@
+/*
+ * Minio Cloud Storage, (C) 2017 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cmd
+
+import (
+	"fmt"
+	"net/http"
+	"net/url"
+	"strconv"
+	"strings"
+)
+
+// Writes S3 compatible copy part range error.
+func writeCopyPartErr(w http.ResponseWriter, err error, url *url.URL) {
+	switch err {
+	case errInvalidRange:
+		writeErrorResponse(w, ErrInvalidCopyPartRange, url)
+		return
+	case errInvalidRangeSource:
+		writeErrorResponse(w, ErrInvalidCopyPartRangeSource, url)
+		return
+	default:
+		writeErrorResponse(w, ErrInternalError, url)
+		return
+	}
+}
+
+// Parses x-amz-copy-source-range for CopyObjectPart API. Specifically written to
+// differentiate the behavior between regular httpRange header v/s x-amz-copy-source-range.
+// The range of bytes to copy from the source object. The range value must use the form
+// bytes=first-last, where the first and last are the zero-based byte offsets to copy.
+// For example, bytes=0-9 indicates that you want to copy the first ten bytes of the source.
+// http://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadUploadPartCopy.html
+func parseCopyPartRange(rangeString string, resourceSize int64) (hrange *httpRange, err error) {
+	// Return error if given range string doesn't start with byte range prefix.
+	if !strings.HasPrefix(rangeString, byteRangePrefix) {
+		return nil, fmt.Errorf("'%s' does not start with '%s'", rangeString, byteRangePrefix)
+	}
+
+	// Trim byte range prefix.
+	byteRangeString := strings.TrimPrefix(rangeString, byteRangePrefix)
+
+	// Check if range string contains delimiter '-', else return error. eg. "bytes=8"
+	sepIndex := strings.Index(byteRangeString, "-")
+	if sepIndex == -1 {
+		return nil, errInvalidRange
+	}
+
+	offsetBeginString := byteRangeString[:sepIndex]
+	offsetBegin := int64(-1)
+	// Convert offsetBeginString only if its not empty.
+	if len(offsetBeginString) > 0 {
+		if !validBytePos.MatchString(offsetBeginString) {
+			return nil, errInvalidRange
+		}
+		if offsetBegin, err = strconv.ParseInt(offsetBeginString, 10, 64); err != nil {
+			return nil, errInvalidRange
+		}
+	}
+
+	offsetEndString := byteRangeString[sepIndex+1:]
+	offsetEnd := int64(-1)
+	// Convert offsetEndString only if its not empty.
+	if len(offsetEndString) > 0 {
+		if !validBytePos.MatchString(offsetEndString) {
+			return nil, errInvalidRange
+		}
+		if offsetEnd, err = strconv.ParseInt(offsetEndString, 10, 64); err != nil {
+			return nil, errInvalidRange
+		}
+	}
+
+	// rangeString contains first byte positions. eg. "bytes=2-" or
+	// rangeString contains last bye positions. eg. "bytes=-2"
+	if offsetBegin == -1 || offsetEnd == -1 {
+		return nil, errInvalidRange
+	}
+
+	// Last byte position should not be greater than first byte
+	// position. eg. "bytes=5-2"
+	if offsetBegin > offsetEnd {
+		return nil, errInvalidRange
+	}
+
+	// First and last byte positions should not be >= resourceSize.
+	if offsetBegin >= resourceSize || offsetEnd >= resourceSize {
+		return nil, errInvalidRangeSource
+	}
+
+	// Success..
+	return &httpRange{offsetBegin, offsetEnd, resourceSize}, nil
+}
--- a/cmd/fs-v1-multipart.go
+++ b/cmd/fs-v1-multipart.go
@@ -463,7 +463,6 @@ func (fs fsObjects) CopyObjectPart(srcBucket, srcObject, dstBucket, dstObject, u
 	pipeReader, pipeWriter := io.Pipe()
 
 	go func() {
-		var startOffset int64 // Read the whole file.
 		if gerr := fs.GetObject(srcBucket, srcObject, startOffset, length, pipeWriter); gerr != nil {
 			errorIf(gerr, "Unable to read %s/%s.", srcBucket, srcObject)
 			pipeWriter.CloseWithError(gerr)
--- a/cmd/object-handlers.go
+++ b/cmd/object-handlers.go
@@ -601,16 +601,12 @@ func (api objectAPIHandlers) CopyObjectPartHandler(w http.ResponseWriter, r *htt
 	var hrange *httpRange
 	rangeHeader := r.Header.Get("x-amz-copy-source-range")
 	if rangeHeader != "" {
-		if hrange, err = parseRequestRange(rangeHeader, objInfo.Size); err != nil {
+		if hrange, err = parseCopyPartRange(rangeHeader, objInfo.Size); err != nil {
 			// Handle only errInvalidRange
 			// Ignore other parse error and treat it as regular Get request like Amazon S3.
-			if err == errInvalidRange {
-				writeErrorResponse(w, ErrInvalidRange, r.URL)
-				return
-			}
-
-			// log the error.
-			errorIf(err, "Invalid request range")
+			errorIf(err, "Unable to extract range %s", rangeHeader)
+			writeCopyPartErr(w, err, r.URL)
+			return
 		}
 	}
 
@@ -623,8 +619,8 @@ func (api objectAPIHandlers) CopyObjectPartHandler(w http.ResponseWriter, r *htt
 	var startOffset int64
 	length := objInfo.Size
 	if hrange != nil {
-		startOffset = hrange.offsetBegin
 		length = hrange.getLength()
+		startOffset = hrange.offsetBegin
 	}
 
 	/// maximum copy size for multipart objects in a single operation
@@ -658,6 +654,12 @@ func (api objectAPIHandlers) PutObjectPartHandler(w http.ResponseWriter, r *http
 	objectAPI := api.ObjectAPI()
 	if objectAPI == nil {
 		writeErrorResponse(w, ErrServerNotInitialized, r.URL)
+		return
+	}
+
+	// X-Amz-Copy-Source shouldn't be set for this call.
+	if _, ok := r.Header["X-Amz-Copy-Source"]; ok {
+		writeErrorResponse(w, ErrInvalidCopySource, r.URL)
 		return
 	}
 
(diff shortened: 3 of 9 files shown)


cmd/object-handlers_test.go
```diff
@@ -939,6 +940,125 @@ func testAPIPutObjectHandler(obj ObjectLayer, instanceType, bucketName string, a
...
+			t.Fatalf("Test failed to create HTTP request for copy object part: <ERROR> %v", err)
+		}
+
+		// "X-Amz-Copy-Source" header contains the information about the source bucket and the object to copied.
+		req.Header.Set("X-Amz-Copy-Source", url.QueryEscape(pathJoin(bucketName, objectName)))
+		var byteRange string
+		if b > 0 {
+			byteRange = fmt.Sprintf("bytes=%d-%d", a, b)
+		} else {
+			byteRange = fmt.Sprintf("bytes=%d-", a)
```
donatello: This form is not actually supported by the AWS API - see https://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadUploadPartCopy.html:

> The range of bytes to copy from the source object. The range value must use the form bytes=first-last, where the first and last are the zero-based byte offsets to copy. For example, bytes=0-9 indicates that you want to copy the first ten bytes of the source.

As a consequence it appears that the S3 API would not allow copying a 0-byte object, but if we do this, Minio would allow it.

Is it a conscious decision?

donatello (commented):
LGTM code-wise (but have a comment). Will do some testing as well.

cmd/object-handlers_test.go
harshavardhana: Well i wouldn't say conscious - it was assumption that this header followed the RFC for httpRanges. Supporting full http ranges is not a problem for us, but we can stick to S3 behavior here. Will update the PR.

donatello (approved):
Confirm that this fixes the issue - this program https://gist.github.com/donatello/2579c2b1116214ebfcc6e6c4e368845e using minio-hs master branch says that server-side copy is corrected after the fix (initially it fails).

cmd/fs-v1-multipart.go
```diff
@@ -463,7 +463,6 @@ func (fs fsObjects) CopyObjectPart(srcBucket, srcObject, dstBucket, dstObject, u
 	pipeReader, pipeWriter := io.Pipe()
 
 	go func() {
-		var startOffset int64 // Read the whole file.
```
vadmeste, line 466: `vet shadow` wasn't saying anything here? :)

harshavardhana: Yes, that's a `go vet` bug.

cmd/copy-part-range.go
```diff
@@ -0,0 +1,90 @@
...
+		if offsetBegin, err = strconv.ParseInt(offsetBeginString, 10, 64); err != nil {
+			return nil, errInvalidRange
+		}
+	}
+
+	offsetEndString := byteRangeString[sepIndex+1:]
+	offsetEnd := int64(-1)
+	// Convert offsetEndString only if its not empty.
+	if len(offsetEndString) > 0 {
+		if !validBytePos.MatchString(offsetEndString) {
```
donatello, line 79: Seems unnecessary - validBytePos is just:

```go
var validBytePos = regexp.MustCompile(`^[0-9]+$`)
```

which is also checked by ParseInt immediately below.

cmd/copy-part-range.go
```diff
@@ -0,0 +1,90 @@
...
+	sepIndex := strings.Index(byteRangeString, "-")
+	if sepIndex == -1 {
+		return nil, errInvalidRange
+	}
+
+	offsetBeginString := byteRangeString[:sepIndex]
+	offsetBegin := int64(-1)
+	// Convert offsetBeginString only if its not empty.
+	if len(offsetBeginString) > 0 {
+		if !validBytePos.MatchString(offsetBeginString) {
```
donatello, line 67: Same as for next validBytePos comment below.

cmd/copy-part-range.go
```diff
@@ -0,0 +1,90 @@
...
+	// Return error if given range string doesn't start with byte range prefix.
+	if !strings.HasPrefix(rangeString, byteRangePrefix) {
+		return nil, fmt.Errorf("'%s' does not start with '%s'", rangeString, byteRangePrefix)
+	}
+
+	// Trim byte range prefix.
+	byteRangeString := strings.TrimPrefix(rangeString, byteRangePrefix)
+
+	// Check if range string contains delimiter '-', else return error. eg. "bytes=8"
+	sepIndex := strings.Index(byteRangeString, "-")
```
donatello, line 58: Better approach is to use `strings.Split(byteRangeString, "-")`

cmd/copy-part-range.go
```diff
@@ -0,0 +1,90 @@
...
+		return nil, errInvalidRange
+	}
+
+	if offsetBegin > offsetEnd {
+		// Last byte position is not greater than first byte position. eg. "bytes=5-2"
+		return nil, errInvalidRange
+	}
+
+	// First and last byte positions should not be >= resourceSize.
+	if offsetBegin >= resourceSize {
```
donatello: This if block is not needed - offsetBegin <= offsetEnd and offsetEnd < resourceSize implies the condition checked in this block. The previous if block and the next if block check these two conditions.

donatello (changes requested):
Code may be simplified.

cmd/copy-part-range.go
harshavardhana: Strings split actually doesn't cater for a bytes range where `bytes=8`

cmd/copy-part-range.go
harshavardhana: Not at all for a byte range like below `bytes=+2-5` you need that check.

cmd/copy-part-range.go
harshavardhana: Not at all for a byte range like below `bytes=+2-5` you need that check.

cmd/copy-part-range.go
harshavardhana: Done

cmd/copy-part-range.go
harshavardhana: Or even this `bytes=2-+5`

cmd/copy-part-range.go
harshavardhana: You have to know strings like '+5', '-2' are parsed properly by strconv.ParseInt as valid integer representations.

cmd/copy-part-range.go
```diff
@@ -0,0 +1,88 @@
...
+	"strconv"
+	"strings"
+)
+
+// Parses x-amz-copy-source-range for CopyObjectPart API. Specifically written to
+// differentiate the behavior between regular httpRange header v/s x-amz-copy-source-range.
+// The range of bytes to copy from the source object. The range value must use the form
+// bytes=first-last, where the first and last are the zero-based byte offsets to copy.
+// For example, bytes=0-9 indicates that you want to copy the first ten bytes of the source.
+func parseCopyPartRange(rangeString string, resourceSize int64) (hrange *httpRange, err error) {
```
vadmeste: Is there any link which specify this range header for S3 ?

harshavardhana: Yes there is, let me add..

cmd/object-handlers.go
```diff
@@ -601,11 +601,14 @@ func (api objectAPIHandlers) CopyObjectPartHandler(w http.ResponseWriter, r *htt
...
 	if rangeHeader != "" {
-		if hrange, err = parseRequestRange(rangeHeader, objInfo.Size); err != nil {
+		if hrange, err = parseCopyPartRange(rangeHeader, objInfo.Size); err != nil {
 			// Handle only errInvalidRange
 			// Ignore other parse error and treat it as regular Get request like Amazon S3.
 			if err == errInvalidRange {
-				writeErrorResponse(w, ErrInvalidRange, r.URL)
+				writeErrorResponse(w, ErrInvalidCopyPartRange, r.URL)
+				return
+			} else if err == errInvalidRangeSource {
```
vadmeste: For the sake of precaution, can you add an else which returns error unexpected ?

cmd/object-handlers.go
```diff
@@ -661,6 +664,12 @@ func (api objectAPIHandlers) PutObjectPartHandler(w http.ResponseWriter, r *http
 		return
 	}
 
+	// X-Amz-Copy-Source shouldn't be set for this call.
+	if _, ok := r.Header["X-Amz-Copy-Source"]; ok {
```
vadmeste, line 661: r.Header.Get() is case insensitive. I think it would be better to use it here.

harshavardhana: There reason to do that is "Get()" is not useful it would still return empty string for both cases when

`X-Amz-Copy-Source` is specified with empty string or `X-Amz-Copy-Source` is not specified. 

In this scenario we need to know if there is a header only not the value, so this is a correct style.

harshavardhana: Also all headers in Go are converted to upper-case.

cmd/copy-part-range.go
```diff
@@ -0,0 +1,106 @@
...
+	if len(offsetBeginString) > 0 {
+		if !validBytePos.MatchString(offsetBeginString) {
+			return nil, errInvalidRange
+		}
+		if offsetBegin, err = strconv.ParseInt(offsetBeginString, 10, 64); err != nil {
+			return nil, errInvalidRange
+		}
+	}
+
+	offsetEndString := byteRangeString[sepIndex+1:]
```
vadmeste, line 75: Doesn't this crash when the user sends "bytes=-" ?

vadmeste: > Doesn't this crash when the user sends "bytes=-" ?

It doesn't, I thought it will rise slice out of range error.

harshavardhana: Nope! @vadmeste it is already tested.. look at the tests.

vadmeste (approved):

PR #3855, opened 2017-03-06
Remove commands and commandsTree global variables.

balamurugana:
## How Has This Been Tested?

`make test`

## Types of changes

## Checklist:


main.go
```diff
@@ -65,5 +65,5 @@ func main() {
 		console.Fatalln("Go runtime version check failed.", err)
 	}
 
-	minio.Main(os.Args, os.Exit)
+	minio.Main(os.Args)
```
vadmeste, line 68: I know passing os.Exit looks odd here but we do that to be able to test it. Running os.Exit(0) while testing will stop all tests and return success status.

balamurugana: I couldn't able to get `os.Exit(0)` and related tests.  Are you talking about functional tests?

vadmeste: > Are you talking about functional tests?

Yes.

The goal is to be able to test minio.Main() function. If we let this latter calls os.Exit(), `go test` will just stop and report success though it should run all of our written tests.

balamurugana: I see `main.Main()` will be tested by `recover()` because `fatalIf()`, `fatalln()`  are used,  and `cli` package itself does `os.Exit()`.   Even if we pass a exit function, it won't help here.

harshavardhana: You can't test using recover() only panic() can be tested and panic() doesn't happen inside CLI niether does fatalIf calls panic()

cmd/main.go
```diff
@@ -120,89 +117,28 @@ func registerApp() *cli.App {
...
-	// Start profiler if env is set.
-	if profiler := os.Getenv("_MINIO_PROFILER"); profiler != "" {
-		globalProfiler = startProfiler(profiler)
-	}
+func Main(args []string) {
+	app := newApp()
 
 	// Run the app - exit on error.
 	if err := app.Run(args); err != nil {
-		exitFn(1)
```
harshavardhana: Keep the exitFn

cmd/main.go
```diff
@@ -61,53 +58,53 @@ VERSION:
...
+	// Collection of minio commands currently supported in a trie tree.
+	commandsTree := trie.NewTrie()
 
-func findClosestCommands(command string) []string {
-	var closestCommands []string
-	for _, value := range commandsTree.PrefixMatch(command) {
-		closestCommands = append(closestCommands, value.(string))
+	// registerCommand registers a cli command.
+	appendCommand := func(command cli.Command) {
+		commands = append(commands, command)
```
harshavardhana: The name and comment mismatch. Keep the name of the function as is I.e `registerCommand`

harshavardhana (changes requested):

cmd/main.go
balamurugana: Done

harshavardhana (approved):

PR #3863, opened 2017-03-07
Make unit testable cert parsing functions.

balamurugana:
## How Has This Been Tested?

`make test`

## Types of changes

## Checklist:


harshavardhana:
Ping @vadmeste can you review ?

cmd/config.go
```diff
@@ -46,13 +56,42 @@ func (config *ConfigDir) Get() string {
...
+	return filepath.Join(config.getCertsDir(), certsCADir)
+}
+
+// Create - creates configuration directory tree.
+func (config *ConfigDir) Create() error {
+	return mkdirAll(config.GetCADir(), 0700)
+}
+
+// GetMinioConfigFile - returns absolute path of config.json file.
+func (config *ConfigDir) GetMinioConfigFile() string {
```
vadmeste: I think GetMinioConfigPath() is more expressive than GetMinioConfigFile(), the same thing for GetPublicCertFile() and GetPrivateKeyFile()

cmd/server-main.go
```diff
@@ -209,10 +209,10 @@ func initServerConfig(c *cli.Context) {
 	minioInit(c)
 
 	// Create certs path.
-	fatalIf(createCertsPath(), "Unable to create \"certs\" directory.")
+	fatalIf(createConfigDir(), "Unable to create \"certs\" directory.")
```
vadmeste: I think it is better to move createConfigDir() inside minioInit() since config dir is already created there, it would make sense to put all creation process in one place

cmd/certs.go
```diff
@@ -19,120 +19,90 @@ package cmd
...
+		err = fmt.Errorf("Empty public certificate file %s", certFile)
+	}
 
-// isSSL - returns true with both cert and key exists.
-func isSSL() bool {
-	return isCertFileExists() && isKeyFileExists()
+	return certs, err
 }
 
 // Reads certificated file and returns a list of parsed certificates.
```
donatello: s/certificated/certificate/

donatello (approved):

cmd/config.go
balamurugana: `GetMinioConfigPath()` looks to me like path to minio configuration where some configuration files are stored whereas the function returning a absolute path of `config.json` file.

cmd/server-main.go

cmd/certs.go
balamurugana: Done

cmd/config.go
```diff
@@ -24,6 +24,16 @@ import (
 	"github.com/minio/mc/pkg/console"
 )
```
harshavardhana: This file name can be change to be more appropriate, `config.go` is means it holds `config` data structure - these are just fileutil functions for config.

cmd/config.go
```diff
@@ -24,6 +24,16 @@ import (
...
 
+const (
```
harshavardhana: Add a comment for the constants..

cmd/server-main.go
```diff
@@ -144,18 +144,20 @@ func initConfig() {
...
 // Generic Minio initialization to create/load config, prepare loggers, etc..
 func minioInit(ctx *cli.Context) {
-	// Is TLS configured?.
-	globalIsSSL = isSSL()
-
-	// Initialize minio server config.
-	initConfig()
-
 	// Enable all loggers by now so we can use errorIf() and fatalIf()
 	enableLoggers()
```
harshavardhana: How would one enable logging before initializing a config file? and what happens after the config has been initialized? - initConfig() also loads previously saved config - you would never set the logging properly with this approach.

harshavardhana (changes requested):

cmd/config.go
balamurugana: Done

cmd/config.go
balamurugana: I see this file contains configuration directory which contains configuration files and  subdirectories.  May be configdir.go?

cmd/server-main.go
balamurugana: There is a problem here.

Without initializing log, `errorIf()` and `fatalIf()` functions won't work.  There are functions which use those logger functions.

I will move back these statements like before and address the above problem later.

harshavardhana: No you should use `console.Fatalln()` until the logging is initialized.

cmd/config.go
harshavardhana: config-dir.go is fine @balamurugana

balamurugana: Done

cmd/server-main.go
balamurugana: Yes. But some those inter functions are used at run time also.  We can't log them into console later.  I can fix wherever possible in different PR.

harshavardhana (approved):

Issue #3759, opened 2017-02-18
Add support for healing missing `uploads.json` and `xl.json`


harshavardhana:
## Expected Behavior
Heal objects along with healing their incomplete upload parts.

## Current Behavior
Current healing object doesn't provide a way to heal corrupted or missing `uploads.json`

https://github.com/minio/minio/blob/master/cmd/xl-v1-healing.go#L291
```
 // TODO :-                                                                                                                                       
 // - add support for healing dangling `uploads.json`.                                                                                            
 // - add support for healing dangling `xl.json`.   
```

## Possible Solution
`quickHeal()` function should heal `uploads.json` since the incomplete uploads are pending transaction entires for a bucket. 

`xl.json` should be healed along with object healing.

## Steps to Reproduce (for bugs)
Quite easy just delete `uploads.json` from one the backends uploading fails sporadically.

## Context
Healing functionality is not complete without this.

## Your Environment
Current master

GarimaKapoor:
@kp are you working on this currently?

krisis:
@garima Yes I am working on this issue. Healing of an object when xl.json is missing in some disks is fixed. Healing of uploads remains to be fixed. It involves the following sub-tasks,
- Listing of uploads of a given bucket and prefix that needs to be healed.
- Healing of all uploads given the bucket and object.

harshavardhana:
Fixed by #3885 #3914

PR #3885, opened 2017-03-11
Implement list uploads heal admin API

krisis:
## Description

Add listing of ongoing multipart uploads that needs healing. This is similar to what we have for objects.

## Motivation and Context

See #3759. This fixes one part of the issue.

## How Has This Been Tested?

Manually and with unit tests where applicable.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/admin-handlers.go
+++ b/cmd/admin-handlers.go
@@ -36,16 +36,19 @@ const (
 // Type-safe query params.
 type mgmtQueryKey string
 
-// Only valid query params for list/clear locks management APIs.
+// Only valid query params for mgmt admin APIs.
 const (
-	mgmtBucket       mgmtQueryKey = "bucket"
-	mgmtObject       mgmtQueryKey = "object"
-	mgmtPrefix       mgmtQueryKey = "prefix"
-	mgmtLockDuration mgmtQueryKey = "duration"
-	mgmtDelimiter    mgmtQueryKey = "delimiter"
-	mgmtMarker       mgmtQueryKey = "marker"
-	mgmtMaxKey       mgmtQueryKey = "max-key"
-	mgmtDryRun       mgmtQueryKey = "dry-run"
+	mgmtBucket         mgmtQueryKey = "bucket"
+	mgmtObject         mgmtQueryKey = "object"
+	mgmtPrefix         mgmtQueryKey = "prefix"
+	mgmtLockDuration   mgmtQueryKey = "duration"
+	mgmtDelimiter      mgmtQueryKey = "delimiter"
+	mgmtMarker         mgmtQueryKey = "marker"
+	mgmtKeyMarker      mgmtQueryKey = "key-marker"
+	mgmtMaxKey         mgmtQueryKey = "max-key"
+	mgmtDryRun         mgmtQueryKey = "dry-run"
+	mgmtUploadIDMarker mgmtQueryKey = "upload-id-marker"
+	mgmtMaxUploads     mgmtQueryKey = "max-uploads"
 )
 
 // ServerVersion - server version
@@ -400,8 +403,57 @@ func (adminAPI adminAPIHandlers) ClearLocksHandler(w http.ResponseWriter, r *htt
 	writeSuccessResponseJSON(w, jsonBytes)
 }
 
-// validateHealQueryParams - Validates query params for heal list management API.
-func validateHealQueryParams(vars url.Values) (string, string, string, string, int, APIErrorCode) {
+// ListUploadsHealHandler - similar to listObjectsHealHandler
+// GET
+// /?heal&bucket=mybucket&prefix=myprefix&key-marker=mymarker&upload-id-marker=myuploadid&delimiter=mydelimiter&max-uploads=1000
+// - bucket is mandatory query parameter
+// - rest are optional query parameters List upto maxKey objects that
+// need healing in a given bucket matching the given prefix.
+func (adminAPI adminAPIHandlers) ListUploadsHealHandler(w http.ResponseWriter, r *http.Request) {
+	// Get object layer instance.
+	objLayer := newObjectLayerFn()
+	if objLayer == nil {
+		writeErrorResponse(w, ErrServerNotInitialized, r.URL)
+		return
+	}
+
+	// Validate request signature.
+	adminAPIErr := checkRequestAuthType(r, "", "", "")
+	if adminAPIErr != ErrNone {
+		writeErrorResponse(w, adminAPIErr, r.URL)
+		return
+	}
+
+	// Validate query params.
+	vars := r.URL.Query()
+	bucket := vars.Get(string(mgmtBucket))
+	prefix, keyMarker, uploadIDMarker, delimiter, maxUploads, _ := getBucketMultipartResources(r.URL.Query())
+
+	if err := checkListMultipartArgs(bucket, prefix, keyMarker, uploadIDMarker, delimiter, objLayer); err != nil {
+		writeErrorResponse(w, toAPIErrorCode(err), r.URL)
+		return
+	}
+
+	if maxUploads <= 0 || maxUploads > maxUploadsList {
+		writeErrorResponse(w, ErrInvalidMaxUploads, r.URL)
+		return
+	}
+
+	// Get the list objects to be healed.
+	listMultipartInfos, err := objLayer.ListUploadsHeal(bucket, prefix,
+		keyMarker, uploadIDMarker, delimiter, maxUploads)
+	if err != nil {
+		writeErrorResponse(w, toAPIErrorCode(err), r.URL)
+		return
+	}
+
+	listResponse := generateListMultipartUploadsResponse(bucket, listMultipartInfos)
+	// Write success response.
+	writeSuccessResponseXML(w, encodeResponse(listResponse))
+}
+
+// extractListObjectsHealQuery - Validates query params for heal objects list management API.
+func extractListObjectsHealQuery(vars url.Values) (string, string, string, string, int, APIErrorCode) {
 	bucket := vars.Get(string(mgmtBucket))
 	prefix := vars.Get(string(mgmtPrefix))
 	marker := vars.Get(string(mgmtMarker))
@@ -418,10 +470,13 @@ func validateHealQueryParams(vars url.Values) (string, string, string, string, i
 		return "", "", "", "", 0, ErrInvalidObjectName
 	}
 
-	// check if maxKey is a valid integer.
-	maxKey, err := strconv.Atoi(maxKeyStr)
-	if err != nil {
-		return "", "", "", "", 0, ErrInvalidMaxKeys
+	// check if maxKey is a valid integer, if present.
+	var maxKey int
+	var err error
+	if maxKeyStr != "" {
+		if maxKey, err = strconv.Atoi(maxKeyStr); err != nil {
+			return "", "", "", "", 0, ErrInvalidMaxKeys
+		}
 	}
 
 	// Validate prefix, marker, delimiter and maxKey.
@@ -454,7 +509,7 @@ func (adminAPI adminAPIHandlers) ListObjectsHealHandler(w http.ResponseWriter, r
 
 	// Validate query params.
 	vars := r.URL.Query()
-	bucket, prefix, marker, delimiter, maxKey, adminAPIErr := validateHealQueryParams(vars)
+	bucket, prefix, marker, delimiter, maxKey, adminAPIErr := extractListObjectsHealQuery(vars)
 	if adminAPIErr != ErrNone {
 		writeErrorResponse(w, adminAPIErr, r.URL)
 		return
--- a/cmd/xl-v1-list-objects-heal.go
+++ b/cmd/xl-v1-list-objects-heal.go
@@ -17,6 +17,7 @@
 package cmd
 
 import (
+	"path/filepath"
 	"sort"
 	"strings"
 )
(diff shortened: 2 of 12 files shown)


cmd/xl-v1-list-objects-heal.go
```diff
@@ -205,3 +206,252 @@ func (xl xlObjects) ListObjectsHeal(bucket, prefix, marker, delimiter string, ma
...
+	if delimiter == slashSeparator && prefix == slashSeparator {
+		return ListMultipartsInfo{}, nil
+	}
+
+	// If maxUploads is out of bounds reset it to maxUploadsList
+	if maxUploads < 0 || maxUploads > maxUploadsList {
+		maxUploads = maxUploadsList
+	}
+
+	// Initiate a list operation, if successful filter and return quickly.
```
donatello: What is meant by filtering after success here?

cmd/xl-v1-list-objects-heal.go
```diff
@@ -205,3 +206,252 @@ func (xl xlObjects) ListObjectsHeal(bucket, prefix, marker, delimiter string, ma
...
+			}
+			uploads, _, err = listMultipartUploadIDs(bucket, keyMarker,
+				uploadIDMarker, maxUploads, disk)
+			if err == nil {
+				break
+			}
+			if isErrIgnored(err, objMetadataOpIgnoredErrs...) {
+				continue
+			}
+			break
```
donatello: Can eliminate `break` by flipping the logical check to `!isErrIgnored(...)` in previous block.

donatello: This is even more concise, reducing an extra if block.
```go
if err == nil || !isErrIgnored(...) {
        break
}
```

cmd/xl-v1-list-objects-heal.go
```diff
@@ -205,3 +206,252 @@ func (xl xlObjects) ListObjectsHeal(bucket, prefix, marker, delimiter string, ma
...
+				}
+				newUploads, end, err = listMultipartUploadIDs(bucket, entry,
+					uploadIDMarker, maxUploads, disk)
+				if err == nil {
+					break
+				}
+				if isErrIgnored(err, objMetadataOpIgnoredErrs...) {
+					continue
+				}
+				break
```
donatello: Similar if block simplification like above can be done here.

cmd/xl-v1-list-objects-heal.go
```diff
@@ -205,3 +206,252 @@ func (xl xlObjects) ListObjectsHeal(bucket, prefix, marker, delimiter string, ma
...
+			}
+			entryLock.RUnlock()
+			if err != nil {
+				if isErrIgnored(err, xlTreeWalkIgnoredErrs...) {
+					continue
+				}
+				return ListMultipartsInfo{}, err
+			}
+			uploads = append(uploads, newUploads...)
+			maxUploads -= len(newUploads)
```
donatello: Looks like `maxUploads` can go negative here, putting in more entries in `uploads` than was requested.

cmd/xl-v1-list-objects-heal.go
```diff
@@ -205,3 +206,252 @@ func (xl xlObjects) ListObjectsHeal(bucket, prefix, marker, delimiter string, ma
...
+	// For delimiter and prefix as '/' we do not list anything at all
+	// since according to s3 spec we stop at the 'delimiter' along
+	// with the prefix. On a flat namespace with 'prefix' as '/'
+	// we don't have any entries, since all the keys are of form 'keyName/...'
+	if delimiter == slashSeparator && prefix == slashSeparator {
+		return ListMultipartsInfo{}, nil
+	}
+
+	// If maxUploads is out of bounds reset it to maxUploadsList
+	if maxUploads < 0 || maxUploads > maxUploadsList {
```
donatello: `maxUploads < 0` should be a user error - that they gave invalid input. There is no reason to give a negative value and it may be a bug in user code - so it makes sense to check for that.

cmd/xl-v1-list-objects-heal.go
```diff
@@ -205,3 +206,252 @@ func (xl xlObjects) ListObjectsHeal(bucket, prefix, marker, delimiter string, ma
...
+			if isErrIgnored(err, objMetadataOpIgnoredErrs...) {
+				continue
+			}
+			break
+		}
+		keyMarkerLock.RUnlock()
+		if err != nil {
+			return ListMultipartsInfo{}, err
+		}
+		maxUploads = maxUploads - len(uploads)
```
donatello: `maxUploads` can go negative here.

cmd/xl-v1-list-objects-heal.go
```diff
@@ -205,3 +206,252 @@ func (xl xlObjects) ListObjectsHeal(bucket, prefix, marker, delimiter string, ma
...
+			isLeaf := xl.isMultipartUpload
+			listDir := listDirFactory(isLeaf, xlTreeWalkIgnoredErrs,
+				xl.getLoadBalancedDisks()...)
+			walkerCh = startTreeWalk(minioMetaMultipartBucket,
+				multipartPrefixPath, multipartMarkerPath,
+				recursive, listDir, isLeaf, walkerDoneCh)
+		}
+		// Collect uploads until maxUploads limit is reached.
+		for maxUploads > 0 {
+			walkResult, ok := <-walkerCh
```
donatello: Looks like this for loop can just directly read off the channel like:

```go
for walkResult := range walkerCh {
//...
}
```

Simplifies the inner block quite a bit.

cmd/admin-handlers.go
```diff
@@ -400,8 +403,71 @@ func (adminAPI adminAPIHandlers) ClearLocksHandler(w http.ResponseWriter, r *htt
 	writeSuccessResponseJSON(w, jsonBytes)
 }
 
-// validateHealQueryParams - Validates query params for heal list management API.
-func validateHealQueryParams(vars url.Values) (string, string, string, string, int, APIErrorCode) {
+// ListUploadsHealHandler - similar to listObjectsHealHandler
+// ListUploadsHealHandler -
```
vadmeste: This is not needed.

cmd/admin-handlers.go
```diff
@@ -400,8 +403,71 @@ func (adminAPI adminAPIHandlers) ClearLocksHandler(w http.ResponseWriter, r *htt
...
+		return
+	}
+
+	listResponse := generateListMultipartUploadsResponse(bucket, listMultipartInfos)
+	// Write success response.
+	writeSuccessResponseXML(w, encodeResponse(listResponse))
+}
+
+// validateListObjectsHealQuery - Validates query params for heal objects list management API.
+func validateListObjectsHealQuery(vars url.Values) (string, string, string, string, int, APIErrorCode) {
```
vadmeste: extractListObjectsHealQuery is better than validateListObjectsHealQuery since it is also returning values.

pkg/madmin/heal-commands.go
```diff
@@ -432,3 +493,141 @@ func (adm *AdminClient) HealFormat(dryrun bool) error {
...
+					Err: err,
+				}
+				return
+			}
+
+			// If contents are available loop through and
+			// send over channel.
+			for _, upload := range result.Uploads {
+				// Save the marker.
+				marker = upload.Key
```
vadmeste: Not sure if I understand exactly why market is set here and also little below with result.keyMarker

pkg/madmin/heal-commands.go
```diff
@@ -432,3 +493,141 @@ func (adm *AdminClient) HealFormat(dryrun bool) error {
...
+	delimiter := "/"
+	if recursive {
+		// If recursive we do not delimit.
+		delimiter = ""
+	}
+
+	uploadIDMarker := ""
+
+	// Allocate new list objects channel.
+	uploadStatCh := make(chan UploadInfo)
```
vadmeste: I think it would be good if you set the capacity of this channel to 1000, so the code will fetch the next 1000 entries before the caller of this API finishes consuming the current 1000 entries. It will be faster like that.

pkg/madmin/heal-commands.go
krisis: `marker` is used in `listUploadsHeal` call to list the batch of uploads that need healing. I don't see a point in setting it here. We should only use the marker returned by the API.

pkg/madmin/heal-commands.go
krisis: Makes sense, will do.

cmd/xl-v1-list-objects-heal.go
krisis: `maxUploads` will not be negative. This is because we ask for at most `maxUploads` entries from `listMultipartUploadIDs`.

cmd/xl-v1-list-objects-heal.go
krisis: done.

cmd/xl-v1-list-objects-heal.go
krisis: done.

cmd/xl-v1-list-objects-heal.go
krisis: `listMultipartUploadIDs` is asked to return at most `maxUploads` entries. So, `maxUploads` won't become negative.

cmd/xl-v1-list-objects-heal.go
krisis: done.

cmd/xl-v1-list-objects-heal.go
krisis: This code snippet was 'inspired' from listObjectsHeal. This comment needs to be modified since it doesn't make sense after the modifications I did to the version.

cmd/xl-v1-list-objects-heal.go
krisis: done.

cmd/admin-handlers.go
krisis: done.

cmd/admin-handlers.go
krisis: done.

krisis (commented):
I have made the recommended changes. I will be submitting the updated PR after testing the changes locally.

cmd/xl-v1-list-objects-heal.go
```diff
@@ -205,3 +206,237 @@ func (xl xlObjects) ListObjectsHeal(bucket, prefix, marker, delimiter string, ma
...
+	}
+
+	switch {
+	// With max uploads of zero we have reached eof, return right here.
+	case maxUploads == 0:
+		return ListMultipartsInfo{}, nil
+	case maxUploads > maxUploadsList, maxUploads < 0:
+		return ListMultipartsInfo{}, InvalidMaxUpload{maxUploads}
+	}
+
```
donatello: Just noticed that `ListUploadsHealHandler` is the only caller to `xl.ListUploadsHeal` - it appears that this checking of `maxUploads` is repeated in both functions. Also, why does the case of `maxUploads == 0` have to be handled at all? Shouldn't it be a user error and shouldn't the API spec also disallow this kind of useless exchange with the server?

For example, AWS list-multipart-uploads API does not accept zero for max-uploads - https://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadListMPUpload.html

cmd/xl-v1-list-objects-heal.go
```diff
@@ -205,3 +206,237 @@ func (xl xlObjects) ListObjectsHeal(bucket, prefix, marker, delimiter string, ma
...
+		for _, disk := range xl.getLoadBalancedDisks() {
+			if disk == nil {
+				continue
+			}
+			uploads, _, err = listMultipartUploadIDs(bucket, keyMarker,
+				uploadIDMarker, maxUploads, disk)
+			if err != nil && isErrIgnored(err, objMetadataOpIgnoredErrs...) {
+				continue
+			}
+			break
```
donatello: Above (four lines) can be simplified as:

```go
			if err == nil || !isErrIgnored(err, objMetadataOpIgnoredErrs...) {
				break
			}
```

cmd/xl-v1-list-objects-heal.go
```diff
@@ -205,3 +206,237 @@ func (xl xlObjects) ListObjectsHeal(bucket, prefix, marker, delimiter string, ma
...
+			for _, disk := range xl.getLoadBalancedDisks() {
+				if disk == nil {
+					continue
+				}
+				newUploads, end, err = listMultipartUploadIDs(bucket, entry,
+					uploadIDMarker, maxUploads, disk)
+				if err != nil && isErrIgnored(err, objMetadataOpIgnoredErrs...) {
+					continue
+				}
+				break
```
donatello: Again (last four lines):

```go
			if err == nil || !isErrIgnored(err, objMetadataOpIgnoredErrs...) {
				break
			}
```

krisis: Done.

cmd/xl-v1-list-objects-heal.go
krisis: Done.

cmd/xl-v1-list-objects-heal.go
krisis: The check is moved to `admin-handlers.go` and is the only place we validate the query parameters. `maxUploads == 0` in the updated PR would return `InvalidMaxUploads` error.

donatello (approved):

harshavardhana (approved):

PR #3893, opened 2017-03-13
Simplify credential usage.

balamurugana:
## How Has This Been Tested?

`make test`

## Types of changes

## Checklist:

### DIFF
--- a/cmd/credential.go
+++ b/cmd/credential.go
@@ -1,5 +1,5 @@
 /*
- * Minio Cloud Storage, (C) 2015, 2016 Minio, Inc.
+ * Minio Cloud Storage, (C) 2015, 2016, 2017 Minio, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -20,7 +20,6 @@ import (
 	"crypto/rand"
 	"encoding/base64"
 	"errors"
-	"os"
 
 	"github.com/minio/mc/pkg/console"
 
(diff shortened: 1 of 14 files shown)


cmd/credential.go
```diff
@@ -75,65 +58,71 @@ type credential struct {
...
+	return isAccessKeyValid(cred.AccessKey) && isSecretKeyValid(cred.SecretKey)
 }
 
-// Initialize a new credential object
-func newCredential() credential {
-	return newCredentialWithKeys(mustGetAccessKey(), mustGetSecretKey())
-}
+// Equals - returns whether two credentials are equal or not.
+func (cred credential) Equal(ccred credential) bool {
+	if !ccred.IsValid() {
```
donatello: This check seems unnecessary in the Equal() function, because `createCredential` is already validating.

balamurugana: Passed `ccred` may be from empty or invalid due to JSON unmarshaling.  This check requires in this case.

krishnasrinivas: @balamurugana can you point the code flow? because `authenticateJWT` is calling `createCredential`

balamurugana: @krishnasrinivas I am unable to understand what code flow you are asking.

This method checks for equality of two credentials.  Those credentials could be created with different method than `createCredential()` or `mustGetNewCredential()`.  `credential.Equal()` prevents such error case by validating passed `ccred` is valid or not.

In short, if both `cred` and `ccred` are empty, this method makes equality check failure.

cmd/credential.go
```diff
@@ -75,65 +58,71 @@ type credential struct {
...
+		return false
+	}
 
-func newCredentialWithKeys(accessKey, secretKey string) credential {
-	secretHash := mustGetHashedSecretKey(secretKey)
-	return credential{accessKey, secretKey, secretHash}
-}
+	if cred.secretKeyHash == nil {
+		secretKeyHash, err := bcrypt.GenerateFromPassword([]byte(cred.SecretKey), bcrypt.DefaultCost)
+		if err != nil {
```
harshavardhana: Since this error is lost, it would be better we can log it.. to know why even bcrypt failed..

harshavardhana (approved):
One minor change..

harshavardhana:
ping @balamurugana - can you resolve the conflicts.. ?

cmd/credential.go
balamurugana: Done

harshavardhana (approved):

PR #3912, opened 2017-03-16
config: Relax browser and region to be empty.

harshavardhana:
## Description
- If browser field is missing or empty
  then default to "on".

- If region field is empty or missing then
  default to "us-east-1" (S3 spec behavior)

## Motivation and Context
Starting from slack message 

```
@y4m4b4 if browser field is missing in config.json, it should not error out. It should instead proceed with the default value.
```

## How Has This Been Tested?
Manually and using `go test`

## Types of changes

## Checklist:

### DIFF
--- a/cmd/server-main.go
+++ b/cmd/server-main.go
@@ -128,7 +128,7 @@ func initConfig() {
 			console.Fatalf("Invalid access/secret Key set in environment. Err: %s.\n", err)
 		}
 
-		// Envs are set globally.
+		// credential Envs are set globally.
 		globalIsEnvCreds = true
 	}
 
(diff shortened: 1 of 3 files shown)


cmd/server-main.go
```diff
@@ -138,7 +138,12 @@ func initConfig() {
 			console.Fatalf("Invalid value ‘%s’ in MINIO_BROWSER environment variable.", browser)
 		}
 
-		globalIsEnvBrowser = strings.EqualFold(browser, "off")
+		// browser Envs are set globally.
+		globalIsEnvBrowser = true
+	} else {
+		// browser envs are not through env,
+		// but by default browser is on nevertheless.
+		browser = "on"
```
vadmeste: Not clear why this is needed, besides, we are setting this `browser` variable to `envParams`, I think this could lead to a more confusion. envParams should only contain exact env variables.

harshavardhana: We can do without this change as well

vadmeste (approved):

PR #3914, opened 2017-03-16
Implement heal upload admin API.

krisis:
## Description

Add healing of ongoing multipart uploads. This is similar to what exists for objects.

## Motivation and Context

See #3759. This fixes the issue with #3885.

## How Has This Been Tested?

Manually and with unit tests where applicable.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/admin-handlers.go
+++ b/cmd/admin-handlers.go
@@ -24,6 +24,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/url"
+	"path"
 	"strconv"
 	"time"
 )
@@ -49,6 +50,7 @@ const (
 	mgmtDryRun         mgmtQueryKey = "dry-run"
 	mgmtUploadIDMarker mgmtQueryKey = "upload-id-marker"
 	mgmtMaxUploads     mgmtQueryKey = "max-uploads"
+	mgmtUploadID       mgmtQueryKey = "upload-id"
 )
 
 // ServerVersion - server version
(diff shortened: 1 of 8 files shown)


cmd/xl-v1-healing.go
```diff
@@ -477,10 +477,6 @@ func healObject(storageDisks []StorageAPI, bucket string, object string, quorum
 // and later the disk comes back up again, heal on the object
 // should delete it.
 func (xl xlObjects) HealObject(bucket, object string) error {
-	if err := checkGetObjArgs(bucket, object); err != nil {
```
harshavardhana, line 480: Any reason this was removed ?

krisis:
@harshavardhana, the check is moved to admin-handlers. I saw a fixme note elsewhere which said we should move these checks to object-handlers.go. So, now Healobject handler calls the check* family of functions before call this function.

cmd/xl-v1-healing.go
krishnasrinivas: probably because length can go beyond 1024 bytes?

harshavardhana: No actually @krisis already wrote it at a separate comment..

krishnasrinivas (approved):

cmd/admin-handlers.go
```diff
@@ -654,6 +656,67 @@ func (adminAPI adminAPIHandlers) HealObjectHandler(w http.ResponseWriter, r *htt
...
+	}
+
+	vars := r.URL.Query()
+	bucket := vars.Get(string(mgmtBucket))
+	object := vars.Get(string(mgmtObject))
+	uploadID := vars.Get(string(mgmtUploadID))
+	uploadObj := path.Join(bucket, object, uploadID)
+
+	// Validate bucket and object names.
+	if err := checkBucketAndObjectNames(minioMetaMultipartBucket,
```
krishnasrinivas: for validation, shouldn't the bucket be `bucket` and object be `object`

because if bucket is sent as `a` it does not fail with invalid bucket name.

krisis: >

for validation, shouldn't the bucket be bucket and object be object

because if bucket is sent as a it does not fail with invalid bucket name.

Makes sense. The reason I check for `uploadObjName` is to avoid failures later in healing of an upload. I will add additional checks for the bucket and object as passed via query parameters.

cmd/admin-handlers.go
```diff
@@ -654,6 +656,75 @@ func (adminAPI adminAPIHandlers) HealObjectHandler(w http.ResponseWriter, r *htt
...
+// HealUploadHandler - POST /?heal&bucket=mybucket&object=myobject&upload-id=myuploadID&dry-run
+// - x-minio-operation = upload
+// - bucket, object and upload-id are mandatory query parameters
+// Heal a given upload, if present.
+func (adminAPI adminAPIHandlers) HealUploadHandler(w http.ResponseWriter, r *http.Request) {
+	// Get object layer instance.
+	objLayer := newObjectLayerFn()
+	if objLayer == nil {
+		writeErrorResponse(w, ErrServerNotInitialized, r.URL)
+		return
```
harshavardhana, line 668: Can you add a test case of object layer not initialized?

cmd/admin-handlers.go
```diff
@@ -654,6 +656,75 @@ func (adminAPI adminAPIHandlers) HealObjectHandler(w http.ResponseWriter, r *htt
...
+	}
+
+	// Validate request signature.
+	adminAPIErr := checkRequestAuthType(r, "", "", "")
+	if adminAPIErr != ErrNone {
```
harshavardhana, line 673: Also for signature mismatch?

cmd/admin-handlers.go
```diff
@@ -654,6 +656,75 @@ func (adminAPI adminAPIHandlers) HealObjectHandler(w http.ResponseWriter, r *htt
...
+		writeSuccessResponseHeadersOnly(w)
+		return
+	}
+
+	//We are able to use HealObject for healing an upload since an
+	//ongoing upload has the same backend representation as an
+	//object.  The 'object' corresponding to a given bucket,
+	//object and uploadID is
+	//.minio.sys/multipart/bucket/object/uploadID.
+	err := objLayer.HealObject(minioMetaMultipartBucket, uploadObj)
```
harshavardhana, line 718: Can you capture the failure conditions in tests?

harshavardhana (changes requested):

cmd/admin-handlers.go
krisis: done.

cmd/admin-handlers.go
krisis: done.

cmd/admin-handlers.go
krisis: done, added as part of unit tests in `xl-v1-healing_test.go`.

harshavardhana (approved):

PR #3932, opened 2017-03-18
	heal: Set truncate when no more walk entries

vadmeste:
## Description
List Uploads Heal sets truncate to false when tree walk returns
file not found error for any reason, for example, in case of the
absence of the bucket directory inside .minio.sys/multipart

## Motivation and Context
truncate variable is not set when ListUploadsHeal is called and we don't have
the corresponding bucket directory inside multipart directory

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:

### DIFF
--- a/cmd/xl-v1-list-objects-heal.go
+++ b/cmd/xl-v1-list-objects-heal.go
@@ -303,6 +303,8 @@ func (xl xlObjects) listMultipartUploadsHeal(bucket, prefix, keyMarker,
 	var walkerDoneCh chan struct{}
 	// Check if we have room left to send more uploads.
 	if maxUploads > 0 {
+		uploadsLeft := maxUploads
+
 		walkerCh, walkerDoneCh = xl.listPool.Release(listParams{
 			bucket:    minioMetaMultipartBucket,
 			recursive: recursive,
@@ -319,10 +321,14 @@ func (xl xlObjects) listMultipartUploadsHeal(bucket, prefix, keyMarker,
 				multipartPrefixPath, multipartMarkerPath,
 				recursive, listDir, isLeaf, walkerDoneCh)
 		}
-		// Collect uploads until maxUploads limit is reached.
-		for walkResult := range walkerCh {
-			// For any error during tree walk we should
-			// return right away.
+		// Collect uploads until leftUploads limit is reached.
+		for {
+			walkResult, ok := <-walkerCh
+			if !ok {
+				truncated = false
+				break
+			}
+			// For any error during tree walk, we should return right away.
 			if walkResult.err != nil {
 				return ListMultipartsInfo{}, walkResult.err
 			}
@@ -334,8 +340,8 @@ func (xl xlObjects) listMultipartUploadsHeal(bucket, prefix, keyMarker,
 				uploads = append(uploads, uploadMetadata{
 					Object: entry,
 				})
-				maxUploads--
-				if maxUploads == 0 {
+				uploadsLeft--
+				if uploadsLeft == 0 {
 					break
 				}
 				continue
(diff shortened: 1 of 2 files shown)


cmd/xl-v1-list-objects-heal.go
```diff
@@ -325,15 +325,14 @@ func (xl xlObjects) listMultipartUploadsHeal(bucket, prefix, keyMarker,
...
-			// Ignore errors like errDiskNotFound
-			// and errFileNotFound.
-			if isErrIgnored(walkResult.err,
-				xlTreeWalkIgnoredErrs...) {
-				continue
-			}
-			// For any error during tree walk we should
-			// return right away.
+
+			// For any error during tree walk, except for file not found,
```
harshavardhana: Can you explain this with a test case ?

harshavardhana (changes requested):

cmd/xl-v1-list-objects-heal.go
vadmeste: > Can you explain this with a test case ?

Done, I've added some tests in admin-handlers_test.go: Test 1 and 2 succeed in this PR but not in the current master. isTruncated is supposed to return False but it is returning True.

krisis: @vadmeste `listDirFactory` shouldn't return an ignored error. `errFileNotFound` is an ignored error.  See PR https://github.com/minio/minio/pull/3935. With that being the case, we won't be able to check if `err == errFileNotFound`. The fix is to add the following snippet after the for loop over the `walkResultCh`. This checks if the for-loop was exited on iterating through all entries or on `maxUploads` becoming `0`.
```go
if maxUploads > 0 {
  truncated = false
}
```

vadmeste: > if maxUploads > 0 {
>   truncated = false
> }

I see, in that case, I would rename maxUploads to leftUploads, what do you think ?

krisis: :+1: I think `remainingUploads` or `uploadsLeft` reads nice.

vadmeste:
This PR doesn't work only after merging #3935.

@krisis, I made some changes, can you review again ?

krisis:
@vadmeste, the changes look good.

krishnasrinivas:
```
		// Mark as not truncated if we have no more uploads to return
		// but still have rooms of them.
		if uploadsLeft > 0 {
			truncated = false
		}
```

this would be incorrect way to decide if the listing is truncated, it should be based on if we have more entries to send to the client after returning, it's not based on if we were able to send the number of entries that were requested.

also i think this patch is not needed after https://github.com/minio/minio/pull/3935 because the reader of the channel will not get the error, if it does then we can return error to the client

vadmeste:
> this would be incorrect way to decide if the listing is truncated, it should be based on if we have more entries to send to the client after returning, it's not based on if we were able to send the number of entries that were requested.

Incorrect or could be confusing ?

> also i think this patch is not needed after #3935 because the reader of the channel will not get the error, if it does then we can return error to the client

No, we still need this patch, listDir() doesn't know when we should end because there is another function which fetches multipart entries.

krishnasrinivas:
>> this would be incorrect way to decide if the listing is truncated, it should be based on if we have more entries to send to the client after returning, it's not based on if we were able to send the number of entries that were requested.
> Incorrect or could be confusing ?

Incorrect. check `listObjectsHeal` on how IsTruncated is decided on, it's based on if we have more entries  after returning the call. The client would send the List call again with the marker till the IsTruncated is false. What am I missing?

>> also i think this patch is not needed after #3935 because the reader of the channel will not get the error, if it does then we can return error to the client
> No, we still need this patch, listDir() doesn't know when we should end because there is another function which fetches multipart entries.

I did not get you @vadmeste is there something different in listMultipartUploadsHeal behavior when compared to listObjects/listObjectsHeal because in those calls we decide the value of IsTruncated differently.

vadmeste:
> Incorrect. check listObjectsHeal on how IsTruncated is decided on, it's based on if we have more entries after returning the call. The client would send the List call again with the marker till the IsTruncated is false. What am I missing?

Just wanted to know if you are seeing a bug that leads to an incorrect behavior or it is just about a bad code.

I modified a little the code, now it sets truncated to false when we can't receive new entries, please review again.

krishnasrinivas:
> I modified a little the code, now it sets truncated to false when we can't receive new entries, please review again.

:+1:

krishnasrinivas (approved):

harshavardhana (approved):

PR #3935, opened 2017-03-20
XL: Don't return ignored errors in listDirFactory

krisis:
## Description

This change fixes `listDirFactory` used for listing in Erasure backend. Previously, it was possible that `listDirFactory` could return an error which was explicitly asked to be ignored. With this change, it will return `nil`.

## How Has This Been Tested?

Unit test

## Types of changes

## Checklist:

### DIFF
--- a/cmd/xl-v1-list-objects.go
+++ b/cmd/xl-v1-list-objects.go
@@ -27,19 +27,20 @@ func listDirFactory(isLeaf isLeafFunc, treeWalkIgnoredErrs []error, disks ...Sto
 				continue
 			}
 			entries, err = disk.ListDir(bucket, prefixDir)
-			if err == nil {
-				entries, delayIsLeaf = filterListEntries(bucket, prefixDir, entries, prefixEntry, isLeaf)
-				return entries, delayIsLeaf, nil
+			if err != nil {
+				// For any reason disk was deleted or goes offline, continue
+				// and list from other disks if possible.
+				if isErrIgnored(err, treeWalkIgnoredErrs...) {
+					continue
+				}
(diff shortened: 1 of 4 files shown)


harshavardhana:
ping @krishnasrinivas need your reviews here..

krishnasrinivas (approved):

cmd/xl-v1-list-objects.go
```diff
@@ -34,6 +34,7 @@ func listDirFactory(isLeaf isLeafFunc, treeWalkIgnoredErrs []error, disks ...Sto
 			// For any reason disk was deleted or goes offline, continue
 			// and list from other disks if possible.
 			if isErrIgnored(err, treeWalkIgnoredErrs...) {
+				err = nil
```
vadmeste: I recommand to even simplify this function, like this way:

```
 24     listDir := func(bucket, prefixDir, prefixEntry string) (entries []string, delayIsLeaf bool, err error) {
 25         for _, disk := range disks {
 26             if disk == nil {
 27                 continue
 28             }
 29             entries, err = disk.ListDir(bucket, prefixDir)
 30             if err != nil {
 31                 if isErrIgnored(err, treeWalkIgnoredErrs...) {
 32                     continue
 33                 }
 34                 return nil, false, traceError(err)
 35             }
 36             entries, delayIsLeaf = filterListEntries(bucket, prefixDir, entries, prefixEntry, isLeaf)
 37             return entries, delayIsLeaf, nil
 38         }
 39         // Nothing found in all disks
 40         return nil, false, nil
 41     }
```

harshavardhana (approved):

PR #3940, opened 2017-03-20
Implement gateway S3 support

nl5887:
Implement gateway s3 backend.

* the s3 gateway layer will connect to minio or s3 endpoints


krishnasrinivas:
@nl5887 is the PR ready for review?

krishnasrinivas:
@nl5887 lint errors need to be fixed. Check the travis log.

cmd/gateway-main.go
```diff
@@ -55,7 +55,26 @@ var gatewayCmd = cli.Command{
 	Flags: append(serverFlags, cli.BoolFlag{
 		Name:  "quiet",
 		Usage: "Disable startup banner.",
-	}),
+	},
+		cli.StringFlag{
+			Name:  "s3-endpoint",
```
krishnasrinivas: `endpoint` is better than `s3-endpoint`

cmd/gateway-main.go
```diff
@@ -55,7 +55,26 @@ var gatewayCmd = cli.Command{
...
+			Usage: "The endpoint.",
+			Value: "play.minio.io:9000",
+		},
+		cli.StringFlag{
+			Name:  "s3-location",
```
krishnasrinivas: `location` is not needed as minio-go figures it out automatically

cmd/gateway-main.go
```diff
@@ -83,11 +103,14 @@ func mustGetGatewayCredsFromEnv() (accessKey, secretKey string) {
 //
 // - Azure Blob Storage.
 // - Add your favorite backend here.
-func newGatewayLayer(backendType, accessKey, secretKey string) (GatewayLayer, error) {
-	if gatewayBackend(backendType) != azureBackend {
+func newGatewayLayer(backendType, endpoint, location, accessKey, secretKey string) (GatewayLayer, error) {
+	if gatewayBackend(backendType) == azureBackend {
```
krishnasrinivas: switch statement will be better

cmd/gateway-main.go
```diff
@@ -55,7 +55,26 @@ var gatewayCmd = cli.Command{
 	Flags: append(serverFlags, cli.BoolFlag{
 		Name:  "quiet",
 		Usage: "Disable startup banner.",
-	}),
+	},
+		cli.StringFlag{
+			Name:  "s3-endpoint",
+			Usage: "The endpoint.",
+			Value: "play.minio.io:9000",
```
krishnasrinivas: default should be `s3.amazonaws.com`

cmd/s3-layer-anonymous.go
```diff
@@ -0,0 +1,34 @@
+package cmd
+
+import "io"
+
+// AnonGetObject - Get object anonymously
+func (l *S3Layer) AnonGetObject(bucket string, key string, startOffset int64, length int64, writer io.Writer) (err error) {
+	// already an s3 object error
+	return l.GetObject(bucket, key, startOffset, length, writer)
```
krishnasrinivas: it's doing an authenticated operation here

cmd/s3-layer-anonymous.go
```diff
@@ -0,0 +1,34 @@
...
+}
+
+// AnonGetObjectInfo - Get object info anonymously
+func (l *S3Layer) AnonGetObjectInfo(bucket string, object string) (objInfo ObjectInfo, err error) {
+	// already an s3 object error
+	return l.GetObjectInfo(bucket, object)
```
krishnasrinivas: it's doing an authenticated operation here

cmd/s3-layer-anonymous.go
```diff
@@ -0,0 +1,34 @@
...
+}
+
+// AnonListObjects - List objects anonymously
+func (l *S3Layer) AnonListObjects(bucket string, prefix string, marker string, delimiter string, maxKeys int) (result ListObjectsInfo, err error) {
+	result, err = l.ListObjects(bucket, prefix, marker, delimiter, maxKeys)
```
krishnasrinivas: it's doing an authenticated operation here

cmd/s3-layer-anonymous.go
```diff
@@ -0,0 +1,34 @@
...
+	if err != nil {
+		return ListObjectsInfo{}, s3ToObjectError(traceError(err), bucket)
+	}
+	return
+}
+
+// AnonGetBucketInfo - Get bucket metadata anonymously.
+func (l *S3Layer) AnonGetBucketInfo(bucket string) (bucketInfo BucketInfo, err error) {
+	bucketInfo, err = l.GetBucketInfo(bucket)
```
krishnasrinivas: it's doing an authenticated operation here

cmd/s3-layer.go
```diff
@@ -0,0 +1,508 @@
...
+		err = BucketNotFound{Bucket: bucket}
+	case "NoSuchKey":
+		if object != "" {
+			err = ObjectNotFound{Bucket: bucket, Object: object}
+		} else {
+			err = BucketNotFound{Bucket: bucket}
+		}
+	case "XMinioInvalidObjectName":
+		err = ObjectNameInvalid{}
+	default:
```
krishnasrinivas: default statement (and hence fmt.Println statement) can be removed

cmd/s3-layer.go
```diff
@@ -0,0 +1,508 @@
...
+	SecretAccessKey string
+	UseSSL          bool
+}
+
+// S3Layer - Implements Object layer for S3 and Minio blob storage.
+type S3Layer struct {
+	s3LayerConfig
+
+	Client *minio.Core
+}
```
krishnasrinivas: s3Layer can be renamed to s3Gateway because we refer to http handlers as s3Layer during discussions. It can contain just two fields:

```
type struct s3Gateway {
    client *minio.Core                   // For authenticated requests
    anonClient *minio.Core          // For unauthenticated requests
}
```

There is no need to have the struct `s3LayerConfig`

cmd/s3-layer.go
```diff
@@ -0,0 +1,508 @@
...
+
+func newS3Layer(endpoint, location, accessKey, secretKey string) (GatewayLayer, error) {
```
krishnasrinivas: all functions need comments (even though it's redundant sometimes, we follow this convention)

cmd/s3-layer.go
```diff
@@ -0,0 +1,508 @@
...
+	doneCh := make(chan struct{})
+	defer close(doneCh)
+
+	loi := ListObjectsInfo{}
+
+	loi.Objects = make([]ObjectInfo, maxKeys)
+
+	i := 0
+
+	for message := range l.Client.ListObjectsV2(bucket, prefix, true, doneCh) {
```
krishnasrinivas: you can use https://github.com/minio/minio-go/blob/master/core.go : ListObjects()

cmd/s3-layer.go
```diff
@@ -0,0 +1,508 @@
...
+	if err != nil {
+		return ObjectInfo{}, s3ToObjectError(traceError(err), bucket, object)
+	}
+
+	return oi, nil
+}
+
+// CopyObject - Copies a blob from source container to destination container.
+func (l *S3Layer) CopyObject(srcBucket string, srcObject string, destBucket string, destObject string, metadata map[string]string) (ObjectInfo, error) {
+	err := l.Client.CopyObject(destBucket, destObject, fmt.Sprintf("%s/%s", srcBucket, srcObject), minio.CopyConditions{})
```
krishnasrinivas: you can use pathJoin() helper instead of fmt.Sprintf

cmd/s3-layer.go
```diff
@@ -0,0 +1,508 @@
...
+}
+
+// NewMultipartUpload - upload object in multiple parts
+func (l *S3Layer) NewMultipartUpload(bucket string, object string, metadata map[string]string) (uploadID string, err error) {
+	return l.Client.NewMultipartUpload(bucket, object, toMinioClientMetadata(metadata))
+}
+
+// CopyObjectPart - copy part of object to other bucket and object
+func (l *S3Layer) CopyObjectPart(srcBucket string, srcObject string, destBucket string, destObject string, uploadID string, partID int, startOffset int64, length int64) (info PartInfo, err error) {
+	panic("CopyObjectPart: not implemented")
```
krishnasrinivas: you can remove panic() statements here and at other parts of the code and instead return errors.

cmd/s3-layer.go
```diff
@@ -0,0 +1,508 @@
...
+	err := l.Client.CompleteMultipartUpload(bucket, object, uploadID, toMinioClientCompleteParts(uploadedParts))
+	if err != nil {
+		return ObjectInfo{}, nil
+	}
+
+	return l.GetObjectInfo(bucket, object)
+}
+
+// SetBucketPolicies - Set policy on bucket
+func (l *S3Layer) SetBucketPolicies(string, []BucketAccessPolicy) error {
```
krishnasrinivas: this should be implemented.

cmd/s3-layer.go
```diff
@@ -0,0 +1,508 @@
...
+	return traceError(NotImplemented{})
+}
+
+// GetBucketPolicies - Get policy on bucket
+func (l *S3Layer) GetBucketPolicies(string) ([]BucketAccessPolicy, error) {
```
krishnasrinivas: this should be implemented.

cmd/s3-layer.go
```diff
@@ -0,0 +1,508 @@
...
+	return []BucketAccessPolicy{}, traceError(NotImplemented{})
+}
+
+// DeleteBucketPolicies - Delete all policies on bucket
+func (l *S3Layer) DeleteBucketPolicies(string) error {
```
krishnasrinivas: this should be implemented.

krishnasrinivas (changes requested):
I will review the caching code next.

krisis:
@nl5887 Could you add notes describing the architecture/design of gateway cache? It will help in reviewing the PR.

krishnasrinivas:
Current limitations of the cache:
* Object name cannot have `/` in it
* Cache directory should have a format.json that determines the version number of the on-disk format of the cache directory. i.e if we change the backend format we should be able to detect older backend-formats so that either it can be cleaned up or migrated.
* If there is any network error when putting the object into the cache we end up with a partial object in the cache. So for any future requests we would reply with partial object.
* It is not caching PutObject requests
* Does not cater to Anonymous requests
* Does not validate cache with backend
* Grows the disk cache indefinitely
* If the backend is down, it should behave intelligently. i.e it should allow anonymous GET requests if the object was cached during a previous anonymous request.

Cache will need significant changes to accommodate the above requirements.

krishnasrinivas:
Can you guys review this: https://github.com/krishnasrinivas/wikinotes/wiki/Caching-gateway-proposal

vadmeste:
> Can you guys review this: https://github.com/krishnasrinivas/wikinotes/wiki/Caching-gateway-proposal

For the sake of completeness, can you tell when an object will be cached ? (on the fly when a client asks for it?) and when objects will be removed from the cache.

cmd/s3-layer.go
nl5887: Still need to implement the BucketPolicy functions, all other issues have been solved.

nl5887:
When the s3 gateway is completed and through review, I'll implement the caching layer completely. This was just an implementation to show how it will be implemented as a wrapping layer.

krishnasrinivas:
>> Can you guys review this: https://github.com/krishnasrinivas/wikinotes/wiki/Caching-gateway-proposal

> For the sake of completeness, can you tell when an object will be cached ? (on the fly when a client asks for it?) and when objects will be removed from the cache.

We can fill the cache during:
* GetObject - if the object is not in the cache
* PutObject

cmd/s3-layer-anonymous.go
```diff
@@ -3,32 +3,62 @@ package cmd
...
+	if err != nil {
+		return s3ToObjectError(traceError(err), bucket, key)
+	}
+
+	object.Seek(startOffset, io.SeekStart)
+	if _, err := io.CopyN(writer, object, length); err != nil {
+		return s3ToObjectError(traceError(err), bucket, key)
+	}
+
+	object.Close()
```
krishnasrinivas: `defer object.Close()` should be done and moved to the line 11

cmd/s3-layer-anonymous.go
```diff
@@ -3,32 +3,62 @@ package cmd
...
-func (l *S3Layer) AnonListObjects(bucket string, prefix string, marker string, delimiter string, maxKeys int) (result ListObjectsInfo, err error) {
-	result, err = l.ListObjects(bucket, prefix, marker, delimiter, maxKeys)
+func (l *s3Gateway) AnonListObjects(bucket string, prefix string, marker string, delimiter string, maxKeys int) (ListObjectsInfo, error) {
+	loi := ListObjectsInfo{}
+
+	loi.Objects = make([]ObjectInfo, maxKeys)
+
+	result, err := l.anonClient.ListObjects(bucket, prefix, marker, delimiter, maxKeys)
 	if err != nil {
-		return ListObjectsInfo{}, s3ToObjectError(traceError(err), bucket)
```
krishnasrinivas: `return ListObjectsInfo{}, s3ToObjectError(traceError(err), bucket)` should be done?

cmd/s3-layer-anonymous.go
```diff
@@ -3,32 +3,62 @@ package cmd
...
 	}
-	return
+
+	return fromMinioClientListBucketResult(bucket, result), nil
 }
 
 // AnonGetBucketInfo - Get bucket metadata anonymously.
-func (l *S3Layer) AnonGetBucketInfo(bucket string) (bucketInfo BucketInfo, err error) {
-	bucketInfo, err = l.GetBucketInfo(bucket)
+func (l *s3Gateway) AnonGetBucketInfo(bucket string) (BucketInfo, error) {
```
krishnasrinivas: Since S3 does not allow anonymous listing of buckets, what you can do is do `l.anonClient.BucketExists()` if there is no error then do authenticated ListBuckets to get the `creationDate`, if there was any error from `l.anonClient.BucketExists()` then we can just return error.

cmd/s3-layer.go
```diff
@@ -77,73 +78,68 @@ func s3ToObjectError(err error, params ...string) error {
...
 		err = ObjectNameInvalid{}
+	case "AccessDenied":
+		err = PrefixAccessDenied{
+			Bucket: bucket,
+			Object: object,
+		}
 	default:
-		fmt.Println("Undefined s3 object error:", bucket, object, minioErr.Code)
+		// should not be reached
+		err = fmt.Errorf("Undefined s3 object error: bucket=%s object=%s code=%s", bucket, object, minioErr.Code)
```
krishnasrinivas: This will cause the original error to be replaced with fmt.Errorf() error, hence you could just remove the entire `default` section from the switch statement. There is a high chance of us reaching default statement as number `if` cases in the switch statement is limited.

cmd/s3-layer.go
```diff
@@ -77,73 +78,68 @@ func s3ToObjectError(err error, params ...string) error {
...
-	config := s3LayerConfig{
-		Endpoint:        endpoint,
-		Location:        location,
-		AccessKeyID:     accessKey,
-		SecretAccessKey: secretKey,
-		UseSSL:          true,
+// newS3Gateway returns s3 gatewaylayer
+func newS3Gateway(endpoint, accessKey, secretKey string) (GatewayLayer, error) {
+	// Initialize minio client object.
+	client, err := minio.NewCore(endpoint, accessKey, secretKey, true)
```
krishnasrinivas: since we need to allow for http and https, endpoint would be `http://host.com` so endpoint needs to be parsed to see if the scheme is http or https and accordingly call `minio.NewCore` with the isSecure option

cmd/s3-layer.go
```diff
@@ -191,34 +187,35 @@ func (l *S3Layer) DeleteBucket(bucket string) error {
...
 
-	i := 0
-
-	for message := range l.Client.ListObjectsV2(bucket, prefix, true, doneCh) {
-		if message.Err != nil {
-			return ListObjectsInfo{}, s3ToObjectError(traceError(message.Err), bucket)
-		}
+	result, err := l.Client.ListObjects(bucket, prefix, marker, delimiter, maxKeys)
+	if err != nil {
+		return ListObjectsInfo{}, err
```
krishnasrinivas: return s3ToObjectError(traceError(message.Err), bucket) ?

krishnasrinivas (changes requested):

krishnasrinivas:
Discussed with @abperiasamy and @harshavardhana , updated document: https://github.com/krishnasrinivas/wikinotes/wiki/Caching-gateway-proposal

krishnasrinivas:
@nl5887 do you think it is better to have two separate PRs, one for S3 backend and one for Caching. That way Caching can be implemented/merged separately.

krishnasrinivas:
We had another round of discussion and I have updated the [document](https://github.com/krishnasrinivas/wikinotes/wiki/Caching-gateway-proposal)

nl5887:
I think we should separate the s3 gateway and the caching layer. The s3 gateway is almost ready and can be merged.

krishnasrinivas:
> I think we should separate the s3 gateway and the caching layer. The s3 gateway is almost ready and can be merged.

Cool. You can remove the caching related code so that this PR can be merged.

cmd/s3-layer.go
```diff
@@ -0,0 +1,511 @@
...
+}
+
+// StorageInfo - Not relevant to S3 backend.
+func (l *s3Gateway) StorageInfo() StorageInfo {
+	return StorageInfo{}
+}
+
+// MakeBucket - Create a new container on S3 backend.
+func (l *s3Gateway) MakeBucket(bucket string) error {
+	err := l.Client.MakeBucket(bucket, "us-east-1")
```
krisis: we should make "us-east-1" a constant.

cmd/s3-layer.go
```diff
@@ -0,0 +1,511 @@
...
+		return s3ToObjectError(traceError(err), bucket)
+	}
+	return nil
+}
+
+// ListObjects - lists all blobs in S3 bucket filtered by prefix
+func (l *s3Gateway) ListObjects(bucket string, prefix string, marker string, delimiter string, maxKeys int) (ListObjectsInfo, error) {
+	loi := ListObjectsInfo{}
+
+	loi.Objects = make([]ObjectInfo, maxKeys)
```
krisis: `loi` is unused.

cmd/s3-layer.go
```diff
@@ -0,0 +1,511 @@
...
+		Size:         op.Size,
+		ETag:         op.ETag,
+		LastModified: op.LastModified,
+		PartNumber:   op.PartNumber,
+	}
+}
+
+// PutObjectPart puts a part of object in bucket
+func (l *s3Gateway) PutObjectPart(bucket string, object string, uploadID string, partID int, size int64, data io.Reader, md5Hex string, sha256sum string) (PartInfo, error) {
+	md5HexBytes, err := hex.DecodeString(md5Hex)
```
donatello, line 461: Why is `md5Hex` argument hex encoded here? It is getting decoded here, and then getting re-encoded as base64 again inside minio-go

cmd/s3-layer.go
```diff
@@ -0,0 +1,511 @@
...
+	if err != nil {
+		return PartInfo{}, err
+	}
+
+	sha256sumBytes, err := hex.DecodeString(sha256sum)
```
donatello, line 466: Why is `sha256sum` argument hex encoded here? It is getting decoded here, and then getting re-encoded as base64 again inside minio-go

cmd/s3-layer.go
```diff
@@ -0,0 +1,511 @@
...
+	}
+
+	delete(metadata, "md5Sum")
+
+	err := l.Client.PutObject(bucket, object, size, teeReader, toMinioClientMetadata(metadata))
+	if err != nil {
+		return ObjectInfo{}, s3ToObjectError(traceError(err), bucket, object)
+	}
+
+	if sha256sum != "" {
```
krisis, line 336: minio-go computes sha256 sum, why do we need to compute it at the gateway too?

cmd/s3-layer.go
harshavardhana: should be made configurable.

nl5887: had this configurable, this is comment of @krishnasrinivas:

```location is not needed as minio-go figures it out automatically```

File: cmd/s3-layer.go (hunk 27 above)
Comment (nl5887) [reply cmd/s3-layer.go:466]: the minio-go core expects []byte if we want to pass it as base64 string we need to adjust core. In my opinion it is better to pass it as []byte and just settle with the de- and encoding.

File: cmd/s3-layer.go (hunk 28 above)
Comment (nl5887) [reply cmd/s3-layer.go:336]: just tested it, the Mint TestSHA256Mismatch fails without this code. The sha 256 mismatch error isn't being returned by minio-go.

Comment (krisis) [reply cmd/s3-layer.go:336]: interesting, this https://github.com/minio/minio-go/blob/master/api-put-object.go#L204 made me believe that for V4 signature and non-HTTPS client, minio-go SDK computes sha256.

Comment (nl5887) [reply cmd/s3-layer.go:336]: yeah, I was looking at the same. It calculates them, but won't return an error at all.

File: cmd/s3-layer.go (hunk 24 above)
Comment (harshavardhana) [reply cmd/s3-layer.go]: It is useful for bucket creates.  Here it defaults to `us-east-1` perhaps even need to configure during newGatewayConfig() for different S3 endpoints. Through region as ENV. We need to discuss @krishnasrinivas 

Otherwise al requests would go to same endpoint us-east.

File: cmd/s3-layer.go (hunk 28 above)
Comment (harshavardhana) [reply cmd/s3-layer.go:336]: Server doing Sha is correct we do this with Azure as well. Double checksum is fine. Usually gateway is talking to an SSL enabled remote S3 server in those cases minio-go will not compute sha256. We can document this as a caveat but we storngly recommend using a secure remote S3. 

With Amazon S3 as backend it's already solved.

Comment (harshavardhana) [reply cmd/s3-layer.go:336]: Your streaming signature change would help here @krisis as well.

File: cmd/s3-layer.go (hunk 24 above)
Comment (krishnasrinivas) [reply cmd/s3-layer.go]: @harshavardhana yes it's needed for bucket creation, actually we can use the location info sent by the client:
```
// createBucketConfiguration container for bucket configuration.                                                                                                                        
type createBucketConfiguration struct {                                                                                                                                                 
        XMLName  xml.Name `xml:"http://s3.amazonaws.com/doc/2006-03-01/ CreateBucketConfiguration" json:"-"`                                                                            
        Location string   `xml:"LocationConstraint"`                                                                                                                                    
}
```

File: cmd/s3-layer.go (hunk 28 above)
Comment (krishnasrinivas) [reply cmd/s3-layer.go:336]: we can expose the following function in minio-go:
```
func (c Client) putObjectDo(bucketName, objectName string, reader io.Reader, md5Sum []byte, sha256Sum []byte, size int64, metaData map[string][]string) (ObjectInfo, error) {
```

Comment (harshavardhana) [reply cmd/s3-layer.go:336]: We don't need do it doesn't add any benefit.  A sha should be validated by the gateway not minio-go.  minio-go is just calculating upon incoming stream and uploading.

File: cmd/s3-layer.go (hunk 24 above)
Comment (harshavardhana) [reply cmd/s3-layer.go]: This might be sufficient.. for location.

File: cmd/s3-layer.go (hunk 28 above)
Comment (krishnasrinivas) [reply cmd/s3-layer.go:336]: It adds benefit, i.e the sha256 verification will be done by the backend S3 server not minio-go. Else gateway will have to compute sha256 twice - once in object-layer and once in minio-go.

Comment (harshavardhana) [reply cmd/s3-layer.go:336]: Again looking at the code it makes sense to add what you are saying, avoids the RemoveObject() and returns ObjectInfo{} is one call.

File: cmd/s3-layer.go (hunk 27 above)
Comment (krishnasrinivas) [reply cmd/s3-layer.go:466]: At the API level it would be more appropriate to represent md5/sha as `string` ... I will make those changes to minio-go. It will make S3 gateway code simpler too, it would just pass along these values instead of converting from string->[]byte

Comment (donatello) [reply cmd/s3-layer.go:466]: Just had a better look at this - sha256sum is used hex encoded and md5sum is used base64 encoded ("used" => S3 API requirement). There is some unneeded/repeated conversion between binary ([]byte) to string (hex/base64 encodings) representations and vice versa. 

Ideally, hashes need an explicit type that store just bits (using []byte) and have methods for conversion to various formats like hex or base64 (and this can be taken up outside this patch).

File: cmd/gateway-main.go [hunk 29]
```diff
@@ -83,11 +91,23 @@ func mustGetGatewayCredsFromEnv() (accessKey, secretKey string) {
...
+	case s3Backend:
+		useHTTPS := true
+		if u, err := url.Parse(endpoint); err != nil {
+			return nil, err
+		} else if u.Scheme == "http" || u.Scheme == "https" {
+			useHTTPS = (u.Scheme == "https")
+			endpoint = u.Host
+		}
+
+		return newS3Gateway(endpoint, useHTTPS, accessKey, secretKey)
```
donatello:
```go
case s3Backend:
	if u, err := url.Parse(endpoint); err != nil {
		return nil, err
	}

	return newS3Gateway(u.Host, u.Scheme == "https", accessKey, secretKey)
```

I think concise is better.

cmd/s3-layer-anonymous.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	return fromMinioClientObjectInfo(bucket, oi), nil
+}
+
+// AnonListObjects - List objects anonymously
+func (l *s3Gateway) AnonListObjects(bucket string, prefix string, marker string, delimiter string, maxKeys int) (ListObjectsInfo, error) {
+	loi := ListObjectsInfo{}
+
+	loi.Objects = make([]ObjectInfo, maxKeys)
```
donatello: `loi` is unused.

cmd/s3-layer.go
```diff
@@ -0,0 +1,511 @@
...
+		Size:            oi.Size,
+		MD5Sum:          oi.ETag,
+		UserDefined:     userDefined,
+		ContentType:     oi.ContentType,
+		ContentEncoding: oi.Metadata.Get("Content-Encoding"),
+	}
+}
+
+// GetObjectInfo - reads blob metadata properties and replies back ObjectInfo,
+// uses zure equivalent GetBlobProperties.
```
donatello: Outdated comment.

cmd/gateway-main.go
nl5887: Current code allows urls without scheme as well. Eg. both http://play.minio.io:9000 and play.minio.io:9000 are supported now.

cmd/azure.go
```diff
@@ -151,7 +151,7 @@ func (a AzureObjects) StorageInfo() StorageInfo {
 }
 
 // MakeBucket - Create a new container on azure backend.
-func (a AzureObjects) MakeBucket(bucket string) error {
+func (a AzureObjects) MakeBucket(location, bucket string) error {
```
krishnasrinivas: you can override MakeBucket in gateway-interface as location does not make sense for fs/xl

cmd/s3-layer.go
krishnasrinivas: that is a better way to represent md5 and sha256:
`[md5.Size]byte` and `[sha256.Size]byte`

cmd/fs-v1.go
```diff
@@ -203,7 +203,7 @@ func (fs fsObjects) statBucketDir(bucket string) (os.FileInfo, error) {
 
 // MakeBucket - create a new bucket, returns if it
 // already exists.
-func (fs fsObjects) MakeBucket(bucket string) error {
+func (fs fsObjects) MakeBucket(location, bucket string) error {
```
harshavardhana: Location should be the last argument .. All optional arguments go to the end.   Also this method signature change is not needed.

cmd/handler-utils.go
```diff
@@ -27,7 +27,7 @@ import (
 // Validates location constraint in PutBucket request body.
 // The location value in the request body should match the
 // region configured at serverConfig, otherwise error is returned.
-func isValidLocationConstraint(r *http.Request) (s3Error APIErrorCode) {
+func isValidLocationConstraint(r *http.Request) (location string, s3Error APIErrorCode) {
```
harshavardhana: isValidLocationConstraint should never return `location`   change this function to take just the `location` string and parse the location outside in a different function as 

```go
func parseLocationConstraint(r *http.Request)  (location string, s3Error APIErrorCode) {
        // If the request has no body with content-length set to 0,
	// we do not have to validate location constraint. Bucket will
	// be created at default region.
	locationConstraint := createBucketLocationConfiguration{}
	err = xmlDecoder(r.Body, &locationConstraint, r.ContentLength)
 	       errorIf(err, "Unable to xml decode location constraint")
               return "", ErrMalformedXML
        }
        // Successfully decoded, proceed to verify the region.
	// Once region has been obtained we proceed to verify it.
	location = locationConstraint.Location
	if location == "" {
		// Location constraint is empty for region globalMinioDefaultRegion,
		// in accordance with protocol.
		location = globalMinioDefaultRegion
	}
        return location, nil 
}

func isValidLocation(location string) (s3Error APIErrorCode) {
         if serverConfig.GetRegion() != location {
                  return ErrInvalidRegion
         }
         return ErrNone
}

```

harshavardhana (changes requested):

cmd/azure.go
nl5887: That will imply a lot of changes:

* ObjectLayer cannot be implemented anymore within GatewayLayer, because MakeBucket has different signature within ObjectLayer and GatewayLayer and cannot override. So need to copy all methods from interface ObjectLayer to GatewayLayer
* globalObjectAPI cannot be used anymore, because we have ObjectLayer and GatewayLayer difererent signature for MakeBucket, need to find other solution for that.

Now I've chosen to add a MakeBucketWithLocation to comply to GatewayLayer interface and MakeBucket to comply to ObjectLayer interface which will call the default region. The MakeBucket will be never called because PutBucket has been overridden for the GatewayLayer.

cmd/s3-layer.go
nl5887: I'd oppose to changing minio-go to strings, because those are undefined (we don't know if they are valid strings). I'd say you want to pass byte hashes instead of strings. And make the appropriate conversions where necessary, eg in the gateway.

cmd/gateway-main.go
harshavardhana: The url.Parse fails for address in format "play.minio.io:9000" in go1.8 , i would suggest we avoid that kind of support or if needed explicitly handle it using `net.SplitHostPort()` 

https://github.com/golang/go/commit/c5ccbdd22bdbdc43d541b7e7d4ed66ceb559030e#comments

nl5887: We need to try to parse the endpoint anyhow to see get the scheme. I'll remove the err!=nil check to pass and default to use https.

harshavardhana:
Can the description and title of the PR be changed since the caching is moved out now?

cmd/handler-utils.go
```diff
@@ -49,11 +47,28 @@ func isValidLocationConstraint(r *http.Request) (s3Error APIErrorCode) {
...
 	// Treat all other failures as XML parsing errors.
-	return ErrMalformedXML
+	return "", ErrMalformedXML
+
+}
+
+// Validates location constraint in PutBucket request body.
+// The location value in the request body should match the
+// region configured at serverConfig, otherwise error is returned.
+func isValidLocationConstraint(r *http.Request) (s3Error APIErrorCode) {
```
harshavardhana: This is changed in the master to simplify this code.. you might have to rebase.

cmd/s3-layer-anonymous.go
```diff
@@ -0,0 +1,67 @@
+package cmd
```
harshavardhana, line 17: Add license header..

cmd/s3-layer-anonymous.go
```diff
@@ -0,0 +1,67 @@
...
+	if err != nil {
+		return ListObjectsInfo{}, s3ToObjectError(traceError(err), bucket)
+	}
+
+	return fromMinioClientListBucketResult(bucket, result), nil
+}
+
+// AnonGetBucketInfo - Get bucket metadata anonymously.
+func (l *s3Gateway) AnonGetBucketInfo(bucket string) (BucketInfo, error) {
+	if exists, err := l.anonClient.BucketExists(bucket); err != nil {
```
harshavardhana, line 60: To optimize this perhaps we don't need to do `ListBuckets()` just to fill in `BucketInfo()` - since this call is used only in HeadBucketHandler() we can simply change the API to not return BucketInfo() at all. 

In all its usage i see that `BucketInfo{}` is not used at all.

cmd/s3-layer-unsupported.go
```diff
@@ -0,0 +1,26 @@
+package cmd
```
harshavardhana, line 17: license header..

cmd/s3-layer.go
```diff
@@ -0,0 +1,512 @@
...
+			Bucket: bucket,
+			Object: object,
+		}
+	}
+
+	e.e = err
+	return e
+}
+
+// s3Gateway - Implements gateway for S3 and Minio blob storage.
```
harshavardhana: s/"Implements gateway for Minio and S3 compatible object storage servers."/

cmd/s3-layer.go
```diff
@@ -0,0 +1,512 @@
...
+	err := l.Client.RemoveBucket(bucket)
+	if err != nil {
+		return s3ToObjectError(traceError(err), bucket)
+	}
+	return nil
+}
+
+// ListObjects - lists all blobs in S3 bucket filtered by prefix
+func (l *s3Gateway) ListObjects(bucket string, prefix string, marker string, delimiter string, maxKeys int) (ListObjectsInfo, error) {
+	result, err := l.Client.ListObjects(bucket, prefix, marker, delimiter, maxKeys)
```
harshavardhana, line 198: Perhaps we should use the latest V2 ListObjects API. Needs changes in `minio-go` core. // cc @krishnasrinivas

harshavardhana: We should also implement it at gateway handler layer. Because some aws-sdks might start making V2 List default. So we need to bring it in even for Azure gateway. // cc @krishnasrinivas

cmd/s3-layer.go
```diff
@@ -0,0 +1,512 @@
...
+func (l *s3Gateway) GetObjectInfo(bucket string, object string) (objInfo ObjectInfo, err error) {
+	oi, err := l.Client.StatObject(bucket, object)
+	if err != nil {
+		return ObjectInfo{}, s3ToObjectError(traceError(err), bucket, object)
+	}
+
+	return fromMinioClientObjectInfo(bucket, oi), nil
+}
+
+// PutObject - Create a new blob with the incoming data,
```
harshavardhana: Let's call it `blob` it is only of Azure its called `blob`, we should call it object here in S3 context.

cmd/s3-layer.go
```diff
@@ -0,0 +1,512 @@
...
+	return mm
+}
+
+// NewMultipartUpload - upload object in multiple parts
+func (l *s3Gateway) NewMultipartUpload(bucket string, object string, metadata map[string]string) (uploadID string, err error) {
+	return l.Client.NewMultipartUpload(bucket, object, toMinioClientMetadata(metadata))
+}
+
+// CopyObjectPart - copy part of object to other bucket and object
+func (l *s3Gateway) CopyObjectPart(srcBucket string, srcObject string, destBucket string, destObject string, uploadID string, partID int, startOffset int64, length int64) (info PartInfo, err error) {
```
harshavardhana, line 444: Perhaps add a FIXME here, so that we can open relevant bugs and address this.

cmd/s3-layer.go
```diff
@@ -0,0 +1,512 @@
...
+}
+
+// GetBucketPolicies - Get policy on bucket
+func (l *s3Gateway) GetBucketPolicies(bucket string) ([]BucketAccessPolicy, error) {
+	return []BucketAccessPolicy{}, traceError(NotImplemented{})
+}
+
+// DeleteBucketPolicies - Delete all policies on bucket
+func (l *s3Gateway) DeleteBucketPolicies(string) error {
+	return traceError(NotImplemented{})
```
harshavardhana: We need to implement this..

cmd/s3-layer_test.go
```diff
@@ -0,0 +1 @@
+package cmd
```
harshavardhana, line 17: License header missing.

harshavardhana (changes requested):

cmd/s3-layer-anonymous.go
krishnasrinivas: > In all its usage i see that BucketInfo{} is not used at all.

meaning?
`mc ls s3/bucketname` issues a HEAD on the bucket

so returning a BucketInfo{} will be necessary ... right?

cmd/s3-layer.go
krishnasrinivas: yes @harshavardhana https://github.com/minio/minio/issues/4043 currently only authenticated list-v2 works.

cmd/s3-layer-anonymous.go
harshavardhana: No it is not necessary BucketInfo{} is never used HEAD has no body. @krishnasrinivas we only need to see if Bucket indeed exists.. GetBucketInfo() was implemented for XL and FS which implement the Stat() call.  Doing `ListBuckets()` and filtering is not needed..

krishnasrinivas:
@nl5887 can you
* update minio-go package
* fix conflicts
* make sure travis CI runs fine (`make test`)
* address remaining comments
* add bucket policy functionality

Can this be done tomorrow? I will co-ordinate with India team for testing.

cmd/s3-layer-anonymous.go
nl5887: The function is called GetBucketInfo so we should return a valid BucketInfo with correct fields if we keep GetBucketInfo. So we'd better remove GetBucketInfo for gateway and replace it with BucketExists.

cmd/s3-layer.go
nl5887: Don't you think the policies are too backend specific (eg if we're going to add more and more backends, with all their specifics? I assume people using the gateway will use it for main functionality, not for configuration. They'll use the backend directly.)

nl5887:
* minio-go has been updated
* rebased to master
* tests fail on coverage, not on other tests. We need to add more tests to both s3, azure and gateway to get above the 80% again.
* remaining comments has been addressed

two remaining issues:
* implement listobjectsv2, needs to be added to azure and gateway as well
* implement bucket policies (but have an outstanding question about this) -> (```Don't you think the policies are too backend specific (eg if we're going to add more and more backends, with all their specifics? I assume people using the gateway will use it for main functionality, not for configuration. They'll use the backend directly.)```

I think new issues needs to be made for the two remaining issues, which can be implemented separately.

cmd/s3-layer.go
krishnasrinivas: @nl5887 we'll only support a subset of {private, read-only, read-write, write-only} depending on what the backend supports.
for example for azure we support {private, read-only}

krishnasrinivas:
> implement listobjectsv2, needs to be added to azure and gateway as well
> implement bucket policies (but have an outstanding question about this) -> (Don't you think the policies are too backend specific (eg if we're going to add more and more backends, with all their specifics? I assume people using the gateway will use it for main functionality, not for configuration. They'll use the backend directly.)

these can be implemented before we start testing. I can send PR for bucket-policies to your branch. Let me know.

cmd/s3-layer-anonymous.go
harshavardhana: Agreed we should do that makes more sense in gateway context.

cmd/s3-layer.go
```diff
@@ -0,0 +1,588 @@
...
+	objects := make([]ObjectInfo, len(result.Contents))
+
+	for i, oi := range result.Contents {
+		objects[i] = fromMinioClientObjectInfo(bucket, oi)
+	}
+
+	return ListObjectsV2Info{
+		IsTruncated: result.IsTruncated,
+		Prefixes: []string{
+			result.Prefix,
```
vadmeste: I think you want to fill Prefixes with result.CommonPrefixes elements instead.

cmd/s3-layer.go
```diff
@@ -0,0 +1,588 @@
...
+	objects := make([]ObjectInfo, len(result.Contents))
+
+	for i, oi := range result.Contents {
+		objects[i] = fromMinioClientObjectInfo(bucket, oi)
+	}
+
+	return ListObjectsInfo{
+		IsTruncated: result.IsTruncated,
+		NextMarker:  result.NextMarker,
+		Prefixes: []string{
```
vadmeste: I think you want to fill Prefixes with result.CommonPrefixes elements instead.

cmd/s3-layer.go
```diff
@@ -0,0 +1,588 @@
...
+// length indicates the total length of the object.
+func (l *s3Gateway) GetObject(bucket string, key string, startOffset int64, length int64, writer io.Writer) error {
+	object, err := l.Client.GetObject(bucket, key)
+	if err != nil {
+		return s3ToObjectError(traceError(err), bucket, key)
+	}
+
+	defer object.Close()
+
+	object.Seek(startOffset, io.SeekStart)
```
vadmeste: Better to check errors returned from object.Seek()

cmd/s3-layer.go
```diff
@@ -0,0 +1,588 @@
...
+		mparts[i] = toMinioClientCompletePart(part)
+	}
+	return mparts
+}
+
+// CompleteMultipartUpload completes ongoing multipart upload and finalizes object
+func (l *s3Gateway) CompleteMultipartUpload(bucket string, object string, uploadID string, uploadedParts []completePart) (ObjectInfo, error) {
+	err := l.Client.CompleteMultipartUpload(bucket, object, uploadID, toMinioClientCompleteParts(uploadedParts))
+	if err != nil {
+		return ObjectInfo{}, nil
```
vadmeste: We should return an error here, not nil.

harshavardhana:
@nl5887 we need to re-vendorize minio-go with the latest changes.

krishnasrinivas:
@nl5887 this change is needed:
```
krishna@escape:~/gocode/src/github.com/minio/minio$ git diff
diff --git a/cmd/s3-layer.go b/cmd/s3-layer.go
index c592503..f38d7be 100644
--- a/cmd/s3-layer.go
+++ b/cmd/s3-layer.go
@@ -84,6 +84,8 @@ func s3ToObjectError(err error, params ...string) error {
                        Bucket: bucket,
                        Object: object,
                }
+       case "XAmzContentSHA256Mismatch":
+               err = SHA256Mismatch{}
        }
 
        e.e = err

```

to fix:
```
krishna@escape:~/gocode/src/github.com/minio/mint/functional-test/minio-functional-test$ go test -run=TestSHA256Mismatch
--- FAIL: TestSHA256Mismatch (0.00s)
	server_test.go:304: Expected response code to be XAmzContentSHA256Mismatch, got InternalError.
	server_test.go:307: Expected response Message to be The provided 'x-amz-content-sha256' header does not match what was computed., got We encountered an internal error, please try again..
	server_test.go:310: Expected response status code to be 400, got 500.
FAIL
exit status 1
FAIL	github.com/minio/mint/functional-test/minio-functional-test	0.007s
krishna@escape:~/gocode/src/github.com/minio/mint/functional-test/minio-functional-test$ 
```

cmd/s3-layer.go
```diff
@@ -0,0 +1,597 @@
...
+		if err := l.Client.SetBucketPolicy(bucket, prefix, policy.Policy); err != nil {
+			return s3ToObjectError(traceError(err), bucket, policy.Prefix)
+		}
+	}
+	return nil
+}
+
+// fromMinioClientListBucketPolicies converts from map[string]policy.BucketPolicy to []BucketAccessPolicy
+func fromMinioClientListBucketPolicies(bucket string, bucketPolicies map[string]policy.BucketPolicy) []BucketAccessPolicy {
+	policies := make([]BucketAccessPolicy, 0)
```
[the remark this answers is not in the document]
harshavardhana: Can this be fixed https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices?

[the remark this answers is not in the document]
harshavardhana: Can this be fixed? - https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices

nl5887:
After giving it some thought, I decided to add an extra error, ErrNotSupported. This will be used for functionality that is either not supported by the defined backend or impossible to implement unambiguously. It is better to be clear about what is supported and what is not, then having unclear expectations on what the results will be on the backend. I also expect in the case of the gateway that users will use it for objects and bucket related operations and not for notifications or policy modifications. Those will be set using the own backend tooling / ui's.

harshavardhana:
Can you fix the build CI  ? @nl5887

harshavardhana (approved):

krishnasrinivas (approved):

PR #3951, opened 2017-03-22
pkg/quick: add Save() function and other enhancements.

balamurugana:
* Add a new function Save() which saves given configuration into given file.
* Simplify Load() function.
* Remove unused CheckVersion().
* CheckData() is a private function now.
* quick_test.go is part of quick package now.
* minio server uses top level quick.Load() and quick.Save() functions.

## How Has This Been Tested?

`make test`

## Types of changes

## Checklist:

### DIFF
--- a/cmd/config-migrate.go
+++ b/cmd/config-migrate.go
@@ -88,33 +88,34 @@ func migrateConfig() error {
 
 // Version '1' is not supported anymore and deprecated, safe to delete.
 func purgeV1() error {
-	cv1, err := loadConfigV1()
-	if err != nil {
-		if os.IsNotExist(err) {
-			return nil
-		}
-		return fmt.Errorf("Unable to load config version ‘1’. %v", err)
+	configFile := filepath.Join(getConfigDir(), "fsUsers.json")
 
-	}
-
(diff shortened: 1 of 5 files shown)


cmd/config-migrate.go
```diff
@@ -88,33 +88,34 @@ func migrateConfig() error {
...
-
-	}
+	configFile := filepath.Join(getConfigDir(), "fsUsers.json")
 
-	if cv1.Version == "1" {
-		// Purge old fsUsers.json file
-		configFile := filepath.Join(getConfigDir(), "fsUsers.json")
-		removeAll(configFile)
-		console.Println("Removed unsupported config version ‘1’.")
+	cv1 := &configV1{Version: "1"}
```
vadmeste: Looks like setting version here is not useful anymore (qc.Load() doesn't check for version header anymore)

vadmeste:
LGTM except one minor comment.

cmd/config-migrate.go
balamurugana: Yes. Doesn't check and the check didn't work previously in `Load()`.

donatello (approved):
:+1:

Issue #3928, opened 2017-03-18
notification: [feature] Support `access` event format for database backends. 


harshavardhana:
Currently on our database targets 

- redis
- elasticsearch
- postgres
- mysql 

Update the target table / entries - keeping the state of the bucket namespace replicated, while this is one of required features. There is also another growing need to save all the generate events in the database while not replicating the namespace but instead we are saving the access information.

For this to happen we need to support different `formats` for database targets. 

`format: access`, 
`format: namespace`

Each of the config entries should have a new field called as `format` which would describe the way a given notification target would behave. 

Under `namespace` no new code needs to be added the format is exactly what we are doing today, but for format as `access` we need to ensure that rather than saving the entry in database by the Bucket/Key as the unique resource we need to store the events by their time.

PR #3953, opened 2017-03-22
Support `access` format for db targets

donatello:
This PR adds:

1. format config parameter for database kind of notification targets.
2. postgres access format functionality.
3. mysql access format functionality.

## Description

When the _namespace_ format is used, Minio synchronizes objects in the bucket with rows in the table. It creates rows with two columns: key_name and value. The key_name is the bucket and object name of an object that exists in Minio. The value is JSON encoded event data about the operation that created/replaced the object in Minio. When objects are updated or deleted, the corresponding row from this table is updated or deleted respectively.

When the _access_ format is used, Minio appends events to a table. It creates rows with two columns: event_time and event_data. The event_time is the time at which the event occurred in the Minio server. The event_data is the JSON encoded event data about the operation on an object. No rows are deleted or modified in this format.

## Motivation and Context

Towards fixing https://github.com/minio/minio/issues/3928

## Types of changes

## Checklist:


cmd/config-migrate.go
```diff
@@ -1110,3 +1114,145 @@ func migrateV14ToV15() error {
...
+
+// Version '15' to '16' migration. Adds "format" configuration
+// parameter for database targets.
+func migrateV15ToV16() error {
+	cv15, err := loadConfigV15()
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil
+		}
+		return fmt.Errorf("Unable to load config version ‘14’. %v", err)
```
vadmeste: version 15 :)

cmd/config-migrate.go
```diff
@@ -1110,3 +1114,145 @@ func migrateV14ToV15() error {
...
+	}
+	if len(cv15.Notify.ElasticSearch) == 0 {
+		srvConfig.Notify.ElasticSearch = make(map[string]elasticSearchNotify)
+		srvConfig.Notify.ElasticSearch["1"] = elasticSearchNotify{}
+	} else {
+		// IMPORTANT NOTE: Future migrations should remove
+		// this as existing configuration will already contain
+		// a value for the "format" parameter.
+		for k, v := range cv15.Notify.ElasticSearch.Clone() {
+			if v.Enable {
```
vadmeste: I am not sure I understand what's going on here, there is not default value but we are enforcing users to set it ?

cmd/config-migrate.go
donatello: Thanks :)

cmd/config-migrate.go
donatello: We are adding a parameter called "format" to 4 targets (postgres, mysql, elasticsearch and redis). The values can be "namespace" (existing behaviour) and "access" (new behaviour where all events are appended in the dbs). For existing user configurations, the value has to be "namespace", so that is what the migration in this commit is doing. There is no default, as the user is expected to choose the kind of behaviour they want, and the parameter is ignored if Enable=false. For future migrations (say version 16 to 17), we no longer need to migrate for "format" values and the indicated loop above should be removed - we should just be copying over existing format values.

donatello:
Just rebased the PR and handled conflicts after @balamurugana change - please check migration is being done correctly here.

cmd/config-migrate.go
vadmeste: > and the parameter is ignored if Enable=false.

I think it would be better to put `namespace` whether Enable is activated or not, because:
1. No harm to do so.
2. If the user manually enables later, it will get an error indicating that he should set format field, right?

donatello: Ok, I think your idea is better. It also let's us split up this change into multiple PRs (one per db maybe) to ensure quality. I am making changes for this.

donatello:
Added `access` format for PostgreSQL and updated documentation.

harshavardhana:
Ping @vadmeste can you review?

vadmeste:
> Ping @vadmeste can you review?

@donatello there are some conflicts in this PR, can you address them?

donatello:
> @donatello there are some conflicts in this PR, can you address them?

@vadmeste Fixed now - please check.


@harshavardhana Could me merge this change as it is? It would help reduce the number of merge conflicts due to changes in the configuration files. The changes for elasticsearch, redis and mysql to add access format, would not require a config migration - as their format parameter is already added - we will just be adding an additional allowed value for the format parameter.

donatello:
Added MySQL "access" format support. It is very similar to PostgreSQL.

I would prefer to do Redis and Elasticsearch in different PRs as they are quite different in terms of testing. 

For Redis:

1. Need to change the stored internal representation from golang printf encoding to JSON/MsgPack/BSON (@harshavardhana @abperiasamy decision pending here)
2. For namespace mode we will use a hash type (key=bucket and object, value is encoded representation of event data.
3. For access mode it seems appropriate to either use a redis list (allows traversal in the order of events getting added, but keeps growing unless user deletes from the list), or redis pub/sub (this drops events once they are delivered to all listeners, but is similar to listen-bucket-notifications). I lean towards using a redis-list where each item is heterogeneous list - [`timestamp-string`, `event-data-object`] - encoded in JSON/MsgPack.

For Elasticsearch:

1. It may be time to switch to the [newest version](https://github.com/olivere/elastic#releases) of the library so that we can support the latest major release line of elasticsearch. However, this could break some backwards compatibility.
2. In access format, in ES, we will not set the `Id` and let ES auto-generate that. We can set explicit time stamp as the event time.

harshavardhana:
> Need to change the stored internal representation from golang printf encoding to JSON/MsgPack/BSON (@harshavardhana @abperiasamy decision pending here)

We decided to do hmset, hmget style with native redis representation. So in case of namespace the unique top level key is bucket+object, whereas in case of access this unique key is either unix time stamp etc (leaving this upto you).  redis list is better for access logs. 

> It may be time to switch to the newest version of the library so that we can support the latest major release line of elasticsearch. However, this could break some backwards compatibility.
> In access format, in ES, we will not set the Id and let ES auto-generate that. We can set explicit time stamp as the event time.

Can you validate if there is any migration that needs to be done, if there is we can reference it in our docs and move to the latest. :+1:

harshavardhana:
ping @balamurugana - please review this patch.

donatello:
Just squashed commits and cleaned up commit message. It is ready for review and then merge.

cmd/config-migrate.go
```diff
@@ -1082,3 +1086,124 @@ func migrateV15ToV16() error {
...
+	}
+	if len(cv16.Notify.ElasticSearch) == 0 {
+		srvConfig.Notify.ElasticSearch = make(map[string]elasticSearchNotify)
+		srvConfig.Notify.ElasticSearch["1"] = elasticSearchNotify{}
+	} else {
+		// IMPORTANT NOTE: Future migrations should remove
+		// this as existing configuration will already contain
+		// a value for the "format" parameter.
+		for k, v := range cv16.Notify.ElasticSearch.Clone() {
+			v.Format = "namespace"
```
krisis: nit: We could make the format types as string constants like,
```go
const (
  nsFormat = "namespace"
  accessFormat = "access"
)
```

cmd/notify-mysql.go
```diff
@@ -86,21 +126,24 @@ func (m *mySQLNotify) Validate() error {
 	if !m.Enable {
 		return nil
 	}
+	if m.Format != "namespace" && m.Format != "access" {
```
krisis: We should use string constants for "namespace" and "access".

docs/bucket/notifications/README.md
```diff
@@ -372,36 +372,72 @@ go run nats.go
 <a name="PostgreSQL"></a>
 ## Publish Minio events via PostgreSQL
 
-Install PostgreSQL from [here](https://www.postgresql.org/).
+Install [PostgreSQL](https://www.postgresql.org/) database server. For illustrative pruposes, we have set the "postgres" user password as `password` and created a database called `minio_events` to store the events.
```
krisis: s/pruposes/purposes/

harshavardhana:
ping @balamurugana can you review this?

krisis (commented):
LGTM, only minor comments and a typo.

cmd/config-migrate.go
donatello: Done - added as `formatNamespace` and `formatAccess`.

cmd/notify-mysql.go
donatello: Done - added as `formatNamespace` and `formatAccess`.

docs/bucket/notifications/README.md
donatello: Done!

krisis (approved):

docs/bucket/notifications/README.md
```diff
@@ -372,36 +372,72 @@ go run nats.go
...
+The steps below show how to use this notification target in `namespace` format. The other format is very similar and is omitted for brevity.
 
 ### Step 1: Add PostgreSQL endpoint to Minio
 
-The default location of Minio server configuration file is ``~/.minio/config.json``. Update the PostgreSQL configuration block in ``config.json`` as follows:
+The default location of Minio server configuration file is ``~/.minio/config.json``. The PostgreSQL configuration is located in the `postgresql` key under the `notify` top-level key. Create a configuration key-value pair here for your PostgreSQL instance. The key is a name for your PostgreSQL endpoint, and the value is a collection of key-value parameters described in the table below.
+
+| Parameter | Value type | Description |
+|:---|:---|:---|
+| `enable` | Boolean | (Required) Is this server endpoint configuration active/enabled? |
```
harshavardhana: This value can be represented with Go's native type. as `_bool_` and `_string_` .  Also you can just call 'Value type' as just 'Type'

docs/bucket/notifications/README.md
```diff
@@ -479,7 +525,7 @@ With the `mc` tool, the configuration is very simple to add. Let us say that the
 # Create bucket named `images` in myminio
 mc mb myminio/images
 # Add notification configuration on the `images` bucket using the MySQL ARN. The --suffix argument filters events.
-mc events add  myminio/images arn:minio:sqs:us-east-1:1:postgresql --suffix .jpg
+mc events add myminio/images arn:minio:sqs:us-east-1:1:postgresql --suffix .jpg
```
harshavardhana, line 528: BTW also update the `docs/config/README.md` with the new config v17.

harshavardhana (changes requested):

docs/bucket/notifications/README.md
donatello: Done.

docs/bucket/notifications/README.md
donatello: Not sure what needs to be done here. The doc has link to configuration about mysql/postgresql and those links have the updated information (updated in this PR).

harshavardhana: Just edit the file from v16 to v17 ?

donatello: Ah got it. Done.

harshavardhana (approved):

PR #3963, opened 2017-03-23
browser-flag: wrapped bool type denotes browser on/off flag.

balamurugana:
## How Has This Been Tested?

`make test`

## Types of changes

## Checklist:

### DIFF
--- a/cmd/browser-flag.go
+++ b/cmd/browser-flag.go
@@ -1,0 +1,68 @@
+/*
+ * Minio Cloud Storage, (C) 2017 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cmd
+
+import (
+	"encoding/json"
+	"fmt"
+)
+
+// BrowserFlag - wrapper bool type.
+type BrowserFlag bool
+
+// String - returns string of BrowserFlag.
+func (bf BrowserFlag) String() string {
+	if bf {
+		return "on"
+	}
+
+	return "off"
+}
+
+// MarshalJSON - converts BrowserFlag into JSON data.
+func (bf BrowserFlag) MarshalJSON() ([]byte, error) {
+	return json.Marshal(bf.String())
+}
+
+// UnmarshalJSON - parses given data into BrowserFlag.
+func (bf *BrowserFlag) UnmarshalJSON(data []byte) (err error) {
+	var s string
+	if err = json.Unmarshal(data, &s); err == nil {
+		b := BrowserFlag(true)
+		if s == "" {
+			// Empty string is treated as valid.
+			*bf = b
+		} else if b, err = ParseBrowserFlag(s); err == nil {
+			*bf = b
+		}
+	}
+
+	return err
+}
+
+// ParseBrowserFlag - parses string into BrowserFlag.
+func ParseBrowserFlag(s string) (bf BrowserFlag, err error) {
+	if s == "on" {
+		bf = true
+	} else if s == "off" {
+		bf = false
+	} else {
+		err = fmt.Errorf("invalid value ‘%s’ for BrowserFlag", s)
+	}
+
+	return bf, err
+}
--- a/cmd/browser-flag_test.go
+++ b/cmd/browser-flag_test.go
@@ -1,0 +1,137 @@
+/*
+ * Minio Cloud Storage, (C) 2017 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cmd
+
+import (
+	"errors"
+	"testing"
+)
+
+// Test BrowserFlag.String()
+func TestBrowserFlagString(t *testing.T) {
+	var bf BrowserFlag
+
+	testCases := []struct {
+		flag           BrowserFlag
+		expectedResult string
+	}{
+		{bf, "off"},
+		{BrowserFlag(true), "on"},
+		{BrowserFlag(false), "off"},
+	}
+
+	for _, testCase := range testCases {
+		str := testCase.flag.String()
+		if testCase.expectedResult != str {
+			t.Fatalf("expected: %v, got: %v", testCase.expectedResult, str)
+		}
+	}
+}
+
+// Test BrowserFlag.MarshalJSON()
+func TestBrowserFlagMarshalJSON(t *testing.T) {
+	var bf BrowserFlag
+
+	testCases := []struct {
+		flag           BrowserFlag
+		expectedResult string
+	}{
+		{bf, `"off"`},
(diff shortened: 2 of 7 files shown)


harshavardhana (commented):
Why is this kind of change needed? what benefit does a boolean type provide here?

balamurugana:
> Why is this kind of change needed? what benefit does a boolean type provide here?

Statically typed `BrowserFlag` prevents any arbitrary string value usage.  The wrapped bool marshal/unmarshal JSON according to the typed value.

harshavardhana:
> Statically typed BrowserFlag prevents any arbitrary string value usage. The wrapped bool marshal/unmarshal JSON according to the typed value.

I feel that this change is not useful, any arbitrary string can be easily validated by just looking at the string value, a typed approach means your in-memory representation is different than the one of the disk. Also just a special type for this value doesn't make sense. 

Since MINIO_BROWSER is expected to have only `on` or `off` i am thinking that it is valid to represent the in-memory value as enum

```
type BrowserFlag string 
const {
       BrowserOff BrowserFlag = "off"
       BrowserOn BrowserFlag = "on"
}
```

Writing custom marshal and unmarshal is not needed.. we don't have such a complex requirement just for browser flag.

balamurugana:
> I feel that this change is not useful, any arbitrary string can be easily validated by just looking at the string value, a typed approach means your in-memory representation is different than the one of the disk. Also just a special type for this value doesn't make sense.
> 

Why in-memory value should be same as the disk?

> Since MINIO_BROWSER is expected to have only on or off i am thinking that it is valid to represent the in-memory value as enum
> 
> type BrowserFlag string 
> const {
>        BrowserOff BrowserFlag = "off"
>        BrowserOn BrowserFlag = "on"
> }
> 

How extended `string` type prevents setting other than `on` and `off` in any situation? As you want to static type to allow only boolean values `on` and `off`, `bool` is the type we should extend.

> Writing custom marshal and unmarshal is not needed.. we don't have such a complex requirement just for browser flag.

I couldn't able to understand what complexity with this approach.

harshavardhana:
> How extended string type prevents setting other than on and off in any situation? As you want to static type to allow only boolean values on and off, bool is the type we should extend.

It only helps but keeping it as boolean is better since it itself can hold only maximum two different types of values. 

> I couldn't able to understand what complexity with this approach.

The choice is not clear why convert from one form to another.  My only concern is that It keeps the code simple nothing special to understand, no new types of follow. 

It is a very simple check IMHO to see if browser value is either "on" or "off" . 

```
func isValidBrowserValue(browser string)  bool {
     return strings.EqualFold(browser) == BrowserOn || strings.EqualForld(browser) == BrowserOff
}
```

Parsing it from one form to another writing a custom JSON marshal looks clearly complicated for a simple situation which could have been done in few lines.  I don't see how the wrapped type approach communicates any better to someone reading the code. 

Clearly you must have found a case which current code doesn't fit and i just want to know why - Is there a case where some odd value is given and minio server still starts fine without any validation? 

Another thing this patch is doing is forcing `browser` to be have a value which is not the expected situation, currently it is written such that if browser value is empty it is treated as if it is `on` , even if the `browser` param is missing in the config it is treated as `on`.

balamurugana:
> It only helps but keeping it as boolean is better since it itself can hold only maximum two different types of values.

Thats what we need.  The type itself prevents to store other than two states.



> The choice is not clear why convert from one form to another.

I don't see any type conversion happening.  `BooleanFlag` represents true as `"on" ` and false as `"off"` strings in JSON like the way it is now.



> func isValidBrowserValue(browser string) bool { return strings.EqualFold(browser) == BrowserOn || strings.EqualForld(browser) == BrowserOff }

> Parsing it from one form to another writing a custom JSON marshal looks clearly complicated for a simple situation which could have been done in few lines. I don't see how the wrapped type approach communicates any better to someone reading the code.

My question is whether we want to store specific values only or check and fail later.




> Clearly you must have found a case which current code doesn't fit and i just want to know why - Is there a case where some odd value is given and minio server still starts fine without any validation?

This PR is not a fix.  It prevents any failure now and then.



> Another thing this patch is doing is forcing browser to be have a value which is not the expected situation, currently it is written such that if browser value is empty it is treated as if it is on , even if the browser param is missing in the config it is treated as on.

This can be fixed.  But we want to work without `browser` element not with wrong value like empty string.  Isn't this wanted?

harshavardhana:
> This can be fixed. But we want to work without browser element not with wrong value like empty string. Isn't this wanted?

Yes empty either the field exists or not if its just empty string we just treat it as `on` its not invalid. 

> My question is whether we want to store specific values only or check and fail later.

Either way is fine, but i found to be an overkill to bring a new type. You could make it a generic boolean flag   type so that it can be used elsewhere as well. Not just `BrowserFlag`

balamurugana:
> Yes empty either the field exists or not if its just empty string we just treat it as on its not invalid.

This can be done, but see the below about generic boolean type

> Either way is fine, but i found to be an overkill to bring a new type.

Advantage of typed system is to prevent storing other than allowed values.  Even though browser behavior is a boolean, but we represent it as string in JSON  `"on" / "off"`.  If we want to keep it as string in config `struct`, it opens way to store unknown value, even empty string.


> You could make it a generic boolean flag type so that it can be used elsewhere as well. Not just BrowserFlag

What do you mean by generic boolean flag?  Does it mean boolean true is one of JSON value `"on", "enable", "yes"` etc and false is `"off", "disable", "no"` etc?

In browser flag case, empty string is treated as `true` but some other cases it would be wanted as `false` or error.

harshavardhana:
> What do you mean by generic boolean flag? Does it mean boolean true is one of JSON value "on", "enable", "yes" etc and false is "off", "disable", "no" etc?
> 
> In browser flag case, empty string is treated as true but some other cases it would be wanted as false or error.

Meaning like this .. 

```go
// Boolean - wrapper bool type.
type Boolean bool

// String - returns string of Boolean.
func (bf Boolean) String() string {
	if bf {
		return "on"
	}

	return "off"
}

// MarshalJSON - converts Boolean into JSON data.
func (bf Boolean) MarshalJSON() ([]byte, error) {
	return json.Marshal(bf.String())
}

// UnmarshalJSON - parses given data into Boolean.
func (bf *Boolean) UnmarshalJSON(data []byte) (err error) {
	var s string
	if err = json.Unmarshal(data, &s); err == nil {
		var b Boolean
		if b, err = ParseBoolean(s); err == nil {
			*bf = b
		}
	}

	return err
}

var ErrInvalidBoolean = fmt.Errorf("invalid value ‘%s’ for Boolean", s)
var ErrEmptyBoolean = errors.New("Boolean value is empty")

// ParseBoolean - parses string into Boolean type.
func ParseBoolean(s string) (bf Boolean, err error) {
	if s == "on" {
		bf = true
	} else if s == "off" {
		bf = false
	} else s == "" {
                err = ErrEmptyBoolean
	} else {
                err = ErrInvalidBoolean
        }
	return bf, err
}
```

Then use it 

```go
// This is specific to our requirements
func isValidBrowserFlag(flag string) (bool, error) {
       ok, err := ParseBoolean(flag) 
       if err == nil {
            return ok, nil
       } // else {
       // We treat empty flag as valid and treat it as true.
       if err == ErrEmptyBoolean {
              ok = true
              return ok, nil
        }
        return false, err   
}
```

balamurugana:
1. `func (bf *Boolean) UnmarshalJSON(data []byte) (err error)` fails on empty string for `"browser"` element in JSON.  Is this wanted? Previously you said you want empty string as `true`

2. Boolean type is not generic that it works `"on"` and `"off`"  values ie it doesn't fit if `"enable"`/`"disable"` values for a flag.

3. The generic name is not a good idea where it covers for browser on/off flag.  I wouldn't recommend this approach.

4. Currently I don't see any other requirement of bool wrapper.  If we encounter such situation we think about making generic.

harshavardhana:
I give up, perhaps i don't see how this change really helps. It doesn't add anything which present code doesn't do.

But in any case empty string either field exists or not, should behave how it behaves right now. This behavior we shouldn't change or throw an error if `brower: "".

balamurugana:
> I give up, perhaps i don't see how this change really helps. It doesn't add anything which present code doesn't do.

Current code is not bug free WRT `Browser` field in `serverConfig` struct. Interesting two methods are `GetBrowser()` and `SetBrowser()`.  No validation is required in them.


> But in any case empty string either field exists or not, should behave how it behaves right now. This behavior we shouldn't change or throw an error if `brower: "".

Done

harshavardhana:
> Current code is not bug free WRT Browser field in serverConfig struct. Interesting two methods are GetBrowser() and SetBrowser(). No validation is required in them.

Okay Agreed.

harshavardhana (approved):

PR #3964, opened 2017-03-23
server-mux: Return 503 when reqs num hits the max

### LINKED ISSUE #3883: server: Minio crashed due to thread exhaustion under high load.
here is the log 

```
runtime: program exceeds 10000-thread limit
fatal error: thread exhaustion

runtime stack:
runtime.throw(0xca2800, 0x11)
        /home/builder/.gimme/versions/go1.7.5.linux.amd64/src/runtime/panic.go:566 +0x95
runtime.checkmcount()
        /home/builder/.gimme/versions/go1.7.5.linux.amd64/src/runtime/proc.go:486 +0xa4
runtime.mcommoninit(0xc42b94b000)
        /home/builder/.gimme/versions/go1.7.5.linux.amd64/src/runtime/proc.go:506 +0xd5
runtime.allocm(0xc42002ca00, 0xd11508, 0x7efe00000001)
        /home/builder/.gimme/versions/go1.7.5.linux.amd64/src/runtime/proc.go:1286 +0x9b
runtime.newm(0xd11508, 0xc42002ca00)
        /home/builder/.gimme/versions/go1.7.5.linux.amd64/src/runtime/proc.go:1555 +0x39
runtime.startm(0xc42002ca00, 0x100000001)
        /home/builder/.gimme/versions/go1.7.5.linux.amd64/src/runtime/proc.go:1642 +0x181
runtime.wakep()
        /home/builder/.gimme/versions/go1.7.5.linux.amd64/src/runtime/proc.go:1723 +0x57
runtime.newproc1(0xd100b8, 0xc4202bae40, 0x10, 0x56d0d5, 0x7f1487ac8d78)
        /home/builder/.gimme/versions/go1.7.5.linux.amd64/src/runtime/proc.go:2810 +0x334
runtime.newproc.func1()
        /home/builder/.gimme/versions/go1.7.5.linux.amd64/src/runtime/proc.go:2712 +0x51
runtime.systemstack(0xc420027500)
        /home/builder/.gimme/versions/go1.7.5.linux.amd64/src/runtime/asm_amd64.s:298 +0x79
runtime.mstart()
        /home/builder/.gimme/versions/go1.7.5.linux.amd64/src/runtime/proc.go:1079

goroutine 26 [running]:
runtime.systemstack_switch()
        /home/builder/.gimme/versions/go1.7.5.linux.amd64/src/runtime/asm_amd64.s:252 fp=0xc4202bade8 sp=0xc4202bade0
runtime.newproc(0x10, 0xd100b8)
        /home/builder/.gimme/versions/go1.7.5.linux.amd64/src/runtime/proc.go:2713 +0x8b fp=0xc4202bae30 sp=0xc4202bade8
github.com/minio/minio/cmd.newListenerMux.func1(0xc4203c6090)
        /home/builder/mygo/src/github.com/minio/minio/cmd/server-mux.go:263 +0x2c5 fp=0xc4202bafa8 sp=0xc4202bae30
runtime.goexit()
```

1.
2.
3.
4.

## Your Environment

* Version used:
VERSION:
  2017-02-16T01:47:30Z

* Server type and version:
* Link to your project:

vadmeste:
## Description
Parallel requests should have a maximum number to avoid
crashing our server when we have too many of them. Go, at least,
intentionally crashes the server when we hit 10k threads.

## Motivation and Context
Fixes #3883 

## How Has This Been Tested?
go test

## Types of changes

## Checklist:

### DIFF
--- a/cmd/server-mux.go
+++ b/cmd/server-mux.go
@@ -32,6 +32,10 @@ import (
 
 const (
 	serverShutdownPoll = 500 * time.Millisecond
+
+	// Maximum number of concurrent connections, to avoid creating
+	// too many threads in the OS and get killed by golang runtime
+	serverMaxConcurrentConns = 9000
 )
 
 // The value chosen below is longest word chosen
@@ -331,6 +335,9 @@ type ServerMux struct {
 
 	// Current number of concurrent http requests
 	currentReqs int32
+	// Max number of concurrent http requests
+	maxConcurrentReqs int32
+
 	// Time to wait before forcing server shutdown
 	gracefulTimeout time.Duration
 
@@ -341,8 +348,9 @@ type ServerMux struct {
 // NewServerMux constructor to create a ServerMux
 func NewServerMux(addr string, handler http.Handler) *ServerMux {
 	m := &ServerMux{
-		Addr:    addr,
-		handler: handler,
+		Addr:              addr,
+		handler:           handler,
+		maxConcurrentReqs: serverMaxConcurrentConns,
 		// Wait for 5 seconds for new incoming connnections, otherwise
 		// forcibly close them during graceful stop or restart.
 		gracefulTimeout: 5 * time.Second,
@@ -465,6 +473,12 @@ func (m *ServerMux) ListenAndServe(certFile, keyFile string) (err error) {
 			closing := m.closing
 			m.mu.RUnlock()
 			if closing {
+				w.WriteHeader(http.StatusServiceUnavailable)
+				return
+			}
+
+			// Return HTTP service unavailable when we have reached maximum concurrent HTTP requests.
+			if atomic.LoadInt32(&m.currentReqs) >= m.maxConcurrentReqs {
 				w.WriteHeader(http.StatusServiceUnavailable)
 				return
 			}


cmd/server-rlimit.go
```diff
@@ -30,6 +34,9 @@ func setMaxResources() (err error) {
 		return err
 	}
 
+	// Set new max threads limit
+	increaseMaxThreads(int(maxLimit))
```
krisis: shouldn't we check if `increaseMaxThreads` returns `maxLimit` and return error if it doesn't?

krishnasrinivas:
https://github.com/minio/minio/issues/3883 :

> Yes. we have about 4000-5000 concurrent uploads going on at the same time, and about 4-5 mts. Minio service crashed. could you please let me know how to increase the thread ?

for 5000 concurrent uploads, assuming all the goroutines are blocking, there would be roughly 5000 threads, right? but if the thread count is 10000 then there must be some other underlying problem? does the thread count come back to original number once the concurrent uploads stops? (when tested with a high concurrent upload count not enough to crash minio i.e)

cmd/server-rlimit.go
vadmeste: > shouldn't we check if increaseMaxThreads returns maxLimit and return error if it doesn't?

Actually no. maxLimit could be less than the original value of MaxThreads. This function is not supposed to fail.

vadmeste:
> for 5000 concurrent uploads, assuming all the goroutines are blocking, there would be roughly 5000 threads, right?

No. In addition to the thread which is processing the client request, the server creates an additional goroutine here  https://github.com/minio/minio/blob/master/cmd/server-mux.go#L238. I know this latter is supposed to die but if I test in my local machine, I can see threads number increasing to roughly the double.

How I tested ?

1. I added a simple goroutine in Minio server which prints number of threads each 500 ms https://golang.org/pkg/runtime/#NumGoroutine

2. I uploaded a 1kb file to the server and made it public.

3. I run this command with httperf `httperf --ssl --server=127.0.0.1 --port=9000 --hog --timeout 300.0 --num-conns 5000 --rate=5000 --uri=/testbucket/public/file`

httperf shows me that the server was able to treat around 400 requests per seconds, at the same time, I can see the number of threads reaching until 800 or 1000.

harshavardhana (commented):
Lets bring the change of blocking connections or rejecting new connections also in this patch. 

So that we can avoid run-away effect of crashing the server.

vadmeste:
> Lets bring the change of blocking connections or rejecting new connections also in this patch.
> 
> So that we can avoid run-away effect of crashing the server.

:+1:, will do.

vadmeste:
I've added a new commit which returns http 503 / service unavailable when the number of concurrent requests reaches 9k, this is less than 10k.

By the way, I was wrong interpreting the problem in the first place. Having 4000 parallels connections means we should have 8000 maximum goroutines but not threads. Number of threads is usually less than the number of goroutines, especially when many goroutines are blocked. But anyway, this patch will protect Minio from all use cases I can think of.

krishnasrinivas:
> By the way, I was wrong interpreting the problem in the first place. Having 4000 parallels connections means we should have 8000 maximum goroutines but not threads. Number of threads is usually less than the number of goroutines, especially when many goroutines are blocked. But anyway, this patch will protect Minio from all use cases I can think of.

@vadmeste hence my previous comment:
```
for 5000 concurrent uploads, assuming all the goroutines are blocking, there would be roughly 5000 threads, right? but if the thread count is 10000 then there must be some other underlying problem? does the thread count come back to original number once the concurrent uploads stops? (when tested with a high concurrent upload count not enough to crash minio i.e)
```

Hope there is no underlying problem in minio causing high thread count.
Does the thread count come back to normalcy after the load completes?

cmd/server-rlimit.go
```diff
@@ -39,6 +46,21 @@ func setMaxResources() (err error) {
...
+// go routines if it is greater than the old one.
+func increaseMaxThreads(max int) int {
+	if max <= 0 {
+		return max
+	}
+	// There is no way to know the current number of max threads but
+	// only after setting the new number. If the latter turned out
+	// to be less than the old one, revert the operation.
+	if oldMax := debug.SetMaxThreads(max); oldMax > max {
+		return debug.SetMaxThreads(oldMax)
```
balamurugana: As per https://golang.org/pkg/runtime/debug/#SetMaxThreads, the default value is 10,000 threads. If max value is < 10,000, you reset back to 10,000 threads.

In the above `maxLimit` of `sys.GetMaxOpenFileLimit()` is used which is 4096 in Linux by default.   In this case, the function sets back 10,000 again.  In MS Windows, max open file limit is not applicable.

I think this function doesn't work practically.

Have you tried with other *nix?

krishnasrinivas:
@vadmeste you can undo the increaseMaxThreads related changes in the previous commit

vadmeste:
@krishnasrinivas, actually there is a difference between number of goroutines and number of threads. I've tested the following example that I wanted to share with you:

sample.go: 

```
package main

import (
        "fmt"
        "sync"
        "time"
)

func main() {

        wg := sync.WaitGroup{}

        for i := 0; i < 20000; i++ {
                wg.Add(1)
                go func() {
                        defer wg.Done()
                        time.Sleep(2 * 60 * time.Second)
                }()
        }
        fmt.Println("waiting for all finishing sleeping..")
        wg.Wait()
}
```

`$ go build sample.go`
`$ ./sample`

sample will create 20k goroutines sleep for about 2 minutes.

In another term:
`$ cat /proc/$(pidof sample)/status | grep Threads`

You can notice that the number of threads is very low. It is 7 threads in my case.

cmd/server-rlimit.go
vadmeste: > In the above maxLimit of sys.GetMaxOpenFileLimit() is used which is 4096 in Linux by default. In this case, the function sets back 10,000 again.

In my case, maxLimit is 65k, (ubuntu 16.04, 64 bits)

> Have you tried with other *nix?

Actually no, that's why this code is generic and does revert SetMaxThreads when need.

vadmeste:
> Does the thread count come back to normalcy after the load completes?

Number of threads desn't come back to normal but it was always low (~ 30 threads)

vadmeste:
> @vadmeste you can undo the increaseMaxThreads related changes in the previous commit

Done

krishnasrinivas (approved):

cmd/server-mux.go
```diff
@@ -31,7 +31,8 @@ import (
 )
 
 const (
-	serverShutdownPoll = 500 * time.Millisecond
+	serverShutdownPoll     = 500 * time.Millisecond
+	serverMaxParallelConns = 9000
```
harshavardhana: Can you add a comment on why this is chosen, perhaps even reference the github issue here for understanding.

cmd/server-mux.go
```diff
@@ -331,6 +332,9 @@ type ServerMux struct {
 
 	// Current number of concurrent http requests
 	currentReqs int32
+	// Max number of concurrent http requests
+	maxParallelReqs int32
```
harshavardhana: maxConcurrentReqs perhaps is better.. than parallel.

cmd/server-mux.go
```diff
@@ -469,6 +474,13 @@ func (m *ServerMux) ListenAndServe(certFile, keyFile string) (err error) {
 				return
 			}
 
+			// Return ServiceUnavailable when we hit a limit of number
```
harshavardhana: This is better --> `// Return HTTP service unavailable when we have reached maximum concurrent HTTP requests.`

harshavardhana (changes requested):

vadmeste:
@harshavardhana, done

harshavardhana (approved):

harshavardhana:
We should close this PR @vadmeste - we don't need this. We can bring this back as needed.

PR #3969, opened 2017-03-24
[WIP] browser: Update UI with new components and elements

rushenn:
## Description
New UI for browser
ui

**Todo**
- Image preview
- Mobile support (Responsive)
- IE testing

## Types of changes

## Checklist:


vadmeste:
:1st_place_medal:

harshavardhana (approved):

krishnasrinivas:
@harshavardhana i could send a separate fix for `the svgs are returning index.html` problem.

harshavardhana:
> @harshavardhana i could send a separate fix for the svgs are returning index.html problem.

Rushen already fixed it.. i tested can you test on your end?

krishnasrinivas:
> Rushen already fixed it.. i tested can you test on your end?

ah, will do.

README.md
```diff
@@ -77,9 +77,7 @@ You can run Minio on FreeBSD with FreeNAS storage-backend - see [here](https://d
 Source installation is only intended for developers and advanced users. If you do not have a working Golang environment, please follow [How to install Golang](https://docs.minio.io/docs/how-to-install-golang).
```
harshavardhana: This change is unrelated to this can you rebase properly locally @rushenn ?

harshavardhana (changes requested):

rushenn:
@harshavardhana Rebased, please check.

harshavardhana (approved):

krishnasrinivas:
path


Previously `>` was `/` which looks more intuitive as a path separator.

i.e the path would look as `test / 1 / 2 / 3`

krishnasrinivas:
click-outside1

Here it is not obvious that you have to click somewhere outside to remove the "share-url" and "delete" buttons from the view.

Older style:


click-outside2

Here it is obvious that you have to click outside .... as the "share-url" and "delete" buttons looks overlaid on a background.

krishnasrinivas:
create

On mouse-over it should show a tool tip like how it was previously. Right now the icons are not clear on its purpose.

krishnasrinivas:
policy

Previously we had three dots and it was intuitive that you had to click it for more options. Now it is not intuitive that you have to click the circle icon (marked in the image)

krishnasrinivas:
bucketlist

Will it be better to darken the background of the bucket selected as well?

krishnasrinivas:
miniobrowser1
miniobrowser2


does the lower case look better?

krishnasrinivas:
topicons

Needs tooltip

krishnasrinivas:
select-file

It will be better to change the file-select icon to circle when the mouse is moved over the icon, now it changes to circle if the mouse is anywhere on the row.

krishnasrinivas:
signout

you can use the `sign-out` icon instead of `power-off` icon

rushenn:
@krishnasrinivas Thank you, I will make sure to include these fixes as part of my next push.

rushenn:
> It will be better to change the file-select icon to circle when the mouse is moved over the icon, now it changes to circle if the mouse is anywhere on the row.

If the checkbox toggle is set to the icon only, it won't be noticed much. Since the checkboxes are hidden by default, i think it's better to keep the toggle set to the entire row.

deekoder:
Update PR with responsiveness changes.

PR #3972, opened 2017-03-25
config: Appropriate error message when newer config file is found

krishnasrinivas:
## Description

If newer config file is run with older binary it would do:
errors.New("bad config version, expected: " + v16)

which is a confusing error message. Provide a better error message of:
`"Expected config version: %s, newer config version found: %s"`
so that the user gets a hint on what is happening.

## How Has This Been Tested?

manually

## Types of changes

## Checklist:

### DIFF
--- a/cmd/config-v17.go
+++ b/cmd/config-v17.go
@@ -215,9 +215,10 @@ func validateConfig() error {
 		return err
 	}
 
-	// Check if config version is valid
 	if srvCfg.Version != v17 {
-		return errors.New("bad config version, expected: " + v17)
+		// Older binary but newer config version.
+		// Config version can never be older as it would have migrated.
+		return fmt.Errorf("Expected config version: %s, newer config version found: %s", v17, srvCfg.Version)
 	}
 
 	// Load config file json and check for duplication json keys
--- a/cmd/server-main.go
+++ b/cmd/server-main.go
@@ -169,7 +169,7 @@ func initConfig() {
 
 	// Validate config file
 	err = validateConfig()
-	fatalIf(err, "Cannot validate configuration file")
+	fatalIf(err, "Unable to validate configuration file")
 
 	// Once we have migrated all the old config, now load them.
 	err = loadConfig(envs)


harshavardhana:
Looks like build failure @krishnasrinivas

```
/home/travis/gopath/src/github.com/minio/minio/cmd/config-v16.go:213:10: should replace errors.New(fmt.Sprintf(...)) with fmt.Errorf(...)
Found 1 lint suggestions; failing.
make: *** [lint] Error 1
The command "make" exited with 2.
```

cmd/config-v16.go
```diff
@@ -208,6 +208,11 @@ func validateConfig() error {
 		return err
 	}
 
+	if srvCfg.Version > v16 {
```
vadmeste: Actually, this is a string comparison, not integer.

By the way, since we are always migrating older versions, we will never have the case srvCfg.Version < v16. I think this could be merge with the next if condition.

harshavardhana: ping @krishnasrinivas

krishnasrinivas: > Actually, this is a string comparison, not integer.

`It is usually clearer and always faster to use the built-in string comparison operators ==, <, >, and so on.`

https://golang.org/pkg/strings/#Compare

> By the way, since we are always migrating older versions, we will never have the case srvCfg.Version < v16. I think this could be merge with the next if condition.

I will retain the `!=` itself then.

vadmeste: > It is usually clearer and always faster to use the built-in string comparison operators ==, <, >, and so on.
> 
> https://golang.org/pkg/strings/#Compare

Yeah but with string comparaison, "128" < "20", this is what I meant.

krishnasrinivas: @vadmeste 
```
        if srvCfg.Version != v16 {                                                                                                                                                           
                // Older binary but newer config version.                                                                                                                                    
                return fmt.Errorf("Expected config version: %s, newer config version found: %s", v16, srvCfg.Version)                                                                        
        }
```

Even though srvCfg.Version is never less than v16 the above code affects readability i.e comparison is inequality `if srvCfg.Version != v16` but the error message is `newer config version found: %s` 

So I think the style followed in this commit is better

krishnasrinivas: > Yeah but with string comparaison, "128" < "20", this is what I meant.

ah! will do it `!=` way then and add a comment.

cmd/config-v16.go
```diff
@@ -208,9 +208,10 @@ func validateConfig() error {
 		return err
 	}
 
-	// Check if config version is valid
```
harshavardhana: Should be now v17.

harshavardhana (changes requested):

harshavardhana (approved):

harshavardhana (approved):

PR #3973, opened 2017-03-25
build: Improve build messaging, say where we built Minio.

harshavardhana:
## Description
Improve build messaging and say where we built minio in.

## Motivation and Context
Currently we don't indicate where the binary is built into. Helpful for new comers.

## How Has This Been Tested?
Manually using `make`

## Types of changes

## Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,42 +56,55 @@ endif
 all: install
 
 checks:
-	@echo "Checking deps:"
+	@echo -n "Check deps: "
 	@(env bash $(PWD)/buildscripts/checkdeps.sh)
+	@echo "Done."
+	@echo -n "Checking project is in GOPATH: "
 	@(env bash $(PWD)/buildscripts/checkgopath.sh)
+	@echo "Done."
 
 getdeps: checks
-	@echo "Installing golint:" && go get -u github.com/golang/lint/golint
-	@echo "Installing gocyclo:" && go get -u github.com/fzipp/gocyclo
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -56,42 +56,55 @@ endif
...
 	@(env bash $(PWD)/buildscripts/checkgopath.sh)
+	@echo "Done."
 
 getdeps: checks
-	@echo "Installing golint:" && go get -u github.com/golang/lint/golint
-	@echo "Installing gocyclo:" && go get -u github.com/fzipp/gocyclo
-	@echo "Installing deadcode:" && go get -u github.com/remyoudompheng/go-misc/deadcode
-	@echo "Installing misspell:" && go get -u github.com/client9/misspell/cmd/misspell
-	@echo "Installing ineffassign:" && go get -u github.com/gordonklaus/ineffassign
+	@echo -n "Installing golint: " && go get -u github.com/golang/lint/golint
```
krisis, line 67: `go get` commands requires internet connectivity. This makes it impossible to use `make` when offline. Could we check if `golint` and similar binaries are already in `$PATH` before performing `go get`?

harshavardhana: How big of a problem is internet connectivity? also how many users are offline writing code?

krisis: I have faced this problem. I don't know if it bothers many. I can do without using `make` when offline.

krisis (approved):

PR #3975, opened 2017-03-26
server: handle command line and env variables at one place.

balamurugana:
## How Has This Been Tested?

- `make test`
- manually

## Types of changes

## Checklist:

### DIFF
--- a/cmd/certs.go
+++ b/cmd/certs.go
@@ -24,11 +24,6 @@ import (
 	"path/filepath"
 )
 
-// isSSL - returns true with both cert and key exists.
-func isSSL() bool {
-	return isFile(getPublicCertFile()) && isFile(getPrivateKeyFile())
-}
-
 func parsePublicCertFile(certFile string) (certs []*x509.Certificate, err error) {
 	var bytes []byte
 
@@ -58,11 +53,6 @@ func parsePublicCertFile(certFile string) (certs []*x509.Certificate, err error)
 	}
 
 	return certs, err
-}
-
-// Reads certificate file and returns a list of parsed certificates.
-func readCertificateChain() ([]*x509.Certificate, error) {
-	return parsePublicCertFile(getPublicCertFile())
 }
 
 func getRootCAs(certsCAsDir string) (*x509.CertPool, error) {
@@ -100,9 +90,19 @@ func getRootCAs(certsCAsDir string) (*x509.CertPool, error) {
 	return rootCAs, nil
 }
 
-// loadRootCAs fetches CA files provided in minio config and adds them to globalRootCAs
-// Currently under Windows, there is no way to load system + user CAs at the same time
-func loadRootCAs() (err error) {
-	globalRootCAs, err = getRootCAs(getCADir())
-	return err
+func getSSLConfig() (publicCerts []*x509.Certificate, rootCAs *x509.CertPool, secureConn bool, err error) {
+	if !(isFile(getPublicCertFile()) && isFile(getPrivateKeyFile())) {
+		return publicCerts, rootCAs, secureConn, err
+	}
+
+	if publicCerts, err = parsePublicCertFile(getPublicCertFile()); err != nil {
+		return publicCerts, rootCAs, secureConn, err
+	}
+
+	if rootCAs, err = getRootCAs(getCADir()); err != nil {
+		return publicCerts, rootCAs, secureConn, err
+	}
+
+	secureConn = true
+	return publicCerts, rootCAs, secureConn, err
 }
--- a/cmd/globals.go
+++ b/cmd/globals.go
@@ -109,6 +109,9 @@ var (
 	// Time when object layer was initialized on start up.
 	globalBootTime time.Time
 
+	globalActiveCred         credential
+	globalPublicCerts        []*x509.Certificate
+	globalXLObjCacheDisabled bool
 	// Add new variable global values here.
 )
 
--- a/cmd/server-main.go
+++ b/cmd/server-main.go
@@ -94,20 +94,6 @@ func checkUpdate(mode string) {
 	}
 }
 
-// envParams holds all env parameters
-type envParams struct {
-	creds   credential
-	browser BrowserFlag
-}
-
-func migrate() {
-	// Migrate config file
-	err := migrateConfig()
-	fatalIf(err, "Config migration failed.")
-
-	// Migrate other configs here.
-}
-
 func enableLoggers() {
 	fileLogTarget := serverConfig.Logger.GetFile()
 	if fileLogTarget.Enable {
@@ -122,77 +108,6 @@ func enableLoggers() {
 	}
 
 	log.SetConsoleTarget(consoleLogTarget)
-}
-
-// Initializes a new config if it doesn't exist, else migrates any old config
-// to newer config and finally loads the config to memory.
-func initConfig() {
-	accessKey := os.Getenv("MINIO_ACCESS_KEY")
-	secretKey := os.Getenv("MINIO_SECRET_KEY")
-
-	var cred credential
-	var err error
-	if accessKey != "" && secretKey != "" {
-		cred, err = createCredential(accessKey, secretKey)
-		fatalIf(err, "Invalid access/secret Key set in environment.")
-
-		// credential Envs are set globally.
-		globalIsEnvCreds = true
-	}
-
-	envs := envParams{
-		creds:   cred,
-		browser: BrowserFlag(true),
-	}
-
-	if browser := os.Getenv("MINIO_BROWSER"); browser != "" {
-		browserFlag, err := ParseBrowserFlag(browser)
-		if err != nil {
-			fatalIf(errors.New("invalid value"), "Unknown value ‘%s’ in MINIO_BROWSER environment variable.", browser)
-		}
-
-		// browser Envs are set globally, this does not represent
-		// if browser is turned off or on.
-		globalIsEnvBrowser = true
-		envs.browser = browserFlag
-	}
-
-	// Config file does not exist, we create it fresh and return upon success.
-	if !isConfigFileExists() {
-		err := newConfig(envs)
-		fatalIf(err, "Unable to initialize minio config for the first time.")
-		log.Println("Created minio configuration file successfully at " + getConfigDir())
-		return
-	}
-
-	// Migrate any old version of config / state files to newer format.
-	migrate()
-
-	// Validate config file
-	err = validateConfig()
-	fatalIf(err, "Unable to validate configuration file")
-
-	// Once we have migrated all the old config, now load them.
-	err = loadConfig(envs)
-	fatalIf(err, "Unable to initialize minio config")
-}
-
-// Generic Minio initialization to create/load config, prepare loggers, etc..
-func minioInit(ctx *cli.Context) {
-	// Create certs path.
-	fatalIf(createConfigDir(), "Unable to create \"certs\" directory.")
-
-	// Is TLS configured?.
-	globalIsSSL = isSSL()
-
-	// Initialize minio server config.
-	initConfig()
-
-	// Enable all loggers by now so we can use errorIf() and fatalIf()
-	enableLoggers()
-
-	// Init the error tracing module.
-	initError()
 }
 
 type serverCmdConfig struct {
@@ -238,18 +153,6 @@ func parseStorageEndpoints(eps []string) (endpoints []*url.URL, err error) {
 		endpoints = append(endpoints, u)
 	}
 	return endpoints, nil
-}
-
-// initServer initialize server config.
-func initServerConfig(c *cli.Context) {
-	// Initialization such as config generating/loading config, enable logging, ..
-	minioInit(c)
-
-	// Load user supplied root CAs
-	fatalIf(loadRootCAs(), "Unable to load a CA files")
-
-	// Set system resources to maximum.
-	errorIf(setMaxResources(), "Unable to change resource limit")
 }
 
 // Validate if input disks are sufficient for initializing XL.
@@ -344,21 +247,9 @@ func checkEndpointsSyntax(eps []*url.URL, disks []string) error {
 }
 
 // Make sure all the command line parameters are OK and exit in case of invalid parameters.
-func checkServerSyntax(c *cli.Context) {
-	serverAddr := c.String("address")
-
-	host, portStr, err := net.SplitHostPort(serverAddr)
-	fatalIf(err, "Unable to parse %s.", serverAddr)
-
-	// Verify syntax for all the XL disks.
-	disks := c.Args()
-
-	// Parse disks check if they comply with expected URI style.
-	endpoints, err := parseStorageEndpoints(disks)
-	fatalIf(err, "Unable to parse storage endpoints %s", strings.Join(disks, " "))
-
+func checkServerSyntax(endpoints []*url.URL, disks []string) {
 	// Validate if endpoints follow the expected syntax.
-	err = checkEndpointsSyntax(endpoints, disks)
+	err := checkEndpointsSyntax(endpoints, disks)
 	fatalIf(err, "Invalid endpoints found %s", strings.Join(disks, " "))
 
 	// Validate for duplicate endpoints are supplied.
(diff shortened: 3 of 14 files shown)


cmd/config-v16.go
```diff
@@ -131,15 +131,15 @@ func loadConfig(envParams envParams) error {
 
 	// If env is set override the credentials from config file.
 	if globalIsEnvCreds {
-		srvCfg.SetCredential(envParams.creds)
+		srvCfg.SetCredential(globalActiveCred)
+	} else {
+		globalActiveCred = srvCfg.GetCredential()
```
vadmeste: This makes a little confusion about whether we should use globalActiveCred or srvCfg.GetCredential() later. I think we should reduce global variables instead if possible.

vadmeste: What about writing a global structure which holds all our env variables ?

balamurugana:
> What about writing a global structure which holds all our env variables ?

I am already working on the patch. To make it work, there are several fixes required in current code base.  Look at https://github.com/minio/minio/pull/3643

cmd/config-v16.go
balamurugana: > This makes a little confusion about whether we should use globalActiveCred or srvCfg.GetCredential() later. I think we should reduce global variables instead if possible.

All functions should use `globalActiveCred` which is the right credential even though config file has different or same credential.  ATM I follow existing logic of having configuration in sync with env ie `config` struct contains replaced credentials values on load, so there is no change in other functions/method.  Eventually they will be using `globalActiveCred`.  My actual plan is to remove getter/setter of `config` structure wherever possible in down the line.

Spilling env or command line arguments information across code base is not a good idea.

I am favor of removing all globals. I already have a patch ready but the patch is huge and reviewing would be harder.  I split the patch into multiples to ease out for acceptance.  This PR is one of such kind.

cmd/generic-handlers.go
```diff
@@ -141,7 +141,7 @@ func setBrowserCacheControlHandler(h http.Handler) http.Handler {
 }
 
 func (h cacheControlHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	if r.Method == httpGET && guessIsBrowserReq(r) && globalIsBrowserEnabled {
+	if r.Method == httpGET && guessIsBrowserReq(r) && globalIsBrowserEnabled == true {
```
krishnasrinivas: isn't previous style more readable?

harshavardhana: Previous style is preferred..

balamurugana: Done

balamurugana:
> What about writing a global structure which holds all our env variables ?

I think it won't add any improvements.  I prefer removing all globals and put in a struct like
```go
// Setup minio setup information
type Setup struct {
	mutex       *sync.Mutex
	serverAddr  string
	endpoints   EndpointList
	setupType   SetupType
	configDir   string
	profileMode string
	cred        credential
	isEnvCred   bool
	profiler    interface {
		Stop()
	}
	isCacheDisabled   bool
	isBrowserDisabled bool

	serverConfig *serverConfigV14

	publicCrtFile  string
	privateKeyFile string
	publicCerts    []*x509.Certificate
	rootCAs        *x509.CertPool
	secureConn     bool

	bootTime time.Time
}
```

This will bring running unit/functional tests in parallel.

cmd/globals.go
```diff
@@ -109,6 +109,9 @@ var (
 	// Time when object layer was initialized on start up.
 	globalBootTime time.Time
 
+	globalActiveCred       credential
+	globalPublicCerts      []*x509.Certificate
+	globalObjCacheDisabled bool
```
krishnasrinivas: since this is XL specific, better put it in xlObjects

balamurugana: in `xlObjects` struct, there is `objCacheEnabled bool`.  I couldn't able to get what you mean by `put it in xlObjects`?

cmd/server-main.go
```diff
@@ -239,18 +154,6 @@ func parseStorageEndpoints(eps []string) (endpoints []*url.URL, err error) {
...
-// initServer initialize server config.
-func initServerConfig(c *cli.Context) {
-	// Initialization such as config generating/loading config, enable logging, ..
-	minioInit(c)
-
-	// Load user supplied root CAs
-	fatalIf(loadRootCAs(), "Unable to load a CA files")
-
-	// Set system resources to maximum.
-	errorIf(setMaxResources(), "Unable to change resource limit")
```
krishnasrinivas, line 252: `setMaxResources()` should be called from one of the init functions.

cmd/globals.go
krishnasrinivas: What I mean is, we should use `xlObjects.objCacheEnabled` and `globalObjCacheDisabled` should be removed

cmd/server-main.go
balamurugana: Done

cmd/globals.go
balamurugana: We would need to keep it as global because,

1. We would need to validate how multi-node minio is started.  This check needs to be done in post login (currently not).

2. I see flow to `newXLObjects()` is `serverMain()` -> `newObjectLayer()` -> `newXLObjectLayer()` -> `newXLObjects()`.  Passing `globalObjCacheDisabled` flag over these functions is not a good idea.

krishnasrinivas: > We would need to validate how multi-node minio is started. This check needs to be done in post login (currently not).

_MINIO_CACHE is an undocumented env-var mainly to be used by developers for testing erasure code (i.e so that we don't actually hit the cache and ignore the underlying problem)

Even otherwise _MINIO_CACHE is an minio process specific setting and hence need not be compared with how the peers are run with. i.e I should be able to run one server with caching and one without caching.

krishnasrinivas: > I see flow to newXLObjects() is serverMain() -> newObjectLayer() -> newXLObjectLayer() -> newXLObjects(). Passing globalObjCacheDisabled flag over these functions is not a good idea.

This style need not be done, we can just do how it was previously done in xl-v1.go:
```
objCacheDisabled := strings.EqualFold(os.Getenv("_MINIO_CACHE"), "off")
```

balamurugana:
> _MINIO_CACHE is an undocumented env-var mainly to be used by developers for testing erasure code (i.e so that we don't actually hit the cache and ignore the underlying problem)
> 
> Even otherwise _MINIO_CACHE is an minio process specific setting and hence need not be compared with how the peers are run with. i.e I should be able to run one server with caching and one without caching.

The cache is disabled either by `_MINIO_CACHE` or by < 8GB memory.  Do you say we support mixed cache mode of minio servers (some with cache enabled and some are not)?

balamurugana:
> This style need not be done, we can just do how it was previously done in xl-v1.go:
> 
> objCacheDisabled := strings.EqualFold(os.Getenv("_MINIO_CACHE"), "off")

Spilling env variable handling across code base is not a good idea.  This is not a good practice. Also It would lead to write complex functional/unit tests.

harshavardhana:
> Spilling env variable handling across code base is not a good idea. This is not a good practice. Also It would lead to write complex functional/unit tests.

The reason to not expose that is because it is not documented and shouldn't be available outside to be manipulated. From what it exists today this patch makes the env first class i.e treating it like a regular changing entity.  `_MINIO_CACHE` and `_MINIO_PROFILER` are treated specially and do not have any meaning outside. 

What you did only makes sense when we bring in the caching layer which is going to move outside of XL and becomes part of the S3 handler layer.

krishnasrinivas:
> The cache is disabled either by _MINIO_CACHE or by < 8GB memory. Do you say we support mixed cache mode of minio servers (some with cache enabled and some are not)?

Yes ... because there is no reason not to.

balamurugana:
> The reason to not expose that is because it is not documented and shouldn't be available outside to be manipulated. From what it exists today this patch makes the env first class i.e treating it like a regular changing entity. _MINIO_CACHE and _MINIO_PROFILER are treated specially and do not have any meaning outside.

I couldn't get the regular changing entity.  As a common practice, env variables to any program is optional.  Programs could fall back to default values.  minio does the same thing.

I understand that `_MINIO_CACHE` and `_MINIO_PROFILER` are special and hidden environment variables (by not documenting it) and they are used by developers.  That doesn't mean they have to be handled some layer beneath.

Handling env variables at one place gives clearer understanding and eases add/remove of them by anyone.

balamurugana:
> The cache is disabled either by _MINIO_CACHE or by < 8GB memory. Do you say we support mixed cache mode of minio servers (some with cache enabled and some are not)?
> 
> Yes ... because there is no reason not to.

Looks like a design call here.

krishnasrinivas:
> Handling env variables at one place gives clearer understanding and eases add/remove of them by anyone.
>  Hide all checks

OK @balamurugana in that case name it globalXLObjCacheDisabled

balamurugana:
> OK @balamurugana in that case name it globalXLObjCacheDisabled

Done

cmd/certs.go
```diff
@@ -24,11 +24,6 @@ import (
 	"path/filepath"
 )
 
-// isSSL - returns true with both cert and key exists.
-func isSSL() bool {
-	return isFile(getPublicCertFile()) && isFile(getPrivateKeyFile())
```
harshavardhana, line 29: is this helper function not useful?  why not?

cmd/server-main.go
```diff
@@ -485,62 +377,115 @@ func serverMain(c *cli.Context) {
...
+
+		// browser Envs are set globally, this does not represent
+		// if browser is turned off or on.
+		globalIsEnvBrowser = true
+		globalIsBrowserEnabled = bool(browserFlag)
+	}
+}
+
+// serverMain handler called for 'minio server' command.
+func serverMain(c *cli.Context) {
```
harshavardhana: If c is renamed as ctx lets use it everywhere..

cmd/server-main.go
```diff
@@ -485,62 +377,115 @@ func serverMain(c *cli.Context) {
...
+	}
+
+	// Get quiet flag from command line argument.
+	quietFlag := c.Bool("quiet") || c.GlobalBool("quiet")
+	if quietFlag {
+		log.EnableQuiet()
+	}
+
+	handleCmdArgs(c)
+	handleEnvVars()
```
harshavardhana: prefix this with `serverHandleCmdArgs()` so that avoiding namespace conflict in future for other commands.

cmd/server-main.go
```diff
@@ -485,62 +377,115 @@ func serverMain(c *cli.Context) {
...
+
+	// Create certs path.
+	fatalIf(createConfigDir(), "Unable to create configuration directories.")
+
+	initConfig()
+
+	// Enable all loggers by now so we can use errorIf() and fatalIf()
```
harshavardhana: what happens to fatalIf() and errorIf() until here before in handleEnvVars()? and initConfig() ? .

cmd/server-main.go
```diff
@@ -485,62 +377,115 @@ func serverMain(c *cli.Context) {
...
+	enableLoggers()
+
+	// Init the error tracing module.
+	initError()
+
+	// Check and load SSL certificates.
+	if isFile(getPublicCertFile()) && isFile(getPrivateKeyFile()) {
```
harshavardhana: Retaining `isSSLEnabled()` is better here, i don't see why you removed `isSSL()` - you could even set `globalIsSSL`  inside the function perhaps like `enableSSL()`  - this function sets `globalIsSSL = true` only if its possible and validates the public/private key.

harshavardhana (changes requested):

harshavardhana:
> I couldn't get the regular changing entity. As a common practice, env variables to any program is optional. Programs could fall back to default values. minio does the same thing.

What you are not seeing is that caching is only XL concept and at object layer, it has no reason for it spill out into serverMain() .

Currently this patch is buggy it sets the value of globalObjCacheEnabled to false and true regardless of XL/Distributed XL is used or not. A developer cannot make any conclusion from what is being changed.

So the flow itself is completely wrong either you handle the case where _MINIO_CACHE is only honored if XL is set or not. or keep it the same way as it is today.

cmd/certs.go
balamurugana: The function is incomplete ie existent of two files do not say SSL is true.  Below logic is correct.
```golang
	if isFile(getPublicCertFile()) && isFile(getPrivateKeyFile()) {
		var err error
		globalPublicCerts, err = parsePublicCertFile(getPublicCertFile())
		fatalIf(err, "Unable to parse public certificate file %s", getPublicCertFile())

		globalRootCAs, err = getRootCAs(getCADir())
		fatalIf(err, "Unable to load CA files in %s directory", getCADir())

		globalIsSSL = true
	}
```

cmd/server-main.go

cmd/server-main.go

cmd/server-main.go
balamurugana: Wrong comment from previous code.

balamurugana:
> What you are not seeing is that caching is only XL concept and at object layer, it has no reason for it spill out into serverMain() .
> 

As `serverMain()` is entry point to object layer and it is getting command line arguments and other startup initialization, it is the right place to parse/validate command line arguments and environment variables.  Reading `_MINIO_CACHE` and setting a global variable at  `serverMain()` is not a spill.  It sets a flag and does not do any specific logic how caching for XL object layer.

> Currently this patch is buggy it sets the value of globalObjCacheEnabled to false and true regardless of XL/Distributed XL is used or not. A developer cannot make any conclusion from what is being changed.
> 
> So the flow itself is completely wrong either you handle the case where _MINIO_CACHE is only honored if XL is set or not. or keep it the same way as it is today.

The right way of `newXLObjects()` is to accept tunable as arguments and work on them.  This way we could send various values in unit test and know how the function behaves.  However this is not the case now.

Having a environment variable as a control flag inside a function would prevent running unit/functional  tests ie one test may run with cache enabled and another one is not.  However globals really do not solve this. Eventually I am going to remove all globals down the line.

harshavardhana:
> The right way of newXLObjects() is to accept tunable as arguments and work on them. This way we could send various values in unit test and know how the function behaves. However this is not the case now.
> 
> Having a environment variable as a control flag inside a function would prevent running unit/functional tests ie one test may run with cache enabled and another one is not. However globals really do not solve this. Eventually I am going to remove all globals down the line.

Okay if you are removing globals later then its fine.  This is considered a part fix for an eventual fix.

cmd/certs.go
harshavardhana: > if isFile(getPublicCertFile()) && isFile(getPrivateKeyFile()) {
> 		var err error
> 		globalPublicCerts, err = parsePublicCertFile(getPublicCertFile())
> 		fatalIf(err, "Unable to parse public certificate file %s", getPublicCertFile())
> 
> 		globalRootCAs, err = getRootCAs(getCADir())
> 		fatalIf(err, "Unable to load CA files in %s directory", getCADir())
> 
> 		globalIsSSL = true
> 	}

Perhaps this can be moved into a more enableSSL() function?

balamurugana: Done

harshavardhana (approved):

cmd/generic-handlers.go
```diff
@@ -118,7 +118,7 @@ func (h redirectHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	// Re-direct only for JWT and anonymous requests from browser.
 	if aType == authTypeJWT || aType == authTypeAnonymous {
 		// Re-direction is handled specifically for browser requests.
-		if guessIsBrowserReq(r) && globalIsBrowserEnabled {
+		if guessIsBrowserReq(r) && globalIsBrowserEnabled == true {
```
harshavardhana: Looks like you forgot to change this.

harshavardhana (changes requested):
One minor..Comment.

cmd/generic-handlers.go
balamurugana: Done. Fixing it.

krishnasrinivas (approved):

harshavardhana (approved):

PR #3977, opened 2017-03-26
docker: Support docker swarm secrets.

### LINKED ISSUE #3896: docker: Minio in Docker should be able to read swarm secrets
## Expected Behavior

When deploying Minio as a Docker container in Swarm Mode, I should be able to create a secure swarm secret with the access and secret keys. When minio starts, an additional flag should be able to specify the filename to read for secrets.

## Current Behavior

Current behavior generates secrets at start-up and these have to be edited manually outside the container, or created outside the container and bind-mounted.

This is understandable because Docker Swarm Secrets were made available only recently. 

## Possible Solution

Optional flag to minio binary:

```
minio server --secrets-from=/run/secrets/minio_server_1
```

Docker Swarm does not provide secrets through environmental variables because these can be leaked easily - the `/run/secrets/<secret_name>` convention is used. To create a secret use `docker secret create`.

Here is the documentation for Swarm secrets - https://docs.docker.com/engine/swarm/secrets/

CC @NitishT @abperiasamy 

If accepted, I would be open to working on this feature/addition and it would help adoption of Minio with Docker Swarm.

harshavardhana:
## Description
Support reading swarm secrets

## Motivation and Context
Fixes #3896 

## How Has This Been Tested?
Manually with
```
echo "minio" | docker secret create access_key -
echo "minio123" | docker secret create secret_key -
```

```
docker service create --name="minio" --secret="access_key" \
      --secret="secret_key" minio/minio server /export
```

## Types of changes

## Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,5 +15,13 @@ RUN  \
      rm -rf /go/pkg /go/src /usr/local/go && apk del .build-deps
 
 EXPOSE 9000
-ENTRYPOINT ["minio"]
+
+COPY buildscripts/docker-entrypoint.sh /usr/bin/
+
+RUN chmod +x /usr/bin/docker-entrypoint.sh
+
+ENTRYPOINT ["/usr/bin/docker-entrypoint.sh"]
+
 VOLUME ["/export"]
+
+CMD ["minio"]
--- a/buildscripts/docker-entrypoint.sh
+++ b/buildscripts/docker-entrypoint.sh
@@ -1,0 +1,43 @@
+#!/bin/sh
+#
+# Minio Cloud Storage, (C) 2017 Minio, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# if command starts with an option, prepend minio
+if [ "${1}" != "minio" ]; then
+    if [ -n "${1}" ]; then
+        set -- minio "$@"
+    fi
+fi
+
+## Look for swarm secrets in default location.
+file_env() {
+    local MINIO_ACCESS_KEY_FILE="/run/secrets/access_key"
+    local MINIO_SECRET_KEY_FILE="/run/secrets/secret_key"
+
+    if [ -f $MINIO_ACCESS_KEY_FILE -a -f $MINIO_SECRET_KEY_FILE ]; then
+        if [ -f $MINIO_ACCESS_KEY_FILE ]; then
+            export MINIO_ACCESS_KEY="$(cat "$MINIO_ACCESS_KEY_FILE")"
+        fi
+        if [ -f $MINIO_SECRET_KEY_FILE ]; then
+            export MINIO_SECRET_KEY="$(cat "$MINIO_SECRET_KEY_FILE")"
+        fi
+    fi
+}
+
+## Set env if necessary.
+file_env
+
+exec "$@"
--- a/Dockerfile.aarch64
+++ b/Dockerfile.aarch64
@@ -15,5 +15,13 @@ RUN  \
      rm -rf /go/pkg /go/src /usr/local/go && apk del .build-deps
 
 EXPOSE 9000
-ENTRYPOINT ["minio"]
+
+COPY buildscripts/docker-entrypoint.sh /usr/bin/
+
+RUN chmod +x /usr/bin/docker-entrypoint.sh
+
+ENTRYPOINT ["/usr/bin/docker-entrypoint.sh"]
+
 VOLUME ["/export"]
+
+CMD ["minio"]
--- a/Dockerfile.armhf
+++ b/Dockerfile.armhf
@@ -15,5 +15,13 @@ RUN  \
      rm -rf /go/pkg /go/src /usr/local/go && apk del .build-deps
 
 EXPOSE 9000
-ENTRYPOINT ["minio"]
+
+COPY buildscripts/docker-entrypoint.sh /usr/bin/
+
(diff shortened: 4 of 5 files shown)


Dockerfile
```diff
@@ -15,5 +15,12 @@ RUN  \
...
-ENTRYPOINT ["minio"]
+
+COPY buildscripts/docker-entrypoint.sh /usr/bin/
+RUN ln -s usr/bin/docker-entrypoint.sh /entrypoint.sh && \
+     chmod +x /entrypoint.sh && \
+     chmod +x /usr/bin/docker-entrypoint.sh # backwards compat
+ENTRYPOINT ["docker-entrypoint.sh"]
+
 VOLUME ["/export"]
+CMD ["minio"]
```
alexellis: Do you need a CMD and an ENTRYPOINT? Normally I used one or the other.

buildscripts/docker-entrypoint.sh
```diff
@@ -0,0 +1,43 @@
...
+fi
+
+## Look for swarm secrets in default location.
+file_env() {
+    local MINIO_ACCESS_KEY_FILE="/run/secrets/access_key"
+    local MINIO_SECRET_KEY_FILE="/run/secrets/secret_key"
+
+    if [ -f $MINIO_ACCESS_KEY_FILE -a -f $MINIO_SECRET_KEY_FILE ]; then 
+        if [ -f $MINIO_ACCESS_KEY_FILE ]; then
+            export MINIO_ACCESS_KEY="$(< "$MINIO_ACCESS_KEY_FILE")"
```
alexellis: I wonder if this env variable can get leaked through the process/container? It would be more secure I think if Minio read this on start-up from the disk if present.

alexellis: Check out this page under "Build support for Docker Secrets into your images"

https://docs.docker.com/engine/swarm/secrets/#build-support-for-docker-secrets-into-your-images

Dockerfile
harshavardhana: CMD is the default argument to entrypoint. Since entrypoint here is a script  `minio` becomes the default argument.

buildscripts/docker-entrypoint.sh
harshavardhana: ENVs never get leaked, unless there is an exploit into the process itself. if that ever happens all bets are off anyways.

harshavardhana: Minio reading from disk and ENVs are the same in terms of security.  The way to extract info either regarding the file or the ENVs is to have a shell exploit. So making minio read these files is not going to be any secure. 

Docker secrets by themselves i don't see them to be secure, the secret files are in plain text and quite readable since you are root anyways inside the container. If the party is concerned with security even the container itself should be exploit proof which is also not the case.

Overall minio access and secret keys are not the same as GPG passphrases and can be changed quite easily from a WebUI etc and take wide range of values.  If you are planning to expose Minio on a public network then it is upto you to use SSL and also share keys carefully. There are no software solutions to human errors :-).

deekoder:
@abperiasamy approved the proposed solution by @harshavardhana

buildscripts/docker-entrypoint.sh
harshavardhana: Also tested with the ENVs leaking process envs run through a script are well inside the process itself so an entrypoint.sh contains it from being exposed, since these are not set container wide.

```
/data # env
HOSTNAME=1f3d3b6c405b
SHLVL=1
HOME=/root
MINIO_TEST.1.PGIKYY0C35OEJLYKI4V6XMZPW_ENV_GOPATH=/go
MINIO_TEST.1.PGIKYY0C35OEJLYKI4V6XMZPW_PORT_9000_TCP_ADDR=172.17.0.2
MINIO_TEST.1.PGIKYY0C35OEJLYKI4V6XMZPW_ENV_CGO_ENABLED=0
MINIO_TEST.1.PGIKYY0C35OEJLYKI4V6XMZPW_PORT_9000_TCP_PORT=9000
REDIS_DOWNLOAD_SHA1=6780d1abb66f33a97aad0edbe020403d0a15b67f
MINIO_TEST.1.PGIKYY0C35OEJLYKI4V6XMZPW_PORT_9000_TCP_PROTO=tcp
TERM=xterm
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
MINIO_TEST.1.PGIKYY0C35OEJLYKI4V6XMZPW_PORT_9000_TCP=tcp://172.17.0.2:9000
MINIO_TEST.1.PGIKYY0C35OEJLYKI4V6XMZPW_PORT=tcp://172.17.0.2:9000
REDIS_DOWNLOAD_URL=http://download.redis.io/releases/redis-3.2.8.tar.gz
MINIO_TEST.1.PGIKYY0C35OEJLYKI4V6XMZPW_NAME=/amazing_aryabhata/minio-test.1.pgikyy0c35oejlyki4v6xmzpw
REDIS_VERSION=3.2.8
```

```
docker run --rm --link minio-test.1.pgikyy0c35oejlyki4v6xmzpw redis:alpine
```

So even with links the secrets are not exposed. So the kernel is making sure of that ENVs don't spill over.

nitisht (approved):

vadmeste (approved):

buildscripts/docker-entrypoint.sh
alexellis: Hi - I caught up with Diogo the security lead at Docker. 

The argument against environmental variables is that they can be leaked easily through log files and stack traces / dumps. I think where these could leak is from the minio Golang process if there was any debugging / output of env vars.

Bit more reading here: https://www.linkedin.com/pulse/why-you-shouldnt-use-env-variables-secret-data-diogo-m%C3%B3nica

harshavardhana: That way mysql, postgres all have problems. I don't think its such a big deal. If a process itself has envs in its own namespace isolated by the kernel. 

**To access the env's one need to exploit the server itself. Now if the server is exploited reading ENVs should be least of our worries.**

Argument that log files, stack traces and dumps means a programmer error - there is no solution for that without doing an audit on the code itself.  

Additionally we also print the access/secret it is visible in the `docker logs` . To avoid that you need to specify `--quiet`.  IMHO Go doesn't allow leaking ENVs while debugging a process -
 that would be a failure of their design.

Issue #3980, opened 2017-03-27
admin-api: Issue when heal is attempted while a node is down.


hackintoshrao:
Incorrect error message/false-detection when heal is attempted on a setup in which a node down.
## Expected Behavior

A message which could convey that heal has to be attempted after the node is back.
## Current Behavior

An attempt to heal shows that the object is healed.
## Possible Solution

- The issue could just at the presentation layer or it could be more deeper problem in detection.
## Steps to Reproduce (for bugs)

1. Start a 4 node Minio distributed.
2. Bring a node down.
3. Upload an object.
4. Now try to heal the bucket.

Note: Since the node to be healed is still not up, an attempt to heal should inform that to the user.

## Context

`mc admin heal` testing.
## Your Environment

* Version used: 6e63904048e9d7318c5454268b418b92317ba871

krisis:
@harshavardhana I am working on this.

deekoder:
If the server is capable of healing and can establish quorum, it should. If we can't return error message.

krisis:
## Background
Healing algorithm returns failure only if an attempt to heal any of the disks failed. It returns success even if all disks needing heal were offline.
In its current form, heal admin REST APIs returns the above boolean state making it impossible for clients like `mc` to distinguish between an object being healed completely, partially and not.

## Solution
The fix (so far) is to return `struct { HealedCount int, OfflineCount int}` in addition to `error` for a heal-object and heal-upload operation. `mc` can make use of this to guide their users better.

## Future-proof solution
This brings us to whether it is right by design to provide internal details of the healing process in admin API. @abperiasamy, IIUC we could return an enumerated constant in heal-object and heal-upload API that captures the 3 result states of the heal operation, namely, `CompletelyHealed`, `PartiallyHealed` and `NotHealed`. In addition, we would return `error` to indicate if the healing operation failed to due to other I/O related errors. Does this approach sound right?

krisis:
@deekoder I have updated admin-api docs in this PR - https://github.com/minio/minio/pull/4045/files#diff-b8edc1fe50490a19c5d5fd2849477ad9.

krisis:
This issue can be closed once `mc` uses these heal states to provide user with rich(er) information about a heal-object or heal-upload admin operation. Waiting on #4045 to be merged. The pkg/madmin changes need to be vendored in.

harshavardhana:
[...]

PR #3990, opened 2017-03-28
Add disksUnavailable healStatus const

krisis:
This is one of a two part fix for #3980. `mc admin heal` needs a change to interpret `canPartiallyHeal` heal status and inform user if an object can be healed or some disks with missing object parts are offline.

## Description

`canPartiallyHeal` healStatus constant indicates that a given object needs healing but one or more of disks requiring heal are offline. This can be used by admin heal API consumers to distinguish between a successful heal and a no-op since the outdated disks were offline. In addition, heal-object and heal-upload admin APIs now return the number of disks the object/upload was healed and the number of disks that were offline during the heal operation.

## Motivation and Context

See #3980 
## How Has This Been Tested?

Manually tested the scenario described in #3980 
## Types of changes

## Checklist:

### DIFF
--- a/cmd/object-api-datatypes.go
+++ b/cmd/object-api-datatypes.go
@@ -57,6 +57,7 @@ const (
 	canHeal                             // Object can be healed
 	corrupted                           // Object can't be healed
 	quorumUnavailable                   // Object can't be healed until read quorum is available
+	canPartiallyHeal                    // Object can't be healed completely until outdated disk(s) are online.
 )
 
 // HealBucketInfo - represents healing related information of a bucket.
@@ -78,9 +79,9 @@ type BucketInfo struct {
 
 // HealObjectInfo - represents healing related information of an object.
 type HealObjectInfo struct {
-	Status              healStatus
-	MissingDataCount    int
-	MissingPartityCount int
+	Status             healStatus
+	MissingDataCount   int
+	MissingParityCount int
 }
 
 // ObjectInfo - represents object metadata.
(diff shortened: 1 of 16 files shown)


cmd/object-api-datatypes.go
```diff
@@ -57,6 +57,7 @@ const (
 	canHeal                             // Object can be healed
 	corrupted                           // Object can't be healed
 	quorumUnavailable                   // Object can't be healed until read quorum is available
+	disksUnavailable                    // Object can't be healed completely until outdated disk(s) are online.
```
vadmeste: Probably a better name for this is `canPartiallyHeal`

krishnasrinivas (commented):
If 1 disk needs healing and 1 disk is offline, we will still be able to heal 1 disk right? i.e this change only affects the user message?

krisis:
>If 1 disk needs healing and 1 disk is offline, we will still be able to heal 1 disk right? i.e this change only affects the user message?

This change allows `mc` like applications to inform users whether the object will be healed completely or partially. In addition to this, I will be changing heal-object admin API to return number of disks healed and number of disks offline during heal-object. This is to help `mc` to inform users the specifics of the heal operation.

krishnasrinivas (approved):

cmd/object-api-datatypes.go
harshavardhana: any comments on this @krisis ?

krisis: I have one more change to this PR where I will address this comment.

PR #3996, opened 2017-03-29
config/path: Figure out absolute paths properly on windows.

### LINKED ISSUE #3991: windows: "minio -C ./some_dir server ./dir" leads to error in initializing minio server
## Expected Behavior
`minio -C ./some_dir server ./dir` should work even if there is not directory `./some_dir`

## Current Behavior
`minio -C ./some_dir server ./dir`

leads to the following error:

`minio: <ERROR> Unable to initialize minio config for the first time. Error: mkdir \\?\some_dir\certs: The system cannot find the path specified..`

## Possible Solution
NA

1. `minio -C ./some_dir server ./dir`
2.
3.
4.

## Your Environment
* Version used:
```
Version: 2017-03-16T21:50:32Z
Release-Tag: RELEASE.2017-03-16T21-50-32Z
Commit-ID: 5311eb22fd681a8cd4a46e2a872d46c2352c64e8
```
* Server type and version:
`Windows 10`
* Link to your project:

harshavardhana:
## Description
The following form of arguments such as

```
minio.exe -C some_dir server dir
```

has stopped working because of lack of handling of
absolute paths for config directory. Always calculate
absolute path for any relative paths on any operating
system.

The following fix converts all config directory relative
paths into absolute paths.

## Motivation and Context
Fixes #3991 

## How Has This Been Tested?
Manually in Windows 2012 R2

## Types of changes

## Checklist:

### DIFF
--- a/cmd/server-main.go
+++ b/cmd/server-main.go
@@ -23,6 +23,7 @@ import (
 	"net/url"
 	"os"
 	"path"
+	"path/filepath"
 	"sort"
 	"strconv"
 	"strings"
@@ -480,6 +481,14 @@ func serverMain(c *cli.Context) {
 	}
 	if configDir == "" {
 		fatalIf(errors.New("empty directory"), "Configuration directory cannot be empty.")
+	}
+
+	// Disallow relative paths, figure out absolute paths.
+	{
+		configDirAbs, err := filepath.Abs(configDir)
+		fatalIf(err, "Unable to fetch absolute path for config directory %s", configDir)
+
+		configDir = configDirAbs
 	}
 
 	// Set configuration directory.


cmd/server-main.go
```diff
@@ -482,8 +483,12 @@ func serverMain(c *cli.Context) {
 		fatalIf(errors.New("empty directory"), "Configuration directory cannot be empty.")
 	}
 
+	// Disallow relative paths, figure out absolute paths.
+	configDirAbs, err := filepath.Abs(configDir)
```
balamurugana: you can assign it back to `configDir` or make this into a block so that `configDirAbs` is available further below.

harshavardhana: Upon error `configDir` gets replaced when printing.. to empty string. and what do you mean "block" it?

harshavardhana: done

balamurugana (approved):

cmd/server-main.go
```diff
@@ -482,6 +483,14 @@ func serverMain(c *cli.Context) {
 		fatalIf(errors.New("empty directory"), "Configuration directory cannot be empty.")
 	}
 
+	// Disallow relative paths, figure out absolute paths.
```
krisis, line 486: Suggestion: if we need to place this inside a block, we could use,
```go
if configDirAbs, err := file path.Abs(configDir); err != nil {
 // Failure code goes here
} else {
  // Set configDir here
}

krisis (approved):
The changes look good to me. Suggestion is optional.

cmd/server-main.go
balamurugana: Generic: If failure `//Failure: code goes here` does `return`, `golint` will throw warning.

krisis: @bala, `configDirAbs`'s scope ends in the if-else block. To avoid that `configDirAbs` should be defined earlier, defeating the point of restricting its scope.

harshavardhana: Since we use fatalIf currently style is more preferred. It collapses the `if err != nil { `

PR #4002, opened 2017-03-29
sys/stats: return cgroup mem limit, fall back to sysinfo()

### LINKED ISSUE #4001: caching: Out of memory crashes on cgroup limit
I have seem some crashes on my host with the following message.

[34549785.782753] Memory cgroup out of memory: Kill process 35945 (minio) score 995 or sacrifice child

[34549785.800895] Killed process 35945 (minio) total-vm:8804576kB, anon-rss:3962816kB, file-rss:0kB

The minio server is running in LXC container which is limited to 4GB ram.

It is possible that minio sees the HOST RAM which is much higher: 250+GB

wsadmin@wscwbs-storage06-prod:~$ free -g
             total       used       free     shared    buffers     cached
Mem:           251        250          1          0          8         46

Work-around: I will now start the servers with _MINIO_CACHE_=off.

harshavardhana:
## Description
This is necessary where in certain environments where
cgroup is used to limit memory usage of a container or
a particular process.

GetStats() is used by caching module to figure out the
optimal cacheable size in memory with cgroup limits
what sysinfo reports might not be the right value set
for a given process.

## Motivation and Context
Fixes #4001

## How Has This Been Tested?
Manually using cgroup limits.

## Types of changes

## Checklist:

### DIFF
--- a/pkg/cgroup/linux.go
+++ b/pkg/cgroup/linux.go
@@ -1,0 +1,177 @@
+// +build linux
+
+/*
+ * Minio Cloud Storage, (C) 2017 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Package cgroup implements parsing for all the cgroup
+// categories and functionality in a simple way.
+package cgroup
+
+import (
+	"bufio"
+	"bytes"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strconv"
+	"strings"
+)
+
+// DO NOT EDIT following constants are chosen defaults for any kernel
+// after 3.x, please open a github issue https://github.com/minio/minio/issues
+// and discuss first if you wish to change this.
+const (
+	// Default string for looking for kernel memory param.
+	memoryLimitKernelParam = "memory.limit_in_bytes"
+
+	// Points to sys path memory path.
+	cgroupMemSysPath = "/sys/fs/cgroup/memory"
+
+	// Default docker prefix.
+	dockerPrefixName = "/docker/"
+
+	// Proc controller group path.
+	cgroupFileTemplate = "/proc/%d/cgroup"
+)
+
+// CGEntries - represents all the entries in a process cgroup file
+// at /proc/<pid>/cgroup as key value pairs.
+type CGEntries map[string]string
+
+// GetEntries reads and parses all the cgroup entries for a given process.
+func GetEntries(pid int) (CGEntries, error) {
+	r, err := os.Open(fmt.Sprintf(cgroupFileTemplate, pid))
+	if err != nil {
+		return nil, err
+	}
+	defer r.Close()
+	return parseProcCGroup(r)
+}
+
+// parseProcCGroup - cgroups are always in the following
+// format once enabled you need to know the pid of the
+// application you are looking for so that the the
+// following parsing logic only parses the file located
+// at /proc/<pid>/cgroup.
+//
+// CGROUP entries id, component and path are always in
+// the following format. ``ID:COMPONENT:PATH``
+//
+// Following code block parses this information and
+// returns a procCGroup which is a parsed list of all
+// the line by line entires from /proc/<pid>/cgroup.
+func parseProcCGroup(r io.Reader) (CGEntries, error) {
+	var cgEntries = CGEntries{}
+
+	// Start reading cgroup categories line by line
+	// and process them into procCGroup structure.
+	scanner := bufio.NewScanner(r)
+	for scanner.Scan() {
+		line := scanner.Text()
+
+		tokens := strings.SplitN(line, ":", 3)
+		if len(tokens) < 3 {
+			continue
+		}
+
+		name, path := tokens[1], tokens[2]
+		for _, token := range strings.Split(name, ",") {
+			name = strings.TrimPrefix(token, "name=")
+			cgEntries[name] = path
+		}
+	}
+
+	// Return upon any error while reading the cgroup categories.
+	if err := scanner.Err(); err != nil {
+		return nil, err
+	}
+
+	return cgEntries, nil
+}
+
+// Fetch value of the cgroup kernel param from the cgroup manager,
+// if cgroup manager is configured we should just rely on `cgm` cli
+// to fetch all the values for us.
+func getManagerKernValue(cname, path, kernParam string) (limit uint64, err error) {
+
+	cmd := exec.Command("cgm", "getvalue", cname, path, kernParam)
+	var out bytes.Buffer
+	cmd.Stdout = &out
+	if err = cmd.Run(); err != nil {
+		return 0, err
+	}
+
+	// Parse the cgm output.
+	limit, err = strconv.ParseUint(strings.TrimSpace(out.String()), 10, 64)
+	return limit, err
+}
+
+// Get cgroup memory limit file path.
+func getMemoryLimitFilePath(cgPath string) string {
+	path := cgroupMemSysPath
+
+	// Docker generates weird cgroup paths that don't
+	// really exist on the file system.
+	//
+	// For example on regular Linux OS :
+	// `/user.slice/user-1000.slice/session-1.scope`
+	//
+	// But they exist as a bind mount on Docker and
+	// are not accessible :  `/docker/<hash>`
+	//
+	// We we will just ignore if there is `/docker` in the
+	// path ignore and fall back to :
+	// `/sys/fs/cgroup/memory/memory.limit_in_bytes`
+	if !strings.HasPrefix(cgPath, dockerPrefixName) {
+		path = filepath.Join(path, cgPath)
+	}
+
+	// Final path.
+	return filepath.Join(path, memoryLimitKernelParam)
+}
+
+// GetMemoryLimit - Fetches cgroup memory limit either from
+// a file path at '/sys/fs/cgroup/memory', if path fails then
+// fallback to querying cgroup manager.
+func GetMemoryLimit(pid int) (limit uint64, err error) {
+	var cg CGEntries
+	cg, err = GetEntries(pid)
+	if err != nil {
+		return 0, err
+	}
+
+	path := cg["memory"]
+
+	limit, err = getManagerKernValue("memory", path, memoryLimitKernelParam)
+	if err != nil {
+
+		// Upon any failure returned from `cgm`, on some systems cgm
+		// might not be installed. We fallback to using the the sysfs
+		// path instead to lookup memory limits.
+		var b []byte
+		b, err = ioutil.ReadFile(getMemoryLimitFilePath(path))
+		if err != nil {
+			return 0, err
+		}
+
+		limit, err = strconv.ParseUint(strings.TrimSpace(string(b)), 10, 64)
+	}
+
+	return limit, err
+}
--- a/pkg/cgroup/linux_test.go
+++ b/pkg/cgroup/linux_test.go
@@ -1,0 +1,140 @@
+// +build linux
+
+/*
+ * Minio Cloud Storage, (C) 2017 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cgroup
+
+import (
+	"io/ioutil"
+	"os"
+	"testing"
+)
+
+// Testing parsing correctness for various process cgroup files.
+func TestProcCGroup(t *testing.T) {
+	tmpPath, err := ioutil.TempFile("", "cgroup")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.Remove(tmpPath.Name())
+
+	cgroup := `
+11:memory:/user.slice
+10:blkio:/user.slice
+9:hugetlb:/
+8:net_cls,net_prio:/
+7:perf_event:/
+6:pids:/user.slice/user-1000.slice
+5:devices:/user.slice
+4:cpuset:/
+3:cpu,cpuacct:/user.slice
+2:freezer:/
+1:name=systemd:/user.slice/user-1000.slice/session-1.scope
+`
+	_, err = tmpPath.WriteString(cgroup)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Seek back to read from the beginning.
+	tmpPath.Seek(0, 0)
+
+	cg, err := parseProcCGroup(tmpPath)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	path := cg["memory"]
+	if len(path) == 0 {
+		t.Fatal("Path component cannot be empty")
+	}
+
+	if path != "/user.slice" {
+		t.Fatal("Path component cannot be empty")
+	}
+
+	path = cg["systemd"]
+	if path != "/user.slice/user-1000.slice/session-1.scope" {
+		t.Fatal("Path component cannot be empty")
+	}
+
+	// Mixed cgroups with different group names.
+	cgroup = `
+11:memory:/newtest/newtest
+10:blkio:/user.slice
+9:hugetlb:/
+8:net_cls,net_prio:/
+7:perf_event:/
+6:pids:/user.slice/user-1000.slice
+5:devices:/user.slice
+4:cpuset:/
+3:cpu,cpuacct:/newtest/newtest
+2:freezer:/
+1:name=systemd:/user.slice/user-1000.slice/session-1.scope
+`
+
+	// Seek back to read from the beginning.
+	tmpPath.Seek(0, 0)
+
+	_, err = tmpPath.WriteString(cgroup)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Seek back to read from the beginning.
+	tmpPath.Seek(0, 0)
+
+	cg, err = parseProcCGroup(tmpPath)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	path = cg["memory"]
+	if path != "/newtest/newtest" {
+		t.Fatal("Path component cannot be empty")
+	}
+
+	path = cg["systemd"]
+	if path != "/user.slice/user-1000.slice/session-1.scope" {
+		t.Fatal("Path component cannot be empty")
+	}
+
+}
+
+// Tests cgroup memory limit path construction.
+func TestMemoryLimitPath(t *testing.T) {
+	testCases := []struct {
+		cgroupPath   string
+		expectedPath string
+	}{
+		{
+			cgroupPath:   "/user.slice",
+			expectedPath: "/sys/fs/cgroup/memory/user.slice/memory.limit_in_bytes",
+		},
+		{
+			cgroupPath:   "/docker/testing",
+			expectedPath: "/sys/fs/cgroup/memory/memory.limit_in_bytes",
+		},
+	}
+
+	for i, testCase := range testCases {
+		actualPath := getMemoryLimitFilePath(testCase.cgroupPath)
+		if actualPath != testCase.expectedPath {
+			t.Fatalf("Test: %d: Expected: %s, got %s", i+1, testCase.expectedPath, actualPath)
+		}
+	}
+}
--- a/pkg/cgroup/others.go
+++ b/pkg/cgroup/others.go
@@ -1,5 +1,7 @@
+// +build !linux
+
 /*
- * Minio Cloud Storage, (C) 2016, 2017 Minio, Inc.
+ * Minio Cloud Storage, (C) 2017 Minio, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -14,17 +16,4 @@
  * limitations under the License.
  */
 
-package sys
-
-import "testing"
-
-// Test get stats result.
-func TestGetStats(t *testing.T) {
-	stats, err := GetStats()
-	if err != nil {
-		t.Errorf("Tests: Expected `nil`, Got %s", err)
-	}
-	if stats.TotalRAM == 0 {
-		t.Errorf("Tests: Expected `n > 0`, Got %d", stats.TotalRAM)
-	}
-}
+package cgroup
--- a/pkg/sys/stats_linux.go
+++ b/pkg/sys/stats_linux.go
@@ -6,7 +6,7 @@
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- *shouldP
+ *
  *     http://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
@@ -18,14 +18,65 @@
 
 package sys
 
-import "syscall"
+import (
+	"os"
+	"syscall"
 
-// GetStats - return system statistics.
-func GetStats() (stats Stats, err error) {
-	var si syscall.Sysinfo_t
-	if err = syscall.Sysinfo(&si); err == nil {
-		stats.TotalRAM = uint64(si.Totalram)
+	"github.com/minio/minio/pkg/cgroup"
+)
+
+// Get the final system memory limit chosen by the user.
+// by default without any configuration on a vanilla Linux
+// system you would see physical RAM limit. If cgroup
+// is configured at some point in time this function
+// would return the memory limit chosen for the given pid.
+func getMemoryLimit() (sysLimit uint64, err error) {
+	if sysLimit, err = getSysinfoMemoryLimit(); err != nil {
+		// Physical memory info is not accessible, just exit here.
+		return 0, err
 	}
 
-	return stats, err
+	// Following code is deliberately ignoring the error.
+	cGroupLimit, gerr := cgroup.GetMemoryLimit(os.Getpid())
+	if gerr != nil {
+		// Upon error just return system limit.
+		return sysLimit, nil
+	}
+
+	// cgroup limit is lesser than system limit means
+	// user wants to limit the memory usage further
+	// treat cgroup limit as the system limit.
+	if cGroupLimit <= sysLimit {
+		sysLimit = cGroupLimit
+	}
+
+	// Final system limit.
+	return sysLimit, nil
+
+}
+
+// Get physical RAM size of the node.
+func getSysinfoMemoryLimit() (limit uint64, err error) {
+	var si syscall.Sysinfo_t
+	if err = syscall.Sysinfo(&si); err != nil {
+		return 0, err
+	}
+
+	// Total RAM is always the multiplicative value
+	// of unit size and total ram.
+	limit = uint64(si.Unit) * si.Totalram
+	return limit, nil
+}
+
+// GetStats - return system statistics, currently only
+// supported value is TotalRAM.
+func GetStats() (stats Stats, err error) {
+	var limit uint64
+	limit, err = getMemoryLimit()
+	if err != nil {
+		return Stats{}, err
+	}
+
+	stats.TotalRAM = limit
+	return stats, nil
 }


harshavardhana:
Tested this with local docker container 

```yml
version: '2'

# starts 4 docker containers running minio server instances. Each
# minio server's web interface will be accessible on the host at port
# 9001 through 9004.
services:
  minio1:
    image: y4m4/minio
    mem_limit: 100m
    ports:
      - "9001:9000"
    volumes:
      - /mnt/export/minio1:/export
    environment:
      MINIO_ACCESS_KEY: minio
      MINIO_SECRET_KEY: minio123
    command: server http://minio1/export http://minio2/export http://minio3/export http://minio4/export        
  minio2:
    image: y4m4/minio
    mem_limit: 100m    
    ports:
      - "9002:9000"
    volumes:
      - /mnt/export/minio2:/export
    environment:
      MINIO_ACCESS_KEY: minio
      MINIO_SECRET_KEY: minio123
    command: server http://minio1/export http://minio2/export http://minio3/export http://minio4/export    
  minio3:
    image: y4m4/minio
    mem_limit: 100m    
    ports:
      - "9003:9000"
    volumes:
      - /mnt/export/minio3:/export
    environment:
      MINIO_ACCESS_KEY: minio
      MINIO_SECRET_KEY: minio123
    command: server http://minio1/export http://minio2/export http://minio3/export http://minio4/export
  minio4:
    image: y4m4/minio
    mem_limit: 100m  
    ports:
      - "9004:9000"
    volumes:
      - /mnt/export/minio4:/export
    environment:
      MINIO_ACCESS_KEY: minio
      MINIO_SECRET_KEY: minio123
    command: server http://minio1/export http://minio2/export http://minio3/export http://minio4/export
```

harshavardhana:
Create a local docker container with this patch.

``` 
FROM ubuntu
COPY minio /usr/bin/minio

ENTRYPOINT ["/usr/bin/minio"]
```

pkg/sys/cgroup_linux.go
```diff
@@ -0,0 +1,189 @@
...
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strconv"
+	"strings"
+)
+
+// Splits special only splits based on the first occurrence
+// of the separator.
+func splitSpecial(text string, sep string) (head string, tail string) {
```
krisis: `s/splitSpecial/splitBy`
`splitSpecial` doesn't convey what is special. OTOH `splitBy` suggests that it is parametrized by something, a separator in this case.

pkg/sys/cgroup_linux.go
```diff
@@ -0,0 +1,189 @@
...
+// you need to know the pid of the application you are
+// looking for so that the the following parsing logic
+// only parses the file located at /proc/<pid>/cgroup.
+//
+// CGROUP entries id, component and path are always in
+// the following format. ``ID:COMPONENT:PATH``
+//
+// Following code block parses this information and
+// returns a procCGroup which is a parsed list of all
+// the line by line entires from /proc/<pid>cgroup.
```
krisis: typo: `s/<pid>group/<pid>\/cgroup`

pkg/sys/cgroup_linux.go
```diff
@@ -0,0 +1,189 @@
...
+				return proc, err
+			}
+
+			proc = append(proc, cGroup{ID: iid, Name: name, Path: path})
+			name = next
+		}
+	}
+
+	if err := bio.Err(); err != nil {
+		return proc, err
```
krisis: It should be `return nil, err`.

pkg/sys/cgroup_linux.go
```diff
@@ -0,0 +1,189 @@
...
+		if memory := cgroups.getByName("memory"); len(memory) != 0 {
+			// Pick the first one.
+			limit, err = getMemoryCGroupMemoryLimit(memory[0])
+		}
+	}
+	return limit, err
+}
+
+// Default value for any error or if cgroup is not set.
+const unlimitedMemoryLimit = 9223372036854771712
```
krisis: Shouldn't the default limit be only as large as the RAM?

pkg/sys/cgroup_linux.go
balamurugana: isn't `strings.SplitN("hello:gnu:world", ":", 2)` different?

cmd/server-rlimit.go
```diff
@@ -65,7 +69,7 @@ func GetMaxCacheSize() (cacheSize uint64, err error) {
 
 	// Get total RAM.
 	var stats sys.Stats
-	if stats, err = sys.GetStats(); err != nil {
+	if stats, err = sys.GetStats(os.Getpid()); err != nil {
```
balamurugana: As `sys.GetStats()` returns current PID's stat, there is no need to pass PID as argument.

pkg/sys/cgroup_linux.go
```diff
@@ -0,0 +1,189 @@
...
+		head, tail = text[:i], text[i+1:]
+	} else {
+		head = text
+	}
+	head = strings.TrimSpace(head)
+	tail = strings.TrimSpace(tail)
+	return
+}
+
+type procCGroup []cGroup
```
balamurugana: this type could be after `cGroup` type for better readability.

pkg/sys/cgroup_linux.go
```diff
@@ -0,0 +1,189 @@
...
+
+type cGroup struct {
```
balamurugana: I think `cgroup` would be better name.

pkg/sys/cgroup_linux.go
```diff
@@ -0,0 +1,189 @@
...
+// CGROUP entries id, component and path are always in
+// the following format. ``ID:COMPONENT:PATH``
+//
+// Following code block parses this information and
+// returns a procCGroup which is a parsed list of all
+// the line by line entires from /proc/<pid>cgroup.
+func parseProcCGroup(r io.Reader) (procCGroup, error) {
+	var proc procCGroup
+
+	bio := bufio.NewScanner(r)
```
balamurugana: `scanner` is better name than `bio`

balamurugana (changes requested):
I see below is the essence of finding memory size of lxc.
In bash;
```bash
mempathseg=${awk -F: '$2 ~ /^memory$/ {print $3}'  /proc/1/cgroup}

mempath=/sys/fs/cgroup/memory/
if [[ "$mempathseg" != /docker/* ]]; then
    mempath="$mempath$mempathseg"
fi
mempath="$mempath/memory.limit_in_bytes"

memsize=$(cat $mempath)
```

I couldn't able to get why too many layers of code.

pkg/sys/cgroup_linux.go
harshavardhana: Not in cgroup   it weirdly keeps this value.

cmd/server-rlimit.go
harshavardhana: Pid is needed to know the right cgroup rules for the given process. As each process can have a different cgroup.

krishnasrinivas:
```
cmd := exec.Command("cgm", "getvalue", cgroup.Name, cgroup.Path, "memory.limit_i\
    n_bytes")
```

On host:
```
root@escape:~# cgm getvalue memory /docker/f6234f456d76086b7da2e758a5b860653be3c5b3503d456525972abc104cb379 memory.limit_in_bytes
call to cgmanager_get_value_sync failed: invalid request
root@escape:~# 
```

krishnasrinivas:
"return early" style results in better indentation and more readable code:
for ex:
```
func getProcCGroup(pid int) (proc procCGroup, err error) {                                   
        var r *os.File                                                                       
        if r, err = os.Open(fmt.Sprintf("/proc/%d/cgroup", pid)); err == nil {               
                defer r.Close()                                                              
                return parseProcCGroup(r)                                                    
        }                                                                                    
        return proc, err                                                                     
}
```
Vs
```
func getProcCGroup(pid int) (proc procCGroup, err error) {                                   
        r, err := os.Open(fmt.Sprintf("/proc/%d/cgroup", pid))                               
        if err != nil {                                                                      
                return proc, err                                                             
        }                                                                                    
        defer r.Close()                                                                      
        return parseProcCGroup(r)                                                            
}
```

Another ex:
```
func getCGroupMemoryLimit(pid int) (limit uint64, err error) {                               
        if cgroups, err := getProcCGroup(pid); err == nil {                                  
                if memory := cgroups.getByName("memory"); len(memory) != 0 {                 
                        // Pick the first one.                                               
                        limit, err = getMemoryCGroupMemoryLimit(memory[0])                   
                }                                                                            
        }                                                                                    
        return limit, err                                                                    
} 
```
Vs
```
func getCGroupMemoryLimit(pid int) (limit uint64, err error) {                               
        cgroups, err := getProcCGroup(pid)                                                   
        if err != nil {                                                                      
                return 0, err                                                                
        }                                                                                    
        memory := cgroups.getByName("memory")                                                
        if len(memory) == 0 {                                                                
                return 0, nil                                                                
        }                                                                                    
        // Pick the first one.                                                               
        return getMemoryCGroupMemoryLimit(memory[0])                                         
}
```

harshavardhana:
> mempathseg=${awk -F: '$2 ~ /^memory$/ {print $3}'  /proc/1/cgroup}
> 
> mempath=/sys/fs/cgroup/memory/
> if [[ "$mempathseg" != /docker/* ]]; then
>     mempath="$mempath$mempathseg"
> fi
> mempath="$mempath/memory.limit_in_bytes"
> 
> memsize=$(cat $mempath)

You cannot always just assume the proc to be '1' always each PID can have its own limits based on which groupname the process is associated with.  The current code handles all scenarios not just defaulting to the init process memory limit alone.  

Process id is needed to know the exact groupname as well while on docker it is not necessary since the process you are running i.e minio is always pid "1"  - But this code caters for process outside containers i.e on a regular operating system as well.

pkg/sys/cgroup_linux.go
harshavardhana: done

pkg/sys/cgroup_linux.go
harshavardhana: done

krishnasrinivas (approved):

balamurugana:
> You cannot always just assume the proc to be '1' always each PID can have its own limits based on which groupname the process is associated with. The current code handles all scenarios not just defaulting to the init process memory limit alone.
> 
> Process id is needed to know the exact groupname as well while on docker it is not necessary since the process you are running i.e minio is always pid "1" - But this code caters for process outside containers i.e on a regular operating system as well.

I was merely showing a logic not to use PID 1.

This patch has unnecessary types and parsing code to find PID's memory path segment.

harshavardhana:
> This patch has unnecessary types and parsing code to find PID's memory path segment.

Let me see if i can further simplify this.

balamurugana:
Below is the equivalent go code.  These functions are better unit testable.

```go
func getProcMemPathSegment() (memPathSegment string, err error) {
	pidCGroupFile := fmt.Sprintf("/proc/%d/cgroup", os.Getpid())
	file, err := os.Open(pidCGroupFile)
	if err != nil {
		return memPathSegment, err
	}
	defer file.Close()

	scanner := bufio.NewScanner(file)
	for scanner.Scan() {
		line := scanner.Text()

		tokens := strings.SplitN(line, ":", 2)
		if len(tokens) != 2 {
			continue
		}

		tokens = strings.SplitN(line, ":", 2)
		if len(tokens) != 2 {
			continue
		}

		if tokens[0] == "memory" {
			memPathSegment = tokens[1]
			break
		}
	}
	err = scanner.Err()

	return memPathSegment, err
}

const cgroupMemoryPathPrefix = "/sys/fs/cgroup/memory"

func getProcMemFilename() (string, error) {
	memPathSegment, err := getProcMemPathSegment()
	if err != nil {
		return "", err
	}

	memFile := cgroupMemoryPathPrefix
	if !strings.HasPrefix(memPathSegment, "/docker/") {
		memFile = filepath.Join(memFile, memPathSegment)
	}
	memFile = filepath.Join(memFile, "memory.limit_in_bytes")

	return memFile, nil
}

func getProcMemLimit(memFile string) (limit int, err error) {
	data, err := ioutil.ReadFile(memFile)
	if err != nil {
		return limit, err
	}

	dataString := strings.TrimSpace(string(data))
	limit, err = strconv.Atoi(dataString)

	return limit, err
}
```

harshavardhana:
It would be neat to send a PR to my branch @balamurugana

harshavardhana:
> tokens := strings.SplitN(line, ":", 2)
> 		if len(tokens) != 2 {
> 			continue
> 		}
> 
> 		tokens = strings.SplitN(line, ":", 2)
> 		if len(tokens) != 2 {
> 			continue
> 		}
> 
> 		if tokens[0] == "memory" {
> 			memPathSegment = tokens[1]
> 			break
> 		}

There are bugs in this there is a changes that tokens have "name=" in some of the cgroup versions, which leads to an issue where "memory" string is never found. 

Let me bring changes in cgroup only to look for `memory`

harshavardhana:
can you review this again @balamurugana .

cmd/server-rlimit.go
balamurugana: But why can't it do `os.Getpid()` inside `sys.GetStats()` ?

harshavardhana: It can @balamurugana but it didn't look idiomatic to me. It looks little clumsy.

balamurugana: As `sys` package is minio specific, I don't think generic PID specific is necessary.  Better `sys` should return cgroup memory limit of current process than some other process.

pkg/cgroup/cgroup_linux.go
```diff
@@ -0,0 +1,260 @@
...
+	// Default docker prefix.
+	dockerPrefixName = "/docker/"
+
+	// Proc controller group path.
+	procCGroupPath = "/proc/%d/cgroup"
+
+	// CGroup entry separator
+	cgroupEntrySep   = ":"
+	cgroupNameSep    = ","
+	cgroupNamePrefix = "name="
```
balamurugana: As they are specific to file format, these constants are overkill.  They make `parseProcCGroup()` function harder to read and understand.

pkg/cgroup/cgroup_linux.go
```diff
@@ -0,0 +1,260 @@
...
+// DO NOT EDIT following constants are chosen defaults for any kernel
+// after 3.x, please open a github issue https://github.com/minio/minio/issues
+// and discuss first if you wish to change this.
+const (
+	// Default value for any error or if cgroup is not set.
+	// this is the defautl if cgroups are not used/configured.
+	unlimitedMemoryLimit = 9223372036854771712
+
+	// Default string for looking for kernel memory param.
+	memoryLimitInBytes = "memory.limit_in_bytes"
```
balamurugana: I see this is a kernel parameter name.  Better name could be `memoryLimitKernelParam`?

pkg/cgroup/cgroup_linux.go
```diff
@@ -0,0 +1,260 @@
...
+
+	// Points to sys path memory path.
+	cgroupMemSysPath = "/sys/fs/cgroup/memory"
+
+	// Default mem controller name.
+	cgroupMemName = "memory"
```
balamurugana: As this constant is used at only one place, there is no need to be a constant.

pkg/cgroup/cgroup_linux.go
```diff
@@ -0,0 +1,260 @@
...
+// at /proc/<pid>/cgroup.
+type CGroup struct {
+	ID   int    // Unique cgroup ID (Assigned usually by kernel).
+	Name string // Name of cgroup type.
+	Path string // Path in /sys/fs/cgroup.
+}
+
+// GetCGByName - returns all the cgroups matching the controller name
+// of the cgroup entry.
+func (pcg ProcCGroup) GetCGByName(name string) (res []CGroup) {
```
balamurugana: Group this function to the type.

pkg/cgroup/cgroup_linux.go
```diff
@@ -0,0 +1,260 @@
...
+//
+// Following code block parses this information and
+// returns a procCGroup which is a parsed list of all
+// the line by line entires from /proc/<pid>/cgroup.
+func parseProcCGroup(r io.Reader) (ProcCGroup, error) {
+	var proc ProcCGroup
+
+	// Start reading cgroup categories line by line
+	// and process them into procCGroup structure.
+	fileScanner := bufio.NewScanner(r)
```
balamurugana: As this function gets a `io.Reader`, `scanner` is appropriate.

pkg/cgroup/cgroup_linux.go
```diff
@@ -0,0 +1,260 @@
...
+)
+
+// ProcCGroup - represents all the entries in a process cgroup file
+// at /proc/<pid>/cgroup.
+type ProcCGroup []CGroup
+
+// CGroup - represents the each cgroup entry in process cgroup file
+// at /proc/<pid>/cgroup.
+type CGroup struct {
+	ID   int    // Unique cgroup ID (Assigned usually by kernel).
```
balamurugana: * How useful `ID` is? I don't think it's useful practically
* If `ID` is removed, it would better to have `CGroup`  as `map[string][]string`
* I feel `CGroup` name is too generic.  As the file `/proc/$PID/cgroup` contains process's cgroup membership, how about better naming?

pkg/cgroup/cgroup_linux.go
```diff
@@ -0,0 +1,260 @@
...
+	for fileScanner.Scan() {
+		line := fileScanner.Text()
+
+		tokens := strings.SplitN(line, cgroupEntrySep, 3)
+		if len(tokens) < 3 {
+			continue
+		}
+
+		id, name, path := tokens[0], tokens[1], tokens[2]
+		for len(name) != 0 {
```
balamurugana: Below is better than `strings.SplitN()` for loop.

```go
for token := range strings.Split(name, cgroupNameSep) {
    name := strings.TrimPrefix(token, cgroupNamePrefix)
    n, err := strconv.Atoi(id)
    if err != nil {
        return nil, err
    }
    proc = append(proc, CGroup{id, name, path})
}
```

pkg/cgroup/cgroup_linux.go
```diff
@@ -0,0 +1,260 @@
...
+	cgmProgGetValue = "getvalue"
+)
+
+// Fetch value of the cgroup kernel param from the cgroup manager,
+// if cgroup manager is configured we should just rely on `cgm` cli
+// to fetch all the values for us.
+func getCGroupManagerKernValue(cg CGroup, kernParam string) (limit uint64, err error) {
+	limit = unlimitedMemoryLimit
+
+	cmd := exec.Command(cgmProgName, cgmProgGetValue, cg.Name, cg.Path, kernParam)
```
balamurugana: You could use string literal than the constant to better readability.

pkg/cgroup/cgroup_linux.go
```diff
@@ -0,0 +1,260 @@
...
+	var out bytes.Buffer
+	cmd.Stdout = &out
+	if err = cmd.Run(); err != nil {
+		return limit, err
+	}
+
+	var v uint64
+	v, err = strconv.ParseUint(strings.TrimSpace(out.String()), 10, 64)
```
balamurugana: As this function is generic enough, for any kernelParam, is the output always uint64?

pkg/cgroup/cgroup_linux.go
```diff
@@ -0,0 +1,260 @@
...
+	// Action param name used to fetch value for a particular
+	// controller.
+	cgmProgGetValue = "getvalue"
+)
+
+// Fetch value of the cgroup kernel param from the cgroup manager,
+// if cgroup manager is configured we should just rely on `cgm` cli
+// to fetch all the values for us.
+func getCGroupManagerKernValue(cg CGroup, kernParam string) (limit uint64, err error) {
+	limit = unlimitedMemoryLimit
```
balamurugana: As `unlimitedMemoryLimit` is treated as return value when `err != nil`, you could better return `0` than `uint64(-1)`.  There is no meaning to `limit` when `err != nil`

pkg/cgroup/cgroup_linux.go
```diff
@@ -0,0 +1,260 @@
...
+		return limit, err
+	}
+
+	memory := pcg.GetCGByName(cgroupMemName)
+	if len(memory) == 0 {
+		return limit, errors.New("Unable to locate any memory limits")
+	}
+
+	// Pick the first entry in memory controller.
+	cg := memory[0]
```
balamurugana: Why ignore others except first entry?  If you are expecting only one entry for memory, you could error out ie multiple memory entires is a unknown situation this may send wrong information.

pkg/cgroup/cgroup_linux.go
```diff
@@ -0,0 +1,260 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Package cgroup implements parsing for all the cgroup
+// categories and functionality in a simple way.
+package cgroup
```
balamurugana: As functions in this package are scoped to its limit, there is no need to have them `CGroup` prefixed

balamurugana (changes requested):

pkg/cgroup/cgroup_linux.go
harshavardhana: Yes that is correct this value is always non-negative uint64 value.

pkg/cgroup/cgroup_linux.go
harshavardhana: Expectation is only one entry,  error is better.

pkg/cgroup/cgroup_linux.go
harshavardhana: Sure

pkg/cgroup/cgroup_linux.go
harshavardhana: Sure

pkg/cgroup/cgroup_linux.go
harshavardhana: Sure

pkg/cgroup/cgroup_linux.go
harshavardhana: Yeah better.

pkg/cgroup/cgroup_linux.go
harshavardhana: Will fix it.

pkg/cgroup/cgroup_linux.go
harshavardhana: unlimitedMemoryLimit - is the default value if cgroup is not configured 

```
$ cat /sys/fs/cgroup/memory/memory.limit_in_bytes 
9223372036854771712
```

Just helps in handling the top level code.

pkg/cgroup/cgroup_linux.go
harshavardhana: Yep will do.

pkg/cgroup/cgroup_linux.go
harshavardhana: Sure.. the whole thing is completely changed now.

pkg/cgroup/cgroup_linux.go
harshavardhana: Sure

pkg/cgroup/cgroup_linux.go
harshavardhana: Makes sense.

harshavardhana:
Fixed all the comments @balamurugana - ready for one more round.

harshavardhana:
Ping @balamurugana anything else needed here ?

pkg/cgroup/linux.go
```diff
@@ -0,0 +1,213 @@
...
+	"strings"
+)
+
+// DO NOT EDIT following constants are chosen defaults for any kernel
+// after 3.x, please open a github issue https://github.com/minio/minio/issues
+// and discuss first if you wish to change this.
+const (
+	// Default value for any error or if cgroup is not set.
+	// this is the defautl if cgroups are not used/configured.
+	unlimitedMemoryLimit = 9223372036854771712
```
balamurugana: As you see consumer functions of this const, its used to be returned when `err != nil` ie on success file read or command execution, it returns the read value.

In an error situation, returning any value as limit is (has to be) discarded.  So there is no need of this const.

pkg/cgroup/linux.go
```diff
@@ -0,0 +1,213 @@
...
+	// Proc controller group path.
+	procCGroupPath = "/proc/%d/cgroup"
+)
+
+// CGEntries - represents all the entries in a process cgroup file
+// at /proc/<pid>/cgroup.
+type CGEntries map[string][]string
+
+// GetCGPathByName - returns the path matching the controller name.
+func (cg CGEntries) GetCGPathByName(cname string) (path string) {
```
balamurugana: I think this function is not necessary. As map is its base type.  The consumer can directly do `cg[key]`, its up to the consumer what to do with value as string slice.

In case of `GetCGroupMemoryLimit()`, it could error out if more than one value in slice as we agreed previously.

pkg/cgroup/linux.go
```diff
@@ -0,0 +1,213 @@
...
+	memoryLimitKernelParam = "memory.limit_in_bytes"
+
+	// Points to sys path memory path.
+	cgroupMemSysPath = "/sys/fs/cgroup/memory"
+
+	// Default docker prefix.
+	dockerPrefixName = "/docker/"
+
+	// Proc controller group path.
+	procCGroupPath = "/proc/%d/cgroup"
```
balamurugana: `cgroupFileTemplate` would be better.

pkg/cgroup/linux.go
```diff
@@ -0,0 +1,213 @@
...
+
+		tokens := strings.SplitN(line, ":", 3)
+		if len(tokens) < 3 {
+			continue
+		}
+
+		_, name, path := tokens[0], tokens[1], tokens[2]
+		for _, token := range strings.Split(name, ",") {
+			name = strings.TrimPrefix(token, "name=")
+			proc[name] = []string{path}
```
balamurugana: if `name` already exists in `proc`, previous value is lost.

pkg/cgroup/linux.go
```diff
@@ -0,0 +1,213 @@
...
+
+// Control group manager CLI defaults.
+const (
+	// Change this if at all in future control group manager
+	// changes their cli name.
+	cgmProgName = "cgm"
+
+	// Action param name used to fetch value for a particular
+	// controller.
+	cgmProgGetValue = "getvalue"
```
balamurugana: I think you said of removing these consts.

pkg/cgroup/linux.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+
+	var v uint64
+	v, err = strconv.ParseUint(strings.TrimSpace(out.String()), 10, 64)
+	if err != nil {
+		return limit, err
+	}
+
+	limit = v
+	return limit, nil
```
balamurugana: Below code is enough, as no one supposes to use `limit` in case of error is set.

```go
limit, err = strconv.ParseUint(strings.TrimSpace(out.String()), 10, 64)
return limit, err
```

little line saver is
```go
    var out bytes.Buffer
    cmd := exec.Command(cgmProgName, cgmProgGetValue, cname, path, kernParam)
    cmd.Stdout = &out
    if err = cmd.Run(); err == nil {
        limit, err = strconv.ParseUint(strings.TrimSpace(out.String()), 10, 64)
    }

    return limit, err
```

pkg/cgroup/linux.go
```diff
@@ -0,0 +1,213 @@
...
+		var b []byte
+		b, err = ioutil.ReadFile(getCGroupMemoryLimitFilePath(path))
+		if err != nil {
+			return limit, err
+		}
+
+		v, err = strconv.ParseUint(strings.TrimSpace(string(b)), 10, 64)
+		if err != nil {
+			return limit, err
+		}
```
balamurugana: Below code is enough, as no one supposes to use `limit` in case of error is set.

```go
if limit, err = getCGroupManagerKernValue("memory", path, memoryLimitKernelParam); err == nil {
    return limit, err
}

var b []byte
if b, err = ioutil.ReadFile(getCGroupMemoryLimitFilePath(path)); err == nil {
    limit, err = strconv.ParseUint(strings.TrimSpace(string(b)), 10, 64)
}
return limit, err
```

pkg/cgroup/linux_test.go
```diff
@@ -0,0 +1,94 @@
...
+9:hugetlb:/
+8:net_cls,net_prio:/
+7:perf_event:/
+6:pids:/user.slice/user-1000.slice
+5:devices:/user.slice
+4:cpuset:/
+3:cpu,cpuacct:/user.slice
+2:freezer:/
+1:name=systemd:/user.slice/user-1000.slice/session-1.scope
+`
```
balamurugana, line 47: You could add more case for multiple values for same keys.

pkg/cgroup/linux.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+
+	// Final path.
+	return filepath.Join(path, memoryLimitKernelParam)
+}
+
+// GetCGroupMemoryLimit - Fetches cgroup memory limit either
+// from a file path at '/sys/fs/cgroup/memory', if path fails then
+// fallback to querying cgroup manager.
+func GetCGroupMemoryLimit(pid int) (limit uint64, err error) {
```
balamurugana: This exposed function `GetCGroupMemoryLimit` can be `GetMemoryLimit` to be used as `cgroup.GetMemoryLimit()`

pkg/cgroup/linux.go
```diff
@@ -0,0 +1,213 @@
...
+
+	// Action param name used to fetch value for a particular
+	// controller.
+	cgmProgGetValue = "getvalue"
+)
+
+// Fetch value of the cgroup kernel param from the cgroup manager,
+// if cgroup manager is configured we should just rely on `cgm` cli
+// to fetch all the values for us.
+func getCGroupManagerKernValue(cname, path, kernParam string) (limit uint64, err error) {
```
balamurugana: Does `cgm` gives other than `/sys/fs/cgroup/$MEMPATH/memory.limit_in_bytes`?
Is there a chance of missing `/sys/fs/cgroup/$MEMPATH/memory.limit_in_bytes` file?

pkg/sys/stats_bsd.go
```diff
@@ -39,7 +39,7 @@ func getHwPhysmem() (uint64, error) {
 }
 
 // GetStats - return system statistics for bsd.
-func GetStats() (stats Stats, err error) {
+func GetStats(pid int) (stats Stats, err error) {
```
balamurugana: `pid` parameter should be removed.

balamurugana (changes requested):

pkg/cgroup/linux.go
harshavardhana: No there isn't.

pkg/sys/stats_bsd.go
harshavardhana: Oh yes.

pkg/cgroup/linux.go
harshavardhana: Agreed.

pkg/cgroup/linux.go
harshavardhana: Cgm doesn't read this file it makes UNIX socket connection to cgm manager which returns it over a DBus call.

pkg/cgroup/linux.go
harshavardhana: This style was avoided on purpose as Krishna suggested on readability loss.

pkg/cgroup/linux_test.go
harshavardhana: Will do

pkg/cgroup/linux.go
harshavardhana: This style of coding is avoided because of readability argument Krishna posted earlier. Let me see how I can refactor it properly.

pkg/cgroup/linux.go
balamurugana: I couldn't able to get what readability lost.

harshavardhana: basically the idea is to handle the error case first not success.

pkg/cgroup/linux.go
harshavardhana: Sure

pkg/cgroup/linux.go
harshavardhana: Yes, forgot.

pkg/cgroup/linux.go
harshavardhana: Fixed.

pkg/cgroup/linux_test.go
harshavardhana: Multiple values for same keys are not possible, but multiple keys are possible for the same value.  Like `cpu,cpuacct`

harshavardhana: Same process can have different controllers in different groups as well. But ending path is always unique

pkg/cgroup/linux.go
harshavardhana: Fixed

pkg/cgroup/linux.go
harshavardhana: Fixed

pkg/cgroup/linux.go
harshavardhana: Agreed

pkg/cgroup/linux.go
harshavardhana: Sure

harshavardhana:
ping @balamurugana - PR updated once again with the changes requested. Can you check again?

pkg/cgroup/linux.go
```diff
@@ -0,0 +1,177 @@
...
+	// Default docker prefix.
+	dockerPrefixName = "/docker/"
+
+	// Proc controller group path.
+	cgroupFileTemplate = "/proc/%d/cgroup"
+)
+
+// CGEntries - represents all the entries in a process cgroup file
+// at /proc/<pid>/cgroup as key value pairs.
+type CGEntries map[string]string
```
balamurugana, line 55: :+1:

balamurugana (approved):

PR #4006, opened 2017-03-30
Add `access` format support for Elasticsearch notification target

### LINKED ISSUE #3928: notification: [feature] Support `access` event format for database backends.

donatello:
Support `access` format for elasticsearch and upgrade support to version 5.x of elasticsearch.

## Description

This change adds `access` format support for notifications to a
Elasticsearch server, and it refactors `namespace` format support.

In the case of `access` format, for each event in Minio, a JSON
document is inserted into Elasticsearch with its timestamp set to the
event's timestamp, and with the ID generated automatically by
elasticsearch. No events are modified or deleted in this mode.

In the case of `namespace` format, for each event in Minio, a JSON
document is keyed together by the bucket and object name is updated in
Elasticsearch. In the case of an object being created or over-written
in Minio, a new document or an existing document is inserted into the
Elasticsearch index. If an object is deleted in Minio, the
corresponding document is deleted from the Elasticsearch index.

Additionally, this change upgrades Elasticsearch support to the 5.x
series. This is a breaking change, and users of previous elasticsearch
versions should upgrade.

Also updates documentation on Elasticsearch notification target usage
and has a link to an elasticsearch upgrade guide.

This is the last patch that finally resolves #3928.

## Motivation and Context

#3928 

## How Has This Been Tested?

Manually tested using elasticsearch docker image from https://www.elastic.co/guide/en/elasticsearch/reference/current/docker.html with a command like:

```
docker run -p 9200:9200 -e "http.host=0.0.0.0" -e "transport.host=127.0.0.1" -e "xpack.security.enabled=false" docker.elastic.co/elasticsearch/elasticsearch:5.3.0
```

The docker images comes bundled with basic authentication through Xpack, which was disabled based on the docs at https://www.elastic.co/guide/en/x-pack/5.3/security-settings.html 

## Types of changes

## Checklist:


vadmeste (approved):
LGTM

cmd/notify-elasticsearch.go
```diff
@@ -129,25 +135,43 @@ func (q elasticClient) Fire(entry *logrus.Entry) error {
...
 		}
-		return nil
-	} // else we update elastic index or create a new one.
-	_, err := q.Client.Index().Index(q.params.Index).
-		Type("event").
-		BodyJson(map[string]interface{}{
-			"Records": entry.Data["Records"],
-		}).Id(keyStr).Do()
-	return err
+		eventTimeStr := fmt.Sprintf("%d", eventTime.UnixNano()/1000)
```
harshavardhana: why not just use `eventTime.Unix()` ?

harshavardhana (changes requested):

cmd/notify-elasticsearch.go
```diff
@@ -129,25 +135,43 @@ func (q elasticClient) Fire(entry *logrus.Entry) error {
...
+			Type("event").
+			BodyJson(map[string]interface{}{
+				"Records": entry.Data["Records"],
+			}).Id(keyStr).Do(context.Background())
+	case formatAccess:
+		// eventTime is taken from the first entry in the
+		// records.
+		events, ok := entry.Data["Records"].([]NotificationEvent)
+		if !ok {
+			return makeESError("unable to extract event time due to conversion error of entry.Data[\"Records\"]=%v", entry.Data["Records"])
```
harshavardhana: "unable" --> "Unable"

cmd/notify-elasticsearch.go
```diff
@@ -129,25 +135,43 @@ func (q elasticClient) Fire(entry *logrus.Entry) error {
...
+		}
+		var eventTime time.Time
+		eventTime, err = time.Parse(timeFormatAMZ, events[0].EventTime)
 		if err != nil {
-			return err
+			return makeESError("unable to parse event time \"%s\": %v",
```
harshavardhana: "unable" --> "Unable"

harshavardhana (changes requested):

cmd/notify-elasticsearch.go
donatello: Because ES wants it in milliseconds.

cmd/notify-elasticsearch.go
donatello: Done.

cmd/notify-elasticsearch.go
donatello: Done.

cmd/notify-elasticsearch.go
harshavardhana: Perhaps better to add a comment

donatello: Done.

harshavardhana (approved):

cmd/notify-elasticsearch.go
```diff
@@ -17,14 +17,20 @@
...
+	"time"
 
 	"github.com/Sirupsen/logrus"
-	"github.com/minio/sha256-simd"
-	"gopkg.in/olivere/elastic.v3"
+	"gopkg.in/olivere/elastic.v5"
+)
+
+var (
+	makeESError = makeNotificationErrorMaker("Elasticsearch Notifier Error")
```
harshavardhana: Actually we should just keep it as "Elasticsearch:" and for everything as well.  Calling it "Elasticsearch Notifier Error" is just verbose and redundant. 

Because "Elasticsearch" already means that it is notifier and "Error:" is not needed since you are encapsulating these messages inside `fmr.Errorf()` any ways. Unless we have another set of messages indicating "Success:" we don't it.

cmd/notifiers.go
```diff
@@ -221,3 +222,16 @@ func filterRuleMatch(object string, frs []filterRule) bool {
 	}
 	return prefixMatch && suffixMatch
 }
+
+// A type to represent dynamic error maker functions for
+// notifications.
+type notificationErrorMaker func(string, ...interface{}) error
```
harshavardhana: This should be called  `notificationErrorFactoryFunc` instead. Factory is a more common word than Maker.

cmd/notifiers.go
```diff
@@ -221,3 +222,16 @@ func filterRuleMatch(object string, frs []filterRule) bool {
...
+
+// A function to build dynamic error functions for notifications by
+// setting an error prefix string.
+func makeNotificationErrorMaker(prefix string) notificationErrorMaker {
```
harshavardhana: this can be just 

```
newNotificationErrorFactory(prefix string) notificationErrorFactoryFunc {
}
```

Take a look at `listDirFactory()`

cmd/notify-elasticsearch.go
```diff
@@ -17,14 +17,20 @@
...
+
+	esNFormatError = makeESError(`"format" value is invalid - it must be one of "%s" or "%s".`, formatNamespace, formatAccess)
```
harshavardhana: the format for error strings is . `err` is started at the beginning - for example look at `typed-errors.go` 

```
errESFormat
errESIndex
```

cmd/notify-mysql.go
```diff
@@ -90,12 +90,9 @@ VALUES (?, ?);`
 	tableExistsMySQL = `SELECT 1 FROM %s;`
 )
 
-func makeMySQLError(msg string, a ...interface{}) error {
-	s := fmt.Sprintf(msg, a...)
-	return fmt.Errorf("MySQL Notifier Error: %s", s)
-}
-
 var (
+	makeMySQLError = makeNotificationErrorMaker("MySQL Notifier Error")
```
harshavardhana: This becomes 

```
mysqlErrFunc = newNotificationErrorFactory("MySQL")

errMySQLFormat = mysqlErrFunc("....")
```

cmd/notify-redis.go
```diff
@@ -18,20 +18,16 @@ package cmd
...
 	"github.com/garyburd/redigo/redis"
 )
 
-func makeRedisError(msg string, a ...interface{}) error {
-	s := fmt.Sprintf(msg, a...)
-	return fmt.Errorf("Redis Notifier Error: %s", s)
-}
-
 var (
+	makeRedisError = makeNotificationErrorMaker("Redis Notifier Error")
```
harshavardhana: You have to follow the above format mentioned everywhere..

harshavardhana (changes requested):

donatello:
@harshavardhana Did all the changes now.

donatello:
Fixed milliseconds since epoch calculation for elasticsearch timestamp.

harshavardhana (approved):

PR #4008, opened 2017-03-30
Add `deliveryMode` parameter for AMQP notfication target

### LINKED ISSUE #3982: [notify:amqp] Missing DeliveryMode
There is no way to use DeliveryMode.

## Expected Behavior
```
  "notify": {
    "amqp": {
      "1": {
        "enable": true,
        "url": "amqp://rabbitmq:rabbitmq@192.168.99.100:5672",
        "exchange": "bucketevents",
        "routingKey": "import",
        "exchangeType": "direct",
-->   "deliveryMode": 2, // queue implementation use - non-persistent (1) or persistent (2)` <---
        "mandatory": false,
        "immediate": false,
        "durable": true,
        "internal": false,
        "noWait": false,
        "autoDeleted": false
      }
}
```

## Current Behavior
 ```
 "notify": {
    "amqp": {
      "1": {
        "enable": true,
        "url": "amqp://rabbitmq:rabbitmq@192.168.99.100:5672",
        "exchange": "bucketevents",
        "routingKey": "import",
        "exchangeType": "direct",
        "mandatory": false,
        "immediate": false,
        "durable": true,
        "internal": false,
        "noWait": false,
        "autoDeleted": false
      }
}
```

## Possible Solution
Add the field deliveryMode to supported config fields. It's implemented in streadway/amqp
So the implemtation should look like this.

```
	amqp.Publishing{
			Body:            []byte(body),
			DeliveryMode:    amqp.Transient, // 1=non-persistent, 2=persistent
		},
```

## Context
I need to store rabbitmq messages persistent to disk. This is not possible without setting the deliveryMode. 
Sorry i'm no go expert and can't do a pull request.

### PR DESCRIPTION
**Pull request by donatello**
Fixes #3982

## Description

Adds deliveryMode parameter for amqp.

## Motivation and Context

Fixes #3982

## Types of changes

## Checklist:

### CODE REVIEW

File: cmd/notify-amqp.go [hunk 1]
```diff
@@ -32,6 +32,7 @@ type amqpNotify struct {
 	Exchange     string `json:"exchange"`
 	RoutingKey   string `json:"routingKey"`
 	ExchangeType string `json:"exchangeType"`
+	DeliveryMode uint8  `json:"deliveryMode"`
```
vadmeste, line 35: Can we add a documentation somewhere about the possible values that this field can take?. Besides there is no validation of this field, which seems to take only (0, 1) for Transient and 2 for Persistent.

krishnasrinivas (approved):

cmd/notify-amqp.go
donatello: It is just using the types in the underlying amqp library. I am not sure wrapping it as a bool (with false for transient and true for persistent) is appropriate because I don't know if this parameter may change in the future. The default value of 0 is implicit and works correctly.

I will add some documentation about this parameter.

donatello:
Added documentation about parameters for AMQP

donatello:
Configuration doc & example also updated.

vadmeste (approved):

harshavardhana (approved):

PR #4011, opened 2017-03-30
notifiers: Stop using url.Parse in validating address format.

harshavardhana:
## Description
url.Parse() wrongly parses an address of format "address:port"
which is fixed in go1.8.  This inculcates a breaking change
on our end. We should fix this wrong usage everywhere so that
migrating to go1.8 eventually becomes smoother.

## Motivation and Context
Fixes future go1.8 issues - https://github.com/golang/go/commit/c5ccbdd22bdbdc43d541b7e7d4ed66ceb559030e#comments

## How Has This Been Tested?
Manually using go1.8

## Types of changes

## Checklist:

### DIFF
--- a/cmd/notify-nats.go
+++ b/cmd/notify-nats.go
@@ -18,6 +18,7 @@ package cmd
 
 import (
 	"io/ioutil"
+	"net"
 
 	"github.com/Sirupsen/logrus"
 	"github.com/nats-io/go-nats-streaming"
@@ -52,7 +53,7 @@ func (n *natsNotify) Validate() error {
 	if !n.Enable {
 		return nil
 	}
-	if _, err := checkURL(n.Address); err != nil {
+	if _, _, err := net.SplitHostPort(n.Address); err != nil {
 		return err
 	}
 	return nil
--- a/cmd/admin-handlers_test.go
+++ b/cmd/admin-handlers_test.go
@@ -1377,7 +1377,7 @@ func TestWriteSetConfigResponse(t *testing.T) {
 		},
 	}
 
-	testURL, err := url.Parse("dummy.com")
+	testURL, err := url.Parse("http://dummy.com")
 	if err != nil {
 		t.Fatalf("Failed to parse a place-holder url")
 	}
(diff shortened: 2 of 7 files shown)


cmd/notify-nats.go
```diff
@@ -52,7 +53,7 @@ func (n *natsNotify) Validate() error {
 	if !n.Enable {
 		return nil
 	}
-	if _, err := checkURL(n.Address); err != nil {
+	if _, _, err := net.SplitHostPort(n.Address); err != nil {
```
vadmeste, line 56: :+1:, NATs already supports only this format.

vadmeste: > , NATs already supports only this format.

SplitHostPort requires port, doesn't it? it means it can break existing configuration.

vadmeste:
I think the best solution for this is to require scheme in all addresses of all notification targets, the scheme can indicate if this is plain or secured connection, so it can be useful.

cmd/notify-nats.go
harshavardhana: Yes port should be provided..  existing configs are broken anyways because of the url usage.

harshavardhana:
> I think the best solution for this is to require scheme in all addresses of all notification targets, the scheme can indicate if this is plain or secured connection, so it can be useful.

Scheme will not be useful because redis, nats, kafka are not performing HTTP based requests and do not have HTTP endpoints. So a secure connection is not meaningful.

vadmeste:
> Scheme will not be useful because redis, nats, kafka are not performing HTTP based requests and do not have HTTP endpoints. So a secure connection is not meaningful.

Scheme could be something other than http. NATs already requires "nats://" or "tls://" for secure connections but actually Minio adds that scheme before calling NATs API.

Anyway, the only thing that we can gain from adding a scheme is to enhance validation of urls. Otherwise I don't see any other proper way to validate.

harshavardhana:
> Scheme could be something other than http. NATs already requires "nats://" or "tls://" for secure connections but actually Minio adds that scheme before calling NATs API.

Okay so that's for NATS - how about for redis and kafka?

vadmeste (approved):

cmd/notify-nats.go
krisis: @harshavardhana we can have default ports for the service if port is not specified as opposed to failing due to `net.SplitHostPort`'s requirement right? Why penalise the user for library quirks?

harshavardhana: Can do need to note down default ports for these targets.

krisis, line 56: shouldn't we ignore "port missing" error in this case too?

harshavardhana: oops yet.. forgot.

krisis (approved):

PR #4014, opened 2017-03-31
fix: use its own lock in serverConfigV17

balamurugana:
Previously serverConfigV17 used a global lock that made any instance of
serverConfigV17 depended on single global serverConfigMu.

This patch fixes by having individual lock per instances.

## How Has This Been Tested?

`make test`

## Types of changes

## Checklist:

### DIFF
--- a/cmd/config-v17.go
+++ b/cmd/config-v17.go
@@ -26,17 +26,21 @@ import (
 	"github.com/tidwall/gjson"
 )
 
-// Read Write mutex for safe access to ServerConfig.
-var serverConfigMu sync.RWMutex
-
 // Config version
 const v17 = "17"
+
+var (
+	// serverConfig server config.
+	serverConfig   *serverConfigV17
+	serverConfigMu sync.RWMutex
+)
 
 // serverConfigV17 server configuration version '17' which is like
 // version '16' except it adds support for "format" parameter in
 // database event notification targets: PostgreSQL, MySQL, Redis and
 // Elasticsearch.
 type serverConfigV17 struct {
+	sync.RWMutex
 	Version string `json:"version"`
 
 	// S3 API configuration.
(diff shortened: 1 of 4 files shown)


cmd/gateway-main.go
```diff
@@ -94,7 +94,7 @@ func newGatewayLayer(backendType, accessKey, secretKey string) (GatewayLayer, er
 //
 // DO NOT save this config, this is meant to be
 // only used in memory.
-func newGatewayConfig(accessKey, secretKey, region string) error {
```
harshavardhana: why is this changed? this is needed when we bring the S3 backend it was kept as a placeholder so that region is addressed properly for S3 backends. 

Please keep it the way it was and only move `createConfigDir()` outside.

harshavardhana (changes requested):

cmd/config-v17.go
```diff
@@ -26,17 +26,21 @@ import (
...
-// Read Write mutex for safe access to ServerConfig.
-var serverConfigMu sync.RWMutex
-
 // Config version
 const v17 = "17"
 
+var (
+	// serverConfig server config.
+	serverConfig      *serverConfigV17
+	serverConfigMutex sync.RWMutex
```
krishnasrinivas: in Go, a general convention that I have seen people follow is - for object named xxxx - if it needs lock then the lock is named as xxxxMu

cmd/config-v17.go
```diff
@@ -91,67 +162,25 @@ func newConfig() error {
...
 	}
-
 	if globalIsEnvBrowser {
 		srvCfg.SetBrowser(globalIsBrowserEnabled)
 	}
 
-	// Create config path.
-	if err := createConfigDir(); err != nil {
-		return err
-	}
```
krishnasrinivas, line 106: I see that `createConfigDir()` is getting called in gateway mode, but unable to see it getting called for server mode

harshavardhana: Its called in server-main.go

cmd/gateway-main.go
balamurugana: Done.

Why not add when needed than unused parameter? I think you want default to be passed and set in the config struct even though new config carries the same.

cmd/config-v17.go
balamurugana: Done

harshavardhana (approved):

PR #4019, opened 2017-03-31
Add extensive endpoints validation

balamurugana:
Previously `url.URL` is used as endpoint which does not apply for FS and
XL type of endpoints.  This patch makes type `Endpoint` which is
extended type of url.URL plus additional field of `IsLocal` to know
whether the endpoint points to local host or not.

## How Has This Been Tested?

- make test
- manually

## Types of changes

## Checklist:


cmd/admin-handlers_test.go
```diff
@@ -1308,6 +1269,8 @@ func TestSetConfigHandler(t *testing.T) {
 		t.Fatalf("Failed to decode set config result json %v", err)
 	}
 
+	fmt.Printf("%+v\n", result)
```
donatello: Stray print statement?

harshavardhana:
ping @krishnasrinivas @donatello @krisis - can you guys review this PR?

harshavardhana:
and @vadmeste as well.

cmd/endpoint.go
```diff
@@ -0,0 +1,413 @@
...
+	// PathEndpointType - path style endpoint type enum.
+	PathEndpointType EndpointType = iota + 1
+
+	// URLEndpointType - URL style endpoint type enum.
+	URLEndpointType
+)
+
+// Endpoint - any type of endpoint.
+type Endpoint struct {
+	value   string
```
vadmeste: We probably can merge value and URL fields, then we test on url.Host to see if this is a net URL or a local path

balamurugana: :+1: looks like a good idea. My little worry is if URL object behaver is changed in future, we would need to fix ATM

krishnasrinivas:
@balamurugana the diff is huge, looks tough to review. If it makes sense can you provide any documentation that can help in reviewing the diff?

balamurugana:
> @balamurugana the diff is huge, looks tough to review. If it makes sense can you provide any documentation that can help in reviewing the diff?

I think you are asking what this patch does.  Below is the essence

Previously `url.URL` is used as endpoint which does not apply for FS and
XL type of endpoints.  This patch makes type `Endpoint` which is
extended type of `url.URL` plus additional field of `IsLocal` to know
whether the endpoint points to local host or not.

cmd/admin-handlers_test.go
```diff
@@ -530,11 +507,7 @@ func TestClearLocksHandler(t *testing.T) {
 	defer adminTestBed.TearDown()
 
 	// Initialize admin peers to make admin RPC calls.
-	eps, err := parseStorageEndpoints([]string{"http://127.0.0.1"})
-	if err != nil {
-		t.Fatalf("Failed to parse storage end point - %v", err)
-	}
-	initGlobalAdminPeers(eps)
+	initGlobalAdminPeers(mustGetNewEndpointList("http://127.0.0.1/d1"))
```
krisis: shouldn't we be setting `globalMinioAddr` here?

cmd/admin-handlers_test.go
krisis: This should be removed.

cmd/endpoint.go
```diff
@@ -0,0 +1,414 @@
...
+}
+
+// CreateEndpoints - validates and creates new endpoints for given args.
+func CreateEndpoints(serverAddr string, args ...string) (string, EndpointList, SetupType, error) {
+	var endpoints EndpointList
+	var setupType SetupType
+	var err error
+
+	// Check whether serverAddr is valid for this host.
+	isServerAddrEmpty := (serverAddr == "")
```
krisis: nit: there is not need for paranthesis

cmd/signature-v2.go
```diff
@@ -85,8 +85,12 @@ func doesPresignV2SignatureMatch(r *http.Request) APIErrorCode {
 	cred := serverConfig.GetCredential()
 
 	// r.RequestURI will have raw encoded URI as sent by the client.
-	splits := splitStr(r.RequestURI, "?", 2)
-	encodedResource, encodedQuery := splits[0], splits[1]
+	tokens := strings.SplitN(r.RequestURI, "?", 2)
```
vadmeste, line 88: Not sure why this is changed like that, it looks like the same behavior but with more lines. Can you explain ?

balamurugana: I removed `splitStr()` as its used in two places only and used exactly for two element tokens.  `splitStr()` is necessary utility function.

cmd/endpoint.go
```diff
@@ -0,0 +1,425 @@
...
+
+	"github.com/minio/minio-go/pkg/set"
+)
+
+// EndpointType - enum for endpoint type.
+type EndpointType int
+
+const (
+	// PathEndpointType - path style endpoint type enum.
+	PathEndpointType EndpointType = iota + 1
```
donatello, line 36: Any reason `EndpointType` should not just start with `iota` instead of `iota + 1`?

cmd/signature-v2.go
vadmeste: > I removed splitStr() as its used in two places only and used exactly for two element tokens. splitStr() is necessary utility function.

I guess the real reason is that you don't understand the exact purpose of splitStr function from its name and the code is actually more readable without it, right?

balamurugana: I also read what `splitStr()` does  irrespective of its name and where it was used.  Of course the code is more readable without it.

cmd/endpoint.go
```diff
@@ -0,0 +1,425 @@
...
+		endpoints = append(endpoints, endpoint)
+	}
+
+	sort.Sort(endpoints)
+
+	return endpoints, nil
+}
+
+// CreateEndpoints - validates and creates new endpoints for given args.
+func CreateEndpoints(serverAddr string, args ...string) (string, EndpointList, SetupType, error) {
```
donatello, line 229: This function seems too long at ~200 lines! Since it uses multiple explicit scope blocks (code in braces `{...}` without an outer block statement like `for/switch/if ... {...}`) it seems straightforward to separate them into their own usefully named functions.

cmd/endpoint.go
```diff
@@ -0,0 +1,425 @@
...
+			return serverAddr, endpoints, setupType, fmt.Errorf("port in server address does not match with local endpoints")
+		}
+	}
+
+	setupType = DistXLSetupType
+	return serverAddr, endpoints, setupType, nil
+}
+
+// GetRemotePeers - get hosts information other than this minio service.
+func GetRemotePeers(endpoints EndpointList) []string {
```
donatello, line 364: Better to make this function a method of the `EndPointList` type?

cmd/net.go
```diff
@@ -0,0 +1,196 @@
...
+	host, port, err := net.SplitHostPort(serverAddr)
+	if err != nil {
+		return err
+	}
+
+	// Check whether port is a valid port number.
+	p, err := strconv.Atoi(port)
+	if err != nil {
+		return fmt.Errorf("invalid port number")
+	} else if p < 1 {
```
donatello: p also needs to be less than 65536

cmd/setup-type.go
```diff
@@ -0,0 +1,54 @@
...
+ */
+
+package cmd
+
+// SetupType - enum for setup type.
+type SetupType int
+
+const (
+	// FSSetupType - FS setup type enum.
+	FSSetupType SetupType = iota + 1
```
donatello: Seems like there is no reason not to just start with `iota`

cmd/admin-handlers_test.go
balamurugana: I don't know. The previous code didn't do.

cmd/admin-handlers_test.go
balamurugana: Done

cmd/endpoint.go
balamurugana: Nothing special than to avoid default value of int

cmd/endpoint.go
balamurugana: Do you mean a closure functions?  I thought its not that useful and readable.

cmd/endpoint.go
balamurugana: Yes. Its for readability.  doesn't look good?

cmd/endpoint.go
balamurugana: It was in previous refresh. As its using globals, its not a good idea to be in a type.

cmd/net.go
balamurugana: Done

cmd/setup-type.go
balamurugana: `iota + 1` to avoid default int value `0`

harshavardhana:
ping @krishnasrinivas @krisis @donatello @vadmeste - what is the final review on this PR?

cmd/server-main.go
```diff
@@ -365,62 +120,45 @@ func initConfig() {
...
+		// to IPv6 address ie minio will start listening on IPv6 address whereas another
+		// (non-)minio process is listening on IPv4 of given port.
+		// To avoid this error sutiation we check for port availability only for macOS.
+		fatalIf(checkPortAvailability(globalMinioPort), "Port %d already in use", globalMinioPort)
 	}
 
-	// Set endpoints of []*url.URL type to globalEndpoints.
-	globalEndpoints = endpoints
+	globalIsDistXL = (setupType == DistXLSetupType)
+	globalIsXL = (setupType == XLSetupType)
```
krishnasrinivas: globalIsXL should be true even for dist XL setup:
admin-handlers.go:
```
770         // Check if this setup is an erasure code backend, since                                                                                                                    
771         // heal-format is only applicable to single node XL and                                                                                                                     
772         // distributed XL setup.                                                                                                                                                    
773         if !globalIsXL {                                                                                                                                                            
774                 writeErrorResponse(w, ErrNotImplemented, r.URL)                                                                                                                     
775                 return                                                                                                                                                              
776         }
```

cmd/net.go
```diff
@@ -0,0 +1,196 @@
...
+	if err != nil {
+		// Ignore "missing port in address" error as we use default port.
+		// This happens when hostPort = "" or hostPort = "server".
+		if !strings.Contains(err.Error(), "missing port in address") {
+			return "", "", err
+		}
+
+		host = hostPort
+		port = defaultPort
+		err = nil
```
krishnasrinivas: `err = nil` is not needed

cmd/net.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+
+	p, err := strconv.Atoi(port)
+	if err != nil {
+		return "", "", fmt.Errorf("invalid port number")
+	} else if p < 1 || p > 65536 {
+		return "", "", fmt.Errorf("port number must be between 1 to 65536")
+	}
+
+	return host, port, err
```
krishnasrinivas: can be `return host, port, nil`

cmd/endpoint.go
```diff
@@ -0,0 +1,430 @@
...
+}
+
+// CreateEndpoints - validates and creates new endpoints for given args.
+func CreateEndpoints(serverAddr string, args ...string) (string, EndpointList, SetupType, error) {
+	var endpoints EndpointList
+	var setupType SetupType
+	var err error
+
+	// Check whether serverAddr is valid for this host.
+	isServerAddrEmpty := (serverAddr == "")
```
krishnasrinivas: `serverAddr` will never be `""`

cmd/endpoint.go
```diff
@@ -0,0 +1,430 @@
...
+	if !isServerAddrEmpty {
+		if err = CheckLocalServerAddr(serverAddr); err != nil {
```
krishnasrinivas: this check is already done in server-main.go

cmd/server-main.go
```diff
@@ -365,62 +120,45 @@ func initConfig() {
...
-	sort.Sort(byHostPath(endpoints))
-
-	checkServerSyntax(endpoints, ctx.Args())
-
-	// Should exit gracefully if none of the endpoints passed
-	// as command line args are local to this server.
-	if !isAnyEndpointLocal(endpoints) {
-		fatalIf(errInvalidArgument, "None of the disks passed as command line args are local to this server.")
+	serverAddr := ctx.String("address")
+	if serverAddr == "" {
```
krishnasrinivas: serverAddr will never be empty string

cmd/endpoint.go
```diff
@@ -0,0 +1,430 @@
...
+			return serverAddr, endpoints, setupType, err
+		}
+	}
+
+	// Normalize server address.
+	serverAddrHost, serverAddrPort := mustSplitHostPort(serverAddr)
+	serverAddr = net.JoinHostPort(serverAddrHost, serverAddrPort)
```
krishnasrinivas: `serverAddr = net.JoinHostPort(serverAddrHost, serverAddrPort)`

This is redundant. Because serverAddr will be of the form `:port` or `host:port` split+join will basically result in the same string.

cmd/endpoint.go
```diff
@@ -0,0 +1,430 @@
...
+				endpoints = append(endpoints, endpoint)
+				setupType = FSSetupType
+			} else {
+				err = fmt.Errorf("use path style endpoint for FS setup")
+			}
+		}
+
+		return serverAddr, endpoints, setupType, err
+	}
+
```
krishnasrinivas: this style is more readable:
```
        if len(args) == 1 {                                                                                                                                                             
                endpoint, err := NewEndpoint(args[0])                                                                                                                                   
                if err != nil {                                                                                                                                                         
                        return serverAddr, endpoints, setupType, err                                                                                                                    
                }                                                                                                                                                                       
                if endpoint.Type() != PathEndpointType {                                                                                                                                
                        return serverAddr, endpoints, setupType, fmt.Errorf("use path style endpoint for FS setup")                                                                     
                }                                                                                                                                                                       
                endpoints = append(endpoints, endpoint)
                return serverAddr, endpoints, FSSetupType, nil                                                                                                                            
        }
```

cmd/endpoint.go
```diff
@@ -0,0 +1,430 @@
...
+		sset := set.NewStringSet()
+		for i, endpoint := range endpoints {
+			sset.Add(endpoint.Host)
+			normArgs[i] = endpoint.Path
+		}
+		uniqueHosts = sset.ToSlice()
+	}
+
+	// URL style endpoints are used for XL.
+	if len(uniqueHosts) == 1 {
```
krishnasrinivas: this code is not needed, check https://github.com/balamurugana/minio/pull/5 7b7f24b for details.

cmd/endpoint.go
```diff
@@ -0,0 +1,430 @@
...
+	if endpoints, err = NewEndpointList(args...); err != nil {
+		return serverAddr, endpoints, setupType, err
+	}
+
+	// Return XL setup when all endpoints are path style.
+	if endpoints[0].Type() == PathEndpointType {
+		setupType = XLSetupType
+		return serverAddr, endpoints, setupType, nil
+	}
+
```
krishnasrinivas, line 268: After this point it is only distributed XL is possible

krishnasrinivas (commented):
Review not complete yet. @balamurugana can you take a look at this PR: https://github.com/balamurugana/minio/pull/5 many of the commits there are still valid.

Also it will be easier for reviewers to review fixes if you push review fixes as a new commit instead of amending and force pushing it.

cmd/server-main.go
balamurugana: Why? `--server=` would be empty.

cmd/server-main.go

cmd/net.go
balamurugana: Done

cmd/endpoint.go
balamurugana: It is empty when unit tests call this function.

cmd/endpoint.go
balamurugana: Yes. It is unit test friendly. In later multiple EndpointList will be created to represent different kind of setup in tests.

cmd/endpoint.go
balamurugana: `mustSplitHostPort()`  returns host="" and port=defaultPort when serverAddr="".  This is not a redundant check.

Here is what I think about serverAddr.  If serverAddr is empty, we could bind to all interfaces with `defaultPort`.  If this is not wanted, we could enforce `serverAddr` must be only `":port" or "host:port"` (in server main as well).

Is this wanted?

cmd/endpoint.go
balamurugana: Done

cmd/endpoint.go
balamurugana: We should treat it as XL for below endpoints.

```
[]string{"http://localhost/d1", "http://localhost/d2", "http://localhost/d3", "http://localhost/d4"}
```

balamurugana:
@krishnasrinivas 

> Also it will be easier for reviewers to review fixes if you push review fixes as a new commit instead of amending and force pushing it.

:+1:

donatello (approved):

donatello:
Looks good to me currently, but looks like there are more changes coming to this PR.

krishnasrinivas:
@balamurugana 
https://github.com/krishnasrinivas/wikinotes/wiki/Running-distributed-XL-on-localhost

We should be able to simulate distributed XL on localhost as described in that document.

Using this PR it simulates singlenode-XL when minio is run as described in the above document.

Try it out using a binary from master branch and see how it behaves.

krishnasrinivas:
Discussed with AB on how we should behave when command is run as:
```
minio server http://locahost/mnt/export/disk1 http://locahost/mnt/export/disk2 http://locahost/mnt/export/disk3 http://locahost/mnt/export/disk4
```

The code should flow naturally into a distributed XL code (instead of returning early with setupType as as XLSetupType we should return DistXLSetupType) and work as expected.

But this can be done in a different patch if you think it will complicate this patch.

balamurugana:
> $ minio server http://locahost/mnt/export/disk1 http://locahost/mnt/export/disk2 http://locahost/mnt/export/disk3 http://locahost/mnt/export/disk4
>
> The code should flow naturally into a distributed XL code (instead of returning early with setupType as as XLSetupType we should return DistXLSetupType) and work as expected.
> 
> But this can be done in a different patch if you think it will complicate this patch.

I am not able to understand this.
Are you saying this is DistXL setup, but later it should be taken care as XL setup due to all endpoints are localhost?

cmd/update-notifier.go
```diff
@@ -27,6 +27,58 @@ import (
 	"github.com/fatih/color"
 )
 
+// humanizeDuration converts time.Duration to human readable duration.
```
harshavardhana: The structure is kept previously so that it can be passed around easily just the same argument as why BrowserFlag should be boolean wrapper. I don't know how is the choice driven but i would prefer humanized duration to be a proper parsed value and to have a String() method which makes it print properly in a set format.

All the more this PR has nothing to do with this, i am not why you are bringing unrelated changes in this PR.

harshavardhana (changes requested):

cmd/update-notifier.go
balamurugana: Reverting this commit and making it as a fresh PR.

I see the struct `humanizedDuration()` is used only in `colorizeUpdateMessage()` for printing duration and nowhere I see its used.  For that a conversion function is good enough to go than passing around a struct.

harshavardhana: We can perhaps remove humanize duration and just use https://github.com/dustin/go-humanize/blob/master/times.go

balamurugana: The new PR is https://github.com/minio/minio/pull/4071

I have slightly changed humanization by having only higher component of time.Duration.  If its good to go, the package can't be used.

harshavardhana:
@balamurugana can you resolve conflicts in this PR? . @krishnasrinivas @donatello @krisis @vadmeste can you guys review it once more?

krishnasrinivas:
@harshavardhana yes

cmd/setup-type.go
```diff
@@ -0,0 +1,54 @@
...
+
+	// XLSetupType - XL setup type enum.
+	XLSetupType
+
+	// DistXLSetupType - Distributed XL setup type enum.
+	DistXLSetupType
+)
+
```
krishnasrinivas: @balamurugana looks like the below two functions String() and IsValid() are not used anywhere?

balamurugana: * Removing `IsValid()`.
* `String()` is used automagically in print functions.

krishnasrinivas:
@balamurugana have you tested on windows? if not can you some basic tests on windows to see if different styles start minio properly?

when command prompt is on C:
`minio server d:\`    <---- this will fail with the current patch I think


When command prompt is on D:
`minio server \` <----- this will also fail I think


Relative path:
`minio server export`

Absolute path:
`minio server d:\export`

Absolute path:
`minio server \export`

krishnasrinivas:
@harshavardhana should we support this style on windows:

minio server d:\
minio server \

master branch supports. But not sure if it's a good idea ... because we dont support exposing root directory on nix.

We definitely have to disallow `minio server C:\`

harshavardhana:
> minio server d:
> minio server \

It is a requirement on windows when a CIFS/SMB share is mounted on a `Z:\` drive so we need to work. Otherwise we need to educate users to do that which might come as a surprise. 

Also there is no reason to avoid running as "root directory on nix." we can just warn and continue to run.

balamurugana:
> It is a requirement on windows when a CIFS/SMB share is mounted on a Z:\ drive so we need to work. Otherwise we need to educate users to do that which might come as a surprise.
> 
> Also there is no reason to avoid running as "root directory on nix." we can just warn and continue to run.

Currently it works for `D:` `D:\`, not to `\`.

cmd/endpoint.go
```diff
@@ -0,0 +1,382 @@
...
+				return Endpoint{}, fmt.Errorf("invalid URL endpoint format: %s", err)
+			}
+
+			host = u.Host
+		} else {
+			var p int
+			p, err = strconv.Atoi(port)
+			if err != nil {
+				return Endpoint{}, fmt.Errorf("invalid URL endpoint format: invalid port number")
+			} else if p < 1 || p > 65536 {
```
donatello: Max port value is 2^16-1 = 65535, so it should say `p > 65535` here.

cmd/endpoint.go
```diff
@@ -0,0 +1,382 @@
...
+		}
+	}
+
+	// No local endpoint found.
+	if localEndpointCount == 0 {
+		return serverAddr, endpoints, setupType, fmt.Errorf("no endpoint found for this host")
+	}
+
+	// Check whether same path is not used in endpoints of a host.
+	{
```
donatello, line 297: For example, this block is just a check and can be separated out to be a bit more readable.

cmd/endpoint.go
```diff
@@ -0,0 +1,382 @@
...
+			hostIPSet, _ := getHostIP4(host)
+			if IPSet, ok := pathIPMap[endpoint.Path]; ok {
+				if !IPSet.Intersection(hostIPSet).IsEmpty() {
+					err = fmt.Errorf("path '%s' can not be served from different address/port", endpoint.Path)
+					return serverAddr, endpoints, setupType, err
+				}
+			} else {
+				pathIPMap[endpoint.Path] = hostIPSet
+			}
+		}
```
donatello, line 314: In the above, the algorithm seems incorrect, and the error message is also misleading.

It is legitimate to serve a path like `/export` from different hosts. It is not legitimate to serve `/export` from multiple minio servers on the same host. The error message is conveying the opposite of this.

The algorithm is incorrect - e.g. endpoints:

```
10.0.0.1:9000/export
10.0.0.2:9000/export
10.0.0.2:9001/export
10.0.0.3:9000/export
```

After first iteration, the pathIPMap will have an entry for `/export` with stringset value containing just the IP `10.0.0.1`. In the second iteration, the IP seen is `10.0.0.2` and no error is raised. In the third iteration, the same IP 10.0.0.2 is seen - this should return an error, but this will also pass undetected because the intersection operation will give an empty set.

cmd/net.go
```diff
@@ -0,0 +1,167 @@
...
+			// As we are able to listen on this network, the port is not in use.
+			// Close the listener and continue check other networks.
+			if err = l.Close(); err != nil {
+				return err
+			}
+		} else if isAddrInUseErr(err) {
+			// As we got EADDRINUSE error, the port is in use by other process.
+			// Return the error.
+			return err
+		}
```
donatello, line 134: Here, is it ok to return success if `err != nil` and `!isAddrInUseErr(err)` is true?

cmd/net.go
```diff
@@ -0,0 +1,167 @@
...
+	host, port, err := net.SplitHostPort(serverAddr)
+	if err != nil {
+		return err
+	}
+
+	// Check whether port is a valid port number.
+	p, err := strconv.Atoi(port)
+	if err != nil {
+		return fmt.Errorf("invalid port number")
+	} else if p < 1 || p > 65536 {
```
donatello: `s/65536/65535` - as reasoned earlier in this review.

donatello (changes requested):

cmd/endpoint.go
balamurugana: Done

cmd/net.go
balamurugana: Done

cmd/endpoint.go
balamurugana: > In the above, the algorithm seems incorrect, and the error message is also misleading.

What is the right error message you are expecting?



> It is legitimate to serve a path like /export from different hosts. It is not legitimate to serve /export from multiple minio servers on the same host. The error message is conveying the opposite of this.
> 
> The algorithm is incorrect - e.g. endpoints:
> 
> 10.0.0.1:9000/export
> 10.0.0.2:9000/export
> 10.0.0.2:9001/export
> 10.0.0.3:9000/export
> 
> After first iteration, the pathIPMap will have an entry for /export with stringset value containing just the IP 10.0.0.1. In the second iteration, the IP seen is 10.0.0.2 and no error is raised. In the third iteration, the same IP 10.0.0.2 is seen - this should return an error, but this will also pass undetected because the intersection operation will give an empty set.


It seems you lost something there.

Below test case passes
```go
	case6u1, _ := url.Parse("http://10.0.0.1:9000/export")
	case6u2, _ := url.Parse("http://10.0.0.2:9000/export")
	case6u3, _ := url.Parse("http://10.0.0.3:9000/export")
	case6u4, _ := url.Parse("http://localhost:9001/export")

		{":9001", []string{"http://10.0.0.1:9000/export", "http://10.0.0.2:9000/export", "http://localhost:9001/export", "http://10.0.0.3:9000/export"}, ":9001", EndpointList{
			Endpoint{URL: case6u1, IsLocal: false},
			Endpoint{URL: case6u2, IsLocal: false},
			Endpoint{URL: case6u3, IsLocal: false},
			Endpoint{URL: case6u4, IsLocal: true},
		}, DistXLSetupType, nil},
```

cmd/net.go
balamurugana: This function returns `nil` only if all `[]string{"tcp", "tcp4", "tcp6"}` have free port.

harshavardhana:
Weird unrelated test case failure

```
Running cyclo: Done.
Running all coverage for minio: ok  	github.com/minio/minio	0.010s	coverage: 66.7% of statements
?   	github.com/minio/minio/browser	[no test files]
OK: 201 passed
--- FAIL: TestListObjectPartsDiskNotFound (0.19s)
	object-api-multipart_test.go:1460: Test 5: XL: Expected to fail with error "Bucket not found: volatile-bucket-1", but instead failed with error "disk not found" instead
	object-api-multipart_test.go:1460: Test 6: XL: Expected to fail with error "Bucket not found: volatile-bucket-2", but instead failed with error "disk not found" instead
FAIL
coverage: 75.2% of statements
FAIL	github.com/minio/minio/cmd	81.275s
make: *** [coverage] Error 1
```

balamurugana:
> Weird unrelated test case failure

Not sure why. It works fine if I run locally

harshavardhana:
> Not sure why. It works fine if I run locally

Opening a new bug, will rebuild this PR.

harshavardhana (approved):

harshavardhana:
Need one more approval here @krishnasrinivas @vadmeste @donatello @krisis - can i get one? are there any more changes?

vadmeste (approved):

krishnasrinivas (approved):

cmd/endpoint.go
donatello: Please try the test case I provided rather than the above. The bug is in the test case that I provided.

donatello:
This patch is merged with a bug.

cmd/endpoint.go
balamurugana: @donatello I got the problem and fixed in https://github.com/minio/minio/pull/4105

balamurugana:
@donatello Check the fix in https://github.com/minio/minio/pull/4105

Issue #3939, opened 2017-03-20
build: Reduce release binary sizes for minio releases


harshavardhana:
## Expected Behavior
Binary sizes can be made as small as 5.6MB from 24MB current release using upx.

## Current Behavior
Currently no binary size reduction is attempted. 

## Possible Solution
Use - https://upx.github.io/ and also along with `go build --ldflags "-s -w"`

## Steps to Reproduce (for bugs)
Just `make` and check the resulting binary size.

## Context
This is for faster to download and deploy times overall user experience etc.

## Your Environment
Current master.

harshavardhana:
// cc @hackintoshrao once this change is done - need your help in validating the sanctity of the binary so that we don't expect any unexpected issues.

matrixik:
UPX on Go binaries is not working now: https://github.com/golang/go/issues/16706

harshavardhana:
Looks like something happened between 1.7 and 1.8 it's regression. It works with our current release btw which is built with 1.7.5

PR #4027, opened 2017-04-01
build: Reduce binary size by using -s -w

harshavardhana:
## Description
Reduce binary size for all minio source builds through `make`

## Motivation and Context
Refer #3939

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 LDFLAGS := $(shell go run buildscripts/gen-ldflags.go)
 PWD := $(shell pwd)
 GOPATH := $(shell go env GOPATH)
-BUILD_LDFLAGS := '$(LDFLAGS)'
+BUILD_LDFLAGS := '$(LDFLAGS) -s -w'
 TAG := latest
 
 HOST ?= $(shell uname)


harshavardhana:
Upx needs to be done by the minio-release tool.

donatello (approved):
Looks good - I see the binary decreasing in size from 27MB to 17MB with this change.

I checked if log statements printing their source line is affected by removing either symbol table/DWARF symbols - it does not affect them.

So LGTM.

harshavardhana:
> Looks good - I see the binary decreasing in size from 27MB to 17MB with this change.
> 
> I checked if log statements printing their source line is affected by removing either symbol table/DWARF symbols - it does not affect them.

Yes next step is upx 

```
$ upx --brute minio
                       Ultimate Packer for eXecutables
                          Copyright (C) 1996 - 2013
UPX 3.91        Markus Oberhumer, Laszlo Molnar & John Reiser   Sep 30th 2013

        File size         Ratio      Format      Name
   --------------------   ------   -----------   -----------
  17191936 ->   4303540   25.03%  linux/ElfAMD   minio                         

Packed 1 file.
$ du -sh minio
4.2M	minio
```

vadmeste:
Shouldn't we check performance penalty after reducing size ?

harshavardhana:
> Shouldn't we check performance penalty after reducing size ?

What performance penalty?

vadmeste:
> Shouldn't we check performance penalty after reducing size ?
> What performance penalty?

Reducing size has usually impact on speed performance, but I see here that we are passing -s parameter to the linker and not compiler, so this is OK.

PR #4040, opened 2017-04-03
handler: simplify parsing valid location constraint.

harshavardhana:
## Description
Separate out validating v/s parsing logic in
isValidLocationConstraint() into parseLocationConstraint()
and isValidLocation()

## Motivation and Context
Simplify.

## How Has This Been Tested?
Manually.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/handler-utils.go
+++ b/cmd/handler-utils.go
@@ -24,36 +24,29 @@ import (
 	"strings"
 )
 
-// Validates location constraint in PutBucket request body.
-// The location value in the request body should match the
-// region configured at serverConfig, otherwise error is returned.
-func isValidLocationConstraint(r *http.Request) (s3Error APIErrorCode) {
-	serverRegion := serverConfig.GetRegion()
+// Parses location constraint from the incoming reader.
+func parseLocationConstraint(r *http.Request) (location string, s3Error APIErrorCode) {
 	// If the request has no body with content-length set to 0,
 	// we do not have to validate location constraint. Bucket will
 	// be created at default region.
 	locationConstraint := createBucketLocationConfiguration{}
 	err := xmlDecoder(r.Body, &locationConstraint, r.ContentLength)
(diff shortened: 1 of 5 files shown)


cmd/handler-utils.go
```diff
@@ -24,36 +24,29 @@ import (
...
-	errorIf(err, "Unable to xml decode location constraint")
-	// Treat all other failures as XML parsing errors.
-	return ErrMalformedXML
+	return location, ErrNone
+}
+
+// Validates input location is same as configured region
+// of Minio server.
+func isValidLocation(location string) bool {
+	return serverConfig.GetRegion() != location
```
vadmeste: This should be equal and not unequal to follow the function name.

harshavardhana: oops yes..

cmd/handler-utils.go
```diff
@@ -24,36 +24,29 @@ import (
...
-			incomingRegion = globalMinioDefaultRegion
-		}
-		// Return errInvalidRegion if location constraint does not match
-		// with configured region.
-		s3Error = ErrNone
-		if serverRegion != incomingRegion {
-			s3Error = ErrInvalidRegion
-		}
-		return s3Error
+	if err != nil && err != io.EOF {
```
vadmeste, line 34: I think `err != io.EOF` is not useful here, I know it is not harmful but it makes reading little more difficult.

vadmeste: LGTM otherwise.

harshavardhana: You can't do it otherwise io.EOF is a valid condition. @vadmeste when no location is sent.

vadmeste: > You can't do it otherwise io.EOF is a valid condition. @vadmeste when no location is sent.

:+1:, better to add a comment here

vadmeste (approved):

PR #4041, opened 2017-04-03
cache: Increasing caching GC percent from 20 to 50.

harshavardhana:
## Description
Previous value was set to avoid large cache value build
up but we can clearly see this can cause lots of GC
pauses which can lead to significant drop in performance.

Change this value to 50% and decrease the value to 25%
once the 75% cache size is used. To have a larger
window for GC pauses.

Another change is to only allow caching if a server has
more than 24GB of RAM instead of 8GB.

## Motivation and Context
Simplify and fixes an apparent bug. 

## How Has This Been Tested?
Manually using `dicomimport` tool. 

## Types of changes

## Checklist:


fwessels (approved):
Looks good to me

krishnasrinivas (approved):

hackintoshrao (approved):

krishnasrinivas (approved):

PR #4045, opened 2017-04-04
Return possible states a heal operation

krisis:
## Background
Healing algorithm returns failure only if an attempt to heal any of the disks failed. It returns success even if all disks needing heal were offline.
In its current form, heal admin REST APIs returns the above boolean state making it impossible for clients like `mc` to distinguish between an object being healed completely, partially and not.

## Solution
The fix (so far) is to return `struct { HealedCount int, OfflineCount int}` in addition to `error` for a heal-object and heal-upload operation. `mc` can make use of this to guide users better.

## Future-proof solution
This brings us to whether it is right by design to provide internal details of the healing process in admin API. We return an enumerated constant in heal-object and heal-upload API that captures the 3 result states of the heal operation, namely, `CompletelyHealed`, `PartiallyHealed` and `NotHealed`. In addition, we would return `error` to indicate if the healing operation failed to due to other I/O related errors.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/admin-handlers.go
+++ b/cmd/admin-handlers.go
@@ -630,9 +630,40 @@ func isDryRun(qval url.Values) bool {
 	return false
 }
 
-type healObjectResult struct {
-	HealedCount  int
-	OfflineCount int
+// healResult - represents result of a heal operation like
+// heal-object, heal-upload.
+type healResult struct {
+	State healState `json:"state"`
+}
+
+// healState - different states of heal operation
+type healState int
+
+const (
+	// healNone - none of the disks healed
+	healNone healState = iota
+	// healPartial - some disks were healed, others were offline
+	healPartial
+	// healOK - all disks were healed
+	healOK
+)
+
+// newHealResult - returns healResult given number of disks healed and
+// number of disks offline
+func newHealResult(numHealedDisks, numOfflineDisks int) healResult {
+	var state healState
+	switch {
+	case numHealedDisks == 0:
+		state = healNone
+
+	case numOfflineDisks > 0:
+		state = healPartial
+
+	default:
+		state = healOK
+	}
+
+	return healResult{State: state}
 }
 
 // HealObjectHandler - POST /?heal&bucket=mybucket&object=myobject&dry-run
@@ -683,10 +714,7 @@ func (adminAPI adminAPIHandlers) HealObjectHandler(w http.ResponseWriter, r *htt
 		return
 	}
 
-	jsonBytes, err := json.Marshal(healObjectResult{
-		HealedCount:  numHealedDisks,
-		OfflineCount: numOfflineDisks,
-	})
+	jsonBytes, err := json.Marshal(newHealResult(numHealedDisks, numOfflineDisks))
 	if err != nil {
 		writeErrorResponse(w, toAPIErrorCode(err), r.URL)
 		return
@@ -761,10 +789,7 @@ func (adminAPI adminAPIHandlers) HealUploadHandler(w http.ResponseWriter, r *htt
 		return
 	}
 
-	jsonBytes, err := json.Marshal(healObjectResult{
-		HealedCount:  numHealedDisks,
-		OfflineCount: numOfflineDisks,
-	})
+	jsonBytes, err := json.Marshal(newHealResult(numHealedDisks, numOfflineDisks))
 	if err != nil {
 		writeErrorResponse(w, toAPIErrorCode(err), r.URL)
 		return
--- a/pkg/madmin/API.md
+++ b/pkg/madmin/API.md
@@ -245,6 +245,18 @@ __Example__
 ### HealObject(bucket, object string, isDryRun bool) (HealResult, error)
 If object is successfully healed returns nil, otherwise returns error indicating the reason for failure. If isDryRun is true, then the object is not healed, but heal object request is validated by the server. e.g, if the object exists, if object name is valid etc.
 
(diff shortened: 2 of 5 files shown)


deekoder:
Can you update the API doc https://github.com/minio/minio/tree/master/docs/admin-api with the details?

deekoder (changes requested):
Change looks ok. But please update the admin-api docs as mentioned in the comments.

krisis:
@deekoder I have added necessary admin-API docs.

cmd/admin-handlers.go
```diff
@@ -630,9 +630,40 @@ func isDryRun(qval url.Values) bool {
...
+	// notHealed - none of the disks healed
+	notHealed healState = iota
+	// partiallyHealed - some disks were healed, others were offline
+	partiallyHealed
+	// completelyHealed - all disks were healed
+	completelyHealed
+)
+
+// newHealResult - returns healResult given number of disks healed and
+// number of disks offline
```
harshavardhana: Can you add a test case for this alone?

harshavardhana (changes requested):

cmd/admin-handlers.go
krisis: @harshavardhana Added a unit test.

harshavardhana (approved):

harshavardhana:
@deekoder @krisis has fixed the admin API docs can you review again?

harshavardhana:
ping @vadmeste need your review here..

pkg/madmin/heal-commands.go
```diff
@@ -494,10 +494,21 @@ func (adm *AdminClient) HealUpload(bucket, object, uploadID string, dryrun bool)
 
 // HealResult - represents result of heal-object admin API.
 type HealResult struct {
-	HealedCount  int // number of disks that were healed.
-	OfflineCount int // number of disks that needed healing but were offline.
+	State HealState
```
vadmeste: Can you set a json tag here? `json:"state"`

vadmeste (changes requested):
One comment, LGTM otherwise

deekoder:
I will schedule the gist review today with @abperiasamy

harshavardhana:
> I will schedule the gist review today with @abperiasamy

Is this PR waiting on the gist review? - we can merge this and proceed with gist review with a separate PR as a doc cleanup.

vadmeste (approved):

cmd/admin-handlers.go
```diff
@@ -630,9 +630,40 @@ func isDryRun(qval url.Values) bool {
...
+type healResult struct {
+	State healState `json:"state"`
+}
+
+// healState - different states of heal operation
+type healState int
+
+const (
+	// notHealed - none of the disks healed
+	notHealed healState = iota
```
deekoder: @abperiasamy :  By prefixing heal, there is namespace pollution. Use healOK, healFailed, healPartial everywhere. Always group them with common prefix to help IDEs autocomplete.

pkg/madmin/heal-commands.go
```diff
@@ -494,10 +494,21 @@ func (adm *AdminClient) HealUpload(bucket, object, uploadID string, dryrun bool)
 
 // HealResult - represents result of heal-object admin API.
```
deekoder, line 495: @abperiasamy :  healState is duplicated in cmd/admin-handlers.go

harshavardhana: This is on purpose the struct on cmd/ cannot be used with `madmin` package.. Just like how `minio-go` has ListObjects struct which is similar to what server has as ListObjectInfo. All elements are same. 

Since `madmin` is imported into `mc` it is not required for minio server to be vendorized in `mc` to get admin API working. So it is written like an independent SDK capable of parsing all the Admin API REST responses and requests.

harshavardhana:
> @abperiasamy : healState is duplicated in cmd/admin-handlers.go

This is on purpose the struct on cmd/ cannot be used with madmin package.. Just like how minio-go has ListObjects struct which is similar to what server has as ListObjects.

Since madmin is imported into mc it is not required for minio server to be vendorized in mc to get admin API working. So it is written like an independent SDK capable of parsing all the Admin API REST responses and requests.

pkg/madmin/API.md
```diff
@@ -245,6 +245,18 @@ __Example__
...
 If object is successfully healed returns nil, otherwise returns error indicating the reason for failure. If isDryRun is true, then the object is not healed, but heal object request is validated by the server. e.g, if the object exists, if object name is valid etc.
 
+| Param  | Type  | Description  |
+|---|---|---|
+|`h.State` | _HealState_ | Represents the result of heal operation. It could be one of `NotHealed`, `PartiallyHealed` or `CompletelyHealed`. |
+
+
+| Value | Description |
+|---|---|
+|`NotHealed` | Object/Upload wasn't healed on any of the disks |
```
harshavardhana: This should be fixed as well @krisis ?

harshavardhana (changes requested):

krisis:
@deekoder I have changed the `healState` values to be `HealNone`, `HealPartial` and `HealOK` respectively.

pkg/madmin/API.md
krisis: @harshavardhana yes, thanks for catching that. I have updated the PR fixing this.

harshavardhana (approved):

deekoder (approved):

PR #4049, opened 2017-04-05
admin/info: Add HTTPStats value as part of serverInfo() struct.

harshavardhana:
## Description
Remove our counter implementation instead use atomic external
package which supports more types and methods.

## Motivation and Context
Add HTTPStats value as part of serverInfo() struct.

## How Has This Been Tested?
Manually.. 

## Types of changes

## Checklist:


cmd/generic-handlers.go
```diff
@@ -401,9 +401,18 @@ func (h httpStatsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 
+	// Time after call has completed.
+	tAfter := UTCNow()
+
+	// Time duration since the call started.
+	duration := tAfter.Sub(tBefore)
+
 	// Update http statistics
-	globalHTTPStats.updateStats(r, ww)
+	globalHTTPStats.updateStats(r, ww, duration.Seconds())
```
vadmeste: Can you add a comment saying that `duration.Seconds()` returns a floating and non rounded seconds ?

cmd/http-stats.go
```diff
@@ -0,0 +1,184 @@
...
+	totalDELETEs   HTTPMethodStats
+	successDELETEs HTTPMethodStats
+}
+
+// Converts http stats into struct to be sent back to the client.
+func (st HTTPStats) toServerHTTPStats() ServerHTTPStats {
+	serverStats := ServerHTTPStats{}
+	serverStats.TotalHEADStats = ServerHTTPMethodStats{
+		Count:       st.totalHEADs.Counter.Load(),
+		AvgDuration: fmt.Sprint(time.Duration(st.totalHEADs.Duration.Load()/float64(st.totalHEADs.Counter.Load())) * time.Second),
```
vadmeste: Can you avoid duplicate fmt.Sprint()? It would be easier to review.

vadmeste (approved):
LGTM, only minor comments.

cmd/generic-handlers.go
harshavardhana: Done

cmd/http-stats.go
harshavardhana: Done

vadmeste (approved):

harshavardhana:
ping @krisis - can you review.

harshavardhana:
Since @krisis is not around @donatello can you review.

krisis (approved):

cmd/generic-handlers.go
```diff
@@ -401,9 +401,21 @@ func (h httpStatsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 	h.handler.ServeHTTP(ww, r)
 
+	// Time after call has completed.
+	tAfter := UTCNow()
+
+	// Time duration in secs since the call started.
+	//
+	// We don't need to do nanosecond precision in this
+	// simply for the fact that it is not human readable.
+	durationSecs := tAfter.Sub(tBefore).Seconds()
```
mtparet: Does it means that duration below 1s are ignored ?

PR #4060, opened 2017-04-07
notify: Webhook endpoints can fail, but we must start the server.

### LINKED ISSUE #4050: Startup fails if webhook endpoint is offline.
In my development environment my webhook endpoint was offline.  During this time startup of a Minio node would fail.

## Expected Behavior
I'm not entirely sure if this is operating as intended or a bug, but it seems to me having Minio up and running is preferable to stopping due to a down webhook endpoint.

Minio should log the error and continue with initialization.

## Current Behavior
Minio logs the following error and stops:

` time="2017-04-03T11:14:01Z" level=fatal msg="Unable to initialize event notification." cause="dial tcp 192.168.1.5:443: getsockopt: no route to host" source="[xl-v1.go:91:newXLObjectLayer()]"`

## Possible Solution
Maybe make this a non-fatal error by adding some handling somewhere around here:
https://github.com/minio/minio/blob/master/cmd/xl-v1.go#L87

1. Add an invalid webhook endpoint to the server config.
2. Restart minio.

## Your Environment
* Version used: 2017-03-16T21:50:32Z

harshavardhana:
## Description
Ignore any network errors when registering a webhook
notifier during Minio startup sequence. This way server
can be started even if the webhook endpoint is not available
and unreachable.

## Motivation and Context
Fixes #4050

## How Has This Been Tested?
Manually with wrong config.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/notify-webhook.go
+++ b/cmd/notify-webhook.go
@@ -17,12 +17,14 @@
 package cmd
 
 import (
+	"bytes"
 	"crypto/tls"
 	"fmt"
 	"io/ioutil"
 	"net"
 	"net/http"
 	"net/url"
+	"strings"
 	"time"
 
 	"github.com/Sirupsen/logrus"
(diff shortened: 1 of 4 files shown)


cmd/notify-webhook.go
```diff
@@ -48,33 +50,119 @@ type httpConn struct {
...
+			return derr
+		}
+		if resp == nil {
+			return fmt.Errorf("No response from server to download URL %s", urlStr)
+		}
+		resp.Body.Close()
+
+		// Redo the request with the new redirect url if http 307
+		// is returned, quit the loop otherwise
+		if resp.StatusCode == http.StatusTemporaryRedirect {
```
vadmeste, line 141: Can you add a TODO here to check later if golang 1.8 follows 307 after POST ? 
https://github.com/golang/go/issues/7912

cmd/notify-webhook.go
```diff
@@ -48,33 +50,119 @@ type httpConn struct {
...
+		},
+	}
+
+	// Set content-type.
+	req.Header.Set("Content-Type", "application/json")
+
+	// Set proper server user-agent.
+	req.Header.Set("User-Agent", globalServerUserAgent)
+
+	for {
```
vadmeste: Looks like 'for' here is only needed to follow redirect urls, can you also add a comment here ?

cmd/notify-webhook.go
```diff
@@ -48,33 +50,119 @@ type httpConn struct {
...
+var successStatus = []int{
+	http.StatusOK,
+	http.StatusAccepted,
+	http.StatusContinue,
+	http.StatusNoContent,
+	http.StatusPartialContent,
+}
+
+// isNetErrorRetryable - is network error retryable.
+func isNetErrorRetryable(err error) bool {
```
vadmeste: This PR is supposed to ignore error like 'connection refused' for example when remote port is closed, can you add more comments to understand ?

cmd/notify-webhook.go
harshavardhana: There is nothing do for since its already implemented in go1.8?

cmd/notify-webhook.go
harshavardhana: Connection refused is already `*net.OpError` ..

cmd/notify-webhook.go
vadmeste: > There is nothing do for since its already implemented in go1.8?

Yes

vadmeste (approved):

donatello (approved):

Issue #3185, opened 2016-11-05
erasure: Without read or write quorum object layer hangs with no errors to client.


krishnasrinivas:
## Expected Behavior
`mc cat` should return error when quorum is not available.
`mc cp` should return error when quorum is not available.

## Current Behavior
`mc cat` hangs.
`mc cp` hangs 

## Steps to Reproduce (for bugs)

``mc cat`` hangs 
1. Create a 3:3 distributed-XL setup
2. mc cp /etc/passwd myminio/testbucket
3. bringdown 4 servers
4. `mc cat myminio/testbucket/passwd` hangs

``mc cp`` hangs
1. Create a 3:3 distributed-XL setup
2. bring down 3 servers
3. mc cp /etc/passwd myminio/testbucket hangs

harshavardhana:
> I think Get http://localhost:9001/test/?location= request hangs when trying to get lock.

I think it is expected in distributed Lock right now, may be we can have a canceler timeout once it has attempted enough.  // cc @fwessels

harshavardhana:
Actually Lock cannot timeout, if it does it would proceed forward. There needs to be a way to actively cancel the ObjectLayer operation and return error when the quorum is not available.  We don't even attempt to lock with this functionality - note it would need to make additional calls to make sure if we have enough quorum even to begin with.

harshavardhana:
Merged #3184 with this bug and changed the overall description.

abperiasamy:
With N/2 redundancy, it is a likely condition only in a desperate situation. Ideally the handlers should time out at the higher level and return appropriate response. Locks with deadlines are dangerous. `context` package is meant for this purpose, however context adds more complexity to minio for a tiny little benefit. It is not all that hard to implement deadline for top level handler go routines alone. This is not a blocker for this release.

harshavardhana:
> This is not a blocker for this release.

Moving it out.

fwessels:
If you would want to cancel/error out, to Go way of doing it is with an additional go routine that has a timeout and in that way abort the blocking operation.

We can decide on a case-by-case basis when we want to do this. Or come up with some sort of 'template' that adds this functionality.

deekoder:
Discuss and finalize where we stand on these class of bugs.

deekoder:
Duplicate #3234

donatello:
[...]

PR #4066, opened 2017-04-07
Allow cluster to start when only n/2 servers are up

### LINKED ISSUE #3234: erasure: Replace n/2 disks, new servers do not come up

donatello:
## Description

Fixes #3234.

Relaxes the quorum requirement to start the object layer, and skips
quick-healing at start-up (as no write quorum is present).

This fix allows reads to work after start-up, but it does hang on writes. The hang is because no write quorum servers are available and the lock interface currently does not have a possibility of failing/returning an error. In my view, this is a separate issue - #3185 ought to be re-opened.

## Motivation and Context

Toward resolving #3234. 

## How Has This Been Tested?

Manually.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/prepare-storage.go
+++ b/cmd/prepare-storage.go
@@ -132,13 +132,14 @@ func prepForInitXL(firstDisk bool, sErrs []error, diskCount int) InitActions {
 	}
 
 	quorum := diskCount/2 + 1
+	readQuorum := diskCount / 2
 	disksOffline := errMap[errDiskNotFound]
 	disksFormatted := errMap[nil]
 	disksUnformatted := errMap[errUnformattedDisk]
 	disksCorrupted := errMap[errCorruptedFormat]
 
 	// No Quorum lots of offline disks, wait for quorum.
-	if disksOffline >= quorum {
+	if disksOffline > readQuorum {
 		return WaitForQuorum
 	}
 
@@ -168,7 +169,7 @@ func prepForInitXL(firstDisk bool, sErrs []error, diskCount int) InitActions {
 	}
 
 	// Already formatted and in quorum, proceed to initialization of object layer.
-	if disksFormatted >= quorum {
+	if disksFormatted >= readQuorum {
 		if disksFormatted+disksOffline == diskCount {
 			return InitObjectLayer
 		}
@@ -293,7 +294,7 @@ func initStorageDisks(endpoints []*url.URL) ([]StorageAPI, error) {
 	return storageDisks, nil
 }
 
-// Format disks before initialization object layer.
+// Format disks before initialization of object layer.
 func waitForFormatXLDisks(firstDisk bool, endpoints []*url.URL, storageDisks []StorageAPI) (formattedDisks []StorageAPI, err error) {
 	if len(endpoints) == 0 {
 		return nil, errInvalidArgument
--- a/cmd/xl-v1.go
+++ b/cmd/xl-v1.go
@@ -155,6 +155,16 @@ func newXLObjects(storageDisks []StorageAPI) (ObjectLayer, error) {
 	xl.readQuorum = readQuorum
 	xl.writeQuorum = writeQuorum
 
+	// If the number of offline servers is equal to the readQuorum
+	// (i.e. the number of online servers also equals the
+	// readQuorum), we cannot perform quick-heal (no
+	// write-quorum). However reads may still be possible, so we
+	// skip quick-heal in this case, and continue.
+	offlineCount := len(newStorageDisks) - diskCount(newStorageDisks)
+	if offlineCount == readQuorum {
+		return xl, nil
+	}
+
 	// Do a quick heal on the buckets themselves for any discrepancies.
 	return xl, quickHeal(xl.storageDisks, xl.writeQuorum, xl.readQuorum)
 }


krisis (approved):

cmd/prepare-storage.go
```diff
@@ -150,29 +151,35 @@ func prepForInitXL(firstDisk bool, sErrs []error, diskCount int) InitActions {
...
+		// Only the first server formats an uninitialized
+		// setup, others wait for notification.
+		if firstDisk {
 			return FormatDisks
 		}
 		return WaitForFormatting
 	}
 
-	// Total disks unformatted are in quorum verify if we have some offline disks.
+	// Total disks unformatted are in quorum verify if we have
```
harshavardhana: These are not related changes to PR, can we avoid it?

cmd/prepare-storage.go
```diff
@@ -150,29 +151,35 @@ func prepForInitXL(firstDisk bool, sErrs []error, diskCount int) InitActions {
...
-	if disksFormatted >= quorum {
+	// Already formatted and having read-quorum, proceed to
+	// initialization of object layer.
+	if disksFormatted >= readQuorum {
 		if disksFormatted+disksOffline == diskCount {
 			return InitObjectLayer
 		}
-		// Some of the formatted disks are possibly corrupted or unformatted, heal them.
+		// Some of the formatted disks are possibly corrupted
+		// or unformatted, heal them.
```
harshavardhana: If we still return `WaitForHeal` should the caller proceed to heal? under ReadQuorum?

cmd/xl-v1.go
```diff
@@ -155,6 +155,21 @@ func newXLObjects(storageDisks []StorageAPI) (ObjectLayer, error) {
 	xl.readQuorum = readQuorum
 	xl.writeQuorum = writeQuorum
 
+	// If the number of offline servers is equal to the readQuorum
+	// (i.e. the number of online servers also equals the
+	// readQuorum), we cannot perform quick-heal (no
+	// write-quorum). However reads may still be possible, so we
+	// skip quick-heal in this case, and continue.
+	offlineCount := 0
```
harshavardhana: perhaps you can use `len(newStorageDisks) - diskCount(newStorageDisks)`  or create a function similar to `diskCount` and rename it as `onlineDiskCount` and this one has `offlineDiskCount`

harshavardhana (changes requested):

cmd/prepare-storage.go
donatello: Done.

cmd/xl-v1.go
donatello: Done.

cmd/prepare-storage.go
donatello: From the current implementation in `healFormatXL`, this seems ok as no quorum appears to be required in it - I am not sure if that is entirely correct, but in this case, there seems to be no harm in allowing to heal format.json on a unformatted/corrupted disk if there are readQuorum other formatted disks. E.g.: in a fully working 8-node 8-disk cluster, if 4-disks are replaced, there is no data-loss (due to erasure-coding) and healing should be allowed.

donatello:
@harshavardhana Your comments are addressed - please check now.

harshavardhana (approved):

harshavardhana:
Ping @krishnasrinivas  can you review this once ?

cmd/prepare-storage.go
```diff
@@ -132,13 +132,14 @@ func prepForInitXL(firstDisk bool, sErrs []error, diskCount int) InitActions {
...
 	quorum := diskCount/2 + 1
+	readQuorum := diskCount / 2
 	disksOffline := errMap[errDiskNotFound]
 	disksFormatted := errMap[nil]
 	disksUnformatted := errMap[errUnformattedDisk]
 	disksCorrupted := errMap[errCorruptedFormat]
 
 	// No Quorum lots of offline disks, wait for quorum.
-	if disksOffline >= quorum {
+	if disksOffline > readQuorum {
```
krishnasrinivas, line 142: Not related to this commit. But when deciding on quorum we have to consider online-disks rather than offline-disks.

Because when data != parity this comparison will be incorrect.

For ex. consider 6-data and 2-parity. readQuorum will be 6. If 3 disks are down, the statement becomes:
```
if 3 > 6 {
    return WaitForQuorum
}
```
Which is incorrect because we should have returned WaitForQuorum.

If we talk in terms of online-disks, it will be like this (which is also more readable):
```
if disksOnline < readQuorum {
    return WaitForQuorum
}
```
which will work as intended.

If/when we make data/parity configurable we will have to fix these things.

krishnasrinivas (approved):

cmd/prepare-storage.go
harshavardhana: Can this be captured in a bug @krishnasrinivas ?

krishnasrinivas: @harshavardhana opened an umbrella issue https://github.com/minio/minio/issues/4079

PR #4086, opened 2017-04-10
gateway: Support for custom endpoint.

### LINKED ISSUE #4083: gateway: Support for custom endpoint
Support endpoint in the format:

minio gateway azure [endpoint]
default would connect to Azure backend.

minio gateway s3 [endpoint]
default would connect to AWS-S3.

krishnasrinivas:
fixes #4083

### DIFF
--- a/cmd/gateway-main.go
+++ b/cmd/gateway-main.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"net/url"
 	"os"
+	"regexp"
 
 	"github.com/gorilla/mux"
 	"github.com/minio/cli"
@@ -118,7 +119,16 @@ func newGatewayConfig(accessKey, secretKey, region string) error {
 	return nil
 }
 
-func getGatewayEndpoint(arg string) (endPoint string, secure bool, err error) {
+// Return endpoint.
+func parseGatewayEndpoint(arg string) (endPoint string, secure bool, err error) {
+	schemeSpecified, err := regexp.MatchString(".*://", endPoint)
+	if err != nil {
+		return "", false, err
+	}
+	if !schemeSpecified {
+		// Default connection will be "secure".
+		endPoint = "https://" + endPoint
+	}
 	u, err := url.Parse(arg)
 	if err != nil {
 		return "", false, err
@@ -129,9 +139,6 @@ func getGatewayEndpoint(arg string) (endPoint string, secure bool, err error) {
 		return u.Host, true, nil
 	case "https":
 		return u.Host, false, nil
-	case "":
-		// url.Parse("play.minio.io") puts play.minio.io in u.Path
-		return u.Path, true, nil
 	default:
 		return "", false, fmt.Errorf("Unrecognized scheme %s", u.Scheme)
 	}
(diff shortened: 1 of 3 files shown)


cmd/gateway-main.go
```diff
@@ -142,10 +142,14 @@ func gatewayMain(ctx *cli.Context) {
 	// First argument is selected backend type.
 	backendType := ctx.Args().First()
 
+	// Second argument is endpoint.	If no endpoint is specified then the
+	// gateway implementation should use a default setting.
+	endPoint := ctx.Args().Get(1)
```
harshavardhana: is there no input validation? - I think we should.

harshavardhana (changes requested):

cmd/gateway-main.go
```diff
@@ -117,6 +118,25 @@ func newGatewayConfig(accessKey, secretKey, region string) error {
 	return nil
 }
```
harshavardhana, line 121: perhaps a better name is `parseGatewayEndpoint`  and also a comment as well.

cmd/gateway-main.go
```diff
@@ -117,6 +118,25 @@ func newGatewayConfig(accessKey, secretKey, region string) error {
...
+		return "", false, err
+	}
+
+	switch u.Scheme {
+	case "http":
+		return u.Host, true, nil
+	case "https":
+		return u.Host, false, nil
+	case "":
+		// url.Parse("play.minio.io") puts play.minio.io in u.Path
```
harshavardhana: In go1.8 this wouldn't happen anymore and an error will be generated, if we need to support a scenario like this. We should handle it like `net.SplitHostPort()` style.

krishnasrinivas: but it seems to work:
```
krishna@escape:~/tmp$ cat url.go
package main

import (
	"fmt"
	"net/url"
)

var asd = "asd"

func main() {
	u, err := url.Parse("play.minio.io")
	fmt.Println(err, u.Path)
}
krishna@escape:~/tmp$ go version
go version go1.8.1 linux/amd64
krishna@escape:~/tmp$ go run url.go
<nil> play.minio.io
krishna@escape:~/tmp$ 

```

harshavardhana: Doesn't work with IP @krishnasrinivas 

```
$ go run url.go
parse 192.168.1.106:9000: first path segment in URL cannot contain colon <nil>
```

cmd/gateway-main.go
```diff
@@ -117,6 +119,31 @@ func newGatewayConfig(accessKey, secretKey, region string) error {
 	return nil
 }
 
+// Return endpoint.
+func parseGatewayEndpoint(arg string) (endPoint string, secure bool, err error) {
```
harshavardhana, line 123: add a test case? @krishnasrinivas

cmd/gateway-main.go
```diff
@@ -117,6 +119,31 @@ func newGatewayConfig(accessKey, secretKey, region string) error {
...
+	schemeSpecified, err := regexp.MatchString(".*://", endPoint)
```
donatello: `arg` instead of `endPoint`? `endPoint` is part of return and at this point it is an empty string.

donatello: Also, for a simple check like this, it is not necessary to use regular expressions:

```
schemeSpecified := strings.Contains(arg, "://")
```

This is faster, and there is no error checking needed.

donatello (changes requested):

harshavardhana:
can you add a test case @krishnasrinivas ?

harshavardhana (approved):

harshavardhana:
@krishnasrinivas error in tests 

```
Installing ineffassign: Done.
Running vet: cmd/gateway-main_test.go:45: wrong number of args for format in Errorf call: 7 needed but 8 args
make: *** [vet] Error 1
```

harshavardhana (approved):

PR #4087, opened 2017-04-11
conditions: datetime headers supports new TZs

### LINKED ISSUE #4082: server: Precondition modified headers should be ignored if not in http.TimeFormat
Continuing from - https://github.com/minio/minio/issues/4080

## Expected Behavior
Should ignore ModifiedSince and UnmodifiedSince headers if they are not in http.TimeFormat and proceed to perform the action instead of returning an error to the client. 

## Current Behavior

```go
// returns true if object was modified after givenTime.
func ifModifiedSince(objTime time.Time, givenTimeStr string) bool {
        givenTime, err := time.Parse(http.TimeFormat, givenTimeStr)
        if err != nil {
                return true
        }
        // The Date-Modified header truncates sub-second precision, so
        // use mtime < t+1s instead of mtime <= t to check for unmodified.
        if objTime.After(givenTime.Add(1 * time.Second)) {
                return true
        }
        return false
}
```

Returns `true` which fails conditions when a client sends headers in invalid `http.TimeFormat`

## Possible Solution
```
        givenTime, err := time.Parse(http.TimeFormat, givenTimeStr)
        if err != nil {
                return true
        }
```

Should return `false` instead of `true`.

Clone the https://github.com/minio/minio-java master branch. Below minio-java functional test fails in play.minio.io whereas it succeeds in amazon s3.

```java
    CopyConditions copyConditions = new CopyConditions();
    DateTime dateRepresentation = new DateTime(2015, Calendar.MAY, 3, 10, 10);
    copyConditions.setUnmodified(dateRepresentation);
    client.copyObject(bucketName, filename, destBucketName, copyConditions);
```

Failure message:
```
ErrorResponse(code=PreconditionFailed, message=At least one of the pre-conditions you specified did not hold, bucketName=null, objectName=null, resource=/minio-java-test-2ffm6th/minio-java-test-g9dvi7, requestId=3L137, hostId=3L137)
request={method=PUT, url=https://play.minio.io:9000/minio-java-test-2ffm6th/minio-java-test-g9dvi7, headers=x-amz-copy-source-if-unmodified-since: Fri, 03 Apr 2015 17:10:00 +00:00
x-amz-copy-source: minio-java-test-3n2cg3a%2Fminio-java-test-g9dvi7
Content-MD5: 1B2M2Y8AsgTpgAmY7PhCfg==
Host: play.minio.io:9000
User-Agent: Minio (amd64; amd64) minio-java/dev
x-amz-content-sha256: UNSIGNED-PAYLOAD
x-amz-date: 20170409T131324Z
Authorization: AWS4-HMAC-SHA256 Credential=Q3AM3UQ867SPQQA43P2F/20170409/us-east-1/s3/aws4_request, SignedHeaders=content-md5;host;x-amz-content-sha256;x-amz-copy-source;x-amz-copy-source-if-unmodified-since;x-amz-date, Signature=4e5b1404d533fcafbe9468abdbb11b310184ddbcf1abb1c5fb281b061f3549e8
Content-Length: 0
}
response={code=412, headers=Accept-Ranges: bytes
Content-Type: application/xml
Etag: "f7f14b0dc4fc67f17f7590be179ad2de"
Last-Modified: Sun, 09 Apr 2017 13:13:24 GMT
Server: Minio/DEVELOPMENT.2017-04-06T07-23-25Z (linux; amd64)
Vary: Origin
X-Amz-Bucket-Region: us-east-1
X-Amz-Request-Id: 14B3BCE78F6F4807
Date: Sun, 09 Apr 2017 13:13:24 GMT
Transfer-Encoding: chunked
OkHttp-Sent-Millis: 1491743604320
OkHttp-Received-Millis: 1491743604327
```

## Context
To behave in accordance with HTTP specification 

https://tools.ietf.org/html/rfc7232#section-3.3
```
 A recipient MUST ignore the If-Modified-Since header field if the
 received field-value is not a valid HTTP-date.
```

https://tools.ietf.org/html/rfc7232#section-3.4
```
 A recipient MUST ignore the If-Unmodified-Since header field if the
 received field-value is not a valid HTTP-date.
```

## Your Environment
Current master branch.

vadmeste:
## Description
Some headers of type date time, like If-Unmodified-Since, If-Modified-Since
and other can support timezones other than GMT, like UTC, PST, CEST etc..
Though the S3 spec explicitly requires GMT, AWS S3 in practice accepts
time.RFC1123 date format, so this PR follows the same behavior.

## Motivation and Context
Fixes https://github.com/minio/minio/issues/4082

## Types of changes

## Checklist:

### DIFF
--- a/cmd/time.go
+++ b/cmd/time.go
@@ -1,0 +1,67 @@
+/*
+ * Minio Cloud Storage, (C) 2017 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cmd
+
+import "time"
+
+var minioTZ = map[string]int{
+
+	"PST": -8 * 3600,
+	"PDT": -7 * 3600,
+
+	"EST": -5 * 3600,
+	"EDT": -4 * 3600,
+
+	"GMT": 0,
+	"UTC": 0,
+
+	"CET":  1 * 3600,
+	"CEST": 2 * 3600,
+}
+
+// parseTime returns the correct time when a timezone is provided
+// with the time string, if the specified timezone is not found in
+// minioTZ, call only the standard golang time parsing.
+func parseTime(layout, value string) (time.Time, error) {
+	// Standard golang time parsing
+	t, err := time.Parse(layout, value)
+	if err != nil {
+		return time.Time{}, err
+	}
+
+	// Fetch the location of the passed time
+	loc := t.Location()
+
+	if loc == nil || loc == time.UTC || loc == time.Local {
+		// Nothing to do, even when location is set to time.Local
+		// since time will be obviously correct with the local
+		// machine's timezone
+		return t, nil
+	}
+
+	// Fetch the time offset associated to the passed timezone.
+	// If not found, return golang std time parser result.
+	offset, ok := minioTZ[loc.String()]
+	if !ok {
+		return t, nil
+	}
+
+	// Calculate the new time
+	newTime := t.Add(-time.Duration(offset) * time.Second)
+
+	return newTime, nil
+}
--- a/cmd/object-handlers-common.go
+++ b/cmd/object-handlers-common.go
@@ -66,11 +66,13 @@ func checkCopyObjectPreconditions(w http.ResponseWriter, r *http.Request, objInf
 	// since the specified time otherwise return 412 (precondition failed).
 	ifModifiedSinceHeader := r.Header.Get("x-amz-copy-source-if-modified-since")
 	if ifModifiedSinceHeader != "" {
-		if !ifModifiedSince(objInfo.ModTime, ifModifiedSinceHeader) {
-			// If the object is not modified since the specified time.
-			writeHeaders()
-			writeErrorResponse(w, ErrPreconditionFailed, r.URL)
-			return true
+		if givenTime, err := parseTime(time.RFC1123, ifModifiedSinceHeader); err == nil {
+			if !ifModifiedSince(objInfo.ModTime, givenTime) {
+				// If the object is not modified since the specified time.
+				writeHeaders()
+				writeErrorResponse(w, ErrPreconditionFailed, r.URL)
+				return true
+			}
 		}
 	}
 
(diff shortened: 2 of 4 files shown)


cmd/time.go
```diff
@@ -0,0 +1,67 @@
...
+	"PDT": -7 * 3600,
+
+	"EST": -5 * 3600,
+	"EDT": -4 * 3600,
+
+	"GMT": 0,
+	"UTC": 0,
+
+	"CET":  1 * 3600,
+	"CEST": 2 * 3600,
```
balamurugana, line 33: Why only these timezone? what is 3600? why can't we use `time.Hour` constant?

cmd/time.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+// parseTime returns the correct time when a timezone is provided
+// with the time string, if the specified timezone is not found in
+// minioTZ, call only the standard golang time parsing.
+func parseTime(layout, value string) (time.Time, error) {
+	// Standard golang time parsing
+	t, err := time.Parse(layout, value)
```
balamurugana, line 41: Why can we use https://golang.org/pkg/time/#ParseInLocation?

cmd/time.go
```diff
@@ -0,0 +1,67 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cmd
+
+import "time"
+
+var minioTZ = map[string]int{
```
harshavardhana, line 21: We need to expand this.. but i think we are better off finding a better approach here.  This will make it harder to add new timezones etc.

harshavardhana (changes requested):

cmd/time.go
vadmeste: > Why can we use https://golang.org/pkg/time/#ParseInLocation?

We can't, ParseInLocation needs location as parameter and the purpose of this PR is to guess the location from the date string. 

For example, if we have this date `Fri, 03 Apr 2015 17:10:00 PST`, golang can parse this string but it doesn't consider PST because it doesn't know it. The parsed date will automatically have UTC timezone which is wrong.

cmd/time.go
vadmeste: > i think we are better off finding a better approach here. This will make it harder to add new timezones etc.

I tried, but there is no other way.

> We need to expand this..

We can add a new dependency for a more complete list.

cmd/time.go
balamurugana: My observation on time/timezone in Go is below

* Go has an [issue](https://github.com/golang/go/issues/18012) supporting abbreviated timezone.
* `time.Parse()` sets abbreviated timezone text, not correct value to know time difference with UTC.
* However `time.ParseInLocation()` works properly when timezone is passed as argument.
* As RFC1123 carries abbreviated timezone we have problem with name collision eg. CST refers to multiple timezone.

Based on the above we could do
* No support other than GMT as per RFC.
* If we support other than GMT, we would need to support all of them (Note: there is no solution to abbreviated timezone name collision).

If we go with option 2,  the right fix would just have a map of abbreviated name to full name and use `time.ParseInLocation()` like below
```go
var NameLocationMap = map[string]string{"PST":"US/Pacific", ...}
t, err = time.Parse(time.RFC1123, timeString)
l, err = t.Location()
location, ok = NameLocationMap[l.String()]
if ok {
    l, err = time.LoadLocation(location)
    t, err = time.ParseInLocation(time.RFC1123, timeString, l)
}

fmt.Println("local time:", t.Local())
fmt.Println("UTC time:", t.UTC())
```

harshavardhana:
> No support other than GMT as per RFC.

IMO we should stick to GMT only as per RFC, and perhaps address this case by case basis as and when the issue arises.  It is expected that SDKs should set the proper values in accordance with HTTP standard.

Otherwise we will end up writing a lot of boilerplate code which would require us to mimic a new time package for parsing and can be done as a separate package.

cmd/time.go
vadmeste: @balamurugana, the problem with locations is that they are taking care of summer time shift, here  US/Pacific can be PST or PDT, that's why I opted the time offset solution (-8, -7, ..)

I knew about timezone collision, but AWS S3 is already doing that..

vadmeste:
> and when the issue arises.

Yes, probably we will never see a client sending a date with a timezone other than GMT, besides the current state is complying with the S3 spec.

Closing this..

PR #4088, opened 2017-04-11
lock/rpc: change rpcPath to be called serviceEndpoint.

harshavardhana:
## Description
lock/rpc: change rpcPath to be called serviceEndpoint.

## Motivation and Context
This is a cleanup to ensure proper naming.

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:

### DIFF
--- a/cmd/lock-rpc-server.go
+++ b/cmd/lock-rpc-server.go
@@ -30,7 +30,10 @@ import (
 
 const (
 	// Lock rpc server endpoint.
-	lockRPCPath = "/lock"
+	lockServicePath = "/lock"
+
+	// Lock rpc service name.
+	lockServiceName = "Dsync"
 
 	// Lock maintenance interval.
 	lockMaintenanceInterval = 1 * time.Minute // 1 minute.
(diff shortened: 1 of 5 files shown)


cmd/lock-rpc-server.go
```diff
@@ -20,7 +20,7 @@ import (
 	"fmt"
 	"math/rand"
 	"net/rpc"
-	"path"
+	pathutil "path"
```
balamurugana: why we need to alias `path` package into `pathutl`?

harshavardhana: the problem is that `path` variable generally just conflicts with the package name and sadly Go doesn't complain properly .. it is just better as a safety. I can keep it `path` as well i don't forse usage if `path` as a variable anytime in future.

balamurugana: I do not see any `path` variable currently.  Even in future, why someone will bring generic `path` variable name to conflict with `path` package?

harshavardhana: It happens in posix.go but it is also naturally possible to use a local variable by someone . which in turn generates this problem. I can keep it `path` for now. 

One more place is namespace-lock.go `nsParam`

balamurugana: a struct having `path` field is not a problem but a variable naming `path` will confuse in general.  That should be avoided.

harshavardhana: Look at the usage of lock() , unlock() see the whole code.. @balamurugana

balamurugana (approved):

fwessels (approved):

PR #4100, opened 2017-04-11
server: Fix a regression in printing startup banner.

### LINKED ISSUE #4099: server: [startup] Startup message endpoints list is not sorted properly.
## Expected Behavior
Endpoints listed in the startup banner should be properly sorted based on last octet.

```
http://192.168.1.106:9000  http://127.0.0.1:9000  http://172.23.0.1:9000  http://172.19.0.1:9000  http://172.24.0.1:9000  http://172.20.0.1:9000  http://172.17.0.1:9000  http://172.21.0.1:9000  http://172.25.0.1:9000  http://172.22.0.1:9000  http://172.18.0.1:9000  http://172.26.0.1:9000  http://172.27.0.1:9000
```

## Current Behavior
Last octet based sorting is lost.

```
Endpoint:  http://127.0.0.1:9000  http://172.17.0.1:9000  http://172.18.0.1:9000  http://172.19.0.1:9000  http://172.20.0.1:9000  http://172.21.0.1:9000  http://172.22.0.1:9000  http://172.23.0.1:9000  http://172.24.0.1:9000  http://172.25.0.1:9000  http://172.26.0.1:9000  http://172.27.0.1:9000  http://192.168.1.106:9000
```

## Possible Solution
Fix `net.go` 

```
Endpoint:  http://127.0.0.1:9000  http://172.17.0.1:9000  http://172.18.0.1:9000  http://172.19.0.1:9000  http://172.20.0.1:9000  http://172.21.0.1:9000  http://172.22.0.1:9000  http://172.23.0.1:9000  http://172.24.0.1:9000  http://172.25.0.1:9000  http://172.26.0.1:9000  http://172.27.0.1:9000  http://192.168.1.106:9000
```

127.0.0.1:9000 is a wrong ip to be preferred as the first entry.

## Context
Startup banners need to be consistent , this is a regression. 

## Your Environment
master branch de204a0a52e6b13fec2d25b73102239891e6de35

harshavardhana:
## Description
Octet based sorting was lost in the previous commit

de204a0a52e6b13fec2d25b73102239891e6de35

## Motivation and Context
This PR fixes a regression - fixes #4099

## How Has This Been Tested?
Manually and using `go test`

## Types of changes

## Checklist:

### DIFF
--- a/cmd/net.go
+++ b/cmd/net.go
@@ -77,17 +77,66 @@ func getHostIP4(host string) (ipList set.StringSet, err error) {
 	return ipList, err
 }
 
+// byLastOctetValue implements sort.Interface used in sorting a list
+// of ip address by their last octet value in descending order.
+type byLastOctetValue []net.IP
+
+func (n byLastOctetValue) Len() int      { return len(n) }
+func (n byLastOctetValue) Swap(i, j int) { n[i], n[j] = n[j], n[i] }
+func (n byLastOctetValue) Less(i, j int) bool {
+	// This case is needed when all ips in the list
+	// have same last octets, Following just ensures that
+	// 127.0.0.1 is moved to the end of the list.
+	if n[i].String() == "127.0.0.1" {
+		return false
+	}
+	if n[j].String() == "127.0.0.1" {
+		return true
+	}
+	return []byte(n[i].To4())[3] > []byte(n[j].To4())[3]
+}
+
+// sortIPs - sort ips based on higher octects.
+// The logic to sort by last octet is implemented to
+// prefer CIDRs with higher octects, this in-turn skips the
+// localhost/loopback address to be not preferred as the
+// first ip on the list. Subsequently this list helps us print
+// a user friendly message with appropriate values.
+func sortIPs(ipList []string) []string {
+	if len(ipList) == 1 {
+		return ipList
+	}
+
+	var ipV4s []net.IP
+	var nonIPs []string
+	for _, ip := range ipList {
+		nip := net.ParseIP(ip)
+		if nip != nil {
+			ipV4s = append(ipV4s, nip)
+		} else {
+			nonIPs = append(nonIPs, ip)
+		}
+	}
+
+	sort.Sort(byLastOctetValue(ipV4s))
+
+	var ips []string
+	for _, ip := range ipV4s {
+		ips = append(ips, ip.String())
+	}
+
+	return append(nonIPs, ips...)
+}
+
 func getAPIEndpoints(serverAddr string) (apiEndpoints []string) {
 	host, port := mustSplitHostPort(serverAddr)
 
 	var ipList []string
 	if host == "" {
-		ipList = localIP4.ToSlice()
+		ipList = sortIPs(localIP4.ToSlice())
 	} else {
 		ipList = []string{host}
 	}
-
-	sort.Strings(ipList)
 
 	scheme := httpScheme
 	if globalIsSSL {
--- a/cmd/net_test.go
+++ b/cmd/net_test.go
@@ -19,6 +19,7 @@ package cmd
 import (
 	"fmt"
 	"net"
+	"reflect"
 	"runtime"
 	"testing"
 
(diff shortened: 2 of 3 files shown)


cmd/net.go
```diff
@@ -37,42 +37,77 @@ func mustSplitHostPort(hostPort string) (host, port string) {
...
+func (n byLastOctetValue) Len() int      { return len(n) }
+func (n byLastOctetValue) Swap(i, j int) { n[i], n[j] = n[j], n[i] }
+func (n byLastOctetValue) Less(i, j int) bool {
+	return []byte(n[i].To4())[3] > []byte(n[j].To4())[3]
+}
+
 // mustGetLocalIP4 returns IPv4 addresses of local host.  It panics on error.
-func mustGetLocalIP4() (ipList set.StringSet) {
-	ipList = set.NewStringSet()
+func mustGetLocalIP4() (ipList []string) {
```
balamurugana: I know local IPs need to be sorted for printing.  Why not do the sorting at printing?

cmd/net.go
```diff
@@ -37,42 +37,77 @@ func mustSplitHostPort(hostPort string) (host, port string) {
...
+	// being the preferred endpoint based on string
+	// sorting, we should choose a higher octet
+	// value instead.
 	for _, ip := range ips {
 		if ip.To4() != nil {
-			ipList.Add(ip.String())
+			ipv4s = append(ipv4s, ip.To4())
 		}
 	}
+	sort.Sort(byLastOctetValue(ipv4s))
```
balamurugana: Where is it useful for non-local hosts?

balamurugana (changes requested):
1. Why do we need octet based sorting over `sort.String()`?
2. What is the expected behavior for below endpoints?
```
http://localhost/d1 
http://127.0.0.1/d2
http://192.168.122.1:10000/d3
http://mylocalhostname:20000/d4
http://remotehostname/d1
http://192.168.63.34:9001/d2
```

In the above `localhost`, `127.0.0.1`, `192.168.122.1` and `mylocalhostrname` are for local host, but `mylocalhostrname` is only accessible by `remotehostname`.  Also `remotehostname` and `192,168.63.34` point to same host and `remotehostname` resolves to `192.168.63.34`

harshavardhana:
> http://localhost/d1 
> http://127.0.0.1/d2
> http://192.168.122.1:10000/d3
> http://mylocalhostname:20000/d4
> http://remotehostname/d1
> http://192.168.63.34:9001/d2

You will never have these type of endpoints.. or mix and match of these.

cmd/net.go
harshavardhana: We need to sort and save it in the form we are going to use.  That is the only purpose of APIEndpoints() .  why would i parse again it into a net.IP form and then print it.

harshavardhana:
Try the patch locally and see the overall change @balamurugana - show me a case where we will print endpoints the above combination.

cmd/net.go
harshavardhana: Not sure i understand..

harshavardhana:
> You will never have these type of endpoints.. or mix and match of these.

Additionally what we are looking at is just IPs there is no Path component etc.

harshavardhana:
ping @balamurugana this PR is updated.

cmd/net.go
```diff
@@ -77,18 +77,58 @@ func getHostIP4(host string) (ipList set.StringSet, err error) {
 	return ipList, err
 }
 
+// byLastOctetValue implements sort.Interface used in sorting a list
+// of ip address by their last octet value.
+type byLastOctetValue []net.IP
```
donatello: Please define the sorting order explicitly, so we don't forget the reason for sorting again - e.g. "Sort a list of IP addresses by the last octet in **descending** order". I feel this is necessary, because it is a weird (non-standard) sorting order.

cmd/net.go
```diff
@@ -77,18 +77,58 @@ func getHostIP4(host string) (ipList set.StringSet, err error) {
...
+
+func (n byLastOctetValue) Len() int      { return len(n) }
+func (n byLastOctetValue) Swap(i, j int) { n[i], n[j] = n[j], n[i] }
+func (n byLastOctetValue) Less(i, j int) bool {
+	return []byte(n[i].To4())[3] > []byte(n[j].To4())[3]
+}
```
donatello: If the stated purpose is to not prefer the loopback address as the first entry, this patch (obviously) does not work for all cases - e.g.

```
127.0.0.1
10.0.0.1
```

Some test code for the above case here - https://play.golang.org/p/26yn5fAGMi

If we really shouldn't print 127.0.0.1 as the first IP, then an appropriate way to sort would be something like:

```go
func (n byLastOctetValue) Less(i, j int) bool {
	if n[i].String() == "127.0.0.1" {
		return false
	}
	if n[j].String() == "127.0.0.1" {
		return true
	}
	return []byte(n[i].To4())[3] > []byte(n[j].To4())[3]
}
```

donatello (changes requested):

cmd/net.go
balamurugana: Basically we want to print `127.xxx.xxx.xxx` at the end of list.  I would prefer custom sorting by checking `strings.HasPrefix(IPStr, "127.")` .  Number sorting would lead to failure like the one you mentioned below.

cmd/net.go
harshavardhana: It is usually based on the last octet.  It is not meant to be not a perfect replacement neither it is going to fully eradicate 127.0.0.1 to be printed as the first element,  it is for a best case scenario only. 

But your change is easy to incorporate let me do that.

cmd/net.go
harshavardhana: Will define explicitly @donatello. 

@balamurugana That is incorrect. A string based mechanism is not what we want here. All we want is a highest octet on the system to be the first in the element not just by string comparing and moving localhost to the end.

balamurugana: > All we want is a highest octet on the system to be the first in the element not just by string comparing and moving localhost to the end.

Personally the highest octet is printed as first doesn't make any value to me and I see its just a matter of choice.  However I don't have any issue with this algo.

donatello (approved):

balamurugana (approved):

Issue #4110, opened 2017-04-13
lock instrumentation shouldn't return duration which can be computed from time of lock


krisis:
## Expected Behavior

Ref #4108 
>Lock APIs

  >  In the list api do we need options since and duration? Need more info.
    older-than : Specify time format with an example.

Lock instrumentation should return time since when the namespace lock is held alone. The time elapsed since can be derived.

## Current Behavior

We return both time since when the lock is held and the duration computed at the time of `List Locks` admin API call.

## Possible Solution

Return only the time since when the lock is held on Minio server.

## Steps to Reproduce (for bugs)

1. `mc admin lock list alias/bucket`
2.
3.
4.

## Your Environment

* Version used:
* Environment name and version (e.g. nginx 1.9.1):
* Server type and version:
* Operating System and version:
* Link to your project:

harshavardhana:
Fixed by #4111

PR #4111, opened 2017-04-13
Remove duration field from lock instrumentation

krisis:
## Description

Duration for which a lock was held can be computed from the `Since`
field of `OpsLockState`. It is the difference between current time and
time at which the namespace lock was held. This change avoids
superfluous instrumentation.

## Motivation and Context

See #4110 
## Types of changes

## Checklist:

### DIFF
--- a/cmd/lock-instrument_test.go
+++ b/cmd/lock-instrument_test.go
@@ -141,7 +141,6 @@ func getSystemLockState() (SystemLockState, error) {
 				LockType:    lockInfo.lType,
 				Status:      lockInfo.status,
 				Since:       lockInfo.since,
-				Duration:    UTCNow().Sub(lockInfo.since),
 			})
 		}
 		lockState.LocksInfoPerObject = append(lockState.LocksInfoPerObject, volLockInfo)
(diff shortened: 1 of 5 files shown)


vadmeste:
@krisis umm, are we sure? doesn't that require from the client to have the correct time to correctly compute the duration ?

donatello:
Tests need fix.

krisis:
@vadmeste 
>@krisis umm, are we sure? doesn't that require from the client to have the correct time to correctly compute the duration ?

Interesting point you raise. If we can't expect clients to be in (time) sync with Minio server, then we will need both the time when the lock was held (as per Minio server) and duration. 

@harshavardhana @krishnasrinivas Is it correct to expect Minio server and clients like `mc` to have their time in sync?

harshavardhana:
> @harshavardhana @krishnasrinivas Is it correct to expect Minio server and clients like mc to have their time in sync?

They can be different timezones but they should be in UTC sync - 15mins is the window allowed for time skew between `mc` and `minio`

vadmeste:
Or we do the reverse, we un-export Since field and keep Duration

krisis:
@vadmeste, Like we discussed, `Since` and `Duration` fields are both equivalent, given that the server and clients can't be farther than 15 minutes, I have decided to retain `Since` and remove `Duration` as it is in this PR. Could you provide further review comments if any?

vadmeste (approved):

harshavardhana (approved):

PR #4115, opened 2017-04-13
Enable build on solaris

ptribble:
## Description

The minio server fails to build on solaris.

To enable solaris, I have enabled solaris for rlimit-file and rlimit-memory, and added and implementation of stats_solaris.go

## Motivation and Context

With these changes, minio builds on solaris.

## How Has This Been Tested?

A build using 'go get -u' produces a minio executable on solaris

I have run and tested minio server - I can list buckets, list files, upload and download files, create buckets.

(The alternative build using checkout and make doesn't work properly on solaris. That's a slightly different issue unrelated to code or functionality that I would also like to fix.)

## Types of changes

## Checklist:

I'm not quite sure how to run the tests. I have ensured that 'make verifiers' is clean, though. And I'm fairly sure that I haven't done anything that will affect any other platforms.


pkg/sys/stats_solaris.go
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	"os"
+)
+
+/*
+#include <unistd.h>
+*/
+import (
+	"C"
```
harshavardhana: We shouldn't use CGO based dependencies in Minio server. You should implement something Go native. @ptribble

harshavardhana (changes requested):

pkg/sys/stats_solaris.go
```diff
@@ -0,0 +1,41 @@
+/*
+ * Minio Cloud Storage, (C) 2016,2017 Minio, Inc.
```
balamurugana: New file needs to have `2017` only.

pkg/sys/stats_solaris.go
harshavardhana: Generally as a practice we would like to avoid this, because this causes requirements of `cgo`. Minio is generally built with CGO_ENABLED=0 and we want to keep it that way for simplicity and avoid depending on gcc + libc altogether. 

There might be some limitations on how much Solaris is supported but currently by Go - you can see an example on how it should be implemented here 

https://github.com/golang/sys/blob/master/unix/syscall_solaris.go#L706

harshavardhana: Another way is simply return error as not implemented and avoid this altogether. This package is used only by caching layer which is enabled during erasure or distributed erasure minio server setup. For regular `fs` setups this is not needed 

We can keep this as a FIXME and address this in subsequent time. @ptribble

harshavardhana (changes requested):

ptribble:
Thanks for the comments.

OK, let me think about this. I haven't yet found a way to do this without cgo, but I'll dig deeper.

(On solaris, things are a bit different. You always have to have libc, because that [not syscall] is the interface to the system.)

harshavardhana:
> Thanks for the comments.
> 
> OK, let me think about this. I haven't yet found a way to do this without cgo, but I'll dig deeper.
> 
> (On solaris, things are a bit different. You always have to have libc, because that [not syscall] is the interface to the system.)

I would suggest we do just return `return 0, errors.New("Not implemented")` for now so we can get this built properly and work on proper version using `x/sys/unix` at a later stage?  @ptribble

harshavardhana:
@ptribble are you still working on this PR? were you able to look at my suggestion above?

deekoder:
@ptribble I suggest we close this PR and you can send us a new one when you can without the cgo dependency. Would you be in agreement with this proposal?

ptribble:
Sorry about the delay; holiday weekend.

I've updated the PR with a stub implementation in stats_solaris.go, and corrected the date.

The build completes, the make checks are clean, and minio server has been tested in fs mode successfully.

I agree, we can defer a full implementation until later.

harshavardhana:
> I've updated the PR with a stub implementation in stats_solaris.go, and corrected the date.

Thanks @ptribble - taking a look.

harshavardhana (approved):

balamurugana (changes requested):
solaris cross build fails.

```bash
$ GOOS=solaris go build 
vendor/github.com/mattn/go-isatty/isatty_solaris.go:7:2: cannot find package "golang.org/x/sys/unix" in any of:
	/home/bala/golang/gopath/src/github.com/minio/minio/vendor/golang.org/x/sys/unix (vendor tree)
	/home/bala/golang/goroot/src/golang.org/x/sys/unix (from $GOROOT)
	/home/bala/golang/gopath/src/golang.org/x/sys/unix (from $GOPATH)
```

my go version
```bash
$ go version
go version go1.7.5 linux/amd64
```

harshavardhana:
> $ GOOS=solaris go build

I know the problem.. it is perhaps because we have failed to vendorize `golang.org/x/sys/unix`

harshavardhana:
@ptribble can you merge this PR in here https://github.com/minio/minio/pull/4161 - this would completely fix all the build issues on Solaris.

ptribble:
Sure, I can do that. May be a couple of day before I can get to it.

ptribble:
Ok, I've merged 4161, and also fixed up checkdeps.sh for Solaris (which appears to immediately cause the make style build to fail).

harshavardhana:
> Ok, I've merged 4161, and also fixed up checkdeps.sh for Solaris (which appears to immediately cause the make style build to fail).

Thanks a lot @ptribble

harshavardhana (approved):

buildscripts/checkdeps.sh
```diff
@@ -25,6 +25,11 @@ _init() {
     OSX_VERSION="10.8"
     KNAME=$(uname -s)
     ARCH=$(uname -m)
```
balamurugana, line 27: does `uname -m` exist in Solaris? does `isainfo -k` provide different value over `uname -m`?

ptribble: On Solaris, uname -m returns the type of processor. However, Solaris is dual 32/64-bit so all that uname -m tells you is that you're running on Intel (as opposed to sparc or whatever) - it doesn't tell you whether you're 32 or 64-bit. The isainfo command tells you which instruction set architectures are available; isainfo -k tells you which variant the kernel is running in.

See below for the output on my system:

% uname -m
i86pc
% isainfo
amd64 i386
% isainfo -k
amd64

The checkdeps.sh script is looking for what isainfo -k gives on a Solaris system.

balamurugana: @ptribble got it. thanks.

balamurugana (approved):

harshavardhana:
Thanks for the efforts here @ptribble

Issue #4140, opened 2017-04-17
The remote volumes will not be found when adding new nodes into minio distributed mode


afterhill:
When adding new new node into minio distributed modes,  the following error was repeated in living container and the entire minio system crashed:

```
$ sudo docker logs -f 4cfe
Created minio configuration file successfully at /root/.minio
time="2017-04-17T06:37:43Z" level=error msg="Failed to lookup host" cause="lookup minio1 on 127.0.0.11:53: server misbehaving" source="[object-api-common.go:149:isLocalStorage()]" 
.... repeat 20+ times....
Initializing data volume for first time. Waiting for other servers to come online (elapsed 22s)
.... repeat endlessly...
```

## Expected Behavior
The new node was merged properly.

## Current Behavior
The entire minio system will crash and all of running container will repeat the following error message:

```
$ sudo docker logs -f 4cfe
Initializing data volume for first time. Waiting for other servers to come online (elapsed 22s)
.... repeat endlessly...
```

## Possible Solution
Remove all of `.minio.sys` folder will bring the minio system up immediately!

## Steps to Reproduce (for bugs)
1. 2 virtual box VMs, 3 disks for each VM.
2.  $docker-machine ls:
    ```
    minio-swarm-01   -        generic   Running   tcp://10.25.3.11:2376           v17.04.0-ce   
    minio-swarm-02   -        generic   Running   tcp://10.25.3.12:2376           v17.04.0-ce 
    ```
3. lsblk VM1:

**/data1 has some existing data and /data1 has the directory .minio.sys, /data2 and /data3 are empty**

```
NAME   MAJ:MIN RM   SIZE RO TYPE MOUNTPOINT
sda      8:0    0    40G  0 disk 
└─sda1   8:1    0    40G  0 part /
sdb      8:16   0   1.9T  0 disk 
└─sdb1   8:17   0   1.9T  0 part /data1
sdc      8:32   0   1.9T  0 disk 
└─sdc1   8:33   0   1.9T  0 part /data2
sdd      8:48   0   1.9T  0 disk 
└─sdd1   8:49   0   1.9T  0 part /data3
```

4. lsblk VM2:

**All of /data1 /data2 /data3 are new and empty**

```
NAME   MAJ:MIN RM   SIZE RO TYPE MOUNTPOINT
sda      8:0    0    40G  0 disk 
└─sda1   8:1    0    40G  0 part /
sdb      8:16   0   1.9T  0 disk 
[...]

PR #4142, opened 2017-04-17
server: Fix message when corrupted or unsupported format is found.

harshavardhana:
## Description
This is a fix to provide a little more elaborate message.

## Motivation and Context
Refer https://github.com/minio/minio/issues/4140

## How Has This Been Tested?
Manually.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/prepare-storage.go
+++ b/cmd/prepare-storage.go
@@ -188,6 +188,9 @@ func printRetryMsg(sErrs []error, storageDisks []StorageAPI) {
 	}
 }
 
+// Maximum retry attempts.
+const maxRetryAttempts = 5
+
 // Implements a jitter backoff loop for formatting all disks during
 // initialization of the server.
 func retryFormattingXLDisks(firstDisk bool, endpoints EndpointList, storageDisks []StorageAPI) error {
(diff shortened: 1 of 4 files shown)


cmd/storage-rpc-client.go
```diff
@@ -116,7 +116,8 @@ func newStorageRPC(endpoint Endpoint) StorageAPI {
 // Stringer interface compatible representation of network device.
 func (n *networkStorage) String() string {
 	// Remove the storage RPC path prefix, internal paths are meaningless.
-	serviceEndpoint := strings.TrimPrefix(n.rpcClient.ServiceEndpoint(), storageRPCPath)
+	serviceEndpoint := strings.TrimPrefix(n.rpcClient.ServiceEndpoint(),
+		path.Join(minioReservedBucketPath, storageRPCPath))
 	return n.rpcClient.ServerAddr() + ":" + serviceEndpoint
```
krisis: Could you add a unit test to test `String()` function of network storage for a bunch of endpoint paths?

cmd/prepare-storage.go
```diff
@@ -229,7 +229,11 @@ func retryFormattingXLDisks(firstDisk bool, endpoints EndpointList, storageDisks
 			// most part unless one of the formats is not consistent
 			// with expected XL format. For example if a user is trying
 			// to pool FS backend.
-			if err := checkFormatXLValues(formatConfigs); err != nil {
+			if index, err := checkFormatXLValues(formatConfigs); err != nil {
+				if retryCount < 5 {
```
krisis: We should use a `const` instead of a `int` literal.

cmd/storage-rpc-client.go
harshavardhana: Sure.

cmd/prepare-storage.go
harshavardhana: Sure!

cmd/prepare-storage.go
```diff
@@ -219,17 +222,22 @@ func retryFormattingXLDisks(firstDisk bool, endpoints EndpointList, storageDisks
...
-			// with expected XL format. For example if a user is trying
-			// to pool FS backend.
-			if err := checkFormatXLValues(formatConfigs); err != nil {
+			// with expected XL format. For example if a user is
+			// trying to pool FS backend.
+			if index, err := checkFormatXLValues(formatConfigs); err != nil {
+				if retryCount > maxRetryAttempts {
+					errorIf(err, "Detected disk (%s) in unexpected format",
+						storageDisks[index])
+					continue
```
vadmeste: Why we 'continue' here ? Shouldn't we should immediately quit if format xl format is corrupted.

cmd/storage-rpc-client.go
```diff
@@ -116,8 +116,13 @@ func newStorageRPC(endpoint Endpoint) StorageAPI {
 // Stringer interface compatible representation of network device.
 func (n *networkStorage) String() string {
 	// Remove the storage RPC path prefix, internal paths are meaningless.
-	serviceEndpoint := strings.TrimPrefix(n.rpcClient.ServiceEndpoint(), storageRPCPath)
-	return n.rpcClient.ServerAddr() + ":" + serviceEndpoint
+	serviceEndpoint := strings.TrimPrefix(n.rpcClient.ServiceEndpoint(),
```
vadmeste, line 119: This looks like needed to be sent in a separate PR. Also no change in rpc server to support the new rpc path ?

harshavardhana: The path is only canonicalized here. Server doesn't need any change. See this is a String method.

cmd/prepare-storage.go
harshavardhana: We do is to avoid timing issue of how we need to deal with servers coming online. See the original issue.

vadmeste: yes, I understand, could you add a comment?

harshavardhana: Sure..

vadmeste (approved):

krisis (approved):

Issue #4139, opened 2017-04-17
server: Distributed erasure coded leaves blocked locks even after client cancelled the request


harshavardhana:
SIGABRT crash report - [retryLoop](https://github.com/minio/minio/files/924985/filename.txt)

## Expected Behavior
No hang to be observed when running dicom import tool. 

## Current Behavior
Stale locks can be observed. 
```
dicom/b0/7e6acceb2b75864085763066bee2f520ceb638e6f6f8522d538c30c927de54ba7da43cab94f5ade07b55e7cbf01f12919efdd69035de96de7e33ef1ffc8c14 (LocksOnObject: 0, locksAcquiredOnObject: 0, totalBlockLocks:0):   {OperationID:8RJQGJ5SI04TRWSA LockSource:[object-handlers.go:491:objectAPIHandlers.PutObjectHandler()] LockType:WLock Status:Blocked Since:2017-04-17 05:19:49.390435825 +0000 UTC Duration:0s}

dicom/29/945e8a5c7e6401831dd65c1b08db005a9dfa230598cbff3a5e723660b48106ee9319a85d0e332210870e3b4e45c87c8dc2ac02a96815eab0d0712be794c51d (LocksOnObject: 0, locksAcquiredOnObject: 0, totalBlockLocks:0):   {OperationID:HZU063CUNAI0MZSW LockSource:[object-handlers.go:491:objectAPIHandlers.PutObjectHandler()] LockType:WLock Status:Blocked Since:2017-04-17 05:19:49.232126599 +0000 UTC Duration:0s}

dicom/1d/b520e3957b6b2e0a01b798904d0efea05f423c837a1867040608fdd2aaa9252fede5141c929b317a54ad0cb22c3c49428fa841c93ecad42adc98e42bf2ab51 (LocksOnObject: 0, locksAcquiredOnObject: 0, totalBlockLocks:0):   {OperationID:GO0U4ACFRDYEG519 LockSource:[object-handlers.go:491:objectAPIHandlers.PutObjectHandler()] LockType:WLock Status:Blocked Since:2017-04-17 05:20:03.222637985 +0000 UTC Duration:0s}

[...]

PR #4146, opened 2017-04-19
server/mux: Fix serverMux to set deadlines based on UTC time.

harshavardhana:
## Description
Avoid using `time.Now()` instead rely on UTC time
for the final deadline, this is to be consistent with
all our internal functions.

Reduce the default read timeout to 15 seconds
in lieu with a newly discovered issue
   - https://github.com/minio/minio/issues/4139

Additionally also change the Read() conn wrapper
to set deadline only upon successful Reads().

## Motivation and Context
https://github.com/minio/minio/issues/4139

## How Has This Been Tested?
Manually with large sets of parallel requests around 10,000 concurrent operations. 

## Types of changes

## Checklist:

### DIFF
--- a/cmd/server-mux.go
+++ b/cmd/server-mux.go
@@ -113,30 +113,41 @@ func (c *ConnMux) PeekProtocol() (string, error) {
 	return protocolTLS, nil
 }
 
-// Read - streams the ConnMux buffer when reset flag is activated, otherwise
-// streams from the incoming network connection
-func (c *ConnMux) Read(b []byte) (n int, e error) {
-	// Push read deadline
-	c.Conn.SetReadDeadline(time.Now().Add(defaultTCPReadTimeout))
+// Read reads from the tcp session for data sent by
+// the client, additionally sets deadline for 15 secs
+// after each successful read. Deadline cancels and
+// returns error if the client does not send any
+// data in 15 secs. Also keeps track of the total
+// bytes received from the client.
+func (c *ConnMux) Read(b []byte) (n int, err error) {
+	// Update total incoming number of bytes.
+	defer globalConnStats.incInputBytes(n)
 
-	// Update server's connection statistics
-	defer func() {
-		globalConnStats.incInputBytes(n)
-	}()
+	n, err = c.peeker.Read(b)
+	if err != nil {
+		return n, err
+	}
 
-	return c.peeker.Read(b)
+	// Read deadline was already set previously, set again
+	// after a successful read operation for future read
+	// operations.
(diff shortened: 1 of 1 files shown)


cmd/server-mux.go
```diff
@@ -113,30 +113,46 @@ func (c *ConnMux) PeekProtocol() (string, error) {
...
+		return n, err
+	}
+
+	// Read deadline was already set previously, set again
+	// after a successful read operation for future read
+	// operations.
+	c.Conn.SetReadDeadline(UTCNow().Add(defaultTCPReadTimeout))
+
+	// Update total incoming number of bytes.
+	globalConnStats.incInputBytes(n)
```
vadmeste: Updating stats only when Read() is successful will lose some counting accuracy here, like when a client sends some data and prematurely closes the connection.

cmd/server-mux.go
```diff
@@ -113,30 +113,46 @@ func (c *ConnMux) PeekProtocol() (string, error) {
...
+// Write to the client over a tcp session, additionally
+// keeps track of the total bytes written by the server.
+func (c *ConnMux) Write(b []byte) (n int, err error) {
+	n, err = c.Conn.Write(b)
+	if err != nil {
+		return n, err
+	}
+
+	// Update total outgoing number of bytes.
+	globalConnStats.incOutputBytes(n)
```
vadmeste: Updating stats only when Write() is successful will lose some counting accuracy here, like when a client  prematurely closes the connection.

harshavardhana: Yes Only success is counted because we don't need remember data which we didn't write to disk.

vadmeste: > Yes Only success is counted because we don't need remember data which we didn't write to disk.

I see. But our stats (input/output bytes) won't look reliable that way. Besides, if a user wants to see data written in the disk, he won't need the exact values, the approximation will be sufficient for him.

cmd/server-mux.go
harshavardhana: In my understanding that is how it was even now with current master. It only makes sense to track success. Otherwise this would give misleading values since we need to track error separately.

cmd/server-mux.go
harshavardhana: The only way to do is separate out error and success counter then combine them before writing to client.

cmd/server-mux.go
vadmeste: ConnStats() counts all bytes incoming/outgoing in all sockets

harshavardhana: Look at the current usage of n instead of len(b)

vadmeste (approved):

donatello (approved):

PR #4156, opened 2017-04-20
Disk caching support

krishnasrinivas:
## Description

DiskCache for gateway and server

## Motivation and Context

Gateway can use local disk for caching providing performance boost.

Note: Unit test cases are minimal as I don't know how we can unit test caching without manual intervention. I will see what other tests can be added.,If there are any suggestions on unit test cases please let me know.

## How Has This Been Tested?

manually

## Types of changes

## Checklist:


cmd/disk-cache.go
```diff
@@ -9,14 +9,29 @@ import (
...
-	"syscall"
 	"time"
 
+	"github.com/boltdb/bolt"
 	"github.com/minio/minio/pkg/disk"
 
 	"encoding/json"
 )
 
+const diskCacheBackendVersion = "1"
```
harshavardhana: Can you put this like and add a comment for each constant. 

```
const (
      diskCacheBackendVersion = "1"
      diskCacheObjectMetaVersion = "1.0.0"
)
```

cmd/disk-cache.go
```diff
@@ -9,14 +9,29 @@ import (
...
+const diskCacheObjectMetaVersion = "1.0.0"
+
+type diskCacheBoltdbEntry map[string]string
+
```
harshavardhana: Add comments..

harshavardhana (changes requested):

cmd/api-router.go
```diff
@@ -21,13 +21,22 @@ import router "github.com/gorilla/mux"
...
+	cache     *cacheObjects
 }
 
 // registerAPIRouter - registers S3 compatible APIs.
 func registerAPIRouter(mux *router.Router) {
+	var cache *cacheObjects
+	var err error
+	if globalCacheDir != "" {
+		cache, err = newServerCacheObjects(globalObjectAPI, globalCacheDir, globalCacheMax, 0)
+		fatalIf(err, "Unable to init cache")
```
harshavardhana: "Unable to initialize disk caching."

cmd/disk-cache_test.go
```diff
@@ -0,0 +1,88 @@
+package cmd
```
harshavardhana, line 17: License header missing.

cmd/gateway-main.go
```diff
@@ -213,6 +217,9 @@ func gatewayMain(ctx *cli.Context) {
 
 	initNSLock(false) // Enable local namespace lock.
 
+	globalCacheDir = ctx.String("cache-dir")
+	globalCacheMax = ctx.Int("cache-max")
```
harshavardhana, line 221: Mention what does this signify, both the options with a comment.

cmd/gateway-router.go
```diff
@@ -50,11 +50,16 @@ type gatewayAPIHandlers struct {
 
 // registerAPIRouter - registers S3 compatible APIs.
 func registerGatewayAPIRouter(mux *router.Router, gw GatewayLayer) {
+	cache, err := newGatewayCacheObjects(gw, globalCacheDir, globalCacheMax, 0)
+	if err != nil {
```
harshavardhana: No need to do 'if err != nil`

cmd/gateway-s3.go
```diff
@@ -52,6 +54,14 @@ func s3ToObjectError(err error, params ...string) error {
 		object = params[1]
 	}
 
+	var urlErr *url.Error
```
harshavardhana: This is more complete. 

```
// isNetworkError - is network error.. 
func isNetworkError(err error) bool {
        if err == nil {
                return false
        }
        switch err.(type) {
        case net.Error:
                switch err.(type) {
                case *net.DNSError, *net.OpError, net.UnknownNetworkError:
                        return true
                case *url.Error:
                        // For a URL error, where it replies back "connection closed"
                        if strings.Contains(err.Error(), "Connection closed by foreign host") {
                                return true
                        }
                default:
                        if strings.Contains(err.Error(), "net/http: TLS handshake timeout") {
                                // If error is - tlsHandshakeTimeoutError,.
                                return true
                        } else if strings.Contains(err.Error(), "i/o timeout") {
                                // If error is - tcp timeoutError.
                                return true
                        } else if strings.Contains(err.Error(), "connection timed out") {
                                // If err is a net.Dial timeout.
                                return true
                        }
                }
        }
        return false
}
```

cmd/globals.go
```diff
@@ -114,6 +114,9 @@ var (
 	globalActiveCred         credential
 	globalPublicCerts        []*x509.Certificate
 	globalXLObjCacheDisabled bool
+
```
harshavardhana, line 117: Comments needed.. for each global option.

cmd/server-main.go
```diff
@@ -33,6 +33,16 @@ var serverFlags = []cli.Flag{
...
 	},
+	cli.StringFlag{
+		Name:  "cache-dir",
+		Value: "",
+		Usage: "Cache directory.",
+	},
+	cli.StringFlag{
+		Name:  "cache-max",
+		Value: "80",
+		Usage: "Percent of disk space for Cache.",
```
harshavardhana: "Percent of disk space used for caching."

cmd/server-main.go
```diff
@@ -33,6 +33,16 @@ var serverFlags = []cli.Flag{
 		Value: ":9000",
 		Usage: "Bind to a specific ADDRESS:PORT, ADDRESS can be an IP or hostname.",
 	},
+	cli.StringFlag{
+		Name:  "cache-dir",
+		Value: "",
+		Usage: "Cache directory.",
```
harshavardhana: "Location of cache directory"

cmd/server-main.go
```diff
@@ -250,6 +263,13 @@ func serverMain(ctx *cli.Context) {
 	// Initialize name space lock.
 	initNSLock(globalIsDistXL)
 
+	newObject, err := newObjectLayer(globalEndpoints)
```
harshavardhana: Moving this here won't work for distributed setup. For the disks to connect you need HTTP service to run first.

harshavardhana: If no good reason then move back to after HTTP server has initialized.

harshavardhana (changes requested):
First round of review.. haven't looked into disk-cache.go yet.

cmd/disk-cache.go
```diff
@@ -0,0 +1,674 @@
+package cmd
```
harshavardhana, line 1: License header missing.

cmd/disk-cache.go
```diff
@@ -0,0 +1,674 @@
...
+		errorIf(err, "Error getting disk information on %s", c.dir)
+		return false
+	}
+	usedPercent := (di.Total - di.Free) * 100 / di.Total
+	return int(usedPercent) < minUsage
+}
+
+// Retusn if the disk usage is high.
+// Disk usage is high if disk used is > maxUsage
+func (c diskCache) diskUsageHigh() bool {
```
harshavardhana, line 161: Can above two functions be combined? as single function which basically gives high/low behavior based on a boolean input?

cmd/disk-cache.go
```diff
@@ -0,0 +1,674 @@
...
+func (c diskCache) encodedMetaPath(bucket, object string) string {
+	return path.Join(c.dataDir, fmt.Sprintf("%x.json", sha256.Sum256([]byte(path.Join(bucket, object)))))
+}
+
+// Commit the cached object - rename from tmp directory to data directory.
+func (c diskCache) Commit(f *os.File, objInfo ObjectInfo, anon bool) error {
+	bucket := objInfo.Bucket
+	object := objInfo.Name
+
+	objectLock := globalNSMutex.NewNSLock(diskCacheLockingPrefix+bucket, object)
```
harshavardhana: why not pathJoin here? rather than +

cmd/disk-cache.go
```diff
@@ -0,0 +1,674 @@
...
+	objectLock.Lock()
+	defer objectLock.Unlock()
+
+	encPath := c.encodedPath(bucket, object)
+	if err := os.Rename(f.Name(), encPath); err != nil {
```
harshavardhana, line 248: This won't work on windows.. renaming an active fd. You should use the `lock.LockedOpenFile()` which is allowed to be renamed in this manner.

cmd/disk-cache.go
```diff
@@ -0,0 +1,674 @@
...
+	objectLock.RLock()
+	defer objectLock.RUnlock()
+
+	metaPath := c.encodedMetaPath(bucket, object)
+	objMeta := diskCacheObjectMeta{}
+	metaBytes, err := ioutil.ReadFile(metaPath)
+	if err != nil {
+		return nil, ObjectInfo{}, false, err
+	}
+	if err = json.Unmarshal(metaBytes, &objMeta); err != nil {
```
harshavardhana, line 323: Use gson to unmarshal for better performance.. look at xl-v1-metadata.go and fs-v1-metadata.go

cmd/disk-cache.go
```diff
@@ -0,0 +1,674 @@
...
+		return nil, ObjectInfo{}, false, err
+	}
+	if objMeta.Version != diskCacheObjectMetaVersion {
+		return nil, ObjectInfo{}, false, errors.New("format not supported")
```
harshavardhana, line 327: Use `errCorruptedFormat`

cmd/disk-cache.go
```diff
@@ -0,0 +1,674 @@
...
+type diskCacheFormat struct {
+	Version string    `json:"version"`
+	Format  string    `json:"format"`
+	Time    time.Time `json:"createTime"`
+}
+
+// Inits the cache-dir if it is not init'ed already.
+// Initializing implies creation of cache-dir, data-dir, tmp-dir and format.json
+func newDiskCache(dir string, maxUsage, expiry int) (*diskCache, error) {
+	if err := os.MkdirAll(dir, 0766); err != nil {
```
harshavardhana, line 396: Is it allowed to have simultaneous minio process to write to same cache location? if not can we prevent it by holding lock properly like how we do in `fs` ?

cmd/disk-cache.go
```diff
@@ -0,0 +1,674 @@
...
+type cacheObjects struct {
+	dcache *diskCache
+
+	// Object functions pointing to the corresponding functions of backend implementation.
+	GetObjectFn         func(bucket, object string, startOffset int64, length int64, writer io.Writer) (err error)
+	GetObjectInfoFn     func(bucket, object string) (objInfo ObjectInfo, err error)
+	PutObjectFn         func(bucket, object string, size int64, data io.Reader, metadata map[string]string, sha256sum string) (objInfo ObjectInfo, err error)
+	AnonGetObjectFn     func(bucket, object string, startOffset int64, length int64, writer io.Writer) (err error)
+	AnonGetObjectInfoFn func(bucket, object string) (objInfo ObjectInfo, err error)
+	AnonPutObjectFn     func(bucket, object string, size int64, data io.Reader, metadata map[string]string, sha256sum string) (objInfo ObjectInfo, err error)
```
harshavardhana, line 468: Is multipart not supported?

cmd/disk-cache.go
```diff
@@ -0,0 +1,674 @@
...
+	GetObjectFn := c.GetObjectFn
+	GetObjectInfoFn := c.GetObjectInfoFn
+	GetObjectInfo := c.GetObjectInfo
+	if anonReq {
+		GetObjectFn = c.AnonGetObjectFn
+		GetObjectInfoFn = c.AnonGetObjectInfoFn
+		GetObjectInfo = c.AnonGetObjectInfo
+	}
+	objInfo, err := GetObjectInfo(bucket, object)
+	_, backendDown := errorCause(err).(BackendDown)
```
harshavardhana, line 482: This needs to be a function `isErrBackendDown`

cmd/disk-cache.go
```diff
@@ -0,0 +1,674 @@
...
+	if err != nil && !backendDown {
+		if _, ok := errorCause(err).(ObjectNotFound); ok {
+			// Delte the cached entry if backend object was deleted.
```
harshavardhana: Typo `delete`

cmd/disk-cache.go
```diff
@@ -0,0 +1,674 @@
...
+		return err
+	}
+	r, cachedObjInfo, anon, err := c.dcache.Get(bucket, object)
+	if err == nil {
+		defer r.Close()
+		if backendDown {
+			if anonReq {
+				if anon {
+					_, err = io.Copy(writer, io.NewSectionReader(r, startOffset, length))
+					return err
```
harshavardhana, line 497: Highly nested if conditions, can you find a way to simplify this?

harshavardhana (changes requested):
Initial review for disk-cache.go

cmd/gateway-handlers.go
```diff
@@ -243,28 +255,44 @@ func (api gatewayAPIHandlers) PutObjectHandler(w http.ResponseWriter, r *http.Re
...
 	var objInfo ObjectInfo
 	switch reqAuthType {
 	case authTypeAnonymous:
 		// Create anonymous object.
-		objInfo, err = objectAPI.AnonPutObject(bucket, object, size, r.Body, metadata, sha256sum)
+		putObject = objectAPI.AnonPutObject
+		if api.cache != nil {
+			putObject = api.cache.AnonPutObject
+		}
+		// objInfo, err = objectAPI.AnonPutObject(bucket, object, size, r.Body, metadata, sha256sum)
```
vadmeste: Useless comment ?

cmd/api-router.go
krishnasrinivas: done

cmd/disk-cache_test.go
krishnasrinivas: done

cmd/disk-cache_test.go
```diff
@@ -1,3 +1,19 @@
+/*
+ * Minio Cloud Storage, (C) 2016 Minio, Inc.
```
harshavardhana, line 2: s/2016/2017

cmd/server-main.go
krishnasrinivas: moved back but changes were needed to newServerCacheObjects to accomodate the move back

harshavardhana (changes requested):

cmd/disk-cache.go
krishnasrinivas: no, its more readable with Low/High words than with boolean, also implementation is clearer with separate functions

cmd/disk-cache.go
krishnasrinivas: errCorruptedFormat will not be right because it gives the impression that the format file is itself corrupted. But this is a case where older binary running on newer formats.

cmd/disk-cache.go
krishnasrinivas: we do locking already in disk-cache

cmd/disk-cache.go
krishnasrinivas: no

cmd/disk-cache.go
krishnasrinivas: using isNetworkOrHostDown() as function name, using implementation provided by you in one of the comments

cmd/disk-cache.go
harshavardhana: Then use `errFormatNotSupported`

cmd/disk-cache.go
harshavardhana: can you point me to the right location?

cmd/disk-cache.go
harshavardhana: Can you write that why its not supported, so that someone doesn't implement it future.

cmd/disk-cache.go
krishnasrinivas: yes this piece is highly nested, could not simplify it further

cmd/disk-cache.go
```diff
@@ -0,0 +1,680 @@
...
+	_, backendDown := errorCause(err).(BackendDown)
+	if err != nil && !backendDown {
+		if _, ok := errorCause(err).(ObjectNotFound); ok {
+			// Delete the cached entry if backend object was deleted.
+			c.dcache.Delete(bucket, object)
+		}
+		return err
+	}
+	r, cachedObjInfo, anon, err := c.dcache.Get(bucket, object)
+	if err == nil {
```
harshavardhana, line 491: In this condition is it okay to ignore `err != nil` ? and we should definitely simplify the following code.

cmd/disk-cache.go
```diff
@@ -0,0 +1,680 @@
...
+		if _, ok := errorCause(err).(BackendDown); !ok {
+			if _, ok = errorCause(err).(ObjectNotFound); ok {
+				// Delte the cached entry if backend object was deleted.
+				c.dcache.Delete(bucket, object)
+			}
+			return ObjectInfo{}, err
+		}
+		cachedObjInfo, anon, infoErr := c.dcache.GetObjectInfo(bucket, object)
+		if infoErr == nil {
+			if !anonReq {
```
harshavardhana, line 564: What is the difference between anon and anonReq?  perhaps we can choose a better variable name to differentiate their purpose.

harshavardhana (changes requested):
Larger change here is boltdb is not maintained anymore https://github.com/boltdb/bolt#a-message-from-the-author 

We should move to the fork by coreos - https://github.com/coreos/bbolt

donatello:
Since I am going to take over this PR, shall i open a new one? I managed to rebase this PR after squashing commits on master, however, I cannot force push to this PR as it is in @krishnasrinivas's fork. The plan would be to continue work on my branch, and finally squash and send two commits for final merge to master, one would be krishna's patch (this PR with all commits squashed into one) and the other would be my changes to it (also in a single commit). Thus, two commits by different authors would be added to master. Is this acceptable?

krishnasrinivas:
@donatello sure go ahead

harshavardhana:
> Since I am going to take over this PR, shall i open a new one? I managed to rebase this PR after squashing commits on master, however, I cannot force push to this PR as it is in @krishnasrinivas's fork. The plan would be to continue work on my branch, and finally squash and send two commits for final merge to master, one would be krishna's patch (this PR with all commits squashed into one) and the other would be my changes to it (also in a single commit). Thus, two commits by different authors would be added to master. Is this acceptable?

@krishnasrinivas you can close this PR.

donatello:
This PR is being integrated in #4922 where it is currently rebased on latest master.

PR #4157, opened 2017-04-20
Remove hardcoded min and max limit for erasure coding

fwessels:
## Description
Remove hardcoded min and max limit for erasure coding

## How Has This Been Tested?
Tested manually 

## Checklist:

### DIFF
--- a/cmd/endpoint.go
+++ b/cmd/endpoint.go
@@ -185,23 +185,9 @@ func NewEndpointList(args ...string) (endpoints EndpointList, err error) {
 		return (count >= minErasureBlocks && count <= maxErasureBlocks && count%2 == 0)
 	}
 
-	// getValidErasureBlocks returns a string containing the valid values for erasure coding.
-	getValidErasureBlocks := func() string {
-		list := ""
-		for blocks := minErasureBlocks; blocks <= maxErasureBlocks; blocks += 2 {
-			list += fmt.Sprintf("%d", blocks)
-			if blocks < maxErasureBlocks-2 {
-				list += ", "
-			} else if blocks == maxErasureBlocks-2 {
-				list += " or "
-			}
-		}
-		return list
-	}
-
 	// Check whether no. of args are valid for XL distribution.
 	if !isValidDistribution(len(args)) {
-		return nil, fmt.Errorf("A total of %d endpoints were found. For erasure mode it should be %s", len(args), getValidErasureBlocks())
+		return nil, fmt.Errorf("A total of %d endpoints were found. For erasure mode it should be an even number between %d and %d", len(args), minErasureBlocks, maxErasureBlocks)
 	}
 
 	var endpointType EndpointType
--- a/cmd/endpoint.go
+++ b/cmd/endpoint.go
@@ -201,7 +201,7 @@ func NewEndpointList(args ...string) (endpoints EndpointList, err error) {
 
 	// Check whether no. of args are valid for XL distribution.
 	if !isValidDistribution(len(args)) {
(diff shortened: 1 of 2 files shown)


cmd/endpoint.go
```diff
@@ -182,12 +182,26 @@ func (endpoints EndpointList) SetHTTP() {
...
+				list += " or "
+			}
+		}
+		return list
 	}
 
 	// Check whether no. of args are valid for XL distribution.
 	if !isValidDistribution(len(args)) {
-		return nil, fmt.Errorf("total endpoints %d found. For XL/Distribute, it should be 4, 6, 8, 10, 12, 14 or 16", len(args))
+		return nil, fmt.Errorf("total endpoints %d found. For XL/Distribute, it should be %s", len(args), getValidErasureBlocks())
```
donatello: Error message could be better - e.g. `"A total of %d endpoints were found. For XL/distributed mode, it should be %s"`.

donatello (commented):
LGTM except error message could be better.

cmd/endpoint.go
harshavardhana: We can think about removing XL from logging .. to avoid confusion with our docs. We can just say `erasure` 

```A total of %d endpoints were found. For erasure mode it should be %d```

Comment (fwessels) [reply cmd/endpoint.go]: I like the last one the best, will change it to that.

**Review by donatello — APPROVED** [review]

File: cmd/endpoint.go [hunk 2]
```diff
@@ -182,12 +182,26 @@ func (endpoints EndpointList) SetHTTP() {
...
+		list := ""
+		for blocks := minErasureBlocks; blocks <= maxErasureBlocks; blocks += 2 {
+			list += fmt.Sprintf("%d", blocks)
+			if blocks < maxErasureBlocks-2 {
+				list += ", "
+			} else if blocks == maxErasureBlocks-2 {
+				list += " or "
+			}
+		}
+		return list
```
balamurugana: You could have `validErasureBlocks` slice and,  `isValidDistribution()` and error message can use it.
```go
validErasureBlocks := []int{}
for block := minErasureBlocks; block <= maxErasureBlocks; block += 2 {
  validErasureBlocks = append(validErasureBlocks, block)
}

isValidDistribution := func(count int) bool {
  for block := range validErasureBlocks {
    if block == count {
      return true
    }
  }
  return false
}

err = fmt.Errorf("A total of %d endpoints were found. For erasure mode it should be one of %s", len(args), validErasureBlocks)
```

balamurugana (changes requested):
Unit test needs to be fixed.

deekoder:
Do we have to add/modify any unit tests in erasure tests to support this PR?

cmd/endpoint.go
fwessels: This way it would be more expensive as the slice is always computed whereas now it is only done in the case of error.

balamurugana: As this function is called at startup only, its OK to be expensive.    If its also a concern, we could have `validErasureBlocks` along with `{min|max}ErasureBlocks`.

What I understood from this PR is that, `{min|max}ErasureBlocks` can be changed in later time and `NewEndpointList()` would work without a change.   To extend that I would like to move out `divisible by 2` logic from `NewEndpointList()`.

balamurugana:
> Do we have to add/modify any unit tests in erasure tests to support this PR?

This PR makes some unit test failure.  You can check travis/appveyor CI about those failure.

balamurugana (approved):

harshavardhana:
This PR is blocked until we implement a proper solution for issue #3883

PR #4163, opened 2017-04-22
bit-rot: Use faster hash functions for bit-rot detection

### LINKED ISSUE #3946: bit-rot: Use faster hash functions for bit-rot detection
Minio's erasure coding / bit-rot detection currently relies on BLAKE2b (and SHA-256 on arm).
It's not necessary to use a cryptographic secure general purpose hash function - weaker hash function properties are sufficient for minio's purposes.  

## Expected Behavior
The bit-rot detection requires following properties:
 - The hash must be collision resistant. (see: what means collision resistant)
 - The hash must be long enough - this is also related to collision resistance

### What means collision resistant
In this context collision resistant does not mean collision resistant from a cryptographers perspective.
The hash should not collide for special structured messages - e.g. zero blocks. For example a hash function consisting of modulo multiplication - like `hash(i) = (hash(i-1) * message) mod poly` produces always the zero value if a zero-block occurs, regardless what data comes after the zero-block.     

### Some math
A "good" decision requires some math:  
The birthday paradox tells us: 
 - the probability of a collision is >= 50 % if you hash `2^(hashsize / 2) different messages.
 
Or in oder words: you have to expect a collision after hashing 2^(hashsize / 2) different messages. 
But this only applies for an ideal hash function. If you use a hash function, that relies on simple mathematical operations (like addition or multiplication), a collision can occur because of some structure within the message. On way to be "sure", that a bit flip in the message results in a different hash value, is to relay on hash functions using the [avalanche effect](https://en.wikipedia.org/wiki/Avalanche_effect).

So if we rely on a hash function with the "avalanche property", we can be confident, that the hash function will give us a different values for slightly different messages  - otherwise we found a collision of the hash function. Further it should be very unlikely to get two different messages, which results in the same hash value - this is pretty similar to a cryptographic hash function.

Now let us assume, that we have found such a good hash function - how long should the hash value be?
The computation for minio is a bit different because of:
 - for minio a pure collision is not a problem
 - only a collision of a message and its bit-rot version is problem
 - minio is limited to 16 storage disks  

```
min_msg_size := minimal size of uploaded / stored files (e.g. 4KB)
bitrot rate br:= probability of a disk failure - depends on disks / mode of operation etc. 
collision rate cr := ~ 1 /  2 ^ (hashsize / 2) 

We can now compute the number of bytes - a collision statistically appears after:     
(br ^ -1) * (cr ^ -1) * min_msg_size = (br ^ -1) * (2 ^ (hashsize / 2)) * min_msg_size bytes
in the worst case - only min_msg_size long messages are stored
```
But notice, that this number of bytes refers to ***all minio instances over the world together - not just a single instances***
Let use some concrete numbers:
```
hash_size = 128 bit
min_msg_size = 1K
br = 0.01% (I've no clue whether this is realistic - probably not, realistic values are much lower e.g. < 0.00001%)

=> (br ^ -1) * (cr ^ -1) * min_msg_size = 1,717 * 10^14 TB =~ 163834 EB (exabyte)
if we assume we have 1 billion Minio instances out there, we get following bytes / instance: 
( (br ^ -1) * (cr ^ -1) * min_msg_size ) / 10000000000  =~ 171798,69 TB / instance
```
This is far beyond current store capabilities. 

## Possible Solution
On hash function candidate is [SipHash](https://131002.net/siphash/siphash.pdf) - It's fast on amd64, 386 and ARM - see [implementation](https://github.com/aead/siphash).

aead:
This change introduces the SipHash hash function as default bit-rot detection mechanism.

## Description
This change sets the default hash algorithm to SipHash. Therefore the siphash implementation is vendored (github.com/aead/siphash). Further the doc is updated and now says that Siphash is used instead of BLAKE2b.

## Motivation and Context
SipHash is much faster than BLAKE2b or SHA-2.
Fixes #3946 

## How Has This Been Tested?
run `go test -v` within /cmd

## Types of changes

## Checklist:


cmd/xl-v1-metadata.go
```diff
@@ -60,30 +59,20 @@ const (
 	sha256Algo = "sha256"
 	// Rest of the platforms default to blake2b.
```
harshavardhana: Comment needs to change.

cmd/erasure-utils_test.go
```diff
@@ -21,6 +21,8 @@ import (
 	"errors"
 	"io"
 	"testing"
+
+	"github.com/aead/siphash"
```
harshavardhana, line 25: I thought we are going to be pointing this to `golang.org/x/crypto`

cmd/xl-v1-metadata.go
```diff
@@ -60,30 +59,20 @@ const (
...
-		// blake2b is sub-optimal and slower on ARM64.
-		// This would also allows erasure coded writes
-		// on ARM64 servers to be on-par with their
-		// counter-part X86_64 servers.
-		return sha256Algo
-	default:
-		// Default for all other architectures we use blake2b.
-		return blake2bAlgo
-	}
+	return siphash128Algo
```
harshavardhana, line 77: Is sipHash better on arm64 as well ? If not we should keep sha256 for arm64.

harshavardhana (changes requested):

cmd/erasure-utils_test.go
aead: Me too, but I think Adam (agl) does not want to include SipHash (yet). I think the reason for this is that it is currently not widely used and the arguments about not including to many stuff into the /x repos - See [17244](https://github.com/golang/go/issues/17244)

cmd/xl-v1-metadata.go
aead: I think so, I cannot test it because I don't have an arm64 device (just an arm). But I would be really surprised if SHA-2 is faster on arm64.
Further there's currently no arm64 assembly version of SipHash.

cmd/erasure-utils_test.go
aead: But we can make siphash a Minio repository...

cmd/xl-v1-metadata.go
aead: Done

cmd/xl-v1-metadata.go
harshavardhana: Will test it.. can get you an ARM64 in meantime.

aead:
@harshavardhana The benchmark of your ARM64-machine:
sha256-simd
```
BenchmarkHash8Bytes-96    	 2000000	       701 ns/op	  11.40 MB/s
BenchmarkHash1K-96        	  500000	      3867 ns/op	 264.76 MB/s
BenchmarkHash8K-96        	   50000	     26545 ns/op	 308.61 MB/s
BenchmarkHash1M-96        	     500	   3308938 ns/op	 316.89 MB/s
```
siphash
```
BenchmarkWrite_8-96      	20000000	      87.8 ns/op	  91.17 MB/s
BenchmarkWrite_1K-96     	 1000000	      1677 ns/op	 610.43 MB/s
BenchmarkSum64_8-96      	10000000	       159 ns/op	  50.31 MB/s
BenchmarkSum64_1K-96     	 1000000	      1748 ns/op	 585.62 MB/s
BenchmarkSum128_8-96     	10000000	       196 ns/op	  40.63 MB/s
BenchmarkSum128_1K-96    	 1000000	      1786 ns/op	 573.31 MB/s
```
But remember: SipHash does not use optimized assembly, so it is probably possible to improve performance further.

harshavardhana:
> But we can make siphash a Minio repository...

This seems like a good idea as well. Let me talk to @abperiasamy first and we can move.

fwessels (commented):
Couple of questions/remarks (after going through the paper):
- "SipHash is optimized for short inputs" -- what does short mean, is it really short, kilobytes?
- We comment that SipHash is not meant to be, and (obviously) is not, collision resistant.
- We propose that hash tables switch to SipHash as a hash function.
- General-purpose cryptographic hash functions perform many extra computations for the goal of collision resistance on public inputs, while MACs have secret keys and do not need collision resistance

Note also that BLAKE2b is 512 bit vs 128 bit for SipHash.

It is obviously faster but I'm not sure if it is overall a better choice in our case.

aead:
@fwessels 
1. "Optimized for short inputs" means that SipHash is also fast for short inputs / messages.
SipHash was designed with hash-flooding attacks in mind. If you use SipHash as the hash function of a hashmap implementation you often hash short inputs (the hashmap key). E.g. lookup an int or a short string. For example BLAKE2b use a compression function that takes 128 bytes (the block size) while SipHash takes just 8 bytes. This results in a much cheaper state update.

2. There are two perspectives of "collision resistance". From a cryptographers perspective SipHash is (obviously) not collision resistant, because of the "to short hash size". As a cryptographer you want to "ensure" that an attacker must do at least 2^128 steps to break a crypto-primitive. This means that any hash function, that produces hash values < 256 bit is not (strong) collision resistant. 
As an example you can take a look at the BLAKE2b package - [RFC 7693](https://tools.ietf.org/html/rfc7693#section-4) specifies BLAKE2b-160 but it's not implemented because people may use it in cases where birthday attacks can be applied.
Further a cryptographer is worried about mathematical attacks. Therefore general purpose hash functions perform many extra computations to defend against mathematical attacks like differential analysis. A cryptographer will always say: SipHash is not collision resistant - but SipHash is not designed to be a general purpose cryptographic (collision resistant) hash function.

The use case of minio is different.  The bit-rot detection must only recognize: Okay, this message has changed, because the hash I computed after reading the msg from disk is different from the one I computed when I wrote the msg to disk. So there was a bit-flip / bit-rot.
Looks like a general purpose hash function is a good choice and that's right. But a general purpose hash function provides properties minio does not use/need.
For example an arbitrary collision is not a problem for the bit-rot detection. Only a collision of a message and a bit-flipped version of the message is a problem. The probability of such a collision can be computed (under the assumption SipHash exploits the avalanche effect /+ is collision resistant - not from a cryptographer's perspective) - I've tried to show this within the issue #3946.

As far as I know SipHash is collision resistant except for the hash size.
Maybe a comparison shows what I mean:
*Let's assume we use CRC-64. Using more or less simple math we can find collisions very quickly. That's because CRC-64 does not use design principles like confusion and diffusion but a mathematical approach - so CRC-64 is not collision resistant. For SipHash the task is much harder. Of course we can hash 2^32 messages (for SipHash-64) and we will find a collision with a probability about 50%. (birthday paradox) But we cannot (as far as I know) do a fancy mathematical trick to create two (slightly) different messages with the same SipHash hash. - That's meant by collision resistance*

Of course it's probably much easer to find such an attack against SipHash than against BLAKE2 (or SHA-2) because SipHash has a much simpler compression function with lesser security margin. 

> Note also that BLAKE2b is 512 bit vs 128 bit for SipHash.

Of course your right. BLAKE2b produce much longer hash values than SipHash. (Just take into account that BLAKE2b-256 would be secure enough anyway - as long as BLAKE2b is secure of course). That's a statistically decision - see the math within the issue.

fwessels:
One more thing came to mind, in order to verify whether bit-rot has occurred you need to know the key-phrase. So for an external tool to verify the hash you will have to input the key-phrase for SipHash (in contrast to SHA256/BLAKE2 etc where you can just do something like `openssl dgst -sha256`).

Also if anybody changes the key phrase in the source then GETs for previously stored objects will start to fail.

aead:
> One more thing came to mind, in order to verify whether bit-rot has occurred you need to know the key-phrase. So for an external tool to verify the hash you will have to input the key-phrase for SipHash (in contrast to SHA256/BLAKE2 etc where you can just do something like openssl dgst -sha256).
Also if anybody changes the key phrase in the source then GETs for previously stored objects will start to fail.

That's right. I just added the key because "minio is awesome" matches 16 bytes. We can also use a zero key or change the siphash API...
Or maybe we can find a way to use a different key per block / message...

Maybe a stupid question: Does minio provide a mechanism to protect the hash values from getting corrupted by bit-rot? Because if not, how can we be sure that a bit flip does not effect the (somewhere) stored hash (and the message)?

fwessels:
> Maybe a stupid question: Does minio provide a mechanism to protect the hash values from getting corrupted by bit-rot? Because if not, how can we be sure that a bit flip does not effect the (somewhere) stored hash (and the message)?

Good question, no, that is not the case.
The json message is not "signed" or anything, so a user (with access to the back-end) can change the json.
However, bit-rot happening in the hash itself as stored in the json file is also detected (unless the data blob would also have bit-rot leading to the same outcome as the bit-rotted hash in the json file but  let's no go this far...)

aead:
> However, bit-rot happening in the hash itself as stored in the json file is also detected (unless the data blob would also have bit-rot leading to the same outcome as the bit-rotted hash in the json file but let's no go this far...)

Ah okay - but than you actually never know whether the bit-rot happened in the hash itself or in the blob, right?! - whenever a bit-rot is detected of course. I guess disk failures occur somehow randomly distributed over the hole disk - so the probability that it affects the blob is significant higher... But this is (if it's an issue at all) a problem which cannot be solved by the hash function...

aead:
The current plan is to migrate to [GHASH](https://en.wikipedia.org/wiki/Galois/Counter_Mode) so I close this.

PR #4170, opened 2017-04-24
server: Validate path for bad components in a handler.

harshavardhana:
## Description
Validate path for bad components in a handler.

## How Has This Been Tested?
Manually and using `go test`

## Types of changes

## Checklist:

### DIFF
--- a/cmd/generic-handlers.go
+++ b/cmd/generic-handlers.go
@@ -419,3 +419,52 @@ func (h httpStatsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	// Update http statistics
 	globalHTTPStats.updateStats(r, ww, durationSecs)
 }
+
+// pathValidityHandler validates all the incoming paths for
+// any bad components and rejects them.
+type pathValidityHandler struct {
+	handler http.Handler
+}
+
+func setPathValidityHandler(h http.Handler) http.Handler {
+	return pathValidityHandler{handler: h}
+}
+
+// Bad path components to be rejected by the path validity handler.
+const (
+	dotdotComponent = ".."
+	dotComponent    = "."
+)
+
+// Check if the incoming path has bad path components,
+// such as ".." and "."
+func hasBadPathComponent(path string) bool {
+	path = strings.TrimSpace(path)
+	for _, p := range strings.Split(path, slashSeparator) {
+		switch strings.TrimSpace(p) {
+		case dotdotComponent:
+			return true
+		case dotComponent:
+			return true
+		}
+	}
+	return false
+}
+
+func (h pathValidityHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	// Check for bad components in URL path.
+	if hasBadPathComponent(r.URL.Path) {
+		writeErrorResponse(w, ErrInvalidResourceName, r.URL)
+		return
+	}
+	// Check for bad components in URL query values.
+	for _, vv := range r.URL.Query() {
+		for _, v := range vv {
+			if hasBadPathComponent(v) {
+				writeErrorResponse(w, ErrInvalidResourceName, r.URL)
+				return
+			}
+		}
+	}
+	h.handler.ServeHTTP(w, r)
+}
--- a/cmd/object-api-utils.go
+++ b/cmd/object-api-utils.go
@@ -131,6 +131,9 @@ func IsValidObjectName(object string) bool {
 // IsValidObjectPrefix verifies whether the prefix is a valid object name.
 // Its valid to have a empty prefix.
 func IsValidObjectPrefix(object string) bool {
+	if hasBadPathComponent(object) {
+		return false
+	}
 	if len(object) > 1024 {
 		return false
 	}
--- a/cmd/api-errors.go
+++ b/cmd/api-errors.go
@@ -139,6 +139,7 @@ const (
 	ErrObjectExistsAsDirectory
 	ErrPolicyNesting
 	ErrInvalidObjectName
+	ErrInvalidResourceName
 	ErrServerNotInitialized
 	// Add new extended error codes here.
 	// Please open a https://github.com/minio/minio/issues before adding
@@ -588,7 +589,12 @@ var errorCodeResponse = map[APIErrorCode]APIError{
 	},
 	ErrInvalidObjectName: {
 		Code:           "XMinioInvalidObjectName",
-		Description:    "Object name contains unsupported characters. Unsupported characters are `^*|\\\"",
+		Description:    "Object name contains unsupported characters.",
+		HTTPStatusCode: http.StatusBadRequest,
+	},
+	ErrInvalidResourceName: {
+		Code:           "XMinioInvalidResourceName",
+		Description:    "Resource name contains bad components such as \"..\" or \".\".",
 		HTTPStatusCode: http.StatusBadRequest,
 	},
 	ErrServerNotInitialized: {
--- a/cmd/object-api-utils_test.go
+++ b/cmd/object-api-utils_test.go
@@ -100,12 +100,22 @@ func TestIsValidObjectName(t *testing.T) {
 		{"contains-|-pipe", true},
 		{"contains-\"-quote", true},
 		{"contains-`-tick", true},
+		{"..test", true},
+		{".. test", true},
+		{". test", true},
+		{".test", true},
 		{"There are far too many object names, and far too few bucket names!", true},
 		// cases for which test should fail.
 		// passing invalid object names.
 		{"", false},
 		{"a/b/c/", false},
 		{"/a/b/c", false},
+		{"../../etc", false},
+		{"../../", false},
+		{"/../../etc", false},
+		{" ../etc", false},
+		{"./././", false},
+		{"./etc", false},
 		{"contains-\\-backslash", false},
 		{string([]byte{0xff, 0xfe, 0xfd}), false},
 	}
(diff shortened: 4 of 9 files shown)


cmd/object-api-utils.go
```diff
@@ -163,6 +163,29 @@ func pathJoin(elem ...string) string {
...
 
+// pathClean - is path.Clean() but retains trailing "/" and returns
+// empty string for empty strings not ".".
+func pathClean(p string) string {
+	dotdot := strings.HasPrefix(p, "..")
+	if dotdot {
+		p = slashSeparator + p
+	}
+	// Check path ends with separate, save it.
+	sp := hasSuffix(p, slashSeparator)
```
donatello: This line can be after the next if-block - it is not used until then.

cmd/object-api-utils.go
```diff
@@ -163,6 +163,29 @@ func pathJoin(elem ...string) string {
...
+	p = path.Clean(p)
+	if p == "." {
```
vadmeste: We are cleaning object names with pathClean() function, if we remove '.', it means we are not able to upload an object with name '.' though this is already supported in S3.

cmd/object-handlers.go
```diff
@@ -709,7 +709,7 @@ func (api objectAPIHandlers) CopyObjectPartHandler(w http.ResponseWriter, r *htt
 func (api objectAPIHandlers) PutObjectPartHandler(w http.ResponseWriter, r *http.Request) {
 	vars := mux.Vars(r)
 	bucket := vars["bucket"]
-	object := vars["object"]
+	object := pathClean(pathClean(vars["object"]))
```
donatello: Why twice?

cmd/object-handlers.go
```diff
@@ -833,7 +833,7 @@ func (api objectAPIHandlers) PutObjectPartHandler(w http.ResponseWriter, r *http
 func (api objectAPIHandlers) AbortMultipartUploadHandler(w http.ResponseWriter, r *http.Request) {
 	vars := mux.Vars(r)
 	bucket := vars["bucket"]
-	object := vars["object"]
+	object := pathClean(pathClean(vars["object"]))
```
donatello: Why twice? Here and multiple places below.

donatello (changes requested):

cmd/object-api-utils.go
harshavardhana: We can't support that already. @vadmeste on Linux "." means directory.

cmd/object-handlers.go
harshavardhana: oops string replace.

cmd/object-handlers.go
harshavardhana: Fixed.. string replace mistake.

cmd/object-api-utils.go
harshavardhana: No we need to path.Clean() removes the ending separator.

donatello: Ah ok.

cmd/web-handlers.go
```diff
@@ -485,7 +485,7 @@ func (web *webAPIHandlers) Upload(w http.ResponseWriter, r *http.Request) {
 
 	vars := mux.Vars(r)
 	bucket := vars["bucket"]
-	object := vars["object"]
+	object := pathClean(vars["object"])
```
vadmeste: Not sure why we don't also clean object in RemoveObject and listing etc..

vadmeste:
Overall, I think we should check for paths and clean them in only one place, and it seems to me it is safe to do that in posix primitives.

harshavardhana:
PR updated @vadmeste @donatello @krishnasrinivas

harshavardhana:
Adding tests..

cmd/object-api-utils.go
```diff
@@ -234,7 +267,7 @@ func isMinioReservedBucket(bucketName string) bool {
 	return bucketName == minioReservedBucket
 }
 
-// byBucketName is a collection satisfying sort.Interface.
+// byBucketName is a collection satisfying sort.Iterface.
```
donatello: Typo.

cmd/posix-prepare-path.go
```diff
@@ -17,25 +17,59 @@
...
+func filepathClean(path string) string {
+	dotdot := strings.HasPrefix(path, "../")
+	if dotdot {
+		// Make sure we add a separator to
+		// facilitate filepath.Clean behavior
+		// to kick in.
+		path = string(os.PathSeparator) + path
+	}
+
+	// Check path ends with separate, save it.
```
donatello: s/separate/separator/

cmd/posix-prepare-path.go
```diff
@@ -17,25 +17,59 @@
...
+	// Make sure we do this before filepath.Clean()
+	// filepath.Clean() removes trailing separator.
+	sp := hasSuffix(path, string(os.PathSeparator))
+
+	// Clean the incoming path for all relative paths.
+	path = filepath.Clean(path)
+	if path == "." {
+		// a path with "." has no meaning
+		// on object storage, treat is a empty.
```
donatello: The path "." has no meaning on object storage; treat as empty.

cmd/object-api-utils.go
```diff
@@ -163,6 +164,38 @@ func pathJoin(elem ...string) string {
...
+		// Make sure we add a separator to
+		// facilitate path.Clean behavior
+		// to kick in.
+		p = slashSeparator + p
+	}
+
+	// Check path ends with separator, save it.
+	// Make sure we do this before path.Clean()
+	// path.Clean() removes trailing separator.
+	sp := hasSuffix(p, slashSeparator)
```
donatello: We are using `hasSuffix` here and `strings.HasPrefix` above - when a `hasPrefix` that works for windows is available. In either case because there is nothing to lowercase there is no problem, but we can be consistent and use either the std lib funcs or our own windows compatible funcs.

donatello:
Tests broke because of the error message change.

harshavardhana:
> Tests broke because of the error message change.

Should be fixed..

donatello:
@harshavardhana Compile errors...

cmd/object-api-utils.go
harshavardhana: good idea.

cmd/generic-handlers.go
```diff
@@ -419,3 +419,52 @@ func (h httpStatsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+}
+
+func setPathValidityHandler(h http.Handler) http.Handler {
+	return pathValidityHandler{handler: h}
+}
+
+// Bad path components to be rejected by the path validity handler.
+const (
+	relPath = ".."
+	curPath = "."
```
krishnasrinivas: can be:
dotdotComponent = ".."
dotComponent = "."

cmd/generic-handlers.go
```diff
@@ -419,3 +419,52 @@ func (h httpStatsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+func hasBadPathComponent(path string) bool {
+	path = strings.TrimSpace(path)
+	for _, p := range strings.Split(path, slashSeparator) {
+		p = strings.TrimSpace(p)
+		if p == relPath {
+			return true
+		} else if p == curPath {
+			return true
+		}
+	}
```
krishnasrinivas: can be
```
		if p == relPath {
			return true
		}
                 if p == curPath {
			return true
		}
```

harshavardhana: why? @krishnasrinivas  it is more idiomatic to have if and else in this scenario.

cmd/generic-handlers.go
harshavardhana: Sure..

cmd/generic-handlers.go
harshavardhana: Changed to switch.

krishnasrinivas: `if` statements are easier on eyes than `if else if` no?

`idiomatic scenario` wise both if or if-else-if fit

harshavardhana:
Updated with tests and also all other fixes.. can be reviewed again @donatello @krishnasrinivas

cmd/posix.go
```diff
@@ -55,16 +56,17 @@ func checkPathLength(pathName string) error {
...
-	// Check each path segment length is > 255
-	for len(pathName) > 0 && pathName != "." && pathName != "/" {
-		dir, file := slashpath.Dir(pathName), slashpath.Base(pathName)
+	// Check each path component if they have "." or "..".
+	if hasBadPathComponent(pathName) {
+		return errFileComponentInvalid
+	}
 
-		if len(file) > 255 {
+	// Check each path segment length is > 255.
```
donatello: `Check each path segment length is <= 255`

krishnasrinivas (approved):

cmd/posix.go
```diff
@@ -55,16 +56,17 @@ func checkPathLength(pathName string) error {
 	// Convert any '\' to '/'.
```
krishnasrinivas: checkPathLength can be renamed to isPathValid because it's doing more than path length check

harshavardhana: Renamed as checkPathValidity.

donatello (approved):

krishnasrinivas:
xml error messge says:
```
Resource name contains bad components such as &#34;..&#34; or &#34;.&#34;.
```

is that expected?

harshavardhana:
> Resource name contains bad components such as &#34;..&#34; or &#34;.&#34;.

Yes you need unmarshal it, on the wire that is the correct value. @krishnasrinivas see the listobjects ETag output.

donatello:
LGTM.

Issue #4160, opened 2017-04-21
[tests] Investigate TestSuiteCommon.TestGetPartialObjectMisAligned rare crash


vadmeste:
**This crash is found while compiling this PR https://github.com/minio/minio/pull/4150, However, commits which triggered this crash disappeared now.**

## Expected Behavior
Successful go test.

## Current Behavior
Test crashed.

```
Running all coverage for minio: ok  	github.com/minio/minio	0.009s	coverage: 66.7% of statements
?   	github.com/minio/minio/browser	[no test files]
----------------------------------------------------------------------
PANIC: server_test.go:1821: TestSuiteCommon.TestGetPartialObjectMisAligned
... Panic: runtime error: slice bounds out of range (PC=0x4621BC)
/home/travis/.gimme/versions/go1.7.5.linux.amd64/src/runtime/panic.go:458
  in gopanic
/home/travis/.gimme/versions/go1.7.5.linux.amd64/src/runtime/panic.go:34
  in panicslice
server_test.go:1875
  in TestSuiteCommon.TestGetPartialObjectMisAligned
/home/travis/.gimme/versions/go1.7.5.linux.amd64/src/reflect/value.go:302
  in Value.Call
/home/travis/.gimme/versions/go1.7.5.linux.amd64/src/runtime/asm_amd64.s:2086
  in goexit
OOPS: 200 passed, 1 PANICKED
--- FAIL: Test (13.62s)
FAIL
coverage: 74.9% of statements
FAIL	github.com/minio/minio/cmd	82.608s
make: *** [coverage] Error 1
```

## Possible Solution
No idea

## Context
Bug found

## Your Environment
All

deekoder:
@hackintoshrao - please take a look at this.

hackintoshrao:
- It looks like the issue has originated from [this line](https://github.com/vadmeste/minio/blob/issues/4138/cmd/server_test.go#L1875) where the contents of the slice has been accessed. 
- The size of the string slice depends on the output of the random function set in the loop (and inside the loop too) [here](https://github.com/vadmeste/minio/blob/issues/4138/cmd/server_test.go#L1849). 
- In cases where the random number generators used returns small numbers (0,1), there is the possibility wherein the slice would have very few elements in it and this will result in out of bound error. 
- We might have to guarantee the presence of 13 bytes of characters atleast in the slice.

harshavardhana:
[...]

PR #4175, opened 2017-04-25
Fix rare test crash.

hackintoshrao:
## Description

- Fix rare crash of TestSuiteCommon.TestGetPartialObjectMisAligned
- Since the size of the buffer used depends on the random number
  generated the guarantee of 13 bytes of data is not assured.
- This is fixed by initially writing 13 bytes of data into the buffer.
## Motivation and Context

Fixing https://github.com/minio/minio/issues/4160
## How Has This Been Tested?

`make test`
## Types of changes

## Checklist:

### DIFF
--- a/cmd/server_test.go
+++ b/cmd/server_test.go
@@ -1833,25 +1833,23 @@ func (s *TestSuiteCommon) TestGetPartialObjectMisAligned(c *C) {
 	c.Assert(response.StatusCode, Equals, http.StatusOK)
 
 	var buffer bytes.Buffer
-	line := `1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,
-	1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,
-	1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,
-	1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,
-	1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,
-	1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,
-	1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,
-	1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,
-	1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,
-	1234567890,1234567890,1234567890,123`
-
-	rand.Seed(UTCNow().UnixNano())
-	// Create a misalgined data.
-	for i := 0; i < 13*rand.Intn(1<<16); i++ {
-		buffer.WriteString(fmt.Sprintf("[%05d] %s\n", i, line[:rand.Intn(1<<8)]))
+	// data to be written into buffer.
+	data := "1234567890"
+	// seed the random number generator once.
+	rand.Seed(3)
+	// generate a random number between 13 and 200.
+	randInt := getRandomRange(13, 200, -1)
+	// write into buffer till length of the buffer is greater than the generated random number.
+	for i := 0; i <= randInt; i += 10 {
+		buffer.WriteString(data)
 	}
-	putContent := buffer.String()
-	buf := bytes.NewReader([]byte(putContent))
-
+	// String content which is used for put object range test.
+	putBytes := buffer.Bytes()
+	putBytes = putBytes[:randInt]
+	// randomize the order of bytes in the byte array and create a reader.
+	putBytes = randomizeBytes(putBytes, -1)
+	buf := bytes.NewReader(putBytes)
+	putContent := string(putBytes)
 	objectName := "test-big-file"
 	// HTTP request to upload the object.
 	request, err = newTestSignedRequest("PUT", getPutObjectURL(s.endPoint, bucketName, objectName),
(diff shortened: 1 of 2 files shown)


harshavardhana (approved):

cmd/server_test.go
```diff
@@ -1843,6 +1843,8 @@ func (s *TestSuiteCommon) TestGetPartialObjectMisAligned(c *C) {
 	1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,
 	1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,1234567890,
 	1234567890,1234567890,1234567890,123`
+	// ensuring that the buffer contains atleast 13 bytes of data.
+	buffer.WriteString(line[:13])
```
donatello: This function is incredibly complex for what is being done here. The requirement is just to check that the requested range of an object returns the correct range of data.

The approach used to generate `putContent` (the string that is ultimately uploaded and parts of which are retrieved) is just obfuscated and has no method to it - this is why there is a weird crash and a weird special patch to handle it.

Here is a simple way to generate `putContent` that has better random properties:

1. Select a random number between 12 and (say) 100 randomly. Call it `n`. This is the length of the string.
2. Append "123456890" to an empty string slice `s` repeatedly until its length exceeds `n`. Then set `s = s[:n]`.
3. Run the [Knuth shuffle (also called Fisher-Yates shuffle)](https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle) on `s`. The shuffle algo can be a simple utility function available for reuse in other places too.

That's it - use `s` as `putContent`.

As it stands the current code is unreadable and unmaintainable.

donatello (changes requested):
Requesting changes to use a better design. A possible design is described above.

cmd/server_test.go
harshavardhana: Part of erasure index distribution which we used to have can be used here

```
// randInts - uses Knuth Fisher-Yates shuffle algorithm for generating uniform shuffling.
func randInts(count int) []int {
        rand.Seed(time.Now().UTC().UnixNano()) // Seed with current time.
        ints := make([]int, count)
        for i := 0; i < count; i++ {
                ints[i] = i + 1
        }
        for i := 0; i < count; i++ {
                // Choose index uniformly in [i, count-1]
                r := i + rand.Intn(count-i)
                ints[r], ints[i] = ints[i], ints[r]
        }
        return ints
}
```

donatello: [Here](https://play.golang.org/p/wZxs9e1Auw) is an implementation that provides repeatable in-place randomization for byte-arrays.

The repeatability is useful because it ensures that in every run the same sequence of random values are generated - without having repeatable random numbers, if a test finds a bug, it will not be easily reproducible because in the next run the random numbers would change and the bug may not manifest.

donatello: Repeatability comes from setting a seed value. Each different seed value, causes a different random sequence, but the keeping the seed constant causes the same sequence of random numbers to be generated.

harshavardhana:
@hackintoshrao have you looked @donatello comments?

hackintoshrao:
Made the requested changes.

harshavardhana (dismissed):

cmd/test-utils_test.go
```diff
@@ -2361,3 +2361,24 @@ func getEndpointsLocalAddr(endpoints EndpointList) string {
...
 	return globalMinioHost + ":" + globalMinioPort
 }
+
+// fetches a random number between range min-maz
+func getRandomRange(min, max int) int {
+	rand.Seed(time.Now().Unix())
+	return rand.Intn(max-min) + min
+}
+
+// randomizes the order of bytes in the byte array.
```
donatello: Add a comment attributing the algorithm "Knuth Fisher-Yates shuffle" - so its clear where it comes from.

donatello (changes requested):

cmd/test-utils_test.go
hackintoshrao: Done.

harshavardhana (dismissed):

cmd/test-utils_test.go
```diff
@@ -2361,3 +2361,25 @@ func getEndpointsLocalAddr(endpoints EndpointList) string {
...
+// usign Knuth Fisher-Yates shuffle algorithm.
```
donatello: Fix spelling.

cmd/test-utils_test.go
donatello: Let's take some care in writing comments - capitalize "randomizes", replace period with comma at the end here. The sentence continues on the next line.

cmd/test-utils_test.go
```diff
@@ -2361,3 +2361,25 @@ func getEndpointsLocalAddr(endpoints EndpointList) string {
 
 	return globalMinioHost + ":" + globalMinioPort
 }
+
+// fetches a random number between range min-maz
```
donatello: maz?

cmd/test-utils_test.go
```diff
@@ -2361,3 +2361,25 @@ func getEndpointsLocalAddr(endpoints EndpointList) string {
...
+func getRandomRange(min, max int) int {
+	rand.Seed(time.Now().Unix())
```
donatello: Looks like my earlier point about seeding the RNG was not understood. If you pick a seed like this, the value of this test is low - the seed changes based on time and each time this code runs it will pick a new seed. This means that each time the generated random number is different and even if there is a bug in the code that you are testing, and it is found via this test, it will not be reproducible, because in the next run the test failure may disappear.

Using random numbers in testing should not be taken lightly. If you must use a randomization based approach, it needs to be done properly. Here the seed value needs to be picked in the calling code only once - that sets the sequence of random numbers that will be generated in all further randomization steps. So it would work like:

1. Seed RNG.
2. Select length of string randomly (**without resetting the seed!**)
3. Randomly permute the byte-array (**without resetting the seed!**)
4. Verify results of the code being tested.

donatello (changes requested):

donatello (approved):
👍

balamurugana (approved):
Verified as per @donatello

harshavardhana (approved):

PR #4179, opened 2017-04-26
web: Simplify and converge common functions in web/obj API.

harshavardhana:
## Description
RemoveObject() in webAPI currently re-implements some part
of the code to remove objects combine them for simplicity
and code convergence.

## Motivation and Context
web: Simplify and converge common functions in web/obj API.

## How Has This Been Tested?
Manually.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/object-handlers-common.go
+++ b/cmd/object-handlers-common.go
@@ -17,6 +17,7 @@
 package cmd
 
 import (
+	"net"
 	"net/http"
 	"strings"
 	"time"
@@ -223,4 +224,37 @@ func canonicalizeETag(etag string) string {
 // are equal, false otherwise
 func isETagEqual(left, right string) bool {
 	return canonicalizeETag(left) == canonicalizeETag(right)
+}
+
+// deleteObject is a convenient wrapper to delete an object, this
+// is a common function to be called from object handlers and
+// web handlers.
+func deleteObject(obj ObjectLayer, bucket, object string, r *http.Request) (err error) {
+	// Acquire a write lock before deleting the object.
+	objectLock := globalNSMutex.NewNSLock(bucket, object)
+	objectLock.Lock()
+	defer objectLock.Unlock()
+
+	// Proceed to delete the object.
+	if err = obj.DeleteObject(bucket, object); err != nil {
+		return err
+	}
+
+	// Get host and port from Request.RemoteAddr.
+	host, port, _ := net.SplitHostPort(r.RemoteAddr)
+
+	// Notify object deleted event.
+	eventNotify(eventData{
+		Type:   ObjectRemovedDelete,
+		Bucket: bucket,
+		ObjInfo: ObjectInfo{
+			Name: object,
+		},
+		ReqParams: extractReqParams(r),
+		UserAgent: r.UserAgent(),
+		Host:      host,
+		Port:      port,
+	})
+
+	return nil
 }
--- a/cmd/web-handlers.go
+++ b/cmd/web-handlers.go
@@ -255,16 +255,22 @@ func (web *webAPIHandlers) ListObjects(r *http.Request, args *ListObjectsArgs, r
 	return nil
 }
 
-// RemoveObjectArgs - args to remove an object
-// JSON will look like:
-// '{"bucketname":"testbucket","objects":["photos/hawaii/","photos/maldives/","photos/sanjose.jpg"]}'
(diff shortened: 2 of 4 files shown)


krishnasrinivas (approved):

cmd/object-handlers-common.go
```diff
@@ -224,3 +225,38 @@ func canonicalizeETag(etag string) string {
...
+// DeleteObject is a convenient wrapper to delete an object, this
+// is a common function to be called from object handlers and
+// web handlers.
+func deleteObject(api ObjectLayer, bucket, object string, r *http.Request) (err error) {
+	// Acquire a write lock before deleting the object.
+	objectLock := globalNSMutex.NewNSLock(bucket, object)
+	objectLock.Lock()
+	defer objectLock.Unlock()
+
+	// Stacke purposefully after Unlock(), to use defer's execution
```
krisis: typo: Stacked?

cmd/object-handlers-common.go
```diff
@@ -224,3 +225,38 @@ func canonicalizeETag(etag string) string {
 func isETagEqual(left, right string) bool {
 	return canonicalizeETag(left) == canonicalizeETag(right)
 }
+
+// DeleteObject is a convenient wrapper to delete an object, this
```
krisis: s/DeleteObject/deleteObject/

cmd/object-handlers-common.go
```diff
@@ -224,3 +225,38 @@ func canonicalizeETag(etag string) string {
...
+func deleteObject(api ObjectLayer, bucket, object string, r *http.Request) (err error) {
+	// Acquire a write lock before deleting the object.
+	objectLock := globalNSMutex.NewNSLock(bucket, object)
+	objectLock.Lock()
+	defer objectLock.Unlock()
+
+	// Stacke purposefully after Unlock(), to use defer's execution
+	// order (Last in first out).
+	defer func() {
+		if err == nil {
```
krisis: we can simply write the body of the function after `err = api.DeleteObject(...)` instead of using `defer` and having to explain the evaluation order.

cmd/object-handlers-common.go
```diff
@@ -224,3 +225,38 @@ func canonicalizeETag(etag string) string {
 func isETagEqual(left, right string) bool {
 	return canonicalizeETag(left) == canonicalizeETag(right)
 }
+
+// DeleteObject is a convenient wrapper to delete an object, this
+// is a common function to be called from object handlers and
+// web handlers.
+func deleteObject(api ObjectLayer, bucket, object string, r *http.Request) (err error) {
```
krisis: How about name it `obj` instead of `api` for a parameter of type `ObjectLayer`? `api` seems out of place.

cmd/web-handlers.go
```diff
@@ -972,6 +962,12 @@ func toWebAPIError(err error) APIError {
 			HTTPStatusCode: http.StatusForbidden,
 			Description:    err.Error(),
 		}
+	} else if err == errInvalidArgument {
```
krisis, line 965: Why is only `errInvalidArgument` being checked?

krisis (changes requested):

cmd/web-handlers.go
harshavardhana: Its a new addition. we don't need to add all, just add what is needed as needed.

krisis: @harshavardhana could you add a comment on when this error will be seen? Could you provide a test to verify that we receive `errInvalidArgument` as expected?

cmd/object-handlers-common.go
harshavardhana: Done

cmd/object-handlers-common.go
harshavardhana: Removed.

cmd/object-handlers-common.go
harshavardhana: Done

cmd/object-handlers-common.go
harshavardhana: done

cmd/web-handlers.go
harshavardhana: Already added @krisis - look at the web-handlers_test.go

krisis (approved):

PR #4195, opened 2017-04-28
fs: Migrate object metadata to objects directory.

### LINKED ISSUE #3352: backend: Namespace conflict with policy.json and actual object called policy.json
`mc policy both myminio/test` fails, refer to steps-to-repro for details.

## Expected Behavior
`mc policy both myminio/test` should work

```
krishna@escape:~$ mc mb myminio/test
Bucket created successfully ‘myminio/test’.
krishna@escape:~$ mc cp /etc/passwd myminio/test/policy.json
/etc/passwd:    2.20 KB / 2.20 KB ┃▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓┃ 100.00% 340.99 KB/s 0skrishna@escape:~$ mc policy both myminio/test
mc: <ERROR> Unable to set policy ‘both’ for ‘myminio/test’. 500 Internal Server Error
krishna@escape:~$ 
```

minio server logs:
```
ERRO[0018] Unable to set policy for the bucket test      cause=rename /home/krishna/export/.minio.sys/tmp/474f105d-7067-4e18-9de5-dd35caf250ac /home/krishna/export/.minio.sys/buckets/test/policy.json: is a directory source=[bucket-policy.go:211:writeBucketPolicy()] stack=fs-v1.go:460:fsObjects.PutObject <autogenerated>:195:(*fsObjects).PutObject bucket-policy.go:210:writeBucketPolicy bucket-policy-handlers.go:218:persistAndNotifyBucketPolicyChange bucket-policy-handlers.go:185:objectAPIHandlers.PutBucketPolicyHandler api-router.go:79:(objectAPIHandlers).PutBucketPolicyHandler-fm

```

harshavardhana:
## Description
Migrate object metadata to objects directory.

## Motivation and Context
Fixes #3352

## How Has This Been Tested?
Manually.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/format-config-v1.go
+++ b/cmd/format-config-v1.go
@@ -20,8 +20,12 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
+	"io/ioutil"
 	"reflect"
 	"sync"
+
+	"github.com/minio/minio/pkg/lock"
 )
 
 // fsFormat - structure holding 'fs' format.
@@ -45,6 +49,43 @@ type formatConfigV1 struct {
 	Format string    `json:"format"`
 	FS     *fsFormat `json:"fs,omitempty"` // FS field holds fs format.
 	XL     *xlFormat `json:"xl,omitempty"` // XL field holds xl format.
+}
+
+func (f *formatConfigV1) WriteTo(lk *lock.LockedFile) (n int64, err error) {
+	// Serialize to prepare to write to disk.
+	var fbytes []byte
+	fbytes, err = json.Marshal(f)
+	if err != nil {
+		return 0, traceError(err)
+	}
+	if err = lk.Truncate(0); err != nil {
+		return 0, traceError(err)
+	}
+	_, err = lk.Write(fbytes)
+	if err != nil {
+		return 0, traceError(err)
+	}
+	return int64(len(fbytes)), nil
+}
+
+func (f *formatConfigV1) ReadFrom(lk *lock.LockedFile) (n int64, err error) {
+	var fbytes []byte
+	fi, err := lk.Stat()
+	if err != nil {
+		return 0, traceError(err)
+	}
+	fbytes, err = ioutil.ReadAll(io.NewSectionReader(lk, 0, fi.Size()))
+	if err != nil {
+		return 0, traceError(err)
+	}
+	if len(fbytes) == 0 {
+		return 0, traceError(io.EOF)
+	}
+	// Decode `format.json`.
+	if err = json.Unmarshal(fbytes, f); err != nil {
+		return 0, traceError(err)
+	}
+	return int64(len(fbytes)), nil
 }
 
 /*
--- a/cmd/fs-v1-helpers.go
+++ b/cmd/fs-v1-helpers.go
@@ -230,7 +230,7 @@ func fsOpenFile(readPath string, offset int64) (io.ReadCloser, int64, error) {
 
 // Creates a file and copies data from incoming reader. Staging buffer is used by io.CopyBuffer.
 func fsCreateFile(filePath string, reader io.Reader, buf []byte, fallocSize int64) (int64, error) {
-	if filePath == "" || reader == nil || buf == nil {
+	if filePath == "" || reader == nil {
 		return 0, traceError(errInvalidArgument)
 	}
 
@@ -263,11 +263,18 @@ func fsCreateFile(filePath string, reader io.Reader, buf []byte, fallocSize int6
 		}
 	}
 
-	bytesWritten, err := io.CopyBuffer(writer, reader, buf)
-	if err != nil {
-		return 0, traceError(err)
+	var bytesWritten int64
+	if buf != nil {
+		bytesWritten, err = io.CopyBuffer(writer, reader, buf)
+		if err != nil {
+			return 0, traceError(err)
+		}
+	} else {
+		bytesWritten, err = io.Copy(writer, reader)
+		if err != nil {
+			return 0, traceError(err)
+		}
 	}
-
 	return bytesWritten, nil
 }
 
@@ -275,6 +282,12 @@ func fsCreateFile(filePath string, reader io.Reader, buf []byte, fallocSize int6
 func fsRemoveUploadIDPath(basePath, uploadIDPath string) error {
 	if basePath == "" || uploadIDPath == "" {
 		return traceError(errInvalidArgument)
+	}
+	if err := checkPathLength(basePath); err != nil {
+		return traceError(err)
+	}
+	if err := checkPathLength(uploadIDPath); err != nil {
+		return traceError(err)
 	}
 
 	// List all the entries in uploadID.
@@ -319,6 +332,26 @@ func fsFAllocate(fd int, offset int64, len int64) (err error) {
 // Renames source path to destination path, creates all the
 // missing parents if they don't exist.
 func fsRenameFile(sourcePath, destPath string) error {
+	if err := checkPathLength(sourcePath); err != nil {
+		return traceError(err)
+	}
+	if err := checkPathLength(destPath); err != nil {
+		return traceError(err)
+	}
+	// Verify if source path exists.
+	if _, err := os.Stat(preparePath(sourcePath)); err != nil {
+		if os.IsNotExist(err) {
+			return traceError(errFileNotFound)
+		} else if os.IsPermission(err) {
+			return traceError(errFileAccessDenied)
+		} else if isSysErrPathNotFound(err) {
+			return traceError(errFileNotFound)
+		} else if isSysErrNotDir(err) {
+			// File path cannot be verified since one of the parents is a file.
+			return traceError(errFileAccessDenied)
+		}
+		return traceError(err)
+	}
 	if err := mkdirAll(pathutil.Dir(destPath), 0777); err != nil {
 		return traceError(err)
 	}
--- a/cmd/fs-v1-metadata.go
+++ b/cmd/fs-v1-metadata.go
@@ -18,6 +18,8 @@ package cmd
 
 import (
 	"encoding/json"
+	"errors"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
@@ -25,6 +27,7 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/minio/minio-go/pkg/set"
 	"github.com/minio/minio/pkg/lock"
 	"github.com/minio/minio/pkg/mimedb"
 	"github.com/tidwall/gjson"
@@ -225,7 +228,18 @@ const (
 	// FS backend meta format.
 	fsMetaFormat = "fs"
 
+	// FS backend format version.
+	fsFormatVersion = fsFormatV2
+
 	// Add more constants here.
+)
+
+// FS format version strings.
+const (
+	fsFormatV1 = "1" // Previous format.
+	fsFormatV2 = "2" // Current format.
+	// Proceed to add "3" when we
+	// change the backend format in future.
 )
 
 // newFSMetaV1 - initializes new fsMetaV1.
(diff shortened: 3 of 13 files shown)


harshavardhana:
Will be adding tests. But ready for review.

harshavardhana:
There is a windows issue i am still debugging.. rest of the code can be reviewed @vadmeste @krishnasrinivas

cmd/fs-v1.go
```diff
@@ -72,15 +72,83 @@ func initMetaVolumeFS(fsPath, fsUUID string) error {
...
+// Attempts to migrate old object metadata files to newer format
+//
+// i.e
+//    -------------------------------------------------------
+//    .minio.sys/buckets/<bucket_name>/<object_path>/fs.json
+//    -------------------------------------------------------
+//    .minio.sys/buckets/<bucket_name>/objects/<object_path>/fs.json
+//    -------------------------------------------------------
+//
+func migrateFSObjectMetadataDir(fsPath, fsUUID string) error {
```
vadmeste: I can imagine that if a user has too many objects, he will press control-c in the middle of upgrade and it will be a problem. A better approach (unfortunately it also has its drawbacks) is to copy fs.json and not rename them. Once everything is finished, we upgrade fs format to v2.

cmd/fs-v1.go
```diff
@@ -110,17 +178,25 @@ func newFSObjectLayer(fsPath string) (ObjectLayer, error) {
...
 	}
 
 	// If the `format.json` doesn't exist create one.
-	if err == errUnformattedDisk {
+	if err == errUnformattedDisk || err == errFSFormatOld {
 		fsFormatPath := pathJoin(fsPath, minioMetaBucket, fsFormatJSONFile)
+
 		// Initialize format.json, if already exists overwrite it.
-		if serr := saveFormatFS(fsFormatPath, newFSFormatV1()); serr != nil {
+		if serr := saveFormatFS(fsFormatPath, newFSFormatV2()); serr != nil {
```
vadmeste: saveFormatFS() should be called after migrateFSObjectMetadataDir() and when this latter is successful.

harshavardhana: The reason to do that is if migrate fails we don't need to attempt again.

harshavardhana: We can then help the user manually .. migrate

cmd/fs-v1.go
harshavardhana: Yes we can even block control-c but .. I think it would make sense to print a message . That would make it easier.

harshavardhana: Copying each fs.json can take ages. We don't need to do that. Renames are atomic in nature so either it's do or die.. if we print a message about migration then all of this works out nicely.

cmd/fs-v1.go
harshavardhana: I know it is cleaner to do later. But that was the only reason as I stated above.

cmd/fs-v1.go
vadmeste: > Copying each fs.json can take ages.

Yeah, that's the drawback.

harshavardhana:
Also brought in locking to avoid multiple migrations running simultaneously in shared backend mode.

harshavardhana:
Windows build has been fixed as well.  Can I get more reviews ?

cmd/fs-v1.go
```diff
@@ -72,15 +72,131 @@ func initMetaVolumeFS(fsPath, fsUUID string) error {
 
 }
 
+// List all buckets at meta bucket prefix in `.minio.sys/buckets/` path.
+func readDirBuckets(metaBucket string) ([]string, error) {
```
vadmeste: I think you can move this function to fs-v1-helpers.go, this function is already generic.

cmd/fs-v1.go
```diff
@@ -110,17 +226,25 @@ func newFSObjectLayer(fsPath string) (ObjectLayer, error) {
...
+	if err != nil && err != errUnformattedDisk && err != errFSFormatOld {
 		return nil, fmt.Errorf("Unable to load 'format.json', %s", err)
 	}
 
 	// If the `format.json` doesn't exist create one.
-	if err == errUnformattedDisk {
+	if err == errUnformattedDisk || err == errFSFormatOld {
+		// Disk is in old format migrate object metadata.
+		if err == errFSFormatOld {
+			if merr := migrateFSObjectMetadataDir(fsPath, fsUUID); merr != nil {
```
vadmeste: Can we do migrate with the following style?
```
migrateFSObject()
|___ migrateFSObjectV1toV2()
```

That way we will always remember users that have real old installation.

cmd/fs-v1-metadata.go
```diff
@@ -269,6 +284,12 @@ func loadFormatFS(fsPath string) (*formatConfigV1, error) {
 		return nil, err
 	}
 
+	// Check if format is V2.
+	if !isFSFormatV2(format) {
```
vadmeste: I think it is important if we return two errors when we check for FS format version:
- Old version so we can migrate
- Newer unsupported version (like, 3, 4, ..), so we stop here since things start to be unpredictable.

cmd/fs-v1.go
```diff
@@ -72,15 +72,131 @@ func initMetaVolumeFS(fsPath, fsUUID string) error {
...
+//    .minio.sys/buckets/<bucket_name>/objects/<object_path>/fs.json
+//    -------------------------------------------------------
+//
+func migrateFSObjectMetadataDir(fsPath, fsUUID string) (err error) {
+	metaBucket := pathJoin(fsPath, minioMetaBucket, bucketConfigPrefix)
+	migratePlaceHolder := "." + fsUUID
+
+	// Locked for writing ".migrate" file under the buckets folder.
+	var lk *lock.LockedFile
+	lk, err = lock.LockedOpenFile(preparePath(pathJoin(metaBucket, migratePlaceHolder)), os.O_CREATE|os.O_WRONLY, 0600)
```
vadmeste: This line can fail if `.minio.sys/buckets` doesn't exist yet. In that case, user can't migrate to format V2

harshavardhana: This line already fails.. @vadmeste

cmd/fs-v1-metadata.go
harshavardhana: Good idea.

cmd/fs-v1.go
harshavardhana: Sure

cmd/fs-v1.go
harshavardhana: Good idea.

harshavardhana:
Changes have been done @vadmeste @krishnasrinivas .. Can you review and test again?

cmd/fs-v1-metadata.go
```diff
@@ -237,17 +238,45 @@ func newFSMetaV1() (fsMeta fsMetaV1) {
...
+func isFSValidFormat(formatCfg *formatConfigV1) bool {
+	if formatCfg == nil {
+		return false
+	}
+	if formatCfg.FS == nil {
+		return false
+	}
+
+	// Check for supported format versions.
+	for _, version := range supportedFormatVersions {
```
vadmeste: I think it is easier to check if the found version is <= 2 than doing this style

and 2 could be a global variable.

cmd/fs-v1.go
```diff
@@ -72,15 +73,124 @@ func initMetaVolumeFS(fsPath, fsUUID string) error {
...
+	defer func() {
+		if err == nil {
+			log.Println("SUCCESS.")
+		} else {
+			log.Printf("FAILED with %s.\n", err)
+		}
+	}()
+
+	var buckets []string
+	buckets, err = fsReaddirBuckets(fsPath)
```
vadmeste: Shouldn't we check for buckets inside '.minio.sys/buckets', rather than fsPath ? otherwise we will fail later in `fsRenameFile(pathJoin(metaBucket, bucket), pathJoin(metaBucket, tmpBucket))`

cmd/fs-v1-metadata.go
harshavardhana: The reason to do this is quite different.. it is to indicate if we are looking at the right versions supported by Minio.

cmd/fs-v1.go
harshavardhana: We are constructing it internally @vadmeste

cmd/fs-v1-metadata.go
harshavardhana: The lexical ordering is lost if we do `<=` it is safer to not do that and do direct comparison.

cmd/fs-v1.go
vadmeste: > We are constructing it internally @vadmeste

You mean we are going to create all buckets inside .minio.sys/buckets?

Anyway, it will fail later as I said in fsRenameFile(), I just tested it. pathJoin(metaBucket, bucket) doesn't always exist and it returns an error at that stage.

cmd/fs-v1.go
```diff
@@ -110,17 +219,25 @@ func newFSObjectLayer(fsPath string) (ObjectLayer, error) {
...
+	if err != nil && err != errUnformattedDisk && err != errFSFormatOld {
 		return nil, fmt.Errorf("Unable to load 'format.json', %s", err)
 	}
 
 	// If the `format.json` doesn't exist create one.
-	if err == errUnformattedDisk {
+	if err == errUnformattedDisk || err == errFSFormatOld {
+		// Disk is in old format migrate object metadata.
+		if err == errFSFormatOld {
+			if merr := migrateFSObject(fsPath, fsUUID); merr != nil {
```
krishnasrinivas: fsUUID is not shared with other minio processes, hence you could flock on format.json itself during the entire process

harshavardhana: Ah yes good catch, should have used that way.. instead.

harshavardhana: Fixed.

cmd/fs-v1-helpers.go
```diff
@@ -424,3 +457,18 @@ func fsRemoveMeta(basePath, deletePath, tmpDir string) error {
 	}
 	return fsDeleteFile(basePath, deletePath)
 }
+
+// List all buckets at meta bucket prefix in `.minio.sys/buckets/` path.
+// This is implemented to avoid a bug on windows with using readDir().
+func fsReaddirBuckets(fsPath string) ([]string, error) {
```
krishnasrinivas: Note that this function gets used only for migration. Not during regular FS operations. fsReaddirBuckets reads `fsPath/.minio.sys/buckets` not `fsPath` hence fsReaddirMetaBuckets sounds more appropriate.

Also better move this closer to `migrateFSObject()`

cmd/fs-v1-metadata.go
```diff
@@ -269,6 +309,17 @@ func loadFormatFS(fsPath string) (*formatConfigV1, error) {
...
 
+	// Check if format is currently supported.
+	if !isFSValidFormat(format) {
+		return nil, errCorruptedFormat
+	}
+
+	// Check for format version is current.
+	if format.FS.Version != fsFormatVersion {
+		return nil, errFSFormatOld
+	}
```
krishnasrinivas: Better do these checks version in the caller `loadFormatFS()` because as a code reader I expected loadFormatFS to just load it and not do validation of versions.

cmd/fs-v1.go
```diff
@@ -72,15 +72,100 @@ func initMetaVolumeFS(fsPath, fsUUID string) error {
...
+	log.Println("PLEASE DO NOT STOP THE SERVER")
+	log.Printf("Migrating old bucket metadata format at \"%s\" \"1\" to newer format \"2\"... ", fsPath)
+
+	defer func() {
+		if err == nil {
+			log.Println("SUCCESS.")
+		} else {
+			log.Printf("FAILED with %s.\n", err)
+		}
+	}()
```
krishnasrinivas: Instead of doing this style, the more conventional style is better:
```
err = migrateFSObject()
if err {...}
else {...}
```

cmd/fs-v1.go
```diff
@@ -110,17 +195,25 @@ func newFSObjectLayer(fsPath string) (ObjectLayer, error) {
...
+				return nil, merr
+			}
+		}
+
 		fsFormatPath := pathJoin(fsPath, minioMetaBucket, fsFormatJSONFile)
 		// Initialize format.json, if already exists overwrite it.
-		if serr := saveFormatFS(fsFormatPath, newFSFormatV1()); serr != nil {
+		if serr := saveFormatFS(fsFormatPath, newFSFormatV2()); serr != nil {
 			return nil, fmt.Errorf("Unable to initialize 'format.json', %s", serr)
 		}
```
krishnasrinivas: Lock should be held for the entire duration of load, migrate, save

cmd/fs-v1.go
```diff
@@ -110,17 +195,25 @@ func newFSObjectLayer(fsPath string) (ObjectLayer, error) {
...
+		if err == errFSFormatOld {
+			if merr := migrateFSObject(fsPath, fsUUID); merr != nil {
+				return nil, merr
+			}
+		}
+
 		fsFormatPath := pathJoin(fsPath, minioMetaBucket, fsFormatJSONFile)
 		// Initialize format.json, if already exists overwrite it.
-		if serr := saveFormatFS(fsFormatPath, newFSFormatV1()); serr != nil {
+		if serr := saveFormatFS(fsFormatPath, newFSFormatV2()); serr != nil {
```
krishnasrinivas: saveFormatFS() is not truncating previous contents. We will end up with corrupt format.json if the current json's size is less than previous contents.

cmd/fs-v1-helpers.go
harshavardhana: Sure it was there and moved to helpers..

cmd/fs-v1-metadata.go
harshavardhana: Sure

cmd/fs-v1.go
harshavardhana: Yes good thing..

cmd/fs-v1.go
harshavardhana: Perfect..

harshavardhana:
All changes requested done @krishnasrinivas

cmd/format-config-v1.go
```diff
@@ -47,6 +51,43 @@ type formatConfigV1 struct {
...
+	_, err = lk.Write(fbytes)
+	if err != nil {
+		return 0, traceError(err)
+	}
+	return int64(len(fbytes)), nil
+}
+
+func (f *formatConfigV1) ReadFrom(lk *lock.LockedFile) (n int64, err error) {
+	var fbytes []byte
+	fi, err := lk.Stat()
```
vadmeste, line 73: Why Stat() ? since we have ioutil.ReadAll later, is that to synchronize access to the file ?

harshavardhana: We need to know the exact size to read.. there is a reason for it but i do not remember why.

vadmeste:
I wasn't able to break this code. Probably one last thing is to disable signals as you suggested

```
          ch := make(chan os.Signal)
  
          signal.Notify(ch, syscall.SIGINT, syscall.SIGTERM)
  
          go func() {
                  for {
                          _, ok := <-ch
                          if !ok {
                                  break
                          }
                          log.Println("Please wait until the server finish upgrading..")
                  }
          }()
  
          // Working.. busy update
          time.Sleep(10 * time.Second)
          // Update finished.
  
          signal.Stop(ch)
          close(ch)
```

harshavardhana:
> ch := make(chan os.Signal)
>   
>           signal.Notify(ch, syscall.SIGINT, syscall.SIGTERM)
>   
>           go func() {
>                   for {
>                           _, ok := <-ch
>                           if !ok {
>                                   break
>                           }
>                           log.Println("Please wait until the server finish upgrading..")
>                   }
>           }()
>   
>           // Working.. busy update
>           time.Sleep(10 * time.Second)
>           // Update finished.
>   
>           signal.Stop(ch)
>           close(ch)

We don't need to do that because the server exit with block on `globalServiceDoneCh` and wouldn't exit.

vadmeste:
> We don't need to do that because the server exit with block on globalServiceDoneCh and wouldn't exit.

We are probably missing something, I've just tested with 10,000 buckets and ctrl-c stopped Minio before upgrading to format v2.

cmd/fs-v1-metadata.go
```diff
@@ -237,58 +250,114 @@ func newFSMetaV1() (fsMeta fsMetaV1) {
...
+
+	// Disk is in old format migrate object metadata.
+	if err == errFSFormatOld {
+		if err = migrateFSObject(fsPath, fsUUID); err != nil {
+			return err
+		}
+		format = newFSFormatV2()
+	}
+
+	// Initialize format.json, if already exists overwrite it.
```
krishnasrinivas: we should overwrite only if we migrated from V1 to V2, else leave it as it is

harshavardhana:
> We are probably missing something, I've just tested with 10,000 buckets and ctrl-c stopped Minio before upgrading to format v2.

Let me check.. it shouldn't happen from current code point of view.

cmd/fs-v1-metadata.go
```diff
@@ -237,58 +250,117 @@ func newFSMetaV1() (fsMeta fsMetaV1) {
...
+		}
+
+		// Initialize format v2.
+		format = newFSFormatV2()
+
+		// Initialize format.json.
+		if _, err = format.WriteTo(lk); err != nil {
+			return fmt.Errorf("Unable to initialize 'format.json', %s", err)
+		}
+	}
```
krishnasrinivas: still a problem, because on a fresh export directory we are not writing format to disk, i.e we endup with empty file created at line 322

harshavardhana: Oops.. yes.

harshavardhana:
Fixed and added CTRL-C catcher.. as well. @vadmeste @krishnasrinivas - let me know what else is needed.

krishnasrinivas:
Another problem :-(

If older binary is started on version-2 format.json it runs without exiting.

harshavardhana:
Nothing can be done for older binaries @krishnasrinivas.

krishnasrinivas (approved):

harshavardhana:
@vadmeste we need to review this as well after the release..

harshavardhana:
@vadmeste waiting on your :+1: .. let me know if you have more comments.

vadmeste (approved):

vadmeste:
I tested again. LGTM.

PR #4197, opened 2017-04-28
Add PKCS#12 certificate support for HTTPS server.

### LINKED ISSUE #3572: ssl: Add support of Windows PFX certificates
## Expected Behavior
To be able to detect and load pfx file inside .minio/certs/ directory 

## Current Behavior
PFX file is not recognizable

## Possible Solution
Use golang pfx API support

## Context
Adding new feature

## Your Environment
All

balamurugana:
Fixes #3572

## How Has This Been Tested?

* `make test`
* manually

## Types of changes

## Checklist:

### DIFF
--- a/cmd/certs.go
+++ b/cmd/certs.go
@@ -17,6 +17,7 @@
 package cmd
 
 import (
+	"bytes"
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/base64"
@@ -102,7 +103,7 @@ func parsePKCS12CertFile(pkcs12File string) (x509Certs []*x509.Certificate, tlsC
 
 	// Decode if the content is base64 encoded.
 	var decodedData []byte
-	if _, err = base64.StdEncoding.Decode(decodedData, data); err == nil {
+	if decodedData, err = ioutil.ReadAll(base64.NewDecoder(base64.StdEncoding, bytes.NewBuffer(data))); err == nil {
 		data = decodedData
 	}
 
--- a/cmd/certs.go
+++ b/cmd/certs.go
@@ -17,42 +17,45 @@
 package cmd
 
 import (
+	"crypto/tls"
 	"crypto/x509"
+	"encoding/base64"
 	"encoding/pem"
 	"fmt"
 	"io/ioutil"
 	"path/filepath"
+
+	"golang.org/x/crypto/pkcs12"
 )
 
-func parsePublicCertFile(certFile string) (certs []*x509.Certificate, err error) {
-	var bytes []byte
-
-	if bytes, err = ioutil.ReadFile(certFile); err != nil {
-		return certs, err
+func parsePublicCertFile(certFile string) (x509Certs []*x509.Certificate, err error) {
+	// Read certificate file.
+	var data []byte
+	if data, err = ioutil.ReadFile(certFile); err != nil {
+		return nil, err
 	}
 
 	// Parse all certs in the chain.
-	var block *pem.Block
-	var cert *x509.Certificate
-	current := bytes
+	current := data
 	for len(current) > 0 {
-		if block, current = pem.Decode(current); block == nil {
-			err = fmt.Errorf("Could not read PEM block from file %s", certFile)
-			return certs, err
+		var pemBlock *pem.Block
+		if pemBlock, current = pem.Decode(current); pemBlock == nil {
+			return nil, fmt.Errorf("Could not read PEM block from file %s", certFile)
 		}
 
-		if cert, err = x509.ParseCertificate(block.Bytes); err != nil {
-			return certs, err
+		var x509Cert *x509.Certificate
+		if x509Cert, err = x509.ParseCertificate(pemBlock.Bytes); err != nil {
+			return nil, err
 		}
 
-		certs = append(certs, cert)
+		x509Certs = append(x509Certs, x509Cert)
 	}
 
-	if len(certs) == 0 {
-		err = fmt.Errorf("Empty public certificate file %s", certFile)
+	if len(x509Certs) == 0 {
+		return nil, fmt.Errorf("Empty public certificate file %s", certFile)
 	}
 
-	return certs, err
+	return x509Certs, nil
 }
 
 func getRootCAs(certsCAsDir string) (*x509.CertPool, error) {
@@ -81,7 +84,7 @@ func getRootCAs(certsCAsDir string) (*x509.CertPool, error) {
 	for _, caFile := range caFiles {
 		caCert, err := ioutil.ReadFile(caFile)
 		if err != nil {
-			return rootCAs, err
+			return nil, err
 		}
 
 		rootCAs.AppendCertsFromPEM(caCert)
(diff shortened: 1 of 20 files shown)


cmd/server-mux.go
```diff
@@ -1,531 +0,0 @@
-/*
```
vadmeste: Can you do server-mux refactoring in a separate PR ? (if adding pcks#12 support doesn't require that)

harshavardhana: I concur @balamurugana we should avoid server-mux refactoring in this PR.

balamurugana:
@vadmeste @harshavardhana removed refactored HTTP server.

cmd/config-dir.go
```diff
@@ -41,6 +41,10 @@ const (
 
 	// Private key file for HTTPS.
 	privateKeyFile = "private.key"
+
+	// PKCS#12 certificate file for HTTPS.
+	minioPFXFile = "minio.pfx"
```
vadmeste: Probably this name is better 
`pfxFile = "cert.pfx"`

cmd/config-dir.go
```diff
@@ -41,6 +41,10 @@ const (
...
+	minioP12File = "minio.p12"
```
vadmeste: p12 and pfx looks like to be the same exact format, I think we can keep only one extension.

cmd/globals.go
```diff
@@ -94,6 +95,8 @@ var (
 	// IsSSL indicates if the server is configured with SSL.
 	globalIsSSL bool
 
+	globalTLSCertificate *tls.Certificate
```
vadmeste, line 98: I think this name is better `globalSrvTLSCertificate`

cmd/server-mux.go
```diff
@@ -408,33 +408,26 @@ func initListeners(serverAddr string, tls *tls.Config) ([]*ListenerMux, error) {
...
-		PreferServerCipherSuites: true,
-		// Set minimum version to TLS 1.2
-		MinVersion: tls.VersionTLS12,
-	} // Always instantiate.
-
-	if tlsEnabled {
-		// Configure TLS in the server
-		if config.NextProtos == nil {
-			config.NextProtos = []string{"http/1.1", "h2"}
+func (m *ServerMux) ListenAndServe(tlsCert *tls.Certificate) (err error) {
```
vadmeste, line 411: I think `srvTLSCert` name is better

cmd/server-mux.go
```diff
@@ -408,33 +408,26 @@ func initListeners(serverAddr string, tls *tls.Config) ([]*ListenerMux, error) {
...
+	var tlsConfig *tls.Config
```
vadmeste, line 412: I think `srvTLSConfig` name is better.

vadmeste:
Missing some documentation, you can add some paragraphs here `docs/tls/README.md`

By the way @balamurugana, how to generate this pfx file ?

I tested with this `openssl pkcs12 -export -nodes -out minio.pfx -inkey private.key -in public.crt -passout pass:` which is supposed to generate passwordless minio.pfx file from private.key and public.key.

vadmeste (changes requested):
Documentation is missing

cmd/config-dir.go
balamurugana: Done

cmd/config-dir.go
balamurugana: It is UX decision. But I am going with `.pfx`

cmd/globals.go
balamurugana: I think `Srv` is a duplicate meaning as this certificate is used only in server and there is no client TLS certificate we have now.

cmd/server-mux.go
balamurugana: I think `Srv` is a duplicate meaning as this certificate is used only in server and there is no client TLS certificate we have now.

cmd/server-mux.go
balamurugana: I think `Srv` is a duplicate meaning as this certificate is used only in server and there is no client TLS certificate we have now.

balamurugana:
> Missing some documentation, you can add some paragraphs here docs/tls/README.md

Done



> I tested with this openssl pkcs12 -export -nodes -out minio.pfx -inkey private.key -in public.crt -passout pass: which is supposed to generate passwordless minio.pfx file from private.key and public.key.

:+1:

balamurugana:
> Documentation is missing

What are other documentation than `docs/tls/README.md`?

vadmeste:
> I tested with this openssl pkcs12 -export -nodes -out minio.pfx -inkey private.key -in public.crt -passout pass: which is supposed to generate passwordless minio.pfx file from private.key and public.key.

sorry @balamurugana, I believe that I unintentionally removed one sentence, I wanted to say that it didn't work with openssl, I am not sure what I am missing.

balamurugana:
> sorry @balamurugana, I believe that I unintentionally removed one sentence, I wanted to say that it didn't work with openssl, I am not sure what I am missing.

You can try below commands
```sh
openssl genrsa -out server.key 2048
openssl req -new -key server.key -out server.csr
openssl x509 -req -days 3650 -in server.csr -signkey server.key -out server.crt
openssl pkcs12 -keypbe PBE-SHA1-3DES -certpbe PBE-SHA1-3DES -export -in server.crt -inkey server.key -out cert.pfx -name "my-cert-name"
```

cmd/certs.go
```diff
@@ -90,19 +94,70 @@ func getRootCAs(certsCAsDir string) (*x509.CertPool, error) {
...
+func parsePKCS12CertFile(pkcs12File string) (x509Certs []*x509.Certificate, tlsCert *tls.Certificate, err error) {
+	// Read pkcs12 file.
+	var data []byte
+	if data, err = ioutil.ReadFile(pkcs12File); err != nil {
+		return nil, nil, err
+	}
+
+	// Decode if the content is base64 encoded.
+	var decodedData []byte
+	if decodedData, err = ioutil.ReadAll(base64.NewDecoder(base64.StdEncoding, bytes.NewBuffer(data))); err == nil {
```
harshavardhana, line 106: It is better to fail here if err != nil  - rather than getting overwritten by the next `pkcs12.ToPEM()`

cmd/certs.go
```diff
@@ -90,19 +94,70 @@ func getRootCAs(certsCAsDir string) (*x509.CertPool, error) {
...
+		x509Certs, tlsCert, err = parsePKCS12CertFile(pkcs12File)
+	} else if isFile(getPublicCertFile()) && isFile(getPrivateKeyFile()) {
+		if x509Certs, err = parsePublicCertFile(getPublicCertFile()); err == nil {
+			var cert tls.Certificate
+			if cert, err = tls.LoadX509KeyPair(getPublicCertFile(), getPrivateKeyFile()); err == nil {
+				tlsCert = &cert
+			}
+		}
+	} else {
+		return nil, nil, nil, false, nil
```
harshavardhana: It is more readable if we can use a switch statement here i think.

cmd/certs.go
```diff
@@ -90,19 +94,70 @@ func getRootCAs(certsCAsDir string) (*x509.CertPool, error) {
...
 	}
 
-	if publicCerts, err = parsePublicCertFile(getPublicCertFile()); err != nil {
-		return publicCerts, rootCAs, secureConn, err
+	if err == nil {
```
harshavardhana: Why not do `if err != nil {` and return and then do the rest?

cmd/server-mux_test.go
```diff
@@ -370,7 +370,13 @@ func TestServerListenAndServeTLS(t *testing.T) {
 	}
 
 	// ListenAndServe in a goroutine, but we don't know when it's ready
-	go func() { errc <- m.ListenAndServe(certFile, keyFile) }()
+	var tlsCert tls.Certificate
+	if tlsCert, err = tls.LoadX509KeyPair(certFile, keyFile); err != nil {
+		t.Error(err)
```
harshavardhana: You can do 't.Fatal(err)` instead of 'return' it is more idiomatic Go test.

docs/tls/README.md
```diff
@@ -8,7 +8,7 @@ In this document, we will configure Minio servers with TLS certificates for both
 
 ## 2. Configure with existing certificates
 
-Assuming that you are already having private and public certificates, you will need to copy them under `certs` in your Minio config directory using the names `private.key` and `public.crt` for key and public certificates respectively.
+Assuming that you are already having private and public certificates, or passwordless PKCS#12 certificate, you will need to copy them under `certs` in your Minio config directory using the names `private.key` and `public.crt` for key and public certificates respectively, or `cert.pfx` in case of PKCS#12 certificate.
```
harshavardhana, line 11: Is there a password based approach as well? should we mention that we don't support it?.  Have you explored about can we support it? @balamurugana

harshavardhana (changes requested):

cmd/certs.go
balamurugana: No. PKCS#12 certificate can be in binary format or base64 encoded.   Here `data` is tried to base64 decode if successful decoded data is taken, else the read data is used.

cmd/certs.go
balamurugana: Done

cmd/certs.go
balamurugana: Done

cmd/server-mux_test.go
balamurugana: Both does different where `t.Fatal()` aborts test run whereas `t.Error(); return` exits `TestServerListenAndServeTLS()`.

However I am doing `t.Fatal()` now

docs/tls/README.md
balamurugana: > Is there a password based approach as well?

Yes.

> should we mention that we don't support it?

I mentioned as `passwordless`. Do you want more elaborate message?

> Have you explored about can we support it? @balamurugana

As a password, for security reasons, it should be read from stdin   Other ways are insecure.  Its UX decision how password needs to be read.

harshavardhana: We can discuss that @balamurugana - should we even do it. Can we detect if a given pfx file requires password?

balamurugana: IMO storing password of the certificate for every run is as insecure as passwordless certificate.  Asking password for every run will lead some service functionality breakage eg. restart.  


> Can we detect if a given pfx file requires password?

Its possible to do

deekoder:
Document using openssl.

deekoder:
We just need to document this. I will open one and assign it to @remco.

deekoder:
Closing this PR. @nl5887  will help us document this.

PR #4209, opened 2017-05-01
gateway: Restore bucket policy functionality for Azure

krishnasrinivas:
## Description

Bucket policy code for azure were removed by mistake, restoring it.

## How Has This Been Tested?

manually

## Types of changes

## Checklist:

### DIFF
--- a/cmd/gateway-azure.go
+++ b/cmd/gateway-azure.go
@@ -599,15 +599,56 @@ func azureListBlobsGetParameters(p storage.ListBlobsParameters) url.Values {
 // As the common denominator for minio and azure is readonly and none, we support
 // these two policies at the bucket level.
 func (a AzureObjects) SetBucketPolicies(bucket string, policyInfo policy.BucketAccessPolicy) error {
-	return traceError(NotSupported{})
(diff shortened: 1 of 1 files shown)


harshavardhana (approved):

cmd/gateway-azure.go
```diff
@@ -599,15 +599,56 @@ func azureListBlobsGetParameters(p storage.ListBlobsParameters) url.Values {
 // As the common denominator for minio and azure is readonly and none, we support
 // these two policies at the bucket level.
 func (a AzureObjects) SetBucketPolicies(bucket string, policyInfo policy.BucketAccessPolicy) error {
```
donatello, line 601: Though comments say that the "none" policy is supported, it seems like this function will return a `NotImplemented` error if the policy is not read-only.

Is it actually working?

krishnasrinivas: for `none` client calls `DELETE /testbucket/?policy=` which translates to DeleteBucketPolicies

donatello (approved):

PR #4213, opened 2017-05-01
Cleanup code that checks for updates

donatello:
## Description

This patch introduces a new type `updateParams` to parameterize the
code that checks for updates.

* Eliminate passing variables as-is across multiple function (e.g. the
  `mode` string in the earlier code).

* Remove trivial functions (e.g. earlier version of `DownloadReleaseData()` and
  `IsSourceBuild()`).

* Maintain testability of the code.

## Motivation and Context

Cleanup for better readability and maintainability.

## How Has This Been Tested?

Existing test cases pass (`make test`)

## Types of changes

## Checklist:

### DIFF
--- a/cmd/update-main.go
+++ b/cmd/update-main.go
@@ -22,7 +22,6 @@ import (
 	"io/ioutil"
 	"net/http"
 	"os"
-	"os/exec"
 	"path/filepath"
 	"runtime"
 	"strings"
@@ -68,61 +67,84 @@ const (
 	minioReleaseURL           = "https://dl.minio.io/server/minio/release/" + runtime.GOOS + "-" + runtime.GOARCH + "/"
 )
 
-func getCurrentReleaseTime(minioVersion, minioBinaryPath string) (releaseTime time.Time, err error) {
-	if releaseTime, err = time.Parse(time.RFC3339, minioVersion); err == nil {
-		return releaseTime, err
+var (
+	// Newer official download info URLs appear earlier below.
+	minioReleaseInfoURLs = []string{
+		minioReleaseURL + "minio.sha256sum",
+		minioReleaseURL + "minio.shasum",
 	}
+)
 
-	if !filepath.IsAbs(minioBinaryPath) {
-		// Make sure to look for the absolute path of the binary.
-		minioBinaryPath, err = exec.LookPath(minioBinaryPath)
-		if err != nil {
-			return releaseTime, err
-		}
+// minioVersionToReleaseTime - parses a standard official released
+// Minio version string. An Official binary's version string is the
+// release time formatted with RFC3339 (but without the timezone) -
+// see minioReleaseTagTimeLayout for the precise format.
+func minioVersionToReleaseTime(version string) (releaseTime time.Time, err error) {
+	return time.Parse(time.RFC3339, version)
+}
+
+// releaseTimeToReleaseTag - converts a release time to an official
+// Minio release tag.
+func releaseTimeToReleaseTag(releaseTime time.Time) string {
+	return "RELEASE." + releaseTime.Format(minioReleaseTagTimeLayout)
+}
+
+// releaseTagToReleaseTime - reverse of `releaseTimeToReleaseTag()`
+func releaseTagToReleaseTime(releaseTag string) (releaseTime time.Time, err error) {
+	tagTimePart := strings.TrimPrefix(releaseTag, "RELEASE.")
+	if tagTimePart == releaseTag {
+		return releaseTime, fmt.Errorf("%s is not a valid release tag", releaseTag)
 	}
+	return time.Parse(minioReleaseTagTimeLayout, tagTimePart)
+}
 
-	// Looks like version is minio non-standard, we use minio binary's ModTime as release time.
-	fi, err := osStat(minioBinaryPath)
+// getModTime - get the file modification time of `path`
+func getModTime(path string) (t time.Time, err error) {
+	// Convert to absolute path
+	absPath, err := filepath.Abs(path)
 	if err != nil {
-		err = fmt.Errorf("Unable to get ModTime of %s. %s", minioBinaryPath, err)
-	} else {
-		releaseTime = fi.ModTime().UTC()
+		return t, fmt.Errorf("Unable to get absolute path of %s. %s", path, err)
 	}
 
-	return releaseTime, err
+	// Get Stat info
+	fi, err := osStat(absPath)
+	if err != nil {
+		return t, fmt.Errorf("Unable to get ModTime of %s. %s", absPath, err)
+	}
+
+	// Return the ModTime
+	return fi.ModTime().UTC(), nil
 }
 
 // GetCurrentReleaseTime - returns this process's release time.  If it is official minio version,
 // parsed version is returned else minio binary's mod time is returned.
 func GetCurrentReleaseTime() (releaseTime time.Time, err error) {
-	return getCurrentReleaseTime(Version, os.Args[0])
+	if releaseTime, err = minioVersionToReleaseTime(Version); err == nil {
+		return releaseTime, err
+	}
+
+	// Looks like version is minio non-standard, we use minio
+	// binary's ModTime as release time:
+	return getModTime(os.Args[0])
 }
 
-// Check if we are indeed inside docker.
+// IsDocker - returns if the environment minio is running in docker or
+// not. The check is a simple file existence check.
+//
 // https://github.com/moby/moby/blob/master/daemon/initlayer/setup_unix.go#L25
 //
 //     "/.dockerenv":      "file",
 //
-func isDocker(dockerEnvFile string) (ok bool, err error) {
-	_, err = os.Stat(dockerEnvFile)
-	if err != nil {
-		if os.IsNotExist(err) {
-			err = nil
-		}
-		return false, err
-	}
-	return true, nil
-}
-
-// IsDocker - returns if the environment minio is running
-// is docker or not.
 func IsDocker() bool {
-	found, err := isDocker("/.dockerenv")
-	// We don't need to fail for this check, log
-	// an error and return false.
+	_, err := osStat("/.dockerenv")
+	if os.IsNotExist(err) {
+		return false
+	}
+
+	// Log error, as we will not propagate it to caller
 	errorIf(err, "Error in docker check.")
 
-	return found
+	return err == nil
 }
 
 // IsDCOS returns true if minio is running in DCOS.
@@ -165,46 +187,54 @@ func getHelmVersion(helmInfoFilePath string) string {
 	return ""
 }
 
-func isSourceBuild(minioVersion string) bool {
-	_, err := time.Parse(time.RFC3339, minioVersion)
-	return err != nil
-}
-
-// IsSourceBuild - returns if this binary is made from source or not.
+// IsSourceBuild - returns if this binary is a non-official build from
+// source code.
 func IsSourceBuild() bool {
-	return isSourceBuild(Version)
+	_, err := minioVersionToReleaseTime(Version)
+	return err != nil
 }
 
 // DO NOT CHANGE USER AGENT STYLE.
 // The style should be
 //
-//   Minio (<OS>; <ARCH>[; dcos][; kubernetes][; docker][; source]) Minio/<VERSION> Minio/<RELEASE-TAG> Minio/<COMMIT-ID> [Minio/univers-<PACKAGE_NAME>]
+//   Minio (<OS>; <ARCH>[; <MODE>][; dcos][; kubernetes][; docker][; source]) Minio/<VERSION> Minio/<RELEASE-TAG> Minio/<COMMIT-ID> [Minio/universe-<PACKAGE-NAME>] [Minio/helm-<HELM-VERSION>]
 //
-// For any change here should be discussed by openning an issue at https://github.com/minio/minio/issues.
+// Any change here should be discussed by opening an issue at https://github.com/minio/minio/issues.
 func getUserAgent(mode string) string {
-	userAgent := "Minio (" + runtime.GOOS + "; " + runtime.GOARCH
-	if mode != "" {
-		userAgent += "; " + mode
-	}
-	if IsDCOS() {
-		userAgent += "; dcos"
-	}
-	if IsKubernetes() {
-		userAgent += "; kubernetes"
-	}
-	if IsDocker() {
-		userAgent += "; docker"
-	}
-	if IsSourceBuild() {
-		userAgent += "; source"
+
+	userAgentParts := []string{}
+	// Helper function to concisely append a pair of strings to a
+	// the user-agent slice.
+	uaAppend := func(p, q string) {
+		userAgentParts = append(userAgentParts, p, q)
 	}
 
-	userAgent += ") Minio/" + Version + " Minio/" + ReleaseTag + " Minio/" + CommitID
+	uaAppend("Minio (", runtime.GOOS)
+	uaAppend("; ", runtime.GOARCH)
+	if mode != "" {
+		uaAppend("; ", mode)
+	}
+	if IsDCOS() {
+		uaAppend("; ", "dcos")
+	}
+	if IsKubernetes() {
+		uaAppend("; ", "kubernetes")
+	}
+	if IsDocker() {
+		uaAppend("; ", "docker")
+	}
+	if IsSourceBuild() {
+		uaAppend("; ", "source")
+	}
+
+	uaAppend(") Minio/", Version)
+	uaAppend(" Minio/", ReleaseTag)
+	uaAppend(" Minio/", CommitID)
 	if IsDCOS() {
 		universePkgVersion := os.Getenv("MARATHON_APP_LABEL_DCOS_PACKAGE_VERSION")
 		// On DC/OS environment try to the get universe package version.
 		if universePkgVersion != "" {
-			userAgent += " Minio/" + "universe-" + universePkgVersion
+			uaAppend(" Minio/universe-", universePkgVersion)
 		}
 	}
 
@@ -212,17 +242,17 @@ func getUserAgent(mode string) string {
 		// In Kubernetes environment, try to fetch the helm package version
 		helmChartVersion := getHelmVersion("/podinfo/labels")
 		if helmChartVersion != "" {
-			userAgent += " Minio/" + "helm-" + helmChartVersion
+			uaAppend(" Minio/helm-", helmChartVersion)
 		}
 	}
 
-	return userAgent
+	return strings.Join(userAgentParts, "")
 }
 
-func downloadReleaseData(releaseChecksumURL string, timeout time.Duration, mode string) (data string, err error) {
+func downloadReleaseURL(releaseChecksumURL string, timeout time.Duration, mode string) (content string, err error) {
 	req, err := http.NewRequest("GET", releaseChecksumURL, nil)
 	if err != nil {
-		return data, err
+		return content, err
 	}
 	req.Header.Set("User-Agent", getUserAgent(mode))
 
@@ -236,34 +266,43 @@ func downloadReleaseData(releaseChecksumURL string, timeout time.Duration, mode 
 
 	resp, err := client.Do(req)
 	if err != nil {
-		return data, err
+		return content, err
 	}
 	if resp == nil {
-		return data, fmt.Errorf("No response from server to download URL %s", releaseChecksumURL)
+		return content, fmt.Errorf("No response from server to download URL %s", releaseChecksumURL)
 	}
 	defer resp.Body.Close()
 
 	if resp.StatusCode != http.StatusOK {
-		return data, fmt.Errorf("Error downloading URL %s. Response: %v", releaseChecksumURL, resp.Status)
+		return content, fmt.Errorf("Error downloading URL %s. Response: %v", releaseChecksumURL, resp.Status)
 	}
-	dataBytes, err := ioutil.ReadAll(resp.Body)
+	contentBytes, err := ioutil.ReadAll(resp.Body)
 	if err != nil {
-		return data, fmt.Errorf("Error reading response. %s", err)
+		return content, fmt.Errorf("Error reading response. %s", err)
 	}
 
-	data = string(dataBytes)
-	return data, err
+	return string(contentBytes), err
 }
 
 // DownloadReleaseData - downloads release data from minio official server.
 func DownloadReleaseData(timeout time.Duration, mode string) (data string, err error) {
-	data, err = downloadReleaseData(minioReleaseURL+"minio.shasum", timeout, mode)
-	if err == nil {
-		return data, nil
+	for _, url := range minioReleaseInfoURLs {
+		data, err = downloadReleaseURL(url, timeout, mode)
+		if err != nil {
+			return data, err
+		}
 	}
-	return downloadReleaseData(minioReleaseURL+"minio.sha256sum", timeout, mode)
+	return data, fmt.Errorf("Failed to fetch release URL - last error: %s", err)
 }
 
+// parseReleaseData - parses release info file content fetched from
+// official minio download server.
+//
+// The expected format is a single line with two words like:
+//
+// fbe246edbd382902db9a4035df7dce8cb441357d minio.RELEASE.2016-10-07T01-16-39Z
+//
+// The second word must be `minio.` appended to a standard release tag.
 func parseReleaseData(data string) (releaseTime time.Time, err error) {
 	fields := strings.Fields(data)
 	if len(fields) != 2 {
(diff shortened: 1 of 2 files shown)


donatello:
Fixed conflicts. @krishnasrinivas @balamurugana Please review.

cmd/update-main.go
```diff
@@ -68,90 +68,112 @@ const (
...
 	return bytes.Contains(cgroup, []byte("docker")), err
 }
 
-// IsDocker - returns if the environment is docker or not.
-func IsDocker() bool {
-	found, err := isDocker("/proc/self/cgroup")
+// isDocker - returns if the environment is docker or not.
+func (u *updateParams) isDocker() bool {
+	found, err := u.isDockerInternal()
 	fatalIf(err, "Error in docker check.")
```
nitisht: Probably a better error message here will be `Error in reading cgroup file` ?

donatello: Improved it - we dont have to make it specific to the cgroup file as the check may change in the future.

cmd/update-main.go
```diff
@@ -68,90 +68,110 @@ const (
...
 
-func getCurrentReleaseTime(minioVersion, minioBinaryPath string) (releaseTime time.Time, err error) {
-	if releaseTime, err = time.Parse(time.RFC3339, minioVersion); err == nil {
+type updateParams struct {
+	versionStr         string
+	serverBinaryPath   string
+	cgroupFile         string
+	releaseChecksumURL string
+	serverMode         string
+}
```
krishnasrinivas: `updateParams` does not look like the right name for this.
i.e For example why is cgroupFile part of updateParams
why is isDockerInternal a method on updateParams

If the structure name and method names are changed so that they look related then it will look better

cmd/update-main.go
```diff
@@ -68,90 +68,110 @@ const (
 	minioReleaseURL           = "https://dl.minio.io/server/minio/release/" + runtime.GOOS + "-" + runtime.GOARCH + "/"
 )
 
-func getCurrentReleaseTime(minioVersion, minioBinaryPath string) (releaseTime time.Time, err error) {
-	if releaseTime, err = time.Parse(time.RFC3339, minioVersion); err == nil {
+type updateParams struct {
+	versionStr         string
```
balamurugana: `version` is better than `versionStr`.  `version` is anyway `string` type.

cmd/update-main.go
```diff
@@ -68,90 +68,110 @@ const (
...
+	serverBinaryPath   string
+	cgroupFile         string
+	releaseChecksumURL string
+	serverMode         string
+}
+
+// newUpdateParams - returns appropriate default value for
+// updateParams
+func newUpdateParams() *updateParams {
```
balamurugana: Why pointer is returned than value?

cmd/update-main.go
```diff
@@ -68,90 +68,110 @@ const (
...
 //
-// For any change here should be discussed by openning an issue at https://github.com/minio/minio/issues.
-func getUserAgent(mode string) string {
-	userAgent := "Minio (" + runtime.GOOS + "; " + runtime.GOARCH
-	if mode != "" {
-		userAgent += "; " + mode
+// For any change here should be discussed by openning an issue at
+// https://github.com/minio/minio/issues.
+func (u *updateParams) getUserAgent() string {
+	uaPieces := []string{runtime.GOOS, runtime.GOARCH}
```
balamurugana: `tokens` or `envTokens` is better than `uaPieces`

cmd/update-main.go
```diff
@@ -68,90 +68,110 @@ const (
 	minioReleaseURL           = "https://dl.minio.io/server/minio/release/" + runtime.GOOS + "-" + runtime.GOARCH + "/"
 )
 
-func getCurrentReleaseTime(minioVersion, minioBinaryPath string) (releaseTime time.Time, err error) {
-	if releaseTime, err = time.Parse(time.RFC3339, minioVersion); err == nil {
+type updateParams struct {
```
balamurugana: How about `updateChecker` than `updateParams`?

cmd/update-main.go
```diff
@@ -68,90 +68,110 @@ const (
...
 }
 
-// IsDocker - returns if the environment is docker or not.
-func IsDocker() bool {
-	found, err := isDocker("/proc/self/cgroup")
-	fatalIf(err, "Error in docker check.")
-
+// isDocker - returns if the environment is docker or not.
+func (u *updateParams) isDocker() bool {
+	found, err := u.isDockerInternal()
```
balamurugana: `isDockerEnv()` is better than `isDockerInternal()`.

cmd/update-main.go
```diff
@@ -235,20 +257,22 @@ func getDownloadURL() (downloadURL string) {
 	return minioReleaseURL + "minio"
 }
 
-func getUpdateInfo(timeout time.Duration, mode string) (older time.Duration, downloadURL string, err error) {
-	currentReleaseTime, err := GetCurrentReleaseTime()
+func (u *updateParams) getUpdateInfo(timeout time.Duration) (
```
balamurugana: IMO this method needs to be exported ie `GetUpdateInfo()`.  This will give better readability at consumer side.  but its your call.

balamurugana (changes requested):
Minor comments

cmd/update-main.go
donatello: Added more comments - this set of parameters enables writing tests for various pieces of logic. It should be more readable now.

cmd/update-main.go
donatello: Changed to the more explicit `updateCheckParams` :)

cmd/update-main.go
donatello: :+1:

cmd/update-main.go
donatello: Refactored a bit more and eliminated this function.

cmd/update-main.go
donatello: :+1:

cmd/update-main.go
donatello: :+1: for envTokens

cmd/update-main.go
donatello: Changed it to not depend on the struct - since the struct can be initialized here. No need to export I think unless we are following a particular convention.

cmd/server-main.go
```diff
@@ -80,7 +80,7 @@ EXAMPLES:
 // Check for updates and print a notification message
 func checkUpdate(mode string) {
 	// Its OK to ignore any errors during getUpdateInfo() here.
-	if older, downloadURL, err := getUpdateInfo(1*time.Second, mode); err == nil {
+	if older, downloadURL, err := getUpdateInfo(mode, 1*time.Second); err == nil {
```
balamurugana: Now I don't see any improvement/fixes the PR does than another way to implement with a struct.

balamurugana (changes requested):

cmd/server-main.go
donatello: < moved comment to main-thread so as to not lose it when/if the commit is modified >

donatello:
> Now I don't see any improvement/fixes the PR does than another way to implement with a struct.

@balamurugana This PR is only trying to improve the understandability of the code. It is not trying to fix any bugs. Let me try to list the improvements that are being attempted.

The struct is helping for the following reasons: 

### Avoid threading the `mode` argument through 5 functions

Excerpts from existing code:

```go
func getUpdateInfo(timeout time.Duration, mode string) (older time.Duration, downloadURL string, err error) {...}
func getLatestReleaseTime(timeout time.Duration, mode string) (releaseTime time.Time, err error) {...}
func DownloadReleaseData(timeout time.Duration, mode string) (data string, err error) {...}
func downloadReleaseData(releaseChecksumURL string, timeout time.Duration, mode string) (data string, err error) {...}
func getUserAgent(mode string) string {...}
```

Guess how many of the above functions actually use `mode` than just pass it around?

Just one.


### Avoid trivial functions to just hide and pass-in arguments to functions

1. `IsSourceBuild()`
```go
func isSourceBuild(minioVersion string) bool {
	_, err := time.Parse(time.RFC3339, minioVersion)
	return err != nil
}

func IsSourceBuild() bool {
	return isSourceBuild(Version)
}
```

2. `GetCurrentReleaseTime()`

```go
func GetCurrentReleaseTime() (releaseTime time.Time, err error) {
	return getCurrentReleaseTime(Version, os.Args[0])
}
```

3. `DownloadReleaseData()`

```go
func DownloadReleaseData(timeout time.Duration, mode string) (data string, err error) {
	return downloadReleaseData(minioReleaseURL+"minio.shasum", timeout, mode)
}

```

In this change, I am eliminating these multiple similarly named functions.


### Remove excessive string concatenation

In addition to changes related to introducing the struct, `getUserAgent()` is refactored to be significantly more readable (and less importantly, to have better performance).

```go
// Old
func getUserAgent(mode string) string {
	userAgent := "Minio (" + runtime.GOOS + "; " + runtime.GOARCH
	if mode != "" {
		userAgent += "; " + mode
	}
	if IsDocker() {
		userAgent += "; docker"
	}
	if IsSourceBuild() {
		userAgent += "; source"
	}
	userAgent += ") " + " Minio/" + Version + " Minio/" + ReleaseTag + " Minio/" + CommitID

	return userAgent
}

// New
func (u *updateCheckParams) getUserAgent() string {
	envTokens := []string{runtime.GOOS, runtime.GOARCH}
	if u.serverMode != "" {
		envTokens = append(envTokens, u.serverMode)
	}
	if u.isDocker() {
		envTokens = append(envTokens, "docker")
	}
	if u.isSourceBuild() {
		envTokens = append(envTokens, "source")
	}
	envInfo := strings.Join(envTokens, "; ")
	return fmt.Sprintf("Minio (%s)  Minio/%s Minio/%s Minio/%s",
		envInfo, Version, ReleaseTag, CommitID)
}
```

I hope the points above better illuminate the improvements. They are already mentioned briefly in the commit message. I hope your argument is not that these changes are too trivial to be included.

balamurugana:
> I hope the points above better illuminate the improvements.

:+1: 


> They are already mentioned briefly in the commit message. I hope your argument is not that these changes are too trivial to be included.

The commit message talks about a type `updateParams` which is not exposed/used now other than `getUpdateInfo()`.  You would need to fix that

cmd/server-main.go
balamurugana: As `mode` and `timeout` are required to be passed to `getUpdateInfo()`, No point in changing the order.

cmd/update-main.go
```diff
@@ -68,90 +68,108 @@ const (
 	minioReleaseURL           = "https://dl.minio.io/server/minio/release/" + runtime.GOOS + "-" + runtime.GOARCH + "/"
 )
 
-func getCurrentReleaseTime(minioVersion, minioBinaryPath string) (releaseTime time.Time, err error) {
-	if releaseTime, err = time.Parse(time.RFC3339, minioVersion); err == nil {
+// updateCheckParams - collection of parameters used to check if the
+// version of the current server process is older than the latest
+// available official release. They enable the testability of various
+// routines in the update checking process.
+type updateCheckParams struct {
```
balamurugana: the mid-path name `updateCheckParams` looks odd to me.  As `getUpdateInfo()` is not a method anymore, `updateParams` is enough.

cmd/update-main.go
```diff
@@ -68,90 +68,108 @@ const (
...
+	// release)
+	releaseChecksumURL string
+	// current process run mode
+	serverMode string
+}
+
+// getCurrentReleaseTime - returns this process's release time.  If it
+// is official minio version, parsed version is returned else minio
+// binary's mod time is returned.
+func (u *updateCheckParams) getCurrentReleaseTime() (releaseTime time.Time,
```
balamurugana: Having these methods to pointer is incorrect.

cmd/update-main_test.go
```diff
@@ -119,7 +119,10 @@ func TestGetCurrentReleaseTime(t *testing.T) {
 	}
```
balamurugana, line 107: You can add unit tests for `getUserAgent()` and `getDownloadURL()` for codecov more.

donatello:
> The commit message talks about a type `updateParams` which is not exposed/used now other than `getUpdateInfo()`. You would need to fix that

Done.

cmd/update-main.go
```diff
@@ -183,12 +203,10 @@ func downloadReleaseData(releaseChecksumURL string, timeout time.Duration, mode
 	return data, err
 }
 
-// DownloadReleaseData - downloads release data from minio official server.
-func DownloadReleaseData(timeout time.Duration, mode string) (data string, err error) {
-	return downloadReleaseData(minioReleaseURL+"minio.shasum", timeout, mode)
-}
+// parseReleaseData - helper to parse time of release.
+func parseReleaseData(data string) (releaseTime time.Time,
```
balamurugana: You could group methods of a type together for better readability.

donatello:
@krishnasrinivas Could you continue with your review?

balamurugana (changes requested):

cmd/update-main.go
donatello: Why is it incorrect?

balamurugana: You can try below
```go
package main

import (
	"fmt"
)

type mytype struct {
	name string
}

func (t mytype) hello() {
	fmt.Println("hello", t.name)
}

func (t *mytype) welcome() {
	fmt.Println("welcome", t.name)
}

func main() {
	t1 := mytype{"world"}
	t1.hello()
	t1.welcome()

	var t2 *mytype
	t2.hello()
	t2.welcome()
}
```

cmd/server-main.go
donatello: Done.

cmd/update-main.go
donatello: No, I think making the name explicit is good - it just needs to be read as `update-check params` meaning "parameters used for doing an update-check". The comment also makes this clear.

cmd/update-main.go
donatello: In the above code sample, `t2` is not assigned, so it crashes. Seems like an argument to not have pointer receivers in the Go language! Not sure what your point is, because in my code, the struct is initialized in only place in the application (and few more times in tests), in the same file and not used elsewhere. Obviously it cannot be a nil pointer (by trivial inspection).

Doing some research on this, I found review guidelines in the [Go Wiki](https://github.com/golang/go/wiki/CodeReviewComments#receiver-type) - some cherry-picked points that seem relevant (emphasis my own):

> * If the receiver is a large struct or array, a pointer receiver is more efficient. How large is large? Assume it's equivalent to passing all its elements as arguments to the method. If that feels too large, it's also too large for the receiver.
> * If the receiver is a small array or struct that is naturally a value type (for instance, something like the time.Time type), with no mutable fields and no pointers, or is just a simple basic type such as int or string, a value receiver makes sense. A value receiver can reduce the amount of garbage that can be generated; if a value is passed to a value method, an on-stack copy can be used instead of allocating on the heap. (The compiler tries to be smart about avoiding this allocation, but it can't always succeed.) **Don't choose a value receiver type for this reason without profiling first.**
> * Finally, when in doubt, use a pointer receiver.
> 

Though this is not performance sensitive, for curiosity sake, I did some profiling too using this code https://play.golang.org/p/PAuPwIWojW - it does not run in play.golang.org (takes too long), but running on my laptop I get:

```
$ ./testprog 
PointerReceiverResult
2000000000	         0.41 ns/op
ValueReceiverResult
500000000	         3.05 ns/op
```

The program uses a struct that is identical to `updateCheckParams` (has 5 string properties). Clearly the pointer receiver method is faster by like 7X. In summary, it is ok to use pointer receivers here.

cmd/update-main.go
donatello: Done.

cmd/update-main_test.go
donatello: Done.

donatello:
@balamurugana @krishnasrinivas Review points are addressed - please take a look again now.

cmd/update-main.go
```diff
@@ -235,20 +256,33 @@ func getDownloadURL() (downloadURL string) {
...
+func getUpdateInfo(mode string, timeout time.Duration) (
+	older time.Duration, downloadURL string, err error) {
+
+	u := updateCheckParams{
+		version:            Version,
+		serverBinaryPath:   os.Args[0],
+		cgroupFile:         "/proc/self/cgroup",
+		releaseChecksumURL: minioReleaseURL + "minio.shasum",
+		serverMode:         mode,
+	}
```
krishnasrinivas: You can notice that all the fields of `updateCheckParams` (except mode) are globally available and hence there is no reason for these fields to be part of this struct.

The intention of code cleanup is correct, the previous code was looking bad i.e mode was being passed all around and in the end to be used by a function somewhere down deep. This can be fixed in a better way by having `globalMinioMode` that gets set during init. Global variables are not a problem as long as no one modifies it, so having mode as globally available is fine. Also Bala is working putting all globals into a single struct so that it looks cleaner.

So you can do away with `updateCheckParams` and have mode globally available.

donatello: Yes, other than `mode` all fields are constants. If you see the previous code, this is clear there too (so why did the earlier code become complex anyway?!) - the reason for putting constant-like values as argument(s) to various functions is to be able to write tests. If these values are constants or are hard-coded then it is not possible to pass different values for the purposes of tests. See for example the test for `isDockerEnv()` - it is quite difficult to write it if the file path string "constant" `/proc/self/cgroup` is hard-coded - but clearly the test is useful and important here and explains why the function is implemented in the way it is (no need to even open `/proc/self/cgroup`!).

The problem with global variables (even those that do not change even during the lifetime of the process) is also related to writing tests. Global values impact multiple tests at the same time and so the test author needs to clearly understand which global variables are used or are affected by a function/code that is to be tested - the test author needs to ensure that before the test they are appropriately set and after the test cleaned up so it doesnt affect subsequent tests. This is generally a pain. This was discussed quite a lot in BLR office some months back with @balamurugana and @krisis - the problem with globals and why they need to be eliminated.

IMO, this is a clean implementation - serving the purpose of maintainability and testability.

krishnasrinivas: There are two options:

1) Passing variables as function parameters so that the functions are unit testable. Downside is longer function signature (or use structs like updateCheckParams and have methods on it)

2) Use global variables, harder to unit test but upside is simpler function signature.

We use 2nd style in minio all over, let's discuss what style we should follow so that we can be consistent.

donatello: @harshavardhana @deekoder Requesting to mark this PR as blocked on @abperiasamy

krishnasrinivas: An issue related to this discussion: https://github.com/minio/minio/issues/3198

cmd/update-main.go
balamurugana: * As you told WRT `updateCheckParams` is not performance sensitive. the pointer type is good for methods changing fields of the type.  This case is not applicable to the struct as none of the methods are updating fields.

* As I indicated in the example, pointer type must be initialized to avoid run time panic.  This is avoidable if methods are to value type.

* method to pointer type is even accessible by `nil` pointer.   Having any methods to pointer type will lead to panic.
```go
func (t *mytype) greet() {
	fmt.Println("spock hand")
}

var t2 *mytype
t1.greet()
```

* Choosing pointer or value type for method is by considering performance and/or changing field values.

donatello: > * As you told WRT updateCheckParams is not performance sensitive. the pointer type is good for methods changing fields of the type. This case is not applicable to the struct as none of the methods are updating fields.

Pointer receivers are **required** for methods changing fields of the struct/type. There is no choice in the matter there. The choice is only in the case of methods not updating the struct/type they are attached to.

So what are the considerations in choosing between pointer/value receiver types for such methods? The [Go Code Review Comments](https://github.com/golang/go/wiki/CodeReviewComments#receiver-type) say that it makes sense for small structs. It also says that you have to think of it as passing all properties of the struct as arguments to each value-receiver method. To me passing 5 variables to so many methods is excessive. So I find a pointer type to be better here. It is not a performance sensitivity argument - but more of an aesthetic to not be wasteful where it is trivial to do so.

You say having pointer receiver has the issue of passing an uninitialized pointer to a method:

> * As I indicated in the example, pointer type must be initialized to avoid run time panic. This is avoidable if methods are to value type.

However, such programmer mistakes are not easily overcome by simply robotically following a convention - see this code:

``` go
package main

import (
	"fmt"
)

type mytype struct{}

func (t mytype) greet() {
	fmt.Println("spock hand")
}

func doSomething(k mytype) {
	k.greet()
}

func main() {
	var t1 *mytype
	doSomething(*t1)
}
```

We have neatly followed the convention and used value-receiver types. However, this also crashes because the caller sent in an unintialized pointer. **The conclusion is that the programmer has to be careful to avoid such pitfalls (especially in Golang). There is no *holy-cow* convention that will help avoid shooting yourself in the foot.**

Again, in this change, the scope of the struct is just one file (ignoring test files for obvious reasons). **It is quite clear that pointers will be always be initialized in this code.** So I still say that it is a fine choice to use pointer-receiver here.


> * method to pointer type is even accessible by nil pointer. Having any methods to pointer type will lead to panic.
> 
> ``` go
> func (t *mytype) greet() {
> 	fmt.Println("spock hand")
> }
> 
> // (fixed a bug here to make it compile)
> var t1 *mytype
> t1.greet()
> ```

Again, I don't understand this point - you say `"Having any methods to pointer type will lead to panic."` but just gave an example yourself of a pointer-receiver method that will [not panic or crash](https://play.golang.org/p/8NKAuImyP1). Here, it appears that it doesn't matter if the method is pointer/value-receiver, because it makes no use of any properties in the type. **Additionally, all the code in the present change are counter-examples to the statement "pointer-receiver methods will lead to panic".**

> * Choosing pointer or value type for method is by considering performance and/or changing field values.

This is not true - as I said earlier, if the method needs to change fields in the struct it has to be a pointer-receiver type. This is compulsory. There is no choice here.

The choice comes only in the case of methods not changing their type's fields. The main consideration in Go coding guidelines *is performance*! Please take some time to go through the [Go Code Review Comments](https://github.com/golang/go/wiki/CodeReviewComments#receiver-type) - they in fact say to default to pointer types if in doubt.

krishnasrinivas: From [here](https://github.com/golang/go/wiki/CodeReviewComments#receiver-type):

> If the receiver is a large struct or array, a pointer receiver is more efficient. How large is large? Assume it's equivalent to passing all its elements as arguments to the method. If that feels too large, it's also too large for the receiver.

We'll need some sort of convention to decide when a struct is considered "large". We can check with AB on this.

> If the receiver is a small array or struct that is naturally a value type (for instance, something like the time.Time type), with no mutable fields and no pointers, or is just a simple basic type such as int or string, a value receiver makes sense.

This suggests that for anything other than basic types like int, string or a "naturally value type" like time.Time we should use pointer receiver.

> A value receiver can reduce the amount of garbage that can be generated; if a value is passed to a value method, an on-stack copy can be used instead of allocating on the heap. (The compiler tries to be smart about avoiding this allocation, but it can't always succeed.) Don't choose a value receiver type for this reason without profiling first.


_Don't choose a value receiver type for this reason without profiling first._

Very confusing, are they saying use pointers even for basic types like int, string and then perform profiling and then change it to value receivers if there is a performance problem?

donatello: Another official resource on this topic is here - https://golang.org/doc/faq#methods_on_values_or_pointers - appears that value receivers are good (i.e. efficient and clear) for small structs according to the doc.

Also, in some cases, for value receivers, a copy of the struct may end up getting allocated on the heap rather than the stack. This creates more work for the GC as items allocated on the function's call stack are freed at the end of a function call (and don't need the GC). Go determines where to put the allocation based on [Escape Analysis](https://en.wikipedia.org/wiki/Escape_analysis). I couldn't find any official docs about the implementation but there are references to it in golang issue tracker, reddit, mailing list and other places.

deekoder:
@donatello  - Will have a discussion today with @abperiasamy , @krishnasrinivas  and @balamurugana

deekoder:
ping @krishnasrinivas @balamurugana  and @nitisht on this. @donatello can you rebase again.

donatello:
Updated this PR for latest master. This is refactored from scratch, so reviewers may disregard the discussion above up to this point. I have not added a new struct this time, since it was contentious last time.

Some tests are removed - this is following @abperiasamy's advice about not writing code in a weird way simply for the sake of writing unit tests.

harshavardhana:
> Updated this PR for latest master. This is refactored from scratch, so reviewers may disregard the discussion above up to this point. I have not added a new struct this time, since it was contentious last time.

@donatello perhaps you could re-submit the PR to discard the history and freshly review?

donatello:
> @donatello perhaps you could re-submit the PR to discard the history and freshly review?
@harshavardhana ok!

PR #4217, opened 2017-05-02
Refactor HTTP server to address bugs

balamurugana:
* Remove unnecessary goroutine to start multiple TCP listeners.
* HTTP server waits for shutdown to maximum of Server.ShutdownTimeout than per serverShutdownPoll.
* Handles new connection errors properly.
* Handles error on start of HTTP server properly by exiting minio process.

## How Has This Been Tested?

* `make test`
* manually

## Types of changes

## Checklist:


harshavardhana:
Can you list the various bug this addresses in the description @balamurugana ??.   So we can validate.

pkg/http/listener.go
```diff
@@ -0,0 +1,284 @@
...
+				listener.errorLogFunc(err,
+					"Error in reading from new connection %s at server %s",
+					bufconn.RemoteAddr(), bufconn.LocalAddr())
+			}
+			bufconn.Close()
+			return nil, nil
+		}
+
+		// Return bufconn if read data is a valid HTTP method.
+		tokens := strings.SplitN(string(data), " ", 2)
```
harshavardhana: Have you profiled this code? @balamurugana - i wonder if we can avoid using `SplitN()` or simply use bytes directly instead of strings for this. We need to avoid allocation.

harshavardhana (changes requested):
Can you also fix the CI build?

donatello:
@balamurugana Could you please fix conflicts and the build if this PR is ready?

deekoder:
Let's discuss on thu. We have to merge or close this pr soon imo.

pkg/http/server.go
```diff
@@ -0,0 +1,195 @@
...
+		shutdownTimeout := srv.ShutdownTimeout
+		shutdownTimeoutCh := time.After(shutdownTimeout)
+		ticker := time.NewTicker(serverShutdownPoll)
+		defer ticker.Stop()
+		for {
+			select {
+			case <-shutdownTimeoutCh:
+				return errors.New("timed out. some connections are still active. doing abnormal shutdown")
+			case <-ticker.C:
+				if atomic.LoadInt32(&srv.requestCount) <= 0 {
```
harshavardhana: There a problem in this which is exposed in this PR and fixed https://github.com/minio/minio/pull/4500#issuecomment-306980377 

requestCount if at all is `1` then this loop will never honor `shutdownTimeoutCh` since `time.After()` only gets invoked if the other case is waiting for 5secs > . 

Basically this would wait for ConnTimeout i.e 30secs until when requestCount will be `1` for Go rpc connections. So gracefulShutdownTimeout is never invoked. Just like how it is in current master.

Can you check @balamurugana

harshavardhana (changes requested):

pkg/http/server.go
balamurugana: As per our discussion in slack, https://github.com/minio/minio/pull/4500 is already fixed here.

pkg/http/bufconn.go
```diff
@@ -0,0 +1,88 @@
...
+type bufConn struct {
+	net.Conn
+	bufReader              *bufio.Reader // buffered reader wraps reader in net.Conn.
+	readTimeout            time.Duration // sets the read timeout in the connection.
+	writeTimeout           time.Duration // sets the write timeout in the connection.
+	updateBytesReadFunc    func(int)     // function to be called to update bytes read.
+	updateBytesWrittenFunc func(int)     // function to be called to update bytes written.
+}
+
+func (c *bufConn) setReadTimeout() {
```
harshavardhana, line 35: We shouldn't set this timeout for RPC operations, since deadlines like this for both Read and Write lead to issues where `net.Conn` prematurely returns UnexpectedEOF on rpcClient and server rejects the connection.

We need to come up with a way to avoid these before we get in this refactor. I would suggest a prototype work towards migrating to grpc might be needed here.

harshavardhana (changes requested):

pkg/http/listener.go
```diff
@@ -0,0 +1,294 @@
...
+	"io"
+	"net"
+	"strings"
+	"sync"
+	"time"
+)
+
+// HTTP methods.
+var methods = []string{
+	"OPTIONS",
```
harshavardhana: Use the strings from golang http standard library to avoid mistakes..  `http.MethodOptions` etc.

pkg/http/listener.go
```diff
@@ -0,0 +1,294 @@
...
+		}(listener.tcpListeners[i], listener.doneChs[i])
+	}
+	wg.Wait()
+
+	listener.doneChs = nil
+	return nil
+}
+
+// Addr - net.Listener interface compatible method returns net.Addr.  In case of multiple TCP listeners, it returns '0.0.0.0' as IP address.
+func (listener *httpListener) Addr() (addr net.Addr) {
```
harshavardhana, line 249: why is this needed? just to be net.Listener compatible?

pkg/http/listener.go
```diff
@@ -0,0 +1,294 @@
...
+	if ip := net.ParseIP("0.0.0.0"); ip != nil {
+		tcpAddr.IP = ip
+	}
+
+	addr = tcpAddr
+	return addr
+}
+
+// Addrs - returns all address information of TCP listeners.
+func (listener *httpListener) Addrs() (addrs []net.Addr) {
```
harshavardhana, line 265: same as above why?

pkg/http/listener.go
```diff
@@ -0,0 +1,294 @@
...
+	for i := range listener.tcpListeners {
+		addrs = append(addrs, listener.tcpListeners[i].Addr())
+	}
+
+	return addrs
+}
+
```
harshavardhana, line 272: Needs a comment.. on the functionality this provides.

pkg/http/listener.go
```diff
@@ -0,0 +1,294 @@
...
+	tcpKeepAliveTimeout time.Duration,
+	readTimeout time.Duration,
+	writeTimeout time.Duration,
+	updateBytesReadFunc func(int),
+	updateBytesWrittenFunc func(int),
+	errorLogFunc func(error, string, ...interface{})) (listener *httpListener, err error) {
+
+	var tcpListeners []*net.TCPListener
+	// Close all opened listeners on error
+	defer func() {
```
harshavardhana, line 288: Avoid `defer func()`  stack problems by passing the arguments to `func()` explicitly.. like 

```
defer func(listeners []*net.TCPListener) {
}
```

pkg/http/server.go
```diff
@@ -0,0 +1,195 @@
...
+
+	// Wrap given handler to do additional
+	// * automatic redirect to HTTPS if given TLS is configured.
+	// * return 503 (service unavailable) if the server in shutdown.
+	wrappedHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		atomic.AddInt32(&srv.requestCount, 1)
+		defer atomic.AddInt32(&srv.requestCount, -1)
+
+		// If TLS is configured and the request is sent as non-TLS, return 307 (temporary redirect).
+		if tlsConfig != nil && r.TLS == nil {
```
harshavardhana: Rebase this to ensure that we are rejecting Plain-text connections for a TLS server.

pkg/http/bufconn.go
balamurugana: I understand the problem in RPC with timeouts.  I couldn't able to get why grpc evaluation blocks this PR. This PR addresses http server issues and provided way to override timeouts.

pkg/http/server.go
```diff
@@ -0,0 +1,195 @@
...
+		srv.inShutdownMutex.Unlock()
+
+		// Close underneath HTTP listener.
+		srv.listenerMutex.Lock()
+		err := srv.listener.Close()
+		srv.listenerMutex.Unlock()
+
+		// Wait for opened connection to be closed up to Shutdown timeout.
+		shutdownTimeout := srv.ShutdownTimeout
+		shutdownTimeoutCh := time.After(shutdownTimeout)
```
harshavardhana: Better to use time.NewTimer() instead of `time.After()` since they won't be garbage collected.  Its a Shutdown() routine so it is generally fine. But it is your call..

pkg/http/server.go
```diff
@@ -0,0 +1,195 @@
...
+	if certificate != nil {
+		tlsConfig = &tls.Config{
+			PreferServerCipherSuites: true,
+			MinVersion:               tls.VersionTLS12,
+			NextProtos:               []string{"http/1.1", "h2"},
+		}
+		tlsConfig.Certificates = append(tlsConfig.Certificates, *certificate)
+	}
+
+	httpServer := Server{}
```
harshavardhana: Why keep it separately like this.. why not this style?

```
return &Server{
    Handler: handler,
    TLSConfig: tlsconfig,
.....
....
}
```

harshavardhana (changes requested):

pkg/http/listener.go
balamurugana: Done

pkg/http/listener.go
balamurugana: Yes.  As mentioned in the comment.

pkg/http/listener.go
balamurugana: `net.Listener` interface returns only single address with `Addr()` function whereas this `httpListener` handles multiple addresses, `Addrs()` function is enhanced version returns multiple net.Addr in an array.

pkg/http/listener.go
balamurugana: Done

pkg/http/listener.go
balamurugana: Unable to understand why we need to do it here.  Can you explain?

pkg/http/server.go
balamurugana: Done. 403 (like Microsoft IIS) is returned for insecure access on secure server.

pkg/http/bufconn.go
harshavardhana: Correct but we need to figure out a solution for RPC as well.

pkg/http/server.go
balamurugana: Done. using `time.NewTimer()`

pkg/http/server.go
balamurugana: Using this style to applicable fields. As `Server` is an extended type of `http.Server` some fields are not available in the initialization.

pkg/http/bufconn.go
balamurugana: Yep. IMO this PR is not getting blocked with that.

pkg/http/listener.go
harshavardhana: Would not happen what i said in this case..

pkg/http/server.go
```diff
@@ -0,0 +1,192 @@
...
+	// * automatic redirect to HTTPS if given TLS is configured.
+	// * return 503 (service unavailable) if the server in shutdown.
+	wrappedHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		atomic.AddInt32(&srv.requestCount, 1)
+		defer atomic.AddInt32(&srv.requestCount, -1)
+
+		// If TLS is configured and the request is sent as non-TLS, return 403 (forbidden).
+		if tlsConfig != nil && r.TLS == nil {
+			w.WriteHeader(http.StatusForbidden)
+			w.Write([]byte("Use HTTPS scheme"))
```
vadmeste: The old code does this:
```
		if tlsEnabled && r.TLS == nil {
			// TLS is enabled but request is not TLS
			// configured - return error to client.
			writeErrorResponse(w, ErrInsecureClientRequest, &url.URL{})
		}
```

Can you see if you can follow it ?

pkg/http/server.go
```diff
@@ -0,0 +1,192 @@
...
+	DefaultReadTimeout = 30 * time.Second
+
+	// DefaultWriteTimeout - default timout to write data to accepted connection.
+	DefaultWriteTimeout = 30 * time.Second
+
+	// DefaultMaxHeaderBytes - default maximum HTTP header size in bytes.
+	DefaultMaxHeaderBytes = 1 * humanize.MiByte
+)
+
+// Server - extended http.Server supports multiple addresses and transparent redirection from insecure to secure connection.
```
vadmeste: We don't have transparent redirection from insecure to secure connection anymore

pkg/http/server.go
balamurugana: 1. The correct error code for this situation is `http.StatusForbidden` (as mentioned above, Microsoft IIS does with extended code).
2. Returning minio specific error xml would be confusing if the package is used outside of minio ie I would like to keep it generic.
3. Current error message `"Cannot respond to plain-text request from TLS-encrypted server"` is slightly confusing.  I would take modified message of it like `SSL Required` (like IIS) or `"Server refuses to serve non-encrypted requests"`.

pkg/http/server.go
balamurugana: Done

cmd/signals.go
```diff
@@ -18,32 +18,72 @@ package cmd
...
+				rerr := restartProcess()
+				errorIf(rerr, "Unable to restart the server.")
+
+				if err == nil {
+					err = rerr
+				}
+
+				exit(err)
+			case serviceStop:
+				exit(stopService())
```
vadmeste: We had this message `Received signal to exit.` when Control-C is pressed, I guess you can add it here.

balamurugana: Done

harshavardhana:
@donatello can you review this PR as well?

pkg/http/server.go
```diff
@@ -0,0 +1,195 @@
...
+	srv.listener = listener
+	srv.listenerMutex.Unlock()
+
+	// Start servicing with listener.
+	return srv.Server.Serve(listener)
+}
+
+// Shutdown - shuts down HTTP server.
+func (srv *Server) Shutdown() error {
+	srv.inShutdownMutex.Lock()
```
donatello, line 125: Need to defer to the mutex unlock immediately? There are return statements in the if-block that may bypass unlocking.

pkg/http/server.go
```diff
@@ -0,0 +1,195 @@
...
+	// Wrap given handler to do additional
+	// * return 403 (forbidden) with 'SSL required' error message if non-encrypted
+	//   request received for secured server.
+	// * return 503 (service unavailable) if the server in shutdown.
+	wrappedHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		atomic.AddInt32(&srv.requestCount, 1)
+		defer atomic.AddInt32(&srv.requestCount, -1)
+
+		// If TLS is configured and the request is sent as non-TLS, return 403 (forbidden).
+		if tlsConfig != nil && r.TLS == nil {
```
donatello: Since TLS handshake will fail with a non-TLS client, will this block ever be executed? Or is not the case that the handshake always fails if server is TLS, and client is not?

donatello (changes requested):
Initial review.

pkg/http/server.go
balamurugana: Yesterday's change in `acceptTCP()` closure makes this check obsolete.  Fixing now.

pkg/http/server.go
balamurugana: No. `inShutdownMutex` guards `inShutdown` field read/write access.  It is unlocked properly after read and/or write of the field.

donatello: Ok got it.

donatello:
**EDIT**: My bad. The errors I got were due to some mix up in my local checkouts. Please disregard this comment.

~~@balamurugana After compiling your branch and running locally (without TLS), I tried to run minio-go tests against it, but it fails - I get following errors logged on server console:~~

cmd/globals.go
```diff
@@ -96,6 +99,12 @@ var (
 	// IsSSL indicates if the server is configured with SSL.
 	globalIsSSL bool
 
+	globalTLSCertificate *tls.Certificate
+
+	globalHTTPServer        *miniohttp.Server
+	globalHTTPServerErrorCh = make(chan error)
+	globalOSSignalCh        = make(chan os.Signal, 1)
```
harshavardhana, line 116: Is this ever closed? and why not ?

cmd/signals.go
```diff
@@ -18,32 +18,75 @@ package cmd
...
 
-	return trapCh
+			exit(err)
+		case osSignal := <-globalOSSignalCh:
+			log.Printf("Exiting on signal %v\n", osSignal)
+			exit(stopService())
+		case signal := <-globalServiceSignalCh:
+			switch signal {
+			case serviceStatus:
+				// Ignore this ignore at the moment.
```
harshavardhana: Looks like a typo.

cmd/signals.go
```diff
@@ -18,32 +18,75 @@ package cmd
...
+				errorIf(rerr, "Unable to restart the server.")
+
+				if err == nil {
+					err = rerr
+				}
+
+				exit(err)
+			case serviceStop:
+				log.Println("Stopping on service signal")
+				exit(stopService())
```
harshavardhana: can you say this a stopProcess()? or rename restartProcess() as restartService() ?

cmd/signals.go
```diff
@@ -18,32 +18,75 @@ package cmd
...
+				rerr := restartProcess()
+				errorIf(rerr, "Unable to restart the server.")
+
+				if err == nil {
+					err = rerr
+				}
+
+				exit(err)
+			case serviceStop:
+				log.Println("Stopping on service signal")
```
harshavardhana, line 87: We can keep this similar to how it is today. `log.Println("Received signal to exit.")`

pkg/http/bufconn.go
```diff
@@ -0,0 +1,88 @@
...
+	if c.readTimeout != 0 {
```
harshavardhana, line 36: You should stop setting readTimeouts after the fact that we have started reading from the Conn successfully.  The reason being that it doesn't really help to have deadlines in this manner which provide no proper control. 

This is carefully visible in my previous comment on how net/rpc can disconnect and requires continuous reconnects every 30secs.

pkg/http/bufconn.go
```diff
@@ -0,0 +1,88 @@
...
+
+// Peek - returns the next n bytes without advancing the reader.  It just wraps bufio.Reader.Peek().
+func (c *bufConn) Peek(n int) ([]byte, error) {
+	c.setReadTimeout()
+	return c.bufReader.Peek(n)
+}
+
+// Read - reads data from the connection using wrapped buffered reader.
+func (c *bufConn) Read(b []byte) (n int, err error) {
+	c.setReadTimeout()
```
harshavardhana, line 55: Lets not do this after first successful read.

pkg/http/server.go
```diff
@@ -0,0 +1,184 @@
...
+		Addrs:               addrs,
+		ShutdownTimeout:     DefaultShutdownTimeout,
+		TCPKeepAliveTimeout: DefaultTCPKeepAliveTimeout,
+		listenerMutex:       &sync.Mutex{},
+		inShutdownMutex:     &sync.Mutex{},
+	}
+	httpServer.Handler = handler
+	httpServer.TLSConfig = tlsConfig
+	httpServer.ReadTimeout = DefaultReadTimeout
+	httpServer.WriteTimeout = DefaultWriteTimeout
```
harshavardhana, line 180: We don't need to set this value it is not useful and would cause frequent timeouts for slow clients. 30 secs is still a small number for situations minio is deployed.

pkg/http/server.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+
+	httpServer := &Server{
+		Addrs:               addrs,
+		ShutdownTimeout:     DefaultShutdownTimeout,
+		TCPKeepAliveTimeout: DefaultTCPKeepAliveTimeout,
+		listenerMutex:       &sync.Mutex{},
+		inShutdownMutex:     &sync.Mutex{},
+	}
+	httpServer.Handler = handler
```
harshavardhana, line 177: Do not see a reason why initialization of `httpServer` and tlsConfig and timeouts is written differently. Why can't this be moved in into a single line.. ? 

We don't even need to allocate httpServer as a variable here, simply do this. 

```
return &Server{
 		Addrs:               addrs,
 		ShutdownTimeout:     DefaultShutdownTimeout,
 		TCPKeepAliveTimeout: DefaultTCPKeepAliveTimeout,
 		listenerMutex:       &sync.Mutex{},
 		inShutdownMutex:     &sync.Mutex{},
                Handler: handler,
                TLSConfig: tlsConfig,
                ReadTimeout: DefaultReadTimeout,
                MaxHeaderBytes: DefaultMaxHeaderBytes,
 	}
```

pkg/http/server_test.go
```diff
@@ -0,0 +1,103 @@
...
+package http
+
+import (
+	"crypto/tls"
+	"fmt"
+	"net/http"
+	"reflect"
+	"testing"
+)
+
```
harshavardhana, line 26: Add a comment on purpose of the test.

pkg/http/listener.go
```diff
@@ -0,0 +1,326 @@
...
+			// As stop signal is received, close accepted connection.
+			if result.conn != nil {
+				result.conn.Close()
+			}
+			return false
+		}
+	}
+
+	// Closure to accept single connection.
+	acceptTCP := func(tcpConn *net.TCPConn, doneCh <-chan struct{}) {
```
harshavardhana, line 111: why closures @balamurugana - why not stateless functions?

pkg/http/listener.go
```diff
@@ -0,0 +1,326 @@
...
+	"errors"
+	"io"
+	"net"
+	"net/http"
+	"strings"
+	"sync"
+	"time"
+)
+
+var sslRequiredErrMsg = []byte("HTTP/1.0 403 Forbidden\r\n\r\nSSL required")
```
harshavardhana, line 30: is this part of RFC or something if yes then can we state that here?

pkg/http/listener.go
```diff
@@ -0,0 +1,326 @@
...
+				bufconn.Close()
+				return
+			}
+
+			// Check whether the connection contains HTTP request or not.
+			bufconn = newBufConn(tlsConn, listener.readTimeout, listener.writeTimeout,
+				listener.updateBytesReadFunc, listener.updateBytesWrittenFunc)
+
+			// Peek bytes of maximum length of all HTTP methods.
+			data, err := bufconn.Peek(methodMaxLen)
```
harshavardhana, line 163: Why is there a double Peek() ? there is a Peek() before.. Previous implementation has only one.

pkg/http/listener.go
```diff
@@ -0,0 +1,326 @@
...
+				send(acceptResult{bufconn, nil}, doneCh)
+			} else {
+				// As TLS is configured and we got plain text HTTP request,
+				// return 403 (forbidden) error.
+				bufconn.Write(sslRequiredErrMsg)
+				bufconn.Close()
+			}
+			return
+		}
+
```
harshavardhana, line 144: There is a buffConn() leak here @balamurugana the tlsConfig verification only happens if the its one of the valid HTTP methods. If its not a valid http method `tlsConfig != nil`  would proceed without calling bufconn.Close() on an existing plain-text connection.

harshavardhana (commented):
One more set of reviews.. also testing locally.

cmd/globals.go
balamurugana: Reading of this channel is done only by `handleSignals()` and once value on the channel received, it calls `os.Exit()`.  I think closing channel is not useful.  Moreover there is a chance of panic `write on closed channel` by writer `signal.Notify()`

cmd/signals.go
balamurugana: Done

cmd/signals.go
balamurugana: Technically it is not a signal, it receives a service request to stop.  I kept this to differentiate signal and service request.  Let me know you still want it current way.

cmd/signals.go
balamurugana: Unable to understand.  Do you want to name the closure as `stopProcess`?

harshavardhana: Correct.

cmd/globals.go
harshavardhana: Once the signal.Notify gets control we should also close the signal handler such that signal.Stop is called.  This is achieved in signals.go . Not sure why keeping it global helps.. it can still be allocated inside the signalTrap and returned.

All channels would still receive since kernel signal handler is common no matter how many signal.Notify you invoke.

pkg/http/listener.go
balamurugana: There is no spec in the RFC.  I borrowed the idea from Microsoft IIS.  This is already discussed in the PR previously.  With lake of github review system its hard to see previous conversions.

pkg/http/listener.go
balamurugana: If they are methods to `httpListener`, they should be called only by `start()` ie they have no meaning outside `start()` and there is no control on whether they are called only by `start()`.  Secondly they can't be unit tested.

pkg/http/listener.go
balamurugana: I am not able to understand.  Below is the logic of this block.

1. Tune `TCPConn` and make `bufConn`
2. Peek bytes of maximum length of all HTTP methods irrespective of TLS configuration.
3. If valid HTTP method is found and TLS is not configured, we got valid client connection and return.
4. If valid HTTP method is found and TLS is configured, return error `sslRequiredErrMsg` indicating client should use TLS and drop the connection.
5. If invalid HTTP method is found and TLS is configured, attempt for TLS handshake.     
  i. If handshake fails, drop the connection. may be junk data sent by the client. 
  ii. If handshake succeeds, peek bytes of maximum length of all HTTP methods to check whether the client sends valid HTTP request via TLS. If its valid, we got valid client connection and return
6.  Finally drop the connection because of junk data is sent by the client in TLS/non-TLS mode.

With this logic, I don't find the leak problem in the implementation.

pkg/http/listener.go
balamurugana: Please check the above reply.

pkg/http/server.go
balamurugana: This was asked previously and I replied with the detail ie some fields are not available only after object creation because of extended type.

pkg/http/server_test.go
balamurugana: Do you want obvious comment like `Test NewServer()`

cmd/signals.go
balamurugana: Done

cmd/globals.go
balamurugana: I understand calling `signal.Stop()` stops writing to the channel.  `signal.Notify()` is called in `serverMain()` after a goroutine starting http server and before object layer initialization.  This means, signals occur in between two codes also handled.

> All channels would still receive since kernel signal handler is common no matter how many signal.Notify you invoke.

Not sure I understand this correctly.  If you say to avoid global variable `globalOSSignalCh`, it can be done.

pkg/http/listener.go
harshavardhana: ~~The problem you are not seeing is that bufconn if TLS comes into the `if` block (given server is also tls configured) and it gets overwritten by a new bufConn without closing the connection. You use previous buffConn to create a tlsConn and that tlsConn replaces the an active bufConn with another bufConn.~~

~~So there is a leak.~~

pkg/http/listener.go
harshavardhana: Multiple Peeks are not useful @balamurugana this function path can be re-implemented.  We should only have one Peek() and that Peek() is sufficient to tell us the type of incoming connection is.

pkg/http/server_test.go
harshavardhana: No the non obvious what this test is really doing. I don't understand TestNewServer() - what it does, does it test just NewServer() init? does it test if the NewServer() is usable? etc. 

The non-obvious is what we need to bring in here..

pkg/http/server.go
harshavardhana: I must have missed can you post it again? @balamurugana

pkg/http/listener.go
balamurugana: Could you suggest a better path?

pkg/http/server.go
balamurugana: @harshavardhana those fields are available only after creating new objects because of extended type. hence its done this way.

harshavardhana: Still don't understand when i see the function @balamurugana .- how does with in this function that is affecting?

balamurugana: You need to fix compilation errors like
```
pkg/http/server.go:176: unknown Server field 'Handler' in struct literal
```

harshavardhana: that is simple @balamurugana .. it should be like this. 

```
 return &Server{
                 Server: &http.Server{
                         Handler:        handler,
                         TLSConfig:      tlsConfig,
                         ReadTimeout:    DefaultReadTimeout,
                         WriteTimeout:   DefaultWriteTimeout,
                         MaxHeaderBytes: DefaultMaxHeaderBytes,
                 },
                 Addrs:               addrs,
                 ShutdownTimeout:     DefaultShutdownTimeout,
                 TCPKeepAliveTimeout: DefaultTCPKeepAliveTimeout,
                 listenerMutex:       &sync.Mutex{},
                 inShutdownMutex:     &sync.Mutex{},
-        }
```

balamurugana: Why do you want to do that?

harshavardhana: This is the idiomatic style, what you have done requires someone to really know what is going on. You are making it complicated for no reason. I would have never guessed this is the reason, it is harder to follow.

pkg/http/listener.go
harshavardhana: There is a better way.. 

```diff
+var methodMaxLen = 7
 
 type acceptResult struct {
        conn net.Conn
@@ -107,6 +99,24 @@ func (listener *httpListener) start() {
                }
        }
 
+       guessTLSProtocol := func(buf []byte) bool {
+               // Check for HTTP2 methods first.
+               for _, m := range defaultHTTP2Methods {
+                       if strings.HasPrefix(string(buf), m) {
+                               return false
+                       }
+               }
+
+               // Check for HTTP1 methods.
+               for _, m := range defaultHTTP1Methods {
+                       if strings.HasPrefix(string(buf), m) {
+                               return false
+                       }
+               }
+
+               return true
+       }
+
        // Closure to accept single connection.
        acceptTCP := func(tcpConn *net.TCPConn, doneCh <-chan struct{}) {
                // Tune accepted TCP connection.
@@ -128,65 +138,44 @@ func (listener *httpListener) start() {
                        return
                }
 
-               // Return bufconn if read data is a valid HTTP method.
-               tokens := strings.SplitN(string(data), " ", 2)
-               if isHTTPMethod(tokens[0]) {
-                       if listener.tlsConfig == nil {
-                               send(acceptResult{bufconn, nil}, doneCh)
-                       } else {
+               if !guessTLSProtocol(data) {
+                       // Return bufconn if read data is a valid HTTP method.
+                       if listener.tlsConfig != nil {
                                // As TLS is configured and we got plain text HTTP request,
                                // return 403 (forbidden) error.
                                bufconn.Write(sslRequiredErrMsg)
                                bufconn.Close()
+                               return
                        }
+                       send(acceptResult{bufconn, nil}, doneCh)
                        return
                }
 
-               if listener.tlsConfig != nil {
-                       // As the listener is configured with TLS, try to do TLS handshake, drop the connection if it fails.
-                       tlsConn := tls.Server(bufconn, listener.tlsConfig)
-                       if err := tlsConn.Handshake(); err != nil {
-                               if listener.errorLogFunc != nil {
-                                       listener.errorLogFunc(err,
-                                               "TLS handshake failed with new connection %s at server %s",
-                                               bufconn.RemoteAddr(), bufconn.LocalAddr())
-                               }
-                               bufconn.Close()
-                               return
-                       }
-
-                       // Check whether the connection contains HTTP request or not.
-                       bufconn = newBufConn(tlsConn, listener.readTimeout, listener.writeTimeout,
-                               listener.updateBytesReadFunc, listener.updateBytesWrittenFunc)
-
-                       // Peek bytes of maximum length of all HTTP methods.
-                       data, err := bufconn.Peek(methodMaxLen)
-                       if err != nil {
-                               if listener.errorLogFunc != nil {
-                                       listener.errorLogFunc(err,
-                                               "Error in reading from new TLS connection %s at server %s",
-                                               bufconn.RemoteAddr(), bufconn.LocalAddr())
-                               }
-                               bufconn.Close()
-                               return
-                       }
-
-                       // Return bufconn if read data is a valid HTTP method.
-                       tokens := strings.SplitN(string(data), " ", 2)
-                       if isHTTPMethod(tokens[0]) {
-                               send(acceptResult{bufconn, nil}, doneCh)
-                               return
-                       }
+               if listener.tlsConfig == nil {
+                       // Client is TLS and server is not TLS configured.
+                       // return 403 (forbidden) error.
+                       bufconn.Write(httpRequiredErrMsg)
+                       bufconn.Close()
+                       return
                }
 
-               if listener.errorLogFunc != nil {
-                       listener.errorLogFunc(errors.New("junk message"),
-                               "Received non-HTTP message from new connection %s at server %s",
-                               bufconn.RemoteAddr(), bufconn.LocalAddr())
+               // As the listener is configured with TLS, try to do TLS
+               // handshake, drop the connection if it fails.
+               tlsConn := tls.Server(bufconn, listener.tlsConfig)
+               if err = tlsConn.Handshake(); err != nil {
+                       if listener.errorLogFunc != nil {
+                               listener.errorLogFunc(err,
+                                       "TLS handshake failed with new connection %s at server %s",
+                                       bufconn.RemoteAddr(), bufconn.LocalAddr())
+                       }
+                       bufconn.Close()
+                       return
                }
-
-               bufconn.Close()
-               return
+               send(acceptResult{newBufConn(tlsConn,
+                       listener.readTimeout,
+                       listener.writeTimeout,
+                       listener.updateBytesReadFunc,
+                       listener.updateBytesWrittenFunc), nil}, doneCh)
        }
 
        // Closure to handle new connections till done channel is not closed.
@@ -198,9 +187,9 @@ func (listener *httpListener) start() {
                                if !send(acceptResult{nil, err}, doneCh) {
                                        return
                                }
-                       } else {
-                               go acceptTCP(tcpConn, doneCh)
+                               continue
                        }
+                       go acceptTCP(tcpConn, doneCh)
                }
        }
```

balamurugana: Can you send me the code than the diff? Thanks.

pkg/http/server.go
balamurugana: What do you mean by `what you have done requires someone to really know what is going on`?  Why assignment statements are complicated?  I completely disagree

harshavardhana: Basically its hidden that Handler should be set outside after allocation - anyone writing Go would try to move it out to simplify.  Without you telling me, i had no idea that was the case.

pkg/http/listener.go
harshavardhana: https://github.com/balamurugana/minio/pull/9 here you go @balamurugana

abperiasamy:
This is a very poor PR review process. It is also not helpful to send competing PRs instead of writing helpful reviews.

balamurugana:
Resubmitting fresh PR for better review

PR #4255, opened 2017-05-04
Webhook endpoints can fail, we must start the server.

harshavardhana:
## Description
Webhook endpoints can fail, we must start the server.

## Motivation and Context
This PR fixes a regression introduced in #4060

## How Has This Been Tested?
Manually.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/notify-webhook.go
+++ b/cmd/notify-webhook.go
@@ -69,10 +69,15 @@ func isNetErrorIgnored(err error) bool {
 	}
 	switch err.(type) {
 	case net.Error:
-		switch err.(type) {
+		switch e := err.(type) {
 		case *net.DNSError, *net.OpError, net.UnknownNetworkError:
 			return true
 		case *url.Error:
+			// Fixes https://github.com/minio/minio/issues/4050
+			switch e.Err.(type) {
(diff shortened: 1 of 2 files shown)


cmd/notify-webhook.go
```diff
@@ -69,10 +69,15 @@ func isNetErrorIgnored(err error) bool {
 	}
 	switch err.(type) {
 	case net.Error:
-		switch err.(type) {
+		switch e := err.(type) {
 		case *net.DNSError, *net.OpError, net.UnknownNetworkError:
 			return true
 		case *url.Error:
+			// Fixes https://github.com/minio/minio/issues/4050
```
vadmeste, line 76: Can you add a log when we ignore the error ? (in lookupEndpoint) it will be showed only one time. That would be better since we never try to send a notification again so the user needs to be aware. (also since there is some code that will be removed in go 1.8 in lookupEndpoint, you don't need a perfect solution for the moment).

harshavardhana: Done

vadmeste (approved):

krishnasrinivas (approved):

PR #4264, opened 2017-05-05
config: Do not migrate config file if not needed.

### LINKED ISSUE #4233: [bug] Cryptic error message due to bad value in config file
When there is a bad value in the configuration file, the server appears to report an error unrelated to the bad configuration value.

When setting the kafka configuration value `notify.kafka.brokers` whose value has type list, a simple string was given `"localhost:9092"`, but the error was cryptic:

```
$ minio server /tmp/newd
FATA[0000] Config migration failed.                      cause=Unable to load config version ‘6’. json: cannot unmarshal string into Go value of type []string source=[server-main.go:109:initConfig()]
```

While the type error makes sense (since we know the problem already), it does not say where the error happened. Also the error message says it is related to configuration migration - however the configuration was already at version 18 (latest afaik).

## Expected Behavior

Error message should not say configuration migration at the very least. It would be good if it says which item in the configuration file had a parsing error.

## Current Behavior

As described above.

## Possible Solution

Unknown.

1. Enable kafka in the configuration and set the `brokers` value as described above.
2. Start the server.

## Context

Happened during testing of something unrelated to this.

## Your Environment

```
Version: 2017-05-03T19:23:01Z
Release-Tag: DEVELOPMENT.2017-05-03T19-23-01Z
Commit-ID: f0b5c0ec7cbd6604d4a43ed0869318741823a40a
```
* Environment name and version (e.g. nginx 1.9.1): Linux

harshavardhana:
## Description
config: Do not migrate config file if not needed. Also improve the error message returned by `pkg/quick`.

## Motivation and Context
Fixes #4233

## How Has This Been Tested?
Manually.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/config-migrate.go
+++ b/cmd/config-migrate.go
@@ -27,77 +27,125 @@ import (
 // DO NOT EDIT following message template, please open a github issue to discuss instead.
 var configMigrateMSGTemplate = "Configuration file %s migrated from version '%s' to '%s' successfully.\n"
 
+// Migrates all config versions from "1" to "18".
 func migrateConfig() error {
-	// Purge all configs with version '1'.
+	// Purge all configs with version '1',
+	// this is a special case since version '1' used
+	// to be a filename 'fsUsers.json' not 'config.json'.
 	if err := purgeV1(); err != nil {
 		return err
 	}
-	// Migrate version '2' to '3'.
-	if err := migrateV2ToV3(); err != nil {
-		return err
-	}
-	// Migrate version '3' to '4'.
-	if err := migrateV3ToV4(); err != nil {
-		return err
-	}
-	// Migrate version '4' to '5'.
-	if err := migrateV4ToV5(); err != nil {
-		return err
-	}
-	// Migrate version '5' to '6.
-	if err := migrateV5ToV6(); err != nil {
-		return err
-	}
-	// Migrate version '6' to '7'.
-	if err := migrateV6ToV7(); err != nil {
-		return err
-	}
-	// Migrate version '7' to '8'.
-	if err := migrateV7ToV8(); err != nil {
-		return err
-	}
-	// Migrate version '8' to '9'.
-	if err := migrateV8ToV9(); err != nil {
-		return err
-	}
-	// Migrate version '9' to '10'.
-	if err := migrateV9ToV10(); err != nil {
-		return err
-	}
-	// Migrate version '10' to '11'.
-	if err := migrateV10ToV11(); err != nil {
-		return err
-	}
-	// Migrate version '11' to '12'.
-	if err := migrateV11ToV12(); err != nil {
-		return err
-	}
-	// Migrate version '12' to '13'.
-	if err := migrateV12ToV13(); err != nil {
-		return err
-	}
-	// Migrate version '13' to '14'.
-	if err := migrateV13ToV14(); err != nil {
-		return err
-	}
-	// Migrate version '14' to '15'.
-	if err := migrateV14ToV15(); err != nil {
-		return err
-	}
-	// Migrate version '15' to '16'.
-	if err := migrateV15ToV16(); err != nil {
-		return err
-	}
-	// Migrate version '16' to '17'.
-	if err := migrateV16ToV17(); err != nil {
-		return err
-	}
-	// Migrate version '17' to '18'.
-	if err := migrateV17ToV18(); err != nil {
+
+	// Load only config version information.
+	version, err := quick.GetVersion(getConfigFile())
+	if err != nil {
 		return err
 	}
 
-	return nil
+	// Conditional to migrate only relevant config versions.
+	// Upon success migration continues to the next version in sequence.
+	switch version {
+	case "2":
+		// Migrate version '2' to '3'.
+		if err = migrateV2ToV3(); err != nil {
+			return err
+		}
+		fallthrough
+	case "3":
+		// Migrate version '3' to '4'.
+		if err = migrateV3ToV4(); err != nil {
+			return err
+		}
+		fallthrough
+	case "4":
+		// Migrate version '4' to '5'.
+		if err = migrateV4ToV5(); err != nil {
+			return err
+		}
+		fallthrough
+	case "5":
+		// Migrate version '5' to '6.
+		if err = migrateV5ToV6(); err != nil {
+			return err
+		}
+		fallthrough
+	case "6":
+		// Migrate version '6' to '7'.
+		if err = migrateV6ToV7(); err != nil {
+			return err
+		}
+		fallthrough
+	case "7":
+		// Migrate version '7' to '8'.
+		if err = migrateV7ToV8(); err != nil {
+			return err
+		}
+		fallthrough
+	case "8":
+		// Migrate version '8' to '9'.
+		if err = migrateV8ToV9(); err != nil {
+			return err
+		}
+		fallthrough
+	case "9":
+		// Migrate version '9' to '10'.
+		if err = migrateV9ToV10(); err != nil {
+			return err
+		}
+		fallthrough
+	case "10":
+		// Migrate version '10' to '11'.
+		if err = migrateV10ToV11(); err != nil {
+			return err
+		}
+		fallthrough
+	case "11":
+		// Migrate version '11' to '12'.
+		if err = migrateV11ToV12(); err != nil {
+			return err
+		}
+		fallthrough
+	case "12":
+		// Migrate version '12' to '13'.
+		if err = migrateV12ToV13(); err != nil {
+			return err
+		}
+		fallthrough
+	case "13":
+		// Migrate version '13' to '14'.
+		if err = migrateV13ToV14(); err != nil {
+			return err
+		}
+		fallthrough
+	case "14":
+		// Migrate version '14' to '15'.
+		if err = migrateV14ToV15(); err != nil {
+			return err
+		}
+		fallthrough
+	case "15":
+		// Migrate version '15' to '16'.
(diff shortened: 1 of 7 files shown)


harshavardhana:
Here is how this can be validated 

Create a config file with version '14'.
```
$ ./minio.RELEASE.2017-03-16T21-50-32Z server --config-dir /tmp/cc /tmp/ll
``` 

```
$ minio server --config-dir /tmp/cc /tmp/ll
Configuration file /tmp/cc/config.json migrated from version '14' to '15' successfully.
Configuration file /tmp/cc/config.json migrated from version '15' to '16' successfully.
Configuration file /tmp/cc/config.json migrated from version '16' to '17' successfully.
Configuration file /tmp/cc/config.json migrated from version '17' to '18' successfully.
...
```

If the config is already at '18' then if there is a type error in the `config.json` , error message should indicate that the config migration was not attempted but instead the error is about loading config file. 

```
$ minio server ~/test
FATA[0000] Unable to load minio config file              cause=Unable to parse JSON, type 'number' cannot be converted into the Go 'string' type source=[server-main.go:114:initConfig()]
```

vadmeste (approved):

cmd/server-main.go
```diff
@@ -106,8 +106,12 @@ func enableLoggers() {
...
 	// Config file does not exist, we create it fresh and return upon success.
 	if isFile(getConfigFile()) {
-		fatalIf(migrateConfig(), "Config migration failed.")
-		fatalIf(loadConfig(), "Unable to load minio config file")
+		// Check if we should migrate the config,
+		// if not proceed to load the config.
+		if shouldMigrate() {
+			fatalIf(migrateConfig(), "Config migration failed.")
+		}
+		fatalIf(loadConfig(), "Unable to load config version %s", v18)
```
balamurugana: How about below?
```go
if err = loadConfig(); err != nil {
  fatalIf(migrateConfig(), "Failed to migrate configuration file")
  fatalIf(loadConfig(), "Unable to load minio config file")
}
```

balamurugana (changes requested):

cmd/server-main.go
harshavardhana: Load config will fail @balamurugana for older releases, we just want to see the version not rest of the fields.

donatello (changes requested):
The error is almost right..
```
$ minio server /tmp/newd
FATA[0000] Config migration failed.                      cause=Unable to load config version ‘6’. Unable to parse JSON, type 'string' cannot be converted into the Go '' type source=[server-main.go:110:initConfig()]
```
It shows destination type as empty string.

harshavardhana:
> The error is almost right..
> 
> $ minio server /tmp/newd
> FATA[0000] Config migration failed.                      cause=Unable to load config version ‘6’. Unable to parse JSON, type 'string' cannot be converted into the Go '' type source=[server-main.go:110:initConfig()]
> It shows destination type as empty string.

can you show how you reproduce this? @donatello

cmd/server-main.go
harshavardhana: Done @balamurugana

donatello (approved):
:+1: 

```
$ minio server /tmp/nwed
FATA[0000] Unable to load config version:(18).           cause=Unable to parse JSON, type 'string' cannot be converted into the Go '[]string' type source=[server-main.go:117:initConfig()]
```

cmd/config-migrate.go
```diff
@@ -27,6 +28,39 @@ import (
...
+	if srvCfgVersion.Version == v18 {
+		return nil
+	}
+
+	// Config is readable but older, proceed to migrate.
+	return errConfigOld
+}
+
+// Migrates all config versions from "1" to "18".
 func migrateConfig() error {
```
balamurugana: This patch addresses only to the latest version. However the problem is still there in migration because of we do 1 to N version migration irrespective of current version of the config.  For example,
```go
type ConfigV6 struct {
  Version string
  Value string
}

type ConfigV7 struct {
  Version string
  Value []string
}
```
When config file is not current version and which is in version 7 like
```json
{
     "Version" : "7",
    "Value": ["myval1",  "myval2"]
}
```

the current logic still breaks
```go
	// Migrate version '6' to '7'.
	if err := migrateV6ToV7(); err != nil {
		return err
	}
```

cmd/config-migrate.go
```diff
@@ -27,6 +28,39 @@ import (
...
+// proceed to migrate.
+func checkConfigForMigrate() error {
+	configFile := getConfigFile()
+
+	type serverConfigVersion struct {
+		Version string
+	}
+
+	srvCfgVersion := &serverConfigVersion{}
+	_, err := quick.Load(configFile, srvCfgVersion)
```
balamurugana: You could add `quick.GetVersion(configFile string)` in `quick` package and use it here.

balamurugana (changes requested):

cmd/config-migrate.go
harshavardhana: Ah yes sure..

cmd/config-migrate.go
harshavardhana: Current is the only thing which this tries to address.. That is the not intention of this PR @balamurugana. We will do the other ones later which requires more work.

balamurugana: If so, the PR title/description doesn't convey that.  However migration fix is not that complex
```go
func migrateConfig() error {
	version, err := quick.GetVersion(getConfigFile())
	if err != nil {
		return err
	}

	switch version {
	case "1":
		err = purgeV1()
	case "2":
		if err = migrateV2ToV3(); err == nil {
			fallthrough
		}
	case "3":
		if err = migrateV3ToV4(); err == nil {
			fallthrough
		}
	case "4":
		if err = migrateV4ToV5(); err == nil {
			fallthrough
		}
	case "5":
		if err = migrateV5ToV6(); err == nil {
			fallthrough
		}
	case "6":
		if err = migrateV6ToV7(); err == nil {
			fallthrough
		}
	case "7":
		if err = migrateV7ToV8(); err == nil {
			fallthrough
		}
	case "8":
		if err = migrateV8ToV9(); err == nil {
			fallthrough
		}
	case "9":
		if err = migrateV9ToV10(); err == nil {
			fallthrough
		}
	case "10":
		if err = migrateV10ToV11(); err == nil {
			fallthrough
		}
	case "11":
		if err = migrateV11ToV12(); err == nil {
			fallthrough
		}
	case "12":
		if err = migrateV12ToV13(); err == nil {
			fallthrough
		}
	case "13":
		if err = migrateV13ToV14(); err == nil {
			fallthrough
		}
	case "14":
		if err = migrateV14ToV15(); err == nil {
			fallthrough
		}
	case "15":
		if err = migrateV15ToV16(); err == nil {
			fallthrough
		}
	case "16":
		if err = migrateV16ToV17(); err == nil {
			fallthrough
		}
	case "17":
		if err = migrateV17ToV18(); err == nil {
			fallthrough
		}
	}

	return err
}
```

cmd/config-migrate.go
```diff
@@ -27,6 +28,39 @@ import (
...
+// proceed to migrate.
+func checkConfigForMigrate() error {
+	configFile := getConfigFile()
+
+	type serverConfigVersion struct {
+		Version string
+	}
+
+	srvCfgVersion := &serverConfigVersion{}
+	version, err := quick.GetVersion(configFile, srvCfgVersion)
```
balamurugana: Any quick friendly configuration file must have `Version` field.  There is no need to define a struct type to do here.  Its quick package responsibility to read version from the file. I still suggest `quick.GetVersion(configFile string)`

balamurugana (changes requested):

cmd/config-migrate.go
harshavardhana: Agreed.. let me do that.

cmd/config-migrate.go
harshavardhana: This makes sense .. let me do here @balamurugana i didn't think about it. Good idea.

balamurugana: this way there is no need of function `shouldMigrate()`

harshavardhana:
> if err = migrateV17ToV18(); err == nil {
> 			fallthrough
> 		}
> 	}

Fallthrough doesn't work apparently @balamurugana inside a conditional.

harshavardhana:
Done updated with all requested changes @balamurugana

balamurugana:
> Fallthrough doesn't work apparently @balamurugana inside a conditional.

you could do
```go
case "6":
  if err = migrateV6ToV7(); err != nil {
    return err
  }
  fallthrough
```

harshavardhana:
> case "6":
>   if err = migrateV6ToV7(); err != nil {
>     return err
>   }
>   fallthrough

Already done @balamurugana , was just FYI.

pkg/quick/quick.go
```diff
@@ -196,12 +196,23 @@ func New(data interface{}) (Config, error) {
...
+// GetVersion - extracts the version information.
+func GetVersion(filename string) (version string, err error) {
+	var qc Config
+	type quickVersion struct {
+		Version string
+	}
+	if qc, err = New(&quickVersion{}); err == nil {
+		err = qc.Load(filename)
+	}
+	return qc.Version(), err
```
balamurugana: How about below?
```go
	baseConfig := struct {
		Version string
	}{}
	_, err := Load(configFile, &baseConfig)
	return baseConfig.Version, err
```

harshavardhana: Done @balamurugana

balamurugana (approved):

PR #4266, opened 2017-05-05
config: Support kafka brokers to be specified dynamically.

harshavardhana:
## Description
Currently the brokers have a strict []string requirement. To
make it easier for users to consume it is easier to allow this
field to be dynamic.

This PR adds support for following styles

  - "addr:port" as a single broker address.
  - [ "addr1:port1", "addr2:port2"... ] as a list of brokers.
  - "addr1:port1, addr2:port2, ..." as a comma separated list of brokers.

## Motivation and Context
Ease of usability for user.

## How Has This Been Tested?
Manually.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/config-v18.go
+++ b/cmd/config-v18.go
@@ -22,6 +22,7 @@ import (
 	"io/ioutil"
 	"sync"
 
+	"github.com/minio/minio-go/pkg/set"
 	"github.com/minio/minio/pkg/quick"
 	"github.com/tidwall/gjson"
 )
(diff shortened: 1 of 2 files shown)


vadmeste (approved):

balamurugana:
@harshavardhana 

> "addr1:port1, addr2:port2, ..." as a comma separated list of brokers.

Why do we need one more representation of array?  As per JSON, its non-standard. Many options will lead to confusion.

harshavardhana:
> Why do we need one more representation of array? As per JSON, its non-standard. Many options will lead to confusion.

The reason is that brokers are usually provided as comma separated.

balamurugana:
> The reason is that brokers are usually provided as comma separated.

Where is it specified like this?

balamurugana:
@harshavardhana basically its not a good idea to have dynamic JSON in config.  If we need one more values, I would prefer array ie
```json
{ "value": ["one"] }
```
or
```json
{ "value": ["one", "two"] }
```

harshavardhana:
> Where is it specified like this?

It is usually how kafka clients take broker list for example here https://github.com/Shopify/sarama/blob/master/examples/http_server/http_server.go#L21 - this is the library we use.

balamurugana:
> It is usually how kafka clients take broker list for example here https://github.com/Shopify/sarama/blob/master/examples/http_server/http_server.go#L21 - this is the library we use.

As its an example code reading from values from env or command line argument, it may deal that style.  However as a JSON configuration file, I won't recommend dynamic JSON.

harshavardhana:
> As its an example code reading from values from env or command line argument, it may deal that style. However as a JSON configuration file, I won't recommend dynamic JSON.

I agree, it is more of a proposal to see if we indeed need to support. I know the typed is more preferred than typeless.

harshavardhana:
Closing this.. as this is not needed for the time being.

PR #4298, opened 2017-05-08
Implementing Gateway backend Google Cloud Storage

nl5887:


cmd/gateway-main.go
```diff
@@ -190,15 +199,10 @@ func gatewayMain(ctx *cli.Context) {
...
-	// Second argument is endpoint.	If no endpoint is specified then the
-	// gateway implementation should use a default setting.
-	endPoint, secure, err := parseGatewayEndpoint(ctx.Args().Get(1))
-	fatalIf(err, "Unable to parse endpoint")
-
 	// Create certs path for SSL configuration.
 	fatalIf(createConfigDir(), "Unable to create configuration directory")
 
-	newObject, err := newGatewayLayer(backendType, endPoint, accessKey, secretKey, secure)
+	newObject, err := newGatewayLayer(backendType, ctx.Args()[1:])
```
harshavardhana, line 206: You can use `ctx.Args().Tail()` is more meaningful in this context and also pass in as `newGatewayLayer(backendType, cli.Args)`

harshavardhana: cli.Args have helpers which implement things like `cli.Args.First()`, `cli.Args.Get(0)`, `cli.Args.Get(1)`

harshavardhana: These are safer methods and they don't crash.. if there are less args on the CLI.

harshavardhana (changes requested):

donatello (approved):

cmd/gateway-gcs-layer.go
```diff
@@ -153,7 +161,20 @@ type gcsGateway struct {
...
 // newGCSGateway returns gcs gatewaylayer
-func newGCSGateway(endpoint string, projectID, secretKey string, secure bool) (GatewayLayer, error) {
+func newGCSGateway(args []string) (GatewayLayer, error) {
+	if len(args) != 1 {
+		return nil, fmt.Errorf("ProjectID expected")
+	}
+
+	profile := "default"
+	// TODO: don't know where to set profile yet
+	_ = profile
```
harshavardhana, line 171: I think there is no API or way right now, perhaps the way to customize this with GCS IAM [1] is that to use service account file. 

https://cloud.google.com/iam/docs/creating-managing-service-account-keys
```
gcloud iam service-accounts keys create \
    ~/key.json \
    --iam-account my-sa-123@my-project-123.iam.gserviceaccount.com
```

[1] - https://github.com/GoogleCloudPlatform/google-cloud-go#authorization

cmd/gateway-gcs-layer.go
```diff
@@ -162,14 +183,14 @@ func newGCSGateway(endpoint string, projectID, secretKey string, secure bool) (G
 		return nil, err
 	}
 
-	anonClient, err := minio.NewCore("storage.googleapis.com", "", "", secure)
+	anonClient, err := minio.NewCore(endpoint, "", "", secure)
```
harshavardhana, line 186: Azure uses basic http client for anonymous operations, we should use that instead here as well. Using two SDKs for the GCS backend is not a good idea IMHO.

harshavardhana (changes requested):

cmd/gateway-gcs-layer.go
nl5887: I agree, but this adds a lot of extra code where we will parse S3 XML output. Also we need to maintain the same code in and the minio-go client and here. Minio-go has been added as library already by the s3 gateway.

cmd/gateway-gcs-layer.go
```diff
@@ -197,10 +218,28 @@ func (l *gcsGateway) MakeBucket(bucket string) error {
 func (l *gcsGateway) MakeBucketWithLocation(bucket, location string) error {
 	bkt := l.client.Bucket(bucket)
 
-	// TODO: location
-	_ = location
+	// this will map s3 regions to google multi regions
+	if v, ok := map[string]string{
```
fwessels: Maybe just look at the first part of the AWS region, i.e. all 'ap's to asia, 'eu's to eu, etc.
This way it is a bit more resilient against future addtions of AWS regions.

fwessels (approved):

cmd/gateway-gcs-layer.go
nl5887: Yes, but we need to make exceptions then for ca- and sa-. Thought this was cleaner.

fwessels: I see, and ok.

harshavardhana:
Something failed on travis-ci @nl5887 can you take a look.. If the build is fixed we can merge this to gateway-gcs branch.

nl5887:
This is an interesting case, as the linting is not right here. The other if branches are passing, I'll see if I can rewrite the code or make a PR for the linting.

deekoder:
@nl5887 - Can you fix the travis test. We cannot approve the PR till it passes.

harshavardhana (approved):

cmd/gateway-azure.go
```diff
@@ -123,10 +125,22 @@ func azureToObjectError(err error, params ...string) error {
 }
 
 // Inits azure blob storage client and returns AzureObjects.
-func newAzureLayer(endPoint string, account, key string, secure bool) (GatewayLayer, error) {
+func newAzureLayer(args []string) (GatewayLayer, error) {
+	endPoint, secure, err := parseGatewayEndpoint(args[0])
```
vadmeste, line 129: If each gateway will take care of its arguments, you need to check if `len(args) > 0` here

cmd/gateway-s3.go
```diff
@@ -98,12 +100,23 @@ type s3Gateway struct {
 }
 
 // newS3Gateway returns s3 gatewaylayer
-func newS3Gateway(endpoint string, accessKey, secretKey string, secure bool) (GatewayLayer, error) {
+func newS3Gateway(args []string) (GatewayLayer, error) {
+	endpoint, secure, err := parseGatewayEndpoint(args[0])
```
vadmeste, line 104: You need to check if args has elements as described above.

harshavardhana:
The PR is merged @vadmeste send fixes to the feature-gcs branch.

PR #4309, opened 2017-05-09
web-handler: Allow anonymous download of zip

### LINKED ISSUE #4230: Browser: Object download is not working in guest mode
Object download is not working in guest mode

## Expected Behavior
Objects should be downloadable in guest mode 

## Current Behavior
When trying to download one or multiple objects from the browser, it fires an error message in the console 
```
Failed to load resource: the server responded with a status of 403 (Forbidden)  ..../minio/zip?token=undefined 
```

1. Open Browser in guest mode.
2. Select one or more objects.
3. Click 'Download all as zip' from the top action header.

## Your Environment
* Version used: 2017-03-15T17:49:51Z
* Operating System and version: macOS Sierra 10.12.4

krishnasrinivas:
## Description

Allow download of zip if download policy is set.

## Motivation and Context

fixes #4230

## How Has This Been Tested?

manually

## Types of changes

## Checklist:

### DIFF
--- a/cmd/web-handlers.go
+++ b/cmd/web-handlers.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"net/http"
 	"os"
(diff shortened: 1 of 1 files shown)


cmd/web-handlers.go
```diff
@@ -569,17 +569,22 @@ func (web *webAPIHandlers) DownloadZip(w http.ResponseWriter, r *http.Request) {
...
-	if !isAuthTokenValid(token) {
-		writeWebErrorResponse(w, errAuthentication)
-		return
-	}
 	var args DownloadZipArgs
 	decodeErr := json.NewDecoder(r.Body).Decode(&args)
 	if decodeErr != nil {
 		writeWebErrorResponse(w, decodeErr)
 		return
 	}
```
harshavardhana, line 580: We might have to add some comment here on why we are validating the token later i.e we are reading from the body that the client sent and then failing. Perhaps we can also do a `io.LimitReader(r.Body)` on the maximum allowed.. body header for this one. So that the server doesn't crash due to some erroneous handcrafted client which can happen now. Since token is validated later..

poornas (approved):
Reviewed and tested locally.

harshavardhana (approved):

PR #4315, opened 2017-05-10
gateway: Use default params when no args provided

### LINKED ISSUE #4314: [gateway] Crash if no arguments are provided
## Expected Behavior
Using default parameters when no argument is provided when running Gateway s3 and Azure

## Current Behavior
Crash:
```
$ MINIO_ACCESS_KEY="minio" MINIO_SECRET_KEY="minio123" minio gateway s3
panic: runtime error: index out of range

goroutine 1 [running]:
panic(0xe2c920, 0xc420016150)
	/home/vadmeste/work/go/src/runtime/panic.go:500 +0x1a1
github.com/minio/minio/cmd.newS3Gateway(0xc42000c350, 0x0, 0x0, 0x2, 0xc420179b60, 0x1, 0xf86c5a)
	/home/vadmeste/work/gospace/src/github.com/minio/minio/cmd/gateway-s3.go:104 +0x3e0
github.com/minio/minio/cmd.newGatewayLayer(0x7fffec26af5d, 0x2, 0xc42000c350, 0x0, 0x0, 0x0, 0x0, 0x0, 0x5)
	/home/vadmeste/work/gospace/src/github.com/minio/minio/cmd/gateway-main.go:122 +0x8a
github.com/minio/minio/cmd.gatewayMain(0xc4202ec160)
	/home/vadmeste/work/gospace/src/github.com/minio/minio/cmd/gateway-main.go:206 +0x1fb
github.com/minio/minio/vendor/github.com/minio/cli.HandleAction(0xdcd8c0, 0x10004f8, 0xc4202ec160, 0xc420183200, 0x0)
	/home/vadmeste/work/gospace/src/github.com/minio/minio/vendor/github.com/minio/cli/app.go:499 +0x7c
github.com/minio/minio/vendor/github.com/minio/cli.Command.Run(0xf6190d, 0x7, 0x0, 0x0, 0x0, 0x0, 0x0, 0xf7b0d4, 0x1d, 0x0, ...)
	/home/vadmeste/work/gospace/src/github.com/minio/minio/vendor/github.com/minio/cli/command.go:214 +0xb9a
github.com/minio/minio/vendor/github.com/minio/cli.(*App).Run(0xc4202b2680, 0xc42000c330, 0x3, 0x3, 0x0, 0x0)
	/home/vadmeste/work/gospace/src/github.com/minio/minio/vendor/github.com/minio/cli/app.go:260 +0x816
github.com/minio/minio/cmd.Main(0xc42000c330, 0x3, 0x3)
	/home/vadmeste/work/gospace/src/github.com/minio/minio/cmd/main.go:143 +0x4d
main.main()
	/home/vadmeste/work/gospace/src/github.com/minio/minio/main.go:68 +0x7b
```

## Possible Solution
Small fix in the code

1. `MINIO_ACCESS_KEY="access" MINIO_SECRET_KEY="secret" minio gateway s3`

## Your Environment
Detected in gcs-feature branch
All environments

vadmeste:
## Description
For S3 & Azure, use default parameters when no arguments (endpoint) are
provided. This also avoids a crash.

## Motivation and Context
Fixes #4314 

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:

### DIFF
--- a/cmd/gateway-azure.go
+++ b/cmd/gateway-azure.go
@@ -32,6 +32,7 @@ import (
 	"time"
 
 	"github.com/Azure/azure-sdk-for-go/storage"
+	"github.com/minio/cli"
 	"github.com/minio/minio-go/pkg/policy"
 	"github.com/minio/sha256-simd"
 )
@@ -125,14 +126,20 @@ func azureToObjectError(err error, params ...string) error {
 }
 
 // Inits azure blob storage client and returns AzureObjects.
-func newAzureLayer(args []string) (GatewayLayer, error) {
-	endPoint, secure, err := parseGatewayEndpoint(args[0])
-	if err != nil {
-		return nil, err
-	}
+func newAzureLayer(args cli.Args) (GatewayLayer, error) {
 
-	if endPoint == "" {
-		endPoint = storage.DefaultBaseURL
+	var err error
+
+	// Default endpoint parameters
+	endPoint := storage.DefaultBaseURL
+	secure := true
+
+	// If user provided some parameters
+	if len(args) > 0 {
+		endPoint, secure, err = parseGatewayEndpoint(args.First())
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	account := os.Getenv("MINIO_ACCESS_KEY")
--- a/cmd/gateway-gcs-layer.go
+++ b/cmd/gateway-gcs-layer.go
@@ -33,6 +33,7 @@ import (
 	"google.golang.org/api/googleapi"
 	"google.golang.org/api/iterator"
 
+	"github.com/minio/cli"
 	minio "github.com/minio/minio-go"
 	"github.com/minio/minio-go/pkg/policy"
 )
@@ -161,7 +162,7 @@ type gcsGateway struct {
 }
 
 // newGCSGateway returns gcs gatewaylayer
-func newGCSGateway(args []string) (GatewayLayer, error) {
+func newGCSGateway(args cli.Args) (GatewayLayer, error) {
 	if len(args) != 1 {
 		return nil, fmt.Errorf("ProjectID expected")
 	}
@@ -173,7 +174,7 @@ func newGCSGateway(args []string) (GatewayLayer, error) {
 	endpoint := "storage.googleapis.com"
 	secure := true
 
-	projectID := args[0]
+	projectID := args.First()
 
 	ctx := context.Background()
 
--- a/cmd/gateway-s3.go
+++ b/cmd/gateway-s3.go
@@ -25,6 +25,7 @@ import (
 
 	"encoding/hex"
 
+	"github.com/minio/cli"
 	minio "github.com/minio/minio-go"
 	"github.com/minio/minio-go/pkg/policy"
 )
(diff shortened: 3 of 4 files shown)


cmd/gateway-azure.go
```diff
@@ -126,12 +126,18 @@ func azureToObjectError(err error, params ...string) error {
...
+	var secure bool
+	var err error
+
+	if len(args) > 0 {
+		endPoint, secure, err = parseGatewayEndpoint(args[0])
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		// Provide default parameters if no arguments were provided
```
krishnasrinivas: @vadmeste this style is more concise:

```
setting := defaultSetting
if settingShouldChange {
    setting = newSetting
}
```

vadmeste: Done,

I made another change which is enabling secure by default for Azure, does that look correct ?

krishnasrinivas: @vadmeste yes

krishnasrinivas (approved):

harshavardhana (approved):

cmd/gateway-azure.go
```diff
@@ -126,13 +126,18 @@ func azureToObjectError(err error, params ...string) error {
...
-	}
 
-	if endPoint == "" {
-		endPoint = storage.DefaultBaseURL
+	var err error
+
+	// Default endpoint parameters
+	endPoint := storage.DefaultBaseURL
+	secure := true
+
```
harshavardhana: Actually if you passed around cl.Args helper we won't have this crash issue at all.

harshavardhana (changes requested):

cmd/gateway-azure.go
vadmeste: > Actually if you passed around cl.Args helper we won't have this crash issue at all.

Yes, are you suggesting to pass ctx.Args() instead of []string in gateway init ?

harshavardhana: Yes

vadmeste: Done

cmd/gateway-azure.go
```diff
@@ -125,14 +126,19 @@ func azureToObjectError(err error, params ...string) error {
...
+	var err error
 
-	if endPoint == "" {
-		endPoint = storage.DefaultBaseURL
+	// Default endpoint parameters
+	endPoint := storage.DefaultBaseURL
+	secure := true
+
+	if len(args) > 0 {
```
harshavardhana: There is no need to do this.. it won't crash. args.Get(0) is safe, returns empty if not found.  In-fact just use `args.First()`

cmd/gateway-gcs-layer.go
```diff
@@ -173,7 +174,7 @@ func newGCSGateway(args []string) (GatewayLayer, error) {
 	endpoint := "storage.googleapis.com"
 	secure := true
 
-	projectID := args[0]
+	projectID := args.Get(0)
```
harshavardhana: Use `args.First()`

cmd/gateway-s3.go
```diff
@@ -100,15 +101,20 @@ type s3Gateway struct {
...
+	var err error
+
+	// Default endpoint parameters
+	endpoint := "s3.amazonaws.com"
+	secure := true
 
-	if endpoint == "" {
-		endpoint = "s3.amazonaws.com"
-		secure = true
+	if len(args) > 0 {
```
harshavardhana: Same as above.

harshavardhana (changes requested):

cmd/gateway-azure.go
vadmeste: > There is no need to do this.. it won't crash. args.Get(0) is safe, returns empty if not found. In-fact just use args.First()

yes but we don't like to call parseGatewayEndpoint() with empty string. I think we need to keep len(args) and use only args.First().

harshavardhana: Just call it with empty string and let parseGatewayEndpoint() return an error. there is no harm in that.

vadmeste: Updated with args.First(). Let's see as discussed if we need to check if there is another way to provide default parameters.

harshavardhana (approved):

PR #4319, opened 2017-05-11
fs/erasure: Rename meta 'md5Sum' as 'etag'.

### LINKED ISSUE #4312: erasure/backend: Change xl.json meta key from md5Sum to etag
Current `xl.json` user defined meta contains `md5Sum` key.

```
  "meta": {
    "md5Sum": "d41d8cd98f00b204e9800998ecf8427e-12"
  },
```

It should be changed such that to reflect it is ETag instead of MD5Sum. 

```
  "meta": {
    "etag": "d41d8cd98f00b204e9800998ecf8427e-12"
  },
```

This requires a format change such that we are moving from current `xl.json` version i.e `1.0.0` to the newer format of `1.0.1` which would cater for this change. 

Existing files under `1.0.0` should be allowed to be read properly, if re-written should be written with the newer `1.0.1` format.

harshavardhana:
## Description
This PR also does backend format change to 1.0.1
from 1.0.0.  Backward compatible changes are still
kept to read the 'md5Sum' key. But all new objects
will be stored with the same details under 'etag'.

## Motivation and Context
Fixes #4312

## How Has This Been Tested?
Manually using `go test`

## Types of changes

## Checklist:

### DIFF
--- a/cmd/xl-v1-metadata.go
+++ b/cmd/xl-v1-metadata.go
@@ -146,7 +146,10 @@ type xlMetaV1 struct {
 // XL metadata constants.
 const (
 	// XL meta version.
-	xlMetaVersion = "1.0.0"
+	xlMetaVersion = "1.0.1"
+
+	// XL meta version.
+	xlMetaVersion100 = "1.0.0"
 
 	// XL meta format string.
 	xlMetaFormat = "xl"
@@ -173,7 +176,38 @@ func newXLMetaV1(object string, dataBlocks, parityBlocks int) (xlMeta xlMetaV1) 
 // IsValid - tells if the format is sane by validating the version
 // string and format style.
 func (m xlMetaV1) IsValid() bool {
-	return m.Version == xlMetaVersion && m.Format == xlMetaFormat
+	return isXLMetaValid(m.Version, m.Format)
+}
+
+// Verifies if the backend format metadata is sane by validating
+// the version string and format style.
+func isXLMetaValid(version, format string) bool {
+	return ((version == xlMetaVersion || version == xlMetaVersion100) &&
+		format == xlMetaFormat)
+}
+
+// Converts metadata to object info.
+func (m xlMetaV1) ToObjectInfo(bucket, object string) ObjectInfo {
+	objInfo := ObjectInfo{
+		IsDir:           false,
+		Bucket:          bucket,
+		Name:            object,
+		Size:            m.Stat.Size,
+		ModTime:         m.Stat.ModTime,
+		ContentType:     m.Meta["content-type"],
+		ContentEncoding: m.Meta["content-encoding"],
+	}
+
+	// Extract etag from metadata.
+	objInfo.ETag = extractETag(m.Meta)
+
+	// etag/md5Sum has already been extracted. We need to
+	// remove to avoid it from appearing as part of
+	// response headers. e.g, X-Minio-* or X-Amz-*.
+	objInfo.UserDefined = cleanMetaETag(m.Meta)
+
+	// Success.
+	return objInfo
 }
 
 // objectPartIndex - returns the index of matching object part number.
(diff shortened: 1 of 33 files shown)


cmd/xl-v1-object.go
```diff
@@ -108,14 +108,10 @@ func (xl xlObjects) CopyObject(srcBucket, srcObject, dstBucket, dstObject string
...
-			MD5Sum:          xlMeta.Meta["md5Sum"],
 			ContentType:     xlMeta.Meta["content-type"],
 			ContentEncoding: xlMeta.Meta["content-encoding"],
 		}
-		// md5Sum has already been extracted into objInfo.MD5Sum.  We
-		// need to remove it from xlMetaMap to avoid it from appearing as
-		// part of response headers. e.g, X-Minio-* or X-Amz-*.
-		delete(xlMeta.Meta, "md5Sum")
+		objInfo.ETag = extractETag(xlMeta.Meta)
 		objInfo.UserDefined = xlMeta.Meta
```
vadmeste: Doesn't UserDefined need etag cleaning here ?

vadmeste (commented):
Manually tested, +1

cmd/xl-v1-object.go
harshavardhana: The reason it works fine is that the info is not used. But we should clean it.

cmd/xl-v1-object.go
```diff
@@ -108,15 +108,11 @@ func (xl xlObjects) CopyObject(srcBucket, srcObject, dstBucket, dstObject string
...
 			ContentType:     xlMeta.Meta["content-type"],
 			ContentEncoding: xlMeta.Meta["content-encoding"],
 		}
-		// md5Sum has already been extracted into objInfo.MD5Sum.  We
-		// need to remove it from xlMetaMap to avoid it from appearing as
-		// part of response headers. e.g, X-Minio-* or X-Amz-*.
-		delete(xlMeta.Meta, "md5Sum")
-		objInfo.UserDefined = xlMeta.Meta
+		objInfo.ETag = extractETag(xlMeta.Meta)
+		objInfo.UserDefined = cleanMetaETag(xlMeta.Meta)
```
krishnasrinivas: xlMetaV1.ToObjectInfo() just like fsMetaV1.ToObjectInfo() will look cleaner

harshavardhana: Yeah correct...

cmd/xl-v1-metadata.go
```diff
@@ -173,7 +176,8 @@ func newXLMetaV1(object string, dataBlocks, parityBlocks int) (xlMeta xlMetaV1)
 // IsValid - tells if the format is sane by validating the version
 // string and format style.
 func (m xlMetaV1) IsValid() bool {
```
krishnasrinivas, line 178: similar check is needed for fs too. i.e older binaries should fail reading from newer fs.json

krishnasrinivas: Running older binaries on newer files does not fail for FS before this commit.

harshavardhana: There is no problem with that.  We are going to read both we shouldn't fail. We don't fail in xl either.  We should read all files.. in newer version. 

We only fail for format.json because directory structure was not recognized.

harshavardhana: One caveat is that etag will be missing if users run older binaries with newer file..

harshavardhana:
@vadmeste @krishnasrinivas any more reviews?

vadmeste (approved):
Reviewed and tested with FS/XL while removing some parts.

cmd/xl-v1-metadata.go
krishnasrinivas: > One caveat is that etag will be missing if users run older binaries with newer file..

exactly this is the reason it should have failed. All the clients which depend on etag will be affected. There can be more dangerous consequences which we can not predict at this time. You can confirm with AB about this if you'd like.

harshavardhana: Yes but you are not supposed to depend on ETag in the first place. Let's discuss with @abperiasamy . We need to do this for erasure backend as well.

harshavardhana: Actually i added this verification.. @krishnasrinivas . You are right it is necessary.

harshavardhana:
@krishnasrinivas @vadmeste i added corrupted metadata verification for both erasure and fs backend. Can you review again?

vadmeste (approved):
LGTM

krishnasrinivas (approved):

PR #4327, opened 2017-05-11
Feature gcs

nl5887:
Closes several issues


cmd/gateway-gcs.go
```diff
@@ -631,6 +633,9 @@ func (l *gcsGateway) ListMultipartUploads(bucket string, prefix string, keyMarke
 }
 
 func fromGCSMultipartKey(s string) (key, uploadID string, partID int, err error) {
+	// remove prefixes
+	s = filepath.Base(s)
```
vadmeste: I saw the code filepath.Base() and it seems it relies on os.IsPathSeparator ('/' for unix and '\' on Windows). As I understood, it seems it doesn't fill well in this case if we want windows support.

harshavardhana: Yes it won't we need to always use path.Base() not filepath.  All incoming object names are expected to be separated by "/" not by "\\" even on windows.

cmd/gateway-gcs.go
```diff
@@ -779,26 +779,30 @@ func (l *gcsGateway) ListObjectParts(bucket string, key string, uploadID string,
 
 // AbortMultipartUpload aborts a ongoing multipart upload
 func (l *gcsGateway) AbortMultipartUpload(bucket string, key string, uploadID string) error {
-	prefix := fmt.Sprintf("%s/multipart-%s-%s", ZZZZMinioPrefix, key, uploadID)
-	delimiter := "/"
 
-	// delete part zero, ignoring errors here, we want to clean up all remains
-	_ = l.client.Bucket(bucket).Object(toGCSMultipartKey(key, uploadID, 0)).Delete(l.ctx)
+	delimiter := "/"
```
harshavardhana: We already have this variable.. as `slashSeparator`  use it.

cmd/gateway-gcs.go
```diff
@@ -779,26 +779,30 @@ func (l *gcsGateway) ListObjectParts(bucket string, key string, uploadID string,
...
+	prefix := fmt.Sprintf("%s/multipart-%s-%s", ZZZZMinioPrefix, key, uploadID)
```
harshavardhana: Use `pathJoin()`  helper.

harshavardhana (changes requested):

cmd/gateway-gcs.go
```diff
@@ -601,18 +659,32 @@ func fromGCSMultipartKey(s string) (key, uploadID string, partID int, err error)
...
+	s = strings.Replace(s, "-", "%2D", -1)
+	return s
+}
 
+func toGCSMultipartKey(key string, uploadID string, partID int) string {
 	// parts are allowed to be numbered from 1 to 10,000 (inclusive)
-	return fmt.Sprintf("%s/multipart-%s-%s-%05d", ZZZZMinioPrefix, key, uploadID, partID)
+
+	// we need to encode the key because of possible slashes
+	return fmt.Sprintf("%s/multipart-%s-%s-%05d", ZZZZMinioPrefix, escape(key), uploadID, partID)
```
harshavardhana: use `pathJoin(ZZZMinipPrefix, fmt.Sprintf("multipart-%s-%s-%05d", escape(key), uploadID, partID)` instead.

cmd/gateway-gcs.go
```diff
@@ -601,18 +659,32 @@ func fromGCSMultipartKey(s string) (key, uploadID string, partID int, err error)
...
 }
 
 // NewMultipartUpload - upload object in multiple parts
 func (l *gcsGateway) NewMultipartUpload(bucket string, key string, metadata map[string]string) (uploadID string, err error) {
 	// generate new uploadid
 	uploadID = mustGetUUID()
+	uploadID = strings.Replace(uploadID, "-", "", -1)
```
harshavardhana, line 666: You can put this into a function and get a different UUID representation. @nl5887

cmd/gcs/gateway-gcs-url.go
```diff
@@ -0,0 +1,1053 @@
...
+		if shouldEscape(c, mode) {
+			if c == ' ' && mode == encodeQueryComponent {
+				spaceCount++
+			} else {
+				hexCount++
+			}
+		}
+	}
+
+	if spaceCount == 0 && hexCount == 0 {
```
harshavardhana: why do you need this @nl5887 ?

harshavardhana (changes requested):

cmd/gcs/gateway-gcs-url.go
nl5887: Has been removed already, doing it differently.

harshavardhana: Cool.

harshavardhana:
You might have to rebase @nl5887

nl5887:
@harshavardhana your opinion about the bucket translation? The reason / advantage of doing this is to make all s3 compatible libraries compatible with this feature. Libraries that don't support custom regions, won't work otherwise.

harshavardhana:
> @harshavardhana your opinion about the bucket translation? The reason / advantage of doing this is to make all s3 compatible libraries compatible with this feature. Libraries that don't support custom regions, won't work otherwise.

@krishnasrinivas @nl5887 - looks like the advantage of doing region mapping for GCS is that it creates multi region buckets. Which in a way is nicer and also default recommended way to create bucket on the Google Cloud Console. 

Disadvantage is that it disallows doing anything else other than multi regions. We can see what is the default behavior that we can follow here with @abperiasamy

nl5887:
@harshavardhana @krishnasrinivas  Actually, we don't have that issue. The naming of google regions is different (https://cloud.google.com/storage/docs/bucket-locations), so you can choose specific regions (like us-central1).

harshavardhana:
> @harshavardhana @krishnasrinivas Actually, we don't have that issue. The naming of google regions is different (https://cloud.google.com/storage/docs/bucket-locations), so you can choose specific regions (like us-central1).

Oh i see so the change here is done to allow for S3 specific region mapping to GCS.

krishnasrinivas:
> @harshavardhana your opinion about the bucket translation? The reason / advantage of doing this is to make all s3 compatible libraries compatible with this feature. Libraries that don't support custom regions, won't work otherwise.

@nl5887 S3 client libs will not be tied to regions else they will not be "forward" compatible with new regions being added to AWS.

The thought of doing the translation makes sense when you are doing it for object metadata (i.e x-amz-meta-key should be translated to x-ms-meta-key)

But doing so for regions has cons and no pros.
cons:
* if GCS introduces new region `australia` then it won't have a corresponding AWS region
* GCS gateway users have to be aware of this map to create region specific buckets

If AWS had specifications on the region string format (like `xx-yyyy-N`) then we would have to do the translation and we would have no choice, but since region is just a string, using GCS regions itself will be better.

nl5887:
@krishnasrinivas There are all kind of legacy applications, that should want to migrate to Minio but cannot because they cannot modify the software (anymore). Anyhow, I think enough has been said about this, time for @harshavardhana and @abperiasamy to decide on this.

krishnasrinivas:
Discussed with @abperiasamy the mapping should be there only for default region. i.e `us-east-1` -> `us` because clients can call makebucket with default region `us-east-1`

The s3-libs are only tied to default region (when regions are not specified) and no other regions.
Legacy applications should never have hardcoded region settings, if indeed we come across such applications, then we can re-visit this.

harshavardhana (approved):

PR #4329, opened 2017-05-11
signature: use region from Auth header if server's region not configured

krishnasrinivas:
## Description

This change allows server/gateway to be run with empty region, in which case for signature calculation we use the region specified in the Authz header.
If server/gateway is run with non-empty region then this configured region is used for signature calculation.

## Motivation and Context

S3 Gateway needs the ability to handle buckets from multiple regions. Hence gateway is run without any region setting.

## How Has This Been Tested?

`go test` and manual testing

## Types of changes

## Checklist:

### DIFF
--- a/cmd/bucket-notification-utils.go
+++ b/cmd/bucket-notification-utils.go
@@ -17,6 +17,7 @@
 package cmd
 
 import (
+	"errors"
 	"strings"
 
 	"github.com/minio/minio-go/pkg/set"
@@ -111,19 +112,20 @@ func checkARN(arn, arnType string) APIErrorCode {
 	if !strings.HasPrefix(arn, arnType) {
 		return ErrARNNotification
 	}
-	if !strings.HasPrefix(arn, arnType+serverConfig.GetRegion()+":") {
-		return ErrRegionNotification
-	}
-	account := strings.SplitN(strings.TrimPrefix(arn, arnType+serverConfig.GetRegion()+":"), ":", 2)
-	switch len(account) {
-	case 1:
-		// This means ARN is malformed, account should have min of 2elements.
+	strs := strings.SplitN(arn, ":", -1)
+	if len(strs) != 6 {
 		return ErrARNNotification
-	case 2:
-		// Account topic id or topic name cannot be empty.
-		if account[0] == "" || account[1] == "" {
-			return ErrARNNotification
+	}
+	if serverConfig.GetRegion() != "" {
+		region := strs[3]
+		if region != serverConfig.GetRegion() {
+			return ErrRegionNotification
 		}
+	}
+	accountID := strs[4]
+	resource := strs[5]
+	if accountID == "" || resource == "" {
+		return ErrARNNotification
 	}
 	return ErrNone
 }
@@ -258,28 +260,39 @@ func validateNotificationConfig(nConfig notificationConfig) APIErrorCode {
 // - webhook
 func unmarshalSqsARN(queueARN string) (mSqs arnSQS) {
 	mSqs = arnSQS{}
-	if !strings.HasPrefix(queueARN, minioSqs+serverConfig.GetRegion()+":") {
+	strs := strings.SplitN(queueARN, ":", -1)
+	if len(strs) != 6 {
 		return mSqs
 	}
-	sqsType := strings.TrimPrefix(queueARN, minioSqs+serverConfig.GetRegion()+":")
-	switch {
-	case hasSuffix(sqsType, queueTypeAMQP):
+	if serverConfig.GetRegion() != "" {
+		region := strs[3]
+		if region != serverConfig.GetRegion() {
+			return mSqs
+		}
+	}
+	sqsType := strs[5]
+	switch sqsType {
+	case queueTypeAMQP:
 		mSqs.Type = queueTypeAMQP
-	case hasSuffix(sqsType, queueTypeNATS):
+	case queueTypeNATS:
 		mSqs.Type = queueTypeNATS
-	case hasSuffix(sqsType, queueTypeElastic):
+	case queueTypeElastic:
 		mSqs.Type = queueTypeElastic
-	case hasSuffix(sqsType, queueTypeRedis):
+	case queueTypeRedis:
 		mSqs.Type = queueTypeRedis
-	case hasSuffix(sqsType, queueTypePostgreSQL):
+	case queueTypePostgreSQL:
 		mSqs.Type = queueTypePostgreSQL
-	case hasSuffix(sqsType, queueTypeMySQL):
+	case queueTypeMySQL:
 		mSqs.Type = queueTypeMySQL
-	case hasSuffix(sqsType, queueTypeKafka):
+	case queueTypeKafka:
 		mSqs.Type = queueTypeKafka
-	case hasSuffix(sqsType, queueTypeWebhook):
+	case queueTypeWebhook:
 		mSqs.Type = queueTypeWebhook
+	default:
+		errorIf(errors.New("invalid SQS type"), "SQS type: %s", sqsType)
 	} // Add more queues here.
-	mSqs.AccountID = strings.TrimSuffix(sqsType, ":"+mSqs.Type)
+
+	mSqs.AccountID = strs[4]
+
 	return mSqs
 }
--- a/cmd/bucket-notification-utils_test.go
+++ b/cmd/bucket-notification-utils_test.go
@@ -259,11 +259,6 @@ func TestQueueARN(t *testing.T) {
 			queueARN: "arn:minio:sns:us-east-1:1:listen",
 			errCode:  ErrARNNotification,
 		},
-		// Invalid region 'us-west-1' in queue arn.
-		{
-			queueARN: "arn:minio:sqs:us-west-1:1:redis",
-			errCode:  ErrRegionNotification,
-		},
 		// Invalid queue name empty in queue arn.
 		{
 			queueARN: "arn:minio:sqs:us-east-1:1:",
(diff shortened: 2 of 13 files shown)


cmd/bucket-notification-utils_test.go
```diff
@@ -259,11 +259,6 @@ func TestQueueARN(t *testing.T) {
 			queueARN: "arn:minio:sns:us-east-1:1:listen",
 			errCode:  ErrARNNotification,
 		},
-		// Invalid region 'us-west-1' in queue arn.
-		{
-			queueARN: "arn:minio:sqs:us-west-1:1:redis",
-			errCode:  ErrRegionNotification,
```
harshavardhana, line 265: why remote this test?

cmd/bucket-notification-utils_test.go
```diff
@@ -337,11 +332,6 @@ func TestUnmarshalSQSARN(t *testing.T) {
 			queueARN: "",
 			Type:     "",
 		},
-		// Invalid region 'us-west-1' in queue arn.
-		{
-			queueARN: "arn:minio:sqs:us-west-1:1:redis",
```
harshavardhana, line 342: Why remove this test?

cmd/handler-utils.go
```diff
@@ -38,15 +38,18 @@ func parseLocationConstraint(r *http.Request) (location string, s3Error APIError
...
 }
 
 // Validates input location is same as configured region
 // of Minio server.
 func isValidLocation(location string) bool {
-	return serverConfig.GetRegion() == location
+	if serverConfig.GetRegion() != "" {
+		return serverConfig.GetRegion() == location
+	}
+	return true
```
harshavardhana: A one line would work here? 

```
return serverConfig.GetRegion() == "" || serverConfig.GetRegion() == location
```

cmd/signature-v4_test.go
```diff
@@ -54,14 +54,7 @@ func TestDoesPolicySignatureMatch(t *testing.T) {
 			},
 			expected: ErrInvalidAccessKeyID,
 		},
-		// (2) It should fail if the region is invalid.
-		{
-			form: http.Header{
-				"X-Amz-Credential": []string{fmt.Sprintf(credentialTemplate, accessKey, now.Format(yyyymmdd), "invalidregion")},
```
harshavardhana, line 60: Why remove this test?

cmd/signature-v4_test.go
```diff
@@ -135,21 +128,7 @@ func TestDoesPresignedSignatureMatch(t *testing.T) {
...
 			expected: ErrInvalidAccessKeyID,
 		},
-		// (2) Should fail with an invalid region.
-		{
-			queryParams: map[string]string{
-				"X-Amz-Algorithm":      signV4Algorithm,
-				"X-Amz-Date":           now.Format(iso8601Format),
-				"X-Amz-Expires":        "60",
-				"X-Amz-Signature":      "badsignature",
-				"X-Amz-SignedHeaders":  "host;x-amz-content-sha256;x-amz-date",
```
harshavardhana, line 145: Why remove this test?

harshavardhana (changes requested):

cmd/bucket-notification-utils_test.go
krishnasrinivas: Tests run in empty region and this test does not fail anymore, hence removed it.

cmd/handler-utils.go
krishnasrinivas: sure

cmd/bucket-notification-utils_test.go
harshavardhana: Well we can set the region and test it.. So that we know that case is covered as well.

krishnasrinivas: this case is already covered by other cases which uses `us-east-1` which does not get special treatment anymore

cmd/bucket-notification-utils_test.go
krishnasrinivas: this test gets covered by other test cases which uses `us-east-1` which is not treated specially anymore

cmd/signature-v4_test.go
krishnasrinivas: This was failing with ErrSignatureDoesNotMatch and but this test case was not looking like it was written for ErrSignatureDoesNotMatch. Hence it would cause confusion to keep it.

There is another test case for ErrSignatureDoesNotMatch.

cmd/signature-v4_test.go
krishnasrinivas: This test now returns `ErrUnsignedHeaders` but the following two tests are similar and tests for the same error. Hence removed this.

harshavardhana: Ok

cmd/signature-v4_test.go
harshavardhana: Yeah it would fail with signature does not match. since we allowed the region.. let's say region is not set on server and client sets it.. Then this error case shouldn't fail. So we need to figure out why it is reporting as such.

cmd/bucket-notification-utils_test.go
harshavardhana: Are you sure that there is a test case which covers are scenario where client sets a region which is different than the one set on the server?. I am not seeing it. @krishnasrinivas

cmd/bucket-notification-utils.go
```diff
@@ -111,20 +112,21 @@ func checkARN(arn, arnType string) APIErrorCode {
...
-	case 2:
-		// Account topic id or topic name cannot be empty.
-		if account[0] == "" || account[1] == "" {
-			return ErrARNNotification
+	}
+	if serverConfig.GetRegion() != "" {
+		region := strs[3]
+		if region != serverConfig.GetRegion() {
+			return ErrRegionNotification
 		}
```
harshavardhana, line 123: From codecoverage its visible that this area is not covered, basically we don't have tests where a region is set on the server.

cmd/bucket-notification-utils.go
```diff
@@ -258,28 +260,39 @@ func validateNotificationConfig(nConfig notificationConfig) APIErrorCode {
...
+	if len(strs) != 6 {
 		return mSqs
 	}
-	sqsType := strings.TrimPrefix(queueARN, minioSqs+serverConfig.GetRegion()+":")
-	switch {
-	case hasSuffix(sqsType, queueTypeAMQP):
+	if serverConfig.GetRegion() != "" {
+		region := strs[3]
+		if region != serverConfig.GetRegion() {
+			return mSqs
```
harshavardhana, line 270: Same codecoverage issue here.. because the equivalent tests have been removed..

harshavardhana (changes requested):

cmd/signature-v4_test.go
krishnasrinivas: it will fail, we are not even sending `X-Amz-Signature`

harshavardhana: Yes but the case i am talking about is when server has a region set and a malformed X-Amz-Credential is sent, we should send ErrInvalidRegion in that scenario. Such a case is not covered here. IMHO.

cmd/signature-v4-parser.go
```diff
@@ -69,9 +69,6 @@ func parseCredentialHeader(credElement string) (credentialHeader, APIErrorCode)
 	if e != nil {
 		return credentialHeader{}, ErrMalformedCredentialDate
 	}
-	if credElements[2] == "" {
-		return credentialHeader{}, ErrMalformedCredentialRegion
```
harshavardhana, line 73: For example this situation we can agree that we have to relax and we will never send malformed credential anymore.

harshavardhana (changes requested):

cmd/bucket-notification-utils_test.go
krishnasrinivas: @harshavardhana previously in tests server would run with region `us-east-1` not it runs with empty region. Because of this change:

```
--- a/cmd/globals.go
+++ b/cmd/globals.go
@@ -29,7 +29,7 @@ import (
 const (
        globalMinioCertExpireWarnDays = time.Hour * 24 * 30 // 30 days.
 
-       globalMinioDefaultRegion       = "us-east-1"
+       globalMinioDefaultRegion       = ""
        globalMinioDefaultOwnerID      = "minio"
        globalMinioDefaultStorageClass = "STANDARD"
        globalWindowsOSName            = "windows"
```

This is also causing the code coverage problem that you commented about.

We should have tests where server runs in non-empty region and clients fail when it does not use this region. I can fix the tests TestQueueARN and TestUnmarshalSQSARN to take care of the coverage issue. But we'll need some rework in test infra where it tests when server runs in non-empty region and clients signs using a different region

harshavardhana: Okay @krishnasrinivas

cmd/signature-v4_test.go
krishnasrinivas: yes, agreed. it will need changes to the way the tests are run (have a look at my other comment `previously in tests server would run with region us-east-1.....`)

harshavardhana (dismissed):

vadmeste:
@krishnasrinivas, can you fix the startup print message and shows `<none>` when no region is set ?

harshavardhana:
> @krishnasrinivas, can you fix the startup print message and shows <none> when no region is set ?

in-fact @abperiasamy asked that Region to be not even printed if not set.

harshavardhana:
@krishnasrinivas can you fix the CI build issue?

harshavardhana (dismissed):

vadmeste (approved):
Reviewed and basic operations with s2/v4 auth successfuly tested.

harshavardhana:
Build failure @krishnasrinivas 

```
Installing ineffassign: Done.
Running vet: cmd/signature-v4-parser_test.go:194: arg errorCodeResponse[testCase.expectedErrCode] for printf verb %s of wrong type: cmd.APIError
make: *** [vet] Error 1
```

harshavardhana (approved):

PR #4335, opened 2017-05-12
gateway: allow buckets to be created in regions requested by client

### LINKED ISSUE #4241: S3 gateway does not support accessing multiple regions
After starting the gateway in us-east-1 region, if we try to create a bucket in a different region, we get an error from mc.

## Expected Behavior

Region should pass through the gateway to the backend and the request should work

## Current Behavior

We get error.

## Possible Solution

Unknown.

1. Start gateway like:
```
MINIO_ACCESS_KEY="<redacted>" MINIO_SECRET_KEY="<redacted>" minio gateway s3 https://s3.amazonaws.com
```
2. Add alias s3gw for the gateway into `mc`. I used S3V2.
3. Run:
```
$ mc mb --region us-west-2 s3gw/miniostestbucket
mc: <ERROR> Unable to make bucket `s3gw/miniostestbucket`. Region does not match.
```

## Context

Testing

## Your Environment

* Version used (`minio version`): Latest master.
* Server type and version:
* Link to your project:

krishnasrinivas:
## Description

Allow buckets to be created in the client requested region in gateway mode.

## Motivation and Context

fixes #4241

## How Has This Been Tested?

manually

## Types of changes

## Checklist:

### DIFF
--- a/cmd/gateway-handlers.go
+++ b/cmd/gateway-handlers.go
@@ -644,15 +644,6 @@ func (api gatewayAPIHandlers) PutBucketHandler(w http.ResponseWriter, r *http.Re
 		return
 	}
 
-	// validating region here, because isValidLocationConstraint
-	// reads body which has been read already. So only validating
-	// region here.
-	serverRegion := serverConfig.GetRegion()
-	if serverRegion != location {
-		writeErrorResponse(w, ErrInvalidRegion, r.URL)
-		return
-	}
-
 	bucketLock := globalNSMutex.NewNSLock(bucket, "")
 	bucketLock.Lock()
 	defer bucketLock.Unlock()


cmd/gateway-handlers.go
```diff
@@ -644,15 +644,6 @@ func (api gatewayAPIHandlers) PutBucketHandler(w http.ResponseWriter, r *http.Re
 		return
 	}
 
-	// validating region here, because isValidLocationConstraint
-	// reads body which has been read already. So only validating
-	// region here.
-	serverRegion := serverConfig.GetRegion()
-	if serverRegion != location {
```
harshavardhana, line 651: We can use isValidLocation instead here. And return error properly. Since we need to honor if gateway region was set manually.

harshavardhana (changes requested):

cmd/gateway-handlers.go
krishnasrinivas: > Its a behavior problem @krishnasrinivas if user says that this is my region. We are ought to honor it IMHO. For the most part they won't set it. But we willfully ignoring what was set by user is not right.

I see what you mean, if user needs to create a bucket in specific region he should run the gateway in empty reason.

we can use isValidLocation from here:
https://github.com/minio/minio/pull/4329/commits/b58048f1e6ebe980160de40feb819c6a77a931c1

nl5887:
This one is already included in https://github.com/dutchcoders/minio/commit/0954addaad8a6395fca57de0b69d0296dc12b8f3, with fix as discussed. When region has been set, it will verify the region with set region, otherwise it will use the region in the request.

krishnasrinivas:
@nl5887 will close this PR then

PR #4349, opened 2017-05-15
Add new ReadFileWithVerify storage-layer API

### LINKED ISSUE #4236: Request an object by ranges of 5MB traffic between minio rose to +/- 600/5*600MB
Hi, I have a file of 600Mo, I request this file by ranges of 5Mo and it looks like the traffic between the minio is +/- 600/5*600Mo ....

```
{"version":"1.0.0","format":"xl","stat":{"size":610184874,"modTime":"2017-05-03T13:20:24.810639588Z"},"erasure":{"algorithm":"klauspost/reedsolomon/vandermonde","data":2,"parity":2,"blockSize":10485760,"index":4,"distribution":[3,4,1,2],"checksum":[{"name":"part.1","algorithm":"blake2b","hash":"c0e27462affda2512fbc7ed2edf3f3466e79dbb2391cd8f778ca6930e6af7b405d191f85e1c701bacada042f3529f65196166d4e9ff8351c62adbfa7340c2d33"},{"name":"part.2","algorithm":"blake2b","hash":"68a37018c5b379d50c5ce428b2256a32bb7cb222d279ed4751e8da4645076ee590f431dc6f1fc03a91fab0e797b93842ed39faefce50459834a238239c6bb35d"},{"name":"part.3","algorithm":"blake2b","hash":"38f6570629a3f645da83d742950078e4e386a2be1f2dbda5e818a5ec291288498acea7339869b11a818f046d84abc6c901b4d2814f91d30167f647bfd57adbe5"},{"name":"part.4","algorithm":"blake2b","hash":"2859055291d3104b6310c7436536a742ce5e5bea208c401fae544b73ebdcb1714e658e7591fbe769e2d4954eb605b3bf4919d11768e5ef134c42b493bd78dc02"},{"name":"part.5","algorithm":"blake2b","hash":"292c2f4cf4a0034f7998ca9f39ddc6247e58f7fd67da972bf347233bb593e4e21e3957297ac21101c9af279555b901756dd1d89d5deb621821fcaeefdd25bb41"},{"name":"part.6","algorithm":"blake2b","hash":"bc2174bf07052abf51766d10048215cc0fcfbfbb9a264ad1a9613e8b55aa8e4f75da48e412562728e0aacb833062df4b8e87f5ec9bacfdef7da3021b6b3a6e5c"},{"name":"part.7","algorithm":"blake2b","hash":"c818d362cbbbaf6c3238aaf26b98aad5843afe4ea87aa974417fa44acd84f7de2e8a6b5e7260d5c2506fec2d4832d1e383fa136d8143a4a4a2002e8b577c670e"},{"name":"part.8","algorithm":"blake2b","hash":"c52345ea710fbe005c5462a8d0e248e7c54e1a404630f4955bd7d84861dfa03bc336753c81d4a9adceba8c0a5273f3e817cba03f9a1b75acc16f13e3b7d0b308"},{"name":"part.9","algorithm":"blake2b","hash":"efde022268693378b89bbb856c4cf59725177ad21a2e4d07e533a4a3a4bed59da6659e8075533ef8db4f284621e4c004a45f7fe8a26bce0e8387b75107854df8"},{"name":"part.10","algorithm":"blake2b","hash":"0a7b323d618bd0f079cde5ecb82b3b73a901f9bc46791bc4742e5f81b38c57b9160dbec7cbfa458c5953d47b536d4929618ef28dfee61074729f07b9898e93a8"}]},"minio":{"release":"DEVELOPMENT.2017-04-27T19-43-28Z"},"meta":{"md5Sum":"e2508ace73dbf331c3cb3e3e95227cc5"},"parts":[{"number":1,"name":"part.1","etag":"","size":67108864},{"number":2,"name":"part.2","etag":"","size":67108864},{"number":3,"name":"part.3","etag":"","size":67108864},{"number":4,"name":"part.4","etag":"","size":67108864},{"number":5,"name":"part.5","etag":"","size":67108864},{"number":6,"name":"part.6","etag":"","size":67108864},{"number":7,"name":"part.7","etag":"","size":67108864},{"number":8,"name":"part.8","etag":"","size":67108864},{"number":9,"name":"part.9","etag":"","size":67108864},{"number":10,"name":"part.10","etag":"","size":6205098}]}
```

Using https://nginx.org/en/docs/http/ngx_http_slice_module.html with size 5M and it's a sequence from the begining to the end of the file. It should be fast and should not generate a traffic of an average of 700Mbits during more than a minute

## Expected Behavior
Expected to not transfer large amounts of data for only 5Mb worth of range reads.

## Current Behavior
Data transfer is large even for smaller range request.

## Possible Solution
investigate the GetObject() method. 

## Steps to Reproduce (for bug)
Reproduced by user @dcharbonnier 

## Context
Performance improvements.. 

## Your Environment
Current release.

donatello:
## Description

This is an enhancement to the XL/distributed-XL mode. FS mode is
unaffected.

The ReadFileWithVerify storage-layer call is similar to ReadFile with
the additional functionality of performing bit-rot checking. It
accepts additional parameters for a hashing algorithm to use and the
expected hex-encoded hash string.

This patch provides significant performance improvement because:

1. combines the step of reading the file (during
erasure-decoding/reconstruction) with bit-rot verification;

2. limits the number of file-reads; and

3. avoids transferring the file over the network for bit-rot
verification.

ReadFile API is implemented as ReadFileWithVerify with empty hashing
arguments.

Credits to AB and Harsha for the algorithmic improvement.

Fixes #4236.

## Motivation and Context

#4236 and performance improvement.

## How Has This Been Tested?

Tests pass. @harshavardhana benchmarked it to find a ~2X improvement on a 2 node, 16 disk distributed XL setup on GET requests.

## Types of changes

## Checklist:


cmd/posix.go
```diff
@@ -38,6 +41,20 @@ const (
 	maxAllowedIOError = 5
 )
 
+// hashMisMatchError - represents a bit-rot hash verification failure
+// error.
```
harshavardhana: Move this error to storage-errors.go ?

cmd/posix.go
```diff
@@ -571,19 +600,61 @@ func (s *posix) ReadFile(volume string, path string, offset int64, buf []byte) (
...
+		// Continue computing hash until end of file.
+		_, err = io.Copy(hasher, file)
+		if err != nil {
+			return 0, err
+		}
+
+		// Verify the computed hash
+		computedHash := hex.EncodeToString(hasher.Sum(nil))
+		if computedHash != expectedHash {
+			return 0, hashMismatchError{expectedHash, computedHash}
```
harshavardhana: Add a testcase..

cmd/storage-rpc-client.go
```diff
@@ -262,6 +262,35 @@ func (n *networkStorage) ReadFile(volume string, path string, offset int64, buff
 	return int64(len(result)), toStorageErr(err)
 }
 
+// ReadFileWithVerify - reads a file at remote path and fills the buffer.
+func (n *networkStorage) ReadFileWithVerify(volume string, path string, offset int64,
+	buffer []byte, algo HashAlgo, expectedHash string) (m int64, err error) {
```
harshavardhana, line 267: you need to add a test case for this..

harshavardhana (commented):
Can you add more tests?

cmd/erasure-readfile.go
```diff
@@ -125,11 +126,15 @@ func parallelRead(volume, path string, readDisks []StorageAPI, orderedDisks []St
...
-			if !bitRotVerify(index) {
-				// So that we don't read from this disk for the next block.
-				orderedDisks[index] = nil
-				return
+			// evaluate if we need to perform bit-rot checking
+			needBitRotVerification := true
+			if brVerifiers[index].isVerified {
+				needBitRotVerification = false
+				// if file has bit-rot, do not reuse disk
+				if !brVerifiers[index].hasNoBitRot {
```
fwessels: Consider renaming hasNoBitRot to simply hasBitRot?
Then we can simply do `if brVerifiers[index].hasBitRot` here instead of 'double' negation.
Also down in line 164 we don't need need to invert
(and the default false value would, for most cases, better reflect the real-world status)

cmd/posix.go
```diff
@@ -515,8 +532,20 @@ func (s *posix) ReadAll(volume, path string) (buf []byte, err error) {
...
-// ReadFile symantics are same as io.ReadFull
-func (s *posix) ReadFile(volume string, path string, offset int64, buf []byte) (n int64, err error) {
+// ReadFile semantics are same as io.ReadFull.
+func (s *posix) ReadFile(volume, path string, offset int64, buf []byte) (n int64, err error) {
+
+	return s.ReadFileWithVerify(volume, path, offset, buf, "", "")
+}
+
+// ReadFileWithVerify is the same as ReadFile but with hashsum
+// verification. The operation may fail if the hash verification
```
fwessels: The operation **will** fail if the verification fails, correct?

cmd/posix.go
```diff
@@ -571,19 +600,61 @@ func (s *posix) ReadFile(volume string, path string, offset int64, buf []byte) (
 		return 0, err
 	}
 
-	// Verify if its not a regular file, since subsequent Seek is undefined.
+	// Verify if its not a regular file, since subsequent Seek is
```
fwessels: its --> it is

cmd/posix.go
```diff
@@ -571,19 +600,61 @@ func (s *posix) ReadFile(volume string, path string, offset int64, buf []byte) (
...
 		return 0, errIsNotRegular
 	}
 
-	// Seek to requested offset.
-	_, err = file.Seek(offset, os.SEEK_SET)
-	if err != nil {
-		return 0, err
+	// If empty hash algo or an expected hash string is empty we
+	// do not perform hash verification.
+	needToHash := !(algo == HashAlgo("") || expectedHash == "")
```
fwessels: Isn't it an error if the either the algo is set or the expectedHash is set (but not both)? Currently this would be silently ignored.

fwessels (approved):
See comment, for the rest LGTM

cmd/posix.go
donatello: Done.

cmd/posix.go
donatello: Done.

cmd/storage-rpc-client.go
donatello: Done.

cmd/erasure-readfile.go
donatello: Done.

cmd/posix.go
donatello: Done.

cmd/posix.go
donatello: :+1:

cmd/posix.go
donatello: :+1:  good catch!

I've modified the behaviour now to skip hashing if and only if `expectedHash == ""`. If `algo == ""` then it defaults to using Blake2b hashing. This way there is no silent error.

fwessels: Good solution, this way it will not slip through.

cmd/posix.go
```diff
@@ -512,11 +514,30 @@ func (s *posix) ReadAll(volume, path string) (buf []byte, err error) {
...
+// fails.
+//
+// The `expectedHash` is the expected hex-encoded hash string for
+// verification. With an empty expected hash string, hash verification
+// is skipped. An empty HashAlgo defaults to `blake2b`.
+//
+// The function takes care to minimize the number of disk read
+// operations.
+func (s *posix) ReadFileWithVerify(volume, path string, offset int64, buf []byte,
+	algo HashAlgo, expectedHash string) (n int64, err error) {
```
balamurugana, line 539: Is it good idea to have only one `ReadFile(volume, path string, offset int64, buf []byte, algo HashAlgo, expectedHash string) (n int64, err error)` where `algo`, `expectedHash` are optional?

To interpret `expectedHash`, its required to know what hasher was used.  I would suggest  to have a type which bundles those values together.

balamurugana (changes requested):

cmd/posix.go
harshavardhana: ReadFile() just uses ReadWithVerify() without algo, a Verify suffix was added to indicate what it does.

donatello: @balamurugana in the most recent version, only `expectedHash` can be empty (and that would imply to skip bit-rot verification). The `algo` is always interpreted and will cause either blake2b or sha256 to be used (currently). So if `algo` is empty and `expectedHash` is non-empty, hash verification is performed (and an error will be returned if there is a mismatch). Is this better?

balamurugana: > So if algo is empty and expectedHash is non-empty, hash verification is performed (and an error will be returned if there is a mismatch). Is this better?

As algo is empty, how do we know what algo needs to be used to validate `expectedHash`?

balamurugana:
> ReadFile() just uses ReadWithVerify() without algo, a Verify suffix was added to indicate what it does.

Yes. My question is why one more `ReadWithVerify()` than adding those new optional params into `ReadFile()` itself

harshavardhana:
> Yes. My question is why one more ReadWithVerify() than adding those new optional params into ReadFile() itself

ReadFile() by itself is not explaining the purpose.. So it is better to name it with a purpose. ReadFile() doesn't indicate that.

cmd/posix.go
harshavardhana: @donatello we can make algo to be non-empty and error out in this scenario because since its a Read operation we should fail if we cannot find the right algo. Rather than write where we can pick the right algo in a platform specific way.

donatello: The `newHash` function used to create a hasher instance already does some interpretation of the algo string - so i thought it would be fine to interpret the algo - it is probably being used in this way in other parts of the code.

donatello: > As algo is empty, how do we know what algo needs to be used to validate expectedHash?

It is defaulting to blake2b because of the existing `newHash` function implementation. If that defaulting behaviour was ok earlier, it should be ok here too?

balamurugana: @donatello does it mean there is a chance that passed `expectedHash` may be verified by different algo?

donatello: I think if code clean-up is required, it can be taken up on a separate basis - I think combining it with every nearby change is a bit distracting on the main purpose of a change. Currently, I think it is not being clearly articulated what is wrong/can be better here.

harshavardhana: @donatello it is doing that because after checksum calculation it will fail anyways. checksum algo is always set for a Read operation so we should treat that as failure here

```
                        // Set checksum algo only once, while it is possible to have
                        // different algos per block because of our `xl.json`.
                        // It is not a requirement, set this only once for all the disks.
                        if ckSumAlgo == "" {
                                ckSumAlgo = ckSumInfo.Algorithm
                        }
```

donatello: Ok, I have changed this so that if `expectedHash == ""` and `algo` is invalid, then (a new) error is returned.

donatello:
Any further comments on this one? Comments so far are addressed.

harshavardhana (approved):

Issue #4345, opened 2017-05-13
Profile distributed Minio setup for performance bottlenecks


harshavardhana:
Distributed Minio Setup is going through a lot of changes 

- Offloading of bit-rot check to server side.
- Dsync related changes for timeout etc.

After some initial benchmarks and results - some bottlenecks have been observed.  Single node erasure code right now is better performing than distributed erasure code even in a 2 node setup with 20Gb bonded  10gig network. 

This issue is to profile and find these bottlenecks and open relevant Github issues.  @deekoder can you assign this to me. I will be taking over this task.

harshavardhana:
Preliminary results are here 

```
(pprof) top --cum
1.42days of 3.24days total (43.81%)
Dropped 299 nodes (cum <= 0.02days)
Showing top 10 nodes out of 107 (cum >= 1.24days)
      flat  flat%   sum%        cum   cum%
         0     0%     0%      2days 61.82%  runtime.goexit
  1.42days 43.81% 43.81%   1.42days 43.81%  sync.(*Mutex).Lock
         0     0% 43.81%   1.24days 38.18%  github.com/minio/minio/cmd.pathValidityHandler.ServeHTTP
         0     0% 43.81%   1.24days 38.18%  github.com/minio/minio/vendor/github.com/gorilla/mux.(*Router).ServeHTTP
         0     0% 43.81%   1.24days 38.18%  github.com/minio/minio/cmd.(*pathValidityHandler).ServeHTTP
         0     0% 43.81%   1.24days 38.18%  github.com/minio/minio/cmd.httpStatsHandler.ServeHTTP
         0     0% 43.81%   1.24days 38.18%  net/http.HandlerFunc.ServeHTTP
         0     0% 43.81%   1.24days 38.18%  github.com/minio/minio/cmd.(*httpStatsHandler).ServeHTTP
         0     0% 43.81%   1.24days 38.18%  github.com/minio/minio/cmd.requestSizeLimitHandler.ServeHTTP
         0     0% 43.81%   1.24days 38.18%  github.com/minio/minio/cmd.(*requestSizeLimitHandler).ServeHTTP
```

```
(pprof) top --cum
42.94hrs of 54.34hrs total (79.02%)
Dropped 324 nodes (cum <= 0.27hrs)
Showing top 10 nodes out of 110 (cum >= 10.48hrs)
      flat  flat%   sum%        cum   cum%
         0     0%     0%   43.90hrs 80.78%  runtime.goexit
[...]

PR #4352, opened 2017-05-15
auth/rpc: Add RWMutex instead of Mutex for granular locking.

harshavardhana:
Refer https://github.com/minio/minio/issues/4345

## Description
auth/rpc: Add RWMutex instead of Mutex for granular locking.

## Motivation and Context
Improvement in terms of overall locking region. 

## How Has This Been Tested?
Manually on a distributed cluster.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/auth-rpc-client.go
+++ b/cmd/auth-rpc-client.go
@@ -52,10 +52,10 @@ type authConfig struct {
 
 // AuthRPCClient is a authenticated RPC client which does authentication before doing Call().
 type AuthRPCClient struct {
-	sync.Mutex            // Mutex to lock this object.
-	rpcClient  *RPCClient // Reconnectable RPC client to make any RPC call.
-	config     authConfig // Authentication configuration information.
-	authToken  string     // Authentication token.
+	sync.RWMutex            // Mutex to lock this object.
+	rpcClient    *RPCClient // Reconnectable RPC client to make any RPC call.
+	config       authConfig // Authentication configuration information.
+	authToken    string     // Authentication token.
 }
 
 // newAuthRPCClient - returns a JWT based authenticated (go) rpc client, which does automatic reconnect.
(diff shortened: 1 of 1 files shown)


cmd/auth-rpc-client.go
```diff
@@ -80,13 +80,13 @@ func newAuthRPCClient(config authConfig) *AuthRPCClient {
...
-	authClient.Lock()
-	defer authClient.Unlock()
-
+	authClient.RLock()
 	// Return if already logged in.
 	if authClient.authToken != "" {
+		authClient.RUnlock()
 		return nil
 	}
+	authClient.RUnlock()
```
balamurugana: This makes login to be called in parallel by multiple goroutines.

What is the purpose of calling `authClient.rpcClient.Call(serviceMethod, &args, &reply);` in parallel?

In the profiling result of referred issue, I don't see a hit of this method.  As `Login()` is called only once upon connect/reconnect, why it is making performance penalty?

balamurugana (changes requested):

cmd/auth-rpc-client.go
harshavardhana: It is to use right kind of locks for reading authToken() currently they block on each other. when using Lock() instead of RLock().  We can make login RPC inside to be blocking.. But reading shouldn't be blocking.

fwessels (approved):
LGTM

cmd/auth-rpc-client.go
balamurugana: I understood `RLock()` is right thing for `call()`, but for `Login()` the whole method needs to be protected by `Lock()`

harshavardhana: I disagree @balamurugana we are reading authToken as well inside Login() so we need to add RLock() there. We need to only add Lock() where its needed which i just changed take a look again.

balamurugana: Now I see the `Lock()` is moved above `authClient.rpcClient.Call(loginMethod, &loginArgs, &loginReply);`.  This doesn't prevent parallel logins either ie the first thread is in process of login and other thread will do login once `Unlock()`.  I am still favor of the previous approach wrt `Login()`.

As you defer, I would recommend someone needs to give +1 wrt `Login()`.  Other changes are good to me.

balamurugana (commented):
LGTM other than `Login()`

cmd/auth-rpc-client.go
harshavardhana: I am refactoring by moving the authToken check outside.. of the Login's scope.

balamurugana: Basically we would need to prevent parallel `Login()`.  Would like to see how that will be prevented.

krishnasrinivas (approved):

cmd/auth-rpc-client.go
harshavardhana: Yes you are right there is no way to prevent this in the current code style, Login() is sort of mixed together with Login() as a separate function and also to be part of the call() it calls for racy situations. I have kept the code as is from previous work and just added RLock() for reading auth token.

Another idea that i explored which works is using a factory pattern for Login. By creating a loginFactory 

```
// loginFactory a JWT based authentication is performed with rpc server.
type loginFactory func() (token string, err error)

func newLoginFactory(rpcClient *RPCClient, config authConfig) loginFactory {
        var authToken string
        var mutex = &sync.Mutex{}
        return func() (string, error) {
                // Login should be attempted one at a time.
                //
                // The reason for large region lock here is
                // to avoid two simultaneous login attempts
                // racing over each other.
                //
                // #1 Login() gets the lock proceeds to login.
                // #2 Login() waits for the unlock to happen
                //    after login in #1.
                // #1 Successfully completes login saves the
                //    newly acquired token.
                // #2 Successfully gets the lock and proceeds,
                //    but since we have acquired the token
                //    already the call quickly returns.
                mutex.Lock()
                defer mutex.Unlock()

                fmt.Println("AuthToken Set", authToken != "")
                if authToken != "" {
                        return authToken, nil
                }

                // Login to authenticate and acquire a new auth token.
                var (
                        loginMethod = config.serviceName + loginMethodName
                        loginArgs   = LoginRPCArgs{
                                Username:    config.accessKey,
                                Password:    config.secretKey,
                                Version:     Version,
                                RequestTime: UTCNow(),
                        }
                        loginReply = LoginRPCReply{}
                )

                if err := rpcClient.Call(loginMethod, &loginArgs, &loginReply); err != nil {
                        return "", err
                }

                authToken = loginReply.AuthToken

                // Success.
                return authToken, nil

        }
}
```

But won't make this change for now. We will just use RLock() for authToken reading instead.

cmd/auth-rpc-client.go
```diff
@@ -78,51 +78,61 @@ func newAuthRPCClient(config authConfig) *AuthRPCClient {
...
-
-		// Do RPC call.
-		err = authClient.rpcClient.Call(serviceMethod, args, reply)
-	}
-	return err
+	if err = authClient.Login(); err != nil {
+		return err
+	} // On successful login, execute RPC call.
+
+	authClient.RLock()
```
donatello, line 129: I don't think this is correct. It is taking a read-lock to perform a write operation (the set-auth-token operation). Does it really work?

PR #4361, opened 2017-05-17
Fix and cleanup update message and improve related tests

### LINKED ISSUE #4232: Message : You are running an older version of Minio released 5 hours from now
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃ You are running an older version of Minio released 5 hours from now ┃
┃ Update: https://dl.minio.io/server/minio/release/darwin-amd64/minio ┃
┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
## Expected Behavior

No such message since I am running the test binaries. 

## Current Behavior
The above sentence does not sound intuitive.
 
## Possible Solution
 Don't know.

1. Type `minio gateway azure` with keys
 
## Context
It's ambiguous now for me because I think i need to update to the latest released version of minio. However since i am already using a release candidate, why does it ask me update. Perhaps it should know that the current version is > than the released versions and not show me this message.

## Your Environment
 
* Version used (`minio version`):  2017-04-28T19:24:01Z
* Environment name and version (e.g. nginx 1.9.1): macOS
* Server type and version: Test Build, Gateway
* Operating System and version (`uname -a`): 16.5.0 Darwin Kernel Version 16.5.0
* Link to your project:

donatello:
Fixes #4232

## How Has This Been Tested?

Manually and updated tests

## Types of changes

## Checklist:

### DIFF
--- a/cmd/server-main.go
+++ b/cmd/server-main.go
@@ -81,8 +81,8 @@ EXAMPLES:
 func checkUpdate(mode string) {
 	// Its OK to ignore any errors during getUpdateInfo() here.
 	if older, downloadURL, err := getUpdateInfo(1*time.Second, mode); err == nil {
-		if older > time.Duration(0) {
-			log.Println(colorizeUpdateMessage(downloadURL, older))
+		if updateMsg := computeUpdateMessage(downloadURL, older); updateMsg != "" {
+			log.Println(updateMsg)
 		}
 	}
 }
--- a/cmd/update-main.go
+++ b/cmd/update-main.go
@@ -238,13 +238,16 @@ func getDownloadURL() (downloadURL string) {
 	return minioReleaseURL + "minio"
 }
 
-func getUpdateInfo(timeout time.Duration, mode string) (older time.Duration, downloadURL string, err error) {
-	currentReleaseTime, err := GetCurrentReleaseTime()
+func getUpdateInfo(timeout time.Duration, mode string) (older time.Duration,
+	downloadURL string, err error) {
+
+	var currentReleaseTime, latestReleaseTime time.Time
+	currentReleaseTime, err = GetCurrentReleaseTime()
 	if err != nil {
 		return older, downloadURL, err
 	}
 
-	latestReleaseTime, err := getLatestReleaseTime(timeout, mode)
+	latestReleaseTime, err = getLatestReleaseTime(timeout, mode)
 	if err != nil {
 		return older, downloadURL, err
 	}
@@ -274,8 +277,8 @@ func mainUpdate(ctx *cli.Context) {
 		os.Exit(-1)
 	}
 
-	if older != time.Duration(0) {
-		log.Println(colorizeUpdateMessage(downloadURL, older))
+	if updateMsg := computeUpdateMessage(downloadURL, older); updateMsg != "" {
+		log.Println(updateMsg)
 		os.Exit(1)
 	}
 
--- a/cmd/update-notifier.go
+++ b/cmd/update-notifier.go
@@ -28,25 +28,39 @@ import (
 	"github.com/fatih/color"
 )
 
+// computeUpdateMessage - calculates the update message, only if a
+// newer version is available.
+func computeUpdateMessage(downloadURL string, older time.Duration) string {
+	if downloadURL == "" || older <= 0 {
+		return ""
+	}
+
+	// Compute friendly duration string to indicate time
+	// difference between newer and current release.
+	t := time.Time{}
+	newerThan := humanize.RelTime(t, t.Add(older), "ago", "")
+
+	// Return the nicely colored and formatted update message.
+	return colorizeUpdateMessage(downloadURL, newerThan)
+}
+
 // colorizeUpdateMessage - inspired from Yeoman project npm package https://github.com/yeoman/update-notifier
-func colorizeUpdateMessage(updateString string, newerThan time.Duration) string {
+func colorizeUpdateMessage(updateString string, newerThan string) string {
 	// Initialize coloring.
 	cyan := color.New(color.FgCyan, color.Bold).SprintFunc()
 	yellow := color.New(color.FgYellow, color.Bold).SprintfFunc()
 
-	// Calculate length without color coding, due to ANSI color
-	// characters padded to actual string the final length is wrong
-	// than the original string length.
-	newerThanStr := humanize.Time(UTCNow().Add(newerThan))
+	msgLine1Fmt := " You are running an older version of Minio released %s "
+	msgLine2Fmt := " Update: %s "
 
-	line1Str := fmt.Sprintf(" You are running an older version of Minio released %s ", newerThanStr)
-	line2Str := fmt.Sprintf(" Update: %s ", updateString)
-	line1Length := len(line1Str)
-	line2Length := len(line2Str)
+	// Calculate length *without* color coding: with ANSI terminal
+	// color characters, the result is incorrect.
+	line1Length := len(fmt.Sprintf(msgLine1Fmt, newerThan))
+	line2Length := len(fmt.Sprintf(msgLine2Fmt, updateString))
 
 	// Populate lines with color coding.
-	line1InColor := fmt.Sprintf(" You are running an older version of Minio released %s ", yellow(newerThanStr))
-	line2InColor := fmt.Sprintf(" Update: %s ", cyan(updateString))
+	line1InColor := fmt.Sprintf(msgLine1Fmt, yellow(newerThan))
+	line2InColor := fmt.Sprintf(msgLine2Fmt, cyan(updateString))
 
 	// calculate the rectangular box size.
 	maxContentWidth := int(math.Max(float64(line1Length), float64(line2Length)))
@@ -60,7 +74,7 @@ func colorizeUpdateMessage(updateString string, newerThan time.Duration) string 
 
 	// Box cannot be printed if terminal width is small than maxContentWidth
 	if maxContentWidth > termWidth {
-		return "\n" + line1InColor + "\n" + line2InColor + "\n" + "\n"
+		return "\n" + line1InColor + "\n" + line2InColor + "\n\n"
 	}
 
 	topLeftChar := "┏"
@@ -79,14 +93,11 @@ func colorizeUpdateMessage(updateString string, newerThan time.Duration) string 
 		vertBarChar = "|"
 	}
 
-	message := "\n"
-	// Add top line
-	message += yellow(topLeftChar+strings.Repeat(horizBarChar, maxContentWidth)+topRightChar) + "\n"
-	// Add message lines
-	message += vertBarChar + line1InColor + strings.Repeat(" ", maxContentWidth-line1Length) + vertBarChar + "\n"
-	message += vertBarChar + line2InColor + strings.Repeat(" ", maxContentWidth-line2Length) + vertBarChar + "\n"
-	// Add bottom line
-	message += yellow(bottomLeftChar+strings.Repeat(horizBarChar, maxContentWidth)+bottomRightChar) + "\n"
-
-	return message
+	lines := []string{
+		yellow(topLeftChar + strings.Repeat(horizBarChar, maxContentWidth) + topRightChar),
+		vertBarChar + line1InColor + strings.Repeat(" ", maxContentWidth-line1Length) + vertBarChar,
+		vertBarChar + line2InColor + strings.Repeat(" ", maxContentWidth-line2Length) + vertBarChar,
+		yellow(bottomLeftChar + strings.Repeat(horizBarChar, maxContentWidth) + bottomRightChar),
+	}
+	return "\n" + strings.Join(lines, "\n") + "\n"
 }
--- a/cmd/update-notifier_test.go
+++ b/cmd/update-notifier_test.go
@@ -17,7 +17,7 @@
 package cmd
 
 import (
-	"runtime"
+	"fmt"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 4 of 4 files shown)


cmd/update-notifier.go
```diff
@@ -28,25 +28,35 @@ import (
...
 
+// computeUpdateMessage - calculates the update message, only if a
+// newer version is available.
+func computeUpdateMessage(downloadURL string, oldVersion, newVersion time.Time) (string, bool) {
+	if !newVersion.After(oldVersion) || downloadURL == "" {
+		return "", false
+	}
+	// Last argument below is never actually used as (newVersion >
+	// oldVersion).
+	newerThanStr := humanize.RelTime(oldVersion, newVersion, "ago", "huh?")
```
harshavardhana: This `huh?` is not nice.. We need to find something else. In-fact when this happens we can throw an error i.e `return "", false` which kind of makes sense to me that no update is needed.. It can happen so that users can build it themselves using `make release` .

harshavardhana (changes requested):

cmd/update-notifier.go
donatello: > This huh? is not nice.. We need to find something else. In-fact when this happens we can throw an error i.e return "", false which kind of makes sense to me that no update is needed.. It can happen so that users can build it themselves using make release .

I think then it is better to switch back to `humanize.Time()`, since we are explicitly handling the "relative" aspects of the time. We are already returning from the function when `newVersion <= oldVersion`, so the `huh?` string is never used. Let me amend this patch.

donatello: My bad, can't use `humanize.Time()` as that is relative to the current time.

So I just made it an empty string - there is a comment explaining that the 4th argument is never used, so I think this is ok.

cmd/server-main.go
```diff
@@ -80,9 +80,9 @@ EXAMPLES:
 // Check for updates and print a notification message
 func checkUpdate(mode string) {
 	// Its OK to ignore any errors during getUpdateInfo() here.
-	if older, downloadURL, err := getUpdateInfo(1*time.Second, mode); err == nil {
-		if older > time.Duration(0) {
-			log.Println(colorizeUpdateMessage(downloadURL, older))
+	if currentReleaseTime, latestReleaseTime, downloadURL, err := getUpdateInfo(1*time.Second, mode); err == nil {
+		if updateMsg, isOld := computeUpdateMessage(downloadURL, currentReleaseTime, latestReleaseTime); isOld {
```
balamurugana: Could you share why `currentReleaseTime` and `latestReleaseTime` are preferred over `older`?

cmd/update-notifier.go
```diff
@@ -28,25 +28,35 @@ import (
 	"github.com/fatih/color"
 )
 
+// computeUpdateMessage - calculates the update message, only if a
+// newer version is available.
+func computeUpdateMessage(downloadURL string, oldVersion, newVersion time.Time) (string, bool) {
+	if !newVersion.After(oldVersion) || downloadURL == "" {
+		return "", false
```
balamurugana: You can treat `""` as false too.  no need to have two false values.

balamurugana (changes requested):

cmd/server-main.go
donatello: This was just to be able to give input to the `humanize.RelTime` function that handles finding friendly duration strings between two given times.

harshavardhana (dismissed):

cmd/server-main.go
balamurugana: You could do the same with `older` too.
```go
t1 := UTCNow()
t2 := t1.Add(-older)
humanize.RelTime(t1, t2)
```

As returned values are not used in this function, I would prefer `older`.  Moreover returning multiple arguments looks ugly to me in this case.

cmd/update-notifier.go
donatello: :+1:

cmd/server-main.go
donatello: Agreed! I have updated the PR.

donatello:
@balamurugana @harshavardhana Review comments addressed - please approve if there are no further comments.

cmd/update-notifier.go
```diff
@@ -28,25 +28,39 @@ import (
...
+// computeUpdateMessage - calculates the update message, only if a
+// newer version is available.
+func computeUpdateMessage(downloadURL string, older time.Duration) string {
+	if downloadURL == "" || older <= 0 {
+		return ""
+	}
+
+	// Compute friendly duration string to indicate time
+	// difference between newer and current release.
+	t := time.Now()
```
harshavardhana: You can use UTCNow() function helper here..

harshavardhana (changes requested):

cmd/update-notifier.go
balamurugana: technically no difference its going to make.

cmd/update-main.go
```diff
@@ -274,8 +277,8 @@ func mainUpdate(ctx *cli.Context) {
 		os.Exit(-1)
 	}
 
-	if older != time.Duration(0) {
-		log.Println(colorizeUpdateMessage(downloadURL, older))
+	if updateMsg := computeUpdateMessage(downloadURL, older); updateMsg != "" {
```
balamurugana, line 280: Technically `downloadURL` never be empty.  I prefer previous check is better.  But its your call.

cmd/update-notifier.go
```diff
@@ -60,7 +74,7 @@ func colorizeUpdateMessage(updateString string, newerThan time.Duration) string
 
 	// Box cannot be printed if terminal width is small than maxContentWidth
 	if maxContentWidth > termWidth {
-		return "\n" + line1InColor + "\n" + line2InColor + "\n" + "\n"
+		return fmt.Sprint("\n", line1InColor, "\n", line2InColor, "\n\n")
```
balamurugana: 1. Is it good idea to make
```
>
>  {line1InColor}
>  {line2InColor}
>
```
2. As all values are strings, I am not sure why `fmt.Sprint()` is used over `+`
3. You could use string array and join for better readability.

cmd/update-notifier.go
```diff
@@ -79,14 +93,11 @@ func colorizeUpdateMessage(updateString string, newerThan time.Duration) string
...
-	message += yellow(bottomLeftChar+strings.Repeat(horizBarChar, maxContentWidth)+bottomRightChar) + "\n"
-
-	return message
+	lines := []string{
+		yellow(fmt.Sprint(topLeftChar, strings.Repeat(horizBarChar, maxContentWidth), topRightChar)),
+		fmt.Sprint(vertBarChar, line1InColor, strings.Repeat(" ", maxContentWidth-line1Length), vertBarChar),
+		fmt.Sprint(vertBarChar, line2InColor, strings.Repeat(" ", maxContentWidth-line2Length), vertBarChar),
+		yellow(fmt.Sprint(bottomLeftChar, strings.Repeat(horizBarChar, maxContentWidth), bottomRightChar)),
+	}
+	return fmt.Sprint("\n", strings.Join(lines, "\n"), "\n")
```
balamurugana: Same as above

cmd/update-notifier_test.go
```diff
@@ -17,34 +17,46 @@
...
+		dlURL string
 
-	if !(strings.Contains(updateMsg, plainMsg) || strings.Contains(updateMsg, colorMsg)) {
-		t.Fatal("Duration string not found in colorized update message", updateMsg)
+		expectedSubStr string
+	}{
+		{72 * time.Hour, "my_download_url", "days"},
+		{3 * time.Hour, "https://my_download_url_is_huge/aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa", "hours"},
+		{-72 * time.Hour, "another_update_url", ""},
+		{0, "another_update_url", ""},
```
balamurugana: Can you add cases like `1 hour, 37 minutes`, `12 minutes, 37 seconds`, `1 second` etc?

balamurugana (changes requested):
What are the reasons to move away from `updateParams` struct?

cmd/update-notifier.go
harshavardhana: It is to be consistent with our internal usage of time. We only deal with UTC everywhere.. @balamurugana

cmd/update-notifier.go
donatello: I thought doing so many concatenations looking inefficient, so I used `fmt.Sprint` - but some simple benchmarking later, it appears the inefficiency is only apparent for a larger number of strings. I switched back to the `+` operation in this code. 

Also here, `strings.Join` is not making it more readable as there are `"\n"` near the beginning and end and extra and there are just two lines with content. So `+` seems fine here.

cmd/update-notifier.go
donatello: Also updated.

cmd/update-main.go
donatello: `downloadURL` can be empty -> whenever it is the latest release or a dev-build (with timestamp greater than most recent release).

cmd/update-notifier_test.go
donatello: Ok done. The humanize lib is not that smart. It always says the same thing. Also with 1 second the rel time func prints "now" - I guess this is because of some rounding issue. It is not a problem.

cmd/update-notifier.go
donatello: Done

donatello:
@balamurugana @harshavardhana Review comments addressed - please approve if there are no further comments.

donatello:
> What are the reasons to move away from `updateParams` struct?

@balamurugana This PR is distinct from https://github.com/minio/minio/pull/4213 - the question does not arise, since that is not what is being fixed in this PR. This PR's purpose is to fix the update message string. Once, this PR is merged, I will update the other one, since it will be on top of this change. I will also explain some of the decisions about simplifications that were done after discussion here with AB, Krishna and Harsha in that PR itself.

harshavardhana (dismissed):

cmd/update-main.go
balamurugana: The key is `older` and the decision needs to be taken based on that ie if `older` is non-zero, there should be `downloadURL`.  `getDownloadURL()` always returns value.

cmd/update-notifier_test.go
balamurugana: I know this problem.  thats why I am asking that corner case to be addressed properly.

balamurugana (changes requested):

cmd/update-notifier_test.go
donatello: Checking humanize souce code, this is not due to a rounding issue, but it is a deliberate choice in the library.

Introducing a test with 1 second time difference prints a message like below:

```
		┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
		┃ You are running an older version of Minio released now ┃
		┃ Update: my_download_url                                ┃
		┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
```

In our usage, `now` is printed when the time difference is 1 second (exactly - stable release and current release happening like this has a super low probability). Calling it `now` in the message is not a huge inaccuracy - it is ok. Trying to fix this requires using `humanize.CustomRelTime` with a `magnitudes` argument that is only slightly different from `humanize.defaultMagnitudes` (an unexposed library variable). This is excessively complex for a very minor issue (and I don't really see this as a problem at all).

I suggest ignoring this problem and moving on.

cmd/update-main.go
donatello: The `older` value is being used in the `computeUpdateMessage` to return an appropriate `updateMsg`:

```
	if downloadURL == "" || older <= 0 {
		return ""
	}
```

(would it help if condition were written as `older <= 0 || downloadURL == ""`? I don't think its a big difference - we are checking downloadURL == "" just as an additional (sanity) check).

 If we check `older` outside here again, and then conditionally call `computeUpdateMessage` you are just introducing another 0 value in a different way that you already said could be avoided. I think this is fine.

cmd/update-notifier_test.go
balamurugana: I am not able to understand the problem of adding this into the unit test.

donatello: Oh, you said to address the problem. You seemed to mean that the behavior is not correct

donatello: Also adding this case in unit test is a special case code that just enforces the `now` printing behavior; this is unnecessary. It is not adding anything valuable and it is just making tests complicated to read and maintain. It is not good.

balamurugana: The test is not for `now`,  the test is about how corner cases are handled.  If the `go-humanize` behavior changes in future, this case will make test failure.

donatello: Refactored tests and made it exhaustive now - though i don't believe it is useful (see comment in main thread below).

donatello:
@balamurugana @harshavardhana Refactored tests, and avoided calling `UTCNow()` using instead the zero value for time - this is a bit more nice.

On the topic of testing, this change has become more about writing tests for `go-humanize` than our logic. It has been a relatively pointless exercise - fears about `go-humanize` changing in the future are unfounded because the package is frozen in our code via vendoring. We need to focus more on testing complicated pieces of our application logic than this. I hope we can have a larger discussion about these aspects soon.

If there are no other _glaring_ issues (and appealing for more pragmatism), could we please approve this change and move on?

cmd/update-notifier.go
```diff
@@ -28,25 +28,39 @@ import (
...
+// newer version is available.
+func computeUpdateMessage(downloadURL string, older time.Duration) string {
+	if downloadURL == "" || older <= 0 {
+		return ""
+	}
+
+	// Compute friendly duration string to indicate time
+	// difference between newer and current release.
+	t := time.Time{}
+	newerThan := humanize.RelTime(t, t.Add(older), "ago", "")
```
harshavardhana, line 41: This makes sense.. :+1:

harshavardhana (dismissed):

balamurugana:
> On the topic of testing, this change has become more about writing tests for go-humanize than our logic. It has been a relatively pointless exercise - fears about go-humanize changing in the future are unfounded because the package is frozen in our code via vendoring. We need to focus more on testing complicated pieces of our application logic than this. I hope we can have a larger discussion about these aspects soon.
> 

The test cases are not for `go-humanize`, but for 
```go
// Compute friendly duration string
```

The tests are still checking for a string `minute`, `days` etc which is not right.   Previously it was tested properly for only one case.

As I commented  `> Can you add cases like 1 hour, 37 minutes, 12 minutes, 37 seconds, 1 second etc?` some durations are rounded and some are not.  This makes everyone to understand how exactly the update message printed now and future.

> If there are no other glaring issues (and appealing for more pragmatism), could we please approve this change and move on?

My whole intention is quality of the code by unit tests and functional tests.  PR doing additions/modifications should have proper unit tests and functional tests for whatever functions it does irrespective of the changes it made.

harshavardhana:
@balamurugana do you have further concerns or can this be merged?

balamurugana:
> @balamurugana do you have further concerns or can this be merged?

In my previous comment, I mentioned in detail about why current unit test is not doing right things.

harshavardhana:
> In my previous comment, I mentioned in detail about why current unit test is not doing right things.

I thought those are fixed in the updated tests? @balamurugana - can you check again?

balamurugana:
> I thought those are fixed in the updated tests? @balamurugana - can you check again?

I am referring below.  Am I missing something?

```
The tests are still checking for a string minute, days etc which is not right. Previously it was tested properly for only one case.
 
As I commented > Can you add cases like 1 hour, 37 minutes, 12 minutes, 37 seconds, 1 second etc? some durations are rounded and some are not. This makes everyone to understand how exactly the update message printed now and future.
```

harshavardhana:
> I am referring below. Am I missing something?

Okay will wait for @donatello to respond.

harshavardhana:
> Okay will wait for @donatello to respond.

@donatello anything that you can add to what @balamurugana is asking..

harshavardhana:
> @donatello anything that you can add to what @balamurugana is asking..

What is pending on this one? @donatello

donatello:
PR updated.

balamurugana (approved):

harshavardhana (approved):

PR #4373, opened 2017-05-19
add traefik as load balancer for swarm

morph027:
## Description

Added [Træfik](https://traefik.io/) load balancer to distributed swarm which allows us to access swarm internal minio using a generic name and also expose one port only (Træfik's port) which does the internal routing to minio's then.

## Motivation and Context

I could not access minio using a generic name from inside swarm (other containers), as each service is named `minio_x`. Adding this, i can have another container attached to network `minio_distributed` accessing minio at `http://minioproxy:80` and Træfik together with swarm makes sure it always is accessible. Bonus is we can expose only a single port.

## How Has This Been Tested?

* create attachable network

```
docker network create --attachable --driver overlay minio_distributed
```

* Spin up minio stack

```
docker stack deploy -c minio/distributed/docker-compose.yml minio
```

* On one of my swarm nodes, i've attached mc client

```
docker run --rm -it --network minio_distributed --entrypoint=/bin/sh minio/mc
mc config host add myminio http://minioproxy:80 foo bar
```

## Types of changes

- [x] contrib/3rdparty improvement

## Checklist:

### DIFF
--- a/docs/orchestration/docker-swarm/README.md
+++ b/docs/orchestration/docker-swarm/README.md
@@ -57,7 +57,7 @@ docker stack rm minio_stack
 
 * By default the services use `local` volume driver. Refer to [Docker documentation](https://docs.docker.com/compose/compose-file/#/volume-configuration-reference) to explore further options.
 
-* Minio services in the Docker compose file expose ports 9001 to 9004. This allows multiple services to run on a host. Explore other configuration options in [Docker documentation](https://docs.docker.com/compose/compose-file/#/ports).
+* Minio services in the Docker compose file will be loadbalanced using [Træfik](https://traefik.io/) which is exposing port 9000.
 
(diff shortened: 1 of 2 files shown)


nitisht:
@morph027 thank you for the contribution. This is a great idea and will be surely helpful for the users. 

I have just one concern, since this is our official documentation on how to run Minio on Docker Swarm, we stayed away from adding any external dependency in the compose file. 

I think better place for this will be in the [Minio Cookbook](https://github.com/minio/cookbook). Could you please add this to our cookbook?

vadmeste:
> I think better place for this will be in the Minio Cookbook.

I agree.

morph027:
Ok, will do :+1: Probably going to ask in Slack channel next time before :grin:

nitisht:
Thanks a lot @morph027 .

PR #4376, opened 2017-05-19
Validate gateway arguments

### LINKED ISSUE #4355: gateway: endpoint and address command line arguments need to properly validated
Currently endpoint and address arguments from command line are not validated properly.  below cases need to be addressed
1. address argument should be valid with port and points to local host only and endpoint should be non-empty.
2. host in endpoint should be remote host
3. for testing, host in endpoint can be local host whereas it should not be using same port as gateway accordingly to binding. For example, `localhost` and `myhostname` points to same local host.  Valid command line argument is
  - `minio gateway --address=:9000 http://localhost:9000` should fail as gateway can not bind to port 9000 in localhost as its already taken by `http://localhost:9000` S3 service.
  - `minio gateway --address=:9000 http://localhost:10000` should work
  - `minio gateway --address=localhost:9000 http://myhostname:9000` should work as both remote S3 and gateway bind to different interfaces

See also https://github.com/minio/minio/issues/4307

## Expected Behavior

Cases in description should be handled accordingly.

1.
2.
3.
4.

## Your Environment

* Server type and version:
* Link to your project:

krishnasrinivas:
## Description

gateway endpoints validation.

## Motivation and Context

fixes #4355

## How Has This Been Tested?

manually

## Types of changes

## Checklist:

### DIFF
--- a/cmd/gateway-main.go
+++ b/cmd/gateway-main.go
@@ -240,7 +240,7 @@ func gatewayMain(ctx *cli.Context) {
 	serverAddr := ctx.String("address")
 
 	err := validateGatewayArguments(serverAddr, endpointAddr)
-	fatalIf(err, "Unable to validate")
+	fatalIf(err, "Invalid argument")
 
 	// Second argument is endpoint.	If no endpoint is specified then the
 	// gateway implementation should use a default setting.
--- a/cmd/gateway-main.go
+++ b/cmd/gateway-main.go
@@ -23,6 +23,7 @@ import (
 	"github.com/minio/cli"
 	"net/url"
 	"os"
+	"runtime"
 	"strings"
 )
 
@@ -205,8 +206,18 @@ func gatewayMain(ctx *cli.Context) {
 	backendType := ctx.Args().Get(0)
 	// Second argument is the endpoint address (optional)
 	endpointAddr := ctx.Args().Get(1)
-	// Third argument is the address flag
+
 	serverAddr := ctx.String("address")
+
+	fatalIf(CheckLocalServerAddr(serverAddr), "Invalid address %s in command line argument.", serverAddr)
+	if runtime.GOOS == "darwin" {
+		host, port := mustSplitHostPort(serverAddr)
+		// On macOS, if a process already listens on LOCALIPADDR:PORT, net.Listen() falls back
+		// to IPv6 address ie minio will start listening on IPv6 address whereas another
+		// (non-)minio process is listening on IPv4 of given port.
+		// To avoid this error sutiation we check for port availability only for macOS.
+		fatalIf(checkPortAvailability(host), "Port %d already in use", port)
+	}
 
 	if endpointAddr != "" {
 		// Reject the endpoint if it points to the gateway handler itself.
--- a/cmd/gateway-main.go
+++ b/cmd/gateway-main.go
@@ -180,6 +180,36 @@ func parseGatewayEndpoint(arg string) (endPoint string, secure bool, err error) 
 	}
 }
 
+// Validate gateway arguments.
+func validateGatewayArguments(serverAddr, endpointAddr string) error {
+	if err := CheckLocalServerAddr(serverAddr); err != nil {
+		return err
+	}
+
+	if runtime.GOOS == "darwin" {
+		_, port := mustSplitHostPort(serverAddr)
+		// On macOS, if a process already listens on LOCALIPADDR:PORT, net.Listen() falls back
+		// to IPv6 address i.e minio will start listening on IPv6 address whereas another
+		// (non-)minio process is listening on IPv4 of given port.
+		// To avoid this error situation we check for port availability only for macOS.
+		if err := checkPortAvailability(port); err != nil {
+			return err
+		}
+	}
+
+	if endpointAddr != "" {
+		// Reject the endpoint if it points to the gateway handler itself.
+		sameTarget, err := sameLocalAddrs(endpointAddr, serverAddr)
+		if err != nil {
+			return err
+		}
+		if sameTarget {
+			return errors.New("endpoint points to the local gateway")
+		}
+	}
+	return nil
+}
+
 // Handler for 'minio gateway'.
 func gatewayMain(ctx *cli.Context) {
 	if !ctx.Args().Present() || ctx.Args().First() == "help" {
@@ -209,24 +239,8 @@ func gatewayMain(ctx *cli.Context) {
 
 	serverAddr := ctx.String("address")
 
-	fatalIf(CheckLocalServerAddr(serverAddr), "Invalid address %s in command line argument.", serverAddr)
-	if runtime.GOOS == "darwin" {
-		host, port := mustSplitHostPort(serverAddr)
-		// On macOS, if a process already listens on LOCALIPADDR:PORT, net.Listen() falls back
-		// to IPv6 address ie minio will start listening on IPv6 address whereas another
-		// (non-)minio process is listening on IPv4 of given port.
-		// To avoid this error sutiation we check for port availability only for macOS.
-		fatalIf(checkPortAvailability(host), "Port %d already in use", port)
-	}
-
-	if endpointAddr != "" {
-		// Reject the endpoint if it points to the gateway handler itself.
-		sameTarget, err := sameLocalAddrs(endpointAddr, serverAddr)
-		fatalIf(err, "Unable to compare server and endpoint addresses.")
-		if sameTarget {
-			fatalIf(errors.New("endpoint points to the local gateway"), "Endpoint url is not allowed")
-		}
-	}
+	err := validateGatewayArguments(serverAddr, endpointAddr)
+	fatalIf(err, "Unable to validate")
 
 	// Second argument is endpoint.	If no endpoint is specified then the
 	// gateway implementation should use a default setting.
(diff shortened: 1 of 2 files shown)


balamurugana:
> Do you mean when myhostname is the same server as the gateway is being run but bound to different interface? I think it is OK not to allow it because code will be complicated to allow it to run, but more importantly allowing such a setup would be an erroneous setup as there is no reason for a user to run it that way and would mean that user is doing a mistake.

As I mentioned in the issue, it is test setup, not some user is going to use it that way.  I am OK if such kind of test setup is not allowed.

balamurugana:
> We'll allow endpoint to be localhost (so that it is easy for testing) but same port for both gateway and server is not allowed (fixed by anis)

`minio gateway --address=localhost:9000 http://myhostname:9000` is allowed to work.  Expected is it should not.

krishnasrinivas:
> minio gateway --address=localhost:9000 http://myhostname:9000 is allowed to work. Expected is it should not.

why @balamurugana ? I am unable to see what I am missing.

balamurugana:
>> minio gateway --address=localhost:9000 http://myhostname:9000 is allowed to work. Expected is it should not.
>>
> why @balamurugana ? I am unable to see what I am missing.

the case is `localhost` and `myhostname` resolve to same local host but different IPs.  This is not a valid case as per your comment in the description.

krishnasrinivas:
> the case is localhost and myhostname resolve to same local host but different IPs. This is not a valid case as per your comment in the description.

But if myhostname resolves to localhost then anis' patch already takes care of not allowing the setup.

balamurugana:
> But if myhostname resolves to localhost then anis' patch already takes care of not allowing the setup.

Looks like its doing, but no unit tests show that.  That needs to be added.  `TestCreateEndpoints()` uses a technique to find non-loop back IPs of local host.

harshavardhana:
Can you fix the commit message typo @krishnasrinivas

vadmeste (approved):
Tested (on linux)

cmd/gateway-main.go
```diff
@@ -188,9 +189,19 @@ func gatewayMain(ctx *cli.Context) {
...
 	endpointAddr := ctx.Args().Get(1)
-	// Third argument is the address flag
+
 	serverAddr := ctx.String("address")
 
+	fatalIf(CheckLocalServerAddr(serverAddr), "Invalid address %s in command line argument.", serverAddr)
+	if runtime.GOOS == "darwin" {
+		host, port := mustSplitHostPort(serverAddr)
+		// On macOS, if a process already listens on LOCALIPADDR:PORT, net.Listen() falls back
+		// to IPv6 address ie minio will start listening on IPv6 address whereas another
```
krisis: s/ie/i.e

cmd/gateway-main.go
```diff
@@ -188,9 +189,19 @@ func gatewayMain(ctx *cli.Context) {
...
+		// (non-)minio process is listening on IPv4 of given port.
+		// To avoid this error sutiation we check for port availability only for macOS.
```
krisis: s/sutiation/situation

krisis (changes requested):
- Please fix the typo in title of the PR.
- PR's description looks more like a comment. You could record it as a comment [here](https://github.com/minio/minio/issues/4355). The description should describe the changes and how it fixes the issue.

balamurugana (changes requested):
Needs to add unit test to address case
`minio gateway --address=localhost:9000 http://mylocalhostname:9000` where `localhost` and `mylocalhostname` point different IPs of local hosts.

krishnasrinivas:
> PR's description looks more like a comment. You could record it as a comment here. The description should describe the changes and how it fixes the issue.

Oops, I don't know how it ended up in the description. It was intended to be a comment in the issue.

krishnasrinivas:
@balamurugana @krisis please check now

cmd/gateway-main_test.go
```diff
@@ -48,3 +51,34 @@ func TestParseGatewayEndpoint(t *testing.T) {
...
+
+	testCases := []struct {
+		serverAddr   string
+		endpointAddr string
+		valid        bool
+	}{
+		{":9000", "http://localhost:9001", true},
+		{"123.123.123.123:9000", "http://localhost:9000", false},
+		{":9000", "http://localhost:9000", false},
+		{":9000", nonLoopBackIP + ":9000", false},
```
balamurugana: I would recommend to validate error messages than `valid` bool.

cmd/gateway-main_test.go
```diff
@@ -48,3 +51,34 @@ func TestParseGatewayEndpoint(t *testing.T) {
...
+		t.Fatalf("No non-loop back IP address found for this host")
+	}
+	nonLoopBackIP := nonLoopBackIPs.ToSlice()[0]
+
+	testCases := []struct {
+		serverAddr   string
+		endpointAddr string
+		valid        bool
+	}{
+		{":9000", "http://localhost:9001", true},
```
balamurugana, line 93: you could add a case where `endpointAddr` is a remote host.  you could use `example.org` domains.

balamurugana (changes requested):
Test cases need a fix.

cmd/gateway-main_test.go
```diff
@@ -62,23 +62,24 @@ func TestValidateGatewayArguments(t *testing.T) {
 	}
 	nonLoopBackIP := nonLoopBackIPs.ToSlice()[0]
 
+	errMsg := "endpoint points to the local gateway"
+
 	testCases := []struct {
 		serverAddr   string
 		endpointAddr string
-		valid        bool
+		errMsg       stringB
```
harshavardhana: `stringB` is invalid, compilation issue here.

cmd/gateway-main_test.go
```diff
@@ -62,23 +62,24 @@ func TestValidateGatewayArguments(t *testing.T) {
 	}
 	nonLoopBackIP := nonLoopBackIPs.ToSlice()[0]
 
+	errMsg := "endpoint points to the local gateway"
```
harshavardhana: You can re-purpose the message like `errEndpointLoop` ? which is returned by the actual function and you are just using errEndpointLoop.Error() to compare. Naked strings like this can invite issues like typo etc avoids changing at multiple places.

harshavardhana (changes requested):

cmd/gateway-main_test.go
krishnasrinivas: we had a discussion here about this bala with @harshavardhana  , the problem with doing string comparison is the code becomes brittle, i.e if we change the error message string the test case breaks. So I have kept it as "valid bool" for now. But we can have a discussion in one of the meetings so that we can consistently follow one method

krishnasrinivas: we had a discussion here about this bala with @harshavardhana  , the problem with doing string comparison is the code becomes brittle, i.e if we change the error message string the test case breaks. So I have kept it as "valid bool" for now. But we can have a discussion in one of the meetings so that we can consistently follow one method

krishnasrinivas:
> I would recommend to validate error messages than valid bool.

@balamurugana  had a discussion here about this with @harshavardhana , the problem with doing string comparison is the code becomes brittle, i.e if we change the error message string the test case breaks. So I have kept it as "valid bool" for now. But we can have a discussion in one of the meetings so that we can consistently follow it everywhere in our code.

harshavardhana (dismissed):

cmd/gateway-main.go
```diff
@@ -188,17 +219,11 @@ func gatewayMain(ctx *cli.Context) {
...
-	if endpointAddr != "" {
-		// Reject the endpoint if it points to the gateway handler itself.
-		sameTarget, err := sameLocalAddrs(endpointAddr, serverAddr)
-		fatalIf(err, "Unable to compare server and endpoint addresses.")
-		if sameTarget {
-			fatalIf(errors.New("endpoint points to the local gateway"), "Endpoint url is not allowed")
-		}
-	}
+	err := validateGatewayArguments(serverAddr, endpointAddr)
+	fatalIf(err, "Unable to validate")
```
krisis: "Unable to validate" is not correct. We did validate and found the arguments invalid. We could say something like "Invalid arguments: custom endpoint and gateway addresses are identical."

cmd/gateway-main_test.go
```diff
@@ -48,3 +51,35 @@ func TestParseGatewayEndpoint(t *testing.T) {
...
+		{":9000", "http://localhost:9001", true},
+		{":9000", "http://google.com", true},
+		{"123.123.123.123:9000", "http://localhost:9000", false},
+		{":9000", "http://localhost:9000", false},
+		{":9000", nonLoopBackIP + ":9000", false},
+	}
+	for i, test := range testCases {
+		err := validateGatewayArguments(test.serverAddr, test.endpointAddr)
+		if test.valid && err != nil {
+			t.Errorf("Test %d expected not to return erro but got %s", i+1, err)
```
krisis: typo: s/erro/error

krisis (changes requested):

cmd/gateway-main.go
```diff
@@ -179,6 +180,36 @@ func parseGatewayEndpoint(arg string) (endPoint string, secure bool, err error)
 	}
 }
 
+// Validate gateway arguments.
+func validateGatewayArguments(serverAddr, endpointAddr string) error {
+	if err := CheckLocalServerAddr(serverAddr); err != nil {
```
harshavardhana, line 185: @krisis can you take a look.. ?

krisis: @harshavardhana yes, this addresses the problem my PR fixes. You can close my PR in favour.

harshavardhana (approved):

Issue #4367, opened 2017-05-18
Investigate dynamic timeouts


fwessels:
Investigate how to best adapt (global) timeouts during the course of the running of the application.

That is, based on how closely or widely a timeout is missed, either increase or decrease the initial timeout value.

PR #4377, opened 2017-05-20
Add support for timeouts for locks

fwessels:
This is a (p)review of timeout support for the namespace-lock. In addition to an updated `dsync`, also a new package `lsync` is introduced. Both packages support the same API for locks that can time out.

When a timeout occurs this is propagated back up  so that an error can be communicated to the client instead of the current behaviour where long/infinite hangs occur.

The default timeout value has been set at 10 minutes, we may want to use different timeout values in different places (or even simulate a blocking lock by passing in 'indefinite' in some cases).

Preliminary testing has been done, but more testing needs to be done.


balamurugana:
@fwessels It would be very helpful why `sync.{RW}Mutex` does not fit in cases where `lsync` is better to go WRT minio?  Unfortunately I don't see any issue# related to this PR and `lsync`

fwessels:
@bala the default sync package has just blocking behaviour which is in most cases (in combination with spinning up a Go routine) a good way of handling concurrent locks.

In our case it can however lead to "indefinite" locks when e.g. quorum is missing or there is a long or stale lock. By allowing locks to time out we can properly report an error back to the client. See for instance #3185.

harshavardhana:
Locking is a critical piece in the stack - @deekoder this PR requires a separate testing cycle not just PR review and local testing, just like `feature-gcs` to validate this in all necessary conditions. 

Especially to go through [chaos](https://github.com/minio/chaos), [mint](https://github.com/minio/mint), [perftest](https://github.com/minio/perftest) etc.

deekoder:
ok @harshavardhana are we blocked from beginning review ?

harshavardhana:
> ok @harshavardhana are we blocked from beginning review ?

Reviewing point of view the PR is fine and it is doing what is expected from the originally discussed idea, what essentially needs to be done is validating this at scale like how we did for distributed server testing on perhaps a good number of servers.

It is also required that @abperiasamy takes a look at this and provides his feedback on the type of timeouts we are using, should they be made configurable under various scenarios etc. 

- Object API calls should have lower timeout than 10minutes default? say in matter of 30secs?
- Healing API calls should have higher timeout or defaulted to 10minutes? 
- Listing, Stat operations should have lower timeout than even 30secs? like 10secs? 

We need to discuss and decide on this.

fwessels:
Also ties in to https://github.com/minio/minio/issues/4367 (dynamic timeouts)

deekoder:
Marking it blocked since we have to extensively test after the release.

fwessels:
Rebased this PR on master.

## `dynamicTimeout`

Other changes include the addition of `dynamicTimeout` for dynamically adapting timeout values. A `dynamicTimeout` object is instantiated with an initial value and a minimum timeout value. Based on the actual results of using the timeout value, it can be increased or lowered, like for instance in: 

```
func (li *lockInstance) GetLock(timeout *dynamicTimeout) (timedOutErr error) {
	lockSource := getSource()
	start := UTCNow()
	readLock := false
	if !li.ns.lock(li.volume, li.path, lockSource, li.opsID, readLock, timeout.Timeout()) {
		timeout.LogFailure()
		return OperationTimedOut{Path: li.path}
	}
	timeout.LogSuccess(UTCNow().Sub(start))
	return
}
```

`LogFailure` means that a lock has not been acquired (and thus the timeout value can be considered too short), in case a lock is acquired the actual time is passed in logged via `LogSuccess`. After registering a series of either failures or successes the timeout value is then adjusted.

## Initial timeout values

Four categories of timeouts have been defined, each with their own timeout values:

```
	globalListingTimeout   = newDynamicTimeout(30*time.Second, 5*time.Second)  // timeout for listing related ops
	globalObjectTimeout    = newDynamicTimeout(1*time.Minute, 10*time.Second)  // timeout for Object API related ops
	globalOperationTimeout = newDynamicTimeout(10*time.Minute, 30*time.Second) // default timeout for general ops
	globalHealingTimeout   = newDynamicTimeout(30*time.Minute, 1*time.Minute)  // timeout for healing related ops
```

However I would proposed to do the initial release with longer timeouts, as per:

```
	globalListingTimeout   = newDynamicTimeout( /*30*/ 600*time.Second /*5*/, 600*time.Second) // timeout for listing related ops
	globalObjectTimeout    = newDynamicTimeout( /*1*/ 10*time.Minute /*10*/, 600*time.Second)  // timeout for Object API related ops
	globalOperationTimeout = newDynamicTimeout(10*time.Minute /*30*/, 600*time.Second)         // default timeout for general ops
	globalHealingTimeout   = newDynamicTimeout(30*time.Minute /*1*/, 30*time.Minute)           // timeout for healing related ops
```
(thus all 10 minutes, plus healing at 30 minutes).

We can then first see if there are no “logical” errors without too tight timeout values, once this has been found okay in practice, we can then gradually “tighten” the timeout values in a couple of steps to the eventual values as we see fit. (And note that using very large timeout values would effectively imply no functional changes at all, as the timeouts would never be hit.)

## Testing 

As testing with long timeouts is kind of hard/tedious, I’ve actually tested manually (both XL and distributed) with a custom build with low values (eg. 3 seconds for timeouts above), so as to force timeout occurring.

deekoder:
@fwessels  we are ready to review. Can you please rebase once again.

balamurugana:
Conflict needs to be resolved.

vendor/github.com/minio/lsync/lmutex.go
```diff
@@ -0,0 +1,78 @@
...
+func (lm *LMutex) GetLock(timeout time.Duration) (locked bool) {
+	return lm.lockLoop(timeout)
+}
+
+// lockLoop will acquire either a read or a write lock
+//
+// The call will block until the lock is granted using a built-in
+// timing randomized back-off algorithm to try again until successful
+func (lm *LMutex) lockLoop(timeout time.Duration) bool {
+	doneCh, start := make(chan struct{}), time.Now()
```
harshavardhana: Perhaps UTC is better?

vendor/github.com/minio/lsync/lrwmutex.go
```diff
@@ -0,0 +1,178 @@
...
+ */
+
+package lsync
+
+import (
+	"sync"
+	"time"
+)
+
+const WRITELOCK = -1
```
harshavardhana: Perhaps constants can be defined like this? 

```
const (
     WRITELOCK = -1 + iota
     NOLOCKS
     READLOCKS
)
```

vendor/github.com/minio/lsync/lrwmutex.go
```diff
@@ -0,0 +1,178 @@
...
+	isWriteLock := false
+	return lm.lockLoop(timeout, isWriteLock)
+}
+
+// lockLoop will acquire either a read or a write lock
+//
+// The call will block until the lock is granted using a built-in
+// timing randomized back-off algorithm to try again until successful
+func (lm *LRWMutex) lockLoop(timeout time.Duration, isWriteLock bool) bool {
+	doneCh, start := make(chan struct{}), time.Now()
```
harshavardhana: UTC as well.

harshavardhana (changes requested):

vendor/github.com/minio/lsync/lmutex.go
fwessels: Will do.

vendor/github.com/minio/lsync/lrwmutex.go
fwessels: Yes, this is nicer, will change.

vendor/github.com/minio/lsync/lrwmutex.go
fwessels: Will do.

cmd/dynamic-timeouts.go
```diff
@@ -0,0 +1,120 @@
+/*
+ * Minio Cloud Storage, (C) 2016 Minio, Inc.
```
harshavardhana: Copyright should be 2017 here.

cmd/dynamic-timeouts_test.go
```diff
@@ -0,0 +1,207 @@
+/*
+ * Minio Cloud Storage, (C) 2016 Minio, Inc.
```
harshavardhana: Same here..

harshavardhana (changes requested):
Minor things, codewise LGTM.  Continuing to test.

cmd/dynamic-timeouts.go
fwessels: Will do

cmd/dynamic-timeouts_test.go
fwessels: Will do

cmd/namespace-lock.go
```diff
@@ -152,9 +162,37 @@ func (n *nsLockMap) lock(volume, path string, lockSource, opsID string, readLock
...
+		// Changing the status of the operation from blocked to none
+		if err := n.statusBlockedToNone(param, lockSource, opsID, readLock); err != nil {
+			errorIf(err, "Failed to clear the lock state")
+		}
+
+		nsLk.ref-- // Decrement ref count since we failed to get the lock
+		// delete the lock state entry for given operation ID.
+		err := n.deleteLockInfoEntryForOps(param, opsID)
+		if err != nil {
+			errorIf(err, "Failed to delete lock info entry")
```
krisis: we could add `params` and `opsID` to the error log. It would come handy during debugging.

cmd/namespace-lock.go
```diff
@@ -152,9 +162,37 @@ func (n *nsLockMap) lock(volume, path string, lockSource, opsID string, readLock
 
 	// Locking here can block.
```
krisis: minor nit: Locking here can block until timeout?

cmd/namespace-lock.go
```diff
@@ -152,9 +162,37 @@ func (n *nsLockMap) lock(volume, path string, lockSource, opsID string, readLock
...
+		}
+		if nsLk.ref == 0 {
+			// Remove from the map if there are no more references.
+			delete(n.lockMap, param)
+
+			// delete the lock state entry for given
+			// <volume, path> pair.
+			err := n.deleteLockInfoEntryForVolumePath(param)
+			if err != nil {
+				errorIf(err, "Failed to delete lock info entry")
```
krisis: printing `param` in error log will be useful during debugging.

cmd/namespace-lock_test.go
```diff
@@ -63,7 +63,9 @@ func TestNamespaceLockTest(t *testing.T) {
 
 	// Write lock tests.
 	testCase := testCases[0]
-	testCase.lk("a", "b", "c") // lock once.
+	if !testCase.lk("a", "b", "c", 60*time.Second) { // lock once.
```
krisis, line 64: we should check for the lock entry only if we acquired the lock.

cmd/namespace-lock_test.go
```diff
@@ -128,7 +140,51 @@ func TestNamespaceLockTest(t *testing.T) {
...
+	if !globalNSMutex.RLock("my-bucket", "my-object", "nop", 60*time.Second) {
+		t.Errorf("Failed to acquire second read lock")
+	}
+
+	// Release both read locks
+	globalNSMutex.RUnlock("my-bucket", "my-object", "ghi")
+	globalNSMutex.RUnlock("my-bucket", "my-object", "nop")
+}
+
+// This test fails -- disabled for now
```
krisis: Is this still the case?

cmd/namespace-lock_test.go
```diff
@@ -339,16 +399,18 @@ func TestLockStats(t *testing.T) {
 		syncChan <- struct{}{}
 	}()
 	// waiting so that the write locks in the above go routines are held.
-	// sleeping so that we can predict the order of the write locks held.
+	// sleeping so that we can predict the order of the write locks held. <---- order cannot be predicted
```
krisis: should we just rewrite this test differently if it is non-deterministic?

krisis (changes requested):
These are some initial review comments. Overall the changes look OK, I will be running some tests locally.

cmd/namespace-lock.go
fwessels: Will update comment

cmd/namespace-lock.go
fwessels: Will do.

cmd/namespace-lock.go
fwessels: Will do.

cmd/namespace-lock_test.go
fwessels: Will change to Fatalf()

cmd/namespace-lock_test.go
fwessels: This test case is flawed as it is making implicit assumptions about the order of events (plus explicit synchronisation) so it is actually a bit of surprise that it is working.
Removing this testcase (also in lieu of additional test cases in lsync).

cmd/namespace-lock_test.go
fwessels: See above, removing

balamurugana (commented):
LGTM

harshavardhana:
Sorry @fwessels lagging behind on testing this. Will be doing it this week.

fwessels:
@harshavardhana no worries, better safe than sorry on this one...

krisis (approved):
Tested locally, LGTM :+1:

balamurugana (approved):
mint succeeded

harshavardhana (approved):

harshavardhana (approved):
Approving despite decrease in code coverage. Large parts of the code can only be tested outside in mint.

PR #4379, opened 2017-05-20
Add nsswitch.conf to our docker image

harshavardhana:
## Description
Add nsswitch.conf to our docker image

## Motivation and Context
This will cause the alpine images to resolve DNS 
using /etc/hosts along with the normal DNS resolver.

## How Has This Been Tested?
Manually setting /etc/hosts

## Types of changes

## Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,7 @@
 FROM alpine:3.5
 
+MAINTAINER Minio Inc <dev@minio.io>
+
 ENV GOPATH /go
 ENV PATH $PATH:$GOPATH/bin
 ENV CGO_ENABLED 0
@@ -9,6 +11,7 @@ WORKDIR /go/src/github.com/minio/
 RUN  \
      apk add --no-cache ca-certificates && \
      apk add --no-cache --virtual .build-deps git go musl-dev && \
+     echo 'hosts: files mdns4_minimal [NOTFOUND=return] dns mdns4' >> /etc/nsswitch.conf && \     
      go get -v -d github.com/minio/minio && \
      cd /go/src/github.com/minio/minio && \
      go install -v -ldflags "$(go run buildscripts/gen-ldflags.go)" && \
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,5 +1,7 @@
 FROM alpine:3.5
 
+RUN echo 'hosts: files dns' >> /etc/nsswitch.conf
```
balamurugana: Is it good idea to conditionally add/edit `hosts: files dns` than forceful append?

harshavardhana: In-fact this file doesn't even exist @balamurugana inside the container. It is meant for only one purpose.

Dockerfile.aarch64
```diff
@@ -9,6 +13,7 @@ WORKDIR /go/src/github.com/minio/
 RUN  \
      apk add --no-cache ca-certificates && \
      apk add --no-cache --virtual .build-deps git go musl-dev && \
+     echo 'hosts: files mdns4_minimal [NOTFOUND=return] dns mdns4' >> /etc/nsswitch.conf && \
```
balamurugana, line 14: Two `RUN` have `/etc/nsswitch.conf` updated.  Is it expected?

Dockerfile.armhf
```diff
@@ -9,6 +13,7 @@ WORKDIR /go/src/github.com/minio/
 RUN  \
      apk add --no-cache ca-certificates && \
      apk add --no-cache --virtual .build-deps git go musl-dev && \
+     echo 'hosts: files mdns4_minimal [NOTFOUND=return] dns mdns4' >> /etc/nsswitch.conf && \
```
balamurugana, line 14: same as above.

balamurugana (changes requested):

Dockerfile.aarch64
harshavardhana: Should be updated my mistake

nitisht (approved):

harshavardhana:
@balamurugana can you review once again?

balamurugana (approved):

PR #4402, opened 2017-05-23
Adjust /export permissions for arbitary-user support

### LINKED ISSUE #4381: Drop root user in Docker image
I would like to be able to drop the root user in the Minio docker image

## Expected Behavior
```
FROM minio/minio
adduser -u 1000 -G root default
USER default
```

## Current Behavior
Permission denied on the volume /export

It is not possible to change the permissions on an existing volume except in the base image and at run time.

## Possible Solution

There are a few solutions:

- Add a minio user and chown the volume:
https://github.com/docker-library/redis/blob/master/3.2/alpine/Dockerfile#L4
https://github.com/docker-library/redis/blob/master/3.2/alpine/Dockerfile#L47

- Simply chmod 777 /export

## Context

Docker recommends running containers without root. On some platforms like openshift, the ability to run containers as root are heavily restricted.

andrewklau:
## Description
The export volume created in the base image is only allowing root to write to the volume 

Docker recommends running containers without root. On platforms like openshift, the root user is not allowed. Instead an arbitrary user ID attached to root is provided.

## Motivation and Context
Fixes https://github.com/minio/minio/issues/4381

## How Has This Been Tested?
```
FROM minio/minio
RUN adduser -S -u 1001 -G root default
USER default
```

## Types of changes

## Checklist:

### DIFF
--- a/buildscripts/docker-entrypoint.sh
+++ b/buildscripts/docker-entrypoint.sh
@@ -78,4 +78,10 @@ docker_secrets_env
 ## Wait for all the hosts to come online.
 docker_wait_hosts "$@"
 
+# allow the container to be started with `--user`
+if [ "$1" = 'minio' -a "$(id -u)" = '0' ]; then
+  chown -R minio .
+	exec su-exec minio "$0" "$@"
+fi
+
 exec "$@"
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,6 +5,9 @@ MAINTAINER Minio Inc <dev@minio.io>
 # add our user and group first to make sure their IDs get assigned consistently, regardless of whatever dependencies get added
 RUN addgroup -S minio && adduser -S -G minio minio
 
+# grab su-exec for easy step-down from root
+RUN apk add --no-cache 'su-exec>=0.2'
+
 ENV GOPATH /go
 ENV PATH $PATH:$GOPATH/bin
 ENV CGO_ENABLED 0
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,6 +2,9 @@ FROM alpine:3.5
 
 MAINTAINER Minio Inc <dev@minio.io>
 
+# add our user and group first to make sure their IDs get assigned consistently, regardless of whatever dependencies get added
+RUN addgroup -S minio && adduser -S -G minio minio
+
 ENV GOPATH /go
 ENV PATH $PATH:$GOPATH/bin
 ENV CGO_ENABLED 0
@@ -11,11 +14,12 @@ WORKDIR /go/src/github.com/minio/
 RUN  \
      apk add --no-cache ca-certificates && \
      apk add --no-cache --virtual .build-deps git go musl-dev && \
-     echo 'hosts: files mdns4_minimal [NOTFOUND=return] dns mdns4' >> /etc/nsswitch.conf && \     
+     echo 'hosts: files mdns4_minimal [NOTFOUND=return] dns mdns4' >> /etc/nsswitch.conf && \
      go get -v -d github.com/minio/minio && \
      cd /go/src/github.com/minio/minio && \
      go install -v -ldflags "$(go run buildscripts/gen-ldflags.go)" && \
-     rm -rf /go/pkg /go/src /usr/local/go && apk del .build-deps
+     rm -rf /go/pkg /go/src /usr/local/go && apk del .build-deps && \
+     mkdir /export && chown minio:minio /export
 
 EXPOSE 9000
 
--- a/Dockerfile.aarch64
+++ b/Dockerfile.aarch64
@@ -5,6 +5,9 @@ MAINTAINER Minio Inc <dev@minio.io>
 # add our user and group first to make sure their IDs get assigned consistently, regardless of whatever dependencies get added
 RUN addgroup -S minio && adduser -S -G minio minio
 
+# grab su-exec for easy step-down from root
+RUN apk add --no-cache 'su-exec>=0.2'
+
 ENV GOPATH /go
 ENV PATH $PATH:$GOPATH/bin
 ENV CGO_ENABLED 0
--- a/Dockerfile.aarch64
+++ b/Dockerfile.aarch64
@@ -2,6 +2,9 @@ FROM resin/aarch64-alpine:3.5
 
 MAINTAINER Minio Inc <dev@minio.io>
 
+# add our user and group first to make sure their IDs get assigned consistently, regardless of whatever dependencies get added
+RUN addgroup -S minio && adduser -S -G minio minio
+
 ENV GOPATH /go
 ENV PATH $PATH:$GOPATH/bin
 ENV CGO_ENABLED 0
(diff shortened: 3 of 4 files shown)


fwessels (approved):
LGTM

nitisht:
Need to test with orchestration tools before going ahead. Will do that today and add review. LGTM code wise.

nitisht:
Thank you for the contribution @andrewklau . I had a discussion with @harshavardhana , while we acknowledge the issue #4381, we are not convinced if this is the best way to go about it. Ideally we'd like to mode towards https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#user

We still need to discuss internally, will get back about with the conclusion.

andrewklau:
I also think the useradd method would be the best approach, I was just not sure how well that would be received as it might have a higher chance of causing some incompatibility issues.

nitisht:
Thanks @andrewklau this looks like a better approach, but we will need to explore all the scenarios before we go ahead with this.

andrewklau:
No worries! I'll just build my own image in the meantime. Thanks for the consideration.

nitisht:
Tried building this locally, Docker build fails 

```
$ sudo docker build .
Sending build context to Docker daemon  111.3MB
Step 1/15 : FROM alpine:3.5
 ---> 75b63e430bd1
Step 2/15 : MAINTAINER Minio Inc <dev@minio.io>
 ---> Using cache
 ---> e7bf5ec89bc5
Step 3/15 : RUN addgroup -S minio && adduser -S -G minio minio
 ---> Using cache
 ---> 820d05e67e62
Step 4/15 : RUN apk add --no-cache 'su-exec>=0.2'
 ---> Running in 3de157a13dbc
fetch http://dl-cdn.alpinelinux.org/alpine/v3.5/main/x86_64/APKINDEX.tar.gz
WARNING: Ignoring http://dl-cdn.alpinelinux.org/alpine/v3.5/main/x86_64/APKINDEX.tar.gz: temporary error (try again later)
fetch http://dl-cdn.alpinelinux.org/alpine/v3.5/community/x86_64/APKINDEX.tar.gz
WARNING: Ignoring http://dl-cdn.alpinelinux.org/alpine/v3.5/community/x86_64/APKINDEX.tar.gz: temporary error (try again later)
ERROR: unsatisfiable constraints:
  su-exec (missing):
    required by: world[su-exec>=0.2]
The command '/bin/sh -c apk add --no-cache 'su-exec>=0.2'' returned a non-zero code: 1
```

andrewklau:
Looks like a temporary error with the alpine repositories. I seem to hit that every now and then too. Try again later?

nitisht:
I see. Let me try again in a while

nitisht:
I tried it again and got the same failure. 

Thank you for the contribution @andrewklau . Do you mind if we close this for now and take the ideas discussed here under the issue https://github.com/minio/minio/issues/4565

andrewklau:
Mine seems to work:

```
Status: Downloaded newer image for docker.io/alpine:3.5
 ---> 074d602a59d7
Step 2/15 : MAINTAINER Minio Inc <dev@minio.io>
 ---> Running in 465988338f59
 ---> c39a477595ae
Removing intermediate container 465988338f59
Step 3/15 : RUN addgroup -S minio && adduser -S -G minio minio
 ---> Running in 98366c1b540d
 ---> 63f912a6778e
Removing intermediate container 98366c1b540d
Step 4/15 : RUN apk add --no-cache 'su-exec>=0.2'
 ---> Running in c828d7a242d3
fetch http://dl-cdn.alpinelinux.org/alpine/v3.5/main/x86_64/APKINDEX.tar.gz
fetch http://dl-cdn.alpinelinux.org/alpine/v3.5/community/x86_64/APKINDEX.tar.gz
(1/1) Installing su-exec (0.2-r0)
Executing busybox-1.25.1-r0.trigger
OK: 4 MiB in 12 packages
 ---> 24d046cf77f2
```

I do often see that alpine mirror error on our CI systems too.. and it's very intermittent (alpine mirror connectivity issue). 

We can close this if you feel there are better approaches. Although this seems to cover what's being done in the official images.

nitisht:
@andrewklau so I got to test this finally. The problem with this approach is that when someone updates Minio image, permissions don't match. Here is how you can reproduce it

- Launch Minio with latest Docker Image based on a host volume:

```
$ sudo docker run -p 9000:9000 -v /home/nitish/minio_volume:/export --net=host minio/minio:RELEASE.2017-06-13T19-01-01Z server /export

Created minio configuration file successfully at /root/.minio

Endpoint:  http://192.168.1.80:9000  http://172.17.0.1:9000  http://127.0.0.1:9000
AccessKey: ADWHD1TFPXX375L7S8O5 
SecretKey: 1kf/BYrVHmP+tBDwRKHfMfN4ua3A3OSsfq2HVhpU 

Browser Access:
   http://192.168.1.80:9000  http://172.17.0.1:9000  http://127.0.0.1:9000

   $ mc config host add myminio http://192.168.1.80:9000 ADWHD1TFPXX375L7S8O5 1kf/BYrVHmP+tBDwRKHfMfN4ua3A3OSsfq2HVhpU

   .NET:       https://docs.minio.io/docs/dotnet-client-quickstart-guide

Drive Capacity: 206 GiB Free, 233 GiB Total
```
- Stop the container and launch a new container based on this Dockerfile. 

```
$ sudo docker build . -t minio:4402
$ sudo docker run -p 9000:9000 -v /home/nitish/minio_volume:/export --net=host minio:4402 server /export
```

This is the error 
```
time="2017-07-06T01:32:31Z" level=fatal msg="Initializing object layer failed" cause="Unable to initialize '.minio.sys' meta volume, mkdir /export/.minio.sys/tmp/c78b0f42-63a9-43c6-b5e5-69b4b3b9a3c5: permission denied" source="[server-main.go:201:serverMain()]" 
```

nitisht (changes requested):
Need to fix the issue as described in the comment.

buildscripts/docker-entrypoint.sh
```diff
@@ -78,4 +78,10 @@ docker_secrets_env
 ## Wait for all the hosts to come online.
 docker_wait_hosts "$@"
 
+# allow the container to be started with `--user`
+if [ "$1" = 'minio' -a "$(id -u)" = '0' ]; then
+  chown -R minio .
```
harshavardhana, line 83: This should be done for the "/export" - also looks like the tabbing is off.. Also don't need to provide this option.  Let's by default move to user style approach.  `su-exec` is not needed. 

All directories provided and local to this instance should chown itself. It would require properly parsing minio command line though.

harshavardhana (changes requested):

andrewklau:
I'm travelling right now so getting back to this might be a bit slower.

The issue @NitishT saw we can fix by running the chmod/chown before the `su-exec`. This is inline with what Docker official images are doing. I forgot to include that.

But my understanding from @hackintoshrao comment is we should go with something like:

Dockerfile
```
USER minio
```

This would break the backwards comparability, as existing volumes that are not running as the `minio` user we will no longer be able to chmod/chown

andrewklau:
After thinking this through I don't believe we can use `USER minio` as this will break backwards compatibility. This is probably why most of the current official docker images also don't use USER

Then if we use `su-exec` it will require chowning all the existing files which is also undesirable and may have bootup performance issues for existing users.

Both seem to break backwards compatibility some what but `su-exec` will at least still allow it to run.

nitisht:
Closing this as we decided not to go with this approach. Another approach is being attempted here #4877

PR #4409, opened 2017-05-23
Set default ETag value if vendor returns empty md5 string

### LINKED ISSUE #4397: [gateway-gcs] Empty ETag returned during  a Put object call
## Expected Behavior

Running a gcs server locally  and uploading a  78.23Mb file to a bucket using python SDK. Expect script to complete successfully.
## Current Behavior

python script crashes because ETag returned by gcs gateway is empty. minio-python sdk expects a nonempty ETag in CompleteMultipartUpload method. Note that the file upload works ok with mc command line.
## Possible Solution

No idea. However server log doesn't show any errors.

1. Start minio gateway gcs server locally
2. Run a fput_object call with gcs credentials from minio-py sdk for a large object
3. The SDK crashes because of missing Etag

$ python3 make_bucket.py
received data b'<?xml version="1.0" encoding="UTF-8"?>\n<CompleteMultipartUploadResult xmlns="http://s3.amazonaws.com/doc/2006-03-01/"><Location>/aquaduck1/object2</Location><Bucket>aquaduck1</Bucket><Key>object2</Key><ETag></ETag> </CompleteMultipartUploadResult>'   <=== xml returned by gateway GCS
None
Traceback (most recent call last):
  File "make_bucket.py", line 37, in <module>
    '/Users/krispy/Downloads/opencv-3.2.0.zip', content_type='application/zip')
  File "/Users/krispy/GMinio/minio-py/minio/api.py", line 687, in fput_object
    uploaded_parts)
  File "/Users/krispy/GMinio/minio-py/minio/api.py", line 1753, in _complete_multipart_upload
    return parse_multipart_upload_result(response.data)
  File "/Users/krispy/GMinio/minio-py/minio/parsers.py", line 162, in parse_multipart_upload_result
    root.get_etag_elem()
  File "/Users/krispy/GMinio/minio-py/minio/parsers.py", line 127, in get_etag_elem
    return self.get_child_text('ETag', strict).replace('"', '')
AttributeError: 'NoneType' object has no attribute 'replace'

==same upload from mc ==
$ mc cp ~/Downloads/opencv-3.2.0.zip gcs2/aquaduck1 
...opencv-3.2.0.zip:  78.23 MB / 78.23 MB  ▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓  100.00% 2.27 MB/s 34s

## Your Environment

* Server type and version:
* Link to your project:
Version: 2017-05-23T00:51:58Z
Release-Tag: DEVELOPMENT.2017-05-23T00-51-58Z
Commit-ID: 2659ff77b65e30224f73e805e1f49c11f5ba17a5

minio-py sdk from master commit a60bf9336aa3d20308c563356ffbed94fed95cd0

poornas:
## Description

The ETag is constructed from md5 atttribute of object attributes returned by the vendor's Composer. The md5 attribute comes back as nil for large uploads. Instead the CRC32C should be used. Refer to https://cloud.google.com/storage/docs/hashes-etags 
Fixes #4397

## Motivation and Context

Large uploads fail because client SDK is not receiving ETag in CompleteMultipartUpload response. 
## How Has This Been Tested?

Ran unit tests locally
## Types of changes

## Checklist:

### DIFF
--- a/cmd/gateway-gcs.go
+++ b/cmd/gateway-gcs.go
@@ -17,7 +17,6 @@
 package cmd
 
 import (
-	"bytes"
 	"cloud.google.com/go/storage"
 	"context"
 	"crypto/sha256"
@@ -356,7 +355,7 @@ func (l *gcsGateway) ListObjects(bucket string, prefix string, marker string, de
 			Bucket:          attrs.Bucket,
 			ModTime:         attrs.Updated,
 			Size:            attrs.Size,
-			ETag:            hex.EncodeToString(attrs.MD5),
+			ETag:            fmt.Sprintf("%d", attrs.CRC32C),
 			UserDefined:     attrs.Metadata,
 			ContentType:     attrs.ContentType,
 			ContentEncoding: attrs.ContentEncoding,
@@ -401,7 +400,7 @@ func (l *gcsGateway) ListObjectsV2(bucket, prefix, continuationToken string, fet
 			continue
 		}
 
-		objects = append(objects, fromGCSObjectInfo(attrs))
+		objects = append(objects, fromGCSAttrsToObjectInfo(attrs))
 	}
 
 	return ListObjectsV2Info{
@@ -441,18 +440,22 @@ func (l *gcsGateway) GetObject(bucket string, key string, startOffset int64, len
 	return nil
 }
 
-// fromGCSObjectInfo converts GCS BucketAttrs to gateway ObjectInfo
-func fromGCSObjectInfo(attrs *storage.ObjectAttrs) ObjectInfo {
-	return ObjectInfo{
+// fromGCSAttrsToObjectInfo converts GCS BucketAttrs to gateway ObjectInfo
+func fromGCSAttrsToObjectInfo(attrs *storage.ObjectAttrs) ObjectInfo {
+	// All google cloud storage objects have a CRC32c hash, whereas composite objects may not have a MD5 hash
+	// Refer https://cloud.google.com/storage/docs/hashes-etags. Use CRC32C for ETag
+	return  ObjectInfo{
 		Name:            attrs.Name,
 		Bucket:          attrs.Bucket,
 		ModTime:         attrs.Updated,
 		Size:            attrs.Size,
-		ETag:            hex.EncodeToString(attrs.MD5),
+		ETag: 			 fmt.Sprintf("%d", attrs.CRC32C),
 		UserDefined:     attrs.Metadata,
 		ContentType:     attrs.ContentType,
 		ContentEncoding: attrs.ContentEncoding,
 	}
+	
+	 
 }
 
 // GetObjectInfo - reads object info and replies back ObjectInfo
@@ -468,7 +471,7 @@ func (l *gcsGateway) GetObjectInfo(bucket string, object string) (ObjectInfo, er
 	if err != nil {
 		return ObjectInfo{}, gcsToObjectError(traceError(err), bucket, object)
 	}
-	objInfo := fromGCSObjectInfo(attrs)
+	objInfo := fromGCSAttrsToObjectInfo(attrs)
 	objInfo.ETag = fmt.Sprintf("%d", attrs.CRC32C)
 
 	return objInfo, nil
@@ -502,6 +505,14 @@ func (l *gcsGateway) PutObject(bucket string, key string, size int64, data io.Re
 
 	w.ContentType = metadata["content-type"]
 	w.ContentEncoding = metadata["content-encoding"]
+	if md5sum =="" {
+	} else if md5, err := hex.DecodeString(md5sum); err != nil {
+		return ObjectInfo{}, gcsToObjectError(traceError(err), bucket, key)
+	} else {
+		w.MD5 = md5
+	}
+	
+
 	w.Metadata = metadata
 
 	_, err := io.Copy(w, teeReader)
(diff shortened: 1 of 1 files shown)


cmd/gateway-gcs.go
```diff
@@ -835,7 +835,11 @@ func (l *gcsGateway) CompleteMultipartUpload(bucket string, key string, uploadID
 
 	partZero.Delete(l.ctx)
 
-	return fromGCSObjectInfo(attrs), gcsToObjectError(traceError(err), bucket, key)
+	objInfo := fromGCSObjectInfo(attrs)
+	if objInfo.ETag == "" {
+		objInfo.ETag = bucket + key + uploadID
```
harshavardhana: you can do a md5Sum of the "bucket + key" - uploadID is gone after the upload is complete, so we don't need this to be part of the ETag. 

```
h := md5.New()
io.WriteString(objInfo.Name+objectInfo.Key+objectInfo.Size)
hex.EncodeToString(h.Sum(nil))
```

^^ this would be better.

harshavardhana (changes requested):

cmd/gateway-gcs.go
```diff
@@ -835,7 +835,14 @@ func (l *gcsGateway) CompleteMultipartUpload(bucket string, key string, uploadID
 
 	partZero.Delete(l.ctx)
 
-	return fromGCSObjectInfo(attrs), gcsToObjectError(traceError(err), bucket, key)
+	objectInfo := fromGCSObjectInfo(attrs)
+	if objectInfo.ETag == "" {
+		h := md5.New()
+		io.WriteString(h, objectInfo.Bucket+objectInfo.Name+strconv.FormatInt(objectInfo.Size, 10))
```
harshavardhana: May be a simple strconv.Atoi would work? or perhaps just use `fmt.Sprintf("%s-%s-%d")`. 

One fundamental problem i see with this approach is that ETag needs to be persistent, if we cannot say that this value is persistent i.e say when 

GetObject() asks for `If-Match`  then we are in trouble. Can you check what do we respond for a GetObject() or a HeadObject() in case of GCS, we can basically use the same ETag i.e sent back to the client.

krishnasrinivas (changes requested):
```
objectInfo.ETag = fmt.Sprintf("%d", attrs.CRC32C)
```

you can do this inside `fromGCSObjectInfo()` that way PutObject and CopyObject are also taken care of

cmd/gateway-gcs.go
```diff
@@ -827,15 +829,18 @@ func (l *gcsGateway) CompleteMultipartUpload(bucket string, key string, uploadID
...
 	// cleanup, delete all parts
 	for _, uploadedPart := range uploadedParts {
 		l.client.Bucket(bucket).Object(toGCSMultipartKey(key, uploadID, uploadedPart.PartNumber)).Delete(l.ctx)
 	}
 
 	partZero.Delete(l.ctx)
 
-	return fromGCSObjectInfo(attrs), gcsToObjectError(traceError(err), bucket, key)
+	objectInfo := fromGCSObjectInfo(attrs)
+	// All google cloud storage objects have a CRC32c hash, whereas composite obejcts may not have a MD5 hash
```
harshavardhana: A typo here `s/obejcts/objects`

cmd/gateway-gcs.go
```diff
@@ -459,11 +458,14 @@ func (l *gcsGateway) GetObjectInfo(bucket string, object string) (ObjectInfo, er
 	}
 
 	attrs, err := l.client.Bucket(bucket).Object(object).Attrs(l.ctx)
+
 	if err != nil {
 		return ObjectInfo{}, gcsToObjectError(traceError(err), bucket, object)
 	}
+	objInfo := fromGCSObjectInfo(attrs)
+	objInfo.ETag = fmt.Sprintf("%d", attrs.CRC32C)
```
harshavardhana: As @krishnasrinivas suggested we could potentially move this to `fromGCSObjectInfo()` function itself. where we will pick like this. 

Also it looks like the function name is wrong it should be `toGCSObjectInfo` not `fromGCSObjectInfo`

This can be combined it should work for all scenarios.
```
// toGCSObjectInfo converts GCS BucketAttrs to gateway ObjectInfo                                                                              
 func toGCSObjectInfo(attrs *storage.ObjectAttrs) ObjectInfo {
         objInfo := ObjectInfo{
                 Name:            attrs.Name,
                 Bucket:          attrs.Bucket,
                 ModTime:         attrs.Updated,
-                Size:            attrs.Size,
                 UserDefined:     attrs.Metadata,
                 ContentType:     attrs.ContentType,
                 ContentEncoding: attrs.ContentEncoding,
         }
+        objInfo.ETag = fmt.Sprintf("%d", attrs.CRC32C)
+        if attrs.MD5 != nil {
+                objInfo.ETag = hex.EncodeToString(attrs.MD5)
+        }
+        return	objInfo
 }
}
```

harshavardhana (changes requested):

cmd/gateway-gcs.go
poornas: I think the function naming is correct, isn't it? It is constructed from GCS attributes - maybe fromGCSToObjectInfo?

harshavardhana: Yes `fromGCSAttrsToObjectInfo()` would work as well.

krishnasrinivas: Better make CRC32 as ETag always (irrespective of whether MD5 is set) so that ETag looks "consistent" for all objects. Also there is no particular advantage in making ETag as MD5 when it is available.

vadmeste:
@poornas, can you rebase this ?

cmd/gateway-gcs.go
```diff
@@ -17,23 +17,21 @@
...
+	"google.golang.org/api/googleapi"
+	"google.golang.org/api/iterator"
 	"hash"
 	"io"
 	"path"
 	"regexp"
 	"strconv"
 	"strings"
 
-	"cloud.google.com/go/storage"
```
harshavardhana: We should avoid imports being organized in between other standard libraries.. https://github.com/golang/go/wiki/CodeReviewComments#imports

```Imports are organized in groups, with blank lines between them. The standard library packages are always in the first group.```

harshavardhana (changes requested):

krishnasrinivas (approved):

harshavardhana (approved):

harshavardhana (approved):

Issue #4417, opened 2017-05-24
[gateway-gcs] Support for PROFILE


krishnasrinivas:
```
 minio gateway:

  4. Start minio gateway server for GCS backend.
      $ gcloud init --console-only # this will configure your google account
      $ export MINIO_ACCESS_KEY=accesskey
      $ export MINIO_SECRET_KEY=secretkey
      $ minio gateway gcs [PROFILE:]PROJECTID
```

Support for PROFILE is not available as shown in the help here

deekoder:
I believe project id is the only thing needed since its unique. But its better to check on this. Krishna, have you any test where you wanted to use the profile and were unable to?

harshavardhana:
> minio gateway:
> 
>   4. Start minio gateway server for GCS backend.
>       $ gcloud init --console-only # this will configure your google account
>       $ export MINIO_ACCESS_KEY=accesskey
>       $ export MINIO_SECRET_KEY=secretkey
>       $ minio gateway gcs [PROFILE:]PROJECTID
> Support for PROFILE is not available as shown in the help here

There is no easier way to support PROFILE AFAIK, atleast from SDK point of view it doesn't differentiate. Only parameter it takes is `projectID`.

deekoder:
We do have to verify that this is the case.

harshavardhana:
> We do have to verify that this is the case.

@nl5887 already verified this before and said its not possible.

krishnasrinivas:
> @nl5887 already verified this before and said its not possible.

Cool, In that case the help message needs to be fixed.

deekoder:
ok then gr8.

harshavardhana:
Closing this.. as fixed in #4421

PR #4420, opened 2017-05-24
Fix gateway start up message to remove profile option for gcs

poornas:
## Description

minio gateway command currently lists profile as an optional argument to start the gcs gateway. Only projectid is supported, and this PR removes the profile option from help message to make it consistent
## Motivation and Context

See #4417
## Types of changes

## Checklist:


harshavardhana:
the branch is wrong @poornas can you close the PR and send it to `feature-gcs` branch.

deekoder:
Please add how this was tested. Basically can we remove this option safely? Can there be same project ids with different profile names in gcs.

cmd/gateway-anonymous.go
```diff
@@ -0,0 +1,49 @@
...
+import "net/http"
+
+func anonErrToObjectErr(statusCode int, params ...string) error {
+	bucket := ""
+	object := ""
+	if len(params) >= 1 {
+		bucket = params[0]
+	}
+	if len(params) == 2 {
+		object = params[1]
```
balamurugana, line 28: optional `params` would lead more than two arguments ie beyond bucket, object, its not useful.

According to this function, bucket is always needed and object might not come.  I would prefer
```go
func anonErrToObjectErr(statusCode int, bucket. object string) error
```
where object can be "" treat it as nil

cmd/gateway-anonymous.go
```diff
@@ -0,0 +1,49 @@
...
+		}
+		return BucketNotFound{Bucket: bucket}
+	case http.StatusBadRequest:
+		if object != "" {
+			return ObjectNameInvalid{bucket, object}
+		}
+		return BucketNameInvalid{Bucket: bucket}
+	case http.StatusForbidden:
+		fallthrough
+	case http.StatusUnauthorized:
```
balamurugana, line 44: may be `case http.StatusForbidden, http.StatusUnauthorized:` better?

harshavardhana:
> Please add how this was tested. Basically can we remove this option safely? Can there be same project ids with different profile names in gcs.

You cannot basically if you created a project with non-default it won't work.  It has to be default only IMO.

cmd/gateway-gcs-anonymous.go
```diff
@@ -0,0 +1,131 @@
...
+	req, err := http.NewRequest("GET", toGCSPublicURL(bucket, object), nil)
+	if err != nil {
+		return gcsToObjectError(traceError(err), bucket, object)
+	}
+
+	if length > 0 && startOffset > 0 {
+		req.Header.Add("Range", fmt.Sprintf("bytes=%d-%d", startOffset, startOffset+length-1))
+	} else if startOffset > 0 {
+		req.Header.Add("Range", fmt.Sprintf("bytes=%d-", startOffset))
+	}
```
balamurugana, line 48: may be below more readable?

```go
if startOffset > 0 {
  if length > 0 {
    req.Header.Add("Range", fmt.Sprintf("bytes=%d-%d", startOffset, startOffset+length-1))
  } else {
    req.Header.Add("Range", fmt.Sprintf("bytes=%d-", startOffset))
  }
}
```

cmd/gateway-gcs-anonymous.go
```diff
@@ -0,0 +1,131 @@
...
+
+	objInfo.ModTime = t
+	objInfo.Bucket = bucket
+	objInfo.UserDefined = make(map[string]string)
+	if resp.Header.Get("Content-Encoding") != "" {
+		objInfo.UserDefined["Content-Encoding"] = resp.Header.Get("Content-Encoding")
+	}
+	objInfo.UserDefined["Content-Type"] = resp.Header.Get("Content-Type")
+	objInfo.ETag = resp.Header.Get("Etag")
+	objInfo.ModTime = t
```
balamurugana, line 98: Its already assigned above.

cmd/gateway-gcs_test.go
```diff
@@ -0,0 +1,97 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cmd
+
+import "testing"
+
+func TestEscape(t *testing.T) {
```
balamurugana, line 21: As `escape()` and `unescape()` are tested in this function, may be `TestEscapeUnescape()` better

cmd/gateway-handlers.go
```diff
@@ -180,6 +180,8 @@ func (api gatewayAPIHandlers) PutObjectHandler(w http.ResponseWriter, r *http.Re
 	bucket = vars["bucket"]
 	object = vars["object"]
 
+	// TODO: we should validate the object name here
```
balamurugana, line 183: Can we open an issue to track this?

cmd/gateway-gcs.go
```diff
@@ -0,0 +1,917 @@
...
+	reason := googleAPIErr.Errors[0].Reason
+	message := googleAPIErr.Errors[0].Message
+
+	switch reason {
+	case "required":
+		// Anonymous users does not have storage.xyz access to project 123.
+		fallthrough
+	case "keyInvalid":
+		fallthrough
+	case "forbidden":
```
balamurugana, line 111: As cases "required" and "keyInvalid" do nothing, you could have `case "required", "keyInvalid", "forbidden":`

cmd/gateway-gcs.go
```diff
@@ -0,0 +1,917 @@
...
+				Bucket: bucket,
+			}
+		}
+	default:
+		err = fmt.Errorf("Unsupported error reason: %s", reason)
+
+	}
+
+	_ = bucket
+	_ = object
```
balamurugana, line 151: why we need to do this?

cmd/gateway-gcs.go
```diff
@@ -0,0 +1,917 @@
...
+	err = e.e
+
+	bucket := ""
+	object := ""
+	if len(params) >= 1 {
+		bucket = params[0]
+	}
+	if len(params) == 2 {
+		object = params[1]
+	}
```
balamurugana, line 76: same as above in cmd/gateway-anonymous.go

balamurugana (changes requested):
Partial review

poornas:
closing PR and resubmitting to feature-gcs branch

PR #4422, opened 2017-05-24
Do not send envVars in ServerInfo()

harshavardhana:
## Description
Sending envVars along with access and secret
exposes the entire minio server's sensitive
information. This will be an unexpected
situation for all users.

If at all we need to look for things like if
credentials are set through env, we should
only have access to only this information
not the entire set of system envs.

## Motivation and Context
Security and avoid sending sensitive information

## How Has This Been Tested?
Manually. 

## Types of changes

## Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -124,7 +124,7 @@ test: build
 	@echo "Done."
 
 coverage: build
-	@echo -n "Running all coverage for minio: "
+	@echo "Running all coverage for minio: "
 	@./buildscripts/go-coverage.sh
 	@echo "Done."
 
--- a/browser/app/js/components/Browse.js
+++ b/browser/app/js/components/Browse.js
@@ -68,7 +68,7 @@ export default class Browse extends React.Component {
           memory: res.MinioMemory,
           platform: res.MinioPlatform,
           runtime: res.MinioRuntime,
-          envVars: res.MinioEnvVars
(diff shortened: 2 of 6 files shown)


vadmeste:
I tested this PR, I was expecting to see 'xxx' in access/secret keys fields in Settings dialog but I am still seeing the actual access/secret keys. Isn't this the expected behavior? at least this is what I understood from the code.

I did never test any PR related to the browser since moving the latter under `minio/minio`

So this is what I did:
1. Downloaded the PR, 
2. `cd browser/; yarn; yarn dev`
3. `minio server /tmp/fs` in another terminal
4. Opening http://localhost:8080

Am I missing something ?

harshavardhana:
> I did never test any PR related to the browser since moving the latter under minio/minio
> 
> So this is what I did:
> 
> Downloaded the PR,
> cd browser/; yarn; yarn dev
> minio server /tmp/fs in another terminal
> Opening http://localhost:8080

You have to set the access and secret through env @vadmeste for it to be disabled.

harshavardhana:
> You have to set the access and secret through env @vadmeste for it to be disabled.

Also the data-structure has changed..did you build `minio server` again?

vadmeste (approved):
LGTM and tested.

fwessels (approved):
Tested and OK

Issue #4391, opened 2017-05-22
No indication to user that they need to login to gcs


deekoder:
There is no way to know for the user that they have to do 
```
gcloud beta auth application-default login
```
## Expected Behavior
 Error message should have above command to reauth or gcloud init if the administrator is setting minio with gcs for the first time.

## Current Behavior
fails without accurate error message

## Possible Solution
Have custom error messages for GCS or the usability is significantly harder to figure out.

## Steps to Reproduce (for bugs)
1. Don't login to gcs and try to do any bucket/object operations. 

## Context
 Testing

## Your Environment

* Version used (`minio version`): minio gateway gcs branch
* Environment name and version (e.g. nginx 1.9.1): 
* Server type and version: 
* Operating System and version (`uname -a`): Darwin macOS
* Link to your project:

krishnasrinivas:
There is a better way to authenticate for an application like minio.

When users want to deploy minio gateway on VMs or using docker containers we can't expect them to do "gloud init" the deployment would have to be automated/scriptable.

https://cloud.google.com/storage/docs/authentication the section `Other Environments` points to https://developers.google.com/identity/protocols/application-default-credentials#howtheywork which has instructions on how to use `GOOGLE_APPLICATION_CREDENTIALS` This is most suitable for applications like minio.

I have edited the help message accordingly. We can use the same way for our testing as well, i.e the admin will have to generate and give us credentials.json for testing.

and:
```
export GOOGLE_APPLICATION_CREDENTIALS=/home/krishna/Downloads/miniotest-d4530e7be663.json
minio gateway gcs miniotesting
```

```

nitisht:
This is an example: https://gist.github.com/NitishT/ae31e00695788f55f498c5534b35c29f#run-gcs-gateway-on-docker

krishnasrinivas:
@NitishT ah cool, you had figured it out. For the GCS document for community we can write the document not requiring them to install google cloud sdk and generate using `gcloud beta auth application-default login` instead we can provide instructions on how to generate credentials.json from the google cloud console UI. It will be easier.

nitisht:
Yes, just the `projectid` is required and user should login using `gcloud beta auth application-default login` and set the `GOOGLE_APPLICATION_CREDENTIALS`.

krishnasrinivas:
@NitishT I meant it's better to provide instructions to generate credentials.json from GCS web console, not `gcloud login` as users don't have to install gcloud to generate credentials.json

[...]

PR #4424, opened 2017-05-24
Fix help message for GCS

krishnasrinivas:
## Description

`gcloud init` or `gcloud beta auth application-default login` way of providing credentials in not suitable for applications like minio.

When users want to deploy minio gateway on VMs or using docker containers we can't expect them to do "gloud init" the deployment would have to be automated/scriptable.

https://cloud.google.com/storage/docs/authentication the section Other Environments points to https://developers.google.com/identity/protocols/application-default-credentials#howtheywork which has instructions on how to use GOOGLE_APPLICATION_CREDENTIALS This is most suitable for applications like minio.

I have edited the help message accordingly. We can use the same way for our testing as well, i.e the admin will have to generate and give us credentials.json for testing.

and:

```
export GOOGLE_APPLICATION_CREDENTIALS=/home/krishna/Downloads/miniotest-d4530e7be663.json
minio gateway gcs PROJECTID
```

## Motivation and Context

https://github.com/minio/minio/issues/4391

## Types of changes

## Checklist:


balamurugana (commented):
What will happen if `gcloud init --console-only` is used instead of `GOOGLE_APPLICATION_CREDENTIALS`?

cmd/gateway-main.go
```diff
@@ -63,7 +63,8 @@ EXAMPLES:
       $ {{.HelpName}} s3 https://play.minio.io:9000
 
   4. Start minio gateway server for GCS backend.
-      $ gcloud init --console-only # this will configure your google account
+      $ export GOOGLE_APPLICATION_CREDENTIALS=/path/to/credentials.json
```
krisis: What a user needs to do for Minio GCS gateway to successfully authenticate with the GCS account depends on the environment it is being run. This message applies only when the gateway is run _outside_ GCP. Is it possible for the gateway to know if it's running in a GCP instance? If not, we could modify the message to,

> (When run outside Google Cloud Platform)
$ export GOOGLE_APPLICATION_CREDENTIALS=/path/to/credentials.json

krisis (changes requested):

harshavardhana:
@krishnasrinivas can you rebase this ?

cmd/gateway-main.go
krishnasrinivas: Not true @krisis this step works everywhere

krishnasrinivas:
This PR is temporary fix till we fix https://github.com/minio/minio/issues/4457

krishnasrinivas:
@harshavardhana done

harshavardhana:
Is this still valid @krishnasrinivas ?

PR #4426, opened 2017-05-25
[proposal] Support creating empty directories.

harshavardhana:
## Description
Every so often we get requirements for creating
directories/prefixes and we end up rejecting
such requirements. This PR is a propoal to show
that its possible to in-fact have empty directories
and be able to manage them like without much
backend format changes etc.

Two new additional APIs were implemented on StorageAPI
layer to facilitate this for distributed and standalone
erasure coded backend.

- DeletePrefix(vol, prefix)
- MakePrefix(vol, prefix)

DeletePrefix() deletes prefix and also its parent directories
if they are empty just like how DeleteFile() does.

MakePrefix() creates prefix all the time as if overwriting it
also creates all the top level directories as well along the
way. Behavior is similar to `mkdir -p`

## Motivation and Context
More S3 compatibility. 

## How Has This Been Tested?
Manually and using `go test` 

## Types of changes

## Checklist:

### DIFF
--- a/cmd/fs-v1-helpers.go
+++ b/cmd/fs-v1-helpers.go
@@ -123,6 +123,24 @@ func fsMkdir(dirPath string) (err error) {
 	return nil
 }
 
+func fsStatAll(statLoc string) (os.FileInfo, error) {
+	fi, err := fsStat(statLoc)
+	if err != nil {
+		err = errorCause(err)
+		if os.IsNotExist(err) {
+			return nil, traceError(errFileNotFound)
+		} else if os.IsPermission(err) {
+			return nil, traceError(errFileAccessDenied)
+		} else if isSysErrNotDir(err) {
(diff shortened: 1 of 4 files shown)


harshavardhana:
This needs to be blocked @deekoder until we discuss with @abperiasamy

krishnasrinivas:
Waiting for confirmation if this is :+1:  from @abperiasamy

harshavardhana:
I explored many ways to do this locally for erasure coding as well but it won't work since an empty directory without `xl.json` is considered as object that needs healing. Currently we can only support FS but it would lead to a different backend behavior for erasure coding if we are okay with allowing FS to look like gateway mode. 

We need to discuss this since having this feature has sort of lot of benefits, we will be able to support large number of use cases.

harshavardhana:
Closing this as not needed for now.

PR #4428, opened 2017-05-25
Refactor service stop signal message.

deekoder:
Changed to show : 
```
./minio server ~/Downloads/miniodata/

Endpoint:  http://192.168.1.225:9000  http://192.168.99.1:9000  http://127.0.0.1:9000
AccessKey: <> 
SecretKey:<>
Region:    us-east-1

Browser Access:
   http://192.168.1.225:9000  http://192.168.99.1:9000  http://127.0.0.1:9000

   $ mc config host add myminio http://192.168.1.225:9000 <> <>

   .NET:       https://docs.minio.io/docs/dotnet-client-quickstart-guide

Drive Capacity: 193 GiB Free, 465 GiB Total
^CExiting
```
## Description
Simple message change.

## Motivation and Context
Gracefully Stopping ... (press Ctrl+C again to force) was not accurate messaging.
 ```

Drive Capacity: 193 GiB Free, 465 GiB Total
^CGracefully stopping... (press Ctrl+C again to force)
```

## How Has This Been Tested?
 Manual testing of regular ^C and it should say "Exiting."
Also tested when a large file was being uploaded and tried to shut down server in while putobject was happening. Minio server still exits with the appropriate messaging and show press ^C again only if the exit still does not happen on time.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/service.go
+++ b/cmd/service.go
@@ -106,7 +106,7 @@ func (m *ServerMux) handleServiceSignals() error {
 			log.Println(" Exiting")
 			go func() {
 				time.Sleep(time.Millisecond * 600)
-				log.Println("Press ^C again to exit immediately")
+				log.Println("Waiting for active connections to terminate, press ^C again to exit immediately.")
 			}()
 			if err := m.Close(); err != nil {
 				errorIf(err, "Unable to close server gracefully")
--- a/cmd/service.go
+++ b/cmd/service.go
@@ -102,7 +102,8 @@ func (m *ServerMux) handleServiceSignals() error {
 			}
 			runExitFn(nil)
 		case serviceStop:
-			log.Println("Exiting")
+			// Extra space will prevent showing "^CExiting" and rather show "^C Exiting" instead.
+			log.Println(" Exiting")
 			go func() {
 				time.Sleep(time.Millisecond * 600)
 				log.Println("Press ^C again to exit immediately")
--- a/cmd/service.go
+++ b/cmd/service.go
@@ -102,11 +102,10 @@ func (m *ServerMux) handleServiceSignals() error {
 			}
 			runExitFn(nil)
 		case serviceStop:
-			// Extra space will prevent showing "^CExiting" and rather show "^C Exiting" instead.
-			log.Println(" Exiting")
+			log.Println("Received signal to exit.")
 			go func() {
-				time.Sleep(time.Millisecond * 600)
-				log.Println("Waiting for active connections to terminate, press ^C again to exit immediately.")
+				time.Sleep(serverShutdownPoll + time.Millisecond*100)
+				log.Println("Waiting for active connections to terminate - press Ctrl+C to quit immediately.")
 			}()
 			if err := m.Close(); err != nil {
 				errorIf(err, "Unable to close server gracefully")
--- a/cmd/service.go
+++ b/cmd/service.go
@@ -102,7 +102,7 @@ func (m *ServerMux) handleServiceSignals() error {
 			}
 			runExitFn(nil)
 		case serviceStop:
-			log.Println("Exiting.")
+			log.Println("Exiting")
 			go func() {
 				time.Sleep(time.Millisecond * 600)
 				log.Println("press ^C again")
--- a/cmd/service.go
+++ b/cmd/service.go
@@ -105,7 +105,7 @@ func (m *ServerMux) handleServiceSignals() error {
 			log.Println("Exiting")
 			go func() {
 				time.Sleep(time.Millisecond * 600)
-				log.Println("press ^C again")
+				log.Println("Press ^C again to exit immediately")
 			}()
 			if err := m.Close(); err != nil {
 				errorIf(err, "Unable to close server gracefully")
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by harshavardhana — COMMENTED** [review]
@krishnasrinivas @donatello can you guys give your feedback.. from original issue?

File: cmd/service.go [hunk 1]
```diff
@@ -102,7 +102,11 @@ func (m *ServerMux) handleServiceSignals() error {
 			}
 			runExitFn(nil)
 		case serviceStop:
-			log.Println("Gracefully stopping... (press Ctrl+C again to force)")
+			log.Println("Exiting.")
+			go func() {
+				time.Sleep(time.Millisecond * 600)
```
krishnasrinivas: You can put `serverShutdownPoll + time.Mullisecond*100`

cmd/service.go
```diff
@@ -102,7 +102,11 @@ func (m *ServerMux) handleServiceSignals() error {
...
+				log.Println("press ^C again")
```
krishnasrinivas: `press ^C again`

Just saying this will make the user think "why is it asking me to press ^C again?"

You can put the message as:
`Waiting for active connections to terminate, press ^C again to stop.`

krishnasrinivas (approved):

poornas (commented):
"press ^C again" is a vague message as well - 
maybe "press ^C again to exit immediately" would be clearer. 

And ... after the Exiting to show server exit in progress.

cmd/service.go
```diff
@@ -102,7 +102,11 @@ func (m *ServerMux) handleServiceSignals() error {
 			}
 			runExitFn(nil)
 		case serviceStop:
-			log.Println("Gracefully stopping... (press Ctrl+C again to force)")
+			log.Println("Exiting")
```
krishnasrinivas: Also a space before "Exiting" will be nice so that it prints
^C Exiting

and not

^CExiting

deekoder:
As per @krishnasrinivas  it would be more self explanatory to say time.Sleep(serverShutdownPoll + time.Millisecond*100). Will incorporate this since it helps readability further.

cmd/service.go
```diff
@@ -102,7 +102,12 @@ func (m *ServerMux) handleServiceSignals() error {
 			}
 			runExitFn(nil)
 		case serviceStop:
-			log.Println("Gracefully stopping... (press Ctrl+C again to force)")
+			// Extra space will prevent showing "^CExiting" and rather show "^C Exiting" instead.
+			log.Println(" Exiting")
```
donatello: The `^C` is not the only way to get to get to this line in the code. A SIGTERM may also be sent to cause this line to execute. In that case, printing with a leading space is weird - also, there is no info on why it is exiting.

cmd/service.go
```diff
@@ -102,7 +102,12 @@ func (m *ServerMux) handleServiceSignals() error {
...
+			go func() {
+				time.Sleep(time.Millisecond * 600)
+				log.Println("Waiting for active connections to terminate, press ^C again to exit immediately.")
```
donatello: There appears to be another case - the current signal handling setup does not distinguish between a keyboard interrupt and being sent SIGTERM - they are not the same thing. It's actually incorrect to say  `press ^C again`, when the user may not have pressed `^C` in the first place. Also `^C` is not a widely intelligible term - it should be `Ctrl+C`.

donatello (changes requested):

cmd/service.go
krishnasrinivas: That is OK, because otherwise we will have to write separate signal handlers just for this.

In case you have better alternative solution we can go for it.

krishnasrinivas: > also, there is no info on why it is exiting.

any suggestions?

cmd/service.go
krishnasrinivas: We can print just `Waiting for active connections to terminate` in that case.

cmd/service.go
donatello: Separate signal handlers is not that difficult, but perhaps we can settle for a more generic message like, `Received exit signal (Ctrl+C/SIGTERM/etc)! Attempting to gracefully exit.`

cmd/service.go
donatello: Alternatively, `Waiting for active connections to terminate - press Ctrl+C to quit immediately` would also be fine.

cmd/service.go
krishnasrinivas: `Received exit signal (Ctrl+C/SIGTERM/etc)! Attempting to gracefully exit.` is too long and verbose.
`Received signal to exit.` will be better

deekoder: Received signal to exit is fine too. But simple "Exiting" is even better imo.

cmd/service.go
deekoder: Ok making this change

harshavardhana (approved):

donatello (approved):

cmd/service.go
krishnasrinivas: I would prefer a simple `Exiting` too

PR #4431, opened 2017-05-25
GCS does manual params validation

### LINKED ISSUE #4396: [gateway-gcs] GCS doesn't ask for projectID when not provided
## Expected Behavior
Error message should be gcs specific, there are no credentials provided with GCS.

## Current Behavior

Asks for users to set `Access and secret keys are mandatory to run Minio gateway server`

```
$ minio gateway gcs 
FATA[0000] Access and secret keys are mandatory to run Minio gateway server.  cause=Missing credentials source=[gateway-main.go:102:mustGetGatewayConfigFromEnv()]
```

## Possible Solution
Fix the error message.

Just type `gcs` without arguments.
```
minio gateway gcs
``` 

## Context
GCS testing. 

## Your Environment
```
minio version
Version: 2017-05-23T02:27:58Z
Release-Tag: DEVELOPMENT.2017-05-23T02-27-58Z
Commit-ID: 37549f20b3b1d2d618f7944895c745b3c2092427
```

vadmeste:
## Description
To control the order of validation, gcs gateway checks project id
then access/secret keys.

```
$ minio gateway gcs
FATA[0000] Unable to initialize gateway layer            cause=GCS project ID expected source=[gateway-main.go:222:gatewayMain()]
```

## Motivation and Context
Fixed #4396 

## How Has This Been Tested?
go test + Manual

## Types of changes

- [] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

### DIFF
--- a/cmd/gateway-gcs.go
+++ b/cmd/gateway-gcs.go
@@ -22,9 +22,11 @@ import (
 	"encoding/base64"
 	"encoding/hex"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"hash"
 	"io"
+	"os"
 	"regexp"
 	"strings"
 	"time"
@@ -207,6 +209,14 @@ const googleStorageEndpoint = "storage.googleapis.com"
 // newGCSGateway returns gcs gatewaylayer
 func newGCSGateway(projectID string) (GatewayLayer, error) {
 	ctx := context.Background()
+
+	if !isValidGCSProjectID(projectID) {
+		fatalIf(errGCSInvalidProjectID, "Unable to initialize GCS gateway")
+	}
+
+	if os.Getenv("MINIO_ACCESS_KEY") == "" || os.Getenv("MINIO_SECRET_KEY") == "" {
+		return nil, errors.New("No credentials set")
+	}
 
 	// Initialize a GCS client.
 	client, err := storage.NewClient(ctx)
--- a/cmd/gateway-startup-msg.go
+++ b/cmd/gateway-startup-msg.go
@@ -50,8 +50,10 @@ func printGatewayCommonMsg(apiEndpoints []string) {
 	apiEndpointStr := strings.Join(apiEndpoints, "  ")
 	// Colorize the message and print.
 	log.Println(colorBlue("\nEndpoint: ") + colorBold(fmt.Sprintf(getFormatStr(len(apiEndpointStr), 1), apiEndpointStr)))
-	log.Println(colorBlue("AccessKey: ") + colorBold(fmt.Sprintf("%s ", cred.AccessKey)))
-	log.Println(colorBlue("SecretKey: ") + colorBold(fmt.Sprintf("%s ", cred.SecretKey)))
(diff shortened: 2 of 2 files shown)


cmd/gateway-gcs.go
```diff
@@ -183,6 +185,10 @@ func newGCSGateway(args cli.Args) (GatewayLayer, error) {
 		fatalIf(errGCSInvalidProjectID, "Unable to initialize GCS gateway")
 	}
 
+	if os.Getenv("MINIO_ACCESS_KEY") == "" || os.Getenv("MINIO_SECRET_KEY") == "" {
+		return nil, errors.New("No GCS credentials set")
```
harshavardhana: `No credentials set` is correct instead of `No GCS credentials set`

harshavardhana (changes requested):

cmd/gateway-gcs.go
vadmeste: oups yes.

vadmeste: > No credentials set is correct instead of No GCS credentials set

Done

vadmeste:
Comments addressed and rebased.

harshavardhana:
Can you rebase this PR @vadmeste ?

vadmeste:
> Can you rebase this PR @vadmeste ?

Done

krishnasrinivas (approved):
Can you rebase @vadmeste

vadmeste:
> Can you rebase @vadmeste

@krishnasrinivas: I rebased, but since this PR is quite old, we still need to see if we need it.

Currently this PR checks if gcs project id is provided and if access/secret keys are set in the environment. In contrary to S3 and Azure gateways which can work with credentials from config.json and not necessarily from the environment.

We need to check if we still need this.

harshavardhana:
> n contrary to S3 and Azure gateways which can work with credentials from config.json and not necessarily from the environment.

Actually it is expected that for AWS S3 and Azure credentials are always provided through ENV regardless of them being saved in the config.json, GCS also has same requirements.

You can test and update us here if the feature-gcs branch is doing all the things which are handled here.

vadmeste:
@harshavardhana, all gateways in feature-gcs branch don't require access and secret keys set in the environment.

harshavardhana:
Then its a bug @vadmeste . All gateways require env to be set regardless of saving them in config file. Config file was introduced for configs like bucket notifications which we are going to support next. But setting envs is a must and it should fail otherwise.

vadmeste:
> Then its a bug @vadmeste . All gateways require env to be set regardless of saving them in config file. Config file was introduced for configs like bucket notifications which we are going to support next. But setting envs is a must and it should fail otherwise.

Minio master requires access/secret keys from env when runnig s3 and azure, but feature-gcs branch doesn't. This PR also requires keys from env. So I guess things will be fine once feature-gcs got merged.

vadmeste:
Overall, I think we can merge this PR now if no more comments.

PR #4438, opened 2017-05-26
xl: Fix rare freeze after many disk/network errors

vadmeste:
## Description
xl.storageDisks is passed to some low-level XL functions where some disks of
xl.storageDisks are set to nil when they encounter some errors. This means all
elements in xl.storageDisks will be nil after some time which lead to an unusable XL.

## Motivation and Context
Manual testing.

I manually modified posix.go to return randomly errors before I discovered this bug.

## How Has This Been Tested?
Manual testing

## Types of changes

## Checklist:


cmd/xl-v1-utils.go
```diff
@@ -357,6 +357,23 @@ func shuffleDisks(disks []StorageAPI, distribution []int) (shuffledDisks []Stora
 	return shuffledDisks
 }
 
+// evalDisks - returns a new slice of disks where nil is set if
+// the correspond error in errs slice is not nil
+func evalDisks(disks []StorageAPI, errs []error) []StorageAPI {
+	if len(errs) != len(disks) {
```
harshavardhana, line 363: This should never happen, returning a nil should accompany a log message here with errorIf() indicating something really unexpected has happened.

cmd/xl-v1-utils.go
```diff
@@ -357,6 +357,23 @@ func shuffleDisks(disks []StorageAPI, distribution []int) (shuffledDisks []Stora
...
+		return nil
```
harshavardhana: Add a test case for this as well separately with its own unit tests..

harshavardhana (changes requested):

vadmeste:
Comments adressed

cmd/xl-v1-utils_test.go
```diff
@@ -429,3 +430,53 @@ func testShuffleDisks(t *testing.T, xl *xlObjects) {
...
+
+	diskErr := errors.New("some disk error")
+	errs := []error{
+		diskErr, nil, nil, nil,
+		nil, diskErr, nil, nil,
+		diskErr, nil, nil, nil,
+		nil, nil, nil, diskErr,
+	}
+
+	newDisks := evalDisks(disks, errs)
```
harshavardhana: why not have a test for errs to be smaller than disks?

harshavardhana (changes requested):

vadmeste:
Fixed, can we also add more reviewers ? @deekoder

harshavardhana (dismissed):

krishnasrinivas (commented):
@vadmeste xl.storageDisks is never supposed to get modified because listOnlineDisks() and shuffleDisks() always create new slice. Need to review where exactly xl.storageDisks get modified, can you check?

vadmeste:
> @vadmeste xl.storageDisks is never supposed to get modified because listOnlineDisks() and shuffleDisks() always create new slice. Need to review where exactly xl.storageDisks get modified, can you check?

https://github.com/minio/minio/pull/4438/commits/2b5e0ac17eecc8c825487e59ed16775e3684432d#diff-949516a155a3e0e700699d2ff77d6fb9L504

This is one example.

I still think the new style is better than the old one. Modifying a slice deep inside many function levels is not good as the developer will always forget about this behavior.

krishnasrinivas:
```
577                 partSizeWritten, checkSums, erasureErr := erasureCreateFile(onlineDisks, minioMetaTmpBucket, tempErasureObj, partReader\
    , allowEmptyPart, partsMetadata[0].Erasure.BlockSize, partsMetadata[0].Erasure.DataBlocks, partsMetadata[0].Erasure.ParityBlocks, bitRo\
    tAlgo, xl.writeQuorum)
```

Since this does not return `[]StorageAPI` we might end up with missing parts because later we do:
```
		if onlineDisks, err = renameXLMetadata(onlineDisks, minioMetaTmpBucket, tempObj, srcBucket, srcObject, xl.writeQuorum); err != nil {
  			return ObjectInfo{}, toObjectErr(err, srcBucket, srcObject)		  			return ObjectInfo{}, toObjectErr(err, srcBucket, srcObject)
  		}		  		}
```

I will be reviewing more.

krishnasrinivas:
We can make sure that we follow this pattern everywhere:

```
function() {                                                                                                                                                                                                       
        onlineDisks, err := f(onlineDisks, ....)                                                                                                                                                                   
        onlineDisks, err := g(onlineDisks, ....)                                                                                                                                                                   
        onlineDisks, err := h(onlineDisks, ....)                                                                                                                                                                   
}
```

There are two exceptions, one in PutObject and the other in CompleteMultipartUpload where it uses xl.storageDisks to purge the previously uploaded object (before renaming from tmp location to the actual location)

Here are the list of comments:

* erasureCreateFile should return `disks`
* In PutObject and CompleteMultipartUpload, after erasureCreateFile() subsequent calls should use the `disks` returned by erasureCreateFile
* In newMultipartUpload() you are ignoring returned disks from writeSameXLMetadata() and renameObject() There is no harm in this particular situation. But to be consistent you need not ignore, you can use the disks returned by writeSameXLMetadata() in renameObject() to be consistent. Also it will be more performant as renameObject() will not try on disks that are down.

vadmeste:
All your comments are addressed @krishnasrinivas

krishnasrinivas (approved):

krisis:
@vadmeste could you provide the steps to manually test this PR? Also, what is the behaviour user observes when more than quorum number of `xl.StorageDisks` are `nil` in current master (in your manual testing)?

vadmeste:
> could you provide the steps to manually test this PR? Also, what is the behaviour user observes when more than quorum number of xl.storageDisks are nil in current master (in your manual testing)?

I'll post the patch in some minutes. I simply added some lines in posix.go which returns an error with a chance of 10% in some calls (rename, delete, append, ..)

vadmeste:
[occasional_disk_error.patch.txt](https://github.com/minio/minio/files/1052063/occasional_disk_error.patch.txt)

This is the patch (remove .txt) which returns fake disk errors.

To reproduce:
```
$ cd $GOPATH/src/github.com/minio/minio
$ patch -p1 < occasional_disk_error.patch
$ go install
$ minio server /tmp/xl{1..4}
```
In another terminal
```
$ mc mb myminio/testbucket
$ dd if=/dev/urandom of=6mb.file bs=1024 count=$[1024*6]
$ while true; do mc cp 6mb.file myminio/testbucket ; done
```
After some time, you'll notice that mc will never be able to upload again.

By the way, I've just noticed that I put freeze word to describe the issue, it is not 100% correct, I meant  unusable instead.

krisis (approved):
Tested locally with @vadmeste patch in comments.

harshavardhana (approved):

PR #4443, opened 2017-05-29
log: Startup banner should strip standard ports.

harshavardhana:
## Description
APIEndpoints list should strip off standard ports
to avoid confusion with clients.

## Motivation and Context
Printing consistent useful information.

## How Has This Been Tested?
Manually and with `go test`

## Types of changes

## Checklist:

### DIFF
--- a/cmd/server-startup-msg.go
+++ b/cmd/server-startup-msg.go
@@ -19,6 +19,7 @@ package cmd
 import (
 	"crypto/x509"
 	"fmt"
+	"net/url"
 	"runtime"
 	"strings"
 
@@ -44,12 +45,14 @@ func getFormatStr(strLen int, padding int) string {
 // Prints the formatted startup message.
 func printStartupMessage(apiEndPoints []string) {
 
+	strippedAPIEndpoints := stripStandardPorts(apiEndPoints)
+
 	// Prints credential, region and browser access.
-	printServerCommonMsg(apiEndPoints)
+	printServerCommonMsg(strippedAPIEndpoints)
 
 	// Prints `mc` cli configuration message chooses
 	// first endpoint as default.
-	printCLIAccessMsg(apiEndPoints[0])
+	printCLIAccessMsg(strippedAPIEndpoints[0])
 
 	// Prints documentation message.
 	printObjectAPIMsg()
@@ -67,6 +70,34 @@ func printStartupMessage(apiEndPoints []string) {
 	}
 }
 
+// strip api endpoints list with standard ports such as
+// port "80" and "443" before displaying on the startup
+// banner.  Returns a new list of API endpoints.
+func stripStandardPorts(apiEndpoints []string) (newAPIEndpoints []string) {
+	newAPIEndpoints = make([]string, len(apiEndpoints))
+	// Check all API endpoints for standard ports and strip them.
+	for i, apiEndpoint := range apiEndpoints {
+		url, err := url.Parse(apiEndpoint)
+		if err != nil {
+			newAPIEndpoints[i] = apiEndpoint
+			continue
+		}
+		host, port := mustSplitHostPort(url.Host)
+		// For standard HTTP(s) ports such as "80" and "443"
+		// apiEndpoints should only be host without port.
+		switch {
+		case url.Scheme == "http" && port == "80":
+			fallthrough
+		case url.Scheme == "https" && port == "443":
+			url.Host = host
+			newAPIEndpoints[i] = url.String()
+		default:
+			newAPIEndpoints[i] = apiEndpoint
+		}
+	}
+	return newAPIEndpoints
+}
+
 // Prints common server startup message. Prints credential, region and browser access.
 func printServerCommonMsg(apiEndpoints []string) {
 	// Get saved credentials.
@@ -76,6 +107,7 @@ func printServerCommonMsg(apiEndpoints []string) {
 	region := serverConfig.GetRegion()
 
 	apiEndpointStr := strings.Join(apiEndpoints, "  ")
+
 	// Colorize the message and print.
 	log.Println(colorBlue("\nEndpoint: ") + colorBold(fmt.Sprintf(getFormatStr(len(apiEndpointStr), 1), apiEndpointStr)))
 	log.Println(colorBlue("AccessKey: ") + colorBold(fmt.Sprintf("%s ", cred.AccessKey)))
--- a/cmd/server-startup-msg_test.go
+++ b/cmd/server-startup-msg_test.go
@@ -20,6 +20,7 @@ import (
 	"crypto/x509"
 	"crypto/x509/pkix"
 	"fmt"
+	"reflect"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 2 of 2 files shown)


balamurugana (changes requested):
> APIEndpoints list should strip off standard ports
> to avoid confusion with clients.

I am unable to understand the `confusion` part.  What exactly we are trying to solve?

harshavardhana:
> I am unable to understand the confusion part. What exactly we are trying to solve?

The reasoning is that standard ports should never be printed as part of the banner, some clients for example aws-sdk-go never strips of port '80' and it causes signature failure with presigned URLs. Users not aware of this detail they just end up using it directly. 

curl and browsers strip off ports for 80 and 443.  So you in-turn get signature mismatch since on server side we just look for what are available in the headers.

```go
package main

import (
	"fmt"
	"time"

	"github.com/aws/aws-sdk-go/aws"
	"github.com/aws/aws-sdk-go/aws/credentials"
	"github.com/aws/aws-sdk-go/aws/session"
	"github.com/aws/aws-sdk-go/service/s3"
)

func main() {
	creds := credentials.NewStaticCredentials("USWUXHGYZQYFYFFIT3RE", "MOJRH0mkL1IPauahWITSVvyDrQbEEIwljvmxdq03", "")
	newSession := session.New()
	s3Config := &aws.Config{
		Credentials:      creds,
		Endpoint:         aws.String("http://10.0.0.13:80"),
		Region:           aws.String("us-east-1"),
		DisableSSL:       aws.Bool(true),
		S3ForcePathStyle: aws.Bool(true),
	}
	// Create an S3 service object in the default region.
	s3Client := s3.New(newSession, s3Config)
	params := &s3.PutObjectInput{
		Bucket: aws.String("testbucket"),
		Key:    aws.String("object"),
	}
	req, _ := s3Client.PutObjectRequest(params)
	url, err := req.Presign(10 * time.Minute)
	if err != nil {
		fmt.Println(err)
		return
	}
	fmt.Println(url)

	gparams := &s3.GetObjectInput{
		Bucket: aws.String("testbucket"),
		Key:    aws.String("object"),
	}
	greq, _ := s3Client.GetObjectRequest(gparams)
	url, err = greq.Presign(10 * time.Minute)
	if err != nil {
		fmt.Println(err)
		return
	}
	fmt.Println(url)

}
```

```
ERRO[0043] {"method":"GET","path":"/testbucket/object","query":"X-Amz-Algorithm=AWS4-HMAC-SHA256\u0026X-Amz-Credential=USWUXHGYZQYFYFFIT3RE%!F(MISSING)20170529%!F(MISSING)us-east-1%!F(MISSING)s3%!F(MISSING)aws4_request\u0026X-Amz-Date=20170529T190139Z\u0026X-Amz-Expires=600\u0026X-Amz-SignedHeaders=host\u0026X-Amz-Signature=19b58080999df54b446fc97304eb8dda60d3df1812ae97f3e8783351bfd9781d","header":{"Accept":["*/*"],"Host":["10.0.0.13"],"User-Agent":["curl/7.47.0"]}}  cause=Signature does not match source=[auth-handler.go:122:checkRequestAuthType()]
```

Internally we can always use port, but this is a startup banner and useful for the clients to know what is the right endpoint. It is better to show them in canonical form. 

A similar type of issue happens if you use to `Endpoint` as `http://s3.amazonaws.com:80` @balamurugana

cmd/server-startup-msg.go
```diff
@@ -67,6 +71,31 @@ func printStartupMessage(apiEndPoints []string) {
...
+	for i, apiEndpoint := range apiEndpoints {
+		url, err := url.Parse(apiEndpoint)
+		if err != nil {
+			newAPIEndpoints[i] = apiEndpoint
+			continue
+		}
+		host, port, _ := net.SplitHostPort(url.Host)
+		// For standard HTTP(s) ports such as "80" and "443"
+		// apiEndpoints should only be host without port.
+		if port == "80" || port == "443" {
```
vadmeste: Can you enforce scheme to http if port is 80 and https if port is 443, even when no tls certificates are provided. I know this is controversial but let's imagine if the user enters --address ":443", in that case :443 will be stripped and the resulted url will still be http://host and not https://host.

harshavardhana: Yes did that.. good catch @vadmeste

pkg/x/os/stat_windows_test.go
```diff
@@ -71,6 +71,9 @@ func sameFile(fi1, fi2 os1.FileInfo) bool {
 }
 
 func TestNetworkSymbolicLink(t *testing.T) {
+	// Skipping this test for now - appveyor builds are failing.
+	t.Skip()
```
balamurugana: Why is it required to change in this PR?  I would recommend to fix it properly in separate PR than skipping it.

cmd/server-startup-msg.go
```diff
@@ -67,6 +71,34 @@ func printStartupMessage(apiEndPoints []string) {
...
+func stripStandardPorts(apiEndpoints []string) (newAPIEndpoints []string) {
+	newAPIEndpoints = make([]string, len(apiEndpoints))
+	// Check all API endpoints for standard ports and strip them.
+	for i, apiEndpoint := range apiEndpoints {
+		url, err := url.Parse(apiEndpoint)
+		if err != nil {
+			newAPIEndpoints[i] = apiEndpoint
+			continue
+		}
+		host, port, _ := net.SplitHostPort(url.Host)
```
balamurugana: you could use `mustSplitHostPort()` than skipping error

cmd/server-startup-msg.go
```diff
@@ -67,6 +71,34 @@ func printStartupMessage(apiEndPoints []string) {
...
+		// For standard HTTP(s) ports such as "80" and "443"
+		// apiEndpoints should only be host without port.
+		switch {
+		case url.Scheme == "http" && port == "80":
+			fallthrough
+		case url.Scheme == "https" && port == "443":
+			url.Host = host
+			newAPIEndpoints[i] = url.String()
```
balamurugana, line 93: This means its possible to have valid endpoints `https://server:80` and `http://server:443`.  If TLS is configured in minio it works for both secure/insecure mode where insecure redirects to secure port  ie what is the behavior for `https://server:80` is accessed by `http://server:80`?

balamurugana (changes requested):

balamurugana:
> The reasoning is that standard ports should never be printed as part of the banner, some clients for example aws-sdk-go never strips of port '80' and it causes signature failure with presigned URLs. Users not aware of this detail they just end up using it directly.

Isn't it a bug in aws-sdk-go?  Not sure how much useful this PR for such buggy clients.

pkg/x/os/stat_windows_test.go
harshavardhana: We should skip it because the issue is unknown, and started happening regardless of any change on our end. Will send a separate PR for it, i wanted this PR to be built properly.

cmd/server-startup-msg.go
harshavardhana: I don't expect port to exist always, mustSplitHostPort() would crash the server.

cmd/server-startup-msg.go
harshavardhana: This i am not sure.. will check.

harshavardhana:
> Isn't it a bug in aws-sdk-go? Not sure how much useful this PR for such buggy clients.

Yes it is a bug in aws-sdk-go. But my point is not to address buggy clients its regardless of that we shouldn't print what we know is not the norm. Confusion arises when a norm is not followed. API publishers for example do not print out standard ports as part of the endpoints, it is unexpected from a users perspective which is why we should always print what is right for users not from what is convenient for us.

harshavardhana:
> This means its possible to have valid endpoints https://server:80 and http://server:443. If TLS is configured in minio it works for both secure/insecure mode where insecure redirects to secure port ie what is the behavior for https://server:80 is accessed by http://server:80?

So 

- port with "443" works without SSL 
- port with "80" only works with SSL

@balamurugana ^^

balamurugana:
>> This means its possible to have valid endpoints https://server:80 and http://server:443. If TLS is configured in minio it works for both secure/insecure mode where insecure redirects to secure port ie what is the behavior for https://server:80 is accessed by http://server:80?
> 
> So
> 
>     port with "443" works without SSL
>     port with "80" only works with SSL
> 
> @balamurugana ^^

I know the meaning of those endpoints.  My question is how redirection works in those cases.  I suspect browser/some clients will fail to connect because 80/443 are meant for insecure/secure only.

harshavardhana:
> I know the meaning of those endpoints. My question is how redirection works in those cases. I suspect browser/some clients will fail to connect because 80/443 are meant for insecure/secure only.

I just tested that @balamurugana - that is not true, browsers work fine and do not employ redirection the way you are mentioning. Neither do they assume ports based on just the scheme.  If port is specified it is honored regardless of the scheme in question.

harshavardhana:
> I just tested that @balamurugana - that is not true, browsers work fine and do not employ redirection the way you are mentioning. Neither do they assume ports based on just the scheme. If port is specified it is honored regardless of the scheme in question.

The problem you are mentioning exists if accessed with "http://ip:80" on a https configured server, we are re-directing wrongly.

harshavardhana:
> The problem you are mentioning exists if accessed with "http://ip:80" on a https configured server, we are re-directing wrongly.

Sent a separate PR to fix this issue #4445

donatello (approved):
LGTM :+1:

harshavardhana:
@balamurugana do you have any pending concerns?

balamurugana (approved):

PR #4445, opened 2017-05-30
server: Redirection should use globalMinioPort with host without port.

harshavardhana:
## Description
Currently redirection doesn't work in following scenarios

 - server started with port ":80" and TLS is configured
   client requested insecure request on port "80"
   gets redirected to port 443 and fails.

## Motivation and Context
Refer discussion on #4443 

## How Has This Been Tested?
Manually.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/server-mux.go
+++ b/cmd/server-mux.go
@@ -449,12 +449,20 @@ func (m *ServerMux) ListenAndServe(certFile, keyFile string) (err error) {
 	// All http requests start to be processed by httpHandler
 	httpHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		if tlsEnabled && r.TLS == nil {
+			// It is expected that r.Host might not have port
+			// for standard ports such as "80" and "443".
+			host, port, _ := net.SplitHostPort(r.Host)
+			if port == "" {
+				host = net.JoinHostPort(r.Host, globalMinioPort)
+			} else {
+				host = r.Host
+			}
 			// TLS is enabled but Request is not TLS configured
 			u := url.URL{
 				Scheme:   httpsScheme,
 				Opaque:   r.URL.Opaque,
 				User:     r.URL.User,
-				Host:     r.Host,
+				Host:     host,
 				Path:     r.URL.Path,
 				RawQuery: r.URL.RawQuery,
 				Fragment: r.URL.Fragment,


donatello:
I think this behaviour is making the server insecure - if the client is sent an error instead of being transparently redirected, then the client can fix their misconfiguration (of using http against a https server). Each such attempt of accessing a https server over http is a security issue because credentials info is sent in the clear (either the V4 authorization header or the browser's JSON RPC based login info).

In case of the V4 header, it is considered secure to be transmitted in the clear, but still the server-operator would be expecting it to be transport-layer-encrypted. In case of the JSON RPC based browser login, it is positively insecure.

The initial request over http is sending credentials in the clear, then the client is being sent a HTTP redirect, and the client just follows that and completes its request and doesnt realize that there is a configuration/security issue going on. The initial http request has potentially compromised the credentials, and this keeps happening at _every_ request - because the client does not realize the problem.

The better way is to immediately return an error if the request is not TLS encrypted and the server is configured with TLS. This way, the client can at least fix their configuration. Otherwise, this is a insecure-by-default kind of configuration.

harshavardhana:
> I think this behaviour is making the server insecure - if the client is sent an error instead of being transparently redirected, then the client can fix their misconfiguration (of using http against a https server). Each such attempt of accessing a https server over http is a security issue because credentials info is sent in the clear (either the V4 authorization header or the browser's JSON RPC based login info).
> 
> In case of the V4 header, it is considered secure to be transmitted in the clear, but still the server-operator would be expecting it to be transport-layer-encrypted. In case of the JSON RPC based browser login, it is positively insecure.
> 
> The initial request over http is sending credentials in the clear, then the client is being sent a HTTP redirect, and the client just follows that and completes its request and doesnt realize that there is a configuration/security issue going on. The initial http request has potentially compromised the credentials, and this keeps happening at every request - because the client does not realize the problem.
> 
> The better way is to immediately return an error if the request is not TLS encrypted and the server is configured with TLS. This way, the client can at least fix their configuration. Otherwise, this is a insecure-by-default kind of configuration.

That is a larger issue @donatello this PR is not meant to solve that.  Re-direction was agreed that is good to have since AWS Signature v4 is secure by itself.  

Now regarding the browser we have agreed to do it differently using a similar signature technique.

-- Harsha

vadmeste (approved):
LGTM & tested.

donatello:
> That is a larger issue @donatello this PR is not meant to solve that. Re-direction was agreed that is good to have since AWS Signature v4 is secure by itself.

> Now regarding the browser we have agreed to do it differently using a similar signature technique.

I agree this PR is not intended to address the potential security issue. Let me think about it a bit more - I will create a new issue if further discussion is needed.

balamurugana:
@donatello 

> The initial request over http is sending credentials in the clear, then the client is being sent a HTTP redirect, and the client just follows that and completes its request and doesnt realize that there is a configuration/security issue going on. The initial http request has potentially compromised the credentials, and this keeps happening at every request - because the client does not realize the problem.
> 
> The better way is to immediately return an error if the request is not TLS encrypted and the server is configured with TLS. This way, the client can at least fix their configuration. Otherwise, this is a insecure-by-default kind of configuration.

:+1:  This is what I was thinking when https://github.com/minio/minio-java/pull/571 and https://github.com/minio/minio/pull/4217

As this is design decision @abperiasamy needs to clarify on this.

balamurugana (approved):

PR #4463, opened 2017-06-02
Add support for reading and saving config on Gateway.

harshavardhana:
## Description
Add support for reading and saving config on Gateway.

## Motivation and Context
This is also a first step towards supporting bucket notification for gateway.

## How Has This Been Tested?
Manually.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/gateway-azure.go
+++ b/cmd/gateway-azure.go
@@ -26,7 +26,6 @@ import (
 	"io"
 	"net/http"
 	"net/url"
-	"os"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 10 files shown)


poornas (changes requested):
Tested with azure - config gets saved, but when the environment vars are unset, gateway server doesn't use saved credentials to  start server.

harshavardhana:
> Tested with azure - config gets saved, but when the environment vars are unset, gateway server doesn't use saved credentials to start server.

Should be fixed now.. @poornas

poornas:
Saved config gets used now when gateway credentials are stored in the config.json. 

However, server generates random access key/secret key when gateway server is started for the first time( both config.json and env. vars are not present). This is misleading behavior and should instead fail if env. variables are missing.

poornas (approved):
LGTM.

krishnasrinivas (approved):

PR #4465, opened 2017-06-02
Bypass network in lock requests to local server

### LINKED ISSUE #4451: Unable to create a bucket with two Minio instances (4 disks) + self-signed certificates
It is a weird behavior and I could be missing something but I was not able to create a bucket in my two Minio instances when https is enabled. It works if I disable https. I saw that the code is blocked in bucket  namespace lock (https://github.com/minio/minio/blob/master/cmd/bucket-handlers.go#L388)

It is important to test first and see if you can reproduce it.  

## Expected Behavior
To be able to create a bucket in a cluster of two Minio instances

## Current Behavior
mc mb doesn't respond. Code blocks in https://github.com/minio/minio/blob/master/cmd/bucket-handlers.go#L388

## Possible Solution
No idea

1. Prepare two self signed certificates
2. Run two instances of Minio servers, each instance contains two disks, both are configured with tls certificates. Do not forget to copy public certificates in CAs directory of both instances.
3. mc mb myminio/testbucket

## Context
Testing

## Your Environment
- master
- Ubuntu 16.04
- Bare metal

donatello:
This makes lock RPCs similar to other RPCs where requests to the local
server bypass the network. Requests to the local lock-subsystem may
bypass the network layer and directly access the locking
data-structures.

This incidentally fixes #4451.

## Description

With this change, there should finally be no more scenarios in which the minio server sends a request to itself via the network - instead it directly accesses the associated data-structures in-process.

This resolves #4451 - the CA for own TLS certificate need not be added as a trusted root certificate. WIthout this change, it is required for the sole (unnecessary) purpose of making network requests to same process.

## Motivation and Context

Makes all RPCs similar - remote requests actually perform a RPC call, and local requests directly access the concerned data-structures in-process, bypassing the network.

## How Has This Been Tested?

~~Manually tested - but some tests are yet to be done. **Please do not merge yet.**~~
**EDIT:** Manually tested and validated. Ready for review.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/namespace-lock.go
+++ b/cmd/namespace-lock.go
@@ -27,6 +27,9 @@ import (
 // Global name space lock.
 var globalNSMutex *nsLockMap
 
+// Global lock servers
+var globalLockServers []*lockServer
+
 // RWLocker - locker interface extends sync.Locker
 // to introduce RLock, RUnlock.
 type RWLocker interface {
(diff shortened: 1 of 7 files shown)


cmd/namespace-lock.go
```diff
@@ -42,15 +45,21 @@ func initDsyncNodes() error {
...
 	for index, endpoint := range globalEndpoints {
-		clnts[index] = newLockRPCClient(authConfig{
-			accessKey:       cred.AccessKey,
-			secretKey:       cred.SecretKey,
-			serverAddr:      endpoint.Host,
-			secureConn:      globalIsSSL,
-			serviceEndpoint: pathutil.Join(minioReservedBucketPath, lockServicePath, endpoint.Path),
-			serviceName:     lockServiceName,
-		})
+		if !endpoint.IsLocal {
```
harshavardhana: You can use `if endpoint.Local { ... do the work. ... ; continue }` style.  That way `if else is just harder`  to read inside a for loop.

harshavardhana: This also helps moving the check `if endpoint.IsLocal && mynode == -1 { myNode = index }` together to the same `if` condition.

harshavardhana (changes requested):
Cursory review,, will do full review when this moves from WIP --> no-WIPh

harshavardhana:
// cc Also want @fwessels to take a look at the implementation here.. This will conflict with the lsync timeout PR submitted for lock with timeouts.

cmd/namespace-lock.go
```diff
@@ -42,15 +45,21 @@ func initDsyncNodes() error {
...
+			clnts[index] = newLockRPCClient(authConfig{
+				accessKey:       cred.AccessKey,
+				secretKey:       cred.SecretKey,
+				serverAddr:      endpoint.Host,
+				secureConn:      globalIsSSL,
+				serviceEndpoint: pathutil.Join(minioReservedBucketPath, lockServicePath, endpoint.Path),
+				serviceName:     lockServiceName,
+			})
+		} else {
+			clnts[index] = &(globalLockServers[localLockServerIndex].ll)
```
vadmeste: I really didn't understand the purpose of `localLockServerIndex`, don't we have only one local ?

donatello: This is changed in latest version.

donatello:
I have tested the change locally now by creating a 4 node cluster - m{1,2,3,4}.localhost - each of these domains is set in `/etc/hosts` to resolve to 127.0.0.1, and self-signed certs were generated for them.

For server m1, only certs for m2, m3 and m4 were copied into m1's `certs/CAs` directory, and so on for each of the other servers.

Finally, make bucket, put object and get object were tested with mc and found to be working.

cmd/namespace-lock.go
donatello: This is done in latest version of the PR.

fwessels:
Looked in depth at the code, looks good, should not result in any issues with locking and lock maintenance AFAIC.
Have not actually run the code.

donatello:
@vadmeste @harshavardhana - this PR is ready for review.

vadmeste (approved):
Tested with distributed/standalone erasure

harshavardhana (approved):

fwessels (approved):
As discussed this morning LGTM (and as tested by Harsha and Anis)

PR #4466, opened 2017-06-02
config: Avoid stale credentials in memory.

harshavardhana:
## Description
Avoid saving stale credentials in memory.

## Motivation and Context
Stale credentials can cause cluster failure in case of issues in config save.

## How Has This Been Tested?
Manually, by triggering of saving config.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/admin-handlers.go
+++ b/cmd/admin-handlers.go
@@ -181,8 +181,14 @@ func (adminAPI adminAPIHandlers) ServiceCredentialsHandler(w http.ResponseWriter
 	}
 
 	// Update local credentials in memory.
-	serverConfig.SetCredential(creds)
+	prevCred := serverConfig.SetCredential(creds)
+
+	// Save credentials to config file
 	if err = serverConfig.Save(); err != nil {
+		// Save the current creds when failed to update.
+		serverConfig.SetCredential(prevCred)
+
+		errorIf(err, "Unable to update the config with new credentials.")
 		writeErrorResponse(w, ErrInternalError, r.URL)
 		return
 	}
--- a/cmd/config-v19.go
+++ b/cmd/config-v19.go
@@ -61,11 +61,12 @@ func (s *serverConfigV19) GetVersion() string {
 	return s.Version
 }
 
-// SetRegion set new region.
+// SetRegion set a new region.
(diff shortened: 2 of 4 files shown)


harshavardhana:
can you guys review this? @donatello @balamurugana ?

cmd/admin-handlers.go
```diff
@@ -181,8 +181,14 @@ func (adminAPI adminAPIHandlers) ServiceCredentialsHandler(w http.ResponseWriter
 	}
 
 	// Update local credentials in memory.
-	serverConfig.SetCredential(creds)
+	currentCred := serverConfig.SetCredential(creds)
```
donatello: I think this should be called `prevCred` - the operation will succeed if the lock is acquired. This also makes the error handling for `.Save()` more understandable.

cmd/config-v19.go
```diff
@@ -61,12 +61,19 @@ func (s *serverConfigV19) GetVersion() string {
 	return s.Version
 }
 
-// SetRegion set new region.
-func (s *serverConfigV19) SetRegion(region string) {
+// SetRegion set new region. SetRegion retruns previous region.
```
donatello: Typo `returns`

cmd/config-v19.go
```diff
@@ -61,12 +61,19 @@ func (s *serverConfigV19) GetVersion() string {
...
+func (s *serverConfigV19) SetRegion(region string) (prevRegion string) {
```
donatello: Since `prevRegion` returned is not used at any of the call-sites, it appears unnecessary to add this change in.

donatello (changes requested):

cmd/config-v19.go
harshavardhana: Sure

cmd/config-v19.go
harshavardhana: Done

cmd/admin-handlers.go
harshavardhana: Done

donatello (approved):

balamurugana (approved):

PR #4472, opened 2017-06-03
disable settings change on browser in gateway mode

poornas:
## Description

Browser should not allow accesskey /secret key to be changed in gateway mode for azure and s3
## Motivation and Context

Since settings for gateway mode are controlled through environment vars, browser should not allow changing these credentials
## Types of changes

## Checklist:

### DIFF
--- a/cmd/gateway-main.go
+++ b/cmd/gateway-main.go
@@ -233,6 +233,11 @@ func gatewayMain(ctx *cli.Context) {
 
 	router := mux.NewRouter().SkipClean(true)
 
+	if gatewayBackend(backendType) == azureBackend ||
+		gatewayBackend(backendType) == s3Backend {
+		globalIsEnvCreds = true
+	}
+
 	// Register web router when its enabled.
 	if globalIsBrowserEnabled {
 		aerr := registerWebRouter(router)
--- a/cmd/gateway-main.go
+++ b/cmd/gateway-main.go
@@ -233,10 +233,8 @@ func gatewayMain(ctx *cli.Context) {
 
 	router := mux.NewRouter().SkipClean(true)
 
-	if gatewayBackend(backendType) == azureBackend ||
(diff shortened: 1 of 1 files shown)


harshavardhana (changes requested):
Actually we need to fix this.. globalIsEnvCreds should be set only when the ENVs are set. It shouldn't be based on the type of backend IMHO. 

This would work for GCS since GCS has custom access/secret keys.

cmd/gateway-main.go
```diff
@@ -95,6 +95,8 @@ func mustGetGatewayCredsFromEnv() (accessKey, secretKey string) {
 	if accessKey == "" || secretKey == "" {
 		fatalIf(errors.New("Missing credentials"), "Access and secret keys are mandatory to run Minio gateway server.")
 	}
+	globalIsEnvCreds = true
```
harshavardhana: Add a comment.. here @poornas - perhaps in the unit test you can check for this value to be set appropriately?

harshavardhana (changes requested):

krishnasrinivas (approved):

cmd/gateway-main.go
```diff
@@ -154,7 +163,7 @@ func newGatewayConfig(accessKey, secretKey, region string) error {
 	serverConfig = srvCfg
 	serverConfigMu.Unlock()
 
-	return nil
+	return accessKey, secretKey, nil
```
harshavardhana: We don't need to return `accessKey, secretKey`  here in `printGatewayStartupMessage` should use similar to how `printServerStartupMessage`  calling `serverConfig.GetCredential()` 

I see that this type restructuring can wait in another PR @poornas - you can keep the code as is and send the one line change that you did in `mustGetGatewayCredsFromEnv()`

harshavardhana (changes requested):

harshavardhana (approved):

PR #4482, opened 2017-06-06
fs: Migration should handle bucketConfigs as regular objects.

### LINKED ISSUE #4478: Unable to load all bucket policies
Trying to set a policy after creating an object called policy.json causes minio server to not want to restart.
(and after upgrading format.json to V2 a different error message is shown.)

## Possible Solution
See further analysis below.

```
$mkdir fstest2
$ ./minio-05-05 server fstest2
...................................
Drive Capacity: 46 GiB Free, 373 GiB Total
^Z
[1]+  Stopped                 ./minio-05-05 server fstest2
$ bg
[1]+ ./minio-05-05 server fstest2 &
$ mc mb myminio/test
Bucket created successfully `myminio/test`.
$ mc cp /etc/passwd myminio/test/policy.json
/etc/passwd:                         5.13 KB / 5.13 KB  ▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓  100.00% 469.50 KB/s 0s
$ mc policy public myminio/test
mc: <ERROR> Unable to set policy `public` for `myminio/test`. We encountered an internal error, please try again.
$ fg
./minio-05-05 server fstest2
^C
$ ./minio-05-05 server fstest2/
FATA[0000] Initializing object layer failed              cause=Unable to load all bucket policies. Object exists on : .minio.sys as directory buckets/test/policy.json source=[server-main.go:276:serverMain()]
```

If we are to upgrade to the newer version the migration itself is successful but then a corrupted backend (understandably) is also reported.

```
$ ./minio-06-02 server fstest2/
Please do not stop the server.
Migrating bucket metadata format from "1" to newer format "2"... completed successfully.FATA[0000] Initializing object layer failed              cause=corrupted backend format, please join https://slack.minio.io for assistance source=[server-main.go:276:serverMain()]
$ ./minio-06-02 server fstest2/
FATA[0000] Initializing object layer failed              cause=corrupted backend format, please join https://slack.minio.io for assistance source=[server-main.go:276:serverMain()]
$
 ```

## Analysis

Further analysis shows that (in `checkFormatSanityFSV2`) the following check fails: 
```
		// Look for the difference between entries and the
		// expected config set, resulting entries if they
		// intersect with original entries set we know
		// that the backend has unexpected files.
		if !entriesSet.Difference(expectedConfigsSet).IsEmpty() {
			fmt.Println(entriesSet)
			fmt.Println(expectedConfigsSet)
			return errCorruptedFormat
		}
```
which gives this output
```
[objects/ policy.json/]
[listener.json notification.xml objects/ policy.json]
```

### PR DESCRIPTION
**Pull request by harshavardhana**
## Description
Current code failed to anticipate the existence of files
which could have been created to corrupt the namespace such
as `policy.json` file created at the bucket top level.

In the current release creating such as file conflicts
with the namespace for future bucket policy operations.
We implemented migration of backend format to avoid situations
such as these.

This PR handles this situation, makes sure that the
erroneous files should have been moved properly.

## Motivation and Context
Fixes #4478

## How Has This Been Tested?
Manually and also with `go test`

## Types of changes

## Checklist:

### DIFF
--- a/cmd/fs-v1.go
+++ b/cmd/fs-v1.go
@@ -113,10 +113,41 @@ func fsReaddirMetaBuckets(fsPath string) ([]string, error) {
 	return f.Readdirnames(-1)
 }
 
+// List of all bucket metadata configs.
 var bucketMetadataConfigs = []string{
 	bucketNotificationConfig,
 	bucketListenerConfig,
 	bucketPolicyConfig,
+}
+
+// Migrates bucket metadata configs, ignores all other files.
+func migrateBucketMetadataConfigs(metaBucket, bucket, tmpBucket string) error {
+	for _, bucketMetaFile := range bucketMetadataConfigs {
+		fi, err := fsStat(pathJoin(metaBucket, tmpBucket, bucketMetaFile))
+		if err != nil {
+			// There are no such files or directories found,
+			// proceed to next bucket metadata config.
+			if os.IsNotExist(errorCause(err)) {
+				continue
+			}
+			return err
+		}
+
+		// Bucket metadata is a file, move it as an actual bucket config.
+		if fi.Mode().IsRegular() {
+			if err = fsRenameFile(pathJoin(metaBucket, tmpBucket, bucketMetaFile),
+				pathJoin(metaBucket, bucket, bucketMetaFile)); err != nil {
+				if errorCause(err) != errFileNotFound {
+					return err
+				}
+			}
+		}
+
+		// All other file types are ignored.
+	}
+
+	// Success.
+	return nil
 }
 
 // Attempts to migrate old object metadata files to newer format
@@ -152,14 +183,9 @@ func migrateFSFormatV1ToV2(fsPath, fsUUID string) (err error) {
 			return err
 		}
 
-		///  Rename all bucket metadata files to newly created `bucket`.
-		for _, bucketMetaFile := range bucketMetadataConfigs {
-			if err = fsRenameFile(pathJoin(metaBucket, tmpBucket, bucketMetaFile),
-				pathJoin(metaBucket, bucket, bucketMetaFile)); err != nil {
-				if errorCause(err) != errFileNotFound {
-					return err
-				}
-			}
+		// Migrate all the bucket metadata configs.
+		if err = migrateBucketMetadataConfigs(metaBucket, bucket, tmpBucket); err != nil {
+			return err
 		}
 
 		// Finally rename the temporary bucket to `bucket/objects` directory.
@@ -169,6 +195,7 @@ func migrateFSFormatV1ToV2(fsPath, fsUUID string) (err error) {
 				return err
 			}
 		}
+
 	}
 
 	log.Printf("Migrating bucket metadata format from \"%s\" to newer format \"%s\"... completed successfully.", fsFormatV1, fsFormatV2)
--- a/cmd/fs-v1-helpers.go
+++ b/cmd/fs-v1-helpers.go
@@ -123,24 +123,31 @@ func fsMkdir(dirPath string) (err error) {
 	return nil
 }
 
+func fsStat(statLoc string) (os.FileInfo, error) {
+	if statLoc == "" {
+		return nil, traceError(errInvalidArgument)
+	}
+	if err := checkPathLength(statLoc); err != nil {
+		return nil, traceError(err)
+	}
+	fi, err := osStat(preparePath(statLoc))
+	if err != nil {
+		return nil, traceError(err)
+	}
+	return fi, nil
+}
+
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: cmd/fs-v1.go [hunk 1]
```diff
@@ -113,12 +113,48 @@ func fsReaddirMetaBuckets(fsPath string) ([]string, error) {
...
+				}
+			}
+			continue
+		} // else { bucket metadata is a regular file.
+		if err = fsRenameFile(pathJoin(metaBucket, tmpBucket, bucketMetaFile),
+			pathJoin(metaBucket, bucket, objectMetaPrefix, bucketMetaFile)); err != nil {
+			if errorCause(err) != errFileNotFound {
+				return err
+			}
+		}
```
krishnasrinivas: the above if conditions body should be reversed, now this is what happens after migration:
```
krishna@escape:~$ ls export/.minio.sys/buckets/test/objects/
policy.json
krishna@escape:~$ 
```

fwessels (approved):
LGTM and verified that #4478 is fixed.

Also tested for the existence of a file called `notification.xml` as well as a dir/file combination of `policy.json/policy.json`.

cmd/fs-v1.go
```diff
@@ -113,12 +113,48 @@ func fsReaddirMetaBuckets(fsPath string) ([]string, error) {
...
+// List of all bucket metadata configs.
 var bucketMetadataConfigs = []string{
 	bucketNotificationConfig,
 	bucketListenerConfig,
 	bucketPolicyConfig,
 }
 
+func migrateBucketMetadataConfigs(metaBucket, bucket, tmpBucket string) error {
+	for _, bucketMetaFile := range bucketMetadataConfigs {
+		_, err := fsStatDir(pathJoin(metaBucket, tmpBucket, bucketMetaFile))
```
krishnasrinivas: The code will become simpler/readable if you use os.Stat() and use the conditions fi.IsDir() and golang errors

It is better not to use fs-v1-helpers.go abstraction in the migration so that we have finer control

harshavardhana: No we shouldn't os.Stat() is not the right thing to do here. We are making sure that this code works across windows as well. `fsStat()` is nothing but `os.Stat()` but with minio only layered requirements. 

I don't see how os.Stat() would provide finer control.

cmd/fs-v1.go
harshavardhana: Done fixed.

cmd/fs-v1.go
```diff
@@ -113,12 +113,55 @@ func fsReaddirMetaBuckets(fsPath string) ([]string, error) {
...
+				if errorCause(err) != errFileNotFound {
+					return err
+				}
+			}
+		}
+	}
+
+	// Success.
+	return nil
+}
```
krishnasrinivas, line 151: You can compare the readability if os.Stat() is used:

```
func migrateBucketMetadataConfigs(metaBucket, bucket, tmpBucket string) error {
        for _, bucketMetaFile := range bucketMetadataConfigs {
                fi, err := os.Stat(pathJoin(metaBucket, tmpBucket, bucketMetaFile))
                if os.IsNotExist(err) {
                        continue
                }
                if err != nil {
                        return err
                }
                if fi.IsDir() {
                        err = fsRenameFile(pathJoin(metaBucket, tmpBucket, bucketMetaFile),
                                pathJoin(metaBucket, bucket, objectMetaPrefix, bucketMetaFile))
                        if err != nil && errorCause(err) != errFileNotFound {
                                return err
                        }
                        continue
                }
                err = fsRenameFile(pathJoin(metaBucket, tmpBucket, bucketMetaFile),
                        pathJoin(metaBucket, bucket, bucketMetaFile))
                if err != nil && errorCause(err) != errFileNotFound {
                        return err
                }
        }

        // Success.
        return nil
}
```

harshavardhana: The problem with this that it is not catering for the issues we address with fs-v1-helpers.go readability doesn't change when we use a mix of os.Stat() and fsRenameFile() - it creates more confusion.  Going to add a lower level variant of `fsStat()` to be used by both fsStatFile and fsStatDir.

krishnasrinivas (approved):

PR #4483, opened 2017-06-06
Fix browser download returning zero bytes for s3

### LINKED ISSUE #4479: [gateway-s3] Browser download gives 0-byte file
When accessing the browser interface for Minio server in gateway mode for S3 backend, clicking an object within any bucket downloads a 0-byte object.

## Expected Behavior

Download should return appropriate data.

## Current Behavior

Downloads a 0-byte file.

1. Start Minio server in gateway mode for s3
2. Open Minio browser interface in your web-browser and login.
3. Click any object in any bucket in main content panel - this is to download the file.
4. A 0-byte file is returned.

## Context

Found during release candidate testing.

poornas:
## Description

Clicking on download object from browser returns 0 bytes for s3
## Motivation and Context

The browser's web handler sends -1 as length for full downloads. The s3 api expects the actual length of the object in the GetObject call. This fixes #4479 
## How Has This Been Tested?

Manually on s3, azure and server mode.
## Types of changes

## Checklist:

### DIFF
--- a/cmd/gateway-s3.go
+++ b/cmd/gateway-s3.go
@@ -261,8 +261,15 @@ func fromMinioClientListBucketResult(bucket string, result minio.ListBucketResul
 // length indicates the total length of the object.
 func (l *s3Objects) GetObject(bucket string, key string, startOffset int64, length int64, writer io.Writer) error {
 	r := minio.NewGetReqHeaders()
-	if err := r.SetRange(startOffset, startOffset+length-1); err != nil {
-		return s3ToObjectError(traceError(err), bucket, key)
+
+	if length < 0 && length != -1 {
+		return s3ToObjectError(traceError(errInvalidArgument), bucket, key)
+	}
+
+	if startOffset >= 0 && length >= 0 {
+		if err := r.SetRange(startOffset, startOffset+length-1); err != nil {
+			return s3ToObjectError(traceError(err), bucket, key)
+		}
 	}
 	object, _, err := l.Client.GetObject(bucket, key, r)
 	if err != nil {
@@ -271,10 +278,9 @@ func (l *s3Objects) GetObject(bucket string, key string, startOffset int64, leng
 
 	defer object.Close()
 
-	if _, err := io.CopyN(writer, object, length); err != nil {
+	if _, err := io.Copy(writer, object); err != nil {
 		return s3ToObjectError(traceError(err), bucket, key)
 	}
-
 	return nil
 }
 


harshavardhana (approved):

cmd/web-handlers.go
```diff
@@ -546,10 +546,15 @@ func (web *webAPIHandlers) Download(w http.ResponseWriter, r *http.Request) {
 	objectLock.RLock()
 	defer objectLock.RUnlock()
 
-	if err := objectAPI.GetObject(bucket, object, 0, -1, w); err != nil {
+	info, err := objectAPI.GetObjectInfo(bucket, object)
```
harshavardhana: One problem here is that this would increase an API call to the backend server. We should avoid that handle this appropriately at the client SDK used by the s3 backend.

harshavardhana (changes requested):

cmd/web-handlers.go
donatello: A simple GET Object via the S3 API is served by making object layer calls to [GetObjectInfo followed by GetObject](https://github.com/minio/minio/blob/master/cmd/object-handlers.go#L101) - it is not a single call. I think it is acceptable to do the same here. These calls are happening within the backend server and not being sent from browser to backend server.

donatello (approved):

cmd/web-handlers.go
donatello: After discussion with @harshavardhana - it will be better to modify in https://github.com/minio/minio/blob/master/cmd/gateway-s3.go#L262 - it avoids a network call to the backend gateway. We can make the change specifically for s3 gateway

donatello (changes requested):

cmd/gateway-s3.go
```diff
@@ -261,9 +261,16 @@ func fromMinioClientListBucketResult(bucket string, result minio.ListBucketResul
 // length indicates the total length of the object.
 func (l *s3Objects) GetObject(bucket string, key string, startOffset int64, length int64, writer io.Writer) error {
 	r := minio.NewGetReqHeaders()
-	if err := r.SetRange(startOffset, startOffset+length-1); err != nil {
-		return s3ToObjectError(traceError(err), bucket, key)
+	length = -2
```
harshavardhana: what is this for? a typo?

cmd/gateway-s3.go
```diff
@@ -261,9 +261,16 @@ func fromMinioClientListBucketResult(bucket string, result minio.ListBucketResul
...
+	if length < 0 && length != -1 {
+		return s3ToObjectError(traceError(errInvalidArgument), bucket, key)
+	}
+	if length != -1 {
```
harshavardhana: this should be `if length >= 0` - more readable.

cmd/web-handlers.go
harshavardhana: You forgot to remove GetObjectInfo() usage @poornas , revert the changes for web-handlers.go

harshavardhana (changes requested):

cmd/gateway-s3.go
```diff
@@ -261,20 +261,32 @@ func fromMinioClientListBucketResult(bucket string, result minio.ListBucketResul
 // length indicates the total length of the object.
 func (l *s3Objects) GetObject(bucket string, key string, startOffset int64, length int64, writer io.Writer) error {
 	r := minio.NewGetReqHeaders()
-	if err := r.SetRange(startOffset, startOffset+length-1); err != nil {
-		return s3ToObjectError(traceError(err), bucket, key)
+
+	if length < 0 && length != -1 {
+		return s3ToObjectError(traceError(errInvalidArgument), bucket, key)
 	}
+	if length != -1 {
```
harshavardhana: This should be more readable if you add `if startOffset >= 0 && length >= 0`

cmd/gateway-s3.go
```diff
@@ -261,20 +261,32 @@ func fromMinioClientListBucketResult(bucket string, result minio.ListBucketResul
...
 	}
 
 	defer object.Close()
 
-	if _, err := io.CopyN(writer, object, length); err != nil {
-		return s3ToObjectError(traceError(err), bucket, key)
+	if length == -1 {
+		_, err = io.Copy(writer, object)
+	} else {
+		_, err = io.CopyN(writer, object, length)
```
harshavardhana: Actually we don't need `io.CopyN()` we already requested the right amount of file size from the backend we can just do `io.Copy()` instead - this will simplify the code further.

harshavardhana (changes requested):

donatello:
I will test this after all changes are finalized - otherwise, github will keep invalidating my review as changes are made to the PR.

poornas:
incorporated review comments

cmd/gateway-s3.go
```diff
@@ -261,8 +261,15 @@ func fromMinioClientListBucketResult(bucket string, result minio.ListBucketResul
 // length indicates the total length of the object.
 func (l *s3Objects) GetObject(bucket string, key string, startOffset int64, length int64, writer io.Writer) error {
 	r := minio.NewGetReqHeaders()
-	if err := r.SetRange(startOffset, startOffset+length-1); err != nil {
-		return s3ToObjectError(traceError(err), bucket, key)
+
+	if length < 0 && length != -1 {
```
krishnasrinivas, line 265: `if length < -1` will do the same thing

donatello: No if `length < -1` we return error. The only negative length to interpret is -1 as a special case meaning the whole object.

donatello (approved):
Looks good; tested and verified the fix.

harshavardhana (approved):

PR #4490, opened 2017-06-06
Prevent unnecessary (superfluous) initialization of return variable

fwessels:
While unmarshalling the SQS ARN, the return variable is unnecessarily reinstantiated (has already been done as part of defining the named return value).

## Motivation and Context
Found during testing/code review.

## How Has This Been Tested?
Manually.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/bucket-notification-utils.go
+++ b/cmd/bucket-notification-utils.go
@@ -259,15 +259,14 @@ func validateNotificationConfig(nConfig notificationConfig) APIErrorCode {
 // - kafka
 // - webhook
 func unmarshalSqsARN(queueARN string) (mSqs arnSQS) {
-	mSqs = arnSQS{}
 	strs := strings.SplitN(queueARN, ":", -1)
 	if len(strs) != 6 {
-		return mSqs
+		return
 	}
 	if serverConfig.GetRegion() != "" {
 		region := strs[3]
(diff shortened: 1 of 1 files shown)


donatello (approved):

harshavardhana (approved):

cmd/bucket-notification-utils.go
```diff
@@ -259,15 +259,14 @@ func validateNotificationConfig(nConfig notificationConfig) APIErrorCode {
 // - kafka
 // - webhook
 func unmarshalSqsARN(queueARN string) (mSqs arnSQS) {
-	mSqs = arnSQS{}
 	strs := strings.SplitN(queueARN, ":", -1)
 	if len(strs) != 6 {
-		return mSqs
+		return
```
balamurugana, line 264: Even though it is supported, we do not use this style ie we use `return mSqs` style everywhere in minio code base.  Is it changed now?

balamurugana (changes requested):

cmd/bucket-notification-utils.go
fwessels: Am investigating this in a little wider perspective (impact on actual code that is generated), and in the light of that it may make sense to change how we use named return variables and/or return statements from functions.

Stay tuned for more info.

PR #4491, opened 2017-06-06
Detect future versions of on-disk formats and error out accordingly

### LINKED ISSUE #4486: Enhance storage/metadata version handling: prevent older minio versions from corrupting a newer minio's on-disk format
This is to improve error messages when an older version of Minio server is run with a disk which has data in a newer format created by a newer version of Minio.

Currently, when running the release candidate binary in a data-disk on which:

1. `.minio.sys/format.json`-> `fs.version` is bumped up from 2 (current supported) to 3 (unknown future version), then we get start-up error:
```
FATA[0000] Initializing object layer failed              cause=Unable to validate 'format.json', corrupted backend format, please join https://slack.minio.io for assistance source=[server-main.go:276:serverMain()] stack=fs-v1-metadata.go:414:initFormatFS fs-v1.go:229:newFSObjectLayer server-main.go:299:newObjectLayer server-main.go:275:serverMain /home/builder/mygo/src/github.com/minio/minio/vendor/github.com/minio/cli/app.go:499:github.com/minio/minio/vendor/github.com/minio/cli.HandleAction /home/builder/mygo/src/github.com/minio/minio/vendor/github.com/minio/cli/command.go:214:github.com/minio/minio/vendor/github.com/minio/cli.Command.Run /home/builder/mygo/src/github.com/minio/minio/vendor/github.com/minio/cli/app.go:260:github.com/minio/minio/vendor/github.com/minio/cli.(*App).Run main.go:143:Main /home/builder/mygo/src/github.com/minio/minio/main.go:68:main.main /home/builder/.gimme/versions/go1.7.5.linux.amd64/src/runtime/proc.go:183:runtime.main /home/builder/.gimme/versions/go1.7.5.linux.amd64/src/runtime/asm_amd64.s:2086:runtime.goexit
```
2. `.minio.sys/format.json`-> `version` is bumped up from 1 (current) to 2 (future), then no error is returned - this is a **bug**, as it allows an older version minio to corrupt a newer minio's data disk.

## Expected Behavior

1. When seeing an unknown version (i.e. version number greater than what is known to the current version), server needs to display a message like "Future on-disk format detected" and error out, so as to not corrupt the disk.
2. When seeing an older version, it should be migrated to the current/latest version (this is the current behaviour - nothing to implement here).

## Current Behavior

As described in near the beginning here.

## Context

Testing for release candidate of FS policy/notifications related migrations.

donatello:
Additionally, enhance error messages in distributed mode by displaying
the index of the server that had an error in the log.

Fixes #4486.

## How Has This Been Tested?

Manually and unit tests pass.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/format-config-v1.go
+++ b/cmd/format-config-v1.go
@@ -318,7 +318,7 @@ func isSavedUUIDInOrder(uuid string, formatConfigs []*formatConfigV1) bool {
 }
 
 // checkDisksConsistency - checks if all disks are consistent with all JBOD entries on all disks.
-func checkDisksConsistency(formatConfigs []*formatConfigV1) error {
+func checkDisksConsistency(formatConfigs []*formatConfigV1) (int, error) {
 	var disks = make([]string, len(formatConfigs))
 	// Collect currently available disk uuids.
 	for index, formatConfig := range formatConfigs {
@@ -329,20 +329,20 @@ func checkDisksConsistency(formatConfigs []*formatConfigV1) error {
 		disks[index] = formatConfig.XL.Disk
 	}
 	// Validate collected uuids and verify JBOD.
-	for _, uuid := range disks {
+	for idx, uuid := range disks {
 		if uuid == "" {
 			continue
 		}
 		// Is uuid present on all JBOD ?.
 		if !isSavedUUIDInOrder(uuid, formatConfigs) {
-			return fmt.Errorf("%s disk not found in JBOD", uuid)
+			return idx, fmt.Errorf("%s disk not found in JBOD", uuid)
 		}
 	}
-	return nil
+	return -1, nil
 }
 
 // checkJBODConsistency - validate xl jbod order if they are consistent.
-func checkJBODConsistency(formatConfigs []*formatConfigV1) error {
+func checkJBODConsistency(formatConfigs []*formatConfigV1) (int, error) {
 	var sentinelJBOD []string
 	// Extract first valid JBOD.
 	for _, format := range formatConfigs {
@@ -352,16 +352,16 @@ func checkJBODConsistency(formatConfigs []*formatConfigV1) error {
 		sentinelJBOD = format.XL.JBOD
 		break
 	}
-	for _, format := range formatConfigs {
+	for idx, format := range formatConfigs {
 		if format == nil {
 			continue
 		}
 		currentJBOD := format.XL.JBOD
 		if !reflect.DeepEqual(sentinelJBOD, currentJBOD) {
-			return errors.New("Inconsistent JBOD found")
+			return idx, errors.New("Inconsistent JBOD found")
 		}
 	}
-	return nil
+	return -1, nil
 }
 
 // findDiskIndex returns position of disk in JBOD.
@@ -812,13 +812,13 @@ func loadFormatXL(bootstrapDisks []StorageAPI, readQuorum int) (disks []StorageA
 func checkFormatXLValue(formatXL *formatConfigV1) error {
 	// Validate format version and format type.
 	if formatXL.Version != "1" {
-		return fmt.Errorf("Unsupported version of backend format [%s] found", formatXL.Version)
+		return fmt.Errorf("unknown version of backend format `%s` found; please upgrade Minio server; if you are running the latest version, you may have a corrupt file - please join https://slack.minio.io for assistance", formatXL.Version)
 	}
 	if formatXL.Format != "xl" {
-		return fmt.Errorf("Unsupported backend format [%s] found", formatXL.Format)
+		return fmt.Errorf("unknown backend format `%s` found", formatXL.Format)
 	}
 	if formatXL.XL.Version != "1" {
-		return fmt.Errorf("Unsupported XL backend format found [%s]", formatXL.XL.Version)
+		return fmt.Errorf("unknown XL backend format version `%s` found; please upgrade Minio server; if you are running the latest version, you may have a corrupt file - please join https://slack.minio.io for assistance", formatXL.XL.Version)
 	}
 	return nil
 }
@@ -839,15 +839,22 @@ func checkFormatXLValues(formatConfigs []*formatConfigV1) (int, error) {
 	return -1, nil
 }
 
-// checkFormatXL - verifies if format.json format is intact.
+// checkFormatXL - verifies if format.json format is intact. If an
+// error is found, the disk that caused the first error is also
+// returned in the error message.
 func checkFormatXL(formatConfigs []*formatConfigV1) error {
-	if _, err := checkFormatXLValues(formatConfigs); err != nil {
-		return err
+	var err error
+	errCheckCall := func(f func([]*formatConfigV1) (int, error)) {
+		if err == nil {
+			if i, e := f(formatConfigs); e != nil {
+				err = fmt.Errorf("error on disk %d: %v", i+1, e)
+			}
+		}
 	}
-	if err := checkJBODConsistency(formatConfigs); err != nil {
-		return err
-	}
-	return checkDisksConsistency(formatConfigs)
+	errCheckCall(checkFormatXLValues)
+	errCheckCall(checkJBODConsistency)
+	errCheckCall(checkDisksConsistency)
+	return err
 }
 
 // saveFormatXL - populates `format.json` on disks in its order.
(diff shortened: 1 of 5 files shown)


cmd/fs-v1-metadata.go
```diff
@@ -375,6 +383,8 @@ func checkFormatSanityFSV2(fsPath string) error {
 // Check for sanity of FS format for a given version.
 func checkFormatSanityFS(fsPath string, fsFormatVersion string) (err error) {
 	switch fsFormatVersion {
+	case fsFormatV1:
```
krishnasrinivas: checkFormatSanityFS() and checkFormatSanityFSV1() can be removed. Because we would have already migrated from previous versions to the latest version by now.

and in initFormatFS() you can call checkFormatSanityFSV2() directly

donatello: On thinking about this, I think we can make some more usage of a sanity checker - I am going to refactor to run the sanity checker when an old version is detected and also after a migration. This provides some additional error case checking. Pinging @harshavardhana for any related inputs about this.

donatello: So I just updated the patch. Though FSV1's check is a no-op currently, it helps as a reference for future development - this way every version's sanity check can be useful.

harshavardhana: That is why it was kept there.. while it is not going to be used or doing anything in the first place.

cmd/fs-v1-metadata.go
```diff
@@ -313,17 +313,25 @@ func checkFormatFS(format *formatConfigV1, formatVersion string) error {
 		return errUnexpected
 	}
 
+	// Currently there are no older format versions, so just
+	// validate the version; no migration to be done here.
+	if format.Version != "1" {
+		return fmt.Errorf("unknown format file version %s found; please upgrade Minio server; if you are running the latest version, you may have a corrupt file - please join https://slack.minio.io for assistance", format.Version)
```
harshavardhana, line 319: Can you add a test for this?

cmd/fs-v1-metadata.go
```diff
@@ -417,6 +426,12 @@ func initFormatFS(fsPath, fsUUID string) (err error) {
 
 	// Disk is in old format migrate object metadata.
 	if err == errFSFormatOld {
+		// check sanity of older disk format.
+		if serr := checkFormatSanityFS(fsPath, format.FS.Version); serr != nil {
```
harshavardhana: Add a failure test for this?

cmd/xl-v1.go
```diff
@@ -103,7 +103,7 @@ func newXLObjects(storageDisks []StorageAPI) (ObjectLayer, error) {
 	// Load saved XL format.json and validate.
 	newStorageDisks, err := loadFormatXL(storageDisks, readQuorum)
 	if err != nil {
-		return nil, fmt.Errorf("Unable to recognize backend format, %s", err)
+		return nil, fmt.Errorf("Unable to recognize backend format - %s", err)
```
harshavardhana, line 106: Add a failure test case for this..

harshavardhana (changes requested):

cmd/fs-v1-metadata.go
krishnasrinivas: @harshavardhana  That's not how we had decided when we discussed with @abperiasamy 

what we decided was to do:
(Lets say currently it's V5)
```
func initFormatFS() {
....
..
....
checkFormatSanityFSV5()
}
```

Here checkFormatSanityFSV5() will itself be no-op if it does not need to do anything. (But it might need to do sanity checks just like the current checkFormatSanityFSV2()) If it is a no-op then it will be a place holder so that when there is a version increment we know that we have to delete checkFormatSanityFSV5() and implement new checkFormatSanityFSV6()

There is no use of maintaining older sanity checking functions. For place holder we will just have the latest sanity checking function as no-op. (if it actually needs to do any sanity checks then it will not be a no-op)

harshavardhana: Ah yes you are right @krishnasrinivas we don't need it.

donatello: > There is no use of maintaining older sanity checking functions. 

> Ah yes you are right @krishnasrinivas we don't need it.

Could you explain what the technical reason behind it is? I was not part of the discussion alluded to above (or I do not remember), but an actual reason is much stronger than recalling just the conclusion of a discussion from memory.

It looks to me like sanity checking protects us from failing to migrate an older version in case some assumption is not satisfied (and that assumption is checked in the sanity checker function).

If the argument against the above is that the migration function will have to handle all possible cases correctly, then why have a post-migration sanity check at all (since the migration function took care of everything)?

krishnasrinivas: Explained the reason in person to @donatello 
The reason checkFormatSanityFS exists is, in the following sequence of steps:
run minio fs version 1
run minio fs version 2 (does migration)
run minio fs version 1 - this starts instead of erroring out. Hence writing in older format in .minio.sys which never gets corrected.

The sanity checking function takes care of making sure that such a corruption has not happened. If it has happened user will have to clean it up before starting minio in version 2

cmd/fs-v1-metadata.go
donatello: Done

cmd/xl-v1.go
donatello: Done

donatello:
I've refactored and added a function for the V1 corruption check and added a test for it. Ready for further review @krishnasrinivas @harshavardhana

cmd/format-config-v1.go
```diff
@@ -839,15 +839,22 @@ func checkFormatXLValues(formatConfigs []*formatConfigV1) (int, error) {
...
-	if _, err := checkFormatXLValues(formatConfigs); err != nil {
-		return err
-	}
-	if err := checkJBODConsistency(formatConfigs); err != nil {
-		return err
+	var err error
+	errCheckCall := func(f func([]*formatConfigV1) (int, error)) {
+		if err == nil {
+			if i, e := f(formatConfigs); e != nil {
+				err = fmt.Errorf("error on disk %d: %v", i, e)
```
harshavardhana: While returning here we can write the `index+1` to point at the right disk number instead of the slice index number.

cmd/xl-v1_test.go
```diff
@@ -166,3 +167,109 @@ func TestNewXL(t *testing.T) {
...
+	}
+
+	// corrupt format file version
+	format.Version = "2"
+	err = saveXLToDisk(disk, format)
+	if err != nil {
+		t.Fatal("unexpected error: ", err)
+	}
+	_, err = newXLObjects(formattedDisks)
+	if err == nil || err.Error() != "Unable to recognize backend format - error on disk 15: unknown version of backend format `2` found; please upgrade Minio server; if you are running the latest version, you may have a corrupt file - please join https://slack.minio.io for assistance" {
```
harshavardhana: Just like above here the disk 15 is indeed disk 16 we should keep the index appropriately so that the user is able to see exact disk position in the CLI.

harshavardhana (changes requested):

cmd/format-config-v1.go
donatello: Done!

cmd/xl-v1_test.go
donatello: Tests also fixed.

cmd/fs-v1-metadata.go
```diff
@@ -286,16 +286,16 @@ func newFSFormatV2() (format *formatConfigV1) {
 }
 
 // Checks if input format is version 1 and 2.
-func isFSValidFormat(formatCfg *formatConfigV1) bool {
-	// Supported format versions.
-	var supportedFormatVersions = []string{
+func isKnownFSFormat(formatCfg *formatConfigV1) bool {
+	// Known format versions.
+	var knownFormatVersions = []string{
```
balamurugana, line 291: You could move this as global of the function for managing the array better for future versions.

cmd/fs-v1-metadata.go
```diff
@@ -286,16 +286,16 @@ func newFSFormatV2() (format *formatConfigV1) {
...
 		fsFormatV1,
 		fsFormatV2,
-		// New supported versions here.
+		// Add new versions here.
 	}
 
 	// Check for supported format versions.
-	for _, version := range supportedFormatVersions {
+	for _, version := range knownFormatVersions {
```
balamurugana, line 298: If `knownFormatVersions` is `set.StringSet`, it is one liner like `return knownFormatVersions.Contains(version)`

balamurugana (changes requested):

deekoder:
No longer relevant.

PR #4499, opened 2017-06-08
Run select statement in separate goroutine to fix TestListenAndServeTLS

fwessels:
This is a (likely) fix for #4461 where a test case fails under rare circumstances. In case when both conditions for the select statement would be true after the wg.Wait(), the error clause could be selected.

## Description 
Run select statement in separate goroutine (instead of after the wg.Wait() so as to make sure that the 'earliest' of the two select case that becomes active is selected.

As TestServerListenAndServePlain follows the same paradigm, it is updated as well.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/server-mux_test.go
+++ b/cmd/server-mux_test.go
@@ -309,34 +309,36 @@ func TestServerListenAndServePlain(t *testing.T) {
 	// ListenAndServe in a goroutine, but we don't know when it's ready
 	go func() { errc <- m.ListenAndServe("", "") }()
 
-	wg := &sync.WaitGroup{}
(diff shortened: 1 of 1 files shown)


vadmeste (approved):
LGTM, but I really wasn't able to reproduce the bug. (go test -count=1000 -run=TestListenAndServeTLS several times did not show any error)

harshavardhana (approved):

fwessels:
@vadmeste I never encountered the bug myself as well, but it is clearly happening occasionally so let's hope that this tackles it.

Issue #4503, opened 2017-06-08
gateway throws different error messages for invalid --address argument


krisis:
Gateway with invalid --address argument throws different failure messages in different runtime environments.

### Docker
```
➜  minio git:(master) ✗ docker run --rm -e MINIO_ACCESS_KEY=minio -e MINIO_SECRET_KEY=minio123 -p 9001:9001 minio/minio:edge gateway --address 9001 s3 http://localhost:9000
Waiting for all hosts to resolve....All hosts are resolving proceeding to initialize Minio.
time="2017-06-08T09:36:40Z" level=fatal msg="Unable to compare server and endpoint addresses." cause="lookup 9001: invalid domain name" source="[gateway-main.go:216:gatewayMain()]" 
```

### Host Operating System
```
➜  minio git:(master) ✗ minio gateway --address 9001 s3 http://localhost:9000
FATA[0000] Failed to start minio server                  cause=missing port in address 9001 source=[gateway-main.go:289:gatewayMain.func1()]
```
## Expected Behavior

gateway should fail with the same error message for the same (class of) invalid argument supplied.
## Current Behavior

See issue description for more details.
## Possible Solution

Check for port in --address argument.
 
## Steps to Reproduce (for bugs)

1. See description for more details
2.
3.
4.

## Your Environment

* Version used (`minio version`):
* Environment name and version (e.g. nginx 1.9.1):
* Server type and version:
* Operating System and version (`uname -a`):
* Link to your project:

krisis:
@deekoder I am working on this.

harshavardhana:
Fixed by #4376

PR #4504, opened 2017-06-08
gateway: serverAddr must be local and have a port

krisis:
## Description

This change validates serverAddr command line argument to make the failure message independent of the runtime.

## Motivation and Context

See #4503
## How Has This Been Tested?

Manually

## Types of changes

## Checklist:

### DIFF
--- a/cmd/gateway-main.go
+++ b/cmd/gateway-main.go
@@ -209,6 +209,7 @@ func gatewayMain(ctx *cli.Context) {
 	endpointAddr := ctx.Args().Get(1)
 	// Third argument is the address flag
 	serverAddr := ctx.String("address")
+	fatalIf(CheckLocalServerAddr(serverAddr), "Invalid address ‘%s’ in command line argument.", serverAddr)
 
 	if endpointAddr != "" {
 		// Reject the endpoint if it points to the gateway handler itself.


balamurugana (approved):

balamurugana (approved):
Please fix travis-ci failure.

krisis:
@bala, from initial investigation of the travis-ci it appears that `TestInitListeners` is bound to fail unpredictably. We use `getFreePort` to find a port which is available and assume it would remain unused during the test. This is incorrect.

In summary, the CI failure is not related to this PR.

cmd/gateway-main.go
```diff
@@ -209,6 +209,7 @@ func gatewayMain(ctx *cli.Context) {
 	endpointAddr := ctx.Args().Get(1)
 	// Third argument is the address flag
 	serverAddr := ctx.String("address")
+	fatalIf(CheckLocalServerAddr(serverAddr), "Invalid address ‘%s’ in command line argument.", serverAddr)
```
harshavardhana, line 212: This is fixed in the GCS branch . Can you check @krisis

harshavardhana (changes requested):

cmd/gateway-main.go
krisis: @harshavardhana, are we merging GCS branch onto master before the next release? If so I can close this PR.

harshavardhana: The better question to ask is is this a show stopper fix for this release. GCS is the next after this one..so  it doesn't make sense to duplicate a fix. If needed IMO the original fix sent to GCS should be sent here by the same submitter.

krisis: @harshavardhana, I don't understand why the original submitter alone should fix this. Is there any specific reason for this policy?

Are we only fixing show stopper issues now? This fix is simple and it's easy to understand its impact. I don't see a reason to wait for another release. Further, the point about duplicate fix is weak for a one line fix. Close the PR if you still feel strongly.

harshavardhana:
@vadmeste can you take a look at this change?

harshavardhana:
> @harshavardhana, I don't understand why the original submitter alone should fix this. Is there any specific reason for this policy?
> 

It is to avoid causing history and choosing commits. In such a scenario this commit would become dummy and lost in git history. If the original fix was sent here then its better to rebase and handle.  I leave it upto you. This PR is fine as is my only opinion was about keeping the history to avoid two commits addressing the same problem.

> Are we only fixing show stopper issues now? This fix is simple and it's easy to understand its impact. I don't see a reason to wait for another release. Further, the point about duplicate fix is weak for a one line fix. Close the PR if you still feel strongly.

Generally yes show stopper issues, but since its a UI issue it is a show stopper.

krisis:
Closing this PR in favour of https://github.com/minio/minio/pull/4376

Issue #4505, opened 2017-06-08
TestInitListeners unit test failed on travis-ci


krisis:
Travis CI output
```
--- FAIL: TestInitListeners (0.00s)

	server-mux_test.go:143: Test 2: Unable to initialize listeners listen tcp :38416: bind: address already in use
```
## Expected Behavior

Unit tests should pass
## Current Behavior

`TestInitListeners` unit test failed.

## Possible Solution

Not known yet
## Steps to Reproduce (for bugs)

1. Seen once on Travis CI - https://travis-ci.org/minio/minio/jobs/240720887
2.
3.
4.

## Your Environment

* Version used (`minio version`):
* Environment name and version (e.g. nginx 1.9.1):
* Server type and version:
* Operating System and version (`uname -a`):
* Link to your project:

harshavardhana:
Fixed by #4508

PR #4508, opened 2017-06-08
TestInitListeners: Use port 0 pick available port

krisis:
## Description

With this change TestInitListeners picks port "0" to avoid failing with "Port already in use".

## Motivation and Context

See #4505 

## Types of changes

## Checklist:

### DIFF
--- a/cmd/server-mux_test.go
+++ b/cmd/server-mux_test.go
@@ -109,20 +109,18 @@ func dial(addr string) error {
 
 // Tests initializing listeners.
 func TestInitListeners(t *testing.T) {
-	portTest1 := getFreePort()
-	portTest2 := getFreePort()
 	testCases := []struct {
 		serverAddr string
 		shouldPass bool
 	}{
 		// Test 1 with ip and port.
 		{
-			serverAddr: "127.0.0.1:" + portTest1,
+			serverAddr: net.JoinHostPort("127.0.0.1", "0"),
 			shouldPass: true,
 		},
 		// Test 2 only port.
 		{
-			serverAddr: ":" + portTest2,
+			serverAddr: net.JoinHostPort("", "0"),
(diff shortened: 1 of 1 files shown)


cmd/server-mux_test.go
```diff
@@ -109,20 +109,18 @@ func dial(addr string) error {
 
 // Tests initializing listeners.
 func TestInitListeners(t *testing.T) {
-	portTest1 := getFreePort()
-	portTest2 := getFreePort()
```
harshavardhana, line 113: GetFreePort uses port '0' to get the free port. GetFreePort should work appropriately in the first place did you investigate why it doesn't work ?

harshavardhana (changes requested):

cmd/server-mux_test.go
krisis: @harshavardhana I am aware of how getFreePort works. It is obvious from the travis-ci failure that the port returned by getFreePort wasn't really free. getFreePort closes the listener (and the underlying socket) leaving the port free for other applications to bind to. So, getFreePort doesn't guarantee that the port it finds free to be available indefinitely. This is why the unit test failed. Whether port will remain free during the test is non-deterministic and can't be relied on. Does this satisfactorily explain the problem?

harshavardhana: Yes now it makes sense.. just wanted to see why its racy.

harshavardhana (approved):

cmd/server-mux_test.go
donatello: Yes, `getFreePort` is inherently racy and should not be implemented like this. The right way is to write a function that creates a socket bound to a free port and return that. In the meanwhile lets get this change in.

donatello (approved):
LGTM

PR #4512, opened 2017-06-09
Show message about blocking lock call before migration.

### LINKED ISSUE #4511: In shared-fs mode upgrade may hang during migration
## Expected Behavior
In share-fs mode upgrade may hang during migration, we need to provide a user actionable message.

## Current Behavior
In shared-fs mode upgrade may hang during migration - if other instances are running older version, we don't show user any actionable message.

## Possible Solution
Add a log message to indicate user that he needs to restart the other servers. If the `format.json` write lock is blocked.

## Context
UI enhancements.. 

## Your Environment
Current master.

krishnasrinivas:
fixes #4511

Show this message in case we need to migrate:
```
Acquiring lock for migration. (If this takes a long time then another minio process might be holding the lock). - Acquired lock!
```

## Motivation and Context

In distributed-FS setup, if minio1 and minio2 is running on fs-version-2, now if minio1 is upgraded and runs fs-version-3 it tries to hold write lock before migration and blocks till minio2 releases the readlock that it has held. In such a situation we display the above message.

@harshavardhana I would suggest that this patch be merged after the release, because right now this patch is not useful because in older versions we never held read lock on format.json

## How Has This Been Tested?

manual and go test.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/fs-v1-metadata.go
+++ b/cmd/fs-v1-metadata.go
@@ -383,11 +383,38 @@ func checkFormatSanityFS(fsPath string, fsFormatVersion string) (err error) {
 	return err
 }
 
+// Check if backend format needs migration to the latest version.
+func formatFSNeedsMigration(fsPath string) bool {
+	fsFormatPath := pathJoin(fsPath, minioMetaBucket, fsFormatJSONFile)
+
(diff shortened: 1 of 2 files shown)


krisis:
@krishnasrinivas this PR can be closed since we are not implementing format.json migration for this release.

Issue #4517, opened 2017-06-09
Minio hangs when run in shared-fs mode


krisis:
When Minio server is run in shared-fs mode like,
```
minio server --address ":9000" ~/dir
minio server --address ":9001" ~/dir
```
One of the above server instances hangs.

## Expected Behavior

Minio server instances when run in shared-fs mode shouldn't hang.

## Current Behavior

The second (or later) launched Minio server instances hangs indefinitely.

## Possible Solution

Not known yet.
## Steps to Reproduce (for bugs)

1. minio server --address ":9000" ~/dir
2. minio server --address ":9001" ~/dir
3.
4.

## Context

Shared-fs mode testing
## Your Environment

* Version used (`minio version`): commit id `3dfe254a1112384c336d14734f2cb23366d92ab3`
* Environment name and version (e.g. nginx 1.9.1):
* Server type and version:
* Operating System and version (`uname -a`):
* Link to your project:

krisis:
## Root cause analysis
The second (or subsequent) Minio server instances block trying to acquire a write lock (flock(3)) on format.json, since the running instance is holding a read lock for its lifetime. This read lock is to avoid future migrations (from format version 2 to 3 or later) while older Minio version is still operating on the shared filesystem. So, this read lock can't be simply removed.

## Solution 
N B The following solution requires https://github.com/minio/minio/pull/4512 for it to work.

One way to fix this is to hold a non-blocking write lock (i.e, LOCK_EX | LOCK_NB). If the lock isn't granted (i.e, we get EWOULDBLOCK) we should exit with the message - `There may be older version of the server still running. Stop all the older instances and upgrade them before migrating to new format version`. If the lock is granted migration can happen and server starts without blocking indefinitely.

deekoder:
#4523  fixes.

PR #4520, opened 2017-06-09
fs: Do not block during writing `format.json`.

harshavardhana:
## Description
This PR changes the behavior of the write lock attempted
over `format.json` for the purposes to avoid the situation
where the server would indefinitely hang waiting on each
other

## Motivation and Context
A possible fix for #4511 and #4517

## How Has This Been Tested?
Manually.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/fs-v1.go
+++ b/cmd/fs-v1.go
@@ -254,6 +254,11 @@ func newFSObjectLayer(fsPath string) (ObjectLayer, error) {
 
 	// Initialize `format.json`.
 	if err = initFormatFS(fsPath, fsUUID); err != nil {
+		// The follow error would occur if there is a conflicting lock held.
+		// we should simply proceed to attempt a read lock.
+		if errorCause(err) == lock.ErrLocked {
+			return nil, fmt.Errorf("Another server already has an exclusive lock on `format.json`, please wait until you start this server. %s", err)
+		}
 		return nil, err
 	}
 
--- a/pkg/lock/lock_windows.go
+++ b/pkg/lock/lock_windows.go
@@ -36,19 +36,22 @@ var (
 )
 
 const (
+	// https://msdn.microsoft.com/en-us/library/windows/desktop/aa365203(v=vs.85).aspx
+	lockFileExclusiveLock   = 2
+	lockFileFailImmediately = 1
+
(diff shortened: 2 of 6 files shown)


harshavardhana:
This is not a work in progress anymore. @deekoder - reviewers can be assigned.

pkg/lock/lock_windows.go
```diff
@@ -121,18 +139,18 @@ func open(path string, flag int, perm os.FileMode) (*os.File, error) {
...
-	var flag uint32 = 2 // Lockfile exlusive.
+	var flag uint32 = lockFileExclusiveLock // Lockfile exlusive.
 	flag |= flags
 
 	if fd == syscall.InvalidHandle {
 		return nil
 	}
 
-	err := lockFileEx(fd, flag, 1, 0, &syscall.Overlapped{})
+	err := lockFileEx(fd, flags, 1, 0, &syscall.Overlapped{})
```
krisis: shouldn't it be `lockFileEx(fd, flag, ...)`?

cmd/fs-v1.go
```diff
@@ -254,7 +254,11 @@ func newFSObjectLayer(fsPath string) (ObjectLayer, error) {
 
 	// Initialize `format.json`.
 	if err = initFormatFS(fsPath, fsUUID); err != nil {
-		return nil, err
+		// The follow error would occur if there is a conflicting lock held.
+		// we should simply proceed to attempt a read lock.
+		if errorCause(err) != lock.ErrLocked {
+			return nil, err
+		}
```
krisis: If we proceed to acquire read lock, we would still have format.json at version 1 but the new binary would be storing objects uploaded via it in the new directory structure causing requests routed to the older server to fail. This should not be allowed. Instead, m2 should exit printing a message that older version may be running and exit. To make certain that we were conflicted only by older version we need https://github.com/minio/minio/pull/4512, which proceeds to migration only if we still have older version of servers running.

krisis (changes requested):

pkg/lock/lock_windows.go
harshavardhana: Oh yes flag.

cmd/fs-v1.go
harshavardhana: Yes agreed.

harshavardhana:
Not doing this anymore. Closing the PR.

PR #4523, opened 2017-06-09
fs: Add safe locking semantics for `format.json` 

harshavardhana:
## Description
This patch also reverts previous changes which were
merged for migration to the newer disk format. We will
be bringing these changes in subsequent releases. But
we wish to add protection in this release such that
future release migrations are protected.

Revert "fs: Migration should handle bucketConfigs as regular objects. (#4482)"
This reverts commit 976870a.

Revert "fs: Migrate object metadata to objects directory. (#4195)"
This reverts commit 76f4f20.

## Motivation and Context
To address the problems of current release and to cater 
for future migrations. We are reverting back the migration 
changes instead bringing in safety checks in v1 itself. 
Such that we can perform migrations properly in the future.

## How Has This Been Tested?
Manually and tests have been added as well.

## Types of changes

## Checklist:


cmd/fs-v1-metadata.go
```diff
@@ -41,6 +39,16 @@ const (
 	fsFormatJSONFile = "format.json"
 )
 
+// FS format version strings.
+const (
+	fsFormatV1 = "1" // Current format.
+
+	// FS backend format version.
+	fsFormatVersion = fsFormatV1
+
```
krishnasrinivas: we can name the constants like this to make it more readable:
```
        // formatConfigV1.Version
        formatFileV1 = "1"                                                                                                                                                                                       
        formatFileCurrent = formatFileV1                                                                                                                                                                       

        // formatConfigV1.fsFormat.Version                                                                                                                                                                                                                   
        fsFormatBackendV1 = "1"                                                                                                                                                                                    
        fsFormatBackendCurrent = fsFormatBackendV1 
```

krishnasrinivas (commented):
There is one more race that can be prevented by this diff:
```
diff --git a/cmd/fs-v1-metadata.go b/cmd/fs-v1-metadata.go
index 8adab07..159b2c0 100644
--- a/cmd/fs-v1-metadata.go
+++ b/cmd/fs-v1-metadata.go
@@ -327,33 +327,35 @@ func loadFormatFS(formatCfg *formatConfigV1, lk *lock.LockedFile) error {
 }
 
 // Check if disk has already a valid format.
-func checkValidFormatFS(fsFormatPath string) error {
+func checkValidFormatFS(fsFormatPath string) (*RLockedFile, error) {
 	rlk, err := lock.RLockedOpenFile(preparePath(fsFormatPath))
 	if err != nil {
 		if os.IsNotExist(err) {
 			// If format.json not found then
 			// its an unformatted disk.
-			return traceError(errUnformattedDisk)
+			return nil, traceError(errUnformattedDisk)
 		}
-		return traceError(err)
+		return nil, traceError(err)
 	}
-	defer rlk.Close()
 
 	var format = &formatConfigV1{}
 	if err = loadFormatFS(format, rlk.LockedFile); err != nil {
-		return traceError(err)
+		return nil, traceError(err)
 	}
-	return checkFormatFS(format)
+	return rlk, checkFormatFS(format)
 }
 
 func initFormatFS(fsPath, fsUUID string) error {
 	fsFormatPath := pathJoin(fsPath, minioMetaBucket, fsFormatJSONFile)
 
 	for {
-		err := checkValidFormatFS(fsFormatPath)
+		rlk, err := checkValidFormatFS(fsFormatPath)
 		if err != nil && errorCause(err) != errUnformattedDisk {
 			return err
 		}
+		if err == nil {
+			rlk.Close()
+		}
 		if errorCause(err) == errUnformattedDisk {
 			// fsFormatJSONFile  format.json file stored in minioMetaBucket(.minio.sys) directory.
 			lk, err := lock.TryLockedOpenFile(preparePath(fsFormatPath), os.O_RDWR|os.O_CREATE, 0600)
diff --git a/cmd/fs-v1.go b/cmd/fs-v1.go
index ea7008c..ed29c3c 100644
--- a/cmd/fs-v1.go
+++ b/cmd/fs-v1.go
@@ -127,11 +127,13 @@ func newFSObjectLayer(fsPath string) (ObjectLayer, error) {
 		return nil, err
 	}
 
-	// Once initialized hold read lock for the entire operation
-	// of filesystem backend.
-	if _, err = fs.rwPool.Open(pathJoin(fsPath, minioMetaBucket, fsFormatJSONFile)); err != nil {
+	rlk, err := checkValidFormatFS(fsFormatPath)
+	if err != nil {
 		return nil, err
 	}
+	// Once initialized hold read lock for the entire operation
+	// of filesystem backend.
+	_ = rlk
 
 	// Initialize and load bucket policies.
 	err = initBucketPolicies(fs)
```

cmd/fs-v1-metadata.go
harshavardhana: Good idea.

cmd/fs-v1-metadata.go
```diff
@@ -274,167 +271,119 @@ func newFSMetaV1() (fsMeta fsMetaV1) {
...
-	// fsFormatJSONFile - format.json file stored in minioMetaBucket(.minio.sys) directory.
-	lk, err := lock.LockedOpenFile(preparePath(fsFormatPath), os.O_RDWR|os.O_CREATE, 0600)
+// Check if disk has already a valid format.
+func checkValidFormatFS(fsFormatPath string) error {
+	rlk, err := lock.RLockedOpenFile(preparePath(fsFormatPath))
 	if err != nil {
+		if os.IsNotExist(err) {
+			// If format.json not found then
+			// its an unformatted disk.
+			return traceError(errUnformattedDisk)
```
fwessels: Why do we need to wrap with traceError() here ?

harshavardhana: Its a general practice.. that all low level errors have a traceError() associated for example if at all the caller needs to print the stack they can.  We don't need to return errors with or without stack in the same function just to be consistent.

cmd/fs-v1-metadata.go
```diff
@@ -274,167 +271,119 @@ func newFSMetaV1() (fsMeta fsMetaV1) {
...
+func initFormatFS(fsPath, fsUUID string) error {
+	fsFormatPath := pathJoin(fsPath, minioMetaBucket, fsFormatJSONFile)
 
-	// Disk is in old format migrate object metadata.
-	if err == errFSFormatOld {
-		if merr := migrateFSObject(fsPath, fsUUID); merr != nil {
-			return merr
+	for {
+		err := checkValidFormatFS(fsFormatPath)
+		if err != nil && errorCause(err) != errUnformattedDisk {
```
fwessels: Could move `return err` to `else if err != nil` clause in line 380 (saves one  `errorCause(err) == errUnformattedDisk` comparison)

harshavardhana: Good idea..

cmd/fs-v1-metadata.go
```diff
@@ -274,167 +271,119 @@ func newFSMetaV1() (fsMeta fsMetaV1) {
...
-
-	// Rewrite or write format.json depending on if disk
-	// unformatted and if format is old.
-	if err == errUnformattedDisk || err == errFSFormatOld {
-		if _, err = format.WriteTo(lk); err != nil {
-			return traceError(fmt.Errorf("Unable to initialize 'format.json', %s", err))
+			var format = &formatConfigV1{}
+			if err = loadFormatFS(format, lk); err != nil {
+				if errorCause(err) != errUnformattedDisk {
+					return err
```
fwessels: Can combine both `if` statements, i.e. `err != nil && errorCause(err) != errUnformattedDisk`

harshavardhana: :+1:

harshavardhana:
> There is one more race that can be prevented by this diff:

This race has been addressed @krishnasrinivas - Thanks for the insights.

krishnasrinivas:
will check @harshavardhana

cmd/fs-v1-metadata.go
```diff
@@ -274,167 +261,134 @@ func newFSMetaV1() (fsMeta fsMetaV1) {
...
+		return nil, traceError(err)
 	}
-	defer lk.Close()
 
 	var format = &formatConfigV1{}
-	_, err = format.ReadFrom(lk)
-	// For all unexpected errors, we return.
-	if err != nil && errorCause(err) != io.EOF {
-		return traceError(fmt.Errorf("Unable to load 'format.json', %s", err))
+	if err = loadFormatFS(format, rlk.LockedFile); err != nil {
```
krishnasrinivas: rlk.Close() can be called here

cmd/fs-v1-metadata.go
```diff
@@ -274,167 +261,134 @@ func newFSMetaV1() (fsMeta fsMetaV1) {
...
-		format = newFSFormatV2()
-	} else {
-		// Validate loaded `format.json`.
-		err = checkFormatFS(format, fsFormatVersion)
-		if err != nil && err != errFSFormatOld {
-			return traceError(fmt.Errorf("Unable to validate 'format.json', %s", err))
-		}
-	}
+	// Check format FS.
+	err = checkFormatFS(format)
```
krishnasrinivas: if err != nil then rlk.Close() can be called

pkg/lock/lock.go
```diff
@@ -19,10 +19,16 @@
...
 
 import (
+	"errors"
 	"os"
 	"sync"
 )
 
+var (
+	// ErrLocked is returned if the underlying fd is already locked.
+	ErrLocked = errors.New("lock: file already locked")
```
krishnasrinivas: ErrAlreadyLocked will make it more readable

cmd/fs-v1-metadata.go
```diff
@@ -274,167 +261,134 @@ func newFSMetaV1() (fsMeta fsMetaV1) {
 	return fsMeta
 }
 
-// newFSFormatV2 - initializes new formatConfigV1 with FS format version 2.
-func newFSFormatV2() (format *formatConfigV1) {
+// newFSFormatV1 - initializes new formatConfigV1 with FS format info.
+func newFSFormatV1() (format *formatConfigV1) {
```
balamurugana: This function needs to be in `format-config-v1.go`

cmd/fs-v1-metadata.go
```diff
@@ -274,167 +261,134 @@ func newFSMetaV1() (fsMeta fsMetaV1) {
...
 	return &formatConfigV1{
-		Version: "1",
-		Format:  "fs",
+		Version: formatFileCurrent,
```
balamurugana: As we create `formatConfigV1`, `Version` should be `formatFileV1`.  `formatFileCurrent` would point to different version in future.

cmd/fs-v1-metadata.go
```diff
@@ -274,167 +261,134 @@ func newFSMetaV1() (fsMeta fsMetaV1) {
...
+		Format:  formatBackendFS,
 		FS: &fsFormat{
-			Version: fsFormatV2,
+			Version: fsFormatBackendCurrent,
```
balamurugana: As we are in `newFSFormatV1()`, `Version` should be `fsFormatBackendV1`.  `fsFormatBackendCurrent` would point to different version in future.

cmd/fs-v1-metadata.go
```diff
@@ -274,167 +261,134 @@ func newFSMetaV1() (fsMeta fsMetaV1) {
...
-	return false
-}
-
-// errFSFormatOld- old fs format.
-var errFSFormatOld = errors.New("old FS format found")
-
 // Checks if the loaded `format.json` is valid and
 // is expected to be of the requested version.
-func checkFormatFS(format *formatConfigV1, formatVersion string) error {
+func checkFormatFS(format *formatConfigV1) error {
```
balamurugana: 1. I think function needs to be in `format-config-v1.go`
2. How about this function is a method to `formatConfigV1` type?

cmd/fs-v1-metadata.go
```diff
@@ -274,167 +261,134 @@ func newFSMetaV1() (fsMeta fsMetaV1) {
...
-// check for FS format in version "1".
-func checkFormatSanityFSV1(fsPath string) error {
 	return nil
 }
 
-// Check for sanity of FS format in version "2".
-func checkFormatSanityFSV2(fsPath string) error {
-	buckets, err := readDir(pathJoin(fsPath, minioMetaBucket, bucketConfigPrefix))
-	if err != nil && err != errFileNotFound {
+func loadFormatFS(formatCfg *formatConfigV1, lk *lock.LockedFile) error {
```
balamurugana: I think function needs to be in `format-config-v1.go`

cmd/fs-v1-metadata.go
```diff
@@ -274,167 +261,134 @@ func newFSMetaV1() (fsMeta fsMetaV1) {
...
-		return fmt.Errorf("Unable to recognize backend format, Disk is not in FS format. %s", format.Format)
+	if format.FS == nil {
+		return errUnexpected
 	}
 
-	// Check if format is currently supported.
-	if !isFSValidFormat(format) {
-		return errCorruptedFormat
+	// Validate if format config version is v1.
+	if format.Version != formatFileCurrent {
```
balamurugana: As this function accepts `formatConfigV1`, the expected version should be `formatFileV1`

cmd/fs-v1-metadata.go
```diff
@@ -274,167 +261,134 @@ func newFSMetaV1() (fsMeta fsMetaV1) {
...
+	// Validate if we have the same format.
+	if format.Format != fsMetaFormat {
+		return fmt.Errorf("Unable to recognize backend FS format. Expected '%s', got '%s'",
+			fsMetaFormat, format.Format)
 	}
 
-	return nil
-}
+	// Check if format is currently supported.
+	if format.FS.Version != fsFormatBackendCurrent {
```
balamurugana: 1. How will we have this check for `fsFormatBackendV2`?
2. I think `{check|load|save}FormatFS()` functions calling the current (latest) version of format is better over `fsFormatBackendCurrent` const ie
```
func checkFormatFS(format *formatConfigV1) {
    return checkFormatFSV1(format);
}
```

balamurugana (changes requested):

pkg/lock/lock.go
balamurugana: I think prefix `lock:` is not a good idea. the consumer will add any prefix for this error.

cmd/fs-v1-metadata.go
harshavardhana: done

cmd/fs-v1-metadata.go
harshavardhana: done

pkg/lock/lock.go
harshavardhana: done

cmd/fs-v1-metadata.go
harshavardhana: done

pkg/lock/lock_solaris.go
```diff
@@ -24,12 +24,8 @@ import (
...
 )
 
-// LockedOpenFile - initializes a new lock and protects
-// the file from concurrent access across mount points.
-// This implementation doesn't support all the open
-// flags and shouldn't be considered as replacement
-// for os.OpenFile().
-func LockedOpenFile(path string, flag int, perm os.FileMode) (*LockedFile, error) {
+// lockedOpenFile is an internal function.
+func lockedOpenFile(path string, flag int, perm os.FileMode, alock int) (*LockedFile, error) {
```
balamurugana: 1. `alock` name doesn't convey the meaning.
2. By default `var lock syscall.Flock_t` created fields with zero value.  Not sure why you initialize `Start`, `Len`, `Pid` explicitly.

cmd/fs-v1-metadata.go
harshavardhana: done

pkg/lock/lock_windows.go
```diff
@@ -36,19 +36,22 @@ var (
 )
```
balamurugana, line 33: `errLocked` could be named better like `errAlreadyLocked`

pkg/lock/lock_windows.go
```diff
@@ -36,19 +36,22 @@ var (
...
+
 	// see https://msdn.microsoft.com/en-us/library/windows/desktop/ms681382(v=vs.85).aspx
 	errLockViolation syscall.Errno = 0x21
 )
 
-// LockedOpenFile - initializes a new lock and protects
-// the file from concurrent access.
-func LockedOpenFile(path string, flag int, perm os.FileMode) (*LockedFile, error) {
+// lockedOpenFile is an internal function.
+func lockedOpenFile(path string, flag int, perm os.FileMode, alock uint32) (*LockedFile, error) {
```
balamurugana: What is the meaning of `alock`?

balamurugana (changes requested):

cmd/fs-v1-metadata.go
harshavardhana: done.

cmd/fs-v1-metadata.go
harshavardhana: done

cmd/fs-v1-metadata.go
harshavardhana: done

cmd/fs-v1-metadata.go
harshavardhana: done

pkg/lock/lock_solaris.go
harshavardhana: It is to convey that we are taking a proper range long which is entire file.

pkg/lock/lock_windows.go
harshavardhana: done

pkg/lock/lock_windows.go
harshavardhana: advisory lock. @balamurugana

harshavardhana: Changing it to lockType.

pkg/lock/lock_solaris.go
harshavardhana: Let me change it differently..

cmd/fs-v1-metadata.go
harshavardhana: done

harshavardhana:
All the review changes have been addressed @balamurugana @krishnasrinivas - can you guys take a look again?

pkg/lock/lock_solaris.go
```diff
@@ -51,16 +42,24 @@ func LockedOpenFile(path string, flag int, perm os.FileMode) (*LockedFile, error
...
 	}
 
-	lock.Type = lockType
-	lock.Whence = 0
+	var lock = syscall.Flock_t{
+		Start:  0,
+		Len:    0,
+		Pid:    0,
+		Type:   lockType,
+		Whence: 0,
```
balamurugana, line 50: Do we still need to initialize fields to '0'?

cmd/fs-v1.go
```diff
@@ -257,12 +127,17 @@ func newFSObjectLayer(fsPath string) (ObjectLayer, error) {
...
-	if _, err = fs.rwPool.Open(pathJoin(fsPath, minioMetaBucket, fsFormatJSONFile)); err != nil {
+	// Validate the `format.json` again after the initialization
+	// to further validate `format.json` entries to be valid.
+	rlk, err := checkLockedValidFormatFS(fsPath)
+	if err != nil {
 		return nil, err
 	}
 
+	// Once the filesystem has initialized hold the read lock for
+	// the entire operation.
```
balamurugana: We would need to explain why we hold read lock for entire life time.

cmd/format-config-v1.go
```diff
@@ -51,6 +79,88 @@ type formatConfigV1 struct {
...
+	// Represents FS backend.
+	formatBackendFS = "fs"
+
+	// Represents XL backend.
+	formatBackendXL = "xl"
+)
+
+// CheckFSV1 if the format is FS and is valid with right values
+// returns appropriate errors otherwise.
+func (f *formatConfigV1) CheckFSV1() error {
```
balamurugana: `V1` in function name looks odd as `formatConfigV1` already carries `V1`.

cmd/format-config-v1.go
```diff
@@ -51,6 +79,88 @@ type formatConfigV1 struct {
...
+			fsFormatBackendV1, f.FS.Version)
+	}
+
+	// Success.
+	return nil
+}
+
+// LoadFormatV1 - loads format config v1, returns `errUnformattedDisk`
+// if reading format.json fails with io.EOF.
+func (f *formatConfigV1) LoadFormatV1(lk *lock.LockedFile) error {
```
balamurugana: same as above

cmd/format-config-v1.go
```diff
@@ -51,6 +79,88 @@ type formatConfigV1 struct {
...
+		return errUnexpected
+	}
+
+	if f.FS == nil {
+		return errUnexpected
+	}
+
+	// Validate if format config version is v1.
+	if f.Version != formatFileV1 {
+		return fmt.Errorf("Unable to recognize format file version. Expected '%s', got '%s'",
```
balamurugana: I think `unknown format file version %s` is better,

cmd/format-config-v1.go
```diff
@@ -51,6 +79,88 @@ type formatConfigV1 struct {
...
+			formatFileV1, f.Version)
+	}
+
+	// Validate if we have the same f.
+	if f.Format != formatBackendFS {
+		return fmt.Errorf("Unable to recognize backend FS format. Expected '%s', got '%s'",
```
balamurugana: I think `FS backend format required. Found %s` is better,

cmd/format-config-v1.go
```diff
@@ -51,6 +79,88 @@ type formatConfigV1 struct {
...
+			formatBackendFS, f.Format)
+	}
+
+	// Check if format is currently supported.
+	if f.FS.Version != fsFormatBackendV1 {
+		return fmt.Errorf("Unable to recognize backend FS format version. Expected '%s', got '%s'",
```
balamurugana: I think `unknown backend FS format version %s` is better,

cmd/format-config-v1.go
```diff
@@ -916,10 +1037,10 @@ func initFormatXL(storageDisks []StorageAPI) (err error) {
 		}
 		// Allocate format config.
 		formats[index] = &formatConfigV1{
-			Version: "1",
-			Format:  "xl",
+			Version: formatFileCurrent,
+			Format:  formatBackendXL,
 			XL: &xlFormat{
-				Version: "1",
+				Version: xlFormatBackendCurrent,
```
balamurugana: 1. `formatFileCurrent` and `xlFormatBackendCurrent` need to be `formatFileV1` and `xlFormatBackendV1` respectively for `formatConfigV1`.
2. You could have a separate `newXLFormatV1()` and use it here.

balamurugana (changes requested):

pkg/lock/lock_solaris.go
harshavardhana: Yes it is indicate a meaning. That we are reading the right ranges and they outght to be these values.

cmd/format-config-v1.go
harshavardhana: Done

cmd/fs-v1.go
harshavardhana: Done

harshavardhana:
Addressed all comments @balamurugana

cmd/format-config-v1.go
```diff
@@ -51,6 +69,82 @@ type formatConfigV1 struct {
...
+	if f.FS == nil {
+		return errUnexpected
+	}
+
+	// Validate if format config version is v1.
+	if f.Version != formatFileV1 {
+		return fmt.Errorf("Unknown format file version '%s'", f.Version)
+	}
+
+	// Validate if we have the same f.
```
krisis: What does "same f" mean here?

cmd/format-config-v1.go
```diff
@@ -51,6 +69,82 @@ type formatConfigV1 struct {
...
+	// Check if format is currently supported.
+	if f.FS.Version != fsFormatBackendV1 {
+		return fmt.Errorf("Unknown backend FS format version '%s'", f.FS.Version)
+	}
+
+	// Success.
+	return nil
+}
+
+// LoadFormatV1 - loads format config v1, returns `errUnformattedDisk`
```
krisis: The function in comment and code are different. Should it be LoadFormatV1 or LoadFormat?

cmd/format-config-v1_test.go
```diff
@@ -682,6 +686,187 @@ func TestGenericFormatCheckXL(t *testing.T) {
 	}
 }
 
+// TestFSCheckFormatFSErr - test loadFormatFS loading older format.
+func TestFSCheckFormatFSErr(t *testing.T) {
+	// Prepare for testing
```
krisis, line 692: This test can be written in a "table-driven" fashion avoiding repetitive code. E.g,
```go
testCases := []struct{
  formatBackend string
  formatVersion string
} {...}
...
```

cmd/fs-v1-metadata.go
```diff
@@ -274,167 +259,81 @@ func newFSMetaV1() (fsMeta fsMetaV1) {
...
+					continue
+				}
+				err = traceError(lerr)
+				break
+			}
+			defer lk.Close()
+
+			// Load format to validate again.
+			var format = &formatConfigV1{}
+			if lerr = format.LoadFormat(lk); lerr != nil && errorCause(lerr) == errUnformattedDisk {
```
krisis: if we used `err` instead of `lerr`, we can avoid the following assignment `err = lerr`.

cmd/fs-v1.go
```diff
@@ -257,12 +127,21 @@ func newFSObjectLayer(fsPath string) (ObjectLayer, error) {
 		return nil, err
 	}
 
-	// Once initialized hold read lock for the entire operation
-	// of filesystem backend.
-	if _, err = fs.rwPool.Open(pathJoin(fsPath, minioMetaBucket, fsFormatJSONFile)); err != nil {
+	// Validate the `format.json` again after the initialization
+	// to further validate `format.json` entries to be valid.
+	rlk, err := checkLockedValidFormatFS(fsPath)
```
krisis: Why is it required to check format validity again?

krisis (changes requested):

krisis:
`initFormatFS` can be simplified by using write lock to validate existing (if any) format.json and create on if absent. This avoids going back and forth between read and write locks.

EDIT: This approach wouldn't work since waiting for a write lock for validating format.json may hang indefinitely. So, for format validation read lock should be used and write lock is required for creating a new format.json.

krisis:
@krishnasrinivas,  What is the race you refer to [here](https://github.com/minio/minio/pull/4523#pullrequestreview-43286747)?

krisis:
- Tested shared-FS mode configuration with 3 Minio server instances sharing a GlusterFS mount (`glusterfs 3.7.6`). Works as expected.

- Tested the above configuration on a local NFS server. Works as expected.

harshavardhana:
> @krishnasrinivas, What is the race you refer to here?

Added some explanation in the code @krisis after initFormatFS()

cmd/fs-v1-metadata.go
```diff
@@ -274,167 +259,81 @@ func newFSMetaV1() (fsMeta fsMetaV1) {
...
-	if err == errUnformattedDisk || err == errFSFormatOld {
-		if _, err = format.WriteTo(lk); err != nil {
-			return traceError(fmt.Errorf("Unable to initialize 'format.json', %s", err))
+		if err != nil && errorCause(err) == errUnformattedDisk {
+			// Attempt a write lock on formatConfigFile `format.json`
+			// file stored in minioMetaBucket(.minio.sys) directory.
+			lk, lerr := lock.TryLockedOpenFile(preparePath(fsFormatPath), os.O_RDWR|os.O_CREATE, 0600)
+			if lerr != nil {
+				// Existing write locks detected.
+				if lerr == lock.ErrAlreadyLocked {
```
krishnasrinivas: Better sleep for 100ms to avoid CPU usage going to 100%

cmd/format-config-v1_test.go
harshavardhana: done

cmd/fs-v1-metadata.go
harshavardhana: There was a reason, shadowing and complications of err being `nil` somewhere down the line.

harshavardhana:
> EDIT: This approach wouldn't work since waiting for a write lock for validating format.json may hang indefinitely. So, for format validation read lock should be used and write lock is required for creating a new format.json.

:+1:

cmd/fs-v1-metadata.go
harshavardhana: Sure..

krishnasrinivas (approved):
Tested on ext4

cmd/fs-v1-metadata.go
```diff
@@ -274,167 +260,83 @@ func newFSMetaV1() (fsMeta fsMetaV1) {
...
+
+	// This loop validates format.json by holding a read lock and
+	// proceeds to holding a TryWriteLock by validating if the
+	// disk is unformatted. If for some reason TryWriteLock fails
+	// and the error returned is lock.ErrAlreadyLocked i.e some other
+	// process is holding a lock we make sure that we try again
+	// in the loop by holding a read lock.
+	for {
+		var rlk *lock.RLockedFile
+		rlk, err = checkLockedValidFormatFS(fsPath)
```
vadmeste: checkLockedValidFormatFS() can also return lock.ErrAlreadyLocked, shouldn't we also take care of that?

harshavardhana: It won't return @vadmeste lock.ErrAlreadyLocked is returned by Try `checkLockedValidFormatFS()` will only block if there is an active WriteLock.

vadmeste: > It won't return @vadmeste lock.ErrAlreadyLocked is returned by Try checkLockedValidFormatFS() will only block if there is an active WriteLock.

Yes, you are right, syscall.LOCK_NB  is not passed in the code path of checkLockedValidFormatFS call

cmd/fs-v1-metadata.go
```diff
@@ -274,167 +260,83 @@ func newFSMetaV1() (fsMeta fsMetaV1) {
...
-			return merr
+	fsFormatPath := pathJoin(fsPath, minioMetaBucket, formatConfigFile)
+
+	// This loop validates format.json by holding a read lock and
+	// proceeds to holding a TryWriteLock by validating if the
+	// disk is unformatted. If for some reason TryWriteLock fails
+	// and the error returned is lock.ErrAlreadyLocked i.e some other
+	// process is holding a lock we make sure that we try again
+	// in the loop by holding a read lock.
+	for {
```
vadmeste: It is still not clear for me why we are using `for` and retry lock later thing. Why we can't just wait until the file lock will get released by another process if any ?

harshavardhana: Because there is no way to unblock a blocking write call.

cmd/fs-v1-metadata.go
```diff
@@ -274,167 +260,83 @@ func newFSMetaV1() (fsMeta fsMetaV1) {
...
+			}
+			err = lerr
 		}
+		break
 	}
-
-	// Check for sanity.
-	return checkFormatSanityFS(fsPath, format.FS.Version)
+	return
 }
```
donatello: This approach for `initFormatFS()` requires at-least two-locks in the best case. In the best case, the disk is already formatted (and is of the latest version, when V2 is introduced). This implementation takes a read lock inside `initFormatFS()` and then superfluously releases it, and then after `initFormatFS()` acquires and keeps the lock for the lifetime of the process.

We can do better than this.

Here is some code illustrating the idea of a state machine implementation that acquires locks and performs actions (formatting/migration) at well defined conditions and finally when the loop is exited, the disk is formatted (and in future V2 servers, migrations are also performed). The read lock is acquired just once in the best case, and held till the end of the process life-time.

```diff
diff --git a/cmd/fs-v1-metadata.go b/cmd/fs-v1-metadata.go
index 4be01ec..90a4eb0 100644
--- a/cmd/fs-v1-metadata.go
+++ b/cmd/fs-v1-metadata.go
@@ -25,6 +25,7 @@ import (
 	pathutil "path"
 	"sort"
 	"strings"
+	"time"
 
 	"github.com/minio/minio/pkg/lock"
 	"github.com/minio/minio/pkg/mimedb"
@@ -290,50 +291,67 @@ func checkLockedValidFormatFS(fsPath string) (*lock.RLockedFile, error) {
 	return rlk, traceError(err)
 }
 
+var errOldFormat = fmt.Errorf("Version is not latest!")
+
 func initFormatFS(fsPath, fsUUID string) (err error) {
 	fsFormatPath := pathJoin(fsPath, minioMetaBucket, formatConfigFile)
 
-	// This loop validates format.json by holding a read lock and
-	// proceeds to holding a TryWriteLock by validating if the
-	// disk is unformatted. If for some reason TryWriteLock fails
-	// and the error returned is lock.ErrAlreadyLocked i.e some other
-	// process is holding a lock we make sure that we try again
-	// in the loop by holding a read lock.
+	var rlk *lock.RLockedFile
 	for {
-		var rlk *lock.RLockedFile
 		rlk, err = checkLockedValidFormatFS(fsPath)
-		if err == nil {
-			// Format is valid, close the read lock.
-			rlk.Close()
-		}
-		if err != nil && errorCause(err) == errUnformattedDisk {
-			// Attempt a write lock on formatConfigFile `format.json`
-			// file stored in minioMetaBucket(.minio.sys) directory.
+		switch {
+		case err == nil:
+			// Disk is formatted, and has latest
+			// version. Hold the lock for lifetime of the
+			// process and exit initialization.
+			_ = rlk
+			return err
+		case errCause(err) == errUnformattedDisk || errCause(err) == errOldFormat:
+			// format.json does not exist or is of old version
 			lk, lerr := lock.TryLockedOpenFile(preparePath(fsFormatPath), os.O_RDWR|os.O_CREATE, 0600)
-			if lerr != nil {
-				// Existing write locks detected.
-				if lerr == lock.ErrAlreadyLocked {
-					// Continue and attempt write lock again.
-					continue
-				}
-				err = traceError(lerr)
-				break
+			if lerr == lock.ErrAlreadyLocked {
+				// lock is already taken - sleep and retry
+				time.Sleep(100 * time.Millisecond)
+				continue
+			} else if lerr != nil {
+				// unexpected error - quit
+				return traceError(lerr)
 			}
+
+			// lock is acquired!
 			defer lk.Close()
 
-			// Load format to validate again.
+			// Load format - and check if someone
+			// formatted/migrated before we got the lock!
 			var format = &formatConfigV1{}
-			if lerr = format.LoadFormat(lk); lerr != nil && errorCause(lerr) == errUnformattedDisk {
+			if lerr = format.LoadFormat(lk); errorCause(lerr) == errUnformattedDisk {
 				if _, lerr = newFSFormat().WriteTo(lk); lerr != nil {
-					err = traceError(fmt.Errorf("Unable to initialize 'format.json', %s", lerr))
-					break
+					return traceError(fmt.Errorf("Unable to initialize 'format.json', %s", lerr))
 				}
+				// formatting worked fine. Acquire
+				// read-lock and exit init
+				continue
+			} else if lerr != nil {
+				// unexpected error - exit.
+				return traceError(lerr)
+			}
+
+			// validate the loaded format.
+			if lerr = format.CheckFS(); lerr == errOldFormat {
+				// we have write lock - perform migration!
+
+				// currently just a no-op. after
+				// migration, acquire read lock and
+				// exit init.
+			} else if lerr != nil {
+				// unexpected error - exit.
+				return traceError(lerr)
 			}
-			err = lerr
+		default:
+			// unexpected error received.
+			return traceError(err)
 		}
-		break
 	}
-	return
 }
 
 // Return if the part info in uploadedParts and completeParts are same.
diff --git a/cmd/fs-v1.go b/cmd/fs-v1.go
index 94e1dbf..6f9d0a9 100644
--- a/cmd/fs-v1.go
+++ b/cmd/fs-v1.go
@@ -127,24 +127,6 @@ func newFSObjectLayer(fsPath string) (ObjectLayer, error) {
 		return nil, err
 	}
 
-	// Validate the `format.json` again after the initialization
-	// to further validate `format.json` entries to be expected.
-	//
-	// Once the filesystem has initialized hold the read lock for
-	// the life time of the server. This is done to ensure that under
-	// shared backend mode for FS, remote servers do not migrate
-	// or cause changes on backend format after initFormatFS().
-	// There is a slight window for this to happen.
-	rlk, err := checkLockedValidFormatFS(fsPath)
-	if err != nil {
-		return nil, err
-	}
-
-	// Holding a read lock ensures that any write lock operation
-	// is blocked if attempted in-turn avoiding corruption on
-	// the backend disk.
-	_ = rlk // Hold the lock on `format.json` until server dies.
-
 	// Initialize and load bucket policies.
 	err = initBucketPolicies(fs)
 	if err != nil {
```

donatello (changes requested):
The below is the most-significant feedback. But I also have some code-improvement suggestions coming up.

cmd/format-config-v1.go
```diff
@@ -51,6 +69,82 @@ type formatConfigV1 struct {
...
+	formatBackendFS = "fs"
+
+	// Represents XL backend.
+	formatBackendXL = "xl"
+)
+
+// CheckFS if the format is FS and is valid with right values
+// returns appropriate errors otherwise.
+func (f *formatConfigV1) CheckFS() error {
+	if f == nil {
```
donatello: Why check for `nil` here? It appears inconsistent, that all the other methods of `formatConfigV1` trust that `f` is not `nil`. If there is no actual use for this (in the current code, there is no nil pointer dereference on this type), we can remove it in the interest of simplicity.

cmd/format-config-v1.go
```diff
@@ -51,6 +69,82 @@ type formatConfigV1 struct {
...
+
+	// No data on disk `format.json` still empty treat
+	// it as unformatted disk.
+	if errorCause(err) == io.EOF {
+		return traceError(errUnformattedDisk)
+	}
+
+	// Success.
+	return nil
+}
```
donatello, line 132: The above error checking pattern (appears to be used quite often) is not very readable. It can be simplified as:

```go
_, err := f.ReadFrom(lk)
if errorCause(err) == io.EOF {
    return traceError(errUnformattedDisk)
}

return err
```

cmd/fs-v1-metadata.go
harshavardhana: Done updating.

donatello (changes requested):

cmd/format-config-v1.go
harshavardhana: Done

cmd/format-config-v1.go
harshavardhana: Done

cmd/fs-v1-metadata.go
```diff
@@ -274,167 +259,109 @@ func newFSMetaV1() (fsMeta fsMetaV1) {
...
+			// Load format to validate again, checks if we are unformatted
+			// writes the new format.json
+			var format = &formatConfigV1{}
+			lerr = format.LoadFormat(lk)
+			switch {
+			case lerr == nil:
+				// Formatting completed by another process, continue
+				// to attempt a read-lock and validate `format.json`.
+				lk.Close()
+				continue
```
donatello: lk.Close() can be after switch statement - this simplifies a couple of cases here.

harshavardhana: done

cmd/fs-v1-metadata.go
```diff
@@ -274,167 +259,106 @@ func newFSMetaV1() (fsMeta fsMetaV1) {
...
+				return traceError(err)
+			}
 
-		// Initialize format v2.
-		format = newFSFormatV2()
-	}
+			// Write new format.
+			if err = writeFormatFS(lk); err != nil {
+				return err
+			}
```
donatello: A comment here saying that we continue the loop to get a read-lock and then finally exit the loop would be helpful to readers.

donatello (approved):
This looks quite good to me - just have a final comment to aid readability.

PR #4525, opened 2017-06-11
Add support for DCOS host detection, improve Docker detection.

### LINKED ISSUE #4456: Update useragent string to indicate where minio was installed on
This would be very helpful for debugging and supportability.

harshavardhana:
## Description
Add support for DCOS host detection, improve Docker detection

isDocker was currently reading from `/proc/cgroup` file. But
this file alone is rather not conclusive evidence. Docker
internally has `.dockerenv` as a special file which we should
use instead.

## Motivation and Context
Fixes #4456

## How Has This Been Tested?
Manually and with tests. 

## Types of changes

## Checklist:

### DIFF
--- a/cmd/update-main.go
+++ b/cmd/update-main.go
@@ -17,7 +17,6 @@
 package cmd
 
 import (
-	"bytes"
 	"fmt"
 	"io/ioutil"
 	"net/http"
@@ -98,26 +97,41 @@ func GetCurrentReleaseTime() (releaseTime time.Time, err error) {
 	return getCurrentReleaseTime(Version, os.Args[0])
 }
 
-func isDocker(cgroupFile string) (bool, error) {
-	cgroup, err := ioutil.ReadFile(cgroupFile)
-	if os.IsNotExist(err) {
-		err = nil
+// Check if we are indeed inside docker.
+// https://github.com/moby/moby/blob/master/daemon/initlayer/setup_unix.go#L25
+//
+//     "/.dockerenv":      "file",
+//
+func isDocker(dockerEnvFile string) (ok bool, err error) {
+	_, err = os.Stat(dockerEnvFile)
+	if err != nil {
+		if os.IsNotExist(err) {
+			err = nil
+		}
+		return false, err
 	}
-
-	return bytes.Contains(cgroup, []byte("docker")), err
+	return true, nil
 }
 
 // IsDocker - returns if the environment minio is running
 // is docker or not.
 func IsDocker() bool {
-	found, err := isDocker("/proc/self/cgroup")
-	fatalIf(err, "Error in docker check.")
+	found, err := isDocker("/.dockerenv")
+	// We don't need to fail for this check, log
+	// an error and return false.
+	errorIf(err, "Error in docker check.")
 
 	return found
 }
 
-// IsKubernetes returns if the environment minio is
-// running is kubernetes or not.
+// IsDCOS returns true if minio is running in DCOS.
+func IsDCOS() bool {
+	// http://mesos.apache.org/documentation/latest/docker-containerizer/
+	// Mesos docker containerizer sets this value
+	return os.Getenv("MESOS_CONTAINER_NAME") != ""
+}
+
+// IsKubernetes returns true if minio is running in kubernetes.
 func IsKubernetes() bool {
 	// Kubernetes env used to validate if we are
 	// indeed running inside a kubernetes pod
@@ -139,13 +153,16 @@ func IsSourceBuild() bool {
 // DO NOT CHANGE USER AGENT STYLE.
 // The style should be
 //
-//   Minio (<OS>; <ARCH>[; kubernetes][; docker][; source]) Minio/<VERSION> Minio/<RELEASE-TAG> Minio/<COMMIT-ID>
+//   Minio (<OS>; <ARCH>[; dcos][; kubernetes][; docker][; source]) Minio/<VERSION> Minio/<RELEASE-TAG> Minio/<COMMIT-ID> [Minio/univers-<PACKAGE_NAME>]
 //
 // For any change here should be discussed by openning an issue at https://github.com/minio/minio/issues.
 func getUserAgent(mode string) string {
 	userAgent := "Minio (" + runtime.GOOS + "; " + runtime.GOARCH
 	if mode != "" {
 		userAgent += "; " + mode
+	}
+	if IsDCOS() {
+		userAgent += "; dcos"
 	}
 	if IsKubernetes() {
 		userAgent += "; kubernetes"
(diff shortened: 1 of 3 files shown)


cmd/update-main.go
```diff
@@ -111,13 +111,19 @@ func isDocker(cgroupFile string) (bool, error) {
 // is docker or not.
 func IsDocker() bool {
 	found, err := isDocker("/proc/self/cgroup")
-	fatalIf(err, "Error in docker check.")
+	// We don't need to fail for this check, print
+	// and error and return false.
+	errorIf(err, "Error in docker check.")
```
balamurugana: If we are unable to read `/proc/self/cgroup` why do we need to continue here?

cmd/update-main.go
```diff
@@ -100,7 +100,7 @@ func GetCurrentReleaseTime() (releaseTime time.Time, err error) {
 
 func isDocker(cgroupFile string) (bool, error) {
 	cgroup, err := ioutil.ReadFile(cgroupFile)
-	if os.IsNotExist(err) {
+	if os.IsNotExist(err) || os.IsPermission(err) {
```
balamurugana: Why do we need to ignore permission error?

balamurugana (changes requested):

cmd/update-main.go
harshavardhana: Because it is perfectly valid.

cmd/update-main.go
harshavardhana: Because we should and not fail the server. It is not binding requirement for this to be readable to start the server.

cmd/update-main.go
balamurugana: Do you mean in any other container system `/proc/self/cgroup` might not be permission to read the file?

harshavardhana: Yes. If you run as a regular user inside the container it's a possibility.

balamurugana: As per my findings, `/proc/self/cgroup` in lxc/docker have permission like
```
-r--r--r-- 1 bala bala 0 Jun 12 07:25 /proc/self/cgroup
```
or
```
-r--r--r-- 1 root root 0 Jun 12 01:55 /proc/self/cgroup
```
I don't know how DC/OS does.

This means the file is always readable by non-root.   I think permission error should not occur, if so there is some setup problem which is hidden by this check.

harshavardhana: I don't think we can infer that its a setup issue @balamurugana by the condition that its not readable. There is no set document that indicates that this file should be readable by everyone, it is readable perhaps is purely incidental and has nothing to do with it. 

Neither its a requirement for Minio server that this file should be readable, it is a guess work at best and works well for the most part because docker makes this file available with some `docker` strings in it.  We will be changing all this code to make a proper API call to know we are indeed inside docker.

Let me not ignore the permission error for now until we find a better solution.

harshavardhana: I even took the liberty of changing the `docker` check to use `.dockerenv` instead of reading `/proc/cgroup` file. Let me know what you think @balamurugana

balamurugana: :+1: Please update commit log/PR title/description accordingly.

harshavardhana: Done @balamurugana

cmd/update-main.go
```diff
@@ -98,26 +97,39 @@ func GetCurrentReleaseTime() (releaseTime time.Time, err error) {
...
+	_, err = os.Stat(dockerEnvFile)
+	if err != nil {
+		if os.IsNotExist(err) {
+			err = nil
+		}
+		return false, err
 	}
-
-	return bytes.Contains(cgroup, []byte("docker")), err
+	return true, nil
```
balamurugana, line 113: returning err is good enough and bool is not required. if it is accepted, below `if` is enough.
```go
if os.IsNotExist(err) {
    err = nil
}
```

cmd/update-main_test.go
```diff
@@ -158,6 +164,59 @@ func TestGetCurrentReleaseTime(t *testing.T) {
 	}
 }
 
+// Tests user agent string.
+func TestUserAgent(t *testing.T) {
+	customUserAgent := func(mode string, extraMode string) string {
```
balamurugana: I would recommend to have a constant string comparison over string generation functions.  This avoids making test code breakage.

balamurugana (changes requested):

cmd/update-main.go
```diff
@@ -98,26 +97,39 @@ func GetCurrentReleaseTime() (releaseTime time.Time, err error) {
...
 }
 
-func isDocker(cgroupFile string) (bool, error) {
-	cgroup, err := ioutil.ReadFile(cgroupFile)
-	if os.IsNotExist(err) {
-		err = nil
+// Check if we are indeed inside docker.
+// https://github.com/moby/moby/blob/master/daemon/initlayer/setup_unix.go#L25
+//
+//     "/.dockerenv":      "file",
```
krisis, line 103: Is this a stable interface that applications running on Docker can rely on?

cmd/update-main.go
```diff
@@ -98,26 +97,39 @@ func GetCurrentReleaseTime() (releaseTime time.Time, err error) {
...
+	errorIf(err, "Error in docker check.")
 
 	return found
 }
 
-// IsKubernetes returns if the environment minio is
-// running is kubernetes or not.
+// IsDCOS returns true if minio is running in DCOS.
+func IsDCOS() bool {
+	return os.Getenv("MARATHON_APP_LABEL_DCOS_PACKAGE_NAME") == "minio"
```
krisis: Is this a stable interface? Please add the relevant document describing the above environment variable usage in comments.

krisis (changes requested):

cmd/update-main.go
harshavardhana: Yes @krisis

cmd/update-main.go
harshavardhana: Yes @krisis

cmd/update-main_test.go
harshavardhana: Why would it break . I added it so that it works for all envs. I don't want to have verbatim copies for each platform.

cmd/update-main.go
harshavardhana: Returning bool is needed to know if it's docker or not. We can treat isNotExist as true.

balamurugana: My bad. I was trying to say returning bool and have `errorIf()` here.  Its ok to continue this way to avoid error message printing when unit test is run.

cmd/update-main_test.go
balamurugana: Because this function accepts switches, there is a possibility of breakage if wrong inputs are passed. Tabular cases and string comparison are clean and simple.

balamurugana (changes requested):

cmd/update-main_test.go
harshavardhana: That is upto us @balamurugana this is not a consumer function and we know the output format precisely. Having verbatim strings is verbose and can bring in more mistakes IMO.

cmd/update-main.go
```diff
@@ -156,8 +173,15 @@ func getUserAgent(mode string) string {
 	if IsSourceBuild() {
 		userAgent += "; source"
 	}
-	userAgent += ") " + " Minio/" + Version + " Minio/" + ReleaseTag + " Minio/" + CommitID
 
+	userAgent += ") Minio/" + Version + " Minio/" + ReleaseTag + " Minio/" + CommitID
+	if IsDCOS() {
+		dcosPkgVersion := os.Getenv("MARATHON_APP_LABEL_DCOS_PACKAGE_VERSION")
```
nitisht: probably `universePkgVersion` will be more suitable here.

nitisht:
LGTM except one comment.

nitisht (approved):

PR #4532, opened 2017-06-13
fs: Hold `format.json` readLock ref to avoid GC.

### LINKED ISSUE #4530: format.json FD gets closed
Start minio server in the usual manner and check that the format.json is kept open:

krishna@escape:~$ ls -l /proc/13134/fd
total 0
lrwx------ 1 krishna krishna 64 Jun 13 00:52 0 -> /dev/pts/0
lrwx------ 1 krishna krishna 64 Jun 13 00:52 1 -> /dev/pts/0
lrwx------ 1 krishna krishna 64 Jun 13 00:52 2 -> /dev/pts/0
lr-x------ 1 krishna krishna 64 Jun 13 00:52 3 -> /dev/tty
lrwx------ 1 krishna krishna 64 Jun 13 00:52 4 -> socket:[270135]
lrwx------ 1 krishna krishna 64 Jun 13 00:52 5 -> anon_inode:[eventpoll]
lr-x------ 1 krishna krishna 64 Jun 13 00:52 6 -> /home/krishna/export/.minio.sys/format.json

But after sometime (5-10 mins) the FD gets closed automatically:
krishna@escape:~$ ls -l /proc/13134/fd
total 0
lrwx------ 1 krishna krishna 64 Jun 13 00:52 0 -> /dev/pts/0
lrwx------ 1 krishna krishna 64 Jun 13 00:52 1 -> /dev/pts/0
lrwx------ 1 krishna krishna 64 Jun 13 00:52 2 -> /dev/pts/0
lr-x------ 1 krishna krishna 64 Jun 13 00:52 3 -> /dev/tty
lrwx------ 1 krishna krishna 64 Jun 13 00:52 4 -> socket:[270135]
lrwx------ 1 krishna krishna 64 Jun 13 00:52 5 -> anon_inode:[eventpoll]
krishna@escape:~$ 

Need to figure out how the FD is getting closed.

harshavardhana:
## Description
Looks like if we follow pattern such as

```
_ = rlk
```

Go can potentially kick in GC and close the fd when
the reference is lost, only speculation is that
the cause here is `SetFinalizer` which is set on
`os.close()` internally in `os` stdlib.

This is unexpected and unsual endeavour for Go, but
we have to make sure the reference is never lost
and always dies with the server.

## Motivation and Context
Fixes #4530

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:

### DIFF
--- a/cmd/fs-v1-metadata.go
+++ b/cmd/fs-v1-metadata.go
@@ -290,17 +290,23 @@ func checkLockedValidFormatFS(fsPath string) (*lock.RLockedFile, error) {
 	return rlk, traceError(err)
 }
 
-// Writes the new format.json if unformatted,
-// otherwise closes the input locked file
-// and returns any error.
-func writeFormatFS(lk *lock.LockedFile) error {
+// Creates a new format.json if unformatted.
+func createFormatFS(fsPath string) error {
+	fsFormatPath := pathJoin(fsPath, minioMetaBucket, formatConfigFile)
+
+	// Attempt a write lock on formatConfigFile `format.json`
+	// file stored in minioMetaBucket(.minio.sys) directory.
+	lk, err := lock.TryLockedOpenFile(preparePath(fsFormatPath), os.O_RDWR|os.O_CREATE, 0600)
+	if err != nil {
+		return traceError(err)
+	}
 	// Close the locked file upon return.
 	defer lk.Close()
 
 	// Load format on disk, checks if we are unformatted
 	// writes the new format.json
 	var format = &formatConfigV1{}
-	err := format.LoadFormat(lk)
+	err = format.LoadFormat(lk)
 	if errorCause(err) == errUnformattedDisk {
 		_, err = newFSFormat().WriteTo(lk)
 		return err
@@ -308,20 +314,12 @@ func writeFormatFS(lk *lock.LockedFile) error {
 	return err
 }
 
-func initFormatFS(fsPath, fsUUID string) (err error) {
-	fsFormatPath := pathJoin(fsPath, minioMetaBucket, formatConfigFile)
-
-	// Once the filesystem has initialized hold the read lock for
-	// the life time of the server. This is done to ensure that under
-	// shared backend mode for FS, remote servers do not migrate
-	// or cause changes on backend format.
-
+func initFormatFS(fsPath string) (rlk *lock.RLockedFile, err error) {
 	// This loop validates format.json by holding a read lock and
 	// proceeds if disk unformatted to hold non-blocking WriteLock
 	// If for some reason non-blocking WriteLock fails and the error
 	// is lock.ErrAlreadyLocked i.e some other process is holding a
 	// lock we retry in the loop again.
-	var rlk *lock.RLockedFile
 	for {
 		// Validate the `format.json` for expected values.
 		rlk, err = checkLockedValidFormatFS(fsPath)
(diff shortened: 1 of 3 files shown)


cmd/fs-v1-metadata.go
```diff
@@ -290,38 +290,38 @@ func checkLockedValidFormatFS(fsPath string) (*lock.RLockedFile, error) {
 	return rlk, traceError(err)
 }
 
-// Writes the new format.json if unformatted,
+// Creates a new format.json if unformatted,
 // otherwise closes the input locked file
```
krisis: The following comment doesn't make sense for `createFormatFS`
>closes the input locked file

cmd/fs-v1.go
```diff
@@ -42,6 +42,9 @@ type fsObjects struct {
 	// temporary transactions.
 	fsUUID string
 
+	// This value shouldn't be touched, once initialized.
+	fsFormatRlk *lock.RLockedFile // Is a read lock remembered on `format.json`.
```
krisis: suggestion: "Is a read lock on `format.json`" reads better. The value being remembered is implicit.

cmd/fs-v1.go
```diff
@@ -122,10 +131,11 @@ func newFSObjectLayer(fsPath string) (ObjectLayer, error) {
...
 
-	// Initialize `format.json`.
-	if err = initFormatFS(fsPath, fsUUID); err != nil {
-		return nil, err
-	}
+	// Once the filesystem has initialized hold the read lock for
+	// the life time of the server. This is done to ensure that under
+	// shared backend mode for FS, remote servers do not migrate
+	// or cause changes on backend format.
+	fs.fsFormatRlk = rlk
```
krisis, line 138: suggestion: we could assign `rlk` as part of `fs := &fsObjects {...}` object literal.

krisis (changes requested):
LGTM except for minor comments. Tested locally and works as expected.

cmd/fs-v1.go
harshavardhana: Kept it outside to add a laborious comment, so its readable.

cmd/fs-v1.go
harshavardhana: Sure

cmd/fs-v1-metadata.go
harshavardhana: Sure.

krisis (approved):

balamurugana (approved):
Tested

vadmeste:
Tested with this script which runs 16 shared instances.

```
#!/bin/bash

INSTANCES=16
FS_DIR="/tmp/fs-dir-test-lock"
MINIO_BIN="$GOPATH/bin/minio"
SCHEME="http" # or https

rm -rf $FS_DIR

echo "Starting instances..."

(for i in $(seq 1 $INSTANCES); do echo "$[9000+$i]" ; done | \
	parallel --no-notice -I{} -j $INSTANCES $MINIO_BIN server --address :{} $FS_DIR >/dev/null ) &

sleep 5

for i in $(seq 1 $INSTANCES); do
	resp=$(curl "$SCHEME://localhost:$[9000+$i]/" 2>/dev/null)
	if [[ "$resp" != *"<Error><Code>AccessDenied</Code>"* ]]; then
		echo "Unexpected response from one instance: $resp"
		exit
	fi
done

echo "Tests passed.."
```
If we call this script test_shared.sh, it is better to run it this way:
`$ killall minio && bash test_shared.sh`

vadmeste (approved):

PR #4543, opened 2017-06-15
gateway/azure: Parse error responses for anonymous requests

### LINKED ISSUE #4481: [gateway-azure] curl to a bucket causes 500 internal error
## Expected Behavior

curl should not cause 500 errors in the server.

## Current Behavior

curl causes 500 errors when accessing many urls in Azure gateway mode.

## Possible Solution

Improve validation.

1. Start Minio in Azure gateway mode.
2. Use to make any request to a "bucket" url:
```
$ curl -v -s "http://localhost:9000/asdfasdf"
*   Trying 127.0.0.1...
* Connected to localhost (127.0.0.1) port 9000 (#0)
> GET /asdfasdf HTTP/1.1
> Host: localhost:9000
> User-Agent: curl/7.47.0
> Accept: */*
> 
< HTTP/1.1 500 Internal Server Error
< Accept-Ranges: bytes
< Content-Type: application/xml
< Server: Minio/RELEASE.2017-06-02T22-13-34Z (linux; amd64)
< Vary: Origin
< X-Amz-Bucket-Region: 
< X-Amz-Request-Id: 14C56205E722FB40
< Date: Tue, 06 Jun 2017 00:50:28 GMT
< Transfer-Encoding: chunked
< 
* Connection #0 to host localhost left intact
<Error><Code>InternalError</Code><Message>We encountered an internal error, please try again.</Message><Key></Key><BucketName></BucketName><Resource>/asdfasdf</Resource><RequestId>3L137</RequestId><HostId>3L137</HostId></Error>
```
On server, we see an error log like:
```
ERRO[0085] Unable to list objects.                       cause=expected element type <EnumerationResults> but have <Error> source=[gateway-handlers.go:767:gatewayAPIHandlers.ListObjectsV1Handler()] stack=home/builder/mygo/src/github.com/minio/minio/cmd/gateway-azure-anonymous.go:170:(*azureObjects).AnonListObjects home/builder/mygo/src/github.com/minio/minio/cmd/gateway-handlers.go:760:(GatewayLayer).AnonListObjects-fm home/builder/mygo/src/github.com/minio/minio/cmd/gateway-handlers.go:765:gatewayAPIHandlers.ListObjectsV1Handler home/builder/mygo/src/github.com/minio/minio/cmd/gateway-router.go:105:(gatewayAPIHandlers).ListObjectsV1Handler-fm
```
3. On plain old Minio server this is perfectly handled:
```
$ curl -v -s "http://localhost:9000/asdfasdf"
*   Trying 127.0.0.1...
* Connected to localhost (127.0.0.1) port 9000 (#0)
> GET /asdfasdf HTTP/1.1
> Host: localhost:9000
> User-Agent: curl/7.47.0
> Accept: */*
> 
< HTTP/1.1 404 Not Found
< Accept-Ranges: bytes
< Content-Type: application/xml
< Server: Minio/DEVELOPMENT.2017-06-05T18-57-45Z (linux; amd64)
< Vary: Origin
< X-Amz-Bucket-Region: us-east-1
< X-Amz-Request-Id: 14C562C6A068935A
< Date: Tue, 06 Jun 2017 01:04:16 GMT
< Transfer-Encoding: chunked
< 
* Connection #0 to host localhost left intact
<Error><Code>NoSuchBucket</Code><Message>The specified bucket does not exist</Message><Key></Key><BucketName></BucketName><Resource>/asdfasdf</Resource><RequestId>3L137</RequestId><HostId>3L137</HostId></Error>
```
4.

## Your Environment

* Server type and version:
* Link to your project:

krishnasrinivas:
fixes #4481

## Description

For anonymous requests to Azure we were not parsing the Error XML responses.

## How Has This Been Tested?

manually

## Types of changes

## Checklist:

### DIFF
--- a/cmd/gateway-azure-anonymous.go
+++ b/cmd/gateway-azure-anonymous.go
@@ -29,6 +29,47 @@ import (
 	"github.com/Azure/azure-sdk-for-go/storage"
 )
 
+// Make anonymous HTTP request to azure endpoint.
+func azureAnonRequest(verb, urlStr string, header http.Header) (*http.Response, error) {
+	req, err := http.NewRequest(verb, urlStr, nil)
+	if err != nil {
+		return nil, err
+	}
+	if header != nil {
+		req.Header = header
+	}
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return nil, err
+	}
+
+	if resp.StatusCode >= 400 && resp.StatusCode <= 505 {
+		defer resp.Body.Close()
+		respBody, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			return nil, err
+		}
+
+		if len(respBody) == 0 {
+			// no error in response body, might happen in HEAD requests
+			return nil, storage.AzureStorageServiceError{
+				StatusCode: resp.StatusCode,
+				Code:       resp.Status,
+				Message:    "no response body was available for error status code",
+			}
+		}
+		// Response contains Azure storage service error object.
+		var storageErr storage.AzureStorageServiceError
+		if err := xml.Unmarshal(respBody, &storageErr); err != nil {
+			return nil, err
+		}
+		storageErr.StatusCode = resp.StatusCode
+		return nil, storageErr
+	}
+
+	return resp, nil
+}
+
 // AnonGetBucketInfo - Get bucket metadata from azure anonymously.
 func (a *azureObjects) AnonGetBucketInfo(bucket string) (bucketInfo BucketInfo, err error) {
 	url, err := url.Parse(a.client.GetBlobURL(bucket, ""))
@@ -36,11 +77,11 @@ func (a *azureObjects) AnonGetBucketInfo(bucket string) (bucketInfo BucketInfo, 
 		return bucketInfo, azureToObjectError(traceError(err))
 	}
 	url.RawQuery = "restype=container"
-	resp, err := http.Head(url.String())
+	resp, err := azureAnonRequest(httpHEAD, url.String(), nil)
 	if err != nil {
 		return bucketInfo, azureToObjectError(traceError(err), bucket)
 	}
-	resp.Body.Close()
+	defer resp.Body.Close()
 
 	if resp.StatusCode != http.StatusOK {
 		return bucketInfo, azureToObjectError(traceError(anonErrToObjectErr(resp.StatusCode, bucket)), bucket)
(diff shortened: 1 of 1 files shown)


harshavardhana:
@krishnasrinivas can you add unit tests for the changes?

krishnasrinivas:
@harshavardhana azureAnonRequest can't be unit tested as it needs to talk to azure backend. This would be covered under mint.

harshavardhana:
> @harshavardhana azureAnonRequest can't be unit tested as it needs to talk to azure backend. This would be covered under mint.

The parsing part if possible can go in unit test since it is specific for azure to s3 translation.

krishnasrinivas:
> The parsing part if possible can go in unit test since it is specific for azure to s3 translation.

```
+		if len(respBody) == 0 {
 +			// no error in response body, might happen in HEAD requests
 +			return nil, storage.AzureStorageServiceError{
 +				StatusCode: resp.StatusCode,
 +				Code:       resp.Status,
 +				Message:    "no response body was available for error status code",
 +			}
 +		}
 +		// Response contains Azure storage service error object.
 +		var storageErr storage.AzureStorageServiceError
 +		if err := xml.Unmarshal(respBody, &storageErr); err != nil {
 +			return nil, err
 +		}
 +		storageErr.StatusCode = resp.StatusCode
 +		return nil, storageErr
```

This part can be separated out into a function, but the downside is it will create a small function which is not reused. Let me check with @abperiasamy once what he thinks

harshavardhana:
> This part can be separated out into a function, but the downside is it will create a small function which is not reused. Let me check with @abperiasamy once what he thinks

It doesn't need to be reused the problem is that we have many functions like this which are not unit tested because they are deep inside, if we do have one function for this we can add test cases for all common and corner cases where this can fail. 

Say if we do reflection or something internally.  My suggestion is we add tests..

cmd/gateway-azure-anonymous.go
```diff
@@ -29,18 +29,59 @@ import (
...
+	}
+	if header != nil {
+		req.Header = header
+	}
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return nil, err
+	}
+
+	if resp.StatusCode >= 400 && resp.StatusCode <= 505 {
```
krisis: A comment explaining why we pick status code between 400 and 505 is required.

krisis (changes requested):
Overall the changes look good to me. @krishnasrinivas do you plan to add unit tests? I see that there was a discussion but no decision mentioned in the conversation.

cmd/gateway-azure-anonymous.go
```diff
@@ -29,18 +29,59 @@ import (
 	"github.com/Azure/azure-sdk-for-go/storage"
 )
 
+// Make anonymous HTTP request to azure endpoint.
+func azureAnonRequest(verb, urlStr string, header http.Header) (*http.Response, error) {
```
balamurugana, line 33: Please add unit test to this function.  You could run a test http server in the unit test and address various cases.

cmd/gateway-azure-anonymous.go
```diff
@@ -29,18 +29,59 @@ import (
...
+		if err != nil {
+			return nil, err
+		}
+
+		if len(respBody) == 0 {
+			// no error in response body, might happen in HEAD requests
+			return nil, storage.AzureStorageServiceError{
+				StatusCode: resp.StatusCode,
+				Code:       resp.Status,
+				Message:    "no response body was available for error status code",
```
balamurugana, line 59: Why is missing response body wrong?

balamurugana (changes requested):

cmd/gateway-azure-anonymous.go
krishnasrinivas: We had decided not to do mock tests. i.e we should not start a mock server and test instead should use the actual backend in functional tests in mint to cover cases like these.

cmd/gateway-azure-anonymous.go
krishnasrinivas: I just picked this code from azure sdk
https://en.wikipedia.org/wiki/List_of_HTTP_status_codes says 4xx 5xx are error codes

cmd/gateway-azure-anonymous.go
krishnasrinivas: @balamurugana note that this is inside the if condition: 
```
if resp.StatusCode >= 400 && resp.StatusCode <= 505
```

hence it is an error condition regardless of whether body is available or not

krishnasrinivas:
> Overall the changes look good to me. @krishnasrinivas do you plan to add unit tests? I see that there was a discussion but no decision mentioned in the conversation.

@krisis AB said functions like this which needs to talk to GCS backend should be covered under functional tests (to test against the actual backend) and that we should not do mock/unit tests.

balamurugana (approved):
Not tested

krisis (approved):
Looks like we have decided what to do about unit tests. Lgtm otherwise.

PR #4548, opened 2017-06-16
fs: Add logging if `format.json` takes too long.

### LINKED ISSUE #4533: minio looks hung when format.json is opened with write lock.
When `format.json` is opened by another process and holding write lock, starting minio looks hung ie nothing printed.

## Expected Behavior

We would need to show some message on retries to get the lock.

## Current Behavior

Nothing printed.

1. run below code before starting minio
```go
	wlockedFile, err := lock.TryLockedOpenFile("d1/.minio.sys/format.json", os.O_RDWR|os.O_CREATE, 0600)
	if err != nil {
		fmt.Println("unable to open wlocked file", err)
		return
	}

	fmt.Println("got wlockedFile", wlockedFile)
	time.Sleep(1 * time.Minute)
```

2.
3.
4.

## Your Environment

* Server type and version:
* Link to your project:

harshavardhana:
## Description
Add logging if `format.json` takes too long.

## Motivation and Context
Fixes #4533

## How Has This Been Tested?
Manually using the test code provided #4533 

## Types of changes

## Checklist:

### DIFF
--- a/cmd/fs-v1-metadata.go
+++ b/cmd/fs-v1-metadata.go
@@ -26,6 +26,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/minio/mc/pkg/console"
 	"github.com/minio/minio/pkg/lock"
 	"github.com/minio/minio/pkg/mimedb"
 	"github.com/tidwall/gjson"
@@ -315,16 +316,23 @@ func createFormatFS(fsPath string) error {
 }
 
 func initFormatFS(fsPath string) (rlk *lock.RLockedFile, err error) {
-	// This loop validates format.json by holding a read lock and
-	// proceeds if disk unformatted to hold non-blocking WriteLock
-	// If for some reason non-blocking WriteLock fails and the error
-	// is lock.ErrAlreadyLocked i.e some other process is holding a
-	// lock we retry in the loop again.
 	for {
+		printedCh := delayPrint("Waiting to acquire lock.. ", time.Second)
+
 		// Validate the `format.json` for expected values.
 		rlk, err = checkLockedValidFormatFS(fsPath)
+
+		// Waits on delayed printer to send back notification
+		// if the previous message was indeed printed.
+		printed := <-printedCh
 		switch {
 		case err == nil:
+			// Print this message only if previously
+			// read-lock was waiting and blocked.
+			if printed {
+				console.Println("Lock acquired.")
+			}
+
 			// Holding a read lock ensures that any write lock operation
 			// is blocked if attempted in-turn avoiding corruption on
 			// the backend disk.
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -30,6 +30,7 @@ import (
 	"time"
 
 	humanize "github.com/dustin/go-humanize"
+	"github.com/minio/mc/pkg/console"
 	"github.com/pkg/profile"
 )
 
@@ -158,6 +159,24 @@ func startProfiler(profiler string) interface {
 // Global profiler to be used by service go-routine.
 var globalProfiler interface {
 	Stop()
+}
+
+// Prints a given message after a requested delay.
+func delayPrint(message string, delay time.Duration) chan bool {
+	printedCh := make(chan bool)
+	go func() {
+		defer close(printedCh)
+		select {
+		case printedCh <- false:
+			return
+		case <-time.After(delay):
+			console.Print(message)
+			// Sends true to indicate that we
+			// printed successfully to console.
+			printedCh <- true
+		}
+	}()
+	return printedCh
 }
 
 // dump the request into a string in JSON format.


cmd/fs-v1-metadata.go
```diff
@@ -315,37 +316,55 @@ func createFormatFS(fsPath string) error {
...
-			return rlk, nil
-		case errorCause(err) == errUnformattedDisk:
-			if err = createFormatFS(fsPath); err != nil {
-				// Existing write locks detected.
-				if errorCause(err) == lock.ErrAlreadyLocked {
-					// Lock already present, sleep and attempt again.
-					time.Sleep(100 * time.Millisecond)
-					continue
+		select {
+		case _ = <-retryTimerCh:
```
krishnasrinivas: ~~If we receive more than once on this channel then we will end up calling checkLockedValidFormatFS() more than once.~~

harshavardhana: why would we receive more?

krishnasrinivas: scratched my comment

krishnasrinivas:
Printing this every time does not look good:

```
Initializing FS backend format.. Waiting for lock to be acquired.
Lock acquired, safely proceeding to intialize object layer.
```

Instead we can log only when we have not been able to acquire lock within sometime, this change does that:
```
diff --git a/cmd/fs-v1-metadata.go b/cmd/fs-v1-metadata.go
index f375c4d..f7c9584 100644
--- a/cmd/fs-v1-metadata.go
+++ b/cmd/fs-v1-metadata.go
@@ -25,6 +25,7 @@ import (
        pathutil "path"
        "sort"
        "strings"
+       "time"
 
        "github.com/minio/mc/pkg/console"
        "github.com/minio/minio/pkg/lock"
@@ -315,6 +316,20 @@ func createFormatFS(fsPath string) error {
        return err
 }
 
+func delayedPrint(msg string, delay time.Duration) chan bool {
+       ch := make(chan bool)
+       go func() {
+               select {
+               case ch <- false:
+                       return
+               case <-time.After(delay):
+                       console.Println(msg)
+                       ch <- true
+               }
+       }()
+       return ch
+}
+
 func initFormatFS(fsPath string) (rlk *lock.RLockedFile, err error) {
        // Done channel is used to close any lingering retry routine, as soon
        // as this function returns.
@@ -328,20 +343,22 @@ func initFormatFS(fsPath string) (rlk *lock.RLockedFile, err error) {
        // If for some reason non-blocking WriteLock fails and the error
        // is lock.ErrAlreadyLocked i.e some other process is holding a
        // lock we retry in the loop again.
-       console.Printf("Initializing FS backend format.. ")
 
        // Wait on the jitter retry loop.
        retryTimerCh := newRetryTimerSimple(doneCh)
        for {
                select {
                case _ = <-retryTimerCh:
-                       console.Println("Waiting for lock to be acquired.")
 
+                       ch := delayedPrint("Waiting to acquire lock", time.Second)
                        // Validate the `format.json` for expected values.
                        rlk, err = checkLockedValidFormatFS(fsPath)
+                       if <-ch {
+                               console.Println("Lock acquired.")
+                       }
+
                        switch {
                        case err == nil:
-                               console.Println("Lock acquired, safely proceeding to intialize object layer.")
                                // Holding a read lock ensures that any write lock operation
                                // is blocked if attempted in-turn avoiding corruption on
                                // the backend disk.
```

harshavardhana:
> Initializing FS backend format.. Waiting for lock to be acquired.
> Lock acquired, safely proceeding to intialize object layer.

This is not printed everytime @krishnasrinivas - it prints at max twice in almost all cases. I would avoid to adding new select loops and go-routines for printing alone. 

This is how it is going to look like usually in the worst case.. 
```
Initializing FS backend format.. Waiting for lock to be acquired.
Waiting for lock to be acquired.
Lock acquired, safely proceeding to intialize object layer.
```

harshavardhana:
This is updated - please verify @donatello @krishnasrinivas

cmd/utils.go
```diff
@@ -160,6 +161,24 @@ var globalProfiler interface {
 	Stop()
 }
 
+// Prints a given message after a requested delay.
+func delayPrint(message string, delay time.Duration) chan bool {
```
balamurugana, line 165: Why is it needed?

cmd/service.go
```diff
@@ -102,10 +104,10 @@ func (m *ServerMux) handleServiceSignals() error {
 			}
 			runExitFn(nil)
 		case serviceStop:
-			log.Println("Received signal to exit.")
+			console.Println("Received signal to exit.")
 			go func() {
 				time.Sleep(serverShutdownPoll + time.Millisecond*100)
-				log.Println("Waiting for active connections to terminate - press Ctrl+C to quit immediately.")
+				console.Println("Waiting for active connections to terminate - press Ctrl+C to quit immediately.")
```
balamurugana: Why do you use `console` than `log`?

cmd/fs-v1-metadata.go
```diff
@@ -315,37 +317,63 @@ func createFormatFS(fsPath string) error {
...
+				// Unhandled error return.
 				return nil, err
 			}
-
-			// Loop will continue to attempt a read-lock on `format.json`.
-		default:
-			// Unhandled error return.
-			return nil, err
+		case <-globalServiceDoneCh:
+			return nil, errors.New("Initializing FS backend stopped successfully")
```
balamurugana: I am not able to understand the meaning of the message.  Are you saying `FS backend initialization canceled`

cmd/fs-v1-metadata.go
```diff
@@ -315,37 +317,63 @@ func createFormatFS(fsPath string) error {
 }
 
 func initFormatFS(fsPath string) (rlk *lock.RLockedFile, err error) {
+	// Done channel is used to close any lingering retry routine, as soon
```
balamurugana: Please update/add respective unit tests.

balamurugana (changes requested):
Why unrelated fix are done in this PR?

cmd/service.go
harshavardhana: To avoid formatting of messages.

cmd/utils.go
harshavardhana: It is needed to print in the situation when perhaps format.json is locked unusually longer and user is not able to see what is going on with minio server.

cmd/service.go
balamurugana: What do you mean by `formatting of messages`?  Have you seen `log.Println()` doing extra formatting?

cmd/fs-v1-metadata.go
harshavardhana: They can't be added.. because you would require a separate process, - this will be taken care in our functional tests.

cmd/fs-v1-metadata.go
harshavardhana: its a error message returned during cancelled init, perhaps the string you mentioned is better.

cmd/service.go
harshavardhana: Just want to avoid that anyways.. log is usually a sort of formatted printer.

harshavardhana:
> Why unrelated fix are done in this PR?

What is unrelated? @balamurugana

cmd/utils.go
balamurugana: Why not print a message before lock attempt than delayed printing?

cmd/fs-v1-metadata.go
balamurugana: What do you mean by separate process?  What is the limitation of adding unit tests?

harshavardhana: Basically the conditions which don't reach in codecoverage right now need to have a situation where locks are held by a separate process so that the fcntl locks can be honored. Holdings fcntl locks on the same process have no meaning @balamurugana 

So it can't be added..

cmd/utils.go
harshavardhana: Because it will be printed all the time we need to avoid only when there is a blocking lock and some amount of time has been spent.

cmd/service.go
harshavardhana: Reverted this for now.. @balamurugana since its unrelated to this PR.

balamurugana:
>> Why unrelated fix are done in this PR?
> 
> What is unrelated?

Why `switch` block is changed in `initFormatFS()`?

harshavardhana:
> Why switch block is changed in initFormatFS()?

There was an existing problem @balamurugana we couldn't stop a server if it was waiting too long based based on `globalServiceDoneCh`  - Switch is still there its just moved inside `select` statement.

cmd/fs-v1-metadata.go
balamurugana: the unit test is a separate program holds a lock.  It has no difference whether its a separate process or not.  Unable to understand `no meaning` part.

cmd/utils.go
balamurugana: By the time the print initiated, there is a change of lock acquisition.  This doesn't give much value.  We just need to print before lock attempt.  That's good enough.

harshavardhana: It actually does.. if lock was acquired quickly we never print it. Our real estate on console is minimal we should just avoid printing it if it's not needed. So we needed this style.

balamurugana: For tiny benefit I wonder why we need such goroutine/channel code.  What do you mean by `our real estate on console is minimal`?

harshavardhana: No the benefit to UI was indicated by @krishnasrinivas .  So it was incorporated . Let me ask him to chime in here.

krishnasrinivas: > For tiny benefit I wonder why we need such goroutine/channel code

It is not a tiny benifit @balamurugana it affects the UI on the console. Refer to my comments earlier in the PR on why this go-routine was added.

Basically in this commit previously we were printing:
```
Initializing FS backend format.. Waiting for lock to be acquired.
Lock acquired, safely proceeding to intialize object layer.
```

everytime minio was started. This message is only useful in the rare case of another process migrating the backend and your lock will block. Hence it is worth a little complexity in the code to print it only when it is needed and not print all the time.

balamurugana: I agree the previous messaging is cumbersome.  Printing just a message like `Initializing FS backend` or `Initializing backend format` is good enough.   Complex code for this would cause penalty in code maintainability.

harshavardhana: > This message is only useful in the rare case of another process migrating the backend and your lock will block. Hence it is worth a little complexity in the code to print it only when it is needed and not print all the time.

@balamurugana you didn't read the message properly, the delayed print is added to avoid printing unless its necessary. Say if the lock was acquired with in a second which is most common this is never printed.  Its not really about what the message string should look like.

balamurugana: @harshavardhana I am aware of that.  My implicit question is why not print a clean message all the time than complex print logic.

harshavardhana: I am okay with that. But I want to get a conclusion from @krishnasrinivas . So current PR is basically nothing but what @krishnasrinivas proposed.

krishnasrinivas: I would prefer not printing it all the time. What we can do is quickly get @donatello 's vote and move forward.

cmd/utils.go
```diff
@@ -160,6 +161,24 @@ var globalProfiler interface {
...
+	printedCh := make(chan bool)
+	go func() {
+		defer close(printedCh)
```
donatello, line 168: This line is superfluous.

cmd/utils.go
```diff
@@ -160,6 +161,24 @@ var globalProfiler interface {
 	Stop()
 }
 
+// Prints a given message after a requested delay.
```
donatello, line 164: I am ok with a little complexity to print the message only when needed, this function is not very reusable without documenting the "protocol" to use it (i.e. need to listen on the returned channel, etc). Perhaps it can just be placed in the file where it is used, alternatively, for better reusability, I suggest:

1. making the returned channel - a "cancellation" channel - i.e. if we write to the channel the printing is cancelled (if it has not already happened) [rather than the current pattern of reading from it]
2. buffer the returned channel to size 1 - this makes sure that there is no unnecessary blocking when either sender/receiver is not listening on the returned channel - it makes sure that cancelling the print is an optional behaviour.

donatello (changes requested):

cmd/utils.go
krishnasrinivas: > making the returned channel - a "cancellation" channel - i.e. if we write to the channel the printing is cancelled (if it has not already happened) [rather than the current pattern of reading from it]

@donatello we print "Lock acquired" only if we had printed "Trying to acquire lock" hence we read from the channel to see the go routine has printed. I think cancellation channel can't fulfill this requirement.

donatello: This looks like it will work:

```go
// Prints a given message after a requested delay. The printing can be
// (optionally) cancelled by writing to or closing the returned
// channel. If `optFollowUp` is non-empty, then the returned channel
// is used as a signal to print a follow-up message (i.e. when a
// signal is received, if the printing has occurred, the follow-up
// message will be printed.)
func delayPrint1(message string, delay time.Duration, optFollowUp string) chan struct{} {
	cancelCh := make(chan struct{}, 1)
	go func() {
		select {
		case <-cancelCh:
		case <-time.After(delay):
			console.Print(message)
			// Is there a follow-up message to print?
			if len(optFollowUp) > 0 {
				// Listen for the signal on the cancel
				// channel
				<-cancelCh
				// Print follow-up message
				console.Print(msgFollowUp)
			}
		}
		return
	}()
	return cancelCh
}
```

Usage:

```go
cancelCh := delayPrint1("Acquiring lock...", time.Second, "done acquiring lock!")
// Do something that *may* take more than a second.
//...
//...
// Now close the channel.
close(cancelCh)
```

Notice that the function also works as a delayed printer too:

```go
delayPrint1("This line was asked to be printed just a second ago", time.Second, "")
// do something for at least a second for the message to appear.
time.Sleep(1)
```

cmd/fs-v1-metadata.go
```diff
@@ -315,16 +316,23 @@ func createFormatFS(fsPath string) error {
...
 		rlk, err = checkLockedValidFormatFS(fsPath)
+
+		// Waits on delayed printer to send back notification
+		// if the previous message was indeed printed.
+		printed := <-printedCh
 		switch {
 		case err == nil:
+			// Print this message only if previously
+			// read-lock was waiting and blocked.
+			if printed {
```
donatello, line 332: Even in this implementation, I think there is a bug. The readlock may block when a write lock is acquired by someone else, but if a write lock is attempted, there are sleep statements and formatting actions that could take a while - in this case the message needs to be printed. So the `printedCh` should be listened on only in the `err == nil` case. In the current implementation, the sleep-after-writelock-attempt and formatting steps might take time too, but the message would not be printed.

donatello (changes requested):

harshavardhana:
Closing this as not needed in lieu of new backend change which @krishnasrinivas is bringing in the next set of changes.  This change was needed only for situation when we have a long running process of backend migration which we might not do with in the same process.

PR #4554, opened 2017-06-19
fix: OwnerID in response should be 64 character in length.

### LINKED ISSUE #4553: OwnerID too short; does not match AWS spec
Minio reports the OwnerID as just "minio", but this causes at least one S3 library to fail, as they expect a 64 character string.

The AWS specifications say:
> The canonical user ID is the Amazon S3–only concept. It is s
> 64-character obfuscated version of the account ID.

That quote taken from: http://docs.aws.amazon.com/AmazonS3/latest/dev/example-walkthroughs-managing-access-example4.html

Failing S3 client: https://github.com/rustyconover/net-amazon-s3/issues/18
My patch for it: https://github.com/strategicdata/net-amazon-s3/commit/1b9b52d3aece937cb853b1f9f25bb3bab2713310

I'm not sure this really needs to be fixed in Minio, but I thought I would alert you to the issue.

harshavardhana:
## Description
Rather than sending a custom "minio" string, we can
change this to `sha256('arn:aws:iam::minio:user/admin')`.

## Motivation and Context
Fixes #4553

## How Has This Been Tested?
Manually and using `go test`

## Types of changes

## Checklist:

### DIFF
--- a/cmd/globals.go
+++ b/cmd/globals.go
@@ -29,8 +29,17 @@ import (
 const (
 	globalMinioCertExpireWarnDays = time.Hour * 24 * 30 // 30 days.
 
-	globalMinioDefaultRegion       = ""
-	globalMinioDefaultOwnerID      = "minio"
+	globalMinioDefaultRegion = ""
+	// This is a sha256 output of ``arn:aws:iam::minio:user/admin``,
+	// this is kept in present form to be compatible with S3 owner ID
+	// requirements -
+	//
+	// ```
+	//    The canonical user ID is the Amazon S3–only concept.
+	//    It is 64-character obfuscated version of the account ID.
+	// ```
+	// http://docs.aws.amazon.com/AmazonS3/latest/dev/example-walkthroughs-managing-access-example4.html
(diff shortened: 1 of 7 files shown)


cmd/globals.go
```diff
@@ -29,8 +29,18 @@ import (
...
-	globalMinioDefaultRegion       = ""
-	globalMinioDefaultOwnerID      = "minio"
+	globalMinioDefaultRegion = ""
+	// This is a sha256 output of ``arn:aws:iam::minio:user/admin``,
+	// this is kept in present form to be compatible with S3 owner ID
+	// requirements -
+	//
+	// ```
+	//    The canonical user ID is the Amazon S3–only concept.
+	//    It is s 64-character obfuscated version of the account
```
poornas: typo here-> It is s 64-character

poornas (approved):

harshavardhana:
ping @vadmeste can you review this?

vadmeste (approved):
Tested with basic mc operations, (put, get, policy, download)

PR #4556, opened 2017-06-19
Add e.g for setting MINIO_REGION env variable

### LINKED ISSUE #4555: specify minio server region using environment variable
Is there a way to specify the region using an environment variable, rather than the config.json?

I went through the docs, but found the method to set region by modifying the config.json.
That works fine, but was wondering if there is any `MINIO_REGION` type of variable that can be used?

My use case is that modifying the config.json at runtime may not be possible, but setting environment variables would be during the start of `minio`.

Thanks and Regards,
Shantanu

krisis:
Fixes #4555 
## Description

Add an example shell snippet to show how to set region for minio server via environment variable.

## Motivation and Context

See #4555 
## How Has This Been Tested?

Manually
## Types of changes

## Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -48,7 +48,7 @@ minio server ~/Photos
 #### Region
 |Field|Type|Description|
 |:---|:---|:---|
-|``region``| _string_ | `region` describes the physical location of the server. By default it is set to `us-east-1`, which is same as AWS3's default region. You may override this field with `MINIO_REGION` environment variable. If you are unsure leave it to default.|
+|``region``| _string_ | `region` describes the physical location of the server. By default it is set to `us-east-1`, which is same as AWS S3's default region. You may override this field with `MINIO_REGION` environment variable. If you are unsure leave it unset.|
 
 Example:
 
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -48,7 +48,7 @@ minio server ~/Photos
 #### Region
 |Field|Type|Description|
 |:---|:---|:---|
-|``region``| _string_ | `region` describes the physical location of the server. By default it is set to `us-east-1`, which is same as AWS3's default region. If you are unsure leave it to default.|
+|``region``| _string_ | `region` describes the physical location of the server. By default it is set to `us-east-1`, which is same as AWS3's default region. You may override this field with `MINIO_REGION` environment variable. If you are unsure leave it to default.|
 
 Example:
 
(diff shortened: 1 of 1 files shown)


docs/config/README.md
```diff
@@ -50,6 +50,13 @@ minio server ~/Photos
 |:---|:---|:---|
 |``region``| _string_ | `region` describes the physical location of the server. By default it is set to `us-east-1`, which is same as AWS3's default region. If you are unsure leave it to default.|
```
donatello: The environment variable also needs to be mentioned here (in the table, like for MINIO_ACCESS_KEY)

donatello (changes requested):

docs/config/README.md
harshavardhana: Since we decided to document this.. we should add it with `minio server` 

```
$ minio server <command help>
ENVIRONMENT VARIABLES:
  ACCESS:
     MINIO_ACCESS_KEY: Custom username or access key of 5 to 20 characters in length.
     MINIO_SECRET_KEY: Custom password or secret key of 8 to 40 characters in length.

  BROWSER:
     MINIO_BROWSER: To disable web browser access, set this value to "off".
```

harshavardhana (changes requested):

krisis:
@harshavardhana Good point, I shall send a PR to minio server adding this to the banner.

docs/config/README.md
```diff
@@ -48,7 +48,14 @@ minio server ~/Photos
 #### Region
 |Field|Type|Description|
 |:---|:---|:---|
-|``region``| _string_ | `region` describes the physical location of the server. By default it is set to `us-east-1`, which is same as AWS3's default region. If you are unsure leave it to default.|
```
donatello, line 51: s/AWS3/AWS S3/

s/If you are unsure leave it to default./If you are unsure leave it unset./

donatello (changes requested):

donatello (approved):

harshavardhana (approved):

donatello:
Looks like https://github.com/minio/minio/pull/4558 can be added in this PR itself? It's very related and makes a coherent single commit.

krisis:
>Looks like #4558 can be added in this PR itself? It's very related and makes a coherent single commit.

@donatello that makes sense. I will be adding these documentation commits to #4558.

PR #4562, opened 2017-06-19
gateway-gcs: use minio.sys.temp/multipart/v1 as url base

krishnasrinivas:
## Description

This helps when we want to use a different backend format in future.

As discussed in today's meeting we will handle the backend format this way. i.e when we move to a different format we will start using minio.sys.temp/multipart/v2 and not worry about migrating the data in v1

## Motivation and Context

To support change in backend format for multipart files.

## How Has This Been Tested?

manual

## Types of changes

## Checklist:

### DIFF
--- a/cmd/gateway-gcs.go
+++ b/cmd/gateway-gcs.go
@@ -43,7 +43,9 @@ const (
 	// ListObjects we filter out this entry.
 	gcsMinioPath = "minio.sys.temp"
 	// Path where multipart objects are saved.
-	gcsMinioMultipartPath = gcsMinioPath + "/multipart"
+	// If we change the backend format we will use a different url path like /multipart/v2
+	// but we will not migrate old data.
+	gcsMinioMultipartPathV1 = gcsMinioPath + "/multipart/v1"
 	// Multipart meta file.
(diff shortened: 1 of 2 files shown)


cmd/gateway-gcs.go
```diff
@@ -43,7 +43,9 @@ const (
 	// ListObjects we filter out this entry.
 	gcsMinioPath = "minio.sys.temp"
 	// Path where multipart objects are saved.
-	gcsMinioMultipartPath = gcsMinioPath + "/multipart"
+	// If we change the backend format we will use a different url path like /multipart/v2
+	// but we will not migrate old data.
+	gcsMinioMultipartPath = gcsMinioPath + "/multipart/v1"
```
harshavardhana: Perhaps call this `gcsMinioMultipartPathV1` ?

harshavardhana (changes requested):

cmd/gateway-gcs.go
krishnasrinivas: will do that ... to be consistent with the rest of the code base

harshavardhana (approved):
LGTM and works.

balamurugana (approved):
Not tested

PR #4576, opened 2017-06-20
Naming of return values 

fwessels:
## Description
Name return values to prevent the need (and unnecessary code bloat) to explicitly instantiate objects for every return statement.

## Motivation and Context
It improves the code and generates less code by eliminating the explicit instantiation of an object for every return statement.

## Checklist:

### DIFF
--- a/cmd/admin-rpc-client.go
+++ b/cmd/admin-rpc-client.go
@@ -113,15 +113,15 @@ func (rc remoteAdminClient) ReInitDisks() error {
 }
 
 // ServerInfoData - Returns the server info of this server.
-func (lc localAdminClient) ServerInfoData() (ServerInfoData, error) {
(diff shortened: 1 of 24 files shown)


krishnasrinivas (approved):

harshavardhana:
@balamurugana @donatello can you guys provide your feedback on this PR?

fwessels:
Pls note that this change was actually automated via the fwessels/go-diet repo (see personnel account).
The rational being that if a) return values were previously unnamed and b) a return statement instantiates an empty object (thus just the type name followed by {}), then it is completely safe to make the change. The reason being that, because the corresponding return value was unnamed, it is necessarily impossible that this value was changed, so doing the replacement for the return statement does not change the behavior in any way.

donatello (approved):

PR #4585, opened 2017-06-23
build: Error out properly when unsupported arch is given.

harshavardhana:
## Description
Release build with unsupported arch selected should error out.

## Motivation and Context
Release build cleanup.

## How Has This Been Tested?
Manually with `make release`

## Types of changes

## Checklist:

### DIFF
--- a/buildscripts/build.sh
+++ b/buildscripts/build.sh
@@ -90,9 +90,18 @@ main() {
             go_build ${each_osarch}
         done
     else
-        for each_osarch in $(echo $chosen_osarch | sed 's/,/ /g'); do
-            go_build ${each_osarch}
+        local found=0
+        for each_osarch in ${SUPPORTED_OSARCH}; do
+            if [ "$chosen_osarch" = "$each_osarch" ]; then
+                found=1
+            fi
         done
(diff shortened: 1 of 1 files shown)


donatello (approved):

buildscripts/build.sh
```diff
@@ -90,9 +90,18 @@ main() {
             go_build ${each_osarch}
         done
     else
-        for each_osarch in $(echo $chosen_osarch | sed 's/,/ /g'); do
-            go_build ${each_osarch}
+        local found=0
+        for each_osarch in ${SUPPORTED_OSARCH}; do
+            if [ "$chosen_osarch" = "$each_osarch" ]; then
```
nitisht, line 95: Probably more readable with `==` instead of `=`

nitisht (changes requested):

buildscripts/build.sh
harshavardhana: No @NitishT - test it locally `=` is also same as `==`  in bash.

nitisht: I see. Let me build locally.

nitisht (approved):
LGTM.

PR #4586, opened 2017-06-23
api: No need to set x-amz-bucket-region if region is empty.

harshavardhana:
## Description
No need to set x-amz-bucket-region if region is empty.

## Motivation and Context
Observed while testing.

## How Has This Been Tested?
Manually on a fresh minio install. 

## Types of changes

## Checklist:


krishnasrinivas (approved):
@harshavardhana approving on the assumption that aws client work fine without this response header

harshavardhana:
> @harshavardhana approving on the assumption that aws client work fine without this response header

x-amz-bucket-region is not interpreted by aws clients. Its only for our clients which make use of this for retrying a failed request.  But since default fresh installation of minio sets region as empty, it should be set.

poornas (approved):

Issue #4416, opened 2017-05-24
Implement server side encryption AES 256


deekoder:

harshavardhana:
http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingServerSideEncryption.html

An example to test it working.  Following example works with S3. 

https://github.com/minio/minio-go/blob/master/examples/s3/putobject-getobject-sse.go

vadmeste:
First prototype: a working server encryption under FS with an aws sdk client, ETA: 24 May.

vadmeste:
In server-side encryption mode, the client doesn't need to know anything to upload and download objects. It means that, if access/secret keys are compromised, all data will also be compromised. The developer can still use client side encryption to enforce his data protection strategy. 

The most secure way that I see for the moment when using server-side encryption, is to avoid storing a plain version of the access/secret keys. They can be shown to the user only the first time Minio is executed and the user has to store them safely. But that also means that if access/secret keys are lost, all encrypted data will be lost.

We probably can search for a framework that solves this problem with the best manner.

For the moment, I will continue with a master key = (access+secret) to simplify the first version of the PR.

harshavardhana:
@vadmeste there are three types of mechanisms here 

 - http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingKMSEncryption.html
 - http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingServerSideEncryption.html
 - http://docs.aws.amazon.com/AmazonS3/latest/dev/ServerSideEncryptionCustomerKeys.html

What we need to implement is SSE-C that is the last one in the list where custom provides the keys to decrypt the data. If these headers are set we are supposed to reject the operation if its not on SSL connection.  

KMSEncryption requires KMS service which is a separate API layer which we don't need to implement, SSE-S3 is also another way where the keys are managed by the server itself which we don't need to implement.

vadmeste:
yes @harshavardhana, SSE-S3 is more difficult to maintain its security. I'll stick with SSE-C as you suggested.

aead:
[...]

PR #4589, opened 2017-06-23
erasure: replace hash-based bitrot detection with generic mechanism

aead:
This change introduces the new pkg/bitrot package and include it into the minio server.

## Description
This change replace the old hash-based bitrot detection approach with a new (more generic) solution.
Now it is possible to use hash-functions or authenticators (esp. polynomial authenticators).
Therefore this change adds a new field in the XL format (checkSumInfo now has also a key).

## Motivation and Context
This is the first step to support server-side-encryption (SSE).
Updates #4416 

## How Has This Been Tested?
Add test to ensure that currently just hash-functions (or similar non-key based techniques) can be used.

## Types of changes

## Checklist:


pkg/bitrot/bitrot.go
```diff
@@ -0,0 +1,122 @@
+package bitrot
```
harshavardhana, line 18: License header needed.

cmd/xl-v1-object.go
```diff
@@ -266,7 +267,7 @@ func (xl xlObjects) GetObject(bucket, object string, startOffset int64, length i
 
 		// Get the checksums of the current part.
 		checkSums := make([]string, len(onlineDisks))
-		var ckSumAlgo HashAlgo
+		var ckSumAlgo bitrot.Algorithm = 0xffffffff
 		for index, disk := range onlineDisks {
```
harshavardhana: Any reason why choosing `0xffff` to indicate a default value . Can we use something else ?

cmd/erasure-utils_test.go
```diff
@@ -158,3 +161,21 @@ func TestCopyBuffer(t *testing.T) {
 		}
 	}
 }
+
```
harshavardhana, line 164: Add a comment on the non obvious that this functionality is testing.

harshavardhana (changes requested):

cmd/xl-v1-object.go
aead: It is an invalid value (because it much to high) Currently valid is 
 - 0: Poly1305
 - 1: GHASH
 - 2: BLAKE2b512
 - 3: SHA-256

Your right, that should be changed - Maybe making [unknownAlgorithm](https://github.com/aead/minio/blob/erasure/pkg/bitrot/bitrot.go#L27) visible? Then it should be clear...

pkg/bitrot/bitrot.go
aead: Done

cmd/erasure-utils_test.go
aead: Done

cmd/xl-v1-object.go
harshavardhana: Yes it is better that way @aead

cmd/posix.go
```diff
@@ -521,8 +521,7 @@ func (s *posix) ReadAll(volume, path string) (buf []byte, err error) {
 // Additionally ReadFile also starts reading from an offset. ReadFile
 // semantics are same as io.ReadFull.
 func (s *posix) ReadFile(volume, path string, offset int64, buf []byte) (n int64, err error) {
-
-	return s.ReadFileWithVerify(volume, path, offset, buf, "", "")
+	return s.ReadFileWithVerify(volume, path, offset, buf, 0xfffffff, "")
```
fwessels: Why are we passing (ugly) constant 0xffffffff (obviously -1) in ?
Make a proper const definition?

cmd/posix_test.go
```diff
@@ -1062,44 +1064,44 @@ func TestPosixReadFileWithVerify(t *testing.T) {
...
 		// Hash verification success with partial content and Sha256 - 6
 		{
-			"myobject", 7, 4, HashSha256, sha256Hash("Hello, world!"),
+			"myobject", 7, 4, bitrot.SHA256, sha256Hash("Hello, world!"),
 			[]byte("worl"), nil,
 		},
 		// Empty hash-algo returns error - 7
 		{
-			"myobject", 7, 4, "", blakeHash("Hello, world!"),
+			"myobject", 7, 4, 0xffffffff, blakeHash("Hello, world!"),
```
fwessels: Same here (and in line 1104)

pkg/bitrot/bitrot.go
```diff
@@ -0,0 +1,141 @@
...
+	SHA256:     "sha256",
+}
+
+var hashes = make([]func([]byte) Hash, UnknownAlgorithm)
+
+// New returns a new Hash implementing the given algorithm.
+// It returns an error if the given key cannot be used by the algorithm.
+// New panics if the algorithm is unknown.
+func (a Algorithm) New(key []byte) (Hash, error) {
+	if keysize := a.KeySize(); keysize != len(key) {
```
fwessels: Why not simply do `if a.KeySize() != len(key) {` etc.

pkg/bitrot/bitrot.go
```diff
@@ -0,0 +1,141 @@
...
+	if f != nil {
+		return f(key), nil
+	}
+	panic("bitrot: requested algorithm  #" + strconv.Itoa(int(a)) + " is not available")
+}
+
+// KeySize returns the size of valid keys for the given algorithm in bytes.
+// It panics if the algorithm is unknown.
+func (a Algorithm) KeySize() int {
+	if a >= 0 && a < UnknownAlgorithm {
```
fwessels: a is a `uint` so `a >= 0` is a NOP.
Also you can better test against len(keysizes) -- in case somebody would add it to the const array but forget to add it to the keysizes array.

pkg/bitrot/bitrot.go
```diff
@@ -0,0 +1,141 @@
...
+
+// Available reports whether the given algorithm is registered.
+func (a Algorithm) Available() bool {
+	return a < UnknownAlgorithm && hashes[a] != nil
+}
+
+// String returns a string representation of the algorithm.
+// It panics if the algorithm is unknown.
+func (a Algorithm) String() string {
+	if a >= 0 && a < UnknownAlgorithm {
```
fwessels: Idem

pkg/bitrot/bitrot.go
```diff
@@ -0,0 +1,141 @@
...
+
+// AlgorithmFromString returns an algorithm based on the provided string representation.
+// It returns an error if the given string does not match a known algorithm.
+func AlgorithmFromString(s string) (Algorithm, error) {
+	for i, v := range names {
+		if s == v {
+			return Algorithm(i), nil
+		}
+	}
+	return UnknownAlgorithm, fmt.Errorf("bitrot: algorithm #%s is unknown", s)
```
fwessels: `#%s` --> `%s` as s is not a number

vendor/github.com/aead/poly/poly.go
```diff
@@ -0,0 +1,28 @@
+package poly
```
fwessels, line 5: May consider adding license header

vendor/github.com/aead/poly1305/poly1305_amd64.go
```diff
@@ -0,0 +1,108 @@
...
+			p.off += copy(p.buf[p.off:], msg)
+			return n, nil
+		}
+		copy(p.buf[p.off:], msg[:dif])
+		update(&(p.state), p.buf[:])
+		msg = msg[dif:]
+		p.off = 0
+	}
+
+	if nn := len(msg) & (^(TagSize - 1)); nn > 0 {
```
fwessels: Add comment why you are rounding up to multiple of TagSize here?

vendor/github.com/aead/poly1305/poly1305_amd64.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	if len(msg) > 0 {
+		p.off += copy(p.buf[p.off:], msg)
+	}
+
+	return n, nil
+}
+
+// Sum appends the Pol1305 hash of the previously
```
fwessels: Pol1305 --> Poly1305

vendor/github.com/aead/poly1305/poly1305_amd64.go
```diff
@@ -0,0 +1,108 @@
...
+// processed data to b and returns the resulting slice.
+// It is safe to call this function multiple times.
+func (p *Hash) Sum(b []byte) []byte {
+	var out [TagSize]byte
+	state := p.state
+
+	if p.off > 0 {
+		update(&state, p.buf[:p.off])
```
fwessels, line 103: reset p.off to 0 here?

vendor/github.com/aead/poly1305/poly1305_amd64.s
```diff
@@ -0,0 +1,48 @@
...
+DATA ·poly1305Mask<>+0x08(SB)/8, $0x0FFFFFFC0FFFFFFC
+GLOBL ·poly1305Mask<>(SB), RODATA, $16
+
+// func initialize(state *[7]uint64, key *[32]byte)
+TEXT ·initialize(SB), $0-16
+	MOVQ state+0(FP), DI
+	MOVQ key+8(FP), SI
+
+
+	// state[0...7] is initialized with zero
```
fwessels: This comment is not correct -- or do you mean: state[0...7] has been initialized with zero
(previously)

vendor/github.com/aead/poly1305/poly1305_ref.go
```diff
@@ -0,0 +1,228 @@
...
+
+	initialize(&r, &s, &key)
+
+	n := len(msg) & (^(TagSize - 1))
+	if n > 0 {
+		update(msg[:n], msgBlock, &h, &r)
+		msg = msg[n:]
+	}
+	if len(msg) > 0 {
+		off += copy(block[:], msg)
```
fwessels: `off +=` is equal to `off =` ?

vendor/github.com/aead/poly1305/poly1305_ref.go
```diff
@@ -0,0 +1,228 @@
...
+)
+
+// Sum generates an authenticator for msg using a one-time key and returns the
+// 16-byte result. Authenticating two different messages with the same key allows
+// an attacker to forge messages at will.
+func Sum(msg []byte, key [32]byte) [TagSize]byte {
+	var (
+		h, r  [5]uint32
+		s     [4]uint32
+		block [TagSize]byte
```
fwessels: Move `block` and `off` inside of `if` statement in line 35?

vendor/github.com/aead/poly1305/poly1305_ref.go
```diff
@@ -0,0 +1,228 @@
...
+// Notice that Poly1305 is insecure if one key is used twice.
+func New(key [32]byte) *Hash {
+	p := new(Hash)
+	initialize(&(p.r), &(p.s), &key)
+	return p
+}
+
+// Hash implements a Poly1305 writer interface.
+// Poly1305 cannot be used like common hash.Hash implementations,
+// because of using a poly1305 key twice breaks its security.
```
fwessels: because of using --> because using

vendor/github.com/aead/poly1305/poly1305_ref.go
```diff
@@ -0,0 +1,228 @@
...
+func (p *Hash) Sum(b []byte) []byte {
+	var out [TagSize]byte
+	h := p.h
+
+	if p.off > 0 {
+		var buf [TagSize]byte
+		copy(buf[:], p.buf[:p.off])
+		buf[p.off] = 1 // invariant: p.off < TagSize
+
+		update(buf[:], finalBlock, &h, &(p.r))
```
fwessels, line 115: reset p.off to 0 ?

vendor/github.com/aead/poly1305/poly1305_go17_amd64.s
```diff
@@ -0,0 +1,197 @@
...
+
+DONE:
+	MOVQ R8, 0(DI)
+	MOVQ R9, 8(DI)
+	MOVQ R10, 16(DI)
+	RET
+
+// func supportsBMI2() bool
+TEXT ·supportsBMI2(SB), 4, $0-1
+    MOVQ runtime·support_avx2(SB), BX
```
fwessels: Formatting

cmd/xl-v1-object.go
```diff
@@ -266,7 +267,7 @@ func (xl xlObjects) GetObject(bucket, object string, startOffset int64, length i
 
 		// Get the checksums of the current part.
 		checkSums := make([]string, len(onlineDisks))
-		var ckSumAlgo HashAlgo
+		var ckSumAlgo bitrot.Algorithm = bitrot.UnknownAlgorithm
```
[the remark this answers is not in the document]
aead: Done

cmd/posix.go
aead: Same as Harsha already said, good catch! Done.

cmd/posix_test.go
aead: Done

pkg/bitrot/bitrot.go
aead: Done

pkg/bitrot/bitrot.go
aead: Done

pkg/bitrot/bitrot.go
aead: Done

pkg/bitrot/bitrot.go
aead: Done

vendor/github.com/aead/poly/poly.go
aead: Oh, yes, thank you!

vendor/github.com/aead/poly1305/poly1305_amd64.go
aead: Maybe a good idea :wink:

vendor/github.com/aead/poly1305/poly1305_amd64.go
aead: Done

vendor/github.com/aead/poly1305/poly1305_amd64.go
aead: No, that would break the commented guarantee:
> // It is safe to call this function multiple times.

If we set `p.off = 0` the second call to Sum would return a different hash. That's the reason why we copy the state (see> `state := p.state`)

vendor/github.com/aead/poly1305/poly1305_amd64.s
aead: Yes, I mean that it is initialized with zeros when `initialize` is invoked (golang memory rule).
Otherwise a `PXOR X0, X0` and `PXOR X1, X1` would be necessary.

vendor/github.com/aead/poly1305/poly1305_ref.go
aead: Yes, it should... off is zero - I change this to `off = copy...`

vendor/github.com/aead/poly1305/poly1305_ref.go
aead: Can be done, is this better? Does the compiler not alloc the memory if it is bound to the if scope?

vendor/github.com/aead/poly1305/poly1305_ref.go
aead: Done, thanks.

vendor/github.com/aead/poly1305/poly1305_ref.go
aead: Same here: // It is safe to call this function multiple times.

vendor/github.com/aead/poly1305/poly1305_go17_amd64.s
aead: Done

pkg/bitrot/bitrot.go
```diff
@@ -0,0 +1,141 @@
...
+
+// Available reports whether the given algorithm is registered.
+func (a Algorithm) Available() bool {
+	return a < UnknownAlgorithm && hashes[a] != nil
+}
+
+// String returns a string representation of the algorithm.
+// It panics if the algorithm is unknown.
+func (a Algorithm) String() string {
+	if a < UnknownAlgorithm && int(a) < len(keysizes) {
```
fwessels: this should be `len(names)`

vendor/github.com/aead/poly1305/poly1305_amd64.go
fwessels: Ok, clear.

vendor/github.com/aead/poly1305/poly1305_amd64.s
fwessels: Ok

vendor/github.com/aead/poly1305/poly1305_ref.go
fwessels: It is not necessarily better (compiler-wise), maybe a little neater to allocate more closely to where you need it (and bound to scope).

aead:
Sry for the additional changes, but I found a conceptional problem which required a change to the API.
But now it should fulfill minio's requirements.

pkg/bitrot/bitrot.go
```diff
@@ -23,6 +23,16 @@ import (
 	"strconv"
 )
 
+// Mode specifies how the Hash will be used - either for protection or for verification.
+type Mode bool
+
+const (
+	// Protect indicates that the Hash is used for computing a checksum.
+	Protect Mode = true
+	// Verify indicates that the Hash is used for verifying a previous computed checksum.
```
fwessels: previous --> previously

vendor/github.com/aead/poly1305/poly1305_ref.go
```diff
@@ -34,7 +32,8 @@ func Sum(msg []byte, key [32]byte) [TagSize]byte {
 		msg = msg[n:]
 	}
 	if len(msg) > 0 {
-		off = copy(block[:], msg)
+		var block [TagSize]byte
```
fwessels: If len(msg) is a multiple of 16 bytes (and this the remainder = 0, there is no update() using the finalBlock?
Is that correct behaviour?

vendor/github.com/aead/poly1305/poly1305_ref.go
```diff
@@ -34,7 +32,8 @@ func Sum(msg []byte, key [32]byte) [TagSize]byte {
...
+		off := copy(block[:], msg)
```
fwessels: if len(msg) ever is TagSize, then the following statement `block[off] = 1` would cause a buffer overrun -- may want to protect against this?

aead: No, I don't think so: If `len(msg) == TagSize` than `len(msg) & (^(TagSize - 1)) == 16` 
I also ran:
```
func TestOverflow(t *testing.T) {
	msg := make([]byte, TagSize)
	var key [32]byte
	Sum(msg, key)
}
with 'export GOARCH=386' and 'go run -v -run=TestOverflow'

vendor/github.com/aead/poly1305/poly1305_ref.go
aead: Yes, this is intended behavior.

pkg/bitrot/bitrot.go
aead: Done

aead:
Closed: We use a separate branch for this. See: [encryption branch](https://github.com/minio/minio/tree/encryption)

PR #4599, opened 2017-06-26
gateway-gcs: remove files older than 2 weeks in minio.sys.temp

krishnasrinivas:
## Description

Every 12 hours we check minio.sys.temp of all the buckets and delete files that are older than 2 weeks.

## How Has This Been Tested?

manually (by changing 1 week to 1 minute and see that it indeed purges the files)

## Types of changes

## Checklist:

### DIFF
--- a/cmd/gateway-gcs.go
+++ b/cmd/gateway-gcs.go
@@ -295,10 +295,16 @@ func newGCSGateway(projectID string) (GatewayLayer, error) {
 func (l *gcsGateway) CleanupGCSMinioPathBucket(bucket string) {
 	it := l.client.Bucket(bucket).Objects(l.ctx, &storage.Query{Prefix: gcsMinioTempPath, Versions: false})
 	for {
+		select {
+		case <-l.cleanupEndCh:
+			// Stop cleanup process if Shutdown() was called.
+			return
+		default:
+		}
 		attrs, err := it.Next()
 		if err != nil {
 			if err != iterator.Done {
-				errorIf(err, "Object listing error on bucket %s", bucket)
+				errorIf(err, "Object listing error on bucket %s during purging of old files in minio.sys.temp", bucket)
 			}
 			return
 		}
@@ -306,7 +312,7 @@ func (l *gcsGateway) CleanupGCSMinioPathBucket(bucket string) {
 			// Delete files older than 2 weeks.
 			err := l.client.Bucket(bucket).Object(attrs.Name).Delete(l.ctx)
 			if err != nil {
-				errorIf(err, "Unable to delete the object %s", attrs.Name)
+				errorIf(err, "Unable to delete the object %s during purging of old files in minio.sys.temp", attrs.Name)
 				return
 			}
 		}
@@ -318,10 +324,16 @@ func (l *gcsGateway) CleanupGCSMinioPath() {
 	for {
 		it := l.client.Buckets(l.ctx, l.projectID)
 		for {
+			select {
+			case <-l.cleanupEndCh:
+				// Stop cleanup process if Shutdown() was called.
+				return
+			default:
+			}
 			attrs, err := it.Next()
 			if err != nil {
 				if err != iterator.Done {
-					errorIf(err, "Bucket listing error")
+					errorIf(err, "Bucket listing error during purging of old files in minio.sys.temp")
 				}
 				break
 			}
@@ -329,7 +341,7 @@ func (l *gcsGateway) CleanupGCSMinioPath() {
 		}
 		select {
 		case <-l.cleanupEndCh:
-			// If shutdown is called we end the cleanup process.
+			// Stop cleanup process if Shutdown() was called.
 			return
 		case <-time.After(gcsCleanupInterval):
 		}
--- a/cmd/gateway-gcs.go
+++ b/cmd/gateway-gcs.go
@@ -69,7 +69,7 @@ const (
 	// gcsMaxPartCount - maximum multipart parts GCS supports which is 32 x 32 = 1024.
 	gcsMaxPartCount = 1024
 
-	// Every 12 hours we scan minio.sys.temp to delete expired multiparts.
+	// Every 24 hours we scan minio.sys.temp to delete expired multiparts.
 	gcsCleanupInterval = time.Hour * 24
 
 	// Expiry time after which the multipart gets deleted.
@@ -249,11 +249,10 @@ func checkGCSProjectID(ctx context.Context, projectID string) error {
 
 // gcsGateway - Implements gateway for Minio and GCS compatible object storage servers.
 type gcsGateway struct {
-	client       *storage.Client
-	anonClient   *minio.Core
-	projectID    string
-	cleanupEndCh chan struct{}
-	ctx          context.Context
+	client     *storage.Client
+	anonClient *minio.Core
+	projectID  string
+	ctx        context.Context
 }
 
 const googleStorageEndpoint = "storage.googleapis.com"
@@ -280,11 +279,10 @@ func newGCSGateway(projectID string) (GatewayLayer, error) {
 	}
 
 	gateway := &gcsGateway{
-		client:       client,
-		projectID:    projectID,
-		ctx:          ctx,
-		anonClient:   anonClient,
-		cleanupEndCh: make(chan struct{}),
+		client:     client,
+		projectID:  projectID,
+		ctx:        ctx,
+		anonClient: anonClient,
 	}
 	// Start background process to cleanup old files in minio.sys.temp
 	go gateway.CleanupGCSMinioPath()
@@ -295,12 +293,6 @@ func newGCSGateway(projectID string) (GatewayLayer, error) {
 func (l *gcsGateway) CleanupGCSMinioPathBucket(bucket string) {
 	it := l.client.Bucket(bucket).Objects(l.ctx, &storage.Query{Prefix: gcsMinioTempPath, Versions: false})
 	for {
-		select {
-		case <-l.cleanupEndCh:
-			// Stop cleanup process if Shutdown() was called.
-			return
-		default:
-		}
 		attrs, err := it.Next()
 		if err != nil {
 			if err != iterator.Done {
@@ -312,7 +304,7 @@ func (l *gcsGateway) CleanupGCSMinioPathBucket(bucket string) {
 			// Delete files older than 2 weeks.
 			err := l.client.Bucket(bucket).Object(attrs.Name).Delete(l.ctx)
 			if err != nil {
-				errorIf(err, "Unable to delete the object %s during purging of old files in minio.sys.temp", attrs.Name)
+				errorIf(err, "Unable to delete %s/%s during purging of old files in minio.sys.temp", bucket, attrs.Name)
 				return
 			}
 		}
@@ -324,12 +316,6 @@ func (l *gcsGateway) CleanupGCSMinioPath() {
 	for {
 		it := l.client.Buckets(l.ctx, l.projectID)
 		for {
-			select {
-			case <-l.cleanupEndCh:
-				// Stop cleanup process if Shutdown() was called.
-				return
-			default:
-			}
 			attrs, err := it.Next()
 			if err != nil {
 				if err != iterator.Done {
@@ -339,19 +325,14 @@ func (l *gcsGateway) CleanupGCSMinioPath() {
 			}
 			l.CleanupGCSMinioPathBucket(attrs.Name)
 		}
-		select {
-		case <-l.cleanupEndCh:
-			// Stop cleanup process if Shutdown() was called.
-			return
-		case <-time.After(gcsCleanupInterval):
-		}
+		// Run the cleanup loop every 1 day.
+		time.Sleep(gcsCleanupInterval)
 	}
 }
 
 // Shutdown - save any gateway metadata to disk
 // if necessary and reload upon next restart.
 func (l *gcsGateway) Shutdown() error {
-	l.cleanupEndCh <- struct{}{}
 	return nil
 }
 
--- a/cmd/gateway-gcs.go
+++ b/cmd/gateway-gcs.go
@@ -45,12 +45,12 @@ const (
 	// gcsMinioMeta is used for multiparts. We have "minio.sys.temp" prefix so that
 	// listing on the GCS lists this entry in the end. Also in the gateway
 	// ListObjects we filter out this entry.
-	gcsMinioPath = "minio.sys.temp/"
+	gcsMinioTempPath = "minio.sys.temp/"
 
 	// Path where multipart objects are saved.
 	// If we change the backend format we will use a different url path like /multipart/v2
 	// but we will not migrate old data.
-	gcsMinioMultipartPathV1 = gcsMinioPath + "multipart/v1"
+	gcsMinioMultipartPathV1 = gcsMinioTempPath + "multipart/v1"
 
 	// Multipart meta file.
 	gcsMinioMultipartMeta = "gcs.json"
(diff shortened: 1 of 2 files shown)


cmd/gateway-gcs.go
```diff
@@ -260,12 +260,49 @@ func newGCSGateway(projectID string) (GatewayLayer, error) {
...
+	for {
+		it := l.client.Buckets(l.ctx, l.projectID)
+		for {
+			attrs, err := it.Next()
+			if err != nil {
+				break
+			}
+			l.CleanupGCSMinioPathBucket(attrs.Name)
+		}
+		time.Sleep(time.Hour * 12)
```
harshavardhana: Make `12Hour` a constant.

cmd/gateway-gcs.go
```diff
@@ -260,12 +260,49 @@ func newGCSGateway(projectID string) (GatewayLayer, error) {
...
+
+// Cleanup old files in minio.sys.temp of the given bucket.
+func (l *gcsGateway) CleanupGCSMinioPathBucket(bucket string) {
+	it := l.client.Bucket(bucket).Objects(l.ctx, &storage.Query{Prefix: gcsMinioPath, Versions: false})
+	for {
+		attrs, err := it.Next()
+		if err != nil {
+			return
+		}
+		if time.Since(attrs.Updated) > time.Hour*24*7 {
```
harshavardhana: Make 7 days constant. Have a select loop so that when `serviceDoneCh` is called, this can be cleanly shutdown.

harshavardhana (changes requested):

deekoder:
@krishnasrinivas does this pr fix a specific issue?

cmd/gateway-gcs.go
```diff
@@ -260,12 +260,48 @@ func newGCSGateway(projectID string) (GatewayLayer, error) {
...
+}
+
+// Cleanup old files in minio.sys.temp of all buckets.
+func (l *gcsGateway) CleanupGCSMinioPath() {
+	for {
+		it := l.client.Buckets(l.ctx, l.projectID)
+		for {
+			attrs, err := it.Next()
+			if err != nil {
+				break
```
donatello: No error logging?

cmd/gateway-gcs.go
```diff
@@ -260,12 +260,48 @@ func newGCSGateway(projectID string) (GatewayLayer, error) {
...
+	for {
+		attrs, err := it.Next()
+		if err != nil {
+			return
+		}
+		if time.Since(attrs.Updated) > time.Hour*24*7 {
+			// Delete files older than 1 week.
+			err := l.client.Bucket(bucket).Object(attrs.Name).Delete(l.ctx)
+			if err != nil {
+				return
```
donatello: Should log error here so at least it is known what is going wrong.

cmd/gateway-gcs.go
```diff
@@ -260,12 +260,48 @@ func newGCSGateway(projectID string) (GatewayLayer, error) {
...
+	return gateway, nil
+}
+
+// Cleanup old files in minio.sys.temp of the given bucket.
+func (l *gcsGateway) CleanupGCSMinioPathBucket(bucket string) {
+	it := l.client.Bucket(bucket).Objects(l.ctx, &storage.Query{Prefix: gcsMinioPath, Versions: false})
+	for {
+		attrs, err := it.Next()
+		if err != nil {
+			return
```
donatello: Error log?

donatello (changes requested):

krishnasrinivas:
@harshavardhana @donatello please check now

@deekoder no, do you want one to be created?

cmd/gateway-gcs.go
```diff
@@ -268,17 +279,67 @@ func newGCSGateway(projectID string) (GatewayLayer, error) {
...
+			if err != iterator.Done {
+				errorIf(err, "Object listing error on bucket %s", bucket)
+			}
+			return
+		}
+		if time.Since(attrs.Updated) > gcsMultipartExpiry {
+			// Delete files older than 1 week.
+			err := l.client.Bucket(bucket).Object(attrs.Name).Delete(l.ctx)
+			if err != nil {
+				errorIf(err, "Unable to delete the object %", attrs.Name)
```
donatello: `s/%/%s/`

cmd/gateway-gcs.go
```diff
@@ -268,17 +279,67 @@ func newGCSGateway(projectID string) (GatewayLayer, error) {
...
+			return
+		case <-time.After(gcsCleanupInterval):
+		}
+	}
 }
 
 // Shutdown - save any gateway metadata to disk
 // if necessary and reload upon next restart.
 func (l *gcsGateway) Shutdown() error {
+	close(l.cleanupEndCh)
```
donatello: The purpose here seems to be to make the cleanup go-routine quit, however, if it is in the middle of scanning buckets it will fail to quit immediately, and in the meanwhile the rest of the process will quit. So it does not seem to be needed.

donatello (changes requested):

harshavardhana (changes requested):
This PR should be updated to handle `.minio.sys.temp/tmp` path as well for cleanup.

krishnasrinivas:
@harshavardhana i see that the patch already does this:
```
it := l.client.Bucket(bucket).Objects(l.ctx, &storage.Query{Prefix: gcsMinioPath, Versions: false})
```

harshavardhana (approved):

donatello (commented):
@krishnasrinivas waiting on you to address my previous comments.

krishnasrinivas:
@donatello done.

Assigning @krisis as reviewer as @donatello is travelling.

cmd/gateway-gcs.go
```diff
@@ -60,20 +60,20 @@ const (
...
+	// gcsMaxPartCount - maximum multipart parts GCS supports which is 32 x 32 = 1024.
+	gcsMaxPartCount = 1024
 
 	// Every 12 hours we scan minio.sys.temp to delete expired multiparts.
-	gcsCleanupInterval = time.Hour * 12
+	gcsCleanupInterval = time.Hour * 24
 
 	// Expiry time after which the multipart gets deleted.
-	gcsMultipartExpiry = time.Hour * 24 * 7
+	gcsMultipartExpiry = time.Hour * 24 * 14
```
harshavardhana: So it's 14 days now why an arbitrary extra week?

cmd/gateway-gcs.go
```diff
@@ -268,17 +279,67 @@ func newGCSGateway(projectID string) (GatewayLayer, error) {
...
+
+// Cleanup old files in minio.sys.temp of all buckets.
+func (l *gcsGateway) CleanupGCSMinioPath() {
+	for {
+		it := l.client.Buckets(l.ctx, l.projectID)
+		for {
+			attrs, err := it.Next()
+			if err != nil {
+				if err != iterator.Done {
+					errorIf(err, "Bucket listing error")
```
krisis: The error message is very generic. This bucket listing error pertains to cleaning up of multipart uploads older than the expiry time. We could log something like,
"Couldn't list buckets while trying to cleanup object parts uploaded a week ago".

cmd/gateway-gcs.go
```diff
@@ -268,17 +279,67 @@ func newGCSGateway(projectID string) (GatewayLayer, error) {
...
+			return
+		case <-time.After(gcsCleanupInterval):
+		}
+	}
 }
 
 // Shutdown - save any gateway metadata to disk
 // if necessary and reload upon next restart.
 func (l *gcsGateway) Shutdown() error {
+	l.cleanupEndCh <- struct{}{}
```
krisis: Isn't `close(ch)` the idiomatic way to signal "done" from the sender's side?

cmd/gateway-gcs.go
```diff
@@ -268,17 +279,67 @@ func newGCSGateway(projectID string) (GatewayLayer, error) {
...
+			if err != nil {
+				if err != iterator.Done {
+					errorIf(err, "Bucket listing error")
+				}
+				break
+			}
+			l.CleanupGCSMinioPathBucket(attrs.Name)
+		}
+		select {
+		case <-l.cleanupEndCh:
```
krisis: Is it OK to block shutdown of object server until cleanup is complete?

krisis (commented):
Not tested yet, but here are some initial comments.

cmd/gateway-gcs.go
krishnasrinivas: I can change it, sure.

cmd/gateway-gcs.go
krishnasrinivas: It was close() previously, have a look at aditya's comment on why it was changed to channel write

cmd/gateway-gcs.go
krishnasrinivas: It was close() previously, have a look at aditya's comment on why it was changed to channel write

krisis:
@krishnasrinivas PR description says uploaded parts of an incomplete upload older than 1week will be cleaned up but `gcsMultipartExpiry` is 2 weeks (`time.Hour * 24 * 14`). Which one is the intended value?

cmd/gateway-gcs.go
krishnasrinivas: We can let the cleanup complete. Cleanup will be a quick process as the number of entries in minio.sys.temp will be less.

krishnasrinivas:
> @krishnasrinivas PR description says uploaded parts of an incomplete upload older than 1week will be cleaned up but gcsMultipartExpiry is 2 weeks (time.Hour * 24 * 14). Which one is the intended value?

Initially it was 1 week, it was changed to 2 weeks today, will update the commit message

cmd/gateway-gcs.go
krisis: Quoting @donatello,
> The purpose here seems to be to make the cleanup go-routine quit, however, if it is in the middle of scanning buckets it will fail to quit immediately, and in the meanwhile the rest of the process will quit. So it does not seem to be needed

IIUC, he says closing the channel is not needed. 

Sending a value over a unbuffered channel will block the go-routine calling Shutdown until the go-routine running cleanup receives from the CleanupEndCh. Is it important for the go-routine calling Shutdown to block until the cleanup routine receives the "done" signal? If so, why is this behaviour required?

cmd/gateway-gcs.go
krisis: There will be one entry for every part, for all the uploads under a given bucket.
E.g, If there are 3 uploads with each 100 parts uploaded like `bucket/obj1`, `bucket/obj2` and `bucket/obj3`, there would be 303 entries (incl. `gcs.json` for every upload) under `bucket`. I.e, for `N` uploads of objects in a given bucket with `K` parts in each, the total number of entries to be deleted in the bucket would be, `N * (K+1)`. Further, in the GCS go SDK we use, there is no provision to delete multiple objects in a single API request (unlike AWS), we will send API requests equal to number of internal objects to be deleted.

For this process to be quick, one or more of the following should be true,
1. There are very few incomplete uploads during cleanup.
2. There are very few parts in an upload.

I am OK with the current blocking behaviour if you know that the above conditions are likely to be met.

cmd/gateway-gcs.go
krishnasrinivas: > Is it important for the go-routine calling Shutdown to block until the cleanup routine receives the "done" signal? If so, why is this behaviour required?

It is important, else `mc admin shutdown` tells the user that minio has shutdown but it will still be doing cleanup activity.

cmd/gateway-gcs.go
krishnasrinivas: Check now, the cleanup go-routine returns as soon as it can.

cmd/gateway-gcs.go
```diff
@@ -268,17 +279,79 @@ func newGCSGateway(projectID string) (GatewayLayer, error) {
...
+	go gateway.CleanupGCSMinioPath()
+	return gateway, nil
+}
+
+// Cleanup old files in minio.sys.temp of the given bucket.
+func (l *gcsGateway) CleanupGCSMinioPathBucket(bucket string) {
+	it := l.client.Bucket(bucket).Objects(l.ctx, &storage.Query{Prefix: gcsMinioTempPath, Versions: false})
+	for {
+		select {
+		case <-l.cleanupEndCh:
```
donatello: Once the `struct{}{}` send in the Shutdown() code happens, if it is consumed on this line, this cleanup go-routine is going to get stuck in the the select block containing `case <-time.After(gcsCleanupInterval):`. This is a serious bug that will stop the shutdown of the server.

cmd/gateway-gcs.go
```diff
@@ -268,17 +279,79 @@ func newGCSGateway(projectID string) (GatewayLayer, error) {
...
+			if err != iterator.Done {
+				errorIf(err, "Object listing error on bucket %s during purging of old files in minio.sys.temp", bucket)
+			}
+			return
+		}
+		if time.Since(attrs.Updated) > gcsMultipartExpiry {
+			// Delete files older than 2 weeks.
+			err := l.client.Bucket(bucket).Object(attrs.Name).Delete(l.ctx)
+			if err != nil {
+				errorIf(err, "Unable to delete the object %s during purging of old files in minio.sys.temp", attrs.Name)
```
donatello: Need to include bucket name also in the message - object alone is not complete info.

cmd/gateway-gcs.go
```diff
@@ -44,26 +45,35 @@ const (
...
-	maxComponents = 32
+	gcsMaxComponents = 32
+
+	// gcsMaxPartCount - maximum multipart parts GCS supports which is 32 x 32 = 1024.
+	gcsMaxPartCount = 1024
 
-	// maxPartCount - maximum multipart parts GCS supports which is 32 x 32 = 1024.
-	maxPartCount = 1024
+	// Every 12 hours we scan minio.sys.temp to delete expired multiparts.
+	gcsCleanupInterval = time.Hour * 24
```
donatello: Comment and value are not matching.

donatello (changes requested):
I think I found a serious issue. Please see inline comment.

Other than that this patch has too many select blocks doing the same things to try and exit the go-routine cleanly and quickly in case of shutdown. However, they do not really work - all network calls must be treated as indefinitely blocking, and once one of them is entered, there is no guarantee when the cleanup routine will exit.

Doing this correctly, requires use of the context objects that the gcs package is making use of. We could discuss how it would work over a call if you prefer.

cmd/gateway-gcs.go
krishnasrinivas: AB asked me to change it to 14 days.

krishnasrinivas:
@donatello as we need to take in the cleanup-go-routine for today's build I have removed the handling of Shutdown() as it needs more work (and discussion). It can be fixed in a later PR as it is not critical. We should fix it before we support `mc admin shutdown` though.

deekoder:
Tested it on gcs and it worked for me.

krishnasrinivas:
@abperiasamy approved, we can take this in @deekoder

harshavardhana:
There were no approvals on this ? Why was it merged ?

harshavardhana:
Also merged without proper squashing this is the current commit message now.

```
* gateway-gcs: remove files older than 2 weeks in minio.sys.temp

* Review fixes

* Fix review comments

* Fix AB's review comments

* KP's comments

* Remove termination of cleanup-go-routine on Shutdown(). To be fixed in a separate PR later.

* Rename minio.sys.temp to minio.sys.tmp
Rename ##minio## to {minio}

* Fix test case
```

harshavardhana:
We need to fix this commit message ASAP before we merge more PRs.

donatello:
@harshavardhana agree with you, but I think since it is a protected branch we won't be able push force to it. Might have to do it yourself.

Also an issue needs to be opened for the cleanup at shutdown part that was not addressed in this PR @krishnasrinivas.

krishnasrinivas:
> @abperiasamy approved, we can take this in @deekoder

> There were no approvals on this ? Why was it merged ?

@harshavardhana @abperiasamy reviewed and approved this and said this could be merged, he was in a meeting and hence could not approve on github

PR #4609, opened 2017-06-29
Added new optimized images.

deekoder:
Added new images for docs. Will remove old ones after changing markdowns.


nitisht (approved):

harshavardhana (commented):
LGTM. 

On a side note we should avoid creating branches on the main repo @deekoder unless they are large changes like

- feature-gcs
- saml

Create a branch on your fork and then submit a PR instead.

harshavardhana (approved):

Issue #4617, opened 2017-07-02
giving 4 backing storage devices with differing free space results in wrong capacity


niv:
Provide 4 backing disk images/directories with differing sizes:
* 1GB
* 2GB
* +2GB (any)
* +2GB (any)

Example setup:
```
Initializing data volume for the first time.
[01/04] /Volumes/minio - 954 MiB online
[02/04] /Volumes/miniobig - 1.7 GiB online
[03/04] a - 465 GiB online
[04/04] b - 465 GiB online

Endpoint:  http://10.47.1.113:9000  http://10.2.1.107:9000  http://127.0.0.1:9000
AccessKey: access
SecretKey: secretsecret

Browser Access:
   http://10.47.1.113:9000  http://10.2.1.107:9000  http://127.0.0.1:9000

Command-line Access: https://docs.minio.io/docs/minio-client-quickstart-guide
   $ mc config host add myminio http://10.47.1.113:9000 access secretsecret

Object API (Amazon S3 compatible):
   Go:         https://docs.minio.io/docs/golang-client-quickstart-guide
   Java:       https://docs.minio.io/docs/java-client-quickstart-guide
   Python:     https://docs.minio.io/docs/python-client-quickstart-guide
   JavaScript: https://docs.minio.io/docs/javascript-client-quickstart-guide
   .NET:       https://docs.minio.io/docs/dotnet-client-quickstart-guide

Drive Capacity: 1.8 GiB Free, 1.9 GiB Total
Status:         4 Online, 0 Offline. We can withstand [2] more drive failure(s).
```

## Expected Behavior

* Capacity should say 1GB, as that is the smallest "disk" available to minio.
* Uploading a >1GB file should fail immediately.

## Current Behavior

* Capacity says 2GB.
* Uploading a file litters the file system with part.XX files. No UI feedback. File does not show up in the web interface. No error on stdout.

## Your Environment
  
* Version used (`minio version`): 
Version: 2017-06-13T19:01:01Z
Release-Tag: RELEASE.2017-06-13T19-01-01Z
Commit-ID: b9f622824ac17d3e0577d8a9b81a707666cc8cf1

Test setup on OSX Sierra, installed via homebrew.

harshavardhana:
Erasure behaves in this manner N disks are divided into N/2 data and N/2 parity lets take an example homogeneous disk sizes, each disk is 1TB on a 4 disk setup would effectively result to 2TB.  So the algorithm of calculating the effective disk size is simply 

```
[...]

PR #4618, opened 2017-07-02
posix: Check for min disk space and inodes

harshavardhana:
## Description
This is needed such that we don't start or
allow writing to a posix disk which doesn't
have minimum disk free available.

## Motivation and Context
One part fix for #4617

## How Has This Been Tested?
Manually using a disk of size smaller than 1GiB

## Types of changes

## Checklist:

### DIFF
--- a/cmd/fs-v1.go
+++ b/cmd/fs-v1.go
@@ -103,6 +103,16 @@ func newFSObjectLayer(fsPath string) (ObjectLayer, error) {
 		}
 	}
 
+	di, err := getDiskInfo(preparePath(fsPath))
+	if err != nil {
+		return nil, err
+	}
+
+	// Check if disk has minimum required total space.
+	if err = checkDiskMinTotal(di); err != nil {
+		return nil, err
+	}
+
 	// Assign a new UUID for FS minio mode. Each server instance
 	// gets its own UUID for temporary file transaction.
 	fsUUID := mustGetUUID()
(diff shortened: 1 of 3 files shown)


cmd/posix.go
```diff
@@ -155,11 +164,40 @@ var ignoreDiskFreeOS = []string{
...
+	// Remove 5% from total space for cumulative disk space
+	// used for journalling, inodes etc.
+	totalDiskSpace := float64(di.Total) * 0.95
+	if int64(totalDiskSpace) <= diskMinTotalSpace {
+		return errDiskFull
+	}
+
+	// Some filesystems do not implement a way to provide total inodes available, instead inodes
+	// are allocated based on available disk space. For example CephDISK, StoreNext CVDISK, AzureFile driver.
+	// Allow for the available disk to be separately validate and we will validate inodes only if
```
fwessels: separately validate --> separately validated (?)

cmd/posix.go
```diff
@@ -155,11 +164,40 @@ var ignoreDiskFreeOS = []string{
...
+
+	return nil
+}
+
 // checkDiskFree verifies if disk path has sufficient minimum free disk space and files.
 func checkDiskFree(diskPath string, neededSpace int64) (err error) {
 	// We don't validate disk space or inode utilization on windows.
-	// Each windows calls to 'GetVolumeInformationW' takes around 3-5seconds.
-	// And StatFS is not supported by Go for solaris and netbsd.
+	// Each windows calls to 'GetVolumeInformationW' takes around
```
fwessels: calls -> call

fwessels (approved):
Tested and LGTM

cmd/fs-v1.go
```diff
@@ -103,6 +103,11 @@ func newFSObjectLayer(fsPath string) (ObjectLayer, error) {
 		}
 	}
 
+	// Check if disk has minimum required total space.
+	if err = checkDiskTotalMin(fsPath); err != nil {
+		return nil, err
+	}
```
krishnasrinivas: Instead of adding another function `checkDiskTotalMin`, can't we reuse `checkDiskFree`?
like: checkDiskFree(fsPath, 1GB)

harshavardhana: @krishnasrinivas this function uses Total disk space. checkDiskFree for free space only. If you use this during the startup what can happen is if a Disk i.e 100GB and upto 95% full can lead to `minio server` not starting at all. 

Total disk is needed so that we know we can't even have 1GB worth of free space anyways.

krishnasrinivas (approved):

PR #4623, opened 2017-07-03
fix confusing code for http.Header handling

aead:
The extractMetadataFromHeader function should return an error if
the http.Header contains a non-canonicalized key.

## Motivation and Context
The problem is that if someone sets a key of http.Header manually (through a map access)
like `request["a-lower-case-key"] = "something"` the http.Header.Get function still searches for a canonicalized key. This can lead to hard-to-find bugs.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/handler-utils.go
+++ b/cmd/handler-utils.go
@@ -98,9 +98,9 @@ func path2BucketAndObject(path string) (bucket, object string) {
 }
 
 // extractMetadataFromHeader extracts metadata from HTTP header.
-func extractMetadataFromHeader(header http.Header) map[string]string {
+func extractMetadataFromHeader(header http.Header) (map[string]string, error) {
 	if header == nil {
-		return nil
+		return nil, nil
 	}
 	metadata := make(map[string]string)
 	// Save standard supported headers.
@@ -116,16 +116,17 @@ func extractMetadataFromHeader(header http.Header) map[string]string {
 	}
 	// Go through all other headers for any additional headers that needs to be saved.
 	for key := range header {
-		cKey := http.CanonicalHeaderKey(key)
-		if strings.HasPrefix(cKey, "X-Amz-Meta-") {
-			metadata[cKey] = header.Get(key)
-		} else if strings.HasPrefix(key, "X-Minio-Meta-") {
-			metadata[cKey] = header.Get(key)
+		if key != http.CanonicalHeaderKey(key) {
+			return nil, errInvalidArgument
+		}
+		if strings.HasPrefix(key, "X-Amz-Meta-") {
+			metadata[key] = header.Get(key)
+		}
+		if strings.HasPrefix(key, "X-Minio-Meta-") {
+			metadata[key] = header.Get(key)
 		}
 	}
-
-	// Success.
-	return metadata
+	return metadata, nil
 }
 
 // The Query string for the redirect URL the client is
@@ -166,11 +167,6 @@ func trimAwsChunkedContentEncoding(contentEnc string) (trimmedContentEnc string)
 		}
 	}
 	return strings.Join(newEncs, ",")
-}
-
-// extractMetadataFromForm extracts metadata from Post Form.
-func extractMetadataFromForm(formValues http.Header) map[string]string {
-	return extractMetadataFromHeader(formValues)
 }
 
 // Validate form field size for s3 specification requirement.
--- a/cmd/handler-utils.go
+++ b/cmd/handler-utils.go
@@ -100,7 +100,7 @@ func path2BucketAndObject(path string) (bucket, object string) {
 // extractMetadataFromHeader extracts metadata from HTTP header.
 func extractMetadataFromHeader(header http.Header) (map[string]string, error) {
 	if header == nil {
-		return nil, nil
+		return nil, traceError(errInvalidArgument)
 	}
 	metadata := make(map[string]string)
 	// Save standard supported headers.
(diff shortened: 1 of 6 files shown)


krishnasrinivas (approved):

cmd/handler-utils.go
```diff
@@ -116,11 +116,10 @@ func extractMetadataFromHeader(header http.Header) map[string]string {
 	}
 	// Go through all other headers for any additional headers that needs to be saved.
 	for key := range header {
-		cKey := http.CanonicalHeaderKey(key)
-		if strings.HasPrefix(cKey, "X-Amz-Meta-") {
-			metadata[cKey] = header.Get(key)
+		if strings.HasPrefix(key, "X-Amz-Meta-") {
+			metadata[key] = header.Get(key)
```
harshavardhana: are you sure that key is always Canonical? @aead  - this code was added on purpose. and why is it confusing?

harshavardhana (changes requested):

cmd/handler-utils.go
harshavardhana: https://play.golang.org/p/hCJ6n1zLZQ 

```go
package main

import (
	"fmt"
	"net/http"
)

func main() {
	headers := http.Header{}
	headers.Set("content-type", "test")
	headers["x-amz-meta-x-amz-key"] = []string{"vvv"}
	for k, v := range headers {
		fmt.Println(k, v)
	}
}
```

How do you handle a situation like this? is it expected that `headers` is not modified by the caller?

harshavardhana: So here is the reason why this is done.. https://play.golang.org/p/KRWnukcwVd

```go
package main

import (
	"fmt"
	"net/http"
	"strings"
)

func main() {
	headers := http.Header{}
	headers.Set("content-type", "test")
	headers["x-amz-meta-x-amz-key"] = []string{"vvv"}
	for k, v := range headers {
		key := http.CanonicalHeaderKey(k)
		if strings.HasPrefix(key, "X-Amz-Meta") {
			fmt.Println(key, v)
		}
	}
}
```

cmd/handler-utils.go
```diff
@@ -116,16 +116,17 @@ func extractMetadataFromHeader(header http.Header) map[string]string {
 	}
 	// Go through all other headers for any additional headers that needs to be saved.
 	for key := range header {
-		cKey := http.CanonicalHeaderKey(key)
-		if strings.HasPrefix(cKey, "X-Amz-Meta-") {
-			metadata[cKey] = header.Get(key)
-		} else if strings.HasPrefix(key, "X-Minio-Meta-") {
-			metadata[cKey] = header.Get(key)
+		if key != http.CanonicalHeaderKey(key) {
```
donatello, line 119: I don't understand why this is needed - the only way to get a non-canonical header is if we add code that changes request headers. Request headers are not supposed to be modified - any code that does so is introducing bugs. So checking for this and returning errors to the user does not appear to be correct, as they could not have caused such a problem in the first place.

donatello (changes requested):

cmd/handler-utils.go
harshavardhana: It is quite possible @donatello because this is a function and http.Header by nature is not protected. So anyone can modify and use it. This function is defining its minimum conditions not hoping that http.Header itself is never modified. 

IMO it is better to error out if we do indeed see an unexpected key since this function cannot operate in any other way.

cmd/handler-utils.go
```diff
@@ -98,9 +98,9 @@ func path2BucketAndObject(path string) (bucket, object string) {
 }
 
 // extractMetadataFromHeader extracts metadata from HTTP header.
-func extractMetadataFromHeader(header http.Header) map[string]string {
+func extractMetadataFromHeader(header http.Header) (map[string]string, error) {
 	if header == nil {
-		return nil
+		return nil, nil
```
harshavardhana: Even this is `errInvalidArgument`

harshavardhana (changes requested):

cmd/web-handlers.go
```diff
@@ -497,7 +497,11 @@ func (web *webAPIHandlers) Upload(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// Extract incoming metadata if any.
-	metadata := extractMetadataFromHeader(r.Header)
+	metadata, err := extractMetadataFromHeader(r.Header)
+	if err != nil {
+		writeErrorResponse(w, ErrInvalidMetadataDirective, r.URL)
```
donatello: The error being returned here is not correct - `extractMetadataFromHeader` returns an `ErrInvalidArgument` but we return an error about metadata directive? It should be an internal server error with a message about an invalid condition being found in the server. An invalid canonical header cannot be caused by the user or an application, so it is not correct so send such an error back to the user.

donatello (changes requested):

harshavardhana (dismissed):

cmd/handler-utils.go
deekoder: Any thoughts @aead  further?

harshavardhana: It's already addressed @deekoder look at the patch itself.

donatello (approved):

aead:
I think the reason for decreasing codecov are checks like:
```
metadata, err := extractMetadataFromHeader(r.Header)
if err != nil {
	errorIf(err, "found invalid http request header")
	writeErrorResponse(w, ErrInternalError, r.URL)
	return
}
```
The change itself LGTM - (CI passes). So from my point of view: Can be merged..

cmd/handler-utils.go
deekoder: @y4m4 wanted to confirm with @aead. Please see Travis failed.

aead:
Go cover says:

The hole project:
Before change: `coverage: 70.1% of statements`
After change:    `coverage: 70.0% of statements`

The change itself:
Before change: `coverage: 0.5% of statements`
After change:    `coverage: 0.6% of statements`

See `go test -v -cover -run=TestExtractMetadataHeaders`

aead:
So the coverage decrease is caused by an introduced error check:
`err =  extractMetadataFromHeader(r.Heade)`
The error branch cannot be reached because of the way Go handles `http.Request` && `http.Header`
The only reason for this checks is, that we don't want to ignore errors - So it should be fine to merge this.

PR #4646, opened 2017-07-07
Implement an offline mode for a distributed node

### LINKED ISSUE #4183: Performance degradation with distributed minio and an offline cluster.
## Expected Behavior
Writes should continue to perform well when a node is offline.

## Current Behavior
There is a noticeable delay with put_object when a node if offline (even though a write quorum exists). 

## Possible Solution
Adding a circuit breaker (https://en.wikipedia.org/wiki/Circuit_breaker_design_pattern) would allow a minio cluster to perform well when a node was know to be down. The initial write(s) would still be slower until the circuit breaker was triggered but thereafter it would perform well. This would enable maintenance on individual nodes without negatively impacting the performance of clients.

1. Start a 4 node distributed cluster (I used the docker-compose https://raw.githubusercontent.com/minio/minio/master/docs/orchestration/docker-compose/docker-compose.yaml)
2. Configure a host: mc config host add node1 http://127.0.0.1:9001 minio minio123 S3v4
3. Create a bucket: mc mb node1/test
4. Create a file: dd if=/dev/zero of=./file bs=1024 count=1024
5. Copy that file to the bucket:  time mc cp ./file node1/test/first
6. Note how fast it uploads
7. Stop a node: docker-compose stop minio2
8. Re-copy that file to the bucket:  time mc cp ./file node1/test/second
9. Note it's much slower.
10. Repeating 8-10 always takes a long time.

## Context
I'm evaluating minio for use in a production environment. We'll need to deploy periodic security updates on minio nodes and will not want this to have an adverse effect on performance.

## Your Environment
Version: 2017-04-25T01:27:49Z
Release-Tag: RELEASE.2017-04-25T01-27-49Z
Commit-ID: 710db6bdadb1b228bf6dee7e2a6958000091125b
* Environment name and version (e.g. nginx 1.9.1): N/A
* Server type and version: N/A
Running in docker on a linux host (see docker-compose file linked above)
Host: Linux xenial 4.4.0-72-generic #93-Ubuntu SMP Fri Mar 31 14:07:41 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
* Link to your project: N/A

fwessels:
Implement an offline mode for retryStorage to cache the offline status of a node in order to prevent network calls that are bound to fail.

After a time interval an attempt will be made to restore the connection and mark the node as online if successful.

## Motivation and Context
This fixes #4183 where a performance degradation is reported with distributed minio and an offline cluster.

## How Has This Been Tested?
Manually in a 4 node cluster by bringing down a server and verifying that the node is marked as down. And after starting the node again, that the node is marked as online again (and repeated multiple times).

## Types of changes

## Checklist:


krishnasrinivas (commented):
will review

cmd/retry-storage.go
```diff
@@ -30,210 +30,276 @@ const (
...
+	// Only after initialization and minimum of one interval
+	// has passed (to prevent marking a node as offline right
+	// after initialization), check whether node has gone offline
+	if f.maxRetryAttempts == globalStorageRetryThreshold &&
+		UTCNow().Sub(f.offlineTimestamp) >= globalStorageHealthCheckInterval {
+		if e == errDiskNotFoundFromNetError { // Make node offline due to network error
+			f.offline = true // Marking node offline
+			f.offlineTimestamp = UTCNow()
+			return errDiskNotFound
+		} else if e == errDiskNotFoundFromRPCShutdown {
```
aead: Is this for explicitness or why is there a no-op `e == errDiskNotFoundFromRPCShutdown` check?

fwessels: These are the two types of specific DiskNotFound errors returned from `toStorageErr`, so both are checked.

But looking at it again, it is probably cleaner to remove the else if clause and let the code continue anyway.

aead (approved):
LGTM

krisis (approved):

krisis:
Tested locally following steps to reproduce from https://github.com/minio/minio/issues/4183. Works as expected.

harshavardhana (approved):

PR #4636, opened 2017-07-06
Refactor HTTP server to address bugs

balamurugana:
## Description

* Remove unnecessary goroutine to start multiple TCP listeners.
* HTTP server waits for shutdown to maximum of Server.ShutdownTimeout
  than per serverShutdownPoll.
* Handles new connection errors properly.
* Handles read and write timeout properly.
* Handles error on start of HTTP server properly by exiting minio
  process.

Fixes #4494 #4476
## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #4647, opened 2017-07-08
rpc: Do not use deadlines for rpc connections.

### LINKED ISSUE #4626: Distributed Docker Minio keeps throwing disk errors
I spoke briefly with kp in the Slack channel about this issue.

I have four servers, and one Minio instance on each, running as a four-disk distributed cluster.

It works for a short time, then starts throwing errors a lot, like:

`msg="Unable to fetch disk info for &cmd.networkStorage{rpcClient:(*cmd.AuthRPCClient)(0xc4202a9760)}" cause="disk not found" source="[xl-v1.go:199:getDisksInfo()]"`

Running `mc admin info` says that three of the four nodes see three disks offline each; but after some seconds, it goes back to saying all disks online. But attempting to make a bucket fails if run again.

This is the setup. Assume you have four nodes, called m1, m2, m3, m4, then run this on each one:

```
mkdir -p /opt/minio/export
docker run --rm --hostname={{ ansible_fqdn}} \
    -p 9000:9000 \
    -v /opt/minio/export:/export \
    -e MINIO_ACCESS_KEY=something \
    -e MINIO_SECRET_KEY=somethingsecret \
    minio/minio server --address :9000 \
    http://m1:9000/export \
    http://m2:9000/export \
    http://m3:9000/export \
    http://m4:9000/export

```

Configure `mc` to talk to it, and just start creating buckets and adding files to them, removing files, and removing the bucket, with a delay. I guess you have test scripts like these that are more powerful than mine! (which is essentially a few lines, a counter, and a sleep, so I end up with `mc mb test/test$n` etc)

I find it starts failing quite quickly, especially if I run `mc admin heal test` a few times.

harshavardhana:
## Description
Disable deadlines on RPC connections.

## Motivation and Context
Fixes #4626

## How Has This Been Tested?
Manually using a distributed setup.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/rpc-server.go
+++ b/cmd/rpc-server.go
@@ -1,0 +1,58 @@
+/*
+ * Minio Cloud Storage, (C) 2017 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cmd
+
+import (
+	"io"
+	"net/http"
+	"net/rpc"
+
+	miniohttp "github.com/minio/minio/pkg/http"
+)
+
+// ServeHTTP implements an http.Handler that answers RPC requests,
+// hijacks the underlying connection and clears all deadlines if any.
+func (server *rpcServer) ServeHTTP(w http.ResponseWriter, req *http.Request) {
+	if req.Method != http.MethodConnect {
+		w.WriteHeader(http.StatusMethodNotAllowed)
+		return
+	}
+
+	conn, _, err := w.(http.Hijacker).Hijack()
+	if err != nil {
+		errorIf(err, "rpc hijacking failed for: %s", req.RemoteAddr)
+		return
+	}
+
+	// Overrides Read/Write deadlines if any.
+	bufConn, ok := conn.(*miniohttp.BufConn)
+	if ok {
+		bufConn.RemoveTimeout()
+		conn = bufConn
+	}
+
+	// Can connect to RPC service using HTTP CONNECT to rpcPath.
+	io.WriteString(conn, "HTTP/1.0 200 Connected to Go RPC\n\n")
+	server.ServeConn(conn)
+}
+
+type rpcServer struct{ *rpc.Server }
+
+// Similar to rpc.NewServer() provides a custom ServeHTTP override.
+func newRPCServer() *rpcServer {
+	return &rpcServer{rpc.NewServer()}
+}
--- a/cmd/rpc-server_test.go
+++ b/cmd/rpc-server_test.go
@@ -1,0 +1,76 @@
+/*
+ * Minio Cloud Storage, (C) 2017 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cmd
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	router "github.com/gorilla/mux"
+)
+
+type ArithArgs struct {
+	A, B int
+}
+
+type ArithReply struct {
+	C int
+}
+
+type Arith int
+
+// Some of Arith's methods have value args, some have pointer args. That's deliberate.
+
+func (t *Arith) Add(args ArithArgs, reply *ArithReply) error {
+	reply.C = args.A + args.B
+	return nil
+}
+
+func TestGoHTTPRPC(t *testing.T) {
+	newServer := newRPCServer()
+	newServer.Register(new(Arith))
+
+	mux := router.NewRouter().SkipClean(true)
+	mux.Path("/foo").Handler(newServer)
+
+	httpServer := httptest.NewServer(mux)
+	defer httpServer.Close()
+
+	client := newRPCClient(httpServer.Listener.Addr().String(), "/foo", false)
+	defer client.Close()
+
+	// Synchronous calls
+	args := &ArithArgs{7, 8}
+	reply := new(ArithReply)
+	if err := client.Call("Arith.Add", args, reply); err != nil {
+		t.Errorf("Add: expected no error but got string %v", err)
+	}
+
+	if reply.C != args.A+args.B {
+		t.Errorf("Add: expected %d got %d", reply.C, args.A+args.B)
+	}
+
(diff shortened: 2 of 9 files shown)


deekoder:
As per @harshavardhana blocked on this until #4636 is merged.

harshavardhana:
This PR is updated and its not blocked anymore.. ready for reviews.

krishnasrinivas:
@harshavardhana 
compile failed
```
krishna@escape:~/gocode/src/github.com/minio/minio$ go install
# github.com/minio/minio/cmd
cmd/rpc-server.go:77: undefined: "github.com/minio/minio/pkg/http".BuffConn

```

cmd/rpc-server.go
```diff
@@ -0,0 +1,82 @@
...
+	if req.Method != http.MethodConnect {
+		w.Header().Set("Content-Type", "text/plain; charset=utf-8")
+		w.WriteHeader(http.StatusMethodNotAllowed)
+		io.WriteString(w, fmt.Sprintf("405 must %s\n", http.MethodConnect))
+		return
+	}
+
+	conn, _, err := w.(http.Hijacker).Hijack()
+	if err != nil {
+		log.Printf("rpc hijacking %s: %s\n", req.RemoteAddr, err.Error())
```
krishnasrinivas: log.Printf should be removed?

krishnasrinivas:
The problem does not seem to be fixed @harshavardhana 

if I ctrl-C in distributed mode, I should see :
```
^CExiting on signal interrupt

ERRO[0044] Unable to shutdown http server                cause=timed out. some connections are still active. doing abnormal shutdown source=[signals.go:42:handleSignals.func2()]
```

but I don't see it when I ctrl-C

cmd/rpc-server.go
```diff
@@ -0,0 +1,67 @@
...
+	"io"
+	"net"
+	"net/http"
+	"net/rpc"
+
+	http2 "github.com/minio/minio/pkg/http"
+)
+
+// Can connect to RPC service using HTTP CONNECT to rpcPath.
+var connected = "200 Connected to Go RPC"
```
balamurugana: This needs to be `const` and name could be `rpcConnectReply`

harshavardhana:
Done with all the final changes @balamurugana @donatello @krishnasrinivas . RPC connections are holding up as ESTABLISHED. Any questions let me know we can discuss further.

cmd/rpc-server.go
```diff
@@ -0,0 +1,67 @@
...
+package cmd
+
+import (
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"net/rpc"
+
+	http2 "github.com/minio/minio/pkg/http"
```
balamurugana: you can name it as `miniohttp`

cmd/rpc-server.go
```diff
@@ -0,0 +1,67 @@
...
+)
+
+// Can connect to RPC service using HTTP CONNECT to rpcPath.
+const rpcConnectReply = "200 Connected to Go RPC"
+
+func (server *rpcServer) ServeHTTP(w http.ResponseWriter, req *http.Request) {
+	if req.Method != http.MethodConnect {
+		w.Header().Set("Content-Type", "text/plain; charset=utf-8")
+		w.WriteHeader(http.StatusMethodNotAllowed)
+		io.WriteString(w, fmt.Sprintf("405 must %s\n", http.MethodConnect))
```
balamurugana: I think we don't need to send a body. If we do we can remove header setting also.

cmd/rpc-server.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	conn, _, err := w.(http.Hijacker).Hijack()
+	if err != nil {
+		errorIf(err, "rpc hijacking: %s", req.RemoteAddr)
+		return
+	}
+
+	// Overrides Read/Write deadlines if any.
+	conn = newRPCConn(conn)
```
balamurugana: this function is not required. you could directly use it like
```go
bufConn, ok := conn.(*miniohttp.BufConn)
if ok {
    bufConn.DisableTimeout()
    conn = bufConn
}
```

cmd/rpc-server.go
```diff
@@ -0,0 +1,67 @@
...
+
+	io.WriteString(conn, "HTTP/1.0 "+rpcConnectReply+"\n\n")
```
balamurugana: You could keep the whole string ie `HTTP/1.0 200 Connected to Go RPC\r\n` in `rpcConnectReply`

cmd/rpc-server.go
```diff
@@ -0,0 +1,67 @@
...
+	server.ServeConn(conn)
+}
+
+type rpcServer struct {
+	*rpc.Server
```
balamurugana: If `rpcServer` extends `rpc.Server`, then `newRPCServer()` is simply `return &rpcServer{}`

harshavardhana: we need to do rpc.NewServer() we can't do `&rpcServer{}` it won't allocate otherwise.

balamurugana: Got it.

cmd/rpc-server.go
```diff
@@ -0,0 +1,58 @@
...
+	"io"
+	"net/http"
+	"net/rpc"
+
+	miniohttp "github.com/minio/minio/pkg/http"
+)
+
+func (server *rpcServer) ServeHTTP(w http.ResponseWriter, req *http.Request) {
+	if req.Method != http.MethodConnect {
+		w.Header().Set("Content-Type", "text/plain; charset=utf-8")
```
balamurugana: Header setting is not useful anymore.

harshavardhana:
The bug reported #4677 is also fixed by this PR. Since this PR test case was blocking on it..

cmd/rpc-server_test.go
```diff
@@ -0,0 +1,106 @@
...
+		t.Fatal(err)
+	}
+
+	if resp.StatusCode != http.StatusMethodNotAllowed {
+		t.Errorf("Expected %d, got %d", http.StatusMethodNotAllowed, resp.StatusCode)
+	}
+}
+
+/*
+func TestHTTPRPC(t *testing.T) {
```
krishnasrinivas: this should be removed?

cmd/rpc-server.go
```diff
@@ -0,0 +1,58 @@
...
+	bufConn, ok := conn.(*miniohttp.BufConn)
+	if ok {
+		bufConn.RemoveTimeout()
+		conn = bufConn
+	}
+
+	// Can connect to RPC service using HTTP CONNECT to rpcPath.
+	io.WriteString(conn, "HTTP/1.0 200 Connected to Go RPC\n\n")
+	server.ServeConn(conn)
+}
```
krishnasrinivas, line 51: instead you can just do this:
```
func (server *rpcServer) ServeHTTP(w http.ResponseWriter, req *http.Request) {                                                                                                  
        conn, _, err := w.(http.Hijacker).Hijack()                                                                                                                              
        if err != nil {                                                                                                                                                         
                errorIf(err, "rpc hijacking failed for: %s", req.RemoteAddr)                                                                                                    
                return                                                                                                                                                          
        }                                                                                                                                                                       
                                                                                                                                                                                
        // Overrides Read/Write deadlines if any.                                                                                                                               
        bufConn, ok := conn.(*miniohttp.BufConn)                                                                                                                                
        if ok {                                                                                                                                                                 
                bufConn.RemoveTimeout()                                                                                                                                         
        }                                                                                                                                                                       
                                                                                                                                                                                
        server.Server.ServeHTTP(w, req)                                                                                                                                         
}
```

harshavardhana: Once hijacked you cant put it back.. you have to now manage the net.Conn

cmd/rpc-server_test.go
harshavardhana: You are looking at an old PR.

cmd/rpc-server.go
krishnasrinivas: ok

balamurugana (approved):

krishnasrinivas (approved):

PR #4651, opened 2017-07-09
gcs: Fix writer/reader go-routine leaks and code re-use

harshavardhana:
## Description
This PR serves to fix following things in GCS gateway.

- fixes leaks in object reader and writer, not getting closed
  under certain situations. This led to go-routine leaks.

- apparent confusing issue in case of complete multipart upload,
  where it is currently possible for an entirely different
  object name to concatenate parts of a different object name
  if you happen to know the upload-id and parts of the object.
  This is a very rare scenario but it is possible.

- succint usage of certain parts of code base and re-use.
  

## Motivation and Context
Cleanup change. 

## How Has This Been Tested?
Manually.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/gateway-gcs.go
+++ b/cmd/gateway-gcs.go
@@ -28,7 +28,6 @@ import (
 	"math"
 	"regexp"
 	"strings"
-	"time"
 
 	"golang.org/x/oauth2/google"
 
@@ -53,7 +52,8 @@ const (
 	// Multipart meta file.
 	gcsMinioMultipartMeta = "gcs.json"
 	// gcs.json version number
-	gcsMultipartMetaCurrentVersion = "1"
+	gcsMinioMultipartMetaCurrentVersion = "1"
+
 	// token prefixed with GCS returned marker to differentiate
 	// from user supplied marker.
 	gcsTokenPrefix = "##minio"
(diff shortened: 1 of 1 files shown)


cmd/gateway-gcs.go
```diff
@@ -684,30 +692,24 @@ func (l *gcsGateway) PutObject(bucket string, key string, size int64, data io.Re
 
 	w.ContentType = metadata["content-type"]
 	w.ContentEncoding = metadata["content-encoding"]
+	w.Metadata = metadata
```
krisis, line 694: Is there any reason we are not resorting to `defer w.Close()`?

cmd/gateway-gcs.go
```diff
@@ -785,51 +792,54 @@ func (l *gcsGateway) CopyObjectPart(srcBucket string, srcObject string, destBuck
...
 		sha256Writer = sha256.New()
 		reader = io.TeeReader(data, sha256Writer)
 	}
 
-	dataName := gcsMultipartDataName(uploadID, etag)
-
-	object = l.client.Bucket(bucket).Object(dataName)
-
+	object := l.client.Bucket(bucket).Object(gcsMultipartDataName(uploadID, etag))
 	w := object.NewWriter(l.ctx)
```
krisis, line 826: We can consistently use `defer w.Close()` in these situations.

cmd/gateway-gcs.go
```diff
@@ -908,29 +910,40 @@ func (l *gcsGateway) CompleteMultipartUpload(bucket string, key string, uploadID
 	if err != nil {
 		return ObjectInfo{}, gcsToObjectError(traceError(err), bucket, key)
 	}
+
+	if len(uploadedParts) > maxPartCount {
```
krisis: We should check if sum of component count of all components exceed 1024.

cmd/gateway-gcs.go
```diff
@@ -908,29 +910,40 @@ func (l *gcsGateway) CompleteMultipartUpload(bucket string, key string, uploadID
...
+		return ObjectInfo{}, traceError(NotSupported{})
+	}
+
 	r, err := object.NewReader(l.ctx)
```
krisis: We can use `defer r.Close()` instead of closing the `ReadCloser` in every failure branch individually.

krisis (changes requested):

cmd/gateway-gcs.go
harshavardhana: The reason is that only after w.Close() we can validate the sha256sum. i.e after a flush.

cmd/gateway-gcs.go
harshavardhana: Explained above why.

cmd/gateway-gcs.go
harshavardhana: If we have input list > 1024 we can't concatenate anyways.. not sure if i understand you correct.

cmd/gateway-gcs.go
harshavardhana: We can do that.. for readers.

harshavardhana: Actually i want to close the reader ASAP not in defer.. it is just to purge resources quickly.  Because we do much more in the later calls. 

It is sort of a pattern we have used with writer anyways..

balamurugana (approved):
Not tested

krisis (approved):

PR #4661, opened 2017-07-10
erasure: fix erasureCreateFile behavior for empty part

### LINKED ISSUE #4660: erasure: createFile does not return an error if empty part is not allowed
If you create an erasure coded file with 0 content and the allowEmpty flag is false the sever should return an io.EOF but it returns nil (success)

## Expected Behavior
sever should return io.EOF

## Current Behavior
server returns success

## Possible Solution
Fix erasureCreateFile

add this to erasure-create_test.go
```
_, size, _, err := erasureCreateFile(disks, "testbucket", "testobject1", bytes.NewReader(data[:0]), false, blockSize, dataBlocks, parityBlocks, defaultBitRotAlgorithm, dataBlocks+1)
if err != io.EOF {
	t.Fatal(err)
}
```

aead:
Fixes #4660

## Description
See issue

## Types of changes

## Checklist:

### DIFF
--- a/cmd/erasure-createfile.go
+++ b/cmd/erasure-createfile.go
@@ -46,6 +46,9 @@ func erasureCreateFile(disks []StorageAPI, volume, path string, reader io.Reader
 			return nil, 0, nil, traceError(rErr)
 		}
 		if rErr == io.EOF {
+			if bytesWritten == 0 && !allowEmpty {
+				return nil, 0, nil, rErr
+			}
 			// We have reached EOF on the first byte read, io.Reader
 			// must be 0bytes, we don't need to erasure code
 			// data. Will create a 0byte file instead.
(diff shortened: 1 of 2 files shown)


cmd/erasure-createfile.go
```diff
@@ -46,6 +46,9 @@ func erasureCreateFile(disks []StorageAPI, volume, path string, reader io.Reader
 			return nil, 0, nil, traceError(rErr)
 		}
 		if rErr == io.EOF {
+			if bytesWritten == 0 && !allowEmpty {
```
harshavardhana, line 49: Under what scenario it is true that allowEmpty is false?

harshavardhana (changes requested):

cmd/erasure-createfile.go
aead: PutObject: first part is can be empty, all other parts not
Even if higher label may prevent this (?!), I would treat this as a bug.
Found this by adding test cases for erasreCreateFile...

aead: *layer

aead:
Will be fixed for encryption, can never occur for sever (top layer checks)

PR #4668, opened 2017-07-12
(WIP) require old credentials for RPC SetAuth

### LINKED ISSUE #4182: [security] Improve browser-based password change flow
Currently, on login the browser receives a token generated by the server and after this the client-side code no longer retains access to the main credentials (access and secret key). The token is used to authorize all  actions performed via the browser. It has a limited validity for security. However, a credentials change action can be performed with just the token - this gives **permanent** access to the minio server.

While this is not an easily exploitable vulnerability (AFAIK), especially when HTTPS is used, security may be improved by requiring the previous credentials to be given (and verified) before credentials can be changed.

## Expected Behavior

Password change should require the old password to be verified before being allowed.

## Current Behavior

Password can be changed with just the JWT received by the browser after logging in.

brendanashworth:
## Description
This commit requires the old server credentials for the SetAuth RPC
call. It will now fail if the client does not present the previous
credentials in addition to the new ones.

This commit also changes the browser to provide those credentials when
calling `SetAuth` (via an additional `GetAuth` call).

Previous to this, only an access token was necessary to set the auth.
However, access tokens are temporary while auth is permanent. Thus, the
client must have both the token and the previous auth, so as to ensure
they have access to change it.

## Motivation and Context

Fixes: https://github.com/minio/minio/issues/4182
## How Has This Been Tested?

Tests have been added.
## Types of changes

Backwards compatibility shouldn't be an issue because the commit also updates the browser code, which AFAIK is the only thing that calls RPC `SetAuth`.

## Checklist:


krishnasrinivas:
Discussed with @brendanashworth that this patch does not fix the intent of the #4182

We have two requirements that conflict with each other:
* Ability to show current access/secret keys on the browser
* Ask for current secret key to change the access/secret key

Need to see how we can solve this.

brendanashworth:
What is above ^ is correct, using `GetAuth` only requires a token so it's kind of pointless to use that before `SetAuth`. I'll close this PR for now because it's quite misguided and can't really be salvaged.

PR #4671, opened 2017-07-12
browser: fix deleting '*' bucket policy

### LINKED ISSUE #4498: [browser] Removing a bucket policy after an error popup doesn't have an effect
## Expected Behavior
Removing a bucket policy *really* removes it.

## Current Behavior
Removing a bucket policy doesn't have an effect after an error popup

## Possible Solution
Don't know

1. Start from fresh FS disk
2. Login into Minio Browser
3. Create a bucket
4. Show bucket policy popup
5. Add '*'  policy
6. Add 'foo' policy
7. Add '*' policy again
(an error popup is shown to say that the user is adding an incompatible policy rule which is correct)
8. Now remove all existing policies
9. Hide policy dialog 
10. Show policy dialog  again

You'll notice that you have a policy rule though you removed all existing rules.

## Context
Testing before release

## Your Environment
minio master, all environments

brendanashworth:
## Description

This commit fixes deleting a bucket policy with a prefix of `*`
Previously, it would rewrite the prefix to '' (empty string), which
wouldn't match the set policy on `*`. By now rewriting the empty string
to `*`, it will allow the server to correctly remove the policy in
question.

## Motivation and Context

Fixes: https://github.com/minio/minio/issues/4498

## How Has This Been Tested?

just on the browser

## Types of changes

## Checklist:


krishnasrinivas:
This might not be the right fix, will review more.

rushenn (approved):

krishnasrinivas:
There is underlying problem which is causing complication and confusion in the web browser code.

for the policy package:
policy.GetPolicies returns a map[string]string i.e it maps prefix to policy-name (like readonly, readwrite..)
Here prefix does not have a trailing `*`
Here prefix also includes bucket name

policy.SetPolicy takes bucket and object-prefix as separate arguments and object-prefix is not expected to have trailing `*` as the package adds it.

These two inconsistencies need to be fixed first and the browser code should be fixed accordingly.

Fix is:
1) GetPolicies should return bucket and prefix as separate elements
2) SetPolicy should not expect trailing `*` for prefix
3) All the other code should be fixed accordingly

The fix in this commit does not fix the problem. I have updated the issue https://github.com/minio/minio/issues/4498 (Step 6 in the repro steps should have given an error)

brendanashworth:
@krishnasrinivas is right, the browser needs a more complete fix. Right now it's just throwing extra `*`s under the rug, but sometimes object names can contain stars, etc, and there's some internal differences between how we handle trailing asterisks. This fix isn't enough.

PR #4673, opened 2017-07-13
(wip) browser: remove token from GET URLs

brendanashworth:
## Description
This commit removes the token from all browser-operated GET URLs. It
moves the tokens into the Authorization: header. This should be more
idiomatic and slightly more secure. This commit updates the tests.

## Motivation and Context

From discussion with @krishnasrinivas 

## Types of changes

## Checklist:

### DIFF
--- a/browser/app/js/components/Browse.js
+++ b/browser/app/js/components/Browse.js
@@ -145,12 +145,43 @@ export default class Browse extends React.Component {
   selectPrefix(e, prefix) {
     e.preventDefault()
     const {dispatch, currentPath, web, currentBucket} = this.props
-    const encPrefix = encodeURI(prefix)
+    const objectName = encodeURI(prefix)
     if (prefix.endsWith('/') || prefix === '') {
       if (prefix === currentPath) return
-      browserHistory.push(utils.pathJoin(currentBucket, encPrefix))
+      browserHistory.push(utils.pathJoin(currentBucket, objectName))
     } else {
-      window.location = `${window.location.origin}/minio/download/${currentBucket}/${encPrefix}?token=${storage.getItem('token')}`
+      let xhr = new XMLHttpRequest()
+      let url = `${location.origin}/minio/download/${currentBucket}/${objectName}`
+
+      xhr.open('GET', url, true)
+      xhr.setRequestHeader('Authorization', 'Bearer ' + storage.getItem('token'))
+
+      // The response must allow binary data.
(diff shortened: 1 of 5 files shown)


brendanashworth:
ping @krishnasrinivas this should look better now

browser/app/js/components/Browse.js
```diff
@@ -145,12 +145,43 @@ export default class Browse extends React.Component {
...
     } else {
-      window.location = `${window.location.origin}/minio/download/${currentBucket}/${encPrefix}?token=${storage.getItem('token')}`
+      let xhr = new XMLHttpRequest()
+      let url = `${location.origin}/minio/download/${currentBucket}/${objectName}`
+
+      xhr.open('GET', url, true)
+      xhr.setRequestHeader('Authorization', 'Bearer ' + storage.getItem('token'))
+
+      // The response must allow binary data.
+      xhr.responseType = 'arraybuffer'
```
krishnasrinivas, line 160: Can you check what is better for us here `arraybuffer` or `blob`? Because we don't want the browser to cache everything in memory first before writing to the disk. Browser should save it to the disk in a streaming fashion.

brendanashworth:
As of right now, the current PR has platform-dependent issues (Safari stores downloaded XHR blobs in memory, so it crashes on large objects). We need to find a better solution.

harshavardhana:
https://github.com/eligrey/FileSaver.js/issues/12 nice history about arraybuffer and blob on Safari

brendanashworth:
@harshavardhana thanks for the link — but that doesn't even take into account the memory issues. [This library](https://github.com/jimmywarting/StreamSaver.js) is intended to be an alternative to FileSaver ("But there is one obstacle - The RAM it can hold and the max blob size limitation") because that library doesn't handle the memory limits.

krishnasrinivas:
@harshavardhana choices are:

* use Cookies for Auth and support downloads on all browsers. Wouldn't prefer this mainly because we have to be careful about csrf bugs. Also we have to maintain sessions on the disk on the server side.
* use Token as part of the URL itself, wouldn't prefer this as it is not a good practice.
* support only FF/Chrome
* support other browsers, but support downloads only on Chrome/FF

krishnasrinivas:
@brendanashworth discussed with @abperiasamy , for this we can create a short lived tokens that can be used in the URL

brendanashworth:
Closing in favor of https://github.com/minio/minio/pull/4684.

PR #4678, opened 2017-07-14
erasure: replace bitrot protection mechanism

### LINKED ISSUE #4359: Create exhaustive unit tests for erasure code

### LINKED ISSUE #4649: Organize packages related to bitrot and hashing algorithms
Vendorize external packages already published.

aead:
This change:
 - Replaces the hash.Hash based bit-rot protection mechanism with a more generic approach (`bitrot.Hash`) - see `/pkg/bitrot`
  - Adds a implementation for [Poly1305](https://en.wikipedia.org/wiki/Poly1305) - see `vendor/github.com/aead/poly1305`
 - Makes poly1305 the default bit-rot protection algorithm
 - Provide new implementations for the file operations (create, read, heal) of the XL backend
 - Add new tests for those file operations
  
## Description
The change does basically two this:
 - It introduces a new bit-rot protection approach - based on polynomial authenticators. To provide a generic interface for bit-rot protection and support hash functions as well as poly. authenticators the general concept of detection data-manipulation is abstracted by the new `bitrot` package. Further the new concept makes it very easy to support new algorithms / different implementations. 

- The second part are the erasure file operations (create, read, heal). This change provides new implementations for those operations and adds more test cases.

## Motivation and Context
 - This change will reduce the CPU load while protecting/verifying erasure-coded data.
 - Simplify erasure file operation implementations. 

## How Has This Been Tested?
New test cases are added for create, read and heal.
All new / existing tests pass.

## Types of changes

## Checklist:

Fixes #4649 
Fixes #4359


harshavardhana (approved):
We reviewed this previously and merged. LGTM :+1:

harshavardhana (changes requested):
We need to make one change undoing my :+1: we haven't bumped the XL version number in this one. We need to address that and also test backward compatible changes.

aead:
@harshavardhana 
@abperiasamy reviewed the XL format change and told me to not bump the version because the change does not add any incompatibility. "New" sever will set key to zero value if "old" XL format is used...

cmd/erasure.go
```diff
@@ -0,0 +1,125 @@
+package cmd
```
fwessels, line 17: Header missing

pkg/bitrot/bitrot.go
```diff
@@ -0,0 +1,151 @@
...
+	Poly1305:   32,
+	GHASH:      32,
+	BLAKE2b512: 0,
+	SHA256:     0,
+}
+
+var names = []string{
+	Poly1305:   "poly1305",
+	GHASH:      "ghash",
+	BLAKE2b512: "blake2b",
```
fwessels: Consider renaming constant `BLAKE2b512' to 'BLAKE2b' in order to match the string name (it is now the odd one out here, and blake2b is what is current written in the json).

cmd/erasure.go
```diff
@@ -0,0 +1,135 @@
...
+	"github.com/minio/minio/pkg/bitrot"
+)
+
+// OfflineDisk represents an unavailable disk.
+var OfflineDisk StorageAPI // zero value is nil
+
+// MissingShard indicates a missing data/parity shard for erasure coding
+var MissingShard []byte // zero value is nil
+
+// ErasureFileInfo contains information about a erasure file operation (create, read, heal).
```
fwessels: about a erasure --> about an erasure

cmd/erasure.go
```diff
@@ -0,0 +1,135 @@
...
+type ErasureFileInfo struct {
+	Size      int64
+	Algorithm bitrot.Algorithm
+	Keys      [][]byte
+	Checksums [][]byte
+}
+
+// XLStorage represents an array of disks.
+// The disks contains erasure coded and bitrot-protected data.
```
fwessels: contains --> contain

cmd/erasure.go
```diff
@@ -0,0 +1,135 @@
...
+type XLStorage []StorageAPI
+
+// Clone creates a copy of storage.
+func (s XLStorage) Clone() XLStorage {
+	s0 := make(XLStorage, len(s))
+	copy(s0, s)
+	return s0
+}
+
+// AppendWriters creates one writer which for every disk of the storage array.
```
fwessels: remove which

cmd/erasure.go
```diff
@@ -0,0 +1,135 @@
...
+// Every writer append data to the corresponding storage disk.
```
fwessels: append --> appends

cmd/erasure.go
```diff
@@ -0,0 +1,135 @@
...
+	for i := range writers {
+		writers[i] = StorageWriter(s[i], volume, path)
+	}
+	return writers
+}
+
+// ErasureEncode encodes the given data and returns the erasure-coded data.
+// It returns an error if the erasure coding failed.
+func (s XLStorage) ErasureEncode(data []byte) ([][]byte, error) {
+	encoder, err := reedsolomon.New(len(s)/2, len(s)/2)
```
fwessels: We should be using xlObjects.dataBlocks and xlObjects.parityBlocks here.

cmd/erasure.go
```diff
@@ -0,0 +1,135 @@
...
+	if err = encoder.Encode(encoded); err != nil {
+		return nil, Errorf("failed to encode data: %v", err)
+	}
+	return encoded, nil
+}
+
+// ErasureDecode decodes the given erasure-coded data.
+// It returns an error if the decoding failed.
+func (s XLStorage) ErasureDecode(data [][]byte) error {
+	decoder, err := reedsolomon.New(len(s)/2, len(s)/2)
```
fwessels: Idem as above.

cmd/erasure.go
```diff
@@ -0,0 +1,135 @@
...
+	if err != nil {
+		return Errorf("failed to create erasure coding: %v", err)
+	}
+	if err = decoder.Reconstruct(data); err != nil {
+		return Errorf("failed to split data: %v", err)
```
fwessels: failed to reconstruct data

cmd/erasure.go
aead: Done

pkg/bitrot/bitrot.go
aead: Done

cmd/erasure.go
aead: Done

cmd/erasure.go
aead: Done

cmd/erasure.go
aead: Done

cmd/erasure.go
aead: Done

cmd/erasure.go
aead: Done

cmd/erasure.go
aead: Done

cmd/erasure.go
aead: Done

cmd/erasure.go
```diff
@@ -0,0 +1,151 @@
...
+type BitrotVerifier struct {
+	bitrot.Hash
+
+	algorithm bitrot.Algorithm
+	key       []byte
+	sum       []byte
+	verified  bool
+}
+
+// Verify returns true iff the computed checksum of the verifier matches the the checksum provided when the verifier
```
fwessels, line 137: typo in iff

fwessels, line 137: the repeated twice

cmd/erasure.go
```diff
@@ -0,0 +1,151 @@
...
+// was created.
+func (v *BitrotVerifier) Verify() bool {
+	v.verified = true
+	return subtle.ConstantTimeCompare(v.Sum(nil), v.sum) == 1
+}
+
+// IsVerified returns true iff Verify was called at least one.
```
fwessels: iff....
one --> once

cmd/erasure.go
```diff
@@ -0,0 +1,151 @@
...
+func (v *BitrotVerifier) IsVerified() bool { return v.verified }
```
fwessels: Wouldn't you want to call Verify() implicitly as part of IsVerified?

cmd/erasure-createfile.go
```diff
@@ -17,128 +17,76 @@
...
-			if bytesWritten == 0 && allowEmpty {
-				blocks = make([][]byte, len(disks))
-				newDisks, rErr = appendFile(disks, volume, path, blocks, hashWriters, writeQuorum)
-				if rErr != nil {
-					return nil, 0, nil, rErr
-				}
-			} // else we have reached EOF after few reads, no need to
-			// add an additional 0bytes at the end.
-			break
+// CreateFile creates a new bitrot encoded file spreed over all available disks. CreateFile will create
```
fwessels: spreed --> spread

cmd/erasure-createfile.go
```diff
@@ -17,128 +17,76 @@
...
+// the file at the given volume and path. It will read from src until an io.EOF occurs. The given algorithm will
+// be used to protected the erasure encoded file. The random reader should return random data (if it's nil the system PRNG will be used).
```
fwessels: to protected --> to protect

cmd/erasure.go
```diff
@@ -0,0 +1,167 @@
...
+		return nil, Errorf("failed to split data: %v", err)
+	}
+	if err = s.erasure.Encode(encoded); err != nil {
+		return nil, Errorf("failed to encode data: %v", err)
+	}
+	return encoded, nil
+}
+
+// ErasureDecodeDataBlocks decodes the given erasure-coded data.
+// It only decodes the data and parity blocks but does not verify them.
```
fwessels: I'd remove "and parity" as that is what we want...

cmd/erasure.go
```diff
@@ -0,0 +1,167 @@
...
+
+// ErasureDecodeDataAndParityBlocks decodes the given erasure-coded data and verifies it.
+// It returns an error if the decoding failed.
+func (s *XLStorage) ErasureDecodeDataAndParityBlocks(data [][]byte) error {
+	if err := s.erasure.Reconstruct(data); err != nil {
+		return Errorf("failed to reconstruct data: %v", err)
+	}
+	if ok, err := s.erasure.Verify(data); !ok || err != nil {
+		if err != nil {
+			return Errorf("failed to verify data: %v", err)
```
fwessels, line 107: why not use else if !ok { } further down below are remove the double comparison for err != nil ?

cmd/erasure-readfile.go
```diff
@@ -17,328 +17,142 @@
...
-	return bytesWritten, nil
-}
-
-// decodeData - decode encoded blocks.
-func decodeData(enBlocks [][]byte, dataBlocks, parityBlocks int) error {
-	// Initialized reedsolomon.
-	rs, err := reedsolomon.New(dataBlocks, parityBlocks)
-	if err != nil {
-		return traceError(err)
+	if err = reduceReadQuorumErrs(errs, []error{}, (s.dataBlocks+s.parityBlocks)/2); err != nil {
```
fwessels: Use xlObjects.readQuorum for read quorum

cmd/erasure.go
aead: `iff` is the short form for `if and only if`... (a common crypto doc abbreviation)  - I can write it out if this is not clear...

aead: same as above

cmd/erasure.go
aead: same..
Done

cmd/erasure.go
aead: Done

cmd/erasure-createfile.go
aead: Done

cmd/erasure-createfile.go
aead: Done

cmd/erasure.go
aead: Okay, sure - done.

cmd/erasure.go
aead: Because there are two different error conditions:
1. for example if we pass two few shards -> ErrTooFewShards
2. the data verification failed because e.g. the data is (after decode) is not correct.
If we remove one check we return the "wrong" error...

We need no !ok because if we enter the first if (`!ok || err != nil`)  but not the second (`err != nil`)
this implies that `ok == false`. 
Or didn't I get your point...

cmd/erasure-readfile.go
aead: Done

cmd/erasure-readfile.go
```diff
@@ -17,328 +17,142 @@
...
-		if !isSuccessDataBlocks(enBlocks, dataBlocks) {
-			// Reconstruct the missing data blocks.
-			if err := decodeData(enBlocks, dataBlocks, parityBlocks); err != nil {
-				return bytesWritten, err
-			}
+		writeLength := blocksize - startOffset
+		if length < writeLength {
+			writeLength = length
 		}
+		n, err := writeDataBlocks(writer, blocks, (s.dataBlocks+s.parityBlocks)/2, startOffset, writeLength)
```
fwessels: We should be using implicit read quorum of `(s.dataBlocks+s.parityBlocks)/2` here
(applies multiple times in this function)

cmd/erasure-readfile.go
```diff
@@ -17,328 +17,142 @@
...
+		}
 	}
+	return nil
+}
 
-	// Verify reconstructed blocks (parity).
-	ok, err := rs.Verify(enBlocks)
-	if err != nil {
-		return traceError(err)
+func erasureReadBlocksConcurrent(disks []StorageAPI, volume, path string, offset int64, blocks [][]byte, verifiers []*BitrotVerifier, errors []error, locks []chan error) {
```
fwessels: Please add comments (even though private function)

cmd/erasure-readfile.go
```diff
@@ -17,328 +17,142 @@
...
+		if errors[i] != nil {
+			disks[i] = OfflineDisk
+			blocks[i] = MissingShard
+		}
 	}
+}
 
-	// Success.
-	return nil
+func erasureReadFromFile(disk StorageAPI, volume, path string, offset int64, buffer []byte, verifier *BitrotVerifier, lock chan<- error) {
```
fwessels: Please add comments (even though private function)

cmd/erasure-readfile.go
```diff
@@ -17,328 +17,142 @@
...
 	}
-	if !ok {
-		// Blocks cannot be reconstructed, corrupted data.
-		err = errors.New("Verification failed after reconstruction, data likely corrupted")
-		return traceError(err)
+	for i := range locks {
+		errors[i] = <-locks[i]
+		if errors[i] != nil {
+			disks[i] = OfflineDisk
+			blocks[i] = MissingShard
```
fwessels: Is this thread safe? (assigning to blocks[i])

cmd/erasure-readfile.go
aead: in this case it must be just DataBlocks

But we are computing the chunkisze here, so we must use the avg of dataBlocks and parityBlocks

cmd/erasure-readfile.go
aead: Added a comment - done.

cmd/erasure-readfile.go
aead: done.

cmd/erasure-readfile.go
aead: done.

cmd/erasure-readfile.go
aead: Never mind - using dataBlocks is the obvious fix for erasure coding if data != parity - done.

cmd/xl-v1-metadata.go
```diff
@@ -50,7 +51,17 @@ func init() {
 	}
 	newPoly1305 := func(key []byte, mode bitrot.Mode) bitrot.Hash {
 		var pKey [32]byte
-		copy(pKey[:], key) // this is safe because bitrot.New will check the len of key
+
+		// If the r part of the key (key[:16]) is zero, the poly1305 multiply has no effect. It's
+		// very unlikely because the key will be randomly generated but we should prevent this.
+		// In this case me map the zero key to the inverse one (11111...).
+		if bytes.Equal(key[:16], pKey[:16]) {
```
harshavardhana: Quite a tricky code to understand the gist would require little more explanation IMHO.

harshavardhana (changes requested):

cmd/xl-v1-metadata.go
aead: Okay maybe something like this?

>>Poly1305 uses the first 16 byte of the key for multiplication. If this part of the key is ever zero the multiply has no effect on the checksum. It is very unlikely that this will ever happen because the key is randomly generated, but even tough we should prevent this from happening. If the part of the key is zero, we now use the inverse - all bit's set. This will guarantee, that the multiplication effects the checksum.

harshavardhana: Yes that would be better.. @aead

cmd/storage-interface.go
```diff
@@ -48,3 +51,48 @@ type StorageAPI interface {
 	// Read all.
 	ReadAll(volume string, path string) (buf []byte, err error)
 }
+
+// storageReader is a io.Reader view of a disk
```
fwessels: a -> an

cmd/storage-interface.go
```diff
@@ -48,3 +51,48 @@ type StorageAPI interface {
...
+func (r *storageReader) Read(p []byte) (n int, err error) {
+	nn, err := r.storage.ReadFile(r.volume, r.path, r.offset, p)
+	n = int(nn)
+	r.offset += nn
+
+	if err == io.ErrUnexpectedEOF && nn > 0 {
+		err = io.EOF
+	}
+	return int(n), err
+
```
fwessels: empty line

cmd/storage-interface.go
```diff
@@ -48,3 +51,48 @@ type StorageAPI interface {
...
+func (r *storageReader) Read(p []byte) (n int, err error) {
+	nn, err := r.storage.ReadFile(r.volume, r.path, r.offset, p)
+	n = int(nn)
+	r.offset += nn
+
+	if err == io.ErrUnexpectedEOF && nn > 0 {
+		err = io.EOF
+	}
+	return int(n), err
```
fwessels: n is already an int ? 
Or you can even `return` ?

cmd/storage-rpc-server-datatypes.go
```diff
@@ -79,11 +81,13 @@ type ReadFileWithVerifyArgs struct {
 	Buffer []byte
 
 	// Algorithm used in bit-rot hash computation.
-	Algo HashAlgo
+	Algo bitrot.Algorithm
+
+	// Stored key used to compute the hash value.
```
fwessels: Add this: (in case needed by algorithm)

cmd/erasure.go
```diff
@@ -0,0 +1,167 @@
...
+	Algorithm bitrot.Algorithm
+	Keys      [][]byte
+	Checksums [][]byte
+}
+
+// XLStorage represents an array of disks.
+// The disks contain erasure coded and bitrot-protected data.
+type XLStorage struct {
+	disks                    []StorageAPI
+	erasure                  reedsolomon.Encoder
```
fwessels, line 42: Consider moving `erasure` to xlObjects and removing XLStorage altogether (too much overlap)

cmd/xl-v1-metadata.go
```diff
@@ -17,20 +17,61 @@
...
+	"golang.org/x/crypto/blake2b"
 )
 
 const (
 	// Erasure related constants.
 	erasureAlgorithmKlauspost = "klauspost/reedsolomon/vandermonde"
 )
 
+// DefaultBitrotAlgorithm is the default algorithm used for bitrot protection
+const DefaultBitrotAlgorithm = bitrot.Poly1305
```
fwessels, line 41: We need to test on ARM whether Poly1305 is faster than SHA256 (which is accelerated)

aead: In general yes - on arm64 github.com/aead/poly1305 will use the reference implementation. If this is slower than SHA-256, we can implement it in assembly: See [sum_arm.s](https://github.com/golang/crypto/blob/master/poly1305/sum_arm.s)

cmd/storage-rpc-server-datatypes.go
aead: Done

cmd/storage-interface.go
aead: Done

cmd/storage-interface.go
aead: Done

cmd/storage-interface.go
aead: Done

cmd/xl-v1-metadata.go
fwessels: The reference implementation will **not** be fast for ARM since the go compiler has some way to go for ARM.
And especially also because the SHA256 uses the crypto instructions to speed up hashing speed to BLAKE2 like performance.
We need to get some benchmark numbers in order to make a decision

cmd/erasure.go
aead: Discussed - Decided to keep it separated because of some corner case problems - revisit in future...

cmd/erasure-readfile.go
```diff
@@ -17,328 +17,151 @@
...
+			disks[i] = OfflineDisk
+			blocks[i] = nil
+		}
 	}
+}
 
-	// Success.
-	return nil
+// erasureReadReadFromFile reads data from the disk to buffer in parallel.
+// It sends the returned error through the lock channel.
```
fwessels, line 158: ReadRead repeated twice.

aead:
Close this since recent changes related to highwayhash and encryption

PR #4679, opened 2017-07-14
build: ditch verifiers on make

brendanashworth:
## Description
This commit ditches running verifiers automatically when just building
the server. It retains the verifiers when running tests.

There is very little point to running the verifiers each time a
developer builds the library but has no intent of running the tests.
They're expensive in time; this commit halves the build time on my
system, from 57 seconds to 29 seconds. This is because verifiers updates
the libraries from GitHub each time, which is slightly wasteful.
Additionally, computing cyclomatic complexity is expensive
computationally and isn't necessary to build the library.

Additionally, this allows the library to be built offline. It no longer
requires internet to run make.

## Motivation and Context
When the internet died I had to build the library

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -67,7 +67,7 @@ coverage: build
 
 # Builds minio locally.
 build:
-	@echo "Building minio to $(PWD)/minio"
+	@echo "Building minio to $(PWD)/minio ..."
 	@CGO_ENABLED=0 go build --ldflags $(BUILD_LDFLAGS) -o $(PWD)/minio
 
 pkg-add:
@@ -87,8 +87,9 @@ pkg-list:
 
 # Builds minio and installs it to $GOPATH/bin.
 install: build
-	@echo "Installing minio at $(GOPATH)/bin/minio"
+	@echo "Installing minio at $(GOPATH)/bin/minio ..."
 	@cp $(PWD)/minio $(GOPATH)/bin/minio
+	@echo "Minio installed successfully. Use 'minio server export' to start."
 
 release: verifiers
 	@MINIO_RELEASE=RELEASE ./buildscripts/build.sh
--- a/Makefile
+++ b/Makefile
@@ -89,7 +89,7 @@ pkg-list:
 install: build
 	@echo "Installing minio at $(GOPATH)/bin/minio ..."
 	@cp $(PWD)/minio $(GOPATH)/bin/minio
-	@echo "Minio installed successfully. Use 'minio server export' to start."
+	@echo "Check 'minio -h' for help."
 
 release: verifiers
 	@MINIO_RELEASE=RELEASE ./buildscripts/build.sh
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ LDFLAGS := $(shell go run buildscripts/gen-ldflags.go)
 
 BUILD_LDFLAGS := '$(LDFLAGS)'
 
-all: install
+all: build
 
 checks:
 	@echo "Check deps"
@@ -45,8 +45,6 @@ cyclo:
 	@${GOPATH}/bin/gocyclo -over 100 cmd
 	@${GOPATH}/bin/gocyclo -over 100 pkg
 
-build: $(UI_ASSETS)
-
 deadcode:
 	@${GOPATH}/bin/deadcode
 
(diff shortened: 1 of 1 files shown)


krishnasrinivas:
It's better if default `make` does everything including tests, linting etc. During development I just do `go install` go see if it compiles and builds the binary. i.e I don't do make as it is time consuming. However just before doing `git push` I do `make` to make sure that everything passes.

brendanashworth:
@krishnasrinivas I'm not sure if thats very idiomatic, [convention](https://www.gnu.org/prep/standards/html_node/Standard-Targets.html#Standard-Targets) would be to run tests with `make test` and just compile the program with `make`, no? To me, it doesn't make sense to just run the verifiers on `make` but not tests, and running tests without `make test` might seem unexpected to someone unfamiliar with the project.

krishnasrinivas:
> @krishnasrinivas I'm not sure if thats very idiomatic, convention would be to run tests with make test and just compile the program with make, no?

https://www.gnu.org/prep/standards/html_node/Standard-Targets.html#Standard-Targets

Then we will need
make - just build
make check - lint, test and build
make install - build and install

@harshavardhana what do you think?

harshavardhana:
> Then we will need
> make - just build
> make check - lint, test and build
> make install - build and install
> 
> @harshavardhana what do you think?

This makes sense for me.. :+1:

brendanashworth:
I can do that. Should we worry about people who might expect `make` to install the binary to $GOPATH/bin? (as it does right now)

krishnasrinivas:
>  Should we worry about people who might expect make to install the binary to $GOPATH/bin? 

`make install` should install the binary to $GOPATH/bin

harshavardhana:
If we indicate to user the path where the binary was built into then we will good. We should say perhaps ` ./minio server export` ;-)

Makefile
```diff
@@ -84,7 +85,10 @@ pkg-remove:
 pkg-list:
 	@$(GOPATH)/bin/govendor list
 
-install: gomake-all
+# Builds minio and installs it to $GOPATH/bin.
+install: build
+	@echo "Installing minio at $(GOPATH)/bin/minio"
+	@cp $(PWD)/minio $(GOPATH)/bin/minio
```
harshavardhana: We can print some friendly message like `Minio compiled successfully  'minio server export' to start.`

harshavardhana (changes requested):

Makefile
krishnasrinivas: users might type `minio server export` in the source directory which is not preferred. Telling users to do `minio server ~/export` will also not be correct for windows users.

We could just leave it at this. So users will type "minio -h" for help

brendanashworth: Perhaps it could say "Minio installed successfully. Use 'minio -h' for usage."?

harshavardhana: We can say minio -h

brendanashworth: So like this?
```
$ make install
Building minio to /.../Go/src/github.com/minio/minio/minio ...
Installing minio at /.../Go/bin/minio ...

NAME:
  minio - Cloud Storage Server.

DESCRIPTION:
  Minio is an Amazon S3 compatible object storage server. Use it to store photos, videos, VMs, containers, log files, or any blob of data as objects.

USAGE:
  minio [FLAGS] COMMAND [ARGS...]

COMMANDS:
  server   Start object storage server.
  version  Print version.
  update   Check for a new software update.
  gateway  Start object storage gateway.
  
FLAGS:
  --config-dir value, -C value  Path to configuration directory. (default: "/Users/brendanashworth/.minio")
  --quiet                       Disable startup information.
  --help, -h                    Show help.
  
VERSION:
  2017-07-15T00:46:01Z
```

krishnasrinivas: no, like this:
```
$ make install
Building minio to /.../Go/src/github.com/minio/minio/minio ...
Installing minio at /.../Go/bin/minio ...
Check "minio -h" for help
```

brendanashworth: Done!

balamurugana (approved):
This simplifies my two commands (make && make test) to one (make check)

Makefile
```diff
@@ -55,7 +53,9 @@ spelling:
 	@${GOPATH}/bin/misspell -error `find pkg/`
 	@${GOPATH}/bin/misspell -error `find docs/`
 
-test: build
+# Builds minio, runs the verifiers then runs the tests.
+check: test
+test: verifiers build
```
krishnasrinivas, line 58: verifiers can also be under `check`

harshavardhana (approved):

PR #4680, opened 2017-07-15
restirct max size of http header and user metadata

### LINKED ISSUE #4634: PutObject user metadata size should follow AWS S3 restriction
## Expected Behavior
The PUT request header is limited to 8 KB in size. Within the PUT request header, the user-defined metadata is limited to 2 KB in size. The size of user-defined metadata is measured by taking the sum of the number of bytes in the UTF-8 encoding of each key and value.

AWS S3 returns 

```
X-Amz-Meta-993: a
X-Amz-Meta-994: a
X-Amz-Meta-995: a
X-Amz-Meta-996: a
X-Amz-Meta-997: a
X-Amz-Meta-998: a
X-Amz-Meta-999: a
Accept-Encoding: gzip

HTTP/1.1 400 Bad Request
Connection: close
Content-Length: 0
Date: Thu, 06 Jul 17 02:36:35 GMT
X-Amz-Id-2: Rk/GF494BZyBF9VOBdfJpfu2adqkFkyoVMAQJWfu8YYXCSdslP8eG7dqTa1h69rnpWLrdIzZ/60EkbvNMHQfZvgSxB9P1Av9
X-Amz-Request-Id: 53AADA523CAC335C
---------END-HTTP---------
2017/07/05 19:36:35 400 Bad Request
```

## Current Behavior
We do not validate this restriction, it is allowed to create a PUT object with large number of headers and there is no such limit on the incoming request headers. 

## Possible Solution
PutObjectHandler should handle this properly. 

Set large set of request headers using minio-go. 

```go
 package main

 import (
         "fmt"
         "log"
         "os"

         "github.com/minio/minio-go"
 )

 func main() {
         s3Client, err := minio.New("localhost:9000", "USWUXHGYZQYFYFFIT3RE", "MOJRH0mkL1IPauahWITSVvyDrQbEEIwljvmxdq03", false)
         if err != nil {
                 log.Fatalln(err)
         }

         object, err := os.Open("/etc/issue")
         if err != nil {
                 log.Fatalln(err)
         }
         defer object.Close()

         metadata := make(map[string][]string)
         for i := 0; i < 2049; i++ {
                 metadata[fmt.Sprintf("X-Amz-Meta-%d", i)] = []string{"a"}
         }
         n, err := s3Client.PutObjectWithMetadata("my-bucketname", "my-objectname", object, metadata, nil)
         if err != nil {
                 log.Fatalln(err)
         }
         log.Println("Uploaded", "my-objectname", " of size: ", n, "Successfully.")
 }
```

## Context
S3 compatibility. 

## Your Environment
master branch.

aead:
## Description

S3 only allows http headers with a size of 8 KB and user-defined metadata
with a size of 2 KB. This change adds a new API error and returns this
error to clients which sends to large http requests.

## Motivation and Context
Fixes #4634 

## How Has This Been Tested?
Add test cases for PutObjectHandler

## Types of changes

## Checklist:

### DIFF
--- a/cmd/api-errors.go
+++ b/cmd/api-errors.go
@@ -116,6 +116,7 @@ const (
 	ErrInvalidDuration
 	ErrNotSupported
 	ErrBucketAlreadyExists
+	ErrMetadataTooLarge
 	// Add new error codes here.
 
 	// Bucket notification related errors.
@@ -630,7 +631,11 @@ var errorCodeResponse = map[APIErrorCode]APIError{
 		Description:    "Cannot respond to plain-text request from TLS-encrypted server",
 		HTTPStatusCode: http.StatusBadRequest,
 	},
-
+	ErrMetadataTooLarge: {
+		Code:           "InvalidArgument",
+		Description:    "Your metadata headers exceed the maximum allowed metadata size.",
+		HTTPStatusCode: http.StatusBadRequest,
+	},
 	// Add your error structure here.
 }
 
--- a/cmd/generic-handlers.go
+++ b/cmd/generic-handlers.go
@@ -62,6 +62,52 @@ func (h requestSizeLimitHandler) ServeHTTP(w http.ResponseWriter, r *http.Reques
 	h.handler.ServeHTTP(w, r)
 }
 
+const (
+	// Maximum size for http headers - See: https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html
+	maxHeaderSize = 8 * 1024
+	// Maximum size for user-defined metadata - See: https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html
+	maxUserDataSize = 2 * 1024
+)
+
+type requestHeaderSizeLimitHandler struct {
+	http.Handler
+}
+
+func setRequestHeaderSizeLimitHandler(h http.Handler) http.Handler {
+	return requestHeaderSizeLimitHandler{h}
+}
+
+// ServeHTTP restricts the size of the http header to 8 KB and the size
+// of the user-defined metadata to 2 KB.
+func (h requestHeaderSizeLimitHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if isHTTPHeaderSizeTooLarge(r.Header) {
+		writeErrorResponse(w, ErrMetadataTooLarge, r.URL)
+		return
+	}
+	h.Handler.ServeHTTP(w, r)
+}
+
+// isHTTPHeaderSizeTooLarge returns true if the provided
+// header is larger than 8 KB or the user-defined metadata
+// is larger than 2 KB.
+func isHTTPHeaderSizeTooLarge(header http.Header) bool {
+	var size, usersize int
+	for key := range header {
+		length := len(key) + len(header.Get(key))
+		size += length
+		for _, prefix := range userMetadataKeyPrefixes {
+			if strings.HasPrefix(key, prefix) {
+				usersize += length
+				break
+			}
+		}
+		if usersize > maxUserDataSize || size > maxHeaderSize {
+			return true
+		}
+	}
+	return false
+}
+
 // Reserved bucket.
 const (
 	minioReservedBucket     = "minio"
--- a/cmd/handler-utils.go
+++ b/cmd/handler-utils.go
@@ -97,6 +97,14 @@ func path2BucketAndObject(path string) (bucket, object string) {
 	return bucket, object
 }
 
+// userMetadataKeyPrefixes contains the prefixes of used-defined metadata keys.
+// All values stored with a key starting with one of the following prefixes
+// must be extracted from the header.
+var userMetadataKeyPrefixes = []string{
+	"X-Amz-Meta-",
+	"X-Minio-Meta-",
+}
+
 // extractMetadataFromHeader extracts metadata from HTTP header.
 func extractMetadataFromHeader(header http.Header) (map[string]string, error) {
 	if header == nil {
@@ -119,11 +127,11 @@ func extractMetadataFromHeader(header http.Header) (map[string]string, error) {
 		if key != http.CanonicalHeaderKey(key) {
 			return nil, traceError(errInvalidArgument)
 		}
-		if strings.HasPrefix(key, "X-Amz-Meta-") {
-			metadata[key] = header.Get(key)
-		}
-		if strings.HasPrefix(key, "X-Minio-Meta-") {
-			metadata[key] = header.Get(key)
+		for _, prefix := range userMetadataKeyPrefixes {
+			if strings.HasPrefix(key, prefix) {
+				metadata[key] = header.Get(key)
+				break
+			}
 		}
 	}
 	return metadata, nil
--- a/cmd/generic-handlers_test.go
+++ b/cmd/generic-handlers_test.go
@@ -18,6 +18,7 @@ package cmd
 
 import (
 	"net/http"
+	"strconv"
 	"testing"
 )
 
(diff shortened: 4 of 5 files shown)


cmd/api-errors.go
```diff
@@ -561,6 +562,11 @@ var errorCodeResponse = map[APIErrorCode]APIError{
 		Description:    "Range specified is not valid for source object",
 		HTTPStatusCode: http.StatusBadRequest,
 	},
+	ErrMetadataTooLarge: {
+		Code:           "InvalidArgument",
+		Description:    "Your metadata headers exceed the maximum allowed metadata size.",
```
krisis: If we intend to have the string representation of `errMetadataTooLarge` same as `ErrMetadataTooLarge`'s description, then we should do the following,
```go
ErrMetadataTooLarge: {
	Description:    errMetadataTooLarge.String(),
},
```

cmd/handler-utils.go
```diff
@@ -114,16 +114,24 @@ func extractMetadataFromHeader(header http.Header) (map[string]string, error) {
 			metadata[supportedHeader] = header.Get(supportedHeader)
 		}
 	}
+	const MaxHeaderSize = 8 * 1024
```
harshavardhana: We should define the constants here.. ideally these should be global. With more descriptive naming 

```
putObjectMaxHeaderSize   = 8 * humanize.KiByte
putObjectMaxUserDataSize = 2 * humanize.KiByte
```

harshavardhana: A relevant AWS S3 docs reference URL should also be commented to indicate this spec restriction.

krisis (approved):

harshavardhana:
Also there is a typo in the commit message `s/restirct/restrict`

krishnasrinivas (commented):
@aead as discussed in person header size restrictions applies to all the calls, hence headerSanityCheck() should be done for all the calls

harshavardhana:
If it's for all calls we should add this as part of generic-handlers.go - thoughts?

krishnasrinivas:
> If it's for all calls we should add this as part of generic-handlers.go - thoughts?

that would be better

balamurugana (commented):
is it good idea to use `MaxHeaderBytes` of HTTP server?  Ref: https://golang.org/pkg/net/http/#Server

krishnasrinivas:
> is it good idea to use MaxHeaderBytes of HTTP server? Ref: https://golang.org/pkg/net/http/#Server

We discussed about it, but the issue with it is:
1) we wont be able to reply S3 XML error
2) @aead  mentioned Go http server calculates header length limit differently than S3 spec. For ex. S3 spec leaves out `:`

harshavardhana:
Why will not be able to reply S3 XML error can you elaborate ? 

We already do this with

- pathVailidity 
- unsupportedResources 
- timeValidity

aead:
@balamurugana 
I had the same idea about `MaxHeaderBytes` but the "problem" is that AWS-doc says:
> The size of the header is computed as the sum of lengths of all keys and values encoded as UTF-8 strings. 

It's accentually no problem if we allow slightly different sizes as AWS because it doesn't matter if the client stores 2KB or e.g. 2KB+20Bytes but I think we should follow AWS-spec. Otherwise a user may send exactly 8KB and we may reject the request because Go computes the header size slightly different...

My current approach is to check the header size for every call (GET, PUT, HEAD,....) and return an error if the header is to large... 
**Any objections?**

krishnasrinivas:
> Why will not be able to reply S3 XML error can you elaborate ?

My response was to @balamurugana suggestion that it we can use http.Server.MaxHeaderBytes

balamurugana:
@aead @krishnasrinivas understood the limitation of using `MaxHeaderBytes`

cmd/generic-handlers.go
```diff
@@ -65,6 +65,48 @@ func (h requestSizeLimitHandler) ServeHTTP(w http.ResponseWriter, r *http.Reques
...
+}
+func (h requestHeaderSizeLimitHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if err := h.checkHTTPHeaderSize(r.Header); err != nil {
+		writeErrorResponse(w, toAPIErrorCode(err), r.URL)
+		return
+	}
+	h.handler.ServeHTTP(w, r)
+}
+
+func (h requestHeaderSizeLimitHandler) checkHTTPHeaderSize(header http.Header) error {
```
krishnasrinivas: checkHTTPHeaderSize need not be a method on requestHeaderSizeLimitHandler, it can be an independent func.
If you rename it to isHTTPHeaderSizeTooLarge, then it can just return `bool` and you dont need to introduce the error errMetadataTooLarge

cmd/generic-handlers.go
```diff
@@ -65,6 +65,44 @@ func (h requestSizeLimitHandler) ServeHTTP(w http.ResponseWriter, r *http.Reques
...
+}
+func (h requestHeaderSizeLimitHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if isHTTPHeaderSizeTooLarge(r.Header) {
+		writeErrorResponse(w, ErrMetadataTooLarge, r.URL)
+		return
+	}
+	h.Handler.ServeHTTP(w, r)
+}
+
+func isHTTPHeaderSizeTooLarge(header http.Header) bool {
```
krishnasrinivas: Should have pointed this out earlier, this function needs unit tests

aead: done

cmd/generic-handlers.go
```diff
@@ -62,6 +62,44 @@ func (h requestSizeLimitHandler) ServeHTTP(w http.ResponseWriter, r *http.Reques
 	h.handler.ServeHTTP(w, r)
 }
 
+const (
+	maxHeaderSize   = 8 * 1024
```
harshavardhana: Write comments and point to the relevant AWS S3 documentation.

cmd/generic-handlers.go
```diff
@@ -62,6 +62,44 @@ func (h requestSizeLimitHandler) ServeHTTP(w http.ResponseWriter, r *http.Reques
...
+
+type requestHeaderSizeLimitHandler struct {
+	http.Handler
+}
+
+func setRequestHeaderSizeLimitHandler(h http.Handler) http.Handler {
+	return requestHeaderSizeLimitHandler{h}
+}
+func (h requestHeaderSizeLimitHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if isHTTPHeaderSizeTooLarge(r.Header) {
```
harshavardhana: Comments..

cmd/handler-utils.go
```diff
@@ -97,6 +97,11 @@ func path2BucketAndObject(path string) (bucket, object string) {
 	return bucket, object
 }
 
+var userMetadataKeyPrefixes = []string{
```
harshavardhana: A comment.. would be necessary for a global variable. If this is not used anywhere define them inside the extractMetadataFromHeader() .

harshavardhana (changes requested):

cmd/generic-handlers.go
aead: done

cmd/handler-utils.go
aead: done

cmd/generic-handlers.go
aead: done

harshavardhana (approved):

harshavardhana:
Can you squash the commits @aead ?

aead:
done

deekoder:
Waiting for codecov to complete.

PR #4683, opened 2017-07-17
Prevent unnecessary verification of parity blocks while reading

fwessels:
While reading an erasure coded file, in case any of the data blocks is missing and we need to reconstruct data, we are unnecessarily verifying the integrity of the parity blocks.

As bit-rot protection shields us from any changes in either the data or parity blocks, it is pointless to verify the integrity of any existing (not reconstructed) parity blocks as these cannot be incorrect (or we would have a bug in the galois field arithmetic or matrix manipulation...). 

Moreover we are not using the parity blocks while reading an erasure coded file, as the data blocks represent the contents of the file.
 
Note that this is not the case for healing where we would also want to reconstruct any missing parity blocks.

Furthermore added a TODO to add a data block reconstruction only mode to Reed Solomon as a further optimization.

## Motivation and Context
Performance optimization by preventing unnecessary verification.

## Types of changes

## Checklist:


aead:
#4758 address this too. 
@fwessels Should we try to merge this because the code will be removed if we merge #4758

cmd/erasure_test.go
```diff
@@ -124,26 +124,54 @@ func TestErasureDecode(t *testing.T) {
...
+				if err != nil && testCase.shouldPass {
+					t.Errorf("Test %d: Expected to pass by failed instead with %s", i+1, err)
+				}
+
+				// Validate if decoded data is what we expected.
+				if bytes.Equal(decodedDataWriter.Bytes(), data) != testCase.shouldPass {
+					err := errUnexpected
+					t.Errorf("Test %d: Expected to pass by failed instead %s", i+1, err)
+				}
 			}
```
donatello, line 150: Nice style of using blocks here to be able to cleanly redeclare and reuse variable names. I will use this in tests in the future! :+1:

donatello (approved):
LGTM - please take decision to merge after considering @aead's comment.

aead (approved):

deekoder:
@fwessels waiting on more info on codecoverage on erasure_readfile.go

cmd/erasure-readfile.go
```diff
@@ -328,14 +340,23 @@ func decodeData(enBlocks [][]byte, dataBlocks, parityBlocks int) error {
...
+		// - either one (or more) of the data blocks is corrupted,
+		// - or one (or more) of the parity blocks is corrupted.
+		//
+		// Unfortunately it is not possible to state which one of the two possibilities
+		// it is, so while the data may still be intact, we have no way of saying this
+		// definitively.
+		//
+		// Note that in the presence of bit-rot protection (as is present in Minio)
+		// you really should never see this message.
+		err = errors.New("Unable to verify parity blocks, either existing data or parity blocks corrupted")
```
harshavardhana: We can perhaps add a test case for this.. just for the failure case of Verify() ? @fwessels

fwessels: Discuss this and decided to remove the Verification altogether since in our case we are protected by bit rot protection.

PR #4689, opened 2017-07-18
[DO NOT MERGE YET] Move to go1.8

vadmeste:
## Description
Support of go 1.8

Once all issues under POD move to go 1.8 are closed, we can merge this PR. 

## Motivation and Context
POD move to go 1.8

## How Has This Been Tested?
go test + Manully

## Types of changes

## Checklist:


cmd/posix-prepare-path.go
```diff
@@ -16,41 +16,7 @@
...
 package cmd
 
-import (
-	"path/filepath"
-	"runtime"
-	"strings"
-)
-
 // preparePath rewrites path to handle any OS specific details.
 func preparePath(path string) string {
```
fwessels: Remove function altogether?

vadmeste: > Remove function altogether?

Done

cmd/endpoint_test.go
```diff
@@ -232,7 +232,7 @@ func TestCreateEndpoints(t *testing.T) {
 		expectedSetupType  SetupType
 		expectedErr        error
 	}{
-		{"localhost", []string{}, "", EndpointList{}, -1, fmt.Errorf("missing port in address localhost")},
+		{"localhost", []string{}, "", EndpointList{}, -1, fmt.Errorf("address localhost: missing port in address")},
```
harshavardhana, line 223: We should remove these style of error checks.. in future.

pkg/http/quirkconn.go
```diff
@@ -0,0 +1,45 @@
...
+// following the go bug reported here https://github.com/golang/go/issues/21133.
+// Once the bug will be fixed, we can remove this structure and replaces it with
+// the standard net.Conn
+type QuirkConn struct {
+	net.Conn
+	hadReadDeadlineInPast int32 // atomic
+}
+
+// SetReadDeadline - implements a workaround of SetReadDeadline go bug
+func (q *QuirkConn) SetReadDeadline(t time.Time) error {
```
harshavardhana: We should add a test for this.. Looks like currently this is not used. Perhaps you can also rename the file as ```conn_bug_21133.go```  and makes it easy to delete.

harshavardhana (changes requested):

pkg/http/quirkconn.go
vadmeste: @harshavardhana yeah we are right, it was a mistake and I fixed it..

By the way, I really tried to make QuirkConn fully transparent.. but it wasn't possible, bufConn should be aware about the read deadline problem and should check by itself if it is allowed to set read deadline to avoid causing the go 1.8 bug.

The reason is that both golang API and bufConn calls SetReadDeadline() and we really need to synchronize those calls due to the go 1.8 bug.

Issue #4620, opened 2017-07-03
JSON parsing for xl.json gjson lags behind jsoninter


harshavardhana:
Comparing standard library v/s jsoniter 
```
$ benchcmp /tmp/std.txt /tmp/jsoniter.txt 
benchmark                          old ns/op     new ns/op     delta
BenchmarkParseUnmarshal10-4        74356         16404         -77.94%
BenchmarkParseUnmarshal100-4       572028        113695        -80.12%
BenchmarkParseUnmarshal500-4       2760211       500776        -81.86%
BenchmarkParseUnmarshal1000-4      5532915       985919        -82.18%
BenchmarkParseUnmarshal5000-4      27897519      7296728       -73.84%
BenchmarkParseUnmarshal10000-4     55900504      14337416      -74.35%
```

Comparing gjson v/s jsoniter 
```
benchcmp /tmp/gjson.txt /tmp/jsoniter.txt 
benchmark                          old ns/op     new ns/op     delta
BenchmarkParseUnmarshal10-4        57561         16404         -71.50%
BenchmarkParseUnmarshal100-4       458900        113695        -75.22%
BenchmarkParseUnmarshal500-4       2159754       500776        -76.81%
BenchmarkParseUnmarshal1000-4      4335041       985919        -77.26%
BenchmarkParseUnmarshal5000-4      21955116      7296728       -66.77%
BenchmarkParseUnmarshal10000-4     43759317      14337416      -67.24%
```

## Expected Behavior
Use a high performant json parser. This is needed because we are json heavy and we need to only use the best parser available to avoid any form of latencies that can incur due to json parsing. 

## Current Behavior
Current usage of gjson is performing lower in benchmarks compared to https://github.com/json-iterator/go which is 100% compatible with the go standard library. 

## Steps to Reproduce (for bugs)
Download https://gist.github.com/harshavardhana/51126d9df806c5fcf64c5cbd01e77129
```
PARSER=gjson go test -run=NONE -bench .
```

## Context
Performance improvements, look for better things on a timely basis. 

## Your Environment
Current master code.

deekoder:
Won't fix for now. Will revisit later in a more comprehensive manner.

Issue #4388, opened 2017-05-22
Replace pkg/quick with gjson library


deekoder:
- Also review sync and cksum support.

fwessels:
Things to sort out:
- [ ] performance impact of writing with O_SYNC 
- [ ] Figure out how "fingerprint" the contents of xl.json to detect external tampering of xl.json

kannappanr:
Closing it as this will not be needed for now.

PR #4692, opened 2017-07-19
(WIP) use json-iterator for performance

brendanashworth:
These sets of commits adds the [json-iterator library](https://github.com/json-iterator/go) for performance when reading JSON files in the server.

It adds the parser for both fs and xl. Benchmarks are added for xl and show significant improvements in `ReadAllXLMetadata` (even with IO performed), as 25% faster. Equivalent improvement is expected in fs. No real change was made for `writeXLMetadata` (+ fs as well) because it was slower than `encoding/json` stdlib.

Each commit has a full description for the change included.

Fixes: https://github.com/minio/minio/issues/4620
cc @harshavardhana @krishnasrinivas 

https://github.com/minio/minio/issues/4388 needs to be discussed. As of this PR, GJSON is very nearly completely removed. The entire library can be removed from `vendor/` as of a few small changes to the config files. However, we check the config file for duplicate keys (`{ a: 1, a: 2}`), and error, even though most other JSON libraries do not error. [We can implement this ourselves](https://stackoverflow.com/a/38645297/2708970), but we could also consider dropping it.

### DIFF
--- a/cmd/fs-v1-metadata.go
+++ b/cmd/fs-v1-metadata.go
@@ -26,9 +26,9 @@ import (
 	"strings"
 	"time"
 
+	"github.com/json-iterator/go"
 	"github.com/minio/minio/pkg/lock"
 	"github.com/minio/minio/pkg/mimedb"
-	"github.com/tidwall/gjson"
 )
 
 // FS format, and object metadata.
@@ -172,43 +172,6 @@ func (m *fsMetaV1) WriteTo(lk *lock.LockedFile) (n int64, err error) {
 	return int64(len(metadataBytes)), nil
 }
 
-func parseFSVersion(fsMetaBuf []byte) string {
-	return gjson.GetBytes(fsMetaBuf, "version").String()
-}
-
-func parseFSFormat(fsMetaBuf []byte) string {
-	return gjson.GetBytes(fsMetaBuf, "format").String()
-}
-
-func parseFSRelease(fsMetaBuf []byte) string {
-	return gjson.GetBytes(fsMetaBuf, "minio.release").String()
-}
-
-func parseFSMetaMap(fsMetaBuf []byte) map[string]string {
-	// Get xlMetaV1.Meta map.
-	metaMapResult := gjson.GetBytes(fsMetaBuf, "meta").Map()
-	metaMap := make(map[string]string)
-	for key, valResult := range metaMapResult {
-		metaMap[key] = valResult.String()
-	}
-	return metaMap
-}
-
-func parseFSParts(fsMetaBuf []byte) []objectPartInfo {
-	// Parse the FS Parts.
-	partsResult := gjson.GetBytes(fsMetaBuf, "parts").Array()
-	partInfo := make([]objectPartInfo, len(partsResult))
-	for i, p := range partsResult {
-		info := objectPartInfo{}
-		info.Number = int(p.Get("number").Int())
-		info.Name = p.Get("name").String()
-		info.ETag = p.Get("etag").String()
-		info.Size = p.Get("size").Int()
-		partInfo[i] = info
-	}
-	return partInfo
-}
-
 func (m *fsMetaV1) ReadFrom(lk *lock.LockedFile) (n int64, err error) {
 	var fsMetaBuf []byte
 	fi, err := lk.Stat()
@@ -225,26 +188,16 @@ func (m *fsMetaV1) ReadFrom(lk *lock.LockedFile) (n int64, err error) {
 		return 0, traceError(io.EOF)
 	}
 
-	// obtain version.
-	m.Version = parseFSVersion(fsMetaBuf)
-
-	// obtain format.
-	m.Format = parseFSFormat(fsMetaBuf)
+	err = jsoniter.Unmarshal(fsMetaBuf, &m)
+	if err != nil {
+		return 0, traceError(err)
+	}
 
 	// Verify if the format is valid, return corrupted format
 	// for unrecognized formats.
-	if !isFSMetaValid(m.Version, m.Format) {
+	if !m.IsValid() {
 		return 0, traceError(errCorruptedFormat)
 	}
-
-	// obtain metadata.
-	m.Meta = parseFSMetaMap(fsMetaBuf)
-
-	// obtain parts info list.
-	m.Parts = parseFSParts(fsMetaBuf)
-
-	// obtain minio release date.
-	m.Minio.Release = parseFSRelease(fsMetaBuf)
 
 	// Success.
 	return int64(len(fsMetaBuf)), nil
--- a/cmd/fs-v1.go
+++ b/cmd/fs-v1.go
@@ -22,7 +22,6 @@ import (
 	"fmt"
 	"hash"
 	"io"
-	"io/ioutil"
 	"os"
 	"path"
 	"path/filepath"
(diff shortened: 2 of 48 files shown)


harshavardhana:
It would be good to show some benchmarks on the improvements it brings.

brendanashworth:
@harshavardhana [here](https://github.com/brendanashworth/minio/commit/7654faa8e897ccb148220d9331d0503fb1d56acf) are the results from the benchmarks I added, which also run IO and some other things. The [original issue](https://github.com/minio/minio/issues/4620) had some that benched just the library; that shows the real improvement on just JSON parsing, instead of the metadata call itself.

```
benchmark                          old ns/op     new ns/op     delta
BenchmarkReadAllXLMetadata-4       483240        355385        -26.46%
BenchmarkWriteSameXLMetadata-4     7243787       6233366       -13.95%

benchmark                          old allocs     new allocs     delta
BenchmarkReadAllXLMetadata-4       595            675            +13.45%
BenchmarkWriteSameXLMetadata-4     571            571            +0.00%

benchmark                          old bytes     new bytes     delta
BenchmarkReadAllXLMetadata-4       101541        69791         -31.27%
BenchmarkWriteSameXLMetadata-4     16573475      16563761      -0.06%
```

harshavardhana:
The general benchmark here is fine but we need to add cases where we see progressive improvement. Look at original issue where the benchmarking was achieved by generating 100 to 10000 parts `xl.json`.  A `xl.json` which holds around 10,000 parts is actually around 2MB+ in current format. With encryption changes coming in this will become bigger perhaps around 5MB.

harshavardhana:
Also with gjson we used to only extract a subset of json and navigate by keys. We need to see for example readXLMetaStat takes a hit if gjson is still better under those conditions.

brendanashworth:
@harshavardhana I just did some benchmarks. GJSON still has the upper edge when retrieving a non-majority part of a JSON file, i.e., getting `Meta` and `Stat` when it has 1000+ parts loaded (=2k JSON objects outside of Meta and Stat). In this case, GJSON outperforms loading the whole struct by 2-3x. When reading the whole JSON file however, json-iterator is about 40% faster than GJSON.

**edit**: to be academic and sound smart. GJSON can read specific values with a complexity of `O(m * log(n))`, where `n` is the size of the entire JSON file and `m` is the size of the requested value. json-iterator does not have that functionality, thus can only read the whole file. This has a complexity of `O(n)`. So, GJSON is faster when `m` is considerably smaller than `n`, but otherwise json-iterator is faster.

```
benchmark                           old ns/op     new ns/op     delta
BenchmarkReadXLMetaParts2-4         30037         40226         +33.92%
BenchmarkReadXLMetaParts10-4        47714         54137         +13.46%
BenchmarkReadXLMetaParts500-4       964007        768341        -20.30%
BenchmarkReadXLMetaParts10000-4     21762325      18470214      -15.13%
BenchmarkReadXLMetaStat2-4          35676         39899         +11.84%
BenchmarkReadXLMetaStat10-4         39352         58671         +49.09%
BenchmarkReadXLMetaStat500-4        275451        770202        +179.61%
BenchmarkReadXLMetaStat10000-4      4960938       18971903      +282.43%
BenchmarkReadAllXLMetadata-4        623469        374010        -40.01%
BenchmarkWriteSameXLMetadata-4      8252591       6848950       -17.01%
```

Should I just port over the change when reading the whole JSON file, and keep GJSON for reading the smaller portions? Or, should we just take the performance hit because adding the json-iterator library into the repository doesn't outweigh the 40% loss?

harshavardhana:
> Should I just port over the change when reading the whole JSON file, and keep GJSON for reading the smaller portions? Or, should we just take the performance hit because adding the json-iterator library into the repository doesn't outweigh the 40% loss?

We should port over everything or keep gjson we need to see the frequency of operations where gjson is currently useful and where jsoniter is faster i.e what it optimizes and which code path is more frequently hit. 

We need to perhaps do one more an outline of the calls benefited with this change and list the APIs down which would perform better or slower.  For an example

Say ListObjects() was only interested in ReadXLMetaStat() whereas GetObject() is interested in the entire `xl.json`

brendanashworth:
Here's a summary of some of the impacts:

- [`xl.ListObjects`](https://github.com/minio/minio/blob/46897b11004a16e3817d1611d1e1d3927a4e5776/cmd/xl-v1-list-objects.go#L125) will take a [significant hit](https://github.com/minio/minio/blob/46897b11004a16e3817d1611d1e1d3927a4e5776/cmd/xl-v1-list-objects.go#L89)
- [`xl.ListObjectParts`](https://github.com/minio/minio/blob/ec5293ce29ef03726c7174d9e3304bc156259c4e/cmd/xl-v1-multipart.go#L840) will take a [hit on small files, will improve for large files](https://github.com/minio/minio/blob/ec5293ce29ef03726c7174d9e3304bc156259c4e/cmd/xl-v1-multipart.go#L777)
- [`xl.GetObjectInfo`](https://github.com/minio/minio/blob/ec5293ce29ef03726c7174d9e3304bc156259c4e/cmd/xl-v1-object.go#L306) will take a [somewhat significant hit](https://github.com/minio/minio/blob/ec5293ce29ef03726c7174d9e3304bc156259c4e/cmd/xl-v1-object.go#L321)
- [`xl.CopyObject`](https://github.com/minio/minio/blob/ec5293ce29ef03726c7174d9e3304bc156259c4e/cmd/xl-v1-object.go#L62) will receive a [relatively significant improvement](https://github.com/minio/minio/blob/ec5293ce29ef03726c7174d9e3304bc156259c4e/cmd/xl-v1-object.go#L64)
- [`xl.GetObject`](https://github.com/minio/minio/blob/ec5293ce29ef03726c7174d9e3304bc156259c4e/cmd/xl-v1-object.go#L137) will receive a [relatively significant improvement](https://github.com/minio/minio/blob/ec5293ce29ef03726c7174d9e3304bc156259c4e/cmd/xl-v1-object.go#L153)
- [`xl.PutObjectPart`](https://github.com/minio/minio/blob/ec5293ce29ef03726c7174d9e3304bc156259c4e/cmd/xl-v1-multipart.go#L571) will receive a [relatively significant improvement](https://github.com/minio/minio/blob/ec5293ce29ef03726c7174d9e3304bc156259c4e/cmd/xl-v1-multipart.go#L590)
- [`xl.CompleteMultipartUpload`](https://github.com/minio/minio/blob/ec5293ce29ef03726c7174d9e3304bc156259c4e/cmd/xl-v1-multipart.go#L865) will receive a [relatively significant improvement](https://github.com/minio/minio/blob/ec5293ce29ef03726c7174d9e3304bc156259c4e/cmd/xl-v1-multipart.go#L900)

FS will receive significant performance improvements across the board, as almost all API calls load the whole JSON file. The only exception is [`fs.ListObjects`](https://github.com/minio/minio/blob/ec5293ce29ef03726c7174d9e3304bc156259c4e/cmd/fs-v1.go#L770), which will take a [relatively significant hit](https://github.com/minio/minio/blob/ec5293ce29ef03726c7174d9e3304bc156259c4e/cmd/fs-v1.go#L817).

I'm going to run some benchmarks to see just how these performance changes play out, hold on.

donatello:
Is this still WIP or can we consider it for review? @harshavardhana @brendanashworth

brendanashworth:
@donatello still a work in progress, please do not review. This is blocked by https://github.com/json-iterator/go/issues/140.

deekoder:
we have decided to revisit this later.

Issue #4564, opened 2017-06-19
ListMultipartUploads and ListObjectParts changes in FS backend


krishnasrinivas:
As discussed in today's meeting we are moving to a simpler backend for multipart related files.

ListMultupartUploads and ListObjectParts will return empty listing as response.

Backend will look more like GCS' multipart backend.

This will rid a lot of complicated code from our code base.

@deekoder can you assign this to me

krishnasrinivas:
@krisis reassigning this to you as you are looking into this.

PR #4694, opened 2017-07-19
ListMultipartUploads, ListObjectParts return empty response

krisis:
## Description

This change makes ListMultipartUploads and ListObjectParts return empty response. It _doesn't_ change the backend structure.
## Motivation and Context

See #4564 
## How Has This Been Tested?

Using `mc ls -I [-r]` and existing unit tests.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/fs-v1-multipart.go
+++ b/cmd/fs-v1-multipart.go
@@ -31,6 +31,11 @@ import (
 	"github.com/minio/sha256-simd"
 )
 
+const (
+	fsMultipartExpiry        = time.Hour * 24 * 14
+	fsMultipartCleanupPeriod = time.Hour * 24
+)
+
 // Returns if the prefix is a multipart upload.
 func (fs fsObjects) isMultipartUpload(bucket, prefix string) bool {
 	uploadsIDPath := pathJoin(fs.fsPath, bucket, prefix, uploadsJSONFile)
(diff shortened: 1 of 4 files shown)


krishnasrinivas (approved):

krishnasrinivas:
@krisis one more thing we need to take care of is cleaning up files older than 2 weeks just like gcs.

krisis:
> @krisis one more thing we need to take care of is cleaning up files older than 2 weeks just like gcs.

Is it required as part of this PR?

fwessels (approved):
LGTM

harshavardhana (commented):
Need to know more about this, why are we removing this code without migrating to `.minio.sys` format? is it expected to be removed in this manner without the backend change?

krisis:
> Need to know more about this, why are we removing this code without migrating to .minio.sys format? is it expected to be removed in this manner without the backend change?

@harshavardhana, we decided to do this change before (and independent of) FS backend format changes like possibly using extended attributes for storing user-metadata and bucket policy and the multipart backend to avoid relying for fcntl(3) locks for coordinating among many minio servers in a shared mode.

harshavardhana:
> @harshavardhana, we decided to do this change before (and independent of) FS backend format changes like possibly using extended attributes for storing user-metadata and bucket policy and the multipart backend to avoid relying for fcntl(3) locks for coordinating among many minio servers in a shared mode.

Okay understood now.

krisis:
@krishnasrinivas I have added the periodic cleanup routine similar to GCS gateway as well.

deekoder:
@krishnasrinivas  @fwessels  since the pr is updated since approval reset reviewers.

cmd/fs-v1-multipart.go
```diff
@@ -31,6 +31,11 @@ import (
 	"github.com/minio/sha256-simd"
 )
 
+const (
+	fsMultipartExpiry        = time.Hour * 24 * 14
```
fwessels, line 35: Maybe use time.Day * 14 ?

cmd/fs-v1-multipart.go
```diff
@@ -978,3 +936,58 @@ func (fs fsObjects) AbortMultipartUpload(bucket, object, uploadID string) error
 
 	return nil
 }
+
+// Remove multipart uploads left unattended in a given bucket older than `fsMultipartExpiry`
+func (fs fsObjects) cleanupStaleMultipartUpload(bucket string) error {
+	var (
+		err error
```
fwessels: You could name the return variable to avoid defining the variable explicitly

cmd/fs-v1-multipart.go
```diff
@@ -978,3 +936,58 @@ func (fs fsObjects) AbortMultipartUpload(bucket, object, uploadID string) error
...
+			return
+		}
+
+		for _, bucketInfo := range bucketInfos {
+			fs.cleanupStaleMultipartUpload(bucketInfo.Name)
+		}
+
+		// Schedule for the next multipart backend cleanup
+		time.Sleep(fsMultipartCleanupPeriod)
+	}
```
fwessels, line 989: Do we need to guard this thread from crashing? (Or restart it when it crashes?)

fwessels (approved):
Few minor comments, for the rest LGTM

cmd/fs-v1-multipart.go
krisis: @fwessels time.Day is not available. I (too) wish it were available.

cmd/fs-v1-multipart.go
krisis: Will do.

cmd/fs-v1-multipart.go
krisis: For now we can keep it simple and not restart on crash. This routine only frees up disk space consumed by 'stale' or 'forgotten' incomplete uploads.

harshavardhana (approved):

harshavardhana:
Making it a point to test this properly with Cloud Berry Client before the release.

PR #4700, opened 2017-07-19
Prevent minio server starting in standalone erasure mode when  distributed erasure was intended

### LINKED ISSUE #4686: Minio runs as stand-alone erasure when distributed-erasure is intended
I have not followed the issue template as explaining this issue in a free form makes more sense.

This issue is reported on slack. If user runs this command:

```
minio server 192.168.220.242:/export1 192.168.220.244:/export1 192.168.220.245:/export1 192.168.220.246:/export1 
```

i.e without `http://` prefix, we create directories 192.168.220.242:/export1 192.168.220.244:/export1 192.168.220.245:/export1 192.168.220.246:/export1 and run stand-alone erasure. But the intention was to run a distributed erasure. So we should either throw an error or assume `http://` prefix and run distributed erasure.

deekoder:
Fixes #4686

When there are multiple args, we assume and start in standalone erasure mode when scheme is absent for ips or localhost. Ask user to provide scheme to start correctly in distributed erasure mode.

## Description
Ask explicitly for a scheme to be specified when starting in a distributed erasure mode so that the minio server wont think that the user intends to run this in standalone distributed mode.

## Motivation and Context

Without this change : 
```
minio server 192.168.1.210:9000/tmp/dir0 192.168.1.210:9000/tmp/dir1 192.168.1.210:9000/tmp/dir2 192.168.110:9000/tmp/dir3
```
The server starts giving the user that minio started in distributed erasure mode when you can see this locally : 

```
find 192.168.1.210\:9000/
192.168.1.210:9000/
192.168.1.210:9000/tmp
192.168.1.210:9000/tmp/dir1
192.168.1.210:9000/tmp/dir1/.minio.sys
192.168.1.210:9000/tmp/dir1/.minio.sys/format.json
192.168.1.210:9000/tmp/dir1/.minio.sys/tmp
192.168.1.210:9000/tmp/dir1/.minio.sys/multipart
192.168.1.210:9000/tmp/dir2
192.168.1.210:9000/tmp/dir2/.minio.sys
192.168.1.210:9000/tmp/dir2/.minio.sys/format.json
192.168.1.210:9000/tmp/dir2/.minio.sys/tmp
192.168.1.210:9000/tmp/dir2/.minio.sys/multipart
192.168.1.210:9000/tmp/dir0
192.168.1.210:9000/tmp/dir0/.minio.sys
192.168.1.210:9000/tmp/dir0/.minio.sys/format.json
192.168.1.210:9000/tmp/dir0/.minio.sys/tmp
192.168.1.210:9000/tmp/dir0/.minio.sys/multipart
```

#4686 

## How Has This Been Tested?

Manually tested after change : 
```
./minio server 192.168.1.210:9000/tmp/dir0 192.168.1.210:9000/tmp/dir1 192.168.1.210:9000/tmp/dir2 192.168.110:9000/tmp/dir3
FATA[0000] Invalid command line arguments server=‘:9000’, args=[192.168.1.210:9000/tmp/dir0 192.168.1.210:9000/tmp/dir1 192.168.1.210:9000/tmp/dir2 192.168.110:9000/tmp/dir3]  cause='192.168.1.210:9000/tmp/dir0': invalid URL endpoint format: missing scheme http or https source=[server-main.go:95:serverHandleCmdArgs()]
```
and 
```
./minio server localhost:9000/tmp/dir0 localhost:9000/tmp/dir1 localhost:9000/tmp/dir2 localhost:9000/tmp/dir3
FATA[0000] Invalid command line arguments server=‘:9000’, args=[localhost:9000/tmp/dir0 localhost:9000/tmp/dir1 localhost:9000/tmp/dir2 localhost:9000/tmp/dir3]  cause='localhost:9000/tmp/dir0': invalid URL endpoint format: missing scheme http or https source=[server-main.go:95:serverHandleCmdArgs()]
```

## Types of changes

## Checklist:

- [X] My code follows the code style of this project.

### DIFF
--- a/cmd/endpoint.go
+++ b/cmd/endpoint.go
@@ -132,6 +132,12 @@ func NewEndpoint(arg string) (ep Endpoint, e error) {
 			return ep, err
 		}
 	} else {
+		// Only check if the arg is an ip address and ask for scheme since its absent.
+		// localhost, example.com, any FQDN cannot be disambiguated from a regular file path such as
+		// /mnt/export1. So we go ahead and start the minio server in FS modes in these cases.
+		if isHostIPv4(arg) {
+			return ep, fmt.Errorf("invalid URL endpoint format: missing scheme http or https")
+		}
 		u = &url.URL{Path: path.Clean(arg)}
 		isLocal = true
 	}
@@ -212,7 +218,6 @@ func NewEndpointList(args ...string) (endpoints EndpointList, err error) {
 			return nil, fmt.Errorf("duplicate endpoints found")
 		}
 		uniqueArgs.Add(arg)
-
 		endpoints = append(endpoints, endpoint)
 	}
 
@@ -267,6 +272,7 @@ func CreateEndpoints(serverAddr string, args ...string) (string, EndpointList, S
 	localEndpointCount := 0
 	localServerAddrSet := set.NewStringSet()
 	localPortSet := set.NewStringSet()
+
 	for _, endpoint := range endpoints {
 		endpointPathSet.Add(endpoint.Path)
 		if endpoint.IsLocal {
--- a/cmd/endpoint_test.go
+++ b/cmd/endpoint_test.go
@@ -70,6 +70,7 @@ func TestNewEndpoint(t *testing.T) {
 		{"http://server:8080//", Endpoint{}, -1, fmt.Errorf("empty or root path is not supported in URL endpoint")},
 		{"http://server:8080/", Endpoint{}, -1, fmt.Errorf("empty or root path is not supported in URL endpoint")},
 		{"http://server/path", Endpoint{}, -1, fmt.Errorf("lookup server" + errMsg)},
+		{"192.168.1.210:9000", Endpoint{}, -1, fmt.Errorf("invalid URL endpoint format: missing scheme http or https")},
 	}
 
 	for _, testCase := range testCases {
@@ -122,6 +123,7 @@ func TestNewEndpointList(t *testing.T) {
 		{[]string{"ftp://server/d1", "http://server/d2", "http://server/d3", "http://server/d4"}, fmt.Errorf("'ftp://server/d1': invalid URL endpoint format")},
 		{[]string{"d1", "http://localhost/d2", "d3", "d4"}, fmt.Errorf("mixed style endpoints are not supported")},
 		{[]string{"http://example.org/d1", "https://example.com/d1", "http://example.net/d1", "https://example.edut/d1"}, fmt.Errorf("mixed scheme is not supported")},
+		{[]string{"192.168.1.210:9000/tmp/dir0", "192.168.1.210:9000/tmp/dir1", "192.168.1.210:9000/tmp/dir2", "192.168.110:9000/tmp/dir3"}, fmt.Errorf("'192.168.1.210:9000/tmp/dir0': invalid URL endpoint format: missing scheme http or https")},
 	}
 
 	for _, testCase := range testCases {
--- a/cmd/net.go
+++ b/cmd/net.go
@@ -153,6 +153,15 @@ func getAPIEndpoints(serverAddr string) (apiEndpoints []string) {
 	return apiEndpoints
 }
 
+// isHostIPv4 - helper for validating if the provided arg is an ip address.
+func isHostIPv4(ipAddress string) bool {
+	host, _, err := net.SplitHostPort(ipAddress)
+	if err != nil {
+		host = ipAddress
+	}
+	return net.ParseIP(host) != nil
+}
+
 // checkPortAvailability - check if given port is already in use.
 // Note: The check method tries to listen on given port and closes it.
 // It is possible to have a disconnected client in this tiny window of time.
--- a/cmd/net_test.go
+++ b/cmd/net_test.go
@@ -317,3 +317,23 @@ func TestSameLocalAddrs(t *testing.T) {
 		}
 	}
 }
+func TestIsHostIPv4(t *testing.T) {
+	testCases := []struct {
+		args           string
+		expectedResult bool
+	}{
+		{"localhost", false},
+		{"localhost:9000", false},
+		{"example.com", false},
+		{"http://192.168.1.0", false},
+		{"http://192.168.1.0:9000", false},
+		{"192.168.1.0", true},
+	}
+
+	for _, testCase := range testCases {
+		ret := isHostIPv4(testCase.args)
+		if testCase.expectedResult != ret {
+			t.Fatalf("expected: %v , got: %v", testCase.expectedResult, ret)
+		}
+	}
+}


deekoder:
@harshavardhana  The suggested pattern is not catching all ips. So it's not an exact brief replacement for what i have now.

cmd/endpoint.go
```diff
@@ -55,6 +57,7 @@ func (endpoint Endpoint) String() string {
 
 // Type - returns type of endpoint.
 func (endpoint Endpoint) Type() EndpointType {
+
```
balamurugana: You could remove this newline addition

cmd/endpoint.go
```diff
@@ -133,6 +136,14 @@ func NewEndpoint(arg string) (ep Endpoint, e error) {
 		}
 	} else {
 		u = &url.URL{Path: path.Clean(arg)}
+
+		// Check if the arg is a localhost or an IP without a scheme.
+		if strings.Contains(arg, "localhost") || (checkIP4(arg) == true) {
```
donatello: Brackets around the part after the `||` are extraneous and may be removed.

cmd/endpoint.go
```diff
@@ -133,6 +136,14 @@ func NewEndpoint(arg string) (ep Endpoint, e error) {
...
+			if !((u.Scheme == "http" || u.Scheme == "https") &&
+				u.User == nil && u.Opaque == "" && u.ForceQuery == false && u.RawQuery == "" && u.Fragment == "") {
+				return ep, fmt.Errorf("invalid URL endpoint format: missing scheme http or https")
```
donatello: This if block appears to be incorrect. `u` is being assigned as `u = &url.URL{Path: path.Clean(arg)}` where only the Path part of the URL struct is being set. All [other items](https://golang.org/pkg/net/url/#URL) in the struct are not assigned . So they will have their 0 values by default. In fact `u.Scheme == ""` and that is why the scheme check is not doing anything. The `u` needs to be assigned by parsing the `arg` like `u, err = url.Parse(arg)` or so.

donatello (changes requested):
Please add tests to make sure this change is working.

cmd/endpoint.go
```diff
@@ -142,6 +153,14 @@ func NewEndpoint(arg string) (ep Endpoint, e error) {
 	}, nil
 }
 
+// checkIP4 - helper for validating if the provided arg is an ip address.
+func checkIP4(ipAddress string) bool {
```
donatello: https://golang.org/pkg/net/#ParseIP - there is already a library function to check for IP address - it will be better to use that than write another one ourself.

harshavardhana:
Can you paste which ips it doesn't work curious to know ? Adding a test case would help.

Even better is use net.ParseIP as suggested by @donatello .

cmd/endpoint.go
deekoder: ok really great catch. Actually once the control flow gets here, it means that the string was not splittable as a regular URL with scheme and host etc but localhost or the ip is erroneously taken as a regular filesystem path. Infact it creates a local directory with the ip address or localhost.  
So removed all this and send this PR again.

deekoder:
@harshavardhana  the replaced pattern didnt work for most ips i tried like 192.168.1.210. But this is mute now since I plan to use the inbuilt net.ParseIP.

harshavardhana:
https://play.golang.org/p/kGY9HqDhMj not sure about that. Here is the example it seems to be working fine.. is this how you used ?. Yes agree net.ParseIP is much better.

deekoder:
@harshavardhana  now i know. It failed when you add :9000. Anyway, moving to net.ParseIP

harshavardhana:
net.ParseIP would fail if you add port so you might have to remove that .

deekoder:
Addressed all the changes. Simplified and tested with localhost, ip address and also localhost without port and ip address without port. Seems to work. Please review and see if its a good fix.

cmd/endpoint.go
```diff
@@ -142,6 +146,27 @@ func NewEndpoint(arg string) (ep Endpoint, e error) {
...
+		// to check if it is an IP. Doing it this way allows us to not use regex to do a crude pattern matching
+		// for an IP address with port string.
+		// TODO: See if checkIPv4 can be reused where SplitHostPort is often called to check if it's an IP.
+		if !strings.Contains(err.Error(), "missing port in address") {
+			return false
+		}
+	}
+	// parse the host part of the ipAddress string without port to check if it is an IP without the scheme.
+	userIP := net.ParseIP(host)
+	if userIP == nil {
```
harshavardhana: You can just do this in one line.. `return net.ParseIP(host) != nil`

cmd/endpoint.go
```diff
@@ -142,6 +146,27 @@ func NewEndpoint(arg string) (ep Endpoint, e error) {
...
+// checkIPv4 - helper for validating if the provided arg is an ip address.
+func checkIPv4(ipAddress string) bool {
+	host, _, err := net.SplitHostPort(ipAddress)
+	if err != nil {
+		// If the error is anything other than missing port in address return false.
+		// If it is missing port in address we want to go ahead and parse the host string without address
+		// to check if it is an IP. Doing it this way allows us to not use regex to do a crude pattern matching
+		// for an IP address with port string.
+		// TODO: See if checkIPv4 can be reused where SplitHostPort is often called to check if it's an IP.
+		if !strings.Contains(err.Error(), "missing port in address") {
```
harshavardhana: This type of code might not work in future releases of go so we can be more diligent about this and handle it properly as we have seen with the recent go1.8 port work. Just to check for IPv4 in my opinion we can do this to simplify and fail using net.ParseIP behavior. 

```
host, _, _ := net.SplitHostPort(ipAddress)
if err != nil {
       host = ipAddress
}
return net.ParseIP(host) != nil 
```

harshavardhana: Similar changes needs to be achieved for other places we are using this pattern..  But in this case alone we can simplify.

cmd/endpoint.go
```diff
@@ -132,7 +132,11 @@ func NewEndpoint(arg string) (ep Endpoint, e error) {
 			return ep, err
 		}
 	} else {
-		u = &url.URL{Path: path.Clean(arg)}
```
harshavardhana: why is this change gone? it is needed to indicate that we treating the `arg` as Path only. `u` will be `nil` as otherwise.

```
return Endpoint{
                URL:     u,
                IsLocal: isLocal,
        }, nil
```

harshavardhana (changes requested):

cmd/endpoint.go
deekoder: ok.

cmd/endpoint.go
deekoder: If the error is something other than missing port we should return false since its not a proper ip.

deekoder: I dont know about 1.8 but if its any error other than port was not specified, I think we have to return with false.

harshavardhana: Correct but it will return false @deekoder net.ParseIP() can do that for you when there is an error in net.SplitHostPort().

deekoder: ah ok. Let me test it to make sure.

deekoder:
@harshavardhana  fixed and tested with localhost, ip address with and without 9000 port.

harshavardhana (approved):

cmd/endpoint.go
```diff
@@ -133,6 +133,11 @@ func NewEndpoint(arg string) (ep Endpoint, e error) {
 		}
 	} else {
 		u = &url.URL{Path: path.Clean(arg)}
+		// Check if the arg is a localhost or an IP without a scheme, then ask user to start with a scheme.
+		if strings.Contains(arg, "localhost") || checkIPv4(arg) {
```
balamurugana: 1. Any reason why `localhost` in entire path (substring) is treated as invalid?
2. It is good to do this check before `u = &url.URL{Path: path.Clean(arg)}`

cmd/endpoint.go
```diff
@@ -133,6 +133,11 @@ func NewEndpoint(arg string) (ep Endpoint, e error) {
...
+			return ep, fmt.Errorf("invalid URL endpoint format: missing scheme http or https")
+		}
+		// Only set isLocal to true after getting a valid localhost or IP address with the scheme.
```
balamurugana: This comment is no meaning to this context i.e. this section of the code knows given `arg` is a path.  You could remove this.

cmd/endpoint.go
```diff
@@ -142,6 +147,16 @@ func NewEndpoint(arg string) (ep Endpoint, e error) {
 	}, nil
 }
 
+// checkIPv4 - helper for validating if the provided arg is an ip address.
+func checkIPv4(ipAddress string) bool {
```
balamurugana: 1. You could name this function as `isHostIPv4()` or `isHostIP4()`. Function naming `check` usually return error.
2. As given arg is split, you could use `serverAddr` than `ipAddress`.  This means it doesn't receive ip address.
3. you could consider moving this function to net.go

cmd/endpoint.go
```diff
@@ -142,6 +147,16 @@ func NewEndpoint(arg string) (ep Endpoint, e error) {
...
+	host, _, err := net.SplitHostPort(ipAddress)
+	if err != nil {
+		host = ipAddress
+	}
+	// parse the host part of the ipAddress string without port to check if it is an IP without the scheme.
```
balamurugana: This comment is unclear to the context.  Either remove the comment or fix it to specific to the context.

cmd/endpoint.go
```diff
@@ -133,6 +133,11 @@ func NewEndpoint(arg string) (ep Endpoint, e error) {
 		}
 	} else {
 		u = &url.URL{Path: path.Clean(arg)}
+		// Check if the arg is a localhost or an IP without a scheme, then ask user to start with a scheme.
```
balamurugana: I would strongly recommend to add respective unit tests.

balamurugana (changes requested):

cmd/endpoint.go
deekoder: yes will add.

cmd/endpoint.go
deekoder: 1. So bala, control will arrive here if there was no http or https specifed. if the arg string has localhost or say localhost:8080 or localhost:9000 we should still say please give us scheme. 
2. Sure. We can move that around.

cmd/endpoint.go
deekoder: ok sure. I dont have to state the obvious here.

cmd/endpoint.go
deekoder: 1. Ok sure. 
2. Not clear what you mean here.
3. ah i think we can do this later when we compact the code in entire endpoint.go. If you see bala we do this check in other places in this file too. How about we wait to remove it properly everywhere and then move this somewhere else appropriate such as net.go

cmd/endpoint.go
deekoder: Will improve this comment. I think we need a comment here.

deekoder:
@balamurugana  @donatello  : addressed feedback. Please take a look.

cmd/endpoint.go
```diff
@@ -142,6 +148,15 @@ func NewEndpoint(arg string) (ep Endpoint, e error) {
 	}, nil
 }
 
+// isHostIPv4 - helper for validating if the provided arg is an ip address.
+func isHostIPv4(ipAddress string) bool {
+	host, _, err := net.SplitHostPort(ipAddress)
```
balamurugana: I see given argument is in the form of IP:/path,  `ipAddress` as  argument is not correct.  The returned value of `net.SplitHostPort()` could be `IP` and `/path' (as port).  I think it would be more appropriate to swap names i.e. `ipAddress` becomes `host` vice versa.

cmd/endpoint.go
```diff
@@ -142,6 +148,15 @@ func NewEndpoint(arg string) (ep Endpoint, e error) {
 	}, nil
 }
 
+// isHostIPv4 - helper for validating if the provided arg is an ip address.
+func isHostIPv4(ipAddress string) bool {
```
balamurugana: As this function is introduced in this PR, I couldn't able to understand why we need another PR for moving it to `net.go`

cmd/endpoint.go
```diff
@@ -132,6 +132,12 @@ func NewEndpoint(arg string) (ep Endpoint, e error) {
 			return ep, err
 		}
 	} else {
+		// Only check if the arg is an ip address and ask for scheme since its absent.
+		// localhost, example.com, any FQDN cannot be disambiguated from a regular file path such as
+		// /mnt/export1. So we go ahead and start the minio server in FS modes in these cases.
+		if isHostIPv4(arg) {
```
balamurugana, line 138: You would need to add test cases for this check in `TestNewEndpoint()`

cmd/endpoint_test.go
```diff
@@ -369,3 +369,24 @@ func TestGetRemotePeers(t *testing.T) {
...
+	testCases := []struct {
+		args           string
+		expectedResult bool
+	}{
+		{"localhost", false},
+		{"localhost:9000", false},
+		{"example.com", false},
+		{"192.168.1.0", true},
+		{"http://192.168.1.0", true},
+		{"http://192.168.1.0:9000", true},
```
balamurugana: You would need to add actual reported case i.e. `192.168.1.1:/path`

cmd/endpoint.go
donatello: Argument can be renamed to `address` or `networkAddress` if you prefer.

donatello (approved):

deekoder:
I will move isHostIPV4 to net.go today and then merge.

balamurugana (approved):

PR #4189, opened 2017-04-28
Add hound CI to allow jshint for ES6

harshavardhana:
## Description
Add hound CI to allow jshint for ES6

## Motivation and Context
HoundCI incorrect comments on PR #4188 

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #4188, opened 2017-04-28
browser: Listing should append instead of replacing previous listing

krishnasrinivas:
fixes #4144

## Description

If the number of objects are say 1001 in a bucket, browser would show first 1000 objects. Upon scroll down it would load the remaining 1 object. Instead of appending it would replace the current 1000 objects with the newly listed object.

This fix defines new actions RESET_OBJECTS and APPEND_OBJECTS

## Motivation and Context

fixes #4144

## How Has This Been Tested?

manually

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #4708, opened 2017-07-20
ci: rename to .hound.yml

brendanashworth:
As of 1ea53b4, a hound.yml file was added to disable JavaScript linting
with jshint. This is the behavior we want, as the browser follows its
own linting style that isn't JSHint. The correct name for the file is
.hound.yml, so nothing happened.

This also removes .jshintrc, because we don't use jshint to lint.

Refs: https://github.com/minio/minio/pull/4189
Refs: https://github.com/minio/minio/pull/4188
See: https://houndci.com/configuration#javascript


poornas (approved):
LGTM

brendanashworth:
@deekoder this targets `newux` now.

krishnasrinivas (approved):

Issue #4718, opened 2017-07-24
Azure should return NoSuchBucketPolicy if policy not found


poornas:
## Expected Behavior
Mint fails for Minio-py sdk because Azure returns { "Version":  "2012-10-17", "Statement": None} instead of NoSuchBucketPolicy if bucket policy does not exist. Refer https://github.com/minio/minio-py/issues/558

This may be an issue with GCS as well.

## Steps to Reproduce (for bugs)

1. Run minio gateway azure 
2. Run ./tests/functional_test.sh py3 for minio-py 2.2.4 release with azure credentials and localhost as SERVER_ENDPOINT
3.
4.
NoneType is not iterable error for get_policy 
## Context

Mint tests fail for Azure gateway. Azure and GCS response for get bucket policy should be 404 error similar to S3 and  minio server.
minio server
```
<Error><Code>NoSuchBucketPolicy</Code><Message>The bucket policy does not exist</Message><Key></Key><BucketName></BucketName><Resource>/bucket/</Resource><RequestId>3L137</RequestId><HostId>3L137</HostId></Error>
```

and s3:

```
<?xml version="1.0" encoding="UTF-8"?>\n<Error><Code>NoSuchBucketPolicy</Code><Message>The bucket policy does not exist</Message><BucketName>vadmeste</BucketName><RequestId>B93FA31B56838F07</RequestId><HostId>4vGdcClynoPiT8MKAKoN5NsBq2RHMhVOr1m5SoU9lvOTWeG1hmNL5o6Y++Vu6sV/</HostId></Error>
```

## Your Environment

* Version used (`minio version`): 2017-07-24T18:27:35Z
* Environment name and version (e.g. nginx 1.9.1): 
* Server type and version: 
* Operating System and version (`uname -a`):Linux kp-ThinkPad-X1-Carbon-5th 4.10.0-28-generic #32-Ubuntu SMP Fri Jun 30 05:32:18 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
* Link to your project:

poornas:
@deekoder, this can be closed

PR #4729, opened 2017-07-27
azure: When container access type is private treat as no policy set

poornas:
## Description

Azure gateway's response to GetBucketPolicies call is not consistent with 
S3 and server response when a bucket does not have a public policy 
defined for it. Both S3 And minio server respond with NoSuchBucketPolicy 
error, which is interpeted by the client SDK's as policy type None
In the case of Azure, by default container is set to access private 
meaning no anonymous access.  Azure gateway currently returns 
a policy object with empty statements when container is private 
access. Instead it should be interpreted as policy not found. 

## Motivation and Context
 

Azure gateway and minio server handling of GetBucketPolicies needs to be consistent . See #4718 

## How Has This Been Tested?

Manually by running server in gateway mode for azure and testing with minio-py client SDK to connect to server and get bucket policy for a newly created bucket with no policy defined on it.
## Types of changes

## Checklist:

### DIFF
--- a/cmd/gateway-azure.go
+++ b/cmd/gateway-azure.go
@@ -738,7 +738,7 @@ func (a *azureObjects) GetBucketPolicies(bucket string) (policy.BucketAccessPoli
 	}
 	switch perm.AccessType {
 	case storage.ContainerAccessTypePrivate:
-		// Do nothing
+		return policy.BucketAccessPolicy{}, traceError(PolicyNotFound{Bucket: bucket})
 	case storage.ContainerAccessTypeContainer:
 		policyInfo.Statements = policy.SetPolicy(policyInfo.Statements, policy.BucketPolicyReadOnly, bucket, "")
 	default:


harshavardhana:
@poornas you can keep the PR title to be more concise and not exceed 50 characters. This PR if you see has wrapped around in a weird manner and overflowed into the description of the PR.

harshavardhana (changes requested):
@poornas you can keep the PR title to be more concise and not exceed 50 characters. This PR if you see has wrapped around in a weird manner and overflowed into the description of the PR.

PR title not just github PR title, you would need to fix also your commit message.

krishnasrinivas (approved):

harshavardhana (commented):
@poornas have you seen my previous comment?

poornas:
@harshavardhana , updated commit as well.

harshavardhana (approved):

Issue #4721, opened 2017-07-25
Update Mesos & Kubernetes with latest release tag


deekoder:
Since we have a new release. We need to send PRs to Mesos & Kubernetes . @krisis  could we also add these steps to the release doc.

deekoder:
https://github.com/kubernetes/charts/pull/1304/files
https://github.com/mesosphere/universe/pull/1169/files

nitisht:
@deekoder I can sent the PR for k8s/helm

nitisht:
https://github.com/kubernetes/charts/pull/1546

ebozduman:
I created a PR for required mesosphere/universe changes.
https://github.com/mesosphere/universe/pull/1290
Waiting for its approval and merge.
I also updated the [release doc](https://docs.google.com/document/d/1yRrWmAa2yjNNjmfnlL7oQfuskZZk5Vy2BjvZR3fZ2qo/edit#heading=h.r2p5sjd4d9jf) about what changes need to be done at kubernetes and mesosphere repos when there is a new minio release.

PR #4730, opened 2017-07-28
Removes max limit requirement on accessKey and secretKey length

ebozduman:
## Description
Fixes Issue #4721 by removing the max limit variable and check 
so that there is no max limit on both accessKey and secretKey. 
We are still using minio specific max length to generate accessKey 
and secretKey. Corresponding tests and README file are also modified.

## Motivation and Context
Issue #4721: Increase access key to 24 chars for compatibility with Azure

## How Has This Been Tested?
- Started minio in gateway (azure) and standalone modes with a 30 byte long accessKey successfully without getting max length error message.
- Ran with less than 4 bytes accessKey and less than 8 secretKey to make sure correct error messages are displayed.
- Ran existing test cases.

## Types of changes

## Checklist:
- [x] I have updated existing tests to cover my changes.

### DIFF
--- a/cmd/credential.go
+++ b/cmd/credential.go
@@ -29,6 +29,7 @@ const (
 	accessKeyMinLen = 5
 
 	// Maximum length for Minio access key.
+	// There is no max length enforcement for access keys
 	accessKeyMaxLen = 20
 
 	// Minimum length for Minio secret key for both server and gateway mode.
@@ -36,11 +37,8 @@ const (
 
 	// Maximum secret key length for Minio, this
 	// is used when autogenerating new credentials.
-	secretKeyMaxLenMinio = 40
-
-	// Maximum secret key length allowed from client side
-	// caters for both server and gateway mode.
-	secretKeyMaxLen = 100
+	// There is no max length enforcement for secret keys
+	secretKeyMaxLen = 40
 
 	// Alpha numeric table used for generating access keys.
 	alphaNumericTable = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ"
@@ -51,18 +49,18 @@ const (
 
 // Common errors generated for access and secret key validation.
 var (
-	errInvalidAccessKeyLength = errors.New("Invalid access key, access key should be 5 to 20 characters in length")
-	errInvalidSecretKeyLength = errors.New("Invalid secret key, secret key should be 8 to 100 characters in length")
+	errInvalidAccessKeyLength = errors.New("Invalid access key, access key should be minimum 5 characters in length")
+	errInvalidSecretKeyLength = errors.New("Invalid secret key, secret key should be minimum 8 characters in length")
 )
 
 // isAccessKeyValid - validate access key for right length.
 func isAccessKeyValid(accessKey string) bool {
-	return len(accessKey) >= accessKeyMinLen && len(accessKey) <= accessKeyMaxLen
+	return len(accessKey) >= accessKeyMinLen
 }
 
 // isSecretKeyValid - validate secret key for right length.
 func isSecretKeyValid(secretKey string) bool {
-	return len(secretKey) >= secretKeyMinLen && len(secretKey) <= secretKeyMaxLen
+	return len(secretKey) >= secretKeyMinLen
 }
 
 // credential container for access and secret keys.
@@ -116,24 +114,35 @@ func createCredential(accessKey, secretKey string) (cred credential, err error) 
 }
 
 // Initialize a new credential object
-func mustGetNewCredential() credential {
+func getNewCredential(accessKeyLen, secretKeyLen int) (cred credential, err error) {
 	// Generate access key.
-	keyBytes := make([]byte, accessKeyMaxLen)
-	_, err := rand.Read(keyBytes)
-	fatalIf(err, "Unable to generate access key.")
-	for i := 0; i < accessKeyMaxLen; i++ {
+	keyBytes := make([]byte, accessKeyLen)
+	_, err = rand.Read(keyBytes)
+	if err != nil {
+		return cred, err
+	}
+
+	for i := 0; i < accessKeyLen; i++ {
 		keyBytes[i] = alphaNumericTable[keyBytes[i]%alphaNumericTableLen]
 	}
 	accessKey := string(keyBytes)
 
 	// Generate secret key.
-	keyBytes = make([]byte, secretKeyMaxLenMinio)
+	keyBytes = make([]byte, secretKeyLen)
 	_, err = rand.Read(keyBytes)
-	fatalIf(err, "Unable to generate secret key.")
-	secretKey := string([]byte(base64.StdEncoding.EncodeToString(keyBytes))[:secretKeyMaxLenMinio])
+	if err != nil {
+		return cred, err
+	}
 
-	cred, err := createCredential(accessKey, secretKey)
-	fatalIf(err, "Unable to generate new credential.")
+	secretKey := string([]byte(base64.StdEncoding.EncodeToString(keyBytes))[:secretKeyLen])
+	cred, err = createCredential(accessKey, secretKey)
 
+	return cred, err
+}
+
+func mustGetNewCredential() credential {
+	// Generate Minio credentials with Minio key max lengths.
+	cred, err := getNewCredential(accessKeyMaxLen, secretKeyMaxLen)
+	fatalIf(err, "Unable to generate new credentials.")
 	return cred
 }
--- a/cmd/jwt_test.go
+++ b/cmd/jwt_test.go
@@ -24,8 +24,12 @@ func testAuthenticate(authType string, t *testing.T) {
 		t.Fatalf("unable initialize config file, %s", err)
 	}
 	defer removeAll(testPath)
-
-	serverCred := serverConfig.GetCredential()
+	// Create access and secret keys in length, 300 and 600
+	cred, err := getNewCredential(300, 600)
+	if err != nil {
+		t.Fatalf("unable to get new credential, %v", err)
+	}
+	serverConfig.SetCredential(cred)
 
 	// Define test cases.
 	testCases := []struct {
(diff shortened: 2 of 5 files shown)


cmd/credential.go
```diff
@@ -29,7 +29,7 @@ const (
 	accessKeyMinLen = 5
 
 	// Maximum length for Minio access key.
-	accessKeyMaxLen = 20
+	accessKeyMaxLenMinio = 20
```
balamurugana: I don't think we need to suffix `Minio` here. it looks odd.

cmd/jwt_test.go
```diff
@@ -35,16 +35,10 @@ func testAuthenticate(authType string, t *testing.T) {
...
-		// Access key too long.
-		{"user12345678901234567", "pass", errInvalidAccessKeyLength},
 		// Access key contains unsuppported characters.
 		{"!@#$", "pass", errInvalidAccessKeyLength},
-		// Success when access key contains leading/trailing spaces.
-		{" " + serverCred.AccessKey + " ", serverCred.SecretKey, errInvalidAccessKeyLength},
 		// Secret key too small.
 		{"myuser", "pass", errInvalidSecretKeyLength},
-		// Secret key too long.
-		{"myuser", "pass1234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890", errInvalidSecretKeyLength},
```
balamurugana: I think we need to keep these cases for jwt aspect because access/secret keys are passed outside of minio like browser etc.

balamurugana (changes requested):

cmd/credential.go
ebozduman: For consistency purposes I suffixed it, just like the original code had it before for `secretKeyMaxLenMinio`.
The other implication by having it there is the fact that this max length variable and value is only needed when generating Minio access and secret keys. So, it is kind of Minio specific. In other words, we don't use them for validating user provided access and secret keys. That's why it made sense to me to add the suffix.

cmd/jwt_test.go
ebozduman: Followings are the reasons for removal of 3 test cases:
1.
		```
		// Access key too long.
		{"user12345678901234567", "pass", errInvalidAccessKeyLength},
		```
There is no "too long" case after this fix, that is; we don't check max length after removal of max length check. Even if we would like to keep it, test gets `errInvalidAccessKeyID = "The access key ID you provided does not exist in our records"`.
2.
		```
 		// Success when access key contains leading/trailing spaces.
		{" " + serverCred.AccessKey + " ", serverCred.SecretKey, errInvalidAccessKeyLength},
		```
This is a left over test after the removal of the cleanup code for leading and trailing spaces in accessKey. (Anyways, this test case was not correct since it was supposed to expect a Pass instead of expecting byte length failure error message.)
3.
		```
		// Secret key too long.
		{"myuser", "pass1234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890", errInvalidSecretKeyLength},
		```
There is no "too long" case after this fix, removal of max length check. Same as #1 bullet above.

ebozduman: @balamurugana 
Please let me know if you agree/disagree.

krisis (approved):

poornas (approved):

cmd/credential.go
balamurugana: Even this case `Minio` suffix (for secretKeyMaxLenMinio` too) is not conveying anything without reading the place it used.  IMO `accessKeyMaxLen` itself is sufficient.  If it looks generic `accessKeyGenMaxLen` would work.

cmd/jwt_test.go
balamurugana: IMO, for case 1 and 3, you would need to check for the success than removing it.  These positive cases tell what inputs are valid.

cmd/credential.go
ebozduman: Done

cmd/jwt_test.go
ebozduman: There is actually no difference between the last "Success" test case and having other random number of positive tests doing exactly the same thing. 

I mean, as long as the key lengths are longer than the minimum values, then they should pass the test, and there is no magic length number we specifically would like to try. 
So, what I changed is to have a longer than before access and secret keys recorded as valid values in the server config, so that the last "Success" test case can use these "long" key values instead of the Minio default key length values, 20/40.

Another thing I've noticed is that the "unsupported characters" test case actually is meaningless since it still tests the length. I also checked the error messages `api-errors.go` file to see if there is a dedicated error message for this condition or not and there was none.

If we would like to implement it and capture a different error message as it deserves, then it needs to be addressed in a separate issue imho, since this specific issue is not about fixing test cases, but to eliminate the max length checks for both keys.

cmd/credential.go
```diff
@@ -116,21 +112,21 @@ func createCredential(accessKey, secretKey string) (cred credential, err error)
 }
 
 // Initialize a new credential object
-func mustGetNewCredential() credential {
+func mustGetNewCredential(accessKeyLen, secretKeyLen int) credential {
```
balamurugana: I see this signature is used only in jwt_test.go.  You can leave the current signature and create another function like
```go
func GetNewCredential(accessKeyLen, secretKeyLen int) (credential, error) {
// same as this mustGetNewCredential() except returning credential and error
}
```

then, you could have `mustGetNewCredential()` like
```go
func mustGetNewCredential() credential {
    cred, err := GetNewCredential(accessKeyMaxLen, secretKeyMaxLen)
    fatalIf(err, "Unable to generate new credential.")
    return cred
}
```

In jwt_test.go, you could use `GetNewCredential()` than `mustGetNewCredential()`

cmd/jwt_test.go
```diff
@@ -33,18 +34,10 @@ func testAuthenticate(authType string, t *testing.T) {
...
+		{"user", serverCred.SecretKey, errInvalidAccessKeyLength},
 		// Access key contains unsuppported characters.
-		{"!@#$", "pass", errInvalidAccessKeyLength},
-		// Success when access key contains leading/trailing spaces.
-		{" " + serverCred.AccessKey + " ", serverCred.SecretKey, errInvalidAccessKeyLength},
-		// Secret key too small.
-		{"myuser", "pass", errInvalidSecretKeyLength},
-		// Secret key too long.
-		{"myuser", "pass1234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890", errInvalidSecretKeyLength},
+		{serverCred.AccessKey, "pass", errInvalidSecretKeyLength},
```
balamurugana: The above comment is not appropriate to this case.

cmd/credential.go
ebozduman: Done

cmd/jwt_test.go
ebozduman: Done

cmd/credential.go
```diff
@@ -116,24 +114,30 @@ func createCredential(accessKey, secretKey string) (cred credential, err error)
...
-	}
-	accessKey := string(keyBytes)
-
-	// Generate secret key.
-	keyBytes = make([]byte, secretKeyMaxLenMinio)
+	keyBytes := make([]byte, accessKeyLen)
 	_, err = rand.Read(keyBytes)
-	fatalIf(err, "Unable to generate secret key.")
-	secretKey := string([]byte(base64.StdEncoding.EncodeToString(keyBytes))[:secretKeyMaxLenMinio])
+	if err == nil {
```
balamurugana: Actually we prefer `err != nil` convention.

cmd/credential.go
```diff
@@ -116,24 +114,30 @@ func createCredential(accessKey, secretKey string) (cred credential, err error)
...
+			keyBytes[i] = alphaNumericTable[keyBytes[i]%alphaNumericTableLen]
+		}
+		accessKey := string(keyBytes)
 
-	cred, err := createCredential(accessKey, secretKey)
-	fatalIf(err, "Unable to generate new credential.")
+		// Generate secret key.
+		keyBytes = make([]byte, secretKeyLen)
+		_, err = rand.Read(keyBytes)
+		if err == nil {
```
balamurugana: same as above

cmd/jwt_test.go
```diff
@@ -24,7 +24,8 @@ func testAuthenticate(authType string, t *testing.T) {
 		t.Fatalf("unable initialize config file, %s", err)
 	}
 	defer removeAll(testPath)
-
+	// Create access and secret keys in length, 300 and 600
+	getNewCredential(300, 600)
```
balamurugana: As you don't keep the return values, I am not sure what exactly you are trying to do here.

balamurugana (changes requested):

cmd/credential.go
ebozduman: If I use `err != nil` in both places, I need to `return` 3 times, 3 exit points in the function. I thought 1 exit is much better and cleaner, but since this is the standard I'll make the change and push it.

cmd/credential.go
ebozduman: Will do

cmd/jwt_test.go
ebozduman: You are right. I need to capture the returned credentials and set them in the server configuration before start running the tests.

cmd/credential.go
harshavardhana: Three exit points is fine. Its a limitation of Go language.  Other languages do error handling more cleanly.

cmd/jwt_test.go
```diff
@@ -24,33 +24,26 @@ func testAuthenticate(authType string, t *testing.T) {
 		t.Fatalf("unable initialize config file, %s", err)
 	}
 	defer removeAll(testPath)
-
-	serverCred := serverConfig.GetCredential()
+	// Create access and secret keys in length, 300 and 600
+	cred, _ := getNewCredential(300, 600)
```
balamurugana: You could handle error here like

```go
cred, err := getNewCredential(300, 600)
if err != nil {
    t.Fatalf("unable to get new credential, %v", err)
}
```

ebozduman: Good idea!

balamurugana (approved):

PR #4736, opened 2017-07-31
bucket-policy: Add IPAddress/NotIPAddress conditions support

### LINKED ISSUE #4369: [feature] Support restricting access by IP address via bucket policy
This may be a useful feature for Minio users that S3 also supports: 

http://docs.aws.amazon.com/AmazonS3/latest/dev/example-bucket-policies.html#example-bucket-policies-use-case-3

## Expected Behavior
Blacklist and whitelist of IP addresses should be possible.

## Current Behavior
Not possible currently. We support a related feature: restricting by a specific HTTP referrer: https://docs.aws.amazon.com/AmazonS3/latest/dev/example-bucket-policies.html#example-bucket-policies-use-case-4

awwalker:
## Description
Adds the ability to restrict bucket access by IP addresses. See for more detail.
https://docs.aws.amazon.com/AmazonS3/latest/dev/example-bucket-policies.html#example-bucket-policies-use-case-3
## Motivation and Context
Fixes: #4369 

## How Has This Been Tested?
Manually and Unit tests.
## Types of changes

## Checklist:

### DIFF
--- a/cmd/object-handlers.go
+++ b/cmd/object-handlers.go
@@ -25,7 +25,6 @@ import (
 	"net/url"
 	"sort"
 	"strconv"
-	"strings"
 
 	mux "github.com/gorilla/mux"
 )
@@ -51,20 +50,14 @@ func setGetRespHeaders(w http.ResponseWriter, reqParams url.Values) {
 
 // getSourceIPAddress - get the source ip address of the request.
 func getSourceIPAddress(r *http.Request) string {
-	// Attempt to get ip from standard headers.
 	var ip string
-	for _, h := range []string{"X-Forwarded-For", "X-Real-Ip"} {
-		addresses := r.Header[h]
-		// Get most recently used ip.
-		for i := len(addresses) - 1; i >= 0; i-- {
-			ip = strings.TrimSpace(addresses[i])
-			parsedIP := net.ParseIP(ip)
-			// Skip non valid IP address.
-			if parsedIP == nil {
-				continue
-			}
-			return ip
-		}
+	// Attempt to get ip from standard headers.
+	// Do not support X-Forwarded-For because it is easy to spoof.
+	ip = r.Header.Get("X-Real-Ip")
+	parsedIP := net.ParseIP(ip)
+	// Skip non valid IP address.
+	if parsedIP != nil {
+		return ip
 	}
 	// Default to remote address if headers not set.
 	ip, _, _ = net.SplitHostPort(r.RemoteAddr)
--- a/cmd/object-handlers.go
+++ b/cmd/object-handlers.go
@@ -52,11 +52,12 @@ func setGetRespHeaders(w http.ResponseWriter, reqParams url.Values) {
 // getSourceIPAddress - get the source ip address of the request.
 func getSourceIPAddress(r *http.Request) string {
 	// Attempt to get ip from standard headers.
+	var ip string
 	for _, h := range []string{"X-Forwarded-For", "X-Real-Ip"} {
 		addresses := strings.Split(r.Header.Get(h), ",")
 		// Get most recently used ip.
 		for i := len(addresses) - 1; i >= 0; i-- {
-			ip := strings.TrimSpace(addresses[i])
+			ip = strings.TrimSpace(addresses[i])
 			parsedIP := net.ParseIP(ip)
 			// Skip non valid IP address.
 			if parsedIP == nil {
@@ -66,7 +67,8 @@ func getSourceIPAddress(r *http.Request) string {
 		}
 	}
 	// Default to remote address if headers not set.
-	return r.RemoteAddr
+	ip, _, _ = net.SplitHostPort(r.RemoteAddr)
+	return ip
 }
 
 // errAllowableNotFound - For an anon user, return 404 if have ListBucket, 403 otherwise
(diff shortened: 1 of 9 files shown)


harshavardhana:
Nice work . Thanks @awwalker a long pending feature :-)

krisis:
@awwalker I was testing this PR locally using `aws s3api` and couldn't get the `IpAddress` condition working. I tried to debug the issue and found that `isIPInCIDR` check tries to match the `sourceIP` which includes the source port too. E.g, 127.0.0.1:56164 127.0.0.1/24, where the IP address with the subnet is part of the bucket policy and the IP address with the port is the request's source IP. Is there something I am missing?

Following is the policy json I used (generated using https://awspolicygen.s3.amazonaws.com/policygen.html),
```json
{
  "Id": "Policy1501543727867",
  "Version": "2012-10-17",
  "Statement": [
    {
      "Sid": "Stmt1501543723428",
      "Action": [
        "s3:GetObject"
      ],
      "Effect": "Allow",
      "Resource": "arn:aws:s3:::test/*",
      "Condition": {
        "IpAddress": {
          "aws:SourceIp": "127.0.0.1/24"
        }
      },
      "Principal": "*"
    }
  ]
}
```

harshavardhana:
The problem is there is a bug in the PR @krisis - sourceIPAddress is not used instead only X-Forward-IP and X-Real-IP are used to look for source address. This needs to be fixed.

cmd/object-handlers.go
```diff
@@ -48,16 +49,37 @@ func setGetRespHeaders(w http.ResponseWriter, reqParams url.Values) {
 	}
 }
 
+// getSourceIPAddress - get the source ip address of the request.
+func getSourceIPAddress(r *http.Request) string {
+	// Attempt to get ip from standard headers.
+	for _, h := range []string{"X-Forwarded-For", "X-Real-Ip"} {
```
harshavardhana: We need to look for r.RemoteAddr, i don't see if we need to look for these type of headers. it should be simple as 

```
        // Get host and port from Request.RemoteAddr.
        host, _, _ := net.SplitHostPort(r.RemoteAddr)
```

harshavardhana (changes requested):

cmd/object-handlers.go
awwalker: I used those headers in case a proxy was in front of the request. I thought if that was the case remote address would refer to the proxy not the source of the request. If none of the headers are used it defaults to remoteAddr. Am I wrong?

awwalker: I do see the bug regarding splitting host from port and can fix that.

awwalker:
@harshavardhana @krisis Can you take a look again? I fixed the splitting of host and port and added a new test

cmd/object-handlers.go
```diff
@@ -48,16 +49,39 @@ func setGetRespHeaders(w http.ResponseWriter, reqParams url.Values) {
 	}
 }
 
+// getSourceIPAddress - get the source ip address of the request.
+func getSourceIPAddress(r *http.Request) string {
+	// Attempt to get ip from standard headers.
+	var ip string
+	for _, h := range []string{"X-Forwarded-For", "X-Real-Ip"} {
```
harshavardhana: Actually there is a problem if you directly use canonical strings for `X-Forwarded-For` It would be easier to just use r.Header.Get(h) instead of looping through multiple addresses..  It is also possible to spoof this information easily. 

>The general format of the field is:
>> X-Forwarded-For: client, proxy1, proxy2[3]

> where the value is a comma+space separated list of IP addresses, the left-most being the original client, and each successive proxy that passed the request adding the IP address where it received the request from. In this example, the request passed through proxy1, proxy2, and then proxy3 (not shown in the header). proxy3 appears as remote address of the request. Since it is easy to forge an X-Forwarded-For field the given information should be used with care. The last IP address is always the IP address that connects to the last proxy, which means it is the most reliable source of information. X-Forwarded-For data can be used in a forward or reverse proxy scenario. Just logging the X-Forwarded-For field is not always enough as the last proxy IP address in a chain is not contained within the X-Forwarded-For field, it is in the actual IP header. A web server should log BOTH the request's source IP address and the X-Forwarded-For field information for completeness.

Some more story here 
- https://serverfault.com/questions/314574/nginx-real-ip-header-and-x-forwarded-for-seems-wrong/414166#414166
- https://distinctplace.com/2014/04/23/story-behind-x-forwarded-for-and-x-real-ip-headers/

So from what it looks like `X-Forwarded-For` can be easily spoofed and will lead to allowing access to objects on malicious clients with this in mind i suggest we use `X-Real-Ip` only.  So it means that 

header.Get(`X-Real-Ip`) is all you need to do, since X-Real-Ip is only set by proxies and cannot be changed/spoofed by the client and are unset by the nginx at termination.

harshavardhana (changes requested):
Found a security issue in using X-Forwarded-For header.. please review @awwalker

cmd/object-handlers.go
awwalker: Ok awesome (not the easy spoofing). I will change it to only support Real-Ip header. Thanks for the links.

krisis (approved):
LGTM. The test works locally. Thanks for the PR :+1:

harshavardhana (approved):

PR #4739, opened 2017-07-31
Fix testcase to not overflow int type

### LINKED ISSUE #4738: Int overflow in TestListObjects on 32-bit systems
When running the tests on a 32-bit system I end up with a int overflow.

## Expected Behavior
Test shouldn't fail even if ran on a 32-bit systems.

## Current Behavior
Currently I end up with the following behaviour when running the TestListObjects test.

```
$ go test -run TestListObjects
# github.com/minio/minio/cmd
./object-api-listobjects_test.go:449: constant 1111000000000000 overflows int
./object-api-listobjects_test.go:455: constant 1234567891011 overflows int
FAIL	github.com/minio/minio/cmd [build failed]
```

This occurs due to the fact that the maxKeys property is defined as an int and the values that are being sent in are bigger then a 32-bit int.

According to the [go tour](https://tour.golang.org/basics/11) : `The int, uint, and uintptr types are usually 32 bits wide on 32-bit systems and 64 bits wide on 64-bit systems.` which means that the type will vary between the systems.

## Possible Solution
There are 2 possible solutions as I see it.

1. Change the maxValue data type to int64.
2. Use smaller values then [2147483647](https://en.wikipedia.org/wiki/2,147,483,647#In_computing) in the tests cases.

1. Use a 32-bit machine like Raspberry Pi 2 Model B v1.1
2. Clone repository `go get github.com/minio/minio`
3. Change directory to cmd `cd $GOPATH/src/github.com/minio/minio/cmd`
3. Run ListObjects test `go test -run TestListObjects`

## Context
I was trying to run the tests in order to see if they work as expected on 32-bit arm architecture.

## Your Environment
Latest checkout of master branch currently at commit: `ec5293ce29ef03726c7174d9e3304bc156259c4e`
* Server type and version:
`go version go1.7.6 linux/arm`
Raspberry Pi 2 Model B v1.1
Debian 8 Jessie
`Linux rpi 4.9.28-v7+ #998 SMP Mon May 15 16:55:39 BST 2017 armv7l GNU/Linux`

donatello:
The int type is only 32-bits wide on 32-bit CPUs.

Fixes #4738

## Motivation and Context

#4738 

## How Has This Been Tested?

Tests would pass on 32-bit systems too as the test constants fit in signed 32-bit ints.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/object-api-listobjects_test.go
+++ b/cmd/object-api-listobjects_test.go
@@ -412,7 +412,7 @@ func testListObjects(obj ObjectLayer, instanceType string, t TestErrHandler) {
 		prefix     string
 		marker     string
 		delimeter  string
-		maxKeys    int
+		maxKeys    int32
 		// Expected output of ListObjects.
 		result ListObjectsInfo
 		err    error
@@ -446,13 +446,13 @@ func testListObjects(obj ObjectLayer, instanceType string, t TestErrHandler) {
 		{"empty-bucket", "", "", "", -1, ListObjectsInfo{}, nil, true},
 		{"empty-bucket", "", "", "", 1, ListObjectsInfo{}, nil, true},
 		// Setting maxKeys to a very large value (17).
-		{"empty-bucket", "", "", "", 1111000000000000, ListObjectsInfo{}, nil, true},
+		{"empty-bucket", "", "", "", 111100000, ListObjectsInfo{}, nil, true},
 		// Testing for all 7 objects in the bucket (18).
 		{"test-bucket-list-object", "", "", "", 9, resultCases[0], nil, true},
 		//Testing for negative value of maxKey, this should set maxKeys to listObjectsLimit (19).
 		{"test-bucket-list-object", "", "", "", -1, resultCases[0], nil, true},
 		// Testing for very large value of maxKey, this should set maxKeys to listObjectsLimit (20).
-		{"test-bucket-list-object", "", "", "", 1234567891011, resultCases[0], nil, true},
+		{"test-bucket-list-object", "", "", "", 1234567890, resultCases[0], nil, true},
 		// Testing for trancated value (21-24).
 		{"test-bucket-list-object", "", "", "", 5, resultCases[1], nil, true},
 		{"test-bucket-list-object", "", "", "", 4, resultCases[2], nil, true},
@@ -520,7 +520,7 @@ func testListObjects(obj ObjectLayer, instanceType string, t TestErrHandler) {
 	}
 
 	for i, testCase := range testCases {
-		result, err := obj.ListObjects(testCase.bucketName, testCase.prefix, testCase.marker, testCase.delimeter, testCase.maxKeys)
+		result, err := obj.ListObjects(testCase.bucketName, testCase.prefix, testCase.marker, testCase.delimeter, int(testCase.maxKeys))
 		if err != nil && testCase.shouldPass {
 			t.Errorf("Test %d: %s:  Expected to pass, but failed with: <ERROR> %s", i+1, instanceType, err.Error())
 		}


krisis (approved):

cmd/object-api-listobjects_test.go
```diff
@@ -446,13 +446,13 @@ func testListObjects(obj ObjectLayer, instanceType string, t TestErrHandler) {
...
 		// Setting maxKeys to a very large value (17).
-		{"empty-bucket", "", "", "", 1111000000000000, ListObjectsInfo{}, nil, true},
+		{"empty-bucket", "", "", "", 111100000, ListObjectsInfo{}, nil, true},
 		// Testing for all 7 objects in the bucket (18).
 		{"test-bucket-list-object", "", "", "", 9, resultCases[0], nil, true},
 		//Testing for negative value of maxKey, this should set maxKeys to listObjectsLimit (19).
 		{"test-bucket-list-object", "", "", "", -1, resultCases[0], nil, true},
 		// Testing for very large value of maxKey, this should set maxKeys to listObjectsLimit (20).
-		{"test-bucket-list-object", "", "", "", 1234567891011, resultCases[0], nil, true},
+		{"test-bucket-list-object", "", "", "", 1234567890, resultCases[0], nil, true},
```
balamurugana, line 455: For `List Objects`, max key value is 1000 as default.  However the request may contain <= 1000.   According to these test cases, it can be > 1000 and server shouldn't fail (looks like) rather returns 1000 entries if available.

Similarly according to the spec, `max-keys` is a String containing numbers.  When max-keys string is converted to int32 or int64 (depending on arch) we would need to return proper value to below layers as int

With this understanding, I am not able to understand these test cases.

harshavardhana: AWS S3 returns an error for maxKeys > 2147483647 or being negative.  If maxKeys is > 1000 and with in the say  0 < maxKeys < 2147483647 then AWS S3 doesn't throw any error it just ends up defaulting for maxKeys == 1000 

```
Argument maxKeys must be an integer between 0 and 2147483647
```

These tests are historical and should be removed eventually.. @balamurugana perhaps this is fix is okay for now?

balamurugana: This means `maxKey` in test case should be `int32` where `maxKey` should not be negative number.

harshavardhana: Yes @balamurugana

donatello: Fixing it by changing the type to `uint32` is relatively complex change; I don't think it is urgent to do that, because all valid max-keys parameters will work, and the worst problem is that the max-keys parameter fails to parse, leading to a max-keys value of 0, and finally ending up with a error response.

I suggest taking it up later. In the meanwhile, shall we merge this so that the test breakage on 32-bit CPUs is fixed?

harshavardhana: Sure.. @donatello

harshavardhana: Waiting on @balamurugana

balamurugana: @donatello as per our discussion, you could change (at least)  for `maxKeys int32` in test cases. That could be interim fix than changing value.  This way no one can add/modify maxKeys value beyond `int32`

donatello: Ah you meant to change type in the tests - now got your point. Will look into it.

donatello: Done now.

harshavardhana (approved):

balamurugana (approved):

PR #4741, opened 2017-07-31
Fix sysctl proposed values

lnicola:
`sched_wakeup_granularity_ns` and `sched_wakeup_granularity_ns` are measured in `ns`, so a value of `10` or `15` is way too low.


harshavardhana (approved):
LGTM - thanks for the fix.

nitisht (approved):
LGTM

Issue #4727, opened 2017-07-26
completeMultipartUpload completes the upload but reports NoSuchKey error


kyleburnett:
## Expected Behavior
Theoretically, the test.js file in the gist provided should run forever.

## Current Behavior
However, eventually it outputs the following error:

```
NoSuchKey: The specified key does not exist.
    at Request.extractError (/Users/kyle/Documents/scripts/minio-test/node_modules/aws-sdk/lib/services/s3.js:568:35)
    at Request.callListeners (/Users/kyle/Documents/scripts/minio-test/node_modules/aws-sdk/lib/sequential_executor.js:105:20)
    at Request.emit (/Users/kyle/Documents/scripts/minio-test/node_modules/aws-sdk/lib/sequential_executor.js:77:10)
    at Request.emit (/Users/kyle/Documents/scripts/minio-test/node_modules/aws-sdk/lib/request.js:682:14)
    at Request.transition (/Users/kyle/Documents/scripts/minio-test/node_modules/aws-sdk/lib/request.js:22:10)
    at AcceptorStateMachine.runTo (/Users/kyle/Documents/scripts/minio-test/node_modules/aws-sdk/lib/state_machine.js:14:12)
    at /Users/kyle/Documents/scripts/minio-test/node_modules/aws-sdk/lib/state_machine.js:26:10
    at Request.<anonymous> (/Users/kyle/Documents/scripts/minio-test/node_modules/aws-sdk/lib/request.js:38:9)
    at Request.<anonymous> (/Users/kyle/Documents/scripts/minio-test/node_modules/aws-sdk/lib/request.js:684:12)
    at Request.callListeners (/Users/kyle/Documents/scripts/minio-test/node_modules/aws-sdk/lib/sequential_executor.js:115:18)
    at Request.emit (/Users/kyle/Documents/scripts/minio-test/node_modules/aws-sdk/lib/sequential_executor.js:77:10)
    at Request.emit (/Users/kyle/Documents/scripts/minio-test/node_modules/aws-sdk/lib/request.js:682:14)
    at Request.transition (/Users/kyle/Documents/scripts/minio-test/node_modules/aws-sdk/lib/request.js:22:10)
    at AcceptorStateMachine.runTo (/Users/kyle/Documents/scripts/minio-test/node_modules/aws-sdk/lib/state_machine.js:14:12)
    at /Users/kyle/Documents/scripts/minio-test/node_modules/aws-sdk/lib/state_machine.js:26:10
[...]

Issue #4658, opened 2017-07-10
Explore go cover tools for code coverage


deekoder:
Our code coverage plugin is showing conflicting numbers. It also shows failures in unrelated files not part of the PR. Having this be unreliable makes it really hard for the PR merger to merge with confidence.

brendanashworth:
The issue here doesn't lie with codecov (= switching to go cover will not help), it lies with our own tests. A few of our tests (file system related especially) won't always cover the exact same code path when they execute. There may be a race condition here or there that will trigger a different code path while still having the same end result, which gives different code coverage between equal runs while still passing the tests.

A good example of this is in `posix.go`. There is a single [`return nil` statement](https://codecov.io/gh/minio/minio/src/ecc7113b0257f37e026f65aaa20df05066904fef/cmd/posix.go#L95) which is only called under [a race condition](https://github.com/minio/minio/issues/3011). The race condition handling code was [added](https://github.com/minio/minio/commit/95567c68bfdc1c198125480bcb1e13d9d280c77f), but it wasn't added with a specific test that always triggered that code path. This means that, out of the [20,000 times `isDirEmpty` will be called](https://codecov.io/gh/minio/minio/src/ecc7113b0257f37e026f65aaa20df05066904fef/cmd/posix.go#L76) throughout the whole test suite, the race condition may only appear once, and it will do so at random. If it happened on master but not on a pull request, the coverage will decrease minutely; if it didn't happen on master but did on a pull request, it will increase minutely.

Because of these tests that have somewhat flaky code paths, the test coverage should have an error margin. Roughly within `~0.05%`, the test coverage is valid and should be taken into consideration, especially if the files are related to the change. If the code coverage is something like `+0.01%`, it is probably unrelated.

deekoder:
Thanks brendan! This explains the rootcause really well.

brendanashworth:
Here's a list of common culprit files:

- [x] ~~[`cmd/posix.go`](https://github.com/minio/minio/blob/ec5293ce29ef03726c7174d9e3304bc156259c4e/cmd/posix.go#L95) ([example](https://codecov.io/gh/minio/minio/pull/4695/src/cmd/posix.go#L95))~~ https://github.com/minio/minio/commit/28bc5899fdea9dd433c10ae480320c58ea48f093
[...]

PR #4742, opened 2017-07-31
cmd: fix race in fsDeleteFile, make isDirEmpty deterministic, drop stat in deleteFile, DRY

brendanashworth:
## Description
```
commit fdbb13799b411b539fca7d9e5b9c05211da2545d
Author: Brendan Ashworth <brendan.ashworth@me.com>
Date:   Mon Jul 31 15:54:22 2017 -0700

    posix: do not upstream errors in deleteFile
    
    This commit changes posix's deleteFile() to not upstream errors from
    removing parent directories. This fixes a race condition.
    
    The race condition occurs when multiple deleteFile()s are called on the
    same parent directory, but different child files. Because deleteFile()
    recursively removes parent directories if they are empty, but
    deleteFile() errors if the selected deletePath does not exist, there was
    an opportunity for a race condition. The two processes would remove the
    child directories successfully, then depend on the parent directory
    still existing. In some cases this is an invalid assumption, because
    other processes can remove the parent directory beforehand. This commit
    changes deleteFile() to not upstream an error if one occurs, because the
    only required error should be from the immediate deletePath, not from a
    parent path.
    
    In the specific bug report, multiple CompleteMultipartUpload requests
    would launch multiple deleteFile() requests. Because they chain up on
    parent directories, ultimately at the end, there would be multiple
    remove files for the ultimate parent directory,
    .minio.sys/multipart/{bucket}. Because only one will succeed and one
    will fail, an error would be upstreamed saying that the file does not
    exist, and the CompleteMultipartUpload code interpreted this as
    NoSuchKey, or that the object/part id doesn't exist. This was faulty
    behavior and is now fixed.
    
    Fixes: https://github.com/minio/minio/issues/4727

commit 52e058583012560c1d7b26a191ee904e422fbb5f
Author: Brendan Ashworth <brendan.ashworth@me.com>
Date:   Fri Jul 28 15:34:01 2017 -0700

    fs: drop Stat() call from fsDeleteFile,deleteFile
    
    This commit makes fsDeleteFile() simply call deleteFile() after calling
    the relevant path length checking functions. This DRYs the code base.
    
    This commit removes the Stat() call from deleteFile(). This improves
    performance and removes any possibility of a race condition.
    
    This additionally adds tests and a benchmark for said function. Here are
    the results from before and after this commit:
    
    benchmark                   old ns/op     new ns/op     delta
    BenchmarkFSDeleteFile-4     2694884       1900661       -29.47%

commit 1778101f5e830f0afea18de67d5c8ce10e4af972
Author: Brendan Ashworth <brendan.ashworth@me.com>
Date:   Thu Jul 20 17:13:15 2017 -0700

    posix: test isDirEmpty, change error conditional
    
    This commit adds a new test for isDirEmpty (for code coverage) and
    changes around the error conditional. Previously, there was a `return
    nil` statement that would only be triggered under a race condition and
    would trip up our test coverage for no reason. With this new error
    conditional, there's no awkward 'else'-esque condition, which means test
    coverage will not change between runs for no reason in this specific
    test. It's also a cleaner read.
```

## Motivation and Context
Fixes the race condition in https://github.com/minio/minio/issues/4727.

Fixes `posix.go` and `fs-v1-helpers.go` in https://github.com/minio/minio/issues/4658#issuecomment-318495587.

## How Has This Been Tested?

well

## Types of changes

## Checklist:

### DIFF
--- a/cmd/fs-v1-helpers.go
+++ b/cmd/fs-v1-helpers.go
@@ -363,8 +363,7 @@ func fsRenameFile(sourcePath, destPath string) error {
 	return nil
 }
 
-// Delete a file and its parent if it is empty at the destination path.
-// this function additionally protects the basePath from being deleted.
+// fsDeleteFile is a wrapper for deleteFile(), after checking the path length.
 func fsDeleteFile(basePath, deletePath string) error {
 	if err := checkPathLength(basePath); err != nil {
 		return traceError(err)
@@ -374,44 +373,7 @@ func fsDeleteFile(basePath, deletePath string) error {
 		return traceError(err)
 	}
 
-	if basePath == deletePath {
-		return nil
-	}
-
-	// Verify if the path exists.
-	pathSt, err := osStat(preparePath(deletePath))
-	if err != nil {
-		if os.IsNotExist(err) {
-			return traceError(errFileNotFound)
-		} else if os.IsPermission(err) {
-			return traceError(errFileAccessDenied)
-		}
-		return traceError(err)
-	}
-
-	if pathSt.IsDir() && !isDirEmpty(deletePath) {
-		// Verify if directory is empty.
-		return nil
-	}
-
-	// Attempt to remove path.
-	if err = os.Remove(preparePath(deletePath)); err != nil {
-		if os.IsNotExist(err) {
-			return traceError(errFileNotFound)
-		} else if os.IsPermission(err) {
-			return traceError(errFileAccessDenied)
-		} else if isSysErrNotEmpty(err) {
-			return traceError(errVolumeNotEmpty)
-		}
-		return traceError(err)
-	}
-
-	// Recursively go down the next path and delete again.
-	if err := fsDeleteFile(basePath, pathutil.Dir(deletePath)); err != nil {
-		return err
-	}
-
-	return nil
+	return deleteFile(basePath, deletePath)
 }
 
 // fsRemoveMeta safely removes a locked file and takes care of Windows special case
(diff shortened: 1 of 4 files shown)


cmd/fs-v1-helpers.go
```diff
@@ -365,6 +365,7 @@ func fsRenameFile(sourcePath, destPath string) error {
 
 // Delete a file and its parent if it is empty at the destination path.
 // this function additionally protects the basePath from being deleted.
+// This will return nil if called on a non-empty directory.
```
harshavardhana: We need to talk about this.. can you explain the reasoning? @brendanashworth

brendanashworth: Yes, this isn't actually a change in behavior. I added this comment to clarify what the function does (hence the confusion 😄 ), as [before](https://github.com/minio/minio/blob/ec5293ce29ef03726c7174d9e3304bc156259c4e/cmd/fs-v1-helpers.go#L392-L395) it would `return nil` from the `Stat()` call. This actually led to an interesting race condition (never became an issue afaik), because if `Stat()` returned a non-empty directory, but was empty by the time the `Remove()` call came around, [it would error](https://github.com/minio/minio/blob/ec5293ce29ef03726c7174d9e3304bc156259c4e/cmd/fs-v1-helpers.go#L404). So technically this PR fixes that race condition, even though I doubt it would ever come to fruition in a production environment.

Also, all tests that are written for this commit pass both before and after the changes. This was done to prevent any change in behavior.

brendanashworth:
Running the CI uncovers a porting issue on Windows, looking into this now.

harshavardhana:
Its better to send three separate PRs here @brendanashworth - they are all unrelated.

brendanashworth:
Yes @harshavardhana I will do so, thanks.

PR #4743, opened 2017-07-31
posix: test isDirEmpty, change error conditional

brendanashworth:
## Description
This commit adds a new test for `isDirEmpty` (for code coverage) and
changes around the error conditional. Previously, there was a `return
nil` statement that would only be triggered under a race condition and
would trip up our test coverage for no reason. With this new error
conditional, there's no awkward 'else'-esque condition, which means test
coverage will not change between runs for no reason in this specific
test. It's also a cleaner read.

## Motivation and Context
Fixes `cmd/posix.go` in https://github.com/minio/minio/issues/4658.

## How Has This Been Tested?
Tests have been added.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/posix_test.go
+++ b/cmd/posix_test.go
@@ -73,6 +73,42 @@ func TestPosixGetDiskInfo(t *testing.T) {
 	}
 }
 
+func TestPosixIsDirEmpty(t *testing.T) {
+	tmp, err := ioutil.TempDir(globalTestTmpDir, "minio-")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer removeAll(tmp)
+
+	// Should give false on non-existent directory.
+	dir1 := slashpath.Join(tmp, "non-existent-directory")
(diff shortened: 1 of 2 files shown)


aead (approved):
LGTM, nice find!

cmd/posix_test.go
```diff
@@ -73,6 +73,31 @@ func TestPosixGetDiskInfo(t *testing.T) {
 	}
 }
 
+func TestPosixIsDirEmpty(t *testing.T) {
```
vadmeste, line 76: Can you add a successful test case? an empty directory and isEmpty() returning true.

vadmeste (changes requested):
One minor comment

cmd/posix_test.go
brendanashworth: Ah yes of course! Added, thanks.

vadmeste (approved):
LGTM

PR #4744, opened 2017-07-31
fs: drop Stat() call from fsDeleteFile,deleteFile

brendanashworth:
## Description
This commit makes `fsDeleteFile()` simply call `deleteFile()` after calling
the relevant path length checking functions. This DRYs the code base.

This commit removes the `Stat()` call from `deleteFile()`. This improves
performance and removes any possibility of a race condition.

This additionally adds tests and a benchmark for said function. The
results aren't very consistent, although I'd expect this commit to make
it faster.

Also uses `io.StartSeek`.

## Motivation and Context
Fixes `cmd/fs-v1-helpers.go` in https://github.com/minio/minio/issues/4658#issuecomment-318495587.

## How Has This Been Tested?
Tests are added.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/fs-v1-helpers_test.go
+++ b/cmd/fs-v1-helpers_test.go
@@ -18,6 +18,7 @@ package cmd
 
 import (
 	"bytes"
+	"io"
 	"io/ioutil"
 	"os"
 	"path"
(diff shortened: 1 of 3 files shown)


cmd/fs-v1-helpers_test.go
```diff
@@ -265,48 +265,121 @@ func TestFSDeletes(t *testing.T) {
 	// Seek back.
 	reader.Seek(0, 0)
```
aead: We may want to use io.SeekStart  (0 is a magic constant)
It's not directly in your PR but maybe you can patch it, too?

aead (changes requested):

cmd/fs-v1-helpers_test.go
brendanashworth: So this, right?
```go
reader.Seek(0, io.SeekStart)
```

harshavardhana: > reader.Seek(0, io.SeekStart)

:+1:

brendanashworth:
Updated, PTAL @aead @harshavardhana

aead (approved):
LGTM
Nice simplification :+1:

cmd/posix.go
```diff
@@ -915,39 +915,28 @@ func (s *posix) StatFile(volume, path string) (file FileInfo, err error) {
...
-		return nil
-	}
+
 	// Attempt to remove path.
 	if err := os.Remove(preparePath(deletePath)); err != nil {
 		if os.IsNotExist(err) {
 			return errFileNotFound
 		} else if os.IsPermission(err) {
 			return errFileAccessDenied
+		} else if isSysErrNotEmpty(err) {
```
harshavardhana: We need to talk about this .. why we need to return `nil` in this scenario.

harshavardhana (changes requested):

cmd/posix.go
```diff
@@ -925,12 +925,14 @@ func deleteFile(basePath, deletePath string) error {
 
 	// Attempt to remove path.
 	if err := os.Remove(preparePath(deletePath)); err != nil {
+		if isSysErrNotEmpty(err) {
```
harshavardhana: We need to add a comment.. on why this is an ignored error.

brendanashworth: Commented added, ptal

harshavardhana (approved):

Issue #4715, opened 2017-07-24
Crash in FS mode when Erasure coded mode is already running


harshavardhana:
## Expected Behavior
```
ERRO[0000] Initializing object layer failed              cause=FS backend format required. Found 'xl' source=[server-main.go:206:serverMain()]
```

## Current Behavior
```
$ minio server /tmp/1
ERRO[0000] Initializing object layer failed              cause=FS backend format required. Found 'xl' source=[server-main.go:206:serverMain()]
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x8daaf1]

goroutine 1 [running]:
panic(0xe898c0, 0xc420010120)
	/home/harsha/.gimme/versions/go1.7.5.linux.amd64/src/runtime/panic.go:500 +0x1a1
github.com/minio/minio/pkg/http.(*httpListener).Close(0x0, 0x0, 0x0)
	/home/harsha/mygo/src/github.com/minio/minio/pkg/http/listener.go:227 +0x41
github.com/minio/minio/pkg/http.(*Server).Shutdown(0xc4201784e0, 0x0, 0x0)
	/home/harsha/mygo/src/github.com/minio/minio/pkg/http/server.go:128 +0xa6
github.com/minio/minio/cmd.serverMain(0xc4201e7340)
	/home/harsha/mygo/src/github.com/minio/minio/cmd/server-main.go:207 +0x89e
github.com/minio/minio/vendor/github.com/minio/cli.HandleAction(0xe239e0, 0x1071ba8, 0xc4201e7340, 0xc420181600, 0x0)
	/home/harsha/mygo/src/github.com/minio/minio/vendor/github.com/minio/cli/app.go:499 +0x7c
github.com/minio/minio/vendor/github.com/minio/cli.Command.Run(0xfccc8f, 0x6, 0x0, 0x0, 0x0, 0x0, 0x0, 0xfe6756, 0x1c, 0x0, ...)
	/home/harsha/mygo/src/github.com/minio/minio/vendor/github.com/minio/cli/command.go:214 +0xb9a
github.com/minio/minio/vendor/github.com/minio/cli.(*App).Run(0xc420065d40, 0xc42000a270, 0x3, 0x3, 0x0, 0x0)
	/home/harsha/mygo/src/github.com/minio/minio/vendor/github.com/minio/cli/app.go:260 +0x816
[...]

PR #4753, opened 2017-08-03
Stop attempting to close nil Listener

ebozduman:
## Description
If HTTP server is not up and running and if the Listener is attempted to be closed when the server could not even started up, then the offending panic/crash happens. 
To prevent crash from happening, this fix introduces a new HTTP server variable `upAndRunning` to keep track of the server's state. When we try to shut down the server, we just check if it was up before moving ahead with cleaning things up. This prevents the panic causing close attempt on the non-existing Listener.

## Motivation and Context
Crash in FS mode when Erasure coded mode is already running [#4715](https://github.com/minio/minio/issues/4715)

## How Has This Been Tested?
Started Minio in distributed mode and tried to start a 2nd Minio with one of the disks/resources that has been already used by the other Minio instance.
Also tested the opposite case.

## Types of changes

## Checklist:

### DIFF
--- a/pkg/http/server.go
+++ b/pkg/http/server.go
@@ -118,6 +118,10 @@ func (srv *Server) Start() (err error) {
 
 // Shutdown - shuts down HTTP server.
 func (srv *Server) Shutdown() error {
+	if srv.listener == nil {
+		return errors.New("server not initialized")
+	}
+
 	if atomic.AddUint32(&srv.inShutdown, 1) > 1 {
 		// shutdown in progress
 		return errors.New("http server already in shutdown")


pkg/http/server.go
```diff
@@ -123,9 +125,14 @@ func (srv *Server) Shutdown() error {
 		return errors.New("http server already in shutdown")
 	}
 
+	if atomic.LoadInt32(&srv.upAndRunning) <= 0 {
+		return nil
+	}
+
 	// Close underneath HTTP listener.
 	srv.listenerMutex.Lock()
 	err := srv.listener.Close()
```
balamurugana: You could check `srv.listener == nil` than bringing additional field to the structure.

balamurugana (changes requested):

pkg/http/server.go
ebozduman: Checking Listener for nil was my first approach, but then I discussed it with @harshavardhana and he suggested to implement the fix this way by introducing additional field in the structure. 
As a continuation, the goal is to refactor the code more. So, the plan is to get rid of the mutex logic from Listener code/design. We can discuss it further on the phone if you'd like.

harshavardhana: I see now @ebozduman  checking for listener as nil is fine if we are not resetting or reading the value elsewhere for this fix.

The refactor discussions requires lot more work we can defer that discussion for later date.

ebozduman: Done. 
@balamurugana , @harshavardhana , @donatello ,
Please review it again.
Thanks.

pkg/http/server.go
```diff
@@ -123,6 +123,10 @@ func (srv *Server) Shutdown() error {
 		return errors.New("http server already in shutdown")
 	}
 
+	if srv.listener == nil {
+		return nil
```
balamurugana: 1. A racy Start() and Stop() would set `srv.inShutdown` incremented.  It would look like the server is in shutdown. You could move this check above `if atomic.AddUint32(...` block.

2. How about returning an error `server not initialized` than `nil`?

balamurugana (changes requested):

balamurugana (approved):

balamurugana:
commit message like `Stop attempting to close nil Listener` much clearer.

donatello (approved):

PR #4760, opened 2017-08-03
XL: Simplify heal-format operations

donatello:
This is in preparation for updated admin heal API.

* Improve case analysis of healFormatXL() - fixes a case where disks
  could have unhandled errors.

* Simplify healFormatXLFreshDisks() and healFormatXLCorruptedDisks()
  to share more code and handle fewer cases for improved simplicity
  and reduced code repetition.

* Fix test cases.

## Description

Existing code is very complex and with a lot of repetition. This patch simplifies the code and fixes cases where disks may have unhandled errors (i.e. other than disk-not-found, corrupted-format and unformatted-disk).

## Motivation and Context

When implementing the Admin heal APIs it was necessary to refactor lower-level heal functionality to make it provide more details about items being healed. This code simplification was done while working on the admin-heal APIs, but however was found to be easy to integrate independently of the heal APIs

## How Has This Been Tested?

No new functionality is added and existing tests have been fixed. A new test is added for the unexpected error handling case.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/format-config-v1.go
+++ b/cmd/format-config-v1.go
@@ -241,48 +241,32 @@ else // No healing at this point forward, some disks are offline or dead.
 fi
 */
 
-// error returned when some disks are found to be unformatted.
-var errSomeDiskUnformatted = errors.New("some disks are found to be unformatted")
-
 // error returned when some disks are offline.
 var errSomeDiskOffline = errors.New("some disks are offline")
 
 // errDiskOrderMismatch - returned when disk UUID is not in consistent JBOD order.
 var errDiskOrderMismatch = errors.New("disk order mismatch")
 
-// Returns error slice into understandable errors.
-func reduceFormatErrs(errs []error, diskCount int) (err error) {
-	var errUnformattedDiskCount = 0
-	var errDiskNotFoundCount = 0
-	var errCorruptedFormatCount = 0
-	for _, dErr := range errs {
-		if dErr == errUnformattedDisk {
-			errUnformattedDiskCount++
-		} else if dErr == errDiskNotFound {
-			errDiskNotFoundCount++
-		} else if dErr == errCorruptedFormat {
-			errCorruptedFormatCount++
+// formatErrsSummary - summarizes errors into different classes
+func formatErrsSummary(errs []error) (formatCount, unformattedDiskCount,
+	diskNotFoundCount, corruptedFormatCount, otherErrCount int) {
+
+	for _, err := range errs {
+		switch err {
+		case errDiskNotFound:
+			diskNotFoundCount++
+		case errUnformattedDisk:
+			unformattedDiskCount++
+		case errCorruptedFormat:
+			corruptedFormatCount++
+		case nil:
+			// implies that format is not nil
+			formatCount++
+		default:
+			otherErrCount++
 		}
 	}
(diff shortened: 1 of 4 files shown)


cmd/xl-v1-healing.go
```diff
@@ -36,21 +36,41 @@ func healFormatXL(storageDisks []StorageAPI) (err error) {
...
+		// all unformatted.
+		if err = initFormatXL(storageDisks); err != nil {
+			return err
+		}
+
+	// No format heal action to take?!
+	case unformattedDiskCount == 0 && corruptedFormatCount == 0:
+		return nil
+
+	case diskNotFoundCount > 0:
```
vadmeste: I am thinking that we should test this case `case diskNotFoundCount > 0:` before this `case unformattedDiskCount == 0 && corruptedFormatCount == 0:`, in we can return an error to the user

cmd/format-config-v1.go
```diff
@@ -469,29 +453,47 @@ func findDiskIndex(disk string, jbod []string) int {
...
-			return nil, errors.New("Unrecognized uuid " + format.XL.Disk + " found")
+			return nil, nil, errors.New("Unrecognized uuid " + format.XL.Disk + " found")
 		}
 		newDisks[jIndex] = bootstrapDisks[fIndex]
 	}
-	return newDisks, nil
+
+	// Based on orderedDisks generate new UUIDs in the ref. config
+	// for disks without format-configs.
+	for index, disk := range newDisks {
```
vadmeste: I am not sure if we need to to assign new format config to nil disks.. at least this doesn't comply with function name orderDisks()

cmd/xl-v1-healing.go
donatello: Agreed - updated code.

cmd/format-config-v1.go
donatello: This was done to improve code sharing between `healFormatXL*()` funcs, but this function is also called from `loadFormatXL()` which has no need of the assigning UUIDs. So now I've added an extra flag to skip this step in case of loadFormatXL. The function is not renamed, as it would be too long to call it something like `reorderDisksAndAssignUUIDsToUnassignedDisks` - but the extra flag is now documented in the function description.

donatello:
@vadmeste @krisis Review comments addressed!

krisis (approved):

cmd/format-config-v1.go
```diff
@@ -658,37 +618,37 @@ func healFormatXLFreshDisks(storageDisks []StorageAPI) error {
...
 	}
-	return assignedDisks, unAssignedDisks
+	return uDisks
 }
 
-// Inspect the content of all disks to guess the right order according to the format files.
-// The right order is represented in orderedDisks
-func reorderDisksByInspection(orderedDisks, storageDisks []StorageAPI, formatConfigs []*formatConfigV1) ([]StorageAPI, error) {
+// Inspect the content of all disks to guess the right order according
+// to the format files. The right order is represented in orderedDisks
```
harshavardhana, line 648: We can avoid these formatting changes @donatello and do it later as and when we are modifying the code paths. Perhaps no reason to bring them here?

vadmeste (approved):
LGTM & tested

Issue #4494, opened 2017-06-07
Unable to peek into incoming protocol


deekoder:
Sudden printing of these messages. Not really helpful for me as the user to figure out why this message shows up. 

## Expected Behavior
Some indication of what caused it would be more insightful to consume for me as the user. Also they seem to always print in pairs.
 
## Possible Solution
Not print it ?
 
## Steps to Reproduce (for bugs)
 Unsure, it is printed often when the server starts and minio browser is loaded. I am not sure of the exact steps
 

## Your Environment

* Version used (`minio version`): 2017-06-02T22 
* Environment name and version (e.g. nginx 1.9.1):
* Server type and version: 
* Operating System and version (`uname -a`): Windows Server 2016 on aws.
* Link to your project:

balamurugana:
@deekoder Do you mean error message "Unable to peek into incoming protocol" is printed when minio server is running?

trivita:
i have the same issue here
```
ERRO[0259] Unable to peek into incoming protocol         cause=read tcp 10.100.10.107:9000->10.4.19.35:51106: i/o timeout source=[server-mux.go:262:newListenerMux.func1.1()]
ERRO[0260] Unable to peek into incoming protocol         cause=read tcp 10.100.10.107:9000->10.4.19.229:51475: i/o timeout source=[server-mux.go:262:newListenerMux.func1.1()]
ERRO[0278] Unable to peek into incoming protocol         cause=read tcp 10.100.10.107:9000->10.4.19.35:55431: i/o timeout source=[server-mux.go:262:newListenerMux.func1.1()]
ERRO[0278] Unable to peek into incoming protocol         cause=read tcp 10.100.10.107:9000->10.4.19.35:55432: i/o timeout source=[server-mux.go:262:newListenerMux.func1.1()]

```

harshavardhana:
Re-opening this bug as re-surfaced with a newer log message. We need to fix this properly.  // cc @abperiasamy 

```
ERRO[0016] Error in reading from new TLS connection 67.180.17.191:41370 at server 171.67.92.55:9000  cause=EOF source=[listener.go:170:github.com/minio/minio/pkg/http.(*httpListener).start.func2()]
ERRO[0039] Error in reading from new TLS connection 67.180.17.191:41408 at server 171.67.92.55:9000  cause=EOF source=[listener.go:170:github.com/minio/minio/pkg/http.(*httpListener).start.func2()]
[...]

Issue #4748, opened 2017-08-02
Closing browser tab should warn user about background activity


balamurugana:
Based on https://github.com/minio/minio/issues/4670 and https://github.com/minio/minio/issues/4494, Closing browser tab should warn user about background activity like GMail.  Currently it causes unexpected EOF on client connections at server side .

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

* Version used (`minio version`):
* Environment name and version (e.g. nginx 1.9.1):
* Server type and version:
* Operating System and version (`uname -a`):
* Link to your project:

deekoder:
@balamurugana  i assume this is for the error message printed on the server. However I am not sure how you want to tell the user about the background processes running when they close the browser. Couldnt we simple stop the background activity if they decided to close the tab. I think we can take the tab closure as an intent to close background processes?

balamurugana:
@deekoder I think we would need to do exactly what GMail handles this case.

deekoder:
Can you elaborate how gmail handles it. Is there a warning pop up on the ux for eg?

balamurugana:
Yes. GMail shows warning popup about background tasks on tab closing.

krisis:
@balamurugana what is the background task in the case of minio browser?

balamurugana:
@krisis I don't know.  If you see issues https://github.com/minio/minio/issues/4670 and https://github.com/minio/minio/issues/4494, closing browser tab causes EOF client connections (continuously).  This is clear evidence that browser is doing some tasks.

deekoder:
BTW that happens for me even without closing the browser.

brendanashworth:
I'm looking into the issue. This only happens when using chrome.

brendanashworth:
There are two issues here. The first is closing the browser when backgrounds tasks are occurring. This is separate from the "Error in reading from new connection" error in https://github.com/minio/minio/pull/4762. The following explanation is for the "reading from new connection" bug.

I believe the "issue" here is that Chrome is opening up connections pre-emptively, so as to make connecting to the web page faster. These EOFs need to be handled gracefully, as when a pre-emptive connection isn't used, Chrome will just give it EOF.

[...]

PR #4762, opened 2017-08-03
Avoid superfluous error messages after connect

krisis:
Peek could fail legitimately when clients abruptly close connection.

## Description

Previously, minio/pkg/http server would log errors if clients closed connection immediately. These are not actionable errors. This change removes logging of these errors.
 
## Motivation and Context

See #4494 

## How Has This Been Tested?

Manually

## Types of changes

## Checklist:

### DIFF
--- a/pkg/http/listener.go
+++ b/pkg/http/listener.go
@@ -20,6 +20,7 @@ import (
 	"crypto/tls"
 	"errors"
 	"fmt"
+	"io"
 	"net"
 	"net/http"
 	"strings"
@@ -88,6 +89,16 @@ type httpListener struct {
 	errorLogFunc           func(error, string, ...interface{}) // function to be called on errors.
 }
 
+// ignoreErr returns true if error is due to a network timeout or
+// io.EOF and false otherwise
+func ignoreErr(err error) bool {
+	if nErr, ok := err.(*net.OpError); ok {
+		return nErr.Timeout()
+	}
+
+	return err == io.EOF
+}
+
 // start - starts separate goroutine for each TCP listener.  A valid insecure/TLS HTTP new connection is passed to httpListener.acceptCh.
 func (listener *httpListener) start() {
 	listener.acceptCh = make(chan acceptResult)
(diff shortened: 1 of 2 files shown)


brendanashworth:
@krisis take a look at my explanation as to what I think this issue is: https://github.com/minio/minio/issues/4748#issuecomment-320114898. We don't need to comment out this code, instead we should check if the error is `io.EOF`, and if it is, ignore it.

pkg/http/listener.go
```diff
@@ -88,6 +89,23 @@ type httpListener struct {
 	errorLogFunc           func(error, string, ...interface{}) // function to be called on errors.
 }
 
+// shouldLogError returns false if error is due to a network timeout or
+// io.EOF and true otherwise
+func shouldLogError(err error) bool {
```
brendanashworth: Because no logger is present in the package scope (pkg/http), perhaps we could name this function `isUnexpected()`?

pkg/http/listener.go
```diff
@@ -88,6 +89,23 @@ type httpListener struct {
...
+	switch err := err.(type) {
```
brendanashworth: I think this can be done more concisely like so:
```go
netErr, ok := err.(net.Error)
if ok {
        if err.Timeout() {
                return false
        }
}

return err != io.EOF
```

brendanashworth (changes requested):

brendanashworth (approved):
This looks nice. The commit message could use a little bit of expansion, maybe like "This now swallows EOF and network timeout errors when no data is sent."

pkg/http/listener.go
```diff
@@ -88,6 +89,19 @@ type httpListener struct {
 	errorLogFunc           func(error, string, ...interface{}) // function to be called on errors.
 }
 
+// isExpectedErr returns true if error is due to a network timeout or
+// io.EOF and false otherwise
+func isExpectedErr(err error) bool {
```
balamurugana: 1. I think this function as closure is more appropriate (you could remove unit test also)
2. you could name this function as 'ignoreErr()`

pkg/http/listener.go
```diff
@@ -88,6 +89,19 @@ type httpListener struct {
...
+	if nErr, ok := err.(*net.OpError); ok {
+		if nErr.Timeout() {
```
balamurugana: return `nErr.Timeout()` is enough.

balamurugana (changes requested):

pkg/http/listener.go
harshavardhana: :+1:

harshavardhana (changes requested):

pkg/http/listener.go
krisis: @balamurugana  these unit tests additionally document the errors we avoid logging via test cases. So, I intend to retain it as a separate function with the unit tests.

harshavardhana (approved):

balamurugana (approved):

PR #4767, opened 2017-08-04
set http transport config for gateway (#4765)

### LINKED ISSUE #4765: gateway: MaxIdleConns is not configured
In gateway-mode minio does not set the transport configuration. The server uses the client side functionality to talk to the GCS/Azure/S3 but does not configure the http.Transport settings (MaxIdleConns).

## Expected Behavior
Minio server sets transport configuration   

## Current Behavior
Minio server does not set transport configuration   

## Possible Solution
Use client side config in server

aead:
## Description

This change sets the http config for the minio client used by the
minio server in gateway mode.

Fixes #4765

## Motivation and Context
The server should set not garbage-collect all connections (except 2 - default). This is a performance
optimization  for the gateway layer.

## Checklist:

### DIFF
--- a/cmd/gateway-anonymous.go
+++ b/cmd/gateway-anonymous.go
@@ -16,7 +16,12 @@
 
 package cmd
 
-import "net/http"
+import (
+	"crypto/tls"
+	"net"
+	"net/http"
+	"time"
+)
 
 func anonErrToObjectErr(statusCode int, params ...string) error {
 	bucket := ""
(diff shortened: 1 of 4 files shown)


cmd/gateway-s3.go
```diff
@@ -130,6 +133,19 @@ func newS3Gateway(host string) (GatewayLayer, error) {
...
+		DialContext: (&net.Dialer{
+			Timeout:   30 * time.Second,
+			KeepAlive: 30 * time.Second,
+		}).DialContext,
+		MaxIdleConns:          100,
+		IdleConnTimeout:       90 * time.Second,
+		TLSHandshakeTimeout:   10 * time.Second,
+		ExpectContinueTimeout: 1 * time.Second,
+		TLSClientConfig:       &tls.Config{RootCAs: globalRootCAs},
+		DisableCompression:    true,
```
balamurugana: How about a function `newCustomHTTPTransport()` returning this new object?  This would avoid repetition of code.

aead: Good point - done.

balamurugana (commented):
LGTM

cmd/gateway-anonymous.go
```diff
@@ -47,3 +52,19 @@ func anonErrToObjectErr(statusCode int, params ...string) error {
 
 	return errUnexpected
 }
+
```
harshavardhana, line 55: A comment would be better..

aead: done

harshavardhana (approved):

cmd/gateway-anonymous.go
harshavardhana: Perhaps we can say why we added this like we added this since default Go sets MaxIdleConns as 2

aead: sure, done.

harshavardhana (approved):

balamurugana (approved):

PR #4771, opened 2017-08-04
posix: do not upstream errors in deleteFile

### LINKED ISSUE #4727: completeMultipartUpload completes the upload but reports NoSuchKey error

brendanashworth:
## Description
This commit changes posix's `deleteFile()` to not upstream errors from
removing parent directories. This fixes a race condition.

The race condition occurs when multiple `deleteFile()`s are called on the
same parent directory, but different child files. Because `deleteFile()`
recursively removes parent directories if they are empty, but
`deleteFile()` errors if the selected `deletePath` does not exist, there was
an opportunity for a race condition. The two processes would remove the
child directories successfully, then depend on the parent directory
still existing. In some cases this is an invalid assumption, because
other processes can remove the parent directory beforehand. This commit
changes `deleteFile()` to not upstream an error if one occurs, because the
only required error should be from the immediate `deletePath`, not from a
parent path.

In the specific bug report, multiple `CompleteMultipartUpload` requests
would launch multiple `deleteFile()` requests. Because they chain up on
parent directories, ultimately at the end, there would be multiple
remove files for the ultimate parent directory,
`.minio.sys/multipart/{bucket}`. Because only one will succeed and one
will fail, an error would be upstreamed saying that the file does not
exist, and the `CompleteMultipartUpload` code interpreted this as
`NoSuchKey`, or that the object/part id doesn't exist. This was faulty
behavior and is now fixed.

The added test fails before this change and passes after this change.

## Motivation and Context

Fixes: https://github.com/minio/minio/issues/4727

## How Has This Been Tested?
Locally (upstream.patch has fix to `cmd/posix.go`):
```
brendanashworth in ~/Go/src/github.com/minio/minio/cmd $ go test -test.run PosixDeleteFile -v
=== RUN   TestPosixDeleteFile
--- FAIL: TestPosixDeleteFile (0.34s)
	posix_test.go:795: TestPosix case 7: Expected: "%!s(<nil>)", got: "file access denied"
FAIL
exit status 1
FAIL	github.com/minio/minio/cmd	0.493s
brendanashworth in ~/Go/src/github.com/minio/minio/cmd $ git apply ../upstream.patch
brendanashworth in ~/Go/src/github.com/minio/minio/cmd $ go test -test.run PosixDeleteFile -v
=== RUN   TestPosixDeleteFile
--- PASS: TestPosixDeleteFile (0.18s)
PASS
ok  	github.com/minio/minio/cmd	0.261s
```

The test uses permissions (parent directory lacks write permission) to fail, but it fails similarly to how a non-existent directory fails.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/posix.go
+++ b/cmd/posix.go
@@ -916,7 +916,10 @@ func deleteFile(basePath, deletePath string) error {
 	}
 
 	// Recursively go down the next path and delete again.
-	return deleteFile(basePath, slashpath.Dir(deletePath))
+	// Errors for parent directories shouldn't trickle down.
+	deleteFile(basePath, slashpath.Dir(deletePath))
+
+	return nil
 }
 
 // DeleteFile - delete a file at path.
--- a/cmd/posix_test.go
+++ b/cmd/posix_test.go
@@ -706,6 +706,17 @@ func TestPosixDeleteFile(t *testing.T) {
 		t.Fatalf("Unable to create file, %s", err)
 	}
 
+	if err = posixStorage.MakeVol("no-permissions"); err != nil {
+		t.Fatalf("Unable to create volume, %s", err.Error())
+	}
+	if err = posixStorage.AppendFile("no-permissions", "dir/file", []byte("Hello, world")); err != nil {
+		t.Fatalf("Unable to create file, %s", err.Error())
+	}
+	// Parent directory must have write permissions, this is read + execute.
+	if err = os.Chmod(pathJoin(path, "no-permissions"), 0555); err != nil {
+		t.Fatalf("Unable to chmod directory, %s", err.Error())
+	}
+
 	testCases := []struct {
 		srcVol      string
 		srcPath     string
@@ -759,6 +770,15 @@ func TestPosixDeleteFile(t *testing.T) {
 			srcPath:     "my-obj-del-0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001",
 			ioErrCnt:    0,
 			expectedErr: errFileNameTooLong,
+		},
+		// TestPosix case - 7.
+		// TestPosix case with undeletable parent directory.
+		// File can delete, dir cannot delete because no-permissions doesn't have write perms.
+		{
+			srcVol:      "no-permissions",
+			srcPath:     "dir/file",
+			ioErrCnt:    0,
+			expectedErr: nil,
 		},
 	}
 


harshavardhana (approved):

cmd/posix.go
```diff
@@ -916,7 +916,10 @@ func deleteFile(basePath, deletePath string) error {
 	}
 
 	// Recursively go down the next path and delete again.
-	return deleteFile(basePath, slashpath.Dir(deletePath))
+	// Errors for parent directories shouldn't trickle down.
+	deleteFile(basePath, slashpath.Dir(deletePath))
```
krishnasrinivas, line 920: The problem with this is if we fail to delete the file for some reason we never return error.

A better fix is to call this function `deleteDir()` which never returns error like how we do here. But delete the file instead in `DeleteFile()`

harshavardhana: We don't @krishnasrinivas os.Remove() should have removed the expected file or directory by the time it reaches here. This is only the other empty/non-empty directories. If they failed for some reason it is okay to leave them as is.. since it wouldn't hurt IMO. 

If we didn't even return error for os.Remove() then it would mean something else entirely and that is not the case here.  The problem here is that there is no locking involved when the directories are being recursively deleted upwards, which is the reason why inherently deleteFile() is racy and has unexpected results when a concurrent deleteFile() is running along with any i/o with in the same prefix.

I would suggest that we can safely do this change in most cases since our expected nature was to delete an empty directory or file.   If we did so we hope that it is possible for us to delete the top level dirs if they are empty as well, if its not say someone created a file while it is being recursively traversed we are okay to not return an error since as a context we have never held a lock on that part of the namespace anyways.

harshavardhana: In-fact this is quite nicely explained by @brendanashworth in the commit message that if deleteFile() is to work appropriately we are not supposed to return error for the top-level folders. Since by nature we have never locked that region of namespace anyways - we are only doing it based on a guess-work at that point in time.

krishnasrinivas: I see

krishnasrinivas (approved):

Issue #4766, opened 2017-08-04
Bucket URI conflicts with login page's URI when named 'login'


samluthebrave:
URI of bucket named 'login' and URI of login page are identical, making the bucket unaccessible.

## Expected Behavior
Any successfully created bucket can be accessed.

## Current Behavior
You can successfully create a bucket named 'login', and it does appear on the sidebar, but click on it won't jump into it.

## Possible Solution
Change login URI or reserve name 'login' from bucket creating.

## Steps to Reproduce (for bugs)
1. create a bucket named 'login'
2. find it on the sidebar and click it

nitisht:
@samluthebrave thank you for finding the issue. We are taking a fresh look at how to handle such scenarios. Here are the details https://github.com/minio/minio/issues/4809

ckuetbach:
Regarding [Amazon documentation](https://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html) bucket names must be at least 3 chars long.

If all UI is accessable wothin /ui/* all valid bucket names are accessable.

deekoder:
Refactoring as per https://github.com/minio/minio/issues/4809 will take care of this.

deekoder:
Whenever we refactor for newux pod we have to do so such that /login is not used in the URL.

PR #4777, opened 2017-08-05
Reserve login bucketname from web browser access.

nitisht:
## Description
A bucket named `login` will conflict with login page URI on Minio Browser. This PR reserves the bucket name `login` for browser access. 

- An attempt to create a bucket called `login` will be rejected on browser with error `All access to this Bucket is disabled`. 
- An existing directory called `login` in FS mode will be ignored in listing buckets. 
- Attempts to create `login` bucket via `mc` or `SDKs` will be successful. 

## Motivation and Context
See issue https://github.com/minio/minio/issues/4766

## How Has This Been Tested?
manually and `go test`

## Types of changes

## Checklist:

### DIFF
--- a/cmd/generic-handlers.go
+++ b/cmd/generic-handlers.go
@@ -65,10 +65,10 @@ func (h requestSizeLimitHandler) ServeHTTP(w http.ResponseWriter, r *http.Reques
 	h.handler.ServeHTTP(w, r)
 }
 
-// Reserved bucket.
+// Reserved keyword and paths.
 const (
-	minioReservedBucket     = "minio"
-	minioReservedBucketPath = "/" + minioReservedBucket
+	minioReservedKeyword    = "minio"
+	minioReservedBucketPath = "/" + minioReservedKeyword
 )
 
 // Adds redirect rules for incoming requests.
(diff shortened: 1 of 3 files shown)


brendanashworth:
@NitishT why not disable the `/minio/login` route if the user is already logged in?

nitisht:
> why not disable the /minio/login route if the user is already logged in?

In that case also, user will not be able to access the login bucket, I think functionally it has the same effect on the user.

cmd/generic-handlers.go
```diff
@@ -65,10 +65,10 @@ func (h requestSizeLimitHandler) ServeHTTP(w http.ResponseWriter, r *http.Reques
 	h.handler.ServeHTTP(w, r)
 }
 
-// Reserved bucket.
+// Reserved keyword and paths.
 const (
-	minioReservedBucket     = "minio"
-	minioReservedBucketPath = "/" + minioReservedBucket
+	minioReservedKeyword    = "minio"
```
harshavardhana, line 70: why change this? only change that is needed is in `isMinioReservedBucket`  also add a global list 

```
var reservedBucketNames  = []string{"minio", "login"}
```

harshavardhana (commented):
Also will confirm with @abperiasamy about this situation.

cmd/generic-handlers.go
nitisht: I felt it confusing to have two variables named similar to `reservedBuckets`. So changed `minioReservedBucket` to  `minioReservedKeyword`. Also `minioReservedBucket` is being used more as a reserved key word here. Hence the change. 

I will add a global list instead of the local list in `isMinioReservedBucket`

nitisht:
As discussed in https://github.com/minio/minio/issues/4809 we will take another approach to fix this issue. Final decision on the approach still needs to be taken. Will send a fresh PR based on the approach we finalize. Closing this PR.

Issue #4669, opened 2017-07-12
Minio server should wait for distributed hosts to come up


nitisht:
## Expected Behavior
In orchestration platforms like Kubernetes, remote hosts are started sequentially and all the hosts are not up initially, though they are expected to come up in a short time frame. Minio server should wait for all the hosts to come up instead of failing immediately. 

## Current Behavior
Minio server fails on remote peer lookup failure immediately. Currently Minio `docker-entrypoint.sh` script has a method to wait for hosts to come up (https://github.com/minio/minio/blob/master/buildscripts/docker-entrypoint.sh).

## Possible Solution
Bring back URL waiting in the Minio server argument validation. Wait indefinitely for URLs to come up . Remove the `docker_wait_hosts` method from `docker-entrypoint.sh` script. 

## Steps to Reproduce (for bugs)
Deploy Minio distributed mode on Kubernetes.

nitisht:
Taking this.

PR #4783, opened 2017-08-07
Wait for remote hosts to resolve

nitisht:
## Description
This PR changes the server behaviour to wait for remote hosts to resolve instead of failing on first host resolution error. Waiting is based on `exponentialbackoff` mechanism described [here](https://www.awsarchitectureblog.com/2015/03/backoff.html).

This is needed because in orchestration platforms like Kubernetes, remote hosts are started sequentially and all the hosts are not up initially, though they are expected to come up in a short time frame. 

It is difficult to identify a cap on the waiting time due to non-deterministic nature of infrastructure platforms, so the server waits infinitely for the hosts to come up, while logging the error messages to the console.  

## Motivation and Context
See issue https://github.com/minio/minio/issues/4669

## How Has This Been Tested?
manually using minikube

## Types of changes

## Checklist:


cmd/net.go
```diff
@@ -66,18 +67,43 @@ func mustGetLocalIP4() (ipList set.StringSet) {
...
+	for {
+		select {
+		case <-retryTimeCh:
+			// try to resolve remote host
+			ips, err = net.LookupIP(host)
+			if err != nil {
+				// log error and continue. We dont fail on host resolution, just log
+				// the message to console about the host not being resolveable. It is
+				// expected that user provides resolveable hosts
+				errorIf(err, "Unable to resolve host %s (elapsed %s)", host, getElapsedTime())
```
donatello: The closure does not appear to be needed - `getElapsedTime()` could just accept the start time argument.

donatello (changes requested):

cmd/net.go
nitisht: Done

cmd/endpoint_test.go
```diff
@@ -68,8 +68,6 @@ func TestNewEndpoint(t *testing.T) {
 		{"http://server:/path", Endpoint{}, -1, fmt.Errorf("invalid URL endpoint format: invalid port number")},
 		{"https://93.184.216.34:808080/path", Endpoint{}, -1, fmt.Errorf("invalid URL endpoint format: port number must be between 1 to 65535")},
 		{"http://server:8080//", Endpoint{}, -1, fmt.Errorf("empty or root path is not supported in URL endpoint")},
-		{"http://server:8080/", Endpoint{}, -1, fmt.Errorf("empty or root path is not supported in URL endpoint")},
-		{"http://server/path", Endpoint{}, -1, fmt.Errorf("lookup server" + errMsg)},
```
balamurugana: Any reason why you removed the above cases?

cmd/net.go
```diff
@@ -66,18 +67,37 @@ func mustGetLocalIP4() (ipList set.StringSet) {
...
+			} else {
+				// add to list if successfully resolved
+				for _, ip := range ips {
+					if ip.To4() != nil {
+						ipList.Add(ip.String())
+					}
+				}
+				return ipList, err
+			}
+		case <-globalServiceDoneCh:
```
balamurugana: Why do you want to handle this channel here? For this function its not required.

below code is enough
```go
doneCh := make(chan struct{})
defer close(doneCh)
var ips []net.IP
for _ := range newRetryTimerSimple(doneCh) {
    ips, err := net.LookupIP(host)
    errorIf(err, "Unable to resolve host %s (elapsed %s)", host, getElapsedTime(time.Now().Round(time.Second)))
    if err == nil {
        break
    }
}
```

cmd/net.go
```diff
@@ -66,18 +67,37 @@ func mustGetLocalIP4() (ipList set.StringSet) {
...
-			ipList.Add(ip.String())
+	//variable to list resolved IP addresses
+	var ips []net.IP
+	// channel to indicate completion of host resolution
+	doneCh := make(chan struct{})
+	// Indicate to our retry routine to exit cleanly, upon this function return.
+	defer close(doneCh)
+	// Wait on the jitter retry loop
+	retryTimeCh := newRetryTimerSimple(doneCh)
+	for {
```
balamurugana: Infinitely retrying could be problem in non-container environment.  I think we could do only for kubernetes and docker swam.

cmd/net.go
```diff
@@ -66,18 +67,37 @@ func mustGetLocalIP4() (ipList set.StringSet) {
...
+		select {
+		case <-retryTimeCh:
+			// try to resolve remote host
+			ips, err = net.LookupIP(host)
+			if err != nil {
+				// log error and continue. We dont fail on host resolution, just log
+				// the message to console about the host not being resolveable. It is
+				// expected that user provides resolveable hosts
+				errorIf(err, "Unable to resolve host %s (elapsed %s)", host, getElapsedTime(time.Now().Round(time.Second)))
```
balamurugana: 1. Why do you need to print elapsed time?
2. elapsed time calculation is wrong here.
3. you don't need a separate function elapsed time calculation.

balamurugana (changes requested):

cmd/net.go
nitisht: okay, we can do that.

cmd/net.go
nitisht: 1. I think it is important to convey to the user how long Minio server has been waiting to resolve the host. This will help to user to make a decision whether it is okay to wait further or there is some issue with the remote hosts they provided to Minio

2. How is it incorrect @balamurugana ?

cmd/net.go
nitisht: okay will update.

cmd/endpoint_test.go
nitisht: These tests were failing because server keeps on waiting for `http://server:8080/` and `"http://server/path` to resolve. So they were removed. I think it can be added back if we take the container only approach you mentioned below.

cmd/net.go
balamurugana: You are not calculating how much time resolve took it.  You are just calculating for two immediate time values.

nitisht: Now I see it, mixed it up while fixing https://github.com/minio/minio/pull/4783#discussion_r131892189, will update.

nitisht:
Address review comments, PTAL @donatello @balamurugana

cmd/net.go
```diff
@@ -66,17 +67,42 @@ func mustGetLocalIP4() (ipList set.StringSet) {
...
+		defer close(doneCh)
+		// Mark the starting time
+		startTime := time.Now().Round(time.Second)
+		// Wait on the jitter retry loop
+		for _ = range newRetryTimerSimple(doneCh) {
+			// try to resolve remote host
+			ips, err = net.LookupIP(host)
+			// log error and continue. We dont fail on host resolution, just log
+			// the message to console about the host not being resolveable.
+			errorIf(err, "Unable to resolve host %s (elapsed %s)", host, time.Now().Round(time.Second).Sub(startTime).String())
```
donatello: Please break long lines (>80 chars) wherever possible and convenient - it is easier to read. Here, the line can be broken after the host argument.

cmd/net.go
```diff
@@ -66,17 +67,42 @@ func mustGetLocalIP4() (ipList set.StringSet) {
...
+			if err == nil {
+				break
+			}
+		}
+	} else {
+		// It is expected that user provides resolveable hosts in non container environments, else we fail
```
donatello: This line can be easily broken too.

donatello (approved):

nitisht:
Closing this PR due to meddled git history, created a new PR https://github.com/minio/minio/pull/4800

PR #4798, opened 2017-08-09
fs: Avoid non-idempotent code flow in ListBuckets()

harshavardhana:
## Description
Under the call flow

```
Readdir
   +
   |
   |
   | path-entry
   |
   |
   v
StatDir
```

Existing code was written in a manner where say
a bucket/top-level directory was indeed deleted
between Readdir() and before StatDir() we would
ignore certain errors. This is not a plausible
situation and might not happen in almost all
practical cases. We do not have to look for
or interpret these errors returned by StatDir()
instead we can just collect the successful
values and return back to the client. We do not
need to pre-maturely decide on bucket access
we just let filesystem decide subsequently for
real I/O operations.

## Motivation and Context
Refer #4658

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:

### DIFF
--- a/cmd/fs-v1.go
+++ b/cmd/fs-v1.go
@@ -257,17 +257,14 @@ func (fs fsObjects) ListBuckets() ([]BucketInfo, error) {
 		}
 		var fi os.FileInfo
 		fi, err = fsStatDir(pathJoin(fs.fsPath, entry))
+		// There seems like no practical reason to check for errors
+		// at this point, if there are indeed errors we can simply
+		// just ignore such buckets and list only those which
+		// return proper Stat information instead.
 		if err != nil {
-			// If the directory does not exist, skip the entry.
-			if errorCause(err) == errVolumeNotFound {
-				continue
-			} else if errorCause(err) == errVolumeAccessDenied {
-				// Skip the entry if its a file.
-				continue
-			}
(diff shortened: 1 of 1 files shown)


cmd/fs-v1.go
```diff
@@ -255,24 +255,18 @@ func (fs fsObjects) ListBuckets() ([]BucketInfo, error) {
...
-				// Skip the entry if its a file.
-				continue
-			}
-			return nil, err
+		fi, _ := fsStatDir(pathJoin(fs.fsPath, entry))
+		// There seems like no practical reason to check for errors
+		// at this point, if there are indeed errors we can simply
+		// just ignore such buckets and list only those which
+		// return proper Stat information instead.
+		if fi != nil {
```
balamurugana: checking `err` is appropriate way than `fi` i.e. `err == nil`.  If `err` is not nil, we don't even consider what `fi` has.

balamurugana (changes requested):

cmd/fs-v1.go
harshavardhana: done @balamurugana

balamurugana (approved):

krishnasrinivas (approved):

PR #4800, opened 2017-08-09
Retry name lookup for kubernetes and docker swarm environment

nitisht:
## Description
This PR changes the server behaviour to wait for remote hosts to resolve instead of failing on first host resolution error. Waiting is based on `exponentialbackoff` mechanism described [here](https://www.awsarchitectureblog.com/2015/03/backoff.html).

This is needed because in orchestration platforms like Kubernetes, remote hosts are started sequentially and all the hosts are not up initially, though they are expected to come up in a short time frame. 

It is difficult to identify a cap on the waiting time due to non-deterministic nature of infrastructure platforms, so the server waits infinitely for the hosts to come up, while logging the error messages to the console.  

## Motivation and Context
See issue https://github.com/minio/minio/issues/4669

## How Has This Been Tested?
manually using minikube

## Types of changes

## Checklist:


nitisht:
@donatello @balamurugana addressed comments in https://github.com/minio/minio/pull/4783 here. 
Please take a look.

donatello (approved):

cmd/net.go
```diff
@@ -65,12 +67,42 @@ func mustGetLocalIP4() (ipList set.StringSet) {
...
+			// not available immediately.
+			if ips, err = net.LookupIP(host); err == nil {
+				break
+			}
+			// log error only if more than 1s elapsed
+			if time.Since(startTime) > time.Second {
+				// log the message to console about the host not being
+				// resolveable.
+				errorIf(err, "Unable to resolve host %s (%s)", host,
+					humanize.RelTime(startTime, time.Now(), "elapsed", ""))
```
balamurugana: Instead of using `time.Now()` twice (`time.Since()` is short form of `time.Now().Sub(t)`), you could use below technique
```go
startTime := time.Now()
timeZero := time.Time{}
....
....
older := time.Since(startTime)
if older > time.Second {
    errorIf(err, "Unable to resolve host %s (%s)", host, humanize.RelTime(timeZero, timeZero.Add(older), "elapsed", ""))
}
```

balamurugana (changes requested):
Minor comment related to optimization

nitisht:
Updated @balamurugana

balamurugana (approved):

deekoder:
@NitishT I think codecov/patch is stuck. I am waiting for it to complete before merging.

harshavardhana:
> @NitishT I think codecov/patch is stuck. I am waiting for it to complete before merging.

It won't complete its lost, @NitishT has to amend the commit and push again.

nitisht:
@deekoder Codecov fails here due to small decrease in coverage. This is because tests were not added in this case. Since the `IsDocker()` `IsKubernetes()` method will always fail in test environment, and new code was added after that check. Also `net.LookupHost()` will fail for all wrong host values.

PR #4807, opened 2017-08-10
Fail to start server if detected cross-device mounts.

### LINKED ISSUE #4764: Exit during boot-up in case of cross device buckets
The issue is explained better in free form rather than issue template.

Currently if a bucket is a mount point, rename from `.minio.sys/tmp` to the bucket will fail with error EXDEV (Invalid cross-device link) Hence PutObject will fail during runtime.

Hence during bootup we should check if any of the bucket's `st_dev` obtained from the stat() syscall is different from the export directory. If any such bucket is found we should fail during the boot itself.

harshavardhana:
## Description
Fail to start server if detected cross-device mounts.

## Motivation and Context
Fixes #4764

## How Has This Been Tested?
Manually on both Linux and Windows.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/server-main.go
+++ b/cmd/server-main.go
@@ -91,6 +91,7 @@ func serverHandleCmdArgs(ctx *cli.Context) {
 
 	var setupType SetupType
 	var err error
+
 	globalMinioAddr, globalEndpoints, setupType, err = CreateEndpoints(serverAddr, ctx.Args()...)
 	fatalIf(err, "Invalid command line arguments server=‘%s’, args=%s", serverAddr, ctx.Args())
 	globalMinioHost, globalMinioPort = mustSplitHostPort(globalMinioAddr)
--- a/pkg/mountinfo/mountinfo.go
+++ b/pkg/mountinfo/mountinfo.go
@@ -1,0 +1,34 @@
+/*
+ * Minio Cloud Storage, (C) 2017 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package mountinfo
+
+// mountInfo - This represents a single line in /proc/mounts.
+type mountInfo struct {
+	Device  string
+	Path    string
+	FSType  string
+	Options []string
+	Freq    string
+	Pass    string
+}
+
+func (m mountInfo) String() string {
+	return m.Path
+}
+
+// mountInfos - This represents the entire /proc/mounts.
+type mountInfos []mountInfo
--- a/pkg/mountinfo/mountinfo_linux.go
+++ b/pkg/mountinfo/mountinfo_linux.go
@@ -1,0 +1,131 @@
+// +build linux
+
+/*
+ * Minio Cloud Storage, (C) 2017 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package mountinfo
+
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+)
+
+const (
+	// Number of fields per line in /proc/mounts as per the fstab man page.
+	expectedNumFieldsPerLine = 6
+	// Location of the mount file to use
+	procMountsPath = "/proc/mounts"
+)
+
+// CheckCrossDevice - check if any list of paths has any sub-mounts at /proc/mounts.
+func CheckCrossDevice(absPaths []string) error {
+	return checkCrossDevice(absPaths, procMountsPath)
+}
+
+// Check cross device is an internal function.
+func checkCrossDevice(absPaths []string, mountsPath string) error {
+	mounts, err := readProcMounts(mountsPath)
+	if err != nil {
+		return err
+	}
+	for _, path := range absPaths {
+		if err := mounts.checkCrossMounts(path); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// CheckCrossDevice - check if given path has any sub-mounts in the input mounts list.
+func (mts mountInfos) checkCrossMounts(path string) error {
+	if !filepath.IsAbs(path) {
+		return fmt.Errorf("Invalid argument, path (%s) is expected to be absolute", path)
+	}
+	var crossMounts mountInfos
+	for _, mount := range mts {
+		// Add a separator to indicate that this is a proper mount-point.
+		// This is to avoid a situation where prefix is '/tmp/fsmount'
+		// and mount path is /tmp/fs. In such a scenario we need to check for
+		// `/tmp/fs/` to be a common prefix amount other mounts.
+		mpath := strings.TrimSuffix(mount.Path, "/") + "/"
+		ppath := strings.TrimSuffix(path, "/") + "/"
+		if strings.HasPrefix(mpath, ppath) {
+			// At this point if the mount point has a common prefix two conditions can happen.
+			// - mount.Path matches exact with `path` means we can proceed no error here.
+			// - mount.Path doesn't match (means cross-device mount), should error out.
+			if mount.Path != path {
+				crossMounts = append(crossMounts, mount)
+			}
+		}
+	}
+	msg := `Cross-device mounts detected on path (%s) at following locations %s. Export path should not have any sub-mounts, refusing to start.`
+	if len(crossMounts) > 0 {
+		// if paths didn't match then we do have cross-device mount.
+		return fmt.Errorf(msg, path, crossMounts)
+	}
+	return nil
+}
+
+// readProcMounts reads the given mountFilePath (normally /proc/mounts) and produces a hash
+// of the contents.  If the out argument is not nil, this fills it with MountPoint structs.
+func readProcMounts(mountFilePath string) (mountInfos, error) {
+	file, err := os.Open(mountFilePath)
+	if err != nil {
+		return nil, err
+	}
+	defer file.Close()
+	return parseMountFrom(file)
+}
+
+func parseMountFrom(file io.Reader) (mountInfos, error) {
+	var mounts = mountInfos{}
+	scanner := bufio.NewReader(file)
+	for {
+		line, err := scanner.ReadString('\n')
+		if err == io.EOF {
+			break
+		}
+		fields := strings.Fields(line)
+		if len(fields) != expectedNumFieldsPerLine {
+			return nil, fmt.Errorf("wrong number of fields (expected %d, got %d): %s", expectedNumFieldsPerLine, len(fields), line)
+		}
+
+		// Freq should be an integer.
+		if _, err := strconv.Atoi(fields[4]); err != nil {
+			return nil, err
+		}
+
+		// Pass should be an integer.
+		if _, err := strconv.Atoi(fields[5]); err != nil {
+			return nil, err
+		}
+
+		mounts = append(mounts, mountInfo{
+			Device:  fields[0],
+			Path:    fields[1],
+			FSType:  fields[2],
+			Options: strings.Split(fields[3], ","),
+			Freq:    fields[4],
+			Pass:    fields[5],
+		})
+	}
+	return mounts, nil
+}
--- a/cmd/endpoint.go
+++ b/cmd/endpoint.go
@@ -21,11 +21,13 @@ import (
 	"net"
 	"net/url"
 	"path"
+	"path/filepath"
 	"sort"
 	"strconv"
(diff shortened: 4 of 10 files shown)


krishnasrinivas (approved):

krishnasrinivas:
Thinking more about the fix, it filters the bucket during list-buckets but does not do anything for other ops like putobject, newmultipart etc.

Better fix is we should show a user friendly message so that they can fix the problem themselves.

So better would be to show a message during bootup like:
`Ignoring bucket "soandso" because it is a mount point.`

harshavardhana:
> Thinking more about the fix, it filters the bucket during list-buckets but does not do anything for other ops like putobject, newmultipart etc.
> 
> Better fix is we should show a user friendly message so that they can fix the problem themselves.
> 
> So better would be to show a message during bootup like:
> Ignoring bucket "soandso" because it is a mount point.

We had this messaging before about all ignored buckets but we couldn't arrive at a constructive sentence which would make sense and is actionable to the user.

krishnasrinivas:
> We had this messaging before about all ignored buckets but we couldn't arrive at a constructive sentence which would make sense and is actionable to the user.

OK, we can discuss further on this issue with AB since just ignoring the buckets during list-buckets does not solve the problem i.e we don't give the user a message on what the problem is.

harshavardhana:
Discussed this with @abperiasamy so we are going to refuse to start the server by detecting cross-device sub-mounts if they match with our export path.  Change the commit and have added code to look for this information. 

Please review @krishnasrinivas @vadmeste 

NOTE: Windows junction files are not supported yet. Haven't found an API to list and validate them just yet.

cmd/server-main.go
```diff
@@ -223,8 +226,57 @@ func serverMain(ctx *cli.Context) {
...
+		// or minio has no permission to read it. In such cases
+		// we just treat it as success and hopefully report
+		// cross-device issues if any at a later stage.
+		if os.IsNotExist(err) || os.IsPermission(err) {
+			return nil, nil
+		}
+		return nil, err
+	}
+
+	var newMounts []mountinfo.MountPoint
```
vadmeste: @harshavardhana, you need to add the following code:
```
        prefix, err = filepath.Abs(prefix)
        if err != nil {
                return nil, err
        }
        prefix += string(filepath.Separator)
```

Using Abs() is more robust when comparing with mounted paths.

Besides, you need to add a separator because otherwise, the user cannot run 'minio server /tmp/fs' if there is another mounted directory which starts by fs, like /tmp/fsmount.

vadmeste:
Isn't it an overkill to treat such case? I am in favor to add a FAQ in minio page which explains different errors that might be difficult to be understood by users.

cmd/server-main.go
harshavardhana: Yes you are right. Let me add it.

harshavardhana:
I think the reasoning is that it is perhaps better for us to detect these things since for the most part people do not read/understand Linux behavior and they deploy in remote instances where access to logging is hard to come by. If we indeed fail early on - mistakes can be avoided.

vadmeste:
> I think the reasoning is that it is perhaps better for us to detect these things since for the most part people do not read/understand Linux behavior and they deploy in remote instances where access to logging is hard to come by. If we indeed fail early on - mistakes can be avoided.

Yeah, my point is that the code will become harder to read because it treats use cases that are not easy to understand in the first place.

harshavardhana:
> Yeah, my point is that the code will become harder to read because it treats use cases that are not easy to understand in the first place.

which part is harder to read? you have to elaborate here so that it can be made easy on the eyes.

pkg/mountinfo/mountinfo.go
```diff
@@ -0,0 +1,31 @@
...
+package mountinfo
+
+// MountPoint - This represents a single line in /proc/mounts.
+type MountPoint struct {
+	Device string
+	Path   string
+	Type   string
+	Opts   []string
+	Freq   int
+	Pass   int
```
balamurugana: You could keep freq and pass as string only to avoid parsing errors.  Let the consumer takes care of it.

pkg/mountinfo/mountinfo.go
```diff
@@ -0,0 +1,31 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package mountinfo
+
+// MountPoint - This represents a single line in /proc/mounts.
+type MountPoint struct {
```
balamurugana: `MountInfo` is more appropriate than `MountInfo`

pkg/mountinfo/mountinfo.go
```diff
@@ -0,0 +1,31 @@
...
+	Device string
+	Path   string
+	Type   string
```
balamurugana: `FSType` is more appropriate

pkg/mountinfo/mountinfo.go
```diff
@@ -0,0 +1,31 @@
...
+	Opts   []string
```
balamurugana: You could expand the name to `Options`

pkg/mountinfo/mountinfo_linux.go
```diff
@@ -0,0 +1,90 @@
...
+
+// MountInfo - returns a list of all mounted filesystems.
+func MountInfo() (mps []MountPoint, err error) {
+	err = readProcMounts(procMountsPath, &mps)
+	return mps, err
+}
+
+// readProcMounts reads the given mountFilePath (normally /proc/mounts) and produces a hash
+// of the contents.  If the out argument is not nil, this fills it with MountPoint structs.
+func readProcMounts(mountFilePath string, out *[]MountPoint) error {
```
balamurugana: `readProcMounts()` returning `[]MountPoint` is much cleaner signature.

pkg/mountinfo/mountinfo_linux.go
```diff
@@ -0,0 +1,90 @@
...
+	file, err := os.Open(mountFilePath)
+	if err != nil {
+		return err
+	}
+	defer file.Close()
+	return readProcMountsFrom(file, out)
+}
+
+func readProcMountsFrom(file io.Reader, out *[]MountPoint) error {
```
balamurugana: You could name this function as `parseProcMounts()` and returning `[]MountPoint` is more appropriate

pkg/mountinfo/mountinfo_linux.go
```diff
@@ -0,0 +1,90 @@
...
+		}
+		fields := strings.Fields(line)
+		if len(fields) != expectedNumFieldsPerLine {
+			return fmt.Errorf("wrong number of fields (expected %d, got %d): %s", expectedNumFieldsPerLine, len(fields), line)
+		}
+
+		if out != nil {
+			mp := MountPoint{
+				Device: fields[0],
+				Path:   fields[1] + "/",
```
balamurugana: Appending `/` is unusual.

pkg/mountinfo/mountinfo_linux.go
```diff
@@ -0,0 +1,90 @@
...
+
+const (
+	// Number of fields per line in /proc/mounts as per the fstab man page.
+	expectedNumFieldsPerLine = 6
+	// Location of the mount file to use
+	procMountsPath = "/proc/mounts"
+)
+
+// MountInfo - returns a list of all mounted filesystems.
+func MountInfo() (mps []MountPoint, err error) {
```
balamurugana: `GetMountInfo()` is more cleaner.

cmd/server-main.go
```diff
@@ -223,8 +226,57 @@ func serverMain(ctx *cli.Context) {
...
+			// if paths didn't match then we do have cross-device mount.
+			err = fmt.Errorf(msg, path, mounts)
+		}
+	}
+	return err
+}
+
 // Initialize object layer with the supplied disks, objectLayer is nil upon any error.
 func newObjectLayer(endpoints EndpointList) (newObject ObjectLayer, err error) {
+	for _, endpoint := range endpoints {
```
balamurugana: You could add this check to single endpoint creation ie `NewEndpoint()` should do this validation.

balamurugana (dismissed):

harshavardhana:
Addressed all your inquiries @balamurugana

pkg/mountinfo/mountinfo_linux.go
```diff
@@ -0,0 +1,135 @@
...
+	// Number of fields per line in /proc/mounts as per the fstab man page.
+	expectedNumFieldsPerLine = 6
+	// Location of the mount file to use
+	procMountsPath = "/proc/mounts"
+)
+
+// List - returns a list of all mounted filesystems, if prefix
+// is provided all matching prefixes for your mounts are listed
+// instead.
+func List(prefix string) ([]MountInfo, error) {
```
krishnasrinivas: doesn't seem to be used outside the package?

krishnasrinivas (commented):
~~mountinfo.CheckCrossDevice does not seem to be there for non linux OSes?~~

pkg/mountinfo/mountinfo.go
```diff
@@ -0,0 +1,31 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package mountinfo
+
+// MountInfo - This represents a single line in /proc/mounts.
```
krishnasrinivas: does not seem to be used outside package?

pkg/mountinfo/mountinfo_linux.go
harshavardhana: This a mountinfo package so it implements basic things which can be useful in future.

pkg/mountinfo/mountinfo.go
harshavardhana: It is exported because List() is exported.

harshavardhana:
> mountinfo.CheckCrossDevice does not seem to be there for non linux OSes?

Yes its okay for now, we can catch up later @krishnasrinivas

pkg/mountinfo/mountinfo_linux.go
krishnasrinivas: better to keep code minimal and add it only when needed

harshavardhana: Sure.. let me change.

pkg/mountinfo/mountinfo_linux.go
```diff
@@ -0,0 +1,128 @@
...
+	mounts, err := readProcMounts(procMountsPath)
+	if err != nil {
+		return nil, err
+	}
+	var filteredMounts []mountInfo
+	for _, mount := range mounts {
+		// Add a separator to indicate that this is a proper directory/device.
+		// This is to avoid a situation where prefix is '/tmp/fsmount'
+		// and mount path is /tmp/fs. In-fact we need to check for `/tmp/fs/`
+		if strings.HasPrefix(strings.TrimSuffix(mount.Path, "/")+"/",
```
vadmeste: path.Abs() is still needed here, otherwise sometimes a mounted dir inside the bucket won't be detected,

```
$ mkdir -b /tmp/fs/bucket
$ sudo mount /dev/your_disk /tmp/fs/bucket
$ cd /tmp/fs/
$ minio server .
```

harshavardhana: Actually let me fail. if the path is absolute.

vadmeste: > Actually let me fail. if the path is absolute.

Probably it is better if we change it to absolute in the beginning of the code path.

harshavardhana: No that is callers responsibility. We don't want to mask usage bugs in that manner.

vadmeste: yeah, I mean change it to absolute very early, probably when endpoints are prepared

harshavardhana: I thought NewEndpoints already does that.

pkg/mountinfo/mountinfo_linux.go
```diff
@@ -0,0 +1,132 @@
...
+func CheckCrossDevice(path string) error {
+	return checkCrossDeviceProcMounts(procMountsPath, path)
+}
+
+// internal function implements sthe same functionality to check if given path has multiple sub-mounts.
+func checkCrossDeviceProcMounts(procMountsPath, path string) error {
+	if !filepath.IsAbs(path) {
+		return fmt.Errorf("Invalid argument, path (%s) is expected to be absolute", path)
+	}
+	mounts, err := listProcMounts(procMountsPath, path)
```
balamurugana: `/proc/mounts` is parsed every time `CheckCrossDevice()` called.  May be you could move this checker into endpoints.go.  This way you can call/get  []MountInfo once and reuse

harshavardhana: Added a global instance.. and using that instead. @balamurugana

harshavardhana:
@vadmeste @krishnasrinivas waiting on further reviews..

vadmeste (approved):
LGTM

harshavardhana:
~~why is this blocked @deekoder ?~~

balamurugana (approved):

krishnasrinivas:
>> better to keep code minimal and add it only when needed

> Sure.. let me change.

List and MountInfo are still exported @harshavardhana

pkg/mountinfo/mountinfo_linux.go
```diff
@@ -0,0 +1,117 @@
...
+	procMountsPath = "/proc/mounts"
+)
+
+// List - lists all mounts.
+func List() (Mounts, error) {
+	return readProcMounts(procMountsPath)
+}
+
+// CheckCrossDevice - check if given path has any sub-mounts in the input mounts list.
+func (mts Mounts) CheckCrossDevice(path string) error {
```
krishnasrinivas: if the signature of this function is:

```
CheckCrossDevice(paths []string) error
```

We can avoid exposing Mount (as this struct is not useful outside the package)

i.e the package can expose only one function

harshavardhana: We did that to avoid doing parsing again and again. @krishnasrinivas

krishnasrinivas (approved):

Issue #4710, opened 2017-07-21
Explore alternatives for browser ux


deekoder:
What could we use instead of react (in case license issue with react does not resolve).

brendanashworth:
There's [no shortage of options](http://stateofjs.com/2016/frontend/) for [JS frontend frameworks](https://github.com/showcases/front-end-javascript-frameworks): [ember.js](https://emberjs.com/), [backbone.js](http://backbonejs.org/#), [vue.js](http://vuejs.org/), [Polymer](https://www.polymer-project.org/) and [angular.js](https://angularjs.org/) to name some of the most popular besides react. Selecting a framework can be very dependent on the team and intended use case, as well as opinionated. It's a deep hole that one can spend a lot of time in.

However our situation is that, should the license situation be an issue, we'd need to port the current browser over to a new framework. Similarities between react.js and the new framework would be what we want the most, and one option thats promising would be Google's [Polymer](https://www.polymer-project.org/) ([on github](https://github.com/Polymer/polymer)). Polymer is licensed under the BSD license.

I'd say the browser's three core parts are Redux, HTML extending components, and RPC.

- HTML extending components is also [Polymer's mojo](https://www.polymer-project.org/2.0/start/quick-tour). We should be able to keep all of our [components](https://github.com/minio/minio/tree/4785555d34a6a481ed9247e25e560585216b73fe/browser/app/js/components) without a major refactor, just slight ports in the actual code. This is a major plus, and I'm not sure if another library does it as similar as react does it.
- Redux is how the app manages [state and dispatching actions](https://github.com/minio/minio/blob/4785555d34a6a481ed9247e25e560585216b73fe/browser/app/js/actions.js). Polymer has a [Redux library](https://github.com/tur-nr/polymer-redux) that should allow us to keep some code the same, without reworking the entire state integration with the rest of the app.
- We have our own RPC implementation.

Of course, the best case scenario would be we keep React as is and don't have to change anything. The [PATENTS](https://www.elcaminolegal.com/single-post/2016/10/04/Facebook-Reactjs-License) situation needs to be discussed, and if it's deemed to be an issue we can look further into a new alternative.

brendanashworth:
After the discussion, we've decided to move to Vuejs. This can be closed.

harshavardhana:
We can move a new issue for migration instead @brendanashworth ? and then we can close this.

GarimaKapoor:
@deekoder since we have already discussed on this issue. We should close it.

[...]

PR #4810, opened 2017-08-12
browser: move from react to vuejs

brendanashworth:
## Description
This commit moves most of the web browser from React to the VueJS
framework. This is about 90% complete. 

This commit contains the entire rewrite. The app maintains much of its
old architecture, just this time with a new framework and a more modular
approach to the Browse component. The app state is much slimmer now, as
components should maintain as much state as they can when it isn't
relevant to other components. If it is, only then should we keep it in
our global app state.

Because this uses Vue templates instead of JSX, we can take advantage of
their superior event handling, and this is visible throughout the code
base. Additionally the separation between <template> and <script> make
the browser code much more readable and understandable.

The JSON RPC and Web action handling stays the same, as does some
actions. The UI is almost entirely unchanged; opening the new and old
browser in two tabs shows negligible difference, if you notice it.

Because this rewrite moves much of the app state from the global state,
this allows the individual components themselves to control the viewing
of the global state. For example, ObjectsList now performs object
sorting on the fly when objects are updated, as opposed to storing the
objects themselves as a sorted array. This is much better structured.
Another example is the search bucket functionality. Before, we'd have a
list of "visible buckets" and regular buckets. Now we store a search
query in the SideBar component, and buckets are dynamically searched
through when they're viewed. This should make the browser easier to
maintain in the future.

Another perk of this rewrite is that it contains the newux improvements.
This commit contains the delete bucket from the UI changes, as well as
object preview (= image thumbnails).

There are however a few things that aren't implemented yet. These
include the most important:
- object uploading
- bulk object downloads, by checked objects
- login route needs fixing, shouldn't be its own route
- dropzone and infinite scroll are not implemented
- share modal is not implemented
- unauthenticated views aren't completely ready

This is not a comprehensive list. There is a "TODO" in the code for
everything that I believe needs fixing / implementation. Besides these
TODOs, the app needs comprehensive testing before publication, but I
believe the current app state to be very close to ready.

~ with <3

## Motivation and Context
This was done because of licensing issues with React; see:
https://github.com/minio/minio/issues/4710


harshavardhana:
:heart: :heart:

PR #4815, opened 2017-08-14
Minor corrections in comments for xl utils

fwessels:
## Description
Just some minor corrections in comments for xl utils (found as part of bit rot review)

## Types of changes

## Checklist:

### DIFF
--- a/cmd/xl-v1-utils.go
+++ b/cmd/xl-v1-utils.go
@@ -27,7 +27,7 @@ import (
 )
 
 // Returns number of errors that occurred the most (incl. nil) and the
-// corresponding error value. N B when there is more than one error value that
+// corresponding error value. NB When there is more than one error value that
 // occurs maximum number of times, the error value returned depends on how
 // golang's map orders keys. This doesn't affect correctness as long as quorum
 // value is greater than or equal to simple majority, since none of the equally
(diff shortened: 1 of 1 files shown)


cmd/xl-v1-utils.go
```diff
@@ -27,7 +27,7 @@ import (
 )
 
 // Returns number of errors that occurred the most (incl. nil) and the
-// corresponding error value. N B when there is more than one error value that
+// corresponding error value. NB When there is more than one error value that
```
harshavardhana, line 30: Isn't this usually `N.B` not `NB` ?  :1st_place_medal:

vadmeste (approved):
LGTM

cmd/xl-v1-utils.go
fwessels: Then it should be N.B. -- it is a Latin abbreviation for 'Nota Bene' meaning "note well", or something along these lines of "watch out" / "be careful" ...

Of course we could also use PS which means "Post scriptum" which refers to an "afterthrough", or even PSS which is "Post scriptum scriptum" which is an "afterthrough" on an "afterthrough" but we are starting to digress now...    😉

harshavardhana (approved):

PR #4817, opened 2017-08-15
Fix for DeleteMultipleObjects access denied response

wd256:
## Description
This pull request addresses the case where the requesting user does not have access to the bucket they are attempting to delete objects from and should receive an 'Access Denied' error for each object they are attempting to delete rather than the generic 403 AccessDenied error response.

## Motivation and Context
While using the minio-go client to issue an anonymous multiple object delete request to Minio, I noticed a difference between how Amazon and Minio respond.

In the case of Minio, if an anonymous user makes a delete multiple object request and is denied access, they will receive a response with a 403 status code and the generic error response body with an "AccessDenied" code.

In the case of Amazon, if an anonymous user makes a delete multiple object request and is denied access, a response with a 200 status code is returned which includes a DeleteResponse object and a list of errors for each item which the user attempted to delete and was denied access.

See the sample response for Example 1:
http://docs.aws.amazon.com/AmazonS3/latest/API/multiobjectdeleteapi.html#multiobjectdeleteapi-examples

## How Has This Been Tested?
- Manually compared responses from Minio and Amazon
- Added a test case and ran all existing tests successfully
- Tested on Ubuntu 17.04 with go 1.8.3

## Types of changes

## Checklist:

### DIFF
--- a/cmd/bucket-handlers.go
+++ b/cmd/bucket-handlers.go
@@ -240,9 +240,14 @@ func (api objectAPIHandlers) DeleteMultipleObjectsHandler(w http.ResponseWriter,
 		return
 	}
 
-	if s3Error := checkRequestAuthType(r, bucket, "s3:DeleteObject", serverConfig.GetRegion()); s3Error != ErrNone {
-		writeErrorResponse(w, s3Error, r.URL)
-		return
+	var authError APIErrorCode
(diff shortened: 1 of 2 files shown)


balamurugana (approved):

harshavardhana (approved):
// cc @NitishT this is a good case for mint as well.

nitisht:
:+1: @harshavardhana https://github.com/minio/mint/issues/115

PR #4823, opened 2017-08-16
fix: [minor] functions should take inputs with required functionality.

harshavardhana:
## Description
Functions should take inputs with required functionality.

## Motivation and Context
Code-review and cleanup.

## How Has This Been Tested?
Manually.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/generic-handlers.go
+++ b/cmd/generic-handlers.go
@@ -24,20 +24,17 @@ import (
 	"time"
 
 	humanize "github.com/dustin/go-humanize"
-	router "github.com/gorilla/mux"
 	"github.com/rs/cors"
 )
 
 // HandlerFunc - useful to chain different middleware http.Handler
 type HandlerFunc func(http.Handler) http.Handler
 
-func registerHandlers(mux *router.Router, handlerFns ...HandlerFunc) http.Handler {
-	var f http.Handler
-	f = mux
+func registerHandlers(h http.Handler, handlerFns ...HandlerFunc) http.Handler {
 	for _, hFn := range handlerFns {
-		f = hFn(f)
+		h = hFn(h)
 	}
-	return f
+	return h
 }
 
 // Adds limiting body size middleware


cmd/erasure-createfile.go
```diff
@@ -80,7 +80,7 @@ func (s *ErasureStorage) CreateFile(src io.Reader, volume, path string, buffer [
 
 // erasureAppendFile appends the content of buf to the file on the given disk and updates computes
 // the hash of the written data. It sends the write error (or nil) over the error channel.
-func erasureAppendFile(disk StorageAPI, volume, path string, hash hash.Hash, buf []byte, errChan chan<- error) {
+func erasureAppendFile(disk StorageAPI, volume, path string, hashWriter io.Writer, buf []byte, errChan chan<- error) {
```
aead: In this case I'm not sure because the semantic of erasureAppendFile requires a hash function (actually not necessarily but minio only supports hash functions for bitrot protection). The variable name / doc gives some hint but a type would be more explicit...
The question here is: explicitness vs. abstraction...

cmd/generic-handlers.go
```diff
@@ -24,20 +24,17 @@ import (
...
 	"github.com/rs/cors"
 )
 
 // HandlerFunc - useful to chain different middleware http.Handler
 type HandlerFunc func(http.Handler) http.Handler
 
-func registerHandlers(mux *router.Router, handlerFns ...HandlerFunc) http.Handler {
-	var f http.Handler
-	f = mux
+func registerHandlers(h http.Handler, handlerFns ...HandlerFunc) http.Handler {
```
aead, line 33: This is fine because the semantics of registerHandlers does not require a router.Router. A Handler is sufficient.

cmd/erasure-createfile.go
harshavardhana: Reverted..

balamurugana (approved):

aead (approved):
LGTM

PR #4826, opened 2017-08-16
New Admin API implementation

donatello:
* Move Admin APIs to new endpoint starting with `/minio/admin/v1`.
* Add admin version endpoint at `/minio/admin/version`
* Change heal related APIs to work like a server-side background service. Now it has heal-start, heal-stop and heal-status-fetch operations.

## Description

This change is a rewrite of the Minio server administration APIs.

Admin APIs are now versioned, and this change brings in version 1.

Technical details about the REST API implemented in this PR are documented temporarily at https://gist.github.com/donatello/e520308bf83adddc70366ea01ef505e5

Non-heal admin APIs are moved to new endpoints and have a more REST API style using JSON to encode parameters.

Heal APIs are reimplemented like a server-side background service. Three operations are possible - start a heal sequence (with various options at a path), stop a heal sequence (stops when the server reaches the earliest safe point), and to fetch status/progress information on a heal sequence.
 
Each heal sequence starts by healing disk formats, then heals the bucket specified or all buckets if none are specified, and then heals objects at and under a given (possible empty) prefix. Two heal operations on overlapping paths (i.e. specified bucket and object prefix combination) may not be started (and the implementation will return an error in such cases).

A heal sequence may be stopped by invoking the heal-stop API. In this case, the server aborts the heal operation only at the earliest safe point reached (i.e. between healing buckets or objects), since it may be in the process of mutating on-disk data. 

Heal results are sent to the client when the client invokes the fetch-heal-status API. Heal results include details of each operation performed by the server, in the form of object level heal-result records. In addition statistics of the aggregate of heal operations performed in this heal sequence are also sent back in each fetch-status request.

The client is required to acknowledge results it receives from the server, so that the server can remove consumed records. The server accumulates records until the client consumes them, upto a limit of 1000 records at which point, the server pauses the heal sequence waiting for the client to catch up. After a timeout with no records consumed by the client, the heal sequence aborts.

When the `statisticsOnly` option is specified, no details of heal-operations are sent back to the client. Only statistics are sent. In this case, the client need not acknowledge consumption of results.

~TODOs:~
* ~Add support for healing parts in an in-progress multipart upload.~
* ~Update `madmin` package documentation in codebase.~

EDIT: Handling of unrecoverable objects requires some discussion and is deferred to after getting this change in.

## Motivation and Context

Brings REST-style Admin APIs. Heal APIs are now easier to use, consisting of just three high-level operations - start a heal sequence, stop a heal sequence and fetch status data about a heal operation.

Context of earlier heal proposal - https://gist.github.com/krishnasrinivas/4b216a6ec03925645253788d8e0e241d

## How Has This Been Tested?

Manually tested. 
~TODO: more unit tests to be added.~
New test cases are added.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/admin-handlers.go
+++ b/cmd/admin-handlers.go
@@ -23,11 +23,11 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/url"
-	"path"
 	"strconv"
 	"sync"
 	"time"
 
+	"github.com/gorilla/mux"
 	"github.com/minio/minio/pkg/auth"
 	"github.com/minio/minio/pkg/madmin"
 )
@@ -41,18 +41,10 @@ type mgmtQueryKey string
 
 // Only valid query params for mgmt admin APIs.
 const (
-	mgmtBucket         mgmtQueryKey = "bucket"
-	mgmtObject         mgmtQueryKey = "object"
-	mgmtPrefix         mgmtQueryKey = "prefix"
-	mgmtLockOlderThan  mgmtQueryKey = "older-than"
-	mgmtDelimiter      mgmtQueryKey = "delimiter"
-	mgmtMarker         mgmtQueryKey = "marker"
-	mgmtKeyMarker      mgmtQueryKey = "key-marker"
-	mgmtMaxKey         mgmtQueryKey = "max-key"
-	mgmtDryRun         mgmtQueryKey = "dry-run"
-	mgmtUploadIDMarker mgmtQueryKey = "upload-id-marker"
-	mgmtMaxUploads     mgmtQueryKey = "max-uploads"
-	mgmtUploadID       mgmtQueryKey = "upload-id"
+	mgmtBucket        mgmtQueryKey = "bucket"
+	mgmtPrefix        mgmtQueryKey = "prefix"
+	mgmtLockOlderThan mgmtQueryKey = "older-than"
+	mgmtMarker        mgmtQueryKey = "marker"
 )
 
 var (
@@ -61,8 +53,10 @@ var (
 	adminAPIVersionInfo = madmin.AdminAPIVersionInfo{"1"}
 )
 
-func (adminAPI adminAPIHandlers) VersionHandler(w http.ResponseWriter,
-	r *http.Request) {
+// VersionHandler - GET /minio/admin/version
+// -----------
+// Returns Administration API version
+func (a adminAPIHandlers) VersionHandler(w http.ResponseWriter, r *http.Request) {
 
 	adminAPIErr := checkRequestAuthType(r, "", "", "")
 	if adminAPIErr != ErrNone {
@@ -72,7 +66,7 @@ func (adminAPI adminAPIHandlers) VersionHandler(w http.ResponseWriter,
 
 	jsonBytes, err := json.Marshal(adminAPIVersionInfo)
 	if err != nil {
-		writeErrorResponse(w, ErrInternalError, r.URL)
+		writeErrorResponseJSON(w, ErrInternalError, r.URL)
 		errorIf(err, "Failed to marshal Admin API Version to JSON.")
 		return
 	}
@@ -83,10 +77,10 @@ func (adminAPI adminAPIHandlers) VersionHandler(w http.ResponseWriter,
 // ServiceStatusHandler - GET /minio/admin/v1/service
 // ----------
 // Returns server version and uptime.
-func (adminAPI adminAPIHandlers) ServiceStatusHandler(w http.ResponseWriter, r *http.Request) {
+func (a adminAPIHandlers) ServiceStatusHandler(w http.ResponseWriter, r *http.Request) {
 	adminAPIErr := checkRequestAuthType(r, "", "", "")
 	if adminAPIErr != ErrNone {
-		writeErrorResponse(w, adminAPIErr, r.URL)
+		writeErrorResponseJSON(w, adminAPIErr, r.URL)
 		return
 	}
 
@@ -100,7 +94,7 @@ func (adminAPI adminAPIHandlers) ServiceStatusHandler(w http.ResponseWriter, r *
 	// of read-quorum availability.
 	uptime, err := getPeerUptimes(globalAdminPeers)
 	if err != nil {
-		writeErrorResponse(w, toAPIErrorCode(err), r.URL)
+		writeErrorResponseJSON(w, toAPIErrorCode(err), r.URL)
 		errorIf(err, "Possibly failed to get uptime from majority of servers.")
 		return
 	}
@@ -114,7 +108,7 @@ func (adminAPI adminAPIHandlers) ServiceStatusHandler(w http.ResponseWriter, r *
 	// Marshal API response
 	jsonBytes, err := json.Marshal(serverStatus)
 	if err != nil {
-		writeErrorResponse(w, ErrInternalError, r.URL)
+		writeErrorResponseJSON(w, ErrInternalError, r.URL)
 		errorIf(err, "Failed to marshal storage info into json.")
 		return
 	}
@@ -128,24 +122,24 @@ func (adminAPI adminAPIHandlers) ServiceStatusHandler(w http.ResponseWriter, r *
 // ----------
 // Restarts/Stops minio server gracefully. In a distributed setup,
 // restarts all the servers in the cluster.
-func (adminAPI adminAPIHandlers) ServiceStopNRestartHandler(w http.ResponseWriter, r *http.Request) {
+func (a adminAPIHandlers) ServiceStopNRestartHandler(w http.ResponseWriter, r *http.Request) {
 	adminAPIErr := checkRequestAuthType(r, "", "", "")
 	if adminAPIErr != ErrNone {
-		writeErrorResponse(w, adminAPIErr, r.URL)
+		writeErrorResponseJSON(w, adminAPIErr, r.URL)
 		return
 	}
 
 	body, err := ioutil.ReadAll(r.Body)
 	if err != nil {
 		errorIf(err, "Failed to read request body")
-		writeErrorResponse(w, toAPIErrorCode(err), r.URL)
+		writeErrorResponseJSON(w, toAPIErrorCode(err), r.URL)
 		return
 	}
 
 	var sa madmin.ServiceAction
 	err = json.Unmarshal(body, &sa)
 	if err != nil {
-		writeErrorResponse(w, ErrMalformedPOSTRequest, r.URL)
+		writeErrorResponseJSON(w, ErrMalformedPOSTRequest, r.URL)
 		errorIf(err, "Error parsing body JSON")
 		return
 	}
@@ -157,7 +151,7 @@ func (adminAPI adminAPIHandlers) ServiceStopNRestartHandler(w http.ResponseWrite
 	case madmin.ServiceActionValueStop:
 		serviceSig = serviceStop
 	default:
-		writeErrorResponse(w, ErrMalformedPOSTRequest, r.URL)
+		writeErrorResponseJSON(w, ErrMalformedPOSTRequest, r.URL)
 		errorIf(err, "Invalid service action received")
 		return
 	}
@@ -226,11 +220,11 @@ type ServerInfo struct {
 // ServerInfoHandler - GET /minio/admin/v1/info
 // ----------
 // Get server information
-func (adminAPI adminAPIHandlers) ServerInfoHandler(w http.ResponseWriter, r *http.Request) {
+func (a adminAPIHandlers) ServerInfoHandler(w http.ResponseWriter, r *http.Request) {
 	// Authenticate request
 	adminAPIErr := checkRequestAuthType(r, "", "", "")
 	if adminAPIErr != ErrNone {
-		writeErrorResponse(w, adminAPIErr, r.URL)
+		writeErrorResponseJSON(w, adminAPIErr, r.URL)
 		return
 	}
 
@@ -266,7 +260,7 @@ func (adminAPI adminAPIHandlers) ServerInfoHandler(w http.ResponseWriter, r *htt
 	// Marshal API response
 	jsonBytes, err := json.Marshal(reply)
 	if err != nil {
-		writeErrorResponse(w, ErrInternalError, r.URL)
+		writeErrorResponseJSON(w, ErrInternalError, r.URL)
 		errorIf(err, "Failed to marshal storage info into json.")
 		return
 	}
@@ -313,19 +307,18 @@ func validateLockQueryParams(vars url.Values) (string, string, time.Duration,
 // - prefix and older-than are optional query parameters
 // ---------
 // Lists locks held on a given bucket, prefix and duration it was held for.
-func (adminAPI adminAPIHandlers) ListLocksHandler(w http.ResponseWriter,
-	r *http.Request) {
+func (a adminAPIHandlers) ListLocksHandler(w http.ResponseWriter, r *http.Request) {
 
 	adminAPIErr := checkRequestAuthType(r, "", "", "")
 	if adminAPIErr != ErrNone {
-		writeErrorResponse(w, adminAPIErr, r.URL)
+		writeErrorResponseJSON(w, adminAPIErr, r.URL)
 		return
 	}
 
 	vars := r.URL.Query()
 	bucket, prefix, duration, adminAPIErr := validateLockQueryParams(vars)
 	if adminAPIErr != ErrNone {
-		writeErrorResponse(w, adminAPIErr, r.URL)
+		writeErrorResponseJSON(w, adminAPIErr, r.URL)
 		return
 	}
 
@@ -334,7 +327,7 @@ func (adminAPI adminAPIHandlers) ListLocksHandler(w http.ResponseWriter,
 	volLocks, err := listPeerLocksInfo(globalAdminPeers, bucket, prefix,
 		duration)
 	if err != nil {
-		writeErrorResponse(w, ErrInternalError, r.URL)
+		writeErrorResponseJSON(w, ErrInternalError, r.URL)
 		errorIf(err, "Failed to fetch lock information from remote nodes.")
 		return
 	}
@@ -342,7 +335,7 @@ func (adminAPI adminAPIHandlers) ListLocksHandler(w http.ResponseWriter,
 	// Marshal list of locks as json.
 	jsonBytes, err := json.Marshal(volLocks)
 	if err != nil {
-		writeErrorResponse(w, ErrInternalError, r.URL)
+		writeErrorResponseJSON(w, ErrInternalError, r.URL)
 		errorIf(err, "Failed to marshal lock information into json.")
 		return
 	}
@@ -358,19 +351,18 @@ func (adminAPI adminAPIHandlers) ListLocksHandler(w http.ResponseWriter,
 // HTTP header x-minio-operation: clear
 // ---------
 // Clear locks held on a given bucket, prefix and duration it was held for.
-func (adminAPI adminAPIHandlers) ClearLocksHandler(w http.ResponseWriter,
-	r *http.Request) {
+func (a adminAPIHandlers) ClearLocksHandler(w http.ResponseWriter, r *http.Request) {
 
 	adminAPIErr := checkRequestAuthType(r, "", "", "")
 	if adminAPIErr != ErrNone {
-		writeErrorResponse(w, adminAPIErr, r.URL)
+		writeErrorResponseJSON(w, adminAPIErr, r.URL)
 		return
 	}
 
 	vars := r.URL.Query()
 	bucket, prefix, duration, adminAPIErr := validateLockQueryParams(vars)
 	if adminAPIErr != ErrNone {
-		writeErrorResponse(w, adminAPIErr, r.URL)
+		writeErrorResponseJSON(w, adminAPIErr, r.URL)
 		return
 	}
 
@@ -379,7 +371,7 @@ func (adminAPI adminAPIHandlers) ClearLocksHandler(w http.ResponseWriter,
 	volLocks, err := listPeerLocksInfo(globalAdminPeers, bucket, prefix,
 		duration)
 	if err != nil {
-		writeErrorResponse(w, ErrInternalError, r.URL)
+		writeErrorResponseJSON(w, ErrInternalError, r.URL)
 		errorIf(err, "Failed to fetch lock information from remote nodes.")
 		return
 	}
@@ -387,7 +379,7 @@ func (adminAPI adminAPIHandlers) ClearLocksHandler(w http.ResponseWriter,
 	// Marshal list of locks as json.
 	jsonBytes, err := json.Marshal(volLocks)
 	if err != nil {
-		writeErrorResponse(w, ErrInternalError, r.URL)
+		writeErrorResponseJSON(w, ErrInternalError, r.URL)
 		errorIf(err, "Failed to marshal lock information into json.")
 		return
 	}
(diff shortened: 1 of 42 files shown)


cmd/admin-handlers.go
```diff
@@ -54,23 +54,34 @@ const (
...
+	// This struct literal represents the Admin API version that
+	// the server uses.
+	adminAPIVersionInfo = madmin.AdminAPIVersionInfo{"1"}
+)
 
-// ServerStatus - contains the response of service status API
-type ServerStatus struct {
-	ServerVersion ServerVersion `json:"serverVersion"`
-	Uptime        time.Duration `json:"uptime"`
+func (adminAPI adminAPIHandlers) VersionHandler(w http.ResponseWriter,
```
krisis: Include a comment for this handler with the REST API syntax like GET /minio/admin/version

cmd/admin-handlers.go
```diff
@@ -422,455 +400,220 @@ func (adminAPI adminAPIHandlers) ClearLocksHandler(w http.ResponseWriter, r *htt
...
-		return
-	}
+	// find number of disk in the setup
+	info := objLayer.StorageInfo()
+	numDisks := info.Backend.OfflineDisks + info.Backend.OnlineDisks
 
-	// if dry-run is present in query-params, then only perform validations and return success.
-	if isDryRun(vars) {
-		writeSuccessResponseHeadersOnly(w)
+	fmt.Printf("Starting heal sequence on %s/%s with settings %v\n", bucket,
```
krisis: Shouldn't this use `log.Printf` instead of `fmt.Printf`?

cmd/admin-handlers.go
```diff
@@ -422,455 +400,220 @@ func (adminAPI adminAPIHandlers) ClearLocksHandler(w http.ResponseWriter, r *htt
...
+	nh := newHealSequence(bucket, objPrefix, numDisks, hs)
+	if err := globalAllHealState.AddNewHealSequence(nh); err != nil {
+		writeErrorResponseJSON(w, ErrInternalError, r.URL)
+		errorIf(err, "Failed to start heal sequence.")
 		return
 	}
 
-	// Heal the given bucket.
-	err := objLayer.HealBucket(bucket)
+	// Marshal list of locks as json.
```
krisis: Marshal list of locks? Shouldn't this be list of objects to be healed or similar?

cmd/admin-heal-ops.go
```diff
@@ -0,0 +1,659 @@
...
+	// for this timeout duration, the heal sequence is aborted.
+	healUnconsumedTimeout = time.Hour * 24
+
+	// time-duration to keep heal sequence state after it
+	// completes.
+	keepHealSeqStateDuration = time.Hour
+)
+
+var (
+	errHealIdleTimeout          = fmt.Errorf("healing results were unconsumed for too long")
```
krisis: s/unconsumed/not consumed ?

cmd/admin-heal-ops.go
```diff
@@ -0,0 +1,659 @@
...
+	errHealServerNotInitialized = fmt.Errorf("server is not initialized - please try again")
```
krisis: we already have an error value for server not initialized yet - `errServerNotInitialized`. Why do we need another one?

cmd/admin-heal-ops.go
```diff
@@ -0,0 +1,659 @@
...
+		errorIf(err, "Error updating heal statistics")
+	}
+}
+
+type healSequenceStatus struct {
+	Summary       healStatusSummary `json:"Summary"`
+	FailureDetail string            `json:"Detail,omitempty"`
+	HealSettings  healSettings      `json:"Settings"`
+
+	updateLock *sync.Mutex
```
krisis: is there a reason to have updateLock be of type `*sync.Mutex`?

cmd/admin-heal-ops.go
```diff
@@ -0,0 +1,659 @@
...
+
+	healSeqMap map[string]*healSequence
+}
+
+var (
+	globalAllHealState allHealState
+)
+
+// initAllHealState - initialize healing apparatus
+func initAllHealState(isErasureMore bool) {
```
krisis: `isErasureMore` - should it be `isErasureMode`?

krisis (dismissed):
First round of comments. Will look further into heal API related go-routines.

cmd/admin-heal-ops.go
harshavardhana: https://dave.cheney.net/2013/01/19/what-is-the-zero-value-and-why-is-it-useful

donatello: `updateLock` is a pointer to avoid copying of the lock when the enclosing structure is passed around/returned (by value).

cmd/admin-handlers.go
donatello: Removed the line.

cmd/admin-heal-ops.go
donatello: I thought it may be useful to distinguish it but then I decided against it and removed it.

donatello:
When trying out this change, please use `mc` from this PR https://github.com/minio/mc/pull/2239 - it now prints out heal status information in table format.

donatello:
Pinging for reviews @harshavardhana @krishnasrinivas @krisis

cmd/generic-handlers.go
```diff
@@ -219,8 +225,9 @@ func setReservedBucketHandler(h http.Handler) http.Handler {
 }
 
 func (h minioReservedBucketHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	if !guessIsRPCReq(r) && !guessIsBrowserReq(r) {
-		// For all non browser, non RPC requests, reject access to 'minioReservedBucketPath'.
+	if !guessIsRPCReq(r) && !guessIsBrowserReq(r) && !isAdminAPIRequest(r) {
+		// For all non browser, non RPC requests, reject
+		// access to 'minioReservedBucketPath'.
```
krisis: This can be simplified as,
```go
switch {
  case guessIsRPCReq(r), guessIsBrowserReq(r), isAdminAPIRequest(r):
  // Do nothing
  default:
  // Code to disallow access to /minio goes here
      return
}
h.handler.ServeHTTP(w, r)

```

cmd/generic-handlers.go
```diff
@@ -209,6 +209,12 @@ func (h cacheControlHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	h.handler.ServeHTTP(w, r)
 }
 
+// Check to allow access to the reserved "bucket" `/minio` for Admin
+// API requests.
+func isAdminAPIRequest(r *http.Request) bool {
```
krisis: Suggestion: rename this function as `isAdminReq` to rhyme with `guessIsBrowserReq` and `guessIsRPCReq`.

pkg/madmin/heal-commands.go
```diff
@@ -435,295 +108,123 @@ func (adm *AdminClient) HealBucket(bucket string, dryrun bool) error {
...
 	defer closeResponse(resp)
 	if err != nil {
-		return listUploadsHealResponse{}, err
+		return res, err
 	}
 
 	if resp.StatusCode != http.StatusOK {
-		return toBeHealedUploads, httpRespToErrorResponse(resp)
-
+		return res, httpRespToErrorResponse(resp)
```
krisis, line 280: `httpRespToErrorResponse` assumes that error response is in XML.

krisis:
While testing I found the following issues,
-  `mc admin heal status s1/test` doesn't report the disk where the object is missing
  The object directory was removed from the backend in disk 2 in a 4 disk setup.

- `mc admin heal stop s1/test` on a previously stopped heal failed with the following (partial) backtrace
```
2017/09/06 12:25:41 http: panic serving 127.0.0.1:55800: close of closed channel ---> bug?
goroutine 328 [running]:
net/http.(*conn).serve.func1(0xc420548000)
	/home/kp/.gimme/versions/go/src/net/http/server.go:1721 +0xd0
panic(0xef9a80, 0xc42046ee60)
	/home/kp/.gimme/versions/go/src/runtime/panic.go:489 +0x2cf
github.com/minio/minio/cmd.(*allHealState).StopHealSequence(0x1891b30, 0xc420313438, 0x5, 0x0)
	/home/kp/code/go1.7.1/workspace/src/github.com/minio/minio/cmd/admin-heal-ops.go:390 +0xb9
github.com/minio/minio/cmd.adminAPIHandlers.HealStopHandler(0x15ead40, 0xc4204690a0, 0xc42030e300)
	/home/kp/code/go1.7.1/workspace/src/github.com/minio/minio/cmd/admin-handlers.go:603 +0x18d
github.com/minio/minio/cmd.(adminAPIHandlers).HealStopHandler-fm(0x15ead40, 0xc4204690a0, 0xc42030e300)
	/home/kp/code/go1.7.1/workspace/src/github.com/minio/minio/cmd/admin-router.go:74 +0x45
```

krisis (changes requested):
Sharing some initial remarks. I found these while testing locally.

pkg/madmin/heal-commands.go
donatello: Fixed - switched over to parsing json as errors are also expected to be json

cmd/generic-handlers.go
donatello: done

cmd/generic-handlers.go
donatello: done

donatello:
@krisis addressed your comments.

> *  mc admin heal status s1/test doesn't report the disk where the object is missing
    The object directory was removed from the backend in disk 2 in a 4 disk setup.

I tried many scenarios but I am not able to reproduce. The healed object does appear in the first table with disk parts looking like `#H##` meaning on disk 2 the object was healed. I am testing with a 4-node setup under docker. If you are able to reproduce it, please give me more details so I can track it down. 

>  * mc admin heal stop s1/test on a previously stopped heal failed with the following (partial) backtrace

This is fixed now - a second stop command on the same heal sequence was causing two close calls on a channel. Now it is fixed to check if the heal sequence has already ended before calling close().

donatello:
Current AppVeyor build failure appears unrelated - nothing was changed related to posix file reading:

```
=== RUN   TestPosixReadFile
--- FAIL: TestPosixReadFile (0.01s)
	posix_test.go:1096: Case: 11 struct { volume string; fileName string; offset int64; bufSize int; expectedBuf []uint8; expectedErr error }{volume:"success-vol", fileName:"myobject", offset:-1, bufSize:5, expectedBuf:[]uint8(nil), expectedErr:(*os.PathError)(0xc04e116000)}, expected: read C:\Users\appveyor\AppData\Local\Temp\1\minio-538410652/success-vol/myobject: The parameter is incorrect., got: readat C:\Users\appveyor\AppData\Local\Temp\1\minio-538410652/success-vol/myobject: negative offset
```

@harshavardhana @krishnasrinivas Nudge nudge. Could I get some reviews please?

donatello:
After my most recent push, there were changes in madmin pkg too - so mc will need to be updated and rebuilt from my PR at https://github.com/minio/mc/pull/2239 to be able to test this PR conveniently.

harshavardhana:
The appveyor failure is because the default Go version has moved to 1.9 you should fix `appveyor.yml` . To point to go18 instead. @donatello

harshavardhana:
Sorry for being late on testing this.. will test this Monday.

donatello:
Bug https://github.com/minio/minio/issues/4905 needs to be fixed as it is a lower-level issue for the admin API.

donatello:
> The appveyor failure is because the default Go version has moved to 1.9 you should fix appveyor.yml. To point to go18 instead. @donatello

@harshavardhana tried your suggestion, but there are other weird problems with appveyor - not sure how to fix atm.

harshavardhana:
> @harshavardhana tried your suggestion, but there are other weird problems with appveyor - not sure how to fix atm.

Actually it is being fixed in https://github.com/minio/minio/pull/4904

krishnasrinivas:
@donatello https://gist.github.com/donatello/e520308bf83adddc70366ea01ef505e5 added some comments

donatello:
Reply to @krishnasrinivas comments on the spec gist (I am replying here as comments on the gist do not seem to generate notifications) - all the quoted parts below are from his [comment](https://gist.github.com/donatello/e520308bf83adddc70366ea01ef505e5#gistcomment-2200915):

> Fetch status of a heal seqeuence
> 
> Request:
> There are 3 variations similar to the two previous Heal APIs. The path provided, identifies the specific heal sequence for fetching status.
> 
> POST /minio/admin/v1/heal-status/
> POST /minio/admin/v1/heal-status/<bucket-name>
> POST /minio/admin/v1/heal-status/<bucket-name>/<object-prefix>
> 
> For this you can use GET HTTP Verb like: (i.e there is no need for heal-status in /minio/admin/v1/heal-status/..., you can use /minio/admin/v1/heal/...
> 
> GET /minio/admin/v1/heal/
> GET /minio/admin/v1/heal/<bucket-name>
> GET /minio/admin/v1/heal/<bucket-name>/<object-prefix>
>
> is ackIndex related to resultID of the response?

The heal-status fetching API also sends acknowledgements to the server via the `ackIndex` parameter. This represents the index of the highest numbered heal-record (i.e. the record index; it is also an ID but index is more accurate as these "IDs" are all serial numbers) received by the client. The server uses it to remove older heal records from its memory. This is thus a state changing operation, and so it uses POST. 

While technically, a GET could also be used, it would not be idempotent or safe (the precise meaning of these terms is given in section 9.1.1 and 9.1.2 in the [HTTP RFC](https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html)):
- non-idempotent because two requests with the same ackIndex can return different results - by the time the second request is received, the server may have more heal records to return.
- non-safe because making a request with an older ackIndex than one that is already sent to the server in a previous request would not work, as the server would have removed older records.

Based on these properties, POST is the correct HTTP verb to use here.

> AB also wanted an advanced mode which does bitrot check as well.

The reason it is removed is that the process of healing data at the core is simply decoding erasure coded blocks of data, reconstructing them and writing the blocks back again. Bitrot checking (i.e. checking if erasure coded blocks are modified due to disk errors, etc) is an essential part of checking that erasure coded blocks are sane - even the implementation of decoding erasure coded blocks (`erasure.HealFile()`) performs this checking. Without bitrot checking, simply reconstructing data can lead to garbage output as the input data blocks are not sanitized - this means that without bitrot checking we cannot ensure the integrity of data after healing.

The result is that healing just does bitrot verification by default, and so there is no advanced mode. This was also discussed with AB more recently.

> Update server configuration and Update server credentials have same endpoint

~I will check and fix this.~
EDIT: This was a bug in the spec - I have fixed the gist now.

donatello:
This PR depends on https://github.com/minio/minio/pull/4913 being merged into master to work correctly for multipart objects to be healed correctly.

harshavardhana:
@donatello i have some feedback for the related PR on the client side, once those are discussed and addressed will continue further review.  Sorry for being late on this.

pkg/madmin/constants.go
```diff
@@ -20,6 +20,6 @@ const (
 	// Unsigned payload.
 	unsignedPayload = "UNSIGNED-PAYLOAD"
 
-	// Admin operation header.
+	// Admin operation header. TODO: remove this
 	minioAdminOpHeader = "X-Minio-Operation"
```
harshavardhana: why keep it?

pkg/madmin/generic-commands.go
```diff
@@ -19,36 +19,33 @@ package madmin
...
 		return err
 	}
-	reqData.contentBody = bytes.NewReader(body)
-	reqData.contentLength = int64(len(body))
-	reqData.contentMD5Bytes = sumMD5(body)
-	reqData.contentSHA256Bytes = sum256(body)
+
+	// Setup new request
+	reqData := requestData{
+		relPath:            "/v1/config/credential",
```
harshavardhana: We should fail this operation if not on TLS connection, are we doing that?

pkg/madmin/service-commands.go
```diff
@@ -18,69 +18,79 @@
...
 	}
 
-	err = json.Unmarshal(respBytes, &serviceStatus)
-	if err != nil {
-		return ServiceStatusMetadata{}, err
-	}
-
-	return serviceStatus, nil
+	err = json.Unmarshal(respBytes, &ss)
+	return ss, nil
```
harshavardhana: Error is ignored? for json.Unmarshal?

cmd/admin-heal-ops.go
```diff
@@ -0,0 +1,836 @@
...
+
+const (
+	// a heal sequence with this many un-consumed heal result
+	// items blocks until heal-status consumption resumes or is
+	// aborted due to timeout.
+	maxUnconsumedHealResultItems = 1000
+
+	// if no heal-results are consumed (via the heal-status API)
+	// for this timeout duration, the heal sequence is aborted.
+	healUnconsumedTimeout = time.Hour * 24
```
harshavardhana: Perhaps `24 * time.Hour` is better style than `time.Hour * 24` ?

cmd/admin-heal-ops.go
```diff
@@ -0,0 +1,836 @@
...
+	// Launch top-level background heal go-routine
+	go h.healSequenceStart()
+
+	// Launch clean-up routine to remove this heal sequence (after
+	// it ends) from the global state after
+	// keepHealSeqStateDuration has elapsed.
+	go func() {
+		for {
+			select {
+			case <-time.After(time.Minute):
```
harshavardhana: Use `time.NewTicker()` for initiating events and we should also wait on `globalServiceDoneCh` otherwise there is no clean way to close this go-routine.

harshavardhana (changes requested):
Still need to review more .. some initial comments.

krishnasrinivas:
> The heal-status fetching API also sends acknowledgements to the server via the ackIndex parameter. This represents the index of the highest numbered heal-record (i.e. the record index; it is also an ID but index is more accurate as these "IDs" are all serial numbers) received by the client. The server uses it to remove older heal records from its memory. This is thus a state changing operation, and so it uses POST.
> 
> While technically, a GET could also be used, it would not be idempotent or safe (the precise meaning of these terms is given in section 9.1.1 and 9.1.2 in the HTTP RFC):
> 
> non-idempotent because two requests with the same ackIndex can return different results - by the time the second request is received, the server may have more heal records to return.
> non-safe because making a request with an older ackIndex than one that is already sent to the server in a previous request would not work, as the server would have removed older records.
> Based on these properties, POST is the correct HTTP verb to use here.

General conventions for HTTP verbs for REST operations:
GET - any kind of read operation on entities
POST, PUT - creation of a new entity
PATCH - update an existing entity
DELETE - delete an entity

Looking at the POST definition in the RFC:
```
The POST method is used to request that the origin server accept the entity enclosed in the request as a new subordinate of the resource identified by the Request-URI in the Request-Line. POST is designed to allow a uniform method to cover the following functions:

      - Annotation of existing resources;
      - Posting a message to a bulletin board, newsgroup, mailing list,
        or similar group of articles;
      - Providing a block of data, such as the result of submitting a
        form, to a data-handling process;
      - Extending a database through an append operation.
```

i.e using POST to get the status of healing is very incorrect.

GET is the natural verb for getting the status of healing. You can design GET for heal-status to be idempotent and safe if you want to be strictly consistent with RFC. But doing so might have downsides. IMO we can be flexible in the implementation to suit our needs. If you'd like I can check with AB/Harsha if we strictly want to be idempotent/safe or be flexible in implementation.

> The reason it is removed is that the process of healing data at the core is simply decoding erasure coded blocks of data, reconstructing them and writing the blocks back again. Bitrot checking (i.e. checking if erasure coded blocks are modified due to disk errors, etc) is an essential part of checking that erasure coded blocks are sane - even the implementation of decoding erasure coded blocks (erasure.HealFile()) performs this checking. Without bitrot checking, simply reconstructing data can lead to garbage output as the input data blocks are not sanitized - this means that without bitrot checking we cannot ensure the integrity of data after healing.

We check bit-rot only for s.dataBlocks files, not all. `advanced` mode is to check all the disks. Or is there anything I am missing? Or are you saying we will change current heal code to do bitrot check on all the disks? In which case we don't need advanced mode.

pkg/madmin/constants.go
donatello: Fixed.

pkg/madmin/generic-commands.go
donatello: Fixed in both madmin and server handler to maximize chances that creds wont be sent over insecure connection. Good catch!

pkg/madmin/service-commands.go
donatello: Fixed.

cmd/admin-heal-ops.go
donatello: Agreed and fixed.

cmd/admin-heal-ops.go
donatello: Ok, we need to do this only because when `globalServiceDoneCh` triggers the server could be restarting/reloading (i.e. without the process exiting). If the server was only shutting down, it is not needed, as there is nothing to clean up. Fixed and added a new commit (that will later be squashed).

donatello:
@krishnasrinivas Your points are noted - I will respond after thinking about it a bit.

pkg/madmin/generic-commands.go
donatello: TODO: Need to figure out how to fix the tests as a result of this change though.

donatello: Fixed with a little hack.

donatello:
@krishnasrinivas

On researching about the problem of POST vs GET, it seems that neither GET or POST in this situation would be RESTful design. A RESTful design would look something like this:

```
GET /heal/
```
```json
{
  "status": "running",
  "resultsReady": true,
  "link": "/heal-records/123"
}
```

The GET request to the heal endpoint would just give the status and ask the client to fetch heal status records from another link. If no results are available, it would not return a link! Assuming a link is returned:

```
GET /heal-records/123
```
```json
{
  "status": "running",
  ...
  "Items": [
    {"id": 123, ...},
    // more records
    {"id": 150, ...}
  ]
}
```

Some records are returned. Now to allow the server to clean up such records, the client is required to call DELETE (needed because the client needs to let the server know explicitly that records are safe to delete):

```
DELETE /heal-records/123
```
```
{
   "result": "success",
   "message": "Records 123 through 150 were deleted"
}
```

To resume the client would again request the main endpoint to find the next page of results:

```
GET /heal/
```
```
{
  "status": "running",
  "resultsReady": false,
  "message": "processing"
}
```

As you said, this is quite complex and probably unnecessary.

> GET is the natural verb for getting the status of healing. You can design GET for heal-status to be idempotent and safe if you want to be strictly consistent with RFC. But doing so might have downsides. IMO we can be flexible in the implementation to suit our needs. If you'd like I can check with AB/Harsha if we strictly want to be idempotent/safe or be flexible in implementation.

It is probably worth a discussion. So far, I could see that `ListenBucketNotification` is not RESTful either - so it is perhaps ok for our purposes to deviate from a strictly RESTful design. As you say, GET for fetching status does look more natural. So I've switched back to GET for record fetching. The companion mc PR is also updated.

> We check bit-rot only for s.dataBlocks files, not all. advanced mode is to check all the disks. Or is there anything I am missing? Or are you saying we will change current heal code to do bitrot check on all the disks? In which case we don't need advanced mode.

The `disksWithAllParts()` function in `xl-v1-healing-common.go` checks bitrot on all parts of all disks already. This is happening before `ErasureStorage.HealFile()` is called to determine the disks that need healing. So we don't need advanced mode.

donatello:
@harshavardhana @krisis Could you continue with your reviews?

cmd/admin-heal-ops.go
```diff
@@ -0,0 +1,852 @@
...
+//
+// Heal results are persisted in server memory for
+// `keepHealSeqStateDuration`. This function also launches a
+// background routine to clean up heal results after the
+// aforementioned duration.
+func (ahs *allHealState) LaunchNewHealSequence(h *healSequence) error {
+	ahs.Lock()
+	defer ahs.Unlock()
+
+	// Check if new heal sequence overaps with any existing,
```
krisis: s/overaps/overlaps

cmd/admin-heal-ops.go
```diff
@@ -0,0 +1,852 @@
...
+	// running sequence
+	for k, hSeq := range ahs.healSeqMap {
+		if !hSeq.hasEnded() && (strings.HasPrefix(k, h.path) ||
```
krisis: shouldn't we start a new heal sequence when,
`h.path = "/bucket/dir1"` and `k == "/bucket/dir1/dir2"`?

I.e, the heal sequence keyed by `k` may not heal objects under `"bucket/dir1/dir3"` say.

cmd/admin-heal-ops.go
```diff
@@ -0,0 +1,852 @@
...
+		for {
+			select {
+			// Check every minute if heal sequence has ended.
+			case <-everyMinute:
+				if h.hasEnded() {
+					keepStateTimeout = time.After(keepHealSeqStateDuration)
+					everyMinute = nil
+				}
+
+			// This case does not fire until the above one
```
krisis: s/until the above one/until the heal sequence completes ?

cmd/admin-heal-ops.go
```diff
@@ -0,0 +1,852 @@
...
+		return status, ErrHealNoSuchProcess
+	}
+
+	// For a statistics only heal sequence, just return it.
+	if h.settings.StatisticsOnly {
+		return h.currentStatus, ErrNone
+	}
+
+	// Check if ackIndex is valid
+	if ackIndex == -1 {
```
krisis: `ackIndex < 0` would cover all negative numbers supplied accidentally or otherwise.

cmd/admin-handlers.go
```diff
@@ -54,52 +54,69 @@ const (
...
+	// This struct literal represents the Admin API version that
+	// the server uses.
+	adminAPIVersionInfo = madmin.AdminAPIVersionInfo{"1"}
+)
+
+// VersionHandler - GET /minio/admin/version
+// -----------
+// Returns Administration API version
+func (adminAPI adminAPIHandlers) VersionHandler(w http.ResponseWriter,
+	r *http.Request) {
```
harshavardhana: This style of odd indentation is not needed in case of VersionHandler()  just change the `(adminAPI adminAPIHandlers)` to simply `(a adminAPIHandlers)`

cmd/admin-handlers.go
```diff
@@ -422,474 +403,245 @@ func (adminAPI adminAPIHandlers) ClearLocksHandler(w http.ResponseWriter, r *htt
...
-	case numOfflineDisks > 0:
-		state = healPartial
+	vars := mux.Vars(r)
+	bucket, objPrefix := vars["bucket"], vars["prefix"]
+	healPath = bucket + "/" + objPrefix
 
-	default:
-		state = healOK
+	qParams := r.URL.Query()
+	ackIndexStrs := qParams["ackIndex"]
```
harshavardhana: Is there a better name for this? why not simply `marker`

pkg/madmin/API.md
```diff
@@ -62,8 +62,25 @@ __Parameters__
 |`secretAccessKey`  | _string_  |Secret key for the object storage endpoint.   |
 |`ssl`   | _bool_  | Set this value to 'true' to enable secure (HTTPS) access.  |
 
+## 2. Admin API Version
```
harshavardhana: Get rid of APIs from docs which are not available anymore like HealUpload() ?

cmd/admin-router.go
```diff
@@ -27,50 +33,54 @@ func registerAdminRouter(mux *router.Router) {
...
+	// Version handler
+	adminRouter.Methods("GET").Path("/version").HandlerFunc(adminAPI.VersionHandler)
+
+	adminV1Router := adminRouter.PathPrefix("/v1").Subrouter()
 
 	/// Service operations
 
 	// Service status
-	adminRouter.Methods("GET").Queries("service", "").Headers(minioAdminOpHeader, "status").HandlerFunc(adminAPI.ServiceStatusHandler)
+	adminV1Router.Methods("GET").Path("/service").HandlerFunc(adminAPI.ServiceStatusHandler)
```
harshavardhana: Lets use ``s/"GET"/http.MethodGet`` everywhere for all http methods in this file.

harshavardhana (changes requested):
Testing this right now..

cmd/admin-heal-ops.go
```diff
@@ -0,0 +1,852 @@
...
+	// path is just bucket + "/" + objPrefix
+	path string
+
+	// heal settings applied to this heal sequence
+	settings healSettings
+
+	// current accumulated status of the heal sequence
+	currentStatus healSequenceStatus
+
+	// channel signalled by background routine when traveral has
```
krisis: s/traveral/traversal

harshavardhana (changes requested):
Also are you not updating pkg/admin-api/README.md in this PR? @donatello

donatello:
EDITED: Fixed service stop API bug, and config set handler API bug identified by @harshavardhana at https://github.com/minio/mc/pull/2239#issuecomment-329582726

mc PR is also updated to be in sync with this one.

donatello:
> Also are you not updating pkg/admin-api/README.md in this PR? @donatello

@harshavardhana I can update it, along with examples, but it may be preferable to do it in a follow-up PR as this one is already quite long. Whatever you suggest is fine.

cmd/admin-handlers.go
donatello: Done.

cmd/admin-handlers.go
donatello: Yes, good idea, it will be more similar to parameters used in AWS S3 APIs. Updated the PR.

pkg/madmin/API.md
donatello: Fixed the table - that was the place where old funcs were still present.

cmd/admin-router.go
donatello: Done.

harshavardhana:
> @harshavardhana I can update it, along with examples, but it may be preferable to do it in a follow-up PR as this one is already quite long. Whatever you suggest is fine.

Since we are doing a big change anyways, i guess its okay.

harshavardhana (dismissed):
Overall the changes are good and work, But we need to decide on the `mc` output UI which is a bigger work IMHO than server side API :-). LGTM

donatello:
Rebased this PR on master. There seems to be an unrelated build failure on AppVeyor.

krisis (changes requested):
Here are some comments, I will continue reviewing more recent changes.

cmd/admin-heal-ops.go
donatello: > I.e, the heal sequence keyed by k may not heal objects under "bucket/dir1/dir3" say.

However, the running heal sequence at `h.path` could already be scanning `bucket/dir1/dir3`. This rule is only to make sure the heal sequences do not compete with other for locks on the same objects.

cmd/admin-heal-ops.go
donatello: Fixed.

cmd/admin-heal-ops.go
donatello: Fixed.

cmd/admin-heal-ops.go
donatello: Fixed.

donatello:
@krishnasrinivas Waiting for your review.

cmd/admin-heal-ops.go
krisis: > However, the running heal sequence at h.path could already be scanning bucket/dir1/dir3. This rule is only to make sure the heal sequences do not compete with other for locks on the same objects.

In this e.g `h` is to be launched, so there isn't an heal sequence scanning `"/bucket/dir1/dir3"` (yet).

donatello: Ah, I misunderstood. But the same problem exists, in this case too - if the set of objects scanned by two heal sequences overlaps, they may compete for locks on the same object(s). In addition, the heal sequence that scans the same object for the second time is duplicating the work of the previous one. It is good to have this rule to facilitate some amount user-driven parallelism in healing objects.

krisis: OK, that makes sense.

cmd/admin-router.go
```diff
@@ -27,50 +35,54 @@ func registerAdminRouter(mux *router.Router) {
...
 
+	// Update credentials
+	adminV1Router.Methods(http.MethodPost).Path("/config/credential").HandlerFunc(adminAPI.UpdateCredentialsHandler)
 	// Get config
-	adminRouter.Methods("GET").Queries("config", "").Headers(minioAdminOpHeader, "get").HandlerFunc(adminAPI.GetConfigHandler)
-	// Set Config
-	adminRouter.Methods("PUT").Queries("config", "").Headers(minioAdminOpHeader, "set").HandlerFunc(adminAPI.SetConfigHandler)
+	adminV1Router.Methods(http.MethodGet).Path("/config").HandlerFunc(adminAPI.GetConfigHandler)
+	// Set config
+	adminV1Router.Methods(http.MethodPut).Path("/config").HandlerFunc(adminAPI.SetConfigHandler)
```
krishnasrinivas, line 87: the spec https://gist.github.com/donatello/e520308bf83adddc70366ea01ef505e5 says POST for this.

POST will be good as the calls are either GET or POST or DELETE. There is no PUT, so we'll never get confused on if a call is PUT or POST.

donatello: @krishnasrinivas PUT seems to me to be appropriate here because it completely replaces the configuration and is safe to repeat (it is idempotent) (if the config update succeeds at least once, it can be repeated with no ill-effects) - this is not in general true of POST. I think the issue of confusing with POST is a somewhat lesser one because the spec will be corrected and most users will use the client tools or the madmin package.

donatello:
This PR is rebased with latest master - conflicts fixed. Please continue reviewing @krishnasrinivas @krisis 

ADDENDUM: appveyor build failure appears to be unrelated. Please disregard it for now.

krishnasrinivas (dismissed):

cmd/admin-heal-ops.go
```diff
@@ -0,0 +1,854 @@
...
+
+	// Launch top-level background heal go-routine
+	go h.healSequenceStart()
+
+	// Launch clean-up routine to remove this heal sequence (after
+	// it ends) from the global state after timeout has elapsed.
+	go func() {
+		var keepStateTimeout <-chan time.Time
+		ticker := time.NewTicker(time.Minute)
+		everyMinute := ticker.C
```
krisis: > greater than zero; if not, NewTicker will panic. Stop the ticker to release associated resources. 

we need to stop the ticker to release associated resources.
```go
everyMinute := ticker.C
defer ticker.Stop()
```

krisis (commented):
Overall PR LGTM. There is a minor leak. If you confirm it is not an issue, I will approve.

harshavardhana:
> ADDENDUM: appveyor build failure appears to be unrelated. Please disregard it for now.

https://github.com/minio/minio/pull/5009 this PR mitigates the problem of windows build for the time being.  We should reduce number of files and merge files if possible @donatello

cmd/admin-heal-ops.go
donatello: Fixed.

deekoder:
blocked till we discuss and close on https://github.com/minio/mc/pull/2239

donatello:
This PR is updated with latest master again, modulo build results that are yet to come.

harshavardhana:
Looks like appveyor failed @donatello can you take a look?

donatello:
> Looks like appveyor failed @donatello can you take a look?

Fixed a race - should pass now.

donatello:
Closing this one, as further discussions with @abperiasamy has lead to another re-write of the heal APIs.  The new heal API is now at https://github.com/minio/minio/pull/5351

PR #4828, opened 2017-08-16
fix: [minor] Avoid unnecessary typecasting.

harshavardhana:
## Description
Avoid unnecessary typecasting.

## Motivation and Context
We don't need to typecast identifiers from
their base to type to same type again. This
is not a bug and compiler is fine to skip
it but it is better to avoid if not needed.

## How Has This Been Tested?
Manually using unconvert tool. 

## Types of changes

## Checklist:


balamurugana (approved):

fwessels (approved):
LGTM

PR #4833, opened 2017-08-19
tests: fix error conditions to be used properly.

harshavardhana:
## Description
fix error conditions to be used properly.

## Motivation and Context
Code review and cleanup. 

## How Has This Been Tested?
Manually.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/xl-v1-object.go
+++ b/cmd/xl-v1-object.go
@@ -98,7 +98,7 @@ func (xl xlObjects) CopyObject(srcBucket, srcObject, dstBucket, dstObject string
 			return oi, toObjectErr(err, srcBucket, srcObject)
 		}
 		// Rename atomically `xl.json` from tmp location to destination for each disk.
-		if onlineDisks, err = renameXLMetadata(onlineDisks, minioMetaTmpBucket, tempObj, srcBucket, srcObject, xl.writeQuorum); err != nil {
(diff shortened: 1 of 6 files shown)


cmd/xl-v1-object.go
```diff
@@ -680,8 +680,7 @@ func (xl xlObjects) PutObject(bucket string, object string, size int64, data io.
 	}
 
 	// Rename the successfully written temporary object to final location.
-	onlineDisks, err = renameObject(onlineDisks, minioMetaTmpBucket, tempObj, bucket, object, xl.writeQuorum)
-	if err != nil {
+	if _, err = renameObject(onlineDisks, minioMetaTmpBucket, tempObj, bucket, object, xl.writeQuorum); err != nil {
```
vadmeste, line 683: I don't encourage removing the update of onlineDisks here and in other places since we can forget this stuff when someone introduces new code after this line, it is better if we keep onlineDisks always updated.

harshavardhana: If we are not going to use it further its not useful, a function signature already communicates what it does.  Is onlineDisks being used again ? if not we shouldn't set it.  We should review when such a code is added in future to use onlineDisks by using the value set by the renameObject() 

Otherwise it doesn't make sense to set it IMO if we are not going to use it further.

harshavardhana:
@fwessels any reviews?

fwessels (approved):
LGTM

vadmeste (approved):
LGTM

deekoder:
Codecoverage should run. Blocked on merging.

PR #4845, opened 2017-08-22
remove bcrypt code from code-base (#4844)

### LINKED ISSUE #4844: do not compare credentials using bcrypt
Currently bcrypt is used to generate a hash from the secret key. This hash is used to compare the secret key send by clients during login. The whole bcrypt code is not necessary and uses the wrong primitive for this use case. A simple constant-time-compare is sufficient.

   
## Expected Behavior
Use constant time compare of secret key

## Current Behavior
Compute bcrypt value of secret key and use the equal function of bcrypt to compare the hash values.

## Possible Solution
Use subtle.ConstantTimeCompare

aead:
Bcrypt is not neccessary and not used properly. This change
replace the whole bcrypt hash computation through a constant time
compare and removes bcrypt from the code base.

Fixes #4844 

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:


harshavardhana:
> hole bcrypt hash computation

what is a `hole bcrypt hash` ? do you mean `whole` ?

aead:
Ahm, yes xD - I fix it!

balamurugana (commented):
LGTM for removing codes

One question.
Does `subtle.ConstantTimeCompare()` prevent timing attack?

aead:
Yes, that's exactly the function used in bcrypt for comparing the computed hash values. It compares two byte slices in constant time. This prevents timing attacks.

aead:
Closed until we decide what to do...

aead:
This PR should fulfill the requirements - reopened

cmd/credential.go
```diff
@@ -80,64 +80,44 @@ func (cred credential) Equal(ccred credential) bool {
...
-		}
-
-		cred.secretKeyHash = secretKeyHash
+	if subtle.ConstantTimeCompare([]byte(cred.AccessKey), []byte(ccred.AccessKey)) != 1 {
+		return false
 	}
-
-	return (cred.AccessKey == ccred.AccessKey &&
-		bcrypt.CompareHashAndPassword(cred.secretKeyHash, []byte(ccred.SecretKey)) == nil)
+	return subtle.ConstantTimeCompare([]byte(cred.SecretKey), []byte(ccred.SecretKey)) == 1
```
balamurugana: How about doing below?
```golang
return (cred.AccessKey == ccred.AccessKey && subtle.ConstantTimeCompare([]byte(cred.SecretKey), []byte(ccred.SecretKey)) == 1)
```

cmd/credential.go
```diff
@@ -80,64 +80,44 @@ func (cred credential) Equal(ccred credential) bool {
...
 }
 
 // Initialize a new credential object
 func getNewCredential(accessKeyLen, secretKeyLen int) (cred credential, err error) {
-	// Generate access key.
-	keyBytes := make([]byte, accessKeyLen)
-	_, err = rand.Read(keyBytes)
-	if err != nil {
-		return cred, err
+	// Generate access key - the access key consits only of alpha-numeric digits.
```
balamurugana: I think you can keep function unchanged to minimize this PR

cmd/credential.go
aead: The access key is public information so this is fine. However for me it looks a lite bit wired when credentials are compared and some parts are constant time and some aren't. It's just easier to miss something in the future - but the decision is up to you...

cmd/credential.go
aead: This PR also cleans up the code a bit - like unnecessary if-else. aso.

cmd/credential.go
balamurugana: I think you can do the change.

cmd/credential.go
balamurugana: Yep. Its better to send different PR doing cleanup.

cmd/credential.go
aead: done

cmd/credential.go
aead: done

harshavardhana (approved):

balamurugana (approved):

harshavardhana:
Can you squash your commits? @aead

aead:
Squashed

PR #4846, opened 2017-08-22
fs: ListObjects() was reading ETag at wrong offsets

### LINKED ISSUE #4842: Api and mc return 0 objects / output when downloading from the same dir at the sametime
## Expected Behavior
I would expect the api and mc to beable to do and ls at the same time as a download from a bucket and if they are unable to access not silently fail i.e. the api returning an empty list and mc to none 0.

## Current Behavior
While downloading a 100Mb file on one client if a second does a ls in the same bucket and path it will randomly give no results but does not give an error either

1. start to down load a larger file from minio
2. in a loop run mc ls on the files containing path

## Context
I'm trying to use minio in a CI system which needs to check that all the needed files have been uploaded before it can start testing. The files follow a naming convention but i need to beable to use regex when looking for them so need the ls step before trying to download.

## Your Environment
Standalone docker minio container running under docker-ce on centos7

@vadmeste was able to quickly reproduce while I was talking to him in the slack channel

harshavardhana:
## Description
Current code was just using io.ReadAll() on an fd()
which might have moved underneath due to a concurrent
read operation. Subsequent read will result in EOF
We should always seek back and read again. pread()
is allowed on all platforms use io.SectionReader to
read from the beginning of the file.

## Motivation and Context
Fixes #4842

## How Has This Been Tested?
Manually and using the reproducer. 

## Types of changes

## Checklist:

### DIFF
--- a/cmd/fs-v1.go
+++ b/cmd/fs-v1.go
@@ -749,14 +749,24 @@ func (fs fsObjects) getObjectETag(bucket, entry string) (string, error) {
 	// Read from fs metadata only if it exists.
 	defer fs.rwPool.Close(fsMetaPath)
 
-	fsMetaBuf, err := ioutil.ReadAll(rlk.LockedFile)
+	// Fetch the size of the underlying file.
+	fi, err := rlk.LockedFile.Stat()
 	if err != nil {
-		// `fs.json` can be empty due to previously failed
-		// PutObject() transaction, if we arrive at such
-		// a situation we just ignore and continue.
-		if errorCause(err) != io.EOF {
-			return "", toObjectErr(err, bucket, entry)
-		}
+		return "", toObjectErr(traceError(err), bucket, entry)
+	}
+
+	// `fs.json` can be empty due to previously failed
+	// PutObject() transaction, if we arrive at such
+	// a situation we just ignore and continue.
+	if fi.Size() == 0 {
+		return "", nil
+	}
+
+	// Wrap the locked file in a ReadAt() backend section reader to
+	// make sure the underlying offsets don't move.
+	fsMetaBuf, err := ioutil.ReadAll(io.NewSectionReader(rlk.LockedFile, 0, fi.Size()))
+	if err != nil {
+		return "", traceError(err)
 	}
 
 	// Check if FS metadata is valid, if not return error.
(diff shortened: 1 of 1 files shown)


vadmeste:
This fixes the issue as I tested it. You mean, in a GNU/Linux, a call to a read() of a moved file returns io.EOF ?

harshavardhana:
> This fixes the issue as I tested it. You mean, in a GNU/Linux, a call to a read() of a moved file returns io.EOF ?

Well the underlying offset would have moved if there is a concurrent read. io.SectionReader helps in seeking back the offsets for the requested Size() and Offset.

cmd/fs-v1.go
```diff
@@ -749,14 +749,21 @@ func (fs fsObjects) getObjectETag(bucket, entry string) (string, error) {
...
+
+	fsMetaBuf, err := ioutil.ReadAll(io.NewSectionReader(rlk.LockedFile, 0, fi.Size()))
+	if err != nil {
+		return "", traceError(err)
+	}
+
+	// `fs.json` can be empty due to previously failed
+	// PutObject() transaction, if we arrive at such
+	// a situation we just ignore and continue.
+	if len(fsMetaBuf) == 0 {
```
balamurugana: You could use `fi.Size()` to validate before `ioutil.ReadAll()`

cmd/fs-v1.go
```diff
@@ -749,14 +749,21 @@ func (fs fsObjects) getObjectETag(bucket, entry string) (string, error) {
 	// Read from fs metadata only if it exists.
 	defer fs.rwPool.Close(fsMetaPath)
 
-	fsMetaBuf, err := ioutil.ReadAll(rlk.LockedFile)
+	fi, err := rlk.LockedFile.Stat()
```
balamurugana: You could add a comment why `Stat()` then `ioutil.ReadAll()` approach is required with `ReadAt()`

balamurugana (changes requested):

harshavardhana:
@balamurugana can you take a look again?

balamurugana (approved):

ebozduman (approved):
LGTM

PR #4853, opened 2017-08-24
add the mount-observe plug to the snap

come-maiz:
## Description
Allow the snap to observe mounts, and added the command the user needs to execute to connect this interface.

## Motivation and Context
Some users try to mount directories inside the minio backend, which leads to errors during basic operations, like put object. Minio reads /proc/mounts to be sure that the backend doesn't contain any mounted dirs.
Rename() between partitions doesn't work, that's why we need to do that check

Thanks to @vadmeste for that explanation.

The mount observe is not autoconnect unlike the other plugs minio is using, because it gives privileged read access to mount arguments and should only be used with trusted apps.

You can ask to become a trusted app and get this plug autoconnected by explaining in the [forum](https://forum.snapcraft.io) how you are using it. Our security team will review the post, your code, and the continuous delivery to grant it, and the user will always be in control to disconnect it.

## How Has This Been Tested?
I tested without the plug connected and got an error.
I connected the plug and could start the minio server without errors.

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,9 +30,9 @@ minio server ~/Photos
 If you previously installed minio using `brew install minio` then reinstall minio from `minio/stable/minio` official repo. Homebrew builds are unstable due to golang 1.8 bugs.
 
 ```
-brew uninstall minio 
+brew uninstall minio
 brew install minio/stable/minio
-```  
+```
 
 ### Binary Download
 | Platform| Architecture | URL|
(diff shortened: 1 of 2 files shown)


harshavardhana (commented):
LGTM

harshavardhana (approved):

harshavardhana:
ping @donatello

donatello (approved):

PR #4864, opened 2017-08-28
fs: Convert repeated code in rwpool.Open() into a single function.

harshavardhana:
## Description
fs: Convert repeated code in rwpool.Open() into a single function.

## Motivation and Context
Refer https://github.com/minio/minio/issues/4658 for more information.

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:

### DIFF
--- a/cmd/fs-v1-rwpool.go
+++ b/cmd/fs-v1-rwpool.go
@@ -31,24 +31,16 @@ type fsIOPool struct {
 	readersMap map[string]*lock.RLockedFile
 }
 
-// Open is a wrapper call to read locked file which
-// returns a ReadAtCloser.
+// lookupToRead - looks up an fd from readers map and
+// returns read locked fd for caller to read from, if
+// fd found increments the reference count. If the fd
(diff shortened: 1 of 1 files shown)


balamurugana (approved):

donatello (approved):
Change looks good to me and is good to do on its own, but I fail to understand what this change has to do with code-coverage or the racey tests described in #4658 - the lock usage appears to be exactly the same as before in this change.

harshavardhana:
> Change looks good to me and is good to do on its own, but I fail to understand what this change has to do with code-coverage or the racey tests described in #4658 - the lock usage appears to be exactly the same as before in this change.

Coverage on a micro level works at source code navigation, since the lock usage is exactly the same we are moving the code such that if at all it happens it happens under a self contained function. This is one step closer to narrowing down the real cause of the issue, without this refactor its hard to reason around existing code due to repetition.  One of the reason why this PR doesn't explicitly say anything related to fixing the situation presented at #4658

PR #4870, opened 2017-08-30
admin: Raise error if config and env credentials mismatch

### LINKED ISSUE #4866: Disallow admin setconfig
This came as a requirement during discussion for Cloud Foundry. But it is not related to CF in particular.

When access/secret keys are set through env vars, currently we allow setconfig through admin API even when the config's access/secret mismatch with the ones set through env vars . This should be disallowed. i.e adminAPIHandlers.SetConfigHandler() should return error when config's creds mismatch with the ones set through env vars.

poornas:
Fixes #4866. This PR raises an error when the config credentials passed to admin set config request does not match the accesskey and secretkey set via env. variables.
## Types of changes

## Checklist:

### DIFF
--- a/cmd/admin-handlers.go
+++ b/cmd/admin-handlers.go
@@ -979,6 +979,24 @@ func (adminAPI adminAPIHandlers) SetConfigHandler(w http.ResponseWriter, r *http
 		return
 	}
 
+	var config serverConfigV19
+	err = json.Unmarshal(configBytes, &config)
+
+	if err != nil {
+		errorIf(err, "Failed to unmarshal config from request body.")
+		writeErrorResponse(w, toAPIErrorCode(err), r.URL)
+		return
+	}
+
+	if globalIsEnvCreds {
+		creds := serverConfig.GetCredential()
+		if config.Credential.AccessKey != creds.AccessKey ||
+			config.Credential.SecretKey != creds.SecretKey {
+			writeErrorResponse(w, ErrAdminCredentialsMismatch, r.URL)
+			return
+		}
+	}
+
 	// Write config received from request onto a temporary file on
 	// all nodes.
 	tmpFileName := fmt.Sprintf(minioConfigTmpFormat, mustGetUUID())
--- a/cmd/api-errors.go
+++ b/cmd/api-errors.go
@@ -152,6 +152,7 @@ const (
 	ErrAdminInvalidAccessKey
 	ErrAdminInvalidSecretKey
 	ErrAdminConfigNoQuorum
+	ErrAdminCredentialsMismatch
 	ErrInsecureClientRequest
 )
 
@@ -630,6 +631,11 @@ var errorCodeResponse = map[APIErrorCode]APIError{
 	ErrAdminConfigNoQuorum: {
 		Code:           "XMinioAdminConfigNoQuorum",
 		Description:    "Configuration update failed because server quorum was not met",
+		HTTPStatusCode: http.StatusServiceUnavailable,
+	},
+	ErrAdminCredentialsMismatch: {
+		Code:           "XMinioAdminCredentialsMismatch",
+		Description:    "Credentials in config mismatch with server environment variables",
 		HTTPStatusCode: http.StatusServiceUnavailable,
 	},
 	ErrInsecureClientRequest: {


cmd/admin-handlers.go
```diff
@@ -979,6 +980,28 @@ func (adminAPI adminAPIHandlers) SetConfigHandler(w http.ResponseWriter, r *http
 		return
 	}
 
+	var config serverConfigV19
```
krisis: Thought (not to be treated as a suggestion necessarily):
Committing to a concrete type like `serverConfigV19` puts `pkg/madmin` API handler in a spot every time server configuration type is modified (version19 tells us something :) ). To get out of this situation we could do,
```go
       var creds struct {
		Credential struct {
			AccessKey string `json:"accessKey"`
			SecretKey string `json:"secretKey"`
		} `json:"credential"`
	}
	err := json.Unmarshal(configBytes, &creds)
	if err != nil {
		// error handling goes here
	}

```

The above snippet focuses only on that portion of the server configuration type that we require to validate here. Note: using `serverConfigV19` doesn't provide any type-safety since `pkg/madmin` and server reside in different packages.

cmd/admin-handlers.go
```diff
@@ -979,6 +980,28 @@ func (adminAPI adminAPIHandlers) SetConfigHandler(w http.ResponseWriter, r *http
...
+	err = json.Unmarshal(configBytes, &config)
+
+	if err != nil {
+		errorIf(err, "Failed to unmarshall config from request body.")
```
krisis: typo: s/unmarshall/unmarshal

cmd/api-errors.go
```diff
@@ -632,6 +633,11 @@ var errorCodeResponse = map[APIErrorCode]APIError{
 		Description:    "Configuration update failed because server quorum was not met",
 		HTTPStatusCode: http.StatusServiceUnavailable,
 	},
+	ErrAdminCredentialsMismatch: {
+		Code:           "XMinioAdminCredentialsMismatch",
+		Description:    "Credentials in config mismatch with server env vars",
```
krisis: s/env vars/environmental variables

I suggested using the expanded form since this is part of the descriptive error message.

krisis (changes requested):

krisis (commented):
Removed my review comment suggesting using anonymous struct. It was incorrect.

cmd/api-errors.go
poornas: done.

cmd/admin-handlers.go
poornas: done.

cmd/admin-handlers.go
```diff
@@ -979,6 +980,28 @@ func (adminAPI adminAPIHandlers) SetConfigHandler(w http.ResponseWriter, r *http
...
+			writeErrorResponse(w, ErrAdminCredentialsMismatch, r.URL)
+			return
+		}
+	}
+	if secretKey := os.Getenv("MINIO_SECRET_KEY"); secretKey != "" {
+		if secretKey != config.Credential.SecretKey {
+			writeErrorResponse(w, ErrAdminCredentialsMismatch, r.URL)
+			return
+		}
+	}
```
krishnasrinivas: we touch env vars only in handleCommonEnvVars()
have a look at handleCommonEnvVars() where we set globalIsEnvCreds
so you can do
```
if globalIsEnvCreds {
    creds := serverConfig.GetCredential()
    if config.Credential.AccessKey != creds.AccessKey {...}
    if config.Credential.SecretKey != creds.SecretKey {...}
}
```

poornas: that's a good suggestion.done

cmd/admin-handlers.go
```diff
@@ -979,6 +979,27 @@ func (adminAPI adminAPIHandlers) SetConfigHandler(w http.ResponseWriter, r *http
...
+		return
+	}
+
+	if globalIsEnvCreds {
+		creds := serverConfig.GetCredential()
+		if config.Credential.AccessKey != creds.AccessKey {
+			writeErrorResponse(w, ErrAdminCredentialsMismatch, r.URL)
+			return
+		}
+		if config.Credential.SecretKey != creds.SecretKey {
```
donatello: An `||` clause to combine the two if blocks will be better.

donatello (changes requested):

cmd/admin-handlers.go
poornas: done.

donatello (approved):

krishnasrinivas (approved):

poornas:
@kp, this PR awaiting your review.

krisis (approved):

poornas:
@deekoder, this can go forward.

PR #4874, opened 2017-08-31
azure: add stateless gateway support

balamurugana:
Previously init multipart upload stores metadata of an object which is
used for complete multipart.  This patch makes azure gateway to store
metadata information of init multipart object in azure in the name of
'minio.sys.tmp/multipart/v1/<UPLOAD-ID>/meta.json' and uses this
information on complete multipart.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/gateway-azure.go
+++ b/cmd/gateway-azure.go
@@ -17,13 +17,19 @@
 package cmd
 
 import (
+	"bytes"
+	"crypto/rand"
+	"crypto/sha256"
 	"encoding/base64"
+	"encoding/hex"
+	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
 	"net/url"
+	"strconv"
 	"strings"
-	"sync"
 	"time"
 
 	"github.com/Azure/azure-sdk-for-go/storage"
@@ -33,6 +39,7 @@ import (
 
 const globalAzureAPIVersion = "2016-05-31"
 const azureBlockSize = 100 * humanize.MiByte
+const metadataObjectNameTemplate = globalMinioSysTmp + "multipart/v1/%s.%x/azure.json"
 
 // Canonicalize the metadata headers, without this azure-sdk calculates
 // incorrect signature. This attempt to canonicalize is to convert
@@ -87,38 +94,9 @@ func azureToS3ETag(etag string) string {
 	return canonicalizeETag(etag) + "-1"
 }
 
-// To store metadata during NewMultipartUpload which will be used after
-// CompleteMultipartUpload to call SetBlobMetadata.
-type azureMultipartMetaInfo struct {
-	meta map[string]map[string]string
-	*sync.Mutex
-}
-
-// Return metadata map of the multipart object.
-func (a *azureMultipartMetaInfo) get(key string) map[string]string {
-	a.Lock()
-	defer a.Unlock()
-	return a.meta[key]
-}
-
-// Set metadata map for the multipart object.
-func (a *azureMultipartMetaInfo) set(key string, value map[string]string) {
-	a.Lock()
-	defer a.Unlock()
-	a.meta[key] = value
-}
-
-// Delete metadata map for the multipart object.
-func (a *azureMultipartMetaInfo) del(key string) {
-	a.Lock()
-	defer a.Unlock()
-	delete(a.meta, key)
-}
-
 // azureObjects - Implements Object layer for Azure blob storage.
 type azureObjects struct {
-	client   storage.BlobStorageClient // Azure sdk client
-	metaInfo azureMultipartMetaInfo
+	client storage.BlobStorageClient // Azure sdk client
 }
 
 // Convert azure errors to minio object layer errors.
@@ -177,6 +155,68 @@ func azureToObjectError(err error, params ...string) error {
 	return e
 }
 
+// mustGetAzureUploadID - returns new upload ID which is hex encoded 8 bytes random value.
+func mustGetAzureUploadID() string {
+	var id [8]byte
+
+	n, err := io.ReadFull(rand.Reader, id[:])
+	if err != nil {
+		panic(fmt.Errorf("unable to generate upload ID for azure. %s", err))
+	}
+	if n != len(id) {
+		panic(fmt.Errorf("insufficient random data (expected: %d, read: %d)", len(id), n))
+	}
+
+	return fmt.Sprintf("%x", id[:])
+}
+
+// checkAzureUploadID - returns error in case of given string is upload ID.
+func checkAzureUploadID(uploadID string) (err error) {
+	if len(uploadID) != 16 {
+		return traceError(MalformedUploadID{uploadID})
+	}
+
+	if _, err = hex.DecodeString(uploadID); err != nil {
+		return traceError(MalformedUploadID{uploadID})
+	}
+
+	return nil
+}
+
+// Encode partID, subPartNumber, uploadID and md5Hex to blockID.
+func azureGetBlockID(partID, subPartNumber int, uploadID, md5Hex string) string {
+	return base64.StdEncoding.EncodeToString([]byte(fmt.Sprintf("%05d.%02d.%s.%s", partID, subPartNumber, uploadID, md5Hex)))
+}
+
+// Parse blockID into partID, subPartNumber and md5Hex.
+func azureParseBlockID(blockID string) (partID, subPartNumber int, uploadID, md5Hex string, err error) {
+	var blockIDBytes []byte
+	if blockIDBytes, err = base64.StdEncoding.DecodeString(blockID); err != nil {
+		return
+	}
+
+	tokens := strings.Split(string(blockIDBytes), ".")
+	if len(tokens) != 4 {
+		err = fmt.Errorf("invalid block id '%s'", string(blockIDBytes))
+		return
+	}
+
+	if partID, err = strconv.Atoi(tokens[0]); err != nil || partID <= 0 {
+		err = fmt.Errorf("invalid part number in block id '%s'", string(blockIDBytes))
+		return
+	}
+
+	if subPartNumber, err = strconv.Atoi(tokens[1]); err != nil || subPartNumber <= 0 {
+		err = fmt.Errorf("invalid sub-part number in block id '%s'", string(blockIDBytes))
+		return
+	}
+
+	uploadID = tokens[2]
+	md5Hex = tokens[3]
+
+	return
+}
+
 // Inits azure blob storage client and returns AzureObjects.
 func newAzureLayer(host string) (GatewayLayer, error) {
 	var err error
@@ -200,10 +240,6 @@ func newAzureLayer(host string) (GatewayLayer, error) {
 
 	return &azureObjects{
 		client: c.GetBlobService(),
-		metaInfo: azureMultipartMetaInfo{
-			meta:  make(map[string]map[string]string),
-			Mutex: &sync.Mutex{},
-		},
 	}, nil
 }
 
@@ -286,6 +322,9 @@ func (a *azureObjects) ListObjects(bucket, prefix, marker, delimiter string, max
 	result.IsTruncated = resp.NextMarker != ""
 	result.NextMarker = resp.NextMarker
 	for _, object := range resp.Blobs {
+		if strings.HasPrefix(object.Name, globalMinioSysTmp) {
+			continue
+		}
 		t, e := time.Parse(time.RFC1123, object.Properties.LastModified)
 		if e != nil {
 			continue
@@ -299,6 +338,14 @@ func (a *azureObjects) ListObjects(bucket, prefix, marker, delimiter string, max
 			ContentType:     object.Properties.ContentType,
 			ContentEncoding: object.Properties.ContentEncoding,
 		})
+	}
+
+	// Remove minio.sys.tmp prefix.
+	for i, prefix := range resp.BlobPrefixes {
+		if prefix == globalMinioSysTmp {
+			resp.BlobPrefixes = append(resp.BlobPrefixes[:i], resp.BlobPrefixes[i+1:]...)
+			break
+		}
 	}
 	result.Prefixes = resp.BlobPrefixes
 	return result, nil
@@ -321,6 +368,9 @@ func (a *azureObjects) ListObjectsV2(bucket, prefix, continuationToken string, f
 		result.NextContinuationToken = resp.NextMarker
 	}
 	for _, object := range resp.Blobs {
+		if strings.HasPrefix(object.Name, globalMinioSysTmp) {
+			continue
+		}
 		t, e := time.Parse(time.RFC1123, object.Properties.LastModified)
 		if e != nil {
 			continue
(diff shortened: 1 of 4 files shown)


krishnasrinivas (commented):
Since uploadID (UUID) is not encoded in the blockID, ListObjectParts will return block entries from all the uploadIDs instead of returning entries for the requested uploadID

So uploadIDs must be part of blockIDs.

cmd/gateway-azure.go
```diff
@@ -161,6 +136,75 @@ func azureToObjectError(err error, params ...string) error {
 	return e
 }
 
+type azureUploadID [8]byte
+
+func (id azureUploadID) String() string {
+	return fmt.Sprintf("%x", id[:])
+}
+
+func mustGetAzureUploadID() *azureUploadID {
```
krishnasrinivas: this can return a `string` and `azureUploadID` type can be removed as defining a new type for this is not being useful anywhere in the code.

cmd/gateway-azure.go
```diff
@@ -457,70 +497,58 @@ func (a *azureObjects) DeleteObject(bucket, object string) error {
...
+func (a *azureObjects) CopyObjectPart(srcBucket, srcObject, destBucket, destObject string, uploadID string, partID int, startOffset int64, length int64) (info PartInfo, err error) {
+	return info, traceError(NotImplemented{})
 }
 
 // PutObjectPart - Use Azure equivalent PutBlockWithLength.
 func (a *azureObjects) PutObjectPart(bucket, object, uploadID string, partID int, size int64, data io.Reader, md5Hex string, sha256sum string) (info PartInfo, err error) {
-	if meta := a.metaInfo.get(uploadID); meta == nil {
+	metadataObject := fmt.Sprintf(metadataObjectTemplate, uploadID)
+	if _, err = a.client.GetBlobMetadata(bucket, metadataObject); err != nil {
+		errorIf(err, "unable to read metadata object for upload ID %s", uploadID)
```
krishnasrinivas: `errorIf` statements can be removed as all the logging happens at the S3 layer

cmd/gateway-azure.go
```diff
@@ -37,6 +40,7 @@ import (
 
 const globalAzureAPIVersion = "2016-05-31"
 const azureBlockSize = 100 * humanize.MiByte
+const metadataObjectTemplate = "minio.sys.tmp/multipart/v1/%s/meta.json"
```
krishnasrinivas: instead you can have

```
func getAzureMultipartMetadataPath(bucket, object, uploadID string) string {
    id := sha256(bucket+object) + "." + uploadID
    return fmt.Sprintf("minio.sys.tmp/multipart/v1/%s/azure.json", id)
}
```
since uploadID is just 8 bytes having sha256 of bucket and object will make this more hash collision resistant
you can name the file `azure.json` as we have `gcs.json` for gcs

cmd/gateway-azure.go
```diff
@@ -457,70 +497,58 @@ func (a *azureObjects) DeleteObject(bucket, object string) error {
...
-}
-
-// Parse blockID into partID, subPartNumber and md5Hex.
-func azureParseBlockID(blockID string) (partID, subPartNumber int, md5Hex string, err error) {
-	var blockIDBytes []byte
-	if blockIDBytes, err = base64.StdEncoding.DecodeString(blockID); err != nil {
-		return
+	var jsonData []byte
+	if jsonData, err = json.Marshal(multipartMetadata{Name: object, Metadata: metadata}); err != nil {
+		return "", err
```
krishnasrinivas: trace this error

cmd/gateway-azure.go
```diff
@@ -592,66 +620,64 @@ func (a *azureObjects) PutObjectPart(bucket, object, uploadID string, partID int
...
+func (a *azureObjects) AbortMultipartUpload(bucket, object, uploadID string) (err error) {
+	metadataObject := fmt.Sprintf(metadataObjectTemplate, uploadID)
+	if _, err = a.client.GetBlobMetadata(bucket, metadataObject); err != nil {
+		errorIf(err, "unable to read metadata object for upload ID %s", uploadID)
+		return traceError(InvalidUploadID{})
+	}
+
+	err = a.client.DeleteBlob(bucket, metadataObject, nil)
+
+	return err
```
krishnasrinivas: can be one line:
`return a.client.DeleteBlob(bucket, metadataObject, nil)`

balamurugana: Done

cmd/gateway-azure.go
balamurugana: I can't understand it. As `return info, traceError(InvalidUploadID{})` is done, I am unable to understand the problem of using `errorIf()`

cmd/gateway-azure.go
balamurugana: Done

cmd/gateway-azure.go
balamurugana: Done

cmd/gateway-azure.go
balamurugana: Encoding bucket and object name would prevent recovery case using upload ID.   However this does not prevent the problem completely in situation like in another instance of gateway gets previously generated upload ID for same bucket/object would lead to meta data loss.

To avoid these problem (to some extent), I added upload id existent check in new multipart upload call.

> you can name the file azure.json as we have gcs.json for gcs

Done

cmd/gateway-azure.go
krishnasrinivas: Convention that we follow is we don't log twice for the same error. We just log once at the S3 layer. Doing traceError() would mean we record the line number where it happened and hence know what the error is about.

If we are unable to return traceError for some reason then we log using errorIf

cmd/gateway-azure.go
krishnasrinivas: > Encoding bucket and object name would prevent recovery case using upload ID.

it can still be done as the uploadID is part of the metadataPath (metadataPath also contains sha256 which is additional info)

> However this does not prevent the problem completely in situation like in another instance of gateway gets previously generated upload ID for same bucket/object would lead to meta data loss.

That is right, what I pointed out is that we can make hash collision more resistant (than what it is currently) by including sha256 of `bucket/object`, 

> To avoid these problem (to some extent), I added upload id existent check in new multipart upload call.

OK, if this is done then it is good enough.

cmd/gateway-azure.go
```diff
@@ -457,38 +487,54 @@ func (a *azureObjects) DeleteObject(bucket, object string) error {
...
+	Metadata map[string]string `json:"metadata"`
+}
+
+func getMetadataObjectName(objectName, uploadID string) string {
+	return fmt.Sprintf("minio.sys.tmp/multipart/v1/%s.%x/azure.json", uploadID, sha256.Sum256([]byte(objectName)))
+}
+
+func (a *azureObjects) checkUploadIDExists(bucketName, objectName, uploadID string) (err error) {
+	_, err = a.client.GetBlobMetadata(bucketName, getMetadataObjectName(objectName, uploadID))
+	err = traceError(azureToObjectError(err))
```
krishnasrinivas: Change the above line to:
```
err = azureToObjectError(traceError(err), bucketName, objectName)
```

azureToObjectError expects traced error according to convention.

cmd/gateway-azure.go
```diff
@@ -457,38 +487,54 @@ func (a *azureObjects) DeleteObject(bucket, object string) error {
...
+	metadata = s3ToAzureHeaders(metadata)
+
+	var jsonData []byte
+	if jsonData, err = json.Marshal(multipartMetadata{Name: object, Metadata: metadata}); err != nil {
+		return "", traceError(err)
+	}
+
+	err = a.client.CreateBlockBlobFromReader(bucket, metadataObject, uint64(len(jsonData)), bytes.NewBuffer(jsonData), nil)
+	if err != nil {
+		return "", err
```
krishnasrinivas: return azureToObjectError(traceError(err))

cmd/gateway-azure.go
```diff
@@ -457,38 +487,54 @@ func (a *azureObjects) DeleteObject(bucket, object string) error {
...
-		// the client to do a fresh upload so that any metadata like content-type are sent again in the
-		// NewMultipartUpload.
-		return result, nil
-	}
-	result.Uploads = []uploadMetadata{{prefix, prefix, UTCNow(), "", nil}}
+	// It's decided not to support List Multipart Uploads, hence returning empty result.
 	return result, nil
 }
 
+type multipartMetadata struct {
```
krishnasrinivas: rename to `azureMultipartMetadata` since it is a global type, needs to have azure prefix as it is azure specific

cmd/gateway-azure.go
balamurugana: Done

cmd/gateway-azure.go
```diff
@@ -457,38 +487,54 @@ func (a *azureObjects) DeleteObject(bucket, object string) error {
...
+	Name     string            `json:"name"`
+	Metadata map[string]string `json:"metadata"`
+}
+
+func getMetadataObjectName(objectName, uploadID string) string {
```
krishnasrinivas: rename to getAzureMultipartMetadataPath() so that it is consistent with what it returns (which is azureMultipartMetadata's path)

you can also use getAzureMultipartMetadataObjectName() if you are particular about using `ObjectName` instead of `Path`

cmd/gateway-azure.go
balamurugana: Done

cmd/gateway-azure.go
balamurugana: Done

cmd/gateway-azure.go
balamurugana: As this function returns `minio.sys.tmp/multipart/v1/%s.%x/azure.json` which is metadata object name of given object, I think `getAzureMetadataObjectName()` is clear enough

krishnasrinivas (approved):

cmd/gateway-azure.go
```diff
@@ -181,6 +156,65 @@ func azureToObjectError(err error, params ...string) error {
 	return e
 }
 
+func mustGetAzureUploadID() string {
```
harshavardhana: Can we not use mustGetUUID() ?  if not can we write a comment on this new implementation?

cmd/gateway-azure.go
```diff
@@ -181,6 +156,65 @@ func azureToObjectError(err error, params ...string) error {
...
+	fatalIf(err, "unable to generate upload ID for azure")
+	if n != len(id) {
+		fatalIf(fmt.Errorf("insufficient random data (expected: %d, read: %d)", len(id), n),
+			"unable to generate upload ID for azure")
+	}
+
+	return fmt.Sprintf("%x", id[:])
+}
+
+func checkAzureUploadID(uploadID string) (err error) {
```
harshavardhana: We would need some unit tests for these functions..

cmd/gateway-azure.go
```diff
@@ -181,6 +156,65 @@ func azureToObjectError(err error, params ...string) error {
 	return e
 }
 
+func mustGetAzureUploadID() string {
+	var id [8]byte
+
+	n, err := io.ReadFull(rand.Reader, id[:])
+	fatalIf(err, "unable to generate upload ID for azure")
+	if n != len(id) {
+		fatalIf(fmt.Errorf("insufficient random data (expected: %d, read: %d)", len(id), n),
```
harshavardhana: fatalIf is a bad idea we should just panic() instead - http server will catch panics and gracefully handle it.

harshavardhana (changes requested):

cmd/gateway-azure.go
balamurugana: Added comment

cmd/gateway-azure.go
balamurugana: Done

cmd/gateway-azure.go
balamurugana: done

cmd/gateway-azure.go
```diff
@@ -181,6 +156,66 @@ func azureToObjectError(err error, params ...string) error {
 	return e
 }
 
+// mustGetAzureUploadID - returns new upload ID which is hex encoded 8 bytes random value.
+func mustGetAzureUploadID() string {
+	var id [8]byte
+
+	n, err := io.ReadFull(rand.Reader, id[:])
+	panic(fmt.Errorf("unable to generate upload ID for azure. %s", err))
```
harshavardhana: This might panic all the time @balamurugana if will have to do `if err != nil` on this one.

cmd/gateway-azure.go
```diff
@@ -304,12 +342,24 @@ func (a *azureObjects) ListObjects(bucket, prefix, marker, delimiter string, max
...
+			break
+		}
+	}
 	result.Prefixes = resp.BlobPrefixes
 	return result, nil
 }
 
 // ListObjectsV2 - list all blobs in Azure bucket filtered by prefix
 func (a *azureObjects) ListObjectsV2(bucket, prefix, continuationToken string, fetchOwner bool, delimiter string, maxKeys int) (result ListObjectsV2Info, err error) {
+	if strings.HasPrefix(prefix, "minio.sys.tmp/") {
```
harshavardhana: Isn't there a static global constant for this? `gcsMinioSysTmp` ? perhaps we can have a common `gatewayMinioSysTmp` ?

cmd/gateway-azure.go
```diff
@@ -477,38 +538,54 @@ func (a *azureObjects) DeleteObject(bucket, object string) error {
...
 	return result, nil
 }
 
+type azureMultipartMetadata struct {
+	Name     string            `json:"name"`
+	Metadata map[string]string `json:"metadata"`
+}
+
+func getAzureMetadataObjectName(objectName, uploadID string) string {
+	return fmt.Sprintf("minio.sys.tmp/multipart/v1/%s.%x/azure.json", uploadID, sha256.Sum256([]byte(objectName)))
```
harshavardhana: GCS seems to have constants for the prefix portion of the string here.. such as  `gcsMinioMultipartPathV1` can we do the same here?

harshavardhana (changes requested):

cmd/gateway-azure.go
balamurugana: Done

cmd/gateway-azure.go
balamurugana: Done

cmd/gateway-azure.go
balamurugana: Not a good idea. GCS and Azure may be in different version of handling metadata

krishnasrinivas: @balamurugana @harshavardhana is saying we can use:
`const azureMinioMultipartPathV1 gatewayMinioSysTmp + "/multipart/v1"`

i.e not `gcsMinioMultipartPathV1` itself

harshavardhana (approved):
LGTM

cmd/gateway-azure.go
```diff
@@ -304,12 +345,24 @@ func (a *azureObjects) ListObjects(bucket, prefix, marker, delimiter string, max
...
 		})
 	}
+
+	// Remove minio.sys.tmp prefix.
+	for i, prefix := range resp.BlobPrefixes {
+		if prefix == globalMinioSysTmpPrefix {
+			resp.BlobPrefixes = append(resp.BlobPrefixes[:i], resp.BlobPrefixes[i+1:]...)
+			break
+		}
+	}
```
krishnasrinivas: We need not blindly remove these entries. because it will help us debugging if we allow listing when we do:
`mc ls myazure/testbucket/minio.sys.tmp/`

i.e we should not filter out the entries if the user is specifically listing under the minio.sys.tmp prefix.

You can take a look at gateway-gcs.go ListObjects on how we take care of this

balamurugana: Done on appropriate section of the code.

balamurugana:
Tested and passed

PR #4877, opened 2017-08-31
[WIP]: Switch to non-root user for new Minio Docker deployments

nitisht:
## Description
Docker [recommends](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#user) running as non-root if the service can run without privileges.
With this PR, we switch to non-root (minio) user while running inside a Docker
container, if there isn't an existing Minio data directory or if its an empty
data directory.

We don't switch to minio user if there is an existing data directory, this is because
we'll need to chown all the files, which may take too long, ruining user's experience.

## Motivation and Context
See: https://github.com/minio/minio/issues/4381

## How Has This Been Tested?
Manually with Docker. Testing ongoing with Kubernetes/Docker Swarm

## Types of changes

## Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,13 +6,20 @@ ENV GOPATH /go
 ENV PATH $PATH:$GOPATH/bin
 ENV CGO_ENABLED 0
 
+COPY . /go/src/github.com/minio/minio
 WORKDIR /go/src/github.com/minio/
+
+# add our user and group first with fixed uid and gid, this ensures
+# multiple updates to the minio image dont have permission issues.
+RUN addgroup -g 190 -S minio && adduser -u 190 -S -G minio minio
+
+# grab su-exec for easy step-down from root
+RUN apk add --no-cache 'su-exec>=0.2'
 
 RUN  \
      apk add --no-cache ca-certificates && \
      apk add --no-cache --virtual .build-deps git go musl-dev && \
      echo 'hosts: files mdns4_minimal [NOTFOUND=return] dns mdns4' >> /etc/nsswitch.conf && \
-     go get -v -d github.com/minio/minio && \
      cd /go/src/github.com/minio/minio && \
      go install -v -ldflags "$(go run buildscripts/gen-ldflags.go)" && \
      rm -rf /go/pkg /go/src /usr/local/go && apk del .build-deps
--- a/buildscripts/docker-entrypoint.sh
+++ b/buildscripts/docker-entrypoint.sh
@@ -22,6 +22,78 @@ if [ "${1}" != "minio" ]; then
     fi
 fi
 
+# helper function to check if a string contains another string
+string_match () { 
+    case "$2" in *$1*)
+        return 0
+    esac
+    return 1
+}
+
+# switch to non root (minio) user if there isn't an existing Minio data directory or
+# if its an empty data directory. We dont switch to minio user if there is an existing
+# data directory, this is because we'll need to chown all the files -- which may take 
+# too long, ruining the user's experience.
+docker_switch_non_root() {
+    arguments_processed=0
+    # loop through all the arguements
+    for var in "$@"
+    do
+        echo "1"
+        # ignore minio, server, gateway or minio server flags
+        if [ "${var}" = "minio" ] || [ "${var}" = "server" ] \
+            || [ "${var}" = "gateway" ] || string_match '--*' "${var}" || \
+            string_match ':*' "${var}" ; then
+            # count as processed
+            arguments_processed=$((arguments_processed+1))
+            echo "2"
+            continue
+        fi
+        # strip scheme and hostname if present in an arg
+        if [ "${var:0:4}" = "http" ]; then
+            echo "3"
+            echo "${var}"
+            if ! string_match "$(hostname)" "${var}"; then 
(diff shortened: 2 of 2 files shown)


nitisht:
An update from current released Minio Docker image to a Docker image based on these changes fails on Kubernetes due to permission issues, looking into it. 

For whoever's interested, here are the commands to test the setup on Kubernetes:

1. Preconditions: an existing Minio distributed deployment running on Kubernetes (with root user inside Docker container). 

2. Use ` kubectl patch statefulset minio -p '{"spec":{"updateStrategy":{"type":"RollingUpdate"}}}'` to enable rolling updates. 

3. Use `kubectl patch statefulset minio --type='json' -p='[{"op": "replace", "path": "/spec/template/spec/containers/0/image", "value":"<image-with-current-changes>"}]'` to update the deployment to the new image. 

4. You should be able to access the pre-existing data on Minio deployment.

buildscripts/docker-entrypoint.sh
```diff
@@ -22,6 +22,69 @@ if [ "${1}" != "minio" ]; then
...
+    case "$2" in 
+        $1) return 0 ;; 
+        *) return 1 ;;
+    esac ; 
+}
+
+# switch to non root (minio) user if there isn't an existing Minio data directory or
+# if its an empty data directory. We dont switch to minio user if there is an existing
+# data directory, this is because we'll need to chown all the files -- which may take 
+# too long, ruining the user's experience.
```
harshavardhana, line 36: "which will take a long time, we do not want that."

buildscripts/docker-entrypoint.sh
```diff
@@ -22,6 +22,69 @@ if [ "${1}" != "minio" ]; then
...
+docker_switch_non_root() {
+    arguments_processed=0
+    # loop through all the arguements
```
harshavardhana, line 39: s/arguements/arguments/

buildscripts/docker-entrypoint.sh
```diff
@@ -22,6 +22,69 @@ if [ "${1}" != "minio" ]; then
...
+            # if the directory is empty
+            if [ "$(ls -A $var)" = "" ]; then
+                # change owner to minio user
+                chown -R minio:minio "${var}"
+            fi 
+            # find the owner of the directory
+            owner=$(ls -ld "${var}" | awk '{print $3}')
+            # if minio is the owner, count as processed
+            if [ "${owner}" = "minio" ];then
+                echo "previous owner = minio"
```
harshavardhana, line 75: All messages should be capitalized when printed and also should be a bit more descriptive of what they did rather than looking like a "return" statement.

buildscripts/docker-entrypoint.sh
```diff
@@ -22,6 +22,69 @@ if [ "${1}" != "minio" ]; then
...
+                arguments_processed=$((arguments_processed+1))
+            fi
+        else
+            echo "Created $var directory"
```
harshavardhana, line 79: Using `Created` before `mkdir` is not correct. We can use present continuous here..

buildscripts/docker-entrypoint.sh
```diff
@@ -39,5 +102,5 @@ docker_secrets_env() {
 
 ## Set access env from secrets if necessary.
 docker_secrets_env
-
-exec "$@"
+## Switch to non root user minio if applicable.
```
harshavardhana, line 114: Switch to non root user minio if possible.

Dockerfile
```diff
@@ -8,6 +8,13 @@ ENV CGO_ENABLED 0
 
 WORKDIR /go/src/github.com/minio/
 
+# add our user and group first with fixed uid and gid, this ensures
+# multiple updates to the minio image dont have permission issues.
+RUN addgroup -g 190 -S minio && adduser -u 190 -S -G minio minio
```
harshavardhana, line 14: why 190? you can perhaps set it to 1000

Dockerfile
```diff
@@ -8,6 +8,13 @@ ENV CGO_ENABLED 0
...
+
+# grab su-exec for easy step-down from root
+RUN apk add --no-cache 'su-exec>=0.2'
```
harshavardhana, line 17: Should this be in a new layer? why not add it to the next line below?

buildscripts/docker-entrypoint.sh
nitisht: This was added for debugging, I inadvertently committed. I don't intend to add echo messages, do you think we should have messages for the user here?

buildscripts/docker-entrypoint.sh
nitisht: This was added for debugging, I inadvertently committed. I don't intend to add echo messages, do you think we should have messages for the user here?

Dockerfile
nitisht: idea is to add a system account, so a user on the machine does not get access to Minio data directory unknowingly

Dockerfile
nitisht: yes, will add it as single layer

nitisht:
Closing this as looks like there are permission issues in getting this working on Kubernetes. Will revisit if needed later.

Issue #4821, opened 2017-08-16
minio startup with error message: Unable to access uploads.json, MultipartUpload()


lifematrix:
Hey, I am new to minio.
I just upload multiple large files to minio server for testing, and interrupt it by ctrl-C.
When I restart minio, it raise a  error message like:

```
time="2017-08-16T10:09:18+08:00" 
level=error 
msg="Unable to access uploads.json /workvol/home1/minio/repos/.minio.sys/multipart/.... /uploads.json" 
cause="file not found" 
source="[fs-v1-multipart.go:47:fsObjects.isMultipartUpload()]" 
stack="fs-v1-helpers.go:168:fsStatFile fs-v1-multipart.go:42:fsObjects.isMultipartUpload fs-v1-multipart.go:218:(fsObjects).(github.com/minio/minio/cmd.isMultipartUpload)-fm tree-walk.go:167:doTreeWalk tree-walk.go:238:startTreeWalk.func1 /home/builder/.gimme/versions/go1.7.5.linux.amd64/src/runtime/asm_amd64.s:2086:runtime.goexit"
```
I guess it is caused by incomplete uploads, and run the command

```
mc rm --incomplete --recursive --force   BUCKET_NAME
```
and restart minio again. However,  the error message still appear.

What  should I do to solve it?

Thanks for any suggestion.

Steven

harshavardhana:
This is a harmless message @lifematrix and can be ignored - we need to suppress and not print this and it is not useful for user to do anything at this point in time. 

It is because recursively some other routine has already deleted the necessary files. Just to be sure can you do `ls /workvol/home1/minio/repos/.minio.sys/multipart/*/*` and see if there are lots of files left over? 

Thanks

lifematrix:
Thanks for you quick reply.

I have deleted files  in ```/workvol/home1/minio/repos/.minio.sys/multipart/*/*``` recursively, and error message disappear.

By the way, I wonder if  a long-time uploads is interrupted or fail , what should do to prevent any problems?

Thanks

harshavardhana:
> By the way, I wonder if a long-time uploads is interrupted or fail , what should do to prevent any problems?
> 
> Thanks

It shouldn't there is an active routine which cleans up on a timely basis - it must be that there is a bug somewhere which leaves certain stale files do you know exactly how to reproduce this? 

Thanks @lifematrix

Julusian:
I have found that I am affected by this bug to a much more extreme level.
[...]

PR #4883, opened 2017-08-31
Suppress 'Unable to access uploads.json' error by fixing the incorrect if condition/comparison.

ebozduman:
Multipart upload throws an error message `Unable to access uploads.json` when, somehow, the uploads.json file cannot be found while there are still partial uploaded files lingering around. Although normally we should never end up in this situation and the scenario in which this could happen needs to be investigated and prevented, we still have a simple bug in the code, which when fixed, correctly suppresses the  error message, and makes this harmless condition invisible to the end user. 

## Description
The fix is in the if condition of `isMultipartUpload` function, where the code compares structure type var `err`  with error cause `errFileNotFound`. So, the fix is to replace var `err` with `errorCause(err)` to compare correct types. This fix automatically fixes the issue by returning false for `isMultipartUpload` function.

## Motivation and Context
Thrown `Unable to access uploads.json` error message when minio server starts is alarming to the user although this condition is harmless and not going to break anything.

https://github.com/minio/minio/issues/4821

## How Has This Been Tested?
Since we don't know the condition how we end up with no `uploads.json` file, I had to manually delete it and restart the minio server to reproduce the issue. To be able to see the error message right away, I also had to change the multipart clean up interval, `fsMultipartCleanupInterval`, and set it to 1 second modifying its regular value, 24 hours. Applying the fix and trying to reproduce the issue the same way again verified that the error message is not thrown/displayed anymore.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/fs-v1-multipart.go
+++ b/cmd/fs-v1-multipart.go
@@ -43,7 +43,7 @@ func (fs fsObjects) isMultipartUpload(bucket, prefix string) bool {
 	uploadsIDPath := pathJoin(fs.fsPath, bucket, prefix, uploadsJSONFile)
 	_, err := fsStatFile(uploadsIDPath)
 	if err != nil {
-		if err == errFileNotFound {
+		if errorCause(err) == errFileNotFound {
 			return false
 		}
 		errorIf(err, "Unable to access uploads.json "+uploadsIDPath)


harshavardhana (commented):
Thanks @ebozduman this fix is good can you fix the PR title to be a more cleaner descriptive rather than just 'Mutipart upload issue #4821' .

For a nicer introduction on commit messages you can read https://chris.beams.io/posts/git-commit/  - Thanks for @donatello for sharing this link.

krishnasrinivas (approved):

harshavardhana (approved):

Issue #4650, opened 2017-07-08
Implement server-side-encryption for all backend formats (FS, XL, ...)


aead:
Server-side-encryption can be implemented in the Object-Layer. This will lead to some overhead for XLs bitrot protection but simplifies encryption support for FS and other backends.

See aslo #4416 ,  #4020 , #4616 ,  #4649 

### Possible Solution
Server-side-encryption can be "easily" implemented in the Object-Layer. It only requires small changes to the backend formats and is transparent for erasure-coding / bitrot protection.P
Proposal:
 - use ChaCha20Poly1305 and/or AES-GCM as cipher.
 - store encrypted objects in slices/parts which are smaller than `N` and use an unique/random nonce per slice/nonce

@harshavardhana, @abperiasamy
We must think about the Impacts on other backends (RPC, GCS, Azure)...

harshavardhana:
GCS, Azure and AWS S3 backends can be a pass through utilizing the relevant server side encryption APIs from all these relevant providers.

aead:
Current progress:
#4894 Add basic support for:
  - GET (without ranges)
  - PUT
  - HEAD

The following operations are still missing:
 - GET with ranges - requires: part size and number of parts for multipart objects
 - InitMultipart
 - PUT Multipart - requires: XL / FS must save part size and number of parts in meta data
 - EndMultipart - requires: check that all uploaded parts have the same size - fail otherwise
 - Copy - Explore exact AWS behavior

aead:
basic work is done - I will open a separate issue for each feature

PR #4894, opened 2017-09-05
Add SSE-C support for basic operations (GET, PUT, HEAD)

aead:
This change adds server-side-encryption support for HEAD, GET and PUT
operations. This PR only addresses single-part PUTs and GETs without
HTTP ranges.

Further this change adds the concept of reserved object metadata which is required
to make encrypted objects tamper-proof and provide API compatibility to AWS S3.
This PR adds the following reserved metadata entries:
- X-Minio-Internal-Server-Side-Encryption-Iv          ('guarantees' tamper-proof property)
- X-Minio-Internal-Server-Side-Encryption-Kdf         (makes Key-MAC computation negotiable in future)
- X-Minio-Internal-Server-Side-Encryption-Key-Mac     (provides AWS S3 API compatibility)

The prefix `X-Minio_Internal` specifies an internal metadata entry which must not
send to clients. All client requests containing a metadata key starting with `X-Minio-Internal`
must also rejected. This is implemented by a generic-handler.

This PR implements SSE-C separated from client-side-encryption (CSE). This means clients
cannot decrypt server-side-encrypted objects on the client-side. However, clients can encrypted the same object with CSE and SSE-C.

This PR does not address:
 - SSE-C Copy and Copy part
 - SSE-C GET with HTTP ranges
 - SSE-C multipart PUT
 - SSE-C Gateway

Each point must be addressed in a separate PR.

Added to vendor dir:
 - x/crypto/chacha20poly1305
 - x/crypto/poly1305
 - github.com/minio/sio

Update #4650 #4416 

## How Has This Been Tested?
manually and against the AWS S3 API implementation

## Types of changes

## Checklist:

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -40,17 +40,19 @@ test_script:
   - mkdir build\coverage
   - go test -v -timeout 17m -race github.com/minio/minio/cmd...
   - go test -v -race github.com/minio/minio/pkg...
-  - go test -v -timeout 17m -coverprofile=build\coverage\coverage.txt -covermode=atomic github.com/minio/minio/cmd
+  # FIXME(aead): enable codecov after issue https://github.com/golang/go/issues/18468 is solved.
+  # - go test -v -timeout 17m -coverprofile=build\coverage\coverage.txt -covermode=atomic github.com/minio/minio/cmd
   - ps: Update-AppveyorTest "Unit Tests" -Outcome Passed
 
 after_test:
-  - go tool cover -html=build\coverage\coverage.txt -o build\coverage\coverage.html
-  - ps: Push-AppveyorArtifact build\coverage\coverage.txt
-  - ps: Push-AppveyorArtifact build\coverage\coverage.html
+  # FIXME(aead): enable codecov after issue https://github.com/golang/go/issues/18468 is solved.
+  # - go tool cover -html=build\coverage\coverage.txt -o build\coverage\coverage.html
+  # - ps: Push-AppveyorArtifact build\coverage\coverage.txt
+  # - ps: Push-AppveyorArtifact build\coverage\coverage.html
   # Upload coverage report.
-  - "SET PATH=C:\\Python34;C:\\Python34\\Scripts;%PATH%"
-  - pip install codecov
-  - codecov -X gcov -f "build\coverage\coverage.txt"
+  # - "SET PATH=C:\\Python34;C:\\Python34\\Scripts;%PATH%"
+  # - pip install codecov
+  # - codecov -X gcov -f "build\coverage\coverage.txt"
 
 # to disable deployment
 deploy: off
--- a/cmd/api-errors.go
+++ b/cmd/api-errors.go
@@ -121,6 +121,17 @@ const (
 	ErrUnsupportedMetadata
 	// Add new error codes here.
 
+	// Server-Side-Encryption (with Customer provided key) related API errors.
+
+	ErrInsecureSSECustomerRequest
+	ErrSSEEncryptedObject
+	ErrInvalidEncryptionParameters
+	ErrInvalidSSECustomerAlgorithm
+	ErrInvalidSSECustomerKey
+	ErrMissingSSECustomerKey
+	ErrMissingSSECustomerKeyMD5
+	ErrSSECustomerKeyMD5Mismatch
+
 	// Bucket notification related errors.
 	ErrEventNotification
 	ErrARNNotification
@@ -158,6 +169,7 @@ const (
 	ErrAdminConfigNoQuorum
 	ErrAdminCredentialsMismatch
 	ErrInsecureClientRequest
+	ErrObjectTampered
 )
 
 // error code to APIError structure, these fields carry respective
@@ -573,6 +585,51 @@ var errorCodeResponse = map[APIErrorCode]APIError{
 		Description:    "Range specified is not valid for source object",
 		HTTPStatusCode: http.StatusBadRequest,
 	},
+	ErrMetadataTooLarge: {
+		Code:           "InvalidArgument",
+		Description:    "Your metadata headers exceed the maximum allowed metadata size.",
+		HTTPStatusCode: http.StatusBadRequest,
+	},
+	ErrInsecureSSECustomerRequest: {
+		Code:           "InvalidRequest",
+		Description:    errInsecureSSERequest.Error(),
+		HTTPStatusCode: http.StatusBadRequest,
+	},
+	ErrSSEEncryptedObject: {
+		Code:           "InvalidRequest",
+		Description:    errEncryptedObject.Error(),
+		HTTPStatusCode: http.StatusBadRequest,
+	},
+	ErrInvalidEncryptionParameters: {
+		Code:           "InvalidRequest",
+		Description:    "The encryption parameters are not applicable to this object.",
+		HTTPStatusCode: http.StatusBadRequest,
+	},
+	ErrInvalidSSECustomerAlgorithm: {
+		Code:           "InvalidArgument",
+		Description:    errInvalidSSEAlgorithm.Error(),
+		HTTPStatusCode: http.StatusBadRequest,
+	},
+	ErrInvalidSSECustomerKey: {
+		Code:           "InvalidArgument",
+		Description:    errInvalidSSEKey.Error(),
+		HTTPStatusCode: http.StatusBadRequest,
+	},
+	ErrMissingSSECustomerKey: {
+		Code:           "InvalidArgument",
+		Description:    errMissingSSEKey.Error(),
+		HTTPStatusCode: http.StatusBadRequest,
+	},
+	ErrMissingSSECustomerKeyMD5: {
+		Code:           "InvalidArgument",
+		Description:    errMissingSSEKeyMD5.Error(),
+		HTTPStatusCode: http.StatusBadRequest,
+	},
+	ErrSSECustomerKeyMD5Mismatch: {
+		Code:           "InvalidArgument",
+		Description:    errSSEKeyMD5Mismatch.Error(),
+		HTTPStatusCode: http.StatusBadRequest,
+	},
 
 	/// S3 extensions.
 	ErrContentSHA256Mismatch: {
@@ -652,11 +709,6 @@ var errorCodeResponse = map[APIErrorCode]APIError{
 		Description:    "A timeout occurred while trying to lock a resource",
 		HTTPStatusCode: http.StatusRequestTimeout,
 	},
-	ErrMetadataTooLarge: {
-		Code:           "InvalidArgument",
-		Description:    "Your metadata headers exceed the maximum allowed metadata size.",
-		HTTPStatusCode: http.StatusBadRequest,
-	},
 	ErrUnsupportedMetadata: {
 		Code:           "InvalidArgument",
 		Description:    "Your metadata headers are not supported.",
@@ -666,6 +718,11 @@ var errorCodeResponse = map[APIErrorCode]APIError{
 		Code:           "XMinioPartsSizeUnequal",
 		Description:    "All parts except the last part should be of the same size.",
 		HTTPStatusCode: http.StatusBadRequest,
+	},
+	ErrObjectTampered: {
+		Code:           "XMinioObjectTampered",
+		Description:    errObjectTampered.Error(),
+		HTTPStatusCode: http.StatusPartialContent,
 	},
 	// Add your error structure here.
 }
@@ -696,6 +753,27 @@ func toAPIErrorCode(err error) (apiErr APIErrorCode) {
 	if apiErr != ErrNone {
 		// If there was a match in the above switch case.
 		return apiErr
+	}
+
+	switch err { // SSE errors
+	case errInsecureSSERequest:
+		return ErrInsecureSSECustomerRequest
+	case errInvalidSSEAlgorithm:
+		return ErrInvalidSSECustomerAlgorithm
+	case errInvalidSSEKey:
+		return ErrInvalidSSECustomerKey
+	case errMissingSSEKey:
+		return ErrMissingSSECustomerKey
+	case errMissingSSEKeyMD5:
+		return ErrMissingSSECustomerKeyMD5
+	case errSSEKeyMD5Mismatch:
+		return ErrSSECustomerKeyMD5Mismatch
+	case errObjectTampered:
+		return ErrObjectTampered
+	case errEncryptedObject:
+		return ErrSSEEncryptedObject
+	case errSSEKeyMismatch:
+		return ErrAccessDenied // no access without correct key
 	}
 
 	switch err.(type) {
--- a/cmd/encryption-v1.go
+++ b/cmd/encryption-v1.go
@@ -1,0 +1,300 @@
+/*
+ * Minio Cloud Storage, (C) 2017 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cmd
+
+import (
+	"bytes"
+	"crypto/hmac"
+	"crypto/md5"
+	"crypto/rand"
+	"encoding/base64"
+	"errors"
+	"io"
+	"net/http"
+
+	sha256 "github.com/minio/sha256-simd"
+	"github.com/minio/sio"
+)
+
+var (
+	// AWS errors for invalid SSE-C requests.
+	errInsecureSSERequest  = errors.New("Requests specifying Server Side Encryption with Customer provided keys must be made over a secure connection")
+	errEncryptedObject     = errors.New("The object was stored using a form of Server Side Encryption. The correct parameters must be provided to retrieve the object")
+	errInvalidSSEAlgorithm = errors.New("Requests specifying Server Side Encryption with Customer provided keys must provide a valid encryption algorithm")
+	errMissingSSEKey       = errors.New("Requests specifying Server Side Encryption with Customer provided keys must provide an appropriate secret key")
+	errInvalidSSEKey       = errors.New("The secret key was invalid for the specified algorithm")
+	errMissingSSEKeyMD5    = errors.New("Requests specifying Server Side Encryption with Customer provided keys must provide the client calculated MD5 of the secret key")
+	errSSEKeyMD5Mismatch   = errors.New("The calculated MD5 hash of the key did not match the hash that was provided")
+	errSSEKeyMismatch      = errors.New("The client provided key does not match the key provided when the object was encrypted") // this msg is not shown to the client
+
+	// Additional Minio errors for SSE-C requests.
+	errObjectTampered = errors.New("The requested object was modified and may be compromised")
+)
+
+const (
+	// SSECustomerAlgorithm is the AWS SSE-C algorithm HTTP header key.
+	SSECustomerAlgorithm = "X-Amz-Server-Side-Encryption-Customer-Algorithm"
+	// SSECustomerKey is the AWS SSE-C encryption key HTTP header key.
+	SSECustomerKey = "X-Amz-Server-Side-Encryption-Customer-Key"
+	// SSECustomerKeyMD5 is the AWS SSE-C encryption key MD5 HTTP header key.
+	SSECustomerKeyMD5 = "X-Amz-Server-Side-Encryption-Customer-Key-MD5"
+)
+
+const (
+	// SSECustomerKeySize is the size of valid client provided encryption keys in bytes.
+	// Currently AWS supports only AES256. So the SSE-C key size is fixed to 32 bytes.
+	SSECustomerKeySize = 32
+
+	// SSECustomerAlgorithmAES256 the only valid S3 SSE-C encryption algorithm identifier.
+	SSECustomerAlgorithmAES256 = "AES256"
+)
+
+// SSE-C key derivation:
+// H: Hash function, M: MAC function
+//
+// key     := 32 bytes              # client provided key
+// r       := H(random(32 bytes))   # saved as object metadata [ServerSideEncryptionIV]
+// key_mac := M(H(key), r)          # saved as object metadata [ServerSideEncryptionKeyMAC]
+// enc_key := M(key, key_mac)
+//
+//
+// SSE-C key verification:
+// H: Hash function, M: MAC function
+//
+// key      := 32 bytes             # client provided key
+// r        := object metadata [ServerSideEncryptionIV]
+// key_mac  := object metadata [ServerSideEncryptionKeyMAC]
+// key_mac' := M(H(key), r)
+//
+// check:   key_mac != key_mac' => fail with invalid key
+//
+// enc_key  := M(key, key_mac')
+const (
+	// ServerSideEncryptionIV is a 32 byte randomly generated IV used to derive an
+	// unique encryption key from the client provided key. The combination of this value
+	// and the client-provided key must be unique to provide the DARE tamper-proof property.
+	ServerSideEncryptionIV = ReservedMetadataPrefix + "Server-Side-Encryption-Iv"
+
+	// ServerSideEncryptionKDF is the combination of a hash and MAC function used to derive
+	// the SSE-C encryption key from the user-provided key.
+	ServerSideEncryptionKDF = ReservedMetadataPrefix + "Server-Side-Encryption-Kdf"
+
+	// ServerSideEncryptionKeyMAC is the MAC of the hash of the client-provided key and the
+	// X-Minio-Server-Side-Encryption-Iv. This value must be used to verify that the client
+	// provided the correct key to follow S3 spec.
+	ServerSideEncryptionKeyMAC = ReservedMetadataPrefix + "Server-Side-Encryption-Key-Mac"
+)
+
+// SSEKeyDerivationHmacSha256 specifies SHA-256 as hash function and HMAC-SHA256 as MAC function
+// as the functions used to derive the SSE-C encryption keys from the client-provided key.
+const SSEKeyDerivationHmacSha256 = "HMAC-SHA256"
+
+// IsSSECustomerRequest returns true if the given HTTP header
+// contains server-side-encryption with customer provided key fields.
+func IsSSECustomerRequest(header http.Header) bool {
+	return header.Get(SSECustomerAlgorithm) != "" || header.Get(SSECustomerKey) != "" || header.Get(SSECustomerKeyMD5) != ""
+}
+
+// ParseSSECustomerRequest parses the SSE-C header fields of the provided request.
+// It returns the client provided key on success.
+func ParseSSECustomerRequest(r *http.Request) (key []byte, err error) {
+	if !globalIsSSL { // minio only supports HTTP or HTTPS requests not both at the same time
+		// we cannot use r.TLS == nil here because Go's http implementation reflects on
+		// the net.Conn and sets the TLS field of http.Request only if it's an tls.Conn.
+		// Minio uses a BufConn (wrapping a tls.Conn) so the type check within the http package
+		// will always fail -> r.TLS is always nil even for TLS requests.
+		return nil, errInsecureSSERequest
+	}
+	header := r.Header
+	if algorithm := header.Get(SSECustomerAlgorithm); algorithm != SSECustomerAlgorithmAES256 {
+		return nil, errInvalidSSEAlgorithm
+	}
+	if header.Get(SSECustomerKey) == "" {
+		return nil, errMissingSSEKey
+	}
+	if header.Get(SSECustomerKeyMD5) == "" {
+		return nil, errMissingSSEKeyMD5
+	}
+
+	key, err = base64.StdEncoding.DecodeString(header.Get(SSECustomerKey))
+	if err != nil {
+		return nil, errInvalidSSEKey
+	}
+	header.Del(SSECustomerKey) // make sure we do not save the key by accident
+
+	if len(key) != SSECustomerKeySize {
+		return nil, errInvalidSSEKey
+	}
+
+	keyMD5, err := base64.StdEncoding.DecodeString(header.Get(SSECustomerKeyMD5))
+	if err != nil {
+		return nil, errSSEKeyMD5Mismatch
+	}
+	if md5Sum := md5.Sum(key); !bytes.Equal(md5Sum[:], keyMD5) {
+		return nil, errSSEKeyMD5Mismatch
+	}
+	return key, nil
+}
+
+// EncryptRequest takes the client provided content and encrypts the data
+// with the client provided key. It also marks the object as client-side-encrypted
+// and sets the correct headers.
+func EncryptRequest(content io.Reader, r *http.Request, metadata map[string]string) (io.Reader, error) {
+	key, err := ParseSSECustomerRequest(r)
+	if err != nil {
+		return nil, err
+	}
+	delete(metadata, SSECustomerKey) // make sure we do not save the key by accident
+
+	// security notice:
+	// Reusing a tuple (nonce, client provided key) will produce the same encryption key
+	// twice and breaks the tamper-proof property. However objects are still confidential.
+	// Therefore the nonce must be unique but need not to be undistinguishable from true
+	// randomness.
+	nonce := make([]byte, 32) // generate random nonce to derive encryption key
+	if _, err = io.ReadFull(rand.Reader, nonce); err != nil {
+		return nil, err
+	}
+	iv := sha256.Sum256(nonce) // hash output to not reveal any stat. weaknesses of the PRNG
+
+	keyHash := sha256.Sum256(key) // derive MAC of the client-provided key
+	mac := hmac.New(sha256.New, keyHash[:])
+	mac.Write(iv[:])
+	keyMAC := mac.Sum(nil)
+
+	mac = hmac.New(sha256.New, key) // derive encryption key
+	mac.Write(keyMAC)
+	reader, err := sio.EncryptReader(content, sio.Config{Key: mac.Sum(nil)})
+	if err != nil {
+		return nil, errInvalidSSEKey
+	}
+
+	metadata[ServerSideEncryptionIV] = base64.StdEncoding.EncodeToString(iv[:])
+	metadata[ServerSideEncryptionKDF] = SSEKeyDerivationHmacSha256
+	metadata[ServerSideEncryptionKeyMAC] = base64.StdEncoding.EncodeToString(keyMAC)
+	return reader, nil
+}
+
+// DecryptRequest decrypts the object with the client provided key. It also removes
+// the client-side-encryption metadata from the object and sets the correct headers.
+func DecryptRequest(client io.Writer, r *http.Request, metadata map[string]string) (io.WriteCloser, error) {
+	key, err := ParseSSECustomerRequest(r)
+	if err != nil {
+		return nil, err
+	}
+	delete(metadata, SSECustomerKey) // make sure we do not save the key by accident
+
+	if metadata[ServerSideEncryptionKDF] != SSEKeyDerivationHmacSha256 { // currently HMAC-SHA256 is the only option
+		return nil, errObjectTampered
+	}
+	nonce, err := base64.StdEncoding.DecodeString(metadata[ServerSideEncryptionIV])
+	if err != nil || len(nonce) != 32 {
+		return nil, errObjectTampered
+	}
+	keyMAC, err := base64.StdEncoding.DecodeString(metadata[ServerSideEncryptionKeyMAC])
+	if err != nil || len(keyMAC) != 32 {
+		return nil, errObjectTampered
+	}
+
+	keyHash := sha256.Sum256(key) // verify that client provided correct key
+	mac := hmac.New(sha256.New, keyHash[:])
+	mac.Write(nonce)
+	if !hmac.Equal(keyMAC, mac.Sum(nil)) {
+		return nil, errSSEKeyMismatch // client-provided key is wrong or object metadata was modified
+	}
+
+	mac = hmac.New(sha256.New, key) // derive decryption key
+	mac.Write(keyMAC)
+	writer, err := sio.DecryptWriter(client, sio.Config{Key: mac.Sum(nil)})
+	if err != nil {
+		return nil, errInvalidSSEKey
+	}
+
+	delete(metadata, ServerSideEncryptionIV)
+	delete(metadata, ServerSideEncryptionKDF)
+	delete(metadata, ServerSideEncryptionKeyMAC)
+	return writer, nil
+}
+
+// IsEncrypted returns true if the object is marked as encrypted.
+func (o *ObjectInfo) IsEncrypted() bool {
+	if _, ok := o.UserDefined[ServerSideEncryptionIV]; ok {
+		return true
+	}
+	if _, ok := o.UserDefined[ServerSideEncryptionKDF]; ok {
+		return true
+	}
+	if _, ok := o.UserDefined[ServerSideEncryptionKeyMAC]; ok {
+		return true
+	}
+	return false
+}
+
+// DecryptedSize returns the size of the object after decryption in bytes.
+// It returns an error if the object is not encrypted or marked as encrypted
+// but has an invalid size.
+// DecryptedSize panics if the referred object is not encrypted.
+func (o *ObjectInfo) DecryptedSize() (int64, error) {
+	if !o.IsEncrypted() {
+		panic("cannot compute decrypted size of an object which is not encrypted")
+	}
+	if o.Size == 0 {
+		return o.Size, nil
+	}
+	size := (o.Size / (32 + 64*1024)) * (64 * 1024)
+	if mod := o.Size % (32 + 64*1024); mod > 0 {
+		if mod < 33 {
+			return -1, errObjectTampered // object is not 0 size but smaller than the smallest valid encrypted object
+		}
+		size += mod - 32
+	}
+	return size, nil
+}
+
+// EncryptedSize returns the size of the object after encryption.
+// An encrypted object is always larger than a plain object
+// except for zero size objects.
+func (o *ObjectInfo) EncryptedSize() int64 {
+	size := (o.Size / (64 * 1024)) * (32 + 64*1024)
+	if mod := o.Size % (64 * 1024); mod > 0 {
+		size += mod + 32
+	}
+	return size
+}
+
+// DecryptObjectInfo tries to decrypt the provided object if it is encrypted.
+// It fails if the object is encrypted and the HTTP headers don't contain
+// SSE-C headers or the object is not encrypted but SSE-C headers are provided. (AWS behavior)
+// DecryptObjectInfo returns 'ErrNone' if the object is not encrypted or the
+// decryption succeeded.
+//
+// DecryptObjectInfo also returns whether the object is encrypted or not.
+func DecryptObjectInfo(info *ObjectInfo, headers http.Header) (apiErr APIErrorCode, encrypted bool) {
+	if apiErr, encrypted = ErrNone, info.IsEncrypted(); !encrypted && IsSSECustomerRequest(headers) {
+		apiErr = ErrInvalidEncryptionParameters
+	} else if encrypted {
+		if !IsSSECustomerRequest(headers) {
+			apiErr = ErrSSEEncryptedObject
+			return
+		}
+		var err error
+		if info.Size, err = info.DecryptedSize(); err != nil {
+			apiErr = toAPIErrorCode(err)
+		}
+	}
+	return
+}
--- a/cmd/gateway-gcs.go
+++ b/cmd/gateway-gcs.go
@@ -872,7 +872,6 @@ func (l *gcsGateway) PutObjectPart(bucket string, key string, uploadID string, p
 	}
 	// Make sure to close the object writer upon success.
 	w.Close()
-
 	return PartInfo{
 		PartNumber:   partNumber,
 		ETag:         etag,
--- a/cmd/gateway-handlers.go
+++ b/cmd/gateway-handlers.go
@@ -18,7 +18,7 @@ package cmd
 
 import (
 	"io"
-	"io/ioutil"
+	goioutil "io/ioutil"
 	"net"
 	"net/http"
 	"strconv"
@@ -29,6 +29,7 @@ import (
 	router "github.com/gorilla/mux"
 	"github.com/minio/minio-go/pkg/policy"
 	"github.com/minio/minio/pkg/hash"
+	"github.com/minio/minio/pkg/ioutil"
 )
 
 // GetObjectHandler - GET Object
@@ -118,32 +119,19 @@ func (api gatewayAPIHandlers) GetObjectHandler(w http.ResponseWriter, r *http.Re
 		startOffset = hrange.offsetBegin
 		length = hrange.getLength()
 	}
-	// Indicates if any data was written to the http.ResponseWriter
-	dataWritten := false
-	// io.Writer type which keeps track if any data was written.
-	writer := funcToWriter(func(p []byte) (int, error) {
-		if !dataWritten {
-			// Set headers on the first write.
-			// Set standard object headers.
-			setObjectHeaders(w, objInfo, hrange)
-
-			// Set any additional requested response headers.
-			setHeadGetRespHeaders(w, r.URL.Query())
-
-			dataWritten = true
-		}
-		return w.Write(p)
-	})
 
 	getObject := objectAPI.GetObject
 	if reqAuthType == authTypeAnonymous {
 		getObject = objectAPI.AnonGetObject
 	}
 
+	setObjectHeaders(w, objInfo, hrange)
+	setHeadGetRespHeaders(w, r.URL.Query())
+	httpWriter := ioutil.WriteOnClose(w)
 	// Reads the object at startOffset and writes to mw.
-	if err = getObject(bucket, object, startOffset, length, writer); err != nil {
+	if err = getObject(bucket, object, startOffset, length, httpWriter); err != nil {
 		errorIf(err, "Unable to write to client.")
-		if !dataWritten {
+		if !httpWriter.HasWritten() {
 			// Error response only if no data has been written to client yet. i.e if
 			// partial data has already been written before an error
 			// occurred then no point in setting StatusCode and
@@ -152,11 +140,11 @@ func (api gatewayAPIHandlers) GetObjectHandler(w http.ResponseWriter, r *http.Re
 		}
 		return
 	}
-	if !dataWritten {
-		// If ObjectAPI.GetObject did not return error and no data has
-		// been written it would mean that it is a 0-byte object.
-		// call wrter.Write(nil) to set appropriate headers.
-		writer.Write(nil)
+	if err = httpWriter.Close(); err != nil {
+		if !httpWriter.HasWritten() { // write error response only if no data has been written to client yet
+			writeErrorResponse(w, toAPIErrorCode(err), r.URL)
+			return
+		}
 	}
 
 	// Get host and port from Request.RemoteAddr.
@@ -472,7 +460,7 @@ func (api gatewayAPIHandlers) PutBucketPolicyHandler(w http.ResponseWriter, r *h
 	// Read access policy up to maxAccessPolicySize.
 	// http://docs.aws.amazon.com/AmazonS3/latest/dev/access-policy-language-overview.html
 	// bucket policies are limited to 20KB in size, using a limit reader.
-	policyBytes, err := ioutil.ReadAll(io.LimitReader(r.Body, maxAccessPolicySize))
+	policyBytes, err := goioutil.ReadAll(io.LimitReader(r.Body, maxAccessPolicySize))
 	if err != nil {
 		errorIf(err, "Unable to read from client.")
 		writeErrorResponse(w, toAPIErrorCode(err), r.URL)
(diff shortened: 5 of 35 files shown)


harshavardhana:
:cake:

cmd/encryption.go
```diff
@@ -0,0 +1,115 @@
+package cmd
```
fwessels: Add license header

cmd/encryption.go
```diff
@@ -0,0 +1,115 @@
...
+	errInvalidSSEKey       = errors.New("The secret key was invalid for the specified algorithm")
+	errMissingSSEKeyMD5    = errors.New("Requests specifying Server Side Encryption with Customer provided keys must provide the client calculated MD5 of the secret key")
+	errSSEKeyMD5Mismatch   = errors.New("The calculated MD5 hash of the key did not match the hash that was provided")
+)
+
+func isSSERequest(header http.Header) bool {
+	if header.Get("X-Amz-Server-Side-Encryption-Customer-Algorithm") != "" {
+		return true
+	}
+	if header.Get("X-Amz-Server-Side-Encryption-Customer-Key") != "" {
```
fwessels: Maybe combine into a single `if` ?

cmd/encryption.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+	return false
+}
+
+func parseSSERequest(header http.Header) ([]byte, error) {
+	algorithm := header.Get("X-Amz-Server-Side-Encryption-Customer-Algorithm")
+	if algorithm != "AES256" {
+		return nil, errInvalidSSEAlgorithm
+	}
+	key, err := base64.StdEncoding.DecodeString(header.Get("X-Amz-Server-Side-Encryption-Customer-Key"))
```
fwessels: Use const definition for header value (repeated multiple times)

cmd/encryption.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+func (sse sseObject) GetObjectInfo(bucket string, object string) (ObjectInfo, error) {
+	info, err := sse.ObjectLayer.GetObjectInfo(bucket, object)
+	if err != nil {
+		return ObjectInfo{}, err
+	}
+	size := (info.Size / (32 + 64*1024)) * (64 * 1024)
+	if mod := info.Size % (32 + 64*1024); mod > 0 {
+		size += mod - 32
```
fwessels: Can mod ever be smaller than 32? Then you would effectively decrease the size...

cmd/object-api-utils.go
```diff
@@ -269,27 +268,3 @@ type byBucketName []BucketInfo
...
-	Max    int64     // max bytes remaining
-}
-
-func (l *rangeReader) Read(p []byte) (n int, err error) {
-	n, err = l.Reader.Read(p)
-	l.Max -= int64(n)
-	l.Min -= int64(n)
-	if l.Max < 0 {
-		// If more data is available than what is expected we return error.
-		return 0, errDataTooLarge
```
harshavardhana: Removal of this struct and implementation send this cleanup as PR separately, not as part of this PR.

cmd/encryption.go
```diff
@@ -0,0 +1,186 @@
...
+	SSECustomerKeyMD5    = "X-Amz-Server-Side-Encryption-Customer-Key-MD5"
+)
+
+// SSECustomerKeySize is the size of valid client provided encryption keys in bytes.
+// Currently AWS supports only AES256. So the SSE-C key size is fixed to 32 bytes.
+const SSECustomerKeySize = 32
+
+// IsSSECustomerRequest returns true if the given HTTP header
+// contains server-side-encryption with customer provided key fields.
+func IsSSECustomerRequest(header http.Header) bool {
```
harshavardhana: You could add unit tests for this function..

cmd/encryption.go
```diff
@@ -0,0 +1,186 @@
...
+	if header.Get(SSECustomerAlgorithm) != "" {
+		return true
+	}
+	if header.Get(SSECustomerKey) != "" {
+		return true
+	}
+	if header.Get(SSECustomerKeyMD5) != "" {
```
harshavardhana: Collapse this `if` into simply. 

``` return header.Get(SSECustomerKeyMD5) != "" ```

File: cmd/encryption.go [hunk 8]
```diff
@@ -0,0 +1,186 @@
...
+		return true
+	}
+	return false
+}
+
+// ParseSSECustomerRequest parses the SSE-C header fields of the provided header.
+// It returns the client provided key on success.
+func ParseSSECustomerRequest(header http.Header) (key []byte, err error) {
```
harshavardhana: Needs a unit test.

cmd/encryption.go
```diff
@@ -0,0 +1,186 @@
...
+		return nil, errSSEKeyMD5Mismatch
+	}
+	return key, nil
+}
+
+// RangeWriter returns an io.Writer which ignores all data written
+// to it before start and all data written to it after end.
+func RangeWriter(w io.Writer, start, end int64) io.Writer {
+	if start < 0 {
+		panic("start cannot be negative")
```
harshavardhana: We shouldn't panic in our server code only under dire situations, especially for argument validation we should simply return errInvalidArgument in this case.

cmd/encryption.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+	if end < 0 {
+		panic("end cannot be negative")
+	}
+	if start > end {
+		panic("start cannot greater than end")
+	}
+	return &rangeWriter{dst: w, start: start, end: end}
+}
+
```
harshavardhana: Needs comments and unit tests..

cmd/encryption.go
```diff
@@ -0,0 +1,186 @@
+package cmd
```
harshavardhana: Rename this file as just `encryption-v1.go` this just follows the current convention and also indicates that this is the object layer implementation.

cmd/encryption.go
```diff
@@ -0,0 +1,186 @@
...
+		}
+		p = p[w.end:]
+		w.end = 0
+		n += nn
+	}
+
+	return n + len(p), err
+}
+
+type sseObject struct {
```
harshavardhana: I guess we don't need to say sseObject instead we can simply say `encryptedObject` since the file is `encryption.go` ?

cmd/object-handlers.go
```diff
@@ -176,6 +187,30 @@ func (api objectAPIHandlers) GetObjectHandler(w http.ResponseWriter, r *http.Req
 		return w.Write(p)
 	})
 
+	if IsSSECustomerRequest(r.Header) {
+		size := (objInfo.Size / (64 * 1024)) * (32 + 64*1024)
```
harshavardhana: Can the following block move into a function? and add a unit test?

cmd/encryption.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+// ParseSSECustomerRequest parses the SSE-C header fields of the provided request.
+// It returns the client provided key on success.
+func ParseSSECustomerRequest(r *http.Request) (key []byte, err error) {
+	if r.TLS == nil {
+		return nil, errInsecureSSERequest
+	}
+	header := r.Header
+	if algorithm := header.Get(SSECustomerAlgorithm); algorithm != "AES256" { // Currently AWS supports only AES256
```
fwessels: Maybe add a define/const for AES256?

cmd/encryption.go
```diff
@@ -0,0 +1,121 @@
...
+	if err != nil {
+		return nil, errInvalidSSEKey
+	}
+	if len(key) != SSECustomerKeySize {
+		return nil, errInvalidSSEKey
+	}
+
+	keyMD5, err := base64.StdEncoding.DecodeString(header.Get(SSECustomerKeyMD5))
+	if err != nil {
+		return nil, errSSEKeyMD5Mismatch
```
fwessels: Technically this may not be a mismatch, maybe report a different error?

cmd/encryption.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+	if md5Sum := md5.Sum(key); !bytes.Equal(md5Sum[:], keyMD5) {
+		return nil, errSSEKeyMD5Mismatch
+	}
+	return key, nil
+}
+
+// DecryptSSECustomerObjectInfo converts an object info of an SSE-C encrypted object into
+// an plain object info. It returns an error if the object info is not a valid encrypted
```
fwessels: an plain --> a plain
a valid encrypted --> a validly encrypted

cmd/encryption.go
```diff
@@ -0,0 +1,121 @@
...
+// object info.
+//
+// DecryptSSECustomerObjectInfo computes the real object size of an encrypted object.
+func DecryptSSECustomerObjectInfo(info ObjectInfo) (ObjectInfo, error) {
+	size := (info.Size / (32 + 64*1024)) * (64 * 1024)
```
fwessels: Add defines for 32 (ie `SioFormatOverhead`) and 64K (`SioPacketSize`) ?

cmd/encryption.go
```diff
@@ -0,0 +1,121 @@
...
+		if mod < 33 {
+			return info, errObjectTampered
+		}
+		size += mod - 32
+	}
+	info.Size = size
+	return info, nil
+}
+
+// EncryptSSECustomerObjectInfo converts an object info of plain object into into an
```
fwessels: into repeated twice.

cmd/encryption.go
```diff
@@ -0,0 +1,121 @@
...
+// encrypted object info.
+//
+// EncryptSSECustomerObjectInfo computes the encrypted object size of an plain object.
+func EncryptSSECustomerObjectInfo(info ObjectInfo) ObjectInfo {
+	size := (info.Size / (64 * 1024)) * (32 + 64*1024)
+	if mod := info.Size % (64 * 1024); mod > 0 {
```
fwessels: Same for defines for 32 and 64K

cmd/encryption.go
aead: In general yes - but we may move to a variable packet size approach: See minio/sio#9

cmd/encryption.go
aead: That's AWS behavior.

cmd/gateway-azure.go
```diff
@@ -580,33 +520,20 @@ func (a *azureObjects) PutObjectPart(bucket, object, uploadID string, partID int
...
-		}
-	}
-
-	if sha256sum != "" {
-		newSHA256sum := hex.EncodeToString(sha256Writer.Sum(nil))
-		if newSHA256sum != sha256sum {
-			return PartInfo{}, azureToObjectError(traceError(SHA256Mismatch{}))
-		}
+	if err = data.Verify(); err != nil {
+		return info, azureToObjectError(traceError(err), bucket, object)
```
fwessels: Should we remove the blob? (`DeleteBlob()` called in line 594 in old code but not in line 602)

cmd/gateway-gcs.go
```diff
@@ -747,30 +738,17 @@ func (l *gcsGateway) PutObject(bucket string, key string, size int64, data io.Re
...
 	w.Close()
 
-	// Verify sha256sum after close.
-	if sha256sum != "" {
-		if hex.EncodeToString(sha256Writer.Sum(nil)) != sha256sum {
-			object.Delete(l.ctx)
-			return ObjectInfo{}, traceError(SHA256Mismatch{})
-		}
+	if err := data.Verify(); err != nil { // Verify sha256sum after close.
+		return ObjectInfo{}, gcsToObjectError(traceError(err), bucket, key)
```
fwessels: Line 771 in old code calls `object.Delete()` -- should we do that here too?

cmd/object-handlers.go
```diff
@@ -88,6 +90,43 @@ func (f funcToWriter) Write(p []byte) (int, error) {
 	return f(p)
 }
 
+// HTTPWriteCloser writes to the underlying io.Writer
+// and also writes nil to the underlying io.Writer when
+// Close is called and no data was written.
+//
+// In case of a http.ResponseWriter the Close will set the
```
fwessels: a --> an

cmd/object-handlers.go
```diff
@@ -88,6 +90,43 @@ func (f funcToWriter) Write(p []byte) (int, error) {
...
+// HTTP headers depending when no data was written to the client.
```
fwessels: depending --> depending upon

cmd/object-handlers.go
```diff
@@ -88,6 +90,43 @@ func (f funcToWriter) Write(p []byte) (int, error) {
...
+type HTTPWriteCloser struct {
+	io.Writer
+	hasWritten bool
+}
+
+// Write writes data to the HTTPWriteCloser.
+func (w *HTTPWriteCloser) Write(p []byte) (n int, err error) {
+	if w.hasWritten {
```
fwessels: condition needs to be inverted. Or alternatively, always set it to `true`, remove if statement altogether.

cmd/object-handlers.go
```diff
@@ -88,6 +90,43 @@ func (f funcToWriter) Write(p []byte) (int, error) {
...
+		w.hasWritten = true
+	}
+	return w.Writer.Write(p)
+}
+
+// HasWritten returns true if data was written to the HTTPWriteCloser.
```
fwessels: to the --> to by the

cmd/object-handlers.go
```diff
@@ -88,6 +90,43 @@ func (f funcToWriter) Write(p []byte) (int, error) {
...
+func (w *HTTPWriteCloser) HasWritten() bool { return w.hasWritten }
+
+// Close closes the HTTPWriteCloser and sets the HTTP headers if the underlying
+// Writer is a http.ResponseWriter and nothing was written to it.
```
fwessels: a --> an

cmd/object-handlers.go
```diff
@@ -128,6 +167,14 @@ func (api objectAPIHandlers) GetObjectHandler(w http.ResponseWriter, r *http.Req
 		writeErrorResponse(w, apiErr, r.URL)
 		return
 	}
+	if IsSSECustomerRequest(r.Header) {
+		objInfo, err = DecryptSSECustomerObjectInfo(objInfo)
+		if err != nil {
+			// object seems to be smaller than a valid encrypt object - fail with custom API error
```
fwessels: than a valid encrypt object --> than a validly encrypted object

cmd/object-handlers.go
```diff
@@ -251,6 +301,14 @@ func (api objectAPIHandlers) HeadObjectHandler(w http.ResponseWriter, r *http.Re
 		writeErrorResponseHeadersOnly(w, apiErr)
 		return
 	}
+	if IsSSECustomerRequest(r.Header) {
+		objInfo, err = DecryptSSECustomerObjectInfo(objInfo)
+		if err != nil {
+			// object seems to be smaller than a valid encrypt object - fail with custom API error
```
fwessels: a valid encrypt object --> a validly encrypted object

fwessels:
:closed_lock_with_key:
Couple of remarks, code looks good in general

harshavardhana:
re-opening :-)

cmd/object-handlers.go
aead: done

cmd/object-handlers.go
aead: done

cmd/object-handlers.go
aead: Same, really?!

cmd/object-handlers.go
aead: done, thanks :+1:

cmd/object-handlers.go
aead: done

cmd/object-handlers.go
aead: Really? Yes it sounds better but I thought 'an' just before vocals?!

fwessels: No, it is a bit more complicated than that, so also for 'silent' consonants, see http://english.blogoverflow.com/2011/11/articles-a-vs-an/

cmd/encryption.go
```diff
@@ -0,0 +1,199 @@
...
+	// SSECustomerKeyMD5 is the AWS SSE-C encryption key MD5 HTTP header key.
+	SSECustomerKeyMD5 = "X-Amz-Server-Side-Encryption-Customer-Key-MD5"
+)
+
+const (
+	// ClientSideEncryptionIV is the IV generated by the client on client-side-encryption
+	// or by the server on server-side-encryption.
+	ClientSideEncryptionIV = "X-Amz-Meta-X-Amz-Iv"
+
+	// ContentEncryptionAlgorithm is the algorithm used to encrypted the content.
```
fwessels: used to encrypted --> used to encrypt

cmd/encryption.go
fwessels: Ok

cmd/encryption.go
```diff
@@ -0,0 +1,199 @@
...
+	nonce := make([]byte, 32) // generate random nonce to derive encryption key
+	if _, err = io.ReadFull(rand.Reader, nonce); err != nil {
+		return nil, err
+	}
+
+	mac := hmac.New(sha256.New, key)
+	mac.Write(nonce)
+	content, err = sio.EncryptReader(content, sio.Config{Key: mac.Sum(nil)})
+	if err != nil {
+		return nil, errInvalidSSEKey
```
fwessels: Is this the appropriate error response (invalid SSE Key)?

cmd/encryption.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+	nonce, err := base64.StdEncoding.DecodeString(metadata[ClientSideEncryptionIV])
+	if err != nil {
+		return nil, errObjectTampered
+	}
+
+	mac := hmac.New(sha256.New, key)
+	mac.Write(nonce)
+	writer, err := sio.DecryptWriter(client, sio.Config{Key: mac.Sum(nil)})
+	fmt.Println(hex.EncodeToString(mac.Sum(nil)))
```
fwessels: Remove print statement

cmd/encryption.go
```diff
@@ -0,0 +1,199 @@
...
+		size += mod - 32
+	}
+	info.Size = size
+	return info, nil
+}
+
+// EncryptSSECustomerObjectInfo converts an object info of plain object into an
+// encrypted object info.
+//
+// EncryptSSECustomerObjectInfo computes the encrypted object size of an plain object.
```
fwessels: an plain --> a plain

cmd/object-handlers.go
```diff
@@ -88,6 +89,41 @@ func (f funcToWriter) Write(p []byte) (int, error) {
 	return f(p)
 }
 
+// HTTPWriteCloser writes to the underlying io.Writer
+// and also writes nil to the underlying io.Writer when
+// Close is called and no data was written.
+//
+// In case of a http.ResponseWriter the Close will set the
+// HTTP headers depending upon when no data was written to the client.
```
fwessels: when --> whether

cmd/object-handlers.go
```diff
@@ -159,40 +203,41 @@ func (api objectAPIHandlers) GetObjectHandler(w http.ResponseWriter, r *http.Req
...
-		return w.Write(p)
-	})
+		length = EncryptSSECustomerObjectInfo(objInfo).Size
+	}
+
+	setObjectHeaders(w, objInfo, hrange)
+	setHeadGetRespHeaders(w, r.URL.Query())
 
+	httpWriter := &HTTPWriteCloser{writer, false}
 	// Reads the object at startOffset and writes to mw.
```
fwessels: "writes to mw": outdated comment?

fwessels:
Code LGTM in general, nice work.

However shouldn't eg `PutObjectPartHandler` detect the presence of customer provided keys, and error out with a not-implemented for now? (Otherwise we may give the impression that we are encrypting whereas we are really not).

Same goes for `CopyObjectHandler` / `CopyObjectPartHandler` ...

As for `DeleteObjectHandler`, it is possible (in AWS) to delete an encrypted object without providing the encryption keys ?

cmd/object-handlers.go
aead: Oh, okay thanks

cmd/encryption.go
aead: done

cmd/encryption.go
aead: done

cmd/encryption.go
aead: Yes, and no - basically this should never happen because the size of the MAC is 32 bytes (SHA-256)
Currently `EncryptReader` will only return an error if the key is not 32 bytes long (all other config options are the default) which can never happen because SHA-256 produces 32 byte hash values. From my understanding the correct approach would be a panic - indicate a logical bug. However we agreed on: no panic in minio server code so it seems to be the correct error (the key has invalid size)
But we may can trace the error?

cmd/encryption.go
aead: Thanks, debug artifact...

cmd/encryption.go
aead: done

cmd/object-handlers.go
aead: done

cmd/object-handlers.go
aead: yes outdated comment

aead:
>However shouldn't eg PutObjectPartHandler detect the presence of customer provided keys, and error out with a not-implemented for now? (Otherwise we may give the impression that we are encrypting whereas we are really not).
>
>Same goes for CopyObjectHandler / CopyObjectPartHandler ...

I would agree.

>As for DeleteObjectHandler, it is possible (in AWS) to delete an encrypted object without providing the encryption keys ?

Good question. Cannot verify yet - @harshavardhana can you give me access to AWS again? (acc/sec key)?

aead:
Hm, github does not update the PR - however the patches are committed to the branch of my fork...
Try to fix otherwise I have to close and reopen...

harshavardhana:
You must have changed the branch name.. @aead

vadmeste:
> Hm, github does not update the PR - however the patches are committed to the branch of my fork...
> Try to fix otherwise I have to close and reopen...

I have the same problem.. I guess it is a temporary problem in github, it happens sometimes

aead:
@harshavardhana Same branch name... local, origin are the same: `encryption`? 
If you manually follow the link:
> Add more commits by pushing to the encryption branch on aead/minio.

It shows the changes in the fork but github does not update the PR.
Very wired, never seen this before...

aead:
Okay, rebase seems to update it... WTF?! Now the UI shows the commits... everything is "fine"

harshavardhana:
> Okay, rebase seems to update it... WTF?! Now the UI shows the commits... everything is "fine"

You might have to fix a conflict. @aead

harshavardhana:
Sorry for the delay on review will review this by EOD tomorrow my time.

cmd/api-errors.go
```diff
@@ -117,6 +117,13 @@ const (
 	ErrBucketAlreadyExists
 	ErrMetadataTooLarge
 	ErrUnsupportedMetadata
+	ErrInsecureSSECustomerRequest
```
harshavardhana: Add a line separating saying this is for `encryption` related errors..

cmd/api-errors.go
```diff
@@ -649,16 +692,16 @@ var errorCodeResponse = map[APIErrorCode]APIError{
...
-		Description:    "Your metadata headers exceed the maximum allowed metadata size.",
-		HTTPStatusCode: http.StatusBadRequest,
-	},
 	ErrUnsupportedMetadata: {
 		Code:           "InvalidArgument",
 		Description:    "Your metadata headers are not supported.",
 		HTTPStatusCode: http.StatusBadRequest,
 	},
+	ErrObjectTampered: {
+		Code:           "XMinioObjectModified",
```
harshavardhana: `XMinioObjectTampered` is better?

cmd/encryption-v1.go
```diff
@@ -0,0 +1,196 @@
...
+	"encoding/base64"
+	"errors"
+	"io"
+	"net/http"
+
+	sha256 "github.com/minio/sha256-simd"
+	"github.com/minio/sio"
+)
+
+// AWS errors for invalid SSE-C requests.
```
harshavardhana: Perhaps move the comment inside like 

```
var (
    // AWS errors..
    ..
    ... 
    // Minio errors.. 
    errObjectTampered = errors.New("The requested object was modified and may be compromised")
)
```

cmd/encryption-v1.go
```diff
@@ -0,0 +1,196 @@
...
+	if md5Sum := md5.Sum(key); !bytes.Equal(md5Sum[:], keyMD5) {
+		return nil, errSSEKeyMD5Mismatch
+	}
+	return key, nil
+}
+
+// EncryptRequest takes the client provided content and encrypts the data
+// with the client provided key. It also marks the object as client-side-encrypted
+// and sets the correct headers.
+func EncryptRequest(content io.Reader, size int64, r *http.Request, metadata map[string]string, sha256sum string) (io.Reader, error) {
```
harshavardhana: Each of these functions require a unit test case @aead

cmd/encryption-v1.go
```diff
@@ -0,0 +1,196 @@
...
+	reader := NewHashReader(content, size, metadata["etag"], sha256sum)
+
+	metadata[ClientSideEncryptionIV] = base64.StdEncoding.EncodeToString(nonce)
+	metadata[ContentEncryptionAlgorithm] = DareHmacSha256
+	return reader, nil
+}
+
+// DecryptRequest decrypts the object with the client provided key. It also removes
+// the client-side-encryption metadata from the object and sets the correct headers.
+func DecryptRequest(client io.Writer, r *http.Request, metadata map[string]string) (io.WriteCloser, error) {
```
harshavardhana, line 194: Each of these functions require a unit test case @aead

cmd/object-handlers.go
```diff
@@ -88,6 +89,41 @@ func (f funcToWriter) Write(p []byte) (int, error) {
 	return f(p)
 }
 
+// HTTPWriteCloser writes to the underlying io.Writer
+// and also writes nil to the underlying io.Writer when
+// Close is called and no data was written.
+//
+// In case of an http.ResponseWriter the Close will set the
+// HTTP headers depending upon whether no data was written to the client.
+type HTTPWriteCloser struct {
```
harshavardhana: We should move this into a package of its own @aead and make it generic of sort.  Not just `HTTPWriteCloser` ?

cmd/object-handlers.go
```diff
@@ -88,6 +89,41 @@ func (f funcToWriter) Write(p []byte) (int, error) {
 	return f(p)
```
harshavardhana: We can get rid of the above code if not used any more @aead

harshavardhana (changes requested):

harshavardhana:
> There are some things to do:
> 
> Support for Copy and CopyPart
> Support for PutPart
> Support for Get with HTTP ranges
> Error handling with AWS API compatible errors.
> Testing

Does this still hold true for this PR?

cmd/object-handlers.go
aead: Currently it's still used in gateway-handlers.go I left it because I tried to keep the PR small (focus on FS and XL first)
but I can replace the code in gateway too if required?!

harshavardhana: Yes we should..if this is better.

cmd/object-handlers.go
aead: We can - `pkg/ioutil`?! I can also take a look whether the functionality is already implemented somewhere but maybe a custom `ioutil` package is not a bad idea - there's probably more io related stuff we can extract an reuse...

cmd/object-handlers.go
aead: okay

pkg/ioutil/ioutil.go
```diff
@@ -0,0 +1,60 @@
...
+	io.Writer
+	hasWritten bool
+}
+
+func (w *WriteOnCloser) Write(p []byte) (int, error) {
+	w.hasWritten = true
+	return w.Writer.Write(p)
+}
+
+func (w *WriteOnCloser) Close() error {
```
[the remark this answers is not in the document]
aead: I implement io.Closer here...?!

aead:
About the wired AppVeyor: golang/go#18468

pkg/ioutil/ioutil.go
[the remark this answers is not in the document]
harshavardhana: `golint` doesn't understand..

harshavardhana:
> About the wired AppVeyor: golang/go#18468

yes planning to fix it by moving files as packages for external functionality @aead for the time being you can disable the `coverage` for windows line in appveyor.yml so that the tests succeed.

aead:
>  for the time being you can disable the coverage for windows line in appveyor.yml so that the tests succeed

Okay, will do

cmd/api-errors.go
```diff
@@ -698,6 +743,23 @@ func toAPIErrorCode(err error) (apiErr APIErrorCode) {
 		return apiErr
 	}
 
+	switch err { // SSE errors
+	case errInsecureSSERequest:
```
harshavardhana, line 759: Perhaps add test cases for these entries for their proper translation to right errorCode.

cmd/gateway-handlers.go
```diff
@@ -117,32 +118,19 @@ func (api gatewayAPIHandlers) GetObjectHandler(w http.ResponseWriter, r *http.Re
...
-	})
 
 	getObject := objectAPI.GetObject
 	if reqAuthType == authTypeAnonymous {
 		getObject = objectAPI.AnonGetObject
 	}
 
+	setObjectHeaders(w, objInfo, hrange)
+	setHeadGetRespHeaders(w, r.URL.Query())
+	httpWriter := ioutil.WriteOnClose(w)
```
harshavardhana, line 130: :+1: looks good now..

harshavardhana (changes requested):

appveyor.yml
```diff
@@ -40,17 +40,17 @@ test_script:
   - mkdir build\coverage
   - go test -v -timeout 17m -race github.com/minio/minio/cmd...
   - go test -v -race github.com/minio/minio/pkg...
-  - go test -v -timeout 17m -coverprofile=build\coverage\coverage.txt -covermode=atomic github.com/minio/minio/cmd
+  # - go test -v -timeout 17m -coverprofile=build\coverage\coverage.txt -covermode=atomic github.com/minio/minio/cmd
```
harshavardhana: We should add a FIXME with the issue reference.. here such that we can enable it in future.

harshavardhana (changes requested):
Other than minor couple of things this PR related to tests, once they are done we can take this in.

cmd/api-errors.go
aead: done

cmd/encryption-v1.go
aead: Will do

pkg/ioutil/ioutil.go
```diff
@@ -0,0 +1,61 @@
...
+	io.Writer
+	hasWritten bool
+}
+
+func (w *WriteOnCloser) Write(p []byte) (int, error) {
+	w.hasWritten = true
+	return w.Writer.Write(p)
+}
+
+// Close closes the WirteOnCloser. It behaves like io.Closer.
```
fwessels: WirteOnCloser --> WriteOnCloser

pkg/ioutil/ioutil_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	goioutil "io/ioutil"
+	"testing"
+)
+
+func TestCloseOnWriter(t *testing.T) {
+	writer := WriteOnClose(goioutil.Discard)
+	if writer.HasWritten() {
+		t.Error("WriteOnCloser must not been marked as HasWritten")
```
fwessels: must not been marked --> must not be marked

pkg/ioutil/ioutil_test.go
```diff
@@ -0,0 +1,39 @@
...
+	}
+	writer.Write(nil)
+	if !writer.HasWritten() {
+		t.Error("WriteOnCloser must been marked as HasWritten")
```
fwessels: Idem (and line 37 also)

fwessels (dismissed):
Apart from few minor changes, LGTM

pkg/ioutil/ioutil_test.go
aead: done

pkg/ioutil/ioutil.go
aead: done

pkg/ioutil/ioutil_test.go
aead: done

harshavardhana:
Still testing need one more day @aead

appveyor.yml
```diff
@@ -40,17 +40,19 @@ test_script:
   - mkdir build\coverage
   - go test -v -timeout 17m -race github.com/minio/minio/cmd...
   - go test -v -race github.com/minio/minio/pkg...
-  - go test -v -timeout 17m -coverprofile=build\coverage\coverage.txt -covermode=atomic github.com/minio/minio/cmd
+  # FIXME(aead): enable codecov after issue https://github.com/golang/go/issues/18468 is solved.
```
harshavardhana, line 43: Can you open a bug for this? we can fix this by moving generic files into their own packages. @aead

harshavardhana (changes requested):
There are two bugs in this implementation one is 

- non-TLS connections are not rejected if they don't comply with our server HMAC algo. For example the following code doesn't return any error and silently ignores. 

```
                                                                                                                                                                       
         content := bytes.NewReader([]byte("Hello again"))                                                                                                                    
         key := []byte("32byteslongsecretkeymustprovided")                                                                                                                    
         h := md5.New()                                                                                                                                                       
         h.Write(key)                                                                                                                                                         
         encryptionKey := base64.StdEncoding.EncodeToString(key)                                                                                                              
         encryptionKeyMD5 := base64.StdEncoding.EncodeToString(h.Sum(nil))                                                                                                    
                                                                                                                                                                              
         // Amazon S3 does not store the encryption key you provide.                                                                                                          
         // Instead S3 stores a randomly salted HMAC value of the                                                                                                             
         // encryption key in order to validate future requests.                                                                                                              
         // The salted HMAC value cannot be used to derive the value                                                                                                          
         // of the encryption key or to decrypt the contents of the                                                                                                           
         // encrypted object. That means, if you lose the encryption                                                                                                          
         // key, you lose the object.                                                                                                                                         
         var metadata = map[string]string{                                                                                                                                    
                 "x-amz-server-side-encryption-customer-algorithm": "AES256",                                                                                                 
                 "x-amz-server-side-encryption-customer-key":       encryptionKey,                                                                                            
                 "x-amz-server-side-encryption-customer-key-MD5":   encryptionKeyMD5,                                                                                         
         }                                                                                                                                                                    
                                                                                                                                                                              
║        minioClient.TraceOn(os.Stderr)                                                                                                                                       
║                                                                                                                                                                             
         // minioClient.TraceOn(os.Stderr) // Enable to debug.                                                                                                                
╫        _, err = minioClient.PutObject("testbucket", "my-encrypted-object.txt", content, 11, minio.PutObjectOptions{UserMetadata: metadata})                                 
         if err != nil {                                                                                                                                                      
                 log.Fatalln(err)                                                                                                                                             
         }                                                           
```

S3 behaves wrongly by ignoring the sent headers, instead we should simply not save the PutObject()  IMO and we should reject upon an error in algorithm.

In any case even if the PutObject succeeds on AWS S3, GetObject errors with wrong algorithm error. 

```
GET /my-encrypted-object.txt HTTP/1.1
Host: testbucket.s3.amazonaws.com
User-Agent: Minio (linux; amd64) minio-go/4.0.0
Authorization: AWS4-HMAC-SHA256 Credential=**REDACTED**/20171026/us-east-1/s3/aws4_request, SignedHeaders=host;x-amz-content-sha256;x-amz-date;x-amz-server-side-encryption-customer-algorithm;x-amz-server-side-encryption-customer-key;x-amz-server-side-encryption-customer-key-md5, Signature=**REDACTED**
X-Amz-Date: 20171026T051543Z
X-Amz-Server-Side-Encryption-Customer-Algorithm: AES251
X-Amz-Server-Side-Encryption-Customer-Key: MzJieXRlc2xvbmdzZWNyZXRrZXltdXN0cHJvdmlkZWQ=
X-Amz-Server-Side-Encryption-Customer-Key-Md5: 7PpPLAK26ONlVUGOWlusfg==
Accept-Encoding: gzip

HTTP/1.1 400 Bad Request
Connection: close
Transfer-Encoding: chunked
Content-Type: application/xml
Date: Thu, 26 Oct 2017 05:15:44 GMT
Server: AmazonS3
X-Amz-Id-2: mE+HSB3A7v7+0iLi0PXyuLEoSuxGfBxomG6zq+4ZuxQPapGYCJne72LY7+oirkrHDXe2ypoemDs=
X-Amz-Request-Id: A3556AD20473C95A

1a2
<Error><Code>InvalidEncryptionAlgorithmError</Code><Message>The Encryption request you specified is not valid. Supported value: AES256.</Message><ArgumentName>x-amz-server-side-encryption</ArgumentName><ArgumentValue>AES251</ArgumentValue><RequestId>A3556AD20473C95A</RequestId><HostId>mE+HSB3A7v7+0iLi0PXyuLEoSuxGfBxomG6zq+4ZuxQPapGYCJne72LY7+oirkrHDXe2ypoemDs=</HostId></Error>
```

- One more issue is GetObject() is returning back invalid headers 

```
---------START-HTTP---------
GET /testbucket/my-encrypted-object.txt HTTP/1.1
Host: localhost:9000
User-Agent: Minio (linux; amd64) minio-go/4.0.0
Authorization: AWS4-HMAC-SHA256 Credential=**REDACTED**/20171026/us-east-1/s3/aws4_request, SignedHeaders=host;x-amz-content-sha256;x-amz-date;x-amz-server-side-encryption-customer-algorithm;x-amz-server-side-encryption-customer-key;x-amz-server-side-encryption-customer-key-md5, Signature=**REDACTED**
X-Amz-Date: 20171026T051805Z
X-Amz-Server-Side-Encryption-Customer-Algorithm: AES256
X-Amz-Server-Side-Encryption-Customer-Key: MzJieXRlc2xvbmdzZWNyZXRrZXltdXN0cHJvdmlkZWQ=
X-Amz-Server-Side-Encryption-Customer-Key-Md5: 7PpPLAK26ONlVUGOWlusfg==
Accept-Encoding: gzip

HTTP/1.1 206 Partial Content
Transfer-Encoding: chunked
Accept-Ranges: bytes
Content-Type: application/xml
Date: Thu, 26 Oct 2017 05:18:05 GMT
Server: Minio/DEVELOPMENT.2017-10-26T04-55-33Z (linux; amd64)
Vary: Origin
X-Amz-Request-Id: 14F10708FB5C2DBF
---------END-HTTP---------
```

The returned value is sort of XML too which is also wrong and also it returns partial content in-fact it looks like its perhaps an error in this situation.

harshavardhana:
Here is the sample code 

```go
package main

import (
	"bytes"
	"crypto/md5"
	"encoding/base64"
	"io/ioutil"
	"log"
	"os"

	minio "github.com/minio/minio-go"
)

func main() {
	// Note: YOUR-ACCESSKEYID, YOUR-SECRETACCESSKEY, my-testfile, my-bucketname and
	// my-objectname are dummy values, please replace them with original values.

	// New returns an Amazon S3 compatible client object. API compatibility (v2 or v4) is automatically
	// determined based on the Endpoint value.
	minioClient, err := minio.New("localhost:9000", "USWUXHGYZQYFYFFIT3RE", "MOJRH0mkL1IPauahWITSVvyDrQbEEIwljvmxdq03", false)
	//minioClient, err := minio.New("s3.amazonaws.com", os.Getenv("ACCESS_KEY"), os.Getenv("SECRET_KEY"), false)
	if err != nil {
		log.Fatalln(err)
	}

	content := bytes.NewReader([]byte("Hello again"))
	key := []byte("32byteslongsecretkeymustprovided")
	h := md5.New()
	h.Write(key)
	encryptionKey := base64.StdEncoding.EncodeToString(key)
	encryptionKeyMD5 := base64.StdEncoding.EncodeToString(h.Sum(nil))

	// Amazon S3 does not store the encryption key you provide.
	// Instead S3 stores a randomly salted HMAC value of the
	// encryption key in order to validate future requests.
	// The salted HMAC value cannot be used to derive the value
	// of the encryption key or to decrypt the contents of the
	// encrypted object. That means, if you lose the encryption
	// key, you lose the object.
	var metadata = map[string]string{
		"x-amz-server-side-encryption-customer-algorithm": "AES256",
		"x-amz-server-side-encryption-customer-key":       encryptionKey,
		"x-amz-server-side-encryption-customer-key-MD5":   encryptionKeyMD5,
	}

	minioClient.TraceOn(os.Stderr)

	// minioClient.TraceOn(os.Stderr) // Enable to debug.
	_, err = minioClient.PutObject("testbucket", "my-encrypted-object.txt", content, 11, minio.PutObjectOptions{UserMetadata: metadata})
	if err != nil {
		log.Fatalln(err)
	}

	opts := minio.GetObjectOptions{}
	for k, v := range metadata {
		opts.Set(k, v)
	}
	coreClient := minio.Core{minioClient}
	reader, _, err := coreClient.GetObject("testbucket", "my-encrypted-object.txt", opts)
	if err != nil {
		log.Fatalln(err)
	}
	defer reader.Close()

	decBytes, err := ioutil.ReadAll(reader)
	if err != nil {
		log.Fatalln(err)
	}
	if !bytes.Equal(decBytes, []byte("Hello again")) {
		log.Fatalln("Expected \"Hello, world\", got %s", string(decBytes))
	}
}
```

aead:
@harshavardhana I think your example code does not work.
Reason: The PUT / GET options will convert the headers to `X-Amz-Meta-X-Amz-Server-Side...`
See e.g. [PutObjectOptions](https://github.com/minio/minio-go/blob/master/api-put-object.go#L81) I had this issue, too - Fixed my local version of minio-go for testing...

Will open an issue for windows build.

EDIT: Will investigate returned GET header issue...

harshavardhana:
There is one more issue in the PR @aead GetObject on an encrypted should fail with appropriate error like AWS S3 

```
---------START-HTTP---------
GET /my-encrypted-object.txt HTTP/1.1
Host: vadmeste.s3.amazonaws.com
User-Agent: Minio (linux; amd64) minio-go/4.0.0
Authorization: AWS4-HMAC-SHA256 Credential=**REDACTED**/20171026/us-east-1/s3/aws4_request, SignedHeaders=host;x-amz-content-sha256;x-amz-date, Signature=**REDACTED**
X-Amz-Date: 20171026T235312Z
Accept-Encoding: gzip

HTTP/1.1 400 Bad Request
Connection: close
Transfer-Encoding: chunked
Content-Type: application/xml
Date: Thu, 26 Oct 2017 23:53:12 GMT
Server: AmazonS3
X-Amz-Id-2: 4w5TxhKt9AXLA7nIQ8zkc9XXRZdojrpKL/wLcE6nvKufzRR7yHM9BVDFlMSpyPh4NUa77MEUZ+8=
X-Amz-Request-Id: F3A9C1CDC51BCDBA

165
<Error><Code>InvalidRequest</Code><Message>The object was stored using a form of Server Side Encryption. The correct parameters must be provided to retrieve the object.</Message><RequestId>F3A9C1CDC51BCDBA</RequestId><HostId>4w5TxhKt9AXLA7nIQ8zkc9XXRZdojrpKL/wLcE6nvKufzRR7yHM9BVDFlMSpyPh4NUa77MEUZ+8=</HostId></Error>
0
---------END-HTTP---------
2017/10/26 16:53:12 The object was stored using a form of Server Side Encryption. The correct parameters must be provided to retrieve the object.
exit status 1
``` 

Where as this PR is not doing that sends back HTTP 200 with encrypted data. 

```
---------START-HTTP---------
GET /testbucket/my-encrypted-object.txt HTTP/1.1
Host: localhost:9000
User-Agent: Minio (linux; amd64) minio-go/4.0.0
Authorization: AWS4-HMAC-SHA256 Credential=**REDACTED**/20171026/us-east-1/s3/aws4_request, SignedHeaders=host;x-amz-content-sha256;x-amz-date, Signature=**REDACTED**
X-Amz-Date: 20171026T235545Z
Accept-Encoding: gzip

HTTP/1.1 200 OK
Content-Length: 43
Accept-Ranges: bytes
Date: Thu, 26 Oct 2017 23:55:45 GMT
Etag: "77ff064f98d741e499b1abc12c144645"
Last-Modified: Thu, 26 Oct 2017 23:55:45 GMT
Server: Minio/DEVELOPMENT.2017-10-26T23-47-56Z (linux; amd64)
Vary: Origin
X-Amz-Meta-X-Amz-Cek-Alg: DARE-HMAC-SHA256
X-Amz-Meta-X-Amz-Iv: CJC7bfSPGuHZxlN8ZwGlEocDMgiGkKpoqCrrdsEMGog=
X-Amz-Request-Id: 14F1440689D31FCF
---------END-HTTP---------
2017/10/26 16:55:45 Expected "Hello, world", got %s 
Z��z6�u���&�8[��;����-
�wZ��&��P
exit status 1
``` 

Was it decide to do this way?  - This might be a bit confusing IMO i see that client side encryption headers are set appropriately, perhaps the idea was that it will be interpreted and decrypted on the client side?

harshavardhana:
IMO i think behaving like AWS S3 is correct otherwise the problem is that we might erroneously send encrypted data to clients which cannot interpret them properly.  Or the clients won't know what to set, atleast `aws-cli` and aws-sdks know what needs to be set appropriately to get the SSE-C working. 

We will loose the ability, expecting them to implement DARE256 is going to be harder.

aead:
@harshavardhana 
The server responds with encrypted data because server-side-encryption of minio and client-side-encryption are exchangeable while AWS does not provide this functionality (for AWS client and server side encryption are two completely separate things).
You can encrypt data at the client and decrypt data at the server or the other way around.
I agree that it could confuse clients, the argument was that users who use encryption must know (at least to some extend - manage pwds/keys a.s.o) what they do.
If we restrict the server to fail if keys are not provided we lose this functionality - on the other side we may can introduce a flag to control behavior...

EDIT: Correct, decision was to differ from AWS because of interoperability between client/server side encryption

aead:
BTW The question whether the server should / should not respond with encrypted data is not directly related to DARE. That's just a S3 restriction and not directly related to DARE

harshavardhana:
I will need to run it by AB once. It is quite confusing IMO to send encrypted data.

cmd/encryption-v1.go
```diff
@@ -161,51 +198,76 @@ func DecryptRequest(client io.Writer, r *http.Request, metadata map[string]strin
...
-// DecryptSSECustomerObjectInfo computes the real object size of an encrypted object.
-func DecryptSSECustomerObjectInfo(info ObjectInfo) (ObjectInfo, error) {
-	size := (info.Size / (32 + 64*1024)) * (64 * 1024)
-	if mod := info.Size % (32 + 64*1024); mod > 0 {
-		if mod < 33 {
-			return info, errObjectTampered
+// IsEncrypted returns true if the object is marked as encrypted.
+func (o *ObjectInfo) IsEncrypted() bool {
+	if _, ok := o.UserDefined[ServerSideEncryptionIV]; ok {
+		return true
```
harshavardhana: We should validate all keys and return true, just one key doesn't satisfy the object to be encrypted? @aead

cmd/encryption-v1.go
```diff
@@ -161,51 +198,76 @@ func DecryptRequest(client io.Writer, r *http.Request, metadata map[string]strin
...
+	}
+	return false
+}
+
+// DecryptedSize returns the size of the object after decryption in bytes.
+// It returns an error if the object is not encrypted or marked as encrypted
+// but has an invalid size.
+func (o *ObjectInfo) DecryptedSize() (int64, error) {
+	if !o.IsEncrypted() {
+		return -1, errObjectTampered // TODO(aead): think about that...
```
harshavardhana: what we do think about? this shouldn't be called based on the contract so caller if they are calling they are on their own and the decryption will fail, we don't need to return an error here IMHO.

harshavardhana: Also objectIsNotTampered is incorrect - we can say objectNotEcnrypted.

cmd/object-handlers.go
```diff
@@ -124,10 +124,19 @@ func (api objectAPIHandlers) GetObjectHandler(w http.ResponseWriter, r *http.Req
...
+		if IsSSECustomerRequest(r.Header) {
+			writeErrorResponse(w, ErrInvalidEncryptionParameters, r.URL)
+			return
+		}
+	case true:
+		if !IsSSECustomerRequest(r.Header) {
+			writeErrorResponse(w, ErrSSEEncryptedObject, r.URL)
+			return
+		}
+		objInfo.Size, err = objInfo.DecryptedSize() // we have to use pain size to get a valid range result
```
harshavardhana: Comment has a typo.

cmd/object-handlers.go
```diff
@@ -256,10 +265,23 @@ func (api objectAPIHandlers) HeadObjectHandler(w http.ResponseWriter, r *http.Re
 		writeErrorResponseHeadersOnly(w, apiErr)
 		return
 	}
-	if IsSSECustomerRequest(r.Header) {
-		objInfo, err = DecryptSSECustomerObjectInfo(objInfo)
+	switch objInfo.IsEncrypted() {
```
harshavardhana: Perhaps this repeated function can become a function? like `checkEncrypted` ?

harshavardhana (changes requested):

cmd/encryption-v1.go
aead: Exactly that were my thoughts ^^
Good catch - Okay so your saying we can ignore the case that the object may not be encrypted?
From my point of view a `panic` would be appropriate but I think we agreed on: no panic in codebase, right?

cmd/object-handlers.go
aead: Also thought about that but the HEAD must do the extra check whether the client key is correct while the GET checks this later (after range parsing) - Let me see if I can unify them...

cmd/object-handlers.go
aead: done

cmd/encryption-v1.go
harshavardhana: A panic is fine in this i think, since its unexpected usage of the API itself and potentially would warn developers directly. @aead

aead: Agree - okay, will add panic 👍

harshavardhana:
is this ready for review again? @aead  if yes can you squash all the changes? We are ready to get this in.

aead:
yes @harshavardhana, ready for review.
Squashed commits.

harshavardhana (dismissed):
All changes LGTM - tested the PR as well both with 

 - SSE-C headers using aws-sdk-go, minio-go, aws-cli

```
$ aws --no-verify-ssl --endpoint-url https://localhost:9000 s3api put-object --bucket testbucket \
    --key my-encrypted-object.txt --sse-customer-algorithm AES256 \
    --sse-customer-key MzJieXRlc2xvbmdzZWNyZXRrZXltdXN0cHJvdmlkZWQ=\
    --sse-customer-key-md5 7PpPLAK26ONlVUGOWlusfg== --body /etc/issue

$ aws --no-verify-ssl --endpoint-url https://localhost:9000 s3api get-object --bucket testbucket \
     --key my-encrypted-object.txt --sse-customer-algorithm AES256 \
     --sse-customer-key MzJieXRlc2xvbmdzZWNyZXRrZXltdXN0cHJvdmlkZWQ= \
     --sse-customer-key-md5 7PpPLAK26ONlVUGOWlusfg==
```

 - Client encryption headers using minio-go 

Both work fine and return appropriate errors providing strict compatibility with AWS S3. 

Nice stuff @aead

cmd/encryption-v1.go
```diff
@@ -0,0 +1,299 @@
...
+//
+// enc_key  := M(key, key_mac')
+const (
+	// ServerSideEncryptionIV is a 32 byte randomly generated IV used to derive an
+	// unique encryption key from the client provided key. The combination of this value
+	// and the client-provided key must be unique to provide the DARE tamper-proof property.
+	ServerSideEncryptionIV = ReservedMetadataPrefix + "Server-Side-Encryption-Iv"
+
+	// ServerSideEncryptionKDF is the combination of a hash and MAC function used to derive
+	// the SSE-C encryption key from the user-provied key.
```
fwessels: provied --> provided

cmd/encryption-v1.go
```diff
@@ -0,0 +1,299 @@
...
+	nonce, err := base64.StdEncoding.DecodeString(metadata[ServerSideEncryptionIV])
+	if err != nil || len(nonce) != 32 {
+		return nil, errObjectTampered
+	}
+	keyMAC, err := base64.StdEncoding.DecodeString(metadata[ServerSideEncryptionKeyMAC])
+	if err != nil || len(keyMAC) != 32 {
+		return nil, errObjectTampered
+	}
+
+	keyHash := sha256.Sum256(key) // verify that client provided is correct key
```
fwessels: verify that client provided is correct key --> verify that client provided correct key

cmd/object-handlers.go
```diff
@@ -666,6 +697,12 @@ func (api objectAPIHandlers) CopyObjectPartHandler(w http.ResponseWriter, r *htt
 		return
 	}
 
+	if IsSSECustomerRequest(r.Header) { // handle SSE-C requests
+		// SSE-C is not implemented for multipart operations yet
```
fwessels: multipart --> copy object part

fwessels (dismissed):
LGTM (few minor comments, nothing serious).

Great work, nice addition !

cmd/encryption-v1.go
aead: Thanks, done.

cmd/encryption-v1.go
aead: done

aead:
>Both work fine and return appropriate errors providing strict compatibility with AWS S3.
Nice stuff @aead

>Great work, nice addition !

Thanks, glad that it works in all the tested scenarios @harshavardhana 
Addressed the typos in the latest commit. I think we can get it in 🎉

harshavardhana (approved):
LGTM, approving again.

fwessels (approved):
Approve again, ready to merge

harshavardhana:
Codecov we can ignore, rest of the testing needs to be added as part of mint. We need to open bugs for this in mint  // cc @aead

PR #4913, opened 2017-09-13
Fix bug in ErasureStorage.HealFile

### LINKED ISSUE #4905: Bug in `ErasureStorage.HealFile()`
## Expected Behavior

Data should be reconstructed successfully.

## Current Behavior

Data reconstruction fails.

## Possible Solution

Not sure. There appears to be a bug in the way data is healed block-by-block near here: https://github.com/minio/minio/blob/master/cmd/erasure-healfile.go#L54 

Each "chunk" is being read from the disks and reconstructed, however, the offsets appear wrong - `offset` is incrementing by `blocksize` (10MiB usually), but the `chunksize` (amount being read from disk) is usually 5MiB - causing many parts of the file to not be read at all. 

In addition, due to the break statement at https://github.com/minio/minio/blob/master/cmd/erasure-healfile.go#L64 - the `blocks` slice is not cleared properly before each iteration of the for-loop at https://github.com/minio/minio/blob/master/cmd/erasure-healfile.go#L47.

Apply the following patch on master:

```patch
From 9b0872fa550f0c4168fc8882edac6e242a1234f9 Mon Sep 17 00:00:00 2001
From: Aditya Manthramurthy <aditya@minio.io>
Date: Mon, 11 Sep 2017 19:55:45 +0530
Subject: [PATCH] add failing test case

---
 cmd/erasure-healfile_test.go | 2 ++
 1 file changed, 2 insertions(+)

diff --git a/cmd/erasure-healfile_test.go b/cmd/erasure-healfile_test.go
index 65e23dc7..9dc6c28e 100644
--- a/cmd/erasure-healfile_test.go
+++ b/cmd/erasure-healfile_test.go
@@ -53,6 +53,8 @@ var erasureHealFileTests = []struct {
 	{dataBlocks: 12, disks: 16, offDisks: 2, badDisks: 1, badOffDisks: 0, blocksize: int64(blockSizeV1), size: oneMiByte, algorithm: DefaultBitrotAlgorithm, shouldFail: false, shouldFailQuorum: false}, // 18
 	{dataBlocks: 6, disks: 8, offDisks: 1, badDisks: 0, badOffDisks: 0, blocksize: int64(blockSizeV1), size: oneMiByte, algorithm: BLAKE2b512, shouldFail: false, shouldFailQuorum: false},               // 19
 	{dataBlocks: 7, disks: 10, offDisks: 1, badDisks: 0, badOffDisks: 0, blocksize: int64(blockSizeV1), size: oneMiByte, algorithm: 0, shouldFail: true, shouldFailQuorum: false},                        // 20
+	{dataBlocks: 2, disks: 4, offDisks: 1, badDisks: 0, badOffDisks: 0, blocksize: int64(blockSizeV1), size: oneMiByte * 64, algorithm: SHA256, shouldFail: false, shouldFailQuorum: false},             // 21
+
 }
 
 func TestErasureHealFile(t *testing.T) {
-- 
2.11.0

```

and run the following in `cmd/`:

```
$ go test -v -run TestErasureHealFile
=== RUN   TestErasureHealFile
--- FAIL: TestErasureHealFile (2.92s)
	erasure-healfile_test.go:93: Test 21: should pass but it failed with: failed to reconstruct data: no shard data
FAIL
exit status 1
FAIL	github.com/minio/minio/cmd	3.013s
```

## Context

When testing the Admin API PR https://github.com/minio/minio/pull/4826 with a file about ~100MB in size, there is an error from the `HealFile` function.

donatello:
Also adds test case that shows why the previous version was buggy and
refactors code for readability.

## Motivation and Context

Fixes #4905

## How Has This Been Tested?

New test case added, and tested manually (with changes in #4826)

## Types of changes

## Checklist:

### DIFF
--- a/cmd/erasure-healfile.go
+++ b/cmd/erasure-healfile.go
@@ -16,72 +16,140 @@
 
 package cmd
 
-import "hash"
+import (
+	"hash"
+)
 
-// HealFile tries to reconstruct a bitrot encoded file spread over all available disks. HealFile will read the valid parts of the file,
-// reconstruct the missing data and write the reconstructed parts back to the disks.
-// It will try to read the valid parts from the file under the given volume and path and tries to reconstruct the file under the given
-// healVolume and healPath. The given algorithm will be used to verify the valid parts and to protect the reconstructed file.
-func (s ErasureStorage) HealFile(offlineDisks []StorageAPI, volume, path string, blocksize int64, healVolume, healPath string, size int64, algorithm BitrotAlgorithm, checksums [][]byte) (f ErasureFileInfo, err error) {
+// HealFile tries to reconstruct an erasure-coded file spread over all
+// available disks. HealFile will read the valid parts of the file,
+// reconstruct the missing data and write the reconstructed parts back
+// to `staleDisks`.
+//
+// `staleDisks` is a slice of disks where each non-nil entry has stale
+// or no data, and so will be healed.
+//
+// It is required that `s.disks` have a (read-quorum) majority of
+// disks with valid data for healing to work.
+//
+// In addition, `staleDisks` and `s.disks` must have the same ordering
+// of disks w.r.t. erasure coding of the object.
+//
+// The function will try to read the valid parts from the file under
+// the given volume and path and tries to reconstruct the file under
+// the given healVolume and healPath (on staleDisks). The given
+// algorithm will be used to verify the valid parts and to protect the
+// reconstructed file.
+//
+// It returns bitrot checksums for the non-nil staleDisks.
+func (s ErasureStorage) HealFile(staleDisks []StorageAPI, volume, path string,
+	blocksize int64, healVolume, healPath string, size int64,
+	algorithm BitrotAlgorithm, checksums [][]byte) (f ErasureFileInfo,
+	err error) {
+
 	if !algorithm.Available() {
 		return f, traceError(errBitrotHashAlgoInvalid)
 	}
+
+	// Initialization
 	f.Checksums = make([][]byte, len(s.disks))
-	hashers, verifiers := make([]hash.Hash, len(s.disks)), make([]*BitrotVerifier, len(s.disks))
+	hashers := make([]hash.Hash, len(s.disks))
+	verifiers := make([]*BitrotVerifier, len(s.disks))
 	for i, disk := range s.disks {
-		if disk == OfflineDisk {
+		switch {
+		case staleDisks[i] != nil:
 			hashers[i] = algorithm.New()
-		} else {
+		case disk == nil:
+			// disregard unavailable disk
+			continue
+		default:
 			verifiers[i] = NewBitrotVerifier(algorithm, checksums[i])
 			f.Checksums[i] = checksums[i]
 		}
 	}
-	blocks := make([][]byte, len(s.disks))
+
+	// Scan part files on disk, block-by-block reconstruct it and
+	// write to stale disks.
 	chunksize := getChunkSize(blocksize, s.dataBlocks)
-	for offset := int64(0); offset < size; offset += blocksize {
-		if size < blocksize {
-			blocksize = size
+	var chunkOffset, blockOffset int64
+	for ; blockOffset < size; blockOffset += blocksize {
+		// last iteration may have less than blocksize data
+		// left, so chunksize needs to be recomputed.
+		if size < blockOffset+blocksize {
+			blocksize = size - blockOffset
 			chunksize = getChunkSize(blocksize, s.dataBlocks)
 		}
+
+		// read a chunk from each disk, until we have
+		// `s.dataBlocks` number of chunks set to non-nil in
+		// `blocks`
+		blocks := make([][]byte, len(s.disks))
+		var buffer []byte
 		numReads := 0
 		for i, disk := range s.disks {
-			if disk != OfflineDisk {
-				if blocks[i] == nil {
-					blocks[i] = make([]byte, chunksize)
-				}
-				blocks[i] = blocks[i][:chunksize]
-				if !verifiers[i].IsVerified() {
-					_, err = disk.ReadFileWithVerify(volume, path, offset, blocks[i], verifiers[i])
-				} else {
-					_, err = disk.ReadFile(volume, path, offset, blocks[i])
-				}
-				if err != nil {
-					blocks[i] = nil
-				} else {
-					numReads++
-				}
-				if numReads == s.dataBlocks { // we have enough data to reconstruct
-					break
-				}
+			// skip reading from unavailable or stale disks
+			if disk == nil || staleDisks[i] != nil {
+				continue
+			}
+			// allocate buffer only when needed - when
+			// reads fail, the buffer can be reused
+			if int64(len(buffer)) != chunksize {
+				buffer = make([]byte, chunksize)
+			}
+			if !verifiers[i].IsVerified() {
+				_, err = disk.ReadFileWithVerify(volume, path,
+					chunkOffset, buffer, verifiers[i])
+			} else {
+				_, err = disk.ReadFile(volume, path,
+					chunkOffset, buffer)
+			}
+			if err != nil {
+				// LOG FIXME: add a conditional log
+				// for read failures, once per-disk
+				// per-function-invocation.
+				continue
+			}
+
+			// read was successful, so set the buffer as
+			// blocks[i], and reset buffer to nil to force
+			// allocation on next iteration
+			blocks[i], buffer = buffer, nil
+
+			numReads++
+			if numReads == s.dataBlocks {
+				// we have enough data to reconstruct
+				break
 			}
 		}
+
+		// advance the chunk offset to prepare for next loop
+		// iteration
+		chunkOffset += chunksize
+
+		// reconstruct data - this computes all data and parity shards
 		if err = s.ErasureDecodeDataAndParityBlocks(blocks); err != nil {
 			return f, err
 		}
-		for i, disk := range s.disks {
-			if disk != OfflineDisk {
+
+		// write computed shards as chunks on file in each
+		// stale disk
+		for i, disk := range staleDisks {
+			if disk == nil {
 				continue
 			}
-			if err = offlineDisks[i].AppendFile(healVolume, healPath, blocks[i]); err != nil {
+
+			err = disk.AppendFile(healVolume, healPath, blocks[i])
+			if err != nil {
 				return f, traceError(err)
 			}
 			hashers[i].Write(blocks[i])
 		}
 	}
+
+	// copy computed file hashes into output variable
 	f.Size = size
 	f.Algorithm = algorithm
-	for i, disk := range s.disks {
-		if disk != OfflineDisk {
+	for i, disk := range staleDisks {
+		if disk == nil {
 			continue
 		}
 		f.Checksums[i] = hashers[i].Sum(nil)
--- a/cmd/erasure-healfile_test.go
+++ b/cmd/erasure-healfile_test.go
@@ -25,38 +25,51 @@ import (
 )
 
 var erasureHealFileTests = []struct {
-	dataBlocks                             int
-	disks, offDisks, badDisks, badOffDisks int
-	blocksize, size                        int64
-	algorithm                              BitrotAlgorithm
-	shouldFail                             bool
-	shouldFailQuorum                       bool
+	dataBlocks, disks int
+
+	// number of offline disks is also number of staleDisks for
+	// erasure reconstruction in this test
+	offDisks int
+
+	// bad disks are online disks which return errors
+	badDisks, badStaleDisks int
+
+	blocksize, size int64
+	algorithm       BitrotAlgorithm
+	shouldFail      bool
 }{
-	{dataBlocks: 2, disks: 4, offDisks: 1, badDisks: 0, badOffDisks: 0, blocksize: int64(blockSizeV1), size: oneMiByte, algorithm: SHA256, shouldFail: false, shouldFailQuorum: false},                   // 0
-	{dataBlocks: 3, disks: 6, offDisks: 2, badDisks: 0, badOffDisks: 0, blocksize: int64(blockSizeV1), size: oneMiByte, algorithm: BLAKE2b512, shouldFail: false, shouldFailQuorum: false},               // 1
-	{dataBlocks: 4, disks: 8, offDisks: 2, badDisks: 1, badOffDisks: 0, blocksize: int64(blockSizeV1), size: oneMiByte, algorithm: BLAKE2b512, shouldFail: false, shouldFailQuorum: false},               // 2
-	{dataBlocks: 5, disks: 10, offDisks: 3, badDisks: 1, badOffDisks: 0, blocksize: int64(blockSizeV1), size: oneMiByte, algorithm: DefaultBitrotAlgorithm, shouldFail: false, shouldFailQuorum: false},  // 3
-	{dataBlocks: 6, disks: 12, offDisks: 2, badDisks: 3, badOffDisks: 0, blocksize: int64(blockSizeV1), size: oneMiByte, algorithm: SHA256, shouldFail: false, shouldFailQuorum: false},                  // 4
-	{dataBlocks: 7, disks: 14, offDisks: 4, badDisks: 1, badOffDisks: 0, blocksize: int64(blockSizeV1), size: oneMiByte, algorithm: DefaultBitrotAlgorithm, shouldFail: false, shouldFailQuorum: false},  // 5
-	{dataBlocks: 8, disks: 16, offDisks: 6, badDisks: 1, badOffDisks: 1, blocksize: int64(blockSizeV1), size: oneMiByte, algorithm: DefaultBitrotAlgorithm, shouldFail: false, shouldFailQuorum: true},   // 6
-	{dataBlocks: 7, disks: 14, offDisks: 2, badDisks: 3, badOffDisks: 0, blocksize: int64(oneMiByte / 2), size: oneMiByte, algorithm: BLAKE2b512, shouldFail: true, shouldFailQuorum: false},             // 7
-	{dataBlocks: 6, disks: 12, offDisks: 1, badDisks: 0, badOffDisks: 1, blocksize: int64(oneMiByte - 1), size: oneMiByte, algorithm: DefaultBitrotAlgorithm, shouldFail: true, shouldFailQuorum: false}, // 8
-	{dataBlocks: 5, disks: 10, offDisks: 3, badDisks: 0, badOffDisks: 3, blocksize: int64(oneMiByte / 2), size: oneMiByte, algorithm: SHA256, shouldFail: true, shouldFailQuorum: false},                 // 9
-	{dataBlocks: 4, disks: 8, offDisks: 1, badDisks: 1, badOffDisks: 0, blocksize: int64(blockSizeV1), size: oneMiByte, algorithm: DefaultBitrotAlgorithm, shouldFail: false, shouldFailQuorum: false},   // 10
-	{dataBlocks: 2, disks: 4, offDisks: 1, badDisks: 0, badOffDisks: 1, blocksize: int64(blockSizeV1), size: oneMiByte, algorithm: DefaultBitrotAlgorithm, shouldFail: false, shouldFailQuorum: true},    // 11
-	{dataBlocks: 6, disks: 12, offDisks: 8, badDisks: 3, badOffDisks: 0, blocksize: int64(blockSizeV1), size: oneMiByte, algorithm: DefaultBitrotAlgorithm, shouldFail: false, shouldFailQuorum: true},   // 12
-	{dataBlocks: 7, disks: 14, offDisks: 3, badDisks: 4, badOffDisks: 0, blocksize: int64(blockSizeV1), size: oneMiByte, algorithm: BLAKE2b512, shouldFail: false, shouldFailQuorum: false},              // 13
-	{dataBlocks: 7, disks: 14, offDisks: 6, badDisks: 1, badOffDisks: 0, blocksize: int64(blockSizeV1), size: oneMiByte, algorithm: DefaultBitrotAlgorithm, shouldFail: false, shouldFailQuorum: false},  // 14
-	{dataBlocks: 8, disks: 16, offDisks: 4, badDisks: 5, badOffDisks: 0, blocksize: int64(blockSizeV1), size: oneMiByte, algorithm: DefaultBitrotAlgorithm, shouldFail: false, shouldFailQuorum: true},   // 15
-	{dataBlocks: 2, disks: 4, offDisks: 0, badDisks: 0, badOffDisks: 0, blocksize: int64(blockSizeV1), size: oneMiByte, algorithm: DefaultBitrotAlgorithm, shouldFail: false, shouldFailQuorum: false},   // 16
-	{dataBlocks: 2, disks: 4, offDisks: 0, badDisks: 0, badOffDisks: 0, blocksize: int64(blockSizeV1), size: oneMiByte, algorithm: 0, shouldFail: true, shouldFailQuorum: false},                         // 17
-	{dataBlocks: 12, disks: 16, offDisks: 2, badDisks: 1, badOffDisks: 0, blocksize: int64(blockSizeV1), size: oneMiByte, algorithm: DefaultBitrotAlgorithm, shouldFail: false, shouldFailQuorum: false}, // 18
-	{dataBlocks: 6, disks: 8, offDisks: 1, badDisks: 0, badOffDisks: 0, blocksize: int64(blockSizeV1), size: oneMiByte, algorithm: BLAKE2b512, shouldFail: false, shouldFailQuorum: false},               // 19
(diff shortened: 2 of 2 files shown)


donatello:
Marking as high priority as this is needed for https://github.com/minio/minio/pull/4826

cmd/erasure-healfile.go
```diff
@@ -36,48 +50,72 @@ func (s ErasureStorage) HealFile(offlineDisks []StorageAPI, volume, path string,
...
+
+			err = staleDisks[i].AppendFile(healVolume, healPath,
+				blocks[i])
+			if err != nil {
 				return f, traceError(err)
 			}
 			hashers[i].Write(blocks[i])
 		}
+
+		chunkOffset += chunksize
```
aead: Can we move this up? It is highly related to blockOffset but is incremented on the "other end" of the function....

cmd/erasure-healfile.go
```diff
@@ -36,48 +50,72 @@ func (s ErasureStorage) HealFile(offlineDisks []StorageAPI, volume, path string,
...
-				} else {
-					numReads++
-				}
-				if numReads == s.dataBlocks { // we have enough data to reconstruct
-					break
-				}
+			if disk == OfflineDisk {
+				continue
+			}
+			blocks[i] = make([]byte, chunksize)
```
aead:
```
if blocks[i] == nil || len(blocks[i]) < int(chunksize) {
     blocks[i] = make([]byte, chunksize)
}
```
To avoid unnecessary memory alloc.

aead (changes requested):

cmd/erasure-healfile.go
donatello: `blocks[i] == nil` is guaranteed just before this line, so I don't see why your suggestion reduces allocation. However, I am updating the PR with a better way to reduce allocations.

cmd/erasure-healfile.go
donatello: Sure, I'll move this up as far as possible.

It is also possible to write like `for i, j = 0, 0; i<10; i, j = i+di, j+dj {` but this way does not look good with longer variable names.

donatello:
Comments addressed @aead.

donatello:
@harshavardhana Please comment in the issue (as review) that on the commit - your previous comments and mine just got lost because I force pushed the branch :-(

~I hope you at least got an email about my reasoning about allocations. Otherwise I could rewrite it here.~

EDIT: Nevermind, it is not lost (yet): https://github.com/minio/minio/commit/f28e723ea82c222bda62549857746a97c277041c#commitcomment-24306122 - but github might remove it later as this stale commti may get pruned.

cmd/erasure-healfile.go
aead: Yes, but if one read fails (let's say one disk goes down while healing) the buffer would be set to nil again. ~~But your updated solution is also fine. Both are equivalent.~~

donatello:
Pinging for review @krishnasrinivas - and @aead do you have further feedback?

cmd/erasure-healfile.go
```diff
@@ -36,48 +50,86 @@ func (s ErasureStorage) HealFile(offlineDisks []StorageAPI, volume, path string,
...
-				}
-				if numReads == s.dataBlocks { // we have enough data to reconstruct
-					break
-				}
+			if disk == OfflineDisk {
+				continue
+			}
+			// allocate buffer only when needed - when
+			// reads fail, the buffer can be reused
+			if int64(len(buffer)) != chunksize {
```
aead, line 95: This must be:
```
if  len(blocks[i]) == nil {
    blocks[i] = make([]byte, chunksize)
}
```

cmd/erasure-healfile.go
```diff
@@ -36,48 +50,86 @@ func (s ErasureStorage) HealFile(offlineDisks []StorageAPI, volume, path string,
...
+				buffer = make([]byte, chunksize)
+			}
+			if !verifiers[i].IsVerified() {
+				_, err = disk.ReadFileWithVerify(volume, path,
+					chunkOffset, buffer, verifiers[i])
```
aead, line 100: And here: `blocks[i][:chunksize]`

cmd/erasure-healfile.go
```diff
@@ -36,48 +50,86 @@ func (s ErasureStorage) HealFile(offlineDisks []StorageAPI, volume, path string,
...
+			} else {
+				_, err = disk.ReadFile(volume, path,
+					chunkOffset, buffer)
```
aead, line 103: And here: `blocks[i][:chunksize]`

krishnasrinivas:
@donatello  yet to review, will review

cmd/erasure-healfile.go
```diff
@@ -36,48 +50,86 @@ func (s ErasureStorage) HealFile(offlineDisks []StorageAPI, volume, path string,
...
+				// LOG FIXME: add a conditional log
+				// for read failures, once per-disk
+				// per-function-invocation.
+				continue
+			}
+
+			// read was successful, so set the buffer as
+			// blocks[i], and reset buffer to nil to force
+			// allocation on next iteration
+			blocks[i], buffer = buffer, nil
```
aead, line 115: This does not work. The problem is that this solution will allocate new buffers all the time. `buffer` is always set to nil so `int64(len(buffer)) != chunksize` always fails and the buffer gets allocated again.

donatello:
@krishnasrinivas you meant me, but notified someone else above.

aead (changes requested):

cmd/erasure-healfile.go
harshavardhana: Correction `len(blocks[i]) == 0`

cmd/erasure-healfile.go
donatello: Not sure if your last comment means you are happy with the allocations behaviour - I think it is ok. My thoughts about further optimization for this are here - https://github.com/minio/minio/commit/f28e723ea82c222bda62549857746a97c277041c#commitcomment-24306122.

cmd/erasure-healfile.go
donatello: No, it is correct as written. The desired size of the buffer is how much we want to read. That is exactly `chunksize`. Remember that for last iteration chunksize may change - it is not a constant throughout the for-loop.

cmd/erasure-healfile.go
donatello: Not needed. Can you explain why you feel it is wrong?

cmd/erasure-healfile.go
donatello: I hope you have read all the changes in this commit - the blocks slice itself is reallocated just before each file-reads loop.

cmd/erasure-healfile.go
donatello: It does work. See https://play.golang.org/p/abVips4qTA

Also, buffer is set to nil only when the read succeeds. When it fails, the buffer is reused. As noted in the comments.

cmd/erasure-healfile.go
donatello: Actually, I agree that @harshavardhana 's suggestion above will work as buffer is redeclared before the file-read loop. However, it is also correct as written, so I don't feel it necessary to change it.

cmd/erasure-healfile.go
aead: I'm not saying it does not work in the sense of correctness - I'm saying it does not reduce the number of allocations.
The reason is that your solution allocates always a new block on a successful read - which should happen much more often than a fail (the disks are already filtered: online | offline). See this simpler comparison of both solutions: [https://play.golang.org/p/0t5FxQ0k5n](https://play.golang.org/p/0t5FxQ0k5n)

cmd/erasure-healfile.go
aead: As said below it works - it is just not efficient in the sense of mem-allocs. On successful read buffer is always allocated. On `chunksize < len(buffer)` buffer is allocated. Instead of allocating the buffer on expected situations the buffer should be allocated on unexpected situations (read fails). Further if the`chunksize < len(buffer)` there is no reason to alloc a buffer. The buffer is large enough to hold to the data - it must only be truncated which is fine because the a smaller chunksize can only happen on last read.

cmd/erasure-healfile.go
donatello: @aead Thank you for the detailed explanation. However, your solution appears to have a bug. When `blocks` is passed to `s.ErasureDecodeDataAndParityBlocks()`, it is required to have the following structure - any `s.dataBlocks` number of slices in `blocks` must be containing valid data (therefore not set to nil), all the other slices are required to be set to nil to indicate that they are missing blocks (i.e. erasure reconstruction will allocate and reconstruct them) - [(doc)](https://github.com/donatello/minio/blob/429aa1e0ab3b736268ae12365ba6fda3a8428d67/vendor/github.com/klauspost/reedsolomon/reedsolomon.go#L50).

Now, read failures can happen to any subset of the online disks for any blocks (specially, even between blocks in the same object - for e.g. because disk errors can happen randomly even to online disks). 

Consider a situation where after reading chunk number `i` successfully, the file-reading loop breaks because `s.dataBlocks` number of chunks have been collected. At this point, any of the slices in `blocks` with slice index > `i` may not be nil (due to sharing `blocks` across iterations of the outer for-loop`). If this happens, erasure reconstruction will receive garbage and return garbage.

Due to this possibility, in my solution, `blocks` is not shared across iterations of the outer for-loop. A difficulty in further reducing memory allocs is that the lower-level erasure reconstruction function always allocates the blocks it is reconstructing. Given this limitation the only allocations that can be saved are the ones due to file-read failures.

aead: @donatello Yes, I see the problem with Reconstruct. In general correct - we need to mark all shards as missing. I looked at the reedsolomon code and there is actually a [problem](https://github.com/klauspost/reedsolomon/blob/master/reedsolomon.go#L685) - it should check whether the capacity is large enough to hold the shard. I'll open an issue and send a patch.
Without this library fix your solution is completely fine and solves the issues. As soon we get this fix into reedsolomon we can optimize HealFile correctly - for now there is nothing we can do about it.
Thanks for the hint :+1:

aead (approved):
LGTM - further mem-optimization could be applied after fixing:
[reedsolomon](https://github.com/klauspost/reedsolomon/blob/master/reedsolomon.go#L685)

cmd/erasure-healfile.go
```diff
@@ -36,48 +50,86 @@ func (s ErasureStorage) HealFile(offlineDisks []StorageAPI, volume, path string,
...
 		if err = s.ErasureDecodeDataAndParityBlocks(blocks); err != nil {
 			return f, err
 		}
+
+		// write computed shards as chunks on file in each
+		// stale disk
 		for i, disk := range s.disks {
 			if disk != OfflineDisk {
 				continue
 			}
```
krishnasrinivas: above for condition should be:
```
for i, disk := range staleDisks {
    if disk == nil {
        continue
    }
....
```
because, else, we might endup trying to write to a disk that is down and return error

cmd/erasure-healfile.go
```diff
@@ -36,48 +50,86 @@ func (s ErasureStorage) HealFile(offlineDisks []StorageAPI, volume, path string,
...
+			if err != nil {
+				// LOG FIXME: add a conditional log
+				// for read failures, once per-disk
+				// per-function-invocation.
+				continue
+			}
+
+			// read was successful, so set the buffer as
+			// blocks[i], and reset buffer to nil to force
+			// allocation on next iteration
```
krishnasrinivas, line 114: > blocks[i], buffer = buffer, nil

blocks[i] = buffer
buffer = nil

doing it in two separate lines will be more readable because reading from and writing to `buffer` in the same line is little difficult to read.

cmd/erasure-healfile.go
```diff
@@ -36,48 +50,86 @@ func (s ErasureStorage) HealFile(offlineDisks []StorageAPI, volume, path string,
...
 				return f, traceError(err)
 			}
 			hashers[i].Write(blocks[i])
 		}
 	}
+
+	// copy computed file hashes into output variable
 	f.Size = size
 	f.Algorithm = algorithm
 	for i, disk := range s.disks {
```
krishnasrinivas: `for i, disk := range staleDisks {` here as well

cmd/erasure-healfile.go
harshavardhana: We don't use nil anymore. It's renamed to OfflineDisk. Assuming that you didn't see it.

krishnasrinivas: for staleDisks we cannot use OfflineDisk for nil i.e if it is `nil` it can also mean that it is online but not stale.

harshavardhana: > for staleDisks we cannot use OfflineDisk for nil i.e if it is nil it can also mean that it is online but not stale.

Okay i see now why staleDisks can be nil.  Currently i think `disk != offlineDisk` is working somehow incidentally and there is a bug out there which will cause eventually the below code to crash. i.e when `staleDisks[i] == nil`

donatello: Yes, @krishnasrinivas is right here. We should only be writing non-nil stale disks. And yes, when making changes it exposes some other issues - I am checking and fixing.

donatello:
This patch is updated. The new commit message reads:

    Fix bug in ErasureStorage.HealFile
    
    This change includes the following:
    
    * Fix for chunk and block scanning bug in `ErasureStorage.HealFile()`.
    
    * Fix for allocation and population of data in the `blocks` slice that
      is the argument to `ErasureDecodeDataAndParityBlocks()`.
    
    * Adds a test-case for the above.
    
    * Correct `ErasureStorage.HealFile()` semantics to only write data to
      the stale disks passed as an argument.
    
    * Fix for test cases to work with the above mentioned modified
      semantics.
    
    Fixes #4905

To incorporate the semantic change of writing only to stale disks in HealFile(), the test cases had to be changed. It is not valid to call HealFile with same set of disks as the underlying `ErasureStorage` struct. `HealFile()` now ignores stale disks when reading files for erasure reconstruction.

The tests are changed as follows:
- the previous `shouldFail` test parameter has been removed, and the previous `shouldFailQuorum` parameter has been renamed to `shouldFail`.
- tests only check that checksums for healed disks is valid (and not for all disks like previously), as this is what HealFile() now returns.

cmd/erasure-healfile.go
donatello: I disagree on this. Any potential reader is expected to know some golang syntax. This style is idiomatic in golang and is used for example, to swap two variables - it is used in our codebase like this (try running `git grep '.*,[^:]*=[^{,(]*,[^,)}]*$'` to find all double-assignment statements in minio).

In addition, (1) this line is concise (and perhaps even elegant), and (2) it has a clear explanatory comment just above. 

Writing it as a swap operation would also be correct and idiomatic, but less obviously understandable - `blocks[i], buffer = buffer, blocks[i]`.

If you feel strongly about this, please let me know the reason. Even if one more reviewer feels it is difficult to read because it is a multi-assignment statement, I am ok to change it.

cmd/erasure-healfile.go
donatello: Fixed.

cmd/erasure-healfile.go
donatello: Fixed and patch updated.

cmd/erasure-healfile.go
```diff
@@ -16,72 +16,136 @@
...
+
+	// Initialization
 	f.Checksums = make([][]byte, len(s.disks))
-	hashers, verifiers := make([]hash.Hash, len(s.disks)), make([]*BitrotVerifier, len(s.disks))
+	hashers := make([]hash.Hash, len(s.disks))
+	verifiers := make([]*BitrotVerifier, len(s.disks))
 	for i, disk := range s.disks {
-		if disk == OfflineDisk {
+		switch {
+		case staleDisks[i] != nil:
```
aead, line 59: ~~Those two cases are identical?! `staleDisk != nil` whenever `s.disk == OfflineDisk` and vice versa.~~ Further we agreed on using `OfflineDisk` as indicator whether disk is available -> s/`staleDisk != nil`/`staleDisk != OfflineDisk`

cmd/erasure-healfile.go
```diff
@@ -16,72 +16,136 @@
...
-				if err != nil {
-					blocks[i] = nil
-				} else {
-					numReads++
-				}
-				if numReads == s.dataBlocks { // we have enough data to reconstruct
-					break
-				}
+			// skip reading from offline or stale disks
+			if disk == OfflineDisk || staleDisks[i] != nil {
```
aead: Same as above - disk should be marked as offline whenever it's stale

aead:
@donatello Can you explain why it is not valid to use the ErasureStorage set of disks? All disks of ErasureStorage which should be healed are marked as OfflineDisk - this should also include stale disks...

aead (changes requested):

aead:
Understood the semantics now - thanks @harshavardhana

cmd/erasure-healfile.go
```diff
@@ -16,72 +16,136 @@
...
+)
+
+// HealFile tries to reconstruct a bitrot encoded file spread over all
+// available disks. HealFile will read the valid parts of the file,
+// reconstruct the missing data and write the reconstructed parts back
+// to `staleDisks`.
+//
+// `staleDisks` is a slice of disks where each non-nil entry
+// represents an online disk that is known to have stale or no data,
+// and so will be healed.
```
aead: Maybe add:
> `staleDisks` are used to avoid writes to unavailable dsiks and support partial healing

This was my idea of HealFile - I thought we try to heal all disks which are marked as offline and fail if at least on write fails -> no partial heal. Now I see that heal tries to heal a set of disks which are explicitly marked as "healable". However, we still fail if one of the "healable" disks goes down.

aead (changes requested):

cmd/erasure-healfile.go
krishnasrinivas: > Writing it as a swap operation would also be correct and idiomatic, but less obviously understandable - blocks[i], buffer = buffer, blocks[i].

Swapping somehow looks easy on the eyes. Maybe because it is a recognised pattern.

> blocks[i], buffer = buffer, nil

This is a little harder to perceive immediately. My subjective view, hence you can ignore my comment.

cmd/erasure-healfile.go
donatello: The slice of disks mean different things - `staleDisks` is non-nil when the disk is known to have out-of-date data. It is better to not conflate offline with nil.

donatello:
> @donatello Can you explain why it is not valid to use the ErasureStorage set of disks? All disks of ErasureStorage which should be healed are marked as OfflineDisk - this should also include stale disks...

@aead The ErasureStorage disks are a list of disks where the first `dataDisks` are data shards and the others are parity shards in a particular ordering that is per-object. Those are that are nil in these disks are offline here. Among the non-nil ones, some are out-of-date and some are up-to-date. When healing an object, we first check which shards are online and up-to-date and the out-of-date online ones are healed via HealFile in this patch. In the earlier version HealFile, it was over-writing even up-to-date shards - this is unnecessary and introduces the possibility of new errors.

Thus, it is enough to write to disks that are known to be stale. This is checked before HealFile is called.

cmd/erasure-healfile.go
donatello: Partial healing is supported in the sense that disks that are offline may be having out-of-date data, but we still continue and heal online disks with stale data. The comment as it stands, IMO  expresses what the function is doing correctly, and what your comment says is already implicitly implied.

The point about failing if one of the stale disks goes down while writing is well-taken. It can occasionally happen, though the caller of HealFile already confirms that they are online. i will look at enhancing this in a subsequent patch to return errors per-disk in case of write failures, and also optimizing the calling code that appears to do multiple passes over the data - many of these optimizations are already present in the new admin API PR. However, it would be better to separate out those optimizations into a new patch, and reduce the amount of low level changes in the admin API PR.

donatello:
This patch appears complete now and fixes multiple bugs. Further optimizations are possible, and I intend to send a patch that optimize the calling code to do only a single pass over the data.

Reviewers, do you have any more changes to be requested on this patch?

aead (approved):
LGTM

cmd/erasure-healfile.go
krishnasrinivas: Thinking more about this, `disk == OfflineDisk` is incorrect, because if `disk` is `nil` it can also mean that the disk is online but outdated.

So the comparison is `disk == nil` will be correct.

Also you don't need `staleDisks[i] != nil` because if `disk` is not `nil` it already means that it is not outdated.

You could use the same term `outDatedDisks` that is used in the caller instead of `staleDisks` for better readability.

donatello: @krishnasrinivas 
This loop is over the `s.disks` slice. This is a list of disks built before any verification of data on the disks based on the erasure ordering determined from `xl.json`. The caller ensures that a majority of this slice is non-nil. There is nothing wrong with considering the nil disks in this slice as offline (also in the code `OfflineDisk == nil` - I am not sure if you knew). Changing `OfflineDisk` to `nil` or vice-versa has no impact on the correctness of the algorithm.

`staleDisks` as defined in the function's doc (at the beginning) is a slice with non-nil items being online disks with stale (or no) data. These are the only disks on which healing is attempted here.

The check `staleDisks[i] != nil` is useful, because we need not try to read a valid shard (for reconstruction) from a disk that is confirmed to be stale (the caller checks that the disk is outdated). We only get a verification error.

In addition, if `staleDisks[i] != nil`, then `disk != nil` as staleDisks is a subset of `s.disks`, however `disk != nil` has no implication on the i-th staleDisk being nil or not.

Now, with this clarification, I think the algorithm implemented in this function should be clear - to me it looks correct.

krishnasrinivas: > This loop is over the s.disks slice. This is a list of disks built before any verification of data on the disks based on the erasure ordering determined from xl.json. The caller ensures that a majority of this slice is non-nil.

I think you are assuming that `s.disks` contains all the online disks but just ordered correctly.
Check out `listOnlineDisks()` implementation, it returns the latest disks based on common modTime. Perhaps we can rename this function to `listLatestDisks()` as it does not list Online disks which are outdated.

> Changing OfflineDisk to nil or vice-versa has no impact on the correctness of the algorithm.

I was pointing to the readability because `if disk == OfflineDisk` reads as `if the disk is Offline` which is incorrect as the particular disk can be online but outdated.

> In addition, if staleDisks[i] != nil, then disk != nil as staleDisks is a subset of s.disks

Incorrect, `s.disks` and `staleDisks` are actually mutually exclusive *except* for the disks that are offline.

> however disk != nil has no implication on the i-th staleDisk being nil or not

if `s.disk[i] != nil` it definitely implies that `staleDisk[i] == nil` however if `s.disks[i] == nil` it does not imply that `staleDisk[i] != nil` as that disk may be offline.

donatello: I am starting over the clarification in the main thread. See below.

donatello:
Let me start over - I realize that some of the things I said earlier about subset relationships among the disks are not entirely accurate. So let's disregard all that.

I will state the assumptions made about the disks in `HealFile()`, so that we can be sure that the implemented algorithm is correct.

1. `s.disks` is a slice of disks where the non-nil entries include a read-quorum of disks that have valid data.
2. `staleDisks` is a slice of disks, where each non-nil entry has stale or no data.
3. Both the above mentioned slices are ordered appropriately for erasure coding.

The above are the minimum requirements. It is the caller's responsibility that these requirements are met. It is `HealFile()`'s responsibility that whenever these assumptions are met and there are no disk errors (during reading/writing of data) the healing of `staleDisks` will work correctly. (When the assumptions are not met, as is usual, we impose no requirement on `HealFile()`'s output and side-effects.)

The assumptions above also obviate the need to characterize disks as offline or online. This is not really useful as disks can change from being online to offline or vice versa at *any* time. So only non-nil disks are used for reading data (with verification for reconstruction) and only non-nil entries are written to for healing.

To remove confusion related to offline vs online, I am removing usage of the constant `OfflineDisk` from `erasure-healfile.go`. It was being used to mean "unavailable", but since the term is a bit overloaded, it is leading to some confusion, as is evident from this thread.

To justify why the assumptions are met when `HealFile()` is called:

1. The caller sets `s.disks` as disks with the latest version of xl.json (i.e. `latestDisks` in the caller). Further, it is ensuring that the subset `availableDisks` computed by filtering `latestDisks` in `disksWithAllParts()` has valid data, and that there are read-quorum number of such disks. This ensures that assumption 1 is met.
2. In the caller `outdatedDisks` is computed as the set difference "storageDisks - availableDisks". Thus, `outdatedDisks` is the same as `storageDisks` but with disks with valid data removed. Among these disks, we attempt to heal the non-nil entries. This ensures assumption 2.
3. The last assumption is trivially met by the caller due to the calls to `shuffleDisks()`.

The patch is updated, now including the assumptions in the function's doc. The algorithm/code remains the same.

donatello:
The build error is due to an update to golint - there is already a fix for it at https://github.com/minio/minio/pull/4941 - please disregard the failure when reviewing.

krishnasrinivas (approved):

PR #4917, opened 2017-09-14
fs: make multipart clean config configurable

fangyuxiangGL:
make multipart clean config configurable

## Description
we can configure multipart clean config in config.json, which is smarter. And a new configuration entry 'unit' is also added to control the time unit of the config.

## Motivation and Context
currently, multipart clean config is static written in codes, we have to rebuild the executable if we want to change it

## Types of changes

## Checklist:

### DIFF
--- a/cmd/config-v19.go
+++ b/cmd/config-v19.go
@@ -51,6 +51,7 @@ type serverConfigV19 struct {
 
 	// Notification queue configuration.
 	Notify *notifier `json:"notify"`
+	Multipart *fsMultipart `json:"multipart"`
 }
 
 // GetVersion get current config version.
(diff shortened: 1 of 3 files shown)


harshavardhana:
Thanks for the PR but this is not correct we don't wish to configure these values in this manner. `config.json` is a carefully curated key value and is expanded only with careful understanding.

Feel free to discuss this in our slack channel.

fangyuxiangGL:
@harshavardhana 
ok ! thanks for your remind

harshavardhana:
This we need to discuss internally @fangyuxiangGL wishes to configure this we have two main options 
- Adding ENV variables on the lines of _MINIO_CACHE=off which are restricted envs just for this. 
- Implementing right API
   - http://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketPUTlifecycle.html 
   - http://docs.aws.amazon.com/AmazonS3/latest/dev/intro-lifecycle-rules.html 
   - http://docs.aws.amazon.com/AmazonS3/latest/dev/mpuoverview.html#mpu-abort-incomplete-mpu-lifecycle-config

PR #4922, opened 2017-09-14
[WIP] Disk cache

donatello:
This PR continues from https://github.com/minio/minio/pull/4156

It is still WIP - I will update this as progress is made.

It appears to work on master now, but there are things to be done:

TODOs (identified so far):
- complete review and and fix any bugs found
- update boltdb to a maintained version
- add support to cache on PUTs too.

## Types of changes

## Checklist:


donatello:
@harshavardhana For an older commit you mentioned locked-file semantics (to allow rename) - that led me onto another problem. I don't think caching (with shared cache-dir) can use Bolt DB as it exclusively locks the DB file - this would be a problem in shared FS mode. In this case, I would think that it is required to use a separate cache-dir for each minio instance sharing the same disk in FS mode. If a separate cache-dir is used, it is not required to have locked files within the cache-dir.

CC: @krishnasrinivas

PR #4925, opened 2017-09-15
refactor ObjectLayer PutObject and PutObjectPart (4923)

### LINKED ISSUE #4923: refactor ObjectLayer PutObject and PutObjectPart
PutObject and PutObjectPart of ObjectLayer take a io.Reader and a size to read as many bytes as specified (size) form the client. Further Minio has to verify the content send by client. Therefore Minio computes the MD5 sum (and SHA256 sum if requested) of the content. This verification is done inside every PutObject and PutObjectPart implementation of every ObjecLayer impl. (S3, XL, FS, GCS, Azure). 

## Expected Behavior
The content verification should be centralized in the API layer (S3).
 
## Current Behavior
The content verification is done in the ObjectLayer

## Possible Solution
Change ObjectLayer API to take a `HashReader` which hold some state to do the verification at any time (Inside ObjectLayer or API Layer). 

## Context
This is currently a problem for SSE-C. The server must verify the content of the client in SSE-C requests. The SSE-C encryption is happening in the API Layer. The content verification is currently done in the ObjecLayer. The encryption changes the data (encrypt it) -> content verification fails.

Also related: #4648 

## Your Environment
Current master + all OS

aead:
## Description

This change refactor the ObjectLayer PutObject and PutObjectPart
functions. Instead of passing an io.Reader and a size to PUT operations
ObejectLayer expects an HashReader.
A HashReader verifies the MD5 sum (and SHA256 sum if required) of the object.
This change updates all all PutObject(Part) calls and removes unnecessary code
in all ObjectLayer implementations.

## Motivation and Context

Fixes #4923

## Types of changes

## Checklist:


cmd/fs-v1-multipart.go
```diff
@@ -523,44 +520,22 @@ func (fs fsObjects) PutObjectPart(bucket, object, uploadID string, partID int, s
...
-	}
-
-	teeReader := io.TeeReader(limitDataReader, multiWriter)
 	bufSize := int64(readSizeV1)
-	if size > 0 && bufSize > size {
+	if size := data.Size(); size > 0 && bufSize > size {
 		bufSize = size
 	}
-	buf := make([]byte, int(bufSize))
+	buf := make([]byte, readSizeV1)
```
harshavardhana: This should be `bufSize`  it is not needed to buffer more for smaller files.

aead: Oh yes, thanks :+1:

harshavardhana (approved):

cmd/fs-v1_test.go
```diff
@@ -146,19 +143,19 @@ func TestFSPutObject(t *testing.T) {
 	}
 
 	// With a directory object.
-	_, err = obj.PutObject(bucketName+"non-existent", objectName+"/", int64(0), bytes.NewReader([]byte("")), nil, sha256sum)
+	_, err = obj.PutObject(bucketName+"non-existent", objectName+"/", NewHashReader(bytes.NewReader([]byte("abcd")), 0, "", ""), nil)
```
fwessels, line 146: bytes.NewReader should be empty?

cmd/fs-v1_test.go
```diff
@@ -173,7 +170,7 @@ func TestFSPutObject(t *testing.T) {
 		}
 	}
 
-	_, err = obj.PutObject(bucketName, objectName+"/1/", 0, bytes.NewReader([]byte("")), nil, sha256sum)
+	_, err = obj.PutObject(bucketName, objectName+"/1/", NewHashReader(bytes.NewReader([]byte("abcd")), 0, "", ""), nil)
```
fwessels, line 173: byes.NewReader should be empty?

cmd/gateway-azure.go
```diff
@@ -580,33 +520,20 @@ func (a *azureObjects) PutObjectPart(bucket, object, uploadID string, partID int
...
-		}
-	}
-
-	if sha256sum != "" {
-		newSHA256sum := hex.EncodeToString(sha256Writer.Sum(nil))
-		if newSHA256sum != sha256sum {
-			return PartInfo{}, azureToObjectError(traceError(SHA256Mismatch{}))
-		}
+	if err = data.Verify(); err != nil {
+		return info, azureToObjectError(traceError(err), bucket, object)
```
fwessels: Old code calls `a.client.Deleteblob` in line 594 -- should we not do this?

cmd/gateway-gcs.go
```diff
@@ -747,30 +738,17 @@ func (l *gcsGateway) PutObject(bucket string, key string, size int64, data io.Re
...
 	w.Close()
 
-	// Verify sha256sum after close.
-	if sha256sum != "" {
-		if hex.EncodeToString(sha256Writer.Sum(nil)) != sha256sum {
-			object.Delete(l.ctx)
-			return ObjectInfo{}, traceError(SHA256Mismatch{})
-		}
+	if err := data.Verify(); err != nil { // Verify sha256sum after close.
+		return ObjectInfo{}, gcsToObjectError(traceError(err), bucket, key)
```
fwessels: Old code called `object.Delete(l.ctx)` -- should we not do this?

cmd/gateway-gcs.go
```diff
@@ -855,65 +833,36 @@ func (l *gcsGateway) checkUploadIDExists(bucket string, key string, uploadID str
...
 	w.Close()
 
-	// Verify sha256sum after Close().
-	if sha256sum != "" {
-		if hex.EncodeToString(sha256Writer.Sum(nil)) != sha256sum {
-			object.Delete(l.ctx)
-			return PartInfo{}, traceError(SHA256Mismatch{})
-		}
+	if err := data.Verify(); err != nil {
+		return PartInfo{}, gcsToObjectError(traceError(err), bucket, key)
```
fwessels: Old code calls `object.Delete(l.ctx)` in lien 907 -- should we not do this?

cmd/fs-v1_test.go
aead: Actually correct but I would still keep the new version because the old test don't really test the 0-size content length. It basically return io.EOF immediately on the first right. The internal LimitReader wouldn't be tested.

cmd/fs-v1_test.go
aead: same

cmd/gateway-azure.go
aead: same

cmd/gateway-s3.go
```diff
@@ -330,32 +330,17 @@ func (l *s3Objects) GetObjectInfo(bucket string, object string) (objInfo ObjectI
...
+func (l *s3Objects) PutObject(bucket string, object string, data *HashReader, metadata map[string]string) (objInfo ObjectInfo, err error) {
+	sha256sumBytes, err := hex.DecodeString(data.sha256Sum)
 	if err != nil {
 		return objInfo, s3ToObjectError(traceError(err), bucket, object)
 	}
-	delete(metadata, "etag")
-
-	oi, err := l.Client.PutObject(bucket, object, size, data, md5Bytes, sha256Bytes, toMinioClientMetadata(metadata))
+	md5sumBytes, err := hex.DecodeString(metadata["etag"])
+	if err != nil {
```
fwessels: Old code deletes etag (`delete(metadata, "etag")`), must't we do the same?

aead: Correct, good catch!

cmd/object-api-interface.go
```diff
@@ -55,3 +63,81 @@ type ObjectLayer interface {
 	ListUploadsHeal(bucket, prefix, marker, uploadIDMarker,
 		delimiter string, maxUploads int) (ListMultipartsInfo, error)
 }
+
+// HashReader writes what it reads from an io.Reader to a MD5 and SHA256 hash.Hash.
```
fwessels: a MD5 --> an MD5

cmd/object-api-interface.go
```diff
@@ -55,3 +63,81 @@ type ObjectLayer interface {
...
+	return
+}
+
+// Size returns the absolute number of bytes the HashReader
+// will return during reading. It returns -1 for unlimited
+// data.
+func (r *HashReader) Size() int64 { return r.size }
+
+// MD5 returns the MD5 sum of the processed data. Any
+// further read will change the MD5 sum.
```
fwessels: read --> reads

fwessels (approved):
LGTM, and nice code simplification.

cmd/object-api-interface.go
aead: done

cmd/object-api-interface.go
aead: done

PR #4926, opened 2017-09-15
api/PostPolicy: Allow location header fully qualified URL

### LINKED ISSUE #4910: Allow full domain+path returned in Location header for single-endpoint configurations
When making a file create request against a minio server configuration with one endpoint

## Expected Behavior (as returned by AWS S3)
Location header returns something like this:
 `<Location>https://{configured-domain}/our-bucket/images/1505248844958-happy_turtle.png</Location>`

## Current Behavior
 `<Location>/our-bucket/images/1505248844958-happy_turtle.png</Location>`

## Context
Our current client doesn't support manipulating this, so we have to change our client to add add the domain to the beginning after processing the request.

harshavardhana:
## Description
req.Host is used to construct the final object location.

## Motivation and Context
Fixes #4910

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:

### DIFF
--- a/cmd/api-response.go
+++ b/cmd/api-response.go
@@ -275,9 +275,25 @@ func getLocation(r *http.Request) string {
 	return path.Clean(r.URL.Path) // Clean any trailing slashes.
 }
 
-// getObjectLocation gets the relative URL for an object
-func getObjectLocation(bucketName string, key string) string {
-	return "/" + bucketName + "/" + key
+// returns "https" if the tls boolean is true, "http" otherwise.
+func getURLScheme(tls bool) string {
+	if tls {
+		return httpsScheme
+	}
+	return httpScheme
+}
+
+// getObjectLocation gets the fully qualified URL of an object.
+func getObjectLocation(host, proto, bucket, object string) string {
+	if proto == "" {
+		proto = getURLScheme(globalIsSSL)
+	}
+	u := url.URL{
+		Host:   host,
+		Path:   path.Join(slashSeparator, bucket, object),
+		Scheme: proto,
+	}
+	return u.String()
 }
 
 // generates ListBucketsResponse from array of BucketInfo which can be
--- a/cmd/bucket-handlers.go
+++ b/cmd/bucket-handlers.go
@@ -569,8 +569,10 @@ func (api objectAPIHandlers) PostPolicyBucketHandler(w http.ResponseWriter, r *h
 		return
 	}
 
+	port := r.Header.Get("X-Forward-Proto")
+	location := getObjectLocation(r.Host, port, bucket, object)
 	w.Header().Set("ETag", `"`+objInfo.ETag+`"`)
-	w.Header().Set("Location", getObjectLocation(bucket, object))
+	w.Header().Set("Location", location)
 
 	// Get host and port from Request.RemoteAddr.
 	host, port, err := net.SplitHostPort(r.RemoteAddr)
@@ -603,7 +605,7 @@ func (api objectAPIHandlers) PostPolicyBucketHandler(w http.ResponseWriter, r *h
 			Bucket:   objInfo.Bucket,
 			Key:      objInfo.Name,
 			ETag:     `"` + objInfo.ETag + `"`,
-			Location: getObjectLocation(objInfo.Bucket, objInfo.Name),
+			Location: location,
 		})
 		writeResponse(w, http.StatusCreated, resp, "application/xml")
 	case "200":
--- a/cmd/api-response_test.go
+++ b/cmd/api-response_test.go
@@ -1,0 +1,74 @@
+/*
+ * Minio Cloud Storage, (C) 2017 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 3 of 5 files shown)


harshavardhana:
Adding tests for `getObjectLocation` function, will be adding them soon.

cmd/api-response.go
```diff
@@ -275,9 +275,26 @@ func getLocation(r *http.Request) string {
...
+	if tls {
+		return httpsScheme
+	}
+	return httpScheme
+}
+
+// getObjectLocation gets the fully URI of an object if
+// server is bound to a specific IP or hostname. If not
+// the URI is relative.
+func getObjectLocation(host, bucket, object string) string {
```
balamurugana: We would need to use something like below
```go
u := URL{
Scheme: getURLScheme(),
Host: ...,
Path: ...,
}

return u.String()
```

balamurugana (changes requested):

cmd/api-response.go
harshavardhana: Not as default .. see the reasoning why it's based on `--address` alone. Otherwise we need it as relative..

harshavardhana: @balamurugana can you take look again?

harshavardhana:
ping @donatello waiting on your review as well.

cmd/api-response.go
```diff
@@ -275,9 +275,26 @@ func getLocation(r *http.Request) string {
...
-	return "/" + bucketName + "/" + key
+// returns "https" if the tls boolean is true, "http" otherwise.
+func getURLScheme(tls bool) string {
+	if tls {
+		return httpsScheme
+	}
+	return httpScheme
+}
+
+// getObjectLocation gets the fully URI of an object if
```
donatello: s/fully/fully qualified/

donatello (approved):
LGTM - however commit message needs to mention that the location header is related to post-policy-handler. Otherwise, some context appears to be missing.

balamurugana:
As this PR fixes the issue partially i.e. its hard to provide usable fully qualified URL.  Discussion is needed.

cmd/bucket-handlers.go
```diff
@@ -603,7 +603,7 @@ func (api objectAPIHandlers) PostPolicyBucketHandler(w http.ResponseWriter, r *h
 			Bucket:   objInfo.Bucket,
 			Key:      objInfo.Name,
 			ETag:     `"` + objInfo.ETag + `"`,
-			Location: getObjectLocation(objInfo.Bucket, objInfo.Name),
+			Location: getObjectLocation(globalMinioAddr, objInfo.Bucket, objInfo.Name),
```
krishnasrinivas: globalMinioAddr gets set to `:9000` if no --address is passed

harshavardhana: Correct.. this is waiting on a different architectural problem see above.

harshavardhana:
The PR is now updated to use MINIO_ENDPOINT environment value after discussion @donatello @balamurugana  - please check.

cmd/server-main.go
```diff
@@ -61,6 +61,9 @@ ENVIRONMENT VARIABLES:
   REGION:
      MINIO_REGION: To set custom region. By default it is "us-east-1".
 
+  ENDPOINT:
+     MINIO_ENDPOINT: To set location endpoint of the instance.
```
balamurugana: I think `MINIO_HOST_NAME` would be nice.  However we will discuss and take it forward

harshavardhana: The reason i didn't want it as HOST_NAME is because ENDPOINT encourages providing PORT as well @balamurugana

donatello (approved):

harshavardhana:
@balamurugana we can remove blocked label from this PR since the PR addresses the agreed style we want to address this?

balamurugana:
@harshavardhana @abperiasamy is looking into the environment variable naming. Rest of the code is good to go.

harshavardhana:
> @harshavardhana @abperiasamy is looking into the environment variable naming. Rest of the code is good to go.

I changed to MINIO_ACCESS_ENDPOINT

harshavardhana:
The PR is updated to use `r.Host` as the endpoint and leverage X-Forward-Proto as scheme hint, if not default to what is configured by the server.  Avoids envs and addresses all the concerns.

donatello (approved):

cmd/api-response.go
```diff
@@ -275,9 +275,25 @@ func getLocation(r *http.Request) string {
...
+}
+
+// getObjectLocation gets the fully qualified URL of an object.
+func getObjectLocation(host, proto, bucket, object string) string {
+	if proto == "" {
+		proto = getURLScheme(globalIsSSL)
+	}
+	u := url.URL{
+		Host:   host,
+		Path:   slashSeparator + pathJoin(bucket, object),
```
balamurugana: I think we could use  `path.Join("/", bucket, object)`?

balamurugana (commented):
Minor comment

balamurugana (approved):
LGTM

PR #4936, opened 2017-09-18
Fix signature v2 and presigned query unescaping.

### LINKED ISSUE #4927: [sign-v2] presigned operations do not work anymore with response headers.
## Expected Behavior
Signature V2 works fine with presigned response headers.

## Current Behavior
Signature V2 doesn't work if response headers are set with special character file names.
```go
package main

import (
	"log"
	"net/url"
	"time"

	"github.com/minio/minio-go"
)

func main() {
	// Note: YOUR-ACCESSKEYID, YOUR-SECRETACCESSKEY, my-bucketname and my-objectname
	// are dummy values, please replace them with original values.

	// Requests are always secure (HTTPS) by default. Set secure=false to enable insecure (HTTP) access.
	// This boolean value is the last argument for New().

	// New returns an Amazon S3 compatible client object. API compatibility (v2 or v4) is automatically
	// determined based on the Endpoint value.
	s3Client, err := minio.NewV2("localhost:9000", "USWUXHGYZQYFYFFIT3RE", "MOJRH0mkL1IPauahWITSVvyDrQbEEIwljvmxdq03", false)
	if err != nil {
		log.Fatalln(err)
	}

	// Set request parameters
	reqParams := make(url.Values)
	reqParams.Set("response-content-disposition", "attachment; filename=\"4K%2d4M.txt\"")

	// Gernerate presigned get object url.
	presignedURL, err := s3Client.PresignedGetObject("testbucket", "issue", time.Duration(1000)*time.Second, reqParams)
	if err != nil {
		log.Fatalln(err)
	}
	log.Println(presignedURL)
}
$ go run /tmp/presignedgetobject.go
$ curl "http://localhost:9000/testbucket/issue?AWSAccessKeyId=USWUXHGYZQYFYFFIT3RE&Expires=1505469132&response-content-disposition=attachment%3B%20filename%3D%224K%252d4M.txt%22&Signature=3NuKNqSbpEoGRqPI/JGP7C4K67E%3D"
<Error><Code>SignatureDoesNotMatch</Code><Message>The request signature we calculated does not match the signature you provided. Check your key and signing method.</Message><Key></Key><BucketName></BucketName><Resource>/testbucket/issue</Resource><RequestId>3L137</RequestId><HostId>3L137</HostId></Error>
```

## Possible Solution
[Fix signature v2 and presigned query unexcaping](https://github.com/minio/minio/pull/4916#discussion_r139101174)

Already mentioned above.

## Context
AWS S3 compliance.

## Your Environment
master version.

harshavardhana:
## Description
Fix signature v2 and presigned query unescaping.

Simplifies the testing code by using s3signer
package from minio-go library.

## Motivation and Context
Fixes #4927

## How Has This Been Tested?
Manually using `go test` 

## Types of changes

## Checklist:

### DIFF
--- a/cmd/bucket-handlers_test.go
+++ b/cmd/bucket-handlers_test.go
@@ -455,7 +455,6 @@ func testListMultipartUploadsHandler(obj ObjectLayer, instanceType, bucketName s
 
 		// verify response for V2 signed HTTP request.
 		reqV2, err := newTestSignedRequestV2("GET", u, 0, nil, testCase.accessKey, testCase.secretKey)
-
 		if err != nil {
 			t.Fatalf("Test %d: %s: Failed to create HTTP request for PutBucketPolicyHandler: <ERROR> %v", i+1, instanceType, err)
 		}
--- a/cmd/object-handlers_test.go
+++ b/cmd/object-handlers_test.go
@@ -2813,6 +2813,7 @@ func testAPIPutObjectPartHandlerPreSign(obj ObjectLayer, instanceType, bucketNam
 		t.Fatalf("[%s] - Failed to create an unsigned request to put object part for %s/%s <ERROR> %v",
 			instanceType, bucketName, testObject, err)
 	}
+
 	err = preSignV2(req, credentials.AccessKey, credentials.SecretKey, int64(10*60*60))
 	if err != nil {
 		t.Fatalf("[%s] - Failed to presign an unsigned request to put object part for %s/%s <ERROR> %v",
@@ -3301,6 +3302,7 @@ func testAPIListObjectPartsHandlerPreSign(obj ObjectLayer, instanceType, bucketN
 			instanceType, bucketName, mpartResp.UploadID)
 	}
 
+	req.Header = http.Header{}
 	err = preSignV2(req, credentials.AccessKey, credentials.SecretKey, int64(10*60*60))
 	if err != nil {
 		t.Fatalf("[%s] - Failed to presignV2 an unsigned request to list object parts for bucket %s, uploadId %s",
--- a/cmd/signature-v2.go
+++ b/cmd/signature-v2.go
@@ -77,6 +77,20 @@ func doesPolicySignatureV2Match(formValues http.Header) APIErrorCode {
 	return ErrNone
 }
 
+// Escape encodedQuery string into unescaped list of query params, returns error
+// if any while unescaping the values.
+func unescapeQueries(encodedQuery string) (unescapedQueries []string, err error) {
+	for _, query := range strings.Split(encodedQuery, "&") {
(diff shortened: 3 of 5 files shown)


krishnasrinivas (commented):
https://github.com/krishnasrinivas/minio/commit/e258cb647c1ad6bcf2600ceff6727eeca7bcce1f

would be a simpler fix. i.e do the unescaping at one place - `canonicalizedResourceV2()`

harshavardhana:
> krishnasrinivas/minio@e258cb6
> 
> would be a simpler fix. i.e do the unescaping at one place - canonicalizedResourceV2()

@krishnasrinivas it is better to comment on the PR we have decided to not do parallel coding in different branches. Unless its a big change to show - it is not conducive to the environment of review process.

harshavardhana:
> krishnasrinivas/minio@e258cb6
> 
> would be a simpler fix. i.e do the unescaping at one place - canonicalizedResourceV2()

Also this is incorrect @krishnasrinivas we have to Unescape because of the issue previously reproduced by @vadmeste 

```
commit 52d8f564bfb8fb8cda4953e0051fb196718524b7
Author: Anis Elleuch <vadmeste@gmail.com>
Date:   Tue Apr 4 01:55:14 2017 +0100

    sigv2: Unespace canonicalized resources values (#4034)
    
    Values of canonicalized query resources should be unescaped before calculating
    the signature. This bug is not noticed before because partNumber and uploadID
    values in Minio doesn't have characters that need to be escaped.
```

harshavardhana:
Interestingly while using s3signer.SignV2 found bugs in our minio-go implementation as well, which should be fixed separately.

krishnasrinivas:
> Also this is incorrect @krishnasrinivas we have to Unescape because of the issue previously reproduced by @vadmeste

That is right, what I was pointing out was we can keep code simpler by doing the unescaping at one place instead of two. i.e canonicalizedResourceV2() already unescapes the query parameter. We can keep it there and remove the unescaping code in the calling functions.

harshavardhana:
> That is right, what I was pointing out was we can keep code simpler by doing the unescaping at one place instead of two. i.e canonicalizedResourceV2() already unescapes the query parameter. We can keep it there and remove the unescaping code in the calling functio

Unescaping is done earlier to fail early and properly. We can further combine these checks.. let me do them as well.

harshavardhana:
@vadmeste @donatello can i get reviews?

vadmeste (approved):
LGTM & tested with some unicode and special characters

donatello (approved):
LGTM

harshavardhana:
PR is updated @krishnasrinivas

krishnasrinivas (approved):

PR #4942, opened 2017-09-19
Remove list multipart uploads for XL

fwessels:
Bring XL in sync with FS by removing ListMultipartUploads and ListObjectParts

## Types of changes

## Checklist:


vadmeste (approved):
LGTM & tested

harshavardhana:
As discussed yesterday with @abperiasamy // cc @deekoder @krishnasrinivas - we are not going to remove this instead we are going to simplify the backend to support 

- ListMultipartUploads()
- ListObjectParts()

Also the cleanup of stale parts will be done every week which needs to be brought into erasure code mode as well.  Gateway requirements are different and will have their own backend implementations for now there are no further changes needed there as agreed.

PR #4943, opened 2017-09-19
Provide the correct free block size volume/disk information

tuhtah:
## Description

On *NIX platforms the statfs(2) system call returns a struct containing both the
free blocks in the filesystem (Statfs_t.Bfree) and the free blocks available to
the unprivileged or non-superuser (Statfs_t.Bavail).

The `Bfree` and `Bavail` fields (with `Bfree >= Bavail`) will be set to
different values on e.g. filesystems such as ext4 that reserve a certain
percentage of the filesystem blocks which may only be allocated by admnistrative
privileged processes.

The calculations for the `Total` disk space need to subtract the difference
between the `Bfree` and `Bavail` fields for it to correctly show the total
available storage space available for unprivileged users.

This implicitly fixes a bug where the `Used = Total - Free` calculation yielded
different (and also incorrect) results for identical contents stored when only
the sizes of the disks or backing volumes differed. (as can be witnessed in the
`Used:` value displayed in the Minio browser)

See:
- https://wiki.archlinux.org/index.php/ext4#Reserved_blocks
- http://man7.org/linux/man-pages/man2/statfs.2.html
- https://man.openbsd.org/statfs
- http://lingrok.org/xref/coreutils/src/df.c#893

## Motivation and Context

When mirroring the exact contents of 2 Minio S3 instances I noticed that the output of the `Used:` Information in the Minio Browser differed when the backing storage sizes were different (e.g. 100G vs 500G). 

After verifying that the contents were transmitted correctly to the 2nd instance with the smaller backing storage and finding them identical, the output of a `du -hs /export` also showed the exact same used count.

Suspecting initially a bug in the calculation in the user interface, I finally found that the root cause was actually the incorrect usage of the _Bavail_ field instead of the _Bfree_ field as explained in the commit message.

Validating in how the `df` utility calculates its displayed _Used_ value, I devised the contributed patch.

## How Has This Been Tested?

I built a Linux minio binary (CGO_ENABLED) using the patch from this PR using _Go 1.9_. I uploaded and tested it on both affected Minio setups where I then visually verified that the number of the `Used: ` output in the Minio Browser matches the _Used_  column of a `df -h /export` invocation.

## Types of changes

**N.B.:** This might subtly break external tests if they relied on the previous wrong value using e.g. the StorageInfo JSON-RPC call. 

## Checklist:

I couldn't find any information on _StorageInfo_ in the documentation so this change should not affect it.

I solely compiled and tested on Linux, but applied the fix for both BSD/Darwin (same field name as on Linux) and the OpenBSD flavor (F_bfree).

However, on both BSD platforms the typecast to _uint64_ is not strictly necessary anymore as the _Bfree_ and _F_bfree_ fields are of type _uint64_ different to their _Bavail_ fields that were of type _int64_, see:

- https://golang.org/src/syscall/ztypes_linux_amd64.go?#L123
- https://golang.org/src/syscall/ztypes_freebsd_amd64.go?#L103 
- https://golang.org/src/syscall/ztypes_openbsd_amd64.go?#L102

I tested with _Go 1.9_ where I ran into an (I believe) unrelated Go runtime specific implementation change. Negative offset errors are now seemingly reported more specifically from earlier Go versions and use `readat` as the operation (see: https://golang.org/src/os/file.go?h=negative+offset#L107):

```patch
diff --git a/cmd/posix_test.go b/cmd/posix_test.go
index d366fce7..91b87f74 100644
--- a/cmd/posix_test.go
+++ b/cmd/posix_test.go
@@ -20,6 +20,7 @@ import (
        "bytes"
        "crypto/rand"
        "encoding/hex"
+       "errors"
        "fmt"
        "io"
        "io/ioutil"
@@ -1030,9 +1031,9 @@ func TestPosixReadFile(t *testing.T) {
                                        }
                                }
                                return &os.PathError{
-                                       Op:   "read",
+                                       Op:   "readat",
                                        Path: slashpath.Join(path, "success-vol", "myobject"),
-                                       Err:  os.ErrInvalid,
+                                       Err:  errors.New("negative offset"),
                                }
                        }(),
                },
```

### DIFF
--- a/pkg/disk/stat_bsd.go
+++ b/pkg/disk/stat_bsd.go
@@ -29,11 +29,13 @@ func GetInfo(path string) (info Info, err error) {
 	if err != nil {
 		return Info{}, err
 	}
-	info = Info{}
-	info.Total = uint64(s.Bsize) * uint64(s.Blocks)
-	info.Free = uint64(s.Bsize) * uint64(s.Bavail)
-	info.Files = uint64(s.Files)
-	info.Ffree = uint64(s.Ffree)
-	info.FSType = getFSType(s.Fstypename)
+	fsReservedBlocks := uint64(s.Bfree) - uint64(s.Bavail)
+	info = Info{
+		Total:  uint64(s.Bsize) * (uint64(s.Blocks) - fsReservedBlocks),
+		Free:   uint64(s.Bsize) * uint64(s.Bavail),
+		Files:  uint64(s.Files),
+		Ffree:  uint64(s.Ffree),
+		FSType: getFSType(s.Fstypename),
+	}
 	return info, nil
 }
--- a/pkg/disk/stat_linux.go
+++ b/pkg/disk/stat_linux.go
@@ -29,11 +29,13 @@ func GetInfo(path string) (info Info, err error) {
 	if err != nil {
 		return Info{}, err
 	}
-	info = Info{}
-	info.Total = uint64(s.Bsize) * uint64(s.Blocks)
-	info.Free = uint64(s.Bsize) * uint64(s.Bavail)
-	info.Files = uint64(s.Files)
-	info.Ffree = uint64(s.Ffree)
-	info.FSType = getFSType(int64(s.Type))
+	fsReservedBlocks := uint64(s.Bfree) - uint64(s.Bavail)
+	info = Info{
+		Total:  uint64(s.Bsize) * (uint64(s.Blocks) - fsReservedBlocks),
+		Free:   uint64(s.Bsize) * uint64(s.Bavail),
+		Files:  uint64(s.Files),
+		Ffree:  uint64(s.Ffree),
+		FSType: getFSType(int64(s.Type)),
+	}
 	return info, nil
 }
--- a/pkg/disk/stat_openbsd.go
+++ b/pkg/disk/stat_openbsd.go
@@ -29,11 +29,13 @@ func GetInfo(path string) (info Info, err error) {
 	if err != nil {
 		return Info{}, err
 	}
-	info = Info{}
-	info.Total = uint64(s.F_bsize) * uint64(s.F_blocks)
-	info.Free = uint64(s.F_bsize) * uint64(s.F_bavail)
-	info.Files = uint64(s.F_files)
-	info.Ffree = uint64(s.F_ffree)
-	info.FSType = getFSType(s.F_fstypename)
+	fsReservedBlocks := uint64(s.F_bfree) - uint64(s.F_bavail)
+	info = Info{
+		Total:  uint64(s.F_bsize) * (uint64(s.F_blocks) - fsReservedBlocks),
+		Free:   uint64(s.F_bsize) * uint64(s.F_bavail),
+		Files:  uint64(s.F_files),
+		Ffree:  uint64(s.F_ffree),
+		FSType: getFSType(s.F_fstypename),
+	}
 	return info, nil
 }


tuhtah:
Forgot to mention, I have not checked if the Windows implementation and code path exhibits the same problem as with the *NIX variants.

harshavardhana:
I am not sure if this is a correct fix @tuhtah you are essentially reverting the changes done in previous PR https://github.com/minio/minio/pull/2788  and the relevant bug that it fixes https://github.com/minio/minio/issues/2780

Can you shed some light on why is this a correct fix?  // cc @vadmeste

tuhtah:
@harshavardhana thx for taking a look and pointing out the previous PR, I wasn't aware of it.

At a first quick glance though, something seems off with the current code in posix.go which assumes a [hardcoded 5% value](https://github.com/vadmeste/minio/blob/master/cmd/posix.go#L168) that gets subtracted from the total value to check if there is enough space.

Afaiu, this is a filesystem tuneable and can be set to any value, where the default for _ext4_ happens to be 5%. (see: `-m reserved-blocks-percentage` in https://linux.die.net/man/8/tune2fs)

For XFS reserved blocks can be set through the xfs_io(8) command (see: https://linux.die.net/man/8/xfs_io), and its semantics are apparently different to ext4: https://www.spinics.net/lists/xfs/msg30776.html

The UFS2 filesystem on FreeBSD seems to default to an 8% value (see: `-m minfree` in https://www.freebsd.org/cgi/man.cgi?query=tunefs&sektion=8), OpenBSD again has a default of 5% (see: https://man.openbsd.org/newfs.8#m)

And for a filesystem such as ZFS reserved space becomes even harder to determine: https://github.com/zfsonlinux/zfs/issues/3014

Hmm, also how much space is effectively available to _minio_ in docker containers when it runs a _root_ there and [user namespacing](http://man7.org/linux/man-pages/man7/user_namespaces.7.html) is not in effect?

But I believe you are very right, reverting the fix from the PR alone can't be the whole story.

It could very well be that the current implementation on the server side is correct and the Disk usage information on the client needs to be adapted, let's see.

I will read through the previous discussion that motivated the fix and study the PR and come back to you later with a suggestion or a better reproducible description of the problem I am experiencing.

tuhtah:
@harshavardhana here is a gist that contains a typescript of a terminal session that hopefully makes it easier to demonstrate the problem I am seeing: 
- https://gist.github.com/tuhtah/345eaef2e46fad7a4a2d8efccff5be3c

There I create 2 separate loopback filesystems formatted with ext4, one 50G and another larger 100G. 

After everything has been setup and Minio started on an empty filesystem, the _Minio Browser_ then reports the following:
- 50G: Used: **2.57** GB Free: 46.40 GB
- 100G: Used: **5.07** GB Free: 92.86 GB

Where I believe the Used values should display rather:
- 50G: **0.05** GB
- 100G: **0.06** GB

I created a minimal example in Go (statfs.go) that I used in the shell session transcript and which exercises the relevant APIs:

- https://gist.github.com/tuhtah/b04fcd9b8225085f33e2b9a8bd402877

The minimal example has been extracted and put together from the Minio code found in _pkg/disk/_. It outputs both calculations for comparison once with _Statfs_t.Bfree_ and another with _Statfs_t.Bavail_.

I am currently wondering if the proper fix wouldn't better include both Statfs_t fields, i.e. _Statfs_t.Bfree_ and _Statfs_t.Bavail_ and then use the right value where it is needed. (the PR you referenced seems to depend on _Statfs_t.Bavail_ and the Minio Browser seems to need _Statfs_t.Bfree_)

But before I continue down this road, can you reproduce my problem and confirm that you agree that the output from the Minio Browser needs to actually get fixed?

vadmeste:
Thanks for your PR @tuhtah,

I wanted to show users the exact amount of free space that users have. That's why I introduced `Bavail` instead instead of `BFree`.

Now, let's understand what's going on exactly,

You see different 'Used' value in Minio Browser for two different  disks which contain the same data, right? and the size of these disks are different, right ? 

Then I think the real problem here is the calculation of `info.Total`, since `Used = Total - Free`

What happened is that each of your disks has a different number of blocks representing the reserved area of the disk.

So I think we need to fix `info.Total`. I think if we substract the reserved blocks number for info.Total, further calculation will be correct.

```
reservedBlocks = s.BFree-s.Bavail
info.Total = uint64(s.Bsize) * (uint64(s.Blocks-reservedBlocks))
```

I didn't test that actually @tuhtah :) 

Can you do so and update us ?

tuhtah:
@vadmeste Seems we just posted about the same time, I added a shell transcript that demonstrates the problem I am seeing! ;)

To answer your question: Yes, if you start 2 separate Minio instances that have 2 different disk sizes then you will get different "Used" values (it's being already displayed wrongly when we start empty with them as demonstrated in the gist).

I also felt that something might be additionally fishy with the `info.Total` calculation, they don't match up exactly with what one would see for a `df -h /export` invocation.

I will go ahead and test out your suggested `info.Total` fix and report back, looks like it could do the trick... ;)

tuhtah:
@vadmeste Applying your fix for `info.Total` to my statfs.go example yields also the correct value.

```shell
$ go run statfs.go /storage/minio/mnt/minio_50G
{Total:49874960384 Free:52521521152 Avail:49820389376 Files:3276800 Ffree:3276784 FSType:EXT4}

Bsize: 4096
Blocks: 12835960
---
Total: 46.45
Free: 48.91
Avail: 46.40
Used: -2.46
Used (avail): 0.05
Files: 3276800
Ffree: 3276784
```
But I am still not entirely sure what the correct thing would be to do, i.e. should we include and pass both fields _Bfree_ and _Bavail_ in the info struct and not tamper with the Total value or do the calculation right away.

In other words should the output mirror the way the coreutils `df -h` utility calculates them or not?
I would still have to check how the _Total_ value is calculated there, it doesn't seem to subtract the reserved blocks (in the Size field):

```shell
$ df -h /storage/minio/mnt/minio_50G 
Filesystem      Size  Used Avail Use% Mounted on
/dev/loop0       49G   53M   47G   1% /storage/minio/mnt/minio_50G
```

What do you think?

vadmeste:
@tutah, I am not sure what df wants to show.

But for Minio, I think it is more correct to show the space for unprivileged users since, most of the time, Minio will be run as a non-root.

tuhtah:
Thinking about it a bit longer, I actually like the fix and suggested approach by @vadmeste 

First of all it addresses what a user actually would be interested in: How much storage do I actually really have available?

The _reserved blocks_ information is an operational/administrative setting and concern. If we needed to get at it in Minio, we would need to do this over a specialized API (see my earlier tune2fs ramblings) for different filesystems.

@vadmeste's approach side steps all this complexity actually quite nicely.

And if it is running in docker as root (at which point it could actually use those 5% reserved blocks), then because Minio sees a different lower total value, we again don't run danger to use up that space.

If you all concur with this, then I would proceed by incorporating @vadmeste's fix and update the PR in this regard.

pkg/disk/stat_bsd.go
```diff
@@ -30,7 +30,7 @@ func GetInfo(path string) (info Info, err error) {
 		return Info{}, err
 	}
 	info = Info{}
-	info.Total = uint64(s.Bsize) * uint64(s.Blocks)
+	info.Total = uint64(s.Bsize) * (uint64(s.Blocks) - (uint64(s.Bfree) - uint64(s.Bavail)))
```
vadmeste: Can you add another variable `fsReservedBlocks := uint64(s.Bfree) - uint64(s.Bavail)` and use it ?

That will make easier to understand the code

tuhtah:
@harshavardhana @vadmeste I have rebased, squashed, updated and "perfumed" the PR to calculate the `Total` taking the `Bfree` and `Bavail` fields into account according to our previous discussion.

I have retested this on my setup using ext4 and the `Used:` values are still not identical but are now correctly calculated and displayed.

That they still slightly differ (which might trip others and is confusing) is to be expected as e.g. on ext4 even a freshly created empty filesystem starts with having blocks allocated already for the filesysystem's bookkeeping structures such as the journal. This initial value is a percentage of the filesystem and will therefore vary with the size of the created filesystem. (you can see this in the `df` outputs of the [gist I posted](https://gist.github.com/tuhtah/345eaef2e46fad7a4a2d8efccff5be3c))

Still, I believe this is the correct fix now and ready from my side for review and merging. Kudos to @vadmeste for his help in tracking down the real root cause and suggesting the correct fix.

I have locally rerun the Minio unit tests and btw., I decided against removing the _uint64()_ type casts where they weren't strictly necessary, in fact I have added 2 more of them explicitly around the `Bfree` and `Bavail` fields in the newly added code. The types in the Statfs_t struct for the fields we use vary across different platforms and architectures, so a removal would have probably broken something in the end and does otherwise no harm while keeping the code consistent and easier to diff across platforms.

pkg/disk/stat_bsd.go
tuhtah: Sounds good, while we are at it, will also simplify the Info struct initialization through the help of a [composite literal](https://golang.org/doc/effective_go.html#composite_literals).

tuhtah: Done, also squashed all of it back into a single commit.
Quickly retesting it one more time to make sure I didn't mess up something.

tuhtah: Retested again on Linux and also built it and took it for a spin on FreeBSD. Looks good on my side.

vadmeste:
@tuhtah, the `used` field means disk usage and not how much space a Minio storage consumes.. so sometimes, it is not useful to compare Used value of two different disks, however this PR is crucial especially for large disks (5% of 1 TB is 50 GB).

vadmeste (approved):
LGTM

harshavardhana (approved):

PR #4944, opened 2017-09-20
fs: optimize multipart clean work

fangyuxiangGL:
when multipart clean auto run or we use abort multipart explicitly, data part file and uploads.json will be cleaned, but some directory will left.

e.g object named A will leave a directory /path/to/bucket/A/
object named A/B/C/D will leave directories 
```
 /path/to/bucket/A/
 /path/to/bucket/A/B/
 /path/to/bucket/A/B/C/
 /path/to/bucket/A/B/C/D/
```

those residual directories will cause performance degraded as time goes..
the goal is to clean up all of the residual directories.

## Description
When do multipart clean work, also clean directory brought in by multipart upload

## Motivation and Context
Fixing multipart cleanup.

## How Has This Been Tested?
Already do test to verify this.
e.g if we have multipart residual as bellow, which is caused by upload failed of object A/B/C/D and A/BC/E
```
/path/to/bucket/A/B/C/D/
/path/to/bucket/A/B/C/E/
```

Result:
there will be nothing under /path/to/bucket/ once clean work is triggered 

## Types of changes

## Checklist:

### DIFF
--- a/cmd/fs-v1-multipart.go
+++ b/cmd/fs-v1-multipart.go
@@ -62,11 +62,11 @@ func (fs fsObjects) deleteUploadsJSON(bucket, object, uploadID string) error {
 
 // Removes the uploadID, called either by CompleteMultipart of AbortMultipart. If the resuling uploads
 // slice is empty then we remove/purge the file.
-func (fs fsObjects) removeUploadID(bucket, object, uploadID string, rwlk *lock.LockedFile) error {
+func (fs fsObjects) removeUploadID(bucket, object, uploadID string, rwlk *lock.LockedFile) (bool, error) {
 	uploadIDs := uploadsV1{}
 	_, err := uploadIDs.ReadFrom(rwlk)
 	if err != nil {
-		return err
+		return false, err
 	}
 
 	// Removes upload id from the uploads list.
@@ -75,12 +75,12 @@ func (fs fsObjects) removeUploadID(bucket, object, uploadID string, rwlk *lock.L
 	// Check this is the last entry.
 	if uploadIDs.IsEmpty() {
 		// No more uploads left, so we delete `uploads.json` file.
-		return fs.deleteUploadsJSON(bucket, object, uploadID)
+		return true, fs.deleteUploadsJSON(bucket, object, uploadID)
 	} // else not empty
 
 	// Write update `uploads.json`.
 	_, err = uploadIDs.WriteTo(rwlk)
-	return err
+	return false, err
 }
 
 // Adds a new uploadID if no previous `uploads.json` is
@@ -656,6 +656,7 @@ func (fs fsObjects) CompleteMultipartUpload(bucket string, object string, upload
 	}
 
 	uploadIDPath := pathJoin(bucket, object, uploadID)
+	var removeObjectDir bool
 
 	// Hold the lock so that two parallel complete-multipart-uploads
 	// do not leave a stale uploads.json behind.
@@ -663,7 +664,15 @@ func (fs fsObjects) CompleteMultipartUpload(bucket string, object string, upload
 	if err = objectMPartPathLock.GetLock(globalOperationTimeout); err != nil {
 		return oi, err
 	}
-	defer objectMPartPathLock.Unlock()
+
+	defer func() {
+		if removeObjectDir {
+			basePath := pathJoin(fs.fsPath, minioMetaMultipartBucket, bucket)
+			derr := fsDeleteFile(basePath, pathJoin(basePath, object))
+			errorIf(derr, "unable to remove %s in %s", pathJoin(basePath, object), basePath)
+		}
+		objectMPartPathLock.Unlock()
+	}()
 
 	fsMetaPathMultipart := pathJoin(fs.fsPath, minioMetaMultipartBucket, uploadIDPath, fsMetaJSONFile)
 	rlk, err := fs.rwPool.Open(fsMetaPathMultipart)
@@ -830,7 +839,8 @@ func (fs fsObjects) CompleteMultipartUpload(bucket string, object string, upload
 	}
 
 	// Remove entry from `uploads.json`.
-	if err = fs.removeUploadID(bucket, object, uploadID, rwlk); err != nil {
+	removeObjectDir, err = fs.removeUploadID(bucket, object, uploadID, rwlk)
+	if err != nil {
 		return oi, toObjectErr(err, minioMetaMultipartBucket, pathutil.Join(bucket, object))
 	}
 
@@ -865,6 +875,7 @@ func (fs fsObjects) AbortMultipartUpload(bucket, object, uploadID string) error 
 	}
 
 	uploadIDPath := pathJoin(bucket, object, uploadID)
+	var removeObjectDir bool
 
 	// Hold the lock so that two parallel complete-multipart-uploads
 	// do not leave a stale uploads.json behind.
@@ -873,7 +884,15 @@ func (fs fsObjects) AbortMultipartUpload(bucket, object, uploadID string) error 
 	if err := objectMPartPathLock.GetLock(globalOperationTimeout); err != nil {
 		return err
 	}
-	defer objectMPartPathLock.Unlock()
+
+	defer func() {
+		if removeObjectDir {
+			basePath := pathJoin(fs.fsPath, minioMetaMultipartBucket, bucket)
+			derr := fsDeleteFile(basePath, pathJoin(basePath, object))
+			errorIf(derr, "unable to remove %s in %s", pathJoin(basePath, object), basePath)
+		}
+		objectMPartPathLock.Unlock()
+	}()
 
 	fsMetaPath := pathJoin(fs.fsPath, minioMetaMultipartBucket, uploadIDPath, fsMetaJSONFile)
 	if _, err := fs.rwPool.Open(fsMetaPath); err != nil {
@@ -910,7 +929,8 @@ func (fs fsObjects) AbortMultipartUpload(bucket, object, uploadID string) error 
 	}
 
 	// Remove entry from `uploads.json`.
-	if err = fs.removeUploadID(bucket, object, uploadID, rwlk); err != nil {
+	removeObjectDir, err = fs.removeUploadID(bucket, object, uploadID, rwlk)
+	if err != nil {
 		return toObjectErr(err, bucket, object)
 	}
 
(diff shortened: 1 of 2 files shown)


balamurugana:
@fangyuxiangGL you could fix gofmt errors

cmd/fs-v1-multipart.go
```diff
@@ -910,7 +930,8 @@ func (fs fsObjects) AbortMultipartUpload(bucket, object, uploadID string) error
 	}
 
 	// Remove entry from `uploads.json`.
-	if err = fs.removeUploadID(bucket, object, uploadID, rwlk); err != nil {
+        removeObjectDir, err = fs.removeUploadID(bucket, object, uploadID, rwlk)
```
balamurugana: As `removeObjectDir` is set only here, you could move out of usage from `defer` and have the logic here.

cmd/fs-v1-multipart.go
```diff
@@ -830,7 +842,8 @@ func (fs fsObjects) CompleteMultipartUpload(bucket string, object string, upload
 	}
 
 	// Remove entry from `uploads.json`.
-	if err = fs.removeUploadID(bucket, object, uploadID, rwlk); err != nil {
+	removeObjectDir, err= fs.removeUploadID(bucket, object, uploadID, rwlk)
```
balamurugana: As `removeObjectDir` is set only here, you could move out of usage from `defer` and have the logic here.  You could check below comments for this pattern of logic

cmd/fs-v1-multipart.go
```diff
@@ -656,14 +661,21 @@ func (fs fsObjects) CompleteMultipartUpload(bucket string, object string, upload
...
 	// do not leave a stale uploads.json behind.
 	objectMPartPathLock := globalNSMutex.NewNSLock(minioMetaMultipartBucket, pathJoin(bucket, object))
 	if err = objectMPartPathLock.GetLock(globalOperationTimeout); err != nil {
 		return oi, err
 	}
-	defer objectMPartPathLock.Unlock()
+
+        defer func() {
+		objectMPartPathLock.Unlock()
+		if removeObjectDir == true {
```
krisis: Suggestion:
The following is more idiomatic Go.
```go
if removeObjectDir {
}
```

cmd/fs-v1-multipart.go
```diff
@@ -873,7 +887,13 @@ func (fs fsObjects) AbortMultipartUpload(bucket, object, uploadID string) error
 	if err := objectMPartPathLock.GetLock(globalOperationTimeout); err != nil {
 		return err
 	}
-	defer objectMPartPathLock.Unlock()
+
+        defer func() {
+		objectMPartPathLock.Unlock()
+		if removeObjectDir == true {
```
krisis: Same suggestion as above.

krisis (changes requested):
Overall changes LGTM. Thanks for the PR. I have couple of minor suggestions.

cmd/fs-v1-multipart.go
```diff
@@ -865,6 +878,7 @@ func (fs fsObjects) AbortMultipartUpload(bucket, object, uploadID string) error
 	}
 
 	uploadIDPath := pathJoin(bucket, object, uploadID)
+	var removeObjectDir bool
```
balamurugana, line 878: You could move this declaration where it is used.

cmd/fs-v1-multipart.go
```diff
@@ -60,13 +60,18 @@ func (fs fsObjects) deleteUploadsJSON(bucket, object, uploadID string) error {
...
 
+func (fs fsObjects) deleteObjectUploadsDir(bucket, object string) error {
+	basePath := pathJoin(fs.fsPath, minioMetaMultipartBucket, bucket)
+	return fsDeleteFile(basePath, pathJoin(basePath, object))
+}
+
 // Removes the uploadID, called either by CompleteMultipart of AbortMultipart. If the resuling uploads
 // slice is empty then we remove/purge the file.
-func (fs fsObjects) removeUploadID(bucket, object, uploadID string, rwlk *lock.LockedFile) error {
+func (fs fsObjects) removeUploadID(bucket, object, uploadID string, rwlk *lock.LockedFile) (bool, error) {
```
balamurugana: I think this method could directly use logic in `deleteObjectUploadsDir()` in `defer` lilke 
```go
defer func() {
  if (err == nil) {
    basePath := pathJoin(fs.fsPath, minioMetaMultipartBucket, bucket)
    derr := fsDeleteFile(basePath, pathJoin(basePath, object))
    errorIf(derr, "unable to remove %s in %s", pathJoin(basePath, object), basePath)
  }
}
```

balamurugana (changes requested):

cmd/fs-v1-multipart.go
```diff
@@ -60,13 +60,18 @@ func (fs fsObjects) deleteUploadsJSON(bucket, object, uploadID string) error {
...
 
+func (fs fsObjects) deleteObjectUploadsDir(bucket, object string) error {
+	basePath := pathJoin(fs.fsPath, minioMetaMultipartBucket, bucket)
+	return fsDeleteFile(basePath, pathJoin(basePath, object))
+}
+
 // Removes the uploadID, called either by CompleteMultipart of AbortMultipart. If the resuling uploads
 // slice is empty then we remove/purge the file.
-func (fs fsObjects) removeUploadID(bucket, object, uploadID string, rwlk *lock.LockedFile) error {
+func (fs fsObjects) removeUploadID(bucket, object, uploadID string, rwlk *lock.LockedFile) (error, bool) {
```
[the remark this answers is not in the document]
fangyuxiangGL: @balamurugana 

did u mean this? already addressed

cmd/fs-v1-multipart.go
fangyuxiangGL: @balamurugana 

the reason why I put deleteObjectUploadsDir is that we must do it after call objectMPartPathLock.Unlock(), or it will fail

cmd/fs-v1-multipart.go
fangyuxiangGL: @krisis  addressed

cmd/fs-v1-multipart.go
fangyuxiangGL: @krisis  addressed

cmd/fs-v1-multipart.go
fangyuxiangGL: addressed

cmd/fs-v1-multipart.go
balamurugana: A lock is required to do any operation else it would lead to unprotected operation.  Could you tell me what failure you see?

fangyuxiangGL: @balamurugana 

Hi, you are right, it has nothing to do with `objectMPartPathLock.Unlock()`, but it has something to do with `defer rwlk.Close()` in codes bellow it.

we must delete the Dirs after call `defer rwlk.Close()`

balamurugana: @fangyuxiangGL Yep. `rwlk.Close()` needs to be done prior to `objectMPartPathLock.Unlock()`.

cmd/fs-v1-multipart.go
```diff
@@ -75,12 +75,12 @@ func (fs fsObjects) removeUploadID(bucket, object, uploadID string, rwlk *lock.L
 	// Check this is the last entry.
 	if uploadIDs.IsEmpty() {
 		// No more uploads left, so we delete `uploads.json` file.
-		return fs.deleteUploadsJSON(bucket, object, uploadID)
+		return true, fs.deleteUploadsJSON(bucket, object, uploadID)
```
balamurugana, line 78: I see there is no need to return `(bool, error)` because  `bool` is exactly the same result of `err == nil`

fangyuxiangGL: @balamurugana 

don't u think we need a explicit flag to express uploadIDs is empty?

fangyuxiangGL: @balamurugana 

even `err == nil`， we have two kinds of return:
1. false, nil -- has any uploads left
2. true, nil -- has no uploads left

logically, I just think that we should clean up the Dirs with result 2. Of course, we can also clean force with no matter what it returned until we got failed because of non empty Dirs.

krisis (approved):

cmd/fs-v1-multipart.go
balamurugana: I was thinking about why not `removeUploadID()` (or a cleanup function) is not doing the object directory cleanup than handling at individual methods.  As we pass `rwlk` to `removeUploadID()` (which is not clean), this prevents doing that.

As this kind of fix is not the scope of this PR, we could take this PR in.

balamurugana (commented):
@fangyuxiangGL you would need to fix travis ci failures.

fangyuxiangGL:
@balamurugana 

will do it, but I just can't open the detail page

balamurugana:
@fangyuxiangGL click on `Details` link placed after `Required`

fangyuxiangGL:
@balamurugana 

why is it so hard to fix?

harshavardhana:
```diff
--- /dev/fd/63	2017-09-27 08:26:45.855478165 +0000
+++ /dev/fd/62	2017-09-27 08:26:45.855478165 +0000
@@ -1,21 +0,0 @@
diff cmd/fs-v1-multipart.go gofmt/cmd/fs-v1-multipart.go
--- /tmp/gofmt953516395	2017-09-27 08:26:46.027566473 +0000
+++ /tmp/gofmt935663566	2017-09-27 08:26:46.027566473 +0000
@@ -672,7 +672,7 @@
			errorIf(derr, "unable to remove %s in %s", pathJoin(basePath, object), basePath)
 		}
 		objectMPartPathLock.Unlock()
-	} ()
+	}()
 
 	fsMetaPathMultipart := pathJoin(fs.fsPath, minioMetaMultipartBucket, uploadIDPath, fsMetaJSONFile)
 	rlk, err := fs.rwPool.Open(fsMetaPathMultipart)
@@ -892,7 +892,7 @@
 			errorIf(derr, "unable to remove %s in %s", pathJoin(basePath, object), basePath)
 		}
 		objectMPartPathLock.Unlock()
-	} ()
+	}()
 
 	fsMetaPath := pathJoin(fs.fsPath, minioMetaMultipartBucket, uploadIDPath, fsMetaJSONFile)
 	if _, err := fs.rwPool.Open(fsMetaPath); err != nil {
```

@fangyuxiangGL  please fix the formatting issues using `gofmt`

fangyuxiangGL:
@harshavardhana @balamurugana 

thanks for your remind! addressed

balamurugana (approved):

PR #4872, opened 2017-08-30
fix build failure for go1.9

balamurugana:
## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #4945, opened 2017-09-20
travis: update go versions

dvrkps:


harshavardhana:
Thanks for the PR @dvrkps we are not upgrading to go1.9.x yet, there is a fix available already in this regard - https://github.com/minio/minio/pull/4872

harshavardhana (approved):

.travis.yml
```diff
@@ -27,4 +27,4 @@ after_success:
 - bash <(curl -s https://codecov.io/bash)
 
 go:
-- 1.8.3
+- 1.8.x
```
balamurugana, line 30: One question is why we would need to use latest version of 1.8 than 1.8.3 as default.

harshavardhana: This will pick the latest. So it could be 1.8.4 so it's an okay patch. Not needed strictly.

vadmeste (approved):
LGTM

PR #4953, opened 2017-09-21
azure: handle list objects properly

balamurugana:
When removing `minio.sys.tmp` prefixed entries and objects/prefixes is
empty, populate till we get all valid entries.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/gateway-azure.go
+++ b/cmd/gateway-azure.go
@@ -385,84 +385,73 @@ func (a *azureObjects) DeleteBucket(bucket string) error {
 // ListObjects - lists all blobs on azure with in a container filtered by prefix
 // and marker, uses Azure equivalent ListBlobs.
 func (a *azureObjects) ListObjects(bucket, prefix, marker, delimiter string, maxKeys int) (result ListObjectsInfo, err error) {
+	var objects []ObjectInfo
+	var prefixes []string
 	container := a.client.GetContainerReference(bucket)
-	resp, err := container.ListBlobs(storage.ListBlobsParameters{
-		Prefix:     prefix,
-		Marker:     marker,
-		Delimiter:  delimiter,
-		MaxResults: uint(maxKeys),
-	})
-	if err != nil {
-		return result, azureToObjectError(traceError(err), bucket, prefix)
-	}
-	result.IsTruncated = resp.NextMarker != ""
-	result.NextMarker = resp.NextMarker
-	for _, object := range resp.Blobs {
-		if strings.HasPrefix(object.Name, globalMinioSysTmp) {
-			continue
-		}
-		result.Objects = append(result.Objects, ObjectInfo{
-			Bucket:          bucket,
-			Name:            object.Name,
-			ModTime:         time.Time(object.Properties.LastModified),
-			Size:            object.Properties.ContentLength,
-			ETag:            azureToS3ETag(object.Properties.Etag),
-			ContentType:     object.Properties.ContentType,
-			ContentEncoding: object.Properties.ContentEncoding,
+	for len(objects) == 0 && len(prefixes) == 0 {
+		resp, err := container.ListBlobs(storage.ListBlobsParameters{
+			Prefix:     prefix,
+			Marker:     marker,
+			Delimiter:  delimiter,
+			MaxResults: uint(maxKeys),
 		})
-	}
+		if err != nil {
+			return result, azureToObjectError(traceError(err), bucket, prefix)
+		}
 
-	// Remove minio.sys.tmp prefix.
-	for i, prefix := range resp.BlobPrefixes {
-		if prefix == globalMinioSysTmp {
-			resp.BlobPrefixes = append(resp.BlobPrefixes[:i], resp.BlobPrefixes[i+1:]...)
+		for _, object := range resp.Blobs {
+			if strings.HasPrefix(object.Name, globalMinioSysTmp) {
+				continue
+			}
+			objects = append(objects, ObjectInfo{
+				Bucket:          bucket,
+				Name:            object.Name,
+				ModTime:         time.Time(object.Properties.LastModified),
+				Size:            object.Properties.ContentLength,
+				ETag:            azureToS3ETag(object.Properties.Etag),
+				ContentType:     object.Properties.ContentType,
+				ContentEncoding: object.Properties.ContentEncoding,
+			})
+		}
+
+		// Remove minio.sys.tmp prefix.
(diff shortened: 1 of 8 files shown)


cmd/gateway-azure.go
```diff
@@ -307,96 +307,102 @@ func (a *azureObjects) DeleteBucket(bucket string) error {
...
+func (a *azureObjects) ListObjects(bucket, prefix, marker, delimiter string, maxKeys int) (result ListObjectsInfo, err error) {
+	var r interface{}
+	r, err = a.listObjects(bucket, prefix, marker, delimiter, maxKeys, 1)
+	result = r.(ListObjectsInfo)
+
+	return result, err
+}
+
+// ListObjectsV2 - list all blobs in Azure bucket filtered by prefix
+func (a *azureObjects) ListObjectsV2(bucket, prefix, continuationToken string, fetchOwner bool, delimiter string, maxKeys int) (result ListObjectsV2Info, err error) {
```
krishnasrinivas: Simpler implementation (i.e you can avoid `a.listObjects`, passing of version number and returning catch-all interface) by doing this
```
func (a *azureObjects) ListObjectsV2() {
    resultV1 := a.ListObjects() // Call V1
    resultV2 := listObjectsV1ToV2(resultV1)
    return resultV1
}
```

balamurugana: Unable to understand why it is simpler than avoiding interface.

cmd/gateway-azure.go
```diff
@@ -310,90 +310,82 @@ func (a *azureObjects) DeleteBucket(bucket string) error {
...
-		return result, azureToObjectError(traceError(err), bucket, prefix)
-	}
-	result.IsTruncated = resp.NextMarker != ""
-	result.NextMarker = resp.NextMarker
-	for _, object := range resp.Blobs {
-		if strings.HasPrefix(object.Name, globalMinioSysTmp) {
-			continue
+	var objects []ObjectInfo
+	var prefixes []string
+	for len(objects) == 0 && len(prefixes) == 0 {
```
krisis: Issue: We may give lesser than `maxKeys` objects matching `prefix` starting after `marker` as we break the `for` loop once we have non-zero `objects` or `prefixes`.

Possible solution:
```go
for len(objects) < maxKeys {
...
}
```

cmd/gateway-router.go
```diff
@@ -36,8 +36,8 @@ type GatewayLayer interface {
 	GetBucketPolicies(string) (policy.BucketAccessPolicy, error)
 	DeleteBucketPolicies(string) error
 	AnonListObjects(bucket, prefix, marker, delimiter string, maxKeys int) (result ListObjectsInfo, err error)
-	AnonListObjectsV2(bucket, prefix, continuationToken string, fetchOwner bool, delimiter string, maxKeys int) (result ListObjectsV2Info, err error)
-	ListObjectsV2(bucket, prefix, continuationToken string, fetchOwner bool, delimiter string, maxKeys int) (result ListObjectsV2Info, err error)
+	AnonListObjectsV2(bucket, prefix, continuationToken, delimiter string, maxKeys int, fetchOwner bool, startAfter string) (result ListObjectsV2Info, err error)
```
krisis, line 39: we could keep `startAfter` with other `string` type arguments.

Suggestion:
```go
AnonListObjectsV2(bucket, prefix, continuationToken, delimiter, startAfter string, maxKeys int, fetchOwner bool) (result ListObjectsV2Info, err error)
```

krisis (changes requested):

cmd/gateway-azure.go
balamurugana: This is indeed an expected behavior i.e. we return less items than requested max keys with `IsTruncated = true`.

Returning 1000 keys with `IsTruncated - True` would lead to complicated logic.

cmd/gateway-router.go
balamurugana: It can, but I wanted to keep less used params at end.  Let me know whether we still need to move it to string argument group.

cmd/gateway-azure.go
```diff
@@ -310,90 +310,82 @@ func (a *azureObjects) DeleteBucket(bucket string) error {
...
+			if prefix != globalMinioSysTmp {
+				prefixes = append(prefixes, prefix)
+			}
+		}
+
+		if resp.NextMarker == "" {
 			break
 		}
+
+		marker = resp.NextMarker
```
krisis: `marker` needs to be set to `""` before breaking out of the `for` loop.

cmd/gateway-azure.go
krisis: @balamurugana this makes sense.

cmd/gateway-router.go
krisis: @balamurugana OK, you can leave arguments as is.

cmd/gateway-azure.go
balamurugana: Done

cmd/gateway-azure.go
```diff
@@ -310,90 +310,81 @@ func (a *azureObjects) DeleteBucket(bucket string) error {
...
-		result.Objects = append(result.Objects, ObjectInfo{
-			Bucket:          bucket,
-			Name:            object.Name,
-			ModTime:         t,
-			Size:            object.Properties.ContentLength,
-			ETag:            azureToS3ETag(object.Properties.Etag),
-			ContentType:     object.Properties.ContentType,
-			ContentEncoding: object.Properties.ContentEncoding,
-		})
+		result.NextContinuationToken = continuationToken
```
krishnasrinivas: should be:
result.NextContinuationToken = resultV1.NextMarker

balamurugana: Done. Also removed `if` block

krishnasrinivas (approved):

cmd/gateway-azure.go
harshavardhana: This is not true with our server mode such as FS and Erasure Code, is this agreed across all gateway implementations?

balamurugana: Unable to understand.  Could you tell me the problem exactly?

krisis (approved):

cmd/gateway-azure.go
harshavardhana: @balamurugana i meant in FS and Erasure code mode we return exact 1000 if max-keys=1000 . Are you saying that Azure will return 999 instead?

PR #4954, opened 2017-09-22
fs: support data IO sync through env MINIO_FSYNC

fangyuxiangGL:
Now, minio use buffer io when write temp data file, which might bring consistent issue. we add support for syncing IO through env in order to enhance consistence.

## Description
call Fsync when MINIO_FSYNC == true once all data in single http request is written into temp data file.
when MINIO_FSYNC is not specified to be true,  Fsync will not be called

## Motivation and Context
enhance consistence for minio over fs

## How Has This Been Tested?
once can do it as bellow:
MINIO_FSYNC=true ./minio server /mnt/minio/ -C ./

## Types of changes

## Checklist:

### DIFF
--- a/cmd/fs-v1-background-append.go
+++ b/cmd/fs-v1-background-append.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"io"
 	"os"
+	"syscall"
 	"reflect"
 	"sync"
 	"time"
@@ -236,5 +237,8 @@ func (fs fsObjects) appendPart(bucket, object, uploadID string, part objectPartI
 	}
 
 	_, err = io.CopyBuffer(wfile, file, buf)
+	if fsUseSync() {
+		syscall.Fsync(int(wfile.Fd()))
+	}
 	return err
 }
--- a/cmd/fs-v1-helpers.go
+++ b/cmd/fs-v1-helpers.go
@@ -17,9 +17,11 @@
 package cmd
 
 import (
+	"strings"
 	"fmt"
 	"io"
 	"os"
+	"syscall"
 	pathutil "path"
 	"runtime"
 )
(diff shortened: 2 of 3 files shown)


fangyuxiangGL:
@harshavardhana 

please take a look, thanks.

harshavardhana:
Thanks @fangyuxiangGL we will let you know on what we decide on this.  Current recommended approach is you mount with `-osync` option your backend disk.

cmd/fs-v1-background-append.go
```diff
@@ -236,5 +237,8 @@ func (fs fsObjects) appendPart(bucket, object, uploadID string, part objectPartI
 	}
 
 	_, err = io.CopyBuffer(wfile, file, buf)
+	if fsUseSync() {
+		syscall.Fsync(int(wfile.Fd()))
```
harshavardhana, line 241: You shouldn't use `syscall.Fsync` simply called `wfile.Sync()` instead.

cmd/fs-v1-helpers.go
```diff
@@ -280,6 +286,10 @@ func fsCreateFile(filePath string, reader io.Reader, buf []byte, fallocSize int6
 			return 0, traceError(err)
 		}
 	}
+
+	if fsUseSync() {
+		syscall.Fsync(int(writer.Fd()))
```
harshavardhana, line 291: You shouldn't use `syscall.Fsync` simply called `wfile.Sync()` instead.

cmd/fs-v1-multipart.go
```diff
@@ -792,6 +793,10 @@ func (fs fsObjects) CompleteMultipartUpload(bucket string, object string, upload
 				return oi, toObjectErr(traceError(err), bucket, object)
 			}
 
+		        if fsUseSync() {
+				syscall.Fsync(int(wfile.Fd()))
```
harshavardhana, line 797: You shouldn't use `syscall.Fsync` simply called `wfile.Sync()` instead.

cmd/fs-v1-multipart.go
```diff
@@ -792,6 +793,10 @@ func (fs fsObjects) CompleteMultipartUpload(bucket string, object string, upload
...
+		        }
+
 			wfile.Close()
 			reader.Close()
```
harshavardhana, line 801: There is also `posix.go` which requires this change @fangyuxiangGL are you comfortable doing that?

fangyuxiangGL:
@harshavardhana got u , thanks!

PR #4964, opened 2017-09-25
optimize memory allocs during reconstruct

aead:
## Description
The reedsolomon library now avoids allocations during reconstruction.
This change exploits that to reduce memory allocs and GC preasure during
healing and reading.

## Motivation and Context
This change will reduce the GC load of a XL setup when disks are down

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:


aead:
Here quick benchstats of a 2x2 setup 2 disks down and 100MB object
```
name        old time/op    new time/op    delta
ReadFile-4     180ms ± 0%     164ms ± 0%   ~     (p=1.000 n=1+1)

name        old alloc/op   new alloc/op   delta
ReadFile-4     109MB ± 0%      22MB ± 0%   ~     (p=1.000 n=1+1)

name        old allocs/op  new allocs/op  delta
ReadFile-4       433 ± 0%       415 ± 0%   ~     (p=1.000 n=1+1)
```

fwessels (approved):
LGTM and tested

donatello:
@aead You don't seem to be following the line length rule (at least in changed lines) that you wanted to have. 100 columns? We may bring up this point in the next meeting again, as some were not present last time. In addition, need to decide if review is the right place to point out this issue.

fwessels:
@donatello We can discuss this again, but let's also please separate main causes and side effects. Code looks perfectly fine (and improved).

donatello:
> @donatello We can discuss this again, but let's also please separate main causes and side effects. Code looks perfectly fine (and improved).

Agreed. I just wanted to check about how to proceed with the line length issue.

donatello (approved):
LGTM.

aead:
@donatello Your right about the length / column issue.
The problem is what do we do with e.g. [this](https://github.com/minio/minio/pull/4964/files#diff-cc000e8f9417c174d87c19bd6bf030d1R26). If I understood the discussion here correctly I should open an issue for refactoring it. Just blindly wrapping the line after 100CL was considered as not a sufficient approach (which I agree). So I open an issue for this.

aead:
#4988

PR #4965, opened 2017-09-26
Fix signature v2 handling for resource names

harshavardhana:
## Description
Previously we were wrongly adding `?` as part
of the resource name, add a test case to check
if this is handled properly.

Thanks to @kannappanr for reproducing this.

Without this change presigned URL generated with following
command would fail with signature mismatch.
```
aws s3 presign s3://testbucket/functional-tests.sh
```

## Motivation and Context
Bug reproduced by @kannappanr 

## How Has This Been Tested?
Manually and also using `go test`

## Types of changes

## Checklist:

### DIFF
--- a/cmd/signature-v2.go
+++ b/cmd/signature-v2.go
@@ -287,7 +287,7 @@ func canonicalizedAmzHeadersV2(headers http.Header) string {
 }
 
 // Return canonical resource string.
-func canonicalizedResourceV2(encodedQuery string) string {
+func canonicalizedResourceV2(encodedResource, encodedQuery string) string {
 	queries := strings.Split(encodedQuery, "&")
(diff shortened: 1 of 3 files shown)


cmd/signature-v2.go
```diff
@@ -287,7 +287,7 @@ func canonicalizedAmzHeadersV2(headers http.Header) string {
 }
 
 // Return canonical resource string.
-func canonicalizedResourceV2(encodedQuery string) string {
+func canonicalizedResourceV2(encodedResouce, encodedQuery string) string {
```
vadmeste: This should be `encodedResource` not `encodedResouce`

krishnasrinivas (approved):

donatello (approved):

PR #4969, opened 2017-09-26
api: MakeBucket() should honor regions properly.

### LINKED ISSUE #4967: MakeBucket call should fail if region is set in config.json and the region sent by client does not match
Server should validate the region parameter in MakeBucket call, if it is set in the server side

## Expected Behavior
If wrong region is sent by the client, MakeBucket call should fail.

## Current Behavior
Currently MakeBucket call succeeds, even if the region is set to say us-west-2 in config.json and the client sets the region to us-east-1

## Possible Solution
Server should fail to create the bucket if the region sent by the client does not match the region information is set in config.json. This check can be ignored if there is no region information set in config.json.

1. Set region value to "region": "us-west-2" in .minio/config.json
2. Start your Minio server
3. Use the follwoing code to reproduce the issue
```
package main

import (
	"fmt"
	"log"
	"math/rand"
	"time"
	minio "github.com/minio/minio-go"
)

func main() {
	// initialize logging params
	startTime := time.Now()
	function := "GetObject(bucketName, objectName)"
	args := map[string]interface{}{
		"bucketName": "",
		"objectName": "",
	}

	// Seed random based on current time.
	rand.Seed(time.Now().Unix())

	// Instantiate new minio client object.
	c, err := minio.New("localhost:9000", "minio", "minio123", false)
	if err != nil {
		log.Fatalln(function, args, startTime, "", "Minio client object creation failed", err)
	}

	// Generate a new random bucket name.
	bucketName := fmt.Sprintf("minio-go-test.%d", rand.Int())
	args["bucketName"] = bucketName

	// Make a new bucket.
	err = c.MakeBucket(bucketName, "us-east-1")
	if err != nil {
		fmt.Println("MakeBucket failed as expected")
	} else {
		log.Fatalln("MakeBucket success")
	}
	
}
```
## Your Environment
Linux ubuntu-ltp 4.10.0-35-generic #39-Ubuntu SMP Wed Sep 13 07:46:59 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux

harshavardhana:
## Description
api: MakeBucket() should honor regions properly.

## Motivation and Context
Fixes #4967

## Types of changes

## Checklist:

### DIFF
--- a/cmd/bucket-handlers.go
+++ b/cmd/bucket-handlers.go
@@ -380,11 +380,7 @@ func (api objectAPIHandlers) PutBucketHandler(w http.ResponseWriter, r *http.Req
 	}
 
 	// PutBucket does not have any bucket action.
-	s3Error := checkRequestAuthType(r, "", "", globalMinioDefaultRegion)
-	if s3Error == ErrInvalidRegion {
-		// Clients like boto3 send putBucket() call signed with region that is configured.
-		s3Error = checkRequestAuthType(r, "", "", serverConfig.GetRegion())
-	}
+	s3Error := checkRequestAuthType(r, "", "", serverConfig.GetRegion())
 	if s3Error != ErrNone {
 		writeErrorResponse(w, s3Error, r.URL)
 		return
--- a/cmd/gateway-handlers.go
+++ b/cmd/gateway-handlers.go
@@ -603,11 +603,7 @@ func (api gatewayAPIHandlers) PutBucketHandler(w http.ResponseWriter, r *http.Re
 	}
 
 	// PutBucket does not have any bucket action.
-	s3Error := checkRequestAuthType(r, "", "", globalMinioDefaultRegion)
-	if s3Error == ErrInvalidRegion {
-		// Clients like boto3 send putBucket() call signed with region that is configured.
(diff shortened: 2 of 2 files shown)


cmd/bucket-handlers.go
```diff
@@ -380,7 +380,7 @@ func (api objectAPIHandlers) PutBucketHandler(w http.ResponseWriter, r *http.Req
 	}
 
 	// PutBucket does not have any bucket action.
-	s3Error := checkRequestAuthType(r, "", "", globalMinioDefaultRegion)
```
vadmeste: As per my understanding, if we do this, it means the following `if` will not be needed

cmd/bucket-handlers.go
```diff
@@ -380,7 +380,7 @@ func (api objectAPIHandlers) PutBucketHandler(w http.ResponseWriter, r *http.Req
...
+	s3Error := checkRequestAuthType(r, "", "", globalServerRegion)
```
vadmeste: As per my understanding, if we do this, it means the following `if` will not be needed

harshavardhana: Yes that is correct, it was never called anymore @vadmeste

donatello (approved):

krishnasrinivas (approved):

PR #4972, opened 2017-09-26
Fix S3-Gateway PutObject

nitisht:
## Description
PutObject and PutObjectPart should parse md5 and shaSum only if present in metadata/data. 

## Motivation and Context
See https://github.com/minio/minio-java/issues/615

## How Has This Been Tested?
Running `minio-java` functional tests. 

## Types of changes

## Checklist:

### DIFF
--- a/cmd/gateway-s3.go
+++ b/cmd/gateway-s3.go
@@ -302,6 +302,23 @@ func (l *s3Objects) GetObject(bucket string, key string, startOffset int64, leng
 	return nil
 }
 
+// Decodes hex encoded md5, sha256 into their raw byte representations.
+func getMD5AndSha256SumBytes(md5Hex, sha256Hex string) (md5Bytes, sha256Bytes []byte, err error) {
+	if md5Hex != "" {
+		md5Bytes, err = hex.DecodeString(md5Hex)
+		if err != nil {
+			return nil, nil, err
+		}
+	}
+	if sha256Hex != "" {
+		sha256Bytes, err = hex.DecodeString(sha256Hex)
+		if err != nil {
+			return nil, nil, err
+		}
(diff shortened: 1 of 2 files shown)


harshavardhana (approved):

aead (changes requested):
This change is not needed as far as I can see.
We can discuss whether we should use a separate function for parsing MD5 and SHA256.

hex.DecodeString("") procudes []byte{}

nitisht:
> hex.DecodeString("") procudes []byte{}

yes, @aead . The function `getMD5AndSha256SumBytes` returns `nil, nil, nil` if both `md5Hex, sha256Hex` are "" (https://play.golang.org/p/LXnBEteiPa).

Not adding a method will cause same checks to be added in PutObject and PutObjectPart so a function is needed IMHO.

aead:
I agree on the function argument - but we don't need a this check: `md5Hex == ""`.
Instead of differencing between nil and []byte{} we should fix those places to len(slice) == 0.

nitisht:
Discussed with @aead and his suggestion is that `minio-go` should not differentiate between empty byte arrays ("") and `nil` byte arrays. As HTTP headers do not differentiate between value is not set and value is `""`. 

A better approach to fix this issue would be to fix the `contentMD5Bytes` []byte handling in `minio-go`. Will send a PR to `minio-go`.

nitisht:
> Will send a PR to minio-go

PR sent to `minio-go` : https://github.com/minio/minio-go/pull/828 If minio-go PR is accepted, we can close current PR.

deekoder:
Closing this since https://github.com/minio/minio-go/pull/828 is merged.

PR #5002, opened 2017-10-02
Implement backblaze-b2 gateway support

### LINKED ISSUE #4072: server: [gateway] Support for Backblaze B2
Hello

I think a lot of user could find a S3 gateway for B2 very useful, as there is a lot of backups tools for S3 and B2 is very cheap (0,005$/GB).

Rclone Go package could be used for integrating with the API: https://godoc.org/github.com/ncw/rclone/b2 https://godoc.org/github.com/ncw/rclone/b2/api

harshavardhana:
## Description
Implement backblaze-b2 gateway support

## Motivation and Context
Fixes #4072

## How Has This Been Tested?
Manually testing and mint.

## Types of changes

## Checklist:


harshavardhana:
Consistent unrelated failure on appveyor  ignore windows error for now.

```
go test -v -timeout 17m -coverprofile=c.txt -covermode=atomic github.com/minio/minio/cmd
go build github.com/minio/minio/cmd: c:\go18\pkg\tool\windows_amd64\compile.exe: fork/exec c:\go18\pkg\tool\windows_amd64\compile.exe: The filename or extension is too long.
```

Talking to appveyor support at the moment.

harshavardhana:
Looks like with this PR we are expanding beyond 32K limit on windows shell PATH_MAX - https://github.com/golang/go/issues/18468 

Work-around is i will try to merge the files reducing the overall length and lets see if it fixes.

harshavardhana:
> Looks like with this PR we are expanding beyond 32K limit on windows shell PATH_MAX - golang/go#18468
> 
> Work-around is i will try to merge the files reducing the overall length and lets see if it fixes.

Was able to fix it by merging some files.

cmd/gateway-s3.go
```diff
@@ -134,8 +135,9 @@ func newS3Gateway(host string) (GatewayLayer, error) {
 	anonClient.SetCustomTransport(newCustomHTTPTransport())
 
 	return &s3Objects{
-		Client:     client,
-		anonClient: anonClient,
+		unsupportedAPIs: unsupportedAPIs{},
```
krishnasrinivas: `unsupportedAPIs: unsupportedAPIs{},` can be left out

cmd/gateway-unsupported.go
```diff
@@ -16,27 +16,30 @@
 
 package cmd
 
+type unsupportedAPIs struct{}
```
krishnasrinivas: can be renamed to more specific `gatewayUnsupported`

cmd/gateway-b2.go
```diff
@@ -0,0 +1,652 @@
...
+	lpi = ListPartsInfo{
+		Bucket:           bucket,
+		Object:           object,
+		UploadID:         uploadID,
+		MaxParts:         maxParts,
+		PartNumberMarker: partNumberMarker,
+	}
+	// startPartNumber must be in the range 1 - 10000 for B2.
+	if partNumberMarker == 0 {
+		partNumberMarker = 1
```
krishnasrinivas: this will cause the list response to always skip the first part

harshavardhana: It won't @krishnasrinivas . Not in case of Backblaze.

harshavardhana: startPartNumber includes 1 and later, it is not the same style as S3.

krishnasrinivas: in that case we need to do partNumberMarker++ always because:

```
krishna@escape:~/dev/minio-tile-2$ miniodebug multipart listparts --bucket krishna1980 --object testobject --uploadid 4_z1efea6cee041ea395aeb021f_f205a55e6b4215c82_d20171003_m202115_c001_v0001093_t0023 --partmarker 2
{
  "Bucket": "krishna1980",
  "Key": "testobject",
  "UploadID": "4_z1efea6cee041ea395aeb021f_f205a55e6b4215c82_d20171003_m202115_c001_v0001093_t0023",
  "Initiator": {
    "ID": "02d6176db174dc93cb1b899f7c6078f08654445fe8cf1b6ce98d8855f66bdbf4",
    "DisplayName": ""
  },
  "Owner": {
    "DisplayName": "",
    "ID": "02d6176db174dc93cb1b899f7c6078f08654445fe8cf1b6ce98d8855f66bdbf4"
  },
  "StorageClass": "STANDARD",
  "PartNumberMarker": 2,
  "NextPartNumberMarker": 0,
  "MaxParts": 1000,
  "IsTruncated": false,
  "ObjectParts": [
    {
      "PartNumber": 2,
      "LastModified": "0001-01-01T00:00:00Z",
      "ETag": "\"5da54579c335464b6ba52090bf57d063384bb1e9\"",
      "Size": 2461
    },
    {
      "PartNumber": 3,
      "LastModified": "0001-01-01T00:00:00Z",
      "ETag": "\"5da54579c335464b6ba52090bf57d063384bb1e9\"",
      "Size": 2461
    }
  ],
  "EncodingType": ""
}
```

harshavardhana: Oh yes you are right. @krishnasrinivas

cmd/gateway-b2-anonymous.go
```diff
@@ -0,0 +1,144 @@
...
+	if offset == 0 && size == 0 {
+		return ""
+	}
+	if size == 0 {
+		return fmt.Sprintf("%s%d-", byteRangePrefix, offset)
+	}
+	return fmt.Sprintf("%s%d-%d", byteRangePrefix, offset, offset+size-1)
+}
+
+// AnonGetObject - Get object anonymously
```
krisis: This comment doesn't say anything more than the name of the function. The comment could be `AnonGetObject - Get object using a plain GET request.` or something to that effect.

cmd/gateway-b2-anonymous.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+	defer resp.Body.Close()
+	_, err = io.Copy(writer, resp.Body)
+	return err
+}
+
+// Converts http Header into ObjectInfo. This function looks for all the
+// standard Backblaze B2 headers to convert into ObjectInfo.
+//
+// Content-Length is converted to Size.
```
krisis, line 68: nit: we could order the headers in the comment such that `X-Bz-...` appear together.

cmd/gateway-b2-anonymous.go
```diff
@@ -0,0 +1,144 @@
...
+	clen, err := strconv.ParseInt(header.Get("Content-Length"), 10, 64)
+	if err != nil {
+		return objInfo, err
+	}
+	timeStamp, err := strconv.ParseInt(header.Get("X-Bz-Upload-Timestamp"), 10, 64)
+	if err != nil {
+		return objInfo, err
+	}
+
+	info := make(map[string]string)
```
krisis: we could call this `userDefinedHdrs` instead.

cmd/gateway-b2-anonymous.go
```diff
@@ -0,0 +1,144 @@
...
+		if err != nil {
+			return objInfo, err
+		}
+		info[name] = val
+	}
+	objInfo = ObjectInfo{
+		Bucket:      bucket,
+		Name:        object,
+		ContentType: header.Get("Content-Type"),
+		ModTime:     time.Unix(timeStamp/1000, timeStamp%1000*1e6),
```
krisis: `ModTime` can be idiomatically computed like `time.Unix(0, 0).Add(timeStamp * time.Millisecond)`.

cmd/gateway-b2-anonymous.go
```diff
@@ -0,0 +1,144 @@
...
+	req, err := http.NewRequest("HEAD", uri, nil)
+	if err != nil {
+		return objInfo, err
+	}
+	resp, err := l.anonClient.Do(req)
+	if err != nil {
+		return objInfo, err
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != 200 && resp.StatusCode != 206 {
```
krisis: do we expect partial content for a HEAD request?

cmd/gateway-b2.go
```diff
@@ -0,0 +1,650 @@
...
+		bucket = params[0]
+	}
+	if len(params) == 2 {
+		object = params[1]
+	}
+	if len(params) == 3 {
+		uploadID = params[2]
+	}
+
+	// Following code is a non-exhaustive list.
```
krisis: nit: `Following code is a non-exhaustive list` is out of place since there is no list of codes below.

cmd/gateway-b2.go
```diff
@@ -0,0 +1,650 @@
...
+
+// StorageInfo is not relevant to B2 backend.
+func (l *b2Objects) StorageInfo() (si StorageInfo) {
+	return si
+}
+
+// MakeBucket creates a new container on B2 backend.
+func (l *b2Objects) MakeBucketWithLocation(bucket, location string) error {
+	info := make(map[string]string)
+	info["x-minio-bucket-createTime"] = UTCNow().Format(time.RFC1123)
```
krisis: "x-minio-bucket-createTime" should be a const.

cmd/gateway-b2.go
```diff
@@ -0,0 +1,650 @@
...
+	// All buckets are set to private by default.
+	_, err := l.b2Client.CreateBucket(l.ctx, bucket, "allPrivate", info, nil)
+	return b2ToObjectError(traceError(err), bucket)
+}
+
+// Bucket - bucket
```
krisis: this comment can be expanded.

cmd/gateway-b2.go
```diff
@@ -0,0 +1,650 @@
...
+	for _, bkt := range bktList {
+		if bkt.Name == bucket {
+			return bkt, nil
+		}
+	}
+	return nil, traceError(BucketNotFound{Bucket: bucket})
+}
+
+// Looks for x-minio-bucket-createTime as part of bucket metadata, if key
+// is absent then returns current UTC time instead. Returns error if
```
krisis: When creation time is unavailable we should use a sentinel value like `time.Unix(0,0) to indicate that we couldn't determine the actual time of creation. Returning current time may be misleading.

krisis (changes requested):
These are some initial comments.

cmd/gateway-b2-anonymous.go
harshavardhana: Yes but why is your approach more idiomatic? @krisis .

cmd/gateway-b2-anonymous.go
harshavardhana: Yes if you set the Range header.

cmd/gateway-b2.go
harshavardhana: time.Unix(0,0) is not a good idea because of the tests that expect CreatTime to be always within some acceptable boundary of time.  The reason to use currentTime here is to provide lastAccessedTime instead.

cmd/gateway-b2.go
harshavardhana: Sure

cmd/gateway-b2.go
harshavardhana: Agreed

cmd/gateway-b2.go
krisis: > time.Unix(0,0) is not a good idea because of the tests that expect CreatTime to be always within some acceptable boundary of time

Which tests are you referring to? If it is unit tests we can (and should) change it as fit. This suggestion is to avoid any confusion to the user, who knows that the bucket was created much before the time we might show (i.e, last accessed time like you mention above).

cmd/gateway-b2-anonymous.go
krisis: It is idiomatic since it uses standard Go types like time.Duration to work with milliseconds. To work with milliseconds available as string in header, we could do the following,
```go
timeStamp, err := time.ParseDuration(header.Get("X-Bz-Upload-Timestamp"))
if err != nil {
}
modTime := time.Unix(0, 0).Add(timeStamp)
```
This way we use the appropriate Go types to work with time in milliseconds instead of dropping down to low-level types like int64, int.

cmd/gateway-b2-anonymous.go
harshavardhana: Sure

cmd/gateway-b2-anonymous.go
harshavardhana: They are written in the order they are handled inside the code..

cmd/gateway-b2-anonymous.go
krisis: we aren't setting range headers here. should we still check?

cmd/gateway-b2-anonymous.go
harshavardhana: Sure

cmd/gateway-b2.go
harshavardhana: I am talking about mint tests.. Setting time.Unix(0, 0) will be more confusing to user IMO. I would prefer not make this change and keep it the way it is. We can even discuss about this but i would like to avoid epoch date.

For the most part users using buckets from minio gateway will not have any problems, B2 users aren't expecting this value anyways. So i choose to set this value to something more meaningful than Epoch time.

harshavardhana: As a matter of fact perhaps i can remove the x-minio-bucket-createTime and avoid such intelligence in the first place.

krisis: > Setting time.Unix(0, 0) will be more confusing to user IMO. I would prefer not make this 

The idea behind `time.Unix(0,0)` is that it is a constant and doesn't change with time like `time.Now()`. We could choose a different constant `time.Time` value if that's the only issue.

> As a matter of fact perhaps i can remove the x-minio-bucket-createTime and avoid such intelligence in the first place.

what would `InitiatedTime` value for `BucketInfo` be?

cmd/gateway-b2-anonymous.go
harshavardhana: we don't need then

cmd/gateway-b2.go
harshavardhana: Nothing just time.Unix(0, 0) there is no such info available from B2 API so we don't return anything.

harshavardhana:
New change in this PR is regular PutObject() is doing multipart if the input stream size is bigger than MinPartSize . This is to ensure that we don't corrupt the namespace if there were errors during data transfer.

krishnasrinivas:
> New change in this PR is regular PutObject() is doing multipart if the input stream size is bigger than MinPartSize . This is to ensure that we don't corrupt the namespace if there were errors during data transfer.

Simpler thing would be not to send SHA1 in the end if there were any errors?

harshavardhana:
> Simpler thing would be not to send SHA1 in the end if there were any errors?

Yes that is possible but that would require a very interwoven logic with HashReader() .. Let me explore that possibility.

harshavardhana:
> Simpler thing would be not to send SHA1 in the end if there were any errors?

That is indeed better @krishnasrinivas good suggestion and its incorporated. Perhaps something similar can be done in other gateway backends as well to avoid namespace clutter when we bring in DARE encryption.

cmd/gateway-b2.go
```diff
@@ -0,0 +1,634 @@
...
+// content reader.
+func newHexDigitsEndReader(r *HashReader, size int64) *hexEndReader {
+	return &hexEndReader{
+		r:    r,
+		size: int(size),
+		hsh:  sha1.New(),
+	}
+}
+
+type hexEndReader struct {
```
krishnasrinivas: could be renamed to `b2Hasher` as `hexEndReader` would be a generic name

cmd/gateway-b2.go
```diff
@@ -0,0 +1,634 @@
...
+		return pi, err
+	}
+
+	fc, err := bkt.File(uploadID, object).CompileParts(0, nil).GetUploadPartURL(l.ctx)
+	if err != nil {
+		return pi, b2ToObjectError(traceError(err), bucket, object, uploadID)
+	}
+
+	hr := newHexDigitsEndReader(data, data.Size())
+	sha1, err := fc.UploadPart(l.ctx, hr, "hex_digits_at_end", hr.Len(), partID)
```
krishnasrinivas: "hex_digits_at_end" should be a `const`

cmd/gateway-b2-anonymous.go
```diff
@@ -0,0 +1,146 @@
...
+		}
+		userMetadata[name] = val
+	}
+	objInfo = ObjectInfo{
+		Bucket:      bucket,
+		Name:        object,
+		ContentType: header.Get("Content-Type"),
+		ModTime:     time.Unix(0, 0).Add(time.Duration(timeStamp) * time.Millisecond),
+		Size:        clen,
+		ETag:        header.Get("X-Bz-Content-Sha1"),
```
krishnasrinivas: Looks like fileID is the right candidate for ETag, because the functionality of `If-Match` is supported by `b2_download_file_by_id` and `b2_get_file_info` both of which take fileID, there is no SHA1 support for these calls

krishnasrinivas (approved):

cmd/gateway-b2-anonymous.go
```diff
@@ -0,0 +1,123 @@
...
+// Converts http Header into ObjectInfo. This function looks for all the
+// standard Backblaze B2 headers to convert into ObjectInfo.
+//
+// Content-Length is converted to Size.
+// X-Bz-Upload-Timestamp is converted to ModTime.
+// X-Bz-Info-<header>:<value> is converted to <header>:<value>
+// Content-Type is converted to ContentType.
+// X-Bz-Content-Sha1 is converted to ETag.
+func headerToObjectInfo(bucket, object string, header http.Header) (objInfo ObjectInfo, err error) {
+	clen, err := strconv.ParseInt(header.Get("Content-Length"), 10, 64)
```
krisis, line 74: Needs a comment mentioning what is going on here. E.g, `// Converting upload timestamp in milliseconds to a time.Time value for ObjectInfo.ModTime`.

cmd/gateway-b2-anonymous.go
```diff
@@ -0,0 +1,123 @@
...
+	if err != nil {
+		return objInfo, err
+	}
+	timeStamp, err := strconv.ParseInt(header.Get("X-Bz-Upload-Timestamp"), 10, 64)
+	if err != nil {
+		return objInfo, err
+	}
+
+	userMetadata := make(map[string]string)
```
krisis: A comment explaining why were skipping headers that don't start with `X-Bz-Info` is needed.

cmd/gateway-b2-anonymous.go
```diff
@@ -0,0 +1,123 @@
...
+	req, err := http.NewRequest("HEAD", uri, nil)
+	if err != nil {
+		return objInfo, err
+	}
+	resp, err := l.anonClient.Do(req)
+	if err != nil {
+		return objInfo, err
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != 200 {
```
krisis: why are we returning a new error value, i.e `errors.New(resp.Status)` while returning the error as is in `AnonGetObject`?

cmd/gateway-b2.go
```diff
@@ -0,0 +1,660 @@
...
+// MakeBucket creates a new container on B2 backend.
+func (l *b2Objects) MakeBucketWithLocation(bucket, location string) error {
+	// location is ignored for B2 backend.
+
+	// All buckets are set to private by default.
+	_, err := l.b2Client.CreateBucket(l.ctx, bucket, bucketTypePrivate, nil, nil)
+	return b2ToObjectError(traceError(err), bucket)
+}
+
+// Bucket - bucket
```
krisis: Could we name this function similar to other functions like `GetBucket`? Is this function required to be exported? Please add a comment too.

cmd/gateway-b2.go
```diff
@@ -0,0 +1,660 @@
...
+		switch file.Status {
+		case "folder":
+			loi.Prefixes = append(loi.Prefixes, file.Name)
+		case "upload":
+			loi.Objects = append(loi.Objects, ObjectInfo{
+				Bucket:      bucket,
+				Name:        file.Name,
+				ModTime:     file.Timestamp,
+				Size:        file.Size,
+				ETag:        file.Info.SHA1,
```
krisis: Shouldn't `file.FileID` be returned for `ETag`?

cmd/gateway-b2.go
```diff
@@ -0,0 +1,660 @@
...
```
krisis: Shouldn't `file.FileID` be returned for `ETag`?

cmd/gateway-b2.go
```diff
@@ -0,0 +1,660 @@
...
+
+// B2Reader - is a Reader wraps the HashReader which will emit out the sha1
+// hex digits at io.EOF. It also means that your overall content size is
+// now original size + 40 bytes. Additionally this reader also verifies
+// Hash encapsulated inside HashReader at io.EOF if the verification
+// failed we return an error and do not send the content to server.
+type B2Reader struct {
+	r    *HashReader
+	size int64
+	hsh  hash.Hash
```
krisis: we can call it `sha1Sum` instead of `hsh`.

cmd/gateway-b2.go
```diff
@@ -0,0 +1,660 @@
...
+func (l *b2Objects) DeleteObject(bucket string, object string) error {
+	bkt, err := l.Bucket(bucket)
+	if err != nil {
+		return err
+	}
+	reader, err := bkt.DownloadFileByName(l.ctx, object, 0, 1)
+	if err != nil {
+		return b2ToObjectError(traceError(err), bucket, object)
+	}
+	io.Copy(ioutil.Discard, reader)
```
krisis, line 483: can we not Close `reader` without copying?

cmd/gateway-b2.go
```diff
@@ -0,0 +1,660 @@
...
+		lpi.Parts = append(lpi.Parts, PartInfo{
+			PartNumber: part.Number,
+			ETag:       part.SHA1,
+			Size:       part.Size,
+		})
+	}
+	return lpi, nil
+}
+
+// AbortMultipartUpload aborts a ongoing multipart upload, uses B2's LargeFile upload API.
```
krisis: typo: s/a ongoing/an ongoing/

cmd/gateway-b2.go
```diff
@@ -0,0 +1,660 @@
...
+	if err != nil {
+		return oi, err
+	}
+	hashes := make(map[int]string)
+	for _, uploadedPart := range uploadedParts {
+		hashes[uploadedPart.PartNumber] = uploadedPart.ETag
+	}
+	_, err = bkt.File(uploadID, object).CompileParts(0, hashes).FinishLargeFile(l.ctx)
+	if err != nil {
+		return oi, err
```
krisis: shouldn't we call `b2ToObjectError(traceError(err), ...)`?

krisis (changes requested):

cmd/gateway-b2-anonymous.go
harshavardhana: what is the error that you would return as is?

cmd/gateway-b2-anonymous.go
harshavardhana: Done

cmd/gateway-b2-anonymous.go
harshavardhana: done

cmd/gateway-b2.go
harshavardhana: done

cmd/gateway-b2.go
harshavardhana: done

cmd/gateway-b2.go
harshavardhana: done

cmd/gateway-b2.go
harshavardhana: done

cmd/gateway-b2.go
harshavardhana: done

cmd/gateway-b2.go
harshavardhana: done

cmd/gateway-b2.go
harshavardhana: done

cmd/gateway-b2.go
```diff
@@ -0,0 +1,699 @@
...
+	"fmt"
+	"hash"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"strings"
+	"sync"
+	"time"
+
+	b2 "github.com/minio/blazer/base"
```
krisis, line 31: why are we not using "github.com/minio/blazer/b2" client, which has reauthorization built into it?

harshavardhana: We need low level API @krisis

krisis (approved):

PR #5006, opened 2017-10-03
Translate s3 gateway errors at object layer

krisis:
## Description

S3 gateway uses minio-go's `Core` type for accessing S3-like object storage services. This returns error values of type `ErrorResponse` which is different from what object-layer already handles.

## Motivation and Context

Refer https://github.com/minio/mint/issues/168

## Types of changes

## Checklist:

### DIFF
--- a/cmd/gateway-s3.go
+++ b/cmd/gateway-s3.go
@@ -85,6 +85,10 @@ func s3ToObjectError(err error, params ...string) error {
 		}
 	case "XAmzContentSHA256Mismatch":
 		err = SHA256Mismatch{}
+	case "NoSuchUpload":
+		err = InvalidUploadID{}
+	case "EntityTooSmall":
+		err = PartTooSmall{}
 	}
 
 	e.e = err
@@ -195,7 +199,7 @@ func (l *s3Objects) GetBucketInfo(bucket string) (bi BucketInfo, e error) {
 func (l *s3Objects) ListBuckets() ([]BucketInfo, error) {
 	buckets, err := l.Client.ListBuckets()
 	if err != nil {
-		return nil, err
+		return nil, s3ToObjectError(traceError(err))
 	}
 
 	b := make([]BucketInfo, len(buckets))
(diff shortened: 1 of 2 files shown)


donatello (changes requested):
Looks it needs to be done for `NewMultipartUpload` too?

harshavardhana (approved):
LGTM

cmd/gateway-s3.go
```diff
@@ -85,6 +85,10 @@ func s3ToObjectError(err error, params ...string) error {
 		}
 	case "XAmzContentSHA256Mismatch":
 		err = SHA256Mismatch{}
+	case "NoSuchUpload":
```
harshavardhana, line 88: Can we bring in a test case for s3ToObjectError() ? @krisis

harshavardhana (changes requested):

cmd/gateway-s3.go
krisis: @harshavardhana  done

harshavardhana (approved):

cmd/gateway-s3.go
```diff
@@ -450,7 +454,11 @@ func toMinioClientMetadata(metadata map[string]string) map[string]string {
 func (l *s3Objects) NewMultipartUpload(bucket string, object string, metadata map[string]string) (uploadID string, err error) {
 	// Create PutObject options
 	opts := minio.PutObjectOptions{UserMetadata: metadata}
-	return l.Client.NewMultipartUpload(bucket, object, opts)
+	uploadID, err = l.Client.NewMultipartUpload(bucket, object, opts)
+	if err != nil {
+		return uploadID, s3ToObjectError(traceError(err), bucket, object)
+	}
+	return uploadID, nil
```
donatello, line 461: The if-block can be avoided. A single return statement would be enough as `s3ToObjectError` would return nil of the error is nil.

donatello (approved):
Just a minor comment - approving either way.

PR #5015, opened 2017-10-04
Return errors in PutObject()/PutObjectPart() if input size is -1.

harshavardhana:
## Description
Amazon S3 API expects all incoming stream has a content-length
It was superflous for us to support object layer which supports
unknown sized stream as well, this PR removes such requirements
and explicitly error out if input stream is less than zero.

## Motivation and Context
Strictness in behavior and avoid 2 ways of interpreting the stream.

## How Has This Been Tested?
Manually and using `go test`

## Types of changes

## Checklist:


cmd/xl-v1-utils.go
```diff
@@ -382,35 +382,37 @@ func evalDisks(disks []StorageAPI, errs []error) []StorageAPI {
...
-		return totalSize % partSize, nil
-	default:
-		return 0, nil
+	if totalSize > 0 {
+		// Compute the total count of parts
+		partsCount := totalSize/partSize + 1
+		// Return the part's size
+		switch {
+		case int64(partIndex) < partsCount:
+			totalSize = partSize
```
vadmeste: Can you add a new variable `currPartSize` and avoid setting totalSize because its meaning is changed here

harshavardhana: Will do

krishnasrinivas (approved):

vadmeste (approved):
LGTM & tested

PR #5028, opened 2017-10-07
Avoid blocking send on channels inside mutex

### LINKED ISSUE #5023: minio client hangs on certain buckets after certain operations
When you do a combination of "mc find watch" on one terminal and "mc cp" on another, mc hangs on certain commands on the bucket where this operation was performed.

## Expected Behavior
At any given point, mc cp and ls command should not hang on any bucket

## Current Behavior
Once the server gets into this state, mc cp and mc ls commands hang on those buckets, on which these operations were performed.

From then on, whenever you do a mc cp onto any other bucket on the server, that also starts exhibiting the same behavior.

Environment: Connecting to my local Minio server

Bucket Name: ``` test ```
Bucket  has the following files
```
datafile-100-kB
datafile-11-MB
datafile-33-kB
datafile-6-MB
datafile-1.03-MB
datafile-1-b
datafile-5-MB
listfiles.go
datafile-10-kB
datafile-1-MB
datafile-65-MB
list.go
listfiles.go
```
____________________________________________
On terminal 1
enter the following command
```
mc find myminio/test --watch
```
You should only see
```
/test/datafile-1-MB
```
Now keep pressing Ctrl +C, until it displays the last file
```
/test/datafile-1-MB
^C
/test/datafile-1-b
^C
/test/datafile-1.03-MB
^C
/test/datafile-10-kB
^C
/test/datafile-100-kB
^C
/test/datafile-11-MB
^C
/test/datafile-33-kB
^C
/test/datafile-5-MB
^C
/test/datafile-6-MB
^C
/test/datafile-65-MB
^C
/test/list.go
^C
/test/listfiles.go
```

At this point on Terminal 2, enter the following command
```
mc cp list.go myminio/test
```
on Terminal 1, you should see this
```
panic: close of closed channel

goroutine 40 [running]:
github.com/minio/mc/cmd.(*watchObject).Close(0xc4202a49e0)
	/home/test/gopath/src/github.com/minio/mc/cmd/watch.go:84 +0x2f
github.com/minio/mc/cmd.(*s3Client).Watch.func2(0xc4202a49e0, 0xc4202b0f00, 0xc4202b0de0, 0xc42030f060, 0xc4202b0d80)
	/home/test/gopath/src/github.com/minio/mc/cmd/client-s3.go:495 +0xc9a
created by github.com/minio/mc/cmd.(*s3Client).Watch
	/home/test/gopath/src/github.com/minio/mc/cmd/client-s3.go:495 +0x90e
```

-------------------------------------
On terminal 1, enter the following command
```
mc find myminio/test --watch
```
you should see
```
/test/datafile-1-MB
```
now, on terminal 2, enter the following command multiple times

```
mc cp list.go myminio/test
```
on the first terminal you will see
```
/test/datafile-1-MB
myminiomyminio/test/list.go
myminiomyminio/test/list.go
myminiomyminio/test/list.go
myminiomyminio/test/list.go
myminiomyminio/test/list.go
myminiomyminio/test/list.go
myminiomyminio/test/list.go
myminiomyminio/test/list.go
myminiomyminio/test/list.go
myminiomyminio/test/list.go
^C
/test/datafile-1-b
panic: send on closed channel
	panic: close of closed channel

goroutine 5 [running]:
github.com/minio/mc/cmd.(*watchObject).Close(0xc42000c4c0)
	/home/test/gopath/src/github.com/minio/mc/cmd/watch.go:84 +0x2f
panic(0x8aa440, 0xc42032cac0)
	/home/test/go/src/runtime/panic.go:489 +0x2cf
github.com/minio/mc/cmd.(*s3Client).Watch.func2(0xc42000c4c0, 0xc42001a2a0, 0xc42001a180, 0xc4202f19a0, 0xc42001a120)
	/home/test/gopath/src/github.com/minio/mc/cmd/client-s3.go:461 +0x964
created by github.com/minio/mc/cmd.(*s3Client).Watch
	/home/test/gopath/src/github.com/minio/mc/cmd/client-s3.go:495 +0x90e

```
------------------------
on terminal 1, enter the following command 
```
mc find myminio/test --watch
```
you should see
```
/test/datafile-1-MB
```
now, on terminal 2, enter the following command multiple times

on the first terminal you will see
```
/test/datafile-1-MB
myminiomyminio/test/list.go
myminiomyminio/test/list.go
myminiomyminio/test/list.go
myminiomyminio/test/list.go
myminiomyminio/test/list.go
myminiomyminio/test/list.go
myminiomyminio/test/list.go
myminiomyminio/test/list.go
myminiomyminio/test/list.go
myminiomyminio/test/list.go
myminiomyminio/test/list.go
myminiomyminio/test/list.go
myminiomyminio/test/list.go
myminiomyminio/test/list.go
myminiomyminio/test/list.go
myminiomyminio/test/list.go
myminiomyminio/test/list.go
```
Basically, for every single mc cp on terminal 2, now you will start seeing multiple lines being displayed on terminal 1. This is an indication that something is wrong in the server.

now press Ctrl +C on terminal 1

you should see
```
/test/datafile-1-b
```
on terminal 2, enter the following command
```
mc cp list.go myminio/test
```
you will see the following panic on terminal 1
```
panic: send on closed channel
	panic: close of closed channel

goroutine 37 [running]:
github.com/minio/mc/cmd.(*watchObject).Close(0xc4202ed920)
	/home/test/gopath/src/github.com/minio/mc/cmd/watch.go:84 +0x2f
panic(0x8aa440, 0xc42035e500)
	/home/test/go/src/runtime/panic.go:489 +0x2cf
github.com/minio/mc/cmd.(*s3Client).Watch.func2(0xc4202ed920, 0xc4202c9560, 0xc4202c9440, 0xc4202ed560, 0xc4202c93e0)
	/home/test/gopath/src/github.com/minio/mc/cmd/client-s3.go:461 +0x964
created by github.com/minio/mc/cmd.(*s3Client).Watch
	/home/test/gopath/src/github.com/minio/mc/cmd/client-s3.go:495 +0x90e
```

-------------------

Now on terminal 1, enter the following command

```
mc find myminio/test --watch
```
```
/test/datafile-1-MB
```
press Ctrl + C again

On terminal 2, enter the mc cp command again

Now the cp command should hang

press Ctrl + C on terminal 2 

On terminal 3, enter 
```
mc ls myminio/test 
```
you should see the command hang

## Your Environment

* Version used (`minio version`): 2017-09-29T19:16:56Z
* Server type and version:
Linux ubuntu-ltp 4.10.0-35-generic #39-Ubuntu SMP Wed Sep 13 07:46:59 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
* Link to your project:

krisis:
## Description

The server side implementation of ListenBucketNotification uses channels to send notification events from the origin of the event like PutObject to the ListenBucketHandler (used by mc find --watch internally). The channels are stored in a map which is protected by a mutex. ListenBucketHandler does the following when a client closes connection.

- sendBucketNotification returns immediately after a failed write (over HTTP). This means we have a channel that has a sender with no receiver
- removes reference to the channel from the map
- closes the channel.

On the other hand, a concurrent PutObjectHandler does the following,

- send notification event on channels read from the map (note: this is inside the critical section)
- release the namespace mutex lock

In this scenario, PutObjectHandler is blocked, holding the mutex protecting the map of channels and the namespace mutex since the channels used to send notification is blocked (remember, it has no receiver).

This change removes the cyclic dependency between the event channel and map of ARN to channels by using a separate done channel to signal that the client has quit.

Fixes #5023 
## Motivation and Context

https://github.com/minio/minio/issues/5023
## How Has This Been Tested?

Not yet.
## Types of changes

## Checklist:

### DIFF
--- a/cmd/bucket-notification-handlers.go
+++ b/cmd/bucket-notification-handlers.go
@@ -22,7 +22,9 @@ import (
 	"encoding/xml"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
+	"syscall"
 	"time"
 
 	"github.com/gorilla/mux"
@@ -213,14 +215,6 @@ func writeNotification(w http.ResponseWriter, notification map[string][]Notifica
 		return err
 	}
 
-	// https://github.com/containous/traefik/issues/560
-	// https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events
-	//
-	// Proxies might buffer the connection to avoid this we
-	// need the proper MIME type before writing to client.
-	// This MIME header tells the proxies to avoid buffering
-	w.Header().Set("Content-Type", "text/event-stream")
-
 	// Add additional CRLF characters for client to
 	// differentiate the individual events properly.
 	_, err = w.Write(append(notificationBytes, crlf...))
@@ -232,25 +226,61 @@ func writeNotification(w http.ResponseWriter, notification map[string][]Notifica
 // CRLF character used for chunked transfer in accordance with HTTP standards.
 var crlf = []byte("\r\n")
 
-// sendBucketNotification - writes notification back to client on the response writer
-// for each notification input, otherwise writes whitespace characters periodically
-// to keep the connection active. Each notification messages are terminated by CRLF
-// character. Upon any error received on response writer the for loop exits.
-func sendBucketNotification(w http.ResponseWriter, arnListenerCh <-chan []NotificationEvent) {
-	var dummyEvents = map[string][]NotificationEvent{"Records": nil}
-	// Continuously write to client either timely empty structures
-	// every 5 seconds, or return back the notifications.
+// listenChan A `listenChan` provides a data channel to send event
+// notifications on and `doneCh` to signal that events are no longer
+// being received. It also sends empty events (whitespace) to keep the
+// underlying connection alive.
+type listenChan struct {
+	doneCh chan struct{}
+	dataCh chan []NotificationEvent
+}
+
+// newListenChan returns a listenChan with properly initialized
+// unbuffered channels.
+func newListenChan() *listenChan {
+	return &listenChan{
+		doneCh: make(chan struct{}),
+		dataCh: make(chan []NotificationEvent),
+	}
+}
+
+// sendNotificationEvent sends notification events on the data channel
+// unless doneCh is not closed
+func (l *listenChan) sendNotificationEvent(events []NotificationEvent) {
+	select {
+	// Returns immediately if receiver has quit.
+	case <-l.doneCh:
+	// Blocks until receiver is available.
+	case l.dataCh <- events:
+	}
+}
+
+// waitForListener writes event notification OR whitespaces on
+// ResponseWriter until client closes connection
+func (l *listenChan) waitForListener(w http.ResponseWriter) {
+
+	// Logs errors other than EPIPE and ECONNRESET.
+	// EPIPE and ECONNRESET indicate that the client stopped
+	// listening to notification events.
+	logClientError := func(err error, msg string) {
+		if oe, ok := err.(*net.OpError); ok && (oe.Err == syscall.EPIPE || oe.Err ==
+			syscall.ECONNRESET) {
+			errorIf(err, msg)
+		}
+	}
+
+	emptyEvent := map[string][]NotificationEvent{"Records": nil}
+	defer close(l.doneCh)
 	for {
 		select {
-		case events := <-arnListenerCh:
+		case events := <-l.dataCh:
 			if err := writeNotification(w, map[string][]NotificationEvent{"Records": events}); err != nil {
-				errorIf(err, "Unable to write notification to client.")
+				logClientError(err, "Unable to write notification")
 				return
 			}
-		case <-time.After(globalSNSConnAlive): // Wait for global conn active seconds.
-			if err := writeNotification(w, dummyEvents); err != nil {
-				// FIXME - do not log for all errors.
-				errorIf(err, "Unable to write notification to client.")
+		case <-time.After(globalSNSConnAlive):
+			if err := writeNotification(w, emptyEvent); err != nil {
+				logClientError(err, "Unable to write empty notification")
 				return
 			}
 		}
@@ -346,12 +376,11 @@ func (api objectAPIHandlers) ListenBucketNotificationHandler(w http.ResponseWrit
 		},
 	}
 
-	// Setup a listening channel that will receive notifications
-	// from the RPC handler.
-	nEventCh := make(chan []NotificationEvent)
-	defer close(nEventCh)
+	// Setup a listen channel to receive notifications like
+	// s3:ObjectCreated, s3:ObjectDeleted etc.
+	nListenCh := newListenChan()
 	// Add channel for listener events
-	if err = globalEventNotifier.AddListenerChan(accountARN, nEventCh); err != nil {
+	if err = globalEventNotifier.AddListenerChan(accountARN, nListenCh); err != nil {
 		errorIf(err, "Error adding a listener!")
 		writeErrorResponse(w, toAPIErrorCode(err), r.URL)
 		return
@@ -361,8 +390,8 @@ func (api objectAPIHandlers) ListenBucketNotificationHandler(w http.ResponseWrit
 	defer globalEventNotifier.RemoveListenerChan(accountARN)
 
 	// Update topic config to bucket config and persist - as soon
-	// as this call compelets, events may start appearing in
-	// nEventCh
+	// as this call completes, events may start appearing in
+	// nListenCh
 	lc := listenerConfig{
 		TopicConfig:  *topicCfg,
 		TargetServer: targetServer,
@@ -378,8 +407,16 @@ func (api objectAPIHandlers) ListenBucketNotificationHandler(w http.ResponseWrit
 	// Add all common headers.
 	setCommonHeaders(w)
 
-	// Start sending bucket notifications.
-	sendBucketNotification(w, nEventCh)
+	// https://github.com/containous/traefik/issues/560
+	// https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events
+	//
+	// Proxies might buffer the connection to avoid this we
+	// need the proper MIME type before writing to client.
+	// This MIME header tells the proxies to avoid buffering
+	w.Header().Set("Content-Type", "text/event-stream")
+
+	// Start writing bucket notifications to ResponseWriter.
+	nListenCh.waitForListener(w)
 }
 
 // AddBucketListenerConfig - Updates on disk state of listeners, and
--- a/cmd/event-notifier.go
+++ b/cmd/event-notifier.go
@@ -60,7 +60,7 @@ type internalNotifier struct {
 	// Connected listeners is a map of listener ARNs to channels
 	// on which the ListenBucket API handler go routine is waiting
 	// for events to send to a client.
-	connectedListeners map[string]chan []NotificationEvent
+	connectedListeners map[string]*listenChan
 
 	rwMutex *sync.RWMutex
 }
@@ -206,7 +206,7 @@ func (en eventNotifier) GetInternalTarget(arn string) *listenerLogger {
 }
 
 // Set a new sns target for an input sns ARN.
-func (en *eventNotifier) AddListenerChan(snsARN string, listenerCh chan []NotificationEvent) error {
+func (en *eventNotifier) AddListenerChan(snsARN string, listenerCh *listenChan) error {
 	if listenerCh == nil {
 		return errInvalidArgument
 	}
@@ -229,9 +229,9 @@ func (en *eventNotifier) SendListenerEvent(arn string, event []NotificationEvent
 	en.internal.rwMutex.Lock()
 	defer en.internal.rwMutex.Unlock()
 
-	ch, ok := en.internal.connectedListeners[arn]
+	listenChan, ok := en.internal.connectedListeners[arn]
 	if ok {
-		ch <- event
+		listenChan.sendNotificationEvent(event)
 	}
 	// If the channel is not present we ignore the event.
 	return nil
@@ -833,7 +833,7 @@ func initEventNotifier(objAPI ObjectLayer) error {
 			rwMutex:            &sync.RWMutex{},
 			targets:            listenTargets,
 			listenerConfigs:    lConfigs,
-			connectedListeners: make(map[string]chan []NotificationEvent),
+			connectedListeners: make(map[string]*listenChan),
 		},
 	}
 
--- a/cmd/server_test.go
+++ b/cmd/server_test.go
@@ -522,32 +522,6 @@ func (s *TestSuiteCommon) TestListenBucketNotificationHandler(c *check) {
 	if s.signer == signerV4 {
 		verifyError(c, response, "XAmzContentSHA256Mismatch", "The provided 'x-amz-content-sha256' header does not match what was computed.", http.StatusBadRequest)
 	}
-
-	// Change global value from 5 second to 100millisecond.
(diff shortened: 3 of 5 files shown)


harshavardhana (approved):

cmd/event-notifier.go
```diff
@@ -230,7 +230,14 @@ func (en *eventNotifier) SendListenerEvent(arn string, event []NotificationEvent
...
 	ch, ok := en.internal.connectedListeners[arn]
 	if ok {
-		ch <- event
+		// To avoid blocking of send event over the channel
+		// while inside critical section.
+		//FIXME: The intermediary channel can be replaced by
+		//the actual http.ResponseWriter.
+		select {
+		case ch <- event:
+		default:
```
harshavardhana: Can we log an error here?

cmd/event-notifier.go
```diff
@@ -230,7 +230,14 @@ func (en *eventNotifier) SendListenerEvent(arn string, event []NotificationEvent
 
 	ch, ok := en.internal.connectedListeners[arn]
 	if ok {
-		ch <- event
+		// To avoid blocking of send event over the channel
+		// while inside critical section.
+		//FIXME: The intermediary channel can be replaced by
```
harshavardhana: Also comment can have a space after `//`

harshavardhana (changes requested):

kannappanr:
This fix doesn't do much for the behavior. Please see the attached file
[krisis_fix_failure_1.txt](https://github.com/minio/minio/files/1366644/krisis_fix_failure_1.txt)

harshavardhana:
> This fix doesn't do much for the behavior. Please see the attached file
> krisis_fix_failure_1.txt

I don't think its fully finished yet @kannappanr

harshavardhana:
> I don't think its fully finished yet @kannappanr

My bad it should work @kannappanr can you double check if you are running the right build? Because of this line
```
        /home/kannappan/gopath/src/github.com/minio/minio/cmd/event-notifier.go:233 +0x124
```
This still shows old code. Instead we have moved to line 238.

kannappanr:
Yes, with this build, I am unable to reproduce the issue.

harshavardhana:
Thanks @kannappanr

cmd/event-notifier.go
krisis: @harshavardhana What would the error message be? What do we intend to inform our users here?

harshavardhana:
```
errorIf(errors.New("blocked channel"), "Unable to send the event to %s, possibly no receiver", arn) 
```
This log will be removed by moving to http.ResponseWriter.

harshavardhana (approved):

cmd/event-notifier.go
```diff
@@ -230,7 +230,19 @@ func (en *eventNotifier) SendListenerEvent(arn string, event []NotificationEvent
...
-		ch <- event
+		// To avoid blocking of send event over the channel
+		// while inside critical section.
+		// FIXME: The intermediary channel can be replaced by
+		// the actual http.ResponseWriter.
+		select {
+		case ch <- event:
+		default:
+			// When there is no receiver at the other end
+			// of the event channel, the channel blocks
```
donatello: I don't think fix is correct - it would introduce a new problem. The ListenBucketNotificationHandler is the only go routine that listens on the channel and writes any received events to the http response writer. This means that between iterations of this loop, sending on the channel is blocked (because there is no listener). This in turn means that this select block will drop events, thus losing them.

See https://play.golang.org/p/oPmqFoaNA7 for a demonstration.

donatello (changes requested):
This bug needs some rethinking - possibly directly writing to the response-writer will work, but it needs to be synchronized with other writers.

cmd/event-notifier.go
harshavardhana: Loosing is acceptable in this scenario @donatello , we are aware that it will happen. So there is a comment on why it is okay ignore. The reasoning is simple this only happens on a unbuffered channel when indeed there is no receiver. This part of the code is also scheduled to be quickly fixed by moving to response writers instead of channel.

harshavardhana: Also this situation only happens when there are other active listeners.. which are non-blocking so it is highly unlikely that all of them are blocking and not receiving. Single listener `default:` should never get executed.

harshavardhana:
> This bug needs some rethinking - possibly directly writing to the response-writer will work, but it needs to be synchronized with other writers.

This is out of scope in this PR - IMO.

harshavardhana:
If the stop gap solution is not acceptable, i guess our current server is fine and it is pretty rare to reproduce this issue. We can fully implement the responseWriter style which does io.MultiWriter will work in this scenario and we can close this PR. @donatello @krisis

cmd/bucket-notification-handlers.go
```diff
@@ -226,23 +226,14 @@ func writeNotification(w http.ResponseWriter, notification map[string][]Notifica
...
 		select {
-		case events := <-arnListenerCh:
-			if err := writeNotification(w, map[string][]NotificationEvent{"Records": events}); err != nil {
-				errorIf(err, "Unable to write notification to client.")
-				return
-			}
-		case <-time.After(globalSNSConnAlive): // Wait for global conn active seconds.
-			if err := writeNotification(w, dummyEvents); err != nil {
+		case <-time.After(globalSNSConnAlive):
+			if err := writeNotification(w, emptyEvent); err != nil {
```
donatello: It is not clear to me if `ResponseWriter` is concurrent-safe. We are writing in one thread here, and in another when events are sent.

cmd/bucket-notification-handlers.go
```diff
@@ -226,25 +218,51 @@ func writeNotification(w http.ResponseWriter, notification map[string][]Notifica
...
+// unbuffered channels.
+func newListenChan() *listenChan {
+	return &listenChan{
+		doneCh: make(chan struct{}),
+		dataCh: make(chan []NotificationEvent),
+	}
+}
+
+// sendNotificationEvent sends notification events on the data channel
+// until doneCh is not closed
```
donatello: s/until doneCh is not closed/unless doneCh is closed/

cmd/bucket-notification-handlers.go
```diff
@@ -226,25 +218,51 @@ func writeNotification(w http.ResponseWriter, notification map[string][]Notifica
...
+	case <-l.doneCh:
+	// Blocks until receiver is available.
+	case l.dataCh <- events:
+	}
+}
+
+// waitForListener writes event notification OR whitespaces on
+// ResponseWriter until client closes connection
+func (l *listenChan) waitForListener(w http.ResponseWriter) {
+	emptyEvent := map[string][]NotificationEvent{"Records": nil}
```
donatello: Can be simplified to just: 
`emptyEvent := map[string][]NotificationEvent{}`
since nil is a zero value anyway.

cmd/bucket-notification-handlers.go
```diff
@@ -226,25 +218,51 @@ func writeNotification(w http.ResponseWriter, notification map[string][]Notifica
...
 				return
 			}
-		case <-time.After(globalSNSConnAlive): // Wait for global conn active seconds.
-			if err := writeNotification(w, dummyEvents); err != nil {
-				// FIXME - do not log for all errors.
-				errorIf(err, "Unable to write notification to client.")
+		case <-time.After(globalSNSConnAlive):
+			if err := writeNotification(w, emptyEvent); err != nil {
+				errorIf(err, "Unable to write empty notification")
+				close(l.doneCh)
```
donatello: The close statement can be factored out from both cases and made into a defer statement at the beginning of the function.

donatello (dismissed):
LGTM - suggesting minor changes.

deekoder:
can reviewers also examine code coverage failures?

cmd/bucket-notification-handlers.go
harshavardhana: You need "Records" otherwise the client unmarshal will fail. @donatello

cmd/bucket-notification-handlers.go
```diff
@@ -226,25 +218,51 @@ func writeNotification(w http.ResponseWriter, notification map[string][]Notifica
...
+		dataCh: make(chan []NotificationEvent),
+	}
+}
+
+// sendNotificationEvent sends notification events on the data channel
+// until doneCh is not closed
+func (l *listenChan) sendNotificationEvent(events []NotificationEvent) {
+	select {
+	// Returns immediately if receiver has quit.
+	case <-l.doneCh:
```
harshavardhana: We can add a test case for `doneCh` here.

cmd/bucket-notification-handlers.go
```diff
@@ -226,25 +218,51 @@ func writeNotification(w http.ResponseWriter, notification map[string][]Notifica
...
+				close(l.doneCh)
 				return
 			}
-		case <-time.After(globalSNSConnAlive): // Wait for global conn active seconds.
-			if err := writeNotification(w, dummyEvents); err != nil {
-				// FIXME - do not log for all errors.
-				errorIf(err, "Unable to write notification to client.")
+		case <-time.After(globalSNSConnAlive):
+			if err := writeNotification(w, emptyEvent); err != nil {
+				errorIf(err, "Unable to write empty notification")
```
harshavardhana: Perhaps this log is not needed, most probably client actually died/disconnected here.. Its a good case of us.

harshavardhana (changes requested):

cmd/bucket-notification-handlers.go
krisis: done

cmd/bucket-notification-handlers.go
krisis: done

cmd/bucket-notification-handlers.go
krisis: I have added a test case.

cmd/bucket-notification-handlers.go
krisis: This log captures failure of server to write to the client when it was writing empty notification events as opposed actual events like in the previous case statement. Both these error message indicate that the client has quit, but the different error messages allow us to uniquely identify the exact case.

harshavardhana: Correct but client dead is a good case so we don't need to log that message, any other we can log.

krisis: Both the error logs assume that any error during Write on http.ResponseWriter is due to client quitting ListenBucketNotification. Are you suggesting we should only log error values other than the `ECONNRESET` equivalent `net.OpError`?

harshavardhana: Yes..any broken pipe. We can ignore where the connection is dead sort of.

donatello:
This change LGTM but the build seems to be failing in Windows. Looks like some kind of file-lock related issue (unrelated to this change itself?).

krisis:
@harshavardhana @donatello could you guys review again?

harshavardhana (approved):
LGTM tested and working.

cmd/server_test.go
```diff
@@ -522,32 +522,6 @@ func (s *TestSuiteCommon) TestListenBucketNotificationHandler(c *check) {
 	if s.signer == signerV4 {
 		verifyError(c, response, "XAmzContentSHA256Mismatch", "The provided 'x-amz-content-sha256' header does not match what was computed.", http.StatusBadRequest)
 	}
-
-	// Change global value from 5 second to 100millisecond.
-	globalSNSConnAlive = 100 * time.Millisecond
-	req, err = newTestSignedRequest("GET",
```
harshavardhana, line 528: Why not leave this here for now? @krisis - this can be fixed when the Bala's refactor change is merged.

harshavardhana (changes requested):

cmd/server_test.go
krisis: It serves no purpose. It sends a http request to the listenBucketNotication handler and nothing else. So, there is never a bad time to remove it.

donatello (approved):

harshavardhana (approved):

PR #5029, opened 2017-10-07
Figure out projectID for GCS automatically from credentials.json

### LINKED ISSUE #5027: Figure out projectID for GCS automatically from credentials.json
when GOOGLE_APPLICATION_CREDENTIALS=/path/to/credentials.json is set where credentials.json is downloaded from https://console.developers.google.com/project/_/apis/credentials then we don't need the user to pass the project ID additionally in the command line. Because we can parse projectID from the credentials.json

## Expected Behavior
project ID in the command line is optional if GOOGLE_APPLICATION_CREDENTIALS is set

## Current Behavior
project ID is currently mandatory in the command line.

## Your Environment

* Server type and version:
* Link to your project:

krishnasrinivas:
fixes #5027

## Description

We parse projectID from credentials.json file, hence it need not be passed in the command line if GOOGLE_APPLICATION_CREDENTIALS is set to the path of credentials.json

## How Has This Been Tested?

manual, unit testing

## Types of changes

## Checklist:

### DIFF
--- a/cmd/gateway-gcs.go
+++ b/cmd/gateway-gcs.go
@@ -23,7 +23,9 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"math"
+	"os"
 	"regexp"
 	"strings"
 	"time"
@@ -64,6 +66,9 @@ const (
 
 	// The cleanup routine deletes files older than 2 weeks in minio.sys.tmp
 	gcsMultipartExpiry = time.Hour * 24 * 14
+
+	// Project ID key in credentials.json
+	gcsProjectIDKey = "project_id"
 )
 
 // Stored in gcs.json - Contents of this file is not used anywhere. It can be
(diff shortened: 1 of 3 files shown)


pkg/http/listener_test.go
```diff
@@ -778,12 +777,14 @@ func TestHTTPListenerAcceptParallel(t *testing.T) {
 			if err != nil {
 				t.Fatalf("Test %d: accept: expected = <nil>, got = %v", i+1, err)
 			}
```
harshavardhana: Don't do these changes in your PR @krishnasrinivas - you should send a separate fix.

cmd/gateway-gcs.go
```diff
@@ -248,11 +252,36 @@ type gcsGateway struct {
 
 const googleStorageEndpoint = "storage.googleapis.com"
 
+// Returns projectID from the GOOGLE_APPLICATION_CREDENTIALS file.
+func gcsParseProjectID(path string) (projectID string, err error) {
```
balamurugana: `credsJSONFile` would be more clear than `path`

cmd/gateway-main.go
```diff
@@ -277,7 +280,12 @@ func gcsGatewayMain(ctx *cli.Context) {
 		cli.ShowCommandHelpAndExit(ctx, "gcs", 1)
 	}
 
-	if !isValidGCSProjectIDFormat(ctx.Args().First()) {
+	projectID := ctx.Args().First()
+	if projectID == "" && os.Getenv("GOOGLE_APPLICATION_CREDENTIALS") == "" {
+		errorIf(errGCSProjectIDNotFound, "project-id should be provided as argument or GOOGLE_APPLICATION_CREDENTIALS should be set with path to credentials.json")
+		cli.ShowCommandHelpAndExit(ctx, "gcs", 1)
+	}
+	if projectID != "" && !isValidGCSProjectIDFormat(projectID) {
```
balamurugana, line 288: As project ID from command line has higher precedence over env var `GOOGLE_APPLICATION_CREDENTIALS`.   I think below `if` block would be clear

```go
credsJSONFile := os.Getenv("GOOGLE_APPLICATION_CREDENTIALS")
if projectID == "" {
    if credsJSONFIle == "" {
       fatalIf(...)
    }
    if projectID, err = gcsParseProjectID(credsJSONFile); err != nil {
        fatalIf(err, ...)
    }
}
```

This way `newGCSGateway()` could use `projectID` argument without knowing about env var `GOOGLE_APPLICATION_CREDENTIALS`

balamurugana (changes requested):

cmd/gateway-main.go
krishnasrinivas: I tried to do this @balamurugana but our source code structure does not allow this flow, it needs a refactor to allow it.

i.e I tried to make the flow such that newGCSGateway() does not access GOOGLE_APPLICATION_CREDENTIALS but it was not possible. It needs a separate refactor of the functions in gateway-main.go

cmd/gateway-gcs.go
krishnasrinivas: I have renamed it to `credsFile`

cmd/gateway-main.go
balamurugana: OK. Probably we need to open an issue?

balamurugana (approved):

harshavardhana (approved):

PR #5034, opened 2017-10-07
tests: Remove dependency on check.v1

harshavardhana:
## Description
tests: An attempt to remove gocheck deps.

## Motivation and Context
This PR addresses a long standing dependency on
`gopkg.in/check.v1` project used for our tests.
All tests are re-written to use the go default
testing framework instead.

There was no reason for us to use an external
package where Go tools are sufficient for this.

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:


vadmeste:
A little warning:
`main_test.go:24:6: TestCheckGoVersion is unused `

harshavardhana:
That is a false positive ignore it for now. It's `deadcode` bug happens on master as well.

cmd/net_test.go
```diff
@@ -175,6 +175,21 @@ func TestGetAPIEndpoints(t *testing.T) {
 	}
 }
 
+// Ask the kernel for a free open port.
+func getFreePort() string {
```
balamurugana, line 179: Rename this function to `mustGetFreePort()`.

harshavardhana: I would defer to make such changes in a dependency removal PR.

cmd/browser-peer-rpc_test.go
```diff
@@ -41,9 +41,8 @@ func (s *TestRPCBrowserPeerSuite) SetUpSuite(c *testing.T) {
 }
 
 // No longer used with gocheck, but used in explicit teardown code in
-// each test function. // Called implicitly by "gopkg.in/check.v1"
-// after all tests are run.
-func (s *TestRPCBrowserPeerSuite) TearDownSuite(c *testing.T) {
+// each test function. Called implicitly by after all tests are run.
```
kannappanr: Maybe a typo. Comment needs to be clear.

harshavardhana: Sure

vadmeste (approved):
LGTM

cmd/object_api_suite_test.go
```diff
@@ -1,5 +1,5 @@
 /*
- * Minio Cloud Storage, (C) 2015, 2016 Minio, Inc.
+ * Minio Cloud Storage, (C) 2015, 2016, 2017 Minio, Int.
```
kannappanr: Typo. Should be Minio Cloud Storage, (C) 2015, 2016, 2017 Minio, Inc.

pkg/quick/quick_test.go
```diff
@@ -19,77 +19,95 @@
...
 
 	err = ioutil.WriteFile("test.json", []byte("{ \"version\":2,"), 0644)
-	c.Assert(err, IsNil)
+	if err != nil {
+		t.Fatal(err)
+	}
 
 	_, err = GetVersion("test.json")
-	c.Assert(err, Not(IsNil))
+	if err == nil {
```
kannappanr, line 71: Log can have more information. That way, the user knows where exactly this failure happened and why. This needs to be checked in all the places where there is not error thrown, but the actual test has failed.

pkg/quick/quick_test.go
```diff
@@ -130,38 +154,63 @@ func (s *MySuite) TestLoadFile(c *C) {
 	}
 	saveMe := myStruct{}
 	_, err := Load("test.json", &saveMe)
-	c.Assert(err, Not(IsNil))
+	if err == nil {
```
kannappanr, line 157: There is no error to log. Should be changed to a text, more like, Unexpected, should have failed while loading test.json.

harshavardhana: Fixed.

pkg/quick/quick_test.go
harshavardhana: Fixed

pkg/quick/quick_test.go
```diff
@@ -189,28 +238,43 @@ directories:
 
 	// Save format using
 	config, err := New(&saveMe)
-	c.Assert(err, IsNil)
-	c.Assert(config, Not(IsNil))
+	if err != nil {
+		t.Fatal(err)
+	}
```
kannappanr, line 244: check for config itself not being null, has been removed. Is it OK, if we check just for no error case?

pkg/quick/quick_test.go
```diff
@@ -240,28 +304,42 @@ func (s *MySuite) TestJSONFormat(c *C) {
...
```
kannappanr, line 311: Similar comment to the previous one. Is it OK, to not check for config being not nil

pkg/safe/safe_test.go
```diff
@@ -23,82 +23,172 @@ import (
...
+
 	f, err := CreateFile(path.Join(s.root, "testfile-abort"))
-	c.Assert(err, IsNil)
+	if err != nil {
+		t.Fatal(err)
+	}
+
 	_, err = os2.Stat(path.Join(s.root, "testfile-abort"))
-	c.Assert(err, Not(IsNil))
+	if !os.IsNotExist(err) {
```
kannappanr, line 58: What other errors are OK, to be skipped?

pkg/safe/safe_test.go
```diff
@@ -23,82 +23,172 @@ import (
...
+
 	err = f.Close()
-	c.Assert(err, IsNil)
+	if err != nil {
+		t.Fatal(err)
+	}
+
 	_, err = os2.Stat(path.Join(s.root, "testfile-close"))
-	c.Assert(err, IsNil)
+	if err != nil {
```
kannappanr, line 94: Why are we not using the following check

if !os.IsNotExist(err) {

like in the other places above?

pkg/quick/quick_test.go
harshavardhana: That is enough in this scenario.

pkg/safe/safe_test.go
```diff
@@ -23,82 +23,172 @@ import (
...
-	c.Assert(err.Error(), Equals, "close on closed file")
+	if err != nil {
+		if err.Error() != "close on closed file" {
+			t.Fatal(err)
+		}
+	}
+
 	_, err = os2.Stat(path.Join(s.root, "testfile-safe"))
-	c.Assert(err, IsNil)
+	if err != nil {
```
kannappanr, line 146: why not check for ``` if !os.IsNotExist(err) { ``` here, instead of just ``` err != nil ```

pkg/quick/quick_test.go
harshavardhana: Same as above.

pkg/safe/safe_test.go
harshavardhana: Almost all other errors..

harshavardhana: Here if we get NotExist is a correct error but if its anything other than NotExist then its a bug.

pkg/safe/safe_test.go
harshavardhana: In this case we don't expect error, we expect the file be available.. and successful.

pkg/safe/safe_test.go
harshavardhana: In this case we don't expect error, we expect the file be available.. and successful.

kannappanr (approved):
LGTM

PR #5038, opened 2017-10-08
Fix azure metadata handling for all incoming PUT requests

harshavardhana:
## Description
s3cmd cli fails when trying to upload a file to azure gateway.
Previous fixes in azure to handle client side encryption alone
did not completely address the problem.

We need to possibilly convert all the x-amz-meta-\<name\>
i.e specifically \<name\> should be converted into a
C# identifier as mentioned in the docs for `put-blob`.

https://docs.microsoft.com/en-us/rest/api/storageservices/put-blob

```
s3cmd put README.md s3://myanis/
upload: 'README.md' -> 's3://myanis/README.md'  [1 of 1]
 4598 of 4598   100% in    0s    47.24 kB/s  done
upload: 'README.md' -> 's3://myanis/README.md'  [1 of 1]
 4598 of 4598   100% in    0s    50.47 kB/s  done
ERROR: S3 error: 400 (InvalidArgument): Your metadata headers are not supported.
```

There is a separate issue with s3cmd after this fix is applied where
the ETag is wrongly validated https://github.com/s3tools/s3cmd/issues/880
But that is an upstream s3cmd problem which wrongly interprets ETag
to be md5sum of the content that was uploaded.

## Motivation and Context
Compatibility fixes.

## How Has This Been Tested?
Manually using s3cmd.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/gateway-azure.go
+++ b/cmd/gateway-azure.go
@@ -55,31 +55,37 @@ const metadataObjectNameTemplate = globalMinioSysTmp + "multipart/v1/%s.%x/azure
 //
 // Header names are canonicalized as in http.Header.
 func s3MetaToAzureProperties(s3Metadata map[string]string) (storage.BlobMetadata,
-	storage.BlobProperties) {
-
-	// Azure does not permit user-defined metadata key names to
-	// contain hyphens. So we map hyphens to underscores for
-	// encryption headers. More such headers may be added in the
-	// future.
-	gatewayHeaders := map[string]string{
-		"X-Amz-Meta-X-Amz-Key":     "X-Amz-Meta-x_minio_key",
-		"X-Amz-Meta-X-Amz-Matdesc": "X-Amz-Meta-x_minio_matdesc",
-		"X-Amz-Meta-X-Amz-Iv":      "X-Amz-Meta-x_minio_iv",
+	storage.BlobProperties, error) {
+	for k := range s3Metadata {
+		if strings.Contains(k, "--") {
+			return storage.BlobMetadata{}, storage.BlobProperties{}, traceError(UnsupportedMetadata{})
+		}
 	}
 
+	// Encoding technique for each key is used here is as follows
+	// Each '-' is converted to '_'
+	// Each '_' is converted to '__'
+	// With this basic assumption here are some of the expected
+	// translations for these keys.
+	// i: 'x-S3cmd_attrs' -> o: 'x_s3cmd__attrs' (mixed)
+	// i: 'x__test__value' -> o: 'x____test____value' (double '_')
+	encodeKey := func(key string) string {
+		tokens := strings.Split(key, "_")
+		for i := range tokens {
+			tokens[i] = strings.Replace(tokens[i], "-", "_", -1)
+		}
+		return strings.Join(tokens, "__")
+	}
 	var blobMeta storage.BlobMetadata = make(map[string]string)
 	var props storage.BlobProperties
 	for k, v := range s3Metadata {
 		k = http.CanonicalHeaderKey(k)
-		if nk, ok := gatewayHeaders[k]; ok {
-			k = nk
-		}
 		switch {
 		case strings.HasPrefix(k, "X-Amz-Meta-"):
 			// Strip header prefix, to let Azure SDK
 			// handle it for storage.
 			k = strings.Replace(k, "X-Amz-Meta-", "", 1)
-			blobMeta[k] = v
+			blobMeta[encodeKey(k)] = v
 
 		// All cases below, extract common metadata that is
 		// accepted by S3 into BlobProperties for setting on
@@ -100,7 +106,7 @@ func s3MetaToAzureProperties(s3Metadata map[string]string) (storage.BlobMetadata
 			props.ContentType = v
 		}
 	}
-	return blobMeta, props
+	return blobMeta, props, nil
 }
 
 // azurePropertiesToS3Meta converts Azure metadata/properties to S3
@@ -108,22 +114,26 @@ func s3MetaToAzureProperties(s3Metadata map[string]string) (storage.BlobMetadata
 // `.GetMetadata()` lower-cases all header keys, so this is taken into
 // account by this function.
 func azurePropertiesToS3Meta(meta storage.BlobMetadata, props storage.BlobProperties) map[string]string {
-	// Remap underscores to hyphens to restore encryption
-	// headers. See s3MetaToAzureProperties for details.
-	gatewayHeaders := map[string]string{
-		"X-Amz-Meta-x_minio_key":     "X-Amz-Meta-X-Amz-Key",
-		"X-Amz-Meta-x_minio_matdesc": "X-Amz-Meta-X-Amz-Matdesc",
-		"X-Amz-Meta-x_minio_iv":      "X-Amz-Meta-X-Amz-Iv",
+	// Decoding technique for each key is used here is as follows
+	// Each '_' is converted to '-'
+	// Each '__' is converted to '_'
+	// With this basic assumption here are some of the expected
+	// translations for these keys.
+	// i: 'x_s3cmd__attrs' -> o: 'x-s3cmd_attrs' (mixed)
+	// i: 'x____test____value' -> o: 'x__test__value' (double '_')
+	decodeKey := func(key string) string {
+		tokens := strings.Split(key, "__")
+		for i := range tokens {
+			tokens[i] = strings.Replace(tokens[i], "_", "-", -1)
+		}
+		return strings.Join(tokens, "_")
 	}
 
 	s3Metadata := make(map[string]string)
 	for k, v := range meta {
 		// k's `x-ms-meta-` prefix is already stripped by
 		// Azure SDK, so we add the AMZ prefix.
-		k = "X-Amz-Meta-" + k
-		if nk, ok := gatewayHeaders[k]; ok {
-			k = nk
-		}
+		k = "X-Amz-Meta-" + decodeKey(k)
 		k = http.CanonicalHeaderKey(k)
 		s3Metadata[k] = v
 	}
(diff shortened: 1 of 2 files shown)


cmd/gateway-azure.go
```diff
@@ -79,6 +79,7 @@ func s3MetaToAzureProperties(s3Metadata map[string]string) (storage.BlobMetadata
 			// Strip header prefix, to let Azure SDK
 			// handle it for storage.
 			k = strings.Replace(k, "X-Amz-Meta-", "", 1)
+			k = strings.Replace(strings.ToLower(k), "-", "_", -1)
```
balamurugana: We decided not to convert all headers into C# friendly names.   This PR requires further discussion.

balamurugana (changes requested):

cmd/gateway-azure.go
harshavardhana: Well our assumptions are wrong. I don't see why we decided what we decided. There are clearly many applications that breaks without this change.

balamurugana: This is known behavior. That is consciously decided.  However I blocked it for further discussion.

harshavardhana: Can you clarify here why such an assumption is made? I am not aware why we decided to do this.

harshavardhana:
FWIW  submitted relevant tests for s3cmd as well https://github.com/minio/mint/pull/182

cmd/gateway-azure.go
harshavardhana: I would like to know the assumptions and update relevant documents on why this was assumed to be okay and expected behavior.

harshavardhana:
Removing blocked as we have agreed to support these scenarios.

cmd/gateway-azure.go
```diff
@@ -57,29 +57,21 @@ const metadataObjectNameTemplate = globalMinioSysTmp + "multipart/v1/%s.%x/azure
...
 
-	// Azure does not permit user-defined metadata key names to
-	// contain hyphens. So we map hyphens to underscores for
-	// encryption headers. More such headers may be added in the
-	// future.
-	gatewayHeaders := map[string]string{
-		"X-Amz-Meta-X-Amz-Key":     "X-Amz-Meta-x_minio_key",
-		"X-Amz-Meta-X-Amz-Matdesc": "X-Amz-Meta-x_minio_matdesc",
-		"X-Amz-Meta-X-Amz-Iv":      "X-Amz-Meta-x_minio_iv",
-	}
```
balamurugana: We need to keep this to have identified for minio gateway usage.

cmd/gateway-azure.go
```diff
@@ -57,29 +57,21 @@ const metadataObjectNameTemplate = globalMinioSysTmp + "multipart/v1/%s.%x/azure
...
 		switch {
 		case strings.HasPrefix(k, "X-Amz-Meta-"):
 			// Strip header prefix, to let Azure SDK
 			// handle it for storage.
 			k = strings.Replace(k, "X-Amz-Meta-", "", 1)
-			blobMeta[k] = v
+			if strings.Contains(k, "_") {
+				k = strings.Replace(k, "_", "__", -1)
+			} else {
+				k = strings.Replace(k, "-", "_", -1)
```
balamurugana: You would need to do `_` and `-` translation irrespective of the condition i.e. it should work for `X-Amz-Meta-My-Metadata_Name`

cmd/gateway-azure.go
```diff
@@ -108,21 +100,15 @@ func s3MetaToAzureProperties(s3Metadata map[string]string) (storage.BlobMetadata
...
 	for k, v := range meta {
 		// k's `x-ms-meta-` prefix is already stripped by
 		// Azure SDK, so we add the AMZ prefix.
 		k = "X-Amz-Meta-" + k
-		if nk, ok := gatewayHeaders[k]; ok {
-			k = nk
+		if strings.Contains(k, "__") {
+			k = strings.Replace(k, "__", "_", -1)
+		} else {
+			k = strings.Replace(k, "_", "-", -1)
```
balamurugana: This won't work for case like `My_Metadata__Name` which is expected to be `My-Metadata_name`

balamurugana (changes requested):

cmd/gateway-azure.go
```diff
@@ -57,29 +70,24 @@ const metadataObjectNameTemplate = globalMinioSysTmp + "multipart/v1/%s.%x/azure
...
-	gatewayHeaders := map[string]string{
-		"X-Amz-Meta-X-Amz-Key":     "X-Amz-Meta-x_minio_key",
-		"X-Amz-Meta-X-Amz-Matdesc": "X-Amz-Meta-x_minio_matdesc",
-		"X-Amz-Meta-X-Amz-Iv":      "X-Amz-Meta-x_minio_iv",
+	encodeKey := func(key string) string {
+		tokens := strings.Split(key, "_")
+		for i := range tokens {
+			tokens[i] = strings.Replace(tokens[i], "-", "_", -1)
+		}
+		return strings.Join(tokens, "__")
```
donatello, line 77: A commented description of the encoding is required so it is clear to future readers that the intended encoding is indeed correctly implemented.

cmd/gateway-azure.go
```diff
@@ -40,6 +40,19 @@ const globalAzureAPIVersion = "2016-05-31"
...
+// without introducing a new encoding style.
+func isValidAzureMetadata(s3Metadata map[string]string) (match bool) {
+	match = false
+	for k := range s3Metadata {
+		if strings.Contains(k, "--") {
+			match = true
+			break
+		}
+	}
+	return match
```
donatello: The naming of the return variable can be avoided for a more concise implementation using just a single return statement in the if block and a return statement at the end of the function.

cmd/gateway-azure.go
```diff
@@ -40,6 +40,19 @@ const globalAzureAPIVersion = "2016-05-31"
 const azureBlockSize = 100 * humanize.MiByte
 const metadataObjectNameTemplate = globalMinioSysTmp + "multipart/v1/%s.%x/azure.json"
 
+// Tests if s3Metadata has double "-", we cannot handle this
+// without introducing a new encoding style.
+func isValidAzureMetadata(s3Metadata map[string]string) (match bool) {
```
donatello: The name is inappropriate - what is meant is `hasUnsupportedAzureMetadata`.

cmd/gateway-azure.go
```diff
@@ -512,7 +517,13 @@ func (a *azureObjects) GetObjectInfo(bucket, object string) (objInfo ObjectInfo,
 // PutObject - Create a new blob with the incoming data,
 // uses Azure equivalent CreateBlockBlobFromReader.
 func (a *azureObjects) PutObject(bucket, object string, data *HashReader, metadata map[string]string) (objInfo ObjectInfo, err error) {
+	if metadata == nil {
+		metadata = make(map[string]string)
```
donatello: Neither deleting from a nil map or ranging over a nil map are errors - so it is not clear why this allocation is needed.

cmd/gateway-azure.go
```diff
@@ -590,6 +601,9 @@ func (a *azureObjects) NewMultipartUpload(bucket, object string, metadata map[st
 	if metadata == nil {
 		metadata = make(map[string]string)
```
donatello: Neither deleting from a nil map or ranging over a nil map are errors - so it is not clear why this allocation is needed.

donatello (changes requested):

cmd/gateway-azure.go
harshavardhana: Yes i know, old code not removing in this PR.

cmd/gateway-azure.go
harshavardhana: Yes i know, old code not removing in this PR.

cmd/gateway-azure.go
harshavardhana: how is that concise?

cmd/gateway-azure.go
harshavardhana: ah correct will fix.

cmd/gateway-azure.go
balamurugana: @harshavardhana I guess below code would avoid the confusion
```go
func hasUnsupportedAzureMetadata(s3Metadata map[string]string) (matched bool) {
	for k := range s3Metadata {
		if matched = strings.Contains(k, "--"); matched {
			break
		}
	}
	return matched
}
```

Another suggestion is this function can be a closure to `s3MetaToAzureProperties()` and `s3MetaToAzureProperties()` could return error additionally.  This way we could avoid repetitive check in other places.

balamurugana (changes requested):

cmd/gateway-azure.go
harshavardhana: > Another suggestion is this function can be a closure to s3MetaToAzureProperties() and s3MetaToAzureProperties() could return error additionally. This way we could avoid repetitive check in other places.

I wanted to avoid too many changes in this PR, but i guess we can do that.

balamurugana: I think this approach would come with minimal diff i.e. within `s3MetaToAzureProperties()`

cmd/gateway-azure.go
```diff
@@ -55,31 +55,38 @@ const metadataObjectNameTemplate = globalMinioSysTmp + "multipart/v1/%s.%x/azure
...
-			k = nk
-		}
 		switch {
 		case strings.HasPrefix(k, "X-Amz-Meta-"):
 			// Strip header prefix, to let Azure SDK
 			// handle it for storage.
 			k = strings.Replace(k, "X-Amz-Meta-", "", 1)
-			blobMeta[k] = v
+			k = encodeKey(k)
+			blobMeta[strings.ToLower(k)] = v
```
balamurugana: My recommendation is avoid `strings.ToLower()` which should be taken care by next layer.  This way you could do `blobMeta[encodeKey(k)] = v`

harshavardhana: done

balamurugana (approved):
LGTM

donatello (approved):

Issue #4998, opened 2017-09-29
Support all values in minio config to be honored through envs


harshavardhana:
Currently values of access key, secret key, region and browser flag are supported by environment variables `MINIO_ACCESS_KEY`, `MINIO_SECRET_KEY`, `MINIO_BROWSER` and `MINIO_REGION` respectively.

We would need to leverage the support to all fields in minio configuration file as environment variables.  Below is the table indicates JSON field and respective environement variable.
```
                  JSON field                |     Environment variable
--------------------------------------------+--------------------------------
{                                           |
    "version": "19",                        |
    "credential": {                         |
        "accessKey": "myuser",              | MINIO_ACCESS_KEY
        "secretKey": "mypassword",          | MINIO_SECRET_KEY
    },                                      | MINIO_REGION
    "region": "",                           | MINIO_BROWSER
    "browser": "on",                        |
    "logger": {                             |
        "console": {                        |
            "enable": true                  | MINIO_LOGGER_CONSOLE_ENABLE
        },                                  |
        "file": {                           |
            "enable": false,                | MINIO_LOGGER_FILE_ENABLE
            "filename": ""                  | MINIO_LOGGER_FILE_FILENAME
        }                                   |
    },                                      |
    "notify": {                             |
        "amqp": {                           |
            "1": {                          |
                "enable": false,            | MINIO_NOTIFY_AMQP_<String>_ENABLE
                "url": "",                  | MINIO_NOTIFY_AMQP_<String>_URL
                "exchange": "",             | MINIO_NOTIFY_AMQP_<String>_EXCHANGE
                "routingKey": "",           | MINIO_NOTIFY_AMQP_<String>_ROUTINGKEY
                "exchangeType": "",         | MINIO_NOTIFY_AMQP_<String>_EXCHANGETYPE
                "deliveryMode": 0,          | MINIO_NOTIFY_AMQP_<String>_DELIVERYMODE
                "mandatory": false,         | MINIO_NOTIFY_AMQP_<String>_MANDATORY
                "immediate": false,         | MINIO_NOTIFY_AMQP_<String>_IMMEDIATE
                "durable": false,           | MINIO_NOTIFY_AMQP_<String>_DURABLE
                "internal": false,          | MINIO_NOTIFY_AMQP_<String>_INTERNAL
                "noWait": false,            | MINIO_NOTIFY_AMQP_<String>_NOWAIT
[...]

PR #5039, opened 2017-10-08
Support all values in minio config to be honored through envs

zerkms:
This is a possible solution for the #4998. It does not have all the necessary unit tests (it would take some time to write heaps of those), so this PR at the moment is more to confirm I'm moving the right direction.

## Description

I made a huge refactoring on how the configuration is managed: I think this implementation is slightly more clear and is scalable to adding more environment variables.

## Motivation and Context

Motivation is explained in #4998

## How Has This Been Tested?

During development it was tested manually, and unit tests are WIP (it would take time to implement those and I need a confirmation this is an acceptable implementation).

## Types of changes

## Checklist:

### DIFF
--- a/cmd/config-env.go
+++ b/cmd/config-env.go
@@ -1,0 +1,590 @@
+package cmd
+
+import (
+	"errors"
+	"os"
+	"strconv"
+)
+
+type envConfigV19 struct {
+	Credential *credential
+	Region     *string
+	Browser    *BrowserFlag
+
+	Notify envNotifier
+}
+
+type envNotifier struct {
+	AMQP          envNotifyAMQP
+	NATS          envNotifyNATS
+	ElasticSearch envNotifyElasticSearch
+	Redis         envNotifyRedis
+	PostgreSQL    envNotifyPostgreSQL
+	Kafka         envNotifyKafka
+	Webhook       envNotifyWebhook
+	MySQL         envNotifyMySQL
+	MQTT          envNotifyMQTT
+}
+
+type envNotifyAMQP struct {
+	Enable       *bool
+	URL          *string
+	Exchange     *string
+	RoutingKey   *string
+	ExchangeType *string
+	DeliveryMode *uint8
+	Mandatory    *bool
+	Immediate    *bool
+	Durable      *bool
+	Internal     *bool
+	NoWait       *bool
+	AutoDeleted  *bool
+}
+
+type envNotifyNATSStreaming struct {
+	Enable             *bool
+	ClusterID          *string
+	ClientID           *string
+	Async              *bool
+	MaxPubAcksInflight *int
+}
+
+type envNotifyNATS struct {
+	Enable       *bool
+	Address      *string
+	Subject      *string
+	Username     *string
+	Password     *string
+	Token        *string
+	Secure       *bool
+	PingInterval *int64
+	Streaming    envNotifyNATSStreaming
+}
+
+type envNotifyElasticSearch struct {
+	Enable *bool
+	Format *string
+	URL    *string
+	Index  *string
+}
+
+type envNotifyRedis struct {
+	Enable   *bool
+	Format   *string
+	Addr     *string
+	Password *string
+	Key      *string
+}
+
+type envNotifyPostgreSQL struct {
+	Enable           *bool
+	Format           *string
+	ConnectionString *string
+	Table            *string
+	Host             *string
+	Port             *string
+	User             *string
+	Password         *string
+	Database         *string
+}
+
+type envNotifyKafka struct {
+	Enable *bool
+	Topic *string
+}
+
+type envNotifyWebhook struct {
+	Enable   *bool
+	Endpoint *string
+}
+
+type envNotifyMySQL struct {
+	Enable    *bool
+	Format    *string
+	DsnString *string
+	Table     *string
+	Host      *string
+	Port      *string
+	User      *string
+	Password  *string
+	Database  *string
+}
+
+type envNotifyMQTT struct {
+	Enable   *bool
+	Broker   *string
+	Topic    *string
+	QoS      *int
+	ClientID *string
+	User     *string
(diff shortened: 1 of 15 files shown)


harshavardhana (commented):
PR title needs a change should follow one liner description of the PR itself.   PR titles like "A possible solution ... " are not recommended..

cmd/config-env.go
```diff
@@ -0,0 +1,590 @@
+package cmd
+
+import (
+	"errors"
+	"os"
+	"strconv"
+)
+
+type envConfigV19 struct {
```
balamurugana, line 9: We could populate values from environment variables to struct `serverConfigV19`.  No new type is required.

balamurugana (changes requested):

cmd/config-env.go
zerkms: @balamurugana values of what structure? You cannot fill the server config directly: some other places rely on the knowledge whether it comes from environment or config.

balamurugana: @zerkms Actually I am trying to say we would need to add new function in `config-v19.go` something like
```go
func loadConfigEnv() error {
    // Create new `serverConfigV19{}` and fill each fields from the ENV.
    //  If any non-optional field is missing or has wrong value, return error.
}
```

In `server-main.go` we call this funcion first, on error, read config file.  new type `envConfigV19` is not necessary.

vadmeste:
We should probably add a new unit test which checks that all JSON keys in config can be converted to a variable name in Bash

cmd/config-env.go
zerkms: @balamurugana and as I mentioned before - it won't work: the current implementation (the current `master`) relies on the knowledge whether a value came from the environment or from the config file.

Example: https://github.com/minio/minio/blob/master/cmd/admin-handlers.go#L150

So you MUST keep track of environment variables somewhere somehow. At the moment it is implemented as bunch of global variables. I refactored it to a single type and a single global variable, which is presumably more clearer.

zerkms:
@vadmeste totally: I would add all the necessary tests as soon as the solution is proven to be acceptable (it's a lot of work and I don't want to waste my time if the implementation is rejected).

cmd/config-env.go
balamurugana: > So you MUST keep track of environment variables somewhere somehow. At the moment it is implemented as bunch of global variables.

Yes. There are three flags (`globalIsEnvBrowser`, `globalIsEnvCreds` and `globalIsEnvRegion`) we have now to know whether those values are from env vars.

> I refactored it to a single type and a single global variable, which is presumably more clearer.

Could you do this change as separate PR (Previously I was attempting to have and removing global variables led to 1000+ lines change) ?  We encourage to have multiple small PRs for easy reviewing.

Also you are welcome to minio.slack.com where we can have longer discussions.

zerkms: @balamurugana

> Could you do this change as separate PR

I though about it initially, but then I would have to reimplement the whole thing twice (given that I do like the solution with a new separated type for environment variables).

Anyway, if you think 

```go
func loadConfigEnv() error {
    // Create new `serverConfigV19{}` and fill each fields from the ENV.
    //  If any non-optional field is missing or has wrong value, return error.
}
```

would work I might invest some of my time more to reimplement it, but I personally don't like it at all:

1. It mutates the global variables (hence it's harder to test)
2. It is validated separately (in my implementation the validation is done just once in one place): maintaining something twice is twice as more error prone than doing something once
3. Because of separate validation - you won't be able to override just some config values but have to provide a full "block" of environment variables: so it makes it much less flexible in use.

PS: I'm in the slack :-)

deekoder:
@balamurugana @krishnasrinivas can this be re reviewed again and help @zerkms  complete this PR?

zerkms:
@deekoder at the moment I decided for myself that I will provide an alternative solution that follows the ideas from @balamurugana and see which is better so that we could then make one nice one out of the two.

Sorry for the delay - I've had a terrible issues with windows 10 fall creators upgrade and it was consuming all my spare time recently.

balamurugana:
@deekoder adding this feature into our configuration system is not straight forward, I was started to work on giving a skeleton code to @zerkms, unfortunately I couldn't with current code.  Hence I am working on https://github.com/balamurugana/minio/tree/make-notification-package

ehiggs:
@zerkms, @balamurugana, any updates?

Thanks!

zerkms:
@ehiggs it's the middle of the unusually hot summer here, so my time to spend with a computer limited (to almost zero) 🌞🚵‍♂️

balamurugana:
I have a separate PR for notification system https://github.com/minio/minio/pull/5294 which makes things easier to add support to ENV support by adding methods similar to `MarshalJSON()/UnmarshalJSON()`, `MarshalXML()/UnmarshalXML()` etc.

The PR review is in progress.

deekoder:
@zerkms it should be easier now with @balamurugana PR on simplifying notification subsystem https://github.com/minio/minio/pull/5294 We think we will merge 5294 soon. So we would love your PR again. Sorry its taken so long but this is a fairly big contribution from you as a community member (which we love). Hoping you are not discouraged that it took so long.

zerkms:
@deekoder not at all - it's summer here and I spend my evenings cycling, not at the PC ;-)

deekoder:
@zerkms - So jealous. We are freezing here. But when its winter there again, i hope you will come back to us with a PR. We really want to help you make this PR happen!

harshavardhana:
@balamurugana since notification refactor is merged - is it possible to do this now?

harshavardhana:
@zerkms do you have time to redo this PR with latest master? If not should we close this PR and based on this idea one of us can rework this later.

zerkms:
Yep, at this moment my focus has shifted somewhere else, so unfortunately I cannot invest enough time to redo it again, sorry :-(

PR #5045, opened 2017-10-10
fs: All parts except the last part should be of the same size

### LINKED ISSUE #4881: FS: return error if part size differ
For multipart, currently we do background append. This is inefficient, better approach is to use `pwrite()` and write the part to the right offset. When the client sends complete multipart upload call we should have the full file ready to be renamed to the main namespace.

There are limitations with this approach, i.e clients can't send variable part sizes. Hence the change can break clients if they send variable part sizes. Before we commit to this way of doing things we can test waters by sending error if client send variable sized parts and see if users report any error (just like how we send empty response for listparts and listincompleteuploads now)

If we don't hear of any critical client breakages we can go ahead and implement pwrite() style multipart operation.

krishnasrinivas:
fixes #4881

## Description

https://github.com/minio/minio/issues/4881 has detailed explanation on this patch.

## How Has This Been Tested?

go test

## Types of changes

## Checklist:

### DIFF
--- a/cmd/fs-v1-multipart.go
+++ b/cmd/fs-v1-multipart.go
@@ -777,6 +777,7 @@ func (fs fsObjects) CompleteMultipartUpload(bucket string, object string, upload
 		return oi, toObjectErr(err, minioMetaMultipartBucket, fsMetaPathMultipart)
 	}
 
+	partSize := int64(-1) // Used later to ensure that all parts sizes are same.
 	// Validate all parts and then commit to disk.
 	for i, part := range parts {
 		partIdx := fsMeta.ObjectPartIndex(part.PartNumber)
@@ -798,6 +799,20 @@ func (fs fsObjects) CompleteMultipartUpload(bucket string, object string, upload
 				PartSize:   fsMeta.Parts[partIdx].Size,
 				PartETag:   part.ETag,
 			})
+		}
+		if partSize == -1 {
+			partSize = fsMeta.Parts[partIdx].Size
+		}
+		// TODO: Make necessary changes in future as explained in the below comment.
+		// All parts except the last part has to be of same size. We are introducing this
+		// check to see if any clients break. If clients do not break then we can optimize
+		// multipart PutObjectPart by writing the part at the right offset using pwrite()
+		// so that we don't need to do background append at all. i.e by the time we get
+		// CompleteMultipartUpload we already have the full file available which can be
+		// renamed to the main name-space.
+		if (i < len(parts)-1) && partSize != fsMeta.Parts[partIdx].Size {
+			fs.rwPool.Close(fsMetaPathMultipart)
+			return oi, traceError(PartsSizeUnequal{})
 		}
 	}
 
--- a/cmd/api-errors.go
+++ b/cmd/api-errors.go
@@ -147,6 +147,7 @@ const (
 	ErrInvalidResourceName
 	ErrServerNotInitialized
 	ErrOperationTimedOut
+	ErrPartsSizeUnequal
 	// Add new extended error codes here.
 	// Please open a https://github.com/minio/minio/issues before adding
 	// new error codes here.
@@ -660,6 +661,11 @@ var errorCodeResponse = map[APIErrorCode]APIError{
 		Description:    "Your metadata headers are not supported.",
 		HTTPStatusCode: http.StatusBadRequest,
 	},
+	ErrPartsSizeUnequal: {
+		Code:           "XMinioPartsSizeUnequal",
+		Description:    "All parts except the last part should be of the same size.",
+		HTTPStatusCode: http.StatusBadRequest,
+	},
 	// Add your error structure here.
 }
 
@@ -758,6 +764,8 @@ func toAPIErrorCode(err error) (apiErr APIErrorCode) {
 		apiErr = ErrEntityTooLarge
 	case UnsupportedMetadata:
 		apiErr = ErrUnsupportedMetadata
+	case PartsSizeUnequal:
+		apiErr = ErrPartsSizeUnequal
 	default:
 		apiErr = ErrInternalError
 	}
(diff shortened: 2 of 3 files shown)


cmd/fs-v1-multipart.go
```diff
@@ -777,6 +777,7 @@ func (fs fsObjects) CompleteMultipartUpload(bucket string, object string, upload
 		return oi, toObjectErr(err, minioMetaMultipartBucket, fsMetaPathMultipart)
 	}
 
+	partSize := int64(-1)
```
kannappanr: Why is part size set to Int64, why not Uint64?

cmd/fs-v1-multipart.go
```diff
@@ -799,6 +800,19 @@ func (fs fsObjects) CompleteMultipartUpload(bucket string, object string, upload
 				PartETag:   part.ETag,
 			})
 		}
+		if partSize == -1 {
```
kannappanr, line 803: A comment like, you set the partsize value to first available size and you would be using that variable to make sure that all the remaining parts are of same size.

cmd/fs-v1-multipart.go
```diff
@@ -799,6 +800,19 @@ func (fs fsObjects) CompleteMultipartUpload(bucket string, object string, upload
...
+			partSize = fsMeta.Parts[partIdx].Size
+		}
+		// All parts except the last part has to be of same size. We are introducing this
+		// check to see if any clients break. If clients do not break then we can optimize
+		// multipart PutObjectPart by writing the part at the right offset using pwrite()
+		// so that we don't need to do background append at all. i.e by the time we get
+		// CompleteMultipartUpload we already have the full file available which can be
+		// renamed to the main name-space.
+		if (i < len(parts)-1) && partSize != fsMeta.Parts[partIdx].Size {
```
kannappanr, line 813: Should this check be done on upload multipart, instead of waiting till complete multipart?

krishnasrinivas: I tried to do this, but it affected test cases, fixing which was tricky and needs more effort.

Hence we can do this check in PutObjectPart once we know that no clients break and we incorporate the pwrite() style PutObjectPart. During which we can invest time in fixing the test cases as well. We have to fix test cases carefully as we allow parts with different sizes for erasure backend.

cmd/fs-v1-multipart.go
krishnasrinivas: I have added a comment

kannappanr (approved):
LGTM

krisis:
The following steps result in `2017/10/11 11:51:42 We encountered an internal error, please try again.
exit status 1` in client side.

On the server side I see the following error,
`ERRO[0701] Unable to complete multipart upload.          cause=All parts except the last part should be of the same size source=[object-handlers.go:1007:objectAPIHandlers.CompleteMultipartUploadHandler()] stack=fs-v1-multipart.go:814:fsObjects.CompleteMultipartUpload <autogenerated>:1:(*fsObjects).CompleteMultipartUpload object-handlers.go:1005:objectAPIHandlers.CompleteMultipartUploadHandler api-router.go:50:(objectAPIHandlers).CompleteMultipartUploadHandler-fm`
1. `mc mb ALIAS/test`
2. `go run` the following sample program

```go
package main

import (
	"bytes"
	"io"
	"log"

	"github.com/minio/minio-go"
)

type state struct {
	reader io.Reader
	size   int
}

func main() {
	endpoint := "localhost:9000"
	accessKeyID := "minio"
	secretAccessKey := "minio123"

	mc, err := minio.NewCore(endpoint, accessKeyID, secretAccessKey, false)
	if err != nil {
		log.Fatalln(err)
	}

	uid, err := mc.NewMultipartUpload("test", "bigobj", minio.PutObjectOptions{})
	if err != nil {
		log.Fatalln(err)
	}

	mb := 1024 * 1024
	sixMB := 6 * mb
	fiveMB := 5 * mb
	partStates := []state{
		state{reader: bytes.NewReader(bytes.Repeat([]byte("a"), sixMB)), size: sixMB},
		state{reader: bytes.NewReader(bytes.Repeat([]byte("a"), fiveMB)), size: fiveMB},
		state{reader: bytes.NewReader(bytes.Repeat([]byte("a"), fiveMB)), size: fiveMB},
	}

	var parts []minio.CompletePart
	for i, partState := range partStates {
		part, err := mc.PutObjectPart("test", "bigobj", uid, i+1, partState.reader, int64(partState.size), nil, nil)
		if err != nil {
			log.Fatalln(err)
		}
		parts = append(parts, minio.CompletePart{PartNumber: part.PartNumber, ETag: part.ETag})
	}

	err = mc.CompleteMultipartUpload("test", "bigobj", uid, parts)
	if err != nil {
		log.Fatalln(err)
	}
}

```

krishnasrinivas:
@krisis yes, that is the expected behaviour. Or are you indicating that we should return different error than `Internal Error`?

harshavardhana:
> @krisis yes, that is the expected behaviour. Or are you indicating that we should return different error than Internal Error?

Yes @krishnasrinivas if we are introducing a divergent functionality we can introduce a new XMinio error in this scenario.

krisis:
> @krisis yes, that is the expected behaviour. Or are you indicating that we should return different error than Internal Error?

yes, especially HTTP status code 5xx indicates that the problem is temporary and may go away on retry.

krishnasrinivas:
with @krisis test code:
```
rishna@escape:~/dev$ go run put-multipart.go
2017/10/11 14:43:35 All parts except the last part should be of the same size.
exit status 1
krishna@escape:~/dev$
```

krisis (approved):

deekoder:
waiting on @krishnasrinivas  on codecov failing

krishnasrinivas:
@deekoder As discussed in the meeting this is a temporary change and hence did not add test case. Instead I have added TODO so that we can make a decision later if this change is the way forward and we can add tests to cover it.

Issue #4761, opened 2017-08-03
Failed to lookup minio1 on Docker Swarm


rolandrajko:
Hi,

I would like to deploy a Minio stack on Docker Swarm, but I have got only the following message every time: `Failed to resolve hosts... Failed to lookup minio1`

## Expected Behavior
I followed your instructions in [docs/Deploy Minio on Docker Swarm](https://docs.minio.io/docs/deploy-minio-on-docker-swarm), and read the article [The complete guide to attach a Docker volume with Minio on your Docker Swarm Cluster](http://firepress.org/blog/the-complete-guide-to-attach-a-docker-volume-with-minio-on-your-docker-swarm-cluster/), so I expected it will start up quickly and successfully.

## Current Behavior
In the logs I get continously the following message: 
```
minio1.1.xik0ml96hy6l@manager1    | .Failed to lookup minio1
minio1.1.xik0ml96hy6l@manager1    | Failed to resolve hosts.. retrying after 1 second.
```
If I try to ping itself (or another element of the Minio stack) with the command `docker exec -it [container_id] ping minio1`, I get the response `ping: bad address 'minio1'`, but if I ping another service on the same overlay network, it's okay.

## Steps to Reproduce (for bugs)
For now I kept only a few lines below in my `docker-minio.yml` file:
```
version: "3.2"

services:
  minio1:
    image: minio/minio:RELEASE.2017-07-24T18-27-35Z
    volumes:
      - /mnt/minio/minio1:/export
    ports:
      - "9001:9000"
    networks:
      - ntw_minio
    deploy:
      placement:
        constraints: [node.hostname == manager1]
      restart_policy:
        delay: 10s
        max_attempts: 10
        window: 60s
    command: server http://minio1/export
    secrets:
      - access_key
      - secret_key

networks:
  ntw_minio:
    external: true

secrets:
  secret_key:
    external: true
  access_key:
    external: true
```
(I tried with named volumes as you described too.)

Steps:
1. `rm -rf /mnt/minio/minio1`
2. `mkdir -p /mnt/minio/minio1`
3. `echo "minio_access_key" | docker secret create access_key -`
4. `echo "minio_secret_key" | docker secret create secret_key -`
5. `docker stack deploy -c docker-minio.yml minio_stack`

I also tried with the following command:
 ```
docker service create \
    --name minio1 \
    --network ntw_minio \
    --replicas "1" \
    -p "9001:9000" \
    --constraint 'node.hostname==manager1' \
[...]

PR #5051, opened 2017-10-10
Fixed docker-compose file for Minio on Docker Swarm.

fungiboletus:
Hei.

[If you follow the documentation ](https://docs.minio.io/docs/deploy-minio-on-docker-swarm), you end up with the issue #4761 

Therefore, I updated the version in the docker-compose file.

### DIFF
--- a/docs/orchestration/docker-swarm/docker-compose-secrets.yaml
+++ b/docs/orchestration/docker-swarm/docker-compose-secrets.yaml
@@ -2,7 +2,7 @@ version: '3.1'
 
 services:
   minio1:
-    image: minio/minio:RELEASE.2017-08-05T00-00-53Z
+    image: minio/minio:latest
     volumes:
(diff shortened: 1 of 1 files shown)


vadmeste:
Thanks for this PR @yellowiscool

vadmeste (approved):
LGTM

nitisht:
@yellowiscool Thank you for the PR, but `latest` tag is not the recommended approach as it has its own set of [problems](http://container-solutions.com/docker-latest-confusion/)

So, we tag releases and use them accordingly in docs. We intend to keep the docs updated with each release, however there was a delay this time, I have now updated the docs.

nitisht:
@yellowiscool as the doc is now updated, feel free to close this PR.

fungiboletus:
@nitisht Ok thanks for the link about the `latest` problems.

PR #5054, opened 2017-10-10
Update Minio version in docs and example yaml files

nitisht:
## Description
Update the Minio version in docs and yaml files 

## Motivation and Context
After Minio release `RELEASE.2017-09-29T19-16-56Z` docs and yaml files need to be updated.

## Types of changes

## Checklist:

### DIFF
--- a/docs/orchestration/docker-swarm/docker-compose-secrets.yaml
+++ b/docs/orchestration/docker-swarm/docker-compose-secrets.yaml
@@ -2,7 +2,7 @@ version: '3.1'
 
 services:
   minio1:
-    image: minio/minio:RELEASE.2017-08-05T00-00-53Z
+    image: minio/minio:RELEASE.2017-09-29T19-16-56Z
     volumes:
       - minio1-data:/export
     ports:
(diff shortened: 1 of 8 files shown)


harshavardhana (approved):

vadmeste:
@nitisht, why we don't use minio/minio:latest instead of putting a specific Minio image version ?

harshavardhana:
> @nitisht, why we don't use minio/minio:latest instead of putting a specific Minio image version ?

AFAIK - It is generally considered bad practice in prod deployments to use "latest" directly.  It also disallows accidental updates to the image if the stack is being restarted etc.

vadmeste:
> AFAIK - It is generally considered bad practice in prod deployments to use "latest" directly. It also disallows accidental updates to the image if the stack is being restarted etc.

yes, that makes sense.. but we need to keep this documentation always updated, we had one user complaining about a bug which was already fixed in the last _release_.

vadmeste (approved):
LGTM

nitisht:
> why we don't use minio/minio:latest instead of putting a specific Minio image version

@vadmeste here is more detailed explanation of the problem with `latest` tag : http://container-solutions.com/docker-latest-confusion/

I agree we should keep these docs always updated, probably we can do it as a part of build automation. Let me add a point there.

PR #5067, opened 2017-10-12
Remove body from POST request in webhook

### LINKED ISSUE #5066: webhook lookup invalid json in body
I have a minio server which uses webhooks for notification.
When I'm starting the server, it sends a request to the webhook where the body is not a JSON however the content-type is specified as application/json.

## Expected Behavior
Body would respect the Content-type header. If Content-type says application/json then the body would be JSON also.

## Current Behavior
The following request is sent to my configured webhook when the minio server is starting:

```
POST /webhook HTTP/1.1
Host: 10.30.0.64:5000
User-Agent: Minio/DEVELOPMENT.2017-10-11T17-19-39Z (linux; amd64)
Content-Length: 32
Content-Type: application/json
Accept-Encoding: gzip
Connection: close

DEVELOPMENT.2017-10-11T17-19-39Z
```

Here, the body is not a valid JSON but the Content-Type says that it is a JSON, which is contradiction.

1. use minio/minio:edge, but also works with minio/minio docker image
2. configure webhook as described in the documentation 
3. start the webhook daemon (here, I used `netcat -l -p 5000` for debugging purposes)
4. start minio server

nitisht:
## Description
If webhook notification is configured, Minio server tries to lookup the
webhook endpoint by making a POST request with body set as releasetag.
We can remove the body from the POST request as the POST body does not
add any specific value.

This discussion on IETF group says empty POSTs are okay
http://lists.w3.org/Archives/Public/ietf-http-wg/2010JulSep/0272.html

## Motivation and Context
Fixes: https://github.com/minio/minio/issues/5066

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:

### DIFF
--- a/cmd/notify-webhook.go
+++ b/cmd/notify-webhook.go
@@ -91,10 +91,9 @@ func isNetErrorIgnored(err error) bool {
 }
 
 // Lookup endpoint address by successfully POSTting
-// a JSON which would send out minio release.
+// empty body.
 func lookupEndpoint(urlStr string) error {
-	minioRelease := bytes.NewReader([]byte(ReleaseTag))
-	req, err := http.NewRequest("POST", urlStr, minioRelease)
+	req, err := http.NewRequest("POST", urlStr, bytes.NewReader([]byte("")))
 	if err != nil {
 		return err
 	}
@@ -107,8 +106,8 @@ func lookupEndpoint(urlStr string) error {
 		},
 	}
 
-	// Set content-type.
-	req.Header.Set("Content-Type", "application/json")
+	// Set content-length to zero as there is no payload.
+	req.ContentLength = 0
 
 	// Set proper server user-agent.
 	req.Header.Set("User-Agent", globalServerUserAgent)


vadmeste:
I don't think webhook is working in master.

Besides, I was wondering why we are not sending clients notification data as we do with other notifications hooks.

harshavardhana:
> I don't think webhook is working in master.
> 

Webhook is working fine. 

> Besides, I was wondering why we are not sending clients notification data as we do with other notifications hooks.

Not sure what does this mean..

vadmeste:
> Webhook is working fine.

yes sorry.. one typo in mc alias, which was considered as local directory



> Not sure what does this mean..

I mean why are we sending empty data? we can send notification information instead as we do in NATs, Redis, MySQL etc.. (this is out of scope of this PR)

vadmeste (approved):
LGTM

harshavardhana:
> I mean why are we sending empty data? we can send notification information instead as we do in NATs, Redis, MySQL etc.. (this is out of scope of this PR)

No we are sending empty to validate the endpoint @vadmeste is correctly configured.

cmd/notify-webhook.go
```diff
@@ -107,8 +107,8 @@ func lookupEndpoint(urlStr string) error {
 		},
 	}
 
-	// Set content-type.
-	req.Header.Set("Content-Type", "application/json")
+	// Set content-length to zero as there is no payload.
+	req.Header.Set("Content-Length", "0")
```
harshavardhana: we don't need to set like this explicitly. just set it as `req.ContentLength = 0`

cmd/notify-webhook.go
```diff
@@ -91,10 +91,10 @@ func isNetErrorIgnored(err error) bool {
 }
 
 // Lookup endpoint address by successfully POSTting
-// a JSON which would send out minio release.
+// empty body.
 func lookupEndpoint(urlStr string) error {
-	minioRelease := bytes.NewReader([]byte(ReleaseTag))
-	req, err := http.NewRequest("POST", urlStr, minioRelease)
+	emptyBody := bytes.NewReader([]byte(""))
```
harshavardhana: No need to have a new variable for this just do this 

```
http.NewRequest("POST", urlStr, bytes.NewReader([]byte("")))
```

harshavardhana (changes requested):

cmd/notify-webhook.go
nitisht: done

cmd/notify-webhook.go
nitisht: done

harshavardhana (approved):

Issue #4994, opened 2017-09-29
[gateway/azure] validate whether gateway sends Content-MD5 or checksum to server


balamurugana:
Currently Content-MD5 or other checksum is not sent to Azure server for `PUT`, `POST` methods.  Its required to send at least one checksum header to server to verify transferred body.

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

* Version used (`minio version`):
* Environment name and version (e.g. nginx 1.9.1):
* Server type and version:
* Operating System and version (`uname -a`):
* Link to your project:

deekoder:
We need to do this in the next release.

jmaitrehenry:
Actually, I found 1 bug about this issue:
```
func s3MetaToAzureProperties(s3Metadata map[string]string) (storage.BlobMetadata,
	storage.BlobProperties, error) {
[...]
	for k, v := range s3Metadata {
		k = http.CanonicalHeaderKey(k)
		switch {
-		case k == "Content-MD5":
+		case k == "Content-Md5":
			props.ContentMD5 = v
		}
	}
	return blobMeta, props, nil
}
```

And it should already works if the client sent the `Content-MD5` header.
I didn't think Minio should calculate the md5 from `r.Body` and send it to Azure because we don't have an easy way to to it without reading the Reader as Azure need the `io.Reader`.

I will send a PR for the bug

harshavardhana:
:+1:  @jmaitrehenry nice find.

harshavardhana:
There is a reason we still need to calculate Md5, when `Content-Md5` is not present in the incoming request. But your PR will address one case though.

harshavardhana:
One interesting point from Azure doc is here to notice https://blogs.msdn.microsoft.com/windowsazurestorage/2011/02/17/windows-azure-blob-md5-overview/ 
> MD5 checking is optional on both PUT and GET operations; however it does provide a convenience facility to ensure data integrity across the network when using HTTP. Additionally since HTTPS provides transport layer security additional MD5 checking is not needed while connecting over HTTPS as it would be redundant.

So looks like for Azure the recommended approach is if you are using HTTPS then don't set MD5, i am sure that we are using MD5 so if the Content-Md5 really doesn't exist we don't need to set it ourselves as indicated/recommended by the Azure docs and it is repeatedly mentioned in their docs.

jmaitrehenry:
[...]

PR #5068, opened 2017-10-13
Fix s3MetaToAzureProperties Content-Md5 key

jmaitrehenry:
## Description
The key was not correctly checked because we `CanonicalHeaderKey`ise it just before

## Motivation and Context
Fix a part of #4994 

## How Has This Been Tested?
Testing by putting some new object on Azure

## Types of changes

## Checklist:

### DIFF
--- a/cmd/gateway-azure.go
+++ b/cmd/gateway-azure.go
@@ -100,7 +100,7 @@ func s3MetaToAzureProperties(s3Metadata map[string]string) (storage.BlobMetadata
 		case k == "Content-Length":
 			// assume this doesn't fail
 			props.ContentLength, _ = strconv.ParseInt(v, 10, 64)
-		case k == "Content-MD5":
+		case k == "Content-Md5":
 			props.ContentMD5 = v
 		case k == "Content-Type":
 			props.ContentType = v
--- a/cmd/gateway-azure_test.go
+++ b/cmd/gateway-azure_test.go
@@ -82,6 +82,14 @@ func TestS3MetaToAzureProperties(t *testing.T) {
 			t.Fatalf("Test failed with unexpected error %s, expected UnsupportedMetadata", err)
 		}
 	}
+
+	headers = map[string]string{
+		"Content-MD5": "d41d8cd98f00b204e9800998ecf8427e",
+	}
+	_, props, _ := s3MetaToAzureProperties(headers)
+	if props.ContentMD5 != headers["Content-MD5"] {
+		t.Fatalf("Test failed, expected %s, got %s", headers["Content-MD5"], props.ContentMD5)
+	}
 }
 
 func TestAzurePropertiesToS3Meta(t *testing.T) {
--- a/cmd/gateway-azure_test.go
+++ b/cmd/gateway-azure_test.go
@@ -86,7 +86,10 @@ func TestS3MetaToAzureProperties(t *testing.T) {
 	headers = map[string]string{
 		"content-md5": "Dce7bmCX61zvxzP5QmfelQ==",
 	}
-	_, props, _ := s3MetaToAzureProperties(headers)
+	_, props, err := s3MetaToAzureProperties(headers)
+	if err != nil {
+		t.Fatalf("Test failed, with %s", err)
+	}
 	if props.ContentMD5 != headers["content-md5"] {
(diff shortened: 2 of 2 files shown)


cmd/gateway-azure.go
```diff
@@ -100,7 +100,7 @@ func s3MetaToAzureProperties(s3Metadata map[string]string) (storage.BlobMetadata
 		case k == "Content-Length":
 			// assume this doesn't fail
 			props.ContentLength, _ = strconv.ParseInt(v, 10, 64)
-		case k == "Content-MD5":
+		case k == "Content-Md5":
```
harshavardhana, line 103: Can you add a unit test for this?

jmaitrehenry:
I will try :)
Minio is my first project in Go and I want to learn more. It could take a little time before having a test working. (Or maybe not, it will be my first test in go)

jmaitrehenry:
@harshavardhana It's really easy to add a test finally

cmd/gateway-azure_test.go
```diff
@@ -82,6 +82,14 @@ func TestS3MetaToAzureProperties(t *testing.T) {
 			t.Fatalf("Test failed with unexpected error %s, expected UnsupportedMetadata", err)
 		}
 	}
+
+	headers = map[string]string{
+		"Content-MD5": "d41d8cd98f00b204e9800998ecf8427e",
+	}
+	_, props, _ := s3MetaToAzureProperties(headers)
+	if props.ContentMD5 != headers["Content-MD5"] {
```
harshavardhana: This should have failed right?

harshavardhana (changes requested):

jmaitrehenry:
Without the fix, it failed.
The http header is Content-MD5 and should works.

harshavardhana:
> Without the fix, it failed.
> The http header is Content-MD5 and should works.

Yes but the fix is looking for `Content-Md5` not `Content-MD5`  looks like its being converted using http.CanonicalHeaderKey - perhaps test should have `Content-Md5` not `Content-MD5` to avoid confusion. Also `Content-Md5` should be base64 not `hex` of md5sum.

jmaitrehenry:
The fix look for the `CanonicalHeaderKey` version of the header.
The tests check for the right header sent by client `Content-MD5 ` wich is transformed to `Content-Md5` by minio, and set to `props.ContentMD5` for Azure blob.

> Also Content-Md5 should be base64 not hex of md5sum
Right, I could change it for a real md5 for Azure, but, most client will send a md5 in "8209ed86ac433cee4e5218e0d93b7d6c" format.

I could add something that recognize this kind of md5 and trasnform it for azure.

harshavardhana:
Content-MD5 is never sent by the client if it's following HTTP RFC. Go http server also marshals the errant key names like this into proper canonical names in req.Header.

So our transformation test rather looks unexpected that is why we can keep it simply `content-md5` since all metadata map entries are small cases.

jmaitrehenry:
@harshavardhana fixed

harshavardhana (approved):

cmd/gateway-azure_test.go
```diff
@@ -82,6 +82,14 @@ func TestS3MetaToAzureProperties(t *testing.T) {
 			t.Fatalf("Test failed with unexpected error %s, expected UnsupportedMetadata", err)
 		}
 	}
+
+	headers = map[string]string{
+		"content-md5": "Dce7bmCX61zvxzP5QmfelQ==",
+	}
+	_, props, _ := s3MetaToAzureProperties(headers)
```
balamurugana: You would need to get and check error also

balamurugana (changes requested):
minor error in test case

balamurugana (approved):
As discussed https://github.com/minio/minio/pull/5068/files/caf8aff52e189e62543f9738b9815af2f4a709da#diff-2ee5453b2351c76bb4c5b94d087d5a96R89 can be taken into separate PR

cmd/gateway-azure_test.go
jmaitrehenry: Will do tonight

jmaitrehenry:
Sorry for the delay @balamurugana, it's now fixed.

balamurugana (approved):

deekoder:
Ignoring codecoverage as spurious. Test has been added. Merging hence.

PR #5071, opened 2017-10-14
Simplify data verification with HashReader.

harshavardhana:
## Description
Verify() was being called by caller after the data
has been successfully read after io.EOF. This disconnection
opens a race under concurrent access to such an object.
Verification is not necessary outside of Read() call,
we can simply just do checksum verification right inside
Read() call at io.EOF.

## Motivation and Context
Simplification of usage

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:

### DIFF
--- a/cmd/gateway-s3.go
+++ b/cmd/gateway-s3.go
@@ -20,11 +20,10 @@ import (
 	"io"
 	"net/http"
 
-	"encoding/hex"
-
 	minio "github.com/minio/minio-go"
 	"github.com/minio/minio-go/pkg/policy"
 	"github.com/minio/minio-go/pkg/s3utils"
+	"github.com/minio/minio/pkg/hash"
 )
 
 // s3ToObjectError converts Minio errors to minio object layer errors.
@@ -84,7 +83,7 @@ func s3ToObjectError(err error, params ...string) error {
 			Object: object,
 		}
 	case "XAmzContentSHA256Mismatch":
-		err = SHA256Mismatch{}
+		err = hash.SHA256Mismatch{}
 	case "NoSuchUpload":
 		err = InvalidUploadID{}
 	case "EntityTooSmall":
@@ -343,17 +342,8 @@ func (l *s3Objects) GetObjectInfo(bucket string, object string) (objInfo ObjectI
 }
 
 // PutObject creates a new object with the incoming data,
-func (l *s3Objects) PutObject(bucket string, object string, data *HashReader, metadata map[string]string) (objInfo ObjectInfo, err error) {
-	sha256sumBytes, err := hex.DecodeString(data.sha256Sum)
-	if err != nil {
-		return objInfo, s3ToObjectError(traceError(err), bucket, object)
-	}
-	md5sumBytes, err := hex.DecodeString(metadata["etag"])
-	if err != nil {
-		return objInfo, s3ToObjectError(traceError(err), bucket, object)
-	}
-	delete(metadata, "etag")
-	oi, err := l.Client.PutObject(bucket, object, data, data.Size(), md5sumBytes, sha256sumBytes, toMinioClientMetadata(metadata))
+func (l *s3Objects) PutObject(bucket string, object string, data *hash.Reader, metadata map[string]string) (objInfo ObjectInfo, err error) {
+	oi, err := l.Client.PutObject(bucket, object, data, data.Size(), data.MD5(), data.SHA256(), toMinioClientMetadata(metadata))
 	if err != nil {
 		return objInfo, s3ToObjectError(traceError(err), bucket, object)
 	}
@@ -472,18 +462,8 @@ func fromMinioClientObjectPart(op minio.ObjectPart) PartInfo {
 }
 
 // PutObjectPart puts a part of object in bucket
-func (l *s3Objects) PutObjectPart(bucket string, object string, uploadID string, partID int, data *HashReader) (pi PartInfo, e error) {
-	md5HexBytes, err := hex.DecodeString(data.md5Sum)
-	if err != nil {
-		return pi, s3ToObjectError(traceError(err), bucket, object)
-	}
-
-	sha256sumBytes, err := hex.DecodeString(data.sha256Sum)
-	if err != nil {
-		return pi, err
-	}
-
-	info, err := l.Client.PutObjectPart(bucket, object, uploadID, partID, data, data.Size(), md5HexBytes, sha256sumBytes)
+func (l *s3Objects) PutObjectPart(bucket string, object string, uploadID string, partID int, data *hash.Reader) (pi PartInfo, e error) {
+	info, err := l.Client.PutObjectPart(bucket, object, uploadID, partID, data, data.Size(), data.MD5(), data.SHA256())
 	if err != nil {
 		return pi, s3ToObjectError(traceError(err), bucket, object)
 	}
--- a/cmd/object-api-interface.go
+++ b/cmd/object-api-interface.go
@@ -17,13 +17,9 @@
 package cmd
 
 import (
-	"bytes"
-	"crypto/md5"
-	"encoding/hex"
-	"hash"
 	"io"
 
-	sha256 "github.com/minio/sha256-simd"
+	"github.com/minio/minio/pkg/hash"
 )
 
 // ObjectLayer implements primitives for object API layer.
@@ -42,7 +38,7 @@ type ObjectLayer interface {
 	// Object operations.
 	GetObject(bucket, object string, startOffset int64, length int64, writer io.Writer) (err error)
 	GetObjectInfo(bucket, object string) (objInfo ObjectInfo, err error)
-	PutObject(bucket, object string, data *HashReader, metadata map[string]string) (objInfo ObjectInfo, err error)
+	PutObject(bucket, object string, data *hash.Reader, metadata map[string]string) (objInfo ObjectInfo, err error)
 	CopyObject(srcBucket, srcObject, destBucket, destObject string, metadata map[string]string) (objInfo ObjectInfo, err error)
 	DeleteObject(bucket, object string) error
 
@@ -50,7 +46,7 @@ type ObjectLayer interface {
 	ListMultipartUploads(bucket, prefix, keyMarker, uploadIDMarker, delimiter string, maxUploads int) (result ListMultipartsInfo, err error)
 	NewMultipartUpload(bucket, object string, metadata map[string]string) (uploadID string, err error)
 	CopyObjectPart(srcBucket, srcObject, destBucket, destObject string, uploadID string, partID int, startOffset int64, length int64) (info PartInfo, err error)
-	PutObjectPart(bucket, object, uploadID string, partID int, data *HashReader) (info PartInfo, err error)
+	PutObjectPart(bucket, object, uploadID string, partID int, data *hash.Reader) (info PartInfo, err error)
 	ListObjectParts(bucket, object, uploadID string, partNumberMarker int, maxParts int) (result ListPartsInfo, err error)
 	AbortMultipartUpload(bucket, object, uploadID string) error
 	CompleteMultipartUpload(bucket, object, uploadID string, uploadedParts []completePart) (objInfo ObjectInfo, err error)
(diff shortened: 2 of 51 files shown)


cmd/xl-v1-object.go
```diff
@@ -531,8 +530,8 @@ func (xl xlObjects) PutObject(bucket string, object string, data *HashReader, me
 		// Hint the filesystem to pre-allocate one continuous large block.
 		// This is only an optimization.
 		if curPartSize > 0 {
-			pErr := xl.prepareFile(minioMetaTmpBucket, tempErasureObj, curPartSize, storage.disks, xlMeta.Erasure.BlockSize, xlMeta.Erasure.DataBlocks)
-			if pErr != nil {
+			if pErr := xl.prepareFile(minioMetaTmpBucket, tempErasureObj, curPartSize, storage.disks,
```
balamurugana: why do need to change this line?

cmd/xl-v1-object.go
```diff
@@ -113,7 +112,7 @@ func (xl xlObjects) CopyObject(srcBucket, srcObject, dstBucket, dstObject string
 		pipeWriter.Close() // Close writer explicitly signalling we wrote all data.
 	}()
 
-	objInfo, err := xl.PutObject(dstBucket, dstObject, NewHashReader(pipeReader, length, metadata["etag"], ""), metadata)
+	objInfo, err := xl.PutObject(dstBucket, dstObject, NewHashReader(pipeReader, length, "", ""), metadata)
```
balamurugana: why etag is not passed?

cmd/test-utils_test.go
```diff
@@ -1975,6 +1975,7 @@ func ExecObjectLayerTest(t TestErrHandler, objTest objTestType) {
 	if err != nil {
 		t.Fatalf("Initialization of object layer failed for single node setup: %s", err)
 	}
+
```
balamurugana, line 1987: My recommendation is to avoid unrelated diffs

cmd/object-handlers.go
```diff
@@ -287,6 +287,9 @@ func (api objectAPIHandlers) HeadObjectHandler(w http.ResponseWriter, r *http.Re
 // Extract metadata relevant for an CopyObject operation based on conditional
 // header values specified in X-Amz-Metadata-Directive.
 func getCpObjMetadataFromHeader(header http.Header, defaultMeta map[string]string) (map[string]string, error) {
+	// Make sure to remove saved etag if any, CopyObject calculates a new one.
+	delete(defaultMeta, "etag")
```
balamurugana, line 292: passed `defaultMeta` is a pointer.  Do we want to modify the source?

cmd/object-handlers.go
```diff
@@ -576,7 +573,7 @@ func (api objectAPIHandlers) PutObjectHandler(w http.ResponseWriter, r *http.Req
 			sha256sum = r.Header.Get("X-Amz-Content-Sha256")
 		}
 		// Create object.
-		objInfo, err = objectAPI.PutObject(bucket, object, NewHashReader(r.Body, size, metadata["etag"], sha256sum), metadata)
+		objInfo, err = objectAPI.PutObject(bucket, object, NewHashReader(r.Body, size, md5sum, sha256sum), metadata)
```
balamurugana: These changes are just diff increaser.  You can optimize things is different PR IMO

cmd/object-api-interface.go
```diff
@@ -114,29 +130,25 @@ func (r *HashReader) Read(p []byte) (n int, err error) {
...
-func (r *HashReader) MD5() []byte { return r.md5Hash.Sum(nil) }
-
-// Verify verifies if the computed MD5 sum - and SHA256 sum - are
+// verify verifies if the computed MD5 sum - and SHA256 sum - are
 // equal to the ones specified when creating the HashReader.
-func (r *HashReader) Verify() error {
+func (r *HashReader) verify() error {
 	if r.sha256Hash != nil {
-		sha256Sum, err := hex.DecodeString(r.sha256Sum)
+		sha256Sum, err := hex.DecodeString(r.SHA256Sum)
```
balamurugana: This check should be in `NewHashReader()` function.

cmd/object-api-interface.go
```diff
@@ -114,29 +130,25 @@ func (r *HashReader) Read(p []byte) (n int, err error) {
...
-		if !bytes.Equal(sha256Sum, r.sha256Hash.Sum(nil)) {
-			return errContentSHA256Mismatch
+		if sum := r.sha256Hash.Sum(nil); !bytes.Equal(sha256Sum, sum) {
+			return SHA256Mismatch{r.SHA256Sum, hex.EncodeToString(sum)}
 		}
 	}
-	if r.md5Hash != nil && r.md5Sum != "" {
-		md5Sum, err := hex.DecodeString(r.md5Sum)
+	if r.md5Hash != nil {
+		md5Sum, err := hex.DecodeString(r.MD5Sum)
```
balamurugana: same as above

cmd/object-api-interface.go
```diff
@@ -67,30 +67,35 @@ type ObjectLayer interface {
 // HashReader writes what it reads from an io.Reader to an MD5 and SHA256 hash.Hash.
 // HashReader verifies that the content of the io.Reader matches the expected checksums.
 type HashReader struct {
```
balamurugana: Move out `HashReader` from `object-api-interface.go` for better code wake through, may be a package if file list is limited by go

cmd/object-api-interface.go
```diff
@@ -100,12 +105,23 @@ func (r *HashReader) Read(p []byte) (n int, err error) {
 	if err != nil && err != io.EOF {
```
balamurugana: you would need to handle `io.UnexpectedEOF` something like
```go
if err == nil || err == io.UnexpectedEOF {
 // update md5 and sha256 hashers
}

if err == io.EOF || err == io.UnexpectedEOF {
//  call HashReader.Verify()
}

return
```

cmd/gateway-s3.go
```diff
@@ -344,15 +344,14 @@ func (l *s3Objects) GetObjectInfo(bucket string, object string) (objInfo ObjectI
 
 // PutObject creates a new object with the incoming data,
 func (l *s3Objects) PutObject(bucket string, object string, data *HashReader, metadata map[string]string) (objInfo ObjectInfo, err error) {
-	sha256sumBytes, err := hex.DecodeString(data.sha256Sum)
+	sha256sumBytes, err := hex.DecodeString(data.SHA256Sum)
```
balamurugana: I believe SHA256 and MD5 hashes are required to be string with fixed encoding.  You could have respective methods in `HashReader` than public fields.

balamurugana (changes requested):

cmd/xl-v1-object.go
harshavardhana: We can revert this.

cmd/xl-v1-object.go
harshavardhana: Its not needed and incorrect multipart Object copied to new location would validate a wrong ETag and will fail.

cmd/test-utils_test.go
harshavardhana: Removing..

cmd/object-handlers.go
harshavardhana: This is not unrelated @balamurugana - we shouldn't use "etag" in this manner and also this change comes with the change we need to make under HashReader

cmd/object-handlers.go
harshavardhana: Yes we need to remove "etag" we can make it safer.

cmd/object-api-interface.go
harshavardhana: Read() never returns io.UnexpectedEOF @balamurugana - UnexpectedEOF is only a higher level construct.

cmd/gateway-s3.go
harshavardhana: This change be done..

cmd/object-api-interface.go
harshavardhana: This is unrelated to this PR, but i can do that..

cmd/object-api-interface.go
harshavardhana: Yeah NewHashReader is not implemented in this manner.. @balamurugana that would require a larger change.

cmd/object-api-interface.go
harshavardhana: Also this change would make the PR much bigger.

cmd/object-api-interface.go
balamurugana: I guess its related fix only because `HashReader` does verification now which changes the usability.

cmd/object-handlers.go
balamurugana: what do you mean by safer?  what is the problem with passed ETag?

cmd/object-handlers.go
balamurugana: No.  In the above `md5sum = metadata["etag"]`. you are just optimizing map access.

cmd/object-handlers.go
harshavardhana: you shouldn't pass the "etag" in CopyObject @balamurugana

cmd/object-handlers.go
harshavardhana: no there is no such thing as "etag" saved in `metadata` it is only saved internally at ObjectLayer. There is no reason to add it prematurely.  You have to see `md5sum` is never set to a map. Also taking "etag" as the md5sum value is incorrect from readability point of view.

pkg/hashreader/hashreader.go
```diff
@@ -0,0 +1,120 @@
...
+	md5sum, sha256sum []byte // Byte values of md5sum, sha256sum of client sent values.
+
+	src                 io.Reader
+	size                int64
+	md5Hash, sha256Hash hash.Hash
+}
+
+// New returns a new HashReader computing the MD5 sum and SHA256 sum
+// (if set) of the provided io.Reader.
+func New(src io.Reader, size int64, md5Hex, sha256Hex string) (*HashReader, error) {
```
balamurugana: I think size can be omitted i.e. consumer needs to restrict by size using `io.LimitReader`

pkg/hashreader/hashreader.go
```diff
@@ -0,0 +1,120 @@
+package hashreader
```
balamurugana: I think this package can be named `hasher` and having a public struct `Reader`

pkg/hashreader/hashreader.go
```diff
@@ -0,0 +1,120 @@
...
+	"io"
+)
+
+// HashReader verifies that the content of the io.Reader matches the expected checksums.
+type HashReader struct {
+	md5sum, sha256sum []byte // Byte values of md5sum, sha256sum of client sent values.
+
+	src                 io.Reader
+	size                int64
+	md5Hash, sha256Hash hash.Hash
```
balamurugana: I think checksum and hasher need to be grouped i.e. md5sum and md5hash grouped together.  Also I personally like `md5Hasher` over `md5Hash`

pkg/hashreader/hashreader.go
```diff
@@ -0,0 +1,120 @@
...
+	sha256sum, err := hex.DecodeString(sha256Hex)
+	if err != nil {
+		return nil, err
+	}
+	md5sum, err := hex.DecodeString(md5Hex)
+	if err != nil {
+		return nil, err
+	}
+	if hr, ok := src.(*HashReader); ok {
+		return hr, nil
```
balamurugana: As we found given md5/sha256 are valid and if `src` is an `HashReader` I guess we need to error out about nesting is not allowed.  This way we can avoid confusion i.e. user might think newly provided checksums are used in the returned `HashReader`

pkg/hashreader/hashreader.go
```diff
@@ -0,0 +1,120 @@
...
+		sha256sum:  sha256sum,
+		src:        io.LimitReader(src, size),
+		size:       size,
+		md5Hash:    md5Hash,
+		sha256Hash: sha256Hash,
+	}, nil
+}
+
+func (r *HashReader) Read(p []byte) (n int, err error) {
+	n, err = r.src.Read(p)
```
balamurugana: `n > 0` may be true in `err != nil`.  I think we need to fix this case as well.

pkg/hashreader/hashreader.go
```diff
@@ -0,0 +1,120 @@
...
+		r.sha256Hash.Write(p[:n])
+	}
+
+	// At io.EOF verify if the checksums are right.
+	if err == io.EOF {
+		cerr := r.Verify()
+		if cerr != nil {
+			return 0, cerr
+		}
+		return n, err
```
balamurugana: this return statement is not required

pkg/hashreader/hashreader.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+// SHA256Hex returns hex sha256 value.
+func (r *HashReader) SHA256Hex() string {
+	return hex.EncodeToString(r.sha256sum)
+}
+
+// Verify verifies if the computed MD5 sum and SHA256 sum are
+// equal to the ones specified when creating the HashReader.
+func (r *HashReader) Verify() error {
```
balamurugana: Do we need this method as public?  There is no meaning someone calls in the middle

pkg/hashreader/hashreader.go
```diff
@@ -0,0 +1,120 @@
...
+	return r.md5sum
+}
+
+// SHA256 - returns byte sha256 value
+func (r *HashReader) SHA256() []byte {
+	return r.sha256sum
+}
+
+// MD5Hex returns hex md5 value.
+func (r *HashReader) MD5Hex() string {
```
balamurugana: I think `MD5HexString()` conveys more meaning

pkg/hashreader/hashreader.go
harshavardhana: Will keep it as hashtee instead hasher means closer to hash.Hash which is not the case here.

pkg/hashreader/hashreader.go
harshavardhana: The case doesn't need to be handled if Read returns any other error other than io.EOF or nil then its a catastrophic.

pkg/hashreader/hashreader.go
harshavardhana: This is still needed there is a bug in erasure code which is not fixed yet.

pkg/hashreader/hashreader.go
harshavardhana: hasher is not go style here we just keep it similar to hash.Hash

pkg/hashreader/hashreader.go
harshavardhana: Yes good idea.

pkg/hashreader/hashreader.go
harshavardhana: Size is needed since we do not pass Size param anymore into our object interface.

pkg/hashreader/hashreader.go
harshavardhana: done

pkg/hashreader/hashreader.go
harshavardhana: > An instance of this general case is that a Reader returning a non-zero number of bytes at the end of the input stream may return either err == EOF or err == nil. The next Read should return 0, EOF. 

This line explains this.

balamurugana: Here is the text from https://golang.org/pkg/io/#Reader
```
When Read encounters an error or end-of-file condition after successfully reading n > 0 bytes, it returns the number of bytes read. It may return the (non-nil) error from the same call or return the error (and n == 0) from a subsequent call. An instance of this general case is that a Reader returning a non-zero number of bytes at the end of the input stream may return either err == EOF or err == nil. The next Read should return 0, EOF. 
```

harshavardhana: Correct but we don't need to do anything @balamurugana we already handle this.. it doesn't matter if n > 0 and err is anything other than io.EOF or nil it is a catastrophic error.  We only need to handle things for correct cases which are only two i.e either io.EOF or nil. 

So Reader implementations if n > 0 always `return either err == EOF or err == nil`  any other error is considered catastrophic and should be quickly failed.

harshavardhana:
@balamurugana @donatello please take a look at the updated PR,  package is `pkg/hashtee` with tests. 

@balamurugana i didn't make the change of moving the size param as the caller responsibility of providing the reader, as there were lot of complications in our tests and certain part of the code base. I have left the change for another time perhaps.

harshavardhana:
hashtee is now just `hash` with New renamed to NewReader.  @balamurugana @donatello please take a look.

donatello (approved):

balamurugana (approved):
LGTM

balamurugana (changes requested):
CI fails.  Please fix it.

balamurugana (approved):
LGTM

PR #5090, opened 2017-10-20
Bucket policies parser should use minio-go/pkg/policy instead.

harshavardhana:
## Description
Bucket policies parser should use minio-go/pkg/policy instead.

## Motivation and Context
Removing redundant parser logic.

## How Has This Been Tested?
Manually, using `go test`.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/bucket-policy-parser.go
+++ b/cmd/bucket-policy-parser.go
@@ -26,8 +26,11 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/minio/minio-go/pkg/policy"
 	"github.com/minio/minio-go/pkg/set"
 )
+
+var emptyBucketPolicy = policy.BucketAccessPolicy{}
 
 var conditionKeyActionMap = map[string]set.StringSet{
 	"s3:prefix":   set.CreateStringSet("s3:ListBucket"),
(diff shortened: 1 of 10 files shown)


cmd/bucket-policy-parser.go
```diff
@@ -26,9 +26,12 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/minio/minio-go/pkg/policy"
 	"github.com/minio/minio-go/pkg/set"
 )
 
+var sentinelBucketPolicy = policy.BucketAccessPolicy{}
```
donatello: Perhaps `emptyBucketPolicy` is a more straightforward name for this? Sentinels are usually used in the context of sorting as a sort of zero/border value.

donatello (dismissed):
Looks like a straightforward change. LGTM - have a comment about renaming an empty policy.

cmd/bucket-policy-parser.go
harshavardhana: I can do that @donatello - i used sentinel in context of default value.

krishnasrinivas (approved):

cmd/bucket-policy-parser.go
harshavardhana: @donatello please review again.

donatello (approved):

PR #5095, opened 2017-10-21
Virtual host style S3 requests

### LINKED ISSUE #4681: Support for DNS style S3 Bucket naming
Simple question regarding Bucket addressing convention and minio :+1: 

AWS S3 defines two way to address a bucket : http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingBucket.html

- Path style. With something like  http://server.adress/%(bucket)

- Name / DNS style. With something like  http://%(bucket).server.adress/

Is Minio supporting the DNS / Name style ? I do not find any documentation related to that.
If not, do you plan to support it in future ?

Adev

krishnasrinivas:
fixes #4681

## Description

This is to support requests like:
```
GET /passwd
Host: test.mydomain.com:9000
User-Agent: aws-sdk-nodejs/2.100.0 linux/v8.1.3 callback
Content-Length: 0
X-Amz-Date: 20171021T074937Z
Authorization: AWS4-HMAC-SHA256 Credential=SXO8VW2OFKKP2OG7AC85/20171021/us-east-1/s3/aws4_request, SignedHeaders=host;x-amz-content-sha256;x-amz-date, Signature=93e3ff11d695702736fc33ba394d09afb01efbb31f6009a0242d4fe7ff0f5314
Connection: close
```

Where bucket is specified in the host header.

This patch is developed on top of https://github.com/minio/minio/pull/4804 as I used it for debugging. Hence this PR can be taken after PR #4804 is taken in.

Will mark it as WIP till #4804 is taken in. But this can be reviewed in the mean time.

## How Has This Been Tested?

Manual

## Types of changes

## Checklist:

### DIFF
--- a/cmd/api-router.go
+++ b/cmd/api-router.go
@@ -17,6 +17,7 @@
 package cmd
 
 import router "github.com/gorilla/mux"
+import "net/http"
 
 // objectAPIHandler implements and provides http handlers for S3 API.
 type objectAPIHandlers struct {
@@ -87,9 +88,9 @@ func registerAPIRouter(mux *router.Router) {
 		// HeadBucket
 		bucket.Methods("HEAD").HandlerFunc(httpTraceAll(api.HeadBucketHandler))
 		// PostPolicy
-		bucket.Methods("POST").HeadersRegexp("Content-Type", "multipart/form-data*").HandlerFunc(httpTraceAll(api.PostPolicyBucketHandler))
+		bucket.Methods("POST").Path("/").HeadersRegexp("Content-Type", "multipart/form-data*").HandlerFunc(httpTraceAll(api.PostPolicyBucketHandler))
 		// DeleteMultipleObjects
-		bucket.Methods("POST").HandlerFunc(httpTraceAll(api.DeleteMultipleObjectsHandler))
+		bucket.Methods("POST").HandlerFunc(httpTraceAll(api.DeleteMultipleObjectsHandler)).Queries("delete", "")
 		// DeleteBucketPolicy
 		bucket.Methods("DELETE").HandlerFunc(httpTraceAll(api.DeleteBucketPolicyHandler)).Queries("policy", "")
 		// DeleteBucket
@@ -99,5 +100,8 @@ func registerAPIRouter(mux *router.Router) {
 	/// Root operation
 
 	// ListBuckets
-	apiRouter.Methods("GET").HandlerFunc(httpTraceAll(api.ListBucketsHandler))
+	apiRouter.Methods("GET").Path("/").HandlerFunc(httpTraceAll(api.ListBucketsHandler))
+
+	// If none of the routes match.
+	apiRouter.NotFoundHandler = http.HandlerFunc(httpTraceAll(notFoundHandler))
 }
--- a/cmd/api-router.go
+++ b/cmd/api-router.go
@@ -16,9 +16,7 @@
 
 package cmd
 
-import (
-	router "github.com/gorilla/mux"
-)
+import router "github.com/gorilla/mux"
 
 // objectAPIHandler implements and provides http handlers for S3 API.
 type objectAPIHandlers struct {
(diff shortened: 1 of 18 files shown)


nitisht:
Blocked/WIP till #4804 is merged, but review can be done meanwhile.

harshavardhana:
Can you fix the conflicts @krishnasrinivas ?

nitisht:
https://github.com/minio/minio/pull/4804 is merged, so unblocking this.

cmd/common-main.go
```diff
@@ -66,6 +66,8 @@ func initConfig() {
 func handleCommonCmdArgs(ctx *cli.Context) {
 	// Set configuration directory.
 	{
+		// Root domain name for virtual-host-style requests.
+		globalDomainName = ctx.String("domain")
```
harshavardhana: I was thinking we don't need a domain name command flag @krishnasrinivas, instead we can have a environment variable for this? like `MINIO_VIRTUAL_HOST_DOMAIN=` ?  and save it in the config file? 

From the code it looks like the server will honor both virtual style and path style even if virtual style is enabled.  Just like how we have `brower: "on"` we can have a `virtualDomain: "minio.io"` entry?

cmd/utils.go
```diff
@@ -229,3 +230,25 @@ func toS3ETag(etag string) string {
...
+func getResource(path string, host string, domain string) (string, error) {
+	if domain == "" {
+		return path, nil
+	}
+
+	// If virtual-host-style is enabled construct the "resource" properly.
+	if strings.Contains(host, ":") {
+		// In bucket.mydomain.com:9000, strip out :9000
+		var err error
+		if host, _, err = net.SplitHostPort(host); err != nil {
```
harshavardhana: why not use `extractHostPort()` helper? you don't need to use `strings.Contains()` function here it would further simplify this.

cmd/utils.go
```diff
@@ -229,3 +230,25 @@ func toS3ETag(etag string) string {
...
+			errorIf(err, "Unable to split %s", host)
+			return "", err
+		}
+	}
+	if !strings.HasSuffix(host, "."+domain) {
+		return path, nil
+	}
+	bucket := strings.TrimSuffix(host, "."+domain)
```
harshavardhana: I think we can validate the bucket name here as soon as we extract it using `isValidBucketName()`

cmd/signature-v2.go
```diff
@@ -157,6 +157,11 @@ func doesPresignV2SignatureMatch(r *http.Request) APIErrorCode {
 		return ErrExpiredPresignRequest
 	}
 
+	encodedResource, err = getResource(encodedResource, r.Host, globalDomainName)
+	if err != nil {
+		return ErrInternalError
```
harshavardhana: getResource can return APIErrorCode here @krishnasrinivas so that we don't need to mask the error here, we can return back proper errors..

cmd/signature-v2.go
```diff
@@ -237,6 +242,11 @@ func doesSignV2Match(r *http.Request) APIErrorCode {
 		return ErrInvalidQueryParams
 	}
 
+	encodedResource, err = getResource(encodedResource, r.Host, globalDomainName)
+	if err != nil {
```
harshavardhana, line 246: Same as above.

cmd/utils.go
```diff
@@ -229,3 +230,25 @@ func toS3ETag(etag string) string {
 
 	return etag
 }
+
+// Returns "/bucketName/objectName" for path-style or virtual-host-style requests.
+func getResource(path string, host string, domain string) (string, error) {
```
harshavardhana: We can move this function to handler-utils.go

cmd/api-router.go
```diff
@@ -35,66 +33,69 @@ func registerAPIRouter(mux *router.Router) {
...
-	bucket.Methods("GET").Path("/{object:.+}").HandlerFunc(httpTraceHdrs(api.GetObjectHandler))
-	// CopyObject
-	bucket.Methods("PUT").Path("/{object:.+}").HeadersRegexp("X-Amz-Copy-Source", ".*?(\\/|%2F).*?").HandlerFunc(httpTraceAll(api.CopyObjectHandler))
-	// PutObject
-	bucket.Methods("PUT").Path("/{object:.+}").HandlerFunc(httpTraceHdrs(api.PutObjectHandler))
-	// DeleteObject
-	bucket.Methods("DELETE").Path("/{object:.+}").HandlerFunc(httpTraceAll(api.DeleteObjectHandler))
+	var routers []*router.Router
+	if globalDomainName != "" {
+		routers = append(routers, apiRouter.Host("{bucket:[A-Za-z0-9]+}."+globalDomainName).Subrouter())
```
harshavardhana: is the regex good-enough? `buckets` can also have `-` in between? like capital is not allowed really.  So it should be a bit stricter? 

```
[a-z0-9][a-z0-9\.\-]{1,61}[a-z0-9]
```

cmd/signature-v2.go
krishnasrinivas: We don't have a suitable APIErrorCode for this and hence returned the generic internal error. Did not introduce a new APIErrorCode as this would be a highly unlikely error (but we log the error using `errorIf`). You think we should introduce a new APIErrorCode?

cmd/utils.go
krishnasrinivas: Better leave this validation to object layer. Because object layer can better control what kind of bucket names are allowed. For example Azure does not allow bucket names with `.` but aws-s3 allows. Also this is the pattern that I see for other code parts ... i.e object layer is doing the bucket name validation.

cmd/utils.go
krishnasrinivas: extractHostPort gives error if host is `bucket.mydomain.com` i.e without port or scheme

cmd/signature-v2.go
harshavardhana: Was thinking we can return something like ErrInvalidArgument ..

krishnasrinivas: Added a more generic `ErrInvalidRequest` (http.StatusBadRequest) which should be judiciously used.

cmd/api-router.go
krishnasrinivas: I have made it `{object:.+}` as IsValidBucketName validation is done later

harshavardhana: You mean `{bucket:.+}`

krishnasrinivas: yes

harshavardhana (changes requested):
Can you update the docs at `docs/config` ? with the new changes?

cmd/api-router.go
```diff
@@ -34,67 +32,69 @@ func registerAPIRouter(mux *router.Router) {
 
 	// API Router
 	apiRouter := mux.NewRoute().PathPrefix("/").Subrouter()
+	var routers []*router.Router
+	if globalDomainName != "" {
+		routers = append(routers, apiRouter.Host("{bucket:.+}."+globalDomainName).Subrouter())
```
nitisht, line 38: Gateway has router added as 

```
routers = append(routers, apiRouter.Host("{bucket:[A-Za-z0-9]+}."+globalDomainName).Subrouter())
```

While here it is `"{bucket:.+}"`, is there a reason for this difference

krishnasrinivas: good catch @nitisht gateway-router needs to be corrected

harshavardhana:
Can you rebase the PR @krishnasrinivas

harshavardhana (approved):

balamurugana (approved):
Not tested

PR #5097, opened 2017-10-21
Ignore logging some tcp routine errors

### LINKED ISSUE #5096: Error log after accessing to Minio Browser using Chrome/Firefox when TLS is enabled
## Expected Behavior
No log since there is no error.

## Current Behavior
```
time="2017-10-21T11:26:33Z" level=error msg="Error in reading from new TLS connection 84.232.207.221:56911 at server 172.17.0.2:9000" cause=EOF source="[listener.go:188:github.com/minio/minio/pkg/http.(*httpListener).start.func2()]"
time="2017-10-21T11:26:33Z" level=error msg="Error in reading from new TLS connection 84.232.207.221:56910 at server 172.17.0.2:9000" cause=EOF source="[listener.go:188:github.com/minio/minio/pkg/http.(*httpListener).start.func2()]"
time="2017-10-21T11:26:33Z" level=error msg="Error in reading from new TLS connection 84.232.207.221:56913 at server 172.17.0.2:9000" cause=EOF source="[listener.go:188:github.com/minio/minio/pkg/http.(*httpListener).start.func2()]"
time="2017-10-21T11:26:33Z" level=error msg="Error in reading from new TLS connection 84.232.207.221:56912 at server 172.17.0.2:9000" cause=EOF source="[listener.go:188:github.com/minio/minio/pkg/http.(*httpListener).start.func2()]"
```

## Possible Solution
Disabling peeking after a successful handshake, browsers seem to open a TLS connection but doesn't really send any data 

1. Configure a Minio server with TLS certificate
2. Open Minio Browser with Chrome or Firefox (and probably other browsers)

## Context
Fixing a bug

## Your Environment
* Version used (`minio version`): 65a817fe8c49564abf7662cc2cbc9e09009661c7
* Operating System and version (`uname -a`):  Ubuntu 16.04

vadmeste:
## Description
This will avoid printing non useful logs when browsers open connections
but doesn't send any data or prematurely close connections

## Motivation and Context
Fixes #5096 

## How Has This Been Tested?
go test + Manually

## Types of changes

## Checklist:

### DIFF
--- a/pkg/http/listener.go
+++ b/pkg/http/listener.go
@@ -23,6 +23,7 @@ import (
 	"io"
 	"net"
 	"net/http"
+	"os"
 	"strings"
 	"sync"
 	"syscall"
@@ -89,13 +90,19 @@ type httpListener struct {
 	errorLogFunc           func(error, string, ...interface{}) // function to be called on errors.
 }
 
-// ignoreErr returns true if error is due to a network timeout or
-// io.EOF and false otherwise
-func ignoreErr(err error) bool {
+// isRoutineNetErr returns true if error is due to a network timeout,
+// connect reset or io.EOF and false otherwise
+func isRoutineNetErr(err error) bool {
 	if nErr, ok := err.(*net.OpError); ok {
+		// Check if the error is a tcp connection reset
+		if syscallErr, ok := nErr.Err.(*os.SyscallError); ok {
+			if errno, ok := syscallErr.Err.(syscall.Errno); ok {
+				return errno == syscall.ECONNRESET
+			}
+		}
+		// Check if the error is a timeout
 		return nErr.Timeout()
 	}
-
 	return err == io.EOF
 }
 
@@ -138,7 +145,7 @@ func (listener *httpListener) start() {
 				// speed up loading of a web page. Peek may also fail due to network
 				// saturation on a transport with read timeout set. All other kind of
 				// errors should be logged for further investigation. Thanks @brendanashworth.
-				if !ignoreErr(err) {
+				if !isRoutineNetErr(err) {
 					listener.errorLogFunc(err,
 						"Error in reading from new connection %s at server %s",
 						bufconn.RemoteAddr(), bufconn.LocalAddr())
@@ -182,7 +189,7 @@ func (listener *httpListener) start() {
 			// Peek bytes of maximum length of all HTTP methods.
 			data, err := bufconn.Peek(methodMaxLen)
 			if err != nil {
-				if listener.errorLogFunc != nil {
+				if !isRoutineNetErr(err) && listener.errorLogFunc != nil {
 					listener.errorLogFunc(err,
 						"Error in reading from new TLS connection %s at server %s",
 						bufconn.RemoteAddr(), bufconn.LocalAddr())
--- a/pkg/http/listener_test.go
+++ b/pkg/http/listener_test.go
@@ -826,7 +826,7 @@ func TestIgnoreErr(t *testing.T) {
 	}
 
 	for i, tc := range testCases {
-		if actual := ignoreErr(tc.err); actual != tc.want {
+		if actual := isRoutineNetErr(tc.err); actual != tc.want {
 			t.Errorf("Test case %d: Expected %v but got %v for %v", i+1,
 				tc.want, actual, tc.err)
 		}


pkg/http/listener.go
```diff
@@ -179,24 +179,8 @@ func (listener *httpListener) start() {
...
-				bufconn.Close()
-				return
-			}
-
-			// Return bufconn if read data is a valid HTTP method.
-			tokens := strings.SplitN(string(data), " ", 2)
-			if isHTTPMethod(tokens[0]) {
-				send(acceptResult{bufconn, nil}, doneCh)
-				return
-			}
```
harshavardhana: This would be incorrect @vadmeste Peek is fine but we don't need to print for `io.EOF` that's all. Read the handleConn above..

```
  // Peek could fail legitimately when clients abruptly close
  // connection. E.g. Chrome browser opens connections speculatively to
  // speed up loading of a web page. Peek may also fail due to network
  // saturation on a transport with read timeout set. All other kind of
  // errors should be logged for further investigation. 
  if !ignoreErr(err) {
```

harshavardhana (changes requested):

pkg/http/listener.go
vadmeste: We don't need peeking at this stage @harshavardhana, peeking is only to differentiate between TLS & plain HTTP, but here we are sure that this is a TLS connection because the handshake was already successful. Golang HTTP can handle HTTP errors already and we don't need to add our own code for that.

harshavardhana: No it is on purpose @vadmeste we are also validating if the HTTP method sent is correct as well.

vadmeste: > No it is on purpose @vadmeste we are also validating if the HTTP method sent is correct as well.

Won't golang http also validate that ?

harshavardhana: We are not worried about that @vadmeste `pkg/http` is a lower level implementation with our own requirements.  It is because when you manage a `conn`  for HTTP you are supposed to handle everything related to HTTP at this level. Passing the issue to Go and expect it to fail in unexpected ways would be incosistent.  

`pkg/http` is not a thin layer implementation it is purposefully built to handle these conditions.

vadmeste: > Passing the issue to Go and expect it to fail in unexpected ways would be incosistent.

Are you saying that pkg/http should validate all the http protocol ?

harshavardhana: I mean the Conn validation. Not HTTP parsing. Simple fix is what I mentioned above. 

Removing method validation is incorrect.

vadmeste:
PR updated.

pkg/http/listener.go
```diff
@@ -89,13 +90,22 @@ type httpListener struct {
 	errorLogFunc           func(error, string, ...interface{}) // function to be called on errors.
 }
 
-// ignoreErr returns true if error is due to a network timeout or
+// isRoutineNetErr returns true if error is due to a network timeout or
 // io.EOF and false otherwise
-func ignoreErr(err error) bool {
+func isRoutineNetErr(err error) bool {
```
balamurugana: I think you can retain the function name as `ignoreErr()`. It is clear enough I guess.

pkg/http/listener.go
```diff
@@ -89,13 +90,22 @@ type httpListener struct {
...
+		if nErr.Timeout() {
+			return true
+		}
+		// Check if the error is a tcp connection reset
+		if syscallErr, ok := nErr.Err.(*os.SyscallError); ok {
+			if errno, ok := syscallErr.Err.(syscall.Errno); ok {
+				return errno == syscall.ECONNRESET
+			}
+		}
+		return false
```
balamurugana: You could do `syscall.ECONNREST` first.  This return statement can be avoided

pkg/http/listener.go
```diff
@@ -182,10 +192,12 @@ func (listener *httpListener) start() {
 			// Peek bytes of maximum length of all HTTP methods.
 			data, err := bufconn.Peek(methodMaxLen)
 			if err != nil {
-				if listener.errorLogFunc != nil {
-					listener.errorLogFunc(err,
-						"Error in reading from new TLS connection %s at server %s",
-						bufconn.RemoteAddr(), bufconn.LocalAddr())
+				if !isRoutineNetErr(err) {
+					if listener.errorLogFunc != nil {
```
balamurugana: You could do `if listener.errorLogFunc != nil && !ignoreErr(err)` to avoid necessary nesting `if`

vadmeste: Done

pkg/http/listener.go
vadmeste: Emm.. Actually ignoreErr() is too generic for me, isRoutineNetErr() is confusing ?

nitisht:
Is this ready for review again @vadmeste

harshavardhana (changes requested):
Looks like this PR has unexpected changes from other commits. @vadmeste

harshavardhana:
Can you rebase this PR @vadmeste ?

vadmeste:
PR updated.

harshavardhana (approved):

harshavardhana:
@balamurugana can you take a look?

balamurugana (approved):
LGTM

PR #5099, opened 2017-10-22
convert ETag properly for all gateways

balamurugana:
Previously ID/ETag from backend service is used as is which causes
failure on s3cmd like tools where those tools use ETag as checksum to
validate data.  This is fixed by prepending "-1".

Refer minio/mint#193 minio/mint#201

## Types of changes

## Checklist:

### DIFF
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -211,3 +211,21 @@ func checkURL(urlStr string) (*url.URL, error) {
 func UTCNow() time.Time {
 	return time.Now().UTC()
 }
+
+// genETag - generate UUID based ETag
+func genETag() string {
+	return toS3ETag(getMD5Hash([]byte(mustGetUUID())))
+}
+
+// toS3ETag - return checksum to ETag
+func toS3ETag(etag string) string {
+	etag = canonicalizeETag(etag)
+
+	if !strings.HasSuffix(etag, "-1") {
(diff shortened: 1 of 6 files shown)


harshavardhana (dismissed):

harshavardhana:
Tested the changes with azure gateway, etag returned is working fine with latest master of s3cmd.

cmd/utils.go
```diff
@@ -211,3 +211,21 @@ func checkURL(urlStr string) (*url.URL, error) {
 func UTCNow() time.Time {
 	return time.Now().UTC()
 }
+
+// genETag - generate UUID based ETag
+func genETag() string {
```
krisis, line 216: can we call this function genRandomETag or getUniqueETag?

harshavardhana:
This PR is release blocking as well @deekoder - we need to do this for s3cmd to work properly we have mint tests for this, since this is a minor change as well.

harshavardhana (approved):
Works tested across Azure, GCS, B2

krisis (approved):

cmd/utils.go
balamurugana: Most of the time we generate ETag.  This means there is no need to explicitly tell about randomness or uniqueness

deekoder:
Taking this in as per discussions. Ignoring code coverage as expected behavior on gateway.

PR #5101, opened 2017-10-23
Update documenation to reflect correct default region of ''

kula:
Ever since commit 5db1e9f3dd74ecce06cbab1cf1c52930a62a302f the default region as been '', instead of 'us-east-1'. Update documentation to reflect this, in particular, documentation about notifications.

## Description

Two documents were updated to reflect the current default for the `region` setting.

## Motivation and Context

If you try to follow the documentation to set up notifications, you'll be prompted to install event arns with a region set to `us-east-1`, which will cause event notifications to silently not work, as `minio` believes it is in the `` region.

## How Has This Been Tested?

No testing, documentation changes only.

## Types of changes

## Checklist:

### DIFF
--- a/docs/bucket/notifications/README.md
+++ b/docs/bucket/notifications/README.md
@@ -65,7 +65,7 @@ An example configuration for RabbitMQ is shown below:
 }
 ```
 
-After updating the configuration file, restart the Minio server to put the changes into effect. The server will print a line like `SQS ARNs:  arn:minio:sqs:us-east-1:1:amqp` at start-up if there were no errors.
+After updating the configuration file, restart the Minio server to put the changes into effect. The server will print a line like `SQS ARNs:  arn:minio:sqs::1:amqp` at start-up if there were no errors.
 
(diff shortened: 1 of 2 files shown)


harshavardhana:
Thanks for the PR but this might be a bug.. when region is not set all regions should be honored..

deekoder (approved):

deekoder (approved):
I found that this issue also happened in sqs arn. I had to remove us-east-1. So i approve.

krisis (approved):
Thanks for the PR @kula. This document change reflects the current behaviour. We can take this in and address server behaviour later, if required.

deekoder:
I will open a separate issue for the server also being ok with us-east-1 in the arn.

harshavardhana:
us-east-1 ARN seems to be not the problem here 

```
mc events add myminio/testbucket arn:minio:sqs:us-east-1:test1:redis
Successfully added arn:minio:sqs:us-east-1:test1:redis
```

This is successful even with server region being empty, i see that we already have code to handle this correctly. But there is a problem when sending events though. 

While it is saved as is the notification is picked from a map which doesn't store it in the same manner, this is a bug. Doc fix is correct but the correct fix is .. https://github.com/minio/minio/pull/5119

PR #5106, opened 2017-10-23
Set etag properly to a calculated value

### LINKED ISSUE #5100: server generates emtpy etag when client does not set content-md5
## Expected Behavior
Currently the server should compute the MD5 sum of the object and use this value as etag if the client does not set the content-md5. (Actually we can set a random value as etag - see #4385 #4641)  

fs.json
```

{
   "version":"1.0.1",
   "format":"fs","
    minio":{ 
             "release":"DEVELOPMENT.GOGET"
    },
   "meta":{
            "content-type":"application/octet-stream",
            "etag":"5eb63bbbe01eeed093cb22bb8f5acdc3"
   }
}
```

## Current Behavior
The server saves an empty value as etag. 

fs.json
```

{
   "version":"1.0.1",
   "format":"fs","
    minio":{ 
             "release":"DEVELOPMENT.GOGET"
    },
   "meta":{
            "content-type":"application/octet-stream",
            "etag":""
   }
}
```

## Possible Solution
The problem is within the changes of #5071. 
1. The `HashReader` should return the computed MD5 not the MD5 sum provided by the client (it may be emtpy) - See [MD5()](https://github.com/minio/minio/blob/master/pkg/hash/reader.go#L104)
2. The etag (metadata field) should be set after the client has uploaded the whole object (see Expected Behavior) - See  FS [Put](https://github.com/minio/minio/blob/master/cmd/fs-v1.go#L596)

## Context
Encrypted objects "don't have" an etag value any more. 

## Your Environment
current master

harshavardhana:
## Description
Set etag properly to a calculated value

## Motivation and Context
Fixes https://github.com/minio/minio/issues/5100

## How Has This Been Tested?
Manually without setting `content-md5` 

## Types of changes

## Checklist:

### DIFF
--- a/cmd/fs-v1.go
+++ b/cmd/fs-v1.go
@@ -17,6 +17,7 @@
 package cmd
 
 import (
+	"encoding/hex"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -593,8 +594,6 @@ func (fs fsObjects) PutObject(bucket string, object string, data *hash.Reader, m
 		return ObjectInfo{}, traceError(errInvalidArgument)
 	}
 
-	metadata["etag"] = data.MD5HexString()
-
 	var wlk *lock.LockedFile
 	if bucket != minioMetaBucket {
 		bucketMetaDir := pathJoin(fs.fsPath, minioMetaBucket, bucketMetaPrefix)
@@ -633,6 +632,8 @@ func (fs fsObjects) PutObject(bucket string, object string, data *hash.Reader, m
 		errorIf(err, "Failed to create object %s/%s", bucket, object)
 		return ObjectInfo{}, toObjectErr(err, bucket, object)
 	}
+
+	metadata["etag"] = hex.EncodeToString(data.MD5Current())
 
 	// Should return IncompleteBody{} error when reader has fewer
 	// bytes than specified in request header.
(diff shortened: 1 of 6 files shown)


cmd/fs-v1.go
```diff
@@ -593,7 +594,7 @@ func (fs fsObjects) PutObject(bucket string, object string, data *hash.Reader, m
 		return ObjectInfo{}, traceError(errInvalidArgument)
 	}
 
-	metadata["etag"] = data.MD5HexString()
```
aead, line 596: IMO we still have to move this line below the the `fsCreate` Otherwise the etag is always the same for PUTs without content-md5

harshavardhana: Done @aead

pkg/hash/reader.go
```diff
@@ -52,13 +52,12 @@ func NewReader(src io.Reader, size int64, md5Hex, sha256Hex string) (*Reader, er
 	if err != nil {
 		return nil, err
 	}
+
 	var (
 		md5Hash    hash.Hash
```
aead: md5Hash is not required - we can directly use `md5Hash: md5.New()` during Reader init.

aead (changes requested):
Everything else looks good to me 👍 
We have to revisit this again when we change etag with `-1` or `-part-number`.

pkg/hash/reader.go
harshavardhana: Sure let me do that.

vadmeste (approved):
LGTM & Tested

aead (approved):
LGTM

Issue #5093, opened 2017-10-20
Error too many open files (ECONNRESET)


danielfdsilva:
## Expected Behavior
It should be able to delete multiple files without crashing.

## Current Behavior
When trying to remove multiple files from minio (circa 10.000) the client (minio-js) throws an error:
```
Unhandled rejection Error: getaddrinfo ENOTFOUND localhost localhost:9000
    at errnoException (dns.js:28:10)
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (dns.js:76:26)

Error: read ECONNRESET
    at exports._errnoException (util.js:1020:11)
    at TCP.onread (net.js:568:26)
```
and the server logs:
```
ERRO[0336] Unable to delete an object rra/tiles/b/14/6472/8685.pbf  cause=open /guts/projects/rra/.resources/test-minio/minio-data/.minio.sys/buckets/rra/tiles/b/14/6472/8685.pbf/fs.json: too many open files in system source=[object-handlers.go:1077:objectAPIHandlers.DeleteObjectHandler()] stack=fs-v1.go:647:fsObjects.DeleteObject <autogenerated>:424:(*fsObjects).DeleteObject object-handlers-common.go:241:deleteObject object-handlers.go:1076:objectAPIHandlers.DeleteObjectHandler api-router.go:62:(objectAPIHandlers).DeleteObjectHandler-fm
```

## Steps to Reproduce (for bugs)
To delete the files I'm getting the list with `listObjectsV2` (storing them in an array) and then looping over the array and deleting each file with `removeObject`.

## Your Environment
Happens both with the minio docker container and installed locally on a Mac OS Sierra 10.12.6
Doesn't seem to happen if I use an S3 bucket with the js client.

harshavardhana:
what version of minio server are you running? @danielfdsilva  with `docker run --rm minio/minio version` ?  also what can you provide the system limit here? like `ulimit -n` ?

danielfdsilva:
@harshavardhana I installed the server today using homebrew:
```
Version: 2017-09-29T19:16:56Z
Release-Tag: RELEASE.2017-09-29T19-16-56Z
Commit-ID: 60cc6184d253efee4a3120683517028342229e21
```
The docker version is the same:
```
Version: 2017-09-29T19:16:56Z
Release-Tag: RELEASE.2017-09-29T19-16-56Z
Commit-ID: 60cc6184d253efee4a3120683517028342229e21
```

`ulimit -n` = 7128

[...]

PR #5109, opened 2017-10-24
Limit number of connections upto system maxlimit

harshavardhana:
## Description
Limit number of connections upto system maxlimit

## Motivation and Context
Refer #5093 

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:

### DIFF
--- a/cmd/api-errors.go
+++ b/cmd/api-errors.go
@@ -122,10 +122,10 @@ const (
 	ErrMetadataTooLarge
 	ErrUnsupportedMetadata
 	ErrMaximumExpires
+	ErrSlowDown
 	// Add new error codes here.
 
 	// Server-Side-Encryption (with Customer provided key) related API errors.
-
 	ErrInsecureSSECustomerRequest
 	ErrSSEEncryptedObject
 	ErrInvalidEncryptionParameters
@@ -505,6 +505,11 @@ var errorCodeResponse = map[APIErrorCode]APIError{
 		Code:           "AccessDenied",
 		Description:    "Request is not valid yet",
 		HTTPStatusCode: http.StatusForbidden,
+	},
+	ErrSlowDown: {
+		Code:           "SlowDown",
+		Description:    "Please reduce your request",
+		HTTPStatusCode: http.StatusServiceUnavailable,
 	},
 	// FIXME: Actual XML error response also contains the header which missed in list of signed header parameters.
 	ErrUnsignedHeaders: {
--- a/cmd/generic-handlers.go
+++ b/cmd/generic-handlers.go
@@ -18,13 +18,16 @@ package cmd
 
 import (
 	"bufio"
+	"context"
 	"net"
 	"net/http"
 	"strings"
 	"time"
 
 	humanize "github.com/dustin/go-humanize"
+	"github.com/minio/minio/pkg/sys"
 	"github.com/rs/cors"
+	"golang.org/x/time/rate"
 )
 
 // HandlerFunc - useful to chain different middleware http.Handler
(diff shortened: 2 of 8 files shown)


pkg/http/listener.go
```diff
@@ -99,9 +101,11 @@ func ignoreErr(err error) bool {
...
 }
 
-// start - starts separate goroutine for each TCP listener.  A valid insecure/TLS HTTP new connection is passed to httpListener.acceptCh.
-func (listener *httpListener) start() {
-	listener.acceptCh = make(chan acceptResult)
+// start - starts separate goroutine for each TCP listener.  A valid insecure/TLS HTTP new
+// connection is passed to httpListener.acceptCh, acceptch is also limited by input n number
+// of connections.
+func (listener *httpListener) start(nConn uint64) {
+	listener.acceptCh = make(chan acceptResult, nConn)
```
balamurugana: httpListener works same as regular listener i.e it blocks next connection till accepted connection is passed to upper layer. This change would lead incompatible problem.  I think this limitation needs to be handled upper layer than listener

cmd/generic-handlers.go
```diff
@@ -525,3 +528,45 @@ func (h pathValidityHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 	h.handler.ServeHTTP(w, r)
 }
+
+// setRateLimitHandler middleware limits the throughput to h using a
+// rate.Limiter token bucket configured with maxOpenFileLimit/2 and
```
fwessels: Limit is set to MaxOpenFileLimit (line 544), correct? (And not /2)

cmd/generic-handlers.go
```diff
@@ -525,3 +528,45 @@ func (h pathValidityHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+}
+
+func (l rateLimit) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	// create a new context from the request with the wait timeout
+	ctx, cancel := context.WithTimeout(r.Context(), 1*time.Second)
+	defer cancel() // always cancel the context!
+
+	// Wait errors out if the request cannot be processed within
+	// the deadline. time/rate tries to reserve a slot if possible
+	// with in the given duration if its not possible then Wait(ctx)
```
fwessels: its --> it's

cmd/generic-handlers.go
```diff
@@ -525,3 +528,45 @@ func (h pathValidityHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	// returns an error and we cancel the request with ErrSlowDown
+	// error message to the client. This context wait also ensures
+	// requests doomed to fail are terminated early, preventing a
+	// potential pileup on the server.
+	if err := l.Wait(ctx); err != nil {
+		// Send an S3 compatible error, SlowDown.
+		writeErrorResponse(w, ErrSlowDown, r.URL)
```
fwessels, line 596: would it make sense to add an hint in the form of a Retry-After header ? (eg. default 60 seconds or 120 seconds)

cmd/generic-handlers.go
```diff
@@ -525,3 +528,45 @@ func (h pathValidityHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+
+	// Wait errors out if the request cannot be processed within
+	// the deadline. time/rate tries to reserve a slot if possible
+	// with in the given duration if its not possible then Wait(ctx)
+	// returns an error and we cancel the request with ErrSlowDown
+	// error message to the client. This context wait also ensures
+	// requests doomed to fail are terminated early, preventing a
+	// potential pileup on the server.
+	if err := l.Wait(ctx); err != nil {
+		// Send an S3 compatible error, SlowDown.
```
fwessels, line 595: Do we want to send a generic error when err != 'NoSlot' (or whatever the error is when no slot is available).

fwessels (dismissed):
Apart from minor comments, LGTM

cmd/generic-handlers.go
harshavardhana: Wait is actually an external code. So no control over what they are sending. Also Wait returns only if no slot is available.

fwessels: ok, then we keep it as as.

harshavardhana:
Addressed your comments @fwessels @vadmeste  PTAL.

harshavardhana:
> Addressed your comments @fwessels @vadmeste PTAL.

Addressed your comments @fwessels @vadmeste PTAL

cmd/api-errors.go
```diff
@@ -506,6 +506,11 @@ var errorCodeResponse = map[APIErrorCode]APIError{
 		Description:    "Request is not valid yet",
 		HTTPStatusCode: http.StatusForbidden,
 	},
+	ErrSlowDown: {
```
vadmeste, line 509: I think 'ServiceUnavailable' is better than SlowDown here.. isn't slow down returned to a specific user who is sending too many requests ?

vadmeste: > I think 'ServiceUnavailable' is better than SlowDown here.. isn't slow down returned to a specific user who is sending too many requests ?

I withdraw what I said. We already have only one user.

harshavardhana: the reason for slowdown is AWS S3 error compliance @vadmeste the http status is correct and in accordance with `Retry-After` http header - https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After

vadmeste:
This PR is too complicated for me to understand.. let's discuss this when you have time @harshavardhana

harshavardhana:
> This PR is too complicated for me to understand.. let's discuss this when you have time @harshavardhana

Whenever you want to talk please reach out to me on slack

vadmeste (approved):
LGTM

harshavardhana:
@fwessels you might have to review this again.

fwessels:
LGTM, and nice implementation of Retry-After

harshavardhana:
> LGTM, and nice implementation of Retry-After

@fwessels you might have to click 'approve'

fwessels (approved):

PR #5103, opened 2017-10-23
Implement Alibaba Cloud OSS gateway support

timonwong:
## Motivation and Context

Implement Alibaba Cloud (Aliyun) [OSS](https://www.alibabacloud.com/product/oss) gateway support.

API Documentation (very similar to AWS S3): https://www.alibabacloud.com/help/doc-detail/31947.htm

## How Has This Been Tested?

Manually & Mint

Notes about failed mint tests:
- Since OSS doesn't support underscores(`"_"`) in custom metadata headers, so `aws-sdk-php` will fail. (Workaround by `sed -i 's/Param_1/Param-1/' /mint/run/core/aws-sdk-php/quick-tests.php`)
- minio-dotnet failed during put objects (tons of connect failed errors, may be caused by too many file descriptors)

Besides, in order to get mint tests pass, there is a check in `CompleteMultipartUpload()` which iterate uploaded parts, and error out if uploadedParts except last part sizing < 5MiB.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #5111, opened 2017-10-24
Simplify gateway backend registration

timonwong:
## Motivation and Context

When I'm trying to add a new gateway backend in #5103, several files need to be touched, which IMHO is not very convenient.
In this pull request, gateway backend implementations only need to focus their own scope, and 
 just invoke `mustRegisterGatewayBackend()` function in their `init()`.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/gateway-azure.go
+++ b/cmd/gateway-azure.go
@@ -33,13 +33,86 @@ import (
 
 	"github.com/Azure/azure-sdk-for-go/storage"
 	humanize "github.com/dustin/go-humanize"
+	"github.com/minio/cli"
 	"github.com/minio/minio-go/pkg/policy"
 	"github.com/minio/minio/pkg/hash"
 )
 
-const globalAzureAPIVersion = "2016-05-31"
-const azureBlockSize = 100 * humanize.MiByte
-const metadataObjectNameTemplate = globalMinioSysTmp + "multipart/v1/%s.%x/azure.json"
+const (
+	globalAzureAPIVersion      = "2016-05-31"
+	azureBlockSize             = 100 * humanize.MiByte
+	metadataObjectNameTemplate = globalMinioSysTmp + "multipart/v1/%s.%x/azure.json"
+	azureBackend               = "azure"
+)
+
+func init() {
+	const azureGatewayTemplate = `NAME:
+  {{.HelpName}} - {{.Usage}}
+
+USAGE:
+  {{.HelpName}} {{if .VisibleFlags}}[FLAGS]{{end}} [ENDPOINT]
+{{if .VisibleFlags}}
+FLAGS:
+  {{range .VisibleFlags}}{{.}}
+  {{end}}{{end}}
+ENDPOINT:
+  Azure server endpoint. Default ENDPOINT is https://core.windows.net
+
+ENVIRONMENT VARIABLES:
+  ACCESS:
+     MINIO_ACCESS_KEY: Username or access key of Azure storage.
+     MINIO_SECRET_KEY: Password or secret key of Azure storage.
+
+  BROWSER:
+     MINIO_BROWSER: To disable web browser access, set this value to "off".
+
+EXAMPLES:
+  1. Start minio gateway server for Azure Blob Storage backend.
+      $ export MINIO_ACCESS_KEY=azureaccountname
+      $ export MINIO_SECRET_KEY=azureaccountkey
+      $ {{.HelpName}}
+
+  2. Start minio gateway server for Azure Blob Storage backend on custom endpoint.
+      $ export MINIO_ACCESS_KEY=azureaccountname
+      $ export MINIO_SECRET_KEY=azureaccountkey
+      $ {{.HelpName}} https://azure.example.com
+
+`
+
+	MustRegisterGatewayCommand(cli.Command{
+		Name:               azureBackend,
+		Usage:              "Microsoft Azure Blob Storage.",
+		Action:             azureGatewayMain,
+		CustomHelpTemplate: azureGatewayTemplate,
+		Flags:              append(serverFlags, globalFlags...),
+		HideHelpCommand:    true,
+	})
+}
+
+// Handler for 'minio gateway azure' command line.
+func azureGatewayMain(ctx *cli.Context) {
+	// Validate gateway arguments.
+	host := ctx.Args().First()
+	// Validate gateway arguments.
+	fatalIf(validateGatewayArguments(ctx.GlobalString("address"), host), "Invalid argument")
+
+	startGateway(ctx, &AzureGateway{host})
+}
+
+// AzureGateway implements Gateway.
+type AzureGateway struct {
+	host string
+}
+
+// Name implements Gateway interface.
+func (g *AzureGateway) Name() string {
+	return azureBackend
+}
+
+// NewGatewayLayer initializes azure blob storage client and returns AzureObjects.
+func (g *AzureGateway) NewGatewayLayer() (GatewayLayer, error) {
+	return newAzureLayer(g.host)
+}
 
 // s3MetaToAzureProperties converts metadata meant for S3 PUT/COPY
 // object into Azure data structures - BlobMetadata and
--- a/cmd/gateway-main.go
+++ b/cmd/gateway-main.go
@@ -31,212 +31,41 @@ import (
 	miniohttp "github.com/minio/minio/pkg/http"
 )
 
-const azureGatewayTemplate = `NAME:
-  {{.HelpName}} - {{.Usage}}
-
-USAGE:
-  {{.HelpName}} {{if .VisibleFlags}}[FLAGS]{{end}} [ENDPOINT]
-{{if .VisibleFlags}}
-FLAGS:
-  {{range .VisibleFlags}}{{.}}
-  {{end}}{{end}}
-ENDPOINT:
-  Azure server endpoint. Default ENDPOINT is https://core.windows.net
-
-ENVIRONMENT VARIABLES:
-  ACCESS:
-     MINIO_ACCESS_KEY: Username or access key of Azure storage.
-     MINIO_SECRET_KEY: Password or secret key of Azure storage.
-
-  BROWSER:
-     MINIO_BROWSER: To disable web browser access, set this value to "off".
-
-EXAMPLES:
-  1. Start minio gateway server for Azure Blob Storage backend.
-      $ export MINIO_ACCESS_KEY=azureaccountname
-      $ export MINIO_SECRET_KEY=azureaccountkey
-      $ {{.HelpName}}
-
-  2. Start minio gateway server for Azure Blob Storage backend on custom endpoint.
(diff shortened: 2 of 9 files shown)


cmd/gateway-main.go
```diff
@@ -31,212 +31,77 @@ import (
...
+func newGatewayLayer(backendType gatewayBackend, args cli.Args) (GatewayLayer, error) {
+	if reg, ok := gatewayBackendReg[backendType]; ok {
+		if reg.IsExperimental {
+			log.Println(colorYellow("\n               *** Warning: Not Ready for Production ***"))
+		}
+		return reg.New(args)
 	}
 
-	return nil, fmt.Errorf("Unrecognized backend type %s", backendType)
+	return nil, fmt.Errorf("unrecognized backend type %s", backendType)
```
harshavardhana: Capitalize the strings for errors, this is to be consistent with our code base.

cmd/gateway-s3.go
```diff
@@ -20,12 +20,71 @@ import (
...
+      $ {{.HelpName}}
+
+  2. Start minio gateway server for S3 backend on custom endpoint.
+      $ export MINIO_ACCESS_KEY=Q3AM3UQ867SPQQA43P2F
+      $ export MINIO_SECRET_KEY=zuf+tfteSlswRu7BJ86wekitnifILbZam1KYY3TG
+      $ {{.HelpName}} https://play.minio.io:9000
+`
+
+	mustRegisterGatewayBackend(gatewayBackendEntry{
+		Name: "s3",
```
harshavardhana: we can use `s3Backend` const here..

cmd/gateway-s3.go
```diff
@@ -20,12 +20,71 @@ import (
...
+		Cmd: cli.Command{
+			Name:               "s3",
```
harshavardhana: same as above and can be replicated for all gateway backends..

cmd/gateway-s3.go
```diff
@@ -102,12 +161,13 @@ type s3Objects struct {
 }
 
 // newS3Gateway returns s3 gatewaylayer
-func newS3Gateway(host string) (GatewayLayer, error) {
+func newS3Gateway(args cli.Args) (GatewayLayer, error) {
```
harshavardhana: if a gateway takes only one argument then we don't need to expand the scope here by adding more args. This refactor may have led to that.  But taking in input param as `host` is more meaningful here.

cmd/gateway-main.go
```diff
@@ -31,212 +31,77 @@ import (
...
-	gcsBackend   gatewayBackend = "gcs"
-	b2Backend    gatewayBackend = "b2"
-	// Add more backends here.
-)
+// Represents the type of each item in gateway backend registry.
+type gatewayBackendEntry struct {
+	Name           gatewayBackend
+	Cmd            cli.Command
+	New            func(args cli.Args) (GatewayLayer, error)
+	IsExperimental bool
```
harshavardhana: We don't need to call this `IsExperimental` just say it as `Experimental` is fine.

cmd/gateway-main.go
```diff
@@ -31,212 +31,77 @@ import (
...
-
 	gatewayCmd = cli.Command{
 		Name:            "gateway",
 		Usage:           "Start object storage gateway.",
 		Flags:           append(serverFlags, globalFlags...),
 		HideHelpCommand: true,
-		Subcommands:     []cli.Command{azureBackendCmd, s3BackendCmd, gcsBackendCmd, b2BackendCmd},
 	}
+
+	gatewayBackendReg = map[gatewayBackend]gatewayBackendEntry{}
```
harshavardhana: We can avoid map and make this more safer by using a slice instead?

cmd/gateway-main.go
```diff
@@ -31,212 +31,77 @@ import (
...
-const (
-	azureBackend gatewayBackend = "azure"
-	s3Backend    gatewayBackend = "s3"
-	gcsBackend   gatewayBackend = "gcs"
-	b2Backend    gatewayBackend = "b2"
-	// Add more backends here.
-)
+// Represents the type of each item in gateway backend registry.
+type gatewayBackendEntry struct {
+	Name           gatewayBackend
```
harshavardhana: there is no reason for this type if there are no constants anymore.

cmd/gateway-main.go
```diff
@@ -31,212 +31,77 @@ import (
...
+func (e *gatewayBackendEntry) Validate() error {
+	if e.Name == "" {
+		return errors.New("field 'Name' cannot be empty")
+	}
+
+	if e.New == nil {
+		return errors.New("field 'New' cannot be nil")
+	}
+
+	if e.Cmd.Action != nil {
```
harshavardhana: This check is sort of incorrect with the overall usage and pattern,  There is no reason we should add this check as it might be frivolous indeed. We should run a Validate() func right after the Action is registered, or just remove verifying Action.

cmd/gateway-main.go
```diff
@@ -31,212 +31,77 @@ import (
...
+	if err := entry.Validate(); err != nil {
+		return err
+	}
+
+	backendType := entry.Name
+	if _, found := gatewayBackendReg[backendType]; found {
+		return fmt.Errorf("duplicate backend type found in registry: %s", string(backendType))
+	}
+
+	entry.Cmd.Action = gatewayMainAction(backendType)
```
harshavardhana: This additional code here is not needed, modifying input structs causing side-affects is not a good idea usually. We should simply move this code to mustRegisterGatewayBackend() - `gatewayMainAction("s3")` for example.

harshavardhana: So all you would have to do would to do is just simply do validation here and avoid mutation.

cmd/gateway-azure.go
```diff
@@ -292,12 +354,13 @@ func azureParseBlockID(blockID string) (partID, subPartNumber int, uploadID, md5
 }
 
 // Inits azure blob storage client and returns AzureObjects.
-func newAzureLayer(host string) (GatewayLayer, error) {
+func newAzureLayer(args cli.Args) (GatewayLayer, error) {
```
harshavardhana: The issue about changing input params similar to my comment on s3 gateway

harshavardhana (changes requested):
Overall a nice change, but some comments about style and structure.

cmd/gateway-main.go
timonwong: > We can avoid map and make this more safer by using a slice instead?

Not sure, would you mind telling why a slice is more safer in this scenario?

balamurugana (changes requested):
We could have an `interface` approach to solve the problem.  That way each gateway backend is completely independent.  Below is the approach

gateway-main.go:
* Add an interface
```go
type Gateway interface{
    Name() string
    NewGatewayLayer() (GatewayLayer, error)
}
```
* Rename `gatewayMain()` to `startGateway()`

gateway-s3.go:
* In `init()` add subcommand by `gatewayCmd.Subcommands = append(gatewayCmd.Subcommands, cli.Command{...})`
* Have `S3Gateway`  compatible to `Gateway` interface
```go
type S3Gateway struct {
    host string
}

func (g *S3Gateway) Name() string {
    return "s3"
}

func (g *S3Gateway) NewObjectLayer() (GatewayLayer, error) {
    // log.Println(colorYellow("\n               *** Warning: Not Ready for Production ***"))
    // have newS3Gateway() here
}
```
This logic needs to be done in all gateway code

timonwong:
@balamurugana 

I think we should not access  `gatewayCmd.Subcommands` directly in each gateway implementation.
My ultimate goal (perhaps) is the possibility of adding "plugin" support, so we can easily opt-in/opt-out.

What about this?

```go
type GatewayCommand interface {
	// Name is the global unique name for each gateway.
	Name() string
	// A short description of the usage of this command.
	Description() string
	// HelpTemplate returns the text template for the command help topic.
	HelpTemplate() string
	// Parse parse the command line context.
	Parse(ctx *cli.Context) error
	// NewGatewayLayer returns a new gateway layer.
	NewGatewayLayer() (GatewayLayer, error)
}

func RegisterGatewayCommand(command GatewayCommand) error {
	// We should not have multiple subcommands with same name.
	for _, c := range gatewayCmd.Subcommands {
		if c.Name == command.Name() {
			return fmt.Errorf("duplicate gateway command: %s", command.Name())
		}
	}

	gatewayCmd.Subcommands = append(gatewayCmd.Subcommands, cli.Command{
		Name:               command.Name(),
		Usage:              command.Description(),
		Action:             gatewayCommandMain(command),
		CustomHelpTemplate: command.HelpTemplate(),
		Flags:              append(serverFlags, globalFlags...),
		HideHelpCommand:    true,
	})
	return nil
}
```

balamurugana:
@timonwong This approach brings a cyclic usage i.e. `gateway-main.go` calls a method on registration.  To make things simple, you can have a `registerCommand()` in `gateway-main.go` and every gateway layer calls this function for its registration for `cli` package.  This way each knowledge is abstracted within each gateway layer.

timonwong:
Can you have a look latest commit, though I'm afraid it's not very feasible, because:
1. RegisterGatewayCommand() requires a GatewayCommandDesc with command line usage, help, and a function to parse command line and create a `Gateway`. (Because I don't want to call `gatewayMain` in each gateway implementation)
2. gateway-main.go handles actual command action, then call `Gateway:NewObjectLayer()` to fully bring up a gateway layer.

cmd/gateway-main.go
```diff
@@ -31,212 +31,61 @@ import (
...
 
-const (
-	azureBackend gatewayBackend = "azure"
-	s3Backend    gatewayBackend = "s3"
-	gcsBackend   gatewayBackend = "gcs"
-	b2Backend    gatewayBackend = "b2"
-	// Add more backends here.
-)
+// GatewayCommandDesc is descriptor for gateway command.
+type GatewayCommandDesc struct {
```
balamurugana: Other than `cli` package, this information is nowhere required.  You could directly pass `cli.Command{}` from each gateway to `MustRegisterGatewayCommand()` and `RegisterGatewayCommand()`.

balamurugana:
> Can you have a look latest commit, though I'm afraid it's not very feasible, because:
> 
>     RegisterGatewayCommand() requires a GatewayCommandDesc with command line usage, help, and a function to parse command line and create a Gateway. (Because I don't want to call gatewayMain in each gateway implementation)

1. `GatewayCommandDesc` is not required.  See my above comment.
2. `gatewayMain()` is a common code to know how to start a gateway.  Calling `startGateway()` in each gateway is correct

>     gateway-main.go handles actual command action, then call Gateway:NewObjectLayer() to fully bring up a gateway layer.

Note that all gateway-main and gateway-*.go are same package.  Its not required to visualize as separate layer.

timonwong:
@balamurugana Done, just check the latest commit

harshavardhana (approved):

balamurugana (approved):
LGTM

PR #4802, opened 2017-08-09
Sia Integration

dvstate:
## Description
This code change adds a Sia gateway to Minio server, allowing data to be stored on the Sia network using Minio as a frontend. It changes nothing functional with the existing Minio codebase. It simply adds an additional gateway to allow users to store content on Sia.

Documentation for using the Sia gateway, which can be incorporated into Minio's documentation, can be found here:
https://github.com/dvstate/minio/blob/master/README.md

## Motivation and Context
To allow Minio users to utilize Sia for inexpensive backend storage.

## How Has This Been Tested?
New methods were individually tested as they were written.
Basic Go tests have been written for some of the Sia functionality.
Live servers using the Sia gateway have been operational for a couple weeks now, and all reported bugs, issues, and annoyances have been addressed.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #5114, opened 2017-10-25
Implement sia gateway

dvstate:
## Description
This PR is a continuation of the first "Sia Integration" PR, with the commit history cleaned up. Adds a gateway for using Sia as a backend storage provider.

## Motivation and Context
Adds support for Sia as a backend storage provider.

## How Has This Been Tested?
Light testing by myself and various other Sia community members.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/gateway-sia.go
+++ b/cmd/gateway-sia.go
@@ -1,0 +1,524 @@
+/*
+ * Minio Cloud Storage, (C) 2017 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cmd
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
+	"time"
+
+	"github.com/minio/cli"
+	"github.com/minio/minio-go/pkg/set"
+	"github.com/minio/minio/pkg/hash"
+)
+
+const (
+	siaBackend = "sia"
+)
+
+type siaObjects struct {
+	gatewayUnsupported
+	Address  string // Address and port of Sia Daemon.
+	TempDir  string // Temporary storage location for file transfers.
+	RootDir  string // Root directory to store files on Sia.
+	password string // Sia password for uploading content in authenticated manner.
+}
+
+func init() {
+	const siaGatewayTemplate = `NAME:
+  {{.HelpName}} - {{.Usage}}
+
+USAGE:
+  {{.HelpName}} {{if .VisibleFlags}}[FLAGS]{{end}} [SIA_DAEMON_ADDR]
+{{if .VisibleFlags}}
+FLAGS:
+  {{range .VisibleFlags}}{{.}}
+  {{end}}{{end}}
+ENVIRONMENT VARIABLES: (Default values in parenthesis)
+  ACCESS:
+     MINIO_ACCESS_KEY: Custom access key (Do not reuse same access keys on all instances)
+     MINIO_SECRET_KEY: Custom secret key (Do not reuse same secret keys on all instances)
+
+  SIA_TEMP_DIR:        The name of the local Sia temporary storage directory. (.sia_temp)
+  SIA_API_PASSWORD:    API password for Sia daemon. (default is empty)
+
+EXAMPLES:
+  1. Start minio gateway server for Sia backend.
+      $ {{.HelpName}}
+
+`
+
+	MustRegisterGatewayCommand(cli.Command{
+		Name:               siaBackend,
+		Usage:              "Sia Decentralized Cloud.",
+		Action:             siaGatewayMain,
+		CustomHelpTemplate: siaGatewayTemplate,
+		Flags:              append(serverFlags, globalFlags...),
+		HideHelpCommand:    true,
+	})
+}
+
+// Handler for 'minio gateway sia' command line.
+func siaGatewayMain(ctx *cli.Context) {
+	// Validate gateway arguments.
+	host := ctx.Args().First()
+	// Validate gateway arguments.
+	fatalIf(validateGatewayArguments(ctx.GlobalString("address"), host), "Invalid argument")
+
+	startGateway(ctx, &SiaGateway{host})
+}
+
+// SiaGateway implements Gateway.
+type SiaGateway struct {
+	host string // Sia daemon host address
+}
+
+// Name implements Gateway interface.
+func (g *SiaGateway) Name() string {
+	return siaBackend
+}
+
+// NewGatewayLayer returns b2 gateway layer, implements GatewayLayer interface to
+// talk to B2 remote backend.
+func (g *SiaGateway) NewGatewayLayer() (GatewayLayer, error) {
+	log.Println(colorYellow("\n               *** Warning: Not Ready for Production ***"))
+	return newSiaGatewayLayer(g.host)
+}
+
+// non2xx returns true for non-success HTTP status codes.
+func non2xx(code int) bool {
+	return code < 200 || code > 299
+}
+
+// decodeError returns the api.Error from a API response. This method should
+// only be called if the response's status code is non-2xx. The error returned
+// may not be of type api.Error in the event of an error unmarshalling the
+// JSON.
+type siaError struct {
+	// Message describes the error in English. Typically it is set to
+	// `err.Error()`. This field is required.
+	Message string `json:"message"`
+}
+
+func (s siaError) Error() string {
+	return s.Message
+}
+
+func decodeError(resp *http.Response) error {
+	// Error is a type that is encoded as JSON and returned in an API response in
+	// the event of an error. Only the Message field is required. More fields may
+	// be added to this struct in the future for better error reporting.
+	var apiErr siaError
+	if err := json.NewDecoder(resp.Body).Decode(&apiErr); err != nil {
+		return err
+	}
+	return apiErr
+}
+
+// SiaMethodNotSupported - returned if call returned error.
+type SiaMethodNotSupported struct {
+	method string
+}
+
+func (s SiaMethodNotSupported) Error() string {
+	return fmt.Sprintf("API call not recognized: %s", s.method)
+}
+
+// apiGet wraps a GET request with a status code check, such that if the GET does
+// not return 2xx, the error will be read and returned. The response body is
+// not closed.
+func apiGet(addr, call, apiPassword string) (*http.Response, error) {
+	req, err := http.NewRequest("GET", "http://"+addr+call, nil)
+	if err != nil {
+		return nil, traceError(err)
+	}
+	req.Header.Set("User-Agent", "Sia-Agent")
+	if apiPassword != "" {
+		req.SetBasicAuth("", apiPassword)
+	}
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return nil, traceError(err)
+	}
+	if resp.StatusCode == http.StatusNotFound {
+		resp.Body.Close()
+		return nil, SiaMethodNotSupported{call}
+	}
+	if non2xx(resp.StatusCode) {
+		err := decodeError(resp)
+		resp.Body.Close()
+		return nil, err
+	}
+	return resp, nil
+}
+
+// apiPost wraps a POST request with a status code check, such that if the POST
+// does not return 2xx, the error will be read and returned. The response body
+// is not closed.
+func apiPost(addr, call, vals, apiPassword string) (*http.Response, error) {
+	req, err := http.NewRequest("POST", "http://"+addr+call, strings.NewReader(vals))
+	if err != nil {
+		return nil, err
+	}
+	req.Header.Set("User-Agent", "Sia-Agent")
+	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+	if apiPassword != "" {
+		req.SetBasicAuth("", apiPassword)
+	}
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return nil, err
+	}
+
+	if resp.StatusCode == http.StatusNotFound {
+		resp.Body.Close()
+		return nil, SiaMethodNotSupported{call}
+	}
+
+	if non2xx(resp.StatusCode) {
+		err := decodeError(resp)
+		resp.Body.Close()
+		return nil, err
+	}
+	return resp, nil
+}
+
+// post makes an API call and discards the response. An error is returned if
+// the response status is not 2xx.
+func post(addr, call, vals, apiPassword string) error {
+	resp, err := apiPost(addr, call, vals, apiPassword)
+	if err != nil {
+		return err
+	}
+	resp.Body.Close()
+	return nil
+}
+
+// list makes a lists all the uploaded files, decodes the json response.
+func list(addr string, apiPassword string, obj *renterFiles) error {
+	resp, err := apiGet(addr, "/renter/files", apiPassword)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode == http.StatusNoContent {
+		return errors.New("Expecting a response, but API returned status code 204 No Content")
+	}
+
+	return json.NewDecoder(resp.Body).Decode(obj)
+}
+
+// get makes an API call and discards the response. An error is returned if the
+// responsee status is not 2xx.
+func get(addr, call, apiPassword string) error {
+	resp, err := apiGet(addr, call, apiPassword)
+	if err != nil {
+		return err
+	}
+	resp.Body.Close()
+	return nil
+}
+
+// newSiaGatewayLayer returns Sia gatewaylayer
+func newSiaGatewayLayer(host string) (GatewayLayer, error) {
+	sia := &siaObjects{
+		Address: host,
+		// RootDir uses access key directly, provides partitioning for
+		// concurrent users talking to same sia daemon.
+		RootDir:  os.Getenv("MINIO_ACCESS_KEY"),
+		TempDir:  os.Getenv("SIA_TEMP_DIR"),
+		password: os.Getenv("SIA_API_PASSWORD"),
+	}
+
+	// If Address not provided on command line or ENV, default to:
+	if sia.Address == "" {
+		sia.Address = "127.0.0.1:9980"
+	}
+
+	// If local Sia temp directory not specified, default to:
+	if sia.TempDir == "" {
+		sia.TempDir = ".sia_temp"
+	}
+
+	var err error
+	sia.TempDir, err = filepath.Abs(sia.TempDir)
+	if err != nil {
+		return nil, err
+	}
+
+	// Create the temp directory with proper permissions.
+	// Ignore error when dir already exists.
+	if err = os.MkdirAll(sia.TempDir, 0700); err != nil {
+		return nil, err
+	}
+
+	log.Println(colorBlue("\nSia Gateway Configuration:"))
+	log.Println(colorBlue("  Sia Daemon API Address:") + colorBold(fmt.Sprintf(" %s\n", sia.Address)))
+	log.Println(colorBlue("  Sia Temp Directory:") + colorBold(fmt.Sprintf(" %s\n", sia.TempDir)))
+	return sia, nil
+}
+
+// Shutdown saves any gateway metadata to disk
+// if necessary and reload upon next restart.
+func (s *siaObjects) Shutdown() error {
+	return nil
+}
+
+// StorageInfo is not relevant to Sia backend.
+func (s *siaObjects) StorageInfo() (si StorageInfo) {
+	return si
+}
+
+// MakeBucket creates a new container on Sia backend.
+func (s *siaObjects) MakeBucketWithLocation(bucket, location string) error {
+	return nil
+}
+
+// GetBucketInfo gets bucket metadata.
+func (s *siaObjects) GetBucketInfo(bucket string) (bi BucketInfo, err error) {
+	// Until Sia support buckets/directories, must return here that all buckets exist.
+	bi.Name = bucket
+	return bi, nil
+}
+
+// ListBuckets will detect and return existing buckets on Sia.
+func (s *siaObjects) ListBuckets() (buckets []BucketInfo, err error) {
+	sObjs, serr := s.listRenterFiles("")
+	if serr != nil {
+		return buckets, serr
+	}
+
+	m := make(set.StringSet)
+
+	prefix := s.RootDir + "/"
+	for _, sObj := range sObjs {
+		if strings.HasPrefix(sObj.SiaPath, prefix) {
+			siaObj := strings.TrimPrefix(sObj.SiaPath, prefix)
+			idx := strings.Index(siaObj, "/")
+			if idx > 0 {
+				m.Add(siaObj[0:idx])
+			}
+		}
+	}
+
+	for _, bktName := range m.ToSlice() {
+		buckets = append(buckets, BucketInfo{
+			Name:    bktName,
+			Created: timeSentinel,
+		})
+	}
+
+	return buckets, nil
+}
+
+// DeleteBucket deletes a bucket on Sia.
+func (s *siaObjects) DeleteBucket(bucket string) error {
+	return nil
+}
+
+func (s *siaObjects) ListObjects(bucket string, prefix string, marker string, delimiter string, maxKeys int) (loi ListObjectsInfo, err error) {
+	siaObjs, siaErr := s.listRenterFiles(bucket)
+	if siaErr != nil {
+		return loi, siaErr
+	}
+
+	loi.IsTruncated = false
+	loi.NextMarker = ""
+
+	root := s.RootDir + "/"
+
+	// FIXME(harsha) - No paginated output supported for Sia backend right now, only prefix
+	// based filtering. Once list renter files API supports paginated output we can support
+	// paginated results here as well - until then Listing is an expensive operation.
+	for _, sObj := range siaObjs {
+		name := strings.TrimPrefix(sObj.SiaPath, pathJoin(root, bucket, "/"))
+		if strings.HasPrefix(name, prefix) {
+			loi.Objects = append(loi.Objects, ObjectInfo{
+				Bucket: bucket,
+				Name:   name,
+				Size:   int64(sObj.Filesize),
+				IsDir:  false,
+			})
+		}
+	}
+	return loi, nil
+}
+
+func (s *siaObjects) GetObject(bucket string, object string, startOffset int64, length int64, writer io.Writer) error {
+	if !isValidObjectName(object) {
+		return traceError(ObjectNameInvalid{bucket, object})
+	}
+
+	dstFile := pathJoin(s.TempDir, mustGetUUID())
+	defer fsRemoveFile(dstFile)
+
+	var siaObj = pathJoin(s.RootDir, bucket, object)
+	if err := get(s.Address, "/renter/download/"+siaObj+"?destination="+url.QueryEscape(dstFile), s.password); err != nil {
+		return err
+	}
+
+	reader, size, err := fsOpenFile(dstFile, startOffset)
+	if err != nil {
+		return toObjectErr(err, bucket, object)
+	}
+	defer reader.Close()
+
+	// For negative length we read everything.
+	if length < 0 {
+		length = size - startOffset
+	}
+
+	bufSize := int64(readSizeV1)
+	if bufSize > length {
+		bufSize = length
+	}
+
+	// Reply back invalid range if the input offset and length fall out of range.
+	if startOffset > size || startOffset+length > size {
+		return traceError(InvalidRange{startOffset, length, size})
+	}
+
+	// Allocate a staging buffer.
+	buf := make([]byte, int(bufSize))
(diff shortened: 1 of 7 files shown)


mtlynch:
It looks like it's expecting environment variables to be set and won't start with default values:

```bash
$ ./minio gateway sia
FATA[0000]                                               cause=Access and Secret keys should be set through ENVs for backend [sia] source=[gateway-main.go:422:gatewayMain()]
```

Edit: I can get it to work if I set the keys explicitly:

```
$ export MINIO_ACCESS_KEY=minioaccesskey
$ export MINIO_SECRET_KEY=miniosecretkey
$ ./minio gateway sia                 
```

Edit2: I was able to successfully upload ~10 files totaling ~1.5 GB. I downloaded one and it failed the first time. I was watching the download get to 100% in Sia-UI, but then this appeared in the Minio output:

```
ERRO[2098] Unable to write to client.                    cause=write tcp 10.0.0.202:9000->10.0.0.100:54010: i/o timeout source=[gateway-handlers.go:145:gatewayAPIHandlers.GetObjectHandler()]
```

My S3 client automatically retried and it worked second time around.

harshavardhana:
> Edit: I can get it to work if I set the keys explicitly:
> 
> $ export MINIO_ACCESS_KEY=minioaccesskey
> $ export MINIO_SECRET_KEY=miniosecretkey
> $ ./minio gateway sia

This is expected and should be set @mtlynch in-fact the password here should point to the password you would use with your Sia daemon.

mtlynch:
>This is expected and should be set @mtlynch in-fact the password here should point to the password you would use with your Sia daemon.

It's using these credentials for Minio's access key and secret key credentials (i.e., it accepts them via the Minio web frontend and via the S3 interface). My Sia instance is running passwordless (which is the default behavior), so Minio can't use them there. I think this documentation is incorrect:

```
ENVIRONMENT VARIABLES: (Default values in parenthesis)
  ACCESS:
     MINIO_ACCESS_KEY: Custom sia access key
     MINIO_SECRET_KEY: Sia password
```

ericflo:
I think the best thing would be to use SIA_API_PASSWORD, which is what Sia now uses for its own environment variable: https://github.com/NebulousLabs/Sia/pull/2371

Here's my patch for minio implementing this: https://github.com/dvstate/minio/pull/8/commits/29c869df2d7bb192c9d5037cfd5ed1a13787c6ad

harshavardhana:
> I think the best thing would be to use SIA_API_PASSWORD, which is what Sia now uses for its own environment variable: NebulousLabs/Sia#2371
> 
> Here's my patch for minio implementing this: dvstate/minio@29c869d

Do not add more envs which are SIA only,  just like other gateways we should have two credentials.  One is access key, secret key etc.  So secret key can be re-purposed and use it with the same value as set for the sia daemon. 

> It's using these credentials for Minio's access key and secret key credentials (i.e., it accepts them via the Minio web frontend and via the S3 interface). My Sia instance is running passwordless (which is the default behavior), so Minio can't use them there. I think this documentation is incorrect:
> 
> ENVIRONMENT VARIABLES: (Default values in parenthesis)
>   ACCESS:
>      MINIO_ACCESS_KEY: Custom sia access key
>      MINIO_SECRET_KEY: Sia password

This is correct @mtlynch we are using if the MINIO_SECRET_KEY is set for talking to sia daemon as API password, looks like if its set passwordless then it allows every request. It is expected that this secret key is set to the value what you set with Sia daemon.

ericflo:
The problem with this is that now you *must* use the same key Sia uses as the Minio key. I'm in a position where I'm running Sia nodes that I do not expose to users, but I do want to expose Minio to the users. I would not like them to know the password I'm using on the Sia daemon.

harshavardhana:
> The problem with this is that now you must use the same key Sia uses as the Minio key. I'm in a position where I'm running Sia nodes that I do not expose to users, but I do want to expose Minio to the users. I would not like them to know the password I'm using on the Sia daemon.

Are you saying that you would allow concurrent access to multiple users to same sia daemon? how are conflicts handled in terms of same object being uploaded by multiple minio servers?   There is no locking here so you are on your own if you deploy in that manner and might lead to overwriting each others objects even reading wrong content. 

I would suggest that you deploy each minio with a new sia daemon - which brings its own password. Gateway is designed to leverage requirements of the backend not to multiplex authentication.

ericflo:
No, concurrent access isn't the main goal - the problem is that users are paying with USD $. We fund their single Sia node with enough SiaCoins to begin using storage, but we can't let them access their Sia node, or it would make those coins fungible & would be legally tricky.

So we need to lock down access to the Sia instance, which we do by firewall rules but also by taking advantage of the secret API password stored in an environment variable. While the firewall rules are probably sufficient, I'd rather not leak the secret API info out to the user.

dvstate:
I'm not sure how ericflo has things configured, but my reasoning for the SIA_ROOT_DIR env was to provide a way to partition the storage space of a Sia node so that multiple Minio instances could use the same Sia daemon without conflict.

Such as:
/minio1/bucket/file.cpp
/minio1/bucket/file.h
/minio2/bucket/music.mp3
/minio2/bucket/movie.mov

And if you only want a Minio instance to access a specific partition (/minio1), providing the users of minio1 with the actual password for the backend daemon is a huge security risk.

ericflo:
@dvstate That is fantastic!

harshavardhana:
> No, concurrent access isn't the main goal - the problem is that users are paying with USD $. We fund their single Sia node with enough SiaCoins to begin using storage, but we can't let them access their Sia node, or it would make those coins fungible & would be legally tricky.
> 
> So we need to lock down access to the Sia instance, which we do by firewall rules but also by taking advantage of the secret API password stored in an environment variable. While the firewall rules are probably sufficient, I'd rather not leak the secret API info out to the user.

Now i see so if concurrency is not a problem and @dvstate has already provided a prefix based partition i see that these requirements can be fixed. 

So MINIO_ACCESS_KEY, SECRET_KEY should still be set to custom values on as many instances as necessary but can definitely talk to single sia daemon by using the partitioning technique.  But this has to be made explicit IMO.  If this is the best practice we should ask every SIA user of this gateway to have set this value appropriately.

We can make this a function of ACCESS_KEY or we can have a separate ENV (which is already there) but we need to ask user to set this otherwise we shouldn't start the gateway.  This is to ensure that they don't accidentally corrupt each other files unknowingly.

ericflo:
Sorry - I definitely didn't mean to start a bikeshed discussion here.  I'm happy to maintain a patch for my own purposes.

For clarification's sake, the common case is probably running a local Sia node and that defaults to not having an API password. However, when you tell Sia to listen on anything other than localhost, it *forces* you to put an API password. So anyone who wants to have a network between Sia and Minio will be running with an API password. It's not a common case to run it in a multitenant way. I'm not doing that now, but I'd like to do so in the future, so it's great to see that prefix concept being thought of.  

It doesn't seem like a bad idea to me, to force a prefix and password env var if there is a non-localhost domain name. But that's just my 2c - happy with whatever you all decide.

harshavardhana:
A PR is sent implementing this as well https://github.com/dvstate/minio/pull/10 // cc @dvstate

mannp:
We seem to have lost the SIA_DAEMON_ADDR / --api-addr env variable to point the gateway to a separate Sia docker rather than localhost.

Is that by design and can we direct another way?

> Sia Gateway Configuration:
> Sia Daemon API Address: **127.0.0.1:9980**
> Sia Temp Directory: /mnt/minio-data/sia_temp

Thanks in advance

Also things stop when starting up, at 'mkdir /mnt/minio-data/sia_temp' if the directory already exists rather than dealing with it.

> time="2017-10-28T14:36:23+01:00" level=fatal msg="Unable to initialize gateway layer" cause="mkdir /mnt/minio-data/sia_temp: file exists" source="[gateway-main.go:445:gatewayMain()]"

harshavardhana:
Sia addr is provided by `minio gateway sia <sia_daemon_addr>` . The temp issue should be fixed.

mannp:
So got things going with a number of things to point out, but won't swamp things while its still in development :)

mannp:
Not sure where else to log, so adding here ...

restic (https://github.com/restic/restic - @fd0) to minio create repository gives the following error;

create key in backend at s3:http://192.168.1.99:9001/restic failed: client.PutObject: A header you provided implies functionality that is not implemented

> time="2017-11-01T21:00:03Z" level=error msg="Unable to initiate new multipart upload id." cause="Not Implemented" source="[object-handlers.go:640:objectAPIHandlers.NewMultipartUploadHandler()]" stack="gateway-unsupported.go:35:gatewayUnsupported.NewMultipartUpload <autogenerated>:1:(*siaObjects).NewMultipartUpload object-handlers.go:638:objectAPIHandlers.NewMultipartUploadHandler api-router.go:54:(objectAPIHandlers).NewMultipartUploadHandler-fm"

It works perfectly and very fast with standalone minio.

harshavardhana:
It's because multipart uploads are not yet supported .

mannp:
OK thanks. It didn't give the error in the previous v2 version. Will that be supported in the merged PR?

harshavardhana:
> OK thanks. It didn't give the error in the previous v2 version. Will that be supported in the merged PR?

There is a plan to re-implement it, the reason it was removed was to cleanup the usage of multiple layers. The real pain point is the Sia APIs are primitive unlike cloud storage backends like azure, backblaze.  So we will have to maintain the state locally and munge the data to look like a single contiguous entity and you will not be able to load balance the state across independent instances. 

This is a sort of a complex functionality, ideally if Sia has APIs which can be leveraged - this doesn't seem to be case at the moment.

mannp:
@harshavardhana, thanks for the explanation; understood :)

mtlynch:
Thanks for the update @dvstate!

I tested the latest version and I notice that it's prepending the access key to the Sia path. Is that intended behavior?

image

Aside from that, I'm uploading and downloading successfully.

Here are the latest build steps:

```bash
go get github.com/minio/minio/...
cd $GOPATH/src/github.com/minio/minio/
git remote add dvstate https://github.com/dvstate/minio.git
git checkout -b sia_integration_v2
git reset --hard HEAD~50 # Back up a bit to avoid merge conflicts
git pull dvstate sia_integration_v2
make
export MINIO_ACCESS_KEY=minioaccesskey # Change to a random value
export MINIO_SECRET_KEY=miniosecretkey # Change to a random value
./minio gateway sia
```

mannp:
@mtlynch  can you perhaps add what you've had success with please? What is your basis for testing here? mc uploads, s3 clients, minio gui uploads as I have a number of issues.

@harshavardhana does minio request a base set of tests before merging, in order that minio's supported s3 clients are supported etc? Is that something we should be waiting for here?

1) The upload dialog in the minio gui remains and stays at 0% even though the file is uploaded, it never updates or disappears.

2) Once uploaded all file times are set to Jan 01.

3) I am using a remote sia node and have to map the directories that minio and sia share, so that files do get seen. eg /mnt/minio-data/ which is where .sia_temp is set.

What is the designed config for this scenario, so that I can confirm that mine is correct and not causing further issues?

mtlynch:
>can you perhaps add what you've had success with please? What is your basis for testing here? mc uploads, s3 clients, minio gui uploads as I have a number of issues.

I'm testing through the Minio web UI and with a third-party S3 client ([Cloudberry Backup Pro](https://www.cloudberrylab.com/backup/windows/amazon-s3.aspx)).

>Once uploaded all file times are set to Jan 01.

I see this behavior as well.

>The upload dialog in the minio gui remains and stays at 0% even though the file is uploaded, it never updates or disappears.

>I am using a remote sia node and have to map the directories that minio and sia share, so that files do get seen. eg /mnt/minio-data/ which is where .sia_temp is set.

I recommend using a local Sia node. The issue you're seeing sounds like Minio isn't able to upload successfully to Sia.

The remote Sia node is going to introduce a lot more complexity into testing. I'd much rather support a local Sia node now and iterate on the remote node scenario after the initial PR.

mannp:
Ok thanks.

Well when I say a remote sia node, its sia, minio and nextcloud running in separate dockers on the same server, connected via shared directories, which is a requirement of the nextcloud config.

Based on your setup for your synology you posted.

Perhaps but the full file is present in the .sia_temp directory and is queued for upload, so it seems to have pushed the file to minio correctly?

I will download a free trial of cloudberry and see if it works for me then.

harshavardhana:
> @harshavardhana does minio request a base set of tests before merging, in order that minio's supported s3 clients are supported etc? Is that something we should be waiting for here?
> 

That is the case for most gateways, but with Sia that will not happen since we can never make this fully S3 compatible.  The reason to wait here is that its not yet ready to be merged, there are few things that needs to be fixed as well such as multipart support etc.  Without that this gateway is pretty much useless for any real S3 usage. 

Apart from that we need some basic tests to pass for Sia.  Just so everyone knows that this gateway will not be fully S3 compatible and perhaps it never will be unless backend provides APIs which provides the flexibility to ensure this compatibility.

The basic intention of Gateway is following

 - Share nothing architecture, IPC of any sort should be implemented leveraging the remote backend.
 - Simple translations if possible, leverage existing APIs. If existing APIs do not provide equivalent functionality say so by returning "NotImplemented"
 - No complex state management on the minio's end - this is to ensure that you don't have a data loss.

With these guidelines in mind Sia gateway was re-implemented to ensure that we don't carry a lot of state, to some extent it is possible now and we just have to finish the multipart support.  So lets say as of now if you are expecting such as `s3cmd` will work with Sia gateway with all its options, it will not work.  May be some options will never work.

> The upload dialog in the minio gui remains and stays at 0% even though the file is uploaded, it never updates or disappears.
> 

This is perhaps the problem of how the uploading indeed works, there is a double upload so if the upload to Sia backend is slow then it will not be completed and perhaps it just causes a bug on the minio browser.  There was discussion with @dvstate to tell the client that it uploaded correctly and return GET only when it really finds the object. Basically to have No strict write to read consistency.  @dvstate thinks it will be confusing i do agree it is confusing but there is nothing much we can do here, if we really need to make PutObject look atomic we have to only return true if the object was successfully uploaded to Sia backend.

> Once uploaded all file times are set to Jan 01.
> 

This is expected Sia backend doesn't preserve any time, so we cannot either. Unless you add some sort of namespace database which increases the complexity and defeats the fundamentals of Gateway. 

> I am using a remote sia node and have to map the directories that minio and sia share, so that files do get seen. eg /mnt/minio-data/ which is where .sia_temp is set.

Is this a question? or suggestion?  I am sorry i couldn't follow. 

> I tested the latest version and I notice that it's prepending the access key to the Sia path. Is that intended behavior?

@mtlynch this is intended to avoid conflicts, say when you run multiple minio gateways and you need to separate each such user.

harshavardhana:
> Once uploaded all file times are set to Jan 01.
> This is expected Sia backend doesn't preserve any time, so we cannot either. Unless you add some sort of namespace database which increases the complexity and defeats the fundamentals of Gateway.

If you really need all of this, there is a possibility of encoding all of this information into the path-name of an object itself for example like, this requires a custom backend format something like this. 

```
<root_dir>/<bucket>/object-name/part.[1..10000]
<multiparts are uploaded as individual parts and are updated as part of sia.json, to read from>

<root_dir>/<bucket>/object-name/sia.json
<sia.json carries all the metadata associated with the object-name>
```

If you see in this scenario we are still not implementing anything locally, instead leveraging the backend itself to implement a custom format on it.  The major caveat in this is you need to use minio to copy the data out in future and there will be no easier way to reconstruct this data back unless you have a tooling which understands this format. My personal choice is to not go this route because it causes a vendor lock-in and we intend to keep minio's role here as minimal as possible. 

// cc @dvstate

mannp:
@harshavardhana Thanks very much for taking the time for a full and informative post. I understand better where things are at now.

Personally I do not need date stamps on the minio gui as I will mainly be using it with restic backup which I would envisage would not mind about the dates either.

> The reason to wait here is that its not yet ready to be merged, there are few things that needs to be fixed as well such as multipart support etc. Without that this gateway is pretty much useless for any real S3 usage.

Restic would though need multipart support with the error I reported above.

> Is this a question? or suggestion? I am sorry i couldn't follow.

Its my config setup and was just to state that this may not be part of the design criteria for the gateway. That said being able to direct the gateway to a remote address for sia would suggest that I should be able to use a remote sia node.

> Sia addr is provided by minio gateway sia <sia_daemon_addr> .

mtlynch:
@harshavardhana would it be possible to establish what the acceptance criteria is for this PR? My understanding was that the only request from #4802 was to remove the caching layer. The caching layer is gone, but now it feels like we're moving the goalposts again with the requirements to implement additional tests and support multipart.

Is multipart a firm requirement or can that wait until a subsequent PR? I feel like the broader we increase the scope, the harder it is for the minio devs to review because this is already a giant PR. It also keeps forcing @dvstate to keep fixing merge conflicts as the `master` branch changes out from under him.

harshavardhana:
> @harshavardhana would it be possible to establish what the acceptance criteria is for this PR? My understanding was that the only request from #4802 was to remove the caching layer. The caching layer is gone, but now it feels like we're moving the goalposts again with the requirements to implement additional tests and support multipart.

The acceptance criteria is it implements basic S3 functionalities which includes multipart, without that it is pretty much useless. I am working on them right now, should be finished by next week. 

> Is multipart a firm requirement or can that wait until a subsequent PR? I feel like the broader we increase the scope, the harder it is for the minio devs to review because this is already a giant PR. It also keeps forcing @dvstate to keep fixing merge conflicts as the master branch changes out from under him.

There is one more change that is being brought in to make the gateway backends more pluggable.  Such that this code never needs to reside in Minio and can be pulled in as needed by users when they build the binaries.

mtlynch:
>The acceptance criteria is it implements basic S3 functionalities which includes multipart, without that it is pretty much useless. I am working on them right now, should be finished by next week.

It's obviously your decision, but Sia support without multipart would still be very useful to me as a Minio user. I'm using dvstate@'s branch now to back up my DVD collection with Sia as a backend. I just disable the multipart option in my client and it works fine. `s3cmd` also supports the `--disable-multipart` switch so that the tool would still work even without multipart in Sia.

mannp:
Minio is an open source object storage server with Amazon S3 compatible API. 

> as well such as multipart support etc. Without that this gateway is pretty much useless for any real S3 usage.

Its inevitable that the minio code base will change when a PR is being worked on, on and off over the past 5 months.

A clear understanding of the outstanding items may help, but they also need completing in a timely manner, otherwise the same issue will be the case again.

harshavardhana:
> Its inevitable that the minio code base will change when a PR is being worked on, on and off over the past 5 months.
> 
> A clear understanding of the outstanding may help, but they also need completing in a timely manner, otherwise the same issue will be the case again.

This is all that needs to be implemented but agreed that it doesn't need to wait though and can be merged as a separate PR later. Now if @dvstate can rebase this with latest master we can do a final review and merge.

mannp:
> This is all that needs to be implemented but agreed that it doesn't need to wait though and can be merged as a separate PR later. Now if @dvstate can rebase this with latest master we can do a final review and merge.

I have waited 5 months for this, so another week for multipart would be fine by me, as dvstates limited availability has been part of the reason for the 5 months and once merged I can imagine that time will be further restricted and we will be left with a partly complete solution.

@harshavardhana this all said it will be your team that has to deal with questions and queries regarding any sia related minio issues, so its obviously your decision where its at before merging.

@mtlynch multipart was supported in the earlier release of this gateway, so it is not added fuctionaility, expanding scope, but functionaility that has actually been removed.

harshavardhana (changes requested):
LGTM @dvstate almost there finally.

One last remaining change is would you mind writing a doc about this in docs/gateway/sia.md 

This would explain scenarios on how to run Sia gateway and an example of using `s3cmd` like tools to configure it.

mannp:
@harshavardhana & @dvstate nice job and thanks for your monumental efforts in getting this ready for us eager users :) 

It really feels like its close now 👍

cmd/gateway-sia.go
```diff
@@ -20,6 +20,9 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/minio/cli"
+	"github.com/minio/minio-go/pkg/set"
+	"github.com/minio/minio/pkg/hash"
```
harshavardhana: Keep the imports in their previous location here. @dvstate you should run `goimports` on gateway-sia.go for proper styling.

harshavardhana (changes requested):

mtlynch:
Are there any outstanding changes required before this can be merged?

harshavardhana:
There are no outstanding changes we are just testing right now on an ec2 instance.

harshavardhana (dismissed):
LGTM tested and working . Can you take a look @krisis ?

dvstate:
Just to update everyone that may be following this PR, I setup an Amazon EC2 instance running a live Sia Daemon 1.3.0 and this Minio PR, and gave the Minio team full access for testing. We've made additional improvements on both sides. The Sia gateway has been tested with the Minio browser, the mc command line tool, and the official s3cmd tool. Basic object operations such as put,get, and delete have been verified to work using all 3 tools.

However, there have been some performance related issues with the Sia Daemon that is beyond our control. Harshavardhana observed that uploading many files at once can cause the daemon to bog down and even lock up. This may be due to the daemon's HTTP interface performing work synchronously.

The Sia team indicated that they will be releasing 1.3.1 of the daemon within a couple weeks, and that numerous improvements have been made. We're hoping that 1.3.1 will eliminate all remaining issues.

mannp:
Looks like some further progress, which is great to see.

Not clear if you plan on merging now, with a note on the outstanding issue, then doing an additional PR for 1.3.1 compatibility or waiting for 1.3.1 to do the merge with further testing?

harshavardhana:
We are not waiting for that once @krisis approves we will be merging this PR

cmd/gateway-sia.go
```diff
@@ -0,0 +1,508 @@
...
+
+USAGE:
+  {{.HelpName}} {{if .VisibleFlags}}[FLAGS]{{end}} [SIA_DAEMON_ADDR]
+{{if .VisibleFlags}}
+FLAGS:
+  {{range .VisibleFlags}}{{.}}
+  {{end}}{{end}}
+ENVIRONMENT VARIABLES: (Default values in parenthesis)
+  ACCESS:
+     MINIO_ACCESS_KEY: Custom access key (Do not re-use same access keys on all instances)
```
krisis: s/re-use/reuse

cmd/gateway-sia.go
```diff
@@ -0,0 +1,508 @@
...
+     MINIO_SECRET_KEY: Custom secret key (Do not re-use same secret keys on all instances)
```
krisis: s/re-use/reuse

cmd/gateway-sia.go
```diff
@@ -0,0 +1,508 @@
...
+		req.SetBasicAuth("", apiPassword)
+	}
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return nil, err
+	}
+
+	if resp.StatusCode == http.StatusNotFound {
+		resp.Body.Close()
+		return nil, errors.New("API call not recognized: " + call)
```
krisis: errors.New("API call not recognised") looks like a commonly occurring error. How about,
```go
type SiaMethodNotSupported {
  method string
}

func (s SiaMethodNotSupported) Error() string {
  return fmt.Sprintf("API call not recognized: %s", s.method)
}
```

So the above return statement would be,
 ```go
return nil, SiaMethodNotSupported{method: call}
```

File: cmd/gateway-sia.go [hunk 5]
```diff
@@ -0,0 +1,508 @@
...
+	resp, err := apiPost(addr, call, vals, apiPassword)
+	if err != nil {
+		return err
+	}
+	resp.Body.Close()
+	return nil
+}
+
+// list makes a lists all the uploaded files, decodes the json response.
+func list(addr string, apiPassword string, obj interface{}) error {
```
krisis: why aren't we defining a type for `obj`?

cmd/gateway-sia.go
```diff
@@ -0,0 +1,508 @@
...
+
+	var err error
+	sia.TempDir, err = filepath.Abs(sia.TempDir)
+	if err != nil {
+		return nil, err
+	}
+
+	// Create the temp directory with proper permissions.
+	// Ignore error when dir already exists.
+	_ = os.Mkdir(sia.TempDir, 0700)
```
krisis: we should check if `os.IsExist(err)` is true before we ignore the error.

cmd/gateway-sia.go
```diff
@@ -0,0 +1,508 @@
...
+			Created: timeSentinel,
+		})
+	}
+
+	return buckets, nil
+}
+
+// DeleteBucket deletes a bucket on Sia.
+func (s *siaObjects) DeleteBucket(bucket string) error {
+	return nil
```
krisis, line 339: Shouldn't we return error saying API method is not implemented?

cmd/gateway-sia.go
```diff
@@ -0,0 +1,508 @@
...
+		return loi, siaErr
+	}
+
+	loi.IsTruncated = false
+	loi.NextMarker = ""
+
+	root := s.RootDir + "/"
+
+	for _, sObj := range siaObjs {
+		name := strings.TrimPrefix(sObj.SiaPath, pathJoin(root, bucket, "/"))
```
krisis: don't we need to skip objects that occur before (and including) marker?

cmd/gateway-sia.go
```diff
@@ -0,0 +1,508 @@
...
+	}
+
+	reader, size, err := fsOpenFile(dstFile, startOffset)
+	if err != nil {
+		return toObjectErr(err, bucket, object)
+	}
+	defer reader.Close()
+
+	bufSize := int64(readSizeV1)
+	if length > 0 && bufSize > length {
```
krisis: length  > 0 check seems unnecessary

docs/gateway/sia.md
```diff
@@ -0,0 +1,61 @@
...
+- Bucket policies are not currently supported.
+
+Other limitations:
+
+- Bucket notification APIs are not supported.
+
+## Explore Further
+- [`mc` command-line interface](https://docs.minio.io/docs/minio-client-quickstart-guide)
+- [`aws` command-line interface](https://docs.minio.io/docs/aws-cli-with-minio)
+- [`minfs` filesystem interface](http://docs.minio.io/docs/minfs-quickstart-guide)
```
krisis: is a link to minfs required, since it is deprecated?

docs/gateway/sia.md
```diff
@@ -0,0 +1,61 @@
...
+`SIA_API_PASSWORD` | The API password required to access the Sia daemon, if needed. | (empty)
+
+### Running Minio with Sia Gateway
+```
+export MINIO_ACCESS_KEY=minioaccesskey
+export MINIO_SECRET_KEY=miniosecretkey
+minio gateway sia [SIA_DAEMON_ADDRESS]
+```
+The [SIA_DAEMON_ADDRESS] is optional, and it defaults to "127.0.0.1:9980".
+Access information should then be presented on-screen. To connect to the server and upload files using your web browser, open a web browser and point it to the address displayed for "Browser Access." Then log in using the "AccessKey" and "SecretKey" that are also displayed on-screen. You should then be able to create buckets (folders) and upload files.
```
krisis, line 41: How does one create a bucket when `MakeBucketWithLocation` doesn't do anything in the Sia backend?

cmd/gateway-sia.go
```diff
@@ -0,0 +1,508 @@
...
+	return nil
+}
+
+// StorageInfo is not relevant to Sia backend.
+func (s *siaObjects) StorageInfo() (si StorageInfo) {
+	return si
+}
+
+// MakeBucket creates a new container on Sia backend.
+func (s *siaObjects) MakeBucketWithLocation(bucket, location string) error {
```
krisis, line 296: shouldn't this return NotImplemented?

krisis (changes requested):

docs/gateway/sia.md
harshavardhana: MakeBucket is fake on Sia

cmd/gateway-sia.go
harshavardhana: MakeBucket is fake.

cmd/gateway-sia.go
harshavardhana: Done

cmd/gateway-sia.go
harshavardhana: Done

cmd/gateway-sia.go
harshavardhana: done

cmd/gateway-sia.go
harshavardhana: done

cmd/gateway-sia.go
harshavardhana: done using os.MkdirAll

cmd/gateway-sia.go
harshavardhana: DeleteBucket is fake as well.

cmd/gateway-sia.go
harshavardhana: No such things are not honored, we just list everything right now. So if Sia has 10,000 objects all of them will be listed simultaneously. Future optimizations will be done when backend supports paginated output.

cmd/gateway-sia.go
harshavardhana: Done

docs/gateway/sia.md
harshavardhana: Done

harshavardhana:
@krisis PR is updated, PTAL

harshavardhana (dismissed):
LGTM

krisis (approved):

harshavardhana (approved):

DidgeriDude:
Somehow I get this error.

`./minio gateway sia
`
`‘sia’ is not a minio sub-command. See ‘minio --help’.`

I checked and my version is 2017-11-22T19:55:46Z. I installed from the instructions on the main page.

nitisht:
@DidgeriDude this PR was merged after the release, so its not available in the release. If you want, you can build from master and give it a try.

mannp:
@DidgeriDude it also appears to be available in the edge release of the minio docker.

harshavardhana:
> @DidgeriDude it also appears to be available in the edge release of the minio docker.

Each commit merged to master is built into a docker image as `minio/minio:edge`

PR #5115, opened 2017-10-25
move credentials as separate package

balamurugana:
## Types of changes

## Checklist:


cmd/lock-instrument.go
```diff
@@ -258,13 +257,5 @@ func (n *nsLockMap) deleteLockInfoEntryForOps(param nsParam, opsID string) error
...
-	const opsIDLen = 16
-	opsIDBytes := make([]byte, opsIDLen)
-	if _, err := rand.Read(opsIDBytes); err != nil {
-		panic(err)
-	}
-	for i := 0; i < opsIDLen; i++ {
-		opsIDBytes[i] = alphaNumericTable[opsIDBytes[i]%alphaNumericTableLen]
-	}
-	return string(opsIDBytes)
+	return mustGetUUID()
```
donatello, line 260: This seems to be unrelated to moving credentials as a separate package - could we note it in the commit message?

donatello (approved):
LGTM - there seems to be a unrelated change though that we should note in the commit message.

cmd/lock-instrument.go
balamurugana: No. Previously its using `alphaNumericTable` from credential.go.  As its moved to separate package in this PR, its required to do this change as well.

harshavardhana (approved):

PR #5116, opened 2017-10-25
Support for pcf user-agent when querying for updates.

krishnasrinivas:
## Description

Minio PCF startup script should set MINIO_PCF=1 in the env var which will be used for constructing user-agent.

## How Has This Been Tested?

No straight forward way of testing this fix.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/update-main.go
+++ b/cmd/update-main.go
@@ -169,6 +169,20 @@ func IsKubernetes() bool {
 	return os.Getenv("KUBERNETES_SERVICE_HOST") != ""
 }
 
+// IsBOSH returns true if minio is deployed from a bosh package
+func IsBOSH() bool {
+	// "/var/vcap/bosh" exists in BOSH deployed instance.
+	_, err := os.Stat("/var/vcap/bosh")
+	if os.IsNotExist(err) {
+		return false
+	}
+
+	// Log error, as we will not propagate it to caller
+	errorIf(err, "Error in BOSH check.")
+
+	return err == nil
+}
+
 // Minio Helm chart uses DownwardAPIFile to write pod label info to /podinfo/labels
 // More info: https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/#store-pod-fields
 // Check if this is Helm package installation and report helm chart version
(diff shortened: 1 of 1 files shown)


cmd/update-main.go
```diff
@@ -169,6 +169,12 @@ func IsKubernetes() bool {
 	return os.Getenv("KUBERNETES_SERVICE_HOST") != ""
 }
 
+// IsPCF returns true if minio is running on Pivotal Cloud Foundry.
+func IsPCF() bool {
+	// Minio PCF start-up script should set MINIO_PCF.
+	return os.Getenv("MINIO_PCF") != ""
```
harshavardhana: Is this is the only way to know? isn't there something PCF specific so that we don't have to add?

krishnasrinivas: These are the env vars set:
```
MONIT_DATE=Thu, 26 Oct 2017 00:16:53 +0000
MONIT_HOST=localhost
PATH=/bin:/usr/bin:/sbin:/usr/sbin
PWD=/etc/sv/monit
MONIT_PROCESS_PID=0
MONIT_EVENT=Started
MONIT_PROCESS_MEMORY=0
SHLVL=1
MONIT_PROCESS_CPU_PERCENT=0
MONIT_SERVICE=minio-server
MONIT_PROCESS_CHILDREN=0
MONIT_DESCRIPTION=Started
_=/usr/bin/printenv
```

But not sure if we should rely on the `MONIT_*` vars as I could not find documentation on these vars. Safer to depend on a var of our own.

harshavardhana: https://mmonit.com/monit/ is actually a different project, looks like there is no other way.

nitisht (dismissed):
LGTM

cmd/update-main.go
harshavardhana: At least can we have a more MINIO_PIVOTAL_CLOUD_FOUNDRY - an expanded name to avoid any future confusion ?

harshavardhana (changes requested):
Minor comment env namspace can be protected and made more explanatory .. also avoiding any future conflicts.

harshavardhana (approved):

balamurugana (approved):
LGTM

PR #5119, opened 2017-10-26
Allow event notifications to work without region

harshavardhana:
## Description
Allow event notifications to work without region

## Motivation and Context
Refer #5101

## How Has This Been Tested?
Manually using `mc event add` 

## Types of changes

## Checklist:

### DIFF
--- a/cmd/bucket-notification-handlers.go
+++ b/cmd/bucket-notification-handlers.go
@@ -150,6 +150,12 @@ func (api objectAPIHandlers) PutBucketNotificationHandler(w http.ResponseWriter,
 		return
 	}
 
+	// Convert the incoming ARNs properly to the GetRegion().
+	for i, queueConfig := range notificationCfg.QueueConfigs {
+		queueConfig.QueueARN = unmarshalSqsARN(queueConfig.QueueARN).String()
+		notificationCfg.QueueConfigs[i] = queueConfig
+	}
+
 	// Put bucket notification config.
 	err = PutBucketNotificationConfig(bucket, &notificationCfg, objectAPI)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


cmd/event-notifier.go
```diff
@@ -291,7 +292,7 @@ func eventNotifyForBucketNotifications(eventType, objectName, bucketName string,
 		eventMatch := eventMatch(eventType, qConfig.Events)
 		ruleMatch := filterRuleMatch(objectName, qConfig.Filter.Key.FilterRules)
 		if eventMatch && ruleMatch {
-			targetLog := globalEventNotifier.GetExternalTarget(qConfig.QueueARN)
+			targetLog := globalEventNotifier.GetExternalTarget(unmarshalSqsARN(qConfig.QueueARN).String())
```
vadmeste: @harshavardhana, I think it is better to update qConfig.QueueARN just after ARN validation, I think it would be cleaner that way.

cmd/bucket-notification-handlers.go
```diff
@@ -144,6 +144,12 @@ func (api objectAPIHandlers) PutBucketNotificationHandler(w http.ResponseWriter,
 		return
 	} // Successfully marshalled notification configuration.
 
+	// Convert the incoming ARNs properly to the GetRegion().
```
krisis: This comment is not clear. How about the following?
`// Set server's region as the ARN region for all queue notifications configured`

harshavardhana: yes agreed @krisis

krisis (approved):

vadmeste (approved):
LGTM & tested

PR #5138, opened 2017-11-02
Make GCS multipart upload failures S3-compatible

### LINKED ISSUE #206: aws-sdk-php `testMultipartUploadFailure` fails on GCS gateway
The test fails on GCS gateway since part size validation is missing.

```
Running aws-sdk-php tests ... FAILED in 49 seconds
{
  "name": "aws-sdk-php",
  "function": "uploadPart ( array $params = [] )",
  "args": {
    "Bucket": "aws-sdk-php-bucket-14203",
    "Object": "obj1"
  },
  "duration": "6079",
  "status": "FAIL",
  "error": "Expected completeMultipartUpload to fail with EntityTooSmall"
}
```

krisis:
## Description

This change translates GCS error responses from multipart upload failures to appropriate AWS S3 compatible errors.

## Motivation and Context

Fixes minio/mint#206

## How Has This Been Tested?

Running mint aws-sdk-php tests locally

## Types of changes

## Checklist:

### DIFF
--- a/cmd/gateway-gcs.go
+++ b/cmd/gateway-gcs.go
@@ -192,11 +192,15 @@ func gcsToObjectError(err error, params ...string) error {
 
 	bucket := ""
 	object := ""
+	uploadID := ""
 	if len(params) >= 1 {
 		bucket = params[0]
 	}
 	if len(params) == 2 {
 		object = params[1]
+	}
+	if len(params) == 3 {
+		uploadID = params[2]
 	}
 
 	// in some cases just a plain error is being returned
@@ -208,9 +212,15 @@ func gcsToObjectError(err error, params ...string) error {
 		e.e = err
 		return e
 	case "storage: object doesn't exist":
-		err = ObjectNotFound{
-			Bucket: bucket,
-			Object: object,
+		if uploadID != "" {
+			err = InvalidUploadID{
+				UploadID: uploadID,
+			}
+		} else {
+			err = ObjectNotFound{
+				Bucket: bucket,
+				Object: object,
+			}
 		}
 		e.e = err
 		return e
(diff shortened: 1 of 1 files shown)


balamurugana (approved):
LGTM

cmd/gateway-gcs.go
```diff
@@ -1036,9 +1046,26 @@ func (l *gcsGateway) CompleteMultipartUpload(bucket string, key string, uploadID
 	}
 
 	var parts []*storage.ObjectHandle
-	for _, uploadedPart := range uploadedParts {
+	partSizes := make([]int64, len(uploadedParts))
+	for i, uploadedPart := range uploadedParts {
 		parts = append(parts, l.client.Bucket(bucket).Object(gcsMultipartDataName(uploadID,
 			uploadedPart.PartNumber, uploadedPart.ETag)))
+		partAttr, pErr := l.client.Bucket(bucket).Object(gcsMultipartDataName(uploadID, uploadedPart.PartNumber, uploadedPart.ETag)).Attrs(l.ctx)
```
krisis: > Object returns an ObjectHandle, which provides operations on the named object. This call does not perform any network operations.

A network call to fetch object attributes is made when `ObjectHandle.Attrs` method is called. So, there is only one call to the backend here.

cmd/gateway-gcs.go
```diff
@@ -192,12 +192,16 @@ func gcsToObjectError(err error, params ...string) error {
 
 	bucket := ""
 	object := ""
+	uploadID := ""
 	if len(params) >= 1 {
 		bucket = params[0]
 	}
 	if len(params) == 2 {
 		object = params[1]
```
krisis: `bucket` and `object` values aren't used when we return error of type `InvalidUpload`, which is the only place where we use `uploadID`. This is intentionally left as is.

donatello, line 200: This `if` block needs to be changed to `if len(params) >=2` or `object` will be empty when len(params) == 3.

cmd/gateway-gcs.go
donatello, line 1053: Looks like there is an extra backend call here - the `i`-th part is being fetched twice (before the call to `.Attrs()`). The result from the first call should be reused.

donatello (changes requested):

donatello (approved):

deekoder:
ignoring codecov as spurious failure.

PR #5157, opened 2017-11-08
Update logrus and use lowercase sirupsen/logrus import

jawnsy:
**NOTE: build is currently broken due to an API change between logrus 0.11 and 1.0.3**

logrus.Entry no longer has a Reader() function

## Description

Update to the latest version of logrus and use the lowercase form of the import (`github.com/sirupsen/logrus` vs `github.com/Sirupsen/logrus` as was done previously). The project now recommends using the lowercase form, so this change updates all references to the lowercase sirupsen/logrus.

* Updates to the latest sirupsen/logrus version using govendor (go get github.com/sirupsen/logrus, make pkg-update PKG=github.com/sirupsen/logrus)
* Updates references to `github.com/Sirupsen/logrus` to `github.com/sirupsen/logrus`
* Renames the directory in `vendor/` to the lowercase form

Signed-off-by: Jonathan Yu <jonathan.i.yu@gmail.com>

## Motivation and Context

When adding `github.com/minio/minio` as a dependency to `dep` in a project that imports logrus using the new form, the following message is logged by dep:

(22)  ✗   case-only variation in dependency on "github.com/Sirupsen/logrus"; "github.com/sirupsen/logrus" already established by:
(22)    (root)

The tool then iterates over all the old releases of Minio instead of using the latest one, but all the old releases have the same problem (since they, too, use an older version of logrus).

## How Has This Been Tested?

This hasn't been tested yet. I plan to run unit tests.

## Types of changes

## Checklist:


kannappanr (changes requested):
This PR is not needed. We are moving away from logrus for both logging and notification. We will be done with that soon.

PR #5168, opened 2017-11-10
replace SSE-C key derivation scheme

### LINKED ISSUE #5167: SSE-C key derivation is not invertible
The current SSE-C key derivation derives an unique encryption key per object from the client-provided key. The key derivation uses only the hash function SHA256 (and HMAC-SHA256). The key derivation is not invertible. That means that a client cannot change its key without changing the derived object encryption key. (The object encryption key is derived from the client key and a random value).

## Expected Behavior
The SSE-C key derivation must be invertible.

## Current Behavior
The SSE-C key derivation is not invertible.

## Possible Solution
Make SSE-C key derivation invertible.
That can be achieved by sealing/encrypting the object encryption key:
 - The object encryption key is derived once when the object is created
 - The object encryption key is stored in a sealed form as part of the object metadata  

## Context
In general a non-invertible key derivation is the default/standard approach to derive unique key material from a long-term / master key and the non-invertible property is not a problem but a feature.
However AWS-S3 allows users to change there SSE-C key by sending a COPY request to an SSE-C encrypted object where `copy-source` == `copy-destination`. This means that AWS will update just the metadata of the object and does not copy the object (which would be expensive).

Furthermore there are special headers for COPY requests to also encrypt the copied object.
See [https://docs.aws.amazon.com/en_en/AmazonS3/latest/dev/ServerSideEncryptionCustomerKeys.html](https://docs.aws.amazon.com/en_en/AmazonS3/latest/dev/ServerSideEncryptionCustomerKeys.html)  

That means that a client can do a SSE-C server-side COPY where source == destination but with different SSE-C keys. (That's the SSE-C key rotation 'feature'):
```
aws s3api copy-object \
> --endpoint-url https://s3.amazonaws.com \
> --copy-source aead/my-secret-diary \
> --bucket aead \
> --key my-secret-diary \
> --sse-customer-algorithm AES256 \
> --sse-customer-key MzJieXRlc2xvbmdzZWNyZXRrZXltdXN0cHJvdmlkZWQ= \
> --sse-customer-key-md5 7PpPLAK26ONlVUGOWlusfg== \
> --copy-source-sse-customer-algorithm AES256 \
> --copy-source-sse-customer-key XAm0dRrJsEsyPb1UuFNezv1bl9hxuYsgUVC/MUctE2k= \
> --copy-source-sse-customer-key-md5 bY4wkxQejw9mUJfo72k53A=

{
    "SSECustomerAlgorithm": "AES256",
    "CopyObjectResult": {
        "ETag": "\"12a863d8d21c1abedbb4759621f510dc\"",
        "LastModified": "2017-11-10T10:40:19.000Z"
    },
    "SSECustomerKeyMD5": "bY4wkxQejw9mUJfo72k53A="
}
```

If this behavior is not changed the minio server would have to copy the whole object instead of updating the metadata - which is not efficient.

aead:
## Description

This change replaces the current SSE-C key derivation scheme. The 'old'
scheme derives an unique object encryption key from the client provided key.
This key derivation was not invertible. That means that a client cannot change
its key without changing the object encryption key.

AWS S3 allows users to update there SSE-C keys by executing a SSE-C COPY with
source == destination. AWS probably updates just the metadata (which is a very
cheap operation). The old key derivation scheme would require a complete copy
of the object because the minio server would not be able to derive the same
object encryption key from a different client provided key (without breaking
the crypto. hash function).

This change makes the key derivation invertible.

## Motivation and Context
Fixes #5167 

## How Has This Been Tested?
Manually by using aws-cli as described [here](https://github.com/minio/cookbook/blob/master/docs/how-to-use-minio-server-side-encryption-with-aws-cli.md) 

## Types of changes

## Checklist:

### DIFF
--- a/cmd/encryption-v1.go
+++ b/cmd/encryption-v1.go
@@ -18,7 +18,6 @@ package cmd
 
 import (
 	"bytes"
-	"crypto/hmac"
 	"crypto/md5"
 	"crypto/rand"
 	"encoding/base64"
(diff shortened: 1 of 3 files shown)


harshavardhana:
Why is it severe priority ? @deekoder ?

deekoder:
As per @aead  we need to take this fix to ship the next release.

harshavardhana (dismissed):
Understood from @aead about the necessities of this PR and also the breaking changes here.

Here are the security implications of this change as worded by @aead 

> 1. adversary has access to server and copies object metadata.
> 2. adversary compromises client key
> 3. client changes client-key
> => Adversary can still decrypt object because object encryption key is not changed.

> That would not possible if the client updates client-key with non-invertible KDF.
> But this only applies if the adversary copies "just" metadata not entire object which is kind of not realistic...
> If adversary copies metadata & object and compromises client key he can always decrypt - per definition...
> So not really a big difference. Security is (except for theoretical/formal analysis) the same...

Tested aws-cli things are working fine, expected change is existing encrypted objects cannot be decrypted.  Since the key-derivation function has changed.

cmd/encryption-v1.go
```diff
@@ -161,31 +181,44 @@ func EncryptRequest(content io.Reader, r *http.Request, metadata map[string]stri
...
+	iv := sha256.Sum256(nonce[32:]) // derive key encryption key
+	sha = sha256.New()
+	sha.Write(key)
+	sha.Write(iv[:])
+	keyEncryptionKey := sha.Sum(nil)
 
-	mac = hmac.New(sha256.New, key) // derive encryption key
-	mac.Write(keyMAC)
-	reader, err := sio.EncryptReader(content, sio.Config{Key: mac.Sum(nil)})
+	sealedKey := bytes.NewBuffer(make([]byte, 0, 64)) // 16 byte header + 32 byte payload + 16 byte tag
```
fwessels: could declare (fixed length buffer of 64 bytes) on stack

cmd/encryption-v1.go
```diff
@@ -161,31 +181,44 @@ func EncryptRequest(content io.Reader, r *http.Request, metadata map[string]stri
...
+	n, err := sio.Encrypt(sealedKey, bytes.NewReader(objectEncryptionKey), sio.Config{
+		Key: keyEncryptionKey,
+	})
+	if n != 64 || err != nil {
+		return nil, errors.New("failed to seal object encryption key") // if this happens there's a bug in the code (may panic ?)
```
fwessels, line 211: or memory shortage?

cmd/encryption-v1.go
```diff
@@ -198,35 +231,39 @@ func DecryptRequest(client io.Writer, r *http.Request, metadata map[string]strin
...
-	mac := hmac.New(sha256.New, keyHash[:])
-	mac.Write(nonce)
-	if !hmac.Equal(keyMAC, mac.Sum(nil)) {
-		return nil, errSSEKeyMismatch // client-provided key is wrong or object metadata was modified
+	sha := sha256.New() // derive key encryption key
+	sha.Write(key)
+	sha.Write(iv)
+	keyEncryptionKey := sha.Sum(nil)
+
+	objectEncryptionKey := bytes.NewBuffer(make([]byte, 0, 32)) // decrypt object encryption key
```
fwessels: idem: on stack?

fwessels (dismissed):
LGTM (few minor comments)

cmd/encryption-v1.go
aead: Tried that before but the growing functionality of `bytes.Buffer` allocates a new buffer 
(yes, if I declare:
```
var sealedKey [64]byte
n, err := sio.Encrypt(bytes.NewBuffer(sealedKey[:0]), ...
```
)

I can explore why / reason...
The problem is that if we use for example:
```
var sealedKey [64]byte // or any other size instead of 64
n, err := sio.Encrypt(bytes.NewBuffer(sealedKey[:0]), ...)
...

metadata[ServerSideEncryptionSealedKey] = base64.StdEncoding.EncodeToString(sealedKey[:]) // or sealedKey[:64]  
```
and the `bytes.Buffer` made heap copy because of growing than `sealedKey[:]` contains zeros and object cannot be decrypted. So we would depend on a non-obvious implementation detail here. That can cause horrible bugs IMO... 

Nevertheless I try to figure out why `bytes.Buffer` grows because from my understanding there should be no reason...

cmd/encryption-v1.go
aead: You mean the Go runtime runs OoM? That would cause a panic...

cmd/encryption-v1.go
aead: Okay, reason is:
1. bytes.Buffer implements `io.ReadFrom`. `io.Copy` (which is used by `sio.Encrypt`) prefers
   `ReadFrom` function of an `io.Writer` over doing the copy with a buffer.
2. bytes.Buffer's `ReadFrom` implementation allocates buffer always if remaining cap is smaller than 512. See [ReadFrom](https://golang.org/src/bytes/buffer.go?s=6465:6524#L197) 

So (currently) bytes.Buffer will always alloc a new byte slice on the heap if `bytes.Buffer`'s cap is smaller than 512. (like with `var sealedKey [64]byte`). But this is a specific implementation detail and if it's changed in the future our code would not work anymore...

harshavardhana: Bytes buffer implementation is meant to be grown and it will grow with your control 2x times.

aead: Yes, but you "cannot" predict whether it grows. For example the given slice should have enough capacity to hold all data -> no need to grow but the `ReadFrom` function grows always if `cap - len` <  512 even if there would be enough space.

If you write: 
```
buffer := bytes.NewBuffer(make([]byte, 0, 100))
reader := bytes.NewReader(make([]byte, 99))
io.Copy(buffer, reader)
```
the buffer will still grow even though it would not be necessary.

cmd/encryption-v1.go
fwessels: Yes, I guess so. So let's leave it like this.

cmd/encryption-v1.go
fwessels: Ok, fine.

harshavardhana (approved):
LGTM tested

fwessels (approved):
LGTMOFA

PR #5187, opened 2017-11-15
fix a typo

silenceshell:
'atleast' does not exist, it should be 'at least'


nitisht (approved):

krisis (approved):
Thanks for the PR @silenceshell.

PR #5198, opened 2017-11-16
Support for ListObjectParts in azure-gateway

krishnasrinivas:
fixes #5169

## How Has This Been Tested?

manual. mint functional test should be added: https://github.com/minio/mint/issues/223

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added unit tests to cover my changes.
- [ ] I have added/updated functional tests in [mint](https://github.com/minio/mint). (If yes, add `mint` PR # here: )
- [ ] All new and existing tests passed.

Issue #5169, opened 2017-11-10
[feature] Azure Gateway: Implement ListObjectParts


pauln:
## Expected Behavior

`ListObjectParts` returns a list of parts which can be used to generate a subsequent `CompleteMultipartUpload` request

## Current Behavior

`ListObjectParts` always returns an empty list, as per Known Limitations list in Azure Gateway docs.

## Possible Solution

The `ListObjectParts` stub method in Azure Gateway has the comment `ListObjectParts - Use Azure equivalent GetBlockList`, suggesting that it may be possible to implement `ListObjectParts` using existing Azure API methods. 

## Context

While attempting to set up a `tusd` server which stores uploaded files on Azure, I came across Minio Azure Gateway and thought it seemed like a neat way to achieve this as `tusd` supports S3 (but doesn't currently support Azure).  However, `tusd` currently fails to complete the multipart upload as it relies on S3 to provide the list of parts.  I [raised an issue](https://github.com/tus/tusd/issues/156) with `tusd` to see if it was possible to track the parts there, but its architecture makes it infeasible to do so - especially if `tusd` is scaled horizontally (i.e. there are multiple instances of `tusd` potentially handling different chunks of the same upload).

## Your Environment

* Version used (`minio version`): RELEASE.2017-10-27T18-59-02Z
* Environment name and version (e.g. nginx 1.9.1): `minio/minio` Docker image

krishnasrinivas:
We will discuss this internally and get back to you @pauln

krishnasrinivas:
@pauln it is tricky to implement ListObjectParts on Azure:

Here is a sample response:
```
  <Part>
    <PartNumber>2</PartNumber>
    <LastModified>2010-11-10T20:48:34.000Z</LastModified>
    <ETag>"7778aef83f66abc1fa1e8477f296d394"</ETag>
    <Size>10485760</Size>
  </Part>
  <Part>
    <PartNumber>3</PartNumber>
    <LastModified>2010-11-10T20:48:33.000Z</LastModified>
    <ETag>"aaaa18db4cc2f85cedef654fccc4a4x8"</ETag>
    <Size>10485760</Size>
  </Part>
```

We won't be able to derive `LastModified` from Azure as it does not provide this info.

Also if there are two PutObjectPart with same part number, ListObjectParts will not be able to differentiate which is the latest part.

Also https://github.com/aws/aws-sdk-go/issues/1518 ListObjectParts should not be relied upon on Amazon S3.

[...]

PR #5226, opened 2017-11-23
Azure Gateway: Create empty BlockBlob when creating multipart upload

pauln:
I thought that this had already been merged as part of #5198, as @krishnasrinivas [merged it into `krishnasrinivas:azure-listparts`](https://github.com/krishnasrinivas/minio/pull/14), which was the subject of #5198 - but it seems that it was removed from that branch before the PR was merged.  See discussion in #5169.  If there's any chance that a new release could be published after this PR is merged, I would be very appreciative.

## Description
This allows requests such as GetInfo and ListObjectParts to succeed before the first block has been uploaded - thus matching S3 more closely, as S3 does not return a 404 in this situation (as Minio Azure Gateway currently does).

## Motivation and Context
As discussed in #5169, this is required for [tus/tusd](https://github.com/tus/tusd) to be able to use Minio Azure Gateway.

## How Has This Been Tested?
Tested as part of testing #5198.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/gateway-azure.go
+++ b/cmd/gateway-azure.go
@@ -682,10 +682,17 @@ func (a *azureObjects) NewMultipartUpload(bucket, object string, metadata map[st
 		return "", traceError(err)
 	}
 
-	blob := a.client.GetContainerReference(bucket).GetBlobReference(metadataObject)
-	err = blob.CreateBlockBlobFromReader(bytes.NewBuffer(jsonData), nil)
+	metaBlob := a.client.GetContainerReference(bucket).GetBlobReference(metadataObject)
+	err = metaBlob.CreateBlockBlobFromReader(bytes.NewBuffer(jsonData), nil)
 	if err != nil {
 		return "", azureToObjectError(traceError(err), bucket, metadataObject)
+	}
+
+	// Create empty BlockBlob so that attempts to get info and block list don't 404.
+	blob := a.client.GetContainerReference(bucket).GetBlobReference(object)
+	err = blob.CreateBlockBlob(nil)
+	if err != nil {
+		return "", azureToObjectError(traceError(err), bucket, object)
 	}
 
 	return uploadID, nil


krishnasrinivas:
Not sure how it got missed out, but looks like that fix is not correct. Because it is creating a 0 byte file in the main name space.

A better fix is:
```
+++ b/cmd/gateway-azure.go
@@ -746,6 +746,11 @@ func (a *azureObjects) ListObjectParts(bucket, object, uploadID string, partNumb
 
        objBlob := a.client.GetContainerReference(bucket).GetBlobReference(object)
        resp, err := objBlob.GetBlockList(storage.BlockListTypeUncommitted, nil)
+       azureErr, ok := err.(storage.AzureStorageServiceError)
+       if ok && azureErr.StatusCode == http.StatusNotFound {
+               // If no parts are uploaded yet then we return empty list.
+               return result, nil
+       }
        if err != nil {
                return result, azureToObjectError(errors.TraceError(err), bucket, object)
        }
```

pauln:
> looks like that fix is not correct. Because it is creating a 0 byte file in the main name space

It's creating an empty BlockBlob, into which the blocks can subsequently be uploaded (and then concatenated).  As it currently stands (without this patch), the BlockBlob is only created when the first chunk is uploaded

> A better fix is:

I did think about implementing it that way when the issue came up, but:

- It would mean implementing the workaround in multiple methods - at least `GetInfo` as well as `ListObjectParts` - rather than simply creating the empty BlockBlob, which is what I'd expect `NewMultipartUpload` to be doing anyway
- It would result in the aforementioned methods then failing to return an error if called on a file/upload which genuinely doesn't exist (which it probably shouldn't)

krishnasrinivas:
@pauln no, empty BlockBlob would not be desirable.

what I mean is, if you call newmultipartupload and not upload any parts, you will create a 0 byte file:
```
krishna@escape:~$ miniodebug multipart new --bucket test --object testobject
{
  "Bucket": "test",
  "Key": "testobject",
  "UploadID": "54c35031301ff13a"
}
krishna@escape:~$ mc ls myazure/test
[2017-11-22 18:45:57 PST]     0B testobject
krishna@escape:~$ 
```

> It would result in the aforementioned methods then failing to return an error if called on a file/upload which genuinely doesn't exist (which it probably shouldn't)

No actually, we have a check that will fail:
```
        if err = a.checkUploadIDExists(bucket, object, uploadID); err != nil {                          
                return result, err                                                                      
        }                                                                                               
```

pauln:
It still seems strange to me that `NewMultipartUpload` doesn't actually create the BlockBlob - but having dug a bit deeper, the `a.checkUploadIDExists` check does indeed seem to cover the case where the multipart upload doesn't exist at all as it's checking for the existence of the metadata object rather than the actual object.

I've tested your suggestion on top of master, and `tusd` seems to be happy with it.  Would you like me to update this PR to use your patch instead, or would you prefer that I close this PR and leave it to you to handle getting your patch in?

krishnasrinivas:
> Would you like me to update this PR to use your patch instead

Yes please, you can send the PR.

pauln:
Since your fix is a completely different approach to dealing with the original issue, I've opened a fresh PR for it (#5230) and am now closing this one.

PR #5228, opened 2017-11-23
dynamic port in docker’s healthcheck

dehy:
Because :9000 is not always the way it is

## Description
Made the minio port in docker's healthcheck dynamic. Falls back to 9000 if can't define the port. If you find a better (more elegant) way to dynamically find the listening port, feel free to bash me!

## Motivation and Context
When starting a docker container with `server --address ":80" /export` command for instance, the healthcheck.sh script fails after the 120s grace period and marks the container as "unhealthy". And an unhealthy container is not dns-resolvable anymore. And a non-resolvable container is often not a good thing.

## How Has This Been Tested?
I rebuilt an image and ran it in a docker-compose setup with a custom listener port command (`command: server --address ":80" /data`). I waited for 120s and yay, the container was still healthy!

## Types of changes

## Checklist:

### DIFF
--- a/dockerscripts/healthcheck.sh
+++ b/dockerscripts/healthcheck.sh
@@ -19,7 +19,7 @@ set -x
 
 _init () {
     scheme="http://"
-    address="127.0.0.1:9000"
+    address="$(netstat -nplt 2>/dev/null | awk ' /(.*\/minio)/ { gsub(":::","127.0.0.1:",$4); print $4}')"
     resource="/minio/index.html"
     start=$(stat -c "%Y" /proc/1)
 }


dockerscripts/healthcheck.sh
```diff
@@ -19,7 +19,8 @@ set -x
 
 _init () {
     scheme="http://"
-    address="127.0.0.1:9000"
+    port="$(netstat -npta | grep 1/minio | grep LISTEN | awk '{ print $4 }' | cut -d: -f4)"
```
harshavardhana: Good bug fix - we need to actually expand the scope a bit, following code would work fine.

```
address=$(netstat -nplt 2>/dev/null | awk ' /(.*\/minio)/ { gsub(":::","127.0.0.1:",$4); print $4}')
```

The reason is we need to support two different type of scenarios here

- Listening on tcp6 by default i.e and user chooses a different port. 
- Listening on tcp only i.e user provided `--address` the right IP or hostname

harshavardhana (changes requested):

harshavardhana:
Would you mind squashing your commits ?

harshavardhana (dismissed):

nitisht (approved):
LGTM & Tested

harshavardhana (approved):

PR #5247, opened 2017-11-30
Fix hash order to generate more even distribution

harshavardhana:
## Description
The problem in existing code was the following line

```
start := int(keyCrc%uint32(cardinality)) | 1
```

A given a value of N cardinality the ending result
because of the the bitwise '|' would lead to always
higher affinity to odd sequences.

As can be seen from the test cases that this can
lead to many objects being allocated the same set
of disks or atleast the first disk is an odd disk
always.  This introduces a performance problem
for majority of the objects under concurrent load.

Remove `| 1` to provide a more cleaner distribution
and the new code will be.
```
start := int(keyCrc % uint32(cardinality))
```

Thanks to Krishna Srinivas for pointing out the bitwise
situation here.

## Motivation and Context
While working on erasure code pools/packs https://github.com/minio/minio/projects/14 this bug was reproduced and needs to be fixed.

## How Has This Been Tested?
Manually and using `go test`

## Types of changes

## Checklist:

### DIFF
--- a/cmd/xl-v1-utils.go
+++ b/cmd/xl-v1-utils.go
@@ -118,7 +118,7 @@ func hashOrder(key string, cardinality int) []int {
 	nums := make([]int, cardinality)
 	keyCrc := crc32.Checksum([]byte(key), crc32.IEEETable)
 
-	start := int(keyCrc%uint32(cardinality)) | 1
+	start := int(keyCrc % uint32(cardinality))
 	for i := 1; i <= cardinality; i++ {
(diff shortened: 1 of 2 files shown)


fwessels (approved):
LGTM. 

This does not have any retrospective consequences, correct? (as it is recorded in json anyways).

harshavardhana:
No it does not @fwessels existing data is fine.

krishnasrinivas (approved):

PR #4996, opened 2017-09-29
Enable ListMultipartUploads and ListObjectParts for FS

krishnasrinivas:
Prefix based ListMultipartUploads is not supported.

## Description

Previously we had disabled ListMultipartUploads and ListObjectParts to see if any clients break. Docker registry broke. This patch enables ListMultipartUploads and ListObjectParts, however ListMultipartUploads with prefix based listing is not supported (which is not used by docker registry anyway). 

In the near future we will simplify our backend to support this behaviour.

required reviewer @krisis 

## How Has This Been Tested?

Tested with docker registry. Manual testing also done.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #5248, opened 2017-11-30
erasure: Remove prefix based listing support on ListMultipartUploads

harshavardhana:
## Description
Previously we have removed this support under FS on #4996,
deprecate this on erasure coded backend as well to simplify.

## Motivation and Context
This is needed as part of https://github.com/minio/minio/projects/14 

## How Has This Been Tested?
Manually and using `mc ls -I` 

## Types of changes

## Checklist:

### DIFF
--- a/cmd/xl-v1-multipart.go
+++ b/cmd/xl-v1-multipart.go
@@ -272,189 +272,54 @@ func commitXLMetadata(disks []StorageAPI, srcBucket, srcPrefix, dstBucket, dstPr
 	return evalDisks(disks, mErrs), err
 }
 
-// listMultipartUploads - lists all multipart uploads.
-func (xl xlObjects) listMultipartUploads(bucket, prefix, keyMarker, uploadIDMarker, delimiter string, maxUploads int) (lmi ListMultipartsInfo, e error) {
-	result := ListMultipartsInfo{
-		IsTruncated: true,
-		MaxUploads:  maxUploads,
-		KeyMarker:   keyMarker,
-		Prefix:      prefix,
-		Delimiter:   delimiter,
-	}
-
-	recursive := true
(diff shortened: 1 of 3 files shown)


krishnasrinivas (changes requested):
@harshavardhana ideally cleanup code will also needs to be part of the PR, if not, the cleanup code should also go in before the release. Else multipart files will never get deleted.

harshavardhana:
> @harshavardhana ideally cleanup code will also needs to be part of the PR, if not, the cleanup code should also go in before the release. Else multipart files will never get deleted.

I thought we want to do cleanup after this PR, since they are both separate functionalities. Since abort is still supported in the API.

krishnasrinivas (approved):

cmd/xl-v1-multipart.go
```diff
@@ -272,189 +272,54 @@ func commitXLMetadata(disks []StorageAPI, srcBucket, srcPrefix, dstBucket, dstPr
...
-		} else {
-			uploadID = upload.UploadID
-			objectName = upload.Object
+
+		result.NextKeyMarker = object
+		// Loop through all the received uploads fill in the multiparts result.
+		for _, upload := range uploads {
+			uploadID := upload.UploadID
 			result.Uploads = append(result.Uploads, upload)
+			result.NextUploadIDMarker = uploadID
```
poornas, line 310: you are not restricting total uploads across disks to maxUploads. Should decrement len(uploads) from maxUploads after each iteration.

poornas (changes requested):

cmd/xl-v1-multipart.go
poornas: not an issue -because of the break statement on line 319

poornas (approved):

PR #5252, opened 2017-12-01
Check if SSL is configured to validate input arguments

### LINKED ISSUE #5251: Minio runs unencrypted when https URLs are used but certificates are not installed
## Expected Behavior
Minio should error out when listening on or connecting to remote hosts using https:// URLs.

## Current Behavior
It's possible to start a distributed minio cluster while specifying https URLs but which will not operate with TLS enabled.

## Steps to Reproduce
1. On 4 completely empty nodes, install minio.
2. On each machine run "minio server https://node1/data https://node2/data https://node3/data https://node4/data".

## Context
This caused me to incorrectly think minio was performing some sort of cross-node encryption even though it wasn't and was operating with no secure connectivity.

## Your Environment
* Version 2017-11-22T19:55:46Z
* Linux x64

harshavardhana:
## Description
This PR handles following situations

- secure endpoints provided, server should fail to start
  if TLS is not configured
- insecure endpoints provided, server starts ignoring
  if TLS is configured or not.

## Motivation and Context
Fixes #5251

## How Has This Been Tested?
Manually using `go test -v` and following script

```sh
#!/bin/bash

uuid1=$(uuid)
uuid2=$(uuid)
uuid3=$(uuid)
uuid4=$(uuid)

mkdir -p "/tmp/$uuid1" "/tmp/$uuid2" "/tmp/$uuid3" "/tmp/$uuid4"
echo "https://localhost:9001/tmp/${uuid1}" "https://localhost:9002/tmp/${uuid2}" "https://localhost:9003/tmp/${uuid3}" "https://localhost:9004/tmp/${uuid4}" >> /tmp/slots

minio server --address localhost:9001 "https://localhost:9001/tmp/${uuid1}" "https://localhost:9002/tmp/${uuid2}" "https://localhost:9003/tmp/${uuid3}" "https://localhost:9004/tmp/${uuid4}"&
minio server --address localhost:9002 "https://localhost:9001/tmp/${uuid1}" "https://localhost:9002/tmp/${uuid2}" "https://localhost:9003/tmp/${uuid3}" "https://localhost:9004/tmp/${uuid4}"&
minio server --address localhost:9003 "https://localhost:9001/tmp/${uuid1}" "https://localhost:9002/tmp/${uuid2}" "https://localhost:9003/tmp/${uuid3}" "https://localhost:9004/tmp/${uuid4}"&
minio server --address localhost:9004 "https://localhost:9001/tmp/${uuid1}" "https://localhost:9002/tmp/${uuid2}" "https://localhost:9003/tmp/${uuid3}" "https://localhost:9004/tmp/${uuid4}"
```

## Types of changes

## Checklist:

### DIFF
--- a/cmd/endpoint.go
+++ b/cmd/endpoint.go
@@ -65,18 +65,9 @@ func (endpoint Endpoint) Type() EndpointType {
 	return URLEndpointType
 }
 
-// SetHTTPS - sets secure http for URLEndpointType.
-func (endpoint Endpoint) SetHTTPS() {
-	if endpoint.Host != "" {
-		endpoint.Scheme = "https"
-	}
-}
-
-// SetHTTP - sets insecure http for URLEndpointType.
-func (endpoint Endpoint) SetHTTP() {
-	if endpoint.Host != "" {
-		endpoint.Scheme = "http"
-	}
+// IsHTTPS - returns true if secure for URLEndpointType.
+func (endpoint Endpoint) IsHTTPS() bool {
+	return endpoint.Scheme == "https"
 }
 
 // NewEndpoint - returns new endpoint based on given arguments.
@@ -190,18 +181,9 @@ func (endpoints EndpointList) Less(i, j int) bool {
 	return endpoints[i].String() < endpoints[j].String()
 }
 
-// SetHTTPS - sets secure http for URLEndpointType.
-func (endpoints EndpointList) SetHTTPS() {
-	for i := range endpoints {
-		endpoints[i].SetHTTPS()
-	}
-}
-
-// SetHTTP - sets insecure http for URLEndpointType.
-func (endpoints EndpointList) SetHTTP() {
-	for i := range endpoints {
-		endpoints[i].SetHTTP()
-	}
+// IsHTTPS - returns true if secure for URLEndpointType.
+func (endpoints EndpointList) IsHTTPS() bool {
+	return endpoints[0].IsHTTPS()
 }
 
 // NewEndpointList - returns new endpoint list based on input args.
--- a/cmd/server-main.go
+++ b/cmd/server-main.go
@@ -158,6 +158,11 @@ func serverMain(ctx *cli.Context) {
 	globalPublicCerts, globalRootCAs, globalTLSCertificate, globalIsSSL, err = getSSLConfig()
 	fatalIf(err, "Invalid SSL certificate file")
 
+	// Is distributed setup, error out if no certificates are found for HTTPS endpoints.
+	if globalIsDistXL && globalEndpoints.IsHTTPS() && !globalIsSSL {
+		fatalIf(errInvalidArgument, "No certificates found for HTTPS endpoints (%s)", globalEndpoints)
+	}
+
 	if !quietFlag {
 		// Check for new updates from dl.minio.io.
 		mode := globalMinioModeFS
@@ -182,7 +187,6 @@ func serverMain(ctx *cli.Context) {
 	initNSLock(globalIsDistXL)
 
 	// Configure server.
-	// Declare handler to avoid lint errors.
 	var handler http.Handler
 	handler, err = configureServerHandler(globalEndpoints)
 	fatalIf(err, "Unable to configure one of server's RPC services.")


cmd/endpoint.go
```diff
@@ -235,6 +207,8 @@ func NewEndpointList(args ...string) (endpoints EndpointList, err error) {
 			return nil, fmt.Errorf("mixed style endpoints are not supported")
 		} else if endpoint.Scheme != scheme {
 			return nil, fmt.Errorf("mixed scheme is not supported")
+		} else if endpoint.Scheme == "https" && scheme == "https" && !globalIsSSL {
```
balamurugana: endpoint should not worry about any globals.  This check should be outside of endpoint creation.

balamurugana (changes requested):
The check should not be in endpoint creation

harshavardhana:
Can you suggest where the check would be ideally ?

harshavardhana:
> Can you suggest where the check would be ideally ?

No need moved the check into newStorageRPC() instead. @balamurugana PTAL

balamurugana:
@harshavardhana 
> Can you suggest where the check would be ideally ?

Below is correct way to fit the fix
1. Add a method `IsSecure()` or `IsHTTPS()` to `EndpointList` and `Endpoint`
2. in `server-main.go` call `globalPublicCerts, globalRootCAs, globalTLSCertificate, globalIsSSL, err = getSSLConfig()` only if `globalEndpoints.IsSecure() == true`
3. modify `getSSLConfig()` to return `(x509Certs []*x509.Certificate, rootCAs *x509.CertPool, tlsCert *tls.Certificate, err error)` always i.e. it returns error if `getPublicCertFile()` and/or `getPrivateKeyFile()` missing.

harshavardhana:
> Below is correct way to fit the fix
> 
> Add a method IsSecure() or IsHTTPS() to EndpointList and Endpoint
> in server-main.go call globalPublicCerts, globalRootCAs, globalTLSCertificate, globalIsSSL, err = getSSLConfig() only if globalEndpoints.IsSecure() == true
> modify getSSLConfig() to return (x509Certs []*x509.Certificate, rootCAs *x509.CertPool, tlsCert *tls.Certificate, err error) always i.e. it returns error if getPublicCertFile() and/or getPrivateKeyFile() missing.

Don't you think adding a check in newStorageRPC is a correct behavior for this? since that is where we perform an actual request eventually anyways?

It is sort of similar to how newPosix() would validate the disk.  Because the PR needs to address two cases not just TLS

Currently there are two bugs

- server is configured with TLS and insecure endpoints are provided.
- server is not configured with TLS and secure endpoints are provided. 

Doing the fix that you mentioned would not address the `first` case.

harshavardhana:
> Add a method IsSecure() or IsHTTPS() to EndpointList and Endpoint
> in server-main.go call globalPublicCerts, globalRootCAs, globalTLSCertificate, globalIsSSL, err = getSSLConfig() only if globalEndpoints.IsSecure() == true
> modify getSSLConfig() to return (x509Certs []*x509.Certificate, rootCAs *x509.CertPool, tlsCert *tls.Certificate, err error) always i.e. it returns error if getPublicCertFile() and/or getPrivateKeyFile() missing.

Changed to this style @balamurugana PTAL

cmd/common-main.go
```diff
@@ -123,3 +123,23 @@ func handleCommonEnvVars() {
...
+	if ssl := globalEndpoints.IsHTTPS(); ssl {
+		// Check and load SSL certificates under distributed mode.
+		var err error
+		globalPublicCerts, globalRootCAs, globalTLSCertificate, err = getSSLConfig()
+
+		// Set globalIsSSL to true only if certs can be loaded properly.
+		globalIsSSL = err == nil
+
+		// When not running in distributed mode, for example in gateway we ignore
+		// if we cannot load certs, reset the error as success.
```
balamurugana: Can't understand why.  Isn't it possible to run a insecure gateway with TLS backend? similarly secure gateway with insecure backend?

cmd/common-main.go
```diff
@@ -123,3 +123,23 @@ func handleCommonEnvVars() {
 		globalIsEnvDomainName = true
 	}
 }
+
+func loadCerts() {
+	// Check for certificates only if endpoints are secure.
+	if ssl := globalEndpoints.IsHTTPS(); ssl {
```
balamurugana: actually `globalIsSSL = globalEndpoints.IsHTTPS()`.  If possible `globalIsSSL` can be removed.

cmd/endpoint.go
```diff
@@ -65,18 +65,20 @@ func (endpoint Endpoint) Type() EndpointType {
...
 
-// SetHTTP - sets insecure http for URLEndpointType.
-func (endpoint Endpoint) SetHTTP() {
+// IsHTTP - returns true if insecure for URLEndpointType.
+func (endpoint Endpoint) IsHTTP() bool {
 	if endpoint.Host != "" {
-		endpoint.Scheme = "http"
+		return endpoint.Scheme == "http"
 	}
+	return false
```
balamurugana: I see you are moving `SetHTTP()` and `SetHTTPS()` methods.  You can remove them completely and add only one method `IsHTTPS()` i.e. `IsHTTP()` is negative duplicate.

cmd/endpoint.go
```diff
@@ -190,18 +192,22 @@ func (endpoints EndpointList) Less(i, j int) bool {
...
 }
 
-// SetHTTPS - sets secure http for URLEndpointType.
-func (endpoints EndpointList) SetHTTPS() {
+// IsHTTPS - returns true if secure for URLEndpointType.
+func (endpoints EndpointList) IsHTTPS() bool {
+	var ok = true
 	for i := range endpoints {
-		endpoints[i].SetHTTPS()
+		ok = ok && endpoints[i].IsHTTPS()
```
balamurugana: Its not possible to have mixed scheme unless handcrafted `EndpointList`

cmd/common-main.go
harshavardhana: Yes but we shouldn't fail. For gateway if do not add the check we will fail because globalEndpoints has the TLS backend. 

This code doesn't invalidate current behvaioe. It just relaxes secure gateway to be mandatory for secure backend.

cmd/common-main.go
harshavardhana: No it's not true for gateway or FS or standalone Erasure.

harshavardhana:
We needed to handle multiple situations here now it is handled properly @balamurugana @krishnasrinivas PTAL

 The situation is complex because gateway mode also re-purposes globalEndpoints for its remote as well.  This leads to a situation where certs check would fail since we changed the `getSSLConfig()` behavior. 

There is no change needed in getSSLConfig() in-fact the behavior doesn't need to change. We need to validate instead if distributed mode has incorrect arguments in accordance with the current environment.

cmd/endpoint.go
```diff
@@ -190,18 +184,13 @@ func (endpoints EndpointList) Less(i, j int) bool {
...
-// SetHTTP - sets insecure http for URLEndpointType.
-func (endpoints EndpointList) SetHTTP() {
+// IsHTTPS - returns true if secure for URLEndpointType.
+func (endpoints EndpointList) IsHTTPS() bool {
+	var ok = true
 	for i := range endpoints {
-		endpoints[i].SetHTTP()
+		ok = ok && endpoints[i].IsHTTPS()
 	}
+	return ok
```
krishnasrinivas: easier would be:
```
{
for i := range endpoints {
    if !endpoints[i].IsHTTPS() {
        return false
    }
return true
}
```

harshavardhana: How is this easier ?

krishnasrinivas: Second function is easier to read and also more efficient (since we return earlier):
```
func (endpoints EndpointList) IsHTTPS() bool {
        var ok = true
        for i := range endpoints {
                ok = ok && endpoints[i].IsHTTPS()
        }
        return ok
}
```

```
func (endpoints EndpointList) IsHTTPS() bool {
        for _, endpoint := range endpoints {
                if endpoint.IsHTTPS() == false {
                        return false
                }
        }
        return true
}
```

cmd/common-main.go
```diff
@@ -123,3 +123,27 @@ func handleCommonEnvVars() {
 		globalIsEnvDomainName = true
 	}
 }
+
+func loadCerts() {
```
balamurugana: As gateway and server have different logic of handling TLS, this function is not required i.e. handle cases individually in gateway and server.

harshavardhana: Okay

cmd/endpoint.go
```diff
@@ -65,18 +65,12 @@ func (endpoint Endpoint) Type() EndpointType {
...
 	if endpoint.Host != "" {
-		endpoint.Scheme = "https"
-	}
-}
-
-// SetHTTP - sets insecure http for URLEndpointType.
-func (endpoint Endpoint) SetHTTP() {
-	if endpoint.Host != "" {
-		endpoint.Scheme = "http"
+		return endpoint.Scheme == "https"
```
balamurugana: No need to have `if`. Just have this one liner for any type of endpoint.

cmd/endpoint.go
```diff
@@ -190,18 +184,14 @@ func (endpoints EndpointList) Less(i, j int) bool {
...
-
-// SetHTTP - sets insecure http for URLEndpointType.
-func (endpoints EndpointList) SetHTTP() {
-	for i := range endpoints {
-		endpoints[i].SetHTTP()
+		if !endpoints[i].IsHTTPS() {
+			return false
+		}
 	}
+	return true
```
balamurugana: Just do one-liner `return endpoints[0].IsHTTPS()`

harshavardhana: done

cmd/endpoint.go
harshavardhana: done

harshavardhana:
Added all your recommendations @balamurugana

cmd/server-main.go
```diff
@@ -183,8 +190,7 @@ func serverMain(ctx *cli.Context) {
 
 	// Configure server.
 	// Declare handler to avoid lint errors.
-	var handler http.Handler
-	handler, err = configureServerHandler(globalEndpoints)
+	handler, err := configureServerHandler(globalEndpoints)
```
balamurugana: why do you need to change this line? Is above comment not applicable anymore?

harshavardhana: I don't see any linter errors, it was changed because of loadCerts() since err moved into that function. I can change it back.

harshavardhana: if you see few lines below `newObject, err :=` so current master has this sort of frivolous comments IMO.

cmd/server-main.go
```diff
@@ -158,6 +157,14 @@ func serverMain(ctx *cli.Context) {
 	globalPublicCerts, globalRootCAs, globalTLSCertificate, globalIsSSL, err = getSSLConfig()
 	fatalIf(err, "Invalid SSL certificate file")
 
+	// Check if certficates are available, and endpoints are secured as well. We
+	// should validate this only under distributed setup since distributed setup
+	// can have multiple secure endpoints. We shouldn't validate and fail for any
+	// other mode.
```
balamurugana: Below comment is good enough.
```
// In DistXL setup, error out if no certificate found for HTTPS endpoints.
```

cmd/server-main.go
```diff
@@ -158,6 +157,14 @@ func serverMain(ctx *cli.Context) {
...
+	if !globalIsSSL && globalEndpoints.IsHTTPS() && globalIsDistXL {
```
balamurugana: You could change order conditions like `globalIsDistXL && globalEndpoints.IsHTTPS() && !globalIsSSL` to convey the flow.

cmd/server-main.go
harshavardhana: Done

cmd/server-main.go
harshavardhana: Done

cmd/server-main.go
balamurugana: I explicitly added the comment to indicate `golint` error.  Later someone will fix/add the same.  Do you see any issue to left the line as is?

harshavardhana: No i can leave it as is, revert it back to how it was.

harshavardhana:
Fixed @balamurugana

balamurugana (approved):
LGTM

krishnasrinivas (approved):

PR #5257, opened 2017-12-03
fix: Restore empty files when healing

### LINKED ISSUE #5069: Heal reports false success on healing 0 byte files
0 Byte Files are repored as successfully healed but will not actually get the erasure coded part files restored, instead the heal command will report a successful heal on every run.

## Expected Behavior
healing 0 byte files should succeed and actually restore the part files as reliable as files with conent, otherwise admins will think the file is empty because data loss occured.

## Current Behavior
The healing is reported as successful but the part files will not be restored, the checking of successful healing with mc heal --fake will report the same file as still broken and cause heart attac for admins, especially if they dont know the file actually was empty.

## Possible Solution
properly restore the part files even if the are empty.

To reproduce
- add an empty/ 0 byte file (i used the name 'content-script.js') to a 4 node erasure coded minio server (local discs)
- open 2 of the discs directly and delete the folder representation of the erasure coded 0 byte file
- use mc heal -r on the bucket to heal
- run the same command again and notice it reports the same file to be successfully healed again, that we just did repair 

## Your Environment
Version: 2017-09-29T19:16:56Z
Release-Tag: RELEASE.2017-09-29T19-16-56Z
Commit-ID: 60cc6184d253efee4a3120683517028342229e21
run directly on macOS

vadmeste:
## Description
HealFile() does not process the case when an empty file is lost in
some disks. Since, Reedsolomon erasure doesn't handle restoring empty
data, HealFile() will create empty files similarly to CreateFile().

## Motivation and Context
Fixes #5069 

## How Has This Been Tested?
go test + Manually

## Types of changes

## Checklist:

### DIFF
--- a/cmd/erasure-healfile.go
+++ b/cmd/erasure-healfile.go
@@ -78,12 +78,14 @@ func (s ErasureStorage) HealFile(staleDisks []StorageAPI, volume, path string, b
 		blocks[i] = make([]byte, chunksize)
 	}
 	var chunkOffset, blockOffset int64
-	for ; blockOffset < size; blockOffset += blocksize {
+
+	// The for loop below is entered when size == 0 and
+	// blockOffset == 0 to allow for reconstructing empty files.
+	for ; blockOffset == 0 || blockOffset < size; blockOffset += blocksize {
 		// last iteration may have less than blocksize data
 		// left, so chunksize needs to be recomputed.
 		if size < blockOffset+blocksize {
-			blocksize = size - blockOffset
-			chunksize = getChunkSize(blocksize, s.dataBlocks)
+			chunksize = getChunkSize(size-blockOffset, s.dataBlocks)
 			for i := range blocks {
 				blocks[i] = blocks[i][:chunksize]
 			}
@@ -121,9 +123,13 @@ func (s ErasureStorage) HealFile(staleDisks []StorageAPI, volume, path string, b
 		// iteration
 		chunkOffset += chunksize
 
-		// reconstruct data - this computes all data and parity shards
-		if err = s.ErasureDecodeDataAndParityBlocks(blocks); err != nil {
-			return f, err
+		// reconstruct data - this computes all data and
+		// parity shards - but we skip this step if we are
+		// reconstructing an empty file.
+		if chunksize > 0 {
+			if err = s.ErasureDecodeDataAndParityBlocks(blocks); err != nil {
+				return f, err
+			}
 		}
 
 		// write computed shards as chunks on file in each


cmd/erasure-healfile.go
```diff
@@ -117,13 +116,18 @@ func (s ErasureStorage) HealFile(staleDisks []StorageAPI, volume, path string, b
...
-		if err = s.ErasureDecodeDataAndParityBlocks(blocks); err != nil {
-			return f, err
+		if chunksize > 0 {
+			// Reedsolomon does not know how to encode/decode empty data,
+			// therefore, we should only proceed to reconstructing blocks
+			// if we are sure we have some real data waiting for us.
+			if err = s.ErasureDecodeDataAndParityBlocks(blocks); err != nil {
+				return f, err
+			}
+		} else {
```
aead: I think this else is not required. If `chunksize == 0` than `len(blocks[i]) == 0.

aead (changes requested):

cmd/erasure-healfile.go
vadmeste: > I think this else is not required. If chunksize == 0 than `len(blocks[i]) == 0.

Yeah.. but sometimes blocks[i] is empty and sometimes it is nil because the previous loop tries to exit as fast as possible. My policy is to make things consistent in a such random behavior to avoid bugs..

vadmeste:
@donatello can you review this ?

@aead did I convince you ?

cmd/erasure-healfile.go
aead: I do not agree that it would be sometimes nil and sometimes empty. It should be always empty?! We never assign `nil` to a block, right?! However I think it's still fine for explicitness.

aead (approved):
LGTM

vadmeste:
@aead, I modified the PR again.. to make it little simpler.. Kindly review again..

aead:
@vadmeste One question: Can't we extract the logic handling empty data in advance? 
Like:
```
for ; blockOffset == 0 || blockOffset < size; blockOffset += blocksize {
if chunksize == 0 && blockOffset == 0 { // handling zero size objects
     for i := range blocks {
           blocks[i] = blocks[i][:0]
     }
     break
}

if size < blockOffset+blocksize {
     blocksize = size - blockOffset
     chunksize = getChunkSize(size-blockOffset, s.dataBlocks)
     chunksize = getChunkSize(blocksize, s.dataBlocks)
     for i := range blocks {
 		blocks[i] = blocks[i][:chunksize]
 		blocks[i] = blocks[i][:chunksize]	
     }
}
// read a chunk from each disk, until we have
 ....
}
```

vadmeste:
@aead, Not easy to do.. only if we use goto directive. The for block here reads by 10 Mb chunk.. so writing in the disk occurs multiple times. We can't add 'break' as you suggested.

aead:
You're completely right, I missed that the write happens within the for loop. Somehow I thought that the read happens before the write which is of course not the case...

aead (dismissed):

donatello:
Not sure if I'm making a mistake somewhere, but I don't think we need to make such a large change to handle 0-sized files.

We need to add `blockOffset == 0 ||` (as you have already done) to the outer for loop condition and just wrap the Reed Solomon reconstruction call so that it is called only when `chunksize > 0`.

IMO, we do not need to be writing special code to:
- avoid reading 0-sized files etc (it is a premature optimization!)
- make the `blocks[i] == []byte{}` explicitly - it is naturally handled in the existing code

donatello (changes requested):
See comment https://github.com/minio/minio/pull/5257#issuecomment-350129004

vadmeste:
@donatello the problem is more with the readability.. I will follow your suggestion and add more comments

vadmeste:
@donatello I think the change is worthy to keep. It simplifies reading.. because otherwise, many questions will be raised, like why the code is reading 0-bytes data from disks ? I like narrowing scope using if {} else {} to have a clear mind.

It is a big change only because I moved a large block (without modifying it) inside a new conditional if.

vadmeste:
@donatello, are you not convinced ? if yes, let's mark it as blocked, I'll ping @abperiasamy

vadmeste:
@donatello I did what you asked, please review it again.

donatello (dismissed):
LGTM

cmd/erasure-healfile.go
```diff
@@ -121,9 +123,14 @@ func (s ErasureStorage) HealFile(staleDisks []StorageAPI, volume, path string, b
...
 		chunkOffset += chunksize
 
-		// reconstruct data - this computes all data and parity shards
-		if err = s.ErasureDecodeDataAndParityBlocks(blocks); err != nil {
-			return f, err
+		// reconstruct data - this computes all data and
+		// parity shards - but we skip this step if we are
+		// reconstructing an empty file.
+		if chunksize > 0 {
+			// reconstruct data - this computes all data and parity shards
```
donatello: We could remove this extra comment.

vadmeste: > We could remove this extra comment.

Done

donatello (approved):

aead (approved):
LGTM

kannappanr:
@vadmeste Have you run mint with a distributed setup for this PR?

vadmeste:
@kannappanr, I tested with mint, all successful, only minio-go fails due to a known issue (https://github.com/minio/mint/issues/241)

```
Running with
SERVER_ENDPOINT: 192.168.1.92:9013
ACCESS_KEY:      abcd123
SECRET_KEY:      ***REDACTED***
ENABLE_HTTPS:    0
SERVER_REGION:   us-east-1
MINT_DATA_DIR:   /mint/data
MINT_MODE:       full

To get logs, run 'sudo docker cp d0b5dcb24f7d:/mint/log /tmp/mint-logs'
(1/11) Running awscli tests ... done in 29 seconds
(2/11) Running aws-sdk-go tests ... done in 0 seconds
(3/11) Running aws-sdk-php tests ... done in 41 seconds
(4/11) Running aws-sdk-ruby tests ... done in 2 seconds
(5/11) Running mc tests ... done in 9 seconds
(6/11) Running minio-dotnet tests ... done in 43 seconds
(7/11) Running minio-java tests ... done in 51 seconds
(8/11) Running minio-js tests ... done in 21 seconds
(9/11) Running minio-py tests ... done in 10 seconds
(10/11) Running s3cmd tests ... done in 10 seconds
(11/11) Running minio-go tests ... FAILED in 3 seconds
{
  "args": {
    "bucketName": "minio-go-test-c2zjbai496tizr9w",
    "ctx": {
      "Context": 0
    },
    "objectName": "qfkhaa56i516kacfu9se43yy1fbcow"
  },
  "duration": 17,
  "function": "GetObjectWithContext(ctx, bucketName, objectName)",
  "message": "GetObjectWithContext should fail on short timeout",
  "name": "minio-go: testGetObjectWithContextV2",
  "status": "FAIL"
}

Executed 10 out of 11 tests successfully.
```

PR #5280, opened 2017-12-08
Add JSON flag support

### LINKED ISSUE #5197: Add json flag to get json output
This is a new feature, that will be implemented as part of overall improvement to server side logging

Adding this feature will give the option of displaying json formatted logs on the server console, when this flag is enabled.

kannappanr:
Add json flag in the cli
Modify logging function to create a json output,
if json flag is passed.
Fixes #5197

## Types of changes

## Checklist:

### DIFF
--- a/cmd/gateway-main.go
+++ b/cmd/gateway-main.go
@@ -112,6 +112,13 @@ func StartGateway(ctx *cli.Context, gw Gateway) {
 		cli.ShowCommandHelpAndExit(ctx, gatewayName, 1)
 	}
 
+	// Get "json" flag from command line argument.
+	// Set quiet mode if json flag is set.
+	jsonFlag := ctx.Bool("json") || ctx.GlobalBool("json")
+	if jsonFlag {
+		log.EnableJSON()
+	}
+
 	// Get quiet flag from command line argument.
 	quietFlag := ctx.Bool("quiet") || ctx.GlobalBool("quiet")
 	if quietFlag {
--- a/cmd/logger.go
+++ b/cmd/logger.go
@@ -17,11 +17,14 @@
 package cmd
 
 import (
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"os"
 	"path"
 	"runtime"
 	"strings"
+	"time"
 
 	"github.com/Sirupsen/logrus"
 	"github.com/minio/mc/pkg/console"
@@ -29,6 +32,15 @@ import (
 )
 
 var log = NewLogger()
+
+type logEntry struct {
+	Level   string   `json:"type"`
+	Source  string   `json:"source"`
+	Cause   string   `json:"cause"`
+	Message string   `json:"message"`
+	Time    string   `json:"time"`
+	Stack   []string `json:"stack"`
+}
 
 // LogTarget - interface for log target.
 type LogTarget interface {
@@ -48,6 +60,7 @@ type Logger struct {
 	consoleTarget ConsoleLogger
 	targets       []LogTarget
 	quiet         bool
+	json          bool
 }
 
 // AddTarget - add logger to this hook.
@@ -89,6 +102,12 @@ func (log *Logger) Levels() []logrus.Level {
 
 // EnableQuiet - sets quiet option.
 func (log *Logger) EnableQuiet() {
+	log.quiet = true
+}
+
+// EnableJSON - outputs logs in json format.
+func (log *Logger) EnableJSON() {
+	log.json = true
 	log.quiet = true
 }
 
@@ -137,6 +156,26 @@ func getSource() string {
 	return fmt.Sprintf("[%s:%d:%s()]", filename, lineNum, funcName)
 }
 
+func printJSON(level logrus.Level, source, cause, msg string, stack []string, data ...interface{}) {
+	logJSON, err := json.Marshal(&logEntry{
+		Level:   strings.Title(level.String()),
+		Source:  source,
+		Stack:   stack,
+		Cause:   cause,
+		Time:    UTCNow().Format(time.RFC3339Nano),
+		Message: fmt.Sprintf(msg, data...),
+	})
+
+	if err != nil {
+		panic(err)
+	}
+	fmt.Println(string(logJSON))
+
+	if level == logrus.FatalLevel {
+		os.Exit(1)
+	}
+}
+
 func logIf(level logrus.Level, source string, err error, msg string, data ...interface{}) {
 	isErrIgnored := func(err error) (ok bool) {
 		err = errors.Cause(err)
@@ -155,13 +194,23 @@ func logIf(level logrus.Level, source string, err error, msg string, data ...int
 		return
 	}
 
+	var stack []string
+	if terr, ok := err.(*errors.Error); ok {
+		stack = terr.Stack()
+	}
+
+	if log.json {
+		printJSON(level, source, err.Error(), msg, stack, data...)
+		return
+	}
+
 	fields := logrus.Fields{
 		"source": source,
 		"cause":  err.Error(),
 	}
 
-	if terr, ok := err.(*errors.Error); ok {
-		fields["stack"] = strings.Join(terr.Stack(), " ")
+	if len(stack) > 0 {
+		fields["stack"] = strings.Join(stack, " ")
 	}
 
 	switch level {
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -38,6 +38,10 @@ var globalFlags = []cli.Flag{
 		Name:  "quiet",
 		Usage: "Disable startup information.",
 	},
+	cli.BoolFlag{
+		Name:  "json",
+		Usage: "Output server logs and startup information in json format.",
+	},
 }
 
 // Help template for minio.
--- a/cmd/server-main.go
+++ b/cmd/server-main.go
@@ -129,6 +129,13 @@ func serverMain(ctx *cli.Context) {
 		cli.ShowCommandHelpAndExit(ctx, "server", 1)
 	}
 
+	// Get "json" flag from command line argument.
+	// Set quiet mode if json flag is set.
+	jsonFlag := ctx.Bool("json") || ctx.GlobalBool("json")
+	if jsonFlag {
+		log.EnableJSON()
+	}
+
 	// Get quiet flag from command line argument.
 	quietFlag := ctx.Bool("quiet") || ctx.GlobalBool("quiet")
 	if quietFlag {


cmd/logger.go
```diff
@@ -136,7 +155,26 @@ func getSource() string {
...
+		Source:  source,
+		Stack:   stack,
+		Cause:   cause,
+		Time:    time.Now().UTC().String(),
+		Message: fmt.Sprintf(msg, data...),
+	}
+	buf := new(bytes.Buffer)
+	enc := json.NewEncoder(buf)
+	enc.SetEscapeHTML(false)
+	_ = enc.Encode(log)
```
ebozduman: `_ = enc.Encode(log)` => `enc.Encode(log)`

cmd/logger.go
```diff
@@ -136,7 +155,26 @@ func getSource() string {
...
+	fmt.Println(string(buf.String()))
```
ebozduman: You don't need to convert into string twice, i.e.
`fmt.Println(string(buf.String()))` => `fmt.Println(buf.String())`

Additionally, assuming `buf` is in json format, I'd think `buf` is already of string type, hence 
`fmt.Println(buf)` should be good enough.

cmd/logger.go
```diff
@@ -136,7 +155,26 @@ func getSource() string {
...
+	if fatal {
+		level = "Fatal"
+		defer os.Exit(1)
+	}
+	log := &logEntry{
+		Level:   level,
+		Source:  source,
+		Stack:   stack,
+		Cause:   cause,
+		Time:    time.Now().UTC().String(),
```
ebozduman: `time.Now().UTC().String(),` => `time.Now().UTC().Format(time.RFC3339Nano),`

ebozduman (changes requested):

ebozduman (dismissed):

cmd/logger.go
```diff
@@ -30,6 +34,15 @@ import (
 
 var log = NewLogger()
 
+type logEntry struct {
+	Level   string `json:"level"`
+	Source  string `json:"source"`
+	Cause   string `json:"cause"`
+	Message string `json:"message"`
+	Time    string `json:"time"`
+	Stack   string `json:"stack"`
```
balamurugana: If `Stack` is stacktrace, array of string is better than single string.

cmd/logger.go
```diff
@@ -92,16 +106,21 @@ func (log *Logger) EnableQuiet() {
...
 
+// EnableJSON - outputs logs in json format.
+func (log *Logger) EnableJSON() {
+	log.json = true
+}
+
 // Println - wrapper to console.Println() with quiet flag.
 func (log *Logger) Println(args ...interface{}) {
-	if !log.quiet {
+	if !log.quiet && !log.json {
```
balamurugana: isn't `quiet` overriding `json`?  unable to understand why this check is needed.

cmd/logger.go
```diff
@@ -92,16 +106,21 @@ func (log *Logger) EnableQuiet() {
...
 		console.Println(args...)
 	}
 }
 
 // Printf - wrapper to console.Printf() with quiet flag.
 func (log *Logger) Printf(format string, args ...interface{}) {
-	if !log.quiet {
+	if !log.quiet && !log.json {
```
balamurugana: isn't `quiet` overriding `json`?  unable to understand why this check is needed.

cmd/logger.go
```diff
@@ -137,6 +156,27 @@ func getSource() string {
 	return fmt.Sprintf("[%s:%d:%s()]", filename, lineNum, funcName)
 }
 
+func printJSON(fatal bool, source, cause, stack, msg string, data ...interface{}) {
+	level := "Error"
+	if fatal {
```
balamurugana: Pass `level` as string than `fatal` bool

cmd/logger.go
```diff
@@ -137,6 +156,27 @@ func getSource() string {
...
+		Level:   level,
+		Source:  source,
+		Stack:   stack,
+		Cause:   cause,
+		Time:    time.Now().UTC().Format(time.RFC3339Nano),
+		Message: fmt.Sprintf(msg, data...),
+	}
+	buf := new(bytes.Buffer)
+	enc := json.NewEncoder(buf)
+	enc.SetEscapeHTML(false)
```
balamurugana: this is not required IMO.  consumer knows it is JSON encoded.

cmd/logger.go
```diff
@@ -160,23 +200,29 @@ func logIf(level logrus.Level, source string, err error, msg string, data ...int
...
-	case logrus.ErrorLevel:
-		log.logger.WithFields(fields).Errorf(msg, data...)
-	case logrus.WarnLevel:
-		log.logger.WithFields(fields).Warnf(msg, data...)
-	case logrus.InfoLevel:
-		log.logger.WithFields(fields).Infof(msg, data...)
-	default:
-		log.logger.WithFields(fields).Debugf(msg, data...)
+	if log.json {
+		printJSON(level == logrus.FatalLevel, source, err.Error(), stack, msg, data...)
```
balamurugana: `return` and avoid `else`

cmd/logger.go
kannappanr: Without this, characters like `<` , `>` , `&` appear as their UTF encoded value

cmd/logger.go
kannappanr: We do not want to print comments like `Config has been Created` etc to be displayed, irrespective of whether `quiet` flag is set or not, if `json` mode flag is set

cmd/logger.go
kannappanr: Will change it

cmd/logger.go
kannappanr: OK

cmd/logger.go
kannappanr: OK

cmd/logger.go
balamurugana: yes. JSON is not for human.  What is the problem?

cmd/logger.go
balamurugana: then do `EnableQuiet()` when `EnableJSON()`?

cmd/logger.go
kannappanr: For e.g, If the text within a JSON field is `<autogenerated>`, this will show up as `\u003cautogenerated\u003e` . If this is fine, I can do that.

cmd/logger.go
kannappanr: You mean, internally?

balamurugana: yes

cmd/logger.go
balamurugana: thats fine because consumer knows how to read e.g. `jq`

kannappanr: done

cmd/logger.go
kannappanr: done

cmd/logger.go
kannappanr: done

cmd/logger.go
kannappanr: done

cmd/gateway-main.go
```diff
@@ -112,9 +112,16 @@ func StartGateway(ctx *cli.Context, gw Gateway) {
 		cli.ShowCommandHelpAndExit(ctx, gatewayName, 1)
 	}
 
+	// Get "json" flag from command line argument.
+	jsonFlag := ctx.Bool("json") || ctx.GlobalBool("json")
+	if jsonFlag {
+		log.EnableJSON()
```
ebozduman: We can have `EnableQuiet ` call here if jsonFlag is turned on.
I think this way, it'll be easier to understand and complete instead of having `jsonFlag` checked one more time at line#124 together with `quietFlag`.
```
	if jsonFlag {
 		log.EnableJSON()
	}
```
 =>
```
	if jsonFlag {
 		log.EnableJSON()
 		log.EnableQuiet()
	}
```

cmd/logger.go
```diff
@@ -137,6 +155,24 @@ func getSource() string {
...
+	log := &logEntry{
+		Level:   level,
+		Source:  source,
+		Stack:   stack,
+		Cause:   cause,
+		Time:    time.Now().UTC().Format(time.RFC3339Nano),
+		Message: fmt.Sprintf(msg, data...),
+	}
+
+	buf, _ := json.Marshal(log)
```
ebozduman: Variable name `buf` is meaningless here. 
Call it something more meaningful in this context, like `log_json` or something.

cmd/gateway-main.go
kannappanr: done

cmd/logger.go
kannappanr: done

cmd/logger.go
```diff
@@ -160,10 +196,20 @@ func logIf(level logrus.Level, source string, err error, msg string, data ...int
...
 	}
 
+	var stack []string
 	if terr, ok := err.(*errors.Error); ok {
 		fields["stack"] = strings.Join(terr.Stack(), " ")
+		stack = terr.Stack()
 	}
 
+	if log.json {
+		levelString := "Fatal"
```
harshavardhana: You don't need a level String to be like this, you can create a new type for as 

```go
type logLevel string

const (
      logLevelFatal logLevel = "Fatal"
      ...
)
```

cmd/logger.go
```diff
@@ -160,10 +196,20 @@ func logIf(level logrus.Level, source string, err error, msg string, data ...int
 		"cause":  err.Error(),
 	}
 
+	var stack []string
 	if terr, ok := err.(*errors.Error); ok {
 		fields["stack"] = strings.Join(terr.Stack(), " ")
```
harshavardhana: Rather than calling `terr.Stack()` multiple times can we call it once in this `if` conditional? 

```go
var stack []string
if terr, ok := err.(*errors.Error); ok {
         stack = terr.Stack()
}

if len(stack) > 0 {
       fields["stack"] = strings.Join(terr.Stack(), " ")
}
```

cmd/logger.go
```diff
@@ -160,10 +196,20 @@ func logIf(level logrus.Level, source string, err error, msg string, data ...int
 		"cause":  err.Error(),
 	}
```
harshavardhana: fields is not used during `json` mode move this struct below after the `if log.json` conditional.

harshavardhana (changes requested):

cmd/logger.go
```diff
@@ -137,6 +155,24 @@ func getSource() string {
...
+	log := &logEntry{
+		Level:   level,
+		Source:  source,
+		Stack:   stack,
+		Cause:   cause,
+		Time:    time.Now().UTC().Format(time.RFC3339Nano),
+		Message: fmt.Sprintf(msg, data...),
+	}
+
+	logJSON, _ := json.Marshal(log)
```
harshavardhana: If there is an error Marshalling we can `panic(err)` i think. Rather than printing an empty message, we are expecting all our values to be JSON friendly.

cmd/logger.go
```diff
@@ -137,6 +155,24 @@ func getSource() string {
 	return fmt.Sprintf("[%s:%d:%s()]", filename, lineNum, funcName)
 }
 
+func printJSON(level, source, cause, msg string, stack []string, data ...interface{}) {
+	log := &logEntry{
+		Level:   level,
+		Source:  source,
+		Stack:   stack,
+		Cause:   cause,
+		Time:    time.Now().UTC().Format(time.RFC3339Nano),
```
harshavardhana: Use `UTCNow()` helper instead here.

cmd/logger.go
```diff
@@ -137,6 +155,24 @@ func getSource() string {
...
+		Message: fmt.Sprintf(msg, data...),
+	}
+
+	logJSON, _ := json.Marshal(log)
+	fmt.Println(string(logJSON))
+
+	if level == "Fatal" {
```
harshavardhana: This would become `if level == logLevelFatal` after the change.

cmd/logger.go
```diff
@@ -137,6 +155,24 @@ func getSource() string {
 	return fmt.Sprintf("[%s:%d:%s()]", filename, lineNum, funcName)
 }
 
+func printJSON(level, source, cause, msg string, stack []string, data ...interface{}) {
+	log := &logEntry{
```
harshavardhana: No need to write this as a separate line, you can directly do `json.Marshal(&logEntry{})`

harshavardhana (changes requested):

cmd/gateway-main.go
```diff
@@ -112,6 +112,14 @@ func StartGateway(ctx *cli.Context, gw Gateway) {
 		cli.ShowCommandHelpAndExit(ctx, gatewayName, 1)
 	}
 
+	// Get "json" flag from command line argument.
+	// Set quiet mode if json flag is set.
+	jsonFlag := ctx.Bool("json") || ctx.GlobalBool("json")
+	if jsonFlag {
+		log.EnableJSON()
+		log.EnableQuiet()
```
balamurugana: This is incorrect way to address.  `EnableJSON()` should do `EnableQuiet()` not gateway/server/something else

cmd/logger.go
```diff
@@ -137,6 +155,24 @@ func getSource() string {
 	return fmt.Sprintf("[%s:%d:%s()]", filename, lineNum, funcName)
 }
 
+func printJSON(level, source, cause, msg string, stack []string, data ...interface{}) {
```
balamurugana: `level` should be `logrus.Level` and use it as `strings.Title(level.String())` for string.

kannappanr:
@harshavardhana @ebozduman @balamurugana PTAL again

cmd/logger.go
```diff
@@ -30,6 +33,22 @@ import (
 
 var log = NewLogger()
 
+type logLevel string
+
+const (
+	logLevelFatal logLevel = "Fatal"
+	logLevelError logLevel = "Error"
```
balamurugana: -1. use `logrus.Level.String()`

cmd/logger.go
```diff
@@ -155,13 +201,27 @@ func logIf(level logrus.Level, source string, err error, msg string, data ...int
...
+
 	fields := logrus.Fields{
 		"source": source,
 		"cause":  err.Error(),
 	}
 
-	if terr, ok := err.(*errors.Error); ok {
-		fields["stack"] = strings.Join(terr.Stack(), " ")
+	if len(stack) > 0 {
+		fields["stack"] = strings.Join(stack, " ")
```
balamurugana: just do `strings.Join(stack, " ")`.  `len` check is unnecessary.

cmd/logger.go
```diff
@@ -137,6 +163,26 @@ func getSource() string {
 	return fmt.Sprintf("[%s:%d:%s()]", filename, lineNum, funcName)
 }
 
+func printJSON(level logLevel, source, cause, msg string, stack []string, data ...interface{}) {
```
balamurugana: pass `logrus.Level`.  logrus removal PR supposes to bring its own type, not this PR

kannappanr:
@harshavardhana @balamurugana @ebozduman PTAL

balamurugana (commented):
LGTM

harshavardhana (approved):

kannappanr:
Closing this PR as #5293 is also taking care of this issue

PR #5289, opened 2017-12-12
Integrate existing remove bucket functionality from newux to current UI

### LINKED ISSUE #4166: remove a bucket from gui?
In the minio webapp, why there is no way to delete a bucket from the graphical interface?

kaankabalak:
## Description

This commit takes the existing remove bucket functionality written by
brendanashworth, integrates it to the current UI with a dropdown for
each bucket, and fixes small issues that were present, like the dropdown
not disappearing after the user clicks on 'Delete' for certain buckets.
This feature only deletes a bucket that is empty (that has no objects).

## Motivation and Context

Fixes #4166

## How Has This Been Tested?

The changes to the front-end for this commit were tested manually. The remove bucket functionality has been tested for both empty and non-empty buckets. The code removes empty buckets successfully, and returns an error message if the bucket is not empty. A manual check was successfully completed to see whether the dropdowns closed after choosing an option, which was an issue fixed in this commit. As after deleting a bucket, the current bucket becomes empty, a test was also done for the case where the user tries to upload a file immediately after deleting a bucket. It was verified that the code handles this case by not allowing the user to upload a file without choosing a bucket and returning an error message. 
For the back-end changes that were brought in from the newux branch, the relevant unit tests were adapted and updated accordingly to the code in this branch. These tests were ran using the `go test -race ./...` command.

## Types of changes

## Checklist:


browser/app/js/actions.js
```diff
@@ -314,6 +337,37 @@ export const selectBucket = (newCurrentBucket, prefix) => {
...
+          type: 'info',
+          message: `Bucket '${bucket}' has been deleted.`
+        }))
+        dispatch(removeBucket(bucket))
+      })
+      .catch(err => {
+        let message = err.message
+
+        // Show a custom "bucket not empty" message, as it can be confusing.
+        if (/Bucket not empty/.test(err.message)) {
```
vadmeste: I think we don't need to do this for the moment, we can show `Bucket not empty` and let's care about UI error messages in a separate PR with a generic solution.

vadmeste:
@kaankabalak, not able to find how to remove a bucket from the UI (without reading the code), can you explain how ?

kaankabalak:
@vadmeste On the sidebar, you hover over the bucket that you want to delete, click on the dropdown menu that appears (the three vertical dots on the right), and click on 'Delete'. Please let me know if you have any more questions.

screen shot 2017-12-12 at 9 36 01 am

screen shot 2017-12-12 at 9 36 57 am

vadmeste:
Thanks @kaankabalak for the screenshots..

Yeah but I am not having it.. are you sure you committed everything ?

To test your PR, I suppose I only have to type `make` and then run the binary, am I missing something ?

kaankabalak:
@vadmeste, could you try typing `yarn release` in the minio/browser directory, and then going back to the minio directory and typing `make` there? Then you can run the server on the minio directory (I use the `./minio server /tmp` command), go back to the minio/browser directory and type `yarn dev` there to see the changes on http://localhost:8080/minio/.

harshavardhana:
You could update ui-assets.go as well along with this PR.  Makes it easier for folks to just download the PR and test.

kaankabalak:
Thanks @harshavardhana for the recommendation, I thought I already updated the ui-assets.go file while I was working on the commit, but apparently I missed something. So, should I run `yarn release` on the minio/browser directory and update the commit?

vadmeste:
@kaankabalak, I was able to see* the change in `yarn dev` but not with `yarn release && make`

README inside the browser directory is probably missing further instructions to make this work in the release version.

kaankabalak:
@vadmeste, good to hear, please let me know if you have any further questions.

browser/app/js/actions.js
kaankabalak: So I should remove the `if` statement with the `test()` method and just keep the error message as `Bucket '${bucket}' must be empty to delete.`, right?

vadmeste: No, I meant we don't need to enhance the error message for now. It is better if we do this in a separate PR. You can remove the if block.

kaankabalak: Thanks for the explanation, I will apply the relevant changes.

krishnasrinivas (approved):
tested

vadmeste (approved):
LGTM & tested

Issue #4995, opened 2017-09-29
[gateway/gcs] validate whether gateway sends Content-MD5 or checksum to server


balamurugana:
Currently Content-MD5 or other checksum is not sent to GCS server for `PUT`, `POST` methods. Its required to send at least one checksum header to server to verify transferred body.

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

* Version used (`minio version`):
* Environment name and version (e.g. nginx 1.9.1):
* Server type and version:
* Operating System and version (`uname -a`):
* Link to your project:

deekoder:
We need to do this in the next release. Not critical for release.

Praveenrajmani:
The same block as discussed in [https://github.com/minio/minio/issues/4994#issuecomment-434682933](url)

harshavardhana:
Closing this gateway issue to be done as a future milestone.

PR #5315, opened 2017-12-21
Add base64 encoded MD5 output for Hash Reader

donatello:
## Description

- Use base64 encoded MD5 function of Hash Reader to send the Content-MD5 header correctly encoded to S3 Gateway

- Fixes a bug in PutObject with S3 Gateway found when testing with
  Mint.

## Motivation and Context

Found a bug during testing for server release.

## How Has This Been Tested?

By running mint and make sure the failing test passes - specifically this one passed after the fix was applied:

```
{
  "name": "aws-sdk-ruby",
  "function": "putObject(bucket_name, file)",
  "args": {
    "bucket_name": "731612681f16",
    "file": "/mint/data/datafile-1-MB"
  },
  "duration": 2285.67,
  "error": "We encountered an internal error, please try again.",
  "status": "FAIL"
}
```

## Types of changes

## Checklist:

### DIFF
--- a/cmd/gateway/s3/gateway-s3.go
+++ b/cmd/gateway/s3/gateway-s3.go
@@ -289,7 +289,7 @@ func (l *s3Objects) GetObjectInfo(bucket string, object string) (objInfo minio.O
 
 // PutObject creates a new object with the incoming data,
 func (l *s3Objects) PutObject(bucket string, object string, data *hash.Reader, metadata map[string]string) (objInfo minio.ObjectInfo, err error) {
-	oi, err := l.Client.PutObject(bucket, object, data, data.Size(), data.MD5HexString(), data.SHA256HexString(), minio.ToMinioClientMetadata(metadata))
+	oi, err := l.Client.PutObject(bucket, object, data, data.Size(), data.MD5Base64String(), data.SHA256HexString(), minio.ToMinioClientMetadata(metadata))
 	if err != nil {
 		return objInfo, minio.ErrorRespToObjectError(errors.Trace(err), bucket, object)
 	}
@@ -343,7 +343,7 @@ func (l *s3Objects) NewMultipartUpload(bucket string, object string, metadata ma
 
 // PutObjectPart puts a part of object in bucket
 func (l *s3Objects) PutObjectPart(bucket string, object string, uploadID string, partID int, data *hash.Reader) (pi minio.PartInfo, e error) {
-	info, err := l.Client.PutObjectPart(bucket, object, uploadID, partID, data, data.Size(), data.MD5HexString(), data.SHA256HexString())
+	info, err := l.Client.PutObjectPart(bucket, object, uploadID, partID, data, data.Size(), data.MD5Base64String(), data.SHA256HexString())
 	if err != nil {
 		return pi, minio.ErrorRespToObjectError(errors.Trace(err), bucket, object)
 	}
--- a/pkg/hash/reader.go
+++ b/pkg/hash/reader.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"crypto/md5"
 	"crypto/sha256"
+	"encoding/base64"
 	"encoding/hex"
 	"errors"
 	"hash"
@@ -115,6 +116,11 @@ func (r *Reader) SHA256() []byte {
 // MD5HexString returns hex md5 value.
 func (r *Reader) MD5HexString() string {
 	return hex.EncodeToString(r.md5sum)
+}
+
+// MD5Base64String returns base64 encoded MD5sum value.
+func (r *Reader) MD5Base64String() string {
+	return base64.StdEncoding.EncodeToString(r.md5sum)
 }
 
 // SHA256HexString returns hex sha256 value.
--- a/cmd/gateway/s3/gateway-s3-anonymous.go
+++ b/cmd/gateway/s3/gateway-s3-anonymous.go
@@ -28,7 +28,7 @@ import (
 
 // AnonPutObject creates a new object anonymously with the incoming data,
 func (l *s3Objects) AnonPutObject(bucket string, object string, data *hash.Reader, metadata map[string]string) (objInfo minio.ObjectInfo, e error) {
-	oi, err := l.anonClient.PutObject(bucket, object, data, data.Size(), data.MD5HexString(), data.SHA256HexString(), minio.ToMinioClientMetadata(metadata))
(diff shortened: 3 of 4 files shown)


cmd/gateway/s3/gateway-s3.go
```diff
@@ -289,7 +289,7 @@ func (l *s3Objects) GetObjectInfo(bucket string, object string) (objInfo minio.O
 
 // PutObject creates a new object with the incoming data,
 func (l *s3Objects) PutObject(bucket string, object string, data *hash.Reader, metadata map[string]string) (objInfo minio.ObjectInfo, err error) {
-	oi, err := l.Client.PutObject(bucket, object, data, data.Size(), data.MD5HexString(), data.SHA256HexString(), minio.ToMinioClientMetadata(metadata))
+	oi, err := l.Client.PutObject(bucket, object, data, data.Size(), data.MD5Base64String(), data.SHA256HexString(), minio.ToMinioClientMetadata(metadata))
```
krishnasrinivas, line 292: this is why previously md5sum and sha256 were of the type `[]byte` so that it is hex/base64 representation independent and bugs like this are avoided .... we need to revert back to that api in minio-go. however we can take this patch in.

krishnasrinivas (dismissed):

krishnasrinivas (dismissed):

krishnasrinivas (dismissed):

kannappanr (dismissed):
LGTM

pkg/hash/reader.go
```diff
@@ -117,6 +118,11 @@ func (r *Reader) MD5HexString() string {
 	return hex.EncodeToString(r.md5sum)
 }
 
+// MD5Base64String return base64 encoded MD5sum value.
+func (r *Reader) MD5Base64String() string {
```
kannappanr: Typo should be //MD5Base64String returns base64 encoded MD5sum value

kannappanr:
@donatello Can you add unit test for MD5Base64String, that way code coverage will not go down.

cmd/gateway/s3/gateway-s3.go
harshavardhana: We shouldnt send byte that is incorrect. We should give the right encoding this PR is correct. The problem is that we are doing double validation that can be fixed when we truly send the right values to the gateway which we didn't so this PR fixes.

For example let's say we remove hashReader for gateway completely we will be sending the stream directly, if the headers have Content-MD5 we should preserve it and send it directly.

donatello:
@kannappanr @krishnasrinivas Updated unit tests to exercise the new hash reader code too.

krishnasrinivas (dismissed):

harshavardhana (dismissed):
@donatello you have to fix gateway-s3-anonymous.go as well

donatello:
>  @donatello you have to fix gateway-s3-anonymous.go as well

@harshavardhana Fixed, thanks.

pkg/hash/reader.go
donatello: Fixed!

krishnasrinivas (approved):

cmd/gateway/s3/gateway-s3.go
harshavardhana: > so that it is hex/base64 representation independent and bugs like this are avoided

In-fact bugs like this are good it exposes the flow, we should always send the right encoding to a low level API and not expect it to convert to a right encoding.  The encoded values are already there the problem we have is double validation in gateway.  A stream is not necessary to be re-validated since all backends have a way to validate the incoming stream. 

For example 
- https://github.com/minio/minio/issues/4995
- https://github.com/minio/minio/issues/4994

We are not using the gcs/azure features to validate `Content-Md5`

kannappanr (approved):
LGTM

cmd/gateway/s3/gateway-s3.go
krishnasrinivas: if an API takes input in canonical form then programmer bugs like this can be avoided.

> For example let's say we remove hashReader for gateway completely we will be sending the stream directly, if the headers have Content-MD5 we should preserve it and send it directly.

you are thinking about `core.PutObject` API from what is convenient for the gateway i.e gateway can directly pass the `Content-Md5` to PutObject(). But if you think of `core.PutObject` as an API for applications `[]byte` is the canonical way to represent a hash. i.e developers who are using core APIs directly will now have to be careful regarding what type of encoding they do for md5 and sha256.

harshavardhana: Canonical form is not []byte, its the right encoding for HTTP headers is string. The interpretation you are putting forward is not correct here.  Look at the PutObject() APIs in aws-sdk-go they require applications to set the right Content-Md5 and they right fully take string not []byte. Because that is how the S3 API is documented.  It would be a misleading API and not representing the right requirements of S3 API.

harshavardhana: In-fact if we used aws-sdk-go this argument wouldn't exist.  All aws-sdk's right fully document this correctly, neither of them take a byte representation.  Here API is documenting what is required from the user we also named it correctly If someone reads our godoc. 

https://godoc.org/github.com/minio/minio-go#Core.PutObject

```
func (c Core) PutObject(bucket, object string, data io.Reader, size int64, md5Base64, sha256Hex string, metadata map[string]string) (ObjectInfo, error)
```

```
func (c Core) PutObjectPart(bucket, object, uploadID string, partID int, data io.Reader, size int64, md5Base64, sha256Hex string) (ObjectPart, error)
```

One problem is that disconnection between exposed PutObject() v/s core PutObject() i would prefer a consolidation effort where PutObject core API takes *PutObjectOptions which in-fact has a method SetContentMD5() 

http://docs.aws.amazon.com/sdk-for-go/api/service/s3/#PutObjectInput.SetContentMD5

krishnasrinivas: > Canonical form is not []byte

in Go `[]byte` is the canonical form for md5 sum, infact it is `[16]byte` which provides even better typing.

>  Look at the PutObject() APIs in aws-sdk-go

Wouldn't lookup to aws-sdk-go for API design

Let me check with AB what he thinks.

harshavardhana: Well not just aws-sdk-go it's for all SDKs. A canonical form term should be associated to the API/Protocol in question not the language IMO.

PutObjectOptions for example is the right place for these values it already has other values and we can extend that.

Issue #5334, opened 2017-12-29
V4/V2 signature verification should happen in constant time


aead:
The signature verification should happen in constant time (using `subtle.ConstantTimeCompare` not `==`) - otherwise the server provides a timing side-channel which can be used to learn valid signatures for requests.

## Expected Behavior
The signature verification should use `subtle.ConstantTimeCompare`

## Current Behavior
The signature verification uses `==`

## Possible Solution
Replace `==` checks with `subtle.ConstantTimeCompare`

## Context
Tracking issue - described privately on slack. 

## Your Environment
current master

PR #5335, opened 2017-12-29
Prevent timing attack against signature (V2/V4) verification

aead:
## Description
This PR replaces the non-constant time signature (V2 / V4) verification with an constant time verification using `subtle.ConstantTimeCompare`
More details are submitted privately on the slack channel.

## Motivation and Context
#5334 

## How Has This Been Tested?
manually

## Types of changes

## Checklist:

### DIFF
--- a/cmd/signature-v2.go
+++ b/cmd/signature-v2.go
@@ -19,6 +19,7 @@ package cmd
 import (
 	"crypto/hmac"
 	"crypto/sha1"
+	"crypto/subtle"
 	"encoding/base64"
 	"fmt"
 	"net/http"
@@ -71,7 +72,7 @@ func doesPolicySignatureV2Match(formValues http.Header) APIErrorCode {
 	}
 	policy := formValues.Get("Policy")
 	signature := formValues.Get("Signature")
-	if signature != calculateSignatureV2(policy, cred.SecretKey) {
+	if !compareSignatureV2(signature, calculateSignatureV2(policy, cred.SecretKey)) {
 		return ErrSignatureDoesNotMatch
 	}
 	return ErrNone
@@ -124,6 +125,9 @@ func doesPresignV2SignatureMatch(r *http.Request) APIErrorCode {
 	// Extract the necessary values from presigned query, construct a list of new filtered queries.
 	for _, query := range unescapedQueries {
 		keyval := strings.SplitN(query, "=", 2)
+		if len(keyval) != 2 {
+			return ErrInvalidQueryParams
+		}
 		switch keyval[0] {
 		case "AWSAccessKeyId":
 			accessKey = keyval[1]
@@ -163,7 +167,7 @@ func doesPresignV2SignatureMatch(r *http.Request) APIErrorCode {
 	}
 
 	expectedSignature := preSignatureV2(r.Method, encodedResource, strings.Join(filteredQueries, "&"), r.Header, expires)
-	if gotSignature != expectedSignature {
+	if !compareSignatureV2(gotSignature, expectedSignature) {
 		return ErrSignatureDoesNotMatch
 	}
 
@@ -247,11 +251,15 @@ func doesSignV2Match(r *http.Request) APIErrorCode {
 		return ErrInvalidRequest
 	}
 
-	expectedAuth := signatureV2(r.Method, encodedResource, strings.Join(unescapedQueries, "&"), r.Header)
-	if v2Auth != expectedAuth {
+	prefix := fmt.Sprintf("%s %s:", signV2Algorithm, globalServerConfig.GetCredential().AccessKey)
+	if !strings.HasPrefix(v2Auth, prefix) {
 		return ErrSignatureDoesNotMatch
 	}
-
+	v2Auth = v2Auth[len(prefix):]
+	expectedAuth := signatureV2(r.Method, encodedResource, strings.Join(unescapedQueries, "&"), r.Header)
+	if !compareSignatureV2(v2Auth, expectedAuth) {
+		return ErrSignatureDoesNotMatch
+	}
 	return ErrNone
 }
 
@@ -268,12 +276,30 @@ func preSignatureV2(method string, encodedResource string, encodedQuery string, 
 	return calculateSignatureV2(stringToSign, cred.SecretKey)
 }
 
-// Return signature-v2 authrization header.
+// Return the signature v2 of a given request.
 func signatureV2(method string, encodedResource string, encodedQuery string, headers http.Header) string {
 	cred := globalServerConfig.GetCredential()
 	stringToSign := getStringToSignV2(method, encodedResource, encodedQuery, headers, "")
 	signature := calculateSignatureV2(stringToSign, cred.SecretKey)
-	return fmt.Sprintf("%s %s:%s", signV2Algorithm, cred.AccessKey, signature)
+	return signature
+}
+
+// compareSignatureV2 returns true if and only if both signatures
+// are equal. The signatures are expected to be base64 encoded strings
+// according to the AWS S3 signature V2 spec.
+func compareSignatureV2(sig1, sig2 string) bool {
+	// Decode signature string to binary byte-sequence representation is required
+	// as Base64 encoding of a value is not unique:
+	// For example "aGVsbG8=" and "aGVsbG8=\r" will result in the same byte slice.
+	signature1, err := base64.StdEncoding.DecodeString(sig1)
+	if err != nil {
+		return false
+	}
+	signature2, err := base64.StdEncoding.DecodeString(sig2)
+	if err != nil {
+		return false
+	}
+	return subtle.ConstantTimeCompare(signature1, signature2) == 1
 }
 
 // Return canonical headers.
--- a/cmd/signature-v4.go
+++ b/cmd/signature-v4.go
@@ -26,6 +26,7 @@ package cmd
 
 import (
 	"bytes"
+	"crypto/subtle"
 	"encoding/hex"
 	"net/http"
 	"net/url"
@@ -34,7 +35,7 @@ import (
 	"strings"
 	"time"
 
-	"github.com/minio/sha256-simd"
+	sha256 "github.com/minio/sha256-simd"
 )
 
 // AWS Signature Version '4' constants.
@@ -146,6 +147,15 @@ func doesPolicySignatureMatch(formValues http.Header) APIErrorCode {
 	return doesPolicySignatureV4Match(formValues)
 }
 
+// compareSignatureV4 returns true if and only if both signatures
+// are equal. The signatures are expected to be HEX encoded strings
+// according to the AWS S3 signature V4 spec.
+func compareSignatureV4(sig1, sig2 string) bool {
+	// The CTC using []byte(str) works because the hex encoding
+	// is unique for a sequence of bytes. See also compareSignatureV2.
+	return subtle.ConstantTimeCompare([]byte(sig1), []byte(sig2)) == 1
+}
+
 // doesPolicySignatureMatch - Verify query headers with post policy
 //     - http://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-HTTPPOSTConstructPolicy.html
 // returns ErrNone if the signature matches.
@@ -180,7 +190,7 @@ func doesPolicySignatureV4Match(formValues http.Header) APIErrorCode {
 	newSignature := getSignature(signingKey, formValues.Get("Policy"))
 
 	// Verify signature.
-	if newSignature != formValues.Get("X-Amz-Signature") {
+	if !compareSignatureV4(newSignature, formValues.Get("X-Amz-Signature")) {
 		return ErrSignatureDoesNotMatch
 	}
 
@@ -301,7 +311,7 @@ func doesPresignedSignatureMatch(hashedPayload string, r *http.Request, region s
 	newSignature := getSignature(presignedSigningKey, presignedStringToSign)
 
 	// Verify signature.
-	if req.URL.Query().Get("X-Amz-Signature") != newSignature {
+	if !compareSignatureV4(req.URL.Query().Get("X-Amz-Signature"), newSignature) {
 		return ErrSignatureDoesNotMatch
 	}
 	return ErrNone
@@ -380,7 +390,7 @@ func doesSignatureMatch(hashedPayload string, r *http.Request, region string) AP
 	newSignature := getSignature(signingKey, stringToSign)
 
 	// Verify if signature match.
-	if newSignature != signV4Values.Signature {
+	if !compareSignatureV4(newSignature, signV4Values.Signature) {
 		return ErrSignatureDoesNotMatch
 	}
 
--- a/cmd/signature-v4-parser_test.go
+++ b/cmd/signature-v4-parser_test.go
@@ -427,7 +427,7 @@ func TestParseSignV4(t *testing.T) {
 			validateCredentialfields(t, i+1, testCase.expectedAuthField.Credential, parsedAuthField.Credential)
 
 			// validating the extraction/parsing of signature field.
-			if testCase.expectedAuthField.Signature != parsedAuthField.Signature {
+			if !compareSignatureV4(testCase.expectedAuthField.Signature, parsedAuthField.Signature) {
 				t.Errorf("Test %d: Parsed Signature field mismatch: Expected \"%s\", got \"%s\"", i+1, testCase.expectedAuthField.Signature, parsedAuthField.Signature)
 			}
 
@@ -795,7 +795,7 @@ func TestParsePreSignV4(t *testing.T) {
 				t.Errorf("Test %d: Expected the result to be \"%v\", but got \"%v\". ", i+1, testCase.expectedPreSignValues.SignedHeaders, parsedPreSign.SignedHeaders)
 			}
 			// validating signature field.
-			if testCase.expectedPreSignValues.Signature != parsedPreSign.Signature {
+			if !compareSignatureV4(testCase.expectedPreSignValues.Signature, parsedPreSign.Signature) {
 				t.Errorf("Test %d: Signature field mismatch: Expected \"%s\", got \"%s\"", i+1, testCase.expectedPreSignValues.Signature, parsedPreSign.Signature)
 			}
 			// validating expiry duration.
--- a/cmd/streaming-signature-v4.go
+++ b/cmd/streaming-signature-v4.go
@@ -29,7 +29,7 @@ import (
 	"time"
 
 	humanize "github.com/dustin/go-humanize"
-	"github.com/minio/sha256-simd"
+	sha256 "github.com/minio/sha256-simd"
 )
 
 // Streaming AWS Signature Version '4' constants.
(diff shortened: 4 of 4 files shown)


cmd/signature-v2.go
```diff
@@ -124,55 +125,63 @@ func doesPresignV2SignatureMatch(r *http.Request) APIErrorCode {
 	// Extract the necessary values from presigned query, construct a list of new filtered queries.
 	for _, query := range unescapedQueries {
 		keyval := strings.SplitN(query, "=", 2)
+		if len(keyval) != 2 {
```
donatello, line 128: Good catch!

donatello (dismissed):
LGTM

cmd/signature-v2.go
aead: Thanks 😃 
Without the check you can cause panics if you send pre-signed requests without e.g. a signature.
Luckily the server survives the panic - it only kills the go routine. Otherwise we would have an additional DoS vul.

krishnasrinivas (commented):
@aead would be easier to review if clean up can be part of another PR and this PR could just be about using ConstantTimeCompare instead of ==

Also in the cleanup PR you could write what is being cleaned up and why so that it's easy to review.

harshavardhana (changes requested):
Refactors should be avoided in a bug fix change. It is easier to review just the timing attack. Actual refactor is being done here https://github.com/harshavardhana/minio/tree/move-signature

Please refrain from doing minor wording changes.

harshavardhana:
Also the PR title can be a bit better than "Signature timing attack" it looks like an issue description.

aead:
@krishnasrinivas 
The problem is that a the signatures are used as strings everywhere - if you don't want to convert it back and forth all the time you have to change quite some code. (It becomes quite difficult to understand what's going on if you have to think about whether the signature is no base64, or hex or binary...) However @harshavardhana  fixes in general by moving to a `signer` package - which makes changes obsolete.

@harshavardhana @krishnasrinivas @donatello 
Changed PR to minimal bugfix

harshavardhana (dismissed):
LGTM

cmd/signature-v2.go
```diff
@@ -247,11 +251,15 @@ func doesSignV2Match(r *http.Request) APIErrorCode {
 		return ErrInvalidRequest
 	}
 
+	prefix := fmt.Sprintf("%s %s:", signV2Algorithm, globalServerConfig.GetCredential().AccessKey)
+	if !strings.HasPrefix(v2Auth, prefix) {
+		return ErrSignatureDoesNotMatch
+	}
+	v2Auth = v2Auth[len(prefix):]
```
krishnasrinivas, line 258: you can keep the code simple, you can remove the above changes and then, below, you can do:
`subtle.ConstantTimeCompare([]byte(v2Auth), []byte(expectedAuth))`

a new function compareSignatureV2() is not needed

krishnasrinivas: If you'd like to have a wrapper function, you can just call it:
```
constantTimeStrCompare(str1, str2 string) {
return subtle.ConstantTimeCompare([]byte(str1), []byte(str2))
}
```
i.e there is no need to do decodestring()

also you can use the same function for sig-v2 and v4

aead: That does not work - since because of padding. See for example:
https://play.golang.org/p/McT2-0_BwLR

For signature V4 it would work because the signature is represented as HEX - but we could easily run into edge cases because somebody must know how a string is represented (and whether its okay to use string conversion) whenever constant time compare is used.

Therefore it's not recommended to do things like:
`subtle.ConstantTimeCompare(string(a), string(b))`

krishnasrinivas: > That does not work - since because of padding. See for example:
https://play.golang.org/p/McT2-0_BwLR

what you are indicating here is, for base64, there can be multiple base64 representations for the same []byte. In this case I agree, it is safer to convert base64 to a canonical byte form before comparing.

You can add a comment saying why it is converted to byte form before comparing. Else it looks redundant to the regular eyes.

> For signature V4 it would work because the signature is represented as HEX - but we could easily run into edge cases because somebody must know how a string is represented (and whether its okay to use string conversion) whenever constant time compare is used.

Since hex representation does not have the problem of base64 and also V4 is always hex, you can safely do `subtle.ConstantTimeCompare([]byte(got), []byte(expected))` Because if you are decoding it to byte before comparing, it will be confusing to the reader as to why it is being done so when it is not providing any advantage.

Note that getSignature() always returns the signature in hex.

aead: >what you are indicating here is, for base64, there can be multiple base64 representations for the same []byte. In this case I agree, it is safer to convert base64 to a canonical byte form before comparing.

Correct.

>You can add a comment saying why it is converted to byte form before comparing. Else it looks redundant to the regular eyes.

Will add comments.

>Since hex representation does not have the problem of base64 and also V4 is always hex, you can safely do subtle.ConstantTimeCompare([]byte(got), []byte(expected)) Because if you are decoding it to byte before comparing, it will be confusing to the reader as to why it is being done so when it is not providing any advantage.
Note that getSignature() always returns the signature in hex.

Right, but that's exactly the behavior I changed in the previous version of this PR. In the previous version there were no conversions to strings any more. (I dropped this bc @harshavardhana is working on a complete refactoring of the signature verification code - moving to a package)

I completely agree that the current code looks like converting strings to bytes back an forth.
But that's IMO an issue of the existing code - I fixed that in the previous version of this PR but dropped it to make review easier. As mentioned it's is not recommended to do
`subtle.ConstantTimeCompare([]byte(str1), []byte(str2))` because different strings can be decoded into the same binary representation - depending on the encoding.
Even if it's safe to do `subtle.ConstantTimeCompare([]byte(str1), []byte(str2))` if both strings are e.g. HEX values the code would implicitly require that the encoding is unambiguous.

IMO if we would use `subtle.ConstantTimeCompare([]byte(str1), []byte(str2))` for V4 we can safe about 8 lines of code but include a "bad" usage of `subtle.ConstantTimeCompare` - even though it would work. 

>Because if you are decoding it to byte before comparing, it will be confusing to the reader as to why it is being done so when it is not providing any advantage.

A reader of `subtle.ConstantTimeCompare([]byte(str1), []byte(str2))` could assume that it's okay to implement such checks always this way - which is not correct.

krishnasrinivas: > IMO if we would use subtle.ConstantTimeCompare([]byte(str1), []byte(str2)) for V4 we can safe about 8 lines of code but include a "bad" usage of subtle.ConstantTimeCompare - even though it would work.

It is not "bad" usage of subtle.ConstantTimeCompare() at all, subtle.ConstantTimeCompare is a primitive and should not be concerned about whether the byte slice is encoded or not.

```
func compareSignatureV4(sig1, sig2 string) bool {
	// convert signature to binary representation
	// subtle.ConstantTimeCompare([]byte(sig1), []byte(sig2)) is not
	// recommended even though it would work for HEX values.
	signature1, err := hex.DecodeString(sig1)
	if err != nil {
		return false
	}
	signature2, err := hex.DecodeString(sig2)
	if err != nil {
		return false
	}
	return subtle.ConstantTimeCompare(signature1, signature2) == 1
}
```

This function is equivalent to:
```
func compareSignatureV4(sig1, sig2 string) bool {
	return subtle.ConstantTimeCompare([]byte(signature1), []byte(signature2)) == 1
}
```

because of the property that:
if a == b then []byte(hexEncoding(a)) == []byte(hexEncoding(b))

```
	// convert signature to binary representation
	// subtle.ConstantTimeCompare([]byte(sig1), []byte(sig2)) is not
	// recommended even though it would work for HEX values.
```

If you try to explain why it is not recommended and also say that v4 signature is hex you will see that it will confuse the reader.

Since V4 is always hex and name of this function is compareSignatureV4, we should just do subtle.ConstantTimeCompare([]byte(sig1), []byte(sig2)) and not add any unnecessary code.

That said, someone else can approve the code to move things forward as I can't approve this.

aead:
I've no clue why adding comments lead to `TestPostPolicyForm` failure...

cmd/signature-v2.go
aead: I don't doubt that the functionality is equivalent for hex values.
The problem I have with this is that the solution you're proposing **only** works because of the fact the hex encoding is unambiguous. Somebody who read that code **MUST** be aware of that. I'm saying that `subtle.ConstantTimeCompare([]byte(str1), []byte(str2)) is not a good style because your implicitly assuming something about str1 and str2 which is not at all obvious.
Saving a few lines doesn't compensate this IMO.

The confusion comes (IMO) from the existing code which turns the computed signature into a hex value instead of decoding the received signature.

aead:
Merging is blocked because of #5336

harshavardhana:
> Merging is blocked because of #5336

This PR https://github.com/minio/minio/pull/5337 should unblock this PR.

cmd/signature-v2.go
harshavardhana: For now i think existing code for hex is fine @krishnasrinivas  @aead idea to make it look consistent and more forth coming is okay in this situation since it is a careful fix.

nitisht:
@aead #5337 is merged, please re run the tests.

cmd/signature-v2.go
donatello: > The problem I have with this is that the solution you're proposing only works because of the fact the hex encoding is unambiguous. Somebody who read that code MUST be aware of that.

@aead: This argument that the user/reader needs to be aware of the properties of certain encodings appears specious to me. For V4, the comment states:

```
+// compareSignatureV4 returns true if and only if both signatures
+// are equal. The signatures are excepted to be HEX encoded strings
+// according to the AWS S3 signature V4 spec.
+func compareSignatureV4(sig1, sig2 string) bool { 
```

From the V4 spec it is already required that the user be aware the signature is encoded as hex and from V2 spec that it is encoded as Base64. When this is the case, and we use separate functions for V2 and V4, separately handling the encoding issues are fine. While the properties may not be obvious they are quite easy to learn once it is mentioned.

For V2 we can just add a comment saying that the Base64 encoding of a byte-sequence is not unique and therefore, decoding is required, and for V4 we just add a comment saying that since hexadecimal encoding of a byte-sequence is unique, no decoding is required (and we can do a direct CTC). Now we are making the reason for decoding/not decoding clear. Saying that something is "not recommended" is an [argument from authority](https://en.wikipedia.org/wiki/Argument_from_authority), and makes the code more difficult to understand.

cmd/signature-v2.go
```diff
@@ -268,12 +276,30 @@ func preSignatureV2(method string, encodedResource string, encodedQuery string,
...
 func signatureV2(method string, encodedResource string, encodedQuery string, headers http.Header) string {
 	cred := globalServerConfig.GetCredential()
 	stringToSign := getStringToSignV2(method, encodedResource, encodedQuery, headers, "")
 	signature := calculateSignatureV2(stringToSign, cred.SecretKey)
-	return fmt.Sprintf("%s %s:%s", signV2Algorithm, cred.AccessKey, signature)
+	return signature
+}
+
+// compareSignatureV2 returns true if and only if both signatures
+// are equal. The signatures are excepted to be base64 encoded strings
```
donatello: s/excepted/expected

cmd/signature-v2.go
```diff
@@ -268,12 +276,30 @@ func preSignatureV2(method string, encodedResource string, encodedQuery string,
...
+// according to the AWS S3 signature V2 spec.
+func compareSignatureV2(sig1, sig2 string) bool {
+	// convert signature to binary representation
+	// subtle.ConstantTimeCompare([]byte(sig1), []byte(sig2)) will not
+	// work correctly and is not recommended because of ambiguous endcoding
```
donatello: This comment is better restated as:

```
// Decode signature string to binary byte-sequence representation is required as Base64
// encoding of a value is not unique (<link or explanation>).
```

cmd/signature-v4.go
```diff
@@ -146,6 +147,24 @@ func doesPolicySignatureMatch(formValues http.Header) APIErrorCode {
 	return doesPolicySignatureV4Match(formValues)
 }
 
+// compareSignatureV4 returns true if and only if both signatures
+// are equal. The signatures are excepted to be HEX encoded strings
+// according to the AWS S3 signature V4 spec.
+func compareSignatureV4(sig1, sig2 string) bool {
+	// convert signature to binary representation
+	// subtle.ConstantTimeCompare([]byte(sig1), []byte(sig2)) is not
+	// recommended even though it would work for HEX values.
```
donatello: I suggest scrapping this comment, and the unnecessary decoding of hexadecimal to binary. Just add a comment that since hexadecimal encodings are unique for a sequence of bytes, decoding is not required.

donatello (changes requested):

cmd/signature-v4.go
```diff
@@ -146,6 +147,24 @@ func doesPolicySignatureMatch(formValues http.Header) APIErrorCode {
 	return doesPolicySignatureV4Match(formValues)
 }
 
+// compareSignatureV4 returns true if and only if both signatures
+// are equal. The signatures are excepted to be HEX encoded strings
```
donatello: s/excepted/expected

donatello (changes requested):

cmd/signature-v2.go
aead: @donatello 
>From the V4 spec it is already required that the user be aware the signature is encoded as hex and from V2 spec that it is encoded as Base64. When this is the case, and we use separate functions for V2 and V4, separately handling the encoding issues are fine.

Correct - The `compareSignatureV4` function expects a signature as hex - as mentioned in the docs - and `subtle.ConstantTimeCompare([]byte(str1), []byte(str2))` will work because of this fact. So in **this** specific case it will work as expected. Why I'm saying doing it like this is not recommended is that a reader of the code has to be aware of this implicit dependency (CTC only works because of hex decoding is not ambiguous) to understand why the code would be correct.
 I'm much more making a general case here saying: "Avoid `subtle.ConstantTimeCompare([]byte(str1), []byte(str2))` because it can cause subtle bugs"
It's a general recommendation based on experience.

In this case (V4) we can add a comment saying something like `// this only works because  the hex encoding is not ambiguous` and use the shorter version.
As @harshavardhana said I'm trying to avoid implicit assumptions to make reasoning about the code simple.
From my point of view the better approach to avoid en/decoding multiple times would be to let return getSignatureV4 a slice  - not an hex string.
BTW []byte(str) does a copy anyway (it's not just a type cast) - So the code is shorter but the data is copied too.

@harshavardhana @donatello @krishnasrinivas
But I would accept `subtle.ConstantTimeCompare([]byte(str1), []byte(str2)) with a comment to get this fix merged. General refactoring can be done later on.

cmd/signature-v2.go
aead: Done, thanks

cmd/signature-v4.go
donatello: Spelling needs fix.

cmd/signature-v2.go
donatello: > Why I'm saying doing it like this is not recommended is that a reader of the code has to be aware of this implicit dependency (CTC only works because of hex decoding is not ambiguous) to understand why the code would be correct.

My point simply put, is that the "dependency" is no longer implicit, once it is mentioned (as a comment). 

> I'm much more making a general case here saying: "Avoid subtle.ConstantTimeCompare([]byte(str1), []byte(str2)) because it can cause subtle bugs"
> It's a general recommendation based on experience.

This statement requires a large amount of further qualification. For e.g. if we wanted to compare two strings with CTC for some (cryptographic) reason, this would be the only way because in Go, we would not implement CTC for more types such as strings, []rune, etc.

harshavardhana (dismissed):
LGTM

cmd/signature-v4.go
aead: Oh sry, missed that - fixed

cmd/signature-v2.go
krishnasrinivas: > @harshavardhana @donatello @krishnasrinivas
But I would accept `subtle.ConstantTimeCompare([]byte(str1), []byte(str2)) with a comment to get this fix merged. General refactoring can be done later on.

A comment at compareSignatureV4 will not hurt. But comment at compareSignatureV2 is definitely needed (the comment that you have put explains well) explaining on why decoding is done before comparing because the RFC https://tools.ietf.org/html/rfc4648 does not seem to indicate any ambiguity, only golang's base64 implementation allows ambiguous decoding. Wondering why golang just does not return an error upon encountering characters which are not allowed.

aead: >My point simply put, is that the "dependency" is no longer implicit, once it is mentioned (as a comment).

That I agree. But that's probably a school-of thought issue - "self-explaining code" vs. "clarifying comments". Like vim vs. emacs or tabs vs. spaces 😉    

>This statement requires a large amount of further qualification. For e.g. if we wanted to compare two strings with CTC for some (cryptographic) reason, this would be the only way because in Go, we would not implement CTC for more types such as strings, []rune, etc.

True, but that's the typical 90%-fit approach of Go. Further this is a general equality problem.
E.g. two Base64 strings can have the same meaning (represent the same value) even though the are not equal. Therefore the recommendation is to "normalize" the values (if necessary) and than compare them. Imagine two strings one encoded as utf-8 (like all go `string`) and one as utf-16 - they can represent the same value but may not be equal.  
I completely agree that in our case the solution works and is well understood. However I would still recommend to:
 - decode the client provided signature
 - compute the signature on the server side (as byte slice)
 - compare them

instead of:
 - compute the signature on the server side (as byte slice)
 - encode the server side signature as hex string
 - convert client and server signature from hex string to byte slice
 - compare them

aead: @krishnasrinivas 
I just took a quick look at the RFC... Go allows arbitrary new-lines in the string - don't know exactly why... The only thing the RFC mentioned are line feeds (3.1.  Line Feeds in Encoded Data)
EDIT: The (12.  Security Considerations) talk about ignored characters... maybe Go's behavior has compatibility reasons...

krishnasrinivas: > However I would still recommend to:
> 
> decode the client provided signature
> compute the signature on the server side (as byte slice)
> compare them
> instead of:
> 
> compute the signature on the server side (as byte slice)
> encode the server side signature as hex string
> convert client and server signature from hex string to byte slice
> compare them

agreed @aead there will be no confusion in the code if we always represent something in its canonical form. In golang, for signature, canonical form would be `[]byte`, if the code is refactored to do what you have said the above the current discussion would not have been needed. That is what you tried to do in the initial PR? Because, it was not obvious what was being done in that PR and it looked like a lot of changes for a simple fix.

harshavardhana: I would suggest we avoid long form comments like these keep it for slack discussions, this PR is ready to be merged.

harshavardhana: Simply use https://golang.org/pkg/encoding/base64/#Encoding.Strict for stricter decoding behavior. It is part of RFC to allow non strict behavior as well.

harshavardhana: For more information look here https://github.com/golang/go/issues/15656

krishnasrinivas (approved):

harshavardhana (approved):

cmd/signature-v2.go
aead: @krishnasrinivas 
>That is what you tried to do in the initial PR? 

Yes, exactly

Thanks @harshavardhana for sharing!

nitisht:
@donatello are any of your review comments pending here?

donatello (approved):

Issue #5341, opened 2018-01-02
In what folder/directory does Minio store its objects/files on local disk?


billmei:
I'm running a Minio instance on my local machine, on http://localhost:9000. I created a bucket called `foo`, and uploaded a single file `test.txt` to it.

Since this Minio instance is running on my local machine, where is the "uploaded" binary `test.txt` file actually stored in my local filesystem? That is, is it possible to locate this `test.txt` file in a folder somewhere using the GUI filesystem browser? For example, `/Users/myusername/some/hidden/folder/minio/buckets/foo/test.txt`

I looked through [the docs](https://docs.minio.io/docs/minio-client-complete-guide) and it only has information about using the `mc` command via `mc ls` to view the files in the "remote" Minio machine via the command line. This is not what I'm looking for, I want to see under what folder the `test.txt` file actually resides on my local machine. Is this possible, or am I forced to only view an abstraction of the `test.txt` through the `mc` command?

(Cross-post of the StackOverflow question: https://stackoverflow.com/questions/47973268/in-what-folder-directory-does-minio-store-its-objects-files-on-local-disk)

nitisht:
@billmei the file is stored in the mount you pass to start `minio` server. For example if you started `minio` by `minio server /tmp`, create a bucket `foo`, and upload object `bar.txt`. The file will be present at `/tmp/foo/bar.txt`

billmei:
I see it now, thanks @nitisht! I went ahead and added this to the documentation: https://github.com/minio/minio/pull/5342

nitisht:
Closing this as answered, @billmei feel free to join us on [Slack](https://minio.slack.com) for further questions..

PR #5342, opened 2018-01-02
Add info about where Minio mounts its filesystem

billmei:
Fixes https://github.com/minio/minio/issues/5341 by adding clarifying documentation information.

See the issue for more details.

I'm not sure if this is the right location for this documentation, but am happy to move it to a different file if there is a more appropriate place to put this.

## Checklist:

### DIFF
--- a/docs/browser/README.md
+++ b/docs/browser/README.md
@@ -15,6 +15,8 @@ Currently these tokens expire after 10hrs, this is not configurable yet.
 minio server /data
 ```
 
+This will mount the filesystem that Minio uses in the `/data` directory of the host machine. For example, if you run Minio locally and create a bucket called `foo`, then you can access any files you "uploaded" to the `foo` bucket by navigating to the `/data/foo` directory on your local disk.
+
 ### JSON RPC APIs.
 
 JSON RPC namespace is `Web`.


docs/browser/README.md
```diff
@@ -15,6 +15,8 @@ Currently these tokens expire after 10hrs, this is not configurable yet.
 minio server /data
 ```
 
+This will mount the filesystem that Minio uses in the `/data` directory of the host machine. For example, if you run Minio locally and create a bucket called `foo`, then you can access any files you "uploaded" to the `foo` bucket by navigating to the `/data/foo` directory on your local disk.
```
harshavardhana, line 18: Minio doesn't mount anything, minio simply uses what is already present on the disk. Minio has other modes of operation as well so the line indicating accessing backend is not correct. In-fact it is discouraged to access the backend i.e /data directly even if its possible. 

Minio server manages concurrent access to backends - accessing backend directly would defeat this and lead to data loss and application issues.

harshavardhana (changes requested):

docs/browser/README.md
billmei: Interesting... My use case is to use minio as a scriptable front-end interface to my local filesystem, as I'm building an app on top of minio an alternative to using Finder.

harshavardhana: You can @billmei but this doc change might not be needed IMO.

billmei: Ok, I will close this pull request.

PR #5466, opened 2018-01-29
FS: Always read bucket policy directly from the backend

### LINKED ISSUE #4458: Policies are not shared between all instances of minio
Problem:
Bucket policy is just set to one instance

Expected:
Shared state of policy.

Environment:
We are running a nomad/consul cluster with 3 docker instances of minio:latest and a fabio load balancer. All minio instances mount the same storage endpoint into the container. The load balancers are using round robin to route to minio and are set as a TLS endpoint.

If I change bucket policy to public:

```
mc policy s3/conf
Access permission for ‘s3/conf’ is ‘public
```

I can download files approx. every 3rd reload. 

```
curl https://s3.ourdomain.inc/conf/index.html

<!DOCTYPE html>
<html>
....
```

```
curl https://s3.ourdomain.inc/conf/index.html

<Error><Code>AccessDenied</Code><Message>Access Denied.</Message><Key></Key><BucketName></BucketName><Resource>/conf/index.html</Resource><RequestId>3L137</RequestId><HostId>3L137</HostId></Error>

```

Is this expected and minio is not capable of using a shared storage space?

vadmeste:
## Description
Multiple instances of shared FS have no way to share in realtime
a bucket policy change. This PR changes the behavior of GetBucketPolicy
to always read and parse bucket policy document from the backend
if this is an FS.

## Motivation and Context
Fixes #4458 

## How Has This Been Tested?
go test + Manually

## Types of changes

## Checklist:

### DIFF
--- a/cmd/bucket-policy.go
+++ b/cmd/bucket-policy.go
@@ -43,9 +43,13 @@ var globalBucketPolicies *bucketPolicies
 // Global bucket policies list, policies are enforced on each bucket looking
 // through the policies here.
 type bucketPolicies struct {
-	rwMutex *sync.RWMutex
+	// Object API to get/set bucket policies
+	objAPI ObjectLayer
+	// Always read policy from the backend
+	alwaysRefresh bool
 
-	// Collection of 'bucket' policies.
+	// Memory version of 'bucket' policies.
+	rwMutex             *sync.RWMutex
 	bucketPolicyConfigs map[string]policy.BucketAccessPolicy
 }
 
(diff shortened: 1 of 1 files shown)


cmd/bucket-policy.go
```diff
@@ -137,6 +152,8 @@ func initBucketPolicies(objAPI ObjectLayer) error {
 	globalBucketPolicies = &bucketPolicies{
 		rwMutex:             &sync.RWMutex{},
 		bucketPolicyConfigs: policies,
+		objAPI:              objAPI,
+		alwaysRefresh:       objAPI.StorageInfo().Backend.Type == FS,
```
donatello, line 156: Would it not be better to use `globalObjectAPI` instead of storing it inside the bucket policies struct?

donatello (changes requested):

harshavardhana:
This PR is not needed @poornas is converging the layers for both Gateway and Object API where this style of functionality will be handling by NAS gateway. 

Implementation would move the code for reading and accessing BucketPolicy to the objectLayer code such as caching etc. will move to the lower layer.  Top layer simply performs GetBucketPolicy() and applies the policy returned to the incoming request. 

Now underneath if the GetBucketPolicy() implementation caches or not is implementation dependent such as erasure coding, single disk fs mode all will cache at their respective object layers where as NAS will directly read from the backend.

vadmeste:
@harshavardhana makes sense.. closing this.

Issue #5462, opened 2018-01-29
Title case is used wrongly in the error


balamurugana:
Currently log message is title cased.  

Example error message:
```
[2018-01-29T08:14:51.373640438Z] [ERROR] Unable to lookup webhook endpoint http://localhost:12345 (Post Http://Localhost:12345: Dial Tcp 127.0.0.1:12345: Getsockopt: Connection Refused)
```

kannappanr:
#5468 has been merged. Hence closing the issue

PR #5450, opened 2018-01-26
[WIP] Move error strings into a file

kannappanr:
Moves all the error strings to a single file. Each error string
is contained within a corresponding error logging function.

Fixes #5264

## How Has This Been Tested?

Manually

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added unit tests to cover my changes.
- [ ] I have added/updated functional tests in [mint](https://github.com/minio/mint). (If yes, add `mint` PR # here: )
- [x] All new and existing tests passed.

PR #5468, opened 2018-01-30
Removes Title() for capitalization of error cause

ebozduman:
`Title()` library function was used for error cause string and it was capitalizing the first letters of every word in the error cause string.

## Description
`Title()` library function call is removed and no capitalization is needed. Error cause messages are expected to be defined correctly when they are created.

## Motivation and Context
#5462

## How Has This Been Tested?
minio server is rebuilt locally and ran against mint test

## Types of changes

## Checklist:

### DIFF
--- a/cmd/logger.go
+++ b/cmd/logger.go
@@ -172,7 +172,8 @@ func logIf(level Level, err error, msg string,
 	if err == nil || isErrIgnored(err) {
 		return
 	}
-	cause := strings.Title(err.Error())
+	// Get the cause for the Error
+	cause := err.Error()
 	// Get full stack trace
 	trace := getTrace(3)
 	// Get time


kannappanr (dismissed):
LGTM

balamurugana:
Capitalizing any error message is not correct.  We shouldn't change user/system inputs/names e.g. for error message `foo/bar: The system cannot find the path specified`, we shouldn't change the case.

IMO you would need to fix error message than fixing generically.

ebozduman:
@balamurugana ,
I agree that ideally and in general all error and cause messages are supposed to be created/written with correct wording and format. However, there'll be always the possibility of some stray messages with improper capitalization.

When we were testing the new logger mechanism, there were cause messages that did not adhere the expected capitalization format, and @abperiasamy suggested handling it without going through all the error and cause messages one by one for practical reasons.

@kannappanr's effort as part of the PR #5450 will address error messages (not cause messages though) and will make sure they are worded and formatted correctly. 

I guess, the question  is; Is it too costly to capitalize cause messages this way as a second layer of protection? 
Although it is not ideal, I think the cost is negligible and having another layer/assurance to catch uncapitalized messages this way is ok for pragmatic reasons.

balamurugana:
@ebozduman 

* If another PR is taking care of proper capitalized cause message, we don't need to capitalize the message in this PR.

* There is nothing costlier.  We have to fix it in right place.  Capitalizing blindly is not a good idea.  You could leave the message as is.

ebozduman:
@balamurugana 

OK.
- The other PR handles only the error messages, NOT the cause messages. To have cause messages reviewed and fixed, we have to have a separate issue filed to address inconsistencies in cause messages.
- I'll use this PR/Issue to remove the capitalization code in the logger.

nitisht (approved):
LGTM

kannappanr (commented):
LGTM

PR #5475, opened 2018-01-30
Quorum based listing for XL

### LINKED ISSUE #5380: Offline node make cluster ListObjects inconsistent when it's back online
1. Create 4 node cluster, 1 disk per node, wait for it all ready
2. mc cp 10000 small(10M) files to the cluster
3. On all 4 nodes, use mc ls local | wc -l to check files count, all return 10000 consistently
4. Bring down 1 node, then cp another 5000 files to the cluster
5. On the running 3 nodes, use mc ls local | wc -l to check files count, all return 15000 consistently
6. Bring the offline node back online, and execute mc admin heal on it (with and without --recursive)
7. On all 4 nodes, use mc ls local | wc -l to check files count

## Expected Behavior

All nodes should report 15000 consistently

## Current Behavior

All nodes report 10000 ~ 15000 randomly

## Possible Solution

None

## Context

This inconsistent behaviour is unaccpetable for production uses, especially for docker registry garbage collect procedure which will fail when any file is inaccessible, the inconsistent file list makes gc impossible to finish successfully.

## Your Environment

* Version used (`minio version`): RELEASE.2018-01-02T23-07-00Z
* Server type and version: Linux
* Operating System and version (`uname -a`): Linux controller 3.10.0-514.26.2.el7.x86_64 #1 SMP Mon Dec 4 15:49:59 CST 2017 x86_64 x86_64 GNU/Linux

krishnasrinivas:
fixes #5380

## Description

Previously for listing we would consult just one disk (which might have outdated contents), now we consult all the disks. This will make listing slow but correct.

## Motivation and Context

https://github.com/minio/minio/issues/5380

## How Has This Been Tested?

make test

## Types of changes

## Checklist:

### DIFF
--- a/cmd/xl-v1-list-objects.go
+++ b/cmd/xl-v1-list-objects.go
@@ -16,18 +16,24 @@
 
 package cmd
 
-import "github.com/minio/minio/pkg/errors"
+import (
+	"sort"
+
+	"github.com/minio/minio/pkg/errors"
+)
 
 // Returns function "listDir" of the type listDirFunc.
 // isLeaf - is used by listDir function to check if an entry is a leaf or non-leaf entry.
-// disks - used for doing disk.ListDir(). FS passes single disk argument, XL passes a list of disks.
+// disks - used for doing disk.ListDir()
 func listDirFactory(isLeaf isLeafFunc, treeWalkIgnoredErrs []error, disks ...StorageAPI) listDirFunc {
-	// listDir - lists all the entries at a given prefix and given entry in the prefix.
-	listDir := func(bucket, prefixDir, prefixEntry string) (entries []string, delayIsLeaf bool, err error) {
+	// Returns sorted merged entries from all the disks.
+	listDir := func(bucket, prefixDir, prefixEntry string) (mergedEntries []string, delayIsLeaf bool, err error) {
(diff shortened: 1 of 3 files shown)


cmd/xl-v1-list-objects.go
```diff
@@ -91,7 +110,7 @@ func (xl xlObjects) listObjects(bucket, prefix, marker, delimiter string, maxKey
 			objInfo, err = xl.getObjectInfo(bucket, entry)
 			if err != nil {
 				// Ignore errFileNotFound
```
harshavardhana: Comment should be fixed.. perhaps we can now use `isErrIgnored()` ?

krishnasrinivas: Actually, now, here only errXLReadQuorum should be ignored. Rest of the ignorable errors are taken care of inside listDir

krishnasrinivas: ummm, errFileNotFound should be ignored too because the object might have got deleted in the interim period of listing and getObjectInfo()

harshavardhana (approved):
LGTM

balamurugana (approved):
LGTM

PR #5597, opened 2018-03-01
check that the server supports encryption for SSE-C requests

### LINKED ISSUE #5511: Server should fail if SSE-C headers are provided but backend does not support encryption
## Expected Behavior
The server should fail if the backend does not support encryption *but* SSE-C headers are provided.

## Current Behavior
The server stores the object unencrypted.

## Possible Solution
Add check in object handler layer (S3 -layer)

1.  Disable encryption for e.g. FS backend.
2. Build and start server (with TLS) on FS
3.  `aws s3api put-object --no-verify-ssl --endpoint-url https://localhost:9000 --bucket my-bucket --key my-secret-diary --sse-customer-algorithm AES256 --sse-customer-key MzJieXRlc2xvbmdzZWNyZXRrZXltdXN0cHJvdmlkZWQ= --sse-customer-key-md5 7PpPLAK26ONlVUGOWlusfg== --body ~/my-diary.txt`

## Context
Encryption, SSE-C

## Your Environment
master

aead:
## Description

This change adds a check to verify that the server backend does support
encryption if a request contains SSE-C headers. If not the server
should fail - with Not Implemented - instead of given the impression
of e.g. encrypting the object while it actually stores the object in plaintext.

## Motivation and Context
Fixes #5511

## How Has This Been Tested?
manually - as described in the issue. 

## Types of changes

## Checklist:

### DIFF
--- a/cmd/object-handlers.go
+++ b/cmd/object-handlers.go
@@ -111,6 +111,10 @@ func (api objectAPIHandlers) GetObjectHandler(w http.ResponseWriter, r *http.Req
 		writeErrorResponse(w, s3Error, r.URL)
 		return
 	}
+	if IsSSECustomerRequest(r.Header) && !objectAPI.IsEncryptionSupported() { // SSE-C request but backend does not support encryption
+		writeErrorResponse(w, ErrNotImplemented, r.URL)
+		return
+	}
 
 	objInfo, err := objectAPI.GetObjectInfo(bucket, object)
 	if err != nil {
@@ -236,6 +240,10 @@ func (api objectAPIHandlers) HeadObjectHandler(w http.ResponseWriter, r *http.Re
 		writeErrorResponseHeadersOnly(w, s3Error)
 		return
 	}
+	if IsSSECustomerRequest(r.Header) && !objectAPI.IsEncryptionSupported() { // SSE-C request but backend does not support encryption
+		writeErrorResponse(w, ErrNotImplemented, r.URL)
+		return
+	}
 
 	objInfo, err := objectAPI.GetObjectInfo(bucket, object)
 	if err != nil {
@@ -334,6 +342,10 @@ func (api objectAPIHandlers) CopyObjectHandler(w http.ResponseWriter, r *http.Re
 
 	if s3Error := checkRequestAuthType(r, dstBucket, "s3:PutObject", globalServerConfig.GetRegion()); s3Error != ErrNone {
 		writeErrorResponse(w, s3Error, r.URL)
+		return
+	}
+	if (IsSSECustomerRequest(r.Header) || IsSSECopyCustomerRequest(r.Header)) && !objectAPI.IsEncryptionSupported() { // SSE-C request but backend does not support encryption
+		writeErrorResponse(w, ErrNotImplemented, r.URL)
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


harshavardhana (approved):
LGTM

harshavardhana:
One problem here is that @aead we will not be able to translate SSE headers back to gateway backends - how do you think we should address that?

aead:
> One problem here is that @aead we will not be able to translate SSE headers back to gateway backends - how do you think we should address that?

Do you mean that we cannot check whether the gateway supports SSE-C? Or that we cannot send SSE-C headers to gateway?

harshavardhana:
> Do you mean that we cannot check whether the gateway supports SSE-C? Or that we cannot send SSE-C headers to gateway?

IsObjectEncryptionEnabled() to true would mean that DARE will be created on gateway backend. But we want to perhaps do pass through for now as well. So rejecting in SSE-C headers are set would actually make it short handed? is what i think.

aead:
Okay, so it depends - if we want to terminate SSE-C on gateway, than everything is fine - backend has no clue that we store encrypted blobs. It does not know whether the client requested SSE-C.

If we want to do passthrough we have to think about proper solution...

harshavardhana:
> Okay, so it depends - if we want to terminate SSE-C on gateway, than everything is fine - backend has no clue that we store encrypted blobs. It does not know whether the client requested SSE-C.

That is why perhaps existing behavior is fine for now.

aead:
> That is why perhaps existing behavior is fine for now.

For a gateway implementation (e.g. S3 gateway) we can return always true for `objectAPI.IsEncryptionSupported()` - So handler-layer will always create DARE objects if SSE-C is provided. So basically **all** gateway backend implementations support encryption - since the server does it for them.

The only thing we have to decide is whether we want to do passthrough as well...

harshavardhana:
> For a gateway implementation (e.g. S3 gateway) we can return always true for objectAPI.IsEncryptionSupported() - So handler-layer will always create DARE objects if SSE-C is provided. So basically all gateway backend implementations support encryption - since the server does it for them.

Yes but doing this is something @abperiasamy doesn't want yet - perhaps you can ping him and ask. But i see that it could be a hindrance if we don't support pass through.

vadmeste (approved):
LGTM

aead:
>Yes but doing this is something @abperiasamy doesn't want yet - perhaps you can ping him and ask. But i see that it could be a hindrance if we don't support pass through.

I'll ask him. ~But the point is: at the moment we do this already - So AFAIKS if someone starts gateway and sends SSE-C requests we create DARE objects.~ So the only thing we have to add is a check whether we want to do pass-through (set by ENV_VAR or ....) and if so delegate directly to backend implementation.

EDIT: I see what you mean will check.

aead:
I put this on hold until we have a decision.

balamurugana:
## Mint Automation
| Test | Result |
|:--- |:--- |

harshavardhana (changes requested):
-1 until this PR gets unblocked

harshavardhana:
We should close this @aead and re-send if we decide to do. Since current implementation is expected.

aead:
@harshavardhana Agree, closing.

PR #5658, opened 2018-03-15
[refactor] simplify en/decrypted size computation

### LINKED ISSUE #5657: [refactor]: simplify en/decrypted object size computation
## Description

Since [sio](https://github.com/minio/sio) provides functionality to compute the size of encrypted/decrypted data streams we can simplify the current encryption logic in the server.   

This includes:
 - [decryptedSize](https://github.com/minio/minio/blob/master/cmd/encryption-v1.go#L737)
 - [EncryptedSize](https://github.com/minio/minio/blob/master/cmd/encryption-v1.go#L766)

And all constants / tests used for these functions.

## Context
SSE-C, encryption, refactor

## Your Environment
master

aead:
## Description

This commit replaces the en/decrypted size computation
with functions from the `sio` package.

## Motivation and Context
Fixes #5657

## How Has This Been Tested?
unit tests

## Types of changes

## Checklist:

### DIFF
--- a/cmd/encryption-v1.go
+++ b/cmd/encryption-v1.go
@@ -623,9 +623,9 @@ func DecryptBlocksRequest(client io.Writer, r *http.Request, startOffset, length
 	var partStartOffset = startOffset
 	// Skip parts until final offset maps to a particular part offset.
 	for i, part := range objInfo.Parts {
-		decryptedSize, err := decryptedSize(part.Size)
+		decryptedSize, err := sio.DecryptedSize(uint64(part.Size))
 		if err != nil {
-			return nil, -1, -1, err
+			return nil, -1, -1, errObjectTampered
 		}
 
 		partStartIndex = i
@@ -633,12 +633,12 @@ func DecryptBlocksRequest(client io.Writer, r *http.Request, startOffset, length
 		// Offset is smaller than size we have reached the
 		// proper part offset, break out we start from
 		// this part index.
-		if partStartOffset < decryptedSize {
+		if partStartOffset < int64(decryptedSize) {
 			break
 		}
 
 		// Continue to look for next part.
-		partStartOffset -= decryptedSize
+		partStartOffset -= int64(decryptedSize)
 	}
 
 	startSeqNum := partStartOffset / sseDAREPackageBlockSize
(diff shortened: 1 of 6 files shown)


balamurugana:
## Mint Automation
| Test | Result |
|:--- |:--- |

cmd/encryption-v1.go
```diff
@@ -756,19 +742,22 @@ func (o *ObjectInfo) DecryptedSize() (int64, error) {
...
 // EncryptedSize returns the size of the object after encryption.
 // An encrypted object is always larger than a plain object
 // except for zero size objects.
 func (o *ObjectInfo) EncryptedSize() int64 {
-	size := (o.Size / sseDAREPackageBlockSize) * (sseDAREPackageBlockSize + sseDAREPackageMetaSize)
-	if mod := o.Size % (sseDAREPackageBlockSize); mod > 0 {
-		size += mod + sseDAREPackageMetaSize
+	size, err := sio.EncryptedSize(uint64(o.Size))
+	if err != nil {
+		panic(err) // Since AWS S3 allows parts to be 5GB at most this cannot happen - sio max. size is 256 TB
```
vadmeste, line 758: Looks like the server will panic if data in the backend are corrupted. In general we don't like the server to crash. I think it makes sense to also return an error here and make callers check the result. I checked EncryptedSize() is called only few times.

harshavardhana: Server will not crash only the goroutine to the handler crashes. Go http server survives recoverable panics. This panic will only occur at 256TB and above since it is unlikely a panic here is okay.

harshavardhana: Unlikely with the sense that we have other checks to prevent object > 5TB

vadmeste: oh so that's the only case when EncryptedSize returns an error then

vadmeste (approved):
LGTM

cmd/encryption-v1.go
```diff
@@ -623,22 +623,22 @@ func DecryptBlocksRequest(client io.Writer, r *http.Request, startOffset, length
 	var partStartOffset = startOffset
 	// Skip parts until final offset maps to a particular part offset.
 	for i, part := range objInfo.Parts {
-		decryptedSize, err := decryptedSize(part.Size)
+		decryptedSize, err := sio.DecryptedSize(uint64(part.Size))
 		if err != nil {
-			return nil, -1, -1, err
+			return nil, -1, -1, errObjectTampered
```
nitisht, line 628: Not sure if its okay to assume `errObjectTampered` for all `err` thrown by `DecryptedSize`?

aead: It is fine - `DecryptedSize` can fail for two reasons:
 1. The size is greater than `2^48 + 2^37` which cannot happen if the object is a valid S3 object. `2^48 + 2^37 > 256 TB` - S3 allows only 5 TB
 2. The size would imply a last package which is smaller than 33 bytes. That cannot happen if the object is a valid encrypt object.

-> The object is accidentally/maliciously tampered.

nitisht (approved):

PR #5920, opened 2018-05-10
fix: introduce isLeafDir in healing to fix the crash

### LINKED ISSUE #5917: Minio server crashes when healing recursively
## Expected Behavior
Calling `mc admin heal -r myminio` should not crash the server

## Current Behavior
Server exits with the following message:

```
API: Heal()
Time: 09:55:22 UTC 05/10/2018
RemoteHost: 10.255.0.2
Error: disk not found

       1: cmd/xl-v1-healing.go:111:cmd.healBucket()
       2: cmd/xl-v1-healing.go:82:cmd.xlObjects.HealBucket()
       3: cmd/xl-sets.go:1184:cmd.(*xlSets).HealBucket()
       4: cmd/admin-heal-ops.go:597:cmd.(*healSequence).healBucket()
       5: cmd/admin-heal-ops.go:577:cmd.(*healSequence).healBuckets()
       6: cmd/admin-heal-ops.go:523:cmd.healBuckets)-fm()
       7: cmd/admin-heal-ops.go:516:cmd.(*healSequence).traverseAndHeal.func1()
       8: cmd/admin-heal-ops.go:523:cmd.(*healSequence).traverseAndHeal()
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xc0e063]

goroutine 614 [running]:
github.com/minio/minio/cmd.doTreeWalk(0x121d300, 0xc420182f00, 0xc420502a4a, 0x5, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
        /q/.q/sources/gopath/src/github.com/minio/minio/cmd/tree-walk.go:186 +0x873
github.com/minio/minio/cmd.startTreeWalk.func1(0x121d300, 0xc420182f00, 0xc420502a4a, 0x5, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
        /q/.q/sources/gopath/src/github.com/minio/minio/cmd/tree-walk.go:259 +0x10a
created by github.com/minio/minio/cmd.startTreeWalk
        /q/.q/sources/gopath/src/github.com/minio/minio/cmd/tree-walk.go:257 +0x1c3

```

1. Using Docker swarm deploy 2 services with 4 exports each in distributed mode
2. Remove the content from 2 exports inside a single service
3. Use `mc` to heal the missing 2 exports
4. Minio server crashes

## Context
I was trying to move half of the exports to a second server and mount it using NFS, but when I tried to issue the `heal` command the server crashed, then I rolled back to using local disks, removed the same content locally and tried again, with the same crash result (meaning that NFS played no role in the issue).

The server has around 150MB of random files I had laying around on my personal computer, one folder has 530 files.

## Your Environment
```
Version: 2018-05-10T00:00:42Z
Release-Tag: RELEASE.2018-05-10T00-00-42Z
Commit-ID: cc8178cdc4096ebf66fc2f467fd8f3be9ff1b99b
```
* Docker version: 
```
Client:
 Version:       17.12.1-ce
 API version:   1.35
 Go version:    go1.9.4
 Git commit:    3dfb8343b139d6342acfd9975d7f1068b5b1c3d3
 Built: Tue Apr  3 23:37:44 2018
 OS/Arch:       linux/amd64

Server:
 Engine:
  Version:      17.12.1-ce
  API version:  1.35 (minimum version 1.12)
  Go version:   go1.9.4
  Git commit:   7390fc6/17.12.1-ce
  Built:        Tue Apr  3 23:38:52 2018
  OS/Arch:      linux/amd64
  Experimental: false

```
* Environment name and version: Minio was the only thing running inside a docker swarm environment
* Server type and version: Amazon default Linux AMI on a micro instance: 
* Operating System and version (`uname -a`): `Linux 4.14.33-51.37.amzn1.x86_64 #1 SMP Thu May 3 20:07:43 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux`
* Link to your project: [docker-swarm.yml](https://gist.github.com/rafaelsierra/eab0e3d06cfcdfb169a1179a2022371c) being used with `docker stack deploy`

harshavardhana:
## Description
fix: introduce isLeafDir in healing to fix the crash

This PR also supports healing directories.

## Motivation and Context
Fixes #5917

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:

### DIFF
--- a/cmd/tree-walk.go
+++ b/cmd/tree-walk.go
@@ -153,6 +153,12 @@ func doTreeWalk(ctx context.Context, bucket, prefixDir, entryPrefixMatch, marker
 			return err
 		}
 	}
+
+	// When isleaf check is delayed, make sure that it is set correctly here.
+	if delayIsLeaf && isLeaf == nil {
+		return errInvalidArgument
+	}
+
 	// For an empty list return right here.
 	if len(entries) == 0 {
 		return nil
@@ -169,6 +175,7 @@ func doTreeWalk(ctx context.Context, bucket, prefixDir, entryPrefixMatch, marker
 	if len(entries) == 0 {
 		return nil
 	}
+
 	for i, entry := range entries {
 		var leaf, leafDir bool
 
@@ -187,7 +194,6 @@ func doTreeWalk(ctx context.Context, bucket, prefixDir, entryPrefixMatch, marker
 		}
 
 		isDir := !leafDir && !leaf
-
 		if i == 0 && markerDir == entry {
 			if !recursive {
 				// Skip as the marker would already be listed in the previous listing.
--- a/cmd/xl-v1-healing.go
+++ b/cmd/xl-v1-healing.go
@@ -539,12 +539,74 @@ func healObject(ctx context.Context, storageDisks []StorageAPI, bucket string, o
 	return result, nil
 }
 
+// healObjectDir - heals object directory specifically, this special call
+// is needed since we do not have a special backend format for directories.
+func (xl xlObjects) healObjectDir(ctx context.Context, bucket, object string, dryRun bool) (hr madmin.HealResultItem, err error) {
+	storageDisks := xl.getDisks()
+
+	// Initialize heal result object
+	hr = madmin.HealResultItem{
+		Type:         madmin.HealItemObject,
+		Bucket:       bucket,
+		Object:       object,
+		DiskCount:    len(storageDisks),
+		ParityBlocks: len(storageDisks) / 2,
+		DataBlocks:   len(storageDisks) / 2,
+		ObjectSize:   0,
+	}
+
+	// Prepare object creation in all disks
+	for _, disk := range storageDisks {
+		if disk == nil {
+			hr.Before.Drives = append(hr.Before.Drives, madmin.HealDriveInfo{
+				UUID:  "",
+				State: madmin.DriveStateOffline,
+			})
+			hr.After.Drives = append(hr.After.Drives, madmin.HealDriveInfo{
+				UUID:  "",
+				State: madmin.DriveStateMissing,
+			})
+			continue
+		}
+
+		drive := disk.String()
+		hr.Before.Drives = append(hr.Before.Drives, madmin.HealDriveInfo{
+			UUID:     "",
+			Endpoint: drive,
+			State:    madmin.DriveStateMissing,
+		})
+		hr.After.Drives = append(hr.After.Drives, madmin.HealDriveInfo{
+			UUID:     "",
+			Endpoint: drive,
+			State:    madmin.DriveStateMissing,
+		})
+
+		if !dryRun {
+			if err := disk.MakeVol(pathJoin(bucket, object)); err != nil && err != errVolumeExists {
+				return hr, toObjectErr(err, bucket, object)
+			}
+
+			for i, v := range hr.Before.Drives {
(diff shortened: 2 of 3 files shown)


cmd/xl-v1-healing.go
```diff
@@ -539,12 +539,72 @@ func healObject(ctx context.Context, storageDisks []StorageAPI, bucket string, o
...
+
 // HealObject - heal the given object.
 //
 // FIXME: If an object object was deleted and one disk was down,
 // and later the disk comes back up again, heal on the object
 // should delete it.
 func (xl xlObjects) HealObject(ctx context.Context, bucket, object string, dryRun bool) (hr madmin.HealResultItem, err error) {
+	// Healing directories handle it separately.
+	if hasSuffix(object, slashSeparator) {
+		return xl.healObjectDir(ctx, bucket, object)
```
vadmeste: This needs a dry-run parameter

cmd/tree-walk.go
```diff
@@ -183,11 +184,13 @@ func doTreeWalk(ctx context.Context, bucket, prefixDir, entryPrefixMatch, marker
 		}
 
 		if strings.HasSuffix(entry, slashSeparator) {
-			leafDir = isLeafDir(bucket, pathJoin(prefixDir, entry))
+			leafDir = true
```
vadmeste: 'leafDir' should be false by default, so 'isDir' would be able to be set to true if 'entry' has a trailing slash.

harshavardhana: that won't work for empty directories.. The reason to set this value is because we are considering leafDir is true for all trailing slash.  isLeafDir sets it to false otherwise.

cmd/tree-walk.go
```diff
@@ -169,11 +169,12 @@ func doTreeWalk(ctx context.Context, bucket, prefixDir, entryPrefixMatch, marker
 	if len(entries) == 0 {
 		return nil
 	}
+
 	for i, entry := range entries {
 		var leaf, leafDir bool
 
 		// Decision to do isLeaf check was pushed from listDir() to here.
-		if delayIsLeaf {
+		if delayIsLeaf && isLeaf != nil {
```
vadmeste: Can you also add similar check for leafDir in line 186 ? it makes sense since doTreeWalk is called from many places

harshavardhana: We can but that function is never called... by healing code.

vadmeste (approved):
LGTM

cmd/tree-walk.go
krishnasrinivas: this change is not correct, because it gives an impression to the reader that delayIsLeaf can be true and isLeaf can be nil (in which case the code will not strip the trailing `/` and not work as expected)

Hence it's better to leave it as before.

Or, do
```
if delayIsLeaf {
    if isLeaf == nil { return error }
```

harshavardhana: yes, it can be @krishnasrinivas isLeaf being nil is not a wrong case. It is just that if delayIsLeaf is true isLeaf should be better be set otherwise we will crash. But not having a value for isLeaf is also true in case of healing code. It provides rather more meaning. In-fact previous code used to be 

```
if delayIsLeaf && isLeaf(entry) {
```

krishnasrinivas (approved):

PR #5926, opened 2018-05-12
Better error message for the PartSizeUnequal error.

thomasetter:
Only the standalone single directory mode emits the PartSizeUnequal error and is therefore incompatible with some client libraries (e.g. https://github.com/hubrick/vertx-s3-client).

The new error message makes this clearer and avoids deterring potential users because of perceived incompatibilities.

### DIFF
--- a/cmd/object-api-errors.go
+++ b/cmd/object-api-errors.go
@@ -324,7 +324,8 @@ func (e InvalidPart) Error() string {
 type PartsSizeUnequal struct{}
 
 func (e PartsSizeUnequal) Error() string {
-	return "All parts except the last part should be of the same size"
+	return "All parts except the last part should be of the same size. If you see this, you are probably using the standalone single directory mode. " +
+		"You can workaround this by specifying four data directories instead of one."
 }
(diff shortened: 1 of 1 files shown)


krishnasrinivas (commented):
For a user who wants to use FS backend this workaround will not be useful.
It is not right to put such details in an error string.  (there are exceptions, but not this one) 
Previous error message is accurate and good enough to tell the user about the problem.

thomasetter:
When I saw this error message as a user, I assumed that this is a general limitation of Minio so there are some users for whom the current error message is not sufficient.
If I had not found the workaround by writing to the Minio Slack channel, I would have switched to other solutions because I was under the impression that S3 compatibility was no longer a concern for the project.

harshavardhana:
@krishnasrinivas we can discuss this with @abperiasamy and move forward?

harshavardhana:
We are planning to revert back changes for this error anyways, so this https://github.com/minio/minio/pull/6183 supersedes this PR. Closing.

PR #5963, opened 2018-05-20
Implement auto cert reloading

harshavardhana:
## Description
Originating from the idea in #5610 by @ewdurbin

## Motivation and Context
Deployment simplicity

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:


kannappanr:
ping @donatello and @vadmeste

cmd/signals.go
```diff
@@ -76,13 +79,11 @@ func handleSignals() {
 				// Ignore this at the moment.
 			case serviceRestart:
 				logger.Info("Restarting on service signal")
-				err := globalHTTPServer.Shutdown()
-				logger.LogIf(context.Background(), err)
 				stopHTTPTrace()
+				stop := stopProcess()
```
vadmeste, line 83: I would say it is better to include `globalHTTPServer.Shutdown()` in `restartProcess()` and avoid adding this line (stopProcess + restartProcess doesn't make sense)

pkg/certs/certs.go
```diff
@@ -0,0 +1,120 @@
...
+func (c *Certs) watch() (err error) {
+	if err = notify.Watch(c.certFile, c.e, eventWrite...); err != nil {
+		return err
+	}
+
+	if err = notify.Watch(c.keyFile, c.e, eventWrite...); err != nil {
+		return err
+	}
+	c.Lock()
+	c.cert, err = c.loadCert(c.certFile, c.keyFile)
```
vadmeste, line 85: I think loading certificate should be run before watching to quit as early as possible.

vadmeste (dismissed):
Asking for minor changes

pkg/certs/certs.go
donatello: In addition to loading the certificate before watching, this function can automatically call `notify.Stop()` (in case one of the Watch calls succeeds before an error is encountered). This would make the certs package a bit safer/easier to use (e.g. user of this module need not call Stop in case of errors).

donatello (dismissed):

pkg/certs/certs.go
harshavardhana: if watch returns an error there is no need to notify.Stop() since the returned value of `certs` is nil anyway. The application is dead and all watches for that application have been relinquished automatically by the kernel.

harshavardhana: Well watch will also fail if the files don't exist @vadmeste so the end result is same.

harshavardhana: even if the files are corrupted this function simply returns an error during New() so subsequent caller never has to do notify.Stop()

harshavardhana: The reason loading is done later is also to ensure that watching is already present and we load the cert right before we look for new events. Since some events can be buffered it is to avoid any races that may occur due to the timing.

harshavardhana: Added notify.Stop() return loadCert error to cleanup any on-going watches.

vadmeste (dismissed):
LGTM & tested

pkg/certs/certs.go
donatello: Well it is fine and works for now. However, we need something like:

```
defer func() {
    if (err != nil) {
        notify.Stop(c.e)
    }
}
```

at the beginning of the function to make it properly correct. Though the certs package is being used like this today, a future change may not exit the application in case of error - this might lead to a leaking watch. The certs package should not require/assume such behaviour on the application IMO.

harshavardhana:
updated @vadmeste @donatello PTAL

donatello (approved):

vadmeste (approved):
LGTM & tested

Issue #5305, opened 2017-12-18
Multiple access and secret key pairs


mvrhov:
## Expected Behavior
Have the ability to specif multiple Access/secret key pairs.

## Current Behavior
Only one access and secret key available.

## Context
We would like to give the access to the bucket to multiple 3rd parties. Each of them should have it's own pair that we can disable if he starts to abuse the storage access

nitisht:
@mvrhov We are going to implement https://github.com/minio/mfs can you please take a look there

harshavardhana:
MSF is a minio federation project currently under progress, which would implement support for multi access keys as one of the feature.

mvrhov:
Thanks for letting me know. Any chance this remains open until this is implemented. So I can automatically get a notification.

nitisht:
@mvrhov you can just _watch_ https://github.com/minio/mfs for notifications on that project. IMO we can close this issue.

mvrhov:
Yeah, but watching the project pollutes my inbox

nitisht:
> Yeah, but watching the project pollutes my inbox

@mvrhov then creating an issue here https://github.com/minio/mfs will be better.

nogweii:
Since msf is marked as deprecated and it's ideas incorporated into minio itself (rather than existing separately) should this issue be re-opened? minio/msf#2 hasn't been marked resolved yet.

harshavardhana:
Yes good thinking @evaryont

erma07:
When will this feature be implemented? Any chance credential pair permissions could also be defined per bucket?

tnatanael:
Any news?

harshavardhana:
> Any news?

This is in future @thiagonatanael @erkimarks currently we are working on other priorities.

tnatanael:
Sure, really thx! The software is awesone!

craSH:
Hello, this is a feature that I would like / need before I can roll out Minio for some internal uses (which are all inherently multi-user) at my company, please consider this an additional request to prioritize the feature. Thanks!

diogojesus95:
+1 We also need this to implement private buckets over different co-workers for internal use

devinmartin:
I can certainly appreciate keeping Minio a "single responsibility" micro-service. Minio is responsible in my mind for storage durability and availability.

[...]

Issue #5199, opened 2017-11-17
What's the point of multiple buckets without multiple access keys?


lflare:
As I have mentioned in the title, what is the point of allowing for multiple buckets when we are not able to use multiple access/secret keys for each different bucket? Most services/projects I have seen tend towards using only one bucket for all purposes, yet to use the same access/secret keys for multiple projects is also a security disaster waiting to happen.

Before anyone tells me to spin up multiple minio instances to handle multi-identities, I know that is a solution but it is also a shoddy solution. If Minio is modeled after the S3 API, at the very least, maybe allow for multiple secret keys with the same access keys? I should not be needed to spin up multiple instances of Minio to handle multi-identity, when the overhead of doing so is greater than simply allowing for multiple secret keys to access different buckets.

Back to the title, what's the point of multiple buckets without multiple access keys?

krishnasrinivas:
> Back to the title, what's the point of multiple buckets without multiple access keys?

Data organisation. Almost all of Minio's users have multiple buckets. Most of software that use AWS-S3 (for example cloudberry backup client) make ListBucket/MakeBucket calls and we need to support it.

Coming to supporting multiple access/secret keys, we have not had many requests for this. We have had discussions on this, however we can not commit anything at this point.

lflare:
Hello,

Thanks for replying fast. I have some further questions regarding this issue,

> Almost all of Minio's users have multiple buckets.

How does the average user go about segregating access controls to prevent one compromised access key from completely killing their Minio's security? I assume that each bucket is for each different service, how do people sleep at night knowing that if a single service is compromised, all their buckets are equally as compromised?

> Coming to supporting multiple access/secret keys, we have not had many requests for this. We have had discussions on this, however we can not commit anything at this point.

I would've thought this would be quite an important issue that security people would be flocking about but perhaps I'm wrong?

krishnasrinivas:
> How does the average user go about segregating access controls to prevent one compromised access key from completely killing their Minio's security? I assume that each bucket is for each different service, how do people sleep at night knowing that if a single service is compromised, all their buckets are equally as compromised?

We advise one minio instance per tenant. Different tenants can have different access/secret keys. Minio integrates well with micro-service/container architecture. Hence spinning up new instances is easy.

Some documents:
https://docs.minio.io/docs/multi-tenant-minio-deployment-guide
https://github.com/krishnasrinivas/cookbook/blob/multi-tenancy/docs/multi-tenancy-in-minio.md

[...]

Issue #4186, opened 2017-04-27
Allow support for multiple credentials for the minio server


efarrer:
Minio should be updated to support multiple sets of credentials.

## Expected Behavior
It would be nice if a minio cluster could support multiple sets of keys (credentials). This would allow for access/secret keys to be rotated without downtime as follows:
1. New keys are **added** to individual nodes in the cluster and each node is restarted (or the configuration hot reloaded).
2. Clients are updated to use the new keys.
3. The old keys are removed.

## Current Behavior
Only one set of keys are supported

## Possible Solution
Two potential solutions:
  1. accessKeyNext and secretKeyNext along with relevant env vars will be supported. Either the original credentials or the *Next credentials can be used to authenticate.

  2. An unlimited number of credentials could be added via a new configuration key "credentials" who's value is a list of objects that each have "accessKey" and "secretKey" entries. Sequentially numbered env variables such as MINIO_ACCESS_KEY0, MINIO_SECRET_KEY0, MINIO_ACCESS_KEY1, MINIO_SECRET_KEY1 will be also be supported (without numeric gaps for simplicity).

Solution number 1 satisfies the support for key rotation, but number 2 would allow greater flexibility (like having different users have different keys and being able to revoke keys), but adds more complexity and would make the "credential" key in the config.json obsolete/deprecated.

## Context
  Company tries to rotate credentials frequently to minimize the risk of leaked credentials. This would help us rotate credentials without downtime.

## Your Environment
* Version used (`minio version`): Version: 2017-04-27T06:30:39Z

deekoder:
We are studying amazon standards to implement this comprehensively including rotating credentials/password. I am not able to give you an ETA on this. But we are very much on getting this in our future release.

checketts:
@harshavardhana I see you moved this to `Completed` on https://github.com/minio/minio/projects/7#card-4102146 does that mean this feature is implemented?

harshavardhana:
https://github.com/minio/federator is where the project has moved .

deekoder:
Closing this as we will handle these as part of the federator project

checketts:
As an aside, I wish there was a simple non-federator way of rotating credentials. I have several projects that were setup with hardcoded keys (to use in a docker compose like setup) But unfortunately without being able to set a secondary key I can't rotate the dev credentials.

[...]

Issue #2998, opened 2016-10-19
server: Support multiple api keys


theredcat:
Is minio going to implement multiple accesskey?

I am aware of #811, but minio-xl repo doesn't exsits anymore.
Also aware of the workaround using multi tenant as described in #2751 (which will be fixed in docs by #2752 ), but this approach need externals tools to be implemented.

Do you plan to implement such a feature in minio (even in a distant future) ?

harshavardhana:
> Do you plan to implement such a feature in minio (even in a distant future) ?

We will not be implementing multiple access key support. #2752 is what holds good for the future as well.

pdf:
@harshavardhana multiple instances do not solve the same problems as multiple keys, unless multiple instances are safe to point at the same data (the multi-tenant docs appear to suggest otherwise)?  For example, compromised keys require rolling credentials on every client, or the need for differing access levels for shared data.

If it is not possible to point multiple instances at the same data (which is in any case quite painful to maintain), is there something technical (rather than philosophical) that makes it particularly difficult to support multiple keys?

PR #5968, opened 2018-05-22
Per-bucket access keys

evangreen:
This change enables support for having a separate access key/secret per bucket.

## Motivation
Minio is quite useful, but it's challenging to properly manage bucket access when there is only a single root-level API access key. Handing out the all-access-pass master key to users is risky, as it invites abuse or mistakes from any actor using the service. Users have also been asking for such a feature, as in #5199, #5305, #4186, #2998, #811, and #2751. Responses to these requests have usually been a "no" with responses such as "this feature doesn't fit in the microservices architecture", "it's in the federated service", or "spin up multiple instances instead". I believe that with only minor changes to the overall architecture it's possible to add support for multiple keys, in a way that doesn't bloat the code or ruin the microservice flavor of minio.

## Architecture
 * The primary set of master credentials is still used for administrative backend tasks, RPCs, and as an all-access key to all buckets.
 * In addition, a single key/secret can be set for each bucket. This key can be used on the web, in v2 and v4 requests, and with the streaming API. The per-bucket keys cannot enumerate the contents or existence of other buckets. The per-bucket key is currently stored in the main config.json file. 
 * The per-bucket key cannot be used to delete the bucket or change the key.
 * For the purposes of ListBuckets, if the same API key is set for multiple buckets, then ListBuckets shows all buckets that correspond to the key used.

While this might not satisfy every possible use case, I believe it will be useful for a large population of users asking for this feature. Or at least me :)

## Outstanding issues
 * Currently only one key per bucket can be stored. This could probably be expanded with a little more effort, and some UI changes.
 * I've hijacked the password change web UI since I'm no expert in web stuff. The password change UI now controls the per-bucket key, leaving no way to change the master key/secret. This could probably be tweaked with a minimal UI change (eg a switch saying "Change per-bucket key", or maybe two buttons instead of one for "Update Bucket Key" and "Update Master Key").
 * No unit tests.
 * No documentation.

## How Has This Been Tested?
I've tested this on my local machine manually, with the web UI and the "mc" tool.

## Types of changes
- New feature (non-breaking change which adds functionality)

## Checklist:


harshavardhana:
> While this might not satisfy every possible use case, I believe it will be useful for a large population of users asking for this feature. Or at least me :)

@evangreen thanks for the commendable effort really appreciate the time you spent implementing this feature. To start with let me explain a few things here for more clarity

- Our implementation is going to diverge from per bucket level access keys. Minio is designed on purpose for a single tenant model i.e single access/secret key and this PR conflicts with our design model.
- We are planning to bring external identity providers like OpenID, SAML etc this would provide temporary credentials (Amazon STS) for each user but such a user conceptually doesn't exist on Minio and Minio doesn't manager users. Access keys are simply temporary and expire.  (this still keeps Minio in line with our single tenant model)

We will take this PR into consideration discuss this with the team and see how we can fit this into our design model. But for now, multiple keys per bucket is going to complicate our simple behavior.

krishnasrinivas:
@evangreen :1st_place_medal: for the effort

evangreen:
Thanks for taking a look at this PR. I'll be sure to check out your external identity provider hookups once they're up on Github. Sounds like this is a no for now, which is admittedly disappointing. If you do decide to take some or all of this later, I'll be around. Thanks again for making minio!

guanzo:
@harshavardhana Any idea when the identity feature will be released?

pruiz:
It is a shame this pull-req has been rejected, as on somewhat large deployments (with many nodes, using erasure, etc.), it is not easy to spin multiple instances on each cluster-node w/o a bit of deployment & maintenance hell.

Being able to have a single minio deployment with per-bucket keys would benefit allot in those cases.. :(

harshavardhana:
It has not been rejected it is being discussed,  what was stated above are design challenges. We will keep everyone posted.

harshavardhana:
@evangreen I have been trying to reach out to you regarding this PR. Would you be able to spend some time on https://slack.minio.io to discuss this? . Feel free to ping me at 'y4m4'.

evangreen:
Whoops, I'm sorry. I was travelling when I received your last message, and lost of track of it. I've connected now on LinkedIn, so I'm happy to talk here or there.

krishnasrinivas:
We are going to support multi access/secret keys supporting Client-Credentials-Flow of OpenID. Here is the rough idea on how it will work https://github.com/aarushiarya/minio-idp/blob/master/README.md

Implementation is underway.

evangreen:
Will I be able to specify a simple text file of keys and buckets, without having to set up additional services? One of the things I love about minio is that it's got everything out of the box in a single binary.

krishnasrinivas:
Not sure how it will look in the end.

Different implementation options for us without any additional IDP services:
1) `minio server --iam-policy policy.json /mnt/export`

policy.json might look like this:
```
{
  "version": 1,
  "users": [
    {
      "accessKey": "user1",
      "secretKey": "secret1",
      "policy": {
	"Version": "2012-10-17",
	"Statement": [
	  {
	    "Sid": "Stmt1530559274230",
	    "Action": [
              "s3:CreateBucket",
              "s3:DeleteBucket",
              "s3:DeleteObject",
              "s3:GetObject",
              "s3:PutObject"
	    ],
	    "Effect": "Allow",
	    "Resource": "arn:aws:s3:::user1bucket"
	  }
	]
      }
    },
    {
      "accessKey": "user2",
      "secretKey": "secret2",
      "policy": {
	"Version": "2012-10-17",
	"Statement": [
	  {
	    "Sid": "Stmt1530559274231",
	    "Action": [
              "s3:CreateBucket",
              "s3:DeleteBucket",
              "s3:DeleteObject",
              "s3:GetObject",
              "s3:PutObject"
	    ],
	    "Effect": "Allow",
	    "Resource": "arn:aws:s3:::user2bucket"
	  },
	  {
	    "Sid": "Stmt1530559274232",
	    "Action": [
              "s3:GetObject"
	    ],
	    "Effect": "Allow",
	    "Resource": "arn:aws:s3:::user1bucket"  <-- user2 can do GetObject on user1's bucket
	  },
	]
      }
    }
  ]
}
```

the individual policies here were generated by https://awspolicygen.s3.amazonaws.com/policygen.html Choose "IAM Policy" from the dropdown there.

2) the policy.json above can be embedded in config.json itself (downside is every time a user is added deleted we have to touch config.json)

3) We might have an "~/.minio/iam/" directory containing a number of user policy files. One policy file per user.

harshavardhana:
Closing this PR we have first draft of the changes in - there is still more work that needs to be done but everyone can follow the PR here https://github.com/minio/minio/pull/6176

PR #5987, opened 2018-05-29
`Minio Gateway` should expose `Prometheus` metrics similarly to `Mini…

yskopets:
…o Server`

## Description

`Minio Gateway` should expose `Prometheus` metrics similarly to `Minio Server`.

## Motivation and Context

Having `Prometheus` endpoint is important when `Minio Gateway` is deployed on `Kubernetes` (so that it could be monitored consistently with the other infrastructure).

## How Has This Been Tested?

Manual testing

## Types of changes

## Checklist:

- [x ] My code follows the code style of this project.

### DIFF
--- a/cmd/metrics.go
+++ b/cmd/metrics.go
@@ -64,23 +64,7 @@ func (c *minioCollector) Describe(ch chan<- *prometheus.Desc) {
 // Collect is called by the Prometheus registry when collecting metrics.
 func (c *minioCollector) Collect(ch chan<- prometheus.Metric) {
 
-	// Fetch disk space info
-	objLayer := newObjectLayerFn()
-	// Service not initialized yet
-	if objLayer == nil {
-		return
-	}
-	s := objLayer.StorageInfo(context.Background())
-
-	var totalDisks, offlineDisks int
-	// Setting totalDisks to 1 and offlineDisks to 0 in FS mode
-	if s.Backend.Type == FS {
-		totalDisks = 1
-		offlineDisks = 0
-	} else {
-		offlineDisks = s.Backend.OfflineDisks
-		totalDisks = s.Backend.OfflineDisks + s.Backend.OnlineDisks
-	}
+	// Always expose network stats
 
 	// Network Sent/Received Bytes
 	ch <- prometheus.MustNewConstMetric(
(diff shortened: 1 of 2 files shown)


cmd/metrics.go
```diff
@@ -100,6 +101,11 @@ func (c *minioCollector) Collect(ch chan<- prometheus.Metric) {
 		float64(globalConnStats.getTotalInputBytes()),
 	)
 
+	// don't expose disk-related metrics if an ObjectLayer doesn't provide respective info, e.g. gateways
+	if reflect.DeepEqual(s, StorageInfo{}) {
```
harshavardhana: We shouldn't do this inside `Collect` instead simply look for s.Backend.Type

harshavardhana (changes requested):

cmd/metrics.go
yskopets: Sorry, I don't get it.

* Should I look for `s.Backend.Type` inside `Collect` ?
* If I start dropping disk metrics when, let's say, `if s.Backend.Type == Unknown`, then I'll make the following `else` branch inaccessible (see below). Is it ok ?

```
	var totalDisks, offlineDisks int
	// Setting totalDisks to 1 and offlineDisks to 0 in FS mode
	if s.Backend.Type == FS {
                ...
	} else {
		offlineDisks = s.Backend.OfflineDisks
		totalDisks = s.Backend.OfflineDisks + s.Backend.OnlineDisks
	}
```

In other words, the idea behind `if reflect.DeepEqual(s, StorageInfo{})` was to have a check equivalent to `if s == nil`. Which would preserve existing behaviour in a case where `s.Backend.Type == Unknown` together with `s.Backend.OfflineDisks > 0` or `s.Backend.OnlineDisks > 0`.

harshavardhana: The problem is the way all the metrics are listed . All networking related ones should come before the StorageInfo() since they are available always. 

There is no set order so you can make those changes ..

harshavardhana (approved):
LGTM

nitisht (approved):

nitisht:
Thanks for the PR @yskopets , great addition! LGTM and tested

yskopets:
Thank you!

PR #6028, opened 2018-06-07
Revert "NewUX: On filename click, select file instead of downloading"

deekoder:
Reverts minio/minio#6023
Needs more in-depth discussion on how we can do this and allow downloading a folder as a zip and select for batch operations.


kaankabalak (approved):

harshavardhana:
@kaankabalak to send new PR with changes, closed this PR.

PR #6033, opened 2018-06-11
limit memory allocation during multiple object deletion

aead:
## Description
This commit limits the amount of memory allocated by the
S3 Multi-Object-Delete-API. The server used to allocate as
many bytes as provided by the client using Content-Length.

S3 specifies that the S3 Multi-Object-Delete-API can delete
at most 1000 objects using a single request.
(See: https://docs.aws.amazon.com/AmazonS3/latest/API/multiobjectdeleteapi.html)
Since the maximum S3 object name is limited to 1024 bytes the
XML body sent by the client can only contain up to 1000 * 1024
bytes (excluding XML format overhead).

This commit limits the size of the parsed XML for the S3
Multi-Object-Delete-API to 2 MB. This fixes a DoS
vulnerability since (auth.) clients, MitM-adversaries
(without TLS) and un-auth. users accessing buckets allowing
multi-delete by policy can kill the server.

This behavior is similar to the AWS-S3 implementation.

## Motivation and Context
Security, DoS

## How Has This Been Tested?
manually

## Types of changes

## Checklist:

### DIFF
--- a/cmd/bucket-handlers.go
+++ b/cmd/bucket-handlers.go
@@ -282,7 +282,13 @@ func (api objectAPIHandlers) DeleteMultipleObjectsHandler(w http.ResponseWriter,
 	}
 
 	// Allocate incoming content length bytes.
-	deleteXMLBytes := make([]byte, r.ContentLength)
+	var deleteXMLBytes []byte
+	const maxBodySize = 2 * 1000 * 1024 // The max. XML contains 1000 object names (each at most 1024 bytes long) + XML overhead
+	if r.ContentLength > maxBodySize {  // Only allocated memory for at most 1000 objects
+		deleteXMLBytes = make([]byte, maxBodySize)
+	} else {
+		deleteXMLBytes = make([]byte, r.ContentLength)
+	}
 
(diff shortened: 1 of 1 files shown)


annanay25:
I guess this would have to be tested with a large (>2MB) string here - [test-utils_test.go](https://github.com/minio/minio/blob/3dc13323e51dc7038232f5f02f55b37b388c59c2/cmd/test-utils_test.go#L1565)

aead:
@annanay25 At the linked line we cannot insert anything which could test this.
A negative test would be to send a > 2MB request body (valid XML & non-valid-XML) to the server and expect that it succeed - but delete less objects - / fail because of non-valid XML after sending ~2MB.

I think these tests should be implemented in mint, but we have to agree on that first...

harshavardhana:
> I think these tests should be implemented in mint, but we have to agree on that first...

Yes this is correct @aead

harshavardhana (approved):

donatello (approved):
Restarted build as it seemed to have failed due to play being down.

PR #6035, opened 2018-06-11
Avoid allocating larger buffer if the incoming body is smaller

harshavardhana:
## Description
Avoid allocating a larger buffer if the incoming body is smaller

## Motivation and Context
Cleanup

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:

### DIFF
--- a/cmd/xl-v1-multipart.go
+++ b/cmd/xl-v1-multipart.go
@@ -375,8 +375,17 @@ func (xl xlObjects) PutObjectPart(ctx context.Context, bucket, object, uploadID 
 	}
 
 	// Fetch buffer for I/O, returns from the pool if not allocates a new one and returns.
-	buffer := xl.bp.Get()
-	defer xl.bp.Put(buffer)
+	var buffer []byte
+	switch size := data.Size(); {
+	case size == 0:
+		buffer = make([]byte, 1) // Allocate atleast a byte to reach EOF
+	case size < blockSizeV1:
+		// No need to allocate fully blockSizeV1 buffer if the incoming data is smaller.
+		buffer = make([]byte, size, 2*size)
+	default:
+		buffer = xl.bp.Get()
+		defer xl.bp.Put(buffer)
+	}
 
 	file, err := storage.CreateFile(ctx, data, minioMetaTmpBucket, tmpPartPath, buffer, DefaultBitrotAlgorithm, writeQuorum)
 	if err != nil {
--- a/cmd/xl-v1-object.go
+++ b/cmd/xl-v1-object.go
@@ -611,8 +611,17 @@ func (xl xlObjects) putObject(ctx context.Context, bucket string, object string,
 	}
 
 	// Fetch buffer for I/O, returns from the pool if not allocates a new one and returns.
-	buffer := xl.bp.Get()
-	defer xl.bp.Put(buffer)
+	var buffer []byte
+	switch size := data.Size(); {
+	case size == 0:
+		buffer = make([]byte, 1) // Allocate atleast a byte to reach EOF
+	case size < blockSizeV1:
+		// No need to allocate fully blockSizeV1 buffer if the incoming data is smaller.
+		buffer = make([]byte, size, 2*size)
+	default:
+		buffer = xl.bp.Get()
+		defer xl.bp.Put(buffer)
+	}
 
 	// Read data and split into parts - similar to multipart mechanism
 	for partIdx := 1; ; partIdx++ {


cmd/xl-v1-multipart.go
```diff
@@ -375,8 +375,19 @@ func (xl xlObjects) PutObjectPart(ctx context.Context, bucket, object, uploadID
 	}
 
 	// Fetch buffer for I/O, returns from the pool if not allocates a new one and returns.
-	buffer := xl.bp.Get()
-	defer xl.bp.Put(buffer)
+	var buffer []byte
+	if data.Size() < blockSizeV1 {
```
aead: Suggestion: Replace `if-else` cascade with `switch`:
```
switch size := data.Size(); {
    case size == 0:  
          buffer = make([]byte, 1) // Allocate atleast a byte to reach EOF
    case size < blockSizeV1: 
          buffer = make([]byte, data.Size())
    default: 
         buffer = xl.bp.Get()
         defer xl.bp.Put(buffer)
}
```

cmd/test-utils_test.go
```diff
@@ -2057,6 +2057,7 @@ func ExecObjectLayerTest(t TestErrHandler, objTest objTestType) {
 	if err != nil {
 		t.Fatalf("Initialization of object layer failed for XL setup: %s", err)
 	}
+
```
aead: Unnecessary?!

aead (commented):
General idea LGTM - I would suggest using a `switch` clause...

aead (dismissed):
LGTM

krishnasrinivas (commented):
Are you sure this helps memory? because the new code is not taking advantage of pools for smaller files and hence more pressure on garbage collector.

Are there any tests to show which is better?

harshavardhana:
> Are you sure this helps memory? because the new code is not taking advantage of pools for smaller files and hence more pressure on garbage collector.
> 
> Are there any tests to show which is better?

The pool doesn't have an absolute cap @krishnasrinivas - so if the pool is full it does create a new buffer and send it - this change is to ensure that we do not allocate more than we need for smaller files.  Relinquishing a smaller memory is not a big deal for Go - the issue is really about allocating a 10MiB buffer and relinquishing it when we will only need perhaps few KBs

cmd/xl-v1-multipart.go
```diff
@@ -375,8 +375,17 @@ func (xl xlObjects) PutObjectPart(ctx context.Context, bucket, object, uploadID
...
 	// Fetch buffer for I/O, returns from the pool if not allocates a new one and returns.
-	buffer := xl.bp.Get()
-	defer xl.bp.Put(buffer)
+	var buffer []byte
+	switch size := data.Size(); {
+	case size == 0:
+		buffer = make([]byte, 1) // Allocate atleast a byte to reach EOF
+	case size < blockSizeV1:
+		// No need to allocate fully blockSizeV1 buffer if the incoming data is smaller.
+		buffer = make([]byte, size)
```
aead: We should allocate 2*size here because the erasure coding will use [0:size] for data blocks and [size:2*size] for parity.
Otherwise we will allocate memory during erasure coding.
So `make([]byte, size, 2*size)`

aead:
I tested this using a smarter implementation of the `BytePool` which produces buffers only as large as requested, the solution of this PR and the old `BytePool` implementation used before this PR.

I simulated high load by consuming more buffers than available pool-capacity:

```
BenchmarkGetSmartPool-4                    10000            161755 ns/op         1050673 B/op          0 allocs/op
BenchmarkGet-4                              1000           1629169 ns/op        10485760 B/op          0 allocs/op
BenchmarkGetDirectAllocSmall-4             10000            163241 ns/op         1050673 B/op          0 allocs/op
```  
Using 10MB/20MB regular buffers (`blockSizeV1`) and 1 / 2 MB for smaller buffers.
Both solutions, smarter pool and allocating smaller buffers directly are ~10x faster

cmd/xl-v1-multipart.go
harshavardhana: Ah i see @aead

harshavardhana:
> I tested this using a smarter implementation of the BytePool which produces buffers only as large as requested, the solution of this PR and the old BytePool implementation used before this PR.

can you share the smarter bytepool implementation?

cmd/xl-v1-multipart.go
harshavardhana: This is what the bytepool does

```
        bp := bpool.NewBytePoolCap(setCount*drivesPerSet, blockSizeV1, blockSizeV1*2)
```

harshavardhana:
> BenchmarkGet-4                              1000           1629169 ns/op        10485760 B/op          0 allocs/op

The slowness is perhaps due to channels being used @aead

cmd/xl-v1-multipart.go
aead: Yes, so we should do the same for smaller buffers.

aead:
>The slowness is perhaps due to channels being used @aead

No, it is because the byte-pool keeps allocating 10/20 MB buffers instead of 1/2 MB.

> can you share the smarter bytepool implementation?

Sure.

aead:
```
type BytePool struct {
	mux  sync.RWMutex
	pool map[int]chan []byte
	size int
	cap  int
}

// NewBytePool returns a new BytePool with the given size and capacity.
// The size controls how many buffers of a certain length are available in the pool
// while the cap controls how many different pools are managed by the BytePool.
// The total number of available byte buffers is: size x cap.
func NewBytePool(size, cap int) *BytePool {
	return &BytePool{
		pool: make(map[int]chan []byte, cap),
		size: size,
		cap:  cap,
	}
}

// enqueue tries to add b to the channel. It returns true on success.
func enqueue(bchan chan<- []byte, b []byte) bool {
	select {
	case bchan <- b:
		return true
	default:
		return false
	}
}

// Get returns a new byte buffer of the given size. It allocates a new one
// if the pool does not contain a sufficient buffer. The capacity of the
// returned buffer is equal to its length.
func (bp *BytePool) Get(size int) ([]byte, bool) { return bp.GetWithCap(size, size) }

// GetWithCap returns a new byte buffer of the given size and capacity. It
// allocates a new one if the pool does not contain a sufficient buffer.
func (bp *BytePool) GetWithCap(size, capacity int) ([]byte, bool) {
	bp.mux.RLock()
	bchan, ok := bp.pool[size]
	if !ok && bp.cap > 0 && len(bp.pool) < bp.cap {
		bp.mux.RUnlock()
		bp.mux.Lock()
		if bchan, ok = bp.pool[size]; !ok && bp.cap > 0 && len(bp.pool) < bp.cap { // in between another caller could have created a pool[size]
			bchan = make(chan []byte, bp.size)
		}
		bp.pool[size] = bchan
		bp.mux.Unlock()

		b := make([]byte, size, capacity)
		return b, enqueue(bchan, b)
	} else if !ok {
		bp.mux.RUnlock()
		return make([]byte, size, capacity), false
	}
	bp.mux.RUnlock()

	var b []byte
	select {
	case b = <-bchan:
	default:
	}

	if cap(b) < capacity {
		bp.Put(b)
		return make([]byte, size, capacity), false
	}
	return b[:size:capacity], true
}

// Put returns b to the BytePool and returns true if the BytePool
// has enough capacity to take the buffer. Otherwise the
// buffer is not added to the pool and can still be used by the
// caller.
func (bp *BytePool) Put(b []byte) bool {
	size := len(b)
	bp.mux.RLock()
	bchan, ok := bp.pool[size]
	if !ok && bp.cap > 0 && len(bp.pool) < bp.cap {
		bp.mux.RUnlock()
		bp.mux.Lock()
		if bchan, ok = bp.pool[size]; !ok && bp.cap > 0 && len(bp.pool) < bp.cap { // in between another caller could have created a pool[size]
			bchan = make(chan []byte, bp.size)
		}
		bp.pool[size] = bchan
		bp.mux.Unlock()
		return enqueue(bchan, b)
	} else if !ok {
		bp.mux.RUnlock()
		return false
	}
	bp.mux.RUnlock()
	return enqueue(bchan, b)
}
```

harshavardhana:
> No, it is because the byte-pool keeps allocating 10/20 MB buffers instead of 1/2 MB.

Ah I see makes sense, changed the PR to allow a cap for the smaller sizes as well.

aead (approved):

krishnasrinivas (approved):

PR #6145, opened 2018-07-12
Support multiple values for x-amz-meta header

### LINKED ISSUE #5595: Minio fails to store multiple values for user metadata keys
## Expected Behavior
When multiple headers with same key are set, we should collect all the values into a comma separated string and save.

## Current Behavior
With multiple headers with same key are set, we only save the first value. 

## Possible Solution
Metadata extraction function should simply store all the values set in user metadata not just `one` value. 

```go
// extractMetadataFromHeader extracts metadata from HTTP header.
func extractMetadataFromHeader(header http.Header) (map[string]string, error) {
        if header == nil {
                return nil, errors.Trace(errInvalidArgument)
        }
        metadata := make(map[string]string)

        // Save all supported headers.
        for _, supportedHeader := range supportedHeaders {
                canonicalHeader := http.CanonicalHeaderKey(supportedHeader)
                // HTTP headers are case insensitive, look for both canonical
                // and non canonical entries.
                if _, ok := header[canonicalHeader]; ok {
                        metadata[supportedHeader] = header.Get(canonicalHeader)
                } else if _, ok := header[supportedHeader]; ok {
                        metadata[supportedHeader] = header.Get(supportedHeader)
                }
        }

        // Go through all other headers for any additional headers that needs to be saved.
        for key := range header {
                if key != http.CanonicalHeaderKey(key) {
                        return nil, errors.Trace(errInvalidArgument)
                }
                for _, prefix := range userMetadataKeyPrefixes {
                        if strings.HasPrefix(key, prefix) {
                                metadata[key] = header.Get(key) ---> we should simply do `header[key]` instead
                                break
                        }
                }
        }
        return metadata, nil
}
```

```php
<?php

// Include the SDK using the Composer autoloader
date_default_timezone_set('America/Los_Angeles');
require 'vendor/autoload.php';

$s3 = new Aws\S3\S3Client([
        'version' => 'latest',
        'region'  => 'us-east-1',
        'endpoint' => 'https://s3.amazonaws.com',
        'use_path_style_endpoint' => true,
        'credentials' => [
                'key'    => 'minio',
                'secret' => 'minio123',
        ],
]);


// Send a PutObject request and get the result object.
$insert = $s3->putObject([
        'Bucket' => 'testbucket',
        'Key'    => 'testkey',
        'Body'   => 'Hello from Minio!!',
        'Metadata'     => array(    
                'param1' => 'value 1',
                'Param1' => 'value 2',
                'paRam1' => 'value 3',
        )        
]);

// Get a command object from the client
$command = $s3->getCommand('GetObject', [
            'Bucket' => 'testbucket',
            'Key'    => 'testkey'
        ]);

// Create a pre-signed URL for a request with duration of 10 miniutes
$presignedRequest = $s3->createPresignedRequest($command, '+10 minutes');

// Get the actual presigned-url
$presignedUrl =  (string)  $presignedRequest->getUri();
echo $presignedUrl;
```
```
curl -v `php example.php` 
``` 

NOTE: you need to configure and install aws-sdk-php first. Please follow our docs on this https://docs.minio.io/docs/how-to-use-aws-sdk-for-php-with-minio-server

## Context
AWS S3 compatibility testing

## Your Environment
master and all other previous releases of minio.

harshavardhana:
## Description
Support multiple values for x-amz-meta header

## Motivation and Context
Fixes #5595

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:

### DIFF
--- a/cmd/handler-utils.go
+++ b/cmd/handler-utils.go
@@ -156,7 +156,7 @@ func extractMetadataFromMap(ctx context.Context, v map[string][]string, m map[st
 			}
 			value, ok := v[key]
 			if ok {
-				m[key] = value[0]
+				m[key] = strings.Join(value, ",")
 				break
 			}
 		}


nitisht (approved):
LGTM, Can we add a mint test for this?

harshavardhana:
We can definitely add @nitisht

Praveenrajmani (approved):

Issue #3314, opened 2016-11-22
Running Minio on port "0" should print the right port allocated by the OS, not "0"


yhafri:
I'm trying to use Minio without specifying any fixed port number during startup.
For this, I'm relying on the OS to give me back a free port to use.

## Expected Behavior
$ minio server --address "127.0.0.1:0" /tmp/.s3share

Endpoint:  http://127.0.0.1:57088
AccessKey: KM6074YUX3OB5BK06D02
SecretKey: LuQwctJmvJQUyPA6ogYWKtkMLFo9uEBez6/5Y/o5
Region:    us-east-1
SqsARNs:

Browser Access:
   http://127.0.0.1:57088

Command-line Access: https://docs.minio.io/docs/minio-client-quickstart-guide
   $ mc config host add myminio http://127.0.0.1:57088 KM6074YUX3OB5BK06D02 LuQwctJmvJQUyPA6ogYWKtkMLFo9uEBez6/5Y/o5

And as you can see, it works as expected.

## Current Behavior

What i'm getting now is this (which is OK but kind of useless):

$ minio server --address "127.0.0.1:0" /tmp/.s3share

Endpoint:  http://127.0.0.1:0
AccessKey: KM6074YUX3OB5BK06D02
SecretKey: LuQwctJmvJQUyPA6ogYWKtkMLFo9uEBez6/5Y/o5
Region:    us-east-1
SqsARNs:

Browser Access:
   http://127.0.0.1:0

Command-line Access: https://docs.minio.io/docs/minio-client-quickstart-guide
   $ mc config host add myminio http://127.0.0.1:0 KM6074YUX3OB5BK06D02 LuQwctJmvJQUyPA6ogYWKtkMLFo9uEBez6/5Y/o5

## Possible Solution
Once the OS gave us an automatic free port (in this case "57088"), use that one instead of "0". 

Actually, I'm using netstat to discover which port Minio is using, which is very tedious:
I'm trying to use Minio without specifying any fixed port number during startup.
For this, I'm relying on the OS to give back me a free port to use.

## Expected Behavior
$ minio server --address "127.0.0.1:0" /tmp/.s3share

Endpoint:  http://127.0.0.1:57088
AccessKey: KM6074YUX3OB5BK06D02
SecretKey: LuQwctJmvJQUyPA6ogYWKtkMLFo9uEBez6/5Y/o5
Region:    us-east-1
SqsARNs:

Browser Access:
   http://127.0.0.1:57088

[...]

PR #6179, opened 2018-07-20
Do not forbid listening on port 0

vincentbernat:
## Description

Allow a user to specify ":0" as a listening address.

## Motivation and Context

This is useful to bind on ":0" as it enables one to run minio on any
free port. This is convenient when running tests as it ensure the test
won't fail because a port is already in use.

## How Has This Been Tested?

Started minio with `./minio server --address 127.0.0.1:0`, checked we can connect to the provided endpoint.

## Types of changes

## Checklist:

- [x] I have added unit tests to cover my changes (well, modified an existing one).

### DIFF
--- a/cmd/http/listener.go
+++ b/cmd/http/listener.go
@@ -297,7 +297,7 @@ func newHTTPListener(serverAddrs []string,
 	readTimeout time.Duration,
 	writeTimeout time.Duration,
 	updateBytesReadFunc func(int),
-	updateBytesWrittenFunc func(int)) (listener *httpListener, err error) {
+	updateBytesWrittenFunc func(int)) (listener *httpListener, listeningAddrs []string, err error) {
 
 	var tcpListeners []*net.TCPListener
 	// Close all opened listeners on error
@@ -315,15 +315,16 @@ func newHTTPListener(serverAddrs []string,
 	for _, serverAddr := range serverAddrs {
 		var l net.Listener
 		if l, err = net.Listen("tcp", serverAddr); err != nil {
-			return nil, err
+			return nil, nil, err
 		}
 
 		tcpListener, ok := l.(*net.TCPListener)
 		if !ok {
-			return nil, fmt.Errorf("unexpected listener type found %v, expected net.TCPListener", l)
+			return nil, nil, fmt.Errorf("unexpected listener type found %v, expected net.TCPListener", l)
 		}
 
 		tcpListeners = append(tcpListeners, tcpListener)
+		listeningAddrs = append(listeningAddrs, tcpListener.Addr().String())
 	}
 
 	listener = &httpListener{
@@ -337,5 +338,5 @@ func newHTTPListener(serverAddrs []string,
 	}
 	listener.start()
 
-	return listener, nil
+	return listener, listeningAddrs, nil
 }
--- a/cmd/http/listener_test.go
+++ b/cmd/http/listener_test.go
@@ -228,7 +228,7 @@ func TestNewHTTPListener(t *testing.T) {
 	}
 
 	for _, testCase := range testCases {
-		listener, err := newHTTPListener(
+		listener, _, err := newHTTPListener(
 			testCase.serverAddrs,
 			testCase.tlsConfig,
 			testCase.tcpKeepAliveTimeout,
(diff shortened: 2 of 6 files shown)


harshavardhana:
@vincentbernat Thanks for the PR. We didn't support port ":0" for a specific purpose.

There are general problems with this approach let me state them one by one here.

- Distributed clusters cannot be set up in this manner, we have no way of knowing what port is being assigned by the system and no way to probe. Each restart of a server picks a new port which is detrimental leading to frequent I/O disconnections.

- Port configured when used by clients and applications have to be continuously changed, when there is a restart of a server. This causes significant disruption in the process. There is no way for clients to discover the right port etc with this approach.

> This is useful to bind on ":0" as it enables one to run minio on any
> free port. This is convenient when running tests as it ensure the test
> won't fail because a port is already in use.

- ^^ This is not a sufficient requirement to provide this support, in-fact you can do this in your own test launcher routines to pick a free port and provide it to Minio server.

vincentbernat:
It is only intended to be used during automated tests outside of a cluster. If you run tests in parallel, checking for a free port, then using it is racy. Using port 0 works all the time. I don't see why people would use this "by accident". However, if you are worried about that, we can just remove the code that displays the right endpoint. People not knowing how port 0 works won't be able to connect to the Minio instance and will use another port.

harshavardhana:
> It is only intended to be used during automated tests outside of a cluster. If you run tests in parallel, checking for a free port, then using it is racy. Using port 0 works all the time. I don't see why people would use this "by accident". However, if you are worried about that, we can just remove the code that displays the right endpoint. People not knowing how port 0 works won't be able to connect to the Minio instance and will use another port.

We cannot restrict one way or the other we cannot treat one form to allow one style another form with another style we don't know that you are using it for automated tests or in some other way, whatever is currently implemented is correct in behavior and enforces the right restrictions.

As I suggested since you are launching Minio yourself pick a free port and give it to Minio for tests. Minio should be launched only once per test run, even if the tests are parallel they should use single minio instance. These are architectural things which are outside the scope of Minio behavior. I don't see a good reason behind this PR it needs to be solved outside of Minio.

harshavardhana:
https://github.com/minio/minio/issues/3314 an older issue with the same explanation on why cannot be cleanly achieved.

vincentbernat:
It seems odd to refuse this feature to people with actual use cases while nobody complained against it. In both cases, the workaround with a wrapper cannot work reliably. If you have several wrappers competing to find a free port, they will detect a free port that won't be free at time of use. You can retry several times and eventually get it working, but that's quite complex to compared to just allowing use of port 0.

May we add a warning when using port 0 saying some features may not work when using Minio this way?

harshavardhana:
> It seems odd to refuse this feature to people with actual use cases while nobody complained against it. In both cases, the workaround with a wrapper cannot work reliably. If you have several wrappers competing to find a free port, they will detect a free port that won't be free at time of use. You can retry several times and eventually get it working, but that's quite complex to compared to just allowing use of port 0.

Allowing this opens more mistakes by users, once they configure their clients they cannot go back and change it.  For simple maintenance, let's say they restarted the server port would be changed now breaking all existing configured clients and applications. 

It is not possible to provide this cleanly where clients discover Minio's port automatically, `:0` is a nice feature when the service such as like NFS etc can register themselves with portmapper and NFS clients can discover them through portmapper which runs on a static port 111.  The ephemeral nature of port `:0` is detrimental to all I/O which is bound to a port.

We are restrictive for a reason to avoid users making mistakes. At this point in time I am not convinced that this adds any value.

ywelsch:
Can this decision be revisited? [We're using Minio in our build](https://github.com/elastic/elasticsearch/pull/31688) to test API compatibility of Minio with our Elasticsearch S3 repository plugin. In our tests, we just want to spin up a single instance of minio, have it choose a free port, and then run our tests against it. I get that you would like to avoid having this feature for the reason of avoiding more mistakes by users. The solution we've used for Elasticsearch was to introduce [bootstrap checks](https://www.elastic.co/guide/en/elasticsearch/reference/current/bootstrap-checks.html), which distinguish between development mode and production mode.

harshavardhana:
> Can this decision be revisited? We're using Minio in our build to test API compatibility of Minio with our Elasticsearch S3 repository plugin. In our tests, we just want to spin up a single instance of minio, have it choose a free port, and then run our tests against it. I get that you would like to avoid having this feature for the reason of avoiding more mistakes by users. The solution we've used for Elasticsearch was to introduce bootstrap checks, which distinguish between development mode and production mode.

You can simply do that yourself @ywelsch as explained before, we cannot do this due to important concerns as mentioned before.

PR #6264, opened 2018-08-08
GetObject(): Avoid sending an error after 206 HTTP code

vadmeste:
## Description
When a S3 client sends a GET Object with a range header, 206 http
code is returned indicating success, however the call of the object
layer's GetObject() inside the handler can return an error and will lead
to writing an XML error message, which is obviously wrong since
we already sent 206 http code. So in the case, we just stop sending
data to the S3 client, this latter can still detect if there is no
error when comparing received data with Content-Length header
in the Get Object response.

## Motivation and Context
Fixing bug

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:

### DIFF
--- a/cmd/object-handlers.go
+++ b/cmd/object-handlers.go
@@ -173,16 +173,23 @@ func (api objectAPIHandlers) GetObjectHandler(w http.ResponseWriter, r *http.Req
 
 	setObjectHeaders(w, objInfo, hrange)
 	setHeadGetRespHeaders(w, r.URL.Query())
-	httpWriter := ioutil.WriteOnClose(writer)
 
 	getObject := objectAPI.GetObject
 	if api.CacheAPI() != nil && !hasSSECustomerHeader(r.Header) {
 		getObject = api.CacheAPI().GetObject
 	}
 
+	statusCodeWritten := false
+	httpWriter := ioutil.WriteOnClose(writer)
+
+	if hrange != nil && hrange.offsetBegin > -1 {
+		statusCodeWritten = true
+		w.WriteHeader(http.StatusPartialContent)
+	}
+
 	// Reads the object at startOffset and writes to mw.
 	if err = getObject(ctx, bucket, object, startOffset, length, httpWriter, objInfo.ETag); err != nil {
-		if !httpWriter.HasWritten() { // write error response only if no data has been written to client yet
+		if !httpWriter.HasWritten() && !statusCodeWritten { // write error response only if no data or headers has been written to client yet
 			writeErrorResponse(w, toAPIErrorCode(err), r.URL)
 		}
 		httpWriter.Close()
@@ -190,7 +197,7 @@ func (api objectAPIHandlers) GetObjectHandler(w http.ResponseWriter, r *http.Req
 	}
 
 	if err = httpWriter.Close(); err != nil {
-		if !httpWriter.HasWritten() { // write error response only if no data has been written to client yet
+		if !httpWriter.HasWritten() && !statusCodeWritten { // write error response only if no data or headers has been written to client yet
 			writeErrorResponse(w, toAPIErrorCode(err), r.URL)
 			return
 		}
--- a/cmd/api-headers.go
+++ b/cmd/api-headers.go
@@ -100,6 +100,5 @@ func setObjectHeaders(w http.ResponseWriter, objInfo ObjectInfo, contentRange *h
 		// Override content-length
 		w.Header().Set("Content-Length", strconv.FormatInt(contentRange.getLength(), 10))
 		w.Header().Set("Content-Range", contentRange.String())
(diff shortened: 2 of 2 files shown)


cmd/object-handlers.go
```diff
@@ -171,29 +171,23 @@ func (api objectAPIHandlers) GetObjectHandler(w http.ResponseWriter, r *http.Req
 		}
 	}
 
-	setObjectHeaders(w, objInfo, hrange)
 	setHeadGetRespHeaders(w, r.URL.Query())
-	httpWriter := ioutil.WriteOnClose(writer)
+	setObjectHeaders(w, objInfo, hrange)
```
harshavardhana: This change I don't know what effect might bring in, the previous code is fine as well until here...

cmd/object-handlers.go
```diff
@@ -171,29 +171,23 @@ func (api objectAPIHandlers) GetObjectHandler(w http.ResponseWriter, r *http.Req
...
+	// be returned at this stage.
 
 	getObject := objectAPI.GetObject
 	if api.CacheAPI() != nil && !hasSSECustomerHeader(r.Header) {
 		getObject = api.CacheAPI().GetObject
 	}
 
-	// Reads the object at startOffset and writes to mw.
-	if err = getObject(ctx, bucket, object, startOffset, length, httpWriter, objInfo.ETag); err != nil {
-		if !httpWriter.HasWritten() { // write error response only if no data has been written to client yet
```
harshavardhana: This code was to detect if we started writing if we can keep this and have `httpWriter := ioutil.WriteOnClose(writer)`  but we should detect that we have written atleast the part about http status WriteOnClose()

harshavardhana (changes requested):

cmd/object-handlers.go
vadmeste: > This change I don't know what effect might bring in, the previous code is fine as well until here...

The thing is that setObjectHeaders() writes the header status code and I want that this will be the last thing to do before we forbid sending errors, just for better code but not functional change

harshavardhana (changes requested):
```
$  git diff
diff --git a/cmd/api-headers.go b/cmd/api-headers.go
index ac030c4e6..1ed996d86 100644
--- a/cmd/api-headers.go
+++ b/cmd/api-headers.go
@@ -100,6 +100,5 @@ func setObjectHeaders(w http.ResponseWriter, objInfo ObjectInfo, contentRange *h
                // Override content-length
                w.Header().Set("Content-Length", strconv.FormatInt(contentRange.getLength(), 10))
                w.Header().Set("Content-Range", contentRange.String())
-               w.WriteHeader(http.StatusPartialContent)
        }
 }
diff --git a/cmd/object-handlers.go b/cmd/object-handlers.go
index 8274841d8..c12748c02 100644
--- a/cmd/object-handlers.go
+++ b/cmd/object-handlers.go
@@ -180,6 +180,11 @@ func (api objectAPIHandlers) GetObjectHandler(w http.ResponseWriter, r *http.Req
                getObject = api.CacheAPI().GetObject
        }
 
+       // for providing ranged content
+       if contentRange != nil && contentRange.offsetBegin > -1 {
+               w.WriteHeader(http.StatusPartialContent)
+       }
+
        if err = getObject(ctx, bucket, object, startOffset, length, httpWriter, objInfo.ETag); err != nil {
                if !httpWriter.HasWritten() { // write error response only if no data has been written to client yet
```

We only need to make this change, detect 206 partial content as already written to avoid writing quorum errors.

vadmeste:
@harshavardhana PR updated.

cmd/object-handlers.go
```diff
@@ -173,24 +173,31 @@ func (api objectAPIHandlers) GetObjectHandler(w http.ResponseWriter, r *http.Req
...
 	if api.CacheAPI() != nil && !hasSSECustomerHeader(r.Header) {
 		getObject = api.CacheAPI().GetObject
 	}
 
+	statusCodeWritten := false
+	httpWriter := ioutil.WriteOnClose(writer)
+
+	if hrange != nil && hrange.offsetBegin > -1 {
+		statusCodeWritten = true
+		w.WriteHeader(http.StatusPartialContent)
```
poornas, line 187: for a Get with range header, this throws multiple response.WriteHeader calls warning.

poornas (changes requested):

cmd/object-handlers.go
vadmeste: > for a Get with range header, this throws multiple response.WriteHeader calls warning.

oups yes, fixed.

harshavardhana (approved):

poornas (approved):
LGTM.

minio-ops:
## Mint Automation
| Test | Result |
|:--- |:--- |

PR #6535, opened 2018-09-27
xl: Require full write quorum in rename() as general rule

### LINKED ISSUE #6529: Removing an empty directory not present in all disks fail to remove its parent directory although empty
## Expected Behavior
When I remove an empty directory living alone in its parent directory, the parent directory should be removed in all cases.

## Current Behavior
When an empty directory is not present in all disks (not healthy), removing it won't remove the parent directory although the latter becomes empty

## Possible Solution
Do not know

1. Start Minio standalone erasure with 4 fresh disks
2. aws s3api put-object --bucket testbucket --key dir/emptydir/ --endpoint http://minio:9010
3. Remove emptydir in two disks from the backend
4. mc rm myminio/testbucket/dir/emptydir/
5. mc ls myminio/testbucket/ and mc ls myminio/testbucket/dir/

## Context
Fixing an empty object issue

## Regression
No

## Your Environment
All

vadmeste:
## Description
Simplify the logic of using rename() in xl. Currently, renaming
doesn't require the source object/dir to be existant in at least
read quorum disks, since there is no apparent reason for it
to be as a general rule, this commit will just simplify the
logic to avoid possible inconsistency in the backend in the future.

## Motivation and Context
Simple enhancement in the code

## Regression
No

## How Has This Been Tested?
~1. Start Minio standalone erasure with 4 fresh disks
2. `aws s3api put-object --bucket testbucket --key dir/emptydir/ --endpoint http://minio:9010`
3. Remove emptydir in two disks from the backend
4. `mc rm myminio/testbucket/dir/emptydir/`
5. `mc ls myminio/testbucket/ and mc ls myminio/testbucket/dir/`~

## Types of changes

## Checklist:

### DIFF
--- a/cmd/xl-v1-multipart.go
+++ b/cmd/xl-v1-multipart.go
@@ -220,7 +220,7 @@ func (xl xlObjects) newMultipartUpload(ctx context.Context, bucket string, objec
 	defer xl.deleteObject(ctx, minioMetaTmpBucket, tempUploadIDPath, writeQuorum, false)
 
 	// Attempt to rename temp upload object to actual upload path object
-	_, rErr := renameObject(ctx, disks, minioMetaTmpBucket, tempUploadIDPath, minioMetaMultipartBucket, uploadIDPath, writeQuorum)
+	_, rErr := rename(ctx, disks, minioMetaTmpBucket, tempUploadIDPath, minioMetaMultipartBucket, uploadIDPath, true, writeQuorum, nil)
 	if rErr != nil {
 		return "", toObjectErr(rErr, minioMetaMultipartBucket, uploadIDPath)
 	}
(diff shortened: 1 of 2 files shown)


vadmeste:
Not blocked anymore but the purpose of this PR is changed, tested with mint

harshavardhana (approved):

harshavardhana (approved):

harshavardhana:
ping @donatello

donatello (commented):
This code looks like a refactoring with no changes in behaviour at all as far as I can tell - is this true?

If so, I think we should indicate in the commit title that it is a refactor - otherwise, it looks misleading that now write quorum is enforced*.

vadmeste:
@donatello there is a little simplification in the logic:

rename() is often called in xl API, just before returning a success to the user, for example:
https://github.com/minio/minio/blob/master/cmd/xl-v1-multipart.go#L223

Which means even if renameObject() doesn't find the upload path in at least quorum disks it will still return success with nil error.

Although there is no immediate bug for this I really suggest to simplify the logic

donatello (approved):

vadmeste:
mint tests passed

PR #6539, opened 2018-09-28
Fix racy error communication inside go-routine

harshavardhana:
## Description
Fix racy error communication inside go-routine

## Motivation and Context
Use CloseWithError to communicate errors in a pipe,
this PR also fixes potential shadowing of error

## Regression
No

## How Has This Been Tested?
- make test won't compile due to `go vet` complaints
- Also reproduced this racy behavior by sending short 
  data using `curl` commands basically set a
  high enough content-length but send only a few bytes. 

## Types of changes

## Checklist:

### DIFF
--- a/cmd/object-handlers.go
+++ b/cmd/object-handlers.go
@@ -738,25 +738,20 @@ func (api objectAPIHandlers) CopyObjectHandler(w http.ResponseWriter, r *http.Re
 	// No need to compress for remote etcd calls
 	// Pass the decompressed stream to such calls.
 	if srcInfo.IsCompressed() && !isRemoteCallRequired(ctx, srcBucket, dstBucket, objectAPI) {
-		var sreader io.Reader
-		var swriter io.Writer
-
 		// Open a pipe for compression.
-		// Where snappyWriter is piped to srcInfo.Reader.
-		// gr writes to snappyWriter.
-		snappyReader, snappyWriter := io.Pipe()
-		reader = snappyReader
+		// Where pipeWriter is piped to srcInfo.Reader.
+		// gr writes to pipeWriter.
+		pipeReader, pipeWriter := io.Pipe()
+		reader = pipeReader
 		length = -1
 
-		swriter = snappy.NewWriter(snappyWriter)
-		sreader = gr
+		snappyWriter := snappy.NewWriter(pipeWriter)
 
 		go func() {
-			defer snappyWriter.Close()
 			// Compress the decompressed source object.
-			if _, err = io.Copy(swriter, sreader); err != nil {
-				return
-			}
+			_, cerr := io.Copy(snappyWriter, gr)
+			snappyWriter.Close()
+			pipeWriter.CloseWithError(cerr)
 		}()
 	} else {
 		// Remove the metadata for remote calls.
@@ -1102,7 +1097,6 @@ func (api objectAPIHandlers) PutObjectHandler(w http.ResponseWriter, r *http.Req
 		}
 	}
 
-	var hashError error
 	actualSize := size
 
 	if objectAPI.IsCompressionSupported() && isCompressible(r.Header, object) && size > 0 {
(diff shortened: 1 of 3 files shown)


balamurugana (dismissed):

cmd/object-handlers.go
```diff
@@ -1113,21 +1110,21 @@ func (api objectAPIHandlers) PutObjectHandler(w http.ResponseWriter, r *http.Req
...
-			defer pipeWriter.Close()
-			defer snappyWriter.Close()
 			// Writing to the compressed writer.
-			_, err = io.CopyN(snappyWriter, actualReader, actualSize)
-			if err != nil {
-				hashError = err
-				return
-			}
+			_, cerr := io.CopyN(snappyWriter, actualReader, actualSize)
+			pipeWriter.CloseWithError(cerr)
```
Praveenrajmani: A small suggestion - it is safer to close the `snappyWriter` first and then the `pipeWriter`. The reason for having `pipeWriter.Close()` is because `snappyWriter.Close()` will not close the wrapped writer (`pipeWriter`) here [https://github.com/golang/snappy/blob/2e65f85255dbc3072edf28d6b5b8efc472979f5a/encode.go#L278](url)

So, by closing snappyWriter first , we let snappy do all its closures first (even for future versions) and then close the wrapped writer (`pipeWriter`) after this.

cmd/object-handlers.go
```diff
@@ -1197,15 +1194,6 @@ func (api objectAPIHandlers) PutObjectHandler(w http.ResponseWriter, r *http.Req
 		}
 	}
 
-	if hashError != nil {
```
Praveenrajmani, line 1200: Our current unit tests works such a way that the `putObject()` errors are treated with higher priority. Since unit tests wont run on compression mode we cannot detect such behaviour. 
So by recording the `hashError` and then erroring out if `putObject` err is nil gives higher priority to the putObject errors.
Fine if this is not a problem here.

cmd/object-handlers.go
```diff
@@ -1113,21 +1110,21 @@ func (api objectAPIHandlers) PutObjectHandler(w http.ResponseWriter, r *http.Req
...
 		go func() {
-			defer pipeWriter.Close()
-			defer snappyWriter.Close()
 			// Writing to the compressed writer.
-			_, err = io.CopyN(snappyWriter, actualReader, actualSize)
-			if err != nil {
-				hashError = err
-				return
-			}
+			_, cerr := io.CopyN(snappyWriter, actualReader, actualSize)
```
Praveenrajmani, line 1119: shouldn't we treat `io.ErrUnexpectedEOF` seperately as ErrIncompleteBody ? as per this discussion [https://github.com/minio/minio/pull/6079#discussion_r206027174](url)

cmd/web-handlers.go
```diff
@@ -662,21 +661,20 @@ func (web *webAPIHandlers) Upload(w http.ResponseWriter, r *http.Request) {
...
-			defer snappyWriter.Close()
 			// Writing to the compressed writer.
-			_, err = io.CopyN(snappyWriter, actualReader, actualSize)
-			if err != nil {
-				hashError = err
-				return
-			}
+			_, cerr := io.CopyN(snappyWriter, actualReader, actualSize)
+			pipeWriter.CloseWithError(cerr)
+			snappyWriter.Close()
```
Praveenrajmani, line 674: The same here

Praveenrajmani (changes requested):

cmd/object-handlers.go
harshavardhana: The code was wrong in the first place, you cannot set a value on a variable inside a go-routine and read it outside. You are actively introducing a race, the way to communicate over Pipe is by sending  CloseWithError() this way the error gets propagated through the Pipe to the reader. 

Reader fails appropriately. if you look at the code it was sending error responses after setting some successful encryption headers etc.

cmd/object-handlers.go
harshavardhana: No, we shouldn't PutObject is perfectly capable of returning this error. it is the pipeReader which sees this error.

cmd/object-handlers.go
Praveenrajmani: ok .. understood

cmd/object-handlers.go
Praveenrajmani: ok

cmd/object-handlers.go
harshavardhana: This I will change, we need to fix this in couple of other locations as well.

cmd/web-handlers.go
harshavardhana: This I will change, we need to fix this in couple of other locations as well.

harshavardhana:
PTAL @balamurugana @Praveenrajmani

Praveenrajmani (approved):
LGTM . Tested over mint-compression mode also

balamurugana (approved):

minio-ops:
## Mint Automation
| Test | Result |
|:--- |:--- |

PR #6562, opened 2018-10-03
Avoid unsolicited response over idle http client

harshavardhana:
## Description
Without this PR minio server is writing an erroneous
response to clients on an idle connection which ends
up printing following message

```
Unsolicited response received on idle HTTP channel
```

## Motivation and Context
This PR would avoid sending responses on idle connections
.i.e routine network errors.

## Regression
No

## How Has This Been Tested?
This issue is tricky to reproduce, requires quite a few runs and only in distributed setup 
```sh
#!/bin/bash

uuid1="export1"
uuid2="export2"
uuid3="export3"
uuid4="export4"
mkdir -p "/tmp/$uuid1" "/tmp/$uuid2" "/tmp/$uuid3" "/tmp/$uuid4"
minio server --config-dir /tmp/config --address :9001 "http://127.0.0.1:9001/tmp/${uuid1}" "http://127.0.0.1:9002/tmp/${uuid2}" "http://127.0.0.1:9003/tmp/${uuid3}" "http://127.0.0.1:9004/tmp/${uuid4}"&
minio server --config-dir /tmp/config --address :9002 "http://127.0.0.1:9001/tmp/${uuid1}" "http://127.0.0.1:9002/tmp/${uuid2}" "http://127.0.0.1:9003/tmp/${uuid3}" "http://127.0.0.1:9004/tmp/${uuid4}"&
minio server --config-dir /tmp/config --address :9003 "http://127.0.0.1:9001/tmp/${uuid1}" "http://127.0.0.1:9002/tmp/${uuid2}" "http://127.0.0.1:9003/tmp/${uuid3}" "http://127.0.0.1:9004/tmp/${uuid4}"&
minio server --config-dir /tmp/config --address :9004 "http://127.0.0.1:9001/tmp/${uuid1}" "http://127.0.0.1:9002/tmp/${uuid2}" "http://127.0.0.1:9003/tmp/${uuid3}" "http://127.0.0.1:9004/tmp/${uuid4}"
```

Repeatedly perform `mc admin heal` to reproduce this, @vadmeste suggested even mint tests would reproduce this problem.
```
mc admin heal -r myminio
```

## Types of changes

## Checklist:

### DIFF
--- a/cmd/http/listener.go
+++ b/cmd/http/listener.go
@@ -35,9 +35,7 @@ import (
 
 var sslRequiredErrMsg = []byte("HTTP/1.0 403 Forbidden\r\n\r\nSSL required")
 
-var malformedErrMsgFn = func(data interface{}) string {
-	return fmt.Sprintf("HTTP/1.0 400 Bad Request\r\n\r\n%s", data)
-}
+var badRequestMsg = []byte("HTTP/1.0 400 Bad Request\r\n\r\n")
 
 // HTTP methods.
 var methods = []string{
@@ -97,7 +95,7 @@ func getPlainText(bufConn *BufConn) (bool, error) {
 	return false, nil
 }
 
-func getResourceHost(bufConn *BufConn, maxHeaderBytes int) (resource string, method string, host string, err error) {
+func getMethodResourceHost(bufConn *BufConn, maxHeaderBytes int) (method string, resource string, host string, err error) {
 	defer bufConn.setReadTimeout()
 
 	var data []byte
@@ -120,11 +118,15 @@ func getResourceHost(bufConn *BufConn, maxHeaderBytes int) (resource string, met
 
 		if method == "" && resource == "" {
 			if i := strings.IndexByte(tokens[0], ' '); i == -1 {
-				return "", "", "", fmt.Errorf("malformed HTTP request %s, from %s", tokens[0], bufConn.LocalAddr())
+				return "", "", "", fmt.Errorf("malformed HTTP request from '%s'", bufConn.LocalAddr())
 			}
 			httpTokens := strings.SplitN(tokens[0], " ", 3)
 			if len(httpTokens) < 3 {
-				return "", "", "", fmt.Errorf("malformed HTTP request %s, from %s", tokens[0], bufConn.LocalAddr())
+				return "", "", "", fmt.Errorf("malformed HTTP request from '%s'", bufConn.LocalAddr())
+			}
+			if !isHTTPMethod(httpTokens[0]) {
+				return "", "", "", fmt.Errorf("malformed HTTP request, invalid HTTP method '%s' from '%s'",
+					httpTokens[0], bufConn.LocalAddr())
 			}
 
 			method = httpTokens[0]
@@ -141,7 +143,7 @@ func getResourceHost(bufConn *BufConn, maxHeaderBytes int) (resource string, met
 			token = strings.ToLower(token)
 			if strings.HasPrefix(token, "host: ") {
 				host = strings.TrimPrefix(strings.TrimSuffix(token, "\r"), "host: ")
-				return resource, method, host, nil
+				return method, resource, host, nil
 			}
 		}
 
@@ -234,6 +236,7 @@ func (listener *httpListener) start() {
 				bufconn.Close()
 				return
 			}
+
 			if ok {
 				// As TLS is configured and we got plain text HTTP request,
 				// return 403 (forbidden) error.
@@ -241,8 +244,10 @@ func (listener *httpListener) start() {
 				bufconn.Close()
 				return
 			}
+
 			// As the listener is configured with TLS, try to do TLS handshake, drop the connection if it fails.
 			tlsConn := tls.Server(bufconn, listener.tlsConfig)
+
 			if err := tlsConn.Handshake(); err != nil {
 				reqInfo := (&logger.ReqInfo{}).AppendTags("remoteAddr", bufconn.RemoteAddr().String())
 				reqInfo.AppendTags("localAddr", bufconn.LocalAddr().String())
@@ -251,10 +256,11 @@ func (listener *httpListener) start() {
 				bufconn.Close()
 				return
 			}
+
 			bufconn = newBufConn(tlsConn, listener.readTimeout, listener.writeTimeout)
 		}
 
-		resource, method, host, err := getResourceHost(bufconn, listener.maxHeaderBytes)
+		method, resource, host, err := getMethodResourceHost(bufconn, listener.maxHeaderBytes)
 		if err != nil {
 			// Peek could fail legitimately when clients abruptly close
 			// connection. E.g. Chrome browser opens connections speculatively to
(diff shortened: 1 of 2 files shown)


cmd/http/listener.go
```diff
@@ -120,11 +120,15 @@ func getResourceHost(bufConn *BufConn, maxHeaderBytes int) (resource string, met
...
 			if i := strings.IndexByte(tokens[0], ' '); i == -1 {
-				return "", "", "", fmt.Errorf("malformed HTTP request %s, from %s", tokens[0], bufConn.LocalAddr())
+				return "", "", "", fmt.Errorf("malformed HTTP request, from '%s'", bufConn.LocalAddr())
 			}
 			httpTokens := strings.SplitN(tokens[0], " ", 3)
 			if len(httpTokens) < 3 {
-				return "", "", "", fmt.Errorf("malformed HTTP request %s, from %s", tokens[0], bufConn.LocalAddr())
+				return "", "", "", fmt.Errorf("malformed HTTP request, from '%s'", bufConn.LocalAddr())
+			}
+			if !isHTTPMethod(httpTokens[0]) {
```
balamurugana, line 127: This function is called after fetching HTTP method.  This is not required

harshavardhana: This was simply the old code, are you saying we should let Go HTTP handle this?

cmd/http/listener.go
```diff
@@ -120,11 +120,15 @@ func getResourceHost(bufConn *BufConn, maxHeaderBytes int) (resource string, met
 
 		if method == "" && resource == "" {
 			if i := strings.IndexByte(tokens[0], ' '); i == -1 {
-				return "", "", "", fmt.Errorf("malformed HTTP request %s, from %s", tokens[0], bufConn.LocalAddr())
+				return "", "", "", fmt.Errorf("malformed HTTP request, from '%s'", bufConn.LocalAddr())
```
balamurugana: This message needs to be beautified.

cmd/http/listener.go
harshavardhana: The reason to validate the method is one can generate something like this

```
curl -X XXX http://localhost:9000
``` 

this can clobber up our stats @balamurugana - I just moved the code from outside the function to inside to combine printing log messages.

cmd/http/listener.go
harshavardhana: can you provide an example?

cmd/http/listener.go
```diff
@@ -230,6 +234,7 @@ func (listener *httpListener) start() {
 					reqInfo.AppendTags("localAddr", bufconn.LocalAddr().String())
 					ctx := logger.SetReqInfo(context.Background(), reqInfo)
 					logger.LogIf(ctx, err)
+					bufconn.Write([]byte(malformedErrMsgFn(err)))
```
balamurugana: The decision was `close the connection for invalid HTTP requests` and no need to write error message

harshavardhana: A user already requested this @balamurugana that we are closing connections without sending an error back to the client. I don't remember if that was decided...

cmd/http/listener_test.go
```diff
@@ -657,11 +657,8 @@ func TestHTTPListenerAcceptError(t *testing.T) {
 				continue
 			}
 
-			s, err := bufio.NewReader(conn).ReadString('\n')
-			if err != nil {
-				t.Errorf("Test %d: reply read: expected = <nil>, got = %s", i+1, err)
-			} else if !strings.Contains(s, "HTTP/1.0 400 Bad Request") {
-				t.Errorf("Test %d: reply read: expected = 'HTTP/1.0 400 Bad Request' got = %s", i+1, s)
+			if _, err = bufio.NewReader(conn).ReadString('\n'); err != io.EOF {
```
balamurugana, line 660: Unable to understand what problem current check has

harshavardhana: Well the assumptions changed, the test needs to incorporate that assumption. 

- When there is a routine network timeout we do not write an error to the client which is what this PR fixes. 
- When there is a non-network timeout such as invalid HTTP message or some malformed HTTP we write an error back to the client. 

In this test the client times out @balamurugana so it will see EOF instead of a message from the server.

harshavardhana:
@balamurugana updated the PR

cmd/http/listener.go
balamurugana: Use error message as `fmt.Errorf("malformed HTTP request from '%s'", bufConn.LocalAddr())`

harshavardhana: fixed

cmd/http/listener.go
```diff
@@ -120,11 +118,15 @@ func getResourceHost(bufConn *BufConn, maxHeaderBytes int) (resource string, met
...
+				return "", "", "", fmt.Errorf("malformed HTTP request from '%s'", bufConn.LocalAddr())
 			}
 			httpTokens := strings.SplitN(tokens[0], " ", 3)
 			if len(httpTokens) < 3 {
-				return "", "", "", fmt.Errorf("malformed HTTP request %s, from %s", tokens[0], bufConn.LocalAddr())
+				return "", "", "", fmt.Errorf("malformed HTTP request from '%s'", bufConn.LocalAddr())
+			}
+			if !isHTTPMethod(httpTokens[0]) {
+				return "", "", "", fmt.Errorf("malformed HTTP request, invalid HTTP method '%s' from '%s'",
+					httpTokens[0], bufConn.LocalAddr())
```
balamurugana, line 129: This function needs to be named appropriately i.e. `getMethodResourceHost()` and returns `(method, resource, host string, err error)`

cmd/http/listener.go
```diff
@@ -230,6 +232,7 @@ func (listener *httpListener) start() {
 					reqInfo.AppendTags("localAddr", bufconn.LocalAddr().String())
 					ctx := logger.SetReqInfo(context.Background(), reqInfo)
 					logger.LogIf(ctx, err)
+					bufconn.Write(badRequestMsg)
```
balamurugana: Writing response just before closing is more appropriate.

harshavardhana: We don't need to write for network errors @balamurugana that is why this PR exists, we cannot write a response on an idle connection.

cmd/http/listener.go
harshavardhana: Done

cmd/http/listener.go
balamurugana: IMO write and close is appropriate.  Either close or handle condition differently

cmd/http/listener.go
balamurugana: Are you deferring about `(method, resource, host string, err error)`? If so, change function name accordingly

cmd/http/listener.go
harshavardhana: Not in this case, a network disconnection or a timeout doesn't mean that we should be writing back to the client which is meaningless. that is why code moved in situations only when it's not a network error we can write.

balamurugana (approved):

vadmeste (approved):
LGTM & tested

vadmeste:
Tested with mints and minio-go + tls

Issue #6582, opened 2018-10-07
SSE-KMS - KMS id doesn't exist should throw an error


harshavardhana:
## Expected Behavior
AWS S3 returns an error when a KMS id doesn't exist. 
```
$ s3cmd put /etc/issue --server-side-encryption-kms-id=02de8641-eea1-4115-8aa2-daf8b1e9b52e s3://vadmeste/obj
upload: '/etc/issue' -> 's3://vadmeste/obj'  [1 of 1]
 26 of 26   100% in    0s    37.84 B/s  done

$ s3cmd put /etc/issue --server-side-encryption-kms-id=02de8641-eea1-4115-8aa2-daf8b1e9b52f s3://vadmeste/obj
upload: '/etc/issue' -> 's3://vadmeste/obj'  [1 of 1]
 26 of 26   100% in    0s    44.89 B/s  done
ERROR: S3 error: 400 (KMS.NotFoundException): Key 'arn:aws:kms:us-east-1:804065449417:key/02de8641-eea1-4115-8aa2-daf8b1e9b52f' does not exist
```

## Current Behavior
Minio server doesn't validate the KMS id, I know that we don't have a concept currently of multiple KMS ids. But we should reject requests if the KMS id is wrong regardless of how many KMS ids we support.
```
$ s3cmd put /etc/issue --no-check-certificate --server-side-encryption-kms-id=02de8641-eea1-4115-8aa2-daf8b1e9b52e s3://vadmeste/obj
upload: '/etc/issue' -> 's3://vadmeste/obj'  [1 of 1]
 26 of 26   100% in    0s   372.35 B/s  done

$  s3cmd put /etc/issue --no-check-certificate --server-side-encryption-kms-id=1 s3://vadmeste/obj
upload: '/etc/issue' -> 's3://vadmeste/obj'  [1 of 1]
 26 of 26   100% in    0s   448.41 B/s  done
```

The actual KMS id is `1`
```
MINIO_SSE_MASTER_KEY=1:a948904f2f0f479b8f8197694b30184b0d2ed1c1cd2a1ec0fb85d299a192a447 minio server ~/test1
```

## Possible Solution
Validate the incoming KMS id

## Steps to Reproduce (for bugs)
Already explained.

## Context
SSE compatibility

## Regression
No, implementation issue. 

## Your Environment
master branch + #6488

poornas:
[...]

PR #6587, opened 2018-10-08
Fix: Disallow requests with SSE-KMS headers

poornas:
Addresses issue #6582. Minio server currently does not
have SSE-KMS support. Reject requests with SSE-KMS headers
with NotImplementedErr

## Motivation and Context

minio server doesn't have SSE-KMS support. However current sse-s3 implementation allows ambiguity to creep in by not rejecting requests with SSE-KMS header. This will cause confusion because Put/Copy requests with SSE-KMS headers silently get encrypted with sse-s3 instead.

## Regression

Wrong implementation in #6234 where NotImplementedErr is raised for Put/Copy requests when ObjectLayer doesnt allow encryption, but in fact should be raised irrespective of object layer support for encryption (as server does not yet support it, and encryption headers are not being passed to gateway)

## How Has This Been Tested?

Start minio server with vault as KMS enabled. Any Get/Put/Copy request with SSE-KMS headers should fail with NotImplementedErr
```
$ aws --endpoint-url http://localhost:9000 s3api put-object --body /home/kris/data/datafile-10-kB --bucket tbucket11 --key datafile-10-kB --ssekms-key-id 1
```

## Types of changes

## Checklist:

### DIFF
--- a/cmd/object-handlers.go
+++ b/cmd/object-handlers.go
@@ -84,7 +84,10 @@ func (api objectAPIHandlers) SelectObjectContentHandler(w http.ResponseWriter, r
 		writeErrorResponse(w, ErrServerNotInitialized, r.URL)
 		return
 	}
-
+	if crypto.S3KMS.IsRequested(r.Header) { // SSE-KMS is not supported
+		writeErrorResponse(w, ErrNotImplemented, r.URL)
+		return
+	}
 	vars := mux.Vars(r)
 	bucket := vars["bucket"]
 	object := vars["object"]
@@ -620,7 +623,7 @@ func (api objectAPIHandlers) CopyObjectHandler(w http.ResponseWriter, r *http.Re
 		writeErrorResponse(w, ErrServerNotInitialized, r.URL)
 		return
 	}
-	if !objectAPI.IsEncryptionSupported() && crypto.S3KMS.IsRequested(r.Header) {
+	if crypto.S3KMS.IsRequested(r.Header) {
 		writeErrorResponse(w, ErrNotImplemented, r.URL) // SSE-KMS is not supported
 		return
 	}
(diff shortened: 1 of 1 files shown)


harshavardhana (dismissed):

cmd/object-handlers.go
```diff
@@ -620,7 +623,7 @@ func (api objectAPIHandlers) CopyObjectHandler(w http.ResponseWriter, r *http.Re
 		writeErrorResponse(w, ErrServerNotInitialized, r.URL)
 		return
 	}
-	if !objectAPI.IsEncryptionSupported() && crypto.S3KMS.IsRequested(r.Header) {
```
aead, line 623: We must not remove the `!objectAPI.IsEncryptionSupported()` because gateway backend could support SSE-KMS (AWS).
However the logic was wrong. Correct is: `!objectAPI.IsEncryptionSupported() || crypto.S3KMS.IsRequested(r.Header)`

aead (changes requested):

cmd/object-handlers.go
harshavardhana: We can bring in later when gateway indeed supports such a functionality @aead

harshavardhana: Otherwise `!objectAPI.IsEncryptionSupported() || crypto.S3KMS.IsRequested(r.Header)` we are rejecting all Put/Copy operations if backend doesn't support encrypted.

poornas: @harshavardhana is correct - for passing SSE-KMS headers to backend in the event it supports encryption, we will need to be able to distinguish we are running in gateway mode at the objectAPI interface, but think we can defer that for later

aead: Okay, agree. Actually the gateway functionality is totally broken - if a client sends SSE-KMS the backend does not see these headers even if it would support it (AWS) because of minio doesn't forward it. At least the client sees an error msg now.

aead (dismissed):

cmd/object-handlers.go
```diff
@@ -84,7 +84,10 @@ func (api objectAPIHandlers) SelectObjectContentHandler(w http.ResponseWriter, r
 		writeErrorResponse(w, ErrServerNotInitialized, r.URL)
 		return
 	}
-
+	if crypto.S3KMS.IsRequested(r.Header) { // SSE-KMS is not supported
+		writeErrorResponse(w, ErrBadRequest, r.URL)
```
kannappanr: Should this be returning `ErrNotimplemented` instead of `ErrBadRequest`?

poornas: changed it to make it consistent

harshavardhana (approved):

aead (approved):

kannappanr (approved):

PR #6652, opened 2018-10-17
fix wrong actual part size assignment in CopyObjectPart

aead:
## Description
This commit fixes a wrong assignment to `actualPartSize`.
The `actualPartSize` for an encrypted src object is not `srcInfo.Size`
because that's the encrypted object size which is larger than the
actual object size. So the actual part size for an encrypted
object is the decrypted size of `srcInfo.Size`.

## Motivation and Context
SSE, CopyObjectPart

## Regression
No

## Types of changes

## Checklist:

### DIFF
--- a/cmd/object-handlers.go
+++ b/cmd/object-handlers.go
@@ -1454,11 +1454,17 @@ func (api objectAPIHandlers) CopyObjectPartHandler(w http.ResponseWriter, r *htt
 	defer gr.Close()
 	srcInfo := gr.ObjInfo
 
-	var actualPartSize int64
-	actualPartSize = srcInfo.Size
+	actualPartSize := srcInfo.Size
+	if crypto.IsEncrypted(srcInfo.UserDefined) {
+		actualPartSize, err = srcInfo.DecryptedSize()
+		if err != nil {
+			writeErrorResponse(w, toAPIErrorCode(err), r.URL)
+			return
+		}
+	}
 
 	// Special care for CopyObjectPart
-	if partRangeErr := checkCopyPartRangeWithSize(rs, srcInfo.Size); partRangeErr != nil {
+	if partRangeErr := checkCopyPartRangeWithSize(rs, actualPartSize); partRangeErr != nil {
 		writeCopyPartErr(w, partRangeErr, r.URL)
 		return
 	}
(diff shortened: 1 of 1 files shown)


poornas (approved):
LGTM

vadmeste:
@aead this PR simplifies the code, but I just wanted to make sure what functional change does it do ?

According to my understanding, only checkCopyPartRangeWithSize() will behave correctly but everything else will behave the same, what do you think ?

aead:
@vadmeste The patch also changes the computation of `rs.GetOffsetLength(actualPartSize)`.
If `actualPartSize` is the encrypted size of the object (behavior before this patch) - which is the case if src is an encrypted object - than `length` become that encrypted size.
So the important difference is that:
 1. `actualPartSize` was the encrypted size in case of src is encrypted
 2. `length` is computed based on the `actualPartSize` which is only correct if `actualPartSize` is the real size.

vadmeste:
@aead actually, prior to this PR, length is still updated correctly if encryption is enabled, look at this https://github.com/minio/minio/pull/6652/commits/a9ae3ed0ccf47ecd79786a8dcbbf1901539f211c#diff-d1c4dccc78d8d9f128ca8972a1199176L1485

I just wanted to see if this PR fixes any bug because we have an ongoing investigation about a possible decryption issue, can you imagine a scenario which introduces a bug and this PR would correct it ?

aead:
Actually you're right @vadmeste The length is updated correctly. However the `actualPartSize` is computed the wrong way but this does not affect correctness because the destination object seems to be never compressed (Any reason @Praveenrajmani ?!)

poornas:
@aead , if it is compressed, it would never enter this code path since we don't allow encryption and compression together.

aead:
@poornas Sure, but the source could be encrypted while the destination could be compressed.
In this case it would store the encrypted size of `srcInfo` as `actualPartSize`. (If it would compress)

Praveenrajmani:
> Actually you're right @vadmeste The length is updated correctly. However the `actualPartSize` is computed the wrong way but this does not affect correctness because the destination object seems to be never compressed (Any reason @Praveenrajmani ?!)

Yes @aead  the length is updated correctly, But eventhough the actualpartsize wont affect the encryption flow anywhere (unreachable for encryption), actualpartsize should be assigned correctly with the decrypted size for the future use. 
And for the Copy and CopyPart, The destination object is compressed only if the source object is compressed which is wrong. We should compress the destination based on the config option. Will do a PR for this.
Let me cover this in my PR itself, We can even close this

Praveenrajmani:
> @poornas Sure, but the source could be encrypted while the destination could be compressed.
> In this case it would store the encrypted size of `srcInfo` as `actualPartSize`. (If it would compress)

Yes in this case we should be having actualsize as the decrypted size

aead:
@Praveenrajmani Is it on purpose that I cannot create a compressed destination object using CopyObjectPart API ?

Praveenrajmani:
> @Praveenrajmani Is it on purpose that I cannot create a compressed destination object using CopyObjectPart API ?

No @aead , its not on purpose. The current implementation wont work if the source object is encrypted. This should be fixed. Working on it now.

harshavardhana:
can this PR be taken @vadmeste ?

vadmeste (approved):
LGTM

Praveenrajmani:
This PR#6669 covers this change @harshavardhana @vadmeste

harshavardhana:
@aead can you close it?

harshavardhana:
> This PR#6669 covers this change @harshavardhana @vadmeste

@Praveenrajmani we can take this PR and you work your changes on top of this?

Issue #4812, opened 2017-08-14
[Proposal] Reorganize codebase in a more idiomatc way


aead:
Currently `github.com/minio/minio` contains the main.go (and some build files, etc.) and everything else is located under `github.com/minio/minio/cmd`.
This has the following consequences:
 - It's tough for external contributors to dive into minio and understand the layer-structure of the minio server/gateway - it took me quite a while and I'm working currently full time on the minio server.
 - Build times are quite long - even tough the Go compiler is pretty fast.
 - Currently /cmd exports stuff more or less randomly.
 - It is very hard to test functionality properly because /cmd is one global namespace and you cannot rely on invariants. Somebody can modify every state from everywhere in the code.

Go natural approach to face these kind of problems is the package. Packages provide:
 - A structured way to organize code. It's easier to understand the architecture of a project if you can look at it's components and see how they interact. This helps new (and old :wink: ) contributors to figure out which part of the code is responsible for what.
- Faster build and test times. The go compiler does incremental builds by default. It only compiles packages which have changed. This sames time during development.
- Go supports internal (private) packages. Currently somebody can import `github.com/minio/minio/cmd` and may raise issues because we're breaking the code after every release. We can prevent misuse of minio/cmd by using internal packages.
- A package exports a public API used by other packages to build minio's functionality but we can also protect state and rely on certain invariants and assumptions. Packages will help us to prevent bugs by accidentally changing invariants.

## Expected Behavior
There a many Go projects which follow a package-based approach:
 - [Caddy](https://github.com/mholt/caddy)
 - [Kubernetes](https://github.com/kubernetes/kubernetes)
 - [Prometheus](https://github.com/prometheus/prometheus)
...

## Possible Solution
I propose to introduce packages for minio to face the listed problems. The packages should mirror the architecture of minio. Minio consists of 3 independent layers: storage, object and handler. Further minio supports FS and XL as well as Gateway (GCS, Azure and AWS). Within XL erasure coding and bitrot protection is a special/complex code base.

We should discuss about the right structure of packages and keep the balance between useful abstractions and over-engineering but I think the general structure should look like this:
```
minio/
   - main.go
   - ...
   - storage/
          - storage.go
          - storage_test.go
          - posix/
                 - posix.go
                 - ...
                 - posix_test.go
           - rpc/
                - rpc.go
                - ...
   - object/
         - object.go
         - fs/
               - fs.go
         - xl/
               - xl.go

.... 
```
[...]

PR #6682, opened 2018-10-22
Fix goroutine test fatalf

leitzler:
## Description
This PR replaces `t.Fatal` and `t.Fatalf` with `t.Error` and `t.Errorf` when called from within a goroutine.

## Motivation and Context
`t.Fatal` and `t.Fatalf` must be called from the same goroutine running the Test function, see https://godoc.org/testing#T.

## How Has This Been Tested?
`make test`

## Types of changes

## Checklist:

### DIFF
--- a/cmd/http/bufconn_test.go
+++ b/cmd/http/bufconn_test.go
@@ -47,6 +47,7 @@ func TestBuffConnReadTimeout(t *testing.T) {
 		tcpConn, terr := tcpListener.AcceptTCP()
 		if terr != nil {
 			t.Errorf("failed to accept new connection. %v", terr)
+			return
 		}
 		bufconn := newBufConn(tcpConn, 1*time.Second, 1*time.Second)
 		defer bufconn.Close()
@@ -56,10 +57,12 @@ func TestBuffConnReadTimeout(t *testing.T) {
 		_, terr = bufconn.Read(b)
 		if terr != nil {
 			t.Errorf("failed to read from client. %v", terr)
+			return
 		}
 		received := string(b)
 		if received != "message one\n" {
 			t.Errorf(`server: expected: "message one\n", got: %v`, received)
+			return
 		}
 
 		// Wait for more than read timeout to simulate processing.
@@ -68,16 +71,19 @@ func TestBuffConnReadTimeout(t *testing.T) {
 		_, terr = bufconn.Read(b)
 		if terr != nil {
 			t.Errorf("failed to read from client. %v", terr)
+			return
 		}
 		received = string(b)
 		if received != "message two\n" {
 			t.Errorf(`server: expected: "message two\n", got: %v`, received)
+			return
 		}
 
 		// Send a response.
 		_, terr = io.WriteString(bufconn, "messages received\n")
 		if terr != nil {
 			t.Errorf("failed to write to client. %v", terr)
+			return
 		}
 
 		// Removes all deadlines if any.
--- a/cmd/http/bufconn_test.go
+++ b/cmd/http/bufconn_test.go
@@ -46,7 +46,7 @@ func TestBuffConnReadTimeout(t *testing.T) {
 
 		tcpConn, terr := tcpListener.AcceptTCP()
 		if terr != nil {
-			t.Fatalf("failed to accept new connection. %v", terr)
+			t.Errorf("failed to accept new connection. %v", terr)
 		}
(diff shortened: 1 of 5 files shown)


harshavardhana:
what bug does this fix @leitzler ?

leitzler:
Several tests that isn’t working as intended since they call `t.Fatalf` from within a goroutine that isn’t the one executing the test. That won’t work.
But since the tests pass right now it’s not visible.

leitzler:
Check out the docs link I provided in the description. `t.Error` is ok to call from multiple goroutines, `t.Fatal` (and others that is supposed to fail fast) isn’t ok.

harshavardhana:
> everal tests that isn’t working as intended since they call `t.Fatalf` from within a goroutine that isn’t the one executing the test. That won’t work.
> But since the tests pass right now it’s not visible.

Why would they not work? I don't see a CI issue at this moment this PR fixes. Are you saying this is an improvement?

leitzler:
Yes, if a bug is introduced here the bug (in the tests) will show.

leitzler:
They won’t work since `t.Fatal` is supposed to fail fast, and end the test running. It won’t when called from a spawned goroutine.

balamurugana (changes requested):
@leitzler 
> They won’t work since `t.Fatal` is supposed to fail fast, and end the test running. It won’t when called from a spawned goroutine.

`t.Fatalf()` is cautiously used to fail fast in goroutines.  As there is no problem seen so far with this code, I couldn't able to see why this chance is required.

If you provide a separate test program why bad to use `f.Fatalf()` in goroutine is better.

kannappanr:
@balamurugana There are other projects where they have done this

https://github.com/golang/go/issues/17900

https://github.com/ipfs/go-ipfs/issues/2043

balamurugana:
@kannappanr Why aren't we facing those issues?

leitzler:
@balamurugana That code is only executed if a test fails. And there are no failing tests now.

balamurugana:
@leitzler What will happen on failure?

leitzler:
The change is submitted to increase overall code quality (you could see it as a part of issue #4812), since the documentation clearly states that:
> A test ends when its Test function returns or calls any of the methods FailNow, Fatal, Fatalf, SkipNow, Skip, or Skipf. Those methods, as well as the Parallel method, must be called only from the goroutine running the Test function.

Maintaining good code quality will minimize the risk of introducing new bugs, and will make it easier for new contributors to understand the code.

This PR will not alter the behavior. A common and idiomatic way to handle this is to have an error channel that the goroutine communicates the error from, but that is another change. First of all the code should do what is says, then additional improvement can be added.

Here is an narrowed down concrete example:
```go
package main

import (
	"testing"
	"time"
)

func TestHelloWorld(t *testing.T) {
	go func() {
		t.Fatal("Fail")
	}()

	time.Sleep(time.Second)

	t.Log("This line should not be called since the purpose of Fatal is to fail fast")
}
```

```bash
$ go test . 
--- FAIL: TestHelloWorld (1.00s)
    main_test.go:11: Fail
    main_test.go:16: This line should not be called since the purpose of Fatal is to fail fast
FAIL
FAIL	_/tmp/test	1.003s
```

balamurugana:
Previously I encountered that `testing.T` was not thread-safe and had custom `fail()` function (unfortunately I couldn't able to find in the current master).  In later time, `testing.T` doesn't show up.  Hence `bufconn_test.go` (As I wrote consciously) used `t.Fatal()` in goroutines.

I consider this PR is trying to deal with `testing.T` thread safety, but replacing `t.Fatal()` with `t.Error()` is insufficient i.e. `t.Fatal()` stops executing next checks.  This needs to be handled correctly in this PR.

leitzler:
You are indeed right, it does alter the behaviour in the case that the goroutine will continue. But calling `t.Fatal` won't stop the test as it is supposed to when called from a goroutine. That's why the documentation states that you **must not** use it that way, it doesn't work.

I could add returns after `t.Error` calls within goroutines, then the behaviour won't change but the code will actually do what it says. If you truly want a fail fast behaviour the tests must be redesigned. This PR only solves the fact that the code doesn't do what it says since it is used in an incorrect way.

balamurugana (approved):

minio-ops:
## Mint Automation
| Test | Result |
|:--- |:--- |

harshavardhana (approved):

kannappanr (approved):

Issue #6704, opened 2018-10-25
HTTP2 support is broken / not implemented


leitzler:
## Expected Behavior
Improvement suggestion: minio should support HTTP2.

## Current Behavior
Major browsers as well as curl supports both HTTP/1.1 as well as HTTP2, but due to a bug in `cmd/http/server.go`, minio will never serve HTTP2 if the client support HTTP/1.1.

If a client would only support HTTP2, minio terminate all requests without response while logging `Error: malformed HTTP request from n.n.n.n:nnnn" 

## Possible Solution
Make sure that the server doesn't imply that is supports HTTP2 (as it does right now). I'll submit a PR for this step. This must be done regardless of if HTTP2 support is added.

When (if) HTTP2 support is added, make sure that `tls.Config.NextProtos` is set to `[]string{"h2", "http/1.1"}` instead of `[]string{"http/1.1", "h2"}` since the list is ordered by preference.

## Steps to Reproduce (for bugs)
Create a request using curl with verbose output to see that HTTP2 isn't used:
```bash
$ curl -v https://minio-host/
[...]
* ALPN, offering h2
* ALPN, offering http/1.1
[...]
* ALPN, server accepted to use http/1.1
[...]
GET / HTTP/1.1
[...]
```

You could also try with a minimal H2 client written in Go to reproduce the incorrect handling of H2 requests:
```go
package main

import (
	"crypto/tls"
	"log"
	"net/http"

	"golang.org/x/net/http2"
)

func main() {
	c := &http.Client{
		Transport: &http2.Transport{
			TLSClientConfig: &tls.Config{
				InsecureSkipVerify: true,
				NextProtos:         []string{"h2"},
			},
		},
	}

	_, err := c.Get("https://127.0.0.1:9000")
	if err != nil {
		log.Fatal(err)
	}
}
```
```bash
$ go run main.go
2018/10/25 14:11:00 Get https://127.0.0.1:9000: unexpected EOF
```
```
API: SYSTEM()
Time: 14:11:00 CEST 10/25/2018
Error: malformed HTTP request from 127.0.0.1:9000
       remoteAddr=127.0.0.1:44256, localAddr: 127.0.0.1:9000
```

## Your Environment

* Version used (`minio version`): `latest`
* Environment name and version (e.g. nginx 1.9.1): -
* Server type and version: Linux / macOS
[...]

PR #6705, opened 2018-10-25
Remove h2 from tls.Config.NextProtos as it doesn't work

leitzler:
## Description
This PR removes `h2` support since it is broken/not implemented. It makes sure that H2 clients get an appropriate response instead of just being terminated. 

## Motivation and Context
See #6704

## How Has This Been Tested?
`make test` & requests from a pure H2 client.

## Types of changes

## Checklist:


minio-ops:
## Mint Automation
| Test | Result |
|:--- |:--- |

balamurugana:
Blocking this as `h2` is there from day one.  It requires wider discussion.

vadmeste:
If I recall correctly, removing "h2" will disallow clients to perform HTTP 2 requests while securing it with TLS, since client & server needs to agree on the used protocol in TLS.

@leitzler could you explain how you concluded that http/2 is currently broken ?

harshavardhana (changes requested):
Please provide a test program for proof.

leitzler:
@vadmeste yes, that's right. The issue described in #6704 is two parted. Part one is that `tls.Config.NextProtos` is set to `[]string{"http/1.1", "h2"}` instead of `[]string{"h2", "http/1.1"}`. That's a pure bug. The configuration is a list of preferred order, so "h2" must come before "http/1.1" since almost all clients support both. All major browsers and curl support both and will currently never use HTTP2 since "http/1.1" comes first.

So, if you want to test out HTTP2, change `"http/1.1", "h2"` to `"h2", "http/1.1"` in `cmd/http/server.go`, recompile and start (in https mode). Then try to access minio using curl or a browser.

Then you'll run into the next problem, the one described in #6704. Minio terminates H2 calls since it thinks that the request isn't a valid http request (the error you get is from `cmd/http/listener.go` and the message is `Error: malformed HTTP request from n.n.n.n:nnnn`.

I guess the reason for this is that HTTP2 uses multiplexed frames, and inside `listener.go` there is a `getMethodResourceHost(...)` that is trying to parse HTTP/1.1 requests. It probably gets the HTTP2 connection preface (`PRI * HTTP/2.0\r\n\r\nSM\r\n\r\n`) and then a `SETTINGS` frame that is binary.
And if I understand it correctly, minio uses the `tls` package to decrypt the incoming request, but has no code for handling `HTTP2` frames at all. That's why I wrote #6704.

@harshavardhana sure! Either you could just set up minio in "https-mode" and use the client I provided in the issue:
```go
package main

import (
	"crypto/tls"
	"log"
	"net/http"

	"golang.org/x/net/http2"
)

func main() {
	c := &http.Client{
		Transport: &http2.Transport{
			TLSClientConfig: &tls.Config{
				InsecureSkipVerify: true,
				NextProtos:         []string{"h2"},
			},
		},
	}

	_, err := c.Get("https://127.0.0.1:9000")
	if err != nil {
		log.Fatal(err)
	}
}
```

Or you could fix the ordering bug in `tls.Config.NextProtos`, so that `h2` comes before `http/1.1`. Then you can use any browser or curl.

harshavardhana:
I explored more on this topic there are a few things which makes H2 support a problem for us right now, the problem is due to the fact that we read from the underlying conn well before the request is fully read by the server. We read a few tokens to understand if its a TLS connection etc. 

While the code actually works perfectly fine for http/1.1 - Go http2 implementation cannot handle it, i.e if you provide a conn on which we already did tls.Handshake() 

All these are special requirements for Minio and Go standard package seems to be doing something of its own which leads to odd scenarios where things don't work properly. Surprisingly all of our TLS is working properly because in our NextProtos we default to http/1.1 - this ordering seems to have silently helped us never discover this bug. 

In fact, this bug has existed as long as we have had the custom HTTP implementation where we peek into the incoming connection. 

Removing `blocked` label since it's not applicable, this PR is valid until we properly fix the HTTP 2.0 support.

harshavardhana (approved):

balamurugana (approved):

vadmeste:
I just checked this random version RELEASE.2017-01-25T03-14-52Z, HTTP/2 is working fine with TLS, it looks like we need more sophisticated code to parse http/2

PR #6729, opened 2018-10-29
No error in parse location constraint

diddiman-zz:
Fixed so all types of errors when parsing location constraint is returned

## Description

Just changed from returning only EOF errors to all errors as long as the content length is larger than zero

## Motivation and Context

To be able to return the correct HTTP Error code, we need to know what actually went wrong with the request

## Regression

No

## How Has This Been Tested?

Tested with old and new unit tests

## Types of changes

## Checklist:

### DIFF
--- a/cmd/handler-utils.go
+++ b/cmd/handler-utils.go
@@ -39,7 +39,7 @@ func parseLocationConstraint(r *http.Request) (location string, s3Error APIError
 	// be created at default region.
 	locationConstraint := createBucketLocationConfiguration{}
 	err := xmlDecoder(r.Body, &locationConstraint, r.ContentLength)
-	if err != nil && err != io.EOF {
+	if err != nil && r.ContentLength != 0 {
 		logger.LogIf(context.Background(), err)
 		// Treat all other failures as XML parsing errors.
 		return "", ErrMalformedXML
--- a/cmd/handler-utils_test.go
+++ b/cmd/handler-utils_test.go
@@ -19,6 +19,7 @@ package cmd
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"encoding/xml"
 	"io/ioutil"
 	"net/http"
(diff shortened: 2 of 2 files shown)


cmd/handler-utils.go
```diff
@@ -39,7 +39,7 @@ func parseLocationConstraint(r *http.Request) (location string, s3Error APIError
 	// be created at default region.
 	locationConstraint := createBucketLocationConfiguration{}
 	err := xmlDecoder(r.Body, &locationConstraint, r.ContentLength)
-	if err != nil && err != io.EOF {
+	if err != nil && r.ContentLength != 0 {
```
harshavardhana, line 42: This is not correct, location constraint is optional. The previous code is correct

diddiman-zz: Yeah. The idea was to check if it's bigger than zero, since that should cover the case when it's not passed. must have missed that somewhere in the process

cmd/handler-utils_test.go
```diff
@@ -64,18 +63,37 @@ func TestIsValidLocationContraint(t *testing.T) {
...
+		bucketConfig       interface{}
+		serverConfigRegion string
+		expectedCode       APIErrorCode
 	}{
 		// Test case - 1.
-		{globalMinioDefaultRegion, globalMinioDefaultRegion, ErrNone},
+		{createBucketLocationConfiguration{Location: globalMinioDefaultRegion}, globalMinioDefaultRegion, ErrNone},
 		// Test case - 2.
 		// In case of empty request body ErrNone is returned.
-		{"", globalMinioDefaultRegion, ErrNone},
```
harshavardhana, line 75: This is an expected scenario @diddiman this PR removes this test incorrectly.

harshavardhana (changes requested):
r.ContentLength is already used in io.LimitReader() to limit the reader properly, we don't need to look at r.ContentLength here again

I would suggest you can close this PR

diddiman-zz:
I would suggest that the test case with empty body is re added, because as it is now, that test case should still be valid and give the expected result. Also change from != 0 to > 0 to be safe

The problem that i'm trying to fix is to give back an error when the user passes a body that is incorrect, for example just a string of random content or a json. Because now that returns an ErrNone.

harshavardhana:
A new PR has been sent at #7179 with some more tests and cleanup.

PR #6850, opened 2018-11-22
Fix CloudBerry rename object failed

ssuio:
CloudBerry rename object does PUT request and query acl key. Server  responses not implement causing rename failed.

1. Make acl query pass through dummy api.
2. Create putObject/Bucket dummy api.

## Description
CloudBerry rename object does PUT request in the last step.
1. Add put method when generic-handler checking queries with acl key.
2. Create putObject/Bucket dummy api.

## Motivation and Context
Cloud berry explorer show failed when renaming objects(Actually it did rename). But server still return "Not implement" response.

## Regression
 No

## How Has This Been Tested?
Rename an object by Cloud berry explorer. And make sure no failed msg popup.

## Types of changes

- [x ] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [x ] My code follows the code style of this project.
- [x ] All new and existing tests passed.

### DIFF
--- a/cmd/acl-handlers.go
+++ b/cmd/acl-handlers.go
@@ -99,6 +99,51 @@ func (api objectAPIHandlers) GetBucketACLHandler(w http.ResponseWriter, r *http.
 	w.(http.Flusher).Flush()
 }
 
+func (api objectAPIHandlers) PutBucketACLHandler(w http.ResponseWriter, r *http.Request) {
+	ctx := newContext(r, w, "PutBucketACL")
+
+	defer logger.AuditLog(w, r, "PutBucketACL", mustGetClaimsFromToken(r))
+
+	vars := mux.Vars(r)
+	bucket := vars["bucket"]
+
+	objAPI := api.ObjectAPI()
+	if objAPI == nil {
+		writeErrorResponse(w, ErrServerNotInitialized, r.URL)
+		return
+	}
+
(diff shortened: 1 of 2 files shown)


nitisht:
Thanks for the PR @ssuio

harshavardhana:
We are not going to implement PUT dummy handlers they do not make sense. We only did that for GET because we can return back default value. We cannot do that with PUT since it's a mutation of server state. Faking a PUT operation is hacky and doesn't serve any purpose.

CBB should handle this either on client or you should use Minio specific connector - https://www.cloudberrylab.com/resources/blog/support-for-minio-in-cloudberry-backup-5-6/

harshavardhana (changes requested):
:-1: we shouldn't do this

ssuio:
Got it. But I am using Cloud berry explorer, not backup one. And I think rename is a basic operation.
I tested Openstack swift and Amazon S3, they both can rename successfully. What changing should I do to make Minio support it too ?

harshavardhana:
So here is our take - ACL is legacy functionality and we are not planning to implement it. ACLs are tricky and buggy to implement in S3 API. Amazon S3 recommends apps to move to bucket policies instead. So we implement this instead.

In this situation i would suggest that Cloud Berry explorer should migrate from legacy implementation. Implementing in the manner in this PR will not be acceptable since it is not a proper implementation.

ssuio:
Thanks for the reply.

PR #6861, opened 2018-11-24
Cleanup Kubernetes documentation

nitisht:
## Description
Also add details on why Readiness checks are not recommended for Minio
StatefulSets.

## Motivation and Context
Based on discussion with community members on Slack.

## Regression
No

## How Has This Been Tested?
Using the sample yaml file on local Minikube deployment

## Types of changes

## Checklist:

### DIFF
--- a/docs/orchestration/kubernetes/README.md
+++ b/docs/orchestration/kubernetes/README.md
@@ -8,35 +8,19 @@ Kubernetes concepts like Deployments and StatefulSets provide perfect platform t
 
 ## Table of Contents
 
-- [Prerequisites](#prerequisites)
+- [Prerequisites](#Prerequisites)
 - [Minio Standalone Server Deployment](#minio-standalone-server-deployment)
-    - [Standalone Quickstart](#standalone-quickstart)
-    - [Create Persistent Volume Claim](#create-persistent-volume-claim)
-    - [Create Deployment](#create-minio-deployment)
-    - [Create LoadBalancer Service](#create-minio-service)
-  - [Update existing Minio Deployment](#update-existing-minio-deployment)
-  - [Resource cleanup](#standalone-resource-cleanup)
-
 - [Minio Distributed Server Deployment](#minio-distributed-server-deployment)
-    - [Distributed Quickstart](#distributed-quickstart)
-    - [Create Minio Headless Service](#create-minio-headless-service)
-    - [Create Minio Statefulset](#create-minio-statefulset)
-    - [Create LoadBalancer Service](#create-minio-service)
-  - [Update existing Minio StatefulSet](#update-existing-minio-statefulset)
-  - [Deploying on cluster nodes with local host path](#deploying-on-cluster-nodes-with-local-host-path)
-  - [Resource cleanup](#distributed-resource-cleanup)
-
 - [Minio GCS Gateway Deployment](#minio-gcs-gateway-deployment)
-    - [GCS Gateway Quickstart](#gcs-gateway-quickstart)
-    - [Create GCS Credentials Secret](#create-gcs-credentials-secret)
-    - [Create Minio GCS Gateway Deployment](#create-minio-gcs-gateway-deployment)
-    - [Create Minio LoadBalancer Service](#create-minio-loadbalancer-service)
-    - [Update Existing Minio GCS Deployment](#update-existing-minio-gcs-deployment)
-  - [Resource cleanup](#gcs-gateway-resource-cleanup) 
+- [Monitoring Minio in Kubernetes](#monitoring-minio)
+
+<a name="Prerequisites"></a>
 
 ## Prerequisites
 
 To run this example, you need Kubernetes version >=1.4 cluster installed and running, and that you have installed the [`kubectl`](https://kubernetes.io/docs/tasks/kubectl/install/) command line tool in your path. Please see the [getting started guides](https://kubernetes.io/docs/getting-started-guides/) for installation instructions for your platform.
+
+<a name="minio-standalone-server-deployment"></a>
 
 ## Minio Standalone Server Deployment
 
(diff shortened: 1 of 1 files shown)


harshavardhana (dismissed):
LGTM

docs/orchestration/kubernetes/README.md
```diff
@@ -61,31 +45,9 @@ kubectl create -f https://github.com/minio/minio/blob/master/docs/orchestration/
...
-    # The volume is mounted as read-write by a single node
-    - ReadWriteOnce
-  resources:
-    # This is the request for storage. Should be available in the cluster.
-    requests:
-      storage: 10Gi
-```
-
-Create the PersistentVolumeClaim
+Create a persistent volume claim (PVC) to request storage for the Minio instance. Kubernetes looks out for PVs matching the PVC request in the cluster and binds it to the PVC automatically. Create the PersistentVolumeClaim like so
```
kannappanr: `Create the PersistentVolumeClaim like so` doesn't sound correct.

docs/orchestration/kubernetes/README.md
```diff
@@ -94,75 +56,7 @@ persistentvolumeclaim "minio-pv-claim" created
...
-        livenessProbe:
-          httpGet:
-            path: /minio/health/live
-            port: 9000
-          initialDelaySeconds: 120
-          periodSeconds: 20
-```
-
-Create the Deployment
+A deployment encapsulates replica sets and pods — so, if a pod goes down, replication controller makes sure another pod comes up automatically. This way you won’t need to bother about pod failures and will have a stable Minio service available. Create the Deployment like so
```
kannappanr: Should `A deployment encapsulates replica sets and pods — so, if a pod goes down` be changed to `A deployment encapsulates replica sets and pods. If a pod goes down`. 

`Create the Deployment like so` needs to be modified.

docs/orchestration/kubernetes/README.md
```diff
@@ -173,25 +67,7 @@ deployment "minio-deployment" created
...
-  ports:
-    - port: 9000
-      targetPort: 9000
-      protocol: TCP
-  selector:
-    # Looks for labels `app:minio` in the namespace and applies the spec
-    app: minio
-```
-Create the Minio service
+In this example, we expose the Minio Deployment by creating a LoadBalancer service. Create the Minio service like so
```
kannappanr: `Create the Minio service like so` can be changed to `Create the minio service using the following commands`

docs/orchestration/kubernetes/README.md
```diff
@@ -252,27 +130,7 @@ kubectl create -f https://github.com/minio/minio/blob/master/docs/orchestration/
...
-  clusterIP: None
-  ports:
-    - port: 9000
-      name: minio
-  selector:
-    app: minio
-```
-
-Create the Headless Service
+Headless Service controls the domain within which StatefulSets are created. The domain managed by this Service takes the form: `$(service name).$(namespace).svc.cluster.local` (where “cluster.local” is the cluster domain), and the pods in this domain take the form: `$(pod-name-{i}).$(service name).$(namespace).svc.cluster.local`. This is required to get a DNS resolvable URL for each of the pods created within the Statefulset. Create the Headless Service like so
```
kannappanr: `Create the Headless Service like so` needs to corrected.

docs/orchestration/kubernetes/README.md
```diff
@@ -281,71 +139,7 @@ service "minio" created
...
-    spec:
-      accessModes:
-        - ReadWriteOnce
-      resources:
-        requests:
-          storage: 10Gi
-```
-
-Create the Statefulset
+A StatefulSet provides a deterministic name and a unique identity to each pod, making it easy to deploy stateful distributed applications. To launch distributed Minio you need to pass drive locations as parameters to the minio server command. Then, you’ll need to run the same command on all the participating pods. StatefulSets offer a perfect way to handle this requirement. Create the Statefulset like so
```
kannappanr: `Create the Statefulset like so` needs to be corrected.

docs/orchestration/kubernetes/README.md
```diff
@@ -354,25 +148,7 @@ statefulset "minio" created
...
-  type: LoadBalancer
-  ports:
-    - port: 9000
-      targetPort: 9000
-      protocol: TCP
-  selector:
-    app: minio
-```
-Create the Minio service
+Now that you have a Minio statefulset running, you may either want to access it internally (within the cluster) or expose it as a Service onto an external (outside of your cluster, maybe public internet) IP address, depending on your use case. You can achieve this using Services. There are 3 major service types — default type is ClusterIP, which exposes a service to connection from inside the cluster. NodePort and LoadBalancer are two types that expose services to external traffic. Create the Minio service like so
```
kannappanr: `You can achieve this using Services.` can be changed to `This can be achieved using services.` Also, please fix `Create the minio service like so`

docs/orchestration/kubernetes/README.md
```diff
@@ -552,25 +282,7 @@ deployment "minio-deployment" created
...
-  type: LoadBalancer
-  ports:
-    - port: 9000
-      targetPort: 9000
-      protocol: TCP
-  selector:
-    app: minio
-```
-Create the Minio service
+Now that you have a Minio deployment running, you may either want to access it internally (within the cluster) or expose it as a Service onto an external (outside of your cluster, maybe public internet) IP address, depending on your use case. You can achieve this using Services. There are 3 major service types — default type is ClusterIP, which exposes a service to connection from inside the cluster. NodePort and LoadBalancer are two types that expose services to external traffic. Create the Minio service like so
```
kannappanr: Please fix `Create the Minio service like so`

docs/orchestration/kubernetes/README.md
```diff
@@ -494,56 +273,7 @@ kubectl create secret generic gcs-credentials --from-file=/path/to/gcloud/creden
...
-        - containerPort: 9000
-        # Mount the volume into the pod
-        volumeMounts:
-        - name: gcs-credentials
-          mountPath: "/etc/credentials"
-          readOnly: true
-```
-
-Create the Deployment
+GCS project ID. Create the Deployment like so
```
kannappanr: `Create the Deployment like so` needs to be corrected.

docs/orchestration/kubernetes/README.md
nitisht: > doesn't sound correct.

I have changed it, but like so is completely valid english. (ref: https://english.stackexchange.com/questions/4524/is-like-so-correct, https://www.urbandictionary.com/define.php?term=Like%20So)

docs/orchestration/kubernetes/README.md
nitisht: > You can achieve this using Services. can be changed to This can be achieved using services.

`This can be achieved using services.` is passive voice. It is generally avoided in documentation
https://writing.stackexchange.com/questions/16812/technical-documentation-voice-preference-active-voice-passive-voice-a . I think we can keep it as is.

kannappanr:
@nitisht Sorry, didn't know about `like so`.

kannappanr (approved):
LGTM

nitisht:
> Sorry, didn't know about like so.

No worries at all, I just pointed out to clarify. I understand there is no one correct answer here, but more about what is more readable for majority, and your approach seems more popular.

harshavardhana (approved):

PR #6925, opened 2018-12-05
Provide configurable port for metrics scraping

bpmoneill:
## Description
The metrics port can now be set via MINIO_EXT_PORT environment
variable, exposing metrics on path '/metrics'.  Where MINIO_EXT_PORT
has not been set, the default path of '/minio/prometheus/metrics' will
apply.

## Motivation and Context
Prometheus uses /metrics as the default path to scrape for metrics.
For multi-container Kubernetes PODs, this allows the path to be
consistent across all containers within the POD allowing for
auto-discovery of metrics.

## Regression
Is this PR fixing a regression? No

## How Has This Been Tested?
Local build of Minio Server and Gateway and verified with Prometheus.

### values.yaml
```
- job_name: minio_metrics_9002
    scrape_interval: 10s
    metrics_path: /metrics
    static_configs:
    - targets:
      - 127.0.0.1:9002

  - job_name: minio_metrics_9000
    scrape_interval: 10s
    metrics_path: /minio/prometheus/metrics
    static_configs:
    - targets:
      - 127.0.0.1:9000
```
### Start Prometheus
```
./prometheus --config.file=values.yaml
```

## Server Behaviour
Running Minio Server on 127.0.0.1:9000 and also with a specific ip-address.

#### Start Server
```
export MINIO_EXT_PORT=<test-value>
./minio server /tmp

./minio server --address 192.168.0.13:9000 /tmp
```

| MINIO_EXT_PORT | Path                | Result |
|:---------------|:------------------- |:-------|
| 9000           | /metrics            | Fatal error, Invalid MINIO_EXT_PORT value (:9000): port is already in use. |
| 90000000       | /metrics            | Fatal error, Error: listen tcp4: address 900000000: invalid port |
| -1             | /metrics            | Fatal error, Error: listen tcp4: address -1: invalid port |
| Unset          | /metrics            | metrics available on /minio/prometheus/metrics port 9000 only, Prometheus displays 1 target |
| 9002           | /metrics            | metrics available on /metrics port 9002 and /minio/prometheus/metrics port 9000, Prometheus displays 2 targets |
| 9002           | /<any_other_path>   | 501 Not implemented, A header you provided implies functionality that is not implemented |

## Gateway Behaviour
Running Minio S3 Gateway and Server on 127.0.0.1:9000 and 127.0.0.1:9001 respectively and also with a specific ip-address.

#### Start Gateway
```
export MINIO_EXT_PORT=<test-value>
./minio gateway s3 http://127.0.0.1:9001

./minio gateway --address 192.168.0.13:9000 s3 http://127.0.0.1:9001

```

| MINIO_EXT_PORT | Path                | Result |
|:---------------|:------------------- |:-------|
| 9000           | /metrics            | Fatal error, Invalid MINIO_EXT_PORT value (:9000): port is already in use. |
| 9001           | /metrics            | Fatal error, Invalid MINIO_EXT_PORT value (:9001), port points to the local backend storage: Invalid Extensions Port value. |
| 90000000       | /metrics            | Fatal error, Error: listen tcp4: address 900000000: invalid port |
| -1             | /metrics            | Fatal error, Error: listen tcp4: address -1: invalid port |
| unset          | /metrics            | metrics available on /minio/prometheus/metrics port 9000 only, Prometheus displays 1 target |
| 9002           | /metrics            | metrics available on /metrics port 9002 and /minio/prometheus/metrics port 9000, Prometheus displays 2 targets |
| 9002           | /<any_other_path>   | 501 Not implemented, A header you provided implies functionality that is not implemented |

## Types of changes

## Checklist:
- [x] Go Test passed.
- [x] Make Verifiers passed.

### DIFF
--- a/cmd/common-main.go
+++ b/cmd/common-main.go
@@ -355,4 +355,11 @@ func handleCommonEnvVars() {
 			globalCompressMimeTypes = contenttypes
 		}
 	}
+
+	extensionsPort := os.Getenv("MINIO_EXT_PORT")
+	if extensionsPort == globalMinioPort {
+		errorMsg := "Invalid MINIO_EXT_PORT value (:" + extensionsPort + ")"
+		logger.FatalIf(errors.New("port is already in use"), errorMsg)
+	}
+	globalExtensionsPort = ":" + extensionsPort
 }
--- a/cmd/extension-port-handlers.go
+++ b/cmd/extension-port-handlers.go
@@ -1,0 +1,65 @@
+package cmd
+
+import (
+	"net/http"
+)
+
+const (
+	extPortReservedMetrics     = "metrics"
+	extPortReservedMetricsPath = "/" + extPortReservedMetrics
+)
+
+//
+// Metrics Routing
+//
+func isMetricsOnExtensionPort(path string, host string) bool {
+	_, requestPort := mustSplitHostPort(host)
+	requestPort = ":" + requestPort
+	return path == extPortReservedMetricsPath && globalExtensionsPort != "" && requestPort == globalExtensionsPort
+}
+
+func isSupportedPath(req *http.Request) bool {
+	if req == nil {
+		return false
+	}
+	aType := getRequestAuthType(req)
+	return aType == authTypeAnonymous &&
+		isMetricsOnExtensionPort(req.URL.Path, req.Host)
+}
+
+type extPortReservedPathHandler struct {
+	handler http.Handler
+}
+
+func setExtPortReservedPathHandler(h http.Handler) http.Handler {
+	return extPortReservedPathHandler{h}
+}
+
+func (h extPortReservedPathHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if isSupportedPath(r) {
+		h.handler.ServeHTTP(w, r)
+	} else {
+		writeErrorResponse(w, ErrNotImplemented, r.URL, guessIsBrowserReq(r))
+		return
+	}
+}
+
+//
+// Cache Control
+//
+type extPortCacheControlHandler struct {
+	handler http.Handler
+}
+
+func setExtPortBrowserCacheControlHandler(h http.Handler) http.Handler {
+	return extPortCacheControlHandler{h}
+}
+
+func (h extPortCacheControlHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if r.Method == http.MethodGet && guessIsBrowserReq(r) && globalIsBrowserEnabled {
+		if hasPrefix(r.URL.Path, extPortReservedMetricsPath) {
+			w.Header().Set("Cache-Control", "no-store")
+		}
+	}
+	h.handler.ServeHTTP(w, r)
+}
--- a/cmd/extensions-port-server.go
+++ b/cmd/extensions-port-server.go
@@ -1,0 +1,40 @@
+package cmd
+
+import (
+	"github.com/gorilla/mux"
+	xhttp "github.com/minio/minio/cmd/http"
+	"github.com/minio/minio/cmd/logger"
+	"github.com/minio/minio/pkg/certs"
+)
+
+func initLogger() {
+	logger.Init(GOPATH, GOROOT)
+	logger.RegisterUIError(fmtError)
+}
+
+var extPortHandlers = []HandlerFunc{
+	setExtPortReservedPathHandler,
+	setExtPortBrowserCacheControlHandler,
+	// Add new extension port specific handlers here.
+}
+
+func startExtensionsPortServer(getCert certs.GetCertificateFunc) {
+	initLogger()
+	err := checkPortAvailability(globalExtensionsPort)
+	if err != nil {
+		logger.Fatal(err, "Unable to start the extensions port server")
+	}
+
+	router := mux.NewRouter().SkipClean(true)
+
+	// register all extensions port applicable routers
+	registerMetricsRouter(router, extPortReservedMetricsPath)
+	handler := registerHandlers(registerHandlers(router, globalHandlers...), extPortHandlers...)
+
+	extensionsPortServer := xhttp.NewServer([]string{globalMinioHost + globalExtensionsPort}, criticalErrorHandler{handler}, getCert)
+	extensionsPortServer.UpdateBytesReadFunc = globalConnStats.incInputBytes
+	extensionsPortServer.UpdateBytesWrittenFunc = globalConnStats.incOutputBytes
+	go func() {
+		globalHTTPServerErrorCh <- extensionsPortServer.Start()
+	}()
+}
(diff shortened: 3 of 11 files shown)


harshavardhana:
Moving this to blocked for further discussion

wlan0:
@bpmoneill 

There are multiple implications for such a change:

1. the first path element overridden (`/metrics`) is reserved for bucket names. This change would prevent buckets by the name `metrics` from being created
2. This would block upgrades of existing instances with buckets named `metrics`
3. This would require extensive changes to naming and API conventions 
4. Breaking API conventions come with their own set of overheads and technical debt

The benefit of this change is that it would allow metrics to be scraped without the user having to specify the `prometheus.io/path` annotation. Additionally, In case of multiple containers in a pod, the user would not have to create multiple services or configure relabel rules.

However, this benefit would be diminished if there is no common denominator of metrics paths among the containers in the given pod. 

As you can see, this change adds significant overhead, technical debt and maintenance for little value. 

There are alternate solutions to this problem

1. Create a service for each port that needs monitoring and set `prometheus.io/*` annotations on the service. Brian Brazil, the author of prometheus suggests a solution along these lines as noted here: https://github.com/prometheus/prometheus/issues/1864#issuecomment-236875520

2. Brian suggest the option of using multiple ports corresponding to containers inside the same annotation. This can be extended to `__metrics_path__` as well. However, the relabel_configs for this can get pretty involved.
 https://github.com/prometheus/prometheus/issues/3756#issuecomment-361182179

The right approach for this problem is to enable per container labels in Kubernetes. This has been discussed before here: https://github.com/kubernetes/kubernetes/issues/26275

bpmoneill:
@wlan0, 

> _1. the first path element overridden (/metrics) is reserved for bucket names. This change would prevent buckets by the name metrics from being created_

I do not see that this is the case.  '/metrics' can still be used for bucket names on the server/gateway port, just not on the user defined extensions port.  The implementation separates the server/gateway routing from extension port requests, unless i have missed something.

> _2. This would block all upgrades of existing instances with buckets named metrics._

As stated for the reasons above, i do not believe this is the case.

Given the above, does this also negate points 3 and 4?  If i have missed something in the implementation or understanding please let me know.

> _However, even this benefit would be diminished if there are no common denominator of metrics paths among the containers in the given pod._

Agreed, this resolves a single use case.  We could allow the user to provide the path in addition to the port?  This would further support the idea of a user defined extensions port.  The convention is that certain features, in this case metrics, can be configured by the user for port and path.  Other features could also be added to this port where required.

Thanks for the alternative suggestions.  I do not believe option 1 is valid for us as we do not want to expose all ports as services. I am not sure about option 2, i will need to check if this will work for us.

bpmoneill:
@wlan0, looking into option 2 and i do not see how this helps us.  This solution is to prevent all containers within a pod being scraped for metrics by creating appropriate port label names, but this does assume a single scrape path i.e. prometheus.io.path is defined at the pod level and is the same for all containers within that pod.  This approach does not, from what i can see, solve the following use case:

- a single pod with multiple containers where Minio exposes metrics on '/minio/prometheus/metrics' and all other containers exposing on '/metrics'.

I do not see an approach which will allow us to configure multiple prometheus.io.path definitions for multiple containers within the same pod.

wlan0:
> Agreed, this resolves a single use case. 

Adding platform level features to solve specific use cases would leave the platform unmaintainable. Minio has existing conventions and philosophy for evolving its architecture. This change is architectural, and overrides the existing convention of using a single port. This change cannot be taken lightly, since it affects system attributes, flexibility, fault-tolerance, security, usability and maintainability. 

> We could allow the user to provide the path in addition to the port? This would further support the idea of a user defined extensions port. 

Adding a new port requires much consideration, no less adding new path conventions

1. Existing users would lose the ability to monitor their Minio instances on port `9000`
2. Existing users cannot upgrade without going through the process of opening up the firewall to another port
3. It increases the attack surface of Minio servers
4. It adds management overhead for users 

> Given the above, does this also negate points 3 and 4? If i have missed something in the implementation or understanding please let me know.

API conventions will be broken since the existing `/metrics` endpoint will be moved to the new port.  Such a design decision will ensue in a fallout from lack of uniformity, and broken conventions 

> The convention is that certain features, in this case metrics, can be configured by the user for port and path. Other features could also be added to this port where required.

This introduces the concept of a different port for metrics, and the stage for its evolution has not been set. Will this new port simply be a `metrics` port, or will it be a`control` or `management` port? Will all of the existing `control` features be moved to this port? What do you mean by "certain" features?

>  I do not believe option 1 is valid for us as we do not want to expose all ports as services.

Can you explain further what you mean by "expose all ports as services"? Adding a service does not expose a port. Every port is already exposed and accessible. 

Technically your solution "works", but it is not the right solution. The best course of action is to solve this problem at the configuration layer, not the application layer, for all the reasons laid out above.

In response to your newer comment:

> looking into option 2 and i do not see how this helps us. This solution is to prevent all containers within a pod being scraped for metrics by creating appropriate port label names, but this does assume a single scrape path i.e. prometheus.io.path is defined at the pod level and is the same for all containers within that pod. This approach does not, from what i can see, solve the following use case:

As originally mentioned, relabel_config of a similar kind to what has been done with `__port__` needs to be written for  `__metrics_path__`. Please note the use of `relabel_configs` as the solution.

> Brian suggest the option of using multiple ports corresponding to containers inside the same annotation. This can be extended to __metrics_path__ as well. However, the relabel_configs for this can get pretty involved.

bpmoneill:
> Can you explain further what you mean by "expose all ports as services"? Adding a service does not expose a port. Every port is already exposed and accessible.

Apologies, poor terminology.  What i mean is that logically at least we cannot simply create services for all ports we wish to scrape, this is why we have multi-container pods i.e. sidecar pattern, for helper applications.

> 1. Existing users would lose the ability to monitor their Minio instances on port 9000
> 2. Existing users cannot upgrade without going through the process of opening up the firewall to another port
> 3. It increases the attack surface of Minio servers
> 4. It adds management overhead for users

I understand your concerns but it should be noted that, although possibly not the right solution, this implementation does not change existing behaviour on server/gateway with regards to buckets , metrics, firewall configuration. etc.  Metrics will always be available on '/minio/prometheus/metrics' even if the user defines the extension port.  And for requests made to the server/gateway on port 9000, the bucket path /metrics is also still available, users will not lose their ability to monitor Minio instances on port 9000, the extensions port is additional to existing server/gateway port, not overriding.  If users do not define the port, they will see no difference in behaviour or functionality.  Additionally, if the user does define the port then it is an informed choice and as such they will be aware and accepting of firewall changes and additional management, this is their choice.  

With that said i understand that the architectural choice here may not be correct and certainly requires further discussion to ensure this fits with your plans and is something which provides value to all and is maintainable and extendable.  

It did not look like relabel_config could solve our issue but i am possibly wrong.  Is the intention to write this such that we can pick up on multiple metrics paths using a regex pattern?

harshavardhana:
Since we decided to not do this, we will close this PR for now.

PR #6945, opened 2018-12-10
cmd: remove duplicated switch case

quasilyte:
The `errFileAccessDenied` case was repeated twice.

### DIFF
--- a/cmd/object-api-errors.go
+++ b/cmd/object-api-errors.go
@@ -47,7 +47,7 @@ func toObjectErr(err error, params ...string) error {
 				Object: params[1],
 			}
 		}
-	case errIsNotRegular, errFileAccessDenied:
+	case errIsNotRegular,:
 		if len(params) >= 2 {
 			err = ObjectExistsAsDirectory{
 				Bucket: params[0],


harshavardhana:
Thanks for the PR. @Quasilyte 

So quick question why was the pull request template removed ?. We have no idea what this PR is doing and what is the gist behind the fix. Please resubmit with the template used appropriately.

harshavardhana:
Going to close this, for now, feel free to re-open with a proper PR.

quasilyte:
You have a switch statement with duplicated cases.
It means that either of those 2 cases won't ever execute.
And I found that by using static analysis tool.
This PR fixes a logical bug in the code. You can ignore it, of course, your choice.

quasilyte:
I were editing the file via github web editor and haven't removed the comma.

harshavardhana:
> You have a switch statement with duplicated cases.
> It means that either of those 2 cases won't ever execute.
> And I found that by using static analysis tool.
> This PR fixes a logical bug in the code. You can ignore it, of course, your choice.

Always happy to take contributions but use the template and explain a bit more detail on what the PR does.

quasilyte:
I can't give more context, but I'll try to open an issue.
You're free to fix it if you feel like it.

quasilyte:
I tried to explain it in #6948.
Does it still requires more explanations?
I really lost here and I can't explain it in another way, basically, I give up, but to leave something useful, contributors have an issue and they can take a look, so this issue has a chance to be fixed. :)
Thanks.

harshavardhana:
> I tried to explain it in #6948.
> Does it still requires more explanations?
> I really lost here and I can't explain it in another way, basically, I give up, but to leave something useful, contributors have an issue and they can take a look, so this issue has a chance to be fixed. :)
> Thanks.

Thanks for the issue @Quasilyte we will fix it.

PR #6878, opened 2018-11-27
Add transaction lock when migrating configs

harshavardhana:
## Description
Add transaction lock when migrating configs

## Motivation and Context
When migrating configs it happens often that some
servers fail to start due to version mismatch etc.

Hold a transaction lock such that all servers get
serialized.

## Regression
Not sure

## How Has This Been Tested?
Just test this PR with or without this change. Without this change, some servers will fail to start from the following script.

```sh
#!/bin/bash

export MINIO_ERASURE_SET_DRIVE_COUNT=6
export MINIO_ACCESS_KEY="minio"
export MINIO_SECRET_KEY="minio123"

rm -rf /tmp/export{01..24}

for i in {01..24}; do
    minio server --address ":90${i}" http://127.0.0.1:9001/tmp/export01 http://127.0.0.1:9002/tmp/export02 http://127.0.0.1:9003/tmp/export03 http://127.0.0.1:9004/tmp/export04 http://127.0.0.1:9005/tmp/export05 http://127.0.0.1:9006/tmp/export06 http://127.0.0.1:9007/tmp/export07 http://127.0.0.1:9008/tmp/export08 http://127.0.0.1:9009/tmp/export09 http://127.0.0.1:9010/tmp/export10 http://127.0.0.1:9011/tmp/export11 http://127.0.0.1:9012/tmp/export12 http://127.0.0.1:9013/tmp/export13 http://127.0.0.1:9014/tmp/export14 http://127.0.0.1:9015/tmp/export15 http://127.0.0.1:9016/tmp/export16 http://127.0.0.1:9017/tmp/export17 http://127.0.0.1:9018/tmp/export18 http://127.0.0.1:9019/tmp/export19 http://127.0.0.1:9020/tmp/export20 http://127.0.0.1:9021/tmp/export21 http://127.0.0.1:9022/tmp/export22 http://127.0.0.1:9023/tmp/export23 http://127.0.0.1:9024/tmp/export24 &
done
```

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added unit tests to cover my changes.
- [ ] I have added/updated functional tests in [mint](https://github.com/minio/mint). (If yes, add `mint` PR # here: )
- [x] All new and existing tests passed.

PR #7082, opened 2019-01-15
Avoid holding write lock on config in situations where it is not needed

krishnasrinivas:
On 4 disk distributed setup (4 servers, each with 1 disk) when 2 servers are started, minio does not give a read-only service. It instead tries to hold a write lock on config.json.transaction and hangs (as N/2+1 servers are not available) This hang can be avoided as there is no need to hold write lock if there is no need for config.json migration.

This is to allow the cluster to come up when N/2 number of disks is available.

Yes PR #6878

With the setup detailed in description and tested to make sure that the problem is not seen.


harshavardhana (approved):
LGTM tested

krisis (approved):

PR #7214, opened 2019-02-08
remove unnecessary buffer while discarding stream

wlan0:
Makes no sense to use a buffer while draining http body. We are choking a fast operation by putting a buffer in between. `ioutil.Discard` is not going to push back on the reader, if anything, it'll be the buffer that pushes back.

### DIFF
--- a/cmd/http/close.go
+++ b/cmd/http/close.go
@@ -44,9 +44,7 @@ func DrainBody(respBody io.ReadCloser) {
 		// Without this closing connection would disallow re-using
 		// the same connection for future uses.
 		//  - http://stackoverflow.com/a/17961593/4465767
-		bufp := b512pool.Get().(*[]byte)
-		defer b512pool.Put(bufp)
-		io.CopyBuffer(ioutil.Discard, respBody, *bufp)
-		respBody.Close()
+		defer respBody.Close()
+		io.Copy(ioutil.Discard, respBody)
 	}
 }


minio-ops:
## Mint Automation
| Test | Result |
|:--- |:--- |
| mint-compression-xl.sh | :heavy_check_mark: |
| mint-compression-dist-xl.sh | :heavy_check_mark: |
| mint-compression-fs.sh | :heavy_check_mark: |

cmd/http/close.go
```diff
@@ -44,9 +44,7 @@ func DrainBody(respBody io.ReadCloser) {
 		// Without this closing connection would disallow re-using
 		// the same connection for future uses.
 		//  - http://stackoverflow.com/a/17961593/4465767
-		bufp := b512pool.Get().(*[]byte)
```
harshavardhana, line 47: This implementation was because we frequently drain the body due to the extensive usage of the HTTP client. So this code solves two things.

- One is to avoid using internal buffers used by io.Copy() repeatedly.
- Another is to use a smaller buffer 512 is enough to drain the buffer because practically we would have read the entire body anyways. But we need to do this to satisfy Go connection pooling requirements. So we don't need a 4k buffer we can simply use a 512 buffer.

Changing this to io.Copy is simpler but it adds lots of internal allocations..

wlan0: This is still inefficient. It doesn't make sense to use a fixed size buffer from a pool, when most of the time this call is preceded by a `ioutil.ReadAll()`. We are going through a atomic acquisition, and potentially adding the cost of allocating a 512 byte buffer for each close. 

Add to that, the close is called on an empty stream in most cases - and yet acquiring a buffer. 

The detail most people on stackoverflow and "internet" miss is that io.Copy is just fine because resp.Body is a *io.LimitedReader, and Copy does not allocate a 32K buffer in this case. In the most likely scenario of this being called on an already closed stream, io.Copy will not even allocate the 512 bytes.

The cost of acquiring a buffer on a previously read stream, along with the cost of allocating a new buffer even when it is not needed is a performance penalty, not an improvement.

minio-ops:
## Mint Automation
| Test | Result |
|:--- |:--- |
| mint-compression-xl.sh | :heavy_check_mark: |
| mint-compression-dist-xl.sh | :heavy_check_mark: |
| mint-compression-fs.sh | :heavy_check_mark: |

cmd/http/close.go
harshavardhana: As discussed since io.Copy() does magic inside when its an io.LimitReader() we can use io.Copy() here and it will be simpler.

harshavardhana (approved):

donatello (approved):
Well though we can take this change in, it is not for the reasons discussed above as that's not what happens. Discard actually implements ReadFrom and this would be used by io.Copy - internally it allocates from a sync.Pool an 8kb buffer, which can't really be avoided.

Also it's not clear resp.Body would be a limited reader especially in case of chunked transfer encoding.

harshavardhana:
> Well though we can take this change in, it is not for the reasons discussed above as that's not what happens. Discard actually implements ReadFrom and this would be used by io.Copy - internally it allocates from a sync.Pool an 8kb buffer, which can't really be avoided.
> 
> Also it's not clear resp.Body would be a limited reader especially in case of chunked transfer encoding.

Yes agreed it's not true in all cases wherever we are using io.Copy is fine for now. io.Copy itself has hacks to use io.LimitReader struct private members, but since it is the most used we are okay with this for now.

PR #7475, opened 2019-04-03
Set S3 Select record message length to 128KiB

donatello:
- Previously this limit was a little more than 1MiB, and it broke
  compatibility with AWS SDK Java causing a buffer overflow error.

## Regression

No

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added unit tests to cover my changes.
- [ ] I have added/updated functional tests in [mint](https://github.com/minio/mint). (If yes, add `mint` PR # here: )
- [x] All new and existing tests passed.

PR #8312, opened 2019-09-25
S3 Select: Workaround java buffer size

klauspost:
## Description

Updates #7475

The Java implementation has a 128KB buffer and a message must be emitted before that is used. #7475 therefore limits the message size to 128KB. But up to 256 bytes are written to the buffer in each call. This means we must emit a message before shorter than 128KB.

Therefore we change the limit to 128KB minus 256 bytes.

## How to test this PR?

Triggering the bug in the Java code could also seem to be dependent on network conditions.

## Types of changes

### DIFF
--- a/pkg/s3select/message.go
+++ b/pkg/s3select/message.go
@@ -62,7 +62,12 @@ var recordsHeader = []byte{
 }
 
 const (
-	maxRecordMessageLength = 128 * 1024 // Chosen for compatibility with AWS JAVA SDK
+	// Chosen for compatibility with AWS JAVA SDK
+	// It has a a buffer size of 128K:
+	// https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/internal/eventstreaming/MessageDecoder.java#L26
+	// but we must make sure there is always space to add 256 bytes:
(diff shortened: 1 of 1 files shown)


pkg/s3select/message.go
```diff
@@ -62,7 +62,12 @@ var recordsHeader = []byte{
 }
 
 const (
-	maxRecordMessageLength = 128 * 1024 // Chosen for compatibility with AWS JAVA SDK
+	// Chosen for compatibility with AWS JAVA SDK
+	// It has a a buffer size of 128K:
+	// https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/internal/eventstreaming/MessageDecoder.java#L26
+	// but we must make sure there is always space to add 256 bytes:
+	// https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/model/SelectObjectContentEventStream.java#L197
+	maxRecordMessageLength = (128 << 10) - 256
```
donatello, line 70: Quite a quirk - now we have a buffer that is 127.25 KiB long!

donatello (approved):
LGTM

kannappanr (approved):
LGTM

PR #8457, opened 2019-10-28
S3 Select: Support Square Bracket Lists

### LINKED ISSUE #8422: S3 SELECT - syntax difference between AWS & MinIO
We're using the S3 SELECT to fetch data from CSV files across both AWS and MinIO, with outset in a frontend platform based on the javascript SDK.

We run a standard S3 SELECT query, illustratively as below:

`SELECT ParameterX FROM S3Object WHERE ParameterY in ['X', 'Y']`

This works when fetching from an AWS S3 backend - but when fetching from a MinIO backend, we need to replace the above with the below alternative notation.

`SELECT ParameterX FROM S3Object WHERE ParameterY in ('X', 'Y')`

Based on a reply from Harshavardhana on the Slack channel, this is likely unintentional. It would be great to get this resolved so that we can use consistent syntax.

Martin

klauspost:
## Description

Allows for S3 compatible `SELECT * from s3object s WHERE id IN [3,2]`

## Motivation and Context

Fixes #8422

## How to test this PR?

Tests included

## Types of changes

### DIFF
--- a/pkg/s3select/select_test.go
+++ b/pkg/s3select/select_test.go
@@ -173,6 +173,22 @@ func TestJSONQueries(t *testing.T) {
 {"id":2,"title":"Second Record","desc":"another text","numbers":[2,3,4]}`,
 		},
 		{
+			name:       "indexed-list-square-bracket",
+			query:      `SELECT * from s3object s WHERE [7,8.5,9] = s.nested[1]`,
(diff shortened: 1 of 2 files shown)


sinhaashish (approved):
Tested & LGTM

nitisht (approved):

PR #8518, opened 2019-11-13
Add hidden profiling parameters

klauspost:
## Description

Add profiling parameters. We keep them hidden so they don't clutter the help ui.

Parameters and syntax are similar to the ones in `mc`.

## Motivation and Context

This will enable the possibility to capture various profiles both when testing or when identifying bottlenecks on remote systems.

## How to test this PR?

`minio server -cpuprofile=. disk1`

## Types of changes

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -25,6 +25,7 @@ import (
 	"github.com/minio/mc/pkg/console"
 	"github.com/minio/minio/pkg/trie"
 	"github.com/minio/minio/pkg/words"
+	"github.com/pkg/profile"
 )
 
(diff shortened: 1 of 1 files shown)


harshavardhana:
you need to check with @abperiasamy about this - we don't add new CLI flags without UI approvals.  What is the current limitation of using `mc admin profile` @klauspost ?

klauspost:
So much stuff in there. I looked for references to pprof in the codebase but didn't find any.

This would make a great addition to warp. I'll see if I can add it easily. That way it could just alongside the actual benchmark and only while the benchmark is running.

Closing for now.

PR #8650, opened 2019-12-14
Remove safe mode for invalid entries in config

harshavardhana:
## Description
Remove safe mode for invalid entries in config

## Motivation and Context
The approach is that now safe mode is only invoked when
we cannot read the config or under some catastrophic
situations, but not under situations when config entries
are invalid or unreachable. This allows for maximum
availability for MinIO and not fail on our users unlike
most of our historical releases.

## How to test this PR?
Set valid configs such enable vault endpoint and then 
take the vault endpoint down restart the server. This 
shouldn't take the server into safe-mode - since this
can be a quite common occurrence. Allow server to
continue regardless of the failure and let the internal
sub-system code retry.

## Types of changes

## Checklist:
- [ ] Documentation needed
- [ ] Unit tests needed

### DIFF
--- a/cmd/config/config.go
+++ b/cmd/config/config.go
@@ -32,23 +32,13 @@ import (
 
 // Error config error type
 type Error struct {
-	Kind ErrorKind
-	Err  string
+	Err string
 }
-
-// ErrorKind config error kind
-type ErrorKind int8
-
-// Various error kinds.
-const (
-	ContinueKind ErrorKind = iota + 1
-	SafeModeKind
-)
 
 // Errorf - formats according to a format specifier and returns
 // the string as a value that satisfies error of type config.Error
-func Errorf(errKind ErrorKind, format string, a ...interface{}) error {
-	return Error{Kind: errKind, Err: fmt.Sprintf(format, a...)}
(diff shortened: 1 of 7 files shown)


cmd/config/config.go
```diff
@@ -366,9 +356,9 @@ func LookupCreds(kv KVS) (auth.Credentials, error) {
 	if err := CheckValidKeys(CredentialsSubSys, kv, DefaultCredentialKVS); err != nil {
 		return auth.Credentials{}, err
 	}
-	accessKey := env.Get(EnvAccessKey, kv.Get(AccessKey))
-	secretKey := env.Get(EnvSecretKey, kv.Get(SecretKey))
-	if accessKey == "" && secretKey == "" {
+	accessKey := kv.Get(AccessKey)
```
kannappanr, line 359: Will `kv` load it from the environment variable/config file?

harshavardhana: Yes it does, this code path is only for situations where env is not available

kannappanr (approved):
LGTM

poornas (approved):
LGTM

PR #8652, opened 2019-12-16
Add s390x support for travis.

srajmane:
As Travis CI [officially supports](https://blog.travis-ci.com/2019-11-12-multi-cpu-architecture-ibm-power-ibm-z) s390x builds, adding support for same.


harshavardhana:
We would like to reduce our CI runs, s390x is not what we are interested at this moment. Thanks for the PR but we would not take it for now.

PR #9315, opened 2020-04-10
Allow skipping the credential validation for S3 gateway

jaym:
## Description
Add environment variable to skip credential check for S3 gateway

## Motivation and Context
The credential validation probes a random bucket, which seems to cause
problems for some people using proxies that limit which buckets are
accessible.

## How to test this PR?
The following example will fail to start the gateway because the credentials are invalid:
```
# export MINIO_ACCESS_KEY=fake_aws_s3_access_key
# export MINIO_SECRET_KEY=fake_aws_s3_secret_key
# go run ./ gateway s3
```

The following example will allow the gateway to start:
```
# export MINIO_ACCESS_KEY=fake_aws_s3_access_key
# export MINIO_SECRET_KEY=fake_aws_s3_secret_key
# export SKIP_CREDENTIAL_CHECK=true
# go run ./ gateway s3
```

## Types of changes

## Checklist:
- [ ] Documentation needed
- [ ] Unit tests needed

### DIFF
--- a/cmd/gateway/s3/gateway-s3.go
+++ b/cmd/gateway/s3/gateway-s3.go
@@ -33,13 +33,17 @@ import (
 
 	"github.com/minio/minio-go/v6/pkg/encrypt"
 	"github.com/minio/minio-go/v6/pkg/s3utils"
+	"github.com/minio/minio/cmd/config"
 	"github.com/minio/minio/cmd/logger"
 	"github.com/minio/minio/pkg/auth"
 	"github.com/minio/minio/pkg/bucket/policy"
+	"github.com/minio/minio/pkg/env"
 )
 
 const (
 	s3Backend = "s3"
+
+	envS3SkipCredentialCheck = "SKIP_CREDENTIAL_CHECK"
 )
 
 func init() {
@@ -54,6 +58,9 @@ FLAGS:
   {{end}}{{end}}
 ENDPOINT:
   s3 server endpoint. Default ENDPOINT is https://s3.amazonaws.com
+
+SKIP_CREDENTIAL_CHECK:
+	Skips checking for valid credentials if set to true
 
 EXAMPLES:
   1. Start minio gateway server for AWS S3 backend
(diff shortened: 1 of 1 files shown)


harshavardhana:
What seems to be the problem that it causes using proxies ?

jaym:
We've had some people with internal security policies that only allow access to a whitelisted set of buckets. The way credentials are validated currently in the minio gateway, they probe a random bucket which would not be whitelisted.

harshavardhana:
> We've had some people with internal security policies that only allow access to a whitelisted set of buckets. The way credentials are validated currently in the minio gateway, they probe a random bucket which would not be whitelisted.

Then we should have a code that should automatically proceed, we don't need this type of environment variable. Can you should me what is the error that you see?

jaym:
I'll try to track down the code, but it might be specific to the persons environment (could be that they just 500 all those requests). It might be a few days to get this info back.

Another option instead of disabling it would be if we could specify which bucket the credentials are validated against

harshavardhana:
> I'll try to track down the code, but it might be specific to the persons environment (could be that they just 500 all those requests). It might be a few days to get this info back.

@jaym 500 looks wrong, either it should be 403 or 400 - we should investigate why there is 5xx here.

jaym:
The 5xx wouldn't be coming from minio or S3, but from the proxy. It very well could be a 403. I've asked the person who was having problems and am waiting to hear back.

For more clarity though, the setup looks something like this:

```
minio +---> proxy +----> s3
```

When minio sends a bucket probe to the proxy, it provides a random bucket. The proxy has an internal security policy that says you may only use buckets that are whitelisted. So this request would never be sent to S3. Instead, it could do any number of things such as reset the connection, return a 5xx or 400/403. The proxy is the thing messing with our requests and we don't control what it does when we give it a bucket that isn't in the whitelist.

harshavardhana:
> When minio sends a bucket probe to the proxy, it provides a random bucket. The proxy has an internal security policy that says you may only use buckets that are whitelisted. So this request would never be sent to S3. Instead, it could do any number of things such as reset the connection, return a 5xx or 400/403. The proxy is the thing messing with our requests and we don't control what it does when we give it a bucket that isn't in the whitelist.

So if you send 403 then it will work, the current code handles this already.  If you do send 500 it means there is an error on the backend which is unhandled. 

So until this is clear, I don't think this PR is correct. We would like to limit the number of ENVs that get added as it is a burden for us to maintain and test in future.

jaym:
yea, cant be a 403, but could be one of the other things. I'll go ahead and close this for now and suggest that a 403 be returned for all the `probe-bucket-sign-*` requests through the proxy

PR #9602, opened 2020-05-14
tests: Fix one multi-delete test failure in Windows CI

vadmeste:
## Description
There is a disparency of behavior under Linux & Windows about
the returned error when trying to rename a non existant path.

err := os.Rename("/path/does/not/exist", "/tmp/copy")

Linux:
  isSysErrNotDir(err) = false
  os.IsNotExist(err) = true

Windows:
  isSysErrNotDir(err) = true
  os.IsNotExist(err) = true

ENOTDIR in Linux is returned when the destination path
of the rename call contains a file in one of the middle
segments of the path (e.g. /tmp/file/dst, where /tmp/file
is an actual file not a directory)

However, as shown above, Windows has more scenarios when
it returns ENOTDIR. For example, when the source path contains
an inexistant directory in its path.

In that case, we want errFileNotFound returned and not
errFileAccessDenied, so this commit will add a further check to close
the disparency between Windows & Linux.

## Motivation and Context
Fix one test failure under Windows

## How to test this PR?
go test under Windows

## Types of changes

## Checklist:
- [ ] Documentation needed
- [ ] Unit tests needed

### DIFF
--- a/cmd/os-reliable.go
+++ b/cmd/os-reliable.go
@@ -131,7 +131,11 @@ func renameAll(srcFilePath, dstFilePath string) (err error) {
 
 	if err = reliableRename(srcFilePath, dstFilePath); err != nil {
 		switch {
-		case isSysErrNotDir(err):
+		case isSysErrNotDir(err) && !os.IsNotExist(err):
+			// Windows can have both isSysErrNotDir(err) and os.IsNotExist(err) returning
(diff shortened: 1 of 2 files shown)


klauspost (approved):
Nice catch!

harshavardhana (approved):

PR #9637, opened 2020-05-18
fix: remove LDAP groups claim and store them on server

harshavardhana:
## Description
fix: remove LDAP groups claim and store them on server

## Motivation and Context
Groups information shall be now stored as part of the
credential data structure, this is a more idiomatic
way to support large LDAP groups.

Avoids the complication of setups where LDAP groups
can be in the range of 150+ which may lead to excess
HTTP header size > 8KiB, to reduce such an occurrence
we shall save the group information on the server as
part of the credential data structure.

Bonus change support multiple policies across user types

## How to test this PR?
No functionality should change

## Types of changes

## Checklist:
- [ ] Documentation needed
- [ ] Unit tests needed


harshavardhana:
@Alevsk @donatello @vadmeste can I get some reviews?

Alevsk:
> @Alevsk @donatello @vadmeste can I get some reviews?

sure harsha, reviewing it right now

Alevsk (approved):
LGTM

cmd/iam.go
```diff
@@ -162,16 +162,37 @@ func newGroupInfo(members []string) GroupInfo {
...
+			continue
+		}
+		policies = append(policies, policy)
+	}
+	return policies
 }
 
 func (mp MappedPolicy) policySet() set.StringSet {
-	return set.CreateStringSet(strings.Split(mp.Policy, ",")...)
+	var policies []string
```
vadmeste, line 183: Looks like easier this way:
` return set.CreateStringSet(mp.toSlice()...)`

vadmeste (commented):
One comment, LGTM, not tested yet

cmd/iam.go
harshavardhana: It is to avoid the empty value in map as a key @vadmeste which is a real thing btw.

donatello (approved):
LGTM

PR #9660, opened 2020-05-21
keep bucket metadata fields to be consistent

harshavardhana:
## Description
keep bucket metadata fields to be consistent

## Motivation and Context
added bonus reload bucket metadata always after
a successful MakeBucket, current we were only
doing it with object locking enabled.

## How to test this PR?
Nothing new, just a simple rename and functionality change

## Types of changes

## Checklist:
- [x] Fixes a regression probably with ListBuckets CreateTime value
- [ ] Documentation needed
- [ ] Unit tests needed

### DIFF
--- a/cmd/bucket-metadata.go
+++ b/cmd/bucket-metadata.go
@@ -46,7 +46,7 @@ const (
 )
 
 var (
-	defaultBucketObjectLockConfig = []byte(`<ObjectLockConfiguration xmlns="http://s3.amazonaws.com/doc/2006-03-01/"><ObjectLockEnabled>Enabled</ObjectLockEnabled></ObjectLockConfiguration>`)
+	enabledBucketObjectLockConfig = []byte(`<ObjectLockConfiguration xmlns="http://s3.amazonaws.com/doc/2006-03-01/"><ObjectLockEnabled>Enabled</ObjectLockEnabled></ObjectLockConfiguration>`)
 )
 
 //go:generate msgp -file $GOFILE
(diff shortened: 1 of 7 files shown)


harshavardhana:
This change is already done in the versioning PR, just moving it out here since it can go in here.

kannappanr (approved):
LGTM

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

cmd/bucket-metadata.go
```diff
@@ -46,7 +46,9 @@ const (
 )
 
 var (
-	defaultBucketObjectLockConfig = []byte(`<ObjectLockConfiguration xmlns="http://s3.amazonaws.com/doc/2006-03-01/"><ObjectLockEnabled>Enabled</ObjectLockEnabled></ObjectLockConfiguration>`)
+	enabledBucketObjectLockConfig   = []byte(`<ObjectLockConfiguration xmlns="http://s3.amazonaws.com/doc/2006-03-01/"><ObjectLockEnabled>Enabled</ObjectLockEnabled></ObjectLockConfiguration>`)
```
klauspost: In general I am not a big fan of including default values. 

They are taking up space in memory (and a bit on disk) for no real reason, whereas if `len(field) == 0 {/*default*/}` is easy to check AND doesn't require an unmarshal to get the actual values.

It also makes it clear whether it is a user applied default or a system default, which can be relevant if defaults should change.

harshavardhana: There is nothing we can do about it - S3 APIs some have default values already on the bucket. So this sort of designates that.

For example versioning config is already present just that the value is disabled by default.

harshavardhana: I can move it and keep it like allocating structs individually, I can do that too. Since this is minor memory usage I don't know if we should really worry about it.

klauspost: I know it needs to return it when requested, but unless the user has set it I don't see a big reason to actually store a value until they actually set something. 

I would expect it still would be possible to return a default config if none is set. I may misunderstand a state, but it seems that we add config if nothing has been set.

harshavardhana: Yeah with this I wanted to indicate what is the default state of all bucket metadata, rather than adding it in the handler layer.  Since it's small minor memory i thought this is the right place for it..

harshavardhana: I can change either way, just thought this is more readable.

klauspost (approved):

poornas (approved):

PR #9790, opened 2020-06-09
kes: try to auto. create master key if not present

aead:
## Description
This commit changes the data key generation such that
if a MinIO server/nodes tries to generate a new DEK
but the particular master key does not exist - then
MinIO asks KES to create a new master key and then
requests the DEK again.

From now on, a SSE-S3 master key must not be created
explicitly via: `kes key create <key-name>`.
Instead, it is sufficient to just set the env. var.
```
export MINIO_KMS_KES_KEY_NAME=<key-name>
```

However, the MinIO identity (mTLS client certificate)
must have the permission to access the `/v1/key/create/`
API. Therefore, KES policy for MinIO must look similar to:
```
[
  /v1/key/create/<key-name-pattern>
  /v1/key/generate/<key-name-pattern>
  /v1/key/decrypt/<key-name-pattern>
]
```
However, in our guides we already suggest that.
See e.g.: https://github.com/minio/kes/wiki/MinIO-Object-Storage#kes-server-setup

***

The ability to create master keys on request may also be
necessary / useful in case of SSE-KMS.

## Motivation and Context
KMS, SSE-S3, SSE-KMS

## How to test this PR?
Basically follow: https://github.com/minio/minio/tree/master/docs/kms#quick-start
**1**
```
curl -sSL --tlsv1.2 \
     -O 'https://raw.githubusercontent.com/minio/kes/master/root.key' \
     -O 'https://raw.githubusercontent.com/minio/kes/master/root.cert'
```
**2**
```
export MINIO_KMS_KES_ENDPOINT=https://play.min.io:7373
export MINIO_KMS_KES_KEY_FILE=root.key
export MINIO_KMS_KES_CERT_FILE=root.cert
export MINIO_KMS_KES_KEY_NAME=
```
> Choose an arbitrary key name.

**3**
```
export MINIO_ACCESS_KEY=minio
export MINIO_SECRET_KEY=minio123
export MINIO_KMS_AUTO_ENCRYPTION=on
minio server ~/export
```  

## Types of changes

## Checklist:
- [ ] Documentation needed
- [ ] Unit tests needed


minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

vadmeste:
@aead what happens if the master key is mistakenly replaced by some other values by the administrator ? I feel like the master key is something that the admin should store in his own safe place.

aead:
> what happens if the master key is mistakenly replaced by some other values by the administrator ?

Not sure whether I understand this correctly. The administrator cannot "replace" a master key at KES / the KMS key store. What might happen is that the admin changes the value of `MINIO_KMS_KES_KEY_NAME`. But that is totally fine and already supported. MinIO remembers what master key was used for a particular object. 

> I feel like the master key is something that the admin should store in his own safe place.

Not sure what you mean. The master key is only stored at the KMS key store.

vadmeste:
> Not sure what you mean. The master key is only stored at the KMS key store.

yes right, I've just found out that running KES server in FS mode is not recommended in production. 

Just wanted to protect users who do not exactly understand what they are doing from making mistakes, but I understand now that there is no way we can take any blame here :+1:

cmd/crypto/kes.go
```diff
@@ -235,48 +248,155 @@ type request struct {
...
+		}
+		var response Response
+		if err := json.NewDecoder(io.LimitReader(resp.Body, size)).Decode(&response); err != nil {
+			return err
+		}
+		return NewKESError(resp.StatusCode, response.Message)
+	}
+
+	var sb strings.Builder
+	if _, err := io.CopyN(&sb, resp.Body, size); err != nil {
```
vadmeste: If size == MaxBodySize, io.CopyN will most likely return the error io.EOF, maybe io.Copy() with limit reader is enough ?

vadmeste (commented):
One comment, LGTM otherwise & tested

cmd/crypto/kes.go
aead: But we want an error in this case, right. Otherwise we treat a "too large response" as valid?!

vadmeste: What I wanted to say is that  if the size is unknown, io.CopyN will always return io.EOF. 

If the size is unknown, it will be set to MaxBodySize, therefore io.CopyN will return an error because the body size will be less than MaxBodySize (in normal cases)

aead: Oh I see - yeah, that should be fixed. Will also send a fix to KES (this code comes from the KES SDK). :+1: 
However, it doubt it is an actual issue in the wild since KES has a proper JSON error API and sets the content type.

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

poornas (approved):
tested, LGTM

vadmeste (approved):
LGTM

PR #10184, opened 2020-08-03
feat: allow service accounts to be generated with OpenID STS

harshavardhana:
## Description
feat: allow service accounts to be generated with OpenID STS

## Motivation and Context
Bonus also fix a bug where we did not purge relevant
service accounts generated by rotating credentials
appropriately, service accounts should become invalid
as soon as its corresponding parent user becomes invalid.

Since service account themselves carry parent claim always
we would never reach this problem, as the access get
rejected at IAM policy layer.

## How to test this PR?
You need to test the service account creates based on the 
STS obtained from OpenID auth flow. 

## Types of changes

## Checklist:
- [ ] Documentation needed
- [ ] Unit tests needed

### DIFF
--- a/cmd/iam-object-store.go
+++ b/cmd/iam-object-store.go
@@ -304,6 +304,7 @@ func (iamOS *IAMObjectStore) loadUser(user string, userType IAMUserType, m map[s
 	if u.Credentials.AccessKey == "" {
 		u.Credentials.AccessKey = user
 	}
+
 	m[user] = u.Credentials
 	return nil
 }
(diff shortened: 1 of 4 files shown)


minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

Alevsk (approved):

cmd/iam-object-store.go
```diff
@@ -474,14 +475,30 @@ func (iamOS *IAMObjectStore) loadAll(ctx context.Context, sys *IAMSys) error {
...
 
 	// purge any expired entries which became expired now.
+	var expiredEntries []string
 	for k, v := range sys.iamUsersMap {
 		if v.IsExpired() {
 			delete(sys.iamUsersMap, k)
 			delete(sys.iamUserPolicyMap, k)
+			expiredEntries = append(expiredEntries, k)
+			if v.ParentUser != "" {
+				expiredEntries = append(expiredEntries, v.ParentUser)
```
vadmeste: I did not understand the logic here, why you assume that a parent user is expired if a sub-user is expired ?

vadmeste (commented):
One question

cmd/iam-object-store.go
harshavardhana: Because ParentUser is same as sub-user in OpenID.

harshavardhana: It's a good catch, I think I don't need to use ParentUser here..

harshavardhana:
PTAL @vadmeste

vadmeste (approved):
LGTM

PR #10506, opened 2020-09-17
Add a reasonable default if unable to get total RAM

klauspost:
## Description

Though unlikely we shouldn't skip initializing the API if we cannot get RAM.

Assume 16GiB as a default and log the error.

## Types of changes
- [x] Bug fix (defensive programming)

### DIFF
--- a/cmd/handler-api.go
+++ b/cmd/handler-api.go
@@ -22,6 +22,7 @@ import (
 	"time"
 
 	"github.com/minio/minio/cmd/config/api"
+	"github.com/minio/minio/cmd/logger"
 	"github.com/minio/minio/pkg/sys"
 )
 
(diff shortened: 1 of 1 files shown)


minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally
Error: No such image: minio/minio:10506-a2e4285

cmd/handler-api.go
```diff
@@ -45,7 +46,9 @@ func (t *apiConfig) init(cfg api.Config, setDriveCount int) {
 	if cfg.RequestsMax <= 0 {
 		stats, err := sys.GetStats()
 		if err != nil {
-			return
+			logger.LogIf(GlobalContext, err)
+			// Default to 16 GiB, not critical.
+			stats.TotalRAM = 16 << 30
```
harshavardhana, line 51: the reason for the return is because we don't apply any request limits if we can't get any RAM size whatsoever, I thought that this makes sense we don't know what the system is capable of. 

Setting a conservative 16GiB limit is also okay.

harshavardhana (changes requested):

cmd/handler-api.go
klauspost: It appears unintentional that the last part is skipped. At least this will make it more visible if there is a problem.

Not expecting this code to ever be executed, but just in case some OS is weird.

harshavardhana (approved):

PR #10624, opened 2020-10-02
Document requirements of user for AssumeRole

ad-m:
## Description

It took me a while to discover that a user creation was required. I also tried to create a Policy with the `AssumeRole` operation (which is required in AWS) but it was impossible and any user can `AssumeRole` in Minio.

## Motivation and Context

As above.

## How to test this PR?

Read and fix typo.

## Types of changes

## Checklist:

- [ ] Documentation needed
- [ ] Unit tests needed

### DIFF
--- a/docs/sts/assume-role.md
+++ b/docs/sts/assume-role.md
@@ -96,7 +96,7 @@ $ export MINIO_SECRET_KEY=minio123
 $ minio server ~/test
 ```
 
-Create new users following the multi-user guide [here](https://docs.min.io/docs/minio-multi-user-quickstart-guide.html)
+Create new users following the multi-user guide [here](https://docs.min.io/docs/minio-multi-user-quickstart-guide.html). `AssumeRole` is not supported without user.
 
 Testing with an example
 > Use the same username and password created in the previous steps.


docs/sts/assume-role.md
```diff
@@ -96,7 +96,7 @@ $ export MINIO_SECRET_KEY=minio123
 $ minio server ~/test
 ```
 
-Create new users following the multi-user guide [here](https://docs.min.io/docs/minio-multi-user-quickstart-guide.html)
+Create new users following the multi-user guide [here](https://docs.min.io/docs/minio-multi-user-quickstart-guide.html). `AssumeRole` is not supported without user.
```
harshavardhana, line 99: isn't this line indicative of the fact that you need to `create new users` to use this functionality? - why does adding an extra line help here?

harshavardhana (changes requested):

docs/sts/assume-role.md
ad-m: This may also indicate that using a separate user is best practice. In my case, I wanted to know how the functionality works, so I don't need follow strict secuirity best practice and I skip that step.

harshavardhana: This doc is meant to be followed exactly step by the step, that is what this doc is intended for. I don't know why anyone would assume that these are optional.  Adding a line ``AssumeRole` is not supported without the user.` makes it sound like a limitation. 

Just like we don't say that you need to have `server` online to make an S3 request. That is implicit.

PR #11371, opened 2021-01-28
Add a check to prevent removing an actively used tier

krisis:
## Description
This change prevents removal an actively used remote tier

## Motivation and Context
To avoid users from accidentally removing a remote tier that is being referred to by active ILM rules.

## How to test this PR?
1. Add a remote tier
```
➜  mc git:(storage-class)  ./mc admin tier add s3 minio-tier S3TIER-1 --endpoint https://play.minio.io --access-key Q3AM3UQ867SPQQA43P2F --secret-key zuf+tfteSlswRu7BJ86wekitnifILbZam1KYY3TG --bucket kp-sc-add --prefix s3-infreq
 ```
2. Create a bucket and configure ILM rule on it
```
➜  mc git:(storage-class) ./mc mb minio-tier/mybucket
Bucket created successfully `minio-tier/mybucket`.

➜  mc git:(storage-class) ./mc ilm add --transition-date "2021-01-28" --storage-class "S3TIER-1" minio-tier/mybucket 
Lifecycle configuration rule added with ID `c09ko6va7c2027e2mb40` to minio-tier/mybucket.
```
3. Try to remove the remote tier `S3TIER-1`
```
➜  mc git:(storage-class) ✗ ./mc admin tier rm minio-tier S3TIER-1
mc: <ERROR> Unable to remove remote tier. Specified remote tier is actively used by one or more ILM transition rules.
```
You should see the above error message.
## Types of changes

## Checklist:
- [ ] Documentation updated

### CODE REVIEW

File: cmd/bucket-metadata-sys.go [hunk 1]
```diff
@@ -420,6 +420,35 @@ func (sys *BucketMetadataSys) GetConfig(bucket string) (BucketMetadata, error) {
...
+		}
+		for _, rule := range bucketMeta.lifecycleConfig.Rules {
+			// skip disabled ILM rules
+			if rule.Status == lifecycle.Disabled {
+				continue
+			}
+			if rule.Transition.StorageClass == tierName {
+				return true
+			}
+		}
```
poornas, line 446: this by itself is not sufficient - some objects could have transitioned to this tier and the lifecycle rule could have been disabled or removed entirely after that. This will cause objects transitioned previously from being accessible if the tier config is deleted. Seems to me this has to be cached during scanner run to see if tier is in use.

krisis: yes, that's correct. I shall include additional validation to prevent the case you describe above. The idea is to list objects in the remote tier, allow tier to be deleted only if there was no objects found. This way we ensure that the remote-tier is not used by active ILM rules and that there are no previously transitioned objects. Does this make sense?

poornas:
can you change branch

PR #11403, opened 2021-02-01
Removing a tier is not supported

krisis:
## Description
Supporting safe removal of a remote tier involves,
- ensuring there are no transitioned objects to this tier
- ensuring that there are no active ILM rules referring this tier (through storage-class tag)

The cost of supporting remove tier outweighs its benefits. Later we could provide support for disabling a remote tier to avoid transitioning newer objects to a remote tier which is being 'retired'/'decommissioned'.

## Types of changes

## Checklist:
- [ ] Documentation updated


poornas (approved):
LGTM

krishnasrinivas (approved):

Issue #11421, opened 2021-02-03
Cluster with multiple zones fails to start


ToLToL:
I have Federation with multiple MinIO clusters. One of them has 2 zones, 3 nodes in each one.
I updated to the latest release: `RELEASE.2021-02-01T22-56-52Z` from `RELEASE.2020-10-18T21-54-12Z`.

Clusters with only 1 zone restart rapidly, in ~ 10 secondes.
However the cluster with 2 zones is blocked with this message since 14 hours:

```
Waiting for all MinIO sub-systems to be initialized.. trying to acquire lock
Waiting for all MinIO sub-systems to be initialized.. lock acquired
Verifying if {more than 3K} buckets are consistent across drives...
```

This cluster has:

20 GiB, 3k Buckets, 5k Objects, 24 drives

## Expected Behavior
A cluster with multiple zones should restart normally with the latest release: `RELEASE.2021-02-01T22-56-52Z`

## Current Behavior
Blocked with the following message since 14 hours:

```
Waiting for all MinIO sub-systems to be initialized.. trying to acquire lock
Waiting for all MinIO sub-systems to be initialized.. lock acquired
Verifying if {more than 3K} buckets are consistent across drives...
```

ToLToL:
Maybe it's related to https://github.com/minio/minio/pull/11423 ?

zapto2600:
I am seeing the same issue when we upgraded to `2021-02-01T22:56:52Z` This seems to only affect clusters that are setup as distributed.

harshavardhana:
@ToLToL can you paste your server command line?

ToLToL:
@harshavardhana

`minio server --certs-dir=/certs https://minio{0...2}.xxx.xxx.xxx/mnt/data{0...3} https://minio{3...5}.xxx.xxx.xxx/mnt/data{0...3}`
Worked fine with `RELEASE.2020-10-18T21-54-12Z`

harshavardhana:
> @harshavardhana
> 
> `minio server --certs-dir=/certs https://minio{0...2}.xxx.xxx.xxx/mnt/data{0...3} https://minio{3...5}.xxx.xxx.xxx/mnt/data{0...3}`
> Worked fine with `RELEASE.2020-10-18T21-54-12Z`

Did you upgrade all servers together? I.e all of them were upgraded?

ToLToL:
@harshavardhana yes

zapto2600:
> @harshavardhana yes

[...]

PR #11449, opened 2021-02-04
speed up heal buckets in parallel during startup

harshavardhana:
## Description
speed up heal buckets in parallel during startup

## Motivation and Context
speed up startup 

## How to test this PR?
refer #11421 

## Types of changes

## Checklist:
- [ ] Documentation updated


klauspost (commented):
I don't like this being unbounded. This open a huge number of connections `len(buckets) x drives_per_set` maybe even.

I added an alternative in #11457

harshavardhana:
> I added an alternative in #11457

Yes alternative is what I thought would be better as well. closing.

PR #12141, opened 2021-04-23
return the correct message when a bucket already exists

ghost:
## Description
When trying to create a bucket that already exists, the message can be confusing since you may not have been the one creating the bucket. This changes the error message.

## Motivation and Context
I was confused

## How to test this PR?
A test is included.

## Types of changes
- [ x ] Bug fix (non-breaking change which fixes an issue)

## Checklist:
- [ x ] Fixes a regression (If yes, please add `commit-id` or `PR #` here)
- [ ] Documentation updated

### DIFF
--- a/cmd/typed-errors.go
+++ b/cmd/typed-errors.go
@@ -69,9 +69,6 @@ var errNotFirstDisk = errors.New("Not first disk")
 // error returned by first disk waiting to initialize other servers.
 var errFirstDiskWait = errors.New("Waiting on other disks")
 
-// error returned when a bucket already exists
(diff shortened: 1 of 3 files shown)


harshavardhana:
> When trying to create a bucket that already exists, the message can be confusing since you may not have been the one creating the bucket. This changes the error message.

Please add a sign-off @brentadamson

harshavardhana:
Browser is not taking any more changes, as its code freeze and is scheduled to be removed from MinIO server in couple of weeks. Closing this PR.  All users are advised to start using https://github.com/minio/console

PR #12425, opened 2021-06-03
config: fix typo in KES CA path env. variable

aead:
## Description
This commit fixes a typo in the KES CA path
environment variable.

The CA certificate / directory for verifying
the KES server certificate should be specified
as `MINIO_KMS_KES_CA_PATH`.

## Motivation and Context
KES

## Types of changes

## Checklist:
- [ ] Documentation updated

### DIFF
--- a/internal/config/constants.go
+++ b/internal/config/constants.go
@@ -47,7 +47,7 @@ const (
 	EnvKESKeyName    = "MINIO_KMS_KES_KEY_NAME"
 	EnvKESClientKey  = "MINIO_KMS_KES_KEY_FILE"
 	EnvKESClientCert = "MINIO_KMS_KES_CERT_FILE"
-	EnvKESServerCA   = "MINIO_KMS_KES_CAPATH"
+	EnvKESServerCA   = "MINIO_KMS_KES_CA_PATH"
 
 	EnvEndpoints = "MINIO_ENDPOINTS" // legacy
 	EnvWorm      = "MINIO_WORM"      // legacy


internal/config/constants.go
```diff
@@ -47,7 +47,7 @@ const (
 	EnvKESKeyName    = "MINIO_KMS_KES_KEY_NAME"
 	EnvKESClientKey  = "MINIO_KMS_KES_KEY_FILE"
 	EnvKESClientCert = "MINIO_KMS_KES_CERT_FILE"
-	EnvKESServerCA   = "MINIO_KMS_KES_CAPATH"
+	EnvKESServerCA   = "MINIO_KMS_KES_CA_PATH"
```
harshavardhana, line 50: This is already used in k8s we should have a fallback @aead

harshavardhana (changes requested):

internal/config/constants.go
aead: Do we set this in operator? Because, I don't see it used there...

harshavardhana: So this means this broke MINIO_KMS_KES_CAPATH because k8s uses MINIO_KMS_KES_CA_PATH @aead and we don't have that ENV anymore..

aead:
> So this means this broke MINIO_KMS_KES_CAPATH because k8s uses MINIO_KMS_KES_CA_PATH @aead and we don't have that ENV anymore..

@harshavardhana Okay then we can drop this PR. It's simply a cosmetic fix. We can live with `MINIO_KMS_KES_CAPATH`

harshavardhana:
@aead then we need to fix the k8s to start using the correct ENV for kes deployments currently it uses CA_PATH

PR #12525, opened 2021-06-17
fs: Return faster on no ListObjects results

### LINKED ISSUE #12518: Listing empty bucket is far slower than listing non-empty bucket
## Expected Behavior

I would expect listing an empty bucket to be near-instantaneous.

## Current Behavior

Minio seems to take a minimum of 6ms to list an empty bucket on my system. This is an AMD Ryzen 5 with tmpfs.

The workaround is to ... well ... add a file! Listing a non-empty bucket can take 1ms.

I measured the above timings in my Python+botocore codebase. There, minio CPU usage is 350%, which surprises me because I thought I was requesting in serial. There may be problems in my codebase; but I won't reduce it because I can use `mc` to illustrate the minio-side problem.

Here's an illustration with `mc` timings. The slowdown is smaller when I use `mc` (and minio CPU stays at 100%). The slowdown is significant and easy to reproduce.

```
$ docker run --rm -p 9000:9000 --tmpfs /data minio/minio:RELEASE.2021-06-14T01-29-23Z server /data
$ mc config host add test http://localhost:9000 minioadmin minioadmin --api S3v4
$ mc mb test/abc/ # empty bucket
$ time for i in `seq 1 1000`; do mc ls test/abc/ >/dev/null; done

real    0m7.577s
user    0m3.548s
sys     0m2.106s
$ touch x
$ mc cp x test/abc/x # non-empty bucket
time for i in `seq 1 1000`; do mc ls test/abc/ >/dev/null; done

real    0m6.339s
user    0m3.449s
sys     0m2.196s
```

I would expect the non-empty timings to be imperceptibly slower than the empty-directory timings. Instead, listing an empty directory is ~20% slower than listing a non-empty directory.

## Context

I use minio for unit testing. I have 5 buckets, and I empty them between tests. Most tests take maybe 5ms and don't write anything to minio; so minio slows down my test suite 700%.

## Regression

This is a regression, but I can't pinpoint the exact date. My unit-test suite started taking twice as long as it used to. I think the minio release came after September 2020.

## Your Environment

* Version: `RELEASE.2021-06-14T01-29-23Z`
* Server setup and configuration: Docker+tmpfs
* Operating system

klauspost:
## Description

When no results are sent `result.end` is never sent, so the loop runs until the list is full.

Break immediately when channel is closed.

Fixes #12518

## Types of changes

### DIFF
--- a/cmd/object-api-common.go
+++ b/cmd/object-api-common.go
@@ -22,7 +22,7 @@ import (
 	"strings"
 	"sync"
 
-	humanize "github.com/dustin/go-humanize"
+	"github.com/dustin/go-humanize"
 	"github.com/minio/minio/internal/sync/errgroup"
 )
 
@@ -289,6 +289,7 @@ func listObjects(ctx context.Context, obj ObjectLayer, bucket, prefix, marker, d
 		if !ok {
 			// Closed channel.
 			eof = true
+			break
 		}
 
 		if HasSuffix(walkResult.entry, SlashSeparator) {


minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

harshavardhana (approved):

PR #12564, opened 2021-06-23
docs: add limits for bucket and object name length

iternity-dotcom:
## Description
Bucket names are limited to 63 chars (according to AWS).
Object names are limited to 1024 chars (according to AWS).
A '/' seperated segment of an object id may not exceed 255 chars.

## Motivation and Context
The information is currently missing in the minio-limits.md, but good to know for the users.

## How to test this PR?
Create objects and buckets with appropriate names.

## Types of changes

## Checklist:
- [x] Documentation updated

### DIFF
--- a/docs/minio-limits.md
+++ b/docs/minio-limits.md
@@ -28,6 +28,9 @@
 | Maximum number of parts returned per list parts request                         | 10000                                         |
 | Maximum number of objects returned per list objects request                     | 4500                                          |
 | Maximum number of multipart uploads returned per list multipart uploads request | 1000                                          |
+| Maximum length for bucket names                                                 | 63                                            |
+| Maximum length for object names                                                 | 1024                                          |
+| Maximum length for '/' seperated object name segment                            | 255                                           |
 
 ### List of Amazon S3 API's not supported on MinIO
 We found the following APIs to be redundant or less useful outside of AWS S3. If you have a different view on any of the APIs we missed, please open a [GitHub issue](https://github.com/minio/minio/issues).


minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

docs/minio-limits.md
```diff
@@ -28,6 +28,9 @@
 | Maximum number of parts returned per list parts request                         | 10000                                         |
 | Maximum number of objects returned per list objects request                     | 4500                                          |
 | Maximum number of multipart uploads returned per list multipart uploads request | 1000                                          |
+| Maximum length for bucket names                                                 | 63                                            |
+| Maximum length for object names                                                 | 1024                                          |
+| Maximum length for '/' seperated object name segment                            | 255                                           |
```
ravindk89, line 33: @harshavardhana non-blocking questions: Given the following, are we saying

`/bucket/segment1/segment2/segment3`

each `segment` must be < 255 chars long? Just checking - when I import these docs I will probably need to add an example here.

Also, for object names, does this include the slash delimiters? I assume so, but am thinking whether it starts at `/object-name` or `object`

ravindk89 (approved):

docs/minio-limits.md
harshavardhana: Yes this is correct, we include slash as well.

ravindk89: Good to know. I need to import these limits into next-gen docs, will see about adding examples later if necessary.

harshavardhana (approved):

docs/minio-limits.md
iternity-dotcom: @ravindk89 The only thing we're not 100% sure about is, if the 255 chars per segment is an official AWS limitation or a Minio specific one.
Where 1024 chars for object names and 63 chars for bucket names are official AWS limitations that also applies to Minio.

ravindk89: https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html - 3 -> 63 character bucket name limit

https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html - object key max name of 1024 bytes

These look like AWS limits. For the object name segment, I'll defer to @harshavardhana - I would venture this is a limit we encountered through reverse engineering the AWS API behavior, vs something AWS explicitly documents. But I could be wrong, so have asked for clarity.

Generally speaking, I do not believe we implement limits not imposed by AWS in the first place, whether those limits are explicitly documented or appear as part of observed behavior. If any such limits exist, I'll make it a point to call them out in the next-gen docs.

harshavardhana: The reason for segments on MinIO is because of the filesystem as namespace requirement for MinIO. We are database-free object storage, unlike every other vendor.

ravindk89: OK - that implies that out of those limits, only the segment limit is MinIO-specific. The other two are AWS inherited. @harshavardhana

harshavardhana: > OK - that implies that out of those limits, only the segment limit is MinIO-specific. The other two are AWS inherited. @harshavardhana 

Correct @ravind89

PR #12748, opened 2021-07-19
sts: add support for certificate-based authentication

aead:
## Description
This commit adds a new STS API for X.509 certificate
authentication.

A client can make a HTTP POST request over a TLS connection
and MinIO will verify the provided client certificate,
map it to a S3 policy and return temp. S3 credentials to the
client.

So, this STS API allows clients to authenticate with X.509
certificates over TLS and obtain temp. S3 credentials.

For more details and examples refer to the docs/sts/tls.md
documentation.

## Motivation and Context
TLS, STS

## How to test this PR?
See the docs/sts/tls.md

## Types of changes

## Checklist:
- [x] Documentation updated

### DIFF
--- a/cmd/sts-handlers.go
+++ b/cmd/sts-handlers.go
@@ -20,11 +20,13 @@ package cmd
 import (
 	"bytes"
 	"context"
+	"crypto/x509"
 	"encoding/base64"
 	"errors"
 	"fmt"
 	"net/http"
 	"strings"
+	"time"
 
 	"github.com/gorilla/mux"
 	"github.com/minio/minio/internal/auth"
@@ -48,10 +50,11 @@ const (
 	stsLDAPPassword     = "LDAPPassword"
 
 	// STS API action constants
-	clientGrants = "AssumeRoleWithClientGrants"
-	webIdentity  = "AssumeRoleWithWebIdentity"
-	ldapIdentity = "AssumeRoleWithLDAPIdentity"
-	assumeRole   = "AssumeRole"
+	clientGrants      = "AssumeRoleWithClientGrants"
+	webIdentity       = "AssumeRoleWithWebIdentity"
+	ldapIdentity      = "AssumeRoleWithLDAPIdentity"
+	clientCertificate = "AssumeRoleWithCertificate"
+	assumeRole        = "AssumeRole"
 
 	stsRequestBodyLimit = 10 * (1 << 20) // 10 MiB
 
@@ -124,6 +127,12 @@ func registerSTSRouter(router *mux.Router) {
 		Queries(stsVersion, stsAPIVersion).
 		Queries(stsLDAPUsername, "{LDAPUsername:.*}").
 		Queries(stsLDAPPassword, "{LDAPPassword:.*}")
+
+	// AssumeRoleWithCertificate
+	stsRouter.Methods(http.MethodPost).HandlerFunc(httpTraceAll(sts.AssumeRoleWithCertificate)).
+		Queries(stsAction, clientCertificate).
+		Queries(stsVersion, stsAPIVersion)
+
 }
 
 func checkAssumeRoleAuth(ctx context.Context, r *http.Request) (user auth.Credentials, isErrCodeSTS bool, stsErr STSErrorCode) {
@@ -648,4 +657,82 @@ func (sts *stsAPIHandlers) AssumeRoleWithLDAPIdentity(w http.ResponseWriter, r *
 	encodedSuccessResponse := encodeResponse(ldapIdentityResponse)
 
 	writeSuccessResponseXML(w, encodedSuccessResponse)
+}
+
+// AssumeRoleWithCertificate implements user authentication with client certificates.
+// It verifies the client-provided X.509 certificate, maps the certificate to an S3 policy
+// and returns temp. S3 credentials to the client.
+//
+// API endpoint: https://minio:9000?Action=AssumeRoleWithCertificate&Version=2011-06-15
+func (sts *stsAPIHandlers) AssumeRoleWithCertificate(w http.ResponseWriter, r *http.Request) {
+	var ctx = newContext(r, w, "AssumeRoleWithCertificate")
+
+	if !globalSTSTLSConfig.Enabled {
+		writeSTSErrorResponse(ctx, w, true, ErrSTSNotInitialized, errors.New("STS API 'AssumeRoleWithCertificate' is disabled"))
+		return
+	}
+
+	// We have to establish a TLS connection and the
+	// client must provide exactly one client certificate.
+	// Otherwise, we don't have a certificate to verify or
+	// the policy lookup would ambigious.
+	if r.TLS == nil {
+		writeSTSErrorResponse(ctx, w, true, ErrSTSInsecureConnection, errors.New("No TLS connection attempt"))
+		return
+	}
+	if len(r.TLS.PeerCertificates) == 0 {
+		writeSTSErrorResponse(ctx, w, true, ErrSTSMissingParameter, errors.New("No client certificate provided"))
+		return
+	}
+	if len(r.TLS.PeerCertificates) > 1 {
+		writeSTSErrorResponse(ctx, w, true, ErrSTSInvalidParameterValue, errors.New("More than one client certificate provided"))
+		return
+	}
+
+	var certificate = r.TLS.PeerCertificates[0]
+	if !globalSTSTLSConfig.InsecureSkipVerify {
+		_, err := certificate.Verify(x509.VerifyOptions{
+			KeyUsages: []x509.ExtKeyUsage{
+				x509.ExtKeyUsageClientAuth,
+			},
+			Roots: globalRootCAs,
+		})
+		if err != nil {
+			writeSTSErrorResponse(ctx, w, true, ErrSTSInvalidClientCertificate, err)
+			return
+		}
+	}
+
+	// We set the expiry of the temp. credentials to the minumim of the
+	// configured expiry and the duration until the certificate itself
+	// expires.
+	// We must not issue credentials that out-live the certificate.
+	var expiry = globalSTSTLSConfig.STSExpiry
+	if validUntil := time.Until(certificate.NotAfter); validUntil < expiry {
+		expiry = validUntil
+	}
+
+	tmpCredentials, err := auth.GetNewCredentialsWithMetadata(map[string]interface{}{
+		expClaim: time.Now().UTC().Add(expiry).Unix(),
+	}, globalActiveCred.SecretKey)
+	if err != nil {
+		writeSTSErrorResponse(ctx, w, true, ErrSTSInternalError, err)
+		return
+	}
+
+	// We map the X.509 subject common name to the policy. So, a client
+	// with the common name "foo" will be associated with the policy "foo".
+	// Other mapping functions - e.g. public-key hash based mapping - are
+	// possible but not implement.
+	//
+	// Group mapping is not possible with standard X.509 certificates.
+	if err = globalIAMSys.SetTempUser(tmpCredentials.AccessKey, tmpCredentials, certificate.Subject.CommonName); err != nil {
+		writeSTSErrorResponse(ctx, w, true, ErrSTSInternalError, err)
+		return
+	}
+
+	var response = new(AssumeRoleWithCertificateResponse)
+	response.Result.Credentials = tmpCredentials
+	response.Metadata.RequestID = w.Header().Get(xhttp.AmzRequestID)
+	writeSuccessResponseXML(w, encodeResponse(response))
 }
--- a/cmd/sts-handlers.go
+++ b/cmd/sts-handlers.go
@@ -703,23 +703,6 @@ func (sts *stsAPIHandlers) AssumeRoleWithCertificate(w http.ResponseWriter, r *h
 		}
 	}
 
-	// We set the expiry of the temp. credentials to the minimum of the
-	// configured expiry and the duration until the certificate itself
-	// expires.
-	// We must not issue credentials that out-live the certificate.
-	var expiry = globalSTSTLSConfig.STSExpiry
-	if validUntil := time.Until(certificate.NotAfter); validUntil < expiry {
-		expiry = validUntil
-	}
-
-	tmpCredentials, err := auth.GetNewCredentialsWithMetadata(map[string]interface{}{
-		expClaim: time.Now().UTC().Add(expiry).Unix(),
-	}, globalActiveCred.SecretKey)
-	if err != nil {
-		writeSTSErrorResponse(ctx, w, true, ErrSTSInternalError, err)
-		return
-	}
-
 	// We map the X.509 subject common name to the policy. So, a client
 	// with the common name "foo" will be associated with the policy "foo".
 	// Other mapping functions - e.g. public-key hash based mapping - are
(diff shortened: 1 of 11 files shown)


cmd/sts-handlers.go
```diff
@@ -593,3 +602,81 @@ func (sts *stsAPIHandlers) AssumeRoleWithLDAPIdentity(w http.ResponseWriter, r *
...
+			},
+			Roots: globalRootCAs,
+		})
+		if err != nil {
+			writeSTSErrorResponse(ctx, w, true, ErrSTSInvalidClientCertificate, err)
+			return
+		}
+	}
+
+	// We set the expiry of the temp. credentials to the minumim of the
```
harshavardhana:
```suggestion
	// We set the expiry of the temp. credentials to the minimum of the
```

cmd/sts-handlers.go
```diff
@@ -593,3 +602,81 @@ func (sts *stsAPIHandlers) AssumeRoleWithLDAPIdentity(w http.ResponseWriter, r *
...
+	}, globalActiveCred.SecretKey)
+	if err != nil {
+		writeSTSErrorResponse(ctx, w, true, ErrSTSInternalError, err)
+		return
+	}
+
+	// We map the X.509 subject common name to the policy. So, a client
+	// with the common name "foo" will be associated with the policy "foo".
+	// Other mapping functions - e.g. public-key hash based mapping - are
+	// possible but not implement.
```
harshavardhana:
```suggestion
	// possible but not implemented.
```

cmd/sts-handlers.go
```diff
@@ -593,3 +602,81 @@ func (sts *stsAPIHandlers) AssumeRoleWithLDAPIdentity(w http.ResponseWriter, r *
...
+	}
+
+	tmpCredentials, err := auth.GetNewCredentialsWithMetadata(map[string]interface{}{
+		expClaim: time.Now().UTC().Add(expiry).Unix(),
+	}, globalActiveCred.SecretKey)
+	if err != nil {
+		writeSTSErrorResponse(ctx, w, true, ErrSTSInternalError, err)
+		return
+	}
+
```
harshavardhana, line 745: We should also set `tmpCredentials.ParentUser = "client-tls:" + certificate.Subject.CommonName`  - I am assuming that CommonName is not optional here right?  and can be used to distinctively address the client ?

cmd/sts-handlers.go
```diff
@@ -593,3 +602,81 @@ func (sts *stsAPIHandlers) AssumeRoleWithLDAPIdentity(w http.ResponseWriter, r *
...
+	// We map the X.509 subject common name to the policy. So, a client
+	// with the common name "foo" will be associated with the policy "foo".
+	// Other mapping functions - e.g. public-key hash based mapping - are
+	// possible but not implement.
+	//
+	// Group mapping is not possible with standard X.509 certificates.
+	if err = globalIAMSys.SetTempUser(tmpCredentials.AccessKey, tmpCredentials, certificate.Subject.CommonName); err != nil {
```
harshavardhana: should we validate that `CommonName` is empty or not?

internal/http/server.go
```diff
@@ -172,6 +172,7 @@ func NewServer(addrs []string, handler http.Handler, getCert certs.GetCertificat
 			MinVersion:               tls.VersionTLS12,
 			NextProtos:               []string{"http/1.1", "h2"},
 			GetCertificate:           getCert,
+			ClientAuth:               tls.RequestClientCert,
```
harshavardhana, line 175: does this not break anything? @aead for regular TLS clients ?

harshavardhana (changes requested):

cmd/sts-handlers.go
aead: Doesn't `SetTempUser` check this? What happens if it is emtpy. Does the access key not a have any policy.

If not checked - we should.

internal/http/server.go
aead: Should not. See: https://pkg.go.dev/crypto/tls#ClientAuthType

Also, without requesting a client certificate it will not work. A TLS client can always choose to not send one.

cmd/sts-handlers.go
aead: What is the `ParentUser`?

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

cmd/sts-handlers.go
harshavardhana: ParentUser is needed to associate service accounts..

cmd/sts-handlers.go
harshavardhana: No, we need to enforce this at API level so that we expect that CommonName is always present. Expecting SetTempUser returning error doesn't have the context from the caller to indicate proper errors for the client.

harshavardhana: Also what is the TLS field that is non-optional? that can be used - IMO we should use that.

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

harshavardhana (approved):

harshavardhana (changes requested):
@aead how do we clean up the credentials STS and Service Accounts if the certs are expired on the client? how do we know that client cert is now invalid? 

should we allow accounts to automatically purge once the cert becomes invalid at a future date? I see that we do not save this information anywhere. 

Can you provide some clarity on this?

aead:
> how do we clean up the credentials STS and Service Accounts if the certs are expired on the client? how do we know that client cert is now invalid?

The temp. S3 credentials (access/secret key, session token) have an expiry. So they should be cleaned up once they expire.
In general, a X.509 has a `NotBefore` and `NotAfter` date limiting the time window the certificate is valid.
So, a service account should - at most - live until the `NotAfter` date.

For service accounts, this information (i.e. the `NotAfter` date) must be available without client sending the certificate / doing an STS API call? So creating a service account should not require the client certificate? If so, we need to preserve this information when the client performs the STS `AssumeRoleWithCertificate` call.

cmd/sts-handlers.go
donatello: In this situation, IIUC, each distinct trusted client certificate should be considered a distinct user in minio. This way each STS credential or service account credential generated with the same client certificate are associated together as credentials for the same "user" - this is the ParentUser field. In case the client certificate is revoked is revoked in some sense, then all the associated credentials may be deleted.

cmd/sts-handlers.go
donatello: Under this implementation, the ParentUser = Policy name associated with the user. This is a bit inflexible - you cannot assign multiple clients the same policy as they all need to have the same "CommonName" - is my understanding right?

docs/sts/tls.md
```diff
@@ -0,0 +1,126 @@
...
+The MinIO TLS STS API can be configured via MinIO's standard configuration API (i.e. using `mc admin config set/get`).
+Further, it can be configured via the following environment variables:
+
+```
+mc admin config set myminio identity_tls --env
+KEY:
+identity_tls  enable X.509 TLS certificate SSO support
+
+ARGS:
+MINIO_IDENTITY_TLS_STS_EXPIRY   (duration)  temporary credentials validity duration in s,m,h,d. Default is "1h"
```
donatello: To keep it consistent with other STS APIs, we should use a "DurationSeconds" query parameter in the API to set per-call expiry, and remove this environment configuration. We are also using an upper limit of 1 year validity for all STS creds.

donatello, line 21: To keep it consistent with other STS APIs, we should use a "DurationSeconds" query parameter in the API to set per-call expiry, and remove this environment configuration. We are also using an upper limit of 1 year validity for all STS creds.

docs/sts/tls.md
```diff
@@ -0,0 +1,126 @@
...
+MINIO_IDENTITY_TLS_SKIP_VERIFY  (on|off)    trust client certificates without verification. Defaults to "off" (verify)
```
donatello, line 16: Do we need a skip verify for TLS client auth? The purpose here is to authenticate the client, not just have a secure transport right?

donatello (changes requested):
Ideally we should get a client username and a policy mapping from the STS API call parameters - not sure if its possible though.

cmd/sts-handlers.go
aead: @harshavardhana The `Request.TLS` contains TLS status information (the TLS connection state) about the TLS connection over which the request has been sent/received. It contains, values like the SNI (server name that client has asked for), used ciphersuite, OCSP status information etc.
The entire `Request.TLS` field cannot be used. Imagine it like some status information about a TCP connection. It contains way to much volatile / unrecoverable data.

@donatello 
This is correct. However, there is no "standard" way to assign multiple clients (with different X.509 certificates) the same "value" derived from the certificate. The CN of each certificate must be unique. A CA must not issue the same CN for different clients.
The fact that there is a 1:1 mapping between client and policy is a known limitation.

docs/sts/tls.md
aead: Hm... I don't see such a query parameter for e.g. `AssumeRoleWithLDAPIdentity`. Also, that would mean clients can set the expiry (within a time window themself). That does not seem correct...

docs/sts/tls.md
aead: Yes, both. However, e.g. in case of LDAP an insecure transport is completely broken in the same way w.r.t. authentication. If you send your LDAP credentials over a plaintext connection / resp. a TLS connection without verifying to whom your talking then it's more or less trivial to obtain LDAP credentials. Either via inspecting the network traffic or via MitM.

The idea here is: You can use your self-signed certificates - e.g. to test that it works. Otherwise, you cannot test your client application without an actual certificate from your CA.

docs/sts/tls.md
donatello: It is used in the handler for AssumeRoleWithLDAP and in AssumeRoleWithWebIdentity - to set the expiry. Even AWS STS APIs have this parameter. Clients can set the expiry - AWS has a more restrictive upper limit, but we seem to have set it to 1 year now. @harshavardhana

harshavardhana: yeah AWS has an upper limit of 7 days we moved it to 365 days for a customer.

harshavardhana: > Hm... I don't see such a query parameter for e.g. `AssumeRoleWithLDAPIdentity`. Also, that would mean clients can set the expiry (within a time window themself). That does not seem correct...

We do we don't need a separate ENV to set the expiry - clients can ask for validity, we can of course reject STS requests that go beyond certificate validity if DurationSeconds lets say are beyond certificate not_after..

aead: That model seems odd. So a client can request credentials that are valid for an amount of time that is defined by the client itself. So a client can get credentials for 5 days or 10 days. I would assume the server defines the lifetime of the credentials, not the client.

I based `MINIO_IDENTITY_TLS_STS_EXPIRY` on https://github.com/minio/minio/blob/master/internal/config/identity/ldap/config.go#L99

harshavardhana: That ENV is going to be removed. Not sure why is it odd? 

This duration can never be bigger than client cert validity, that's all i.e the duration cannot go beyond the expiry date.

harshavardhana:
@aead will test this tomorrow

cmd/sts-handlers.go
aead: I don't think this is the expectation. A certificate is issued for a given entity and contains the public key that belongs to that entity. An entity can have multiple certificates at the same time.

A certificate does not necessarily correspond to a user. A user may have 1 or multiple certificates that are all valid.
Therefore, we should associate all service accounts to the entity that the certificate is issued for - e.g. the Subject or a SAN.

docs/sts/tls.md
```diff
@@ -0,0 +1,126 @@
...
+         <SessionToken>eyJhbGciOiJIUzUxMiIsInR5cCI6IkpXVCJ9.eyJhY2Nlc3NLZXkiOiJZQzEyWkJIVVZXNTg4QlFBRTVCTSIsImV4cCI6MTYyNjcyNTQ0NX0.wvMUf3w_x16qpVWgua8WxnV1Sgtv1jOnSu03vbrwOMzV3cI4q3_9WZD9LwlP-34DTsvbsg7gCBGh6YNriMMiQw</SessionToken>
+      </Credentials>
+   </AssumeRoleWithCertificateResult>
+   <ResponseMetadata>
+      <RequestId>169339CD8B3A6948</RequestId>
+   </ResponseMetadata>
+</AssumeRoleWithCertificateResponse>
+```
+
+## Authentication Flow
```
ravindk89, line 51: @aead how does this work with arbitrarily signed x.509 certificates? That is, does this API require that the client specify an x.509 cert signed by a known CA (CA is in ~/.minio/certs/CAs) ?

I'm specifically thinking of a scenario where an arbitrary client uses this API with a self-signed certificate with `CN = consoleAdmin` to gain admin access to the MinIO cluster.

I think most users would want to block this API to only those certificates signed by a trusted CA (at minimum). From there I might want to go farther and only trust those certificates whose CN matches some group filter, e.g.:

```
MINIO_IDENTITY_TLS_DN_FILTER="[OU=OrgUnit], [O=ORG], [L=City],[S=STATE],[C=COUNTRYCODE]
```

This would ensure that only those x.509 certs whose DN includes that specific string pattern after the CN would work.

So if I had:

```
MINIO_IDENTITY_TLS_DN_FILTER="OU=datalake,O=techops"
```

I would only expect the first of the following three x.509 certs to work with this API:

```
Subject: CN = ConsoleAdmin, OU=datalake,O=techops
Subject: CN = ConsoleAdmin, O=techops
Subject: CN = ConsoleAdmin
```

Let me know if this makes sense. This is a really powerful API, but since it is enabled by default IMHO we need some guardrails to prevent "guessing" access to roles by anyone with cert-minting capabilities.

I haven't fully parsed the code (nor would I be able to), so maybe this is already handled - but for the above example, for both trusted or untrusted CA, I think the secure expectation would be that we have controls for ensuring only specific x.509 certs can access MinIO.

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

harshavardhana:
I removed the _STS_EXPIRY environment variable instead we should be using DurationSeconds - and that is properly updated now. 

Also added additional claims for the temporary credentials for future use such as `subClaim, issClaim, audClaim` all populated from the relevant information from TLS certificates. 

@aead please can you update the support for https://github.com/minio/minio-go/pull/1520 for DurationSeconds(). 

This PR works as expected and tested with client certs.

harshavardhana (approved):

docs/sts/tls.md
harshavardhana: I removed the _TLS_EXPIRY..  environment.

harshavardhana:
> For service accounts, this information (i.e. the `NotAfter` date) must be available without client sending the certificate / doing an STS API call? So creating a service account should not require the client certificate? If so, we need to preserve this information when the client performs the STS `AssumeRoleWithCertificate` call.

Yes this needs to be but it can be stored based on the parentUser lookups, we can think about this later. We introduced expiring service accounts only recently.

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

docs/sts/tls.md
harshavardhana: This is something that we can look at it

ravindk89: I'll try testing this week. I'd like to at least confirm we restrict to trusted CA.

harshavardhana: mTLS won't work without it @ravindk89

ravindk89: Yeah, it's more for my own curiosity as well. Easier to document that way too.

dmitrybabeshko:
@harshavardhana @aead
I can't completely disable this with *MINIO_IDENTITY_TLS_SKIP_VERIFY* env variable set.  It is still asking for the client certificate. Had to revert to *RELEASE.2021-09-03T03-56-13Z* prior to this change. Please advise how to disable it.

image

dmytrobabeshko:
@harshavardhana @aead
I can't completely disable this with *MINIO_IDENTITY_TLS_ENABLE* env variable set.  It is still asking for the client certificate. Had to revert to *RELEASE.2021-09-03T03-56-13Z* prior to this change. Please advise how to disable it?

image

aead:
@dmitrybabeshko This is an STS API to obtain temp. S3 access/secret key pairs for regular S3 authentication.

For regular S3 operations you don't need to specify anything. MinIO should not require client certificates (mTLS) for regular S3 operations...

PR #12885, opened 2021-08-05
xl walk: Limit walk concurrent IO

klauspost:
## Description

We are observing heavy system loads, potentially locking the system up for periods when concurrent listing operations are performed.

We place a per-disk lock on walk IO operations. This will minimize the impact of concurrent listing operations on the entire system and de-prioritize them compared to other operations since only 1 will be in IO queue at any time.

Single list operations should remain largely unaffected.

## How to test this PR?

@harshavardhana has a reproducing setup with spinning disks where this is most prevalent.

## Types of changes

### DIFF
--- a/cmd/metacache-walk.go
+++ b/cmd/metacache-walk.go
@@ -121,7 +121,9 @@ func (s *xlStorage) WalkDir(ctx context.Context, opts WalkDirOptions, wr io.Writ
 		if contextCanceled(ctx) {
 			return ctx.Err()
 		}
+		s.walkMu.Lock()
 		entries, err := s.ListDir(ctx, opts.Bucket, current, -1)
+		s.walkMu.Unlock()
 		if err != nil {
 			// Folder could have gone away in-between
 			if err != errVolumeNotFound && err != errFileNotFound {
@@ -171,7 +173,9 @@ func (s *xlStorage) WalkDir(ctx context.Context, opts WalkDirOptions, wr io.Writ
 			// If root was an object return it as such.
 			if HasSuffix(entry, xlStorageFormatFile) {
 				var meta metaCacheEntry
+				s.walkMu.Lock()
 				meta.metadata, err = s.readMetadata(pathJoin(volumeDir, current, entry))
+				s.walkMu.Unlock()
 				if err != nil {
 					logger.LogIf(ctx, err)
 					continue
@@ -186,7 +190,9 @@ func (s *xlStorage) WalkDir(ctx context.Context, opts WalkDirOptions, wr io.Writ
 			// Check legacy.
 			if HasSuffix(entry, xlStorageFormatFileV1) {
 				var meta metaCacheEntry
+				s.walkMu.Lock()
 				meta.metadata, err = xioutil.ReadFile(pathJoin(volumeDir, current, entry))
+				s.walkMu.Unlock()
 				if err != nil {
 					logger.LogIf(ctx, err)
 					continue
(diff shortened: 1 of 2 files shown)


minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

harshavardhana:
Will test this today.

harshavardhana (changes requested):
@klauspost isn't it enough to do Lock() on ListDir()? instead of locking everywhere at ReadFile and readMetadata()? 

Also wouldn't this slowdown scanner-based healing as well? @klauspost - the main issue here is that locking would end up blocking everyone for all requests if there are many requests in progresss

harshavardhana:
I tested this and it helps but the problem now can be that if there are so many requests and waiting it can lead to everyone serialized - i.e normal calls that use any listing are blocked be it recursive or non-recursive. 

IMO I would suggest that we change the code to handle this for non-recursive listing alone such that regular directory navigation always works and has enough room for everyone. The most taxing is always recursive calls.

We can also have a top-level upper bound per bucket if we don't want to do locking based instead we can do bucket level throttling.

klauspost:
@harshavardhana Not sure I see a huge point in preferring one over the other. They get an equal opportunity to produce output and non-recursive can be  just as disruptive as recursive.

Spinning disks can only really serve one request at the time and NVME shouldn't really be too penalized by this. Also, it being a per-disk lock still allows for plenty of concurrent operations to occur on different disks.

So I believe that would be an unneeded complication.

harshavardhana:
> @harshavardhana Not sure I see a huge point in preferring one over the other. They get an equal opportunity to produce output and non-recursive can be just as disruptive as recursive.
> 
> Spinning disks can only really serve one request at the time and NVME shouldn't really be too penalized by this. Also, it being a per-disk lock still allows for plenty of concurrent operations to occur on different disks.
> 
> So I believe that would be an unneeded complication.

After the #12977 fix, everything is behaving as expected in terms of load, as the client's timeout server is not overloaded anymore. 

While this is true, I found the actual problem i.e prefix handling that is the fundamental problem for us.

harshavardhana:
We can table this PR for another rainy day, it's not needed right now.

klauspost:
@harshavardhana I still believe this would be a benefit for spinning disks.

harshavardhana:
> @harshavardhana I still believe this would be a benefit for spinning disks.

I will try this again tomorrow, need access to HDDs @klauspost - I am worried that this would slow down scanner listing.

cmd/metacache-walk.go
```diff
@@ -120,7 +120,9 @@ func (s *xlStorage) WalkDir(ctx context.Context, opts WalkDirOptions, wr io.Writ
 		if contextCanceled(ctx) {
 			return ctx.Err()
 		}
+		s.walkMu.Lock()
 		entries, err := s.ListDir(ctx, opts.Bucket, current, -1)
+		s.walkMu.Unlock()
```
harshavardhana, line 126: Should we only lock ListDir() instead of locking reading metadata since that may serialize a lot of calls that can be independent after ListDir() @klauspost ?

harshavardhana (changes requested):

cmd/metacache-walk.go
klauspost: The reads will be on the same disk on all cases, so not sure why we wouldn't include reading metadata.

For disks with a single ongoing listing, there will be close to no change since uncontested mutexes are fast and it doesn't operate concurrently. 

Again, it is to limit the potential impact of many concurrent listing requests. They will of course not all hit the same disks, but the disks that do should remain more responsive to non-listing requests.

harshavardhana (approved):
This looks good and works fine.

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

PR #12963, opened 2021-08-13
Make LDAP case-insensitive in DNs for username and group names

donatello:
## Description
- Involves changes in loading policy mappings from IAM object store and IAM etcd
store

## How to test this PR?

For an LDAP setup use https://github.com/donatello/minio-ldap-testing

Use `mc admin policy set myminio consoleAdmin user=UID=liZA,ou=people,ou=swengg,dc=min,dc=io` with different casings for the user DN and check that it is case-insensitive.

## Types of changes

## Checklist:
- [ ] Documentation updated

### DIFF
--- a/cmd/iam-dummy-store.go
+++ b/cmd/iam-dummy-store.go
@@ -45,6 +45,9 @@ func (ids *iamDummyStore) runlock() {
 	ids.RUnlock()
 }
 
+func (ids *iamDummyStore) setUsersSysType(u UsersSysType) {
+}
+
 func (ids *iamDummyStore) migrateBackendFormat(context.Context) error {
 	return nil
 }
--- a/cmd/iam-etcd-store.go
+++ b/cmd/iam-etcd-store.go
@@ -62,6 +62,8 @@ func extractPathPrefixAndSuffix(s string, prefix string, suffix string) string {
 type IAMEtcdStore struct {
 	sync.RWMutex
 
+	usersSysType UsersSysType
+
 	client *etcd.Client
 }
 
@@ -83,6 +85,10 @@ func (ies *IAMEtcdStore) rlock() {
 
 func (ies *IAMEtcdStore) runlock() {
 	ies.RUnlock()
+}
+
+func (ies *IAMEtcdStore) setUsersSysType(u UsersSysType) {
+	ies.usersSysType = u
 }
 
 func (ies *IAMEtcdStore) saveIAMConfig(ctx context.Context, item interface{}, itemPath string, opts ...options) error {
@@ -406,20 +412,12 @@ func (ies *IAMEtcdStore) loadMappedPolicy(ctx context.Context, name string, user
 		}
 		return err
 	}
-	m[name] = p
-	return nil
-}
-
-func getMappedPolicy(ctx context.Context, kv *mvccpb.KeyValue, userType IAMUserType, isGroup bool, m map[string]MappedPolicy, basePrefix string) error {
-	var p MappedPolicy
-	err := getIAMConfig(&p, kv.Value, string(kv.Key))
-	if err != nil {
-		if err == errConfigNotFound {
-			return errNoSuchPolicy
-		}
-		return err
+	// To implement LDAP case-insensitivity (and to avoid data migration -
+	// see comments in `loadMappedPolicies`), we lower-case the key name in
+	// the map `m`.
+	if ies.usersSysType == LDAPUsersSysType {
+		name = strings.ToLower(name)
 	}
-	name := extractPathPrefixAndSuffix(string(kv.Key), basePrefix, ".json")
 	m[name] = p
 	return nil
 }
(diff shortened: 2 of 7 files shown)


minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

klauspost:
Casual review, but just thinking if we should add a check to see if we are overwriting, where we have a conflict between `user=Liza`  and `user=liza`, potentially mixing permissions? That could be a security problem, but you might already be handling it.

donatello:
> Casual review, but just thinking if we should add a check to see if we are overwriting, where we have a conflict between `user=Liza` and `user=liza`, potentially mixing permissions? That could be a security problem, but you might already be handling it.

Under LDAP auth, the last argument to `mc admin policy set` must be either a user or group DN (distinguished name - they look like ordered comma separated key-value pairs) from LDAP. For LDAP DNs are completely _case-insensitive_. We should be overwriting for cases like:

```
mc admin policy set myminio policy1 user=UID=liZA,ou=people,ou=swengg,dc=min,dc=io
mc admin policy set myminio policy2 user=uid=liza,ou=people,ou=swengg,dc=min,dc=io
```

The above two commands differ in case for the last `user=xxx` argument but they refer to the same user.

vadmeste:
@donatello do we know that LDAP DN in RFC is case insensitive ? Because users using Active Directory can still overcome the issue by putting the exact same DN as shown by their LDAP server.

klauspost:
@donatello I am mostly worried about existing configs with duplicates when lower cased. Can that not happen?

harshavardhana:
> @donatello I am mostly worried about existing configs with duplicates when lower cased. Can that not happen?

This PR handles that conversion automatically.

donatello:
> @donatello I am mostly worried about existing configs with duplicates when lower cased. Can that not happen?

@klauspost If two DNs only differ in case, they are the same and we can overwrite/remove the existing policy mapping on disk, but to avoid a migration we are not doing that and instead overwriting the value in memory when a lower-case policy mapping key is present.

donatello:
> @donatello do we know that LDAP DN in RFC is case insensitive ? Because users using Active Directory can still overcome the issue by putting the exact same DN as shown by their LDAP server.

@vadmeste Yes, they are by default, but in rare cases there are exceptions according to many places that interpret the RFC:
https://ldapwiki.com/wiki/Distinguished%20Name%20Case%20Sensitivity
https://ldap.com/ldap-dns-and-rdns/
https://stackoverflow.com/questions/29897684/is-ldap-dn-case-insensitive

Also, the ldap library's comparison of DNs is case-insensitive: https://pkg.go.dev/github.com/go-ldap/ldap#DN.Equal

For us, this only appears to matter when applying the policy mapped to a user. 

@harshavardhana  We need to canonicalize the DN - here I chose to use lowercase. However looking at the parsing logic in the go library https://pkg.go.dev/github.com/go-ldap/ldap#ParseDN - it doesn't look like this solution completely correct - for e.g. there can be leading or trailing spaces in attributes and they should be ignored.

I want to discuss if we should be doing this at all - as we could just stick to the value of the DN returned by the LDAP server as before. If we should make it case-insensitive, I think we may need a more careful approach to canonicalize the DN, but this is not specified in the RFC either - quoting from https://datatracker.ietf.org/doc/html/rfc4514 :

> This document does not define a canonical string representation for
   DNs.  Comparison of DNs for equality is to be performed in accordance
   with the distinguishedNameMatch matching rule [RFC4517].

harshavardhana:
> I want to discuss if we should be doing this at all - as we could just stick to the value of the DN returned by the LDAP server as before. If we should make it case-insensitive, I think we may need a more careful approach to canonicalize the DN, but this is not specified in the RFC either - quoting from https://datatracker.ietf.org/doc/html/rfc4514 :
> 
> >

Yeah if we don't need to do it and can be fixed by using the appropriate DNs then we are good, perhaps just documenting would be fine. 

// cc @ravindk89

ravindk89:
I'm taking a *very* rough reading of [RFC4517 4.2.15 - distinguishedNameMatch](https://datatracker.ietf.org/doc/html/rfc4517#section-4.2.15) - but agree that there does not seem to be a hard rule on case sensitivity here.

To confirm @donatello @harshavardhana , our current behavior is case *sensitivity*, where the following DNs are *not* equal:

```
user=UID=liZA,ou=people,ou=swengg,dc=min,dc=io
user=uid=liza,ou=people,ou=swengg,dc=min,dc=io
```

If we keep this behavior, then I can update the documentation [here](https://docs.min.io/minio/baremetal/security/ad-ldap-external-identity-management/external-authentication-with-ad-ldap-identity-provider.html#access-control-for-externally-managed-identities) to explicitly remind users that MinIO is case-sensitive for the purpose of DN matching (as above). AD/LDAP admins are responsible for maintaining consistency of case.

harshavardhana:
> I'm taking a _very_ rough reading of [RFC4517 4.2.15 - distinguishedNameMatch](https://datatracker.ietf.org/doc/html/rfc4517#section-4.2.15) - but agree that there does not seem to be a hard rule on case sensitivity here.
> 
> To confirm @donatello @harshavardhana , our current behavior is case _sensitivity_, where the following DNs are _not_ equal:
> 
> ```
> user=UID=liZA,ou=people,ou=swengg,dc=min,dc=io
> user=uid=liza,ou=people,ou=swengg,dc=min,dc=io
> ```
> 
> If we keep this behavior, then I can update the documentation [here](https://docs.min.io/minio/baremetal/security/ad-ldap-external-identity-management/external-authentication-with-ad-ldap-identity-provider.html#access-control-for-externally-managed-identities) to explicitly remind users that MinIO is case-sensitive for the purpose of DN matching (as above). AD/LDAP admins are responsible for maintaining consistency of case.

@donatello we can slate this for the future for another rainy day, right now we do not have much clarity.

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

klauspost:
Keeping this for a rainy day.

PR #13088, opened 2021-08-27
Transition without re-reading object info

klauspost:
## Description

Queue objects for transition without reading consistent object info, but do that in the worker instead.

This gives faster scanner operation and more up-to-date metadata for the transition operation.

@krisis I don't know if we should re-run the checks against the lifecycle, or if there is enough checks for state changes for it to not do something weird, if some state changed.

Since it is an async process it *should* be rechecking, and I can see it fairly quickly [re-reads the metadata](https://github.com/minio/minio/blob/bbf3576f70390309c4ff73f7f15a20223b383764/cmd/erasure-object.go#L1489), so I assume this is safe.

It might even be safe enough without re-reading in `worker`, but I wanted to double check before doing that.

## How to test this PR?

Regular transition setup.

## Types of changes

### DIFF
--- a/cmd/bucket-lifecycle.go
+++ b/cmd/bucket-lifecycle.go
@@ -180,8 +180,15 @@ func (t *transitionState) worker(ctx context.Context, objectAPI ObjectLayer) {
 				return
 			}
 			atomic.AddInt32(&t.activeTasks, 1)
-			if err := transitionObject(ctx, objectAPI, oi); err != nil {
-				logger.LogIf(ctx, fmt.Errorf("Transition failed for %s/%s version:%s with %w", oi.Bucket, oi.Name, oi.VersionID, err))
+			// Retrieve consistent object info.
+			obj, err := objectAPI.GetObjectInfo(ctx, oi.Bucket, oi.Name, ObjectOptions{
+				VersionID: oi.VersionID,
+			})
+			if err == nil {
+				// Ignore errors, the object may have changed since it was queued.
+				if err := transitionObject(ctx, objectAPI, obj); err != nil {
+					logger.LogIf(ctx, fmt.Errorf("Transition failed for %s/%s version:%s with %w", oi.Bucket, oi.Name, oi.VersionID, err))
+				}
 			}
 			atomic.AddInt32(&t.activeTasks, -1)
 		}
--- a/cmd/data-scanner.go
+++ b/cmd/data-scanner.go
@@ -921,6 +921,9 @@ func (i *scannerItem) applyLifecycle(ctx context.Context, o ObjectLayer, oi Obje
 	switch action {
 	case lifecycle.DeleteAction, lifecycle.DeleteVersionAction:
 	case lifecycle.TransitionAction, lifecycle.TransitionVersionAction:
+		// Apply without re-reading object info.
+		// This is done by the worker.
+		return applyLifecycleAction(ctx, action, o, oi), size
 	case lifecycle.DeleteRestoredAction, lifecycle.DeleteRestoredVersionAction:
 	default:
 		// No action.


minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

cmd/bucket-lifecycle.go
```diff
@@ -180,8 +180,15 @@ func (t *transitionState) worker(ctx context.Context, objectAPI ObjectLayer) {
 				return
 			}
 			atomic.AddInt32(&t.activeTasks, 1)
-			if err := transitionObject(ctx, objectAPI, oi); err != nil {
-				logger.LogIf(ctx, fmt.Errorf("Transition failed for %s/%s version:%s with %w", oi.Bucket, oi.Name, oi.VersionID, err))
+			// Retrieve consistent object info.
+			obj, err := objectAPI.GetObjectInfo(ctx, oi.Bucket, oi.Name, ObjectOptions{
+				VersionID: oi.VersionID,
+			})
```
poornas, line 186: getObjectFileInfo is already done inside TransitionObject in the erasure layer to ensure consistent object info is used, this call is unnecessary.

poornas (changes requested):

cmd/bucket-lifecycle.go
harshavardhana: @poornas this was done for a reason to avoid GetObjectInfo() inside the scanner instead move it to worker

poornas: Understood that, it is just that TransitionObject already fetches object info and returns if there is an error or transition completed - so this GetObjectInfo call is redundant.

krisis: @klaus, like @poornas mentions above, we can avoid this GetObjectInfo call too. Since object version's content/metadata are immutable, we only run the risk of this version being deleted before this worker gets to it. This is handled by erasureObject.TransitionObject method.

krisis (commented):
Great catch!

harshavardhana:
```diff
diff --git a/cmd/data-scanner.go b/cmd/data-scanner.go
index d9b140251..9ec080667 100644
--- a/cmd/data-scanner.go
+++ b/cmd/data-scanner.go
@@ -927,30 +927,9 @@ func (i *scannerItem) applyLifecycle(ctx context.Context, o ObjectLayer, oi Obje
                return false, size
        }
 
-       obj, err := o.GetObjectInfo(ctx, i.bucket, i.objectPath(), ObjectOptions{
-               VersionID: versionID,
-       })
-       if err != nil {
-               switch err.(type) {
-               case MethodNotAllowed: // This happens usually for a delete marker
-                       if !obj.DeleteMarker { // if this is not a delete marker log and return
-                               // Do nothing - heal in the future.
-                               logger.LogIf(ctx, err)
-                               return false, size
-                       }
-               case ObjectNotFound, VersionNotFound:
-                       // object not found or version not found return 0
-                       return false, 0
-               default:
-                       // All other errors proceed.
-                       logger.LogIf(ctx, err)
-                       return false, size
-               }
-       }
-
-       action = evalActionFromLifecycle(ctx, *i.lifeCycle, obj, i.debug)
+       action = evalActionFromLifecycle(ctx, *i.lifeCycle, oi, i.debug)
        if action != lifecycle.NoneAction {
-               applied = applyLifecycleAction(ctx, action, o, obj)
+               applied = applyLifecycleAction(ctx, action, o, oi)
        }
 
        if applied {
```

This change is also useful

harshavardhana:
and the rest in 

```diff
diff --git a/cmd/bucket-lifecycle.go b/cmd/bucket-lifecycle.go
index 8a5d84af1..364af1e66 100644
--- a/cmd/bucket-lifecycle.go
+++ b/cmd/bucket-lifecycle.go
@@ -179,11 +179,17 @@ func (t *transitionState) worker(ctx context.Context, objectAPI ObjectLayer) {
                        if !ok {
                                return
                        }
-                       atomic.AddInt32(&t.activeTasks, 1)
-                       if err := transitionObject(ctx, objectAPI, oi); err != nil {
-                               logger.LogIf(ctx, fmt.Errorf("Transition failed for %s/%s version:%s with %w", oi.Bucket, oi.Name, oi.VersionID, err))
+                       obj, err := objectAPI.GetObjectInfo(ctx, oi.Bucket, oi.Name, ObjectOptions{
+                               VersionID: oi.VersionID,
+                       })
+                       if err == nil {
+                               atomic.AddInt32(&t.activeTasks, 1)
+                               if err := transitionObject(ctx, objectAPI, obj); err != nil {
+                                       logger.LogIf(ctx, fmt.Errorf("Transition failed for %s/%s version:%s with %w", obj.Bucket,
+                                               obj.Name, obj.VersionID, err))
+                               }
+                               atomic.AddInt32(&t.activeTasks, -1)
                        }
-                       atomic.AddInt32(&t.activeTasks, -1)
                }
        }
 }
```

If we can remove this as well we can live with it too AFAICS.

harshavardhana:
@klauspost #13091 would do more on the ILM side including expiration - we can drop this PR and use the newer one.

PR #13231, opened 2021-09-17
de-couple walkMu and walkReadMu for some granularity

harshavardhana:
## Description
de-couple walkMu and walkReadMu for some granularity

## Motivation and Context
This commit brings two locks instead of single lock for
WalkDir() calls on top of c25816eabc7712893b177b29b4820df0cb81da58.
    
The main reason is to avoid contention between readMetadata()
and ListDir() calls, ListDir() can take time on prefixes that
are huge for readdir() but this shouldn't end up blocking
all readMetadata() operations, this allows for more room for
I/O while not overly penalizing all listing operations.

## How to test this PR?
Nothing special, everything should work as expected.

## Types of changes

## Checklist:
- [ ] Documentation updated

### DIFF
--- a/cmd/metacache-walk.go
+++ b/cmd/metacache-walk.go
@@ -121,6 +121,7 @@ func (s *xlStorage) WalkDir(ctx context.Context, opts WalkDirOptions, wr io.Writ
 		if contextCanceled(ctx) {
 			return ctx.Err()
 		}
+
 		s.walkMu.Lock()
 		entries, err := s.ListDir(ctx, opts.Bucket, current, -1)
 		s.walkMu.Unlock()
@@ -173,9 +174,9 @@ func (s *xlStorage) WalkDir(ctx context.Context, opts WalkDirOptions, wr io.Writ
 			// If root was an object return it as such.
 			if HasSuffix(entry, xlStorageFormatFile) {
 				var meta metaCacheEntry
-				s.walkMu.Lock()
+				s.walkReadMu.Lock()
 				meta.metadata, err = s.readMetadata(pathJoin(volumeDir, current, entry))
-				s.walkMu.Unlock()
+				s.walkReadMu.Unlock()
 				if err != nil {
 					logger.LogIf(ctx, err)
 					continue
@@ -190,9 +191,9 @@ func (s *xlStorage) WalkDir(ctx context.Context, opts WalkDirOptions, wr io.Writ
 			// Check legacy.
 			if HasSuffix(entry, xlStorageFormatFileV1) {
 				var meta metaCacheEntry
-				s.walkMu.Lock()
+				s.walkReadMu.Lock()
 				meta.metadata, err = xioutil.ReadFile(pathJoin(volumeDir, current, entry))
-				s.walkMu.Unlock()
+				s.walkReadMu.Unlock()
 				if err != nil {
 					logger.LogIf(ctx, err)
 					continue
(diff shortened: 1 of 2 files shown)


harshavardhana:
The main reason is this to do with why serializing is not working well with the concurrent listing, drives can get blocked on listing when they are concurrently attempted. This can lead to a complication with a large build-up of go-routines that are all blocked fully which can eventually lead to DDOS-type scenarios.

Instead, we can have a token bucket system with an upper limit of tokens available, if no token is available the call shall Wait() for `x` amount of time until new tokens are available.

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

cmd/metacache-walk.go
```diff
@@ -30,8 +30,12 @@ import (
 	xhttp "github.com/minio/minio/internal/http"
 	xioutil "github.com/minio/minio/internal/ioutil"
 	"github.com/minio/minio/internal/logger"
+	"golang.org/x/time/rate"
 )
 
+// Rate is 10 events/sec and permits burst of at most 16 events.
+var limiter = rate.NewLimiter(10, 16)
```
harshavardhana: We can eventually make this configurable as needed...

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

klauspost:
👎🏼 This is strictly worse.

The current scheme is good. As proven by customer reports that "listings are slow, but read/writes are fast". You sent it to me myself. It *doesn't* serialize listings as you state, it just slows each one down so the system doesn't clog up. That is the exact point of it.

Issuing 16 concurrent read requests to HDDs does not make them faster, it just de-prioritizes other reads and writes and makes the whole system slow instead of just the listings. Furthermore completely blocking requests makes it impossible to return *any* listing results while 16 listings are running. It will be impossible for the blocked requests to begin streaming back already found listing results.

When issuing strict listings there *will* be a high IO load on the system, that is after all there is a reason it is an option. However load will be spread across all running listings. That is why the locking is so fine grained and *only* when IO is performed.

Yes, systems will list slowly when you issue a lot of concurrent requests, but that is an IO limit and this is by far the best solution for that. Each request will progress, but slower than when one listing. Naturally.

You can rate limit S3 API calls, if you are concerned about DDOS. Your proposal doesn't solve that anyway, since requests will still be built up.

harshavardhana:
> 👎🏼 This is strictly worse.
> 
> The current scheme is good. As proven by customer reports that "listings are slow, but read/writes are fast". You sent it to me myself. It _doesn't_ serialize listings as you state, it just slows each one down so the system doesn't clog up. That is the exact point of it.
> 
> Issuing 16 concurrent read requests to HDDs does not make them faster, it just de-prioritizes other reads and writes and makes the whole system slow instead of just the listings. Furthermore completely blocking requests makes it impossible to return _any_ listing results while 16 listings are running. It will be impossible for the blocked requests to begin streaming back already found listing results.
> 
> When issuing strict listings there _will_ be a high IO load on the system, that is after all there is a reason it is an option. However load will be spread across all running listings. That is why the locking is so fine grained and _only_ when IO is performed.
> 
> Yes, systems will list slowly when you issue a lot of concurrent requests, but that is an IO limit and this is by far the best solution for that. Each request will progress, but slower than when one listing. Naturally.
> 
> You can rate limit S3 API calls, if you are concerned about DDOS. Your proposal doesn't solve that anyway, since requests will still be built up.

No there is a fundamental problem here @klauspost as you can clearly see when a namespace is bigger a single directory listing can block all other listing.

This PR simply removes that and instead makes it 16 at a time this way we do not block the system WalkDir().

Especially this is true with the recent report that we have seen, almost all of them are blocked on WalkDir()

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

kannappanr (approved):
LGTM

PR #13232, opened 2021-09-17
s3: Put bucket tagging to return an error when bucket is not found

vadmeste:
## Description
loadBucketMetadata() should return an error when the bucket metadata
file doesn't exist. This should not cause a problem since the bucket metadata
file is always created already when a new bucket is created.

## Motivation and Context
PutBucketTagging handler should return bucket not found error when the bucket
doesn't exist.  

## How to test this PR?
aws s3api put-bucket-tagging --debug --endpoint http://localhost:9000 --profile minio --bucket testbucket  --tagging '{"TagSet": [{ "Key": "organization", "Value": "marketing" }]}'

## Types of changes

## Checklist:
- [ ] Documentation updated

### DIFF
--- a/cmd/api-errors.go
+++ b/cmd/api-errors.go
@@ -1796,6 +1796,8 @@ func toAPIErrorCode(ctx context.Context, err error) (apiErr APIErrorCode) {
 	}
 
 	switch err {
+	case errVolumeNotFound:
+		apiErr = ErrNoSuchBucket
 	case errInvalidArgument:
 		apiErr = ErrAdminInvalidArgument
 	case errNoSuchUser:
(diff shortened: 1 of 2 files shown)


minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

klauspost (approved):
LGTM.

harshavardhana:
@vadmeste a mint test might make sense here.

harshavardhana (approved):

caugner:
@harshavardhana Unfortunately, this is the second unannounced **breaking change** in a month.

Previously, it would suffice to create an empty directory with the bucket name in the data directory, now Minio rejects that, because the bucket metadata file is missing. 🤷 

Please reconsider migrating to semantic versioning, and avoid making things stricter in a so-called "minor bugfix release".

klauspost:
@caugner Modifying the backend manually has never, and never will be supported.

If you could describe your problem instead of just complaining we might be able to actually help you.

If this is a problem in NAS mode, then it is a problem that should be fixed AFAICT.

jmsmkn:
Indeed this works fine in NAS mode, a very simple change to my project, and happy to know I'm doing things correctly now!

Once you're at minio's scale every change will break something for someone so no problems with the versioning from me, I only tagged this PR for my own tracking. Thank you for developing and maintaining this excellent software, it's enabled us to do a lot more a lot faster ❤

PR #13243, opened 2021-09-18
Fix goroutine leak when remote target is edited

poornas:
## Motivation and Context
Remote target has a healthcheck goroutine that should be canceled when remote target is modified via `mc admin bucket remote edit` command.

## How to test this PR?
Edit existing remote target, then see number of healthcheck routines with `mc admin profile start|stop`

## Types of changes

## Checklist:
- [ ] Documentation updated


harshavardhana (approved):

harshavardhana (changes requested):
can you pull in the `minio-go` update as well in this PR @poornas

harshavardhana (approved):

PR #13249, opened 2021-09-19
fix: fallback listing on drives that are unformatted, disconnected

harshavardhana:
## Description
fix: fallback listing on drives that are unformatted, disconnected

## Motivation and Context
Make sure to skip the offline disks appropriately.

## How to test this PR?
Nothing should change, generally an optimization. 

## Types of changes

## Checklist:
- [ ] Documentation updated

### DIFF
--- a/cmd/metacache-server-pool.go
+++ b/cmd/metacache-server-pool.go
@@ -209,6 +209,10 @@ func (z *erasureServerPools) listPath(ctx context.Context, o *listPathOptions) (
 			}
 		}()
 		o.ID = ""
+
+		if err != nil {
+			logger.LogIf(ctx, fmt.Errorf("Resuming listing from drives failed %w, proceeding to do raw listing", err))
+		}
 	}
 
 	// Do listing in-place.
--- a/cmd/metacache-set.go
+++ b/cmd/metacache-set.go
@@ -365,12 +365,14 @@ func (er *erasureObjects) streamMetadataParts(ctx context.Context, o listPathOpt
 		}
 
 		const retryDelay = 500 * time.Millisecond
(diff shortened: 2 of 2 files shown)


minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

klauspost (changes requested):
This is needed. The first block is often not written when the client returns, and it may be partially written.

I don't see anything unnecessary.

harshavardhana:
> This is needed. The first block is often not written when the client returns, and it may be partially written.
> 
> I don't see anything unnecessary.

@klauspost are we sure about that? - I don't see a retry is needed here since that involves unnecessary round trips. 

Especially if the resumable client comes before the server has written newer block? isn't this already happening at the same time while we return the results?

I would like to really reproduce the behavior here and see does it really make sense.

klauspost:
> Especially if the resumable client comes before the server has written newer block? isn't this already happening at the same time while we return the results?

No. When we start `listAndSave` the results are split, so the request will get a copy of the first `max_keys` results, which will return as soon as they are done. 

Filling a block requires 5K entries (unless it is done), which most often is more than the first request `max_keys`. So the returning request (typically the 3rd) will usually have to wait for the 1st block to be finished. On fast listing HW this will not be a wait, but on spinning disks it will typically have to wait and retry.

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

krisis (approved):

poornas (approved):

PR #13276, opened 2021-09-22
Allow only account-key updates for azure tier

krisis:
## Description
Editing azure remote tier account-name could result in previously transitioned objects becoming inaccessible.
Edit-tier functionality is meant to only update secret key when they are rotated for this IAM user in the respective cloud storage. This changes removes the possibility of updating storage account name for an azure tier.

## Motivation and Context
To avoid user from accidentally changing azure tier's storage account leaving previously transtioned objects inaccessible.

## How to test this PR?
1. add an azure tier
2. edit this tier's account key (note: this requires an updated mc)

## Types of changes

## Checklist:
- [ ] Documentation updated

### DIFF
--- a/cmd/tier.go
+++ b/cmd/tier.go
@@ -146,7 +146,7 @@ func (config *TierConfigMgr) Edit(ctx context.Context, tierName string, creds ma
 		return errTierNotFound
 	}
 
-	newCfg := config.Tiers[tierName]
+	cfg := config.Tiers[tierName]
 	switch tierType {
 	case madmin.S3:
 		if (creds.AccessKey == "" || creds.SecretKey == "") && !creds.AWSRole {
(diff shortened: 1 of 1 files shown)


minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

harshavardhana (approved):

vadmeste (commented):
LGTM but wondering why we don't do the same thing with S3 backend

harshavardhana:
> LGTM but wondering why we don't do the same thing with S3 backend

Since that can change anytime and it's not associated in the same manner as an account. 

https://github.com/minio/mc/pull/3805#issuecomment-925378370 quoting @krisis here

> This is best explained using an example. Imagine a setup where a user configures an azure type tier and configures bucket ILM policies to transition objects to this tier. Over time we would have objects transitioned to this tier which is backed by a storage account, let's call this storageacct1. If this user decides to update the storage account to storageacct2 then objects that were transitioned earlier would be looked up in storageacct2 but they remain in storageacct1. This problem doesn't exist with other tier types, namely s3 and gcs. In each of these the credentials correspond to an IAM user and doesn't change the part of namespace this tier is referring to.

vadmeste:
> Since that can change anytime and it's not associated in the same manner as an account.

oh yes, account name here is storage account name. Thanks.

vadmeste (approved):
LGTM

PR #13341, opened 2021-10-01
Print log when EINVALID is encountered in storage layer

vadmeste:
EINVALID from the OS is not a common case and should be logger.
## Description
EINVALID from the OS is not a common case and should be logger.

## Motivation and Context
Simple log adding

## How to test this PR?
Trivial

## Types of changes

## Checklist:
- [ ] Documentation updated

### DIFF
--- a/cmd/storage-errors.go
+++ b/cmd/storage-errors.go
@@ -17,7 +17,12 @@
 
 package cmd
 
-import "errors"
+import (
+	"context"
+	"errors"
+
+	"github.com/minio/minio/internal/logger"
+)
 
 // errUnexpected - unexpected error, requires manual intervention.
 var errUnexpected = StorageErr("unexpected error, please report this issue at https://github.com/minio/minio/issues")
@@ -157,6 +162,7 @@ func osErrToFileErr(err error) error {
 		return errFaultyDisk
 	}
 	if isSysErrInvalidArg(err) {
+		logger.LogIf(context.Background(), err)
 		// For some odd calls with O_DIRECT reads
 		// filesystems can return EINVAL, handle
 		// these as FileNotFound instead.


cmd/storage-errors.go
```diff
@@ -157,6 +162,7 @@ func osErrToFileErr(err error) error {
 		return errFaultyDisk
 	}
 	if isSysErrInvalidArg(err) {
+		logger.LogIf(context.Background(), err)
```
klauspost, line 165: We don't really do logging here, and it seems like an unintentional side effect. 

You could argue that a lot of these would warrant a log entry `isSysErrHandleInvalid, isSysErrIO, isSysErrTooManyFiles,isSysErrNoSpace` seems pretty severe, but aren't logged.

I guess it is being ignored since it returns a "harmless" error. Maybe we should "upgrade" it instead, maybe `errFaultyDisk`?

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

cmd/storage-errors.go
vadmeste: yes, the only true reason for logging this is that we convert this error to errFileNotFound, so debugging this will be hard.

harshavardhana: The reason this will happen @vadmeste is very rare and we don't have to log this error. 

This is mainly a filesystem related error due to some flags not implemented etc.

EINVAL to errFileNotFound was added because some random characters in object name with O_DIRECT throw EINVAL and the object doesn't exist..

vadmeste: We actually need to log this because it is very rare. My only aim of this is that it will be easier for us to debug when this happens. If an S3 application complains about a 404 error, we can waste multiple hours debuggining to understand why this happened.

harshavardhana (approved):

klauspost (approved):

PR #13383, opened 2021-10-08
allow disabling strict sha256 validation with some broken clients

harshavardhana:
## Description
allow disabling strict sha256 validation with some broken clients

## Motivation and Context
with some broken clients allow non-strict validation
of sha256 when ContentLength > 0, it has been found in
the wild some applications that need this behavior. This
shall be only allowed if `--no-compat` is used.

## How to test this PR?
Nothing special, there are few proprietary clients that
have this behavior.

## Types of changes

## Checklist:
- [ ] Documentation updated

### DIFF
--- a/cmd/signature-v4-utils.go
+++ b/cmd/signature-v4-utils.go
@@ -55,9 +55,30 @@ func skipContentSha256Cksum(r *http.Request) bool {
 		v, ok = r.Header[xhttp.AmzContentSha256]
 	}
 
+	// Skip if no header was set.
+	if !ok {
+		return true
+	}
+
 	// If x-amz-content-sha256 is set and the value is not
 	// 'UNSIGNED-PAYLOAD' we should validate the content sha256.
-	return !(ok && v[0] != unsignedPayload)
+	switch v[0] {
+	case unsignedPayload:
+		return true
+	case emptySHA256:
+		// some broken clients set empty-sha256
+		// with > 0 content-length in the body,
+		// we should skip such clients and allow
+		// blindly such insecure clients only if
+		// S3 strict compatibility is disabled.
+		if r.ContentLength > 0 && !globalCLIContext.StrictS3Compat {
+			// We return true only in situations when
+			// deployment has asked MinIO to allow for
+			// such broken clients and content-length > 0.
+			return true
+		}
+	}
+	return false
 }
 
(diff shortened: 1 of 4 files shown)


minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

vadmeste (approved):
LGTM

cmd/signature-v4-utils.go
```diff
@@ -57,7 +57,25 @@ func skipContentSha256Cksum(r *http.Request) bool {
...
+			if r.ContentLength > 0 && !globalCLIContext.StrictS3Compat {
+				// We return true only in situations when
+				// deployment has asked MinIO to allow for
+				// such broken clients.
+				return true
+			}
+		}
+		return false
+	}
+	return true
```
klauspost: Simplify and skip check if length is actually 0.

```suggestion
	// Skip if no header was set.
	if !ok {
		return true
	}
	// If x-amz-content-sha256 is set and the value is not
	// 'UNSIGNED-PAYLOAD' we should validate the content sha256.
	switch v[0] {
	case unsignedPayload:
		return true
	case emptySHA256:
		// some broken clients set empty-sha256
		// with > 0 content-length in the body,
		// we should skip such clients and allow
		// blindly such insecure clients.

		// We return true only in situations when
		// deployment has asked MinIO to allow for
		// such broken clients, or if length is actually 0.
		return r.ContentLength == 0 || !globalCLIContext.StrictS3Compat
	}
	return false
```

klauspost (changes requested):

cmd/signature-v4-utils.go
harshavardhana: @klauspost I wrote it this way since it's intricate to show that it's more readable - I can change it to the new style as well

harshavardhana: NOTE: we really don't want to skip content-sha256 in any situation even with contentlength == 0 we want to skip only when contentlength > 0 if the incoming value is `emptySha256`

poornas (approved):
LGTM

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

PR #13388, opened 2021-10-09
fix: disallow invalid x-amz-security-token for root credentials

harshavardhana:
## Description
fix: disallow invalid x-amz-security-token for root credentials

## Motivation and Context
fixes #13335

This was a regression added in #12947 when this part of the
code was refactored to avoid privilege issues with service
accounts with session policy.

## How to test this PR?
Just do 
```
MC_HOST_local=http://minio:minio123:tk@localhost:9000/ mc ls local/
```

This shall not fail with the master branch but it should with this fix. 

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Optimization (provides speedup with no functional changes)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [x] Fixes a regression #12947
- [ ] Documentation updated
- [ ] Unit tests added/updated

PR #13412, opened 2021-10-11
fix: allow service accounts for root credentials

### LINKED ISSUE #13407: All root-owned service account static credentials fail after 2021-10-10?
I have only been running Minio since 2021-10-05, in a small single-instance container which updates to latest releases automatically. The latest release (2021-10-10) seems to have broken my setup, which involves a couple of service accounts attached to root which are used by a variety of clients including [rclone](https://rclone.org/s3/#minio) and [pottava/s3-proxy](https://hub.docker.com/r/pottava/s3-proxy/) which [appears to use](https://github.com/pottava/aws-s3-proxy/blob/master/internal/service/amazon-s3.go) the AWS SDK for Go.

It seems likely that https://github.com/minio/minio/pull/13388 is related to this issue. @klauspost helpfully explained in Slack that these changes should have only affected authentication when using credentials obtained using STS. My understanding is that my setup uses static credentials, and thus should not be subject to this change.

N.B. I don't have a complete understanding of the difference between STS and static credentials, so I'm following my intuition for some of this analysis.

## Expected Behavior
Provisioning a service account under the root user should allow a client to authenticate with that service account's access key and secret key.

## Current Behavior
Attempting to authenticate a client using a root service account's static credentials results in `403 Forbidden`.

As a workaround, if all credentials are known, service accounts can be deleted and recreated under a non-root admin user, which would appear not to be subject to the changes in the referenced PR.

My example using minio and rclone 1.55.0-linux-amd64:

1. Log in to Minio console as root, create service account with policy:

```json
{
   "Version" : "2012-10-17",
   "Statement" : [
      {
         "Effect" : "Allow",
         "Action" : [ "s3:*" ],
         "Resource" : "arn:aws:s3:::my-bucket/*"
      }
   ]
}
```

2. Configure rclone as follows:
```lang-ini
[minio]
type = s3
provider = Minio
env_auth = false
access_key_id = REDACTED
secret_access_key = REDACTED
endpoint = https://my.minio.api.example
acl = private
region = us-east-1
```

3. Attempt to PUT to the service account's allowed bucket `rclone copy my_file.txt minio:my-bucket`
4. Rclone reports
```
2021/10/11 09:05:09 ERROR : Attempt 1/3 failed with 1 errors and: Forbidden: Forbidden
	status code: 403, request id: 16ACFBB28EE016DF, host id:
```

## Context
It's mildly plausible to me that this is intentional since it is the root account after all, but if that's true then I'm not sure the change has been explained or documented effectively. When starting out with minio I would also expect to troubleshoot my own configuration issues by using the root account, so I would find it confusing if this was not possible for some reason.

## Regression
Possibly https://github.com/minio/minio/pull/13388 if this behavior is indeed incorrect.

## Your Environment
* RELEASE.2021-10-10T16-53-30Z
* Deployed on CentOS 8 Linux (`4.18.0-240.1.1.el8_3.x86_64`) on Docker 20.10.7 using docker-compose.

harshavardhana:
## Description
fix: allow service accounts for root credentials

## Motivation and Context
fixes #13407

## How to test this PR?
As per #13407 

## Types of changes

## Checklist:
- [x] Fixes a regression yes and no but valid due to #13388 
- [ ] Documentation updated

### DIFF
--- a/cmd/iam.go
+++ b/cmd/iam.go
@@ -2228,6 +2228,12 @@ func (sys *IAMSys) policyDBGet(name string, isGroup bool) (policies []string, er
 
 	mp, ok := sys.iamUserPolicyMap[name]
 	if !ok {
+		// Service accounts with root credentials, inherit parent permissions
+		if parentName == globalActiveCred.AccessKey && u.IsServiceAccount() {
+			// even if this is set, the claims present in the service
+			// accounts apply the final permissions if any.
+			return []string{"consoleAdmin"}, nil
+		}
 		if parentName != "" {
 			mp = sys.iamUserPolicyMap[parentName]
 		}


minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

donatello (approved):

PR #13496, opened 2021-10-22
fix: server in shutdown should return 503 instead of 403

harshavardhana:
## Description
fix: server in shutdown should return 503 instead of 403

## Motivation and Context
various situations where the client is retrying the request
server going through shutdown might incorrectly send 403
which is a non-retriable error, this PR allows for clients
when they retry an attempt to go to another healthy pod
or server in a distributed cluster - assuming it is a properly
load-balanced setup.

## How to test this PR?
Nothing special just have a retrying client, when the server
is going down you shall see new connection attempts are
receiving 403 which is wrong and should be changed to 503
instead.

## Types of changes

## Checklist:
- [ ] Documentation updated


klauspost (approved):
LGTM

vadmeste (approved):
LGTM

PR #13514, opened 2021-10-26
Update README.md prometheus

titansmc:
## Description
Make users aware that prometheus will add the port to the host header,

## Motivation and Context
Lack of documentation

## Checklist:
- [ x] Documentation updated


docs/metrics/prometheus/README.md
```diff
@@ -114,6 +114,8 @@ Start (or) Restart Prometheus service by running
 
 Here `prometheus.yml` is the name of configuration file. You can now see MinIO metrics in Prometheus dashboard. By default Prometheus dashboard is accessible at `http://localhost:9090`.
 
+In case you are running MiniIO nodes behind a proxy (like HAProxy), consider adjusting the acl rules to match the header, since Prometheus will always add port to the hostname like `hostname:port`.
```
harshavardhana: Can you explain what this is ?

// cc @eco-minio @ravindk89

harshavardhana (changes requested):

docs/metrics/prometheus/README.md
titansmc: so when running Minio behind a LB like HAProxy, prometheus scraper sends the header `Host: server.domain.com:port` , which might not match the rule that you have in your Load Balancer, so you might need to add another rule to match that header too. Regular requests from `mc` or web interface will have `Host: server.domain.com`

ravindk89: I think we want those explicit details / examples here if we want to document this case. Also guessing this might apply in general to any network control plane that has strict routing controls. Minor concern that this kind of documentation can lead to "tell me how to do the <thing>", which lends itself to scope creep over time (we are not an authority on network control), but this level of detail is probably OK.

Spitballing, not completely familiar with this knowledge area so might not be 100% accurate here.

"Prometheus sets the ``Host`` header to ``domain:port`` as part of HTTP operations against the MinIO metrics endpoint. For MinIO deployments behind a load balancer, reverse proxy, or other control plane (HAProxy, nginx, pfsense, opnsense, etc.), ensure the network service supports routing these requests to the deployment."

Or something along those lines. I'm inclined to not specify "ACL rules" because that might not be the only solution here, and it also begs the question of "what should my ACL look like"

titansmc: You seem to have described it better than me. So for haproxy I had to add:
```
acl is_s3 hdr(host) -i s3.domain.com
use_backend s3_ssl_back if is_s3

acl is_s3 hdr(host) -i s3.domain.com:443
use_backend s3_ssl_back if is_s3
```
So the second request would catch only the prometheus probe.

ravindk89: OK - feel free to use what i wrote or tweak slightly. I would prefer not to get into specific configs for specific products, as we cannot meaningfully keep those maintained over time.

titansmc: ok, I think now I have added the second commit which modifies the initial comment.

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

harshavardhana (approved):

ravindk89 (approved):

PR #13934, opened 2021-12-16
heal data shards with padding issues automatically

harshavardhana:
## Description
heal data shards with padding issues automatically

## Motivation and Context
data shards were wrong due to a healing bug in
reported in #13803 in some versions mainly with
M=12, N=4 and 1MiB blockSized erasure blocks.

This PR handles this situation automatically by
looking at the disk mtimes of the xl.meta, as long
as the disk mtimes are same in quorum we can try
to heal the rest of the blocks as outdated.

This should automatically fix such situations.

## How to test this PR?
As per the #13803 corrupt the blocks 

## Types of changes

## Checklist:
- [ ] Documentation updated

### DIFF
--- a/cmd/erasure-metadata.go
+++ b/cmd/erasure-metadata.go
@@ -205,10 +205,14 @@ func (fi FileInfo) TransitionInfoEquals(ofi FileInfo) bool {
 
 // MetadataEquals returns true if FileInfos Metadata maps are equal, false otherwise.
 func (fi FileInfo) MetadataEquals(ofi FileInfo) bool {
-	if len(fi.Metadata) != len(ofi.Metadata) {
-		return false
-	}
 	for k, v := range fi.Metadata {
+		if k == reservedMetadataPrefixLowerDataShardFix {
+			// Skip this since this is a special
+			// metadata that might not exist on all
+			// drives, but that is expected so skip
+			// verifying this.
+			continue
+		}
 		if ov, ok := ofi.Metadata[k]; !ok || ov != v {
 			return false
 		}
@@ -349,6 +353,17 @@ func findFileInfoInQuorum(ctx context.Context, metaArr []FileInfo, modTime time.
 	}
 
 	return FileInfo{}, errErasureReadQuorum
+}
+
+func pickValidDiskTimeWithQuorum(metaArr []FileInfo, errs []error, quorum int) time.Time {
+	diskMTimes := listObjectDiskMtimes(metaArr, errs)
+
+	diskMTime, diskMaxima := commonTimeAndOccurence(diskMTimes)
+	if diskMaxima >= quorum {
+		return diskMTime
+	}
+
+	return timeSentinel
 }
 
 // pickValidFileInfo - picks one valid FileInfo content and returns from a
--- a/cmd/erasure-healing-common.go
+++ b/cmd/erasure-healing-common.go
@@ -26,29 +26,35 @@ import (
 )
 
 // commonTime returns a maximally occurring time from a list of time.
-func commonTime(modTimes []time.Time) (modTime time.Time) {
-	timeOccurenceMap := make(map[int64]int, len(modTimes))
+func commonTimeAndOccurence(times []time.Time) (maxTime time.Time, maxima int) {
+	timeOccurenceMap := make(map[int64]int, len(times))
 	// Ignore the uuid sentinel and count the rest.
-	for _, t := range modTimes {
+	for _, t := range times {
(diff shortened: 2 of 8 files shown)


minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

harshavardhana:
@klauspost there are some other upgrade areas that are not working but - I will find out a way if it's possible to address those as well. 

Please look at this approach and see if there are any red areas that might come up.

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

harshavardhana:
this still doesn't fix the error on the drives. Still working through fixing it.

harshavardhana:
> this still doesn't fix the error on the drives. Still working through fixing it.

Okay tested it works, basically, you need a setup let's say from RELEASE.2021-11-24T23-19-33Z, and create the scenario mentioned in #13803 

Then upgrade the MinIO server with this fix, `mc admin heal -r` will fix the object the first time. 

The resolution window is '1-second' and this healing will trigger only for situations where it is possible i.e we have "quorum".  

So it is best-effort only. The protection is added to ensure that we do not "double" heal and do it only once. 

@klauspost

harshavardhana:
I will add some tests with sample data corpus tomorrow at my time to fully have tests validate the fix.

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

cmd/erasure-metadata.go
```diff
@@ -205,10 +205,14 @@ func (fi FileInfo) TransitionInfoEquals(ofi FileInfo) bool {
 
 // MetadataEquals returns true if FileInfos Metadata maps are equal, false otherwise.
 func (fi FileInfo) MetadataEquals(ofi FileInfo) bool {
-	if len(fi.Metadata) != len(ofi.Metadata) {
-		return false
-	}
 	for k, v := range fi.Metadata {
+		if k == reservedMetadataPrefixLowerDataShardFix {
```
klauspost, line 209: This is problematic, since it will affect quorum calculations for listing at least in [xlMetaV2Object signature](https://github.com/minio/minio/blob/81e43b87c2b9eb9a112e21aaaea04a202c4f4d55/cmd/xl-storage-format-v2.go#L520).

If we want to update metadata it should be done for *all* disks, otherwise we will risk running into all kinds of trouble. Not doing that is a rather high risk change.

klauspost (changes requested):
My big question is: Does this rewrite *all* shards, or just the ones that are detected?

The problem is that erasure codes are calculated with junk in the padding bytes. If we just restore the one that was rewritten the erasure codes will still be unable to reconstruct the data.

The thing is that when we heal, AFAICT we discard the junk bytes when healing. I presume we have the junk bytes, except for when a shard was healed.

If we read shards and they have non-0 bytes as padding all of them should be rewritten.

This needs more investigation AFAICT. We risk making the issue worse. I will go through the code and see what I can figure out.

klauspost:
So AFAICT we read it back as an object and has the writer spit out the shards we need.

Problem is that we still have junk in the shards at the end and erasure codes have this junk in them. 

So unless we are clever about it, we need to rewrite *all* shards (except the first), otherwise erasure codes are still broken for the last block.

harshavardhana:
> My big question is: Does this rewrite _all_ shards, or just the ones that are detected?
> 
> The problem is that erasure codes are calculated with junk in the padding bytes. If we just restore the one that was rewritten the erasure codes will still be unable to reconstruct the data.
> 
> The thing is that when we heal, AFAICT we discard the junk bytes when healing. I presume we have the junk bytes, except for when a shard was healed.

The main idea here is to pick the only entries that are only available in "quorum" for diskMTimes. 

Now, this code skips the offending drives, the assumption here is that we will know that by a heuristic that whatever matches the commonality of diskMTimes has enough quorum to read the objects.

If it can be read as it was originally written then we are 100% sure that we have good content, so we can potentially go ahead and heal the offending drives. After the upgrade with this patch.

That automatically heals all of those content automatically @klauspost 

> Problem is that we still have junk in the shards at the end and erasure codes have this junk in them.

We will not have the assumption here is that reedsolomon + healing will fix them, how else does it work after I remove the shards manually. 

It does quite well actually TBH.

harshavardhana:
Coming up with a different style

PR #14038, opened 2022-01-06
DeleteMultipleObjects: Finish even if cancelled + concurrent sets

klauspost:
## Description

* Process sets concurrently.
* Disconnect context from request.
* Insert context cancellation checks.
* `errFileNotFound` and `errFileVersionNotFound` are ok, unless creating delete markers.

## Motivation and Context

Canceled `DeleteMultipleObjects` has some interesting properties.

Since it is operating independently in bulk, per disk, it tends to leave a lot of inconsistent objects, since each disk may be at a different places when the cancel is received. This leads to a lot of "file not found" on subsequent requests, leading to quorum errors, even if the object/version is deleted.

In overall terms cancelation on bulk requests are questionable, unless you do strictly serial processing. I have done some cleanups, which should make the current processing more stable. I will send this as a PR.

However, I think the correct solution is to disable cancelation propagation and always finish the bulk delete operation, even if the client disconnects while it is going on.

## How to test this PR?

Best test is warp:

λ warp delete --duration=1s --noclear --objects=1000000
(let it finish)

λ warp delete --duration=1s
(ctrl+c after a few seconds, repeat)

## Types of changes

### DIFF
--- a/cmd/bucket-handlers.go
+++ b/cmd/bucket-handlers.go
@@ -576,6 +576,9 @@ func (api objectAPIHandlers) DeleteMultipleObjectsHandler(w http.ResponseWriter,
 		return
 	}
 
+	// Disable timeouts and cancellation
+	ctx = bgContext(ctx)
+
 	deleteList := toNames(objectsToDelete)
 	dObjects, errs := deleteObjectsFn(ctx, bucket, deleteList, ObjectOptions{
 		Versioned:        versioned,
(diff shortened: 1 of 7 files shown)


minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

harshavardhana (approved):

vadmeste (approved):
LGTM

PR #14070, opened 2022-01-10
Avoid reconnect of disk during startup sequence

yinhen:
## Description
For the first call of function connectDisks, lastCheck is zero value, this will lead to reconnect to the endpoint, it's not expected. Also reconnect to the endpoints would lead to close of the previous connect, may lead to offline of the drives with close REST Client of the endpoints; when healing objects in the dir of .minio.sys/config, may meet err: errDiskNotFound。Before the  version https://github.com/minio/minio/commit/b7c5e45fff935994791f89237a4891185c55b91f，the err: errDiskNotFound may trigger  `delete`，which will delete the policy files on the folder of `.minio.sys/config`。Also, I don't agree with that drives can come online lazily leads to errDiskNotFound in commit https://github.com/minio/minio/commit/b7c5e45fff935994791f89237a4891185c55b91f。All the drivers of the sets come online in the call of the function waitForFormatErasure。So，all the drivers come online before call of isObjectDangling。

## Motivation and Context
Fix the bug of that objects  on `.minio.sys/config`
folder deleted unexpectedly ,  due to healing unexpectedly deleting the
files as healObject due to meet errs:errDiskNotFound when call `readAllFileInfo`，because drive comes offline by unexpectedly 
 reconnect in first call connectDisks。 

## How to test this PR?
For the latest verison，start the minio，the REST Clients of each drive be close and  reconnect after newErasureSets；

## Types of changes

### DIFF
--- a/cmd/erasure-sets.go
+++ b/cmd/erasure-sets.go
@@ -106,7 +106,7 @@ func isEndpointConnectionStable(diskMap map[Endpoint]StorageAPI, endpoint Endpoi
 	if !disk.IsOnline() {
 		return false
 	}
-	if disk.LastConn().After(lastCheck) {
+	if !lastCheck.IsZero() && disk.LastConn().After(lastCheck) {
 		return false
 	}
 	return true


minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

vadmeste:
> For the first call of function connectDisks, lastCheck is zero value, this will lead to reconnect to the endpoint

I don't think this can happen. connectDisks() will always update lastConnectDisksOpTime the first time connectDisks() is called (see the code at the beginning of connectDisks() function), therefore lastCheck will only be zero once when the node is first started

yinhen:
> > For the first call of function connectDisks, lastCheck is zero value, this will lead to reconnect to the endpoint
> 
> I don't think this can happen. connectDisks() will always update lastConnectDisksOpTime the first time connectDisks() is called (see the code at the beginning of connectDisks() function), therefore lastCheck will only be zero once when the node is first started

connectDisks only called in monitorAndConnectEndpoints，when calling this，the node has already started。Also，updating lastConnectDisksOpTime at the end of the connectDisks function, not first time.

harshavardhana:
This is not the right fix. However it's okay safe measure.

The issue with diskNotFound is that object layer is initialized very early on even with len(total disks)/2

harshavardhana (approved):

harshavardhana:
Please fix the title @yinhen your PR title is wrong here.

harshavardhana:
> For the first call of function connectDisks, lastCheck is zero value, this will lead to reconnect to the endpoint, it's not expected. Also reconnect to the endpoints would lead to close of the previous connect, may lead to offline of the drives with close REST Client of the endpoints; when healing objects in the dir of .minio.sys/config, may meet err: errDiskNotFound。Before the version [b7c5e45](https://github.com/minio/minio/commit/b7c5e45fff935994791f89237a4891185c55b91f，the) err: errDiskNotFound may trigger `delete`，which will delete the policy files on the folder of `.minio.sys/config`。Also, I don't agree with that drives can come online lazily leads to errDiskNotFound in commit [b7c5e45](https://github.com/minio/minio/commit/b7c5e45fff935994791f89237a4891185c55b91f。All) the drivers of the sets come online in the call of the function waitForFormatErasure。So，all the drivers come online before call of isObjectDangling。

The drives come online lazily because we do not expect all the drives to be online from the get-go what we expect is total disks in the cluster agree on the same format.json atleast len(totalDisks)/2 - i.e `getFormatErasureInQuorum()` dictates this. 

The fix is the right fix in the commit and nowhere else - what you are adding here is an optimization to avoid reconnect() during startup which is good to add but it is not the bug that will get fixed with just this change.

@yinhen so fix the PR title to indicate the optimization that you are doing to avoid "reconnect's()" during startup sequence.

vadmeste:
This change will cause confusion to isEndpointConnectionStable(). Maybe the best way is to remove this line, 

https://github.com/minio/minio/blob/master/cmd/erasure-sets.go#L294

I am not sure why we do that in the first place.

harshavardhana:
> https://github.com/minio/minio/blob/master/cmd/erasure-sets.go#L294
> 
> I am not sure why we do that in the first place.

We do that because its possible that some disks are not yet online and we ask them to connect right away.

yinhen:
> 

Ok，got it. In my test case, all disks comes online, but during the reconnet of the disks in fist call of connectDisks，which would close the disk and  newStorageAPI with the endpoint concurently，between the time interval，another goroutines works as initAutoHeal，will heal the objects  on the folder of .minio.sys/config，then readAllFileInfo would return ErrDiskNoFound in HealObject because of offline disks. So my view about the commit https://github.com/minio/minio/commit/b7c5e45fff935994791f89237a4891185c55b91f is that, it's not only the slow network，but also the reconnet, may lead to ErrDiskNotFound.

harshavardhana:
> Ok，got it. In my test case, all disks comes online, but during the reconnet of the disks in fist call of connectDisks，which would close the disk and newStorageAPI with the endpoint concurently，between the time interval，another goroutines works as initAutoHeal，will heal the objects on the folder of .minio.sys/config，then readAllFileInfo would return ErrDiskNoFound in HealObject because of offline disks. So my view about the commit [b7c5e45](https://github.com/minio/minio/commit/b7c5e45fff935994791f89237a4891185c55b91f) is that, it's not only the slow network，but also the reconnet, may lead to ErrDiskNotFound.

Agreed that is definitely plausible here - but the fix in that commit however is what fixes the expected issue.

harshavardhana (approved):

PR #14095, opened 2022-01-12
lock: Fix decision when a lock needs to be removed

vadmeste:
## Description
The code was not properly deciding if lock needs to be removed when it
doesn't have quorum anymore. After this commit, a lock will be
forcefully unlocked if nodes reporting they are not able to find a lock
internally breaks the quorum.

Simplify the code as well.

## Motivation and Context
Sometimes a lock is not cleared when it doesn't have a quorum

## How to test this PR?
Hard to test

## Types of changes

## Checklist:
- [ ] Documentation updated

### DIFF
--- a/internal/dsync/dsync_test.go
+++ b/internal/dsync/dsync_test.go
@@ -236,10 +236,38 @@ func TestTwoSimultaneousLocksForDifferentResources(t *testing.T) {
 	time.Sleep(10 * time.Millisecond)
 }
 
-// Test refreshing lock
+// Test refreshing lock - refresh should always return true
+//
+func TestSuccessfulLockRefresh(t *testing.T) {
+	dm := NewDRWMutex(ds, "aap")
+	contextCanceled := make(chan struct{})
+
+	ctx, cl := context.WithCancel(context.Background())
+	cancel := func() {
+		cl()
+		close(contextCanceled)
+	}
+
+	if !dm.GetLock(ctx, cancel, id, source, Options{Timeout: 5 * time.Minute}) {
+		t.Fatal("GetLock() should be successful")
+	}
+
+	timer := time.NewTimer(drwMutexRefreshInterval * 2)
+
+	select {
+	case <-contextCanceled:
+		t.Fatal("Lock context canceled which is not expected")
+	case <-timer.C:
+	}
+
+	// Should be safe operation in all cases
+	dm.Unlock()
+}
+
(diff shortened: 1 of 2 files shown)


harshavardhana (changes requested):
It is better to bring a unit test for this logic @vadmeste we can avoid regressions.

internal/dsync/dsync_test.go
```diff
@@ -236,10 +236,38 @@ func TestTwoSimultaneousLocksForDifferentResources(t *testing.T) {
...
+	cancel := func() {
+		cl()
+		close(contextCanceled)
+	}
+
+	if !dm.GetLock(ctx, cancel, id, source, Options{Timeout: 5 * time.Minute}) {
+		t.Fatal("GetLock() should be successful")
+	}
+
+	timer := time.NewTimer(drwMutexRefreshInterval * 2)
```
klauspost, line 259: Does this test take 20 seconds to run?

vadmeste: > Does this test take 20 seconds to run?

This test is checking that a lock is not canceled accidently. I am giving it two rounds of Refresh before we can be sure it is is not buggy.

harshavardhana (approved):
LGTM

internal/dsync/dsync_test.go
klauspost: @vadmeste My point is that no test should take 20 seconds to run.

vadmeste: > @vadmeste My point is that no test should take 20 seconds to run.

Refresh is executed each 10 seconds and I think we don't have a choice here. I can skip it when -short flag is passed

klauspost: > I can skip it when -short flag is passed

If you don't feel like refactoring, then sure. And make it `t.Parallel`.

vadmeste: t.Parallel will need refactoring because tests share locking API and there is some weird RPC code which looks like legacy. I will send another PR with initialize a new locking API for each test so we can run then in parallel. That will reduce test duration in this package from 2 minutes to 20 seconds.

klauspost (approved):

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

PR #14143, opened 2022-01-20
fs: Return prefix as listing marker if no objects

### LINKED ISSUE #14132: Truncated Response without Continuation Token
**Context**
We have a problem regarding the S3 ListObjectV2 request.

**Current Behavior**
Sometimes we have an error in our application indicating that the response from MinIO server is truncated but there is no continuation token.

image

We checked the HTTP packets and in fact it seems that MinIO returns a truncated response without the continuous tokens and key count set to zero.
Check the picture below: In the left side it's the wrong response.

image

In this picture we can see that it doesn't always happen. Some requests return fine (right side) and few milisesond after it returns bad responses for the same S3 prefix.
When we restart the service, it works fine for a few hours and then we observe these responses again.

**Your Environment**
 Minio RELEASE.2022-01-08T03-11-54Z on Centos 7 (Kernel 3.10.0-1160.45.1.el7.x86_64) and installed in standalone mode.

klauspost:
## Description

Fixes #14132

## How to test this PR?

Requesting single keys is probably the easiest. Requires empty directories as far as I can tell.

## Types of changes

### DIFF
--- a/cmd/object-api-common.go
+++ b/cmd/object-api-common.go
@@ -164,6 +164,8 @@ func listObjectsNonSlash(ctx context.Context, bucket, prefix, marker, delimiter 
 		result.IsTruncated = true
 		if len(objInfos) > 0 {
 			result.NextMarker = objInfos[len(objInfos)-1].Name
+		} else if len(result.Prefixes) > 0 {
+			result.NextMarker = result.Prefixes[len(result.Prefixes)-1]
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

harshavardhana (approved):

PR #14229, opened 2022-02-01
Add extra environment variables from secret in Helm

Sea-you:
## Description
Add ability to use extra environment variables from a secret.

## Motivation and Context
There are obvious cases when you don't want to commit some secrets in the values.yaml file, such as LDAP password, etc. For these use cases either you could create such secrets manually or use for example the external-secrets operator. This PR aims to cover these use-cases

## How to test this PR?
Create the secret where the key name (all caps) is the same as the required env var and set the secret name in the values.yaml file as documented.

## Types of changes

## Checklist:
- [ ] Documentation updated

### DIFF
--- a/helm/minio/values.yaml
+++ b/helm/minio/values.yaml
@@ -315,6 +315,10 @@ environment:
   ## MINIO_SUBNET_LICENSE: "License key obtained from https://subnet.min.io"
   ## MINIO_BROWSER: "off"
 
+## The name of a secret in the same kubernetes namespace which contain values to be added to the environment
+## This can be useful for LDAP password, etc
+# envFromSecret: minio-extraenv
+
 networkPolicy:
   enabled: false
   allowExternal: true
--- a/helm/minio/templates/deployment.yaml
+++ b/helm/minio/templates/deployment.yaml
@@ -126,6 +126,11 @@ spec:
             - name: {{ $key }}
               value: {{ $val | quote }}
             {{- end}}
+          {{- if .Values.envFromSecret }}
+          envFrom:
+          - secretRef:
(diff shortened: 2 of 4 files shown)


minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

helm/minio/values.yaml
```diff
@@ -315,6 +315,10 @@ environment:
   ## MINIO_SUBNET_LICENSE: "License key obtained from https://subnet.min.io"
   ## MINIO_BROWSER: "off"
 
+## The name of a secret in the same kubernetes namespace which contain values to be added to the environment
+## This can be useful for LDAP password, etc
+# envFromSecret: minio-extraenv
+
```
harshavardhana, line 321: You can actually mount the entire secret directly as `config.env` all you need to set is `MINIO_CONFIG_ENV_FILE=path/to/env` - MinIO will load the entries in this and set the relevant environment values. 

Expected format is like 

```
ENV_NAME1=value
ENV_NAME2=value
```

This allows overriding all values. It is preferred to have this change over loading from `envFrom`

harshavardhana (changes requested):

helm/minio/values.yaml
Sea-you: Thanks, that would also do what I intend to do here, although I don't see from the templates how would it be possible to mount an additional secret right now. I could change this PR in order to do that if that's ok for you.

harshavardhana: yes that is better way to do it @Sea-you

Sea-you:
closing in favor of #14254

PR #14261, opened 2022-02-07
prometheus: Add number of requests arrived since last scrape

vadmeste:
## Description
Some users running MinIO claim that their system became slow. One way
to investigate is to look at this prometheus history of the number of
the requests reaching the server. The existing current S3 requests metric
is not enough because it can increase of the system really becomes slow, due
to disk issues for example.

## Motivation and Context
Give better visibility of the number of S3 requests arriving to the server

## How to test this PR?
`curl http://localhost:9004/minio/v2/metrics/node  | grep minio_s3_requests_reached_total`

## Types of changes

## Checklist:
- [ ] Documentation updated

### DIFF
--- a/cmd/admin-handlers.go
+++ b/cmd/admin-handlers.go
@@ -284,6 +284,7 @@ type ServerHTTPAPIStats struct {
 // including their average execution time.
 type ServerHTTPStats struct {
 	S3RequestsInQueue      int32              `json:"s3RequestsInQueue"`
+	S3RequestsIncoming     uint64             `json:"s3RequestsIncoming"`
 	CurrentS3Requests      ServerHTTPAPIStats `json:"currentS3Requests"`
 	TotalS3Requests        ServerHTTPAPIStats `json:"totalS3Requests"`
 	TotalS3Errors          ServerHTTPAPIStats `json:"totalS3Errors"`
--- a/cmd/http-stats.go
+++ b/cmd/http-stats.go
@@ -143,6 +143,7 @@ type HTTPStats struct {
 	rejectedRequestsHeader  uint64
 	rejectedRequestsInvalid uint64
 	s3RequestsInQueue       int32
+	s3RequestsIncoming      uint64
 	currentS3Requests       HTTPAPIStats
 	totalS3Requests         HTTPAPIStats
 	totalS3Errors           HTTPAPIStats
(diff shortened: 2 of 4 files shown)


klauspost:
Wouldn't `currentS3Requests` show the metrics you are looking for? Ie. running requests.

cmd/admin-handlers.go
```diff
@@ -284,6 +284,7 @@ type ServerHTTPAPIStats struct {
 // including their average execution time.
 type ServerHTTPStats struct {
 	S3RequestsInQueue      int32              `json:"s3RequestsInQueue"`
+	ReachedS3Requests      uint64             `json:"reachedS3Requests"`
```
harshavardhana: This should be called incoming @vadmeste and has to be changed in this manner elsewhere

harshavardhana (changes requested):

vadmeste:
> Wouldn't currentS3Requests show the metrics you are looking for? Ie. running requests.

No, current S3 requests is the number of the requests currently processed by the server. If the server becomes slow for some reasons, then this metric will increase because each request will take more time to be processed and we won't know if the server is slow because there is more S3 requests or there is another problem. That's why I am adding a new metric here.

cmd/http-stats.go
```diff
@@ -143,6 +143,7 @@ type HTTPStats struct {
 	rejectedRequestsHeader  uint64
 	rejectedRequestsInvalid uint64
 	s3RequestsInQueue       int32
+	incomingS3Requests      uint64
```
harshavardhana:
```suggestion
	s3RequestsInComing      uint64
```

harshavardhana: With same format style as `s3RequestsInQueue`

harshavardhana (changes requested):

cmd/http-stats.go
vadmeste: > s3RequestsInComing

I did this, s3RequestsIncoming

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

harshavardhana (approved):

Issue #10265, opened 2020-08-14
Minio does not support ExistingObjectTag policies


braunsonm:
Object tagging is common practice and AWS S3 supports the key `ExistingObjectTag` in policies since version `2012-10-17` but it seems unsupported in Minio.

## Expected Behavior
The following policy should work:
```
                    {
                        "Effect": "Deny",
                        "Action": "s3:GetObject",
                        "Resource": "arn:aws:s3:::mybucket/*",
                        "Condition": {
                            "StringNotEquals": {
                                "s3:ExistingObjectTag/status": "DONE"
                            }
                        }
                    }
```

## Current Behavior
Minio will fail to parse the policy.

## Possible Solution
Add support for the missing condition documented here: https://docs.aws.amazon.com/AmazonS3/latest/dev/object-tagging.html

## Steps to Reproduce (for bugs)

1. Create a bucket
2. Add the above policy
3. Notice the error that the condition is not supported and cannot be parsed.

## Context
Having some post processing on files that tag them when complete and blocking downloads of those files until the post processing is done is not possible without this condition.
This is also common practice with bucket antivirus solutions: https://github.com/upsidetravel/bucket-antivirus-function

## Your Environment
* Version used (`minio --version`): Any
* Server setup and configuration: Any
* Operating System and version (`uname -a`): Any

harshavardhana:
We do not yet support object tag conditionals..

braunsonm:
I'm happy to aid in testing if you're able to work on it @harshavardhana

braunsonm:
In my opinion this should not be automatically closed. This is an important feature with a lot of implications in S3.

harshavardhana:
> In my opinion this should not be automatically closed. This is an important feature with a lot of implications in S3.

Will add `do not close` label, But we have no cycles right now to work on this @braunsonm so if you have time please send us a fix.

bpedersen2:
Seems like https://github.com/minio/pkg/pull/4 did only add the names, but no usage of the tags?

MartinKavik:
Could I ask why the issue is closed, please? 
`ExistingObjectTag` isn't supported on the latest Minio (`minio version RELEASE.2021-12-10T23-03-39Z`) and the issue has been closed many month ago so I assume there isn't a PR waiting for merging.
Thank you.

harshavardhana:
This will be re-opened as the server component part is not added yet and will be added in future.

ftravagliaocto:
Hello, 

[...]

PR #14333, opened 2022-02-17
Using tag condition on get object

ftravaglia:
## Description
Implementing the use of "s3:ExistingObjectTag" policy conditions on GetObject

## Motivation and Context
Within a bucket, we want to give the possibility to give access or not on an object in reading for a given user. 

For example: If one of the files has the tag 'type:sensitive' and in the iam policies file the policy condition is set to Allow "s3:ExistingObjectTag/type": "sensitive" then the user will not be able to read the file nor download it (it remains visible in the bucket but only at the name level)

Translated with www.DeepL.com/Translator (free version)

## How to test this PR?
- deploy minio
- add policy 'testtagpolicy' 
`{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "s3:*"
            ],
            "Resource": [
                "arn:aws:s3:::*"
            ]
        },
        {
            "Effect": "Deny",
            "Action": [
                "s3:GetObject"
            ],
            "Resource": [
                "arn:aws:s3:::mybucket/*"
            ],
            "Condition": {
                "StringEquals": {
                    "s3:ExistingObjectTag/type": [
                        "sensitive"
                    ]
                }
            }
        }
    ]
}`
- as admin create user 'test'
- as admin link the user 'test' with the 'testtagpolicy'
- as admin create bucket 'mybucket'
- as admin go into the bucket and push a file
- as admin add tag to the file with key 'type' and value 'sensitive'
- as admin add another file without tag
- as test user login to minio
- as test go browse 'mybucket'
- as test try to download the file with tags
- as test try to download the file without tags

Result expected : 
The file with tags can't be downloaded (on firefox it download a file but with 0 bytes)
The file without tags can be downloaded

- as admin remove tag to the file with tags

Result expected : 
The file with tags removed can be downloaded or see into the minio console view

- as admin add tag to the file with key 'type' and value 'sensitive'
The file with tags can't be downloaded (on firefox it download a file but with 0 bytes)

## Types of changes

## Checklist:
- [ ] Documentation updated

See [https://github.com/minio/minio/issues/10265](https://github.com/minio/minio/issues/10265)

### DIFF
--- a/cmd/object-handlers.go
+++ b/cmd/object-handlers.go
@@ -600,8 +600,13 @@ func (api objectAPIHandlers) headObjectHandler(ctx context.Context, objectAPI Ob
 		writeErrorResponseHeadersOnly(w, toAPIError(ctx, err))
 		return
 	}
-
-	if s3Error := checkRequestAuthType(ctx, r, policy.GetObjectAction, bucket, object); s3Error != ErrNone {
+	objAPI := api.ObjectAPI()
+	tags, err := objAPI.GetObjectTags(ctx, bucket, object, opts)
+	if err != nil {
+		writeErrorResponseHeadersOnly(w, toAPIError(ctx, err))
+		return
+	}
+	if s3Error := checkRequestAuthTypeAndTagCondition(ctx, r, policy.GetObjectAction, bucket, object, tags); s3Error != ErrNone {
 		if getRequestAuthType(r) == authTypeAnonymous {
 			// As per "Permission" section in
 			// https://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectHEAD.html
--- a/cmd/auth-handler.go
+++ b/cmd/auth-handler.go
@@ -33,6 +33,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/minio/minio-go/v7/pkg/tags"
 	"github.com/minio/minio/internal/auth"
 	objectlock "github.com/minio/minio/internal/bucket/object/lock"
 	"github.com/minio/minio/internal/etag"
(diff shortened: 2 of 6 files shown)


cmd/object-handlers.go
```diff
@@ -600,8 +600,13 @@ func (api objectAPIHandlers) headObjectHandler(ctx context.Context, objectAPI Ob
 		writeErrorResponseHeadersOnly(w, toAPIError(ctx, err))
 		return
 	}
-
-	if s3Error := checkRequestAuthType(ctx, r, policy.GetObjectAction, bucket, object); s3Error != ErrNone {
+	objAPI := api.ObjectAPI()
+	tags, err := objAPI.GetObjectTags(ctx, bucket, object, opts)
```
klauspost, line 604: Fetching tags, just to fetch object info following is effectively doubling the operations needed.

Fetch the information once.

cmd/object-handlers.go
```diff
@@ -600,8 +600,13 @@ func (api objectAPIHandlers) headObjectHandler(ctx context.Context, objectAPI Ob
...
+	if err != nil {
+		writeErrorResponseHeadersOnly(w, toAPIError(ctx, err))
```
klauspost, line 606: This changes the order of errors and likely changes the errors returned when multiple problems exist.

In this case it seems that we "leak" whether the object exists, before we have check auth. This needs careful implementation.

See also `// As per "Permission" section in....` below.

klauspost (commented):
Thanks for the contribution. Some initial observations on your draft.

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

harshavardhana:
This PR is an incorrect way to implement this @ftravaglia

ftravaglia:
> This PR is an incorrect way to implement this @ftravaglia

@harshavardhana Okay, so can you tell me the correct way to correct this, please? Thanks

harshavardhana:
> > This PR is an incorrect way to implement this @ftravaglia
> 
> @harshavardhana Okay, so can you tell me the correct way to correct this, please? Thanks

First of all 

- Attempting to read tags is wrong at the handler layer this was on purpose not implemented, we cannot loose IOPs the way you have implemented.

- The clean way to implement this to implement this as a callback, i.e tagging based filtering has to be applied in the same manner as Get and Head based pre-conditions are applied for Range, Etag(match).. are implemented.

- No new package dependencies are entertained, such as another ftravaglia/pkg 

If you can make all these deeper changes you are free to attempt this change.

But honestly this is a more involved change than you have led yourself on, so my suggestion would be simply close this PR and wait for us to implement it cleanly.

harshavardhana:
Closed this since this is not a correct change, please re-submit when the mentioned comments are addressed or wait for us to fix it.

PR #14401, opened 2022-02-24
fix: possible wrong path composition for keycloak user lookups

hellivan:
## Description
`LookupUser` method in `internal/config/identity/openid/provider/keycloak.go` misses out a separation slash in URL path composition.

## Motivation and Context
In case that a user defined a value for the `MINIO_IDENTITY_OPENID_KEYCLOAK_REALM` environment
variable, but did not prefixed the value with a slash, the API path for the keycloak user
lookups was not composited correctly. This was also the case if the user did not specify any realm at all since Minio
used realm `master` as default in such cases

## How to test this PR?
For now I did not implement any unit tests since I'm not 100% sure if implementation satisfies your general implementation/coding strategy in Minio. Please let me know if implementation approach is ok for you
I can implement the missing unit tests.

## Types of changes

## Checklist:
- [ ] Documentation updated

### DIFF
--- a/internal/config/identity/openid/provider/keycloak.go
+++ b/internal/config/identity/openid/provider/keycloak.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"path"
 	"strings"
 	"sync"
 )
@@ -83,11 +84,12 @@ func (k *KeycloakProvider) LoginWithClientID(clientID, clientSecret string) erro
 
 // LookupUser lookup user by their userid.
 func (k *KeycloakProvider) LookupUser(userid string) (User, error) {
-	lookupUserID := k.adminURL + "/realms" + k.realm + "/users/" + userid
-	req, err := http.NewRequest(http.MethodGet, lookupUserID, nil)
+	req, err := http.NewRequest(http.MethodGet, k.adminURL, nil)
 	if err != nil {
 		return User{}, err
 	}
+	req.URL.Path = path.Join(req.URL.Path, "realms", k.realm, "users", userid)
+
 	k.Lock()
 	accessToken := k.accessToken
 	k.Unlock()


internal/config/identity/openid/provider/keycloak.go
```diff
@@ -81,9 +81,17 @@ func (k *KeycloakProvider) LoginWithClientID(clientID, clientSecret string) erro
...
+	for _, part := range parts {
+		res = fmt.Sprintf("%s/%s", strings.TrimRight(res, "/"), strings.TrimLeft(part, "/"))
+	}
+	return res
+}
+
 // LookupUser lookup user by their userid.
 func (k *KeycloakProvider) LookupUser(userid string) (User, error) {
-	lookupUserID := k.adminURL + "/realms" + k.realm + "/users/" + userid
+	lookupUserID := joinURLPaths(k.adminURL, "realms", k.realm, "users", userid)
```
harshavardhana: This should be rewritten like this instead, joinURLPaths is a complicated piece that is not necessary.

```
k.adminURL + "/" + path.Join("realms", k.realm, "users", userid)
```

harshavardhana (changes requested):

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

internal/config/identity/openid/provider/keycloak.go
hellivan: I understand.... I implemented the helper function since I was not sure if the content of `adminURL` is validated before so that it does not contain a trailing slash. From my understanding (please correct me if I'm wrong), `adminURL` contains the value a user provided via `MINIO_IDENTITY_OPENID_KEYCLOAK_ADMIN_URL` env variable. If that value contains a trailing slash, the composed URL is messed up again. This is why I thought that we could use this little helper function to compose the whole URL (`path.Join` only seams to work for paths and not for fully qualified urls).

harshavardhana: That really doesn't matter if there are double `//` @hellivan since the HTTP calls filter them out.

harshavardhana: just change it to `path.Join()` joinURLPaths is a complicated piece that is not safe to use.

harshavardhana (changes requested):
```diff
git diff
diff --git a/internal/config/identity/openid/provider/keycloak.go b/internal/config/identity/openid/provider/keycloak.go
index 9b44659c0..6b11d7e1b 100644
--- a/internal/config/identity/openid/provider/keycloak.go
+++ b/internal/config/identity/openid/provider/keycloak.go
@@ -23,6 +23,7 @@ import (
        "fmt"
        "net/http"
        "net/url"
+       "path"
        "strings"
        "sync"
 )
@@ -83,11 +84,12 @@ func (k *KeycloakProvider) LoginWithClientID(clientID, clientSecret string) erro
 
 // LookupUser lookup user by their userid.
 func (k *KeycloakProvider) LookupUser(userid string) (User, error) {
-       lookupUserID := k.adminURL + "/realms" + k.realm + "/users/" + userid
-       req, err := http.NewRequest(http.MethodGet, lookupUserID, nil)
+       req, err := http.NewRequest(http.MethodGet, k.adminURL, nil)
        if err != nil {
                return User{}, err
        }
+       req.URL.Path = path.Join(req.URL.Path, "realms", k.realm, "users", userid)
+
        k.Lock()
        accessToken := k.accessToken
        k.Unlock()
```

Actually I realized a more simpler change and safer is this

hellivan:
@harshavardhana Thanks for the input. That would have been my second option for implementing this PR.
I changed the PR accordingly. However I was not sure if its a good idea to change the `URL.Path` of the created request (see https://github.com/golang/go/issues/18952) which is why I created a dedicated URL struct instead of changing it on the request.

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

internal/config/identity/openid/provider/keycloak.go
```diff
@@ -83,11 +84,17 @@ func (k *KeycloakProvider) LoginWithClientID(clientID, clientSecret string) erro
 
 // LookupUser lookup user by their userid.
 func (k *KeycloakProvider) LookupUser(userid string) (User, error) {
-	lookupUserID := k.adminURL + "/realms" + k.realm + "/users/" + userid
-	req, err := http.NewRequest(http.MethodGet, lookupUserID, nil)
+	userURL, err := url.Parse(k.adminURL)
 	if err != nil {
 		return User{}, err
 	}
+	userURL.Path = path.Join(userURL.Path, "realms", k.realm, "users", userid)
```
harshavardhana: @hellivan this is done automatically by http.NewRequest() not sure why we need to do this again and again NOTE: we are leaking memory here doing url.Parse() again and again. 

http.NewRequest() parses the input URL again.

harshavardhana (changes requested):

harshavardhana:
> I changed the PR accordingly. However I was not sure if its a good idea to change the `URL.Path` of the created request (see [golang/go#18952](https://github.com/golang/go/issues/18952)) which is why I created a dedicated URL struct instead of changing it on the request.

This is legal - for client, the issue you referenced talks about server not client.

hellivan:
@harshavardhana ok, I updated the PR.

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

harshavardhana (approved):

PR #14427, opened 2022-02-28
Return data for disk info when the disk is unformatted

vadmeste:
## Description
A DiskInfo REST call to an unformatted disk returns an error with no
disk information, such as the disk endpoint URL, which is unexpected.

## Motivation and Context
Fix wrong reporting of heal status when there is an unformatted disk (while the cluster is still online)

## How to test this PR?
- Start a distributed setup
- Manually remove one disk content and quickly run `mc admin heal <alias>`

## Types of changes

## Checklist:
- [ ] Documentation updated

### DIFF
--- a/cmd/storage-rest-server.go
+++ b/cmd/storage-rest-server.go
@@ -109,7 +109,7 @@ func storageServerRequestValidate(r *http.Request) error {
 }
 
 // IsValid - To authenticate and verify the time difference.
-func (s *storageRESTServer) IsValid(w http.ResponseWriter, r *http.Request) bool {
+func (s *storageRESTServer) IsAuthValid(w http.ResponseWriter, r *http.Request) bool {
 	if s.storage == nil {
 		s.writeErrorResponse(w, errDiskNotFound)
 		return false
@@ -117,6 +117,15 @@ func (s *storageRESTServer) IsValid(w http.ResponseWriter, r *http.Request) bool
 
 	if err := storageServerRequestValidate(r); err != nil {
 		s.writeErrorResponse(w, err)
+		return false
+	}
+
+	return true
+}
+
+// IsValid - To authenticate and check if the disk-id in the request corresponds to the underlying disk.
+func (s *storageRESTServer) IsValid(w http.ResponseWriter, r *http.Request) bool {
+	if !s.IsAuthValid(w, r) {
 		return false
 	}
 
@@ -155,7 +164,7 @@ func (s *storageRESTServer) HealthHandler(w http.ResponseWriter, r *http.Request
 
 // DiskInfoHandler - returns disk info.
 func (s *storageRESTServer) DiskInfoHandler(w http.ResponseWriter, r *http.Request) {
-	if !s.IsValid(w, r) {
+	if !s.IsAuthValid(w, r) {
 		return
 	}
 	info, err := s.storage.DiskInfo(r.Context())


minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

harshavardhana:
I am not sure what is it that fixes here @vadmeste there are some double condition checks that are not correct.

isValid should be doing isStale as well using it separately is not correct. Can you explain why are you doing these changes?

vadmeste:
@harshavardhana basically DiskInfo() for an unformatted disk doesn't work in a distributed setup. 

DiskInfo() is currently little weird because it returns an error when the disk is unformatted, though I don't think this is necessary. If it is okay to break the storage API then I can add a new field in DiskInfo result structure called 'IsFormatted'

harshavardhana:
> DiskInfo() is currently little weird because it returns an error when the disk is unformatted, though I don't think this is necessary. If it is okay to break the storage API then I can add a new field in DiskInfo result structure called 'IsFormatted'

DiskInfo can return "errUnformatted" while the DiskInfo value is valid @vadmeste this way the top-level code can interpret it properly.

vadmeste:
> DiskInfo can return "errUnformatted" while the DiskInfo value is valid @vadmeste this way the top-level code can interpret it properly.

yes, so this is PR to allow the same behavior in a distributed setup. The issue is that storage server doesn't return any information when the disk is unformatted

harshavardhana:
> yes, so this is PR to allow the same behavior in a distributed setup. The issue is that storage server doesn't return any information when the disk is unformatted

Well, IT shouldn't @vadmeste only some value should be returned that allows for making decisions. The drive should be allowed for "I/O" only when it is not in the "unformatted" stage.

The code you are adding here can get very confusing and how to use it cleanly when adding new storage APIs. That's why DiskInfo() call should be special which allows for all I/O to come in.

vadmeste:
> The code you are adding here can get very confusing and how to use it cleanly when adding new storage APIs. That's why DiskInfo() call should be special which allows for all I/O to come in.

yes, this is what the PR is doing

harshavardhana:
> yes, this is what the PR is doing

But it is making way too many changes @vadmeste I don't think we can afford regressions in this area. My suggestion is to remove IsValid() from DiskInfo() call - the rest of the code should not even be touched.

All we need is an IsValid function just to be used with DiskInfo call that validates the authorization header that's all. 

The kind of code in this PR such as 

```
if s.IsValid() and isStale()
```

IT is harder to document why we need multiple functions.

```diff
git diff
diff --git a/cmd/storage-rest-server.go b/cmd/storage-rest-server.go
index 58c19f408..6f68c3565 100644
--- a/cmd/storage-rest-server.go
+++ b/cmd/storage-rest-server.go
@@ -108,8 +108,7 @@ func storageServerRequestValidate(r *http.Request) error {
        return nil
 }
 
-// IsValid - To authenticate and verify the time difference.
-func (s *storageRESTServer) IsValid(w http.ResponseWriter, r *http.Request) bool {
+func (s *storageRESTServer) IsValidAuth(w http.ResponseWriter, r *http.Request) bool {
        if s.storage == nil {
                s.writeErrorResponse(w, errDiskNotFound)
                return false
@@ -125,6 +124,15 @@ func (s *storageRESTServer) IsValid(w http.ResponseWriter, r *http.Request) bool
                return false
        }
 
+       return true
+}
+
+// IsValid - To authenticate and verify the time difference.
+func (s *storageRESTServer) IsValid(w http.ResponseWriter, r *http.Request) bool {
+       if !s.IsValidAuth(w, r) {
+               return false
+       }
+
        diskID := r.Form.Get(storageRESTDiskID)
        if diskID == "" {
                // Request sent empty disk-id, we allow the request
@@ -155,7 +163,7 @@ func (s *storageRESTServer) HealthHandler(w http.ResponseWriter, r *http.Request
 
 // DiskInfoHandler - returns disk info.
 func (s *storageRESTServer) DiskInfoHandler(w http.ResponseWriter, r *http.Request) {
-       if !s.IsValid(w, r) {
+       if !s.IsValidAuth(w, r) {
                return
        }
        info, err := s.storage.DiskInfo(r.Context())
```

AFAICS all we need is only this many lines of change.

harshavardhana:
Since the concern is DiskInfo() doesn't return a value the handler to be fixed is DiskInfo() alone not other handlers, the PR here does the inverse by changing the "concerns" of other handlers to satisfy DiskInfo() call.

harshavardhana (approved):

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

poornas (approved):

PR #14429, opened 2022-02-28
Enhance the error checking for reset by peer, use errors.Is to check error in cmd/os-readdir_other.go

xieyuschen:
## Description
- Check `connection reset by peer` by `strings.Contains` is not elegant  
The original code to check `connection reset by peer` is this:
```go
func IsConnResetErr(err error) bool {
	if strings.Contains(err.Error(), "connection reset by peer") {
		return true
	}
	// incase if error message is wrapped.
	return errors.Is(err, syscall.ECONNRESET)
}
```
It's not so elegant and is a question asked in [stackoverflow](https://stackoverflow.com/questions/19929386/handling-connection-reset-errors-in-go)
So we can check it better with the help of type asset.  I upgraded it as this:
```go
// IsConnResetErr - Checks for connection reset errors.
func IsConnResetErr(err error) bool {
	operr, ok := err.(*net.OpError)
	if !ok {
		return false
	}
	return errors.Is(operr, syscall.ECONNRESET)
}
```
- As golang 1.13 introduced `errors.Is` to check error, so I think we can use it instead of `==` to make a better error judgement.

## Motivation and Context
N/A(sorry for unknowing how to write down here)

## How to test this PR?
Just fix the error checking way and don't affect the business.

## Types of changes

## Checklist:
- [ ] Documentation updated

### DIFF
--- a/cmd/os-readdir_other.go
+++ b/cmd/os-readdir_other.go
@@ -21,6 +21,7 @@
 package cmd
 
 import (
+	"errors"
 	"io"
 	"os"
 	"syscall"
@@ -49,11 +50,11 @@ func readDirFn(dirPath string, filter func(name string, typ os.FileMode) error) 
 		// Read up to max number of entries.
 		fis, err := d.Readdir(maxEntries)
 		if err != nil {
-			if err == io.EOF {
+			if errors.Is(err, io.EOF) {
 				break
 			}
 			err = osErrToFileErr(err)
-			if err == errFileNotFound {
+			if errors.Is(err, errFileNotFound) {
 				return nil
 			}
 			return err
@@ -77,7 +78,7 @@ func readDirFn(dirPath string, filter func(name string, typ os.FileMode) error) 
 					continue
 				}
 			}
-			if err = filter(fi.Name(), fi.Mode()); err == errDoneForNow {
+			if err = filter(fi.Name(), fi.Mode()); errors.Is(err, errDoneForNow) {
 				// filtering requested to return by caller.
 				return nil
 			}
@@ -106,7 +107,7 @@ func readDirWithOpts(dirPath string, opts readDirOpts) (entries []string, err er
 		// Read up to max number of entries.
 		fis, err := d.Readdir(maxEntries)
 		if err != nil {
-			if err == io.EOF {
+			if errors.Is(err, io.EOF) {
 				break
 			}
 			return nil, osErrToFileErr(err)
--- a/internal/event/target/store.go
+++ b/internal/event/target/store.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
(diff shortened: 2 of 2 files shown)


minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

cmd/os-readdir_other.go
```diff
@@ -49,11 +50,11 @@ func readDirFn(dirPath string, filter func(name string, typ os.FileMode) error)
 		// Read up to max number of entries.
 		fis, err := d.Readdir(maxEntries)
 		if err != nil {
-			if err == io.EOF {
+			if errors.Is(err, io.EOF) {
 				break
```
harshavardhana, line 54: Let's avoid orthogonal changes not pertaining to the title of the PR, if not mention why are you doing these changes

harshavardhana (changes requested):

cmd/os-readdir_other.go
xieyuschen: Ok, it's my carelessness, sorry for that. I already fixed it.

Also, a question is what are `orthogonal changes`? I googled it but found no relevant topic about it.

xieyuschen: I update the PR title to make the PR title contains my changes here.

harshavardhana: can you explain what error are you trying to fix here?

xieyuschen: > can you explain what error are you trying to fix here?

I think it's better to use `errors.Is` to judge error. So I change the code from `err == io.EOF` to `errors.Is(err, io.EOF)`

harshavardhana: That's a big cleanup that we have to do in multiple other places, changing code like this has potential for regressions and we cannot accept a PR without a specific issue that you are trying to fix. 

Sorry, the PR cannot be accepted in the way it is today.

xieyuschen: Ok, thanks for your feedback. So how about creating another PR to enhance the error checking in function `IsConnResetErr`?

harshavardhana: yes but why do you need to do that? @xieyuschen are you really facing any issue?

xieyuschen: > 

hmm, I don't face any issue with it. I just think that the error checking could be better if it uses an elegant way when I read the code about checking `connection reset by peer`. 
I will close this PR and create another one. Thanks for your review of my PR.

xieyuschen:
As using `errors.Is` way to check error instead of `==` is a big cleanup that we have to do in multiple other places, changing code like this has potential for regressions and we cannot accept a PR without a specific issue that you are trying to fix.

PR #13368, opened 2021-10-06
MakeBucket: Delete leftover buckets on error

klauspost:
## Description

In `(erasureServerPools).MakeBucketWithLocation` delete the created buckets if any pool returns an error.

Add `NoRecreate` option, which will not recreate the bucket in `DeleteBucket`, if the operation fails.

Use `context.Background()` for operations we want to succeed even if called canceled.

## Motivation and Context

Buckets could be left on one or more server pools.

## How to test this PR?

Create buckets in a faulty environment.

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)

PR #14452, opened 2022-03-03
do not attempt force delete on bucket

### LINKED ISSUE #14445: CreateBucket during server init causes bucket deletion (causing huge data losses) due to race condition at server initialization
I was able to finally **reproduce** the problem I've been having for the past weeks which caused the **deletion of multiple buckets on every drive and replica**.
During the server initialization it is possible to create buckets even if MinIO hasn't yet figured out which buckets already exist.

I've got a replicated minio with 4 nodes and this is how I reproduced the issue:

**Step 1**
Shutdown the MinIO instance(s).

**Step 2**
Run the following script, which sends bucket creation HTTP requests to all the MinIO replicas:

```
#!/bin/bash

while true 
do
    for i in $(seq 1 4)
    do
        curl -k "http://minio$i:9001/api/v1/buckets" \
          -H "authority: minio$i:9001" \
          -H 'sec-ch-ua: " Not A;Brand";v="99", "Chromium";v="98", "Google Chrome";v="98"' \
          -H 'sec-ch-ua-mobile: ?0' \
          -H 'user-agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/98.0.4758.102 Safari/537.36' \
          -H 'sec-ch-ua-platform: "Linux"' \
          -H 'content-type: application/json' \
          -H 'accept: */*' \
          -H "origin: http://minio$i:9001" \
          -H 'sec-fetch-site: same-origin' \
          -H 'sec-fetch-mode: cors' \
          -H 'sec-fetch-dest: empty' \
          -H "referer: http://minio$i:9001/add-bucket" \
          -H 'accept-language: en-US,en;q=0.9' \
          -H 'cookie: minio_cookie=xxxxxxxxxxxxxxxx; token=xxxxxxxxxxxxxxxxxxx' \
          --data-raw '{"name":"bucketnamehere","versioning":false,"locking":false}' \
          --compressed &

        sleep 0.01
        echo $i
    done
done
```

**Step 3**
Start the instances again (while keeping the script running in background)

---

After all of these steps I was able to **recreate** a bucket that existed already, and as a consequence, few minutes later that bucket was empty as it was been recreated even if it shouldn't have been, both on the filesystem and on MinIO's UI.

Of course my program does not run that script, but it does something similar by creating buckets at server's startup.

Also this problem shouldn't be dependent on the number of available nodes and/or drives as it looks like a problem directly correlated with the ObjectAPI object: the function initServer is called before the buckets metadata are fully loaded, and the HTTP API is already available to be used during this "unready" period.

I've reproduced it on two versions (output of "minio --version"): the latest one and one dated 27 December 2021.

harshavardhana:
## Description
do not attempt force delete on the bucket

## Motivation and Context
the caller needs to ask explicitly for force delete
otherwise, the force delete might end up deleting
an existing bucket with data.

fixes #14445

## How to test this PR?
As per #14445 - in few attempts, you can reproduce 
this issue. 

## Types of changes

## Checklist:
- [x] Fixes a regression yes introduced in PR #13368
- [ ] Documentation updated

### DIFF
--- a/cmd/erasure-server-pool.go
+++ b/cmd/erasure-server-pool.go
@@ -671,7 +671,9 @@ func (z *erasureServerPools) MakeBucketWithLocation(ctx context.Context, bucket 
 		if err != nil {
 			if _, ok := err.(BucketExists); !ok {
 				// Delete created buckets, ignoring errors.
-				z.DeleteBucket(context.Background(), bucket, DeleteBucketOptions{Force: true, NoRecreate: true})
+				z.DeleteBucket(context.Background(), bucket, DeleteBucketOptions{
+					Force: false, NoRecreate: true,
+				})
 			}
 			return err
 		}


poornas (approved):

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

harshavardhana:
Taking this in since this is a self-evident fix, please let me know otherwise.

PR #14606, opened 2022-03-22
ListObjects: Filter lifecycle expired objects

klauspost:
## Description

For ListObjects and ListObjectsV2 perform lifecycle checks on all objects before returning.

This will filter out objects that are pending lifecycle expiration.

Bonus: Cheaper server pool conflict resolution by not converting to FileInfo.

## How to test this PR?

Required expired objects that have not been cleaned. List them.

## Types of changes

### DIFF
--- a/cmd/metacache-server-pool.go
+++ b/cmd/metacache-server-pool.go
@@ -354,9 +354,11 @@ func (z *erasureServerPools) listMerged(ctx context.Context, o listPathOptions, 
 	return nil
 }
 
-// filterLifeCycle will filter out objects if the most recent version should be deleted by lifecycle.
+// filterLifeCycle will filter out objects if the most recent
+// version should be deleted by lifecycle.
 // out will be closed when there are no more results.
-// When 'in' is closed or the context is canceled the function closes out and exits.
+// When 'in' is closed or the context is canceled the
+// function closes 'out' and exits.
 func filterLifeCycle(ctx context.Context, bucket string, lc *lifecycle.Lifecycle, in <-chan metaCacheEntry, out chan<- metaCacheEntry) {
 	defer close(out)
 	for {
--- a/cmd/metacache-server-pool.go
+++ b/cmd/metacache-server-pool.go
@@ -355,8 +355,22 @@ func (z *erasureServerPools) listMerged(ctx context.Context, o listPathOptions, 
 }
 
 // filterLifeCycle will filter out objects if the most recent version should be deleted by lifecycle.
+// out will be closed when there are no more results.
+// When 'in' is closed or the context is canceled the function closes out and exits.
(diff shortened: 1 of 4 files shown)


minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

cmd/metacache-server-pool.go
```diff
@@ -346,6 +354,24 @@ func (z *erasureServerPools) listMerged(ctx context.Context, o listPathOptions,
...
+			continue
+		}
+		objInfo := fi.ToObjectInfo(bucket, obj.name)
+		action := evalActionFromLifecycle(ctx, *lc, objInfo, false)
+		switch action {
+		case lifecycle.DeleteVersionAction, lifecycle.DeleteAction:
+			// Skip this entry.
+			continue
+		}
+		out <- obj
```
harshavardhana: do we not want to do 

```
select {
case <-ctx.Done():
..
```

harshavardhana (changes requested):

cmd/metacache-server-pool.go
klauspost: Could be - double checking call path.

klauspost: @harshavardhana Added for both blocking operations 👍🏼

harshavardhana (approved):

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

WolfspiritM:
@klauspost I've a question about this change. Currently in our case with millions of files the lifecyclepolicy is way too slow and doesn't delete the files in time (even with a faster scanner). We have our files structured with the date in the path/prefix so we made a script that lists all old files and deletes them manually while still having the lifecyclepolicy available.

That script wouldn't work anymore as it won't be able to receive any files after this change, right?
Will the files be deleted right when listed or are they just filtered out and the scanner still needs to reach them?

Cause that would mean we have to disable the lifecyclepolicy completly now.

klauspost:
@WolfspiritM You can list versions. These are not filtered, as they are also needed for deleting all objects in a bucket, so that is not too likely to change.

WolfspiritM:
@klauspost We're not using versioning in that bucket to not create even more metadata for our millions of small files. Does listing versions work for a non versioned bucket? A quick test return 404 Not Found but I'm not sure if the reason is that Versioning is disabled.

harshavardhana:
> @klauspost We're not using versioning in that bucket to not create even more metadata for our millions of small files. Does listing versions work for a non versioned bucket? A quick test return 404 Not Found but I'm not sure if the reason is that Versioning is disabled.

Even without using versioning the API will behave as explained by @klauspost

PR #14631, opened 2022-03-26
etag: fix incorrect multipart detection

aead:
## Description
This commit fixes a subtle bug in the ETag
`IsEncrypted` implementation.

An encrypted ETag may contain random bytes,
i.e. some randomness used for encryption.
This random value can contain a '-' byte
simple due to being randomly generated.

Before, the `IsEncrypted` implementation
incorrectly assumed that an encrypted ETag
cannot contain a '-' since it would be a
multipart ETag. Multipart ETags have a
16 byte value followed by a '-' and the part number.
For example:
```
059ba80b807c3c776fb3bcf3f33e11ae-2
```

However, the following encrypted ETag
```
20000f00db2d90a7b40782d4cff2b41a7799fc1e7ead25972db65150118dfbe2ba76a3c002da28f85c840cd2001a28a9
```
also contains a '-' byte but is not a multipart ETag.

This commit fixes the `IsEncrypted` implementation
simply by checking whether the ETag is at least 32
bytes long. A valid multipart ETag is never 32 bytes
long since a part number must be <= 10000.

However, an encrypted ETag must be at least 32 bytes
long. It contains the encrypted ETag bytes (16 bytes)
and the authentication tag added by the AEAD cipher (again
16 bytes).

## Motivation and Context
ETag, Listing, SSE-S3

## Types of changes

## Checklist:
- [ ] Documentation updated


internal/etag/etag.go
```diff
@@ -143,15 +143,30 @@ func (e ETag) String() string {
...
+	//   return len(e) > 16 && !bytes.ContainsRune(e, '-')
+	//
+	// An encrypted ETag may contain some random bytes - e.g.
+	// and nonce value. This nonce value may contain a '-'
+	// just by its nature of being randomly generated.
+	// The above implementation would incorrectly consider
+	// such an ETag (with a nonce value containing a '-')
+	// as non-encrypted.
+
+	return len(e) >= 32 // We consider all ETags longer than 32 bytes as encrypted
```
poornas, line 161: @aead, should this be len(e) > 32

aead: Why `>`? Based on the code doc an encrypted ETag consists of 16 bytes for its actual value + 16 bytes for the authentication tag. That is `32`.

In general, ETags encrypted using our `sio` package also contain a random value and some cipher / encryption format version information. So they are at least `16 + 16 + 1 + 1 + 12` bytes long. However, that seems like a implementation detail.
If the encryption mechanism for ETags changes, this code may again become incorrect.

A secure encryption of some value will always add at least a 16 byte checksum. So the minium size of encrypted ETag is its plaintext length + 16. Since a ETag must be at least 16 bytes long, `32` is the minimum for an encrypted ETag.

aead: Also, a multipart ETag cannot be `32` bytes long. Then it would not be valid. 

Note that we check the length of the binary ETag, not is hex-encoded length

poornas (approved):
LGTM

kannappanr (approved):

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

PR #14799, opened 2022-04-22
Limit the number of parallel trace/bucket notifications listeners

vadmeste:
## Description
By default, MinIO calculates the number of the maximum parallel
connections based on the available memory. However, there is a use case
when a client sets up too many bucket notifications which can saturate
the number of total S3 requests.

This commit limits the number of trace listeners to 4, console logs
listeners to 4 and bucket notifications listeners to 256. All of them
per server.

## Motivation and Context
Limit the number of concurrent bucket notifications to limit rate requests starvation

## How to test this PR?
Maybe start a simple MinIO server
Start many mc watch commands until you see some erors

## Types of changes

## Checklist:
- [ ] Documentation updated

### DIFF
--- a/internal/pubsub/pubsub.go
+++ b/internal/pubsub/pubsub.go
@@ -18,6 +18,7 @@
 package pubsub
 
 import (
+	"errors"
 	"sync"
 	"sync/atomic"
 )
@@ -32,6 +33,7 @@ type Sub struct {
 type PubSub struct {
 	subs           []*Sub
 	numSubscribers int32
+	maxSubscribers int32
 	sync.RWMutex
 }
 
@@ -53,13 +55,18 @@ func (ps *PubSub) Publish(item interface{}) {
 }
 
 // Subscribe - Adds a subscriber to pubsub system
-func (ps *PubSub) Subscribe(subCh chan interface{}, doneCh <-chan struct{}, filter func(entry interface{}) bool) {
+func (ps *PubSub) Subscribe(subCh chan interface{}, doneCh <-chan struct{}, filter func(entry interface{}) bool) error {
+	totalSubs := atomic.AddInt32(&ps.numSubscribers, 1)
+	if ps.maxSubscribers > 0 && totalSubs > ps.maxSubscribers {
+		atomic.AddInt32(&ps.numSubscribers, -1)
+		return errors.New("maximum clients subscribed exceeded")
+	}
+
 	ps.Lock()
 	defer ps.Unlock()
 
 	sub := &Sub{subCh, filter}
 	ps.subs = append(ps.subs, sub)
-	atomic.AddInt32(&ps.numSubscribers, 1)
 
 	go func() {
 		<-doneCh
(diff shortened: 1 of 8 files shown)


minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

klauspost:
Sounds reasonable with a limit. A few Qs

Are admin requests "S3" requests?
Would it be reasonable for bucket notifications (which are S3 requests) to be a fraction of the s3 request limit?

This is however improvement as it is proposed.

vadmeste:
> Would it be reasonable for bucket notifications (which are S3 requests) to be a fraction of the s3 request limit?

I guess so, maybe I can rewrite this PR to use a different requests queue for admin requests, this looks like a better idea

mshanmu:
@vadmeste , Suggestion, please try to make these limits configurable instead of hardcoding them.

harshavardhana:
> @vadmeste , Suggestion, please try to make these limits configurable instead of hardcoding them.

We should keep them sane values, this doesn't need to be configurable for now @vadmeste .

We need to absolutely sure why anyone needs this configurable because existing limits are already enough and should be sufficient to avoid abuse.

Keeping them as separate limits in API is not necessary. The functionality here is specific to certain calls that keep the client connections Open.. we need to make sure to manage them appropriately.

We should IMO skip the ListenBucket notification from the S3 request queue since it's a zero memory operation and has no I/O. We do not have to allow ListenBucket to take requests from the actual S3 request queue.

This automatically already fixes the problem we saw @vadmeste  - beyond that limiting these APIs to a certain top level number is something that can be done subsequently.

vadmeste:
> We should keep them sane values, this doesn't need to be configurable for now @vadmeste .

I like the idea of a separate queue because some users might be listening to many buckets at the same time. Something like `mc mirror --watch <alias> <destination`

The current value is 256 per node. I can bump to 1000.

vadmeste:
I updated the PR. Trace listeners is limited to 8 listeners per node. No limit for bucket notifications.

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

internal/pubsub/pubsub.go
```diff
@@ -53,13 +55,18 @@ func (ps *PubSub) Publish(item interface{}) {
 }
 
 // Subscribe - Adds a subscriber to pubsub system
-func (ps *PubSub) Subscribe(subCh chan interface{}, doneCh <-chan struct{}, filter func(entry interface{}) bool) {
+func (ps *PubSub) Subscribe(subCh chan interface{}, doneCh <-chan struct{}, filter func(entry interface{}) bool) error {
+	totalSubs := atomic.AddInt32(&ps.numSubscribers, 1)
+	if ps.maxSubscribers > 0 && totalSubs > ps.maxSubscribers {
+		atomic.AddInt32(&ps.numSubscribers, -1)
+		return errors.New("maximum clients subscribed exceeded")
```
harshavardhana: fmt.Errorf() to indicate current subscribers and what is maximum allowed?

harshavardhana (changes requested):

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

harshavardhana (approved):

PR #14818, opened 2022-04-26
Implement purge-on-delete feature for object deletions

Praveenrajmani:
If a special header - `x-minio-purge-on-delete` is passed in the DeleteObject API, the object
will be purged immediately and free up the space

## Description

this is a feature to immediately purge the object while deleting and freeing up the space from `.minio.sys/tmp/trash`

## Motivation and Context

to purge the deleted objects and reclaim the space

## How to test this PR?

- pass a special header `x-minio-purge-on-delete: true` in the delete object API 
- check if the corresponding entries are removed in `.minio.sys/tmp/trash`
- this minio-go PR https://github.com/minio/minio-go/pull/1644 can be used for testing

the benchmark tests by removing 100 entries reported the following

parallel deletes :-

- 0.0006639 ns/op
- memory profile reported around 1024.05 KB (18.05%)

sequential deletes :- 

- 12933971394 ns/op
- 22532.32 KB (79.70%)

## Types of changes

## Checklist:
- [ ] Documentation updated


cmd/xl-storage.go
```diff
@@ -961,17 +961,33 @@ func (s *xlStorage) DeleteVersions(ctx context.Context, volume string, versions
...
+
+	if err := renameFn(filePath, targetPath); err != nil {
+		return err
 	}
-	return Rename(filePath, pathutil.Join(s.diskPath, minioMetaTmpDeletedBucket, pathUUID))
+
+	if purge {
+		go func() {
+			removeAll(targetPath)
+		}()
```
Praveenrajmani: need to benchmark this between parallel and sequential `os.RemoveAll` calls

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

cmd/xl-storage.go
harshavardhana: We should never do this @Praveenrajmani it should be blocking call.. otherwise this can lead to uncontrolled calls on the drive leading to lot more IOPs loss.

harshavardhana: It should be just and the cost should be reflected back for caller for cleaner serializing

```
if purgeImmediately {
      removeAll()
```

harshavardhana (changes requested):

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

harshavardhana:
It is okay have it server side but I am not sure we should add all of this our official SDKs.

This extension will be seldomly used and may even by manually possible to generate.

Conceptually no one would need this functionality other than support related work @Praveenrajmani

harshavardhana:
Actually to make it more proper, perhaps its better to introduce `mc rm --immediate` change to ensure that purge happens immediately.

cmd/bucket-handlers.go
```diff
@@ -474,6 +474,16 @@ func (api objectAPIHandlers) DeleteMultipleObjectsHandler(w http.ResponseWriter,
 	versioned := globalBucketVersioningSys.Enabled(bucket)
 	suspended := globalBucketVersioningSys.Suspended(bucket)
 
+	purgeOnDelete := false
+	if d := r.Header.Get(xhttp.MinIOPurgeOnDelete); d != "" {
+		if b, err := strconv.ParseBool(d); err == nil {
```
harshavardhana, line 488:
```suggestion
		if b, err := strconv.ParseBool(d); err == nil {
```
Lets only support `true, false` not `ParseBool`  - it is too generic.

harshavardhana (changes requested):
There is no internode API update that is needed at `storage-rest-common.go` @Praveenrajmani 

This is also the time we introduce `DeleteVersion(ctx, str, str, fi, opts DeleteOptions)`  which has `forceDelMarker` and `purgeOnDelete` as options.

Praveenrajmani:
> This is also the time we introduce DeleteVersion(ctx, str, str, fi, opts DeleteOptions) which has forceDelMarker and purgeOnDelete as options.

@harshavardhana can we introduce `type DeleteOptions struct` and use it for all Delete calls or keep ObjectOptions and use this instead of having another type?

harshavardhana:
> > This is also the time we introduce DeleteVersion(ctx, str, str, fi, opts DeleteOptions) which has forceDelMarker and purgeOnDelete as options.
> 
> @harshavardhana can we introduce `type DeleteOptions struct` and use it for all Delete calls or keep ObjectOptions and use this instead of having another type?

You need to it specific for storage layer. We can call it StoreOptions as well.

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

cmd/object-api-interface.go
```diff
@@ -78,6 +80,11 @@ type ObjectOptions struct {
 	WalkAscending bool // return Walk results in ascending order of versions
 }
 
+type StoreOptions struct {
+	PurgeOnDelete  bool // set true to enable purge on delete
+	ForceDelMarker bool // force to create a delete marker
+}
+
```
harshavardhana: this should be moved to `storage-datatypes.go` @Praveenrajmani

cmd/bucket-handlers.go
harshavardhana: Actually now to think of this, relaxed behavior is okay. Let's keep it ParseBool for now.

harshavardhana (changes requested):

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

harshavardhana (approved):

poornas:
If bucket is replicated, this purge on delete feature can leave replication targets in inconsistent state. When `mc rm --immediate` is implemented, it should check if replication applies and alert user of the replication side effect and/or require an override.

harshavardhana:
> If bucket is replicated, this purge on delete feature can leave replication targets in inconsistent state. When `mc rm --immediate` is implemented, it should check if replication applies and alert user of the replication side effect and/or require an override.

Thanks @poornas we should perhaps disallow this for `replication`

Praveenrajmani:
@harshavardhana @poornas can we add `--override` flag - `mc rm --immediate --override` for replicated buckets. And for such object deletions, `x-minio-purge-on-delete` header will be set only if `--override` is provided.

and we don't need any special handling for this in the server side?

harshavardhana:
> @harshavardhana @poornas can we add `--override` flag - `mc rm --immediate --override` for replicated buckets. And for such object deletions, `x-minio-purge-on-delete` header will be set only if `--override` is provided.
> 
> and we don't need any special handling for this in the server side?

What @poornas is saying that purge-on-delete defeats replication i.e the entire state needed for replication to work for delete is gone.

Infact to do it safely you need to pass this information differently on replicated buckets. 

i.e save a metadata information about delete on the drive with purge-status and use that status to come to conclusion on deleting entirely or keeping in .trash

Praveenrajmani:
The option that we added in this PR is to immediately purge the object from trash instead of a background thread which will clean up the ".trash" every 5 minutes. Currently, we will just move the entries to the trash and the trash will be cleared in 5mins.

with respective to replication, if the object from primary site is removed with `purge-on-delete` set in DeleteObject API request, we will purge the object immediately and the same object in the target site will be removed as well but won't be purged immediately instead will be cleaned-up by the trash cleanup thread.

my question is, do we want to carry-forward/replicate the purge-on-delete decision to the target also? so, that the object in the target will be purged immediately as well? If this feature is planned to be used only in special situations for support related work, why would we want this decision to be replicated in the target site?

> i.e save a metadata information about delete on the drive with purge-status and use that status to come to conclusion on deleting entirely or keeping in .trash

If we need to replicate the purge-on-delete decision, can we preserve the decision in https://github.com/minio/minio/blob/master/cmd/bucket-replication.go#L1260 (in-memory) instead of saving it in xl.meta (drive)?

let me know if i'm mis-understanding the context here @harshavardhana @poornas @kannappanr

harshavardhana:
> The option that we added in this PR is to immediately purge the object from trash instead of a background thread which will clean up the ".trash" every 5 minutes. Currently, we will just move the entries to the trash and the trash will be cleared in 5mins.
> 
> with respective to replication, if the object from primary site is removed with `purge-on-delete` set in DeleteObject API request, we will purge the object immediately and the same object in the target site will be removed as well but won't be purged immediately instead will be cleaned-up by the trash cleanup thread.
> 
> my question is, do we want to carry-forward/replicate the purge-on-delete decision to the target also? so, that the object in the target will be purged immediately as well? If this feature is planned to be used only in special situations for support related work, why would we want this decision to be replicated in the target site?
> 
> > i.e save a metadata information about delete on the drive with purge-status and use that status to come to conclusion on deleting entirely or keeping in .trash
> 
> If we need to replicate the purge-on-delete decision, can we preserve the decision in https://github.com/minio/minio/blob/master/cmd/bucket-replication.go#L1260 (in-memory) instead of saving it in xl.meta (drive)?
> 
> let me know if i'm mis-understanding the context here @harshavardhana @poornas @kannappanr

You cannot save it in memory for consistency reasons, this PR as it is won't work in replicated buckets and will create an inconsistent situation.

Purge on delete on the source cluster should happen only after the purge on delete has been propagated to replica.

That is the flow that needs to be maintained.

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

harshavardhana (changes requested):
Did @kannappanr fill you in regarding re-use x-minio-force-delete for this? - all we have to do is if x-minio-force-delete is set then we can simply delete the content from the Trash additionally to the rename call.

Praveenrajmani:
> Did @kannappanr fill you in regarding re-use x-minio-force-delete for this? - all we have to do is if x-minio-force-delete is set then we can simply delete the content from the Trash additionally to the rename call.

got it, will close this PR and open a new one

PR #14831, opened 2022-04-28
Do not use inline data size in xl.meta quorum calculation

vadmeste:
## Description
Some users who are having low EC parity are not able to retrieve some objects
because of a lack of quorum, although the object data are still not lost.

Data shards of the same object can have different inline/not-inline decision
in multiple disks. It can be found in  outdated disks after inline
decision changes, such as, enabling bucket versioning configuration
which changes the small file threshold.

In this case and when the parity of an object becomes low, GET object can return 503
because it is not unable to calculate the xl.meta quorum, though the data is still
there and not lost

So this commit will be disable taking the size of the inline data into
consideration when calculating the xl.meta quorum.

## Motivation and Context
Some users who are having low EC parity are not able to retrieve some objects
because of a lack of quorum, though theorically the object data are still not lost.

## How to test this PR?
Not trivial - we inspected some real data

## Types of changes

## Checklist:
- [ ] Documentation updated


klauspost:
@vadmeste Can you re-confirm that the metadata fields are consistent? The sample you sent was, just want to make sure all claim to have inline data.

klauspost:
@vadmeste I would actually expect it to be consistent, if they were called with `readData=true`, but it may be that it doesn't read parts on disk if the inline header was set.

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

harshavardhana:
> @vadmeste I would actually expect it to be consistent, if they were called with `readData=true`, but it may be that it doesn't read parts on disk if the inline header was set.

How about we rely on `xlmeta.data` instead of the header @klauspost so if the .data is `0` then we attempt to read from DataDir/part.1`

harshavardhana:
```go
        if readData {
                if len(fi.Data) > 0 || fi.Size == 0 {
                        if fi.InlineData() {
                                // If written with header we are fine.
                                return fi, nil
                        }
                        if fi.Size == 0 || !(fi.VersionID != "" && fi.VersionID != nullVersionID) {
                                // If versioned we have no conflicts.
                                fi.SetInlineData()
                                return fi, nil
                        }

                        // For overwritten objects without header we might have a conflict with
                        // data written later.
                        // Check the data path if there is a part with data.
                        partPath := fmt.Sprintf("part.%d", fi.Parts[0].Number)
                        dataPath := pathJoin(path, fi.DataDir, partPath)
                        _, err = s.StatInfoFile(ctx, volume, dataPath, false)
                        if err != nil {
                                // Set the inline header, our inlined data is fine.
                                fi.SetInlineData()
                                return fi, nil
                        }
                        // Data exists on disk, remove the version from metadata.
                        fi.Data = nil
                }

                // Reading data for small objects when
                // - object has not yet transitioned
                // - object size lesser than 128KiB
                // - object has maximum of 1 parts
                if fi.TransitionStatus == "" &&
                        fi.DataDir != "" && fi.Size <= smallFileThreshold &&
                        len(fi.Parts) == 1 {
                        partPath := fmt.Sprintf("part.%d", fi.Parts[0].Number)
                        dataPath := pathJoin(volumeDir, path, fi.DataDir, partPath)
                        fi.Data, _, err = s.readAllData(ctx, volumeDir, dataPath)
                        if err != nil {
                                return FileInfo{}, err
                        }
                }
        }
```

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

harshavardhana (approved):

harshavardhana:
@vadmeste can you add a test for the this with sample data?

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

klauspost (approved):
Let's give it a try.

PR #14899, opened 2022-05-11
add support for extra Prometheus labels

### LINKED ISSUE #14353: MINIO_PROMETHEUS_JOB_ID is insufficient when scraping metrics via prometheus-operator when having more minio instances
Right now minio exposes metrics and is possible to scrape it with prometheus with servicemonitor etc.
The problem is that currently minio is using only MINIO_PROMETHEUS_JOB_ID to filter out metrics, which is extremely limiting when using prometheus operator and minio in kubernetes with more than 1 minio installation.
Prometheus operator + ServiceMonitor will cause jobs to be names as service is named on deployment.

Let say we have 2 namespaces (lets say bravo and charlie), with the same minio installs in each namespace.
As a result we have a services minio.bravo.svc and minio.charlie.svc
Now we need to configure minio to use metrics from prometheus, but because how prometheus operator and servicemonitors are created in both cases job name will be 'minio'.
Now minio querying for the metric in the UI will get metrics fro BOTH installations.

What is missing is extra option to pass additional labels to filter out metrics.

The best option would be to add another env var to pass additional labels to be used when querying prometheus.
Something like MINIO_PROMETHEUS_LABEL_FILTER so I could add `namespace="bravo"` and `namespace="charlie"` for each individual config of the minio. This way they could use it to fetch metrics only for specific instance.
This would also allow to pass fine tuned labels such as `namespace="bravo",cluster="echo"` to fine grain metric even more.

- what works now is to actually exploit the way metrics are queried, so that injecting desired labels occurs, in example setting MINIO_PROMETHEUS_JOB_ID to `minio",namespace="beta` and `minio",namespace="charlie` will in effect allow to query with labels `job="minio",namespace="beta"` and `job="minio",namespace="charlie"`, but it's dirty ;)

- differently named minio installs, but that tend to be problematic in certain situations
- some weird metric label renaming which ends in ugly hacks
- prometheus-operator provides `additionalScrapeConfigs` to load extra snippets to prometheus to scrape, the problem is that user may not be able to control prometheus installation for that. There is also ongoing process of defining new CRD for that case https://github.com/prometheus-operator/prometheus-operator/issues/2787 but it will probably take a lot of time before it will be available.

**Additional context**

harshavardhana:
## Description
add support for extra Prometheus labels

## Motivation and Context
fixes #14353

## How to test this PR?
Corresponding PR to console is needed https://github.com/minio/console/pull/1936

## Types of changes

## Checklist:
- [ ] Documentation updated

### DIFF
--- a/cmd/common-main.go
+++ b/cmd/common-main.go
@@ -190,6 +190,10 @@ func minioConfigToConsoleFeatures() {
 		os.Setenv("CONSOLE_PROMETHEUS_URL", value)
 		if value := env.Get("MINIO_PROMETHEUS_JOB_ID", "minio-job"); value != "" {
 			os.Setenv("CONSOLE_PROMETHEUS_JOB_ID", value)
+			// Support additional labels for more granular filtering.
+			if value := env.Get("MINIO_PROMETHEUS_EXTRA_LABELS", ""); value != "" {
+				os.Setenv("CONSOLE_PROMETHEUS_EXTRA_LABELS", value)
+			}
 		}
 	}
 	// Enable if LDAP is enabled.


donatello (approved):

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

poornas (approved):

PR #14986, opened 2022-05-27
fix the hang problem with command ls recursive

DoubleJ213:
## Description
when i use the command mc ls -r xxx. the programe always hangs. 
With too many dirs and files in my cluster, but the command mc ls --versions works well for me.
After debugging, I found that the program is often stuck in the method erasureServerPools.listMerged
when range pool.sets the goroutine is always end with the cancel context

## Types of changes

### DIFF
--- a/cmd/metacache-server-pool.go
+++ b/cmd/metacache-server-pool.go
@@ -238,7 +238,7 @@ func (z *erasureServerPools) listPath(ctx context.Context, o *listPathOptions) (
 	}(*o)
 
 	entries, err = filteredResults()
-	cancelList()
+	defer cancelList()
 	wg.Wait()
 	if listErr != nil && !errors.Is(listErr, context.Canceled) {
 		return entries, listErr
@@ -324,7 +324,7 @@ func (z *erasureServerPools) listMerged(ctx context.Context, o listPathOptions, 
 		return len(oMeta.versions) > len(eMeta.versions)
 	})
 
-	cancelList()
+	//cancelList()
 	wg.Wait()
 	if err != nil {
 		return err


cmd/metacache-server-pool.go
```diff
@@ -238,7 +238,7 @@ func (z *erasureServerPools) listPath(ctx context.Context, o *listPathOptions) (
 	}(*o)
 
 	entries, err = filteredResults()
-	cancelList()
+	defer cancelList()
```
klauspost, line 241: When filteredResults() returns we supposedly have the results we want, and we can cancel. 

Deferring this means the list will run to completion, wasting a lot of resources before we return.

This is definitely not the solution you are looking for,

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

cmd/metacache-server-pool.go
```diff
@@ -324,7 +324,7 @@ func (z *erasureServerPools) listMerged(ctx context.Context, o listPathOptions,
 		return len(oMeta.versions) > len(eMeta.versions)
 	})
 
-	cancelList()
+	//cancelList()
```
klauspost, line 327: Again, this is needed if an error occurs in mergeEntryChannels. For now it is AFAICT only upstream cancelation.

This change should not affect anything.

klauspost (commented):
Could you instead provide a reproducer? I don't see how this fixes the issue you are describing. Are you sure you aren't just comparing cached vs uncached calls?

harshavardhana:
A reproducer is needed here. Please provide one and then submit a PR.

Your PR actually introduces bugs here.

DoubleJ213:
Sorry for no reproducers because i cannot reproduce the issue in my other minio cluster . Not always reproducible even under problematic clusters

Thanks for the explanation @klauspost  ! And apologies that i not deep understanding the code .

i'll check it goes on to find why after filteredResults and upstream cancel action the z.listMerged goroutine is never finished(no debuglog output "listMerged returned with"). so wg.Wait() hangs

btw i can fix the problem with upstrem defer cancellist or add time.sleep 1s  before cancellist. Do you have any suggestions ？

PR #15163, opened 2022-06-23
walk: Simplify forward-to logic

vadmeste:
## Description
Currently, forward can be compared with entry names with __XLDIR__
marker, though the code just works, simplify it further for the upcoming
commit.

## Motivation and Context
Code simplification

## How to test this PR?
Listing, listing with continuation

## Types of changes

## Checklist:
- [ ] Documentation updated

### DIFF
--- a/cmd/metacache-walk.go
+++ b/cmd/metacache-walk.go
@@ -144,36 +144,10 @@ func (s *xlStorage) WalkDir(ctx context.Context, opts WalkDirOptions, wr io.Writ
 		}
 		dirObjects := make(map[string]struct{})
 		for i, entry := range entries {
-			if len(prefix) > 0 && !strings.HasPrefix(entry, prefix) {
-				// Do do not retain the file, since it doesn't
-				// match the prefix.
-				entries[i] = ""
-				continue
-			}
-			if len(forward) > 0 && entry < forward {
-				// Do do not retain the file, since its
-				// lexially smaller than 'forward'
-				entries[i] = ""
-				continue
-			}
-			if strings.HasSuffix(entry, slashSeparator) {
-				if strings.HasSuffix(entry, globalDirSuffixWithSlash) {
-					// Add without extension so it is sorted correctly.
-					entry = strings.TrimSuffix(entry, globalDirSuffixWithSlash) + slashSeparator
-					dirObjects[entry] = struct{}{}
-					entries[i] = entry
-					continue
-				}
-				// Trim slash, since we don't know if this is folder or object.
-				entries[i] = entries[i][:len(entry)-1]
-				continue
-			}
-			// Do do not retain the file.
-			entries[i] = ""
-
 			if contextCanceled(ctx) {
 				return ctx.Err()
 			}
+
 			// If root was an object return it as such.
 			if HasSuffix(entry, xlStorageFormatFile) {
 				var meta metaCacheEntry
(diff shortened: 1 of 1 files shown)


minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

klauspost:
Not really sure I see the simplification - it just looks like the code was moved around.

Please double and triple check that "forward" and "prefix" are completely skipped. If this regresses on that it will have a very negative impact on listing operations.

Since I don't understand the improvement this makes I tend to want to keep the code as-is.

vadmeste:
> Not really sure I see the simplification - it just looks like the code was moved around.

In fact I am trying to send separate PRs to facilitate reviewing. The PR which will come after that will support listing until a given object name and quit early, this is needed to support healing a subset of the namespace (from the beginning until a given object name)

The only confusion in the existing code is that we compare `object__XLDIR__` with 'forward', although it seems it works but this is still making the code hard to read (also it makes hard for the next PR to exist)

This PR:
- Moves the order of code block (should not have an effect)
- Removes __XLDIR__ from the entry name before comparing to 'forward'
- Removes unneeded code at line 251

harshavardhana:
> The only confusion in the existing code is that we compare `object__XLDIR__` with 'forward', although it seems it works but this is still making the code hard to read (also it makes hard for the next PR to exist)

Can you just fix that instead of moving the code around? @vadmeste - let's get that in first and then try to clean things up?

klauspost:
It is quite sensitive wrt ordering. It took several attempts to get the objects ending with `/` ordered correctly, and resume correctly.

So I am a bit hesitant to change too much.

vadmeste:
> It is quite sensitive wrt ordering. It took several attempts to get the objects ending with / ordered correctly, and resume correctly.

> So I am a bit hesitant to change too much.

Yeah I totally understand, I will make more granular changes

cmd/metacache-walk.go
```diff
@@ -194,7 +168,9 @@ func (s *xlStorage) WalkDir(ctx context.Context, opts WalkDirOptions, wr io.Writ
 				meta.name = strings.TrimSuffix(meta.name, SlashSeparator)
 				meta.name = pathJoin(current, meta.name)
 				meta.name = decodeDirObject(meta.name)
-				out <- meta
+				if forward == "" || meta.name >= forward {
```
klauspost, line 171: It is very expensive that we have to read metadata for every file we are *skipping*.

Before the skip would be done purely based on the name. Now we have to decode and clean up before we skip.

We must be able to skip forward without having to read the file.

klauspost (changes requested):

cmd/metacache-walk.go
vadmeste: In fact, this is only executed when we are inside an object (it sees xl.meta and part.*). This happens only if you are listing one directory marker

harshavardhana: @klauspost do you concur? - do we need this PR @vadmeste - does it optimize reading xl.meta?

klauspost: I still don't see the point of the change. Are we fixing anything?

vadmeste: > I still don't see the point of the change. Are we fixing anything?

No. My point is to introduce a new listing feature which is called EndAT, which means list until a given object. But this code is hard, that's why I am sending separate PRs to review. It is also hard to do EndAt without this PR (which avoids comparing xl.meta with 'forward'

harshavardhana: hmm, interesting can you send the entire PR? @vadmeste hard to imagine what it does :-)

harshavardhana:
@vadmeste are you planning to send a larger change?

harshavardhana:
Going to close this since it looks like you are still working on something locally. Please submit when you have something concrete.

PR #15399, opened 2022-07-25
feat: add object prefixes as a hint for Listing()

harshavardhana:
## Description
feat: add object prefixes as a hint for Listing()

## Motivation and Context
this new feature optimizes listing at various prefixes
by listing a single disk when we are absolutely sure
that there are no quorum errors

## How to test this PR?
Nothing should change

## Types of changes

## Checklist:
- [ ] Documentation updated

### DIFF
--- a/cmd/erasure-object.go
+++ b/cmd/erasure-object.go
@@ -108,6 +108,18 @@ func (er erasureObjects) CopyObject(ctx context.Context, srcBucket, srcObject, d
 	// List all online disks.
 	onlineDisks, modTime := listOnlineDisks(storageDisks, metaArr, errs)
 
+	var online int
+	defer func() {
+		parentDir := path.Dir(srcObject)
+		if online == er.setDriveCount {
+			er.listQuorumCache.Add(parentDir, nil)
+		} else {
+			// metadata updates that are partial must
+			// be invalidated for listing.
+			er.listQuorumCache.Remove(parentDir)
+		}
+	}()
+
 	// Pick latest valid metadata.
 	fi, err := pickValidFileInfo(ctx, metaArr, modTime, readQuorum)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

klauspost:
I think we should discuss this. I don't see how this gives any insurance that it will work. In fact I think most inconsistencies would go though, and we'd be in a much worse situation than when we asked 3 drives by default.

The only way I see this is to be *enabled* for a path is when a full, recursive listing returned fully consistent results for all disks. We can also add non-recursive listings, as long as we know it is only for the current prefix.

It can be *disabled* by any of the conditions you list, but a single successful write doesn't really tell you anything about the state of the prefix.

harshavardhana:
> I think we should discuss this. I don't see how this gives any insurance that it will work. In fact I think most inconsistencies would go though, and we'd be in a much worse situation than when we asked 3 drives by default.
> 
> The only way I see this is to be _enabled_ for a path is when a full, recursive listing returned fully consistent results for all disks. We can also add non-recursive listings, as long as we know it is only for the current prefix.
> 
> It can be _disabled_ by any of the conditions you list, but a single successful write doesn't really tell you anything about the state of the prefix.

The idea is that it's just for the same parentDir if you see a quorum loss on it, then you simply invalidate the same parentDir. 

There shouldn't be a situation where you would have incorrect parentDir. To keep the situation sane we can also serialize lookup and addition to LRU between various mutation calls. But this should be sufficient for now. 

It must be done during PUT because listing is batched there is no such thing as full listing in S3 and most applications don't do full recursive listing and wait for the entire results. 

Infact there must be entries during read phase as well, I will be adding those.

harshavardhana:
This will not work in it's current form, closing.

PR #15403, opened 2022-07-26
fix:timer_stop

jiuker:
## Description
According to the timer provided by the official package of golang, the call to the Stop function needs to be done for channels that have not received an end. To prevent excessive changes to the program, this pr simply wraps the timer to handle the behavior of the Stop function call to prevent memory leaks. If this is not handled, a memory leak may occur, which is a potential bug.
## Motivation and Context
fix it
## Types of changes

## Checklist:
- [ ] Documentation updated


harshavardhana (changes requested):
This is wrong change all our timers never expire and only Stop() when MinIO server is exiting.

We won't be able to take this change.

jiuker:
> This is wrong change all our timers never expire and only Stop() when MinIO server is exiting.
> 
> We won't be able to take this change.

In fact, this modification does not modify any logic, but there is a unified stop, which can modify the stop behavior of timer, which is recommended by golang

PR #15546, opened 2022-08-18
fix: increase concurrency of DeleteObjects() to N/10th

harshavardhana:
## Description
fix: increase concurrency of DeleteObjects() to N/10th

## Motivation and Context
instead of keeping the value 10 and static, make
the concurrency a function of incoming number of
objects being deleted.

## How to test this PR?
Nothing should change, only performance difference.

## Types of changes

## Checklist:
- [ ] Documentation updated

### DIFF
--- a/cmd/erasure-server-pool.go
+++ b/cmd/erasure-server-pool.go
@@ -1046,8 +1046,16 @@ func (z *erasureServerPools) DeleteObjects(ctx context.Context, bucket string, o
 	poolObjIdxMap := map[int][]ObjectToDelete{}
 	origIndexMap := map[int][]int{}
 
+	// Always perform 1/10th of the number of objects per delete
+	concurrent := len(objects) / 10
+	if concurrent <= 10 {
+		// if we cannot get 1/10th then choose the number of
+		// objects as concurrent.
+		concurrent = len(objects)
+	}
+
 	var mu sync.Mutex
-	eg := errgroup.WithNErrs(len(objects)).WithConcurrency(10)
+	eg := errgroup.WithNErrs(len(objects)).WithConcurrency(concurrent)
 	for j, obj := range objects {
 		j := j
 		obj := obj


minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

poornas (approved):

cmd/erasure-server-pool.go
```diff
@@ -1046,8 +1046,16 @@ func (z *erasureServerPools) DeleteObjects(ctx context.Context, bucket string, o
 	poolObjIdxMap := map[int][]ObjectToDelete{}
 	origIndexMap := map[int][]int{}
 
+	// Always perform 1/10th of the number of objects per delete
+	concurrent := len(objects) / 10
+	if concurrent <= 0 {
+		// if we cannot get 1/10th then choose the number of
+		// objects as concurrent.
+		concurrent = len(objects)
```
vadmeste, line 1054: I guess let's have a maximum of this as a self defense

vadmeste (commented):
One comment

cmd/erasure-server-pool.go
harshavardhana: how would that be self-defense? @vadmeste provide me a case where defense is needed.

vadmeste: > how would that be self-defense? @vadmeste provide me a case where defense is needed.

We should have a limit of the number of goroutines, even if we don't have a use case where we pass a high number of objects, we may do in the future, so if guess `if concurrent > 1000 { concurrent = 1000 }` is a good self defence code.

harshavardhana: @vadmeste these are supposed to be shortlived and the point is we will never support more than 1000 for that we have DeletePrefix().  Even if we do the logic can be changed here in the future. 

Currently, this is designed to have 1000/10 = 100 at max - so we will never have goroutines beyond 100 at any given point in time.

harshavardhana: For example the logic here is to make sure that the ratio of number of concurrent calls for 800 and 80 are same i.e 

```
800/10 = 80
```

```
80/10 = 8 - its less than 10 choose 80
```

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

vadmeste (approved):
LGTM

PR #15610, opened 2022-08-30
audit: account for response headers separately

krisis:
## Description
Today, audit-log entries account for response headers along with the response body towards `OutputBytes` (seen as `api.tx` in json). With this change, audit logs would account for response header bytes in `HeaderBytes` (seen as `api.txHeaders` in json). 

## How to test this PR?
Requires setting up an audit-log target, like a simple webhook server (https://github.com/krishnasrinivas/minio-webhook)

## Types of changes

## Checklist:
- [ ] Documentation updated

### DIFF
--- a/internal/logger/audit.go
+++ b/internal/logger/audit.go
@@ -45,6 +45,8 @@ type ResponseWriter struct {
 	StartTime       time.Time
 	// number of bytes written
 	bytesWritten int
+	// number of bytes of response headers written
+	headerBytesWritten int
 	// Internal recording buffer
 	headers bytes.Buffer
 	body    bytes.Buffer
@@ -86,10 +88,10 @@ func (lrw *ResponseWriter) Write(p []byte) (int, error) {
 // Write the headers into the given buffer
 func (lrw *ResponseWriter) writeHeaders(w io.Writer, statusCode int, headers http.Header) {
 	n, _ := fmt.Fprintf(w, "%d %s\n", statusCode, http.StatusText(statusCode))
-	lrw.bytesWritten += n
+	lrw.headerBytesWritten += n
 	for k, v := range headers {
 		n, _ := fmt.Fprintf(w, "%s: %s\n", k, v[0])
-		lrw.bytesWritten += n
+		lrw.headerBytesWritten += n
 	}
 }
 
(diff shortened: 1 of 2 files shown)


minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

harshavardhana (approved):

harshavardhana (changes requested):
Can you how this affects s3_sent_bytes? @krisis

internal/logger/audit.go
```diff
@@ -86,10 +88,10 @@ func (lrw *ResponseWriter) Write(p []byte) (int, error) {
 // Write the headers into the given buffer
 func (lrw *ResponseWriter) writeHeaders(w io.Writer, statusCode int, headers http.Header) {
 	n, _ := fmt.Fprintf(w, "%d %s\n", statusCode, http.StatusText(statusCode))
-	lrw.bytesWritten += n
+	lrw.headerBytesWritten += n
```
vadmeste, line 91: maybe we still need to add 'n' to lrw.bytesWritten here. This is not a 'written request body' but 'bytes written'

vadmeste (commented):
One comment, also why this is useful ?

internal/logger/audit.go
krisis: @vadmeste I noticed that this metric included response headers size in addition to response body. If that's the intention, then we don't need this PR.

krisis:
> Can you how this affects s3_sent_bytes? @krisis

This doesn't affect the s3_sent_bytes calculation. s3_sent_bytes comes `globalConnStats` managed by `setHTTPStatsHandler` middleware.

harshavardhana:
> 
> This doesn't affect the s3_sent_bytes calculation. s3_sent_bytes comes `globalConnStats` managed by `setHTTPStatsHandler` middleware.

Okay @krisis

internal/logger/audit.go
harshavardhana: We need them to separate otherwise we may be providing false information.

harshavardhana (approved):

internal/logger/audit.go
vadmeste: > We need them to separate otherwise we may be providing false information.

I think maybe let's avoid counting headers in the response to make it consistent with InputBytes and we do not have to publish txHeaders for now.

harshavardhana: @vadmeste no we need to keep it to have a good understanding of how much is tx vs txHeaders for example in one of the customers the metadata was HTTP headers >> object size

Adding this is harmless..

vadmeste (approved):
LGTM

PR #15649, opened 2022-09-04
verify RenameData() versions to be consistent

harshavardhana:
## Description
verify RenameData() versions to be consistent

## Motivation and Context
xl.meta gets written and never rolled back, however
we definitely need to validate the state that is
persisted on the disk, if there are inconsistencies

- less than write quorum we should return an error
  to the client

- if write quorum was achieved however there are
  inconsistent xl.meta's we should simply trigger
  an MRF on them

## How to test this PR?
Hard to test mostly a generated hand-crafted scenario.

## Types of changes

## Checklist:
- [ ] Documentation updated

### DIFF
--- a/cmd/erasure-metadata-utils.go
+++ b/cmd/erasure-metadata-utils.go
@@ -27,6 +27,30 @@ import (
 	"github.com/minio/minio/internal/sync/errgroup"
 )
 
+// figure out the most commonVersions across disk that satisfies
+// the 'writeQuorum' this function returns '0' if quorum cannot
+// be achieved and disks have too many inconsistent versions.
+func reduceCommonVersions(diskVersions []uint64, writeQuorum int) (commonVersions uint64) {
+	diskVersionsCount := make(map[uint64]int)
+	for _, versions := range diskVersions {
+		diskVersionsCount[versions]++
+	}
+
+	max := 0
+	for versions, count := range diskVersionsCount {
+		if max < count {
+			max = count
+			commonVersions = versions
+		}
+	}
+
+	if max >= writeQuorum {
+		return commonVersions
+	}
+
+	return 0
+}
+
 // Returns number of errors that occurred the most (incl. nil) and the
 // corresponding error value. NB When there is more than one error value that
 // occurs maximum number of times, the error value returned depends on how
--- a/cmd/erasure-healing.go
+++ b/cmd/erasure-healing.go
@@ -625,7 +625,7 @@ func (er erasureObjects) healObject(ctx context.Context, bucket string, object s
 		partsMetadata[i].Erasure.Index = i + 1
 
 		// Attempt a rename now from healed data to final location.
-		if err = disk.RenameData(ctx, minioMetaTmpBucket, tmpID, partsMetadata[i], bucket, object); err != nil {
+		if _, err = disk.RenameData(ctx, minioMetaTmpBucket, tmpID, partsMetadata[i], bucket, object); err != nil {
 			logger.LogIf(ctx, err)
 			return result, err
 		}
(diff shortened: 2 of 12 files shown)


minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

cmd/erasure-metadata-utils.go
```diff
@@ -27,6 +27,27 @@ import (
 	"github.com/minio/minio/internal/sync/errgroup"
 )
 
+func reduceCommonVersions(diskVersions []int, writeQuorum int) (commonVersions int) {
```
klauspost: Document

harshavardhana:
Open to suggestions on what else we can verify! here .. @vadmeste @klauspost

vadmeste:
> if write quorum was achieved however there are
inconsistent xl.meta's we should simply trigger
an MRF on them

I am not sure if this is needed, I mean, returning errors should be enough. Do you see any use case where comparning number of versions is needed ?

klauspost (approved):
Other than docs for a non-obvious function, LGTM.

klauspost:
> Open to suggestions on what else we can verify! here .. @vadmeste @klauspost

@harshavardhana `xlMeta.versions[...].header.Signature` is a signature for each version, that ignores the per-disk variances. You can return a combined hash. For example with github.com/zeebo/xxh3:

```
	combined := xxh3.New()
	for _, ver := range xlMeta.versions {
		combined.Write(ver.header.Signature[:])
	}
	sum := combined.Sum64()
```
.. or if you want something faster, but weaker:

```
	var combined uint32
	for i, ver := range xlMeta.versions {
		x := binary.LittleEndian.Uint32(ver.header.Signature[:])
		x ^= uint32(i+1) * 2654435761 // Multiply i by big prime
		combined += x                 // Allow overflows
	}
```
Since versions are ordered, this should be stable.

harshavardhana:
done @klauspost @vadmeste

klauspost (approved):
LGTM

harshavardhana:
> I am not sure if this is needed, I mean, returning errors should be enough. Do you see any use case where comparning number of versions is needed ?

It is needed to return "error" when we couldn't really do what we were supposed to do i.e if we wrote in write quorum we better have consistent versions on the disk. 

This is to ensure that client knows that their PUT was never successful today there are cases where this can eventually happen on a very flaky system @vadmeste and healing has not yet kicked in.

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

vadmeste (approved):
LGTM

vadmeste:
@harshavardhana I was thinking about this again, this code is aggressive because this will disallow PUT object when there is some versions disparency between xl.meta between disks. This lead to a blocking situation.

In general, a soft delete can be successful if one node in a distributed setup is offline and I think, in the real world, we can have many objects having this disparency. Maybe we should trigger healing instead of disallowing PUT object ?

harshavardhana:
> @harshavardhana I was thinking about this again, this code is aggressive because this will disallow PUT object when there is some versions disparency between xl.meta between disks. This lead to a blocking situation.
> 

Only if we can't achieve quorum @vadmeste 

The point is we can never return success for situations when a read can fail in the next call. This PR reduces the room for such eventualities.

> In general, a soft delete can be successful if one node in a distributed setup is offline and I think, in the real world, we can have many objects having this disparency. Maybe we should trigger healing instead of disallowing PUT object ?

You didn't get the PR right - we only return error if we can't achieve quorum.

harshavardhana:
We need to do this for DELETEs as well..

PR #15838, opened 2022-10-11
Add Data-Dir as criteria to pick the quorum FileInfo

vadmeste:
## Description
With the replication feature, modtime alone is not reliable enough to pick a FileInfo{} which has a quorum in the disks. The reason is that replication sets ModTime when replicating objects. Add Data-Dir criteria for versions with associated data (not delete marker) to pick FileInfo{} with correct related info, such as EcN, EcM and other informations.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/erasure-healing-common.go
+++ b/cmd/erasure-healing-common.go
@@ -112,6 +112,32 @@ func listObjectModtimes(partsMetadata []FileInfo, errs []error) (modTimes []time
 	return modTimes
 }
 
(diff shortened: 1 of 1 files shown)


harshavardhana:
This was removed on purpose @vadmeste in the past please check the history on why it was removed.

vadmeste:
>  This was removed on purpose @vadmeste in the past please check the history on why it was removed.

yes you mean this, https://github.com/minio/minio/commit/c791de0e1eae2f9ae743e1fb866dcf1481748746

I will send another PR after our last discussion to use quorum parity instead

PR #15857, opened 2022-10-13
replication: Attempt delete marker replication after object is replic…

poornas:
…ated

This is to avoid possibility of delete marker replication being marked completed prematurely.

## Motivation and Context
Ensure delete marker replication success, especially since the recent optimizations to heal on HEAD, LIST and GET can force replication attempts on delete marker before underlying object version could be synced

## How to test this PR?
Set up replication and upload one or more objects and set delete marker when target is down. after target comes up, `mc stat alias/bucket/obj`.
 if delete marker is top most, it is marked completed because it is attempted to replicate out of order (with master). 

With this PR both clusters should get in sync

## Types of changes

## Checklist:

### DIFF
--- a/cmd/object-handlers.go
+++ b/cmd/object-handlers.go
@@ -740,6 +740,15 @@ func (api objectAPIHandlers) headObjectHandler(ctx context.Context, objectAPI Ob
 			}
 			QueueReplicationHeal(ctx, bucket, objInfo)
 		}
+		// do an additional verification whether object exists when opts.DeleteMarker is set by source
+		// cluster as part of delete marker replication
+		if opts.DeleteMarker && opts.ProxyHeaderSet {
+			opts.VersionID = ""
+			goi, gerr := getObjectInfo(ctx, bucket, object, opts)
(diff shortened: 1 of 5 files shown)


cmd/object-handlers.go
```diff
@@ -740,6 +740,15 @@ func (api objectAPIHandlers) headObjectHandler(ctx context.Context, objectAPI Ob
 			}
 			QueueReplicationHeal(ctx, bucket, objInfo)
 		}
+		// do an additional verification whether object exists when opts.DeleteMarker is set by source
+		// cluster as part of delete marker replication
+		if opts.DeleteMarker && opts.ProxyHeaderSet {
+			opts.VersionID = ""
+			goi, gerr := getObjectInfo(ctx, bucket, object, opts)
+			if gerr == nil || goi.VersionID != "" { // object layer returned more info because object is deleted
+				w.Header()[xhttp.MinIOTargetReplicationReady] = []string{"true"}
```
```suggestion
				w.Header().Set(xhttp.MinIOTargetReplicationReady, "true")
```

harshavardhana (changes requested):

harshavardhana (approved):

PR #16083, opened 2022-11-17
Distribute replication ops based on names

klauspost:
## Motivation and Context

Replication workers can end up fighting over the same object locks when many versions of an object are processed at the same time.

This leads to huge resource use and general slowdown of the replication.

## Description

Instead of having a single channel shared across many workers we give each worker a channel and distribute objects based on a hash of their bucket+object key.

This will ensure that the same object will end up at the same worker (on each server), thereby minimizing contention, and still ensure random distribution between workers.

Objects are re-distributed when the worker count is re-scaled, but that should be a rather rare occurrence.

## How to test this PR?

Replication setup.

## Types of changes
- [x] Optimization (provides speedup with no functional changes)

PR #16084, opened 2022-11-17
Fix replication autoscaling deadlock

klauspost:
## Description

When auto-scaling replication workers, the operation will deadlock.

An Rlock would be held while `ResizeWorkers` would try to grab a write lock. This will deadlock `mu` completely in ReplicationPool for all future operations.

Also make sure that atomic int32 values are aligned to avoid potential 32 bit crashes.

Fixed in #16083 as well, but this is cleaner to backport.

## Types of changes

### DIFF
--- a/cmd/bucket-replication.go
+++ b/cmd/bucket-replication.go
@@ -1509,6 +1509,10 @@ var (
 
 // ReplicationPool describes replication pool
 type ReplicationPool struct {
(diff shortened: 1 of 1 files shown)


vadmeste (approved):
LGTM

harshavardhana (approved):

PR #15929, opened 2022-10-22
remove gateway completely

harshavardhana:
## Description
remove gateway completely

## Motivation and Context
gateway code has been deprecated for over close
to 2yrs now, with this change we are allowing
for full transition towards server deployments.

## How to test this PR?
Nothing should change 

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Optimization (provides speedup with no functional changes)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [ ] Fixes a regression (If yes, please add `commit-id` or `PR #` here)
- [ ] Unit tests added/updated
- [x] Internal documentation updated
- [ ] Create a documentation update request [here](https://github.com/minio/docs/issues/new?label=doc-change,title=Doc+Updated+Needed+For+PR+github.com%2fminio%2fminio%2fpull%2fNNNNN)

PR #13171, opened 2021-09-09
fix: multipart replication and encrypted etag for sse-s3

harshavardhana:
## Description
fix: multipart replication and encrypted etag for sse-s3

## Motivation and Context
Replication was not working properly for encrypted
objects in single PUT object for preserving etag,

We need to make sure to preserve etag such that replication
works properly and not gets into infinite loops of copying
due to ETag mismatches.

## How to test this PR?
https://gist.github.com/harshavardhana/579e776ab8396d122a22b9f4a36dc0f5

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Optimization (provides speedup with no functional changes)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [x] Fixes a regression yes has been since #12375
- [ ] Documentation updated
- [ ] Unit tests added/updated

PR #16129, opened 2022-11-26
preserve replicated ETag properly on target

poornas:
Fixes regression from PR's #15674, #15929, #13171 and #15346

## Motivation and Context
replication ETag passed in x-minio-source-etag header used to be passed into opts.UserDefined as "etag" key for PutObject and NewMultipartUpload. If such an ETag is present in the metadata PutObject and CompleteMultipart should preserve it from the replication source.

 PR's above brought in changes that broke this and could lead to re-replication attempts if ETag is not matching b/w clusters 

## How to test this PR?
Use a replicated setup with encryption enabled.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/erasure-multipart.go
+++ b/cmd/erasure-multipart.go
@@ -350,7 +350,7 @@ func (er erasureObjects) newMultipartUpload(ctx context.Context, bucket string, 
 		rctx := lkctx.Context()
 		obj, err := er.getObjectInfo(rctx, bucket, object, opts)
 		lk.RUnlock(lkctx.Cancel)
-		if err != nil {
+		if err != nil && !isErrVersionNotFound(err) {
 			return nil, err
 		}
 		if opts.CheckPrecondFn(obj) {
@@ -359,7 +359,9 @@ func (er erasureObjects) newMultipartUpload(ctx context.Context, bucket string, 
 	}
 
 	userDefined := cloneMSS(opts.UserDefined)
-
+	if opts.PreserveETag != "" {
+		userDefined["etag"] = opts.PreserveETag
+	}
 	onlineDisks := er.getDisks()
 	parityDrives := globalStorageClass.GetParityForSC(userDefined[xhttp.AmzStorageClass])
 	if parityDrives < 0 {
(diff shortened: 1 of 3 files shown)


minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

cmd/erasure-multipart.go
```diff
@@ -1151,9 +1153,13 @@ func (er erasureObjects) CompleteMultipartUpload(ctx context.Context, bucket str
...
-	fi.Metadata["etag"] = opts.UserDefined["etag"]
+	// for replica, newMultipartUpload would have already sent the replication ETag
 	if fi.Metadata["etag"] == "" {
-		fi.Metadata["etag"] = getCompleteMultipartMD5(parts)
+		if opts.UserDefined["etag"] != "" {
+			fi.Metadata["etag"] = opts.UserDefined["etag"]
+		} else { // fallback if not already calculated in handler.
+			fi.Metadata["etag"] = getCompleteMultipartMD5(parts)
+		}
 	}
```
harshavardhana, line 1163: unable to understand what does this logically change?

cmd/erasure-object.go
```diff
@@ -943,7 +943,7 @@ func (er erasureObjects) putObject(ctx context.Context, bucket string, object st
 
 	if opts.CheckPrecondFn != nil {
 		obj, err := er.getObjectInfo(ctx, bucket, object, opts)
-		if err != nil {
+		if err != nil && !isErrVersionNotFound(err) {
```
harshavardhana, line 946: shouldn't the caller be ignoring this? not here?

harshavardhana (changes requested):

cmd/erasure-object.go
poornas: ETag was not being passed into the precondition match correctly from opts in #15674, so it worked previously for the wrong reason (precondition function was nil and this match was never tested) - when a replica PUT event is received, the version will not exist and should be treated as ok

cmd/erasure-multipart.go
poornas: for replication, the fi.Metadata already has the replication ETag passed in NewMultipartUpload and should not be overridden if it is set. 
https://github.com/minio/minio/pull/15346 was setting "etag" in in opts.UserDefined at the handler and passing it down to object layer, clobbering what was set in fi.Metadata already for replication

harshavardhana (approved):

PR #15985, opened 2022-11-01
Use the same lock for the scanner and site replication healing

vadmeste:
## Description
Reduce locking refresh chattiness by using only one lock called .minio.sys/leader 
that can be used by both data scanner and site replication healing.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Optimization (provides speedup with no functional changes)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [ ] Fixes a regression (If yes, please add `commit-id` or `PR #` here)
- [ ] Unit tests added/updated
- [ ] Internal documentation updated
- [ ] Create a documentation update request [here](https://github.com/minio/docs/issues/new?label=doc-change,title=Doc+Updated+Needed+For+PR+github.com%2fminio%2fminio%2fpull%2fNNNNN)

PR #16169, opened 2022-12-05
Avoid hot loop when lock is cancelled

klauspost:
## Description

If `lkctx.Context().Done()` is triggered we do not exit the loop.

Instead it becomes a hot loop.

## Types of changes
- [x] Fixes a regression #15985

### DIFF
--- a/cmd/shared-lock.go
+++ b/cmd/shared-lock.go
@@ -40,6 +40,7 @@ func (ld sharedLock) backgroundRoutine(ctx context.Context, objAPI ObjectLayer, 
 			continue
 		}
 
+	keepLock:
 		for {
 			select {
 			case <-ctx.Done():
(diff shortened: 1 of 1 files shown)


vadmeste (approved):
LGTM!

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

harshavardhana (approved):

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

cmd/shared-lock.go
```diff
@@ -40,6 +40,7 @@ func (ld sharedLock) backgroundRoutine(ctx context.Context, objAPI ObjectLayer,
 			continue
 		}
 
+	keepLock:
```
krisis, line 43: Shouldn't we break out to the beginning of the outer for loop, so that we can reacquire the shared lock? We could still get into the same hot loop, since `<-lkctx.Context().Done()` won't block unless the shared lock is reacquired.

harshavardhana: @vadmeste ^^

vadmeste: > Shouldn't we break out to the beginning of the outer for loop, so that we can reacquire the shared lock?

I think this PR is doing exactly that @krisis

krisis: You're right @vadmeste. I had incorrectly assumed that `break` statement worked similar to `goto`.

PR #16262, opened 2022-12-15
lock: Use context with timeout for acquiring lock calls

vadmeste:
## Description
The code is using context.Background() when doing lock calls, 
it does not timeout.

Also add acquiring lock timeout, probably useful in a hyper 
loaded cluster.

## Motivation and Context
Add timeout to lock acquiring

## How to test this PR?
Not trivial

## Types of changes

## Checklist:

### DIFF
--- a/internal/dsync/drwmutex.go
+++ b/internal/dsync/drwmutex.go
@@ -45,7 +45,7 @@ func log(format string, data ...interface{}) {
 
 const (
 	// dRWMutexAcquireTimeout - default tolerance limit to wait for lock acquisition before.
-	drwMutexAcquireTimeout = 1 * time.Second // 1 second.
+	drwMutexAcquireTimeout = 5 * time.Second
 
 	// dRWMutexRefreshTimeout - default timeout for the refresh call
 	drwMutexRefreshCallTimeout = 5 * time.Second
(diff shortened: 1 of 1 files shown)


internal/dsync/drwmutex.go
```diff
@@ -430,11 +430,11 @@ func lock(ctx context.Context, ds *Dsync, locks *[]string, id, source string, is
 			var locked bool
 			var err error
 			if isReadLock {
-				if locked, err = c.RLock(context.Background(), args); err != nil {
+				if locked, err = c.RLock(ctx, args); err != nil {
```
klauspost, line 433: The original motivation is that we get to a more "known" state, meaning we will either 

A) Get the lock (and we know we have to release it) or
B) We don't get the lock and we don't have to release it.

Adding a canceled request into the mix will make it more likely that the remote server actually grants the lock, but the response is never sent back.

So as far as I can tell having a cancelable context here will make it more likely we leak the lock. 

Leaking locks is not impossible with `context.Background()`, but it is less likely.

vadmeste: Locks won't leak. There is a refresh mechanism that will clean up locks that the lock owner cleared internally or when a lock does not have a quorum anymore.

I would prefer to return an error rather than waiting indefinitely, it is easier to troubleshoot and understand what is going on.

klauspost: @vadmeste We shouldn't rely more on cleanup than we have to. It will just create random delays.

We are not waiting indefinitely - this is not a blocking loop. It is async. The loop below is what is blocking, and it returns when the context is canceled.

harshavardhana: context.Background() is needed to ensure that we are not leaking locks and state() elsewhere purposefully.

HTTP contexts are notorious and they can any time disconnect so relying on refresh() is not a good idea.

I think what we have right now is sufficient, it is not the problem of contexts() that is the issue in our dsync/lock.

We would have to re-do the implementation with a different kind of design to be more optimal at the network layer.

PR #16280, opened 2022-12-20
serialize replication stats() only when needed

harshavardhana:
## Description
serialize replication stats() only when needed

## Motivation and Context
also, rename '.minio.sys/buckets/replication' to
   '.minio.sys/buckets/.replication'

## How to test this PR?
Just test by running on a fresh path, create
buckets and let the usage scanner run once. 

Wait for 5mins and observe if any data is 
written at `.minio.sys/buckets/replication/<hostname>.stats/` 

If replication is not configured we do not need to
serialize anything.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/bucket-replication-utils.go
+++ b/cmd/bucket-replication-utils.go
@@ -638,7 +638,7 @@ type replicationResyncer struct {
 }
 
 const (
-	replicationDir      = "replication"
+	replicationDir      = ".replication"
 	resyncFileName      = "resync.bin"
 	resyncMetaFormat    = 1
 	resyncMetaVersionV1 = 1
--- a/cmd/bucket-replication-stats.go
+++ b/cmd/bucket-replication-stats.go
@@ -254,11 +254,10 @@ func (r *ReplicationStats) serializeStats() ([]byte, error) {
 	}
 	r.mostRecentStatsMu.Lock()
 	defer r.mostRecentStatsMu.Unlock()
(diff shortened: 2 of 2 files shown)


cmd/bucket-replication-utils.go
```diff
@@ -638,7 +638,7 @@ type replicationResyncer struct {
 }
 
 const (
-	replicationDir      = "replication"
```
poornas, line 641: better to make this change as a separate PR - both mrf and stats use this directory. It needs to be backwards compatible to move the stats and mrf files to new path

poornas (changes requested):

cmd/bucket-replication-utils.go
harshavardhana: No we don't need to do that @poornas - I think these stats can be completely discarded, they will be regenerated soon enough via scanner.

poornas (approved):

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

klauspost (approved):

klauspost:
@harshavardhana This will drop stats on upgrade, right? I assume you are ok with that.

harshavardhana:
> @harshavardhana This will drop stats on upgrade, right? I assume you are ok with that.

Yes @klauspost it will be re-populated via scanner in 5mins - something we can live with.

PR #16431, opened 2023-01-18
Use async healing in PutObject call

vadmeste:
## Description
PutObject can heal up to two versions when 
it finds a disparency between the list of versions, 
however the heal object call is sync and holds 
a lock as well. Call healing in a goroutine instead.

Also, do not account disks returning an error (offline for example) when
trying to detect if there is a versions disparency for an object during
PutObject.

## Motivation and Context
Fix a successful but long 'mc licence update' call seen by a user.

## How to test this PR?
Hard to test

## Types of changes

## Checklist:

### DIFF
--- a/cmd/erasure-object.go
+++ b/cmd/erasure-object.go
@@ -773,7 +773,10 @@ func renameData(ctx context.Context, disks []StorageAPI, srcBucket, srcEntry str
 	err := reduceWriteQuorumErrs(ctx, errs, objectOpIgnoredErrs, writeQuorum)
 	if err == nil {
 		versions := reduceCommonVersions(diskVersions, writeQuorum)
-		for _, dversions := range diskVersions {
+		for index, dversions := range diskVersions {
+			if errs[index] != nil {
+				continue
+			}
 			if versions != dversions {
 				versionsDisparity = true
 				break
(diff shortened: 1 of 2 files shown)


klauspost (commented):
The change makes sense.

This can potentially spawn a lot of resource intensive async operations. Isn't there a queue we can add it to (and maybe even spill if too long)?

harshavardhana:
IMO we should simply call queueHeal here instead of blocking healObject()

harshavardhana:
IMO we should simply call queueHeal here instead of blocking healObject()

harshavardhana:
- In-fact `healObject()` must be non-blocking already not sure why it would block
- Why would this be called for `config.json` ? there are no multiple versions on config.json?

klauspost:
@harshavardhana 

> In-fact healObject() must be non-blocking already not sure why it would block

Sometimes we need the result. For example when scanning. If a folder was missing, we send it to healing and wait for the result, so we know if we should scan it now.

Sometimes we need async, sometimes not. Having an async option would definitely be a plus.

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

harshavardhana (approved):

harshavardhana:
#16433 should help in other things

PR #16484, opened 2023-01-26
Signed trailers for signature v4

klauspost:
## Motivation and Context

Provide streaming trailer support to have content hash as trailers in PutObject and multipart uploads.

Both signed (HTTP) and unsigned (HTTPS) trailers are supported.

This makes S3 compatibility stricter, and we no longer accept clients that send more content than the indicated Content-Length. This matches both common HTTP and AWS behavior and ensures that Trailing headers are parsed correctly.

## How to test this PR?

Use https://github.com/minio/mc/pull/4297

Functional tests added in https://github.com/minio/minio-go/pull/1817

## Types of changes

## Checklist:


minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

cmd/streaming-v4-unsigned.go
```diff
@@ -0,0 +1,257 @@
+// Copyright (c) 2015-2022 MinIO, Inc.
```
harshavardhana:
```suggestion
// Copyright (c) 2015-2023 MinIO, Inc.
```

cmd/signature-v4-utils.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2015-2021 MinIO, Inc.
+// Copyright (c) 2015-2022 MinIO, Inc.
```
harshavardhana:
```suggestion
// Copyright (c) 2015-2023 MinIO, Inc.
```

internal/ioutil/hardlimitreader.go
```diff
@@ -0,0 +1,37 @@
+package ioutil
```
harshavardhana, line 20: license header missing

harshavardhana (changes requested):

cmd/streaming-v4-unsigned.go
klauspost: Hey, it was correct when I started the PR :D

I will do a final pass before I mark it ready. I want PutObject to also work before that, though.

harshavardhana:
PTAL @donatello @poornas

harshavardhana (approved):

poornas (approved):

cmd/streaming-v4-unsigned.go
```diff
@@ -0,0 +1,257 @@
...
+				cr.err = err
+				return n, cr.err
+			}
+			break
+		}
+
+		// Manually deserialize the size since AWS specified
+		// the chunk size to be of variable width. In particular,
+		// a size of 16 is encoded as `10` while a size of 64 KB
+		// is `10000`.
```
azat, line 121: And actually I tried to use MinIO with `STREAMING-UNSIGNED-PAYLOAD-TRAILER` but failed, so I looked around and wrote the following patch - https://github.com/minio/minio/commit/7d4c0c311c06d296652a035512dd7b22e352d10f

As far as I understand, in case of `STREAMING-UNSIGNED-PAYLOAD-TRAILER`, there are no "aws specified chunks" only regular plain HTTP chunks, so all of this looks superfluous.

Although I'm not sure about checksum part in the patch, apparently the aws-cpp-sdk simply does not support `STREAMING-UNSIGNED-PAYLOAD-TRAILER` properly, since when I tried to send request to the AWS S3, it was complaining:

```
S3Exception name IncompleteBody, Message: Unable to parse ExceptionName: IncompleteBody Message: The request body terminated unexpectedly
```

@klauspost what do you think?

klauspost: @azat 
Our [minio-go](https://github.com/minio/minio-go) implementation is confirmed to work against AWS:

```
PUT /1z5dsknvswcxo92b6sdfajulvwfxh2?partNumber=2&uploadId=mj6EhLRx2c1plArOgolRk3hgCtFQP6408LqkU_JR910Qu.QvFUTJGAwqAUddj3IglD51NgCVNLncjd2WLG6XlkXmFixKz9M1.ifxJagF0pR2d.qQWnEqD_m3M3DkreQW HTTP/1.1
Host: minio-go-test-nu40trprcadk2wmv.s3.dualstack.us-east-1.amazonaws.com
User-Agent: MinIO (windows; amd64) minio-go/v7.0.60 MinIO-go-FunctionalTest/0.1.0
Content-Length: 4889394
Authorization: AWS4-HMAC-SHA256 Credential=**REDACTED**/20231005/us-east-1/s3/aws4_request, SignedHeaders=content-encoding;host;x-amz-content-sha256;x-amz-date;x-amz-decoded-content-length;x-amz-trailer, Signature=**REDACTED**
Content-Encoding: aws-chunked
X-Amz-Content-Sha256: STREAMING-UNSIGNED-PAYLOAD-TRAILER
X-Amz-Date: 20231005T170945Z
X-Amz-Decoded-Content-Length: 4888682
X-Amz-Trailer: x-amz-checksum-crc32c
Accept-Encoding: gzip

HTTP/1.1 200 OK
Content-Length: 0
Date: Thu, 05 Oct 2023 17:09:46 GMT
Etag: "2e17afc721dce2bca8572e594fa4451d"
Server: AmazonS3
X-Amz-Checksum-Crc32c: C/etzg==
X-Amz-Id-2: qYB5u7L3L1T9w18BoUTwYmkhF05TdP6IrVPQxBnvTSp1JiVbaWzQK8Jup7EE2mINcP5CcRFm/DU=
X-Amz-Request-Id: B70CB93P78ZVZBV5
X-Amz-Server-Side-Encryption: AES256
```

klauspost: @azat Since the CPP library fails on AWS, I suspect it doesn't support it.

azat: >Our [minio-go](https://github.com/minio/minio-go) implementation is confirmed to work against AWS:

But minio-go is not the same as minio, right?
Or you are saying that it implements protocol in the same fashion as this implementation in the server, and it send "aws chunks" even for unsigned requests? And it works with AWS S3 and MinIO?

When I tested this `STREAMING-UNSIGNED-PAYLOAD-TRAILER`, using C++ sdk, I got this:
- S3: `Message: Unable to parse ExceptionName: IncompleteBody Message: The request body terminated unexpectedly,`
- GCS: works
- R2 (cloudflare): `Message: Unable to parse ExceptionName: NotImplemented Message: STREAMING-UNSIGNED-PAYLOAD-TRAILER not implemented,`

So maybe the format was wrong indeed, but this means that GCS implemented it not correctly.

P.S. It seems that there is no cli that I can use?
(I'm not a go developer actually, just a needy user)

harshavardhana: minio-go implementation works for both MinIO and AWS S3 - that's what @klauspost meant. 

Meaning our implementation is correct and to the spec.

klauspost: @azat TBH I referenced the documentation and the AWS Java SDK IIRC.

Considering their SDK doesn't work against their own servers, I don't really have much confidence in it. I suggest you discuss that with AWS.

As Harsha mentioned our client implementation works against S3 (and it was pretty strict in what it accepted). It seems reasonable to assume that the matching server implementation is correct as well. I don't really care what GCS does.

jimrobinson:
I saw this branch was merged into minio:master on 2023-05-05, but if it isn't marked as completed.  Does that mean it's still a work in progress?

I ran into some issues with chunk size limitations on my minio instance and I was wondering if they are expected to be a firm limit going forward?  For our minio instance we're actually hoping to encourage our clients to use 5GiB chunks, due to some efficiency concerns we have with our particular post-upload usage requirements.

We can use tools such as aws-cli v2 and configure them to use 5GiB chunks w/o any issue:
```
aws configure --profile test-elm set s3.multipart_chunksize 5GiB
```

With `aws s3 cp` minio is happy to accept uploads, and I can see that the files are created using 5GiB parts.

However, I tried implementing trailing checksum logic into a client that is using the AWS SDK, and ran into an error where my minio instance (RELEASE.2024-07-31T05-46-26Z) wasn't happy with the default chunk sizes being sent, it wanted me to limit them to 16MiB chunks.

I wasn't positive I coding up the client properly, so I tried cloning [amazon-s3-checksum-tool](https://github.com/aws-samples/amazon-s3-checksum-tool), only tweaking it to let me run it against my test minio instance (see the branch [use-path-style](https://github.com/jimrobinson/amazon-s3-checksum-tool/tree/use-path-style) for the changes).

The aws-sample code assumes a default of 64 MiB, which naturally triggers the same sort of rejection.  Once I override the default to use 16MiB minio does accept the upload w/o any issue.

```
$ dd if=/dev/random of=/tmp/x.6 bs=1024M count=6
6+0 records in
6+0 records out
6442450944 bytes (6.4 GB, 6.0 GiB) copied, 25.0648 s, 257 MB/s

$ pwd
/home/jimr/proj/amazon-s3-checksum-tool/cmd/s3checksum

$ go build

$ ./s3checksum upload --file /tmp/x.6 --profile test-elm --bucket jimr --key x.6 --use-path-style
2024/08/01 16:58:31 Beginning upload...
2024/08/01 16:58:32 upload multipart failed, upload id: ZjNmZjgyZGUtOGFmMy00ZGViLTlmMzAtOGY3MDY4N2RlNjcxLjMwMzgzZjhlLTdiMjMtNGM3ZC1iNzA2LTljMmUzZTc0ODBlYngxNzIyNTQ5NTExNjQ4ODM2OTMy, cause: operation error S3: UploadPart, https response error StatusCode: 400, RequestID: 17E7B9A78E01E9BE, HostID: 7e63378ad44bb38bbd22e530765313e809e81cf73e8706de3a00b16c1a63c395, api error BadRequest: chunk too big: choose chunk size <= 16MiB

$ ./s3checksum upload --file /tmp/x.6 --profile test-elm --bucket jimr --key x.6 --use-path-style --chunksize 16
2024/08/01 17:11:19 Beginning upload...
Part: 00001             e4kMqnNr6ClQiTGRJeFHAr6ovMLxyBoWuKLWjAGiqH8=
Part: 00002             6gJidO6gCIj14od1vNy4YLV2+4my9gNvfavneC/W2eM=
...
Part: 00383             AAf0C/ME5d/vWseKy57jNoCN2ZD1p4N9y2BiF3MxE8I=
Part: 00384             TXWch1H2aRr8HW2YCdFwiEQdMCcjdAcJiN9sNV9B9DM=
Amazon S3 SHA256:       rMNibg9WVBT5uBz7eP5Vh4IYKWQUcPMZoRN9ZyE5Qec=-384
Amazon S3 Etag: d5dfd10bb13f38082470e8bb56301154-384
```

harshavardhana:
64MiB chunk size is literally terrible, that kind of malloc is a bit too much.

Never do that, let alone make it 5GiB.

Remember don't confuse this chunk with multipart part size.

This causes large allocation on client and as well as server. Checksums require that you validate them so you need that much buffer as big as your chunk is.

Most efficient would be for trailer checksums to be 4MiB.

jimrobinson:
> don't confuse this chunk with multipart part size

Ah, that's a very interesting correction!  I was using PartSIze and chunk size as interchangeable terms, but your comment indicates that's not correct. My understanding up to this point was that a multi-part object might look like this on the minio filesystem:

```
/<bucket>/<key>/<version>/part.<n>
```

Where `part.<n>` may range in size from 5 MiB to 5 GiB, with the last `part.<n>` in the set allowed to be < 5 MiB.  Am I correct to think this `part.<n>` file size limit is what you are referring to as _multipart part size_?

It sounds from your comment as though _chunksize_ is distinct from this `part.<n>` size, and is something I can adjust separately?  For example, like in HTTP chunked encoding?   If that's correct,  would you be able to point me at some documentation, or tell me some nomenclature I should have been aware of?

When I read https://repost.aws/knowledge-center/s3-upload-large-files it explained the aws-cli settings as including:

**multipart_chunksize** _This value sets the size of each part that AWS CLI uploads in a multipart upload for an individual file. This setting allows you to break down a larger file (for example, 300 MB) into smaller parts for quicker upload speeds. The default value is 8 MB._
**Note**: _A multipart upload requires that a single file is uploaded in fewer than 10,000 distinct parts. Be sure that the chunksize that you set balances the part file size and the number of parts._

Looking at that [aws-samples/amazon-s3-checksum-tool](https://github.com/aws-samples/amazon-s3-checksum-tool/tree/main) code, the command line --chunksize is used to directly set an option variable PartSize

https://github.com/aws-samples/amazon-s3-checksum-tool/blob/25a77aa34a68a93e3f2c1a988befbf30d2bdf83e/cmd/s3checksum/main.go#L80

```
mpf, err := s3checksum.NewMultipartFile(s3checksum.MultipartFileOpts{
	FilePath:         file,
	ManifestFilePath: manifestFile,
	PartSize:         chunksize * 1024 * 1024,
	Threads:          threads,
})
```

This option is then passed to the AWS SDK api as a PartSize argument to the s3 manager Uploader:

https://github.com/aws-samples/amazon-s3-checksum-tool/blob/25a77aa34a68a93e3f2c1a988befbf30d2bdf83e/uploader.go#L55

```
uploader := manager.NewUploader(client, func(u *manager.Uploader) {
	u.PartSize = opts.PartSize
	u.Concurrency = opts.NumRoutines
})
```

So those had led me to assume "chunksize == PartSize". 

The use of PartSize of 64 MiB seems to be explicitly used as an example in the S3 SDK documentation:

https://github.com/aws/aws-sdk-go-v2/blob/a2747951bd66c0b687a0a483195843b73f6448bb/feature/s3/manager/upload.go#L249-L271

The uploader code also seems to work off the PartSize to determine how it builds its upload requests:

- https://github.com/aws/aws-sdk-go-v2/blob/a2747951bd66c0b687a0a483195843b73f6448bb/feature/s3/manager/upload.go#L426-L476
- https://github.com/aws/aws-sdk-go-v2/blob/a2747951bd66c0b687a0a483195843b73f6448bb/feature/s3/manager/upload.go#L695-L753

harshavardhana:
yes but that part and chunk size for signature are different.

jimrobinson:
> yes but that part and chunk size for signature are different.

Great, it's just that I'm failing to spot what knobs I can turn in the aws sdk to allow for 16 MiB _chunksize_ to keep minio happy and still keep 5 GiB `part.<n>`  _multipart part size_.   Can you point out what I need to look at?  Or is this something I should be attempting to do using the minio sdk nstead of the aws sdk?

Edit: As a follow-up, I tested whether or not things worked out-of-the-box with AWS S3, and it does.  I can specify a 5 GiB multipart part size and the expected number of parts are created in S3 w/ their checksums.  The wording throughout [S3 Compatible Storage](https://min.io/product/s3-compatibility) made me assume I could just use the AWS SDK as a drop-in solution for interacting with MinIO, but that's clearly not the case here.

I'll experiment with the MinIO SDK to see if I can get it to work the way we need.

```
# amazon-s3-checksum-tool (use-path-style =)
$ ./s3checksum upload --file /tmp/x.6 --profile srcc-jrobinso --bucket test-trailing-checksums --key x.6 --chunksize 5120
2024/08/02 12:12:48 Beginning upload...
Part: 00001             wFnK7Pbe9FhsSq0qHA66dhOIj9dWi4MKpA8sWutpbRI=
Part: 00002             97bJw4Q08aCSBOeR5s2WacaBR3+Mj/ia/E4mEdw2Uqc=
Amazon S3 SHA256:       RkHEB/J/3X4V7Kx8VnId/wWEOU0QllxZWZymOpyPp+Y=-2
Amazon S3 Etag: 66ee6d9a5c13943747057da7270090b5-2
# amazon-s3-checksum-tool (use-path-style =)
$ aws --profile srcc-jrobinso s3api get-object-attributes --bucket test-trailing-checksums --key x.6 --object-attributes ETag Checksum ObjectParts StorageClass ObjectSize
{
    "LastModified": "2024-08-02T17:12:49+00:00",
    "ETag": "66ee6d9a5c13943747057da7270090b5-2",
    "Checksum": {
        "ChecksumSHA256": "RkHEB/J/3X4V7Kx8VnId/wWEOU0QllxZWZymOpyPp+Y="
    },
    "ObjectParts": {
        "TotalPartsCount": 2,
        "PartNumberMarker": 0,
        "NextPartNumberMarker": 2,
        "MaxParts": 1000,
        "IsTruncated": false,
        "Parts": [
            {
                "PartNumber": 1,
                "Size": 5368709120,
                "ChecksumSHA256": "wFnK7Pbe9FhsSq0qHA66dhOIj9dWi4MKpA8sWutpbRI="
            },
            {
                "PartNumber": 2,
                "Size": 1073741824,
                "ChecksumSHA256": "97bJw4Q08aCSBOeR5s2WacaBR3+Mj/ia/E4mEdw2Uqc="
            }
        ]
    },
    "StorageClass": "STANDARD",
    "ObjectSize": 6442450944
}
```

klauspost:
Your arguments are going in circles. It seems like AWS is mixing parts and chunks in their tool. Chunks is an encoding scheme for [chunked streaming uploads](https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming.html). It has no relation to part sizes and is purely an internal encoding.

Don't use stupid big chunks. Having no chunk limit is opening your server to DOS. In fact a chunk size of something like 32/64KB is probably the most optimal, since chunks can remain in L1/L2 cache on both client and server.

jimrobinson:
@klauspost Thank you, that is the documentation pointer I had been hoping harshavardhana would provide.   I don't think it's fair to say I'm arguing in circles, I think I pretty clearly acknowledged that I was being told the two were distinct, I was asking for pointers on details.

PR #16486, opened 2023-01-26
fix: do not return IAM/Bucket metadata replication errors to client

harshavardhana:
## Description
fix: do not return IAM/Bucket metadata replication errors to client

## Motivation and Context
This is a change to ensure that we behave like object() replication, 
we are only going to log the error instead.

It is expected that customers/users always perform a job, or a 
workload only on one "site" at any given point in time.

Even if there is a GSLB in front of all replicated sites, it is 
expected that applications perform operations on a single site 
at any given point in time.

The intention is we shouldn't be reporting errors to clients 
for transient network errors, clients have no way to interpret 
these server side internal errors.

## How to test this PR?
Nothing new to test, this is what we agreed after discussing with @abperiasamy
will be sending another change regarding reduction in logs.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/bucket-targets.go
+++ b/cmd/bucket-targets.go
@@ -98,7 +98,7 @@ func (sys *BucketTargetSys) heartBeat(ctx context.Context) {
 		select {
 		case <-hcTimer.C:
 			sys.hMutex.RLock()
-			var eps []madmin.ServerProperties
+			eps := make([]madmin.ServerProperties, 0, len(sys.hc))
 			for _, ep := range sys.hc {
 				eps = append(eps, madmin.ServerProperties{Endpoint: ep.Endpoint, Scheme: ep.Scheme})
 			}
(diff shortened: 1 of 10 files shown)


minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

cmd/bucket-targets.go
```diff
@@ -98,24 +98,20 @@ func (sys *BucketTargetSys) heartBeat(ctx context.Context) {
 		select {
 		case <-hcTimer.C:
 			sys.hMutex.RLock()
-			var eps []madmin.ServerProperties
+			eps := make([]madmin.ServerProperties, 0, len(sys.hc))
 			for _, ep := range sys.hc {
 				eps = append(eps, madmin.ServerProperties{Endpoint: ep.Endpoint, Scheme: ep.Scheme})
```
poornas, line 103:
```suggestion
			for i, ep := range sys.hc {
				eps[i] = madmin.ServerProperties{Endpoint: ep.Endpoint, Scheme: ep.Scheme}
```

poornas (changes requested):

cmd/bucket-targets.go
harshavardhana: @poornas notice I am not allocating - I am only indicating the capacity of the slice

> +                       eps := make([]madmin.ServerProperties, 0, len(sys.hc))

It is to let Go know the final length so that it can optimize the slice allocation.

poornas: missed that!

poornas (approved):

krisis (approved):

PR #16504, opened 2023-01-30
Fix truncated list response on deleted replicated objects

klauspost:
## Description

When an object version is deleted it is marked as a "FreeTierVersion" pending replication of the delete.

These will be excluded from listing results since they are deleted, but pending replication.

The issue is that when 1000 keys are requested, we request 1000 keys from each disk, but don't expect all of them to be filtered out. Since all are filtered out, the listing returns as if it has completed.

Do not count all free versions as a returned object when listing with limits.
Also filter objects that are all free versions earlier and not only when converting individual versions to responses.

## How to test this PR?

Requires thousands of un-replicated objects. If all sets only return `TierFreeVersion` objects the response will be truncated.

## Types of changes

### DIFF
--- a/cmd/metacache-entries.go
+++ b/cmd/metacache-entries.go
@@ -189,6 +189,43 @@ func (e *metaCacheEntry) isLatestDeletemarker() bool {
 	return xlMeta.versions[0].header.Type == DeleteType
 }
 
+// isAllFreeVersions returns if all objects are free versions.
+// If metadata is NOT versioned false will always be returned.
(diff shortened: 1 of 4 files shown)


minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

harshavardhana (approved):

poornas (approved):

krisis (approved):
LGTM, a little late to the party :)

PR #16541, opened 2023-02-06
feat: allow offline disks on a fresh start

harshavardhana:
## Description
feat: allow offline disks on a fresh start

## Motivation and Context
We never used to allow offline disks during fresh
setup, but that is not necessary anymore. The reason 
being we expect the order of disks to be exactly the 
same once started.

This new change simply allows starting MinIO with
some offline disks, it still refuses to start if we 
do not have enough disks online, i.e

quorum=(len(disks)/2+1)

## How to test this PR?
Nothing special start with a setup like this

```
~ mkdir -p /tmp/xl{1..4}
~ sudo chown -R root. /tmp/xl2
~ CI=true minio server /tmp/xl{1...4}
```

## Types of changes

## Checklist:

### DIFF
--- a/cmd/prepare-storage.go
+++ b/cmd/prepare-storage.go
@@ -198,8 +198,15 @@ func connectLoadInitFormats(verboseLogging bool, firstDisk bool, endpoints Endpo
 		return nil, nil, err
 	}
 
+	// Return error when quorum unformatted disks - indicating we are
+	// waiting for first server to be online.
+	unformattedDisks := quorumUnformattedDisks(sErrs)
+	if unformattedDisks && !firstDisk {
+		return nil, nil, errNotFirstDisk
+	}
+
 	// All disks report unformatted we should initialized everyone.
-	if shouldInitErasureDisks(sErrs) && firstDisk {
+	if unformattedDisks && firstDisk {
 		logger.Info("Formatting %s pool, %v set(s), %v drives per set.",
 			humanize.Ordinal(poolCount), setCount, setDriveCount)
 
@@ -213,19 +220,6 @@ func connectLoadInitFormats(verboseLogging bool, firstDisk bool, endpoints Endpo
 		// minio server managing the first disk
 		globalDeploymentID = format.ID
 		return storageDisks, format, nil
-	}
-
-	// Return error when quorum unformatted disks - indicating we are
-	// waiting for first server to be online.
-	unformattedDisks := quorumUnformattedDisks(sErrs)
-	if unformattedDisks && !firstDisk {
-		return nil, nil, errNotFirstDisk
-	}
-
-	// Return error when quorum unformatted disks but waiting for rest
-	// of the servers to be online.
-	if unformattedDisks && firstDisk {
-		return nil, nil, errFirstDiskWait
 	}
 
 	// Mark all root disks down
--- a/cmd/format-erasure.go
+++ b/cmd/format-erasure.go
@@ -304,11 +304,6 @@ func countErrs(errs []error, err error) int {
 	return i
 }
 
-// Does all errors indicate we need to initialize all disks?.
-func shouldInitErasureDisks(errs []error) bool {
-	return countErrs(errs, errUnformattedDisk) == len(errs)
-}
-
 // Check if unformatted disks are equal to write quorum.
 func quorumUnformattedDisks(errs []error) bool {
 	return countErrs(errs, errUnformattedDisk) >= (len(errs)/2)+1
@@ -750,6 +745,9 @@ func initFormatErasure(ctx context.Context, storageDisks []StorageAPI, setCount,
 		hostCount := make(map[string]int, setDriveCount)
 		for j := 0; j < setDriveCount; j++ {
 			disk := storageDisks[i*setDriveCount+j]
+			if disk == nil {
+				continue
+			}
 			newFormat := format.Clone()
 			newFormat.Erasure.This = format.Erasure.Sets[i][j]
 			if distributionAlgo != "" {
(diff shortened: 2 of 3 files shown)


minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

cmd/prepare-storage.go
```diff
@@ -198,8 +198,15 @@ func connectLoadInitFormats(verboseLogging bool, firstDisk bool, endpoints Endpo
 		return nil, nil, err
 	}
 
+	// Return error when quorum unformatted disks - indicating we are
+	// waiting for first server to be online.
+	unformattedDisks := quorumUnformattedDisks(sErrs)
```
vadmeste, line 203: This actually means many of new fresh setups will have many unformatted disks at the first.. healing will detect that in 10 seconds and will reformat again but it will print a lot of unnecessary logs etc.. maybe let's ignore the offline disk after 2 minutes ? (time.Since(globalBootTime) > 2 *time.Minute)

vadmeste (commented):
One comment

cmd/prepare-storage.go
harshavardhana: where would I add that? @vadmeste

vadmeste: maybe something like below, again the only point is to generate less logs which can be confusing for users.

```go
diff --git a/cmd/format-erasure.go b/cmd/format-erasure.go
index d535039b8..662887d10 100644
--- a/cmd/format-erasure.go
+++ b/cmd/format-erasure.go
@@ -305,8 +305,9 @@ func countErrs(errs []error, err error) int {
 }
 
 // Check if unformatted disks are equal to write quorum.
-func quorumUnformattedDisks(errs []error) bool {
-       return countErrs(errs, errUnformattedDisk) >= (len(errs)/2)+1
+func quorumUnformattedDisks(errs []error) (bool, int) {
+       unformattedFound := countErrs(errs, errUnformattedDisk)
+       return unformattedFound >= (len(errs)/2)+1, unformattedFound
 }
 
 // loadFormatErasureAll - load all format config from all input disks in parallel.
diff --git a/cmd/prepare-storage.go b/cmd/prepare-storage.go
index 0b6c5dd2f..522328fd7 100644
--- a/cmd/prepare-storage.go
+++ b/cmd/prepare-storage.go
@@ -200,13 +200,13 @@ func connectLoadInitFormats(verboseLogging bool, firstDisk bool, endpoints Endpo
 
        // Return error when quorum unformatted disks - indicating we are
        // waiting for first server to be online.
-       unformattedDisks := quorumUnformattedDisks(sErrs)
-       if unformattedDisks && !firstDisk {
+       quorumUnformatted, unformattedFound := quorumUnformattedDisks(sErrs)
+       if quorumUnformatted && !firstDisk {
                return nil, nil, errNotFirstDisk
        }
 
        // All disks report unformatted we should initialized everyone.
-       if unformattedDisks && firstDisk {
+       if quorumUnformatted && firstDisk && (unformattedFound == len(storageDisks) || time.Since(globalBootTime) > 2*time.Minute) {
                logger.Info("Formatting %s pool, %v set(s), %v drives per set.",
                        humanize.Ordinal(poolCount), setCount, setDriveCount)
 

```

harshavardhana:
@vadmeste your change doesn't reduce any logs, it simply takes 2 minutes to start fresh. I would avoid adding such an unpredictable code.

Reverting it.

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

vadmeste:
> @vadmeste your change doesn't reduce any logs, it simply takes 2 minutes to start fresh. I would avoid adding such an unpredictable code.

What is unpredictable in it ? To reproduce this, just add a sleep of 15 seconds or even after each node start. You will see some healing already going on.

harshavardhana:
> What is unpredictable in it ? To reproduce this, just add a sleep of 15 seconds or even after each node start. You will see some healing already going on.

How the disk is offline @vadmeste there won't be any heal triggered, heal is triggered when the drive is replaced.

vadmeste:
> How the disk is offline @vadmeste there won't be any heal triggered, heal is triggered when the drive is replaced.

The disk is not offline, just unformatted. The disk is unformatted because the first node decided to format before waiting for all nodes to be up though the quorum is there

harshavardhana:
> > How the disk is offline @vadmeste there won't be any heal triggered, heal is triggered when the drive is replaced.
> 
> The disk is not offline, just unformatted. The disk is unformatted because the first node decided to format before waiting for all nodes to be up though the quorum is there

Why is that a big deal? @vadmeste why do we care ? Let them all come online as they place.  That is the whole reason this PR exists.

What you suggested has a problem, it simply keeps waiting up-to 2mins and it looks weird from a customer point of view.

The current PR is what is expected, it can heal as the drives come online as they please.

vadmeste:
> What you suggested has a problem, it simply keeps waiting up-to 2mins and it looks weird from a customer point of view.

Are you saying that with the suggested change, it will always wait for 2 minutes? that is only true if you have offline disks. This doesn't increase any waiting time beyond what we have already right now.

By yes, it is not a big deal, just wanted to make sure we are on the same page.

harshavardhana:
> Are you saying that with the suggested change, it will always wait for 2 minutes? that is only true if you have offline disks. This doesn't increase any waiting time beyond what we have already right now.

Correct because the main purpose is offline disks @vadmeste all other scenarios are business as usual - nothing should change for them. 

Even if they were offline while starting up, we can be sure to heal them which is okay as we have no idea if it was a legitimate startup sequence delay or indeed the node was offline.

harshavardhana:
PTAL @vadmeste with new assumptions.

vadmeste (approved):
LGTM

klauspost (approved):
LGTM

PR #16813, opened 2023-03-15
fix: No need to traverse when the jobID is not empty

jiuker:
## Description

No need to traverse when the jobID is not empty.

## Types of changes

## Checklist:


cmd/batch-handlers.go
```diff
@@ -1212,12 +1212,39 @@ func batchReplicationTrace(d batchReplicationMetric, job string, startTime time.
 	}
 }
 
-func (m *batchJobMetrics) report(jobID string) (metrics *madmin.BatchJobMetrics) {
-	metrics = &madmin.BatchJobMetrics{CollectedAt: time.Now(), Jobs: make(map[string]madmin.JobMetric)}
+func (m *batchJobMetrics) report(jobID string) *madmin.BatchJobMetrics {
```
harshavardhana: it should be rewritten as `

```
if jobID == "" {
    return &madmin.BatchJobMetrics{CollectedAt: time.Now(), Jobs: make(map[string]madmin.JobMetric)}
} // and then the rest..
```

harshavardhana (changes requested):

jiuker:
@harshavardhana It seems the jobID is empty will return all the data before this pr.The change you suggested is right?

cmd/batch-handlers.go
```diff
@@ -1212,12 +1212,39 @@ func batchReplicationTrace(d batchReplicationMetric, job string, startTime time.
...
+						Object:           job.Object,
+						Objects:          job.Objects,
+						ObjectsFailed:    job.ObjectsFailed,
+						BytesTransferred: job.BytesTransferred,
+						BytesFailed:      job.BytesFailed,
+					},
+				},
+			},
+		}
+	}
```
harshavardhana: the reason it was intentionally not written this way is because of the amount of duplicate code here. 

that's why traversing `m.metrics` was okay because that is a cheap operation.

harshavardhana (changes requested):

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

jiuker:
@harshavardhana what can I do for this PR?

cmd/batch-handlers.go
```diff
@@ -1267,33 +1267,37 @@ func batchReplicationTrace(d batchReplicationMetric, job string, startTime time.
...
-			Complete:      job.Complete,
-			Failed:        job.Failed,
-			Replicate: &madmin.ReplicateInfo{
-				Bucket:           job.Bucket,
-				Object:           job.Object,
-				Objects:          job.Objects,
-				ObjectsFailed:    job.ObjectsFailed,
-				BytesTransferred: job.BytesTransferred,
-				BytesFailed:      job.BytesFailed,
+	if jobID == "" {
```
klauspost, line 1272: If `jobID == ""` previously it would return *all* entries, unless I am misreading.

cmd/batch-handlers.go
```diff
@@ -1267,33 +1267,37 @@ func batchReplicationTrace(d batchReplicationMetric, job string, startTime time.
...
+		return &madmin.BatchJobMetrics{CollectedAt: time.Now(), Jobs: make(map[string]madmin.JobMetric)}
+	}
+	job, ok := m.metrics[jobID]
+	if !ok {
+		return nil
```
klauspost, line 1277: Previously, if jobID wasn't found it would return *all* entries. Not sure if that is on purpose.

cmd/batch-handlers.go
```diff
@@ -1267,33 +1267,37 @@ func batchReplicationTrace(d batchReplicationMetric, job string, startTime time.
...
+	}
+	return &madmin.BatchJobMetrics{
```
klauspost, line 1279: Before it would (randomly) return all entries until it found `jobID`. This seems like a bug in the original code.

klauspost (commented):
There are some significant changes that we need to work out before this can be merged.

harshavardhana:
This PR is not correct, closing. Will resubmit this correctly.

PR #16852, opened 2023-03-20
fix: convert '\' to '/' on windows

harshavardhana:
## Description
fix: convert '\' to '/' on windows

## Motivation and Context
This is to ensure on windows we avoid
arbitrary path escalation during PUT
operations.

## How to test this PR?
Needs a windows machine to test

## Types of changes

## Checklist:

### DIFF
--- a/cmd/generic-handlers.go
+++ b/cmd/generic-handlers.go
@@ -22,6 +22,7 @@ import (
 	"net"
 	"net/http"
 	"path"
+	"path/filepath"
 	"runtime/debug"
 	"strings"
 	"sync/atomic"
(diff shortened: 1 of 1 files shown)


donatello (approved):

minio-trusted:
## Mint Automation
| Test | Result |
|:--- |:--- |
Deleting image on docker hub
Deleting image locally

kannappanr (approved):

klauspost:
@harshavardhana @kannappanr We should just reject objects with `\` - at least just on Windows. TBH I assumed we already did.

klauspost:
If we accept it, it will be a constant duality between `/` and `\`, which we shouldn't have. We should only allow `/`

harshavardhana:
> If we accept it, it will be a constant duality between `/` and `\`, which we shouldn't have. We should only allow `/`

The reason is we have seen keys with `\` on Linux so we can't blindly disallow it, however, we could reject the `\` key for windows separately.

klauspost:
> however, we could reject the \ key for windows separately.

Yeah, that was my thought. Otherwise  `\` gets converted to `/` on Windows, which IMO is worse than rejecting it from the start.

harshavardhana:
> Yeah, that was my thought. Otherwise `\` gets converted to `/` on Windows, which IMO is worse than rejecting it from the start.

It won't get converted, it's only validated as a check to look for `..` @klauspost - so it will reject the PUT on windows with `\..` - however outrightly rejected situations like `\/hello` is also needed on Windows. 

Sent a PR #16856

PR #16874, opened 2023-03-22
return quorum error instead of insufficient storage error

vadmeste:
## Description
PutObject calculates available disk space from cached disk info. 
However an insufficient storage error (507 status code) can be 
returned when there is no enough good information from those disks.

Return quorum error instead of insufficient storage error and add
 a log to facilitate debugging issues.

## Motivation and Context
Avoid returning 507 insufficient storage error when it should be 503

## How to test this PR?
Not trivial without modifying some code

## Types of changes

## Checklist:

### DIFF
--- a/cmd/object-api-utils.go
+++ b/cmd/object-api-utils.go
@@ -1061,7 +1061,7 @@ func getDiskInfos(ctx context.Context, disks ...StorageAPI) []*DiskInfo {
 }
 
 // hasSpaceFor returns whether the disks in `di` have space for and object of a given size.
-func hasSpaceFor(di []*DiskInfo, size int64) bool {
+func hasSpaceFor(di []*DiskInfo, size int64) (bool, error) {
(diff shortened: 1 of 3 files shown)


klauspost (approved):
Beautiful

cmd/object-api-utils.go
```diff
@@ -1082,8 +1082,8 @@ func hasSpaceFor(di []*DiskInfo, size int64) bool {
 		available += disk.Total - disk.Used
 	}
 
-	if nDisks == 0 {
-		return false
+	if nDisks < len(di)/2 {
+		return false, fmt.Errorf("no enough online disks to calculate the available space")
```
```suggestion
		return false, fmt.Errorf("not enough online disks to calculate the available space expected(%d)/(%d)", (len(di)/2)+1, nDisks)
```

harshavardhana (changes requested):

harshavardhana (approved):

PR #16881, opened 2023-03-23
Revert query parameter `src` from diag upload if callhome enabled

shtripat:
## Description
This reverts commit 850a945a1873d4c89690c0a4fae3f23dc607bb5c.

## Motivation and Context
Earlier intention was to set a flag in SUBNET if callhome sends diagnostics data for a cluster and the same was achieved using this query parameter.
But we actually want to make diag and callhome synonymous and SUBNET would rather use the information like whether diag had been upadated in last 24h hrs for a cluster or not. This parameters stands of no use in such case and no need to set explicit flag at cluster level just to identify if callhome enabed or not.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/callhome.go
+++ b/cmd/callhome.go
@@ -168,7 +168,7 @@ func sendHealthInfo(ctx context.Context, healthInfo madmin.HealthInfo) error {
 	}
 
 	filename := fmt.Sprintf("health_%s.json.gz", UTCNow().Format("20060102150405"))
-	url += "?filename=" + filename + "&src=callhome"
+	url += "?filename=" + filename
 
 	_, err := globalSubnetConfig.Upload(url, filename, createHealthJSONGzip(ctx, healthInfo))
 	return err


harshavardhana:
Can you explain a bit more about why are we reverting? and why did we add it in the first place?

anjalshireesh (approved):

shtripat:
> Can you explain a bit more about why are we reverting? and why did we add it in the first place?

Added motivation and context details for the change/reversal.

anjalshireesh:
> Can you explain a bit more about why are we reverting? and why did we add it in the first place?

@harshavardhana  We had added this so that subnet can detect that callhome is enabled on the cluster. After further discussion with @abperiasamy it has been decided that subnet should not differentiate in any way between health diagnostics upload from `mc support diag` and callhome.

harshavardhana (approved):

PR #16887, opened 2023-03-24
Set tcp user timeout to clean sockets with data in the buffer

vadmeste:
## Description

TCP Keep Alive is not enough to kill connections with dead clients. 
The reason is that TCP keep alive does not work when there is data 
in the socket buffer waiting to be transmitted. 

To fix this, TCP_USER_TIMEOUT needs to be set. It kills connections 
with non received ACKs in many TCP lifecycle stages.

## Motivation and Context
Kill existing idle connections

## How to test this PR?
- Setup a cluster and create a bucket and upload a large object
- Use iptables to kill ongoing connection
- Check with 'mc support top locks <alias>'

## Types of changes

## Checklist:


klauspost (approved):
Great stuff.

harshavardhana (approved):

PR #16895, opened 2023-03-27
fix:sort will rand by map

jiuker:
## Description

sort will rand by map
```go
package main

import (
	"fmt"
	"sort"
)

type A struct {
	A string
	T string
}

func main() {
	MA := map[string]A{
		"1231": {
			A: "1231",
			T: "inter",
		},
		"1232": {
			A: "1232",
			T: "inter",
		},
		"1233": {
			A: "1233",
			T: "inter",
		},
		"1234": {
			A: "1234",
			T: "inter",
		},
		"1235": {
			A: "1235",
			T: "",
		},
		"1236": {
			A: "1236",
			T: "",
		},
		"1237": {
			A: "1237",
			T: "",
		},
		"1238": {
			A: "1238",
			T: "",
		},
		"1239": {
			A: "1239",
			T: "",
		},
	}
	sa := []A{}
	for _, v := range MA {
		sa = append(sa, v)
	}
	sort.Slice(sa, func(i, j int) bool {
		//if sa[i].T == "inter" && sa[j].T == "inter" {
		//	return sa[i].A < sa[j].A
		//}
		if sa[i].T == "inter" {
			return true
		}
		if sa[j].T == "inter" {
			return false
		}
		return sa[i].A < sa[j].A
	})
	fmt.Println(sa)
}
```
first output:
```shell
[{1234 inter} {1232 inter} {1233 inter} {1231 inter} {1235 } {1236 } {1237 } {1238 } {1239 }]
```
second output:
```shell
[{1234 inter} {1232 inter} {1231 inter} {1233 inter} {1235 } {1236 } {1237 } {1238 } {1239 }]
```
## Types of changes

## Checklist:

### DIFF
--- a/cmd/data-usage-utils.go
+++ b/cmd/data-usage-utils.go
@@ -131,6 +131,9 @@ func (dui DataUsageInfo) tierStats() []madmin.TierInfo {
 	}
 
 	sort.Slice(infos, func(i, j int) bool {
+		if infos[i].Type == "internal" && infos[j].Type == "internal" {
+			return infos[i].Name < infos[j].Name
+		}
 		if infos[i].Type == "internal" {
 			return true
 		}


klauspost (commented):
It is kind of a no-change, since there will only be one "internal", since TierType cannot be "internal".

jiuker:
> It is kind of a no-change, since there will only be one "internal", since TierType cannot be "internal".
@klauspost This is true by context, but it's more rigorous, isn't it?

jiuker:
@klauspost This is true by context, but it's more rigorous, isn't it?

klauspost:
Changing something always has a risk of regressions, no matter how perfect we are. So unless it fixes a problem or a race, there isn't much gained by taking that risk. 

I'm not saying that we shouldn't add this change, but the gains should be clear. In this case it is removing a possible error on future changes.

harshavardhana:
I think we should avoid this change, it's not fixing anything.

PR #16906, opened 2023-03-28
Match existing format in minio-go when listing with metadata

klauspost:
## Description

Allow seamless parsing when sending data to minio-go.

Seems there is already unmarshal code in [minio-go objects](https://github.com/minio/minio-go/blob/b6c32fc05a1939419b983970afcea3bb8b62825d/api-datatypes.go#L127) - match the format of that instead.

Regression caused by #16883 - sorry.

## How to test this PR?

Must be part of `mc` listings.

## Types of changes


vadmeste:
> Regression caused by https://github.com/minio/minio/pull/16883 - sorry.

What regression ? https://github.com/minio/minio/pull/16883  seems to be fine

klauspost:
@vadmeste It doesn't match `minio-go` which has `UserTags map[string]string `json:"userTags"`.

It does however require some further tweaking: https://github.com/minio/minio-go/pull/1795

klauspost:
Dropping for simpler change.

PR #16910, opened 2023-03-28
Generalize the event store using go generics

Praveenrajmani:
## Description

This PR makes the event store generic so that it can be used by other targets like logger webhook targets, audit webhook targets and audit kafka targets etc.

## Motivation and Context

We will be implementing `queue_dir` support for audit and logger targets soon. For this, we need to generalize the current event store.

## How to test this PR?

Just some sanity tests to make sure the `queue_dir` functionality in event notifications work fine.

- Configure any notification target with `queue_dir` enabled
- Test event store functionality. ie, take down the target, send events and check if the events are properly queued in the provided `queue_dir` and then bring the target up and check if the events are getting replayed properly.

## Types of changes

## Checklist:


Praveenrajmani:
This PR requires https://github.com/minio/pkg/pull/62

go.mod
```diff
@@ -50,7 +50,7 @@ require (
 	github.com/minio/madmin-go/v2 v2.0.17
 	github.com/minio/minio-go/v7 v7.0.50
 	github.com/minio/mux v1.9.0
-	github.com/minio/pkg v1.6.5
+	github.com/minio/pkg v1.6.6-0.20230330040824-5db111e5f63c
```
Praveenrajmani, line 53: using `github.com/minio/pkg@main` currently, will change the tag once we make a release of `minio/pkg`

internal/store/queuestore.go
```diff
@@ -15,49 +15,59 @@
...
 	"time"
 
-	"github.com/minio/minio/internal/event"
+	"github.com/google/uuid"
 )
 
 const (
 	defaultLimit = 100000 // Default store limit.
-	eventExt     = ".event"
+	defaultExt   = ".unknown"
```
klauspost, line 35:
```suggestion
	defaultExt   = ".events"
```

klauspost (approved):
Very nice!

internal/store/queuestore.go
Praveenrajmani: This is an extension used for the items saved in the store. Each and every item that uses the store have its own file extension. Bucket notification events use `.event` already. So, i kept `.unknown` as the default one.

klauspost: It is .event*s*  - but it shouldn't really be used.

harshavardhana (approved):

PR #16919, opened 2023-03-29
fix: avoid logging when object/version not found in replication

allanrogerr:
## Description
Simply gracefully recover if unable to update replication metadata, without logging

## Motivation and Context
Logging is unnecessary since the system gracefully recovers in this situation

## Types of changes

## Checklist:

### DIFF
--- a/cmd/bucket-replication.go
+++ b/cmd/bucket-replication.go
@@ -508,7 +508,6 @@ func replicateDelete(ctx context.Context, dobj DeletedObjectReplicationInfo, obj
 		VersionSuspended: globalBucketVersioningSys.Suspended(bucket),
 	})
 	if err != nil && !isErrVersionNotFound(err) { // VersionNotFound would be reported by pool that object version is missing on.
-		logger.LogIf(ctx, fmt.Errorf("Unable to update replication metadata for %s/%s(%s): %s", bucket, dobj.ObjectName, versionID, err))
 		sendEvent(eventArgs{
 			BucketName: bucket,
 			Object: ObjectInfo{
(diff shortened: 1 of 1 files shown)


cmd/bucket-replication.go
```diff
@@ -508,7 +508,6 @@ func replicateDelete(ctx context.Context, dobj DeletedObjectReplicationInfo, obj
 		VersionSuspended: globalBucketVersioningSys.Suspended(bucket),
 	})
 	if err != nil && !isErrVersionNotFound(err) { // VersionNotFound would be reported by pool that object version is missing on.
```
harshavardhana, line 510:
```suggestion
	if err != nil && !isErrVersionNotFound(err) && !isErrObjectNotFound(err){ // VersionNotFound would be reported by pool that object version is missing on.
```

This is perhaps the correct fix here @allanrogerr we still log for other errors.

harshavardhana (changes requested):

cmd/bucket-replication.go
allanrogerr: Thanks - modifying

allanrogerr:
@harshavardhana Do we wish to also do this change here?
https://github.com/minio/minio/blob/master/cmd/bucket-replication.go#L1008-L1011

harshavardhana:
> @harshavardhana Do we wish to also do this change here? https://github.com/minio/minio/blob/master/cmd/bucket-replication.go#L1008-L1011

@allanrogerr yes.

poornas (approved):

harshavardhana (approved):

PR #16926, opened 2023-03-30
fix: potential data conflicts save site-resync metadata

jiuker:
## Description

Here I think we need to finish one batch of saves before we can move on to the next batch of saves.
Close this if on purpose.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/site-replication-utils.go
+++ b/cmd/site-replication-utils.go
@@ -170,6 +170,7 @@ func (sm *siteResyncMetrics) save(ctx context.Context) {
 		case <-sTimer.C:
 			if globalSiteReplicationSys.isEnabled() {
 				sm.Lock()
+				wg := sync.WaitGroup{}
 				for dID, rs := range sm.peerResyncMap {
 					st, ok := sm.resyncStatus[rs.resyncID]
 					if ok {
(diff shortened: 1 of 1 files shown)


poornas (approved):

harshavardhana:
If you have lets say 200 peers this will cause 200 parallel writes? @poornas

poornas:
> If you have lets say 200 peers this will cause 200 parallel writes? @poornas

resync runs on only one node in the cluster

harshavardhana:
> resync runs on only one node in the cluster

Isn't this invoking X callers without a top-level limit? - this PR doesn't look scalable to me. Calling wg.Add(1) for each peer with 300+ servers is going to be trouble for us in the long run.

poornas:
> > resync runs on only one node in the cluster
> 
> Isn't this invoking X callers without a top-level limit? - this PR doesn't look scalable to me. Calling wg.Add(1) for each peer with 300+ servers is going to be trouble for us in the long run. 

Not sure about the concern here - there is only one site resync going on any time per site replication peer, so at the max there are as n-1 resyncs going on where n is number of sites. A leader lock is taken so only one node in the cluster is running site resync

harshavardhana:
So the peersyncmap is actually just the replication peers not MinIO peers?

poornas:
> So the peersyncmap is actually just the replication peers not MinIO peers?

That's right

jiuker:
@harshavardhana What this pr is trying to convey is that if once the previous batch is not completed, because there is no limit, the next batch starts again. If the task of the latter batch writes data before the previous batch succeeds, which overwrites the data, the probability is small, but it has to be considered. If this probability is so small that you think it will not occur, turn off this pr.

harshavardhana:
> @harshavardhana What this pr is trying to convey is that if once the previous batch is not completed, because there is no limit, the next batch starts again. If the task of the latter batch writes data before the previous batch succeeds, which overwrites the data, the probability is small, but it has to be considered. If this probability is so small that you think it will not occur, turn off this pr.

I understand that part, the issue was I wanted to know how many peers will be present in syncMap() - we really cannot afford an additional 100's goroutines because having more goroutines that are actively doing disk I/O - causes some amount of CPU to build up.

But it looks like the peerSyncMap is always quite small since this is for replication.

PR #16937, opened 2023-03-31
read use Rwlock first.

jiuker:
## Description

It looks like we can use a read/write lock to get the data, and once the data doesn't exist, we can use the write lock to check once if it was written by another thread first, and generate a new one if it wasn't. If we use write lock at the beginning, it feels like it will affect the performance.
Close this pr if on purpose.
## Types of changes

## Checklist:

### DIFF
--- a/cmd/tier.go
+++ b/cmd/tier.go
@@ -260,16 +260,21 @@ func (config *TierConfigMgr) Bytes() ([]byte, error) {
 
 // getDriver returns a warmBackend interface object initialized with remote tier config matching tierName
 func (config *TierConfigMgr) getDriver(tierName string) (d WarmBackend, err error) {
+	config.RLock()
+	// Lookup in-memory drivercache
+	d, ok := config.drivercache[tierName]
+	if ok {
+		config.Unlock()
+		return d, nil
+	}
+	config.Unlock()
+
 	config.Lock()
 	defer config.Unlock()
-
-	var ok bool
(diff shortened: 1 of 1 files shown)


klauspost (commented):
Is there contention here? Otherwise there isn't really much gained.

jiuker:
> Is there contention here? Otherwise there isn't really much gained.

It enables getDriver to fetch data faster, and according to the contextual semantics, we fetch more from memory, but behaves as single threaded due to the write lock. After we do this, when there is data inside the memory, we can read the data inside the memory in a multi-threaded way. @klauspost

klauspost:
@jiuker If there is no or little contention on the lock this will be slower. That is why I am asking.

If there *is* contention it will be an improvement. So I am trying to figure out if you have observed contention for the lock.

jiuker:
> 

@klauspost No,But I found this function will be called by some different api.

klauspost:
Sure, but I am asking how often it is called and how much contention we can expect there is for the lock.

jiuker:
@harshavardhana It seems the reviewer didn't review this.

krisis:
@jiuker I agree with @klauspost that the gains by using RWMutex in this case would be very low, since we don't expect a lot of contention in this code path. Like any performance optimisation, it would be good to add this when we see high contention related issues first.

harshavardhana:
Closing this as not needed for now.

PR #16952, opened 2023-04-03
implement support for FTP/SFTP server

harshavardhana:
## Description
implement support for FTP/SFTP server

## Motivation and Context
(S)FTP server designed to use the handler
layer, the intention is to make sure
that (S)FTP server honors all the S3 
requirements including

- SSE (server-side encryption)
- Replication (bucket/site replication)

Passwords are to be validated again via 
signature v4. Credentials also honor
IAM policies - existing permissions 
automatically apply.

## How to test this PR?
Documentation is available at `docs/ftp/README.md` 

## Types of changes

## Checklist:

### DIFF
--- a/cmd/ftp-server-driver.go
+++ b/cmd/ftp-server-driver.go
@@ -1,0 +1,505 @@
+// Copyright (c) 2015-2023 MinIO, Inc.
+//
+// This file is part of MinIO Object Storage stack
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package cmd
+
+import (
+	"bytes"
+	"context"
+	"crypto/subtle"
+	"errors"
+	"fmt"
+	"io"
+	"os"
+	"strings"
+	"time"
+
+	"github.com/minio/madmin-go/v2"
+	"github.com/minio/minio-go/v7"
+	"github.com/minio/minio-go/v7/pkg/credentials"
+	"github.com/minio/minio/internal/auth"
+	"github.com/minio/minio/internal/logger"
+	ftp "goftp.io/server/v2"
+)
+
+var _ ftp.Driver = &ftpDriver{}
+
+// ftpDriver implements ftpDriver to store files in minio
+type ftpDriver struct {
+	endpoint string
+}
+
+// NewFTPDriver implements ftp.Driver interface
+func NewFTPDriver() ftp.Driver {
+	return &ftpDriver{endpoint: fmt.Sprintf("127.0.0.1:%s", globalMinioPort)}
+}
+
+func buildMinioPath(p string) string {
+	return strings.TrimPrefix(p, SlashSeparator)
+}
+
+func buildMinioDir(p string) string {
+	v := buildMinioPath(p)
+	if !strings.HasSuffix(v, SlashSeparator) {
+		return v + SlashSeparator
+	}
+	return v
+}
+
+type minioFileInfo struct {
+	p     string
+	info  minio.ObjectInfo
+	isDir bool
+}
+
+func (m *minioFileInfo) Name() string {
+	return m.p
+}
+
+func (m *minioFileInfo) Size() int64 {
+	return m.info.Size
+}
+
+func (m *minioFileInfo) Mode() os.FileMode {
+	if m.isDir {
+		return os.ModeDir
+	}
+	return os.ModePerm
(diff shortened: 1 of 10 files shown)


cmd/ftp-server-driver.go
```diff
@@ -0,0 +1,446 @@
...
+func (m *minioFileInfo) Name() string {
+	return m.p
+}
+
+func (m *minioFileInfo) Size() int64 {
+	return m.info.Size
+}
+
+func (m *minioFileInfo) Mode() os.FileMode {
+	return os.ModePerm
```
klauspost, line 81: I can't find the mc metadata key we had at one point. If we still have it, should we preserve this as a metadata field?

cmd/ftp-server-driver.go
```diff
@@ -0,0 +1,446 @@
...
+	if err != nil {
+		return err
+	}
+
+	cctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	bucket, prefix := path2BucketObject(path)
+	if bucket == "" {
+		// TODO implement bucket listing.
```
klauspost: Don't we pretty much need this?

cmd/ftp-server.go
```diff
@@ -0,0 +1,102 @@
...
+	if serverDebugLog {
+		logger.Info("%s < %d %s", sessionID, code, message)
+	}
+}
+
+func startFTPServer(c *cli.Context) {
+	tls := c.String("ftp-tls-private-key") != "" &&
+		c.String("ftp-tls-public-cert") != ""
+
+	ftpServer, err := server.NewServer(&server.Options{
```
klauspost: We need the passive port range to be specified, so it doesn't just pick random ones.

klauspost: I looked for a way to completely disabling active connections, but didn't find any.

klauspost (commented):
Nothing major with the implementation. Private overall issues.

cmd/ftp-server-driver.go
harshavardhana: Well, we do but they are not really meaningful here - on the server that's why these are just some dummy values that are okay.

The primary use case here is not full-blown FTP implementation just implementing it enough to satisfy, PUT/GET/LIST

cmd/ftp-server-driver.go
harshavardhana: yes forgot to push the change

cmd/ftp-server.go
harshavardhana: Will add

cmd/ftp-server-driver.go
```diff
@@ -0,0 +1,461 @@
...
+		for object := range clnt.ListObjects(cctx, bucket, opts) {
+			if object.Err != nil {
+				return
+			}
+			objectsCh <- object
+		}
+	}()
+
+	// Call RemoveObjects API
+	for err := range clnt.RemoveObjects(context.Background(), "my-bucketname", objectsCh, minio.RemoveObjectsOptions{}) {
```
poornas:
```suggestion
	for err := range clnt.RemoveObjects(context.Background(), bucket, objectsCh, minio.RemoveObjectsOptions{}) {
```

poornas (changes requested):

cmd/ftp-server-driver.go
```diff
@@ -0,0 +1,461 @@
...
+	}
+
+	dirPath := buildMinioDir(prefix)
+
+	_, err = clnt.PutObject(context.Background(), bucket, dirPath, bytes.NewReader([]byte("")), 0,
+		// Always send Content-MD5 to succeed with bucket with
+		// locking enabled. There is no performance hit since
+		// this is always an empty object
+		minio.PutObjectOptions{SendContentMd5: true},
+	)
```
poornas, line 437: if bucket name does not exist, should this create a bucket?

harshavardhana: Currently assuming buckets are created prior, but MakeBucket can be called in this situation.

cmd/ftp-server.go
harshavardhana: there is a way @klauspost we simply need to disable PORT cmd.

klauspost:
For the regular FTP. How do you control the passive port range?

harshavardhana:
> For the regular FTP. How do you control the passive port range?

This library simply uses Linux features where OS can give you a free-port whatever is available, if needed we can use a range that is just for us. I don't know if we need it, 
I might as well hide the flag and use it only in specific situations. @klauspost

cmd/sftp-server-driver.go
```diff
@@ -0,0 +1,343 @@
...
+			}}, nil
+		}
+
+		bucket, object := path2BucketObject(r.Filepath)
+		if bucket == "" {
+			return nil, errors.New("bucket name cannot be empty")
+		}
+
+		if object == "" {
+			_, err := clnt.BucketExists(context.Background(), bucket)
```
poornas:
```suggestion
			ok, err := clnt.BucketExists(context.Background(), bucket)
```

klauspost:
> This library simply uses Linux features where OS can give you a free-port whatever is available, if needed we can use a range that is just for us. 

For everything not running bare metal, you'd need a port range to expose the ports. Even bare metal users will probably appreciate that they can specify a port range in firewalls.

poornas (approved):

klauspost (approved):

PR #16955, opened 2023-04-03
batch: validate versioning on src/tgt buckets

poornas:
Also tweaking preconditions check to allow batch copy between unversioned buckets.

## Motivation and Context
correctness

## How to test this PR?
`mc batch..`

## Types of changes

## Checklist:

### DIFF
--- a/cmd/batch-handlers.go
+++ b/cmd/batch-handlers.go
@@ -1116,16 +1116,18 @@ func (r *BatchJobReplicateV1) Validate(ctx context.Context, job BatchJobRequest,
 	if r.Source.Bucket == "" {
 		return errInvalidArgument
 	}
+	var isRemoteToLocal bool
 	localBkt := r.Source.Bucket
 	if r.Source.Endpoint != "" {
 		localBkt = r.Target.Bucket
+		isRemoteToLocal = true
 	}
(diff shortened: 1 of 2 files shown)


harshavardhana (approved):

cmd/batch-handlers.go
```diff
@@ -1116,9 +1116,11 @@ func (r *BatchJobReplicateV1) Validate(ctx context.Context, job BatchJobRequest,
 	if r.Source.Bucket == "" {
 		return errInvalidArgument
 	}
+	var isRemoteToLocal bool
 	localBkt := r.Source.Bucket
 	if r.Source.Endpoint != "" {
 		localBkt = r.Target.Bucket
+		isRemoteToLocal = true
```
prakashsvmx, line 1123: could the below error  (line 1130) be updated to match source/target because the error message is not intuitive,  if the bucket is not present in target. ?
it  would also be helpful to include the bucket name as part of the error message.

prakashsvmx (approved):
Tested. LGTM :+1:

cmd/batch-handlers.go
harshavardhana: @poornas ^^

prakashsvmx (approved):

PR #16958, opened 2023-04-04
add isSysErrNotDir to openFileNoSync

jiuker:
## Description

add isSysErrNotDir to openFileNoSync
Here is openFileSync.
```go
func (s *xlStorage) openFileSync(filePath string, mode int) (f *os.File, err error) {
	// Create top level directories if they don't exist.
	// with mode 0777 mkdir honors system umask.
	if err = mkdirAll(pathutil.Dir(filePath), 0o777); err != nil {
		return nil, osErrToFileErr(err)
	}

	w, err := OpenFile(filePath, mode|writeMode, 0o666)
	if err != nil {
		// File path cannot be verified since one of the parents is a file.
		switch {
		case isSysErrIsDir(err):
			return nil, errIsNotRegular
		case osIsPermission(err):
			return nil, errFileAccessDenied
		case isSysErrNotDir(err):
			return nil, errFileAccessDenied
		case isSysErrIO(err):
			return nil, errFaultyDisk
		case isSysErrTooManyFiles(err):
			return nil, errTooManyOpenFiles
		default:
			return nil, err
		}
	}

	return w, nil
}
```
Here is openFileNoSync
```go
func (s *xlStorage) openFileNoSync(filePath string, mode int) (f *os.File, err error) {
	// Create top level directories if they don't exist.
	// with mode 0777 mkdir honors system umask.
	if err = mkdirAll(pathutil.Dir(filePath), 0o777); err != nil {
		return nil, osErrToFileErr(err)
	}

	w, err := OpenFile(filePath, mode, 0o666)
	if err != nil {
		// File path cannot be verified since one of the parents is a file.
		switch {
		case isSysErrIsDir(err):
			return nil, errIsNotRegular
		case osIsPermission(err):
			return nil, errFileAccessDenied
		case isSysErrIO(err):
			return nil, errFaultyDisk
		case isSysErrTooManyFiles(err):
			return nil, errTooManyOpenFiles
		default:
			return nil, err
		}
	}

	return w, nil
}
```
Should we add isSysErrNotDir For `openFileNoSync` ?
## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Optimization (provides speedup with no functional changes)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [ ] Fixes a regression (If yes, please add `commit-id` or `PR #` here)
- [ ] Unit tests added/updated
[...]

PR #16970, opened 2023-04-05
refactor: reuse open file in storage interface

ferhatelmas:
## Description
Implement sync open file via reusing no sync open file.

## Motivation and Context
To prevent divergence in the same implementation and code reuse/dry, related to #16958.

## Types of changes

## Checklist:


cmd/xl-storage.go
```diff
@@ -1672,32 +1672,7 @@ func (s *xlStorage) openFileDirect(path string, mode int) (f *os.File, err error
...
-			return nil, errFaultyDisk
-		case isSysErrTooManyFiles(err):
-			return nil, errTooManyOpenFiles
-		default:
-			return nil, err
-		}
-	}
-
-	return w, nil
+	return s.openFileNoSync(filePath, mode|writeMode)
```
harshavardhana: There is a reason why we want sync to be sync and not `nosync` - This code is not correct @ferhatelmas

harshavardhana (changes requested):

cmd/xl-storage.go
harshavardhana: they have a different function signature, we should change `s.openFileNoSync` to be `s.openFile`

ferhatelmas: so next step is to rename `openFileNoSync` to `openFile` and keep this change (call within sync with right mode) right?

harshavardhana: yes @ferhatelmas

ferhatelmas: thanks, updated

harshavardhana (approved):

klauspost (approved):

PR #16977, opened 2023-04-05
trace: Fix <unknown> func name of requests rejected by max clients

vadmeste:
## Description
Currently, when a request is rejected because it was queued for a long
 time (requests_deadline), the trace event does not appear when filtered
against S3 calls. However we can see the request when running
 'mc admin trace -a', the function name will be `'<unknown>'`

The function name is set as 'unknown' because it did not reach the S3 handler.

Fix this by setting the trace func name in maxClients(), since 
maxClients() is only called for S3 requests.

Also, when the request is canceled, return 499 status code instead 
of 200 (the default one)

## Motivation and Context
Fix weird trace events when requests time out with max_requests feature

## How to test this PR?
Hard to test without modifying the code

## Types of changes

## Checklist:


klauspost (approved):
LGTM

cmd/handler-api.go
```diff
@@ -288,6 +293,10 @@ func maxClients(f http.HandlerFunc) http.HandlerFunc {
 			globalHTTPStats.addRequestsInQueue(-1)
 			return
 		case <-r.Context().Done():
+			// When the client disconnects before getting the S3 handler
+			// status code response, set the status code to 499 so this request
+			// will be properly audited and traced.
+			w.WriteHeader(499)
```
harshavardhana, line 299: I suspect this can lead to double Writes on the responseWriter @vadmeste if ServeHTTP returned then responseWriter must have been written to already.

harshavardhana (changes requested):

cmd/handler-api.go
```diff
@@ -272,6 +273,10 @@ func maxClients(f http.HandlerFunc) http.HandlerFunc {
 
 		globalHTTPStats.addRequestsInQueue(1)
 
+		if tc, ok := r.Context().Value(mcontext.ContextTraceKey).(*mcontext.TraceCtxt); ok {
+			tc.FuncName = "s3.MaxClients"
+		}
```
harshavardhana, line 278: Isn't this fixed via #16964

harshavardhana (changes requested):

cmd/handler-api.go
vadmeste: > Isn't this fixed via https://github.com/minio/minio/pull/16964

This is for mc admin trace, not audit

cmd/handler-api.go
vadmeste: > I suspect this can lead to double Writes on the responseWriter @vadmeste if ServeHTTP returned then responseWriter must have been written to already.

I thought about it, it is not possible to have a double write header. This is a handler that decided to return and not calling handlers in the lower layers, so it is allowed to send a status code. No handler in the upper layer is allowed to send a status code as well.

klauspost: > I suspect this can lead to double Writes

Not here, since we haven't started processing the request yet. it is canceled while in queue.

cmd/handler-api.go
harshavardhana: so this should be generalized? because it looks like other handlers also might not be visible.

vadmeste: > so this should be generalized? because it looks like other handlers also might not be visible.

I don't think there is other handlers without a func name. We did not see this before because it is rare that a request is rejected by the server using the requests_max/requests_deadline feature.

harshavardhana (approved):

PR #16983, opened 2023-04-06
simplify bucketInfo return in GetBucketInfo peer call

jiuker:
## Description

		// bucketInfos:[local,remote...]
		// errs:[nil,remote...]
It seems will return local info always.
```go
	errs := []error{nil}
	bucketInfos[0] = bucketInfo
```

## Types of changes

## Checklist:

### DIFF
--- a/cmd/peer-s3-client.go
+++ b/cmd/peer-s3-client.go
@@ -171,7 +171,9 @@ func (sys *S3PeerSys) GetBucketInfo(ctx context.Context, bucket string, opts Buc
 	}
 
 	for i, err := range errs {
-		if err == nil {
+		// bucketInfos:[local,remote...]
+		// errs:[nil,remote...]
+		if err == nil && i != 0 {
 			bucketInfo = bucketInfos[i]
 			break
 		}
--- a/cmd/peer-s3-client.go
+++ b/cmd/peer-s3-client.go
@@ -169,16 +169,7 @@ func (sys *S3PeerSys) GetBucketInfo(ctx context.Context, bucket string, opts Buc
 	if err = reduceReadQuorumErrs(ctx, errs, bucketOpIgnoredErrs, quorum); err != nil {
 		return BucketInfo{}, toObjectErr(err, bucket)
 	}
-
-	for i, err := range errs {
-		// bucketInfos:[local,remote...]
-		// errs:[nil,remote...]
-		if err == nil && i != 0 {
-			bucketInfo = bucketInfos[i]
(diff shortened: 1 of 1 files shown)


cmd/peer-s3-client.go
```diff
@@ -171,7 +171,9 @@ func (sys *S3PeerSys) GetBucketInfo(ctx context.Context, bucket string, opts Buc
 	}
 
 	for i, err := range errs {
-		if err == nil {
+		// bucketInfos:[local,remote...]
+		// errs:[nil,remote...]
+		if err == nil && i != 0 {
```
harshavardhana: This is not needed @jiuker if local succeeded we always return local on purpose but make sure we verify quorum.

We do not need it to be a different remote endpoints output.

harshavardhana (changes requested):

cmd/peer-s3-client.go
jiuker: Then the loop is meaningless, because surely the first one must be nil @harshavardhana

jiuker:
@harshavardhana Based on the context, it looks like the loop is not needed here.
```go
	errs := []error{nil}
	bucketInfos[0] = bucketInfo
	for idx, client := range sys.peerClients {
		idx := idx
		client := client
		g.Go(func() error {
			if client == nil {
				return errPeerOffline
			}
			bucketInfo, err := client.GetBucketInfo(ctx, bucket, opts)
			if err != nil {
				return err
			}
			bucketInfos[idx+1] = bucketInfo
			return nil
		}, idx)
	}
	errs = append(errs, g.Wait()...)
```

harshavardhana:
No it is needed @jiuker to calculate quorum

cmd/peer-s3-client.go
harshavardhana: No we need to deduce quorum, if quorum fails we do not return the local disk.

jiuker:
@harshavardhana I mean this diff
https://github.com/minio/minio/pull/16983/files#diff-7c81f2846ef59d2ab49f495171f0c4cd7c8703e2b923f6a30e7e568a80d5472a

```diff
-	for i, err := range errs {
-		if err == nil {
-			bucketInfo = bucketInfos[i]
-			break
-		}
-	}
```

harshavardhana (approved):

harshavardhana:
yeah that change is perfectly fine.

harshavardhana (approved):

PR #16987, opened 2023-04-06
snowball: Restrict zstd window size

klauspost:
## Description

Restrict Zstandard window size to 16MB to avoid excessive allocations.

Should not be hit with normal operation.

## Types of changes
- [x] Safety - reduce OOM vectors.


vadmeste (approved):
LGTM

harshavardhana (approved):

PR #16993, opened 2023-04-06
Add trace support for ilm activity

poornas:
## Motivation and Context
Visibility into ilm activity

## How to test this PR?
with companion mc PR https://github.com/minio/mc/pull/4538

## Types of changes

## Checklist:

### DIFF
--- a/cmd/bucket-lifecycle.go
+++ b/cmd/bucket-lifecycle.go
@@ -32,6 +32,7 @@ import (
 	"time"
 
 	"github.com/google/uuid"
+	"github.com/minio/madmin-go/v2"
 	"github.com/minio/minio-go/v7/pkg/tags"
 	"github.com/minio/minio/internal/amztime"
 	sse "github.com/minio/minio/internal/bucket/encryption"
@@ -68,6 +69,30 @@ func (sys *LifecycleSys) Get(bucketName string) (lc *lifecycle.Lifecycle, err er
 // NewLifecycleSys - creates new lifecycle system.
 func NewLifecycleSys() *LifecycleSys {
 	return &LifecycleSys{}
+}
+
+func ilmTrace(startTime time.Time, duration time.Duration, oi ObjectInfo, event string) madmin.TraceInfo {
+	return madmin.TraceInfo{
+		TraceType: madmin.TraceILM,
+		Time:      startTime,
+		NodeName:  globalLocalNodeName,
+		FuncName:  event,
+		Duration:  duration,
+		Path:      pathJoin(oi.Bucket, oi.Name),
+		Error:     "",
+		Message:   getSource(4),
+		Custom:    map[string]string{"version-id": oi.VersionID},
+	}
+}
+
+func (sys *LifecycleSys) trace(oi ObjectInfo) func(event string) {
+	startTime := time.Now()
+	return func(event string) {
+		duration := time.Since(startTime)
+		if globalTrace.NumSubscribers(madmin.TraceILM) > 0 {
+			globalTrace.Publish(ilmTrace(startTime, duration, oi, event))
+		}
+	}
 }
 
 type expiryTask struct {
@@ -360,6 +385,7 @@ const (
 // 1. when a restored (via PostRestoreObject API) object expires.
 // 2. when a transitioned object expires (based on an ILM rule).
 func expireTransitionedObject(ctx context.Context, objectAPI ObjectLayer, oi *ObjectInfo, lcOpts lifecycle.ObjectOpts, action expireAction) error {
+	traceFn := globalLifecycleSys.trace(*oi)
 	var opts ObjectOptions
 	opts.Versioned = globalBucketVersioningSys.PrefixEnabled(oi.Bucket, oi.Name)
 	opts.VersionID = lcOpts.VersionID
@@ -385,7 +411,7 @@ func expireTransitionedObject(ctx context.Context, objectAPI ObjectLayer, oi *Ob
 		}
 
 		// Send audit for the lifecycle delete operation
-		auditLogLifecycle(ctx, *oi, ILMExpiry)
+		defer auditLogLifecycle(ctx, *oi, ILMExpiry, traceFn)
 
 		eventName := event.ObjectRemovedDelete
 		if lcOpts.DeleteMarker {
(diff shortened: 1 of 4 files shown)


cmd/bucket-lifecycle.go
```diff
@@ -360,6 +383,7 @@ const (
 // 1. when a restored (via PostRestoreObject API) object expires.
 // 2. when a transitioned object expires (based on an ILM rule).
 func expireTransitionedObject(ctx context.Context, objectAPI ObjectLayer, oi *ObjectInfo, lcOpts lifecycle.ObjectOpts, action expireAction) error {
+	traceFn := globalLifecycleSys.trace(*oi)
```
harshavardhana, line 388: we should be doing `defer traceFn` such that we can capture the other returns as well.

harshavardhana (changes requested):

cmd/bucket-lifecycle.go
poornas: deferring auditLogLifecycle call since traceFn is passed into this.

harshavardhana:
PTAL at the conflicts @poornas

go.sum
```diff
@@ -774,8 +774,8 @@ github.com/minio/highwayhash v1.0.2/go.mod h1:BQskDq+xkJ12lmlUUi7U0M5Swg3EWR+dLT
 github.com/minio/kes-go v0.1.0 h1:h201DyOYP5sTqajkxFGxmXz/kPbT8HQNX1uh3Yx2PFc=
 github.com/minio/kes-go v0.1.0/go.mod h1:VorHLaIYis9/MxAHAtXN4d8PUMNKhIxTIlvFt0hBOEo=
 github.com/minio/madmin-go v1.6.6/go.mod h1:ATvkBOLiP3av4D++2v1UEHC/QzsGtgXD5kYvvRYzdKs=
-github.com/minio/madmin-go/v2 v2.0.18 h1:5kME44CIL7RlNm99hg6EU0K5AiIX0RtVHFu0nMEWmNE=
-github.com/minio/madmin-go/v2 v2.0.18/go.mod h1:8bL1RMNkblIENFSgGYjeHrzUx9PxROb7OqfNuMU9ivE=
+github.com/minio/madmin-go/v2 v2.0.19-0.20230406175749-1404a84d1612 h1:dsvGtTzHr5K4VgpwdJ75umzq9MimLIPqIZQI5FH/n8I=
+github.com/minio/madmin-go/v2 v2.0.19-0.20230406175749-1404a84d1612/go.mod h1:8bL1RMNkblIENFSgGYjeHrzUx9PxROb7OqfNuMU9ivE=
```
vadmeste: You can use v2.0.19 now

vadmeste (commented):
one comment, LGTM otherwise

cmd/bucket-lifecycle.go
```diff
@@ -70,6 +71,28 @@ func NewLifecycleSys() *LifecycleSys {
...
 }
 
+func ilmTrace(startTime time.Time, duration time.Duration, oi ObjectInfo, event string) madmin.TraceInfo {
+	return madmin.TraceInfo{
+		TraceType: madmin.TraceILM,
+		Time:      startTime,
+		NodeName:  globalLocalNodeName,
+		FuncName:  event,
+		Duration:  duration,
+		Path:      pathJoin(oi.Bucket, oi.Name),
```
krisis, line 81: Can we include `oi.VersionID` into `madmin.TraceInfo.Path` like so?
```suggestion
		Path:      fmt.Sprintf("%s?versionId=%s", pathJoin(oi.Bucket, oi.Name), oi.VersionID)
```

cmd/bucket-lifecycle.go
```diff
@@ -70,6 +71,28 @@ func NewLifecycleSys() *LifecycleSys {
 	return &LifecycleSys{}
 }
 
+func ilmTrace(startTime time.Time, duration time.Duration, oi ObjectInfo, event string) madmin.TraceInfo {
+	return madmin.TraceInfo{
+		TraceType: madmin.TraceILM,
+		Time:      startTime,
+		NodeName:  globalLocalNodeName,
+		FuncName:  event,
```
krisis, line 79: An ILM action may be called from multiple code paths. For example, ILMExpiry may be called during scanning or when serving a HEAD/GET object request on an object version past its expiry. We can use `madmin.TraceInfo.Message` to pass the actual function name that invoked this ILM action.

```suggestion
		FuncName:  event,
		Message, getSource(3),	
```
Note: 3 in `getSource(3)` above depends on how many functions deep `ilmTrace` is from the function calling `globalLifecycleSys.Trace`. Please correct this based on your local testing.

krisis (commented):
Overall PR looks good to me. I have a couple of comments to add more information to the trace log.

cmd/bucket-lifecycle.go
klauspost: I added a `map[string]string` that allows adding custom fields. That would make it easier to parse.

cmd/bucket-lifecycle.go
```diff
@@ -70,6 +71,28 @@ func NewLifecycleSys() *LifecycleSys {
...
+		Path:      pathJoin(oi.Bucket, oi.Name),
+		Error:     "",
+	}
+}
+
+func (sys *LifecycleSys) trace(oi ObjectInfo) func(event string) {
+	startTime := time.Now()
+	return func(event string) {
+		duration := time.Since(startTime)
+		if globalTrace.NumSubscribers(madmin.TraceILM) > 0 {
```
klauspost, line 92: Check this outside and return a sub function if 0.

harshavardhana (approved):

cmd/bucket-lifecycle.go
```diff
@@ -70,6 +71,29 @@ func NewLifecycleSys() *LifecycleSys {
...
 }
 
+func ilmTrace(startTime time.Time, duration time.Duration, oi ObjectInfo, event string) madmin.TraceInfo {
+	return madmin.TraceInfo{
+		TraceType: madmin.TraceILM,
+		Time:      startTime,
+		NodeName:  globalLocalNodeName,
+		FuncName:  event,
+		Duration:  duration,
+		Path:      fmt.Sprintf("%s?versionId=%s", pathJoin(oi.Bucket, oi.Name), oi.VersionID),
```
krisis: Per @klauspost recommendation, we can use the `Custom` map to include details like version-id instead of overloading the path, like I had suggested earlier. 
```suggestion
		Path:      pathJoin(oi.Bucket, oi.Name),
		Custom: map[string]string{"version-id": oi.VersionID},
```

krisis (approved):

PR #17004, opened 2023-04-10
avoid attempting to migrate old configs

harshavardhana:
## Description
avoid attempting to migrate old configs

## Motivation and Context
instead of migration, move them in memory to a
local representation of new config.

In all other cases re-initialize if config's
missing.

## Types of changes

## Checklist:


klauspost (approved):
LGTM. I presume you tested as much as possible.

vadmeste:
I think maybe we should start with adding MinIO version in format.json. It may be useful to prevent direct upgrade from very old versions when we do very old code clean up.

vadmeste (commented):
@harshavardhana LGTM, but maybe if we want to just drop support of older config versions to clean up existing code, we can error out after testing on config version (<=33) and tell that the user needs to upgrade to 'RELEASE.2023-04-07T05-28-58Z' first.

harshavardhana:
> @harshavardhana LGTM, but maybe if we want to just drop support of older config versions to clean up existing code, we can error out after testing on config version (<=33) and tell that the user needs to upgrade to 'RELEASE.2023-04-07T05-28-58Z' first.

We can but it's a bunch of stuff to document @vadmeste and confusing. I would like to avoid that completely.

@klauspost I did I will test more, there is one more change missing here. I am removing migration code for credentials based encryption to KMS based.

harshavardhana:
re-did the decryption migration as well, we will never need to migrate any old content, we simply read them always as their original entity.

But let the WRITEs make sure that we write back in the new format, so we do not need to hold
MinIO startup migrating content.

krisis (approved):

PR #17014, opened 2023-04-11
config: fix duplication of replication priority key

poornas:
## How to test this PR?
upgrade from RELEASE.2022-09-22T18-57-27Z to latest - 
`mc admin config get alias api` should show only one entry for replication_priority

## Types of changes

## Checklist:

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -602,6 +602,17 @@ var renamedSubsys = map[string]string{
 	// Add future sub-system renames
 }
 
+const ( // deprecated keys
+	apiReplicationWorkers       = "replication_workers"
+	apiReplicationFailedWorkers = "replication_failed_workers"
+)
+
+// map of subsystem to deleted keys
+var deletedSubSysKeys = map[string][]string{
+	APISubSys: {apiReplicationWorkers, apiReplicationFailedWorkers},
+	// Add future sub-system deleted keys
+}
+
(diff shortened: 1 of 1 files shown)


harshavardhana (approved):

internal/config/config.go
```diff
@@ -602,6 +602,17 @@ var renamedSubsys = map[string]string{
...
 }
 
+const ( // deprecated keys
+	apiReplicationWorkers       = "replication_workers"
+	apiReplicationFailedWorkers = "replication_failed_workers"
+)
+
+// map of subsystem to deleted keys
+var deletedSubSysKeys = map[string][]string{
+	APISubSys: {apiReplicationWorkers, apiReplicationFailedWorkers},
```
harshavardhana, line 612: the list of keys I think should be provided by the `internal/config/api/config.go` so that - each sub-system can have its own list of keys to deprecate.

harshavardhana (approved):
Minor we can do it later as well.

internal/config/config.go
poornas: internal/config/api/ already imports internal/config leading to circular dependencies

harshavardhana: hmm I think we should do something about that, perhaps need to split deps.

krisis (approved):

PR #17026, opened 2023-04-12
simplify MRF, converge it to regular healing

harshavardhana:
## Description
simplify MRF, converge it to regular healing

## Motivation and Context
MRF doesn't need to exist as an independent 
healing subsystem, instead it simply should 
be piggybacking existing healing workers 
as well as healing functionality.

## How to test this PR?
MRF functionality is tied to taking nodes down, you 
would have to do that locally by setting up a local cluster
something like this

```
#!/bin/bash

set -x

export MINIO_PROMETHEUS_AUTH_TYPE="public"
export MINIO_API_DELETE_CLEANUP_INTERVAL="5s"
export MINIO_CI_CD=1
export MINIO_SCANNER_CYCLE=10s
killall -9 minio
sudo rm -rf ${HOME}/tmp/dist

scheme="http"
nr_servers=4

addr="localhost"
args=""
for ((i=0;i<$[${nr_servers}];i++)); do
    args="$args $scheme://$addr:$[9100+$i]/${HOME}/tmp/dist/path1/$i"
done

echo $args


for ((i=0;i<$[${nr_servers}];i++)); do
    (minio server --address ":$[9100+$i]" $args 2>&1 > /tmp/log$i.txt) &
done
```

Periodically take down minio pids, and restarting them to see 
object healing catch up on the missing entries.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/mrf.go
+++ b/cmd/mrf.go
@@ -20,16 +20,13 @@ package cmd
 import (
 	"context"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	"github.com/minio/madmin-go/v2"
-	"github.com/minio/minio/internal/logger"
 )
 
 const (
-	mrfInfoResetInterval = 10 * time.Second
-	mrfOpsQueueSize      = 10000
+	mrfOpsQueueSize = 100000
 )
 
 // partialOperation is a successful upload/delete of an object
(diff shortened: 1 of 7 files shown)


harshavardhana:
PTAL @vadmeste primarily need your inputs here.

cmd/mrf.go
```diff
@@ -102,140 +71,43 @@ func (m *mrfState) addPartialOp(op partialOperation) {
...
+			// 4 retries in the span 250ms each.
+
+			// Additionally retry healing 3 times, upon success
+			// move to the next.
+			for retry := 0; retry < 3; retry++ {
+				var err error
+				if u.object == "" {
+					err = healBucket(u.bucket, madmin.HealNormalScan)
+				} else {
+					err = healObject(u.bucket, u.object, u.versionID, madmin.HealNormalScan)
```
vadmeste: The retry is not that useful for MRF because healObject() wil always return nil. An error means either nodes are not available or the object is corrupted

Otherwise, instead of the retry logic, you can create a timestamp of the MRF event in 'addPartialOp()' function, that way we can avoid sleeping for one second for each object if the timestamp is older than one second. We can even confomtably increase one second to 5 for example.

vadmeste (commented):
One comment

cmd/mrf.go
harshavardhana: okay.

vadmeste (approved):
LGTM

PR #17035, opened 2023-04-14
Add part count to checksum

klauspost:
## Description

It seems like AWS is now returning the part count on multipart checksums: https://stackoverflow.com/questions/75630708/verifying-aws-s3-vs-minio-checksum

I did not find any documentation on this. It appears to be serverside only, so you are still supposed to send checksums *without* the part count for complete-multipart.

We do not change already uploaded content. This *should* not be breaking, but tests will have to confirm that.

## How to test this PR?

Upload multipart objects. Observe checksums.

Verified by mint tests as well.  `ChecksumCRC32C:"1yLc4A==-9"`.

## Types of changes


vadmeste (approved):
LGTM

harshavardhana (approved):

PR #17036, opened 2023-04-14
Reduce allocations in Walkdir

klauspost:
## Description

Joining paths is one of the most allocation heavy when listing. Reuse a stringbuffer for listing operations.

Furthermore path cleaning as avoided when not needed, which also reduces allocations.

Use a stringbuilder that can be reused to reduce allocations when listing directories.

## How to test this PR?

Regular tests.

## Types of changes

### DIFF
--- a/cmd/metacache-walk.go
+++ b/cmd/metacache-walk.go
@@ -126,6 +126,7 @@ func (s *xlStorage) WalkDir(ctx context.Context, opts WalkDirOptions, wr io.Writ
 
 	scanDir = func(current string) error {
 		// Skip forward, if requested...
+		var sb strings.Builder
 		forward := ""
 		if len(opts.ForwardTo) > 0 && strings.HasPrefix(opts.ForwardTo, current) {
 			forward = strings.TrimPrefix(opts.ForwardTo, current)
@@ -198,7 +199,7 @@ func (s *xlStorage) WalkDir(ctx context.Context, opts WalkDirOptions, wr io.Writ
 			if HasSuffix(entry, xlStorageFormatFile) {
 				var meta metaCacheEntry
 				s.walkReadMu.Lock()
-				meta.metadata, err = s.readMetadata(ctx, pathJoin(volumeDir, current, entry))
+				meta.metadata, err = s.readMetadata(ctx, pathJoinBuf(&sb, volumeDir, current, entry))
 				s.walkReadMu.Unlock()
 				diskHealthCheckOK(ctx, err)
 				if err != nil {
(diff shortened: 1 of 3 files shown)


harshavardhana:
What is the overall alloc reduction @klauspost ?

klauspost:
Didn't check. But I can run warp and get a baseline.

klauspost:
Approximate 50-60% reduction in objects/space allocated for that part:

And WalkDir goes from 47 -> 40% allocs overall.

Before Objects:
image


After Objects:
image


Before SPACE:
image


After SPACE:

image

cmd/metacache-walk.go
```diff
@@ -126,6 +127,7 @@ func (s *xlStorage) WalkDir(ctx context.Context, opts WalkDirOptions, wr io.Writ
 
 	scanDir = func(current string) error {
 		// Skip forward, if requested...
+		var sb bytes.Buffer
```
harshavardhana, line 130: can we not use `strings.Builder` here? or better use github.com/valyala/bytebufferpool ? we already use this - looks like no harm in using it?

harshavardhana (changes requested):

cmd/metacache-walk.go
klauspost: `strings.Builder` allocates on each reset. This will keep the buffer and only allocate on on first use and final output (conversion to string). You are welcome to feed it a byte buffer as seed, but this already eliminate most allocs. This is simple and get us most of the way.

The initial alloc in bytes.Buffer doesn't even show up on the profiles, so the difference would be near zero.

cmd/object-api-utils.go
```diff
@@ -231,13 +232,102 @@ func pathsJoinPrefix(prefix string, elem ...string) (paths []string) {
...
 			trailingSlash = SlashSeparator
 		}
 	}
 	return path.Join(elem...) + trailingSlash
 }
 
+// pathJoinBuf - like path.Join() but retains trailing SlashSeparator of the last element.
+// Provide a string builder to reduce allocation.
+func pathJoinBuf(dst *bytes.Buffer, elem ...string) string {
+	trailingSlash := len(elem) > 0 && hasSuffixByte(elem[len(elem)-1], '/')
```
```suggestion
	trailingSlash := len(elem) > 0 && hasSuffixByte(elem[len(elem)-1], SlashSeparatorChar)
```

cmd/object-api-utils.go
```diff
@@ -231,13 +232,102 @@ func pathsJoinPrefix(prefix string, elem ...string) (paths []string) {
...
+		if added > 0 || e != "" {
+			if added > 0 {
+				dst.WriteRune(SlashSeparatorChar)
+			}
+			dst.WriteString(e)
+			added += len(e)
+		}
+	}
+
+	if pathNeedsClean(dst.Bytes()) {
```
harshavardhana, line 260:
```suggestion
	if s := pathNeedsClean(dst.Bytes()); s ! "" {
	     if trailingSlash {
	         return s + SlashSeparator
	     }
	     return s 
	}
```

is this doable? since pathNeedsClean() already peeks into the bytes.

harshavardhana (changes requested):

cmd/object-api-utils.go
klauspost: It is quite rare it needs cleaning after the slashes were cleaned from the input.

It was too complicated for me to figure out without too much fiddling. So I didn't want to risk breaking it.

klauspost:
New change should remove another 4% of total allocs (HasSuffix->ToLower)

harshavardhana (approved):

PR #17037, opened 2023-04-15
Listen bucket events to send empty events with new line

vadmeste:
## Description
minio-go currently parses listen bucket events by lines but it can fail
when the bucket is idle and the server sends too many whitespaces that
can break minio-go buffer scanner.

Add a query parameter to ask the server to send empty event each
interval instead of whitespaces.

## Motivation and Context
minio-go Listen Bucket Notification complains when the server does not send
a notification for a long time (28 days)

## How to test this PR?
Hard without modification of the code

## Types of changes

## Checklist:

### DIFF
--- a/cmd/listen-notification-handlers.go
+++ b/cmd/listen-notification-handlers.go
@@ -20,6 +20,7 @@ package cmd
 import (
 	"encoding/json"
 	"net/http"
+	"strconv"
 	"time"
 
 	"github.com/minio/minio/internal/event"
@@ -141,8 +142,30 @@ func (api objectAPIHandlers) ListenNotificationHandler(w http.ResponseWriter, r 
 		peer.Listen(listenCh, ctx.Done(), values)
 	}
 
-	keepAliveTicker := time.NewTicker(500 * time.Millisecond)
-	defer keepAliveTicker.Stop()
+	var (
+		emptyEventTicker <-chan time.Time
+		keepAliveTicker  <-chan time.Time
+	)
+
+	if p := values.Get("ping"); p != "" {
+		pingInterval, err := strconv.Atoi(p)
+		if err != nil {
+			writeErrorResponse(ctx, w, errorCodes.ToAPIErr(ErrInvalidQueryParams), r.URL)
+			return
+		}
+		if pingInterval < 1 {
+			writeErrorResponse(ctx, w, errorCodes.ToAPIErr(ErrInvalidQueryParams), r.URL)
+			return
+		}
+		t := time.NewTicker(time.Duration(pingInterval) * time.Second)
+		defer t.Stop()
+		emptyEventTicker = t.C
+	} else {
+		// Deprecated Apr 2023
+		t := time.NewTicker(500 * time.Millisecond)
+		defer t.Stop()
+		keepAliveTicker = t.C
+	}
 
 	enc := json.NewEncoder(w)
 	for {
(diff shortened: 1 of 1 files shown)


cmd/listen-notification-handlers.go
```diff
@@ -141,6 +141,13 @@ func (api objectAPIHandlers) ListenNotificationHandler(w http.ResponseWriter, r
 		peer.Listen(listenCh, ctx.Done(), values)
 	}
 
+	var emptyEventTicker <-chan time.Time
+	if values.Get("ping") == "true" {
+		ticker := time.NewTicker(time.Minute)
+		defer ticker.Stop()
+		emptyEventTicker = ticker.C
+	}
+
```
harshavardhana: You can change keepAlive to send empty events{} entirely @vadmeste

harshavardhana (changes requested):
We don't need an additional flag here, are trying to not break old clients?

cmd/listen-notification-handlers.go
vadmeste: > You can change keepAlive to send empty events{} entirely @vadmeste

This is only possible if we require users to upgrade their minio-go version when they upgrade MinIO version, otherwise it will break.

harshavardhana: No what I meant is never send empty instead send an event called s3:Ping event - clients know exactly what to look for.

No new changes are needed on client side.

harshavardhana: A new flag just for this is an overkill IMO.

vadmeste: @harshavardhana s3:ping may break some deployments, like if an application is listening only for object upload events it will receive other events and will break.

> A new flag just for this is an overkill IMO.

I don't think we have a choice unless to fix minio-go as well. For example, avoid using bufio.Scanner or implement a reader which discards whitespaces between JSON documents (not very obvious since it needs to ignore escaped '{' and '}' as well)

At the other hand, we can take this opportunity to deprecate the whitespace code and send s3:Ping instead. The client can send a request with '?ping=60' to ask the server to send s3:Ping each 60 seconds instead of  the whitespaces.

klauspost:
Isn't this more reasonable to fix on the client side?

Gateways could also be timing out and other random stuff. Having a regular stream of data seems like a good idea.

vadmeste:
> Isn't this more reasonable to fix on the client side?

It is, although it involves some work in minio-go.

Currently the code sends a whitespace and flushes ResponseWriter every 500 ms, I am not sure why it is that aggressive in the first place, also it doesn't scale well if there are many listeners connecting to the server. The idea is to let the client configures its own ping interval > 1 second. That may be better for a better control.

harshavardhana (approved):

Praveenrajmani (approved):

klauspost:
A backtrace would help a lot.

However, it could seem like you are closing upstream (buf) while a read occurs. 

The decompressor decompresses ahead to try to always have a block available, so it will often do a Read call upstream even when no Read is called.

Using the `zstd.WithDecoderConcurrency(1)` option will switch to sync decompression, OR calling `zstdReader.Close()` before you close upstream will make sure that it is safe to close.

klauspost (approved):
Setting the `keepAliveTicker` to 10 seconds should be fine as a bonus.

vadmeste:
> Setting the keepAliveTicker to 10 seconds should be fine as a bonus.

Yes, I kept the old behavior but added a query parameter that the client can specify to ask the server to send a ping regularly

PR #17039, opened 2023-04-17
fix: inlined objects don't need to honor long locks

### LINKED ISSUE #16314: Support concurrent upload and downloads to same object but different versions
## Expected Behavior
When uploading a new version of a versioned file, I would expect it to be able to be uploaded even if users are downloading the previous version.

## Current Behavior
When uploading a new version of a file that is heavily downloaded, I get the following error:
```mc: <ERROR> Failed to copy `MyFile.zip`. A timeout occurred while trying to lock a resource, please reduce your request rate```

## Possible Solution
Place a CDN in front of our MinIO server. Probably a preferable solution, but I don't understand why MinIO doesn't handle this situation as well.

1. Upload a large file to the server. (Ours is 500mb).
2. Make the bucket public.
3. Have many users attempt to download the file at once.
4. Try and `minio cp` a new version on top of the only version.
5. Observer the timeout and error.

## Context
We make games with releases stored on MinIO, and occasionally release large updates. These updates lead to a spike in traffic to our MinIO server.
However, if we need to upload a hotfix, we need to restart the server or otherwise sever the active connections in order to be able to upload a new neversion. Is it possible to only have old `GET` requests continue downloading the old file until the new file is uploaded, (in which case the older requests would still download the old version) and have all new requests download the new file? 

## Regression
No

## Your Environment
Minio instance installed with helm and the MInio Operator, running on Kubernetes. Traefik for routing.
Version `RELEASE.2022-11-11T03-44-20Z`.

harshavardhana:
## Description
fix: inlined objects don't need to honor long locks

## Motivation and Context
getObjectNInfo() currently holds the lock until all
the inlined data is written to disk.
    
It is sufficient to verify xl.meta and its quorum
validity under quorum, once that is achieved we
do not ever have to hold the lock further.
    
Currently, clients speed would dictate the time
the lock is held for READ operation on MinIO,
this is not needed and we can make this lock-free
since the inlined data is already obtained in
a single call.

## How to test this PR?
An improvement for the requirement in #16314 

## Types of changes

## Checklist:

### DIFF
--- a/cmd/batch-handlers.go
+++ b/cmd/batch-handlers.go
@@ -694,7 +694,7 @@ func (r *BatchJobReplicateV1) ReplicateToTarget(ctx context.Context, api ObjectL
 		Versioned:        versioned,
 		VersionSuspended: versionSuspended,
 	}
-	rd, err := api.GetObjectNInfo(ctx, srcBucket, srcObject, nil, http.Header{}, readLock, opts)
+	rd, err := api.GetObjectNInfo(ctx, srcBucket, srcObject, nil, http.Header{}, opts)
 	if err != nil {
 		return err
 	}
--- a/cmd/bucket-replication.go
+++ b/cmd/bucket-replication.go
@@ -1085,7 +1085,7 @@ func (ri ReplicateObjectInfo) replicateObject(ctx context.Context, objectAPI Obj
 	versioned := globalBucketVersioningSys.PrefixEnabled(bucket, object)
 	versionSuspended := globalBucketVersioningSys.PrefixSuspended(bucket, object)
 
-	gr, err := objectAPI.GetObjectNInfo(ctx, bucket, object, nil, http.Header{}, readLock, ObjectOptions{
+	gr, err := objectAPI.GetObjectNInfo(ctx, bucket, object, nil, http.Header{}, ObjectOptions{
 		VersionID:        objInfo.VersionID,
 		Versioned:        versioned,
 		VersionSuspended: versionSuspended,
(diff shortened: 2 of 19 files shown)


klauspost:
How will this not break with deletes? Or block/break healing where that require no reads on writes?

harshavardhana:
> How will this not break with deletes? Or block/break healing where that requires no reads on writes?

It will not @klauspost renames() are allowed on `open fd's` on Linux and the reads will complete on those `fd's`. 

It will also not break healing because healing only heals the missing data dir, there is one situation through which is `bit-rot` scanning that I have yet to verify.

klauspost:
@harshavardhana Multiple parts are multiple files, so a deleted part will cause a failure when downloading.

Also you create a race between reading metadata to opening files, where they could be deleted in the meantime.

Not to mention that this will completely brick Windows, since it doesn't support read+deferred writes.

harshavardhana:
> @harshavardhana Multiple parts are multiple files, so a deleted part will cause a failure when downloading.

I think then perhaps I will go back to my original thinking, that we can do this for inlined objects only @klauspost

klauspost:
@harshavardhana Yeah... You can release the lock at once if the data was returned inline. Good idea!

klauspost:
(lock downgrading - as we are discussing - would allow to write new versions while data is being streamed back)

harshavardhana:
> (lock downgrading - as we are discussing - would allow writing new versions while data is being streamed back)

Yeah wanted something that helps the current implementation this is a good step towards that, there is one more idea that came to me - let's discuss it offline.

klauspost (approved):
LGTM

vadmeste (commented):
LGTM, the only drawback is that GET object with a parallel upload can generate a 503, but still seems acceptable to me

harshavardhana:
> LGTM, the only drawback is that GET object with a parallel upload can generate a 503, but still seems acceptable to me

How can that happen @vadmeste ? Locking is actually for xl.meta reads in case of inlining.

So once we have quorum there is no competition here. The next get for another competing put will simply serialize.

We are not removing locks just changing when we need to unlock.

PR #17046, opened 2023-04-18
fix: MB unit should be 1024

jiuker:
## Description

MB unit should be 1024
Close this if on purpose.
## Types of changes

## Checklist:

### DIFF
--- a/cmd/admin-bucket-handlers.go
+++ b/cmd/admin-bucket-handlers.go
@@ -244,7 +244,7 @@ func (a adminAPIHandlers) SetRemoteTargetHandler(w http.ResponseWriter, r *http.
 	}
 
 	// enforce minimum bandwidth limit as 100MBps
-	if target.BandwidthLimit > 0 && target.BandwidthLimit < 100*1000*1000 {
+	if target.BandwidthLimit > 0 && target.BandwidthLimit < 100*1024*1024 {
 		writeErrorResponseJSON(ctx, w, errorCodes.ToAPIErrWithErr(ErrReplicationBandwidthLimitError, err), r.URL)
 		return
 	}


klauspost:
Take a look at https://en.wikipedia.org/wiki/Data-rate_unit

Megabit per second (symbol Mbit/s or Mb/s, often abbreviated "Mbps") is a unit of data transfer rate equal to:

* 1,000 kilobits per second
* 1,000,000 bits per second
* 125,000 bytes per second
* 125 kilobytes per second

I blame harddrive manufacturers in the 90s switching from base 1024 to 1000 to get it to look as if they were selling drives with bigger capacity than they had. There has been much confusion every since. Especially since the difference is bigger KiB -> KB, MiB -> MB, Gib -> GB, TiB -> TB.

jiuker:
@klauspost Why set 2x1024x1024 here?
https://github.com/minio/madmin-go/blob/main/examples/bucket-target.go#L47-L54
```go
       target := madmin.BucketTarget{
		Endpoint:       "site2:9000",
		Credentials:    creds,
		TargetBucket:   "destbucket",
		IsSSL:          false,
		Type:           madmin.ReplicationArn,
		BandwidthLimit: 2 * 1024 * 1024,
	}
```

harshavardhana:
> @klauspost Why set 2x1024x1024 here? https://github.com/minio/madmin-go/blob/main/examples/bucket-target.go#L47-L54
> 
> ```go
>        target := madmin.BucketTarget{
> 		Endpoint:       "site2:9000",
> 		Credentials:    creds,
> 		TargetBucket:   "destbucket",
> 		IsSSL:          false,
> 		Type:           madmin.ReplicationArn,
> 		BandwidthLimit: 2 * 1024 * 1024,
> 	}
> ```

Bandwidth is network speed, network speeds are measured usually in 1000's not 1024's so the verification in the code is correct and the example must be fixed.

PR #17050, opened 2023-04-19
fix:lcpSuffix(globalDomainNames) use pre instead

jiuker:
## Description
It seems test_case use true.
Here is false
test case here:
https://github.com/minio/minio/blob/8a81e317d66bcc596265db76e02defe3d300e297/cmd/utils_test.go#L399-L419
```go
	fmt.Println("output:" + lcp([]string{"127.0.0.1/yyyyy", "127.0.0.1/zzzzz"}, false))
	fmt.Println("output:" + lcp([]string{"127.0.0.1/yyyyy", "127.0.0.1/zzzzz"}, true))
	fmt.Println("output:" + lcp([]string{"127.0.0.1/yyyyy", "127.0.0.2/yyyyy"}, false))
	fmt.Println("output:" + lcp([]string{"127.0.0.1/yyyyy", "127.0.0.2/yyyyy"}, true))
```
```shell
output:
output:127.0.0.1/
output:/yyyyy
output:127.0.0.
```
here 
https://github.com/minio/minio/blob/8a81e317d66bcc596265db76e02defe3d300e297/cmd/common-main.go#L691-L697
It seems should use true instead.
Domain should like `127.0.0.1/xxxx`
## Types of changes

## Checklist:

### DIFF
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -857,7 +857,7 @@ func restQueries(keys ...string) []string {
 
 // Suffix returns the longest common suffix of the provided strings
 func lcpSuffix(strs []string) string {
-	return lcp(strs, false)
+	return lcp(strs, true)
 }
 
 func lcp(strs []string, pre bool) string {


jiuker:
@harshavardhana cc

jiuker:
@harshavardhana /cc

jiuker:
@harshavardhana how about this?

vadmeste:
@jiuker MINIO_DOMAINS will not contain any path, only domain names.

MINIO_DOMAINS is created to support virtual hosts type requests style, where the URL of the S3 requests is https://bucket-name.example.com and it is not possible to set up this url "http://bucket-name.127.0.0.1", so overall, no IP addresses can be configured in MINIO_DOMAINS

PR #17052, opened 2023-04-19
Use expontential backoff algo for internode reconnections

vadmeste:
## Description
Using expontential backoff will ensure a quick reconnect in 
the first retry with an increasing backoff in each retry. The algorithm 
will ensure that nodes reconnection will spread in a time frame 
which is larger each time.

## Motivation and Context
The current reconnect algorithm makes all nodes reconnect in a random
period less than 200 ms, which can be alot if there are many nodes/disks

The new algorithm will spread the reconnections further in a timeframe
larger in each retry.

## How to test this PR?
Hard to test

## Types of changes

## Checklist:

### DIFF
--- a/internal/rest/client.go
+++ b/internal/rest/client.go
@@ -343,8 +343,16 @@ func (c *Client) LastError() error {
 // computes the exponential backoff duration according to
 // https://www.awsarchitectureblog.com/2015/03/backoff.html
 func exponentialBackoffWait(r *rand.Rand, unit, cap time.Duration) func(uint) time.Duration {
+	if unit > time.Hour {
+		// Protect against integer overflow
+		panic("unit cannot exceed one hour")
+	}
 	return func(attempt uint) time.Duration {
-		// sleep = random_between(0, min(cap, base * 2 ** attempt))
+		if attempt > 16 {
+			// Protect against integer overflow
+			attempt = 16
+		}
+		// sleep = random_between(unit, min(cap, base * 2 ** attempt))
 		sleep := unit * time.Duration(1<<attempt)
 		if sleep > cap {
 			sleep = cap
--- a/internal/rest/client.go
+++ b/internal/rest/client.go
@@ -75,9 +75,9 @@ type Client struct {
 	// is online or offline.
 	HealthCheckFn func() bool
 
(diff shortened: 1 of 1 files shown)


harshavardhana (changes requested):
Instead of this should we do dynamic sleeper @vadmeste basically sleep at minimum of 250ms but also scale up the retry based on the response we get back from the servers?

i.e if we can detect faster we should?

internal/rest/client.go
```diff
@@ -340,6 +340,20 @@ func (c *Client) LastError() error {
...
+// computes the exponential backoff duration according to
+// https://www.awsarchitectureblog.com/2015/03/backoff.html
+func exponentialBackoffWait(r *rand.Rand, unit, cap time.Duration) func(uint) time.Duration {
+	return func(attempt uint) time.Duration {
+		// sleep = random_between(0, min(cap, base * 2 ** attempt))
+		sleep := unit * time.Duration(1<<attempt)
+		if sleep > cap {
+			sleep = cap
+		}
+		sleep -= time.Duration(r.Float64() * float64(sleep))
```
krisis: Any reason why you chose the above formula for `sleep` instead of
```go
sleep = time.Duration(r.Float64() * float64(sleep)
```

harshavardhana:
@klauspost can you PTAL at this PR ?

internal/rest/client.go
klauspost:
```suggestion
		sleep -= time.Duration(r.Float64() * float64(sleep-unit))
```
... at least with this, it makes sense (and we always sleep at least "unit"

klauspost (approved):
With a max of 30s that seems reasonable.

internal/rest/client.go
vadmeste: @krisis the point here is to specify a large time frame in each retry then each server will pick a random time in that.

harshavardhana (approved):

harshavardhana (approved):

vadmeste:
Something wrong in the PR, it looks like some integers overflow after some retries (because of the exponential, sending a fix)

harshavardhana (approved):

PR #17061, opened 2023-04-21
Remove Expect 100-continue in internode communications

vadmeste:
## Description
Requests between nodes are always successful, at least, 
in normal environments. Expect 100-continue creates an 
unnecessary additional roundtrip in internode http requests, 
therefore removing it.

It may have some performance improvement but not tested.

## How to test this PR?
Just regular operations

## Types of changes

## Checklist:

### DIFF
--- a/internal/rest/client.go
+++ b/internal/rest/client.go
@@ -191,9 +191,6 @@ func (c *Client) newRequest(ctx context.Context, u *url.URL, body io.Reader) (*h
 		req.Header.Set("Authorization", "Bearer "+c.newAuthToken(u.RawQuery))
 	}
 	req.Header.Set("X-Minio-Time", time.Now().UTC().Format(time.RFC3339))
-	if body != nil {
-		req.Header.Set("Expect", "100-continue")
-	}
 
 	if tc, ok := ctx.Value(mcontext.ContextTraceKey).(*mcontext.TraceCtxt); ok {
 		req.Header.Set(xhttp.AmzRequestID, tc.AmzReqID)


internal/rest/client.go
```diff
@@ -191,9 +191,6 @@ func (c *Client) newRequest(ctx context.Context, u *url.URL, body io.Reader) (*h
 		req.Header.Set("Authorization", "Bearer "+c.newAuthToken(u.RawQuery))
 	}
 	req.Header.Set("X-Minio-Time", time.Now().UTC().Format(time.RFC3339))
-	if body != nil {
-		req.Header.Set("Expect", "100-continue")
-	}
```
harshavardhana, line 196: This is actually done to ensure we make sure to not send the body if the server is not able to receive requests @vadmeste 

Do you have evidence on how removing this helps? this has been around for the last 3-4yrs already.

harshavardhana (changes requested):

internal/rest/client.go
vadmeste: > This is actually done to ensure we make sure to not send the body if the server is not able to receive requests @vadmeste

This is internode communication, there is no case when a peer needs to validate an incoming request before allowing to receive the body, maybe only in more edge cases, like wrong root credentials, or wrong date/time, but it is not needed in normal environments.

vadmeste: > Do you have evidence on how removing this helps? this has been around for the last 3-4yrs already.

Probably better performance for small files with fast NVME disks. For WAN setups, for sure it will have some benefits

harshavardhana: Can you do some numbers here with some delays and get numbers on why would it matter?

Even a slight change in latency, let's see what 100-continue typically adds.

harshavardhana (approved):

PR #17062, opened 2023-04-21
Support systemd notify protocol

WGH-:
## Description

Add support for systemd notification protocol (https://www.freedesktop.org/software/systemd/man/sd_notify.html)

## Motivation and Context

This will allow to set `Type=notify` in systemd service.

Thanks to this, `systemctl start` will actually wait for the ready notification instead of exiting when process has just started.

This will also allow to use `After=minio.service` for proper startup ordering. If some other service depends on MinIO, it will wait until MinIO is actually up and listening on its port.

Unfortunately, `Server.Serve` function has some additional setup going on instead of just launching serving goroutines, so in some cases systemd may report success with MinIO failing shortly after. Thankfully, listening sockets are still bound before, so proper startup ordering will still be achieved.

"Stopping" notification is mostly decorative, and will be reported in `systemctl status`.

If not running under systemd, or running without `Type=notify`, these changes have no effect.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/signals.go
+++ b/cmd/signals.go
@@ -24,6 +24,7 @@ import (
 	"os"
 	"strings"
 
+	"github.com/coreos/go-systemd/v22/daemon"
 	"github.com/minio/minio/internal/logger"
 )
 
@@ -80,17 +81,23 @@ func handleSignals() {
 			exit(stopProcess())
 		case osSignal := <-globalOSSignalCh:
 			logger.Info("Exiting on signal: %s", strings.ToUpper(osSignal.String()))
+			daemon.SdNotify(false, daemon.SdNotifyStopping)
 			exit(stopProcess())
 		case signal := <-globalServiceSignalCh:
 			switch signal {
 			case serviceRestart:
 				logger.Info("Restarting on service signal")
+				daemon.SdNotify(false, daemon.SdNotifyReloading)
 				stop := stopProcess()
 				rerr := restartProcess()
+				if rerr == nil {
+					daemon.SdNotify(false, daemon.SdNotifyReady)
+				}
 				logger.LogIf(context.Background(), rerr)
 				exit(stop && rerr == nil)
 			case serviceStop:
 				logger.Info("Stopping on service signal")
+				daemon.SdNotify(false, daemon.SdNotifyStopping)
 				exit(stopProcess())
 			}
 		}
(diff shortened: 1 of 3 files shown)


harshavardhana (changes requested):
This needs more than what is being done in the PR @WGH- are you intending to add more things here?

WGH-:
On April 21, 2023 9:29:21 PM GMT+03:00, Harshavardhana ***@***.***> wrote:
>This needs more than what is being done in the PR @WGH- are you intending to add more things here? 

I think no, at least not here. minio.service is in another repo, I think.

cmd/signals.go
```diff
@@ -79,6 +80,7 @@ func handleSignals() {
 			logger.LogIf(context.Background(), err)
 			exit(stopProcess())
 		case osSignal := <-globalOSSignalCh:
+			daemon.SdNotify(false, daemon.SdNotifyStopping)
```
harshavardhana: We should use 

```
stop := stopProcess()
daemon.SdNotify(false, daemon.SdNotifyReloading)
rerr := restartProcess()
if rerr == nil {
       daemon.SdNotify(false, daemon.SdNotifyReady)
}
```

Should be also called for "serviceRestart" IMO 

And also for `serviceSTop` we must have 

```
daemon.SdNotify(false, daemon.SdNotifyStopping)
```

harshavardhana (changes requested):

cmd/signals.go
WGH-: Right, this will allow us to use `Type=notify-reload` in the future for reload synchronization as well (introduced in latest [systemd v253](https://github.com/systemd/systemd/releases/tag/v253)).

And as of missed `case serviceStop:`, my bad, sorry.

Added this and that.

WGH-: Though I think it would be better to send "reloading" notification before calling `stopProcess`?

harshavardhana: Yes that is correct

harshavardhana: Let's not spend too much time on the nuances of reload, if we can do it properly within expected semantics then it's fine. 

Otherwise let's not add it there just add Ready when restart is complete?

harshavardhana: Sure makes sense..

harshavardhana (approved):

PR #17066, opened 2023-04-23
simplify sort.Sort by using sort.Slice

harshavardhana:
## Description
simplify sort.Sort by using sort.Slice

## Motivation and Context
move away from old style

## How to test this PR?
Nothing should change from functionality CI/CD should cover

## Types of changes

## Checklist:

### DIFF
--- a/cmd/net.go
+++ b/cmd/net.go
@@ -111,25 +111,6 @@ func getHostIP(host string) (ipList set.StringSet, err error) {
 	return ipList, err
 }
 
-// byLastOctetValue implements sort.Interface used in sorting a list
-// of ip address by their last octet value in descending order.
-type byLastOctetValue []net.IP
-
-func (n byLastOctetValue) Len() int      { return len(n) }
-func (n byLastOctetValue) Swap(i, j int) { n[i], n[j] = n[j], n[i] }
-func (n byLastOctetValue) Less(i, j int) bool {
-	// This case is needed when all ips in the list
-	// have same last octets, Following just ensures that
-	// 127.0.0.1 is moved to the end of the list.
-	if n[i].IsLoopback() {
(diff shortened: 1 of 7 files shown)


cmd/net.go
```diff
@@ -152,7 +133,18 @@ func sortIPs(ipList []string) []string {
...
+		// This case is needed when all ips in the list
+		// have same last octets, Following just ensures that
+		// 127.0.0.1 is moved to the end of the list.
+		if ipV4s[i].IsLoopback() {
+			return false
+		}
+		if ipV4s[j].IsLoopback() {
+			return true
+		}
+		return []byte(ipV4s[i].To4())[3] > []byte(ipV4s[j].To4())[3]
```
klauspost, line 146: Why only the last octet? Surely for bigger installs we could encounter several subnets?

```
		return binary.BigEndian.Uint32(ipV4s[i].To4()) > binary.LittleEndian.Uint32(ipV4s[j].To4())
```
This will make the sorting stable.

The documentation claims it is the highest octet, but it is actually sorting on the lowest (`?.?.?.X` - it only sorts on X).

If you indeed would like to sort on the highest, switch `LittleEndian` to `BigEndian` above.

You can see the different outputs here: https://go.dev/play/p/1i38P-3wh0z

klauspost: You also have `ip.IsPrivate`, which will allow you to filter out private IPs.

harshavardhana: Yeah well this code has been sort of around for years, didn't want to change that.

May be after this PR

klauspost (approved):
If you want to fix the old bugs in a separate PR, this LGTM.

krisis (approved):

PR #17077, opened 2023-04-25
helm-chart: Use minio service account for post-deploy job if available

### LINKED ISSUE #17076: Helm Chart: Don't run the post-job hook with the default service account
The post-job hook runs as the `default` service account which is not recommended. Since the helm chart has support for creating a new service account for the minio service, I suggest we use that instead if available.

LinAnt:
## Description

fixes https://github.com/minio/minio/issues/17076

### DIFF
--- a/helm/minio/templates/post-job.yaml
+++ b/helm/minio/templates/post-job.yaml
@@ -78,6 +78,9 @@ spec:
               - key: {{ .Values.tls.publicCrt }}
                 path: CAs/public.crt
         {{- end }}
+      {{ if .Values.serviceAccount.create }}
+      serviceAccountName: {{ .Values.serviceAccount.name }}
+      {{- end }}
       containers:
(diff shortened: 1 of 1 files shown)


helm/minio/templates/post-job.yaml
```diff
@@ -78,6 +78,9 @@ spec:
               - key: {{ .Values.tls.publicCrt }}
                 path: CAs/public.crt
         {{- end }}
+      {{ if .Values.serviceAccount.create }}
```
dnskr: Would you mind changing to the following to delete extra new line?
```yaml
{{- if .Values.serviceAccount.create }}
```

LinAnt: Good catch, fixed :)

harshavardhana (approved):

PR #17080, opened 2023-04-25
fix: initialize reverse proxy forwarder with right public certs

harshavardhana:
## Description
fix: initialize reverse proxy forwarder with right public certs

## Motivation and Context
public certs were missing for reverse proxy forwarder 
due to the fact that it was initialized very early on 
even before certs were loaded, move the code such that 
we only initialize the forwarder after TLS has been
successfully loaded.

## How to test this PR?
Configure two sites with self-signed TLS and then
upload one part to site1, second part goes to site2
would succeed (instead of how it fails on the master)
after this PR.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/common-main.go
+++ b/cmd/common-main.go
@@ -59,7 +59,6 @@ import (
 	"github.com/minio/minio/internal/auth"
 	"github.com/minio/minio/internal/color"
 	"github.com/minio/minio/internal/config"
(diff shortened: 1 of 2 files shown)


vadmeste (approved):
LGTM

poornas (approved):

PR #17091, opened 2023-04-27
fix: add correct worker count before startHTTPLogger()

jiuker:
## Description

`go startHTTPLogger()` will call `defer atomic.AddInt64(&h.workers, -1)` why not add 1 here.

## Types of changes

## Checklist:

### DIFF
--- a/internal/logger/target/http/http.go
+++ b/internal/logger/target/http/http.go
@@ -204,6 +204,7 @@ func (h *Target) Init() (err error) {
 		h.workerStartMu.Lock()
 		h.lastStarted = time.Now()
 		h.workerStartMu.Unlock()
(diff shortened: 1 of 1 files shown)


internal/logger/target/http/http.go
```diff
@@ -204,6 +204,7 @@ func (h *Target) Init() (err error) {
 		h.workerStartMu.Lock()
 		h.lastStarted = time.Now()
 		h.workerStartMu.Unlock()
+		atomic.AddInt64(&h.workers, 1)
```
harshavardhana, line 207: A good find, this is actually a bug.

harshavardhana (approved):

klauspost (approved):

PR #17107, opened 2023-04-28
optimization use small blocks up to 64KB

klauspost:
## Description

From https://github.com/minio/minio/discussions/17103 - files up to 1MB would be read in 32KB blocks. Set that threshold to 64KB, so at most there are 2 reads.

Remove the unneeded wrapper in readAllData when we are not using O_DIRECT - we are reading directly into a single buffer, so there is no need for an intermediate buffer. This will save a memcopy in most cases.

Either way, we now quickly reject on unsupported platforms and avoid the copying there.

Bonus: `ReadFileWithFileInfo` will now also use `NO_ATIME` flag for reads.
 
## How to test this PR?

Picked up by regular tests reliably.

## Types of changes

### DIFF
--- a/internal/ioutil/read_file.go
+++ b/internal/ioutil/read_file.go
@@ -37,9 +37,9 @@ var (
 // ReadFileWithFileInfo reads the named file and returns the contents.
 // A successful call returns err == nil, not err == EOF.
 // Because ReadFile reads the whole file, it does not treat an EOF from Read
-// as an error to be reported, additionall returns os.FileInfo
+// as an error to be reported.
 func ReadFileWithFileInfo(name string) ([]byte, fs.FileInfo, error) {
-	f, err := OsOpen(name)
+	f, err := OsOpenFile(name, readMode, 0o666)
 	if err != nil {
 		return nil, nil, err
 	}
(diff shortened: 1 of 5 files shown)


harshavardhana (approved):

internal/ioutil/read_file.go
```diff
@@ -62,6 +62,23 @@ func ReadFileWithFileInfo(name string) ([]byte, fs.FileInfo, error) {
 //
 // passes NOATIME flag for reads on Unix systems to avoid atime updates.
 func ReadFile(name string) ([]byte, error) {
+	if !disk.ODirectPlatform {
+		// Don't wrap with un-needed buffer.
+		// Don't use os.ReadFile, since it doesn't pass NO_ATIME when present.
+		f, err := OpenFileDirectIO(name, readMode, 0o666)
```
vadmeste: Shouldn't we use OpenFileDirectIO since disk.ODirectPlatform == false in this case ?

klauspost: Yeah. It falls back on the implementation, but you are completely correct.

harshavardhana (approved):

krisis (approved):

tdsn0007:
Hi, this optimization will help to read 1MB files (like veeam files)?

klauspost:
@tdsn0007 A tiny bit.

PR #17113, opened 2023-05-02
batch replication now supports arbitrary S3 targets

poornas:
When S3 is the source/target, this behaves like `mc mirror` and
 only latest version will be copied.

## Motivation and Context
Add s3 as supported target from minio <-> s3. since versioning cannot be guaranteed, this behaves like `mc mirror` on server side when one of the object stores is s3.

## How to test this PR?
In the batch replication yaml, change "type" to s3 for source or target leaving other as minio
```
replicate:
  apiVersion: v1
  # target of the objects to be replicated
  target:
    type: s3 # valid values are "minio"
    bucket: s3bucket
    prefix: xx # 'PREFIX' is optional
    # NOTE: if source is remote then target must be "local"
    endpoint: https://s3.amazonaws.com
    credentials:
      accessKey: xxx
      secretKey: yyyyyy
  #  #   sessionToken: SESSION-TOKEN # Optional only available when rotating credentials are used

  # source from where the objects must be replicated
  source:
    type: minio # valid values are "minio"
    bucket: tbucketvr-cp
    prefix: pdx # 'PREFIX' is optional
    # NOTE: if target is remote then source must be "local"
    # endpoint: http://localhost:9004
    # credentials:
    #   accessKey: minio
    #   secretKey: minio123
    # #   sessionToken: SESSION-TOKEN # Optional only available when rotating credentials are used

    retry:
```
## Types of changes

## Checklist:

### DIFF
--- a/cmd/batch-handlers.go
+++ b/cmd/batch-handlers.go
@@ -174,6 +174,7 @@ type BatchJobReplicateResourceType string
 func (t BatchJobReplicateResourceType) Validate() error {
 	switch t {
 	case BatchJobReplicateResourceMinIO:
+	case BatchJobReplicateResourceS3:
 	default:
 		return errInvalidArgument
 	}
@@ -183,6 +184,8 @@ func (t BatchJobReplicateResourceType) Validate() error {
 // Different types of batch jobs..
 const (
 	BatchJobReplicateResourceMinIO BatchJobReplicateResourceType = "minio"
+	BatchJobReplicateResourceS3    BatchJobReplicateResourceType = "s3"
+
 	// add future targets
 )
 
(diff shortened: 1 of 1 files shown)


cmd/batch-handlers.go
```diff
@@ -539,19 +550,42 @@ func (r *BatchJobReplicateV1) StartFromSource(ctx context.Context, api ObjectLay
...
 		for obj := range objInfoCh {
 			oi := toObjectInfo(r.Source.Bucket, obj.Key, obj)
+			if !minioSrc {
+				oi2, err := c.StatObject(ctx, r.Source.Bucket, obj.Key, minio.StatObjectOptions{})
+				if err == nil {
+					oi = toObjectInfo(r.Source.Bucket, obj.Key, oi2)
+				} else {
+					logger.LogIf(ctx, err)
+					cancel()
+					return err
```
harshavardhana, line 579: No need to return an error for objects that got deleted @poornas such as 404?

harshavardhana (changes requested):

cmd/batch-handlers.go
poornas: good point, updated the PR.

cmd/batch-handlers.go
```diff
@@ -539,19 +550,46 @@ func (r *BatchJobReplicateV1) StartFromSource(ctx context.Context, api ObjectLay
...
+				}
+			}
 			if skip(oi) {
 				continue
 			}
+			if obj.Key != prevObj {
+				prevObj = obj.Key
+				skipReplicate = obj.IsDeleteMarker && s3Type
+			}
```
harshavardhana: Let's add a comment here to indicate that we are actually skipping DEL marked objects, basically the latest is DEL marker? right.

harshavardhana (changes requested):

harshavardhana (approved):

PR #17121, opened 2023-05-03
Support persistent queue store for loggers

Praveenrajmani:
## Description

This commit adds the queue store persistence support for MinIO loggers. This enables persisting the logs to the configured queue store which will be replayed by the store replay mechanism. This is similar to the queue store support for bucket notification targets.

This will make sure the logs will not be missed when the logger targets are down. The logs will be persisted in such cases and will be replayed once the target is up.

## Motivation and Context

To persist the logs so that no events will be missed when the target is down. Those interim events will be replayed back when the target is back online.

## How to test this PR?

- Setup webhook logger / audit webhook / audit kafka
- Enable the persistent store in the loggers. For example - `mc admin config set myminio audit_webhook enable=on endpoint=http://localhost:8080/ queue_size=100000 queue_dir=/tmp/webhook-logger`
- Bring down the target
- Generate logs - any s3 operations.
- Check the queue_dir - `mc tree --files /tmp/webhook-logger` - you should be seeing few log files persisted.
- Bring up the target
- After few secs, the events should have been replayed to the target and the queue store should be empty

(NOTE: For configuring kafka for testing, you can refer https://kafka.apache.org/quickstart)

## Types of changes

## Checklist:


internal/store/store.go
```diff
@@ -53,7 +54,7 @@ type Store[I any] interface {
 }
 
 // replayItems - Reads the items from the store and replays.
-func replayItems[I any](store Store[I], doneCh <-chan struct{}, loggerOnce logger.LogOnce, id string) <-chan string {
+func replayItems[I any](store Store[I], doneCh <-chan struct{}, logger logger, id string) <-chan string {
```
anjalshireesh, line 57: `logger logger` - It might be better to have the variable name different from it's type.

internal/store/store.go
```diff
@@ -24,21 +24,22 @@ import (
...
 
-	"github.com/minio/minio/internal/logger"
 	xnet "github.com/minio/pkg/net"
 )
 
 const (
 	retryInterval = 3 * time.Second
 )
 
+type logger = func(ctx context.Context, err error, id string, errKind ...interface{})
```
anjalshireesh, line 34: Is this same as `logger.LogOnce` ?

internal/logger/config.go
```diff
@@ -395,6 +421,25 @@ func lookupAuditKafkaConfig(scfg config.Config, cfg Config) (Config, error) {
 		kafkaArgs.SASL.Password = env.Get(saslPasswordEnv, kv.Get(KafkaSASLPassword))
 		kafkaArgs.SASL.Mechanism = env.Get(saslMechanismEnv, kv.Get(KafkaSASLMechanism))
 
+		queueDirEnv := EnvKafkaQueueDir
+		if k != config.Default {
+			queueDirEnv = queueDirEnv + config.Default + k
+		}
+		kafkaArgs.QueueDir = env.Get(queueDirEnv, kv.Get(KafkaQueueDir))
```
anjalshireesh, line 428: This pattern seems to be repeating quite a few times. Can we introduce a function, something like:

```
func getCfgVal(kv config.KVS, k string, key string, defaultEnv string) string {
	e := defaultEnv
	if k != config.Default {
		e = e + config.Default + k
	}
	return env.Get(e, kv.Get(key))
}
```
and use it e.g.
`kafkaArgs.QueueDir = getCfgVal(kv, k, KafkaQueueDir, EnvKafkaQueueDir)`

?

internal/logger/target/http/http.go
```diff
@@ -125,54 +144,42 @@ func (h *Target) Stats() types.TargetStats {
 	return stats
 }
 
+// This will check if we can reach the remote.
+func (h *Target) isAlive(ctx context.Context) (err error) {
```
anjalshireesh, line 148: The previous name (`checkAlive`) sounds better here as it doesn't return `bool`

internal/logger/target/http/http.go
```diff
@@ -205,19 +212,51 @@ func (h *Target) Init() (err error) {
...
 
-// Accepted HTTP Status Codes
-var acceptedStatusCodeMap = map[int]bool{http.StatusOK: true, http.StatusCreated: true, http.StatusAccepted: true, http.StatusNoContent: true}
+func (h *Target) send(ctx context.Context, payload []byte, timeout time.Duration) (err error) {
+	ctx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost,
+		h.config.Endpoint, bytes.NewReader(payload))
+	if err != nil {
+		return fmt.Errorf("%s returned '%w', please check your endpoint configuration", h.config.Endpoint, err)
```
anjalshireesh, line 226: This error message might not be correct here as the request hasn't been sent to the endpoint yet..

krisis (approved):

harshavardhana (approved):

harshavardhana:
@Praveenrajmani feel free to send PRs for what @anjalshireesh reviewed if it makes sense.

internal/store/store.go
Praveenrajmani: yes, we need to remove the `logger` dependency here as it is cyclic.

PR #17128, opened 2023-05-04
optimize randomly generating AK and SK for new users

### LINKED ISSUE #4562: 'mc admin svcacct add' can't generate random secret if access-key specified
`mc admin user svcacct add` allows randomly generating both access key and secret key by omitting `--access-key` and `--secret-key` arguments. Example:
```
$ mc admin user svcacct add myminio myexistinguser
Access Key: random-access-key-value
Secret Key: random-secret-key-value
```

But if you provide the access key and omit the secret, it returns an error:
```
$ mc admin user svcacct add myminio myexistinguser --access-key "mynewsvcacct"
mc: <ERROR> Unable to add a new service account. The secret key is invalid. (secret key length should be between 8 and 40).
```

Additionally, if you provide only the secret you get random values for both:
```
$ mc admin user svcacct add myminio myexistinguser --secret-key "mysecret"
Access Key: random-access-key-value
Secret Key: random-secret-key-value
```

## Expected Behavior

`mc admin user svcacct add myminio myexistinguser --access-key "mynewsvcacct"` creates a service account with the given access key and a random secret.

## Current Behavior

`mc admin user svcacct add myminio myexistinguser --access-key "mynewsvcacct"` returns an error and the service account is not created.

## Possible Solution

Allow creating a random secret for a given access key. 

Random access key for provided secret could be allowed, but doesn't seem that useful. Ignoring explicit user input (the `--secret-key`) isn't great however.

Nice to have: consistent behavior between `mc admin user svcacct add` (random allowed) and `mc admin user add` (both values required).

1. `mc admin user svcacct add myminio myexistinguser --access-key "mynewsvcacct"`

## Context

Discovered while updating documentation, discussed with @harshavardhana and @vadmeste 

## Regression

Not that I am aware of.

## Your Environment

Ubuntu VM with multipass, on macOS
```
$ minio --version
minio version RELEASE.2023-04-20T17-56-55Z (commit-id=c61c4b71b26aad048b8f3abe0ee24547964fc49f)
Runtime: go1.20.3 linux/arm64
Copyright: 2015-2023 MinIO, Inc.
```
Guest OS
```
$ uname -a
Linux object-lambda 5.15.0-70-generic minio/minio#77-Ubuntu SMP Tue Mar 21 15:58:51 UTC 2023 aarch64 aarch64 aarch64 GNU/Linux
```
Host OS
```
$ uname -a
Darwin Andreas-MacBook-Pro.local 22.3.0 Darwin Kernel Version 22.3.0: Mon Jan 30 20:39:46 PST 2023; root:xnu-8792.81.3~2/RELEASE_ARM64_T6020 arm64
```

dormanze:
fix [#4562](https://github.com/minio/mc/issues/4562)

## Types of changes

## Checklist:

### DIFF
--- a/cmd/iam.go
+++ b/cmd/iam.go
@@ -975,14 +975,17 @@ func (sys *IAMSys) NewServiceAccount(ctx context.Context, parentUser string, gro
 		}
 	}
 
-	var accessKey, secretKey string
-	var err error
-	if len(opts.accessKey) > 0 {
-		accessKey, secretKey = opts.accessKey, opts.secretKey
-	} else {
-		accessKey, secretKey, err = auth.GenerateCredentials()
+	accessKey, secretKey := opts.accessKey, opts.secretKey
+	if len(accessKey) <= 0 || len(secretKey) <= 0 {
+		randomAccessKey, randomSecretKey, err := auth.GenerateCredentials()
 		if err != nil {
 			return auth.Credentials{}, time.Time{}, err
+		}
+		if len(opts.accessKey) <= 0 {
+			accessKey = randomAccessKey
+		}
+		if len(opts.secretKey) <= 0 {
+			secretKey = randomSecretKey
 		}
 	}
 	cred, err := auth.CreateNewCredentialsWithMetadata(accessKey, secretKey, m, secretKey)


shtripat (commented):
LGTM. Verified

harshavardhana:
@dormanze this is actually not a server fix, its actually a client fix. We do not want too much flexibility here from a server API point of view.

dormanze:
> @dormanze this is actually not a server fix, its actually a client fix. We do not want too much flexibility here from a server API point of view.
Is the interface designed to allow users to transfer AK and SK at the same time? Otherwise, it will be randomly generated?
If this is the case, the client needs to verify parameters, and the server needs to restrict the user to transfer both the AK and SK. Instead of separately processing scenarios where only AK are transferred.
What do you think?

harshavardhana:
> > @dormanze this is actually not a server fix, its actually a client fix. We do not want too much flexibility here from a server API point of view.
> > Is the interface designed to allow users to transfer AK and SK at the same time? Otherwise, it will be randomly generated?
> > If this is the case, the client needs to verify parameters, and the server needs to restrict the user to transfer both the AK and SK. Instead of separately processing scenarios where only AK are transferred.
> > What do you think?

Server assumes that if client sets access key it has set secret key. It is not optional - this needs to be addressed on the mc side.

PR #17141, opened 2023-05-04
disallow EC:0 if not set during server startup

harshavardhana:
## Description
disallow EC:0 if not set during server startup

## Motivation and Context
This PR adds also new node-level prometheus
metrics for any alerts if the configured values
are not what it expected.

cluster health check also returns this value
if the storage class/config is not initialized
as an additional HTTP header.

## How to test this PR?
Manual delays are needed to reproduce the problem
this PR would fix this behavior entirely.

## Types of changes

## Checklist:


vadmeste (approved):
LGTM

PR #17142, opened 2023-05-04
Validate if replication config being added is self referential

poornas:
## Motivation and Context
Existing checks to compare target endpoint to local cluster endpoint not sufficient to catch issues if a load balancer endpoint configured incorrectly to point to source cluster.

## How to test this PR?
set up sidekick in front of a minio instance - use this endpoint to configure replication. It should throw an error 

## Types of changes

## Checklist:

### DIFF
--- a/cmd/bucket-replication.go
+++ b/cmd/bucket-replication.go
@@ -26,6 +26,7 @@ import (
 	"math"
 	"math/rand"
 	"net/http"
+	"net/url"
 	"path"
 	"reflect"
 	"strings"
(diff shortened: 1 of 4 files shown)


cmd/bucket-replication.go
```diff
@@ -154,6 +163,46 @@ func validateReplicationDestination(ctx context.Context, bucket string, rCfg *re
...
+	}
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return err
+	}
+	if err == nil {
+		// Drain the connection.
+		io.Copy(ioutil.Discard, resp.Body)
+		resp.Body.Close()
```
harshavardhana:
```
xhttp.DrainBody(resp.Body)
```

harshavardhana (changes requested):

cmd/bucket-replication.go
```diff
@@ -154,6 +163,46 @@ func validateReplicationDestination(ctx context.Context, bucket string, rCfg *re
...
+		Timeout:   10 * time.Second,
+	}
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return err
+	}
+	if err == nil {
+		// Drain the connection.
+		io.Copy(ioutil.Discard, resp.Body)
```
harshavardhana: DrainBody already does the io.Copy() part.

harshavardhana (changes requested):

harshavardhana (approved):

PR #17168, opened 2023-05-09
fix: avoid audit log race protection deadlocks

klauspost:
## Description

Since we hold a read lock on reqInfo, if a write lock request comes in before calling `reqInfo.GetTagsMap()` it will deadlock.

A historical bug 9004d69c6f 

## Types of changes


harshavardhana (approved):

vadmeste (approved):
LGTM

harshavardhana (approved):

PR #17170, opened 2023-05-09
fix: start using pkg/workers to spawn parallel workers

harshavardhana:
## Description
fix: start using pkg/workers to spawn parallel workers

## Motivation and Context
use helper workers package

## How to test this PR?
Nothing should change CI/CD must test it. 

## Types of changes

## Checklist:

### DIFF
--- a/cmd/erasure-server-pool-decom.go
+++ b/cmd/erasure-server-pool-decom.go
@@ -714,6 +714,10 @@ func (z *erasureServerPools) decommissionPool(ctx context.Context, idx int, pool
 		return err
 	}
 
+	// each set get its own thread separate from the concurrent
+	// objects/versions being decommissioned.
+	workerSize += len(pool.sets)
+
 	wk, err := workers.New(workerSize)
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


cmd/erasure-server-pool-decom.go
```diff
@@ -902,21 +901,19 @@ func (z *erasureServerPools) decommissionPool(ctx context.Context, idx int, pool
 			z.poolMetaMutex.Unlock()
 		}
 
-		wg.Add(1)
+		wk.Take()
```
krisis, line 908: We can manage the concurrent, set-level listing of objects to be decommissioned separate from the concurrent decommissioning of objects/versions. We can continue to use `sync.WaitGroup` to manage the concurrent listing as before and use `internal/workers` only for decommissioning entries.

This is not to say that you can't use the same worker-pool to do both. If you wish to go that route, then we should account for one worker per set in addition to what the user configured via the environment variable.
i.e, 
```go
  workerSize += len(pool.sets)
```

krisis (approved):

poornas (approved):

PR #17178, opened 2023-05-11
fix: missing http status error in discovery doc parsing function

deluxghost:
## Description

value of `err` under this condition is `nil`, the return values will be an empty `DiscoveryDoc` and a `nil` error. This will cause outer function skip the err check

## Motivation and Context

I keep seeing this error `Error: Unable to initialize OpenID: no JWKS URI found in your provider's discovery doc` at minio start, but the url is fine. Turns out my web firewall blocks any non-human request with 403 code, but minio's log doesn't mention the status code at all. Thus it's harder to find the root cause of the error

## How to test this PR?

set the openid config url, make it response 4XX code. minio should show error log about the status code instead of missing `jwks_url`

## Types of changes

## Checklist:

### DIFF
--- a/docs/sts/web-identity.go
+++ b/docs/sts/web-identity.go
@@ -90,7 +90,7 @@ func parseDiscoveryDoc(ustr string) (DiscoveryDoc, error) {
 	}
 	defer resp.Body.Close()
 	if resp.StatusCode != http.StatusOK {
-		return d, err
+		return d, errors.New(resp.Status)
 	}
 	dec := json.NewDecoder(resp.Body)
 	if err = dec.Decode(&d); err != nil {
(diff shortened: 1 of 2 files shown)


harshavardhana:
We need a better error `errors.New(resp.Status)` won't work since that is even more intelligible. 

Let us fix it for you, this PR is not complete.

deluxghost:
> We need a better error `errors.New(resp.Status)` won't work since that is even more intelligible.
> 
> Let us fix it for you, this PR is not complete.

I use `errors.New(resp.Status)` because I saw them other places. It's good to use a better error message and also improve other status errors. Thanks!

deluxghost:
@harshavardhana do you have an issue for this?

harshavardhana:
Please open an issue @deluxghost thanks

PR #17180, opened 2023-05-11
Prevent the default configuration from being deleted

jiuker:
## Description

Prevent the default configuration from being deleted
When deleting the default, you will find that it can be deleted successfully and minio will be restarted, but the default does not disappear when the page is displayed and the program is actually running, so why don't we just organize the deletion of the default?
Before:
企业微信截图_4b27a6e6-3fe5-4575-b58f-3e945414d80f
Do delete:
企业微信截图_71e156b1-0f69-484a-80c6-7ae93159bf6c
After:
企业微信截图_f303ef56-edf3-4890-bd38-5e6557aa05df

## Types of changes

## Checklist:

### DIFF
--- a/cmd/admin-handlers-idp-config.go
+++ b/cmd/admin-handlers-idp-config.go
@@ -354,6 +354,11 @@ func (a adminAPIHandlers) DeleteIdentityProviderCfg(w http.ResponseWriter, r *ht
 	var subSys string
 	switch idpCfgType {
 	case madmin.OpenidIDPCfg:
+		// idpCfgType is OpenidIDPCfg,Can't delete _ (Default)
+		if cfgName == madmin.Default {
+			writeErrorResponseJSON(ctx, w, errorCodes.ToAPIErr(ErrAdminInvalidArgument), r.URL)
+			return
+		}
 		subSys = config.IdentityOpenIDSubSys
 		cfgInfos, err := globalIAMSys.OpenIDConfig.GetConfigInfo(cfgCopy, cfgName)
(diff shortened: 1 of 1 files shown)


cmd/admin-handlers-idp-config.go
```diff
@@ -354,6 +354,11 @@ func (a adminAPIHandlers) DeleteIdentityProviderCfg(w http.ResponseWriter, r *ht
 	var subSys string
 	switch idpCfgType {
 	case madmin.OpenidIDPCfg:
+		// idpCfgType is OpenidIDPCfg,Can't delete _ (Default)
+		if cfgName == madmin.Default {
+			writeErrorResponseJSON(ctx, w, errorCodes.ToAPIErr(ErrAdminInvalidArgument), r.URL)
+			return
+		}
```
harshavardhana, line 361: this is an incorrect interpretation, default can be configured with the relevant target. A delete of default simply means "reset" it to default values. 

Each sub-system always has as `default` target of `_` that can never be "deleted" - there is nothing wrong here.

harshavardhana: Console UI should say "resetting" Default target to its default values.

harshavardhana (changes requested):

cmd/admin-handlers-idp-config.go
jiuker: @harshavardhana No, I deleted _ and restarted minio, as shown in minio's log

PR #17182, opened 2023-05-11
Freeze the s3 APIs until the notification sub-system initializes completely

Praveenrajmani:
## Description

Currently, we will miss the events if MinIO server receives the API requests before the notification sub-system is initialized and the bucket target rules are populated.

This PR will freeze the APIs until these subsystems get completely initialized.

## Motivation and Context

- No events should me missed on server restarts
- No interim events should be missed between the bootstrap and notification-subsystem initialization.

## How to test this PR?

- Configure bucket notifications with or without queue_dir
- In `master` apply the following diff to simulate the config subsystem and bucket target rules time lag.

```diff
diff --git a/cmd/server-main.go b/cmd/server-main.go
index e212e13e2..4f5e047e8 100644
--- a/cmd/server-main.go
+++ b/cmd/server-main.go
@@ -417,6 +417,8 @@ func initServer(ctx context.Context, newObject ObjectLayer) error {
 
        r := rand.New(rand.NewSource(time.Now().UnixNano()))
 
+       time.Sleep(3*time.Minute)
+
        for {
                select {
                case <-ctx.Done():
```
- Start MinIO - it will be sleeping for 3 minutes.
- Copy few objects to MinIO - it will be successful but no events can be seen or persisted for those objects.
- With this change in the PR, the API's will freezed until the targets and configs are initialized and then unfreezes once done. Here, we won't miss any events.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/server-main.go
+++ b/cmd/server-main.go
@@ -633,47 +633,27 @@ func serverMain(ctx *cli.Context) {
 		logger.Info(color.RedBold(msg))
 	}
 
-	go func() {
-		if !globalDisableFreezeOnBoot {
-			// Freeze the services until the bucket notification subsystem gets initialized.
-			freezeServices()
-			defer unfreezeServices()
+	if !globalDisableFreezeOnBoot {
+		// Freeze the services until the bucket notification subsystem gets initialized.
+		freezeServices()
+	}
 
-			t := time.AfterFunc(5*time.Minute, func() {
-				logger.Info(color.Yellow("WARNING: Taking more time to initialize the config subsystem. Please set 'MINIO_DISABLE_API_FREEZE_ON_BOOT=true' to not freeze the APIs"))
-			})
-			defer t.Stop()
+	bootstrapTrace("initializing the server")
+	if err = initServer(GlobalContext, newObject); err != nil {
+		var cerr config.Err
+		// For any config error, we don't need to drop into safe-mode
+		// instead its a user error and should be fixed by user.
+		if errors.As(err, &cerr) {
+			logger.FatalIf(err, "Unable to initialize the server")
 		}
 
-		bootstrapTrace("initializing the server")
-		if err = initServer(GlobalContext, newObject); err != nil {
-			var cerr config.Err
-			// For any config error, we don't need to drop into safe-mode
-			// instead its a user error and should be fixed by user.
-			if errors.As(err, &cerr) {
-				logger.FatalIf(err, "Unable to initialize the server")
-			}
-
-			// If context was canceled
-			if errors.Is(err, context.Canceled) {
-				logger.FatalIf(err, "Server startup canceled upon user request")
-			}
-
-			logger.LogIf(GlobalContext, err)
+		// If context was canceled
+		if errors.Is(err, context.Canceled) {
+			logger.FatalIf(err, "Server startup canceled upon user request")
 		}
 
-		// Initialize bucket notification system.
-		logger.LogIf(GlobalContext, globalEventNotifier.InitBucketTargets(GlobalContext, newObject))
-
-		// List buckets to initialize bucket metadata sub-sys.
-		buckets, err := newObject.ListBuckets(GlobalContext, BucketOptions{})
-		if err != nil {
-			logger.LogIf(GlobalContext, fmt.Errorf("Unable to list buckets to initialize bucket metadata sub-system: %w", err))
-		}
-
-		// Initialize bucket metadata sub-system.
-		globalBucketMetadataSys.Init(GlobalContext, buckets, newObject)
-	}()
+		logger.LogIf(GlobalContext, err)
+	}
 
 	// Initialize users credentials and policies in background right after config has initialized.
 	go func() {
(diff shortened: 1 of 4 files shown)


cmd/server-main.go
```diff
@@ -633,22 +633,41 @@ func serverMain(ctx *cli.Context) {
...
-		var cerr config.Err
-		// For any config error, we don't need to drop into safe-mode
-		// instead its a user error and should be fixed by user.
-		if errors.As(err, &cerr) {
-			logger.FatalIf(err, "Unable to initialize the server")
+	var buckets []BucketInfo
+	func() {
+		// Freeze the services until the bucket notification subsystem gets initialized.
+		freezeServices()
+		defer unfreezeServices()
```
harshavardhana: One change we can do here is that for some reason `initServer` or any other Init that fails or for some reason blocks - we shouldn't block S3 API fully forever.

Have some kind of a timeout on this? - i.e allow a provision to restart the pod/service with some hidden ENV that ignores freezing calls?

harshavardhana (changes requested):

cmd/server-main.go
```diff
@@ -633,22 +633,41 @@ func serverMain(ctx *cli.Context) {
...
 	}
 
-	bootstrapTrace("initializing the server")
-	if err = initServer(GlobalContext, newObject); err != nil {
-		var cerr config.Err
-		// For any config error, we don't need to drop into safe-mode
-		// instead its a user error and should be fixed by user.
-		if errors.As(err, &cerr) {
-			logger.FatalIf(err, "Unable to initialize the server")
+	var buckets []BucketInfo
```
harshavardhana: you can't modify `buckets` slice inside a go-routine while someone can read from it in the meantime.? 

Whoever needs the list of buckets must also be initialized in the same goroutine @Praveenrajmani

harshavardhana (changes requested):

cmd/server-main.go
Praveenrajmani: okay

cmd/server-main.go
Praveenrajmani: It is not a go-routine - just a functional block. 

We wait for the notification subsystem to be done and then unfreeze and proceed with other initializations.

harshavardhana: Well that makes this blocking we don't want it blocking server startup

harshavardhana (changes requested):

cmd/server-main.go
harshavardhana: what we need is for this to be run in a goroutine but freeze it and let the server start, admin calls are possible to verify what is going on.

That is the advantage of using freeze is that we can start and admin operations can be performed however all S3 operations are blocked.

Praveenrajmani: understood

Praveenrajmani:
@harshavardhana Can we print the startup message once all the initializations return? Because, with this change, we can't see `SQS ARNs` as the config wouldn't have been loaded yet.

Can we display the startup message only after all the go-routines return (ie, wg.Wait)

harshavardhana:
This is not yet correct @Praveenrajmani let me send an update to this branch.

cmd/server-main.go
```diff
@@ -439,6 +439,10 @@ func initServer(ctx context.Context, newObject ObjectLayer) error {
 				// These messages only meant primarily for distributed setup, so only log during distributed setup.
 				logger.Info("All MinIO sub-systems initialized successfully in %s", time.Since(t1))
 			}
+
+			// Initialize bucket notification system.
+			logger.LogIf(GlobalContext, globalEventNotifier.InitBucketTargets(GlobalContext, newObject))
```
Praveenrajmani: @harshavardhana here, we should also call 

```go
// List buckets to initialize bucket metadata sub-sys.
buckets, err := newObject.ListBuckets(GlobalContext, BucketOptions{})
if err != nil {
	logger.LogIf(GlobalContext, fmt.Errorf("Unable to list buckets to initialize bucket metadata sub-system: %w", err))
}

// Initialize bucket metadata sub-system.
globalBucketMetadataSys.Init(GlobalContext, buckets, newObject)
```

otherwise, the `bucketRulesMap` will be empty and we will miss the events.

cmd/server-main.go
```diff
@@ -413,7 +413,7 @@ func initServer(ctx context.Context, newObject ObjectLayer) error {
 	t1 := time.Now()
 
 	// Once the config is fully loaded, initialize the new object layer.
-	setObjectLayer(newObject)
+	defer setObjectLayer(newObject)
```
Praveenrajmani: looks like the mint tests are failing due to this 

```
{
  "alert": "",
  "args": {
    "bucketName": "aws-sdk-go-test-xuin4gr4m14f0l",
    "expiry": 60000000000,
    "objectName": "presignedTest"
  },
  "duration": 364831,
  "error": "XMinioServerNotInitialized: Server not initialized, please try again.\n\tstatus code: 503, request id: 175FCA4CA107CB03, host id: 87ef3db1a67a60b69bcfee10c0874f8a253fc9bc1bf4e58d7314742c975c4eb6",
  "function": "PresignedPut",
  "message": "AWS SDK Go CreateBucket Failed",
  "name": "aws-sdk-go",
  "status": "FAIL"
}
```

Do we need to fix the tests? or add the API freeze to handle this?

harshavardhana: Actually will take a look why. They should be retried.

harshavardhana:
@Praveenrajmani see what I wanted to do here. Mostly this would work if you can give this a test.

Praveenrajmani:
yes, this works @harshavardhana

harshavardhana (approved):

vadmeste (approved):
LGTM

poornas:
LGTM

PR #17184, opened 2023-05-11
do not count ILM expired objects and other skipped objects

harshavardhana:
## Description
do not count ILM expired objects and other skipped objects

## Motivation and Context
Additionally skip decommissioning of .usage-cache.bin, 
this should be regenerated by scanner.

## How to test this PR?
decommission on pools with expired ILM objects,
we wrongly conclude that decommission failed.

## Types of changes

## Checklist:


cmd/erasure-server-pool-decom.go
```diff
@@ -778,6 +778,11 @@ func (z *erasureServerPools) decommissionPool(ctx context.Context, idx int, pool
 					continue
 				}
 
+				if bi.Name == minioMetaBucket && strings.Contains(version.Name, dataUsageCacheName) {
```
vadmeste: Not sure I understand why ? the code is not ready to decomission data usage cache ?

cmd/erasure-server-pool-decom.go
```diff
@@ -1003,17 +1008,76 @@ func (z *erasureServerPools) checkAfterDecom(ctx context.Context, idx int) error
...
+				// to create the appropriate stack.
+				versionsSorter(fivs.Versions).reverse()
+
+				for _, version := range fivs.Versions {
+					// Apply lifecycle rules on the objects that are expired.
+					if filterLifecycle(bi.Name, version.Name, version) {
+						continue
+					}
+
+					if bi.Name == minioMetaBucket && strings.Contains(version.Name, dataUsageCacheName) {
```
vadmeste, line 1059: This can be outside for block, maybe just after

```
				if !entry.isObject() {
					return
				}
```

cmd/erasure-server-pool-decom.go
```diff
@@ -1003,17 +1008,76 @@ func (z *erasureServerPools) checkAfterDecom(ctx context.Context, idx int) error
...
+						// skipping bucket usage cache name
+						continue
+					}
+
+					// We will skip decommissioning delete markers
+					// with single version, its as good as there
+					// is no data associated with the object.
+					if version.Deleted && len(fivs.Versions) == 1 {
```
vadmeste: This is a little problematic. Let's assume decom was suspeneded in the middle and there is one delete marker left for one object in the pool that we want to decomission, then it should be decom as well and not ignored.

vadmeste (commented):
Few comments

cmd/erasure-server-pool-decom.go
harshavardhana: yes, why would we need to do it? @vadmeste This file is autogenerated, it is used as a temporary value until we write usage.json. There is no reason to move it.

harshavardhana:
PR updated @vadmeste

vadmeste (approved):
LGTM

poornas (approved):

PR #17186, opened 2023-05-11
fix: close the notification and lambda targets after testing

Praveenrajmani:
## Description

`fetchSubSysTargets` function may return a non-empty target list even if `err != nil` - this is because we fetch all the targets from a subsystem (say, notify_webhook) and return err even if one target is down. Targets in this non-empty list should be properly closed. Otherwise we will be leaking some go-routines and connections.

## Motivation and Context

Do not leak connections or go-routines and properly close the subsystem targets after testing.

## How to test this PR?

We can't really test this change. 

## Types of changes

## Checklist:


Praveenrajmani:
Just saw that we return nil if the target is down. This PR is not required.

PR #17192, opened 2023-05-12
fix: make slice cap of right size

jiuker:
## Description

make slice cap right
It seems should be `len*4` not `len`

## Types of changes

## Checklist:

### DIFF
--- a/cmd/metrics-v2.go
+++ b/cmd/metrics-v2.go
@@ -2380,7 +2380,7 @@ func getWebhookMetrics() *MetricsGroup {
 	}
 	mg.RegisterRead(func(ctx context.Context) []Metric {
 		tgts := append(logger.SystemTargets(), logger.AuditTargets()...)
-		metrics := make([]Metric, 0, len(tgts))
+		metrics := make([]Metric, 0, len(tgts)*4)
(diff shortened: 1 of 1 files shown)


cmd/metrics-v2.go
```diff
@@ -2380,7 +2380,7 @@ func getWebhookMetrics() *MetricsGroup {
 	}
 	mg.RegisterRead(func(ctx context.Context) []Metric {
 		tgts := append(logger.SystemTargets(), logger.AuditTargets()...)
-		metrics := make([]Metric, 0, len(tgts))
+		metrics := make([]Metric, 0, len(tgts)*4)
 		for _, t := range tgts {
```
harshavardhana, line 2384: It doesn't really matter here, this is just an optimization.

Eventual slice could be always bigger. Not sure where the 4* comes here.

So this change is not needed.

harshavardhana (changes requested):

klauspost:
@harshavardhana Each loop writes exactly 4 times per target, so the capacity is exact.

klauspost (approved):

harshavardhana (approved):

Issue #17195, opened 2023-05-12
SOS API (Veeam) system.xml is not compatible with Veeam


omarkohl:
## Expected Behavior

Veeam recognizes MinIO as providing SOS API, but it does not.

## Current Behavior

Manually requesting the system.xml object works, but there are some errors in it, which is why Veeam does not accept it.

## Possible Solution

<div class="table-wrap">

XML Tag | Change | Description
-- | -- | --
ProtocolVersion | add enclosing double quotes | That's how it is described in the documentation and it's not working without it.
ModelName | add enclosing double quotes | That's how it is described in the documentation and recommended by Veeam (even if it is currently working without it).
APIEndpoints | remove this tag including subtags | The tag is optional. Since IAM and STS processing is not supported it should be removed.
SystemRecommendations | remove this tag including subtags | The tag is optional and should not be there, unless there are differences in Veeam defaults that have been previously discussed with Veeam.

</div>

<p></p>

I'll submit a pull request.

## Steps to Reproduce (for bugs)

1. Follow "Appendix C: Manual plug-in load test" of the "Veeam Smart Object Storage API SOSAPI v1.0" available in the Veeam partner portal
2. Notice that in the last step "If everything was implemented correctly you will the capacity information when you create a new backup job" does not work i.e. no capacity is displayed.

After the changes I propose, the capacity is displayed correctly.

## Your Environment

* Version used (`minio --version`): RELEASE.2023-05-04T21-44-30Z
* Server setup and configuration:
* Operating System and version (`uname -a`):

PR #17196, opened 2023-05-12
Fix Veeam SOS API system.xml (#17195)

omarkohl:
The system.xml should look like this:

    <SystemInfo>
      <ProtocolVersion>"1.0"</ProtocolVersion>
      <ModelName>"MinIO RELEASE.2023-04-28T18-11-17Z"</ModelName>
      <ProtocolCapabilities>
        <CapacityInfo>true</CapacityInfo>
        <UploadSessions>false</UploadSessions>
        <IAMSTS>false</IAMSTS>
      </ProtocolCapabilities>
    </SystemInfo>

* ProtocolVersion needs to be in double quotes according to the documentation and my tests confirm Veeam won't accept it otherwise.
* ModelName needs to be double quotes according to the documentation (though, it does work without).
* APIEndpoints should be removed because IAM and STS processing is not supported by MinIO. The tag is optional according to the documentation.
* SystemRecommendations should be removed because it is optional and must only be specified if there are differences from Veeam defaults that have been discussed with Veeam. The values previously in there were not valid.

Before this change Veeam can't interact with MinIO's SOS API and after the change it can.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/veeam-sos-api.go
+++ b/cmd/veeam-sos-api.go
@@ -85,16 +85,6 @@ type systemInfo struct {
 		UploadSessions bool `xml:"UploadSessions"`
 		IAMSTS         bool `xml:"IAMSTS"`
 	} `mxl:"ProtocolCapabilities"`
-	APIEndpoints struct {
-		IAMEndpoint string `xml:"IAMEndpoint"`
-		STSEndpoint string `xml:"STSEndpoint"`
-	} `xml:"APIEndpoints"`
-	SystemRecommendations struct {
-		S3ConcurrentTaskLimit    int `xml:"S3ConcurrentTaskLimit"`
-		S3MultiObjectDeleteLimit int `xml:"S3MultiObjectDeleteLimit"`
-		StorageCurrentTaskLimit  int `xml:"StorageCurrentTaskLimit"`
-		KBBlockSize              int `xml:"KbBlockSize"`
-	} `xml:"SystemRecommendations"`
 }
 
 // This optional functionality allows vendors to report space information to Veeam products, and Veeam will make placement
@@ -118,13 +108,10 @@ func veeamSOSAPIGetObject(ctx context.Context, bucket, object string, rs *HTTPRa
 	switch object {
 	case systemXMLObject:
 		si := systemInfo{
-			ProtocolVersion: "1.0",
-			ModelName:       "MinIO " + ReleaseTag,
+			ProtocolVersion: "\"1.0\"",
+			ModelName:       "\"MinIO " + ReleaseTag + "\"",
 		}
 		si.ProtocolCapabilities.CapacityInfo = true
-
-		// Default recommended block size with MinIO
-		si.SystemRecommendations.KBBlockSize = 4096
 
 		buf = encodeResponse(&si)
 	case capacityXMLObject:


cmd/veeam-sos-api.go
```diff
@@ -118,14 +108,11 @@ func veeamSOSAPIGetObject(ctx context.Context, bucket, object string, rs *HTTPRa
...
 		si := systemInfo{
-			ProtocolVersion: "1.0",
-			ModelName:       "MinIO " + ReleaseTag,
+			ProtocolVersion: "\"1.0\"",
+			ModelName:       "\"MinIO " + ReleaseTag + "\"",
 		}
 		si.ProtocolCapabilities.CapacityInfo = true
 
-		// Default recommended block size with MinIO
-		si.SystemRecommendations.KBBlockSize = 4096
```
harshavardhana, line 127: You shouldn't be removing this.. this recommendation from MinIO. 

Default of 1MiB is not a good idea in general.

cmd/veeam-sos-api.go
```diff
@@ -85,16 +85,6 @@ type systemInfo struct {
 		UploadSessions bool `xml:"UploadSessions"`
 		IAMSTS         bool `xml:"IAMSTS"`
 	} `mxl:"ProtocolCapabilities"`
-	APIEndpoints struct {
```
harshavardhana, line 88: You shouldn't be removing this also, if this is optional in the sense it expects them to be empty then it needs to be done differently.

It is kept here for future purposes

harshavardhana (changes requested):
If you want I can send a cleaner change. Don't remove things without knowing the context.

harshavardhana:
#17202 replaces this PR

cmd/veeam-sos-api.go
omarkohl: @harshavardhana could you elaborate on why this is? The Veeam documentation (that is literally quoted in `cmd/veeam-sos-api.go`) states:

> <KbBlockSize>
> Veeam Block Size for backup and restore processing before compression is applied. The higher the block size, the more
> backup space is needed for incremental backups. Larger block sizes also mean less performance for random read restore
> methods like Instant Restore, File Level Recovery, and Database/Application restores. Veeam recommends that vendors
> optimize the storage system for the default value of 1MB minus compression object sizes. The setting simultaneously
> affects read from source, block, file, dedup, and object storage backup targets for a specific Veeam Job.

Furthermore:

> - Optional: Set server preferences for Backup & Replication parallel sessions, batch size of deletes, and block sizes (before
>   compression). This is an optional area; by default, there should be no <SystemRecommendations> section in the
>   system.xml. Vendors can work with Veeam Product Management and the Alliances team on getting approval to integrate
>   specific system recommendations based on current support case statistics and storage performance possibilities.

i.e. summarizing the documentation: Don't change the default value 1024 unless you are really sure and have talked to Veeam about it.

Is the choice of 4096 a conscious decision with full knowledge that this causes disadvantages to Veeam, in particular higher storage usage and worse performance (end to end when using Veeam as the S3 client) at least in some use cases?

Please note that before the introduction of the SOS API, Veeam was and is always using 1024 as the default, so essentially every Veeam instance out there that is using MinIO as a backend is using that value right now! Setting the value to 4096 would modify the default for new Veeam backup jobs!

Please also note that even though the value is called KBBlockSize, it refers to the **size of the S3 objects that Veeam writes**, not to the MinIO-internal block size which of course can't be influenced at this point, just in case that wasn't clear.

Given that the setting sets the S3 object size, it simultaneously affects read from source, block, file, dedup, and object storage backup targets for a specific Veeam job i.e. it has wide-ranging consequences.

harshavardhana: Yes it is a conscious decision this is a recommendation, users are free to choose what they like on their end.

harshavardhana: We are not going to change this decision for now.

klauspost: > Given that the setting sets the S3 object size, it simultaneously affects read from source, block, file, dedup, and object storage backup targets for a specific Veeam job i.e. it has wide-ranging consequences.

It affects dedup efficiency. That's it. The clients will have to read the same amount of data.

Reducing the amount of objects to list/write/update by close to a factor of 4x has a much bigger impact than a potential single-digit-percentage improvement in storage efficiency. Either way it is a default you can override at your own discretion if you are that confident you know better.

omarkohl: OK, thanks for the info. I am aware the value can be overridden on the client's side, but it has to be done for every backup job so having a sane default is valuable. I understand your argument that 4096 is precisely this sane default.

PR #17201, opened 2023-05-12
allow specification of path/virtual style bucket lookup in batch replication

poornas:
## Motivation and Context
for compatibility with other S3 remotes

## How to test this PR?
mc batch start with this yaml to replicate to local minio bucket `vbucket`
```
replicate:
  apiVersion: v1
  # source of the objects to be replicated
  source:
    type: s3 # valid values are "minio", "s3"
    bucket: mybucket
    prefix:   # 'PREFIX' is optional
    # NOTE: if source is remote then target must be "local"
    endpoint: https://storage.googleapis.com
    path: on
    credentials:
      accessKey: xxx
      secretKey:  yyyyy
  #  #   sessionToken: SESSION-TOKEN # Optional only available when rotating credentials are used

  # target where the objects must be replicated
  target:
    type: minio # valid values are "minio"
    bucket: vbucket
    prefix:  # 'PREFIX' is optional
    # NOTE: if target is remote then source must be "local"
    # endpoint: http://localhost:9004
    # credentials:
    #   accessKey: minio
    #   secretKey: minio123
    # #   sessionToken: SESSION-TOKEN # Optional only available when rotating credentials are used

  # NOTE: All flags are optional
  # - filtering criteria only applies for all source objects match the criteria
  # - configurable notification endpoints
  # - configurable retries for the job (each retry skips successfully previously replaced objects)
  flags:
    filter:
    #  newerThan: "8h" # match objects newer than this value (e.g. 7d10h31s)
    #   olderThan: "8h" # match objects older than this value (e.g. 7d10h31s)
    #  createdAfter: "date" # match objects created after "date"
    #  createdBefore: "date" # match objects created before "date"

    #     ## NOTE: tags are not supported when "source" is remote.
    # tags:
    #   - key: "name"
    #     value: "pick*" # match objects with tag 'name', with all values starting with 'pick'

    #     ## NOTE: metadata filter not supported when "source" is non MinIO.
    #     # metadata:
    #     #   - key: "content-type"
    #     #     value: "image/*" # match objects with 'content-type', with all values starting with 'image/'

  #   notify:
  #     endpoint: "https://notify.endpoint" # notification endpoint to receive job status events
  #     token: "Bearer xxxxx" # optional authentication token for the notification endpoint

    retry:

```

## Types of changes

## Checklist:

### DIFF
--- a/cmd/batch-handlers.go
+++ b/cmd/batch-handlers.go
@@ -86,6 +86,7 @@ import (
 //     type: "minio"
 //     bucket: "testbucket1"
 //     endpoint: "https://play.min.io"
+//     path: "on"
 //     credentials:
 //       accessKey: "minioadmin"
 //       secretKey: "minioadmin"
@@ -181,6 +182,10 @@ func (t BatchJobReplicateResourceType) Validate() error {
 	return nil
 }
 
+func (t BatchJobReplicateResourceType) isMinio() bool {
+	return t == BatchJobReplicateResourceMinIO
+}
+
 // Different types of batch jobs..
 const (
 	BatchJobReplicateResourceMinIO BatchJobReplicateResourceType = "minio"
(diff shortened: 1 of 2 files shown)


cmd/batch-handlers.go
```diff
@@ -217,19 +221,41 @@ type BatchJobReplicateTarget struct {
...
 	Prefix   string                        `yaml:"prefix" json:"prefix"`
 	Endpoint string                        `yaml:"endpoint" json:"endpoint"`
+	Path     string                        `yaml:"path" json:"path"`
 	Creds    BatchJobReplicateCredentials  `yaml:"credentials" json:"credentials"`
 }
 
+// ValidPath returns true if path is valid
+func (t BatchJobReplicateTarget) ValidPath() bool {
+	switch t.Path {
+	case "on", "off", "auto", "":
```
harshavardhana: Don't need a switch case for a simple one liner

harshavardhana (changes requested):

cmd/batch-handlers.go
harshavardhana: @poornas ^^

harshavardhana (approved):

vadmeste (approved):
LGTM

PR #17205, opened 2023-05-14
fix: tier handlers must write error only once

jiuker:
## Description

https://github.com/minio/minio/blob/a7f266c9075a8c2eee5fec39dac0b5411763d2a4/cmd/admin-handler-utils.go#L37-L61
It looks like this exception has been handled here.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/tier-handlers.go
+++ b/cmd/tier-handlers.go
@@ -141,11 +141,6 @@ func (api adminAPIHandlers) ListTierHandler(w http.ResponseWriter, r *http.Reque
 		writeErrorResponseJSON(ctx, w, errorCodes.ToAPIErr(ErrServerNotInitialized), r.URL)
 		return
 	}
-
-	if globalTierConfigMgr == nil {
-		writeErrorResponseJSON(ctx, w, errorCodes.ToAPIErr(ErrServerNotInitialized), r.URL)
-		return
-	}
 	objAPI, _ := validateAdminReq(ctx, w, r, iampolicy.ListTierAction)
 	if objAPI == nil {
 		return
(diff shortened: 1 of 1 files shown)


harshavardhana:
We need to make sure we return error for `globalTierConfigMgr` is nil.

jiuker:
@harshavardhana /cc now?

cmd/tier-handlers.go
```diff
@@ -133,6 +137,15 @@ func (api adminAPIHandlers) ListTierHandler(w http.ResponseWriter, r *http.Reque
...
 
+	if globalTierConfigMgr == nil {
+		writeErrorResponseJSON(ctx, w, errorCodes.ToAPIErr(ErrServerNotInitialized), r.URL)
+		return
+	}
+
+	if globalTierConfigMgr == nil {
+		writeErrorResponseJSON(ctx, w, errorCodes.ToAPIErr(ErrServerNotInitialized), r.URL)
+		return
+	}
```
harshavardhana: Looks like copy paste error?

harshavardhana (changes requested):

harshavardhana (approved):

cmd/tier-handlers.go
```diff
@@ -202,11 +211,14 @@ func (api adminAPIHandlers) RemoveTierHandler(w http.ResponseWriter, r *http.Req
 
 	defer logger.AuditLog(ctx, w, r, mustGetClaimsFromToken(r))
 
-	objAPI, _ := validateAdminReq(ctx, w, r, iampolicy.SetTierAction)
-	if objAPI == nil || globalNotificationSys == nil || globalTierConfigMgr == nil {
+	if globalTierConfigMgr == nil {
```
krisis: `globalTierConfigMgr` is [initialized](https://github.com/minio/minio/blob/master/cmd/server-main.go#L366) to a non-nil value before the object layer is [initialized](https://github.com/minio/minio/blob/master/cmd/server-main.go#L416). So, we don't need to check `if globalTierConfigMgr == nil` in any of the tier handlers. I realized this only on reviewing this PR.

krisis (commented):
Good catch!

krisis (approved):

harshavardhana (approved):

PR #17206, opened 2023-05-14
fix: reduce using memory and temporary files.

harshavardhana:
## Description
fix: reduce using memory and temporary files.

## Motivation and Context
This PR changes Post upload API into a complete
streaming implementation. This is the first step
before we attempt fan-out implementation.

## How to test this PR?
use `mc share upload alias/bucket/large-file` without
this change MinIO since it uses Go's APIs which seems
to use `/tmp` folder to store content.

This PR is an attempt to do this in a fully 
streaming manner.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/bucket-handlers.go
+++ b/cmd/bucket-handlers.go
@@ -22,8 +22,10 @@ import (
 	"context"
 	"encoding/base64"
 	"encoding/xml"
+	"errors"
 	"fmt"
 	"io"
+	"mime/multipart"
 	"net/http"
 	"net/textproto"
 	"net/url"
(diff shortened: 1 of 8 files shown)


cmd/bucket-handlers.go
```diff
@@ -912,41 +914,111 @@ func (api objectAPIHandlers) PostPolicyBucketHandler(w http.ResponseWriter, r *h
...
+			writeErrorResponse(ctx, w, apiErr, r.URL)
+			return
+		}
+		maxParts--
 
-	// Extract all form fields
-	fileBody, fileName, fileSize, formValues, err := extractPostPolicyFormValues(ctx, form)
-	if err != nil {
+		name := part.FormName()
+		if name == "" {
```
vadmeste, line 950: is part.Close() needed here ?

vadmeste (commented):
One comment

cmd/bucket-handlers.go
harshavardhana: well it is important

```go
func (p *Part) Close() error {
	io.Copy(io.Discard, p)
	return nil
}
```

Apparently to drain this

vadmeste (approved):
LGTM

krisis (approved):

PR #17208, opened 2023-05-15
fix: workers.New err must be returned

jiuker:
## Description

When workers.New get err should return

## Types of changes

## Checklist:

### DIFF
--- a/cmd/bucket-lifecycle.go
+++ b/cmd/bucket-lifecycle.go
@@ -163,7 +163,7 @@ func initBackgroundExpiry(ctx context.Context, objectAPI ObjectLayer) {
 	nwk, err := workers.New(workerSize)
 	if err != nil {
 		logger.LogIf(ctx, err)
-		return
+		nwk = 4
 	}
 
 	go func() {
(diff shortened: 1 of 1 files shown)


vadmeste (approved):
LGTM

harshavardhana (approved):

cmd/bucket-lifecycle.go
```diff
@@ -157,11 +157,13 @@ func initBackgroundExpiry(ctx context.Context, objectAPI ObjectLayer) {
 	ewk, err := workers.New(workerSize)
 	if err != nil {
 		logger.LogIf(ctx, err)
+		return
```
harshavardhana: actually this PR is wrong we should always initialize ILM and never return errors, initialize this with some random valid value.

```suggestion
		ewk = 4
```

cmd/bucket-lifecycle.go
```diff
@@ -157,11 +157,13 @@ func initBackgroundExpiry(ctx context.Context, objectAPI ObjectLayer) {
...
 	}
 
 	nwk, err := workers.New(workerSize)
 	if err != nil {
 		logger.LogIf(ctx, err)
+		return
```
harshavardhana:
```suggestion
		nwk = 4
```

harshavardhana (changes requested):

jiuker:
@harshavardhana it seems that is wrong which you suggested.Do you mean workerSize should be 4 by default?

harshavardhana:
> @harshavardhana it seems that is wrong which you suggested.Do you mean workerSize should be 4 by default?

Correct if someone sets the wrong value @jiuker - we should never fail to initialize, we should always initialize with sane defaults.

harshavardhana (approved):

klauspost (approved):

PR #17226, opened 2023-05-17
Protecting the read index is not out of bounds

jiuker:
## Description

Protecting the read index is not out of bounds
We can see that erasureDisks can be written by other write locks, and if the index we read for the first time is already at risk of going out of bounds after losing the lock, it is necessary to organize this situation.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/erasure-sets.go
+++ b/cmd/erasure-sets.go
@@ -221,16 +221,14 @@ func (s *erasureSets) connectDisks() {
 			if disk.IsLocal() && disk.Healing() != nil {
 				globalBackgroundHealState.pushHealLocalDisks(disk.Endpoint())
 			}
-			s.erasureDisksMu.RLock()
+			s.erasureDisksMu.Lock()
 			setIndex, diskIndex, err := findDiskIndex(s.format, format)
-			s.erasureDisksMu.RUnlock()
 			if err != nil {
 				printEndpointError(endpoint, err, false)
 				disk.Close()
 				return
 			}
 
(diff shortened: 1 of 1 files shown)


klauspost (approved):
Good find!

harshavardhana:
@jiuker we never rewrite erasureDisks() to go out of bounds, i.e drives don't get reduced automatically in a given deployment.

cmd/erasure-sets.go
```diff
@@ -221,16 +221,14 @@ func (s *erasureSets) connectDisks() {
 			if disk.IsLocal() && disk.Healing() != nil {
 				globalBackgroundHealState.pushHealLocalDisks(disk.Endpoint())
 			}
-			s.erasureDisksMu.RLock()
+			s.erasureDisksMu.Lock()
 			setIndex, diskIndex, err := findDiskIndex(s.format, format)
-			s.erasureDisksMu.RUnlock()
 			if err != nil {
 				printEndpointError(endpoint, err, false)
 				disk.Close()
```
harshavardhana, line 228: You are missing an unlock here causing deadlock @jiuker

harshavardhana (changes requested):

cmd/erasure-sets.go
jiuker: @harshavardhana No.I just move the '223L Lock' to here.Close it if it on purpose.

harshavardhana (approved):

PR #17228, opened 2023-05-17
Check for only network errors in audit webhook for reachability

Praveenrajmani:
## Description

Currently, If the webhook responds with 500 response code, we consider the audit target to be down and ping repeateadly to initialize the target. Instead, we need to just consider actual target reachability error. Just receiving a response from the webhook is a good enough check for reachability.

This commit also addresses the review comments from https://github.com/minio/minio/pull/17121

## Motivation and Context

non network errors should not be considered for target reachability checks  

## How to test this PR?

- Start a webhook server which responds 500 response code for empty payload in the request.
```go
package main

import (
	"fmt"
	"io/ioutil"
	"log"
	"net/http"
)

func main() {

	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
		b, err := ioutil.ReadAll(r.Body)
		defer r.Body.Close()
		if err != nil {
			log.Fatal(err)
			return
		}
		fmt.Println(string(b))
		if string(b) == "{}" {
			fmt.Println("empty payload! Responding back with 500 response code..")
			w.WriteHeader(500)
		} else {
			w.WriteHeader(200)
		}
	})

	log.Printf("listening on http://%s/", "localhost:8080")
	log.Fatal(http.ListenAndServe("localhost:8080", nil))
}

```
- Configure audit_webhook 
```sh
mc admin config set myminio audit_webhook enable=on endpoint=http://localhost:8080/
```
- Restart MinIO server
- You will see repeated pings (every second) and the audit logs will not be sent properly as the audit target is not yet initialized.
- With this PR, you will not see repeated pings and you will see audit logs properly sent.

## Types of changes

## Checklist:

### DIFF
--- a/internal/logger/config.go
+++ b/internal/logger/config.go
@@ -253,6 +253,13 @@ func NewConfig() Config {
 	return cfg
 }
 
+func getCfgVal(envName, key, defaultValue string) string {
+	if key != config.Default {
+		envName = envName + config.Default + key
+	}
+	return env.Get(envName, defaultValue)
+}
+
 func lookupLegacyConfigForSubSys(subSys string) Config {
 	cfg := NewConfig()
 	switch subSys {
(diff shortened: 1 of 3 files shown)


shtripat (approved):
LGTM. Verified and works as expected.

harshavardhana (approved):

anjalshireesh (approved):
LGTM

krisis (approved):

PR #17233, opened 2023-05-17
support fan-out objects via PostUpload()

harshavardhana:
## Description
support fan-out objects via PostUpload()

## Motivation and Context
This PR adds support for `x-minio-fanout-list` that
provides the list of object names that need to be created 
on the namespace from a single input incoming stream.

## How to test this PR?
Requires https://github.com/minio/minio-go/pull/1826

## Types of changes

## Checklist:


cmd/bucket-handlers.go
```diff
@@ -1174,6 +1214,84 @@ func (api objectAPIHandlers) PostPolicyBucketHandler(w http.ResponseWriter, r *h
...
+
+				fanOutResp = append(fanOutResp, minio.PutObjectFanOutResponse{
+					Key:          objInfo.Name,
+					ETag:         objInfo.ETag,
+					VersionID:    objInfo.VersionID,
+					LastModified: &objInfo.ModTime,
+				})
+
+				// Notify object created event.
+				defer sendEvent(eventArgs{
```
klauspost: This will potentially defer thousands of calls.

Since we are blocking on them anyway, we should probably just remove the defer - it will also allow senders to catch up if we are doing multiple loops.

cmd/postpolicyform.go
```diff
@@ -262,7 +262,7 @@ func checkPolicyCond(op string, input1, input2 string) bool {
 // (http://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-HTTPPOSTConstructPolicy.html)
 func checkPostPolicy(formValues http.Header, postPolicyForm PostPolicyForm) error {
 	// Check if policy document expiry date is still not reached
-	if !postPolicyForm.Expiration.After(UTCNow()) {
+	if !postPolicyForm.Expiration.UTC().After(UTCNow()) {
```
klauspost: No need. "After" handles timezones fine.
```suggestion
	if !postPolicyForm.Expiration.After(UTCNow()) {
```

cmd/bucket-handlers.go
```diff
@@ -971,9 +970,30 @@ func (api objectAPIHandlers) PostPolicyBucketHandler(w http.ResponseWriter, r *h
 
 		var b bytes.Buffer
 		if fileName == "" {
+			if http.CanonicalHeaderKey(name) == http.CanonicalHeaderKey("x-minio-fanout-list") {
+				dec := json.NewDecoder(part)
+				dec.DisallowUnknownFields()
```
klauspost: This is a double edged sword. I tend to agree that we want older versions to complain about unknown fields.

cmd/post-policy-fan-out.go
```diff
@@ -0,0 +1,109 @@
...
+			err = errors.Join(err, e)
+		}
+	}
+	return err
+}
+
+// fanOutPutObject takes an input source reader and fans out multiple PUT operations
+// based on the incoming fan-out request, a context cancelation by the caller
+// would ensure all fan-out operations are canceled.
+func fanOutPutObject(ctx context.Context, bucket string, objectAPI ObjectLayer, fanOutReq []minio.PutObjectFanOutRequest, pReader *PutObjReader) ([]ObjectInfo, []error) {
```
klauspost: Sending the content as bytes makes it much easier and removes the need for multiple wrapper layers. Let me do the entire func

klauspost: This eliminates all the reader, closer, writer and does everything within goroutines:

```Go
// fanOutPutObject takes an input source reader and fans out multiple PUT operations
// based on the incoming fan-out request, a context cancelation by the caller
// would ensure all fan-out operations are canceled.
func fanOutPutObject(ctx context.Context, bucket string, objectAPI ObjectLayer, fanOutReq []minio.PutObjectFanOutRequest, content []byte) ([]ObjectInfo, []error) {
	errs := make([]error, len(fanOutReq))
	objInfos := make([]ObjectInfo, len(fanOutReq))

	var wg sync.WaitGroup
	wg.Add(len(fanOutReq))
	for i, req := range fanOutReq {
		go func(idx int, req minio.PutObjectFanOutRequest) {
			defer wg.Done()
			hr, _ := hash.NewReader(bytes.NewReader(content), int64(len(content)), "", "", -1)
			reader := NewPutObjReader(hr)
			defer func() {
				errs[idx] = errors.Join(errs[idx], reader.Close(), hr.Close())
				if errs[idx] != nil {
					objInfos[idx] = ObjectInfo{Name: req.Key}
				}
			}()
			userDefined := make(map[string]string, len(req.UserMetadata))
			for k, v := range req.UserMetadata {
				userDefined[k] = v
			}
			userDefined[xhttp.AmzObjectTagging] = s3utils.TagEncode(req.UserTags)

			objInfos[idx], errs[idx] = objectAPI.PutObject(ctx, bucket, req.Key, reader, ObjectOptions{
				Versioned:        globalBucketVersioningSys.PrefixEnabled(bucket, req.Key),
				VersionSuspended: globalBucketVersioningSys.PrefixSuspended(bucket, req.Key),
				UserDefined:      userDefined,
			})
		}(i, req)
	}
	wg.Wait()

	return objInfos, errs
}
```

klauspost (commented):
Are POST uploads limited to 2MB? Or am misreading `maxMemoryBytes`?

harshavardhana:
> Are POST uploads limited to 2MB? Or am misreading `maxMemoryBytes`?

That is only for field values - not the File body @klauspost

cmd/post-policy-fan-out.go
harshavardhana: Well, there is a reason we need to do that because of the encryption requirement @klauspost that's why I can't pass down bytes as plain-text. 

If SSE headers are set we need a stream that is precisely the encrypted stream.

cmd/bucket-handlers.go
harshavardhana: Yeah this is mostly defensive, can be removed.

cmd/bucket-handlers.go
harshavardhana: I think I will perhaps do this as a bulk call at once instead of many events.

cmd/postpolicyform.go
harshavardhana: I think I did this to fix a bug that I found with Expiration when it is TZ specific.

cmd/post-policy-fan-out.go
klauspost: @harshavardhana Before calling you can read the stream. You are not mutating any data per upload anyway.

If data should have unique keys we need to move the encryption part down here, so targets can have unique keys. Either way sending data as bytes will allow writes to complete independent of each other and not be tied to a single io.Copy loop.

cmd/postpolicyform.go
klauspost: The conversion will not do anything, since the absolute time will be the same.

cmd/bucket-handlers.go
klauspost: Yeah, probably a better approach. Maybe spawn a goroutine with a waitgroup, so it will send while the subsequent loops are running (if multiple), and we only wait once all have completed.

cmd/bucket-handlers.go
klauspost: I think it is reasonable to keep it. It is actually a good thing.

How about if we do something like this:

```
type PutObjectFanOutRequest struct {
	Key                string            `json:"key"`
	...
	Optional       map[string]string `json:"optional,omitempty"`
}
```

This will allow us to add optional values, that can safely be ignored, and we reject all other unknown fields?

harshavardhana: Its okay let me remove it.

cmd/bucket-handlers.go
```diff
@@ -1174,6 +1194,99 @@ func (api objectAPIHandlers) PostPolicyBucketHandler(w http.ResponseWriter, r *h
...
 
+	if len(fanOutReq) > 0 {
+		// Fan-out requires no copying, and must be carried from original source
+		// https://en.wikipedia.org/wiki/Copy_protection so the incoming stream
+		// is always going to be in-memory as we cannot re-read from what we
+		// wrote to disk - since that amounts to "copying" from a "copy"
+		// instead of "copying" from source, we need the stream to be seekable
+		// to ensure that we can make fan-out calls concurrently.
+		buf := bytebufferpool.Get()
+		n, err := io.Copy(buf, pReader)
```
klauspost:
```suggestion
		// Maximum allowed fan-out object size.
		const maxFanOutSize = 16 << 20

		buf := bytebufferpool.Get()
		n, err := io.Copy(buf, ioutil.HardLimitReader(pReader, maxFanOutSize))
```

klauspost (commented):
Final Q: Should Fanout have separate permissions? It seems like a feature you want to control to some extent?

cmd/bucket-handlers.go
harshavardhana: do we really want to limit here?

harshavardhana:
> Final Q: Should Fanout have separate permissions? It seems like a feature you want to control to some extent?

We could add s3:PutObjectFanOut as an action here but I suspect no one really would want this anyways without having a specific requirement. So complicating it to have a custom policy is not worth the effort. 

Just a simple "readwrite" should get them going.

klauspost:
> We could add s3:PutObjectFanOut as an action here but I suspect no one really would want this anyways without having a specific requirement. 

Yeah, that is the issue. Some would like to *disable* this so it cannot be abused for DOS on publicly available systems.

I think a `s3:PostObjectFanOut` as an extended options is the most reasonable.

harshavardhana:
Okay let's add a proper action for this to be explicit @klauspost https://github.com/minio/pkg/pull/65

cmd/bucket-handlers.go
klauspost: @harshavardhana Yes, otherwise users can just fill up memory with data.

harshavardhana:
This PR is fully complete end to end along with encryption settings etc. fixed a bug as well when we had bucket level encryption set - objects were not getting encrypted via PostUpload().

harshavardhana:
PTAL @krisis @poornas

krisis (approved):

poornas (approved):

Issue #17212, opened 2023-05-15
missing http status error in discovery doc parsing function


deluxghost:
https://github.com/minio/minio/blob/684399433b8bd7e164a4b81e57a4bb0d76d0b428/internal/config/identity/openid/jwt.go#L281-L283

value of `err` under this condition is `nil`, the return values will be an empty `DiscoveryDoc` and a `nil` error. This will cause outer function to skip the err check

I keep seeing this error `Error: Unable to initialize OpenID: no JWKS URI found in your provider's discovery doc` at minio start, but the url is fine. Turns out one of the cause is my WAF blocking any non-human request with 403 code, but minio's log doesn't mention the status code at all. Thus it's harder to find the root cause of the error

## Expected Behavior

output error message about the HTTP code and return outer function, instead of showing `jwks_uri` error

## Current Behavior
the outer error check is skipped, and minio shows `jwks_uri` error due to the empty `DiscoveryDoc` struct

## Possible Solution
return correct new `error` or `Errorf` related to the HTTP code. And as @harshavardhana said in #17178, we need a better solution than simple `errors.New(resp.Status)`

## Steps to Reproduce (for bugs)

1. set the openid config url, make it response non-OK code
2. restart minio, it shows `Error: Unable to initialize OpenID: no JWKS URI found in your provider's discovery doc` error, and openid doesn't work

## Your Environment

* Version used (`minio --version`):
* Server setup and configuration:
* Operating System and version (`uname -a`):

harshavardhana:
Thanks @deluxghost

harshavardhana:
fixed via #17242

PR #17242, opened 2023-05-19
fix: return proper error if OIDC Discoverydoc fails to respond

jiuker:
## Description

return err with status

## Types of changes

## Checklist:
- [x] Fixes a regression [issue17212](https://github.com/minio/minio/issues/17212)

### DIFF
--- a/docs/sts/web-identity.go
+++ b/docs/sts/web-identity.go
@@ -90,7 +90,7 @@ func parseDiscoveryDoc(ustr string) (DiscoveryDoc, error) {
 	}
 	defer resp.Body.Close()
 	if resp.StatusCode != http.StatusOK {
-		return d, errors.New(resp.Status)
+		return d, fmt.Errorf("unexpected error returned by %s : status(%s)", ustr, resp.Status)
 	}
 	dec := json.NewDecoder(resp.Body)
 	if err = dec.Decode(&d); err != nil {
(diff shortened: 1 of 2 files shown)


docs/sts/web-identity.go
```diff
@@ -90,7 +90,7 @@ func parseDiscoveryDoc(ustr string) (DiscoveryDoc, error) {
 	}
 	defer resp.Body.Close()
 	if resp.StatusCode != http.StatusOK {
-		return d, err
+		return d, errors.New(resp.Status)
```
shtripat: Suggestion, this way actual error is lost. Should we create final error with response and error both before return?

jiuker: here err always nil @shtripat

internal/config/identity/openid/jwt.go
```diff
@@ -279,7 +279,7 @@ func parseDiscoveryDoc(u *xnet.URL, transport http.RoundTripper, closeRespFn fun
 	}
 	defer closeRespFn(resp.Body)
 	if resp.StatusCode != http.StatusOK {
-		return d, err
+		return d, errors.New(resp.Status)
```
harshavardhana: We should provide more context here by saying some words

```suggestion
		return d, fmt.Errorf("unexpected error returned by %s : status(%s)", u, resp.Status)  
```

harshavardhana (changes requested):

harshavardhana (approved):

shtripat (approved):
LGTM

Issue #17147, opened 2023-05-05
Not able to add webhook notification endpoint when another is offline


SBylemans:
## Expected Behavior
I should be able to add a new endpoint, regardless whether other endpoints are offline.

## Current Behavior
Trying to add via UI gives:
<img width="469" alt="image" src="https://user-images.githubusercontent.com/5698858/236415636-6a2ed6cc-a17b-4292-b92d-c549f4d60d60.png">

And via minio client:
```
mc: <ERROR> Unable to set '<url>' to server: error (<endpoint-name>:webhook): not connected to target server/service.
```
where `endpoint-name` is different from the endpoint I'm trying to configure.

## Steps to Reproduce (for bugs)
1. Add webhook endpoint that is listening
2.Stop the listening server
3.Try to add another webhook endpoint, it fails

## Context
I'm trying to publish events to a NiFi cluster. I added a notification endpoint for every node, but I only want the primary node to process it. Seeing as the primary node of this cluster can change, I can't just add the endpoint for the primary node and be done with it. If the primary node changes, I don't want to reconfigure it. Only the primary node is online according to minio.

Another issue with this, is that I can't configure an endpoint that is offline. What does minio care if the endpoint is online or not? It knows whether the endpoint is online or not, so why not only send to online endpoints? It is not the problem of minio that the endpoint is online or not, you can warn the user. But if the user wants to preconfigure, it is not able to do so.

SBylemans:
The issue is even worse, I'm not even able to delete a notification endpoint when one is offline...

vadmeste:
@SBylemans what version of MinIO is this ?

SBylemans:
@vadmeste RELEASE.2022-12-02T23-48-47Z

harshavardhana:
> @vadmeste RELEASE.2022-12-02T23-48-47Z

@vadmeste please upgrade your setup

SBylemans:
@harshavardhana We upgraded and now we can't see the notification endpoints in the GUI (we can see them using minio client). And we can't reset them because they are offline, I also can't add one because other ones are offline...

Praveenrajmani:
You can just disable a notification target @SBylemans and can't remove it.

You can disable the existing target by `mc admin config set <alias> <targetname> enable=off`

ldeschuyter:
Upgrade to 2023-05-04T21:44:30Z didn't resolve the issue. Now the "event destinations" are not longer visible.

They are still visible using the mc-client

[...]

PR #17250, opened 2023-05-20
fix: ignore existing target status when adding new targets

jiuker:
## Description

ignore exsits target status when add/set/delete

## Types of changes
- [x] Bug fix [#17147 ](https://github.com/minio/minio/issues/17147)

## Checklist:


shtripat:
Tested the changes. It works if old web-hook if offline. But allows setting a new one which is down as well. I understand that's acceptable as its a configuration and runtime while sending the audit, it would fail if target not reachable.
With this assumption, change looks good to me.

shtripat (approved):
lgtm

harshavardhana:
This actually reminds me that we have to the client CLI UI slightly differently here, because we have always expected targets to be online when they are being set.

Basically, we want a solution like this 

- Allow adding new targets while having some existing offline targets
- Allow deletes to happen on any target
- However do not allow offline targets to be added freshly, we want
  them to be online when they are being added - just doesn't make
  sense to add a config that is not valid. 

In this PR the 3rd part is not complete yet.

jiuker:
@harshavardhana review again

jiuker:
@shtripat make a test ,plz

shtripat:
> @shtripat make a test ,plz

How is third point from @harshavardhana being taken care? I am still able to add a target which is offline as below
```
$ mc admin config set ALIAS notify_webhook:3 enable=on endpoint=http://localhost:8000/webhook
Successfully applied new settings.
Please restart your server 'mc admin service restart myminio'.
```
Here `http://localhost:8000/webhook` is not actually up.

Am I missing something?

jiuker:
> > @shtripat make a test ,plz
> 
> How is third point from @harshavardhana being taken care? I am still able to add a target which is offline as below
> 
> ```
> $ mc admin config set ALIAS notify_webhook:3 enable=on endpoint=http://localhost:8000/webhook
> Successfully applied new settings.
> Please restart your server 'mc admin service restart myminio'.
> ```
> 
> Here `http://localhost:8000/webhook` is not actually up.
> 
> Am I missing something?

No，I don't use mc.Thx，I will test again.

jiuker:
@shtripat It's ready.Make a test.Plz

shtripat:
> @shtripat It's ready.Make a test.Plz

There seems to different issue now as unable to set `notify_webhook` value

```
$ mc admin config set myminio notify_webhook enable=on endpoint=http://localhost:8000/webhook
Successfully applied new settings.
Please restart your server 'mc admin service restart myminio'.

$ mc admin service restart myminio 
Restart command successfully sent to `myminio`. Type Ctrl-C to quit or wait to follow the status of the restart process.
...
Restarted `myminio` successfully in 1 seconds

$ mc admin config get myminio audit_webhook
audit_webhook enable=off endpoint= auth_token= client_cert= client_key= queue_size=100000 queue_dir= 
```

Webhook is actually up and able to receive requests

```
$ curl -XPOST -H 'Content-Type: application/json' -d '{"msg": "hello"}' http://localhost:8000/webhook
Webhook received!
```

jiuker:
> > @shtripat It's ready.Make a test.Plz
> 
> There seems to different issue now as unable to set `notify_webhook` value
> 
> ```
> $ mc admin config set myminio notify_webhook enable=on endpoint=http://localhost:8000/webhook
> Successfully applied new settings.
> Please restart your server 'mc admin service restart myminio'.
> 
> $ mc admin service restart myminio 
> Restart command successfully sent to `myminio`. Type Ctrl-C to quit or wait to follow the status of the restart process.
> ...
> Restarted `myminio` successfully in 1 seconds
> 
> $ mc admin config get myminio audit_webhook
> audit_webhook enable=off endpoint= auth_token= client_cert= client_key= queue_size=100000 queue_dir= 
> ```
> 
> Webhook is actually up and able to receive requests
> 
> ```
> $ curl -XPOST -H 'Content-Type: application/json' -d '{"msg": "hello"}' http://localhost:8000/webhook
> Webhook received!
> ```

@shtripat you enable audit_webhook to off? this is part of test?

shtripat:
Oh that seems to be spell mistake from my side. Will re-verify..

shtripat:
> Oh that seems to be spell mistake from my side. Will re-verify..

Sorry for type. It works now and doesn't allow to set one with down state. Also if one of targets is down, it does allow to set one which is up and running.

jiuker:
> > Oh that seems to be spell mistake from my side. Will re-verify..
> 
> Sorry for type. It works now and doesn't allow to set one with down state. Also if one of targets is down, it does allow to set one which is up and running.

@shtripat thanks a lot

harshavardhana:
PTAL @shtripat @Praveenrajmani

Praveenrajmani:
can you fix the lint failure? @jiuker

shtripat (approved):
Other than lint issues, code looks good to me.
Please fix them.

jiuker:
@Praveenrajmani run pipeline agagin.plz

shtripat (approved):
lgtm

harshavardhana (approved):

cmd/admin-handlers-config-kv.go
```diff
@@ -211,7 +211,12 @@ func setConfigKV(ctx context.Context, objectAPI ObjectLayer, kvBytes []byte) (re
 		return
 	}
 
-	if verr := validateConfig(result.Cfg, result.SubSys); verr != nil {
+	tgts, err := config.ParseConfigTargetID(bytes.NewReader(kvBytes))
```
Praveenrajmani, line 214:
```go
result.SubSys, _, target, err = config.GetSubSys(string(kvBytes))
```

^^ already provides the target which is going to be edited/added. We can use this instead.

internal/config/config.go
```diff
@@ -426,6 +426,34 @@ func (c Config) DelFrom(r io.Reader) error {
...
 }
 
+// ContextKeyString is type(string) for contextKey
+type ContextKeyString string
+
+// ContextKeyForTargetFromConfig - key for context for target from config
+const ContextKeyForTargetFromConfig = ContextKeyString("ContextKeyForTargetFromConfig")
+
+// ParseConfigTargetID - read all targetIDs from reader
+func ParseConfigTargetID(r io.Reader) (ids map[string]bool, err error) {
```
Praveenrajmani, line 436: I don't think SetConfigKV API will accept multiple targets. It works with only single target. We don't need this parser.

internal/config/notify/parse.go
```diff
@@ -58,13 +58,24 @@ func TestSubSysNotificationTargets(ctx context.Context, cfg config.Config, subSy
...
+			if err == nil && !yes {
+				err = ErrTargetsOffline
+			}
+			if err != nil {
+				return fmt.Errorf("error (%s): %w", target.ID(), err)
+			}
+		} else {
+			// Just for call init.
+			// Ignore target is online or offline
+			_, _ = target.IsActive()
```
Praveenrajmani, line 78: IMO, We don't need to ping other targets too. This is a side effect and can be avoided. 

`Set|Delete` don't need to care about Init.

Need your help here to decide @harshavardhana

Praveenrajmani (commented):
Thanks for the PR @jiuker 

Just some minor things to check.

cmd/admin-handlers-config-kv.go
jiuker: @Praveenrajmani It used in ParseConfigTargetID

internal/config/config.go
jiuker: @harshavardhana help there too.

PR #17251, opened 2023-05-20
fix: for Target isActive use net.Dial instead

jiuker:
## Description

Target isActive should use StatusCode==200

## Types of changes

## Checklist:


jiuker:
This BUG is found when I reproduct #17147

shtripat (approved):
LGTM. Tested and verified

jiuker:
@shtripat I changed the order to prevent the body of the resp from not being closed.Review again plz

shtripat (approved):

internal/event/target/webhook.go
```diff
@@ -155,7 +155,10 @@ func (target *WebhookTarget) isActive() (bool, error) {
 	}
 	io.Copy(io.Discard, resp.Body)
 	resp.Body.Close()
-	// No network failure i.e response from the target means its up
+	if resp.StatusCode != http.StatusOK {
```
klauspost: Maybe:
```suggestion
	if resp.StatusCode >= http.StatusBadRequest {
```

Only return errors when >= 400

klauspost (approved):
LGTM

internal/event/target/webhook.go
```diff
@@ -155,7 +155,10 @@ func (target *WebhookTarget) isActive() (bool, error) {
 	}
 	io.Copy(io.Discard, resp.Body)
 	resp.Body.Close()
-	// No network failure i.e response from the target means its up
+	if resp.StatusCode >= http.StatusBadRequest {
```
Praveenrajmani: this change will treat 500s (internal server errors) and 403s (access denied) as connection errors. 

`isActive` function is used to check if the target is reachable or not. So, internal server and access denied errors would still mean that the target is reachable and is failing due to some parsing error or access denials.

IMO, we shouldn't treat 403s and 500s as n/w or connection errors.

jiuker: @Praveenrajmani The fact that the target is online means that we can send it a message, in fact, if it is 500 or 403, can we send it?

Praveenrajmani: If the target is reachable, we are good to send a message to that target. If MinIO receives a 500 or 403 as a response - it has to be fixed by the webhook. MinIO will just log that error.

MinIO can't treat 500s or 403s as connection errors.

If you need MinIO to persist and retry such events. you can configure `queue_dir`.

klauspost: @Praveenrajmani I see your point, but that will make proxy errors uncatchable - I see an argument both ways.

Current behavior ignores response code. Keeping it that way is the safe option, but doesn't catch proxy errors or consistent 400/403/500/503, etc codes.

klauspost: Maybe we should make @harshavardhana tie-break. If we keep the current, just close the PR.

klauspost:
@jiuker I asked Harsha to make the decision so no need to close yet.

jiuker:
> asked Harsha to make the decision so no need to close yet.

Ok.

internal/event/target/webhook.go
harshavardhana: I think the argument is kind of valid, we actually don't need to make Head request here, we should simply perform

net.Dial("host:port")

and if we got a successful conn, we conclude that the target is up. Because its entirely plausible that webhook handler doesn't implement HEAD at all which would be constantly logged now as error.

My suggestion is remove http call instead let's just make a TCP connection if that succeeds we can conclude that the target is online.

harshavardhana (changes requested):

Praveenrajmani (approved):

harshavardhana:
PTAL @shtripat

shtripat (approved):
lgtm

klauspost (approved):

jiuker:
@klauspost pipeline failed.It related to this pr?

klauspost:
I don't really see how webhooks could affect this. I'll rerun it.

PR #17252, opened 2023-05-20
offline drives more than 50% of total drives return error

harshavardhana:
## Description
with more offline drives than 50% of drives return error

## Motivation and Context
avoids reading the incoming body by returning error
quickly.

## How to test this PR?
Start with a 4-node setup, you should see the
stream is not even read instead the error
is returned right away.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/erasure-multipart.go
+++ b/cmd/erasure-multipart.go
@@ -39,6 +39,7 @@ import (
 	"github.com/minio/minio/internal/logger"
 	"github.com/minio/pkg/mimedb"
 	"github.com/minio/pkg/sync/errgroup"
+	uatomic "go.uber.org/atomic"
 )
 
 func (er erasureObjects) getUploadIDDir(bucket, object, uploadID string) string {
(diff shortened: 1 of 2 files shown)


cmd/erasure-multipart.go
```diff
@@ -371,29 +372,61 @@ func (er erasureObjects) newMultipartUpload(ctx context.Context, bucket string,
...
 	}
 
+	// If we have offline disks upgrade the number of erasure codes for this object.
 	parityOrig := parityDrives
+
+	atomicParityDrives := uatomic.NewInt64(0)
+	atomicOfflineDrives := uatomic.NewInt64(0)
+
+	// Start with current parityDrives
+	atomicParityDrives.Store(int64(parityDrives))
```
klauspost, line 389: We start with "parityDrives" from storageclass, but when a disk is nil or offline, we *increment* this.

I don't understand the math going on here.

cmd/erasure-object.go
```diff
@@ -1029,30 +1035,42 @@ func (er erasureObjects) putObject(ctx context.Context, bucket string, object st
...
 				di, err := disk.DiskInfo(ctx)
 				if err != nil || di.ID == "" {
+					atomicOfflineDrives.Inc()
 					atomicParityDrives.Inc()
 				}
 			}(disk)
 		}
 		wg.Wait()
 
+		if int(atomicOfflineDrives.Load()) > len(storageDisks)/2 {
```
klauspost: Wouldn't this catch more cases?
```suggestion
		if int(atomicOfflineDrives.Load()) >= len(storageDisks+1)/2 {
```

harshavardhana: It can actually, let me add this suggestion.

cmd/erasure-multipart.go
harshavardhana: This PR is not about any math going wrong - it is simply a faster way to return error instead of trying to use a large parity and reading from body.

cmd/erasure-object.go
```diff
@@ -1029,30 +1035,42 @@ func (er erasureObjects) putObject(ctx context.Context, bucket string, object st
...
 				di, err := disk.DiskInfo(ctx)
 				if err != nil || di.ID == "" {
+					atomicOfflineDrives.Inc()
 					atomicParityDrives.Inc()
 				}
 			}(disk)
 		}
 		wg.Wait()
 
+		if int(atomicOfflineDrives.Load()) >= len(storageDisks+1)/2 {
```
klauspost: typo:
```suggestion
		if int(atomicOfflineDrives.Load()) >= (len(storageDisks)+1)/2 {
```

vadmeste (approved):
LGTM

klauspost (approved):

PR #17259, opened 2023-05-22
Fixed When the MINIO_BROWSER_REDIRECT_URL parameter contains path, the console cannot be accessed normally.

polaris-phecda:
## Description
When the MINIO_BROWSER_REDIRECT_URL parameter contains path, the console cannot be accessed normally.
Related to [https://github.com/minio/console/pull/2725](https://github.com/minio/console/pull/2725) 

## Motivation and Context
When the minio server start, the `basepath ` variable needs to be injected into the swagger spec of the console.

## How to test this PR?
Refer to this PR:
[https://github.com/minio/console/pull/2725](https://github.com/minio/console/pull/2725)

## Types of changes

## Checklist:

### DIFF
--- a/cmd/common-main.go
+++ b/cmd/common-main.go
@@ -260,6 +260,10 @@ func initConsoleServer() (*restapi.Server, error) {
 		return nil, err
 	}
 
+	subPath := restapi.GetSubPath()
+	swaggerSpec.Spec().BasePath = subPath + swaggerSpec.Spec().BasePath
+	swaggerSpec.OrigSpec().BasePath = subPath + swaggerSpec.OrigSpec().BasePath
+
 	api := operations.NewConsoleAPI(swaggerSpec)
 
 	if !serverDebugLog {


klauspost:
We shouldn't magically work around a wrong configuration. The value shows an intent, which may be wrong, but will lead to unpredictable behavior.

Printing a warning would be fine IMO.

klauspost:
cc @harshavardhana

polaris-phecda:
> We shouldn't magically work around a wrong configuration. The value shows an intent, which may be wrong, but will lead to unpredictable behavior.
> 
> Printing a warning would be fine IMO.

This modification is not intended to bypass the incorrect configuration. The MINIO_BROWSER_REDIRECT_URL variable is configured but does not take effect. This is because the correct basepath is missing in the sepc. The correct basepath needs to be injected during startup.
issue at minio/console: [https://github.com/minio/console/issues/2774](https://github.com/minio/console/issues/2774)

cmd/common-main.go
```diff
@@ -260,6 +260,10 @@ func initConsoleServer() (*restapi.Server, error) {
 		return nil, err
 	}
 
+	subPath := restapi.GetSubPath()
```
harshavardhana, line 263: Is this safe to change base paths in this manner?

harshavardhana (changes requested):

cmd/common-main.go
polaris-phecda: The spec is automatically generated by the `swagger API`. Therefore, we can only inject the desired base path when the minio server is starting. The `GetSubPath ` function has a validity check when reading environment, so I think it's safe.

klauspost:
```
# github.com/minio/minio/cmd
Error: cmd/common-main.go:263:21: undefined: restapi.GetSubPath
make: *** [Makefile:130: build] Error 1
Error: Process completed with exit code 2.
```

harshavardhana:
This seem to have been addressed in the Console project already.

PR #16822, opened 2023-03-15
Add number of offline disks in quorum errors

vadmeste:
## Description
It will be quickly assesed what caused a quoroum error, is 
it because of majority of disks are unaccessible or something
 else.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Optimization (provides speedup with no functional changes)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [ ] Fixes a regression (If yes, please add `commit-id` or `PR #` here)
- [ ] Unit tests added/updated
- [ ] Internal documentation updated
- [ ] Create a documentation update request [here](https://github.com/minio/docs/issues/new?label=doc-change,title=Doc+Updated+Needed+For+PR+github.com%2fminio%2fminio%2fpull%2fNNNNN)

PR #17262, opened 2023-05-22
record err time when remote node is offline

jiuker:
## Description

Record err time for Client offline

## Types of changes

## Checklist:


klauspost (approved):
LGTM

harshavardhana (approved):

internal/rest/client.go
```diff
@@ -334,7 +335,7 @@ func (c *Client) LastConn() time.Time {
 func (c *Client) LastError() error {
 	c.RLock()
 	defer c.RUnlock()
-	return c.lastErr
+	return fmt.Errorf("[%s] %w", c.lastErrTime.Format(time.RFC3339), c.lastErr)
```
harshavardhana, line 343: Hmm this may break things where we don't use `errors.Is()` and there may be places where we don't use it. As this error is encapsulated into NetworkError

My suggestion is don't add it here instead add it as `LastError() (error, time.Time)`  or since no one is consuming `lastErrTime` just record it and leave it for now.

harshavardhana (changes requested):

internal/rest/client.go
jiuker: I will change it

klauspost: @harshavardhana The `%w` should wrap the error, so errors.Is can unwrap it.

harshavardhana: Yeah but NetworkError is not implemented in this manner. So it can lead to other side affects without further testing.

Because we don't know really what it might break.

harshavardhana:
With the merge of PR #16822 we can relook at this.

// cc @vadmeste

internal/rest/client.go
```diff
@@ -342,6 +343,13 @@ func (c *Client) LastError() error {
 	return c.lastErr
 }
 
+// LastErrTime returns previous error time
```
harshavardhana: We can bring back the previous change now since we have made safer changes at the caller side.

harshavardhana (changes requested):

jiuker:
@harshavardhana cc

klauspost (approved):
LGTM

harshavardhana (approved):

PR #17264, opened 2023-05-22
fix: do not allow removal of inbuilt policies unless they are already persisted

shtripat:
## Description
The command `mc admin policy remove` for an inbuilt policy like `diagnostics` doesn't have any effect in backend, even when it says policy deleted successfully. This PR removes this wrong messaging and clearly tells that inbuilt policies cannot be removed.

## Motivation and Context
`mc admin policy remove myminio diagnostics` should show

```
mc: <ERROR> Unable to remove policy. We encountered an internal error, please try again. (inbuilt policy `diagnostics` not allowed to be deleted).
```

## Types of changes

## Checklist:


cmd/iam.go
```diff
@@ -532,6 +532,12 @@ func (sys *IAMSys) DeletePolicy(ctx context.Context, policyName string, notifyPe
 		return errServerNotInitialized
 	}
 
+	for _, v := range iampolicy.DefaultPolicies {
+		if v.Name == policyName {
+			return fmt.Errorf("inbuilt policy `%s` not allowed to be deleted", policyName)
+		}
+	}
+
```
harshavardhana: this may be a bit more than we can chew, since we actually allowed default policies to be modified? - we should check if the default policies were persisted or not - if they are persisted we must allow them to be deleted().

harshavardhana (changes requested):

cmd/iam.go
shtripat: ok

cmd/iam.go
```diff
@@ -532,6 +532,17 @@ func (sys *IAMSys) DeletePolicy(ctx context.Context, policyName string, notifyPe
 		return errServerNotInitialized
 	}
 
+	m := map[string]PolicyDoc{}
+	if err := sys.store.loadPolicyDocs(ctx, m); err != nil {
+		return err
+	}
```
vadmeste: I am not sure what this code is needed, is it because it is possible to upload a custom policy with an existing name ?

vadmeste (commented):
One question

cmd/iam.go
shtripat: if policies are persisted, deletion should be allowed, so below `for` loop checks if inbuilt policies are not persisted and we are trying to delete, don't allow. Not sure if its confusing.

shtripat: @vadmeste are you saying this check itself is wrong?

vadmeste: > @vadmeste are you saying this check itself is wrong?

I just don't see why it is needed

shtripat: I understand `loadPolicyDocs` loads persisted policies in the map passed i.e. `m`.
https://github.com/minio/minio/blob/master/cmd/iam-etcd-store.go#L181
https://github.com/minio/minio/blob/master/cmd/iam-object-store.go#L170

harshavardhana: oh wait this can make the entire DeletePolicy() slow. It's better to validate if the default policies persist on disk individually?.

Well, you don't even have to read you could simply "checkIAMConfig" by doing a GetObjectInfo(). 

Instead, reloading all policies into a local map?

shtripat: Thanks @harshavardhana makes sense. So logic could be like below
```
m := map[string]PolicyDoc{}
if err := sys.store.loadIAMConfig(ctx, m, getPolicyDocPath(policyName)); err != nil && err == errConfigNotFound {
	return fmt.Errorf("inbuilt policy `%s` not allowed to be deleted", policyName)
}
```

`GeObjectInfo` part, I didnt get. Can you please elaborate that?

vadmeste: In fact, I still don't understand why sys.store.loadPolicyDocs() is needed. 

My guess is that, currently, users are able to create a custom policy with a 'readwrite' name and the name is one of the inbuilt policy, @shtripat can you confirm if that is the case ?

shtripat: @vadmeste yes, that's correct. We can overwrite inbuilt policy say `readwrite` and then allowed to delete as well. But this delete removes the one with new custom definition and if we do `mc admin policy list myminio` the policy `readwrite` gets listed with original inbuilt definition.

Only thing is if we haven't overridden the inbuilt policy, the remove command gives a false info that user is able to delete. Actually its not and listing again shows that as below
```
$ mc admin policy list myminio
readwrite
writeonly
consoleAdmin
diagnostics
readonly

$ mc admin policy remove myminio readwrite
Removed policy `readwrite` successfully.

$ mc admin policy list myminio
readwrite
writeonly
consoleAdmin
diagnostics
readonly
```

So as long as an inbuilt named policy is not customized yet, if we say removal is not allowed, it could be better message. But if we are fine with existing behavior, we can close this change I feel.

harshavardhana: Actually you are loading entire set of policies which is unexpected. Let me send a PR here to your branch.

harshavardhana (changes requested):

cmd/iam.go
shtripat: Does https://github.com/minio/minio/pull/17264#discussion_r1205293504 make sense?

harshavardhana:
Okay so I couldn't get to this today in timely manner. what we need to do is like this

 - we should check just a stat() call to see if there is a default policies persisted on disk
   then we allow removal of them at that point, if not we do not. We should only look for default policies here.
 - currently, this PR is reloading all the policies on disk which is expensive and not scalable implementation. We should be looking for only "DefaultPolicies" if they exist on disk.

So this is what I would like to see in this PR.

cmd/iam.go
```diff
@@ -532,6 +533,16 @@ func (sys *IAMSys) DeletePolicy(ctx context.Context, policyName string, notifyPe
 		return errServerNotInitialized
 	}
 
+	for _, v := range iampolicy.DefaultPolicies {
+		if v.Name == policyName {
+			// Path would be same on all endpoint
+			policyDocPath := pathJoin(globalEndpoints[0].Endpoints[0].Path, minioMetaBucket, getPolicyDocPath(policyName))
+			if _, err := os.Stat(policyDocPath); err != nil {
+				return fmt.Errorf("inbuilt policy `%s` not allowed to be deleted", policyName)
```
vadmeste: this will not work sometimes, policies are regular objects and sometimes they are not present in all disks

I believe you can use 'loadPolicyDoc()' function instead

harshavardhana: Just implement a function called statConfig() - just like readConfig() so that you want to check for existence on the disk.

harshavardhana (changes requested):

cmd/iam.go
```diff
@@ -532,6 +533,16 @@ func (sys *IAMSys) DeletePolicy(ctx context.Context, policyName string, notifyPe
 		return errServerNotInitialized
 	}
 
+	for _, v := range iampolicy.DefaultPolicies {
+		if v.Name == policyName {
+			// Path would be same on all endpoint
+			policyDocPath := pathJoin(globalEndpoints[0].Endpoints[0].Path, minioMetaBucket, getPolicyDocPath(policyName))
```
harshavardhana: This should be something like

```
statConfig()
```

harshavardhana (changes requested):

cmd/iam.go
shtripat: Ok, I will add function `statConfig` and use it here. Makes sense.

cmd/config-common.go
```diff
@@ -55,6 +56,14 @@ func readConfig(ctx context.Context, store objectIO, configFile string) ([]byte,
 	return buf, err
 }
 
+func statConfig(config string) error {
+	// Path would be same on all endpoint
+	if _, err := os.Stat(pathJoin(globalEndpoints[0].Endpoints[0].Path, minioMetaBucket, config)); err != nil {
```
harshavardhana: You need to use objlayer.GetObjectInfo()

harshavardhana (changes requested):

harshavardhana (approved):

vadmeste (approved):
LGTM

PR #17273, opened 2023-05-25
fix: ListObjectV2 return an empty dir prefix

ooninoo:
## Description

The CommonPrefix returned by ListObjectV2 is an empty folder

## How to test this PR?
Steps to reproduce:
- Create a multi-version bucket
- Create a dir/ object through the S3 API
- Upload an object dir/test.png via S3 API
- Delete the dir/test.png object through the S3 API
- Delete dir/ object via S3 API
- ListObjectV2 still returns dir/ this CommonPrefix

## Types of changes

## Checklist:

### DIFF
--- a/cmd/metacache-entries.go
+++ b/cmd/metacache-entries.go
@@ -564,6 +564,19 @@ func (m *metaCacheEntriesSorted) fileInfos(bucket, prefix, delimiter string) (ob
 
 	vcfg, _ := globalBucketVersioningSys.Get(bucket)
 
+	entryMap := make(map[string]int, 0)
+	for _, entry := range m.o {
+		name := strings.TrimPrefix(entry.name, prefix)
+		if name == "" {
+			continue
+		}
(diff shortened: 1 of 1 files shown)


klauspost:
I don't think this is a proper solution. Listing with a delimiter that is "/" will not recurse into subdirectories, so no objects will be returned from upstream with the directory as prefix.

Running tests to see if anything breaks for now.

vadmeste:
@ooninoo from the issue description, it looks like there an issue in Delete itself, not in listing

klauspost:
@vadmeste He is using a versioned bucket, so he is only placing delete markers.

vadmeste:
> @vadmeste He is using a versioned bucket, so he is only placing delete markers.

oh right, this is a known limitation, I don't think can do anything for this (otherwise a non recursive listing will need to do a recursive listing in the backend looking for a delete-marker)

PR #17274, opened 2023-05-25
avoid decommissioning DEL markers with single versions

harshavardhana:
## Description
avoid decommissioning DEL markers with single versions

## Motivation and Context
ensure that we can decommission all versions including
`null` markers with other versions from the bucket.

## How to test this PR?
Start locally with two pool setup

```
minio server /tmp/xl{1...4} /tmp/xl{5...8}

mc mb myminio/testbucket/

mc cp /etc/hosts myminio/testbucket/

mc version enable myminio/testbucket

mc cp /etc/hosts myminio/testbucket/

mc cp /etc/hosts myminio/testbucket/

mc cp /etc/hosts myminio/testbucket/

mc version suspend myminio/testbucket/

mc rm myminio/testbucket/hosts

mc ls --versions myminio/testbucket/
[2023-05-24 22:45:56 PDT]     0B STANDARD null v4 DEL hosts
[2023-05-24 22:45:50 PDT]   226B STANDARD 77a02c6b-1f02-4d98-8ee6-3eedd3d8f250 v3 PUT hosts
[2023-05-24 22:45:49 PDT]   226B STANDARD 3c7f4afe-dc42-41ec-8acd-3d3db423cef9 v2 PUT hosts
[2023-05-24 22:45:49 PDT]   226B STANDARD ba405e3f-3d3e-4ba8-bfaf-f41f2bda87f5 v1 PUT hosts

mc admin decom start /tmp/xl{5...8}

```

Once the decom starts, it must finish and the final result after the pool removal must be
```
[2023-05-24 22:45:56 PDT]     0B STANDARD null v4 DEL hosts
[2023-05-24 22:45:50 PDT]   226B STANDARD 77a02c6b-1f02-4d98-8ee6-3eedd3d8f250 v3 PUT hosts
[2023-05-24 22:45:49 PDT]   226B STANDARD 3c7f4afe-dc42-41ec-8acd-3d3db423cef9 v2 PUT hosts
[2023-05-24 22:45:49 PDT]   226B STANDARD ba405e3f-3d3e-4ba8-bfaf-f41f2bda87f5 v1 PUT hosts
```

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Optimization (provides speedup with no functional changes)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [ ] Fixes a regression (If yes, please add `commit-id` or `PR #` here)
[...]

PR #17282, opened 2023-05-25
rebalance 'null' delete markers properly

harshavardhana:
## Description
rebalance 'null' delete markers properly

## Motivation and Context
count if all versions expired via ILM, purge them
at the end. If single DEL markers are found, 
purge them as well.

## How to test this PR?
Very similar to previous PR #17274 

## Types of changes

## Checklist:

### DIFF
--- a/cmd/erasure-server-pool-rebalance.go
+++ b/cmd/erasure-server-pool-rebalance.go
@@ -506,16 +506,20 @@ func (z *erasureServerPools) rebalanceBucket(ctx context.Context, bucket string,
 
 				// Apply lifecycle rules on the objects that are expired.
 				if filterLifecycle(bucket, version.Name, version) {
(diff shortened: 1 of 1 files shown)


krisis (approved):
Good catch!

poornas (approved):

ravindk89:
@harshavardhana by "at the end" does that mean we do this en-masse post-rebalancing? Or is the intent that, during rebalancing, if we discover an object that is expirable, we expire it?

harshavardhana:
> @harshavardhana by "at the end" does that mean we do this en-masse post-rebalancing? Or is the intent that, during rebalancing, if we discover an object that is expirable, we expire it?

We don't expire just don't rebalance @ravindk89 in some scenarios.

ravindk89:
Ah OK - do we have any guidance on what those scenarios are, or is it sufficient to state "MinIO typically does not rebalance objects eligible for expiration under configured ILM rules."

PR #17287, opened 2023-05-26
fix: for delete marked objects choose appropriate parity

harshavardhana:
## Description
fix: for delete marked objects choose the appropriate parity

## Motivation and Context
deleted marked objects only need to have N/2 consistent 
copies to be called a good copy.

## How to test this PR?
Needs a specific scenario to be handled here where out of 16 striped deployments, if you have 8 inconsistent DEL markers we should still allow it. Del markers honor n/2 quorum not the same quorum as regular objects.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/erasure-metadata.go
+++ b/cmd/erasure-metadata.go
@@ -448,7 +448,7 @@ func commonParity(parities []int, defaultParityCount int) int {
 		occMap[p]++
 	}
 
-	var maxOcc, commonParity int
+	var maxOcc, cparity int
 	for parity, occ := range occMap {
 		if parity == -1 {
 			// Ignore non defined parity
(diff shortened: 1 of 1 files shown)


klauspost:
>  if you have 8 inconsistent DEL markers we should still allow it

Could you explain a bit more? I would expect 9 confirmed delete markers to not risk split brain.

harshavardhana:
> Could you explain a bit more? I would expect 9 confirmed delete markers to not risk split brain.

That is only when all nodes are online, if you have rolling restarts going on and a delete is initiated.

However nodes didn't get the call, but got the call elsewhere on the nodes. Received an error however the permanent delete has indeed succeeded due to partial rollout.

There is no rollback in this situation. 

The code here basically tries to ensure that on the read side even if you loose 8 copies and 8 has consistent copy we allow responses for Head() calls.

In this scenario btw - mc ls --versions work.

klauspost (approved):

poornas (approved):

PR #17308, opened 2023-05-30
fix: Wait 10min to load globalDNS

jiuker:
## Description

If not in containers, It will cost 10min to to refresh globalDNS first.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/common-main.go
+++ b/cmd/common-main.go
@@ -121,12 +121,11 @@ func init() {
 		} else {
 			t = time.NewTicker(10 * time.Minute)
 		}
-		globalDNSCache.RefreshWithOptions(options)
 		defer t.Stop()
 		for {
+			globalDNSCache.RefreshWithOptions(options)
 			select {
 			case <-t.C:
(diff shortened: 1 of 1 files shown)


harshavardhana (changes requested):
The correct way to do this is perhaps?

```
		for {
                        globalDNSCache.RefreshWithOptions(options)
			select {
			case <-t.C:
			case <-GlobalContext.Done():
				return
			}
		}
```

jiuker:
> The correct way to do this is perhaps?
> 
> ```
> 		for {
>                         globalDNSCache.RefreshWithOptions(options)
> 			select {
> 			case <-t.C:
> 			case <-GlobalContext.Done():
> 				return
> 			}
> 		}
> ```

yes

harshavardhana (commented):
while this is an okay fix, there is no real value in having refresh() attempted right away after DNS resolution.

The first time we have already resolved the DNS, DNS doesn't change at all in bare-metal setups so a 10min refresh can actually wait.

PR #17314, opened 2023-05-30
Draft: allow inheriting sockets from systemd

mupuf:
## Description

This enables running MinIO as a socket-activated systemd service, thus providing the same benefits of having an HTTP reverse proxy on a single node deployment, without the performance cost nor complexity of deployment.

The reason this PR is a draft is because I would like to know if this is something of interest to the minio community before spending some time reworking minio's internals to make this PR less of a hack.

Open issues:

- [ ] Figure out a good name for the sockets
- [ ] Add support for TLS sockets
- [ ] Write the documentation
- [ ] Add tests

## Motivation and Context

By having systemd binding the main socket of MinIO, we can ensure that the new connections never fail due to the service not being up, either at startup time or because the process crashed.

This provides similar benefits to running a reverse proxy, but without the overhead when the service is running on a single node.

## How to test this PR?

$ systemd-socket-activate -l 9002 --fdname server ./minio server data/

Then connect to `http://127.0.0.1:9002` using `mc`, `wget`, or any other client you may want to use.

## Types of changes

## Checklist:

### DIFF
--- a/internal/http/listener.go
+++ b/internal/http/listener.go
@@ -23,6 +23,8 @@ import (
 	"net"
 	"strings"
 	"syscall"
+
+	"github.com/coreos/go-systemd/v22/activation"
 )
 
 type acceptResult struct {
@@ -150,6 +152,21 @@ func newHTTPListener(ctx context.Context, serverAddrs []string, opts TCPOptions)
 			if strings.EqualFold(host, "localhost") {
 				isLocalhost = true
 			}
+		}
+	}
+
+	// Get the list of socket-activated listeners
+	named_listeners, _ := activation.ListenersWithNames()
+	server_listeners, ok := named_listeners["server"]
+	if ok {
+		for _, l := range server_listeners {
+			tcpListener, ok := l.(*net.TCPListener)
+			if !ok {
+				err = fmt.Errorf("unexpected listener type found %v in the socket-activated list", l)
+				return nil, err
+			}
+
+			tcpListeners = append(tcpListeners, tcpListener)
 		}
 	}
 


klauspost:
Have you made any considerations on this with regards to consistency?

We cannot allow 2 instances to run concurrently, since that could compromise write consistency. This seems to break that promise. 

Also I am not sure how locking would react to this situation - I'm not as much worried about that, but want to bring it up.

mupuf:
> Have you made any considerations on this with regards to consistency?
> 
> We cannot allow 2 instances to run concurrently, since that could compromise write consistency. This seems to break that promise.

If I understand you correctly, you are wondering what would happen if systemd were to spawn 2 instances of minio and passed the same fd to both instances? If so, then yeah, that would likely be bad.

That being said, while this is possible to achieve in a systemd unit by setting [KillMode=none](https://www.freedesktop.org/software/systemd/man/systemd.kill.html#KillMode=), this will generate a warning that this is a bad idea and deprecated.

In this light, I don't see it as a major concern. We just need to make sure the recommended systemd unit doesn't include `KillMode=none` (it doesn't), and we can add a warning to tell users to make sure only one instance runs at a time on this socket.

klauspost:
> If I understand you correctly, you are wondering what would happen if systemd were to spawn 2 instances of minio and passed the same fd to both instances?

No. What I assume is happening is that the FD is passed from the old instance to the new, while the old one is being shut down. That means that the old instance is finishing processing requests while the new is accepting new requests.

If that is not the order of operations (ie if the new instance isn't started before the old one has been stopped), then there is no problem. I just want to make sure that we don't have to instances running at any time.

mupuf:
> > If I understand you correctly, you are wondering what would happen if systemd were to spawn 2 instances of minio and passed the same fd to both instances?
> 
> No. What I assume is happening is that the FD is passed from the old instance to the new, while the old one is being shut down. That means that the old instance is finishing processing requests while the new is accepting new requests.

Well, not really. First, the `minio.socket` unit would allocate the socket and keep the endpoint in listen mode. Then, `minio.service` has a `Requires=minio.socket` command, to specify that it wants to pass the fd to the main process. Finally, minio's server cmd gets the information about the fd that was passed to it via environment variables. This means the old and new processes don't need to communicate any data.

> 
> If that is not the order of operations (ie if the new instance isn't started before the old one has been stopped), then there is no problem. I just want to make sure that we don't have to instances running at any time.

Yeah, this is the case: [Service restart requests are implemented as stop operations followed by start operations. This means that ExecStop= and ExecStopPost= are executed during a service restart operation](https://www.freedesktop.org/software/systemd/man/systemd.service.html#ExecStop=)

harshavardhana:
> By having systemd binding the main socket of MinIO, we can ensure that the new connections never fail due to the service not being up, either at startup time or because the process crashed.

So how long do the client connections are queued up? what is the queue size? 

> The reason this PR is a draft is because I would like to know if this is something of interest to the minio community before spending some time reworking minio's internals to make this PR less of a hack.

I don't think we would need another socket file opened up and available. It is a tangential deployment paradigm that we would like to avoid - it is simply confusing and a bit too much "unix" to avoid. In MinIO we already queue up incoming I/O at the handler layer while the server is getting initialized. As soon as server initializes the caller gets to perform their request.

Most of our deployments are in k8s/containers (reverse proxy is baked in) - Baremetal deployments don't need new areas of understanding customers tend to already have the necessary infrastructure to go with bare metal deployments.

mupuf:
>  So how long do the client connections are queued up?

Systemd doesn't do anything fancy here: It just creates the socket we want (in this case, TCP), and calls Listen() on it. Of course, we can tweak all the parameters, but that's essentially it.

So, this means the kernel handles the SYN, SYN/ACK, ACK dance... at which point the connection is made and it will remain active until the client cuts the connection due to inactivity. I would assume ~5 minutes for a browser, not sure about `mc`.

> what is the queue size?

The default queue size is SOMAXCONN, which appears to be 128. It can however be modified, like many other parameters: https://www.freedesktop.org/software/systemd/man/systemd.socket.html#Backlog=

> Most of our deployments are in k8s/containers (reverse proxy is baked in) - Baremetal deployments don't need new areas of understanding customers tend to already have the necessary infrastructure to go with bare metal deployments.

Sounds like the best deployment strategy for high availability clusters! Nothing to change here :)

This PR brings benefits to the SNSD deployments:

 * On-demand startup: lowering RAM usage by ~300 MB on amd64 and improving boot time of developers' machines
 * Higher reliability for single-node systems controlling other machines, like for bare-metal CI for driver development (what I work on)

As you can see, maintaining the code won't be much trouble, since the systemd sockets can be imported as TCPListenners just like the sockets created by MinIO. IMO, the most difficult thing to do is settling up on an interface that can be maintained without hindering future MinIO developments, and documenting it in places like:

 *  https://min.io/docs/minio/linux/operations/install-deploy-manage/deploy-minio-single-node-single-drive.html#create-the-systemd-service-file
 * https://github.com/minio/minio-service/tree/master/linux-systemd.

If you feel like there is a chance for me to pull this off in a way that would be acceptable to you, I am willing to spend the time needed to get this feature done. Otherwise, I understand that you may not want to change your mental model, so no hurt feelings, I'll just maintain a fork with the changes for the purpose of my needs.

harshavardhana:
There are some good things in this PR however if this feature works out of the box, i.e we just deploy the binary and forget and no more changes are expected from systemd configuration or minio service file

We are willing to take it in - otherwise, we are not going to take this in. 

I am going to close this PR since it is incomplete. I cannot guarantee any merge even if the PR is complete - it needs to satisfy some simplicity guarantees first.

Thank you for all the effort.

PR #17329, opened 2023-06-01
fix: mc head will close the connection

jiuker:
## Description

fix: mc head will close the connection

## Types of changes

## Checklist:

### DIFF
--- a/cmd/object-handlers.go
+++ b/cmd/object-handlers.go
@@ -24,6 +24,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"net/http/httptest"
 	"net/url"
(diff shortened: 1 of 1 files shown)


harshavardhana:
This PR doesn't make sense @jiuker mc head will invoke only HeadObject not GetObject you are fixing something in the wrong path.

All client connections close when the handler returns, so not sure what is being fixed here.

cmd/object-handlers.go
```diff
@@ -548,12 +549,20 @@ func (api objectAPIHandlers) getObjectHandler(ctx context.Context, objectAPI Obj
 			writeErrorResponse(ctx, w, toAPIError(ctx, err), r.URL)
 			return
 		}
+		// `mc head` will close the connection, will make this error
```
harshavardhana, line 552: We need to move this type of check into IsNetworkOrHostDown(err) 

So that we can ignore printing the logs instead of introducing a goto jump

harshavardhana (changes requested):

jiuker:
update go depend will fix by auto.

PR #17349, opened 2023-06-05
Add disk usage high watermark

dormanze:
## Description
Many times, we can't pay attention to the disk usage at all times. When the disk usage reaches 100%, it is difficult to clean up some data because there is no available space. When the disk usage reaches 100%, some data may be damaged if the minio restarts. Therefore, I suggest adding two high watermarks to specify the disk usage percentage or the disk usage upper limit to prevent the disk usage from reaching 100%.

## Motivation and Context
To prevent high disk usage, it is difficult to clear data.

## How to test this PR?
step1:
```
export MINIO_ALLOCATION_DISK_HIGH_WATERMARK_PERCENT=90%
or
export MINIO_ALLOCATION_DISK_HIGH_WATERMARK_SIZE=200G
```
step2:
start minio server.

## Types of changes

## Checklist:


jiuker (changes requested):
We can achieve the same effect by using limited capacity. Set `Enable Bucket Quota` instead.

dormanze:
> We can achieve the same effect by using limited capacity. Set `Enable Bucket Quota` instead.

In the absence of a foreseeable upper bucket limit, I think this capability is needed to at least give the user an optional option.

klauspost:
This is what `diskFillFraction` is for. We don't plans any intention to make that configurable.

`diskReserveFraction` is for multiple pools.

dormanze:
> This is what `diskFillFraction` is for. We don't plans any intention to make that configurable.
> 
> `diskReserveFraction` is for multiple pools.

The default value of diskFillFraction is 0.99 , the disk usage may reach 100%. Data cleaning can be difficult and there is a risk of data corruption. So I suggest opening this parameter to have more options without damaging the original features.

klauspost:
@dormanze Adding an option is adding maintenance and one more thing to worry about.

Your issue is basically a monitoring issue. Solving it with another option is not the way forward. The way forward is to set up proper monitoring to ensure you never get close to the limit.

I will leave the final decision to @harshavardhana

harshavardhana:
We will not be taking in more ways to configure this, there are no real benefits beyond what we have already chosen as defaults.

Beyond that monitor your disk usage via Prometheus node_exporter or MinIO

PR #17351, opened 2023-06-05
allow quota enforcement to rely on older values

harshavardhana:
## Description
allow quota enforcement to rely on older values

## Motivation and Context
PUT calls cannot afford to have large latency build-ups due
to contentious `usage.json`, or worse letting them fail with
some unexpected error, this can happen when this file is
concurrently being updated via scanner or it is being
healed during a disk replacement heal.

However, these are fairly quick in theory, stressed clusters
can quickly show visible latency this can add up leading to
invalid errors returned during PUT.

It is perhaps okay for us to relax this error return requirement
instead, make sure that we log that we are proceeding to take in
the requests while the quota is using an older value for the quota
enforcement. These things will reconcile themselves eventually,
via scanner making sure to overwrite the `usage.json`.

Bonus: make sure that storage-rest-client sets ExpectTimeouts to
be 'true', such that DiskInfo() call with contextTimeout does
not prematurely disconnect the servers leading to a longer
healthCheck, back-off routine. This can easily pile up while also
causing active callers to disconnect, leading to quorum loss.

DiskInfo is actively used in the PUT, Multipart call path for
upgrading parity when disks are down, it in-turn shouldn't cause
more disks to go down.

## How to test this PR?
To reproduce the quota error make the following changes in the code
from master branch 

```diff
diff --git a/cmd/data-usage.go b/cmd/data-usage.go
index cecbb25a1..e9959af8b 100644
--- a/cmd/data-usage.go
+++ b/cmd/data-usage.go
@@ -21,6 +21,7 @@ import (
        "context"
        "errors"
        "strings"
+       "time"
 
        jsoniter "github.com/json-iterator/go"
        "github.com/minio/minio/internal/logger"
@@ -92,6 +93,8 @@ func loadPrefixUsageFromBackend(ctx context.Context, objAPI ObjectLayer, bucket
 }
 
 func loadDataUsageFromBackend(ctx context.Context, objAPI ObjectLayer) (DataUsageInfo, error) {
+       time.Sleep(5100 * time.Millisecond)
+
        buf, err := readConfig(ctx, objAPI, dataUsageObjNamePath)
        if err != nil {
                if errors.Is(err, errConfigNotFound) {
```

Then perform the following commands on SNSD or a distributed setup
```
mc mb alias/mybucket
mc cp /etc/hosts alias/mybucket/ (this succeeds all good)
mc quota set alias/mybucket/ --size 10Gi
mc cp /etc/hosts alias/mybucket (this will never complete and fail)
```

To reproduce the DiskInfo taking nodes offline
```diff
diff --git a/cmd/xl-storage.go b/cmd/xl-storage.go
index af148f8c6..229db81ec 100644
--- a/cmd/xl-storage.go
+++ b/cmd/xl-storage.go
@@ -613,6 +613,8 @@ func (s *xlStorage) NSScanner(ctx context.Context, cache dataUsageCache, updates
 // DiskInfo provides current information about disk space usage,
 // total free inodes and underlying filesystem.
 func (s *xlStorage) DiskInfo(_ context.Context) (info DiskInfo, err error) {
+       time.Sleep(5100 * time.Millisecond)
+
        s.diskInfoCache.Once.Do(func() {
                s.diskInfoCache.TTL = time.Second
                s.diskInfoCache.Update = func() (interface{}, error) {
```

This will make remote calls for DiskInfo to mark the disk/node as offline, 
failing an on-going PUT operation with write quorum errors.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/bucket-quota.go
+++ b/cmd/bucket-quota.go
@@ -47,9 +47,13 @@ func NewBucketQuotaSys() *BucketQuotaSys {
 // Init initialize bucket quota.
 func (sys *BucketQuotaSys) Init(objAPI ObjectLayer) {
 	sys.bucketStorageCache.Once.Do(func() {
-		sys.bucketStorageCache.TTL = 1 * time.Second
+		// Set this to 10 secs since its enough, as scanner
+		// does not update the bucket usage values frequently.
+		sys.bucketStorageCache.TTL = 10 * time.Second
+		// Rely on older value if usage loading fails from disk.
+		sys.bucketStorageCache.Relax = true
 		sys.bucketStorageCache.Update = func() (interface{}, error) {
-			ctx, done := context.WithTimeout(context.Background(), 5*time.Second)
+			ctx, done := context.WithTimeout(context.Background(), 1*time.Second)
 			defer done()
 
 			return loadDataUsageFromBackend(ctx, objAPI)
@@ -60,16 +64,20 @@ func (sys *BucketQuotaSys) Init(objAPI ObjectLayer) {
 // GetBucketUsageInfo return bucket usage info for a given bucket
 func (sys *BucketQuotaSys) GetBucketUsageInfo(bucket string) (BucketUsageInfo, error) {
 	v, err := sys.bucketStorageCache.Get()
-	if err != nil {
-		return BucketUsageInfo{}, err
+	if err != nil && v != nil {
+		logger.LogIf(GlobalContext, fmt.Errorf("unable to retrieve usage information for bucket: %s, relying on older value cached in-memory: err(%v)", bucket, err))
 	}
+	if v == nil {
+		logger.LogIf(GlobalContext, errors.New("unable to retrieve usage information for bucket: %s, no reliable usage value available - quota will not be enforced"))
+	}
+
+	var bui BucketUsageInfo
 
 	dui, ok := v.(DataUsageInfo)
-	if !ok {
-		return BucketUsageInfo{}, fmt.Errorf("internal error: Unexpected DUI data type: %T", v)
+	if ok {
+		bui = dui.BucketsUsage[bucket]
 	}
 
-	bui := dui.BucketsUsage[bucket]
 	return bui, nil
 }
 
@@ -100,6 +108,10 @@ func (sys *BucketQuotaSys) enforceQuotaHard(ctx context.Context, bucket string, 
 	}
 
 	if q != nil && q.Type == madmin.HardQuota && q.Quota > 0 {
+		if uint64(size) >= q.Quota { // check if file size already exceeds the quota
+			return BucketQuotaExceeded{Bucket: bucket}
+		}
+
 		bui, err := sys.GetBucketUsageInfo(bucket)
 		if err != nil {
 			return err
--- a/cmd/lock-rest-client.go
+++ b/cmd/lock-rest-client.go
@@ -151,10 +151,8 @@ func newlockRESTClient(endpoint Endpoint) *lockRESTClient {
 	}
 
 	restClient := rest.NewClient(serverURL, globalInternodeTransport, newCachedAuthToken())
-	restClient.ExpectTimeouts = true
 	// Use a separate client to avoid recursive calls.
 	healthClient := rest.NewClient(serverURL, globalInternodeTransport, newCachedAuthToken())
-	healthClient.ExpectTimeouts = true
 	healthClient.NoMetrics = true
 	restClient.HealthCheckFn = func() bool {
 		ctx, cancel := context.WithTimeout(context.Background(), restClient.HealthCheckTimeout)
--- a/cmd/namespace-lock.go
+++ b/cmd/namespace-lock.go
@@ -200,8 +200,8 @@ func (di *distLockInstance) GetRLock(ctx context.Context, timeout *dynamicTimeou
 	}) {
 		timeout.LogFailure()
 		defer cancel()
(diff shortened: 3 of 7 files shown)


cmd/bucket-quota.go
```diff
@@ -48,8 +48,9 @@ func NewBucketQuotaSys() *BucketQuotaSys {
 func (sys *BucketQuotaSys) Init(objAPI ObjectLayer) {
 	sys.bucketStorageCache.Once.Do(func() {
 		sys.bucketStorageCache.TTL = 1 * time.Second
```
vadmeste: I think let's increase TTL as well. One second seems to be unnecessary low

vadmeste (commented):
I am thinking about this:

```diff
diff --git a/internal/rest/client.go b/internal/rest/client.go
index 1fba815f0..1369ffd7d 100644
--- a/internal/rest/client.go
+++ b/internal/rest/client.go
@@ -241,6 +241,11 @@ func (c *Client) Call(ctx context.Context, method string, values url.Values, bod
        u.Path = path.Join(u.Path, method)
        u.RawQuery = values.Encode()
 
+       expectTimeout := c.ExpectTimeouts
+       if _, ok := ctx.Deadline(); ok {
+               expectTimeout = false
+       }
+
        req, err := c.newRequest(ctx, u, body)
        if err != nil {
                return nil, &NetworkError{err}
@@ -254,7 +259,7 @@ func (c *Client) Call(ctx context.Context, method string, values url.Values, bod
 
        resp, err := c.httpClient.Do(req)
        if err != nil {
-               if xnet.IsNetworkOrHostDown(err, c.ExpectTimeouts) {
+               if xnet.IsNetworkOrHostDown(err, expectTimeout) {
                        if !c.NoMetrics {
                                atomic.AddUint64(&globalStats.errs, 1)
                        }
@@ -284,7 +289,7 @@ func (c *Client) Call(ctx context.Context, method string, values url.Values, bod
                // Limit the ReadAll(), just in case, because of a bug, the server responds with large data.
                b, err := io.ReadAll(io.LimitReader(resp.Body, c.MaxErrResponseSize))
                if err != nil {
-                       if xnet.IsNetworkOrHostDown(err, c.ExpectTimeouts) {
+                       if xnet.IsNetworkOrHostDown(err, expectTimeout) {
                                if !c.NoMetrics {
                                        atomic.AddUint64(&globalStats.errs, 1)
                                }
@@ -299,7 +304,7 @@ func (c *Client) Call(ctx context.Context, method string, values url.Values, bod
                }
                return nil, errors.New(resp.Status)
        }
-       if !c.NoMetrics && !c.ExpectTimeouts {
+       if !c.NoMetrics && expectTimeout {
                resp.Body = &respBodyMonitor{ReadCloser: resp.Body, expectTimeouts: c.ExpectTimeouts}
        }
        return resp.Body, nil
```

If the caller sent a context with a deadline, ignore context deadline.. if not, we can honor underlying transport timeout

harshavardhana:
```
diff --git a/internal/rest/client.go b/internal/rest/client.go
index 1fba815f0..1369ffd7d 100644
--- a/internal/rest/client.go
+++ b/internal/rest/client.go
@@ -241,6 +241,11 @@ func (c *Client) Call(ctx context.Context, method string, values url.Values, bod
        u.Path = path.Join(u.Path, method)
        u.RawQuery = values.Encode()
 
+       expectTimeout := c.ExpectTimeouts
+       if _, ok := ctx.Deadline(); ok {
+               expectTimeout = false
+       }
```

This is still causing the deadline to be considered has timeouts @vadmeste - if you see ExpectTimeouts is actually false here for everyone, we need it to be 'True'

cmd/bucket-quota.go
harshavardhana: I can change that.

cmd/bucket-quota.go
```diff
@@ -108,6 +113,10 @@ func (sys *BucketQuotaSys) enforceQuotaHard(ctx context.Context, bucket string,
 		if bui.Size > 0 && ((bui.Size + uint64(size)) >= q.Quota) {
 			return BucketQuotaExceeded{Bucket: bucket}
 		}
+
+		if uint64(size) >= q.Quota {
+			return BucketQuotaExceeded{Bucket: bucket}
+		}
```
poornas: this can be moved to line 108 - before getting the cached usage

vadmeste:
> This is still causing the deadline to be considered has timeouts @vadmeste - if you see ExpectTimeouts is actually false here for everyone, we need it to be 'True'

Yes but this is fine. My idea is just to separate two uses cases:
- First case, the code sets up a deadline, we do not have to mark a node as offline
- Second use case, the code did not set up any deadline, but golang/net will issue a deadline error, that we can honor and mark the node offline and reconnect.

harshavardhana:
> > This is still causing the deadline to be considered has timeouts @vadmeste - if you see ExpectTimeouts is actually false here for everyone, we need it to be 'True'
> 
> Yes but this is fine. My idea is just to separate two uses cases:
> 
> * First case, the code sets up a deadline, we do not have to mark a node as offline
> * Second use case, the code did not set up any deadline, but golang/net will issue a deadline error, that we can honor and mark the node offline and reconnect.

Go will never return context.DeadlineExceeded error @vadmeste since they themselves will never add custom contexts without callers approval.

What you are suggesting is not needed. Expect timeouts only operate for this special error and we need to do this correctly if we use deadline on contexts.

vadmeste (approved):
LGTM

poornas (approved):

PR #17353, opened 2023-06-05
listen: Only error out if not able to bind any interface

vadmeste:
## Description
Quit only when the server does not find any interface to listen on.

Otherwise print logs why it is not able to bind to other interfaces.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/server-main.go
+++ b/cmd/server-main.go
@@ -612,7 +612,14 @@ func serverMain(ctx *cli.Context) {
 
 	httpServer.TCPOptions.Trace = bootstrapTrace
 	go func() {
-		globalHTTPServerErrorCh <- httpServer.Start(GlobalContext)
+		serveFn, err := httpServer.Init(GlobalContext, func(listenAddr string, err error) {
+			logger.LogIf(GlobalContext, fmt.Errorf("Unable to listen on `%s`: %v", listenAddr, err))
+		})
+		if err != nil {
+			globalHTTPServerErrorCh <- err
+			return
+		}
+		globalHTTPServerErrorCh <- serveFn()
 	}()
 
 	bootstrapTrace("setHTTPServer")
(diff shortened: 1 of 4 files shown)


cmd/server-main.go
```diff
@@ -600,7 +600,8 @@ func serverMain(ctx *cli.Context) {
 	}
 
 	bootstrapTrace("xhttp.NewServer")
-	httpServer := xhttp.NewServer(getServerListenAddrs()).
+	listenAddrs := getServerListenAddrs()
```
harshavardhana: If you want current listenAddrs() move this as a function function in `httpServer.ListenAddrs()`  @vadmeste

cmd/server-main.go
```diff
@@ -612,7 +613,18 @@ func serverMain(ctx *cli.Context) {
...
+			globalHTTPServerErrorCh <- err
+			return
+		}
+		for i, e := range listenErrs {
+			if e == nil {
+				continue
+			}
+			logger.LogIf(GlobalContext, fmt.Errorf("Unable to listen on `%s`: %v", listenAddrs[i], listenErrs[i]))
+		}
+		globalHTTPServerErrorCh <- serveFn()
```
harshavardhana: Pass a `function()` to print the errors if any like this

```
    serve, err := httpServer.Init(GlobalContext, func(listenAddr string, err error) {
    if err != nil {
         logger.LogIf(GlobalContext, fmt.Errorf("Unable to listen on `%s`: %v",  listenAddr, err)
    })
    if err != nil {
        globalHTTPServerErrorCh <- err
	return
    }
    globalHTTPServerErrorCh <- serve()
```

harshavardhana (changes requested):

cmd/server-main.go
vadmeste: Done

harshavardhana (approved):

klauspost (approved):
LGTM

PR #17363, opened 2023-06-06
fix: kafka broker pings must not be greater than 1sec

harshavardhana:
## Description
fix: Kafka broker pings must not be greater than 1sec

## Motivation and Context
optimize checking Kafka brokers during the audit

## How to test this PR?
Setup Kafka via environment variable and then setup `nc -l 9092` to see the calls fail

## Types of changes

## Checklist:
- [x] Fixes a regression since #17121

### DIFF
--- a/internal/event/target/kafka.go
+++ b/internal/event/target/kafka.go
@@ -28,6 +28,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"sync"
 	"time"
 
 	"github.com/minio/minio/internal/event"
@@ -161,7 +162,7 @@ func (target *KafkaTarget) IsActive() (bool, error) {
 }
 
 func (target *KafkaTarget) isActive() (bool, error) {
-	if !target.args.pingBrokers() {
+	if err := target.args.pingBrokers(); err != nil {
 		return false, store.ErrNotConnected
 	}
 	return true, nil
@@ -268,14 +269,30 @@ func (target *KafkaTarget) Close() error {
 }
 
 // Check if atleast one broker in cluster is active
-func (k KafkaArgs) pingBrokers() bool {
-	d := net.Dialer{Timeout: 60 * time.Second}
-	for _, broker := range k.Brokers {
-		if _, err := d.Dial("tcp", broker.String()); err == nil {
-			return true
+func (k KafkaArgs) pingBrokers() (err error) {
+	d := net.Dialer{Timeout: 1 * time.Second}
+
+	errs := make([]error, len(k.Brokers))
+	var wg sync.WaitGroup
+	for idx, broker := range k.Brokers {
+		broker := broker
+		idx := idx
+		wg.Add(1)
+		go func(broker xnet.Host, idx int) {
+			defer wg.Done()
+
+			_, errs[idx] = d.Dial("tcp", broker.String())
+		}(broker, idx)
+	}
+	wg.Wait()
+
+	var retErr error
+	for _, err := range errs {
+		if err != nil {
+			retErr = err
 		}
 	}
-	return false
+	return retErr
 }
 
 func (target *KafkaTarget) init() error {
@@ -295,6 +312,7 @@ func (target *KafkaTarget) initKafka() error {
 		config.Version = kafkaVersion
 	}
 
+	config.Net.KeepAlive = 60 * time.Second
 	config.Net.SASL.User = args.SASL.User
 	config.Net.SASL.Password = args.SASL.Password
 	initScramClient(args, config) // initializes configured scram client.
--- a/internal/logger/target/kafka/kafka.go
+++ b/internal/logger/target/kafka/kafka.go
@@ -32,6 +32,7 @@ import (
 	"time"
 
 	"github.com/minio/pkg/logger/message/audit"
+	"github.com/minio/pkg/logger/message/log"
 
 	"github.com/Shopify/sarama"
 	saramatls "github.com/Shopify/sarama/tools/tls"
@@ -75,15 +76,29 @@ type Config struct {
 
 // Check if atleast one broker in cluster is active
 func (k Config) pingBrokers() (err error) {
-	d := net.Dialer{Timeout: 60 * time.Second}
+	d := net.Dialer{Timeout: 1 * time.Second}
 
-	for _, broker := range k.Brokers {
-		_, err = d.Dial("tcp", broker.String())
+	errs := make([]error, len(k.Brokers))
+	var wg sync.WaitGroup
+	for idx, broker := range k.Brokers {
+		broker := broker
+		idx := idx
+		wg.Add(1)
+		go func(broker xnet.Host, idx int) {
+			defer wg.Done()
+
+			_, errs[idx] = d.Dial("tcp", broker.String())
+		}(broker, idx)
+	}
+	wg.Wait()
+
+	var retErr error
+	for _, err := range errs {
 		if err != nil {
-			return err
+			retErr = err
 		}
 	}
-	return nil
+	return retErr
 }
 
 // Target - Kafka target.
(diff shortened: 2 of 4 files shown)


harshavardhana:
This is an important fix, without this if Kafka endpoint is offline all API calls hang for 1min or more.

harshavardhana:
PTAL @Praveenrajmani

internal/event/target/kafka.go
```diff
@@ -268,14 +269,30 @@ func (target *KafkaTarget) Close() error {
...
+	errs := make([]error, len(k.Brokers))
+	var wg sync.WaitGroup
+	for idx, broker := range k.Brokers {
+		broker := broker
+		idx := idx
+		wg.Add(1)
+		go func(broker xnet.Host, idx int) {
+			defer wg.Done()
+
+			_, errs[idx] = d.Dial("tcp", broker.String())
```
Praveenrajmani, line 284: do we need to ping all the brokers here? Can we return if at least one broker is active?

internal/event/target/kafka.go
```diff
@@ -268,14 +269,30 @@ func (target *KafkaTarget) Close() error {
 }
 
 // Check if atleast one broker in cluster is active
-func (k KafkaArgs) pingBrokers() bool {
-	d := net.Dialer{Timeout: 60 * time.Second}
-	for _, broker := range k.Brokers {
-		if _, err := d.Dial("tcp", broker.String()); err == nil {
-			return true
+func (k KafkaArgs) pingBrokers() (err error) {
+	d := net.Dialer{Timeout: 1 * time.Second}
```
Praveenrajmani, line 273: suggestion: can we keep timeout to be 3 secs as 1 sec is very low?

internal/logger/target/kafka/kafka.go
```diff
@@ -75,32 +76,49 @@ type Config struct {
 
 // Check if atleast one broker in cluster is active
 func (k Config) pingBrokers() (err error) {
-	d := net.Dialer{Timeout: 60 * time.Second}
+	d := net.Dialer{Timeout: 1 * time.Second}
```
Praveenrajmani, line 79: suggestion: can we keep timeout to be 3 secs as 1 sec is very low?

internal/logger/target/kafka/kafka.go
```diff
@@ -75,32 +76,49 @@ type Config struct {
...
+	errs := make([]error, len(k.Brokers))
+	var wg sync.WaitGroup
+	for idx, broker := range k.Brokers {
+		broker := broker
+		idx := idx
+		wg.Add(1)
+		go func(broker xnet.Host, idx int) {
+			defer wg.Done()
+
+			_, errs[idx] = d.Dial("tcp", broker.String())
```
Praveenrajmani, line 90: same here - do we need to dial all the brokers?

Praveenrajmani:
I can see the following error when queue_dir is configured

```
API: SYSTEM()
Time: 06:00:50 UTC 06/07/2023
DeploymentID: cb0fc82e-8d45-4870-89b5-7cba32998199
Error: target.SendFromStore() failed with 'unsupported data structure' (*fmt.wrapError)
       5: internal/logger/logonce.go:118:logger.(*logOnceType).logOnceIf()
       4: internal/logger/logonce.go:149:logger.LogOnceIf()
       3: internal/store/store.go:105:store.sendItems.func1()
       2: internal/store/store.go:129:store.sendItems()
       1: internal/store/store.go:144:store.StreamItems[...].func1()
```

internal/event/target/kafka.go
klauspost: Is it, though? Do we expect Kafka servers to be on a different continent?

internal/event/target/kafka.go
```diff
@@ -268,14 +269,30 @@ func (target *KafkaTarget) Close() error {
...
+
+			_, errs[idx] = d.Dial("tcp", broker.String())
+		}(broker, idx)
+	}
+	wg.Wait()
+
+	var retErr error
+	for _, err := range errs {
+		if err != nil {
+			retErr = err
```
klauspost, line 292: This doesn't match the function signature. It says it will succeed if *one* broker is reachable. This means that *all* brokers must be reachable.

Without too much info, I think one should be fine. But YMMV.

internal/event/target/kafka.go
klauspost: Yes on the first point (we should ping all concurrently). IMO, yes.

internal/logger/target/kafka/kafka.go
```diff
@@ -75,32 +76,49 @@ type Config struct {
...
+	// Reading logCh must hold read lock on logChMu (to avoid read race)
+	// Sending a value on logCh must hold read lock on logChMu (to avoid closing)
+	logCh   chan interface{}
+	logChMu sync.RWMutex
 
 	// store to persist and replay the logs to the target
 	// to avoid missing events when the target is down.
-	store              store.Store[audit.Entry]
-	storeCtxCancel     context.CancelFunc
+	store          store.Store[interface{}]
```
klauspost, line 119: Why make it weaker? Are you fixing an issue?

klauspost (changes requested):

internal/event/target/kafka.go
Praveenrajmani: i think we can expect upto 2-3 secs in k8s.. but it would still indicate a bad n/w. we can fix this if needed.

internal/logger/target/kafka/kafka.go
harshavardhana: No 1 sec is important we want Kafka to reply fast.

internal/logger/target/kafka/kafka.go
harshavardhana: yeah because kafka will be used with logger as well @klauspost so same target.

internal/logger/target/kafka/kafka.go
harshavardhana: yes

vadmeste:
@harshavardhana something is missing here, which is decoupling Send() from pinging brokers when kafka is offline. But I can do in another PR, it seems it needs a more involved work.

harshavardhana:
> @harshavardhana something is missing here, which is decoupling Send() from pinging brokers when kafka is offline. But I can do in another PR, it seems it needs a more involved work.

Send() will initialize the Kafka brokers if not present. However it won't block. 

We do not need to initialize Kafka while doing Send() it must be during logEntry() @vadmeste

harshavardhana:
The issue was a regression was added asking Kafka to initialize with 60sec timeout before sending the audit to logCh.

This PR fixes that.

harshavardhana:
This PR also ensures that we can eventually use Kafka for logger target as well.

harshavardhana:
> I can see the following error when queue_dir is configured
> 
> ```
> API: SYSTEM()
> Time: 06:00:50 UTC 06/07/2023
> DeploymentID: cb0fc82e-8d45-4870-89b5-7cba32998199
> Error: target.SendFromStore() failed with 'unsupported data structure' (*fmt.wrapError)
>        5: internal/logger/logonce.go:118:logger.(*logOnceType).logOnceIf()
>        4: internal/logger/logonce.go:149:logger.LogOnceIf()
>        3: internal/store/store.go:105:store.sendItems.func1()
>        2: internal/store/store.go:129:store.sendItems()
>        1: internal/store/store.go:144:store.StreamItems[...].func1()
> ```

This is fixed because of the type loss from marshaling and unmarshaling as interface - latest is actually here https://github.com/harshavardhana/minio/commit/cd3063ba3ff77d6f848b880c0b7c9a0f7eaa5bf4

Unsure why GitHub is not reflecting the changes.

PR #17364, opened 2023-06-06
decom: Parallelize decommissining

vadmeste:
## Types of changes

## Checklist:


vadmeste:
@harshavardhana let's say if we want to decom two pools, 1 and 2 in parallel. Shouldn't the first node of the first pool decommission the first pool, and the first node of the second pool decommission the second pool ?

I did not understand the proxing logic of the old code. I changed the admin API to support only one pool decom request for simplicity and we can make mc sends multiple decom requests. This is only for code simplicity.

Let me know if you think that the old code is working and not buggy.

harshavardhana (approved):

poornas (approved):

PR #17380, opened 2023-06-07
Redact all secrets from config viewing APIs

donatello:
## Description
This change adds a `Secret` property to `HelpKV` to identify secrets like passwords and auth tokens that should not be revealed by the server in its configuration fetching APIs. Functions that retrieve secrets for display purposes now return `***` as the secret values.

## How to test this PR?
Query configuration with mc or UI and ensure that secrets are redacted.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/admin-handlers-config-kv.go
+++ b/cmd/admin-handlers-config-kv.go
@@ -241,6 +241,8 @@ func setConfigKV(ctx context.Context, objectAPI ObjectLayer, kvBytes []byte) (re
 // 1. `subsys:target` -> request for config of a single subsystem and target pair.
 // 2. `subsys:` -> request for config of a single subsystem and the default target.
 // 3. `subsys` -> request for config of all targets for the given subsystem.
+//
+// This is a reporting API and config secrets are redacted in the response.
 func (a adminAPIHandlers) GetConfigKVHandler(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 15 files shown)


harshavardhana (dismissed):
how will use take backups of their config during `mc admin config export`  if we redact them?

donatello:
> how will use take backups of their config during `mc admin config export` if we redact them?

Yeah just noticed, fixing.

harshavardhana:
@donatello wanted one internal discussion on this before we merge and make a release with @abperiasamy - it will be short.

donatello:
This PR is now updated - secrets are never returned by configuration reporting APIs. They are not redacted as `***` or some other dummy string to prevent them from being treated as data or passwords as discussed with @abperiasamy 

The export API which exports configuration for backup purposes is not affected by this change and will continue to work as expected. Access to that API can be controlled via denying the `GetConfig` IAM permission, and with an upcoming change in how policies may be attached to users.

This PR is now ready for review.

harshavardhana (approved):

harshavardhana:
PTAL @allanrogerr @cniackz @cesnietor

vadmeste (approved):
I would say viewing secrets can be allowed to users who have SetConfig permission but this can be confusing to users unless we make mc/Console inform users about this

(whoever can set a new config or doing import IAM should be able to see secrets)

LGTM otherwise

PR #17389, opened 2023-06-09
tar: Avoid storing snowball extraction header in extract objects

vadmeste:
## Description
Extracted objects inside the archive file are not necessarly tar, avoid saving the 
snowball archive header with the extracted objects.

Also avoid storing empty storage class in extarcted objects metadata as well

## Types of changes

## Checklist:

### DIFF
--- a/cmd/object-handlers.go
+++ b/cmd/object-handlers.go
@@ -2106,7 +2106,19 @@ func (api objectAPIHandlers) PutObjectExtractHandler(w http.ResponseWriter, r *h
 		writeErrorResponse(ctx, w, toAPIError(ctx, err), r.URL)
 		return
 	}
-	metadata[xhttp.AmzStorageClass] = sc
+
+	// Remove the snowball headers from metadata from objects
+	// since those are actionable headers and not for storage
+	delete(metadata, xhttp.AmzSnowballExtract)
+	delete(metadata, xhttp.AmzSnowballExtract)
+	delete(metadata, xhttp.MinIOSnowballIgnoreDirs)
+	delete(metadata, xhttp.MinIOSnowballIgnoreErrors)
+	delete(metadata, xhttp.MinIOSnowballPrefix)
+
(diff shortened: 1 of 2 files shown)


cmd/object-handlers.go
```diff
@@ -2106,7 +2106,15 @@ func (api objectAPIHandlers) PutObjectExtractHandler(w http.ResponseWriter, r *h
 		writeErrorResponse(ctx, w, toAPIError(ctx, err), r.URL)
 		return
 	}
-	metadata[xhttp.AmzStorageClass] = sc
+
+	// Remove the snowball header to avoid trying to untar files inside
+	// the archive when they are copied somewhere else later
+	delete(metadata, xhttp.AmzSnowballExtract)
```
harshavardhana, line 2112: There are other headers as well such as specifying prefix, skipping dirs etc.

harshavardhana (changes requested):

cmd/object-handlers.go
harshavardhana: Such as

- X-Amz-Meta-Minio-Snowball-Ignore-Dirs
- X-Amz-Meta-Minio-Snowball-Ignore-Errors
- X-Amz-Meta-Minio-Snowball-Prefix

klauspost: Like *all* `X-Amz-Meta-Minio-`? Are there any we want to keep?

harshavardhana (approved):

klauspost:
@harshavardhana What are the headers `X-Amz-Meta-Minio-` we want to keep?

harshavardhana:
> @harshavardhana What are the headers `X-Amz-Meta-Minio-` we want to keep?

We are removing that we interpret them to be special in the handler layer 
since they do not belong on the object, the rest of them can be saved.

klauspost (approved):

Issue #17390, opened 2023-06-10
SignatureDoesNotMatch while response-content-disposition contains & 


TestMsr:
## NOTE
presigned url with & in response-content-disposition cause download  failed

## Expected Behavior

download succes with correct filename
## Current Behavior

minio return SignatureDoesNotMatch
## Possible Solution

 the preSignatureV2 func use unescaped queries and then join with &
 if using url.QueryEscape with filename ,the request can be successful, but there is document in s3 to show these value shoud be escaped
## Steps to Reproduce (for bugs)

1.generate url with minio-go
v := url.Values{}
v.Set("response-content-disposition", "file&name")
fmt.Println(minioClient.PresignedGetObject(context.Background(), "bucket", "obj", time.Hour, v))
2.get object with presigned url
3.minio return SignatureDoesNotMatch 

## Your Environment

* Version used (`minio --version`):RELEASE.2023-06-02T23-17-26Z (go1.19.9 linux/amd64)
* Server setup and configuration: docker run -p 9000:9000 -p 9001:9001   quay.io/minio/minio server /data --console-address ":9001"
* Operating System and version (`uname -a`):Linux DELL-PC 5.15.90.1-microsoft-standard-WSL2  x86_64 x86_64 x86_64 GNU/Linux

jiuker:
@TestMsr  I'm using the latest minio-go v7.0.56 and minio RELEASE.2023-06-02T23-17-26Z  it doesn't reproduce your issue. could you post the minimized reproduction code, I'll see if it's for other reasons. Thank you very much.

jiuker:
And here is the test code:
```go
package main

import (
	"context"
	"fmt"
	"log"
	"net/url"
	"time"

	"github.com/minio/minio-go/v7"
	"github.com/minio/minio-go/v7/pkg/credentials"
)

func main() {
	ctx := context.Background()
	endpoint := "127.0.0.1:9000"
	accessKeyID := "minioadmin"
	secretAccessKey := "minioadmin"
	useSSL := false

	// Initialize minio client object.
	minioClient, err := minio.New(endpoint, &minio.Options{
		Creds:  credentials.NewStaticV4(accessKeyID, secretAccessKey, ""),
		Secure: useSSL,
	})
	if err != nil {
		log.Fatalln(err)
	}

	// Make a new bucket called mymusic.
	bucketName := "mytest"
	v := url.Values{}
	v.Set("response-content-disposition", "file&name")
[...]

PR #17392, opened 2023-06-10
fix: preserve multiple values for query params

jiuker:
fix:set an error header when url.Value have multi value
related #17390 
## Description

When url.Value define multi values. Will set a wrong header here. 
make it will works for download 
```url
http://127.0.0.1:9000/mytest/R-C.jfif?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=minioadmin%2F20230610%2Fus-east-1%2Fs3%2Faws4_request&X-Amz-Date=20230610T135713Z&X-Amz-Expires=3600&X-Amz-SignedHeaders=host&response-content-disposition=file&response-content-disposition=name&X-Amz-Signature=9cd0138a9ec5632462835575db36f79750dea2e8bb6881f0764cadf05416ca59
```
if  not , the Browser will close the connection by unexpected.
I make it by
```go
        v := url.Values{}
	v["response-content-disposition"] = []string{"file", "name"}
	//v.Set("response-content-disposition", "file&name")
	fmt.Println(minioClient.PresignedGetObject(ctx, bucketName, "R-C.jfif", time.Hour, v))
```
the doc:
[content-disposition](https://developer.mozilla.org/zh-CN/docs/web/http/headers/content-disposition)
## Types of changes

## Checklist:


vadmeste (approved):
LGTM

Can you fix this PR title @jiuker ?

klauspost (approved):

cmd/object-handlers.go
```diff
@@ -90,7 +90,7 @@ const (
 func setHeadGetRespHeaders(w http.ResponseWriter, reqParams url.Values) {
 	for k, v := range reqParams {
 		if header, ok := supportedHeadGetReqParams[strings.ToLower(k)]; ok {
-			w.Header()[header] = v
+			w.Header()[header] = []string{strings.Join(v, ";")}
```
```suggestion
			w.Header()[header] = []string{strings.Join(v, ",")}
```

Multiple values in HTTP header must be `comma` separated

You'll want to take a look at the HTTP spec RFC 2616 where it says:

```
Multiple message-header fields with the same field-name MAY be present in a message if and only if the entire field-value for that header field is defined as a comma-separated list [i.e., #(values)]. It MUST be possible to combine the multiple header fields into one "field-name: field-value" pair, without changing the semantics of the message, by appending each subsequent field-value to the first, each separated by a comma. The order in which header fields with the same field-name are received is therefore significant to the interpretation of the combined field value, and thus a proxy MUST NOT change the order of these field values when a message is forwarded.
```

harshavardhana (changes requested):

harshavardhana (approved):

PR #17401, opened 2023-06-11
various compliance related fixes

harshavardhana:
## Description
various compliance-related fixes

## Motivation and Context
- getObjectTagging to be allowed for anonymous policies
- return correct errors for an invalid retention period
- return a sorted list of tags for an object
- putObjectTagging must return 200 OK not 204 OK

## How to test this PR?
Will add these as tests to our CI/CD but these are part
of S3 compliance tests that we need to address.

## Types of changes

## Checklist:


cmd/object-handlers.go
```diff
@@ -2808,20 +2810,52 @@ func (api objectAPIHandlers) GetObjectTaggingHandler(w http.ResponseWriter, r *h
...
+		}) {
+			writeErrorResponse(ctx, w, errorCodes.ToAPIErr(s3Error), r.URL)
+			return
+		}
+
+		ot, err = tags.ParseObjectTags(objInfo.UserTags)
+	} else {
+		// Get object tags
+		ot, err = objAPI.GetObjectTags(ctx, bucket, object, opts)
+	}
```
vadmeste: I am not sure I understand this else block, it looks like we still continue if there is any error in authenticating the request

vadmeste (changes requested):
One comment, LGTM otherwise

klauspost (approved):
LGTM

poornas (approved):

PR #17402, opened 2023-06-12
fix: set time format right

jiuker:
## Description

fix: set time format right

https://github.com/minio/minio/blob/be45ffd8a432b594ecb9a8733168aa6986212218/cmd/xl-storage-format-v2.go#L1298-L1307
I see the metadata use `time.RFC3339Nano`.
If not , cause some parse error maybe. 

## Types of changes

## Checklist:


cmd/erasure-metadata.go
```diff
@@ -626,10 +626,10 @@ func getInternalReplicationState(m map[string]string) ReplicationState {
 	for k, v := range m {
 		switch {
 		case equals(k, ReservedMetadataPrefixLower+ReplicationTimestamp):
-			tm, _ := time.Parse(http.TimeFormat, v)
+			tm, _ := time.Parse(time.RFC3339Nano, v)
```
klauspost: @poornas For backcompat - should we try both if one fails?

harshavardhana:
Go parses this fine what is bug you are fixing @jiuker ?

cmd/erasure-metadata.go
poornas: yes, that would be better

poornas:
there shouldn't be any parse issues - these timestamps are only used for some metadata overwrite decisions, and we don't really care about nano precision for that.

poornas (approved):

cmd/erasure-metadata.go
```diff
@@ -626,10 +626,16 @@ func getInternalReplicationState(m map[string]string) ReplicationState {
...
+			if tm.IsZero() {
+				tm, _ = time.Parse(http.TimeFormat, v)
+			}
```
harshavardhana: move this to `internal/amztime` package and call this `ParseReplicationTS(str string)` and parse using two different values return which ever succeeds.

harshavardhana (changes requested):

harshavardhana (approved):

klauspost (approved):
LGTM

PR #17406, opened 2023-06-13
Add an option to make bucket notifications synchronous

Praveenrajmani:
## Description

With the current asynchronous behavior in sending notification events to the targets, we can't provide guaranteed delivery as the systems might go for restarts.

For such event-driven use-cases, we can provide an option to enable synchronous events where the APIs wait until the event is successfully sent or persisted.

This commit adds 'MINIO_ENABLE_SYNC_EVENTS' env which when set to 'true' will enable sending/persisting events to targets synchronously.

This issue can be seen by simulating a slow queue_dir mountpoint with the following diff

```diff
 // Stats returns stats for targets.
diff --git a/internal/store/queuestore.go b/internal/store/queuestore.go
index 60e8a09fd..7fa65f79c 100644
--- a/internal/store/queuestore.go
+++ b/internal/store/queuestore.go
@@ -99,6 +99,7 @@ func (store *QueueStore[_]) Open() error {
 
 // write - writes an item to the directory.
 func (store *QueueStore[I]) write(key string, item I) error {
+       time.Sleep(1 * time.Second)
        // Marshalls the item.
        eventData, err := json.Marshal(item)
        if err != nil {
```

After applying the above diff, enable bucket notifications and restart the server while performing a bulk upload with the notification target down.

ie,

```bash
> mc mb myminio/testb
> mc admin config set myminio notify_webhook:1 enable=on endpoint=http://localhost:8080/ queue_dir=/tmp/webhookevents 
> mc admin service restart myminio
> mc event add myminio/testb arn:minio:sqs::1:webhook
> # Enable trace if needed to check the API trace
> # Now stop the webhook server and start uploading objects in parallel.
> for i in {1..300};do mc cp /home/praveen/Downloads/testobj.img myminio/testb/obj$i.jpg &; done
> # stop the server inbetween and check the queue_dir you will see some events missing.
```

this is because, the `sendEvent` is asynchronous - which will start a goroutine and sends the API response back. TERMINATING/INTERRUPTING the server will cancel all these go-routines which will lead to missing events.

## Motivation and Context

To add an option to make event subsystem synchronous so that we won't be missing any events in a non-resilient systems.

## How to test this PR?

STEP 1: Apply the above diff, set the env `export MINIO_API_SYNC_EVENTS=on` and start MinIO server (or) set the config by `mc admin config set myminio api sync_events=on`

STEP 2: Configure bucket notifications

```bash
> mc mb myminio/testb
> mc admin config set myminio notify_webhook:1 enable=on endpoint=http://localhost:8080/ queue_dir=/tmp/webhookevents 
> mc admin service restart myminio
> mc event add myminio/testb arn:minio:sqs::1:webhook
> # Enable trace if needed to check the API trace
> # Now stop the webhook server and start uploading objects in parallel.
```
STEP 3: Use the following minio-java snippet to upload objects in parallel. Also, enable the trace so that we can see the succeeded PUTs and failed PUTs.

```java
lass PutObjectRunnable implements Runnable {
  MinioClient client;
  String bucketName;
  String filename;
  String objectName;

  public PutObjectRunnable(String objectName) {
    this.client = MinioClient.builder()
    .endpoint("http://localhost:9000")
    .credentials("minio", "minio123")
    .build();
    this.bucketName = "testb";
    this.filename = "/etc/issue";
    this.objectName = objectName;
  }

  public void run() {
    try {
      client.uploadObject(
          UploadObjectArgs.builder()
              .bucket(bucketName)
              .object(objectName)
              .filename(filename)
              .build());
      System.out.println(objectName + " is uploaded successfully");
    } catch (Exception e) {
      System.out.println("error occured while uploading "+objectName+": "+e.toString());
      // e.printStackTrace();
    }
  }
}


public class PutObject {
  /** MinioClient.putObjectProgressBar() example. */
  public static void main(String[] args)
      throws IOException, NoSuchAlgorithmException, InvalidKeyException, MinioException, Exception {
    /* play.min.io for test and development. */
    int count = 300;
    Thread[] threads = new Thread[count];

    for (int i = 0; i < count; i++) {
      threads[i] = new Thread(new PutObjectRunnable("obj-"+i));
    }

    for (int i = 0; i < count; i++) {
      threads[i].start();
    }

    // Waiting for threads to complete.
    for (int i = 0; i < count; i++) {
      threads[i].join();
    }

    // All threads are completed.
    System.out.println("all threads completed");
  }
}
```

STEP 4: Stop the MinIO server inbetween the upload and check the queue dir - `ls -l /tmp/webhookevents/minio-webhook-1/ | wc -l` it should be equal or greater than succeeded responses seen in trace or in java console.

(NOTE: `mc ls` might show more objects as MinIO would've successfully written few objects)

## Types of changes

## Checklist:

### DIFF
--- a/cmd/event-notification.go
+++ b/cmd/event-notification.go
@@ -35,20 +35,18 @@ import (
 // EventNotifier - notifies external systems about events in MinIO.
 type EventNotifier struct {
 	sync.RWMutex
-	targetList                 *event.TargetList
-	targetResCh                chan event.TargetIDResult
-	bucketRulesMap             map[string]event.RulesMap
-	bucketRemoteTargetRulesMap map[string]map[event.TargetID]event.RulesMap
+	targetList     *event.TargetList
+	targetResCh    chan event.TargetIDResult
+	bucketRulesMap map[string]event.RulesMap
 }
 
 // NewEventNotifier - creates new event notification object.
 func NewEventNotifier() *EventNotifier {
 	// targetList/bucketRulesMap/bucketRemoteTargetRulesMap are populated by NotificationSys.InitBucketTargets()
 	return &EventNotifier{
-		targetList:                 event.NewTargetList(),
-		targetResCh:                make(chan event.TargetIDResult),
-		bucketRulesMap:             make(map[string]event.RulesMap),
-		bucketRemoteTargetRulesMap: make(map[string]map[event.TargetID]event.RulesMap),
+		targetList:     event.NewTargetList(),
+		targetResCh:    make(chan event.TargetIDResult),
+		bucketRulesMap: make(map[string]event.RulesMap),
 	}
 }
 
@@ -122,10 +120,6 @@ func (evnot *EventNotifier) AddRulesMap(bucketName string, rulesMap event.RulesM
 
 	rulesMap = rulesMap.Clone()
 
-	for _, targetRulesMap := range evnot.bucketRemoteTargetRulesMap[bucketName] {
-		rulesMap.Add(targetRulesMap)
-	}
-
 	// Do not add for an empty rulesMap.
 	if len(rulesMap) == 0 {
 		delete(evnot.bucketRulesMap, bucketName)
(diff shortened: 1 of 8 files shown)


cmd/event-notification.go
```diff
@@ -35,20 +35,18 @@ import (
 // EventNotifier - notifies external systems about events in MinIO.
 type EventNotifier struct {
 	sync.RWMutex
-	targetList                 *event.TargetList
-	targetResCh                chan event.TargetIDResult
-	bucketRulesMap             map[string]event.RulesMap
-	bucketRemoteTargetRulesMap map[string]map[event.TargetID]event.RulesMap
```
Praveenrajmani, line 41: Removed `bucketRemoteTargetRulesMap` as its stale and no longer used.

cmd/common-main.go
```diff
@@ -778,6 +778,8 @@ func handleCommonEnvVars() {
 	}
 
 	globalDisableFreezeOnBoot = env.Get("_MINIO_DISABLE_API_FREEZE_ON_BOOT", "") == "true" || serverDebugLog
+
+	globalEnableSyncEvents = env.Get("MINIO_ENABLE_SYNC_EVENTS", "") == "true"
```
harshavardhana: We should make this an option in the API subsystem?

MINIO_API_SYNC_EVENTS

harshavardhana (changes requested):

cmd/common-main.go
Praveenrajmani: sure, will do this change

harshavardhana (approved):

poornas (approved):

PR #17424, opened 2023-06-14
fallback on etags if they match when mtime is not same

harshavardhana:
## Description
fallback on etags if they match when mtime is not same

## Motivation and Context
on "unversioned" buckets there are situations
when successive concurrent I/O can lead to
an inconsistent state() with mtime while the
etag might be the same for the object on disk.
    
in such a scenario it is possible for us to
allow reading of the object since etag matches
and if etag matches we are guaranteed that we
have enough copies the object will be readable
and same.
    
This PR allows fallback in such scenarios.

## How to test this PR?
Requires a specific scenario, please ask me for the xl.meta

## Types of changes

## Checklist:

### DIFF
--- a/cmd/erasure-healing-common.go
+++ b/cmd/erasure-healing-common.go
@@ -25,6 +25,38 @@ import (
 	"github.com/minio/madmin-go/v2"
 )
 
+func commonETags(etags []string) (etag string, maxima int) {
+	etagOccurrenceMap := make(map[string]int, len(etags))
+
+	// Ignore the uuid sentinel and count the rest.
+	for _, etag := range etags {
+		if etag == "" {
+			continue
+		}
+		etagOccurrenceMap[etag]++
+	}
+
+	maxima = 0 // Counter for remembering max occurrence of elements.
+	latest := ""
+
+	// Find the common cardinality from previously collected
+	// occurrences of elements.
+	for etag, count := range etagOccurrenceMap {
+		if count < maxima {
+			continue
+		}
+
+		// We are at or above maxima
+		if count > maxima {
+			maxima = count
+			latest = etag
+		}
+	}
+
+	// Return the collected common max time, with maxima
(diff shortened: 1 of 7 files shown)


cmd/erasure-healing-common.go
```diff
@@ -102,6 +141,33 @@ func bootModtimes(diskCount int) []time.Time {
...
+	for _, count := range vidMap {
+		// do we have enough common versions
+		// that have enough quorum to satisfy
+		// the etag.
+		if count >= quorum {
+			return etags
+		}
+	}
+
+	return make([]string, len(partsMetadata))
```
vadmeste, line 168: this should be etags, not make()

cmd/erasure-object.go
```diff
@@ -692,12 +687,14 @@ func (er erasureObjects) getObjectFileInfo(ctx context.Context, bucket, object s
 			missingBlocks++
 			continue
 		}
-		if metaArr[i].IsValid() && metaArr[i].ModTime.Equal(fi.ModTime) {
-			continue
+		if etag == "" {
```
vadmeste: Isn't this a little bit downgrade change, before we can detect disks with outdated xl.meta, but now this only works when etag == ""

vadmeste (commented):
Few comments

cmd/erasure-healing-common.go
harshavardhana: This is not etags on purpose

cmd/erasure-object.go
harshavardhana: This will be empty always since modTime is not equal to zero values unless quorum can't be achieved

vadmeste (approved):
LGTM

harshavardhana:
PTAL @poornas

poornas (approved):

PR #17427, opened 2023-06-14
do not save plain-text ETag when encryption is requested

harshavardhana:
## Description
do not save plain-text ETag when encryption is requested

## Motivation and Context
fixes an issue under bucket replication could cause
ETags for replicated SSE-S3 single-part PUT objects,
to fail as we would attempt a decryption while listing,
or stat() operation.

## How to test this PR?
```bash
set -x

export CI=1

make || exit -1

killall -9 minio

rm -rf /tmp/xl/
mkdir -p /tmp/xl/1/ /tmp/xl/2/

export MINIO_KMS_SECRET_KEY="my-minio-key:OSMM+vkKUTCvQs9YL/CVMIMt43HFhkUpqJxTmGl6rYw="

NODES=4

args1=()
args2=()
for i in $(seq 1 $NODES); do
        args1+=("http://localhost:$((9000 + $i))/tmp/xl/1/$i ")
        args2+=("http://localhost:$((9100 + $i))/tmp/xl/2/$i ")
done

for i in $(seq 1 $NODES); do
        ./minio server --address "127.0.0.1:$((9000 + $i))" ${args1[@]} & # | tee /tmp/minio/node.$i &
        ./minio server --address "127.0.0.1:$((9100 + $i))" ${args2[@]} & # | tee /tmp/minio/node.$i &
done

sleep 10

./mc alias set myminio1 http://localhost:9001 minioadmin minioadmin
./mc alias set myminio2 http://localhost:9101 minioadmin minioadmin

sleep 1

./mc mb myminio1/testbucket/ --with-lock
./mc mb myminio2/testbucket/ --with-lock
./mc encrypt set sse-s3 my-minio-key myminio1/testbucket/
./mc encrypt set sse-s3 my-minio-key myminio2/testbucket/

./mc replicate add myminio1/testbucket --remote-bucket http://minioadmin:minioadmin@localhost:9101/testbucket --priority 1

sleep 1

./mc cp internal.tar myminio1/testbucket/dir/1.tar
./mc cp internal.tar myminio2/testbucket/dir/2.tar

sleep 1

./mc ls myminio1/testbucket/dir/

./mc ls myminio2/testbucket/dir/
```

## Types of changes

## Checklist:

### DIFF
--- a/cmd/erasure-object.go
+++ b/cmd/erasure-object.go
@@ -37,6 +37,7 @@ import (
 	"github.com/minio/minio/internal/bucket/lifecycle"
 	"github.com/minio/minio/internal/bucket/object/lock"
 	"github.com/minio/minio/internal/bucket/replication"
+	"github.com/minio/minio/internal/crypto"
 	"github.com/minio/minio/internal/event"
 	"github.com/minio/minio/internal/hash"
 	xhttp "github.com/minio/minio/internal/http"
@@ -1263,8 +1264,16 @@ func (er erasureObjects) putObject(ctx context.Context, bucket string, object st
 	}
 
 	userDefined["etag"] = r.MD5CurrentHexString()
+	kind, _ := crypto.IsEncrypted(userDefined)
 	if opts.PreserveETag != "" {
-		userDefined["etag"] = opts.PreserveETag
+		if !opts.ReplicationRequest {
+			userDefined["etag"] = opts.PreserveETag
+		} else if kind != crypto.S3 {
+			// if we have a replication request
+			// and SSE-S3 is specified do not preserve
+			// the incoming etag.
+			userDefined["etag"] = opts.PreserveETag
+		}
 	}
 
 	// Guess content-type from the extension if possible.
--- a/Makefile
+++ b/Makefile
@@ -74,6 +74,9 @@ test-iam: build ## verify IAM (external IDP, etcd backends)
 	@echo "Running tests for IAM (external IDP, etcd backends) with -race"
 	@MINIO_API_REQUESTS_MAX=10000 GORACE=history_size=7 CGO_ENABLED=1 go test -race -tags kqueue -v -run TestIAM* ./cmd
 
+test-sio-error:
+	@(env bash $(PWD)/docs/bucket/replication/sio-error.sh)
+
 test-replication-2site:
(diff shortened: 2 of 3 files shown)


cmd/erasure-object.go
```diff
@@ -1263,7 +1264,8 @@ func (er erasureObjects) putObject(ctx context.Context, bucket string, object st
 	}
 
 	userDefined["etag"] = r.MD5CurrentHexString()
-	if opts.PreserveETag != "" {
+	_, ok := crypto.IsEncrypted(userDefined)
```
vadmeste: I think S3.IsEncrypted() is better.

We can keep ETags for non encrypted, SSE-KMS, multipart but not SSE-S3

harshavardhana: multipart will never come here @vadmeste for preserveETag and in-case of SSE-KMS also we must save the longer encrypted version.

That's why this code says that it must be plain-text only to be able to preserve a custom etag remotely.

poornas (changes requested):
https://github.com/minio/minio/blob/master/cmd/object-handlers-common.go#L193 
etagMatch should be considered true if it is sse-s3 and etag is empty to avoid re-replicating if there are timing issues

cmd/erasure-object.go
```diff
@@ -1263,8 +1264,16 @@ func (er erasureObjects) putObject(ctx context.Context, bucket string, object st
 	}
 
 	userDefined["etag"] = r.MD5CurrentHexString()
+	kind, _ := crypto.IsEncrypted(userDefined)
 	if opts.PreserveETag != "" {
-		userDefined["etag"] = opts.PreserveETag
+		if !opts.ReplicationRequest {
+			userDefined["etag"] = opts.PreserveETag
```
vadmeste, line 1270: @poornas why would we want to preserve ETag in a PutObject request if this is not a replication request ? is this just a defensive code ?

harshavardhana: Decommissioning, rebalancing forgot? @vadmeste

vadmeste: oh right

vadmeste (approved):
LGTM

poornas (approved):

PR #17436, opened 2023-06-15
do not save any metadata from the headers in tar extract

### LINKED ISSUE #17434: Wrong content type with x-amz-meta-snowball-auto-extract
I have reasons to use the undocumented x-amz-meta-snowball-auto-extract functionality, which gives the ability of uploading many files in one request.

As I noted there are issues with the content type. If I upload a .js file as part of a tar-stream, I will get the content type "binary/octet-stream".
Otherwise it's "application/x-javascript".

I cannot find a solution to set the right content type. In my opinion it should be possible to configure a manual mapping.

harshavardhana:
## Description
do not save any metadata from the headers in tar extract

## Motivation and Context
only preserve the same storage-class as incoming
request other than that rest of them must be
deduced.

## How to test this PR?
fixes #17434 

## Types of changes

## Checklist:

### DIFF
--- a/cmd/object-handlers.go
+++ b/cmd/object-handlers.go
@@ -2111,36 +2111,16 @@ func (api objectAPIHandlers) PutObjectExtractHandler(w http.ResponseWriter, r *h
 		getObjectInfo = api.CacheAPI().GetObjectInfo
 	}
 
-	// Extract request metadata
-	metadata, err := extractMetadata(ctx, r)
-	if err != nil {
-		writeErrorResponse(ctx, w, toAPIError(ctx, err), r.URL)
-		return
-	}
-
-	// Remove the snowball headers from metadata from objects
-	// since those are actionable headers and not for storage
-	delete(metadata, xhttp.AmzSnowballExtract)
-	delete(metadata, xhttp.AmzSnowballExtract)
-	delete(metadata, xhttp.MinIOSnowballIgnoreDirs)
-	delete(metadata, xhttp.MinIOSnowballIgnoreErrors)
-	delete(metadata, xhttp.MinIOSnowballPrefix)
-
-	// Use the same storage class as the original uploaded tar file
-	if sc != "" {
-		metadata[xhttp.AmzStorageClass] = sc
-	}
-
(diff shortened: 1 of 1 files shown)


cmd/object-handlers.go
```diff
@@ -2111,36 +2111,12 @@ func (api objectAPIHandlers) PutObjectExtractHandler(w http.ResponseWriter, r *h
...
-
 	putObjectTar := func(reader io.Reader, info os.FileInfo, object string) error {
 		size := info.Size()
 
-		// Copy metadata and make space for a bit more
-		metaCopy := make(map[string]string, len(metadata)+5)
-		for k, v := range metadata {
-			metaCopy[k] = v
+		metadata := map[string]string{
+			xhttp.AmzStorageClass: sc, // save same storage-class as incoming stream.
```
vadmeste, line 2122: 'sc' in this function can be empty, but usually there is no storage-class set in xl.meta when the user does not specify anything. 'mc stat' will show empty storage class if you set an empty sc here

cmd/object-handlers.go
```diff
@@ -2111,36 +2111,12 @@ func (api objectAPIHandlers) PutObjectExtractHandler(w http.ResponseWriter, r *h
...
-	delete(metadata, xhttp.AmzSnowballExtract)
-	delete(metadata, xhttp.AmzSnowballExtract)
-	delete(metadata, xhttp.MinIOSnowballIgnoreDirs)
-	delete(metadata, xhttp.MinIOSnowballIgnoreErrors)
-	delete(metadata, xhttp.MinIOSnowballPrefix)
-
-	// Use the same storage class as the original uploaded tar file
-	if sc != "" {
-		metadata[xhttp.AmzStorageClass] = sc
-	}
```
poornas, line 2132: shouldn't we hold on to user defined metadata?

harshavardhana: no for tar extract why? - do all objects have same metadata?

poornas (approved):

donatello (approved):

vadmeste (approved):
LGTM

PR #17454, opened 2023-06-18
allow quorum fileInfo to pick same parityBlocks

harshavardhana:
## Description
allow quorum fileInfo to pick same parityBlocks

## Motivation and Context
Bonus: allow replication to proceed without performing 
a HEAD (statObject) first, there is no value in doing 
that instead we should allow PUT to proceed and let it 
fail with pre-condition check.

## How to test this PR?
Requires a special setup, please ping if you want
to test this. 

## Types of changes

## Checklist:


cmd/bucket-replication.go
```diff
@@ -1349,53 +1349,6 @@ func (ri ReplicateObjectInfo) replicateAll(ctx context.Context, objectAPI Object
...
-		logger.LogIf(ctx, fmt.Errorf("unable to replicate %s/%s (%s). Target returned %s error on HEAD", bucket, object, objInfo.VersionID, cerr))
-		sendEvent(eventArgs{
-			EventName:  event.ObjectReplicationNotTracked,
-			BucketName: bucket,
-			Object:     objInfo,
-			UserAgent:  "Internal: [Replication]",
-			Host:       globalLocalNodeName,
-		})
-		return
-	}
```
poornas, line 1398: if the HEAD is removed, every metadata update is a full overwrite - this is not ok. We already optimize to do a HEAD only for replication failure healing, not for inline replication

poornas (changes requested):

cmd/bucket-replication.go
```diff
@@ -1385,16 +1385,23 @@ func (ri ReplicateObjectInfo) replicateAll(ctx context.Context, objectAPI Object
...
-			BucketName: bucket,
-			Object:     objInfo,
-			UserAgent:  "Internal: [Replication]",
-			Host:       globalLocalNodeName,
-		})
-		return
+	if cerr != nil {
+		errResp := minio.ToErrorResponse(cerr)
+		switch errResp.Code {
+		case "NoSuchKey", "NoSuchVersion", "SlowDown":
```
poornas: why try to replicate when other cluster is asking to slow down - we can just leave it for catch up?

cmd/bucket-replication.go
```diff
@@ -1385,16 +1385,23 @@ func (ri ReplicateObjectInfo) replicateAll(ctx context.Context, objectAPI Object
...
-			BucketName: bucket,
-			Object:     objInfo,
-			UserAgent:  "Internal: [Replication]",
-			Host:       globalLocalNodeName,
-		})
-		return
+	if cerr != nil {
+		errResp := minio.ToErrorResponse(cerr)
+		switch errResp.Code {
+		case "NoSuchKey", "NoSuchVersion", "SlowDownRead":
```
harshavardhana, line 1391: It must be a quorum read issue, if that is the case we are better off replicating the new state() if possible @poornas - we shouldn't ignore such situations.

harshavardhana: we should use replication as a "healing" mechanism here to make sure both sites are always in sync.

cmd/erasure-object.go
```diff
@@ -1007,14 +1010,23 @@ func (er erasureObjects) putObject(ctx context.Context, bucket string, object st
...
+	if opts.CheckPrecondFn != nil && !opts.NoLock {
+		ns := er.NewNSLock(bucket, object)
+		lkctx, err := ns.GetLock(ctx, globalOperationTimeout)
+		if err != nil {
+			return ObjectInfo{}, err
 		}
-		if opts.CheckPrecondFn(obj) {
+		ctx = lkctx.Context()
+		defer ns.Unlock(lkctx)
+		opts.NoLock = true
```
poornas: above will not enforce precondition if opts.NoLock is true. 

```
if opts.CheckPrecondFn != nil {
      if !opts.NoLock {
		ns := er.NewNSLock(bucket, object)
		lkctx, err := ns.GetLock(ctx, globalOperationTimeout)
		..
```

poornas (approved):
LGTM

vadmeste (approved):
LGTM but lint is failing

PR #17467, opened 2023-06-19
s3: Make/Delete buckets to use error quorum per pool

vadmeste:
## Description
Creating and delete buckets should calculate action quorum based on pools
e.g.: it should not error out when few nodes (< quorum are offline

## Motivation and Context
Fix an issue when trying to delete a bucket when one node or less than quorum nodes are offline

## How to test this PR?
Start a MinIO cluster (4 nodes, 1 disk/node)
Kill a node and create a bucket

## Types of changes

## Checklist:

### DIFF
--- a/cmd/globals.go
+++ b/cmd/globals.go
@@ -224,9 +224,9 @@ var (
 	// registered listeners
 	globalConsoleSys *HTTPConsoleLoggerSys
 
-	// All drives endpoints of this deployment
+	// All unique drives for this deployment
 	globalEndpoints EndpointServerPools
-	// All nodesof this deployment
+	// All unique nodes for this deployment
 	globalNodes []Node
 
 	// The name of this local node, fetched from arguments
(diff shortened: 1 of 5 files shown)


harshavardhana:
this is unnecessary why are we doing this? @vadmeste

cmd/globals.go
```diff
@@ -224,7 +224,10 @@ var (
 	// registered listeners
 	globalConsoleSys *HTTPConsoleLoggerSys
 
+	// All drives endpoints of this deployment
 	globalEndpoints EndpointServerPools
+	// All nodesof this deployment
```
harshavardhana:
```suggestion
	// All unique nodes for this deployment
```

cmd/globals.go
```diff
@@ -224,7 +224,10 @@ var (
 	// registered listeners
 	globalConsoleSys *HTTPConsoleLoggerSys
 
+	// All drives endpoints of this deployment
```
harshavardhana:
```suggestion
	// All unique drives for this deployment
```

harshavardhana (changes requested):

harshavardhana (approved):

poornas (approved):

PR #17479, opened 2023-06-21
always look for expired versions v/s remaining versions

harshavardhana:
## Description
always look for expired versions v/s remaining versions

## Motivation and Context
while decommissioning it can so happen that the non-current
versions are all expired but there is a DEL marker as the
latest version.
    
For such objects, we should not decommission them instead
calculate the remaining versions and if the remaining versions
is one and that version is a DEL marker consider such
an object not to be scheduled for decommissioning.

## How to test this PR?
You would need a bucket with expired non-current versions
and a DEL marker as the latest version for an object.

Once you have that you can attempt the decommission
the pool where this object and its versions reside.

Decommissioning will fail with 
```
API: SYSTEM()
Time: 23:45:44 UTC 06/20/2023
DeploymentID: ebc4311e-bc9d-486b-9f09-0f5cdedb4278
Error: Object not found: testbucket/hosts (cmd.ObjectNotFound)
       2: internal/logger/logger.go:258:logger.LogIf()
       1: cmd/erasure-server-pool-decom.go:765:cmd.(*erasureServerPools).decommissionPool.func2()
```

Make sure to revert PR #17473 to see this error, this happens
because single DEL marker without a prior `xl.meta` on the disk
is not supported which would end up returning ObjectNotFound.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/erasure-server-pool-decom.go
+++ b/cmd/erasure-server-pool-decom.go
@@ -419,7 +419,9 @@ func (p poolMeta) save(ctx context.Context, pools []*erasureSets) error {
 	// Saves on all pools to make sure decommissioning of first pool is allowed.
 	for i, eset := range pools {
 		if err = saveConfig(ctx, eset, poolMetaName, buf); err != nil {
-			logger.LogIf(ctx, fmt.Errorf("saving pool.bin for pool index %d failed with: %v", i, err))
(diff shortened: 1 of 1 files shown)


vadmeste (approved):
LGTM

poornas (approved):

PR #17426, opened 2023-06-14
Fix locking in policy attach API

donatello:
## Description

For policy attach/detach API to work correctly the server should hold a lock before reading existing policy mapping and until after writing the updated policy mapping. This is fixed in this change.

Bonus: Additionally, the server responds with the actual (or net) changes performed in the attach/detach API call. For e.g. if a user already has policy A applied, and a call to attach policies A and B is performed, the server will respond that B was attached successfully.

## How to test this PR?

Requires mc from https://github.com/minio/mc/pull/4602

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Optimization (provides speedup with no functional changes)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [ ] Fixes a regression (If yes, please add `commit-id` or `PR #` here)
- [ ] Unit tests added/updated
- [ ] Internal documentation updated
- [ ] Create a documentation update request [here](https://github.com/minio/docs/issues/new?label=doc-change,title=Doc+Updated+Needed+For+PR+github.com%2fminio%2fminio%2fpull%2fNNNNN)

PR #17483, opened 2023-06-21
Return additional user metadata in GetUserInfo API

donatello:
## Description

This change stores additional metadata about the source/authentication
of a user and returns this info for builtin IDP and LDAP users in the
GetUserInfo API. This metadata will be used in other informational
user/group related APIs and track each credential stored by the server.

~This needs https://github.com/minio/minio/pull/17426 to be merged first as it is developed on top of that change.~

## Motivation and Context

Motivation here is to make credentials (accounts, STS creds, service accounts) more trackable and to expose the owner of each access key in user info APIs. In this PR, builtin IDP users and LDAP users are clearly identified.

## How to test this PR?

Needs https://github.com/minio/madmin-go/pull/201 and https://github.com/minio/mc/pull/4613

See the mc PR for example outputs with this change.

## Types of changes

## Checklist:


donatello:
This is ready for review

cmd/iam-store.go
```diff
@@ -1,3 +1,5 @@
+//go:generate stringer -type=EntityInfoType,ESType -linecomment
+//
 // Copyright (c) 2015-2021 MinIO, Inc.
```
harshavardhana:
```suggestion
// Copyright (c) 2015-2023 MinIO, Inc.
```

harshavardhana (approved):

donatello:
PTAL @poornas @vadmeste This is ready - the corresponding PR for mc is also approved.

cmd/site-replication.go
```diff
@@ -1430,7 +1430,9 @@ func (c *SiteReplicationSys) PeerSTSAccHandler(ctx context.Context, stsCred *mad
 	}
 
 	// Set these credentials to IAM.
-	if _, err := globalIAMSys.SetTempUser(ctx, cred.AccessKey, cred, stsCred.ParentPolicyMapping); err != nil {
+	//
+	// FIXME(aditya): Currently EntityInfo is not replicated.
```
harshavardhana: isn't this a problem? @donatello

harshavardhana (changes requested):

cmd/site-replication.go
donatello: This PR is only exposing `EntityInfo` information for internal IDP and LDAP user listing and the info for this is replicated.

For LDAP, OpenID, MTLS, CustomToken based STS creds, the info is not yet replicated, but the info is not yet exposed in admin APIs - it will be exposed by updating list-access-keys (currently list-service-accounts) APIs to include STS accounts. The replication of this info is not implemented in this PR, because we need to update SR to replicate the `UserIdentity` type rather than the `Credentials` type and that is a more involved change.

The only user visible impact of not replicating this info is that when the STS account info is exposed in admin APIs, some STS accounts would have the info in Site A (where the cred is created) but not in Site B (a replication destination). However, I will fix the replication bits before we expose the STS account info, so we will have the behavior that (after that version is released), any STS accounts created post updating to that release, will have `EntityInfo` replicated.

If we really cant have this temporary inconsistency in replication, I can update this PR to not save `EntityInfo` for STS accounts in this change itself.

donatello: This https://github.com/minio/minio/pull/17589 is towards fixing site replication so we can replicate STS better with the EntityInfo (in a subsequent change)

donatello: We should take https://github.com/minio/minio/pull/17589 first - I will rework this change.

harshavardhana:
Is this PR ready to be merged? @donatello or is this not needed anymore?

harshavardhana:
Closing this PR @donatello please resubmit after rebasing.

PR #17486, opened 2023-06-22
[helm] add ability to mount a custom volume to the export directory

ricsam:
## Description
This PR adds 1 parameter to the helm values file allowing the user to choose which volume to mount on the exports folder.

## Motivation and Context
I would like to change where the export folder is mounted in the container using the helm chart

## How to test this PR?
You can add the following to the values.yaml file when you deploy the helm chart:
```yml
exportVolumeName: minio-host-path
extraVolumes:
  - name: minio-host-path
    hostPath:
      # directory location on host
      path: /minio-host-path
      type: Directory
```

## Types of changes

## Checklist:

### DIFF
--- a/helm/minio/templates/deployment.yaml
+++ b/helm/minio/templates/deployment.yaml
@@ -77,7 +77,7 @@ spec:
             - name: minio-user
               mountPath: "/tmp/credentials"
               readOnly: true
-            - name: export
+            - name: {{.Values.exportVolumeName | default "export" }}
               mountPath: {{ .Values.mountPath }}
               {{- if and .Values.persistence.enabled .Values.persistence.subPath }}
               subPath: "{{ .Values.persistence.subPath }}"
--- a/helm/minio/values.yaml
+++ b/helm/minio/values.yaml
@@ -49,6 +49,9 @@ extraArgs: []
 ## Additional volumes to minio container
 extraVolumes: []
 
+## If you add additional volumes and you want to mount one of the those volumes to the export directory then update the exportVolumeName to match the custom volume
+exportVolumeName: export
+
 ## Additional volumeMounts to minio container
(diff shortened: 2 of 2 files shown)


helm/minio/templates/deployment.yaml
```diff
@@ -77,7 +77,7 @@ spec:
             - name: minio-user
               mountPath: "/tmp/credentials"
               readOnly: true
-            - name: export
+            - name: {{.Values.exportVolumeName | default "export" }}
```
jiuker, line 80: The volumes still named `export`, that should be the same.

jiuker (changes requested):

jiuker:
Then what is the meaning of this pr, I don't think the user should set the name of this volume, if there is a conflict in the volume name setting, it will definitely not start. I also don't see this name together with the volume configuration, which is the usual practice. @ricsam

jiuker:
And this is related to minio-operator. We will not take this.  @harshavardhana

harshavardhana:
Customizing export volume can conflict, here so configuring this is not allowed in operator - can you explain what were you trying to do @ricsam ?

ricsam:
@jiuker @harshavardhana Thanks for input. My goal is to mount the `{{ .Values.mountPath }}` directory to a host path volume.

The "export" pvc/emptyDir volume will be mounted on the `{{ .Values.mountPath }}` path. I don't want that. I want another volume.

There is still a volume named export created, but that doesn't matter much (for me at least). It is just a dangling pvc/emptyPath volume.

jiuker:
You should create the pv with hostPath, then create the pvc to use that pv, then fill in the name of the pvc to .Values.persistence.existingClaim. Your pr is not the usual practice for helm. @ricsam

PR #17498, opened 2023-06-23
fix: bug in passing Versioned field set for getHealReplicationInfo()

harshavardhana:
## Description
fix: bug in passing Versioned field set for getHealReplicationInfo()

## Motivation and Context
Bonus:

- rejects prefix deletes on object-locked buckets earlier

## How to test this PR?
CI/CD should cover all the tests

## Types of changes

## Checklist:

### DIFF
--- a/cmd/object-handlers.go
+++ b/cmd/object-handlers.go
@@ -2298,6 +2298,14 @@ func (api objectAPIHandlers) DeleteObjectHandler(w http.ResponseWriter, r *http.
 		return
 	}
 
+	replica := r.Header.Get(xhttp.AmzBucketReplicationStatus) == replication.Replica.String()
+	if replica {
+		if s3Error := checkRequestAuthType(ctx, r, policy.ReplicateDeleteAction, bucket, object); s3Error != ErrNone {
+			writeErrorResponse(ctx, w, errorCodes.ToAPIErr(s3Error), r.URL)
+			return
+		}
+	}
+
 	if globalDNSConfig != nil {
 		_, err := globalDNSConfig.Get(bucket)
 		if err != nil && err != dns.ErrNotImplemented {
@@ -2311,10 +2319,12 @@ func (api objectAPIHandlers) DeleteObjectHandler(w http.ResponseWriter, r *http.
 		writeErrorResponse(ctx, w, toAPIError(ctx, err), r.URL)
 		return
 	}
-	var (
-		goi  ObjectInfo
-		gerr error
-	)
+
+	rcfg, _ := globalBucketObjectLockSys.Get(bucket)
+	if rcfg.LockEnabled && opts.DeletePrefix {
+		writeErrorResponse(ctx, w, toAPIError(ctx, errors.New("force-delete is forbidden in a locked-enabled bucket")), r.URL)
+		return
+	}
 
 	getObjectInfo := objectAPI.GetObjectInfo
 	if api.CacheAPI() != nil {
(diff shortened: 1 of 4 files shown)


cmd/object-handlers.go
```diff
@@ -2293,7 +2293,12 @@ func (api objectAPIHandlers) DeleteObjectHandler(w http.ResponseWriter, r *http.
 		return
 	}
 
-	if s3Error := checkRequestAuthType(ctx, r, policy.DeleteObjectAction, bucket, object); s3Error != ErrNone {
+	replica := r.Header.Get(xhttp.AmzBucketReplicationStatus) == replication.Replica.String()
+	action := policy.DeleteObjectAction
+	if replica {
+		action = policy.ReplicateDeleteAction
+	}
+	if s3Error := checkRequestAuthType(ctx, r, policy.Action(action), bucket, object); s3Error != ErrNone {
```
klauspost: AFAICT nothing prevents someone from just setting this property and change the check?

It seems to me we really needs some documentation to make sure that replica users are never mixed with regular users.

harshavardhana: This change is just relaxing the policy requirement, currently it requires both DeleteObject() and ReplicateDelete()

For replicated request we only need ReplicateDelete() that's all

cmd/object-handlers.go
```diff
@@ -2338,17 +2345,10 @@ func (api objectAPIHandlers) DeleteObjectHandler(w http.ResponseWriter, r *http.
...
 	vID := opts.VersionID
-	if r.Header.Get(xhttp.AmzBucketReplicationStatus) == replication.Replica.String() {
-		// check if replica has permission to be deleted.
-		if apiErrCode := checkRequestAuthType(ctx, r, policy.ReplicateDeleteAction, bucket, object); apiErrCode != ErrNone {
-			writeErrorResponse(ctx, w, errorCodes.ToAPIErr(apiErrCode), r.URL)
-			return
-		}
+	if dsc.ReplicateAny() {
+		opts.SetDeleteReplicationState(dsc, opts.VersionID)
 		opts.SetReplicaStatus(replication.Replica)
```
poornas, line 2357: replica status should not be set on the replicating side. Should be set only if replica bool is set to true

poornas (changes requested):

cmd/object-handlers.go
```diff
@@ -2338,42 +2345,31 @@ func (api objectAPIHandlers) DeleteObjectHandler(w http.ResponseWriter, r *http.
...
-		opts.SetReplicaStatus(replication.Replica)
-		if opts.VersionPurgeStatus().Empty() {
-			// opts.VersionID holds delete marker version ID to replicate and not yet present on disk
-			vID = ""
+	if dsc.ReplicateAny() {
+		opts.SetDeleteReplicationState(dsc, opts.VersionID)
+		if replica {
+			opts.SetReplicaStatus(replication.Replica)
+			if opts.VersionPurgeStatus().Empty() {
+				// opts.VersionID holds delete marker version ID to replicate and not yet present on disk
```
poornas: this should be outside the dsc.ReplicateAny() - on replica side, replication decision will be false

cmd/object-handlers.go
klauspost: Yeah, but my point is that you can change the permission required from `DeleteObjectAction` to `ReplicateDeleteAction` by setting a header in the request.

If users are somehow given `ReplicateDeleteAction`, a hostile user could delete objects without `DeleteObjectAction` permission.

harshavardhana: if that confusion can arise, I can simply remove this.

harshavardhana:
I removed the permissions back to what it was to avoid any confusion. We may have to add replica-request as part of the signed headers to avoid spoofing.

harshavardhana:
PTAL @poornas @vadmeste @klauspost

klauspost (approved):
LGTM

harshavardhana:
PTAL @poornas

poornas (approved):

PR #17499, opened 2023-06-23
fix: do not listAndHeal() inline with PutObject()

harshavardhana:
## Description
fix: do not listAndHeal() inline with PutObject()

## Motivation and Context
there is a possibility that slow drives can actually add latency
to the overall call, leading to a large spike in latency.

this can happen if there are other parallel listObjects()
calls to the same drive, in turn causing each other to sort 
of serialize.

this potentially improves performance and makes PutObject()
also non-blocking.

## How to test this PR?
CI/CD

## Types of changes

## Checklist:

### DIFF
--- a/cmd/erasure-healing.go
+++ b/cmd/erasure-healing.go
@@ -23,6 +23,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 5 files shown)


poornas (approved):

cmd/erasure-healing.go
```diff
@@ -66,6 +67,68 @@ func (fi FileInfo) DataShardFixed() bool {
...
+				entry, _ = entries.firstFound()
+			}
+
+			if err := healEntry(bucket, *entry); err != nil {
+				logger.LogIf(ctx, err)
+				cancel()
+				return
+			}
+		},
+		finished: nil,
```
jiuker, line 122: Why don't we output a little log here?

jiuker (approved):

cmd/erasure-healing.go
harshavardhana: Not very useful, that's why it's nil and can be a lot of logs which is not needed.

PR #17518, opened 2023-06-26
CI/CD to stop one node per pool in the two pools mint test

vadmeste:
## Description
This is to make sure that all S3 ops work when there is enough quorum

## Motivation and Context
More tests

## Types of changes

## Checklist:


harshavardhana (approved):

harshavardhana:
As discussed we need to see why the total runtime of the jobs increased from 25-30mins to 50+mins

vadmeste:
> As discussed we need to see why the total runtime of the jobs increased from 25-30mins to 50+mins

One thing I found out is that creating a bucket takes 2 seconds, since three months.. and mint creates a lot of buckets. The reason of the bucket creation slowdown is that the code is retrying LoadBucketMetadata() and do one second sleep in PutBucketHandler().. but I think there are other reasons for the slowdown of mint test here as well

PR #17537, opened 2023-06-29
Using Alpine as based system in Docker Image

galek:
## Description
Hello.  This is an implementation of docker image what based on alpine systems.

## Motivation and Context
Sometimes we have situations we can't use RHEL systems as base system (the legal limits). Using Debian/Ubuntu-based systems is not a good idea for use as base system because image size is longer if compare with alpine based image. 

## How to test this PR?
Just deploy it with your env file.

## Types of changes

## Checklist:
- [] Create a documentation update request [here](https://github.com/minio/docs/issues/new?label=doc-change,title=Doc+Updated+Needed+For+PR+github.com%2fminio%2fminio%2fpull%2fNNNNN)

### DIFF
--- a/Dockerfile.alpine
+++ b/Dockerfile.alpine
@@ -3,31 +3,31 @@ FROM alpine:3.18.2
 RUN apk update && apk upgrade
 RUN apk add su-exec minisign bash bash-completion curl --no-cache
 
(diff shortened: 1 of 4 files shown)


harshavardhana:
Thanks but we are not planning to keep alpine images in our repo.

PR #17539, opened 2023-06-29
Add active disk health checks

klauspost:
## Description

Add check every 2 minutes to see if a write+read operation can complete.

If disk doesn't complete a read+write within 2 minutes or returns errFaultyDisk, take it offline.

## Motivation and Context

Disks are not taken offline until all 512 slots have been exhausted. For a low-traffic site this may not trigger for quite a while.

This should be conservative enough that it shouldn't be triggered unless drive is either faulty or extremely overloaded - in both cases taking the drive offline makes sense.

## How to test this PR?

Requires flaky drives or similar.

## Types of changes

### DIFF
--- a/cmd/xl-storage-disk-id-check.go
+++ b/cmd/xl-storage-disk-id-check.go
@@ -828,19 +828,30 @@ func (p *xlStorageDiskIDCheck) monitorDiskStatus() {
 // monitorDiskStatus should be called once when a drive has been marked offline.
 // Once the disk has been deemed ok, it will return to online status.
 func (p *xlStorageDiskIDCheck) monitorDiskWritable() {
-	// We check every 2 minutes if the disk is writable and we can read back.
-	t := time.NewTicker(2 * time.Minute)
+	const (
+		// We check every 15 seconds if the disk is writable and we can read back.
+		checkEvery = 15 * time.Second
+
+		// Disk has 2 minutes to complete write+read.
+		timeoutOperation = 2 * time.Minute
+
+		// If the disk has completed an operation successfully within last 5 seconds, don't check it.
+		skipIfSuccessBefore = 5 * time.Second
+	)
+
+	t := time.NewTicker(checkEvery)
 	defer t.Stop()
 	fn := mustGetUUID()
 
 	// Be just above directio size.
 	toWrite := []byte{xioutil.DirectioAlignSize + 1: 42}
 	rng := rand.New(rand.NewSource(time.Now().UnixNano()))
+
 	for range t.C {
 		if atomic.LoadInt32(&p.health.status) != diskHealthOK {
 			continue
 		}
-		if time.Since(time.Unix(0, atomic.LoadInt64(&p.health.lastSuccess))) < 5*time.Second {
+		if time.Since(time.Unix(0, atomic.LoadInt64(&p.health.lastSuccess))) < skipIfSuccessBefore {
 			// We recently saw a success - no need to check.
 			continue
 		}
@@ -855,13 +866,17 @@ func (p *xlStorageDiskIDCheck) monitorDiskWritable() {
 		done := make(chan struct{})
 		started := time.Now()
 		go func() {
+			timeout := time.NewTimer(timeoutOperation)
 			select {
 			// Reuse the same trigger
 			// If it triggers again, it took too long.
-			case <-t.C:
+			case <-timeout.C:
 				spent := time.Since(started)
 				goOffline(fmt.Errorf("unable to write+read for %v", spent.Round(time.Millisecond)))
 			case <-done:
+				if !timeout.Stop() {
+					<-timeout.C
+				}
 			}
 		}()
 		func() {
--- a/cmd/xl-storage-disk-id-check.go
+++ b/cmd/xl-storage-disk-id-check.go
@@ -840,6 +840,10 @@ func (p *xlStorageDiskIDCheck) monitorDiskWritable() {
 		if atomic.LoadInt32(&p.health.status) != diskHealthOK {
 			continue
 		}
+		if time.Since(time.Unix(0, atomic.LoadInt64(&p.health.lastSuccess))) < 5*time.Second {
+			// We recently saw a success - no need to check.
+			continue
+		}
 		goOffline := func(err error) {
 			if atomic.CompareAndSwapInt32(&p.health.status, diskHealthOK, diskHealthFaulty) {
 				logger.LogAlwaysIf(context.Background(), fmt.Errorf("node(%s): taking drive %s offline: %v", globalLocalNodeName, p.storage.String(), err))
--- a/cmd/xl-storage-disk-id-check.go
+++ b/cmd/xl-storage-disk-id-check.go
@@ -83,6 +83,8 @@ type xlStorageDiskIDCheck struct {
 	storage      *xlStorage
 	health       *diskHealthTracker
 	metricsCache timedValue
+	diskCtx      context.Context
+	cancel       context.CancelFunc
 }
 
 func (p *xlStorageDiskIDCheck) getMetrics() DiskMetrics {
@@ -136,11 +138,12 @@ func newXLStorageDiskIDCheck(storage *xlStorage, healthCheck bool) *xlStorageDis
 		storage: storage,
 		health:  newDiskHealthTracker(),
 	}
+	xl.diskCtx, xl.cancel = context.WithCancel(context.TODO())
 	for i := range xl.apiLatencies[:] {
 		xl.apiLatencies[i] = &lockedLastMinuteLatency{}
 	}
 	if healthCheck {
-		go xl.monitorDiskWritable()
+		go xl.monitorDiskWritable(xl.diskCtx)
 	}
 	return &xl
 }
@@ -199,6 +202,7 @@ func (p *xlStorageDiskIDCheck) SetDiskLoc(poolIdx, setIdx, diskIdx int) {
 }
 
 func (p *xlStorageDiskIDCheck) Close() error {
+	p.cancel()
 	return p.storage.Close()
 }
 
@@ -827,7 +831,7 @@ func (p *xlStorageDiskIDCheck) monitorDiskStatus() {
 
 // monitorDiskStatus should be called once when a drive has been marked offline.
 // Once the disk has been deemed ok, it will return to online status.
-func (p *xlStorageDiskIDCheck) monitorDiskWritable() {
+func (p *xlStorageDiskIDCheck) monitorDiskWritable(ctx context.Context) {
 	const (
 		// We check every 15 seconds if the disk is writable and we can read back.
 		checkEvery = 15 * time.Second
@@ -848,6 +852,9 @@ func (p *xlStorageDiskIDCheck) monitorDiskWritable() {
 	rng := rand.New(rand.NewSource(time.Now().UnixNano()))
 
 	for range t.C {
+		if contextCanceled(ctx) {
+			break
+		}
 		if atomic.LoadInt32(&p.health.status) != diskHealthOK {
 			continue
 		}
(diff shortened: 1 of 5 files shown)


klauspost:
I can't see anything that would cause these failures.

harshavardhana:
> I can't see anything that would cause these failures.

I think the `go` per drive is causing it to go into this loop ? . It seems persistent. Can't see it in other PRs.

cmd/xl-storage-disk-id-check.go
```diff
@@ -821,6 +825,61 @@ func (p *xlStorageDiskIDCheck) monitorDiskStatus() {
...
+	toWrite := []byte{xioutil.DirectioAlignSize + 1: 42}
+	rng := rand.New(rand.NewSource(time.Now().UnixNano()))
+	for range t.C {
+		if atomic.LoadInt32(&p.health.status) != diskHealthOK {
+			continue
+		}
+		goOffline := func(err error) {
+			if atomic.CompareAndSwapInt32(&p.health.status, diskHealthOK, diskHealthFaulty) {
+				logger.LogAlwaysIf(context.Background(), fmt.Errorf("node(%s): taking drive %s offline: %v", globalLocalNodeName, p.storage.String(), err))
+				go p.monitorDiskStatus()
```
harshavardhana, line 868: I think this is causing some kind of stack overflow on windows @klauspost

cmd/xl-storage-disk-id-check.go
```diff
@@ -821,6 +825,61 @@ func (p *xlStorageDiskIDCheck) monitorDiskStatus() {
...
+// monitorDiskStatus should be called once when a drive has been marked offline.
+// Once the disk has been deemed ok, it will return to online status.
+func (p *xlStorageDiskIDCheck) monitorDiskWritable() {
+	// We check every 2 minutes if the disk is writable and we can read back.
+	t := time.NewTicker(2 * time.Minute)
+	defer t.Stop()
+	fn := mustGetUUID()
+
+	// Be just above directio size.
+	toWrite := []byte{xioutil.DirectioAlignSize + 1: 42}
```
harshavardhana, line 851: do we not need some data to write, are we just writing zeroes?

harshavardhana (changes requested):
IMO we should also keep track of the open fd's during 'writes' and close them actively when we take the drive offline.

I am doing a similar change but I can simply borrow this and build on top.

harshavardhana:
Another thing is if we can add this why do we need to monitor 512 calls per drive to take it offline?

We can quite simply just do this I/O to take drive offline if its not writable or unresponsive. This satisfies all use-cases - in-fact we can reduce this to 5secs even 120secs is too high. 

Adding to this if we keep an open-fd map for reads and writes on the actual caller side. We can simply take all those fd's offline by closing them forcibly.

WDYT?

cmd/xl-storage-disk-id-check.go
klauspost: Nah `atomic.CompareAndSwapInt32` will only trigger once. It is not recursive in case that is what you are thinking.

cmd/xl-storage-disk-id-check.go
klauspost: Well, mostly 0s and a 42 at the end. Shouldn't really make much of a difference.

klauspost:
@harshavardhana 

> We can quite simply just do this I/O to take drive offline if its not writable or unresponsive. This satisfies all use-cases - in-fact we can reduce this to 5secs even 120secs is too high.

Didn't want to add too much load to the system, but if we cross reference with the last operation time we can do it more often. The timeout will have to be in this area at least, otherwise overloaded drives will go offline all the time. I will update this.

> Adding to this if we keep an open-fd map for reads and writes on the actual caller side. We can simply take all those fd's offline by closing them forcibly.

It would need to be a separate system since there isn't always a 1:1 relation between ops and fd's. 

We need *some* number of operations, since there may be some slow-running operations that are more limited by upstream than disks. If we set this limit too low we may get false disconnects.

Let me think if there is a good solution. Maybe as an alternative to this.

klauspost:
@harshavardhana See hhttps://github.com/minio/minio/pull/17552

So to disconnect a drive we currently need the following conditions:

A) Queue must be full. Default 512 requests (still waaaay to high IMO).
B) No new requests have started running within 15 seconds.
C) No running operation can have reported any successful operation within 30 seconds.

If we set the limit of A) to something lower, say default 32 as proposed in the PR, we still have B)

B) is there to make sure that we can compare the last successful operation, which isn't reset when a new operation starts. Imagine a flood of incoming requests that fills the queue.

So starting new operations will "reset" this part of the system. However I sent the PR since I don't really see it doing any harm. It will however not solve the majority of cases. It did give me an idea for improving this further, though.

klauspost:
Test are failing because tests keep creating new drives.

Trying to add a bit of cleanup, but not too optimistic.

harshavardhana:
@klauspost what I am suggesting is that we don't have to track hung calls with this PR anymore.

Since taking offline a drive is a sole responsibility of this healthchecker routine.

This implementation is actually enough to ensure a hung drive is taken offline. 2mins is a long time this also co-incides with hung_task_timeout value on Linux

We don't need any async monitoring of hung tasks at all, since we can take a drive offline more actively.

It simplifies everything in my opinion. I tested your PR with xfs_freeze this PR works and is just enough instead of current implementation.

Another thing beyond this I am bringing in to keep an fd state() which allows for us to close the fd's that are active on a hung drive allowing us to close such callers more effectively.

What do you think?

cmd/xl-storage-disk-id-check.go
```diff
@@ -821,6 +829,83 @@ func (p *xlStorageDiskIDCheck) monitorDiskStatus() {
...
+		// Disk has 2 minutes to complete write+read.
+		timeoutOperation = 2 * time.Minute
+
+		// If the disk has completed an operation successfully within last 5 seconds, don't check it.
+		skipIfSuccessBefore = 5 * time.Second
+	)
+
+	t := time.NewTicker(checkEvery)
+	defer t.Stop()
+	fn := mustGetUUID()
```
harshavardhana, line 848: We can always write new file, allowing filesystem to allocate a new block in some sense?.

cmd/xl-storage-disk-id-check.go
```diff
@@ -821,6 +829,83 @@ func (p *xlStorageDiskIDCheck) monitorDiskStatus() {
...
+		func() {
+			defer close(done)
+			err := p.storage.WriteAll(ctx, minioMetaTmpBucket, fn, toWrite)
+			if err != nil {
+				if osErrToFileErr(err) == errFaultyDisk {
+					goOffline(fmt.Errorf("unable to write: %w", err))
+				}
+				return
+			}
+			b, err := p.storage.ReadAll(context.Background(), minioMetaTmpBucket, fn)
```
harshavardhana, line 896: We should also add Delete() perhaps like how we do the regular monitoring?

harshavardhana (changes requested):

cmd/xl-storage-disk-id-check.go
klauspost: We use fsync, right? The data should be written on every run.

harshavardhana: I meant a new filename in essence, creates a new namespace entry on the fs.

cmd/xl-storage-disk-id-check.go
klauspost: I didn't really see a need, since it will be potentially leaky on restarts anyway. Trying to do *min*IO here :D

I do have a Q, though - we are clearing `.minio.sys/tmp` on every restart and not just `.minio.sys/tmp/.trash`, right?

I didn't want to place it in `.trash`, since we would be racing the cleanup routine - but I wanted it to be deleted on restarts, since it will be leaky (and we can't use a deterministic name, since at least two instances will be doing the operations).

harshavardhana: Yes correct upon every restart.

cmd/xl-storage-disk-id-check.go
klauspost: @harshavardhana Don't want to deal with potential buildup, so IMO overwriting should be fine. Should do plenty of writes to ensure we are testing disk reasonably.

cmd/xl-storage-disk-id-check.go
klauspost: Yeah. No need to delete then.

cmd/xl-storage-disk-id-check.go
harshavardhana: Okay

harshavardhana (approved):

harshavardhana:
I would go ahead and get rid of tracking disk I/O instead rely on this PR alone since this is more than sufficient.

harshavardhana:
PTAL @vadmeste

harshavardhana:
PTAL @vadmeste

klauspost:
Not sure it should replace the existing, but it should be a good fallback when there is low traffic.

cmd/xl-storage-disk-id-check.go
```diff
@@ -821,6 +829,83 @@ func (p *xlStorageDiskIDCheck) monitorDiskStatus() {
...
+		}
+		// Offset checks a bit.
+		time.Sleep(time.Duration(rng.Int63n(int64(1 * time.Second))))
+		done := make(chan struct{})
+		started := time.Now()
+		go func() {
+			timeout := time.NewTimer(timeoutOperation)
+			select {
+			// Reuse the same trigger
+			// If it triggers again, it took too long.
```
krisis:
```suggestion
```
This comment is no longer valid.

cmd/xl-storage-disk-id-check.go
```diff
@@ -821,6 +829,83 @@ func (p *xlStorageDiskIDCheck) monitorDiskStatus() {
...
+				if osErrToFileErr(err) == errFaultyDisk {
+					goOffline(fmt.Errorf("unable to write: %w", err))
+				}
+				return
+			}
+			b, err := p.storage.ReadAll(context.Background(), minioMetaTmpBucket, fn)
+			if err != nil || len(b) != len(toWrite) {
+				if osErrToFileErr(err) == errFaultyDisk {
+					goOffline(fmt.Errorf("unable to read: %w", err))
+				}
```
krisis, line 900: Shouldn't we call `goOffline` even in the case where only `len(b) != len(toWrite)` is true?

vadmeste: yes, I don't understand the logic here, it looks like it can be simplify with only
```
				if osErrToFileErr(err) == errFaultyDisk {
```
or
```
                                 if len(b) != len(toWrite) {
```

cmd/xl-storage-disk-id-check.go
```diff
@@ -821,6 +829,83 @@ func (p *xlStorageDiskIDCheck) monitorDiskStatus() {
 	}
 }
 
+// monitorDiskStatus should be called once when a drive has been marked offline.
```
poornas, line 832:
```suggestion
// monitorDiskWritable should be called once when a drive has been marked offline.
```

cmd/xl-storage-disk-id-check.go
klauspost: @krisis @vadmeste  I am on purpose *very* cautious on when we take a drive offline.

Take for example a disk full situation. That would fail the length check. However if we take the disk offline there is no way to fix the disk full situation - and we and up in a non-operational state. 

Therefore I *only* take the disk offline, when we get the faulty error.

krisis: @klauspost  Agreed, it makes sense to be extra cautious in this case.

krisis (approved):

PR #17548, opened 2023-06-30
experimental feat allow binding to specific NUMA cpu node

harshavardhana:
## Description
experimental feat allow binding to specific NUMA cpu node

## Motivation and Context
this flag is hidden for now

## How to test this PR?
Experimental feature of allowing MinIO to bind to specific
CPU node on NUMA servers `--numa-cpunode=0` for example.

## Types of changes

## Checklist:


cmd/server-main.go
```diff
@@ -495,7 +504,33 @@ func getServerListenAddrs() []string {
...
+		cpuNode := ctx.Int("numa-cpunode")
+		cpumask, err := numa.NodeToCPUMask(cpuNode)
+		if err != nil {
+			return err
+		}
+
+		if err = numa.SetSchedAffinity(0, cpumask); err != nil {
+			return err
+		}
+
```
klauspost, line 518: GOMAXPROCS should be set appropriately to not overprovision threads for cores.
```suggestion
		runtime.GOMAXPROCS(cpumask.OnesCount())
			
```
I seems like this should correctly be set to the logical core count of the CPU.

cmd/server-main.go
```diff
@@ -129,6 +131,13 @@ var ServerFlags = []cli.Flag{
 		Name:  "sftp",
 		Usage: "enable and configure an SFTP server",
 	},
+	cli.IntFlag{
```
klauspost, line 134: For an experiment it is fine, but a `--numa="node=1;mem=1,2"` would be more flexible.

harshavardhana: Correct but didn't want to bring in memory bind as well here. Since by default membind indeed is is spanning across.

klauspost: @harshavardhana It may make sense in some setups. We have only really tested drive read speeds. For most data operations (hashing/erasure coding/general copying) will be much slower if memory is on different cores - though L3 may be able to make up for it.

Just saying that a more flexible param may be more future proof, if we want to add it officially.

(you can resolve for now)

harshavardhana: Yeah I will bring membind as well

cmd/server-main.go
harshavardhana: The reason this is not needed @klauspost is that the `SetSchedAffinity` automatically handles this via syscall.Exec() - go automatically then gets the right "get_schedaffinity" value. 

I observed this via `strace` - thats why this kind of hack is needed. Otherwise Go runtime picks the affinity very early on and its impossible to set it correctly.

But I could this as a safety measure.

harshavardhana:
This is not helping with the object performance, however it helps with dperf. Might need more investigation - looks like reducing in GOMAXPROCS has other consequences. 

Closing this PR for now.

PR #17556, opened 2023-07-01
Close fd as soon as possible

drivebyer:
## Description

Should not use defer in loop. To avolid potential memory leak.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/format-disk-cache.go
+++ b/cmd/format-disk-cache.go
@@ -162,7 +162,6 @@ func loadFormatCache(ctx context.Context, drives []string) ([]*formatCacheV2, bo
 			logger.LogIf(ctx, err)
 			return nil, migrating, err
 		}
-		defer f.Close()
 		format, err := formatMetaCacheV1(f)
 		if err != nil {
 			continue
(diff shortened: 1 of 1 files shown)


cmd/format-disk-cache.go
```diff
@@ -172,6 +171,8 @@ func loadFormatCache(ctx context.Context, drives []string) ([]*formatCacheV2, bo
 			migrating = true
 		}
 		formats[i] = formatV2
+
+		f.Close()
```
harshavardhana, line 175: You are introducing a bug here by not closing the `fd` when the `formatMetaCacheV1()` returns an error. You are basically leaking fd now than before. 

Remember `defer` is stacked up here and this is a `load()` time call there is actually no real memory build up here. 

This happens during startup, and whenever the code finishes all the fd's are closed automatically.

Cached drives are usually like few, even if there are 1000 of them it is hardly anything for a modern system.

harshavardhana (changes requested):

cmd/format-disk-cache.go
drivebyer: You are right, I miss the `formatMetaCacheV1() ` error before

PR #17561, opened 2023-07-02
fix: upon DNS refresh() failure use previous values

harshavardhana:
## Description
fix: upon DNS refresh() failure use previous values

## Motivation and Context
DNS refresh() in-case of MinIO can safely re-use
the previous values on bare-metal setups, since
bare-metal setups do not change DNS in any manner dynamically.

Bonus: on containerized setups do not enable DNS
caching at all, as it may have adverse effects on
the overall effectiveness of k8s DNS design.

k8s DNS systems are dynamic and expect applications 
to avoid managing DNS caching themselves, instead a cleaner container native caching
implementations that must be used. Such as NodeLocalDNS.

This PR simplifies that, we only ever need DNS caching 
on bare-metal setups.

## How to test this PR?
The important change here is the failure to resolve DNS
does not actually cause nodes to disconnect on bare-metal
setups. Currently, in our releases, this would also
cause future DNS refreshes to be delayed causing a 
relatively longer duration of nodes being offline.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/common-main.go
+++ b/cmd/common-main.go
@@ -97,8 +97,13 @@ func init() {
 	initGlobalContext()
 
 	options := dnscache.ResolverRefreshOptions{
-		ClearUnused:      true,
-		PersistOnFailure: false,
+		ClearUnused: true,
+		// We always must persist on failure to effectively utilize
+		// dnscache package. First lookup when the lookup is not persisted
+		// will always provide the correct errors, however if the DNS
+		// has indeed resolved atleast once we must simply use
+		// older value and wait for the refresh window instead.
+		PersistOnFailure: true,
 	}
 
 	t, _ := minioVersionToReleaseTime(Version)
@@ -108,19 +113,10 @@ func init() {
 
 	globalIsCICD = env.Get("MINIO_CI_CD", "") != "" || env.Get("CI", "") != ""
 
-	containers := IsKubernetes() || IsDocker() || IsBOSH() || IsDCOS() || IsPCFTile()
-
-	// Call to refresh will refresh names in cache. If you pass true, it will also
-	// remove cached names not looked up since the last call to Refresh. It is a good idea
-	// to call this method on a regular interval.
+	// Call to refresh will refresh names in cache.
 	go func() {
-		var t *time.Ticker
-		if containers {
-			// k8s DNS TTL is 30s (Attempt a refresh only after)
-			t = time.NewTicker(30 * time.Second)
-		} else {
-			t = time.NewTicker(10 * time.Minute)
-		}
+		// baremetal setups set DNS refresh window to 30 minutes.
+		t := time.NewTicker(30 * time.Minute)
 		defer t.Stop()
 		for {
 			select {
--- a/cmd/common-main.go
+++ b/cmd/common-main.go
@@ -66,7 +66,6 @@ import (
 	"github.com/minio/pkg/ellipses"
 	"github.com/minio/pkg/env"
 	xnet "github.com/minio/pkg/net"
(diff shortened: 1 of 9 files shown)


cmd/utils.go
```diff
@@ -557,8 +557,13 @@ func ToS3ETag(etag string) string {
 
 // GetDefaultConnSettings returns default HTTP connection settings.
 func GetDefaultConnSettings() xhttp.ConnSettings {
+	lookupHost := globalDNSCache.LookupHost
+	if IsKubernetes() || IsDocker() {
```
jiuker, line 561:
```suggestion
	if IsKubernetes() || IsDocker() || IsBOSH() || IsDCOS() || IsPCFTile() {
```

harshavardhana: This was done on purpose @jiuker `IsDCOS, IsBOSH, IsPCFTile` are not really containers.

harshavardhana: also IsDocker() really covers any containerized environments.

jiuker (approved):

klauspost:
> Bonus: on containerized setups do not enable DNS
> caching at all, as it may have adverse effects on
> the overall effectiveness of k8s DNS design.

Can it handle potentially thousands of requests per second?

I would rather that the bonus was that we lowered it to 1 second or so, to at least limit how bad the performance can be. Also - with the other proposed change that would mean that a temporary outage in that system will not affect operations.

harshavardhana:
> Can it handle potentially thousands of requests per second?

yes way more than that is the whole design of https://kubernetes.io/docs/tasks/administer-cluster/nodelocaldns/ @klauspost 

> I would rather that the bonus was that we lowered it to 1 second or so, to at least limit how bad the performance can be. Also - with the other proposed change that would mean that a temporary outage in that system will not affect operations.

That would practically make it useless and overload k8s DNS. Our code was also buggy, because of the dynamic nature of IPs. 

Two tenants on k8s one that has cached the DNS inside MinIO would result in talking to another tenant's pod incorrectly when pods are being deleted etc, we are safe to some extent that they do not have the same root credentials however it can be disastrous in certain situations.

Our code as it was written today was practically speaking useless in containerized setups.

cmd/common-main.go
```diff
@@ -108,19 +113,10 @@ func init() {
...
 	go func() {
-		var t *time.Ticker
-		if containers {
-			// k8s DNS TTL is 30s (Attempt a refresh only after)
-			t = time.NewTicker(30 * time.Second)
-		} else {
-			t = time.NewTicker(10 * time.Minute)
-		}
+		// baremetal setups set DNS refresh window to 30 minutes.
+		t := time.NewTicker(30 * time.Minute)
```
vadmeste: Isn't 30 minutes very high ? especially with persist errors.. this means an error will be persisted for 30 minutes

harshavardhana: For baremetal setups this is not a real problem @vadmeste - since the initial lookup always succeeds once.

harshavardhana: Remember refresh() waits for 30mins to refresh. The issue was that the failure to refresh invalidated previous successful result @vadmeste

vadmeste: > For baremetal setups this is not a real problem @vadmeste - since the initial lookup always succeeds once.

The issue is when the first lookup fails, the library will persist it until someone refreshes it. This means sometimes when you start a cluster a node can take 30 minutes to start because of this. I took a look at the library code, the library doesn't try to look up again if it never had a success before.

vadmeste (approved):
LGTM

PR #17575, opened 2023-07-05
fix: avoid DiskInfo() to fetch metrics for all calls

harshavardhana:
## Description
fix: avoid DiskInfo() to fetch metrics for all calls

## Motivation and Context
introduce a new DiskInfoMetrics() API to fetch metrics
for specific Metrics() interested callers.

DiskInfo() is commonly used in PUT, Multipart phase avoids
calling for metrics which is not useful during these calls.

## How to test this PR?
There nothing much to test here, but there is a possible
improvements here in terms of latency per DiskInfo() call.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/erasure.go
+++ b/cmd/erasure.go
@@ -183,7 +183,7 @@ func getDisksInfo(disks []StorageAPI, endpoints []Endpoint) (disksInfo []madmin.
 				}
 				return nil
 			}
-			info, err := disks[index].DiskInfo(context.TODO())
+			info, err := disks[index].DiskInfoMetrics(context.TODO())
 			di := madmin.Disk{
 				Endpoint:       diskEndpoint,
 				DrivePath:      info.MountPath,
--- a/cmd/naughty-disk_test.go
+++ b/cmd/naughty-disk_test.go
@@ -116,6 +116,13 @@ func (d *naughtyDisk) NSScanner(ctx context.Context, cache dataUsageCache, updat
 	return d.disk.NSScanner(ctx, cache, updates, scanMode)
 }
 
+func (d *naughtyDisk) DiskInfoMetrics(ctx context.Context) (info DiskInfo, err error) {
+	if err := d.calcError(); err != nil {
+		return info, err
+	}
+	return d.disk.DiskInfoMetrics(ctx)
+}
+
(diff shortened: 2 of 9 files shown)


klauspost:
Well, that is why DiskInfo responses were/are cached. Was this removed?

harshavardhana:
> Well, that is why DiskInfo responses were/are cached. Was this removed?

No not really, even if you cache them. you see quick burst of small files I/O starts contending on the lastMinuteMetrics @klauspost 

Basically they contend heavily at 1sec boundaries. I am moving the metrics out such that they are not needed for every DiskInfo call. DiskInfo() call is quite common inside MinIO and it is still cached.

What is moved out is Metrics that is available via the `mc admin info` or Prometheus so nothing in the I/O path.

harshavardhana:
Even with lockless change in #17576 the total() part is contended, which is why it is not needed in the I/O path.

klauspost:
@harshavardhana They are competing with updates (the `add()` function), not eachother. As you showed in your trace most of the blocking was in `add` - that will make `total()` slow since they are competing for the same lock.

With many small operations there will be orders of magnitude more `add()` calls than `total()` calls.

harshavardhana:
> @harshavardhana They are competing with updates (the `add()` function), not eachother. As you showed in your trace most of the blocking was in `add` - that will make `total()` slow since they are competing for the same lock.
> 
> With many small operations there will be orders of magnitude more `add()` calls than `total()` calls.

yes, you are right about that @klauspost even with that it was simply unnecessary to return metrics when it's not useful or used by the caller.

which is most DiskInfo() calls. Only one caller needs metrics i.e LocalStorageInfo() at object layer. 

I am testing right now on the amount of real benefits this can bring in @klauspost

harshavardhana:
The benefits of this https://github.com/minio/minio/pull/17576 outweigh the change here. Unless we are okay with this change I will close this PR as not needed for now. We can live with the metrics we return with DiskInfo()

klauspost:
I am not per se against it. More that if we can avoid a specialization it would be fine.

We could remove it completely and just have the realtime metrics API handle this - but I think there is still a dream that we can select drives based on metrics, even if I am a bit sceptical.

PR #17584, opened 2023-07-06
Reduce allocations

klauspost:
## Description

* Add `stringsHasPrefixFold` which can compare string prefixes, while ignoring case and not allocating with `strings.ToLower`.
* Reuse all msgp.Readers
* Reuse metadata buffers when not reading data.

## How to test this PR?

Regular tests should cover functionality.

## Types of changes

### DIFF
--- a/cmd/storage-rest-client.go
+++ b/cmd/storage-rest-client.go
@@ -227,7 +227,8 @@ func (client *storageRESTClient) NSScanner(ctx context.Context, cache dataUsageC
 		rw.CloseWithError(waitForHTTPStream(respBody, rw))
 	}()
 
-	ms := msgp.NewReader(rr)
+	ms := msgpNewReader(rr)
+	defer readMsgpReaderPool.Put(rr)
 	for {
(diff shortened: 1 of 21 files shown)


cmd/storage-rest-client.go
```diff
@@ -227,7 +227,8 @@ func (client *storageRESTClient) NSScanner(ctx context.Context, cache dataUsageC
 		rw.CloseWithError(waitForHTTPStream(respBody, rw))
 	}()
 
-	ms := msgp.NewReader(rr)
+	ms := msgpNewReader(rr)
+	defer readMsgpReaderPool.Put(rr)
```
vadmeste: Shouldn't this be 'ms' instead 'rr' ?

klauspost: oh yeah... Hmmm.. going to make this type safe. We don't want this mistake elsewhere

harshavardhana (approved):

donatello (approved):

PR #17589, opened 2023-07-06
Replicate UserIdentity instead of user/SA/STS cred

donatello:
## Description
This simplifies site replication - we can remove the types specific to
IAM user, service account and STS creds as this type would cover all
three. It removes per credential type handling as all the required info
is present in the UserIdentity type.

~- [ ] https://github.com/minio/minio/pull/17483~

## Motivation and Context

This change simplifies site-replication, so we can add move info to the `UserIdentity` type and have it replicated automatically without specific handling for each credential type.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/admin-handlers-site-replication.go
+++ b/cmd/admin-handlers-site-replication.go
@@ -182,16 +182,20 @@ func (a adminAPIHandlers) SRPeerReplicateIAMItem(w http.ResponseWriter, r *http.
 				err = globalSiteReplicationSys.PeerAddPolicyHandler(ctx, item.Name, policy, item.UpdatedAt)
 			}
 		}
-	case madmin.SRIAMItemSvcAcc:
-		err = globalSiteReplicationSys.PeerSvcAccChangeHandler(ctx, item.SvcAccChange, item.UpdatedAt)
 	case madmin.SRIAMItemPolicyMapping:
 		err = globalSiteReplicationSys.PeerPolicyMappingHandler(ctx, item.PolicyMapping, item.UpdatedAt)
+	case madmin.SRIAMItemGroupInfo:
+		err = globalSiteReplicationSys.PeerGroupInfoChangeHandler(ctx, item.GroupInfo, item.UpdatedAt)
+	case madmin.SRIAMItemCredential:
+		err = globalSiteReplicationSys.PeerCredInfoChangeHandler(ctx, item.CredentialInfo, item.UpdatedAt)
+
+	// The following cases would be DEPRECATED by the previous case above.
+	case madmin.SRIAMItemSvcAcc:
+		err = globalSiteReplicationSys.PeerSvcAccChangeHandler(ctx, item.SvcAccChange, item.UpdatedAt)
 	case madmin.SRIAMItemSTSAcc:
 		err = globalSiteReplicationSys.PeerSTSAccHandler(ctx, item.STSCredential, item.UpdatedAt)
 	case madmin.SRIAMItemIAMUser:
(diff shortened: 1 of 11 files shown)


donatello:
Needs https://github.com/minio/madmin-go/pull/206

donatello:
PTAL @poornas @harshavardhana this is ready now and does not depend on https://github.com/minio/minio/pull/17483

cmd/admin-handlers-site-replication.go
```diff
@@ -182,16 +182,20 @@ func (a adminAPIHandlers) SRPeerReplicateIAMItem(w http.ResponseWriter, r *http.
...
+	case madmin.SRIAMItemCredential:
+		err = globalSiteReplicationSys.PeerCredInfoChangeHandler(ctx, item.CredentialInfo, item.UpdatedAt)
+
+	// The following cases would be DEPRECATED by the previous case above.
+	case madmin.SRIAMItemSvcAcc:
+		err = globalSiteReplicationSys.PeerSvcAccChangeHandler(ctx, item.SvcAccChange, item.UpdatedAt)
 	case madmin.SRIAMItemSTSAcc:
 		err = globalSiteReplicationSys.PeerSTSAccHandler(ctx, item.STSCredential, item.UpdatedAt)
 	case madmin.SRIAMItemIAMUser:
 		err = globalSiteReplicationSys.PeerIAMUserChangeHandler(ctx, item.IAMUser, item.UpdatedAt)
```
donatello, line 198: After this change, the three cases (SvcAcc, STSAcc, IAMUser) above are no longer needed, and we should remove them to avoid bugs. However this would be a backwards incompatible change - an older site may send these cases and we should ideally return an error, as the current handling is going to lead to out-of-sync info on the sites.

The best way to remove these cases is to remove these string constants (`SRIAMItemSvcAcc` etc) from madmin-go, but that would require a major bump of that library. Is that ok or shall we bump up site replication API version to detect the issue and log errors very early in the server startup? @harshavardhana @poornas 

**EDIT** I expect both may be needed - remove constants to catch erroneous usage of these constants at compile time and bump up site repl API version to log errors so an admin would update all sites sooner rather than later.

donatello: @harshavardhana @poornas PTAL, we need to see how we can upgrade the replication APIs, preventing older sites from sending potentially stale data. If we can get this change in, we can bring further IAM enhancements related to tracking/listing derived credentials - but that is blocked because it would lead to issues with SR. The problem here seems to be that we should be directly replicating the `UserIdentity` type instead of passing info to recreate this object on the remote site, which is leading to a more complicated implementation.

poornas: >Is that ok or shall we bump up site replication API version to detect the issue and log errors very early in the server startup?

Users may not be able to upgrade all clusters participating in site replication to the same release right away. The older API's could be used if the peer is on a release that doesn't support the enhanced API. it should be backward compatible because of the importance of IAM data, imo.

Maybe we can track the SR Version of peer during each call to choose the API's to use (just on server startup is not enough since peer can get upgraded at any time). Will the healing piece be unaffected here?

donatello:
This change should be reworked so that IAM data is replicated at the IAM object level and not at user account/policy content/policy mapping level. That way regardless of how IAM data representation is changed, we would not have to touch site replication related code. When we do that we can handle all the compatibility related issues and then SR would be stable at least IAM-wise.

With the approach in this PR, it is quite hard to maintain compatibility with older SR versions because the newer version would have to support sending info with the older and newer IAM item types.

Closing this PR now, while I work on changing the approach here.

PR #17591, opened 2023-07-07
Added a start script to inspect command output

shtripat:
## Description
Using this script, post decrypt we should be able to bring up the MinIO instance with same configuration.

## Motivation and Context
Using an inspect output, we should be able to bring up MinIO instance with same configuration. This helps to debug any
issues in the setup

## How to test this PR?
Follow the below steps to verify the changes
1. `mc support inspect ALIAS/BUCKET/** --airgap`
2. Decrypt the generated inspect.enc
3. Run the command `./start-minio.sh`

Verify that MinIO instance comes up properly and we are able to create alias and access the details from bucket.
```
$ mc alias set myminio http://localhost:9000 minio minio123
$ ms ls myminio
```

## Types of changes

## Checklist:


cmd/admin-handlers.go
```diff
@@ -2896,6 +2896,36 @@ func (a adminAPIHandlers) InspectDataHandler(w http.ResponseWriter, r *http.Requ
...
+	scrb.WriteString(`#!/usr/bin/env bash
+
+function main() {
+	for file in $(ls -1); do
+		dest_file=$(echo "$file" | cut -d ":" -f1)
+		mv "$file" "$dest_file"
+	done
+
+	# Read content of inspect-input.txt
+	MINIO_OPTS=$(grep "Server command line args" <./inspect-input.txt | sed "s/Server command line args: //g" | sed -r "s#https:\/\/#\.\/#g")
```
shtripat, line 2905: @harshavardhana I understand we should not simply take all the paths here and start server. In extracted dirs of the inspect file, if certain dir dont have a xl.meta file, we should set the permissions of those dirs as 644 so that when server comes up, it doesn't write the missing parts and we can achieve the actual stage where the source server state was.
Kindly comment or guide on this aspect so that this script can incorporate the same.

vadmeste: I am not sure why we are generating a script inside the code itself ? can't we just write a separate script under github.com/minio/minio and pass inspect.enc as argument ?

shtripat: There was a suggestion to include as part of inspect-data.enc, so post extraction user can just run `./start-minio.sh`

harshavardhana: Script is for repeatable and self contained @vadmeste.

This can be easily automated.

harshavardhana (approved):

harshavardhana:
PTAL @krisis

vadmeste (approved):
LGTM

PR #17593, opened 2023-07-07
Avoid per request URL parsing

klauspost:
## Description

Every REST request does a `url.Parse(c.url.String())` to clone a URL. Instead copy a pre-parsed URL.

Host will also be static, so we rewrite that on creation, removing empty port.

## Motivation and Context

Reduce allocations/processing. Observed to cause significant number of allocs in profiles.

## How to test this PR?

Regular operations should cover this.

## Types of changes

### DIFF
--- a/internal/rest/client.go
+++ b/internal/rest/client.go
@@ -126,16 +126,15 @@ func removeEmptyPort(host string) string {
 }
 
 // Copied from http.NewRequest but implemented to ensure we re-use `url.URL` instance.
-func (c *Client) newRequest(ctx context.Context, u *url.URL, body io.Reader) (*http.Request, error) {
+func (c *Client) newRequest(ctx context.Context, u url.URL, body io.Reader) (*http.Request, error) {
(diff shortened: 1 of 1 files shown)


klauspost:
(waiting for tests - still have a comment I didn't move)

internal/rest/client.go
```diff
@@ -379,14 +374,27 @@ func (c *Client) Close() {
 }
 
 // NewClient - returns new REST client.
-func NewClient(url *url.URL, tr http.RoundTripper, newAuthToken func(aud string) string) *Client {
+func NewClient(uu *url.URL, tr http.RoundTripper, newAuthToken func(aud string) string) *Client {
+	connected := int32(online)
```
vadmeste, line 378: casting doesn't seem to be needed, LGTM otherwise

klauspost: It is - `online` is untyped, so unless specified it will be an `int`.

harshavardhana (approved):

PR #17598, opened 2023-07-07
Support mc top net

jiuker:
## Description

Support mc top net

https://github.com/minio/mc/pull/4626
https://github.com/minio/madmin-go/pull/208

- [x] clear debug info

## Types of changes

## Checklist:


cmd/metrics-realtime.go
```diff
@@ -69,6 +71,22 @@ func collectLocalMetrics(types madmin.MetricType, opts collectMetricsOpts) (m ma
 	if types.Contains(madmin.MetricsSiteResync) {
 		m.Aggregated.SiteResync = globalSiteResyncMetrics.report(opts.depID)
 	}
+	if types.Contains(madmin.MetricNet) {
+		m.ByNet = map[string]madmin.NetMetrics{}
+		netStats, _ := net.GetInterfaceNetStats(getGlobalInternodeInterface())
```
klauspost: Check error. If `madmin.NetMetrics` doesn't have an `Error string` field, add one.

cmd/server-main.go
```diff
@@ -461,6 +464,35 @@ func initConfigSubsystem(ctx context.Context, newObject ObjectLayer) error {
...
+			return
+		}
+		ip := "127.0.0.1"
+		host, _ := mustSplitHostPort(globalMinioAddr)
+		if host != "" {
+			ip = host
+		}
+
+		ifs, err := net.Interfaces()
+		logger.FatalIf(err, "Unable to get IP addresses of this host")
```
klauspost: Is this important enough to crash the server?

Also, crash message seems wrong. You are fetching interfaces, not IP addresses - but perhaps it shouldn't be there.

cmd/server-main.go
```diff
@@ -461,6 +464,35 @@ func initConfigSubsystem(ctx context.Context, newObject ObjectLayer) error {
...
+		for _, interf := range ifs {
+			addrs, err := interf.Addrs()
+			if err == nil {
+				for _, addr := range addrs {
+					if strings.SplitN(addr.String(), "/", 2)[0] == ip {
+						globalInternodeInterface = interf.Name
+					}
+				}
+			}
+		}
```
klauspost, line 489: There is a non-zero chance that `globalInternodeInterface` ends up being `""` here.

Maybe set `globalInternodeInterface = ip` in that case.

go.mod
```diff
@@ -33,24 +33,24 @@ require (
 	github.com/hashicorp/golang-lru v0.5.4
 	github.com/inconshreveable/mousetrap v1.1.0
 	github.com/json-iterator/go v1.1.12
-	github.com/klauspost/compress v1.16.7
+	github.com/klauspost/compress v1.16.6
```
klauspost: Some of these are downgrades. Seems like a bad merge.

internal/net/net.go
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) 2015-2021 MinIO, Inc.
```
klauspost:
```suggestion
// Copyright (c) 2015-2023 MinIO, Inc.
```
Faster than Harsha again :D

cmd/metrics-realtime.go
```diff
@@ -69,6 +71,22 @@ func collectLocalMetrics(types madmin.MetricType, opts collectMetricsOpts) (m ma
...
+		endpoint := url.URL{
+			Scheme: "http",
+			Host:   globalLocalNodeName,
+		}
+		if globalIsTLS {
+			endpoint.Scheme = "https"
+		}
+		m.ByNet[endpoint.String()] = madmin.NetMetrics{
```
klauspost: See my comments on https://github.com/minio/madmin-go/pull/208

cmd/globals.go
```diff
@@ -393,6 +393,10 @@ var (
 	// Is _MINIO_DISABLE_API_FREEZE_ON_BOOT set?
 	globalDisableFreezeOnBoot bool
 
+	// Find out which NIC the instance is running on based on Endpoint
```
klauspost:
```suggestion
	// Contains NIC interface name used for internode communication
```

Are we sure there is only ever 1?

klauspost (changes requested):

cmd/metrics-realtime.go
```diff
@@ -69,6 +70,18 @@ func collectLocalMetrics(types madmin.MetricType, opts collectMetricsOpts) (m ma
 	if types.Contains(madmin.MetricsSiteResync) {
 		m.Aggregated.SiteResync = globalSiteResyncMetrics.report(opts.depID)
 	}
+	if types.Contains(madmin.MetricNet) {
+		m.Aggregated.Net = &madmin.NetMetrics{
+			CollectedAt:   UTCNow(),
+			InterfaceName: getGlobalInternodeInterface(),
```
shtripat: Instead of calling the function twice, assign the output to a variable and use it.

cmd/server-main.go
klauspost: If the loop above doesn't match anything we can still end up with an empty string.

Just set `globalInternodeInterface = ip` unconditionally above, instead of depending on the error value.

go.mod
```diff
@@ -240,3 +240,5 @@ require (
 	gopkg.in/square/go-jose.v2 v2.6.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+replace github.com/minio/madmin-go/v3 => github.com/jiuker/madmin-go/v3 v3.0.0-20230711080547-dcff875f13a3
```
klauspost: This can be removed and dependency can be updated.

harshavardhana (approved):

klauspost (approved):
LGTM

PR #17599, opened 2023-07-07
Avoid extra GetObjectInfo call in DeleteObject API

poornas:
on the replicating side. For receiving side, it is just a regular DeleteObject call.

Also fixing a corner case where version being purged is absent on target (either due to replication not yet complete or target version already deleted in a one-way replication or while replication was disabled). In such cases, mark version purge complete.

## Motivation and Context
optimization

## How to test this PR?
regular replication of deletes/delete markers; setting compliance retention on object version should throw errors if deleting the version before retention expiry

## Types of changes

## Checklist:

### DIFF
--- a/cmd/object-handlers.go
+++ b/cmd/object-handlers.go
@@ -2326,33 +2326,29 @@ func (api objectAPIHandlers) DeleteObjectHandler(w http.ResponseWriter, r *http.
 		return
 	}
 
-	getObjectInfo := objectAPI.GetObjectInfo
-	if api.CacheAPI() != nil {
-		getObjectInfo = api.CacheAPI().GetObjectInfo
-	}
-
 	os := newObjSweeper(bucket, object).WithVersion(opts.VersionID).WithVersioning(opts.Versioned, opts.VersionSuspended)
-	// Mutations of objects on versioning suspended buckets
-	// affect its null version. Through opts below we select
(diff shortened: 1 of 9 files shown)


cmd/object-handlers.go
```diff
@@ -2326,59 +2326,44 @@ func (api objectAPIHandlers) DeleteObjectHandler(w http.ResponseWriter, r *http.
...
-
 	os := newObjSweeper(bucket, object).WithVersion(opts.VersionID).WithVersioning(opts.Versioned, opts.VersionSuspended)
-	// Mutations of objects on versioning suspended buckets
-	// affect its null version. Through opts below we select
-	// the null version's remote object to delete if
-	// transitioned.
-	goiOpts := os.GetOpts()
-	goi, gerr := getObjectInfo(ctx, bucket, object, goiOpts)
-	if gerr == nil {
-		os.SetTransitionState(goi.TransitionedObject)
```
harshavardhana, line 2342: Is this not necessary anymore?

harshavardhana (changes requested):

cmd/object-handlers.go
poornas: fixed this

harshavardhana (approved):

klauspost (approved):
LGTM

PR #17602, opened 2023-07-08
fix: set request ID in tracing context key

donatello:
## Description

Since `addCustomerHeaders` middleware was after the `httpTracer` middleware, the request ID was not set in the http tracing context. By reordering these middleware functions, the request ID header becomes available. We also avoid setting the tracing context key again in `newContext`.

Bonus: All middleware functions are renamed with a "Middleware" suffix to avoid confusion with http Handler functions.

## Motivation and Context

We can avoid setting the http tracing context key multiple times - it was being set in `newContext` presumably because the request id was not otherwise available.

## How to test this PR?

Tracing functionality should be unaffected.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/auth-handler.go
+++ b/cmd/auth-handler.go
@@ -595,8 +595,8 @@ func isSupportedS3AuthType(aType authType) bool {
 	return ok
 }
 
-// setAuthHandler to validate authorization header for the incoming request.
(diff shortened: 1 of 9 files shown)


harshavardhana (approved):

vadmeste (approved):
LGTM

poornas (approved):

PR #17484, opened 2023-06-21
Print certain log messages once per error

kaankabalak:
## Description
This commit prints certain log messages once per error instead of multiple times.

## Motivation and Context
Logs being unnecessarily printed multiple times is causing clutter and makes it harder for users to navigate through error messages in their deployments.

## How to test this PR?
Make sure that log messages involved in the PR are printed only once per error and the messages work as expected.

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Optimization (provides speedup with no functional changes)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [ ] Fixes a regression (If yes, please add `commit-id` or `PR #` here)
- [ ] Unit tests added/updated
- [ ] Internal documentation updated
- [ ] Create a documentation update request [here](https://github.com/minio/docs/issues/new?label=doc-change,title=Doc+Updated+Needed+For+PR+github.com%2fminio%2fminio%2fpull%2fNNNNN)

PR #17618, opened 2023-07-10
Further print log messages once per error

kaankabalak:
## Description

This commit prints certain log messages once per error instead of multiple times.

## Motivation and Context

Logs being unnecessarily printed multiple times is causing clutter and makes it harder for users to navigate through error messages in their deployments.

## How to test this PR?

Make sure that log messages involved in the PR are printed only once per error and the messages work as expected.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/bucket-quota.go
+++ b/cmd/bucket-quota.go
@@ -65,10 +65,10 @@ func (sys *BucketQuotaSys) Init(objAPI ObjectLayer) {
 func (sys *BucketQuotaSys) GetBucketUsageInfo(bucket string) (BucketUsageInfo, error) {
 	v, err := sys.bucketStorageCache.Get()
 	if err != nil && v != nil {
-		logger.LogOnceIf(GlobalContext, fmt.Errorf("unable to retrieve usage information for bucket: %s, relying on older value cached in-memory: err(%v)", bucket, err), "bucket-usage-cache"+bucket)
+		logger.LogOnceIf(GlobalContext, fmt.Errorf("unable to retrieve usage information for bucket: %s, relying on older value cached in-memory: err(%v)", bucket, err), "bucket-usage-cache-"+bucket)
 	}
 	if v == nil {
-		logger.LogOnceIf(GlobalContext, errors.New("unable to retrieve usage information for bucket: %s, no reliable usage value available - quota will not be enforced"), "bucket-usage-empty"+bucket)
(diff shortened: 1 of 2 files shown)


klauspost (approved):

cmd/bucket-quota.go
```diff
@@ -65,10 +65,10 @@ func (sys *BucketQuotaSys) Init(objAPI ObjectLayer) {
 func (sys *BucketQuotaSys) GetBucketUsageInfo(bucket string) (BucketUsageInfo, error) {
 	v, err := sys.bucketStorageCache.Get()
 	if err != nil && v != nil {
-		logger.LogIf(GlobalContext, fmt.Errorf("unable to retrieve usage information for bucket: %s, relying on older value cached in-memory: err(%v)", bucket, err))
+		logger.LogOnceIf(GlobalContext, fmt.Errorf("unable to retrieve usage information for bucket: %s, relying on older value cached in-memory: err(%v)", bucket, err), "bucket-usage-cache"+bucket)
```
jiuker: `"bucket-usage-cache"+bucket` have a bad format. We can add `: `?

klauspost: @jiuker This is the token used to identify duplicates, so it doesn't have to be "nice".

jiuker: @klauspost `-` is ok. But there's no distinction at the moment and I don't think it looks good

kaankabalak: I have applied the style recommended in the previous PR, however I will add a slash in between as it looks cleaner. 👍

kaankabalak: Thanks @jiuker, I have added a slash in between, would appreciate it if you could take another look. 👍

When this PR gets merged, I can also add the slash in between for the errors changed in #17484

kaankabalak:
The failing tests seem to be unrelated to the PR, however I'll take a look to see if there's a way to make them pass.

klauspost:
Retriggered the failed tests. Yeah. It is very flaky at the moment.

jiuker (approved):
lgtm

harshavardhana (approved):

harshavardhana:
> Retriggered the failed tests. Yeah. It is very flaky at the moment.

Yeah @poornas is looking into replication related failures

cmd/bucket-quota.go
kaankabalak: Hi @jiuker, I have sent https://github.com/minio/minio/pull/17631 to update the errors changed in #17484, would appreciate it if you could take a look 🙏

PR #17632, opened 2023-07-11
Send AuditLog via new middleware fn for admin APIs

donatello:
## Description

(This PR is sent with two commits for easier reviewing, but both need to go in together)

A new middleware function is added for admin handlers, including options
for modifying certain behaviors. This admin middleware:

- sets the handler context via reflection in the request and sends AuditLog
- checks for object API availability (skipping it if a flag is passed)
- enables gzip compression (skipping it if a flag is passed)
- enables header tracing (adding body tracing if a flag is passed)

While the new function is a middleware, due to the flags used for
conditional behavior modification, it is used in each route registration
call.

To try to ensure that no regressions are introduced, the following
changes were done mechanically mostly with `sed` and regexps:

- Remove defer logger.AuditLog in admin handlers
- Replace newContext() calls with r.Context()
- Update admin routes registration calls

Bonus: remove unused NetSpeedtestHandler

### [Disable object layer check for some handlers](https://github.com/minio/minio/commit/9fd70cf51319e4fdc436b72b211969bd09c83884)

Since the new adminMiddleware function checks for object layer presence
by default, we need to pass the `noObjLayerFlag` explicitly for admin
handlers that should work even when it is not available. The following
admin handlers do not require it:

- ServerInfoHandler
- StartProfilingHandler
- DownloadProfilingHandler
- ProfileHandler
- SiteReplicationDevNull
- SiteReplicationNetPerf
- TraceHandler

For these handlers adminMiddleware does not check for object layer
presence (disabled by passing the `noObjLayerFlag`), and for all other
handlers, the pre-check ensures that the handler is not called when the
object layer is not available - the client would get a
ErrServerNotInitialized and can retry later.

This `noObjLayerFlag` is added based on existing behavior for these
handlers only.

## Motivation and Context

The motivation is to reduce some repetition of important calls (e.g. sending AuditLog) - common functionality for admin handlers can be done in a single function that is a middleware for the actual handler. This let's handler functions focus on specific functionality being provided by them and ensure that middlewares take care of common functionality.

We are no longer setting the function name for AuditLog with an explicit (literal) string in the call to `newContext`. Instead the handler function name is extracted by reflection and sanitized similar to how it is done in http tracing and then passed to `newContext` in the new `adminMiddleware`

## How to test this PR?

There would be no functionality changes introduced by this change. Bugs found while developing this change, were sent as separate changes (https://github.com/minio/minio/pull/17622 and https://github.com/minio/minio/pull/17623).

## Types of changes

## Checklist:

### DIFF
--- a/cmd/admin-router.go
+++ b/cmd/admin-router.go
@@ -158,7 +158,7 @@ func registerAdminRouter(router *mux.Router, enableConfigOps bool) {
 		adminRouter.Methods(http.MethodPost).Path(adminVersion+"/update").HandlerFunc(adminMiddleware(adminAPI.ServerUpdateHandler, traceAllFlag)).Queries("updateURL", "{updateURL:.*}")
 
 		// Info operations
-		adminRouter.Methods(http.MethodGet).Path(adminVersion + "/info").HandlerFunc(adminMiddleware(adminAPI.ServerInfoHandler, traceAllFlag))
+		adminRouter.Methods(http.MethodGet).Path(adminVersion + "/info").HandlerFunc(adminMiddleware(adminAPI.ServerInfoHandler, traceAllFlag, noObjLayerFlag))
 		adminRouter.Methods(http.MethodGet, http.MethodPost).Path(adminVersion + "/inspect-data").HandlerFunc(adminMiddleware(adminAPI.InspectDataHandler, noGZFlag, traceAllFlag))
 
 		// StorageInfo operations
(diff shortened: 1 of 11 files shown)


donatello:
PTAL this is ready.

donatello:
Addressed the merge conflicts - PTAL!

cmd/admin-router.go
```diff
@@ -35,6 +38,106 @@ const (
...
+//
+// Note that, while this is a middleware function (i.e. it takes a handler
+// function and returns one), due to flags being passed based on required
+// conditions, it is done per-"handler function registration" in the router.
+//
+// When no flags are passed, gzip compression, http tracing of headers and
+// checking of object layer availability are all enabled. Use flags to modify
+// this behavior.
+func adminMiddleware(f http.HandlerFunc, flags ...hFlag) http.HandlerFunc {
+	// Collect all flags with bitwise-OR and assign operator
```
krisis, line 95: Suggestion: The caller can do the bitwise-OR and pass a single `hFlag`. We don't need variable number of arguments on this account.

krisis (commented):
Neat!

krisis (approved):

poornas (approved):

cmd/admin-router.go
donatello: The variable arguments make the flags easier to read imo. With bitwise OR gofmt removes the spaces around the operator and makes it slightly harder to read.

PR #17635, opened 2023-07-11
return tags as part of Head/Get calls

harshavardhana:
## Description
return tags as part of Head/Get calls

## Motivation and Context
AWS S3 only returns the number of tag
counts, along with that we must return
the tags as well to avoid another metadata
call to the server.

## How to test this PR?
Verify after attaching new tags on the object
that `mc stat` returns the tags as part
of the Head call.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/api-headers.go
+++ b/cmd/api-headers.go
@@ -23,11 +23,11 @@ import (
 	"encoding/xml"
 	"fmt"
 	"net/http"
-	"net/url"
 	"strconv"
 	"strings"
 	"time"
 
+	"github.com/minio/minio-go/v7/pkg/tags"
 	"github.com/minio/minio/internal/crypto"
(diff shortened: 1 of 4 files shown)


krisis (approved):

poornas (approved):

harshavardhana:
Need to test one scenario with large tag key names and values.

klauspost (changes requested):
Please gate this behind a feature header.

Several if not all reverse proxies have header size limits, and adding up to 8K of extra data is likely to exceed default config.

harshavardhana:
This is not returned by default, however setting `x-amz-tagging-directive: ACCESS` makes the server return the tags with `x-amz-tagging: <tags>`

klauspost (approved):
👍🏼

PR #17640, opened 2023-07-12
fix: do not os.Exit(1) while writing goroutines during shutdown

harshavardhana:
## Description
fix: do not os.Exit(1) while writing goroutines during shutdown

## Motivation and Context
Also shutdown poll add jitter, to verify if the shutdown
sequence can finish before 500ms, this reduces the overall 
time taken during "restart" of the service.

Provides speedup for `mc admin service restart` during
active I/O, also ensures that systemd doesn't treat the
returned 'error' as a failure, certain configurations
in systemd can cause it to 'auto-restart' the process by-itself 
which can interfere with `mc admin service restart`.

It can be observed how now restarting the service is 
much snappier.

## How to test this PR?
via `mc admin service restart alias/`, via `CTRL-C` while
shutting down the process.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/signals.go
+++ b/cmd/signals.go
@@ -46,21 +46,17 @@ func handleSignals() {
 	}
 
 	stopProcess := func() bool {
-		var err, oerr error
-
 		// send signal to various go-routines that they need to quit.
 		cancelGlobalContext()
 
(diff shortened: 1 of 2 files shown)


klauspost (approved):
LGTM

jiuker (approved):

PR #17644, opened 2023-07-12
shuffle buckets randomly before being scanned

harshavardhana:
## Description
shuffle buckets randomly before being scanned

## Motivation and Context
this randomness is needed to avoid scanning
the same buckets across different erasure sets,
in the same order.

allow random buckets to be scanned instead
allowing a wider spread of ILM, replication
checks.

Additionally do not loop over twice to fill
the channel, fill the channel regardless of
having bucket new or old.

## How to test this PR?
This requires 100s of buckets and various
ILM policies and cleaning up objects inside them.

without this PR the affinity toward other buckets
is kind of lost, since bigger buckets can sort 
of slow things down for others, instead of different 
erasure sets should spend time doing things for 
different buckets.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/erasure.go
+++ b/cmd/erasure.go
@@ -277,31 +277,33 @@ func (er erasureObjects) getOnlineDisksWithHealing() (newDisks []StorageAPI, hea
 	var wg sync.WaitGroup
 	disks := er.getDisks()
 	infos := make([]DiskInfo, len(disks))
-	for _, i := range hashOrder(UTCNow().String(), len(disks)) {
+	r := rand.New(rand.NewSource(time.Now().UnixNano()))
+	for _, i := range r.Perm(len(disks)) {
 		i := i
 		wg.Add(1)
 		go func() {
 			defer wg.Done()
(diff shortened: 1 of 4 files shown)


cmd/erasure.go
```diff
@@ -373,23 +375,22 @@ func (er erasureObjects) nsScanner(ctx context.Context, buckets []BucketInfo, wa
 
 	// Put all buckets into channel.
 	bucketCh := make(chan BucketInfo, len(buckets))
-	// Add new buckets first
```
klauspost, line 376: You should keep this functionality. Otherwise new buckets will take quite a while to show up in usage.

It will only be for one cycle anyway.

klauspost (commented):
Let's try it out. Not really sure what effect it will have on usage update frequency, but I think it should be fine, if we keep scanning new buckets first.

krisis (approved):

klauspost (approved):
LGTM

Issue #17409, opened 2023-06-13
Minio Cors config does not work - "Cross-Origin Request Blocked"


richardARPANET:
Version `RELEASE.2023-06-02T23-17-26Z`

I'm doing s3 multipart upload, part way through the uploading of parts I get this error in the web browser console:

```
Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://objectstorage.example.com/my-media/uploads/000ZheRnGHoP.mp4?uploadId=NjQ0NzJkMWEtNzQzMC00OTM5LWI5NzMtMmM4NTY2MGRlMGI0LjUwNjIyMjgzLTM1ZGEtNDNhNC1iOTA0LTJjYWFlNDg5ZjY5ZQ&partNumber=3&AWSAccessKeyId=IdOT27Jz7UWsD2Aq&Signature=PMT6BCUuDCdEaMc%2BmxRxC2XP2MI%3D&Expires=1686764158. (Reason: CORS request did not succeed). Status code: (null).
```

When I'm uploading I'm on the website https://example.com (replaced real hostname with example for privacy).

And in the Minio Console I have set under Settings -> API -> Cors Allow Origin. The following values:

```
*
https://example.com
https://www.example.com
```

In my PUT request which errors I have set the following headers:

```
Accept application/json, text/plain, */*
Accept-Encoding gzip, deflate, br
Accept-Language en-US,en;q=0.5
Cache-Control no-cache
Connection keep-alive
Content-Length 5242880
DNT 1
Host objectstorage.example.com
Origin https://example.com/
Pragma no-cache
Sec-Fetch-Dest empty
Sec-Fetch-Mode cors
Sec-Fetch-Site same-site
TE trailers
User-Agent Mozilla/5.0 (X11; Linux x86_64; rv:109.0) Gecko/20100101 Firefox/111.0
```

Why does this PUT request fail when I have configured Minio seemingly correctly?

harshavardhana:
@richardARPANET can you provide `mc admin config export alias/` output?

richardARPANET:
@harshavardhana I need to ask Devops where I can run that. In the mean time I managed to export this from the console UI using the 'Export' button.

```
subnet license= api_key= proxy= 
# callhome enable=off frequency=24h 
site name= region= 
[...]

PR #17651, opened 2023-07-14
fix when set null orgin will return null back

jiuker:
## Description

CROS must set  AllowedOrigins work for [*]

## How to test this PR?

`curl -v -XOPTIONS -H "Origin: null" https://host:9000/testing/README.md`
You will see the following header!
before:
`Access-Control-Allow-Origin: null`
after:
`Access-Control-Allow-Origin:  *`
## Types of changes
- [x] Bug fix ( #17409 )

## Checklist:

### DIFF
--- a/cmd/api-router.go
+++ b/cmd/api-router.go
@@ -523,14 +523,9 @@ func corsHandler(handler http.Handler) http.Handler {
 		"x-amz*",
 		"*",
 	}
-
-	return cors.New(cors.Options{
+	opts := cors.Options{
 		AllowOriginFunc: func(origin string) bool {
-			allowedOrigins := globalAPIConfig.getCorsAllowOrigins()
-			if len(allowedOrigins) == 0 {
-				allowedOrigins = []string{"*"}
-			}
-			for _, allowedOrigin := range allowedOrigins {
+			for _, allowedOrigin := range globalAPIConfig.getCorsAllowOrigins() {
 				if wildcard.MatchSimple(allowedOrigin, origin) {
 					return true
 				}
@@ -549,6 +544,13 @@ func corsHandler(handler http.Handler) http.Handler {
 		AllowedHeaders:   commonS3Headers,
 		ExposedHeaders:   commonS3Headers,
 		AllowCredentials: true,
-		AllowedOrigins:   globalAPIConfig.getCorsAllowOrigins(),
-	}).Handler(handler)
+	}
+	for _, origin := range globalAPIConfig.getCorsAllowOrigins() {
+		if origin == "*" {
+			opts.AllowOriginFunc = nil
+			opts.AllowedOrigins = globalAPIConfig.getCorsAllowOrigins()
+			break
+		}
+	}
+	return cors.New(opts).Handler(handler)
 }
--- a/cmd/api-router.go
+++ b/cmd/api-router.go
@@ -24,8 +24,6 @@ import (
 
 	"github.com/klauspost/compress/gzhttp"
 	"github.com/minio/console/restapi"
-	"github.com/minio/minio/internal/config"
-	"github.com/minio/minio/internal/config/api"
 	xhttp "github.com/minio/minio/internal/http"
 	"github.com/minio/minio/internal/logger"
 	"github.com/minio/mux"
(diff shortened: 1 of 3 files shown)


internal/config/api/api.go
```diff
@@ -281,3 +277,11 @@ func LookupConfig(kvs config.KVS) (cfg Config, err error) {
 
 	return cfg, nil
 }
+
+func GetCorsAllowOrigin(kvs config.KVS) []string {
+	corsAllowOrigin := strings.Split(env.Get(EnvAPICorsAllowOrigin, kvs.Get(apiCorsAllowOrigin)), ",")
```
shtripat: Should we make the condition as `if len(corsAllowOrigin) == 0 || (len(corsAllowOrigin) == 1 && corsAllowOrigin[0] == "")` for better readability?

shtripat (approved):
lgtm

cmd/api-router.go
```diff
@@ -549,5 +549,6 @@ func corsHandler(handler http.Handler) http.Handler {
 		AllowedHeaders:   commonS3Headers,
 		ExposedHeaders:   commonS3Headers,
 		AllowCredentials: true,
+		AllowedOrigins:   globalAPIConfig.getCorsAllowOrigins(),
```
harshavardhana: This is not correct, please revert.

internal/config/api/api.go
```diff
@@ -194,7 +194,7 @@ func LookupConfig(kvs config.KVS) (cfg Config, err error) {
 	}
 
 	corsAllowOrigin := strings.Split(env.Get(EnvAPICorsAllowOrigin, kvs.Get(apiCorsAllowOrigin)), ",")
-	if len(corsAllowOrigin) == 0 {
+	if len(corsAllowOrigin) == 1 && corsAllowOrigin[0] == "" {
```
harshavardhana:
```suggestion
	if len(corsAllowOrigin) == 1 && corsAllowOrigin[0] == "" {
```

This is not needed we should do it differently here.

```
corsList := env.Get(EnvAPICorsAllowOrigin, kvs.Get(apiCorsAllowOrigin))
if corsList == "" {
   corsAllowOrigin = []string{"*"} // defaults to '*'
} else {
   corsAllowOrigin = strings.Split(corsList, ",")
   for _, cors := range corsAllowOrigin {
       if cors == "" {
           return cfg, errors.New("invalid cors value")
       }
   }
} 
```

harshavardhana (changes requested):

harshavardhana:
The reason why this is wrong

- AllowedOrigins []string: A list of origins a cross-domain request can be executed from. If the special * value is present in the list, all origins will be allowed. An origin may contain a wildcard (*) to replace 0 or more characters (i.e.: http://*.domain.com). The usage of wildcards implies a small performance penalty. Only one wildcard can be used per origin. The default value is *.

- AllowOriginFunc func (origin string) bool: A custom function to validate the origin. It takes the origin as an argument and returns true if allowed, or false otherwise. If this option is set, the content of AllowedOrigins is ignored.

If you look at the `rs/cors` library setting AllowedOrigins will break AllowOriginFunc basically breaking existing users - you need a different change such as this to make this fully complete

```diff
diff --git a/cmd/api-router.go b/cmd/api-router.go
index 517b0dd6d..875c572f5 100644
--- a/cmd/api-router.go
+++ b/cmd/api-router.go
@@ -524,13 +524,9 @@ func corsHandler(handler http.Handler) http.Handler {
                "*",
        }
 
-       return cors.New(cors.Options{
+       opts := cors.Options{
                AllowOriginFunc: func(origin string) bool {
-                       allowedOrigins := globalAPIConfig.getCorsAllowOrigins()
-                       if len(allowedOrigins) == 0 {
-                               allowedOrigins = []string{"*"}
-                       }
-                       for _, allowedOrigin := range allowedOrigins {
+                       for _, allowedOrigin := range globalAPIConfig.getCorsAllowOrigins() {
                                if wildcard.MatchSimple(allowedOrigin, origin) {
                                        return true
                                }
@@ -549,5 +545,15 @@ func corsHandler(handler http.Handler) http.Handler {
                AllowedHeaders:   commonS3Headers,
                ExposedHeaders:   commonS3Headers,
                AllowCredentials: true,
-       }).Handler(handler)
+       }
+
+       for _, origin := range globalAPIConfig.getCorsAllowOrigins() {
+               if origin == "*" {
+                       opts.AllowOriginFunc = nil
+                       opts.AllowedOrigins = globalAPIConfig.getCorsAllowOrigins()
+                       break
+               }
+       }
+
+       return cors.New().Handler(handler)
 }
```

harshavardhana:
However I feel the `rs/cors` library is doing a bit of magic here and is not documenting itself correctly on what is to be expected.

harshavardhana (approved):

PR #17652, opened 2023-07-14
fix: ILM newer noncurrent version limit must return correct versions

harshavardhana:
## Description
fix: ILM newer noncurrent version limit must return correct versions

## Motivation and Context
objects/versions that are not expired via NewerNoncurrentVersions 
must be properly returned to be applied under further ILM actions.

this would cause legitimately expired objects to be missed 
from expiration.

## How to test this PR?
Would require a setup with ILM newer noncurrent version expiration
as well as non-current version expiration. 

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/mint.yml
+++ b/.github/workflows/mint.yml
@@ -3,7 +3,7 @@ name: Mint Tests
 on:
   pull_request:
     branches:
-    - master
+      - master
 
 # This ensures that previous jobs for the PR are canceled when the PR is
 # updated.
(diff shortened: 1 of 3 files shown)


klauspost:
`FAIL: TestDataUsageUpdate (0.04s)` is likely because you randomized buckets (or a sideeffect of that change)

vadmeste (approved):
Good catch!

krisis (approved):
Good catch!

PR #17663, opened 2023-07-15
move bucket centric metrics to /minio/v2/metrics/bucket handlers

harshavardhana:
## Description
move bucket centric metrics to /minio/v2/metrics/bucket handlers

## Motivation and Context
users/customers do not have a reasonable number of buckets anymore, 
this is why we must avoid overpopulating cluster endpoints, instead 
move the bucket monitoring to a separate endpoint.

some of it's a breaking change here for a couple of metrics, but 
it is imperative that we do it to improve the responsiveness of 
our Prometheus cluster endpoint.

## How to test this PR?
Bucket-centric metrics are moved to bucket centric endpoint
for monitoring and scraping.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/metrics-v2.go
+++ b/cmd/metrics-v2.go
@@ -40,24 +40,26 @@ import (
 )
 
 var (
-	nodeCollector     *minioNodeCollector
-	clusterCollector  *minioClusterCollector
-	peerMetricsGroups []*MetricsGroup
+	nodeCollector           *minioNodeCollector
+	clusterCollector        *minioClusterCollector
+	bucketCollector         *minioBucketCollector
+	peerMetricsGroups       []*MetricsGroup
+	bucketPeerMetricsGroups []*MetricsGroup
 )
 
 func init() {
 	clusterMetricsGroups := []*MetricsGroup{
-		getBucketUsageMetrics(),
 		getNodeHealthMetrics(),
 		getClusterStorageMetrics(),
 		getClusterTierMetrics(),
+		getClusterUsageMetrics(),
 		getKMSMetrics(),
 	}
 
 	peerMetricsGroups = []*MetricsGroup{
 		getCacheMetrics(),
 		getGoMetrics(),
-		getHTTPMetrics(),
+		getHTTPMetrics(false),
 		getNotificationMetrics(),
 		getLocalStorageMetrics(),
 		getMinioProcMetrics(),
@@ -82,7 +84,7 @@ func init() {
 		getNodeHealthMetrics(),
 		getLocalDriveStorageMetrics(),
 		getCacheMetrics(),
-		getHTTPMetrics(),
+		getHTTPMetrics(false),
 		getNetworkMetrics(),
 		getMinioVersionMetrics(),
 		getS3TTFBMetric(),
@@ -90,8 +92,20 @@ func init() {
 		getDistLockMetrics(),
 	}
 
+	bucketMetricsGroups := []*MetricsGroup{
+		getBucketUsageMetrics(),
+		getHTTPMetrics(true),
+		getBucketTTFBMetric(),
+	}
+
+	bucketPeerMetricsGroups = []*MetricsGroup{
+		getHTTPMetrics(true),
+		getBucketTTFBMetric(),
+	}
+
 	nodeCollector = newMinioCollectorNode(nodeGroups)
 	clusterCollector = newMinioClusterCollector(allMetricsGroups)
+	bucketCollector = newMinioBucketCollector(bucketMetricsGroups)
 }
 
 // MetricNamespace is top level grouping of metrics to create the metric name.
@@ -121,11 +135,13 @@ const (
 	ioSubsystem               MetricSubsystem = "io"
 	nodesSubsystem            MetricSubsystem = "nodes"
 	objectsSubsystem          MetricSubsystem = "objects"
+	bucketsSubsystem          MetricSubsystem = "bucket"
 	processSubsystem          MetricSubsystem = "process"
 	replicationSubsystem      MetricSubsystem = "replication"
 	requestsSubsystem         MetricSubsystem = "requests"
 	requestsRejectedSubsystem MetricSubsystem = "requests_rejected"
 	timeSubsystem             MetricSubsystem = "time"
+	ttfbSubsystem             MetricSubsystem = "requests_ttfb"
 	trafficSubsystem          MetricSubsystem = "traffic"
 	softwareSubsystem         MetricSubsystem = "software"
 	sysCallSubsystem          MetricSubsystem = "syscall"
@@ -192,7 +208,7 @@ const (
 
 	sizeDistribution    = "size_distribution"
 	versionDistribution = "version_distribution"
-	ttfbDistribution    = "ttfb_seconds_distribution"
+	ttfbDistribution    = "seconds_distribution"
 
 	lastActivityTime = "last_activity_nano_seconds"
 	startTime        = "starttime_seconds"
@@ -306,6 +322,16 @@ func (g *MetricsGroup) Get() (metrics []Metric) {
 		metrics = append(metrics, m[i].copyMetric())
 	}
 	return metrics
+}
+
+func getClusterBucketsTotalMD() MetricDescription {
+	return MetricDescription{
+		Namespace: clusterMetricNamespace,
+		Subsystem: bucketsSubsystem,
+		Name:      total,
+		Help:      "Total number of buckets in the cluster",
+		Type:      gaugeMetric,
+	}
 }
 
 func getClusterCapacityTotalBytesMD() MetricDescription {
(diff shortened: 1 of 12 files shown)


docs/metrics/prometheus/list.md
```diff
@@ -4,113 +4,123 @@ Each metric includes a label for the server that calculated the metric. Each met
...
+| `minio_s3_requests_ttfb_seconds_distribution` | Distribution of the time to first byte across API calls.                                                        |
+| `minio_s3_traffic_received_bytes`             | Total number of s3 bytes received.                                                                              |
+| `minio_s3_traffic_sent_bytes`                 | Total number of s3 bytes sent.                                                                                  |
+| `minio_software_commit_info`                  | Git commit hash for the MinIO release.                                                                          |
+| `minio_software_version_info`                 | MinIO Release tag for the server.                                                                               |
+| `minio_usage_last_activity_nano_seconds`      | Time elapsed (in nano seconds) since last scan activity.                                                        |
 
 
+# List of metrics exported per bucket level
```
shtripat, line 107: For some of the metrics like 
```
minio_bucket_requests_4xx_errors_total
minio_bucket_requests_inflight_total
minio_bucket_requests_total
minio_bucket_requests_ttfb_seconds_distribution
minio_bucket_traffic_sent_bytes
```
other than existing buckets in the instance, I see metrics for two additional buckets with names `api` and `probe-bucket-sign-xicgc15djbyz`. 
Are these inbuilt buckets? If so, do we need these metrics as well to be exposed/listed?

shtripat: This could be because some application is sending non-existent bucket names, so not an issue.

docs/metrics/prometheus/list.md
```diff
@@ -4,113 +4,123 @@ Each metric includes a label for the server that calculated the metric. Each met
...
+
+Each metric includes a label for the server that calculated the metric. Each metric has a label for the server that generated the metric. Each
+metric has a label that distinguishes the bucket.
+
+These metrics can be obtained from any MinIO server once per collection.
+
+| Name                                              | Description                                                                     |
+|:--------------------------------------------------|:--------------------------------------------------------------------------------|
+| `minio_bucket_objects_size_distribution`          | Distribution of object sizes in the bucket, includes label for the bucket name. |
```
shtripat, line 116: minio_bucket_usage_version_total
minio_bucket_objects_version_distribution

should be added to the list

docs/metrics/prometheus/list.md
harshavardhana: yes we capture all those buckets as well which come in @shtripat

harshavardhana: it is added to ensure that we capture whatever hits MinIO at bucket resource.

docs/metrics/prometheus/list.md
harshavardhana: added

shtripat (approved):
LGTM. Verified

harshavardhana:
This has a breaking change that affects Console UI as well @prakashsvmx @bexsoft and also our grafana dashboard which must be updated after this is merged

// cc @ravindk89

cmd/notification.go
```diff
@@ -1101,6 +1101,61 @@ func (sys *NotificationSys) GetBandwidthReports(ctx context.Context, buckets ...
...
+// GetBucketMetrics - gets the cluster level bucket metrics from all nodes excluding self.
+func (sys *NotificationSys) GetBucketMetrics(ctx context.Context) <-chan Metric {
+	if sys == nil {
+		return nil
+	}
+	g := errgroup.WithNErrs(len(sys.peerClients))
+	peerChannels := make([]<-chan Metric, len(sys.peerClients))
+	for index := range sys.peerClients {
+		if sys.peerClients[index] == nil {
+			continue
```
vadmeste: I think you need to return an error here or at least ignore peersChannels[index] with a nil value before running the next goroutine

harshavardhana:
For Console UI you have to remove any `minio_bucket` view on the dashboard since `bucket` centric view has moved into its own scrape endpoint due to scale issues with customers having 1000's of buckets.

Instead Console UI dashboard can be a purely Cluster information view and rest of the bucket centric view can move into "bucket" centric, per bucket view.

Since the latter might require work my suggestion is deprecate the `minio_bucket*` first from the Prometheus dashboard in Console UI. Since the usage information is already available via `object browser`

cmd/notification.go
```diff
@@ -1101,6 +1101,61 @@ func (sys *NotificationSys) GetBandwidthReports(ctx context.Context, buckets ...
...
+		wg.Add(1)
+		go func(ctx context.Context, peerChannel <-chan Metric, wg *sync.WaitGroup) {
+			defer wg.Done()
+			for {
+				select {
+				case m, ok := <-peerChannel:
+					if !ok {
+						return
+					}
+					ch <- m
```
vadmeste: Do you want to add a ctx.Done() listener here so to avoid a leak if this function is not used correctly ?

cmd/notification.go
```diff
@@ -1101,6 +1101,61 @@ func (sys *NotificationSys) GetBandwidthReports(ctx context.Context, buckets ...
...
+				case <-ctx.Done():
+					return
+				}
+			}
+		}(ctx, peerChannels[index], &wg)
+	}
+	go func(wg *sync.WaitGroup, ch chan Metric) {
+		wg.Wait()
+		close(ch)
+	}(&wg, ch)
```
vadmeste, line 1161: It doesn't seem this goroutine is needed and wg are needed, you can simply add 'defer close(ch)' in the previous goroutine

cmd/metrics-v2.go
```diff
@@ -40,14 +40,15 @@ import (
...
+	nodeCollector           *minioNodeCollector
+	clusterCollector        *minioClusterCollector
+	bucketCollector         *minioBucketCollector
+	peerMetricsGroups       []*MetricsGroup
+	bucketPeerMetricsGroups []*MetricsGroup
 )
 
 func init() {
 	clusterMetricsGroups := []*MetricsGroup{
-		getBucketUsageMetrics(),
```
vadmeste, line 50: This looks like a breaking change, is it necessary to remove this one ?

harshavardhana:
Let's get Console UI release aligned with this? @bexsoft @prakashsvmx

vadmeste (commented):
Few comments, not tested yet

cmd/notification.go
harshavardhana: No this is not returning any error @vadmeste this is simply copy paste from GetClusterMetrics()

cmd/notification.go
harshavardhana: Copy paste from GetClusterMetrics() but it can be added to avoid blocking.

cmd/notification.go
harshavardhana: that activity will do later this is a copy paste from GetClusterMetrics()

cmd/metrics-v2.go
harshavardhana: Yes read the comments @vadmeste

cmd/notification.go
harshavardhana: and we also don't need to return errors for `nil` peerClients since it is always nil and we do not need to log it.

harshavardhana:
PTAL @vadmeste made relevant changes.

prakashsvmx:
@harshavardhana 

```
"Data Usage Growth"         ->  minio_bucket_usage_total_bytes
"Object size distribution"  ->  minio_bucket_objects_size_distribution
"Number of Buckets"         ->  minio_bucket_usage_total_bytes
"Number of Objects"         ->  minio_bucket_usage_object_total
```

these are the `minio_bucket*` metrics used in Console to display widgets.   We will plan to remove them. 
@bexsoft

harshavardhana:
> @harshavardhana
> 
> ```
> "Data Usage Growth"         ->  minio_bucket_usage_total_bytes
> "Object size distribution"  ->  minio_bucket_objects_size_distribution
> "Number of Buckets"         ->  minio_bucket_usage_total_bytes
> "Number of Objects"         ->  minio_bucket_usage_object_total
> ```
> 
> these are the `minio_bucket*` metrics used in Console to display widgets. We will plan to remove them. @bexsoft

They will be moved to `minio_cluster_usage` at the cluster endpoint @prakashsvmx so cluster dashboard can still show those values and use them.

harshavardhana:
Take a look at the new list in this PR @prakashsvmx at `list.md` in the docs directory

donatello (approved):
LGTM

cmd/notification.go
```diff
@@ -1101,19 +1103,78 @@ func (sys *NotificationSys) GetBandwidthReports(ctx context.Context, buckets ...
 	return consolidatedReport
 }
 
-// GetClusterMetrics - gets the cluster metrics from all nodes excluding self.
-func (sys *NotificationSys) GetClusterMetrics(ctx context.Context) <-chan Metric {
+// GetBucketMetrics - gets the cluster level bucket metrics from all nodes excluding self.
+func (sys *NotificationSys) GetBucketMetrics(ctx context.Context) <-chan Metric {
```
Praveenrajmani, line 1107:
```suggestion
func (sys *NotificationSys) GetPeerBucketMetrics(ctx context.Context) <-chan Metric {
```

as it only collects bucket metrics from peers?

cmd/metrics-v2.go
```diff
@@ -82,16 +84,28 @@ func init() {
...
 		getDistLockMetrics(),
 	}
 
+	bucketMetricsGroups := []*MetricsGroup{
+		getBucketUsageMetrics(),
+		getHTTPMetrics(true),
+		getBucketTTFBMetric(),
+	}
+
+	bucketPeerMetricsGroups = []*MetricsGroup{
```
Praveenrajmani, line 101: Q: why do we need `bucketPeerMetricsGroups` as we already publish bucket peer metrics in [bucketCollector](https://github.com/minio/minio/pull/17663/files#diff-245d74540f439e91f424f4f2d44e1f3f9385197d33144c8fe67f58f4c2f7afd2R2902)? Also, we don't have a collector for `bucketPeerMetricsGroups`

cmd/notification.go
harshavardhana: Yeah we but we don't call it that way. It's all Peers do it is called a higher level name.

cmd/metrics-v2.go
harshavardhana: That is needed for Peer calls for remote nodes to collect. 

Look at how existing peerMetricsGroups work.

PR #17690, opened 2023-07-20
Get the signed headers in the order they were signed

ruspaul013:
## Description
Get the signed headers in the order they were signed not alphabetically

## Motivation and Context
MinIO should have the same behavior like S3.
The signature is computed twice in inconsistent ways:
1. AWS PHP SDK generates the signature using the given headers with no key manipulation.
2. MinIO (upon verifying the validity of the signature generated by the SDK) regenerates the signature AFTER sorting the keys alphabetically.

This minor inconsistency leads to unexpected, hard to trace behavior when using the aws s3 sdk for uploading files to minio with more than one signed header.
If the metadata used when signing an url, in aws-sdk for php, is not sorted alphabetically, the response to the request will be 403, because MinIO is sorting the signed headers alphabetically when verifying if the signatures match ( `<Error><Code>SignatureDoesNotMatch</Code><Message>The request signature we calculated does not match the signature you provided. Check your key and signing method.</Message>` ) 
Except when there is no metadata or only one property, in this case there is no problems.

This function will get the signed headers in the signed order.

## How to test this PR?
You need aws-sdk for php, create a small script were you generate an presigned url, with metadata, to upload a file and make a request to MinIO.
Here is an example of test file:
[minio-example.txt](https://github.com/minio/minio/files/12080889/minio-example.txt)
Try first with current version of minio and then with this version.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/signature-v4.go
+++ b/cmd/signature-v4.go
@@ -260,7 +260,7 @@ func doesPresignedSignatureMatch(hashedPayload string, r *http.Request, region s
 	// Construct the query.
 	query.Set(xhttp.AmzDate, t.Format(iso8601Format))
 	query.Set(xhttp.AmzExpires, strconv.Itoa(expireSeconds))
-	query.Set(xhttp.AmzSignedHeaders, getSignedHeaders(extractedSignedHeaders))
+	query.Set(xhttp.AmzSignedHeaders, strings.Join(pSignValues.SignedHeaders, ";"))
 	query.Set(xhttp.AmzCredential, cred.AccessKey+SlashSeparator+pSignValues.Credential.getScope())
 
 	defaultSigParams := set.CreateStringSet(


cmd/signature-v4.go
```diff
@@ -260,7 +276,7 @@ func doesPresignedSignatureMatch(hashedPayload string, r *http.Request, region s
 	// Construct the query.
 	query.Set(xhttp.AmzDate, t.Format(iso8601Format))
 	query.Set(xhttp.AmzExpires, strconv.Itoa(expireSeconds))
-	query.Set(xhttp.AmzSignedHeaders, getSignedHeaders(extractedSignedHeaders))
+	query.Set(xhttp.AmzSignedHeaders, getSignedHeadersInSignedOrder(extractedSignedHeaders, pSignValues.SignedHeaders))
```
harshavardhana: This could simply be `strings.Join(pSignValues.SignedHeaders, ";")` There is no value in re-doing the signed headers while going through the HTTP headers.

harshavardhana (changes requested):

cmd/signature-v4.go
ruspaul013: I thought about this but I wasn't sure if it would be the right method. I will make the changes.

harshavardhana: it is right, there is nothing new that is being added here.

harshavardhana: Signed headers are already validated, there is no point in regenerating them in the same order.

harshavardhana (approved):

ruspaul013:
In order to pass all the checks, what should I do? Why does Advanced Tests with Go 1.20.x fail?

vadmeste:
> In order to pass all the checks, what should I do? Why does Advanced Tests with Go 1.20.x fail?

It looks like some test fails occasionally and we need to fix the test itself. I restarted the check again. I will take a deeper look if it still fails.

harshavardhana:
Please provide an example of AWS SDK PHP, I will verify this with other AWS SDKs.

klauspost (approved):

ruspaul013:
@harshavardhana 
> Please provide an example of AWS SDK PHP, I will verify this with other AWS SDKs.

I attached a txt file in the PR, do you need something else?

harshavardhana:
@ruspaul013 please add a test case as well.

ruspaul013:
> @ruspaul013 please add a test case as well.

@harshavardhana I tried to add test for this one, but I could not figure it out how to create them so that I can test what I wrote. I was thinking to do the same as test (8) from  `func TestDoesPresignedSignatureMatch(t *testing.T)` but I saw that the existing tests are not failing so I thought is not necessary to create a new test case. Can you give me a hint on how to create one for this?

I tested locally by making a build and then using dlv to check the values.

PR #17697, opened 2023-07-20
interpret empty version as `null` to evolve with versioning config

harshavardhana:
## Description
interpret empty version as `null` to evolve with versioning config

## Motivation and Context
empty versions such as `null` are empty and `null` depending
on the bucket configuration. This can be problematic for 
count(versions) calculation as we may skip valid versions.

We may also skip passing the correct version in ILM via 
scanner. Always treat buckets as versioned to ensure that 
un-versioned bucket versions are treated as `null`.

## How to test this PR?
very hard, you would need to create a bucket that changes
its configuration while the ILM rule is being applied,
such out-of-band changes are possible. These end up affecting
the ILM rule application.

## Types of changes

## Checklist:


harshavardhana:
PTAL @krisis @poornas

PR #17708, opened 2023-07-21
don't error when asked for 0-based range on empty objects

flisk:
In a reverse proxying setup, a proxy in front of minio may attempt to request objects in slices for enhanced cache efficiency. Since such a proxy cannot have prior knowledge of how large a requested resource is, it usually sends a header of the form:

        Range: 0-$slice_size

... and, depending on the size of the resource, expects either:

- an empty response, if $resource_size == 0
- a full response, if $resource_size <= $slice_size
- a partial response, if $resource_size > $slice_size

Prior to this change, minio would respond `416 Range Not Satisfiable` if a client tried to request a range on an empty resource. This behavior is technically consistent with [RFC9110][1] – however, it renders sliced reverse proxying, such as implemented in nginx, broken in the case of empty files. Nginx itself seems to break this convention to enable "useful" responses in these cases, and minio should probably do that too.

[1]: https://www.rfc-editor.org/rfc/rfc9110#byte.ranges

## How to test this PR?

Using e.g. curl, request ranges on objects served by a minio instance:

```
$ curl -v --range 0-1 http://localhost:12345/testbucket/empty
```

Without the patch applied, the above run against an empty (0-byte) object should fail with `416 Range Not Satisfiable`.

With the patch applied, an empty response should be returned.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/httprange.go
+++ b/cmd/httprange.go
@@ -59,6 +59,9 @@ func (h *HTTPRangeSpec) GetLength(resourceSize int64) (rangeLength int64, err er
 			rangeLength = resourceSize
 		}
 
+	case h.Start == 0 && resourceSize == 0:
+		rangeLength = resourceSize
+
 	case h.Start >= resourceSize:
 		return 0, errInvalidRange
 


harshavardhana (approved):

klauspost:
@harshavardhana This is wrong.

There is no valid way to specify offsets in a 0-byte file. Blame it on the [RFC](https://datatracker.ietf.org/doc/html/rfc7233#section-4.4)

> The 416 (Range Not Satisfiable) status code indicates that none of the ranges in the request's Range header field (Section 3.1) overlap the current extent of the selected resource...

A request for "bytes=0-" is a request for the first byte of a resource. If the resource is 0-length, the range [0,0] does not overlap the resource content and the server responds with an error.

flisk:
@klauspost would you be more amenable to a PR that mirrors nginx' behavior (ignoring the range header and instead serving a full response with a 200 code) in these cases? this behavior should be compliant, as servers are free to ignore range headers that don't make sense.

klauspost:
@flisk No. We do what S3 does - there is thousands of clients out there, and we cannot change behavior because it is convenient.

I suggest you write a module for nginx that removes `Range: bytes=0-` headers, if you don't want the functionality it implies, or just don't send them in the first place.

PR #17709, opened 2023-07-23
allow DNS cache TTL to be configurable

harshavardhana:
## Description
allow DNS cache TTL to be configurable

## Motivation and Context
this is added for now as a hidden variable

## How to test this PR?
There is not much to test here, this flag
is useful in rare circumstances where we
want to increase the default DNS cache TTL 
to a higher value on baremetal setups.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/common-main.go
+++ b/cmd/common-main.go
@@ -102,22 +102,6 @@ func init() {
 
 	globalIsCICD = env.Get("MINIO_CI_CD", "") != "" || env.Get("CI", "") != ""
 
-	// Call to refresh will refresh names in cache.
-	go func() {
-		// Baremetal setups set DNS refresh window to 10 minutes.
-		t := time.NewTicker(10 * time.Minute)
-		defer t.Stop()
-		for {
-			select {
-			case <-t.C:
-				globalDNSCache.Refresh()
-
-			case <-GlobalContext.Done():
-				return
-			}
-		}
-	}()
-
 	console.SetColor("Debug", fcolor.New())
 
 	gob.Register(StorageErr(""))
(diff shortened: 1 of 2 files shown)


klauspost (approved):
`cmd/common-main.go:455: File is not `gofumpt`-ed (gofumpt)` should be fixed, though.

cmd/common-main.go
```diff
@@ -443,6 +427,32 @@ func handleCommonCmdArgs(ctx *cli.Context) {
 	globalCertsCADir = &ConfigDir{path: filepath.Join(globalCertsDir.Get(), certsCADir)}
 
 	logger.FatalIf(mkdirAllIgnorePerm(globalCertsCADir.Get()), "Unable to create certs CA directory at %s", globalCertsCADir.Get())
+
+	// Check if we have configured a customer DNS cache TTL.
+	dnsTTL := ctx.GlobalDuration("dns-cache-ttl")
```
vadmeste: GlobalDuration is not needed because dns-cache-ttl is not defined as a global flag (as other many flags), it is not needed you can drop it.

cmd/common-main.go
```diff
@@ -443,6 +427,32 @@ func handleCommonCmdArgs(ctx *cli.Context) {
...
+	if dnsTTL <= 0 {
+		dnsTTL = ctx.Duration("dns-cache-ttl")
+	}
+	if dnsTTL <= 0 {
+		dnsTTL = 10 * time.Minute
```
vadmeste, line 434: there is already a default value when defining the flag

vadmeste (commented):
Few comments, LGTM otherwise

cmd/common-main.go
harshavardhana: It is when someone sets 0ms

cmd/common-main.go
harshavardhana: that is a code to make sure if the flag gets moved around this is always honored, since this function is specifically for commonFlags()

cmd/common-main.go
```diff
@@ -443,6 +427,31 @@ func handleCommonCmdArgs(ctx *cli.Context) {
 	globalCertsCADir = &ConfigDir{path: filepath.Join(globalCertsDir.Get(), certsCADir)}
 
 	logger.FatalIf(mkdirAllIgnorePerm(globalCertsCADir.Get()), "Unable to create certs CA directory at %s", globalCertsCADir.Get())
+
+	// Check if we have configured a customer DNS cache TTL.
```
harshavardhana:
```suggestion
	// Check if we have configured a custom DNS cache TTL.
```

cmd/common-main.go
vadmeste: > that is a code to make sure if the flag gets moved around this is always honored, since this function is specifically for commonFlags()

It doesn't work @harshavardhana, I already tested it. For this to work, --dns-cache-ttls needs to be define when defining cli.App

harshavardhana: correct it is a defensive code

vadmeste (approved):
LGTM

PR #17717, opened 2023-07-24
Use decrypted object size while computing object size summary

poornas:
Corrects an issue with encrypted versioned objects being reported under `unversioned` bin in the object version histogram

## Motivation and Context
Correctness

## How to test this PR?
on a versioned bucket, turn on default encryption with `mc encrypt set sse-s3..`. 
Upload a version and check prometheus metrics `minio_bucket_objects_version_distribution` - it shows 0 versions with master because of incorrect size accounting

## Types of changes

## Checklist:


krisis (approved):
Nice catch!

harshavardhana (approved):

klauspost:
@poornas Applying ILM (for example deleting) may change the size, so what you are doing will do wrong accounting then.

Please change so downstream returns the correct value instead.

PR #17718, opened 2023-07-25
feat: support perf client to show `TX` from client to server

jiuker:
support_perf_client

## Description

related pr 
https://github.com/minio/mc/pull/4645
https://github.com/minio/madmin-go/pull/212

## Motivation and Context

to `support perf client`.
To show `TX` from client to server

## How to test this PR?

```shell
#> .\mc support perf client minio

Client: ✔

ENDPOINT                TX
http://127.0.0.1:9000   269 MiB/s
```

## Types of changes

## Checklist:

### DIFF
--- a/cmd/admin-handlers.go
+++ b/cmd/admin-handlers.go
@@ -41,6 +41,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync/atomic"
 	"time"
 
 	"github.com/dustin/go-humanize"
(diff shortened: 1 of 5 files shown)


cmd/perf-tests.go
```diff
@@ -422,3 +422,18 @@ func siteNetperf(ctx context.Context, duration time.Duration) madmin.SiteNetPerf
...
 }
+
+func clientPerf(ctx context.Context) madmin.ClientPerfResult {
+	for {
+		if globalSiteNetPerfRX.ActiveConnections() == 0 || contextCanceled(ctx) {
+			break
+		}
+		time.Sleep(time.Second)
+	}
+	rx := globalSiteNetPerfRX.RXSample
```
harshavardhana: We shouldn't reuse this, use a different global value

harshavardhana (changes requested):

klauspost:
1MB of data is not going to give you any real indication of speed.

The overhead of request processing and the overall network latency will dominate way too much.

TCP connections tends to "speed up", meaning they start out slow and then ramp up as they see the host can receive more data.  A 1MB packet will not be close to enough for it to ramp up to full speed.

cmd/admin-handlers.go
```diff
@@ -1172,6 +1172,37 @@ func (a adminAPIHandlers) SitePerfHandler(w http.ResponseWriter, r *http.Request
 	}
 }
 
+// ClientDevNull - everything goes to io.Discard
+// [POST] /minio/admin/v3/speedtest/client/devnull
+func (a adminAPIHandlers) ClientDevNull(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+
+	objectAPI, _ := validateAdminReq(ctx, w, r, iampolicy.HealthInfoAdminAction)
```
harshavardhana: We should introduce perhaps another action here since this is easy to abuse. 

Also have an upper limit of 100G data transfer and also have a global atomic boolean to reject if two requests are sent to same server of ClientDevNull. 

We only allow 1 at a time.

harshavardhana (changes requested):

cmd/admin-handlers.go
```diff
@@ -1172,6 +1172,46 @@ func (a adminAPIHandlers) SitePerfHandler(w http.ResponseWriter, r *http.Request
...
+	defer nsLock.Unlock(lkctx)
+
+	ctx, cancel := context.WithTimeout(ctx, madmin.MaxClientPerfTimeout)
+	defer cancel()
+	totalRx := int64(0)
+	connectTime := time.Now()
+	for {
+		n, err := io.CopyN(io.Discard, r.Body, 128*humanize.KiByte)
+		if err != nil && err != io.EOF && err != io.ErrUnexpectedEOF {
+			// If there is a disconnection before globalNetPerfMinDuration (we give a margin of error of 1 sec)
```
harshavardhana: this comment doesn't make sense here..

cmd/admin-handlers.go
```diff
@@ -1172,6 +1172,46 @@ func (a adminAPIHandlers) SitePerfHandler(w http.ResponseWriter, r *http.Request
...
+	if objectAPI == nil {
+		return
+	}
+
+	nsLock := objectAPI.NewNSLock(minioMetaBucket, "client-perf")
+	lkctx, err := nsLock.GetLock(ctx, globalOperationTimeout)
+	if err != nil {
+		writeErrorResponseJSON(ctx, w, errorCodes.ToAPIErr(toAPIErrorCode(ctx, err)), r.URL)
+		return
+	}
```
harshavardhana, line 1208:
```suggestion
	}
	ctx = lkctx.Context()
```

cmd/admin-handlers.go
```diff
@@ -1172,6 +1172,46 @@ func (a adminAPIHandlers) SitePerfHandler(w http.ResponseWriter, r *http.Request
...
+	defer nsLock.Unlock(lkctx)
```
klauspost: It is a problem that this is actually part of the benchmark. Return the time taken from the lock was acquired until the body completed as a header.

jiuker: Yeah. It is

harshavardhana (changes requested):

harshavardhana (approved):

harshavardhana:
PTAL @shtripat

shtripat (commented):
Changes look fine.  Once verified with `mc` changes, can be approved.

jiuker:
> Changes look fine. Once verified with `mc` changes, can be approved.

You can verify now. Thanks

shtripat (approved):
LGTM. Verified with https://github.com/minio/mc/pull/4645

harshavardhana:
> LGTM. Verified with [minio/mc#4645](https://github.com/minio/mc/pull/4645)
> 
> ```
> ./mc support perf client myminio --airgap --dev
> 
> Client: ✔ 
> 
> ENDPOINT        TX               
>                 230 MiB/s
> mc: MinIO performance report saved at myminio-perf_20230810115721.zip, please upload to SUBNET portal manually
> ```

Looks like endpoint is missing in this output @jiuker is it a client issue?

jiuker:
@harshavardhana I will check it

PR #17701, opened 2023-07-21
Add slow drive timeouts to match with active disk monitoring

harshavardhana:
## Description
Add slow drive timeouts to match with active disk monitoring

## Motivation and Context
allow active disk monitoring to be configurable, and 
use these to add deadlines in various call layers for 
various syscalls.

## How to test this PR?
Will outline this soon, it requires the use of 

```
xfs_freeze -f /mount-point
```
 
To simulate a disk hang. 

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Optimization (provides speedup with no functional changes)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [ ] Fixes a regression (If yes, please add `commit-id` or `PR #` here)
- [ ] Unit tests added/updated
- [ ] Internal documentation updated
- [ ] Create a documentation update request [here](https://github.com/minio/docs/issues/new?label=doc-change,title=Doc+Updated+Needed+For+PR+github.com%2fminio%2fminio%2fpull%2fNNNNN)

PR #17724, opened 2023-07-26
introduce deadlines on READ operations

harshavardhana:
## Description
introduce deadlines on READ operations

## Motivation and Context
similar in spirit to the #17701

## How to test this PR?
Testing this is a bit harder, it requires
mocking of disks. `xfs_freeze` is not useful
since that only disallows WRITE operations.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/xl-storage-disk-id-check.go
+++ b/cmd/xl-storage-disk-id-check.go
@@ -538,7 +538,15 @@ func (p *xlStorageDiskIDCheck) ReadVersion(ctx context.Context, volume, path, ve
 	}
 	defer done(&err)
 
-	return p.storage.ReadVersion(ctx, volume, path, versionID, readData)
+	w := xioutil.NewDeadlineWorker(diskMaxTimeout)
+	rerr := w.Run(func() error {
+		fi, err = p.storage.ReadVersion(ctx, volume, path, versionID, readData)
+		return err
+	})
+	if rerr != nil {
+		return fi, rerr
+	}
+	return fi, err
 }
 
 func (p *xlStorageDiskIDCheck) ReadAll(ctx context.Context, volume string, path string) (buf []byte, err error) {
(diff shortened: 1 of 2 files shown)


cmd/xl-storage-disk-id-check.go
```diff
@@ -466,11 +466,15 @@ func (p *xlStorageDiskIDCheck) DeleteVersions(ctx context.Context, volume string
 	}
 	defer done(&err)
 
-	errs = p.storage.DeleteVersions(ctx, volume, versions)
+	w := xioutil.NewDeadlineWorker(diskMaxTimeout)
+	rerr := w.Run(func() error {
+		errs = p.storage.DeleteVersions(ctx, volume, versions)
```
klauspost: This is a batch call, so I think this should be applied at the lower level for each operation.

Since we know the batch size a lesser optimal version would be to multiply by the number of versions. Wouldn't recommend.

cmd/xl-storage-disk-id-check.go
```diff
@@ -568,34 +596,44 @@ func (p *xlStorageDiskIDCheck) StatInfoFile(ctx context.Context, volume, path st
 	}
 	defer done(&err)
 
-	return p.storage.StatInfoFile(ctx, volume, path, glob)
+	w := xioutil.NewDeadlineWorker(diskMaxTimeout)
+	rerr := w.Run(func() error {
+		stat, err = p.storage.StatInfoFile(ctx, volume, path, glob)
```
klauspost: Again, batch call - the glob can expand to thousands of entries. Feel free to add at a lower lever.

cmd/xl-storage-disk-id-check.go
```diff
@@ -568,34 +596,44 @@ func (p *xlStorageDiskIDCheck) StatInfoFile(ctx context.Context, volume, path st
...
+func (p *xlStorageDiskIDCheck) ReadMultiple(ctx context.Context, req ReadMultipleReq, resp chan<- ReadMultipleResp) (err error) {
 	ctx, done, err := p.TrackDiskHealth(ctx, storageMetricReadMultiple, req.Bucket, req.Prefix)
 	if err != nil {
 		close(resp)
 		return err
 	}
 	defer done(&err)
 
-	return p.storage.ReadMultiple(ctx, req, resp)
+	w := xioutil.NewDeadlineWorker(diskMaxTimeout)
```
klauspost: Batch call.

klauspost (approved):
LGTM

PR #17693, opened 2023-07-20
Check if metadata headers/url values are equal with signed headers

ruspaul013:
## Description
Check if metadata headers/url values are equal with signed headers

## Motivation and Context
MinIO should have the same behavior like S3. 
As it was mention in this isssue #16917, MinIO will upload the file with the metadata header send in the request, even if this is not used when generating the signed URL, while AWS will return an error.

## How to test this PR?
Follow steps from #16917, with current version of minio and then with this version.

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Optimization (provides speedup with no functional changes)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [ ] Fixes a regression (If yes, please add `commit-id` or `PR #` here)
- [ ] Unit tests added/updated
- [ ] Internal documentation updated
- [ ] Create a documentation update request [here](https://github.com/minio/docs/issues/new?label=doc-change,title=Doc+Updated+Needed+For+PR+github.com%2fminio%2fminio%2fpull%2fNNNNN)

Issue #16917, opened 2023-03-29
No error at upload with HTTP headers that were not signed


AndreeaCapatina:
I tried to upload a file to minio server using the pre signed URL obtained with php s3client. Nothing was provided to the Metadata option of the s3client. 
When doing the upload, however,  I added a metadata header `X-Amz-Meta-A: B`.  I notice a different behaviour between minio and AWS in this scenario. 

## Expected Behavior

Minio and AWS should be consistent. Minio should return an error if metadata headers are present in the request but not when generating the signed URL.  
## Current Behavior

AWS returns an error if the upload request contains metadata headers that were not signed. 
With minio, the file is uploaded and the metadata are saved. 

## Steps to Reproduce (for bugs)

1.Make an API call on both AWS and minio to get an upload pre signed URI, like so (here using PHP):
```
$cmd = $client->getCommand('PutObject', [
    'Bucket' => 'mybucket',
    'Key'    => 'image.jpg'
]);

$response = $client->createPresignedRequest($cmd, '+60 minutes')->withMethod('PUT');
$uri = $response->getUri();
```
As we can see, I didn't added any metadata to be signed.
2.  Make an upload request, with at least one metadata header. Here is a simple curl example I used to upload a file:
```
$curl = curl_init();

curl_setopt($curl, CURLOPT_URL,  $uri);

$headers[] = 'X-Amz-Meta-A: B';
curl_setopt($curl, CURLOPT_HTTPHEADER, $headers);

curl_setopt($curl, CURLOPT_SSL_VERIFYPEER, false);
curl_setopt($curl, CURLOPT_SSL_VERIFYHOST, false);
curl_setopt($curl, CURLOPT_RETURNTRANSFER, true);
curl_setopt($curl, CURLOPT_FOLLOWLOCATION, true);
curl_setopt($curl, CURLOPT_PUT, true);

$content = fopen($filePath, 'rb');
curl_setopt($curl, CURLOPT_INFILE, $content);
curl_setopt($curl, CURLOPT_INFILESIZE, filesize($filePath));

$responseBody = curl_exec($curl);
```
3.  For AWS, you should receive an error:
```
<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>AccessDenied</Code><Message>There were headers present in the request which were not signed</Message><HeadersNotSigned>x-amz-meta-a</HeadersNotSigned></Error>
```
4. Minio currently doesn't return any error and the metadata are saved. 
```
[...]

PR #17737, opened 2023-07-27
check if metadata headers/url values are equal with signed headers

ruspaul013:
## Description
Check if metadata headers/url values are equal with signed headers

## Motivation and Context
Open a new PR  ( #17693 - closed )
MinIO should have the same behavior like S3.
As it was mention in this isssue https://github.com/minio/minio/issues/16917, MinIO will upload the file with the metadata header send in the request, even if this is not used when generating the signed URL, while AWS will return an error.

## How to test this PR?
Follow steps from https://github.com/minio/minio/issues/16917, with current version of minio and then with this version.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/signature-v4-utils.go
+++ b/cmd/signature-v4-utils.go
@@ -260,3 +260,28 @@ func signV4TrimAll(input string) string {
 	// unicode.IsSpace() internally here) to one space and return
 	return strings.Join(strings.Fields(input), " ")
 }
+
+// checkMetaHeaders will check if the metadata from header/url is the same with the one from signed headers
+func checkMetaHeaders(signedHeadersMap http.Header, r *http.Request) APIErrorCode {
+	// check values from http header
+	for k, val := range r.Header {
+		if stringsHasPrefixFold(k, "X-Amz-Meta-") {
+			if signedHeadersMap.Get(k) == val[0] {
+				continue
+			}
+			return ErrUnsignedHeaders
+		}
+	}
+
+	// check values from url, if no http header
+	for k, val := range r.Form {
+		if stringsHasPrefixFold(k, "x-amz-meta-") {
+			if signedHeadersMap.Get(http.CanonicalHeaderKey(k)) == val[0] {
+				continue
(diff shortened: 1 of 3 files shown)


ruspaul013:
Hello @jiuker , @klauspost 
Sorry for the inconvenience about last the last PR (#17693), I opened a new one to simplify everything.

klauspost (approved):

cmd/signature-v4-utils.go
```diff
@@ -260,3 +260,28 @@ func signV4TrimAll(input string) string {
...
+			if signedHeadersMap.Get(k) == val[0] {
+				continue
+			}
+			return ErrUnsignedHeaders
+		}
+	}
+
+	// check values from url, if no http header
+	for k, val := range r.Form {
+		if stringsHasPrefixFold(k, "x-amz-meta-") {
```
jiuker, line 278:
```suggestion
		if stringsHasPrefixFold(k, "X-Amz-Meta-") {
```

jiuker (approved):

cmd/signature-v4-utils.go
klauspost: @jiuker We want the most likely case (since it is slightly faster). I believe in forms, the most likely case is lower case, while headers are header capitalized.

So AFAICT this matches the capitalization we expect.

jiuker:
@ruspaul013 Can you change where these three files start? We want it to be 2023.  
`// Copyright (c) 2015-2023 MinIO, Inc.`

cmd/signature-v4-utils.go
ruspaul013: it is ok to let it like this? I know that the url values have the prefix like this `x-amz-meta-` and I saw that the `stringsHasPrefixFold` function is ignoring case. I would like to leave them like this so that someone knows when they look where the value comes from.

jiuker (approved):

harshavardhana (approved):

PR #17745, opened 2023-07-28
add ReadFileStream deadline for disk call

harshavardhana:
## Description
add ReadFileStream deadline for disk call

## Motivation and Context
timeout the reader side if hung via disk max timeout

## How to test this PR?
Not easy, needs mock code. 

## Types of changes

## Checklist:

### DIFF
--- a/internal/ioutil/ioutil.go
+++ b/internal/ioutil/ioutil.go
@@ -73,7 +73,54 @@ type ioret struct {
 	err error
 }
 
-// DeadlineWriter deadline writer with context
+// DeadlineReader deadline reader with timeout
+type DeadlineReader struct {
+	io.ReadCloser
+	timeout time.Duration
+	err     error
+}
+
+// NewDeadlineReader wraps a writer to make it respect given deadline
+// value per Write(). If there is a blocking write, the returned Reader
+// will return whenever the timer hits (the return values are n=0
+// and err=context.DeadlineExceeded.)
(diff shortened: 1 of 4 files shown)


internal/ioutil/ioutil.go
```diff
@@ -73,7 +73,54 @@ type ioret struct {
...
+
+	select {
+	case res := <-c:
+		if !t.Stop() {
+			<-t.C
+		}
+		r.err = res.err
+		return res.n, res.err
+	case <-t.C:
+		r.ReadCloser.Close()
```
vadmeste, line 112: Since this is a very low level, I don't recommend closing this here, let's keep the one done in Close()

Golang *os.File.Close() returns 'file already closed' if it is called the second time.

harshavardhana: why? so that we can leak fd's? @vadmeste

harshavardhana: We should close here, why are you saying otherwise?

vadmeste: > why? so that we can leak fd's? @vadmeste

umm normally it will not be leaked because NewDeadlineReader() returns io.ReadCloser and I think the idometic way to use it is similar to this:
```
f, err := os.Open(filename)
if err != nil {
    panic(err)
}
defer f.Close()
f.Read()...
```

but yeah I see now that we never check for the returned error from f.Close()

vadmeste (approved):
LGTM

krisis (approved):

PR #17747, opened 2023-07-28
logger: Avoid slow calls in http logger Send() function

vadmeste:
## Description
Send() is synchronous and can affect the latency of S3 requests 
when the logger buffer is full.

Avoid checking if the http target is online or not and increase the 
workers anyway since the buffer is already full.

Also avoid logs flooding when the audit target is down.

## Motivation and Context
Address a possible outage when the audit is down and the logger http target has a high latency
Address reduced IO when the http target has a high latency and the logger buffer is always full

## Types of changes

## Checklist:

### DIFF
--- a/internal/logger/audit.go
+++ b/internal/logger/audit.go
@@ -28,7 +28,6 @@ import (
 	"github.com/minio/minio/internal/mcontext"
 	"github.com/minio/pkg/logger/message/audit"
 
-	"github.com/minio/madmin-go/v3"
 	xhttp "github.com/minio/minio/internal/http"
 )
 
@@ -145,7 +144,7 @@ func AuditLog(ctx context.Context, w http.ResponseWriter, r *http.Request, reqCl
 	// Send audit logs only to http targets.
 	for _, t := range auditTgts {
 		if err := t.Send(ctx, entry); err != nil {
-			LogAlwaysIf(context.Background(), fmt.Errorf("event(%v) was not sent to Audit target (%v): %v", entry, t, err), madmin.LogKindAll)
+			LogOnceIf(ctx, fmt.Errorf("Unable to send an audit event to the target `%v`: %v", t, err), "send-audit-event-failure")
 		}
 	}
 }
--- a/internal/logger/target/http/http.go
+++ b/internal/logger/target/http/http.go
@@ -43,7 +43,7 @@ const (
 	// Timeout for the webhook http call
 	webhookCallTimeout = 5 * time.Second
 
-	// maxWorkers is the maximum number of concurrent operations.
+	// maxWorkers is the maximum number of concurrent http loggers
 	maxWorkers = 16
 
 	// the suffix for the configured queue dir where the logs will be persisted.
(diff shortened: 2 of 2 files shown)


internal/logger/audit.go
```diff
@@ -145,7 +144,7 @@ func AuditLog(ctx context.Context, w http.ResponseWriter, r *http.Request, reqCl
 	// Send audit logs only to http targets.
 	for _, t := range auditTgts {
 		if err := t.Send(ctx, entry); err != nil {
-			LogAlwaysIf(context.Background(), fmt.Errorf("event(%v) was not sent to Audit target (%v): %v", entry, t, err), madmin.LogKindAll)
+			LogOnceIf(ctx, fmt.Errorf("Unable to send an audit event to the target `%v`: %v", t, err), "send-audit-event-failure")
```
harshavardhana, line 147: We are logging missed events such that they are not lost, if we do LogOnceIf() then we lose such logs.

internal/logger/target/http/http.go
```diff
@@ -361,32 +361,29 @@ func (h *Target) SendFromStore(key string) (err error) {
...
 
-// Send log message 'e' to http target.
-// If servers are offline messages are queued until queue is full.
+// Send the log message 'entry' to the http target.
+// Messages are queued in the disk if the store is enabled
 // If Cancel has been called the message is ignored.
 func (h *Target) Send(ctx context.Context, entry interface{}) error {
+	if atomic.LoadInt32(&h.status) == statusClosed {
+		return nil
+	}
```
harshavardhana, line 370: moving this here means that we won't be able to store logs if the remote status is offline or closed.

harshavardhana (changes requested):

internal/logger/audit.go
vadmeste: > We are logging missed events such that they are not lost, if we do LogOnceIf() then we lose such logs.

I understand, but audit is really verbose, these are a lot of events. This will also flood the http logger. I am not sure if we should care about printing this audit in the log.

harshavardhana: yeah its a trade-off we don't know if we should preserve or lose these events.

internal/logger/target/http/http.go
vadmeste: > moving this here means that we won't be able to store logs if the remote status is offline or closed.

well, if it is closed, then we don't have to store anything (closed means we explicity asked for the logger to stop working)

harshavardhana: this is also preserved in case of situations where queue_dir is set and target is closed, we can re-enable it and it would requeue all of the past events.

vadmeste: > this is also preserved in case of situations where queue_dir is set and target is closed, we can re-enable it and it would requeue all of the past events.

What do you mean by target closed here ? in the close the state closed is only when you call target.Cancel() (there is another state which is called offline which has another meaning)

harshavardhana: correct target can be renabled later after a Cancel() @vadmeste the code is present to ensure we always preserve events.

harshavardhana: It can argued but this is breaking something that someone might be already relying on.

harshavardhana (approved):

harshavardhana (approved):

PR #17763, opened 2023-07-31
fix: optimize DiskInfo() call avoid metrics when not needed

harshavardhana:
## Description
fix: optimize DiskInfo() call avoid metrics when not needed

## Motivation and Context
DiskInfo() is frequently used, optimize it

## How to test this PR?
Mostly via making multiple calls this PR simply optimizes

- looking up disk-level metrics when its not useful to the caller
- looking up if disk is HDD or SSD when its already known

## Types of changes

## Checklist:

### DIFF
--- a/cmd/storage-rest-client.go
+++ b/cmd/storage-rest-client.go
@@ -272,7 +272,7 @@ func (client *storageRESTClient) SetDiskID(id string) {
 }
 
 // DiskInfo - fetch disk information for a remote disk.
-func (client *storageRESTClient) DiskInfo(_ context.Context) (info DiskInfo, err error) {
+func (client *storageRESTClient) DiskInfo(_ context.Context, metrics bool) (info DiskInfo, err error) {
 	if !client.IsOnline() {
 		// make sure to check if the disk is offline, since the underlying
 		// value is cached we should attempt to invalidate it if such calls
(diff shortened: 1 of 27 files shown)


poornas (approved):

cmd/storage-rest-client.go
```diff
@@ -287,7 +287,10 @@ func (client *storageRESTClient) DiskInfo(_ context.Context) (info DiskInfo, err
 			var info DiskInfo
 			ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
 			defer cancel()
-			respBody, err := client.call(ctx, storageRESTMethodDiskInfo, nil, nil, -1)
+
+			vals := make(url.Values)
+			vals.Set(storageRESTMetrics, strconv.FormatBool(metrics))
```
klauspost, line 292: @harshavardhana The first call then decides (forever) if metrics are fetched. I will fix this for the grid implementation.

harshavardhana: not forever right, it will be for 1sec @klauspost

klauspost: @harshavardhana It is captured in the `sync.Once` closure, so whatever it was then, it will remain that.

PR #17783, opened 2023-08-02
fix: prioritize audit_webhook and logger_webhook ENVs over the config KVS

Praveenrajmani:
## Description

Currently, the ENV setting will not override the corresponding config setting for audit and logger webhooks.

NOTE: This will start honoring the ENV settings over the config settings. If user has a dummy ENV set for audit and logger webhooks, it will override the corresponding kvs config from now.

## Motivation and Context

make sure that the ENV setting always takes the precedence over config KVS

## How to test this PR?
- Configure audit_webook or logger_webhook using `mc admin config set`
- Set an ENV overriding the audit or logger webhook config
- The ENV setting should be honored

## Types of changes

## Checklist:

### DIFF
--- a/cmd/config-current.go
+++ b/cmd/config-current.go
@@ -386,7 +386,7 @@ func validateSubSysConfig(ctx context.Context, s config.Config, subSys string, o
 		}
 	default:
 		if config.LoggerSubSystems.Contains(subSys) {
-			if err := logger.ValidateSubSysConfig(s, subSys); err != nil {
+			if err := logger.ValidateSubSysConfig(ctx, s, subSys); err != nil {
 				return err
 			}
 		}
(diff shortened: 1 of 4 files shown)


harshavardhana:
> This PR also fixes the empty requests sent to the webhooks to check if the target is alive, Instead uses tcp dials.

A TCP dial might not work properly if your webhook endpoint is behind a load balancer, it may not know if the target is alive or not. 

Since the entire HTTP transit has not happened. The HTTP based is used to ensure that the entire proxy stack propagation happens.

harshavardhana (changes requested):
I think you should remove the checkAlive() change and rest are fine for this PR.

harshavardhana (approved):

PR #17784, opened 2023-08-02
Add paramaters in Helm chart to load OIDC clientSecret from Secret Resource

alikhtag:
## Description
Add ability in helm chart to load oidc client secret from kubernetes secret resource. 

## Motivation and Context
One would want to load this OIDC secret from Kubernetes secret resource, so it would be nice to have this option in helm chart. This is useful if one has secret operators to load secrets from keyvault into kubernetes secret resources. 
Tried to name paramater to something similar way in helm chart. 

## How to test this PR?
One can run helm template and see that `statefulSet` and `Deployment` are templated correctly, also feel free to do test deployment
`helm template helm/minio/ --set oidc.enabled=true --set oidc.existingClientSecretKey="testKey" --set oidc.existingClientSecretName="testName"`
and too see that if `clientSecret` is set, `clientSecret` takes precedence
`helm template helm/minio/ --set oidc.enabled=true --set oidc.existingClientSecretKey="testKey" --set oidc.existingClientSecretName="testName" --set oidc.clientSecret="IamFirst"`

## Types of changes

## Checklist:


jiuker (changes requested):
We should set the `secret` name when that's not empty. Not for `else`

alikhtag:
> We should set the `secret` name when that's not empty. Not for `else`

ok will change it, thanks!

jiuker (approved):

Issue #17768, opened 2023-08-01
Minio healing slowing down over time


Jazz9:
When replacing a failed server, the healing process starts off fast but drops to a slow speed. The cpu, network and drives in the cluster are mostly idle.

![image](https://github.com/minio/minio/assets/39378577/26960092-320f-4626-ad5e-32d7dfd40e32)

![image](https://github.com/minio/minio/assets/39378577/73711e1f-83bc-472b-a29f-6413ca279b1c)

I would expect that the healing process would use all available disk resources. At this rate, the replacement of a full host will take months.

I have tried increasing the heal workers, heal max IO, heal sleep to 0 and the scanner speed faster or slower. None of these variables fix the issue.

## Steps to Reproduce (for bugs)
Stop minio on one host
reformat all drives on host
set folder permissions
start minio

## Your Environment
minio.RELEASE.2023-06-29T05-12-28Z
ubuntu 22 5.15.0-58-generic
6 minio hosts, 28 x 8tb XFS drives per host.

harshavardhana:
@Jazz9 can you `kill -ABRT $(pidof minio)` on the replaced host and share the generated backtrace?

Jazz9:
[backtrace.log](https://github.com/minio/minio/files/12225813/backtrace.log)

Unfortunately, I had just rebooted the minio process to see if it would speed up. I've uploaded the backtrace but I think it will be of limited use.

AlexZIX:
Minio really healing data to the replaced drive very slow. I have cluster of 4 hosts with 4 drives on each. Each drive filled with 1 TB of data for now. Yesterday I've replaced one drive and healing process started filling it with data. After 1 day it's filled with 40 GB only. So healing process will finish in 25 days I think.

[I've asked](https://github.com/minio/minio/issues/14198) about it above year ago and this issue now marked "work as intended".

Jazz9:
Objects Healed: 798,408, 161 GiB (9.9%)
Objects Failed: 0
Heal rate: 7 obj/s, 1.5 MiB/s
Estimated Completion: 1 week 

with 7.8 million objects left at 7 objects per second, ETA 81 days left.

@harshavardhana - is there anything else that I can check or log that would assist you?

Jazz9:
After reading the code - I wonder if the heal folder cache is growing over time and lookups are slowing down. Only a guess - i'll continue to read the code and learn more about the process.

@AlexZIX - what kind data did you have on the cluster? Was there lots of 'folders'

[...]

PR #17788, opened 2023-08-02
allow healing to prefer local disks over remote

harshavardhana:
## Description
allow healing to prefer local disks over remote

## Motivation and Context
optimization of healing, refer #17768 

## How to test this PR?
You need a setup with 4 servers and 4 disks in a 16-stripe setup.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/erasure-decode.go
+++ b/cmd/erasure-decode.go
@@ -289,12 +289,15 @@ func (e Erasure) Decode(ctx context.Context, writer io.Writer, readers []io.Read
 }
 
 // Heal reads from readers, reconstruct shards and writes the data to the writers.
-func (e Erasure) Heal(ctx context.Context, writers []io.Writer, readers []io.ReaderAt, totalLength int64) (derr error) {
+func (e Erasure) Heal(ctx context.Context, writers []io.Writer, readers []io.ReaderAt, totalLength int64, prefer []bool) (derr error) {
 	if len(writers) != e.parityBlocks+e.dataBlocks {
 		return errInvalidArgument
 	}
 
 	reader := newParallelReader(readers, e, 0, totalLength)
+	if len(readers) == len(prefer) {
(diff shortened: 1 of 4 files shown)


cmd/erasure-decode.go
```diff
@@ -289,12 +289,15 @@ func (e Erasure) Decode(ctx context.Context, writer io.Writer, readers []io.Read
...
 
 // Heal reads from readers, reconstruct shards and writes the data to the writers.
-func (e Erasure) Heal(ctx context.Context, writers []io.Writer, readers []io.ReaderAt, totalLength int64) (derr error) {
+func (e Erasure) Heal(ctx context.Context, writers []io.Writer, readers []io.ReaderAt, totalLength int64, prefer []bool) (derr error) {
 	if len(writers) != e.parityBlocks+e.dataBlocks {
 		return errInvalidArgument
 	}
 
 	reader := newParallelReader(readers, e, 0, totalLength)
+	if len(readers) == len(prefer) {
```
jiuker, line 298: There is always true. That's on purpose?
```go
readers := make([]io.ReaderAt, len(latestDisks))
prefer := make([]bool, len(latestDisks))
```

jiuker (changes requested):

cmd/erasure-decode.go
harshavardhana: yes it checks if the input `prefer` is non-nil and it matches number of readers.  @jiuker

jiuker: @harshavardhana I mean We should check the `prefer`'s `true` length?

harshavardhana: no not really @jiuker it must be same as available readers if its not then we do not enable preferential reading.

jiuker (approved):

klauspost (approved):
LGTM

PR #17792, opened 2023-08-03
Implement batch expiration for permanent deletes of expired objects

harshavardhana:
## Description
Implement batch expiration for permanent deletes of expired objects

## Motivation and Context
```yaml
expire:
  apiVersion: v1
  bucket: BUCKET
  prefix: PREFIX
  # optional flags-based filtering criteria
  # for all objects
  flags:
    filter:
      newerThan: "7d" # match objects newer than this value (e.g. 7d10h31s)
      olderThan: "7d" # match objects older than this value (e.g. 7d10h31s)
      createdAfter: "date" # match objects created after "date"
      createdBefore: "date" # match objects created before "date"
      tags:
        - key: "name"
          value: "pick*" # match objects with tag 'name', with all values starting with 'pick'
      metadata:
        - key: "content-type"
          value: "image/*" # match objects with 'content-type', with all values starting with 'image/'
    notify:
      endpoint: "https://notify.endpoint" # notification endpoint to receive job status events
      token: "Bearer xxxxx" # optional authentication token for the notification endpoint
    retry:
```

## How to test this PR?
You would need two PRs 

- https://github.com/minio/madmin-go/pull/213
- https://github.com/minio/mc/pull/4650

## Types of changes

## Checklist:


cmd/batch-expire.go
```diff
@@ -0,0 +1,418 @@
...
+	Metadata      []BatchExpireKV `yaml:"metadata,omitempty" json:"metadata"`
+}
+
+// BatchExpireNotification success or failure notification endpoint for each job attempts
+type BatchExpireNotification struct {
+	Endpoint string `yaml:"endpoint" json:"endpoint"`
+	Token    string `yaml:"token" json:"token"`
+}
+
+// BatchJobExpireFlags various configurations for replication job definition currently includes
```
krisis, line 178:
```suggestion
// BatchJobExpireFlags various configurations for expiration job definition currently includes
```

cmd/batch-expire.go
```diff
@@ -0,0 +1,418 @@
...
+
+	xhttp.DrainBody(resp.Body)
+	if resp.StatusCode != http.StatusOK {
+		return errors.New(resp.Status)
+	}
+
+	return nil
+}
+
+// Expire rotates encryption key of an object
```
krisis, line 228:
```suggestion
// Expire deletes an object
```

cmd/batch-expire.go
```diff
@@ -0,0 +1,414 @@
...
+func (r *BatchJobExpireV1) Expire(ctx context.Context, api ObjectLayer, objInfo ObjectInfo) error {
+	srcBucket := r.Bucket
+	srcObject := objInfo.Name
+
+	_, err := api.DeleteObject(ctx, srcBucket, srcObject, ObjectOptions{
+		DeletePrefix: true,
```
poornas, line 234: allowing filters on batch expiry is incorrect imo - since a prefix delete is being done here and will apply to all versions for the matching object, though user intent is to expire specific versions that match the filter critieria. 

Either filters should be removed or this implementation should do a multi delete @harshavardhana

poornas (changes requested):

cmd/batch-expire.go
harshavardhana: No @poornas every filter is applied at latest object, it doesn't matter how many versions you have. That is not the purpose of this implementation.

cmd/batch-expire.go
```diff
@@ -0,0 +1,479 @@
...
+	globalBatchJobsMetrics.save(job.ID, ri)
+	lastObject := ri.Object
+
+	delay := job.Expire.Flags.Retry.Delay
+	if delay == 0 {
+		delay = batchExpireJobDefaultRetryDelay
+	}
+	rnd := rand.New(rand.NewSource(time.Now().UnixNano()))
+
+	skip := func(info FileInfo) (ok bool) {
```
krisis, line 269:
```suggestion
	filter := func(info FileInfo) (ok bool) {
```
`opts.WalkFitler` takes a function which returns true if the entry should be returned as part of the Walk calls results. `skip` gives the opposite impression/idea. This is a minor change but will help future readers understand the intent better.

PR #17796, opened 2023-08-03
sts: validate if iam subsystem initialized in handlers

poornas:
## Motivation and Context
If IAM system is not initialized, return 503 so request can be re-tried.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/sts-handlers.go
+++ b/cmd/sts-handlers.go
@@ -374,6 +374,11 @@ func (sts *stsAPIHandlers) AssumeRoleWithSSO(w http.ResponseWriter, r *http.Requ
 		}
 	}
 
+	if !globalIAMSys.Initialized() {
+		writeSTSErrorResponse(ctx, w, ErrSTSIAMNotInitialized, errIAMNotInitialized)
+		return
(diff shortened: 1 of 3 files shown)


cmd/sts-handlers.go
```diff
@@ -374,6 +374,11 @@ func (sts *stsAPIHandlers) AssumeRoleWithSSO(w http.ResponseWriter, r *http.Requ
 		}
 	}
 
+	if !globalIAMSys.Initialized() {
+		writeSTSErrorResponse(ctx, w, ErrSTSNotInitialized, errServerNotInitialized)
```
harshavardhana: should we bring a new error here saying `ErrSTSIAMNotInitialized` ? since object layer is initialized but not IAM.

harshavardhana (changes requested):

cmd/sts-handlers.go
poornas: that would be better, done

harshavardhana (approved):

donatello (approved):

PR #17803, opened 2023-08-04
error out by default beyond 10000 versions per object

harshavardhana:
## Description
error out by default beyond 10000 versions per object

## Motivation and Context
```
You've exceeded the limit on the number of versions you can create on this object
```

## How to test this PR?
Well, try to upload more than 10,000 versions per object.
```
mc: <ERROR> Failed to copy `/etc/hosts`. You've exceeded the limit on the number of versions you can create on this object
```

## Types of changes

## Checklist:


cmd/xl-storage-format-v2.go
```diff
@@ -37,9 +38,30 @@ import (
...
+var maxObjectVersions = 10000
+
+func init() {
+	v := env.Get("_MINIO_OBJECT_MAX_VERSIONS", "")
+	if v != "" {
+		maxv, err := strconv.Atoi(v)
+		if err != nil {
+			logger.Info("invalid _MINIO_OBJECT_MAX_VERSIONS value: %s, defaulting to '10000'", v)
+			maxObjectVersions = 10000
+		}
```
vadmeste: Better to do 'else' here (otherwise we can have two logger.Info()

vadmeste (commented):
Minor thing, LGTM otherwise

klauspost (commented):
I'd prefer to have this as a config value. Having to reboot the entire cluster to change the value is not very user friendly.

harshavardhana:
> I'd prefer to have this as a config value. Having to reboot the entire cluster to change the value is not very user friendly.

It is kept the way it is right now since we don't know if this needs to be upgraded to a fully supported value. 

Since it is experimental at this point @klauspost - I really do not want users to keep changing this and play around with it at runtime for now until we bake this in properly.

vadmeste (approved):
LGTM

shtripat (commented):
Verified. LGTM

jiuker (approved):

PR #17804, opened 2023-08-04
Ensure tracking 499 for canceled list objects v1, v2, versions calls

vadmeste:
## Description

listPath() is only called from S3 ListObjects variants. The current code 
ignores context.Canceled error and returns a nil error. This doesn't affect 
S3 applications since context.Canceled means that the s3 client already
 disconnected, but this makes tracing shows wrong information

 (200 OK and an empty list instead of 499 and XML error)

## Motivation and Context
Shows 499 status code in the trace and audit when clients disconnect from Listing S3 API variants calls

## Types of changes

## Checklist:

### DIFF
--- a/cmd/erasure-server-pool.go
+++ b/cmd/erasure-server-pool.go
@@ -1254,6 +1254,11 @@ func (z *erasureServerPools) ListObjectVersions(ctx context.Context, bucket, pre
 		return loi, err
 	}
 	defer merged.truncate(0) // Release when returning
+
+	if contextCanceled(ctx) {
+		return ListObjectVersionsInfo{}, ctx.Err()
+	}
+
 	if versionMarker == "" {
 		o := listPathOptions{Marker: marker}
 		// If we are not looking for a specific version skip it.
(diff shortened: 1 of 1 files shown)


cmd/metacache-server-pool.go
```diff
@@ -233,7 +233,7 @@ func (z *erasureServerPools) listPath(ctx context.Context, o *listPathOptions) (
 	entries, err = filteredResults()
 	cancelList()
 	wg.Wait()
-	if listErr != nil && !errors.Is(listErr, context.Canceled) {
+	if listErr != nil {
```
klauspost: We cancel the listing above the wg.Wait, when we have the results we want.

Therefore a `context.Canceled` does not indicate that the parent canceled.

You will start incorrectly returning errors with this.

klauspost: You can check the parent context:
```
	if contextCanceled(ctx) {
		return entries, ctx.Err()
	}
```

harshavardhana (approved):

krisis (approved):

PR #17826, opened 2023-08-09
s3: Pick a pool with >85% if all other pools are in suspended state

vadmeste:
## Types of changes

## Checklist:

### DIFF
--- a/cmd/erasure-server-pool.go
+++ b/cmd/erasure-server-pool.go
@@ -263,8 +263,8 @@ type serverPoolsAvailableSpace []poolAvailableSpace
 
 type poolAvailableSpace struct {
 	Index      int
-	Available  uint64
-	MaxUsedPct int // Used disk percentage of most filled disk, rounded down.
+	Available  uint64 // in bytes
+	MaxUsedPct int    // Used disk percentage of most filled disk, rounded down.
 }
 
 // TotalAvailable - total available space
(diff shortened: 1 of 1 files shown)


cmd/erasure-server-pool.go
```diff
@@ -299,7 +301,7 @@ func (p serverPoolsAvailableSpace) FilterMaxUsed(max int) {
 
 	// Remove entries that are above.
 	for i, z := range p {
-		if z.MaxUsedPct < max {
+		if z.Valid && z.MaxUsedPct < max {
```
```suggestion
		if z.MaxUsedPct > 0 && z.MaxUsedPct < max {
```

harshavardhana (changes requested):

cmd/erasure-server-pool.go
```diff
@@ -265,6 +265,8 @@ type poolAvailableSpace struct {
 	Index      int
 	Available  uint64
 	MaxUsedPct int // Used disk percentage of most filled disk, rounded down.
+
+	Valid bool
```
klauspost: "Valid" without a bit more context isn't very obvious. A comment would be great.

cmd/erasure-server-pool.go
klauspost: @harshavardhana It seems to me this would break on new installs. Maybe @vadmeste can explain the original code.

vadmeste: yeah it is risky for me to ignore pools with 0% of usage, such as (0.11%)

cmd/erasure-server-pool.go
vadmeste: > "Valid" without a bit more context isn't very obvious. A comment would be great.

I removed it. I am using Available now.

harshavardhana (approved):

klauspost (approved):
LGTM

PR #17838, opened 2023-08-10
do not log client canceled events

harshavardhana:
## Description
do not log client-canceled events

## Motivation and Context
avoid logging context canceled/timeout errors 
as audit miss events since they are not.

## How to test this PR?
Nothing special a canceled client conn leads

## Types of changes

## Checklist:

### DIFF
--- a/internal/logger/target/http/http.go
+++ b/internal/logger/target/http/http.go
@@ -377,7 +377,11 @@ func (h *Target) Send(ctx context.Context, entry interface{}) error {
 	select {
 	case h.logCh <- entry:
 	case <-ctx.Done():
-		return ctx.Err()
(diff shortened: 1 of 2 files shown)


vadmeste (commented):
I really think it is better if we ignore logs from the upper level, I can send a PR for that

harshavardhana:
> I really think it is better if we ignore logs from the upper level, I can send a PR for that

yes but we can't return errors for valid context canceled ones since the audit is invalid as the request didn't complete.

harshavardhana:
PTAL @vadmeste @krisis

vadmeste (approved):
LGTM

krisis (approved):

PR #17842, opened 2023-08-11
Fix wrapped io.EOF

klauspost:
## Description

When listing getObjectFileInfo can return `io.EOF` if file is being written.

When we wrap the error it will *not* retry upstream, since `io.EOF` is a valid return value.

Allow one retry before returning errors and canceling the listing.

## How to test this PR?

Low-prob listing error. Couldn't find an obvious regression. Has been in code for a while.

## Types of changes


vadmeste (approved):
LGTM

harshavardhana (approved):

PR #17855, opened 2023-08-15
Fix distributed listing not able to resume

klauspost:
## Description

Two fields in lifecycles made GOB encoding consistently fail with `gob: type lifecycle.Prefix has no exported fields`.

This meant that in distributed systems listings would never be able to continue and would restart on every call.

Fix issues and be sure to log these errors at least once per bucket. We may see some connectivity errors here, but we shouldn't hide them.

## How to test this PR?

Set `_MINIO_SERVER_DEBUG=on`. Observe if listings are reused.

## Types of changes


vadmeste (approved):
LGTM

harshavardhana (approved):

PR #17861, opened 2023-08-16
if object is a delete marker it must skip tags filter in ILM

harshavardhana:
## Contribution License

## Description
if object is a delete marker it must skip tags filter in ILM

## Motivation and Context
With the following rule if the latest version is a DEL marker then the object is not
purged via ExpiredObjectAllVersions at all due to the fact that tags filter out the
rule.

```json
      {
        "Expiration": {
          "Days": 3,
          "ExpiredObjectAllVersions": true
        },
        "ID": "TTL 90",
        "Filter": {
          "Tag": {
            "Key": "ttl",
            "Value": "3"
          }
        },
        "Status": "Enabled"
      },
```

DELETE markers must not be considered with tags since they do not have any.

## How to test this PR?
Requires the above behavior ^^

## Types of changes

## Checklist:

### DIFF
--- a/internal/bucket/lifecycle/lifecycle.go
+++ b/internal/bucket/lifecycle/lifecycle.go
@@ -261,7 +261,7 @@ func (lc Lifecycle) FilterRules(obj ObjectOpts) []Rule {
 		if !strings.HasPrefix(obj.Name, rule.GetPrefix()) {
 			continue
 		}
-		if !rule.Filter.TestTags(obj.UserTags) {
+		if !obj.DeleteMarker && !rule.Filter.TestTags(obj.UserTags) {
 			continue
 		}
 		rules = append(rules, rule)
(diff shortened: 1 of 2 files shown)


klauspost (approved):

jiuker (approved):

internal/bucket/lifecycle/lifecycle.go
```diff
@@ -261,7 +261,7 @@ func (lc Lifecycle) FilterRules(obj ObjectOpts) []Rule {
 		if !strings.HasPrefix(obj.Name, rule.GetPrefix()) {
 			continue
 		}
-		if !rule.Filter.TestTags(obj.UserTags) {
+		if !obj.DeleteMarker && !rule.Filter.TestTags(obj.UserTags) {
```
vadmeste, line 264: @harshavardhana maybe for now, we can check if the current rule contains ExpireAllObjects; I am looking what is the behavior of ILM in this case (I am guess this code is correct after all)

vadmeste (commented):
One comment

internal/bucket/lifecycle/lifecycle.go
harshavardhana: Okay

vadmeste (approved):
LGTM

PR #17879, opened 2023-08-19
fix:before close channel check the length to make sure the data have being sent.

jiuker:
## Contribution License

## Description

before close channel. We can check the channel length to make sure all the data have being sent.
Mostly, `mrfCh` have 4000 size. It will have data before close.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/notification.go
+++ b/cmd/notification.go
@@ -25,6 +25,7 @@ import (
 	"math/rand"
 	"net/http"
 	"net/url"
+	"runtime"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


cmd/notification.go
```diff
@@ -1528,7 +1529,24 @@ func (sys *NotificationSys) GetReplicationMRF(ctx context.Context, bucket, node
...
+			return
+		default:
+			if len(ch) == 0 {
+				return
+			}
+			// make sure all the data have being sent
+			runtime.Gosched()
+		}
+	}
+}
```
harshavardhana, line 1552: why do we need to do this? this will create a busy loop from the `for { select { default: ` there is no delay there so it will shoot up CPU 

Can you explain what you are trying to achieve?

harshavardhana: Because wg.Wait() ensures that all the entries in channel are empty before close()

harshavardhana (changes requested):

PR #17881, opened 2023-08-19
fix:copy the req.body into io.Discard always when it need

jiuker:
## Contribution License

## Description

copy the req.body into io.Discard always.

the same as 
```golang
	        res, err := c.Delete(
			index,
			key,
			c.Delete.WithContext(ctx),
		)
		if err != nil {
			return err
		}
		defer res.Body.Close()
		defer io.Copy(io.Discard, res.Body)
		if res.IsError() {
			return fmt.Errorf("Delete err: %s", res.String())
		}
```

## Types of changes

## Checklist:

### DIFF
--- a/internal/event/target/elasticsearch.go
+++ b/internal/event/target/elasticsearch.go
@@ -23,7 +23,6 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
-	"io"
 	"net/http"
 	"net/url"
 	"os"
@@ -475,12 +474,10 @@ func (c *esClientV7) createIndex(args ElasticsearchArgs) error {
 		if err != nil {
 			return err
 		}
-		defer resp.Body.Close()
+		defer DrainBody(resp.Body)
 		if resp.IsError() {
-			err := fmt.Errorf("Create index err: %s", res.String())
-			return err
+			return fmt.Errorf("Create index err: %s", res.String())
 		}
-		io.Copy(io.Discard, resp.Body)
 		return nil
 	}
 	return nil
@@ -493,8 +490,7 @@ func (c *esClientV7) ping(ctx context.Context, _ ElasticsearchArgs) (bool, error
 	if err != nil {
 		return false, store.ErrNotConnected
 	}
-	io.Copy(io.Discard, resp.Body)
-	resp.Body.Close()
+	DrainBody(resp.Body)
 	return !resp.IsError(), nil
 }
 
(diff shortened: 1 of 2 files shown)


vadmeste (approved):
LGTM

donatello (approved):

internal/event/target/elasticsearch.go
```diff
@@ -476,11 +476,10 @@ func (c *esClientV7) createIndex(args ElasticsearchArgs) error {
 			return err
 		}
 		defer resp.Body.Close()
+		defer io.Copy(io.Discard, resp.Body)
```
harshavardhana: You should drain before and then call Close().

harshavardhana (changes requested):

internal/event/target/elasticsearch.go
harshavardhana: Move them into same defer statement

```
defer func() {
   io.CopyN(io.Discard, resp.Body, humanize.MiByte)
   resp.Body.Close()
}
```

internal/event/target/elasticsearch.go
```diff
@@ -475,12 +474,10 @@ func (c *esClientV7) createIndex(args ElasticsearchArgs) error {
 		if err != nil {
 			return err
 		}
-		defer resp.Body.Close()
+		defer DrainBody(resp.Body)
 		if resp.IsError() {
-			err := fmt.Errorf("Create index err: %s", res.String())
-			return err
+			return fmt.Errorf("Create index err: %s", res.String())
```
harshavardhana:
```suggestion
			return fmt.Errorf("Create index err: %v", res)
```

internal/event/target/webhook.go
```diff
@@ -297,3 +296,23 @@ func NewWebhookTarget(ctx context.Context, id string, args WebhookArgs, loggerOn
...
+	// (typically Transport) may not be able to re-use a persistent TCP
+	// connection to the server for a subsequent "keep-alive" request.
+	if respBody != nil {
+		// Drain any remaining Body and then close the connection.
+		// Without this closing connection would disallow re-using
+		// the same connection for future uses.
+		//  - http://stackoverflow.com/a/17961593/4465767
+		defer respBody.Close()
+		io.Copy(io.Discard, respBody)
+	}
```
harshavardhana, line 318: We should limit the size of io.Copy() otherwise it will never end if respBody doesn't end with EOF. So we can drain up-to 1MiB

harshavardhana (changes requested):

harshavardhana (approved):

Issue #16672, opened 2023-02-21
MinIO web console logs out after a few minutes with oidc


ghost:
## Expected Behavior
MinIO web console stays logged in

## Current Behavior
MinIO web console logs out after a few minutes when using OpenID Connect

## Possible Solution
Unknown

## Steps to Reproduce (for bugs)
1. Setup 4-node cluster with OpenID
2. Login to https://node-1.internal:9090 to access web console
3. Wait a few minutes and refresh, the login page is shown

## Context
Use web console

## Regression
Unsure

## Your Environment
* Version used (`minio --version`): RELEASE.2023-02-17T17-52-43Z
* Server setup and configuration: nodes running in podman containers
* Operating System and version (`uname -a`): `Linux na-tst-minio-1.internal 5.14.0-162.6.1.el9_1.x86_64 #1 SMP PREEMPT_DYNAMIC Fri Sep 30 07:36:03 EDT 2022 x86_64 x86_64 x86_64 GNU/Linux`

I ran a `mc admin trace -a --funcname admin* --funcname sts* -v` and could see the following log when I refreshed the page:

```


na-tst-minio-1.internal:9000 [REQUEST admin.AccountInfo] [2023-02-21T15:37:41.522] [Client IP: 127.0.0.1]
na-tst-minio-1.internal:9000 GET /minio/admin/v3/accountinfo
na-tst-minio-1.internal:9000 Proto: HTTP/1.1
na-tst-minio-1.internal:9000 Host: 127.0.0.1:9000
na-tst-minio-1.internal:9000 Authorization: AWS4-HMAC-SHA256 Credential=<snip>/20230221//s3/aws4_request, SignedHeaders=host;x-amz-content-sha256;x-amz-date;x-amz-security-token, Signature=26b04<snip>612a6
na-tst-minio-1.internal:9000 Content-Length: 0
na-tst-minio-1.internal:9000 User-Agent: MinIO (linux; amd64) madmin-go/2.0.0
na-tst-minio-1.internal:9000 X-Amz-Content-Sha256: e3b0c4<snip>b855
na-tst-minio-1.internal:9000 X-Amz-Date: 20230221T053741Z
na-tst-minio-1.internal:9000 X-Amz-Security-Token: eyJhbG<snip>4RPg
na-tst-minio-1.internal:9000 
na-tst-minio-1.internal:9000 [RESPONSE] [2023-02-21T15:37:41.522] [ Duration 143µs  ↑ 98 B  ↓ 224 B ]
[...]

Issue #16308, opened 2022-12-23
Web Session timeout is very short with OPENID identity Auth method 


abdennour:
## NOTE
If this case is urgent, please subscribe to [Subnet](https://min.io/pricing) so that our 24/7 support team may help you faster.

## Expected Behavior
Regardless the identity provider, the web session timeout should be the same. 

## Current Behavior
When connecting Minio with an identity provider thru OIDC method, the web session expires quickly. If no activities on the website for almost 1 minute, you will get your self logged out. 

## Possible Solution

Expecting to provide an env vars like `MINIO_IDENTITY_OPENID_SESSION_TIMEOUT=3600s`

## Steps to Reproduce (for bugs)

1. Install minio using bitnami helm chart (version `11.10.24`)
2. Configuring new client in the keycloak realm, then Setting the following env vars in minio: 
```yaml
    - name: MINIO_IDENTITY_OPENID_CONFIG_URL
      value: "https://sso.comp.lan/auth/realms/myrealm/.well-known/openid-configuration" 
    - name: MINIO_IDENTITY_OPENID_CLIENT_ID
      value: "minio" # ENV
    - name: MINIO_IDENTITY_OPENID_CLIENT_SECRET
      value: "xxxxxxx-xxx-xxxx-xxxx-xxxxxxx" # ENV
    - name: MINIO_IDENTITY_OPENID_SCOPES
      value: "openid,profile,email,groups"
    - name: MINIO_IDENTITY_OPENID_CLAIM_NAME
      value: "groups"
    - name: MINIO_IDENTITY_OPENID_REDIRECT_URI
      value: https://minio.comp.lan/oauth_callback # ENV
    - name: MINIO_IDENTITY_OPENID_COMMENT
      value: Comp SSO
```
4. Upgrading the helm release & Navigating to Website again

===> the Login works fine however the web session is very short & it expires as soon as you stop navigating/clicking here and there.

## Context
RBAC with OIDC

## Regression
N/A
## Your Environment
* Version used (`minio --version`):
```
minio version DEVELOPMENT.2022-12-12T19-27-27Z (commit-id=a469e6768df4d5d2cb340749fa58e4721a7dee96)
Runtime: go1.18.9 linux/amd64
License: GNU AGPLv3 <https://www.gnu.org/licenses/agpl-3.0.html>
Copyright: 2015-2022 MinIO, Inc.
```

* Server setup and configuration: kubernetes .1.24
[...]

PR #17882, opened 2023-08-21
Add session timeout env variable

mrvanes:
## Description
Add `MINIO_IDENTITY_OPENID_SESSION_TIMEOUT` env var to override OP `exp` claim expiry.

## Motivation and Context
As can be read in issue https://github.com/minio/minio/issues/16308 and https://github.com/minio/minio/issues/16672 there is popular demand for overriding OP (IdP) `exp` claim value governing console session timeout.

Contrary to what @harshavardhana says, the RP (Relying Party or SP) session expiration is not governed by the OP `exp` claim value, this value is used to express the expiry of the token and should be treated as such.
See also https://github.com/minio/minio/issues/16308#issuecomment-1663475027

Even though Minio's interpretation of the OIDC spec is wrong, I left the original behaviour intact for backwards compatibility.

### Caveats
I didn't add the equivalent setting to the console Identity/OpenID Interface.
I'm unsure about the way I access the OIDC ProviderCfgs["_"].SessionTimout member in openid.Config.Validate()? Please correct me if I'm wrong.

## How to test this PR?
- Set env variable `MINIO_IDENTITY_OPENID_SESSION_TIMEOUT` to any integer representing timeout in seconds
- restart minio server
- login using OIDC Provider
- Experience session timeout longer than value set in `exp` claim

## Types of changes

## Checklist:


harshavardhana:
The correct PR is #17885

PR #17894, opened 2023-08-22
Reduce memory profiling rate

klauspost:
## Description

Change profiling from every 4KB to every 128KB, reducing the lock contention by a factor of 32.

## Motivation and Context

Reduce lock contention seen on dual socket systems.

Observed usage was 16% - this should be reduced to at least <0.5% - probably less since there is less contention for the actual lock.

## Types of changes


vadmeste (approved):
LGTM

harshavardhana (approved):

PR #17901, opened 2023-08-23
feat:refactor the perf client for TTFB and TotalResponseTime

jiuker:
## Contribution License

## Description

prev pr for `TTFB` and `TotalResponseTime`
We will have the new request for perfNet. It looks the same.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/perf-tests.go
+++ b/cmd/perf-tests.go
@@ -19,6 +19,7 @@ package cmd
 
 import (
 	"context"
+	"encoding/gob"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


cmd/perf-tests.go
```diff
@@ -422,3 +408,37 @@ func siteNetperf(ctx context.Context, duration time.Duration) madmin.SiteNetPerf
...
+		return "", err
+	}
+	rp := cli.GetEndpointURL()
+	reqURL := &url.URL{
+		Scheme: rp.Scheme,
+		Host:   rp.Host,
+		Path:   reqPath,
+	}
+	defer func() {
+		enpoint = rp.String()
```
klauspost:
```suggestion
		endpoint = rp.String()
```

Why the defer? Why not just set it and return it as a regular value? 'rp' is not changing from here.

shtripat (approved):
lgtm

cmd/perf-tests.go
```diff
@@ -422,3 +408,37 @@ func siteNetperf(ctx context.Context, duration time.Duration) madmin.SiteNetPerf
 		TotalConn:       uint64(connectionsPerPeer),
 	}
 }
+
+// perfNetRequest - reader for http.request.body, result should be a pointer if it need
+func perfNetRequest(ctx context.Context, timeout time.Duration, deploymentID string, reqPath string, reader io.Reader, result interface{}) (enpoint string, err error) {
```
klauspost:
```suggestion
func perfNetRequest(ctx context.Context, timeout time.Duration, deploymentID, reqPath string, reader io.Reader, result any) (endpoint string, err error) {
```

Why are you sending a timeout? Wouldn't attaching the timeout to the context do the same?

cmd/perf-tests.go
jiuker: because there are four place will write as `rp.String(),nil`. The endpoint should be return always. @klauspost

cmd/perf-tests.go
jiuker: Timeout should be used to http.Request. Shouldn't include get the client.

cmd/site-replication.go
```diff
@@ -696,31 +693,16 @@ func (c *SiteReplicationSys) Netperf(ctx context.Context, duration time.Duration
...
-			return gob.NewDecoder(resp.Body).Decode(&result)
+			ctx, cancel := context.WithTimeout(ctx, duration+10*time.Second)
+			defer cancel()
+			result.Endpoint, err = perfNetRequest(
+				ctx,
+				info.DeploymentID,
+				adminPathPrefix+adminAPIVersionPrefix+adminAPISiteReplicationNetPerf,
+				nil,
+				&result,
+			)
```
harshavardhana: allocating of `result` and overwriting it is a bit of an anti-pattern style What we should do is return the `result{}` or have the endpoint be populated before returning this function. 

And this function only returns `err` the new function signature will be like this

```
func perfNetRequest(ctx context.Context, deploymentID, reqPath string, reader io.Reader, result any) (err error) {
```

harshavardhana (changes requested):

cmd/site-replication.go
klauspost: It is a bit like reading C again - sending result pointers :)

jiuker: > It is a bit like reading C again - sending result pointers :)

or wrap it in a `func (body []byte){  // here is the response body byte}` ? @klauspost

harshavardhana (approved):

PR #17925, opened 2023-08-26
Return successorModTime in quorum when available

krisis:

## Description
Return SuccessorModTime in quorum when available

## Motivation and Context
findFileInfoInQuorum function, prior to this change, picks a valid FileInfo which has several fields that are in quorum but leaves out SuccessorModTime. Leaving this out could lead to picking a FileInfo with invalid values in fields like IsLatest, SuccessorModTime leading to incorrect ILM policy evaluation.

## How to test this PR?
Unit tests added

## Types of changes

## Checklist:

### DIFF
--- a/cmd/erasure-metadata.go
+++ b/cmd/erasure-metadata.go
@@ -393,14 +393,39 @@ func findFileInfoInQuorum(ctx context.Context, metaArr []FileInfo, modTime time.
 		return FileInfo{}, errErasureReadQuorum
 	}
 
+	// Find the successor mod time in quorum, otherwise leave the
+	// candidate's successor modTime as found
+	succModTimeMap := make(map[time.Time]int)
+	var candidate FileInfo
+	var found bool
 	for i, hash := range metaHashes {
 		if hash == maxHash {
 			if metaArr[i].IsValid() {
-				return metaArr[i], nil
+				if !found {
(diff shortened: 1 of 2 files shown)


cmd/erasure-metadata.go
```diff
@@ -393,14 +393,39 @@ func findFileInfoInQuorum(ctx context.Context, metaArr []FileInfo, modTime time.
...
+			smodTimeQuorum = true
+			succModTime = smodTime
+			break
+		}
+	}
 
+	if found {
+		if smodTimeQuorum {
+			candidate.SuccessorModTime = succModTime
+			candidate.IsLatest = succModTime.IsZero()
```
harshavardhana, line 425: So will this make a version latest only if it doesn't have a SuccesorModTime?

harshavardhana (changes requested):

cmd/erasure-metadata.go
krisis: Yes. We modify the candidate's successor mod time and isLatest if and only if we found this successor mod time to be in quorum. IsLatest, by definition, is true only when a version's successor mod time is 'zero', i.e undefined.

harshavardhana (approved):

vadmeste (approved):
LGTM

PR #17943, opened 2023-08-29
Websocket Grid Connectivity

klauspost:
## Description

Adds a websocket grid feature, which allows servers to communicate via a single two-way connection.

There are 2 request types: 

* Single requests, which are `[]byte => ([]byte, error)`. This is for efficient small roundtrips with small payloads.
* Streaming requests which are `[]byte, chan []byte => chan []byte (and error)`, which allows for different combinations of full twoway streams with an initial payload.

Only a single stream is created between two machines - and there is as such no server/client relation, since both sides can initiate and handle requests. Which server initiates the request is decided deterministically on the server names.

Requests are done through a mux client and server, which handles message passing, congestion, cancelation, timeouts, etc.

If a connection is lost all requests are canceled and the calling server will try to reconnect.

Handler are registered and can either operate directly on byte slices or use a higher level generics abstraction.

There is no versioning of handlers/clients and incompatible changes should be handled by adding new handlers.

For protocol changes, the request path can be changed to a new one.

First all servers creates a "Manager". The manager must know its own address as well as all remote addresses. This will manage all connections. To get a connection to any remote, ask the manager to provide it given the remote address using `func (m *Manager) Connection(host string) *Connection`.

All serverside handlers must also be registered on the manager. This will make sure that all incoming requests are served. For streaming requests the number of in-flight requests and responses must also be given.

The "Connection" returned manages the mux-clients. Requests issued to the connection will be sent to the remote.

* `func (c *Connection) Request(ctx context.Context, h HandlerID, req []byte) ([]byte, error)`  performs a single request and returns the result. Any deadline provided on the request is forwarded to the server and canceling the context will make the function return at once.

* `func (c *Connection) NewStream(ctx context.Context, h HandlerID, payload []byte) (st *Stream, err error) ` will initiate a remote call and send the initial payload.

```Go
// A Stream is a two-way stream.
// All responses *must* be read by the caller.
// If the call is canceled though the context,
// the appropriate error will be returned.
type Stream struct {
	// Responses from the remote server.
	// Channel will be closed after error or when remote closes.
	// All responses *must* be read by the caller until either an error is returned or the channel is closed.
	// Canceling the context will cause the context cancellation error to be returned.
	Responses <-chan Response

	// Requests sent to the server.
	// If the handler is defined with 0 incoming capacity this will be nil.
	// Channel *must* be closed to signal the end of the stream.
	// If the request context is canceled, the stream will no longer process requests.
	Requests chan<- []byte
}

type Response struct {
	Msg []byte
	Err error
}
```

There are generic versions of the server/client handlers that allows using type safe implementations for data types that support msgpack marshal/unmarshal.

## How to test this PR?

Still WIP. Adding more tests is needed and also an initial implementation.

## Types of changes


internal/grid/manager.go
```diff
@@ -0,0 +1,308 @@
...
+		return ErrUnknownHandler
+	}
+	if m.handlers.hasAny(id) && !id.isTestHandler() {
+		return ErrHandlerAlreadyExists
+	}
+
+	m.handlers.stateless[id] = &h
+	return nil
+}
+*/
```
harshavardhana, line 267: is this needed?

harshavardhana (changes requested):

internal/grid/manager.go
klauspost: @harshavardhana We want stateless, for things like traces, metrics, etc, where messages may be dropped, but we don't have to do congestion handling.

However, I decided to focus on the two other interfaces and make them fully functional/tested first.

PR #17944, opened 2023-08-29
Clarify Community Helm Chart

ravindk89:
There is some consistent confusion between the Community Helm Chart in this repo and the MinIO Kubernetes Operator Helm Chart.

This change seeks to clarify the differences between the two charts and which ones are community maintained vs MinIO maintained.

### DIFF
--- a/helm/minio/README.md
+++ b/helm/minio/README.md
@@ -1,10 +1,12 @@
-# MinIO Helm Chart
+# MinIO Community Helm Chart
 
 [![Slack](https://slack.min.io/slack?type=svg)](https://slack.min.io) [![license](https://img.shields.io/badge/license-AGPL%20V3-blue)](https://github.com/minio/minio/blob/master/LICENSE)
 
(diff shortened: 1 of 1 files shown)


cesnietor (approved):

PKizzle:
Thanks for clarifying that the helm chart is community maintained.
Maybe you could add a sentence about why it even exists (the difference to the officially maintained operator chart).
Also information on how to contribute to the community helm chart would be appreciated.

harshavardhana:
> Thanks for clarifying that the helm chart is community maintained.
> Maybe you could add a sentence about why it even exists (the difference to the officially maintained operator chart).
> Also information on how to contribute to the community helm chart would be appreciated.

It exists because there are users who want to preserve `https://github.com/helm/charts/tree/master/stable/minio

ravindk89:
@PKizzle I'll have to defer contribution guidelines to someone who understands the chart setup better - though the overall guideline should be the same as our general contributors guide w.r.t. commits, reviews, and merges.

PKizzle:
@harshavardhana Is it not also the newer version of https://github.com/minio/charts?. I would add: The old helm chart (and therefore the new one) provide the easiest and smallest possible setup of minio with hardware configurations not supported by the official minio operator based helm chart (number of drives < 4).

PR #17945, opened 2023-08-29
heal: calculate the number of workers based on  NRRequests

vadmeste:

## Types of changes

## Checklist:

### DIFF
--- a/cmd/global-heal.go
+++ b/cmd/global-heal.go
@@ -20,6 +20,7 @@ package cmd
 import (
 	"context"
 	"fmt"
+	"runtime"
 	"sort"
 	"time"
 
@@ -30,6 +31,7 @@ import (
 	"github.com/minio/minio/internal/logger"
 	"github.com/minio/pkg/v2/console"
 	"github.com/minio/pkg/v2/wildcard"
+	"github.com/minio/pkg/v2/workers"
 )
 
 const (
@@ -132,30 +134,8 @@ func getLocalBackgroundHealStatus(ctx context.Context, o ObjectLayer) (madmin.Bg
 	return status, true
 }
 
-func mustGetHealSequence(ctx context.Context) *healSequence {
-	// Get background heal sequence to send elements to heal
-	for {
-		globalHealStateLK.RLock()
-		hstate := globalBackgroundHealState
-		globalHealStateLK.RUnlock()
-
-		if hstate == nil {
-			time.Sleep(time.Second)
-			continue
-		}
-
-		bgSeq, ok := hstate.getHealSequenceByToken(bgHealingUUID)
-		if !ok {
-			time.Sleep(time.Second)
-			continue
-		}
-		return bgSeq
-	}
-}
-
 // healErasureSet lists and heals all objects in a specific erasure set
 func (er *erasureObjects) healErasureSet(ctx context.Context, buckets []string, tracker *healingTracker) error {
-	bgSeq := mustGetHealSequence(ctx)
 	scanMode := madmin.HealNormalScan
 
 	// Make sure to copy since `buckets slice`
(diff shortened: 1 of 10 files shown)


cmd/global-heal.go
```diff
@@ -173,6 +173,21 @@ func (er *erasureObjects) healErasureSet(ctx context.Context, buckets []string,
...
+	}
+
+	numHealers := uint64(1)
+	if info.NRRequests > 0 {
+		numHealers = info.NRRequests
+	}
+	if numHealers > 128 {
+		numHealers = 128
+	}
+
```
harshavardhana:
```suggestion
	numHealers := uint64(4)
	if info.NRRequests > runtime.GOMAXPROCS(0) {
	      numHealers = runtime.GOMAXPROCS(0)/4
	} else {
	      numHealers = info.NRRequests/4
	}
	if numHealers == 0 || numHealers == 1 {
	       numHealers = 4
	}
	// allow overriding this value as well..
	if v := globalAPIConfig.getHealWorkers(); v > 0 {
	     numHealers = v
	}
```

harshavardhana (changes requested):

internal/config/heal/heal.go
```diff
@@ -103,6 +115,11 @@ var DefaultKVS = config.KVS{
 		Key:   IOCount,
 		Value: "100",
 	},
+	config.KV{
+		Key:        Workers,
+		Value:      "",
+		Deprecated: true,
```
harshavardhana:
```suggestion
		Deprecated: true,
```

:P

harshavardhana (changes requested):

internal/config/heal/heal.go
harshavardhana: @vadmeste you need to remove "Deprecated: true"

harshavardhana (changes requested):

vadmeste:
> @vadmeste you need to remove "Deprecated: true"

The point is to make this variable hidden from users, until someone expliclity sets it

harshavardhana:
> The point is to make this variable hidden from users, until someone explicitly sets it

For that we can't reuse this boolean, we should simply have this available to all users but document that it is not needed to be set without expert guidance via SUBNET

cmd/xl-storage.go
```diff
@@ -244,6 +246,11 @@ func newXLStorage(ep Endpoint, cleanUp bool) (s *xlStorage, err error) {
 		diskIndex:  -1,
 	}
 
+	// Sanitaize before setting it
```
harshavardhana:
```suggestion
	// Sanitize before setting it
```

harshavardhana (changes requested):

cmd/global-heal.go
```diff
@@ -173,6 +153,28 @@ func (er *erasureObjects) healErasureSet(ctx context.Context, buckets []string,
...
+	if numHealers == 0 || numHealers == 1 {
+		numHealers = 4
+	}
+	// allow overriding this value as well..
+	if v := globalHealConfig.GetWorkers(); v > 0 {
+		numHealers = uint64(v)
+	}
+
+	jt, _ := workers.New(int(numHealers))
+
```
harshavardhana: let's also update https://github.com/minio/minio/blob/e3fbcaeb729f5dd02f4b7c96c10b952c1d7dfe2a/docs/config/README.md?plain=1#L274

To show this configurable value

harshavardhana: We have to say that this is Linux only optimization, for example on Windows and other OS we will use `number of cores / 4`

harshavardhana (changes requested):

cmd/global-heal.go
```diff
@@ -173,6 +153,28 @@ func (er *erasureObjects) healErasureSet(ctx context.Context, buckets []string,
...
+	}
+
+	var numHealers uint64
+
+	if numCores := uint64(runtime.GOMAXPROCS(0)); info.NRRequests > numCores {
+		numHealers = numCores / 4
+	} else {
+		numHealers = info.NRRequests / 4
+	}
+	if numHealers == 0 || numHealers == 1 {
```
harshavardhana:
```suggestion
	if numHealers < 4 {
```

We expect this to be 4 minimum, unless user wants this to be some other value.

harshavardhana (changes requested):

harshavardhana (approved):

PR #17946, opened 2023-08-29
Implement batch-expiration for objects

krisis:

## Description
Based on an initial PR from Harshavardhana -
https://github.com/minio/minio/pull/17792

Companion PRs
- [ ] mc PR: https://github.com/minio/mc/pull/4677
- [x] madmin-go PR: https://github.com/minio/madmin-go/pull/239
- [x] madmin-go PR: https://github.com/minio/madmin-go/pull/248

## Motivation and Context
ILM policy at a bucket-level allows user to expire objects based on prefix or object tags based filtering and object age-based selection criteria. This doesn't work well when the number of days to expire in since object creation is very small, say less than a week.
Batch-expiry feature works on a bucket level. It allows users to express a selection of objects to be expired based on prefix and/or object tags. It differs from ILM in the following ways,
- MinIO's batch jobs subsystem will immediately identify the selected objects and expire them
- Unlike ILM processing, batch-expiry doesn't yield to application I/O
- Runs only once, to completion; If newer objects matching the selection criteria appear in the object namespace, this batch job may be run again.

## How to test this PR?
https://github.com/krisis/batch-expiry-test contains instructions and commands useful to test batch-expiry.

## Types of changes

## Checklist:


harshavardhana (approved):

cmd/batch-expire.go
```diff
@@ -0,0 +1,634 @@
...
+	Size          struct {
+		LesserThan  BatchJobSize `yaml:"lesserThan,omitempty" json:"lesserThan"`
+		GreaterThan BatchJobSize `yaml:"greaterThan,omitempty" json:"greaterThan"`
+	} `yaml:"size" json:"size"`
+	Type  string              `yaml:"type" json:"type"`
+	Name  string              `yaml:"name" json:"name"`
+	Purge BatchJobExpirePurge `yaml:"purge" json:"purge"`
+}
+
+// Matches returns true if fi matches the filter conditions specified in ef.
```
shtripat: `if obj matches the filter`?

cmd/batch-expire.go
```diff
@@ -0,0 +1,634 @@
...
+			tagMap = t.ToMap()
+		}
+
+		for _, kv := range ef.Tags {
+			// Object (version) must match all tags specified in
+			// the filter
+			var match bool
+			for t, v := range tagMap {
+				if kv.Match(BatchJobKV{Key: t, Value: v}) {
+					match = true
```
shtripat, line 151: Do we need to break once match found here?

cmd/batch-expire.go
```diff
@@ -0,0 +1,634 @@
...
+			// standard metadata headers
+			// specified in the filter
+			var match bool
+			for k, v := range obj.UserDefined {
+				if !stringsHasPrefixFold(k, "x-amz-meta-") && !isStandardHeader(k) {
+					continue
+				}
+				// We only need to match x-amz-meta or standardHeaders
+				if kv.Match(BatchJobKV{Key: k, Value: v}) {
+					match = true
```
shtripat, line 172: Same as above for break

cmd/batch-expire.go
```diff
@@ -0,0 +1,634 @@
...
+				}
+			}
+			if !match {
+				return false
+			}
+		}
+	}
+
+	if ef.Size.LesserThan > 0 {
```
shtripat: I feel for match, by default we can expect object to not match. If conditions met, return true, else false at last. That looks more readable to me.

cmd/batch-expire.go
```diff
@@ -0,0 +1,634 @@
...
+}
+
+const (
+	// BatchJobExpireObject - object type
+	BatchJobExpireObject string = "object"
+	// BatchJobExpireDeleted - delete marker type
+	BatchJobExpireDeleted string = "deleted"
+)
+
+// Validate returns nil if f has valid fields, validation error otherwise.
```
shtripat: `if ef has valid fields` ?

cmd/batch-job-common-types.go
```diff
@@ -81,3 +82,37 @@ func (r BatchJobRetry) Validate() error {
...
+type BatchJobSizeFilter struct {
+	LesserThan  BatchJobSize `yaml:"lesserThan,omitempty" json:"lesserThan"`
+	GreaterThan BatchJobSize `yaml:"greaterThan,omitempty" json:"greaterThan"`
+}
+
+// BatchJobSize supports humanized byte values in yaml files type BatchJobSize uint64
+type BatchJobSize int64
+
+// LesserThan returns true if sz is less than s, false otherwise.
+func (s BatchJobSize) LesserThan(sz int64) bool {
```
shtripat: Semantically I feel `BatchJobSize.LesserThan(arg)` means if current `BatchJobSize` value is lesser than argument value passed. Should we change that way? I think that's the reason I was confused while usage of this function.

cmd/batch-job-common-types.go
```diff
@@ -81,3 +82,37 @@ func (r BatchJobRetry) Validate() error {
...
+	return sz < int64(s)
+}
+
+// GreaterThan returns true if sz is greater than s, false otherwise.
+func (s BatchJobSize) GreaterThan(sz int64) bool {
```
shtripat: Same as above

shtripat (commented):
few minor comments/questions
will verify the changes following the instructions.

shtripat (approved):
Verified and works as expected.
Don't have any other comments other than prior minor ones.
Verified metrics
```
$ mc admin prometheus metrics myminio bucket | grep _expire_objects 
# HELP minio_bucket_batch_expire_objects Get successfully completed batch job expireobjects
# TYPE minio_bucket_batch_expire_objects gauge
minio_bucket_batch_expire_objects{bucket="bucket",jobId="dQT94EHKcAjERSPS5FikkZ:-1",server="127.0.0.1:9000"} 97
# HELP minio_bucket_batch_expire_objects_failed Get failed batch job expireobjects
# TYPE minio_bucket_batch_expire_objects_failed gauge
minio_bucket_batch_expire_objects_failed{bucket="bucket",jobId="dQT94EHKcAjERSPS5FikkZ:-1",server="127.0.0.1:9000"} 0
```

cmd/batch-expire.go
krisis: An object matches the list of tags filter criteria if and only if the object's tags match (glob pattern match) all of them (think AND operation).

cmd/batch-expire.go
krisis: `metadata` filter criteria too can be thought of as an AND operation of `kv.Match` results over all metadata headers specified.

cmd/batch-job-common-types.go
krisis: I agree with you that this is confusing. Let me clean this up.

cmd/batch-expire.go
krisis: This is not possible. For an object is said to have matched the filter criteria only when it matches all of them. Think AND (again). This is also why we can return early if even an object fails to match a single criterion, but we can't declare it as a match unless all the criteria are met.

cmd/batch-expire.go
klauspost: @krisis But the logic here doesn't reflect that. The logic implies OR.

cmd/batch-job-common-types.go
```diff
@@ -81,3 +83,52 @@ func (r BatchJobRetry) Validate() error {
 
 	return nil
 }
+
+// BatchJobSizeFilter supports size based filters - LesserThan and GreaterThan
+type BatchJobSizeFilter struct {
+	UpperBound BatchJobSize `yaml:"lesserThan,omitempty" json:"lesserThan"`
+	LowerBound BatchJobSize `yaml:"greaterThan,omitempty" json:"greaterThan"`
```
krisis: @shtripat I have renamed the internal fields so that the methods on them are easy to understand. I have also added unit tests to `BatchJobSizeFilter` methods to cover typical scenarios and to illustrate its usage.

cmd/erasure-server-pool.go
```diff
@@ -2029,7 +2029,10 @@ func (z *erasureServerPools) Walk(ctx context.Context, bucket, prefix string, re
 								return
 							}
 
-							versionsSorter(fivs.Versions).reverse()
+							// Note: entry.fileInfoVersions returns versions sorted in reverse chronological order based on ModTime
+							if opts.WalkVersionsSort == WalkVersionsSortAsc {
```
krisis: I just realized that this PR introduces support for receiving versions of an object in reverse chronological order of `ModTime`s. To minimize the amount of files that need to be touched, I use the 'zero' value of [`WalkVersionsSortOrder`](https://github.com/minio/minio/pull/17946/files#diff-7410eec4482a2c3834d74dd077538d3b377627a3a50e436a372bdd78934ef837R49) to indicate ascending order as default ordering. I'm drawing attention to this since it affects all consumers of the `ObjectLayer.Walk` method.

cmd/batch-expire.go
```diff
@@ -0,0 +1,622 @@
...
+//       olderThan: 70h # match objects older than this value
+//       createdBefore: "2006-01-02T15:04:05.00Z" # match objects created before "date"
+//       tags:
+//         - key: name
+//           value: pick* # match objects with tag 'name', all values starting with 'pick'
+//       metadata:
+//         - key: content-type
+//           value: image/* # match objects with 'content-type', all values starting with 'image/'
+//       size:
+//         lesserThan: "10MiB" # match objects with size lesser than this value (e.g. 10MiB)
```
klauspost: Gramatically it is `lessThan`. I can live with it, though.

cmd/batch-expire.go
```diff
@@ -0,0 +1,622 @@
...
+	}
+
+	if len(ef.Name) > 0 && !wildcard.Match(ef.Name, obj.Name) {
+		return false
+	}
+	if ef.OlderThan > 0 && now.Sub(obj.ModTime) <= ef.OlderThan {
+		return false
+	}
+
+	if !ef.CreatedBefore.IsZero() && (obj.ModTime.Equal(ef.CreatedBefore) || obj.ModTime.After(ef.CreatedBefore)) {
```
klauspost:
```suggestion
	if !ef.CreatedBefore.IsZero() && !obj.ModTime.Before(ef.CreatedBefore) {
```

cmd/batch-expire.go
```diff
@@ -0,0 +1,622 @@
...
+	if p.RetainVersions < 0 {
+		return errors.New("retainVersions must be >= 0")
+	}
+	return nil
+}
+
+// BatchJobExpireFilter holds all the filters currently supported for batch replication
+type BatchJobExpireFilter struct {
+	OlderThan     time.Duration       `yaml:"olderThan,omitempty" json:"olderThan"`
+	CreatedBefore time.Time           `yaml:"createdBefore,omitempty" json:"createdBefore"`
```
klauspost: If you want omitempty to work, it must be:
```suggestion
	CreatedBefore *time.Time          `yaml:"createdBefore,omitempty" json:"createdBefore"`
```

cmd/batch-expire.go
```diff
@@ -0,0 +1,622 @@
...
+		if err := meta.Validate(); err != nil {
+			return err
+		}
+	}
+	if err := ef.Purge.Validate(); err != nil {
+		return err
+	}
+	if err := ef.Size.Validate(); err != nil {
+		return err
+	}
```
klauspost, line 219: Maybe check if `CreatedBefore` is actually in the past.

cmd/batch-expire.go
```diff
@@ -0,0 +1,622 @@
...
+)
+
+type objInfoCache map[string]*ObjectInfo
+
+func newObjInfoCache() objInfoCache {
+	return objInfoCache(make(map[string]*ObjectInfo))
+}
+
+func (oiCache objInfoCache) Add(toDel ObjectToDelete, oi *ObjectInfo) {
+	oiCache[fmt.Sprintf("%s-%s", toDel.ObjectName, toDel.VersionID)] = oi
```
klauspost, line 296: So this can become *huge* is there any control over max size?

cmd/batch-expire.go
```diff
@@ -0,0 +1,622 @@
...
+}
+
+func (oiCache objInfoCache) Get(toDel ObjectToDelete) (*ObjectInfo, bool) {
+	oi, ok := oiCache[fmt.Sprintf("%s-%s", toDel.ObjectName, toDel.VersionID)]
+	return oi, ok
```
klauspost, line 301:
```suggestion
	return oiCache[fmt.Sprintf("%s-%s", toDel.ObjectName, toDel.VersionID)]
```

cmd/batch-expire.go
```diff
@@ -0,0 +1,622 @@
...
+		return err
+	}
+	if err := ef.Size.Validate(); err != nil {
+		return err
+	}
+	return nil
+}
+
+// BatchJobExpireV1 v1 of batch expiration job
+type BatchJobExpireV1 struct {
```
klauspost: Do we realistically have to rewrite for new versions? Maybe just:
```suggestion
type BatchJobExpire struct {
```

cmd/batch-expire.go
```diff
@@ -0,0 +1,622 @@
...
+	retryAttempts := r.Retry.Attempts
+	delay := job.Expire.Retry.Delay
+	if delay == 0 {
+		delay = batchExpireJobDefaultRetryDelay
+	}
+
+	var i int
+	for toExpire := range expireCh {
+		if i > 0 {
+			if wait := globalBatchConfig.Clone().ExpirationWorkersWait; wait > 0 {
```
klauspost: Can't you just add a func that returns the value, so the Clone can be avoided?

cmd/batch-expire.go
```diff
@@ -0,0 +1,622 @@
...
+
+	vc, _ := globalBucketVersioningSys.Get(r.Bucket)
+	retryAttempts := r.Retry.Attempts
+	delay := job.Expire.Retry.Delay
+	if delay == 0 {
+		delay = batchExpireJobDefaultRetryDelay
+	}
+
+	var i int
+	for toExpire := range expireCh {
```
klauspost, line 313: Check if ctx is done?

cmd/batch-expire.go
```diff
@@ -0,0 +1,622 @@
...
+	oiCache[fmt.Sprintf("%s-%s", toDel.ObjectName, toDel.VersionID)] = oi
+}
+
+func (oiCache objInfoCache) Get(toDel ObjectToDelete) (*ObjectInfo, bool) {
+	oi, ok := oiCache[fmt.Sprintf("%s-%s", toDel.ObjectName, toDel.VersionID)]
+	return oi, ok
+}
+
+func batchObjsForDelete(ctx context.Context, r *BatchJobExpireV1, ri *batchJobInfo, job BatchJobRequest, api ObjectLayer, wk *workers.Workers, expireCh <-chan []expireObjInfo) {
+	defer wk.Give()
```
klauspost: Is the corresponding in the caller? Seems a bit strange.

cmd/batch-expire.go
```diff
@@ -0,0 +1,622 @@
...
+					},
+				}
+				toDel = append(toDel, od)
+				oiCache.Add(od, &exp.ObjectInfo)
+			}
+
+			// DeleteObject(deletePrefix: true) to expire all versions of an object
+			for _, exp := range toExpireAll {
+				var success bool
+				for attempts := 1; attempts <= retryAttempts; attempts++ {
```
klauspost, line 354: Break/return on ctx.Done?

cmd/batch-expire.go
```diff
@@ -0,0 +1,622 @@
...
+						success = true
+						break
+					}
+				}
+				ri.trackMultipleObjectVersions(r.Bucket, exp, success)
+			}
+
+			// DeleteMultiple objects
+			toDelCopy := make([]ObjectToDelete, len(toDel))
+			for attempts := 1; attempts <= retryAttempts; attempts++ {
```
klauspost, line 385: ctx Done

cmd/batch-expire.go
```diff
@@ -0,0 +1,622 @@
...
+	ri.Failed = ri.ObjectsFailed > 0
+	globalBatchJobsMetrics.save(job.ID, ri)
+
+	// Close the saverQuitCh - this also triggers saving in-memory state
+	// immediately one last time before we exit this method.
+	close(saverQuitCh)
+
+	// Notify expire jobs final status to the configured endpoint
+	buf, _ := json.Marshal(ri)
+	if err := r.Notify(ctx, bytes.NewReader(buf)); err != nil {
```
klauspost: ctx may be expired here. Maybe
```suggestion
	if err := r.Notify(context.Background(), bytes.NewReader(buf)); err != nil {
```
?

cmd/batch-handlers.go
```diff
@@ -431,23 +435,27 @@ func (r *BatchJobReplicateV1) StartFromSource(ctx context.Context, api ObjectLay
...
 				} else {
-					stopFn(nil)
+					stopFn(oi, nil)
 				}
 				ri.trackCurrentBucketObject(r.Target.Bucket, oi, success)
 				globalBatchJobsMetrics.save(job.ID, ri)
 				// persist in-memory state to disk after every 10secs.
 				logger.LogIf(ctx, ri.updateAfter(ctx, api, 10*time.Second, job))
+
+				if wait := globalBatchConfig.Clone().ReplicationWorkersWait; wait > 0 {
```
klauspost: Avoid clone if feasible.

cmd/batch-handlers.go
```diff
@@ -996,16 +1031,20 @@ func (r *BatchJobReplicateV1) Start(ctx context.Context, api ObjectLayer, job Ba
...
 				} else {
-					stopFn(nil)
+					stopFn(result, nil)
 				}
 				ri.trackCurrentBucketObject(r.Source.Bucket, result, success)
 				globalBatchJobsMetrics.save(job.ID, ri)
 				// persist in-memory state to disk after every 10secs.
 				logger.LogIf(ctx, ri.updateAfter(ctx, api, 10*time.Second, job))
+
+				if wait := globalBatchConfig.Clone().ReplicationWorkersWait; wait > 0 {
```
klauspost: same

klauspost (commented):
Nothing major! Great stuff!

cmd/batch-expire.go
krisis: This 'cache' would have at most 10 entries in it. We batch up to 10 objects to be expired per batch expiration worker - https://github.com/minio/minio/pull/17946/files#diff-75c25e8d16842df078c17f4d7234993a6982cbae3ecbcb81a5d759f4f398d032R498

cmd/batch-expire.go
krisis: ### Context (optional)
The batch job starts at the `BatchJobExpire.Start` method. This is where we build a pipeline of objects to distribute among worker goroutines for expiration. At the source of the pipeline, we have `ObjectLayer.Walk` method sending `ObjectInfo`s downstream. In `BatchJobExpire.Start` we apply the various filters specified in the batch expiry yaml on the incoming `ObjectInfo`s and send `[]expireObjInfo` (viz. simply annotated `[]ObjectInfo`) of 'qualified' objects downstream. 
The `batchObjsForDelete` function is that component in this pipeline which spawns worker goroutines and assigns each of them a `[]expireObjInfo`. This is the fan-out point.

### Answer
To answer your question, I use the same `worker.Workers` object which we use for limited concurrency among worker goroutines to manage the goroutine running the `batchObjsForDelete` function too. Yes, the `wk.Take()` is called by its caller right before we call `go batchObjsForDelete(...)`.

cmd/batch-expire.go
krisis: I shall address this in this PR. I will need to send a PR to madmin-go for this though.

krisis (commented):
I have addressed most of the review comments. I shall address the remaining in another refresh.

cmd/batch-expire.go
krisis: https://github.com/minio/madmin-go/pull/248

harshavardhana:
PTAL @poornas @vadmeste

internal/config/batch/batch.go
```diff
@@ -0,0 +1,151 @@
...
+		cfg.ReplicationWorkersWait = rduration
+	}
+
+	if kduration > 0 {
+		cfg.KeyRotationWorkersWait = kduration
+	}
+
+	if eduration > 0 {
+		cfg.ExpirationWorkersWait = eduration
+	}
```
vadmeste, line 157: time.Duration can take a negative value; let's error out if thats the ccase

cmd/metrics-v2.go
```diff
@@ -3247,6 +3249,77 @@ func getClusterHealthMetrics() *MetricsGroup {
...
+					Value:          objects,
+					VariableLabels: map[string]string{"bucket": bucket, "jobId": mj.JobID},
+				},
+				Metric{
+					Description: MetricDescription{
+						Namespace: bucketMetricNamespace,
+						Subsystem: "batch",
+						Name:      MetricName(jtype + "_objects_failed"),
+						Help:      "Get failed batch job " + jtype + "objects",
+						Type:      gaugeMetric,
```
vadmeste: I think this is a counter metric, not gauge

cmd/metrics-v2.go
```diff
@@ -3247,6 +3249,77 @@ func getClusterHealthMetrics() *MetricsGroup {
...
+				bucket = mj.Expired.Bucket
+			}
+			metrics = append(metrics,
+				Metric{
+					Description: MetricDescription{
+						Namespace: bucketMetricNamespace,
+						Subsystem: "batch",
+						Name:      MetricName(jtype + "_objects"),
+						Help:      "Get successfully completed batch job " + jtype + "objects",
+						Type:      gaugeMetric,
```
vadmeste: I think this is a counter metric, not gauge

cmd/batch-job-common-types.go
```diff
@@ -81,3 +83,52 @@ func (r BatchJobRetry) Validate() error {
 
 	return nil
 }
+
+// BatchJobSizeFilter supports size based filters - LesserThan and GreaterThan
+type BatchJobSizeFilter struct {
+	UpperBound BatchJobSize `yaml:"lessThan,omitempty" json:"lessThan"`
```
vadmeste: I doubt omitempty is necessary  with a struct; I believe only with a pointer

cmd/batch-job-common-types.go
```diff
@@ -81,3 +83,52 @@ func (r BatchJobRetry) Validate() error {
...
+	var batchExpireSz string
+	err := unmarshal(&batchExpireSz)
+	if err != nil {
+		return err
+	}
+	sz, err := humanize.ParseBytes(batchExpireSz)
+	if err != nil {
+		return err
+	}
+	*s = BatchJobSize(sz)
```
vadmeste, line 161: I guess check if batch job size is >= 0

cmd/batch-expire.go
```diff
@@ -0,0 +1,650 @@
...
+	case BatchJobExpireObject:
+		if obj.DeleteMarker {
+			return false
+		}
+	case BatchJobExpireDeleted:
+		if !obj.DeleteMarker {
+			return false
+		}
+	default:
+		// we should never come here, Validate should have caught this.
```
vadmeste, line 118: maybe let's panic()

cmd/batch-expire.go
```diff
@@ -0,0 +1,650 @@
...
+
+	}
+	if len(ef.Metadata) > 0 && !obj.DeleteMarker {
+		for _, kv := range ef.Metadata {
+			// Object (version) must match all x-amz-meta and
+			// standard metadata headers
+			// specified in the filter
+			var match bool
+			for k, v := range obj.UserDefined {
+				if !stringsHasPrefixFold(k, "x-amz-meta-") && !isStandardHeader(k) {
```
vadmeste, line 167: Why is this needed ? we can leave the user to specify anything

cmd/batch-expire.go
```diff
@@ -0,0 +1,650 @@
...
+	if err := ri.load(ctx, api, job); err != nil {
+		return err
+	}
+
+	globalBatchJobsMetrics.save(job.ID, ri)
+	lastObject := ri.Object
+
+	now := time.Now().UTC()
+
+	workerSize, err := strconv.Atoi(env.Get("_MINIO_BATCH_EXPIRATION_WORKERS", strconv.Itoa(runtime.GOMAXPROCS(0)/2)))
```
vadmeste, line 457: Don't we have a configuration for this ?

cmd/batch-expire.go
```diff
@@ -0,0 +1,650 @@
...
+	go batchObjsForDelete(ctx, r, ri, job, api, wk, expireCh)
+
+	var (
+		prevObj       ObjectInfo
+		matchedFilter BatchJobExpireFilter
+		versionsCount int
+		toDel         []expireObjInfo
+	)
+	for result := range results {
+		result := result
```
vadmeste: This doesn't seem needed

cmd/batch-expire.go
```diff
@@ -0,0 +1,650 @@
...
+				for attempts := 1; attempts <= retryAttempts; attempts++ {
+					select {
+					case <-ctx.Done():
+						done = true
+					default:
+					}
+					stopFn := globalBatchJobsMetrics.trace(batchJobMetricExpire, ri.JobID, attempts)
+					_, err := api.DeleteObject(ctx, exp.Bucket, exp.Name, ObjectOptions{
+						DeletePrefix: true,
+						NoLock:       true,
```
vadmeste: Why no lock ?

vadmeste (commented):
Few comments, not tested yet

harshavardhana:
Please rebase the PR @krisis

cmd/batch-expire.go
krisis: It was decided that we should only ever increase batch-expiration workers in exceptional cases. Making this available as a configuration will encourage the user behavior we wish to avoid. /cc @harshavardhana

cmd/batch-job-common-types.go
krisis: https://pkg.go.dev/github.com/dustin/go-humanize#ParseBytes returns an error if input string contains negative values.

cmd/batch-expire.go
krisis: This object is past its expiry and applications aren't expected to access such objects. It is possible though that an application may be accessing this object while it is being expired. This unlikely application failure is acceptable in this context.

klauspost: Probably better with a lock to avoid edge cases.

cmd/batch-expire.go
klauspost: Maybe move it in here, so it is symmetric - makes the "preconditions" of the function easier to understand.

cmd/batch-expire.go
```diff
@@ -0,0 +1,658 @@
...
+			return batchExpireJobError{
+				Code:           "NoSuchSourceBucket",
+				Description:    "The specified source bucket does not exist",
+				HTTPStatusCode: http.StatusNotFound,
+			}
+		}
+		return err
+	}
+
+	if len(r.Rules) > 10 {
```
klauspost: Why this (seemingly) arbitrary limit?

krisis: The intention here is to have an upper bound on the number of rules. 10 is definitely arbitrary and small. We could keep it higher, say 1000, similar to the limit on number of rules in an ILM policy.

klauspost: Yeah... 1000 would be insane.

krisis: Should we settle at 100 then? :)

cmd/batch-expire.go
krisis: Agree. It is better to be safe than sorry here.

cmd/batch-expire.go
harshavardhana: 50 is fine @krisis

krisis: Updated the maximum rules in a batch expiry job to 50.

cmd/batch-expire.go
harshavardhana: @krisis ^^

cmd/batch-expire.go
harshavardhana: @krisis ^^

cmd/batch-expire.go
harshavardhana: WE do not want them to come up internal headers and somehow match them, this is a way to prevent such mistakes.

cmd/batch-expire.go
krisis: panic() seems a little too much given that we would (safely) ignore rules with invalid filters. I can add an error log to record that something is amiss. Does that sound ok @vadmeste ?

cmd/batch-expire.go
krisis: It dawned on me that `batchObjsForDelete`'s lifetime is bound to `expireCh`. i.e once `expireCh` channel is closed, we have no more objects to work. Thanks for catching this code smell, which turned out to be superfluous.

harshavardhana (approved):
LGTM and tested

PR #17958, opened 2023-08-31
feat: SSE-KMS use uuid instead of read all data to md5.

jiuker:

## Motivation and Context

md5sum is very expensive
We change a little at `PutObjectHandler`. If it works. Will change to other handler.

## Types of changes

## Checklist:


harshavardhana:
@aead your review here is a must. Hopefully, you may like this new approach. It encourages more SSE-KMS usage.

cmd/object-handlers.go
```diff
@@ -1758,8 +1758,10 @@ func (api objectAPIHandlers) PutObjectHandler(w http.ResponseWriter, r *http.Req
 		md5hex = "" // Do not try to verify the content.
 		sha256hex = ""
 	}
-
-	hashReader, err := hash.NewReader(reader, size, md5hex, sha256hex, actualSize)
+	if crypto.S3KMS.IsRequested(r.Header) {
```
harshavardhana: We should also make sure `content-md5` is not requested, if content-md5 is requested we must ensure that we calculate `md5sum` as per protocol regardless of SSE-KMS is specified or not.

harshavardhana (changes requested):

cmd/object-handlers.go
```diff
@@ -1757,8 +1757,10 @@ func (api objectAPIHandlers) PutObjectHandler(w http.ResponseWriter, r *http.Req
 		md5hex = "" // Do not try to verify the content.
 		sha256hex = ""
 	}
-
-	hashReader, err := hash.NewReader(reader, size, md5hex, sha256hex, actualSize)
+	if crypto.S3KMS.IsNotContentMD5Requested(r.Header) {
```
harshavardhana: Another thing I realized is that we can do this for SSE-C requests as well so you can do 

```
if crypto.S3KMS.NoMD5(r.Header) || crypto.SSEC.NoMD5(r.Header) {
     ctx = context.WithValue(ctx, etag.ContextEtagAsUUID, mustGetUUID())
}
```

harshavardhana (changes requested):

internal/etag/etag.go
```diff
@@ -263,6 +263,12 @@ func FromContentMD5(h http.Header) (ETag, error) {
 	return ETag(b), nil
 }
 
+// IsNotContentMD5Requested - for http.request.header is not request Content-Md5
+func IsNotContentMD5Requested(h http.Header) bool {
```
harshavardhana: We can keep the function name a bit simpler

```
func ContentMD5Requested(h http.Header) bool {
   return ok
```

Now use this as inverse in the check

```
!etag.ContentMD5Requested(r.Header) && (crypto.S3KMS.IsRequested(r.Header) || crypto.SSEC.IsRequested(r.Header)) {
```

harshavardhana (changes requested):

internal/etag/reader.go
```diff
@@ -153,3 +167,41 @@ type VerifyError struct {
...
+	return hasher.Sum([]byte(u.uuid))
+}
+
+// Reset -  implement hash.Hash Reset
+func (u UUIDHash) Reset() {
+	return
+}
+
+// Size -  implement hash.Hash Size
+func (u UUIDHash) Size() int {
```
aead, line 186: `Size` must return `16` since we use `md5.New().Sum`. Otherwise, there are subtle bugs possible.

aead (changes requested):

internal/etag/reader.go
```diff
@@ -153,3 +167,41 @@ type VerifyError struct {
 func (v VerifyError) Error() string {
 	return fmt.Sprintf("etag: expected ETag %q does not match computed ETag %q", v.Expected, v.Computed)
 }
+
+// UUIDHash - use uuid to make md5sum
+type UUIDHash struct {
```
harshavardhana, line 166:
```suggestion
type UUIDHash struct {
         hash.Hash
```

and overriding relevant methods is enough? @aead

harshavardhana (changes requested):

harshavardhana:
PTAL @aead ?

harshavardhana:
PTAL @aead

harshavardhana:
@aead your 👍🏼 is a must here, please take a look when you can.  The performance boost from this is significant in most common scenarios for SSE-KMS.

Especially on Optane and NVMe drives the performance doubles for WRITEs - it also makes a much stronger case to use SSE-KMS you would see real performance benefits.

aead (approved):

harshavardhana (approved):

cmd/object-handlers.go
```diff
@@ -1757,8 +1757,11 @@ func (api objectAPIHandlers) PutObjectHandler(w http.ResponseWriter, r *http.Req
 		md5hex = "" // Do not try to verify the content.
 		sha256hex = ""
 	}
-
-	hashReader, err := hash.NewReader(reader, size, md5hex, sha256hex, actualSize)
+	// SSE-KVM and SSE-C if not request Content-Md5. Simple use md5.Sum(uuid) as etag.
+	if !etag.ContentMD5Requested(r.Header) && (crypto.S3KMS.IsRequested(r.Header) || crypto.SSEC.IsRequested(r.Header)) {
+		ctx = context.WithValue(ctx, etag.ContextEtagAsUUID, mustGetUUID())
```
klauspost: We shouldn't start passing functional parameters as context values.

If we do that, we will end up with a huge amount of magic code that does things that are hard to follow.

Pass a regular parameter. You are welcome to add a `type hash.ReaderOptions struct` that combines the parameters, but you can also just add another one.

internal/etag/reader.go
```diff
@@ -153,3 +167,41 @@ type VerifyError struct {
...
+}
+
+// Reset -  implement hash.Hash Reset
+func (u UUIDHash) Reset() {
+	return
+}
+
+// Size -  implement hash.Hash Size
+func (u UUIDHash) Size() int {
+	return md5.Size
```
klauspost: You are returning `len(u.uuid)` in Sum(...), which is the encoded length, not the raw length.

But you probably want to return the raw value anyway.

internal/etag/reader.go
```diff
@@ -153,3 +167,41 @@ type VerifyError struct {
...
+
+// Write -  implement hash.Hash Write
+func (u UUIDHash) Write(p []byte) (n int, err error) {
+	return len(p), nil
+}
+
+// Sum -  implement md5.Sum
+func (u UUIDHash) Sum(b []byte) []byte {
+	hasher := md5.New()
+	return hasher.Sum([]byte(u.uuid))
```
klauspost: `u.uuid` is the string encoding, not the UUID values. Here is how you get the raw UUID:

```Go
// mustGetUUIDBytes - get a random UUID as 16 bytes unencoded.
func mustGetUUIDBytes() []byte {
	u, err := uuid.NewRandom()
	if err != nil {
		logger.CriticalIf(GlobalContext, err)
	}
	return u[:]
}
```

Alternatively you can parse the UUID, but that seems wasteful.

internal/etag/reader.go
jiuker: Not really. md5.Size is a const values which is the encoded length @klauspost

internal/etag/reader.go
jiuker: Yes, It comes from the ctx. `u.uuid = mustGetUUID()`. This is external random generation, I think it can be decided externally here. If the UUID passed in is the same, the value of ETAG will be the same. @klauspost

klauspost: Yes. That is the *encoded* string, not a 16 byte.

The encoded version looks like this: `c59a75c1-e28d-4d07-a3ec-1e7ff63b8ceb` - the UUID values are just 16 values in a `[]byte`  - as is co-incidentally the MD5 before encoding.

cmd/object-handlers.go
harshavardhana: Yeah I thought about it but it may or may not be okay here. But I agree we can change the function signature we are changing it anyways change it to take

```
opts := hash.ReaderOptions{
}

hash.NewReader(ctx, r.Body, opts)
```

Is perhaps a more idiomatic approach @jiuker

jiuker:
@harshavardhana @klauspost @aead PTAL

cmd/object-handlers.go
```diff
@@ -1757,8 +1757,23 @@ func (api objectAPIHandlers) PutObjectHandler(w http.ResponseWriter, r *http.Req
...
+	// SSE-KVM and SSE-C if not request Content-Md5. Simple use md5.Sum(uuid) as etag.
+	if !etag.ContentMD5Requested(r.Header) && (crypto.S3KMS.IsRequested(r.Header) || crypto.SSEC.IsRequested(r.Header)) {
+		hashReader, err = hash.NewReaderWithOpts(ctx, reader, hash.Options{
+			Size:       size,
+			MD5Hex:     md5hex,
+			SHA256Hex:  sha256hex,
+			ActualSize: actualSize,
+			DisableMD5: false,
+			ContentMD5: hash.ContentMD5Opt{
+				NotRequest: true,
```
harshavardhana: Instead of Not request name it as 

```suggestion
				Requested: false,
```

cmd/object-handlers.go
```diff
@@ -1757,8 +1757,23 @@ func (api objectAPIHandlers) PutObjectHandler(w http.ResponseWriter, r *http.Req
...
-	hashReader, err := hash.NewReader(reader, size, md5hex, sha256hex, actualSize)
+	var hashReader *hash.Reader
+	// SSE-KVM and SSE-C if not request Content-Md5. Simple use md5.Sum(uuid) as etag.
+	if !etag.ContentMD5Requested(r.Header) && (crypto.S3KMS.IsRequested(r.Header) || crypto.SSEC.IsRequested(r.Header)) {
+		hashReader, err = hash.NewReaderWithOpts(ctx, reader, hash.Options{
+			Size:       size,
+			MD5Hex:     md5hex,
+			SHA256Hex:  sha256hex,
+			ActualSize: actualSize,
+			DisableMD5: false,
```
harshavardhana, line 1774: This should be true right?

harshavardhana (changes requested):

cmd/object-handlers.go
jiuker: Not really. False.

cmd/object-handlers.go
klauspost: @jiuker In general, it is good practice to never name your bools as a negative. 

When you do, you easily end up with things that are hard to read.

`if !NotRequested {}` takes a lot harder to read than `if Requested` - and it gets worse when you combine it and `&&` and `||` in a longer sequence.

There are exceptions. But that is what we should try to do when we can.

Naming is hard - even when you've been programming a long time. And as for most of us where English is a second or third language it is even harder.

internal/etag/reader.go
```diff
@@ -153,3 +161,42 @@ type VerifyError struct {
...
+func (u UUIDHash) Write(p []byte) (n int, err error) {
+	return len(p), nil
+}
+
+// Sum -  implement md5.Sum
+func (u UUIDHash) Sum(b []byte) []byte {
+	hasher := md5.New()
+	hasher.Write(u.uuid)
+	return hasher.Sum(nil)
+}
```
klauspost:
```suggestion
func (u UUIDHash) Sum(b []byte) []byte {
	return u.uuid
}
```

You are hashing the uuid - not needed.

internal/etag/reader.go
```diff
@@ -153,3 +161,42 @@ type VerifyError struct {
...
+
+// Reset -  implement hash.Hash Reset
+func (u UUIDHash) Reset() {
+	return
+}
+
+// Size -  implement hash.Hash Size
+func (u UUIDHash) Size() int {
+	return md5.Size
```
klauspost:
```suggestion
	return len(u.uuid)
```

cmd/object-handlers.go
klauspost: Is this needed at all? Can't we just check if UUID is set?

You can just remove the struct and call it `ForceMD5 []byte`.

jiuker:
@harshavardhana applied @klauspost suggestions. That should be fine. PTAL.

harshavardhana (approved):

cmd/object-handlers.go
```diff
@@ -1757,8 +1757,20 @@ func (api objectAPIHandlers) PutObjectHandler(w http.ResponseWriter, r *http.Req
 		md5hex = "" // Do not try to verify the content.
 		sha256hex = ""
 	}
-
-	hashReader, err := hash.NewReader(reader, size, md5hex, sha256hex, actualSize)
+	var hashReader *hash.Reader
+	// SSE-KVM and SSE-C if not request Content-Md5. Simple use md5.Sum(uuid) as etag.
```
```suggestion
	// Optimization: If SSE-KMS and SSE-C did not request Content-Md5. Use md5.Sum(uuid) as etag.
```

internal/etag/etag.go
```diff
@@ -263,6 +263,12 @@ func FromContentMD5(h http.Header) (ETag, error) {
 	return ETag(b), nil
 }
 
+// ContentMD5Requested - for http.request.header is not request Content-Md5
+func ContentMD5Requested(h http.Header) bool {
+	_, ok := h["Content-Md5"]
```
harshavardhana:
```suggestion
	_, ok := h[xhttp.ContentMD5]
```

harshavardhana: Please use 

```
import xhttp "github.com/minio/minio/internal/http"
```

harshavardhana (changes requested):

harshavardhana (approved):

klauspost (approved):

PR #17986, opened 2023-09-06
fix: drive rotational calculation status for partitions

vadmeste:

## Description
If a MinIO drive path is mounted to a partition and not a real disk, 
getting the rotational status would fail because Linux does not expose
 that status to partition; 

In other words, /sys/block/drive-partition-name/queue/rotational does not exist;

To fix the issue, the code will search for rotational status of the disk 
that hosts the partition, and this can be calculated from the realpath 
of /sys/class/block/<drive-partition-name>

## Motivation and Context
Fix the calculation of rotational in some corner cases

## Types of changes

## Checklist:

### DIFF
--- a/internal/disk/stat_linux.go
+++ b/internal/disk/stat_linux.go
@@ -24,6 +24,7 @@ import (
 	"bufio"
 	"fmt"
 	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"syscall"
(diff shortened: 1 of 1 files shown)


internal/disk/stat_linux.go
```diff
@@ -71,21 +72,33 @@ func GetInfo(path string, firstTime bool) (info Info, err error) {
...
 					continue
 				}
-				qst, err := bfs.SysBlockDeviceQueueStats(dstat.DeviceName)
-				if err != nil {
-					continue
-				}
-				rot := qst.Rotational == 1 // Rotational is '1' if the device is HDD
 				if dstat.MajorNumber == info.Major && dstat.MinorNumber == info.Minor {
-					info.Name = dstat.DeviceName
+					devName = dstat.DeviceName
```
harshavardhana, line 83: break once you have a devName @vadmeste

harshavardhana (changes requested):

internal/disk/stat_linux.go
vadmeste: > break once you have a devName @vadmeste

done

harshavardhana (approved):

PR #18016, opened 2023-09-12
fix: webhook notify endpoint with standard ports

vadmeste:

## Description
If the endpoint in the notify webhook configuration does not contain 
any port number, the webhook will not work. This commit fixes it by 
adding :80 or :443 in the host; isActive() needs a port number to test 
if the webhook port is open or not.

## Motivation and Context
Fix webhook behavior when the port number is not provided in the endpoint

## Types of changes

## Checklist:

### DIFF
--- a/internal/event/target/webhook.go
+++ b/internal/event/target/webhook.go
@@ -104,6 +104,8 @@ type WebhookTarget struct {
 	loggerOnce logger.LogOnce
 	cancel     context.CancelFunc
 	cancelCh   <-chan struct{}
+
+	addr string // full address ip/dns with a port number, e.g.  x.x.x.x:8080
 }
 
 // ID - returns target ID.
@@ -130,7 +132,7 @@ func (target *WebhookTarget) Store() event.TargetStore {
 }
 
 func (target *WebhookTarget) isActive() (bool, error) {
-	conn, err := net.DialTimeout("tcp", target.args.Endpoint.Host, 5*time.Second)
+	conn, err := net.DialTimeout("tcp", target.addr, 5*time.Second)
 	if err != nil {
 		if xnet.IsNetworkOrHostDown(err, false) {
 			return false, store.ErrNotConnected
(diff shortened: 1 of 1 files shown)


jiuker (approved):

internal/event/target/webhook.go
```diff
@@ -281,6 +281,16 @@ func NewWebhookTarget(ctx context.Context, id string, args WebhookArgs, loggerOn
...
+	// Add :80 or :443 to the endpoint Host since the port is used in isActive() function
+	if _, _, err := net.SplitHostPort(args.Endpoint.Host); err != nil && strings.Contains(err.Error(), "missing port in address") {
+		switch strings.ToLower(args.Endpoint.Scheme) {
+		case "http":
+			args.Endpoint.Host += ":80"
+		case "https":
+			args.Endpoint.Host += ":443"
+		}
+	}
+
```
harshavardhana: strange why is this not automatically added by httpClient?

```
package main

import "net/http"
import "fmt"

func main() {
	fmt.Println(http.Get("https://www.google.com"))
}
```

harshavardhana (changes requested):

internal/event/target/webhook.go
vadmeste: The issue is more in isActive() in internal/event/target/webhook.go where it is creating a TCP connection witout specifying a port number; I did not want to change isActive() to not be too much distruptive; but I am not sure why we are not doing a simple GET request in isActive() function.

harshavardhana: I think that change was by @Praveenrajmani - can you tell us ?

harshavardhana (approved):

internal/event/target/webhook.go
Praveenrajmani: We were doing an empty `{}` request body POST request. This ended up creating too many empty requests causing some parsing failure at the receiver end. 

We changed to TCP dials because of this.

I think we can have empty `{}` requests only because the TCP dials will not bypass the loadbalancer.

Praveenrajmani: https://github.com/minio/minio/commit/443250d1354a7022af76587ea55cbd82470ce1d8

PR #18026, opened 2023-09-13
Move IAM periodic ops to a single go routine

donatello:

## Description

This helps reduce disk operations as these periodic routines would not
run concurrently any more.

Also add expired STS purging periodic operation: Since we do not scan
the on-disk STS credentials (and instead only load them on-demand) a
separate routine is needed to purge expired credentials from storage.
Currently this runs about a quarter as often as IAM refresh.

Also fix a bug where with etcd, STS accounts could get loaded into the
iamUsersMap instead of the iamSTSAccountsMap.

## Motivation and Context

Reduce IOPS related to IAM and add periodic STS creds cleanup operation.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/iam-object-store.go
+++ b/cmd/iam-object-store.go
@@ -32,6 +32,7 @@ import (
 	"github.com/minio/minio-go/v7/pkg/set"
 	"github.com/minio/minio/internal/config"
 	"github.com/minio/minio/internal/kms"
+	"github.com/minio/minio/internal/logger"
 )
 
 // IAMObjectStore implements IAMStorageAPI
@@ -381,6 +382,32 @@ func (iamOS *IAMObjectStore) listAllIAMConfigItems(ctx context.Context) (map[str
 		}
 	}
 	return res, nil
+}
+
+// PurgeExpiredSTS - purge expired STS credentials from object store.
+func (iamOS *IAMObjectStore) PurgeExpiredSTS(ctx context.Context) error {
+	if iamOS.objAPI == nil {
+		return errServerNotInitialized
+	}
+
+	bootstrapTraceMsg("purging expired STS credentials")
+	// Scan STS users on disk and purge expired ones. We do not need to hold a
+	// lock with store.lock() here.
+	for item := range listIAMConfigItems(ctx, iamOS.objAPI, iamConfigPrefix+SlashSeparator+stsListKey) {
+		if item.Err != nil {
+			return item.Err
+		}
+		userName := path.Dir(item.Item)
+		// loadUser() will delete expired user during the load - we do not need
+		// to keep the loaded user around in memory, so we reinitialize the map
+		// each time.
+		m := map[string]UserIdentity{}
+		if err := iamOS.loadUser(ctx, userName, stsUser, m); err != nil && err != errNoSuchUser {
+			logger.LogIf(GlobalContext, fmt.Errorf("unable to load user during STS purge: %w (%s)", err, item.Item))
+		}
+
+	}
+	return nil
 }
 
 // Assumes cache is locked by caller.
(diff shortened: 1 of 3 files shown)


harshavardhana (dismissed):
I think we should avoid new goroutines that again list and wake up the drive. We do IAM load anyways in a routine we can keep expiring credentials in that @donatello

donatello:
> I think we should avoid new goroutines that again list and wake up the drive. We do IAM load anyways in a routine we can keep expiring credentials in that @donatello

The IAM load avoids listing on the STS credentials path, and this new routine only lists that path, and runs only half as often as IAM load. It really doesn't need to run with every IAM load - what you are proposing is more disk activity.

donatello:
We can make this purging routine run after an IAM load so that they wont run simultaneously @harshavardhana ?

harshavardhana:
> We can make this purging routine run after an IAM load so that they wont run simultaneously @harshavardhana ?

Yeah that would be better in the same goroutine, at any given point in time we need to make sure we are not concurrently going through the disks loading IAM and expiring simultaneously.

I am trying to see what is the less IOPs path.

donatello:
PTAL @harshavardhana

donatello:
Added a bugfix for a bug found from the CI tests.

harshavardhana (approved):

cmd/iam-object-store.go
```diff
@@ -383,6 +384,32 @@ func (iamOS *IAMObjectStore) listAllIAMConfigItems(ctx context.Context) (map[str
...
+	if iamOS.objAPI == nil {
+		return errServerNotInitialized
+	}
+
+	bootstrapTraceMsg("purging expired STS credentials")
+	// Scan STS users on disk and purge expired ones. We do not need to hold a
+	// lock with store.lock() here.
+	for item := range listIAMConfigItems(ctx, iamOS.objAPI, iamConfigPrefix+SlashSeparator+stsListKey) {
+		if item.Err != nil {
+			return item.Err
```
vadmeste, line 398: Maybe, let's only log this

cmd/iam-store.go
```diff
@@ -536,13 +546,13 @@ func (store *IAMStoreSys) LoadIAMCache(ctx context.Context) error {
 
 		bootstrapTraceMsg("loading STS users")
 		// load STS temp users
-		if err := store.loadUsers(ctx, stsUser, newCache.iamUsersMap); err != nil {
+		if err := store.loadUsers(ctx, stsUser, newCache.iamSTSAccountsMap); err != nil {
```
vadmeste, line 549: Does this change fix an issue in master ?

vadmeste (commented):
Few comments

cmd/iam-object-store.go
donatello: This is logged by a caller lower in the stack. We cannot continue on this error because the returned channel gets closed, and the range loop will end.

cmd/iam-store.go
donatello: Yes, it is mentioned in the commit also:
> Also fix a bug where with etcd, STS accounts could get loaded into the
> iamUsersMap instead of the iamSTSAccountsMap.

The bug happens in etcd only - we will be deprecating it too, but this will fix the problem for now.

vadmeste (approved):
LGTM

Scapal:
@donatello , we are still suffering from this bug where I have STS user mappings in the user_mappings table (when exporting the cluster iam).
I'm running the latest MinIO release. The new changes are now indeed made in stsuser_mappings but it seems that user_mappings take precedence so that all OIDC users are stuck with their old policies applied.
Is there a way to clean that up so that user_mappings only contains mappings to local users? We only have 1 local user in case the OIDC IDP has configuration issues.
We are not using etcd.

PR #18029, opened 2023-09-14
add backups for usage-caches to rely on upon error

harshavardhana:

## Description
add backups for usage-caches to rely on upon error

## Motivation and Context
This allows the scanner to avoid lengthy scans, skip
things appropriately and also not lose metrics in
any manner.

reduce longer deadlines for usage-cache loads/saves 
to match the disk timeout which is 2 minutes now per 
IOP.

## How to test this PR?
This is more like a fallback plan, tested it locally
by corrupting the usage caches it would fallback
and load existing values. Rely on them for subsequent
scanner cycles.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/bucket-replication.go
+++ b/cmd/bucket-replication.go
@@ -1501,9 +1501,18 @@ func replicateObjectWithMultipart(ctx context.Context, c *minio.Core, bucket, ob
 	var uploadedParts []minio.CompletePart
 	// new multipart must not set mtime as it may lead to erroneous cleanups at various intervals.
 	opts.Internal.SourceMTime = time.Time{} // this value is saved properly in CompleteMultipartUpload()
-	nctx, cancel := context.WithTimeout(ctx, 5*time.Minute)
-	defer cancel()
-	uploadID, err := c.NewMultipartUpload(nctx, bucket, object, opts)
+	var uploadID string
+	attempts := 1
+	for attempts <= 3 {
+		nctx, cancel := context.WithTimeout(ctx, time.Minute)
+		uploadID, err = c.NewMultipartUpload(nctx, bucket, object, opts)
+		cancel()
+		if err == nil {
+			break
+		}
+		attempts++
+		time.Sleep(time.Duration(rand.Int63n(int64(time.Second))))
+	}
 	if err != nil {
 		return err
 	}
@@ -1524,7 +1533,7 @@ func replicateObjectWithMultipart(ctx context.Context, c *minio.Core, bucket, ob
 					fmt.Errorf("trying %s: Unable to cleanup failed multipart replication %s on remote %s/%s: %w - this may consume space on remote cluster",
 						humanize.Ordinal(attempts), uploadID, bucket, object, aerr))
 				attempts++
-				time.Sleep(time.Second)
+				time.Sleep(time.Duration(rand.Int63n(int64(time.Second))))
 			}
 		}
 	}()
--- a/cmd/data-usage-cache.go
+++ b/cmd/data-usage-cache.go
@@ -38,6 +38,7 @@ import (
 	"github.com/minio/minio/internal/hash"
 	"github.com/minio/minio/internal/logger"
 	"github.com/tinylib/msgp/msgp"
+	"github.com/valyala/bytebufferpool"
 )
 
 //go:generate msgp -file $GOFILE -unexported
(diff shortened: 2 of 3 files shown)


cmd/bucket-replication.go
```diff
@@ -1501,11 +1501,17 @@ func replicateObjectWithMultipart(ctx context.Context, c *minio.Core, bucket, ob
...
+	for attempts <= 3 {
+		nctx, cancel := context.WithTimeout(ctx, time.Minute)
+		uploadID, err = c.NewMultipartUpload(nctx, bucket, object, opts)
+		cancel()
+		if err == nil {
+			break
+		}
+		attempts++
+		time.Sleep(time.Duration(rand.Int63n(int64(time.Second))))
 	}
```
harshavardhana: I did this to ensure that we can at least "3" times to get this going @poornas

cmd/data-usage-cache.go
```diff
@@ -967,47 +977,48 @@ var maxConcurrentScannerSaves = make(chan struct{}, 4)
 // save the content of the cache to minioMetaBackgroundOpsBucket with the provided name.
 // Note that no locking is done when saving.
 func (d *dataUsageCache) save(ctx context.Context, store objectIO, name string) error {
-	var r io.Reader
-	maxConcurrentScannerSaves <- struct{}{}
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	case maxConcurrentScannerSaves <- struct{}{}:
+	}
```
harshavardhana, line 983: I did this to ensure that we can honor top-level contexts and not proceed with writing to disks while we are shutting down

cmd/data-usage-cache.go
```diff
@@ -967,47 +977,48 @@ var maxConcurrentScannerSaves = make(chan struct{}, 4)
...
-	size := int64(-1)
-	if len(d.Cache) > 10000 {
-		pr, pw := io.Pipe()
-		go func() {
-			pw.CloseWithError(d.serializeTo(pw))
-		}()
-		defer pr.Close()
-		r = pr
-	} else {
-		var buf bytes.Buffer
```
harshavardhana, line 985: Let's just use only one way to do this.

poornas (approved):

cmd/data-usage-cache.go
```diff
@@ -927,34 +926,45 @@ type objectIO interface {
...
-			*d = dataUsageCache{}
-			return nil
+			return false, toObjectErr(err, dataUsageBucket, name)
+		}
+		err = d.deserialize(r)
+		r.Close()
+		if err != nil {
+			return true, nil
+		}
+		return false, nil
```
klauspost:
```suggestion
		return err != nil, nil
```

cmd/data-usage-cache.go
klauspost: Actually, with the compaction, choosing the buffered approach makes much more sense than the streaming.

cmd/bucket-replication.go
```diff
@@ -1501,9 +1501,18 @@ func replicateObjectWithMultipart(ctx context.Context, c *minio.Core, bucket, ob
 	var uploadedParts []minio.CompletePart
 	// new multipart must not set mtime as it may lead to erroneous cleanups at various intervals.
 	opts.Internal.SourceMTime = time.Time{} // this value is saved properly in CompleteMultipartUpload()
-	nctx, cancel := context.WithTimeout(ctx, 5*time.Minute)
-	defer cancel()
-	uploadID, err := c.NewMultipartUpload(nctx, bucket, object, opts)
+	var uploadID string
```
klauspost, line 1504: Bonus change?

cmd/data-usage-cache.go
```diff
@@ -967,47 +977,48 @@ var maxConcurrentScannerSaves = make(chan struct{}, 4)
...
+		case <-ctx.Done():
+		case <-maxConcurrentScannerSaves:
 		}
-		r = &buf
-		size = int64(buf.Len())
-	}
+	}()
 
-	hr, err := hash.NewReader(r, size, "", "", size)
+	// If big, do streaming...
```
klauspost: Use the other. We no longer have huge caches since they auto-compact now.

cmd/bucket-replication.go
harshavardhana: yes @klauspost as part of changing context timeouts 5 min timeouts are too long anyways.

cmd/data-usage-cache.go
harshavardhana: are you sure? @klauspost because I think avoiding memory alloc whenever possible is a good idea anyway, because the scanner is so relied upon - I really don't want to have another high memory-related problem.

klauspost (approved):
LGTM

PR #18051, opened 2023-09-18
cache usage, prefix-usage, and buckets for AccountInfo up to 10 secs

harshavardhana:

## Description
cache usage, prefix-usage, and buckets for AccountInfo up to 10 secs

## Motivation and Context
AccountInfo is quite frequently called by Console UI login
attempts, when many users are logging in it is important
that we provide them with better responsiveness.
    
- ListBuckets information is cached every second
- Bucket usage info is cached up to 10 secs
- Prefix usage (optional) info is cached up to 10 secs
    
Failure to update after cache expiration, would still
allow login which would end up providing information
previously cached.
    
This allows for seamless responsiveness for the Console UI
logins, and overall responsiveness on a heavily loaded
system.

## How to test this PR?
Nothing should change other than Console UI logins
would be more responsive on loaded systems.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/admin-handlers-users.go
+++ b/cmd/admin-handlers-users.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2015-2021 MinIO, Inc.
+// Copyright (c) 2015-2023 MinIO, Inc.
 //
 // This file is part of MinIO Object Storage stack
 //
@@ -19,6 +19,7 @@ package cmd
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -1192,7 +1193,7 @@ func (a adminAPIHandlers) DeleteServiceAccount(w http.ResponseWriter, r *http.Re
 	writeSuccessNoContent(w)
 }
 
-// AccountInfoHandler returns usage
+// AccountInfoHandler returns usage, permissions and other bucket metadata for incoming us
 func (a adminAPIHandlers) AccountInfoHandler(w http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
 
@@ -1261,12 +1262,30 @@ func (a adminAPIHandlers) AccountInfoHandler(w http.ResponseWriter, r *http.Requ
 		return rd, wr
 	}
 
-	// Load the latest calculated data usage
-	dataUsageInfo, _ := loadDataUsageFromBackend(ctx, objectAPI)
+	bucketStorageCache.Once.Do(func() {
+		// Set this to 10 secs since its enough, as scanner
+		// does not update the bucket usage values frequently.
+		bucketStorageCache.TTL = 10 * time.Second
+
+		// Rely on older value if usage loading fails from disk.
+		bucketStorageCache.Relax = true
+		bucketStorageCache.Update = func() (interface{}, error) {
+			ctx, done := context.WithTimeout(context.Background(), 2*time.Second)
+			defer done()
+
+			return loadDataUsageFromBackend(ctx, objectAPI)
+		}
+	})
+
+	var dataUsageInfo DataUsageInfo
+	v, _ := bucketStorageCache.Get()
+	if v != nil {
+		dataUsageInfo, _ = v.(DataUsageInfo)
+	}
 
 	// If etcd, dns federation configured list buckets from etcd.
-	var buckets []BucketInfo
 	var err error
+	var buckets []BucketInfo
 	if globalDNSConfig != nil && globalBucketFederation {
 		dnsBuckets, err := globalDNSConfig.List()
 		if err != nil && !IsErrIgnored(err,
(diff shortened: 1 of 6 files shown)


cmd/admin-handlers-users.go
```diff
@@ -1261,12 +1262,30 @@ func (a adminAPIHandlers) AccountInfoHandler(w http.ResponseWriter, r *http.Requ
 		return rd, wr
 	}
 
-	// Load the latest calculated data usage
-	dataUsageInfo, _ := loadDataUsageFromBackend(ctx, objectAPI)
+	bucketStorageCache.Once.Do(func() {
+		// Set this to 10 secs since its enough, as scanner
+		// does not update the bucket usage values frequently.
+		bucketStorageCache.TTL = 10 * time.Second
```
vadmeste, line 1268: I think let's increase this to 30 seconds

cmd/admin-handlers-users.go
```diff
@@ -1261,12 +1262,30 @@ func (a adminAPIHandlers) AccountInfoHandler(w http.ResponseWriter, r *http.Requ
...
+
+		// Rely on older value if usage loading fails from disk.
+		bucketStorageCache.Relax = true
+		bucketStorageCache.Update = func() (interface{}, error) {
+			ctx, done := context.WithTimeout(context.Background(), 2*time.Second)
```
vadmeste, line 1273: I think let's increase this to 10 seconds

cmd/bucket-quota.go
```diff
@@ -44,16 +42,18 @@ func NewBucketQuotaSys() *BucketQuotaSys {
...
 		// does not update the bucket usage values frequently.
-		sys.bucketStorageCache.TTL = 10 * time.Second
+		bucketStorageCache.TTL = 10 * time.Second
 		// Rely on older value if usage loading fails from disk.
-		sys.bucketStorageCache.Relax = true
-		sys.bucketStorageCache.Update = func() (interface{}, error) {
-			ctx, done := context.WithTimeout(context.Background(), 1*time.Second)
+		bucketStorageCache.Relax = true
+		bucketStorageCache.Update = func() (interface{}, error) {
+			ctx, done := context.WithTimeout(context.Background(), 2*time.Second)
```
vadmeste, line 56: two seconds seem to be low for me in a busy system

cmd/data-usage.go
```diff
@@ -73,28 +76,45 @@ func loadPrefixUsageFromBackend(ctx context.Context, objAPI ObjectLayer, bucket
...
+		prefixUsageCache.TTL = 30 * time.Second
+
+		// No need to fail upon Update() error, fallback to old value.
+		prefixUsageCache.Relax = true
+		prefixUsageCache.Update = func() (interface{}, error) {
+			m := make(map[string]uint64)
+			for _, pool := range z.serverPools {
+				for _, er := range pool.sets {
+					// Load bucket usage prefixes
+					ctx, done := context.WithTimeout(context.Background(), 2*time.Second)
```
vadmeste, line 89: This can be large if a bucket contains many prefixes at the bucket top level, I think 2 seconds is very low

vadmeste (commented):
Few comments

cmd/bucket-quota.go
harshavardhana: That is perfectly fine, there is nothing we can do about that. We cannot hold a PUT Object more than 2 seconds here during cache invalidation.

cmd/admin-handlers-users.go
harshavardhana: No we should not that is not a correct value.

cmd/admin-handlers-users.go
harshavardhana: 10 seconds is actually enough, 30secs is a bit longer.

cmd/data-usage.go
harshavardhana: That is perfectly fine, none of them will be beyond 1m prefixes @vadmeste and that is fairly small to read in 2secs. Mosty small object reads finish in < 1sec even on HDDs.

Also this call is hardly used.

cmd/admin-handlers-users.go
cesnietor: what makes it a correct vs not correct value? is it just gut feeling?

cesnietor (commented):
can we add the ttl values as constants so that we can have only one place to tweak them ?

harshavardhana:
> can we add the ttl values as constants so that we can have only one place to tweak them ?

We do not want them to be tweaked @cesnietor these are something that should never be tweaked. They should have sane defaults that work in most environments. If you want to test you should modify them as per your needs via source and build it.

The above values are a good trade-off between 

- how fast does this API respond? - have a top-level cap < 10secs

v/s 

- are we comfortable with older values? - yes we are as there is no real
  requirement to keep this "info" in realtime.

vadmeste (approved):
LGTM

poornas (approved):

PR #18069, opened 2023-09-20
remove abandoned children from scanner routine

harshavardhana:

## Description
remove abandoned children since it's not required

## Motivation and Context
- abandonedChildren logic to avoid listing in scanning cycles
- we visit each object anyways and heal it so dangling objects 
  get removed always
- abandonedChildren can be an expensive logic with the listing, 
  while the drives are busy serving other requests

## How to test this PR?
this logic is historical and not necessarily useful

## Types of changes

## Checklist:

### DIFF
--- a/cmd/data-scanner.go
+++ b/cmd/data-scanner.go
@@ -246,9 +246,6 @@ type folderScanner struct {
 	healObjectSelect      uint32 // Do a heal check on an object once every n cycles. Must divide into healFolderInclude
 	scanMode              madmin.HealScanMode
 
-	disks       []StorageAPI
-	disksQuorum int
-
 	// If set updates will be sent regularly to this channel.
 	// Will not be closed when returned.
 	updates    chan<- dataUsageEntry
@@ -299,7 +296,7 @@ type folderScanner struct {
 // The returned cache will always be valid, but may not be updated from the existing.
 // Before each operation sleepDuration is called which can be used to temporarily halt the scanner.
 // If the supplied context is canceled the function will return at the first chance.
-func scanDataFolder(ctx context.Context, disks []StorageAPI, basePath string, cache dataUsageCache, getSize getSizeFn, scanMode madmin.HealScanMode) (dataUsageCache, error) {
+func scanDataFolder(ctx context.Context, basePath string, cache dataUsageCache, getSize getSizeFn, scanMode madmin.HealScanMode) (dataUsageCache, error) {
 	switch cache.Info.Name {
 	case "", dataUsageRoot:
 		return cache, errors.New("internal error: root scan attempted")
@@ -318,8 +315,6 @@ func scanDataFolder(ctx context.Context, disks []StorageAPI, basePath string, ca
 		scanMode:              scanMode,
 		updates:               cache.Info.updates,
 		updateCurrentPath:     updatePath,
-		disks:                 disks,
-		disksQuorum:           len(disks) / 2,
 	}
 
 	// Enable healing in XL mode.
@@ -383,10 +378,6 @@ func (f *folderScanner) scanFolder(ctx context.Context, folder cachedFolder, int
 			return ctx.Err()
 		default:
 		}
-		var abandonedChildren dataUsageHashMap
-		if !into.Compacted {
-			abandonedChildren = f.oldCache.findChildrenCopy(thisHash)
-		}
 
 		// If there are lifecycle rules for the prefix.
 		_, prefix := path2BucketObjectWithBasePath(f.root, folder.name)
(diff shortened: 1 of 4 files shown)


klauspost:
Why would you remove all passive cleanup?

> remove abandoned children since it's not required

Now you have nothing that will clean up abandoned data that was left behind by an incomplete delete operation. You will have dead data lingering for months - and for disks missing data, you will see incorrect data usage when these are scanned.

If there is nothing to fix, it is practically a no-op.

harshavardhana:
> If there is nothing to fix, it is practically a no-op.

It is not a no-op @klauspost we are actually listing while scanning the drives this can create a fan-out-like operation which will in turn use more resources, that is not a good idea and there is no way to control it or cancel it.

> Now you have nothing that will clean up abandoned data that was left behind by an incomplete delete operation. You will have dead data lingering for months - and for disks missing data, you will see incorrect data usage when these are scanned.

I tested this already, every 16 cycles you hit every object anyway and that cleanup happens automatically on dangling paths. `f.getSize()` already calls the relevant portions every 512th cycle (planning to go back to the previous value now that a more expensive portion is removed)

In a flaky environment, our current implementation is causing a quorum listing on per per-erasure set basis, so cases where the different buckets have inconsistent prefixes lead to a large listing operation triggered as if an application is performing an I/O. 

> Why would you remove all passive cleanup?
> 

Simply cost-cutting, we are just being too aggressive in our approach here and it is not necessary TBH.

klauspost:
@harshavardhana 

> It is not a no-op @klauspost we are actually listing while scanning the drives this can create a fan-out-like operation which will in turn use more resources, that is not a good idea and there is no way to control it or cancel it.

Yes. If the folder exists on other and is missing on this drive, then we list for healing. Obviously the folder needs healing and if it isn't usage will be very much wrong. So it is doing work if it needs to.

> I tested this already, every 16 cycles you hit every object anyway and that cleanup happens automatically on dangling paths.

No, not unless you actually hit this code - there is no crossdisk consistency checks outside this code. Usage will fluctuate greatly for every run and missing prefixes will be just that for a long while.

> f.getSize() already calls the relevant portions every 512th cycle (planning to go back to the previous value now that a more expensive portion is removed)

1024 cycles can take months or even years. You will have drives missing parts for all that duration - and if you happen that the healcheck is on the disk with the missing data it will not trigger - and it will only every restore the one missing object. So healing a full missing prefix will likely take years.

Here we *know* an entire directory is missing. It may be 1 object or it may be 10000 objects. If it is deleted, no big deal, we move on. If not, we heal all objects in the prefix.

harshavardhana:
No @klauspost all of those were sound decisions but our main problem is we are using something that has the potential to make things worse just for some corner cases where drives are offline in the middle of an I/O. 

Yes, the usage will be wrong for a bit but it will correct itself if it is allowed to do so under various usage cycles - heal which simply runs its course with other disks in the erasure set in the next cycle.

I can't see the need to do a large prefix-based listing and wait on thousands of objects and versions. With larger dense setups, this is a major problem and this kind of fan-out operation is wrong. This not only affects usage updates but also

- ILM expiration, tiering
- Replication of FAILED/Existing objects

We simply cannot have this kind of approach inside the scanner IMO. Remember for short WRITE failures which are the most common

- we have MRF which does its job
- we have GET object based healing
- we have HEAD object based healing

So I don't know why we need this as well to do an expensive operation, most I/O that missed a WRITE would get healed upon a READ call once the drive is online. Persistent data is always READ() by applications either via a GET or a HEAD call. 

In fact, we can actually even persist some of these MRF failures per erasure set and replay them as needed as well instead of relying on the scanner to catch this up. We already do this for replication failures per node.

> 1024 cycles can take months or even years. You will have drives missing parts for all that duration - and if you happen that the healcheck is on the disk with the missing data it will not trigger - and it will only every restore the one missing object. So healing a full missing prefix will likely take years.

NOTE: I also changed the heal object probability to 256th cycle which is 2x earlier than 512. - there is no such thing as a scanner taking years - in-fact almost everyone uses scanner in its fast form - scanner in-fact even on the largest of all the namespaces only has to scan a "limited" portion of objects per erasure set. 

Even in that every 16th cycle we visit each object and every 256th cycle we invoke the HealObject() that heals missing parts.

These are the main basis of my arguments here. We should at all costs avoid doing Listing() of any kind that requires 

- "quorum" based listing
- list all drives in an erasure set

This not only blocks active I/O on the folder that is currently actively being read where WalkDir() ends up serializing concurrently occurring list calls on HDDs.

With all of this in mind, I very much feel like what we have right now is a very niche case and we have added a rather aggressive approach towards handling this scenario.

klauspost:
@harshavardhana You will have usage fluctuate for months. Casual healing has a nice chance of missing these for many, many cycles - there is no guarantee you will hit it within 256 cycles since you may hit the faulty disk.

When an entire path is missing, the only thing you can do is ask the other disks. If it doesn't exist on others it is just a stat operation. When it exists elsewhere, there is clearly work to be done and just ignoring it will lead to objects existing and not existing based on which disks are hit, and you will have many more questions on why usage goes up and down without changes.

Due to compaction there is only a limited number of paths "remembered" between cycles, so the amount of objects will pretty much always be significant. Removing it will give you many, many more headaches than having it - it is silently fixing a ton of issues.

harshavardhana:
Will change this to avoid listing instead lets do a single drive listing.

PR #18074, opened 2023-09-21
fix: check post policy like AWS S3

### LINKED ISSUE #17391: Pre-signed POST policy check not correct
When using a Pre-signed POST policy it is possible to set fields (like Content-Disposition) that are not mentioned in it. It's also possible to modify the Content-Type even if it's specified.

## Expected Behavior

S3 docs specify
> Each form field that you specify in a form (except x-amz-signature, file, policy, and field names that have an x-ignore- prefix) must appear in the list of conditions.

## Current Behavior

Minio allows and processes field names not mentioned in the policy

## Possible Solution

```ts
client.presignedPutObject(bucket, key, 3600) + '&foo=bar'
```
Results in an 403 error like it should so maybe look there how the check is implemented

```ts
// server.ts
import { Client } from 'minio'
const client = new Client(...)

export function getUploadUrl(name: string, type: string, size: number) {
  const policy = client.newPostPolicy()

  policy.setBucket('bucket')
  policy.setKey(name)
  policy.setContentType(type)
  policy.setContentLengthRange(size, size)
  policy.setExpires(new Date(Date.now() + 1000 * 60 * 10))

  return client.presignedPostPolicy(policy)
}
```
```ts
// client.ts
const file: File = ...
const uploadUrl = ... // getUploadUrl(file.name, file.type, file.size)
const form = new FormData()
const xhr = new XMLHttpRequest()

for (const [name, value] of Object.entries(uploadUrl.formData)) {
  form.append(name, value as string)
}

// is ignored according to spec
form.append('x-amz-ignore-foo, 'bar')

// all of these calls should result in an error but will be accepted

// sets Content-Disposition of the object
form.append('Content-Disposition', 'attachment; filename="file.exe"')

// appends the Content-Type to the one specified in the policy
// if file.type was `image/jpeg` it will be `image/jpeg,application/foo`
form.append('Content-Type', 'application/foo')

// is ignored
form.append('foo, 'bar')


form.append('file', file)
xhr.open('POST', uploadUrl.postURL)
xhr.send(form)
```

## Regression

This issue was previously reported in https://github.com/minio/minio/issues/8498 with a potential implementation in https://github.com/minio/minio/pull/8623 but was closed as stale

## Your Environment

* Version used (`minio --version`):  version RELEASE.2023-06-02T23-17-26Z (commit-id=0649aca2195a8d3aef9b946337b4ca6581e23844)
* Server setup and configuration: docker run quay.io/minio/minio server /data

jiuker:

## Description

fix: https://github.com/minio/minio/issues/17391#issuecomment-1727466126
```log
Check all formValues appear in postPolicyForm or return error.
Check Content-Type check is not correct, probably because it's an array.
```
## How to test this PR?

```go
package main

import (
	"bytes"
	"context"
	"fmt"
	"io"
	"io/ioutil"
	"log"
	"mime/multipart"
	"net/http"
	"os"
	"time"

	minio "github.com/minio/minio-go/v7"
)

func main() {
	endpoint := "127.0.0.1:9000"
	accessKeyID := "minioadmin"
	secretAccessKey := "minioadmin"
	useSSL := false

	minioClient, err := minio.New(endpoint, &minio.Options{
		Creds:  credentials.NewStaticV4(accessKeyID, secretAccessKey, ""),
		Secure: useSSL,
	})
	if err != nil {
		log.Fatalln(err)
	}
	ctx := context.Background()
	policy := minio.NewPostPolicy()
	policy.SetBucket("mytest")

	file, err := os.Open("req")
	if err != nil {
		panic(err)
	}
	fileInfo, err := file.Stat()
	if err != nil {
		panic(err)
	}
	policy.SetKey(file.Name())
	policy.SetContentType("image/png")
	policy.SetContentLengthRange(fileInfo.Size(), fileInfo.Size())
	policy.SetExpires(time.Now().Add(time.Hour * 6))
	policy.SetUserMetadata("myname", "hahaha")

	postUrl, formData, err := minioClient.PresignedPostPolicy(ctx, policy)
	if err != nil {
		panic(err)
	}

	bodyBuffer := &bytes.Buffer{}
	w := multipart.NewWriter(bodyBuffer)
	for k, v := range formData {
		w.WriteField(k, v)
	}

	// is ignored according to spec
	w.WriteField("x-amz-ignore-foo", "bar")

	// all of these calls should result in an error but will be accepted

	// sets Content-Disposition of the object
	w.WriteField("Content-Disposition", `attachment; filename="test.png"`)

	// appends the Content-Type to the one specified in the policy
	// if file.type was `image/jpeg` it will be `image/jpeg,application/foo`
	w.WriteField("Content-Type", "application/png")

	// is ignored
	w.WriteField("foo", "bar")

	fw, err := w.CreateFormFile("file", "test.png")
	if err != nil {
		panic(err)
	}
	_, err = io.Copy(fw, file)
	if err != nil {
		panic(err)
	}
	resp, err := http.Post(postUrl.String(), w.FormDataContentType(), bodyBuffer)
	if err != nil {
		panic(err)
	}
	defer resp.Body.Close()
	body, err := ioutil.ReadAll(resp.Body)
	if err != nil {
		panic(err)
	}
	fmt.Println(resp.Status)
	fmt.Println(resp.StatusCode)
	fmt.Println(string(body))
}
```
We should reject the post for
```go
	// sets Content-Disposition of the object
	w.WriteField("Content-Disposition", `attachment; filename="test.png"`)

	// appends the Content-Type to the one specified in the policy
	// if file.type was `image/jpeg` it will be `image/jpeg,application/foo`
	w.WriteField("Content-Type", "application/png")
```

## Types of changes

## Checklist:


cmd/postpolicyform.go
```diff
@@ -265,6 +265,22 @@ func checkPostPolicy(formValues http.Header, postPolicyForm PostPolicyForm) erro
 	if !postPolicyForm.Expiration.After(UTCNow()) {
 		return fmt.Errorf("Invalid according to Policy: Policy expired")
 	}
+	// check all formValues appear in postPolicyForm or return error. #https://github.com/minio/minio/issues/17391
+	checkHeader := map[string][]string{}
+	ignoreKeys := map[string]bool{}
```
shtripat, line 284: Can we just use a slice of string to maintain list of keys to ignore and check keys against this list?

jiuker: it can be set by `X-Amz-Ignore-Foo` @shtripat

shtripat (commented):
lgtm

cmd/postpolicyform.go
```diff
@@ -270,7 +270,7 @@ func checkPostPolicy(formValues http.Header, postPolicyForm PostPolicyForm) erro
 	ignoreKeys := map[string]bool{}
 	for key, value := range formValues {
 		switch {
-		case ignoreKeys[key], key == "Policy", key == "X-Amz-Signature":
+		case ignoreKeys[key], key == "Policy", key == "X-Amz-Signature", key == "Content-Encoding":
```
harshavardhana: Define constants for these in the handler and use them here..

cmd/postpolicyform.go
```diff
@@ -270,7 +270,7 @@ func checkPostPolicy(formValues http.Header, postPolicyForm PostPolicyForm) erro
...
 			continue
 		case strings.HasPrefix(key, "X-Amz-Ignore-"):
```
harshavardhana, line 289: use `HasPrefix` for case-insensitive check.

harshavardhana (changes requested):

cmd/postpolicyform.go
jiuker: the `formValues ` have case-insensitive keys already.
```go
formValues[http.CanonicalHeaderKey(name)] = append(formValues[http.CanonicalHeaderKey(name)], b.String())
```

cmd/postpolicyform.go
```diff
@@ -31,6 +31,7 @@ import (
 
 	"github.com/bcicen/jstream"
 	"github.com/minio/minio-go/v7/pkg/set"
+	http2 "github.com/minio/minio/internal/http"
```
harshavardhana:
```suggestion
	xhttp "github.com/minio/minio/internal/http"
```

not `http2` http2 means http2.0

cmd/postpolicyform.go
```diff
@@ -51,6 +52,18 @@ var startsWithConds = map[string]bool{
 	"$x-amz-date":              false,
 }
 
+var checkPostPolicyIgnoreKeys = map[string]bool{
+	"Policy":           true,
+	http2.AmzSignature: true,
+	"Content-Encoding": true,
```
harshavardhana:
```suggestion
	xhttp.ContentEncoding: true,
```

cmd/postpolicyform.go
```diff
@@ -51,6 +52,18 @@ var startsWithConds = map[string]bool{
 	"$x-amz-date":              false,
 }
 
+var checkPostPolicyIgnoreKeys = map[string]bool{
```
harshavardhana:
```suggestion
var postPolicyIgnoreKeys = map[string]bool{
```

cmd/postpolicyform.go
```diff
@@ -270,9 +283,9 @@ func checkPostPolicy(formValues http.Header, postPolicyForm PostPolicyForm) erro
 	ignoreKeys := map[string]bool{}
 	for key, value := range formValues {
 		switch {
-		case ignoreKeys[key], key == "Policy", key == "X-Amz-Signature", key == "Content-Encoding":
+		case ignoreKeys[key], checkPostPolicyIgnoreKeys[key]:
```
harshavardhana:
```suggestion
		case ignoreKeys[key], postPolicyIgnoreKeys[key]:
```

harshavardhana (changes requested):

harshavardhana:
Please check 

```
(9/14) Running minio-java tests ... FAILED in 1 minutes and 5 seconds
{
  "name": "minio-java",
  "function": "getPresignedPostFormData()",
  "duration": 29,
  "status": "FAIL",
  "error": "java.lang.Exception: failed to upload object. Response: Response{protocol=http/1.1, code=403, message=Forbidden, url=http://nginx:9000/minio-java-test-2mkg5ho}, Error: <?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<Error><Code>AccessDenied</Code><Message>Access Denied. (Each form field that you specify in a form (except Content-Type) must appear in the list of conditions.)</Message><BucketName>minio-java-test-2mkg5ho</BucketName><Resource>/minio-java-test-2mkg5ho</Resource><RequestId>1786E614343E54C0</RequestId><HostId>e0c385c033c4356721cc9121d3109c9b9bfdefb22fd2747078acd223287[99](https://github.com/minio/minio/actions/runs/6260262125/job/16997967875?pr=18074#step:7:100)e36</HostId></Error> >>> [FunctionalTest.getPresignedPostFormData(FunctionalTest.java:1762), FunctionalTest.runObjectTests(FunctionalTest.java:3762), FunctionalTest.runTests(FunctionalTest.java:3781), FunctionalTest.runEndpointTests(FunctionalTest.java:3879), FunctionalTest.main(FunctionalTest.java:4000)]"
}
```

harshavardhana (approved):

harshavardhana (approved):

cmd/postpolicyform.go
```diff
@@ -311,6 +345,15 @@ func checkPostPolicy(formValues http.Header, postPolicyForm PostPolicyForm) erro
...
 		}
+		delete(checkHeader, formCanonicalName)
+	}
+
+	if len(checkHeader) != 0 {
+		logKeys := make([]string, 0, len(checkHeader))
+		for key := range checkHeader {
+			logKeys = append(logKeys, key)
+		}
+		return fmt.Errorf("Each form field that you specify in a form (except %s) must appear in the list of conditions.", strings.Join(logKeys, ", "))
```
kumy, line 356: Are you sure? I recently hit that error message, and i had to add the keys listed in the "except" list to get it working again.

harshavardhana: Not all fields you need to provide what you upload with standard headers. All in the name of AWS S3 compatibility unfortunately.

kumy: Yes I agree with you but the wording gave me troubles. The word "except" made me think all the fields must be included excepted the one listed in the parenthesis. Maybe just remove that word?

horta: I'm still lost. I keep getting `Access Denied. (Each form field that you specify in a form (except Signature, Awsaccesskeyid) must appear in the list of conditions`.

jiuker: Yeah. You can add them to policy like
```go
        policy.SetKey(file.Name())
	policy.SetContentType("image/png")
``` 
.  Or add the `x-amz-ignore-Signature` or `x-amz-ignore-Awsaccesskeyid` to postfrom. What kind of code do you use? Go or Java or others. Maybe I can found something for you. @horta

cmd/postpolicyform.go
```diff
@@ -51,6 +53,18 @@ var startsWithConds = map[string]bool{
...
+var postPolicyIgnoreKeys = map[string]bool{
+	"Policy":              true,
+	xhttp.AmzSignature:    true,
+	xhttp.ContentEncoding: true,
+	http.CanonicalHeaderKey(xhttp.AmzChecksumAlgo):   true,
+	http.CanonicalHeaderKey(xhttp.AmzChecksumCRC32):  true,
+	http.CanonicalHeaderKey(xhttp.AmzChecksumCRC32C): true,
+	http.CanonicalHeaderKey(xhttp.AmzChecksumSHA1):   true,
+	http.CanonicalHeaderKey(xhttp.AmzChecksumSHA256): true,
+	http.CanonicalHeaderKey(xhttp.AmzChecksumMode):   true,
```
harshavardhana, line 65: this must not be ignored @jiuker @klauspost I think.

klauspost: Yeah. Seems like creating a bug rather than fixing the test.

marktheunissen: While looking at this issue again, I noticed another problem: the handling of X-Amz-Ignore- form keys. It's possible to have no policy about a given field key name, and then provide it, like so:

```
curl -v http://127.0.0.1:9000/test/ \
-F bucket=test \
-F policy=... \
-F x-amz-algorithm=AWS4-HMAC-SHA256 \
-F x-amz-credential=minioadmin/20240924/us-east-1/s3/aws4_request \
-F x-amz-date=20240924T041513Z \
-F x-amz-signature=dcc099ecc88a6790417d37185b917f971985d66b11c69672881d3456f4b5386a \
-F Content-Type=image/sdf \
-F key=buckbunny.jpg \
-F content-disposition='attachment; filename="file.exe"' \
-F x-amz-ignore-content-disposition=""  \
-F file=@buckbunny.jpg
```
For example the above, if the policy has no specific rule about `content-disposition`, then we can bypass the check by providing `-F content-disposition='attachment; filename="file.exe"'` AND `-F x-amz-ignore-content-disposition=""` which basically tricks the policy check into ignoring the fact that we've provided `content-disposition`. 

This only works if the policy itself did not contain any specific restriction for that key name, however S3 would not allow this. 

I think that `X-Amz-Ignore` is incorrectly implemented, and furthermore the S3 docs don't even mention X-Amz-Ignore, they only mention a prefix called `x-ignore`. See https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-HTTPPOSTConstructPolicy.html. 

Just noting all this so I don't forget when looking at it next time.

PR #18081, opened 2023-09-21
replication: drop mrf entries if target has extended downtime

poornas:
Optimization to avoid overloading replication MRF when replication target is offline for more than 10 minutes and MRF entry limit is reached.

## Motivation and Context
this is only applicable for the most common use case of single replication target.

## How to test this PR?
set up replication, take down a target for more than 10 mins - not easy to observe except in an active system

## Types of changes

## Checklist:


cmd/bucket-targets.go
```diff
@@ -91,6 +91,16 @@ func (sys *BucketTargetSys) isOffline(ep *url.URL) bool {
 	return false
 }
 
+// offlineFor returns amount of time remote target has been offline
+func (sys *BucketTargetSys) offlineFor(epHost string) time.Duration {
```
harshavardhana:
```suggestion
func (sys *BucketTargetSys) offlineDuration(epHost string) time.Duration {
```

harshavardhana: Or change the function to take 

```
if globalBucketTargetSys.offlineSince(ep, 10*time.Minute)
```

cmd/bucket-targets.go
```diff
@@ -91,6 +91,16 @@ func (sys *BucketTargetSys) isOffline(ep *url.URL) bool {
...
+	sys.hMutex.RLock()
+	defer sys.hMutex.RUnlock()
```
harshavardhana: return `epHost == ""` before the lock.

cmd/bucket-replication.go
```diff
@@ -3148,19 +3173,23 @@ func (p *ReplicationPool) persistMRF() {
...
+// queueMRFSave queues most recent failures to disk if number of retries not exceed limit and if target
+// is not offline for more than 10 minutes.
+func (p *ReplicationPool) queueMRFSave(entry MRFReplicateEntry, ep string) {
 	if !p.initialized() {
 		return
 	}
-	if entry.RetryCount > mrfRetryLimit { // let scanner catch up if retry count exceeded
+	// drop if retry count exceeded or target is offline for more than 10 minutes and mrf map is full. Note that we may
+	// not always know the target endpoint if failure is early on in the replication flow
+	if entry.RetryCount > mrfRetryLimit || (ep != "" && globalBucketTargetSys.offlineFor(ep) > 10*time.Minute && len(p.mrfEntries) >= mrfMaxEntries) { // let scanner catch up if retry count exceeded
```
harshavardhana:
```suggestion
	if entry.RetryCount > mrfRetryLimit || (len(p.mrfEntries) >= mrfMaxEntries && globalBucketTargetSys.offlineSince(ep, 10*time.Minute)) { // let scanner catch up if retry count exceeded
```

This is an optimized check with `&&` operator if the first entry says `false` Golang optimizes and avoids calling the second operand.

harshavardhana (changes requested):

cmd/bucket-replication.go
```diff
@@ -1706,6 +1730,7 @@ type ReplicationPool struct {
 	mrfSaveCh       chan MRFReplicateEntry
 	mrfStopCh       chan struct{}
 	mrfWorkerSize   int
+	mrfEntries      map[string]MRFReplicateEntry
```
harshavardhana, line 1735: so this is going to be a million entries worst case? @poornas

harshavardhana (changes requested):

cmd/bucket-replication.go
harshavardhana: Having this as map I am not sure is a good idea.

vadmeste:
@poornas it is not clear what do you mean by overloading the MRF ? what is the consequences of it ?

cmd/bucket-replication.go
poornas: there would be duplicates if it were not a map - surely 1M shouldn't be that big of a deal?

poornas:
> @poornas it is not clear what do you mean by overloading the MRF ? what is the consequences of it ?

 This PR basically short circuiting MRF queueing if we know the MRF limit is already reached thus avoiding needless attempts early on.

cmd/bucket-replication.go
harshavardhana: it is a big deal @poornas - map and 1m entry lookups are huge, here the values are fat as well since its a data struct value of MRFReplicateEntry

harshavardhana: I don't think this PR makes sense to add it will end up causing more memory issues AFAICS.

poornas: you are misunderstanding - this is not a new map, just the MRF map moved into ReplicationPool struct so we know how many entries are there in the map early on. If we remove this map, there will be no MRF queueing to speak of.

poornas:
This PR as not needed right now , closing this.

PR #18084, opened 2023-09-21
feat: introduce list_quorum="auto" to prefer quorum drives

harshavardhana:

## Description
feat: introduce list_quorum="auto" to prefer quorum drives

## Types of changes

## Checklist:

### DIFF
--- a/cmd/metacache-set.go
+++ b/cmd/metacache-set.go
@@ -602,13 +602,14 @@ func getListQuorum(quorum string, driveCount int) int {
 	return driveCount
 }
 
-func getQuorumDisks(disks []StorageAPI, infos []DiskInfo, readQuorum int) (newDisks []StorageAPI) {
+func getQuorumDiskInfos(disks []StorageAPI, infos []DiskInfo, readQuorum int) (newDisks []StorageAPI, newInfos []DiskInfo) {
 	calcCommonCounter := func() (commonMutations uint64) {
 		filter := func() (commonCount uint64) {
 			max := 0
 			signatureMap := map[uint64]int{}
 			for _, info := range infos {
-				signatureMap[info.Metrics.TotalDeletes+info.Metrics.TotalWrites]++
+				mutations := info.Metrics.TotalDeletes + info.Metrics.TotalWrites
+				signatureMap[mutations]++
 			}
 			for ops, count := range signatureMap {
 				if max < count && commonCount < ops {
@@ -627,11 +628,18 @@ func getQuorumDisks(disks []StorageAPI, infos []DiskInfo, readQuorum int) (newDi
 
 	commonMutations := calcCommonCounter()
 	for i, info := range infos {
-		if (info.Metrics.TotalWrites + info.Metrics.TotalDeletes) >= commonMutations {
+		mutations := info.Metrics.TotalDeletes + info.Metrics.TotalWrites
+		if mutations >= commonMutations {
 			newDisks = append(newDisks, disks[i])
+			newInfos = append(newInfos, infos[i])
 		}
 	}
 
+	return newDisks, newInfos
+}
+
+func getQuorumDisks(disks []StorageAPI, infos []DiskInfo, readQuorum int) (newDisks []StorageAPI) {
+	newDisks, _ = getQuorumDiskInfos(disks, infos, readQuorum)
 	return newDisks
 }
 
--- a/cmd/metacache-set.go
+++ b/cmd/metacache-set.go
@@ -602,31 +602,65 @@ func getListQuorum(quorum string, driveCount int) int {
 	return driveCount
 }
 
-func getQuorumDiskInfos(disks []StorageAPI, infos []DiskInfo, readQuorum int) (newDisks []StorageAPI, newInfos []DiskInfo) {
-	calcCommonCounter := func() (commonMutations uint64) {
-		filter := func() (commonCount uint64) {
-			max := 0
-			signatureMap := map[uint64]int{}
-			for _, info := range infos {
-				mutations := info.Metrics.TotalDeletes + info.Metrics.TotalWrites
-				signatureMap[mutations]++
-			}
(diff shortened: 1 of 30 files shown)


harshavardhana:
Submitted as a draft for some initial feedback and discussions.

klauspost:
Well, this doesn't really include any of the "hard" part. We should probably discuss that, since I don't really see any way to do that in a useful/scalable way.

harshavardhana:
> Well, this doesn't really include any of the "hard" part. We should probably discuss that, since I don't really see any way to do that in a useful/scalable way.

The next change I will make will be bucket centric and then subsequently try to go all the way to 1 level prefix but I have to test it further.

cmd/metacache-set.go
```diff
@@ -590,25 +590,73 @@ func (er *erasureObjects) streamMetadataParts(ctx context.Context, o listPathOpt
...
+
+	commonMutations := calcCommonCounter()
+	for i, info := range infos {
+		if (info.Metrics.TotalWrites + info.Metrics.TotalDeletes) >= commonMutations {
+			newDisks = append(newDisks, disks[i])
+		}
+	}
+
+	return newDisks
+}
```
Praveenrajmani: :clap: :ok_hand:

cmd/metacache-set.go
```diff
@@ -590,25 +590,73 @@ func (er *erasureObjects) streamMetadataParts(ctx context.Context, o listPathOpt
...
 	case "optimal":
 		return (driveCount + 1) / 2
+	case "auto":
+		return -1
 	}
 	// defaults to 'strict'
 	return driveCount
 }
 
+func getQuorumDisks(disks []StorageAPI, infos []DiskInfo, readQuorum int) (newDisks []StorageAPI) {
```
Praveenrajmani: Question: "auto" list quorum will work only if we see some common mutations in the disks? ie, if the disks go down and come up randomly, we won't be able to get some common mutations on (total deletes + total writes), and we end up choosing all the disks for the listing. Correct?

Praveenrajmani (approved):

cmd/metacache-set.go
harshavardhana: Idea is that eventually all the missed events will get healed. Very flaky systems we can't do much they will behave the same as what it is today.

Remember this is erasure set level what it means is that there are many other erasure sets that may be good so overall it shall benefit. If one or two sets don't qualify it isn't a big deal

harshavardhana: Also for now this feature is kept opt-in, default will be still strict.

klauspost:
@Praveenrajmani My instincts says the same. I don't really see this working - but Harsha convinced me enough to "at least we can give it a try".

My overall impression is that it relies too much on state that cannot be kept consistent and cannot be resynchronized once out-of-sync. If we were a DB with a strict journal "yes" - but we went the scalable direction.

My main issue is that once it desyncs (and it will, just do a rolling restart) it will bring down performance unexpectedly. So if systems are relying on this to keep afloat they will have a rude awakening once it de-syncs.

For a read-only cluster also "yes", but then you might as well just reduce the listing disks once and for all. But let's get some data instead of speculating.

@harshavardhana Is there a way for us to extract the write/delete numbers? 

It would be interesting to see these after running this for a month in various prod environments. Even being able to see how they change over time would be very valuable for evaluating the feasibility.

harshavardhana:
> @Praveenrajmani My instincts says the same. I don't really see this working - but Harsha convinced me enough to "at least we can give it a try".
> 
> My overall impression is that it relies too much on state that cannot be kept consistent and cannot be resynchronized once out-of-sync. If we were a DB with a strict journal "yes" - but we went the scalable direction.
> 
> My main issue is that once it desyncs (and it will, just do a rolling restart) it will bring down performance unexpectedly. So if systems are relying on this to keep afloat they will have a rude awakening once it de-syncs.
> 
> For a read-only cluster also "yes", but then you might as well just reduce the listing disks once and for all. But let's get some data instead of speculating.
> 
> @harshavardhana Is there a way for us to extract the write/delete numbers?
> 
> It would be interesting to see these after running this for a month in various prod environments. Even being able to see how they change over time would be very valuable for evaluating the feasibility.

Yes @klauspost that information is persisted to disk in xattrs so you can use standard tools like getfattr to get those values and you can even manually sync them to a quorum state via setfattr and then restart services.

I have these Go specific tools already, will put these into our containers so that they can be used in case they are ever needed. 

Almost all our customers use strict quorum already and they are chugging along. However for customers who have stable infrastructures and manage their clusters well. Have the scale of many disks and erasure sets this will work fine for the most part.

harshavardhana:
PR is updated with 

- xattr tool added for any debugging that we might have to do
- Handled drive replacement healing situations. Thanks to @krisis for reminding me to do that.

harshavardhana:
The fresh drive replacement scenario requires a more fundamental change, and it will not work correctly in this PR.

This PR can only handle disconnects, and missed events are healed; however, it can't correctly heal a fresh drive replacement scenario.

Also, the persistent nature of this can only be enabled with `export _MINIO_DRIVE_QUORUM=on` environment variable.

harshavardhana:
PTAL @vadmeste @krisis - I have changed this to make it an opt-in feature with `_MINIO_DRIVE_QUORUM=on`

However, disk replacement and healing have not yet been done due to the underlying challenges of re-using the `local disk`.

harshavardhana:
I will take a stab at it once more by tonight and see where I can land with it.

harshavardhana:
To correctly know the right values per disk we need this PR to go in first

https://github.com/minio/minio/pull/18645

harshavardhana:
To use this PR and test you need to use _MINIO_DRIVE_QUORUM=on (to enable this feature), it is a hidden variable until it emerges to be default after some field understanding.

while testing make sure to do that.

minio-trusted (approved):

PR #17994, opened 2023-09-08
Load STS accounts into IAM cache lazily

donatello:
## Community Contribution License
All community contributions in this pull request are licensed to the project maintainers
under the terms of the [Apache 2 license] (https://www.apache.org/licenses/LICENSE-2.0). 
By creating this pull request I represent that I have the right to license the 
contributions to the project maintainers under the Apache 2 license.

## Description
In situations with large number of STS credentials on disk, IAM load time is high. To mitigate this, STS accounts will now be loaded into memory only on demand - i.e. when the credential is used.

In each IAM cache (re)load we skip loading STS credentials and STS policy mappings into memory. Since STS accounts only expire and cannot be deleted, there is no risk of invalid credentials being reused, because credential validity is checked when it is used.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Optimization (provides speedup with no functional changes)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [ ] Fixes a regression (If yes, please add `commit-id` or `PR #` here)
- [ ] Unit tests added/updated
- [ ] Internal documentation updated
- [ ] Create a documentation update request [here](https://github.com/minio/docs/issues/new?label=doc-change,title=Doc+Updated+Needed+For+PR+github.com%2fminio%2fminio%2fpull%2fNNNNN)

PR #18088, opened 2023-09-22
Fix: Update policy mapping properly in notif'n

donatello:

## Description

This is fixing a regression from an earlier change where STS account loading was made lazy.

## Motivation and Context

Fix a bug in AssumeRoleWithCertificate where created credentials dont work with other nodes than the one in which it was created.

## How to test this PR?

1. Setup multi-node MinIO with a self signed certificate (I used https://github.com/FiloSottile/mkcert) and enable AssumeRoleWithCertificate - https://github.com/minio/minio/blob/master/docs/sts/tls.md
2. Create client cert (some openssl commands for this are here https://docs.aws.amazon.com/iot/latest/developerguide/create-device-cert.html - use the self signed CA used in the previous step).
3. Generate creds with AssumeRoleWithCert - example curl command here - https://github.com/minio/minio/blob/master/docs/sts/tls.md#example
4. Test creds against each node of minio - it will work only with the node on which it was created - and after apply this fixes, repeat the test and it will work on all nodes.

## Types of changes

## Checklist:
- [x] Fixes a regression (https://github.com/minio/minio/pull/17994)

### DIFF
--- a/cmd/iam-store.go
+++ b/cmd/iam-store.go
@@ -1583,9 +1583,15 @@ func (store *IAMStoreSys) UserNotificationHandler(ctx context.Context, accessKey
 		m = cache.iamUsersMap
 	}
 	err := store.loadUser(ctx, accessKey, userType, m)
+
 	if err == errNoSuchUser {
 		// User was deleted - we update the cache.
 		delete(m, accessKey)
+
+		// Since cache was updated, we update the timestamp.
+		defer func() {
+			cache.updatedAt = time.Now()
+		}()
 
 		// 1. Start with updating user-group memberships
 		if store.getUsersSysType() == MinIOUsersSysType {
@@ -1618,7 +1624,6 @@ func (store *IAMStoreSys) UserNotificationHandler(ctx context.Context, accessKey
 		// 3. Delete any mapped policy
 		delete(cache.iamUserPolicyMap, accessKey)
 
-		cache.updatedAt = time.Now()
 		return nil
 	}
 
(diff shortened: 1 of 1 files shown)


cmd/iam-store.go
```diff
@@ -1626,37 +1626,38 @@ func (store *IAMStoreSys) UserNotificationHandler(ctx context.Context, accessKey
...
-		u, ok := cache.iamSTSAccountsMap[accessKey]
-		if ok {
-			cred := u.Credentials
-			if cred.ParentUser != "" && cred.ParentUser != globalActiveCred.AccessKey {
-				if _, ok := cache.iamUserPolicyMap[cred.ParentUser]; !ok {
-					cache.iamUserPolicyMap[cred.ParentUser] = cache.iamSTSPolicyMap[accessKey]
-					cache.updatedAt = time.Now()
-				}
-			}
-		}
```
donatello, line 1659: I don't think we need this block any more - not sure what the original purpose was.

harshavardhana: This block for OIDC credentials

donatello: Yeah, I don't think we need it - for sts accounts we lookup iamSTSPolicyMap for the parent (which is "virtual").

cmd/iam-store.go
```diff
@@ -1583,6 +1583,10 @@ func (store *IAMStoreSys) UserNotificationHandler(ctx context.Context, accessKey
 		m = cache.iamUsersMap
 	}
 	err := store.loadUser(ctx, accessKey, userType, m)
+	defer func() {
+		cache.updatedAt = time.Now()
```
vadmeste: It makes sense for me to only update cache.updateAt when this call returns a nil error, /cc @poornas

vadmeste (commented):
One comment

cmd/iam-store.go
harshavardhana: yeah that makes sense @donatello ^^

donatello: That's not completely correct - we do in some cases update the cache and still return non-nil error. I am slightly updating this logic.

harshavardhana (approved):

vadmeste (approved):
LGTM

PR #18107, opened 2023-09-25
scanner: Scan buckets asynchronously

vadmeste:
## Community Contribution License
All community contributions in this pull request are licensed to the project maintainers
under the terms of the [Apache 2 license] (https://www.apache.org/licenses/LICENSE-2.0). 
By creating this pull request I represent that I have the right to license the 
contributions to the project maintainers under the Apache 2 license.

## Description
- Scan buckets asynchrounously; multiple erasure sets scan in parallel, multiple disks in the same erasure set scan different buckets in parallel
- No data format is changed
- Cycle concept changed to be per bucket focused, the cycle is incremented ecah time a bucket is successfully or unsuccesfully scanned
- Next bucket to scan is chosen based on the lowest possible cycle number

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Optimization (provides speedup with no functional changes)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [ ] Fixes a regression (If yes, please add `commit-id` or `PR #` here)
- [ ] Unit tests added/updated
- [ ] Internal documentation updated
- [ ] Create a documentation update request [here](https://github.com/minio/docs/issues/new?label=doc-change,title=Doc+Updated+Needed+For+PR+github.com%2fminio%2fminio%2fpull%2fNNNNN)

PR #18109, opened 2023-09-26
scanner: Allow full throttle if there is no parallel disk ops

vadmeste:

## Description
When there is no ongoing xl storage operations, it is okay to allow the scanner to go
on full speed scan.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/data-scanner.go
+++ b/cmd/data-scanner.go
@@ -246,7 +246,7 @@ type folderScanner struct {
 	healObjectSelect      uint32 // Do a heal check on an object once every n cycles. Must divide into healFolderInclude
 	scanMode              madmin.HealScanMode
 
-	fullThrottle func() bool
+	weSleep func() bool
 
 	disks       []StorageAPI
 	disksQuorum int
@@ -301,7 +301,7 @@ type folderScanner struct {
 // The returned cache will always be valid, but may not be updated from the existing.
 // Before each operation sleepDuration is called which can be used to temporarily halt the scanner.
 // If the supplied context is canceled the function will return at the first chance.
-func scanDataFolder(ctx context.Context, disks []StorageAPI, basePath string, cache dataUsageCache, getSize getSizeFn, scanMode madmin.HealScanMode, fullThrottle func() bool) (dataUsageCache, error) {
+func scanDataFolder(ctx context.Context, disks []StorageAPI, basePath string, cache dataUsageCache, getSize getSizeFn, scanMode madmin.HealScanMode, weSleep func() bool) (dataUsageCache, error) {
 	switch cache.Info.Name {
 	case "", dataUsageRoot:
 		return cache, errors.New("internal error: root scan attempted")
@@ -318,7 +318,7 @@ func scanDataFolder(ctx context.Context, disks []StorageAPI, basePath string, ca
 		dataUsageScannerDebug: false,
 		healObjectSelect:      0,
 		scanMode:              scanMode,
-		fullThrottle:          fullThrottle,
+		weSleep:               weSleep,
 		updates:               cache.Info.updates,
 		updateCurrentPath:     updatePath,
 		disks:                 disks,
@@ -407,7 +407,7 @@ func (f *folderScanner) scanFolder(ctx context.Context, folder cachedFolder, int
 			replicationCfg = f.oldCache.Info.replication
 		}
 
-		if !f.fullThrottle() {
+		if f.weSleep() {
 			scannerSleeper.Sleep(ctx, dataScannerSleepPerFolder)
 		}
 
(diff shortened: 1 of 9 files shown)


cmd/xl-storage-disk-id-check.go
```diff
@@ -665,10 +666,21 @@ func scannerTrace(s scannerMetric, startTime time.Time, duration time.Duration,
 }
 
 // Update storage metrics
-func (p *xlStorageDiskIDCheck) updateStorageMetrics(s storageMetric, paths ...string) func(err *error) {
+func (p *xlStorageDiskIDCheck) updateStorageMetrics(s storageMetric, ctx context.Context, paths ...string) func(err *error) {
 	startTime := time.Now()
 	trace := globalTrace.NumSubscribers(madmin.TraceStorage) > 0
+	incS3Count := false
+
+	if tc, ok := ctx.Value(mcontext.ContextTraceKey).(*mcontext.TraceCtxt); ok && strings.HasPrefix(tc.FuncName, "s3.") {
```
vadmeste: >  strings.HasPrefix(tc.FuncName, "s3.")

This doesn't look robust enough for now; maybe we need to set a new context value that can be spread between nodes using another header

cmd/xl-storage-disk-id-check.go
```diff
@@ -665,10 +666,21 @@ func scannerTrace(s scannerMetric, startTime time.Time, duration time.Duration,
 }
 
 // Update storage metrics
-func (p *xlStorageDiskIDCheck) updateStorageMetrics(s storageMetric, paths ...string) func(err *error) {
+func (p *xlStorageDiskIDCheck) updateStorageMetrics(s storageMetric, ctx context.Context, paths ...string) func(err *error) {
```
harshavardhana: context must be first argument @vadmeste

harshavardhana (changes requested):

klauspost:
The `diskHealthTracker`  in `xlStorageDiskIDCheck` already keeps track of running operations - so when `cap(diskHealthTracker.tokens) == ` len(diskHealthTracker.tokens)` there are no ongoing disk activity.  The scanner is already excluded from this.

QUESTION is if we really want to do that by default?

cmd/xl-storage-disk-id-check.go
klauspost: Not needed. The health tracker already tracks ongoing operations and there is no magic there.

vadmeste:
> QUESTION is if we really want to do that by default?

The main goal is to make scanning goes in full throttle when we know that there is no ongoing S3 operations; Since one S3 operation involves multiple disks in different nodes, I thought the easiest way is to keep track of the number of concurrent S3 operations per disk using context.Context concept; (this seems the cheapest option in my opinion)

So, if an S3 request lands one node; we add an attribute to context.Context which will be passed down to xl-storage, that way we can count the number of concurrent s3 ops per disk. To make sure that remote disks get the same attribute, we can pass that in internode calls headers; 

Now each disk that is currently scanning knows how many concurrent requests that is hitting and dynamically decides if it should go in full throtlle or uses the usual sleeping mechanism

klauspost:
> The main goal is to make scanning goes in full throttle when we know that there is no ongoing S3 operations

Yes. But I am not too sure everybody will appreciate the extra wear on the hardware.

> (monitor) S3 operations

It is a hacky, fragile solution. When we already monitor running disk operations we should hook into that instead.

You can add a parameter in `(p *xlStorageDiskIDCheck) NSScanner(..)` that passes down a function on whether wait should be skipped.

vadmeste:
> Yes. But I am not too sure everybody will appreciate the extra wear on the hardware.

Yes, maybe let's discuss this first; what we can possibly do is to avoid running the scanner when we know there is no write s3 operations since the last scanner run; I actually started to believe we need to merge the other scanner PR first, https://github.com/minio/minio/pull/18107 ; I have few left things to check before making it ready for review

cmd/data-scanner.go
```diff
@@ -246,6 +246,8 @@ type folderScanner struct {
 	healObjectSelect      uint32 // Do a heal check on an object once every n cycles. Must divide into healFolderInclude
 	scanMode              madmin.HealScanMode
 
+	fullThrottle func() bool
```
harshavardhana:
```suggestion
	weSleep func() bool
```

is better @vadmeste which returns `true` when `cap(tokens) != len(tokens)`

harshavardhana (changes requested):

cmd/data-scanner.go
harshavardhana: PTAL @vadmeste

harshavardhana (approved):

harshavardhana:
I have rebased and pushed the changes @vadmeste

shtripat (approved):
lgtm

harshavardhana:
This PR is merged, @vadmeste will send another PR to allow this to be configurable behavior via `scanner speed=default delay=always`

PR #18131, opened 2023-09-28
Listbuckets: Check quorum for each bucket

klauspost:
## Description

Each host does it's own quorum calculation, across disks. This means that if a host can reach quorum, it alone decides if a bucket exists or not.

We switch to a similar quorum calculation like `GetBucketInfo`, where there has to be quorum for a bucket checked across hosts.

## Motivation and Context

If a node misses a bucket delete request, it will keep a "zombie" bucket alive, which cannot otherwise reach quorum.

## How to test this PR?

Disconnect a host. Delete a bucket. Reconnect the host. List buckets.

## Types of changes

### DIFF
--- a/cmd/peer-s3-client.go
+++ b/cmd/peer-s3-client.go
@@ -153,17 +153,18 @@ func (sys *S3PeerSys) ListBuckets(ctx context.Context, opts BucketOptions) (resu
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


harshavardhana:
@klauspost @vadmeste sent a PR already for this.

klauspost:
Replaced by https://github.com/minio/minio/pull/18125

PR #18136, opened 2023-09-29
Fix tier addition validation

poornas:

## Description
preliminary probe was not cleaning up the versioned content created on remote tier.

## Motivation and Context
Fix tier in use error

## How to test this PR?
```
 mc admin tier add minio sitea tier1 --endpoint http://localhost:9004 --access-key minio --secret-key minio123 --bucket bucket --prefix prefix6/
mc: <ERROR> Unable to configure remote tier target. Specified remote tier is already in use.
```

## Types of changes

## Checklist:

### DIFF
--- a/cmd/warm-backend.go
+++ b/cmd/warm-backend.go
@@ -49,7 +49,7 @@ const probeObject = "probeobject"
 // to perform all operations defined in the WarmBackend interface.
 func checkWarmBackend(ctx context.Context, w WarmBackend) error {
 	var empty bytes.Reader
-	_, err := w.Put(ctx, probeObject, &empty, 0)
+	remoteVersionID, err := w.Put(ctx, probeObject, &empty, 0)
 	if err != nil {
 		if _, ok := err.(BackendDown); ok {
 			return err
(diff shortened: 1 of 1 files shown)


harshavardhana:
@poornas I think the idea was we do not expect DeleteVersion permission on the backend bucket. This will bring it back. Can we do cleanup differently? Or not do at all?

vadmeste:
@poornas can you show how to reproduce from scratch ? I **think** the point of the error is that a path in the warm tier is already in use probably by another tiering configuration, so we abort quickly. I do not think we should clean up in the first place.

poornas:
> @poornas can you show how to reproduce from scratch ? I **think** the point of the error is that a path in the warm tier is already in use probably by another tiering configuration, so we abort quickly. I do not think we should clean up in the first place.

clean up was only for the probe object created to test permissions - just try tier addition on a versioned bucket. It will fail with tier in use error

vadmeste:
> clean up was only for the probe object created to test permissions - just try tier addition on a versioned bucket. It will fail with tier in use error

It doesn't fail for me. Can you try with a new versioned bucket without this PR ?

kannappanr (approved):
LGTM

harshavardhana (approved):

vadmeste (approved):
I talked with @kannappanr and I understand more now;

In all cases, we need s3:DeleteObjectVersion in case the user enabled versioning in the warm tier and that may cause a lot of soft deletes not being correctly cleaned up

PR #18151, opened 2023-10-01
Do not require restart when a disk is unreachable during node boot

vadmeste:

## Description
A disk that is not able to initialize when an instance is started 
will never have a handler registered, which means a user will 
need to restart the node after fixing the disk;

This will also prevent showing the wrong 'upgrade is needed' 
error message in that case.

When the disk is still failing; print an error each 30 minutes;
Disk reconnection will be retried each 30 seconds.

## Motivation and Context
Do not require node restart when the disk fails to initialize
during start up

## Types of changes

## Checklist:

### DIFF
--- a/cmd/storage-rest-server.go
+++ b/cmd/storage-rest-server.go
@@ -54,7 +54,18 @@ var errDiskStale = errors.New("drive stale")
 
 // To abstract a disk over network.
 type storageRESTServer struct {
-	storage *xlStorageDiskIDCheck
+	storage atomic.Value
+}
+
+func (s *storageRESTServer) getStorage() *xlStorageDiskIDCheck {
+	if s, ok := s.storage.Load().(*xlStorageDiskIDCheck); ok {
+		return s
+	}
+	return nil
+}
+
+func (s *storageRESTServer) setStorage(xl *xlStorageDiskIDCheck) {
+	s.storage.Store(xl)
 }
 
 func (s *storageRESTServer) writeErrorResponse(w http.ResponseWriter, err error) {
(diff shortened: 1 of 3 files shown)


cmd/storage-rest-server.go
```diff
@@ -52,7 +53,18 @@ var errDiskStale = errors.New("drive stale")
...
-	storage *xlStorageDiskIDCheck
+	storage atomic.Value
+}
+
+func (s *storageRESTServer) getStorage() *xlStorageDiskIDCheck {
+	if s, ok := s.storage.Load().(*xlStorageDiskIDCheck); ok {
+		return s
+	}
+	return nil
+}
```
harshavardhana, line 65: so this has the potential to return `nil` so a crash is okay here? @vadmeste

harshavardhana (changes requested):

cmd/storage-rest-server.go
harshavardhana: This still applies @vadmeste

vadmeste: @harshavardhana the caller function will be responsible for not crashing

harshavardhana: so `s.getStorage().WalkDir()` will crash?

vadmeste: > so s.getStorage().WalkDir() will crash?

what do you mean ?

klauspost: That these will crash if it returns nil: `infos, err := s.getStorage().ListVols(r.Context())`

vadmeste: yes, it will crash

PR #18160, opened 2023-10-03
Fix current_send_in_progress metric always being zero

### LINKED ISSUE #18159: current_send_in_progress metric is always 0
## Expected Behavior
The metric `current_send_in_progress` should provide "Number of concurrent async Send calls active to all targets" and we expect this to be non-zero when 100s of events are being sent concurrently (with no queue).

## Current Behavior
Currently the `current_send_in_progress` metric is always 0.
Adding a log to `targetlist.go` shows that `TargetList.currentSendCalls` is being incremented correctly.
```
2023/10/03 05:18:52 currentSendCalls: 18
2023/10/03 05:18:52 currentSendCalls: 19
2023/10/03 05:18:52 currentSendCalls: 18
2023/10/03 05:18:53 currentSendCalls: 19
2023/10/03 05:18:53 currentSendCalls: 20
2023/10/03 05:18:53 currentSendCalls: 18
2023/10/03 05:18:53 currentSendCalls: 18
2023/10/03 05:18:53 currentSendCalls: 18
2023/10/03 05:18:53 currentSendCalls: 19
2023/10/03 05:18:53 currentSendCalls 20
2023/10/03 05:18:53 currentSendCalls: 19
2023/10/03 05:18:53 currentSendCalls: 20
2023/10/03 05:18:53 currentSendCalls: 21
2023/10/03 05:18:53 currentSendCalls: 22
2023/10/03 05:18:53 currentSendCalls: 23
2023/10/03 05:18:53 currentSendCalls: 24
2023/10/03 05:18:53 currentSendCalls: 25
```
It just seems to be a problem with the cluster metric provided.
```
minio_notify_current_send_in_progress{server="127.0.0.1:9000"} 0
```

## Possible Solution
In `metrics-v2.go` we currently get this metric with `globalNotifyTargetList.Stats()`. However, `globalNotifyTargetList` is a difference instance of `TargetList` from the one that is actually being used to publish events.

Changing `globalNotifyTargetList.Stats()` to `globalEventNotifier.targetList.Stats()` uses the correct `TargetList` instance and seems to fix the issue.

```
minio_notify_current_send_in_progress{server="127.0.0.1:9000"} 16
```

1. Create a Kafka notification target with no queue directory. This can probably be any kind of target but I have only tested with a Kafka target. Not using a queue directory as this increases the lifetime of the goroutines which leads to a higher `currentSendCalls` values. When using a queue this value tends to never go above 1 because the goroutines are so short lived.
2. Add event subscription to a bucket with the above target. Subscribe to all events for all paths.
3. Create/Access some objects in the above bucket in short succession. I created 1000 objects over 30s.
4. Observe that the `current_send_in_progress` stays at 0 and does not increase as expect while there are multiple send go routines running. This can be done with a get request polling `/minio/v2/metrics/cluster` and looking for `minio_notify_current_send_in_progress`.

## Context

We want this metric so we can be sure we are not exceeding `maxConcurrentTargetSendCalls` and dropping events. Without it we cannot tell if our current load is coming close to exceeding the limit.

## Regression

Unsure

## Your Environment

* Version used (`minio --version`): `2023-01-12T02:06:16Z` and `DEVELOPMENT.2023-09-13T19-43-46Z` locally.

mctoohey:

## Description

Fixes https://github.com/minio/minio/issues/18159

## Motivation and Context

See https://github.com/minio/minio/issues/18159

## Types of changes

## Checklist:

### DIFF
--- a/cmd/metrics-v2.go
+++ b/cmd/metrics-v2.go
@@ -2556,7 +2556,7 @@ func getNotificationMetrics() *MetricsGroup {
 		cacheInterval: 10 * time.Second,
 	}
 	mg.RegisterRead(func(ctx context.Context) []Metric {
-		nstats := globalNotifyTargetList.Stats()
+		nstats := globalEventNotifier.targetList.Stats()
 		metrics := make([]Metric, 0, 1+len(nstats.TargetStats))
 		metrics = append(metrics, Metric{
 			Description: MetricDescription{
--- a/cmd/metrics-v2.go
+++ b/cmd/metrics-v2.go
@@ -2556,30 +2556,33 @@ func getNotificationMetrics() *MetricsGroup {
 		cacheInterval: 10 * time.Second,
 	}
 	mg.RegisterRead(func(ctx context.Context) []Metric {
-		nstats := globalEventNotifier.targetList.Stats()
-		metrics := make([]Metric, 0, 1+len(nstats.TargetStats))
-		metrics = append(metrics, Metric{
-			Description: MetricDescription{
-				Namespace: minioNamespace,
-				Subsystem: notifySubsystem,
-				Name:      "current_send_in_progress",
-				Help:      "Number of concurrent async Send calls active to all targets",
-				Type:      gaugeMetric,
-			},
-			Value: float64(nstats.CurrentSendCalls),
-		})
-		for _, st := range nstats.TargetStats {
+		metrics := make([]Metric, 0, 3)
+
+		if globalEventNotifier != nil {
+			nstats := globalEventNotifier.targetList.Stats()
 			metrics = append(metrics, Metric{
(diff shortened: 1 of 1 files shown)


harshavardhana (approved):

harshavardhana (changes requested):
We should add a check such that event notifier if nil we must not crash the handler. Also fix the actual code internally perhaps to return stats() correctly.

mctoohey:
@harshavardhana  I've add a check to only include these metrics if `globalEventNotifier` is not `nil`.


> Also fix the actual code internally perhaps to return stats() correctly.

I'm not quite sure what you would want to see changed for this.

harshavardhana (approved):

PR #14847, opened 2022-05-02
fix: remove embedded-policy as requested by the user

harshavardhana:
## Description
fix: remove embedded-policy as requested by the user

## Motivation and Context
this PR introduces few changes such as

- sessionPolicyName is not reused in extracted manner
  to apply policies for incoming authenticated calls,
  instead uses a different key to designate this
  information for the callers.

- this differentiation is needed to ensure that service
  account updates do not accidentally store JSON representation
  instead of base64 equivalent on the disk.

- relax requirements for Deleting a service account, allow
  deleting a service account that might be unreadable, i.e
  a situation where user might have removed sessionPolicy which
  now carries a JSON representation, making it unparsable.

- introduce some constants to reuse instead of strings.

fixes #14784 
## How to test this PR?
As per #14784 

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Optimization (provides speedup with no functional changes)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [ ] Fixes a regression (If yes, please add `commit-id` or `PR #` here)
- [ ] Documentation updated
- [ ] Unit tests added/updated

PR #18163, opened 2023-10-04
Fix DeleteServiceAccount API behavior

taran-p:

## Description
Changes DeleteServiceAccount behavior so that it returns an error if the service account does not exist.

## Motivation and Context
Currently DeleteServiceAccount will return successfully if the service account does not exist, but will return that the account is not found if it belongs to another user. This pr makes the behavior consistent between the two by saying the account is not found either way.

## How to test this PR?
Delete service accounts that do not exist

## Types of changes

## Checklist:


harshavardhana (changes requested):
All DELETEs must be idempotent, this is not necessary. Think of it like DeleteObject() S3 API behavior. 

HTTP API calls need to be generally idempotent for DELETEs its not useful to do this.

taran-p:
> All DELETEs must be idempotent, this is not necessary. Think of it like DeleteObject() S3 API behavior.
> 
> HTTP API calls need to be generally idempotent for DELETEs its not useful to do this.

It's not stricly necessary, but I think it would make sense since it maintains idempotence but makes the behavior consistent with that of deleteing a user or a group.

This is how the call functioned previous to #14847, which removed the error check due to a bug. This PR takes that into consideration by allowing the delete to continue in any scenario other than a nonexistent access key.

harshavardhana:
> All DELETEs must be idempotent, this is not necessary. Think of it like DeleteObject() S3 API behavior.
> 
> HTTP API calls need to be generally idempotent for DELETEs its not useful to do this.

@donatello WDYT?

donatello:
Looks like aws does return an error in this case.  https://docs.aws.amazon.com/IAM/latest/APIReference/API_DeleteAccessKey.html 


However our API is not s3 compatible anyway so it's possible to argue either way. 

I think it's better to take this in just to be explicit for user creds.

donatello (approved):

harshavardhana (approved):

PR #18164, opened 2023-10-04
Send kafka notification messages in batches when queue_dir is enabled

### LINKED ISSUE #18124: kafka_notify should support batching of messages to Kafka

We are trying to use Kafka bucket notifications to receive notifications for put/delete events on a bucket. We are using a queue (by setting queue_dir) because we want to guarantee delivery of these events.

However, we notice the performance of publishing events from the queue onto Kafka is very poor. Resulting in large delays (up to several hours) from when a bucket event occurs in MinIO to it getting published to Kafka.

It seems like each event is pulled off the queue one at a time and published to Kafka individually. Say it takes 5ms to send a message to Kafka and get a response (including network latency and processing time). This would limit us to 200 messages a second. We need to be able to support 1000s per second for our use case. Kafka has no trouble supporting 10s of thousands of messages per second. I think the throughput we want could be achieved if instead of sending these messages one at a time off the queue, they were sent in batches. 

Kafka supports batching messages and so does sumara (Kafka library used by MinIO) with the SendMessages function (https://github.com/IBM/sarama/pull/677). It would be great if kafka_notify could take advantage of this to send messages from the queues in batches.

We have consider and tried creating multiple kafka_notify targets (split by different events and prefixes) to help increase the throughput of messages to Kafka. While this does help. It is impractical at scale.

**Additional context**

I did some testing with a local build of MinIO from my laptop publishing to kafka with 300ms network latency. This only allowed me to send around 3 events per second to Kafka. I then made some changes to the MinIO code to send messages to Kafka in batches and I was able to send 100s of events per second from my laptop onto Kafka.

Praveenrajmani:

## Description

This PR adds the support to replay the events from `queue_dir` in batches. With this change, if queue dir is configured, the events will be replayed in batches to that kafka target. Kafka library has [SendMessages()](https://github.com/IBM/sarama/blob/v1.38.1/sync_producer.go#L26) which can be used for sending msgs in batch.

## Motivation and Context

Fixes #18124

## How to test this PR?

- Setup Kafka - Refer https://kafka.apache.org/quickstart
- Enable `notify_kafka` with `batch_size`
```
mc admin config set myminio notify_kafka enable=on topic=quickstart-events brokers=localhost:9092 queue_dir=/tmp/events batch_size=10
```
- Add bucket notification using `mc event add`
- Bring down the target and let the events persist in the queue store.
- Bring up the kafka target after sometime, and notice the events getting replayed in batches
- After bringing the target up, there shouldn't be any messages left in the queue store - `ls -l /tmp/events`

## Types of changes

## Checklist:

### DIFF
--- a/internal/event/target/kafka.go
+++ b/internal/event/target/kafka.go
@@ -57,6 +57,7 @@ const (
 	KafkaClientTLSCert = "client_tls_cert"
 	KafkaClientTLSKey  = "client_tls_key"
 	KafkaVersion       = "version"
+	KafkaBatchSize     = "batch_size"
 
 	EnvKafkaEnable        = "MINIO_NOTIFY_KAFKA_ENABLE"
 	EnvKafkaBrokers       = "MINIO_NOTIFY_KAFKA_BROKERS"
@@ -73,6 +74,9 @@ const (
 	EnvKafkaClientTLSCert = "MINIO_NOTIFY_KAFKA_CLIENT_TLS_CERT"
 	EnvKafkaClientTLSKey  = "MINIO_NOTIFY_KAFKA_CLIENT_TLS_KEY"
 	EnvKafkaVersion       = "MINIO_NOTIFY_KAFKA_VERSION"
+	EnvKafkaBatchSize     = "MINIO_NOTIFY_KAFKA_BATCH_SIZE"
+
+	maxBatchLimit = 100
 )
 
 // KafkaArgs - Kafka target arguments.
@@ -83,6 +87,7 @@ type KafkaArgs struct {
 	QueueDir   string      `json:"queueDir"`
 	QueueLimit uint64      `json:"queueLimit"`
 	Version    string      `json:"version"`
+	BatchSize  uint32      `json:"batchSize"`
 	TLS        struct {
 		Enable        bool               `json:"enable"`
 		RootCAs       *x509.CertPool     `json:"-"`
@@ -122,7 +127,20 @@ func (k KafkaArgs) Validate() error {
 			return err
 		}
 	}
+	if k.BatchSize > 1 {
+		if k.QueueDir == "" {
+			return errors.New("batch should be enabled only if queue dir is enabled")
+		}
+		if k.BatchSize > maxBatchLimit {
+			return fmt.Errorf("batch limit should not exceed %d", maxBatchLimit)
+		}
+	}
 	return nil
+}
+
+type batchItem struct {
+	key     string
+	message *sarama.ProducerMessage
 }
 
 // KafkaTarget - Kafka target.
@@ -134,6 +152,7 @@ type KafkaTarget struct {
 	producer   sarama.SyncProducer
 	config     *sarama.Config
 	store      store.Store[event.Event]
+	batch      *store.Batch[batchItem]
 	loggerOnce logger.LogOnce
 	quitCh     chan struct{}
 }
@@ -188,30 +207,16 @@ func (target *KafkaTarget) send(eventData event.Event) error {
 	if target.producer == nil {
 		return store.ErrNotConnected
 	}
-	objectName, err := url.QueryUnescape(eventData.S3.Object.Key)
+	msg, err := target.toProducerMessage(eventData)
 	if err != nil {
 		return err
 	}
-	key := eventData.S3.Bucket.Name + "/" + objectName
-
-	data, err := json.Marshal(event.Log{EventName: eventData.EventName, Key: key, Records: []event.Event{eventData}})
-	if err != nil {
-		return err
-	}
-
-	msg := sarama.ProducerMessage{
-		Topic: target.args.Topic,
-		Key:   sarama.StringEncoder(key),
-		Value: sarama.ByteEncoder(data),
-	}
-
-	_, _, err = target.producer.SendMessage(&msg)
-
+	_, _, err = target.producer.SendMessage(msg)
 	return err
 }
 
 // SendFromStore - reads an event from store and sends it to Kafka.
-func (target *KafkaTarget) SendFromStore(eventKey string) error {
+func (target *KafkaTarget) SendFromStore(key store.Key) error {
 	if err := target.init(); err != nil {
 		return err
 	}
@@ -236,7 +241,7 @@ func (target *KafkaTarget) SendFromStore(eventKey string) error {
 		}
 	}
 
-	eventData, eErr := target.store.Get(eventKey)
+	eventData, eErr := target.store.Get(key.Name)
 	if eErr != nil {
 		// The last event key in a successful batch will be sent in the channel atmost once by the replayEvents()
 		// Such events will not exist and wouldve been already been sent successfully.
@@ -246,17 +251,80 @@ func (target *KafkaTarget) SendFromStore(eventKey string) error {
 		return eErr
 	}
 
+	// If batch is enabled, the event will be batched and
+	// will be committed once the batch is full.
+	if target.batch != nil {
+		msg, err := target.toProducerMessage(eventData)
+		if err != nil {
+			return err
+		}
+		err = target.batch.Add(batchItem{key.Name, msg})
+		if err != nil {
+			if errors.Is(err, store.ErrBatchFull) {
+				// This should not happen, just commit
+				// the batch and return the error.
+				if targetErr := target.commitBatch(); targetErr != nil {
+					err = targetErr
+				}
+			}
+			return err
+		}
+		// commit the batch if the key is the last one present in the store.
+		if key.IsLast || target.batch.IsFull() {
+			return target.commitBatch()
+		}
+		return nil
+	}
+
 	err = target.send(eventData)
 	if err != nil {
-		// Sarama opens the ciruit breaker after 3 consecutive connection failures.
-		if err == sarama.ErrLeaderNotAvailable || err.Error() == "circuit breaker is open" {
+		if isKafkaConnErr(err) {
 			return store.ErrNotConnected
 		}
 		return err
 	}
 
 	// Delete the event from store.
-	return target.store.Del(eventKey)
+	return target.store.Del(key.Name)
+}
+
+func (target *KafkaTarget) commitBatch() error {
+	var keysToDelete []string
+	var msgs []*sarama.ProducerMessage
+
+	for _, item := range target.batch.Get() {
+		keysToDelete = append(keysToDelete, item.key)
+		msgs = append(msgs, item.message)
+	}
+
+	err := target.producer.SendMessages(msgs)
+	if err != nil {
+		if isKafkaConnErr(err) {
+			return store.ErrNotConnected
+		}
+		return err
+	}
+
+	return target.store.DelList(keysToDelete)
+}
+
+func (target *KafkaTarget) toProducerMessage(eventData event.Event) (*sarama.ProducerMessage, error) {
+	objectName, err := url.QueryUnescape(eventData.S3.Object.Key)
+	if err != nil {
+		return nil, err
+	}
+
+	key := eventData.S3.Bucket.Name + "/" + objectName
+	data, err := json.Marshal(event.Log{EventName: eventData.EventName, Key: key, Records: []event.Event{eventData}})
+	if err != nil {
+		return nil, err
+	}
+
+	return &sarama.ProducerMessage{
+		Topic: target.args.Topic,
+		Key:   sarama.StringEncoder(key),
+		Value: sarama.ByteEncoder(data),
+	}, nil
 }
 
 // Close - closes underneath kafka connection.
@@ -394,8 +462,16 @@ func NewKafkaTarget(id string, args KafkaArgs, loggerOnce logger.LogOnce) (*Kafk
 	}
 
 	if target.store != nil {
+		if args.BatchSize > 1 {
+			target.batch = store.NewBatch[batchItem](args.BatchSize)
+		}
 		store.StreamItems(target.store, target, target.quitCh, target.loggerOnce)
 	}
 
 	return target, nil
+}
+
+func isKafkaConnErr(err error) bool {
+	// Sarama opens the ciruit breaker after 3 consecutive connection failures.
+	return err == sarama.ErrLeaderNotAvailable || err.Error() == "circuit breaker is open"
 }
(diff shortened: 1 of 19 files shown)


mctoohey:
@Praveenrajmani thanks for looking into this! 

One thing I found when trying this out was that although the performance of sending messages to Kafka with 'SendMessages()` is very fast. The collecting of the batches seems to be slow. Which surprises me since I would think pulling events off the queue and putting in the batch would be fast.

I added a log before and after calling `SendMessages` which shows it is fast (taking around a second). But the time between sending batches is around 14 seconds.
```
2023/10/04 09:01:09 Starting Send Messages
2023/10/04 09:01:10 Finishing Send Messages
2023/10/04 09:01:23 Starting Send Messages
2023/10/04 09:01:24 Finishing Send Messages
2023/10/04 09:01:51 Starting Send Messages
2023/10/04 09:01:51 Finishing Send Messages
2023/10/04 09:02:05 Starting Send Messages
2023/10/04 09:02:05 Finishing Send Messages
2023/10/04 09:02:19 Starting Send Messages
```

This was after creating 1000 events and using a batch_size of 100.

internal/store/batch.go
```diff
@@ -0,0 +1,86 @@
...
+	return nil
+}
+
+// Get fetches the items and resets the batch
+func (b *Batch[T]) Get() (items []T) {
+	b.Lock()
+	defer b.Unlock()
+
+	items = append([]T(nil), b.items...)
+	b.items = make([]T, 0, b.limit)
```
klauspost: No need to copy AND create a new destination.
```suggestion
	items = append([]T(nil), b.items...)
	b.items = b.items[:0]
```
NewBatch ensures that this will not fail.

internal/store/batch.go
```diff
@@ -0,0 +1,86 @@
...
+
+	if b.isFull() {
+		return ErrBatchFull
+	}
+
+	b.items = append(b.items, item)
+	return nil
+}
+
+// Get fetches the items and resets the batch
```
klauspost:
```suggestion
// Get fetches the items and resets the batch.
// Returned items are not referenced by the batch.
```

klauspost:
@mctoohey We will need some more information. Check the time for `DelList` and whatever else could be taking time.

mctoohey:
@klauspost DelList is also very fast. I did some more digging and found that it is the `SendFromStore` that is slow (even when just batching events).
```
[DEBUG] 2023-10-04T10:25:38.038Z Starting SendFromStore
[DEBUG] 2023-10-04T10:25:38.038Z Checking target.isActive
[DEBUG] 2023-10-04T10:25:38.166Z Done checking target.isActive
[DEBUG] 2023-10-04T10:25:38.166Z Getting Event
[DEBUG] 2023-10-04T10:25:38.166Z Got Event
```
This turns out to be due to the `target.isActive()` check. This makes sense as this check has to deal with the same network latency we are trying to avoid by batching.

I tried removing this check and performance became significantly better!

One possible solution could be when batching is enabled we only do this check before commiting a batch rather than every time an event is read of the store.

Praveenrajmani:
> @klauspost DelList is also very fast. I did some more digging and found that it is the `SendFromStore` that is slow (even when just batching events).
> 
> ```
> [DEBUG] 2023-10-04T10:25:38.038Z Starting SendFromStore
> [DEBUG] 2023-10-04T10:25:38.038Z Checking target.isActive
> [DEBUG] 2023-10-04T10:25:38.166Z Done checking target.isActive
> [DEBUG] 2023-10-04T10:25:38.166Z Getting Event
> [DEBUG] 2023-10-04T10:25:38.166Z Got Event
> ```
> 
> This turns out to be due to the `target.isActive()` check. This makes sense as this check has to deal with the same network latency we are trying to avoid by batching.
> 
> I tried removing this check and performance became significantly better!
> 
> One possible solution could be when batching is enabled we only do this check before commiting a batch rather than every time an event is read of the store.

It is due to the producer config that we have currently @mctoohey 

`SendMessages()` takes some time to open the circuit only for the first initial connection. I was able to see this with the following diff

```diff
diff --git a/internal/event/target/kafka.go b/internal/event/target/kafka.go
index c6010f7b9..278c68802 100644
--- a/internal/event/target/kafka.go
+++ b/internal/event/target/kafka.go
@@ -297,6 +297,7 @@ func (target *KafkaTarget) commitBatch() error {
                msgs = append(msgs, item.message)
        }
 
+       fmt.Printf("\n\n[DEBUG][%v] Sending Messages", time.Now().String())
        err := target.producer.SendMessages(msgs)
        if err != nil {
                if isKafkaConnErr(err) {
@@ -304,6 +305,7 @@ func (target *KafkaTarget) commitBatch() error {
                }
                return err
        }
+       fmt.Printf("\n[DEBUG][%v] Sent Messages", time.Now().String())
 
        return target.store.DelList(keysToDelete)
 }
```

```
[DEBUG][2023-10-04 11:25:03.733370956 +0000 UTC m=+186.208199543] Sending Messages
[DEBUG][2023-10-04 11:25:13.930117267 +0000 UTC m=+196.404945799] Sent Messages

[DEBUG][2023-10-04 11:25:14.039459316 +0000 UTC m=+196.514287887] Sending Messages
[DEBUG][2023-10-04 11:25:14.064997733 +0000 UTC m=+196.539826282] Sent Messages

[DEBUG][2023-10-04 11:25:14.143159728 +0000 UTC m=+196.617988270] Sending Messages
[DEBUG][2023-10-04 11:25:14.15532851 +0000 UTC m=+196.630157042] Sent Messages

[DEBUG][2023-10-04 11:25:15.186648632 +0000 UTC m=+197.661477281] Sending Messages
[DEBUG][2023-10-04 11:25:15.206489339 +0000 UTC m=+197.681317925] Sent Messages

[DEBUG][2023-10-04 11:25:15.241533079 +0000 UTC m=+197.716361629] Sending Messages
[DEBUG][2023-10-04 11:25:15.247207513 +0000 UTC m=+197.722036042] Sent Messages

[DEBUG][2023-10-04 11:25:15.294461843 +0000 UTC m=+197.769290381] Sending Messages
[DEBUG][2023-10-04 11:25:15.300369183 +0000 UTC m=+197.775197720] Sent Messages
```

In the latest commit, i have reduced the retry timeouts. Now, the producer doesn't take much time to initiate the connection.

With the latest commit :- 

```
[DEBUG][2023-10-04 11:44:14.414252626 +0000 UTC m=+126.190305943] Sending Messages
[DEBUG][2023-10-04 11:44:15.522537337 +0000 UTC m=+127.298590576] Sent Messages

[DEBUG][2023-10-04 11:44:15.631660638 +0000 UTC m=+127.407713936] Sending Messages
[DEBUG][2023-10-04 11:44:15.676465783 +0000 UTC m=+127.452519039] Sent Messages

[DEBUG][2023-10-04 11:44:15.718363943 +0000 UTC m=+127.494417184] Sending Messages
[DEBUG][2023-10-04 11:44:15.734146094 +0000 UTC m=+127.510199334] Sent Messages
```

you can try with the latest change @mctoohey

internal/event/target/kafka.go
```diff
@@ -335,16 +403,16 @@ func (target *KafkaTarget) initKafka() error {
 	// These settings are needed to ensure that kafka client doesn't hang on brokers
 	// refer https://github.com/IBM/sarama/issues/765#issuecomment-254333355
 	config.Producer.Retry.Max = 2
-	config.Producer.Retry.Backoff = (10 * time.Second)
+	config.Producer.Retry.Backoff = (1 * time.Second)
```
Praveenrajmani, line 394: I have reduced the retry backoff to 1sec @harshavardhana

mctoohey:
> > @klauspost DelList is also very fast. I did some more digging and found that it is the `SendFromStore` that is slow (even when just batching events).
> > ```
> > [DEBUG] 2023-10-04T10:25:38.038Z Starting SendFromStore
> > [DEBUG] 2023-10-04T10:25:38.038Z Checking target.isActive
> > [DEBUG] 2023-10-04T10:25:38.166Z Done checking target.isActive
> > [DEBUG] 2023-10-04T10:25:38.166Z Getting Event
> > [DEBUG] 2023-10-04T10:25:38.166Z Got Event
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This turns out to be due to the `target.isActive()` check. This makes sense as this check has to deal with the same network latency we are trying to avoid by batching.
> > I tried removing this check and performance became significantly better!
> > One possible solution could be when batching is enabled we only do this check before commiting a batch rather than every time an event is read of the store.
> 
> It is due to the producer config that we have currently @mctoohey
> 
> `SendMessages()` takes some time to open the circuit only for the first initial connection. I was able to see this with the following diff
> 
> ```diff
> diff --git a/internal/event/target/kafka.go b/internal/event/target/kafka.go
> index c6010f7b9..278c68802 100644
> --- a/internal/event/target/kafka.go
> +++ b/internal/event/target/kafka.go
> @@ -297,6 +297,7 @@ func (target *KafkaTarget) commitBatch() error {
>                 msgs = append(msgs, item.message)
>         }
>  
> +       fmt.Printf("\n\n[DEBUG][%v] Sending Messages", time.Now().String())
>         err := target.producer.SendMessages(msgs)
>         if err != nil {
>                 if isKafkaConnErr(err) {
> @@ -304,6 +305,7 @@ func (target *KafkaTarget) commitBatch() error {
>                 }
>                 return err
>         }
> +       fmt.Printf("\n[DEBUG][%v] Sent Messages", time.Now().String())
>  
>         return target.store.DelList(keysToDelete)
>  }
> ```
> 
> ```
> [DEBUG][2023-10-04 11:25:03.733370956 +0000 UTC m=+186.208199543] Sending Messages
> [DEBUG][2023-10-04 11:25:13.930117267 +0000 UTC m=+196.404945799] Sent Messages
> 
> [DEBUG][2023-10-04 11:25:14.039459316 +0000 UTC m=+196.514287887] Sending Messages
> [DEBUG][2023-10-04 11:25:14.064997733 +0000 UTC m=+196.539826282] Sent Messages
> 
> [DEBUG][2023-10-04 11:25:14.143159728 +0000 UTC m=+196.617988270] Sending Messages
> [DEBUG][2023-10-04 11:25:14.15532851 +0000 UTC m=+196.630157042] Sent Messages
> 
> [DEBUG][2023-10-04 11:25:15.186648632 +0000 UTC m=+197.661477281] Sending Messages
> [DEBUG][2023-10-04 11:25:15.206489339 +0000 UTC m=+197.681317925] Sent Messages
> 
> [DEBUG][2023-10-04 11:25:15.241533079 +0000 UTC m=+197.716361629] Sending Messages
> [DEBUG][2023-10-04 11:25:15.247207513 +0000 UTC m=+197.722036042] Sent Messages
> 
> [DEBUG][2023-10-04 11:25:15.294461843 +0000 UTC m=+197.769290381] Sending Messages
> [DEBUG][2023-10-04 11:25:15.300369183 +0000 UTC m=+197.775197720] Sent Messages
> ```
> 
> In the latest commit, i have reduced the retry timeouts. Now, the producer doesn't take much time to initiate the connection.
> 
> With the latest commit :-
> 
> ```
> [DEBUG][2023-10-04 11:44:14.414252626 +0000 UTC m=+126.190305943] Sending Messages
> [DEBUG][2023-10-04 11:44:15.522537337 +0000 UTC m=+127.298590576] Sent Messages
> 
> [DEBUG][2023-10-04 11:44:15.631660638 +0000 UTC m=+127.407713936] Sending Messages
> [DEBUG][2023-10-04 11:44:15.676465783 +0000 UTC m=+127.452519039] Sent Messages
> 
> [DEBUG][2023-10-04 11:44:15.718363943 +0000 UTC m=+127.494417184] Sending Messages
> [DEBUG][2023-10-04 11:44:15.734146094 +0000 UTC m=+127.510199334] Sent Messages
> ```
> 
> you can try with the latest change @mctoohey

I have tried the latest change but am still seeing the same issue.

The performance of `SendMessages()` seems good to me. What I am referring to is poor performance when messages are being added to the batch (and `SendMessages()` is not called).

The bottleneck seems to be the `target.isActive()` call in `SendFromStore` which takes about 100ms (in my case) limiting the rate at which messages can be added to the batch to around 10 per second. Removing this call makes the time to add a message to the batch negligible and greatly increases the throughput.

Praveenrajmani:
> added to the batch to around 10 per second

this was because of the 10 second retry backoff

anyway, we can skip checking the targets when we won't try to `sendMessages()` to the target. will make the change.

internal/store/batch.go
```diff
@@ -0,0 +1,87 @@
...
+	"errors"
+	"sync"
+)
+
+// ErrBatchFull indicates that the batch is full
+var ErrBatchFull = errors.New("batch is full")
+
+// Batch can be used to replay the events in batches
+type Batch[T any] struct {
+	items []T
```
Praveenrajmani: will change this to `map` as we can avoid duplicates.

mctoohey:
> this was because of the 10 second retry backoff

I disagree with this. I do not see how the 10 second retry backoff comes into it when Kafka is up and healthy. I still think the reason I was seeing it take 1 second to add 10 messages to the batch was because `isActive()` takes 0.1 seconds to run. So processing 10 messages takes 10*0.1 seconds.

However, if you make the change to skip the target check when we don't `sendMessages()` I will be happy as this resolves the issue I am seeing 😃

internal/event/target/kafka.go
```diff
@@ -73,6 +74,9 @@ const (
 	EnvKafkaClientTLSCert = "MINIO_NOTIFY_KAFKA_CLIENT_TLS_CERT"
 	EnvKafkaClientTLSKey  = "MINIO_NOTIFY_KAFKA_CLIENT_TLS_KEY"
 	EnvKafkaVersion       = "MINIO_NOTIFY_KAFKA_VERSION"
+	EnvKafkaBatchSize     = "MINIO_NOTIFY_KAFKA_BATCH_SIZE"
+
+	maxBatchLimit = 100
```
mctoohey, line 79: Curious as to the reasoning for choosing 100 as the `maxBatchLimit`. I can potentially see wanting to go higher than this, although in a production setting where Kafka is deployed very close to MinIO 100 should be sufficient.

Praveenrajmani:
> I do not see how the 10 second retry backoff comes into it when Kafka is up and healthy

yes, correct. there won't be any backoff when the target is up and active. If you see here https://github.com/minio/minio/pull/18164#issuecomment-1746743486, there was a 10 sec backoff for the first initial connection. We don't see this for each event.

The latency depends on the target's connectivity too.

Anyway, we don't need to check the connectivity when we know we are going to batch the incoming event.

internal/event/target/kafka.go
Praveenrajmani: the idea is to maintain a threshold on the async threads. we need to decide an optimal value. 100 seems to a "good-enough" value.

harshavardhana (approved):

PR #18173, opened 2023-10-05
Fix: check for disk-level O_DIRECT support

donatello:

## Description

Disk level O_DIRECT support checking at xl storage initialization was conditional on a config setting being enabled. (This never took effect because config initialization happens after ObjectLayer is ready.) This is not necessary as the config setting is dynamic - O_DIRECT should be enabled via runtime config. So we need to do the disk level support check regardless of the config setting.

## Types of changes

## Checklist:
- [x] Fixes a regression (not sure when it was introduced.)


harshavardhana:
This is on purpose @donatello we do not want to check for o_direct support if it's not enabled.

By default o_direct value in config is on so a default value must be on.

harshavardhana (approved):

harshavardhana (approved):

PR #18222, opened 2023-10-12
runtime error: index out of range

r-scheele:


harshavardhana (changes requested):
@r-scheele this is not the correct fix the crash is because

```
panic: runtime error: index out of range [0] with length 0

goroutine 1450 [running]:
github.com/minio/minio/cmd.poolMeta.IsSuspended(...)
	e:/gopath/src/github.com/minio/minio/cmd/erasure-server-pool-decom.go:311
github.com/minio/minio/cmd.(*erasureServerPools).IsSuspended(0x5358c40?, 0xc001c2bb30?)
```

The `poolMeta` is not initialized yet that's why there is a crash, since the server has not initialized yet we should return an error early on here.

And also add protection inside `IsSuspended()` function

This PR doesn't do anything and is not fixing the panic

harshavardhana:
Fixed via #18230

PR #18225, opened 2023-10-12
Healing tweaks

klauspost:
## Description

Null version ID would cause the wrong error to be returned (version not found) on unversioned objects.

Include version id result on errors as well.

## Types of changes

### DIFF
--- a/cmd/erasure-healing.go
+++ b/cmd/erasure-healing.go
@@ -23,6 +23,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


harshavardhana (approved):

cmd/erasure-healing.go
```diff
@@ -380,6 +381,9 @@ func (er *erasureObjects) healObject(ctx context.Context, bucket string, object
 	dryRun := opts.DryRun
 	scanMode := opts.ScanMode
 
+	if versionID == nullVersionID {
+		versionID = ""
+	}
```
harshavardhana: this may not be correct for versioned buckets?

harshavardhana (changes requested):

cmd/erasure-healing.go
klauspost: @harshavardhana I can't tell, but all code below uses "", and "null" IS sent as a value.

klauspost: I will remove it, but someone should really check this and all uses of it below.

vadmeste (approved):
LGTM

harshavardhana (approved):

PR #18270, opened 2023-10-17
cache DiskInfo() metrics call separately

harshavardhana:

## Description
cache DiskInfo() metrics call separately

## Motivation and Context
caching variables is not a good idea,
introduce another cache for metrics
DiskInfo() alone.

## How to test this PR?
Just a bug-fix nothing should change.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/storage-rest-client.go
+++ b/cmd/storage-rest-client.go
@@ -144,7 +144,8 @@ type storageRESTClient struct {
 	// Indexes, will be -1 until assigned a set.
 	poolIndex, setIndex, diskIndex int
 
-	diskInfoCache timedValue
+	diskInfoCache        timedValue
+	diskInfoCacheMetrics timedValue
 }
 
 // Retrieve location indexes.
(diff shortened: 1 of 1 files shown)


klauspost:
I have rewritten this for websockets.

Since we are fetching it only every second, I just always fetch metrics remotely.

harshavardhana:
> I have rewritten this for websockets.
> 
> Since we are fetching it only every second, I just always fetch metrics remotely.

No, we don't need to @klauspost we shouldn't send it when it's not needed. a DiskInfo() is used without this information for the most part. The metrics itself is a large information that we should avoid over the network IMO.

klauspost:
@harshavardhana Sure, if it is that slow. I will integrate it.

klauspost:
@harshavardhana Added here: https://github.com/minio/minio/pull/18263/commits/773630603b01598a771194534cde07076272a3a4 (set a 10 second timeout)

vadmeste (approved):
LGTM

PR #18302, opened 2023-10-23
fix: SR: Add more info when IAM config differs

donatello:
## Motivation and Context

Improve debuggability.

## Types of changes


cmd/site-replication.go
```diff
@@ -641,6 +641,8 @@ func (c *SiteReplicationSys) validateIDPSettings(ctx context.Context, peers []Pe
 
 	for i := 1; i < len(s); i++ {
 		if !reflect.DeepEqual(s[i], s[0]) {
+			logger.Info("IDP settings mismatch for peer \"%s\": %#v expected same as for peer \"%s\" %#v",
+				peers[i].Name, s[i], peers[0].Name, s[0])
```
harshavardhana: we should return an error here adding info() in this manner is not useful.

harshavardhana (changes requested):

cmd/site-replication.go
donatello: Ok done.

poornas (approved):

vadmeste (approved):
LGTM

harshavardhana (approved):

PR #18307, opened 2023-10-24
Improve multipart listing speed

klauspost:
## Description

Store initiated as part of the part ID. Avoid one `ReadVersion` per part.

Not sure if we should just delete the old code. Given the small usage of `ListMultipartUploads` I wouldn't mind.

## How to test this PR?

Regaular tests should cover it.

## Types of changes

### DIFF
--- a/cmd/erasure-multipart.go
+++ b/cmd/erasure-multipart.go
@@ -309,20 +309,36 @@ func (er erasureObjects) ListMultipartUploads(ctx context.Context, bucket, objec
 		if populatedUploadIds.Contains(uploadID) {
 			continue
 		}
-		fi, err := disk.ReadVersion(ctx, minioMetaMultipartBucket, pathJoin(er.getUploadIDDir(bucket, object, uploadID)), "", false)
-		if err != nil {
-			if !IsErrIgnored(err, errFileNotFound, errDiskNotFound) {
-				logger.LogIf(ctx, err)
+		// If present, use time stored in ID.
+		if split := strings.Split(uploadID, "."); len(split) == 2 {
+			startTime, err := strconv.ParseInt(split[1], 10, 64)
+			if err != nil {
+				continue
 			}
-			// Ignore this invalid upload-id since we are listing here
-			continue
+			uploads = append(uploads, MultipartInfo{
+				Bucket:    bucket,
+				Object:    object,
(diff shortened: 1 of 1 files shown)


cmd/erasure-multipart.go
```diff
@@ -309,20 +309,36 @@ func (er erasureObjects) ListMultipartUploads(ctx context.Context, bucket, objec
...
-			// Ignore this invalid upload-id since we are listing here
-			continue
+			uploads = append(uploads, MultipartInfo{
+				Bucket:    bucket,
+				Object:    object,
+				UploadID:  base64.RawURLEncoding.EncodeToString([]byte(fmt.Sprintf("%s.%s", globalDeploymentID(), uploadID))),
+				Initiated: time.Unix(0, startTime),
+			})
+		} else {
+			// No new parts should be created without time.
```
harshavardhana: We don't need to care for existing ones @klauspost we can simply ignore and move on. 

Less code to carry.

harshavardhana (changes requested):

harshavardhana (approved):

cmd/erasure-multipart.go
```diff
@@ -309,20 +309,21 @@ func (er erasureObjects) ListMultipartUploads(ctx context.Context, bucket, objec
...
+		startTime := time.Now()
+		if split := strings.Split(uploadID, "x"); len(split) == 2 {
+			t, err := strconv.ParseInt(split[1], 10, 64)
+			if err == nil {
+				startTime = time.Unix(0, t)
 			}
-			// Ignore this invalid upload-id since we are listing here
-			continue
 		}
-		populatedUploadIds.Add(uploadID)
```
das7pad, line 320: In passing: Was the removal of `populatedUploadIds.Add` intentional? If so, noting is adding to `populatedUploadIds` and it looks like the check `populatedUploadIds.Contains` in line 309 can get removed.

klauspost: Let me check.

klauspost: I can't really see how ListDir can return duplicate entries. But let's re-add it, since it is a cheap sanity check.

klauspost: https://github.com/minio/minio/pull/18339

PR #18310, opened 2023-10-24
Add check for reverse proxy setups

allanrogerr:
Add check for reverse proxy setups, to skip check for paths being served by different port on same address.

## Description
In a minio multi node cluster, we may run multiple minio processes behind multiple reverse proxy servers. In this situation,  the following error is encountered, and prevents minio startup. This error is invalid in a reverse proxy setup:
```
ERROR Invalid command line arguments: same path '/tmp/data1' can not be served by different port on same address
```
See https://github.com/minio/minio/blob/master/cmd/endpoint.go#L917

The solution here is to use a new environment variable `_MINIO_REVERSE_PROXY` to indicate a reverse proxy and therefore skip the above test.  This new variable will be used in conjunction with the existing environment variables: `CI` and `MINIO_CI_CD`.

Essentially, 
if `CI=true or MINIO_CI_CD=true is set and _MINIO_REVERSE_PROXY is set`
then skip check for whether the same path is used in endpoints of a host on different port.

## Motivation and Context
When running multiple reverse proxy servers, the hosts are not all on the same network. Routing may be achieved over the internet using FQDNs. If the same server address handles multiple reverse proxied hosts, then the resolved IP addresses of these reverse proxied hosts are all reported as the same - thus the above error.

This sort of setup should be possible on minio i.e multiple reverse proxy servers handling multiple reverse proxied hosts, where the resolved IP addresses of these reverse proxied hosts are all reported as the same.

## How to test this PR?
Create a minio cluster with multiple reverse proxy servers handling multiple reverse proxied hosts, where the resolved IP addresses of these reverse proxied hosts are all reported as the same.

The cluster should be able to be load balanced and should be accessible. 

## Types of changes

## Checklist:


allanrogerr:
A demonstration here: https://github.com/allanrogerr/public/wiki/minio-vm%E2%80%90broker-load-balance-minio

harshavardhana (approved):

vadmeste (approved):
LGTM

cmd/endpoint.go
```diff
@@ -1113,32 +1115,34 @@ func CreateEndpoints(serverAddr string, args ...[]string) (Endpoints, SetupType,
 	}
 
 	orchestrated := IsKubernetes() || IsDocker()
-	if !orchestrated {
+	reverseProxy := (env.Get("_MINIO_REVERSE_PROXY", "") != "") && ((env.Get("MINIO_CI_CD", "") != "") || (env.Get("CI", "") != ""))
```
harshavardhana, line 1118:
```suggestion
	reverseProxy := (env.Get("_MINIO_REVERSE_PROXY", "") != "") && ((env.Get("MINIO_CI_CD", "") != "") || (env.Get("CI", "") != ""))
```

This can be set as a global value @allanrogerr so that it can re-used elsewhere. 

```
var globalIsReverseProxy = (env.Get("_MINIO_REVERSE_PROXY", "") != "") && ((env.Get("MINIO_CI_CD", "") != "") || (env.Get("CI", "") != ""))
```

in globals.go

harshavardhana (changes requested):

cmd/endpoint.go
harshavardhana: oh I see its already approved by me, let me merge this anyways you can fix it another PR.

harshavardhana (approved):

PR #18327, opened 2023-10-27
Clean stale left out buckets

shtripat:

## Description
If a bucket got removed while few nodes/disks were offline in cluster, while nodes/disks come back online, the stale bucket lies around as there is no bucket heal which gets triggered on its own. While scanning, if we find such stale buckets which dont have quorum anymore, safely we remove them now.

## How to test this PR?
Step-1: Start a 4 node, 1 disk each MinIO setup as below
```
minio server http://127.0.0.1:9000/tmp/bkt-test/1/ http://127.0.0.1:9001/tmp/bkt-test/2/ http://127.0.0.1:9002/tmp/bkt-test/3/ http://127.0.0.1:9003/tmp/bkt-test/4/

minio server http://127.0.0.1:9000/tmp/bkt-test/1/ http://127.0.0.1:9001/tmp/bkt-test/2/ http://127.0.0.1:9002/tmp/bkt-test/3/ http://127.0.0.1:9003/tmp/bkt-test/4/ --address :9001

minio server http://127.0.0.1:9000/tmp/bkt-test/1/ http://127.0.0.1:9001/tmp/bkt-test/2/ http://127.0.0.1:9002/tmp/bkt-test/3/ http://127.0.0.1:9003/tmp/bkt-test/4/ --address :9002

minio server http://127.0.0.1:9000/tmp/bkt-test/1/ http://127.0.0.1:9001/tmp/bkt-test/2/ http://127.0.0.1:9002/tmp/bkt-test/3/ http://127.0.0.1:9003/tmp/bkt-test/4/ --address :9003
```

Step-2: Create two buckets `bkt1` and `bkt2` and load objects to them
```
$ mc alias set myminio http://127.0.0.1:9000 minioadmin minioadmin
$ mc mb myminio/bkt1
$ mc mb myminio/bkt2
$ mc cp /etc/hosts myminio/bkt2
```

Step-3: Now bring down the last MinIO instance and remove first bucket
`mc rb myminio/bkt1 --force`

Step-4: Verify that last disk only contains the stale bucket xl.meta
```
$ ll /tmp/bkt-test/1 
total 4
drwxr-xr-x. 1 shubhendu shubhendu 88 Oct 27 11:23 .minio.sys
drwxr-xr-x. 1 shubhendu shubhendu  8 Oct 27 11:23 ..
drwxr-xr-x. 1 shubhendu shubhendu 10 Oct 27 11:24 bkt2
drwxr-xr-x. 1 shubhendu shubhendu 28 Oct 27 11:25 .

$ ll /tmp/bkt-test/2 
total 4
drwxr-xr-x. 1 shubhendu shubhendu 88 Oct 27 11:23 .minio.sys
drwxr-xr-x. 1 shubhendu shubhendu  8 Oct 27 11:23 ..
drwxr-xr-x. 1 shubhendu shubhendu 10 Oct 27 11:24 bkt2
drwxr-xr-x. 1 shubhendu shubhendu 28 Oct 27 11:25 .

$ ll /tmp/bkt-test/3
total 4
drwxr-xr-x. 1 shubhendu shubhendu 88 Oct 27 11:23 .minio.sys
drwxr-xr-x. 1 shubhendu shubhendu  8 Oct 27 11:23 ..
drwxr-xr-x. 1 shubhendu shubhendu 10 Oct 27 11:24 bkt2
drwxr-xr-x. 1 shubhendu shubhendu 28 Oct 27 11:25 .

$ ll /tmp/bkt-test/4
total 4
drwxr-xr-x. 1 shubhendu shubhendu  8 Oct 27 11:23 ..
drwxr-xr-x. 1 shubhendu shubhendu 36 Oct 27 11:24 .
drwxr-xr-x. 1 shubhendu shubhendu 10 Oct 27 11:24 bkt2
drwxr-xr-x. 1 shubhendu shubhendu 10 Oct 27 11:24 bkt1 <<<<<<<<========== HERE
drwxr-xr-x. 1 shubhendu shubhendu 72 Oct 27 11:24 .minio.sys
```

Step-5: Now start the last MinIO instance back and wait for few secs. Verify that last disk also dosn't contain the removed disk xl.meta anymore
```
$ ll /tmp/bkt-test/4 
total 4
drwxr-xr-x. 1 shubhendu shubhendu  8 Oct 27 11:23 ..
drwxr-xr-x. 1 shubhendu shubhendu 10 Oct 27 11:24 bkt2
drwxr-xr-x. 1 shubhendu shubhendu 72 Oct 27 11:25 .minio.sys
drwxr-xr-x. 1 shubhendu shubhendu 28 Oct 27 11:25 .
```

Without the fix, last disk retains the traces of removed bucket (while 4th MinIO instance was down). This fix as well will remove any dangling bucket which don't contain any data. If bucket contained data, it just logs as error once that dangling bucket cleanup didn't happen as bucket was not empty
```
While deleting dangling Bucket (bkt1), Drive /tmp/bkt-test/4 returned an error (volume is not empty) (*fmt.wrapError)
```

## Types of changes

## Checklist:

### DIFF
--- a/cmd/erasure-healing.go
+++ b/cmd/erasure-healing.go
@@ -224,6 +224,19 @@ func (er erasureObjects) healBucket(ctx context.Context, storageDisks []StorageA
 		})
 	}
 
+	// check dangling and delete bucket only if its not a meta bucket
+	if !isMinioMetaBucketName(bucket) {
+		// read all bucket metadata
+		bktMetadata, errs := readAllBucketInfo(ctx, storageDisks, bucket)
+		if isAllBucketsNotFound(errs) {
+			// nothing to be done, bucket already gone
+			return res, errVolumeNotFound
+		}
+		if err := er.deleteBucketIfDangling(ctx, bucket, bktMetadata, errs); err != nil {
+			return res, err
+		}
+	}
+
 	reducedErr := reduceReadQuorumErrs(ctx, errs, bucketOpIgnoredErrs, res.DataBlocks)
 	if errors.Is(reducedErr, errVolumeNotFound) && !opts.Recreate {
 		for i := range beforeState {
@@ -1019,6 +1032,22 @@ func isAllNotFound(errs []error) bool {
 	return len(errs) > 0
 }
 
+// isAllBucketsNotFound will return true if all the errors are either errFileNotFound
+// or errFileCorrupt
+// A 0 length slice will always return false.
+func isAllBucketsNotFound(errs []error) bool {
+	if len(errs) == 0 {
+		return false
+	}
+	notFoundCount := 0
+	for _, err := range errs {
+		if err != nil && (err.Error() == errFileNotFound.Error() || err.Error() == errFileCorrupt.Error()) {
+			notFoundCount++
+		}
+	}
+	return len(errs) == notFoundCount
+}
+
 // ObjectDir is considered dangling/corrupted if any only
 // if total disks - a combination of corrupted and missing
 // files is lesser than N/2+1 number of disks.
@@ -1044,7 +1073,7 @@ func isObjectDirDangling(errs []error) (ok bool) {
 	return found < notFound && found > 0
 }
 
-// Object is considered dangling/corrupted if any only
+// Object is considered dangling/corrupted if and only
 // if total disks - a combination of corrupted and missing
 // files is lesser than number of data blocks.
 func isObjectDangling(metaArr []FileInfo, errs []error, dataErrs []error) (validMeta FileInfo, ok bool) {
@@ -1131,6 +1160,35 @@ func isObjectDangling(metaArr []FileInfo, errs []error, dataErrs []error) (valid
 
 	// We have valid meta, now verify if we have enough files with parity blocks.
 	return validMeta, totalErrs > validMeta.Erasure.ParityBlocks
+}
+
+// Bucket is considered dangling/corrupted if and only
+// if bucket meta data is missing on more than N/2 disks
+func isBucketDangling(metaArr []VolInfo, errs []error) bool {
+	// We can consider a bucket not reliable
+	// when xl.meta is not found in read quorum disks.
+	// or when xl.meta is not readable in read quorum disks.
+	danglingErrsCount := func(cerrs []error) (int, int, int) {
+		var (
+			notFoundCount     int
+			corruptedCount    int
+			diskNotFoundCount int
+		)
+		for _, readErr := range cerrs {
+			switch {
+			case errors.Is(readErr, errVolumeNotFound):
+				notFoundCount++
+			case errors.Is(readErr, errFileCorrupt):
+				corruptedCount++
+			case errors.Is(readErr, errDiskNotFound):
+				diskNotFoundCount++
+			}
+		}
+		return notFoundCount, corruptedCount, diskNotFoundCount
+	}
+
+	notFoundMetaErrs, corruptedMetaErrs, _ := danglingErrsCount(errs)
+	return corruptedMetaErrs+notFoundMetaErrs > len(errs)/2
 }
 
 // HealObject - heal the given object, automatically deletes the object if stale/corrupted if `remove` is true.
--- a/cmd/erasure-healing.go
+++ b/cmd/erasure-healing.go
@@ -228,12 +228,10 @@ func (er erasureObjects) healBucket(ctx context.Context, storageDisks []StorageA
 	if !isMinioMetaBucketName(bucket) {
 		// read all bucket metadata
 		bktMetadata, errs := readAllBucketInfo(ctx, storageDisks, bucket)
-		if isAllBucketsNotFound(errs) {
-			// nothing to be done, bucket already gone
-			return res, errVolumeNotFound
-		}
-		if err := er.deleteBucketIfDangling(ctx, bucket, bktMetadata, errs); err != nil {
-			return res, err
+		if !isAllBucketsNotFound(errs) {
+			if err := er.deleteBucketIfDangling(ctx, bucket, bktMetadata, errs); err != nil {
+				return res, err
+			}
 		}
 	}
 
@@ -290,6 +288,7 @@ func (er erasureObjects) healBucket(ctx context.Context, storageDisks []StorageA
 			State:    afterState[i],
 		})
 	}
+
 	return res, nil
 }
 
@@ -1041,7 +1040,7 @@ func isAllBucketsNotFound(errs []error) bool {
 	}
 	notFoundCount := 0
 	for _, err := range errs {
-		if err != nil && (err.Error() == errFileNotFound.Error() || err.Error() == errFileCorrupt.Error()) {
+		if err != nil && (err.Error() == errVolumeNotFound.Error() || err.Error() == errFileCorrupt.Error()) {
 			notFoundCount++
 		}
 	}
--- a/cmd/erasure-object.go
+++ b/cmd/erasure-object.go
@@ -587,10 +587,11 @@ func (er erasureObjects) deleteBucketIfDangling(ctx context.Context, bucket stri
 				}
 				err := disks[index].DeleteVol(ctx, bucket, false)
 				if errors.Is(err, errVolumeNotEmpty) {
-					// If volume non empty, force delete as its dangling anyway
-					return disks[index].DeleteVol(ctx, bucket, true)
+					logger.LogOnceIf(ctx, fmt.Errorf("While deleting dangling Bucket (%s), Drive %s returned an error (%w)",
+						bucket, disks[index], err), "delete-dangling-bucket-"+bucket)
+					return err
 				}
-				return err
+				return nil
 			}, index)
 		}
 
--- a/cmd/erasure-object.go
+++ b/cmd/erasure-object.go
@@ -476,6 +476,20 @@ func auditDanglingObjectDeletion(ctx context.Context, bucket, object, versionID 
 	auditLogInternal(ctx, opts)
 }
 
+func auditDanglingBucketDeletion(ctx context.Context, bucket string, tags map[string]interface{}) {
+	if len(logger.AuditTargets()) == 0 {
+		return
+	}
+
+	opts := AuditLogOptions{
+		Event:  "DeleteDanglingBucket",
+		Bucket: bucket,
+		Tags:   tags,
+	}
+
+	auditLogInternal(ctx, opts)
+}
+
 func joinErrors(errs ...error) error {
 	s := make([]string, 0, len(errs))
 	nonNilErrs := make([]any, 0, len(errs))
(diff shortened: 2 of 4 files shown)


cmd/peer-s3-client.go
```diff
@@ -185,6 +185,14 @@ func (sys *S3PeerSys) ListBuckets(ctx context.Context, opts BucketOptions) ([]Bu
 				}
 			}
 		}
+		// loop through buckets and see if some with lost quorum
+		// these could be stale buckets lying around, remove them
+		for bktName, count := range bucketsMap {
+			if count < quorum {
+				// Its safe to remove as its stale bucket lying around
+				sys.DeleteBucket(ctx, bktName, DeleteBucketOptions{Force: true})
```
harshavardhana: quorum issue can happen where nodes are down, we cannot just assume that to be a precursor to delete a bucket entirely.

we should carefully look at the deleteDangling code for objects. We must be absolutely sure of what we are trying to do here.

This PR makes quorum-less a definitive way to purge the entire data automatically which would be unexpected.

shtripat: sure, makes sense. let me explore more and redo this.

vadmeste: I think let's try to avoid calling Deletebucket with Force set to true; It is okay if it fails to delete a bucket which is stale rather than always doing a force delete.

isObjectDangling(), this is a function that decides if an object is dangling or not; I think the logic is we only mark a bucket as dangling in well defined use cases; meaning if we do not have certain errors and secnarios, we should not dare to remove a bucket.

harshavardhana:
You have to basically calculate `errVolumeNotFound` on X copies that is greater than N/2 then we safely attempt to delete the bucket.

Force is not needed we can attempt first without force first, and then if this returns errBucketNotEmpty then we do a force delete.

cmd/erasure-healing.go
```diff
@@ -1019,6 +1032,22 @@ func isAllNotFound(errs []error) bool {
...
+// isAllBucketsNotFound will return true if all the errors are either errFileNotFound
+// or errFileCorrupt
+// A 0 length slice will always return false.
+func isAllBucketsNotFound(errs []error) bool {
+	if len(errs) == 0 {
+		return false
+	}
+	notFoundCount := 0
+	for _, err := range errs {
+		if err != nil && (err.Error() == errFileNotFound.Error() || err.Error() == errFileCorrupt.Error()) {
```
vadmeste: So isAllBucketsNotFound() takes errs from readAllBucketInfo();

readAllBucketInfo() runs xl-storage.StatVol(), which returns errVolumeNotFound, not errFileNotFound; 

errFileCorrupt is not returned as well, so no need to check for it

cmd/erasure-object.go
```diff
@@ -545,6 +559,46 @@ func (er erasureObjects) deleteIfDangling(ctx context.Context, bucket, object st
...
+		for index := range disks {
+			index := index
+			g.Go(func() error {
+				if disks[index] == nil {
+					return errDiskNotFound
+				}
+				err := disks[index].DeleteVol(ctx, bucket, false)
+				if errors.Is(err, errVolumeNotEmpty) {
+					// If volume non empty, force delete as its dangling anyway
+					return disks[index].DeleteVol(ctx, bucket, true)
```
vadmeste: I am really not in favor of doing Force Delete here; the benefit is low comparing to the risk

cmd/erasure-healing.go
```diff
@@ -1133,6 +1162,35 @@ func isObjectDangling(metaArr []FileInfo, errs []error, dataErrs []error) (valid
...
+		var (
+			notFoundCount     int
+			corruptedCount    int
+			diskNotFoundCount int
+		)
+		for _, readErr := range cerrs {
+			switch {
+			case errors.Is(readErr, errVolumeNotFound):
+				notFoundCount++
+			case errors.Is(readErr, errFileCorrupt):
```
vadmeste, line 1180: No file corrupt error in my opinion is returned in this case

cmd/erasure-healing.go
```diff
@@ -1133,6 +1162,35 @@ func isObjectDangling(metaArr []FileInfo, errs []error, dataErrs []error) (valid
...
+// if bucket meta data is missing on more than N/2 disks
+func isBucketDangling(metaArr []VolInfo, errs []error) bool {
+	// We can consider a bucket not reliable
+	// when xl.meta is not found in read quorum disks.
+	// or when xl.meta is not readable in read quorum disks.
+	danglingErrsCount := func(cerrs []error) (int, int, int) {
+		var (
+			notFoundCount     int
+			corruptedCount    int
+			diskNotFoundCount int
```
vadmeste, line 1174: This doesn't seem to be needed, you can remove it

vadmeste (commented):
Few comments

cmd/erasure-healing.go
shtripat: ah yes, that's a mistake. Will change to use `errVolumeNotFound` and remove `errFileCorrupt`

cmd/erasure-object.go
shtripat: ok, so no need to have force delete called at all? will remove.

cmd/erasure-healing.go
shtripat: ack. will remove

cmd/erasure-healing.go
shtripat: ack

shtripat: actually we are concerned about volume not found error `errVolumeNotFound` so we can calculate that count and use. will change accordingly

cmd/erasure-object.go
shtripat: As scanner process wont consider a dangling bucket for clearing up its dangling objects, and we are sure that volume metadata is found on more than N/2 nodes of the cluster, isn't it safe to remove bucket with force after first error with `errVolumeNotEmpty`?
@vadmeste @harshavardhana please suggest.

harshavardhana: My suggestion is @shtripat let's leave out the force part instead just perform normal DelVol if possible if its not they engage support and we investigate.

Keeping it manual is safer.

shtripat: ok, I will remove the force delete part here, but in actual dangling bucket scenario where objects as well left out, this code wont have any effect as bucket wont be deleted due to non empty. We can just log an error once that during removal of dangling buckets, it was not able to delete as bucket was non empty.

harshavardhana:
There is some persistent issue after this change for decommissioning - can you verify @shtripat ? `make test-decom`

shtripat:
> There is some persistent issue after this change for decommissioning - can you verify @shtripat ? `make test-decom`

Sure, I will check and do the needful.

cmd/peer-s3-client.go
```diff
@@ -185,6 +186,21 @@ func (sys *S3PeerSys) ListBuckets(ctx context.Context, opts BucketOptions) ([]Bu
 				}
 			}
 		}
+		// loop through buckets and see if some with lost quorum
+		// these could be stale buckets lying around, remove them
+		for bktName, count := range bucketsMap {
+			if count < quorum {
+				// Queue a bucket heal task
+				globalHealStateLK.Lock()
+				bgSeq, ok := globalBackgroundHealState.getHealSequenceByToken(bgHealingUUID)
```
harshavardhana: We should put this to MRF not global heal queue @shtripat 

A much simpler code to handle it's called er.addPartial()

harshavardhana (changes requested):

cmd/peer-s3-client.go
shtripat: Done

harshavardhana (approved):

cmd/erasure-healing.go
```diff
@@ -1133,6 +1161,35 @@ func isObjectDangling(metaArr []FileInfo, errs []error, dataErrs []error) (valid
 	return validMeta, totalErrs > validMeta.Erasure.ParityBlocks
 }
 
+// Bucket is considered dangling/corrupted if and only
+// if bucket meta data is missing on more than N/2 disks
+func isBucketDangling(metaArr []VolInfo, errs []error) bool {
+	// We can consider a bucket not reliable
+	// when xl.meta is not found in read quorum disks.
+	// or when xl.meta is not readable in read quorum disks.
```
harshavardhana, line 1169: these comments don't make sense for buckets

cmd/erasure-healing.go
```diff
@@ -1133,6 +1161,35 @@ func isObjectDangling(metaArr []FileInfo, errs []error, dataErrs []error) (valid
...
+			switch {
+			case errors.Is(readErr, errVolumeNotFound):
+				notFoundCount++
+			case errors.Is(readErr, errFileCorrupt):
+				corruptedCount++
+			case errors.Is(readErr, errDiskNotFound):
+				diskNotFoundCount++
+			}
+		}
+		return notFoundCount, corruptedCount, diskNotFoundCount
```
harshavardhana, line 1186:
```suggestion
		return notFoundCount, corruptedCount, diskNotFoundCount
```

diskNotFoundCount is not used? if there is any drive that is offline we shouldn't treat it as dangling and return right away.

cmd/erasure-healing.go
```diff
@@ -1133,6 +1161,35 @@ func isObjectDangling(metaArr []FileInfo, errs []error, dataErrs []error) (valid
...
+	}
+
+	notFoundMetaErrs, corruptedMetaErrs, _ := danglingErrsCount(errs)
```
harshavardhana, line 1189:
```suggestion
	notFoundMetaErrs, corruptedMetaErrs, diskNotFoundErrs := danglingErrsCount(errs)
	if diskNotFoundErrs > 0 {
	    return false
	}
```

harshavardhana (changes requested):
Unfortunately looks like we would have to re-implement this as part of the `peer` call such that dangling buckets are cleared based on the per pool quorum.

Look at the implementation at `s3-peer-client.go`

shtripat:
> Unfortunately looks like we would have to re-implement this as part of the `peer` call such that dangling buckets are cleared based on the per pool quorum.
> 
> Look at the implementation at `s3-peer-client.go`

Ok, I will relook and do needful.

cmd/erasure-healing.go
shtripat: Ack

cmd/erasure-healing.go
shtripat: Ack

cmd/erasure-healing.go
shtripat: Will remove

shtripat:
Closing this in as logic changed to move bucket healing to node level. New PR at https://github.com/minio/minio/pull/18504

PR #18329, opened 2023-10-27
switch minio container base image to ubi-mciro

aead:
## Description
This commit changes the container base image
from ubi-minimal to ubi-micro.

The docker build process happens now in two stages. The build stage:
 - downloads the latest CA certificate bundle
 - downloads MinIO binary (for requested version/os/arch)
 - downloads MinIO binary signature and verifies it using minisign

Then it creates an image based on ubi-micro with just the minio binary downloaded and verified during the build stage.

The build stage is simplified to just verifying the minisign signature.

## Motivation and Context
Docker

Size difference:
```
9603f4fc9deb    6 seconds ago   150MB (old)
8b2ef5ae2847   30 minutes ago   124MB (new)
```

## How to test this PR?
For example:
```
docker build --build-arg=RELEASE=RELEASE.2023-10-07T15-07-38Z --build-arg=TARGETARCH=amd64 -f Dockerfile.release .
```

## Types of changes

## Checklist:


Dockerfile.hotfix
```diff
@@ -1,12 +1,33 @@
...
 
 ARG RELEASE
 
 LABEL name="MinIO" \
       vendor="MinIO Inc <dev@min.io>" \
       maintainer="MinIO Inc <dev@min.io>" \
-      version="${RELEASE}" \
-      release="${RELEASE}" \
+      version="${RELEASE}.hotfix" \
+      release="${RELEASE}.hotfix" \
```
harshavardhana:
```suggestion
      version="${RELEASE}.hotfix" \
      release="${RELEASE}.hotfix" \
```

.hotfix is only meant for us we don't need to be added to RELEASE string here IMO

Dockerfile.release
```diff
@@ -18,37 +37,17 @@ ENV MINIO_ACCESS_KEY_FILE=access_key \
...
-RUN \
-     microdnf clean all && \
-     microdnf update --nodocs && \
-     rpm -Uvh https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm && \
-     microdnf install curl ca-certificates shadow-utils util-linux gzip lsof tar net-tools iproute iputils jq minisign --nodocs && \
-     mkdir -p /opt/bin && chmod -R 777 /opt/bin && \
-     curl -s -q https://dl.min.io/server/minio/release/linux-${TARGETARCH}/archive/minio.${RELEASE} -o /opt/bin/minio && \
-     curl -s -q https://dl.min.io/server/minio/release/linux-${TARGETARCH}/archive/minio.${RELEASE}.sha256sum -o /opt/bin/minio.sha256sum && \
-     curl -s -q https://dl.min.io/server/minio/release/linux-${TARGETARCH}/archive/minio.${RELEASE}.minisig -o /opt/bin/minio.minisig && \
-     curl -s -q https://dl.min.io/client/mc/release/linux-${TARGETARCH}/mc -o /opt/bin/mc && \
```
harshavardhana, line 39: We need `mc` inside the pod this cannot be removed.

Dockerfile.release
```diff
@@ -18,37 +37,17 @@ ENV MINIO_ACCESS_KEY_FILE=access_key \
     MINIO_ROOT_PASSWORD_FILE=secret_key \
     MINIO_KMS_SECRET_KEY_FILE=kms_master_key \
     MINIO_UPDATE_MINISIGN_PUBKEY="RWTx5Zr1tiHQLwG9keckT0c45M3AGeHD6IvimQHpyRywVWGbP1aVSGav" \
-    MINIO_CONFIG_ENV_FILE=config.env \
-    MC_CONFIG_DIR=/tmp/.mc \
```
harshavardhana, line 22: this can't be removed either.

Dockerfile.release.fips
```diff
@@ -18,35 +37,17 @@ ENV MINIO_ACCESS_KEY_FILE=access_key \
...
     MINIO_UPDATE_MINISIGN_PUBKEY="RWTx5Zr1tiHQLwG9keckT0c45M3AGeHD6IvimQHpyRywVWGbP1aVSGav" \
-    MINIO_CONFIG_ENV_FILE=config.env \
-    MC_CONFIG_DIR=/tmp/.mc \
-    PATH=/opt/bin:$PATH
+    MINIO_CONFIG_ENV_FILE=config.env
+
+COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
+COPY --from=build /go/bin/minio /usr/bin/minio
 
-COPY dockerscripts/verify-minio.sh /usr/bin/verify-minio.sh
```
harshavardhana, line 25: Remove this file if its no longer needed.

harshavardhana (changes requested):

Dockerfile.hotfix
aead: Okay 👍

Dockerfile.release
aead: Sure. What do we do for fips releases? There is no "latest" `mc.fips`  binary. Should we require a mc version for FIPS builds?

harshavardhana: Yes for FIPs container it's just .fips

harshavardhana (approved):

Dockerfile.release
aead: I'm not following... Do we want to include mc in FIPS builds? If so, which mc version should be fetched? `https://dl.min.io/client/mc/release/linux-amd64/mc.fips` does not exist. Either we have to create it from now on or specify a mc version container arg for the FIPS container.

harshavardhana: We can leave for now, I will add it as part of release process.

levrik:
This broke all our integration tests as `mc` which we use to import test fixtures isn't available on `PATH` anymore.

```
exec: "mc": executable file not found in $PATH: unknown
```

Tests stopped working ~15 minutes ago, the release containing this change went out ~30 minutes ago. We're simply pulling `latest`.

_Edit:_

Trying to run the executable directly with absolute path fails with the following:

```
exec: "/usr/bin/mc": permission denied: unknown
```

Looks like executable bit isn't set?

levrik:
Opened https://github.com/minio/minio/pull/18364 which fixes this regression!

Issue #18235, opened 2023-10-12
`filename` being incorrectly enforced - MalformedPOSTRequest - The file or text content is missing


jonbarrow:
Presigned POST URL is failing to upload multi-part data when `filename` is not present

## Expected Behavior
For the upload to succeed

## Current Behavior
Upload fails with the following error

**EDIT**:

This error is thrown because MinIO is enforcing that `filename` be present during multi-part file uploads. This is not to spec. The spec for multi-part uploads says that `filename` is *not* required https://www.ietf.org/rfc/rfc7578.txt

> For form data that represents the content of a file, a name for the file SHOULD be supplied as well, by using a "filename" parameter of the Content-Disposition header field.  _**The file name isn't mandatory for cases where the file name isn't available or is meaningless or private**_; this might result, for example, when selection or drag-and- drop is used or when the form data content is streamed directly from a device.

In cases like S3, where the "file name" is the same as the objects key in the bucket, "where the file name isn't available or is meaningless" applies

In our specific case, "when the form data content is streamed directly from a device" is what applies to us. Our server provides presigned URLs to a game client, which we do not control, and the console is what makes these upload requests as a way to store user generated content

```xml
<Error>
	<Code>MalformedPOSTRequest</Code>
	<Message>The body of your POST request is not well-formed multipart/form-data. (The file or text content is missing)</Message>
	<BucketName>pn-amaj-d1</BucketName>
	<Resource>/pn-amaj-d1/</Resource>
	<RequestId></RequestId>
	<HostId></HostId>
</Error>
```

## Possible Solution
No longer enforcing `filename` to be present on multi-part uploads, bringing things to spec. My team and I are willing to contribute this ourselves

## Steps to Reproduce (for bugs)
```go
policy := minio.NewPostPolicy()

policy.SetBucket(bucket)
policy.SetKey(key)
policy.SetExpires(time.Now().UTC().Add(lifetime))

URL, formData, err := minioClient.PresignedPostPolicy(context.Background(), policy)
if err != nil {
	panic(err)
}

// POST to the URL with the file named "file", omitting the "filename"
```

## Context
I am working with a client which I do not have control over. The client requests presigned s3 URLs to manage user-generated content. When giving the client a presigned POST URL, the upload fails due to the file name not being found despite it existing. Uploads have worked before in s3 cloud providers, such as DigitalOcean Spaces, just not here

The form upload looks like so (trimmed):

```
POST /pn-amaj-d1/ HTTP/1.1
[...]

PR #18240, opened 2023-10-13
fix: missing fileName will upload failed when PostPolicyBucketHandler

jiuker:
## Community Contribution License
All community contributions in this pull request are licensed to the project maintainers
under the terms of the [Apache 2 license] (https://www.apache.org/licenses/LICENSE-2.0). 
By creating this pull request I represent that I have the right to license the 
contributions to the project maintainers under the Apache 2 license.

## Description
fix https://github.com/minio/minio/issues/18235
fix: missing fileName will upload file failed
We use the keyName[0] as fileName when can't get that.

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Optimization (provides speedup with no functional changes)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [ ] Fixes a regression (If yes, please add `commit-id` or `PR #` here)
- [ ] Unit tests added/updated
- [ ] Internal documentation updated
- [ ] Create a documentation update request [here](https://github.com/minio/docs/issues/new?label=doc-change,title=Doc+Updated+Needed+For+PR+github.com%2fminio%2fminio%2fpull%2fNNNNN)

PR #18336, opened 2023-10-28
fix: relax enforcing filename on PostPolicy

DaniElectra:

## Description
The filename on the `file` field isn't required, but it's being used as reference for checking if we have reached the `file` field on the multipart fields loop. Check the current field using it's name instead.

## Motivation and Context
If the filename isn't set, the `fileName == ""` check is always true and the file reader never gets assigned, which causes a nullptr dereference when calling `PutObject` on `PostPolicyBucketHandler`. This fully fixes issue #18235

## How to test this PR?
```go
package main

import (
	"bytes"
	"context"
	"fmt"
	"log"
	"mime/multipart"
	"net/http"
	"time"

	"github.com/minio/minio-go/v7"
)

func main() {
	endpoint := "example.com"
	accessKeyID := "ACCESS_KEY"
	secretAccessKey := "ACCESS_SECRET"
	useSSL := true

	minioClient, err := minio.New(endpoint, &minio.Options{
		Creds:  credentials.NewStaticV4(accessKeyID, secretAccessKey, ""),
		Secure: useSSL,
	})
	if err != nil {
		log.Fatalln(err)
	}

	// Initialize policy condition config.
	policy := minio.NewPostPolicy()

	// Apply upload policy restrictions:
	policy.SetBucket("bucket")
	policy.SetKey("data/somedata")
	policy.SetExpires(time.Now().UTC().AddDate(0, 0, 10))

	// Get the POST form key/value object:
	url, formData, err := minioClient.PresignedPostPolicy(context.Background(), policy)
	if err != nil {
		fmt.Println(err)
		return
	}

	body := &bytes.Buffer{}
	writer := multipart.NewWriter(body)

	for fieldName, fieldValue := range formData {
		part, _ := writer.CreateFormField(fieldName)
		part.Write([]byte(fieldValue))
	}

	// Set the file field using CreateFormField to skip filename assignment
	part, _ := writer.CreateFormField("file")
	part.Write([]byte{30}) // "0"

	writer.Close()

	r, _ := http.NewRequest("POST", url.String(), body)
	r.Header.Add("Content-Type", writer.FormDataContentType())
	client := &http.Client{}
	client.Do(r)
}
```

## Types of changes

## Checklist:
- [ ] Fixes a regression

### DIFF
--- a/cmd/bucket-handlers.go
+++ b/cmd/bucket-handlers.go
@@ -983,7 +983,7 @@ func (api objectAPIHandlers) PostPolicyBucketHandler(w http.ResponseWriter, r *h
 		}
 
 		var b bytes.Buffer
-		if fileName == "" {
+		if name != "file" {
 			if http.CanonicalHeaderKey(name) == http.CanonicalHeaderKey("x-minio-fanout-list") {
 				dec := json.NewDecoder(part)
 
--- a/cmd/post-policy_test.go
+++ b/cmd/post-policy_test.go
@@ -314,6 +314,7 @@ func testPostPolicyBucketHandler(obj ObjectLayer, instanceType string, t TestErr
 		secretKey          string
 		dates              []interface{}
 		policy             string
+		noFilename         bool
 		corruptedBase64    bool
 		corruptedMultipart bool
 	}{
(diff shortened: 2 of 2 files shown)


jiuker:
You should use 
```go
	fw, err := w.CreateFormFile("file", "test.png")
	if err != nil {
		panic(err)
	}
	_, err = io.Copy(fw, file)
	if err != nil {
		panic(err)
	}
```
To create the file.

jiuker (changes requested):
You should use

	fw, err := w.CreateFormFile("file", "test.png")
	if err != nil {
		panic(err)
	}
	_, err = io.Copy(fw, file)
	if err != nil {
		panic(err)
	}
to create the file.
Don't use CreateFormField instead.
@DaniElectra

DaniElectra:
@jiuker I think there is a misunderstanding. The code that I've shown above is *an example* to test the PR, and doesn't represent the real code that the client is running. I have the same issue as #18235 where I can't control the client that makes the request, and you had made PR #18240 to fix this.

But this case still doesn't work for the reason that I had explained (the file reader not getting assigned). In any case, as it was explained on the issue I mention, the filename **isn't mandatory**, so this helps making the server compliant to the spec.

jiuker:
Go, Java, Python or else. All have this function similar method. @DaniElectra 
I think the file and field are different.

jiuker:
@harshavardhana Any advices here?

harshavardhana:
> @harshavardhana Any advices?

This PR is correct @jiuker `Filename()` is not a required field in AWS S3 spec, however `file` is. So the code change is correct.

@DaniElectra please add a test in `TestPostPolicyBucketHandler`

DaniElectra:
I added the test for the filename. I also removed the ignore key for `"File"` since it wasn't really needed.

harshavardhana (approved):

jiuker (approved):

PR #18343, opened 2023-10-30
fix: HeadBucketHandler should check the `HeadBucketAction`

jiuker:

## Types of changes

## Checklist:

### DIFF
--- a/cmd/bucket-handlers.go
+++ b/cmd/bucket-handlers.go
@@ -1546,7 +1546,7 @@ func (api objectAPIHandlers) HeadBucketHandler(w http.ResponseWriter, r *http.Re
 		return
 	}
 
-	if s3Error := checkRequestAuthType(ctx, r, policy.ListBucketAction, bucket, ""); s3Error != ErrNone {
+	if s3Error := checkRequestAuthType(ctx, r, policy.HeadBucketAction, bucket, ""); s3Error != ErrNone {
 		writeErrorResponseHeadersOnly(w, errorCodes.ToAPIErr(s3Error))
 		return
 	}
--- a/cmd/bucket-handlers.go
+++ b/cmd/bucket-handlers.go
@@ -1546,9 +1546,11 @@ func (api objectAPIHandlers) HeadBucketHandler(w http.ResponseWriter, r *http.Re
 		return
 	}
 
-	if s3Error := checkRequestAuthType(ctx, r, policy.HeadBucketAction, bucket, ""); s3Error != ErrNone {
-		writeErrorResponseHeadersOnly(w, errorCodes.ToAPIErr(s3Error))
-		return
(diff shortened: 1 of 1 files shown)


cmd/bucket-handlers.go
```diff
@@ -1546,7 +1546,7 @@ func (api objectAPIHandlers) HeadBucketHandler(w http.ResponseWriter, r *http.Re
 		return
 	}
 
-	if s3Error := checkRequestAuthType(ctx, r, policy.ListBucketAction, bucket, ""); s3Error != ErrNone {
+	if s3Error := checkRequestAuthType(ctx, r, policy.HeadBucketAction, bucket, ""); s3Error != ErrNone {
 		writeErrorResponseHeadersOnly(w, errorCodes.ToAPIErr(s3Error))
 		return
 	}
```
harshavardhana: We did this on purpose @jiuker to keep things simple on the policy side, such that users don't have to explicitly set `s3:HeadBucket` action.

Since it is sufficient enough to have `ListBucketAction` - because no one really cares for just `HeadBucket`, we can, however, pass both in this scenario as actions.

harshavardhana (changes requested):

cmd/bucket-handlers.go
jiuker: > We did this on purpose @jiuker to keep things simple on the policy side, such that users don't have to explicitly set `s3:HeadBucket` action.
> 
> Since it is sufficient enough to have `ListBucketAction` - because no one really cares for just `HeadBucket`, we can, however, pass both in this scenario as actions.

@harshavardhana You mean checking two permissions at the same time?

harshavardhana (approved):

vadmeste:
@jiuker did you try this with AWS S3, or it is written in their spec already ?

harshavardhana:
> @jiuker did you try this with AWS S3, or it is written in their spec already ?

This is per AWS S3 @vadmeste

vadmeste:
> This is per AWS S3 @vadmeste

I cannot find s3:HeadBucket permission in the S3 spec

jiuker:
> > This is per AWS S3 @vadmeste
> 
> I cannot find s3:HeadBucket permission in the S3 spec

@vadmeste I was look for another issue. Found out it doesn't make sense here.

harshavardhana:
> > This is per AWS S3 @vadmeste
> 
> I cannot find s3:HeadBucket permission in the S3 spec

You are right this PR is not needed

```
To use this operation, you must have permissions to perform the s3:ListBucket action. The bucket owner has this permission by default and can grant this permission to others. For more information about permissions, see Permissions Related to Bucket Subresource Operations and Managing Access Permissions to Your Amazon S3 Resources.
```

PR #18344, opened 2023-10-30
Websocket grid benchmarks & improvements

klauspost:
(against `next`)

## Description

Add stream benchmarks. Reduce allocations.

Fixes a few minor issues as well.

## Types of changes

### DIFF
--- a/cmd/format-erasure.go
+++ b/cmd/format-erasure.go
@@ -31,6 +31,7 @@ import (
 	"github.com/minio/minio/internal/color"
 	"github.com/minio/minio/internal/config"
 	"github.com/minio/minio/internal/config/storageclass"
+	"github.com/minio/minio/internal/grid"
 	xioutil "github.com/minio/minio/internal/ioutil"
 	"github.com/minio/minio/internal/logger"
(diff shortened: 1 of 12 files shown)


harshavardhana (approved):

harshavardhana:
This PR caused some issue with `healing functional tests` - server returning 403's 

https://github.com/minio/minio/actions/runs/6695145724/job/18190054122?pr=18344

Canceled this 4hr run

klauspost:
@harshavardhana It is a race, where `ReadAll` can succeed before the grid is connected. I have reproduced it locally.

Since `loadFormatErasure` relies on response from `ReadAll` for quorum, it may be able to perform the call before the grid is up (it may be in a retry backoff).

I inserted a `DiskInfo` call before the ReadAll. This gets rid of the race.

klauspost:
@harshavardhana You may want to consider adding a check that reformats if the initial formatting is not successful.

harshavardhana:
> @harshavardhana You may want to consider adding a check that reformats if the initial formatting is not successful.

Not sure I follow @klauspost what is this about ?

PR #18400, opened 2023-11-06
do not need to continue after we have found the version

harshavardhana:

## Description
do not need to continue after we have found the version

## Motivation and Context
it looks like we must break in this path

## How to test this PR?
Not sure has been around for a while.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/xl-storage-format-v2.go
+++ b/cmd/xl-storage-format-v2.go
@@ -1788,7 +1788,7 @@ func (x xlMetaV2) ToFileInfo(volume, path, versionID string, inclFreeVers, allPa
 			}
 		}
 		if found {
-			continue
+			break
 		}
 
 		// We need a specific version, skip...


cmd/xl-storage-format-v2.go
```diff
@@ -1788,7 +1788,7 @@ func (x xlMetaV2) ToFileInfo(volume, path, versionID string, inclFreeVers, allPa
 			}
 		}
 		if found {
-			continue
+			break
```
vadmeste, line 1791: I think we need to continue, for now. It looks like we are counting the number of free versions in this xl.meta so we can report correctly the number of versions in line 1827 (not sure why NumVersions is used, mostly for ILM max non current versions count)

harshavardhana: Okay then this is a problem @krisis we need to see why this is the case - because this is expensive.

poornas (approved):

cmd/xl-storage-format-v2.go
klauspost: Not really. It is just checking the flag for each entry. Even with 100K versions that is not really going to be slow.

harshavardhana: Okay then, let's leave this for now.

PR #18402, opened 2023-11-07
Add APIs to create and list access keys for LDAP

taran-p:

## Description
Add `AddServiceAccountLDAP`, which verifies that the given user is in LDAP and also gives clearer error if DN is supplied (which is not allowed).

Add `ListAccessKeysLDAP`, which lists STS and service account access keys for a given user, or just one based on supplied arguments. Either short username or full DN can be used.

Add `ListSTSAccounts` to the IAM store.

## Motivation and Context
Add more flexible options for for retrieving information about LDAP accounts.
Corresponding mc commands will be added/modified for the two new APIs.

## Types of changes

## Checklist:


harshavardhana:
PTAL @krisis @donatello

cmd/admin-handlers-idp-ldap.go
```diff
@@ -175,3 +178,359 @@ func (a adminAPIHandlers) AttachDetachPolicyLDAP(w http.ResponseWriter, r *http.
 
 	writeSuccessResponseJSON(w, encryptedData)
 }
+
+// AddServiceAccountLDAP adds a new service account for provided LDAP username or DN
+//
+// PUT /minio/admin/v3/idp/ldap/add-service-account
+func (a adminAPIHandlers) AddServiceAccountLDAP(w http.ResponseWriter, r *http.Request) {
```
donatello, line 186: Most of the validations here are common with the AddServiceAccount handler and they are important - we should factor out the common parts into one or more smaller functions and reuse the code, so that if we fix something, it is automatically applied to the other API. Basically AddServiceAccountLDAP should be mostly identical to AddServiceAccount except for validating that the user exists in LDAP

internal/config/identity/ldap/ldap.go
```diff
@@ -329,3 +330,61 @@ func (l *Config) LookupGroupMemberships(userDistNames []string, userDNToUsername
 
 	return res, nil
 }
+
+// LookupUsername searches for the username and groups of a given DN, like opposite of LookupUserDN
+func (l *Config) LookupUsername(userDN string) (string, []string, error) {
```
donatello: I don't think there's a robust way to do this. We can probably stick with only recording the DN with the service account if the user provided a DN in the API input. If they provide a short username, we lookup the DN and record both. We can remove this function.

In any situation where the short username is given, we first get the full DN and use that for the operation.

donatello (changes requested):

donatello (approved):

donatello (approved):

harshavardhana (commented):
What would this API change bring and what would break if someone is using an older `mc`?

taran-p:
> What would this API change bring and what would break if someone is using an older `mc`?

`AddServiceAccountLDAP` just verifies the account is in LDAP and adds extra clarification for when a DN is supplied, but is otherwise the same as `AddServiceAccount`

`ListAccessKeysLDAP` verifies the account is in LDAP, adds the ability to use short username as well as user DN, and returns STS accounts alongside service accounts.

It shouldn't break anything in old `mc` versions since existing functions should not have changed functionality

taran-p:
Parts of `AddServiceAccount` were moved to a common function with `AddServiceAccountLDAP`, but the functionality should be the same as before

harshavardhana (approved):
Thanks for confirming @taran-p

PR #18411, opened 2023-11-09
use sendfile/splice implementation to perform DMA

harshavardhana:

## Description
use sendfile/splice implementation to perform DMA

## Motivation and Context
Go stdlib already supports sendfile/splice implementations
for

- Linux
- Windows
- *BSD
- Solaris

Along with this change however O_DIRECT for reads() must be
removed as well since we need to use sendfile() implementation

The main reason to add O_DIRECT for reads was to reduce the
chances of page-cache causing OOMs for MinIO, however it would
seem that avoiding buffer copies from user-space to kernel space
this issue is not a problem anymore.

There is no Go based memory allocation required, and neither
the page-cache is referenced back to MinIO. This page-
cache reference is fully owned by kernel at this point, this
essentially should solve the problem of page-cache build up.

With this now we also support SG - when NIC supports Scatter/Gather
https://en.wikipedia.org/wiki/Gather/scatter_(vector_addressing)

## How to test this PR?
This should be transparent to all callers.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/storage-rest-server.go
+++ b/cmd/storage-rest-server.go
@@ -611,6 +611,8 @@ func (s *storageRESTServer) ReadFileStreamHandler(w http.ResponseWriter, r *http
 		return
 	}
 
+	w.Header().Set(xhttp.ContentLength, strconv.Itoa(length))
+
 	rc, err := s.storage.ReadFileStream(r.Context(), volume, filePath, int64(offset), int64(length))
 	if err != nil {
 		s.writeErrorResponse(w, err)
@@ -618,12 +620,28 @@ func (s *storageRESTServer) ReadFileStreamHandler(w http.ResponseWriter, r *http
 	}
 	defer rc.Close()
 
-	w.Header().Set(xhttp.ContentLength, strconv.Itoa(length))
-	if _, err = xioutil.Copy(w, rc); err != nil {
-		if !xnet.IsNetworkOrHostDown(err, true) { // do not need to log disconnected clients
-			logger.LogIf(r.Context(), err)
+	rf, ok := w.(io.ReaderFrom)
+	if ok {
+		// Attempt to use splice/sendfile() optimization, A very specific behavior mentioned below is necessary.
+		// See https://github.com/golang/go/blob/f7c5cbb82087c55aa82081e931e0142783700ce8/src/net/sendfile_linux.go#L20
+		dr, ok := rc.(*xioutil.DeadlineReader)
+		if ok {
+			sr, ok := dr.ReadCloser.(*sendFileReader)
+			if ok {
+				_, err = rf.ReadFrom(sr.Reader)
+				if !xnet.IsNetworkOrHostDown(err, true) { // do not need to log disconnected clients
+					logger.LogIf(r.Context(), err)
+				}
+				if err == nil || !errors.Is(err, xhttp.ErrNotImplemented) {
+					return
+				}
+			}
 		}
-		return
+	} // Fallback to regular copy
+
+	_, err = xioutil.Copy(w, rc)
+	if !xnet.IsNetworkOrHostDown(err, true) { // do not need to log disconnected clients
+		logger.LogIf(r.Context(), err)
 	}
 }
 
--- a/cmd/xl-storage-disk-id-check.go
+++ b/cmd/xl-storage-disk-id-check.go
@@ -421,7 +421,13 @@ func (p *xlStorageDiskIDCheck) ReadFile(ctx context.Context, volume string, path
 	}
 	defer done(&err)
 
-	return p.storage.ReadFile(ctx, volume, path, offset, buf, verifier)
+	w := xioutil.NewDeadlineWorker(diskMaxTimeout)
+	err = w.Run(func() error {
+		n, err = p.storage.ReadFile(ctx, volume, path, offset, buf, verifier)
+		return err
+	})
+
+	return n, err
 }
 
 // Legacy API - does not have any deadlines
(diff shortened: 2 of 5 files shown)


internal/http/response-recorder.go
```diff
@@ -58,6 +59,16 @@ func NewResponseRecorder(w http.ResponseWriter) *ResponseRecorder {
...
 }
 
+// ReadFrom implements support for calling internal io.ReaderFrom implementations
+// returns an error if the underlying ResponseWriter does not implement io.ReaderFrom
+func (lrw *ResponseRecorder) ReadFrom(r io.Reader) (int64, error) {
+	rf, ok := lrw.ResponseWriter.(io.ReaderFrom)
+	if ok {
+		return rf.ReadFrom(r)
+	}
+	return 0, errors.New("not implemented")
```
klauspost: `http.ResponseWriter` is an interface. You can assert this on creation, by doing your own interface:

```Go
// ResponseWriter extends http.ResponseWriter to always include io.ReaderFrom and flushing.
type ResponseWriter interface {
	http.ResponseWriter
	io.ReaderFrom
	http.Flusher
}

// NewResponseRecorder - returns a wrapped response writer to trap
// http status codes for auditing purposes.
func NewResponseRecorder(w ResponseWriter) *ResponseRecorder {
```

If you don't want to always have the flushing, remove it and do a proper interface check.

Edit: It seems like you are using this as a magic value. If it is not possible to assert for all wrappers (not sure why), then at least make the error an exported `ErrReadFromUnavailable` or similar.

cmd/storage-rest-server.go
```diff
@@ -611,19 +611,38 @@ func (s *storageRESTServer) ReadFileStreamHandler(w http.ResponseWriter, r *http
...
 
-	w.Header().Set(xhttp.ContentLength, strconv.Itoa(length))
-	if _, err = xioutil.Copy(w, rc); err != nil {
-		if !xnet.IsNetworkOrHostDown(err, true) { // do not need to log disconnected clients
-			logger.LogIf(r.Context(), err)
+	rf, ok := w.(io.ReaderFrom)
+	if ok {
+		// Boilerplate to use splice/sendfile() optimization, please do not change.
+		// why all this? because of https://github.com/golang/go/blob/f7c5cbb82087c55aa82081e931e0142783700ce8/src/net/sendfile_linux.go#L20
+		// Go has made it super harder to use any of these optimizations appropriately.
```
klauspost:
```suggestion
		// Attempt to use splice/sendfile. A very specific setup is needed.
		// See https://github.com/golang/go/blob/f7c5cbb82087c55aa82081e931e0142783700ce8/src/net/sendfile_linux.go#L20
```

cmd/storage-rest-server.go
```diff
@@ -611,19 +611,38 @@ func (s *storageRESTServer) ReadFileStreamHandler(w http.ResponseWriter, r *http
...
+				if !xnet.IsNetworkOrHostDown(err, true) { // do not need to log disconnected clients
+					logger.LogIf(r.Context(), err)
+				}
+				if err != nil && err.Error() != "not implemented" {
+					return
+				}
+			}
 		}
-		return
+	} // not ReaderFrom nor xioutil.DeadlineReader nor sendFileReader
```
klauspost:
```suggestion
	} // Fall back to regular copy.
```

cmd/storage-rest-server.go
```diff
@@ -611,19 +611,38 @@ func (s *storageRESTServer) ReadFileStreamHandler(w http.ResponseWriter, r *http
...
+		// Go has made it super harder to use any of these optimizations appropriately.
+		dr, ok := rc.(*xioutil.DeadlineReader)
+		if ok {
+			sr, ok := dr.ReadCloser.(*sendFileReader)
+			if ok {
+				_, err = rf.ReadFrom(sr.Reader)
+				if !xnet.IsNetworkOrHostDown(err, true) { // do not need to log disconnected clients
+					logger.LogIf(r.Context(), err)
+				}
+				if err != nil && err.Error() != "not implemented" {
```
klauspost:
```suggestion
				if err == nil || err.Error() != "not implemented" {
```

Seems like you should return here if no error. (add type for check)

cmd/xl-storage-disk-id-check.go
```diff
@@ -410,7 +410,13 @@ func (p *xlStorageDiskIDCheck) ListDir(ctx context.Context, volume, dirPath stri
 	}
 	defer done(&err)
 
-	return p.storage.ListDir(ctx, volume, dirPath, count)
+	w := xioutil.NewDeadlineWorker(diskMaxTimeout)
```
klauspost: Bonus change I see. 2 minutes may lead to some cases where overfilled directories time out.

cmd/xl-storage-disk-id-check.go
```diff
@@ -452,9 +467,16 @@ func (p *xlStorageDiskIDCheck) ReadFileStream(ctx context.Context, volume, path
 	}
 	defer done(&err)
 
-	rc, err := p.storage.ReadFileStream(ctx, volume, path, offset, length)
+	w := xioutil.NewDeadlineWorker(diskMaxTimeout)
+
+	var rc io.ReadCloser
+	err = w.Run(func() error {
+		var ierr error
+		rc, ierr = p.storage.ReadFileStream(ctx, volume, path, offset, length)
```
klauspost, line 469: This is a streaming operation. You cannot apply a global deadline to this. It could be streaming gigabytes of data.

harshavardhana: This doesn't add deadline to the reader @klauspost it is to the return of file descriptor  from this call. 

Since ReadFileStream() `open()` can block this is a deadline worker. Once the `rc` is available this caller simply returns upwards wrapped around deadline reader.

cmd/xl-storage-disk-id-check.go
harshavardhana: Reverted I might have to add this to a layer beneath.

klauspost (approved):
LGTM

poornas (approved):

PR #18414, opened 2023-11-09
Fix leaking offline disks

klauspost:
## Description

`monitorAndConnectEndpoints` will continue to attempt to reconnect offline disks.

Since disks were never closed, a `MarkOffline` would continue to try to check these disks forever.

Close previous disks.

## How to test this PR?

Have disk(s) offline. Observe growth in `github.com/minio/minio/internal/rest.(*Client).MarkOffline.func1` goroutines.

## Types of changes

### DIFF
--- a/cmd/erasure-sets.go
+++ b/cmd/erasure-sets.go
@@ -217,6 +217,10 @@ func (s *erasureSets) connectDisks() {
 				continue
 			}
 		}
+		if cdisk != nil {
+			// Close previous offline disk.
+			cdisk.Close()
+		}
 
 		wg.Add(1)
 		go func(endpoint Endpoint) {


harshavardhana (approved):

mshanmu:
Anytime I see a leak fix, I am a little bit worried. The description seems to suggest once it is marked offline, it will be always offline. Is my understanding correct? 

Is it a regression? If yes, from which version

harshavardhana:
> Anytime I see a leak fix, I am a little bit worried. The description seems to suggest once it is marked offline, it will be always offline. Is my understanding correct?
> 
> Is it a regression? If yes, from which version

An offline is offline until it becomes online that has nothing to do with this fix.

This fix is to ensure that we do not keep creating new MarkOffline routines that never exit. Which happens when a node is offline for a prolonged period like in days to months.

mshanmu:
> > Anytime I see a leak fix, I am a little bit worried. The description seems to suggest once it is marked offline, it will be always offline. Is my understanding correct?
> > Is it a regression? If yes, from which version
> 
> An offline is offline until it becomes online that has nothing to do with this fix.
> 
> This fix is to ensure that we do not keep creating new MarkOffline routines that never exit. Which happens when a node is offline for a prolonged period like in days to months.

Thanks for the explanation, @harshavardhana !! I can sleep well now :)

cmd/erasure-sets.go
```diff
@@ -217,6 +217,10 @@ func (s *erasureSets) connectDisks() {
 				continue
 			}
 		}
+		if cdisk != nil {
+			// Close previous offline disk.
+			cdisk.Close()
```
vadmeste, line 222: It doesn't look like this is an offline disk; diskMap already contains only onlineDisks ; this also means we are closing a live disk here ?

harshavardhana: No we are closing an offline one @vadmeste since the DiskLoc() returns `-1`

vadmeste: well, DiskLoc() will not return -1 for an offline disk, we are not setting set to -1 when a disk goes offline. I am not able to find that anywhere.

Issue #18415, opened 2023-11-09
File corrupted when uploaded via SFTP


mh-ambition-dk:
[MinIO_test.txt](https://github.com/minio/minio/files/13306872/MinIO_test.txt)

When uploading the attached file via SFTP to MinIO it looks like the file is corrupted.

For example here the file is corrupted from line 883 and onwards:

```
881: Hello World -- this is line 881 of the file... Here are some national letters: ÆØÅæøå
882: Hello World -- this is line 882 of the file... Here are some national letters: ÆØÅæøå
883: Heare some national letters: ÆØÅæøå
884: Hello World -- this is line 1059 of the file... Here are some national letters: ÆØÅæøå
885: Hello World -- this is line 1060 of the file... Here are some national letters: ÆØÅæøå
```

The uncorrupted source file look like this:

```
881: Hello World -- this is line 881 of the file... Here are some national letters: ÆØÅæøå
882: Hello World -- this is line 882 of the file... Here are some national letters: ÆØÅæøå
883: Hello World -- this is line 883 of the file... Here are some national letters: ÆØÅæøå
884: Hello World -- this is line 884 of the file... Here are some national letters: ÆØÅæøå
885: Hello World -- this is line 885 of the file... Here are some national letters: ÆØÅæøå
```

The file size is unchanged but data seem to be shuffeled. Its random if it happens or not and where it happens. So you might have to try a few times before it happens.

The problem seems to happen in the SFTP upload process. If I upload the file using MinIO Console I have no problems downloading with SFTP or MinIO Console. But if the file is uploaded using SFTP the file is corrupted both when downloading using MinIO Console and SFTP.

I have tried uploading both using SSH.NET in C# and using FileZilla. The same corruption seem to happen with both.

zveinn:
hello!

I've been trying to replicate your issue using the same file but I cannot. I tried uploading and downloading multiple times but it's always consistent. 

- Can you describe your MinIO setup a little better ? 
- What are the configurations like ? 
- are you connecting to MinIO over wifi or the internet ? 
- have you tried pinging the MinIO server to see if there is instability on the network ? 

I would love to debug this further but I need a little more information, maybe a test script that can help me reproduce ?

mh-ambition-dk:
Thank you for your fast response!

Here are some feedback from my server admin:

•	What are the configurations like?

helm chart:
NAME                            CHART VERSION   APP VERSION
bitnami/minio                   12.9.1          2023.11.1
[...]

PR #18422, opened 2023-11-10
Fixing a file consistency issue on SFTP upload

zveinn:

## Description
Our community raised an [issue with SFTP](https://github.com/minio/minio/issues/18415) where files would randomly get corrupted on upload. The issue is that SFTP demands underlying implementations to offer multiplex support for file segments, while our SFTP to S3 interface was assuming writes would be in sync. 

## Resolution
In order to ensure file consistency, a file segment queue was created which ensures that our S3 store receives all file segments in order. 

## NOTES
The current queue size is set to 1000 segments, if filled the writer returns an error "sftp segment queue is full".
During testing using local interfaces I would regularly run into a queue size of 1-10, so it is probably safe to assume that under certain circumstances the queue size can grow quite a bit bigger than 10. 

## How to test this PR?
The SFTP module relies on manual testing. I recommend uploading 1GB+ files.

1. Create a 1GB file
2. Get the md5 sum of the file
3. Upload the file and download it again
4. compare the md5sums of the downloaded file and the original

## Types of changes
- [ x ] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/cmd/sftp-server-driver.go
+++ b/cmd/sftp-server-driver.go
@@ -201,19 +201,83 @@ func (f *sftpDriver) Fileread(r *sftp.Request) (ra io.ReaderAt, err error) {
 	return obj, nil
 }
 
-type writerAt struct {
-	w  *io.PipeWriter
-	wg *sync.WaitGroup
-}
-
 func (w *writerAt) Close() error {
+	for i := range w.buffers {
+		w.buffers[i] = nil
+	}
 	err := w.w.Close()
 	w.wg.Wait()
 	return err
 }
 
+const bufferLength = 1000
+
+type writerAt struct {
+	w  *io.PipeWriter
+	wg *sync.WaitGroup
+
+	nextOffset int64
+	remaining  int
+	offsets    [bufferLength]int64
+	buffers    [bufferLength][]byte
+	m          sync.Mutex
+}
+
+func (w *writerAt) placeBytesInBuffer(b []byte, offset int64) (err error) {
+	for i := 0; i < bufferLength; i++ {
+		if len(w.buffers[i]) == 0 || w.offsets[i] == 0 {
+			w.offsets[i] = offset
+			w.buffers[i] = make([]byte, len(b))
+			copy(w.buffers[i], b)
+			w.remaining++
+			return nil
+		}
+	}
+	return errors.New("sftp segment queue is full")
+}
+
+func (w *writerAt) flushBuffers() (err error) {
+	for i := 0; i < bufferLength; i++ {
+		if w.offsets[i] == w.nextOffset {
+			n, err := w.w.Write(w.buffers[i])
+			if err != nil {
+				return err
+			}
+			if n != len(w.buffers[i]) {
+				return fmt.Errorf("expected write size %d but wrote %d bytes", len(w.buffers[i]), n)
+			}
+			w.nextOffset += int64(len(w.buffers[i]))
+			w.offsets[i] = 0
+			w.buffers[i] = nil
+			w.remaining--
+			continue
+		}
+	}
+
+	return nil
+}
+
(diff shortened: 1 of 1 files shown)


harshavardhana (approved):

cmd/sftp-server-driver.go
```diff
@@ -201,19 +201,83 @@ func (f *sftpDriver) Fileread(r *sftp.Request) (ra io.ReaderAt, err error) {
 	return obj, nil
 }
 
+func (w *writerAt) Close() error {
+	for i := range w.buffers {
```
klauspost: Check if w.remaining > 0 - return error if so - and use w.w.CloseWithError.

cmd/sftp-server-driver.go
```diff
@@ -201,19 +201,83 @@ func (f *sftpDriver) Fileread(r *sftp.Request) (ra io.ReaderAt, err error) {
...
+				return err
+			}
+			if n != len(w.buffers[i]) {
+				return fmt.Errorf("expected write size %d but wrote %d bytes", len(w.buffers[i]), n)
+			}
+			w.nextOffset += int64(len(w.buffers[i]))
+			w.offsets[i] = 0
+			w.buffers[i] = nil
+			w.remaining--
+			continue
```
klauspost: You can have unflushed buffer at a lower index, that matches new offset. Either restart the loop or recurse.

If not, there is a risk you don't end flushing everything.

klauspost (changes requested):

cmd/sftp-server-driver.go
```diff
@@ -201,19 +201,83 @@ func (f *sftpDriver) Fileread(r *sftp.Request) (ra io.ReaderAt, err error) {
...
+	}
+
+	if w.remaining > 0 {
+		err = w.flushBuffers()
+		if err != nil {
+			return 0, err
+		}
+	}
+	return
 }
```
harshavardhana, line 256: Now that this has gotten a bit complicated @zveinn can you add tests for boundary conditions?

harshavardhana (changes requested):

cmd/sftp-server-driver.go
zveinn: We actually ran into that problem during our first round of testing.

There are two options here, one is to keep iterating the buffer until we have no offsets left that fit the order OR we can trigger a flush anytime there is anything left in the buffer. Both will result in the full file be written. 

I opted to implement it this way because I felt like it was simpler to just flush anytime there was something in the buffer. Also, there is less time between accepting packets from the SFTP clients that way.. kind of like "1 packet in, 1x flush iteration".

cmd/sftp-server-driver.go
zveinn: will do

cmd/sftp-server-driver.go
zveinn: yep, I'll write up some tests for the writer after the weekend (or during if I find some time)

cmd/sftp-server-driver.go
zveinn: Now that I think about it.. my way of doing it is probably going to scale worse then trying to evict as many as possible on every flush. 

I'll have another look at this.

zveinn: you're actually right and I can't believe I missed that.. I'll fix it asap

zveinn:
I took another look at the problem and realized there was no need for a slice buffer. Please re-view the new map method. 

Will make some tests tomorrow  @klauspost @krisis @harshavardhana

krisis (approved):

harshavardhana (approved):

klauspost:
@zveinn This is much better. Did want to ask for a refactor, since we could always add that later - but this is perfect!

PR #18449, opened 2023-11-15
Load bucket configs during the metadata refresh

Praveenrajmani:

## Description

Currently, If the bucket configs are not loaded initially (bootstrap) due to any failure (for example, quorum loss), no retries are made to set the bucket configs back. As a result, the bucket notification configs will not be loaded on the respective nodes and any subsequent s3 APIs to the nodes will not be sending any events to the targets.

This patch takes care of loading the bucket configs of failed buckets during the periodic refresh. This makes sure the event notifiers and remote bucket targets are properly initialized.

## Motivation and Context

To retry the failures on bucket config loading, and make the event notification system and bucket replication system resilient on bootstrap failures.

## How to test this PR?

- Build an image with the following diff
- 
```diff
diff --git a/cmd/bucket-metadata-sys.go b/cmd/bucket-metadata-sys.go
index 344e8a598..ecaaae924 100644
--- a/cmd/bucket-metadata-sys.go
+++ b/cmd/bucket-metadata-sys.go
@@ -454,6 +454,8 @@ func (sys *BucketMetadataSys) loadBucketMetadata(ctx context.Context, bucket Buc
 func (sys *BucketMetadataSys) concurrentLoad(ctx context.Context, buckets []BucketInfo) {
        g := errgroup.WithNErrs(len(buckets))
        bucketMetas := make([]BucketMetadata, len(buckets))
+       fmt.Println("\n[DEBUG] Sleeping...")
+       time.Sleep(3 * time.Minute)
        for index := range buckets {
                index := index
                g.Go(func() error {

```
- Setup a multi node MinIO with bucket notifications configured. You can use https://webhook.site for this. 
- Pick a node for testing, when this node comes to the "[DEBUG] Sleeping..." state, kill the other MinIO servers - this is to mimic the quorum loss on this node during the bucket loading time.
- Wait for few mins (say 10-20 mins) for the pending calls to timeout
- Start the other MinIO servers
- Set an alias pointing to the targeted node; Try to upload and see if you have any events triggered. There won't be any.
- Apply the fix and repeat the same steps.
- The config will be eventually loaded on periodic refresh. The current periodic refresh is set to 15mins. So, you have to wait for 15minx max to get the configs loaded successfully.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/bucket-metadata-sys.go
+++ b/cmd/bucket-metadata-sys.go
@@ -437,21 +437,8 @@ func (sys *BucketMetadataSys) Init(ctx context.Context, buckets []BucketInfo, ob
 	return nil
 }
 
-func (sys *BucketMetadataSys) loadBucketMetadata(ctx context.Context, bucket BucketInfo) error {
-	meta, err := loadBucketMetadata(ctx, sys.objAPI, bucket.Name)
-	if err != nil {
-		return err
-	}
-
-	sys.Lock()
-	sys.metadataMap[bucket.Name] = meta
-	sys.Unlock()
-
-	return nil
-}
-
 // concurrently load bucket metadata to speed up loading bucket metadata.
-func (sys *BucketMetadataSys) concurrentLoad(ctx context.Context, buckets []BucketInfo) {
+func (sys *BucketMetadataSys) concurrentLoad(ctx context.Context, buckets []BucketInfo, failedBuckets map[string]struct{}) {
 	g := errgroup.WithNErrs(len(buckets))
 	bucketMetas := make([]BucketMetadata, len(buckets))
 	for index := range buckets {
@@ -491,6 +478,10 @@ func (sys *BucketMetadataSys) concurrentLoad(ctx context.Context, buckets []Buck
 
 	for i, meta := range bucketMetas {
 		if errs[i] != nil {
+			if failedBuckets == nil {
+				failedBuckets = make(map[string]struct{})
+			}
+			failedBuckets[buckets[i].Name] = struct{}{}
 			continue
 		}
 		globalEventNotifier.set(buckets[i], meta)   // set notification targets
@@ -498,7 +489,7 @@ func (sys *BucketMetadataSys) concurrentLoad(ctx context.Context, buckets []Buck
 	}
 }
 
-func (sys *BucketMetadataSys) refreshBucketsMetadataLoop(ctx context.Context) {
+func (sys *BucketMetadataSys) refreshBucketsMetadataLoop(ctx context.Context, failedBuckets map[string]struct{}) {
 	const bucketMetadataRefresh = 15 * time.Minute
 
 	sleeper := newDynamicSleeper(2, 150*time.Millisecond, false)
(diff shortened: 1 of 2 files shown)


harshavardhana (commented):
Have you checked if this affects other configuration that are related to bucket metadata?

Praveenrajmani:
@harshavardhana There is no problem with bucket metadata here - it gets loaded fine during the periodic refresh. Just that we missed setting the event rules map and bucket targets map for these failed buckets.

Will also test the bucket replication target settings during such scenarios.

harshavardhana:
> @harshavardhana There is no problem with bucket metadata here - it gets loaded fine during the periodic refresh. Just that we missed setting the event rules map and bucket targets map for these failed buckets.
> 
> Will also test the bucket replication target settings during such scenarios.

Okay thanks @Praveenrajmani

Praveenrajmani:
> Will also test the bucket replication target settings during such scenarios.

Tested this. It works as expected.

Also, the remote bucket targets get lazily refreshed (https://github.com/minio/minio/blob/master/cmd/bucket-targets.go#L503)  when an object is set to be replicated. So, bucket target functions are fine and will not be affected by this issue.

harshavardhana (approved):

cmd/bucket-metadata-sys.go
```diff
@@ -513,7 +504,7 @@ func (sys *BucketMetadataSys) refreshBucketsMetadataLoop(ctx context.Context) {
 			buckets, err := sys.objAPI.ListBuckets(ctx, BucketOptions{})
 			if err != nil {
 				logger.LogIf(ctx, err)
-				continue
+				break
```
vadmeste, line 507: 👍

cmd/bucket-metadata-sys.go
```diff
@@ -525,38 +516,49 @@ func (sys *BucketMetadataSys) refreshBucketsMetadataLoop(ctx context.Context) {
...
 					continue
 				}
 
+				sys.Lock()
+				sys.metadataMap[buckets[i].Name] = meta
+				sys.Unlock()
+
+				// Initialize the failed buckets
+				if _, ok := failedBuckets[buckets[i].Name]; ok {
+					globalEventNotifier.set(buckets[i], meta)
```
vadmeste, line 535: If you ask me, we should also call globalEventNotifer.set() and globalBucktTargetSys.set() even if this is not a failed bucket; because let's say one node did not receive an update from a peer about this configuration change; this will still be not updated. 

I think we need to add a timestamp in all configurations so we know when we update and when not; but not in this PR; in another one instead.

vadmeste (approved):
LGTM

cmd/bucket-metadata-sys.go
harshavardhana: @Praveenrajmani will merge this PR please send an another one.

Praveenrajmani: These `set` calls doesn't do any peer rest calls. It is just a local in-memory update.

Issue #18315, opened 2023-10-25
make-user job in community helm chart broken


mundry:
Hi,

since changing the base image of the `mc` container image from ubi-minimal to ubi-micro ([minio/mc#4725](https://github.com/minio/mc/pull/4725)), the make-user step in the community helm chart is broken when using an existing secret for credentials, as [the step uses `sed`](https://github.com/minio/minio/blob/0594d372309510c5a37b09f6134735398d1efaa1/helm/minio/templates/_helper_create_user.txt#L100), which is not included in ubi-micro.

## Expected Behavior
Running the community helm chart with values for an user and credentials from an existing secret defined to run smoothly and create the user according to the given values.

## Current Behavior
The create-user job fails, as `sed` is missing from the `mc` container image. Hence no user is created. The deployment finishes with an error.
```
Connecting to MinIO server: http://minio:9000
Added `myminio` successfully.
/config/add-user: line 89: sed: command not found
```

## Possible Solution
I see a few possible options to resolve this.
* Rolling back the base image to use ubi-minimal again
* Installing sed in the mc container image, based on ubi-micro
* Refactoring the `sed` line in the make-user job to use an available command.
* Dropping the `sed` line altogether. (I'm not sure on the repercussions of this)

## Steps to Reproduce (for bugs)
Create a minio deployment using the community helm chart and specifying the respective values for an user to be created, with credentials from an existing secret.

## Context
Setting up the minio server worked fine. However, the make-user job failed. not creating an user as specified as a result. 
Other jobs part of the chart remain unaffected, as the minio server deployment is executed as expected.

## Regression
Yes
[minio/mc#4725](https://github.com/minio/mc/pull/4725)

## Your Environment
minio: RELEASE.2023-10-25T06-33-25Z
mc: RELEASE.2023-10-24T05-18-28Z

PKizzle:
As a temporary solution I have replaced `sed -i '$a\' $MINIO_ACCESSKEY_SECRETKEY_TMP` with `echo >> $MINIO_ACCESSKEY_SECRETKEY_TMP`. This will always append a newline at the end of the file instead of only if none existed beforehand but with the very limited tools available in ubi-micro that seems to be the best solution at the moment.

PR #18459, opened 2023-11-16
Fix issue [#18315], resolve the absence of the sed command in ubi-mic…

opencmit2:
…ro by replacing it with echo.

## Description

resolve the absence of the sed command in ubi-micro by replacing it with echo.

## Motivation and Context

Fix issue [#18315].

## Types of changes

## Checklist:


harshavardhana (approved):

helm/minio/templates/_helper_create_user.txt
```diff
@@ -97,7 +97,7 @@ echo {{ tpl .accessKey $global }} > $MINIO_ACCESSKEY_SECRETKEY_TMP
 {{- if .existingSecret }}
 cat /config/secrets/{{ tpl .existingSecret $global }}/{{ tpl .existingSecretKey $global }} >> $MINIO_ACCESSKEY_SECRETKEY_TMP
 # Add a new line if it doesn't exist
-sed -i '$a\' $MINIO_ACCESSKEY_SECRETKEY_TMP
+echo >> $MINIO_ACCESSKEY_SECRETKEY_TMP
```
harshavardhana, line 100: Wait how does it fix usage of sed?

harshavardhana (changes requested):

helm/minio/templates/_helper_create_user.txt
opencmit2: Here, only the method of command replacement is used to make it functional.
I apologize for the misunderstanding.Due to the fact that the quay.io/minio/mc:RELEASE.2023-09-29T16-41-22Z image does not support the sed command, can also use the following two methods.
1、Roll back the base image to use ubi-minimal again.
2、Install sed in the mc container image, based on ubi-micro.

opencmit2: The purpose of the sed command is to insert a blank line at the end of the target file. If the last line is absent in the target file, the command will create a new line at the file's conclusion.

Alternatively, you can achieve the same result using the echo command along with the >> operator. This approach effectively adds a blank line to the end of the file.

harshavardhana: and why do we need that blankline?

WVenus: I used echo just to address the issue of the sed command not being available in the image, causing it to be unusable.
image

For the above repair method, the echo command is indeed not necessary.

opencmit2: I used echo just to address the issue of the sed command not being available in the image, causing it to be unusable.

For the above repair method, the echo command is indeed not necessary.

jiuker: @harshavardhana 
for this 
```shell
cat /config/secrets/{{ tpl .existingSecret $global }}/{{ tpl .existingSecretKey $global }} >> $MINIO_ACCESSKEY_SECRETKEY_TMP
# Add a new line if it doesn't exist
sed -i '$a\' $MINIO_ACCESSKEY_SECRETKEY_TMP
```
before the sed action.
cat will add the environment like
```env
export A=B
```
if we not do this, we will get 
```env
export A=BOthers
```
after:
```env
export A=B
Others
```
I think the fix is correct. cc

jiuker (approved):

helm/minio/templates/_helper_create_user.txt
vadmeste: @jiuker as per my testing `sed -i '$a\'` doesn't have a real effect.

harshavardhana (approved):

helm/minio/templates/_helper_create_user.txt
jiuker: > as per my testing `sed -i '$a\'` doesn't have a real effect.

It depends on whether your secret configuration comes with a new line or not @vadmeste

PR #18461, opened 2023-11-16
Websocket grid functionality

klauspost:
## Description

`next` branch rebased with master.

See API description below

## How to test this PR?

Regular testing.

## Types of changes

## Checklist:

# MinIO Grid

The MinIO Grid is a package that provides two-way communication between servers.
It uses a single two-way connection to send and receive messages between servers.

It includes built in muxing of concurrent requests as well as congestion handling for streams.

Requests can be "Single Payload" or "Streamed".

Use the MinIO Grid for:

* Small, frequent requests with low latency requirements.
* Long-running requests with small/medium payloads.

Do *not* use the MinIO Grid for:

* Large payloads.

Only a single connection is ever made between two servers.
Likely this means that this connection will not be able to saturate network bandwidth.
Therefore, using this for large payloads will likely be slower than using a separate connection,
and other connections will be blocked while the large payload is being sent.

## Handlers & Routes

Handlers have a predefined Handler ID. 
In addition, there can be several *static* subroutes used to differentiate between different handlers of the same ID.
A subroute on a client must match a subroute on the server. So routes cannot be used for dynamic routing, unlike HTTP.

Handlers should remain backwards compatible. If a breaking API change is required, a new handler ID should be created.

## Setup & Configuration

A **Manager** is used to manage all incoming and outgoing connections to a server.

On startup all remote servers must be specified. 
From that individual connections will be spawned to each remote server, 
or incoming requests will be hooked up to the appropriate connection.

To get a connection to a specific server, use `Manager.Connection(host)` to get a connection to the specified host.
From this connection individual requests can be made.

Each handler, with optional subroutes can be registered with the manager using 
`Manager.RegisterXHandler(handlerID, handler, subroutes...)`.

A `Handler()` function provides an HTTP handler, which should be hooked up to the appropriate route on the server.

On startup, the manager will start connecting to remotes and also starts listening for incoming connections.
Until a connection is established, all outgoing requests will return `ErrDisconnected`.

# Usage

## Single Payload Requests

Single payload requests are requests and responses that are sent in a single message.
In essence, they are `[]byte` -> `[]byte, error` functions.

It is not possible to return *both* an error and a response.

Handlers are registered on the manager using `(*Manager).RegisterSingleHandler(id HandlerID, h SingleHandlerFn, subroute ...string)`.

The server handler function has this signature: `type SingleHandlerFn func(payload []byte) ([]byte, *RemoteErr)`.

Sample handler:
```go
    handler :=  func(payload []byte) ([]byte, *grid.RemoteErr) {
        // Do something with payload
        return []byte("response"), nil
    }

    err := manager.RegisterSingleHandler(grid.HandlerDiskInfo, handler)
```

Sample call:
```go
    // Get a connection to the remote host
    conn := manager.Connection(host)
	
    payload := []byte("request")
    response, err := conn.SingleRequest(ctx, grid.HandlerDiskInfo, payload)
```

If the error type is `*RemoteErr`, then the error was returned by the remote server. Otherwise it is a local error.

Context timeouts are propagated, and a default timeout of 1 minute is added if none is specified.

There is no cancellation propagation for single payload requests.
When the context is canceled, the request will return at once with an appropriate error. 
However, the remote call will not see the cancellation - as can be seen from the 'missing' context on the handler.
The result will be discarded.

### Typed handlers

Typed handlers are handlers that have a specific type for the request and response payloads.
These must provide `msgp` serialization and deserialization.

In the examples we use a `MSS` type, which is a `map[string]string` that is `msgp` serializable.

```go
    handler := func(request *grid.MSS) (*grid.MSS, *grid.RemoteErr) {
        fmt.Println("Got request with field", request["myfield"])
        // Do something with payload
        return NewMSSWith(map[string]string{"result": "ok"}), nil
    }
	
    // Create a typed handler.
    // Due to current generics limitations, a constructor of the empty type must be provided.
    instance := grid.NewSingleHandler[*grid.MSS, *grid.MSS](h, grid.NewMSS, grid.NewMSS)
	
    // Register the handler on the manager
    instance.Register(manager, handler)
	
    // The typed instance is also used for calls
    conn := manager.Connection("host")
    resp, err := instance.Call(ctx, conn, grid.NewMSSWith(map[string]string{"myfield": "myvalue"}))
    if err == nil {
        fmt.Println("Got response with field", resp["result"])
    }
```

The wrapper will handle all serialization and de-seralization of the request and response,
and furthermore provides re-use of the structs used for the request and response.

Note that Responses sent for serialization are automatically reused for similar requests.
If the response contains shared data it will cause issues, since each unique response is reused.
To disable this behavior, use `(SingleHandler).WithSharedResponse()` to disable it.

## Streaming Requests

Streams consists of an initial request with payload and allows for full two-way communication between the client and server.

The handler function has this signature.

Sample handler:
```go
    handler :=  func(ctx context.Context, payload []byte, in <-chan []byte, out chan<- []byte) *RemoteErr {
        fmt.Println("Got request with initial payload", p, "from", GetCaller(ctx context.Context))
        fmt.Println("Subroute:", GetSubroute(ctx))
        for {
            select {
            case <-ctx.Done():
                return nil
            case req, ok := <-in:
				if !ok {
					break
                }           
                // Do something with payload
                out <- []byte("response")

                // Return the request for reuse
                grid.PutByteBuffer(req)
            }
        }
        // out is closed by the caller and should never be closed by the handler.
        return nil
    }

    err := manager.RegisterStreamingHandler(grid.HandlerDiskInfo, StreamHandler{
        Handle: handler,
        Subroute: "asubroute",
        OutCapacity: 1,
        InCapacity: 1,
    })
```

Sample call:
```go
    // Get a connection to the remote host
    conn := manager.Connection(host).Subroute("asubroute")
	
    payload := []byte("request")
    stream, err := conn.NewStream(ctx, grid.HandlerDiskInfo, payload)
	if err != nil {
        return err
    }
    // Read results from the stream
    err = stream.Results(func(result []byte) error {
        fmt.Println("Got result", string(result))

        // Return the response for reuse
        grid.PutByteBuffer(result)
        return nil
    })
```

Context cancellation and timeouts are propagated to the handler. 
The client does not wait for the remote handler to finish before returning.
Returning any error will also cancel the stream remotely.

CAREFUL: When utilizing two-way communication, it is important to ensure that the remote handler is not blocked on a send.
If the remote handler is blocked on a send, and the client is trying to send without the remote receiving, 
the operation would become deadlocked if the channels are full.

### Typed handlers

Typed handlers are handlers that have a specific type for the request and response payloads.

```go
    // Create a typed handler.
    handler := func(ctx context.Context, p *Payload, in <-chan *Req, out chan<- *Resp) *RemoteErr {
        fmt.Println("Got request with initial payload", p, "from", GetCaller(ctx context.Context))
		fmt.Println("Subroute:", GetSubroute(ctx))
        for {
            select {
            case <-ctx.Done():
                return nil
            case req, ok := <-in:
                if !ok {
                    break
                }
                fmt.Println("Got request", in)
                // Do something with payload
                out <- Resp{"response"}
            }
            // out is closed by the caller and should never be closed by the handler.
            return nil
    }
	
    // Create a typed handler.
    // Due to current generics limitations, a constructor of the empty type must be provided.
    instance := grid.NewStream[*Payload, *Req, *Resp](h, newPayload, newReq, newResp)
	
    // Tweakable options
    instance.WithPayload = true // default true when newPayload != nil
    instance.OutCapacity = 1    // default
    instance.InCapacity = 1     // default true when newReq != nil
	
    // Register the handler on the manager
    instance.Register(manager, handler, "asubroute")
	
    // The typed instance is also used for calls
    conn := manager.Connection("host").Subroute("asubroute")
    stream, err := instance.Call(ctx, conn, &Payload{"request payload"})
    if err != nil { ... }
	
    // Read results from the stream
    err = stream.Results(func(resp *Resp) error {
        fmt.Println("Got result", resp)
        // Return the response for reuse
		instance.PutResponse(resp)
        return nil
    })
```

There are handlers for requests with:
 * No input stream: `RegisterNoInput`.
 * No initial payload: `RegisterNoPayload`.

Note that Responses sent for serialization are automatically reused for similar requests.
If the response contains shared data it will cause issues, since each unique response is reused.
To disable this behavior, use `(StreamTypeHandler).WithSharedResponse()` to disable it.


cmd/grid.go
```diff
@@ -0,0 +1,83 @@
...
+	var local string
+	for _, ep := range eps {
+		for _, endpoint := range ep.Endpoints {
+			u := endpoint.GridHost()
+			if seenHosts.Contains(u) {
+				continue
+			}
+			seenHosts.Add(u)
+
+			// Only proceed for remote endpoints.
```
vadmeste: It doesn't feel this comment is correct

cmd/grid.go
```diff
@@ -0,0 +1,83 @@
...
+		Dialer:       grid.ContextDialer(xhttp.DialContextWithLookupHost(lookupHost, xhttp.NewInternodeDialContext(rest.DefaultTimeout, globalTCPOptions))),
+		Local:        local,
+		Hosts:        hosts,
+		AddAuth:      newCachedAuthToken(),
+		AuthRequest:  storageServerRequestValidate,
+		BlockConnect: globalGridStart,
+		TLSConfig: &tls.Config{
+			RootCAs:          globalRootCAs,
+			CipherSuites:     fips.TLSCiphers(),
+			CurvePreferences: fips.TLSCurveIDs(),
```
vadmeste, line 71: fips is not always enabled

cmd/lock-rest-client.go
```diff
@@ -18,121 +18,86 @@
...
-	return client.restCall(ctx, lockRESTMethodUnlock, args)
+func (c *lockRESTClient) Unlock(ctx context.Context, args dsync.LockArgs) (reply bool, err error) {
+	return c.call(ctx, lockRPCUnlock, &args)
 }
 
 // ForceUnlock calls force unlock handler to forcibly unlock an active lock.
-func (client *lockRESTClient) ForceUnlock(ctx context.Context, args dsync.LockArgs) (reply bool, err error) {
-	return client.restCall(ctx, lockRESTMethodForceUnlock, args)
+func (c *lockRESTClient) ForceUnlock(ctx context.Context, args dsync.LockArgs) (reply bool, err error) {
+	return c.call(ctx, lockRPCRLock, &args)
```
vadmeste: this doesn't look correct lockRPCUnlock

cmd/lock-rest-server.go
```diff
@@ -19,203 +19,151 @@ package cmd
...
-	case errLockNotFound:
-		statusCode = http.StatusNotFound
-	}
-	w.WriteHeader(statusCode)
-	w.Write([]byte(err.Error()))
-}
-
 // IsValid - To authenticate and verify the time difference.
 func (l *lockRESTServer) IsValid(w http.ResponseWriter, r *http.Request) bool {
-	if l.ll == nil {
```
vadmeste, line 62: IsValid() is always return true here, is there auth somewhere else ?

cmd/storage-rest-client.go
```diff
@@ -856,7 +802,10 @@ func (client *storageRESTClient) CleanAbandonedData(ctx context.Context, volume
 		return err
 	}
 	defer xhttp.DrainBody(respBody)
-	_, err = waitForHTTPResponse(respBody)
+	respReader, err := waitForHTTPResponse(respBody)
+	if err == nil {
+		io.Copy(io.Discard, respReader)
```
vadmeste: we have L804, isn't that the same thing ?

internal/grid/manager.go
```diff
@@ -0,0 +1,322 @@
...
+	m.handlers.subSingle[subID] = h
+	// Copy so clients can also pick it up for other subpaths.
+	m.handlers.subSingle[makeZeroSubHandlerID(id)] = h
+	return nil
+}
+
+/*
+// RegisterStateless will register a stateless handler that serves
+// []byte -> stream of ([]byte, error) requests.
+func (m *Manager) RegisterStateless(id HandlerID, h StatelessHandler) error {
```
vadmeste, line 261: You probably wanted to remove this

internal/grid/manager.go
```diff
@@ -0,0 +1,322 @@
...
+				debug.PrintStack()
+				fmt.Printf("grid: panic: %v\n", r)
+				w.WriteHeader(http.StatusInternalServerError)
+			}
+		}()
+		if debugPrint {
+			fmt.Printf("grid: Got a %s request for: %v\n", req.Method, req.URL)
+		}
+		if err := m.authRequest(req); err != nil {
+			fmt.Printf("grid: auth error: %v\n", err)
```
vadmeste: logger.LogIf() instead ?

internal/grid/muxserver.go
```diff
@@ -0,0 +1,304 @@
...
+		RecvSeq:   msg.Seq + 1,
+		SendSeq:   msg.Seq,
+		ctx:       ctx,
+		cancel:    cancel,
+		parent:    c,
+		LastPing:  time.Now().Unix(),
+		BaseFlags: c.baseFlags,
+	}
+	go func() {
+		// TODO: Handle
```
vadmeste, line 67: Not needed anymore ?

internal/grid/muxserver.go
```diff
@@ -0,0 +1,304 @@
...
+	if !m.checkSeq(seq) {
+		msg := fmt.Sprintf("receive sequence number mismatch. want %d, got %d", m.RecvSeq, seq)
+		return pongMsg{Err: &msg}
+	}
+	select {
+	case <-m.ctx.Done():
+		err := context.Cause(m.ctx).Error()
+		return pongMsg{Err: &err}
+	default:
+		atomic.StoreInt64(&m.LastPing, time.Now().Unix())
```
vadmeste, line 290: maybe, let's use UnixNano() intead; I don't see any benefit with discarding this information to a second

internal/grid/handlers.go
```diff
@@ -0,0 +1,697 @@
...
+
+func makeSubHandlerID(id HandlerID, subRoute string) subHandlerID {
+	b := subHandlerID(sha256.Sum256([]byte(subRoute)))
+	b[0] = byte(id)
+	b[1] = 0 // Reserved
+	return b
+}
+
+func (s subHandlerID) withHandler(id HandlerID) subHandlerID {
+	s[1] = 0 // Reserved
```
vadmeste: In line 204, we set the id to the index zero; not sure if this is a typo or not

vadmeste (commented):
Mostly questions, two possible bugs; I will need a second more deeper round of review

cmd/grid.go
```diff
@@ -0,0 +1,83 @@
...
+		// Record incoming and outgoing bytes.
+		Incoming: globalConnStats.incInternodeInputBytes,
+		Outgoing: globalConnStats.incInternodeOutputBytes,
+		TraceTo:  globalTrace,
+	})
+	if err != nil {
+		return err
+	}
+	globalGrid.Store(g)
+	return err
```
shtripat: non nil err is returned at line#79, so here we can return nil safely I feel.

cmd/handler-utils.go
```diff
@@ -389,12 +389,6 @@ func errorResponseHandler(w http.ResponseWriter, r *http.Request) {
 			Description:    desc,
 			HTTPStatusCode: http.StatusUpgradeRequired,
 		}, r.URL)
-	case strings.HasPrefix(r.URL.Path, lockRESTPrefix):
```
shtripat, line 392: not needed anymore?

cmd/lock-rest-client.go
```diff
@@ -18,121 +18,86 @@
...
-func (client *lockRESTClient) Close() error {
-	client.restClient.Close()
+func (c *lockRESTClient) Close() error {
 	return nil
 }
 
-// restCall makes a call to the lock REST server.
-func (client *lockRESTClient) restCall(ctx context.Context, call string, args dsync.LockArgs) (reply bool, err error) {
-	argsBytes, err := args.MarshalMsg(metaDataPoolGet()[:0])
+// Close - marks the client as closed.
```
shtripat: Comment doesn't match the function

cmd/lock-rest-client_test.go
```diff
@@ -26,14 +26,27 @@ import (
 
 // Tests lock rpc client.
 func TestLockRESTlient(t *testing.T) {
-	endpoint, err := NewEndpoint("http://localhost:9000")
+	// These should not be conbectable.
```
shtripat: spell check `conbectable`

cmd/metacache-walk.go
```diff
@@ -57,6 +54,10 @@ type WalkDirOptions struct {
 
 	// Limit the number of returned objects if > 0.
 	Limit int
+
+	// DiskID contains the disk ID of the disk.
```
shtripat, line 58: `DiskID contains the ID of the disk` ?

cmd/storage-rest-server.go
```diff
@@ -165,93 +167,85 @@ func (s *storageRESTServer) IsValid(w http.ResponseWriter, r *http.Request) bool
 	return true
 }
 
+// IsValid - To authenticate and check if the disk-id in the request corresponds to the underlying disk.
```
shtripat: Comment doesnt match the function

cmd/storage-rest-server.go
```diff
@@ -410,55 +416,35 @@ func (s *storageRESTServer) ReadVersionHandler(w http.ResponseWriter, r *http.Re
...
-	}
+var storageWriteMetadataHandler = grid.NewSingleHandler[*MetadataHandlerParams, grid.NoPayload](grid.HandlerWriteMetadata, func() *MetadataHandlerParams {
+	return &MetadataHandlerParams{}
+}, grid.NewNoPayload)
 
-	var fi FileInfo
-	if err := msgp.Decode(r.Body, &fi); err != nil {
-		s.writeErrorResponse(w, err)
-		return
+// WriteMetadata write new updated metadata.
```
shtripat: Function name in comment doesn't match the actual function name

cmd/storage-rest-server.go
```diff
@@ -410,55 +416,35 @@ func (s *storageRESTServer) ReadVersionHandler(w http.ResponseWriter, r *http.Re
...
-	}
+var storageUpdateMetadataHandler = grid.NewSingleHandler[*MetadataHandlerParams, grid.NoPayload](grid.HandlerUpdateMetadata, func() *MetadataHandlerParams {
+	return &MetadataHandlerParams{}
+}, grid.NewNoPayload)
 
-	var fi FileInfo
-	if err := msgp.Decode(r.Body, &fi); err != nil {
-		s.writeErrorResponse(w, err)
-		return
+// UpdateMetadata update new updated metadata.
```
shtripat: same

cmd/storage-rest-server.go
```diff
@@ -1350,7 +1311,7 @@ func (s *storageRESTServer) ReadMultiple(w http.ResponseWriter, r *http.Request)
 }
 
 // registerStorageRPCRouter - register storage rpc router.
```
shtripat: correct comment inline with actual function name and its working

cmd/xl-storage-format-v2_test.go
```diff
@@ -115,7 +115,7 @@ func TestXLV2FormatData(t *testing.T) {
 
 	serialized, err := xl.AppendTo(nil)
 	failOnErr(err)
-	// Roundtrip data
+	// roundtrip data
```
shtripat: needed?

internal/grid/README.md
```diff
@@ -0,0 +1,251 @@
...
+```go
+    handler :=  func(ctx context.Context, payload []byte, in <-chan []byte, out chan<- []byte) *RemoteErr {
+        fmt.Println("Got request with initial payload", p, "from", GetCaller(ctx context.Context))
+        fmt.Println("Subroute:", GetSubroute(ctx))
+        for {
+            select {
+            case <-ctx.Done():
+                return nil
+            case req, ok := <-in:
+				if !ok {
```
shtripat: minor one to align this code sample properly

cmd/grid.go
klauspost: No, but the package provides fallback AFAICT. That should not be a concern here.

cmd/lock-rest-server.go
klauspost: There is auth when the websocket connection is created. After that no auth is done for individual requests.

klauspost: In fact this function should just be removed now.

internal/grid/manager.go
klauspost: Yeah. Using LogOnce...

internal/grid/manager.go
klauspost: No. This is to be enabled later for stateless connections.

internal/grid/muxserver.go
klauspost: Needed later.

internal/grid/muxserver.go
klauspost: Sub-second precision is not needed. But I see that it isn't actively used, so I will add some protection against leaks if the caller forgets about it.

internal/grid/connection.go
```diff
@@ -0,0 +1,1604 @@
...
+var ErrDone = errors.New("done for now")
+
+var ErrRemoteRestart = errors.New("remote restarted")
+
+
+// Stateless connects to the remote handler and return all packets sent back.
+// If the remote is restarted will return ErrRemoteRestart.
+// If nil will be returned remote call sent EOF or ErrDone is returned by the callback.
+// If ErrDone is returned on cb nil will be returned.
+func (c *Connection) Stateless(ctx context.Context, h HandlerID, req []byte, cb func([]byte) error) error {
```
shtripat, line 496: needed?

cmd/grid.go
klauspost: Yeah. Copy+paste remain.

internal/grid/handlers.go
klauspost: They are just swapped here. I will make index 0 first, so it is clearer.

cmd/storage-rest-client.go
klauspost: Weird. don't remember this change - maybe a merge artifact. Yeah - I don't see why it would be needed. Removing.

internal/grid/handlers.go
```diff
@@ -0,0 +1,697 @@
...
+
+// putRequest will accept a request for reuse.
+// This is not exported, since it shouldn't be needed.
+func (h *SingleHandler[Req, Resp]) putRequest(r Req) {
+	if r != h.nilReq {
+		h.reqPool.Put(r)
+	}
+}
+
+// NewRequest creates a new response.
```
shtripat: creates a new request, right? comment doesnt look aligned.

internal/grid/handlers.go
```diff
@@ -0,0 +1,697 @@
...
+		return payload, nil
+	}, subroute...)
+}
+
+// Requester is able to send requests to a remote.
+type Requester interface {
+	Request(ctx context.Context, h HandlerID, req []byte) ([]byte, error)
+}
+
+// Call the remove with the request and return the response.
```
shtripat: `%s/remove/remote/g`

internal/grid/handlers.go
```diff
@@ -0,0 +1,697 @@
...
+
+// PutRequest will accept a request for reuse.
+// These should be returned by the handler.
+func (h *StreamTypeHandler[Payload, Req, Resp]) PutRequest(r Req) {
+	if r != h.nilReq {
+		h.reqPool.Put(r)
+	}
+}
+
+// PutResponse will accept a request for reuse.
```
shtripat: accept a response , right?

internal/grid/handlers.go
```diff
@@ -0,0 +1,697 @@
...
+			}()
+			rErr := handle(ctx, plT, inT, outT)
+			close(outT)
+			<-outDone
+			return rErr
+		}, OutCapacity: h.OutCapacity, InCapacity: h.InCapacity, Subroute: strings.Join(subroute, "/"),
+	})
+}
+
+// TypedSteam is a stream with specific types.
```
shtripat: `%s/TypedSteam/TypedStream/g`

internal/grid/manager.go
```diff
@@ -0,0 +1,320 @@
...
+	// Bumping this should only be done when overall,
+	// incompatible changes are made, not when adding a new handler
+	// or changing an existing handler.
+	apiVersion = "v1"
+
+	// RoutePath is the remote path to connect to.
+	RoutePath = "/minio/grid/" + apiVersion
+)
+
+// Manager will
```
shtripat: Incomplete comment

internal/grid/manager.go
```diff
@@ -0,0 +1,320 @@
...
+		}
+		remote := m.targets[cReq.Host]
+		if remote == nil {
+			if debugPrint {
+				fmt.Printf("%s: handler: unknown host: %v. Have %v\n", m.local, cReq.Host, m.targets)
+			}
+			logger.LogIf(ctx, fmt.Errorf("handler: unknown host: %v", cReq.Host))
+			return
+		}
+		if debugPrint {
```
shtripat, line 211: Just a suggestion, should we have something like `DebugPrintIf` declared and used. will become one liner everywhere.

internal/grid/manager.go
```diff
@@ -0,0 +1,320 @@
...
+// Targets returns the names of all remote targets.
+func (m *Manager) Targets() []string {
+	var res []string
+	for k := range m.targets {
+		res = append(res, k)
+	}
+	return res
+}
+
+// TestingShutDown will shut down all connections.
```
shtripat: There is no function with name `TestingShutDown`

internal/grid/msg.go
```diff
@@ -0,0 +1,281 @@
...
+const (
+	// OpConnect is a connect request.
+	OpConnect Op = iota + 1
+
+	// OpConnectResponse is a response to a connect request.
+	OpConnectResponse
+
+	// OpPing is a ping request.
+	// If a mux id is specified that mux is pinged.
+	// Clients sends ping requests.
```
shtripat: `Clients send`

internal/grid/msg.go
```diff
@@ -0,0 +1,281 @@
...
+	if f&FlagPayloadIsZero != 0 {
+		res = append(res, "ZERO")
+	}
+	if f&FlagSubroute != 0 {
+		res = append(res, "SUB")
+	}
+	return "[" + strings.Join(res, ",") + "]"
+}
+
+// parse an handleIncoming message.
```
shtripat: do you mean `parse and handle incoming message`?

internal/grid/msg.go
```diff
@@ -0,0 +1,281 @@
...
+type muxConnectError struct {
+	Error string
+}
+
+func (muxConnectError) Op() Op {
+	return OpMuxConnectError
+}
+
+type pongMsg struct {
+	NotFound bool    `msg:"f"`
```
shtripat: `NotFound` fieldname and `f` in MSGP seem reverse of each.

internal/grid/muxclient.go
```diff
@@ -0,0 +1,539 @@
...
+	if msg.Flags&FlagCRCxxh3 != 0 {
+		h := xxh3.Hash(dst)
+		dst = binary.LittleEndian.AppendUint32(dst, uint32(h))
+	}
+	return m.parent.send(dst)
+}
+
+// RequestStateless will send a single payload request and stream back results.
+// req may not be read/written to after calling.
+// TODO: Not implemented
```
shtripat, line 173: TOD needed?

internal/grid/stream.go
```diff
@@ -0,0 +1,94 @@
...
+package grid
+
+import (
+	"context"
+	"errors"
+)
+
+// A Stream is a two-way stream.
+// All responses *must* be read by the caller.
+// If the call is canceled though the context,
```
shtripat: `%s/though/through/g`

internal/grid/stream.go
```diff
@@ -0,0 +1,94 @@
...
+func (s *Stream) Results(next func(b []byte) error) (err error) {
+	done := false
+	defer func() {
+		if !done {
+			if s.cancel != nil {
+				s.cancel(err)
+			}
+			// Drain channel.
+			// A goroutine shouldn't be needed.
+			for range s.responses {
```
shtripat, line 71: something missing here?

internal/grid/trace.go
```diff
@@ -0,0 +1,109 @@
...
+		Publisher: p,
+		TraceType: madmin.TraceInternal,
+		Prefix:    "grid.",
+		Local:     c.Local,
+		Remote:    c.Remote,
+		Subroute:  "",
+	}
+}
+
+// traceRequests adds request tracing to the connection.
```
shtripat: Correct the comments

internal/grid/trace.go
```diff
@@ -0,0 +1,109 @@
...
+
+func (c *muxClient) traceRoundtrip(t *tracer, h HandlerID, req []byte) ([]byte, error) {
+	if t == nil || t.Publisher.NumSubscribers(t.TraceType) == 0 {
+		return c.roundtrip(h, req)
+	}
+	start := time.Now()
+	body := bytesOrLength(req)
+	resp, err := c.roundtrip(h, req)
+	end := time.Now()
+	status := 200
```
shtripat: Should we define constants for status codes?

internal/grid/types.go
```diff
@@ -0,0 +1,172 @@
...
+
+// Get returns the value for the given key.
+func (m *MSS) Get(key string) string {
+	if m == nil {
+		return ""
+	}
+	return (*m)[key]
+}
+
+// UnmarshalMsg deserializes b from the provided byte slice and returns the
```
shtripat: `%s/deserializes b/deserializes bts/g`

shtripat (commented):
few minor ones

cmd/handler-utils.go
klauspost: Nope. All lock REST calls are ported.

cmd/lock-rest-client_test.go
klauspost: But the comment is correct. I am pretty sure that whatever it means, they aren't conbectable! :D

cmd/metacache-walk.go
klauspost: Yes?

cmd/storage-rest-server.go
klauspost: Seems the old code was written this way. Will update.

internal/grid/connection.go
klauspost: Yes. Later.

internal/grid/manager.go
klauspost: No. This will be removed by the compiler since `debugPrint` is static when compiled. That means that objects that could escape through the print operation no longer does. 

Therefore when disabled it is a truly 0-cost function. If the compiler gets fixed to recognize this at some point we can do that.

internal/grid/msg.go
klauspost: Refactor leftover.

internal/grid/muxclient.go
klauspost: Yes.

internal/grid/stream.go
klauspost: Mostly just note to self. Removed.

internal/grid/trace.go
klauspost: They are artificial anyway. But sure - might as well use the http consts for clarity.

cmd/lock-rest-client.go
klauspost: Well, it serves no purpose, but must be there for the interface.

cmd/lock-rest-client_test.go
shtripat: Ah yes.

cmd/metacache-walk.go
shtripat: I meant to change as `disk` word is used twice ;)

cmd/lock-rest-client.go
klauspost: Sorry. Looked at wron fn. You are right 👍🏼

internal/grid/README.md
klauspost: Yeah. My editor is extremely annoying with that.

minio-trusted (approved):

harshavardhana (approved):

PR #18500, opened 2023-11-22
ilm: ObjectSizeLessThan and ObjectSizeGreaterThan

krisis:

## Description
Add support for object size based predicates in bucket lifecycle polices (ILM).
The example in the link below captures where these size based predicates are useful.
https://docs.aws.amazon.com/AmazonS3/latest/userguide/lifecycle-configuration-examples.html#lc-size-rules

## Motivation and Context
Users may want to tier only larger objects to a warm tier. Size based predicates like ObjectSizeLessThan, ObjectSizeGreaterThan allow them to do this.

## How to test this PR?
Companion mc PR coming soon ...

## Types of changes

## Checklist:

### DIFF
--- a/internal/bucket/lifecycle/filter.go
+++ b/internal/bucket/lifecycle/filter.go
@@ -33,6 +33,9 @@ type Filter struct {
 
 	Prefix Prefix
 
+	ObjectSizeGreaterThan int64 `xml:"ObjectSizeGreaterThan,omitempty"`
+	ObjectSizeLessThan    int64 `xml:"ObjectSizeLessThan,omitempty"`
+
 	And    And
 	andSet bool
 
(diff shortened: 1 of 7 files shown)


internal/bucket/lifecycle/filter.go
```diff
@@ -122,32 +150,61 @@ func (f Filter) Validate() error {
...
+		pType = sizeGtPred
+		predCount++
+	}
+	if f.ObjectSizeLessThan > 0 {
+		pType = sizeLtPred
+		predCount++
+	}
+	if predCount > 1 {
+		return errInvalidFilter
+	}
```
harshavardhana, line 188: So in AWS you can only provide one predicate? because then why would at https://github.com/minio/minio/pull/18500/files#diff-f52f08b71d7c7f0ea61a74df360f8e1ea7daf770de643a7108b351e5d31b861bR58 there would be a check for `predCount < 2` ?

harshavardhana (changes requested):

internal/bucket/lifecycle/filter.go
krisis: An ILM rule can have exactly only one of the following predicates at the level of `<Filter></Filter>` such as,
-  `Tag` (singular)
-  `Prefix`
-  `And` (combination of 2 or more predicates)
-  `ObjectSizeLessThan`
-  `ObjectSizeGreaterThan`

So in this case, `predCount > 1` is incorrect.

When a user wishes to combine one or more predicates using the logical AND operator, then they will use `<And>`. This tag allows users to combine **two** or more predicates, which is why we check for `predCount > 2`.

harshavardhana: Add this as a comment since this is unnecessarily complicated for coders and users :-)

vadmeste (approved):
LGTM

harshavardhana (approved):

PR #18533, opened 2023-11-27
avoid run-away goroutine build-up in notification send, use channels

harshavardhana:

## Description
avoid run-away goroutine build-up in notification send, use channels

## Motivation and Context
use memory for async events when necessary and dequeue them as needed, 
for all synchronous events customers must enable

```
MINIO_API_SYNC_EVENTS=on
```

Async events can be lost but is up to the admin to decide what they want, 
we will not create a run-away number of goroutines per event instead we 
will queue them properly.

Currently, the max async workers is set to 500 which is more than sufficient 
in general, but it can be made configurable in the future.

## How to test this PR?
Setup a webhook notification target for your bucket and then
perform operations, batching the operations this way helps
notification targets cool off a bit instead of sending a
barrage of requests towards them.

## Types of changes

## Checklist:


harshavardhana:
20k goroutine was a ridiculous number for concurrent network calls that all can potentially block and wait, instead of that we should send limited concurrent calls but queue them instead to reduce notification endpoint pressure.

So a slow receiver would most definitely be losing events almost all the time on a busy system.

internal/event/targetlist.go
```diff
@@ -228,7 +272,39 @@ func (list *TargetList) Stats() TargetStats {
...
 // NewTargetList - creates TargetList.
-func NewTargetList() *TargetList {
-	return &TargetList{targets: make(map[TargetID]Target)}
+func NewTargetList(ctx context.Context) *TargetList {
+	list := &TargetList{
+		targets: make(map[TargetID]Target),
+		queue:   make(chan asyncEvent, maxConcurrentAsyncSend),
+		ctx:     ctx,
+	}
+	go list.startSendWorkers(500) // TODO:(y4m4) makde this configurable
```
Praveenrajmani:
```suggestion
	go list.startSendWorkers(500) // TODO:(y4m4) make this configurable
```

internal/event/targetlist.go
```diff
@@ -163,46 +183,69 @@ func (list *TargetList) TargetMap() map[TargetID]Target {
...
-				if err := target.Save(event); err != nil {
-					tgtRes.Err = err
-				}
-				resCh <- tgtRes
-			}(id, target)
-		} else {
-			resCh <- TargetIDResult{ID: id}
+		target, ok := list.get(id)
+		if !ok {
+			list.sendResult(resCh, id, nil)
```
Praveenrajmani: can we log this error? 

```suggestion
			list.sendResult(resCh, id, fmt.Errorf("target id %v not found", id))
```

or just continue instead by removing this line.

Praveenrajmani (approved):
Tested and LGTM

internal/event/targetlist.go
harshavardhana: it is to satisfy a test, fixed the test instead.

harshavardhana:
PTAL again @Praveenrajmani @poornas

harshavardhana:
I have updated the behavior and also simplified few other things.

harshavardhana:
PTAL @Praveenrajmani @shtripat

shtripat (approved):
LGTM

harshavardhana:
@Praveenrajmani are you able to take look again?

Praveenrajmani (approved):

Praveenrajmani:
Tested and LGTM

PR #18534, opened 2023-11-27
xl: Cleaning up will delete instead of rename to trash with full disk err

vadmeste:

## Description
moveToTrash() function moves a folder to .trash for example when trying to 
some deletion; for example; when removing a data-dir which contains the object multipart; 
However rename() can return ENOSPC and this may cripple a user trying to remove 
some objects in a full disk situation to free some space in the disk.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/xl-storage.go
+++ b/cmd/xl-storage.go
@@ -1095,18 +1095,27 @@ func (s *xlStorage) DeleteVersions(ctx context.Context, volume string, versions 
 	return errs
 }
 
-func (s *xlStorage) moveToTrash(filePath string, recursive, force bool) error {
+func (s *xlStorage) moveToTrash(filePath string, recursive, force bool) (err error) {
 	pathUUID := mustGetUUID()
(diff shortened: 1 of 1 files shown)


harshavardhana (approved):

cmd/xl-storage.go
```diff
@@ -1095,20 +1095,29 @@ func (s *xlStorage) DeleteVersions(ctx context.Context, volume string, versions
...
+		} else {
+			err = Remove(filePath)
 		}
 	}
 
+	if err != nil {
+		return err
+	}
+
 	// immediately purge the target
```
krisis, line 1121: if we called `remove` or `removeAll` due to disk-full case, we won't need to remove `targetPath` when `force = true`.

krisis (approved):

jiuker (approved):

cmd/xl-storage.go
vadmeste: I think 'force' is not needed here; it seems it is misunderstood

vadmeste: yes, I confirm force flag is not needed

PR #18537, opened 2023-11-27
Skip versions expired by DeleteAllVersionsAction

krisis:

## Description
Object versions expired by `DeleteAllVersionsAction` must not be included toward data-usage accounting. Also, current object versions expired via `DeleteAction`, in a versioned bucket, must not be removed from data-usage accounting as this results in a delete marker on top of the 'expired' version.

## Motivation and Context
Objects expired by `DeleteAllVersionsAction` (MinIO extension) could be accounted incorrectly as non-versioned objects. This could happen when an object has exactly one version at the time of such expiry. 

## Types of changes

## Checklist:

### DIFF
--- a/cmd/data-scanner.go
+++ b/cmd/data-scanner.go
@@ -917,16 +917,17 @@ func (i *scannerItem) applyHealing(ctx context.Context, o ObjectLayer, oi Object
 	return 0
 }
 
-func (i *scannerItem) applyLifecycle(ctx context.Context, o ObjectLayer, oi ObjectInfo) (applied bool, size int64) {
+func (i *scannerItem) applyLifecycle(ctx context.Context, o ObjectLayer, oi ObjectInfo) (action lifecycle.Action, size int64) {
 	size, err := oi.GetActualSize()
 	if i.debug {
 		logger.LogIf(ctx, err)
 	}
 	if i.lifeCycle == nil {
-		return false, size
+		return action, size
 	}
 
 	versionID := oi.VersionID
+	vcfg, _ := globalBucketVersioningSys.Get(i.bucket)
 	rCfg, _ := globalBucketObjectLockSys.Get(i.bucket)
 	replcfg, _ := getReplicationConfig(ctx, i.bucket)
 	lcEvt := evalActionFromLifecycle(ctx, *i.lifeCycle, rCfg, replcfg, oi)
@@ -938,7 +939,7 @@ func (i *scannerItem) applyLifecycle(ctx context.Context, o ObjectLayer, oi Obje
 		}
 	}
 	defer func() {
-		if applied {
+		if lcEvt.Action != lifecycle.NoneAction {
 			numVersions := uint64(1)
 			if lcEvt.Action == lifecycle.DeleteAllVersionsAction {
 				numVersions = uint64(oi.NumVersions)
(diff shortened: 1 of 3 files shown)


cmd/data-scanner.go
```diff
@@ -938,7 +939,7 @@ func (i *scannerItem) applyLifecycle(ctx context.Context, o ObjectLayer, oi Obje
 		}
 	}
 	defer func() {
-		if applied {
+		if lcEvt.Action > lifecycle.NoneAction {
```
vadmeste: can we do '!=' instead of '>' ? (just a little bit more stable if someone plays with the order of the definition of ILM actions)

cmd/xl-storage.go
```diff
@@ -569,9 +569,15 @@ func (s *xlStorage) NSScanner(ctx context.Context, cache dataUsageCache, updates
...
 			done = globalScannerMetrics.time(scannerMetricApplyVersion)
-			sz := item.applyActions(ctx, objAPI, oi, &sizeS)
+			objDeleted, sz := item.applyActions(ctx, objAPI, oi, &sizeS)
 			done()
 
+			// DeleteAllVersionsAction: The object and all its
+			// versions are expired and
+			// doesn't contribute toward data usage.
+			if objDeleted {
+				return sizeSummary{}, errSkipFile
```
vadmeste: shouldn't this be 'break' instead of return ?

vadmeste (commented):
Some comments

cmd/xl-storage.go
krisis: This has to be `return`. `objDeleted` here indicates if all versions of this objects are expired. This is true only when `DeleteAllVersionsAction` flag is enabled in the ILM rule applicable. At this point we have no accounting related actions to be done for this object.

vadmeste: yes, but we still have some "unrelated" code that we can execute meanwhile, such as, ILM transitioning sweeping and trace

krisis: Good point, especially about cleaning up free-versions.  Will update PR.

cmd/xl-storage.go
```diff
@@ -644,6 +652,12 @@ func (s *xlStorage) NSScanner(ctx context.Context, cache dataUsageCache, updates
 				}
 			}
 		}
+		if objDeleted {
+			// we return errSkipFile to signal this function's
+			// callers to skip this object's contribution towards
+			// usage.
+			return sizeSummary{}, errSkipFile
```
vadmeste: I think we should not return an error here; this looks like a clarity simplificiation but this can cause a performance regression; basically all expired objects may go to a healing check performed by the caller using abondonedChildren mechanism (objects found in another disk in the last scan but not found in the current disk, we execute healing of those objects)

/cc @klauspost

vadmeste (changes requested):

cmd/xl-storage.go
krisis: @vadmeste Your concern is valid. @klauspost I need a way to signal from within the `getSize` callback function that this `item` doesn't contribute toward data usage. I have added a commit on top of my PR to address @vadmeste 's concern. Let me know if this is acceptable.

harshavardhana (approved):

cmd/xl-storage.go
```diff
@@ -644,6 +652,12 @@ func (s *xlStorage) NSScanner(ctx context.Context, cache dataUsageCache, updates
 				}
 			}
 		}
+		if objDeleted {
+			// we return errSkipFile to signal this function's
+			// callers to skip this object's contribution towards
+			// usage.
```
harshavardhana:
```suggestion
			// we return errIgnoreFileContrib to signal this function's
			// callers to skip this object's contribution towards
			// usage.
```

harshavardhana (changes requested):

klauspost (approved):

harshavardhana (approved):

PR #18239, opened 2023-10-13
feat: bring new HDD related performance enhancements 

harshavardhana:
## Community Contribution License
All community contributions in this pull request are licensed to the project maintainers
under the terms of the [Apache 2 license] (https://www.apache.org/licenses/LICENSE-2.0). 
By creating this pull request I represent that I have the right to license the 
contributions to the project maintainers under the Apache 2 license.

## Description
Optionally allows customers to enable 

- Enable an external cache to catch GET/HEAD responses 
- Enable skipping disks that are slow to respond in GET/HEAD 
  when we have already achieved a quorum

## Motivation and Context
Quite simply, these optimizations are for HDD-based deployments.

## How to test this PR?
Enable these features via

```
mc admin config set alias/ cache endpoint=http://xx:8080
```

To enable the endpoint, you would need a private binary called `mincache`
available for Enterprise SUBNET customers.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Optimization (provides speedup with no functional changes)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [ ] Fixes a regression (If yes, please add `commit-id` or `PR #` here)
- [ ] Unit tests added/updated
- [ ] Internal documentation updated
- [ ] Create a documentation update request [here](https://github.com/minio/docs/issues/new?label=doc-change,title=Doc+Updated+Needed+For+PR+github.com%2fminio%2fminio%2fpull%2fNNNNN)

PR #18548, opened 2023-11-28
Fix shadowed error

klauspost:
## Description

#18239 introduced a shadowed error. This will result in `done <- err == nil` always returning true for this path, which seems unintentional.

## Types of changes


harshavardhana (approved):

krisis (approved):

vadmeste (approved):
LGTM, thanks for catching this

harshavardhana (approved):

PR #18549, opened 2023-11-28
Improve env var config error reporting

klauspost:
## Description

Env vars that were set on current server but not on remotes were not reported in errors.

Add these.

## How to test this PR?

Create mismatching env vars in cluster.

## Types of changes

### DIFF
--- a/cmd/bootstrap-peer-server.go
+++ b/cmd/bootstrap-peer-server.go
@@ -20,12 +20,14 @@ package cmd
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


harshavardhana (approved):

vadmeste (commented):
I think showing both 'missing' and 'extra' errors in the same node is a little bit confusing; probably a better message would be:

Please ensure those env variables are set and match in all nodes: []missing+extra, []mismatch

vadmeste (approved):
Actually I just realised this error message is per node and not generic 

Approving

PR #18554, opened 2023-11-29
cleanup handling of STS isAllowed and simplifies the PolicyDBGet()

harshavardhana:

## Description
cleanup handling of STS isAllowed and simplifies the PolicyDBGet()

## Motivation and Context
regular cleanup to avoid mistakes and also fix 

- TemporaryAccountInfo() did not honor groups for STS
- TemporaryAccountInfo() did not honor policies from claims when parent mapping is missing

## How to test this PR?
CI/CD should cover the tests

## Types of changes

## Checklist:

### DIFF
--- a/cmd/admin-handlers-users.go
+++ b/cmd/admin-handlers-users.go
@@ -545,14 +545,16 @@ func (a adminAPIHandlers) TemporaryAccountInfo(w http.ResponseWriter, r *http.Re
 		return
 	}
 
-	if !globalIAMSys.IsAllowed(policy.Args{
+	args := policy.Args{
 		AccountName:     cred.AccessKey,
 		Groups:          cred.Groups,
 		Action:          policy.ListTemporaryAccountsAdminAction,
 		ConditionValues: getConditionValues(r, "", cred),
 		IsOwner:         owner,
 		Claims:          cred.Claims,
-	}) {
+	}
+
+	if !globalIAMSys.IsAllowed(args) {
 		writeErrorResponseJSON(ctx, w, errorCodes.ToAPIErr(ErrAccessDenied), r.URL)
 		return
 	}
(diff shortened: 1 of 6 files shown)


donatello (approved):

harshavardhana:
PTAL @vadmeste

cmd/admin-handlers-users.go
```diff
@@ -568,11 +570,16 @@ func (a adminAPIHandlers) TemporaryAccountInfo(w http.ResponseWriter, r *http.Re
...
-		policiesNames, err := globalIAMSys.PolicyDBGet(stsAccount.ParentUser, false)
+		policiesNames, err := globalIAMSys.PolicyDBGet(stsAccount.ParentUser, stsAccount.Groups...)
 		if err != nil {
 			writeErrorResponseJSON(ctx, w, toAdminAPIErr(ctx, err), r.URL)
 			return
 		}
+		if len(policiesNames) == 0 {
+			policySet, _ := args.GetPolicies(iamPolicyClaimNameOpenID())
+			policiesNames = policySet.ToSlice()
+		}
```
vadmeste, line 581: I am not sure I understand this; this is an existing account, which means either it has an explicit policy or it will inherit the policy from the parent user;  if it has an explicit policy, this should be already handled in L569/; if the parent user doesn't have any policy, I think it is enough to stop

harshavardhana: Not really, @vadmeste - there were always situations where we never saved parent-mapped policies in older releases. This is needed for backward compatibility. When customers upgrade from older deployments, we want their STS credentials to work until expired.

NEw STS will follow the parent-mapped policy approach.

vadmeste (approved):
LGTM

PR #18557, opened 2023-11-29
Add TLS cert checksum

klauspost:
## Description

Allows to validate if all certs match across cluster.

Requires https://github.com/minio/madmin-go/pull/251

## How to test this PR?

Submit health check.

## Types of changes

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -49,7 +49,7 @@ require (
 	github.com/minio/dperf v0.5.2
 	github.com/minio/highwayhash v1.0.2
 	github.com/minio/kes-go v0.2.0
-	github.com/minio/madmin-go/v3 v3.0.34
+	github.com/minio/madmin-go/v3 v3.0.35-0.20231130082526-199918d0ff20
 	github.com/minio/minio-go/v7 v7.0.65-0.20231122233251-1f7dd6b7e3e1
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -247,3 +247,5 @@ require (
 	gopkg.in/square/go-jose.v2 v2.6.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+replace github.com/minio/madmin-go/v3 => github.com/klauspost/madmin-go/v3 v3.0.0-20231129155207-e7ce2dbac649
```
aead: The replace should be removed, right?!

klauspost: Yeah. Once https://github.com/minio/madmin-go/pull/251 is merged. It is there so CI can compile.

harshavardhana (approved):

vadmeste (commented):
wondering why we don't use the certificate serial number ? just we do not want to be aggressive?

harshavardhana:
> wondering why we don't use the certificate serial number ? just we do not want to be aggressive?

Serial number apparently can be re-used or shared? @vadmeste

vadmeste (approved):
LGTM

PR #18567, opened 2023-11-30
Fix Mux Connect Error

klauspost:
## Description

`OpMuxConnectError` were not handled correctly. Not released.

Remove local check for single request handler, so they can run before being registered locally.

Bonus: Only log IAM bootstrap on startup.

## How to test this PR?

@Praveenrajmani is our local hero for this.

## Types of changes

### DIFF
--- a/cmd/metacache-set.go
+++ b/cmd/metacache-set.go
@@ -606,8 +606,8 @@ func (er *erasureObjects) listPath(ctx context.Context, o listPathOptions, resul
 	defer close(results)
 	o.debugf(color.Green("listPath:")+" with options: %#v", o)
 
-	// get non-healing disks for listing
-	disks, _ := er.getOnlineDisksWithHealing()
+	// get prioritized non-healing disks for listing
+	disks, _ := er.getOnlineDisksWithHealing(true)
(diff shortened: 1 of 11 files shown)


harshavardhana (approved):

cmd/metacache-set.go
```diff
@@ -606,8 +606,8 @@ func (er *erasureObjects) listPath(ctx context.Context, o listPathOptions, resul
 	defer close(results)
 	o.debugf(color.Green("listPath:")+" with options: %#v", o)
 
-	// get non-healing disks for listing
-	disks, _ := er.getOnlineDisksWithHealing()
+	// get prioritized non-healing disks for listing
+	disks, _ := er.getOnlineDisksWithHealing(true)
```
harshavardhana, line 610: Do we want healing disks to be included? That only has partial content, anyway? - it will sure tax that drive more IMO.

harshavardhana (changes requested):

cmd/metacache-set.go
harshavardhana: Actually I see there is benefit in it.

harshavardhana (approved):

PR #18576, opened 2023-12-01
Fix drives fixed after startup

klauspost:
## Description

Fixes unregistered drives on startup - continuation of #18151 - fixing merge issues and minor things (not as async startup).

All handlers call either IsAuthValid, IsValid (which calls IsAuthValid) or checkID before using the storage layer, which all check if it has been set.

## Motivation and Context

Same as #18151 

## How to test this PR?

@Praveenrajmani Has a good setup. Marking disks read-only before startup.

## Types of changes


harshavardhana (approved):

vadmeste (approved):
LGTM

PR #18581, opened 2023-12-02
avoid hot-tier SLA to be tied to warm-tier SLA

harshavardhana:

## Description
avoid hot-tier SLA being tied to warm-tier SLA

## Motivation and Context
It is okay if the warm tier cannot keep up. We should 
continue taking I/O at the hot tier, only failing it 
or blocking it when the disk is fully used.

Bonus: add a metrics counter for these missed tasks; we 
will know if one of the nodes is lagging or is losing too 
many tasks during transitioning.

## How to test this PR?
Observe if the warm-tier goes down, how much performance
 the PUTs take.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/bucket-lifecycle.go
+++ b/cmd/bucket-lifecycle.go
@@ -219,24 +219,23 @@ type transitionState struct {
 	numWorkers int
 	killCh     chan struct{}
 
-	activeTasks int32
+	activeTasks          atomic.Int64
+	missedImmediateTasks atomic.Int64
 
 	lastDayMu    sync.RWMutex
 	lastDayStats map[string]*lastDayTierStats
 }
 
-func (t *transitionState) queueTransitionTask(oi ObjectInfo, event lifecycle.Event, src lcEventSrc, blocking bool) {
+func (t *transitionState) queueTransitionTask(oi ObjectInfo, event lifecycle.Event, src lcEventSrc) {
 	task := transitionTask{objInfo: oi, event: event, src: src}
-	if blocking {
-		select {
-		case <-t.ctx.Done():
-		case t.transitionCh <- task:
-		}
-	} else {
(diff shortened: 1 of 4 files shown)


cmd/bucket-lifecycle.go
```diff
@@ -219,24 +219,23 @@ type transitionState struct {
 	numWorkers int
 	killCh     chan struct{}
 
-	activeTasks int32
+	activeTasks          atomic.Int64
+	missedImmediateTasks atomic.Int64
```
krisis, line 223: Neat!

krisis (approved):

vadmeste:
the hot-tier SLA tied to the warm tier SLA was only activated when the lifecycle transition days is zero; which I think  it is only needed when you add a warm tier the very first time. Most users who set the transition days to zero are in a hurry and want immediate results. This will also ensure that the hot tier will never get saturated. Do you know any use case when this is a bad idea ?

harshavardhana:
> the hot-tier SLA tied to the warm tier SLA was only activated when the lifecycle transition days is zero; which I think it is only needed when you add a warm tier the very first time. Most users who set the transition days to zero are in a hurry and want immediate results. This will also ensure that the hot tier will never get saturated. Do you know any use case when this is a bad idea ?

Yes always @vadmeste we are now looking at deployments are permanently in immediate transition mode. For such a deployment style we can't have it as a blocking call.

vadmeste:
> Yes always @vadmeste we are now looking at deployments are permanently in immediate transition mode. For such a deployment style we can't have it as a blocking call.

Okay, the other PR was more focusing on slowing down s3 clients when the hot tier is much faster in receiving objects than the warm tier and the transition days is zero; but I understand the trade off as well 👍

PR #18588, opened 2023-12-04
fix: trans cancel error into interval error when context is not cancaled

jiuker:

## Description
before , we have this.
```go
	// Only return ErrClientDisconnected if the provided context is actually canceled.
	// This way downstream context.Canceled will still report ErrRequestTimedout
	if contextCanceled(ctx) && errors.Is(ctx.Err(), context.Canceled) {
		return ErrClientDisconnected
	}
```
So the context is no canceled when:
```go
	case context.Canceled:
		apiErr = ErrClientDisconnected
```
fix: trans cancel error into interval error when context is not cancaled
When we got `Client disconnected before response was ready`, we can't found that if the client do cancel really.
It's  a timeout between the minio-servers maybe.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/api-errors.go
+++ b/cmd/api-errors.go
@@ -2158,7 +2158,8 @@ func toAPIErrorCode(ctx context.Context, err error) (apiErr APIErrorCode) {
 	case errKMSDefaultKeyAlreadyConfigured:
 		apiErr = ErrKMSDefaultKeyAlreadyConfigured
 	case context.Canceled:
-		apiErr = ErrClientDisconnected
+		// the ctx is not Canceled. But the error is Canceled. We should assume that this error is errInternalError
+		apiErr = ErrInternalError
 	case context.DeadlineExceeded:
 		apiErr = ErrRequestTimedout
 	case objectlock.ErrInvalidRetentionDate:


vadmeste:
> When we got Client disconnected before response was ready, we can't found that if the client do cancel really.
It's a timeout between the minio-servers maybe.

A timeout is usually a context.DeadlineExceeded error; However I don't see clearly what is wrong exactly; is there any specific bug you wanted to address ?

jiuker:
> > When we got Client disconnected before response was ready, we can't found that if the client do cancel really.
> > It's a timeout between the minio-servers maybe.
> 
> A timeout is usually a context.DeadlineExceeded error; However I don't see clearly what is wrong exactly; is there any specific bug you wanted to address ?

Yeah. Some issue just log out `Client disconnected before response was ready` at `PutObjectPart`. There have two ways to log this. One is client cancel the request, another one is minio-server interval peer call time-out. we can't found that if the client do cancel really for just one error log out. @vadmeste

klauspost (changes requested):
This is mostly a wrong fix. You are fixing one case, where there are 100 correct ones. 

You need to figure out what is returning the "context canceled" and fix that up at a lower level.

jiuker:
> This is mostly a wrong fix. You are fixing one case, where there are 100 correct ones.
> 
> You need to figure out what is returning the "context canceled" and fix that up at a lower level.

Yeah. @vadmeste should have some thoughts here. @klauspost

PR #18593, opened 2023-12-05
Truncate long traces

klauspost:
## Motivation and Context

Prevent excessively long request traces

## How to test this PR?

Observe traces.

## Types of changes

### DIFF
--- a/internal/grid/trace.go
+++ b/internal/grid/trace.go
@@ -129,7 +129,11 @@ func (c *muxClient) traceRoundtrip(ctx context.Context, t *tracer, h HandlerID, 
 		trace.HTTP.ReqInfo.Path = trace.Path
 	} else if v != nil {
 		// Print exported fields as single request to path.
-		trace.Path = fmt.Sprintf("%s?req=%s", trace.Path, url.QueryEscape(fmt.Sprintf("%+v", v)))
+		obj := fmt.Sprintf("%+v", v)
+		if len(obj) > 1024 {
+			obj = obj[:1024] + "..."
+		}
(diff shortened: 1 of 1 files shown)


harshavardhana (approved):

internal/grid/trace.go
```diff
@@ -129,7 +129,11 @@ func (c *muxClient) traceRoundtrip(ctx context.Context, t *tracer, h HandlerID,
 		trace.HTTP.ReqInfo.Path = trace.Path
 	} else if v != nil {
 		// Print exported fields as single request to path.
-		trace.Path = fmt.Sprintf("%s?req=%s", trace.Path, url.QueryEscape(fmt.Sprintf("%+v", v)))
+		obj := fmt.Sprintf("%+v", v)
+		if len(obj) > 1024 {
```
harshavardhana, line 133:
```suggestion
		if len(obj) > 1024 {
```

still 1k is a lot? @klauspost

klauspost: It is only internal traces... So I think it is safe to say that it is rarely enabled.... What would you like it to be?

harshavardhana: I am thinking that can this be just like 64bytes and not more than that. Seeing entire fileinfo is not a requirement really TBH.

harshavardhana, line 133:
```suggestion
		if len(obj) > 64 {
```

vadmeste (approved):
LGTM

harshavardhana:
Let's take this in ..

PR #18595, opened 2023-12-05
Enable setting throttle config at bucket level

shtripat:

## How to test this PR?
Follow https://github.com/minio/mc/pull/4780

## Types of changes

## Checklist:


cmd/site-replication.go
```diff
@@ -3421,7 +3421,32 @@ func isBktQuotaCfgReplicated(total int, quotaCfgs []*madmin.BucketQuota) bool {
...
+	}
+	return true
+}
+
+func testEqual(a, b []string) bool {
+	if len(a) != len(b) {
+		return false
+	}
+	for i := range a {
+		if a[i] != b[i] {
```
harshavardhana: you are thinking about `slices.Equal()` ?

cmd/bucket-quota.go
```diff
@@ -131,6 +133,10 @@ func (sys *BucketQuotaSys) enforceQuotaHard(ctx context.Context, bucket string,
 		}
 	}
 
+	if err := enforceBucketThrottle(ctx, bucket); err != nil {
+		return err
+	}
+
 	return nil
```
harshavardhana:
```suggestion
	return enforceBucketThrottle(ctx, bucket)
```

cmd/bucket-quota.go
```diff
@@ -140,3 +146,33 @@ func enforceBucketQuotaHard(ctx context.Context, bucket string, size int64) erro
 	}
 	return globalBucketQuotaSys.enforceQuotaHard(ctx, bucket, size)
 }
+
+func enforceBucketThrottle(ctx context.Context, bucket string) error {
+	q, err := globalBucketQuotaSys.Get(ctx, bucket)
```
harshavardhana: do not fetch the config multiple times, fetch it once and pass it around for each enforcement functions.

cmd/bucket-quota.go
```diff
@@ -140,3 +146,33 @@ func enforceBucketQuotaHard(ctx context.Context, bucket string, size int64) erro
...
+	}
+	currS3ReqStats := globalHTTPStats.toServerHTTPStats().CurrentS3Requests.APIStats
+	for api, currReqCount := range currS3ReqStats {
+		for allowedAPI, allowedCount := range s3AllowedReqCountMap {
+			// rules can have exact API names or patterns like `Get*`
+			if matched, _ := regexp.MatchString(strings.ToLower(allowedAPI), strings.ToLower(api)); matched && uint64(currReqCount) > allowedCount {
+				return BucketQuotaExceeded{Bucket: bucket}
+			}
+		}
+	}
```
harshavardhana: Before you loop through, reject the most common when this quota is not set use-case, which is normal. 

Such as `if !cfg.IsBucketThrottled() { return nil }`

cmd/bucket-quota.go
```diff
@@ -140,3 +146,33 @@ func enforceBucketQuotaHard(ctx context.Context, bucket string, size int64) erro
...
+				s3AllowedReqCountMap[api] = rule.ConcurrentRequestsCount
+			}
+		}
+	}
+	currS3ReqStats := globalHTTPStats.toServerHTTPStats().CurrentS3Requests.APIStats
+	for api, currReqCount := range currS3ReqStats {
+		for allowedAPI, allowedCount := range s3AllowedReqCountMap {
+			// rules can have exact API names or patterns like `Get*`
+			if matched, _ := regexp.MatchString(strings.ToLower(allowedAPI), strings.ToLower(api)); matched && uint64(currReqCount) > allowedCount {
+				return BucketQuotaExceeded{Bucket: bucket}
```
harshavardhana: Extend the error to provide more context if the quota is exceeded for size or number of requests.

cmd/bucket-quota.go
```diff
@@ -140,3 +146,33 @@ func enforceBucketQuotaHard(ctx context.Context, bucket string, size int64) erro
...
+			} else {
+				s3AllowedReqCountMap[api] = rule.ConcurrentRequestsCount
+			}
+		}
+	}
+	currS3ReqStats := globalHTTPStats.toServerHTTPStats().CurrentS3Requests.APIStats
+	for api, currReqCount := range currS3ReqStats {
+		for allowedAPI, allowedCount := range s3AllowedReqCountMap {
+			// rules can have exact API names or patterns like `Get*`
+			if matched, _ := regexp.MatchString(strings.ToLower(allowedAPI), strings.ToLower(api)); matched && uint64(currReqCount) > allowedCount {
```
harshavardhana: Don't use `regexp` instead use wildcard package to match - regexp is very slow in hot I/O path.

harshavardhana (changes requested):

cmd/bucket-quota.go
```diff
@@ -140,3 +146,33 @@ func enforceBucketQuotaHard(ctx context.Context, bucket string, size int64) erro
...
 	return globalBucketQuotaSys.enforceQuotaHard(ctx, bucket, size)
 }
+
+func enforceBucketThrottle(ctx context.Context, bucket string) error {
+	q, err := globalBucketQuotaSys.Get(ctx, bucket)
+	if err != nil {
+		return err
+	}
+	// Check the breach of throttle rules
+	s3AllowedReqCountMap := make(map[string]uint64)
```
krisis: This PR allows for users to configure multiple concurrent-requests limits for the same API (or a set of APIs). We resolve this by picking the lowest limit placed on an API, which is fine.

What does the user gain by being able to configure different concurrent requests limit for the same API? I think we should return an error at the time of parsing these rules when we find an API being configured more than one limit value.

cmd/bucket-quota.go
```diff
@@ -140,3 +146,33 @@ func enforceBucketQuotaHard(ctx context.Context, bucket string, size int64) erro
 	}
 	return globalBucketQuotaSys.enforceQuotaHard(ctx, bucket, size)
 }
+
+func enforceBucketThrottle(ctx context.Context, bucket string) error {
```
krisis: This function must take the API name so that we can enforce quota throttle rules for this API alone.

cmd/site-replication.go
shtripat: Actually yes. Will change and use `slices.Equal()`

cmd/bucket-quota.go
shtripat: Thanks for pointing this. Will change this.

cmd/bucket-quota.go
shtripat: Ok, will add more details to it. thanks

cmd/bucket-quota.go
shtripat: Ok. will change

cmd/bucket-quota.go
shtripat: hmm, makes sense. I made it like this just because there could be wildcard rule say for `Get*` and specific one like `GetObject`.
But we can think of making it more versatile like below
1. If say generic wildcard based rule set for `Get*` and another one is being set for `GetObject`, we consider value of `GetObject` calls but any other Get calls we consider `Get*` value
2. Don't allow setting multiple rules for same APIs and if already exists, error out

Does this make sense @krisis @harshavardhana ?

cmd/bucket-quota.go
shtripat: Ack. will change.

cmd/bucket-quota.go
shtripat: I tried this where while adding new rules we would see if there are duplicates using below code
```
// HasDuplicateThrottleRules returns true if throttle rules set more than once
// for same APIs
func (q BucketQuota) HasDuplicateThrottleRules() (bool, string) {
       rulesMap := make(map[string]uint64)
       for _, rule := range q.ThrottleRules {
               for _, api := range rule.APIs {
                       if _, ok := rulesMap[strings.ToLower(api)]; ok {
                               return true, api
                       } else {
                               rulesMap[strings.ToLower(api)] = rule.ConcurrentRequestsCount
                       }
               }
       }
       return false, ""
}
```
With this change say we already have rules for `PutObject,Get*` and limit set as 100. Now if we try to set another rule for `PutObject` say with limit 200, it wont allow as there is limit set already for this API.
But if we want to change the limit as a requirement for `PutObject`, we will have to change for the whole APIs list for `PutObject,Get*` as each rule is treated unique by its list of APIs.
So we can either retain the logic as is and consider the lowest limit value across all rule for an API or let user set new value for the rule using the full APIs list (order of strings doesn't matter as we sort and then compare).
Suggestions??

krisis: >With this change say we already have rules for PutObject,Get* and limit set as 100. Now if we try to set another rule for PutObject say with limit 200, it wont allow as there is limit set already for this API.

Yes. I prefer this to MinIO silently resolving a possible misconfiguration. The motivation here is to make the rules system predictable.

>But if we want to change the limit as a requirement for PutObject, we will have to change for the whole APIs list for PutObject,Get* as each rule is treated unique by its list of APIs.

If they really want `PutObject`s to have a different limit than others in that API list, they should move it to separate rule, despite it being inconvenient, right?
But I understand that we can't completely escape the possibility that there are legitimate cases where an API has more than one rule apply to it. e.g `Get*` in a rule and `GetObject` in a separate rule. I like how you resolve this one, by giving the rule which applies to the API explicitly preference. This feels intuitive to me, but it must be made clear to our users. 

From this discussion, I think the following are the requirements for this feature. I took the liberty of adding observability into the mix. Feel free to change the following as you see correct. 

### Requirements
- Basic
    - A user must be able to set a limit on the maximum concurrent requests for each S3 API for a bucket
    - A user must be able to modify this limit at a later time.

- Convenience
    - A user must be able to such a limit for a group of APIs whose names share a common prefix

- Observability
    - A user should be able to learn more information about a throttled API call
      a. What was the applicable limit at that time?
      b. Which rule was responsible for the limit?
    - Each rule must have its own unique ID.
      - This can be provided by the user, should be generated by MinIO server when absent
      - This ID must be returned in the error client applications may receive when this limit crossed
      - This ID must be present in audit-logs, error logs and bucket events where applicable

shtripat: > If they really want PutObjects to have a different limit than others in that API list, they should move it to separate rule, despite it being inconvenient, right?

With current model actually we are allowed to create new rule for `PutObject` where already we have rule for say `PutObject,Get*` as new rule would be separate list of APIs (`PutObject`) and treated as new rule. Later while applying whatever limit is smaller from these two rules would be applied for `PutObject`.

> But I understand that we can't completely escape the possibility that there are legitimate cases where an API has more than one rule apply to it. e.g Get* in a rule and GetObject in a separate rule. I like how you resolve this one, by giving the rule which applies to the API explicitly preference. This feels intuitive to me, but it must be made clear to our users.

Yes, Agree.

I like the idea of ID for each rule. Allow edit later by ID would be additional feature which brings actually edit under `mc quota` as well.
Regarding observability aspects, is it like HTTP Response should have more details like which rule applied with what limit and why API was rejected? Also adding audit logs, events, error logs while throttle quota breach.

I will change accordingly and update the PR.

shtripat: Now with latest commit, each rule gets an unique ID assigned while creation. Also changing allowed concurrent requests count can be done using this ID.

During a breach now the error message would look something like `Throttle quota exceeded. Applied RuleID: clsogdlb4dtk8pn3p59g, Allowed Requests Count: 10, Current Requets Count: 81, API: PutObject` to the client.

MinIO would log the error as 
```
API: PutObject(bucket=test-bucket, object=file350)
Time: 10:33:56 UTC 12/13/2023
DeploymentID: cd195cc1-ad90-4697-8665-1b0c15952c98
RequestID: 17A05DA845C95CB6
RemoteHost: [::1]
Host: localhost:9000
UserAgent: MinIO (linux; amd64) minio-go/v7.0.63
Error: Throttle quota exceeded. Applied RuleID: clsogdlb4dtk8pn3p59g, Allowed Requests Count: 10, Current Requets Count: 105, API: PutObject (cmd.BucketThrottleQuotaExceeded)
```

Also trace (verbose) would show details as below
```
localhost:9000 [REQUEST s3.PutObject] [2023-12-13T15:52:54.685] [Client IP: [::1]]
localhost:9000 PUT /test-bucket/file463
localhost:9000 Proto: HTTP/1.1
localhost:9000 Host: localhost:9000
localhost:9000 X-Amz-Content-Sha256: STREAMING-AWS4-HMAC-SHA256-PAYLOAD
localhost:9000 X-Amz-Date: 20231213T102254Z
localhost:9000 X-Amz-Decoded-Content-Length: 28
localhost:9000 Authorization: AWS4-HMAC-SHA256 Credential=minioadmin/20231213/us-east-1/s3/aws4_request,SignedHeaders=host;x-amz-content-sha256;x-amz-date;x-amz-decoded-content-length,Signature=6bc6d58e4fc07bcc1cd65adcac203774d13145e0e6580ce3c0f1a5b5509878e3
localhost:9000 Content-Length: 201
localhost:9000 Content-Type: application/octet-stream
localhost:9000 User-Agent: MinIO (linux; amd64) minio-go/v7.0.63
localhost:9000 <BLOB>
localhost:9000 [RESPONSE] [2023-12-13T15:52:54.702] [ Duration 17.519ms TTFB 17.481994ms ↑ 119 B  ↓ 472 B ]
localhost:9000 400 Bad Request
localhost:9000 Accept-Ranges: bytes
localhost:9000 Content-Type: application/xml
localhost:9000 Strict-Transport-Security: max-age=31536000; includeSubDomains
localhost:9000 X-Amz-Id-2: dd9025bab4ad464b049177c95eb6ebf374d3b3fd1af9251148b658df7ac2e3e8
localhost:9000 X-Content-Type-Options: nosniff
localhost:9000 Content-Length: 472
localhost:9000 Server: MinIO
localhost:9000 Vary: Origin,Accept-Encoding
localhost:9000 X-Amz-Request-Id: 17A05D0E31614BC8
localhost:9000 X-Xss-Protection: 1; mode=block
localhost:9000 <?xml version="1.0" encoding="UTF-8"?>
<Error><Code>XMinioAdminBucketThrottleQuotaExceeded</Code><Message>Throttle quota exceeded. Applied RuleID: clsobntb4dtjd30m4arg, Allowed Requests Count: 10, Current Requets Count: 135, API: PutObject</Message><Key>file463</Key><BucketName>test-bucket</BucketName><Resource>/test-bucket/file463</Resource><RequestId>17A05D0E31614BC8</RequestId><HostId>dd9025bab4ad464b049177c95eb6ebf374d3b3fd1af9251148b658df7ac2e3e8</HostId></Error>
```

https://github.com/minio/mc/pull/4780 has verification details.

shtripat:
Closing as not needed.

PR #18619, opened 2023-12-08
fix: remove ODirectReader entirely since we do not need it anymore

harshavardhana:

## Description
fix: remove ODirectReader entirely since we do not need it anymore

## Motivation and Context
We do not need this anymore; we obliterate it.

## How to test this PR?
Nothing should change much

## Types of changes

## Checklist:


vadmeste (commented):
Maybe it is still good to keep odirect read code but disable it by default (if not done already); I suspect that it is better to bypass the kernel cache with fast nvme disks;

klauspost (approved):

klauspost:
I think we can just dig op the code from here if we should need it.

harshavardhana:
> Maybe it is still good to keep odirect read code but disable it by default (if not done already); I suspect that it is better to bypass the kernel cache with fast nvme disks;

There are some other underlying problems - we need to get this merged - there is a 
potential for buffers to be shared.

There is no clear way to do this without overusing memory, so there are only a few benefits.

PR #18631, opened 2023-12-11
MinIO Console security headers

pjuarezd:
…

## Description

New config section `browser` to store Browser (console) dynamic settings

```
mc admin config get ALIAS
KEYS:
...
browser               manage Browser HTTP specific features, such as Security headers, etc.
...

mc admin config get ALIAS
browser csp_policy="default-src 'self' 'unsafe-eval' 'unsafe-inline';" hsts_seconds=0 hsts_include_subdomains=on hsts_preload=on referrer_policy=strict-origin-when-cross-origin
```

Allow set the values  for response headers:
* Content-Security-Policy
* Strict-Transport-Security
* Refferer-Policy

Read settings from env variables:
* MINIO_BROWSER_CONTENT_SECURITY_POLICY
* MINIO_BROWSER_HSTS_SECONDS
* MINIO_BROWSER_HSTS_INCLUDE_SUB_DOMAINS
* MINIO_BROWSER_HSTS_PRELOAD
* MINIO_BROWSER_REFERRER_POLICY

Settings have default values:

`Content-Security-Policy` = `default-src 'self' 'unsafe-eval' 'unsafe-inline';`
`Refferer-Policy` = `strict-origin-when-cross-origin`
`Strict-Transport-Security` not set

This PR depends on https://github.com/minio/madmin-go/pull/255 to be merge first

## Motivation and Context

Application does not respond with security heades, such header help to prevent Cross-Site Scripting attacks or mitigate MITM attacks.

## Types of changes

## Checklist:

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -212,6 +212,7 @@ var SubSystemsSingleTargets = set.CreateStringSet(
 	CallhomeSubSys,
 	DriveSubSys,
 	BatchSubSys,
+	BrowserSubSys,
 )
 
 // Constant separators
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -119,6 +119,7 @@ const (
 	CallhomeSubSys       = madmin.CallhomeSubSys
 	DriveSubSys          = madmin.DriveSubSys
 	BatchSubSys          = madmin.BatchSubSys
+	BrowserSubSys        = madmin.BrowserSubSys
 
 	// Add new constants here (similar to above) if you add new fields to config.
 )
@@ -188,6 +189,7 @@ var SubSystemsDynamic = set.CreateStringSet(
 	StorageClassSubSys,
 	CacheSubSys,
 	BatchSubSys,
+	BrowserSubSys,
 )
 
 // SubSystemsSingleTargets - subsystems which only support single target.
@@ -209,7 +211,6 @@ var SubSystemsSingleTargets = set.CreateStringSet(
 	SubnetSubSys,
 	CallhomeSubSys,
 	DriveSubSys,
-	CacheSubSys,
 	BatchSubSys,
 )
 
--- a/cmd/common-main.go
+++ b/cmd/common-main.go
@@ -178,6 +178,23 @@ func minioConfigToConsoleFeatures() {
 	os.Setenv("CONSOLE_MINIO_REGION", globalSite.Region)
 	os.Setenv("CONSOLE_CERT_PASSWD", env.Get("MINIO_CERT_PASSWD", ""))
 
+	// This section sets Browser (console) stored config
+	if valueSCP := globalBrowserConfig.GetCSPolicy(); valueSCP != "" {
+		os.Setenv("CONSOLE_SECURE_CONTENT_SECURITY_POLICY", valueSCP)
+	}
+
+	if hstsSeconds := globalBrowserConfig.GetHSTSSeconds(); hstsSeconds > 0 {
+		isubdom := globalBrowserConfig.IsHSTSIncludeSubdomains()
+		isprel := globalBrowserConfig.IsHSTSPreload()
+		os.Setenv("CONSOLE_SECURE_STS_SECONDS", strconv.Itoa(hstsSeconds))
+		os.Setenv("CONSOLE_SECURE_STS_INCLUDE_SUB_DOMAINS", isubdom)
+		os.Setenv("CONSOLE_SECURE_STS_PRELOAD", isprel)
+	}
+
+	if valueRefer := globalBrowserConfig.GetReferPolicy(); valueRefer != "" {
+		os.Setenv("CONSOLE_SECURE_REFERRER_POLICY", valueRefer)
+	}
+
 	globalSubnetConfig.ApplyEnv()
 }
 
--- a/cmd/config-current.go
+++ b/cmd/config-current.go
@@ -24,6 +24,8 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/minio/minio/internal/config/browser"
+
 	"github.com/minio/madmin-go/v3"
 	"github.com/minio/minio/internal/config"
 	"github.com/minio/minio/internal/config/api"
@@ -74,6 +76,7 @@ func initHelp() {
 		config.DriveSubSys:          drive.DefaultKVS,
 		config.CacheSubSys:          cache.DefaultKVS,
 		config.BatchSubSys:          batch.DefaultKVS,
+		config.BrowserSubSys:        browser.DefaultKVS,
 	}
 	for k, v := range notify.DefaultNotificationKVS {
 		kvs[k] = v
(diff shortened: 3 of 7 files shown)


cesnietor (approved):

harshavardhana (changes requested):
why does it need to be customizable?

pjuarezd:
> why does it need to be customizable?

CSP is required to further restrict where web resources load from (css, js, etc), specially useful to prevent XSS in MinIO Console.

Is a best practice to set a restrictive CSP, also XSS it's been in OWASP Top 10 for many years and CSP is one of the recommended `Controls`  for XSS, see https://cheatsheetseries.owasp.org/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.html#other-controls

We are also having customers requesting this cause vuln showed up in their 3rd party pentests @harshavardhana

harshavardhana:
> > why does it need to be customizable?
> 
> CSP is required to further restrict where web resources load from (css, js, etc), specially useful to prevent XSS in MinIO Console.
> 
> Is a best practice to set a restrictive CSP, also XSS is been in OWASP Top 10 for many years and CSP is one of the recommended for `Controls` for XSS, see https://cheatsheetseries.owasp.org/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.html#other-controls
> 
> We are also having customers requesting this cause vuln showed up in their 3rd party pentests @harshavardhana

We can set standard headers and be done with it. Why does it need to be customized?

pjuarezd:
> > > why does it need to be customizable?
> > 
> > 
> > CSP is required to further restrict where web resources load from (css, js, etc), specially useful to prevent XSS in MinIO Console.
> > Is a best practice to set a restrictive CSP, also XSS is been in OWASP Top 10 for many years and CSP is one of the recommended for `Controls` for XSS, see https://cheatsheetseries.owasp.org/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.html#other-controls
> > We are also having customers requesting this cause vuln showed up in their 3rd party pentests @harshavardhana
> 
> We can set standard headers and be done with it. Why does it need to be customized?

That won't work for all cases, becasue of third party services like OpenID, LDAP, etc. Anything console need to interact with from the client (browser) direcly, we'll need to know which domains to trust ie: for a XHR call to OpenID

harshavardhana:
> That won't work for all cases, becasue of third party services like OpenID, LDAP, etc. Anything console need to interact with from the client (browser) direcly, we'll need to know which domains to trust ie: for a XHR call to OpenID

Can you provide examples of all the values that can be configured and what is the value that users must set? in those scenarios? 

I don't think adding this like a naked environment variable is going to work, that can take literally any string doesn't make sense.

klauspost:
So it transfers the value from one env var to another. 

Why not just set the correct one from the beginning?

pjuarezd:
> So it transfers the value from one env var to another.
> 
> Why not just set the correct one from the beginning?

We are removing all setted env variables for console and then setting only "the right ones", not my design so I can advocate for it, I am just following the established pattern @klauspost 

https://github.com/minio/minio/blob/4a21dce2b5537e3d2b8840427b6139a8282cad9e/cmd/common-main.go#L212-L217

pjuarezd:
> > That won't work for all cases, becasue of third party services like OpenID, LDAP, etc. Anything console need to interact with from the client (browser) direcly, we'll need to know which domains to trust ie: for a XHR call to OpenID
> 
> Can you provide examples of all the values that can be configured and what is the value that users must set? in those scenarios?

I provided a "starters" example policy as a comment @harshavardhana 
`default-src 'self' 'unsafe-eval' 'unsafe-inline';`

https://github.com/minio/minio/pull/18631/files#diff-4a2ae40b87e0b6d02d439f087737b78754a045535b6c3cf87e2ced69e0c900beR185

Althoug I cannot foresight all scenarios, cause (as the purpose of CSP is,) CSP is to constrain which domain can resources load from, is why we need to allow admin to set this value.

> I don't think adding this like a naked environment variable is going to work, that can take literally any string doesn't make sense.

Which would be a preferred way? ie:  we could be verify the input is a valid CSP, like by using the `mc admin config set` tool

harshavardhana:
> We are removing all setted env variables for console and then setting only "the right ones", not my design so I can advocate for it, I am just following the established pattern @klauspost

That is because the Console's own defaults must be removed, or some user mistakenly set something for CONSOLE_ - we want them to use MINIO_ not CONSOLE_.

harshavardhana:
Is this PR ready, or are we making any more changes? 

- Referrer-Policy Totally fine we can add a default value
- Strict-Transport-Security We will need some UI (CLI and/or web) to allow setting the fields values max-age, includeSubdomains, and preload
- Content-Security-Policy We can set a default policy including all domains MinIO is aware of and a possible artisan way to set a handcrafted value (CLI and web)

There are 3 things that are needed - which one is covered in this PR?

pjuarezd:
> Is this PR ready, or are we making any more changes?

WIP, just rebased to resolve conflicts

pjuarezd:
> Is this PR ready, or are we making any more changes?
> 
> * Referrer-Policy Totally fine we can add a default value
> * Strict-Transport-Security We will need some UI (CLI and/or web) to allow setting the fields values max-age, includeSubdomains, and preload
> * Content-Security-Policy We can set a default policy including all domains MinIO is aware of and a possible artisan way to set a handcrafted value (CLI and web)
> 
> There are 3 things that are needed - which one is covered in this PR?

This is ready @harshavardhana

internal/config/config.go
```diff
@@ -209,7 +211,6 @@ var SubSystemsSingleTargets = set.CreateStringSet(
 	SubnetSubSys,
 	CallhomeSubSys,
 	DriveSubSys,
-	CacheSubSys,
```
harshavardhana, line 212: Wrong to remove this, this is needed and valid.

internal/config/config.go
```diff
@@ -209,7 +211,6 @@ var SubSystemsSingleTargets = set.CreateStringSet(
...
 	BatchSubSys,
```
harshavardhana, line 214: Also missing browser subsys

harshavardhana (changes requested):

internal/config/config.go
pjuarezd: Okay, but `CacheSubSys` seems to be repeated in this list, look at lines 192 and 212, is that expected?:
https://github.com/minio/minio/blob/c1cae51fb52c4dfa3a4ad98dfc034f025d7b9b9e/internal/config/config.go#L194-L214

internal/config/config.go
pjuarezd: Sure!, added

internal/config/config.go
harshavardhana: Oh it's repeated then it's fine.

harshavardhana (approved):

PR #18632, opened 2023-12-11
metrics: Refactor handling of histogram vectors

krisis:

## Description
Refactor handling of metrics of histogram vector type.

## Motivation and Context
Avoid code duplication

## Types of changes

## Checklist:

### DIFF
--- a/cmd/metrics-v2.go
+++ b/cmd/metrics-v2.go
@@ -1598,47 +1598,64 @@ func getGoMetrics() *MetricsGroup {
 	return mg
 }
 
+// Read prometheus histogram data from ch and convert it to internal Metric
+// type values
+func promHistogramToMetric(desc MetricDescription, ch <-chan prometheus.Metric) (metrics []Metric) {
+	for promMetric := range ch {
+		dtoMetric := &dto.Metric{}
+		err := promMetric.Write(dtoMetric)
+		if err != nil {
+			logger.LogIf(GlobalContext, err)
+			return metrics
+		}
+
+		h := dtoMetric.GetHistogram()
+		for _, b := range h.Bucket {
+			labels := make(map[string]string)
+			for _, lp := range dtoMetric.GetLabel() {
+				labels[*lp.Name] = *lp.Value
+			}
+			labels["le"] = fmt.Sprintf("%.3f", *b.UpperBound)
+			metric := Metric{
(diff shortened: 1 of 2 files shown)


cmd/metrics-v2.go
```diff
@@ -1598,47 +1598,64 @@ func getGoMetrics() *MetricsGroup {
...
+		defer wg.Done()
+		// Converts metrics received into internal []Metric type
+		metrics = promHistogramToMetric(desc, ch)
+	}()
+
+	// Collects prometheus metrics from hist and sends it over ch
+	hist.Collect(ch)
+
+	close(ch)
+	wg.Wait()
```
klauspost: I feel this could be cleaner.

Maybe do  `go func() {defer close(ch); hist.Collect(ch)}` and avoid the waitgroup altogether and have the main thread collect the metrics.

krisis: Good point. With that simplification, I didn't find any reason to split the code into two functions like I had earlier.

harshavardhana (approved):

cmd/metrics-v2.go
```diff
@@ -1598,47 +1598,50 @@ func getGoMetrics() *MetricsGroup {
...
+		hist.Collect(ch)
+	}()
+
+	// Converts metrics received into internal []Metric type
+	var metrics []Metric
+	for promMetric := range ch {
+		dtoMetric := &dto.Metric{}
+		err := promMetric.Write(dtoMetric)
+		if err != nil {
+			logger.LogIf(GlobalContext, err)
```
vadmeste, line 1619: will we have a leaked goroutine here if we have this error ? (I am not sure if this error out in the first place)

vadmeste (changes requested):

cmd/metrics-v2.go
harshavardhana: We can actually ignore this error and continue to write the rest and not return an error here.

krisis: @vadmeste yes, good catch! 
@harshavardhana yes, I think that's a simple solution for an extremely rare case. I shall update my PR to address this.

harshavardhana (approved):

vadmeste (approved):
LGTM

PR #18656, opened 2023-12-14
Verify that remote target bucket is on MinIO server for bucket replication 

taran-p:

## Description
Adds an initial health check when adding a remote target for bucket replication. This ensures the target is a MinIO server.

## Motivation and Context
Non-MinIO servers are allowed to be added as remote targets, which is not supported

## Types of changes

## Checklist:


taran-p:
@harshavardhana This is a quick replacement for minio/mc#4747 , I just added a health check on the server side rather than on `mc`'s side. Should this be hitting an authenticated API instead or does this work?

harshavardhana (changes requested):
Is there some code missing ? I don't see any relevant code

cmd/bucket-targets.go
```diff
@@ -351,6 +351,25 @@ func (sys *BucketTargetSys) SetTarget(ctx context.Context, bucket string, tgt *m
 			return BucketRemoteTargetNotVersioned{Bucket: tgt.TargetBucket}
 		}
 	}
+
+	// Check if target is a MinIO server and alive
+	hcCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
```
harshavardhana: This is too long, keep it like 3 secs, it is more than enough. Anywhere where latency is bigger than 3 seconds then it's a unusable cluster anyway.

We can even indicate to the user that their remote target doesn't respond to the expected 3 sec response time.

This way they know what is expected from a replication target.

harshavardhana (changes requested):

taran-p:
Changed the timing and error. Is there anything else to clarify?

harshavardhana (approved):

harshavardhana (approved):

cmd/bucket-targets.go
```diff
@@ -351,6 +351,25 @@ func (sys *BucketTargetSys) SetTarget(ctx context.Context, bucket string, tgt *m
...
+	result := <-sys.hcClient.Alive(hcCtx, madmin.AliveOpts{}, madmin.ServerProperties{
+		Endpoint: tgt.Endpoint,
+		Scheme:   scheme,
+	})
+	cancel()
+	if result.Error != nil {
+		return RemoteTargetConnectionErr{Bucket: tgt.TargetBucket, Err: result.Error, AccessKey: tgt.Credentials.AccessKey}
+	}
+	if !result.Online {
+		return BucketRemoteTargetNotAlive{Bucket: tgt.TargetBucket}
```
vadmeste: It seems to me you don't have to define a new error type, you can still use RemoteTargetConnectionErr for simplicity.

vadmeste (commented):
One comment

cmd/bucket-targets.go
```diff
@@ -351,6 +351,25 @@ func (sys *BucketTargetSys) SetTarget(ctx context.Context, bucket string, tgt *m
...
 		}
 	}
+
+	// Check if target is a MinIO server and alive
+	hcCtx, cancel := context.WithTimeout(ctx, 3*time.Second)
+	scheme := "http"
+	if tgt.Secure {
+		scheme = "https"
+	}
+	result := <-sys.hcClient.Alive(hcCtx, madmin.AliveOpts{}, madmin.ServerProperties{
```
vadmeste, line 362: It looks like madmin.ServerProperties is not needed as well; hcClient has already the URL to ping.

taran-p: hcClient would not have been given the URL of the target replication bucket. The function did not work properly when testing without the ServerProperties

poornas (approved):

PR #18686, opened 2023-12-20
allow pre-allocating buffers to reduce frequent GCs during growth

harshavardhana:

## Description
allow pre-allocating buffers to reduce frequent GCs during growth

## Motivation and Context
This PR also increases per node bpool memory from 1024 entries 
to 2048 entries; along with that, it also moves the byte pool 
centrally instead of being per pool.

## How to test this PR?
Nothing special this is not enabled by default

## Types of changes

## Checklist:

### DIFF
--- a/cmd/server-main.go
+++ b/cmd/server-main.go
@@ -74,6 +74,12 @@ var ServerFlags = []cli.Flag{
 		EnvVar: "MINIO_LISTENERS",
 		Hidden: true,
 	},
+	cli.BoolFlag{
+		Name:   "pre-allocate",
+		Usage:  "Number of 1MiB sized buffers to pre-allocate. Default 2048",
+		EnvVar: "MINIO_PRE_ALLOCATE",
+		Hidden: true,
+	},
 	cli.StringFlag{
 		Name:   "console-address",
 		Usage:  "bind to a specific ADDRESS:PORT for embedded Console UI, ADDRESS can be an IP or hostname",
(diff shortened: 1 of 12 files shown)


klauspost:
Usually garbage collecting is the expensive part. But pre-allocating will make it seem more "stable" in terms of memory usage, so that is a reasonable bonus.

Maybe keep 1024 buffers for SNSD?

cmd/server-main.go
```diff
@@ -74,6 +74,12 @@ var ServerFlags = []cli.Flag{
 		EnvVar: "MINIO_LISTENERS",
 		Hidden: true,
 	},
+	cli.BoolFlag{
+		Name:   "pre-allocate",
+		Usage:  "pre-allocate 2048 - 1MiB sized buffers",
```
klauspost:
```suggestion
		Usage:  "Number of 1MiB sized buffers to pre-allocate. Default 2048",
```

internal/bpool/bpool.go
```diff
@@ -34,6 +34,22 @@ func NewBytePoolCap(maxSize int, width int, capwidth int) (bp *BytePoolCap) {
...
+			// create new buffer
+			if bp.wcap > 0 {
+				b = make([]byte, bp.w, bp.wcap)
+			} else {
+				b = make([]byte, bp.w)
+			}
+			bp.Put(b)
+		}
+	}()
+}
```
klauspost: No need for goroutines - you are wastly overestimatting the cost of allocations.

You can allocate a single slice and cut it into N slices. Even easier, use [AllocAligned](https://pkg.go.dev/github.com/klauspost/reedsolomon#AllocAligned), which also gives all slices and also makes them that are cacheline size aligned.

harshavardhana: > You can allocate a single slice and cut it into N slices. Even easier, use [AllocAligned](https://pkg.go.dev/github.com/klauspost/reedsolomon#AllocAligned), which also gives all slices and also makes them that are cacheline size aligned.

That requires a complete change in the buffers @klauspost i.e [][]byte vs []byte

klauspost (approved):
LGTM

PR #18703, opened 2023-12-22
server-info: Avoid initializing audit/log http/kafka targets

vadmeste:

## Description
This can cause unnecessary ServerInfo() call delay.

## Types of changes

## Checklist:

### DIFF
--- a/internal/logger/targets.go
+++ b/internal/logger/targets.go
@@ -23,7 +23,6 @@ import (
 	"strings"
 	"sync"
 
-	"github.com/minio/madmin-go/v3"
 	"github.com/minio/minio/internal/logger/target/http"
 	"github.com/minio/minio/internal/logger/target/kafka"
 	"github.com/minio/minio/internal/logger/target/types"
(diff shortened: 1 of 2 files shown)


internal/logger/targets.go
```diff
@@ -57,18 +56,6 @@ var (
 	consoleTgt Target
 )
 
-// TargetStatus returns status of the target (online|offline)
-func TargetStatus(ctx context.Context, h Target) madmin.Status {
-	if h.IsOnline(ctx) {
-		return madmin.Status{Status: string(madmin.ItemOnline)}
-	}
-	// Previous initialization had failed. Try again.
-	if e := h.Init(ctx); e == nil {
```
harshavardhana, line 66: I think we could have simply did Init() in a goroutine and then returned offline anyway?

harshavardhana (changes requested):
Allows to retain the ability to reconnect during an Info() call. Of course we should do this only once via sync.Once()

internal/logger/targets.go
vadmeste: Yes, this starts to be confusing. It looks like we need a better contract with the target interface.

But for now, Init() inside ServerInfo() is not needed. Sending a message in both http and kafka target will reinitialize the target if needed and will mark the target online if sending the message is successful. However the first 'mc admin info' will not show the target online because sending the audit event is asynchrounous, but it will be fine in the next call.

Since the code initializes the target during the boot up, I don't mind the users seeing offline sometimes to investigate what is going on anyway rather than complicating our code.

harshavardhana: Okay that's fine, I thought about it. this is not needed.

harshavardhana (approved):

PR #18717, opened 2023-12-27
Upgrade Console to v0.44.0

dvaldivia:
Since console moved it's api package from `restapi` to `api` this also updates the import paths to `consoleapi` to keep it clear what api is being used


cmd/api-router.go
```diff
@@ -23,7 +23,7 @@ import (
 	"net/http"
 
 	"github.com/klauspost/compress/gzhttp"
-	"github.com/minio/console/restapi"
+	consoleapi "github.com/minio/console/api"
```
harshavardhana, line 26:
```suggestion
	console "github.com/minio/console/api"
```

Just this is fine ? @dvaldivia keeps it simple to use.

harshavardhana (changes requested):

cmd/api-router.go
dvaldivia: We would have conflicts with `github.com/minio/pkg/v2/console` in some files

harshavardhana: Ah I see then it's fine. That package needs a rename.

harshavardhana (approved):

PR #18725, opened 2023-12-31
choose max drives per set to be '8' by default

harshavardhana:

## Description
choose max drives per set to be '8' by default

## Motivation and Context
with this, we also change the default parity for '8'
to be 'EC:3' instead of 'EC:4'.

## How to test this PR?
This is a new change.

## Types of changes

## Checklist:


ravindk89:
This is a big change @harshavardhana 

So we still support erasure set sizes of 9-16, but we are defaulting to a max of 8 unless specified otherwise?

Plus Standard EC:3 with max set size instead of EC:4.

Can you walk us through the logic/goal state here and what we gain/lose in this setup? We'll have a lot of guidance to redo.

harshavardhana:
> This is a big change @harshavardhana
> 
> So we still support erasure set sizes of 9-16, but we are defaulting to a max of 8 unless specified otherwise?
> 
> Plus Standard EC:3 with max set size instead of EC:4.
> 
> Can you walk us through the logic/goal state here and what we gain/lose in this setup? We'll have a lot of guidance to redo.

Implications are same as if you set manually storage class standard to EC:3, and use SET_DRIVE_COUNT environment to 8.

Nothing else changes other than this @ravindk89 - instead of manually doing it we are simply agreeing as per "design" that by default we will never do wider stripes.

With our experience over last few years we do not see the value in it anymore. Also reducing stripe quite simply increases IOPs by 2x.

Also increases efficiency of the scanner, everything is more spread out and still provides a good amount of safety net. 

If you do not trust your hardware then you can simply set higher stripe count, with reduced performance.

ravindk89:
Gotcha. We'll have to prep a few things.

A few follow-ups:

Assuming we continue to follow pool expansion logic.

So if I have pool 1 with stripe size of 16 and EC:8, the new pool should also require at least EC:8 and therefore stripe 8, correct? Even with this setting that is.

klauspost:
@harshavardhana Will it prefer 5 over 10, 6 over 12, 7 over 14 - or will this just affect sizes divisible by 16?

harshavardhana:
> @harshavardhana Will it prefer 5 over 10, 6 over 12, 7 over 14 - or will this just affect sizes divisible by 16?

yeah, well, it does right now. That's the question I have that needs clarification, but it is going to be a bit of a hoch-poch to digest after years of running things differently.

klauspost:
Yeah.. Like I get it for 16 -> 8 and 14 -> 7.... But 10 -> 5 and maybe 12 ->6 seems like a rather big drop...

harshavardhana:
This PR will move to EOS

PR #18732, opened 2024-01-02
NEW API: GetObjectAttributes

zveinn:
## Related PRs
- [pkg](https://github.com/minio/pkg/pull/91)
- [minio-go](https://github.com/minio/minio-go/pull/1921)
- [minwall](https://github.com/miniohq/minwall/pull/66)

## About
 [GetObjectAttributes](https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObjectAttributes.html) is a combination of `HeadObject` and `ListParts`.
Within `minio-go` `HeadObject` is represented by [StatObject](https://min.io/docs/minio/linux/developers/go/API.html#StatObject) and `ListParts` is represented by [ListObjectParts](https://github.com/minio/minio-go/blob/56d99496823ac4501d213c0dd3c1d4db5d65535f/core.go#L118). 
We however did not need to use `ListObjectParts` since `StatObject` functions like a combination of `HeadObject` and `ListParts`. 
We were able to extract all required information by just using `StatObject`.

## Notes
This is my first time integrating an S3 API into `minio`, so I would appriciate a high level of scrutiny. 
 There are a few things that are present in the [headObjectHandler](https://github.com/minio/minio/blob/8a0ba093dd0530ec07e3930354078d0ee30b4762/cmd/object-handlers.go#L725) that were not implemented in the `GetObjectAttributes` API:
 - Caching layer
 - Archive support which can be seen in [HeadObjectHandler](https://github.com/minio/minio/blob/8a0ba093dd0530ec07e3930354078d0ee30b4762/cmd/object-handlers.go#L1068C4-L1068C4)
 - Proxy layer calls during active-active replication
If any of these are relevant for this API then I would appriciate some assistance from someone that understands these
 mechanisms better. 

## Things to consider
Some error returns are not compatible with Amazons `GetObjectAttributes` API, but should be consistent with other `minio` APIs. 

## Post PR work
 - [minio-go documentation update](https://github.com/minio/docs/issues/1101)
 - `GetObjectAttributes` has not been implemented in mc (not sure we will do that though).
 - I will also be updating [Minwall](https://github.com/miniohq/minwall/pull/66)

## Testing
The testing for this new API is done using the [functional test suite](https://github.com/minio/minio-go/blob/master/functional_tests.go) in the `minio-go` repository.
Running the suite will use the `play.min.io` credentials by default, which will not work since this API has not been deployed on that environment. 

Including or not including `s3:ListBucket` can change the error response when objects/versions are not found. 
Details can be found under "permissions" within the [S3 documentation](https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObjectAttributes.html).

### Set up minio with HTTPS
To run all the tests related to `GetObjectAttributes` we need HTTPS enabled for `minio`. 
The best way to do that locally is to use [certgen](https://min.io/docs/minio/linux/operations/network-encryption.html) and the `--certs-dir` flag.

### Create Policies
You need to create a policy which contains `S3:GetObjectAttributes` and `S3:GetObjectVersionAttributes`.
```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "s3:ListBucket",
                "s3:GetObject",
                "s3:GetObjectVersion",
                "s3:GetObjectAttributes",
                "s3:GetObjectVersionAttributes"
            ],
            "Resource": [
                "arn:aws:s3:::*"
            ]
        }
    ]
}

```

### Modify the functional_tests.go file
Modification of the `functional_test.go` file is required since it is designed to run against an internal testing site.
<br/> Modify the main function so that the three functions related to `GetObjectAttributes` API are placed above the `isfullMode()` check. 
<br/> Remember to include an `os.Exit(1)` so you don't run the entire suite.
```go
	testGetObjectAttributes()
	testGetObjectAttributesErrorCases()
	testGetObjectAttributesSSECEncryption()
	os.Exit(1)

	// execute tests
	if isFullMode() {
```

### Set environment variables and run the tests
```
export SERVER_ENDPOINT=127.0.0.1:9000
export ACCESS_KEY=[user_with_relevant_policies]
export SECRET_KEY=[user_secret_key]
export ENABLE_HTTPS=true
export MINT_MODE=true
export SKIP_CERT_VALIDATION=true

go run functional_tests.go
```

## Types of changes
- [ x ] New feature (non-breaking change which adds functionality)


cmd/api-errors.go
```diff
@@ -430,6 +430,9 @@ const (
 	ErrLambdaARNInvalid
 	ErrLambdaARNNotFound
 
+	// // New Codes for GetObjectAttributes and GetObjectVersionAttributes
```
klauspost:
```suggestion
	// New Codes for GetObjectAttributes and GetObjectVersionAttributes
```

cmd/object-api-interface.go
```diff
@@ -69,6 +69,9 @@ type ObjectOptions struct {
 	Tagging                 bool // Is only in GET/HEAD operations to return tagging metadata along with regular metadata and body.
 
 	UserDefined         map[string]string   // only set in case of POST/PUT operations
+	ObjectAttributes    map[string]struct{} // Attribute tags defined by the users for the GetObjectAttributes request
+	MaxParts            int                 // used in GetObjectAttributes. Signals how many parts we should return
+	PartNumberMarker    int                 // used in GetObjectAttributes. Signals which
```
klauspost: which?

cmd/object-handlers.go
```diff
@@ -689,6 +689,197 @@ func (api objectAPIHandlers) getObjectHandler(ctx context.Context, objectAPI Obj
...
+				if v.Number <= opts.PartNumberMarker {
+					continue
+				}
+
+				if len(OA.ObjectParts.Parts) == opts.MaxParts {
+					break
+				}
+
+				OA.ObjectParts.NextPartNumberMarker = v.Number
+				objInfo.Parts[i].Checksums = objInfo.decryptChecksums(v.Number)
```
klauspost: Use `func (o *ObjectInfo) decryptPartsChecksums() error`  outside the loop, which should populate `objInfo.[]Parts.Checksums` in a single operation.

cmd/object-handlers.go
```diff
@@ -689,6 +689,197 @@ func (api objectAPIHandlers) getObjectHandler(ctx context.Context, objectAPI Obj
...
+	}
+
+	lastModified := objInfo.ModTime.UTC().Format(http.TimeFormat)
+	w.Header().Set(xhttp.LastModified, lastModified)
+	w.Header().Del(xhttp.ContentType)
+
+	if _, ok := opts.ObjectAttributes[xhttp.Checksum]; ok {
+		chkSums := objInfo.decryptChecksums(0)
+		OA.Checksum = new(objectAttributesChecksum)
+		OA.Checksum.ChecksumCRC32 = strings.Split(chkSums["CRC32"], "-")[0]
```
klauspost:
```suggestion
		// AWS does not appear to append part number on this API call.
		OA.Checksum.ChecksumCRC32 = strings.Split(chkSums["CRC32"], "-")[0]
```

cmd/object-api-datatypes.go
```diff
@@ -612,3 +612,41 @@ type NewMultipartUploadResult struct {
...
+}
+
+type objectAttributesParts struct {
+	IsTruncated          bool
+	MaxParts             int
+	NextPartNumberMarker int
+	PartNumberMarker     int
+	PartsCount           int
+	Parts                []*objectAttributesPart `xml:"Part"`
+}
```
klauspost, line 638:
```suggestion
}

```

cmd/object-handlers.go
```diff
@@ -689,6 +689,197 @@ func (api objectAPIHandlers) getObjectHandler(ctx context.Context, objectAPI Obj
...
+		ObjectName:      object,
+		IsOwner:         reqInfo.Owner,
+		Claims:          reqInfo.Cred.Claims,
+	}) {
+		writeErrorResponse(ctx, w, toAPIError(ctx, outErr), r.URL)
+	} else {
+		writeErrorResponse(ctx, w, errorCodes.ToAPIErr(ErrAccessDenied), r.URL)
+	}
+
+	return true
```
klauspost:
```suggestion
		writeErrorResponse(ctx, w, toAPIError(ctx, outErr), r.URL)
		return true
	} 
	writeErrorResponse(ctx, w, errorCodes.ToAPIErr(ErrAccessDenied), r.URL)
	return true
```

cmd/object-api-interface.go
klauspost:
```suggestion
	PartNumberMarker    int                 // used in GetObjectAttributes. Signals the part number after which results should be returned
```

cmd/object-handlers.go
```diff
@@ -689,6 +689,197 @@ func (api objectAPIHandlers) getObjectHandler(ctx context.Context, objectAPI Obj
 	})
 }
 
+// GetObjectAttributes ...
+// https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObjectAttributes.html#
+func (api objectAPIHandlers) getObjectAttributesHandler(ctx context.Context, objectAPI ObjectLayer, bucket, object string, w http.ResponseWriter, r *http.Request) {
+	if crypto.S3.IsRequested(r.Header) || crypto.S3KMS.IsRequested(r.Header) { // If SSE-S3 or SSE-KMS present -> AWS fails with undefined error
```
klauspost: I am pretty sure this is not how we should behave. I'd expect we should support these. These are fundamental features - and the error is probably just because it wasn't configured on AWS.

cmd/object-handlers.go
```diff
@@ -689,6 +689,197 @@ func (api objectAPIHandlers) getObjectHandler(ctx context.Context, objectAPI Obj
...
+		OA.ObjectParts = new(objectAttributesParts)
+		OA.ObjectParts.PartNumberMarker = opts.PartNumberMarker
+
+		OA.ObjectParts.MaxParts = opts.MaxParts
+		partsLength := len(objInfo.Parts)
+		OA.ObjectParts.PartsCount = partsLength
+
+		if opts.MaxParts > -1 {
+			for i, v := range objInfo.Parts {
+
```
klauspost:
```suggestion
```

cmd/object-handlers.go
```diff
@@ -689,6 +689,197 @@ func (api objectAPIHandlers) getObjectHandler(ctx context.Context, objectAPI Obj
...
+					ChecksumSHA256: objInfo.Parts[i].Checksums["SHA256"],
+					ChecksumCRC32:  objInfo.Parts[i].Checksums["CRC32"],
+					ChecksumCRC32C: objInfo.Parts[i].Checksums["CRC32C"],
+					PartNumber:     objInfo.Parts[i].Number,
+					Size:           objInfo.Parts[i].Size,
+				})
+			}
+		}
+
+		if OA.ObjectParts.NextPartNumberMarker != partsLength {
```
klauspost, line 811: Does this work correctly when continuing a listing and not returning all results?

cmd/object-handlers.go
```diff
@@ -689,6 +689,197 @@ func (api objectAPIHandlers) getObjectHandler(ctx context.Context, objectAPI Obj
...
+			OA.ObjectParts.IsTruncated = true
+		}
+	}
+
+	out, err := xml.Marshal(OA)
```
klauspost: Use a streaming writer to `w`.

klauspost (commented):
Minor stuff...

cmd/object-handlers.go
klauspost: Instead maybe something like this above?
```
				if len(OA.ObjectParts.Parts) == opts.MaxParts {
					OA.ObjectParts.IsTruncated = i == len(objInfo.Parts)-1 
					break
				}
```

cmd/object-handlers.go
zveinn: It seemed to behave similarly to HeadObject so I added these in, but I'll remove them.

cmd/object-handlers.go
zveinn: Yea, StatsObject always returns the full part list so partsLength is always equal to the length of all object. I had multiple functional tests for this before the refactor of the minio-go repo

cmd/object-handlers.go
```diff
@@ -689,6 +689,185 @@ func (api objectAPIHandlers) getObjectHandler(ctx context.Context, objectAPI Obj
...
+		}
+
+		if OA.ObjectParts.NextPartNumberMarker != partsLength {
+			OA.ObjectParts.IsTruncated = true
+		}
+	}
+
+	_, _ = w.Write([]byte(`<?xml version="1.0" encoding="UTF-8"?>`))
+	if err := xml.NewEncoder(w).Encode(OA); err != nil {
+		writeErrorResponseHeadersOnly(w, toAPIError(ctx, err))
```
harshavardhana, line 818: Once we have attempted to write a response, there is no point in looking at the error and writing that error back to the client.

This will simply confuse the response writer, and it ends up logging some strange errors.

You can use `writeResponse()` with mime type XML.

cmd/object-handlers.go
```diff
@@ -689,6 +689,185 @@ func (api objectAPIHandlers) getObjectHandler(ctx context.Context, objectAPI Obj
...
+		AccountName:     reqInfo.Cred.AccessKey,
+		Groups:          reqInfo.Cred.Groups,
+		Action:          policy.ListBucketAction,
+		BucketName:      bucket,
+		ConditionValues: getConditionValues(r, "", reqInfo.Cred),
+		ObjectName:      object,
+		IsOwner:         reqInfo.Owner,
+		Claims:          reqInfo.Cred.Claims,
+	}) {
+		writeErrorResponse(ctx, w, toAPIError(ctx, outErr), r.URL)
```
harshavardhana: When the request is allowed we are writing an error to the client which is empty?

cmd/object-handlers.go
```diff
@@ -689,6 +689,185 @@ func (api objectAPIHandlers) getObjectHandler(ctx context.Context, objectAPI Obj
...
 
+// GetObjectAttributes ...
+func (api objectAPIHandlers) getObjectAttributesHandler(ctx context.Context, objectAPI ObjectLayer, bucket, object string, w http.ResponseWriter, r *http.Request) {
+	opts, valid := getAndValidateAttributesOpts(ctx, w, r, bucket, object)
+	if !valid {
+		return
+	}
+
+	var s3Error APIErrorCode
+	if opts.Versioned {
```
harshavardhana: Instead of this we must only apply this check when versionId is specified. Not when bucket is versioned, since bucket can be suspended as well.

harshavardhana (changes requested):

cmd/object-handlers.go
zveinn: This method is used to detect if we should return a 404 or a 403 depending on if the user has s3:ListBucket permissions. But before that check is done the method will actually check if there is an error at all. 

The current functionality works as intended, however.. since it's hard to process the logic flow at a quick glance I did a re-write. Will push the changes soon.

cmd/object-handlers.go
zveinn: that's how it was before the change to a stream writer in order not to allocate that extra buffer. I'll change it back.

cmd/object-handlers.go
```diff
@@ -689,6 +689,167 @@ func (api objectAPIHandlers) getObjectHandler(ctx context.Context, objectAPI Obj
 	})
 }
 
+// GetObjectAttributes ...
+func (api objectAPIHandlers) getObjectAttributesHandler(ctx context.Context, objectAPI ObjectLayer, bucket, object string, w http.ResponseWriter, r *http.Request) {
+	opts, valid := getAndValidateAttributesOpts(ctx, w, r, bucket, object)
+	if !valid {
+		return
```
vadmeste, line 696: missing writeErrorResponse() ?

cmd/object-handlers.go
```diff
@@ -689,6 +689,167 @@ func (api objectAPIHandlers) getObjectHandler(ctx context.Context, objectAPI Obj
...
+			BucketName:      bucket,
+			ConditionValues: getConditionValues(r, "", reqInfo.Cred),
+			ObjectName:      object,
+			IsOwner:         reqInfo.Owner,
+			Claims:          reqInfo.Cred.Claims,
+		}) {
+			writeErrorResponse(ctx, w, toAPIError(ctx, err), r.URL)
+			return
+		}
+		writeErrorResponse(ctx, w, errorCodes.ToAPIErr(ErrAccessDenied), r.URL)
```
vadmeste: Do we really need this code? I mean wouldn't all permission errors returned by checkRequestAuthType() already ? (or add a comment)

internal/hash/checksum.go
```diff
@@ -240,6 +240,50 @@ func ReadCheckSums(b []byte, part int) map[string]string {
 	return res
 }
 
+// ReadPartCheckSums will read all part checksums from b and return them.
+func ReadPartCheckSums(b []byte) []map[string]string {
+	var res []map[string]string
+	for len(b) > 0 {
+		t, n := binary.Uvarint(b)
+		if n < 0 {
+			break
```
vadmeste, line 248: I think, let's return an error here instead ?

vadmeste (commented):
Few comments

internal/hash/checksum.go
klauspost: No real need. We don't need errors, instead, like the rest it just doesn't add the checksums.

cmd/object-handlers.go
zveinn: The write to w (http.ResponseWriter) is happening within the function. Amazon added some custom fields to the error response and returning that would mean returning 4x variables or 1x variable and 1x new struct customized for this return. That's why writing to w inside the function felt cleaner.

cmd/object-handlers.go
zveinn: you are right, not sure why I didn't opt to use the checkRequestAuthType for that check as well. Will fix.

cmd/object-handlers.go
```diff
@@ -716,23 +716,8 @@ func (api objectAPIHandlers) getObjectAttributesHandler(ctx context.Context, obj
...
-			AccountName:     reqInfo.Cred.AccessKey,
-			Groups:          reqInfo.Cred.Groups,
-			Action:          policy.ListBucketAction,
-			BucketName:      bucket,
-			ConditionValues: getConditionValues(r, "", reqInfo.Cred),
-			ObjectName:      object,
-			IsOwner:         reqInfo.Owner,
-			Claims:          reqInfo.Cred.Claims,
-		}) {
+		s3Error = checkRequestAuthType(ctx, r, policy.ListBucketAction, bucket, object)
```
vadmeste, line 719: I am not sure I fully understand why this is needed, are you copying this from somewhere else ?

internal/hash/checksum.go
vadmeste: oh yes, sorry I did not mean in this particular line; but in general, it is better to log when we have unexpected data, so it will be easier to debug when something bad happens. Wouldn't that be the case here ?

vadmeste (commented):
Few comments

cmd/object-handlers.go
zveinn: Sorry about that.. It's needed because of how the endpoint behaves. 
```
If you have the s3:ListBucket permission on the bucket, Amazon S3 returns an HTTP status code 404 Not Found ("no such key") error.

If you don't have the s3:ListBucket permission, Amazon S3 returns an HTTP status code 403 Forbidden ("access denied") error.
```
[found in the section on permissions](https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObjectAttributes.html)

internal/hash/checksum.go
zveinn: I'm pretty new to the checksum part of our codebase but I'll talk with @klauspost about where/when we could place error logs, assuming they make sense. 

Replying to comment below from @vadmeste 
```
oh yes, sorry I did not mean in this particular line; but in general, it is better to log when we have unexpected data, so it will be easier to debug when something bad happens. Wouldn't that be the case here ? 
```

cmd/object-handlers.go
harshavardhana: It's an weird AWS S3 behavior @vadmeste

harshavardhana (approved):

internal/hash/checksum.go
klauspost: It is kinda non-actionable and not really important...

It is not worth failing over... And users will just be confused about the output.

klauspost (approved):
LGTM

PR #18734, opened 2024-01-03
scanner: Add a config to disable short sleep between objects scan 

vadmeste:

## Description

Add a hidden configuration under the scanner sub section to configure if
the scanner should sleep between two objects scan. The configuration has
only effect when there is no drive activity related to s3 requests or
healing.

By default, the code will keep the current behavior which is doing
sleep between objects.

To forcefully enable the full scan speed in idle mode, you can do this:
   `mc admin config set myminio scanner idle_speed=full`

## Types of changes

## Checklist:

### DIFF
--- a/cmd/xl-storage-disk-id-check.go
+++ b/cmd/xl-storage-disk-id-check.go
@@ -286,8 +286,7 @@ func (p *xlStorageDiskIDCheck) NSScanner(ctx context.Context, cache dataUsageCac
 	}
 
 	weSleep := func() bool {
-		// Entire queue is full, so we sleep.
-		return cap(p.health.tokens) == len(p.health.tokens)
+		return scannerIdleMode.Load() == 0
 	}
 
 	return p.storage.NSScanner(ctx, cache, updates, scanMode, weSleep)
--- a/internal/config/scanner/scanner.go
+++ b/internal/config/scanner/scanner.go
@@ -31,6 +31,9 @@ const (
 	Speed    = "speed"
 	EnvSpeed = "MINIO_SCANNER_SPEED"
 
+	IdleSpeed    = "idle_speed"
+	EnvIdleSpeed = "MINIO_SCANNER_IDLE_SPEED"
+
 	// All below are deprecated in October 2022 and
 	// replaced them with a single speed parameter
 	Delay            = "delay"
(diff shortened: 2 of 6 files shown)


cmd/xl-storage-disk-id-check.go
```diff
@@ -286,8 +286,17 @@ func (p *xlStorageDiskIDCheck) NSScanner(ctx context.Context, cache dataUsageCac
...
+		switch scannerIdleSpeed.Load() {
+		case 0: // full-throttle
+			return false
+		case 1: // sleep as per the scanner config
+			return true
+		default: // it depends if it is SSD or hard disk
+			if p.storage.rotational {
+				return true
+			}
+			return len(p.health.tokens) < cap(p.health.tokens)
```
harshavardhana:
``` suggestion
			return len(p.health.tokens) < cap(p.health.tokens)
```

This means that `1 < 512` will sleep. @vadmeste I think you need to keep this `len(p.health.tokens) == cap(p.health.tokens)` instead - we don't have to sleep for SSDs unless they are overloaded.

harshavardhana (changes requested):

cmd/xl-storage-disk-id-check.go
vadmeste: > This means that 1 < 512 will sleep

yes, this means the drive is doing 511 parallel operations; the disk is very busy

harshavardhana: Oh I see now I get it. Forgot that about channels.

harshavardhana (approved):

internal/config/scanner/scanner.go
```diff
@@ -107,6 +117,17 @@ func LookupConfig(kvs config.KVS) (cfg Config, err error) {
 	default:
 		return cfg, fmt.Errorf("unknown '%s' value", speed)
 	}
+
+	switch idleSpeed := env.Get(EnvIdleSpeed, kvs.GetWithDefault(IdleSpeed, DefaultKVS)); idleSpeed {
+	case "full":
+		cfg.IdleSpeed = 0
+	case "throttled":
+		cfg.IdleSpeed = 1
```
harshavardhana: Let's make this default @vadmeste

harshavardhana (changes requested):

internal/config/scanner/scanner.go
vadmeste: Done

klauspost (approved):
If it keep existing default, this LGTM.

harshavardhana (approved):

PR #18801, opened 2024-01-16
allow renames() for inlined writes without data-dir

harshavardhana:

## Description
allow renames() for inlined writes without data-dir

## Motivation and Context
data-dir not being present is okay. However, we can still 
rely on the `rename()` atomic call instead of relying on the 
`xl.meta` write, which may truncate the file incorrectly.

## How to test this PR?
CI/CD should handle it

## Types of changes

## Checklist:

### DIFF
--- a/cmd/xl-storage.go
+++ b/cmd/xl-storage.go
@@ -2617,67 +2617,53 @@ func (s *xlStorage) RenameData(ctx context.Context, srcVolume, srcPath string, f
 	defer metaDataPoolPut(dstBuf)
 	if err != nil {
 		if legacyPreserved {
-			// Any failed rename calls un-roll previous transaction.
 			s.deleteFile(dstVolumeDir, legacyDataPath, true, false)
 		}
 		return 0, errFileCorrupt
 	}
 
-	if srcDataPath != "" {
-		if err = s.WriteAll(ctx, srcVolume, pathJoin(srcPath, xlStorageFormatFile), dstBuf); err != nil {
-			if legacyPreserved {
-				// Any failed rename calls un-roll previous transaction.
-				s.deleteFile(dstVolumeDir, legacyDataPath, true, false)
-			}
-			return 0, osErrToFileErr(err)
+	if err = s.WriteAll(ctx, srcVolume, pathJoin(srcPath, xlStorageFormatFile), dstBuf); err != nil {
+		if legacyPreserved {
+			s.deleteFile(dstVolumeDir, legacyDataPath, true, false)
 		}
-		diskHealthCheckOK(ctx, err)
+		return 0, osErrToFileErr(err)
+	}
+	diskHealthCheckOK(ctx, err)
 
+	if srcDataPath != "" && len(fi.Data) == 0 && fi.Size > 0 {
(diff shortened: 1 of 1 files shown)


bh4t:
Can someone please review this?

cmd/xl-storage.go
```diff
@@ -2619,68 +2619,56 @@ func (s *xlStorage) RenameData(ctx context.Context, srcVolume, srcPath string, f
...
-		if !fi.Versioned && !fi.Healing() {
-			// Use https://man7.org/linux/man-pages/man2/rename.2.html if possible on unversioned bucket.
-			if err := Rename2(pathutil.Join(srcVolumeDir, srcPath), pathutil.Join(dstVolumeDir, dstPath)); err == nil {
-				return sign, nil
-			} // if Rename2 is not successful fallback.
-		}
+	if !fi.Versioned && !fi.Healing() {
+		// Use https://man7.org/linux/man-pages/man2/rename.2.html if possible on unversioned bucket.
+		if err := Rename2(pathutil.Join(srcVolumeDir, srcPath), pathutil.Join(dstVolumeDir, dstPath)); err == nil {
+			return sign, nil
```
klauspost: Since we are calling directly:
```suggestion
			diskHealthCheckOK(ctx, err)
			return sign, nil
```

klauspost (approved):
LGTM. Didn't test, though.

harshavardhana:
PTAL @vadmeste @poornas @krisis

cmd/xl-storage.go
```diff
@@ -2623,61 +2623,49 @@ func (s *xlStorage) RenameData(ctx context.Context, srcVolume, srcPath string, f
...
-			}
-			s.deleteFile(dstVolumeDir, dstFilePath, false, false)
-			return 0, err
+	// Commit meta-file
+	if err = renameAll(srcFilePath, dstFilePath, dstVolumeDir); err != nil {
+		if legacyPreserved {
+			// Any failed rename calls un-roll previous transaction.
+			s.deleteFile(dstVolumeDir, legacyDataPath, true, false)
 		}
+		s.deleteFile(dstVolumeDir, dstFilePath, false, false)
```
vadmeste: shouldn't this be dstDataPath instead of dstFilePath ?

vadmeste (changes requested):

cmd/xl-storage.go
harshavardhana: Why? You are renaming dstFilePath, dstDataPath is already done above.

vadmeste: > Why? You are renaming dstFilePath, dstDataPath is already done above.

But isn't this xl.meta ? won't we lose old xl.meta content if there is an error with renameAll, such as EIO

harshavardhana: It's in temp path @vadmeste not on actual namespace.

harshavardhana: I think you are right, we should revisit this.

harshavardhana: It must be dstDataPath not dstFilePath @vadmeste updated PTAL.

vadmeste (approved):
LGTM

PR #18813, opened 2024-01-17
Deallocate huge read buffers

klauspost:
## Description

If a message buffer is excessively huge, release it back so it isn't kept around forever.

Saw a couple of ~50MB buffers in a trace, so it seems like some reads can get up there.

## Motivation and Context

Reduce memory kept around

## Types of changes


harshavardhana (approved):

fwessels (approved):
LGTM

PR #18819, opened 2024-01-18
capture I/O in waiting and total tokens in diskMetrics

harshavardhana:

## Description
capture I/O in waiting and total tokens in diskMetrics

## Motivation and Context
This is needed for the subsequent changes
in ServerUpdate(), ServerRestart() etc.

## How to test this PR?
This change mainly captures additional
metrics, for blocked I/O to be responded
back via `DiskInfo()` call for top layer
callers to perform certain maintenance 
activity.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/storage-rest-common.go
+++ b/cmd/storage-rest-common.go
@@ -20,7 +20,7 @@ package cmd
 //go:generate msgp -file $GOFILE -unexported
 
 const (
-	storageRESTVersion       = "v53" // Remove deprecated APIs
+	storageRESTVersion       = "v54" // Add more metrics per drive
 	storageRESTVersionPrefix = SlashSeparator + storageRESTVersion
 	storageRESTPrefix        = minioReservedBucketPath + "/storage"
 )
--- a/cmd/xl-storage-disk-id-check.go
+++ b/cmd/xl-storage-disk-id-check.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2015-2021 MinIO, Inc.
+// Copyright (c) 2015-2024 MinIO, Inc.
 //
 // This file is part of MinIO Object Storage stack
 //
@@ -90,13 +90,13 @@ type xlStorageDiskIDCheck struct {
 	storage      *xlStorage
 	health       *diskHealthTracker
 
-	// diskStartChecking is a threshold above which we will start to check
-	// the state of disks, generally this value is less than diskMaxConcurrent
-	diskStartChecking int
+	// driveStartChecking is a threshold above which we will start to check
+	// the state of disks, generally this value is less than driveMaxConcurrent
+	driveStartChecking int
 
-	// diskMaxConcurrent represents maximum number of running concurrent
+	// driveMaxConcurrent represents maximum number of running concurrent
 	// operations for local and (incoming) remote disk operations.
-	diskMaxConcurrent int
+	driveMaxConcurrent int
 
 	metricsCache timedValue
 	diskCtx      context.Context
@@ -127,8 +127,11 @@ func (p *xlStorageDiskIDCheck) getMetrics() DiskMetrics {
 	}
 
 	// Do not need this value to be cached.
+	diskMetric.TotalTokens = uint32(p.driveMaxConcurrent)
+	diskMetric.TotalWaiting = uint32(p.health.waiting.Load())
 	diskMetric.TotalErrorsTimeout = p.totalErrsTimeout.Load()
 	diskMetric.TotalErrorsAvailability = p.totalErrsAvailability.Load()
+
 	return diskMetric
 }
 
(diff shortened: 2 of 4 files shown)


cmd/storage-rest-common.go
```diff
@@ -20,7 +20,7 @@ package cmd
 //go:generate msgp -file $GOFILE -unexported
 
 const (
-	storageRESTVersion       = "v53" // Remove deprecated APIs
+	storageRESTVersion       = "v54" // Add more metrics per drive
```
klauspost, line 23: I don't think you need to bump version.

If it is just the update to `DiskMetrics` you are good without bump.

cmd/xl-storage-disk-id-check.go
```diff
@@ -1030,8 +1034,9 @@ func (p *xlStorageDiskIDCheck) checkHealth(ctx context.Context) (err error) {
 	// If also more than 15 seconds since last success, take disk offline.
 	t = time.Since(time.Unix(0, atomic.LoadInt64(&p.health.lastSuccess)))
 	if t > maxTimeSinceLastSuccess {
-		if atomic.CompareAndSwapInt32(&p.health.status, diskHealthOK, diskHealthFaulty) {
+		if p.health.status.CompareAndSwap(diskHealthOK, diskHealthFaulty) {
 			logger.LogAlwaysIf(ctx, fmt.Errorf("node(%s): taking drive %s offline, time since last response %v", globalLocalNodeName, p.storage.String(), t.Round(time.Millisecond)))
+			p.health.waiting.Add(1)
```
klauspost, line 1039: Why is this adding to "waiting"? Doesn't seem right to me, that a disk going offline should add to this?

cmd/xl-storage-disk-id-check.go
```diff
@@ -127,8 +127,11 @@ func (p *xlStorageDiskIDCheck) getMetrics() DiskMetrics {
 	}
 
 	// Do not need this value to be cached.
+	diskMetric.TotalTokens = uint32(p.driveMaxConcurrent)
```
klauspost, line 130: You can add `disk.TokensAvailable`, which is `len(p.tokens)`.

cmd/xl-storage-disk-id-check.go
harshavardhana: It is to measure the disk is waiting and error is disk faulty.

Indicating that disk has I/O errors due to blocked/waiting I/O.

Without that we can't tell if faulty disk is due to hung syscall or not.

cmd/storage-rest-common.go
harshavardhana: It is ensure that since DiskInfo() is a tuple and we can unmarshal properly.

cmd/xl-storage-disk-id-check.go
harshavardhana: Let me change it.

cmd/xl-storage-disk-id-check.go
klauspost: Isn't this captured by the state being offline?  It is just the mixing of "offline disks" and "requests waiting for IO" I'd like to avoid.

Otherwise we should probably add a separate counter for "offline disks".

harshavardhana: No just offline is not very useful in deciding if we can send a kill signal to the process or not @klauspost - waiting on I/O on Linux has a behavior that makes the minio process defunct.

So we want to have information here to avoid this.

harshavardhana: I can add a separate counter, but this is good enough since waiting() will always have a non-zero value at the minimum one, even if there are no user calls waiting.

The I/O that is waiting is the I/O performed by the drive monitoring routine.

harshavardhana: We also need to let them know that there are hung syscalls on the disk that can only be cleared after an OS reboot. 

So, it is also an additional observability requirement per drive.

cmd/xl-storage-disk-id-check.go
harshavardhana: No, I am interested in knowing what the configured number of tokens is, I don't really care how many tokens are currently available since that's a useless value.

Not having enough tokens is captured via the number of waiting calls for tokens.

harshavardhana: This would tell us if the disk has "enough tokens" configured that we expect so that we can analyze if the waiting of syscalls is premature or expected behavior.

harshavardhana:
I realize that there are no changes in the code, I have replied to your comments @klauspost please let me know if you want me to rework anything.

klauspost (approved):

poornas (approved):

PR #18834, opened 2024-01-19
grid: Return rejection reason

klauspost:
## Description

When rejecting incoming grid requests fill out the rejection reason and log it once.

This will give more context when startup is failing. Already logged after a retry on caller (and even backwards compatible)

## How to test this PR?

Tested by injecting errors manually.

## Types of changes


harshavardhana (approved):

fwessels (approved):
LGTM

PR #18862, opened 2024-01-24
Add separate grid reconnection mutex

klauspost:
## Description

Give more safety around reconnects and make sure a state change isn't missed. In principle several incoming requests could be processed at the same time under strange network conditions.

Adds separate mutex and doesn't mix in the testing mutex.

## How to test this PR?

Tested with several runs of `λ go test -race -v -count=500` which includes reconnection test.

## Types of changes
- [x] Bug fix/ Safety (non-breaking change which fixes an issue)

### DIFF
--- a/internal/grid/connection.go
+++ b/internal/grid/connection.go
@@ -68,7 +68,8 @@ type Connection struct {
 	id uuid.UUID
 
 	// Remote uuid, if we have been connected.
-	remoteID *uuid.UUID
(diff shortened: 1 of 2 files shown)


harshavardhana (approved):

poornas (approved):

fwessels (commented):
LGTM (but already merged... 😄)

PR #18903, opened 2024-01-29
Improve tracing & notification scalability

klauspost:
## Description

* Perform JSON encoding on remote machines and only forward byte slices.
* Migrate tracing & notification to websockets.

Should improve tracing and notification performance greatly, removing the sender as a processing bottleneck.

We can always move the tracing back to individual connections if it becomes too laggy.

## Motivation and Context

Scalability.

## How to test this PR?

Use `mc admin trace` and `mc watch`. Remember to test other machines than what you are writing to in case of events.

## Types of changes

### DIFF
--- a/cmd/endpoint.go
+++ b/cmd/endpoint.go
@@ -411,6 +411,29 @@ func (l EndpointServerPools) GridHosts() (gridHosts []string, gridLocal string) 
 	return gridHosts, gridLocal
 }
 
+// FindGridHostsFromPeer will returna matching peer from provided peer.
+func (l EndpointServerPools) FindGridHostsFromPeer(peer *xnet.Host) (peerGrid string) {
+	if peer == nil {
+		return ""
+	}
+	for _, ep := range l {
+		for _, endpoint := range ep.Endpoints {
+			if endpoint.IsLocal {
+				continue
+			}
+			host, err := xnet.ParseHost(endpoint.Host)
+			if err != nil {
+				continue
+			}
+
+			if host.String() == peer.String() {
+				return endpoint.GridHost()
+			}
+		}
+	}
+	return ""
+}
+
 // Hostnames - returns list of unique hostnames
 func (l EndpointServerPools) Hostnames() []string {
 	foundSet := set.NewStringSet()
--- a/cmd/endpoint.go
+++ b/cmd/endpoint.go
@@ -411,7 +411,7 @@ func (l EndpointServerPools) GridHosts() (gridHosts []string, gridLocal string) 
 	return gridHosts, gridLocal
 }
 
-// FindGridHostsFromPeer will returna matching peer from provided peer.
+// FindGridHostsFromPeer will return a matching peer from provided peer.
 func (l EndpointServerPools) FindGridHostsFromPeer(peer *xnet.Host) (peerGrid string) {
 	if peer == nil {
 		return ""
--- a/cmd/listen-notification-handlers.go
+++ b/cmd/listen-notification-handlers.go
@@ -123,8 +123,8 @@ func (api objectAPIHandlers) ListenNotificationHandler(w http.ResponseWriter, r 
 
 	// Convert local messages to JSON and send to mergeCh
 	go func() {
-		var buf bytes.Buffer
-		enc := json.NewEncoder(&buf)
+		buf := bytes.NewBuffer(grid.GetByteBuffer()[:0])
+		enc := json.NewEncoder(buf)
 		tmpEvt := struct{ Records []event.Event }{[]event.Event{{}}}
 		for {
 			select {
@@ -137,6 +137,7 @@ func (api objectAPIHandlers) ListenNotificationHandler(w http.ResponseWriter, r 
 				}
 				mergeCh <- append(grid.GetByteBuffer()[:0], buf.Bytes()...)
 			case <-ctx.Done():
+				grid.PutByteBuffer(buf.Bytes())
 				return
 			}
 		}
--- a/cmd/listen-notification-handlers.go
+++ b/cmd/listen-notification-handlers.go
@@ -18,12 +18,14 @@
 package cmd
 
 import (
+	"bytes"
 	"encoding/json"
 	"net/http"
 	"strconv"
 	"time"
 
 	"github.com/minio/minio/internal/event"
+	"github.com/minio/minio/internal/grid"
 	"github.com/minio/minio/internal/logger"
 	"github.com/minio/minio/internal/pubsub"
 	"github.com/minio/mux"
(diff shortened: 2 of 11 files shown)


cmd/endpoint.go
```diff
@@ -411,6 +411,29 @@ func (l EndpointServerPools) GridHosts() (gridHosts []string, gridLocal string)
 	return gridHosts, gridLocal
 }
 
+// FindGridHostsFromPeer will returna matching peer from provided peer.
```
harshavardhana:
```suggestion
// FindGridHostsFromPeer will return a matching peer from provided peer.
```

cmd/listen-notification-handlers.go
```diff
@@ -116,11 +118,31 @@ func (api objectAPIHandlers) ListenNotificationHandler(w http.ResponseWriter, r
...
-	listenCh := make(chan event.Event, globalAPIConfig.getRequestsPoolCapacity()*len(globalEndpoints.Hostnames()))
-
+	mergeCh := make(chan []byte, globalAPIConfig.getRequestsPoolCapacity()*len(globalEndpoints.Hostnames()))
+	localCh := make(chan event.Event, globalAPIConfig.getRequestsPoolCapacity())
+
+	// Convert local messages to JSON and send to mergeCh
+	go func() {
+		var buf bytes.Buffer
+		enc := json.NewEncoder(&buf)
+		tmpEvt := struct{ Records []event.Event }{[]event.Event{{}}}
```
harshavardhana:
```suggestion
		var buf bytes.Buffer
		enc := json.NewEncoder(&buf)
		tmpEvt := struct{ Records []event.Event }{[]event.Event{{}}}
```

you can use `bytebufferpool` here? @klauspost

harshavardhana (changes requested):
LGTM overall my suggestions to use 

- jsoniter instead of Go's json for performance reasons
- also may be bytebufferpool instead of bytes.Buffer

cmd/listen-notification-handlers.go
klauspost: @harshavardhana The `buf bytes.Buffer` is re-used for all writes (this is outside the for loop). But sure, I can init it with a buffer - but it will only save on per RPC call.

Each event uses a pooled buffer. We can avoid the copy, but that would alloc a `bytes.Buffer` and `json.Encoder` for every event.

harshavardhana (approved):

internal/pubsub/pubsub.go
```diff
@@ -96,6 +103,64 @@ func (ps *PubSub[T, M]) Subscribe(mask M, subCh chan T, doneCh <-chan struct{},
...
+	defer ps.Unlock()
+	subChT := make(chan T, 10000)
+	sub := &Sub[T]{ch: subChT, types: Mask(mask.Mask()), filter: filter}
+	ps.subs = append(ps.subs, sub)
+
+	// We hold a lock, so we are safe to update
+	combined := Mask(atomic.LoadUint64(&ps.types))
+	combined.Merge(Mask(mask.Mask()))
+	atomic.StoreUint64(&ps.types, uint64(combined))
+	var encErr atomic.Pointer[error]
```
fwessels: We are keeping track of the error here, but subsequently not doing anything with it (not that I would know what to do with it...).

So we may just as well ignore it (or comment in the code that we're not doing something purposeful with it)

harshavardhana: @klauspost ^^

internal/grid/types.go
```diff
@@ -198,3 +199,133 @@ func (b *Bytes) Msgsize() int {
...
+		(*u)[zb0001] = zb0002
+	}
+	o = bts
+	return
+}
+
+// Msgsize returns an upper bound estimate of the number of bytes occupied by the serialized message
+func (u URLValues) Msgsize() (s int) {
+	s = msgp.MapHeaderSize
+	if u != nil {
```
fwessels, line 321: Can remove nil check -- u passed in by value (not pointer)

internal/grid/types.go
```diff
@@ -198,3 +199,133 @@ func (b *Bytes) Msgsize() int {
...
+
+// UnmarshalMsg implements msgp.Unmarshaler
+func (u *URLValues) UnmarshalMsg(bts []byte) (o []byte, err error) {
+	var zb0004 uint32
+	zb0004, bts, err = msgp.ReadMapHeaderBytes(bts)
+	if err != nil {
+		err = msgp.WrapError(err)
+		return
+	}
+	if (*u) == nil {
```
fwessels: Braces can be removed: `(*u)` --> `*u`

internal/grid/types.go
```diff
@@ -198,3 +199,133 @@ func (b *Bytes) Msgsize() int {
...
+			return
+		}
+		var zb0005 uint32
+		zb0005, bts, err = msgp.ReadArrayHeaderBytes(bts)
+		if err != nil {
+			err = msgp.WrapError(err, zb0001)
+			return
+		}
+		if cap(zb0002) >= int(zb0005) {
+			zb0002 = (zb0002)[:zb0005]
```
fwessels: Braces can be removed: `(zb0002)` --> `zb0002`

internal/pubsub/pubsub.go
harshavardhana:
```suggestion
```

internal/pubsub/pubsub.go
```diff
@@ -96,6 +103,64 @@ func (ps *PubSub[T, M]) Subscribe(mask M, subCh chan T, doneCh <-chan struct{},
...
+			select {
+			case <-doneCh:
+			case v, ok := <-subChT:
+				if !ok {
+					break
+				}
+				buf.Reset()
+				err := enc.Encode(v)
+				if err != nil {
+					encErr.Store(&err)
```
harshavardhana:
```suggestion
```

internal/grid/types.go
harshavardhana:
```suggestion
	if *u == nil {
```

internal/grid/types.go
harshavardhana:
```suggestion
			zb0002 = zb0002[:zb0005]
```

internal/grid/types.go
harshavardhana: I think that's because it's a map, but it's not needed either. `for v := range nil` works fine.

harshavardhana (changes requested):

internal/grid/types.go
harshavardhana: but its harmless overall

harshavardhana (approved):

cmd/peer-rest-client.go
```diff
@@ -618,92 +679,61 @@ func (client *peerRESTClient) LoadTransitionTierConfig(ctx context.Context) erro
...
-		if err := dec.Decode(&ev); err != nil {
-			return
-		}
-		if len(ev.EventVersion) > 0 {
-			select {
-			case listenCh <- ev:
-			default:
-				// Do not block on slow receivers.
-			}
+	_ = st.Results(func(b *grid.Bytes) error {
```
fwessels: Is `_ =` necessary here?

(Note: we're not doing it in line 698)

fwessels (approved):
LGTM (with some minor comments)

cmd/peer-rest-client.go
harshavardhana:
```suggestion
	st.Results(func(b *grid.Bytes) error {
```

fwessels (approved):

PR #18917, opened 2024-01-30
remove checkBucketExist check entirely to avoid fan-out calls

harshavardhana:

## Description
remove checkBucketExist check entirely to avoid fan-out calls

## Motivation and Context
Each Put, List, and Multipart operation heavily relies on 
on making a GetBucketInfo() call to verify whether a bucket 
exists regularly. This has a significant performance cost 
when there are many servers involved.

We did optimize this part by vectorizing the bucket calls; 
however, it's not enough beyond 100 nodes, which becomes 
reasonably visible in terms of performance.

## How to test this PR?
CI/CD has enough tests, its also changing few things that
need attention and careful review.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/erasure-multipart.go
+++ b/cmd/erasure-multipart.go
@@ -61,7 +61,7 @@ func (er erasureObjects) getMultipartSHADir(bucket, object string) string {
 // checkUploadIDExists - verify if a given uploadID exists and is valid.
 func (er erasureObjects) checkUploadIDExists(ctx context.Context, bucket, object, uploadID string, write bool) (fi FileInfo, metArr []FileInfo, err error) {
 	defer func() {
-		if errors.Is(err, errFileNotFound) || errors.Is(err, errVolumeNotFound) {
+		if errors.Is(err, errFileNotFound) {
 			err = errUploadIDNotFound
 		}
 	}()
@@ -71,7 +71,7 @@ func (er erasureObjects) checkUploadIDExists(ctx context.Context, bucket, object
 	storageDisks := er.getDisks()
 
 	// Read metadata associated with the object from all disks.
-	partsMetadata, errs := readAllFileInfo(ctx, storageDisks, minioMetaMultipartBucket,
+	partsMetadata, errs := readAllFileInfo(ctx, storageDisks, bucket, minioMetaMultipartBucket,
 		uploadIDPath, "", false, false)
 
 	readQuorum, writeQuorum, err := objectQuorumFromMeta(ctx, partsMetadata, errs, er.defaultParityCount)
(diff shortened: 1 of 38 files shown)


cmd/erasure-server-pool.go
```diff
@@ -1307,7 +1307,10 @@ func (z *erasureServerPools) ListObjectVersions(ctx context.Context, bucket, pre
 
 	merged, err := z.listPath(ctx, &opts)
 	if err != nil && err != io.EOF {
-		return loi, err
+		if !isErrBucketNotFound(err) {
+			logger.LogOnceIf(ctx, err, "erasure-list-objects-path-"+bucket)
```
klauspost, line 1311: What is this for? Seems like a leftover.

cmd/metacache-server-pool.go
```diff
@@ -313,6 +314,11 @@ func (z *erasureServerPools) listMerged(ctx context.Context, o listPathOptions,
 	}
 
 	if isAllNotFound(errs) {
+		for _, err := range errs {
+			if err != nil && errors.Is(err, errVolumeNotFound) {
+				return err
+			}
+		}
```
klauspost: Potentially a risky change since it is made without quorum check.

cmd/metacache-set.go
```diff
@@ -1083,6 +1084,11 @@ func listPathRaw(ctx context.Context, opts listPathRawOptions) (err error) {
 					errDiskNotFound.Error():
 					atEOF++
 					fnf++
+					// This is a special case, to handle bucket does
+					// not exist situations.
+					if errors.Is(err, errVolumeNotFound) {
+						volumeNotFound = err
+					}
```
klauspost: Maybe count up separately like fnf.

cmd/object-api-input-checks.go
```diff
@@ -40,7 +40,7 @@ func checkDelObjArgs(ctx context.Context, bucket, object string) error {
 // Checks bucket and object name validity, returns nil if both are valid.
 func checkBucketAndObjectNames(ctx context.Context, bucket, object string) error {
 	// Verify if bucket is valid.
-	if !isMinioMetaBucketName(bucket) && s3utils.CheckValidBucketName(bucket) != nil {
+	if !isMinioMetaBucketName(bucket) && s3utils.CheckValidBucketNameStrict(bucket) != nil {
```
klauspost, line 43: Why this change? Seems this could make existing buckets unreachable?

(same for similar change below)

klauspost (commented):
Just some minor stuff.

cmd/object-api-input-checks.go
harshavardhana: Well, this was there for FS mode back in the days when we supported existing buckets; we never allowed the creation of new invalid buckets.

Look at the generic-handlers.go

```
                        // Validate bucket names if it is not empty
                        if bucketName != "" && s3utils.CheckValidBucketNameStrict(bucketName) != nil {
                                if ok {
                                        tc.FuncName = "handler.ValidRequest"
                                        tc.ResponseRecorder.LogErrBody = true
                                }
                                defer logger.AuditLog(r.Context(), w, r, mustGetClaimsFromToken(r))
                                writeErrorResponse(r.Context(), w, errorCodes.ToAPIErr(ErrInvalidBucketName), r.URL)
                                return
                        }
```

cmd/metacache-set.go
harshavardhana: okay I can do  that

cmd/erasure-server-pool.go
harshavardhana: This I copied over from the ListObjects() code @klauspost

cmd/metacache-server-pool.go
harshavardhana: Okay

harshavardhana:
@klauspost fixed some of your comments

klauspost (approved):

cmd/erasure-multipart.go
```diff
@@ -582,6 +580,13 @@ func writeAllDisks(ctx context.Context, disks []StorageAPI, dstBucket, dstEntry
 func (er erasureObjects) PutObjectPart(ctx context.Context, bucket, object, uploadID string, partID int, r *PutObjReader, opts ObjectOptions) (pi PartInfo, err error) {
 	auditObjectErasureSet(ctx, object, &er)
 
+	data := r.Reader
+	// Validate input data size and it can never be less than zero.
+	if data.Size() < -1 {
```
fwessels, line 585: Shouldn't it be `<= -1` or simply `< 0` ?

harshavardhana: no -1 is valid for compressed stream @fwessels

fwessels (approved):
LGTM

vadmeste (approved):
LGTM

PR #18926, opened 2024-01-31
Improve expiration of tiered objects

krisis:

### User facing changes
- Added a new configuration subsystem (`ilm`) to manage number of workers dedicated to expiration and tiering jobs.
```
# To get the current configuration
mc admin config get ALIAS ilm

# To set number of workers for expiration jobs (default is 100)
mc admin config set ALIAS ilm expiration_workers=50
```
- Prometheus Metrics
```
- # HELP minio_node_ilm_expiry_num_workers Number of workers expiring object versions currently
  # TYPE minio_node_ilm_expiry_num_workers gauge

- # HELP minio_node_ilm_expiry_missed_tasks Number of object version expiry missed due to busy system
  # TYPE minio_node_ilm_expiry_missed_tasks counter

(Internal)
- # HELP minio_node_ilm_expiry_missed_freeversions Number of free versions expiry missed due to busy system
 # TYPE minio_node_ilm_expiry_missed_freeversions counter

- # HELP minio_node_ilm_expiry_missed_tierjournal_tasks Number of tier journal entries cleanup missed due to busy system
  # TYPE minio_node_ilm_expiry_missed_tierjournal_tasks counter


```
- Use a shared worker pool for all ILM expiry tasks
- Free version cleanup no longer runs in the scanner goroutine
- Add a free version only if removing the remote object fails
- Add ilm expiry metrics to node namespace

## Types of changes

## Checklist:


harshavardhana (approved):

krisis:
@harshavardhana let's not merge this PR yet. While exploring better ways to make the ILM expiry worker pool dynamically resizeable, I found https://github.com/minio/minio/pull/16083 and realized that ILM expiration too can suffer from this problem. i.e expiring multiple versions of the same object will contend for the locks. I shall update the PR addressing this.

krisis:
@harshavardhana @vadmeste I have updated the PR to use a worker pool model similar to that in bucket-replication. Expiry tasks are distributed to workers based on a hash on their bucket+object or tier_name+tier_object_name as appropriate.

krisis:
@klauspost I have adapted the worker pool abstraction you implemented for bucket replication here - https://github.com/minio/minio/pull/16083. I'd appreciate if you could help reviewing those parts.

harshavardhana:
madmin-go is updated PTAL @krisis

cmd/xl-storage-free-version_test.go
```diff
@@ -228,3 +229,55 @@ func TestFreeVersion(t *testing.T) {
...
+			Index:        1,
+			Distribution: []int{1, 2, 3, 4, 5, 6, 7, 8},
+			Checksums: []ChecksumInfo{{
+				PartNumber: 1,
+				Algorithm:  HighwayHash256S,
+				Hash:       nil,
+			}},
+		},
+		MarkDeleted: false,
+		// DeleteMarkerReplicationStatus: "",
```
shtripat, line 260: Remove commented if not needed

internal/config/api/help.go
```diff
@@ -80,6 +80,12 @@ var (
 			Optional:    true,
 			Type:        "number",
 		},
+		config.HelpKV{
+			Key:         apiExpiryWorkers,
+			Description: `set the number of expiry workers` + defaultHelpPostfix(apiExpiryWorkers),
```
vadmeste, line 85: is this per node ?

cmd/config-current.go
```diff
@@ -711,6 +713,18 @@ func applyDynamicConfigForSubSys(ctx context.Context, objAPI ObjectLayer, s conf
 			return fmt.Errorf("Unable to apply browser config: %w", err)
 		}
 		globalBrowserConfig.Update(browserCfg)
+	case config.ILMSubSys:
+		ilmCfg, err := ilm.LookupConfig(s[config.ILMSubSys][config.Default])
+		if err != nil {
+			return fmt.Errorf("Unable to apply ilm config: %w", err)
+		}
+		if globalTransitionState != nil {
```
vadmeste, line 726: can you add a log when this is nil ?

cmd/config-current.go
```diff
@@ -711,6 +713,18 @@ func applyDynamicConfigForSubSys(ctx context.Context, objAPI ObjectLayer, s conf
...
+			globalTransitionState.UpdateWorkers(ilmCfg.TransitionWorkers)
+		}
+		if globalExpiryState != nil {
```
vadmeste, line 731: Can you add a log when this is nil ?

cmd/bucket-lifecycle.go
```diff
@@ -105,96 +102,266 @@ type expiryTask struct {
...
-func (es *expiryState) close() {
-	es.once.Do(func() {
-		xioutil.SafeClose(es.byDaysCh)
-		xioutil.SafeClose(es.byNewerNoncurrentCh)
-	})
+// enqueueTierJournalEntry enqueues a tier journal entry referring to a remote
+// object corresponding to a 'replaced' object versions. This applies only to
+// non-versioned or version suspended buckets.
+func (es *expiryState) enqueueTierJournalEntry(je jentry) {
+	wrkr := es.getWorkerCh(je.OpHash())
```
vadmeste, line 188: can you elaborte more why we use a hash here ?

vadmeste (commented):
Few comments

cmd/bucket-lifecycle.go
klauspost: Probably the same as replication where we don't want different versions of the same object trying to update at the same time.

cmd/bucket-lifecycle.go
```diff
@@ -105,96 +102,266 @@ type expiryTask struct {
...
 	default:
+		es.stats.missedExpiryTasks.Add(1)
 	}
 }
 
 // enqueueByNewerNoncurrent enqueues object versions expired by
 // NewerNoncurrentVersions limit for expiry.
 func (es *expiryState) enqueueByNewerNoncurrent(bucket string, versions []ObjectToDelete, lcEvent lifecycle.Event) {
+	task := newerNoncurrentTask{bucket: bucket, versions: versions, event: lcEvent}
+	wrkr := es.getWorkerCh(task.OpHash())
```
klauspost, line 237: This and probably similar workers are racy.

The channel returned by `es.getWorkerCh` has no protection against "downscaling" - meaning the channel is closed by `ResizeWorkers` before you can write to it. This is a crash bug.

The best idea I have is to make the "close"  a nil entry instead of actually closing.

cmd/bucket-lifecycle.go
```diff
@@ -105,96 +102,266 @@ type expiryTask struct {
...
+func (es *expiryState) Worker(input <-chan expiryOp) {
+	for {
+		select {
+		case <-es.ctx.Done():
+			return
+		case v, ok := <-input:
+			if !ok {
+				return
+			}
+			switch v := v.(type) {
```
klauspost, line 325: Log unexpected types.

cmd/bucket-lifecycle.go
```diff
@@ -105,96 +102,266 @@ type expiryTask struct {
...
+		objAPI:  objAPI,
+		workers: make([]chan expiryOp, 0, n),
 	}
+	es.ResizeWorkers(n)
+	return es
 }
 
-func initBackgroundExpiry(ctx context.Context, objectAPI ObjectLayer) {
-	globalExpiryState = newExpiryState()
+func (es *expiryState) getWorkerCh(h uint64) chan<- expiryOp {
```
klauspost, line 266: Here is a lockless (for everything but resize) implementation:

```
// expiryState manages all ILM related expiration activities.
type expiryState struct {
	mu      sync.RWMutex
	workers atomic.Pointer[[]chan expiryOp]
}

func (es *expiryState) ResizeWorkers(n int) {
	// Lock to avoid multiple resizes to happen at the same time.
	es.mu.Lock()
	defer es.mu.Unlock()
	var workers []chan expiryOp
	if v := es.workers.Load(); v != nil {
		// Copy to new array.
		workers = append(workers, *v...)
	}

	if n == len(workers) || n < 1 {
		return
	}

	for len(workers) < n {
		input := make(chan expiryOp, 10000)
		workers = append(workers, input)
		go es.Worker(input)
		es.stats.workers.Add(1)
	}

	for len(workers) > n {
		worker := workers[len(workers)-1]
		workers = workers[:len(workers)-1]
		worker <- expiryOp(nil)
		es.stats.workers.Add(-1)
	}
	// Atomically replace workers.
	es.workers.Store(&workers)
}

func (es *expiryState) getWorkerCh(h uint64) chan<- expiryOp {
	w := es.workers.Load()
	if w == nil || len(*w) == 0 {
		return nil
	}
	workers := *w
	return workers[h%uint64(len(workers))]
}

// PendingTasks returns the number of pending ILM expiry tasks.
func (es *expiryState) PendingTasks() int {
	w := es.workers.Load()
	if w == nil || len(*w) == 0 {
		return 0
	}
	var tasks int
	for _, wrkr := range *w {
		tasks += len(wrkr)
	}
	return tasks
}

````

klauspost (changes requested):

cmd/bucket-lifecycle.go
krisis: Thank you for the detailed review and the alternate approach posted above! Now, I see the race between `ResizeWorkers` and any of the `enqueue*` functions. I shall refresh my PR with these changes.

internal/config/api/help.go
krisis: Yes, it is.

cmd/bucket-lifecycle.go
krisis: Yes, it is for the same reason as we do this for replication workers.

harshavardhana (approved):

vadmeste (approved):
LGTM

harshavardhana (approved):

harshavardhana:
Some unit tests failure @krisis

harshavardhana (approved):

PR #18928, opened 2024-01-31
fix: permission checks for editing access keys

donatello:
## Description

With this change, only a user with `UpdateServiceAccountAdminAction` permission is able to edit access keys.

We would like to let a user edit their own access keys, however the feature needs to be re-designed for better security and integration with external systems like AD/LDAP and OpenID.

This change prevents privilege escalation via service accounts.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/admin-handlers-users_test.go
+++ b/cmd/admin-handlers-users_test.go
@@ -206,6 +206,7 @@ func TestIAMInternalIDPServerSuite(t *testing.T) {
 				suite.TestCannedPolicies(c)
 				suite.TestGroupAddRemove(c)
 				suite.TestServiceAccountOpsByAdmin(c)
+				suite.TestServiceAccountPrivilegeEscalationBug(c)
 				suite.TestServiceAccountOpsByUser(c)
 				suite.TestAddServiceAccountPerms(c)
 				suite.TearDownSuite(c)
@@ -896,14 +897,6 @@ func (s *TestSuiteIAM) TestServiceAccountOpsByUser(c *check) {
 	// 3. Check S3 access
 	c.assertSvcAccS3Access(ctx, s, cr, bucket)
 
-	// 4. Check that svc account can restrict the policy, and that the
-	// session policy can be updated.
-	c.assertSvcAccSessionPolicyUpdate(ctx, s, userAdmClient, accessKey, bucket)
-
-	// 4. Check that service account's secret key and account status can be
-	// updated.
-	c.assertSvcAccSecretKeyAndStatusUpdate(ctx, s, userAdmClient, accessKey, bucket)
-
 	// 5. Check that service account can be deleted.
 	c.assertSvcAccDeletion(ctx, s, userAdmClient, accessKey, bucket)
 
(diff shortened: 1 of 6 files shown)


cmd/iam.go
```diff
@@ -1848,37 +1848,18 @@ func (sys *IAMSys) IsAllowedServiceAccount(args policy.Args, parentUser string)
...
-	}
+	// 3. If an inline session-policy is present, evaluate it.
 
-	if subPolicy.Version == "" {
-		return false
+	// Now check if we have a sessionPolicy.
+	hasSessionPolicy, isAllowedSP := isAllowedBySessionPolicy(args)
+	if hasSessionPolicy {
+		result := isAllowedSP && (isOwnerDerived || combinedPolicy.IsAllowed(args))
+		fmt.Printf("IsAllowedSTS: hasSessionPolicy: %v, isAllowedSP: %v, result: %v\n", hasSessionPolicy, isAllowedSP, result)
```
harshavardhana:
```suggestion
```

harshavardhana (changes requested):
some tests might be needed as well for MinIO as IDP users.

donatello:
Tests updated

cmd/admin-handlers-users_test.go
```diff
@@ -896,13 +897,13 @@ func (s *TestSuiteIAM) TestServiceAccountOpsByUser(c *check) {
 	// 3. Check S3 access
 	c.assertSvcAccS3Access(ctx, s, cr, bucket)
 
-	// 4. Check that svc account can restrict the policy, and that the
-	// session policy can be updated.
-	c.assertSvcAccSessionPolicyUpdate(ctx, s, userAdmClient, accessKey, bucket)
+	// // 4. Check that svc account can restrict the policy, and that the
```
klauspost: Just delete all these. We have git history if we need to find them again.

donatello: Yep.

harshavardhana (approved):

cmd/iam.go
```diff
@@ -1848,37 +1848,16 @@ func (sys *IAMSys) IsAllowedServiceAccount(args policy.Args, parentUser string)
...
-		return isOwnerDerived || combinedPolicy.IsAllowed(parentArgs)
-	}
+	// 3. If an inline session-policy is present, evaluate it.
 
-	if subPolicy.Version == "" {
-		return false
+	// Now check if we have a sessionPolicy.
+	hasSessionPolicy, isAllowedSP := isAllowedBySessionPolicy(args)
+	if hasSessionPolicy {
+		return isAllowedSP && (isOwnerDerived || combinedPolicy.IsAllowed(args))
```
vadmeste: I may be missing something, but shouldn't this be parentArgs ?

cmd/iam.go
```diff
@@ -1848,37 +1848,16 @@ func (sys *IAMSys) IsAllowedServiceAccount(args policy.Args, parentUser string)
...
 	}
 
-	return subPolicy.IsAllowed(parentArgs) && (isOwnerDerived || combinedPolicy.IsAllowed(parentArgs))
+	// Sub policy not set. Evaluate only the parent policies.
+	return (isOwnerDerived || combinedPolicy.IsAllowed(args))
```
vadmeste: I may be missing something, but shouldn't this be parentArgs ?

cmd/iam.go
donatello: You're right, fixing!

vadmeste (approved):
LGTM

PR #18940, opened 2024-02-01
Improve object reuse for grid messages

klauspost:
## Description

Allow internal types to support a `Recycler` interface, which will allow for sharing of common types across handlers.

This means that all `grid.MSS` (and similar) objects are shared across in a common pool instead of a per-handler pool.

Add internal request reuse of internal types. Add for safe (pointerless) types explicitly.

Only log params for internal types. Doing Sprint(obj) is just a bit too messy.

Add a few helpers.

## How to test this PR?

Race testing should catch mistakes.

## Types of changes

### DIFF
--- a/internal/grid/types.go
+++ b/internal/grid/types.go
@@ -27,6 +27,14 @@ import (
 	"github.com/tinylib/msgp/msgp"
 )
 
+// Recycler will override the internal reuse in typed handlers.
+// When this is supported, the handler will not do internal pooling of objects,
+// call Recycle() when then object is no longer needed.
+// The recycler should handle nil pointers.
+type Recycler interface {
+	Recycle()
+}
+
 // MSS is a map[string]string that can be serialized.
 // It is not very efficient, but it is only used for easy parameter passing.
 type MSS map[string]string
@@ -37,6 +45,14 @@ func (m *MSS) Get(key string) string {
 		return ""
 	}
 	return (*m)[key]
+}
+
+// Set a key, value pair.
+func (m *MSS) Set(key, value string) {
+	if m == nil {
+		*m = mssPool.Get().(map[string]string)
+	}
+	(*m)[key] = value
 }
 
 // UnmarshalMsg deserializes m from the provided byte slice and returns the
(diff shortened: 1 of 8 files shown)


internal/grid/types.go
```diff
@@ -27,6 +27,14 @@ import (
 	"github.com/tinylib/msgp/msgp"
 )
 
+// Recycler will override the internal reuse in typed handlers.
+// When this is supported, the handler will not do internal pooling of objects,
+// call Recycle() when then object is no longer needed.
```
fwessels: then --> the

harshavardhana:
```suggestion
// call Recycle() when the object is no longer needed.
```

harshavardhana (changes requested):

harshavardhana (approved):

internal/grid/types.go
```diff
@@ -39,6 +47,14 @@ func (m *MSS) Get(key string) string {
 	return (*m)[key]
 }
 
+// Set a key, value pair.
+func (m *MSS) Set(key, value string) {
+	if m == nil {
+		*m = mssPool.Get().(map[string]string)
```
fwessels, line 53: We are getting from the pool here, shouldn't we be clearing the contents before using it for the first time?

fwessels: Or better yet, call `NewMMS()` ?

internal/grid/types.go
```diff
@@ -121,6 +140,20 @@ func NewMSSWith(m map[string]string) *MSS {
...
+var mssPool = sync.Pool{
+	New: func() interface{} {
+		return make(map[string]string, 5)
+	},
+}
+
+// Recycle the underlying map.
+func (m *MSS) Recycle() {
+	if m != nil && *m != nil {
+		mssPool.Put(map[string]string(*m))
```
fwessels, line 152: Alternatively, clear the contents before we insert back into the pool ?
(so we know there never is any left over content).

fwessels (approved):
LGTM (with one comment regarding use of pool, and clearing contents upon putting back into the pool)

internal/grid/types.go
klauspost: I will clear it. Avoids an alloc.

harshavardhana: Going to merge this, please send a new PR @klauspost

PR #18945, opened 2024-02-01
fix: `null` inline policy handling for access keys

### LINKED ISSUE #18939: Docker: Access Keys effective permissions differ from console output
## Desciption

Creating an access key without manually modifying its permissions results in a key which has no permissions.
This breaks existing keys as well as new keys, if there was no policy assigned to the key manually.

## Expected Behavior
The key should inherit the policies assigned to the user unless restricted otherwise, or at least have the permissions shown in the console be accurate.

## Current Behavior
An access key created under a user assigned the default readwrite policy, gets a 403 response on
`mc ls local/`, even though the console shows the following policy as effective for the key:

```
{
 "Version": "2012-10-17",
 "Statement": [
  {
   "Effect": "Allow",
   "Action": [
    "s3:*"
   ],
   "Resource": [
    "arn:aws:s3:::*"
   ]
  }
 ]
}
```

Possible Solution

1. Setup the docker-compose stack as present in docs/orchestration
2. create an access token (with consoleAdmin or other permissions on your user account)
3. add access token as mc alias 
4. get a 403 on mc ls alias/

## Context
This broke during the bugfix release last night and broke some deployments.

## Regression

YES. Broke in RELEASE-2024-01-31

## Your Environment
```
minio version RELEASE.2024-01-31T20-20-33Z (commit-id=24ecc44bac566b25a5bfa9772d946a19ce551035)
Runtime: go1.21.6 linux/amd64
```
Default docker-compose setup as present in docs/orchestration during the release specified above.

## Images
(All data is localhost throwaway setup)
image
image
image
image
image
image

donatello:
## Description

Interpret `null` inline policy for access keys as inheriting parent policy. Since MinIO Console currently sends this value, we need to honor it for now. A larger fix in Console and in the server are required.

Fixes #18939.

## How to test this PR?

Repeat the steps described in the linked issue.

## Types of changes

## Checklist:
- [x] Fixes a regression (https://github.com/minio/minio/pull/18928)


cmd/iam.go
```diff
@@ -1945,6 +1945,67 @@ func (sys *IAMSys) IsAllowedSTS(args policy.Args, parentUser string) bool {
...
+	// such a case, is a security issue. Ideally, we should not allow such
+	// behavior, but for compatibility with the Console, we currently allow it.
+	//
+	// TODO:
+	//
+	// 1. fix console behavior and allow this inheritance for service accounts
+	// created before a certain (TBD) future date.
+	//
+	// 2. do not allow empty statement policies for service accounts.
+	if subPolicy.Version == "" && subPolicy.Statements == nil && subPolicy.ID == "" {
```
harshavardhana, line 1994: Do we not have something like IsEmpty() ?

harshavardhana (changes requested):

cmd/iam.go
donatello: If the inline policy JSON is given as `{"Statement":[]}`, golang json deserializes as a zero-length empty statement slice (and not a nil slice). The intention of the policy here would be to not allow any actions.

The problem is that `.IsEmpty()` in the policy package only checks for `len(iamp.Statements) == 0`, meaning it will treat this as an empty policy -> so minio will treat the service account as inheriting (and do the opposite of user intention). 

As mentioned, a larger fix is needed to do all this a bit more cleanly, I'm working on it.

harshavardhana (approved):

PR #14447, opened 2022-03-02
Add local disk health checks

klauspost:
## Motivation and Context

The main goal of this PR is to solve the situation where disks stops responding to operations.
This generally causes an FD build-up and eventually will crash the server.

## Description

This adds detection of hung disks, where calls on disk gets stuck.

We add functionality to `xlStorageDiskIDCheck` where it keeps track of the number of concurrent requests on a given disk.

A total number of 100 ongoing operations are allowed. If this limit is reached we will block (but not reject) new requests, but we will monitor the state of the disk.

If no requests have completed or updated without error within a 30 second window, we mark the disk as offline.
Requests that are blocked will be unblocked and return an error. New requests will be rejected until the disk is marked online again. As a safety precaution we never mark a disk offline, where an operation has started within 15 seconds.

Once a disk has been marked offline, a check will run every 5 seconds that will attempt to write and read back a file. As long as this fails the disk will remain offline.

To prevent lots of long-running operations to mark the disk offline we implement a callback feature that allows to update the status as parts of these operations are running.

We add a reader and writer wrapper that will update the status on each successful read/write operation. This should allow fine enough granularity that a slow, but still operational disk will not reach 30 seconds where 50 operations have not progressed.

Note that errors itself is not enough to mark a disk offline. A nil (or io.EOF) error will mark a disk as "good".

Future Improvements:

* Unify the REST server and local xlStorageDiskIDCheck. This would also improve stats significantly.
* Add usage stats, concurrent count, blocked operations, etc.

## How to test this PR?

Testing by inserting an infinite block in `ReadFileStream`, and running `warp get -concurrent=500`.

Stuck ops after running:

```
goroutine profile: total 816
200 @ 0xf1d656 0xf2c5ec 0x31bc6dd 0x3188dbf 0x2e8a62d 0x2f7deb2 0xf4d441
#	0x31bc6dc	github.com/minio/minio/cmd.(*xlStorage).ReadFileStream+0x5c		/minio/cmd/xl-storage.go:1634
#	0x3188dbe	github.com/minio/minio/cmd.(*xlStorageDiskIDCheck).ReadFileStream+0x21e	/minio/cmd/xl-storage-disk-id-check.go:331
#	0x2e8a62c	github.com/minio/minio/cmd.(*streamingBitrotReader).ReadAt+0x16c	/minio/cmd/bitrot-streaming.go:154
[...]

PR #18950, opened 2024-02-02
Fix disk stream deadlocks

klauspost:
## Description

When uploading each stream is given a disk token for `CreateFile`. This is done on all drives in a set and will block for all to respond (regular streaming writes).

Under very high load not all streams will be given a token. So the erasure write cannot write to CreateFile, since not all are ready to accept data.... And the ones that got a token is now blocking for other CreateFile calls to get a token. Therefore it can end up in a situation where nobody can do any writing and nobody is releasing for others to be able to write. This is only resolved when disks are taken offline and the blocked CreateFile calls return.

The solution is to unblock on calls that are waiting on network traffic. This means that disks that aren't waiting for network IO can complete. We add this to `CreateFile` (which can block on incoming data), `WalkDir` (which can block on sending data) and `ReadMultiple` (which can block on sending).

We will only perform this check when under heavy load to avoid impacting regular operations.

## How to test this PR?

Tested using distributed xl setup with `_MINIO_DRIVE_MAX_CONCURRENT=1` set and running mint and `warp mixed concurrent=100`. Also tested with `_MINIO_DRIVE_MAX_CONCURRENT=10`.

## Types of changes
- [x] Fixes a regression - probably all the way back to #14447

### DIFF
--- a/cmd/metacache-walk.go
+++ b/cmd/metacache-walk.go
@@ -392,7 +392,7 @@ func (p *xlStorageDiskIDCheck) WalkDir(ctx context.Context, opts WalkDirOptions,
 	}
 	defer done(&err)
 
-	return p.storage.WalkDir(ctx, opts, wr)
+	return p.storage.WalkDir(ctx, opts, diskReleaseWhileWriting(ctx, wr))
 }
 
 // WalkDir will traverse a directory and return all entries found.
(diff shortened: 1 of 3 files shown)


harshavardhana:
As we have removed the token approach, this PR is no longer needed. However, we will re-open as needed by reverting #18947

PR #18982, opened 2024-02-06
Fix tracing send on closed channel

klauspost:
## Description

Depending on when the context cancelation is picked up the handler may return and close the channel before `SubscribeJSON` returns, causing:

```
Feb 05 17:12:00 s3-us-node11 minio[3973657]: panic: send on closed channel
Feb 05 17:12:00 s3-us-node11 minio[3973657]: goroutine 378007076 [running]:
Feb 05 17:12:00 s3-us-node11 minio[3973657]: github.com/minio/minio/internal/pubsub.(*PubSub[...]).SubscribeJSON.func1()
Feb 05 17:12:00 s3-us-node11 minio[3973657]:         github.com/minio/minio/internal/pubsub/pubsub.go:139 +0x12d
Feb 05 17:12:00 s3-us-node11 minio[3973657]: created by github.com/minio/minio/internal/pubsub.(*PubSub[...]).SubscribeJSON in goroutine 378010884
Feb 05 17:12:00 s3-us-node11 minio[3973657]:         github.com/minio/minio/internal/pubsub/pubsub.go:124 +0x352
```

Wait explicitly for the goroutine to exit.

Bonus: Listen for `doneCh` when sending to not risk getting blocked there is channel isn't being emptied.

## How to test this PR?

`mc admin trace` on distributed setup.. Probably needs some load to trigger.

## Types of changes
- [x] Fixes a regression #18903

### DIFF
--- a/internal/pubsub/pubsub.go
+++ b/internal/pubsub/pubsub.go
@@ -104,7 +104,7 @@ func (ps *PubSub[T, M]) Subscribe(mask M, subCh chan T, doneCh <-chan struct{}, 
 }
 
 // SubscribeJSON - Adds a subscriber to pubsub system and returns results with JSON encoding.
-func (ps *PubSub[T, M]) SubscribeJSON(mask M, subCh chan<- []byte, doneCh <-chan struct{}, filter func(entry T) bool) error {
+func (ps *PubSub[T, M]) SubscribeJSON(mask M, subCh chan<- []byte, doneCh <-chan struct{}, filter func(entry T) bool, wg *sync.WaitGroup) error {
 	totalSubs := atomic.AddInt32(&ps.numSubscribers, 1)
 	if ps.maxSubscribers > 0 && totalSubs > ps.maxSubscribers {
 		atomic.AddInt32(&ps.numSubscribers, -1)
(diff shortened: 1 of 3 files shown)


internal/pubsub/pubsub.go
```diff
@@ -136,7 +142,12 @@ func (ps *PubSub[T, M]) SubscribeJSON(mask M, subCh chan<- []byte, doneCh <-chan
 				if err != nil {
 					break
 				}
-				subCh <- append(GetByteBuffer()[:0], buf.Bytes()...)
+
+				select {
+				case subCh <- append(GetByteBuffer()[:0], buf.Bytes()...):
+				case <-doneCh:
+					break loop
```
harshavardhana:
```suggestion
return
```

We can just return here right?

harshavardhana (changes requested):

internal/pubsub/pubsub.go
klauspost: There is some logic after the loop... I was annoyed as well.... Maybe we should defer that logic.

klauspost: Rewrote it. PTAL

harshavardhana (approved):

harshavardhana:
This PR fixes the `go.mod` replace tag https://github.com/minio/minio/pull/18984

klauspost:
Tested with several runs.

vadmeste (approved):
LGTM

PR #18990, opened 2024-02-07
avoid random startup logs in websocket/grid

harshavardhana:

## Description
avoid random startup logs in Websocket/grid

## Motivation and Context
many of these locks are non-actionable and not functional; get rid of them.

## How to test this PR?
Reduce server logging chatter

## Types of changes

## Checklist:

### DIFF
--- a/internal/grid/connection.go
+++ b/internal/grid/connection.go
@@ -959,7 +959,6 @@ func (c *Connection) handleMessages(ctx context.Context, conn net.Conn) {
 			msg, err = readDataInto(msg, conn, c.side, ws.OpBinary)
 			if err != nil {
 				cancel(ErrDisconnected)
-				logger.LogIfNot(ctx, fmt.Errorf("ws read: %w", err), net.ErrClosed, io.EOF)
 				return
 			}
 			if c.incomingBytes != nil {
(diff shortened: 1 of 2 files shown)


kannappanr (approved):

internal/grid/connection.go
```diff
@@ -970,7 +969,6 @@ func (c *Connection) handleMessages(ctx context.Context, conn net.Conn) {
 			var m message
 			subID, remain, err := m.parse(msg)
 			if err != nil {
-				logger.LogIf(ctx, fmt.Errorf("ws parse package: %w", err))
```
klauspost, line 973: None of these are "startup logs" AFAICT.

You are hiding when a significant system potentially is failing. Of course it is my code, so it will never fail (except the reads-writes).

You are welcome to remove all observability, but it seems important for me to know when disconnects are happening, if I am ever going to debug this.

harshavardhana: Yes generally for that I think we should convert them into a metrics problem, the issue is right now these logs are generated for simple events like node disconnection. 

While they may serve some debug purpose - it is quite simply very noisy right now. I could keep it all non-network errors must be logged?

vadmeste: @harshavardhana I have another PR that separate logs of subsystems. Like replication, peers, grid, s3.. Then in each subsystem log, we can configure whatever we want. But overall, I am in favor of having a DEBUG flag that shows verbose log to identify rare issues. I cannot think of another way.

vadmeste: something like this: https://github.com/minio/minio/pull/19002

It would be easier to control logging if we use subsystem concept

harshavardhana: Okay then that will do

PR #19002, opened 2024-02-07
logging: Add subsystem to log API

vadmeste:

## Description
Create new code paths for multiple subsystems in the code. This will
make maintaing this easier later.

Also introduce bugLogIf() for errors that should not happen in the first
place.

Also add SYSTEM.<subsystem-name> in the logs

## Types of changes

## Checklist:

### DIFF
--- a/cmd/logging.go
+++ b/cmd/logging.go
@@ -1,0 +1,195 @@
+package cmd
+
+import (
+	"context"
+
+	"github.com/minio/minio/internal/logger"
+)
+
+func replLogIf(ctx context.Context, err error, errKind ...interface{}) {
+	logger.LogIf(ctx, "replication", err, errKind...)
+}
+
+func replLogOnceIf(ctx context.Context, err error, id string, errKind ...interface{}) {
+	logger.LogOnceIf(ctx, "replication", err, id, errKind...)
+}
+
+func iamLogIf(ctx context.Context, err error, errKind ...interface{}) {
+	logger.LogIf(ctx, "iam", err, errKind...)
+}
+
+func rebalanceLogIf(ctx context.Context, err error, errKind ...interface{}) {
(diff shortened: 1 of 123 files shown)


cmd/logging.go
```diff
@@ -0,0 +1,191 @@
+package cmd
+
+import (
+	"context"
+
+	"github.com/minio/minio/internal/logger"
+)
+
+func replLogIf(ctx context.Context, err error, errKind ...interface{}) {
```
shtripat, line 9: Should we think of defining sub systems as constants and passing them to the functions `logIf` and `logOnceIf`? This way we will have to maintain fewer functions and calling code would have to remember only the functions and they pass the constant value of sub system.
Adding new sub system would be just adding new constant and callers can start using.

internal/config/identity/plugin/config.go
```diff
@@ -38,6 +38,10 @@ import (
 	xnet "github.com/minio/pkg/v2/net"
 )
 
+func authNLogIf(ctx context.Context, err error) {
```
shtripat, line 41: don't we want these functions to be in common place?

cmd/logging.go
vadmeste: Umm, my point of splitting log to subsystems is to make it easier to manage them later. We have around 800 logs here and any further change will be difficult.

For example, let's say we do not want to print internode errors in the first minute of logging since it will always print something when you start a cluster, then it is enough to fix internode/peer/grid subsystem code.

shtripat: ok

harshavardhana:
Please rebase the PR @vadmeste

harshavardhana:
Please rebase the PR or resubmit @vadmeste

klauspost (approved):

shtripat (approved):

Praveenrajmani (approved):

PR #19004, opened 2024-02-07
Convert more peer peer REST calls

klauspost:
## Description

* Clean up in general.
* Add JSON wrapper.
* Add slice wrapper.
* Add option to make handler return nil error if no connection is given, `IgnoreNilConn`.

Converts the following:

```
+	HandlerGetMetrics
+	HandlerGetResourceMetrics
+	HandlerGetMemInfo
+	HandlerGetProcInfo
+	HandlerGetOSInfo
+	HandlerGetPartitions
+	HandlerGetNetInfo
+	HandlerGetCPUs
+	HandlerServerInfo
+	HandlerGetSysConfig
+	HandlerGetSysServices
+	HandlerGetSysErrors
+	HandlerGetAllBucketStats
+	HandlerGetBucketStats
+	HandlerGetSRMetrics
+	HandlerGetPeerMetrics
+	HandlerGetMetacacheListing
+	HandlerUpdateMetacacheListing
+	HandlerGetPeerBucketMetrics
+	HandlerStorageInfo
+	HandlerConsoleLog
+	HandlerGetLocks
+	HandlerBackgroundHealStatus
+	HandlerGetLastDayTierStats
+	HandlerSignalService
+	HandlerGetBandwidth
```

@harshavardhana We should probably postpone this until after next release. Even though I've tested there could be subtle things.

## How to test this PR?

Most are related to health diag, so `mc support diag` and `mc admin info` apply, but other calls apply as well.

`mc admin logs` appear broken on the current version. Needs https://github.com/minio/madmin-go/pull/269 to work.

## Types of changes


klauspost:
Smells like there is a crash somewhere.

Edit: Yeah.. probably a type assertion. Fixed.

cmd/admin-handlers.go
```diff
@@ -2004,49 +2003,70 @@ func (a adminAPIHandlers) ConsoleLogHandler(w http.ResponseWriter, r *http.Reque
 
 	// Avoid reusing tcp connection if read timeout is hit
 	// This is needed to make r.Context().Done() work as
-	// expected in case of read timeout
+	// expected in case of read timeout (????)
```
harshavardhana: This logic is a bit dated, not sure if its still valid.

cmd/admin-handlers.go
```diff
@@ -2004,49 +2003,70 @@ func (a adminAPIHandlers) ConsoleLogHandler(w http.ResponseWriter, r *http.Reque
...
 	w.Header().Set("Connection", "close")
 
 	setEventStreamHeaders(w)
```
harshavardhana, line 2009: We only need to do this that's all no other HTTP headers needed.

cmd/peer-rest-common.go
```diff
@@ -18,68 +18,29 @@
...
-	peerRESTMethodSysConfig                   = "/sysconfig"
-	peerRESTMethodDeleteBucketMetadata        = "/deletebucketmetadata"
-	peerRESTMethodLoadBucketMetadata          = "/loadbucketmetadata"
-	peerRESTMethodGetBucketStats              = "/getbucketstats"
-	peerRESTMethodGetAllBucketStats           = "/getallbucketstats"
 	peerRESTMethodVerifyBinary                = "/verifybinary"
 	peerRESTMethodCommitBinary                = "/commitbinary"
 	peerRESTMethodSignalService               = "/signalservice"
 	peerRESTMethodBackgroundHealStatus        = "/backgroundhealstatus"
 	peerRESTMethodGetLocks                    = "/getlocks"
```
harshavardhana: These can be converted as well

cmd/peer-rest-common.go
```diff
@@ -18,68 +18,29 @@
...
-	peerRESTMethodGetLocalDiskIDs             = "/getlocaldiskids"
 	peerRESTMethodGetBandwidth                = "/bandwidth"
-	peerRESTMethodGetMetacacheListing         = "/getmetacache"
-	peerRESTMethodUpdateMetacacheListing      = "/updatemetacache"
-	peerRESTMethodGetPeerMetrics              = "/peermetrics"
-	peerRESTMethodGetPeerBucketMetrics        = "/peerbucketmetrics"
-	peerRESTMethodLoadTransitionTierConfig    = "/loadtransitiontierconfig"
 	peerRESTMethodSpeedTest                   = "/speedtest"
 	peerRESTMethodDriveSpeedTest              = "/drivespeedtest"
 	peerRESTMethodReloadSiteReplicationConfig = "/reloadsitereplicationconfig"
```
harshavardhana: and these as well, they do not send any data.

internal/grid/handlers.go
```diff
@@ -409,6 +452,17 @@ func (h *SingleHandler[Req, Resp]) AllowCallRequestPool(b bool) *SingleHandler[R
...
 }
 
+// IgnoreNilConn will ignore nil connections when calling.
+// This will make Call return nil instead of ErrDisconnected when the connection is nil.
+// This may only be set ONCE before use.
+func (h *SingleHandler[Req, Resp]) IgnoreNilConn() *SingleHandler[Req, Resp] {
+	if h.ignoreNilConn {
+		logger.LogOnceIf(context.Background(), fmt.Errorf("%s: IgnoreNilConn called twice", h.id.String()), h.id.String()+"IgnoreNilConn")
+	}
+	h.ignoreNilConn = true
```
harshavardhana, line 466: what is the use-case for this?

harshavardhana (changes requested):

cmd/admin-handlers.go
klauspost: Yeah. Seems outdated.

internal/grid/handlers.go
klauspost: This allows removing the `if conn == nil {return nil}` - hit when grid has not yet been initialized - in the client callers. See the changes in all the existing rpc calls.

cmd/peer-rest-common.go
klauspost: I was hesitant for the speedtests.. but I guess the information only ones would be fine.

klauspost: Took the Reload handler.

harshavardhana (approved):

PR #19009, opened 2024-02-08
optimize startup sequence performance

harshavardhana:

## Description
optimize startup sequence performance

## Motivation and Context
- bucket metadata does not need to look for legacy 
  things if b.Created is non-zero

- stagger bucket metadata loads across lots of nodes to avoid 
  the current thundering herd problem.

- Remove deadlines for RenameData and RenameFile. These calls 
  should never be timed out and wait until completion or for 
  client timeout. Do not choose timeouts for applications during 
  the WRITE phase.

## How to test this PR?
Requires a large enough cluster to validate the changes fully

With the current release, this can go on for 2-3mins
```
minio289.minio.quest:9000 [server-main.go:969:serverMain.func17()] globalBucketMetadataSys.Init (duration: 2m)
```

With all the new changes, here is how much it takes now for a single
bucket metadata load across 300 nodes.
```
minio289.minio.quest:9000 [server-main.go:969:serverMain.func17()] globalBucketMetadataSys.Init (duration: 367.643456ms)
```

## Types of changes

## Checklist:

### DIFF
--- a/cmd/bucket-metadata.go
+++ b/cmd/bucket-metadata.go
@@ -182,26 +182,34 @@ func loadBucketMetadataParse(ctx context.Context, objectAPI ObjectLayer, bucket 
 		b.defaultTimestamps()
 	}
 
-	configs, err := b.getAllLegacyConfigs(ctx, objectAPI)
-	if err != nil {
-		return b, err
+	// If bucket metadata is missing look for legacy files,
+	// since we only ever had b.Created as non-zero when
+	// migration was complete in 2020-May release. So this
+	// a check to avoid migrating for buckets that already
+	// have this field set.
+	if b.Created.IsZero() {
+		configs, err := b.getAllLegacyConfigs(ctx, objectAPI)
+		if err != nil {
+			return b, err
+		}
+
+		if len(configs) > 0 {
+			// Old bucket without bucket metadata. Hence we migrate existing settings.
+			if err = b.convertLegacyConfigs(ctx, objectAPI, configs); err != nil {
+				return b, err
+			}
+		}
 	}
 
-	if len(configs) == 0 {
-		if parse {
-			// nothing to update, parse and proceed.
-			err = b.parseAllConfigs(ctx, objectAPI)
+	if parse {
+		// nothing to update, parse and proceed.
+		if err = b.parseAllConfigs(ctx, objectAPI); err != nil {
+			return b, err
 		}
-	} else {
-		// Old bucket without bucket metadata. Hence we migrate existing settings.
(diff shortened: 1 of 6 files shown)


vadmeste (commented):
LGTM, but I am not sure why we are removing the context timeout

internal/grid/connection.go
```diff
@@ -174,9 +174,9 @@ func (c ContextDialer) DialContext(ctx context.Context, network, address string)
 }
 
 const (
-	defaultOutQueue    = 10000
-	readBufferSize     = 16 << 10
-	writeBufferSize    = 16 << 10
+	defaultOutQueue    = 100000
```
klauspost:
```suggestion
	defaultOutQueue    = 10000
```

This is per remote. I don't think it will make any significant difference - but just take more memory.

Bigger buffers are fine, though.

You are welcome to try slightly larger values of `maxMergeMessage` if you are indeed seeing the queue filled up: https://github.com/minio/minio/blob/960d6040136ef64ab9b2fc6f34d925646f5d206f/internal/grid/grid.go#L48

In my benchmarks 20 seemed to be the sweet spot - but maybe with double the buffer doubling this could also help.

klauspost (commented):
Another thing you can try is to reduce the re-dial wait.

https://github.com/minio/minio/blob/9bcc46d93d628633af7745c7d9426489e118a35b/internal/grid/connection.go#L657

You can try `next := dialStarted.Add(sleep/2)`. This will reduce the wait from 2->4 seconds to 1->2 seconds - both minus the time it took to try the connection.

klauspost:
@vadmeste It adds a default 1 minute timeout on non-streaming RPC calls if none is set on the context. So that is the effect of removing it.

harshavardhana:
> LGTM, but I am not sure why we are removing the context timeout

Because we should not timeout unrollable calls. Those should be driven by client and disk timeouts @vadmeste not network.

vadmeste (approved):

poornas (approved):

cmd/bucket-metadata.go
```diff
@@ -182,26 +182,34 @@ func loadBucketMetadataParse(ctx context.Context, objectAPI ObjectLayer, bucket
...
 	}
 
-	configs, err := b.getAllLegacyConfigs(ctx, objectAPI)
-	if err != nil {
-		return b, err
-	}
+	// If bucket metadata is missing look for legacy files,
+	// since we only ever had b.Created as non-zero when
+	// migration was complete in 2020-May release. So this
+	// a check to avoid migrating for buckets that already
```
fwessels, line 188: is a check

fwessels (approved):
LGTM

PR #19016, opened 2024-02-08
fix incorrect disk io stats in k8s environment

anjalshireesh:

## Description

The previous logic of calculating per second values for disk io stats divides the stats by the host uptime. This doesn't work in k8s environment as the uptime is of the pod, but the stats (from /proc/diskstats) are from the host.

Fix this by storing the initial values of uptime and the stats at the timme of server startup, and using the difference between current and initial values when calculating the per second values.

## Motivation and Context

bugfix

## How to test this PR?

- Deploy a minio cluster in k8s with this change
- Configure prometheus to scrape the resource metrics (`/minio/v2/metrics/resource`)
- Run a warp command to simulate heavy disk activity (e.g. `warp mixed --host=<host>:<port> --access-key=minio --secret-key=minio123 --autoterm --tls --insecure`)
- See the charts in prometheus UI for disk io stat related metrics and verify that they reflect the higher activity because of the warp command (and start reducing after the command execution is over)

## Types of changes

## Checklist:

### DIFF
--- a/cmd/metrics-resource.go
+++ b/cmd/metrics-resource.go
@@ -83,6 +83,11 @@ var (
 	// resourceMetricsHelpMap maps metric name to its help string
 	resourceMetricsHelpMap map[MetricName]string
 	resourceMetricsGroups  []*MetricsGroup
+	// initial values for drives (at the time  of server startup)
+	// used for calculating avg values for drive metrics
+	initialDriveStats   map[string]madmin.DiskIOStats
+	initialDriveStatsMu sync.RWMutex
+	initialUptime       uint64
 )
 
 // PeerResourceMetrics represents the resource metrics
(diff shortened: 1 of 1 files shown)


harshavardhana:
Aren't the disk stats still true whether the pod is up or not? Since the disk is managed by the OS, it is still not k8s. It is quite simply just a bind mount into the pod. 

What I was thinking is why do we need to calculate this based on pod's uptime?

anjalshireesh:
> Aren't the disk stats still true whether the pod is up or not? Since the disk is managed by the OS, it is still not k8s. It is quite simply just a bind mount into the pod.
> 
> What I was thinking is why do we need to calculate this based on pod's uptime?

@harshavardhana We ideally want the host's uptime and it will be simpler if we could get that (e.g. reads/sec = readIOs/uptime), but can't get it from inside the pod. We were earlier dividing by `host.Uptime()` to get the per-second values. But that returns the pod's uptime in k8s and hence give wrong per second values. So we are now moving to this approach where we calculate both the stats and uptime values (current - initial) at pod level and use them to calculate the per second values.

cc @Praveenrajmani

harshavardhana:
> @harshavardhana We ideally want the host's uptime and it will be simpler if we could get that (e.g. reads/sec = readIOs/uptime), but can't get it from inside the pod. We were earlier dividing by `host.Uptime()` to get the per-second values. But that returns the pod's uptime in k8s and hence give wrong per second values. So we are now moving to this approach where we calculate both the stats and uptime values (current - initial) at pod level and use them to calculate the per second values.

shouldn't cgroups be managing I/O for drives on per pod basis?

harshavardhana (approved):

PR #19017, opened 2024-02-08
Fix blocked streams blocking reconnects

klauspost:
## Description

We have observed cases where a blocked stream will block for cancellations.

This happens when response channel is blocked and we want to push an error. This will have the response mutex locked, which will prevent all other operations until upstream is unblocked.

Make this behavior non-blocking and if blocked spawn a goroutine that will send the response and close the output.

Still a lot of "dancing". Added a test for this and reviewed.

## How to test this PR?

Long blocked listings (typically from healing) and network disconnects needed to reproduce. Requests must be cancelled while the stream is blocked.

## Types of changes


vadmeste (approved):
LGTM

harshavardhana (approved):

PR #19021, opened 2024-02-09
honor contexts for keep-alive connections when necessary

harshavardhana:

## Description
honor contexts for keep-alive connections when necessary

## Motivation and Context
it may be that in some situations, we are getting into
a loop when we perpetually wait for the connection between
servers.

This is relatively harder to test, so it needs a flaky setup 
environment with some random packet drops.

## How to test this PR?
Very hard, don't even try. 

## Types of changes

## Checklist:

### DIFF
--- a/cmd/storage-rest-server.go
+++ b/cmd/storage-rest-server.go
@@ -717,7 +717,7 @@ func (s *storageRESTServer) DeleteVersionsHandler(w http.ResponseWriter, r *http
 
 	setEventStreamHeaders(w)
 	encoder := gob.NewEncoder(w)
-	done := keepHTTPResponseAlive(w)
+	done := keepHTTPResponseAlive(r.Context(), w)
 
 	opts := DeleteOptions{}
 	errs := s.getStorage().DeleteVersions(r.Context(), volume, versions, opts)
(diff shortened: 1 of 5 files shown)


cmd/storage-rest-server.go
```diff
@@ -897,10 +900,13 @@ func keepHTTPResponseAlive(w http.ResponseWriter) func(error) {
 			}
 		}
 		defer xioutil.SafeClose(doneCh)
+
 		ticker := time.NewTicker(time.Second * 10)
 		defer ticker.Stop()
 		for {
 			select {
+			case <-ctx.Done():
+				return
```
klauspost, line 909: There is a deadlock here. `doneCh` will block in the returned function. 

So below you will need something like:
```
		select {
			case doneCh <- err:
			case <-ctx.Done(): 
				doneCh = nil
				return
		}
```

`keepHTTPReqResponseAlive` above has a similar issue AFAICT.

klauspost (changes requested):

harshavardhana:
Hmm okay, I think we can avoid this for now. There is no real issue here to fix @klauspost .

PR #19023, opened 2024-02-09
introduce reader deadlines for net.Conn

harshavardhana:

## Description
introduce reader deadlines for net.Conn

## Motivation and Context
Bonus: set "retry-after" header for AWS SDKs if possible to honor them.

## How to test this PR?
You need a flaky client 

```go
package main

import (
        "context"
        "io"
        "log"
        "os"
        "time"

        "github.com/minio/minio-go/v7"
)

type slowReader struct {
        times int
        delay time.Duration
}

func (s *slowReader) Read(b []byte) (n int, err error) {
        s.times--
        if s.times < 0 {
                time.Sleep(s.delay)
                err = io.EOF
        }
        return len(b), erer
}

func main() {
        s3Client, err := minio.New("localhost:9000", &minio.Options{
                Creds:  credentials.NewStaticV4("minioadmin", "minioadmin", ""),
                Secure: false,
        })
        if err != nil {
                log.Fatalln(err)
        }

        s3Client.TraceOn(os.Stdout)

        info, err := s3Client.PutObject(context.Background(), "bucket", "object", &slowReader{delay: 4 * time.Minute, times: 3}, 1<<20, minio.PutObjectOptions{})
        if err != nil {
                log.Fatalln(err)
        }
        log.Println("Uploaded", "my-objectname", " of size: ", info.Size, "Successfully.")
}
```

```
minio server ~/test --conn-client-read-deadline 3s
```

```
go run /tmp/putobject.go
```

```
---------START-HTTP---------
PUT /bucket/object HTTP/1.1
Host: localhost:9000
User-Agent: MinIO (linux; amd64) minio-go/v7.0.66
Content-Length: 1050102
Authorization: AWS4-HMAC-SHA256 Credential=minioadmin/20240209/us-east-1/s3/aws4_request,SignedHeaders=host;x-amz-content-sha256;x-amz-date;x-amz-decoded-content-length,Signature=**REDACTED**
X-Amz-Content-Sha256: STREAMING-AWS4-HMAC-SHA256-PAYLOAD
X-Amz-Date: 20240209T115931Z
X-Amz-Decoded-Content-Length: 1048576
Accept-Encoding: gzip

HTTP/1.1 499 status code 499
Connection: close
Content-Length: 352
Accept-Ranges: bytes
Content-Type: application/xml
Date: Fri, 09 Feb 2024 11:59:34 GMT
Server: MinIO
Strict-Transport-Security: max-age=31536000; includeSubDomains
Vary: Origin
Vary: Accept-Encoding
X-Amz-Id-2: dd9025bab4ad464b049177c95eb6ebf374d3b3fd1af9251148b658df7ac2e3e8
X-Amz-Request-Id: 17B22FFCCF9D73D1
X-Content-Type-Options: nosniff
X-Xss-Protection: 1; mode=block

<Error><Code>ClientDisconnected</Code><Message>Client disconnected before response was ready</Message><Key>object</Key><BucketName>bucket</BucketName><Resource>/bucket/object</Resource><RequestId>17B22FFCCF9D73D1</RequestId><HostId>dd9025bab4ad464b049177c95eb6ebf374d3b3fd1af9251148b658df7ac2e3e8</HostId></Error>
```

## Types of changes

## Checklist:


klauspost:
Great! Checking it out.

klauspost:
So this has the same race condition as the previous versions we did.

You will need to read into a similar sized buffer and copy:

```Go
// DeadlineReader deadline reader with timeout
type DeadlineReader struct {
	io.ReadCloser
	timeout time.Duration
	tmp     []byte
	err     error
}

// NewDeadlineReader wraps a readCloser to make it respect the specified
// deadline value per Read(). If there is a blocking read, the returned
// Reader will return whenever the timer hits
// func() (n = 0, err = context.DeadlineExceeded)
func NewDeadlineReader(r io.ReadCloser, timeout time.Duration) io.ReadCloser {
	return &DeadlineReader{ReadCloser: r, timeout: timeout}
}

func (r *DeadlineReader) Read(buf []byte) (int, error) {
	if r.err != nil {
		return 0, r.err
	}
	if cap(r.tmp) < len(buf) {
		r.tmp = make([]byte, len(buf))
	}
	tmp := r.tmp[:len(buf)]

	n, err := WithDeadline[int](context.Background(), r.timeout, func(ctx context.Context) (int, error) {
		return r.ReadCloser.Read(tmp)
	})
	if err != nil {
		r.err = err
		return 0, err
	}
	copy(buf[:n], tmp[:n])
	return n, err
}

// Close closer interface to close the underlying closer
func (r *DeadlineReader) Close() error {
	return r.ReadCloser.Close()
}
```

Careful about recycling the buffer. `r.ReadCloser.Read(tmp)` MUST have completed before you can recycle it.

harshavardhana:
Not going to use the Reader() @klauspost its not useful, the most simpler is net.Conn deadline.

klauspost (approved):
With separate error it will be great!

internal/http/listener.go
```diff
@@ -74,7 +78,8 @@ func (listener *httpListener) Accept() (conn net.Conn, err error) {
 	select {
 	case result, ok := <-listener.acceptCh:
 		if ok {
-			return result.conn, result.err
+			return deadlineconn.New(result.conn).
```
vadmeste, line 81: @klauspost does grid send pings all the time in all situations ?

I was thinking for long listing in a overloaded system, ListDir() does not use grid yet and probably peer-rest-storage can take more than 3 minutes to give a first response. peer-rest-client will most likley be calling Read() and will be stuck there. So we probably need to increase 3 minutes to a higher value.

I can test this theory by adding fake code.

harshavardhana: you can; I just hoped we had a way to tell if it was an incoming S3 API call or an internode call.

vadmeste: In fact, I might be wrong. This is only a listener READ deadline, and I do not remember a scenario where a server receives a new connection but still need to wait legitimately for a long time to read data. Testing anyway.

vadmeste: I can see grid is complaining without changing anything in the code. 

```
API: SYSTEM()                                                                                  
Time: 14:20:07 UTC 02/09/2024                                                                  
DeploymentID: 9ecbdd07-e7d8-48cc-bcd6-93b32426b606                  
Error: ws read: read tcp 127.0.0.1:45768->127.0.0.1:9002: read: connection reset by peer (*fmt.wrapError)                                                                                      
       2: internal/logger/logger.go:257:logger.LogIfNot()                            
       1: internal/grid/connection.go:962:grid.(*Connection).handleMessages.func1() 
```

klauspost: @vadmeste Connections are monitored and reconnected after ~30s if no traffic flows. This is done by the "write" side on each end. Writes therefore have a timeout.

I can add ListDir to https://github.com/minio/minio/pull/19004 if that is what you are seeing.

krisis (approved):

vadmeste (approved):
LGTM

PR #19025, opened 2024-02-09
automatically generate root credentials with KMS

aead:
## Description
With this commit, MinIO generates root credentials automatically and deterministically if:

 - No root credentials have been set.
 - A KMS (KES) is configured.
 - API access for the root credentials is disabled (lockdown mode).

Before, MinIO defaults to `minioadmin` for both access key and secret key. Now, MinIO does generate unique root credentials automatically on startup using the KMS.

Therefore, it uses the KMS HMAC function to generate pseudo-random values. These values never change as long as the KMS key remains the same, and the KMS key must continue to exist since all IAM data is encrypted with it.

Backwards compatibility:

This commit should not cause existing deployments to break. It only changes the root credentials of deployments that have a KMS configured (KES, not a static key) but have not set any admin credentials. Such deployments should be rare or not exist at all.

Even if, the worst case would be to update to root credentials in `mc` or other clients used to administer the cluster. Root credentials are anyway not intended for regular S3 operations.

## Motivation and Context
KMS, auto-generate root credentials

## How to test this PR?
This PR requires the current latest (master branch) or KES. Older KES servers don't provide the HMAC API.

## Types of changes

## Checklist:


aead:
cc @ravindk89 - This will require a doc update

ravindk89:
A few questions:

- Do we persist the random value to file somewhere, or is it kept solely as a config option (thus solving the plaintext problem)
- If a user sets a manual root user/pass later, how do we handle the en/decryption of backend data?
- Extending the above, if we do support auto en/decryption as part of changing keys, do we want to call that out as being an option for key rotation?
- Extending the above, would there be an API to kick this off if necessary?

I also have to assume that with this setup, if the value is persisted to config securely, we would not risk the cluster shutting down due to KMS going down?

aead:
> Do we persist the random value to file somewhere, or is it kept solely as a config option (thus solving the plaintext problem)

Nothing is persistent here. What happens is, that MinIO calls the HMAC API of the KMS. This returns a pseudo-random value. So, for a different key it returns a different result. However, for the same key it returns the same result. 

> If a user sets a manual root user/pass later, how do we handle the en/decryption of backend data?

There is no connection between the backend data and the root credentials. The IAM data is encrypted using the `GenerateDataKey` and `Decrypt` functions of the KMS. The root credentials are not involved at all. The user can simply set their custom credentials and MinIO will honor it.

> Extending the above, if we do support auto en/decryption as part of changing keys, do we want to call that out as being an option for key rotation? Extending the above, would there be an API to kick this off if necessary?

The only "limitation" of this PR is that if you change your `MINIO_KMS_KES_KEY_NAME` env. variable (which you could but nobody does) then your root credentials change. All other data, like objects, data encryption keys etc. are not affected or touched by this PR. Nothing in this regard changes...

ravindk89:
Perfect - this nicely resolves at least one concern we see regarding the root creds.

Between this and hopefully figuring out `systemd-creds` I think we can nicely close the loop around `/etc/default/minio`.

ravindk89:
> Nothing is persistent here. What happens is, that MinIO calls the HMAC API of the KMS. This returns a pseudo-random value. So, for a different key it returns a different result. However, for the same key it returns the same result.

Ah - so this would imply that we need to maintain KES/KMS uptime, correct? If KES goes down, a node cannot retrieve the PRV, and I assume that might cause some issues.

Which isn't a bad thing, it's just the tradeoff

aead:
Technically, we could use the same approach for all root credentials, not just when they're not set manually. (If KMS is present, use the HMAC API to generate them automatically) However, such behavior must be opt-in since it would break existing deployments.

ravindk89:
> Technically, we could use the same approach for all root credentials, not just when they're not set manually. (If KMS is present, use the HMAC API to generate them automatically) However, such behavior must be opt-in since it would break existing deployments.

Eh, if people set the creds manually then there's no reason to override it. 

It's simple enough to document the need to maintain uptime to keep normal operations. At that point users can decide about the tradeoff vs trusting their protection of `root`, etc.

cmd/common-main.go
```diff
@@ -961,6 +956,40 @@ func handleKMSConfig() {
 			logger.Fatal(err, "Unable to initialize a connection to KES as specified by the shell environment")
 		}
 		GlobalKMS = KMS
+
+		// If no root access/secret key has been provided by the user, we generate new random
+		// access/secret key deterministically using the KMS.
+		if globalActiveCred.AccessKey == auth.DefaultCredentials.AccessKey && globalActiveCred.SecretKey == auth.DefaultSecretKey {
```
harshavardhana:
```suggestion
		if globalActiveCred.Equal(auth.DefaultCredentials) {
```

internal/kms/config.go
```diff
@@ -22,7 +22,6 @@ const (
 	EnvKMSSecretKey      = "MINIO_KMS_SECRET_KEY"
 	EnvKMSSecretKeyFile  = "MINIO_KMS_SECRET_KEY_FILE"
 	EnvKESEndpoint       = "MINIO_KMS_KES_ENDPOINT"     // One or multiple KES endpoints, separated by ','
-	EnvKESEnclave        = "MINIO_KMS_KES_ENCLAVE"      // Optional "namespace" within a KES cluster - not required for stateless KES
```
harshavardhana, line 25: breaking change?

cmd/common-main.go
```diff
@@ -961,6 +956,40 @@ func handleKMSConfig() {
...
+			if manager, ok := GlobalKMS.(kms.KeyManager); ok {
+				aKey, err := manager.HMAC(GlobalContext, kmsConf.DefaultKeyID, []byte("root access key"))
```
harshavardhana: We need `logger.Info()` here to say we generate root credentials based on KMS. 

Does this completely change how root credentials can be used with MinIO? How does the client connect to set up the system for the first time?

IMO, we should only do this if the root credential based on the API call is disabled and KMS is set.

This is the correct flow. A customer would set up the cluster properly and want static credentials to work for a while. Once their setup is done, it is a lockdown stage. 

They configure root disable; now, if KMS is set, we simply use a secret credential that is never exposed instead of minioadmin, and we can suggest that the user simply remove the ROOT envs from the config file.

This can be automated in operator as well. I would call that a `lockdown` flag in operator UI or something. 

So it is a step by step process.

harshavardhana (changes requested):

cmd/kms-handlers.go
```diff
@@ -449,22 +445,8 @@ func (a kmsAPIHandlers) KMSAssignPolicyHandler(w http.ResponseWriter, r *http.Re
...
-		writeErrorResponseJSON(ctx, w, toAdminAPIErr(ctx, err), r.URL)
-		return
-	}
-	err := manager.AssignPolicy(ctx, r.Form.Get("policy"), request.Identity)
-	if err != nil {
-		writeErrorResponseJSON(ctx, w, toAdminAPIErr(ctx, err), r.URL)
-		return
-	}
-	writeSuccessResponseHeadersOnly(w)
+	writeErrorResponseJSON(ctx, w, errorCodes.ToAPIErr(ErrNotImplemented), r.URL)
```
shtripat, line 448: Do we still need above validateAdminReq etc if we return just ErrNotImplemented? Cant we just return the error?

internal/auth/credentials.go
```diff
@@ -209,37 +210,68 @@ func ExpToInt64(expI interface{}) (expAt int64, err error) {
...
-	for i := 0; i < accessKeyMaxLen; i++ {
-		keyBytes[i] = alphaNumericTable[keyBytes[i]%alphaNumericTableLen]
+	return accessKey, secretKey, nil
+}
+
+// GenerateAccessKey returns a new access key generated randomly using
+// the given io.Reader. If random is nil, crypto/rand.Reader is used.
+// If length <= 0, the access key length is chosen automatically.
+//
+// GenerateAccessKey returns an error if length is to small for a valid
```
shtripat:
```suggestion
// GenerateAccessKey returns an error if length is too small for a valid
```

internal/auth/credentials.go
```diff
@@ -209,37 +210,68 @@ func ExpToInt64(expI interface{}) (expAt int64, err error) {
...
+		key[i] = alphaNumericTable[key[i]%alphaNumericTableLen]
+	}
+	return string(key), nil
+}
+
+// GenerateSecretKey returns a new secret key generated randomly using
+// the given io.Reader. If random is nil, crypto/rand.Reader is used.
+// If length <= 0, the secret key length is chosen automatically.
+//
+// GenerateSecretKey returns an error if length is to small for a valid
```
shtripat:
```suggestion
// GenerateSecretKey returns an error if length is too small for a valid
```

internal/kms/kes.go
```diff
@@ -346,34 +338,23 @@ func (c *kesClient) DecryptAll(ctx context.Context, keyID string, ciphertexts []
...
-	return c.enclave.DescribePolicy(ctx, policy)
-}
-
-// AssignPolicy assigns a policy to an identity.
-// An identity can have at most one policy while the same policy can be assigned to multiple identities.
-// The assigned policy defines which API calls this identity can perform.
-// It's not possible to assign a policy to the admin identity.
-// Further, an identity cannot assign a policy to itself.
-func (c *kesClient) AssignPolicy(ctx context.Context, policy, identity string) error {
+// DecryptKey decrypts the ciphertext with the key at the KES
```
shtripat: The function description does not match the function name

internal/kms/config.go
aead: Not really because we introduced it for stateful KES which we never did 😉 KES doesn't have anymore (and never honored it)

cmd/common-main.go
aead: > We need logger.Info() here to say we generate root credentials based on KMS.

Sure 👍 

> Does this completely change how root credentials can be used with MinIO? How does the client connect to set up the system for the first time?

The KMS credentials are still "static" as long as the `MINIO_KMS_KES_KEY_NAME` stays the same. They are just not predictable unless how have access to the KMS. So, they don't change. For first access, the user would either need to obtain them via an API call / command from the KMS or look at the console output.

We could restrict is it to be only effective in lock-down mode. However, I don't see any problem with the current behavior or am I missing something? It's not like the credentials keep rotating or depend on any state at MinIO (apart from the KMS key name). Nevertheless if you prefer to only use KMS generated access/secret keys in lock-down mode (as of now) I can change it.

cmd/kms-handlers.go
aead: We are not just returning an error. If you allow anonymous requests you would allow anyone to identify whether or not a KMS is present (`ErrKMSNotConfigured` vs. `ErrNotImplemented`)

internal/kms/kes.go
aead: 👍

cmd/common-main.go
aead: For example, with the following env. vars MinIO starts like this:

```
MINIO_KMS_KES_ENDPOINT=https://127.0.0.1:7373
MINIO_KMS_KES_KEY_NAME=my-key
MINIO_KMS_KES_CAPATH=/tmp/public.crt
MINIO_KMS_KES_API_KEY=kes:v1:AD9E7FSYWrMD+VjhI6q545cYT9YOyFxZb7UnjEepYDRc
```
```
minio server /tmp/1
Automatically generated root access key and secret key with the KMS
Formatting 1st pool, 1 set(s), 1 drives per set.
WARNING: Host local has all drives of set. A host failure will result in data becoming unavailable.
MinIO Object Storage Server
Copyright: 2015-0000 MinIO, Inc.
Version: DEVELOPMENT.GOGET (go1.22.0 darwin/arm64)

Status:         1 Online, 0 Offline.
S3-API: http://192.168.188.79:9000  http://127.0.0.1:9000
RootUser: P63ITYI6L07GY1VKY3ET
RootPass: 4+yRbGik+i+Yf98T0Jrw+8XtlBAtWjYznRvd0kkH+uE

Console: http://192.168.188.79:53656 http://127.0.0.1:53656
RootUser: P63ITYI6L07GY1VKY3ET
RootPass: 4+yRbGik+i+Yf98T0Jrw+8XtlBAtWjYznRvd0kkH+uE

Command-line: https://min.io/docs/minio/linux/reference/minio-mc.html#quickstart
   $ mc alias set 'myminio' 'http://192.168.188.79:9000' 'P63ITYI6L07GY1VKY3ET' '4+yRbGik+i+Yf98T0Jrw+8XtlBAtWjYznRvd0kkH+uE'

Documentation: https://min.io/docs/minio/linux/index.html
Warning: The standard parity is set to 0. This can lead to data loss.
```

The root credentials remain to be `P63ITYI6L07GY1VKY3ET` and `4+yRbGik+i+Yf98T0Jrw+8XtlBAtWjYznRvd0kkH+uE` unless you change `MINIO_KMS_KES_KEY_NAME` to some other key.

harshavardhana: Yeah @aead but this only visible in stdout. But not as a service or inside k8s pods.

We hide credentials there.

ravindk89: I think with this we will still have the issue of users complaining about plaintext output _somewhere_ in the system. You don't necessarily need `root` to do `systemctl status` or `journalctl -u minio` afaik

What about using `mc` ? our assumption here is for this to work, we need the API KEY + KEY NAME. So if you specify both to mc as part of some call (`mc admin user get-root ALIAS APIKEY KEYNAME`) then it's up to the user to appropriately sanitize bash history

and at that point it goes in the `mc` json which is also plaintext so :)

harshavardhana: I think we should not ever automatically do this, we should convert this into a deployment process.

Basically we need to ask users to make sure root is disabled access and KMS is configured to make it "fully secure" i.e no credentials are ever written anywhere.

ravindk89: So basically:

1. Start MinIO normally
2. Create your non-root superuser
3. Ensure KMS is configured
4. Disable root API access
5. Remove `MINIO_ROOT_USER` and `MINIO_ROOT_PASSWORD` from env file/config
6. 

At this point MinIO will auto-generate a new root credential that is completely opaque to the end user, and cannot be found *anywhere*.

is that about right?

cmd/kms-handlers.go
shtripat: ah ok. makes sense.

cmd/common-main.go
aead: Okay, I change the PR to only generate root credentials with the KMS if root API access is disabled.

aead: Updated. Now, MinIO generates the root credentials with the KMS only in lockdown mode. Now, it also does not print root access/secret key.

harshavardhana: Yes this is perfect. This is what we can bring in Operator mode where they can enable this after everything has been configured as per their liking and no more root credentials are needed.

harshavardhana: I would test this out with site replicated setup though as there is one challenge there I want to sort this out 

// cc @poornas 

Can you give this PR a test on site replication? IAM assets for sts have root credentials requirement to be replicated. So have to check how to do this PR safely.

harshavardhana: Now that site replication PR is merged we can rebase this PR @aead and get this ready to be merged.

aead: Done @harshavardhana

harshavardhana (approved):

cmd/config-current.go
```diff
@@ -720,6 +725,55 @@ func applyDynamicConfigForSubSys(ctx context.Context, objAPI ObjectLayer, s conf
...
+// autoGenerateRootCredentials generates root credentials deterministically if
+// a KMS is configured, no manual credentials have been specified and if root
+// access is disabled.
+func autoGenerateRootCredentials() {
+	if GlobalKMS == nil {
+		return
+	}
+	if globalAPIConfig.permitRootAccess() || !globalActiveCred.Equal(auth.DefaultCredentials) {
+		return
+	}
```
donatello, line 737: Should we also check if root creds env vars are set like `MINIO_ROOT_USER` and `MINIO_ROOT_PASSWORD`? If they are I don't think we should generate via KMS

poornas: globalActiveCreds already overrides root creds set via env.var

poornas (approved):

donatello (approved):

datakurre:
@aead As a new user, may I ask the recommended way to use this new feature?

Imagine that I have a fresh MinIO deployment (maybe a server pool) running and I am happy to see it is safe with the logged line

> Automatically generated root access key and secret key with the KMS

How should I now use MinIO? Is there a way to configure e.g. OpenID and its policies declaratively without need for root credentials at all? Thank you :pray: 

Edit: My guess is that

PR #19036, opened 2024-02-12
FIx unexpected behavior when creating service account

### LINKED ISSUE #19003: Likely unintentional behaviour adding service account through API
https://github.com/minio/minio/blob/eead4db1d2d6e91a79d55e6e72f4967696a2df0d/cmd/iam.go#L985-L994

When adding a service account via the API, you can pass an access key and a secret key. You can also _only_ pass a secret key, which gets accepted as input by the server. However, the service account that gets created has a different secret key.

This does work as expected in `mc` as it has an embedded helper function to generate the credentials, and a check if either or `accessKey` or `secretKey` are empty [here](https://github.com/minio/mc/blob/62b12a2770828f9d35ee7c5614d900b95ff6e6ab/cmd/admin-user-svcacct-add.go#L302-L304).

This check should I believe be server-side, not client-side.

Suggestion: the server should not accept a `AddServiceAccount` request if `secretKey` is given, but `accessKey` is not.

I can provide steps to reproduce if desired.

taran-p:

## Description
Fixes bahvior so creating service account will error if secret key is provided without access key.
Fixes #19003 

## Motivation and Context
Explained in #19003 

## How to test this PR?
Attempt to create a service account (access key) using Console with a blank access key but specified secret key.

## Types of changes

## Checklist:


cmd/iam.go
```diff
@@ -983,7 +983,7 @@ func (sys *IAMSys) NewServiceAccount(ctx context.Context, parentUser string, gro
 
 	var accessKey, secretKey string
 	var err error
-	if len(opts.accessKey) > 0 {
+	if len(opts.accessKey) > 0 || len(opts.secretKey) > 0 {
```
vadmeste, line 993: auth.CreateNewCredentialsWithMetadata seems to always fail if access key or secret key is empty. I think it is better to add more check so that we require both are empty or both are non empty and we return an error early in that case.

harshavardhana (approved):

cmd/api-errors.go
```diff
@@ -287,6 +287,8 @@ const (
 	ErrAdminInvalidArgument
 	ErrAdminInvalidAccessKey
 	ErrAdminInvalidSecretKey
+	ErrAdminNoAccessKey
+	ErrAdminNoSecretKey
```
harshavardhana:
```suggestion
	ErrAdminNoAccessKey
	ErrAdminNoSecretKey
```

Generally add it at the end to avoid too many changes

taran-p: End of that section of the list or end of the whole list?

harshavardhana: End of the whole list.

harshavardhana (approved):

vadmeste (approved):
LGTM

PR #19037, opened 2024-02-12
Send a bucket notification event on DeleteObject() for non-existing object

Praveenrajmani:

## Description
Currently, we won't send any event for `204` responses on DeleteObject for non-existing objects. This PR sends a `s3:ObjectRemoved:Delete` event for such requests.

## Motivation and Context
To send an event for `204` responses on DeleteObject on non-existing object

## How to test this PR?
- Configure bucket notficiations
- Perform a DeleteObject on a non-existing object
- You should see an event triggered

## Types of changes

## Checklist:

### DIFF
--- a/cmd/object-handlers.go
+++ b/cmd/object-handlers.go
@@ -2882,6 +2882,18 @@ func (api objectAPIHandlers) DeleteObjectHandler(w http.ResponseWriter, r *http.
 			return
 		}
 		if isErrObjectNotFound(err) || isErrVersionNotFound(err) {
+			// Send an event when the object is not found
+			objInfo.Name = object
+			objInfo.VersionID = opts.VersionID
+			sendEvent(eventArgs{
+				EventName:    event.ObjectRemovedNoOP,
+				BucketName:   bucket,
+				Object:       objInfo,
+				ReqParams:    extractReqParams(r),
+				RespElements: extractRespElements(w),
+				UserAgent:    r.UserAgent(),
+				Host:         handlers.GetSourceIP(r),
+			})
 			writeSuccessNoContent(w)
 			return
 		}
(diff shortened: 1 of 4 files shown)


cmd/object-handlers.go
```diff
@@ -2882,6 +2882,18 @@ func (api objectAPIHandlers) DeleteObjectHandler(w http.ResponseWriter, r *http.
 			return
 		}
 		if isErrObjectNotFound(err) || isErrVersionNotFound(err) {
+			// Send an event when the object is not found
+			objInfo.Name = object
+			objInfo.VersionID = opts.VersionID
+			sendEvent(eventArgs{
+				EventName:    event.ObjectRemovedDelete,
```
Praveenrajmani: AWS S3 creates a DEL marker when DELETE is called on a non-existing object. Thinking if we can send `s3:ObjectRemoved:DeleteMarkerCreated` for such requests on versioned buckets.. 

any suggestions? @harshavardhana

vadmeste: MinIO does not create a delete marker in this case, so we cannot send a notification with reliable information, such as delete-marker version-id, does someone need this ?

harshavardhana: We should send notification that we didn't do anything that's all. 

404 or otherwise we should notify the 204s sent to the client.

In this case nothing is created so it is a no-op of sort where we say that it's 204 but we didn't have the object in the first place.

Add an additional metadata to the notification.

harshavardhana: We should add event.ObjectRemovedNoOP of sort here.

harshavardhana (changes requested):

harshavardhana (approved):

PR #19049, opened 2024-02-13
Refresh tier config periodically

krisis:

## Description
- Increase the parity for tier-config.bin object
- Refresh globalTierConfigMgr cached value once every 15 mins

## Motivation and Context
When tier configuration wasn't loaded at the start up of MinIO server it never retries subsequently. 

## How to test this PR?
- Setup one or more tiers
- `mc ilm tier ls ...` multiple times if ALIAS is configured to a load balanced address to ensure that the list of tiers returned is up to date.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/tier-handlers.go
+++ b/cmd/tier-handlers.go
@@ -140,6 +140,7 @@ func (api adminAPIHandlers) ListTierHandler(w http.ResponseWriter, r *http.Reque
 		return
 	}
 
+	w.Header().Set(tierCfgRefreshAtHdr, globalTierConfigMgr.refreshedAt().String())
 	writeSuccessResponseJSON(w, data)
 }
 
--- a/cmd/tier.go
+++ b/cmd/tier.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2015-2023 MinIO, Inc.
+// Copyright (c) 2015-2024 MinIO, Inc
 //
 // This file is part of MinIO Object Storage stack
 //
@@ -33,6 +33,7 @@ import (
 	"github.com/minio/minio/internal/crypto"
 	"github.com/minio/minio/internal/hash"
 	"github.com/minio/minio/internal/kms"
+	"github.com/minio/minio/internal/logger"
 	"github.com/prometheus/client_golang/prometheus"
 )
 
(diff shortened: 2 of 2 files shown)


cmd/tier-handlers.go
```diff
@@ -140,6 +140,7 @@ func (api adminAPIHandlers) ListTierHandler(w http.ResponseWriter, r *http.Reque
 		return
 	}
 
+	w.Header().Set(tierCfgRefreshAtHdr, globalTierConfigMgr.refreshedAt().String())
```
harshavardhana, line 143: Good thing to add this to other period reloaders as well.

harshavardhana (approved):

cmd/tier.go
```diff
@@ -501,6 +512,27 @@ func NewTierConfigMgr() *TierConfigMgr {
...
+		wait := sleeper.Timer(ctx)
+		select {
+		case <-ctx.Done():
+			return
+		case <-t.C:
+			err := config.Reload(ctx, objAPI)
+			if err != nil {
+				logger.LogIf(ctx, err)
+			}
+			wait()
```
vadmeste: the logic of this wait is a bit weird, it is not that we are not going to sleep here, we still reset the timer here, therefore we are going to sleep for 15 minutes

vadmeste (commented):
Few comments

cmd/tier.go
harshavardhana: I think it's a wait() after the config has hit the disk. Basically an additional time to wait for the servers, and let everyone not read at the same time.

harshavardhana: For that to happen we need it the sleep before Reload() not after.

krisis: @vadmeste good catch. @harshavardhana yes, that was my intention. I now realize that a random sleep of up to 10s is the correct approach for avoiding all MinIO servers contending for the read-locks required to fetch the tier configuration object.
Let me address this.

harshavardhana: if you want to stagger it stagger it like bucketMetadataSys reload() we have to stagger it before the read()

krisis: PTAL @vadmeste @harshavardhana.

harshavardhana (approved):

harshavardhana (approved):

vadmeste (approved):
LGTM

PR #19104, opened 2024-02-21
fix: cache diskInfo to avoid frequent network calls

harshavardhana:

## Description
fix: cache diskInfo to avoid frequent network calls

## Motivation and Context
just improvement in performance under very high
concurrent systems.

## How to test this PR?
CI/CD should do it. 

## Types of changes

## Checklist:

### DIFF
--- a/cmd/erasure.go
+++ b/cmd/erasure.go
@@ -275,9 +275,7 @@ func (er *erasureObjects) LocalStorageInfo(ctx context.Context, metrics bool) St
 // - Non-healing disks
 // - Healing disks (if inclHealing is true)
 func (er *erasureObjects) getOnlineDisksWithHealingAndInfo(inclHealing bool) (newDisks []StorageAPI, newInfos []DiskInfo, healing bool) {
-	disks := er.getDisks()
-
-	infos := func() []DiskInfo {
+	infos := func(disks []StorageAPI) []DiskInfo {
 		var wg sync.WaitGroup
 		dinfos := make([]DiskInfo, len(disks))
 		r := rand.New(rand.NewSource(time.Now().UnixNano()))
@@ -313,17 +311,19 @@ func (er *erasureObjects) getOnlineDisksWithHealingAndInfo(inclHealing bool) (ne
 		er.onlineDisksHealingInfoCache.Once.Do(func() {
 			er.onlineDisksHealingInfoCache.TTL = time.Second
 			er.onlineDisksHealingInfoCache.Update = func() (interface{}, error) {
-				return infos(), nil
+				return infos(er.getDisks()), nil
 			}
 		})
 		v, _ := er.onlineDisksHealingInfoCache.Get()
 		dinfos = v.([]DiskInfo)
 	} else {
-		dinfos = infos()
+		dinfos = infos(er.getDisks())
 	}
 
 	var scanningDisks, healingDisks []StorageAPI
 	var scanningInfos, healingInfos []DiskInfo
+
+	disks := er.getDisks()
 
 	for i, info := range dinfos {
 		// Check if one of the drives in the set is being healed.
@@ -332,6 +332,12 @@ func (er *erasureObjects) getOnlineDisksWithHealingAndInfo(inclHealing bool) (ne
 		if info.Error != "" || disks[i] == nil {
 			continue
 		}
+
+		// Check if the disk is offline
+		if !disks[i].IsOnline() {
+			continue
+		}
+
 		if info.Healing {
 			healing = true
 			if inclHealing {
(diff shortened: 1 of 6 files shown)


harshavardhana:
The reason to move the caching from storage-rest-client -> erasure-objects is because the DiskInfo() caller seems to have some shared buffers causing data races.

IT made sense to add this caching at a higher layer where its most-useful.

cmd/erasure.go
```diff
@@ -272,38 +274,58 @@ func (er erasureObjects) LocalStorageInfo(ctx context.Context, metrics bool) Sto
...
+	}
+
+	var dinfos []DiskInfo
+	if inclHealing {
+		// List/Walk calls ask for all disks including the ones under healing
+		// pick DiskInfo() for those disks as well.
+		er.onlineDisksHealingInfoCache.Once.Do(func() {
+			er.onlineDisksHealingInfoCache.TTL = time.Second
+			er.onlineDisksHealingInfoCache.Update = func() (interface{}, error) {
+				return infos(), nil
```
krisis: `infos` closure binds to a point in time value of `er.getDisks()` via the variable `disks`. This will be called every time `er.onlineDisksHealingInfoCache` requires an update. 
We should not rely on `disks` and instead call `er.getDisks()` inside the closure `infos`.

cmd/storage-rest-client.go
```diff
@@ -306,20 +306,20 @@ func (client *storageRESTClient) DiskInfo(ctx context.Context, opts DiskInfoOpti
...
 		return info, errDiskNotFound
 	}
-	ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
-	defer cancel()
 
 	opts.DiskID = client.diskID
+	ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
+	defer cancel()
 
 	infop, err := storageDiskInfoRPC.Call(ctx, client.gridConn, &opts)
```
vadmeste, line 314: aren't we still doing the frequent network call here ?

harshavardhana: We are doing this once in a second, in the hot path

cmd/erasure.go
```diff
@@ -272,59 +274,85 @@ func (er erasureObjects) LocalStorageInfo(ctx context.Context, metrics bool) Sto
...
+		// pick DiskInfo() for those disks as well.
+		er.onlineDisksHealingInfoCache.Once.Do(func() {
+			er.onlineDisksHealingInfoCache.TTL = time.Second
+			er.onlineDisksHealingInfoCache.Update = func() (interface{}, error) {
+				return infos(er.getDisks()), nil
 			}
-		}()
+		})
+		v, _ := er.onlineDisksHealingInfoCache.Get()
+		dinfos = v.([]DiskInfo)
```
klauspost, line 318: Maybe copy, so we don't risk the caller reshuffling the slice or something.
```suggestion
		dinfos = v.([]DiskInfo)
		dinfos = append(make([]DiskInfo, 0, len(dinfos)), dinfos...)
```

cmd/storage-rest-client.go
klauspost: Main thing is that it is a very specific cache and any upstream changes will get us back to requesting it every time.

For example `PutObject` calls `getDiskInfos`, which returns uncached values:

https://github.com/minio/minio/blob/607cafadbc358b45163de95147c72643022ba302/cmd/erasure-server-pool.go#L1007

So every PutObject has `set_size` roundtrips before it will upload. Seems quite wasteful - even if the RPC calls are very fast :)

klauspost:
Main thing is that it is a very specific cache and any upstream changes will get us back to requesting it every time.

For example `PutObject` calls `getDiskInfos`, which returns uncached values:

https://github.com/minio/minio/blob/607cafadbc358b45163de95147c72643022ba302/cmd/erasure-server-pool.go#L1007

So every PutObject has `set_size` roundtrips before it will upload. Seems quite wasteful - even if the RPC calls are very fast :)

harshavardhana:
> Main thing is that it is a very specific cache and any upstream changes will get us back to requesting it every time.
> 
> For example `PutObject` calls `getDiskInfos`, which returns uncached values:
> 
> https://github.com/minio/minio/blob/607cafadbc358b45163de95147c72643022ba302/cmd/erasure-server-pool.go#L1007
> 
> So every PutObject has `set_size` roundtrips before it will upload. Seems quite wasteful - even if the RPC calls are very fast :)

Correct but I can't get rid of races with cache at the rpc layer. So not sure if there is a cleaner way. 

If you can suggest I would take that route honestly.

klauspost:
Let me take a look.

klauspost:
I try to convert timedValue to a generic handler to get rid of the painful conversion as well.

harshavardhana:
> I try to convert timedValue to a generic handler to get rid of the painful conversion as well.

Yeah that will be great.

harshavardhana:
Sending a different PR with more generic implementation.

PR #19111, opened 2024-02-22
 sr: use site replicator svcacct to sign STS session tokens

poornas:
This change is to decouple need for root credentials to match between
 site replication deployments.

 Also ensuring site replication config initialization is re-tried until
 it succeeds, this dependency is critical to STS flow in site replication
 scenario.

## How to test this PR?
setup site replication with root credentials different across sites
STS tokens generated should replicate

## Types of changes

## Checklist:

### DIFF
--- a/cmd/site-replication.go
+++ b/cmd/site-replication.go
@@ -26,6 +26,7 @@ import (
 	"encoding/xml"
 	"errors"
 	"fmt"
+	"math/rand"
 	"net/url"
 	"reflect"
 	"runtime"
@@ -227,19 +228,30 @@ type srStateData struct {
 // Init - initialize the site replication manager.
 func (c *SiteReplicationSys) Init(ctx context.Context, objAPI ObjectLayer) error {
 	go c.startHealRoutine(ctx, objAPI)
+	r := rand.New(rand.NewSource(time.Now().UnixNano()))
+	for {
+		err := c.loadFromDisk(ctx, objAPI)
+		if err == errConfigNotFound {
+			return nil
+		}
+		if err == nil {
+			break
+		}
+		logger.LogOnceIf(context.Background(), fmt.Errorf("unable to initialize site replication subsystem: (%w)", err), "site-relication-init")
 
-	err := c.loadFromDisk(ctx, objAPI)
-	if err == errConfigNotFound {
-		return nil
+		duration := time.Duration(r.Float64() * float64(time.Minute))
+		if duration < time.Second {
+			// Make sure to sleep at least a second to avoid high CPU ticks.
+			duration = time.Second
+		}
+		time.Sleep(duration)
 	}
-
 	c.RLock()
 	defer c.RUnlock()
 	if c.enabled {
 		logger.Info("Cluster replication initialized")
 	}
-
-	return err
+	return nil
 }
 
 func (c *SiteReplicationSys) loadFromDisk(ctx context.Context, objAPI ObjectLayer) error {
@@ -582,6 +594,9 @@ func (c *SiteReplicationSys) AddPeerClusters(ctx context.Context, psites []madmi
 		}, nil
 	}
 
+	if !globalSiteReplicatorCred.IsValid() {
+		globalSiteReplicatorCred.Set(svcCred)
+	}
 	result := madmin.ReplicateAddStatus{
 		Success: true,
 		Status:  madmin.ReplicateAddStatusSuccess,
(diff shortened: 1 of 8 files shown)


donatello:
We need to document that only new sts creds created will work with SR.

cmd/site-replication.go
```diff
@@ -227,19 +228,28 @@ type srStateData struct {
...
+		}
+		if err == nil {
+			break
+		}
+		duration := time.Duration(r.Float64() * float64(time.Minute))
+		if duration < time.Second {
+			// Make sure to sleep at least a second to avoid high CPU ticks.
+			duration = time.Second
+		}
+		time.Sleep(duration)
```
donatello, line 247: Maybe we should log an error at least once (OnceIf) here because this perpetually retries?

cmd/server-main.go
```diff
@@ -984,7 +984,7 @@ func serverMain(ctx *cli.Context) {
 
 		// Initialize site replication manager after bucket metadata
 		bootstrapTrace("globalSiteReplicationSys.Init", func() {
-			globalSiteReplicationSys.Init(GlobalContext, newObject)
+			go globalSiteReplicationSys.Init(GlobalContext, newObject)
```
harshavardhana: Why is it a goroutine?

cmd/site-replication.go
```diff
@@ -2382,6 +2401,9 @@ func (c *SiteReplicationSys) RemovePeerCluster(ctx context.Context, objectAPI Ob
 	st = madmin.ReplicateRemoveStatus{
 		Status: madmin.ReplicateRemoveStatusSuccess,
 	}
+	if !c.isEnabled() {
+		globalSiteReplicatorCred = auth.Credentials{}
```
harshavardhana: Is this needed?

cmd/sts-handlers.go
```diff
@@ -190,6 +190,18 @@ func parseForm(r *http.Request) error {
 	return nil
 }
 
+// getTokenSigningKey returns secret key used to sign JWT session tokens
+func getTokenSigningKey() (string, error) {
+	secret := globalActiveCred.SecretKey
+	if globalSiteReplicationSys.isEnabled() {
+		if !globalSiteReplicatorCred.IsValid() {
```
harshavardhana: Just like we do loadUser() we must try reloading this from disk and then throw an error if any.

harshavardhana (changes requested):

cmd/server-main.go
poornas: not needed - this is already being. called in a goroutine

cmd/site-replication.go
poornas: can omit this since the config is disabled anyways

docs/site-replication/README.md
```diff
@@ -56,3 +56,6 @@ mc admin replicate add minio1 minio2 minio3
 ```sh
 mc admin replicate info minio1
 ```
+
+** Note **
+Previously, site replication required root credentials of peer sites be identical. This is no longer required due to STS tokens now being signed with the site replicator service account credentials. However, this means that STS tokens signed previously by root credentials will no longer be valid upon upgrade to the latest version with this change. Also, if site replication is ever removed - the STS tokens previously signed by site replicator service account credentials will no longer be valid.
```
harshavardhana:
```suggestion
Previously, site replication required root credentials of peer sites be identical. This is no longer required due to STS tokens now being signed with the site replicator service account credentials. However, this means that STS tokens signed previously by root credentials will no longer be valid upon upgrade to the latest version with this change. Also, if site replication is ever removed - the STS tokens previously signed by site replicator service account credentials will no longer be valid.
```

We should re-word it to indicate why it is done in this manner, basically to allow the flexibility of having different "root" accounts per site. As well as eventually disabling this root account for all practical purposes.

docs/site-replication/README.md
```diff
@@ -56,3 +56,8 @@ mc admin replicate add minio1 minio2 minio3
 ```sh
 mc admin replicate info minio1
 ```
+
+** Note **
+Previously, site replication required root credentials of peer sites be identical. This is no longer required due to STS tokens now being signed with the site replicator service account credentials vs the root credentials, thus allowing flexibility in independent manageement of root accounts across sites, as well as the ability to disable root account eventually.
```
harshavardhana:
```suggestion
Previously, site replication required the root credentials of peer sites to be identical. This is no longer necessary because STS tokens are now signed with the site replicator service account credentials, thus allowing flexibility in the independent management of root accounts across sites and the ability to disable root accounts eventually.
```

docs/site-replication/README.md
```diff
@@ -56,3 +56,8 @@ mc admin replicate add minio1 minio2 minio3
...
+
+ However, this means that STS tokens signed previously by root credentials will no longer be valid upon upgrade to the latest version with this change. Also, if site replication is ever removed - the STS tokens previously signed by site replicator service account credentials will no longer be valid.
```
harshavardhana:
```suggestion
However, this means that STS tokens signed previously by root credentials will no longer be valid upon upgrading to the latest version with this change. Please re-generate them as you usually do. Additionally, if site replication is ever removed - the STS tokens will become invalid, regenerate them as you usually do.
```

docs/site-replication/run-multi-site-oidc.sh
```diff
@@ -6,13 +6,10 @@ exit_1() {
 
 	echo "minio1 ============"
 	cat /tmp/minio1_1.log
-	cat /tmp/minio1_2.log
```
harshavardhana, line 9: ?

harshavardhana (changes requested):

docs/site-replication/run-multi-site-oidc.sh
poornas: this script uses just one pool

harshavardhana (approved):

donatello (approved):

PR #19130, opened 2024-02-26
Improve caching

klauspost:
## Description

* Remove lock for cached operations.
* Rename "Relax" to `ReturnLastGood`.
* Add `CacheError` to allow caching values even on errors.
* Add NoWait that will return current value with async fetching if within 2xTTL.
* Make benchmark somewhat representative.  Result with 1ms for both implementions:

```
Before: BenchmarkCache-12       16408370                63.12 ns/op            0 B/op
After:  BenchmarkCache-12       428282187                2.789 ns/op           0 B/op
```

* Remove `storageRESTClient.scanning`. Nonsensical - RPC clients will not have any idea about scanning.
* Always fetch remote diskinfo metrics and cache them. Seems most calls are requesting metrics.
* Do async fetching of usage caches.

## How to test this PR?

Regular operation.

## Types of changes

### DIFF
--- a/internal/cachevalue/cache.go
+++ b/internal/cachevalue/cache.go
@@ -19,19 +19,20 @@ package cachevalue
 
 import (
 	"sync"
+	"sync/atomic"
 	"time"
 )
 
 // Cache contains a synchronized value that is considered valid
 // for a specific amount of time.
 // An Update function must be set to provide an updated value when needed.
-type Cache[I any] struct {
+type Cache[T any] struct {
 	// Update must return an updated value.
 	// If an error is returned the cached value is not set.
 	// Only one caller will call this function at any time, others will be blocking.
 	// The returned value can no longer be modified once returned.
 	// Should be set before calling Get().
-	Update func() (item I, err error)
+	Update func() (T, error)
 
 	// TTL for a cached value.
 	// If not set 1 second TTL is assumed.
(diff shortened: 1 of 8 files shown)


internal/cachevalue/cache.go
```diff
@@ -61,30 +75,42 @@ func New[I any]() *Cache[I] {
...
-			// to use the returned value based
-			// on error.
-			item, ok = t.get(0)
-			if ok {
-				return item, err
-			}
+	// Fetch new value.
+	if t.NoWait && v != nil && tNow-vTime < ttl.Milliseconds()*2 && (v.e == nil || t.CacheError) {
+		if t.updating.TryLock() {
+			go func() {
```
harshavardhana, line 94: On a hung server wouldn't this cause a run away build-up ? 

How about we only do this one goroutine? Because one success here is applicable to everyone who is waiting on disk.###

internal/cachevalue/cache.go
```diff
@@ -95,25 +121,20 @@ func (t *Cache[_]) ttl() time.Duration {
...
-	defer t.mu.RUnlock()
-	if t.valueSet {
-		item = t.value
-		if ttl <= 0 {
-			return item, true
-		}
-		if time.Since(t.lastUpdate) < ttl {
-			return item, true
+func (t *Cache[T]) update() {
+	val, err := t.Update()
```
harshavardhana, line 125: Since this is locked can we check if the Update() is actually?

With NoWait implementation we have the risk of many updates blocked and happen serially instead we can unblock on first success right away and not have to spend time updating if the last update is already updated.

harshavardhana (changes requested):

internal/cachevalue/cache.go
klauspost: @harshavardhana `TryLock` will only work once per cache. So it cannot build up.

This will block just like the existing if the Update function hangs. This however gives a TTL window where it will keep responding.

internal/cachevalue/cache.go
klauspost: > Since this is locked can we check if the Update() is actually?

Something missing here.

> With NoWait implementation we have the risk of many updates blocked and happen serially instead we can unblock on first success right away and not have to spend time updating if the last update is already updated.

No. There will only ever be one update running. All others will unblock as well when the new result is ready.

harshavardhana (approved):

PR #19134, opened 2024-02-26
ilm: Select an object when all AND tags are satisfied

vadmeste:

## Description
Currently if one object tag matches with one lifecycle tag filter, ILM will select it, 
however this is wrong. All the Tag filters in the lifecycle document should be satisfied.

## Types of changes

## Checklist:

### DIFF
--- a/internal/bucket/lifecycle/filter.go
+++ b/internal/bucket/lifecycle/filter.go
@@ -232,23 +232,20 @@ func (f Filter) TestTags(userTags string) bool {
 	}
 	tagsMap := parsedTags.ToMap()
 
-	// This filter has tags configured but this object
-	// does not have any tag, skip this object
-	if len(tagsMap) == 0 {
+	// Not enough tags on object to satisfy the rule filter's tags
+	if len(tagsMap) < len(f.cachedTags) {
 		return false
 	}
 
-	// Both filter and object have tags, ensure that
-	// all lifecycle tags rules are satisfied
-	var matchCount int
+	var mismatch bool
 	for k, cv := range f.cachedTags {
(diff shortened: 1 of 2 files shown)


internal/bucket/lifecycle/filter.go
```diff
@@ -238,15 +238,17 @@ func (f Filter) TestTags(userTags string) bool {
...
+	var matchCount int
 	for k, cv := range f.cachedTags {
 		v, ok := tagsMap[k]
 		if ok && v == cv {
-			return true
+			matchCount++
 		}
 	}
-	return false
+
```
krisis: We can return at the first mismatch between object tags and the tags in the rule filter.
```go
// Not enough tags on object to satisfy the rule filter's tags
if len(tagsMap) < len(f.cachedTags) {
  return false
}

var mismatch bool
for k, cv := range f.cachedTags {
  v, ok := tagsMap[k]
  if !ok || v != cv {
    mismatch = true
    break
  }
}

return !mismatch
```

internal/bucket/lifecycle/filter_test.go
```diff
@@ -241,3 +241,103 @@ func TestObjectSizeFilters(t *testing.T) {
...
+			filter:   oneTag,
+			userTags: "A=3",
+			want:     false,
+		},
+		{
+			filter:   oneTag,
+			userTags: "FOO=1",
+			want:     true,
+		},
+		{
```
krisis, line 295: Neat!

krisis (commented):
LGTM, except a minor suggestion.

harshavardhana (approved):

krisis (approved):

PR #19137, opened 2024-02-27
cachevalue: simplified exported interface

donatello:
## Description

~~- Also add cache options type, and simplify the exported interface of
the package.~~

- Also add cache options type

## Motivation and Context

~~Passing a context value to the Get* method allows quicker cancellation of the update method if the value is no longer needed.~~

## Types of changes

## Checklist:

### DIFF
--- a/cmd/admin-handlers-users.go
+++ b/cmd/admin-handlers-users.go
@@ -33,6 +33,7 @@ import (
 	"github.com/klauspost/compress/zip"
 	"github.com/minio/madmin-go/v3"
 	"github.com/minio/minio/internal/auth"
+	"github.com/minio/minio/internal/cachevalue"
 	"github.com/minio/minio/internal/config/dns"
 	"github.com/minio/minio/internal/logger"
 	"github.com/minio/mux"
@@ -1197,21 +1198,15 @@ func (a adminAPIHandlers) AccountInfoHandler(w http.ResponseWriter, r *http.Requ
 		return rd, wr
 	}
 
-	bucketStorageCache.Once.Do(func() {
-		// Set this to 10 secs since its enough, as scanner
-		// does not update the bucket usage values frequently.
-		bucketStorageCache.TTL = 10 * time.Second
-
-		// Rely on older value if usage loading fails from disk.
-		bucketStorageCache.ReturnLastGood = true
-		bucketStorageCache.NoWait = true
-		bucketStorageCache.Update = func() (DataUsageInfo, error) {
+	bucketStorageCache.InitOnce(10*time.Second,
+		cachevalue.Opts{ReturnLastGood: true, NoWait: true},
+		func() (DataUsageInfo, error) {
 			ctx, done := context.WithTimeout(context.Background(), 2*time.Second)
 			defer done()
 
 			return loadDataUsageFromBackend(ctx, objectAPI)
-		}
-	})
+		},
+	)
 
 	dataUsageInfo, _ := bucketStorageCache.Get()
 
--- a/internal/cachevalue/cache.go
+++ b/internal/cachevalue/cache.go
@@ -23,22 +23,8 @@ import (
 	"time"
 )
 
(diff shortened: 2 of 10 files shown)


cmd/admin-handlers-users.go
```diff
@@ -1197,23 +1197,7 @@ func (a adminAPIHandlers) AccountInfoHandler(w http.ResponseWriter, r *http.Requ
...
-		bucketStorageCache.NoWait = true
-		bucketStorageCache.Update = func() (DataUsageInfo, error) {
-			ctx, done := context.WithTimeout(context.Background(), 2*time.Second)
-			defer done()
-
-			return loadDataUsageFromBackend(ctx, objectAPI)
-		}
-	})
-
-	dataUsageInfo, _ := bucketStorageCache.Get()
```
donatello, line 1216: There was a double initialization that is removed here (see bucket-quota.go).

vadmeste (approved):
LGTM

cmd/admin-handlers-users.go
klauspost: What if it hits here first? Only one will ever be called anyway.

internal/cachevalue/cache.go
```diff
@@ -67,33 +71,54 @@ type Cache[T any] struct {
...
 	}
 
 	// Fetch new value.
-	if t.NoWait && v != nil && tNow-vTime < ttl.Milliseconds()*2 && (v.e == nil || t.CacheError) {
+	if t.opts.NoWait && v != nil && tNow-vTime < ttl.Milliseconds()*2 && (v.e == nil || t.opts.CacheError) {
 		if t.updating.TryLock() {
 			go func() {
 				defer t.updating.Unlock()
-				t.update()
+				t.update(ctx)
```
klauspost: This context cannot be used async - it will likely be canceled before it is done.

internal/cachevalue/cache.go
```diff
@@ -107,24 +132,22 @@ func (t *Cache[I]) Get() (item I, err error) {
 		t.updating.Unlock()
 		return v.v, v.e
 	}
-	t.update()
+	t.update(ctx)
```
klauspost: The context/lifetime of a request should not affect how the cache is populated.

I looked at this, but actively decided not to go this route. The cache population should not be affected by random cancelation of whatever context happens to be used for the repopulation.

klauspost (changes requested):

klauspost:
You are welcome to keep the initialization alternative. But I disagree on propagating the context as being an improvement.

harshavardhana:
We need the cache to be populated without any hindrance from the top level caller.

cmd/admin-handlers-users.go
donatello: Ah well I thought, the other init spot would always finish first, but I see this is not guaranteed. I'll bring the second init code back and try to remove it in a better way. In server-main.go we initialize many subsystems asynchronously, so perhaps there are more such situations where the subsystem starts getting used before init completes.

internal/cachevalue/cache.go
donatello: I wanted to let the caller have some control over how long the Get+Update could take. An update operation could take an unusually long time - it is better to cancel and retry later. We already take steps for this by setting a constant deadline/timeout within the context created inside the update function (where needed). It seems the amount of the constant timeout being set is arbitrary (we usually cannot foresee how long is "long enough" for the operation in question in all kinds of situations and scales) - thus the reason to make it dependent on the request context.

donatello:
> We need the cache to be populated without any hindrance from the top level caller.

Replied here https://github.com/minio/minio/pull/19137/files#r1504745343

donatello:
Updated the PR and removed the context passing (we can bring it back if we are convinced). Added the duplicate initialization as well.

klauspost (approved):

vadmeste (approved):

PR #19149, opened 2024-02-28
Mux: Send async mux ack and fix stream error responses

klauspost:
## Description

Streams can return errors if the cancelation is picked up before the response stream close is picked up. Under extreme load this could lead to missing responses.

Send server mux ack async so a blocked send cannot block newMuxStream call. Stream will not progress until mux has been acked.

## Motivation and Context

Stability under high load.

## How to test this PR?

High node, latency, high load.

## Types of changes


vadmeste (approved):
LGTM

harshavardhana (approved):

harshavardhana (approved):

PR #19160, opened 2024-02-29
fix: make sure precondition is verified properly before PUTs

Praveenrajmani:

## Description

PutObject with SourceETag and SourceVersionID should fail as expected when the latest object fail to pass precondition checks

## Motivation and Context

PutObject should honor precondition checks properly

## How to test this PR?

Try PutObject with SourceETag set to the etag of the existing object

```go
info, err = s3Client.PutObject(context.Background(), "bucket", "object",
		obj,
		ojStat.Size(),
		minio.PutObjectOptions{
			ContentType: "application/octet-stream",
			Internal: minio.AdvancedPutOptions{
				SourceETag: oi.ETag,
			},
		})
if err != nil {
	log.Fatalln(err)
}
```

## Types of changes

## Checklist:

### DIFF
--- a/cmd/object-handlers-common.go
+++ b/cmd/object-handlers-common.go
@@ -190,9 +190,9 @@ func checkPreconditionsPUT(ctx context.Context, w http.ResponseWriter, r *http.R
 		}
 	}
 
-	etagMatch := opts.PreserveETag != "" && isETagEqual(objInfo.ETag, opts.PreserveETag)
-	vidMatch := opts.VersionID != "" && opts.VersionID == objInfo.VersionID
-	if etagMatch && vidMatch {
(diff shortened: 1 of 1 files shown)


harshavardhana:
We need both conditions to be true. An etag might match but not versionID, are you saying we shouldn't replicate at that point?

harshavardhana:
A precondition check is to ensure that we do not perform the PUT when the etag and vid matches. Given both of them are provided in the request. 

If they are not provided in the request then precondition is a no-op. In your test you are doing it wrong.

Praveenrajmani:
Clarified this with @harshavardhana offline, we can use `If-None-Match` match header to solve this. Thanks @harshavardhana

PR #19205, opened 2024-03-05
batch: Separate batch job request and batch job stats

vadmeste:

## Description
Currently, the progress of the batch job is saved in inside the job request
 object, which is normally not supported by MinIO. Though there is no apparent bug,
 it is better to fix this now.

Batch progress is saved in .minio.sys/batch-jobs/reports/

## Types of changes

## Checklist:

### DIFF
--- a/cmd/batch-handlers.go
+++ b/cmd/batch-handlers.go
@@ -63,7 +63,6 @@ type BatchJobRequest struct {
 	ID        string               `yaml:"-" json:"name"`
 	User      string               `yaml:"-" json:"user"`
 	Started   time.Time            `yaml:"-" json:"started"`
-	Location  string               `yaml:"-" json:"location"`
 	Replicate *BatchJobReplicateV1 `yaml:"replicate" json:"replicate"`
 	KeyRotate *BatchJobKeyRotateV1 `yaml:"keyrotate" json:"keyrotate"`
 	Expire    *BatchJobExpire      `yaml:"expire" json:"expire"`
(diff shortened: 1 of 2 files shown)


harshavardhana (approved):

cmd/batch-handlers.go
```diff
@@ -1671,11 +1676,13 @@ func (a adminAPIHandlers) CancelBatchJob(w http.ResponseWriter, r *http.Request)
...
-		ID:       jobID,
-		Location: pathJoin(batchJobPrefix, jobID),
+		ID: jobID,
 	}
 
-	j.delete(ctx, objectAPI)
+	if err := j.delete(ctx, objectAPI); err != nil {
+		writeErrorResponseJSON(ctx, w, toAPIError(ctx, err), r.URL)
+		return
+	}
```
harshavardhana: The idea here for not returning errors is to avoid clients retrying the operation again and again. It is not necessary TBH and we must simply garbage collect these. 

We should not waste time returning errors to the client because that is not a useful use of a client's time such as `mc`.  Users can move on to new things because it is a cancel operation. The actual job is already canceled, we are just deleting a job ID and its state() file. 

We should be more focussed on a more idempotent behavior.

harshavardhana (changes requested):

harshavardhana:
BTW the original intention of having nested objects here was to simply purge the parent to get rid of all the states for that jobID. 

It allows for a "container" like approach and the objects inside are ephemeral and not visible on the namespace with List() :-)

vadmeste:
We can review and merge this. I am sending another unrelated fix in another PR.

harshavardhana (approved):

PR #11327, opened 2021-01-21
Add PoolNumber field to madmin.ServerProperties

balamurugana:
## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Optimization (provides speedup with no functional changes)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [ ] Fixes a regression (If yes, please add `commit-id` or `PR #` here)
- [ ] Documentation updated
- [ ] Unit tests added/updated

PR #19216, opened 2024-03-07
fix: same server to be part of multiple pools

harshavardhana:

## Description
fix: same server to be part of multiple pools

## Motivation and Context
our PoolNumber calculation was costly,
while we already had this information per 
endpoint, we needed to deduce it appropriately.

This PR addresses this by assigning PoolNumbers 
field that carries all the pool numbers that 
belong to a server.

properties.PoolNumber still carries a valid value 
only when len(properties.PoolNumbers) == 1, otherwise
properties.PoolNumber is set to math.MaxInt (indicating 
that this value is undefined) and then one must rely 
on properties.PoolNumbers for server participation 
in multiple pools.

addresses the issue originating from #11327

## How to test this PR?
You need a large setup.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/admin-server-info.go
+++ b/cmd/admin-server-info.go
@@ -19,10 +19,12 @@ package cmd
 
 import (
 	"context"
+	"math"
 	"net/http"
 	"os"
 	"runtime"
 	"runtime/debug"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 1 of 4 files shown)


cmd/admin-server-info.go
```diff
@@ -42,9 +45,18 @@ func getLocalServerProperty(endpointServerPools EndpointServerPools, r *http.Req
...
-	for _, ep := range endpointServerPools {
+	for idx, ep := range endpointServerPools {
 		for _, endpoint := range ep.Endpoints {
+			if addr == endpoint.Host {
+				poolNumbers[idx+1] = struct{}{}
+			} else if host, err := xnet.ParseHost(addr); err == nil {
+				if host.Name == endpoint.Hostname() {
+					poolNumbers[idx+1] = struct{}{}
+				}
+			}
```
vadmeste: I think another simpler way is to use endpoint.IsLocal, since this function name is called getLocalServerProperty() and returns information about the current node. It is better than parsing those addresses

harshavardhana: Oh yeah that is also doable

vadmeste (approved):
LGTM

PR #19220, opened 2024-03-07
Add Storage Class override

klauspost:
## Description

Add `force_storage_class` api config.

Allows to force a specific storage class to always be returned. Fix for clients that only expect certain storage classes to be returned and using tiering.

## How to test this PR?

Testing: `mc admin config set myminio api force_storage_class=KLAUS` - run listings, etc.

## Types of changes

### DIFF
--- a/cmd/batch-handlers.go
+++ b/cmd/batch-handlers.go
@@ -526,7 +526,7 @@ func toObjectInfo(bucket, object string, objInfo miniogo.ObjectInfo) ObjectInfo 
 
 	_, ok = oi.UserDefined[xhttp.AmzStorageClass]
 	if !ok {
-		oi.UserDefined[xhttp.AmzStorageClass] = objInfo.StorageClass
+		oi.UserDefined[xhttp.AmzStorageClass] = globalAPIConfig.storageClass(objInfo.StorageClass)
 	}
 
 	for k, v := range objInfo.UserMetadata {
--- a/cmd/bucket-replication.go
+++ b/cmd/bucket-replication.go
@@ -2284,7 +2284,7 @@ func proxyHeadToRepTarget(ctx context.Context, bucket, object string, rs *HTTPRa
 			DeleteMarker:              objInfo.IsDeleteMarker,
 			ContentType:               objInfo.ContentType,
 			Expires:                   objInfo.Expires,
-			StorageClass:              objInfo.StorageClass,
+			StorageClass:              globalAPIConfig.storageClass(objInfo.StorageClass),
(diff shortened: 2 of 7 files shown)


cmd/erasure-metadata.go
```diff
@@ -158,11 +158,11 @@ func (fi FileInfo) ToObjectInfo(bucket, object string, versioned bool) ObjectInf
 
 	// Update storage class
 	if fi.TransitionTier != "" {
-		objInfo.StorageClass = fi.TransitionTier
+		objInfo.StorageClass = globalAPIConfig.storageClass(fi.TransitionTier)
```
krisis, line 161: We should leave the storage class unmodified in this layer since FileInfo can be converted to ObjectInfo for internal uses too.

cmd/bucket-replication.go
```diff
@@ -2284,7 +2284,7 @@ func proxyHeadToRepTarget(ctx context.Context, bucket, object string, rs *HTTPRa
 			DeleteMarker:              objInfo.IsDeleteMarker,
 			ContentType:               objInfo.ContentType,
 			Expires:                   objInfo.Expires,
-			StorageClass:              objInfo.StorageClass,
+			StorageClass:              globalAPIConfig.storageClass(objInfo.StorageClass),
```
krisis, line 2287: We should return what the replica site returns here. In the scenario where the user has applications sensitive to the value returned in `X-Amz-Storage-Class` header, `api.force_storage_class` must be configured on the replica too.

cmd/batch-handlers.go
```diff
@@ -526,7 +526,7 @@ func toObjectInfo(bucket, object string, objInfo miniogo.ObjectInfo) ObjectInfo
 
 	_, ok = oi.UserDefined[xhttp.AmzStorageClass]
 	if !ok {
-		oi.UserDefined[xhttp.AmzStorageClass] = objInfo.StorageClass
+		oi.UserDefined[xhttp.AmzStorageClass] = globalAPIConfig.storageClass(objInfo.StorageClass)
```
krisis, line 529: batch-jobs interface is a MinIO interface and unencumbered by external clients. So, in this case we can return the unmodified storage class.

krisis:
@klauspost In the interest of time, I have superseded this PR with https://github.com/minio/minio/pull/19223. It is based on your changes. I have addressed my comments on this PR and also added `x-minio-tier` to the response header with the actual tier name.

harshavardhana:
This changed is merged in #19223

PR #19227, opened 2024-03-08
Fix listing in objects split across pools

klauspost:
## Description

Merging same-object - multiple versions from different pools would not always result in correct ordering.

When merging be sure to keep inputs separated instead of merging flat (unsorted) entries. Also makes sure any duplicates are removed.

```
------ before ------
λ mc ls --versions local/testbucket
[2024-03-05 20:17:19 CET]   228B STANDARD 1f163718-9bc5-4b01-bff7-5d8cf09caf10 v3 PUT hosts
[2024-03-05 20:19:56 CET]  19KiB STANDARD null v2 PUT hosts
[2024-03-05 20:17:15 CET]   228B STANDARD 73c9f651-f023-4566-b012-cc537fdb7ce2 v1 PUT hosts

------ after ------
λ mc ls --versions local/testbucket
[2024-03-05 20:19:56 CET]  19KiB STANDARD null v3 PUT hosts
[2024-03-05 20:17:19 CET]   228B STANDARD 1f163718-9bc5-4b01-bff7-5d8cf09caf10 v2 PUT hosts
[2024-03-05 20:17:15 CET]   228B STANDARD 73c9f651-f023-4566-b012-cc537fdb7ce2 v1 PUT hosts
```

MinIO doesn't split objects, but a partial decommission can end up with this.

## How to test this PR?

Manually split objects. Later versions must be in higher pool number than earlier.

Added unit test.

## Types of changes


harshavardhana:
@kannappanr ^^

harshavardhana (approved):

vadmeste (approved):
o_O

harshavardhana:
With this now we can officially support versions across pools with no problems.

Issue #15536, opened 2022-08-15
Compression configuration does not work


ArcherTrister:
I use the docker compose deployment provided in the document. When I configure the compressed environment variables according to the document, the operation succeeds but the compression configuration does not work.

```yaml
version: '3.7'

# Settings and configurations that are common for all containers
x-minio-common: &minio-common
  image: quay.io/minio/minio:latest
  command: server --console-address ":9001" http://minio{1...4}/data{1...2}
  restart: always
  expose:
    - "9000"
    - "9001"
  environment:
    TZ: Asia/Shanghai
    MINIO_ROOT_USER: admin
    MINIO_ROOT_PASSWORD: xxxxxxxxx
    MINIO_COMPRESSION_ENABLE: "on"
    MINIO_COMPRESS_EXTENSIONS: ".pdf,.doc,.png,.bmp,.jpg,.jpeg,.gif,.mp4"
    MINIO_COMPRESS_MIME_TYPES: "application/pdf,application/ogg,image/gif,image/png,image/jpeg,image/bmp,image/webp"
  healthcheck:
    test: ["CMD", "curl", "-f", "http://localhost:9000/minio/health/live"]
    interval: 30s
    timeout: 20s
    retries: 3

# starts 4 docker containers running minio server instances.
# using nginx reverse proxy, load balancing, you can access
# it through port 9000.
services:
  minio1:
    <<: *minio-common
    hostname: minio1
    volumes:
      - data1-1:/data1
      - data1-2:/data2

  minio2:
    <<: *minio-common
    hostname: minio2
    volumes:
      - data2-1:/data1
      - data2-2:/data2

  minio3:
    <<: *minio-common
    hostname: minio3
    volumes:
      - data3-1:/data1
      - data3-2:/data2

  minio4:
    <<: *minio-common
    hostname: minio4
    volumes:
      - data4-1:/data1
      - data4-2:/data2

  nginx:
    image: nginx:1.19.2-alpine
    hostname: nginx
    volumes:
      - ./nginx.conf:/etc/nginx/nginx.conf:ro
    ports:
      - "9000:9000"
      - "9001:9001"
    depends_on:
      - minio1
      - minio2
      - minio3
      - minio4

## By default this config uses default local driver,
## For custom volumes replace with volume driver configuration.
volumes:
  data1-1:
  data1-2:
  data2-1:
  data2-2:
  data3-1:
  data3-2:
  data4-1:
  data4-2:
```

[...]

PR #19233, opened 2024-03-10
Fix handling of set-but-empty environment variables

twelho:
## Description

Fix handling of set-but-empty environment variables, particularly for configuration entries for which an empty string is a valid and desired value.

## Motivation and Context

The intended way to [enable compression for all compressible objects](https://min.io/docs/minio/linux/administration/object-management/data-compression.html#compress-all-compressible-objects) involves setting `extensions` and `mime_types` to empty strings (arrays). When using [environment variables](https://min.io/docs/minio/linux/reference/minio-server/settings/core.html#id12) for configuration, this involves setting `MINIO_COMPRESSION_EXTENSIONS=""` and `MINIO_COMPRESSION_MIME_TYPES=""`, which get erroneously ignored, leading to the default allowlists of extensions and MIME types being applied instead. This can be verified with `mc` and `console`:

```shell
$ mc admin config get local compression
# MINIO_COMPRESSION_ENABLE=on
compression enable=off allow_encryption=off extensions=.txt,.log,.csv,.json,.tar,.xml,.bin mime_types=text/*,application/json,application/xml,binary/octet-stream
```

image

Observe that `MINIO_COMPRESSION_EXTENSIONS` and `MINIO_COMPRESSION_MIME_TYPES` are absent, and I have verified that compression is not being applied to custom objects due to the default `extensions` and `mime_types` lists being used.

To resolve the issue, https://github.com/minio/pkg/pull/100 is needed for the environment variable handling to be able to (again) distinguish between unset variables and variables set to an empty string. This PR includes minor semantic changes to adapt to `env.LookupEnv`, and also fixes `getTargetEnvs` to handle empty values using its improved output. Variables explicitly marked as `HiddenIfEmpty` are properly ignored.

As a result, launching MinIO with

```
MINIO_COMPRESSION_EXTENSIONS=""
MINIO_COMPRESSION_MIME_TYPES=""
```

now results in the configuration being respected:

```shell
$ mc admin config get local compression
# MINIO_COMPRESSION_ENABLE=on
# MINIO_COMPRESSION_EXTENSIONS=
# MINIO_COMPRESSION_MIME_TYPES=
compression enable=off allow_encryption=off extensions=.txt,.log,.csv,.json,.tar,.xml,.bin mime_types=text/*,application/json,application/xml,binary/octet-stream
```

image

and compression for all compressible objects works once again.

## How to test this PR?

1. Clone https://github.com/twelho/minio-pkg/tree/fix-env-empty-value
2. At the end of `go.mod`, add `replace github.com/minio/pkg/v2 => ./path/to/cloned/minio/pkg`
3. Run MinIO with `MINIO_COMPRESSION_EXTENSIONS=""` and `MINIO_COMPRESSION_MIME_TYPES=""`
4. Observe working behavior described above

## Types of changes

## Checklist:

### DIFF
--- a/cmd/config-current.go
+++ b/cmd/config-current.go
@@ -477,7 +477,7 @@ func validateConfig(ctx context.Context, s config.Config, subSys string) error {
 func lookupConfigs(s config.Config, objAPI ObjectLayer) {
 	ctx := GlobalContext
 
-	dnsURL, dnsUser, dnsPass, err := env.LookupEnv(config.EnvDNSWebhook)
+	dnsURL, _, dnsUser, dnsPass, err := env.LookupEnv(config.EnvDNSWebhook)
 	if err != nil {
 		logger.LogIf(ctx, fmt.Errorf("Unable to initialize remote webhook DNS config %w", err))
 	}
--- a/cmd/server-main.go
+++ b/cmd/server-main.go
@@ -222,13 +222,13 @@ EXAMPLES:
 }
 
 func serverCmdArgs(ctx *cli.Context) []string {
-	v, _, _, err := env.LookupEnv(config.EnvArgs)
+	v, _, _, _, err := env.LookupEnv(config.EnvArgs)
 	if err != nil {
 		logger.FatalIf(err, "Unable to validate passed arguments in %s:%s",
 			config.EnvArgs, os.Getenv(config.EnvArgs))
 	}
 	if v == "" {
-		v, _, _, err = env.LookupEnv(config.EnvVolumes)
+		v, _, _, _, err = env.LookupEnv(config.EnvVolumes)
 		if err != nil {
 			logger.FatalIf(err, "Unable to validate passed arguments in %s:%s",
 				config.EnvVolumes, os.Getenv(config.EnvVolumes))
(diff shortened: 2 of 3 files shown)


twelho:
This is affecting at least one more user in https://github.com/minio/minio/issues/15536, where compression is not being applied to all compressible objects as the user intended, but the default allowlists are erroneously used instead.

harshavardhana:
This is wrong behavior; to set it to all extensions and mime_types, it must be `*` and not an empty value. This defeats and confuses every other config subsystem implementation behavior. 

An empty value != "*"

twelho:
This get pretty confusing. If the user sets `MINIO_COMPRESSION_MIME_TYPES="*"`, everything will be compressed regardless of `MINIO_COMPRESSION_EXTENSIONS`, as intended. However, if they only set `MINIO_COMPRESSION_EXTENSIONS="*"`, the MIME filter will still apply, since the extension check does not parse regex.

twelho:
Also, when using stored configuration, should the options then be `compression extensions=* mime_types=*` instead of `compression extensions= mime_types=` as well?

twelho:
> This is wrong behavior; to set it to all extensions and mime_types, it must be `*` and not an empty value. This defeats and confuses every other config subsystem implementation behavior.

Now implemented in https://github.com/minio/minio/pull/19234.

> Also, when using stored configuration, should the options then be `compression extensions=* mime_types=*` instead of `compression extensions= mime_types=` as well?

Both of these are supported in https://github.com/minio/minio/pull/19234 to maintain existing behavior.

PR #19234, opened 2024-03-10
Improve handling of compression inclusion for objects

twelho:
## Description

Attempt 2 of https://github.com/minio/minio/pull/19233. Instead of adding support for empty environment variables, it is now possible to enable compression for all compressible objects with

```
MINIO_COMPRESSION_EXTENSIONS="*"
MINIO_COMPRESSION_MIME_TYPES="*"
```

either **individually** or **together**. Either one alone is sufficient for catching all objects (that are not part of the fixed built-in exclusion-list). This maintains the existing behavior, with the exception of adding the wildcard match to the extension handling in order for `MINIO_COMPRESSION_EXTENSIONS="*"` to be sufficient on its own for enabling catch-all compression (as the user expects). `hasStringSuffixInSlice` is otherwise only used for the fixed built-in extension list, which does not contain a wildcard, and thus its behavior is unchanged.

For the configuration variables, in addition to setting `compression extensions= mime_types=` as the current preference for enabling catch-all compression, this should also allow `compression extensions=* mime_types=*` to be used.

## How to test this PR?

1. Build MinIO from https://github.com/twelho/minio/tree/fix-compression-envs-2
2. Run with one of or both of the environment variables above, observe desired behavior according to https://github.com/minio/minio/pull/19233.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/object-api-utils.go
+++ b/cmd/object-api-utils.go
@@ -577,22 +577,35 @@ func excludeForCompression(header http.Header, object string, cfg compress.Confi
 	}
 
 	// Filter compression includes.
-	exclude := len(cfg.Extensions) > 0 || len(cfg.MimeTypes) > 0
+	if len(cfg.Extensions) == 0 && len(cfg.MimeTypes) == 0 {
+		// Nothing to filter, include everything.
+		return false
+	}
+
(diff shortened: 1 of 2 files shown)


harshavardhana (approved):
LGTM - this is clean

klauspost (approved):

PR #19259, opened 2024-03-14
add X-Content-Length

mattn:
Some proxy server like cloudflare remove Content-Length header for the cache. So `mc cat` always fail like below.

>specified offset (0) bigger than file (-1)

The mc command gets the file size from the content-length header given in the response of HEAD request, but the proxy server removes the header without knowing that this is the file size that mc command use.

This change add X-Content-Length to recover Content-Length. All of minio client should handle this X-Content-Length if Content-Length is not found.

### DIFF
--- a/cmd/api-response.go
+++ b/cmd/api-response.go
@@ -898,7 +898,9 @@ func writeResponse(w http.ResponseWriter, statusCode int, response []byte, mType
 	if mType != mimeNone {
 		w.Header().Set(xhttp.ContentType, string(mType))
 	}
-	w.Header().Set(xhttp.ContentLength, strconv.Itoa(len(response)))
+	mLength := strconv.Itoa(len(response))
+	w.Header().Set(xhttp.ContentLength, mLength)
+	w.Header().Set(xhttp.XContentLength, mLength)
 	w.WriteHeader(statusCode)
 	if response != nil {
 		w.Write(response)
--- a/internal/http/headers.go
+++ b/internal/http/headers.go
@@ -27,6 +27,7 @@ const (
 	ContentEncoding    = "Content-Encoding"
 	Expires            = "Expires"
 	ContentLength      = "Content-Length"
+	XContentLength     = "X-Content-Length"
 	ContentLanguage    = "Content-Language"
 	ContentRange       = "Content-Range"
 	Connection         = "Connection"


harshavardhana:
Unfortunately we can't make these kind of changes. What if they remove something else tomorrow.

Cloudflare has no business doing this. Nginx, Traefik none of them do it.

Please ask them to fix as they are clearly broken here.

mattn:
https://community.cloudflare.com/t/content-length-is-removed-on-response-of-head-method/628350

mattn:
BTW, mc can be check for this like below.
```
diff --git a/cmd/cat-main.go b/cmd/cat-main.go
index 131c8682..fa600932 100644
--- a/cmd/cat-main.go
+++ b/cmd/cat-main.go
@@ -251,7 +251,7 @@ func catURL(ctx context.Context, sourceURL string, encKeyDB map[string][]prefixS
 				}
 			}
 
-			if client.GetURL().Type == objectStorage {
+			if content.Size != -1 && client.GetURL().Type == objectStorage {
 				size = content.Size - o.startO
 				if size < 0 {
 					err := probe.NewError(fmt.Errorf("specified offset (%d) bigger than file (%d)", o.startO, content.Size))
```
Yet you refuse to make this change?

klauspost:
Yes. We can keep adding workarounds for all kinds of broken proxies. This will break most clients so fixing the minio/mc combination is not going to solve anything significant.

mattn:
Cloudflare says, if the server return `Cache-Control: no-transform` for HEAD request, Content-Length will be returned. 

I think this specification is reasonable; it is not an RFC violation to remove Content-Length. And I think it is the right way for minio to add `Cache-Control: no-transform` to HEAD requests to prevent it.

https://community.cloudflare.com/t/content-length-is-removed-on-response-of-head-method/628350/18?u=mattn.jp

harshavardhana:
None of this AWS S3 behavior, cache-control is part of object metadata we cannot come up new values for objects. So beyond the scope of S3 API. We have no idea what else might break if we just add these headers.

Cloudflare is simply wrong and buggy here. I don't see this with nginx or any other commonly used proxies.

harshavardhana:
They should provide a passthrough option.

PR #19268, opened 2024-03-15
[DRAFT] http webhook refactor 

zveinn:

## Description
NOTE: this description is a short one and temporary. I will write a much better one tomorrow but for now I kind of just want eyes on this.

The audit_webhook implementation was kind of broken, it would miss a lot of important errors and drop messages when configurations were reloaded. 

The new implementation will carry over old messages both from the original logCh and from a temp buffer created to hold messages that are in transit. The configurations can now be updated and removed live without loss of log entries. 
Additionally, the logic around spawning and de-spawning additional log workers was refactored to be a bit more responsive.

This change is applied for ALL http webhooks, not just audit logs. 

I've tested the performance by transferring at full speed on my local system and no problems yet. But I would like this to be tested on a system with more transfer capabilities. 

## Types of changes

### DIFF
--- a/cmd/config-current.go
+++ b/cmd/config-current.go
@@ -646,7 +646,6 @@ func applyDynamicConfigForSubSys(ctx context.Context, objAPI ObjectLayer, s conf
 			loggerCfg.AuditWebhook[n] = l
 		}
 
-		fmt.Println("UPDATING AUDIT FOR SUBSYSTEM", subSys)
 		if errs := logger.UpdateHTTPTargets(ctx, loggerCfg.AuditWebhook); len(errs) > 0 {
 			logger.LogIf(ctx, fmt.Errorf("Unable to update audit webhook targets: %v", errs))
 		}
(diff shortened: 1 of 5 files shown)


klauspost:
I feel the re-use of `logTmpCh` adds a lot of complexity. I will try doing a separate PR that separates "incoming" and "drain" channel just so we can see what differences there would be.

zveinn:
> I feel the re-use of `logTmpCh` adds a lot of complexity. I will try doing a separate PR that separates "incoming" and "drain" channel just so we can see what differences there would be.

The logTmpCh is not re-used, the logTmpCh is drained from the original target to the new. I tried doing it where I drained dirrectly into the logCh, but because the user can modify the size of the queue it can become a blocking opperation. Especially when/if the user provides an invalid config on update. Then you suddenly have a situation where you now have a logCh which is much smaller (f.x.) then the logTmpCh and since nothing is pulling of that channel, the drain mechanism will not complete... at which point when the config is reloaded again, the remaining entires in the original logTmpCh will be lost.

zveinn:
NOTE: there is a bug in the PR regarding updating auditTargets vs systemTargets, me and @klauspost are working on an update (mostly klaus though)

PR #19269, opened 2024-03-15
write anything beyond 4k to be written in 4k pages

harshavardhana:

## Description
write anything beyond 4k to be written in 4k pages

## Motivation and Context
we were prematurely not writing 4k pages while we
could have because most buffers would be multiples 
of 4k up to some number, and there shall be some 
remainder.

We only need to write the remainder without O_DIRECT
not the entire buffer.

## How to test this PR?
CI/CD must test everything end-to-end

## Types of changes

## Checklist:

### DIFF
--- a/internal/ioutil/ioutil.go
+++ b/internal/ioutil/ioutil.go
@@ -20,7 +20,6 @@
 package ioutil
 
 import (
-	"bytes"
 	"context"
 	"errors"
 	"io"
@@ -36,8 +35,8 @@ import (
 // Block sizes constant.
 const (
 	BlockSizeSmall       = 32 * humanize.KiByte // Default r/w block size for smaller objects.
-	BlockSizeLarge       = 2 * humanize.MiByte  // Default r/w block size for larger objects.
-	BlockSizeReallyLarge = 4 * humanize.MiByte  // Default write block size for objects per shard >= 64MiB
+	BlockSizeLarge       = 1 * humanize.MiByte  // Default r/w block size for normal objects.
+	BlockSizeReallyLarge = 4 * humanize.MiByte  // Default r/w block size for very large objects.
 )
 
 // aligned sync.Pool's
@@ -341,19 +340,6 @@ func CopyAligned(w io.Writer, r io.Reader, alignedBuf []byte, totalSize int64, f
 		return 0, nil
 	}
 
-	// Writes remaining bytes in the buffer.
-	writeUnaligned := func(w io.Writer, buf []byte) (remainingWritten int64, err error) {
-		// Disable O_DIRECT on fd's on unaligned buffer
-		// perform an amortized Fdatasync(fd) on the fd at
-		// the end, this is performed by the caller before
-		// closing 'w'.
-		if err = disk.DisableDirectIO(file); err != nil {
-			return remainingWritten, err
-		}
-		// Since w is *os.File io.Copy shall use ReadFrom() call.
-		return io.Copy(w, bytes.NewReader(buf))
-	}
-
 	var written int64
 	for {
 		buf := alignedBuf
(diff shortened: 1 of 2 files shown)


internal/ioutil/ioutil.go
```diff
@@ -371,15 +357,36 @@ func CopyAligned(w io.Writer, r io.Reader, alignedBuf []byte, totalSize int64, f
...
+		if remain == 0 {
 			// buf is aligned for directio write()
 			n, err = w.Write(buf)
 			nw = int64(n)
 		} else {
+			n, err = w.Write(buf[:len(buf)-remain])
+			if err != nil {
+				return written, err
+			}
+			nw = int64(n)
```
klauspost:
```suggestion
			if remain < len(buf) {
				n, err = w.Write(buf[:len(buf)-remain])
				if err != nil {
					return written, err
				}
				nw = int64(n)
			}
```

internal/ioutil/ioutil.go
```diff
@@ -35,9 +34,9 @@ import (
 
 // Block sizes constant.
 const (
-	BlockSizeSmall       = 32 * humanize.KiByte // Default r/w block size for smaller objects.
-	BlockSizeLarge       = 2 * humanize.MiByte  // Default r/w block size for larger objects.
-	BlockSizeReallyLarge = 4 * humanize.MiByte  // Default write block size for objects per shard >= 64MiB
+	BlockSizeSmall       = 128 * humanize.KiByte // Default r/w block size for smaller objects.
+	BlockSizeLarge       = 2 * humanize.MiByte   // Default r/w block size for larger objects.
+	BlockSizeReallyLarge = 4 * humanize.MiByte   // Default write block size for objects per shard >= 64MiB
```
klauspost: I think we are writing too many small files for this to not cause too much overallocation.

So right now "small" is used for < 128K. So we would be looking at max 4 syscalls to write the file.

We can make the `BlockSizeLarge` cheaper and differentiate it more from "ReallyLarge":

```suggestion
	BlockSizeSmall       = 32 * humanize.KiByte // Default r/w block size for smaller objects.
	BlockSizeLarge       = 1 * humanize.MiByte   // Default r/w block size for larger objects.
	BlockSizeReallyLarge = 4 * humanize.MiByte   // Default write block size for big objects.
```

I would then suggest changing this: https://github.com/minio/minio/blob/88a89213ff6663f2d3ee1de7f3293497ab9844e8/cmd/xl-storage.go#L2115-L2126 to

```Go
	switch {
	case fileSize > 0 && fileSize >= xioutil.BlockSizeReallyLarge:
		// use a larger 4MiB buffer for a really large streams.
		bufp = xioutil.ODirectPoolXLarge.Get().(*[]byte)
		defer xioutil.ODirectPoolXLarge.Put(bufp)
	case fileSize <= xioutil.BlockSizeSmall:
		bufp = xioutil.ODirectPoolSmall.Get().(*[]byte)
		defer xioutil.ODirectPoolSmall.Put(bufp)
	default:
		bufp = xioutil.ODirectPoolLarge.Get().(*[]byte)
		defer xioutil.ODirectPoolLarge.Put(bufp)
	}
```

So everything up to 4MiB is written with a single syscall (plus DirectIO splitting).

We could set the switchover to `xioutil.BlockSizeReallyLarge*2` and `xioutil.BlockSizeSmall*2` to reduce to overalloc a bit - but from a pure performance point the above will probably be the best option.

internal/ioutil/ioutil.go
```diff
@@ -371,15 +357,36 @@ func CopyAligned(w io.Writer, r io.Reader, alignedBuf []byte, totalSize int64, f
...
-		if len(buf)%DirectioAlignSize == 0 {
-			var n int
+		var (
+			n  int
+			un int
+			nw int64
+		)
+
+		remain := len(buf) % DirectioAlignSize
+		if remain == 0 {
```
klauspost, line 367: No need for two writes if platform does not have directio.
```suggestion
		if !disk.ODirectPlatform || remain == 0 {
```

klauspost (commented):
I think we can make this "sharper".

internal/ioutil/ioutil.go
harshavardhana: this won't even come here.

klauspost (approved):
👍🏼 Should reduce syscalls nicely!

poornas (approved):

PR #19291, opened 2024-03-18
batch rotate filter() must return true to skip the matching objects

harshavardhana:

## Description
batch rotate filter() must return true to skip the matching objects

## Motivation and Context
the current implementation was doing the reverse

## How to test this PR?
```yaml
keyrotate:
  apiVersion: v1
  bucket: heal-bucket1-1
  encryption:
    type: sse-kms # valid values are sse-s3 and sse-kms
    key: "my-minio-test-key" # valid only for sse-kms

  # optional flags based filtering criteria
  # for all objects
  flags:
    filter:
      createdAfter: "2025-03-15T15:04:05Z" # match objects created after "date"
    retry:
```

The following filter would apply in reverse as createdBefore without this PR.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/batch-rotate.go
+++ b/cmd/batch-rotate.go
@@ -274,7 +274,7 @@ func (r *BatchJobKeyRotateV1) Start(ctx context.Context, api ObjectLayer, job Ba
 
 	rnd := rand.New(rand.NewSource(time.Now().UnixNano()))
 
-	skip := func(info FileInfo) (ok bool) {
+	selectObj := func(info FileInfo) (ok bool) {
 		if r.Flags.Filter.OlderThan > 0 && time.Since(info.ModTime) < r.Flags.Filter.OlderThan {
 			// skip all objects that are newer than specified older duration
 			return false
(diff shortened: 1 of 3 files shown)


cmd/batch-rotate.go
```diff
@@ -277,22 +277,22 @@ func (r *BatchJobKeyRotateV1) Start(ctx context.Context, api ObjectLayer, job Ba
 	skip := func(info FileInfo) (ok bool) {
 		if r.Flags.Filter.OlderThan > 0 && time.Since(info.ModTime) < r.Flags.Filter.OlderThan {
 			// skip all objects that are newer than specified older duration
-			return false
+			return true
```
poornas: skip was written this way to filter by more than one criteria - e.g. metadata match and creation date criteria, which will not work with this change prematurely matching just one.  Perhaps just invert previous when passing to Walk Filter opts?

poornas (changes requested):

poornas (approved):

cmd/batch-rotate.go
```diff
@@ -336,7 +336,7 @@ func (r *BatchJobKeyRotateV1) Start(ctx context.Context, api ObjectLayer, job Ba
 			return false
 		}
 		if r.Flags.Filter.KMSKeyID != "" {
-			if v, ok := info.Metadata[xhttp.AmzServerSideEncryptionKmsID]; ok && strings.TrimPrefix(v, crypto.ARNPrefix) != r.Flags.Filter.KMSKeyID {
+			if v, ok := info.Metadata[xhttp.AmzServerSideEncryptionKmsID]; ok && strings.TrimPrefix(v, crypto.ARNPrefix) == r.Flags.Filter.KMSKeyID {
```
vadmeste, line 339: I do not think this is correct. It is correct to rename skip to select object, but here we are returning false when the kms key id matches?

vadmeste (changes requested):

PR #19305, opened 2024-03-20
Encode dir obj names before expiration

krisis:

## Description
Object names of directory objects qualified for ExpiredObjectAllVersions must be encoded appropriately before calling on deletePrefix on their erasure set.

e.g, a directory object and regular objects with overlapping prefixes could lead to unintended expiration of the regular objects

```
bucket/dir/ ---> directory object
bucket/dir/obj-1
```

When `bucket/dir/` qualifies for expiration, current implementation would remove regular objects under the prefix `bucket/dir/`, in this case, `bucket/dir/obj-1`.

## How to test this PR?
Setup MinIO and run the following script. At the end of the script, `mc ls -r --versions myminio/bucket/dir` must list regular objects (i.e not directory objects).
The test creates a situation where the directory object `dir` under `bucket` qualifies for expiry with overlapping regular objects.
```
#!/bin/bash

mc mb myminio/bucket
mc version enable myminio/bucket
mc anonymous set public myminio/bucket

for _ in $(seq 1 2)
do
   mc cp /etc/hosts myminio/bucket/dir/obj-1
done

curl -X PUT http://127.0.0.1:9000/bucket/dir/  -H 'X-Minio-Source-Mtime: 2024-01-13T12:02:04.115209866-08:00'

mc ilm rule add --expire-days 10 --expire-all-object-versions myminio/bucket/dir/

mc ls -r --versions myminio/bucket/dir
```

## Types of changes

## Checklist:

### DIFF
--- a/cmd/batch-expire.go
+++ b/cmd/batch-expire.go
@@ -427,7 +427,7 @@ func batchObjsForDelete(ctx context.Context, r *BatchJobExpire, ri *batchJobInfo
 					default:
 					}
 					stopFn := globalBatchJobsMetrics.trace(batchJobMetricExpire, ri.JobID, attempts)
-					_, err := api.DeleteObject(ctx, exp.Bucket, exp.Name, ObjectOptions{
+					_, err := api.DeleteObject(ctx, exp.Bucket, encodeDirObject(exp.Name), ObjectOptions{
 						DeletePrefix: true,
 					})
 					if err != nil {
--- a/cmd/data-scanner.go
+++ b/cmd/data-scanner.go
@@ -1261,7 +1261,9 @@ func applyExpiryOnNonTransitionedObjects(ctx context.Context, objLayer ObjectLay
 	if obj.DeleteMarker {
 		eventName = event.ObjectRemovedDeleteMarkerCreated
 	}
-
+	if lcEvent.Action.DeleteAll() {
+		eventName = event.ObjectRemovedDeleteAllVersions
+	}
 	// Notify object deleted event.
 	sendEvent(eventArgs{
 		EventName:  eventName,
--- a/cmd/data-scanner.go
+++ b/cmd/data-scanner.go
@@ -1240,7 +1240,7 @@ func applyExpiryOnNonTransitionedObjects(ctx context.Context, objLayer ObjectLay
 		}
 	}()
 
-	dobj, err = objLayer.DeleteObject(ctx, obj.Bucket, obj.Name, opts)
+	dobj, err = objLayer.DeleteObject(ctx, obj.Bucket, encodeDirObject(obj.Name), opts)
 	if err != nil {
 		if isErrObjectNotFound(err) || isErrVersionNotFound(err) {
 			return false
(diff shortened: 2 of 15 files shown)


harshavardhana (commented):
while we are at this can we add another `event.ObjectRemovedDelete` to be `event.ObjectRemovedDeleteAllVersions` to differentiate ExpireAll action?

krisis:
> while we are at this can we add another `event.ObjectRemovedDelete` to be `event.ObjectRemovedDeleteAllVersions` to differentiate ExpireAll action?

It would be better if we did this in a separate PR. This involves adding more than one event name for lifecycle actions such as DeleteRestoredAction, DeleteRestoredVersionAction, etc.

harshavardhana:
> It would be better if we did this in a separate PR. This involves adding more than one event name for lifecycle actions such as DeleteRestoredAction, DeleteRestoredVersionAction, etc.

My idea was to make this cherry-pick friendly, anyway I pushed the relevant change to this PR.

harshavardhana (approved):

cmd/batch-expire.go
```diff
@@ -427,7 +427,7 @@ func batchObjsForDelete(ctx context.Context, r *BatchJobExpire, ri *batchJobInfo
 					default:
 					}
 					stopFn := globalBatchJobsMetrics.trace(batchJobMetricExpire, ri.JobID, attempts)
-					_, err := api.DeleteObject(ctx, exp.Bucket, exp.Name, ObjectOptions{
+					_, err := api.DeleteObject(ctx, exp.Bucket, encodeDirObject(exp.Name), ObjectOptions{
```
vadmeste, line 430: I can see this is encoded inside erasureServerPools.DeleteObject() as well, I do not think this is needed

cmd/data-scanner.go
```diff
@@ -1240,7 +1240,7 @@ func applyExpiryOnNonTransitionedObjects(ctx context.Context, objLayer ObjectLay
 		}
 	}()
 
-	dobj, err = objLayer.DeleteObject(ctx, obj.Bucket, obj.Name, opts)
+	dobj, err = objLayer.DeleteObject(ctx, obj.Bucket, encodeDirObject(obj.Name), opts)
```
vadmeste, line 1243: I can see this is encoded inside erasureServerPools.DeleteObject() as well, I do not think this is needed

cmd/batch-expire.go
krisis: In erasureServerPools.DeleteObject, encodeDirObject is called only when opts.DeletePrefix is false. In both these cases opts.DeletePrefix is true.

cmd/data-scanner.go
harshavardhana: @vadmeste it is not - it is doing extra work see why this is needed from Decommission code.

cmd/batch-expire.go
harshavardhana: @vadmeste it is not - it is doing extra work see why this is needed from Decommission code.

PR #19336, opened 2024-03-25
allow configuring inline shard size value

harshavardhana:

## Description
allow configuring inline shard size value

## Motivation and Context
performance of inline v/s non-inline seems to be 2x 
difference - allow configuring this in some
situations when a bigger stripe for 1MiB object size
would fall under the inline category.

## How to test this PR?
We need to do performance numbers to see the difference
allow this to be configurable for now. 

## Types of changes

## Checklist:

### DIFF
--- a/internal/config/storageclass/storage-class.go
+++ b/internal/config/storageclass/storage-class.go
@@ -18,13 +18,16 @@
 package storageclass
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"strconv"
 	"strings"
 	"sync"
 
+	"github.com/dustin/go-humanize"
 	"github.com/minio/minio/internal/config"
+	"github.com/minio/minio/internal/logger"
 	"github.com/minio/pkg/v2/env"
 )
 
@@ -40,7 +43,8 @@ const (
 const (
 	ClassStandard = "standard"
 	ClassRRS      = "rrs"
-	ClassOptimize = "optimize"
+	Optimize      = "optimize"
+	InlineBlock   = "inline_block"
 
 	// Reduced redundancy storage class environment variable
 	RRSEnv = "MINIO_STORAGE_CLASS_RRS"
@@ -48,6 +52,14 @@ const (
 	StandardEnv = "MINIO_STORAGE_CLASS_STANDARD"
 	// Optimize storage class environment variable
 	OptimizeEnv = "MINIO_STORAGE_CLASS_OPTIMIZE"
+	// Inline block indicates the size of the shard
+	// that is considered for inlining, remember this
+	// shard value is the value per drive shard it
+	// will vary based on the parity that is configured
+	// for the STANDARD storage_class.
+	// inlining means data and metadata are written
+	// together in a single file i.e xl.meta
+	InlineBlockEnv = "MINIO_STORAGE_CLASS_INLINE_BLOCK"
 
 	// Supported storage class scheme is EC
 	schemePrefix = "EC"
@@ -71,8 +83,13 @@ var (
 			Value: "EC:1",
 		},
 		config.KV{
-			Key:   ClassOptimize,
+			Key:   Optimize,
 			Value: "availability",
+		},
+		config.KV{
+			Key:           InlineBlock,
+			Value:         "",
+			HiddenIfEmpty: true,
 		},
 	}
 )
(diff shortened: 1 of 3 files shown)


poornas (approved):

internal/config/storageclass/storage-class.go
```diff
@@ -40,14 +43,23 @@ const (
...
 
 	// Reduced redundancy storage class environment variable
 	RRSEnv = "MINIO_STORAGE_CLASS_RRS"
 	// Standard storage class environment variable
 	StandardEnv = "MINIO_STORAGE_CLASS_STANDARD"
 	// Optimize storage class environment variable
 	OptimizeEnv = "MINIO_STORAGE_CLASS_OPTIMIZE"
+	// Inline block indicates the size of the shard
+	// that is considered for inlining, remember this
+	// shard value is the value per drive shard it
```
shtripat, line 57:
```suggestion
	// shard value is the value per drive shard. It
```

internal/config/storageclass/storage-class.go
```diff
@@ -355,7 +386,23 @@ func LookupConfig(kvs config.KVS, setDriveCount int) (cfg Config, err error) {
...
+	cfg.Optimize = env.Get(OptimizeEnv, kvs.Get(Optimize))
+
+	inlineBlockStr := env.Get(InlineBlockEnv, kvs.GetWithDefault(InlineBlock, DefaultKVS))
+	if inlineBlockStr != "" {
+		inlineBlock, err := humanize.ParseBytes(inlineBlockStr)
+		if err != nil {
+			return cfg, err
+		}
+		if inlineBlock > 128*humanize.KiByte {
+			logger.LogOnceIf(context.Background(), fmt.Errorf("inline block value bigger than recommended max of 128KiB -> %s, performance may degrade for PUT please benchmark the changes", inlineBlockStr), inlineBlockStr)
```
shtripat, line 399:
```suggestion
			logger.LogOnceIf(context.Background(), fmt.Errorf("inline block value bigger than recommended max of 128KiB -> %s, performance may degrade for PUT. Please benchmark the changes", inlineBlockStr), inlineBlockStr)
```

shtripat (approved):

internal/config/storageclass/storage-class.go
```diff
@@ -40,14 +43,23 @@ const (
...
 	// Optimize storage class environment variable
 	OptimizeEnv = "MINIO_STORAGE_CLASS_OPTIMIZE"
+	// Inline block indicates the size of the shard
+	// that is considered for inlining, remember this
+	// shard value is the value per drive shard. It
+	// will vary based on the parity that is configured
+	// for the STANDARD storage_class.
+	// inlining means data and metadata are written
+	// together in a single file i.e xl.meta
+	InlineBlockEnv = "MINIO_STORAGE_CLASS_INLINE_BLOCK"
```
klauspost, line 62:
```suggestion
	InlineBlockEnv = "_MINIO_STORAGE_CLASS_INLINE_BLOCK"
```

We should not make this an official env var. It should not be needed. We are not C*ph - we should figure out where the difference lies instead of just pushing the problem to the user.

internal/config/storageclass/storage-class.go
```diff
@@ -90,6 +106,8 @@ type Config struct {
 	Standard    StorageClass `json:"standard"`
 	RRS         StorageClass `json:"rrs"`
 	Optimize    string       `json:"optimize"`
+	inlineBlock int64
```
klauspost, line 110: So this isn't stored, correct?

harshavardhana: This config has nothing to do with storing, this is just in-memory representation

harshavardhana:
The feature is now hidden and will not be available via `mc admin config get/set` - this is only useful for exceptional cases where we want to test these things.

Since the config and its ENV is unavailable there is no point adding _MINIO on it.

harshavardhana:
this does not need a doc @feorlen @djwfyi

krisis (approved):
Inline block value is not configurable nor discoverable via `mc admin config get/set` commands. It is only configurable via environment variables, albeit an exported one, i.e starting with `MINIO_`. It is not persisted in MinIO config. These precautions seem sufficient for introducing this experimental setting.

PR #19348, opened 2024-03-26
crank up values for grid/connection for 500 node deployments

harshavardhana:

## Description
crank up values for grid/connection for 500 node deployments

## Motivation and Context
up the default values to handle a bigger scale

## How to test this PR?
Seen to benefit in 500 node deployments with
less contention.

## Types of changes

## Checklist:

### DIFF
--- a/internal/grid/connection.go
+++ b/internal/grid/connection.go
@@ -175,7 +175,7 @@ func (c ContextDialer) DialContext(ctx context.Context, network, address string)
 }
 
 const (
-	defaultOutQueue    = 65535    // kind of close to max open fds per user
+	defaultOutQueue    = 1048576  // (ulimit -Hn) - kind of close to max open fds per user
 	readBufferSize     = 32 << 10 // 32 KiB is the most optimal on Linux
 	writeBufferSize    = 32 << 10 // 32 KiB is the most optimal on Linux
 	defaultDialTimeout = 2 * time.Second
@@ -206,8 +206,8 @@ func newConnection(o connectionParams) *Connection {
 		Local:              o.local,
 		id:                 o.id,
 		ctx:                o.ctx,
-		outgoing:           xsync.NewMapOfPresized[uint64, *muxClient](1000),
-		inStream:           xsync.NewMapOfPresized[uint64, *muxServer](1000),
+		outgoing:           xsync.NewMapOfPresized[uint64, *muxClient](2000),
+		inStream:           xsync.NewMapOfPresized[uint64, *muxServer](2000),
 		outQueue:           make(chan []byte, defaultOutQueue),
 		dialer:             o.dial,
 		side:               ws.StateServerSide,
--- a/internal/grid/grid.go
+++ b/internal/grid/grid.go
@@ -38,7 +38,7 @@ const (
 	minBufferSize = 1 << 10
 
 	// defaultBufferSize is the default buffer allocation size.
-	defaultBufferSize = 4 << 10
+	defaultBufferSize = 8 << 10
 
 	// maxBufferSize is the maximum buffer size.
 	// Buffers larger than this is not reused.


internal/grid/connection.go
```diff
@@ -175,7 +175,7 @@ func (c ContextDialer) DialContext(ctx context.Context, network, address string)
 }
 
 const (
-	defaultOutQueue    = 65535    // kind of close to max open fds per user
+	defaultOutQueue    = 1048576  // (ulimit -Hn) - kind of close to max open fds per user
```
klauspost, line 178:
```suggestion
	defaultOutQueue    = 65535
```
This has nothing to do with FDs, and is a *per connection* property, so there is no need to scale it.

You are just wasting memory.

internal/grid/connection.go
```diff
@@ -206,8 +206,8 @@ func newConnection(o connectionParams) *Connection {
 		Local:              o.local,
 		id:                 o.id,
 		ctx:                o.ctx,
-		outgoing:           xsync.NewMapOfPresized[uint64, *muxClient](1000),
-		inStream:           xsync.NewMapOfPresized[uint64, *muxServer](1000),
+		outgoing:           xsync.NewMapOfPresized[uint64, *muxClient](2000),
+		inStream:           xsync.NewMapOfPresized[uint64, *muxServer](2000),
```
klauspost, line 210:
```suggestion
		outgoing:           xsync.NewMapOfPresized[uint64, *muxClient](1000),
		inStream:           xsync.NewMapOfPresized[uint64, *muxServer](1000),
```

Again a per connection property - and just a pre-alloc anyway.

internal/grid/grid.go
```diff
@@ -38,7 +38,7 @@ const (
 	minBufferSize = 1 << 10
 
 	// defaultBufferSize is the default buffer allocation size.
-	defaultBufferSize = 4 << 10
+	defaultBufferSize = 8 << 10
```
klauspost, line 41: Most messages are significantly below 4K. this is just wasting memory.

```suggestion
	defaultBufferSize = 4 << 10
```

klauspost (changes requested):
None of these changes make sense - in particular for bigger installs.

harshavardhana:
> None of these changes make sense - in particular for bigger installs.

@klauspost, these changes are here to test this on that scale. We have not done anything right now to claim one way or the other.

I increased from 10000 -> 65355 for the 300 node deployment, and this customer is finally not complaining about latencies for most common calls anymore.

This is in line with those changes. When the 500+ node is deployed, we will validate this properly. So, hold before you make guesses; sometimes, we will be surprised how these defaults don't make sense and can come in the way.

klauspost:
@harshavardhana You are seeing snake oil.

None of these values will make any other difference whether it is 5 servers or 500 other than eating up more memory - and it fact the more servers the more memory you are wasting. 

Increasing queue sizes is not going to decrease latency. In fact it will do the opposite. Allocating buffers that are too large just wastes memory and more time will be spent zeroing values that are never used. A 10K message queue between any two servers is more than plenty. If that is filled, having 65K or 1M is not going to decrease latency whatsoever. 

The queue and mux prealloc are *per remote connection* meaning if you have 500 servers the memory used is multiplied by 499. So this scales automatically.

The 65K is "ok", but the 1M is a horrible idea - if you have 1M * 8192 * 499 you have are proposing *4TiB* (plus channel and slice) of queue data, which will just completely kill any server. So you are setting up yourself for failure.

Having more than 1K muxes on a single connections is not very likely - and if it happens the map will just scale anyway. 2K will not make a big difference, but the change is pointless.

klauspost:
@harshavardhana This is a grab from a "300 node cluster" with `defaultOutQueue` at 65535:

image

Math is "65535 * (3  * 8) (size of each queue entry) * 299 (connections) = 470,279,160 bytes".  With your change it will grow to 7.5GB on 300 nodes.  12.5GB on 500 nodes.

PR #19363, opened 2024-03-28
expire STS in Load() instead of re-listing later

harshavardhana:
## Community Contribution License
All community contributions in this pull request are licensed to the project maintainers
under the terms of the [Apache 2 license](https://www.apache.org/licenses/LICENSE-2.0). 
By creating this pull request I represent that I have the right to license the 
contributions to the project maintainers under the Apache 2 license.

## Description
expire STS in Load() instead of re-listing later

## Motivation and Context
avoids unnecessary map allocation and purges STS
in Load() instead of a separate caller.

## How to test this PR?
Simplify expiring STS. 

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Optimization (provides speedup with no functional changes)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [ ] Fixes a regression (If yes, please add `commit-id` or `PR #` here)
- [ ] Unit tests added/updated
- [ ] Internal documentation updated
- [ ] Create a documentation update request [here](https://github.com/minio/docs/issues/new?label=doc-change,title=Doc+Updated+Needed+For+PR+github.com%2fminio%2fminio%2fpull%2fNNNNN)

PR #19376, opened 2024-03-28
fix: sts accounts map refresh and fewer list calls

donatello:
## Description

This fixes a bug where STS Accounts map accumulates accounts in memory
and never removes expired accounts and the STS Policy mappings were not
being refreshed.

The STS purge routine now runs with every IAM credentials load instead
of every 4th time.

The listing of IAM files is now cached on every IAM load operation to
prevent re-listing for STS accounts purging/reload.

Additionally this change makes each server pick a time for IAM loading
that is randomly distributed from a 10 minute interval - this is to
prevent server from thundering while performing the IAM load.

On average, IAM loading will happen between every 5-15min after the
previous IAM load operation completes.

## Motivation and Context

Proposal to replace https://github.com/minio/minio/pull/19363

## Types of changes

## Checklist:

### DIFF
--- a/cmd/iam-object-store.go
+++ b/cmd/iam-object-store.go
@@ -25,12 +25,12 @@ import (
 	"path"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 	"unicode/utf8"
 
 	jsoniter "github.com/json-iterator/go"
 	"github.com/minio/madmin-go/v3"
-	"github.com/minio/minio-go/v7/pkg/set"
 	"github.com/minio/minio/internal/config"
 	xioutil "github.com/minio/minio/internal/ioutil"
 	"github.com/minio/minio/internal/kms"
@@ -44,6 +44,8 @@ type IAMObjectStore struct {
 	sync.RWMutex
 
 	*iamCache
+
+	cachedIAMListing atomic.Value
 
 	usersSysType UsersSysType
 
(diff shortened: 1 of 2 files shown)


harshavardhana (approved):

cmd/iam-object-store.go
```diff
@@ -450,22 +433,48 @@ func (iamOS *IAMObjectStore) PurgeExpiredSTS(ctx context.Context) error {
...
+
 	// Scan STS users on disk and purge expired ones. We do not need to hold a
 	// lock with store.lock() here.
-	for item := range listIAMConfigItems(ctx, iamOS.objAPI, iamConfigPrefix+SlashSeparator+stsListKey) {
-		if item.Err != nil {
-			return item.Err
+	stsAccountsFromStore := map[string]UserIdentity{}
+	stsAccPoliciesFromStore := xsync.NewMapOf[string, MappedPolicy]()
+	for _, item := range iamListing[stsListKey] {
+		fmt.Printf("Purge: item: %s\n", item)
```
poornas: can be removed?

cmd/iam-object-store.go
```diff
@@ -450,22 +433,48 @@ func (iamOS *IAMObjectStore) PurgeExpiredSTS(ctx context.Context) error {
...
+		// to keep the loaded user around in memory, so we reinitialize the map
+		// each time.
+		if err := iamOS.loadUser(ctx, userName, stsUser, stsAccountsFromStore); err != nil && err != errNoSuchUser {
+			logger.LogIf(GlobalContext, fmt.Errorf("unable to load user during STS purge: %w (%s)", err, item))
 		}
-		userName := path.Dir(item.Item)
+
+	}
+	for _, item := range iamListing[policyDBSTSUsersListKey] {
+		fmt.Printf("Purge Pol: item: %s\n", item)
```
poornas:
```suggestion
```

cmd/iam-object-store.go
```diff
@@ -450,22 +433,48 @@ func (iamOS *IAMObjectStore) PurgeExpiredSTS(ctx context.Context) error {
...
+		stsName := strings.TrimSuffix(item, ".json")
 		// loadUser() will delete expired user during the load - we do not need
 		// to keep the loaded user around in memory, so we reinitialize the map
 		// each time.
-		m := map[string]UserIdentity{}
-		if err := iamOS.loadUser(ctx, userName, stsUser, m); err != nil && err != errNoSuchUser {
-			logger.LogIf(GlobalContext, fmt.Errorf("unable to load user during STS purge: %w (%s)", err, item.Item))
+		if err := iamOS.loadMappedPolicy(ctx, stsName, stsUser, false, stsAccPoliciesFromStore); err != nil && err != errNoSuchUser {
```
poornas:
```suggestion
		if err := iamOS.loadMappedPolicy(ctx, stsName, stsUser, false, stsAccPoliciesFromStore); err != nil && !errors.Is(err, errNoSuchPolicy){
```

poornas (changes requested):

poornas (approved):

PR #19378, opened 2024-03-28
Suppress error log for force-deleting object in locked bucket

kaankabalak:

## Description
Suppress error message for force-deleting an object in a lock-enabled bucket.

## Motivation and Context
MinIO is already returning the correct response, therefore, we don't need to return this error message.

## How to test this PR?
Verify that the error message is not being seen when trying to force-delete an object in a bucket with locking enabled.

## Types of changes

## Checklist:


cmd/object-handlers.go
```diff
@@ -2830,7 +2830,6 @@ func (api objectAPIHandlers) DeleteObjectHandler(w http.ResponseWriter, r *http.
 
 	rcfg, _ := globalBucketObjectLockSys.Get(bucket)
 	if rcfg.LockEnabled && opts.DeletePrefix {
-		writeErrorResponse(ctx, w, toAPIError(ctx, errors.New("force-delete is forbidden in a locked-enabled bucket")), r.URL)
```
harshavardhana, line 2829: We have to return this error however we have to convert it into something more meaningful so that the logging doesn't get triggered.

First, do this

```
apiErr := toAPIError(ctx, errInvalidArgument)
apiErr.Description = "force-delete is forbidden on Object Locking enabled buckets" 
writeErrorResponse(ctx, w, apiErr, r.URL)
return
```

harshavardhana (changes requested):

cmd/object-handlers.go
kaankabalak: Thank you very much @harshavardhana, I wasn't aware of the full requirements where we want to keep the error but not log it. I have just pushed the changes you have suggested. 👍

harshavardhana (approved):

kannappanr (approved):

PR #19358, opened 2024-03-27
ldap: improve normalization of DN values

donatello:
## Description

Instead of relying on user input values, we use the DN value returned by the LDAP server.

This handles cases like when a mapping is set on a DN value `uid=svc.algorithm,OU=swengg,DC=min,DC=io` with a user input value (with unicode variation) of `uid=svc﹒algorithm,OU=swengg,DC=min,DC=io`. The LDAP server on lookup of this DN returns the normalized value where the unicode dot character `SMALL FULL STOP` (in the user input), gets replaced with regular full stop.

## How to test this PR?

See included test.

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Optimization (provides speedup with no functional changes)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [ ] Fixes a regression (If yes, please add `commit-id` or `PR #` here)
- [ ] Unit tests added/updated
- [ ] Internal documentation updated
- [ ] Create a documentation update request [here](https://github.com/minio/docs/issues/new?label=doc-change,title=Doc+Updated+Needed+For+PR+github.com%2fminio%2fminio%2fpull%2fNNNNN)

PR #19406, opened 2024-04-03
fix: ldap: use validated base DNs

donatello:
## Description

This fixes a regression from #19358 which prevents policy mappings created in the latest release from being displayed in policy entity listing APIs.

This is due to the possibility that the base DNs in the LDAP config are not in a normalized form and #19358 introduced normalized of mapping keys (user DNs and group DNs). When listing, we check if the policy mappings are on entities that parse as valid DNs that are descendants of the base DNs in the config.

## How to test this PR?

Use a non-normalized DN in the base DN config like `DC=min,dc=io` instead of `dc=min,dc=io` for example.

## Types of changes

## Checklist:
- [x] Fixes a regression (#19358)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -54,7 +54,7 @@ require (
 	github.com/minio/madmin-go/v3 v3.0.50
 	github.com/minio/minio-go/v7 v7.0.69
 	github.com/minio/mux v1.9.0
-	github.com/minio/pkg/v2 v2.0.14
+	github.com/minio/pkg/v2 v2.0.16
 	github.com/minio/selfupdate v0.6.0
 	github.com/minio/sha256-simd v1.0.1
 	github.com/minio/simdjson-go v0.4.5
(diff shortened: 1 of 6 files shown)


harshavardhana (changes requested):
Can you add some unit tests to cover this scenario?

donatello:
Needs https://github.com/minio/pkg/pull/106

go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/minio/minio
 
 go 1.21
 
+replace github.com/minio/pkg/v2 => github.com/donatello/pkg/v2 v2.0.0-20240403174644-37c5f6a91d42
```
harshavardhana:
```suggestion
replace github.com/minio/pkg/v2 => github.com/donatello/pkg/v2 v2.0.0-20240403174644-37c5f6a91d42
```

v2.0.16 released

vadmeste (commented):
LGTM, except the import in go.mod as mentionned by Harsha

harshavardhana (approved):

PR #19410, opened 2024-04-04
heal: Persist MRF queue in the disk during shutdown

vadmeste:

## Types of changes

## Checklist:


cmd/mrf.go
```diff
@@ -15,50 +15,201 @@
...
+			d, err := obj.GetDisks(-1, -1, true)
+			if err != nil {
+				healingLogIf(context.Background(), err)
+				return
+			}
+			localDrives = d
+			break
+		}
+		time.Sleep(5 * time.Second)
+	}
```
harshavardhana: just use globalLocalDrives why do you need this?

cmd/erasure-server-pool.go
```diff
@@ -294,11 +295,30 @@ func (z *erasureServerPools) GetRawData(ctx context.Context, volume, file string
...
+		}
+		disks = z.serverPools[poolIdx].sets[setIdx].getDisks()
+	} else {
+		// poolIdx < 0 means collect all disks in this cluster
+		for pool := range z.serverPools {
+			for set := range z.serverPools[pool].sets {
+				disks = append(disks, z.serverPools[pool].sets[set].getDisks()...)
+			}
+		}
 	}
```
harshavardhana: just use globalLocalDrives why do you need this?

harshavardhana (changes requested):

cmd/erasure-server-pool.go
vadmeste: I like calling object-api.GetDisks() because we know for sure that disks are ready but I see know globalLocalDrives are initialized very early in the code, I will use it

cmd/erasure-server-pool.go
```diff
@@ -215,6 +199,22 @@ func newErasureServerPools(ctx context.Context, endpointServerPools EndpointServ
 		break
 	}
 
+	// initialize the object layer.
+	setObjectLayer(z)
```
harshavardhana: Please don't set the object layer right here.

cmd/erasure-server-pool.go
```diff
@@ -215,6 +199,22 @@ func newErasureServerPools(ctx context.Context, endpointServerPools EndpointServ
...
+	//
+	// - Disk auto healing
+	// - MRF (most recently failed) healing
+	bootstrapTrace("initAutoHeal", func() {
+		initAutoHeal(GlobalContext, z)
+	})
+
+	bootstrapTrace("initHealMRF", func() {
+		go globalMRFState.start()
+		go globalMRFState.healRoutine(z)
```
harshavardhana: can you rename start() to something more `startMRFPersistence()` or something?

cmd/signals.go
```diff
@@ -46,6 +46,9 @@ func handleSignals() {
 	}
 
 	stopProcess := func() bool {
+		globalMRFState.shutdown() // this can take time sometimes, it needs to be executed
```
harshavardhana: we must have a timeout here..

cmd/mrf.go
```diff
@@ -15,50 +15,189 @@
...
+	localDrives := cloneDrives(globalLocalDrives)
+	globalLocalDrivesMu.RUnlock()
+
+	for _, localDrive := range localDrives {
+		r := newReader()
+		err := localDrive.CreateFile(context.Background(), "", minioMetaBucket, pathJoin(healMRFDir, "list.bin"), -1, r)
+		r.Close()
+		if err == nil {
+			break
+		}
```
harshavardhana, line 151: It is perhaps okay to write 2 copies instead of just one IMO.

cmd/erasure-server-pool.go
vadmeste: Healing in general should be executed after the object layer is initialized, in this case many if not all heal MRF entries will be unsuccessfully retried to be healed

I can move this code outside this function anyway but I am not sure, this change does not seem to break anything

harshavardhana: Okay

harshavardhana:
Looks reordering caused some issue in healing tests PTAL @vadmeste

harshavardhana:
@vadmeste please rebase this PR with master.

bh4t:
@vadmeste PTAL when you get a chance.

harshavardhana (approved):

cmd/mrf.go
```diff
@@ -15,51 +15,190 @@
...
 	default:
 	}
 }
 
+// Do not accept new MRF operations anymore and start to save
+// the current heal status in one available disk
+func (m *mrfState) shutdown() {
+	atomic.StoreInt32(&m.closed, 1)
+	m.wg.Wait()
+	close(m.opCh)
```
poornas, line 105: would be safer to use xioutil.SafeClose to close m.opCh 

imo, it looks like closing here will lead to panic if others are still writing to opCh. The close should be on the writer's side

poornas (changes requested):

cmd/mrf.go
```diff
@@ -15,51 +15,192 @@
...
 
 // Add a partial S3 operation (put/delete) when one or more disks are offline.
-func (m *mrfState) addPartialOp(op partialOperation) {
+func (m *mrfState) addPartialOp(op PartialOperation) {
 	if m == nil {
 		return
 	}
 
 	select {
+	case <-m.closing:
```
krisis: This select statement doesn't guarantee that this method won't send `op` over `m.opCh`. This is because once `mrfState.shutdown()` is called, both `case <-m.closing` and `case m.opCh <-op` may be ready and select statement picks one among them at random.
> A select blocks until one of its cases can run, then it executes that case. It chooses one at random if multiple are ready. 
 ref: https://go.dev/tour/concurrency/5


Instead we should do like this,

```go
select {
case <-m.closing:
      m.closeOnce.Do(func() {close(m.opCh))}
default:
      select {
      case m.opCh <- op:
      default:
      }
}
```

cmd/mrf.go
```diff
@@ -15,51 +15,199 @@
...
 	default:
 	}
 }
 
+// Do not accept new MRF operations anymore and start to save
+// the current heal status in one available disk
+func (m *mrfState) shutdown() {
+	atomic.StoreInt32(&m.closing, 1)
+
+	m.wg.Wait()
```
krisis, line 104: The wait group approach to synchronize between the multiple senders, i.e `addPartialOp` and the `shutdown` method won't work since all calls to `wg.Add` must happen before we call `wg.Wait`. This isn't guaranteed.
> Note that calls with a positive delta that occur when the counter is zero must happen before a Wait.
ref: https://pkg.go.dev/sync#WaitGroup.Add

I understand that closing `m.opCh` simplifies processing of outstanding `PartialOp` values onto local drives. But closing a channel which has multiple senders is error prone and complicates the code. 

As an alternate approach, we could use a done/quit channel, which will be closed at the beginning of the `shutdown` method. `addPartialOp` calls will select on this channel and return immediately if they found it closed.
This way we stop more tasks from being sent on `m.opCh`.  Now, we could drain at most `len(m.opCh)` entries from  `m.opCh`. Of course this is only a point in time value of the number of outstanding `PartialOp` values and there could be more sent after we signalled on the quit channel (see above). The downside is that we may drop the 'last' few values. Overall, the implementation will be easier to reason with.

vadmeste: > The wait group approach to synchronize between the multiple senders, i.e addPartialOp and the shutdown method won't work since all calls to wg.Add must happen before we call wg.Wait. This isn't guaranteed.

I do not need a guarantee here. I just wanted to avoid sending new entries to m.opCh during shutdown because m.opCh will be closed by then

cmd/mrf.go
```diff
@@ -15,51 +15,199 @@
...
 // Add a partial S3 operation (put/delete) when one or more disks are offline.
-func (m *mrfState) addPartialOp(op partialOperation) {
+func (m *mrfState) addPartialOp(op PartialOperation) {
 	if m == nil {
 		return
 	}
 
+	m.wg.Add(1)
+	defer m.wg.Done()
+
```
poornas, line 89: can you check if it is shutdown prior to adding to wg as well? Otherwise m.wg.Wait() will need to wait unnecessarily on any ops being queued

krisis (approved):

cmd/mrf.go
harshavardhana: @vadmeste ^^

harshavardhana (approved):

cmd/mrf.go
harshavardhana: @vadmeste ^^

klauspost:
@harshavardhana @vadmeste Well... We don't have "shutdown" anymore. How does this work, then?

harshavardhana:
> @harshavardhana @vadmeste Well... We don't have "shutdown" anymore. How does this work, then?

stopProcess() invokes the necessary shutdown and save required, however it must not be 
blocking in any manner and it doesn't need to be "accurate" by any means.

harshavardhana:
https://github.com/minio/minio/actions/runs/9685968715/job/26727319781 some healing test is failing we must investigate.

vadmeste:
> stopProcess() invokes the necessary shutdown and save required, however it must not be
blocking in any manner and it doesn't need to be "accurate" by any means.

shall we give it one second timeout ? a force killing can generate a corrupted data as well but I am willing to relax that and try to read as much as possible

harshavardhana:
> shall we give it one second timeout ? a force killing can generate a corrupted data as well but I am willing to relax that and try to read as much as possible

? what is this ask about @vadmeste

harshavardhana:
> https://github.com/minio/minio/actions/runs/9685968715/job/26727319781 some healing test is failing we must investigate.

was this investigated?

harshavardhana (approved):

PR #19452, opened 2024-04-09
Add drive metrics in metrics-v3

anjalshireesh:

## Description

Add following metrics:

- used_inodes
- total_inodes
- healing
- online
- reads_per_sec
- reads_kb_per_sec
- reads_await
- writes_per_sec
- writes_kb_per_sec
- writes_await
- perc_util

In order to be able to calculate the `per_sec` values, we maintain the previously captured IOStats related data (along with the time at which they were captured), and compare them against the current values subsequently. This happens every minute along with the cache refresh.

## Motivation and Context

metrics-v3 for drives

ref: https://github.com/miniohq/engineering/issues/1510

## How to test this PR?

`mc admin prometheus metrics <alias> node`

should return the new drive related metrics
(will need some changes in mc to invoke the metrics-v3 api)

## Types of changes

## Checklist:

### DIFF
--- a/cmd/metrics-v3-cache.go
+++ b/cmd/metrics-v3-cache.go
@@ -18,15 +18,20 @@
 package cmd
 
 import (
+	"sync"
 	"time"
 
 	"github.com/minio/madmin-go/v3"
 	"github.com/minio/minio/internal/cachevalue"
 )
 
-// initialDriveIOStats is used to calculate "per second"
-// values for IOStat related disk metrics e.g. reads/sec.
-var initialDriveIOStats *diskIOStats
+var (
+	// prevDriveIOStats is used to calculate "per second"
+	// values for IOStat related disk metrics e.g. reads/sec.
+	prevDriveIOStats            map[string]madmin.DiskIOStats
+	prevDriveIOStatsMu          sync.RWMutex
+	prevDriveIOStatsRefreshedAt time.Time
+)
 
 // metricsCache - cache for metrics.
 //
@@ -66,8 +71,8 @@ func newNodesUpDownCache() *cachevalue.Cache[nodesOnline] {
 }
 
 type diskIOStats struct {
-	stats       map[string]madmin.DiskIOStats
-	collectedAt time.Time
+	stats    map[string]madmin.DiskIOStats
+	duration time.Duration
 }
 
 // storageMetrics - cached storage metrics.
@@ -132,30 +137,35 @@ func newDriveMetricsCache() *cachevalue.Cache[storageMetrics] {
 		onlineDrives, offlineDrives := getOnlineOfflineDisksStats(storageInfo.Disks)
 		totalDrives := onlineDrives.Merge(offlineDrives)
 
-		var ioStats *diskIOStats
-		currentStats := getCurrentDiskMetrics()
-		if initialDriveIOStats == nil {
-			initialDriveIOStats = &diskIOStats{
-				stats:       currentStats,
-				collectedAt: time.Now().UTC(),
-			}
-		} else {
-			diffStats := map[string]madmin.DiskIOStats{}
-			for d, cs := range currentStats {
-				if is, found := initialDriveIOStats.stats[d]; found {
-					diffStats[d] = getDiffStats(is, cs)
+		currentStats := getCurrentDriveIOStats()
+		now := time.Now().UTC()
+
+		prevDriveIOStatsMu.Lock()
+		if prevDriveIOStats != nil {
+			duration := now.Sub(prevDriveIOStatsRefreshedAt)
+			if duration.Seconds() > 1 {
+				diffStats := map[string]madmin.DiskIOStats{}
+				for d, cs := range currentStats {
+					if ps, found := prevDriveIOStats[d]; found {
+						diffStats[d] = getDiffStats(ps, cs)
+					}
+				}
+
+				ioStats := &diskIOStats{stats: diffStats, duration: duration}
+
+				v = storageMetrics{
+					storageInfo:   storageInfo,
+					onlineDrives:  onlineDrives.Sum(),
+					offlineDrives: offlineDrives.Sum(),
+					totalDrives:   totalDrives.Sum(),
+					ioStats:       ioStats,
 				}
 			}
-			ioStats = &diskIOStats{stats: diffStats, collectedAt: time.Now().UTC()}
 		}
 
-		v = storageMetrics{
-			storageInfo:   storageInfo,
-			onlineDrives:  onlineDrives.Sum(),
-			offlineDrives: offlineDrives.Sum(),
-			totalDrives:   totalDrives.Sum(),
-			ioStats:       ioStats,
-		}
+		prevDriveIOStats = currentStats
+		prevDriveIOStatsRefreshedAt = now
+		prevDriveIOStatsMu.Unlock()
 
 		return
 	}
--- a/cmd/metrics-v3-cache.go
+++ b/cmd/metrics-v3-cache.go
@@ -25,14 +25,6 @@ import (
 	"github.com/minio/minio/internal/cachevalue"
 )
 
-var (
-	// prevDriveIOStats is used to calculate "per second"
-	// values for IOStat related disk metrics e.g. reads/sec.
-	prevDriveIOStats            map[string]madmin.DiskIOStats
-	prevDriveIOStatsMu          sync.RWMutex
-	prevDriveIOStatsRefreshedAt time.Time
-)
-
 // metricsCache - cache for metrics.
 //
 // When serving metrics, this cache is passed to the MetricsLoaderFn.
@@ -127,6 +119,14 @@ func getDiffStats(initialStats, currentStats madmin.DiskIOStats) madmin.DiskIOSt
 }
 
 func newDriveMetricsCache() *cachevalue.Cache[storageMetrics] {
+	var (
+		// prevDriveIOStats is used to calculate "per second"
+		// values for IOStat related disk metrics e.g. reads/sec.
+		prevDriveIOStats            map[string]madmin.DiskIOStats
+		prevDriveIOStatsMu          sync.RWMutex
+		prevDriveIOStatsRefreshedAt time.Time
+	)
+
 	loadDriveMetrics := func() (v storageMetrics, err error) {
 		objLayer := newObjectLayerFn()
 		if objLayer == nil {
--- a/cmd/metrics-v3-cache.go
+++ b/cmd/metrics-v3-cache.go
@@ -24,6 +24,10 @@ import (
 	"github.com/minio/minio/internal/cachevalue"
 )
 
+// initialDriveIOStats is used to calculate "per second"
+// values for IOStat related disk metrics e.g. reads/sec.
+var initialDriveIOStats *diskIOStats
+
 // metricsCache - cache for metrics.
 //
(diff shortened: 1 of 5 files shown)


cmd/metrics-v3-cache.go
```diff
@@ -24,6 +24,10 @@ import (
 	"github.com/minio/minio/internal/cachevalue"
 )
 
+// initialDriveIOStats is used to calculate "per second"
+// values for IOStat related disk metrics e.g. reads/sec.
+var initialDriveIOStats *diskIOStats
```
klauspost: It seems like this is set on first request and then never updated again, so this will give the lifetime average of these values? Correct?

If we are just interested in the lifetime you should just initialize it right away and not risk a race.

IMPORTANT Question is if it is lifetime average we are interested in - or a rolling average for a specific time window?

Lifetime becomes increasingly less useful the longer the server is running. And if a server has been running for 2 months, it will take days for an issue to even become visible here.

A rolling sum over last minute would make more sense to me.

cmd/metrics-v3-cache.go
```diff
@@ -108,14 +131,35 @@ func newDriveMetricsCache() *cachevalue.Cache[storageMetrics] {
...
+				collectedAt: time.Now().UTC(),
+			}
+		} else {
+			diffStats := map[string]madmin.DiskIOStats{}
+			for d, cs := range currentStats {
+				if is, found := initialDriveIOStats.stats[d]; found {
+					diffStats[d] = getDiffStats(is, cs)
+				}
+			}
+			ioStats = &diskIOStats{stats: diffStats, collectedAt: time.Now().UTC()}
```
klauspost: It seems like you are missing the duration of the collected data, otherwise the metrics cannot be calculated as a rate over time.

Edit: It seems like you are hacking that in below: `durationSecs := ioStats.collectedAt.Sub(initialDriveIOStats.collectedAt).Seconds()` - do it here.

cmd/metrics-v3-system-drive.go
```diff
@@ -68,15 +90,147 @@ var (
...
+	})
+
+	stats := map[string]madmin.DiskIOStats{}
+	for d, m := range driveRealtimeMetrics.ByDisk {
+		stats[d] = m.IOStats
+	}
+	return stats
+}
+
+func setDriveBasicMetrics(m MetricValues, drive madmin.Disk, labels []string) {
```
klauspost: Pass a pointer, since you are writing to m.
```suggestion
func setDriveBasicMetrics(m *MetricValues, drive madmin.Disk, labels []string) {
```

.. actually you should do...

```suggestion
func (m *MetricValues) setDriveBasicMetrics(drive madmin.Disk, labels []string) {
```

cmd/metrics-v3-system-drive.go
```diff
@@ -68,15 +90,147 @@ var (
...
+	}
+	m.Set(driveHealing, healing, labels...)
+
+	if drive.State == "ok" {
+		online = 1
+	}
+	m.Set(driveOnline, online, labels...)
+}
+
+func setDriveAPIMetrics(m MetricValues, disk madmin.Disk, labels []string) {
```
klauspost:
```suggestion
func (m *MetricValues) setDriveAPIMetrics(disk madmin.Disk, labels []string) {
```

cmd/metrics-v3-cache.go
Praveenrajmani: Is it possible to re-use the existing functions and types? @anjalshireesh

cmd/metrics-v3-system-drive.go
```diff
@@ -68,15 +90,147 @@ var (
...
+	labels = append(labels, "api", "")
+	lastIdx := len(labels) - 1
+	for apiName, latency := range disk.Metrics.LastMinute {
+		labels[lastIdx] = "storage." + apiName
+		m.Set(driveAPILatencyMicros, float64(latency.Avg().Microseconds()),
+			labels...)
+	}
+}
+
+func setDriveIOStatMetrics(m MetricValues, ioStats *diskIOStats, disk madmin.Disk, labels []string) {
```
klauspost:
```suggestion
func (m *MetricValues) setDriveIOStatMetrics(ioStats *diskIOStats, disk madmin.Disk, labels []string) {
```

cmd/metrics-v3-system-drive.go
```diff
@@ -68,15 +90,147 @@ var (
...
+	if ioStats == nil {
+		return
+	}
+
+	durationSecs := ioStats.collectedAt.Sub(initialDriveIOStats.collectedAt).Seconds()
```
klauspost: Do this while calculating.

cmd/metrics-v3-system-drive.go
```diff
@@ -68,15 +90,147 @@ var (
...
+
+	if durationSecs <= 0 {
```
klauspost:
```suggestion
	if durationSecs < 1 {
```

Values less than 1 second will fluctuate too much.

cmd/metrics-v3-cache.go
```diff
@@ -61,8 +65,15 @@ func newNodesUpDownCache() *cachevalue.Cache[nodesOnline] {
 		loadNodesUpDown)
 }
 
+type diskIOStats struct {
+	stats       map[string]madmin.DiskIOStats
+	collectedAt time.Time
```
Praveenrajmani: this can be a global var? like we have `lastDriveStatsRefresh` currently?

klauspost (commented):
I don't think lifetime values are very useful, so we should probably reconsider that aspect.

cmd/metrics-v3-cache.go
anjalshireesh: @Praveenrajmani `metrics-resource.go` implements its own caching mechanism, whereas `metrics-v3` has introduced a standard caching mechanism across all metrics. Because of this most of the existing functions cannot be reused as is. Having said that, please do point out if you find any specific missed reuse opportunities.

cmd/metrics-v3-cache.go
anjalshireesh: @Praveenrajmani I've introduced similar global vars now. This struct still stays as it's purpose is different (caching the latest drive io stats).

cmd/metrics-v3-cache.go
anjalshireesh: Thanks @klauspost , have updated to use "last minute" data instead of lifetime values.

Edit: Actually what I've implemented at this point (using the cache with 1 minute ttl) will behave like `1 minute` rolling window only if prometheus scrape interval is less than a minute. If the interval is bigger, say 2 minutes, then it will become a 2 minutes window. In other words, every scrape will return `per second` values for the interval between the two scrapes. Is that reasonable? If not, I'll have to run a timer to collect the data every minute (like we are doing in metrics-v2 `metrics-resource.go`).

cc @donatello @Praveenrajmani

cmd/metrics-v3-cache.go
anjalshireesh: done, capturing `duration` instead of `collectedAt` now.

cmd/metrics-v3-system-drive.go
anjalshireesh: thanks, done

cmd/metrics-v3-system-drive.go
anjalshireesh: done

cmd/metrics-v3-system-drive.go
anjalshireesh: done

cmd/metrics-v3-system-drive.go
anjalshireesh: done

cmd/metrics-v3-system-drive.go
anjalshireesh: done

cmd/metrics-v3-cache.go
```diff
@@ -18,12 +18,21 @@
...
 )
 
+var (
+	// prevDriveIOStats is used to calculate "per second"
+	// values for IOStat related disk metrics e.g. reads/sec.
+	prevDriveIOStats            map[string]madmin.DiskIOStats
+	prevDriveIOStatsMu          sync.RWMutex
+	prevDriveIOStatsRefreshedAt time.Time
+)
+
```
donatello: I suggest putting these in line 130 (beginning of `newDriveMetricsCache`) as it is only meant for the `loadDriveMetrics` closure there - mainly to reduce bringing in new globals.

donatello (approved):
Minor comment, otherwise LGTM.

cmd/metrics-v3-cache.go
anjalshireesh: thanks, done

cmd/metrics-v3-cache.go
```diff
@@ -108,14 +136,40 @@ func newDriveMetricsCache() *cachevalue.Cache[storageMetrics] {
...
+					}
+				}
+
+				ioStats := &diskIOStats{stats: diffStats, duration: duration}
+
+				v = storageMetrics{
+					storageInfo:   storageInfo,
+					onlineDrives:  onlineDrives.Sum(),
+					offlineDrives: offlineDrives.Sum(),
+					totalDrives:   totalDrives.Sum(),
```
Praveenrajmani: These values will be set only if `prevDriveIOStats != nil`. Can we move this out of the `if` block?

cmd/metrics-v3-cache.go
```diff
@@ -61,8 +62,15 @@ func newNodesUpDownCache() *cachevalue.Cache[nodesOnline] {
...
 
+type diskIOStats struct {
+	stats    map[string]madmin.DiskIOStats
+	duration time.Duration
+}
+
+// storageMetrics - cached storage metrics.
 type storageMetrics struct {
 	storageInfo                              madmin.StorageInfo
+	ioStats                                  *diskIOStats
```
Praveenrajmani: The current implementation is also fine, Just a suggestion here in case if you think it is better.

Can we fix the types to be like the following

```go
type perDriveIOStats struct {
     driveReadsPerSec  float64
     driveReadsKBPerSec float64
     driveReadsAwait float64
     driveWritesPerSec float64
     driveWritesKBPerSec float64
     driveWritesAwait float64
     drivePercUtil float64
}
```

```go
type storageMetrics struct {
     ...
     ioStats      map[string]perDriveIOStats
     ...
}
```
and keep all the calculations here in `loadDriveMetrics()` and just set the calculated values in `setDriveIOStatMetrics()`

Praveenrajmani: There aren't any benefits in the above approach, just that the calculations will be on one side and setMetrics() will just set the metrics and will not do any calculations. This won't spill the calculation logic to more than one place.

Praveenrajmani (approved):
LGTM, Just a small nit.

cmd/metrics-v3-cache.go
anjalshireesh: thanks, done.

cmd/metrics-v3-cache.go
anjalshireesh: @Praveenrajmani I liked this suggestion. Have implemented, just with slightly different names for the struct and its fields. PTAL.

klauspost (approved):
ok, so there is up to 1 minute latency on metrics. But that is fine - we can always move to a sliding window if that makes sense later.

donatello (approved):

Issue #18825, opened 2024-01-18
Support servers running behind loadbalancer (different internal vs. external IPs)


fpytloun:
**Is your feature request related to a problem? Please describe.**
I would like to deploy MiniO cluster where servers are in multiple racks on different internal networks and their ports exposed via loadbalancer. In such case, provided hostnames are not resolving to local IP (eg. IP of k8s pod).
Also when someone attempts this setup, instead of meaningful error, starting minio server just keeps failing.

```
API: SYSTEM()
Time: 15:39:05 UTC 01/18/2024
Error: Read failed. Insufficient number of drives online (*errors.errorString)
       8: internal/logger/logger.go:258:logger.LogIf()
       7: cmd/prepare-storage.go:254:cmd.connectLoadInitFormats()
       6: cmd/prepare-storage.go:312:cmd.waitForFormatErasure()
       5: cmd/erasure-server-pool.go:104:cmd.newErasureServerPools()
       4: cmd/server-main.go:976:cmd.newObjectLayer()
       3: cmd/server-main.go:718:cmd.serverMain.func9()
       2: cmd/server-main.go:434:cmd.bootstrapTrace()
       1: cmd/server-main.go:716:cmd.serverMain()
Waiting for a minimum of 4 drives to come online (elapsed 26s)
```

It is similar to this question: https://github.com/minio/minio/discussions/17374
But not possible to solve with host network.

**Describe the solution you'd like**
Multiple solutions might apply but simplest would be to just have parameter like `MINIO_SERVERNAME` to define current server endpoint so it can find itself within server pool list.

**Describe alternatives you've considered**
Messing with /etc/hosts which will break SSL verification.

I did not try to set with only current server in /etc/hosts pointing to localhost, eg.:
```
127.0.0.1    minio-0.rack-1.domain.tld
```

manju-rn:
@fpytloun   What is your cluster - in K8s? If yes, then you could just define a ingress and point to respective service names

fpytloun:
> @fpytloun What is your cluster - in K8s? If yes, then you could just define a ingress and point to respective service names

My setup are two regions running separate k8s clusters. So minios within one region can use k8s svc but they need to use these LB endpoints to connect to minios in other region.

allanrogerr:
Please provide the following. Otherwise I will use minio latest to review this.
## Your Environment

* Version used (`minio --version`):
* Server setup and configuration:
* Operating System and version (`uname -a`):

[...]

PR #19453, opened 2024-04-09
Allow specifying the local server with env variable _MINIO_SERVER_LOCAL

allanrogerr:

## Description
For containerized systems where the hostname cannot be resolved to local IP, this PR introduces a change allowing the specifying of the local server with env variable `_MINIO_SERVER_LOCAL`

## Motivation and Context
In this sort of containerized system, e.g. a kubernetes cluster within which minio pods is running on nodes on different networks, the internal IP address of the pod is not resolvable from any hostname, accessible externally. Therefore minio is not able to recognize itself within the pool of hosts as no endpoint can be resolved to it's own IP address.

A workaround would be to perform the resolution manually by editing `/etc/hosts/` and assigning a consecutive host resolving to the pod ip address. e.g.
`<ip of k8s pod>     <hostname>.<domain>`
e.g.
`10.42.0.11     minio-18825-1.lab.min.dev`

## How to test this PR?
See example here: https://github.com/allanrogerr/public/wiki/minio%E2%80%9018825
1.- Pull and compile http://github.com/allanrogerr/minio.git on several k8s pods
2.- Run minio on all pods e.g.
```
CI=on \
_MINIO_SERVER_LOCAL=minio-18825-1.lab.min.dev:9000 \
~/github/minio/minio server https://minio-18825-{0...1}.lab.min.dev:9000/tmp/data{1...2} --address :9000 --console-address :9090 --certs-dir ~/.minio/certs
```
3.- Observe the minio cluster online

## Types of changes

## Checklist:

### DIFF
--- a/cmd/endpoint.go
+++ b/cmd/endpoint.go
@@ -807,7 +807,7 @@ func (p PoolEndpointList) UpdateIsLocal() error {
 						continue
 					}
 
-					if endpoint.Host == "" {
+					if endpoint.Host == "" || (env.Get("_MINIO_SERVER_LOCAL", "") == endpoint.Host) {
 						if !foundLocal {
 							foundLocal = true
 						}
--- a/cmd/endpoint.go
+++ b/cmd/endpoint.go
@@ -807,7 +807,7 @@ func (p PoolEndpointList) UpdateIsLocal() error {
 						continue
 					}
 
-					if endpoint.Host == "" || (env.Get("_MINIO_SERVER_LOCAL", "") == endpoint.Host) {
+					if endpoint.Host == "" || (orchestrated && env.Get("_MINIO_SERVER_LOCAL", "") == endpoint.Host) {
 						if !foundLocal {
 							foundLocal = true
 						}


allanrogerr:
Fixes https://github.com/minio/minio/issues/18825

zveinn (approved):

klauspost:
This seems like a can of worms. AFAICT you can really f up an install with this.

> For systems where the hostname cannot be resolved to local IP

It sounds to me like this is the actual issue. It sounds like a nasty workaround to have to push individual (unmatching) env vars to a deployment.  I am by no means an expert on this, but it is a red flag for me.

zveinn:
> This seems like a can of worms. AFAICT you can really f up an install with this.
> 
> > For systems where the hostname cannot be resolved to local IP
> 
> It sounds to me like this is the actual issue. It sounds like a nasty workaround to have to push individual (unmatching) env vars to a deployment. I am by no means an expert on this, but it is a red flag for me.

He's just letting the local be defined right ? .. I mean, sure.. if you're running re-deployments and hostnames change, you're install has a good chance of getting messed up. But I think that applies to many other things too ? 

But I will agree with Klaus on one thing, the original issue seems to be that the local hostname is not resolvable. even within kubernetes the pod/container hostnames should be resolvable ? at least that how I remember it. 

Maybe we're missing more context ?

klauspost:
@zveinn Yeah. Without context it seems like many other things could fail - and per sever env vars just seems like a recipe for misconfiguration.

allanrogerr:
The internal k8s network has its own way of resolving its own hostnames. Within that internal network a pod can know that hostname - so host 10-42-0-10.minio-ns.pod.cluster.local is resolvable locally, but cannot be resolved externally. If we use the the external hostname e.g. node1.min.dev, the dns resolution works for each pod, but each pod cannot determine which of these external hostnames is itself.

I will make this code explicitly work only for containerised solutions. This should minimise any mixups.
Note that the only other workaround is modifying /etc/hosts or perhaps /etc/resolv.conf

harshavardhana:
This issue is about being able to run a cluster where the local for an endpoint cannot be determined by regular means where a DNS resolves to an IP which is not local to you but more elastic in nature.

This applies to many situations however this environment value is just a hack to circumvent that requirement that's why it's an experimental flag.

It's useful only for situations where above is trust and none of the normal deployments would need it.

harshavardhana (approved):

harshavardhana:
And on top of that it's even more self constrained to be honored only for k8s/docker environments which have this problem.

- docker swarm + docker stack deployment face this issue
- k8s multi-cluster deployments

PR #19481, opened 2024-04-11
ilm: Handle DeleteAllVersions action differently for DEL markers

krisis:

## Description
This PR brings about two changes, one adds a new lifecycle action, `DelMarkerExpiration` and another modifies the behavior of an existing lifecycle configuration element, `ExpiredObjectDeleteAllVersions`.

## Companion PRs
- https://github.com/minio/minio-go/pull/1959
- https://github.com/minio/mc/pull/4913
### Addition of `DelMarkerExpiration`
This ILM action removes all versions of an object if its latest version is a DEL marker.
```xml
    <DelMarkerObjectExpiration>
        <Days> 10 </Days>
    </DelMarkerObjectExpiration>
```
1. Applies only to objects whose,

    - latest version is a DEL marker.
    - satisfies the number of days criteria

2. Deletes all versions of this object

3. Associated rule can't have tag-based filtering
 
### Modification to `ExpiredObjectDeleteAllVersions`
Will apply only to objects whose latest version is **not** a DEL marker.

## Motivation and Context

This is a breaking change to how ExpiredObejctDeleteAllVersions functions today. This is necessary to avoid the following highly probable footgun scenario in the future.

Scenario:
User uses tags based filtering to select the time to live(TTL) for an object. Application sometimes deletes objects too, making their latest version a DEL marker. Previous implementation skipped tag based filters if latest version was DEL marker, voiding the tag based TTL. User is surprised to find objects expired sooner than expected.

## Types of changes

## Checklist:

### DIFF
--- a/internal/bucket/lifecycle/delmarker-expiration_test.go
+++ b/internal/bucket/lifecycle/delmarker-expiration_test.go
@@ -1,0 +1,46 @@
+package lifecycle
+
+import (
+	"encoding/xml"
+	"fmt"
+	"testing"
+)
+
+func TestDelMarkerExpParseAndValidate(t *testing.T) {
+	tests := []struct {
+		xml string
+		err error
+	}{
+		{
+			xml: `<DelMarkerExpiration> <Days> 1 </Days> </DelMarkerExpiration>`,
+			err: nil,
+		},
+		{
+			xml: `<DelMarkerExpiration> <Days> -1 </Days> </DelMarkerExpiration>`,
+			err: errInvalidDaysDelMarkerExpiration,
+		},
+	}
+
+	for i, test := range tests {
+		t.Run(fmt.Sprintf("TestDelMarker-%d", i), func(t *testing.T) {
+			var dexp DelMarkerExpiration
+			var fail bool
+			err := xml.Unmarshal([]byte(test.xml), &dexp)
+			if test.err == nil {
+				if err != nil {
+					fail = true
+				}
+			} else {
+				if err == nil {
+					fail = true
+				}
+				if test.err.Error() != err.Error() {
+					fail = true
+				}
+			}
+			if fail {
+				t.Fatalf("Expected %v but got %v", test.err, err)
+			}
+		})
+	}
+}
--- a/internal/bucket/lifecycle/delmarker-expiration_test.go
+++ b/internal/bucket/lifecycle/delmarker-expiration_test.go
@@ -1,3 +1,20 @@
+// Copyright (c) 2024 MinIO, Inc.
+//
+// This file is part of MinIO Object Storage stack
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
 package lifecycle
 
 import (
--- a/internal/bucket/lifecycle/lifecycle.go
+++ b/internal/bucket/lifecycle/lifecycle.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2015-2021 MinIO, Inc.
+// Copyright (c) 2015-2024 MinIO, Inc.
 //
 // This file is part of MinIO Object Storage stack
 //
@@ -279,7 +279,7 @@ func (lc Lifecycle) FilterRules(obj ObjectOpts) []Rule {
 		if !strings.HasPrefix(obj.Name, rule.GetPrefix()) {
 			continue
 		}
-		if !obj.DeleteMarker && !rule.Filter.TestTags(obj.UserTags) {
+		if !rule.Filter.TestTags(obj.UserTags) {
 			continue
 		}
 		if !obj.DeleteMarker && !rule.Filter.BySize(obj.Size) {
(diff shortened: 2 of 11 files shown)


internal/bucket/lifecycle/delmarker-expiration_test.go
```diff
@@ -0,0 +1,46 @@
+package lifecycle
```
harshavardhana, line 18: missing copyright.

harshavardhana (commented):
Also don't forget to update `docs/bucket/lifecycle`

internal/bucket/lifecycle/rule.go
```diff
@@ -158,7 +160,10 @@ func (r Rule) Validate() error {
 	if err := r.validateNoncurrentTransition(); err != nil {
 		return err
 	}
-	if !r.Expiration.set && !r.Transition.set && !r.NoncurrentVersionExpiration.set && !r.NoncurrentVersionTransition.set {
+	if (!r.Filter.Tag.IsEmpty() || len(r.Filter.And.Tags) != 0) && !r.DelMarkerExpiration.Empty() {
+		return errInvalidRuleDelMarkerExpiration
+	}
+	if !r.Expiration.set && !r.Transition.set && !r.NoncurrentVersionExpiration.set && !r.NoncurrentVersionTransition.set && r.DelMarkerExpiration.Empty() {
```
vadmeste, line 166: I think we can leave the old change of this line as it is and add a new 'if' with r.DeMarkerExpiration involved

internal/bucket/lifecycle/lifecycle.go
```diff
@@ -401,6 +385,17 @@ func (lc Lifecycle) eval(obj ObjectOpts, now time.Time) Event {
 			}
 		}
 
+		// DelMarkerExpiration
+		if obj.IsLatest && obj.DeleteMarker && !rule.DelMarkerExpiration.Empty() {
+			if due, ok := rule.DelMarkerExpiration.NextDue(obj); ok && (now.IsZero() || now.After(due)) {
+				return Event{
```
vadmeste: Returning here can confuse the ILM prediction header, it is when now.IsZero() is true, we can have transition + expiration and expiration due time is way after the transition due time. I think the spirit of this code is that we evaluate all rules and then we order them later.

vadmeste (commented):
Few comments

internal/bucket/lifecycle/lifecycle.go
krisis: Ah yes, good catch. I overlooked this.

internal/bucket/lifecycle/rule.go
krisis: I'm not sure what you have in mind. Could you explain with a modified `if`. In this if condition I'm trying to disallow ILM rules which don't have at least one of the above actions in it.

internal/bucket/lifecycle/lifecycle.go
krisis: @vadmeste I have add a couple of unit tests to cover this.

harshavardhana (approved):

poornas:
This feature may interact poorly with replication if replication is set up as active -> passive (read only DR). Depending on the duration of ilm expiry rule above - there is a chance that replication of non current versions is not complete.  Either the docs needs to warn about this possibility or it needs to be addressed by verifying all underlying versions are in COMPLETED status.

poornas:
Another q - Since this op is doing a prefix delete, there may be a window between queuing and acting upon this expiry task where a new version can be uploaded for the same object. Seems like expiry task is not actually verifying that current Objectinfo is indeed the latest before proceeding

Also, seems like if one or more versions in the stack is in tiered status, clean up will not be done

klauspost:
> Associated rule can't have tag-based filtering

Why is this limitation there?

krisis:
> > Associated rule can't have tag-based filtering
> 
> Why is this limitation there?

DelMarkerExpiration applies exclusively to DEL markers. Tag-based filtering doesn't apply to DEL markers, since they can't have tags. Similar restrictions are present in S3's lifecycle elements too. e.g,
>When you use the ExpiredObjectDeleteMarker S3 Lifecycle action, the rule cannot specify a tag-based filter.

(ref: https://docs.aws.amazon.com/AmazonS3/latest/userguide/lifecycle-configuration-examples.html; just above Example 8)

klauspost:
> DelMarkerExpiration applies exclusively to DEL markers. 

👍🏼 It was also my impression that it doesn't make sense. So it isn't a limitation as such - just a natural consequence of delete markers not having tags.

harshavardhana:
> Another q - Since this op is doing a prefix delete, there may be a window between queuing and acting upon this expiry task where a new version can be uploaded for the same object. Seems like expiry task is not actually verifying that current Objectinfo is indeed the latest before proceeding

latest objectInfo() must be verified first and it is already. `opts.DeletePrefix` when set and expiration is "expire=true"

```
        if opts.DeletePrefix {
                if opts.Expiration.Expire {
                        // Expire all versions expiration must still verify the state() on disk
                        // via a getObjectInfo() call as follows, any read quorum issues we
                        // must not proceed further for safety reasons. attempt a MRF heal
                        // while we see such quorum errors.
                        goi, _, gerr := er.getObjectInfoAndQuorum(ctx, bucket, object, opts)
                        if gerr != nil && goi.Name == "" {
                                if _, ok := gerr.(InsufficientReadQuorum); ok {
                                        // Add an MRF heal for next time.
                                        er.addPartial(bucket, object, opts.VersionID)

                                        return objInfo, InsufficientWriteQuorum{}
                                }
                                return objInfo, gerr
                        }
```

poornas:
> > Another q - Since this op is doing a prefix delete, there may be a window between queuing and acting upon this expiry task where a new version can be uploaded for the same object. Seems like expiry task is not actually verifying that current Objectinfo is indeed the latest before proceeding
> 
> latest objectInfo() must be verified first and it is already. `opts.DeletePrefix` when set and expiration is "expire=true"
> 
> ```
>         if opts.DeletePrefix {
>                 if opts.Expiration.Expire {
>                         // Expire all versions expiration must still verify the state() on disk
>                         // via a getObjectInfo() call as follows, any read quorum issues we
>                         // must not proceed further for safety reasons. attempt a MRF heal
>                         // while we see such quorum errors.
>                         goi, _, gerr := er.getObjectInfoAndQuorum(ctx, bucket, object, opts)
>                         if gerr != nil && goi.Name == "" {
>                                 if _, ok := gerr.(InsufficientReadQuorum); ok {
>                                         // Add an MRF heal for next time.
>                                         er.addPartial(bucket, object, opts.VersionID)
> 
>                                         return objInfo, InsufficientWriteQuorum{}
>                                 }
>                                 return objInfo, gerr
>                         }
> ```

while this checks the version id that qualified for ilm exists, it doesn't verify it is the latest version

harshavardhana:
> while this checks the version id that qualified for ilm exists, it doesn't verify it is the latest version

that we should add @krisis

krisis:
> while this checks the version id that qualified for ilm exists, it doesn't verify it is the latest version

Good catch! I have addressed this.

poornas (approved):

PR #19000, opened 2024-02-07
Fix typos

szepeviktor:
## Community Contribution License
All community contributions in this pull request are licensed to the project maintainers
under the terms of the [Apache 2 license](https://www.apache.org/licenses/LICENSE-2.0). 
By creating this pull request I represent that I have the right to license the 
contributions to the project maintainers under the Apache 2 license.

## Description

Found few misspellings.
Please consider switching to `typos`.

Please help me finish this PR: I do not speak go, and I am not a minio user.

PR #19088, opened 2024-02-20
use typos instead of codespell

harshavardhana:
## Community Contribution License
All community contributions in this pull request are licensed to the project maintainers
under the terms of the [Apache 2 license](https://www.apache.org/licenses/LICENSE-2.0). 
By creating this pull request I represent that I have the right to license the 
contributions to the project maintainers under the Apache 2 license.

## Description
use typos instead of codespell

## Motivation and Context
just cleanup

## How to test this PR?
install `typos` tool

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Optimization (provides speedup with no functional changes)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [ ] Fixes a regression (If yes, please add `commit-id` or `PR #` here)
- [x] Unit tests added/updated
- [x] Internal documentation updated
- [ ] Create a documentation update request [here](https://github.com/minio/docs/issues/new?label=doc-change,title=Doc+Updated+Needed+For+PR+github.com%2fminio%2fminio%2fpull%2fNNNNN)

PR #19488, opened 2024-04-12
Fix another round of errors

szepeviktor:
@harshavardhana Thank you for fixing previous typos internally. #19000

Here is another batch of fixes.


szepeviktor:
Minio is already "infected"  #19088 🥳

PR #19492, opened 2024-04-12
Keep an up-to-date copy of the KMS master key

allanrogerr:
- Keep an up-to-date copy of the KMS master key
- Validate KMS master key using a test of encrypting decryption, in order to startup MinIO

## Description
MinIO should automatically request the default master key once every 10 seconds. This matches the KES period for verifying if the KMS is offline. The intention is to have the master key cached, so that the IAM and other cluster metadata can be decrypted even if the KMs goes offline

Currently when the KMS goes offline, if MinIO is restarted for some reason it immediately tests its ability to create a key through KES on the KMS. Since KES cannot contact the KMS, MinIO is unable to start up and responds with:
`ERROR Unable to initialize a connection to KES as specified by the shell environment: failed to create key`

When the KMS is down, KES does not garbage collect any cache, unless `.cache.expiry.offline` is setup. Therefore in order for MinIO to start up, a way was needed to validate that the cached KMS key can be used to encrypt and decrypt.

## Motivation and Context
This is the precise scenario:
* MinIO starts and decrypts its IAM data using the key MINIO_KMS_KES_KEY_NAME. Therefore, KES has to fetch the key from the KMS backend (if not cached already). 

* After some time KES evicts the key MINIO_KMS_KES_KEY_NAME from the cache since its not used by MinIO. See usage of `.cache.expiry.unused`. 
`.cache.expiry.any` is out of scope since it always purges all cache if KES verifies the KMS to be online. This is a race condition.
`.cache.expiry.offline` is also out of scope since it always purges all cache if KES verifies the KMS is offline. This is a race condition.

* The KMS becomes temp. unavailable (network issue, maintance etc.) 

* MinIO gets restarted. This now fails because KES cannot find the MINIO_KMS_KES_KEY_NAME in its cache and it also cannot fetch it from the KMS backend because the KMS is not available.

This fix allows MinIO to start up successfully, with most of its functionality enabled except creating a new KMS key, nor encryption a bucket with cached KMS keys. Only data encrypted with the default key is accessible.

Currently KES listKeys does not use its cache when it fails to access the KMS (https://github.com/minio/kes/blob/master/keystore.go#L342). Therefore, this fix is limited in its scope and while it does permit disabling encryption on future upload, it does not allow a user to enable encryption on buckets while the KMS is down.

## How to test this PR?
See https://github.com/allanrogerr/public/wiki/mineos%E2%80%907
Note: Increase `any` and `offline` to large values to reduce test scope

#### Test 0
```
cache:
  expiry:
    any: 600s
    unused: 30s
    offline: 600s
```
* Start vault, verify KES running, start MinIO, create and encrypt bucket with default key, upload data, kill vault, restart minio after 30s
* Observe minio starts up
* Validate existing minio encrypted data is accessible
* Validate new minio encrypted data is uploadable and accessible
* Validate encryption can be removed and new minio data is uploadable and accessible

#### Test 1
```
cache:
  expiry:
    any: 600s
    unused: 20s
    offline: 600s
```
* Start vault, verify KES running, start MinIO, create and encrypt bucket, upload data, kill vault, restart minio after 10s
* Observe minio starts up sometimes due to the inherent race condition
* Validate existing minio encrypted data is accessible
* Validate new minio encrypted data is uploadable and accessible
* Validate encryption can be removed and new minio data is uploadable and accessible

#### Test 2
```
cache:
  expiry:
    any: 600s
    unused: 10s
    offline: 600s
```
* Start vault, verify KES running, start MinIO, create and encrypt bucket, upload data, kill vault, restart minio after 10s
* Observe minio starts up sometimes due to the inherent race condition

#### Test 3
```
cache:
  expiry:
    any: 600s
    unused: 5s
    offline: 600s
```
* Start vault, verify KES running, start MinIO, create and encrypt bucket, upload data, kill vault, restart minio after 5s
* Observe minio does not start up because KES cache was purged before it could be refreshed by minio

## Types of changes
- [X] Optimization (provides speedup with no functional changes)

## Checklist:


cmd/common-main.go
```diff
@@ -962,12 +962,23 @@ func handleKMSConfig() {
 		if err != nil {
 			logger.Fatal(err, "Unable to initialize a connection to KES as specified by the shell environment")
 		}
-		// We check that the default key ID exists or try to create it otherwise.
-		// This implicitly checks that we can communicate to KES. We don't treat
-		// a policy error as failure condition since MinIO may not have the permission
+		// Try to generate/decrypt a data encryption key. Only try to create key if this fails.
```
aead: Why can we not simply check whether the key exists? For example, like this:
```
		if _, err = KMS.GenerateKey(GlobalContext, env.Get(kms.EnvKESKeyName, ""), kms.Context{}); err != nil && errors.Is(err, kes.ErrKeyNotFound) {
			if err = KMS.CreateKey(GlobalContext, env.Get(kms.EnvKESKeyName, "")); err != nil && !errors.Is(err, kes.ErrKeyExists) {
				logger.Fatal(err, "Unable to initialize a connection to KES as specified by the shell environment")
			}
		}
```

This seems far less implicit instead of passing values throuch `context.Context`.

aead (changes requested):

internal/kms/kes.go
```diff
@@ -416,11 +441,16 @@ func (c *kesClient) Verify(ctx context.Context) []VerifyResult {
...
+		state := kes.State{}
+		// Permission on /status is not required to validate kes key encryption and decryption
+		if ctx.Value(CtxValidKey{}) != "true" {
+			// 1. Get stats for the KES instance
+			clientState, err := client.Status(ctx)
+			if err != nil {
+				results = append(results, VerifyResult{Status: "offline", Endpoint: endpoint})
+				continue
+			}
+			state = clientState
```
aead: This seems not necessary

allanrogerr: @aead I am passing values thru the context in order to reuse the Verify function already existing- without requiring the user to have the /status permission and without modifying any other functions which in turn call Verify.

aead: Yes, but I don't understand why we need to call `Verify` at all. Refer to: https://github.com/minio/minio/pull/19492#discussion_r1562596492

allanrogerr: @aead I call Verify since it already checks if the key can be both generated and also can be used to decrypt. So there is certainty behind the key’s validity. Since Verify already exists I decided to not duplicate the code and simply call it.

If you are sure that the Generate function is sufficient then the code simplification that you suggested makes complete sense. Could you confirm that all that is needed is Generate?

cmd/common-main.go
allanrogerr: Implemented

internal/kms/kes.go
allanrogerr: Implemented

aead (approved):

cmd/common-main.go
```diff
@@ -962,12 +962,14 @@ func handleKMSConfig() {
...
-		// We check that the default key ID exists or try to create it otherwise.
-		// This implicitly checks that we can communicate to KES. We don't treat
-		// a policy error as failure condition since MinIO may not have the permission
+		// Try to generate a data encryption key. Only try to create key if this fails.
+		// This implicitly checks that we can communicate to KES.
+		// We don't treat a policy error as failure condition since MinIO may not have the permission
 		// to create keys - just to generate/decrypt data encryption keys.
-		if err = KMS.CreateKey(context.Background(), env.Get(kms.EnvKESKeyName, "")); err != nil && !errors.Is(err, kes.ErrKeyExists) && !errors.Is(err, kes.ErrNotAllowed) {
-			logger.Fatal(err, "Unable to initialize a connection to KES as specified by the shell environment")
+		if _, err = KMS.GenerateKey(GlobalContext, env.Get(kms.EnvKESKeyName, ""), kms.Context{}); err != nil && errors.Is(err, kes.ErrKeyNotFound) {
```
vadmeste, line 970: Let's say if KES server (not KMS) itself is offline, it doesn't seem we call logger.Fatal() in that case, I think it was failing before this change, can you please confirm ?

allanrogerr: @vadmeste I just tested. If KES is offline then `KMS.CreateKey` fails

internal/kms/kes.go
```diff
@@ -138,9 +139,30 @@ func NewWithConfig(config Config) (KMS, error) {
...
+
+	ticker := time.NewTicker(10 * time.Second)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-ticker.C:
+		case <-ctx.Done():
+			return
+		}
+		_ = c.Verify(ctx)
```
allanrogerr: This requires the /status permission to be granted to KES

cmd/common-main.go
allanrogerr: e.g. `ERROR Unable to initialize a connection to KES as specified by the shell environment: kes: connection error: Post "https://10.214.226.181:9073/v1/key/create/minio-key": dial tcp 10.214.226.181:9073: connect: connection refused`

harshavardhana: Please update the comment with relevant permissions needed on the kes/kms side such as 

```
allow: 
 - /v1/<name>/*
```

internal/kms/kes.go
```diff
@@ -138,9 +139,30 @@ func NewWithConfig(config Config) (KMS, error) {
 			}
 		}
 	}()
+
+	go func() {
+		c.keepKeyInCache()
+	}()
```
harshavardhana:
```suggestion
	go c.refreshKMSMasterKeyCache()
```

internal/kms/kes.go
```diff
@@ -138,9 +139,30 @@ func NewWithConfig(config Config) (KMS, error) {
...
 	return c, nil
 }
 
+// Request KES keep an up-to-date copy of the KMS master key to allow minio to start up even if KMS is down. The
+// cached key may still be evicted if the period of this function is longer than that of KES .cache.expiry.unused
+func (c *kesClient) keepKeyInCache() {
+	ctx := context.Background()
+
+	ticker := time.NewTicker(10 * time.Second)
```
harshavardhana:
```suggestion
	ticker := time.NewTicker(10 * time.Second)
```

Please make this configurable via MINIO_KMS_KEY_CACHE_INTERVAL=10s

harshavardhana: Do not use a ticker because ValidateKey()'s response time would account for the ticker firing, causing misfires.

Use a timer instead to check how `cmd/erasures-sets.go` 

```
// cleanup ".trash/" folder every 5m minutes with sufficient sleep cycles, between each
// deletes a dynamic sleeper is used with a factor of 10 ratio with max delay between
// deletes to be 2 seconds.
func (s *erasureSets) cleanupDeletedObjects(ctx context.Context) {
        timer := time.NewTimer(globalAPIConfig.getDeleteCleanupInterval())
        defer timer.Stop()

```

internal/kms/kes.go
```diff
@@ -450,3 +472,44 @@ func (c *kesClient) Verify(ctx context.Context) []VerifyResult {
 	}
 	return results
 }
+
+// ValidateKey checks the validity of the KMS Master Key
+func (c *kesClient) ValidateKey(ctx context.Context) []VerifyResult {
```
harshavardhana:
```suggestion
func (c *kesClient) RefreshKey(ctx context.Context) []VerifyResult {
```

internal/kms/kes.go
```diff
@@ -138,9 +139,30 @@ func NewWithConfig(config Config) (KMS, error) {
...
+
+	ticker := time.NewTicker(10 * time.Second)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-ticker.C:
+		case <-ctx.Done():
+			return
+		}
+		_ = c.ValidateKey(ctx)
```
harshavardhana:
```suggestion
		_ = c.ValidateKey(ctx)
```

what if ValidateKey returns an error result? who will monitor this?

harshavardhana (changes requested):

internal/kms/kes.go
```diff
@@ -450,3 +472,34 @@ func (c *kesClient) Verify(ctx context.Context) []VerifyResult {
...
+		kmsCtx, err := kmsContext.MarshalText()
+		if err != nil {
+			validKey = false
+			break
+		}
+		_, err = client.GenerateKey(ctx, env.Get(EnvKESKeyName, ""), kmsCtx)
+		if err != nil {
+			validKey = false
+			break
+		}
```
harshavardhana: I think we must use loggerOnce here to log use something like

```go
func kmsLogOnceIf(ctx context.Context, err error, id string, errKind ...interface{}) {
        logger.LogOnceIf(ctx, "kms", err, id, errKind...)
}
```

Add this function to `cmd/logging.go` and use it here

```
kmsLogOnceIf(ctx, err, "refresh-kms-master-key")
```

harshavardhana (changes requested):

internal/kms/kes.go
allanrogerr: It's OK if it returns a result - the point of this code is to refresh the KES cache if it is refreshable i.e. if KES is available. If it fails then KES is unavailable and it does not matter if this code fails. 
I will make this function return a bool and simply invoke it within the timer.

allanrogerr: As you mention below - I will add a loggerOnce to this function

internal/kms/kes.go
allanrogerr: Implemented

internal/kms/kes.go
allanrogerr: Implemented

internal/kms/kes.go
allanrogerr: Implemented

internal/kms/kes.go
allanrogerr: Implemented

internal/kms/kes.go
```diff
@@ -138,9 +140,44 @@ func NewWithConfig(config Config) (KMS, error) {
...
+
+	timer := time.NewTimer(time.Duration(cacheInterval) * time.Second)
+	defer timer.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-timer.C:
+			var wg sync.WaitGroup
```
harshavardhana: Why this wait group? Not very useful you can keep it blocking just like before.

I mean wait group here is blocking as well when you have only 1 goroutine

allanrogerr: True. I've implemented your suggestion.

harshavardhana (approved):

internal/kms/kes.go
```diff
@@ -138,9 +140,38 @@ func NewWithConfig(config Config) (KMS, error) {
...
 	return c, nil
 }
 
+// Request KES keep an up-to-date copy of the KMS master key to allow minio to start up even if KMS is down. The
+// cached key may still be evicted if the period of this function is longer than that of KES .cache.expiry.unused
+func (c *kesClient) refreshKMSMasterKeyCache(logger Logger) {
+	ctx := context.Background()
+
+	defaultCacheInterval := 10
+	cacheInterval, err := env.GetInt("EnvKESKeyCacheInterval", defaultCacheInterval)
```
harshavardhana, line 154: Before release we should fix this, input for ENV interval is generally provided as time.Duration format. 

Taking as int is not easy to document, instead take time.Duration as input and if users < time.Second simply assume the default by logging a warning.

harshavardhana: https://github.com/minio/pkg/pull/108 here is the relevant helper

PR #19507, opened 2024-04-15
Add cluster config metrics in metrics-v3

anjalshireesh:

## Description

endpoint: /minio/metrics/v3/cluster/config
metrics:
- rrs_parity
- standard_parity

## Motivation and Context

https://github.com/miniohq/engineering/issues/1510

## How to test this PR?

`mc admin prometheus metrics <alias> cluster`

should return the new cluster config related metrics
(will need some changes in mc to invoke the metrics-v3 api)

## Types of changes

## Checklist:

### DIFF
--- a/cmd/metrics-v3-cluster-config.go
+++ b/cmd/metrics-v3-cluster-config.go
@@ -1,0 +1,49 @@
+// Copyright (c) 2015-2024 MinIO, Inc.
+//
+// This file is part of MinIO Object Storage stack
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package cmd
+
+import "context"
+
+const (
+	configWriteQuorum    = "write_quorum"
+	configRRSParity      = "rrs_parity"
+	configStandardParity = "standard_parity"
+)
+
+var (
+	configWriteQuorumMD = NewGaugeMD(configWriteQuorum,
+		"Maximum write quorum across all pools and sets")
+	configRRSParityMD = NewGaugeMD(configRRSParity,
+		"Reduced redundancy storage class parity")
+	configStandardParityMD = NewGaugeMD(configStandardParity,
+		"Standard storage class parity")
+)
+
+// loadClusterConfigMetrics - `MetricsLoaderFn` for cluster config
+// such as standard and RRS parity.
+func loadClusterConfigMetrics(ctx context.Context, m MetricValues, c *metricsCache) error {
+	clusterDriveMetrics, _ := c.clusterDriveMetrics.Get()
+	m.Set(configStandardParity, float64(clusterDriveMetrics.storageInfo.Backend.StandardSCParity))
+	m.Set(configRRSParity, float64(clusterDriveMetrics.storageInfo.Backend.RRSCParity))
+
+	objLayer := newObjectLayerFn()
+	result := objLayer.Health(ctx, HealthOptions{})
+	m.Set(configWriteQuorum, float64(result.WriteQuorum))
+
+	return nil
+}
--- a/cmd/metrics-v3-cluster-config.go
+++ b/cmd/metrics-v3-cluster-config.go
@@ -42,8 +42,10 @@ func loadClusterConfigMetrics(ctx context.Context, m MetricValues, c *metricsCac
 	m.Set(configRRSParity, float64(clusterDriveMetrics.storageInfo.Backend.RRSCParity))
 
 	objLayer := newObjectLayerFn()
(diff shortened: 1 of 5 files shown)


cmd/metrics-v3-cluster-config.go
```diff
@@ -0,0 +1,49 @@
...
+	configRRSParityMD = NewGaugeMD(configRRSParity,
+		"Reduced redundancy storage class parity")
+	configStandardParityMD = NewGaugeMD(configStandardParity,
+		"Standard storage class parity")
+)
+
+// loadClusterConfigMetrics - `MetricsLoaderFn` for cluster config
+// such as standard and RRS parity.
+func loadClusterConfigMetrics(ctx context.Context, m MetricValues, c *metricsCache) error {
+	clusterDriveMetrics, _ := c.clusterDriveMetrics.Get()
```
vadmeste: Do we really need to ignore this error ?

cmd/metrics-v3-cluster-config.go
```diff
@@ -0,0 +1,49 @@
...
+	m.Set(configStandardParity, float64(clusterDriveMetrics.storageInfo.Backend.StandardSCParity))
+	m.Set(configRRSParity, float64(clusterDriveMetrics.storageInfo.Backend.RRSCParity))
+
+	objLayer := newObjectLayerFn()
```
vadmeste: Let's return an error if objLayer is nil

harshavardhana: That must be done centrally before even coming here.

cmd/metrics-v3-cluster-config.go
anjalshireesh: @vadmeste my understanding is that `Get()` will always return a value (either already cached one or a newly retrieved one), so the error can be ignored here, but I could be wrong. I've taken this code from `metrics-v3-cluster-health.go`

@donatello can you please confirm what's the right way?

cmd/metrics-v3-cluster-config.go
anjalshireesh: @vadmeste for now changing to use it only if not `nil`

@harshavardhana I didn't understand, could you please elaborate?

minio-trusted: mainly look at metrics-v2.go object layer is checked centrally to avoid passing a nil before populating the values.

anjalshireesh: Thanks @harshavardhana , I've added logic to centrally check if the obj layer is initialized (when the metric group needs it).

This is done at the metric group level, similar to metrics-v2. One issue with this is that, none of the metrics from that group will be loaded till the obj layer is initialized, even when some of them don't really depend on the obj layer e.g. the `standard_parity` and `rrs_parity` in this case.

Let me know what you think..

harshavardhana:
``` suggestion
	objLayer := newObjectLayerFn()
```

This can be nil; pass this via `loader()` as an optional value.

cmd/metrics-v3-types.go
```diff
@@ -408,6 +411,14 @@ func (mg *MetricsGroup) AddExtraLabels(labels ...string) {
 	}
 }
 
+// requiresObjectLayer - use to add dependency on the global object layer API
+// when creating the metrics group
+// e.g. myMG := NewMetricsGroup(...).requiresObjectLayer()
+func (mg *MetricsGroup) requiresObjectLayer(labels ...string) *MetricsGroup {
```
harshavardhana: why label input?

harshavardhana (changes requested):

cmd/metrics-v3-types.go
anjalshireesh: thanks, removed

cmd/metrics-v3-cluster-config.go
anjalshireesh: the function will not be called when objLayer is not initialized; so it cannot be nil here.

harshavardhana: Yeah but that is a very ugly style. In future when this code will move elsewhere it will crash.

anjalshireesh: @harshavardhana I added the central checking because of this comment: https://github.com/minio/minio/pull/19507#discussion_r1580620169

> mainly look at metrics-v2.go object layer is checked centrally to avoid passing a nil before populating the values.

Did I misunderstand it? If so, and if that's not required, then the previous code should just work fine? something like

```
objLayer := newObjectLayerFn()
if objLayer != nil {
 // load metrics
}
```

anjalshireesh: @harshavardhana I found that the the metric for which I was using `objLayer` (`write_quorum`) is already implemented under `/cluster/erasure-set` and would have become a duplicate here. So I have removed this metric (and other changes related to it) altogether from this PR.

cmd/metrics-v3-cluster-config.go
```diff
@@ -0,0 +1,49 @@
...
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package cmd
+
+import "context"
+
+const (
+	configWriteQuorum    = "write_quorum"
```
shtripat: don't see populating this metric. needed?

anjalshireesh: @shtripat thanks, removed.

shtripat (approved):
lgtm

klauspost (approved):

harshavardhana (approved):

PR #19519, opened 2024-04-16
ftp: Return current time for prefixes/directories

klauspost:
## Description

Some clients (ok, Filezilla) doesn't like the zero date on directories and refuses to show entries without a date.

Prefixes do not have a time set. Return current time for these entries.

## How to test this PR?

Observe timestamps on prefixes.

## Types of changes

### DIFF
--- a/cmd/ftp-server-driver.go
+++ b/cmd/ftp-server-driver.go
@@ -82,7 +82,10 @@ func (m *minioFileInfo) Mode() os.FileMode {
 }
 
 func (m *minioFileInfo) ModTime() time.Time {
-	return m.info.LastModified
+	if !m.info.LastModified.IsZero() {
+		return m.info.LastModified
+	}
+	return time.Now().UTC()
 }
 
 func (m *minioFileInfo) IsDir() bool {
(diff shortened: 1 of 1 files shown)


ramondeklein:
Wouldn't it make more sense to return Epoch start instead? That would yield consistent results.

cmd/ftp-server-driver.go
```diff
@@ -82,7 +82,10 @@ func (m *minioFileInfo) Mode() os.FileMode {
 }
 
 func (m *minioFileInfo) ModTime() time.Time {
-	return m.info.LastModified
+	if !m.info.LastModified.IsZero() {
+		return m.info.LastModified
+	}
+	return time.Now().UTC()
```
ramondeklein: Using a fixed timestamp would yield in consistent results during different runs. I would prefer to use 1-1-1970 (epoch), but MS-DOS timestamps only supports dates from 1-1-1980 ([source](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-filetimetodosdatetime#remarks)), so it may be safer to use that instead.
```suggestion
        return minFileDate;
```
Where `minFileDate` is defined as
```golang
var minFileData = time.Date(1980, 1, 1, 0, 0, 0, 0, time.UTC)    // Workaround for Filezilla
```

ramondeklein (changes requested):

zveinn (approved):
lgtm

ramondeklein (approved):
@klauspost If you agree with using a fixed date, then it LGTM

PR #19524, opened 2024-04-16
list: Fix rare listing continuation freeze

vadmeste:

## Description
Reading the list metacache is not protected by a lock, the code retries when it fails to read the metacache object, however it forgot to re-read the metacache object from the drives, which is necessary, especially if the metacache objcet is inlined.

This commit will ensure that we always re-read the metacache object from the drives when it is retrying.

## Motivation and Context
Fix listing freeze

## How to test this PR?
Create a bucket with many objects and then do a recursive listing in a loop and wait until the listing stucks

## Types of changes

## Checklist:

### DIFF
--- a/cmd/metacache-set.go
+++ b/cmd/metacache-set.go
@@ -573,6 +573,7 @@ func (er *erasureObjects) streamMetadataParts(ctx context.Context, o listPathOpt
 					continue
 				case InsufficientReadQuorum:
 					retries++
+					loadedPart = -1
 					time.Sleep(retryDelay250)
 					continue
 				default:


harshavardhana (approved):

klauspost:
Maybe instead set `loadedPart` if there is a read quorum error:

```Go
				case InsufficientReadQuorum:
					retries++
					loadedPart = -1 // Reload metadata
					time.Sleep(retryDelay250)
					continue
```

That would be considerably cheaper. Could you see that working?

vadmeste:
yes, that is possible. I only like retry because it is generic, I can make the change, probably for ObjectNotFound error just above, however I do not see how this can happen

klauspost:
@vadmeste IIRC `ObjectNotFound` will happen until the block is complete - though it may be polling block 0 metadata until it is updated. Can't remember... 

Does it ever inline data? I would kinda expect it not to do that ever.

vadmeste:
> Does it ever inline data? I would kinda expect it not to do that ever.

yes, an metacache object can be inlined, but why is that a problem @klauspost ?

harshavardhana:
Can you make the change so that we can add this for the next release? @vadmeste ?

PR #19528, opened 2024-04-17
ldap: Normalize DNs when importing

donatello:
## Description

This is a change to IAM export/import functionality. For LDAP enabled
setups, it performs additional validations:

- for policy mappings on LDAP users and groups, it ensures that the
corresponding user or group DN exists and if so uses a normalized form
of these DNs for storage

- for access keys (service accounts), it updates (i.e. validates
existence and normalizes) the internally stored parent user DN and group
DNs.

This allows for a migration path for setups in which LDAP mappings have
been stored in previous versions of the server, where the name of the
mapping file stored on drives is not in a normalized form.

An administrator needs to execute:

`mc admin iam export ALIAS`

followed by

`mc admin iam import ALIAS /path/to/export/file`

The validations are more strict and returns errors when multiple
mappings are found for the same user/group DN. This is to ensure the
mappings stored by the server are unambiguous and to reduce the
potential for confusion.

Bonus **bug fix**: IAM export of access keys (service accounts) did not
export key name, description and exipration. This is fixed in this
change too.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/iam.go
+++ b/cmd/iam.go
@@ -32,6 +32,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	libldap "github.com/go-ldap/ldap/v3"
 	"github.com/minio/madmin-go/v3"
 	"github.com/minio/minio-go/v7/pkg/set"
 	"github.com/minio/minio/internal/arn"
@@ -48,6 +49,7 @@ import (
 	xioutil "github.com/minio/minio/internal/ioutil"
 	"github.com/minio/minio/internal/jwt"
 	"github.com/minio/minio/internal/logger"
+	"github.com/minio/pkg/v2/ldap"
 	"github.com/minio/pkg/v2/policy"
 	etcd "go.etcd.io/etcd/client/v3"
 )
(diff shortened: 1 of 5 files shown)


cmd/iam.go
```diff
@@ -1475,6 +1477,87 @@ func (sys *IAMSys) updateGroupMembershipsForLDAP(ctx context.Context) {
...
+	}
+
+	// map of normalized DN keys to original keys.
+	normalizedDNKeysMap := make(map[string][]string)
+	var collectedErrors []error
+	for k := range policyMap {
+		_, err := ldap.NormalizeDN(k)
+		if err != nil {
+			// not a valid DN, ignore.
+			continue
```
vadmeste, line 1599: let's add a log for easier troubleshooting in the future

cmd/iam.go
```diff
@@ -1475,6 +1477,87 @@ func (sys *IAMSys) updateGroupMembershipsForLDAP(ctx context.Context) {
 	}
 }
 
+// ValidateLDAPMappingImport - validates the LDAP policy mappings. Keys in the
+// given map may not correspond to LDAP DNs - these keys are ignored.
+//
+// For validated mappings, it updates the key in the given map to be in
+// normalized form.
+func (sys *IAMSys) ValidateLDAPMappingImport(ctx context.Context, isGroup bool,
```
vadmeste: s/ValidateLDAPMappingImport/NormalizeLDAPMappingImport

Validate can be confusing

vadmeste (commented):
Few comments

harshavardhana (changes requested):
We have to also fix the ParentUser for all existing 

- service accounts
- unexpired STS users

cmd/iam.go
donatello: If they don't parse as DNs it's most likely left over data from builtin IDP, etc. We don;t need to remove or do anything about those, they should still get imported. I am not sure logging is useful here - is there any potential reason?

donatello:
> We have to also fix the ParentUser for all existing
> 
>     * service accounts
> 
>     * unexpired STS users

Added changes to handle service accounts. STS accounts are not handled as we dont support STS account import/export currently.

cmd/iam.go
```diff
@@ -1475,6 +1477,176 @@ func (sys *IAMSys) updateGroupMembershipsForLDAP(ctx context.Context) {
...
+			// not a valid DN, ignore.
+			continue
+		}
+		validatedDN, err := dnValidator(conn, k)
+		if err != nil {
+			collectedErrors = append(collectedErrors, fmt.Errorf("could not validate `%s` exists in LDAP directory: %w", k, err))
+			continue
+		}
+		if validatedDN == "" {
+			err := fmt.Errorf("DN `%s` was not found in the LDAP directory", k)
```
harshavardhana, line 1607: if the DN is not found should we return an error?

cmd/iam.go
```diff
@@ -1475,6 +1477,176 @@ func (sys *IAMSys) updateGroupMembershipsForLDAP(ctx context.Context) {
...
+		var collectedErrMsg strings.Builder
+		collectedErrMsg.WriteString("errors validating LDAP DN: ")
+		for i, err := range collectedErrors {
+			if i > 0 {
+				collectedErrMsg.WriteString("; ")
+			}
+			collectedErrMsg.WriteString(fmt.Sprintf("%d: %v", i, err))
+		}
+		return fmt.Errorf(collectedErrMsg.String())
+	}
```
harshavardhana: you can do perhaps in a single line? 

```
fmt.Errorf("errors validating LDAP DN: %v", errors.Join(collectedErrs...))
```

?

cmd/iam.go
```diff
@@ -1475,6 +1477,176 @@ func (sys *IAMSys) updateGroupMembershipsForLDAP(ctx context.Context) {
...
+
+	for normKey, origKeys := range normalizedDNKeysMap {
+		if len(origKeys) > 1 {
+			return fmt.Errorf("multiple DNs map to the same LDAP DN[%s]: %v; please remove DNs that are not needed",
```
harshavardhana, line 1624: Can we indicate the `mc` command to do this?

harshavardhana (commented):
Small comment, can you also get a sample `iam-exported.zip` file into the repo as assets and hide the entries such that we can see import working with expected credentials?

cmd/iam.go
donatello: because it looks like a dn (parses) we should consider an error i think. Rarely it's going to be because the user is removed from LDAP.

harshavardhana:
Please add 

```diff
diff --git a/.typos.toml b/.typos.toml
index 38bdfb208..9f67ce391 100644
--- a/.typos.toml
+++ b/.typos.toml
@@ -15,6 +15,7 @@ extend-ignore-re = [
     "eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiJ9.*",
     "MIIDBTCCAe2gAwIBAgIQWHw7h.*",
     'http\.Header\{"X-Amz-Server-Side-Encryptio":',
+    "ZoEoZdLlzVbOlT9rbhD7ZN7TLyiYXSAlB79uGEge",
 ]
 
 [default.extend-words]
```

harshavardhana (approved):

PR #19534, opened 2024-04-17
Store object EC in metadata header

klauspost:
## Description

Keep the EC in header, so it can be retrieved easily for exact object quorum calculations.

To not force a full metadata decode on every read, the value will be 0/0 for data written in previous versions.

Size is expected to increase by 2 bytes per version, since all valid values can be represented with 1 byte each.

Example:
```
λ xl-meta xl.meta
{
  "Versions": [
    {
      "Header": {
        "EcM": 4,
        "EcN": 8,
        "Flags": 6,
        "ModTime": "2024-04-17T11:46:25.325613+02:00",
        "Signature": "0a409875",
        "Type": 1,
        "VersionID": "8e03504e11234957b2727bc53eda0d55"
      },
...
```

Not used for operations yet.

## How to test this PR?

Upgrade existing, mainly check that old versions are read correctly (verified locally).

## Types of changes

### DIFF
--- a/cmd/xl-storage-format-v2.go
+++ b/cmd/xl-storage-format-v2.go
@@ -250,15 +250,17 @@ type xlMetaV2VersionHeader struct {
 	Signature [4]byte
 	Type      VersionType
 	Flags     xlFlags
+	EcM, EcN  uint8 // Note that these will be 0/0 for non-v2 objects and older xl.meta
 }
 
 func (x xlMetaV2VersionHeader) String() string {
-	return fmt.Sprintf("Type: %s, VersionID: %s, Signature: %s, ModTime: %s, Flags: %s",
+	return fmt.Sprintf("Type: %s, VersionID: %s, Signature: %s, ModTime: %s, Flags: %s, N: %d, M: %d",
 		x.Type.String(),
 		hex.EncodeToString(x.VersionID[:]),
 		hex.EncodeToString(x.Signature[:]),
 		time.Unix(0, x.ModTime),
 		x.Flags.String(),
+		x.EcN, x.EcM,
 	)
 }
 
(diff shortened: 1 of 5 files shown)


cmd/xl-storage-format-v2.go
```diff
@@ -274,6 +276,11 @@ func (x xlMetaV2VersionHeader) matchesNotStrict(o xlMetaV2VersionHeader) bool {
 		x.Type == o.Type
 }
 
+// hasEC will return true if the version has erasure coding information.
+func (x xlMetaV2VersionHeader) hasEC() bool {
+	return x.EcM > 0 && x.EcN > 0
```
krisis, line 281: Shouldn't `hasEC()` return true in the case of SNSD when it is recorded in xlMeta version header? It would be x.EcM = 1 and x.EcN = 0.

krisis (commented):
Good stuff, much needed. I have only one comment/question. LGTM otherwise.

cmd/xl-storage-format-v2.go
harshavardhana: @klauspost ^^

klauspost: @krisis It probably just the naming that confuses you. Doc says `true if the version has erasure coding information` - key being "information". 

It does not tell you (by itself) if the version has parity shards. That would be `x.hasEC() && x.EcN > 0` - with a fallback to decoding the metadata if you need that information.

krisis: Ah, it is clear from the method docs. Sorry, I didn't read it carefully earlier. LGTM.

krisis (approved):

harshavardhana (approved):

PR #19547, opened 2024-04-18
fix: ldap: avoid unnecessary import errors

donatello:
## Description
Follow up for #19528

If there are multiple existing DN mappings for the same normalized DN, if they all have the same policy mapping value, we pick one of them of them instead of returning an import error.

## How to test this PR?

Test updated.

## Types of changes

## Checklist:


harshavardhana (approved):

harshavardhana:
Can you add a test for this ?

donatello:
> Can you add a test for this ?

Test is already updated for this case. I don't think we need a case that actually errors out.

harshavardhana (approved):

PR #19548, opened 2024-04-18
make renameData() more defensive during overwrites

harshavardhana:

## Description
make renameData() more defensive during overwrites

## Motivation and Context
instead upon any error in renameData(), we still
preserve the existing dataDir in some form for
recoverability in strange situations, such as out
of disk space type errors.

Bonus: avoid running list and heal() instead, allow 
versions disparity to return the actual versions,
UUID to heal. Currently, this can be limited to `0 
versions and fewer disparate objects.

An undo now reverts the `xl.meta` from `xl.meta.bkp` 
during overwrites on such flaky setups.

Bonus: Save N depth syscalls by skipping the parents 
upon overwrites and versioned updates.

Flaky setup examples are stretch clusters with regular 
packet drops, etc. We need to add some defensive code 
to avoid dangling objects.

## How to test this PR?
CI/CD should cover these already.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/erasure-metadata-utils.go
+++ b/cmd/erasure-metadata-utils.go
@@ -19,6 +19,7 @@ package cmd
 
 import (
 	"context"
+	"encoding/binary"
 	"errors"
 	"hash/crc32"
 
@@ -26,14 +27,17 @@ import (
 )
 
 // figure out the most commonVersions across disk that satisfies
-// the 'writeQuorum' this function returns '0' if quorum cannot
+// the 'writeQuorum' this function returns "" if quorum cannot
 // be achieved and disks have too many inconsistent versions.
-func reduceCommonVersions(diskVersions []uint64, writeQuorum int) (commonVersions uint64) {
+func reduceCommonVersions(diskVersions [][]byte, writeQuorum int) (versions []byte) {
 	diskVersionsCount := make(map[uint64]int)
 	for _, versions := range diskVersions {
-		diskVersionsCount[versions]++
+		if len(versions) > 0 {
+			diskVersionsCount[binary.BigEndian.Uint64(versions)]++
+		}
 	}
 
+	var commonVersions uint64
 	max := 0
 	for versions, count := range diskVersionsCount {
 		if max < count {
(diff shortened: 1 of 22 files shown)


cmd/erasure-metadata-utils.go
```diff
@@ -43,10 +47,38 @@ func reduceCommonVersions(diskVersions []uint64, writeQuorum int) (commonVersion
...
+				return versions
+			}
+		}
+	}
+
+	return []byte{}
+}
+
+// figure out the most commonVersions across disk that satisfies
+// the 'writeQuorum' this function returns '0' if quorum cannot
```
klauspost, line 61:
```suggestion
// the 'writeQuorum' this function returns "" if quorum cannot
```

cmd/storage-datatypes.go
```diff
@@ -490,8 +493,14 @@ type WriteAllHandlerParams struct {
...
 
 // RenameDataResp - RenameData()'s response.
+// Provides information about the final state of Rename()
+//   - on xl.meta (array of versions) on disk to check for version disparity
+//   - on rewrite dataDir on disk that must be additionally purged
+//     only after as a 2-phase call, allowing the older dataDir to
+//     hang-around in-case we need some form of recovery.
 type RenameDataResp struct {
-	Signature uint64 `msg:"sig"`
+	Sign       []byte
```
klauspost, line 501: You should add a new `HandlerID` since the response isn't backwards compatible. `HandlerRenameData2` for example.

This will make `HandlerRenameData` calls from previous versions fail and `HandlerRenameData2` fail on newer versions in a mixed env.

Otherwise you will just get the zero value back from there.

cmd/erasure-metadata-utils.go
```diff
@@ -43,10 +47,38 @@ func reduceCommonVersions(diskVersions []uint64, writeQuorum int) (commonVersion
 	}
 
 	if max >= writeQuorum {
-		return commonVersions
+		for _, versions := range diskVersions {
+			if binary.BigEndian.Uint64(versions) == commonVersions {
+				return versions
+			}
+		}
```
vadmeste, line 54: it looks like `return commonVersions` is enough

cmd/xl-storage.go
```diff
@@ -1366,6 +1369,10 @@ func (s *xlStorage) DeleteVersion(ctx context.Context, volume, path string, fi F
 		return s.WriteAll(ctx, volume, pathJoin(path, xlStorageFormatFile), buf)
 	}
 
+	if opts.UndoWrite && fi.OldDataDir != "" {
```
vadmeste: It does not seem it is needed to include OldDataDir in FileInfo{}. We can add it to opts here. It looks more clean that way.

internal/grid/handlers.go
```diff
@@ -56,7 +56,7 @@ const (
 	HandlerUpdateMetadata
 	HandlerWriteMetadata
 	HandlerCheckParts
-	HandlerRenameData
+	HandlerRenameData2
```
klauspost: Renaming will not change the ID. Put it at the end and keep `HandlerRenameData` - maybe as `HandlerRenameDataLegacy`.

```
// Do NOT remove or change the order of existing handlers.
```

klauspost: You are forcing me to make a checksum of all the handler names :D

cmd/xl-storage.go
harshavardhana: Yeah we can let me do that.

cmd/erasure-metadata-utils.go
harshavardhana: Yes that is correct

internal/grid/handlers.go
harshavardhana: Yeah let me do that.

cmd/erasure-metadata-utils.go
harshavardhana: we need to return the []byte slice not uint64

harshavardhana:
PTAL @klauspost @vadmeste - I need this available for the next release. If you don't see a problem with this, we can merge this.

klauspost (approved):
LGTM

vadmeste (approved):
LGTM

PR #19553, opened 2024-04-19
cleanup: Simplify usage of MinIOSourceProxyRequest

rluetzner:

## Description

This replaces a convoluted condition that ultimately evaluated to

"is this HTTP header present in the request or not"

, similar to what happens with MinIOSourceReplicationRequest a few lines further down in `cmd/object-api-options.go`.

We only care that the HTTP header is set, the actual value of it is irrelevant.

## Motivation and Context

Just cleaning up.

## How to test this PR?

-

## Types of changes

## Checklist:

### DIFF
--- a/cmd/object-api-options.go
+++ b/cmd/object-api-options.go
@@ -67,9 +67,7 @@ func getDefaultOpts(header http.Header, copySource bool, metadata map[string]str
 		opts.ServerSideEncryption = encrypt.NewSSE()
 	}
 	_, opts.ProxyRequest = header[xhttp.MinIOSourceProxyRequest]
-	if _, ok := header[xhttp.MinIOSourceReplicationRequest]; ok {
-		opts.ReplicationRequest = true
-	}
+	_, opts.ReplicationRequest = header[xhttp.MinIOSourceReplicationRequest]
 	opts.Speedtest = header.Get(globalObjectPerfUserMetadata) != ""
 	return
 }
--- a/cmd/object-api-options.go
+++ b/cmd/object-api-options.go
@@ -66,10 +66,7 @@ func getDefaultOpts(header http.Header, copySource bool, metadata map[string]str
 	if crypto.S3.IsRequested(header) || (metadata != nil && crypto.S3.IsEncrypted(metadata)) {
 		opts.ServerSideEncryption = encrypt.NewSSE()
 	}
-	if v, ok := header[xhttp.MinIOSourceProxyRequest]; ok {
-		opts.ProxyHeaderSet = true
-		opts.ProxyRequest = strings.Join(v, "") == "true"
-	}
+	_, opts.ProxyRequest = header[xhttp.MinIOSourceProxyRequest]
 	if _, ok := header[xhttp.MinIOSourceReplicationRequest]; ok {
 		opts.ReplicationRequest = true
 	}
--- a/cmd/bucket-replication.go
+++ b/cmd/bucket-replication.go
@@ -2226,7 +2226,7 @@ func getProxyTargets(ctx context.Context, bucket, object string, opts ObjectOpti
 	if opts.VersionSuspended {
 		return &madmin.BucketTargets{}
 	}
-	if opts.ProxyRequest || (opts.ProxyHeaderSet && !opts.ProxyRequest) {
+	if opts.ProxyRequest {
 		return &madmin.BucketTargets{}
 	}
 	cfg, err := getReplicationConfig(ctx, bucket)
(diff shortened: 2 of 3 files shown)


cmd/object-api-options.go
```diff
@@ -66,9 +66,8 @@ func getDefaultOpts(header http.Header, copySource bool, metadata map[string]str
 	if crypto.S3.IsRequested(header) || (metadata != nil && crypto.S3.IsEncrypted(metadata)) {
 		opts.ServerSideEncryption = encrypt.NewSSE()
 	}
-	if v, ok := header[xhttp.MinIOSourceProxyRequest]; ok {
-		opts.ProxyHeaderSet = true
-		opts.ProxyRequest = strings.Join(v, "") == "true"
+	if _, ok := header[xhttp.MinIOSourceProxyRequest]; ok {
+		opts.ProxyRequest = true
```
harshavardhana:
```suggestion
		_, opts.ProxyRequest = header[xhttp.MinIOSourceProxyRequest]
```

you wanted to simplify :-)

harshavardhana (changes requested):

cmd/object-api-options.go
rluetzner: Good catch. 🙂

shtripat (approved):
Looks good. Hope CI tests would cover the path.

harshavardhana (approved):

poornas:
The convoluted condition had a purpose - replication not only needs to proxy in active-active if object is missing on the site that received the request, but also another important task - it needs to figure out whether the object should be replicated [here](https://github.com/minio/minio/blob/928c0181bfaa2fbc4a4d1adc2a5106dd151de600/cmd/bucket-replication.go#L634) and [here](https://github.com/minio/minio/blob/928c0181bfaa2fbc4a4d1adc2a5106dd151de600/cmd/bucket-replication.go#L1401). 

The ProxyHeaderSet to "false" was to avoid re-proxying the request back to primary and prematurely mark replication complete by checking itself. GET/HEAD requests do not have any other identifying headers to know if it is a client request or replication request - it would be preferable to leave this as is without venturing into looking at user agent etc to demarcate a replication request

This PR needs to be reverted - otherwise, we will see the regression below again @harshavardhana 
https://github.com/minio/minio/commit/fd3f02637acf918a2db2bebede7f8bc99c1b9201

rluetzner:
Wow. This is hard to grok. I had to write it down again for me as well.

```go
if opts.ProxyRequest || (opts.ProxyHeaderSet && !opts.ProxyRequest)
```

So either ProxyRequest is true (in which case ProxyHeaderSet is also true) or it is false, in which case we're interested if the header was present in the request.

But let's rethink this.

Header is set to "true", so we get ProxyRequest as true. The second part after `||` doesn't matter. Everything evaluates to true.

Header is missing, so ProxyRequest is false and ProxyHeaderSet is false. Everything evaluates to false.

Header is set to "false", so ProxyRequest is false, but ProxyHeaderSet evaluates to true. The whole expression evaluates to true.

@poornas  Taking all of this, I'm still convinced my change is correct or the previous code was buggy. The only difference is of the HTTP header exists or not.

rluetzner:
In any case, even if the previous code was buggy, the fact that all places use exactly the same condition shows that it _can_ be simplified.

poornas:
This header morphed from a 2 state boolean to track  3 states - "request was proxied", "not set", "do not proxy this" for backward compatibility and to address the regression I mentioned. 

>Header is set to "true", so we get ProxyRequest as true. The second part after || doesn't matter. Everything evaluates to true.

here  GET/HEAD is proxied from peer site

>Header is missing, so ProxyRequest is false and ProxyHeaderSet is false. Everything evaluates to false.

could be client request or replication process checking whether to replicate - if no header is set, it will again proxy back from peer site to primary based on active-active config

> Header is set to "false", so ProxyRequest is false, but ProxyHeaderSet evaluates to true. The whole expression evaluates to true.
 - this was for replication process to use and is preventing premature completion of request and proxying back to self. your PR will essentially scuttle this.

rluetzner:
But again. If the header value is set to true, everything evaluates to true and if the header value is anything else, it'll also evaluate to true.

I can see how it should track three different States, but from the code I still don't see the difference.

poornas:
```
if opts.ProxyRequest || (opts.ProxyHeaderSet && !opts.ProxyRequest) {
    return  //don't proxy further since this already from peer site
}
proxy if conditions applicable..
```

rluetzner:
I've replaced this code

```
if opts.ProxyRequest || (opts.ProxyHeaderSet && !opts.ProxyRequest) {
    return  //don't proxy further since this already from peer site
}
proxy if conditions applicable..
```

with this here

```
if opts.ProxyHeaderSet {
    return  //don't proxy further since this already from peer site
}
proxy if conditions applicable..
```

I also renamed ProxyHeaderSet, which is why this might be confusing. ProxyHeaderSet is always true, when ProxyHeader is true.

And in case that ProxyHeader is false, but ProxyHeaderSet is true, the above code still does exactly the same.

You can still revert this if you want. I concede that there might be a chance that I've missed something or that you are keeping this around to be prepared for a certain situation (which I can't know).

I am however 99.9% sure that my transformation doesn't change anything.

PR #19563, opened 2024-04-20
do not panic on rebalance during server restarts

harshavardhana:

## Description
do not panic on rebalance during server restarts

## Motivation and Context
This PR makes a feasible approach to handle all the scenarios 
that we must face to avoid returning "panic."

Instead, we must return "errServerNotInitialized" when a 
bucketMetadataSys.Get() is called, allowing the caller to 
retry their operation and wait.

Bonus fix the way data-usage-cache stores the object. 
Instead of storing usage-cache.bin with the bucket as 
`.minio.sys/buckets`, the `buckets` must be relative 
to the bucket `.minio.sys` as part of the object name. 

Otherwise, there is no way to decommission entries at
`.minio.sys/buckets` and their final erasure set positions.

A bucket must never have a `/` in it. Adds code to read()
from existing data-usage.bin upon upgrade.

## How to test this PR?
CI/CD should test all the scenarios

## Types of changes

## Checklist:


klauspost (approved):
Wonderful <3

cmd/erasure-server-pool-rebalance.go
```diff
@@ -448,9 +452,11 @@ func (z *erasureServerPools) rebalanceBuckets(ctx context.Context, poolIdx int)
 		}
 
 		stopFn := globalRebalanceMetrics.log(rebalanceMetricRebalanceBucket, poolIdx, bucket)
-		err = z.rebalanceBucket(ctx, bucket, poolIdx)
-		if err != nil {
+		if err = z.rebalanceBucket(ctx, bucket, poolIdx); err != nil {
 			stopFn(err)
+			if errors.Is(err, errServerNotInitialized) || errors.Is(err, errBucketMetadataNotInitialized) {
+				continue
```
krisis, line 458: Ideally, we should retry this bucket again a few times before we move over to the next bucket. This (fix) is an improvement from what we have now. I can send a PR to tackle this subsequently.

krisis (approved):
Neat!

PR #19581, opened 2024-04-22
Fix DeleteObject API for unversioned objects with insufficient read q…

poornas:
…uorum

Since the object is being permanently deleted, lack of read quorum should not matter as long as sufficient disks online to complete the delete with parity requirements.

If several pools have the same object with insufficient read quorum, attempt to delete object from all the pools where it exists

## How to test this PR?
need to simulate drives offline/ inconsistent metadata for a multi pool setup

## Types of changes

## Checklist:

### DIFF
--- a/cmd/erasure-object.go
+++ b/cmd/erasure-object.go
@@ -1912,23 +1912,26 @@ func (er erasureObjects) DeleteObject(ctx context.Context, bucket, object string
 	versionFound := true
 	objInfo = ObjectInfo{VersionID: opts.VersionID} // version id needed in Delete API response.
 	goi, _, gerr := er.getObjectInfoAndQuorum(ctx, bucket, object, opts)
+	tryDel := false
 	if gerr != nil && goi.Name == "" {
 		if _, ok := gerr.(InsufficientReadQuorum); ok {
-			// Add an MRF heal for next time.
-			er.addPartial(bucket, object, opts.VersionID)
-
-			return objInfo, InsufficientWriteQuorum{}
+			if opts.Versioned || opts.VersionSuspended || countOnlineDisks(storageDisks) < len(storageDisks)/2+1 {
+				// Add an MRF heal for next time.
+				er.addPartial(bucket, object, opts.VersionID)
+				return objInfo, InsufficientWriteQuorum{}
+			}
+			tryDel = true // only for unversioned objects if there is write quorum
 		}
 		// For delete marker replication, versionID being replicated will not exist on disk
 		if opts.DeleteMarker {
 			versionFound = false
-		} else {
+		} else if !tryDel {
 			return objInfo, gerr
 		}
 	}
 
 	if opts.EvalMetadataFn != nil {
-		dsc, err := opts.EvalMetadataFn(&goi, err)
+		dsc, err := opts.EvalMetadataFn(&goi, gerr)
 		if err != nil {
 			return ObjectInfo{}, err
 		}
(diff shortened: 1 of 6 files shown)


cmd/erasure-object.go
```diff
@@ -1915,17 +1915,20 @@ func (er erasureObjects) DeleteObject(ctx context.Context, bucket, object string
...
 	if gerr != nil && goi.Name == "" {
 		if _, ok := gerr.(InsufficientReadQuorum); ok {
 			// Add an MRF heal for next time.
 			er.addPartial(bucket, object, opts.VersionID)
-
-			return objInfo, InsufficientWriteQuorum{}
+			if opts.Versioned || opts.VersionSuspended {
+				return objInfo, InsufficientWriteQuorum{}
+			}
+			tryDel = true // only for unversioned objects
```
harshavardhana: if the "version" is specified, can we do a permanent purge by marking xl.meta with "versionPurgeStatus" ?

cmd/erasure-object.go
```diff
@@ -1915,17 +1915,20 @@ func (er erasureObjects) DeleteObject(ctx context.Context, bucket, object string
 	versionFound := true
 	objInfo = ObjectInfo{VersionID: opts.VersionID} // version id needed in Delete API response.
 	goi, _, gerr := er.getObjectInfoAndQuorum(ctx, bucket, object, opts)
+	tryDel := false
 	if gerr != nil && goi.Name == "" {
 		if _, ok := gerr.(InsufficientReadQuorum); ok {
 			// Add an MRF heal for next time.
 			er.addPartial(bucket, object, opts.VersionID)
```
harshavardhana: don't add heal if we are doing to delete

poornas: fixed

cmd/erasure-object.go
poornas: versionPurgeStatus is relevant only for replication because it needs to maintain state. If there is sufficient quorum and metadata can be read, versioned delete can be attempted for non replication cases - perhaps we can check if suff disks online before DeleteVersion?

harshavardhana: Yeah definitely. My point was if delete was issued permanently we can update metadata forcibly?

poornas: my only concern is on whether bucket metadata is correctly loaded at the time of delete - as long as that is known, permanent delete can be forced through for regular versioned objects (barring object lock/replication)

harshavardhana (approved):

cmd/erasure-object.go
```diff
@@ -1912,23 +1912,26 @@ func (er erasureObjects) DeleteObject(ctx context.Context, bucket, object string
...
 	objInfo = ObjectInfo{VersionID: opts.VersionID} // version id needed in Delete API response.
 	goi, _, gerr := er.getObjectInfoAndQuorum(ctx, bucket, object, opts)
+	tryDel := false
 	if gerr != nil && goi.Name == "" {
 		if _, ok := gerr.(InsufficientReadQuorum); ok {
-			// Add an MRF heal for next time.
-			er.addPartial(bucket, object, opts.VersionID)
-
-			return objInfo, InsufficientWriteQuorum{}
+			if opts.Versioned || opts.VersionSuspended || countOnlineDisks(storageDisks) < len(storageDisks)/2+1 {
```
krisis, line 1918: It would be good to handle this without the `tryDel` boolean. It looks like the current code implicitly assumes that an error other than InsufficientReadQuorum is VersionNotFound, as evidenced by `if opts.DeleteMarker { versionFound = false ...`. If we make the error handling for getObjectInfoAndQuorum more robust, we can avoid this boolean flag to alter control flow. If you feel that this change is out of scope to what you wish to accomplish in this change, that is fine too. We can address that subsequently.

krisis (approved):

Issue #19536, opened 2024-04-17
Connection LDAP no longer works after update (normalization of DN names?)


ClementDe:
## NOTE

Hi,
Following the latest update on the normalization of DN names with LDAP, accounts are inaccessible at login.

The tokens are still active but I can no longer access the account.

it seems that the accounts already exist in this form in the policy:

CN=Toto Bou,OU=Users,OU=Blabla,OU=Global,DC=Blabla,DC=lan
does not work anymore

## Current Behavior

The accounts already configured on my minio instance are inaccessible following the update, I have the impression that the standardization has made them inaccessible.

If I run an add cmd again like this:

`mc idp ldap policy attach mgdev consoleAdmin --user="CN=Toto Bou,OU=Users,OU=Blabla,OU=Global,DC=Blabla,DC=lan"`

I have this:

`mc idp ldap policy entities ALIAS`

> Query time: 2024-04-17T12:58:30Z
> Policy -> Entity Mappings:
>    Policy: consoleAdmin
>      User Mappings:
>           CN=Toto Bou,OU=Users,OU=Blabla,OU=Global,DC=Blabla,DC=lan
>           cn=Toto Bou,ou=Users,ou=Blabla,ou=Global,dc=Blabla,dc=lan

instead of a message telling me that the account is already present.

another point I noticed, if I want added a name with an accent like,

`mc idp ldap policy attach mgdev consoleAdmin --user='CN=Clément Bou,OU=Users,OU=Blabla,OU=Global,DC=Blabla,DC=lan'`

I am left with:

`mc idp ldap policy entities ALIAS`

> Query time: 2024-04-17T12:58:30Z
> Policy -> Entity Mappings:
>     Policy: consoleAdmin
>       User Mappings:
>            cn=Cl\c3\a9ment Bou,ou=Users,ou=Blabla,ou=Global,dc=Blabla,dc=lan

Here I cannot connect with the Clément login.

## Possible Solution

it would be necessary to have the possibility of converting the existing one, to no longer see the problem and regain access to existing accounts.
and support the ability to have accents in names.

## Steps to Reproduce (for bugs)

1.Have functional LDAP auth
2.Have an account created with capital letters (so before the update, on a commr version RELEASE.2024-03-26T22-10-45Z I suppose)
3.Update minio to the latest version, you can no longer access the LDAP account
4.See bug description for commands

## Regression
 Yes

## Your Environment

[...]

PR #19591, opened 2024-04-23
fix: avoid some IAM import errors if LDAP enabled

donatello:
## Description

When LDAP is enabled, previously we were:

- rejecting creation of users and groups via the IAM import functionality

- throwing a `not a valid DN` error when non-LDAP group mappings are present

This change allows for these cases as we need to support situations where the MinIO server contains users, groups and policy mappings created before LDAP was enabled.

## Motivation and Context

Fixing remaining issues in IAM import (https://github.com/minio/minio/issues/19536) - follow up for #19528 and #19547.

## How to test this PR?

Tests are updated.

Create some built-in users/group/policy mappings before enabling LDAP. Then enable LDAP, export and import.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/admin-handlers-users.go
+++ b/cmd/admin-handlers-users.go
@@ -271,7 +271,14 @@ func (a adminAPIHandlers) UpdateGroupMembers(w http.ResponseWriter, r *http.Requ
 				return
 			}
 		}
-		updatedAt, err = globalIAMSys.AddUsersToGroup(ctx, updReq.Group, updReq.Members)
+
+		if globalIAMSys.LDAPConfig.Enabled() {
+			// We don't allow internal group manipulation in this API when LDAP
+			// is enabled for now.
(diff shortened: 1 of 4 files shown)


cmd/admin-handlers-users.go
```diff
@@ -2123,6 +2166,7 @@ func (a adminAPIHandlers) ImportIAM(w http.ResponseWriter, r *http.Request) {
 					return
 				}
 				if _, err = globalIAMSys.CreateUser(ctx, accessKey, ureq); err != nil {
+					fmt.Printf("CreateUser FAIL: %#v\n", err)
```
harshavardhana:
```suggestion
					fmt.Printf("CreateUser FAIL: %#v\n", err)
```

Please remove

harshavardhana (approved):

poornas (approved):

PR #19602, opened 2024-04-24
Validates PostgreSQL table name

ramondeklein:
The PostgreSQL store can be for submitting events to a table. This table can be configured, but it is inserted using `Sprintf(...)` to generate `INSERT`, `UPDATE`, `DELETE` statements. Despite it being a configuration value (and not originating from user input) it is still a good idea to validate or sanitize the value.

This PR checks the table name and only allows the use of valid PSQL identifiers. Note that [PostgreSQL identifier rules](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS) also allow diacritics as valid identifier names, but they are only accepted here when used with quotes. Creating a regular expression that accepts diacritics is hard and probably not used often.

### DIFF
--- a/internal/event/target/postgresql.go
+++ b/internal/event/target/postgresql.go
@@ -30,6 +30,7 @@ import (
 	"strconv"
 	"strings"
 	"time"
+	"unicode"
 
 	_ "github.com/lib/pq" // Register postgres driver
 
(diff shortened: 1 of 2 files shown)


internal/event/target/postgresql.go
```diff
@@ -444,3 +449,15 @@ func NewPostgreSQLTarget(id string, args PostgreSQLArgs, loggerOnce logger.LogOn
 
 	return target, nil
 }
+
+var errInvalidPsqlTablename = errors.New("invalid PostgreSQL table")
+
+func validatePsqlTableName(name string) error {
+	// check for simple name or quoted name
+	if match, err := regexp.MatchString("^(([a-zA-Z_][a-zA-Z0-9_$]*)|(\"[^\"]+\"))$", name); err != nil {
```
klauspost: I have only a vague idea what the regex does. Could you add a comment with the intent?

klauspost (commented):
A bit overly dramatic title. Unless something else is wrong this should not be available for anyone but admins to set.

I suggest "Sanitize postgres event table name" or something less "dramatic".

internal/event/target/postgresql.go
ramondeklein: It's in the line above :-) It simply checks for t

klauspost (approved):

ramondeklein:
> A bit overly dramatic title. Unless something else is wrong this should not be available for anyone but admins to set.
> I suggest "Sanitize postgres event table name" or something less "dramatic".

You're right. Updated commit message and PR title.

klauspost:
Good find. Could we imagine any existing table names becoming invalid now (other than the ones containing a quote)?

ramondeklein:
> Good find. Could we imagine any existing table names becoming invalid now (other than the ones containing a quote)?

If you're using a table name with diacritics, then they may be marked as invalid (where previously they were working fine). I do think that's a rare scenario and it can be easily fixed by using quotes (although that's always tricky on command-line and/or YAML).

I did check if I could find an existing function in one of the Go packages that would validate the name, but I couldn't find one. What I could do is first normalize the name using `golang.org/x/text/uniccode/norm` and then check it. Let me try that...

klauspost:
@ramondeklein Yeah. It will probably be pretty annoying to upgrade to this, if valid (even if strange) names would be rejected, since they would have to build a new table.

ramondeklein:
> @ramondeklein Yeah. It will probably be pretty annoying to upgrade to this, if valid (even if strange) names would be rejected, since they would have to build a new table.

They could also update the configuration to use the name with diacritics within double quotes, but I pushed a fix that also allows diacritics.

harshavardhana:
@ramondeklein please make branches on your fork, instead of the origin.

harshavardhana (approved):

ramondeklein:
> @ramondeklein please make branches on your fork, instead of the origin.

Will do that next time.

PR #19605, opened 2024-04-24
Add PutObject Ring Buffer

klauspost:
## Description

Replace the `io.Pipe` from streamingBitrotWriter -> CreateFile with a fixed size ring buffer.

This will add an output buffer for encoded shards to be written to disk - potentially via RPC.

This will remove blocking when `(*streamingBitrotWriter).Write` is called and it writes hashes and data.

With current settings the write looks like this:

```
Outbound
┌───────────────────┐             ┌────────────────┐               ┌───────────────┐                      ┌────────────────┐
│                   │   Parr.     │                │  (http body)  │               │                      │                │
│ Bitrot Hash       │     Write   │      Pipe      │      Read     │  HTTP buffer  │    Write (syscall)   │  TCP Buffer    │
│ Erasure Shard     │ ──────────► │  (unbuffered)  │ ────────────► │   (64K Max)   │ ───────────────────► │    (4MB)       │
│                   │             │                │               │  (io.Copy)    │                      │                │
└───────────────────┘             └────────────────┘               └───────────────┘                      └────────────────┘
```

We write a Hash (32 bytes). Since the pipe is unbuffered, it will block until the 32 bytes have been delivered to the TCP buffer and the next Read hits the Pipe. Then we  write the shard data. This will typically be bigger than 64KB, so it will block until 2 blocks have been read from the pipe.

When we insert a ring buffer:

```
Outbound
┌───────────────────┐             ┌────────────────┐               ┌───────────────┐                      ┌────────────────┐
│                   │             │                │  (http body)  │               │                      │                │
│ Bitrot Hash       │     Write   │  Ring Buffer   │      Read     │  HTTP buffer  │    Write (syscall)   │  TCP Buffer    │
│ Erasure Shard     │ ──────────► │    (2MB)       │ ────────────► │   (64K Max)   │ ───────────────────► │    (4MB)       │
│                   │             │                │               │  (io.Copy)    │                      │                │
└───────────────────┘             └────────────────┘               └───────────────┘                      └────────────────┘
```

The hash+shard will fit within the ring buffer, so writes will not block - but will complete after a memcopy. Reads will be able to fill the 64KB buffer if there is data for it.

If network is congested and the ring buffer will become filled, and all syscalls will be on full buffers. Only when the ring buffer is filled will erasure coding start blocking.

Since there is always "space" to write output data we remove the parallel writing, since we are always writing to memory now and the gorotine synchronization overhead probably isn't worth taking. If output was blocked in the existing, we would still wait for it to unblock in parallelWriter, so it would make no difference there - except now the ringbuffer smoothes out the load.

There are some microoptimizations we could look at later. The biggest is probably that in most cases we could encode directly to the ring buffer - if we are not at a boundary. Also "force filling" the Read requests (ie blocking until a full read can be completed) could be investigated and maybe allow concurrent memcopy on read and write. But if this isn't by itself better I don't think it is overall worth it.

The 2MB are a bit big for my liking - but we don't have 128K-256K buffers, which I would probably go for.

Pending: Actual testing.
For now also vendor https://github.com/smallnest/ringbuffer

## How to test this PR?

Test PutObject/Multipart with bigger files.

## Types of changes


klauspost:
(fixed...)

cmd/erasure-encode.go
```diff
@@ -21,44 +21,36 @@ import (
...
-					p.writers[i] = nil
-				}
-			} else {
+		var n int
+		n, p.errs[i] = p.writers[i].Write(blocks[i])
+		if p.errs[i] == nil {
+			if n != len(blocks[i]) {
+				p.errs[i] = io.ErrShortWrite
 				p.writers[i] = nil
 			}
```
harshavardhana, line 49: are you serializing this on purpose?

harshavardhana: @klauspost ^^

klauspost: We are either writing to a ringbuffer (to disk/net) or a bytes.Buffer (inline).

harshavardhana:
I am backlogged on testing this, will have to take internal help on this.

klauspost:
@harshavardhana No worries. It is not going anywhere.

PR #19647, opened 2024-05-01
support compression after rotation of logs

harshavardhana:

## Description
support compression after rotation of logs

## Motivation and Context
allow gzip compression of logs after
rotation continues on #19641

## How to test this PR?
```
~ minio server ~/test/ --log-dir /tmp/dir --log-size 1024 --log-compress
```

You shall see logs compressed after rotation beyond 1024 bytes per log.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/server-main.go
+++ b/cmd/server-main.go
@@ -205,6 +205,12 @@ var ServerFlags = []cli.Flag{
 		EnvVar: "MINIO_LOG_SIZE",
 		Hidden: true,
 	},
+	cli.BoolFlag{
+		Name:   "log-compress",
+		Usage:  "specify if we want the rotated logs to be gzip compressed or not",
+		EnvVar: "MINIO_LOG_COMPRESS",
+		Hidden: true,
+	},
 }
 
 var serverCmd = cli.Command{
@@ -682,6 +688,7 @@ func initializeLogRotate(ctx *cli.Context) (io.WriteCloser, error) {
 	output, err := logger.NewDir(logger.Options{
 		Directory:       lgDirAbs,
 		MaximumFileSize: int64(lgSize),
+		Compress:        ctx.Bool("log-compress"),
 	})
 	if err != nil {
 		return nil, err
--- a/internal/logger/logrotate.go
+++ b/internal/logger/logrotate.go
@@ -18,6 +18,8 @@
 package logger
 
 import (
+	"compress/gzip"
+	"encoding/json"
 	"fmt"
 	"io"
 	"os"
@@ -25,6 +27,7 @@ import (
 	"time"
 
 	xioutil "github.com/minio/minio/internal/ioutil"
+	"github.com/minio/pkg/v2/logger/message/log"
 )
 
 func defaultFilenameFunc() string {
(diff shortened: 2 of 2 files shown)


cmd/server-main.go
```diff
@@ -205,6 +205,12 @@ var ServerFlags = []cli.Flag{
 		EnvVar: "MINIO_LOG_SIZE",
 		Hidden: true,
 	},
+	cli.BoolFlag{
+		Name:   "log-compress",
+		Usage:  "specify if we want the rotated logs to be gzip compressed or not",
+		EnvVar: "MINIO_LOG_COMPRESS",
+		Hidden: true,
+	},
```
harshavardhana, line 213: hidden flags on purpose to allow them to standardize for a bit, we may potentially change it.

krisis: We can instead have a flag, say `log-compress-algo` a `cli.StringFlag`, which takes the compression algorithm as input, e.g `gzip` initially and more if we choose to support  other algorithms like `zstd`, `bz2`, etc?

internal/logger/logrotate.go
```diff
@@ -86,40 +92,106 @@ func (w *Writer) Close() error {
...
+
 func (w *Writer) listen() {
 	for {
 		var r io.Reader = w.pr
 		if w.opts.MaximumFileSize > 0 {
 			r = io.LimitReader(w.pr, w.opts.MaximumFileSize)
 		}
 		if _, err := io.Copy(w.f, r); err != nil {
-			fmt.Println("Failed to write to log file", err)
+			msg := fmt.Sprintf("unable to write to log file %v: %w", w.f.Name(), err)
```
krisis:
```suggestion
			msg := fmt.Sprintf("unable to write to log file %v: %v", w.f.Name(), err)
```
`%w` doesn't work well outside of `fmt.Errorf`. 
> From Go playground
> ./prog.go:9:14: fmt.Sprintf does not support error-wrapping directive %w

internal/logger/logrotate.go
```diff
@@ -86,40 +92,106 @@ func (w *Writer) Close() error {
...
+			stdErrEnc.Encode(&log.Entry{
+				Level:   ErrorKind,
+				Message: msg,
+				Time:    time.Now().UTC(),
+				Trace:   &log.Trace{Message: msg},
+			})
 		}
 		if err := w.rotate(); err != nil {
-			fmt.Println("Failed to rotate log file", err)
+			msg := fmt.Sprintf("unable to rotate log file %v: %w", w.f.Name(), err)
```
krisis: Same as above

internal/logger/logrotate.go
```diff
@@ -86,40 +92,106 @@ func (w *Writer) Close() error {
...
+		return err
+	}
+
+	var wc io.WriteCloser
+	wc = gzip.NewWriter(gw)
+	if _, err = io.Copy(wc, r); err != nil {
+		return err
+	}
+
+	// close everything before we delete.
```
krisis: There are couple of things missing here.
1. We should call `gzip.Writer`'s Close method before we call fsync on the `.gz` file.
2. `gzip.Writer.Close` method doesn't call `Close` on the underlying `io.WriteCloser`, it simply flushes any pending compressed data. We need to call `gw.Close` explicitly.

cmd/server-main.go
harshavardhana: No I want to keep it gzip and be done. It is not worth adding all others..

internal/logger/logrotate.go
harshavardhana: Oh yes it was renamed from fmt.Errorf()

krisis (approved):

poornas (approved):

PR #19650, opened 2024-05-02
Support user certificate based authentication on SFTP

olljanat:

## Description
Support user certificate based authentication on SFTP.

## Motivation and Context
Picked @klauspost comment from #17585 
> We don't want separate auth mechanism, but it should be possible to create (time limited) certs that link to a user.

SSH server setting `TrustedUserCAKeys` works in way that instead of defining authorized keys per user, server simply outsources that task for trusted certificate authority. This PR implements same logic to MinIO.

## How to test this PR?
One time operations:
1. Create trusted user CA `ssh-keygen -f user_ca`
2. Start MinIO with flag `--sftp=trusted-user-ca-key=/path/to/.ssh/user_ca.pub`

Per user operations:
1. Create user public key in client PC (testuser1 in this example) `ssh-keygen`
2. Provide copy of `/home/testuser1/.ssh/id_rsa.pub` to CA server.
3. Create signature for identity "testuser1" (must match with username in MinIO) which is valid one week `ssh-keygen -s /home/miniouser/.ssh/user_ca -I testuser-2024-05-03 -n testuser1 -V +1w id_rsa.pub`
4. Copy `id_rsa-cert.pub` to `/home/testuser1/.ssh/id_rsa-cert.pub` in client PC.
5. Connect to SFTP server `sftp -P <SFTP port> <server IP>`

After certificate expires, repeat steps 3 and 4.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/sftp-server.go
+++ b/cmd/sftp-server.go
@@ -18,6 +18,7 @@
 package cmd
 
 import (
+	"bytes"
 	"context"
 	"crypto/subtle"
 	"errors"
@@ -162,6 +163,7 @@ func startSFTPServer(args []string) {
 		port          int
 		publicIP      string
 		sshPrivateKey string
+		userCaKeyFile string
 	)
 	allowPubKeys := supportedPubKeyAuthAlgos
 	allowKexAlgos := preferredKexAlgos
@@ -197,6 +199,8 @@ func startSFTPServer(args []string) {
 			allowCiphers = filterAlgos(arg, strings.Split(tokens[1], ","), supportedCiphers)
 		case "mac-algos":
 			allowMACs = filterAlgos(arg, strings.Split(tokens[1], ","), supportedMACs)
+		case "trusted-user-ca-key":
+			userCaKeyFile = tokens[1]
 		}
 	}
 
@@ -276,6 +280,56 @@ func startSFTPServer(args []string) {
 			}
 			return nil, errAuthentication
 		},
+	}
+
+	if userCaKeyFile != "" {
+		keyBytes, err := os.ReadFile(userCaKeyFile)
+		if err != nil {
+			logger.Fatal(fmt.Errorf("invalid arguments passed, trusted user certificate authority public key file is not accessible: %v", err), "unable to start SFTP server")
+		}
+
+		caPublicKey, _, _, _, err := ssh.ParseAuthorizedKey(keyBytes)
+		if err != nil {
+			logger.Fatal(fmt.Errorf("invalid arguments passed, trusted user certificate authority public key file is not parseable: %v", err), "unable to start SFTP server")
+		}
+
+		sshConfig.PublicKeyCallback = func(c ssh.ConnMetadata, key ssh.PublicKey) (*ssh.Permissions, error) {
+			_, ok := globalIAMSys.GetUser(context.Background(), c.User())
+			if !ok {
+				return nil, errNoSuchUser
+			}
+
+			// Verify that client provided certificate, not only public key.
+			cert, ok := key.(*ssh.Certificate)
+			if !ok {
+				return nil, errSftpPublicKeyWithoutCert
+			}
+
+			// ssh.CheckCert called by ssh.Authenticate accepts certificates
+			// with empty principles list so we block those in here.
+			if len(cert.ValidPrincipals) == 0 {
+				return nil, errSftpCertWithoutPrincipals
+			}
+
+			// Verify that certificate provided by user is issued by trusted CA,
+			// username in authentication request matches to identities in certificate
+			// and that certificate type is correct.
+			checker := ssh.CertChecker{}
+			checker.IsUserAuthority = func(k ssh.PublicKey) bool {
+				return bytes.Equal(k.Marshal(), caPublicKey.Marshal())
+			}
+			_, err = checker.Authenticate(c, key)
+			if err != nil {
+				return nil, err
+			}
+
+			return &ssh.Permissions{
+				CriticalOptions: map[string]string{
+					"accessKey": c.User(),
+				},
+				Extensions: make(map[string]string),
+			}, nil
+		}
 	}
 
 	sshConfig.AddHostKey(private)
--- a/cmd/typed-errors.go
+++ b/cmd/typed-errors.go
@@ -119,3 +119,9 @@ var errInvalidMaxParts = errors.New("Part number is greater than the maximum all
 
 // error returned for session policies > 2048
 var errSessionPolicyTooLarge = errors.New("Session policy should not exceed 2048 characters")
+
+// error returned in SFTP when user used public key without certificate
+var errSftpPublicKeyWithoutCert = errors.New("public key authentication without certificate is not accepted")
+
+// error returned in SFTP when user used certificate which does not contain principal(s)
+var errSftpCertWithoutPrincipals = errors.New("certificates without principal(s) are not accepted")
--- a/docs/ftp/README.md
+++ b/docs/ftp/README.md
@@ -242,3 +242,16 @@ hmac-sha1
 hmac-sha1-96
 ```
 
+### Certificate-based authentication
+
+`--sftp=trusted-user-ca-key=...` specifies a file containing public key of certificate authority that is trusted
+to sign user certificates for authentication.
+
+Implementation is identical with "TrustedUserCAKeys" setting in OpenSSH server with exception that only one CA
+key can be defined.
+
+If a certificate is presented for authentication and has its signing CA key is in this file, then it may be
+used for authentication for any user listed in the certificate's principals list. 
+
+Note that certificates that lack a list of principals will not be permitted for authentication using trusted-user-ca-key.
+For more details on certificates, see the CERTIFICATES section in ssh-keygen(1).


cmd/sftp-server.go
```diff
@@ -218,6 +221,16 @@ func startSFTPServer(args []string) {
 		logger.Fatal(fmt.Errorf("invalid arguments passed, private key file is not parseable: %v", err), "unable to start SFTP server")
 	}
 
+	userCaPublicKeyBytes, err := os.ReadFile(userCaPublicKey)
```
klauspost: Shouldn't this be wrapped in an `if userCaPublicKey != "" {...}`?

cmd/sftp-server.go
```diff
@@ -227,6 +240,31 @@ func startSFTPServer(args []string) {
...
+			}
+			if cert.KeyId != c.User() {
+				return nil, errAuthentication
+			}
+			signed := *cert
+			signed.Signature = nil
+			signedData := signed.Marshal()
+			if err := caPublicKey.Verify(signedData[:len(signedData)-4], cert.Signature); err != nil {
+				return nil, errAuthentication
+			}
```
harshavardhana: As you know, some comments here will be needed to maintain this part of the code in the long term.

cmd/sftp-server.go
```diff
@@ -227,6 +240,31 @@ func startSFTPServer(args []string) {
 			MACs:         allowMACs,
 		},
 		PublicKeyAuthAlgorithms: allowPubKeys,
+		PublicKeyCallback: func(c ssh.ConnMetadata, key ssh.PublicKey) (*ssh.Permissions, error) {
+			_, ok := globalIAMSys.GetUser(context.Background(), c.User())
```
klauspost: I would expect `if caPublicKey == nil {return errAuthentication}` or similar.

klauspost (commented):
We probably need @aead to check this - at least I can't claim I understand what is going on.

harshavardhana (changes requested):

olljanat:
Good points, I was too focused to this feature so I totally forgot to check that it does not break old functionality. Now that is the case and I added some comments too.

> We probably need @aead to check this - at least I can't claim I understand what is going on.

Perhaps I can also explain this better. There is two parts on this:

a) Public key authentication with SSH server works in way that user generate private+public key pair with command `ssh-keygen` and add public key to file `/home/<username>/.ssh/authorized_keys` in target server.

That enables public key authentication with commands like `sftp username@servername`
How authentication actually happens then is well described example in [this picture](https://sebastien.saunier.me/images/posts/SSH%20Connection%20explained.png)


2) Challenge is handling those keys because admin needs to manually add those public keys for each user.
When `/home/testuser/.ssh/id_rsa-cert.pub` file is present in client side, it will send it to server instead of `/home/testuser/.ssh/id_rsa-cert.pub` and that gets dynamically allowed as long it is digitally signed by trusted certificate authority.

cmd/sftp-server.go
```diff
@@ -227,6 +243,44 @@ func startSFTPServer(args []string) {
 			MACs:         allowMACs,
 		},
 		PublicKeyAuthAlgorithms: allowPubKeys,
+		PublicKeyCallback: func(c ssh.ConnMetadata, key ssh.PublicKey) (*ssh.Permissions, error) {
+			if caPublicKey == nil {
+				return nil, errAuthentication
+			}
```
harshavardhana: This function must only be enabled if caPublicKey is != nil

cmd/sftp-server.go
```diff
@@ -227,6 +243,44 @@ func startSFTPServer(args []string) {
...
+
+			// Extract data structure without signature from authentication message
+			// and remove last 4 bytes which represent the length of the non-existent
+			// signature field which appears there as result of Marshal()
+			signed := *cert
+			signed.Signature = nil
+			signedData := signed.Marshal()
+			signedData2 := signedData[:len(signedData)-4]
```
harshavardhana: why is there no cleaner way to do this? is this the expected way to verify the incoming pubkey signature ?

harshavardhana (changes requested):

klauspost:
> Perhaps I can also explain this better. There is two parts on this:

Could you add that to [docs/ftp/README.md](https://github.com/minio/minio/pull/19650/files#diff-5fa9ac996654a93cdb322677493edc6525b4a72b3c0947846afe2431c6d62b48)?

harshavardhana (commented):
Instead of taking a single authorized key you must add provisions to take the entire authorized_keys file - otherwise this option is not that useful.

```go

    // Slice to store the authorized keys
    var authorizedKeys []ssh.PublicKey

    // Iterate over the authorized_keys file and parse each key
    for len(authorizedKeysBytes) > 0 {
        // Parse the next key
        pubKey, _, _, rest, err := ssh.ParseAuthorizedKey(authorizedKeysBytes)
        if err != nil {
            log.Fatal(err)
        }

        // Add the authorized key
        authorizedKeys = append(authorizedKeys, pubKey)

        // Update the authorizedKeysBytes slice to point to the next key
        authorizedKeysBytes = rest
    }
```

cmd/sftp-server.go
olljanat: Actually yes, there is [Authenticate](https://cs.opensource.google/go/x/crypto/+/refs/tags/v0.22.0:ssh/certs.go;l=358-381) function and documentation even says that it can be used with `PublicKeyCallback`

Now code most probably is more clear.

olljanat:
> Instead of taking a single authorized key you must add provisions to take the entire authorized_keys file - otherwise this option is not that useful.

Now you are mixing up certificate authority based trust to user based trust. This works already with multiple users in way how it is explained in example. 

This code can be later extended to support that model too but it would require some new way to store user = key mappings.
It is [this](https://cs.opensource.google/go/x/crypto/+/refs/tags/v0.22.0:ssh/certs.go;l=363-365) logic in Authenticate function which would handle those keys.

harshavardhana:
yeah but the option described now in this PR - not sure who will use it. I don't see any valid use-cases.

harshavardhana:
> * Create user CA `ssh-keygen -f /home/miniouser/.ssh/user_ca`
> * Create user public key in client PC (testuser in this example) `ssh-keygen`
> * Provide copy of `/home/testuser/.ssh/id_rsa.pub` to CA server.
> * Create signature for identity "testuser" (must match with username in MinIO) which is valid one week `ssh-keygen -s /home/miniouser/.ssh/user_ca -I testuser -V +1w id_rsa.pub`
> * Copy `id_rsa-cert.pub` to `/home/testuser/.ssh/id_rsa-cert.pub` in client PC.
> * Connect to SFTP server `sftp -P <SFTP port> <server IP>`

All this does is that it works with just one user, MinIO is not deployed for a single user system. Yes it is one of the use-cases but that is not going to satisfy our end users. 

The approach for ssh trust must be via authorized_keys not via providing a custom key that would expire, requiring a service restart. 

This is a heavy touch and cumbersome system.

olljanat:
> All this does is that it works with just one user, MinIO is not deployed for a single user system. Yes it is one of the use-cases but that is not going to satisfy our end users.

Not true. First step you do only once, other steps for as many user you want to support.  No need to restart MinIO when allowing new keys for new users.

> The approach for ssh trust must be via authorized_keys not via providing a custom key that would expire, requiring a service restart.

CA key will not ever expire, unless you separately define expiration for it. User keys you can also make be valid forever by just skipping `-V +1w` parameter.

> yeah but the option described now in this PR - not sure who will use it. I don't see any valid use-cases.

All SFTP clients which I have tested looks to be supporting this feature so I would expect it to be quite widely in use on enterprise environments. But sure, I can use user=authorized key mapping instead of too if you wish.

But where to store those keys? Case LDAP probably would be easiest because we can let user just define which field is used to store those keys but MinIO internal IDP would need more work.

olljanat:
Perhaps "user CA" is confusing term in here. I picked up it from https://dev.to/gvelrajan/how-to-configure-and-setup-ssh-certificates-for-ssh-authentication-b52 but for clarity, that is common certificate authority to for **ALL** user certificates, no matter how many users you have in system.

cmd/sftp-server.go
```diff
@@ -197,6 +199,8 @@ func startSFTPServer(args []string) {
 			allowCiphers = filterAlgos(arg, strings.Split(tokens[1], ","), supportedCiphers)
 		case "mac-algos":
 			allowMACs = filterAlgos(arg, strings.Split(tokens[1], ","), supportedMACs)
+		case "user-ca-public-key":
```
harshavardhana:
```suggestion
		case "trusted-user-ca-keys":
```

This is what it should be IMHO.

harshavardhana (commented):
Thanks for explaining we should call this what SSH calls it TrustedUserCAKeys - asked for the relevant changes.

olljanat:
> Thanks for explaining we should call this what SSH calls it TrustedUserCAKeys

Make sense, however current implementation supports only single CA so updated name and documentation matching to that. Supporting multiple would be of course also possible but it would add quite a lot of complexity so not sure if that is worth of effort. It is recommended to use dedicated CA for this purpose anyway.

cmd/sftp-server.go
```diff
@@ -278,6 +282,56 @@ func startSFTPServer(args []string) {
...
+			// ssh.CheckCert called by ssh.Authenticate accepts certificates
+			// with empty principles list so we block those in here.
+			if len(cert.ValidPrincipals) == 0 {
+				return nil, errSftpCertWithoutPrincipals
+			}
+
+			// Verify that certificate provided by user is issued by trusted CA,
+			// username in authentication request matches to identities in certificate
+			// and that certificate type is correct.
+			checker := ssh.CertChecker{}
```
harshavardhana, line 317: Ah this is exactly the helpers I was looking for.

Now it is clean.

harshavardhana (approved):

cmd/sftp-server.go
olljanat: @harshavardhana  👍  However looks that there was typo which why CI failed. Can you re-trigger it?

klauspost (approved):
LGTM

harshavardhana (approved):

sachinshakya507:
Starting the minio-server with  flag ```--sftp=trusted-user-ca-key=/path/to/.ssh/user_ca.pub``` 
I couldn't password login to sftp server.

olljanat:
@sachinshakya507 After this PR it was possible to use password login when this feature was enabled but looks that refactoring done in #19833 changed behavior.

As far I understand it is because now code is using same `authenticateSSHConnection` function for both password and public key authentications and that code in https://github.com/minio/minio/blob/f736702da8f3fe0967f9a25bab6c8d7e27d002a4/cmd/sftp-server.go#L164-L179 does not check if disablePassAuth is true (flag `--sftp="disable-password-auth=true"` enabled).

ping @zveinn

zveinn:
I'll check it out

zveinn:
@olljanat @sachinshakya507 
https://github.com/minio/minio/pull/20022
This PR will only check the caPublicKey is password if missing

PR #19651, opened 2024-05-02
Use better gzip for log rotate

klauspost:
## Description

Should be 2x faster with same usage.

## How to test this PR?

Enable log rotation

## Types of changes

### DIFF
--- a/internal/logger/logrotate.go
+++ b/internal/logger/logrotate.go
@@ -18,7 +18,6 @@
 package logger
 
 import (
-	"compress/gzip"
 	"encoding/json"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


internal/logger/logrotate.go
```diff
@@ -18,14 +18,14 @@
...
 import (
-	"compress/gzip"
 	"encoding/json"
 	"fmt"
 	"io"
 	"os"
 	"path/filepath"
 	"time"
 
+	"github.com/klauspost/compress/gzip"
```
harshavardhana, line 28: I know that guy.

harshavardhana (approved):

internal/logger/logrotate.go
klauspost: I know him. He is too lazy to add a zstd option, though ;)

PR #19669, opened 2024-05-04
 return appropriate error upon reaching maxClients()

harshavardhana:

## Description
return appropriate error upon reaching maxClients()

## Motivation and Context
return 429 HTTP status instead of 503s at maxClients()

## How to test this PR?
You can test out the functionality using Warp.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/handler-api.go
+++ b/cmd/handler-api.go
@@ -321,13 +321,21 @@ func maxClients(f http.HandlerFunc) http.HandlerFunc {
 			}
 		}
 
+		globalHTTPStats.addRequestsInQueue(1)
+		defer globalHTTPStats.addRequestsInQueue(-1)
+
 		pool, deadline := globalAPIConfig.getRequestsPool()
 		if pool == nil {
 			f.ServeHTTP(w, r)
 			return
 		}
 
-		globalHTTPStats.addRequestsInQueue(1)
+		// No deadline to wait, there is nothing to queue
+		// perform the API call immediately.
+		if deadline <= 0 {
+			f.ServeHTTP(w, r)
+			return
+		}
 
 		if tc, ok := r.Context().Value(mcontext.ContextTraceKey).(*mcontext.TraceCtxt); ok {
 			tc.FuncName = "s3.MaxClients"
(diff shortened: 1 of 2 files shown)


cmd/handler-api.go
```diff
@@ -154,13 +157,32 @@ func (t *apiConfig) init(cfg api.Config, setDriveCounts []int, legacy bool) {
...
+	}
+	if currentMax != apiRequestsMaxPerNode {
+		// requests per node are limited at 95% at once, with 5% of bursts allowed.
+		limit := rate.Limit(float64(apiRequestsMaxPerNode) * 0.95)
+		if limit == 0 {
+			// For some reason limit becomes zero we allow atleast
+			// 10 requests per node
+			limit = rate.Limit(10)
+		}
+		burst := apiRequestsMaxPerNode * (5 / 100)
```
poornas:
```suggestion
		burst := apiRequestsMaxPerNode * .05
```

klauspost:
This is fundamentally different and not the correct approach to limiting.

You are exchanging a *concurrency* limit with a *reqs/s* limit. This is will a) introduce needless latency for fast calls and b) allow long-running to bring down the system.

Consider HeadObject requests:

You are now placing an artificial limit on the throughput of these operations. This will mean that if a server has a limit of 200 and each take 2ms you can serve 200 * (1000/2) = 100K requests per second with the old. Now (with this) you can serve no more than 200 requests per second. 

Consider ListObjects:

If a ListObjects call takes 5s to complete on average, you can serve 200 * (1000/5000) = 40 requests per second. The max running will be 200 (since that is our limit). Switching to a rate limit means that you allow 200 new requests every second and after 30 seconds you will have more than 4000 requests running on your server.

(these assume fixed time storage - but for ListObjects including that would just mean you have even more listings running when the average response rate drops)

I am not sure what you mean by "the added cost of the last the caller is not given to the future caller". So I am not sure what problem you are trying to fix is.

If the problem you are solving is that you want requests to be served in order add a secondary wait queue.

harshavardhana:
> This is fundamentally different and not the correct approach to limiting.
> 
> You are exchanging a _concurrency_ limit with a _reqs/s_ limit. This is will a) introduce needless latency for fast calls and b) allow long-running to bring down the system.
> 
> Consider HeadObject requests:
> 
> You are now placing an artificial limit on the throughput of these operations. This will mean that if a server has a limit of 200 and each take 2ms you can serve 200 * (1000/2) = 100K requests per second with the old. Now (with this) you can serve no more than 200 requests per second.
> 
> Consider ListObjects:
> 
> If a ListObjects call takes 5s to complete on average, you can serve 200 * (1000/5000) = 40 requests per second. The max running will be 200 (since that is our limit). Switching to a rate limit means that you allow 200 new requests every second and after 30 seconds you will have more than 4000 requests running on your server.
> 
> (these assume fixed time storage - but for ListObjects including that would just mean you have even more listings running when the average response rate drops)
> 
> I am not sure what you mean by "the added cost of the last the caller is not given to the future caller". So I am not sure what problem you are trying to fix is.
> 
> If the problem you are solving is that you want requests to be served in order add a secondary wait queue.

yes, you are right. This only works if we do it per API call basis, which inwall is already doing. Doing this globally won't work; our channel-based implementation is sufficient here.

I will change this PR to change instead of responding back with 503 we must respond with 429 in-case of maxclients()

poornas (approved):

klauspost (approved):
LGTM

cmd/handler-api.go
```diff
@@ -321,13 +321,21 @@ func maxClients(f http.HandlerFunc) http.HandlerFunc {
 			}
 		}
 
+		globalHTTPStats.addRequestsInQueue(1)
```
vadmeste, line 324: This makes minio_s3_requests_waiting_total wrong because now this includes requests that are currently processed by the server, was this change only to make the code looks simpler ?

PR #19673, opened 2024-05-06
fix: Ignore AWSAccessKeyId check for SignV2 policy condition

### LINKED ISSUE #19652: Bug in postpolicyform.go signed uploads with POST request always fail with 403 error : must appear in the list of conditions.
When using a presigned_post with minio. Even though all fields and conditions are met we always get the 403 access denied error and thereby the upload fails.

## Expected Behavior
After using the boto3 presigned post request to upload a file it should succeed.

## Current Behavior
No matter what we change in conditions or fields to the generate_presigned_post boto3 call. The actual upload from browser fails.

## Possible Solution
It seems this check is too strict or has a bug in it in the file minio/cmd/postpolicyform.go:
```
	if len(checkHeader) != 0 {
		logKeys := make([]string, 0, len(checkHeader))
		for key := range checkHeader {
			logKeys = append(logKeys, key)
		}
		return fmt.Errorf("Each form field that you specify in a form (except %s) must appear in the list of conditions.", strings.Join(logKeys, ", "))
	}

	return nil

```

Somehow the checkHeader length > 0 but then there should be an entry printed. However the actual error returned is this (meaning there is no missing header key in the logKeys. Or maybe its an empty string in which case that is the bug):
```
Access Denied. (Each form field that you specify in a form (except Awsaccesskeyid, Signature) must appear in the list of conditions.)
```
After 'list of conditions.' We should see what field in the form causes this 403 access denied error to occur. But this is empty.

1. Generate a presigned POST url with boto3:
```
   conditions = [
        ['content-length-range', 1, 100000000],
        ['starts-with', '$content-disposition', ''],
        ['starts-with', '$content-type', ''],
        ['starts-with', '$content-length', ''],
        # trying all possible other headers here does not help
        # ['starts-with', '$user-agent', ''],
        # ['starts-with', '$host', ''],
        # ['starts-with', '$accept', ''],
        # ['starts-with', '$accept-encoding', ''],
        # ['starts-with', '$connection', ''],
        # ['starts-with', '$cookie', ''],
        # ['starts-with', '$origin', ''],
        # ['starts-with', '$referer', ''],
    ]

    # adding more fields here does not help to get rid of the 403 access denied
    # fields = {}
    
    try:
        s3_client = boto3.client(
            's3',
            aws_access_key_id=os.environ.get('S3_TOKEN'),
            aws_secret_access_key=os.environ.get('S3_SECRET'),
            endpoint_url=os.environ.get("S3_ENDPOINT")
            # aws_session_token=SESSION_TOKEN
        )
        response = s3_client.generate_presigned_post(Bucket=os.environ.get("S3_BUCKET"),
                                                     Key=object_name,
                                                     Conditions=conditions,
                                                     Fields=fields,
                                                     ExpiresIn=expiration)
```

We get back our response['fields'] and use this to make a POST request with browser to upload a file using the fields returned from the generate_presigned_post call. Example this form data:
```
key: ce042307309a48d98f92b0149f542e8d.png
AWSAccessKeyId: xGOLjuk2Haq4zS6vGzpr
policy: eyJleHBpcmF0aW9uIjogIjIwMjQtMDUtMDJUMTI6MDY6MTFaIiwgImNvbmRpdGlvbnMiOiBbWyJjb250ZW50LWxlbmd0aC1yYW5nZSIsIDEsIDEwMDAwMDAwMF0sIFsic3RhcnRzLXdpdGgiLCAiJGNvbnRlbnQtZGlzcG9zaXRpb24iLCAiIl0sIFsic3RhcnRzLXdpdGgiLCAiJGNvbnRlbnQtdHlwZSIsICIiXSwgWyJzdGFydHMtd2l0aCIsICIkY29udGVudC1sZW5ndGgiLCAiIl0sIFsic3RhcnRzLXdpdGgiLCAiJHVzZXItYWdlbnQiLCAiIl0sIFsic3RhcnRzLXdpdGgiLCAiJGhvc3QiLCAiIl0sIFsic3RhcnRzLXdpdGgiLCAiJGFjY2VwdCIsICIiXSwgWyJzdGFydHMtd2l0aCIsICIkYWNjZXB0LWVuY29kaW5nIiwgIiJdLCBbInN0YXJ0cy13aXRoIiwgIiRjb25uZWN0aW9uIiwgIiJdLCBbInN0YXJ0cy13aXRoIiwgIiRjb29raWUiLCAiIl0sIFsic3RhcnRzLXdpdGgiLCAiJG9yaWdpbiIsICIiXSwgWyJzdGFydHMtd2l0aCIsICIkcmVmZXJlciIsICIiXSwgeyJidWNrZXQiOiAiZ2l2ZS1yZWZzZXQtcGhvdG9zIn0sIHsia2V5IjogImNlMDQyMzA3MzA5YTQ4ZDk4ZjkyYjAxNDlmNTQyZThkLnBuZyJ9XX0=
signature: 7q6WyPfc6rld2bs9CuVR/s8W6XU=
file: (binary)
```

This used to work fine. With current version of minio we always get a 403 error complaining about form field must appear in list of conditions. But whatever we change to our condition list this error is always happening.

## Context
Uploading with presigned urls to minio is completely broken for us. (Same code and strategy works fine with other servers like wasabi and amazon s3 itself).

## Regression
This is a regression as it worked  in the beginning of 2023 but somewhere the more stricter policy checks were applied to the point that now it never is allowed anymore regardless of the policy or fields configuration.

## Your Environment
* Version used (`minio --version`):
```
/bin/minio --version
minio version RELEASE.2024-03-21T23-13-43Z (commit-id=7fd76dbbb71eeba0dd1d7c16e7d96ec1a9deba52)
Runtime: go1.21.8 linux/amd64
```
* Server setup and configuration: Docker-compose file on macos
```
Linux 532fc3fa91ed 6.4.16-linuxkit #1 SMP PREEMPT_DYNAMIC Thu Nov 16 10:55:59 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux
```

jiuker:

## Description

fix: Ignore AWSAccessKeyId check for SignV2 policy condition

## How to test this PR?

UT should cover this.

## Types of changes
- [x] Bug fix #19652

## Checklist:

### DIFF
--- a/cmd/post-policy_test.go
+++ b/cmd/post-policy_test.go
@@ -210,18 +210,23 @@ func testPostPolicyBucketHandler(obj ObjectLayer, instanceType string, t TestErr
 	// Test cases for signature-V2.
 	testCasesV2 := []struct {
 		expectedStatus int
-		accessKey      string
 		secretKey      string
+		formData       map[string]string
 	}{
-		{http.StatusForbidden, "invalidaccesskey", credentials.SecretKey},
-		{http.StatusForbidden, credentials.AccessKey, "invalidsecretkey"},
-		{http.StatusNoContent, credentials.AccessKey, credentials.SecretKey},
+		{http.StatusForbidden, credentials.SecretKey, map[string]string{"AWSAccessKeyId": "invalidaccesskey"}},
+		{http.StatusForbidden, "invalidsecretkey", map[string]string{"AWSAccessKeyId": credentials.AccessKey}},
+		{http.StatusNoContent, credentials.SecretKey, map[string]string{"AWSAccessKeyId": credentials.AccessKey}},
+		{http.StatusForbidden, credentials.SecretKey, map[string]string{"Awsaccesskeyid": "invalidaccesskey"}},
+		{http.StatusForbidden, "invalidsecretkey", map[string]string{"Awsaccesskeyid": credentials.AccessKey}},
+		{http.StatusNoContent, credentials.SecretKey, map[string]string{"Awsaccesskeyid": credentials.AccessKey}},
+		// Forbidden with key not in policy.conditions for signed requests V2.
+		{http.StatusForbidden, credentials.SecretKey, map[string]string{"Awsaccesskeyid": credentials.AccessKey, "AnotherKey": "AnotherContent"}},
 	}
 
 	for i, test := range testCasesV2 {
 		// initialize HTTP NewRecorder, this records any mutations to response writer inside the handler.
 		rec := httptest.NewRecorder()
-		req, perr := newPostRequestV2("", bucketName, "testobject", test.accessKey, test.secretKey)
+		req, perr := newPostRequestV2("", bucketName, "testobject", test.secretKey, test.formData)
 		if perr != nil {
 			t.Fatalf("Test %d: %s: Failed to create HTTP request for PostPolicyHandler: <ERROR> %v", i+1, instanceType, perr)
 		}
@@ -593,7 +598,7 @@ func postPresignSignatureV4(policyBase64 string, t time.Time, secretAccessKey, l
 	return signature
 }
 
-func newPostRequestV2(endPoint, bucketName, objectName string, accessKey, secretKey string) (*http.Request, error) {
+func newPostRequestV2(endPoint, bucketName, objectName string, secretKey string, formInputData map[string]string) (*http.Request, error) {
 	// Expire the request five minutes from now.
 	expirationTime := UTCNow().Add(time.Minute * 5)
 	// Create a new post policy.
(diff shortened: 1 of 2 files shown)


harshavardhana (commented):
Please add a test case that covers these usecases..

cmd/post-policy_test.go
```diff
@@ -593,7 +598,7 @@ func postPresignSignatureV4(policyBase64 string, t time.Time, secretAccessKey, l
 	return signature
 }
 
-func newPostRequestV2(endPoint, bucketName, objectName string, accessKey, secretKey string) (*http.Request, error) {
+func newPostRequestV2(endPoint, bucketName, objectName string, secretKey string, overWriteFormData map[string]string) (*http.Request, error) {
```
shtripat:
```suggestion
func newPostRequestV2(endPoint, bucketName, objectName string, secretKey string, formInputData map[string]string) (*http.Request, error) {
```

The name `overWriteFormData` looks more like boolean field name

shtripat (approved):
lgtm

harshavardhana (approved):

PR #19677, opened 2024-05-06
Make WalkDir return errors

klauspost:
## Description

This adds errors to WalkDir responses. This means that batch jobs now has an idea of whether they have failed prematurely. This is because if WalkDir terminates it could only close the channel, so it was impossible to detect if there were no more entries or an error occured.

I tried to adapt retry mechanisms reasonably - meaning what I could do without a bigger rewrite.

We now stop updating `RetryAttempts` since some of them were racy. Instead we add "Attempts" which counts total attempts.

We also stop counting failed retries as multiple errors. And if a retry succeeds it will deduct from the failures.

Please review carefully as I am not too deep into these systems.

## How to test this PR?

Batch jobs with unstable network I guess.

## Types of changes

### DIFF
--- a/cmd/batch-handlers.go
+++ b/cmd/batch-handlers.go
@@ -1517,10 +1517,8 @@ func (a adminAPIHandlers) ListBatchJobs(w http.ResponseWriter, r *http.Request) 
 	listResult := madmin.ListBatchJobsResult{}
 	for result := range resultCh {
 		if result.Err != nil {
-			// TODO: We will not receive any more results.
-			// We should probably fail here so client retries.
 			batchLogIf(ctx, result.Err)
-			continue
+			return
 		}
 		req := &BatchJobRequest{}
 		if err := req.load(ctx, objectAPI, result.Item.Name); err != nil {
--- a/cmd/batch-handlers.go
+++ b/cmd/batch-handlers.go
@@ -447,7 +447,7 @@ func (r *BatchJobReplicateV1) StartFromSource(ctx context.Context, api ObjectLay
 				} else {
 					stopFn(oi, nil)
 				}
-				ri.trackCurrentBucketObject(r.Target.Bucket, oi, success)
+				ri.trackCurrentBucketObject(r.Target.Bucket, oi, success, attempts)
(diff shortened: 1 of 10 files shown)


cmd/batch-handlers.go
```diff
@@ -1496,8 +1516,12 @@ func (a adminAPIHandlers) ListBatchJobs(w http.ResponseWriter, r *http.Request)
 
 	listResult := madmin.ListBatchJobsResult{}
 	for result := range resultCh {
+		if result.Err != nil {
+			batchLogIf(ctx, result.Err)
```
vadmeste: this is an admin handler, let's also call writeErrorResponseJSON

cmd/batch-handlers.go
```diff
@@ -833,21 +834,35 @@ func (ri *batchJobInfo) clone() *batchJobInfo {
...
+		Attempts:         ri.Attempts,
 	}
 }
 
-func (ri *batchJobInfo) countItem(size int64, dmarker, success bool) {
+func (ri *batchJobInfo) countItem(size int64, dmarker, success bool, attempt int) {
 	if ri == nil {
 		return
 	}
+	ri.Attempts++
```
vadmeste, line 845: ri.Attempts does not seem to be used here, are we trying to send this information to the user in the future ? also it seems to me it is more useful if we do this:

```
if attempt > 0 {
ri.Attempts++
}
```

vadmeste (commented):
Few comments

cmd/batch-handlers.go
klauspost: @vadmeste Yeah. Had added it. Took it out, so it would return whatever was collected. But I think you are right.

cmd/batch-handlers.go
klauspost: Attempt should always be >0, if not it is a mistake by the caller. 

Yes. It is just to log the total number of attempts.

poornas (approved):
LGTM

harshavardhana (approved):

PR #19679, opened 2024-05-06
Add batch status API

vadmeste:

## Description
Currently the status of a completed or failed batch is held in the memory, a simple restart will lose the status and the user will not have any visibility of the job that was long running.

In addition to the metrics, add a new API that reads the batch status from the drives. A batch job will be cleaned up three days after completion.

Also add the batch type in the batch id, the reason is that the batch job request is removed immediately when the job is finished, then we do not know the type of batch job anymore, hence a difficulty to locate the job report

## Types of changes

## Checklist:

### DIFF
--- a/cmd/batch-handlers.go
+++ b/cmd/batch-handlers.go
@@ -28,6 +28,7 @@ import (
 	"math/rand"
 	"net/http"
 	"net/url"
+	"path/filepath"
 	"runtime"
 	"strconv"
 	"strings"
@@ -56,6 +57,11 @@ import (
 )
 
 var globalBatchConfig batch.Config
+
+const (
+	// Keep the completed/failed job stats 3 days before removing it
+	oldJobsExpiration = 3 * 24 * time.Hour
+)
 
 // BatchJobRequest this is an internal data structure not for external consumption.
 type BatchJobRequest struct {
@@ -262,7 +268,7 @@ func (r *BatchJobReplicateV1) StartFromSource(ctx context.Context, api ObjectLay
 		JobType:   string(job.Type()),
 		StartTime: job.Started,
 	}
-	if err := ri.load(ctx, api, job); err != nil {
+	if err := ri.loadOrInit(ctx, api, job); err != nil {
 		return err
 	}
 	if ri.Complete {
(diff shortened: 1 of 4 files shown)


klauspost:
#19677 also related.

cmd/batch-handlers.go
```diff
@@ -57,6 +58,11 @@ import (
 
 var globalBatchConfig batch.Config
 
+const (
+	// Keep the completed/falied job stats 3 days before removing it
+	oldJobsExpiration = 3 * 24 * time.Hour
```
harshavardhana, line 63: 3 days? 24hr is enough I think.

harshavardhana (changes requested):

cmd/batch-handlers.go
```diff
@@ -57,6 +58,11 @@ import (
 
 var globalBatchConfig batch.Config
 
+const (
+	// Keep the completed/falied job stats 3 days before removing it
```
shtripat:
```suggestion
	// Keep the completed/failed job stats 3 days before removing it
```

cmd/batch-handlers.go
```diff
@@ -722,60 +728,83 @@ const (
...
+	case batchExpireName:
 		version = batchExpireVersionV1
 		format = batchExpireFormat
 	default:
 		return errors.New("no supported batch job request specified")
 	}
-	data, err := readConfig(ctx, api, getJobReportPath(job))
+
+	data, err := readConfig(ctx, api, path)
+	fmt.Println(path, err)
```
shtripat: remove debug print

vadmeste: good catch!

cmd/batch-handlers.go
vadmeste: @harshavardhana the batch job is a long running job, I doubt users will check the batch status everyday, then once the job batch completes or fails + 24 hours, there is no trace at all about what happened. Maybe we should not remove in the first place and add a new API to remove batch reports

harshavardhana: Yeah for that they must configure notifications, we shouldn't have to carry this around.

Can you check what AWS S3 does?

shtripat:
Please resolve the conflicts

cmd/batch-handlers.go
harshavardhana: did you verify? @vadmeste

harshavardhana:
PTAL at the conflicts.

harshavardhana:
rebase the PR please @vadmeste

cmd/batch-handlers.go
vadmeste: @harshavardhana as far as I found, users configure the way to store the result of the job batch, like saving the report in an s3 bucket, but there is no automatic cleaning here

harshavardhana: yeah we can provide such an option as well @vadmeste keeping this around for 3 days for now is okay.

harshavardhana:
> rebase the PR please @vadmeste

ping, please rebase the PR @vadmeste

harshavardhana (approved):

shtripat (approved):
LGTM

PR #19681, opened 2024-05-06
Support 'If-None-Match: *' for PUT case

justincormack:

## Description

The code for If-None-Match works if the etag is specified, but actually the common user case for PUT is to use '*' as the match, which will match any etag, ie will PUT the resource only if there is currently no resource at the location. This is used for simply making sure no resource is overwritten, and is the usual mechanism for handling concurrent updates on object stores. This works elsewhere, eg CLoudflare R2.

Note there are other cases where '*' is allowed, eg in If-match and the copy operations, but these are very rarely used.

## Motivation and Context

Compatibility with other object stores that support conditional updates.

## How to test this PR?

PUT a new object with 'If-None-Match: *' header twice, first time should succeed, second time should fail.

## Types of changes

## Checklist:


harshavardhana:
This is not enough, sending a more comprehensive PR.

PR #19687, opened 2024-05-07
allow caller context during reloads() to cancel

harshavardhana:

## Description
allow caller context during reloads() to cancel

## Motivation and Context

Canceled callers might linger around longer
and can potentially overwhelm the system. Instead,
provider a caller context, and canceled callers
don't hold on to them.
    
Bonus: we have no reason to cache errors; we should
never cache errors otherwise, we can potentially have
quorum errors creeping in unexpectedly. We should
let the cache, when invalidating, hit the actual resources
instead.

## How to test this PR?
Add faster client timeouts, retries
and observe the goroutine build-up on the server.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/admin-handlers-users.go
+++ b/cmd/admin-handlers-users.go
@@ -1217,9 +1217,9 @@ func (a adminAPIHandlers) AccountInfoHandler(w http.ResponseWriter, r *http.Requ
 	}
 
 	bucketStorageCache.InitOnce(10*time.Second,
-		cachevalue.Opts{ReturnLastGood: true, NoWait: true},
-		func() (DataUsageInfo, error) {
-			ctx, done := context.WithTimeout(context.Background(), 2*time.Second)
+		cachevalue.Opts{ReturnLastGood: true},
+		func(ctx context.Context) (DataUsageInfo, error) {
+			ctx, done := context.WithTimeout(ctx, 2*time.Second)
 			defer done()
 
 			return loadDataUsageFromBackend(ctx, objectAPI)
--- a/cmd/bucket-quota.go
+++ b/cmd/bucket-quota.go
@@ -49,8 +49,8 @@ var bucketStorageCache = cachevalue.New[DataUsageInfo]()
 func (sys *BucketQuotaSys) Init(objAPI ObjectLayer) {
 	bucketStorageCache.InitOnce(10*time.Second,
 		cachevalue.Opts{ReturnLastGood: true, NoWait: true},
-		func() (DataUsageInfo, error) {
-			ctx, done := context.WithTimeout(context.Background(), 2*time.Second)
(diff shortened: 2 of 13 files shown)


vadmeste (commented):
Compilation issue for _test.go but other LGTM

klauspost (commented):
AFAICT This would cause a call with a canceled context to stick for the TTL of the cache.

That seems to me like it would de-stabilize running systems. Caller contexts can be canceled at any time for any reason.

I see it would be nice if a canceled context would make the caller return, but it shouldn't be coupled to the validity of the cached value.

klauspost (changes requested):
Much simpler is just to add...
```Go
// GetCtx will return a cached value or fetch a new one.
// The fetch can be cancelled by the context.
func (t *Cache[T]) GetCtx(ctx context.Context) (T, error) {
	var resErr error
	var resVal T
	getDone := make(chan T)
	go func() {
		defer close(getDone)
		resVal, resErr = t.Get()
	}()
	select {
	case <-ctx.Done():
		var t T
		return t, ctx.Err()
	case <-getDone:
		return resVal, resErr
	}
}
```

harshavardhana:
> AFAICT This would cause a call with a canceled context to stick for the TTL of the cache.
> 
> That seems to me like it would de-stabilize running systems. Caller contexts can be canceled at any time for any reason.
> 
> I see it would be nice if a canceled context would make the caller return, but it shouldn't be coupled to the validity of the cached value.

how? the context is only honored for the update() operation, otherwise we will be returning cached values. 

Unlike previous behavior, where we incorrectly passed a cached context, we are not caching the context anymore, so I don't see how this is destabilizing in any manner. 

Remember, I am still not passing the passed context when `NoWait: true` to avoid passing a possible canceled context into a goroutine. 

Instead, it is still passing context.Background() to update the value in a non-blocking manner. However, for the places where `NoWait: false` we apply caller context to ensure that
The resources will be relinquished right away if the caller is gone.

makes sense?

harshavardhana:
you can always pass context.Background() to have this fully blocking and governed via any specific timeout each cached value might have.

klauspost:
My main concern is that the update function should be invariant to the caller context. I will review tomorrow.

harshavardhana:
> My main concern is that the update function should be invariant to the caller context. I will review tomorrow.

I also removed caching of errors, I saw a case where DiskInfo() will keep returning context.Deadline error which is not expected since we want the call to be made fresh upon such errors.

An update() error must be returned unless caller asks for ReportLastGood copy.

harshavardhana:
PTAL @klauspost @vadmeste

vadmeste (approved):
LGTM

poornas (approved):

PR #19699, opened 2024-05-08
Fix incorrect merging of slash-suffixed objects

klauspost:
## Description
If two objects share everything but one object has a slash prefix those would be merged in listings, with secondary properties used to tiebreak.

Example An object with key `prefix/obj` would be merged with an object named `prefix/obj/`. While this violates the [no object can be prefix of another](https://min.io/docs/minio/linux/operations/concepts/thresholds.html#conflicting-objects) let's resolve these.

If we have an object with 'name' and a directory named 'name/' discard the directory only - but allow objects of 'name' and 'name/' (xldir) to be uniquely returned.

Regression from https://github.com/minio/minio/pull/15772

## How to test this PR?

See example. Requires objects to be placed on different sets.

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)

PR #19717, opened 2024-05-10
Add regression test for #19699

klauspost:
## Description

Adds regression test for #19699

Failures are a bit luck based, since it requires objects to be placed on different sets.

However this generates a failure prior to #19699 on distributed.

## How to test this PR?

By running tests.

## Types of changes


harshavardhana:
The original PR I had to revert due to unexpected test failures, can you add that back into this PR?

PR #19719, opened 2024-05-10
Add usage cache cleanup and lower forced top compaction

klauspost:
## Description

Lower forced compaction to 250K entries.

If there is more than 250K entries on the top level force compact it and log an error. (I don't really expect this to be hit. It should be compacted upstream, but as a double safety)

## How to test this PR?

Dump a huge amount of objects in the root of a bucket.

## Types of changes


harshavardhana (approved):

nikpivkin:
@klauspost Hi! Just wondering how the number of entries is chosen?

klauspost:
@nikpivkin it's just a number. This produces a 4mb cache with short names. So seems reasonable.

PR #19741, opened 2024-05-15
fix: add deadlines for all synchronous REST callers

harshavardhana:

## Description
fix: add deadlines for all synchronous REST callers

## Motivation and Context
add deadlines that can be dynamically changed via
the drive max timeout values.

Bonus: optimize "file not found" case and hung drives/network -
circuit, break the check, and return right away instead of waiting.

## How to test this PR?
It allows for the quick detection of hung nodes for all blocking operations.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/erasure-object.go
+++ b/cmd/erasure-object.go
@@ -899,6 +899,20 @@ func (er erasureObjects) getObjectFileInfo(ctx context.Context, bucket, object s
 		if success {
 			validResp++
 		}
+
+		if totalResp >= minDisks && opts.FastGetObjInfo {
+			rw.Lock()
+			ok := countErrs(errs, errFileNotFound) >= minDisks || countErrs(errs, errFileVersionNotFound) >= minDisks
+			rw.Unlock()
+			if ok {
+				err = errFileNotFound
+				if opts.VersionID != "" {
+					err = errFileVersionNotFound
+				}
+				break
+			}
+		}
+
 		if totalResp < er.setDriveCount {
 			if !opts.FastGetObjInfo {
 				continue
--- a/cmd/storage-rest-client.go
+++ b/cmd/storage-rest-client.go
@@ -391,6 +391,9 @@ func (client *storageRESTClient) CreateFile(ctx context.Context, origvolume, vol
 }
 
 func (client *storageRESTClient) WriteMetadata(ctx context.Context, origvolume, volume, path string, fi FileInfo) error {
+	ctx, cancel := context.WithTimeout(ctx, globalDriveConfig.GetMaxTimeout())
+	defer cancel()
+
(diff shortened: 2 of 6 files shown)


cmd/erasure-object.go
```diff
@@ -899,6 +899,20 @@ func (er erasureObjects) getObjectFileInfo(ctx context.Context, bucket, object s
...
+			ok := countErrs(errs, errFileNotFound) >= minDisks || countErrs(errs, errFileVersionNotFound) >= minDisks
+			rw.Unlock()
+			if ok {
+				err = errFileNotFound
+				if opts.VersionID != "" {
+					err = errFileVersionNotFound
+				}
+				break
+			}
+		}
```
harshavardhana, line 914: The idea is that the return file is not found faster; this is better for `Head` like calls.

internal/config/drive/drive.go
```diff
@@ -25,6 +25,10 @@ import (
 	"github.com/minio/pkg/v2/env"
 )
 
+const (
+	EnvMaxDriveTimeout = "MINIO_DRIVE_MAX_TIMEOUT"
```
klauspost: What does the "MAX" refer to? Maybe replace with `OP` or `OPERATTION`  now you have to change it for the linter?

klauspost (commented):
LGTM

cmd/storage-rest-client.go
```diff
@@ -391,6 +391,9 @@ func (client *storageRESTClient) CreateFile(ctx context.Context, origvolume, vol
 }
 
 func (client *storageRESTClient) WriteMetadata(ctx context.Context, origvolume, volume, path string, fi FileInfo) error {
+	ctx, cancel := context.WithTimeout(ctx, globalDriveConfig.GetMaxTimeout())
```
vadmeste, line 394: isn't ping/pong in the grid code enough ? probably we should fix something in the grid code instead

klauspost: @vadmeste Timeouts are strict timeouts and are forwarded to the remote. So both the caller and the remote will hit this timeout.

Ping will keep the request alive forever even with no activity - and are only available on streams.
Same with context cancelation - only for streams - Streams will propagate cancelation.

One-shot requests are assumed to be short lived - and have a default deadline of 1 minute, unless something else is set on the context. Cancelling the context will cancel the local, but the remote will have to hit the timeout.

harshavardhana: We cannot rely on pings. Active callers must timeout.

internal/config/drive/drive.go
harshavardhana: I will send a separate PR for it

krisis (approved):
LGTM

PR #19743, opened 2024-05-15
Tweak one way stream ping

klauspost:
## Description

Do not log errors on oneway streams when sending ping fails. Instead cancel the stream.

This also makes sure pings are sent when blocked on sending responses.

I will do a separate PR that includes this and adds pings to two-way streams as well as tests for pings.

## How to test this PR?

This is double safety for abandoned streams.

## Types of changes


vadmeste (approved):
LGTM

harshavardhana (approved):

PR #19748, opened 2024-05-15
Add Veeam storage class override

klauspost:
## Description

Recent Veeam is very picky about storage class names. Add `_MINIO_VEEAM_FORCE_SC` env var.

It will override the storage class returned by the storage backend if it is non-standard and we detect a Veeam client by checking the User Agent.

Applies to HeadObject/GetObject/ListObject*

## How to test this PR?

Veeam latest version, with a tiered setup. Wait for data to be tiered.

## Types of changes


krisis (approved):
LGTM

harshavardhana (approved):

PR #19756, opened 2024-05-16
Tweak multipart uploads

klauspost:
## Description

Fix minor issues when reviewing  ListMultipartUploads:

* Actually store modtime in upload id, return when listing instead of current time.
* Use modtime for expiring and skip reading the file info.
* Consistent upload sorting in listing (since it now actually has the modtime).
* Try to exclude healing disks.

## How to test this PR?

Sending a mint test to minio-go.

## Types of changes

### DIFF
--- a/cmd/erasure-multipart.go
+++ b/cmd/erasure-multipart.go
@@ -200,11 +200,13 @@ func (er erasureObjects) cleanupStaleUploadsOnDisk(ctx context.Context, disk Sto
 		readDirFn(pathJoin(drivePath, minioMetaMultipartBucket, shaDir), func(uploadIDDir string, typ os.FileMode) error {
 			uploadIDPath := pathJoin(shaDir, uploadIDDir)
 			var modTime time.Time
-			// Upload IDs are of the form <UUID>x<UnixNano>, we can extract the time from the UUID.
-			if split := strings.Split(uploadIDDir, "x"); len(split) == 2 {
-				t, err := strconv.ParseInt(split[1], 10, 64)
-				if err == nil {
-					modTime = time.Unix(0, t)
+			// Upload IDs are of the form base64_url(<UUID>x<UnixNano>), we can extract the time from the UUID.
+			if b64, err := base64.RawURLEncoding.DecodeString(uploadIDDir); err == nil {
+				if split := strings.Split(string(b64), "x"); len(split) == 2 {
+					t, err := strconv.ParseInt(split[1], 10, 64)
+					if err == nil {
+						modTime = time.Unix(0, t)
+					}
 				}
 			}
 			// Fallback for older uploads without time in the ID.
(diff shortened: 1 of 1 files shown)


cmd/erasure-multipart.go
```diff
@@ -199,38 +199,55 @@ func (er erasureObjects) cleanupStaleUploadsOnDisk(ctx context.Context, disk Sto
...
-			fi, err := disk.ReadVersion(ctx, "", minioMetaMultipartBucket, uploadIDPath, "", ReadOptions{})
-			if err != nil {
+			var modTime time.Time
+			// Upload IDs are of the form <UUID>x<UnixNano>, we can extract the time from the UUID.
+			if split := strings.Split(uploadIDDir, "x"); len(split) == 2 {
+				t, err := strconv.ParseInt(split[1], 10, 64)
+				if err == nil {
+					modTime = time.Unix(0, t)
+				}
+			}
```
poornas: uploadID is base64 encoded string - need to decode first

poornas (changes requested):

cmd/erasure-multipart.go
klauspost: Good spot!

cmd/erasure-multipart.go
```diff
@@ -199,38 +199,57 @@ func (er erasureObjects) cleanupStaleUploadsOnDisk(ctx context.Context, disk Sto
 	readDirFn(pathJoin(drivePath, minioMetaMultipartBucket), func(shaDir string, typ os.FileMode) error {
 		readDirFn(pathJoin(drivePath, minioMetaMultipartBucket, shaDir), func(uploadIDDir string, typ os.FileMode) error {
 			uploadIDPath := pathJoin(shaDir, uploadIDDir)
-			fi, err := disk.ReadVersion(ctx, "", minioMetaMultipartBucket, uploadIDPath, "", ReadOptions{})
-			if err != nil {
+			var modTime time.Time
+			// Upload IDs are of the form base64_url(<UUID>x<UnixNano>), we can extract the time from the UUID.
+			if b64, err := base64.RawURLEncoding.DecodeString(uploadIDDir); err == nil {
+				if split := strings.Split(string(b64), "x"); len(split) == 2 {
```
vadmeste, line 205: will 'x' never confuse the base64 decoder ? because x can exist in a base64 output

harshavardhana: @klauspost we need to perhaps add the right "base64" key length to be split instead of `x<UnixNano>` in this case key length for base64 must be constant

klauspost: @vadmeste `x` cannot be present in a UUID. The split is on the *decoded* upload id, not the encoded data.

@harshavardhana Why? It is just an ID.

harshavardhana (approved):

poornas (approved):

PR #19766, opened 2024-05-17
Return NumVersions in quorum when available

krisis:

## Description
This can happen when the ObjectInfo value constructed from a quorum number of drives had its NumVersions field (incorrectly) set to zero. This is possible, however improbable, when the ObjectInfo (from a candidate of drives in quorum)  picked had an incorrect NumVersions value. 

Consider an object which has two versions where the current version is a DEL marker and the noncurrent version is a regular object.
e.g,
```
$ mc ls --versions play/krisis/obj-1
[2024-05-17 10:17:23 PDT]     0B STANDARD 95a561b3-f91f-46cb-a538-87f3b04adbf5 v2 DEL obj-1
[2024-05-17 10:17:21 PDT]    20B STANDARD dd5853a6-6932-4c05-95d9-96fab07eff78 v1 PUT obj-1

```

This can happen when,
- the DEL marker version is present on all the drives in this erasure set.
- the PUT version (noncurrent) is present in quorum drives
- There are a few drives that are present in the first quorum but not in the second quorum.

When MinIO picks a valid ObjectInfo value constructed, it could pick the one from a drive not present in the second quorum (i.e for the PUT version). This affects ObjectInfo fields that are not object specific, like NumVersions.
This PR looks for a quorum agreed value for NumVersions if available. This is similar to https://github.com/minio/minio/pull/17925

## How to test this PR?
Unit tests in erasure-metadata_test.go capture this possibility.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/erasure-metadata-utils.go
+++ b/cmd/erasure-metadata-utils.go
@@ -26,6 +26,20 @@ import (
 	"github.com/minio/pkg/v2/sync/errgroup"
 )
 
+// counterMap type adds GetValueWithQuorum method to a map[T]int used to count occurrences of values of type T.
+type counterMap[T comparable] map[T]int
+
+// GetValueWithQuorum returns the first key which occurs >= quorum number of times.
+func (c counterMap[T]) GetValueWithQuorum(quorum int) (T, bool) {
+	var zero T
+	for x, count := range c {
+		if count >= quorum {
+			return x, true
+		}
+	}
+	return zero, false
+}
+
 // figure out the most commonVersions across disk that satisfies
 // the 'writeQuorum' this function returns "" if quorum cannot
 // be achieved and disks have too many inconsistent versions.
--- a/cmd/erasure-metadata.go
+++ b/cmd/erasure-metadata.go
@@ -344,9 +344,14 @@ func findFileInfoInQuorum(ctx context.Context, metaArr []FileInfo, modTime time.
 		return FileInfo{}, InsufficientReadQuorum{Err: errErasureReadQuorum, Type: RQInconsistentMeta}
 	}
 
-	// Find the successor mod time in quorum, otherwise leave the
-	// candidate's successor modTime as found
-	succModTimeMap := make(map[time.Time]int)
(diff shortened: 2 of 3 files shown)


cmd/erasure-metadata-utils.go
```diff
@@ -26,6 +26,20 @@ import (
 	"github.com/minio/pkg/v2/sync/errgroup"
 )
 
+// counterMap type adds GetValueWithQuorum method to a map[T]int used to count occurrences of values of type T.
+type counterMap[T comparable] map[T]int
```
harshavardhana, line 30: Very nice - I see you learned new tricks.

cmd/erasure-metadata.go
```diff
@@ -357,24 +358,20 @@ func findFileInfoInQuorum(ctx context.Context, metaArr []FileInfo, modTime time.
...
-		if smodTimeQuorum {
+		// Update candidate FileInfo with succModTime and numVersions in quorum when available
+		if succModTime, ok := succModTimeMap.GetValueWithQuorum(quorum); ok {
 			candidate.SuccessorModTime = succModTime
 			candidate.IsLatest = succModTime.IsZero()
 		}
+		if numVersions, ok := numVersionsMap.GetValueWithQuorum(quorum); ok {
+			candidate.NumVersions = numVersions
+		}
 		return candidate, nil
```
harshavardhana: shouldn't we agree on NumVersions, SuccessorModTime ? having them separate would lead to different outcomes? 

It is ever possible they agreed on NumVersions not agree on SUccessorModTime?

krisis: Good point. In practice, there will be only a single quorum so it should not have different outcomes. It is possible that there are no good outcomes, i.e we may not have a quorum about say a second version, which affects both SuccessorModTime and NumVersions. Since it is not possible to have different values anyway, I shall modify the PR to look for quorum for the combination of SuccessorModTime and NumVersions when available.

harshavardhana (approved):

poornas (approved):

PR #19780, opened 2024-05-21
Add multipart combination to xl-meta

klauspost:
## Description

Add automatic combination of multiple parts.

Parts will be reconstructed and saved separately and can manually be combined to the complete object.

Parts will be named `(version_id)-(filename).(partnum).(in)complete`.

Missing non-inline parts will be printed.

Use flags: `-export -combine -xver`

## How to test this PR?

Export inspect data use `xl-meta -export -combine -xver file.zip`.

## Types of changes

### DIFF
--- a/docs/debugging/xl-meta/main.go
+++ b/docs/debugging/xl-meta/main.go
@@ -859,7 +859,7 @@ func readAndMap(files []string, partNum, blockNum int) (*mappedData, error) {
 		if blockNum < m.blocks {
 			m.size = ei.V2Obj.EcBSize
 		} else {
-			m.size = m.size - offset
+			m.size -= offset
 		}
 
 		b, err := os.ReadFile(file)
--- a/docs/debugging/xl-meta/main.go
+++ b/docs/debugging/xl-meta/main.go
@@ -30,6 +30,7 @@ import (
 	"path/filepath"
 	"regexp"
 	"sort"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


harshavardhana (approved):

docs/debugging/xl-meta/main.go
```diff
@@ -976,179 +1012,204 @@ func combine(files []string, out string) error {
...
+						}
+						splitFilled, err := rs.Split(m.filled)
+						if err != nil {
+							return err
+						}
+						ok := len(splitFilled)
+						for i, sh := range splitFilled {
+							for _, v := range sh {
+								if v == 0 {
+									split[i] = nil
```
krisis, line 1145: Neat!

docs/debugging/xl-meta/main.go
```diff
@@ -976,179 +1012,204 @@ func combine(files []string, out string) error {
...
-					}
-					hasParity := 0
-					for idx, sh := range v {
-						split[idx] = sh
-						if idx >= k && len(v) > 0 {
-							hasParity++
+						hasParity := 0
+						for idx, sh := range v {
+							split[idx] = sh
+							if idx >= k && len(v) > 0 {
```
krisis: Shouldn't we be checking if the shard `sh` at drive `idx` is present?
```suggestion
							if idx >= k && len(sh) > 0 {
```

krisis (approved):
Great work! I have only one outstanding question/comment.

docs/debugging/xl-meta/main.go
harshavardhana: @klauspost ^^

klauspost: Would look like it. But I think it is just a needless secondary check, since we on have valid parity shards here anyway.

(also hasParity is only for printing it seems)

PR #19763, opened 2024-05-17
Fix two-way stream cancelation and pings

klauspost:
## Description

@harshavardhana Not urgent. This can wait until after next release.

Rewrite two-way stream ping and add tests. Currently these streams are not used.

## Motivation and Context

Stability and for future features.

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Unit tests added/updated

PR #19796, opened 2024-05-23
Two way streams fixes - take 2

klauspost:
## Description

Re-add #19763 - fixes are individual commits.

Fixes races observed with new tests and tweak tests for flaky operations.

Also fixes a goroutine leak with (unused) two-way streams.

## How to test this PR?

In internal/grid, use `go test -v --count=100 -race -timeout=30m -failfast` or similar.

## Types of changes


harshavardhana (approved):

klauspost:
@krisis @vadmeste PTAL. Now (just post release) would be a good time to get this merged.

internal/grid/connection.go
```diff
@@ -1406,6 +1412,10 @@ func (c *Connection) handleRequest(ctx context.Context, m message, subID *subHan
 }
 
 func (c *Connection) handlePong(ctx context.Context, m message) {
+	if m.MuxID == 0 && m.Payload == nil {
+		atomic.StoreInt64(&c.LastPong, time.Now().Unix())
```
harshavardhana:
```suggestion
		atomic.StoreInt64(&c.LastPong, time.Now().UnixNano())
```

?

harshavardhana (changes requested):

internal/grid/connection.go
klauspost: The one on `Connection` is secs. The one on `muxClient` is nano. Updated it so I could do the tests without them taking minutes.

klauspost: Maybe I should switch Connection to just nano for consistency. It will not overflow for a long time anyway.

klauspost: Friday, April 11, 2262 will be a bad day :D

krisis (approved):
LGTM. Out of curiosity, where would two way streams be used in MinIO's internode communications?

klauspost:
@krisis Locking is most obvious. That way the request will track the lifetime of the lock request to the remote.

harshavardhana (approved):

PR #19820, opened 2024-05-28
policy: More defensive code validating svc:DurationSeconds

vadmeste:

## Description
This does not fix any current issue, but merging https://github.com/minio/madmin-go/pull/282 
can loose the validation of the service account expiration time and create a security issue.

Add more defensive code for now. In the future, we should avoid doing validation in another library.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/admin-handlers-users.go
+++ b/cmd/admin-handlers-users.go
@@ -824,7 +824,11 @@ func (a adminAPIHandlers) UpdateServiceAccount(w http.ResponseWriter, r *http.Re
 	}
 
 	condValues := getConditionValues(r, "", cred)
-	addExpirationToCondValues(updateReq.NewExpiration, condValues)
+	err = addExpirationToCondValues(updateReq.NewExpiration, condValues)
+	if err != nil {
+		writeErrorResponseJSON(ctx, w, toAdminAPIErr(ctx, err), r.URL)
+		return
+	}
 
 	// Permission checks:
 	//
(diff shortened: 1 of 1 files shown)


cmd/admin-handlers-users.go
```diff
@@ -2453,11 +2457,16 @@ func (a adminAPIHandlers) ImportIAM(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
-func addExpirationToCondValues(exp *time.Time, condValues map[string][]string) {
+func addExpirationToCondValues(exp *time.Time, condValues map[string][]string) error {
 	if exp == nil {
-		return
+		return nil
+	}
+	dur := exp.Sub(time.Now())
```
harshavardhana, line 2464:
```suggestion
	dur := exp.Sub(time.Now())
```

check if `exp.IsZero()` is true or `exp.Equal(time.Unix(0,0))`  return `nil`

harshavardhana (changes requested):

cmd/admin-handlers-users.go
klauspost: Both would make the `if dur <= 0 {` below fail AFAICT, so it seems a bit superfluous.

klauspost (approved):
LGTM

cmd/admin-handlers-users.go
vadmeste: @klauspost though Expiration is a pointer, some clients send a non nil expiration which is zero to mean I did not set any expiration in this request. I believe this is what Harsha meant. Fixed it

harshavardhana (approved):

PR #19835, opened 2024-05-29
Add `AccessKeysListLDAPBulk` function to allow listing accesskeys for multiple/all users in one call

taran-p:

## Description
Adds new `AccessKeyListLDAPBulk` API endpoint for listing access keys for multiple or all users in one call.
Adds new `GetAllSTSUserMappings` IAM Store method that retrieves all STS users that have a mapped policy or an associated service account/STS key.
Changes `ListLDAPUsers` to use the `GetAllSTSUserMappings` function.

Relies on minio/madmin-go#284

## Motivation and Context
Currently listing all access keys for LDAP (such as in `mc idp ldap accesskey list`) requires individually querying `AccessKeyListLDAP` for each user. This fixes that, along with improvements like returning the normalized DN.

`ListLDAPUsers` currently only fetches LDAP users with mapped policies in the cache. STS Users are loaded lazily, so this method leads to inconsistent and unreliable results. It also misses LDAP users that have no policy mapped, but are both part of a group with a mapped policy and have a service account/sts key. The new function fixes these problems by fetching the STS policies mappings from storage, as well as checking for users with associated access keys.

## How to test this PR?
Use the updated `mc idp ldap accesskey list` command from **minio/mc#4946**
Using `mc admin user list` will also allow for checking `ListLDAPUsers` new functionality.

## Types of changes

## Checklist:


donatello (approved):

go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/minio/minio
 
 go 1.21
 
+replace github.com/minio/madmin-go/v3 => github.com/taran-p/madmin-go/v3 v3.10.1-0.20240606204813-55a2c38ac0b1
+
```
donatello: New madmin tag is created @taran-p

cmd/admin-handlers-idp-ldap.go
```diff
@@ -479,3 +479,172 @@ func (a adminAPIHandlers) ListAccessKeysLDAP(w http.ResponseWriter, r *http.Requ
...
+		Action:          policy.ListServiceAccountsAdminAction,
+		ConditionValues: getConditionValues(r, "", cred),
+		IsOwner:         owner,
+		Claims:          cred.Claims,
+		DenyOnly:        onlySelf,
+	}) {
+		writeErrorResponseJSON(ctx, w, errorCodes.ToAPIErr(ErrAccessDenied), r.URL)
+		return
+	}
+
```
donatello, line 546: Looks like some permission checking is missing - there are multiple cases:
1. All is set
2. onlySelf is implied (by dnList.len() == 0 and !isAll)
3. A list of userDNs is given.

For 1 and 3 we need ListUsersAdminAction right?

donatello (changes requested):

cmd/admin-handlers-idp-ldap.go
taran-p: Should 3 require ListUsersAdminAction? It's roughly equivalent to calling `ListAccessKeysLDAP` or `ListServiceAccounts` for each user, and neither of those requires ListUsersAdminAction

donatello: Ok sounds good. Does querying one's own access keys work properly when one is not a privileged user?

taran-p: Yes, if the query is onlySelf (by implication or if the only DN in dnList) then DenyOnly should be true (edit: the dn is not validated in this case, I'll add that)

taran-p: Ok, that should be working properly now as well

harshavardhana (approved):

harshavardhana:
PTAL @donatello

donatello (approved):

PR #19839, opened 2024-05-30
avoid pre-populating buffers for deployments < 32GiB memory

harshavardhana:

## Description
avoid pre-populating buffers for deployments < 32GiB memory

## Motivation and Context
stricter behavior and set appropriate limits

## How to test this PR?
You would have to create a `dockerized` environment
with limited memory to see this in effect. 

## Types of changes

## Checklist:

### DIFF
--- a/cmd/common-main.go
+++ b/cmd/common-main.go
@@ -402,6 +402,29 @@ func buildServerCtxt(ctx *cli.Context, ctxt *serverCtxt) (err error) {
 		ctxt.certsDirSet = true
 	}
 
+	memAvailable := availableMemory()
+	if ctx.IsSet("memlimit") || ctx.GlobalIsSet("memlimit") {
+		memlimit := ctx.String("memlimit")
+		if memlimit == "" {
+			memlimit = ctx.GlobalString("memlimit")
+		}
+		mlimit, err := humanize.ParseBytes(memlimit)
+		if err != nil {
+			return err
+		}
+		if mlimit > memAvailable {
+			logger.Info("WARNING: maximum memory available (%s) smaller than specified --memlimit=%s, ignoring --memlimit value",
+				humanize.IBytes(memAvailable), memlimit)
+		}
+		ctxt.MemLimit = mlimit
+	} else {
+		ctxt.MemLimit = memAvailable
+	}
+
+	if memAvailable < ctxt.MemLimit {
+		ctxt.MemLimit = memAvailable
+	}
+
 	ctxt.FTP = ctx.StringSlice("ftp")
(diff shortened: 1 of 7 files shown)


jiuker (approved):

cmd/common-main.go
```diff
@@ -402,6 +402,29 @@ func buildServerCtxt(ctx *cli.Context, ctxt *serverCtxt) (err error) {
 		ctxt.certsDirSet = true
 	}
 
+	memAvailable := availableMemory()
+	if ctx.IsSet("memlimit") {
```
vadmeste:
```suggestion
	if ctx.IsSet("memlimit") || ctx.GlobalIsSet("memlimit") {
```

cmd/common-main.go
```diff
@@ -402,6 +402,29 @@ func buildServerCtxt(ctx *cli.Context, ctxt *serverCtxt) (err error) {
...
+		memlimit := ctx.String("memlimit")
+		if memlimit == "" {
+			memlimit = ctx.GlobalString("memlimit")
+		}
+		mlimit, err := humanize.ParseBytes(memlimit)
+		if err != nil {
+			return err
+		}
+		if mlimit > memAvailable {
+			logger.Info("WARNING: maximum memory available (%s) smaller than specified --memlimit=%s, ignoring --memlimit value",
```
vadmeste, line 416: We are not really ignoring here, right ? we are still setting `ctxt.MemLimit = mlimit` later

cmd/common-main.go
```diff
@@ -402,6 +402,29 @@ func buildServerCtxt(ctx *cli.Context, ctxt *serverCtxt) (err error) {
...
+				humanize.IBytes(memAvailable), memlimit)
+		}
+		ctxt.MemLimit = mlimit
+	} else {
+		ctxt.MemLimit = memAvailable
+	}
+
+	if memAvailable < ctxt.MemLimit {
+		ctxt.MemLimit = memAvailable
+	}
```
vadmeste, line 426: I think let's avoid setting any limit here if --memlimit is not specified, that is the old behavior anyway. If I am not wrong, this will make the GC works very hard even before the memory usage reaches 75%

harshavardhana: If you are using that much memory you better GC @vadmeste instead of crashing. There is nothing wrong in setting this value.

harshavardhana: we should really use this value properly/appropriately because otherwise we risk of OOM in unexpected ways.

vadmeste (approved):
LGTM

PR #19849, opened 2024-05-31
kafka: _MINIO_KAFKA_DEBUG to enable sarama debug messages

vadmeste:

## Types of changes

## Checklist:

### DIFF
--- a/internal/event/target/kafka.go
+++ b/internal/event/target/kafka.go
@@ -24,6 +24,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"log"
 	"net/url"
 	"os"
 	"path/filepath"
@@ -331,6 +332,10 @@ func (target *KafkaTarget) init() error {
 }
 
 func (target *KafkaTarget) initKafka() error {
+	if os.Getenv("_MINIO_KAFKA_DEBUG") != "" {
+		sarama.DebugLogger = log.Default()
+	}
+
 	args := target.args
 
 	config := sarama.NewConfig()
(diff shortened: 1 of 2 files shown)


klauspost (commented):
Would it be crazy to hook it up to the tracer as it's own type?

Not sure if there is a big cost by having it enabled, but just skipping if nothing is listening?

vadmeste:
> Would it be crazy to hook it up to the tracer as it's own type?

I would prefer a new API for verbose debug messages in general

poornas (approved):

klauspost (approved):

klauspost:
@vadmeste Yeah. My main issue is that

A) This requires a restart to enable/disable.
B) Collecting logs may be "challenging" for some users.
C) Logging doesn't really have any type filtering, so we cannot implement something that is (close to) zero cost when not enabled.

Since traces fixes these that was my goto-thought - though it doesn't really match the "text logging" schema. Let's add this and consider our options.

harshavardhana:
> @vadmeste Yeah. My main issue is that
> 
> A) This requires a restart to enable/disable. B) Collecting logs may be "challenging" for some users. C) Logging doesn't really have any type filtering, so we cannot implement something that is (close to) zero cost when not enabled.
> 
> Since traces fixes these that was my goto-thought - though it doesn't really match the "text logging" schema. Let's add this and consider our options.

This env is mostly never going to be used, but it is needed in very rare situations when you are dealing with saas kaka offering as we recently found during debugging.

PR #19861, opened 2024-06-03
fix: log diskerror when detect the disk space failed

jiuker:
fix: log diskerror when delect the disk space failed

## Description

fix: log diskerror when delect the disk space failed
when we got `not enough online disks to calculate the available space`
we don't print the error and provide a way to print the error here.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/object-api-utils.go
+++ b/cmd/object-api-utils.go
@@ -1206,19 +1206,18 @@ func hasSpaceFor(di []*DiskInfo, size int64) (bool, error) {
 	}
 
 	if nDisks < len(di)/2 || nDisks <= 0 {
-		var errs []string
-		for _, disk := range di {
+		var errs []error
+		for index, disk := range di {
 			if disk == nil {
-				continue
-			}
-			if disk.Error != "" {
-				errs = append(errs, fmt.Sprintf("disk %s: %s", disk.Endpoint, disk.Error))
+				errs = append(errs, fmt.Errorf("disk[%d]: offline", index))
+			} else if disk.Error != "" {
+				errs = append(errs, fmt.Errorf("disk %s: %s", disk.Endpoint, disk.Error))
 			} else if disk.Total == 0 {
-				errs = append(errs, fmt.Sprintf("disk %s: total is zero", disk.Endpoint))
+				errs = append(errs, fmt.Errorf("disk %s: total is zero", disk.Endpoint))
 			}
 		}
 		// Log disk errors.
-		peersLogIf(context.Background(), errors.New(strings.Join(errs, ", ")))
+		peersLogIf(context.Background(), errors.Join(errs...))
 		return false, fmt.Errorf("not enough online disks to calculate the available space, need %d, found %d", (len(di)/2)+1, nDisks)
 	}
 
--- a/cmd/object-api-utils.go
+++ b/cmd/object-api-utils.go
@@ -1206,6 +1206,19 @@ func hasSpaceFor(di []*DiskInfo, size int64) (bool, error) {
 	}
 
 	if nDisks < len(di)/2 || nDisks <= 0 {
+		var errs []string
+		for _, disk := range di {
+			if disk == nil {
+				continue
+			}
+			if disk.Error != "" {
+				errs = append(errs, fmt.Sprintf("disk %s: %s", disk.Endpoint, disk.Error))
+			} else if disk.Total == 0 {
+				errs = append(errs, fmt.Sprintf("disk %s: total is zero", disk.Endpoint))
+			}
+		}
+		// Log disk errors.
+		peersLogIf(context.Background(), errors.New(strings.Join(errs, ", ")))
 		return false, fmt.Errorf("not enough online disks to calculate the available space, expected (%d)/(%d)", (len(di)/2)+1, nDisks)
 	}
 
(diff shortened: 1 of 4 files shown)


cmd/object-api-utils.go
```diff
@@ -1206,6 +1206,19 @@ func hasSpaceFor(di []*DiskInfo, size int64) (bool, error) {
 	}
 
 	if nDisks < len(di)/2 || nDisks <= 0 {
+		var errs []string
```
vadmeste: I think we can use the new errors.Join() API (go 1.20), I think it will be easier.

cmd/object-api-utils.go
```diff
@@ -1206,6 +1206,19 @@ func hasSpaceFor(di []*DiskInfo, size int64) (bool, error) {
...
+			}
+			if disk.Error != "" {
+				errs = append(errs, fmt.Sprintf("disk %s: %s", disk.Endpoint, disk.Error))
+			} else if disk.Total == 0 {
+				errs = append(errs, fmt.Sprintf("disk %s: total is zero", disk.Endpoint))
+			}
+		}
+		// Log disk errors.
+		peersLogIf(context.Background(), errors.New(strings.Join(errs, ", ")))
 		return false, fmt.Errorf("not enough online disks to calculate the available space, expected (%d)/(%d)", (len(di)/2)+1, nDisks)
```
vadmeste: while we are here, can you switch the position of `len(di)/2)+1` and `nDisks`, it is confusing right now

cmd/object-api-utils.go
jiuker: @vadmeste errors.Join(errs ....error). 
```go
type DiskInfo struct {
	...
	Error      string // carries the error over the network
}
```
That's string type. So I think it's not easier.

cmd/object-api-utils.go
```diff
@@ -1219,7 +1219,7 @@ func hasSpaceFor(di []*DiskInfo, size int64) (bool, error) {
 		}
 		// Log disk errors.
 		peersLogIf(context.Background(), errors.New(strings.Join(errs, ", ")))
-		return false, fmt.Errorf("not enough online disks to calculate the available space, expected (%d)/(%d)", (len(di)/2)+1, nDisks)
+		return false, fmt.Errorf("not enough online disks to calculate the available space, expected (%d)/(%d)", nDisks, (len(di)/2)+1)
```
klauspost: Your change doesn't really make anything clearer. Maybe instead:
```suggestion
		return false, fmt.Errorf("not enough online disks to calculate the available space, need %d, found %d",  (len(di)/2)+1, nDisks)
```

cmd/object-api-utils.go
```diff
@@ -1206,7 +1206,20 @@ func hasSpaceFor(di []*DiskInfo, size int64) (bool, error) {
 	}
 
 	if nDisks < len(di)/2 || nDisks <= 0 {
-		return false, fmt.Errorf("not enough online disks to calculate the available space, expected (%d)/(%d)", (len(di)/2)+1, nDisks)
+		var errs []string
+		for _, disk := range di {
+			if disk == nil {
+				continue
```
vadmeste: can we also add an error here ? like `a nil disk (index) is found`

cmd/object-api-utils.go
```diff
@@ -1206,7 +1206,20 @@ func hasSpaceFor(di []*DiskInfo, size int64) (bool, error) {
 	}
 
 	if nDisks < len(di)/2 || nDisks <= 0 {
-		return false, fmt.Errorf("not enough online disks to calculate the available space, expected (%d)/(%d)", (len(di)/2)+1, nDisks)
+		var errs []string
```
```suggestion
		var errs []error
```

cmd/object-api-utils.go
```diff
@@ -1206,7 +1206,20 @@ func hasSpaceFor(di []*DiskInfo, size int64) (bool, error) {
...
+		for _, disk := range di {
+			if disk == nil {
+				continue
+			}
+			if disk.Error != "" {
+				errs = append(errs, fmt.Sprintf("disk %s: %s", disk.Endpoint, disk.Error))
```
```suggestion
				errs = append(errs, fmt.Errorf("disk %s: %s", disk.Endpoint, disk.Error))
```

cmd/object-api-utils.go
```diff
@@ -1206,7 +1206,20 @@ func hasSpaceFor(di []*DiskInfo, size int64) (bool, error) {
...
+			} else if disk.Total == 0 {
+				errs = append(errs, fmt.Sprintf("disk %s: total is zero", disk.Endpoint))
```
```suggestion
				errs = append(errs, fmt.Errorf("disk %s: total is zero", disk.Endpoint))
```

cmd/object-api-utils.go
```diff
@@ -1206,7 +1206,20 @@ func hasSpaceFor(di []*DiskInfo, size int64) (bool, error) {
...
+			}
+		}
+		// Log disk errors.
+		peersLogIf(context.Background(), errors.New(strings.Join(errs, ", ")))
```
```suggestion
		peersLogIf(context.Background(), errors.Join(errs...))
```

harshavardhana (changes requested):

cmd/object-api-utils.go
harshavardhana: A `nil` is an offline disk we should just say that.

harshavardhana (approved):

ramondeklein (approved):

PR #19877, opened 2024-06-05
feat: support batchJob for windows

jiuker:
feat: support batch job for windows

## Description

feat: support batchJob for windows
without this pr
we will get this:
```
PS C:\workspace\go\src\subnet-issue> mc batch start minio replicate.yaml
mc.exe: <ERROR> Unable to start job. Object name contains unsupported characters.
```

## Types of changes

## Checklist:

### DIFF
--- a/cmd/batch-job-common-types.go
+++ b/cmd/batch-job-common-types.go
@@ -291,7 +291,7 @@ func (s *BatchJobSize) UnmarshalYAML(unmarshal func(interface{}) error) error {
 }
 
 // getBatchJobIDSeparator - returns the separator to be used in the batch job ID
-// windows requires _ as the separator for : will be an invalid object name
+// windows requires `_` as the separator `:` will be an invalid one
 func getBatchJobIDSeparator() string {
 	if runtime.GOOS == "windows" {
 		return "_"
--- a/cmd/batch-job-common-types.go
+++ b/cmd/batch-job-common-types.go
@@ -293,7 +293,7 @@ func (s *BatchJobSize) UnmarshalYAML(unmarshal func(interface{}) error) error {
 // getBatchJobIDSeparator - returns the separator to be used in the batch job ID
 // windows requires `_` as the separator `:` will be an invalid one
 func getBatchJobIDSeparator() string {
-	if runtime.GOOS == "windows" {
+	if runtime.GOOS == globalWindowsOSName {
 		return "_"
 	}
 	return ":"
--- a/cmd/batch-job-common-types.go
+++ b/cmd/batch-job-common-types.go
@@ -19,6 +19,7 @@ package cmd
 
 import (
 	"fmt"
+	"runtime"
 	"strings"
 	"time"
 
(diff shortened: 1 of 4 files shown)


harshavardhana (approved):

ramondeklein (commented):
Why not always use `_` for consistency? If it works both on Linux, MacOS and Windows then I prefer the same for all platforms.

jiuker:
> Why not always use `_` for consistency? If it works both on Linux, MacOS and Windows then I prefer the same for all platforms.

It's ok for fresh install. But ... @ramondeklein 
@harshavardhana Any advice?

klauspost:
Yeah. We need to keep it `:` for backcompat AFAICT.

klauspost (approved):

cmd/batch-job-common-types.go
```diff
@@ -288,3 +289,12 @@ func (s *BatchJobSize) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	*s = BatchJobSize(sz)
 	return nil
 }
+
+// getBatchJobIDSeparator - returns the separator to be used in the batch job ID
+// windows requires _ as the separator for : will be an invalid object name
```
shtripat:
```suggestion
// windows requires `_` as the separator `:` will be an invalid one
```

shtripat (approved):
LGTM. A minor one

harshavardhana:
> Why not always use `_` for consistency? If it works both on Linux, MacOS and Windows then I prefer the same for all platforms.

this is to make sure we are backward compatible @ramondeklein

harshavardhana (approved):

cmd/batch-job-common-types.go
```diff
@@ -288,3 +289,12 @@ func (s *BatchJobSize) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	*s = BatchJobSize(sz)
 	return nil
 }
+
+// getBatchJobIDSeparator - returns the separator to be used in the batch job ID
+// windows requires `_` as the separator `:` will be an invalid one
+func getBatchJobIDSeparator() string {
+	if runtime.GOOS == "windows" {
```
harshavardhana:
```suggestion
	if runtime.GOOS == globalWindowsOSName {
```

Minor strictly not needed, and please fix

```diff
diff --git a/.github/workflows/vulncheck.yml b/.github/workflows/vulncheck.yml
index d12a20046f..4452575055 100644
--- a/.github/workflows/vulncheck.yml
+++ b/.github/workflows/vulncheck.yml
@@ -21,7 +21,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v5
         with:
-          go-version: 1.22.3
+          go-version: 1.22.4
           check-latest: true
       - name: Get official govulncheck
         run: go install golang.org/x/vuln/cmd/govulncheck@latest
```

klauspost: `1.22.x` - there is no point in bumping every minor version manually

harshavardhana: I explained why it was because of what @aead suggested, the correct way to run `vulncheck` that exposes we must Go upgrade for our releases and make a new release.

klauspost: @harshavardhana @aead IMO we should run this on releases. Running it here is pointless, since we only need to check if the *release* is with the correct Go version - and this is not guaranteed to be the release version.

harshavardhana (approved):

PR #19878, opened 2024-06-05
fix: make sure we list freeVersions like DEL marker with --versions

harshavardhana:

## Description
fix: make sure we list freeVersions like DEL marker with --versions

## Motivation and Context
freeVersions() was being incorrectly skipped, list it as valid objects properly.

## How to test this PR?
You would have to create a situation where only a FreeVersion() is
left on the namespace, for a specific object.

Ask me for an inspect object for the same, if you want to test
the behavior.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/xl-storage-format-utils.go
+++ b/cmd/xl-storage-format-utils.go
@@ -23,22 +23,36 @@ import (
 	"github.com/zeebo/xxh3"
 )
 
-func getFileInfoVersions(xlMetaBuf []byte, volume, path string, allParts bool) (FileInfoVersions, error) {
+func getFileInfoVersions(xlMetaBuf []byte, volume, path string, allParts, inclFreeVersions bool) (FileInfoVersions, error) {
 	fivs, err := getAllFileInfoVersions(xlMetaBuf, volume, path, allParts)
 	if err != nil {
 		return fivs, err
 	}
 	n := 0
+	if !inclFreeVersions {
+		for _, fi := range fivs.Versions {
+			// Filter our tier object delete marker
+			if !fi.TierFreeVersion() {
+				fivs.Versions[n] = fi
+				n++
+			} else {
+				fivs.FreeVersions = append(fivs.FreeVersions, fi)
+			}
+		}
+		fivs.Versions = fivs.Versions[:n]
+		// Update numversions
+		for i := range fivs.Versions {
+			fivs.Versions[i].NumVersions = n
+		}
+		return fivs, nil
+	}
 	for _, fi := range fivs.Versions {
-		// Filter our tier object delete marker
-		if !fi.TierFreeVersion() {
-			fivs.Versions[n] = fi
-			n++
-		} else {
+		if fi.TierFreeVersion() {
 			fivs.FreeVersions = append(fivs.FreeVersions, fi)
+		} else {
+			n++
 		}
 	}
-	fivs.Versions = fivs.Versions[:n]
 	// Update numversions
 	for i := range fivs.Versions {
 		fivs.Versions[i].NumVersions = n
--- a/cmd/xl-storage-format-utils.go
+++ b/cmd/xl-storage-format-utils.go
@@ -23,35 +23,39 @@ import (
 	"github.com/zeebo/xxh3"
 )
 
+// getFileInfoVersions partitions this object's versions such that,
+//   - fivs.Versions has all the non-free versions
+//   - fivs.FreeVersions has all the free versions
+//
+// if inclFreeVersions is true all the versions are in fivs.Versions, free and non-free versions alike.
+//
+// Note: Only the scanner requires fivs.Versions to have exclusively non-free versions. This is used while enforcing NewerNoncurrentVersions lifecycle element.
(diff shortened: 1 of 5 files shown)


klauspost:
@krisis to the rescue pls ;)

Isn't the issue that the delete marker *is* deleted, but the delete is not replicated? AFAICT then it shouldn't be returned.

Though I was never sure I understood "free markers" in the first place.

.github/workflows/vulncheck.yml
```diff
@@ -21,7 +21,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v5
         with:
-          go-version: 1.22.3
+          go-version: 1.22.4
```
klauspost:
```suggestion
          go-version: 1.22.x
```

Is there a point to manually upgrade this?

This isn't linked to the build system anyway. What does that use?

harshavardhana:
> Isn't the issue that the delete marker _is_ deleted, but the delete is not replicated? AFAICT then it shouldn't be returned.

This issue has nothing to do with the DELETE marker @klauspost; it is related to the Tier object when it is deleted. You add a free version marker to track a successful DELETE on the warm tier so that you can delete it once it happens locally on the hot tier.

the free version == DEL marker (but from a tiering point of view)

Delete marker deleted but not replicated is still a delete marker, which must be returned in --versions output.

This is not about regular List(). It's about ListVersions(), which needs to return everything underneath a path. 

It must be consistently the same so that we don't have ghost folders and no way to know what it is.

.github/workflows/vulncheck.yml
harshavardhana: .x doesn't work @klauspost thats why its kept with actual version.

harshavardhana: Its @aead suggestion that we must always run with actual version for vulncheck this basically tells us that we must make new releases as there are vulnchecks failing.

The version here sticks to the version we use to make releases.

harshavardhana: https://github.com/minio/kes/blob/fe54489812091915144bed6c3a793312fce31e09/.github/workflows/go.yml#L73

klauspost: @harshavardhana I don't suppose this is extracted for the build. 

You could run it on the *build* version when doing a release. This tells you nothing, except being annoying every time a minor release is made.

harshavardhana: We can remove this once we add `vulncheck` to release process.

harshavardhana: // cc @donatello

krisis:
@harshavardhana,
tl;dr free-versions are internal to MinIO and shouldn't be visible to applications via S3 API.

Facts about free-versions:
- Internal to MinIO
- Created (only) when MinIO is unable to delete the tiered object version's content due to temporary unavailability of its warm tier
- Hold a reference to a tiered object version's content in its warm tier

Implementation details:
Free version is implemented as an extension to the Delete Marker type version (`xlMetaV2DeleteMarker`) with additional metadata which differentiate it from a regular DEL marker. 
Free versions are only visible to the scanner process for eventual cleanup of references held by them.

harshavardhana:
I disagree @krisis  this PR is valid, all additional versions in xl.meta that we carry must be listed via --versions otherwise it will confuse the heck out of every other code out there.

harshavardhana:
> Implementation details:
> Free version is implemented as an extension to the Delete Marker type version (`xlMetaV2DeleteMarker`) with additional metadata which differentiate it from a regular DEL marker.
> Free versions are only visible to the scanner process for eventual cleanup of references held by them.

DEL marker is the visual representation for the end-user; that is precisely why we don't list this in the regular List(). It is only listed via ListVersions(). 

Visual representation is as essential for supportability. A free version left over is precisely like a DEL marker with no other non-current versions. 

The final behavior is precisely how a DEL marker is visible.

cmd/xl-storage-format-utils.go
```diff
@@ -30,15 +30,12 @@ func getFileInfoVersions(xlMetaBuf []byte, volume, path string, allParts bool) (
...
 	n := 0
 	for _, fi := range fivs.Versions {
-		// Filter our tier object delete marker
-		if !fi.TierFreeVersion() {
-			fivs.Versions[n] = fi
-			n++
-		} else {
+		if fi.TierFreeVersion() {
 			fivs.FreeVersions = append(fivs.FreeVersions, fi)
+		} else {
```
krisis, line 50: `applyVersionActions` (from scanner) expects `getFileInfoVersions` to return `fivs FileInfoVersions` such that, `fivs.Versions` contains only non-free versions and `fivs.FreeVersions` contains only free versions. 
This is used when applying `NewerNoncurrentVersions` lifecycle element.

To address this, we can extend `getFileInfoVersions` to take an argument, `inclFreeVersions bool`, based on which we can populate `fivs.Versions` to include free versions (or not).
```

harshavardhana:
PTAL @krisis @vadmeste @poornas we need to merge this.

krisis (approved):
LGTM. I have changed my views on making free versions visible to the user. While it exposes an implementation detail it does provide visibility into a tiered object version's lifecycle, especially when its expired/deleted while its warm-tier was temporarily offline.

PR #19883, opened 2024-06-05
decom: Fix CurrentSize output when generating the status

vadmeste:

## Description
StartSize starts with the  free usable space of all disks in the given pool, 
however during the status, CurrentSize is not showing the current free 
 space, as expected at least by `mc admin decom status` since it was written.

## Motivation and Context
Fix mc admin decom status

## Types of changes

## Checklist:

### DIFF
--- a/cmd/erasure-server-pool-decom.go
+++ b/cmd/erasure-server-pool-decom.go
@@ -1327,11 +1327,7 @@ func (z *erasureServerPools) Status(ctx context.Context, idx int) (PoolStatus, e
 	poolInfo := z.poolMeta.Pools[idx].Clone()
 	if poolInfo.Decommission != nil {
 		poolInfo.Decommission.TotalSize = pi.Total
-		if poolInfo.Decommission.Failed || poolInfo.Decommission.Canceled {
-			poolInfo.Decommission.CurrentSize = pi.Free
-		} else {
-			poolInfo.Decommission.CurrentSize = poolInfo.Decommission.StartSize + poolInfo.Decommission.BytesDone
-		}
+		poolInfo.Decommission.CurrentSize = pi.Free
 	} else {
 		poolInfo.Decommission = &PoolDecommissionInfo{
 			TotalSize:   pi.Total,


cmd/erasure-server-pool-decom.go
```diff
@@ -1327,11 +1327,7 @@ func (z *erasureServerPools) Status(ctx context.Context, idx int) (PoolStatus, e
 	poolInfo := z.poolMeta.Pools[idx].Clone()
 	if poolInfo.Decommission != nil {
 		poolInfo.Decommission.TotalSize = pi.Total
-		if poolInfo.Decommission.Failed || poolInfo.Decommission.Canceled {
-			poolInfo.Decommission.CurrentSize = pi.Free
-		} else {
-			poolInfo.Decommission.CurrentSize = poolInfo.Decommission.StartSize + poolInfo.Decommission.BytesDone
```
harshavardhana, line 1333: this was done for accuracy reasons, what is the problem here?

harshavardhana (changes requested):

cmd/erasure-server-pool-decom.go
vadmeste: well, I am not able to understand the logic here.. CurrentSize starts is the addition of the free usable space of drives when decom is started, then we add the object sizes that were decommissioned, what is that supposed to mean ?

harshavardhana: StartSize is the first time what was the current Free @vadmeste when the decom was started it should never skew in the manner it skewed for the customer. 

So we keep adding bytesDone to ensure that we keep the Current free size based on the original startSize.

harshavardhana: Eventually the CurrentSize becomes close to TotalSize.

harshavardhana (changes requested):

cmd/erasure-server-pool-decom.go
vadmeste: Okay understood, CurrentSize is close to pi.Free then. 

Something about poolInfo.Decommission.StartSize is that it is a persistant value and it depends on the available number of drives when StartSize was calculated the first time. In my opinion, we could use realtime values instead since we are doing that few lines with z.getDecommissionPoolSpaceInfo(idx)

harshavardhana: yeah if that is the case this entire code is a bit meaningless you just need 

```
		poolInfo.Decommission = &PoolDecommissionInfo{
			TotalSize:   pi.Total,
			CurrentSize: pi.Free,
		}
```

vadmeste: yes, but this is Status() function, we cannot initialize &PoolDecommissionInfo if there is a an ongoing decommissioning, we still want to return other information such as StartTime, etc..

harshavardhana (approved):

PR #19894, opened 2024-06-07
Allow SSE-C on default compressed

klauspost:
## Description

Compression will be disabled by default if SSE-C is specified. So we can still honor SSE-C and there is no need to reject.

## How to test this PR?

Enabled compression. Enable allow compression+encryption. Upload with SSE-C.

## Types of changes


klauspost:
Seems like some test tweaks is needed: `BUG: Loading an SSE-C object to site with compression should fail. Succeeded though.`

We can check the header and see if it was actually compressed.

Can any1 point me to this test?

shtripat:
> Seems like some test tweaks is needed: `BUG: Loading an SSE-C object to site with compression should fail. Succeeded though.`
> 
> We can check the header and see if it was actually compressed.
> 
> Can any1 point me to this test?

docs/site-replication/run-ssec-object-replication-with-compression.sh

klauspost:
Merging into https://github.com/minio/minio/pull/19896

PR #19914, opened 2024-06-11
Listing Policy Mapping Entities will return groups that user is a member of and the group policies

taran-p:

## Description

- Calling `ListLDAPPolicyMappingEntities` or `ListPolicyMappingEntities` for specific users will return groups those users are members of and their associated policies. For LDAP only groups that have policies will be returned.
- `ListLDAPPolicyMappingEntities` now validates and normalizes incoming user or group queries. 
- Outgoing user or group DNs in `ListLDAPPolicyMappingEntities` will be decoded from their internal unicode representation.
- 
## Motivation and Context

- Currently there is no simple way to retrieve associated groups and their policies for LDAP users
- Querying LDAP policy entities for a specific user or group currently does not normalize DNs, meaning the DN must already be formatted as the internal representation of the DN.
- Listing policy entities currently returns DNs with some unicode characters encoded (i.e. `ä` is returned as `\c3\a4`). These characters will now be returned in their human-readable form. 

## How to test this PR?
Use minio/mc#4961 to run `mc idp ldap policy entities` or `mc admin policy entities` on users with groups

## Types of changes

## Checklist:


internal/config/identity/ldap/ldap.go
```diff
@@ -354,3 +403,99 @@ func (l *Config) LookupGroupMemberships(userDistNames []string, userDNToUsername
...
+	}
+	return dn
+}
+
+// Remove leading and trailing spaces from the attribute type and value
+// and unescape any escaped characters in these fields
+//
+// decodeDN is pulled from the go-ldap library
+// https://github.com/go-ldap/ldap/blob/dbdc485259442f987d83e604cd4f5859cfc1be58/dn.go
+func decodeDN(str string) (string, error) {
```
harshavardhana, line 430: why are doing this here?

harshavardhana (changes requested):

internal/config/identity/ldap/ldap.go
taran-p: This is to decode the internal unicode representation of some characters into human-readable characters when they are being sent externally (i.e. `cn=\d0\98\d0\b2\d0\b0\d0\bd\d0\be\d0\b2 \d0\98\d0\b2\d0\b0\d0\bd,cn=Users,dc=min,dc=io` is converted to `cn=Иванов Иван,cn=Users,dc=min,dc=io`)

harshavardhana: Can we move this to `pkg/ldap` and also add some good amount of unit tests, we can't just fork such sensitive code without unit tests for this function.

We should also keep the code as `pkg/ldap/decode_dn_contrib.go` where this file carries an external license header as part of the source file - since its just a copy from the upstream project.

if its a new implementation fresh without copying from upstream then we are good.

harshavardhana (changes requested):

donatello:
@taran-p Of the three motivations listed:

> -  Currently there is no simple way to retrieve associated groups and their policies for LDAP users
>   -  Querying LDAP policy entities for a specific user or group currently does not normalize DNs, meaning the DN must already be formatted as the internal representation of the DN.
> -    Listing policy entities currently returns DNs with some unicode characters encoded (i.e. ä is returned as \c3\a4). These characters will now be returned in their human-readable form.

I think it makes sense to separate out the last two into a separate PR and put the DN decoding logic + tests into the pkg module.

PR #19952, opened 2024-06-19
Do remote unlocks async

klauspost:
## Description

Do not block on distributed unlocks.

* Prevents blocking when losing quorum (common on cluster restarts).
* Time out to prevent endless buildup. Timed out remote locks will be canceled because they miss the refresh anyway.
* Reduces latency for all calls, since the wall time for the roundtrip to remotes no longer add to the requests.

## How to test this PR?

Test restart times in distributed clusters.

## Types of changes


shtripat (approved):
lgtm

harshavardhana (approved):

vadmeste (commented):
LGTM

PR #19956, opened 2024-06-19
Removing timeout on shutdown

zveinn:
## Community Contribution License
All community contributions in this pull request are licensed to the project maintainers
under the terms of the [Apache 2 license](https://www.apache.org/licenses/LICENSE-2.0). 
By creating this pull request I represent that I have the right to license the 
contributions to the project maintainers under the Apache 2 license.

## Description
Stop waiting for requests to finish when shutting down.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Optimization (provides speedup with no functional changes)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [ ] Fixes a regression (If yes, please add `commit-id` or `PR #` here)
- [ ] Unit tests added/updated
- [ ] Internal documentation updated
- [ ] Create a documentation update request [here](https://github.com/minio/docs/issues/new?label=doc-change,title=Doc+Updated+Needed+For+PR+github.com%2fminio%2fminio%2fpull%2fNNNNN)

PR #19960, opened 2024-06-20
Add fixed timed restarts to updates

klauspost:
## Description

Schedule the server restarts to happen at the same time.

## Motivation and Context

Coordinate restarts to happen at approximately the same time for faster startup. Eliminates jitter cause by network latency.

## How to test this PR?

Use `mc admin service restart ALIAS` and `mc admin update ALIAS`.

Please test if possible. I don't have the most common setup.

## Types of changes

### DIFF
--- a/cmd/admin-handlers.go
+++ b/cmd/admin-handlers.go
@@ -543,9 +543,12 @@ func (a adminAPIHandlers) ServiceV2Handler(w http.ResponseWriter, r *http.Reques
 	}
 
 	var objectAPI ObjectLayer
+	var execAt *time.Time
 	switch serviceSig {
 	case serviceRestart:
 		objectAPI, _ = validateAdminReq(ctx, w, r, policy.ServiceRestartAdminAction)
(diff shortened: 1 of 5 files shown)


klauspost:
@harshavardhana @zveinn Should we have a "sanity check", like if difference to current time >10s, ignore it?

vadmeste:
@klauspost I think servers will not restart at the same time because each of those servers will quit gracefully with a  timeout ? Probably we can add a service signal called forceRestart to avoid the graceful shutdown

klauspost:
@vadmeste See https://github.com/minio/minio/pull/19956 - it removes waiting for requests to finish.

vadmeste (approved):
LGTM

harshavardhana (approved):

harshavardhana:
> @harshavardhana @zveinn Should we have a "sanity check", like if difference to current time >10s, ignore it?

We have that for skew time like 15mins?

klauspost:
@harshavardhana Yeah, let's cross that bridge when we get to it.

PR #19971, opened 2024-06-21
Fix error when validating DN that is not under base DN

taran-p:

## Description
Validating a user DN that is not under the configured base DN will now return nil instead of internal error.

## Motivation and Context
Currently if a DN is being validated and it is on the LDAP server but not under the base DN it will return an internal error with `Unable to find user DN: %!w(<nil>)`. This changes the behavior so it is consistent with a user that does not exist

## How to test this PR?

1. Configure LDAP server with user_dn_search_base_dn as `ou=swengg,dc=min,dc=io`
2. `mc idp ldap policy attach` to user `uid=bobfisher,ou=people,ou=hwengg,dc=min,dc=io`
Previously this attach would return an internal error, it should now return `mc: <ERROR> Unable to make LDAP policy association. The specified user does not exist. (Specified user does not exist)`

## Types of changes

## Checklist:

### DIFF
--- a/internal/config/identity/ldap/ldap.go
+++ b/internal/config/identity/ldap/ldap.go
@@ -98,7 +98,7 @@ func (l *Config) GetValidatedDNForUsername(username string) (*xldap.DNSearchResu
 	// under a configured base DN in the LDAP directory.
 	validDN, isUnderBaseDN, err := l.GetValidatedUserDN(conn, username)
 	if err == nil && !isUnderBaseDN {
-		return nil, fmt.Errorf("Unable to find user DN: %w", err)
+		return nil, nil
 	}
 	return validDN, err
 }
--- a/internal/config/identity/ldap/ldap.go
+++ b/internal/config/identity/ldap/ldap.go
@@ -98,6 +98,7 @@ func (l *Config) GetValidatedDNForUsername(username string) (*xldap.DNSearchResu
 	// under a configured base DN in the LDAP directory.
 	validDN, isUnderBaseDN, err := l.GetValidatedUserDN(conn, username)
 	if err == nil && !isUnderBaseDN {
+		// Not under any configured base DN, so treat as not found.
 		return nil, nil
 	}
 	return validDN, err


internal/config/identity/ldap/ldap.go
```diff
@@ -98,7 +98,7 @@ func (l *Config) GetValidatedDNForUsername(username string) (*xldap.DNSearchResu
 	// under a configured base DN in the LDAP directory.
 	validDN, isUnderBaseDN, err := l.GetValidatedUserDN(conn, username)
 	if err == nil && !isUnderBaseDN {
-		return nil, fmt.Errorf("Unable to find user DN: %w", err)
```
harshavardhana, line 101: why is it not an error? and what happens to the caller if they rely on `err == nil` to make sure xldap.DNSearchResult is non-nil

taran-p: The return of GetValidatedDNForUsername on a non-existent username/DN is `nil, nil`, so caller should not be doing that. We treat users that are not under the search base DN as non-existent, so return the same `nil, nil`.

harshavardhana: @donatello PTAL not sure what is the context for returning errors here

donatello: This is correct - we are returning a user not found result as a `nil, nil` instead of a specific` error value like `nil, errNotFound`.

donatello (approved):

internal/config/identity/ldap/ldap.go
harshavardhana: So the question is - is the caller aware of this? That we will be returning nil results with nil error

Please add a comment @taran-p for future users.

taran-p: Added a comment for the change; the overall return behavior is already outlined in the last line of the doc comment

harshavardhana (approved):

PR #20033, opened 2024-07-03
feat: support batch replication prefix slice

jiuker:
feat: support batch job prefix with comma

## Description

feat: support batch replication prefix with comma

## How to test this PR?

testYaml:
V1verion
```yaml
replicate:
  apiVersion: v1
  # source of the objects to be replicated
  source:
    type: minio # valid values are "s3" or "minio"
    bucket: mytest
    prefix: object-prefix1  # 'PREFIX' is optional
    # If your source is the 'local' alias specified to 'mc batch start', then the 'endpoint' and 'credentials' fields are optional and can be omitted
    # Either the 'source' or 'remote' *must* be the "local" deployment
#    endpoint: "http://127.0.0.1:9000"
#    # path: "on|off|auto" # "on" enables path-style bucket lookup. "off" enables virtual host (DNS)-style bucket lookup. Defaults to "auto"
#    credentials:
#      accessKey: minioadmin # Required
#      secretKey: minioadmin # Required
#    # sessionToken: SESSION-TOKEN # Optional only available when rotating credentials are used
    snowball: # automatically activated if the source is local
      disable: false # optionally turn-off snowball archive transfer
#      batch: 100 # upto this many objects per archive
#      inmemory: true # indicates if the archive must be staged locally or in-memory
#      compress: false # S2/Snappy compressed archive
#      smallerThan: 5MiB # create archive for all objects smaller than 5MiB
#      skipErrs: false # skips any source side read() errors

  # target where the objects must be replicated
  target:
    type: minio # valid values are "s3" or "minio"
    bucket: mytest
    prefix: stage # 'PREFIX' is optional
    # If your source is the 'local' alias specified to 'mc batch start', then the 'endpoint' and 'credentials' fields are optional and can be omitted

    # Either the 'source' or 'remote' *must* be the "local" deployment
    endpoint: "http://127.0.0.1:9001"
    # path: "on|off|auto" # "on" enables path-style bucket lookup. "off" enables virtual host (DNS)-style bucket lookup. Defaults to "auto"
    credentials:
      accessKey: minioadmin
      secretKey: minioadmin
    # sessionToken: SESSION-TOKEN # Optional only available when rotating credentials are used

  # NOTE: All flags are optional
  # - filtering criteria only applies for all source objects match the criteria
  # - configurable notification endpoints
  # - configurable retries for the job (each retry skips successfully previously replaced objects)
  flags:
    filter:
      newerThan: "24h" # match objects newer than this value (e.g. 7d10h31s)
#      olderThan: "7d" # match objects older than this value (e.g. 7d10h31s)
#      createdAfter: "date" # match objects created after "date"
#      createdBefore: "date" # match objects created before "date"

      ## NOTE: tags are not supported when "source" is remote.
      # tags:
      #   - key: "name"
      #     value: "pick*" # match objects with tag 'name', with all values starting with 'pick'

      # metadata:
      #   - key: "content-type"
      #     value: "image/*" # match objects with 'content-type', with all values starting with 'image/'

#    notify:
#      endpoint: "https://notify.endpoint" # notification endpoint to receive job status events
#      token: "Bearer xxxxx" # optional authentication token for the notification endpoint
#
#    retry:
#      attempts: 10 # number of retries for the job before giving up
#      delay: "500ms" # least amount of delay between each retry
```
```output
$ mc ls minio9001/mytest -r
[2024-07-03 09:33:45 CST] 3.5KiB STANDARD object-prefix1/sftp.go
[2024-07-03 09:34:02 CST]   945B STANDARD object-prefix2/rlimit.go
[2024-07-03 09:33:45 CST] 3.5KiB STANDARD stage/object-prefix1/sftp.go
[2024-07-03 09:34:02 CST]   945B STANDARD stage/object-prefix2/rlimit.go
```
--- here is a bug for 
```
    snowball: # automatically activated if the source is local
      disable: false   <- if false will not use target prefix
```
V2 version:
```
replicate:
  apiVersion: v2
  # source of the objects to be replicated
  source:
    type: minio # valid values are "s3" or "minio"
    bucket: mytest
    prefix:
     - object-prefix1
     - object-prefix2  # 'PREFIX' is optional
    # If your source is the 'local' alias specified to 'mc batch start', then the 'endpoint' and 'credentials' fields are optional and can be omitted
    # Either the 'source' or 'remote' *must* be the "local" deployment
#    endpoint: "http://127.0.0.1:9000"
#    # path: "on|off|auto" # "on" enables path-style bucket lookup. "off" enables virtual host (DNS)-style bucket lookup. Defaults to "auto"
#    credentials:
#      accessKey: minioadmin # Required
#      secretKey: minioadmin # Required
#    # sessionToken: SESSION-TOKEN # Optional only available when rotating credentials are used
    snowball: # automatically activated if the source is local
      disable: true # optionally turn-off snowball archive transfer
#      batch: 100 # upto this many objects per archive
#      inmemory: true # indicates if the archive must be staged locally or in-memory
#      compress: false # S2/Snappy compressed archive
#      smallerThan: 5MiB # create archive for all objects smaller than 5MiB
#      skipErrs: false # skips any source side read() errors

  # target where the objects must be replicated
  target:
    type: minio # valid values are "s3" or "minio"
    bucket: mytest
    prefix: stage # 'PREFIX' is optional
    # If your source is the 'local' alias specified to 'mc batch start', then the 'endpoint' and 'credentials' fields are optional and can be omitted

    # Either the 'source' or 'remote' *must* be the "local" deployment
    endpoint: "http://127.0.0.1:9001"
    # path: "on|off|auto" # "on" enables path-style bucket lookup. "off" enables virtual host (DNS)-style bucket lookup. Defaults to "auto"
    credentials:
      accessKey: minioadmin
      secretKey: minioadmin
    # sessionToken: SESSION-TOKEN # Optional only available when rotating credentials are used

  # NOTE: All flags are optional
  # - filtering criteria only applies for all source objects match the criteria
  # - configurable notification endpoints
  # - configurable retries for the job (each retry skips successfully previously replaced objects)
  flags:
    filter:
      newerThan: "24h" # match objects newer than this value (e.g. 7d10h31s)
#      olderThan: "7d" # match objects older than this value (e.g. 7d10h31s)
#      createdAfter: "date" # match objects created after "date"
#      createdBefore: "date" # match objects created before "date"

      ## NOTE: tags are not supported when "source" is remote.
      # tags:
      #   - key: "name"
      #     value: "pick*" # match objects with tag 'name', with all values starting with 'pick'

      # metadata:
      #   - key: "content-type"
      #     value: "image/*" # match objects with 'content-type', with all values starting with 'image/'

#    notify:
#      endpoint: "https://notify.endpoint" # notification endpoint to receive job status events
#      token: "Bearer xxxxx" # optional authentication token for the notification endpoint
#
#    retry:
#      attempts: 10 # number of retries for the job before giving up
#      delay: "500ms" # least amount of delay between each retry
```

## Types of changes

## Checklist:


cmd/batch-expire.go
```diff
@@ -538,15 +540,34 @@ func (r *BatchJobExpire) Start(ctx context.Context, api ObjectLayer, job BatchJo
...
 
 	results := make(chan itemOrErr[ObjectInfo], workerSize)
-	if err := api.Walk(ctx, r.Bucket, r.Prefix, results, WalkOptions{
-		Marker:       lastObject,
-		LatestOnly:   false, // we need to visit all versions of the object to implement purge: retainVersions
-		VersionsSort: WalkVersionsSortDesc,
-	}); err != nil {
-		// Do not need to retry if we can't list objects on source.
-		return err
+	prefixs := strings.Split(r.Prefix, ",")
```
harshavardhana:
```suggestion
	prefixes := strings.Split(r.Prefix, ",")
```

cmd/batch-expire.go
```diff
@@ -538,15 +540,34 @@ func (r *BatchJobExpire) Start(ctx context.Context, api ObjectLayer, job BatchJo
...
+			for result := range prefixResultCh {
+				results <- result
+			}
+		}()
 	}
 
+	go func() {
+		wg.Wait()
+		xioutil.SafeClose(results)
+	}()
```
harshavardhana: We don't need to list them in parallel, we should list them individually in the specified order of entries.

harshavardhana (changes requested):

cmd/batch-expire.go
```diff
@@ -534,18 +535,34 @@ func (r *BatchJobExpire) Start(ctx context.Context, api ObjectLayer, job BatchJo
...
 	}
 
-	ctx, cancel := context.WithCancel(ctx)
-	defer cancel()
+	ctx, cancelCancel := context.WithCancelCause(ctx)
+	defer func() {
+		if context.Cause(ctx) == nil {
+			cancelCancel(nil)
+		}
+	}()
```
klauspost: The first error is sticky, so no need to make it complicated:
```suggestion
	defer cancelCancel(nil)
```

klauspost:
```
λ mc ls play/testbucket
[2024-07-03 11:27:36 CEST]     0B prefix,2/
```
Comma is a valid prefix character.

YAML has list/array types, why aren't we using that?

cmd/batch-expire.go
```diff
@@ -534,18 +535,34 @@ func (r *BatchJobExpire) Start(ctx context.Context, api ObjectLayer, job BatchJo
...
+		for _, prefix := range prefixes {
+			prefixResultCh := make(chan itemOrErr[ObjectInfo], workerSize)
+			err := api.Walk(ctx, r.Bucket, strings.TrimSpace(prefix), prefixResultCh, WalkOptions{
+				Marker:       lastObject,
+				LatestOnly:   false, // we need to visit all versions of the object to implement purge: retainVersions
+				VersionsSort: WalkVersionsSortDesc,
+			})
+			if err != nil {
+				xioutil.SafeClose(results)
+				cancelCancel(err)
```
klauspost: You close before you set the error, so you are racing:
```suggestion
				cancelCancel(err)
				xioutil.SafeClose(results)
```

jiuker:
> ```
> λ mc ls play/testbucket
> [2024-07-03 11:27:36 CEST]     0B prefix,2/
> ```
> 
> Comma is a valid prefix character.
> 
> YAML has list/array types, why aren't we using that?

If we add a new fields use slice/array. That is ok.
If we change this field type, I think no.
@klauspost 

@harshavardhana cc.

cmd/batch-expire.go
```diff
@@ -534,18 +535,34 @@ func (r *BatchJobExpire) Start(ctx context.Context, api ObjectLayer, job BatchJo
...
+				return
+			}
+			for result := range prefixResultCh {
+				results <- result
+			}
+		}
+		xioutil.SafeClose(results)
+	}()
```
klauspost, line 561: Below you need to look at

```
			case <-ctx.Done():
				// persist in-memory state immediately before exiting due to context cancellation.
				batchLogIf(ctx, ri.updateAfter(ctx, api, 0, job))
```
A canceled config is being sent to the saveConfig. Pretty sure that does nothing.

cmd/batch-expire.go
```diff
@@ -653,6 +670,11 @@ func (r *BatchJobExpire) Start(ctx context.Context, api ObjectLayer, job BatchJo
 			ObjectInfo: result.Item,
 		})
 	}
+	if context.Cause(ctx) != nil {
+		xioutil.SafeClose(expireCh)
+		xioutil.SafeClose(saverQuitCh)
```
klauspost: If the context is canceled the saver will exit anyway.

klauspost (changes requested):
This needs to be *seriously* vetted for error handling and and races.

I also see a `failed := true` I added that looks like it should have been `failed := false`.

cmd/batch-expire.go
jiuker: You can take a closer look at the diff of this code, and if it behaves differently than before, @klauspost

klauspost:
I am not an expert at YAML, but maybe a second field is the way to make it backwards compatible.

```YAML
    prefixes: 
     - object-prefix1/
     - object-prefix2/
```

cmd/batch-expire.go
klauspost: `context.Cause(ctx) != nil` implies that ctx is canceled... Then observe the code:

```
			case <-ctx.Done():
				// persist in-memory state immediately before exiting due to context cancellation.
				batchLogIf(ctx, ri.updateAfter(ctx, api, 0, job))
				return
			case <-saverQuitCh:
				// persist in-memory state immediately to disk.
				batchLogIf(ctx, ri.updateAfter(ctx, api, 0, job))
				return
```

jiuker: Yesh.

jiuker:
```go
package main

import (
	"context"
	"fmt"
)

func main() {
	pctx, cancel := context.WithCancel(context.Background())
	cancel()
	ctx, _ := context.WithCancelCause(pctx)
	fmt.Println(context.Cause(ctx))
}
```
context canceled
I thought `nil` before. @klauspost  (:

jiuker:
> I am not an expert at YAML, but maybe a second field is the way to make it backwards compatible.
> 
> ```yaml
>     prefixes: 
>      - object-prefix1/
>      - object-prefix2/
> ```

@harshavardhana cc

cmd/batch-expire.go
jiuker: good catch！

harshavardhana (changes requested):
Okay instead of comma separated, since it looks like , can be part of a valid object name. 

To keep it fool proof, introduce a new field that supports multiple prefixes on the source side.

```
prefix:
 - myprefix/
 - myprefix2/
```

Bump the version: field so ensure that we use v2 array for prefixes vs version '1' that supports single string.

harshavardhana (changes requested):
You should split the struct like this 

```
type versionStruct struct{
   APIVersion string
}

type BatchJobExpire struct { --> This is always the latest version (latest version doesn't have V1 or V2 added) only previous ones carry their versions
  versionStruct
}

type BatchJobExpireV1 struct {
  versionStruct
}
```

You first unmarshal the `versionStruct` then get the APIVersion once you get the APIVersion choose the right data structure to `unmarshal` into.

See how this is done in in cmd/format-erasure.go

```
// Returns format Erasure version after reading `format.json`, returns
// successfully the version only if the backend is Erasure.
func formatGetBackendErasureVersion(b []byte) (string, error) {
        meta := &formatMetaV1{}
        if err := json.Unmarshal(b, meta); err != nil {
                return "", err
        }
        if meta.Version != formatMetaVersionV1 {
                return "", fmt.Errorf(`format.Version expected: %s, got: %s`, formatMetaVersionV1, meta.Version)
        }
        if meta.Format != formatBackendErasure && meta.Format != formatBackendErasureSingle {
                return "", fmt.Errorf(`found backend type %s, expected %s or %s - to migrate to a supported backend visit https://min.io/docs/minio/linux/operations/install-deploy-manage/mig
rate-fs-gateway.html`, meta.Format, formatBackendErasure, formatBackendErasureSingle)
        }
        // Erasure backend found, proceed to detect version.
        format := &formatErasureVersionDetect{}
        if err := json.Unmarshal(b, format); err != nil {
                return "", err
        }
        return format.Erasure.Version, nil
}
```

jiuker:
> You should split the struct like this
> 
> ```
> type versionStruct struct{
>    APIVersion string
> }
> 
> type BatchJobExpire struct { --> This is always the latest version (latest version doesn't have V1 or V2 added) only previous ones carry their versions
>   versionStruct
> }
> 
> type BatchJobExpireV1 struct {
>   versionStruct
> }
> ```
> 
> You first unmarshal the `versionStruct` then get the APIVersion once you get the APIVersion choose the right data structure to `unmarshal` into.
> 
> See how this is done in in cmd/format-erasure.go
> 
> ```
> // Returns format Erasure version after reading `format.json`, returns
> // successfully the version only if the backend is Erasure.
> func formatGetBackendErasureVersion(b []byte) (string, error) {
>         meta := &formatMetaV1{}
>         if err := json.Unmarshal(b, meta); err != nil {
>                 return "", err
>         }
>         if meta.Version != formatMetaVersionV1 {
>                 return "", fmt.Errorf(`format.Version expected: %s, got: %s`, formatMetaVersionV1, meta.Version)
>         }
>         if meta.Format != formatBackendErasure && meta.Format != formatBackendErasureSingle {
>                 return "", fmt.Errorf(`found backend type %s, expected %s or %s - to migrate to a supported backend visit https://min.io/docs/minio/linux/operations/install-deploy-manage/mig
> rate-fs-gateway.html`, meta.Format, formatBackendErasure, formatBackendErasureSingle)
>         }
>         // Erasure backend found, proceed to detect version.
>         format := &formatErasureVersionDetect{}
>         if err := json.Unmarshal(b, format); err != nil {
>                 return "", err
>         }
>         return format.Erasure.Version, nil
> }
> ```

`cmd/batch-handlers.go` do that. @harshavardhana

cmd/batch-expire.go
```diff
@@ -286,24 +287,31 @@ type BatchJobExpire struct {
 	Rules           []BatchJobExpireFilter `yaml:"rules" json:"rules"`
 }
 
-var _ yaml.Unmarshaler = &BatchJobExpire{}
+// BatchJobExpireV2 represents configuration parameters for a batch expiration
+// job typically supplied in yaml form
+type BatchJobExpireV2 struct {
```
harshavardhana: You can keep the struct as BatchJobExpire and move the existing one into BatchJobExpireV1

harshavardhana (changes requested):

cmd/batch-expire.go
```diff
@@ -534,18 +542,29 @@ func (r *BatchJobExpire) Start(ctx context.Context, api ObjectLayer, job BatchJo
...
+				cancelCause(err)
+				xioutil.SafeClose(results)
+				return
+			}
+			for result := range prefixResultCh {
+				results <- result
+			}
+		}
+		xioutil.SafeClose(results)
+	}()
```
harshavardhana: I think we should simplify this, by allowing support for Walk() to take multiple prefixes, instead to having a goroutine like this.

harshavardhana (changes requested):

harshavardhana (approved):
Walk() changes can be done later

cmd/batch-expire.go
```diff
@@ -286,24 +287,31 @@ type BatchJobExpire struct {
 	Rules           []BatchJobExpireFilter `yaml:"rules" json:"rules"`
 }
 
-var _ yaml.Unmarshaler = &BatchJobExpire{}
+// BatchJobExpireV2 represents configuration parameters for a batch expiration
+// job typically supplied in yaml form
+type BatchJobExpireV2 struct {
+	BatchJobExpireV1
```
krisis: @jiuker I see that you have used struct embedding in multiple places in this PR. While it saves typing common fields it has the following downsides.
- Shadowing of struct fields makes it harder to read surrounding code for someone new to this feature (maintainability)
- Shadowing affects how various libraries marshal/unmarshal these struct values; It's best to avoid surprises.

I suggest we use separate types for job v1 and v2 without any struct embedding for the sake of reducing repeated fields.

cmd/batch-handlers.go
```diff
@@ -70,10 +70,86 @@ type BatchJobRequest struct {
...
+		fallthrough
+	case "v2":
+		if err = yaml.Unmarshal(buf, job); err != nil {
+			return nil, err
+		}
+	}
+	return
+}
+
+func batchJobRequestMigrateV1ToV2(buf []byte) ([]byte, error) {
```
krisis: You can directly return *BatchJobRequestV2 value after migrating from a v1 value. You don't have to marshal it to only unmarshal it as v2 again.
```suggestion
func batchJobRequestMigrateV1ToV2(buf []byte)(*BatchJobRequestV2, error) {
```

cmd/batch-handlers.go
```diff
@@ -70,10 +70,86 @@ type BatchJobRequest struct {
...
+	if jobV1.KeyRotate != nil {
+		jobV2.KeyRotate = jobV1.KeyRotate
+	}
+	if jobV1.Expire != nil {
+		jobV2.Expire = &BatchJobExpireV2{
+			BatchJobExpireV1: *jobV1.Expire,
+			Prefix:           []string{jobV1.Expire.Prefix}, // string into []string
+		}
+	}
+	return yaml.Marshal(jobV2)
```
krisis:
```suggestion
	return jobV2
```

cmd/batch-handlers.go
```diff
@@ -70,10 +70,86 @@ type BatchJobRequest struct {
...
+	}
+	return
+}
+
+func batchJobRequestMigrateV1ToV2(buf []byte) ([]byte, error) {
+	jobV1 := &BatchJobRequest{}
+	if err := yaml.Unmarshal(buf, jobV1); err != nil {
+		return nil, err
+	}
+	jobV2 := &BatchJobRequestV2{BatchJobRequest: *jobV1}
```
krisis: This is an example of how quickly things can become unwieldly with struct embedding in this case. Here you are copying `jobv2.BatchJobRequest.Replicate` to `jobv2.Replicate`. Before we realize this can become source for bugs in the future.

krisis (commented):
Summary:
- We don't gain anything from struct embedding; Let's use separate types with separate fields
- We need to handle v1 job request when we resume batch jobs after an upgrade to a version with this change.

cmd/batch-expire_test.go
```diff
@@ -68,4 +69,57 @@ expire: # Expire objects that match a condition
 	if err != nil {
 		t.Fatal("Failed to parse batch-job-expire yaml", err)
 	}
+	if !reflect.DeepEqual(job.Expire.Prefix.F(), []string{"myprefix"}) {
```
harshavardhana: Use slices.Equal()

cmd/batch-handlers.go
```diff
@@ -2232,3 +2255,30 @@ func lookupStyle(s string) miniogo.BucketLookupType {
 	}
 	return lookup
 }
+
+// BatchJobPrefix - to support prefix field yaml unmarshalling with string or slice string
```
harshavardhana: Can you add unit tests for this ?

harshavardhana (changes requested):
Let's also add functional tests this time, so that these functionalities are tested.

harshavardhana (approved):

harshavardhana:
PTAL @shtripat

cmd/batch-expire.go
```diff
@@ -535,18 +536,29 @@ func (r *BatchJobExpire) Start(ctx context.Context, api ObjectLayer, job BatchJo
 		return err
 	}
 
-	ctx, cancel := context.WithCancel(ctx)
-	defer cancel()
+	ctx, cancelCancel := context.WithCancelCause(ctx)
```
shtripat:
```suggestion
	ctx, cancelCause := context.WithCancelCause(ctx)
```
lets call it `cancelCause` like other place

cmd/batch-handlers.go
shtripat:
```suggestion
// BatchJobPrefix - to support prefix field yaml unmarshalling with string or slice of strings
```

cmd/batch-handlers.go
```diff
@@ -2232,3 +2255,30 @@ func lookupStyle(s string) miniogo.BucketLookupType {
...
+type BatchJobPrefix []string
+
+var _ yaml.Unmarshaler = &BatchJobPrefix{}
+
+// UnmarshalYAML - to support prefix field yaml unmarshalling with string or slice string
```
shtripat:
```suggestion
// UnmarshalYAML - to support prefix field yaml unmarshalling with string or slice of strings
```

cmd/batch-handlers.go
```diff
@@ -2232,3 +2255,30 @@ func lookupStyle(s string) miniogo.BucketLookupType {
...
+	// try string
+	tmpStr := ""
+	if err := value.Decode(&tmpStr); err == nil {
+		*b = []string{tmpStr}
+		return nil
+	}
+	return fmt.Errorf("unable to decode %s", value.Value)
+}
+
+// F - return prefix as slice
```
shtripat:
```suggestion
// F - return prefix(es) as slice
```

shtripat (commented):
few minor ones

shtripat (approved):

harshavardhana (approved):

PR #20047, opened 2024-07-08
list: Hide objects/versions with pending/failed replicated deletion 

vadmeste:

## Description
In regular listing, this commit will avoid showing an object when its
latest version has a pending or a failed deletion. in replicated setup.
It will also avoid showing older versions in the same case.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/metacache-entries.go
+++ b/cmd/metacache-entries.go
@@ -532,6 +532,9 @@ func (m *metaCacheEntriesSorted) fileInfoVersions(bucket, prefix, delimiter, aft
 			}
 
 			for _, version := range fiVersions {
+				if !version.VersionPurgeStatus().Empty() {
+					continue
+				}
 				versioned := vcfg != nil && vcfg.Versioned(entry.name)
 				versions = append(versions, version.ToObjectInfo(bucket, entry.name, versioned))
 			}
@@ -593,7 +596,7 @@ func (m *metaCacheEntriesSorted) fileInfos(bucket, prefix, delimiter string) (ob
 			}
 
 			fi, err := entry.fileInfo(bucket)
-			if err == nil {
+			if err == nil && fi.VersionPurgeStatus().Empty() {
 				versioned := vcfg != nil && vcfg.Versioned(entry.name)
 				objects = append(objects, fi.ToObjectInfo(bucket, entry.name, versioned))
 			}
--- a/docs/bucket/replication/delete-replication.sh
+++ b/docs/bucket/replication/delete-replication.sh
@@ -52,7 +52,9 @@ export MINIO_ROOT_USER="minioadmin"
 export MINIO_ROOT_PASSWORD="minioadmin"
 
 ./minio server --address ":9001" /tmp/xl/1/{1...4}/ 2>&1 >/tmp/dc1.log &
+pid1=$!
 ./minio server --address ":9002" /tmp/xl/2/{1...4}/ 2>&1 >/tmp/dc2.log &
+pid2=$!
 
 sleep 3
 
@@ -68,6 +70,8 @@ export MC_HOST_myminio2=http://minioadmin:minioadmin@localhost:9002
 ./mc version enable myminio2/testbucket/
 
 ./mc replicate add myminio1/testbucket --remote-bucket http://minioadmin:minioadmin@localhost:9002/testbucket/ --priority 1
+
+# Test replication of delete markers and permanent deletes
 
 ./mc cp README.md myminio1/testbucket/dir/file
 ./mc cp README.md myminio1/testbucket/dir/file
@@ -108,6 +112,34 @@ out=$(diff -qpruN /tmp/myminio1.txt /tmp/myminio2.txt)
 ret=$?
 if [ $ret -ne 0 ]; then
 	echo "BUG: expected no missing entries after replication: $out"
+	exit 1
+fi
+
+# Test listing of non replicated permanent deletes
+
+set -x
+
+./mc mb myminio1/foobucket/ myminio2/foobucket/ --with-versioning
+./mc replicate add myminio1/foobucket --remote-bucket http://minioadmin:minioadmin@localhost:9002/foobucket/ --priority 1
+./mc cp README.md myminio1/foobucket/dir/file
+
+versionId="$(./mc ls --json --versions myminio1/foobucket/dir/ | jq -r .versionId)"
+
+kill ${pid2} && wait ${pid2} || true
+
+aws s3api --endpoint-url http://localhost:9001 --profile minioadmin delete-object --bucket foobucket --key dir/file --version-id "$versionId"
+
+out="$(./mc ls myminio1/foobucket/dir/)"
+if [ "$out" != "" ]; then
+	echo "BUG: non versioned listing should not show pending/failed replicated delete:"
+	echo "$out"
+	exit 1
+fi
+
+out="$(./mc ls --versions myminio1/foobucket/dir/)"
+if [ "$out" != "" ]; then
+	echo "BUG: versioned listing should not show pending/failed replicated deletes:"
+	echo "$out"
 	exit 1
 fi
 


harshavardhana:
Why do we need another flag here ? Version purge status is as good as delete marker - why would that not work?

vadmeste:
> Why do we need another flag here ? Version purge status is as good as delete marker - why would that not work?

Metadata does not seem quickly accessible in the listing code. I think xl header were added to reduce CPU usage during listing, let me check with @klauspost

klauspost (changes requested):
Flags are set at save time, so if you add another and it must work with old data, you must:

A) Bump `xlHeaderVersion`
B) Send full metadata to `(*xlMetaV2VersionHeader) unmarshalV()`
C) Decode the metadata for relevant types and fix up the header before returning.

You can AFAICT do a partial decode into a struct like this:

```Go
//msgp:encode ignore purgeStatusV2Object
//msgp:marshal ignore purgeStatusV2Object
//msgp:decode ignore purgeStatusV2Object
//msgp:size ignore purgeStatusV2Object
type purgeStatusV2Object struct {
	V2Obj struct {
		MetaSys struct {
			PS *string `msg:"x-minio-internal-purgestatus"`
		}
	}
}
```

(adjust to also handle delete markers if it applies to them)

Needless to say this will still be quite CPU intensive. After all the header is there to avoid decoding the metadata.

Of course, you can reuse the existing flag and not bump the headerVersion, but then you would only fix this for new objects.

klauspost:
If this only applies to objects, it would be something like this:
```Go
func (x *xlMetaV2VersionHeader) unmarshalV(v uint8, bts, meta []byte) (o []byte, err error) {
	switch v {
	case 1:
		return x.unmarshalV1(bts)
	case 2:
		x2 := xlMetaV2VersionHeaderV2{xlMetaV2VersionHeader: x}
		return x2.UnmarshalMsg(bts)
	case 3:
		if o, err = x.UnmarshalMsg(bts); err != nil || x.Type != ObjectType {
			return o, err
		}
		// Add new flag
		var tmp purgeStatusV2Object
		_, err := tmp.UnmarshalMsg(meta)
		if err != nil {
			return o, nil
		}
		if tmp.V2Obj.MetaSys.PS != nil {
			x.Flags |= xlFlagIsRemoving
		}
		return o, nil
	case xlHeaderVersion:
		return
	}
	return bts, fmt.Errorf("unknown xlHeaderVersion: %d", v)
}
```

... and you would need to modify the callers to send the metadata, though it seems available at all call sites.

harshavardhana:
We already have a way to handle this, adding more things and expecting new objects with this behavior is not very useful. 

Metadata is definitely available AFAIK, how else will we know etag, size and mtime about the object?

klauspost:
Yeah, if we are only talking about listing, we always decode full metadata anyway at some point.

You can add it, so it gets written to new objects, if you can see rejections happening easier (and you should bump the header version anyway). That would allow checking the flag before decode... but you would still need to check the decoded data, since older versions will not have the flag.

That is the "cheap" method. Almost similar to EC - but the main issue is that we would need to compare the header version as well to know if the flag is reliable (or have an "inverse" flag `xlFlagIsNotRemoving`).

vadmeste:
By the way, the new flag will work once you remove a version, this will add purge-status in the metadata and update the header flags.

I initially thought this concerns a tiny subset of versions, which are the deleted versions but not replicated yet. However this fix is also important for deployments that have ongoing struggle to do a proper replication due to infrastructure issue. So I will use Klaus suggestion here.

harshavardhana (changes requested):
I still think this doesn't warrant a xl.meta header change, this is not downgrade friendly and we cannot have such frequent changes here. 

Either we have to do this once properly so that we can extend this to future. 

In this case however we have enough metadata at some point already it is not correct make xl.meta changes just to cater for Listing()  highly unlikely corner cases.

we are basically making everyone pay the price here which is not needed.

harshavardhana:
Also I am unable to reproduce this with the master branch, can you share the reproducer @vadmeste ?

vadmeste:
@harshavardhana this is the reproducer, just make sure you have any file called 'file' for testing

```
killall minio

rm -rf /tmp/r/

make || exit 0

./minio server --address :9001 --console-address :10000 /tmp/r/1/1/{1...4} /tmp/r/1/2/{1...4}/ &
./minio server --address :9002 --console-address :10001 /tmp/r/2/1/{1...4} /tmp/r/2/2/{1...4}/ &

sleep 3

mc alias set myminio1 http://localhost:9001 minioadmin minioadmin --api=s3v4
mc alias set myminio2 http://localhost:9002 minioadmin minioadmin --api=s3v4

mc ready myminio1/
mc ready myminio2/

mc admin replicate add myminio1/ myminio2/
sleep 1

mc mb myminio1/testbucket/
sleep 1

mc cp file myminio1/testbucket/

vid=$(mc ls myminio1/testbucket/ --versions --json | jq -r .versionId)

ps aux | grep -- "--address :9002" | awk '{ print $2 }' | xargs kill  

mc rm --version-id $vid myminio1/testbucket/file

sleep 5

set -x
mc ls -q --versions myminio1/testbucket/
mc ls -q myminio1/testbucket/
set +x
```

this test script simply runs two independent MinIO, configure site replication, create a versioned bucket, upload a file, then kill the second cluster before removing the uploaded file (creating a delete marker), then you can see that `mc ls`  which is not versioned simply lists that file.

harshavardhana (approved):

harshavardhana:
> @harshavardhana this is the reproducer, just make sure you have any file called 'file' for testing
> 
> ```
> killall minio
> 
> rm -rf /tmp/r/
> 
> make || exit 0
> 
> ./minio server --address :9001 --console-address :10000 /tmp/r/1/1/{1...4} /tmp/r/1/2/{1...4}/ &
> ./minio server --address :9002 --console-address :10001 /tmp/r/2/1/{1...4} /tmp/r/2/2/{1...4}/ &
> 
> sleep 3
> 
> mc alias set myminio1 http://localhost:9001 minioadmin minioadmin --api=s3v4
> mc alias set myminio2 http://localhost:9002 minioadmin minioadmin --api=s3v4
> 
> mc ready myminio1/
> mc ready myminio2/
> 
> mc admin replicate add myminio1/ myminio2/
> sleep 1
> 
> mc mb myminio1/testbucket/
> sleep 1
> 
> mc cp file myminio1/testbucket/
> 
> vid=$(mc ls myminio1/testbucket/ --versions --json | jq -r .versionId)
> 
> ps aux | grep -- "--address :9002" | awk '{ print $2 }' | xargs kill  
> 
> mc rm --version-id $vid myminio1/testbucket/file
> 
> sleep 5
> 
> set -x
> mc ls -q --versions myminio1/testbucket/
> mc ls -q myminio1/testbucket/
> set +x
> ```
> 
> this test script simply runs two independent MinIO, configure site replication, create a versioned bucket, upload a file, then kill the second cluster before removing the uploaded file (creating a delete marker), then you can see that `mc ls` which is not versioned simply lists that file.

Please add this as a test

```shell
#!/bin/bash

pkill minio
make install

rm -rf /tmp/r/

make || exit 0

export MINIO_LOG_DIR=/tmp/logs

minio server --address :9001 --console-address :10000 /tmp/r/1/1/{1...4} /tmp/r/1/2/{1...4}/ &
pid1=$!
minio server --address :9002 --console-address :10001 /tmp/r/2/1/{1...4} /tmp/r/2/2/{1...4}/ &
pid2=$!

export MC_HOST_myminio1=http://minioadmin:minioadmin@localhost:9001
export MC_HOST_myminio2=http://minioadmin:minioadmin@localhost:9002

mc ready myminio1/
mc ready myminio2/

mc admin replicate add myminio1/ myminio2/
sleep 1

mc mb myminio1/testbucket/
sleep 1

mc cp file myminio1/testbucket/

vid=$(mc ls myminio1/testbucket/ --versions --json | jq -r .versionId)

kill ${pid2}

sleep 1

mc rm --version-id $vid myminio1/testbucket/file

sleep 5

set -x
mc ls -q --versions myminio1/testbucket/
mc ls -q myminio1/testbucket/

mc stat --vid ${vid} myminio1/testbucket/file
mc stat ${vid} myminio1/testbucket/file

set +x
```

poornas (approved):

vadmeste:
@harshavardhana I added the equivalent of your test to an existing script, I am also hiding non current versions with a non replicated delete

PR #20062, opened 2024-07-09
Satisfy `aws:SecureTransport` policy behind reverse proxy

### LINKED ISSUE #20061: S3 key "aws:SecureTransport" isn't properly set with Minio SSL Setup
Hi, 

I have configured properly a SSL setup for the minio server. 
We have a simply apache reverse proxy in front of the minio server (+ console) which handles the SSL termination. 
The proxy does locally communicate with HTTP with the minio server, however that shouldn't affect the `aws:SecureTransport` context var. 

https://github.com/minio/console/blob/649c3d74b8a64b102a283e2d493ba8d07faf30cc/api/user_session.go#L112
It should be set to true, as long as the `MINIO_SERVER_URL` starts with `https`. Our url looks like this : `https://<server>:9000`. The `MINIO_BROWSER_REDIRECT_URL`  looks the same just no portnumber at the end (443). 

Given the "only SSL" IAM Policy below it fails to properly work, which leads to my understanding 

## Expected Behavior
When setting the following policy, I expect to be able to access all buckets regularly, given I have the proper SSL setup. Meaning the `aws:SecureTransport` variable should always be set true. 

~~~
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "Allow SSL communication only",
            "Effect": "Deny",
            "Action": [
                "s3:*"
            ],
            "Resource": [
                "arn:aws:s3:::*"
            ],
            "Condition": {
                "Bool": {
                    "aws:SecureTransport": [
                        "false"
                    ]
                }
            }
        }
    ]
}
~~~

## Current Behavior
When enabling that policy for a user it false denies all s3 access / rights. 

Any idea how to progress with debugging or analyzing the problem ? I don't know how to actually check the s3 context keys. 

## Your Environment

* MinIO version used (`minio --version`): minio version RELEASE.2024-02-06T21-36-22Z (commit-id=980fb5e2ab3674287207c9388bb8fda7a142cc64)
Runtime: go1.21.6 linux/amd64
* Server setup and configuration: Linux Ubuntu Server 22.02

ramondeklein:
This PR addresses the issue in #20061.

Use the following setup to reproduce this issue:
- Set up NGINX with proper SSL termination on https://minio.example.com that forwards traffic to http://localhost:9000 (for MinIO) and http://localhost:9001 (for console on https://minio.example.com/minio/ui). Make sure the NGINX configuration follows our [guidelines](https://min.io/docs/minio/linux/integrations/setup-nginx-proxy-with-minio.html).
- Created the following policy and assigned it to user `test`.
  ```json
  {
      "Version": "2012-10-17",
      "Statement": [
          {
              "Effect": "Allow",
              "Action": ["s3:*"],
              "Resource": ["arn:aws:s3:::*"]
          },
          {
              "Sid": "Allow SSL communication only",
              "Effect": "Deny",
              "Action": ["s3:*"],
              "Resource": ["arn:aws:s3:::*"],
              "Condition": {
                  "Bool": {
                      "aws:SecureTransport": [ "false" ]
                  }
              }
          }
      ]
  }
  ```
- Register `mc alias set test https://minio.example.com/ test minio123`.

Running `mc ls test` shows `mc: <ERROR> Unable to list folder. Access Denied.`. When removing the second statement it works fine. Because MinIO is using TLS, I would expect it to succeed.

After code inspection, I found the following ([`bucket-policy.go:131`](https://github.com/minio/minio/blob/380233d646b3e02e8c9f389b8bb02d131d5a06c9/cmd/bucket-policy.go#L131)):
```go
"SecureTransport":  {strconv.FormatBool(r.TLS != nil)},
```
MinIO checks if the request (as MinIO receives it) is using TLS, but it's not checking the headers that may have been added by the reverse proxy. This PR checks if the forwarding headers are set and if so, these take precedence.

### DIFF
--- a/cmd/bucket-policy.go
+++ b/cmd/bucket-policy.go
@@ -125,10 +125,15 @@ func getConditionValues(r *http.Request, lc string, cred auth.Credentials) map[s
 		authtype = "POST"
 	}
 
+	secureTransport := r.TLS != nil
+	if forwardedScheme := handlers.GetSourceScheme(r); forwardedScheme != "" {
+		secureTransport = forwardedScheme == "https"
+	}
+
 	args := map[string][]string{
 		"CurrentTime":      {currTime.Format(time.RFC3339)},
 		"EpochTime":        {strconv.FormatInt(currTime.Unix(), 10)},
-		"SecureTransport":  {strconv.FormatBool(r.TLS != nil)},
+		"SecureTransport":  {strconv.FormatBool(secureTransport)},
 		"SourceIp":         {handlers.GetSourceIPRaw(r)},
 		"UserAgent":        {r.UserAgent()},
 		"Referer":          {r.Referer()},


ramondeklein:
@harshavardhana I think we should take the forwarding headers into account for this check, but maybe it needs to be so strict that even traffic between the reverse proxy and MinIO needs to use TLS too. In that case we should document this behaviour.

A (not very realistic, but possible) configuration could be that MinIO is exposed via HTTP by NGINX, while using TLS for the connection between NGINX and MinIO. In that case the forwarding headers should also be checked to ensure that the condition isn't met.

harshavardhana (approved):

harshavardhana:
We were aware of this and it was never done on purpose since these headers can be spoofed easily.

However it's okay.

klauspost:
I don't really like that we are setting this as `true` when requests are sent as plaintext to the server. This is not a definite veto, but I think we should be careful with this.

ramondeklein:
@harshavardhana I think I agree with @klauspost, because it's easy to spoof the header (didn't think about that), so anyone can bypass this condition by adding the `X-Forwarded-Scheme` header. I rather have this issue then the issue: "I can access via HTTP even though `aws:SecureTransport` was required".

I guess the proper solution is to only expose MinIO via HTTPS.

ramondeklein:
We could add an environment variable that tells us that the `X-Forwarded-Scheme` header is safe to use and only ook at the headers if the variable is set. But I think this is an edge-case that we should ignore.

harshavardhana:
Correct that's why the SecureTransport was kept the way it is, IMO. We do not need to honor `x-forwarded-scheme` here since it makes things vulnerable.

harshavardhana:
I would suggest we close the issue and keep this PR for future purposes if needed, under a branch in your fork @ramondeklein

krisis (approved):

PR #20067, opened 2024-07-10
Clarify post policy error message

klauspost:
## Description

It is not really clear that the listed keys are missing.

Clarify the error

## Types of changes
- [x] UX

### DIFF
--- a/cmd/postpolicyform.go
+++ b/cmd/postpolicyform.go
@@ -364,7 +364,7 @@ func checkPostPolicy(formValues http.Header, postPolicyForm PostPolicyForm) erro
 		for key := range checkHeader {
 			logKeys = append(logKeys, key)
 		}
-		return fmt.Errorf("Each form field that you specify in a form (except %s) must appear in the list of conditions.", strings.Join(logKeys, ", "))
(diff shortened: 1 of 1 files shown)


jiuker:
Let me do a test. This error should the same as `AWS` return

klauspost:
@jiuker Not sure about that. The message should just be readable by a human as per https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html

> The error message contains a generic description of the error condition in English. It is intended for a human audience. Simple programs display the message directly to the end user if they encounter an error condition they don't know how or don't care to handle. Sophisticated programs with more exhaustive error handling and proper internationalization are more likely to ignore the error message.

jiuker (approved):

jiuker:
```
<Error><Code>AccessDenied</Code><Message>Invalid according to Policy: Extra input fields: content-type</Message><RequestId>VP4SV8KE1PE0GHX5</RequestId><HostId>P/9pS/qiWhLYYpctADktIY28o5s/27I7eigmDM1IqX72cbykWjKIIt42RsYQJ0DipyA7XxXNigU=</HostId></Error>
```

harshavardhana (approved):

harshavardhana (approved):

PR #20074, opened 2024-07-11
Fix ListObjects aborting after 3 minute on async request

klauspost:
## Description

When async listing, if the first request does not return within 3 minutes, it is stopped, since it isn't being kept alive.

Keep updating `lastHandout` while we are waiting for the initial request to be fulfilled.

## How to test this PR?

Insert huge delays (1 second+ per entry) in returning entries from storage.

## Types of changes


cmd/metacache-server-pool.go
```diff
@@ -419,6 +419,26 @@ func (z *erasureServerPools) listAndSave(ctx context.Context, o *listPathOptions
 		o.debugln("listAndSave: listing", o.ID, "finished with ", err)
 	}(*o)
 
+	// Keep alive while initial request is running.
+	go func(ctx context.Context, rpc *peerRESTClient, meta metacache) {
+		// Continuously update while we wait.
+		t := time.NewTicker(metacacheMaxClientWait / 10)
+		defer t.Stop()
+		select {
```
marktheunissen: @klauspost Is this missing a `for {` wrapping the select, to keep it running continuously?

harshavardhana: yes

klauspost: Well, since I copy+pasted....

klauspost: Good spot! Moving the shared code to a function.

klauspost:
Tested with `time.Sleep(1000 * time.Millisecond)` when returning entries. Listing is kept alive even if there are 499 (10m timeouts) from mc.

harshavardhana (approved):

cmd/metacache-server-pool.go
```diff
@@ -449,5 +440,10 @@ func (z *erasureServerPools) listAndSave(ctx context.Context, o *listPathOptions
 		xioutil.SafeClose(saveCh)
 	}()
 
-	return filteredResults()
+	entries, err = filteredResults()
+	if err == nil && err != io.EOF {
```
vadmeste: isn't this OR rather than AND ?

harshavardhana:
```suggestion
	if err == nil || err != io.EOF {
```

klauspost: Yeah, it is a bit superfluous. Probably it should be `|| err == io.EOF` just to double check.

klauspost: Let's check it.. Adds some extra safety.

vadmeste (approved):

harshavardhana (approved):

PR #20136, opened 2024-07-23
batch-exp: Avoid failure while expiring an object with too many versions

vadmeste:

## Description
An object that has too many versions can take time to have all its versions to be removed. At the same time, there is a drive IO deadline, which does not take the number of versions into consideration. Deleting a version can involve moving the version data-dir to the trash, which is a costly operation and it can easily accumulate to more than the standard drive deadline.

This PR will scale the deadline with the number of versions per object.

## Motivation and Context
Avoid batch expiration failure when it wants to remove too many versions of a single object

## Types of changes

## Checklist:

### DIFF
--- a/cmd/xl-storage.go
+++ b/cmd/xl-storage.go
@@ -1079,32 +1079,36 @@ func (s *xlStorage) deleteVersions(ctx context.Context, volume, path string, fis
 		return err
 	}
 
-	var legacyJSON bool
-	buf, _, err := s.readAllData(ctx, volume, volumeDir, pathJoin(volumeDir, path, xlStorageFormatFile))
-	if err != nil {
-		if !errors.Is(err, errFileNotFound) {
-			return err
-		}
+	s.RLock()
+	legacy := s.formatLegacy
+	s.RUnlock()
 
-		s.RLock()
-		legacy := s.formatLegacy
-		s.RUnlock()
+	var legacyJSON bool
+	buf, err := xioutil.WithDeadline[[]byte](ctx, globalDriveConfig.GetMaxTimeout(), func(ctx context.Context) ([]byte, error) {
+		buf, _, err := s.readAllData(ctx, volume, volumeDir, pathJoin(volumeDir, path, xlStorageFormatFile))
+		if err != nil && !errors.Is(err, errFileNotFound) {
+			return nil, err
+		}
 		if legacy {
 			buf, _, err = s.readAllData(ctx, volume, volumeDir, pathJoin(volumeDir, path, xlStorageFormatFileV1))
 			if err != nil {
-				return err
+				return nil, err
 			}
 			legacyJSON = true
 		}
-	}
 
-	if len(buf) == 0 {
-		if errors.Is(err, errFileNotFound) && !skipAccessChecks(volume) {
-			if aerr := Access(volumeDir); aerr != nil && osIsNotExist(aerr) {
-				return errVolumeNotFound
+		if len(buf) == 0 {
+			if errors.Is(err, errFileNotFound) && !skipAccessChecks(volume) {
+				if aerr := Access(volumeDir); aerr != nil && osIsNotExist(aerr) {
+					return nil, errVolumeNotFound
+				}
 			}
+			return nil, errFileNotFound
 		}
-		return errFileNotFound
+		return buf, nil
+	})
+	if err != nil {
+		return err
 	}
 
 	if legacyJSON {
@@ -1178,10 +1182,7 @@ func (s *xlStorage) DeleteVersions(ctx context.Context, volume string, versions 
 			errs[i] = ctx.Err()
 			continue
 		}
-		w := xioutil.NewDeadlineWorker(globalDriveConfig.GetMaxTimeout())
-		if err := w.Run(func() error { return s.deleteVersions(ctx, volume, fiv.Name, fiv.Versions...) }); err != nil {
-			errs[i] = err
-		}
+		errs[i] = s.deleteVersions(ctx, volume, fiv.Name, fiv.Versions...)
 		diskHealthCheckOK(ctx, errs[i])
 	}
 
@@ -1212,7 +1213,7 @@ func (s *xlStorage) diskAlmostFilled() bool {
 	return (float64(info.Free)/float64(info.Used)) < almostFilledPercent || (float64(info.FreeInodes)/float64(info.UsedInodes)) < almostFilledPercent
 }
 
-func (s *xlStorage) moveToTrash(filePath string, recursive, immediatePurge bool) (err error) {
+func (s *xlStorage) moveToTrashNoDeadline(filePath string, recursive, immediatePurge bool) (err error) {
 	pathUUID := mustGetUUID()
 	targetPath := pathutil.Join(s.drivePath, minioMetaTmpDeletedBucket, pathUUID)
 
@@ -1265,8 +1266,14 @@ func (s *xlStorage) moveToTrash(filePath string, recursive, immediatePurge bool)
 			}
 		}
 	}
-
 	return nil
+}
+
+func (s *xlStorage) moveToTrash(filePath string, recursive, immediatePurge bool) (err error) {
+	w := xioutil.NewDeadlineWorker(globalDriveConfig.GetMaxTimeout())
+	return w.Run(func() (err error) {
+		return s.moveToTrashNoDeadline(filePath, recursive, immediatePurge)
+	})
 }
 
 // DeleteVersion - deletes FileInfo metadata for path at `xl.meta`. forceDelMarker


cmd/xl-storage.go
```diff
@@ -1180,7 +1180,10 @@ func (s *xlStorage) DeleteVersions(ctx context.Context, volume string, versions
 			errs[i] = ctx.Err()
 			continue
 		}
-		w := xioutil.NewDeadlineWorker(globalDriveConfig.GetMaxTimeout())
+		// Scale the deadline with the number of versions for this object since an object
+		// can contain thousands of versions (though not recommended) and removing a version
+		// can involve moving data-dir to the trash.
+		w := xioutil.NewDeadlineWorker(globalDriveConfig.GetMaxTimeout() + time.Duration(len(fiv.Versions))*time.Second)
```
harshavardhana: this value is too much this is not how it happens internally this needs to be fixed to include only to "specific" object names.

Because deleting X versions of an object is mostly atomic i.e it purges at instant

harshavardhana (changes requested):

cmd/xl-storage.go
vadmeste: the purge is not instant if versions have data-dir. deleteVersions() will move data-dir to the trash folder and that sometimes takes a lot of time, more than the configured deadline (30 seconds is the default one). This means if one object has 10k versions, then this function will fail with a deadline. That is why I wanted to scale it depending on the number of versions of the object.

harshavardhana: yeah but that is literally forever, any drive that takes that long is dead already, there is no point performing this operation - we timeout and let other drives complete. 

This is not the place the optimize TBH.

harshavardhana: we have no reason to hold the the caller this long, a drive takes 30secs to rename you need to throw that drive. 

They are free to configure higher MaxTimeout per drive but this not correct we adding arbitrary number of seconds.

vadmeste: This use case is not a single rename, it is thousands of rename(), users will also have to take an arbitrary number.

It looks to me that a more comprehensive fix is to do a deadline per function inside deleteVersions(), or maybe a better option is to add new API to deadline worker to delay triggering the deadline

harshavardhana: that's what we are doing here

```
// DeleteVersions deletes slice of versions, it can be same object                                                                                            
// or multiple objects.                                                                                                                                       
func (s *xlStorage) DeleteVersions(ctx context.Context, volume string, versions []FileInfoVersions, opts DeleteOptions) []error {
        errs := make([]error, len(versions))

        for i, fiv := range versions {
                if contextCanceled(ctx) {
                        errs[i] = ctx.Err()
                        continue
                }
                w := xioutil.NewDeadlineWorker(globalDriveConfig.GetMaxTimeout())
                if err := w.Run(func() error { return s.deleteVersions(ctx, volume, fiv.Name, fiv.Versions...) }); err != nil {
                        errs[i] = err
                }
                diskHealthCheckOK(ctx, errs[i])
        }

        return errs
}
```

harshavardhana: each object gets a max 30sec to expire,there is no reason to have top level deadline here.

harshavardhana:
```diff
diff --git a/cmd/xl-storage.go b/cmd/xl-storage.go
index cdaa79af1..58682fa7f 100644
--- a/cmd/xl-storage.go
+++ b/cmd/xl-storage.go
@@ -1079,34 +1079,45 @@ func (s *xlStorage) deleteVersions(ctx context.Context, volume, path string, fis
                return err
        }

        discard := true
 
+       s.RLock()
+       legacy := s.formatLegacy
+       s.RUnlock()
+
        var legacyJSON bool
-       buf, _, err := s.readAllData(ctx, volume, volumeDir, pathJoin(volumeDir, path, xlStorageFormatFile), discard)
-       if err != nil {
-               if !errors.Is(err, errFileNotFound) {
-                       return err
+       buf, err := xioutil.WithDeadline[[]byte](ctx, globalDriveConfig.GetMaxTimeout(), func(ctx context.Context) ([]byte, error) {
+               buf, _, err := s.readAllData(ctx, volume, volumeDir, pathJoin(volumeDir, path, xlStorageFormatFile), discard)
+               if err != nil && !errors.Is(err, errFileNotFound) {
+                       return nil, err
                }
 
-               s.RLock()
-               legacy := s.formatLegacy
-               s.RUnlock()
                if legacy {
                        buf, _, err = s.readAllData(ctx, volume, volumeDir, pathJoin(volumeDir, path, xlStorageFormatFileV1), discard)
                        if err != nil {
-                               return err
+                               return nil, err
                        }
                        legacyJSON = true
                }
-       }
 
-       if len(buf) == 0 {
-               if errors.Is(err, errFileNotFound) && !skipAccessChecks(volume) {
-                       if aerr := Access(volumeDir); aerr != nil && osIsNotExist(aerr) {
-                               return errVolumeNotFound
+               if len(buf) == 0 {
+                       if errors.Is(err, errFileNotFound) && !skipAccessChecks(volume) {
+                               if aerr := Access(volumeDir); aerr != nil && osIsNotExist(aerr) {
+                                       return nil, errVolumeNotFound
+                               }
                        }
+                       return nil, errFileNotFound
                }
-               return errFileNotFound
+
+               return buf, nil
+       })
+       if err != nil {
+               return err
        }
 
        if legacyJSON {
@@ -1180,10 +1191,7 @@ func (s *xlStorage) DeleteVersions(ctx context.Context, volume string, versions
                        errs[i] = ctx.Err()
                        continue
                }
-               w := xioutil.NewDeadlineWorker(globalDriveConfig.GetMaxTimeout())
-               if err := w.Run(func() error { return s.deleteVersions(ctx, volume, fiv.Name, fiv.Versions...) }); err != nil {
-                       errs[i] = err
-               }
+               errs[i] = s.deleteVersions(ctx, volume, fiv.Name, fiv.Versions...)
                diskHealthCheckOK(ctx, errs[i])
        }
 
@@ -1214,7 +1222,7 @@ func (s *xlStorage) diskAlmostFilled() bool {
        return (float64(info.Free)/float64(info.Used)) < almostFilledPercent || (float64(info.FreeInodes)/float64(info.UsedInodes)) < almostFilledPercent
 }
 
-func (s *xlStorage) moveToTrash(filePath string, recursive, immediatePurge bool) (err error) {
+func (s *xlStorage) moveToTrashNoDeadline(filePath string, recursive, immediatePurge bool) (err error) {
        pathUUID := mustGetUUID()
        targetPath := pathutil.Join(s.drivePath, minioMetaTmpDeletedBucket, pathUUID)
 
@@ -1267,10 +1275,16 @@ func (s *xlStorage) moveToTrash(filePath string, recursive, immediatePurge bool)
                        }
                }
        }
-
        return nil
 }
 
+func (s *xlStorage) moveToTrash(filePath string, recursive, immediatePurge bool) (err error) {
+       w := xioutil.NewDeadlineWorker(globalDriveConfig.GetMaxTimeout())
+       return w.Run(func() (err error) {
+               return s.moveToTrashNoDeadline(filePath, recursive, immediatePurge)
+       })
+}
+
 // DeleteVersion - deletes FileInfo metadata for path at `xl.meta`. forceDelMarker
 // will force creating a new `xl.meta` to create a new delete marker
 func (s *xlStorage) DeleteVersion(ctx context.Context, volume, path string, fi FileInfo, forceDelMarker bool, opts DeleteOptions) (err error) {
@@ -1776,7 +1790,7 @@ func (s *xlStorage) readAllData(ctx context.Context, volume, volumeDir string, f
                // This discard is mostly true for deletes
                // so we need to make sure we do not keep
                // page-cache references after.
-               defer disk.Fdatasync(f)
+               defer Fdatasync(f)
        }
 
        defer f.Close()
harsha in ~/go/src/github.com/minio/minio on master ● ?38 ~3 λ git diff cmd/xl-storage* > deadline.diff
harsha in ~/go/src/github.com/minio/minio on master ● ?39 ~3 λ mc cp deadline.diff ~/Downloads/
...inio/minio/deadline.diff: 3.77 KiB / 3.77 KiB ┃▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓┃ 2.03 MiB/s 0sharsha in ~/go/src/github.com/minio/minio on master ● ?39 ~3 λ cat ~/Downloads/deadline.diff 
diff --git a/cmd/xl-storage.go b/cmd/xl-storage.go
index cdaa79af1..58682fa7f 100644
--- a/cmd/xl-storage.go
+++ b/cmd/xl-storage.go
@@ -1079,34 +1079,45 @@ func (s *xlStorage) deleteVersions(ctx context.Context, volume, path string, fis
 		return err
 	}
 
+	w := xioutil.NewDeadlineWorker(globalDriveConfig.GetMaxTimeout())
+	if err := w.Run(func() error { return }); err != nil {
+		errs[i] = err
+	}
+
 	discard := true
 
+	s.RLock()
+	legacy := s.formatLegacy
+	s.RUnlock()
+
 	var legacyJSON bool
-	buf, _, err := s.readAllData(ctx, volume, volumeDir, pathJoin(volumeDir, path, xlStorageFormatFile), discard)
-	if err != nil {
-		if !errors.Is(err, errFileNotFound) {
-			return err
+	buf, err := xioutil.WithDeadline[[]byte](ctx, globalDriveConfig.GetMaxTimeout(), func(ctx context.Context) ([]byte, error) {
+		buf, _, err := s.readAllData(ctx, volume, volumeDir, pathJoin(volumeDir, path, xlStorageFormatFile), discard)
+		if err != nil && !errors.Is(err, errFileNotFound) {
+			return nil, err
 		}
 
-		s.RLock()
-		legacy := s.formatLegacy
-		s.RUnlock()
 		if legacy {
 			buf, _, err = s.readAllData(ctx, volume, volumeDir, pathJoin(volumeDir, path, xlStorageFormatFileV1), discard)
 			if err != nil {
-				return err
+				return nil, err
 			}
 			legacyJSON = true
 		}
-	}
 
-	if len(buf) == 0 {
-		if errors.Is(err, errFileNotFound) && !skipAccessChecks(volume) {
-			if aerr := Access(volumeDir); aerr != nil && osIsNotExist(aerr) {
-				return errVolumeNotFound
+		if len(buf) == 0 {
+			if errors.Is(err, errFileNotFound) && !skipAccessChecks(volume) {
+				if aerr := Access(volumeDir); aerr != nil && osIsNotExist(aerr) {
+					return nil, errVolumeNotFound
+				}
 			}
+			return nil, errFileNotFound
 		}
-		return errFileNotFound
+
+		return buf, nil
+	})
+	if err != nil {
+		return err
 	}
 
 	if legacyJSON {
@@ -1180,10 +1191,7 @@ func (s *xlStorage) DeleteVersions(ctx context.Context, volume string, versions
 			errs[i] = ctx.Err()
 			continue
 		}
-		w := xioutil.NewDeadlineWorker(globalDriveConfig.GetMaxTimeout())
-		if err := w.Run(func() error { return s.deleteVersions(ctx, volume, fiv.Name, fiv.Versions...) }); err != nil {
-			errs[i] = err
-		}
+		errs[i] = s.deleteVersions(ctx, volume, fiv.Name, fiv.Versions...)
 		diskHealthCheckOK(ctx, errs[i])
 	}
 
@@ -1214,7 +1222,7 @@ func (s *xlStorage) diskAlmostFilled() bool {
 	return (float64(info.Free)/float64(info.Used)) < almostFilledPercent || (float64(info.FreeInodes)/float64(info.UsedInodes)) < almostFilledPercent
 }
 
-func (s *xlStorage) moveToTrash(filePath string, recursive, immediatePurge bool) (err error) {
+func (s *xlStorage) moveToTrashNoDeadline(filePath string, recursive, immediatePurge bool) (err error) {
 	pathUUID := mustGetUUID()
 	targetPath := pathutil.Join(s.drivePath, minioMetaTmpDeletedBucket, pathUUID)
 
@@ -1267,10 +1275,16 @@ func (s *xlStorage) moveToTrash(filePath string, recursive, immediatePurge bool)
 			}
 		}
 	}
-
 	return nil
 }
 
+func (s *xlStorage) moveToTrash(filePath string, recursive, immediatePurge bool) (err error) {
+	w := xioutil.NewDeadlineWorker(globalDriveConfig.GetMaxTimeout())
+	return w.Run(func() (err error) {
+		return s.moveToTrashNoDeadline(filePath, recursive, immediatePurge)
+	})
+}
+
 // DeleteVersion - deletes FileInfo metadata for path at `xl.meta`. forceDelMarker
 // will force creating a new `xl.meta` to create a new delete marker
 func (s *xlStorage) DeleteVersion(ctx context.Context, volume, path string, fi FileInfo, forceDelMarker bool, opts DeleteOptions) (err error) {
@@ -1776,7 +1790,7 @@ func (s *xlStorage) readAllData(ctx context.Context, volume, volumeDir string, f
 		// This discard is mostly true for deletes
 		// so we need to make sure we do not keep
 		// page-cache references after.
-		defer disk.Fdatasync(f)
+		defer Fdatasync(f)
 	}
 
 	defer f.Close()
```

Here is the diff perhaps

vadmeste:
> Here is the diff perhaps

yes, I tried to change the deadline API to delay the deadline triggering inside deleteVersions(), like in TCP, but it was not easy to do. I will use your suggestion and see but I think it should be fine overall.

harshavardhana (changes requested):
write a new function, do not modify the existing functions signature. Leave them the way they are.

harshavardhana (approved):

harshavardhana (approved):

internal/ioutil/ioutil.go
```diff
@@ -101,21 +101,39 @@ type ioret[V any] struct {
...
+
+// WithDeadlineAndReset will execute a function with a deadline and return a value of a given type.
+// If the deadline/context passes before the function finishes executing, the zero value and the
+// context error is returned. The function can send a message to reset channel to delay deadline
+// triggering by another timeout
+func WithDeadlineAndReset[V any](ctx context.Context, timeout time.Duration, work func(ctx context.Context, reset chan struct{}) (result V, err error)) (result V, err error) {
+	t := time.NewTimer(timeout)
+	defer t.Stop()
 
+	reset := make(chan struct{})
```
krisis:
```suggestion
	reset := make(chan struct{}, 1)
```
`reset` needs to be a buffered to channel to avoid `work` function from being blocked due to an unbuffered channel with no receiver. The same reason why `c chan ioret[V]` is buffered.

cmd/xl-storage.go
```diff
@@ -1153,6 +1153,8 @@ func (s *xlStorage) deleteVersions(ctx context.Context, volume, path string, fis
 					return err
 				}
 			}
+
+			resetDeadline <- struct{}{}
```
krisis: We can send on `resetDeadline` channel inside a select statement with a default case to prevent `deleteVersions` from being blocked.

harshavardhana (commented):
PR is redone with a simpler implementation @krisis as you found more bugs and races in the implementation.

harshavardhana:
Closing this PR, moved the code here to new PR. https://github.com/minio/minio/pull/20185

PR #20149, opened 2024-07-25
fix: check once ready for site-replication

jiuker:
check once ready for site-replication

## Description
Before, we will check very bucket with the remote ready call for site-replication.
Now, we reuse the first respsone.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/bucket-replication.go
+++ b/cmd/bucket-replication.go
@@ -91,7 +91,10 @@ func getReplicationConfig(ctx context.Context, bucketName string) (rc *replicati
 
 // validateReplicationDestination returns error if replication destination bucket missing or not configured
 // It also returns true if replication destination is same as this server.
-func validateReplicationDestination(ctx context.Context, bucket string, rCfg *replication.Config, checkRemote bool) (bool, APIError) {
+func validateReplicationDestination(ctx context.Context, bucket string, rCfg *replication.Config, opts *validateReplicationDestinationOptions) (bool, APIError) {
+	if opts == nil {
+		opts = &validateReplicationDestinationOptions{}
+	}
 	var arns []string
 	if rCfg.RoleArn != "" {
 		arns = append(arns, rCfg.RoleArn)
@@ -113,7 +116,7 @@ func validateReplicationDestination(ctx context.Context, bucket string, rCfg *re
 		if clnt == nil {
 			return sameTarget, toAPIError(ctx, BucketRemoteTargetNotFound{Bucket: bucket})
 		}
-		if checkRemote { // validate remote bucket
+		if opts.CheckRemoteBucket { // validate remote bucket
 			found, err := clnt.BucketExists(ctx, arn.Bucket)
 			if err != nil {
 				return sameTarget, errorCodes.ToAPIErrWithErr(ErrRemoteDestinationNotFoundError, err)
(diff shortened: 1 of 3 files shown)


cmd/bucket-replication.go
```diff
@@ -3741,3 +3739,13 @@ func (p *ReplicationPool) getMRF(ctx context.Context, bucket string) (ch <-chan
...
 	return mrfCh, nil
 }
+
+// validateReplicationDestinationOptions is used to configure the validation of the replication destination.
+// validateReplicationDestination uses this to configure the validation.
+type validateReplicationDestinationOptions struct {
+	CheckRemoteBucket bool
+	CheckReady        bool
+
+	checkReadyResp error
```
shtripat: As this is error better name could be `checkReadyErr`

shtripat (approved):

cmd/bucket-replication.go
```diff
@@ -113,7 +113,7 @@ func validateReplicationDestination(ctx context.Context, bucket string, rCfg *re
 		if clnt == nil {
 			return sameTarget, toAPIError(ctx, BucketRemoteTargetNotFound{Bucket: bucket})
 		}
-		if checkRemote { // validate remote bucket
+		if opts.CheckRemoteBucket { // validate remote bucket
```
ramondeklein, line 119: This will crash when `opts` is `nil`.

cmd/bucket-replication.go
```diff
@@ -91,7 +91,7 @@ func getReplicationConfig(ctx context.Context, bucketName string) (rc *replicati
 
 // validateReplicationDestination returns error if replication destination bucket missing or not configured
 // It also returns true if replication destination is same as this server.
-func validateReplicationDestination(ctx context.Context, bucket string, rCfg *replication.Config, checkRemote bool) (bool, APIError) {
+func validateReplicationDestination(ctx context.Context, bucket string, rCfg *replication.Config, opts *validateReplicationDestinationOptions) (bool, APIError) {
```
ramondeklein, line 94: I would suggest setting `opts` to a default structure at the beginning at this function if its value is `nil`. I'm not really sure why this boolean needed to be changed into a structure.

cmd/bucket-replication.go
```diff
@@ -133,24 +133,22 @@ func validateReplicationDestination(ctx context.Context, bucket string, rCfg *re
...
-				}
-			}
-			if c.EndpointURL().String() == clnt.EndpointURL().String() {
-				selfTarget, _ := isLocalHost(clnt.EndpointURL().Hostname(), clnt.EndpointURL().Port(), globalMinioPort)
-				if !sameTarget {
-					sameTarget = selfTarget
-				}
-				continue
+		// if checked bucket, then check the ready is unnecessary
+		if !opts.CheckRemoteBucket && opts.CheckReady {
```
ramondeklein, line 140: This will crash when `opts` is `nil`.

ramondeklein (changes requested):

cmd/bucket-replication.go
jiuker:
```go
	                opts.checkReadyOnce.Do(func() {
				opts.checkReadyErr = checkRemoteEndpoint(ctx, clnt.EndpointURL())
			})
			switch opts.checkReadyErr.(type) {
			case BucketRemoteIdenticalToSource:
				return true, errorCodes.ToAPIErrWithErr(ErrBucketRemoteIdenticalToSource, fmt.Errorf("remote target endpoint %s is self referential", clnt.EndpointURL().String()))
			default:
			}
```
All is for this. @ramondeklein . Opts will record the first error

cmd/bucket-replication.go
```diff
@@ -113,7 +113,7 @@ func validateReplicationDestination(ctx context.Context, bucket string, rCfg *re
 		if clnt == nil {
 			return sameTarget, toAPIError(ctx, BucketRemoteTargetNotFound{Bucket: bucket})
 		}
-		if opts.CheckRemoteBucket { // validate remote bucket
+		if opts != nil && opts.CheckRemoteBucket { // validate remote bucket
```
ramondeklein: Adding this at the top of the function is simpler and safer (now you need to check every time you access `opts` and it's easy to forget and cause a crash):
```go
if opts == nil {
        opts = &validateReplicationDestinationOptions{}
}
```
Or even simpler... Don't make it a pointer, but pass the structure by value.

cmd/bucket-replication.go
ramondeklein: ~Ah, then you need to keep it a pointer and not pass by value. But you could have used a `*bool` as well instead of the structure. Initially set it to `nil` and let the function set it to `false` if it's `nil`. That's simple and safe.~

Nevermind, there is more in the structure then just this value.

jiuker: Here's another one `CheckReady` @ramondeklein

ramondeklein (approved):
Can't really comment on functionality, but code looks fine.

cmd/bucket-replication.go
```diff
@@ -133,24 +136,22 @@ func validateReplicationDestination(ctx context.Context, bucket string, rCfg *re
...
-			}
-			if c.EndpointURL().String() == clnt.EndpointURL().String() {
-				selfTarget, _ := isLocalHost(clnt.EndpointURL().Hostname(), clnt.EndpointURL().Port(), globalMinioPort)
-				if !sameTarget {
-					sameTarget = selfTarget
-				}
-				continue
+		// if checked bucket, then check the ready is unnecessary
+		if !opts.CheckRemoteBucket && opts.CheckReady {
+			opts.checkReadyOnce.Do(func() {
```
poornas: @jiuker, you are assuming there is just a single replication destination - site replication can be configured to multiple peers, so you should be using a map

poornas (changes requested):

cmd/bucket-replication.go
shtripat: Ah yes, I totally missed that.

poornas (approved):

harshavardhana (approved):

PR #20150, opened 2024-07-25
Implement bucket-level CORS configuration settings with S3 APIs PutBucketCors, etc

marktheunissen:
## Description

This PR adds support for the S3 APIs `PutBucketCors`, `DeleteBucketCors`, `GetBucketCors` and allows the API to use these bucket-specific CORS configurations while falling back to global config where no bucket-specific configuration is given.

## Motivation and Context

MinIO has always supported "Global CORS" settings, but has not supported the S3 Bucket Cors APIS:

- https://min.io/docs/minio/kubernetes/upstream/operations/concepts/thresholds.html#unsupported-s3-bucket-apis

The global configuration is done with `cors_allow_origin` or env var `MINIO_API_CORS_ALLOW_ORIGIN`, and defaults to origins=`*`. This is the only CORS configuration that is currently supported.

S3's API, however, allows a user to put a CORS configuration for a given bucket with a rich configuration language, up to 100 rules may be applied. Docs:
- https://docs.aws.amazon.com/AmazonS3/latest/userguide/ManageCorsUsing.html

For example:

```
<CORSConfiguration>
 <CORSRule>
   <AllowedOrigin>http://www.example1.com</AllowedOrigin>

   <AllowedMethod>PUT</AllowedMethod>
   <AllowedMethod>POST</AllowedMethod>
   <AllowedMethod>DELETE</AllowedMethod>

   <AllowedHeader>*</AllowedHeader>
 </CORSRule>
</CORSConfiguration>
```

The goal of this PR is to match the S3 rules engine as closely as possible, with a few small exceptions where S3 does not follow the fetch() spec. CORS is made for browsers and thus it's the fetch() spec that defines how it should behave. These exceptions are documented in the functional tests in `minio-go`. Fetch spec is here:

- https://fetch.spec.whatwg.org/

Another objective is to allow the global CORS settings to continue functioning, without any action required from users who do not apply bucket CORS.

### Implementation

MinIO currently uses https://github.com/rs/cors to handle Global CORS. 

The intention in this PR is to provide an adapter which can transform a matched S3 Bucket CORS rule, to an `rs/cors` configuration. This is slightly less efficient than implementing a seperate CORS engine just for the bucket configurations, however, it simplifies things, as there is only one actual cors engine, `rs/cors`, not two.

The old implementation applies CORS through a handler wrapper, rather than a mux.Middleware. This PR needs to change that, as we now need the bucket name in the CORS handling logic, and the mux router does not provide the `bucket` var name, unless a route matches.

This PR introduces `corsHandler` as a real Middleware (added to `globalMiddlewares`). `corsHandler` is able to check if a CORS config is applied to a bucket, and if it is, use it instead of the global config. 

The disadvantage of using mux.Middleware, is that it's slightly less efficient, as mux sets up the Middleware chain every time there is a route match. This is in contrast to the current global CORS handler wrapper, which is only called once when the router is initialized. 

One unavoidable change is that currently preflight is evaluated before any auth, so for example, doing `curl -v -XOPTIONS https://play.min.io/minio/kms/v1/status -H "Origin: http://foobar.com/"` gives you 200 OK. After this PR, auth will be evaluated first, so it will be access denied. I believe the new code in this PR is a more correct implementation.

## How to test this PR?

- Automated tests are provided
- The functional tests in the minio-go repo contain 30+ request/response tests that verify CORS correctness. 

## Types of changes

## Checklist:

All related PRs:
- https://github.com/minio/minio/pull/20150
- https://github.com/minio/pkg/pull/123
- https://github.com/minio/madmin-go/pull/290
- https://github.com/minio/minio-go/pull/1987


go.mod
```diff
@@ -260,3 +260,9 @@ require (
 	google.golang.org/grpc v1.65.0 // indirect
 	google.golang.org/protobuf v1.34.2 // indirect
 )
+
+replace github.com/minio/madmin-go/v3 => github.com/marktheunissen/madmin-go/v3 v3.0.0-20240724071817-c292e76a6f95
+
+replace github.com/minio/minio-go/v7 => github.com/marktheunissen/minio-go/v7 v7.0.0-20240725051157-a17359836765
+
+replace github.com/minio/pkg/v3 => github.com/marktheunissen/pkg/v3 v3.0.0-20240725041217-099abbc8b0d4
```
marktheunissen: TODO: Remove all these replace directives.

cmd/admin-bucket-handlers.go
```diff
@@ -933,6 +936,9 @@ func (a adminAPIHandlers) ImportBucketMetadataHandler(w http.ResponseWriter, r *
 			bucketMap[bucket].QuotaConfigJSON = data
 			bucketMap[bucket].QuotaConfigUpdatedAt = updatedAt
 			rpt.SetStatus(bucket, fileName, nil)
+
+		case bucketCorsConfig:
+			// TODO: implement CORS import, test all this
```
klauspost: We probably want this for release.

cmd/api-errors.go
```diff
@@ -2120,6 +2130,36 @@ var errorCodes = errorCodeMap{
 		Description:    "Invalid UTF-8 character detected.",
 		HTTPStatusCode: http.StatusBadRequest,
 	},
+	ErrCORSTooManyRules: {
+		Code:           "InvalidRequest",
+		Description:    "The number of CORS rules should not exceed allowed limit of 100 rules.",
```
klauspost, line 2138:
```suggestion
		Description:    "The number of CORS rules may not exceed allowed limit of 100 rules.",
```

cmd/api-errors.go
```diff
@@ -2120,6 +2130,36 @@ var errorCodes = errorCodeMap{
...
+		HTTPStatusCode: http.StatusBadRequest,
+	},
+	ErrCORSMalformedXML: {
+		Code:           "MalformedXML",
+		Description:    "The XML you provided was not well-formed or did not validate against our published schema",
```
klauspost, line 2143:
```suggestion
		Description:    "The XML you provided was not well-formed or did not validate against our published schema.",
```

cmd/api-errors.go
```diff
@@ -2120,6 +2130,36 @@ var errorCodes = errorCodeMap{
...
+		HTTPStatusCode: http.StatusBadRequest,
+	},
+	ErrCORSAllowedOriginWildcards: {
+		Code:           "InvalidRequest",
+		Description:    "AllowedOrigin can not have more than one wildcard.",
+		HTTPStatusCode: http.StatusBadRequest,
+	},
+	ErrCORSInvalidMethod: {
+		Code:           "InvalidRequest",
+		Description:    "Found unsupported HTTP method in CORS config",
```
klauspost, line 2153:
```suggestion
		Description:    "Unsupported HTTP method in CORS config.",
```

cmd/api-errors.go
```diff
@@ -2120,6 +2130,36 @@ var errorCodes = errorCodeMap{
...
+		HTTPStatusCode: http.StatusBadRequest,
+	},
+	ErrCORSAllowedHeaderWildcards: {
+		Code:           "InvalidRequest",
+		Description:    "AllowedHeader can not have more than one wildcard.",
+		HTTPStatusCode: http.StatusBadRequest,
+	},
+	ErrCORSRequestNotAllowed: {
+		Code:           "AccessForbidden",
+		Description:    "CORSResponse: This CORS request is not allowed. This is usually because the evalution of Origin, request method / Access-Control-Request-Method or Access-Control-Request-Headers are not whitelisted by the resource CORS spec.",
```
klauspost, line 2164:
```suggestion
		Description:    "CORS request not allowed. This is usually because the evaluation of Origin, request method / Access-Control-Request-Method or Access-Control-Request-Headers are not whitelisted by the resource CORS spec.",
```

.typos.toml
```diff
@@ -20,6 +20,7 @@ extend-ignore-re = [
     'http\.Header\{"X-Amz-Server-Side-Encryptio":',
     "ZoEoZdLlzVbOlT9rbhD7ZN7TLyiYXSAlB79uGEge",
     "ERRO:",
+    "because the evalution of Origin",
```
klauspost:
```suggestion
    "because the evaluation of Origin",
```

cmd/routers.go
```diff
@@ -38,6 +38,8 @@ func registerDistErasureRouters(router *mux.Router, endpointServerPools Endpoint
 	// Register distributed namespace lock routers.
 	registerLockRESTHandlers()
 
+	// TODO: We probably don't need OPTIONS preflight matching on this router, but just check.
```
klauspost, line 49: Resolve.

cmd/site-replication.go
```diff
@@ -4473,6 +4505,7 @@ func (c *SiteReplicationSys) healBuckets(ctx context.Context, objAPI ObjectLayer
 			c.healOLockConfigMetadata(ctx, objAPI, bucket, info)
 			c.healSSEMetadata(ctx, objAPI, bucket, info)
 			c.healBucketReplicationConfig(ctx, objAPI, bucket, info)
+			// TODO: Heal Bucket CORS
```
klauspost: Maybe see if @poornas can help you get these in.

klauspost (commented):
Just some minor stuff.

cmd/api-errors.go
marktheunissen: @klauspost These error strings are currently exactly the same as S3, with spelling mistakes and everything. Would you still like them fixed up even if it breaks parity with S3 API?

.typos.toml
marktheunissen: Spelling mistake is present in AWS S3 API's response.

cmd/admin-bucket-handlers.go
marktheunissen: This is done now.

cmd/routers.go
marktheunissen: This TODO is a question for reviewers: Do we need OPTIONS on these? I doubt it because these are `registerDistErasureRouters` which seem to all deal with peer-to-peer endpoints (grid?). I doubt that a browser is using fetch() under cross-origin against one of these APIs?

go.mod
```diff
@@ -260,3 +260,9 @@ require (
 	google.golang.org/grpc v1.65.0 // indirect
 	google.golang.org/protobuf v1.34.2 // indirect
 )
+
+replace github.com/minio/madmin-go/v3 => github.com/marktheunissen/madmin-go/v3 v3.0.0-20240730061730-2d8255671cb3
+
+replace github.com/minio/minio-go/v7 => github.com/marktheunissen/minio-go/v7 v7.0.0-20240726014459-3a5e31b9843d
+
+replace github.com/minio/pkg/v3 => github.com/marktheunissen/pkg/v3 v3.0.0-20240726014105-5a91ee390421
```
marktheunissen: Remove when dependency merged.

cmd/api-errors.go
harshavardhana: Add a comment on each of such strings so that future developers know update it when AWS fixes them.  Point out their mistakes as well.

cmd/api-router-cors.go
```diff
@@ -0,0 +1,234 @@
...
+		vars := mux.Vars(r)
+		bucket := vars["bucket"]
+		if bucket == "" {
+			// No bucket, continue the request using global CORS.
+			corsGlobal.ServeHTTP(w, r)
+			return
+		}
+
+		// Check if the bucket has a CORS config set.
+		config, _, err := globalBucketMetadataSys.GetCorsConfig(bucket)
```
harshavardhana, line 55: can you wrap this around singleflight package so that we can de-duplicate concurrent callers.

cmd/api-router-cors.go
```diff
@@ -0,0 +1,234 @@
...
+		skipCredentialsMethods := []string{http.MethodGet, http.MethodHead}
+		methodToCheck := r.Header.Get("Access-Control-Request-Method")
+		if methodToCheck == "" {
+			methodToCheck = r.Method
+		}
+		if slices.Contains(skipCredentialsMethods, methodToCheck) {
+			opts.AllowCredentials = false
+		}
+
+		rscors.New(*opts).ServeHTTP(w, r, handler.ServeHTTP)
```
harshavardhana, line 105:
```suggestion
		rscors.New(*opts).ServeHTTP(w, r, handler.ServeHTTP)
```

is this okay to cause repeated allocation? any overhead?

cmd/routers.go
harshavardhana: no they won't be, in-fact we should reject all CORS, OPTIONs request at `/minio/` resource entirely.

docs/cors/README.md
```diff
@@ -0,0 +1,21 @@
+# CORS settings for MinIO
+
+There are two ways to apply CORS in MinIO:
```
harshavardhana: maybe move this to `docs/bucket/cors/`

harshavardhana (changes requested):

harshavardhana:
> One unavoidable change is that currently preflight is evaluated before any auth, so for example, doing `curl -v -XOPTIONS https://play.min.io/minio/kms/v1/status -H "Origin: http://foobar.com/"` gives you 200 OK. After this PR, auth will be evaluated first, so it will be access denied. I believe the new code in this PR is a more correct implementation.

This is correct implementation.

cmd/api-errors.go
marktheunissen: Will do.

cmd/routers.go
marktheunissen: @harshavardhana Don't we need to support a use case where someone makes browser-based requests to the MinIO admin API? 

E.g. if a user has a custom web UI that is cross-domain and does some direct requests to MinIO (not using or  bypassing MinIO console)

cmd/api-router-cors.go
marktheunissen: @harshavardhana Yeah this will generate overhead, it's not as efficient as it could be. I would say this PR's approach of using `rs/cors` and the `mux.Middleware` is just the simple thing that works, and has less potential for disruption and maybe less complexity than other alternatives. But it's definitely creating an overhead. 

I have some basic optimizations that I can do for the allocations in `corsGlobalHandler`, I will push up as a new commit.

For this line 107 that you commented on, that's the tricky one to optimize, since it's the case where a CORS Rule (XML rule) has been found for a bucket, and we want to apply it. Going "via" `rs/cors` keeps things consistent, but involves extra allocations. I could remove `rs/cors` itself and then try re-implement the critical parts it in a more efficient way, but not sure how you feel about that.

cmd/api-router-cors.go
marktheunissen: @harshavardhana  I expect the majority of calls to GetCorsConfig to just be a map lookup, with read concurrency supported via RLock here: https://github.com/minio/minio/blob/master/cmd/bucket-metadata-sys.go#L449

Consider that singleflight will lock a mutex here: https://cs.opensource.google/go/x/sync/+/master:singleflight/singleflight.go;l=92,  so if we use it in the wrong place, we restrict the concurrency unnecessarily.

What are we protecting? Maybe we just want to protect the scenario where we have a cache miss (i.e. metadataMap[bucket] not found), here: https://github.com/minio/minio/blob/master/cmd/bucket-metadata-sys.go#L454

In that case I would add the singleflight at that location in `BucketMetadataSys.GetConfig()` and then it would apply to all requests for bucket config, that are cache misses.

Let me know your thoughts

marktheunissen:
@harshavardhana I've pushed some commits for review

cmd/routers.go
harshavardhana: There may be cases, but we have never officially endorsed or documented such things.

There is no reason to have pure browser based requests. But I guess let's leave that for now.

cmd/api-router-cors.go
harshavardhana: We are not protecting we are simply deduplicating the call.

cmd/api-router-cors.go
harshavardhana: Lets keep this for now, this is an esoteric use-case anyway.

harshavardhana (commented):
LGTM please update the dependency.

harshavardhana:
Lets move this PR to mineos only @marktheunissen  https://github.com/miniohq/mineos we are not making this change for public consumption.

marktheunissen:
@harshavardhana How will we handle the divergence in repos like `pkg` and `minio-go`? For example, there are now CORS tests in `minio-go` that require this PR: https://github.com/minio/minio-go/blob/master/functional_tests.go#L13560

harshavardhana:
> @harshavardhana How will we handle the divergence in repos like `pkg` and `minio-go`? For example, there are now CORS tests in `minio-go` that require this PR: https://github.com/minio/minio-go/blob/master/functional_tests.go#L13560

They are clients @marktheunissen and libraries the main functionality stays in EOS.

harshavardhana:
> @harshavardhana How will we handle the divergence in repos like `pkg` and `minio-go`? For example, there are now CORS tests in `minio-go` that require this PR: https://github.com/minio/minio-go/blob/master/functional_tests.go#L13560

They must be run against EOS only.

cmd/api-errors.go
```diff
@@ -2158,7 +2159,8 @@ var errorCodes = errorCodeMap{
 		HTTPStatusCode: http.StatusBadRequest,
 	},
 	ErrCORSRequestNotAllowed: {
-		Code:           "AccessForbidden",
+		Code: "AccessForbidden",
+		// FIXME: 'evalution' is a spelling mistake that is copied verbatim from the S3 API in order to be as compatible as possible.
```
klauspost: We do not need to follow the descriptions letter for letter.

From https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingRESTError.html

> The error message contains a generic description of the error condition in English. It is intended for a human audience. Simple programs display the message directly to the end user if they encounter an error condition they don't know how or don't care to handle.

Please correct spelling mistakes and if there are ways to make the description clearer feel free to do so.

cmd/api-errors.go
```diff
@@ -2133,7 +2133,8 @@ var errorCodes = errorCodeMap{
 		HTTPStatusCode: http.StatusBadRequest,
 	},
 	ErrCORSTooManyRules: {
-		Code:           "InvalidRequest",
+		Code: "InvalidRequest",
+		// FIXME: s/should not/may not/ to be more accurate, this string is copied verbatim from the S3 API error response.
```
klauspost, line 2137: pls fix

cmd/api-errors.go
harshavardhana: Actually this is not going to work @klauspost applications use strings directly in many SDKs and applications by comparing verbatim strings.

I am afraid we can't change it. Unfortunately.

marktheunissen:
Moving to MinEOS

PR #20161, opened 2024-07-25
fix replication of checksum when encryption is enabled

poornas:

## Motivation and Context
part checksum header was not being sent on upload part while replicating encrypted content

This PR also corrects response for GET calls with part-number and checksum-mode specified
## How to test this PR?
1. Added tests to capture the replication of objects with checksum headers
2.  GET/HEAD requests for part on multipart object uploaded with checksum fails with InvalidPartNumber error for master branch
```
aws s3api head-object  --bucket bucket --endpoint-url http://localhost:9001 --profile minio --key mult --part-number 7 --checksum-mode ENABLED

 aws s3api get-object  --bucket bucket --endpoint-url http://localhost:9001 --profile minio --key mult --part-number 1 --checksum-mode sha256  /tmp/file
```

## Types of changes

## Checklist:

### DIFF
--- a/cmd/bucket-replication.go
+++ b/cmd/bucket-replication.go
@@ -774,21 +774,6 @@ func (m caseInsensitiveMap) Lookup(key string) (string, bool) {
 	return "", false
 }
 
-func getCRCMeta(oi ObjectInfo, partNum int, h http.Header) map[string]string {
-	meta := make(map[string]string)
-	cs := oi.decryptChecksums(partNum, h)
-	for k, v := range cs {
-		cksum := hash.NewChecksumString(k, v)
-		if cksum == nil {
-			continue
-		}
-		if cksum.Valid() {
-			meta[cksum.Type.Key()] = v
-		}
-	}
-	return meta
-}
-
 func putReplicationOpts(ctx context.Context, sc string, objInfo ObjectInfo, partNum int) (putOpts minio.PutObjectOptions, err error) {
 	meta := make(map[string]string)
 	isSSEC := crypto.SSEC.IsEncrypted(objInfo.UserDefined)
@@ -797,11 +782,6 @@ func putReplicationOpts(ctx context.Context, sc string, objInfo ObjectInfo, part
 		// In case of SSE-C objects copy the allowed internal headers as well
 		if !isSSEC || !slices.Contains(maps.Keys(validSSEReplicationHeaders), k) {
 			if stringsHasPrefixFold(k, ReservedMetadataPrefixLower) {
-				if strings.EqualFold(k, ReservedMetadataPrefixLower+"crc") {
-					for k, v := range getCRCMeta(objInfo, partNum, nil) {
-						meta[k] = v
-					}
-				}
 				continue
 			}
 			if isStandardHeader(k) {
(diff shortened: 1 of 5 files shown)


cmd/bucket-replication.go
```diff
@@ -1666,7 +1666,7 @@ func replicateObjectWithMultipart(ctx context.Context, c *minio.Core, bucket, ob
 		cHeader := http.Header{}
 		cHeader.Add(xhttp.MinIOSourceReplicationRequest, "true")
 		if !isSSEC {
-			crc := getCRCMeta(objInfo, partInfo.Number, nil) // No SSE-C keys here.
+			crc := getCRCMeta(objInfo, partIdx+1, nil) // No SSE-C keys here.
```
harshavardhana: we should always use partInfo.Number - is it because this is not populated?

harshavardhana (changes requested):

cmd/bucket-replication.go
poornas: yes

harshavardhana: that is strange, it can be pre-populated and there is no reason to make this change here.

poornas: hash.ReadCheckSums is already extracting the right part checksum given the part number. Don't see the need to prepopulate this or save it in xl.meta

harshavardhana: No, I meant whoever returns you PartInfo must already prepopulate this, it doesn't need to be persisted. My point is this is not clean, the code that exists now is clean. 

just prepopulate it when ToObjectInfo() happens.

shtripat (approved):
no comments other than @harshavardhana 's comments

cmd/bucket-replication.go
harshavardhana: because this nuance is unnatural from upper layers to understand. It is safe to add PartNumber such that other callers (there may be others using this would not bother with it to deal with loop index numbers)

klauspost: AFAICT there can be gaps in part numbers, so this does not appear safe. Can you identify in which cases the number is missing and maybe fill it in then?

harshavardhana: eventually, they are all sorted @klauspost and written so its always part.1 is 1st element. If we write checksum encryption based on the final index, then there is no issue.

poornas: there is already a bug with hash.ReadCheckSums because it assumes the parts are sequential. If a CompleteMultipart finalizes parts 4 and 7 - a subsequent GET for that part(4 or 7) with checksum enabled returns InvalidPartNumber error, whereas asking for part number 1 or 2 returns the checksum properly.

harshavardhana: @poornas yeah we need to fix all of that properly.

harshavardhana (approved):

harshavardhana:
```
{
  "time": "2024-07-26T23:35:45.564633616Z",
  "level": "ERROR",
  "name": "minio-go: testPutObjectWithChecksums",
  "duration": 15,
  "function": "PutObject(bucketName, objectName, reader,size, opts)",
  "args": {
    "bucketName": "minio-go-test-2iccnai3dduu0hpy",
    "metadata": {
      "x-amz-checksum-crc32": "jlDeQA=="
    },
    "objectName": "z0hvfztnprfgoldvnpwxwrpw4floyp",
    "opts": "minio.PutObjectOptions{UserMetadata: metadata, Progress: progress}",
    "range": "false"
  },
  "status": "FAIL",
  "message": "checksum mismatch",
  "error": "want jlDeQA==, got "
}
```

looks like something is up with minio-go and this PR.

PR #20181, opened 2024-07-29
Add cache for multipart uploads of an object

shtripat:

## Description
This cache would retain the parts in cache for 24hrs and then evict them.

## Motivation and Context
Maintain a cache of uploaded parts of a objects for listing them overall, by `prefix` and by `upalodIDMarker` etc

## Types of changes

## Checklist:


cmd/erasure-server-pool.go
```diff
@@ -47,6 +47,72 @@ import (
...
+}
+
+// cleanExpiredEntries removes expired entries every hour
+func (c *MultipartUploadCache) cleanExpiredEntries() {
+	for {
+		time.Sleep(1 * time.Hour)
+		c.mutex.Lock()
+		now := time.Now()
+		for key, entry := range c.data {
+			if now.Sub(entry.Timestamp) > 24*time.Hour {
```
```suggestion
			if now.Sub(entry.Timestamp) > globalAPIConfig.getStaleUploadsExpiry() {
```

cmd/erasure-server-pool.go
```diff
@@ -47,6 +47,72 @@ import (
...
+		Timestamp: time.Now(),
+	}
+}
+
+// Get retrieves a value from the cache if it has not expired
+func (c *MultipartUploadCache) Get(key string) (MultipartUpload, bool) {
+	c.mutex.RLock()
+	defer c.mutex.RUnlock()
+	entry, exists := c.data[key]
+	if !exists || time.Since(entry.Timestamp) > 24*time.Hour {
```
```suggestion
	if !exists || time.Since(entry.Timestamp) > globalAPIConfig.getStaleUploadsExpiry() {
```

cmd/erasure-server-pool.go
```diff
@@ -47,6 +47,72 @@ import (
...
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+	c.data[key] = MultipartUploadCacheEntry{
+		Upload:    value,
+		Timestamp: time.Now(),
+	}
+}
+
+// Get retrieves a value from the cache if it has not expired
+func (c *MultipartUploadCache) Get(key string) (MultipartUpload, bool) {
```
```suggestion
func (c *MultipartUploadCache) Get(key string) (MultipartUpload, bool) {
```

Add another helper called `Has(key string) (bool) {`

cmd/erasure-server-pool.go
```diff
@@ -1713,8 +1785,16 @@ func (z *erasureServerPools) ListMultipartUploads(ctx context.Context, bucket, p
 		if err != nil {
 			return result, err
 		}
-		poolResult.Uploads = append(poolResult.Uploads, result.Uploads...)
+		// poolResult.Uploads = append(poolResult.Uploads, result.Uploads...)
+		uploads = append(uploads, result.Uploads...)
+	}
+
+	for _, upload := range uploads {
+		if _, found := z.multipartUploadsCache.Get(upload.UploadID); found {
```
```suggestion
		if ok := z.multipartUploadsCache.Has(upload.UploadID); ok {
```

implement a Has(key) method to make this simpler

harshavardhana (changes requested):

cmd/erasure-server-pool.go
```diff
@@ -47,6 +47,80 @@ import (
...
+		Timestamp: time.Now(),
+	}
+}
+
+// Get retrieves a value from the cache if it has not expired
+func (c *MultipartUploadCache) Get(key string) (MultipartUpload, bool) {
+	c.mutex.RLock()
+	defer c.mutex.RUnlock()
+	entry, exists := c.data[key]
+	if !exists || time.Since(entry.Timestamp) > globalAPIConfig.getStaleUploadsExpiry() {
```
harshavardhana: Get expiry value before the lock.

cmd/erasure-server-pool.go
```diff
@@ -47,6 +47,80 @@ import (
...
+}
+
+// cleanExpiredEntries removes expired entries every hour
+func (c *MultipartUploadCache) cleanExpiredEntries() {
+	for {
+		time.Sleep(1 * time.Hour)
+		c.mutex.Lock()
+		now := time.Now()
+		for key, entry := range c.data {
+			if now.Sub(entry.Timestamp) > globalAPIConfig.getStaleUploadsExpiry() {
```
harshavardhana: Get expiry before the lock?

harshavardhana (changes requested):

cmd/erasure-server-pool.go
shtripat: I thought like get entry from map after getting a lock. Do you suggest like
```
        entry, exists := c.data[key]
        c.mutex.RLock()
        defer c.mutex.RUnlock()
        if !exists || time.Since(entry.Timestamp) > globalAPIConfig.getStaleUploadsExpiry() {
                return MultipartUpload{}, false
        }
        return entry.Upload, true
```
not sure I understood.

```suggestion
        expiry := globalAPIConfig.getStaleUploadsExpiry()
	c.mutex.RLock()
	defer c.mutex.RUnlock()
	entry, exists := c.data[key]
	if !exists || time.Since(entry.Timestamp) > expiry {
```

harshavardhana (commented):
this is what I meant

cmd/erasure-server-pool.go
shtripat: ah ok, sorry for being so naive. Thanks.

cmd/erasure-server-pool.go
```diff
@@ -1791,6 +1881,13 @@ func (z *erasureServerPools) PutObjectPart(ctx context.Context, bucket, object,
 		return PartInfo{}, err
 	}
 
+	// Add the object part to the cache
+	z.multipartUploadsCache.Set(uploadID, MultipartUpload{
+		Bucket:   bucket,
+		Object:   object,
+		UploadID: uploadID,
+	})
+
```
harshavardhana: the multipart uploadID must be set in newMultipartUpload() and 
removed from the map when 

- CompleteMultipart happens
- or AbortMultipart happens

We also need a peer call where we can notify our peers that they need to delete their local uploadID cache in memory.

harshavardhana: It should never be set for every PutObjectPart()

cmd/erasure-server-pool.go
```diff
@@ -47,6 +47,82 @@ import (
...
+		c.mutex.Lock()
+		now := time.Now()
+		for key, entry := range c.data {
+			if now.Sub(entry.Timestamp) > expiry {
+				delete(c.data, key)
+			}
+		}
+		c.mutex.Unlock()
+	}
+}
```
harshavardhana: I think what we can do is rely on cleanupStaleUploadsOnDisk() to return []string of uploadIDs, this uploadIDs you can work on the delete from the map instead of doing a new goroutine here. 

```
func (er erasureObjects) cleanupStaleUploadsOnDisk(ctx context.Context, disk StorageAPI) {
```

harshavardhana:
```go
func (s *erasureSets) cleanupStaleUploads(ctx context.Context) {
        timer := time.NewTimer(globalAPIConfig.getStaleUploadsCleanupInterval())
        defer timer.Stop()

        for {
                select {
                case <-ctx.Done():
                        return
                case <-timer.C:
                        var wg sync.WaitGroup
                        for _, set := range s.sets {
                                wg.Add(1)
                                go func(set *erasureObjects) {
                                        defer wg.Done()
                                        if set == nil {
                                                return
                                        }
                                        set.cleanupStaleUploads(ctx)
                                }(set)
                        }
                        wg.Wait()

                        // Reset for the next interval
                        timer.Reset(globalAPIConfig.getStaleUploadsCleanupInterval())
                }
        }
}
```

cmd/erasure-server-pool.go
shtripat: oh ok, will change accordingly.

cmd/erasure-server-pool.go
```diff
@@ -47,6 +47,87 @@ import (
...
+
+// Remove removes given entry from cache
+func (c *MultipartUploadCache) Remove(key string) {
+	c.mutex.RLock()
+	defer c.mutex.RUnlock()
+	delete(c.data, key)
+}
+
+// cleanExpiredEntries removes expired entries every hour
+func (c *MultipartUploadCache) cleanExpiredEntries(z *erasureServerPools) {
```
harshavardhana: You don't need this anymore

harshavardhana (changes requested):
You would also need to modify to notify peers.

cmd/erasure-server-pool.go
shtripat: Is it because, while Abort and Complete we are removing entries from cache?
Also not clear on usage of returned `[]string` of cleaned up upload ids from `cleanupStaleUploads`

shtripat:
> You would also need to modify to notify peers.

Is it like `[]string` returned from `cleanupStaleUploads` to be sent across peers for their removal?

cmd/erasure-server-pool.go
harshavardhana: why not? if you have two then they will conflict, you only need one source of truth.

harshavardhana:
> Is it like `[]string` returned from `cleanupStaleUploads` to be sent across peers for their removal?

Correct the ones you just expired, you are letting others know that you did that if they have any local reference to remove it as well.

harshavardhana:
Remember newMultipart can happen on one server and complete multipart can happen on another server and abort also can happen on another server, so you would end up with stale UploadIDs on different nodes.

shtripat:
> > Is it like `[]string` returned from `cleanupStaleUploads` to be sent across peers for their removal?
> 
> Correct the ones you just expired, you are letting others know that you did that if they have any local reference to remove it as well.

The last commit, I am sending a notification to other nodes for their cleanup.

cmd/erasure-server-pool.go
```diff
@@ -47,6 +47,71 @@ import (
...
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+	c.data[key] = MultipartUploadCacheEntry{
+		Upload:    value,
+		Timestamp: time.Now(),
+	}
+}
+
+// Get retrieves a value from the cache if it has not expired
+func (c *MultipartUploadCache) Get(key string) (MultipartUpload, bool) {
```
klauspost: Is this called from anywhere?

cmd/erasure-server-pool.go
```diff
@@ -1713,8 +1784,16 @@ func (z *erasureServerPools) ListMultipartUploads(ctx context.Context, bucket, p
...
-		poolResult.Uploads = append(poolResult.Uploads, result.Uploads...)
+		// poolResult.Uploads = append(poolResult.Uploads, result.Uploads...)
+		uploads = append(uploads, result.Uploads...)
 	}
+
+	for _, upload := range uploads {
+		if ok := z.multipartUploadsCache.Has(upload.UploadID); ok {
+			poolResult.Uploads = append(poolResult.Uploads, upload)
+		}
+	}
```
klauspost: I don't understand how this works.

A) The `multipartUploadsCache` is local only, correct? I don't see anything distributing cached entries or any hashing to have it placed dterministically.
B) An upload will only be cached on one server in cluster. How can we filter, if the upload will only be cached on a random server?
C) What if a server was restarted?

I may be missing something, but this seems to assume that the NewMultipartUpload and ListMultipartUploads happen on the same server, which is unlikely.

Both distributing to all servers and asking all servers seems like a non-scalable approach.

klauspost (changes requested):
I feel like I am missing something. Where does the caching come in?

I only see extra code being added for housekeeping (setting/deleting entries), but nothing that provides a speedup by using the cache.

harshavardhana:
> I feel like I am missing something. Where does the caching come in?
> 
> I only see extra code being added for housekeeping (setting/deleting entries), but nothing that provides a speedup by using the cache.

This is not a speed up, it is to provide listMultipartUploads() output when there is no object name provided.

@klauspost

harshavardhana:
This is mainly for fix the behavior of this API for certain section of applications like alluxio underfs.

klauspost:
> This is mainly for fix the behavior of this API for certain section of applications like alluxio underfs.

Okie, but how does it work across a cluster?

harshavardhana:
> Okie, but how does it work across a cluster?

it is just per node basis @klauspost we just have to send some non-empty list here, the values themselves don't need to be "cluster" consistent.

cmd/erasure-server-pool.go
shtripat: Not used, can be removed

klauspost:
Conflicts and 

```
2024-08-09T10:16:09.0411135Z === RUN   TestListMultipartUploadsHandler
2024-08-09T10:16:09.0412005Z     bucket-handlers_test.go:490: Test 2: ErasureSD: Expected the response status to be `404`, but instead found `200`
2024-08-09T10:16:09.0413146Z     bucket-handlers_test.go:507: Test 2: ErasureSD: Expected the response status to be `404`, but instead found `200`
2024-08-09T10:16:09.0414256Z     bucket-handlers_test.go:490: Test 2: Erasure: Expected the response status to be `404`, but instead found `200`
2024-08-09T10:16:09.0415357Z     bucket-handlers_test.go:507: Test 2: Erasure: Expected the response status to be `404`, but instead found `200`
2024-08-09T10:16:09.0416171Z --- FAIL: TestListMultipartUploadsHandler (0.14s)
```

shtripat:
Looks like mint is not happy. will debug and send separate PR for the same.

cmd/bucket-handlers_test.go
```diff
@@ -369,7 +369,7 @@ func testListMultipartUploadsHandler(obj ObjectLayer, instanceType, bucketName s
 			maxUploads:         "0",
 			accessKey:          credentials.AccessKey,
 			secretKey:          credentials.SecretKey,
-			expectedRespStatus: http.StatusNotFound,
```
harshavardhana: why is this changed? a non-existent bucket must return 404

harshavardhana (commented):
A non-existent bucket must return error for ListMultipartUploads()

cmd/bucket-handlers_test.go
shtripat: Test were failing as minio returning 200 OK in this case. I feel it's because we return from cache now.

shtripat:
> A non-existent bucket must return error for ListMultipartUploads()

Oh ok, so need to add a check for bucket existence and return 404 if bucket does not exist, right?

cmd/peer-rest-server.go
```diff
@@ -905,6 +906,23 @@ func (s *peerRESTServer) ReloadPoolMetaHandler(mss *grid.MSS) (np grid.NoPayload
...
+	if objAPI == nil {
+		return np, grid.NewRemoteErr(errServerNotInitialized)
+	}
+
+	pools, ok := objAPI.(*erasureServerPools)
+	if !ok {
+		return np, grid.NewRemoteErr(errors.New("not a pooled setup"))
+	}
+
+	uploadID := mss.Get(peerRESTCleanUploadID)
```
vadmeste, line 920: can we return an error if peerRESTCleanUploID is empty ? (easier for debugging issues)

cmd/erasure-server-pool.go
```diff
@@ -47,6 +47,59 @@ import (
...
+func (c *MultipartUploadCache) Has(key string) bool {
+	c.mutex.RLock()
+	defer c.mutex.RUnlock()
+	_, exists := c.data[key]
+	return exists
+}
+
+// Remove removes given entry from cache
+func (c *MultipartUploadCache) Remove(key string) {
+	c.mutex.RLock()
```
vadmeste: s/RLock/Lock

cmd/erasure-server-pool.go
```diff
@@ -47,6 +47,59 @@ import (
...
+	defer c.mutex.RUnlock()
```
vadmeste: s/RLock/Lock

cmd/erasure-server-pool.go
```diff
@@ -1706,15 +1764,39 @@ func (z *erasureServerPools) ListMultipartUploads(ctx context.Context, bucket, p
...
-
 	poolResult := ListMultipartsInfo{}
 	poolResult.MaxUploads = maxUploads
 	poolResult.KeyMarker = keyMarker
 	poolResult.Prefix = prefix
 	poolResult.Delimiter = delimiter
+
+	// if no prefix provided, return the list from cache
+	if prefix == "" {
+		for _, entry := range z.multipartUploadsCache.data {
```
vadmeste, line 1773: I guess access to data should be protected

vadmeste (commented):
Can you elaborate more what this PR does ? 

Also it looks like we need to remember that this can be a round robin loaded balancer cluster, if I am not mistaken, ListMultipartUploads will return empty when the request lands to a random node.

shtripat:
> Can you elaborate more what this PR does ?
> 
> Also it looks like we need to remember that this can be a round robin loaded balancer cluster, if I am not mistaken, ListMultipartUploads will return empty when the request lands to a random node.

https://github.com/minio/minio/pull/20181#issuecomment-2269565179 explains it. Its node local cache and mainly for fix the behavior of this API for certain section of applications like alluxio underfs.

cmd/peer-rest-server.go
shtripat: Ack. Will add the same

cmd/erasure-server-pool.go
shtripat: Ack

harshavardhana:
> Also it looks like we need to remember that this can be a round robin loaded balancer cluster, if I am not mistaken, ListMultipartUploads will return empty when the request lands to a random node.

the idea is not meant to be a consistent view of everything that is present at the multipart uploads namespace. IT is mainly to satisfy a class of applications such as alluxio/underfs. 

just like how we have done changes for spark/hadoop specific behavior.

harshavardhana:
there are bunch of issues with this PR addressed them and simplified the approach here https://github.com/minio/minio/pull/20407

shtripat:
Closing in favor of https://github.com/minio/minio/pull/20407

PR #20227, opened 2024-08-07
implement a safer completeMultipart implementation

### LINKED ISSUE #17416: [SNSD] A crash __after__ open("xl.meta", O_TRUNC), and __before__ write() would result in lost of metadata and hence the object.
## NOTE
A crash in function  https://github.com/minio/minio/blob/f32efd5429a62bd1fe8a159e9c75e364c27e0f67/cmd/xl-storage.go#LL1925C1-L1925C29 after open with O_TRUNC, before writing new contents to xl.meta, would result in loss of data.

This inline update is currently being done for DeleteVersion and other operations. 

## Expected Behavior

On a crash between open("xl.meta", O_TRUNC) and write(), we should not loss the object.

## Current Behavior

xl.meta is truncated to 0 bytes post crash after opening it with O_TRUNC.

## Possible Solution

Old version of xl.meta is to be replaced atomically with new version prepared in .minio.sys/tmp location. 

Basically, put a break point for write calls to xl.meta in gdb, and kill minio when it hits the breakpoint.

These were the versions originally on a object:

```
$> mc ls local/bktv/ --insecure --versions
[2023-06-14 03:35:02 EDT] 475KiB STANDARD 63804668-26d5-4ad4-80f1-2de36eb10137 v2 PUT CloudStg
[2023-06-14 03:34:50 EDT] 475KiB STANDARD e3e0dbb5-a0be-4aba-a887-c32433adbc0c v1 PUT CloudStg
```

Attached gdb to mino, and put a breakpint on 1953.

```
(gdb) b cmd/xl-storage.go:1953
Breakpoint 6 at 0x339c4ca: file minio-RELEASE.2023-03-13T19-46-17Z/cmd/xl-storage.go, line 1954.
(gdb) c

```
Done delete versions from a client:
`$> mc rm local/bktv/CloudStg --insecure --version-id e3e0dbb5-a0be-4aba-a887-c32433adbc0c`

See that we hit a break point for that particular object's xl.meta:

```
1954            if err != nil {
(gdb) bt
#0  github.com/minio/minio/cmd.(*xlStorage).writeAll (s=0xc003cf2780, ctx=..., volume="bktv",
    path="CloudStg/xl.meta", b=[]uint8 = {...}, sync=true, err=...)
    at /home/xxx/minio-RELEASE.2023-03-13T19-46-17Z/cmd/xl-storage.go:1954
#1  0x000000000339c85f in github.com/minio/minio/cmd.(*xlStorage).WriteAll (s=0xc003cf2780, ctx=...,
    volume="bktv", path="CloudStg/xl.meta", b=[]uint8 = {...}, err=...)
    at /home/xxx/minio-RELEASE.2023-03-13T19-46-17Z/cmd/xl-storage.go:1972
#2  0x000000000338f2df in github.com/minio/minio/cmd.(*xlStorage).deleteVersions (s=0xc003cf2780, ctx=...,
    volume="bktv", path="CloudStg", fis=[]github.com/minio/minio/cmd.FileInfo = {...}, ~r0=...)
    at /home/xxx/minio-RELEASE.2023-03-13T19-46-17Z/cmd/xl-storage.go:1006
#3  0x000000000338f66f in github.com/minio/minio/cmd.(*xlStorage).DeleteVersions (s=0xc003cf2780, ctx=...,
    volume="bktv", versions=[]github.com/minio/minio/cmd.FileInfoVersions = {...}, ~r0=[]error)
    at /home/xxx/minio-RELEASE.2023-03-13T19-46-17Z/cmd/xl-storage.go:1027
#4  0x000000000332cec5 in github.com/minio/minio/cmd.(*xlStorageDiskIDCheck).DeleteVersions (p=0xc00015f680,
    ctx=..., volume="bktv", versions=[]github.com/minio/minio/cmd.FileInfoVersions = {...}, errs=[]error = {...})
    at /home/xxx/minio-RELEASE.2023-03-13T19-46-17Z/cmd/xl-storage-disk-id-check.go:418
#5  0x0000000003017677 in github.com/minio/minio/cmd.erasureObjects.DeleteObjects.func1 (index=0, disk=...)
    at /home/xxx/minio-RELEASE.2023-03-13T19-46-17Z/cmd/erasure-object.go:1437
#6  0x00000000030173f0 in github.com/minio/minio/cmd.erasureObjects.DeleteObjects.func2 ()
--Type <RET> for more, q to quit, c to continue without paging--
   /home//xxx/minio-RELEASE.2023-03-13T19-46-17Z/cmd/erasure-object.go:1452
#7  0x000000000047b961 in runtime.goexit () at /home/xxx/golang/go/src/runtime/asm_amd64.s:1598
#8  0x0000000000000000 in ?? ()
(gdb) q
```

Killed minio process here, (after Open and before Write):

After the kill xl.meta is now a zero byte file as you can see below.

```
$>  ls -l dir/bktv/CloudStg/
total 0
drwxrwxr-x. 2 ratna ratna 28 Jun 14 01:35 ee430c88-6d0b-4d17-b170-bfc974c300b1
-rw-rw-r--. 1 ratna ratna  0 Jun 14 01:44 xl.meta
```

After restarting the minio server, listing versions shows empty output.

```
$> mc ls local/bktv/CloudStg --insecure --versions
$>
```

## Context

We are losing objects because of a crash unrelated to minio process.

## Regression

 
No, it is not a regression.

## Your Environment

* Version used (`minio --version`):
* Server setup and configuration:

setup:

"Single-Node Single-Drive" minio configuration started with "minio server dir"

OS version
`4.18.0-425.10.1.el8_7.x86_64`

Minio version

```
minio version DEVELOPMENT.2023-06-13T05-27-54Z (commit-id=e05a1aa08d11b3b17fb2e04351e919b2fdadaab2)
Runtime: go1.20.5 linux/amd64
Copyright: 2015-2023 MinIO, Inc.
```

### LINKED ISSUE #20091: health checks, multipart uploads, and retries, when Minio node cycles
## Description
I've observed an issue when performing a rolling update to my production Minio cluster, where client uploads ended up seeing request failures, while I had assumed the update would be transparent and fault tolerant. So I then reduced my test case to a smaller configuration.

I have the following simplified setup:

1. 3 node minio cluster
2. Minio sidekick loadbalancer pointing at minio cluster, using `/minio/healthy/ready` check
3. `mc` client performing an `mc cp` command, in a loop, pointing at sidekick instance

With the `mc` performing a copy of a 30MB file in a loop, I cycle one of the minio nodes to simulate a rolling update. The goal is to have uninterrupted copy operations on the client, as sidekick detects and removes a down node, and adds it back later when healthy. 

## Expected Behavior
I would expect to see sidekick respond to the minio node going down. But I would not expect to see any failed requests occurring on the client side during object uploads. I would expect the client retries to handle any temporary failure before becoming fatal to the caller.
I expect to be able to slowly roll nodes in a minio cluster without causing fatal errors to clients.

I would hope to not have to write special error and retry handling in each of our clients, to deal with transient request failures.

## Current Behavior
Sidekick logs "502 Bad Gateway" when minio node goes down. This might be considered expected. 

After bringing the minio node back up, I will see request errors on the Minio node, using `mc admin trace`:

```
[REQUEST s3.CompleteMultipartUpload] [2024-07-12T17:27:49.900] [Client IP: X.X.X.X]
POST /path/key?uploadId=M2Y4MGY3MzMtZWRjMS00YTQ1LWFkZmItNWIxNjY3ZmQwZWVhLjU1NzM3MjA0LTI1YzAtNGVhYy04ZGZiLWI1MWM3MTM1ODljOQ
Proto: HTTP/1.1
Host: <host>:9010
X-Real-Ip: X.X.X.X
Authorization: <...>
Content-Length: 267
X-Forwarded-For: X.X.X.X, X.X.X.X
X-Forwarded-Host: <host>:9010
Accept-Encoding: zstd,gzip
User-Agent: MinIO (linux; amd64) minio-go/v7.0.73 mc/DEVELOPMENT.GOGET
X-Amz-Content-Sha256: 7315c823224b844add1645658a85fc4b5a832930a7ef58a19e6f2531f3604785
X-Amz-Date: 20240712T052749Z
<CompleteMultipartUpload xmlns="http://s3.amazonaws.com/doc/2006-03-01/"><Part><PartNumber>1</PartNumber><ETag>e637d8911e1566418c2d050952051467</ETag></Part><Part><PartNumber>2</PartNumber><ETag>0e27318879caf674aab5c2e35b979b85</ETag></Part></CompleteMultipartUpload>

[RESPONSE] [2024-07-12T17:27:49.902] [ Duration 1.699ms TTFB 1.688781ms ↑ 416 B  ↓ 465 B ]
400 Bad Request
Content-Type: application/xml
Server: MinIO
Vary: Origin,Accept-Encoding
X-Content-Type-Options: nosniff
Accept-Ranges: bytes
Content-Length: 465
X-Amz-Request-Id: 17E15FFC4F62F55B
X-Xss-Protection: 1; mode=block
Strict-Transport-Security: max-age=31536000; includeSubDomains
X-Amz-Id-2: 2de226a8c93748c02709a9643005ddcbb6b35f079c2973a359b7d7309028e074
<Error><Code>InvalidPart</Code><Message>One or more of the specified parts could not be found.  The part may not have been uploaded, or the specified entity tag may not match the part&#39;s entity tag.</Message><Key>key</Key><BucketName>path</BucketName><Resource>/path/key</Resource><RequestId>17E15FFC4F62F55B</RequestId><HostId>2de226a8c93748c02709a9643005ddcbb6b35f079c2973a359b7d7309028e074</HostId></Error>
```

On the client, I get the same fatal failure propagated back. There does not seem to be any form of a retry for the temporary failure. And I can't tell if the failure is caused by the upload being inflight when the server goes down, or the health check on sidekick green-lighting too soon and causing a bad request to start on the node after it is immediately back online.

## Possible Solution
I've done further investigations via the Go SDK and narrowed it down to multipart uploads. The client switches to multipart upload when the object size is greater than 16MB. When I disable multipart, to seems to mitigate the failure when cycling the minio node.  I am not sure if multipart transfers have proper transient failure handling in the clients. The retry logic in `minio/minio-go` suggest a 400 is not something that is retried automatically.

I've also added more extensive error handling on the client to check for codes like 400 InvalidPart, 404 NoSuchUpload, and 503 SlowDownRead, and a set of retries for those errors that can happen when retrying multipart requests. I'm not sure if end users should have to be doing this or if it should be part of the client, or if it should not happen at all in the first place. 

Here is a link to the Go SDK code I have used to test workarounds: https://pastebin.com/L6azFF4P

You will see there are controls for whether multipart should be disabled, and the request size.

1. Start 3 minio nodes
2. Start sidekick
3. Use either:
   1. my test code: `go run test_code.go`
   2. mc cp <30mb.file> minio bucket
5. Alternate between killing one of the minio nodes, and bringing it back up again, and observe client-side errors

## Context
We are releasing a new production stack that uses Minio as a backend storage. it is important for me to be confident that cycling any one Minio node will not cause request failures at the frontend of the application. I want to achieve proper redundancy and I am not sure if I have to implement more custom error and retry handling in each client (we use Go, Python, and C++).

## Regression
No

## Your Environment
* Minio server: docker minio/minio Version: RELEASE.2024-02-26T09-33-48Z (go1.21.7 linux/amd64) with host networking
* Minio sidekick: docker quay.io/minio/sidekick:v7.0.0
* `mc` latest version DEVELOPMENT.GOGET (commit-id=DEVELOPMENT.GOGET)
* Go SDK: github.com/minio/minio-go/v7 v7.0.73

harshavardhana:

## Description
implement a safer completeMultipart implementation

## Motivation and Context
- optimize writing part.N.meta by writing both part.N
  and its meta in sequence without network component.

- remove part.N.meta, part.N which were partially success
  ful, in quorum loss situations during renamePart()

- allow for strict read quorum check arbitrated via ETag
  for the given part number, this makes it double safer
  upon final commit.

- return an appropriate error when read quorum is missing,
  instead of returning InvalidPart{}, which is non-retryable
  error. This kind of situation can happen when many
  nodes are going offline in rotation, an example of such
  a restart() behavior is statefulset updates in k8s.

fixes #20091

## How to test this PR?
CI/CD handles this properly, the way to reproduce issue start 4 node MinIO and continuously restart it 

```sh
#!/bin/bash

alias=$1

minio_wait() {
	mc ready $1
	mc ready $1
	mc ready $1
	mc ready $1
}

bounce() {
	docker stop $1
	sleep 1
	docker start $1
	minio_wait $alias
}

while true; do
	bounce tmp-minio1-1
	bounce tmp-minio2-1
	bounce tmp-minio3-1
	bounce tmp-minio4-1
done
```

```go
package main

import (
	"bytes"
	"context"
	"errors"
	"fmt"
	"log"
	"math/rand"
	"os"
	"os/signal"
	"sync"
	"time"

	mclient "github.com/minio/minio-go/v7"
	mcreds "github.com/minio/minio-go/v7/pkg/credentials"
)

const (
	Host      = "localhost:8080"
	AccessKey = "minioadmin"
	SecretKey = "minioadmin"
)

const (
	NWorkers         = 10
	ReqSize          = 1 << 20 * 10
	DisableMultipart = false
)

func main() {
	mc, err := mclient.New(Host, &mclient.Options{
		Creds:  mcreds.NewStaticV4(AccessKey, SecretKey, ""),
		Secure: false,
	})
	if err != nil {
		log.Fatal(err)
	}

	ctx := context.Background()
	ctx, stop := signal.NotifyContext(ctx, os.Interrupt)
	defer stop()

	msgs := make(chan string, 100)

	var wg sync.WaitGroup
	for i := 0; i < NWorkers; i++ {
		wg.Add(1)

		// start request worker
		go func(i int) {
			defer wg.Done()
			data := make([]byte, ReqSize)
			if _, err := rand.Read(data); err != nil {
				log.Fatalf("data prep err: %v", err)
			}
			buf := bytes.NewReader(data)
			name := fmt.Sprintf("obj%d", i)

			opts := mclient.PutObjectOptions{
				//PartSize:         (1 << 20) * 100,
				DisableMultipart: DisableMultipart,
				PartSize:         (1 << 20) * 5,
			}
			var err error
			for {
				if ctx.Err() != nil {
					return // cancelled
				}

				// retry certain error types
				for tries := 0; tries < 3; tries++ {
					localCtx, done := context.WithTimeout(ctx, 1500*time.Second)
					_, err = mc.PutObject(localCtx, "testing", name, buf, int64(len(data)), opts)
					done()
					if err == nil {
						break
					}

					if errors.Is(err, context.DeadlineExceeded) {
						msgs <- fmt.Sprintf("Deadline (%d: try #%d): %v", i+1, tries, err)
						continue
					}
					eResp := mclient.ToErrorResponse(err)
					if eResp.StatusCode == 400 && eResp.Code == "InvalidPart" {
						msgs <- fmt.Sprintf("400 InvalidPart: (%d: try #%d, obj: %v): %v", i+1, tries, name, err)
						continue
					}
					if eResp.StatusCode == 404 && eResp.Code == "NoSuchUpload" {
						msgs <- fmt.Sprintf("404 NoSuchUpload: (%d: try #%d): %v", i+1, tries, err)
						continue
					}
					if eResp.StatusCode == 503 && eResp.Code == "SlowDownRead" {
						time.Sleep(2 * time.Second)
						msgs <- fmt.Sprintf("503 SlowDownRead: (%d: try #%d): %v", i+1, tries, err)
						continue
					}
					msgs <- fmt.Sprintf("Error: (%d: try #%d): %v", i+1, tries, err)
					break
				}

				if errors.Is(err, context.Canceled) {
					return
				}
				if errors.Is(err, context.DeadlineExceeded) {
					msgs <- fmt.Sprintf("Deadline (%d): %v: skipping request", i+1, err)
					continue
				}
				if err != nil {
					eResp := mclient.ToErrorResponse(err)
					log.Fatalf("Copy error: %#v", eResp)
				}

				msgs <- fmt.Sprintf("OK (%d)", i+1)
				r := rand.Intn(400) + 50
				time.Sleep(time.Duration(r) * time.Millisecond)
			}

		}(i)
	}

	// sync handle messages from workers
	go func() {
		for {
			select {
			case <-ctx.Done():
				return
			case m := <-msgs:
				fmt.Println(m)
			}
		}
	}()

	wg.Wait()
}
```

```yaml
version: '3.7'

# Settings and configurations that are common for all containers
x-minio-common: &minio-common
  image: y4m4/minio:dev
  command: server --address ":9000" --console-address ":9090" http://minio{1...4}/data{1...2}
  expose:
    - "9000"
    - "9090"
  environment:
    MINIO_ROOT_USER: minioadmin
    MINIO_ROOT_PASSWORD: minioadmin
  # healthcheck:
  #  test: ["CMD", "mc", "ready", "local"]
  #  interval: 5s
  #  timeout: 5s
  #  retries: 5

# starts 4 docker containers running minio server instances.
# using nginx reverse proxy, load balancing, you can access
# it through port 9000.
services:
  minio1:
    <<: *minio-common
    hostname: minio1
    volumes:
      - data1-1:/data1
      - data1-2:/data2

  minio2:
    <<: *minio-common
    hostname: minio2
    volumes:
      - data2-1:/data1
      - data2-2:/data2

  minio3:
    <<: *minio-common
    hostname: minio3
    volumes:
      - data3-1:/data1
      - data3-2:/data2

  minio4:
    <<: *minio-common
    hostname: minio4
    volumes:
      - data4-1:/data1
      - data4-2:/data2

  nginx:
    image: nginx:1.19.2-alpine
    hostname: nginx
    volumes:
      - ./nginx.conf:/etc/nginx/nginx.conf:ro
    ports:
      - "9000:9000"
      - "9090:9090"
    depends_on:
      - minio1
      - minio2
      - minio3
      - minio4

  sidekick:
    command: --health-path=/minio/health/cluster --address :8000 http://minio{1...4}:9000
    image: quay.io/minio/sidekick:v7.0.0
    hostname: sidekick
    expose:
    - "8000"
    ports:
      - "8080:8000"
    depends_on:
      - minio1
      - minio2
      - minio3
      - minio4
      - nginx

## By default this config uses default local driver,
## For custom volumes replace with volume driver configuration.
volumes:
  data1-1:
  data1-2:
  data2-1:
  data2-2:
  data3-1:
  data3-2:
  data4-1:
  data4-2:
```

## Types of changes

## Checklist:

### DIFF
--- a/cmd/erasure-multipart.go
+++ b/cmd/erasure-multipart.go
@@ -80,6 +80,14 @@ func (er erasureObjects) checkUploadIDExists(ctx context.Context, bucket, object
 		return fi, nil, err
 	}
 
+	if readQuorum < 0 {
+		return fi, nil, errErasureReadQuorum
+	}
+
+	if writeQuorum < 0 {
+		return fi, nil, errErasureWriteQuorum
+	}
+
 	quorum := readQuorum
 	if write {
 		quorum = writeQuorum
@@ -88,14 +96,13 @@ func (er erasureObjects) checkUploadIDExists(ctx context.Context, bucket, object
 	// List all online disks.
 	_, modTime, etag := listOnlineDisks(storageDisks, partsMetadata, errs, quorum)
 
-	var reducedErr error
 	if write {
-		reducedErr = reduceWriteQuorumErrs(ctx, errs, objectOpIgnoredErrs, writeQuorum)
+		err = reduceWriteQuorumErrs(ctx, errs, objectOpIgnoredErrs, writeQuorum)
 	} else {
-		reducedErr = reduceReadQuorumErrs(ctx, errs, objectOpIgnoredErrs, readQuorum)
+		err = reduceReadQuorumErrs(ctx, errs, objectOpIgnoredErrs, readQuorum)
 	}
-	if reducedErr != nil {
-		return fi, nil, reducedErr
+	if err != nil {
+		return fi, nil, err
 	}
 
 	// Pick one from the first valid metadata.
@@ -490,9 +497,10 @@ func (er erasureObjects) newMultipartUpload(ctx context.Context, bucket string, 
 	uploadIDPath := er.getUploadIDDir(bucket, object, uploadUUID)
 
 	// Write updated `xl.meta` to all disks.
-	if _, err := writeUniqueFileInfo(ctx, onlineDisks, bucket, minioMetaMultipartBucket, uploadIDPath, partsMetadata, writeQuorum); err != nil {
+	if _, err := writeAllMetadata(ctx, onlineDisks, bucket, minioMetaMultipartBucket, uploadIDPath, partsMetadata, writeQuorum); err != nil {
 		return nil, toObjectErr(err, bucket, object)
 	}
+
 	return &NewMultipartUploadResult{
 		UploadID:     uploadID,
 		ChecksumAlgo: userDefined[hash.MinIOMultipartChecksum],
@@ -513,7 +521,7 @@ func (er erasureObjects) NewMultipartUpload(ctx context.Context, bucket, object 
 }
 
 // renamePart - renames multipart part to its relevant location under uploadID.
-func renamePart(ctx context.Context, disks []StorageAPI, srcBucket, srcEntry, dstBucket, dstEntry string, writeQuorum int) ([]StorageAPI, error) {
+func (er erasureObjects) renamePart(ctx context.Context, disks []StorageAPI, srcBucket, srcEntry, dstBucket, dstEntry string, optsMeta []byte, writeQuorum int) ([]StorageAPI, error) {
 	g := errgroup.WithNErrs(len(disks))
 
 	// Rename file on all underlying storage disks.
@@ -523,60 +531,25 @@ func renamePart(ctx context.Context, disks []StorageAPI, srcBucket, srcEntry, ds
 			if disks[index] == nil {
 				return errDiskNotFound
 			}
-			return disks[index].RenameFile(ctx, srcBucket, srcEntry, dstBucket, dstEntry)
+			return disks[index].RenamePart(ctx, srcBucket, srcEntry, dstBucket, dstEntry, optsMeta)
 		}, index)
 	}
 
 	// Wait for all renames to finish.
 	errs := g.Wait()
 
-	// Do not need to undo partial successful operation since those will be cleaned up
-	// in 24hrs via multipart cleaner, never rename() back to `.minio.sys/tmp` as there
-	// is no way to clean them.
-
-	// We can safely allow RenameFile errors up to len(er.getDisks()) - writeQuorum
-	// otherwise return failure. Cleanup successful renames.
-	return evalDisks(disks, errs), reduceWriteQuorumErrs(ctx, errs, objectOpIgnoredErrs, writeQuorum)
-}
-
-// writeAllDisks - writes 'b' to all provided disks.
-// If write cannot reach quorum, the files will be deleted from all disks.
-func writeAllDisks(ctx context.Context, disks []StorageAPI, dstBucket, dstEntry string, b []byte, writeQuorum int) ([]StorageAPI, error) {
-	g := errgroup.WithNErrs(len(disks))
-
-	// Write file to all underlying storage disks.
-	for index := range disks {
-		index := index
-		g.Go(func() error {
-			if disks[index] == nil {
-				return errDiskNotFound
-			}
-			return disks[index].WriteAll(ctx, dstBucket, dstEntry, b)
-		}, index)
+	paths := []string{
+		dstEntry,
+		dstEntry + ".meta",
 	}
 
-	// Wait for all renames to finish.
-	errs := g.Wait()
-
-	// We can safely allow RenameFile errors up to len(er.getDisks()) - writeQuorum
-	// otherwise return failure. Cleanup successful renames.
 	err := reduceWriteQuorumErrs(ctx, errs, objectOpIgnoredErrs, writeQuorum)
-	if errors.Is(err, errErasureWriteQuorum) {
-		// Remove all written
-		g := errgroup.WithNErrs(len(disks))
-		for index := range disks {
-			if disks[index] == nil || errs[index] != nil {
-				continue
-			}
-			index := index
-			g.Go(func() error {
-				return disks[index].Delete(ctx, dstBucket, dstEntry, DeleteOptions{Immediate: true})
-			}, index)
-		}
-		// Ignore these errors.
-		g.WaitErr()
+	if err != nil {
+		er.cleanupMultipartPath(ctx, paths...)
 	}
 
+	// We can safely allow RenameFile errors up to len(er.getDisks()) - writeQuorum
+	// otherwise return failure. Cleanup successful renames.
 	return evalDisks(disks, errs), err
 }
 
@@ -732,19 +705,6 @@ func (er erasureObjects) PutObjectPart(ctx context.Context, bucket, object, uplo
 
 	// Rename temporary part file to its final location.
 	partPath := pathJoin(uploadIDPath, fi.DataDir, partSuffix)
-	onlineDisks, err = renamePart(ctx, onlineDisks, minioMetaTmpBucket, tmpPartPath, minioMetaMultipartBucket, partPath, writeQuorum)
-	if err != nil {
-		if errors.Is(err, errFileNotFound) {
-			// An in-quorum errFileNotFound means that client stream
-			// prematurely closed and we do not find any xl.meta or
-			// part.1's - in such a scenario we must return as if client
-			// disconnected. This means that erasure.Encode() CreateFile()
-			// did not do anything.
-			return pi, IncompleteBody{Bucket: bucket, Object: object}
-		}
-
-		return pi, toObjectErr(err, minioMetaMultipartBucket, partPath)
-	}
 
 	md5hex := r.MD5CurrentHexString()
 	if opts.PreserveETag != "" {
@@ -766,15 +726,22 @@ func (er erasureObjects) PutObjectPart(ctx context.Context, bucket, object, uplo
 		Checksums:  r.ContentCRC(),
 	}
 
-	fi.Parts = []ObjectPartInfo{partInfo}
-	partFI, err := fi.MarshalMsg(nil)
+	partFI, err := partInfo.MarshalMsg(nil)
 	if err != nil {
 		return pi, toObjectErr(err, minioMetaMultipartBucket, partPath)
 	}
 
-	// Write part metadata to all disks.
-	onlineDisks, err = writeAllDisks(ctx, onlineDisks, minioMetaMultipartBucket, partPath+".meta", partFI, writeQuorum)
+	onlineDisks, err = er.renamePart(ctx, onlineDisks, minioMetaTmpBucket, tmpPartPath, minioMetaMultipartBucket, partPath, partFI, writeQuorum)
 	if err != nil {
+		if errors.Is(err, errFileNotFound) {
+			// An in-quorum errFileNotFound means that client stream
+			// prematurely closed and we do not find any xl.meta or
+			// part.1's - in such a scenario we must return as if client
+			// disconnected. This means that erasure.Encode() CreateFile()
+			// did not do anything.
+			return pi, IncompleteBody{Bucket: bucket, Object: object}
+		}
+
 		return pi, toObjectErr(err, minioMetaMultipartBucket, partPath)
 	}
 
@@ -917,7 +884,7 @@ func (er erasureObjects) ListObjectParts(ctx context.Context, bucket, object, up
 
 	g := errgroup.WithNErrs(len(req.Files)).WithConcurrency(32)
 
-	partsInfo := make([]ObjectPartInfo, len(req.Files))
+	partsInfo := make([]*ObjectPartInfo, len(req.Files))
 	for i, file := range req.Files {
 		file := file
 		partN := i + start
@@ -929,21 +896,17 @@ func (er erasureObjects) ListObjectParts(ctx context.Context, bucket, object, up
 				return err
 			}
 
-			var pfi FileInfo
-			_, err = pfi.UnmarshalMsg(buf)
+			pinfo := &ObjectPartInfo{}
+			_, err = pinfo.UnmarshalMsg(buf)
 			if err != nil {
 				return err
 			}
 
-			if len(pfi.Parts) != 1 {
-				return errors.New("invalid number of parts expected 1, got 0")
+			if partN != pinfo.Number {
+				return fmt.Errorf("part.%d.meta has incorrect corresponding part number: expected %d, got %d", partN, partN, pinfo.Number)
 			}
 
-			if partN != pfi.Parts[0].Number {
-				return fmt.Errorf("part.%d.meta has incorrect corresponding part number: expected %d, got %d", partN, partN, pfi.Parts[0].Number)
-			}
-
-			partsInfo[i] = pfi.Parts[0]
+			partsInfo[i] = pinfo
 			return nil
 		}, i)
 	}
@@ -951,7 +914,7 @@ func (er erasureObjects) ListObjectParts(ctx context.Context, bucket, object, up
 	g.Wait()
 
 	for _, part := range partsInfo {
-		if part.Number != 0 && !part.ModTime.IsZero() {
+		if part != nil && part.Number != 0 && !part.ModTime.IsZero() {
 			fi.AddObjectPart(part.Number, part.ETag, part.Size, part.ActualSize, part.ModTime, part.Index, part.Checksums)
 		}
 	}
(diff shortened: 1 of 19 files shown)


harshavardhana:
Thanks to @allanrogerr for a wonderful way to reproduce the issue

cmd/xl-storage.go
```diff
@@ -2113,10 +2119,10 @@ func (s *xlStorage) CreateFile(ctx context.Context, origvolume, volume, path str
 		}
 	}()
 
-	return s.writeAllDirect(ctx, filePath, fileSize, r, os.O_CREATE|os.O_WRONLY|os.O_EXCL, volumeDir)
+	return s.writeAllDirect(ctx, filePath, fileSize, r, os.O_CREATE|os.O_WRONLY|os.O_EXCL, volumeDir, false)
```
harshavardhana, line 2121: O_EXCL means there is no pre-existing file at the destination so we do not need `Truncate()` call.

cmd/xl-storage.go
```diff
@@ -2166,13 +2172,27 @@ func (s *xlStorage) writeAllDirect(ctx context.Context, filePath string, fileSiz
...
 		w.Close()
 		return errMoreData
 	}
 
+	if truncate {
+		// Change the final size of the file, in-case we are overwriting.
+		if err = w.Truncate(written); err != nil {
+			w.Close()
+			return err
+		}
```
harshavardhana: To ensure ftruncate hits the persistent storage Fdatasync() is called

cmd/xl-storage.go
```diff
@@ -2233,10 +2253,17 @@ func (s *xlStorage) writeAll(ctx context.Context, volume string, path string, b
 	}
 
 	if n != len(b) {
+		w.Truncate(0) // to indicate that we did partial write.
 		w.Close()
 		return io.ErrShortWrite
 	}
 
+	// Change the final size of the file, in-case we are overwriting.
+	if err = w.Truncate(int64(n)); err != nil {
```
harshavardhana: Ftruncate with O_SYNC() mostly for xl.meta would persist right away.

allanrogerr:
Reviewing

cmd/xl-storage.go
```diff
@@ -2205,7 +2225,7 @@ func (s *xlStorage) writeAll(ctx context.Context, volume string, path string, b
 		return err
 	}
 
-	flags := os.O_CREATE | os.O_WRONLY | os.O_TRUNC
+	flags := os.O_CREATE | os.O_WRONLY
```
klauspost: It seems like you are adding another syscall to every write, but I don't think that is needed.

If the file cannot be opened. Nothing happens. If the file is opened, it is truncated either way.

It seems to me you should

A) Keep `O_TRUNC`
B) Truncate(0) if the write fails (but *only* if it fails).

AFAICT that would do the same, but there wouldn't be any need for an additional syscall on the "happy path".

Does that make sense?

cmd/storage-rest-common.go
```diff
@@ -20,7 +20,7 @@ package cmd
 //go:generate msgp -file $GOFILE -unexported
 
 const (
-	storageRESTVersion       = "v62" // Introduce DeleteBulk internode API.
+	storageRESTVersion       = "v63" // Introduce RenamePart and ReadParts API
```
klauspost, line 23: We don't really need to bump this any more, right?

allanrogerr (approved):
Tested the functionality. No errors reported with the `reproducer`.

cmd/xl-storage.go
harshavardhana: That's correct, but it allows protection for situations where we don't make `xl.meta` 0 aggressively but safely later. 

Otherwise to do it another way is to do how we do `.minio.sys/tmp` and then rename()

harshavardhana: Let me revert this change, as its not related to this PR at all @klauspost

harshavardhana: The idea was to try to resolve the issue reported for SNSD, but will send a change that uses a temporary path instead.

klauspost: 👍🏼

cmd/erasure-multipart.go
```diff
@@ -987,6 +950,80 @@ func (er erasureObjects) ListObjectParts(ctx context.Context, bucket, object, up
...
+				continue
+			}
+
+			if pinfo.ETag != "" {
+				partMetaQuorumMap[pinfo.ETag]++
+			} else {
+				partMetaQuorumMap[partMetaPaths[pidx]]++
+			}
+		}
+		if pinfo == nil {
```
krisis: It is possible that more than quorum of them had a valid value except the last disk and we'd incorrectly set this part as invalid.

cmd/erasure-multipart.go
```diff
@@ -987,6 +950,80 @@ func (er erasureObjects) ListObjectParts(ctx context.Context, bucket, object, up
...
+	if err := reduceReadQuorumErrs(ctx, g.Wait(), objectOpIgnoredErrs, readQuorum); err != nil {
+		return nil, err
+	}
+
+	partInfosInQuorum := make([]*ObjectPartInfo, len(partMetaPaths))
+	partMetaQuorumMap := make(map[string]int, len(partNumbers))
+	for pidx := range partMetaPaths {
+		var pinfo *ObjectPartInfo
+		for idx := range disks {
+			if objectPartInfos[idx] == nil || len(objectPartInfos[idx]) == 0 {
```
krisis: len(arr) == 0, where arr is nil
```suggestion
			len(objectPartInfos[idx]) == 0 {
```

cmd/erasure-multipart.go
```diff
@@ -987,6 +950,80 @@ func (er erasureObjects) ListObjectParts(ctx context.Context, bucket, object, up
...
+		}
+		if pinfo == nil {
+			partInfosInQuorum[pidx] = &ObjectPartInfo{
+				Error: InvalidPart{
+					PartNumber: partNumbers[pidx],
+				}.Error(),
+			}
+			continue
+		}
+		if pinfo.ETag != "" && partMetaQuorumMap[pinfo.ETag] >= readQuorum {
```
krisis: IIUC, we can't rely on `pinfo` since it holds the last disk's part information. You need to range over the two maps which maintain counts indexed by etag and path names.

cmd/erasure-multipart.go
harshavardhana: I tried to resolve it, please take a look.

harshavardhana:
I want to get this out in the next release, please help me get this merged thank you.

jiuker (approved):

klauspost (approved):
LGTM

cmd/storage-rest-common.go
harshavardhana: ReadParts is not part of websockets @klauspost as it may be slow and longer.

PR #20232, opened 2024-08-08
Relax quorum agreement on DataDir values

krisis:

## Description
Previously we checked if we had quorum on DataDir value. We are removing this check to allow reading objects with different DataDir values in a few drives (due to a rebalance-stop race bug) provided their their etags or ModTimes match.
MinIO server never creates objects with different DataDir values. There are other stronger properties tied to data (etag, ModTime, etc) which ensure that we don't read invalid shards.
Thanks @harshavardhana for the initial patch.

## Motivation and Context
A race between rebalance-stop and ongoing rebalance activity could cause the object to be written to the same location (i.e onto itself) instead of writing it to a different pool. A separate PR to address this will be sent. This PR allows for objects affected by this bug to be read as is. Note: Objects affected are < 128KiB in size (i.e inlined/inlinable).

## How to test this PR?
This requires objects affected due to a race in rebalance-stop.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/erasure-metadata.go
+++ b/cmd/erasure-metadata.go
@@ -289,13 +289,14 @@ func findFileInfoInQuorum(ctx context.Context, metaArr []FileInfo, modTime time.
 		mtimeValid := meta.ModTime.Equal(modTime)
 		if mtimeValid || etagOnly {
 			fmt.Fprintf(h, "%v", meta.XLV1)
-			if !etagOnly {
-				// Verify dataDir is same only when mtime is valid and etag is not considered.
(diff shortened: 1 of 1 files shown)


harshavardhana (commented):
We should do one test where the dataDir chosen is different per drive and see if we can still read the content. 

We can manually create such a scenario by setting dataDir per meta while writing and then trying to read from it.

harshavardhana (approved):

klauspost (approved):

zveinn (approved):

PR #20244, opened 2024-08-12
Add support profile deadlines and concurrent operations

klauspost:
## Description

* Allow max 10 seconds to start profiling operations.
* Download up to 16 profiles concurrently, but only allow 10 seconds to each (does not include write time).
* Add cluster info as the first operation.
* Ignore remote download errors.
* Stop remote profiling if request is terminated.

## Motivation and Context

Make profiles more reliable in unreliable environments.

## How to test this PR?

`mc support profile`

## Types of changes


cmd/notification.go
```diff
@@ -313,28 +316,49 @@ func (sys *NotificationSys) DownloadProfilingData(ctx context.Context, writer io
...
-		}
-
-		profilingDataFound = true
-
-		for typ, data := range data {
-			err := embedFileInZip(zipWriter, fmt.Sprintf("profile-%s-%s", client.host.String(), typ), data, 0o600)
+		ng.Go(ctx, func() error {
+			// Give 10 seconds to each remote call.
+			// Errors are logged but not returned.
+			ctx, cancel := context.WithTimeout(ctx, 10*time.Second)
```
zveinn: The downloads can be quite big right ? maybe give it a little more then 10 seconds ?

klauspost: ok, 15 then.

klauspost: I don't want it higher, since we have max 16 concurrent, so that means that wall time can get quite high. It will be

`wall_time = 15s * int((unresponsive_servers + 15) / 16)`.

zveinn: 5 more is 5 more, I approve.

harshavardhana (commented):
We may some of these timeouts configurable, in-case we need in future. The defaults are okay.

zveinn (approved):

harshavardhana (approved):

PR #20259, opened 2024-08-14
Removing the audit log retry mechanism

zveinn:

## Description
The audit log retry mechanism is causing backpressure when the receiver(aka target) is not able to receive logs fast enough. This causes a buildup, which eventually causes retries to be triggered in the Send function.

The retry mechanism having a sleep timer is sub-optimal, but the cause for minios hard-stop of data ingestion was caused by `minio/cmd/handler-api.go: 535` the above mentioned ratelimiter works on a hard-coded value instead of resource usage, so when the audit logs started retrying, total active requests would build up, and the ratelimiter starts dropping requests. Even if we had enough resources to keep going. 

The audit retry mechanism is not needed anymore since the audit log queues will start spawning more processors at 50% capacity. If we're ever in the situation where the queue is full, a retry won't help. 

## Types of changes

## Checklist:

### DIFF
--- a/internal/logger/target/http/http.go
+++ b/internal/logger/target/http/http.go
@@ -562,7 +562,6 @@ func (h *Target) Send(ctx context.Context, entry interface{}) error {
 		return nil
 	}
 
-retry:
 	select {
 	case h.logCh <- entry:
 		h.totalMessages.Add(1)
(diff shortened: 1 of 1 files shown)


harshavardhana (approved):

zveinn:
@harshavardhana linter having problems ?

harshavardhana:
> @harshavardhana linter having problems ?

not problem we have to fix it.

PR #20285, opened 2024-08-18
heal: Reset healing params when a retry is decided

vadmeste:

## Description
Currently, retry healing of a new drive healing does not reset HealedBuckets, 
which means, the next healing retry will skip those buckets. 
This commit wil fix this behavior.

Also the skipped objects counter will also include objects uploaded that 
are uploaded after the healing is started.

Also print a log when healing Started field is zero (unexpected situation) 
and do not skip any object in that case.

## Motivation and Context
Fix new drive retry healing behavior 

## Types of changes

## Checklist:

### DIFF
--- a/cmd/background-newdisks-heal-ops.go
+++ b/cmd/background-newdisks-heal-ops.go
@@ -369,6 +369,7 @@ func (h *healingTracker) toHealingDisk() madmin.HealingDisk {
 		Object:            h.Object,
 		QueuedBuckets:     h.QueuedBuckets,
 		HealedBuckets:     h.HealedBuckets,
+		RetryAttempts:     h.RetryAttempts,
 
 		ObjectsHealed: h.ItemsHealed, // Deprecated July 2021
 		ObjectsFailed: h.ItemsFailed, // Deprecated July 2021
--- a/cmd/background-newdisks-heal-ops.go
+++ b/cmd/background-newdisks-heal-ops.go
@@ -148,6 +148,26 @@ func initHealingTracker(disk StorageAPI, healID string) *healingTracker {
 	return h
 }
 
+func (h *healingTracker) resetHealing() {
+	h.mu.Lock()
+	defer h.mu.Unlock()
+
+	h.ItemsHealed = 0
+	h.ItemsFailed = 0
+	h.BytesDone = 0
+	h.BytesFailed = 0
+	h.ResumeItemsHealed = 0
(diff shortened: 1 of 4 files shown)


cmd/background-newdisks-heal-ops.go
```diff
@@ -143,6 +143,26 @@ func initHealingTracker(disk StorageAPI, healID string) *healingTracker {
...
+	h.ItemsHealed = 0
+	h.ItemsFailed = 0
+	h.BytesDone = 0
+	h.BytesFailed = 0
+	h.ResumeItemsHealed = 0
+	h.ResumeItemsFailed = 0
+	h.ResumeBytesDone = 0
+	h.ResumeBytesFailed = 0
+	h.ItemsSkipped = 0
+	h.BytesSkipped = 0
```
klauspost, line 164: I will be somewhat unintuitive these values reset, but I see your point.

How about separating these stats and have an accumulated as well as one for the current attempt?

harshavardhana (approved):

cmd/background-newdisks-heal-ops.go
vadmeste: > How about separating these stats and have an accumulated as well as one for the current attempt?

in fact, we already print those stats (items healed/failed/skipped) for each retry in the standard output.. probably it is not needed to have a separate stats that accumulate everything. By the retry here is a retry from scratch (all buckets and all objects again)

klauspost: Yeah, but it will look weird that suddenly progress goes down to 0 for everything. I don't know if you still do ETA, but that will for sure be useless.

Do we count attempts? That would also be fine by my (and probably easier).

vadmeste: oh yes I see what you mean now, yes we send the retry counter information now to mc, I guess we could avoid calculating ETA in mc when we are retrying. By the way, in most use cases, the retry will be much faster because most of the objects will be already healed, the healing will be just listing again and only heal the remaining objects

vadmeste:
@harshavardhana @klauspost  in fact, let's merge this first, this will be needed by mc to know this is a retry attempt to avoid showing ETA in this case

https://github.com/minio/madmin-go/pull/300

harshavardhana (approved):

klauspost (approved):
Sure

Issue #20267, opened 2024-08-15
parseObjectAttributes needs to account for the possibility of repeating headers


jimrobinson:
## NOTE

The RFC 2616 (HTTP/1.1) specification defines a section [4.2 Message Headers](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2) that allows for multiple message-header fields with the same field-name if the field-value for the header is defined as a comma-separated list, as is the case for `X-Amz-Object-Attributes`:

> Multiple message-header fields with the same field-name MAY be present in a message if and only if the entire field-value for that header field is defined as a comma-separated list [i.e., #(values)]. It MUST be possible to combine the multiple header fields into one "field-name: field-value" pair, without changing the semantics of the message, by appending each subsequent field-value to the first, each separated by a comma. The order in which header fields with the same field-name are received is therefore significant to the interpretation of the combined field value, and thus a proxy MUST NOT change the order of these field values when a message is forwarded. 

In minio the [parseObjectAttributes](https://github.com/minio/minio/blob/743ddb196ac071db1f6929c62765ad136112b11f/cmd/object-api-options.go#L226-L235) function has a bug on [line 228](https://github.com/minio/minio/blob/743ddb196ac071db1f6929c62765ad136112b11f/cmd/object-api-options.go#L228) where [http#Header.Get](https://pkg.go.dev/net/http#Header.Get), which can only return a single value, is used instead of [http#Header.Values](https://pkg.go.dev/net/http#Header.Values) which can return multiple values.

If a client's request splits the `X-Amz-Object-Attributes` header across multiple values, as happens in [Go AWS SDK v2](https://github.com/aws/aws-sdk-go-v2)'s [GetObjectAttributes](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/s3#Client.GetObjectAttributes) method, minio will only recognize the first header.

## Expected Behavior

If a `/<bucket>/<key>?attributes` request arrives which splits `X-Amz-Object-Attributes` across multiple headers:

```
X-Amz-Object-Attributes: ETag
X-Amz-Object-Attributes: Checksum
X-Amz-Object-Attributes: ObjectParts
X-Amz-Object-Attributes: StorageClass
X-Amz-Object-Attributes: ObjectSize
```

I expect minio to treat this the same way it treats a single comma-separated list:

```
X-Amz-Object-Attributes: ETag,Checksum,ObjectParts,StorageClass,ObjectSize
```

[...]

PR #20295, opened 2024-08-21
parseObjectAttributes should handle multiple header values when X-Amz…

marktheunissen:
…-Object-Attributes is given more than once

## Description

Detailed description in https://github.com/minio/minio/issues/20267, essentially headers where the key is repeated as follows:

```
X-Amz-Object-Attributes: ETag
X-Amz-Object-Attributes: Checksum
X-Amz-Object-Attributes: ObjectParts
X-Amz-Object-Attributes: StorageClass
X-Amz-Object-Attributes: ObjectSize
```

## How to test this PR?

There are tests in functional-tests.go in minio-go, as well as automated test in this PR.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/object-api-options.go
+++ b/cmd/object-api-options.go
@@ -225,9 +225,11 @@ func getAndValidateAttributesOpts(ctx context.Context, w http.ResponseWriter, r 
 
 func parseObjectAttributes(h http.Header) (attributes map[string]struct{}) {
 	attributes = make(map[string]struct{})
-	for _, v := range strings.Split(strings.TrimSpace(h.Get(xhttp.AmzObjectAttributes)), ",") {
-		if v != "" {
-			attributes[v] = struct{}{}
+	for _, headerVal := range h.Values(xhttp.AmzObjectAttributes) {
+		for _, v := range strings.Split(strings.TrimSpace(headerVal), ",") {
+			if v != "" {
+				attributes[v] = struct{}{}
+			}
 		}
 	}
 
--- a/cmd/object-api-options_test.go
+++ b/cmd/object-api-options_test.go
@@ -1,0 +1,79 @@
+// Copyright (c) 2015-2024 MinIO, Inc.
+//
+// This file is part of MinIO Object Storage stack
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package cmd
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"reflect"
+	"testing"
+
+	xhttp "github.com/minio/minio/internal/http"
+)
+
+// TestGetAndValidateAttributesOpts is currently minimal and covers a subset of getAndValidateAttributesOpts(),
+// it is intended to be expanded when the function is worked on in the future.
+func TestGetAndValidateAttributesOpts(t *testing.T) {
+	globalBucketVersioningSys = &BucketVersioningSys{}
+	bucket := minioMetaBucket
+	ctx := context.Background()
+	testCases := []struct {
+		name            string
+		headers         http.Header
+		wantObjectAttrs map[string]struct{}
+	}{
+		{
+			name:            "empty header",
+			headers:         http.Header{},
+			wantObjectAttrs: map[string]struct{}{},
+		},
+		{
+			name: "single header line",
+			headers: http.Header{
+				xhttp.AmzObjectAttributes: []string{"test1,test2"},
+			},
+			wantObjectAttrs: map[string]struct{}{
+				"test1": {}, "test2": {},
+			},
+		},
+		{
+			name: "multiple header lines with some duplicates",
+			headers: http.Header{
+				xhttp.AmzObjectAttributes: []string{"test1,test2", "test3,test4", "test4,test3"},
+			},
+			wantObjectAttrs: map[string]struct{}{
+				"test1": {}, "test2": {}, "test3": {}, "test4": {},
+			},
+		},
+	}
+
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			rec := httptest.NewRecorder()
+			req := httptest.NewRequest("GET", "/test", nil)
+			req.Header = testCase.headers
+
+			opts, _ := getAndValidateAttributesOpts(ctx, rec, req, bucket, "testobject")
+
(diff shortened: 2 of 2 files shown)


cmd/object-api-options.go
```diff
@@ -225,9 +225,11 @@ func getAndValidateAttributesOpts(ctx context.Context, w http.ResponseWriter, r
 
 func parseObjectAttributes(h http.Header) (attributes map[string]struct{}) {
 	attributes = make(map[string]struct{})
-	for _, v := range strings.Split(strings.TrimSpace(h.Get(xhttp.AmzObjectAttributes)), ",") {
-		if v != "" {
-			attributes[v] = struct{}{}
+	for _, headerVal := range h.Values(xhttp.AmzObjectAttributes) {
+		for _, v := range strings.Split(strings.TrimSpace(headerVal), ",") {
+			if v != "" {
+				attributes[v] = struct{}{}
```
jiuker, line 231: Should `TrimSpace` as well maybe.

klauspost:
Doesn't this also affect [signatures](https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html)?

In particular how `CanonicalHeaders` is calculated. Mainly sorting? They only say "sort by name". Is there any magic concat method?

I seems like `SignedHeaders` should only include it once. But how will it work?

ramondeklein (approved):

cmd/object-api-options.go
```diff
@@ -225,9 +225,11 @@ func getAndValidateAttributesOpts(ctx context.Context, w http.ResponseWriter, r
 
 func parseObjectAttributes(h http.Header) (attributes map[string]struct{}) {
 	attributes = make(map[string]struct{})
-	for _, v := range strings.Split(strings.TrimSpace(h.Get(xhttp.AmzObjectAttributes)), ",") {
-		if v != "" {
-			attributes[v] = struct{}{}
+	for _, headerVal := range h.Values(xhttp.AmzObjectAttributes) {
+		for _, v := range strings.Split(strings.TrimSpace(headerVal), ",") {
```
ramondeklein, line 229: I wasn't correct in my initial comment, where I thought that `http.Header.Values()` would already split the string. It doesn't... It just goes through all the headers with that name. I created a small Go program:
```go
package main

import (
	"log"
	"net/http"
)

func main() {
	http.ListenAndServe("0.0.0.0:9091", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		vv := r.Header.Values("X-TEST")
		log.Printf("%+v", vv)
		w.WriteHeader(http.StatusOK)
	}))
}
```
When invoked with `curl -H "X-TEST: 1" -H "X-TEST: 2" -H "X-TEST: 3,4" http://localhost:9091`, it prints the following headers: `[1 2 3,4]`. So the PR code is correct and it seems that it's necessary to split the values manually.

ramondeklein (changes requested):

cmd/object-api-options.go
ramondeklein: If someone is used spaces here, then it specified an invalid value and we may want to throw an error instead. `400` (bad request) would be appropriate.

cmd/object-api-options.go
klauspost: So `CanonicalHeaders` should be the comma separated, trimmed values I presume.

Can we check we do that?

klauspost:
Coincidence?  Not really sure what actually failed here... (upload or delete)

```
(9/15) Running minio-go tests ... FAILED in 34 seconds
{
  "time": "2024-08-21T07:07:46.842149747Z",
  "level": "ERROR",
  "name": "minio-go: testRemoveObjects",
  "duration": 1337,
  "function": "RemoveObjects(bucketName, objectsCh, opts)",
  "args": {
    "bucketName": "minio-go-test-1euhkbecfxd2josa",
    "objectName": "ipzwzw1b1jvdzhtqwvb1x09xueorr0",
    "objectPrefix": "",
    "recursive": "true"
  },
  "status": "FAIL",
  "message": "Error uploading object",
  "error": "The request signature we calculated does not match the signature you provided. Check your key and signing method."
}
```

cmd/object-api-options.go
ramondeklein: `CanonicalHeaders` has the following format:
```
Lowercase(<HeaderName1>)+":"+Trim(<value>)+"\n"
Lowercase(<HeaderName2>)+":"+Trim(<value>)+"\n"
...
Lowercase(<HeaderNameN>)+":"+Trim(<value>)+"\n"
```
Most important is that headers are in lower-case and sorted. If the header contains multiple values (separated by spaces), then we should just use its trimmed value. If `<value>` contains multiple values, then the entire value must be used (trimmed). This the canonical request that AWS SDK generates for `aws s3api --debug get-object-attributes --bucket test --key xxx --object-attributes "ETag" "Checksum"`:
```
GET
/test/xxx
attributes=
host:localhost:9000
x-amz-content-sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
x-amz-date:20240821T114519Z
x-amz-object-attributes:ETag,Checksum

host;x-amz-content-sha256;x-amz-date;x-amz-object-attributes
e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
2024-08-21 13:45:19,480 - MainThread - botocore.auth - DEBUG - StringToSign:
AWS4-HMAC-SHA256
20240821T114519Z
20240821/us-east-1/s3/aws4_request
26897135b1bd8e4d0bca2c3115e586fef66676802fa57162ac5ce065cc1cce43
```
It set the `x-amz-object-attributes` to `ETag,Checksum`. I don't think we should encounter any issues here. The signing method doesn't even take the multiple values into consideration. It's just a single value for the signing method. It may be different, when you specify the same header multiple times. But that's not allowed according to the HTTP spec, so I think it's fine if that doesn't work. No client should do that...

ramondeklein:
> Coincidence? Not really sure what actually failed here... (upload or delete)

I don't think it can be related. Parsing the incoming header for the `s3:GetObjectAttributes` request has been changed. I don't think it could cause signature failures here. If it reaches the handler, then signature verification has already been done.

cmd/object-api-options.go
ramondeklein: @klauspost I checked the [source](https://github.com/minio/minio/blob/7b239ae1544cd8c8b4cd510a7cf6bbfd53f37bfb/cmd/signature-v4.go#L61-L84) and that seems to work fine. We may want to add the following test to `signature-v4_test.go`:
```go
func TestCanonicalHeaders(t *testing.T) {
	h := http.Header{
		"X-Test": []string{"value1", "value2", "value3.1,value3.2"},
	}
	got := getCanonicalHeaders(h)
	expected := "x-test:value1,value2,value3.1,value3.2\n"
	if got != expected {
		t.Errorf("Expected %s, but got %s", expected, got)
	}
}
```

ramondeklein: I checked and if you specify the following header: `http.Header{"X-TEST": []string{"1", "2", "3.1,3.2"}}`, then the default Go HTTP client will generate three `X-TEST` headers. Each with a single value.

ramondeklein (approved):

harshavardhana (approved):

cmd/object-api-options.go
marktheunissen: yeah, I thought about that but decided probably not worth it.

PR #20301, opened 2024-08-21
s3: DeleteBucket to use listing before returning bucket not empty error

vadmeste:

## Description
Use Walk(), which is a recursive listing with versioning, to check if a 
bucket has some objects before allowing to remove it. This is beneficial 
because the bucket can contain multiple dangling objects in multiple drives, 
in a way that DeleteBucket returns bucket not empty error though listing 
does not list any object.

Also, this will prevent a bug where a bucket is deleted in a deployment that
has many erasure sets but the bucket contains one or few objects not spread
in enough erasures sets.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/erasure-server-pool.go
+++ b/cmd/erasure-server-pool.go
@@ -1980,6 +1980,34 @@ func (z *erasureServerPools) DeleteBucket(ctx context.Context, bucket string, op
 		defer lk.Unlock(lkctx)
 	}
 
+	if !opts.Force {
+		results := make(chan itemOrErr[ObjectInfo])
+
+		ctx, cancel := context.WithTimeout(ctx, time.Minute)
+		defer cancel()
+		err := z.Walk(ctx, bucket, "", results, WalkOptions{Limit: 1})
+		if err != nil {
+			s3LogIf(ctx, fmt.Errorf("unable to verify if the bucket %s is empty: %w", bucket, err))
+			return toObjectErr(err, bucket)
+		}
+
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case r, found := <-results:
+			if found {
+				if r.Err != nil {
+					s3LogIf(ctx, fmt.Errorf("unable to verify if the bucket %s is empty: %w", bucket, r.Err))
+					return toObjectErr(r.Err, bucket)
+				}
+				return toObjectErr(errVolumeNotEmpty, bucket)
+			}
+		}
+
+		// Always pass force to the lower level
+		opts.Force = true
+	}
+
 	err := z.s3Peer.DeleteBucket(ctx, bucket, opts)
 	if err == nil || isErrBucketNotFound(err) {
 		// If site replication is configured, hold on to deleted bucket state until sites sync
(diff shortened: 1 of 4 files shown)


cmd/erasure-server-pool.go
```diff
@@ -1980,6 +1980,24 @@ func (z *erasureServerPools) DeleteBucket(ctx context.Context, bucket string, op
 		defer lk.Unlock(lkctx)
 	}
 
+	if !opts.Force {
+		results := make(chan itemOrErr[ObjectInfo])
+		go z.Walk(ctx, bucket, "", results, WalkOptions{Limit: 1})
```
```suggestion
		go z.Walk(ctx, bucket, "", results, WalkOptions{Limit: 1})
```

pass a context which can canceled upon return, don't wait on client context to cancel this goroutine.

cmd/erasure-server-pool.go
```diff
@@ -1980,6 +1980,24 @@ func (z *erasureServerPools) DeleteBucket(ctx context.Context, bucket string, op
...
+
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case r, found := <-results:
+			if found {
+				if r.Err != nil {
+					s3LogIf(ctx, fmt.Errorf("unable to list the bucket %s: %w", bucket, r.Err))
```
```suggestion
					s3LogIf(ctx, fmt.Errorf("unable to verify if bucket %s is empty: %w", bucket, r.Err))
```

harshavardhana (changes requested):

cmd/erasure-server-pool.go
```diff
@@ -1980,6 +1980,34 @@ func (z *erasureServerPools) DeleteBucket(ctx context.Context, bucket string, op
...
+			return ctx.Err()
+		case r, found := <-results:
+			if found {
+				if r.Err != nil {
+					s3LogIf(ctx, fmt.Errorf("unable to verify if the bucket %s is empty: %w", bucket, r.Err))
+					return toObjectErr(r.Err, bucket)
+				}
+				return toObjectErr(errVolumeNotEmpty, bucket)
+			}
+		}
```
harshavardhana, line 2005: this is a bit unsafe logic, instead of relying on channel `ok` we should rely on valid `r.Err == nil` 

Do it like this

```
case r := <-results:
    if r.Err == nil { 
          return toObjectErr(errVolumeNotEmpty, bucket)
    }
    s3LogIf(ctx, fmt.Errorf("unable to verify if the bucket %s is empty: %w", bucket, r.Err))
    return toObjectErr(r.Err, bucket)
```

harshavardhana (changes requested):

cmd/erasure-server-pool.go
klauspost: If results are closed with no results, r will be the zero value with `r.Err = nil`. AFAICT this is the correct check.

vadmeste: yes agree.. (by the way @harshavardhana you may have reviewed an older buggy code).

harshavardhana (approved):

krisis (approved):

PR #20324, opened 2024-08-25
improve multipart decryption

aead:
## Description
This commit simplifies and optimizes the decryption of large (multipart) objects. This commit does two things:
 - Re-write the init logic for the decryption reader
 - Reduce the number of OEK decryptions

Before, the init logic copied some SSE HTTP request headers to parse them later on. This is simplified to parsing them right away. This removes some fields from the decryption reader struct.

Further, the decryption reader used to decrypt the OEK using the client provided key (SSE-C) or the KMS (SSE-S3 / SSE-KMS) for each part. This is redunant since the OEK is the same for all parts. In particular, a KMS call might be a network request. Now, the OEK is decrypted once when init. the reader. This should improve latency when reading encrypted multipart objects and reduce requests to the KMS.

## Motivation and Context
KMS, SSE

## Types of changes

## Checklist:

### DIFF
--- a/cmd/encryption-v1.go
+++ b/cmd/encryption-v1.go
@@ -601,28 +601,44 @@ func DecryptBlocksRequestR(inputReader io.Reader, h http.Header, seqNumber uint3
 	partEncRelOffset := int64(seqNumber) * (SSEDAREPackageBlockSize + SSEDAREPackageMetaSize)
 
 	w := &DecryptBlocksReader{
-		reader:            inputReader,
-		startSeqNum:       seqNumber,
-		partDecRelOffset:  partDecRelOffset,
-		partEncRelOffset:  partEncRelOffset,
-		parts:             oi.Parts,
-		partIndex:         partStart,
-		header:            h,
-		bucket:            bucket,
-		object:            object,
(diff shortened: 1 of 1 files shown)


harshavardhana:
@aead I sent a PR here already https://github.com/minio/minio/pull/20317 PTAL.

harshavardhana:
> @aead I sent a PR here already #20317 PTAL.

I closed it, lets merge this PR.

harshavardhana (approved):

harshavardhana:
We also should obtain OEK once for both ETag and Data decryption - right now ETag also makes an additional call which we should improve as well. 

But this PR is good we can merge as is.

harshavardhana:
@aead don't submit PRs to origin please. Send a PR from your fork.

aead:
Oh sry - didnt see #20317 PTAL @harshavardhana

aead:
> We also should obtain OEK once for both ETag and Data decryption - right now ETag also makes an additional call which we should improve as well.

Yeah, I'll give the entire crypto logic a round of review / cleanup...

Issue #20321, opened 2024-08-25
Bug: If-Match and If-Unmodified-Since together in the HeadObjectRequest


T-TRz879:
**HeadObject** supports carrying If Match and If Unmodified Sense in requests. If these two parameters are used together, there may be issues in Minio
I upload a object, below is object info
```
{
  AcceptRanges: "bytes",
  ContentLength: 100,
  ContentType: "binary/octet-stream",
  ETag: "\"62ca1808bf6dd0fb37fe0f461f04939e\"",
  LastModified: 2024-08-24 02:34:39 +0000 UTC
}
```
below is my code
```go
func TestHeadObjectMinio(t *testing.T) {
	timeSince1 := time.Date(2024, 8, 24, 10, 34, 38, 0, time.Local)
	timeSince2 := time.Date(2024, 8, 24, 10, 34, 39, 0, time.Local)
	timeSince3 := time.Date(2024, 8, 24, 10, 34, 40, 0, time.Local)
	fmt.Println(timeSince1)
	fmt.Println(timeSince2)
	fmt.Println(timeSince3)
	if res, err := v1Cli.HeadObject(
		&s3.HeadObjectInput{
			Bucket:            aws.String(bucket),
			Key:               aws.String(key),
			IfMatch:           aws.String(`62ca1808bf6dd0fb37fe0f461f04939e`),
			IfUnmodifiedSince: aws.Time(timeSince1),
		}); err != nil {
		assert.Fail(t, err.Error())
	} else {
		fmt.Println(res)
	}
}
```

## Expected Behavior

minio should return 200 to me.

## Current Behavior

minio should return 412 to me.

## Possible Solution

I searched for the usage of this HTTP header in [RFC-9110](https://www.rfc-editor.org/rfc/rfc9110.html#field.if-match). 
<img width="780" alt="image" src="https://github.com/user-attachments/assets/5e9ac4dc-1e3c-4239-837e-0be416155ca3">

## Steps to Reproduce (for bugs)

1. PutObject to minio
2. HeadObject from minio with request If-Match and If-Unmodified-Since, set If-Match=object.ETag, set If-Unmodified-Since=the time before object.LastModified

## Context

If it's a real bug, I can deal with problem, Can you leave this issue to me to handle?

## Regression

No

## Your Environment

* Version used (`minio --version`): RELEASE.2023-11-15T20-43-25Z
* Server setup and configuration: 
[...]

PR #20326, opened 2024-08-26
ignore if-unmodified-since header if if-match is set

T-TRz879:
…ce in the HeadObject

## Description
Detailed description in https://github.com/minio/minio/issues/20321

## How to test this PR?
There are unit test in this PR.

## Types of changes

## Checklist:


go.mod
```diff
@@ -81,6 +81,7 @@ require (
 	github.com/rs/cors v1.11.0
 	github.com/secure-io/sio-go v0.3.1
 	github.com/shirou/gopsutil/v3 v3.24.5
+	github.com/stretchr/testify v1.9.0
```
harshavardhana: don't add testify this is a terrible dependency, just use Go's testing package.

harshavardhana (changes requested):

go.mod
T-TRz879: okay

cmd/object-handlers-common.go
```diff
@@ -268,6 +268,7 @@ func checkPreconditions(ctx context.Context, w http.ResponseWriter, r *http.Requ
 	ifNoneMatchETagHeader := r.Header.Get(xhttp.IfNoneMatch)
 	if ifNoneMatchETagHeader != "" {
 		if isETagEqual(objInfo.ETag, ifNoneMatchETagHeader) {
+			// Do not care If-Modified-Since, if true also return 304 (not modified), other false walk to follow check return 304 (not modified)
```
marktheunissen: This comment is not clear, could you please rewrite it?

cmd/object-handlers-common.go
```diff
@@ -319,11 +326,23 @@ func checkPreconditions(ctx context.Context, w http.ResponseWriter, r *http.Requ
...
+}
+
 // returns true if object was modified after givenTime.
 func ifModifiedSince(objTime time.Time, givenTime time.Time) bool {
-	// The Date-Modified header truncates sub-second precision, so
-	// use mtime < t+1s instead of mtime <= t to check for unmodified.
-	return objTime.After(givenTime.Add(1 * time.Second))
+	// return true if givenTime < objTime (object has been modified since givenTime)
+	// return false if givenTime >= objTime (object has not been modified since givenTime)
+	return givenTime.Before(objTime)
```
marktheunissen: The comments above this return are not necessary as they are just describing exactly what the code does, they don't add any useful information or explain any context. Better to just remove them.

cmd/object-handlers-common_test.go
```diff
@@ -51,3 +57,64 @@ func TestCanonicalizeETag(t *testing.T) {
 		}
 	}
 }
+
+// Tests - CheckPreconditions()
+func TestCheckPreconditions(t *testing.T) {
+	objModTime := time.Date(2024, 8, 26, 02, 01, 01, 0, time.UTC)
+	t.Run("If-None-Match", func(t *testing.T) {
```
marktheunissen: The usual pattern with t.Run is to call it within the range over testCases on line 78, rather than wrapping the entire block. That way each testCase is it's own sub-run and reports individual results.

cmd/object-handlers-common.go
```diff
@@ -319,11 +326,23 @@ func checkPreconditions(ctx context.Context, w http.ResponseWriter, r *http.Requ
 	return false
 }
 
+// checkIfUnmodifiedSince - return true if object was modified after(Including critical values) ifUnmodifiedSinceHeader time
```
marktheunissen: What is a "critical value"?

cmd/object-handlers-common_test.go
```diff
@@ -51,3 +57,64 @@ func TestCanonicalizeETag(t *testing.T) {
...
+			{ifNoneMatch: "aaa", ifModifiedSince: "Sun, 26 Aug 2024 02:01:02 GMT", objInfo: ObjectInfo{ETag: "aa", ModTime: objModTime}, expectedFlag: true, expectedCode: 304},
+		}
+		for i, tc := range testCases {
+			recorder := httptest.NewRecorder()
+			request := httptest.NewRequest(http.MethodHead, "/bucket/a", bytes.NewReader([]byte{}))
+			request.Header.Set(xhttp.IfNoneMatch, tc.ifNoneMatch)
+			request.Header.Set(xhttp.IfModifiedSince, tc.ifModifiedSince)
+			actualFlag := checkPreconditions(context.Background(), recorder, request, tc.objInfo, ObjectOptions{})
+			if tc.expectedFlag != actualFlag {
+				t.Errorf("Test Name:%s, test case:%d is fail", t.Name(), i+1)
```
marktheunissen: Rather than call t.Name(), give each testCase a `name string` and then use that, or just use the integer `i` here to identify which case has failed. Always provide the "got" and "want" to know why the case is failing.

```suggestion
				t.Errorf("test: %s, got flag: %s, want: %s", tc.name, actualFlag, tc.expectedFlag)
```

cmd/object-handlers-common.go
T-TRz879: > What is a "critical value"?

example: obj.Mod = '2024-08-26 02:01:01'

"critical value" => If-Modified-Since = '2024-08-26 02:01:01', in this case, headobject will return 304

cmd/object-handlers-common_test.go
T-TRz879: okay, thanks

cmd/object-handlers-common.go
T-TRz879: okay.

cmd/object-handlers-common_test.go
harshavardhana: actually even better use `t.Run("", func(t *testing.T) {})`

cmd/object-handlers-common.go
```diff
@@ -298,6 +306,12 @@ func checkPreconditions(ctx context.Context, w http.ResponseWriter, r *http.Requ
 			writeHeadersPrecondition(w, objInfo)
 			writeErrorResponse(ctx, w, errorCodes.ToAPIErr(ErrPreconditionFailed), r.URL)
 			return true
+		} else if !checkIfUnmodifiedSince(r, objInfo) {
+			// If both of the If-Match and If-Unmodified-Since headers are  present in the request as follows:
```
harshavardhana:
```suggestion
			// If both of the If-Match and If-Unmodified-Since headers are present in the request as follows:
```

cmd/object-handlers-common.go
```diff
@@ -268,6 +268,14 @@ func checkPreconditions(ctx context.Context, w http.ResponseWriter, r *http.Requ
 	ifNoneMatchETagHeader := r.Header.Get(xhttp.IfNoneMatch)
 	if ifNoneMatchETagHeader != "" {
 		if isETagEqual(objInfo.ETag, ifNoneMatchETagHeader) {
+			// Do not care If-Modified-Since, Because:
+			// 1. If If-Modified-Since condition evaluates to true.
+			//  If both of  the  If-None-Match  and  If-Modified-Since  headers  are present in the request as follows:
```
harshavardhana:
```suggestion
			//  If both of the If-None-Match and If-Modified-Since  headers are present in the request as follows:
```

cmd/object-handlers-common.go
```diff
@@ -319,6 +333,18 @@ func checkPreconditions(ctx context.Context, w http.ResponseWriter, r *http.Requ
 	return false
 }
 
+// checkIfUnmodifiedSince - return true if object was modified after(Including critical values) ifUnmodifiedSinceHeader time
```
harshavardhana: yeah but critical doesn't mean correct in this context.

harshavardhana (changes requested):

cmd/object-handlers-common.go
T-TRz879: Is this the incorrect reason？
“The Date-Modified header truncates sub-second precision, so use mtime < t+1s instead of mtime <= t to check for unmodified.”

cmd/object-handlers-common_test.go
```diff
@@ -51,3 +58,47 @@ func TestCanonicalizeETag(t *testing.T) {
...
+		// If-Match(true) and If-Unmodified-Since(false)
+		{name: "If-Match1", ifMatch: "aa", ifUnmodifiedSince: "Sun, 26 Aug 2024 02:01:00 GMT", objInfo: ObjectInfo{ETag: "aa", ModTime: objModTime}, expectedFlag: false, expectedCode: 200},
+		// If-Unmodified-Since(true)
+		{name: "If-Match2", ifMatch: "aa", ifUnmodifiedSince: "Sun, 26 Aug 2024 02:01:01 GMT", objInfo: ObjectInfo{ETag: "aa", ModTime: objModTime}, expectedFlag: false, expectedCode: 200},
+		{name: "If-Match3", ifMatch: "aa", ifUnmodifiedSince: "Sun, 26 Aug 2024 02:01:02 GMT", objInfo: ObjectInfo{ETag: "aa", ModTime: objModTime}, expectedFlag: false, expectedCode: 200},
+		// If-Match(true)
+		{name: "If-Match4", ifMatch: "aa", objInfo: ObjectInfo{ETag: "aa", ModTime: objModTime}, expectedFlag: false, expectedCode: 200},
+	}
+	for _, tc := range testCases {
+		recorder := httptest.NewRecorder()
```
marktheunissen: t.Run goes here now

T-TRz879: okay

cmd/object-handlers-common.go
harshavardhana: thats why we had 1s leeway

T-TRz879:
Is there anything else that needs to be modified?

cmd/object-handlers-common_test.go
```diff
@@ -51,3 +58,76 @@ func TestCanonicalizeETag(t *testing.T) {
...
+			actualFlag := checkPreconditions(context.Background(), recorder, request, tc.objInfo, ObjectOptions{})
+			if tc.expectedFlag != actualFlag {
+				t.Errorf("test: %s, got flag: %v, want: %v", tc.name, actualFlag, tc.expectedFlag)
+			}
+			if tc.expectedCode != recorder.Code {
+				t.Errorf("test: %s, got code: %d, want: %d", tc.name, recorder.Code, tc.expectedCode)
+			}
+		}
+	})
+	t.Run("If-Match", func(t *testing.T) {
```
marktheunissen: This is where t.Run was originally, and I reviewed asking for it to be moved. Then you removed it entirely, and I asked for it to be put back, except you've put it back in the original place instead of where I was asking for it to be moved to.

marktheunissen (changes requested):
Please follow existing testing conventions in the code. 

Really simple example of what needs to be done can be found in the file: `internal/amztime/iso8601_time_test.go`

Copy that style, don't create a new style.

T-TRz879:
It seems that the GitHub action error was not caused by my modifications, What do I need to do?

harshavardhana (approved):

cmd/object-handlers-common.go
```diff
@@ -298,6 +306,12 @@ func checkPreconditions(ctx context.Context, w http.ResponseWriter, r *http.Requ
 			writeHeadersPrecondition(w, objInfo)
 			writeErrorResponse(ctx, w, errorCodes.ToAPIErr(ErrPreconditionFailed), r.URL)
 			return true
+		} else if !checkIfUnmodifiedSince(r, objInfo) {
```
vadmeste: Isn't easier to avoid adding checkIfUnmodifiedSince function and just simply change the line 321 (little bit below), like this ?
```
--- if ifUnmodifiedSinceHeader != ""
+++  if if ifUnmodifiedSinceHeader != "" && ifMathETagHeader == "" {
```

T-TRz879: If do this
1. `If-Match` is true and `If-Unmodified-Since` is false, return 200.
2. `If-Match` is true and `If-Unmodified-Since` is true, return 200.
3. `If-Match` is true, return 200.
4. `If-Unmodified-Since` is true, return 200.
5. `If-Unmodified-Since` is false, return 412.

I think there's no problem with doing this, Thanks!

marktheunissen:
@T-TRz879 The changes I have asked for are still not done. Do you mind if I push a commit to this branch?

T-TRz879:
Sorry, Is there any other suggestion that I haven't made any changes to? There may be some omissions.

marktheunissen:
@T-TRz879 please see commit https://github.com/minio/minio/pull/20326/commits/d00a2fee4928be13715182ba03d4c6b41212de98 that I have just pushed up.

t.Run must be called in the loop

Ok +1 from me

marktheunissen (approved):

T-TRz879:
Sorry, I didn't get t Run must be called in the loop. I'll know this ut style next time.

harshavardhana (approved):

PR #20333, opened 2024-08-27
Try reconnect IAM systems if failed initially

### LINKED ISSUE #20118: feat: Reconnect OpenID configuration at startup, if fails
## Expected Behavior
Minio should fetch the OpenID configuration and (in case of an error) should occasionally refetch such configuration

## Current Behavior
When Minio in unable to initialize OpenID upon startup it just gives up and never makes an attempt to fetch the config at a later time.

## Possible Solution
Minio should attempt to fetch the OpenID config at a later time, e.g. every 5 minutes, until it gets a valid response.

1. Set up an OpenID provider, e.g. Keycloak, Authentik, or oidc-server-mock
2. Configure Minio to use the OpenID provider with environment variables (i.e. `MINIO_IDENTITY_OPENID_CONFIG_URL` et al)
3. Stop the OpenID provider (to simulate a temporary connectivity error)
4. Start Minio
5. Open the Minio console and confirm that OpenID login is not available
6. Restart the OpenID provider
7. Wait a bit
8. Refresh the Minio console
9. OpenID login is still not available

## Context
This has mostly occurred when both, Minio and our OpenID provider, were restarted at the same time due to updates of the underlying host, containerd, or rolling out several container updates at the same time. In this case Minio starts up before our OpenID provider and fails to obtain the OpenID config.

## Regression
Not that I know of. We have observed this multiple times now since setting up OpenID login.

## Your Environment
* Version used (`minio --version`): `minio version RELEASE.2024-06-13T22-53-53Z (commit-id=20960b6a2ddb9594ee418035b3c7c7fe92ae6a12)`
* Server setup and configuration: Docker compose
* Operating System and version (`uname -a`): `Linux aa2e5237-c6de-466c-aff0-a3923e0af7a2 6.1.0-21-cloud-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.90-1 (2024-05-03) x86_64 GNU/Linux`

shtripat:

## Description
If while start of MinIO server, some of IAM systems (e.g. OpenID/ldap) are not reachable, start a thread to check their availability runtime and re-initialize the configurations properly. This makes sure if openid/ldap comes online later, console would allow using the those creds for login.

## Motivation and Context
Fixes: https://github.com/minio/minio/issues/20118

## How to test this PR?
Follow instructions at https://github.com/minio/minio/issues/20118

## Types of changes

## Checklist:


cmd/globals.go
```diff
@@ -464,6 +464,9 @@ var (
 	globalDynamicAPIPort bool
 
 	// Add new variable global values here.
+
+	// We expect initially that all would go through
+	globalIAMFullyInitialized = true
```
klauspost: You sure you don't want an `atomic.Bool`?

shtripat: hmm, let me try that

cmd/server-main.go
```diff
@@ -972,11 +972,37 @@ func serverMain(ctx *cli.Context) {
 	}
 
 	// Initialize users credentials and policies in background right after config has initialized.
+	globalIAMFullyInitialized.Store(true)
```
klauspost: This is a "trap", since the value doesn't represent the state.

The goroutine reading it will not be started before the value is set anyway, so I don't think you need to set it to true before anyway, since the first call to `globalIAMSys.Init` will always set it. Correct?

cmd/server-main.go
```diff
@@ -972,11 +972,37 @@ func serverMain(ctx *cli.Context) {
...
 
+		// Start a thread to keep checking IAM initialized and re-trigger
+		go func() {
+			ticker := time.NewTicker(10 * time.Second) // Check every 10 seconds
+			defer ticker.Stop()
+			for {
+				select {
+				case <-ticker.C:
+					if !globalIAMFullyInitialized.Load() {
+						globalIAMSys.Init(GlobalContext, newObject, globalEtcdClient, globalRefreshIAMInterval)
```
klauspost: Without checking the details, it seems like `go sys.periodicRoutines(ctx, refreshInterval)` will be called each time you call Init.  Maybe it should only be called if successful?

cmd/server-main.go
```diff
@@ -972,11 +972,37 @@ func serverMain(ctx *cli.Context) {
...
+						if globalIAMFullyInitialized.Load() && globalBrowserEnabled {
+							bootstrapTrace("initConsoleServer", func() {
```
klauspost: Isn't this called twice? We also call it below - and it never waits for IAM, since everything is in a goroutine.

cmd/server-main.go
```diff
@@ -972,11 +972,37 @@ func serverMain(ctx *cli.Context) {
...
 
 	// Initialize users credentials and policies in background right after config has initialized.
+	globalIAMFullyInitialized.Store(true)
 	go func() {
 		bootstrapTrace("globalIAMSys.Init", func() {
 			globalIAMSys.Init(GlobalContext, newObject, globalEtcdClient, globalRefreshIAMInterval)
 		})
 
+		// Start a thread to keep checking IAM initialized and re-trigger
+		go func() {
```
klauspost: What is the point of this goroutine? We are already in a goroutine that is terminated as soon as this has been spawned.

klauspost (changes requested):
PTAL.

cmd/server-main.go
shtripat: hmm, makes sense

cmd/server-main.go
shtripat: yes, I will change accordingly and verify changes.

cmd/server-main.go
shtripat: I felt outer go routine would have exited by this time if all success and the inner one is running a ticker and first Init failed while start of server, it would keep waiting till it succeeds and then we set the console configuration afresh. Without this even if Init was successful console was not updated of openid availability and was not throwing option to login using openid.
Not sure if this understanding is wrong though.

cmd/server-main.go
shtripat: I thought these routines are disjoint in nature and this inner would keep running the ticker and wait for Init to succeed and then set console config fresh. Actually while testing it worked like that where the openid endpoint was down when I started MinIO and this ticker kept waiting. Once I started openid endpoint, it succeeded the Init and then console srarted allowing loin with openid.
Kindly guide if this understanding is wrong.

klauspost: But the goroutine will not be started until `bootstrapTrace("globalIAMSys.Init", func() {` has completed, so there is no need to do anything async here.

You are staring the goroutine inside the other goroutine... Maybe check if you are doing what you expect.

shtripat: ok, let me try without async

shtripat: I tried making this non async as below
```
            {
			ticker := time.NewTicker(10 * time.Second) // Check every 10 seconds
			defer ticker.Stop()
			for {
				select {
				case <-ticker.C:
					if !globalIAMFullyInitialized.Load() {
						globalIAMSys.Init(GlobalContext, newObject, globalEtcdClient, globalRefreshIAMInterval)
						if globalIAMFullyInitialized.Load() && globalBrowserEnabled {
							bootstrapTrace("initConsoleServer", func() {
								srv, err := initConsoleServer()
								if err != nil {
									logger.FatalIf(err, "Unable to initialize console service")
								}

								setConsoleSrv(srv)
							})
							return
						}
					}
				}
			}
		}
```
but in this case it blocks and console doesnt get initialized. Need is that even if openid not reachable keep trying to connect to it in background but at the same time go ahead and allow login using normal creds.

On top of recent updates on PR below change still keeps it working

```
diff --git a/cmd/iam.go b/cmd/iam.go
index 9d1c15a19..6ed42dd76 100644
--- a/cmd/iam.go
+++ b/cmd/iam.go
@@ -371,7 +371,9 @@ func (sys *IAMSys) Init(ctx context.Context, objAPI ObjectLayer, etcdClient *etc
 
        refreshInterval := sys.iamRefreshInterval
 
-       go sys.periodicRoutines(ctx, refreshInterval)
+       if localIAMInitialized {
+               go sys.periodicRoutines(ctx, refreshInterval)
+       }
 
        sys.printIAMRoles()
 
index de0ff1898..7be6eac87 100644
--- a/cmd/server-main.go
+++ b/cmd/server-main.go
@@ -972,7 +972,7 @@ func serverMain(ctx *cli.Context) {
        }
 
        // Initialize users credentials and policies in background right after config has initialized.
-       globalIAMFullyInitialized.Store(true)
+       //globalIAMFullyInitialized.Store(true)
        go func() {
                bootstrapTrace("globalIAMSys.Init", func() {
                        globalIAMSys.Init(GlobalContext, newObject, globalEtcdClient, globalRefreshIAMInterval)
```

klauspost: ok, it looked like the `// Initialize Console UI` was outside this goroutine.

Having the duplicated Console UI init seems weird.

A) It would already have been init'ed outside this goroutine? Why init it twice? Is that safe?
B) If we can initConsoleServer even if IAM hasn't loaded, why wait for it to try once?

harshavardhana:
I simplified the overall behavior @klauspost instead of having many goroutines, it is done in one self contained place PTAL.

harshavardhana:
> I simplified the overall behavior @klauspost instead of having many goroutines, it is done in one self contained place PTAL.

PTAL @shtripat

taran-p (commented):
IAMsys stays uninitialized if one of the subsystems is down, so actions by non-root users fail and IAM `mc` actions hang

harshavardhana:
> IAMsys stays uninitialized if one of the subsystems is down, so actions by non-root users fail and IAM `mc` actions hang

that is to be expected, you need to bring those endpoints online @taran-p

harshavardhana:
the idea is that those endpoints do come online lazily, and we can operate without them being online when the server is coming up.

shtripat:
I tried verifying the latest changes. With this change, when openid is down and we restart MinIO server, it keeps waiting for opendid to come online and keeps logging below error till openid comes online.
```
API: SYSTEM.iam
Time: 04:12:30 UTC 09/09/2024
DeploymentID: 35d9d4e7-0850-4994-ab92-423d2b5e15ba
Error: Unable to initialize OpenID: Get "http://localhost:5557/dex/.well-known/openid-configuration": dial tcp 127.0.0.1:5557: connect: connection refused (*fmt.wrapError)
       6: internal/logger/logger.go:268:logger.LogIf()
       5: cmd/logging.go:29:cmd.iamLogIf()
       4: cmd/iam.go:277:cmd.(*IAMSys).Init()
       3: cmd/server-main.go:980:cmd.serverMain.func14.1()
       2: cmd/server-main.go:561:cmd.bootstrapTrace()
       1: cmd/server-main.go:979:cmd.serverMain.func14()
INFO: Waiting for OpenID to be initialized.. (retrying in 1.280339766s)

API: SYSTEM.authZ
Time: 04:12:31 UTC 09/09/2024
DeploymentID: 35d9d4e7-0850-4994-ab92-423d2b5e15ba
Error: The policies "[readwrite]" mapped to role ARN arn:minio:iam:::role/domXb70kze7Ugc1SaxaeFchhLP4 are not defined - this role may not work as expected. (*errors.errorString)
       7: internal/logger/logger.go:268:logger.LogIf()
       6: cmd/logging.go:54:cmd.authZLogIf()
       5: cmd/iam.go:519:cmd.(*IAMSys).validateAndAddRolePolicyMappings()
       4: cmd/iam.go:366:cmd.(*IAMSys).Init()
       3: cmd/server-main.go:980:cmd.serverMain.func14.1()
       2: cmd/server-main.go:561:cmd.bootstrapTrace()
       1: cmd/server-main.go:979:cmd.serverMain.func14()
```

During this time console doesnt get initialized and we cannot open it. Once openid comes online everything starts working as expected.

If we dont restart MinIO while openid goes offline temporarily, the console falls back to userid/paswd based login and if we bring openid back online, it starts using the same for login.
Hope this is acceptable behavior and as
- while start of MinIO if configured openid down, it doesnt start well
- if openid eventually gets offline while MinIO is running the login sxreen falls back to userid/passwd based login and once openid is online again it starts using the same.

Approving the changes based on the same.

harshavardhana:
> * while start of MinIO if configured openid down, it doesnt start well
> * if openid eventually gets offline while MinIO is running the login sxreen falls back to userid/passwd based login and once openid is online again it starts using the same.

I mean it is precisely doing what we wanted to do, it isn't about not working well. SSO down is catastrophic, we cannot allow access to resources.

We could initialize the console earlier than IAM, which is okay. Then, it will work the same way as when SSO goes offline while the server is running.

harshavardhana:
> We could initialize the console earlier than IAM, which is okay. Then, it will work the same way as when SSO goes offline while the server is running.

@shtripat can you move the code around for starting Console UI in server-main.go

shtripat:
> > We could initialize the console earlier than IAM, which is okay. Then, it will work the same way as when SSO goes offline while the server is running.
> 
> @shtripat can you move the code around for starting Console UI in server-main.go

Juts by moving the console start before IAM init, would start the UI server but still would root creds work for login till we start openid as its already configured?
Also even after successful start and init of openid, we would need to re-init the console to make sure it starts using openid creds for login.

shtripat:
With new commit now behavior is as below
- in start when openid down, it uses root creds
- when we bring openid up in middle, after few secs it starts using openid
- if we bring down openid in middle, it falls back to root creds
- if we bring back openid again, after few secs it again starts using openid creds

shtripat:
> With new commit now behavior is as below
> 
> * in start when openid down, it uses root creds
> * when we bring openid up in middle, after few secs it starts using openid
> * if we bring down openid in middle, it falls back to root creds
> * if we bring back openid again, after few secs it again starts using openid creds

Reverted this change as not too safe way to handle

harshavardhana (approved):
Okay with this change but will confirm once with @abperiasamy

harshavardhana:
@shtripat please move this PR to EOS only

mircea-pavel-anton:
Hi, @harshavardhana !

I see this PR got closed, though I was very interested in this functionality being present, since currently if MinIO starts up before my IDP, then I simply lose all access to the console for my users unless I restart the k8s pod which in turn creates some downtime.

Are there any updates on this? What is the current status?

harshavardhana:
> Hi, @harshavardhana !
> 
> I see this PR got closed, though I was very interested in this functionality being present, since currently if MinIO starts up before my IDP, then I simply lose all access to the console for my users unless I restart the k8s pod which in turn creates some downtime.
> 
> Are there any updates on this? What is the current status?

PR will be made available in our EOS releases.

mircea-pavel-anton:
> > Hi, @harshavardhana !
> > 
> > I see this PR got closed, though I was very interested in this functionality being present, since currently if MinIO starts up before my IDP, then I simply lose all access to the console for my users unless I restart the k8s pod which in turn creates some downtime.
> > 
> > Are there any updates on this? What is the current status?
> 
> PR will be made available in our EOS releases.

So this will be an enterprise only feature?

harshavardhana:
> So this will be an enterprise only feature?

Correct.

mircea-pavel-anton:
> > So this will be an enterprise only feature?
> 
> Correct.

Well that's just a shame... The work was already done over here. I find it weird that SSO is NOT an enterprise only feature but rather it working properly is...

harshavardhana:
> Well that's just a shame... The work was already done over here. I find it weird that SSO is NOT an enterprise only feature but rather it working properly is...

@mircea-pavel-anton you are free to work on it, the code is available keep it in your fork.

donatello (approved):

PR #20350, opened 2024-08-30
Reduce redundant hash computation

arthur-zhang:

## Description

The hashKey function was initially called twice, once with an empty string and once with objects. By setting a default key variable and adjusting it based on the presence of objects, we ensure hashKey is called only once, thus reducing redundant computation.

## Types of changes

## Checklist:
- [x] Create a documentation update request [here](https://github.com/minio/docs/issues/new?label=doc-change,title=Doc+Updated+Needed+For+PR+github.com%2fminio%2fminio%2fpull%2fNNNNN)

### DIFF
--- a/cmd/erasure-server-pool.go
+++ b/cmd/erasure-server-pool.go
@@ -220,10 +220,11 @@ func newErasureServerPools(ctx context.Context, endpointServerPools EndpointServ
 }
 
 func (z *erasureServerPools) NewNSLock(bucket string, objects ...string) RWLocker {
-	poolID := hashKey(z.distributionAlgo, "", len(z.serverPools), z.deploymentID)
+	var key string
 	if len(objects) >= 1 {
-		poolID = hashKey(z.distributionAlgo, objects[0], len(z.serverPools), z.deploymentID)
+		key = objects[0]
 	}
+	poolID := hashKey(z.distributionAlgo, key, len(z.serverPools), z.deploymentID)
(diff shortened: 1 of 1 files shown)


klauspost:
@harshavardhana Looking at this, I don't see how this works reliably for multiple objects. Seems like it should group the objects by pool and do concurrent locks + waitgroup of these.

We only have [one use with multiple objects](https://github.com/minio/minio/blob/73992d2b9f586b90716bd264574f83b353db82a6/cmd/erasure-server-pool.go#L1203), but objects are being locked in the wrong pool, which should be fixed IMO. If you agree I can do a PR.

@arthur-zhang Sorry for hijacking your PR, but that seems like a bigger issue than the couple of extra cycles used for 2 hashes.

harshavardhana:
> We only have [one use with multiple objects](https://github.com/minio/minio/blob/73992d2b9f586b90716bd264574f83b353db82a6/cmd/erasure-server-pool.go#L1203), but objects are being locked in the wrong pool, which should be fixed IMO. If you agree I can do a PR.

yeah its an old change @klauspost we should fix it.

harshavardhana:
> @harshavardhana Looking at this, I don't see how this works reliably for multiple objects. Seems like it should group the objects by pool and do concurrent locks + waitgroup of these.

No it does, with multiple objects it holds the lock on the same pool that's all @klauspost - the actual locks are on objects, this code is just picking what is the expected poolID, we don't need anything more than this.

harshavardhana:
Sending a proper PR.

PR #20374, opened 2024-09-03
Dont start console service if MINIO_BROWSER=off

### LINKED ISSUE #20366: Get free port for console UI error even when MINIO_BROWSER=off
Hi!

I'm looking to reduce the number of ports required for Minio in a large concurrent test suite, and one of the ways I've been able to do this is to set `MINIO_BROWSER=off` to prevent the port (ports?) required to serve the UI. However from what I can tell [this](https://github.com/minio/minio/blob/master/cmd/common-main.go#L464) line of code runs regardless of the state of `MINIO_BROWSER`, and I see this error message in the logs:

```
FATAL Unable to get free port for Console UI on the host: Specified port is already in use
      > Please ensure no other program uses the same address/port
```

or startup failure:

`FATAL Unable to start the server: --console-address cannot be same as --address`

Is this expected, or is it the case that this port isn't required if the console UI isn't used?

## Expected Behavior
`MINIO_BROWSER=off` should prevent code trying to allocate that port

## Current Behavior
Regardless of the state of `MINIO_BROWSER` a port is selected for the console UI

This is easiest to trigger by reducing the ephermal port range on Linux.

1. run a command like: `sudo sysctl -w net.ipv4.ip_local_port_range="50080 50081"` or similar to restrict the range
2. `MINIO_BROWSER=off minio server $(mktemp -d) --address=":50080"`

## Context
Finding dynamic ports for Minio is a bit fiddly at the best of times, and this makes it worse - especially in a concurrent situation where we're already exposed to [this](https://github.com/minio/minio/issues/18301) race condition: 

## Your Environment
```
$ minio --version
minio version RELEASE.2024-07-15T19-02-30Z (commit-id=ded373e6008e1a746875c9c0d15e7e9911ff1549)
Runtime: go1.22.5 linux/amd64

Linux 5.15.153.1-microsoft-standard-WSL2 #1 SMP Fri Mar 29 23:14:13 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux
```

shtripat:

## Description
By default even if MINIO_BROWSER=off set code tries getting free port available for console and hold it. Not needed to avoid need for more ports.

## Motivation and Context
Fixes: https://github.com/minio/minio/issues/20366

## How to test this PR?
Follow https://github.com/minio/minio/issues/20366

## Types of changes

## Checklist:


klauspost (approved):

cmd/common-main.go
```diff
@@ -461,25 +461,31 @@ func handleCommonArgs(ctxt serverCtxt) {
...
-		if err != nil {
-			logger.FatalIf(err, "Unable to get free port for Console UI on the host")
-		}
-		// hold the port
-		l, err := net.Listen("TCP", fmt.Sprintf(":%s", p.String()))
-		if err == nil {
-			defer l.Close()
-		}
-		consoleAddr = net.JoinHostPort("", p.String())
+	browserEnabled, e := config.ParseBool(env.Get(config.EnvBrowser, config.EnableOn))
```
vadmeste: there is a similar code here  in cmd/server-main.go

``` 
        globalBrowserEnabled, err = config.ParseBool(env.Get(config.EnvBrowser, config.EnableOn))
        if err != nil {
                logger.Fatal(config.ErrInvalidBrowserValue(err), "Invalid MINIO_BROWSER value in environment variable")
        }   
```

I think it is better to move that code to new separate function called serverHandleEarlyEnvVars() and then call  serverHandleEarlyEnvVars()  before serverHandleCmdArgs() in server-main.go, then use globalBrowserEnabled flag in the rest of this commit.

shtripat: ok, let me do the change and update the PR. IIUC only the MINIO_BROWSER env var would be read and `globalBrowserEnabled` set in this new function.

vadmeste (approved):

klauspost (approved):

harshavardhana (approved):

PR #20402, opened 2024-09-08
Add support for audit/logger max retry and retry interval

harshavardhana:

## Description
Add support for audit/logger max retry and retry interval

## Motivation and Context
The current implementation retries forever until our
log buffer is full, and we start dropping events.

This PR allows you to set a value until we give
up on existing audit/logger batches to proceed to
process the new ones.

Bonus:
 - do not blow up buffers beyond batchSize value
 - do not leak ticker if worker returns

## How to test this PR?
The PR is self-explanatory.

## Types of changes

## Checklist:

### DIFF
--- a/internal/logger/target/http/http.go
+++ b/internal/logger/target/http/http.go
@@ -340,6 +340,9 @@ func (h *Target) startQueueProcessor(ctx context.Context, mainWorker bool) {
 	globalBuffer := logChBuffers[name]
 	logChLock.Unlock()
 
+	ticker := time.NewTicker(time.Second)
+	defer ticker.Stop()
+
 	var count int
 	for {
 		var (
@@ -348,7 +351,16 @@ func (h *Target) startQueueProcessor(ctx context.Context, mainWorker bool) {
 		)
 
 		if count < h.batchSize {
+			tickered := false
 			select {
+			case _ = <-ticker.C:
+				if len(h.logCh) > 0 || len(globalBuffer) > 0 {
+					// there is something in the log queue
+					// process it first, even if we tickered
+					// first.
+					continue
+				}
+				tickered = true
 			case entry, _ = <-globalBuffer:
 			case entry, ok = <-h.logCh:
 				if !ok {
@@ -358,31 +370,33 @@ func (h *Target) startQueueProcessor(ctx context.Context, mainWorker bool) {
 				return
 			}
 
-			h.totalMessages.Add(1)
-			if !isDirQueue {
-				if err := enc.Encode(&entry); err != nil {
-					h.config.LogOnceIf(
-						ctx,
-						fmt.Errorf("unable to encode webhook log entry, err  '%w' entry: %v\n", err, entry),
-						h.Name(),
-					)
-					h.failedMessages.Add(1)
+			if !tickered {
+				h.totalMessages.Add(1)
+				if !isDirQueue {
+					if err := enc.Encode(&entry); err != nil {
+						h.config.LogOnceIf(
+							ctx,
+							fmt.Errorf("unable to encode webhook log entry, err  '%w' entry: %v\n", err, entry),
+							h.Name(),
+						)
+						h.failedMessages.Add(1)
+						continue
+					}
+				} else {
+					entries = append(entries, entry)
+				}
+				count++
+
+				if len(h.logCh) > 0 || len(globalBuffer) > 0 {
 					continue
 				}
-			} else {
-				entries = append(entries, entry)
-			}
-			count++
 
-			if len(h.logCh) > 0 || len(globalBuffer) > 0 {
-				continue
-			}
-
-			// If we are doing batching, we should wait
-			// at least for a second, before sending.
-			// Even if there is nothing in the queue.
-			if h.batchSize > 1 && time.Since(lastBatchProcess) < time.Second {
-				continue
+				// If we are doing batching, we should wait
+				// at least for a second, before sending.
+				// Even if there is nothing in the queue.
+				if h.batchSize > 1 && time.Since(lastBatchProcess) < time.Second {
+					continue
+				}
 			}
 		} // if we have reached the count send at once or we have crossed last second before batch was processed.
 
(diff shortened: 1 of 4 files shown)


internal/logger/target/http/http.go
```diff
@@ -374,30 +368,31 @@ func (h *Target) startQueueProcessor(ctx context.Context, mainWorker bool) {
...
-				count++
 			}
+			count++
 		}
 
-		if count != h.batchSize {
-			if len(h.logCh) > 0 || len(globalBuffer) > 0 || count == 0 {
-				continue
-			}
+		if len(h.logCh) > 0 || len(globalBuffer) > 0 || count == 0 {
```
donatello: Suppose `count == h.batchSize` before the if block at line 359 (so the if block is not entered), then if line 377 evaluates to true, looks like current `entry` is lost due to being overwritten? Is this expected?

donatello (changes requested):

internal/logger/target/http/http.go
poornas: good point, more than one entry can be lost if less than one second elapsed since last batch

internal/logger/target/http/http.go
```diff
@@ -422,18 +417,25 @@ func (h *Target) startQueueProcessor(ctx context.Context, mainWorker bool) {
...
+				retryIntvl = 3 * time.Second
 			}
 
-			time.Sleep(3 * time.Second)
-			goto retry
+			time.Sleep(retryIntvl)
+			retries++
+			if retries <= maxRetries {
+				goto retry
+			}
```
poornas:
```suggestion
			if retries <= maxRetries  || maxRetries == 0 {
				goto retry
			}
```
maxRetries 0 - shouldn't it be retrying indefinitely

poornas (changes requested):

internal/logger/target/http/http.go
kannappanr: so, it should simply be `if count <= h.batchSize {`

internal/logger/target/http/http.go
harshavardhana: maxRetries == 0 means undefined

internal/logger/target/http/http.go
```diff
@@ -374,30 +368,30 @@ func (h *Target) startQueueProcessor(ctx context.Context, mainWorker bool) {
...
-				// If we are doing batching, we should wait
-				// at least one second before sending.
-				// Even if there is nothing in the queue.
-				if time.Since(lastBatchProcess).Seconds() < 1 {
-					continue
-				}
+			// If we are doing batching, we should wait
+			// at least for 2 seconds before sending.
+			// Even if there is nothing in the queue.
+			if h.batchSize > 1 && time.Since(lastBatchProcess).Seconds() < 2 {
```
vadmeste: we probably need to keep the ticker here, because we may wait for a long time until next event comes in

internal/logger/target/http/http.go
```diff
@@ -361,9 +356,8 @@ func (h *Target) startQueueProcessor(ctx context.Context, mainWorker bool) {
 			return
 		}
 
-		if !isTick {
+		if count < h.batchSize {
```
vadmeste, line 353: I think we can miss some events here.. we are receiving an entry from h.logCh and globalBuffer then we have this condition that will ignore adding this entry to the buffer or entries array

internal/logger/target/http/http.go
harshavardhana: no this only waits since the lastBatchProcess - we sent at once whatever the count is.

vadmeste (approved):

internal/logger/target/http/http.go
```diff
@@ -227,6 +229,7 @@ func (h *Target) send(ctx context.Context, payload []byte, payloadCount int, pay
 			if xnet.IsNetworkOrHostDown(err, false) {
 				h.status.Store(statusOffline)
 			}
+			h.failedMessages.Add(int64(payloadCount))
```
zveinn, line 232: we don't want to register failedMessages on a 401 ? because it'll trigger the same outcome as an error. I can see a scenario where no failed messages are being reported while not receiving an error and the customer won't notice the stdout log. They will instantly open a ticket :P

internal/logger/target/http/http.go
```diff
@@ -343,27 +340,25 @@ func (h *Target) startQueueProcessor(ctx context.Context, mainWorker bool) {
...
-		case entry, _ = <-globalBuffer:
-		case entry, ok = <-h.logCh:
-			if !ok {
+		var (
+			ok    bool
+			entry any
+		)
+
+		if count < h.batchSize {
+			select {
```
zveinn, line 355: It's probably not a good idea to remove the ticker. If someone configures a batch size of lets say 100 but their system is having a slow day, then those audit logs will sit in the queue until they reach 100.  It creates a bigger time window where audit logs could be lost due to hard restarts and we'll probably receive tickets about log delivery being slow from some clients.

internal/logger/target/http/http.go
```diff
@@ -415,25 +409,36 @@ func (h *Target) startQueueProcessor(ctx context.Context, mainWorker bool) {
...
 				context.Background(),
-				fmt.Errorf("unable to send webhook log entry(s) to '%s' err '%w': %d", name, err, count),
+				fmt.Errorf("unable to send audit/log entry(s) to '%s' err '%w': %d", name, err, count),
 				name,
 			)
 
-			if errors.Is(err, context.Canceled) {
-				return
+			if retryIntvl <= 0 {
+				retryIntvl = 3 * time.Second
```
zveinn: This can be initialized outside the loop so we don't have to 'if' for every loop

internal/logger/target/http/http.go
vadmeste, line 355: it looks like a ticker is still needed here, because there is a possiblity that one event is kept in the queue and not sent for a long time when there is no new entries in globalBuffer and h.logCh

vadmeste (changes requested):

internal/logger/target/http/http.go
harshavardhana: why not?

harshavardhana: this code is no different than what we already have, `if err != nil` is true in all scenarios that we have it in master right now. It just adds it in defer statement that's all.

vadmeste (approved):

vadmeste (approved):

Praveenrajmani (approved):

PR #20409, opened 2024-09-09
Don't send a canceled context to Unlock

klauspost:
## Description

AFAICT we send a canceled context to unlock (and thereby releaseAll). This will cause network calls to fail.

Instead use background and add 30s timeout.

@vadmeste You know more about this than I. It seems strange this has been undetected, so I tend to not believe it is the issue. I would expect it to be a bigger issue if this is the case, so I am probably wrong.

PTAL.

## Types of changes

### DIFF
--- a/cmd/namespace-lock.go
+++ b/cmd/namespace-lock.go
@@ -185,7 +185,7 @@ func (di *distLockInstance) Unlock(lc LockContext) {
 	if lc.cancel != nil {
 		lc.cancel()
 	}
-	di.rwMutex.Unlock(lc.ctx)
+	di.rwMutex.Unlock(context.Background())
 }
 
 // RLock - block until read lock is taken or timeout has occurred.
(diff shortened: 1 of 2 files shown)


vadmeste (approved):

harshavardhana (approved):

vadmeste:
@klauspost I think this did not cause a big issue before because the locking subsystem has a refresh mechanism. That mechanism will clear a lock when its owner does not claim its ownership to that lock anymore (just lock uuid not found in its memory). Since the node will call Lock() then Unlock() it, the unlock will be always successful in that node since it is just a memory operation (deleting the lock uuid from the memory)

So to summarize, the locking refresh mechanism will clear this automatically after 30 seconds.

harshavardhana:
> @klauspost I think this did not cause a big issue before because the locking subsystem has a refresh mechanism. That mechanism will clear a lock when its owner does not claim its ownership to that lock anymore (just lock uuid not found in its memory). Since the node will call Lock() then Unlock() it, the unlock will be always successful in that node since it is just a memory operation (deleting the lock uuid from the memory)

I'm even there, but it's holding it longer than it should. Other propagation delays may be adding to faster-failed refresh() unlocks().

I haven't been fortunate enough to reproduce it at the moment. It looks like you need some competing lockers as well as repeated retries, etc.

PR #20410, opened 2024-09-09
Reenable SVE support for Graviton 4

fwessels:
## Description

Update dependency for `reedsolomon` to re-enable ARM SVE support for Graviton 4.


marktheunissen (approved):
Code looks good but I defer to Klaus

klauspost (approved):

Issue #20319, opened 2024-08-25
`Content-MD5` field/header seems to have no effect with a presigned POST url


MrCreosote:
When POSTing with a presigned URL to Minio, I am unable to cause the upload to fail via specifying a known bad `Content-MD5` field and header.

## Expected Behavior
When uploading a file to Minio with a presigned URL and specifying a known bad `Content-MD5`, the upload should fail.

## Current Behavior
The upload succeeds.

## Possible Solution
Possibly I'm doing something wrong - if so, if you could point it out that'd be great, as I'm a bit stumped.
Otherwise presumably Minio needs to check the MD5 of the file against the header / field.

## Steps to Reproduce (for bugs)
* Start Minio, with or without the `--compat` flag
```
MINIO_ROOT_USER=admin MINIO_ROOT_PASSWORD=password ~/minio/RELEASE/2024-08-17T01-24-54Z/minio server ~/minio/data2 --compat
```
* Use the method of your choice to create a bucket called `test`
* Execute the following Python code, which follows [these directions](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/s3-presigned-urls.html#generating-a-presigned-url-to-upload-a-file):
```python
$ ipython
Python 3.11.3 (main, Apr  5 2023, 14:14:40) [GCC 7.5.0]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.26.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from aiobotocore.session import get_session

In [2]: import requests

In [3]: sess = get_session()

In [4]: async with sess.create_client("s3", endpoint_url="http://localhost:9000"
   ...: , aws_access_key_id="admin", aws_secret_access_key="password") as client
   ...: :
   ...:     url = await client.generate_presigned_post(Bucket="test", Key="somef
   ...: ile", Fields={"Content-MD5": "garbage"}, Conditions=[{"Content-MD5": "ga
   ...: rbage"}])
   ...: 

In [5]: headers = {"Content-MD5": "garbage"}

In [6]: url
Out[6]: 
{'url': 'http://localhost:9000/test',
 'fields': {'Content-MD5': 'garbage',
  'key': 'somefile',
  'AWSAccessKeyId': 'admin',
[...]

PR #20417, opened 2024-09-11
convert PostPolicy API into a multipart transaction

harshavardhana:

## Description
convert PostPolicy API into a multipart transaction

## Motivation and Context
recently, we encountered a situation where a client was uploading
over an unreliable connection could potentially hold locks on
the server while the body is yet to be read in a multi-pooled
setup, causing a significant amount of time holding active
locks on the server until we reach an EOF on the body.

Causing cascading occurrences of 499s, and 503s and triggering
various timeouts render server to manage these lengthy locks, 
although this does not incur any significant load, it is a sort 
of a DDOS behavior from an unreliable client.

Bonus: make PutObjectPart() per part, write lock granular, and
only to be held at the renamePart() instead of holding it before 
we read the client stream.

This PR uses multipart's transactional nature and uniqueness 
to avoid lock contention on the same object name() until we 
fully commit the data to the namespace.

Also ends up fixing https://github.com/minio/minio/issues/20319

## How to test this PR?
CI/CD should cover it, but here are the tests.

#### Plain-text test

```
#!/bin/bash

pkill minio
rm -rf /tmp/xl*

minio server /tmp/xl{1...4} &
mc ready myminio
mc mb myminio/testbucket -p

curl -v http://localhost:9000/testbucket/ -F policy=eyJleHBpcmF0aW9uIjoiMjAyNC0wOS0xN1QyMjowNDo1NC4zMDFaIiwiY29uZGl0aW9ucyI6W1siZXEiLCIkYnVja2V0IiwidGVzdGJ1Y2tldCJdLFsic3RhcnRzLXdpdGgiLCIka2V5IiwicHJlZml4LyJdLFsiZXEiLCIkeC1hbXotZGF0ZSIsIjIwMjQwOTEwVDIyMDQ1NFoiXSxbImVxIiwiJHgtYW16LWFsZ29yaXRobSIsIkFXUzQtSE1BQy1TSEEyNTYiXSxbImVxIiwiJHgtYW16LWNyZWRlbnRpYWwiLCJtaW5pb2FkbWluLzIwMjQwOTEwL3VzLWVhc3QtMS9zMy9hd3M0X3JlcXVlc3QiXV19 -F x-amz-algorithm=AWS4-HMAC-SHA256 -F x-amz-credential=minioadmin/20240910/us-east-1/s3/aws4_request -F x-amz-date=20240910T220454Z -F x-amz-signature=3c0be8533b665d94fd7a59f8ac455029f15b447a54aa4bd9d844c24ee926f05c -F bucket=testbucket -F key=prefix/hosts -F file=@/etc/hosts

mc ls myminio/testbucket/prefix/
mc cat myminio/testbucket/prefix/hosts | md5sum
mc stat myminio/testbucket/prefix/hosts
```

#### With SSE-KMS

```
#!/bin/bash

pkill minio
export MINIO_KMS_SECRET_KEY=my-minio-key:OSMM+vkKUTCvQs9YL/CVMIMt43HFhkUpqJxTmGl6rYw=

rm -rf /tmp/xl*

minio server /tmp/xl{1...4} &
mc ready myminio
mc mb myminio/testbucket -p

mc encrypt set sse-kms my-minio-key myminio/testbucket

mc mb myminio/testbucket -p
curl -v http://localhost:9000/testbucket/ -F policy=eyJleHBpcmF0aW9uIjoiMjAyNC0wOS0xN1QyMjowNDo1NC4zMDFaIiwiY29uZGl0aW9ucyI6W1siZXEiLCIkYnVja2V0IiwidGVzdGJ1Y2tldCJdLFsic3RhcnRzLXdpdGgiLCIka2V5IiwicHJlZml4LyJdLFsiZXEiLCIkeC1hbXotZGF0ZSIsIjIwMjQwOTEwVDIyMDQ1NFoiXSxbImVxIiwiJHgtYW16LWFsZ29yaXRobSIsIkFXUzQtSE1BQy1TSEEyNTYiXSxbImVxIiwiJHgtYW16LWNyZWRlbnRpYWwiLCJtaW5pb2FkbWluLzIwMjQwOTEwL3VzLWVhc3QtMS9zMy9hd3M0X3JlcXVlc3QiXV19 -F x-amz-algorithm=AWS4-HMAC-SHA256 -F x-amz-credential=minioadmin/20240910/us-east-1/s3/aws4_request -F x-amz-date=20240910T220454Z -F x-amz-signature=3c0be8533b665d94fd7a59f8ac455029f15b447a54aa4bd9d844c24ee926f05c -F bucket=testbucket -F key=prefix/hosts -F file=@/etc/hosts

mc ls myminio/testbucket/prefix/
mc cat myminio/testbucket/prefix/hosts | md5sum
mc stat myminio/testbucket/prefix/hosts
```

#### With SSE-S3

```
#!/bin/bash

pkill minio
export MINIO_KMS_SECRET_KEY=my-minio-key:OSMM+vkKUTCvQs9YL/CVMIMt43HFhkUpqJxTmGl6rYw=

rm -rf /tmp/xl*

minio server /tmp/xl{1...4} &
mc ready myminio
mc mb myminio/testbucket -p

mc encrypt set sse-s3 myminio/testbucket

curl -v http://localhost:9000/testbucket/ -F policy=eyJleHBpcmF0aW9uIjoiMjAyNC0wOS0xN1QyMjowNDo1NC4zMDFaIiwiY29uZGl0aW9ucyI6W1siZXEiLCIkYnVja2V0IiwidGVzdGJ1Y2tldCJdLFsic3RhcnRzLXdpdGgiLCIka2V5IiwicHJlZml4LyJdLFsiZXEiLCIkeC1hbXotZGF0ZSIsIjIwMjQwOTEwVDIyMDQ1NFoiXSxbImVxIiwiJHgtYW16LWFsZ29yaXRobSIsIkFXUzQtSE1BQy1TSEEyNTYiXSxbImVxIiwiJHgtYW16LWNyZWRlbnRpYWwiLCJtaW5pb2FkbWluLzIwMjQwOTEwL3VzLWVhc3QtMS9zMy9hd3M0X3JlcXVlc3QiXV19 -F x-amz-algorithm=AWS4-HMAC-SHA256 -F x-amz-credential=minioadmin/20240910/us-east-1/s3/aws4_request -F x-amz-date=20240910T220454Z -F x-amz-signature=3c0be8533b665d94fd7a59f8ac455029f15b447a54aa4bd9d844c24ee926f05c -F bucket=testbucket -F key=prefix/hosts -F file=@/etc/hosts

mc ls myminio/testbucket/prefix/
mc cat myminio/testbucket/prefix/hosts | md5sum
mc stat myminio/testbucket/prefix/hosts
```

## Types of changes

## Checklist:

### DIFF
--- a/cmd/erasure-multipart.go
+++ b/cmd/erasure-multipart.go
@@ -41,6 +41,7 @@ import (
 	"github.com/minio/minio/internal/logger"
 	"github.com/minio/pkg/v3/mimedb"
 	"github.com/minio/pkg/v3/sync/errgroup"
+	"github.com/minio/sio"
 )
 
 func (er erasureObjects) getUploadIDDir(bucket, object, uploadID string) string {
@@ -590,19 +591,6 @@ func (er erasureObjects) PutObjectPart(ctx context.Context, bucket, object, uplo
 		return pi, toObjectErr(err, bucket, object, uploadID)
 	}
 
-	// Write lock for this part ID, only hold it if we are planning to read from the
-	// streamto avoid any concurrent updates.
-	//
-	// Must be held throughout this call.
-	partIDLock := er.NewNSLock(bucket, pathJoin(object, uploadID, strconv.Itoa(partID)))
-	plkctx, err := partIDLock.GetLock(ctx, globalOperationTimeout)
-	if err != nil {
-		return PartInfo{}, err
-	}
-
-	ctx = plkctx.Context()
-	defer partIDLock.Unlock(plkctx)
-
 	onlineDisks := er.getDisks()
 	writeQuorum := fi.WriteQuorum(er.defaultWQuorum())
 
@@ -716,11 +704,26 @@ func (er erasureObjects) PutObjectPart(ctx context.Context, bucket, object, uplo
 		index = opts.IndexCB()
 	}
 
+	var actualSize int64
+	_, encrypted := crypto.IsEncrypted(fi.Metadata)
+	compressed := fi.IsCompressed()
+	switch {
+	case compressed:
+		actualSize = data.ActualSize()
+	case encrypted:
+		decSize, err := sio.DecryptedSize(uint64(n))
+		if err == nil {
+			actualSize = int64(decSize)
+		}
+	default:
+		actualSize = n
+	}
+
 	partInfo := ObjectPartInfo{
 		Number:     partID,
 		ETag:       md5hex,
 		Size:       n,
-		ActualSize: data.ActualSize(),
+		ActualSize: actualSize,
 		ModTime:    UTCNow(),
 		Index:      index,
 		Checksums:  r.ContentCRC(),
@@ -730,6 +733,19 @@ func (er erasureObjects) PutObjectPart(ctx context.Context, bucket, object, uplo
 	if err != nil {
 		return pi, toObjectErr(err, minioMetaMultipartBucket, partPath)
 	}
+
+	// Write lock for this part ID, only hold it if we are planning to read from the
+	// stream avoid any concurrent updates.
+	//
+	// Must be held throughout this call.
+	partIDLock := er.NewNSLock(bucket, pathJoin(object, uploadID, strconv.Itoa(partID)))
+	plkctx, err := partIDLock.GetLock(ctx, globalOperationTimeout)
+	if err != nil {
+		return PartInfo{}, err
+	}
+
+	ctx = plkctx.Context()
+	defer partIDLock.Unlock(plkctx)
 
 	onlineDisks, err = er.renamePart(ctx, onlineDisks, minioMetaTmpBucket, tmpPartPath, minioMetaMultipartBucket, partPath, partFI, writeQuorum)
 	if err != nil {
@@ -1154,7 +1170,7 @@ func (er erasureObjects) CompleteMultipartUpload(ctx context.Context, bucket str
 	// However, in case of encryption, the persisted part ETags don't match
 	// what we have sent to the client during PutObjectPart. The reason is
 	// that ETags are encrypted. Hence, the client will send a list of complete
-	// part ETags of which non can match the ETag of any part. For example
+	// part ETags of which may not match the ETag of any part. For example
 	//   ETag (client):          30902184f4e62dd8f98f0aaff810c626
 	//   ETag (server-internal): 20000f00ce5dc16e3f3b124f586ae1d88e9caa1c598415c2759bbb50e84a59f630902184f4e62dd8f98f0aaff810c626
 	//
--- a/cmd/bucket-handlers.go
+++ b/cmd/bucket-handlers.go
@@ -28,6 +28,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"mime"
 	"mime/multipart"
 	"net/http"
 	"net/textproto"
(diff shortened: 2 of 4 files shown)


cmd/erasure-multipart.go
```diff
@@ -716,11 +704,26 @@ func (er erasureObjects) PutObjectPart(ctx context.Context, bucket, object, uplo
...
+	compressed := fi.IsCompressed()
+	switch {
+	case compressed:
+		actualSize = data.ActualSize()
+	case encrypted:
+		decSize, err := sio.DecryptedSize(uint64(n))
+		if err == nil {
+			actualSize = int64(decSize)
+		}
+	default:
```
vadmeste, line 718: shouldn't we add `case compressed && encrypted:` ? 

it should be possible to calculate it,  I think we compress then we encrypt

harshavardhana: you can't @vadmeste this is borrowed from the idea in cmd/object-api-utils.go

klauspost:
@harshavardhana I have added a [Post Object benchmark](https://github.com/minio/warp/pull/338) to warp. I think considering the performance degradation is also a factor in this.

harshavardhana:
> @harshavardhana I have added a [Post Object benchmark](https://github.com/minio/warp/pull/338) to warp. I think considering the performance degradation is also a factor in this.

👍🏼

shtripat (approved):
lgtm, verified

klauspost (commented):
The checksum becoming a multipart checksum is breaking compatibility. It could be overridable, but you'd need to add that AFAICT.

Would it be crazy to have this configurable? Maybe as an opt-out `_MINIO` env var? If checksums are breaking as opt-in.

vadmeste:
```
API: SYSTEM.peers
Time: 14:49:14 UTC 09/11/2024
DeploymentID: 914cb1ef-9105-43cb-b097-f7f848a17f0e
Error: Invalid Handler for type: ClearUploadID (*grid.RemoteErr)
       peerAddress="127.0.0.1:9004"
       4: internal/logger/logonce.go:118:logger.(*logOnceType).logOnceIf()
       3: internal/logger/logonce.go:149:logger.LogOnceIf()
       2: cmd/logging.go:71:cmd.peersLogOnceIf()
       1: cmd/notification.go:140:cmd.(*NotificationGroup).Go.func1()
```

this is printed while testing

harshavardhana:
> The checksum becoming a multipart checksum is breaking compatibility. It could be overridable, but you'd need to add that AFAICT.
> 
> Would it be crazy to have this configurable? Maybe as an opt-out `_MINIO` env var? If checksums are breaking as opt-in.

I can change it to work with checksums the way they work with regular PUT. 

No no _MINIO stuff here we just a code that gives a temporary workaround for now.

klauspost:
@harshavardhana The checksums must be checked and appear as if PutObject - meaning not with `?????-1`. Both when returned, but also when requested afterwards with for example HeadObject.

harshavardhana:
> @harshavardhana The checksums must be checked and appear as if PutObject - meaning not with `?????-1`. Both when returned, but also when requested afterwards with for example HeadObject.

checksums are already checked @klauspost that's how tests pass - what you are asking is if they are stored and returned I will check that.

klauspost:
AFAICT you can set that to whatever you want. I don't think tests for PostObject are that extensive.

harshavardhana:
will bring in a separate change with following API 

- NewPutUpload() that provides a unique ID
- CommitUpload() that commits the unique ID

This way we can avoid concurrent locks.

PR #20434, opened 2024-09-13
hold granular locking for multi-pool PutObject()

harshavardhana:

## Description
hold granular locking for multi-pool PutObject()

## Motivation and Context
- PutObject() for multi-pooled was holding large region locks, 
  which was unnecessary. This affects almost all slowpoke 
  clients and lengthy uploads.

- Re-arrange locks for CompleteMultipart, PutObject to be 
  close to rename()

## How to test this PR?
CI/CD covers everything

## Types of changes

## Checklist:

### DIFF
--- a/cmd/erasure-server-pool.go
+++ b/cmd/erasure-server-pool.go
@@ -1091,18 +1091,7 @@ func (z *erasureServerPools) PutObject(ctx context.Context, bucket string, objec
 		return z.serverPools[0].PutObject(ctx, bucket, object, data, opts)
 	}
 
-	if !opts.NoLock {
-		ns := z.NewNSLock(bucket, object)
-		lkctx, err := ns.GetLock(ctx, globalOperationTimeout)
-		if err != nil {
-			return ObjectInfo{}, err
-		}
-		ctx = lkctx.Context()
-		defer ns.Unlock(lkctx)
-		opts.NoLock = true
-	}
-
-	idx, err := z.getPoolIdxNoLock(ctx, bucket, object, data.Size())
+	idx, err := z.getPoolIdx(ctx, bucket, object, data.Size())
 	if err != nil {
 		return ObjectInfo{}, err
 	}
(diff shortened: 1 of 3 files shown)


cmd/erasure-server-pool.go
```diff
@@ -1091,18 +1091,7 @@ func (z *erasureServerPools) PutObject(ctx context.Context, bucket string, objec
...
-		if err != nil {
-			return ObjectInfo{}, err
-		}
-		ctx = lkctx.Context()
-		defer ns.Unlock(lkctx)
-		opts.NoLock = true
-	}
-
-	idx, err := z.getPoolIdxNoLock(ctx, bucket, object, data.Size())
+	idx, err := z.getPoolIdx(ctx, bucket, object, data.Size())
```
vadmeste, line 1094: with this way, we could have the same object existing in different pools, but I think we support that

vadmeste (approved):
LGTM

the only benefit is to allow other s3 apps to perform other s3 calls on the same object that is currently uploading right ?

cmd/erasure-server-pool.go
harshavardhana: this is no different than NewMultipartUpload() behavior @vadmeste

harshavardhana: we only need READ locks to validate this, we do not need WRITE locks.

harshavardhana:
> LGTM
> 
> the only benefit is to allow other s3 apps to perform other s3 calls on the same object that is currently uploading right ?

We will hold other callers, including DELETEs and READs, once we have reached EOF and are about to update the content. Until then, the state() of the object will still have previous data.

But it's more about avoiding DDOS by a slow client that can hold an object hostage practically forever by sending 1 byte at a time; this is what happened at the customer 
site with duplicate uploads().

krisis (approved):

poornas (approved):

PR #20443, opened 2024-09-16
Changes to Access Key behavior when LDAP is configured then disabled

taran-p:

## Description
Add check in CheckKey that returns `!ok` for LDAP access key when LDAP is disabled

If LDAP config is deleted, `purgeLDAPAccessKeys` runs on server startup to delete any access keys for LDAP accounts

## Motivation and Context
Currently, access keys (both STS and service accounts) for LDAP accounts continue to function even if LDAP configuration has been disabled or removed. We would like to change this behavior different ways depending on scenario:

- If LDAP has been disabled but the configuration not removed, we would like access keys to not work. However, if the configuration is re-enabled, these access keys should begin to work again
- If LDAP has been deleted, all LDAP-associated access keys should be deleted

## How to test this PR?
Enable LDAP and create STS or service accounts for LDAP users. After disabling LDAP, these accounts should not work. Re-enable LDAP, and these accounts should function as normal. Remove LDAP configuration. After re-adding the config, access keys should no longer exist. 

Note that LDAP users in groups get errors if they try to log in when LDAP is enabled currently. To properly show change of PR, test with an LDAP user not in a group, like `ivanov` in `minio-iam-testing`

## Types of changes

## Checklist:


harshavardhana:
@taran-p please move this PR for eos only - it is large enough feature and doesn't need to be here.

PR #20451, opened 2024-09-18
Ensure proper `stale_uploads_cleanup_interval` is used at all times

### LINKED ISSUE #20450: MinIO ignores `stale_uploads_cleanup_interval` at start-up
The `stale_uploads_cleanup_interval` and `stale_uploads_expiry` settings are used to control the interval when abandoned multi-part uploads are being cleaned. During startup the `cmd.(*erasureSets).cleanupStaleUploads()` method is invoked that looks like this:
```go
func (s *erasureSets) cleanupStaleUploads(ctx context.Context) {
	timer := time.NewTimer(globalAPIConfig.getStaleUploadsCleanupInterval())
	defer timer.Stop()

	for {
		select {
		case <-ctx.Done():
			return
		case <-timer.C:
			// ...actual clean-up removed...

			// Reset for the next interval
			timer.Reset(globalAPIConfig.getStaleUploadsCleanupInterval())
		}
	}
}
```
The `globalAPIConfig.getStaleUploadsCleanupInterval()` method returns the interval, but it looks like `cleanupStaleUploads` is called, before the actual API configuration has been loaded. Another issue is that changing the interval doesn't reset the loop, so it still sticks to the old setting until that expiry has passed.

ramondeklein:
Fixes #20450 by resetting the interval when the `stale_uploads_cleanup_interval` is set.

### DIFF
--- a/cmd/erasure-sets.go
+++ b/cmd/erasure-sets.go
@@ -86,6 +86,8 @@ type erasureSets struct {
 	lastConnectDisksOpTime time.Time
 }
 
+var staleUploadsCleanupIntervalChangedCh = make(chan bool)
+
 func (s *erasureSets) getDiskMap() map[Endpoint]StorageAPI {
 	diskMap := make(map[Endpoint]StorageAPI)
 
@@ -532,10 +534,11 @@ func (s *erasureSets) cleanupStaleUploads(ctx context.Context) {
 				}(set)
 			}
 			wg.Wait()
-
-			// Reset for the next interval
-			timer.Reset(globalAPIConfig.getStaleUploadsCleanupInterval())
+		case <-staleUploadsCleanupIntervalChangedCh:
 		}
+
+		// Reset for the next interval
+		timer.Reset(globalAPIConfig.getStaleUploadsCleanupInterval())
 	}
 }
 
--- a/cmd/erasure-sets.go
+++ b/cmd/erasure-sets.go
@@ -86,7 +86,7 @@ type erasureSets struct {
 	lastConnectDisksOpTime time.Time
 }
 
-var staleUploadsCleanupIntervalChangedCh = make(chan bool)
+var staleUploadsCleanupIntervalChangedCh = make(chan struct{})
 
 func (s *erasureSets) getDiskMap() map[Endpoint]StorageAPI {
 	diskMap := make(map[Endpoint]StorageAPI)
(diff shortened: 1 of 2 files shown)


ramondeklein:
Although this fixes the issue, I'm not a big fan of using a channel for this. If -for whatever reason-, the `cleanupStaleUploads` stops listening, then this will block `(*apiConfig).init`. If we already have some kind of eventing mechanism in MinIO, then that would be more appropriate.

cmd/erasure-sets.go
```diff
@@ -86,6 +86,8 @@ type erasureSets struct {
 	lastConnectDisksOpTime time.Time
 }
 
+var staleUploadsCleanupIntervalChangedCh = make(chan bool)
```
klauspost:
```suggestion
var staleUploadsCleanupIntervalChangedCh = make(chan struct{})
```

cmd/handler-api.go
```diff
@@ -183,13 +183,21 @@ func (t *apiConfig) init(cfg api.Config, setDriveCounts []int, legacy bool) {
...
 	t.rootAccess = cfg.RootAccess
 	t.syncEvents = cfg.SyncEvents
 	t.objectMaxVersions = cfg.ObjectMaxVersions
+
+	if t.staleUploadsCleanupInterval != cfg.StaleUploadsCleanupInterval {
+		t.staleUploadsCleanupInterval = cfg.StaleUploadsCleanupInterval
+
+		// signal that cleanup interval has changed
+		if staleUploadsCleanupIntervalChangedCh != nil {
+			staleUploadsCleanupIntervalChangedCh <- true
```
klauspost, line 198:
```suggestion
			select {
				case staleUploadsCleanupIntervalChangedCh <- struct{}:
				default:
			}
```

klauspost (approved):
A channel is by far the cleanest, when there is only one listener. No need for it to be bool, though.

cmd/handler-api.go
```diff
@@ -183,13 +183,21 @@ func (t *apiConfig) init(cfg api.Config, setDriveCounts []int, legacy bool) {
...
+		}
```
klauspost:
```suggestion
		select {
			case staleUploadsCleanupIntervalChangedCh <- struct{}:
			default:
		}
```
Always non-blocking. And doesn't send if no-one is listening - ie if cleanup is running.

klauspost (changes requested):

harshavardhana (approved):

klauspost (approved):

PR #20452, opened 2024-09-18
batch: Set a default retry attempts and a prefix 

vadmeste:

## Description
A batch job will fail if the retry attempt is not provided. The reason
is that the code mistakenly gets the retry attempts from the job status
rather than the job yaml file.

This will also set a default empty prefix in case of batch expiration.

Also this will avoid trimming the prefix since the yaml decoder already
does that if no quotes were provided, and we should not trim if quotes
were provided and the user provided a leading or a trailing space.

## Motivation and Context
Fix unsuccessful batch job with the retry attempts is not provided

## How to test this PR?
Test the following batch expiratoin yaml:
```
expire:
 apiVersion : v1
 bucket: bucket
 prefix: ""
 rules:
  - type: object
    name: ima*
```

## Types of changes

## Checklist:

### DIFF
--- a/cmd/batch-expire.go
+++ b/cmd/batch-expire.go
@@ -27,7 +27,6 @@ import (
 	"net/http"
 	"runtime"
 	"strconv"
-	"strings"
 	"time"
 
 	"github.com/minio/minio-go/v7/pkg/tags"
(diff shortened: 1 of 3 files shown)


klauspost (approved):
lgtm

krisis (approved):

PR #20458, opened 2024-09-19
Show minio version when building from source file

ruspaul013:
## Description
Showing minio version when building from source file.

## Motivation and Context
If minio is build form tar.gz file, when running `minio --version` will show
```
minio version DEVELOPMENT.GOGET (commit-id=DEVELOPMENT.GOGET)
Runtime: go1.23.0 linux/amd64
Copyright: 2015-0000 MinIO, Inc.
```
With this commit, minio will show now
```
minio version CUSTOM.2024-08-17T01-24-54Z (commit-id=DEVELOPMENT.GOGET)
Runtime: go1.23.0 linux/amd64
```

## How to test this PR?
Download the tar.gz file, unzip the file and then run `make`, after that run `./minio --version`.

## Types of changes

## Checklist:

### DIFF
--- a/buildscripts/gen-ldflags.go
+++ b/buildscripts/gen-ldflags.go
@@ -36,8 +36,10 @@ func genLDFlags(version string) string {
 	ldflagsStr += " -X github.com/minio/minio/cmd.Version=" + version
 	ldflagsStr += " -X github.com/minio/minio/cmd.CopyrightYear=" + copyrightYear
 	ldflagsStr += " -X github.com/minio/minio/cmd.ReleaseTag=" + releaseTag
-	ldflagsStr += " -X github.com/minio/minio/cmd.CommitID=" + commitID()
-	ldflagsStr += " -X github.com/minio/minio/cmd.ShortCommitID=" + commitID()[:12]
+	if existGit() {
+		ldflagsStr += " -X github.com/minio/minio/cmd.CommitID=" + commitID()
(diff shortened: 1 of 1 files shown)


klauspost (changes requested):
I know this is well intentioned, but we don't want non-official builds to show as releases.

ruspaul013:
@klauspost Thanks for the reply. 
Instead of RELEASE, can be something else like CUSTOM? I know that `MINIO_RELEASE` env will give that value.

klauspost:
`CUSTOM` would work for me.

harshavardhana:
This is beyond the scope of this project, please install via `go install -v github.com/minio/minio@latest` for source compilation. 

keep this change in your fork if needed.

ruspaul013:
Hello @harshavardhana! 
Thank you for the reply! I understand that this is beyond the scope of the project, but I was thinking that this would help other people that make custom builds for minio, so that they can use `MINIO_VERSION` to add the version and not to see `DEVELOPMENT`.

harshavardhana:
> Thank you for the reply! I understand that this is beyond the scope of the project, but I was thinking that this would help other people that make custom builds for minio, so that they can use `MINIO_VERSION` to add the version and not to see `DEVELOPMENT`.

If you are planning to ship your own binaries of minio you can pass in the flags directly via `go build --ldflags` this approach is not needed. We are never going to use MINIO_VERSION so its not needed in our repo.

PR #20487, opened 2024-09-26
use absolute path for binary checksum verification

harshavardhana:
## Community Contribution License
All community contributions in this pull request are licensed to the project maintainers
under the terms of the [Apache 2 license](https://www.apache.org/licenses/LICENSE-2.0). 
By creating this pull request I represent that I have the right to license the 
contributions to the project maintainers under the Apache 2 license.

## Description
use absolute path for binary checksum verification

## Motivation and Context
minio might be run relative to its ${PATH} in certain 
environments

## How to test this PR?
Run a distributed setup inside container environments

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Optimization (provides speedup with no functional changes)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [x] Fixes a regression since #19968
- [ ] Unit tests added/updated
- [ ] Internal documentation updated
- [ ] Create a documentation update request [here](https://github.com/minio/docs/issues/new?label=doc-change,title=Doc+Updated+Needed+For+PR+github.com%2fminio%2fminio%2fpull%2fNNNNN)

PR #20509, opened 2024-10-01
Fix checksum error during startup when `minio` is loaded via `PATH` environment variable

ramondeklein:

## Description
Fix `ERRO: Calculating checksum failed: open minio: no such file or directory` during start-up when `minio` is started outside the directory where the executable is located, but via the `PATH` variable. This impacts our Docker image, because it uses `/` as the working directory and loads `minio` via the path variable.

PR #20487 already forces to use an absolute path, but `filepath.Abs` will use the current working directory instead of the actual file's directory. This PR will ensure it returns the full path of the executable that was used to start MinIO. There is no need to resolve symlinks, because `os.Open()` will take care of that and read the actual file, instead of the symlink (if any).

## Motivation and Context
MinIO calculates the hash of the binary during start-up. It uses these hashes to prevent having mixed versions inside a single cluster. When the signature cannot be calculated it logs and an error and sets the checksum to all zeros. When this bug is in multiple versions, then it may cause issues when different versions are used but this isn't detected properly.

## How to test this PR?
Run `minio` without specifying the full path and make sure that `minio` isn't in the current directory (so it should be loaded via the `PATH` variable). It should show the error message during start-up.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/api-errors.go
+++ b/cmd/api-errors.go
@@ -978,7 +978,7 @@ var errorCodes = errorCodeMap{
 	},
 	ErrReplicationNoExistingObjects: {
 		Code:           "XMinioReplicationNoExistingObjects",
-		Description:    "No matching ExistingsObjects rule enabled",
+		Description:    "No matching ExistingObjects rule enabled",
 		HTTPStatusCode: http.StatusBadRequest,
 	},
 	ErrRemoteTargetDenyAddError: {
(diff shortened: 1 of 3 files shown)


ramondeklein:
We may want to create a release, because the start-up error in Docker feels awkward and we may mix versions in Kubernetes, because of this.

harshavardhana (approved):

KroArtem:
I stumbled across this problem when updating minio docker image in our codebase. Chances were low to do an update once in three years and get straight to non-working release but it happened :)

I see some folks also struggle with the issue https://forum.rockstor.com/t/minio-not-able-to-start-the-application/9798/8

So maybe it's indeed worth making a new release. JFYI

PR #20577, opened 2024-10-23
heal: large objects fix and avoid .healing.bin corner case premature exit

vadmeste:

## Description
heal: Do not abort updating .healing.bin during ongoing healing 
heal: Scan .minio.sys metadata only during site-wide heall
heal: Avoid deadline error with very large objects

## Types of changes

## Checklist:

### DIFF
--- a/cmd/xl-storage.go
+++ b/cmd/xl-storage.go
@@ -2275,6 +2275,7 @@ func (s *xlStorage) writeAllInternal(ctx context.Context, filePath string, b []b
 
 	_, err = w.Write(b)
 	if err != nil {
+		w.Truncate(0) // to indicate that we did partial write.
 		w.Close()
 		return err
 	}
--- a/cmd/xl-storage.go
+++ b/cmd/xl-storage.go
@@ -432,15 +432,19 @@ func (s *xlStorage) Healing() *healingTracker {
 		bucketMetaPrefix, healingTrackerFilename)
 	b, err := os.ReadFile(healingFile)
 	if err != nil {
+		if !errors.Is(err, os.ErrNotExist) {
+			internalLogIf(GlobalContext, fmt.Errorf("unable to read %s: %w", healingFile, err))
+		}
 		return nil
 	}
 	if len(b) == 0 {
+		internalLogIf(GlobalContext, fmt.Errorf("%s is empty", healingFile))
 		// 'healing.bin' might be truncated
 		return nil
 	}
 	h := newHealingTracker()
 	_, err = h.UnmarshalMsg(b)
-	bugLogIf(GlobalContext, err)
+	internalLogIf(GlobalContext, err)
 	return h
 }
 
@@ -2246,6 +2250,7 @@ func (s *xlStorage) writeAllMeta(ctx context.Context, volume string, path string
 	return renameAll(tmpFilePath, filePath, volumeDir)
 }
 
+// Create or truncate an existing file before writing
 func (s *xlStorage) writeAllInternal(ctx context.Context, filePath string, b []byte, sync bool, skipParent string) (err error) {
 	flags := os.O_CREATE | os.O_WRONLY | os.O_TRUNC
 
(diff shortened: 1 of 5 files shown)


cmd/xl-storage.go
```diff
@@ -2268,18 +2273,12 @@ func (s *xlStorage) writeAllInternal(ctx context.Context, filePath string, b []b
 		return err
 	}
 
-	n, err := w.Write(b)
+	_, err = w.Write(b)
```
klauspost, line 2276: Why remove the short write check? In what cases was it causing issues?

cmd/xl-storage.go
```diff
@@ -2367,6 +2366,41 @@ func (s *xlStorage) AppendFile(ctx context.Context, volume string, path string,
...
+						resp = checkPartVolumeNotFound
+					}
+					return
+				}
+			}
+		}
+		if osErrToFileErr(err) == errFileNotFound {
+			resp = checkPartFileNotFound
+		}
+		return
```
klauspost, line 2391: What about the paths to this return that don't set the response?

(I realize this is just code extracted, but it still seems strange)

cmd/xl-storage.go
```diff
@@ -2385,36 +2419,12 @@ func (s *xlStorage) CheckParts(ctx context.Context, volume string, path string,
 	}
 
 	for i, part := range fi.Parts {
-		partPath := pathJoin(path, fi.DataDir, fmt.Sprintf("part.%d", part.Number))
-		filePath := pathJoin(volumeDir, partPath)
-		st, err := Lstat(filePath)
+		resp.Results[i], err = xioutil.WithDeadline[int](ctx, globalDriveConfig.GetMaxTimeout(), func(ctx context.Context) (int, error) {
```
klauspost, line 2423: What is the default drive timeout, 1m?

klauspost (commented):
A few Qs

cmd/xl-storage.go
vadmeste: the short write check is never called, file.Write already returns an error in that case

cmd/xl-storage.go
vadmeste: By default, the result is unknown, which is zero. We do not act one unknown results

cmd/xl-storage.go
vadmeste: no, it is 30 seconds

cmd/xl-storage.go
klauspost: ok. But it never reads through the part - that is done in other checks, correct? 

If so, I am fine with the deadline and you can just resolve this.

cmd/xl-storage.go
klauspost: I like it as a secondary check, irrespective of the os.File implementation.

cmd/xl-storage.go
klauspost: ok. But it never reads through the part - that is done in other checks, correct? 

If so, I am fine with the deadline and you can just resolve this.

cmd/xl-storage.go
klauspost: ok.

cmd/xl-storage.go
vadmeste: >  But it never reads through the part 

yes, it never reads the part. This is the light healing scan where we do not do bitrot check

cmd/xl-storage.go
vadmeste: os.File.Write() method seems to be a generic call in all the operating systems. I addressed this differently, let me know what do you think

klauspost: Truncate in all error cases then.

I don't really see the point you are trying to make by removing a piece of defensive code from one of the critical write points in MinIO. I don't care if all current code paths make this impossible, but maybe some day it will show up in a stdlib change, and I'd like to handle that gracefully.

harshavardhana: This change must be reverted @vadmeste

vadmeste: it is truncating for all errors now.. yes, I wanted to remove the old code because it confuses debugging but it is better now with your suggestion.

vadmeste: @harshavardhana the code that I removed intially is never called, that is why I removed it in the first place. Now the change will introduce something new.. which is truncating a file if there is a write error, which may or may not work in practice (not sure)

harshavardhana: no we should never do that check what happens in the EOS branch @vadmeste

harshavardhana (approved):

PR #20631, opened 2024-11-11
Harden internode DeadlineConn

klauspost:
## Description

Since DeadlineConn would send deadline updates directly upstream, it would end up racing with Read/Write operations. The stdlib will perform a read, but do an async `SetReadDeadLine(unix(1))` to canceled the Read in `abortPendingRead`. In this case the Read may override the deadline intended to cancel the read.

Stop updating deadlines if a deadline in the past is seen and when Close is called. All upstream deadline calls are now protected by a mutex to avoid races. Should fix short time buildup of...

```
365 @ 0x44112e 0x4756b9 0x475699 0x483525 0x732286 0x737407 0x73816b 0x479601
#	0x475698	sync.runtime_notifyListWait+0x138		runtime/sema.go:569
#	0x483524	sync.(*Cond).Wait+0x84				sync/cond.go:70
#	0x732285	net/http.(*connReader).abortPendingRead+0xa5	net/http/server.go:729
#	0x737406	net/http.(*response).finishRequest+0x86		net/http/server.go:1676
#	0x73816a	net/http.(*conn).serve+0x62a			net/http/server.go:2050
```

AFAICT Only affects internode calls that create a connection (non-grid). These connections will typically die within a minute, but not nice to keep around.

## How to test this PR?

To repro the issue, it requires high traffic.

## Types of changes

### DIFF
--- a/internal/deadlineconn/deadlineconn.go
+++ b/internal/deadlineconn/deadlineconn.go
@@ -19,44 +19,70 @@
 package deadlineconn
 
 import (
+	"context"
 	"net"
+	"sync"
+	"sync/atomic"
 	"time"
 )
 
+// updateInterval is the minimum time between deadline updates.
 const updateInterval = 250 * time.Millisecond
 
 // DeadlineConn - is a generic stream-oriented network connection supporting buffered reader and read/write timeout.
 type DeadlineConn struct {
 	net.Conn
-	readDeadline  time.Duration // sets the read deadline on a connection.
-	readSetAt     time.Time
-	writeDeadline time.Duration // sets the write deadline on a connection.
-	writeSetAt    time.Time
+	readDeadline            time.Duration // sets the read deadline on a connection.
+	readSetAt               time.Time
+	writeDeadline           time.Duration // sets the write deadline on a connection.
+	writeSetAt              time.Time
+	abortReads, abortWrites atomic.Bool // A deadline was set to indicate caller wanted the conn to time out.
+	mu                      sync.Mutex
 }
 
 // Sets read deadline
 func (c *DeadlineConn) setReadDeadline() {
-	if c.readDeadline > 0 {
-		now := time.Now()
-		if now.Sub(c.readSetAt) > updateInterval {
-			c.Conn.SetReadDeadline(now.Add(c.readDeadline + updateInterval))
-			c.readSetAt = now
-		}
+	// Do not set a Read deadline, if upstream wants to cancel all reads.
+	if c.readDeadline <= 0 || c.abortReads.Load() {
+		return
+	}
+
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if c.abortReads.Load() {
+		return
+	}
+
+	now := time.Now()
+	if now.Sub(c.readSetAt) > updateInterval {
+		c.Conn.SetReadDeadline(now.Add(c.readDeadline + updateInterval))
+		c.readSetAt = now
 	}
 }
 
 func (c *DeadlineConn) setWriteDeadline() {
-	if c.writeDeadline > 0 {
-		now := time.Now()
-		if now.Sub(c.writeSetAt) > updateInterval {
-			c.Conn.SetWriteDeadline(now.Add(c.writeDeadline + updateInterval))
-			c.writeSetAt = now
-		}
+	// Do not set a Write deadline, if upstream wants to cancel all reads.
+	if c.writeDeadline <= 0 || c.abortWrites.Load() {
+		return
+	}
+
+	c.mu.Lock()
+	defer c.mu.Unlock()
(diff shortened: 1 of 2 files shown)


internal/deadlineconn/deadlineconn.go
```diff
@@ -19,56 +19,160 @@
...
+	// Do not set a Read deadline, if upstream wants to cancel all reads.
+	if c.readDeadline <= 0 || c.abortReads.Load() {
+		return
+	}
+
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if c.abortReads.Load() {
+		return
+	}
```
vadmeste, line 54: this seems to be already called in the beginning in the function, looks like it is not needed

internal/deadlineconn/deadlineconn.go
```diff
@@ -19,56 +19,160 @@
...
+	// Do not set a Write deadline, if upstream wants to cancel all reads.
+	if c.writeDeadline <= 0 || c.abortWrites.Load() {
+		return
+	}
+
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if c.abortWrites.Load() {
+		return
+	}
```
vadmeste, line 73: this seems to be already called in the beginning in the function, looks like it is not needed

internal/deadlineconn/deadlineconn.go
```diff
@@ -19,56 +19,160 @@
...
+		return 0, context.DeadlineExceeded
+	}
 	c.setWriteDeadline()
 	n, err = c.Conn.Write(b)
 	return n, err
 }
 
+// SetDeadline will set the deadline for reads and writes.
+// A zero value for t means I/O operations will not time out.
+func (c *DeadlineConn) SetDeadline(t time.Time) error {
```
vadmeste, line 103: Who will call SetDeadline/SetReaddeadline/SetWriteDeadline/Close() ?

internal/deadlineconn/deadlineconn.go
klauspost: We must re-check after getting the lock, otherwise we could override a SetReadDeadLine.

internal/deadlineconn/deadlineconn.go
klauspost: Same

internal/deadlineconn/deadlineconn.go
klauspost: It can be called by anyone using the connection, and `net.Conn` explicitly allows concurrent calling of methods, like http (and others) are doing, so multiple goroutines can call different functions.

klauspost:
Linter issues: https://github.com/minio/minio/pull/20633

harshavardhana (approved):

harshavardhana:
@klauspost this PR is not triggering a bug since this is not enabled at all in this branch and also in the eos we use this for listener side which is triggering a bug. 

```
make verify
Checking dependencies
Building minio binary with -race to './minio'
Installing minio binary with -race to '/home/harsha/go/bin/minio'
Verifying build with race
Initializing environment
Testing in FS setup
The cluster 'verify' is unreachable: Get "http://127.0.0.1:9000/minio/health/cluster": dial tcp 127.0.0.1:9000: connect: connection refused
The cluster 'verify' is ready
Dependency validation complete
{"name": "mc", "duration": "306", "function": "test_make_bucket", "status": "FAIL", "error": "/home/harsha/go/src/github.com/miniohq/eos/.verify-30089/mc --config-dir /tmp/.mc-20188 --quiet --no-color rb myminio/mc-test-bucket-17009  >>> \nmc: <ERROR> Failed to remove `myminio/mc-test-bucket-17009`. The specified bucket does not exist."}
INFO: Formatting 1st pool, 1 set(s), 1 drives per set.
INFO: WARNING: Host local has more than 0 drives of set. A host failure will result in data becoming unavailable.
MinIO Object Storage Server
Version: DEVELOPMENT.2024-11-11T17-15-55Z (go1.22.7 linux/amd64)

API: http://10.0.0.185:9000  http://10.0.3.1:9000  http://172.17.0.1:9000  http://172.19.0.1:9000  http://127.0.0.1:9000 
WebUI: http://10.0.0.185:42011 http://10.0.3.1:42011 http://172.17.0.1:42011 http://172.19.0.1:42011 http://127.0.0.1:42011     

Docs: https://docs.min.io
INFO: Exiting on signal: TERMINATED
FAILED
make: *** [Makefile:156: verify] Error 1
```

It is consistently reproducible going to revert the cherry-picked changed.

klauspost:
@harshavardhana It is used here: https://github.com/minio/minio/blob/91805bcab69f270e1b75a755658ce36423ee4f7e/internal/http/dial_linux.go#L130

I will check the other use.

PR #20684, opened 2024-11-25
Fix prefix validation in lifecycle rule

dhananjaykrutika:

## Description

Attempts to create a lifecycle rule using cURL when the xml has no prefix fails with `errXMLNotWellFormed` error:

```
$ curl -X PUT "http://localhost:9001/bucket-1?lifecycle" -H "Content-Type: application/xml" -H "Content-MD5: is0WM2LlKTzQECbf4PMRrQ==" --data '<LifecycleConfiguration><Rule><Expiration><Days>6</Days></Expiration><ID>Expiration-rule</ID><Status>Enabled</Status></Rule></LifecycleConfiguration>'
<Error><Code>InvalidArgument</Code><Message>The XML you provided was not well-formed or did not validate against our published schema</Message><BucketName>bucket-1</BucketName><Resource>/bucket-1</Resource><RequestId>180B2C661117A385</RequestId><HostId>7987905dee74cdeb212432486a178e511309594cee7cb75f892cd53e35f09ea4</HostId></Error>

```

## Motivation and Context

In the now deprecated `PutBucketLifecycle` API, Rule had a mandatory Prefix element and there existed no Filter field.
(See https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutBucketLifecycle.html)
In the newer `PutBucketLifecycleConfiguration` API, Rule has a prefix field that is deprecated, and there exists an optional Filter field, and within it, an optional Prefix field.
(See https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutBucketLifecycleConfiguration.html)

Existing prefix validation fails if Rule.Prefix does not exist and there are no Filters. This behavior is incorrect since Prefix itself is not mandatory in `PutBucketLifecycleConfiguration` API.

A valid rule could be a pre-existing one created using the now deprecated `PutBucketLifecycle`.
Or, a valid rule could also be either a pre-existing or a new rule that is created using `PutBucketLifecycleConfiguration`.

Prefix validation has been modified to check that either Rule.Prefix or Rule.Filter.Prefix exist but not both.
In doing so, we are assuming that any pre-existing rule created using PutBucketLifecycle API is already valid; so validation won't fail if Rule.Prefix is empty.

## Types of changes

## Checklist:

### DIFF
--- a/internal/bucket/lifecycle/rule.go
+++ b/internal/bucket/lifecycle/rule.go
@@ -86,11 +86,11 @@ func (r Rule) validateNoncurrentExpiration() error {
 func (r Rule) validatePrefixAndFilter() error {
 	// In the now deprecated PutBucketLifecycle API, Rule had a mandatory Prefix element and there existed no Filter field.
 	// See https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutBucketLifecycle.html
-	// In the newer PutBucketLifecycleConfiguraiton API, Rule has a prefix field that is deprecated, and there exists an optional
+	// In the newer PutBucketLifecycleConfiguration API, Rule has a prefix field that is deprecated, and there exists an optional
 	// Filter field, and within it, an optional Prefix field.
 	// See https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutBucketLifecycleConfiguration.html
 	// A valid rule could be a pre-existing one created using the now deprecated PutBucketLifecycle.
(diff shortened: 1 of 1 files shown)


dhananjaykrutika:
This PR attempts to fix https://github.com/minio/minio/discussions/20580

klauspost (approved):
If feasible add tests for these set/unset, AFAICT it shouldn't be too crazy to add.

harshavardhana (approved):

PR #20743, opened 2024-12-08
Disable mint full object tests

klauspost:
## Description

Remove expected failures from https://github.com/minio/minio-go/pull/2026

## Types of changes

### DIFF
--- a/internal/hash/reader.go
+++ b/internal/hash/reader.go
@@ -260,7 +260,9 @@ func (r *Reader) Read(p []byte) (int, error) {
 	// If we have reached our expected size,
 	// do one more read to ensure we are at EOF
 	// and that any trailers have been read.
-	if r.size >= 0 && r.bytesRead >= r.size {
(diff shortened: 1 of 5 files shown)


klauspost:
Let me just see if I can fix the mint issue as well.

internal/hash/reader.go
```diff
@@ -257,6 +257,24 @@ func (r *Reader) Read(p []byte) (int, error) {
 		r.contentHasher.Write(p[:n])
 	}
 
+	// If we have reached our expected size,
```
klauspost, line 260: I am not sure if this is really needed, but it is a nice safety net if downstream should add an `io.LimitedReader` that will not hit EOF.

Trailing checksums are relying on hitting EOF to be verified, since they are after the content.

marktheunissen (approved):

harshavardhana (approved):

PR #20761, opened 2024-12-12
Enforce a bucket limit of 100 to v2 metrics calls

klauspost:
## Description

Enforce a bucket count limit on metrics for v2 calls.

If people hit this limit, they should move to v3, as certain calls explode with high bucket count.

Reviewers: This *should* only affect v2 calls, but the complexity is rather overwhelming.

## How to test this PR?

Many buckets, etc.

## Types of changes
- [x] OOM prevention


cmd/metrics-v2.go
```diff
@@ -1834,7 +1838,7 @@ func getGoMetrics() *MetricsGroupV2 {
 //
 // The last parameter is added for compatibility - if true it lowercases the
 // `api` label values.
-func getHistogramMetrics(hist *prometheus.HistogramVec, desc MetricDescription, toLowerAPILabels bool) []MetricV2 {
+func getHistogramMetrics(hist *prometheus.HistogramVec, desc MetricDescription, toLowerAPILabels, limitBuckets bool) []MetricV2 {
```
anjalshireesh, line 1841: Maybe we should update the last line of the comment as `toLowerAPILabels` is not the `last parameter` anymore.

feorlen:
There's a section in the v2 metrics docs for bucket metrics, does this affect all of them or only some?
https://github.com/minio/minio/blob/master/docs/metrics/prometheus/list.md#bucket-metrics

klauspost:
@feorlen I have no idea to be honest.

feorlen:
@klauspost suggestions for how to determine which metrics are affected? 

I presume more thorough testing would uncover it by observation, not sure the best way to accomplish that. This will be an unpleasant surprise if it's not documented. (Likely so in any case, but we do our best.)

klauspost:
@feorlen I don't know, and I don't really care too much tbh. Maybe some of the reviewers know.

A) The v3 api is exactly there for fixing this.
B) This scales horrible with many buckets and will end up just crashing servers unless contained.
C) Metrics is after all secondary to operation of the cluster.

The only thing I really care is if there are more cases than the ones we observe that have similar potential to overload systems.

If you are thinking about documenting, adding a note that some v2 endpoints may restrict the maximum number of bucket metrics and if encountered, start migrating to v3.

shtripat:
@feorlen I think it applies to all the bucket metrics in v2.
But as @klauspost mentioned metrics is second class operation, as long as documented about the  cap on no. of buckets reported in metrics, it should be fine.
If users actually need very bucket metrics, move to v3. But would to hear from other reviewers.

anjalshireesh (approved):
LGTM, one very minor comment

anjalshireesh:
> There's a section in the v2 metrics docs for bucket metrics, does this affect all of them or only some? https://github.com/minio/minio/blob/master/docs/metrics/prometheus/list.md#bucket-metrics

@feorlen based on my understanding of the code, following metrics will be affected:

```
minio_s3_requests_ttfb_seconds_distribution
minio_node_tier_ttlb_seconds_distribution
minio_bucket_requests_ttfb_seconds_distribution
```

Following are bucket level metrics, but are not affected as they don't use the code modified in this PR:

```
minio_bucket_objects_size_distribution
minio_bucket_objects_version_distribution
```

@klauspost do these two also need be changed?

klauspost:
@anjalshireesh So in `getBucketUsageMetrics`?

anjalshireesh:
> @anjalshireesh So in `getBucketUsageMetrics`?

Yes. `minio_bucket_replication_latency_ms` also looks like another bucket level histogram metric (in the same function)

feorlen:
The text in the `# Bucket Metrics` section is automatically picked up by the web docs. It's not ideal, but this should be mentioned somewhere.

Are all the affected metrics in the same section under Bucket Metrics? Give me a few, I'll go look and propose some text.

feorlen:
They are scattered throughout, so no one place to put new text. Best option is to add it to the individual descriptions. Something like:
```
| `metric` | Existing description. This metric is only reported for the first 100 buckets. |
```
How do we determine which ones you get?
_looks at diff_ alpha by bucket name?

Here are where they are in list.md
```
# Cluster metrics
## S3 API Request Metrics
minio_s3_requests_ttfb_seconds_distribution

# Bucket Metrics
## Requests Metrics
minio_bucket_requests_ttfb_seconds_distribution

# Bucket Metrics
## Distribution Metrics
minio_bucket_objects_size_distribution
minio_bucket_objects_version_distribution

# Bucket Metrics
## Replication Metrics
minio_bucket_replication_latency_ms
```

This one is not in list.md
```
minio_node_tier_ttlb_seconds_distribution
```

klauspost:
All should be sorted alphabetically, and first 100 selected. Please be exact with what you want me to change and where. 

We should have this in ASAP. Unless there is a code issue we can deal with the rest later.

feorlen:
Can discuss docs after 👍🏻 

Edit: 
When we do get to it, maybe this in the descriptions:
```
Reported for the first 100 buckets only, sorted alphabetically by bucket name.
```

anjalshireesh:
> This one is not in list.md
> 
> ```
> minio_node_tier_ttlb_seconds_distribution
> ```

@feorlen it is there, but misspelt as `minio_node_tier_tier_ttlb_seconds_distribution`

anjalshireesh (approved):
LGTM

allanrogerr:
LGTM
Note that distribution metrics will appear to exceed the limit due to any additional filtering needed. e.g.
```
minio_bucket_objects_size_distribution{range="BETWEEN_1024B_AND_1_MB"}
minio_bucket_requests_inflight_total{api="getbucketlocation"}
```
Also the sorting is lexicographical so users may not understand immediately why they are not seeing all their buckets in the metrics

allanrogerr (approved):

cmd/metrics-v3-types.go
```diff
@@ -267,7 +267,7 @@ func (m *MetricValues) SetHistogram(name MetricName, hist *prometheus.HistogramV
 		panic(fmt.Sprintf("metric has no description: %s", name))
 	}
 	dummyDesc := MetricDescription{}
-	metricsV2 := getHistogramMetrics(hist, dummyDesc, false)
+	metricsV2 := getHistogramMetrics(hist, dummyDesc, false, false)
```
allanrogerr, line 270: @klauspost @anjalshireesh  I'm not sure why this V2 functionality is invoked by V3 metrics. Will it be limited by this as well. I haven't tested this case as yet.

allanrogerr: Tested all V3 metrics as well. They are not affected.

allanrogerr (approved):

PR #20795, opened 2024-12-26
Helm: Allow users to set storageClass from global parameters

iambugra:

## Description
This allows users to set the storageClass from global parameters.

## Motivation and Context
Global parameters are useful, especially when building a chart with many dependency charts. I would like to set the storageClass parameter in the parent chart once and see it's set on the subcharts, without needing to set that value explicitly on each chart's values.

## How to test this PR?
```sh
cat > .values.yaml <<EOF
global:
  storageClass: "global"

persistence:
  storageClass: "local"
EOF
helm template helm/minio -s templates/statefulset.yaml -f .values.yaml
```

## Types of changes

## Checklist:

### DIFF
--- a/helm/minio/templates/_helpers.tpl
+++ b/helm/minio/templates/_helpers.tpl
@@ -93,6 +93,27 @@ Return the appropriate apiVersion for console ingress.
 {{- end -}}
 
 {{/*
+Return the appropriate storageClassName.
+{{ include "minio.persistence.storageClass" ( dict "persistence" .Values.path.to.the.persistence "global" .Values.global) }}
+*/}}
+{{- define "minio.persistence.storageClass" -}}
+  {{- $storageClass := .persistence.storageClass -}}
+  {{- if .global -}}
+      {{- if .global.storageClass -}}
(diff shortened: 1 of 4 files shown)


allanrogerr:
This functionality is already available in the current chart under https://github.com/iambugra/minio/blob/master/helm/minio/values.yaml#L170.

The net effect is the same ability to set storage class on pvc and statefulset for example, for a single line in `values.yaml`.

What is the actual contribution here?

iambugra:
Yes, setting [here](https://github.com/iambugra/minio/blob/master/helm/minio/values.yaml#L170) sets the storage class of the PV, but with this PR I am adding global parameters, which is better explained in [here](https://helm.sh/docs/chart_template_guide/subcharts_and_globals/). 

Basically, this feature is useful when there is a parent chart, and minio is the subchart of this parent chart. Setting the `.Values.global.storageClass` in the parent chart overrides the storage class of every subchart that implements global stoage class. This is what I want to achieve: setting the storage class in one place and see it's set on every PV of the parent chart and every PV of the subcharts.

Issue #172, opened 2015-02-18
Is pip a real dependency of the project ?


vadmeste:
As I understood, pip is only needed to build mkdocs, so I think it is useless to check it in checkdeps.sh

harshavardhana:
> As I understood, pip is only needed to build mkdocs, so I think it is useless to check it in checkdeps.sh

pip is necessary for mkdocs, checkdeps.sh is executed during the `make`  and so is `mkdocs` with `go generate` - so it is a valid check

vadmeste:
yes, so the project doesn't depend directly on pip, right? I mean isn't it sufficient to check only the presence of mkdocs ? because I guess some people would prefer easy_install which is an alternative of pip..

harshavardhana:
> yes, so the project doesn't depend directly on pip, right? I mean isn't it sufficient to check only the presence of mkdocs ? because I guess some people would prefer easy_install which is an alternative of pip..

You are correct.

PR #20821, opened 2025-01-09
DeleteObjects: Send delete to all pools (#172)

vadmeste:

## Description
Currently, DeleteObjects() tries to find the object's pool before sending a delete request. This only works well when an object has multiple versions in different pools since looking for the pool does not consider the version-id. When an S3 client wants to remove a version-id that exists in pool 2, the delete request will be directed to pool one because it has another version of the same object.

This commit will remove looking for pool logic and will send a delete request to all pools in parallel. This should not cause any performance regression in most of the cases since the object will unlikely exist in only one pool, and the performance price will be similar to getPoolIndex() in that case.

## Motivation and Context
Fix DeleteObjects() of versions of an object that has versions in mutiple pools,

## Types of changes

## Checklist:

### DIFF
--- a/cmd/erasure-object.go
+++ b/cmd/erasure-object.go
@@ -1732,12 +1732,6 @@ func (er erasureObjects) DeleteObjects(ctx context.Context, bucket string, objec
 					continue
 				}
 				for _, v := range dedupVersions[i].Versions {
-					if err == errFileNotFound || err == errFileVersionNotFound {
-						if !dobjects[v.Idx].DeleteMarker {
-							// Not delete marker, if not found, ok.
-							continue
-						}
-					}
 					delObjErrs[index][v.Idx] = err
 				}
 			}
@@ -1757,6 +1751,13 @@ func (er erasureObjects) DeleteObjects(ctx context.Context, bucket string, objec
 			}
 		}
 		err := reduceWriteQuorumErrs(ctx, diskErrs, objectOpIgnoredErrs, writeQuorums[objIndex])
+		if err == nil {
+			dobjects[objIndex].found = true
+		} else if isErrVersionNotFound(err) || isErrObjectNotFound(err) {
+			if !dobjects[objIndex].DeleteMarker {
+				err = nil
+			}
+		}
 		if objects[objIndex].VersionID != "" {
 			errs[objIndex] = toObjectErr(err, bucket, objects[objIndex].ObjectName, objects[objIndex].VersionID)
 		} else {
--- a/cmd/erasure-object.go
+++ b/cmd/erasure-object.go
@@ -27,6 +27,8 @@ import (
 	"net/http"
 	"path"
 	"runtime"
+	"slices"
+	"sort"
 	"strconv"
 	"strings"
 	"sync"
@@ -1706,8 +1708,21 @@ func (er erasureObjects) DeleteObjects(ctx context.Context, bucket string, objec
 	}
 
 	dedupVersions := make([]FileInfoVersions, 0, len(versionsMap))
-	for _, version := range versionsMap {
-		dedupVersions = append(dedupVersions, version)
+	for _, fivs := range versionsMap {
+		// Removal of existing versions and adding a delete marker in the same
+		// request is supported. At the same time, we cannot allow adding
+		// two delete markers on top of any object. To avoid this situation,
+		// we will sort deletions to execute existing deletion first,
+		// then add only one delete marker if requested
+		sort.SliceStable(fivs.Versions, func(i, j int) bool {
+			return !fivs.Versions[i].Deleted
+		})
+		if idx := slices.IndexFunc(fivs.Versions, func(fi FileInfo) bool {
+			return fi.Deleted
+		}); idx > -1 {
+			fivs.Versions = fivs.Versions[:idx+1]
+		}
+		dedupVersions = append(dedupVersions, fivs)
 	}
 
 	// Initialize list of errors.
--- a/cmd/xl-storage.go
+++ b/cmd/xl-storage.go
@@ -30,7 +30,6 @@ import (
 	"path/filepath"
 	"runtime"
 	"slices"
-	"sort"
 	"strconv"
 	"strings"
 	"sync"
@@ -1091,7 +1090,6 @@ func (s *xlStorage) ListDir(ctx context.Context, origvolume, volume, dirPath str
 	return entries, nil
 }
 
-// Delete multiple versions of a single object - no support of replication logic
 func (s *xlStorage) deleteVersions(ctx context.Context, volume, path string, fis ...FileInfo) error {
 	volumeDir, err := s.getVolDir(volume)
 	if err != nil {
@@ -1142,17 +1140,7 @@ func (s *xlStorage) deleteVersions(ctx context.Context, volume, path string, fis
 		return err
 	}
 
-	// Add any delete marker last.
-	sort.SliceStable(fis, func(i, j int) bool {
-		return !fis[i].Deleted
-	})
-
 	for _, fi := range fis {
-		if fi.Deleted && xlMeta.latestType() == DeleteType {
-			// Avoid adding a new delete-marker if
-			// there is one already on top.
-			continue
-		}
 		dataDir, err := xlMeta.DeleteVersion(fi)
 		if err != nil {
 			if !fi.Deleted && (err == errFileNotFound || err == errFileVersionNotFound) {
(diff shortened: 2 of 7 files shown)


klauspost:
> This should not cause any performance regression in most of the cases since the object will unlikely exist in only one pool, and the performance price will be similar to getPoolIndex() in that case.

I was going to call BS on that, but you are actually right :D  Good spot!

cmd/xl-storage.go
```diff
@@ -1140,7 +1141,17 @@ func (s *xlStorage) deleteVersions(ctx context.Context, volume, path string, fis
 		return err
 	}
 
+	latestV, err := xlMeta.getIdx(0)
+	if err != nil {
+		return err
+	}
```
klauspost: No need for a full unmarshal. You can add this helper, when simply do `if xlMeta.latestType() == DeleteType`.

```
// latestType returns the latest version type.
// returns invalidVersionType if no versions exist.
func (x *xlMetaV2) latestType() VersionType {
	if x == nil || len(x.versions) == 0 {
		return invalidVersionType
	}
	return x.versions[0].header.Type
}
```

klauspost: Also doing it in the loop will keep up with mutations (ie as deletes are happening).

A bit worried that the top delete marker could end up getting deleted if the fis contain the delete marker version id.

Couldn't we then end up with an object that *isn't* deleted?

Somehow it feels like we should apply all version-id deletes, and then after that see if we still should add the delete marker... Annoying.. Maybe you can find another way?

harshavardhana: All these changes must be done in the eos first @vadmeste since its a PR based off from that.

harshavardhana: ping @vadmeste ^^

harshavardhana: re-ping @vadmeste ^^

vadmeste: > Also doing it in the loop will keep up with mutations (ie as deletes are happening).

I am not sure I understand, we have object locks already, are you talking about something else ?

> A bit worried that the top delete marker could end up getting deleted if the fis contain the delete marker version id.

FileInfo.Deleted member  is false when you want to remove a delete marker and you know its version-id

@klauspost let's talk in private

klauspost: @vadmeste Sure.

harshavardhana: is there a consensus? @klauspost @vadmeste

klauspost (approved):
LGTM

harshavardhana:
Data race in the new changes

```
==================
WARNING: DATA RACE
Read at 0x00c004070cd0 by goroutine 354738:
  reflect.typedmemmove()
      /opt/hostedtoolcache/go/1.23.5/x64/src/runtime/mbarrier.go:225 +0x0
  internal/reflectlite.typedmemmove()
      /opt/hostedtoolcache/go/1.23.5/x64/src/runtime/mbarrier.go:243 +0x12
  sort.insertionSort_func()
      /opt/hostedtoolcache/go/1.23.5/x64/src/sort/zsortfunc.go:13 +0x83
  sort.stable_func()
      /opt/hostedtoolcache/go/1.23.5/x64/src/sort/zsortfunc.go:343 +0x7e
  sort.SliceStable()
      /opt/hostedtoolcache/go/1.23.5/x64/src/sort/slice.go:44 +0x96
  github.com/minio/minio/cmd.(*xlStorage).deleteVersions()
      /home/runner/work/minio/minio/cmd/xl-storage.go:1146 +0x466
  github.com/minio/minio/cmd.(*xlStorage).DeleteVersions()
      /home/runner/work/minio/minio/cmd/xl-storage.go:1215 +0x25d
  github.com/minio/minio/cmd.(*xlStorageDiskIDCheck).DeleteVersions()
      /home/runner/work/minio/minio/cmd/xl-storage-disk-id-check.go:580 +0x438
  github.com/minio/minio/cmd.erasureObjects.DeleteObjects.func1()
      /home/runner/work/minio/minio/cmd/erasure-object.go:1729 +0x238
  github.com/minio/minio/cmd.erasureObjects.DeleteObjects.gowrap1()
      /home/runner/work/minio/minio/cmd/erasure-object.go:1738 +0x61

Previous write at 0x00c004070cd0 by goroutine 354724:
  reflect.typedmemmove()
      /opt/hostedtoolcache/go/1.23.5/x64/src/runtime/mbarrier.go:225 +0x0
  internal/reflectlite.typedmemmove()
      /opt/hostedtoolcache/go/1.23.5/x64/src/runtime/mbarrier.go:243 +0x12
  sort.insertionSort_func()
      /opt/hostedtoolcache/go/1.23.5/x64/src/sort/zsortfunc.go:13 +0x83
  sort.stable_func()
      /opt/hostedtoolcache/go/1.23.5/x64/src/sort/zsortfunc.go:343 +0x7e
  sort.SliceStable()
      /opt/hostedtoolcache/go/1.23.5/x64/src/sort/slice.go:44 +0x96
  github.com/minio/minio/cmd.(*xlStorage).deleteVersions()
      /home/runner/work/minio/minio/cmd/xl-storage.go:1146 +0x466
  github.com/minio/minio/cmd.(*xlStorage).DeleteVersions()
      /home/runner/work/minio/minio/cmd/xl-storage.go:1215 +0x25d
  github.com/minio/minio/cmd.(*xlStorageDiskIDCheck).DeleteVersions()
      /home/runner/work/minio/minio/cmd/xl-storage-disk-id-check.go:580 +0x438
  github.com/minio/minio/cmd.erasureObjects.DeleteObjects.func1()
      /home/runner/work/minio/minio/cmd/erasure-object.go:1729 +0x238
  github.com/minio/minio/cmd.erasureObjects.DeleteObjects.gowrap1()
      /home/runner/work/minio/minio/cmd/erasure-object.go:1738 +0x61

Goroutine 354738 (running) created at:
  github.com/minio/minio/cmd.erasureObjects.DeleteObjects()
      /home/runner/work/minio/minio/cmd/erasure-object.go:1720 +0x10a6
  github.com/minio/minio/cmd.(*erasureSets).DeleteObjects.func2()
      /home/runner/work/minio/minio/cmd/erasure-sets.go:825 +0x3e6
  github.com/minio/minio/cmd.(*erasureSets).DeleteObjects.gowrap1()
      /home/runner/work/minio/minio/cmd/erasure-sets.go:832 +0x6e

Goroutine 354724 (running) created at:
  github.com/minio/minio/cmd.erasureObjects.DeleteObjects()
      /home/runner/work/minio/minio/cmd/erasure-object.go:1720 +0x10a6
  github.com/minio/minio/cmd.(*erasureSets).DeleteObjects.func2()
      /home/runner/work/minio/minio/cmd/erasure-sets.go:825 +0x3e6
  github.com/minio/minio/cmd.(*erasureSets).DeleteObjects.gowrap1()
      /home/runner/work/minio/minio/cmd/erasure-sets.go:832 +0x6e
==================
    testing.go:1399: race detected during execution of test
````

**Comment by klauspost** [comment]
I guess the same slice is sent to several disks... Maybe sort upstream?

**Comment by vadmeste** [comment]
> I guess the same slice is sent to several disks... Maybe sort upstream?

yes, I moved the sorting to the upper layer, please take a look.

File: cmd/erasure-object.go [hunk 2]
```diff
@@ -1706,8 +1708,21 @@ func (er erasureObjects) DeleteObjects(ctx context.Context, bucket string, objec
...
 	dedupVersions := make([]FileInfoVersions, 0, len(versionsMap))
-	for _, version := range versionsMap {
-		dedupVersions = append(dedupVersions, version)
+	for _, fivs := range versionsMap {
+		// Removal of existing versions and adding a delete marker in the same
+		// request is supported. At the same time, we cannot allow adding
+		// two delete markers on top of any object. To avoid this situation,
+		// we will sort deletions to execute existing deletion first,
+		// then add only one delete marker if requested
+		sort.SliceStable(fivs.Versions, func(i, j int) bool {
```
harshavardhana, line 1717: Stable sort is only needed when you need to use for two different elements to sort. You should just use sort.Slice() 

The API name is confusing sort.SliceStable() doesn't mean that sort.Slice is unstable.

harshavardhana (changes requested):

cmd/erasure-object.go
klauspost: @harshavardhana It only sorts on a bool, so there will be a lot of ties. Therefore I would rather that it retains the order.

It is unstable as output in the two parts will be random. I would rather it just retain the order of these elements than jumble them up a random order.

harshavardhana: ah then it makes sense.

harshavardhana (approved):

harshavardhana:
@vadmeste can you send the updated changes in this PR back to AiStor fork?

PR #20846, opened 2025-01-16
Fix inconsistently written compressed files.

klauspost:
## Description

Before https://github.com/minio/minio/pull/20575 files could pick up indices from unrelated files if no indices were added.

This would result in these files not being consistent across a set.

When loading search for the compression indicators, check if within the problematic date range and clean up any parts that have an index, but shouldn't.

Test validates that the signature matches the one in files stored without an index.

Bumps xlMetaVersion, so this check doesn't have to be made for future versions.

## How to test this PR?

Test included. Requires object where part(s) are < 8MB.

## Types of changes

### DIFF
--- a/cmd/xl-storage-format-v2.go
+++ b/cmd/xl-storage-format-v2.go
@@ -462,7 +462,7 @@ func (j *xlMetaV2Version) ToFileInfo(volume, path string, allParts bool) (fi Fil
 
 const (
 	xlHeaderVersion = 3
-	xlMetaVersion   = 2
+	xlMetaVersion   = 3
 )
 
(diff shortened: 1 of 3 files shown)


harshavardhana (approved):

harshavardhana:
@klauspost this needs backport?

harshavardhana (changes requested):
this needs a change in xl-meta/main.go as well and also inspect?

klauspost:
@harshavardhana Backport to what?

harshavardhana:
> @harshavardhana Backport to what?

miniohq/eos - not obvious?

harshavardhana (approved):

PR #20847, opened 2025-01-16
s3: Provide enough buffer when the object final size is unknown

vadmeste:

## Description
When compression is enabled, the final object size is not calculated, in that case, 
we need to make sure that the provided buffer is always bigger than the shard size 
the bitrot will always calculate the hash of blocks with shard size, except the last block.

## How to test this PR?
Create a file with a length size of 5873240, enable compression+encryption, upload and try to read it.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/erasure-multipart.go
+++ b/cmd/erasure-multipart.go
@@ -626,17 +626,13 @@ func (er erasureObjects) PutObjectPart(ctx context.Context, bucket, object, uplo
 	switch size := data.Size(); {
 	case size == 0:
 		buffer = make([]byte, 1) // Allocate at least a byte to reach EOF
-	case size == -1:
-		if size := data.ActualSize(); size > 0 && size < fi.Erasure.BlockSize {
-			// Account for padding and forced compression overhead and encryption.
-			buffer = make([]byte, data.ActualSize()+256+32+32, data.ActualSize()*2+512)
+	case size >= fi.Erasure.BlockSize || size == -1:
+		if int64(globalBytePoolCap.Load().Width()) < fi.Erasure.BlockSize {
+			buffer = make([]byte, fi.Erasure.BlockSize, 2*fi.Erasure.BlockSize)
 		} else {
 			buffer = globalBytePoolCap.Load().Get()
 			defer globalBytePoolCap.Load().Put(buffer)
 		}
-	case size >= fi.Erasure.BlockSize:
-		buffer = globalBytePoolCap.Load().Get()
-		defer globalBytePoolCap.Load().Put(buffer)
 	case size < fi.Erasure.BlockSize:
 		// No need to allocate fully fi.Erasure.BlockSize buffer if the incoming data is smaller.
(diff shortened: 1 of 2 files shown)


cmd/erasure-multipart.go
```diff
@@ -626,17 +626,13 @@ func (er erasureObjects) PutObjectPart(ctx context.Context, bucket, object, uplo
 	switch size := data.Size(); {
 	case size == 0:
 		buffer = make([]byte, 1) // Allocate at least a byte to reach EOF
-	case size == -1:
-		if size := data.ActualSize(); size > 0 && size < fi.Erasure.BlockSize {
-			// Account for padding and forced compression overhead and encryption.
-			buffer = make([]byte, data.ActualSize()+256+32+32, data.ActualSize()*2+512)
+	case size >= fi.Erasure.BlockSize || size == -1:
+		if int64(globalBytePoolCap.Load().Width()) < fi.Erasure.BlockSize {
+			buffer = make([]byte, fi.Erasure.BlockSize, 2*fi.Erasure.BlockSize)
```
klauspost, line 631: This doesn't account for the overhead added by the encryption and extra bytes for incompressible files, and those will run out of space when N==M.

Could you post an actual reproducer?

klauspost (changes requested):

cmd/erasure-multipart.go
klauspost: (to make it clearer, the output can be somewhat *bigger* than `fi.Erasure.BlockSize`, so when adding erasure codes `fi.Erasure.BlockSize*2` may not be enough).

vadmeste:
```
#!/bin/bash

export MINIO_KMS_SECRET_KEY=my-minio-key:OSMM+vkKUTCvQs9YL/CVMIMt43HFhkUpqJxTmGl6rYw=
export MC_UPLOAD_MULTIPART_SIZE=5245522
export CI=1

./minio server /tmp/x/{1...4}/ &

sleep 5

mc alias set myminio http://localhost:9000 minioadmin minioadmin
mc admin config set myminio compression enable=on extensions=.pdf allow_encryption=on

mc mb myminio/testbucket/
mc encrypt set sse-kms my-minio-key myminio/testbucket

mc cp file.pdf myminio/testbucket/file.pdf
mc cat myminio/testbucket/file.pdf | wc -c

```

I sent you a DM for file.pdf, it is not easy to reproduce with other random file with the same size

Most importantly, the buffer size needs to be always higher than the shard size, otherwise it can confuse the bitrot streaming code.

klauspost: Ah. I see what it is doing now. Yeah. No need for custom buffers unless block size was changed since startup, which AFAICT is the only case where the additional alloc would be hit.

klauspost (approved):

klauspost:
No wait. Let me just double check all if this again.

klauspost:
Yeah. I don't see a block ever being bigger than BlockSize, and BlockSize*2 when erasure codes are added.

harshavardhana (approved):

PR #20860, opened 2025-01-21
update name max limits on windows

dormanze:

## Description
update name max limits on windows and replacement of some constants

## Motivation and Context
refer: https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry

## Types of changes

## Checklist:

### DIFF
--- a/cmd/xl-storage.go
+++ b/cmd/xl-storage.go
@@ -85,7 +85,7 @@ func isValidVolname(volname string) bool {
 		return false
 	}
 
-	if runtime.GOOS == "windows" {
+	if runtime.GOOS == globalWindowsOSName {
 		// Volname shouldn't have reserved characters in Windows.
 		return !strings.ContainsAny(volname, `\:*?\"<>|`)
 	}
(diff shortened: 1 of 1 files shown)


cmd/xl-storage.go
```diff
@@ -131,13 +131,14 @@ type xlStorage struct {
...
+	if runtime.GOOS == globalMacOSName && len(pathName) > 1016 {
 		return errFileNameTooLong
 	}
 
-	// Disallow more than 1024 characters on windows, there
+	// Disallow more than 32,767 characters on windows, there
 	// are no known name_max limits on Windows.
-	if runtime.GOOS == "windows" && len(pathName) > 1024 {
+	// Refer: https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry
+	if runtime.GOOS == globalWindowsOSName && len(pathName) > 32767 {
```
harshavardhana, line 141: this is not suitable for MinIO - we need to return error for object names > 1024 so this is not the same situation as `mc`

harshavardhana (changes requested):

cmd/xl-storage.go
dormanze: I understand you're saying that minio limits the length of an object's name to 1024, which is to verify the length of the entire path, not just the object.
like: /export1/bucket/testfile/xl.meta
object's name check, maybe in https://github.com/minio/minio/blob/master/cmd/object-api-utils.go#L200

harshavardhana: correct but this path is well with in that margin, we don't care that windows supports 32k characters.

harshavardhana:
this PR is not necessary @dormanze

Issue #20862, opened 2025-01-22
MinIO server stuck in a deadlock after a surge in requests


fpetkovski:
## NOTE

We run an analytics engine against a MinIO cluster which can sometimes lead to a lot of range requests getting executed at the same time.
During high load, some MinIO servers would deadlock and become unresponsive. The screenshot bellow illustrates what happens to 4 servers after a spike in requests (our erasure set size is 4). CPU spikes and comes back down to 1 core per server. The number of goroutines remained high until I manually restarted the hanging servers. Any new requests against those 4 servers timed out.

![Image](https://github.com/user-attachments/assets/11a952f8-c61f-44b3-9823-26c4f6f5a361)

I also captured CPU profiles using an eBPF profiler which I am also attaching to this issue: [profile.pb.gz](https://github.com/user-attachments/files/18507398/profile.pb.gz). From a superficial view, it seems like those servers are stuck in some lock removal process as a result of the acute spike in requests.

![Image](https://github.com/user-attachments/assets/500ac6a0-3d17-400a-b6d7-e29e164cf6cc)

## Expected Behavior
I would expect the 4 affected servers to become quickly become responsive.

## Current Behavior
The 4 affected servers become unresponsive.

## Possible Solution
I suspect that the [removeEntry](https://github.com/minio/minio/blob/master/cmd/local-locker.go#L138-L161)  and [expireOldLocks](https://github.com/minio/minio/blob/master/cmd/local-locker.go#L373-L404) compete for the same mutex and block each other. 

I also see that `expireOldLocks` makes conditional progress only when a lock expires, so perhaps that's keeping the mutex locked for a long time.

## Steps to Reproduce (for bugs)
N/A

## Context
The reason why 4 servers are affected is because the receive many requests for a single file at the same time. These 4 servers are part of an erasure set and hold shards of that file.

## Regression
No

## Your Environment
We run in GCP with locally attached SSDs. 
Minio version: RELEASE.2024-11-07T00-52-20Z

klauspost:
Unlocks are still pending/queued. It is not deadlocked.

We are looking into changing the locking for AIStor, but it will not be part of the community release.

fpetkovski:
Here are the locks for an affected server, all metrics are set to 0.

![Image](https://github.com/user-attachments/assets/537a3748-f720-4a7b-a9df-31ed49d5fe69)

Current goroutines

[...]

PR #20867, opened 2025-01-24
Use a timeout when acquiring RLocks to objects.

fpetkovski:
## Description

As illustrated in https://github.com/minio/minio/issues/20862, a large surge of requests against the same object can overwhelm a single minio server. The root cause for this seems to be the lock acquisition process which currently has no timeout on the server owning the requested object. Lock acquisition is still ongoing on the server even when the client request has timed out or failed.

Once lock acquisition in the client times out, the server will still lock the object and return success, causing the client to spend additional time releasing the object.

In order to make the server recover quickly after requests fail, this commit adds a timeout to acquiring read locks. The timeout is set from the server and is optional. Ideally this would be resolved through context cancellation, but contexts do not seem to be used in the RPC framework.

## How to test this PR?

The change is tested through unit tests. 

## Types of changes

## Checklist:

### DIFF
--- a/cmd/local-locker.go
+++ b/cmd/local-locker.go
@@ -164,9 +164,19 @@ func (l *localLocker) RLock(ctx context.Context, args dsync.LockArgs) (reply boo
 	if len(args.Resources) != 1 {
 		return false, fmt.Errorf("internal error: localLocker.RLock called with more than one resource")
 	}
+	if args.TimeoutMillis > 0 {
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithTimeout(ctx, time.Duration(args.TimeoutMillis)*time.Millisecond)
+		defer cancel()
+	}
 
 	l.mutex.Lock()
 	defer l.mutex.Unlock()
+
+	if ctx.Err() != nil {
+		return false, ctx.Err()
+	}
+
 	resource := args.Resources[0]
 	now := UTCNow()
 	lrInfo := lockRequesterInfo{
--- a/cmd/local-locker.go
+++ b/cmd/local-locker.go
@@ -160,21 +160,18 @@ func (l *localLocker) removeEntry(name string, args dsync.LockArgs, lri *[]lockR
 	return false
 }
 
-func (l *localLocker) RLock(ctx context.Context, args dsync.LockArgs) (reply bool, err error) {
+func (l *localLocker) RLock(_ context.Context, args dsync.LockArgs) (reply bool, err error) {
 	if len(args.Resources) != 1 {
 		return false, fmt.Errorf("internal error: localLocker.RLock called with more than one resource")
 	}
-	if args.TimeoutMillis > 0 {
-		var cancel context.CancelFunc
-		ctx, cancel = context.WithTimeout(ctx, time.Duration(args.TimeoutMillis)*time.Millisecond)
-		defer cancel()
-	}
+
+	lockStart := UTCNow()
 
 	l.mutex.Lock()
 	defer l.mutex.Unlock()
 
-	if ctx.Err() != nil {
-		return false, ctx.Err()
+	if time.Since(lockStart) > time.Duration(args.TimeoutMillis)*time.Millisecond {
+		return false, fmt.Errorf("internal error: localLocker.RLock timed out")
 	}
 
 	resource := args.Resources[0]
--- a/internal/dsync/lock-args.go
+++ b/internal/dsync/lock-args.go
@@ -37,6 +37,10 @@ type LockArgs struct {
 
 	// Quorum represents the expected quorum for this lock type.
 	Quorum *int `msgp:"omitempty"`
+
+	// TimeoutMillis is the timeout for acquiring the lock in milliseconds.
+	// A value less than or equal to 0 indicates no timeout.
+	TimeoutMillis int64 `msgp:"omitempty"`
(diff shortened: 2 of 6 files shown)


cmd/local-locker.go
```diff
@@ -164,9 +164,19 @@ func (l *localLocker) RLock(ctx context.Context, args dsync.LockArgs) (reply boo
...
 	}
+	if args.TimeoutMillis > 0 {
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithTimeout(ctx, time.Duration(args.TimeoutMillis)*time.Millisecond)
+		defer cancel()
+	}
 
 	l.mutex.Lock()
 	defer l.mutex.Unlock()
+
```
fpetkovski, line 172: It would be even better to turn the mutex into a channel so that we can wait in parallel for acquiring the lock and for the timeout. I am happy to change that as part of the PR if accepted.

cmd/local-locker.go
```diff
@@ -164,9 +164,19 @@ func (l *localLocker) RLock(ctx context.Context, args dsync.LockArgs) (reply boo
 	if len(args.Resources) != 1 {
 		return false, fmt.Errorf("internal error: localLocker.RLock called with more than one resource")
 	}
+	if args.TimeoutMillis > 0 {
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithTimeout(ctx, time.Duration(args.TimeoutMillis)*time.Millisecond)
+		defer cancel()
+	}
```
klauspost: No need to alloc a new context here. It is simpler to just check time before/after `l.mutex.Lock()`.

internal/dsync/lock-args.go
```diff
@@ -37,6 +37,10 @@ type LockArgs struct {
 
 	// Quorum represents the expected quorum for this lock type.
 	Quorum *int `msgp:"omitempty"`
+
+	// TimeoutMillis is the timeout for acquiring the lock in milliseconds.
+	// A value less than or equal to 0 indicates no timeout.
+	TimeoutMillis int64 `msgp:"omitempty"`
```
klauspost, line 43: The timeout would be the same on all clients, so no real need to xfer it.

Also right now it is only applied to Rlock, right?

klauspost (commented):
In AIStor we added an atomic counter on pending lock requests that instantly rejects new (r)lock requests if it exceeds a threshold, since that also protects against burst requests, and presumably should protect against an un-runnable situation.

It also adds the `dsync.DefaultTimeouts.Acquire` to the context on incoming lock requests.

TBH I would rather backport that (EOS PR 212) than having diverging code. I will leave it to @harshavardhana to decide.

cmd/local-locker.go
klauspost: Doesn't really matter too much, if you do that you would need to spin up a goroutine for every blocked request, that would live until the lock is acquired.

cmd/local-locker.go
fpetkovski: Makes sense, updated.

internal/dsync/lock-args.go
fpetkovski: The timeout is applied on all clients, but I was not sure whether RLock gets called for operations other than GetObject and did not want to change a global default.

> Also right now it is only applied to Rlock, right?

That is correct, we can change it to apply to Lock as well.

cmd/local-locker.go
fpetkovski: My thought was to use something like:
```go 
select {
	case <-ctx.Done():
	    return false, ctx.Err()
	case mutex <- struct{}{}:
            defer func() { <-mutex }()
	}
}
```

I don't know whether performance would improve though, I haven't run any tests.

klauspost: How would you "feed" the mutex channel?

fpetkovski: The `mutex` channel would a field of the `localLocker` struct and is initialized to capacity of 1. This means only one Lock/Rlock invocation can send an item to it. Exiting from the method would remove the item from the channel, allowing pending locks to send a new item. This way a Lock/RLock invocation could return as soon as the timeout expires.

The main difference is that with the current implementation the method still has to lock the mutex even when the timeout has expired, likely causing more contention.

Note that locking the mutex requires sending an item to the channel, and unlocking it requires reading an item from it.

klauspost: I have tested that already and the raw throughput tanks compared to a mutex. You are welcome to try -`Test_localLocker_RUnlock` should be fine for benchmarking baseline.

harshavardhana:
> In AIStor we added an atomic counter on pending lock requests that instantly rejects new (r)lock requests if it exceeds a threshold, since that also protects against burst requests, and presumably should protect against an un-runnable situation.
> 
> It also adds the `dsync.DefaultTimeouts.Acquire` to the context on incoming lock requests.
> 
> TBH I would rather backport that (EOS PR 212) than having diverging code. I will leave it to @harshavardhana to decide.

Yeah adding this backport is fine @klauspost

klauspost:
@fpetkovski I will leave this open and add the backport tomorrow. We can then compare and see what should be brought over so they are comparable.

fpetkovski:
Sounds good to me, thanks for taking a look.

klauspost:
See #20876

PR #20874, opened 2025-01-27
Allow URLs up to 32KB and improve parsing speed

### LINKED ISSUE #20873: Pre-signed URL won't work with tokens being too big
## NOTE
If this case is urgent, please subscribe to [Subnet](https://min.io/pricing) so that our 24/7 support team may help you faster.

## Expected Behavior

Pre-signed URL should still be working for (in our case) download.

## Current Behavior

 Won't work, throwing a "XMinioInvalidResourceName" error with the following message : "Resource name contains bad components such as ".." or "."".

## Possible Solution

On April 2024, a fix to limit the URL to be higher than the max size Linux for a path was introduced : https://github.com/minio/minio/pull/19466/files#diff-28dd2e2e4af0c5a25600cd06909e22a80dc4ad56d203f29f3073110d12240b3d

This also prevents to inject any parameter in the URL, such as the ones required for pre-signed URL. What about limiting the check until the first '?' token is read ?

1. Create a RBAC that contains quite a big token
2. Try to download using a pre-signed URL 

## Context

Just downloading files and sending an URL to the browser and automatically download it without setting headers.

klauspost:
## Description

Before/after...
```
Benchmark_hasBadPathComponent/long-32          	   43936	     27232 ns/op	 146.89 MB/s	   32768 B/op	       1 allocs/op
Benchmark_hasBadPathComponent/long-32          	   89971	     13242 ns/op	 302.07 MB/s	       0 B/op	       0 allocs/op
```

Fixes #20873

## How to test this PR?

Included benchmark, otherwise see #20873

## Types of changes
- [x] Fixes a regression from #19466

### DIFF
--- a/cmd/generic-handlers.go
+++ b/cmd/generic-handlers.go
@@ -293,12 +293,6 @@ func parseAmzDateHeader(req *http.Request) (time.Time, APIErrorCode) {
 	return time.Time{}, ErrMissingDateHeader
 }
 
-// Bad path components to be rejected by the path validity handler.
-const (
-	dotdotComponent = ".."
-	dotComponent    = "."
-)
-
 func hasBadHost(host string) error {
 	if globalIsCICD && strings.TrimSpace(host) == "" {
 		// under CI/CD test setups ignore empty hosts as invalid hosts
--- a/cmd/generic-handlers.go
+++ b/cmd/generic-handlers.go
@@ -22,11 +22,11 @@ import (
 	"net"
 	"net/http"
 	"path"
-	"path/filepath"
 	"runtime/debug"
 	"strings"
 	"sync/atomic"
 	"time"
+	"unicode"
 
 	"github.com/dustin/go-humanize"
 	"github.com/minio/minio-go/v7/pkg/s3utils"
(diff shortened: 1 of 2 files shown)


harshavardhana (approved):

cmd/generic-handlers.go
```diff
@@ -311,21 +305,41 @@ func hasBadHost(host string) error {
...
+	}
+
+	for i < n {
+		// Find the next segment
+		start := i
+		for i < n && path[i] != SlashSeparatorChar && path[i] != '\\' {
+			i++
+		}
+
+		// Trim whitespace of segment
```
marktheunissen, line 326: @klauspost Was the trimming of whitespace a new addition in this PR? https://github.com/minio/minio/issues/21022 reports that S3 considers whitespace before and after `.` to be valid:
```
$ echo hello | aws --profile jrobinso-aws s3 cp - s3://test-jrobinso/\.\ /hello
$ aws --profile jrobinso-aws s3 ls --recursive s3://test-jrobinso/
2025-03-12 11:04:55          6 . /hello
```

```
$ echo hello | aws --profile test-elm s3 cp - s3://test-jrobinso/\.\ /hello
upload failed: - to s3://test-jrobinso/. /hello An error occurred (XMinioInvalidResourceName) when calling the PutObject operation: Resource name contains bad components such as ".." or ".".
```

klauspost: Old code had `path = filepath.ToSlash(strings.TrimSpace(path))`.

klauspost: I don't really care about the case in the linked PR. AWS also says to not use silly object names.

thiell: @klauspost  it came from an actual research user data at Stanford, likely a program creating those weird `. ` path components, that work with AWS, but not when we tested migrating them to MinIO. Just wanted to emphasis that while it looks silly, it is based on actual user data and supported by AWS.

